PR #4987, opened 2017-12-12
Correct highlighted new code that needs to be added

mattsonnhalter:
In the highlighted portion of new code to be added for the section "Display related enrollments on the Details page", the previous 3 lines (32-34) are not highlighted but they are new lines of code that the user should add.


scottaddie:
@mattsonnhalter Very nice catch! Thank you for fixing this.

PR #5013, opened 2017-12-15
Update hosting doc

### LINKED ISSUE #5011: How to set Detailed Errors with environment other than shotgun Development
Per @glennc in #4685
> I noticed that the [detailedStartupErrors ](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/hosting?tabs=aspnetcore2x#detailed-errors) docs I could find never talked about being able to set it via an environment variable, only showed how to do it in code.

Update doc to show how to set with environment directly. Add any security caveats.

Rick-Anderson:
Fixes #5011
[Review URL](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/hosting?branch=pr-en-us-5013&tabs=aspnetcore2x)

[@guardrex hated formatting](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/hosting?branch=pr-en-us-5013&tabs=aspnetcore2x#host-configuration-values)

### DIFF
--- a/aspnetcore/fundamentals/hosting.md
+++ b/aspnetcore/fundamentals/hosting.md
@@ -5,7 +5,7 @@ description: Learn about the web host in ASP.NET Core, which is responsible for 
 keywords: ASP.NET Core,web host,IWebHost,WebHostBuilder,IHostingEnvironment,IApplicationLifetime
 ms.author: riande
 manager: wpickett
-ms.date: 12/16/2017
+ms.date: 09/21/2017
 ms.topic: article
 ms.technology: aspnet
 ms.prod: asp.net-core
@@ -21,7 +21,7 @@ ASP.NET Core apps configure and launch a *host*. The *host* is responsible for a
 
 # [ASP.NET Core 2.x](#tab/aspnetcore2x)
 
-Create a host using an instance of [WebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.webhostbuilder). Creating a host is typically performed in the app's entry point, the `Main` method. In the project templates, `Main` is located in *Program.cs*. A typical *Program.cs* calls [CreateDefaultBuilder](/dotnet/api/microsoft.aspnetcore.webhost.createdefaultbuilder) to start setting up a host:
+Create a host using an instance of [WebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.webhostbuilder). This is typically performed in your app's entry point, the `Main` method. In the project templates, `Main` is located in *Program.cs*. A typical *Program.cs* calls [CreateDefaultBuilder](/dotnet/api/microsoft.aspnetcore.webhost.createdefaultbuilder) to start setting up a host:
 
 [!code-csharp[Main](../common/samples/WebApplication1DotNetCore2.0App/Program.cs?name=snippet_Main)]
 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@guardrex  please review what I've done so far. Should I added env var to all the settings? Also search for TODO
I rewrote some long sentences that may not be correct. In general we should avoid `this` when talking about the previous sentence.

guardrex:
I'll TAL later tonight ... I'm buried in an avalanche of cookie auth code. lol :smile:

btw tho ... I avoided fencing those values on purpose.

When they render on the page, they look *blocky-ish*:tm: imo.

capture

... compare to original (less blocky) ...

capture1

Rick-Anderson:
@guardrex  you OK with the following format on every config key?

**Environment variable**: `ASPNETCORE_CAPTURESTARTUPERRORS`

You mention "Values aren't case sensitive". Shouldn;t that be "Keys and values aren't case sensitive"

guardrex:
Yes, I think that's great. I have a paragraph change that moves the content from the *Ordering importance* section higher in the topic that will better surface the fact that env vars can be used to set the values and that the code approach (`UseX` or `UseSetting`) overrides the env vars.

WRT "Keys and values aren't case sensitive" ... yes, that's true ... because configuration the keys are case insensitive. I'll get this in with the commit, too.

guardrex:
Do you have direction RE the *blockification*:tm: of all that fencing?

My feeling is that since the fencing is most valuable when code is inline (and imo they have tooooooo much padding on fenced code), bullet lists and table cell content sometimes render better without fencing. Not all of the time, of course, just sometimes ... when lots of little blocks show up in the rendered content next to each other. It looks ... well ... **bad** (i.e., distracting). lol :smile:

Rick-Anderson:
>Do you have direction RE the *blockification*:tm: of all that fencing?

I think it's more important to be consistent. I think we should stick with `code fencing` types. @scottaddie  - what's your thoughts

guardrex:
I agree with consistency, but I think when it gets distracting it trumps consistency. I hope it could be addressed in review on a case-by-case basis. If the author detects a rendering problem, they can address it for the 1st draft. Management, *you!*, can say in review ... *Gonnae no! Gonnae no do that?* (in your best Scottish accent, of course).

Rick-Anderson:
@guardrex  I'll let you know when you can make changes. I'll work on it for now.

guardrex:
I was hacking around with the paragraphs in the opening of the *Host configuration values* section. which would allow for removal of the *Ordering importance* section. This is just draft content. It needs work, but the idea was to move the content about env vars *up* in the coverage ... before this section actually lists all of the keys/values.

## Host configuration values

\[WebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.webhostbuilder) provides methods for setting most of the available configuration values for the host, which can also be set directly with \[UseSetting](/dotnet/api/microsoft.aspnetcore.hosting.webhostbuilder.usesetting) and the associated key. When setting a value with \`UseSetting\`, the value is set as a string (in quotes) regardless of the type.

Some settings are first read from environment variables. Any configuration can be set by an environment variable. The environment variables use the format \`ASPNETCORE_{Key}\`.

Override any of the environment variable values by specifying configuration (using \`UseConfiguration\`) or by setting the value explicitly (using \`UseSetting\` or one of the explicit extension methods, such as \`UseUrls\`). The host uses whichever option sets the value last. If you want to programmatically set the default URL to one value but allow it to be overridden with configuration, you can use command-line configuration after setting the URL. See \[Overriding configuration](#overriding-configuration).

scottaddie:
The **Key** and **Type** values in the **Capture Startup Errors** section are formatted inconsistently when compared to the other sections of the doc. My vote is for the less blocky version provided by @guardrex above.

guardrex:
I guessed space savings (less vertical scrolling) at first, but since those lines require the double-space ending to break, maybe it was unintended.

Rick-Anderson:
@guardrex  ready for review.

aspnetcore/fundamentals/hosting.md
```diff
@@ -15,7 +15,7 @@ uid: fundamentals/hosting
 
 By [Luke Latham](https://github.com/guardrex)
 
-ASP.NET Core apps configure and launch a *host*, which is responsible for app startup and lifetime management. At a minimum, the host configures a server and a request processing pipeline.
+ASP.NET Core apps configure and launch a *host*. The *host* is responsible for app startup and lifetime management. At a minimum, the host configures a server and a request processing pipeline.
```
guardrex: The emphasis on "host" can be removed after the first use, so I recommend just "The host is ..."

aspnetcore/fundamentals/hosting.md
```diff
@@ -35,25 +35,25 @@ Create a host using an instance of [WebHostBuilder](/dotnet/api/microsoft.aspnet
...
+The *content root* determines where the host searches for content files, such as MVC view files. <!-- TODO  shorten long sentence-->The default content root is [Directory.GetCurrentDirectory](/dotnet/api/system.io.directory.getcurrentdirectory). The default content root (`Directory.GetCurrentDirectory`) results in using the web project's root folder as the content root when the app is started from the root folder (for example, calling [dotnet run](/dotnet/core/tools/dotnet-run) from the project folder). This is the default used in [Visual Studio](https://www.visualstudio.com/) and the [dotnet new templates](/dotnet/core/tools/dotnet-new).
```
guardrex: I don't have push privileges, so I can't address the TODO on this PR. I suggest removing the TODO, and I'll get this long sentence fixed on a PR after this one. I've made a note for Line 41, and I'll note the others, too.

aspnetcore/fundamentals/hosting.md
```diff
@@ -35,25 +35,25 @@ Create a host using an instance of [WebHostBuilder](/dotnet/api/microsoft.aspnet
...
+The *content root* determines where the host searches for content files, such as MVC view files. The default content root supplied to `UseContentRoot` is [Directory.GetCurrentDirectory](/dotnet/api/system.io.directory.getcurrentdirectory?view=netcore-1.1).<!-- TODO  shorten long sentence--> This results in using the web project's root folder as the content root when the app is started from the root folder (for example, calling [dotnet run](/dotnet/core/tools/dotnet-run) from the project folder). This is the default used in [Visual Studio](https://www.visualstudio.com/) and the [dotnet new templates](/dotnet/core/tools/dotnet-new).
```
guardrex: I've noted Line 56, so you can remove the TODO.

aspnetcore/fundamentals/hosting.md
```diff
@@ -77,7 +77,11 @@ When setting up a host, you can provide [Configure](/dotnet/api/microsoft.aspnet
 
 ## Host configuration values
 
-[WebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.webhostbuilder) provides methods for setting most of the available configuration values for the host, which can also be set directly with [UseSetting](/dotnet/api/microsoft.aspnetcore.hosting.webhostbuilder.usesetting) and the associated key. When setting a value with `UseSetting`, the value is set as a string (in quotes) regardless of the type.
+[WebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.webhostbuilder) provides the following approaches for setting most of the available configuration values for the host:
+
+* Explicit methods, such as `CaptureStartupErrors`.
+* [UseSetting](/dotnet/api/microsoft.aspnetcore.hosting.webhostbuilder.usesetting) and the associated key. When setting a value with `UseSetting`, the value is set as a string regardless of the type.
+* Environment variables with the format `ASPNETCORE_{configurationKey}`. For example, `ASPNETCORE_DETAILEDERRORS`.
```
guardrex: Nit: Maybe make the entry for env vars the first one on this list. We can't number them here because the last one wins when it comes to explicit methods and `UseSetting`, but the env var setting is the primary source.

aspnetcore/fundamentals/hosting.md
```diff
@@ -173,7 +180,8 @@ Sets the app's environment.
 **Key**: environment  
 **Type**: *string*  
 **Default**: Production  
-**Set using**: `UseEnvironment`
+**Set using**: `UseEnvironment`  
+**Environment variable**: `ASPNETCORE_ENVIRONMENT`
 
 You can set the *Environment* to any value. Framework-defined values include `Development`, `Staging`, and `Production`. Values aren't case sensitive. By default, the *Environment* is read from the `ASPNETCORE_ENVIRONMENT` environment variable. When using [Visual Studio](https://www.visualstudio.com/), environment variables may be set in the *launchSettings.json* file. For more information, see [Working with Multiple Environments](xref:fundamentals/environments).
```
guardrex: Note that when I make my pass on the next PR, I'll deal with these little probs (e.g., "You").

aspnetcore/fundamentals/hosting.md
```diff
@@ -494,7 +509,7 @@ dotnet run --urls "http://*:8080"
 ## Ordering importance
 
 Some of the `WebHostBuilder` settings are first read from environment variables, if set. These environment variables use the format `ASPNETCORE_{configurationKey}`. To set the URLs that the server listens on by default, you set `ASPNETCORE_URLS`.
-
+<!-- TODO  shorten long sentence-->
```
guardrex: Noting line 512 for next PR.

aspnetcore/fundamentals/hosting.md
```diff
@@ -813,8 +828,8 @@ The [IApplicationLifetime interface](/aspnet/core/api/microsoft.aspnetcore.hosti
 | Cancellation Token    | Triggered when&#8230; |
 | --------------------- | --------------------- |
 | `ApplicationStarted`  | The host has fully started. |
-| `ApplicationStopping` | The host is performing a graceful shutdown. Requests may still be processing. Shutdown blocks until this event completes. |
-| `ApplicationStopped`  | The host is completing a graceful shutdown. All requests should be completely processed. Shutdown blocks until this event completes. |
+| `ApplicationStopping` | The host is performing a graceful shutdown. Requests may still be processing. Shut down blocks until this event completes. |
```
guardrex: "shut down" ... on the verb usage above (Line 692). :+1:

This one and the one on Line 832: Shouldn't the noun form (a time) remain one word? "Shutdown"

guardrex:
LGTM :shipit: :rocket:

Issue #2846, opened 2017-02-27
Describe language fallback behavior


ryanbrandenburg:
The [localization](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/localization) docs don't mention how we do language fallback for resources, we might want to include a short blurb about that behavior.

Rick-Anderson:
@ryanbrandenburg  can you send me a paragraph on that and I'll get it added.

Rick-Anderson:
@ryanbrandenburg can you send me a paragraph on that and I'll get it added.

PR #5041, opened 2017-12-19
Mention culture fallback

rachelappel:
#2846


rachelappel:
@Rick-Anderson  This paragraph on culture fallback was in there already, but was worth splitting up as its own small section, given the request in #2846.

PR #4909, opened 2017-12-01
Update guidance on creating the Logs folder

guardrex:
Addresses https://github.com/aspnet/websdk/issues/152#issuecomment-348569871
Related: https://github.com/aspnet/AspNetCoreModule/issues/30

Based on the latest guidance for creating and deploying the *Logs* folder :sparkles: auto-magically :sparkles: that covers all manner of deployment strategies.

cc/ @vijayrkn @dasmulli

Thanks to @DaleMckeown and @gpcaretti.

PR #5051, opened 2017-12-20
Refactor instructions on logs folder creation

guardrex:
Noticed in passing that this would need an update for latest guidance (https://github.com/aspnet/Docs/pull/4909). However, it's probably best to keep the instructions in one spot, the *Directory structure* topic, and link to it from here.


Rick-Anderson:
LGTM. @guardrex  does this need additional review/work?

guardrex:
@Rick-Anderson This small change should be fine. We already had #4909 approved, so it should be good to go.

PR #5067, opened 2017-12-22
better write up for warning

yaananth:
see https://github.com/dotnet/docs/pull/4002


Rick-Anderson:
@yaananth  Thanks for the follow up.

PR #5070, opened 2017-12-23
Updated the sample3

rmainhar:
The sample 3 now reflects what the Default.aspx should look like


Rick-Anderson:
@rmainhar  Thanks!

PR #5079, opened 2017-12-29
Fixed confusing but harmless code that had no effect on the Page() display

user135711:
In the razorpages CU-8 there is misleading code that does not do anything in the handleDelete method. Deleted the method because the values displayed come from the ModelState not the Department instance.


user135711:
Don't know what the error means. Builds and runs fine on my machine. :( "Some docset configuration inside publish config has ovelapping output folder. Overlapping output folders: aspnet/ and aspnet/core/"

Rick-Anderson:
Need to close/reopen to fix glitch.

Rick-Anderson:
@user135711  looks great, Thanks!

PR #5089, opened 2018-01-02
add notice regarding undocumented methods notices and updated code

CoolGoose:
Enabling the XML documentation will throw __Missing XML comment for publicly visible type or member__ so I thought it's important to give a heads up in the documentation regarding that. Not sure if referencing this [stack overflow thread](https://stackoverflow.com/questions/203863/missing-xml-comment-for-publicly-visible-type-or-member) is something that could be done in the documentation (or maybe create tab style information area on how to ignore them per editor?)

Also I've updated the reading code to be explicit regarding Path.Combine (so you can just copy paste the code) and also replaced the directory path code PlatformServices.Default.Application.ApplicationBasePath using System.AppContext.BaseDirectory since [according to](http://michaco.net/blog/EnvironmentVariablesAndConfigurationInASPNETCoreApps) that's legacy net 4 code and less future proof.

### DIFF
--- a/aspnetcore/tutorials/web-api-help-pages-using-swagger.md
+++ b/aspnetcore/tutorials/web-api-help-pages-using-swagger.md
@@ -223,7 +223,7 @@ See Visual Studio Code.
 
 ---
 
-* Remember that enabling XML comments will give a notice when debugging in case not all of your methods are documented: __Missing XML comment for publicly visible type or member__
+Enabling XML comments provides debug information for undocumented public methods: __Missing XML comment for publicly visible type or member__
 
 Configure Swagger to use the generated XML file. For Linux or non-Windows operating systems, file names and paths can be case sensitive. For example, a *ToDoApi.XML* file would be found on Windows but not CentOS.
 
--- a/aspnetcore/tutorials/web-api-help-pages-using-swagger.md
+++ b/aspnetcore/tutorials/web-api-help-pages-using-swagger.md
@@ -223,8 +223,7 @@ See Visual Studio Code.
 
 ---
 
-Enabling XML comments provides debug information for undocumented public types and members.
-Undocumented types and members are indicated by the warning message: *Missing XML comment for publicly visible type or member*.
+Enabling XML comments provides debug information for undocumented public types and members. Undocumented types and members are indicated by the warning message: *Missing XML comment for publicly visible type or member*.
 
 Configure Swagger to use the generated XML file. For Linux or non-Windows operating systems, file names and paths can be case sensitive. For example, a *ToDoApi.XML* file would be found on Windows but not CentOS.
 
--- a/aspnetcore/tutorials/web-api-help-pages-using-swagger.md
+++ b/aspnetcore/tutorials/web-api-help-pages-using-swagger.md
@@ -223,7 +223,8 @@ See Visual Studio Code.
 
 ---
 
-Enabling XML comments provides debug information for undocumented public methods: __Missing XML comment for publicly visible type or member__
+Enabling XML comments provides debug information for undocumented public types and members.
+Undocumented types and members are indicated by the warning message: *Missing XML comment for publicly visible type or member*.
 
 Configure Swagger to use the generated XML file. For Linux or non-Windows operating systems, file names and paths can be case sensitive. For example, a *ToDoApi.XML* file would be found on Windows but not CentOS.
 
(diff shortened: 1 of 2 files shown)


aspnetcore/tutorials/web-api-help-pages-using-swagger.md
```diff
@@ -223,6 +223,8 @@ See Visual Studio Code.
 
 ---
 
+Enabling XML comments provides debug information for undocumented public methods: __Missing XML comment for publicly visible type or member__
```
Rick-Anderson: I don't understand 
`__Missing XML comment for publicly visible type or member__`

CoolGoose: That is the error that shows up when  you start the debugger by pressing f5 when you don't have all of your methods with XML comments (or at least that is the error I was getting in VS code)

aspnetcore/tutorials/web-api-help-pages-using-swagger/sample/TodoApi/Startup.cs
```diff
@@ -30,8 +30,8 @@ public void ConfigureServices(IServiceCollection services)
                 });
 
                 // Set the comments path for the Swagger JSON and UI.
-                var basePath = PlatformServices.Default.Application.ApplicationBasePath;
-                var xmlPath = Path.Combine(basePath, "TodoApi.xml"); 
+                var basePath = System.AppContext.BaseDirectory;
+                var xmlPath = System.IO.Path.Combine(basePath, "TodoApi.xml");
```
guardrex: Note that the `System.IO` namespace is in the `using` statements at the top.

CoolGoose: @guardrex I missed that, mostly because on the [documentation page](https://docs.microsoft.com/en-us/aspnet/core/tutorials/web-api-help-pages-using-swagger) in that specific area the head of the file itself isn't shown. 

I guess the best would be to show the use statements at the top somehow

guardrex: Generally, we don't show them and leave it to the reader to follow the "View sample code" link. It's just done this way to keep the length of the code as short and focused as possible to aid in keeping the overall topic as short as possible. It's the same deal with namespaces (and frequently class declarations). You'll notice that the `#region snippet_ConfigureServices` here falls under both, and that's where the code will start to display in the topic.

CoolGoose:
Understood, are the other changes ok? I'll open another merge request in this case

guardrex:
Can you make a commit to your `CoolGoose:swagger_update` branch? The commit will show up here after you push the commit to your branch.

Just a suggestion: You could explain what the error means. For example, something along the lines of ...

> Enabling XML comments provides debug information for undocumented public types and members. Undocumented types and members are indicated by the warning message: \*Missing XML comment for publicly visible type or member\*.

... which will render as ...

> Enabling XML comments provides debug information for undocumented public types and members. Undocumented types and members are indicated by the warning message: *Missing XML comment for publicly visible type or member*.

Note that I replaced "methods" with "types and members."

That's just an example. I show it with a pair of asterisks around the warning for emphasis. When I see a line like that put into inline code with backticks, it makes the text look "blocky" to me, so I suggest italics. They may advise you differently tho. I'm just making a suggestion.

RE: Linking to SO. I just did a search to see how often it has been done. Yes, there are 47 instances of links to SO where the author wanted to tie in SO information. Whether or not they'll want you to do that tho is for them to say. They'll let you know.

aspnetcore/tutorials/web-api-help-pages-using-swagger/sample/TodoApi/Startup.cs
scottaddie: Agreed that we shouldn't include the inline namespaces. It clutters the code samples, and the user can access the working sample app if needed.

I recommend the following changes:
- Add `using System;` to the top of *Startup.cs*, and remove `System.` from the following line:
`var basePath = System.AppContext.BaseDirectory;`
- Remove `System.IO.` from the following line: `var xmlPath = System.IO.Path.Combine(basePath, "TodoApi.xml");

scottaddie:
@CoolGoose Thank you for your time and patience on this PR!

CoolGoose:
@scottaddie thank you 🍻

PR #5098, opened 2018-01-03
Fix text that was incorrectly interpreted as HTML

jennings:
The text `<languagecode2>` was interpreted as an HTML tag, so it was not visible [on the docs.microsoft.com page](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/localization#globalization-and-localization-terms) after converting Markdown to HTML.

Since the examples later in the paragraph are formatted with backticks, I added backticks here so it wouldn't be interpreted as HTML.


scottaddie:
@jennings Thank you for correcting this, and congrats on your 1st contribution to this repo!

PR #5101, opened 2018-01-04
Add port to to the header in Nginx config

mne1991:
Using `$host` only in HTTP header in Nginx config causes that Kestrel only knows that the host name is *hostname.com* **without any port**. That causes a problem in **redirecting** (e.g. to login page). The user will be directed to the hostname without any port (i.e. to 80 or 443 according to the schema HTTP or HTTPS respectively).

I added `$server_port` to HTTP request header in Nginx config, so the host name will become *hostname.com:8080*. That solves any related redirecting problems.
  
I am not sure if this is related to any opened issue.

### DIFF
--- a/aspnetcore/publishing/linuxproduction.md
+++ b/aspnetcore/publishing/linuxproduction.md
@@ -116,7 +116,7 @@ server {
         proxy_http_version 1.1;
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection keep-alive;
-        proxy_set_header Host $host:$server_port;
+        proxy_set_header Host $http_host;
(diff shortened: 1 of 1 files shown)


Tratcher:
https://stackoverflow.com/questions/15414810/whats-the-difference-of-host-and-http-host-in-nginx
$http_host looks more accurate for this scenario.

mne1991:
Correct, I discovered today early. I commited the config with `$http_host`.

scottaddie:
@mne1991 Thank you for bringing this to our attention and for correcting it! Also, congratulations on your first contribution to this repo.

PR #5107, opened 2018-01-04
Removing unnecessary code

BugAgain:
We already have an `using` at the top


scottaddie:
@BugAgain Thank you for cleaning up this code sample as well!

PR #5108, opened 2018-01-04
Removing unnecessary code

BugAgain:
In the same folder. No need full path.


scottaddie:
@BugAgain Thank you for cleaning up this code sample. We really appreciate your help!

PR #5118, opened 2018-01-05
ASP.NET Core Identity sample

rachelappel:
Sample code for #5102


Noor1978:
Dear  Rachelappel please send me a project with perfect authentication i mean with user assign to role and showing full administration page like in Visual studio 2012

rachelappel:
@Noor1978 The sample you would like is outside the scope of an introduction article. If it fits with another doc, I'll open an issue to use it there as that sample.

PR #5129, opened 2018-01-07
Add example to ArgumentException remark

guardrex:
Reacting to a reader comment:

> I can't figure out what "injecting IStartup directly into the dependency injection container" means.

Adding this inline code example should clarify what is meant there.

### DIFF
--- a/aspnetcore/fundamentals/hosting.md
+++ b/aspnetcore/fundamentals/hosting.md
@@ -869,7 +869,17 @@ public class Startup
 
 **Applies to ASP.NET Core 2.0 Only**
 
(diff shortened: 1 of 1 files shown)


guardrex:
@scottaddie How about this format?

scottaddie:
@guardrex Looks much better! Thank you for your help.

guardrex:
> Thank you for your help.

Oh yeah .... oooops ... I forgot I wasn't supposed to be submitting these types of PRs. Let's tell Rick that hackers took over my PC and that *they* are the ones submitting these PRs. *Bad hackers! Bad hackers!* 👊

PR #5150, opened 2018-01-10
Uniform style

SemperPeritus:
Uniform style


scottaddie:
@SemperPeritus Very nice catch! Thank you for taking the time to fix it.

PR #5187, opened 2018-01-16
Razor Pages - Added explanations about the use of `return Page();`

### LINKED ISSUE #4606: Explain return Page()
Per customer: 
OnGet of IndexModel does not return Page() but its page is still rendered and populated with a list of Movie (if any). So what does the return Page() mean? I am totally a newbie.

From https://docs.microsoft.com/en-us/aspnet/core/tutorials/razor-pages/page

Per Ryan

return Page() is the default behavior. This is meant for cases where you don’t do any control flow and just want to set a property or something.

If you are doing control flow, then your handler method will be declared to return IActionResult, so you need `return Page()` so that you have an action result to return. Return null also should work

-- end or ryan

So, for example, some methods have `return RedirectToPage("./Index")`

sputier:
Fixes #4606


Rick-Anderson:
@sputier  great addition. I reworded it a bit.

PR #5193, opened 2018-01-17
Update sample4.cs

badreddine-dlaila:
I think  Dictionary<T, HashSet<string>> _connections should be static.
Otherwise, no records will be saved


BrennanConroy:
The Hub that uses `ConnectionMapping` should make its variable static, the internals of the `ConnectionMapping` class should not be static

scottaddie:
@badreddine-dlaila Your efforts are appreciated; however, I'm closing this PR based on Brennan's recommendation above ☝️.

PR #5201, opened 2018-01-18
Remove unneeded references to VB

Eilon:
Because there is no VBHTML in ASP.NET Core.


scottaddie:
@Eilon Nice catch! Thank you for fixing it.

PR #5220, opened 2018-01-19
Adding a link to the VS snapshot debugging topic

Mikejo5000:
Link to new VS topic on snapshot debugging ASP.NET Core app on Azure App Service.


Rick-Anderson:
@Mikejo5000  Thanks Mike. Do you want to add this to the https://docs.microsoft.com/en-us/aspnet/core/testing/ page? If not, I can add it.

PR #5228, opened 2018-01-20
Update adding-controller2.md

chanmmn:
Just to let the readers know which Method is app.UseMvc belongs to.


Rick-Anderson:
@chanmmn  thanks.
You set the format for routing in the `Configure` method in *Startup.cs* file.

PR #5231, opened 2018-01-20
Update advanced.md

ecso:
orthographic error in a word


Rick-Anderson:
I think original is correct. Do you want to change it from "file" to "open/find"?

PR #5240, opened 2018-01-22
Update url-rewriting.md

snomad:
Minor cleanup to redirect description, done for the purposes of making the text and idea flow a little better.


guardrex:
I recommend killing the "For example," at the beginning of the new paragraph. That might be my fault anyway, as it should read "For example if `/resource` is ..." (no comma). It should just start with "If ..." now. The paragraph split that u suggest @snomad is fine. I just suggest killing the "For example," bit.

scottaddie:
@snomad Thank you very much for the cleanup work!

PR #5241, opened 2018-01-22
Configuration during startup

### LINKED ISSUE #5230: Explain how to get configuration at startup
@nhwilly asks on the [Options topic](https://docs.microsoft.com/aspnet/core/fundamentals/configuration/options) ...

> I need to get at those [Options] values while still in the ConfigureServices method. ... What am I missing?

This would be a great add imo. I just drop back to a built configuration to take them, and I think that will work well in most cases. If they need to use a new configuration builder for a source that isn't OOB, that's fine. We could note that or show it. At the very least, we prob want to provide some basic guidance tho.

[EDIT] I'm thinking this new section should appear at the bottom of the Configuration topic, but then we link a little blurb about it at the bottom of the Options topic over to that Configuration section to surface it in Options.

@HaoK Thoughts?

```csharp
public class Program
{
    public static IConfiguration Configuration { get; set; }

    public static void Main(string[] args)
    {
        BuildWebHost(args).Run();
    }

    public static IWebHost BuildWebHost(string[] args) =>
        WebHost.CreateDefaultBuilder(args)
            .ConfigureAppConfiguration((hostingContext, config) =>
            {
                Configuration = config.Build();
            })
            .ConfigureServices(services =>
            {
                services.Configure<MyOptions>(Configuration);

                Console.WriteLine(".......................................");
                Console.WriteLine($"{Configuration["subsection:suboption1"]}");
                Console.WriteLine(".......................................");

                services.AddMvc();
            })
            .Configure(app =>
            {
                ...
            })
            .Build();
}
```

guardrex:
Fixes #5230 

Internal Review Topics:
[Startup topic](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/startup?branch=pr-en-us-5241)
[Configuration topic](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/index?branch=pr-en-us-5241&tabs=basicconfiguration#accessing-configuration-during-startup)
[Options topic](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?branch=pr-en-us-5241#accessing-options-during-startup)

I found a spot (conventional `Startup` and convenience startup approaches) in app startup to show this. We already touch the concept with `IHostingEnvironment`. It fits well here.

@HaoK, especially head-check me on ...

> Options can't be accessed during startup without building a service provider, which is an anti-pattern because it requires loading the `IOptions` service in code.


aspnetcore/fundamentals/configuration/index.md
```diff
@@ -389,14 +389,18 @@ Left: 1988
...
+
 ## Additional notes
 
 * Dependency Injection (DI) is not set up until after `ConfigureServices` is invoked.
 * The configuration system is not DI aware.
 * `IConfiguration` has two specializations:
   * `IConfigurationRoot` Used for the root node. Can trigger a reload.
   * `IConfigurationSection` Represents a section of configuration values. The `GetSection` and `GetChildren` methods return an `IConfigurationSection`.
-  * Use [IConfigurationRoot](https://docs.microsoft.com/ dotnet/api/microsoft.extensions.configuration.iconfigurationroot) when reloading configuration or need access to each provider. Neither of these situations are common.
+  * Use [IConfigurationRoot](/dotnet/api/microsoft.extensions.configuration.iconfigurationroot) when reloading configuration or need access to each provider. Neither of these situations are common.
```
scottaddie: need --> needing

guardrex: Let's see if "for" works there.

... I see more "you" and "your" here, too. I may as well 🔪 those while I'm here.

guardrex:
ty @scottaddie ... Ready for another look :eyes:.

HaoK:
@guardrex I wouldn't say its an anti-pattern because of service locator.  Its more that you are in the midst of configuring options and building the services in ConfigureServices, so there's always an ordering issue, in that if the build a service provider and use the options, it wouldn't contain any configures that are added later.  

I would tweak the wording a bit, with the main point being something like 

IOptions are fine to use in Configure (since services have been built), but may be in an inconsistent state if manually created from inside of ConfigureServices

guardrex:
@HaoK 

How does this look?

> `IOptions` can be used in `Configure`, since services are built before the `Configure` method executes. If a service provider is built in `ConfigureServices` to access options, it wouldn't contain any options configurations provided after the service provider is built. Therefore, an inconsistent options state may exist due to the ordering of service registrations.

> Since options are typically loaded from configuration, configuration can be used in startup in both `Configure` and `ConfigureServices`. For examples of using configuration during startup, see the [Application startup](xref:fundamentals/startup) topic.

HaoK:
Sounds good @guardrex

aspnetcore/fundamentals/configuration/index.md
```diff
@@ -88,28 +88,28 @@ Configuration considerations:
...
+* If a colon (`:`) can't be used in environment variables on a system, replace the colon (`:`) with a double-underscore (`__`).
 
 ## In-memory provider and binding to a POCO class
 
 The following sample shows how to use the in-memory provider and bind to a class:
 
 [!code-csharp[Main](index/sample/InMemory/Program.cs)]
 
-Configuration values are returned as strings, but binding enables the construction of objects. Binding allows you to retrieve POCO objects or even entire object graphs.
+Configuration values are returned as strings, but binding enables the construction of objects. Binding allows the retrievial of POCO objects or even entire object graphs.
```
scottaddie: "retrievial" --> "retrieval"

aspnetcore/fundamentals/configuration/index.md
```diff
@@ -88,28 +88,28 @@ Configuration considerations:
...
-* If a colon (`:`) can't be used in environment variables on your system, replace the colon (`:`) with a double-underscore (`__`).
+* **Never** store passwords or other sensitive data in configuration provider code or in plain text configuration files. Don't use production secrets in development or test environments. Specify secrets outside of the project so that they can't be accidentally committed to a source code repository. Learn more about [working with multiple environments](xref:fundamentals/environments) and managing [safe storage of app secrets during development](xref:security/app-secrets).
+* If a colon (`:`) can't be used in environment variables on a system, replace the colon (`:`) with a double-underscore (`__`).
```
scottaddie: I recommend adding an example of where the double underscore may be necessary. Maybe we have this documented somewhere already, in which case you could just link to that info.

aspnetcore/fundamentals/configuration/index.md
```diff
@@ -88,28 +88,28 @@ Configuration considerations:
...
+Configuration values are returned as strings, but binding enables the construction of objects. Binding allows the retrievial of POCO objects or even entire object graphs.
 
 ### GetValue
 
 The following sample demonstrates the [GetValue&lt;T&gt;](/dotnet/api/microsoft.extensions.configuration.configurationbinder.get?view=aspnetcore-2.0#Microsoft_Extensions_Configuration_ConfigurationBinder_Get__1_Microsoft_Extensions_Configuration_IConfiguration_) extension method:
 
 [!code-csharp[Main](index/sample/InMemoryGetValue/Program.cs?highlight=31)]
 
-The ConfigurationBinder's `GetValue<T>` method allows you to specify a default value (80 in the sample). `GetValue<T>` is for simple scenarios and does not bind to entire sections. `GetValue<T>` gets scalar values from `GetSection(key).Value` converted to a specific type.
+The ConfigurationBinder's `GetValue<T>` method allows the specification of a default value (80 in the sample). `GetValue<T>` is for simple scenarios and does not bind to entire sections. `GetValue<T>` gets scalar values from `GetSection(key).Value` converted to a specific type.
```
scottaddie: "does not" --> "doesn't"

aspnetcore/fundamentals/configuration/index.md
guardrex: It's in-progress on :point_right: https://github.com/aspnet/Docs/issues/4968

I'll add your comment on the issue referencing this topic.

guardrex:
@scottaddie Got 'um :+1:

scottaddie:
@HaoK Please review one final time.

HaoK:
In the get value sample, there's an extra comment which maybe can be removed, doesn't appear to be any json used in that example

```
// Requires NuGet package
// Microsoft.Extensions.Configuration.Json
```

HaoK:
In Startup topic, you don't actually talk about IHostingStartup at all in here or use it in the examples, yet there's a sentence saying "An alternative to injecting IHostingStartup is"

scottaddie:
@guardrex Please address Hao's feedback ☝️ and then we'll get this merged.

guardrex:
@HaoK ty ... great catch ... that bit was supposed to be referring to the use of `IHostingEnvironment` to base app configuration on the environment ... and going on to say that the dev can use the conventions-based approach if they don't want to set things up that way.

I think I found a 2nd spot where the `Requires NuGet package` comment can be removed.

PR #5249, opened 2018-01-23
Adding missing dotnet new react parameter

mungojam:
Without this, the example won't work


Rick-Anderson:
@mungojam  much appreciated.

PR #5258, opened 2018-01-24
Updates to Razor Pages user data protection doc

scottaddie:
Proposed changes to https://github.com/aspnet/Docs/pull/5190


Rick-Anderson:
Nice work. Do we want to standardize on calling the `PageModel` "Page Model"? Why is it not "page model"?

scottaddie:
@Rick-Anderson I think "page model" probably does make more sense. I'd be okay with standardizing on that.

PR #5270, opened 2018-01-25
Cleaned up a duplicated "the"

csharpfritz:


scottaddie:
@csharpfritz Thank you for taking the time to fix this, Jeff! Your change will go live within the next 24 hours.

PR #5271, opened 2018-01-25
Removed a duplicate "the"

WardDurossette:
Extending my small crusade to eliminate the vile duplicity of "the the".


scottaddie:
@WardDurossette Thank you for fixing this! Also, thank you for tuning in to @csharpfritz's Twitch stream this morning.

PR #5290, opened 2018-01-27
Remove code-behind from RP terminology

guardrex:
I avoid using "page model file" in many spots where it was "code-behind file." It seems sufficient to just state "page model."


Rick-Anderson:
That will make david/damien happy.

PR #5295, opened 2018-01-28
Added missing operating system

SampathDilhan:
_Windows 10_ has not been mentioned in the _Supported Platforms_ for SignalR. That has been added.


guardrex:
Thank you for the fix @SampathDilhan! :rocket:

PR #5301, opened 2018-01-29
Fix minor readme typos for AddRedirect function sample.

adelhelal:
Based on the actual code, the following README documentation needs to be consistent.


scottaddie:
@adelhelal Thank you for fixing this!

PR #5329, opened 2018-02-01
Restructuring of the LoggerMessage topic

guardrex:
[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/loggermessage?branch=pr-en-us-5329)

@syscall implies that the implementation details of `Define` and `DefineScope` should appear next to their definitions (and API links).

The reason the topic has its current layout is that the message template applies to both methods and is worthy of highlighting immediately after the concept of a message template is introduced. To avoid duplication of that text, the topic was created with this structure:

1. Define the methods
1. Describe the message template
1. Describe the implementations

**This PR is is test.**

Let's see how it looks if I group the implementations with the definitions and go ahead and duplicate the message template information. I'm not sure it matters all that much either way, but let's take a look and see how it goes.


guardrex:
I agree with @syscall: I think the small duplication for message template is worth it.

PR #5341, opened 2018-02-02
Fix listener initdata value for Scaleout and Transport log

AndreasHassing:


scottaddie:
@AndreasHassing Thank you for fixing this!

PR #5352, opened 2018-02-03
Fix bold character entities and slashes

guardrex:


Rick-Anderson:
LGTM
We'll find out when we get the next MT build report.

PR #5369, opened 2018-02-05
M17961: Potential missing verbe: "manage"

carolinacmoravia:
Hello, @Rick-Anderson, 

Translator has reported possible source content issue.  There is no verb between 'and to' and 'unit test' the pages.

Please, help to check my proposed file change into the article and help to merge if you agree with fix. If not, please, let me know either if you would like me to fix it in another way within this PR, if you prefer to fix it in another PR, or if I should close this PR as by-design. In case of using another PR, please, let me know of your PR number, so we can confirm and close this PR. 

Many thanks in advance.

  Fixes #Issue_Number


Rodigee:
It looks like this isn't an issue since "test" is a verb in this instance, not a noun.

PR #5384, opened 2018-02-07
Typo in response caching

agriffard:
A client can submit a request that's handled by any server in the group and cached data for the client is available

 if available =>  is available


Rick-Anderson:
@agriffard  Thanks!

guardrex:
That looks like a list of two items:
```
A client can submit a request that's ...
    (1) handled by any server in the group
        and
    (2) cached data for the client is available.
```
I think the "if" breaks the structure of the 2nd list item. It could be rephrased for "if" ...

> A client can submit a request that's handled by any server in the group if cached data for the client is available.

PR #5395, opened 2018-02-08
Update SPA templates references to 2.0.0-rc2-final

SteveSandersonMS:
Hopefully the final release will be in the next week, but in the meantime it's useful to update the docs so people try out the new RC2 build.


scottaddie:
Thank you for the update, @SteveSandersonMS!

PR #5423, opened 2018-02-09
Typo was corrected

zavrakv:
No issue number, just saw a typo on a page


zavrakv:
It's ok :) noticed that while copying and thought that it would be nice to tell you )
Great article btw! helps a lot)

Issue #3352, opened 2017-05-11
Update hosted service topic


danroth27:
See Document IHostStartup (host lightup) #4438

To do background tasks in ASP.NET Core you use IHostedService. We should document that.
[meeting with chris/fowler](https://www.dropbox.com/s/elmr1hg92fzku59/IHostedService%20doc%20-%20Friday%2C%20September%2029%2C%202017%202.36.31%20PM.mp4?dl=0)

Notes;
Low level component for running long running tasks.
Runs a jobs once at startup and runs for the life of the app. If you want to get scoped dependencies - you need to create a scope per invocation. 
Not something you create/destroy while app is running.  It's not [QBWI](https://blogs.msdn.microsoft.com/webdev/2014/06/04/queuebackgroundworkitem-to-reliably-schedule-and-run-background-processes-in-asp-net/) - although it is designed for long running processes. 
Sample goes with controller, not RP. RP are for UI.
14:30 IHostedService jobs don't run in parallel with application. `StartAsync` should return quickly so it doesn't block the web app from starting.  `return Task.CompletedTask;` needs to return when the task is started, not when the task is completed.  So you can't return a task of a long running operation in `StartAsync` .

Rick-Anderson:
@danroth27 which dev should we work with?

danroth27:
@muratg Who can help us out for this doc?

Rick-Anderson:
@muratg can we get a sample doing this?

muratg:
@Rick-Anderson Chris can help with it after Preview2 work winds down.

muratg:
cc @Tratcher

Rick-Anderson:
@Tratcher can you come up with a sample app?

Tratcher:
```
using System;
using System.Collections.Generic;
using System.Linq;
using System.Threading;
using System.Threading.Tasks;
using Microsoft.AspNetCore.Builder;
using Microsoft.AspNetCore.Hosting;
using Microsoft.AspNetCore.Http;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;

namespace WebApplication87
{
    public class Startup
    {
        // This method gets called by the runtime. Use this method to add services to the container.
        // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
        public void ConfigureServices(IServiceCollection services)
        {
            services.AddSingleton<IHostedService, MyBackgroundServiceA>();
[...]

PR #5440, opened 2018-02-11
IHostedService topic

guardrex:
Addresses #3352 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/ihostedservice?branch=pr-en-us-5440)


guardrex:
@Tratcher Four questions:

1. Regarding consuming a scoped service in the hosted service: I'm a little concerned about how the word "scoped" is used in this context and the ordinary use of the word (i.e., a service that's activated per request). Here, `ScopedProcessingService`'s `DoWork` method is called once regardless of repeated requests to the app but can have scoped services injected. Being a RP app, I don't want anyone to think that the service is going to get executed per request. Am I being overly paranoid about potentially confusing terminology?
2. `StopAsync` is triggered when the host is performing a graceful shutdown, but a graceful shutdown doesn't seem to include exiting the debugger in VSC. I'm not getting console output from the hosted services in `StopAsync`. Is there a timing issue in writing those messages when stopping the debugger, or is that kind of app shutdown considered *ungraceful* as far as the host is concerned (i.e., `StopAsync` isn't even called)? Whatever the case, do we need to address it here? I'm just thinking in terms of the reader's expectations when they run the sample. I'd like to see those `StopAsync` messages surface for the reader when they run the sample.
3. In `QueueHostedService` for `catch (Exception) { } // TODO: Log`: To get an `ILogger` in there, do I need to use ye 'olde service locator for it? ... I can't inject it into the `QueuedHostedService` without getting a big :boom:.
4. Can u think of one more additional resource for the *Additional resources* section? The **dreaded single-bullet list** 😠 is a horrible construction imo. :smile: lol

I'm done with the 1st draft. It's ready for a first look :eye:.

aspnetcore/fundamentals/ihostedservice.md
```diff
@@ -0,0 +1,78 @@
...
+
+## IHostedService interface
+
+Hosted services implement the [IHostedService](/dotnet/api/microsoft.extensions.hosting.ihostedservice) interface. The interface defines two methods for objects that are managed by the host:
+
+* [StartAsync(CancellationToken)](/dotnet/api/microsoft.extensions.hosting.ihostedservice.startasync) - Called after the server has started and [IApplicationLifetime.ApplicationStarted](/dotnet/api/microsoft.aspnetcore.hosting.iapplicationlifetime.applicationstarted) is triggered. `StartAsync` contains the logic to start the background task.
+
+* [StopAsync(CancellationToken)](/dotnet/api/microsoft.extensions.hosting.ihostedservice.stopasync) - Triggered when the host is performing a graceful shutdown. `StopAsync` contains the logic to end the background task and dispose of any unmanaged resources.
+
+The hosted service is activated once at app startup and gracefully shutdown at app shutdown.
```
Tratcher: They may also Implement IDisposable and be disposed when the service container is disposed. Note StopAsync may not be called in error cases but Dispose should be.

aspnetcore/fundamentals/ihostedservice.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Timed background tasks
+
+A timed background task makes use of the [System.Threading.Timer](/dotnet/api/system.threading.timer) class. The timer triggers the task's work and is disposed when the service stops:
```
Tratcher: Implement IDisposable, stop the timer in StopAsync, and dispose it in Dispose.

aspnetcore/fundamentals/ihostedservice.md
```diff
@@ -0,0 +1,78 @@
...
+
+[!code-csharp[](ihostedservice/sample/Services/TimedHostedService.cs?name=snippet1&highlight=9,23)]
+
+The service is registered in `Startup.ConfigureServices`:
+
+[!code-csharp[](ihostedservice/sample/Startup.cs?name=snippet1)]
+
+## Consuming a scoped service in a background task
+
+Since an `IHostedService` is started when the app starts, the service is activated only *once*. To use dependency injection with a hosted service, a scope must be created during execution. The scope is used resolve an independently registered service where dependency injection can be used.
```
Tratcher: To use scoped services within an IHostedService a scope must be created. No scope is needed for a hosted service by default.

aspnetcore/fundamentals/ihostedservice.md
```diff
@@ -0,0 +1,78 @@
...
+
+The scoped background task service contains the background task's logic. In the following example, [IHostingEnvironment](/dotnet/api/microsoft.aspnetcore.hosting.ihostingenvironment) is injected into the service:
+
+[!code-csharp[](ihostedservice/sample/Services/ScopedProcessingService.cs?name=snippet1&highlight=8,12,18)]
+
+The hosted service creates a scope to resolve the scoped background task service to call its `DoWork` method:
```
Tratcher: The ScopedHostedService name is confusing, the hosted service isn't scoped, it consumes a scoped service. The example is strange too. If this were the timer example the scope would be created and destroyed inside of DoWork, not in StartAsync.

aspnetcore/fundamentals/ihostedservice.md
```diff
@@ -0,0 +1,78 @@
...
+
+[!code-csharp[](ihostedservice/sample/Services/ScopedHostedService.cs?name=snippet1&highlight=14-20)]
+
+The services are registered in `Startup.ConfigureServices`:
+
+[!code-csharp[](ihostedservice/sample/Startup.cs?name=snippet2)]
+
+## Queued background tasks
+
+A background task queue is built based on .NET 4.x [QueueBackgroundWorkItem](/dotnet/api/system.web.hosting.hostingenvironment.queuebackgroundworkitem) ([tentatively scheduled to be built-in for ASP.NET Core 2.2](https://github.com/aspnet/Hosting/issues/1280)):
```
Tratcher: is ~~built~~ based on

aspnetcore/fundamentals/ihostedservice.md
```diff
@@ -0,0 +1,78 @@
...
+
+[!code-csharp[](ihostedservice/sample/Services/BackgroundTaskQueue.cs?name=snippet1)]
+
+In `QueueHostedService`, background tasks (`workItem`) in the queue are dequeued and executed:
+
+[!code-csharp[](ihostedservice/sample/Services/QueuedHostedService.cs?name=snippet1&highlight=28-29,33)]
+
+The services are registered in `Startup.ConfigureServices`:
```
Tratcher: You don't show how to queue items. E.g. Inject the IBackgroundTaskQueue service into a controller and call it from an action.

Tratcher:
1. Your usage of the term Scoped is correct, it's a general DI concept. I did recommend changing the scope sample to resolve the scope per DoWork iteration. An example where you may do this is with a DBContext.
2. Exiting the debugger is the least graceful shutdown possible. Finalizers don't even get to run, it kills the process immediately at an OS level.
3. Add an ILoggerFactory to QueuedHostedService's constructor and create your logger there. Or switch to console logs as that's what you do everywhere else.
4. No, that works.

aspnetcore/fundamentals/ihostedservice.md
guardrex: I was trying to name them in a way that pertains to what they're demonstrating; and yes, *it went horribly wrong here.* The reason the scope creation is in `StartAsync` is because your sample seemed to want it there: See `StartAsync` at https://github.com/aspnet/Docs/issues/3352#issuecomment-333254142. I'll refactor and work on the naming for the next commit.

aspnetcore/fundamentals/ihostedservice.md
guardrex: I did it this way ... I just didn't get that into the topic. I'll drop it in on the next pass. How does it look? ...

https://github.com/guardrex/Docs/blob/7952051df69fc555ade42199d08df41678c82a08/aspnetcore/fundamentals/ihostedservice/sample/Pages/Index.cshtml.cs

guardrex:
> Exiting the debugger is the least graceful shutdown possible. Finalizers don't even get to run, it kills the process immediately at an OS level.

Is there a decent way to demo my console messages from `StopAsync` ... can I trigger a graceful host shutdown from a page model class (i.e., user click in the UI)?

aspnetcore/fundamentals/ihostedservice.md
Tratcher: Fair enough, my sample worked but wasn't sensible.

aspnetcore/fundamentals/ihostedservice.md
Tratcher: Looks good.

Tratcher:
Command line ctl+c is the most common graceful shutdown.

You can do it in app code by injecting the IApplicationLifetime and calling StopApplication().
https://github.com/aspnet/Hosting/blob/99c742221fd9a2ecaef0ec31e63e951a0ee8b5a6/src/Microsoft.Extensions.Hosting.Abstractions/IApplicationLifetime.cs#L35

guardrex:
@Tratcher 2nd draft is ready for a look.

`StopApplication` is great. That does exactly what I was hoping to do with the sample, and I wonder if we should add a short example in the `IApplicationLifetime` interface section of the Hosting topic to better surface its use. Let's take a look a what that might look like. I'll put up a PR so we can take a look.

aspnetcore/fundamentals/ihostedservice.md
```diff
@@ -29,50 +29,59 @@ Hosted services implement the [IHostedService](/dotnet/api/microsoft.extensions.
 
 * [StartAsync(CancellationToken)](/dotnet/api/microsoft.extensions.hosting.ihostedservice.startasync) - Called after the server has started and [IApplicationLifetime.ApplicationStarted](/dotnet/api/microsoft.aspnetcore.hosting.iapplicationlifetime.applicationstarted) is triggered. `StartAsync` contains the logic to start the background task.
 
-* [StopAsync(CancellationToken)](/dotnet/api/microsoft.extensions.hosting.ihostedservice.stopasync) - Triggered when the host is performing a graceful shutdown. `StopAsync` contains the logic to end the background task and dispose of any unmanaged resources.
+* [StopAsync(CancellationToken)](/dotnet/api/microsoft.extensions.hosting.ihostedservice.stopasync) - Triggered when the host is performing a graceful shutdown. `StopAsync` contains the logic to end the background task and dispose of any unmanaged resources. If an error is thrown during background task execution, `StopAsync` might not be called.
```
Tratcher: I meant if the application shuts down un-gracefully due to an error StopAsync may not be called.

guardrex: I felt like we could give them one (obvious) example. If u have a better example, I can swap that in there.

> If the app shuts down ungracefully (for example, the app's process fails), \`StopAsync\` might not be called.

guardrex:
ty @Tratcher 

@scottaddie If ur free, it's ready for a UE pass.

aspnetcore/fundamentals/ihostedservice.md
```diff
@@ -0,0 +1,87 @@
...
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/fundamentals/ihostedservice/samples/) ([how to download](xref:tutorials/index#how-to-download-a-sample))
+
+## IHostedService interface
+
+Hosted services implement the [IHostedService](/dotnet/api/microsoft.extensions.hosting.ihostedservice) interface. The interface defines two methods for objects that are managed by the host:
+
+* [StartAsync(CancellationToken)](/dotnet/api/microsoft.extensions.hosting.ihostedservice.startasync) - Called after the server has started and [IApplicationLifetime.ApplicationStarted](/dotnet/api/microsoft.aspnetcore.hosting.iapplicationlifetime.applicationstarted) is triggered. `StartAsync` contains the logic to start the background task.
+
+* [StopAsync(CancellationToken)](/dotnet/api/microsoft.extensions.hosting.ihostedservice.stopasync) - Triggered when the host is performing a graceful shutdown. `StopAsync` contains the logic to end the background task and dispose of any unmanaged resources. If the app shuts down ungracefully (for example, the app's process fails), `StopAsync` might not be called.
```
scottaddie: I suggest using a word like "abruptly" or "unexpectedly" instead of "ungracefully". For some reason, "ungracefully" sounds awkward.

aspnetcore/fundamentals/ihostedservice.md
```diff
@@ -0,0 +1,87 @@
...
+
+[!code-csharp[](ihostedservice/sample/Services/ConsumeScopedServiceHostedService.cs?name=snippet1&highlight=23-29)]
+
+The services are registered in `Startup.ConfigureServices`:
+
+[!code-csharp[](ihostedservice/sample/Startup.cs?name=snippet2)]
+
+## Queued background tasks
+
+A background task queue is based on .NET 4.x [QueueBackgroundWorkItem](/dotnet/api/system.web.hosting.hostingenvironment.queuebackgroundworkitem) ([tentatively scheduled to be built-in for ASP.NET Core 2.2](https://github.com/aspnet/Hosting/issues/1280)):
```
scottaddie: based on --> based on the

aspnetcore/fundamentals/ihostedservice/sample/BackgroundTasksSample.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.0" />
```
scottaddie: Upgrade to the latest stable version, which I believe is 2.0.5.

aspnetcore/fundamentals/ihostedservice/sample/Pages/Index.cshtml.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        #region snippet2
+        public IActionResult OnPostAddTask()
+        {
+            Queue.QueueBackgroundWorkItem(async token =>
+            {
+                var guid = Guid.NewGuid().ToString();
+
+                for (int i = 0; i < 3; i++)
```
scottaddie: Please use a more descriptive variable name than `i`.

aspnetcore/fundamentals/ihostedservice.md
```diff
@@ -0,0 +1,87 @@
...
+manager: wpickett
+ms.author: riande
+ms.custom: mvc
+ms.date: 02/12/2018
+ms.prod: asp.net-core
+ms.technology: aspnet
+ms.topic: article
+uid: fundamentals/ihostedservice
+---
+# Background tasks with IHostedService in ASP.NET Core
```
scottaddie: I suggest replacing "IHostedService" in the title with "hosted services". Using "hosted services" or something that isn't a cryptic API name seems a bit more inviting.

guardrex:
ty @scottaddie ... excellent advice as usual ... *especially on the topic/UID naming*. I wrote one earlier that I need to go back and check that probably has the same problem. 🙈

scottaddie:
@davidfowl Please review.

aspnetcore/fundamentals/hosted-services.md
```diff
@@ -0,0 +1,87 @@
...
+
+* [StartAsync(CancellationToken)](/dotnet/api/microsoft.extensions.hosting.ihostedservice.startasync) - Called after the server has started and [IApplicationLifetime.ApplicationStarted](/dotnet/api/microsoft.aspnetcore.hosting.iapplicationlifetime.applicationstarted) is triggered. `StartAsync` contains the logic to start the background task.
+
+* [StopAsync(CancellationToken)](/dotnet/api/microsoft.extensions.hosting.ihostedservice.stopasync) - Triggered when the host is performing a graceful shutdown. `StopAsync` contains the logic to end the background task and dispose of any unmanaged resources. If the app shuts down unexpectedly (for example, the app's process fails), `StopAsync` might not be called.
+
+The hosted service is activated once at app startup and gracefully shutdown at app shutdown. When [IDisposable](/dotnet/api/system.idisposable) is implemented, resources can be disposed when the service container is disposed. If an error is thrown during background task execution, `Dispose` should be called even if `StopAsync` isn't called.
```
davidfowl: Can we fit "The hosted service is a singleton" somewhere in here?

guardrex:
@davidfowl Done ... anything else?

aspnetcore/fundamentals/hosted-services.md
```diff
@@ -0,0 +1,87 @@
...
+
+[!code-csharp[](hosted-services/sample/Services/TimedHostedService.cs?name=snippet1&highlight=9,23,30)]
+
+The service is registered in `Startup.ConfigureServices`:
+
+[!code-csharp[](hosted-services/sample/Startup.cs?name=snippet1)]
+
+## Consuming a scoped service in a background task
+
+To use scoped services within an `IHostedService`, create a scope. No scope is needed for a hosted service by default.
```
davidfowl: No scope is created, I wouldn't say needed. That's implied by the fact that it's a single.

aspnetcore/fundamentals/hosted-services.md
```diff
@@ -0,0 +1,87 @@
...
+ms.prod: asp.net-core
+ms.technology: aspnet
+ms.topic: article
+uid: fundamentals/hosted-services
+---
+# Background tasks with hosted services in ASP.NET Core
+
+By [Luke Latham](https://github.com/guardrex)
+
+In ASP.NET Core, background tasks are implemented as *hosted services*. A hosted service is a class with background task logic that implements the [IHostedService](/dotnet/api/microsoft.extensions.hosting.ihostedservice) interface. This topic provides three hosted service examples:
```
davidfowl: background tasks **can be** implemented as hosted services. Is there an article explaining why that is the case? There might also be questions around how these behave when running in IIS (StopAsync will be called when the application pool is being recycled etc).

aspnetcore/fundamentals/hosted-services/sample/Pages/Index.cshtml.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace BackgroundTasksSample.Pages
+{
+    public class IndexModel : PageModel
+    {
+        private readonly IApplicationLifetime _appLifetime;
+
+        #region snippet1
+        public IndexModel(IBackgroundTaskQueue queue, IApplicationLifetime appLifetime)
+        {
+            Queue = queue ?? throw new ArgumentNullException(nameof(queue));
```
davidfowl: Remove the null check

aspnetcore/fundamentals/hosted-services/sample/Pages/Index.cshtml.cs
```diff
@@ -0,0 +1,57 @@
...
+        #region snippet2
+        public IActionResult OnPostAddTask()
+        {
+            Queue.QueueBackgroundWorkItem(async token =>
+            {
+                var guid = Guid.NewGuid().ToString();
+
+                for (int delayLoop = 0; delayLoop < 3; delayLoop++)
+                {
+                    Console.WriteLine(
```
davidfowl: Logging would be better

aspnetcore/fundamentals/hosted-services/sample/Pages/Index.cshtml.cs
```diff
@@ -0,0 +1,57 @@
...
+                    $"{DateTime.UtcNow} - Queued Background Task {guid} is complete. 3/3");
+            });
+
+            return RedirectToPage();
+        }
+        #endregion
+
+        public void OnPostShutdown()
+        {
+            _appLifetime.StopApplication();
```
davidfowl: I wouldn't do this even in a sample, it's basically a way for somebody to shut your site down from a remote client. We can just describe that on exit the hosted service will stop.

aspnetcore/fundamentals/hosted-services/sample/Program.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            BuildWebHost(args).Run();
+        }
+
+        public static IWebHost BuildWebHost(string[] args) =>
+            WebHost.CreateDefaultBuilder(args)
+                .UseStartup<Startup>()
+                .ConfigureLogging((hostingContext, logging) =>
+                {
+                    // Remove the Debug provider (and other providers) and filter most
```
davidfowl: Why are we doing this in a sample? It seems completely unrelated to the content.

aspnetcore/fundamentals/hosted-services.md
guardrex: > Is there an article explaining why that is the case? 

Not that I'm aware of.

> There might also be questions around how these behave when running in IIS (StopAsync will be called when the application pool is being recycled etc).

Currently, the comment speaks more to an "unexpected" shutdown and implies that other types of shutdowns are "graceful" and will trigger `StopAsync`. If you have specific things that you think it should call out, I can list them (e.g., app pool recycling ... is it of the "ungraceful" variety?) ... what else should be listed?

aspnetcore/fundamentals/hosted-services/sample/Services/ConsumeScopedServiceHostedService.cs
```diff
@@ -0,0 +1,49 @@
...
+using Microsoft.Extensions.Hosting;
+
+namespace BackgroundTasksSample.Services
+{
+    #region snippet1
+    internal class ConsumeScopedServiceHostedService : IHostedService
+    {
+        public ConsumeScopedServiceHostedService(IServiceProvider services)
+        {
+            Services = services ?? throw new ArgumentNullException(nameof(services));
```
davidfowl: Remove the null check

aspnetcore/fundamentals/hosted-services/sample/Services/QueuedHostedService.cs
```diff
@@ -0,0 +1,64 @@
...
+    public class QueuedHostedService : IHostedService
+    {
+        private CancellationTokenSource _shutdown = 
+            new CancellationTokenSource();
+        private Task _backgroundTask;
+        private readonly ILogger _logger;
+
+        public QueuedHostedService(IBackgroundTaskQueue taskQueue, ILoggerFactory loggerFactory)
+        {
+            TaskQueue = taskQueue ??
```
davidfowl: Remove the null check everywhere 😄

aspnetcore/fundamentals/hosted-services/sample/Services/TimedHostedService.cs
```diff
@@ -0,0 +1,42 @@
...
+            Console.WriteLine($"{DateTime.UtcNow} - Timed Background Service is stopping.");
+
+            _timer.Change(Timeout.Infinite, 0);
+
+            return Task.CompletedTask;
+        }
+
+        public void Dispose()
+        {
+            _timer.Dispose();
```
davidfowl: You need to handle the timer being null.

aspnetcore/fundamentals/hosted-services/sample/Services/TimedHostedService.cs
```diff
@@ -0,0 +1,42 @@
...
+        private void DoWork(object state)
+        {
+            Console.WriteLine($"{DateTime.UtcNow} - Timed Background Service is working.");
+        }
+
+        public Task StopAsync(CancellationToken cancellationToken)
+        {
+            Console.WriteLine($"{DateTime.UtcNow} - Timed Background Service is stopping.");
+
+            _timer.Change(Timeout.Infinite, 0);
```
davidfowl: You need to handle the timer being null here

aspnetcore/fundamentals/hosted-services/sample/Pages/Index.cshtml.cs
guardrex: Just wanted to demo the `StopAsync` messages so the dev can :eye: see them.

aspnetcore/fundamentals/hosted-services/sample/Program.cs
guardrex: To clean up the console messages so that the messages written by the methods we're showing don't get lost. We do this on the logger message message topic, too, for the same effect.

aspnetcore/fundamentals/hosted-services/sample/Services/TimedHostedService.cs
guardrex: Ok, but that's not what the doc I read said to do. They say to do it using `Change`. Do you want to see the reference?

aspnetcore/fundamentals/hosted-services/sample/Services/TimedHostedService.cs
guardrex: "handle" ... ?? 😕 Chris said to do it this way. Zero (kill) the timer on `StopAsync` then dispose of it on `Dispose`.

aspnetcore/fundamentals/hosted-services/sample/Program.cs
davidfowl: Ugh

aspnetcore/fundamentals/hosted-services/sample/Services/TimedHostedService.cs
davidfowl: What if the host never starts?

aspnetcore/fundamentals/hosted-services/sample/Program.cs
guardrex: The alternative is to pull the lines and let the dev search for the messages. Would u prefer to do that? I'll strike this here and over in the Logger Message topic.

aspnetcore/fundamentals/hosted-services/sample/Services/TimedHostedService.cs
guardrex: > If dueTime is negative one (-1) milliseconds, the callback method is never invoked; the timer is disabled.

I think u know! lol `Timeout.Infinite` = -1

> When a timer is no longer needed, use the Dispose method to free the resources held by the timer.

aspnetcore/fundamentals/hosted-services/sample/Services/TimedHostedService.cs
guardrex: Ur saying this will fail if the host never starts? ... will it throw an ORE or something?

aspnetcore/fundamentals/hosted-services/sample/Services/QueuedHostedService.cs
Tratcher: Why? Just to make it prettier?

aspnetcore/fundamentals/hosted-services/sample/Services/TimedHostedService.cs
Tratcher: `_timer?.Dispose();` It is possible StartAsync is never called. E.g. if another service throws from it's StartAsync.

aspnetcore/fundamentals/hosted-services/sample/Services/TimedHostedService.cs
Tratcher: @davidfowl ? Stop shouldn't be called without Start.

aspnetcore/fundamentals/hosted-services/sample/Services/TimedHostedService.cs
davidfowl: Correct. If you assign something in start then you have to take care of StartAsync never being called on StopAsync and Dispose

aspnetcore/fundamentals/hosted-services/sample/Pages/Index.cshtml.cs
davidfowl: People copy and paste code from docs into their applications.

aspnetcore/fundamentals/hosted-services/sample/Services/TimedHostedService.cs
davidfowl: This is true. It's only if StartAsync fails before the assignment (which can't happen in this case) that we'll still call StopAsync and it would be null.

guardrex: sry ... I understand now.

aspnetcore/fundamentals/hosted-services/sample/Pages/Index.cshtml.cs
guardrex: The comment is supposed to head that off, but ok ... I'll pull it (here and over in Logger Message).

guardrex: Sure ... I'll pull it.

guardrex:
@davidfowl @Tratcher The only outstanding item remaining (I think) is to convert console message writing over to logging. I need to run out for a few hours, but I'll square that away this afternoon and ping u back.

aspnetcore/fundamentals/hosted-services/sample/Services/QueuedHostedService.cs
```diff
@@ -0,0 +1,63 @@
...
+                var workItem = 
+                    await TaskQueue.DequeueAsync(_shutdown.Token);
+
+                try
+                {
+                    await workItem(_shutdown.Token);
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogError(ex, "An error occurred executing the workItem.");
```
scottaddie: Perhaps this is overkill, but I suggest using `nameof`:
```csharp
$"An error occurred executing the {nameof(workItem)}."
```
If the variable name ever changed, this hard-coded string may be missed. This also makes it clearer to the reader what your intent was with that log message.

guardrex:
@scottaddie Safest if you take one last peek. For one thing, I can't see the review site to check the highlights. (I'll be back to building locally shortly.)

aspnetcore/fundamentals/hosted-services/sample/Services/QueuedHostedService.cs
```diff
@@ -44,7 +44,7 @@ public Task StartAsync(CancellationToken cancellationToken)
                 }
                 catch (Exception ex)
                 {
-                    _logger.LogError(ex, "An error occurred executing the workItem.");
+                    _logger.LogError(ex, $"An error occurred executing the {nameof(workItem)}.");
```
davidfowl: Does this work? I'm not sure it does.

guardrex: I'll run a test ... I'll add a line to throw inside the `try` and see what it does.

guardrex: Seems ok ...

```
System.NotSupportedException: Geebers McScreebers!
...
BackgroundTasksSample.Services.QueuedHostedService:Error: Error occurred executing workItem.
```

PR #5441, opened 2018-02-11
Typos in ef-rp/read-related-data.md

agriffard:
For collections it uses... : add comma

Does 'Eager loading for the `Course` entities and their related `Department` entities is used' means 'Eager loading is used for the `Course` entities and their related `Department` entities'?

'related data from the `Enrollments` entity is displayed' => may be 'are displayed' ?

'clicking on the **Select** link for an instructor when the page directive produces a URL like the following': There must be some missing words.

### DIFF
--- a/aspnetcore/data/ef-rp/read-related-data.md
+++ b/aspnetcore/data/ef-rp/read-related-data.md
@@ -122,7 +122,7 @@ The `OnGetAsync` method loads related data with the `Include` method:
 
 [!code-csharp[Main](intro/samples/cu/Pages/Courses/Index.cshtml.cs?name=snippet_RevisedIndexMethod&highlight=4)]
 
-The `Select` operator loads only the related data needed. For single items, like the `Department.Name` it uses a SQL INNER JOIN. For collections, it uses another database access, but so does the .`Include` operator on collections.
+The `Select` operator loads only the related data needed. For single items, like the `Department.Name` it uses a SQL INNER JOIN. For collections, it uses another database access, but so does the `Include` operator on collections.
 
 The following code loads related data with the `Select` method:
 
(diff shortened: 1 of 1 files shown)


guardrex:
> Does 'Eager loading for the Course entities and their related Department entities is used' means 'Eager loading is used for the Course entities and their related Department entities'?

Yes ... I like your version better.

> 'related data from the Enrollments entity is displayed' => may be 'are displayed' ?

Our style manual calls for only using singular verbs with "data." This one is ok; no change.

> 'clicking on the \*\*Select\*\* link for an instructor when the page directive produces a URL like the following': There must be some missing words.

Yes ... the phrasing is a bit messed up. Use this ...

> For example, clicking on the \*\*Select\*\* link for an instructor with only the \`@page\` directive produces a URL like the following:

aspnetcore/data/ef-rp/read-related-data.md
```diff
@@ -122,7 +122,7 @@ The `OnGetAsync` method loads related data with the `Include` method:
 
 [!code-csharp[Main](intro/samples/cu/Pages/Courses/Index.cshtml.cs?name=snippet_RevisedIndexMethod&highlight=4)]
 
-The `Select` operator loads only the related data needed. For single items, like the `Department.Name` it uses a SQL INNER JOIN. For collections it uses another database access, but so does the .`Include` operator on collections.
+The `Select` operator loads only the related data needed. For single items, like the `Department.Name` it uses a SQL INNER JOIN. For collections, it uses another database access, but so does the .`Include` operator on collections.
```
guardrex: That period at ...

> .\`Include\`

That period can go. 🔪

PR #5445, opened 2018-02-12
Mention InvokeHandlersAfterFailure property in policy auth doc

scottaddie:
Fixes https://github.com/aspnet/Docs/issues/5017

[Internal Review Page](https://review.docs.microsoft.com/en-us/aspnet/core/security/authorization/policies?branch=pr-en-us-5445#what-should-a-handler-return)

### DIFF
--- a/aspnetcore/security/authorization/policies.md
+++ b/aspnetcore/security/authorization/policies.md
@@ -80,7 +80,7 @@ Note that the `Handle` method in the [handler example](#security-authorization-h
 
 * To guarantee failure, even if other requirement handlers succeed, call `context.Fail`.
 
-When set to `false`, the [InvokeHandlersAfterFailure](/dotnet/api/microsoft.aspnetcore.authorization.authorizationoptions.invokehandlersafterfailure#Microsoft_AspNetCore_Authorization_AuthorizationOptions_InvokeHandlersAfterFailure) property (available in ASP.NET Core 1.1 and later) short-circuits the execution of handlers when `context.Fail` is called. `InvokeHandlersAfterFailure` defaults to `true`, in which case all handlers are called when a policy requires the requirement. This allows requirements to produce side effects, such as logging, which always take place even if `context.Fail` has been called in another handler.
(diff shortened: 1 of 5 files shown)


HaoK:
Looks good generally, one tweak tho to the wording here:

"case all handlers are called when a policy requires the requirement."

This sentence is a bit misleading, its probably worth calling out somewhere in this document to clarify that all IAuthenticationHandlers are actually called for all policies/requests (except of course for when this flag is set to false).  The base class implementation of `AuthenticationHandler<T>` is what's filtering out requests for requirements that don't match, but its important to caveat this distinction since its not obvious that handlers don't have to be tied to a specific requirements

scottaddie:
> Maybe this should go in the main Authorization handlers section

@HaoK Are you proposing that the newly added `InvokeHandlersAfterFailure` property text move to that section, or are you proposing that I add text like "Handlers don't have to be tied to a specific requirement." to that section?

HaoK:
Tweaking the handlers section to include some text/change the current text to make it clearer that handlers don't have to derive from `AuthorizationHandler<T>` and can instead directly implement IAuthorizationhandler to handle more than one type of requirement.

Was there another PR where we discussed this before?  I feel like there might be another PR where there was a similar discussion

HaoK:
I'm not sure this property needs its own section, you could just mention it off hand in a sentence in the main authorzation handlers section too.  But I'll leave that to you, very few people will care one way or the other about this flag, so that's kinda why I don't really think it necessarily needs its own section

scottaddie:
@HaoK It was discussed here: https://github.com/aspnet/Docs/pull/4849#issuecomment-351813960.

HaoK:
Ah cool okay, if you were planning on doing these changes in a different PR, sounds good

scottaddie:
@HaoK I'll just tackle it here.

scottaddie:
@HaoK Please take another look.

HaoK:
Looks good, one minor thing to clean up, in the invokeHandlers blurb, you can remove the "when a policy requires the requirement" part since all handlers are called, the `handlers<T>` just automatically ignore calls that don't match the requirement they are looking for, but they still called.

PR #5446, opened 2018-02-12
Added code location to options

user135711:
Added the source location of each code sample of ASP.NET Core Fundamentals [Options ](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options)topic to increase readability. In addition, I would suggest adding a code example using IOptionsMonitor, IOptionsFactory and IPotionsMonitorCache.

### DIFF
--- a/aspnetcore/fundamentals/configuration/options.md
+++ b/aspnetcore/fundamentals/configuration/options.md
@@ -30,7 +30,7 @@ An options class must be non-abstract with a public parameterless constructor. T
 
 [!code-csharp[Main](options/sample/Models/MyOptions.cs?name=snippet1)]
 
-The `MyOptions` class is added to the service container with [IConfigureOptions&lt;TOptions&gt;](/dotnet/api/microsoft.extensions.options.iconfigureoptions-1) and bound to configuration:
+The `MyOptions` class is added to the service container with [IConfigureOptions&lt;TOptions&gt;](/dotnet/api/microsoft.extensions.options.iconfigureoptions-1) and bound to configuration (*Startup.cs*, ConfigureServices(IServiceCollection services)):
 
 [!code-csharp[Main](options/sample/Startup.cs?name=snippet_Example1)]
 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@user135711  thanks for this. The document is not a beginning tutorial and dev's will be able to find the right location. Some files like options don't need to go where you specified.

user135711:
Sure. It was just a novice suggestion. On the same lines, whoever is in charge of the livefyre comments should allow the side comments to be deleted by the commentor because it's easy to make mistakes in comments or realize later that something asked is already in the document or obvious. The comments at the bottom are already self deletable, and I don't see why the side comments aren't.

Rick-Anderson:
We've made that suggesting to LiveFyre many times.  We hope to have that resolved in the next couple of months.

user135711:
Also, the reference source link for ioptionsfactory is still broken at the bottom of the options page.

Rick-Anderson:
Which one exactly? they're all working for me.

user135711:
@Rick-Anderson https://github.com/aspnet/Options/blob/release/2.0.0/src/Microsoft.Extensions.Options/OptionsFactory.cs gives me 404

Rick-Anderson:
Thanks. It should be https://github.com/aspnet/Options/blob/dev/src/Microsoft.Extensions.Options/OptionsFactory.cs
can you create a PR with that fix and ask @guardrex  to review it?

user135711:
Yup.

guardrex:
Hold on a sec @user135711 ... I need to check something.

guardrex:
@user135711 Use this one (the current link is missing an "I" for `IOptionsFactory`. We don't want to doc dev features, so it's best to peg the version to 2.0.

https://github.com/aspnet/Options/blob/release/2.0/src/Microsoft.Extensions.Options/IOptionsFactory.cs

user135711:
@guardrex ok

PR #5455, opened 2018-02-13
fixed ioptionfactory.cs source link

user135711:
I used the 2.0 link suggested. The newer version has more than stubs so when the next full release occurs, people interested in how the type works will probably want to read an updated release link.
Please review @guardrex


guardrex:
> want to read an updated release link

That's correct. We pass over all of the topics each release and update the ref source links (*in theory!* :smile:).

PR #5456, opened 2018-02-13
Add example IApplicationLifetime.StopApplication

guardrex:
This surfaces `StopApplication` with a quick RP example. Feels like this method won't be very useful outside of debugging scenarios, but it might be good to show how it can be used to gracefully shutdown an app from a web app UI. Because this probably releases locked files, it might be useful in some deployment scenarios.

### DIFF
--- a/aspnetcore/fundamentals/hosting.md
+++ b/aspnetcore/fundamentals/hosting.md
@@ -860,7 +860,7 @@ public class Startup
 }
 ```
 
-[StopApplication](/dotnet/api/microsoft.aspnetcore.hosting.iapplicationlifetime.stopapplication) requests termination of the app. The following Razor Pages page model class uses `StopApplication` to gracefully shutdown an app when a shutdown button is selected in the UI:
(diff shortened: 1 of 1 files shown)


aspnetcore/fundamentals/hosting.md
```diff
@@ -864,6 +860,34 @@ public class Startup
 }
 ```
 
+[StopApplication](/dotnet/api/microsoft.aspnetcore.hosting.iapplicationlifetime.stopapplication) requests termination of the app. The following Razor Pages page model class uses `StopApplication` to gracefully shutdown an app when a shutdown button is selected in the UI:
```
scottaddie: I suggest explicitly stating that the `OnPostShutdown` method is invoked.

aspnetcore/fundamentals/hosting.md
```diff
@@ -864,6 +860,34 @@ public class Startup
 }
 ```
 
+[StopApplication](/dotnet/api/microsoft.aspnetcore.hosting.iapplicationlifetime.stopapplication) requests termination of the app. The following [Razor Pages](xref:mvc/razor-pages/index) page model class uses `StopApplication` to gracefully shutdown an app. The `OnPostShutdown` method executes after a shutdown button in the UI is selected:
```
scottaddie: a shutdown button --> the **Shutdown** button

PR #5460, opened 2018-02-13
Patch page model method

guardrex:
Patch required. The file was copied into these projects from a modified template without reverting the method name in the template.


scottaddie:
@guardrex Please resolve the conflict, and then this PR is good to merge.

PR #5468, opened 2018-02-14
PageModel has worng name

Covinsky:
According to the text the page file name is "Pages/Index2.cshtml" but in the code it was written "IndexModel2"


Rick-Anderson:
File name doesn't need to match class name, although that's a good convention. If  you want to make this more consistent you'll need to submit 1 PR with all the changes. Each PR by itself breaks the code.

PR #5469, opened 2018-02-14
PageModel Class has worng name

Covinsky:
According to the text the page file name is "Pages/Index2.cshtml" but in the code it was written "public class IndexModel2 : PageModel". So changed it to "public class Index2Model : PageModel".


Rick-Anderson:
see #5468

PR #5470, opened 2018-02-14
Update working-with-forms.md

pranavkm:
Adding a little bit of text to how ModelExpression is translated. Based on docs for https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.rendering.ihtmlhelper.textarea?view=aspnetcore-2.0#Microsoft_AspNetCore_Mvc_Rendering_IHtmlHelper_TextArea_System_String_System_String_System_Int32_System_Int32_System_Object_

### DIFF
--- a/aspnetcore/mvc/views/working-with-forms.md
+++ b/aspnetcore/mvc/views/working-with-forms.md
@@ -184,7 +184,7 @@ Generates the following:
 
 With collection properties, `asp-for="CollectionProperty[23].Member"` generates the same name as `asp-for="CollectionProperty[i].Member"` when `i` has the value `23`.
 
-The calculated value of a `ModelExpression` is the first non-null result found in the ModelState entry with full name, the ViewData entry with full name, or the expression evaluated against Model, in that order. 
+Calculating the value of a `ModelExpression` looks at multiple sources including `ModelState` and `ViewData`. For instance, for `<input type="text" asp-for="@Name" />`, the calculated value is the first non-null result found in the ModelState entry with full name, the ViewData entry with full name, or the expression evaluated against Model, in that order. 
 
 ### Navigating child properties
 
--- a/aspnetcore/mvc/views/working-with-forms.md
+++ b/aspnetcore/mvc/views/working-with-forms.md
@@ -184,6 +184,8 @@ Generates the following:
 
 With collection properties, `asp-for="CollectionProperty[23].Member"` generates the same name as `asp-for="CollectionProperty[i].Member"` when `i` has the value `23`.
 
+The calculated value of a `ModelExpression` is the first non-null result found in the ModelState entry with full name, the ViewData entry with full name, or the expression evaluated against Model, in that order. 
+
 ### Navigating child properties
 
 You can also navigate to child properties using the property path of the view model. Consider a more complex model class that contains a child `Address` property.
--- a/aspnetcore/mvc/views/working-with-forms.md
+++ b/aspnetcore/mvc/views/working-with-forms.md
@@ -184,7 +184,7 @@ Generates the following:
 
 With collection properties, `asp-for="CollectionProperty[23].Member"` generates the same name as `asp-for="CollectionProperty[i].Member"` when `i` has the value `23`.
 
-Calculating the value of a `ModelExpression` looks at multiple sources including `ModelState` and `ViewData`. For instance, for `<input type="text" asp-for="@Name" />`, the calculated value is the first non-null result found in the ModelState entry with full name, the ViewData entry with full name, or the expression evaluated against Model, in that order. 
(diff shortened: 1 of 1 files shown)


aspnetcore/mvc/views/working-with-forms.md
```diff
@@ -184,6 +184,8 @@ Generates the following:
 
 With collection properties, `asp-for="CollectionProperty[23].Member"` generates the same name as `asp-for="CollectionProperty[i].Member"` when `i` has the value `23`.
 
+The calculated value of a `ModelExpression` is the first non-null result found in the ModelState entry with full name, the ViewData entry with full name, or the expression evaluated against Model, in that order.
```
dougbu: This is not correct in almost all cases. Instead, look at [`TextBoxFor(...)`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.rendering.ihtmlhelper-1.textareafor?view=aspnetcore-2.0) for one fairly-general case.

Also need to be a bit more generic. Different tag helpers and different `<input/>` `type`s handle the expressions differently. For example, the `value` of a generated `<input type="checkbox" .../>` is always `true` and the expression is instead evaluated to determine `checked` (see [`CheckBoxFor(...)`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.rendering.ihtmlhelper-1.checkboxfor?view=aspnetcore-2.0)). `<input type="password" .../>` ignores the expression except to name the generated `<input/>` (see [`PasswordFor(...)`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.rendering.ihtmlhelper-1.passwordfor?view=aspnetcore-2.0)). And, yes, I could go on because there are few all-encompassing rules here.

pranavkm:
🆙 📅

aspnetcore/mvc/views/working-with-forms.md
```diff
@@ -184,7 +184,7 @@ Generates the following:
 
 With collection properties, `asp-for="CollectionProperty[23].Member"` generates the same name as `asp-for="CollectionProperty[i].Member"` when `i` has the value `23`.
 
-The calculated value of a `ModelExpression` is the first non-null result found in the ModelState entry with full name, the ViewData entry with full name, or the expression evaluated against Model, in that order. 
+Calculating the value of a `ModelExpression` looks at multiple sources including `ModelState` and `ViewData`. For instance, for `<input type="text" asp-for="@Name" />`, the calculated value is the first non-null result found in the ModelState entry with full name, the ViewData entry with full name, or the expression evaluated against Model, in that order.
```
dougbu: 1. Keep the back ticks around `ModelState`, `ViewData` and `Model`
2. To clarify the sentence, use more of the specific example e.g. "`ModelState` entry with name `"Name"` (or the full name if this is not a top-level node)" &hellip; "or the result of the expression `Model.Name`."

pranavkm:
🆙 📅

Rick-Anderson:
> looks at multiple sources including `ModelState` and `ViewData`.

Should any of these other source be mentioned?  If not, why don't we say
 When ASP.NET Core calculates the value of `ModelExpression`, it inspects `ModelState` and `ViewData`.

How about the following

When ASP.NET Core calculates the value of `ModelExpression`, it inspects several sources, including `ModelState` and `ViewData`.  Consider `<input type="text" asp-for="@Name" />`. The calculated value is one of the following:

* The first non-null result found in the `ModelState` entry with key "Name".
* The `ViewData` entry with key "Name".
* The result of the expression `Model.Name`.

pranavkm:
Updated

Rick-Anderson:
@pranavkm  assume and very close. We can't say MVC. We could say ASP.NET Core MVC or ASP.NET Core runtime.

Rick-Anderson:
In lieu of several sources, can we say it looks at ModelState, ViewData, and the model expression?

pranavkm:
That bit is a little sketchy. Different tag helpers, or in some cases, the same helper depending on the kind of the `type` look at different sets of sources. I wanted to make this be a non-definitive list

Rick-Anderson:
Can I change Mvc to ASP.NET Core MVC? Or do you prefer the ASP.NET Core runtime?

pranavkm:
The former sounds better. ModelExpressions are an Mvc thing.

dougbu:
Shoot, I missed something. @Rick-Anderson please delete the bullet about `ViewData`.

Similar to the Html helpers that take Linq `Expression`s, ASP.NET Core MVC tag helpers don't use `ViewData`, ever. (`Html.DropDownListFor(...)` and `Html.ListBoxFor(...)` will check `ViewData` for the select list items. `<select>` doesn't even do that.)

---
A lesser issue is "The calculated value of `asp-for`" doesn't make much sense given there's no `asp-for` attribute in the generated HTML. Suggest "The calculated `value` attribute".

PR #5483, opened 2018-02-15
Starting work on HttpClientFactory documentation 

### LINKED ISSUE #5402: IHttpClientFactory - HttpClient as a service, handle cross-cutting concerns like caching, retry logic, timeouts and circuit breakers
aspnet/HttpClientFactory#42

stevejgordon:
Fixes #5402

This is an early WIP PR to share the initial hour of work on this piece of documentation. A majority of it is brought across from the [HttpClientFactory Wiki](https://github.com/aspnet/HttpClientFactory/wiki) by @glennc and @rynowak 

I have not proof read this / finished working on the actual wording yet. It's here purely to get early feedback that I'm heading in the right direction at this stage and structuring things correctly. 

Thoughts very much welcomed from @scottaddie and the product team.

[Internal Review Page](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?branch=pr-en-us-5483)


stevejgordon:
I went with the working title of Making HTTP requests, welcome thoughts on that in terms of discoverability and accuracy as strictly it's not covering much about actually making requests, just accessing a HttpClient to do so!

stevejgordon:
I need to fix up the casing for some of the headings too!

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,241 @@
...
+    public IActionResult Index()
+    {
+        var defaultClient = _httpClientFactory.CreateClient();
+        var gitHubClient = _httpClientFactory.CreateClient("github");
+        return View();
+    }
+}
+```
+
+In the above code the gitHubClient will have the BaseAddress and headers set whereas the defaultClient does not. This provides you the with the ability to have different configurations for different purposes. This may mean different configurations per endpoint/API for example.
```
scottaddie: In the above code --> In the preceding code,
- Prefer the word "preceding" over the word "above"

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,241 @@
...
+
+HttpClientFactory will create, and cache, a single HttpMessageHandler per named client. Meaning that if you were to use netstat or some other tool to view connections on the host machine you would generally see a single TCP connection for each named client, rather than one per instance when you new-up and dispose of a HttpClient manually.
+
+## Typed clients
+
+Typed Clients give you the same capabilities as named clients without the need for using strings as keys. This provides intellisense and compiler help when consuming clients. They also provide a single location to configure and interact with a particular HttpClient. For example, a single typed client might be used for a single backend endpoint and encapsulate all logic which deals with that endpoint.
```
scottaddie: - Typed Clients --> Typed clients
- intellisense --> IntelliSense

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,241 @@
...
+
+    public IActionResult Index()
+    {
+        var result = await _gitHubService.Client.GetStringAsync("/orgs/octokit/repos");
+        return Ok(result);
+    }
+}
+```
+
+In this example we only moved configuration into the type, but we could also have methods with behaviour and not actually expose the HttpClient if we want all access to go through this type.
```
scottaddie: Need a comma after "In this example".

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,241 @@
...
+        c.BaseAddress = new Uri("https://api.github.com/");
+
+        c.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json"); // Github API versioning
+        c.DefaultRequestHeaders.Add("User-Agent", "HttpClientFactory-Sample"); // Github requires a user-agent
+    });
+    services.AddMvc();
+}
+```
+
+If we want to entirely encapsulate the HttpClient in our typed client, rather than exposing it as a property we can define our own methods which control the client
```
scottaddie: - Need a comma after "rather than exposing it as a property".
- Need a period at the end of this sentence.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,241 @@
...
+                return resultObj;
+            }
+            result.EnsureSuccessStatusCode();
+        }        
+        return resultObj;
+    }
+}
+```
+
+In this example the HttpClient is stored as a private field and all access to make external calls goes through the GetValues method.
```
scottaddie: Need a comma after "In this example".

aspnetcore/fundamentals/index.md
```diff
@@ -208,3 +208,7 @@ For more information, see [Choosing between .NET Core and .NET Framework](/dotne
 ## Choose between ASP.NET Core and ASP.NET
 
 For more information on choosing between ASP.NET Core and ASP.NET, see [Choose between ASP.NET Core and ASP.NET](xref:fundamentals/choose-between-aspnet-and-aspnetcore).
+
+## Making HTTP requests
+
+For information about using `HttpClientFactory` to access HttpClient instances to make HTTP requests, see [Making HTTP requests](xref:fundamentals/http-requests).
```
scottaddie: Update the title in the link text to the suggested title (e.g., "Initiate HTTP requests").

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,241 @@
...
+        c.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json"); // Github API versioning
+        c.DefaultRequestHeaders.Add("User-Agent", "HttpClientFactory-Sample"); // Github requires a user-agent
+    });
+    services.AddHttpClient();
+}
+```
+
+Here `AddHttpClient` has been called twice; once with the name 'github' and once without. The github specific client has some default configuration applied, namely the base address and two headers required to work with the GitHub API.
+
+The configuration function here will get called every time CreateClient is called, as a new instance of HttpClient is created each time.
```
scottaddie: CreateClient --> `CreateClient`
- Whenever you're referencing a member name (e.g., class, method, variable, etc.), it should be code-fenced.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,241 @@
...
+        return resultObj;
+    }
+}
+```
+
+In this example the HttpClient is stored as a private field and all access to make external calls goes through the GetValues method.
+
+## Generated clients
+
+TODO - Refit?
```
glennc: Refit integration demo is here: https://github.com/rynowak/HttpClientFactoryDemos/tree/master/RefitSample

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,241 @@
...
+
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
+    services.AddHttpClient();
+    services.AddMvc();
+}
+```
+
+Once registered, you can accept a `IHttpClientFactory` in your constructor which can then be used to create a `HttpClient`.
```
glennc: Probably want to generalize this a bit, make sure people know it is DI magic and can work wherever you can inject services. Not just constructors.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,241 @@
...
+ms.topic: article
+uid: fundamentals/http-requests
+---
+# Making HTTP Requests
+
+By [Glenn Condron](https://github.com/glennc), [Ryan Nowak](https://github.com/rynowak) and [Steve Gordon](https://github.com/stevejgordon) 
+
+A `HttpClientFactory` can be registered and used to configure and consume `HttpClient` instances in your application. It provides several benefits:
+
+1. Provides a central location for naming and configuring logical HttpClients. For example, you may configure a “github” client that is pre-configured to access github and a default client for other purposes.
```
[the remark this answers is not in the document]
guardrex: Replace the smart quotes on "github" with normal straight quotes.

aspnetcore/fundamentals/http-requests.md
guardrex: I think this can be rewritten without the first person.

stevejgordon:
You're welcome @scottaddie. It's nice to be able to get involved and help. I've updated per the feedback above. I'm hoping to carve some time to do some of the later sections soon.

aspnetcore/toc.md
```diff
@@ -136,6 +136,7 @@
 ## [Microsoft.AspNetCore.All metapackage](xref:fundamentals/metapackage)
 ## [Choose between .NET Core and .NET Framework](/dotnet/articles/standard/choosing-core-framework-server)
 ## [Choose between ASP.NET Core and ASP.NET](choose-aspnet-framework.md)
+## [Making HTTP Requests](fundamentals/http-requests.md)
```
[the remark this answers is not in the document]
stevejgordon: Since changing this it seems to throw a warning on the build system?

aspnetcore/fundamentals/http-requests.md
stevejgordon: Added, thanks!

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,360 @@
...
+    public async Task<ActionResult<Reply>> Index()
+    {
+        return await _client.GetMessageAsync();
+    }
+}
+```
+
+## Outgoing request middleware
+
+The `HttpClientFactory` supports registering and chaining `DelegatingHandlers` to easily build an outgoing request middleware pipeline. Each of these handlers is able to perform work before and after the outgoing request, in a very similar pattern to the middleware pipeline in ASP.NET Core. This provides a mechanism to manage cross cutting concerns around the requests an app is making. This includes things such as caching, error handling, serialization and logging.
```
stevejgordon: @scottaddie Do you have a preference around the code backticks around pluralised types? Inclusive of the "s" or not?

aspnetcore/toc.md
```diff
@@ -136,6 +136,7 @@
 ## [Microsoft.AspNetCore.All metapackage](xref:fundamentals/metapackage)
 ## [Choose between .NET Core and .NET Framework](/dotnet/articles/standard/choosing-core-framework-server)
 ## [Choose between ASP.NET Core and ASP.NET](choose-aspnet-framework.md)
+## [Initiate HTTP requests](xref:fundamentals/http-requests.md)
```
scottaddie: Delete the ".md" extension, and this will build fine.

aspnetcore/fundamentals/http-requests.md
scottaddie: I try to avoid the plural form in backticks, as it looks awkward. I suggest rewriting to something like the following:
The `HttpClientFactory` supports registering and chaining a `DelegatingHandler` to easily build an outgoing request middleware pipeline.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,391 @@
...
+
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
+    services.AddHttpClient();
+    services.AddMvc();
+}
+```
+
+Once registered, you can accept a `IHttpClientFactory` whereever services can be injected by the DI framework which can then be used to create a `HttpClient` instance.
```
scottaddie: - whereever --> anywhere
- We try to keep sentences as concise as possible. This improves the machine translation result used for localization of the content. I recommend rewriting this as 2 sentences. Maybe something like the following:
"Once registered, you can accept a `IHttpClientFactory` anywhere services can be injected by the DI framework. The `IHttpClientFactory` can then be used to create a `HttpClient` instance."

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,391 @@
...
+    public IActionResult Index()
+    {
+        var defaultClient = _httpClientFactory.CreateClient();
+        var gitHubClient = _httpClientFactory.CreateClient("github");
+        return View();
+    }
+}
+```
+
+In the preceding code the gitHubClient will have the `BaseAddress` and `DefaultRequestHeaders` set whereas the defaultClient does not. This provides you the with the ability to have different configurations for different purposes. This may mean different configurations per endpoint/API for example.
```
scottaddie: - gitHubClient --> `gitHubClient`
- defaultClient --> `defaultClient`

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,391 @@
...
+        c.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json"); // Github API versioning
+        c.DefaultRequestHeaders.Add("User-Agent", "HttpClientFactory-Sample"); // Github requires a user-agent
+    });
+    services.AddHttpClient();
+}
+```
+
+Here `AddHttpClient` has been called twice, once with the name 'github' and once without. The GitHub-specific client has some default configuration applied, namely the base address and two headers required to work with the GitHub API.
+
+The configuration function here will get called every time ceateClient is called, as a new instance of `HttpClient` is created each time.
```
scottaddie: - ceateClient --> `CreateClient`
- will get called --> is called

scottaddie:
@stevejgordon Can you also please create an associated sample ASP.NET Core 2.x Razor Pages app that targets .NET Core 2.x? Your code snippets can then be imported from the sample app instead of embedding the code in the Markdown. See the following doc for guidance: https://github.com/aspnet/Docs/blob/master/CONTRIBUTING.md#code-snippets. 

This also allows readers to get the working code from GitHub. The sample app can be put in the following folder, which you'll have to create: fundamentals/http-requests/samples/.

Once the sample app is created, add the following to the top of the doc:
```
[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/fundamentals/http-requests/samples) ([how to download](xref:tutorials/index#how-to-download-a-sample))
```

guardrex:
When there is only one samp, I've been using the singular folder name, *sample*. I go plural when there is more than one. Let me know if I should start going 100% plural.

scottaddie:
@guardrex I prefer the plural form to make the folder name more future-proof. For example, when we also add a sample app for ASP.NET Core 3.x.

guardrex:
IIRC we weren't going to keep prior version samples tho ... insane repo build times? @Rick-Anderson pinged me to remove the 1.x sample for the Azure KV config provider topic when I work on a 2.1 update for it.

guardrex:
I'll email you on that ... it's getting too off-topic here.

stevejgordon:
@scottaddie I can look to get a sample app together for sure. At this point though the HttpClientFactory packages are not available via Nuget. I could put something together against the current MyGet nightlies? Is that okay at this stage, pending the preview 1 release which should allow the sample to be more stable?

guardrex:
@stevejgordon A couple more little things for the next pass:
* There's still future tense (e.g., "will") to convert to present tense.
* "You" and "your" aren't banned words, but their use should be avoided if possible.

stevejgordon:
Thanks @guardrex - Sure. I'm going to get eyes on this during my lunch break later today and give it a cleanup pass.

stevejgordon:
@guardrex - I've pushed a couple of cleanup passes which focus on removing "you", "your" and "will". I've also tried to clear up and simplify generally.

@scottaddie / @rynowak - I've started (but not pushed) a sample project. It's a bit flaky when running locally (HTTPS not working for me at the moment). The project is pulling in nightly builds from MyGet.

Are we okay to use GitHub as an endpoint in the examples? If it's going to run locally it needs to target a real endpoint.

I'm not convinced a Razor pages based sample makes the most sense since pulling the data from GitHub and rendering feels a bit contrived. Would an API project be more realistic and suited to this?

I'm also trying to assess how to best have a single sample demo the features. My main question is around the ConfigureServices. I can have all of the various registration methods there, but ideally most snippets want just the ones specific to that section. Just a named or just a typed registration. Any thoughts on the best way to approach that?

guardrex:
> I'm not convinced a Razor pages based sample makes the most sense since pulling the data from GitHub and rendering feels a bit contrived.

Can you put up a repo of the sample you started?

> I can have all of the various registration methods there, but ideally most snippets want just the ones specific to that section.

Normally, we'd put `#region`s around each line or lines for each snippet (i.e., `#region snippet1`-`#endregion`, `#region snippet2`-`#endregion`, etc.) ... are you saying that approach won't work?

stevejgordon:
@guardrex - I've thrown up a quick branch with just the sample. I wanted to avoid polluting this PR whilst we discuss. It's at https://github.com/stevejgordon/Docs/tree/stevejgordon/httpclientfactory-sample/aspnetcore/fundamentals/http-requests/sample

Let me know if you prefer I merge in here?

Note: It's untested and I simply added some registrations on ConfigureServices just for this discussion.

I could add a snippet around the registration in question for the current section. So...

```csharp
 services.AddHttpClient();
```

However, for a sample snippet it loses the context of where it is in the file. i.e. Would it not be better to be 

```csharp
{
	services.AddHttpClient();
}
```

guardrex:
> loses the context of where it is in the file

It's been a challenge occasionally, and the authors have taken different approaches to solve it:
* Use only the line but lead into it by telling the reader where it is. For example: "blah blah blah is in the \`Startup.ConfigureServices\` method:" (code-fence it) ... then just that one line brought in by a `#region`.
* Use a single snippet that captures more than you need to show but use line highlighting to call the reader's attention to only the lines being discussed. An example of this can be seen in the Options topic: Note how the full `IndexModel` is shown from the sample in each Options example and the line highlighting focuses the reader's attention. https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options It was done this way here because (a) I didn't want to butcher the ctor with `#region`s, and (b) I wanted a single page to show how all of the options worked so that the reader doesn't need to navigate to several pages and look at several files to see a fairly simple set of examples. 
* Use an inline code snippet [but this usually only happens when the sample doesn't have the lines that the author wants to show and adding a new sample file (for example, in a `sample_snapshot` folder) would be overkill when just wanting to show two or three lines].

I'll take a look at the repo u put up.

... and for a future pass on the topic: For the first mention of an object/method, link it to the API docs. For example, convert the first instance of \`IHttpClientFactory\` to \[IHttpClientFactory](/dotnet/api/microsoft.extensions.http.ihttpclientfactory) (no styles for link text inside the brackets). The links for 2.1 API won't work (404) for these until 2.1 actually lands, but they'll be done and out of the way.

Also, I saw a couple of methods with `()` ... the topics just get the method name and not the parentheses.

😅 lol ..... it's quite a style learning curve, isn't it!?? :smile:

guardrex:
Sample looks like a fine start.

I've been creating samples that virtually document themselves. It isn't always possible, but I try not to discuss the sample's operation much in the topic (this works for reference topics, not tutorials). I use the landing page to:
1. Explain how the sample works with instructions. 
2. Demo the features of the topic.

This results in a couple of good outcomes:
1. Outside of code snippets, the sample can change without having to update the topic.
2. The reader can take the sample and run it for a full experience without having to constantly flip back-and-forth between the sample and the topic trying to figure out what they're supposed to be looking at in the sample.

That Options sample is a good demo of this approach: https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/configuration/options/sample/Pages/Index.cshtml

Another decent one is the one that just went up for background tasks: https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/hosted-services/sample/Pages/Index.cshtml (oh ... btw ... @scottaddie I need to mod the sample location on that one ... it got caught up in our whole "sample" vs. "samples" folder naming change ... I'll PR it up in a second to get rid of the "s" on the folder).

There's no standing order that samples must be built this way, but I just thought I'd throw that out in case it sounded like a good idea and you wanted to do it with this sample.

WRT this being an API or RP sample:

Could you have a page where the user selects a button, the data is pulled down from GH in a page model method, and the data is displayed in the page?

Can you have both: Leave the API bits in there as another example. One example shows a page model method-based approach and another shows the API endpoint approach (call it from JS in the page)? ... just spitballing here. If you think multiple approaches have merit, you might want to try and show all of them.

guardrex:
Looks like a stuck build ... will close and open to trigger a new build.

stevejgordon:
Thanks for the feedback @guardrex 

I looks like at this point it's worth me getting the sample more complete and then from there it'll be easier to decide on how and where to include snippets in the documentation.

> Use a single snippet that captures more than you need to show but use line highlighting to call the reader's attention to only the lines being discussed.

This sounds and looks like it might be reasonable. I'll see how large the method in the sample gets and assess.

> lol ..... it's quite a style learning curve, isn't it!??

:smile: Sure is, but it all makes sense so happy to learn as I go as long as it's not holding things up.

> Could you have a page where the user selects a button, the data is pulled down from GH in a page model method, and the data is displayed in the page?
> 
> Can you have both: Leave the API bits in there as another example. One example shows a page model method-based approach and another shows the API endpoint approach (call it from JS in the page)? ... just spitballing here. If you think multiple approaches have merit, you might want to try and show all of them.

A combination approach might make sense, so I'll try to come up with some reasonable examples.

After my next pass at the sample, shall I bring it into this PR branch? Trying to avoid too many WIP style commits popping in but if we can squash / cleanup later then perhaps that's not an issue.

guardrex:
> After my next pass at the sample, shall I bring it into this PR branch?

Sure ... that's fine.

> we can squash / cleanup later

That's right. All PR's are squashed when they're merged, so you can add commits. You can leave the commits in the PR (you don't need to squash them if you don't want to), but I often do squash several small commits into one when I'm reacting to feedback. A reviewer may want to look and see a single commit following a round of suggestions to confirm that I've addressed their feedback. Of course, it's easier on the reviewer if they just need to look at one commit.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,392 @@
...
+
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
+    services.AddHttpClient();
+    services.AddMvc();
+}
+```
+
+Once registered, code can accept a `IHttpClientFactory` anywhere services can be injected by the DI framework. The `IHttpClientFactory` can then be used to create a `HttpClient` instance.
```
scottaddie: ~~can be injected by the DI framework.~~
can be injected with dependency injection (DI).

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,392 @@
...
+    public IActionResult Index()
+    {
+        var client = _httpClientFactory.CreateClient();
+        var result = client.GetStringAsync("http://myurl/");
+        return View();
+    }
+}
+```
+
+Using `HttpClientFactory` like this is a good way to start refactoring an existing app, as it has no impact on the way `HttpClient` is used. In places where `HttpClient` instances are currenty created, replace those with a call to `CreateClient()`.
```
scottaddie: currenty --> currently

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,392 @@
...
+        c.BaseAddress = new Uri("https://api.github.com/");
+
+        c.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json"); // Github API versioning
+        c.DefaultRequestHeaders.Add("User-Agent", "HttpClientFactory-Sample"); // Github requires a user-agent
+    });
+    services.AddHttpClient();
+}
+```
+
+Here `AddHttpClient` has been called twice, once with the name 'github' and once without. The GitHub-specific client has some default configuration applied, namely the base address and two headers required to work with the GitHub API.
```
scottaddie: ~~Here `AddHttpClient` has been called twice, once with the name 'github' and once without.~~
In the preceding code, `AddHttpClient` is called twice: once with the name "github" and once without.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,392 @@
...
+    public IActionResult Index()
+    {
+        var defaultClient = _httpClientFactory.CreateClient();
+        var gitHubClient = _httpClientFactory.CreateClient("github");
+        return View();
+    }
+}
+```
+
+In the preceding code, the `gitHubClient` has the `BaseAddress` and `DefaultRequestHeaders` set, whereas the `defaultClient` does not. This provides the ability to have different configurations for different purposes. This may mean different configurations per endpoint/API for example.
```
scottaddie: Add a comma after "endpoint/API"

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,392 @@
...
+
+`HttpClientFactory` creates, and caches, a single `HttpMessageHandler` per named client. Generally a single TCP connection for each named client is be used. When instansiating and disposing of a `HttpClient` manually, a TCP connection is used per instance. Creating multiple connections in this way can lead to socket exhaustion on the host.
+
+## Typed clients
+
+Typed clients provide the same capabilities as named clients without the need to use strings as keys. This provides IntelliSense and compiler help when consuming clients. They also provide a single location to configure and interact with a particular `HttpClient`. For example, a single typed client might be used for a single backend endpoint and encapsulate all logic which deals with that endpoint. 
+
+A typed client is expected to accept a HttpClient via it's constructor.
```
scottaddie: HttpClient --> `HttpClient`

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,392 @@
...
+
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
+    services.AddHttpClient<GitHubService>();
+    services.AddMvc();
+}
+```
+
+The typed client is registered as transient with the DI framework. The typed client can then be injected and consumed directly.
```
scottaddie: ~~with the DI framework.~~
with DI.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,392 @@
...
+        c.BaseAddress = new Uri("https://api.github.com/");
+
+        c.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json"); // GitHub API versioning
+        c.DefaultRequestHeaders.Add("User-Agent", "HttpClientFactory-Sample"); // GitHub requires a user-agent
+    });
+    services.AddMvc();
+}
+```
+
+It is possible to entirely encapsulate the `HttpClient` within a typed client. Rather than exposing it as a property, public methods can be provided which call the `HttpClient` instance internally.
```
scottaddie: - it is --> it's
- Revisit this 2nd sentence. Not sure what it's trying to say. Maybe there's a word missing?

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,392 @@
...
+                return resultObj;
+            }
+            result.EnsureSuccessStatusCode();
+        }        
+        return resultObj;
+    }
+}
+```
+
+In this example, the `HttpClient` is stored as a private field and all access to make external calls goes through the GetValues method.
```
scottaddie: GetValues --> `GetValues`

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,392 @@
...
+    {
+        c.BaseAddress = new Uri("http://localhost:5000");
+    })
+    .AddTypedClient(c => Refit.RestService.For<IHelloClient>(c));
+
+    services.AddMvc();
+}
+```
+
+The defined interface can then be consumed wherever necessary in code, with the implementation provided by the DI framework and Refit.
```
scottaddie: the DI framework --> DI

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,392 @@
...
+    public async Task<ActionResult<Reply>> Index()
+    {
+        return await _client.GetMessageAsync();
+    }
+}
+```
+
+## Outgoing request middleware
+
+`HttpClient` already has the concept of delegating handlers that can be linked together for outgoing HTTP requests. The `HttpClientFactory` makes registering of these per named client more intuitive. It supports registering and chaining multiple handlers together to build an outgoing request middleware pipeline. Each of these handlers is able to perform work before and after the outgoing request, in a very similar pattern to the inbound middleware pipeline in ASP.NET Core. This provides a mechanism to manage cross cutting concerns around HTTP requests, including caching, error handling, serialization and logging for example.
```
scottaddie: You can remove "for example" from this last sentence.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,392 @@
...
+
+    services.AddHttpClient("example", c =>
+    {
+        c.BaseAddress = new Uri("https://localhost:5000/");
+    })
+    .AddHttpMessageHandler<RetryHandler>(); // Retry requests to GitHub using the retry handler
+}
+```
+
+Here the `RetryHandler` is registered as a transient service with the DI framework. Once registered `AddHttpMessageHandler` can be called, passing in the type for the handler.
```
scottaddie: - Here --> In the preceding code,
- the DI framework --> DI
- Add a comma after "Once registered"

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,392 @@
...
+    .AddServerErrorPolicyHandler(p => p.RetryAsync())
+
+    // Build a policy that handles exceptions, 400s, and 500s from the remote server
+    .AddBadRequestPolicyHandler(p => p.RetryAsync());
+}
+```
+
+## DNS and handler rotation
+
+Each time `CreateClient` is called on the `HttpClientFactory` a new instance of a `HttpClient` is returned. The `HttpClientFactory` may reuse the underlying `HttpMessageHandler` when appropriate. The `HttpMessageHandler` is responsible for creating and maintainging the underlying operating system connection. Reusing the `HttpMessageHandler` avoids creating many connections on the host machine which can lead to socket exhaustion.
```
scottaddie: - Add a comma before "a new instance of a"
- Add a comma after "on the host machine"
- avoids creating many --> avoids creating too many

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,392 @@
...
+ms.author: riande
+ms.date: 02/15/2018
+ms.prod: asp.net-core
+ms.technology: aspnet
+ms.topic: article
+uid: fundamentals/http-requests
+---
+# Initiate HTTP requests
+
+By [Glenn Condron](https://github.com/glennc), [Ryan Nowak](https://github.com/rynowak) and [Steve Gordon](https://github.com/stevejgordon)
```
scottaddie: Need a comma before "and"

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,393 @@
...
+public class MyController : Controller
+{
+    private GitHubService _gitHubService;
+
+    public MyController(GitHubService gitHubService)
+    {
+        _gitHubService = gitHubService;
+    }
+
+    public IActionResult Index()
```
scottaddie: Need to mark this action as `async` and return a `Task`.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,393 @@
...
+
+    public MyController(IHttpClientFactory httpClientFactory)
+    {
+        _httpClientFactory = httpClientFactory;
+    }
+
+    public IActionResult Index()
+    {
+        var client = _httpClientFactory.CreateClient();
+        var result = client.GetStringAsync("http://myurl/");
```
scottaddie: Should this line use `await`? If so, also update the action signature.

stevejgordon:
@scottaddie I've added the sample to this main branch. It includes a couple of razor pages to grab and display some data. It needs some more work and examples but it's a start.

scottaddie:
@rynowak and @glennc Please review. Once you sign off on the sample app, we'll get the relevant code snippets incorporated into the doc itself.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,394 @@
...
+ms.prod: asp.net-core
+ms.technology: aspnet
+ms.topic: article
+uid: fundamentals/http-requests
+---
+# Initiate HTTP requests
+
+By [Glenn Condron](https://github.com/glennc), [Ryan Nowak](https://github.com/rynowak), and [Steve Gordon](https://github.com/stevejgordon) 
+
+A `HttpClientFactory` can be registered and used to configure and consume `HttpClient` instances in an app. It offers the following benefits:
```
rynowak: Should be `IHttpClientFactory` not `HttpClientFactory`

> to configure and consume

Can we say create instead of consume? The factory doesn't *do* HTTP it just sets up the clients for you.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,394 @@
...
+
+- Provides a central location for naming and configuring logical `HttpClient` instances. For example, a "github" client can be registered and pre-configured to access GitHub. A default client can be registered for other purposes.
+- Codifies the concept of outgoing middleware via delegating handlers in `HttpClient` and provides extensions for Polly-based middleware to take advantage of that.
+- Manages the pooling and lifetime of underlying `HttpClientMessageHandler` instances to avoid common DNS problems that occur when manually managing `HttpClient` lifetimes.
+
+## Consumption patterns
+
+There are several ways `HttpClientFactory` can be used in an app. None of them are strictly superior to another&mdash;it depends on the app's constraints.
```
rynowak: CTRL+H `HttpClientFactory` -> `IHttpClientFactory`

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,394 @@
...
+    public MyController(IHttpClientFactory httpClientFactory)
+    {
+        _httpClientFactory = httpClientFactory;
+    }
+
+    public IActionResult Index()
+    {
+        var defaultClient = _httpClientFactory.CreateClient();
+        var gitHubClient = _httpClientFactory.CreateClient("github");
+        return View();
```
rynowak: It looks like the wrong line got replaced here, 2 clients are created and neither is used.

rynowak: If you intended to create two clients here, I have a bit of a gripe with it because we'd never tell a user to do this 😆

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,394 @@
...
+    }
+}
+```
+
+In the preceding code, the `gitHubClient` variable has the `BaseAddress` and `DefaultRequestHeaders` set. The `defaultClient` variable doesn't. This provides the ability to have different configurations for different purposes. This could mean different configurations per endpoint/API, for example.
+
+`HttpClientFactory` creates, and caches, a single `HttpMessageHandler` per named client. Generally, a single TCP connection for each named client is used. When instantiating and disposing of a `HttpClient` manually, a TCP connection is used per instance. Creating multiple connections in this way can lead to socket exhaustion on the host.
```
rynowak: > a single TCP connection for each named client is used.

This isn't really correct. Each client handler has its own **pool** of connections, and the behavior is implementation/platform specific.

I think a correct statement to make would be to say that "generally the lifetime of connections used by a handler will not exceed the lifetime of the handler" (based on current implementations) 

Is this really the right time and place to introduce this kind of discussion?

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,394 @@
...
+
+## Typed clients
+
+Typed clients provide the same capabilities as named clients without the need to use strings as keys. This provides IntelliSense and compiler help when consuming clients. They provide a single location to configure and interact with a particular `HttpClient`. For example, a single typed client might be used for a single backend endpoint and encapsulate all logic dealing with that endpoint.
```
rynowak: Another major reason to use typed clients is that they are DI-friendly. You can either put code in the typed client or treat it as a 'handle'. Both are fine.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,394 @@
...
+{
+    public HttpClient Client { get; private set; }
+
+    public GitHubService(HttpClient client)
+    {
+        client.BaseAddress = new Uri("https://api.github.com/");
+        client.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json"); // GitHub API versioning
+        client.DefaultRequestHeaders.Add("User-Agent", "HttpClientFactory-Sample");   // GitHub requires a user-agent
+
+        Client = client;
```
rynowak: /cc @glennc  for visibility. 

I'm not a super fan of this. If you end up needing to use configuration to get these values then you have to find a way to pass that in here. 

Perhaps Glenn and I need to discuss this and come to an understanding.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,394 @@
...
+        _gitHubService = gitHubService;
+    }
+
+    public async Task<IActionResult> Index()
+    {
+        var result = await _gitHubService.Client.GetStringAsync("/orgs/octokit/repos");
+        return Ok(result);
+    }
+}
+```
```
rynowak: This is great 👍

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,394 @@
...
+A typed client can be added, using Refit to generate the implementation:
+
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
+    services.AddHttpClient("hello", c =>
+    {
+        c.BaseAddress = new Uri("http://localhost:5000");
+    })
+    .AddTypedClient(c => Refit.RestService.For<IHelloClient>(c));
```
rynowak: We also have another variant here that will takes `<TInterface, TImpl>`.

One other comment I have about this section is that this style of configuration is in no way specific to refit. It can work with any type that accepts an `HttpClient`, or you can provide your own factory methods.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,394 @@
...
+            }
+        }
+
+        // Unreachable.
+        throw null;
+    }
+}
+```
+
+The preceding code defines a basic retry handler which retries up to five times if a `HttpRequestException` is caught.
```
rynowak: I'm not sure if we have an example of a handler that isn't a dupe of what's in Polly, but our actual guidance will be to use Polly for this kind of thing.

Maybe generating and adding a tracking requestId?

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,394 @@
...
+
+    services.AddHttpClient("example", c =>
+    {
+        c.BaseAddress = new Uri("https://localhost:5000/");
+    })
+    .AddHttpMessageHandler<RetryHandler>(); // Retry requests to GitHub using the retry handler
+}
+```
+
+In the preceding code, the `RetryHandler` is registered as a transient service with DI. Once registered, `AddHttpMessageHandler` can be called, passing in the type for the handler.
```
rynowak: This should say that it **must** be registered as a transient.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,394 @@
...
+    services.AddHttpClient("example", c =>
+    {
+        c.BaseAddress = new Uri("https://localhost:5000/");
+    })
+    .AddHttpMessageHandler<OuterHandler>() // This handler is on the outside and executes first on the way out and last on the way in.
+    .AddHttpMessageHandler<InnerHandler>(); // This handler is on the inside, closest to the request.
+}
+```
+
+## Handling errors with Polly
```
rynowak: I'm going to pass over this section for now since Polly is WIP

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,394 @@
...
+
+    // Build a policy that handles exceptions and 500s from the remote server
+    .AddServerErrorPolicyHandler(p => p.RetryAsync())
+
+    // Build a policy that handles exceptions, 400s, and 500s from the remote server
+    .AddBadRequestPolicyHandler(p => p.RetryAsync());
+}
+```
+
+## DNS and handler rotation
```
rynowak: I think it would be great to call this section 'HttpClient' and lifetime management and introduce some of these concepts a bit more gently.

aspnetcore/fundamentals/http-requests/samples/Startup.cs
```diff
@@ -0,0 +1,70 @@
...
+            services.AddHttpClient("github", c =>
+            {
+                c.BaseAddress = new Uri("https://api.github.com/");
+
+                c.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json"); // Github API versioning
+                c.DefaultRequestHeaders.Add("User-Agent", "HttpClientFactory-Sample"); // Github requires a user-agent
+            });
+
+            // configuration occurs in ctor
+            services.AddHttpClient<GitHubService>();
```
glennc: @stevejgordon do you have a preference between these two styles of registration? Ryan and I have talked about it in the past and I am looking for a third opinion on it :).

aspnetcore/fundamentals/http-requests.md
stevejgordon: I originally took this sample the text below it from @glennc's work on the Wiki. It appeared the intent was to show both options being used within one app.

Happy to factor this out of the final sample app though and focus on each use case in isolation. In the sample the current approach is to register multiple clients to show the various consumption patterns. Do we need to consider that when designing the sample?

aspnetcore/fundamentals/http-requests.md
stevejgordon: Agree, will move, reword this to the later section on lifetimes.

aspnetcore/fundamentals/http-requests.md
stevejgordon: Yeah I was trying to come up with something realistic that wouldn't be solved by Polly and hit a wall. I can look at something which generates a tracking ID and puts it into the header for the sample.

aspnetcore/fundamentals/http-requests/samples/Startup.cs
stevejgordon: @glennc I lean towards the ctor configuration personally. I find in larger real-world apps, the ConfigureServices method can get quite complex, even if things end up split out into extension methods. Putting the configuration into the typed client feels closer to where I'd personally expect to look for it.

stevejgordon:
I've made some tweaks based on @rynowak's feedback. A few questions raised in response.

I was hoping to spend more time on the sample but a few things have come up so far this week. Is what is there on track so far? If so I'll continue along those lines for the other samples.

If the current samples look to be on par I can figure out the snippet inclusion to replace the hard coded samples.

aspnetcore/fundamentals/index.md
```diff
@@ -101,6 +101,10 @@ ASP.NET Core includes a rich set of built-in middleware:
 
 For more information, see [Middleware](xref:fundamentals/middleware/index) and [Open Web Interface for .NET (OWIN)](xref:fundamentals/owin).
 
+## Initiate HTTP requests
+
+For information about using `HttpClientFactory` to access HttpClient instances to make HTTP requests, see [Initiate HTTP requests](xref:fundamentals/http-requests).
```
scottaddie: HttpClientFactory --> IHttpClientFactory

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,390 @@
+---
+title: Initiate HTTP requests
+author: stevejgordon
+description: Learn about using the HttpClientFactory to managed logic HttpClient instances.
```
scottaddie: HttpClientFactory --> IHttpClientFactory

stevejgordon:
@scottaddie / @guardrex - I've started to update the MD doc with the code from the sample. Is there a way to test this locally to ensure the highlighting is correct for example?

aspnetcore/fundamentals/http-requests/samples/GitHub/GitHubService.cs
```diff
@@ -0,0 +1,50 @@
...
+            client.DefaultRequestHeaders.Add("Accept", "application/vnd.github.v3+json"); // GitHub API versioning
+            client.DefaultRequestHeaders.Add("User-Agent", "HttpClientFactory-Sample"); // GitHub requires a user-agent
+
+            Client = client;
+        }
+
+        public async Task<GitHubIssue> GetLatestDocsIssue()
+        {
+            using (var stream = await Client.GetStreamAsync("/repos/aspnet/docs/issues?state=open&sort=created&direction=desc"))
+            using (var streamReader = new StreamReader(stream))
```
stevejgordon: I'd welcome thoughts on whether this is a) correct and b) too complicated for a sample?

I've done it this way as per the advice from @DamianEdwards and @davidfowl during NDC, they highlighted that for performance larger response content should be read as stream, rather than string. @DamianEdwards touched on this again during this week's community stand-up.

This seems to work and I hope is a reasonably valid approach but as per those sources, it's not necessarily easy to do right.

Personally I'd like the sample to try to be real world to some expect to avoid people copy/pasting demoware code, but I understand this may over-complicate the main topic being shown here.

cc// @scottaddie

scottaddie:
@stevejgordon To test this locally, you'll need to use DocFX. See the instructions here: https://github.com/aspnet/Docs/blob/master/CONTRIBUTING.md#test-your-changes-with-docfx.

As an alternative to downloading and unzipping, as described in the instructions, you can install DocFX via Chocolatey:
```console
choco install docfx
```

stevejgordon:
Thanks @scottaddie. Will try that tool.
I'm hoping I can make time to add more sample use cases tomorrow. My last chance before setting off for MVP Summit.

aspnetcore/fundamentals/http-requests.md
rynowak: I think registering multiple clients is fine, my specific gripe is about this method ...

aspnetcore/fundamentals/http-requests.md
rynowak: yeah, the two ideas we have here are to do service discovery or to do something like a tracking header. /cc @glennc  for pointer

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,367 @@
...
+        c.BaseAddress = new Uri("https://localhost:5000/");
+    })
+    .AddHttpMessageHandler<OuterHandler>() // This handler is on the outside and executes first on the way out and last on the way in.
+    .AddHttpMessageHandler<InnerHandler>(); // This handler is on the inside, closest to the request.
+}
+```
+
+## Handle errors with Polly
+
+_NOTE: This is a WIP since the Polly Extensions API is still currently under design._
```
scottaddie: Is this a note to yourself, or would you like the reader to see this? If it's intended for the reader, please use the following syntax:
```
> [!NOTE]
> This section is a work in progress, since the Polly Extensions API is in the design phase.
```

scottaddie:
@stevejgordon Anything you need from me to keep this PR moving along?

stevejgordon:
Sorry for the delay @scottaddie. I got struck down by a bug after Summit and even now I'm not 100% but starting to feel more alive.

I worked on this a little during the Thursday hack day at Summit. I even managed to run part of the sample code by @glennc for a seal of approval. We also talked about timing for this since some of the final elements (Polly integration) are due in preview 2. I think most (all) of those are now completed though so I could switch to nightlies for the sample and build those up to.

I'll aim to get some eyes on that this week once I'm a little more fully recovered.

scottaddie:
@stevejgordon Thank you for the update! Get well soon.

scottaddie:
@stevejgordon I'm just checking in with you. Anything I can do to help push this PR to completion?

stevejgordon:
Hey @scottaddie. Thanks for your patience and sorry for the delay on my part. I'm hoping to get some time on this later tonight after work. Hopefully I can move it forward a fair bit closer.

stevejgordon:
@scottaddie Not a much time to progress as I'd have liked but have moved some of the Polly related docs forward and moved some snippets over to pull from the sample project. Will aim to continue on the sample project which needs more work and continue with the docs tomorrow/Friday.

stevejgordon:
@scottaddie, @rynowak and @glennc  - I _**think**_ this latest commit gets us reasonably close to a version we could ship?

The sample is updated and includes a lot of the feedback from prior discussions. The doc content now uses snippets from the sample in nearly all places.

Is there anything you feel I've missed that needs covering? I'm going to continue diving through the code and will add anything that comes to mind.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,241 @@
...
+# Initiate HTTP requests
+
+By [Glenn Condron](https://github.com/glennc), [Ryan Nowak](https://github.com/rynowak), and [Steve Gordon](https://github.com/stevejgordon)
+
+An `IHttpClientFactory` can be registered and used to configure and create [HttpClient](/dotnet/api/system.net.http.httpclient) instances in an app. It offers the following benefits:
+
+* Provides a central location for naming and configuring logical `HttpClient` instances. For example, a "github" client can be registered and configured to access GitHub. A default client can be registered for other purposes.
+* Codifies the concept of outgoing middleware via delegating handlers in `HttpClient` and provides extensions for Polly-based middleware to take advantage of that.
+* Manages the pooling and lifetime of underlying `HttpClientMessageHandler` instances to avoid common DNS problems that occur when manually managing `HttpClient` lifetimes.
+
```
rynowak: I'd suggest adding another bullet point here:

* Adds a configurable logging experience (via `ILogger`) for all requests sent through clients created by the factory.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,241 @@
...
+[!code-csharp[Main](http-requests/samples/Handlers/ValidateHeaderHandler.cs?name=snippet1)]
+
+The preceding code defines a basic handler. It checks to see if an X-API-KEY header has been included on the request. If the header is missing it can avoid the HTTP call and return a suitable response.
+
+During registration, one or more handlers can be added to the configuration for a `HttpClient`. This task is accomplished via extension methods on the `IHttpClientBuilder`.
+
+[!code-csharp[](http-requests/samples/Startup.cs?name=snippet5)]
+
+In the preceding code, the `ValidateHeaderHandler` is registered as a transient service with DI. Once registered, `AddHttpMessageHandler` can be called, passing in the type for the handler.
```
rynowak: I think it's worth a callout here that a delegating handler **MUST** be registered as a transient.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,241 @@
...
+
+Multiple handlers can be registered in the order that they should execute. Each handler wraps the next handler until the final `HttpClientHandler` executes the request:
+
+[!code-csharp[](http-requests/samples/Startup.cs?name=snippet6)]
+
+## Using Polly-based handlers
+
+`IHttpClientFactory` integrates with a popular third-party library called [Polly](https://github.com/App-vNext/Polly). Polly is a comprehensive resilience and transient fault-handling library for .NET. It allows developers to express policies such as Retry, Circuit Breaker, Timeout, Bulkhead Isolation, and Fallback in a fluent and thread-safe manner.
+
+Extension methods are provided to enable the use of Polly policies with configured `HttpClient` instances. The Polly extensions are available in a Nuget package called 'Microsoft.Extensions.Http.Polly'. To use the extensions, a PackageReference should be included in the project.
```
rynowak: It's worth calling out that `Microsoft.Extensions.Http.Polly` is **not** included by `Microsoft.AspNetCore.App` metapackage, so we expect most apps to add an explicit package reference.

The default experience users will see is that they have access to `IHttpClientFactory` but not the Polly stuff.

aspnetcore/fundamentals/http-requests/samples/GitHub/GitHubService.cs
```diff
@@ -0,0 +1,52 @@
...
+                {
+                    if (jsonReader.TokenType == JsonToken.StartObject)
+                    {
+                        var issue = serializer.Deserialize<GitHubIssue>(jsonReader);
+
+                        if (issue != null)
+                        { 
+                            return issue; // we only want the first object
+                        }
+                    }
```
rynowak: This code should be using `ReadAsAsync<>` https://github.com/rynowak/Samples/blob/dev/MusicStore/src/MusicStoreUI/Services/MusicStoreService.cs#L37

aspnetcore/fundamentals/http-requests/samples/GitHub/RepoService.cs
```diff
@@ -0,0 +1,35 @@
...
+            var result = await _httpClient.GetAsync("aspnet/repos");
+
+            IEnumerable<string> resultObj = null;
+
+            if (result.IsSuccessStatusCode)
+            {
+                resultObj = JsonConvert.DeserializeObject<IEnumerable<string>>(await result.Content.ReadAsStringAsync()).ToList();
+            }
+
+            return resultObj ?? Array.Empty<string>();
```
rynowak: We should show throwing an exception rather than returning null or empty. Just use https://github.com/rynowak/Samples/blob/dev/MusicStore/src/MusicStoreUI/Services/MusicStoreService.cs#L35

aspnetcore/fundamentals/http-requests/samples/Startup.cs
```diff
@@ -0,0 +1,149 @@
...
+            #endregion
+
+            #region snippet8
+            var timeout = Policy.TimeoutAsync<HttpResponseMessage>(TimeSpan.FromSeconds(10));
+            var longTimeout = Policy.TimeoutAsync<HttpResponseMessage>(TimeSpan.FromSeconds(30));
+
+            services.AddHttpClient("conditionalpolicy")
+            // Run some code to select a policy based on the request
+                .AddPolicyHandler(request => request.Method == HttpMethod.Get ? timeout : longTimeout);
+            #endregion
```
rynowak: It's worth calling out somewhere in the Polly part of the docs that users should treat polly policies as long-lived objects. Some policies only function correctly when they are shared or reused. We've worked to make this the idiomatic thing, but it should be covered in the docs as well.

aspnetcore/fundamentals/http-requests/samples/Startup.cs
```diff
@@ -0,0 +1,149 @@
...
+            {
+                app.UseExceptionHandler("/Error");
+                app.UseHsts();
+            }
+
+            app.UseStaticFiles();
+
+            app.UseMvc();
+        }
+    }
```
rynowak: /cc @reisenberger @joelhulen

aspnetcore/fundamentals/http-requests/samples/Startup.cs
joelhulen: Yes, in general we do recommend defining the policy early on and reusing them. Policies are thread-safe and each call to `.Execute` or `.ExecuteAsync` establishes a new, independent underlying policy state. Certain policies, as you say, must be shared for maximum effectiveness, like the `CircuitBreaker`. I can update the Polly and HttpClientFactory doc to reiterate this fact and point to more detailed documentation on the subject.

joelhulen: Actually, I believe this section of the document explains it clearly enough, do you agree? [Use case: scoping CircuitBreakers and Bulkheads](https://github.com/App-vNext/Polly/wiki/Polly-and-HttpClientFactory#use-case-scoping-circuitbreakers-and-bulkheads)

aspnetcore/fundamentals/http-requests/samples/GitHub/GitHubService.cs
stevejgordon: Handy! I wasn't aware ReadAsAsync was available. I believe (while more complex) the code I have is slightly more efficient? In this example it's taking only the first item. A bit of a contrived example. With ReadAsAsync I get all items back in the list. A very rough comparison is ~150kb extra allocated between memory snapshots.

I spoke to @glennc and @DamianEdwards at Summit as I was concerned about this complexity and if in fact this was a reasonable approach. At that time it seemed like we wanted to include something similar.

I'm happy to make the change though?

Side query: Am I correct in thinking that https://github.com/aspnet/AspNetWebStack/blob/master/src/System.Net.Http.Formatting/HttpContentExtensions.cs is the source building Microsoft.AspNet.WebApi.Client?

reisenberger:
Big thanks to @stevejgordon for the work on the HttpClientFactory-Polly side of this doco! I'll aim to review from Polly perspective and feed back, in the next 24-72 hours.

aspnetcore/fundamentals/http-requests/samples/GitHub/GitHubService.cs
rynowak: If what you're saying is true then we should choose a different example. The focus of this doc isn't on the minutae of how to implement special case handling of JSON scenarios, so it should use the idiomatic approach - which is `ReadAsAsync`.

stevejgordon: I can adjust the sample to return the full list and switch to ReadAsAsync. Not a problem from my side so I'll get that done hopefully later today / tomorrow.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,242 @@
...
+
+Extension methods are provided to enable the use of Polly policies with configured `HttpClient` instances. The Polly extensions are available in a NuGet package called 'Microsoft.Extensions.Http.Polly'. This package is not included by default by the 'Microsoft.AspNetCore.App' metapackage. To use the extensions, a PackageReference should be explicitly included in the project.
+
+[!code-csharp[](http-requests/samples/HttpClientFactorySample.csproj?highlight=9)]
+
+After restoring this package, extension methods are available to support adding Polly-based handlers to clients.
+
+### Handle transient faults
+
+The most common faults you may expect to occur when making external HTTP calls are transient. A convenient extension method called `AddTransientHttpErrorPolicy` is included, which allows a policy to be defined. The policy applies when these common transient errors occur. Examples of errors handled by the policy include `HttpRequestException`, HTTP 5xx responses, and HTTP 408 responses.
```
reisenberger: "Examples of errors ... include" may suggest there are other kinds of error not mentioned which are also included, but what is stated _is_ the full set.  Maybe avoid "includes"?

Could:
>Policies configured with this extension method handle `HttpRequestException`, HTTP 5xx responses, and HTTP 408 responses.

reisenberger: Could:

>A convenient extension method [...] is included, which allows a policy to be defined to handle transient errors.

(and lose following sentence)

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,242 @@
...
+
+[!code-csharp[Main](http-requests/samples/Startup.cs?name=snippet10)]
+
+In the preceding code, a PolicyRegistry is added to the `ServiceCollection` and two policies are registered with it. In order to use a policy from the registry, the `AddPolicyHandlerFromRegistry` method is used, passing the name of the policy to apply.
+
+Further information about `IHttpClientFactory` and Polly integrations can be found on the [Polly wiki](https://github.com/App-vNext/Polly/wiki/Polly-and-HttpClientFactory).
+
+## HttpClient and lifetime management
+
+Each time `CreateClient` is called on the `IHttpClientFactory`, a new instance of a `HttpClient` is returned. There's a `HttpMessageHandler` per named client. `IHttpClientFactory` pools the `HttpMessageHandler` instances created by the factory to reduce resource consumption. A `HttpMessageHandler` instance may be reused from the pool when creating new `HttpClient` instance if its lifetime hasn't expired.
```
reisenberger: >when creating new `HttpClient` instance

when creating a new `HttpClient` instance

aspnetcore/fundamentals/http-requests.md
```diff
@@ -0,0 +1,242 @@
...
+
+The log category used for each client includes the name of the client. A client named "MyNamedClient", for example, logs messages with a category of `System.Net.Http.HttpClient.MyNamedClient.LogicalHandler`. Messages with the suffix of "LogicalHandler" occur on the outside of request handler pipeline. On the request, messages are logged before any other handlers in the pipeline have processed it. On the response, messages are logged after any other pipeline handlers have received the response.
+
+Logging also occurs on the inside of the request handler pipeline. In the "MyNameClient" example, those messages are logged against the log category `System.Net.Http.HttpClient.MyNamedClient.ClientHandler`. For the request, logging occurs after all other handlers have run and immediately before the request is sent out on the network. On the response, this logging includes the state of the response before it passes back through the handler pipeline.
```
reisenberger: >In the "MyNameClient" example

"MyNamedClient"

stevejgordon:
Thanks @reisenberger. Good catches. I've fixed up mostly as suggested!

scottaddie:
@stevejgordon Thank you so much for your patience and hard work on this PR. It's always a great feeling to get such high-quality contributions from community members.

stevejgordon:
You're very welcome. It was a pleasure working with you and the team on this. Thanks for your patience in between commits! Hopefully I can find something else to help with in the future.

PR #5488, opened 2018-02-16
Surface XML array indexing for Configuration

guardrex:
React to https://github.com/aspnet/Configuration/issues/770

@HaoK Would u like to address that in the topic? It's true that if one tries to do this ...

```csharp
Configuration["wizard:1:age"]
```

... it won't work with XML.


guardrex:
ping @HaoK

guardrex:
Thanks @HaoK 

@scottaddie *"Meesa lik'in dis. Wha da ya tinks?"* - Jar Jar Binks

Issue #5490, opened 2018-02-16
Servers > ANCM UE pass


guardrex:
https://docs.microsoft.com/aspnet/core/fundamentals/servers/aspnet-core-module

guardrex:
@Rick-Anderson @scottaddie Let's consider reducing the use of the acronym "ANCM" for "ASP.NET Core Module" in this topic. Note that I plan to *reduce* its use ... not eliminate its use.

* It's only rarely used in topics outside of this topic:
   * static-files.md: 2
   * windowsauth.md: 1
   * nano-server.md: 0 (used as part of a script name)
* After the initial reference to the "ASP.NET Core Module" in a section, it's satisfactory to just use "the module" thereafter.
* It's not the name of the module file in IIS (*aspnetcore.dll*) or the name shown in the *Modules* list ("AspNetCoreModule").
* It's not used on the .NET Downloads page or when running the hosting bundle installer.
* Outside of our topics and GH issues, "ancm" doesn't provide relevant results on the first few pages of Google search results.
* Should we be adding yet another acronym to the pantheon of acronyms in .NET 😓 only really useful for IIS insiders? The more acronyms there are, the more cryptic technology becomes.

Rick-Anderson:
AFAIK you need approval - see [this](https://gist.github.com/klaaspieter/12cd68f54bb71a3940eae5cdd4ea1764).

TAWYG!
(Totally Agree With You Guys)

guardrex:
It seems to me that this topic isn't a good spot to halfheartedly duplicate what's in the *Hosting on IIS* topic.

> This section provides an overview of the process for setting up an IIS server and ASP.NET Core app. For detailed instructions, see [Host on Windows with IIS](#).

Why? ... Why not just tell the reader to go access the *Hosting on IIS* topic and the module config reference, where the content is complete (including security coverage)?

I see a few gems in here that should be moved over to the *Hosting on IIS* and module reference topics, but I recommend some of these sections be removed.

This topic should probably just describe the basics of the module (what it is and what it does) and link out to our best coverage for using the module. Other than showing a few lines of code, the sample isn't valuable.

PR #5491, opened 2018-02-16
Add hosting bundle installer desc to ANCM topic

guardrex:
We need a good link point in the topic for the hosting bundle installer that clearly describes what's in the bundle.

I've also opened [Servers > ANCM UE pass (\#5490)](https://github.com/aspnet/Docs/issues/5490) ... this topic needs a pass.

### DIFF
--- a/aspnetcore/fundamentals/servers/aspnet-core-module.md
+++ b/aspnetcore/fundamentals/servers/aspnet-core-module.md
@@ -51,7 +51,7 @@ This section provides an overview of the process for setting up an IIS server an
 
 ### Install ANCM
 
-The ANCM is installed in IIS on Windows Server and in IIS Express on Windows desktop operating systems. For servers and development machines, the ANCM is included in the [.NET Core Windows Server Hosting bundle](https://aka.ms/dotnetcore-2-windowshosting). If installing Visual Studio, the ANCM is automatically installed in IIS Express and in IIS if IIS is already installed on the machine.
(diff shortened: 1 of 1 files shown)


aspnetcore/fundamentals/servers/aspnet-core-module.md
```diff
@@ -51,8 +51,12 @@ This section provides an overview of the process for setting up an IIS server an
 
 ### Install ANCM
 
+The ANCM is installed in IIS on Windows Server and in IIS Express on Windows desktop operating systems. For servers, the ANCM is included in the [.NET Core Windows Server Hosting bundle](https://aka.ms/dotnetcore-2-windowshosting). For development machines, Visual Studio automatically installs the ANCM in IIS Express and in IIS if it's already installed on the machine.
```
scottaddie: What if I use VS Code on my dev machine?

guardrex: Yeah ... I don't like that either. I use VSC, and I run the hosting bundle installer. It's kind'a covered by the topic sentence ...

> The ANCM is installed in IIS on Windows Server and in IIS Express on Windows desktop operating systems. 

... but the last sentence confounds the point. I'll fix this now, and I'll hit the rest of the little probs in the topic on #5490.

guardrex:
@scottaddie The second sentence needed a bit more work to clear up how the "if" condition applied to the rest of the sentence. By using parentheses, there's no question about it.

PR #5498, opened 2018-02-17
Add model to TH sample

Rick-Anderson:
Without model it looks like the TH isn't doing anything other than changing `asp-for` into `for`.

This is not meant to be a clean up, just need to fix this problem.

[review url](https://review.docs.microsoft.com/en-us/aspnet/core/mvc/views/tag-helpers/intro?branch=pr-en-us-5498)


scottaddie:
@Rick-Anderson Please review my PR for proposed changes.

PR #5507, opened 2018-02-19
DI scope validation enabled by default

### LINKED ISSUE #3660: DI: scope validation (now on by default in hosting)

guardrex:
Fixes #3660 

[Internal Review Topic (DI)](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?branch=pr-en-us-5507#scope-validation)
[Internal Review Topic (Hosting)](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/hosting?branch=pr-en-us-5507#scope-validation)

@pakrym The docs don't describe the "root provider" anywhere. I don't think we can put this up without briefly describing it and why scoped services shouldn't be resolved from it. If you'll provide a few notes, I'll wordsmith them in here.

### DIFF
--- a/aspnetcore/fundamentals/dependency-injection.md
+++ b/aspnetcore/fundamentals/dependency-injection.md
@@ -188,6 +188,10 @@ Observe which of the `OperationId` values vary within a request, and between req
 
 * *Singleton* objects are the same for every object and every request (regardless of whether an instance is provided in `ConfigureServices`)
 
+## Scope validation
+
+In ASP.NET Core 2.0 or later, the default service provider performs a check to verify that scoped services aren't resolved from the root provider when the app is running in the Development environment. For more information, see [Scope validation in the Hosting topic](xref:fundamentals/hosting#scope-validation).
+
 ## Request Services
 
 The services available within an ASP.NET request from `HttpContext` are exposed through the `RequestServices` collection.
--- a/aspnetcore/fundamentals/dependency-injection.md
+++ b/aspnetcore/fundamentals/dependency-injection.md
@@ -192,8 +192,12 @@ Observe which of the `OperationId` values vary within a request, and between req
 
 When the app is running in the Development environment on ASP.NET Core 2.0 or later, the default service provider performs checks to verify that:
 
-* Scoped services aren't directly or indirectly resolved from the root provider.
+* Scoped services aren't directly or indirectly resolved from the root service provider.
 * Scoped services aren't directly or indirectly injected into singletons.
+
+The root service provider is created when [BuildServiceProvider](/dotnet/api/microsoft.extensions.dependencyinjection.servicecollectioncontainerbuilderextensions.buildserviceprovider) is called. The root service provider's lifetime usually corresponds to the app/server's lifetime.
+
+Scoped services are disposed by the container that created them. If a scoped service is created in the root container, the service's lifetime is effectively promoted to singleton because it's only disposed by the root container when app/server is shut down. Validating service scopes catches these situations when `BuildServiceProvider` is called.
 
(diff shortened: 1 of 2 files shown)


pakrym:
Root service provider is the one that is created by calling `serviceCollection.BuildServiceProvider` it's lifetime usually coresponds to application/server lifetime.

Scoped services are attached and more importantly disposed with container they were created in. If scoped service gets created by root contaner it's lifetime if effectively promoted to singleton because it will only get disposed with the root container when server is shutdown and this is a thing we are trying to avoid.

guardrex:
ty @pakrym ... how does it look?

guardrex:
Thanks @pakrym

@scottaddie It's ready for a look.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -188,6 +188,19 @@ Observe which of the `OperationId` values vary within a request, and between req
...
 * *Singleton* objects are the same for every object and every request (regardless of whether an instance is provided in `ConfigureServices`)
 
+## Scope validation
+
+When the app is running in the Development environment on ASP.NET Core 2.0 or later, the default service provider performs checks to verify that:
+
+* Scoped services aren't directly or indirectly resolved from the root service provider.
+* Scoped services aren't directly or indirectly injected into singletons.
+
+The root service provider is created when [BuildServiceProvider](/dotnet/api/microsoft.extensions.dependencyinjection.servicecollectioncontainerbuilderextensions.buildserviceprovider) is called. The root service provider's lifetime usually corresponds to the app/server's lifetime.
```
scottaddie: Using the word "usually" here reveals uncertainty. In which situations would the lifetime not correspond to the app/server's lifetime?

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -188,6 +188,19 @@ Observe which of the `OperationId` values vary within a request, and between req
...
+
+Scoped services are disposed by the container that created them. If a scoped service is created in the root container, the service's lifetime is effectively promoted to singleton because it's only disposed by the root container when app/server is shut down. Validating service scopes catches these situations when `BuildServiceProvider` is called.
```
scottaddie: Consider removing the word "effectively". It doesn't seem to add any value.

aspnetcore/fundamentals/dependency-injection.md
guardrex: I presume from @pakrym's remarks that if the dev builds one or more additional "root" containers on-the-fly that it won't exist for the full lifetime of the app/server.

aspnetcore/fundamentals/dependency-injection.md
guardrex: I think it adds some value because it wouldn't be correct to say that the "lifetime *IS* promoted to singleton." It it "*effectively* promoted to singleton" by virtue of the lifetime of the root container.

aspnetcore/fundamentals/dependency-injection.md
guardrex: We could go with something more explicit, like ...

> The root service provider's lifetime corresponds to the app/server's lifetime when the provider starts when the app starts and is disposed when the app shuts down.

aspnetcore/fundamentals/dependency-injection.md
guardrex: I've read this a few times. We can't honestly say that it **IS** promoted to singleton when it's only happening by virtue of the timing of its disposal. If there's another wording to consider, I'm all :ear: :ear: :ear:. I like the sentence tho ... it's clear what the problem is the way that @pakrym stated it.

PR #5510, opened 2018-02-19
Error in cookie parameter default values

ivarne:
Seems like the table got an error.


ivarne:
I have not checked the source to see if this change is actually correct

guardrex:
Yes, those are correct:

https://github.com/aspnet/Security/blob/release/2.0.0/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationDefaults.cs#L39
https://github.com/aspnet/Security/blob/release/2.0.0/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs#L33

Thanks for fixing the topic, @ivarne! :rocket: We'll get this merged as soon as the build completes.

PR #5517, opened 2018-02-20
Document ReturnUrlParameter default value

ivarne:
I was so happy to see a so quick positive response on #5510, so I decided to try another one.

https://github.com/aspnet/Security/blob/99aa3bd35dd5fbe46a93eef8a2c8ab1f9fe8d05b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationDefaults.cs#L44


guardrex:
Thanks @ivarne!

Issue #5391, opened 2018-02-08
Tutorial: Getting Started with ASP.NET Core SignalR


Rick-Anderson:

rachelappel:
- Prereqs
- Get Started with SignalR for ASP.NET Core
- Create an ASP.NET Core project that uses SignalR
- Create the SignalR Hub
- Configure the project to use SignalR
- Create the SignalR client code
- Verify code works
- Next Steps

Rick-Anderson:
@anurse  do you want to recycle [Tutorial: Getting Started with SignalR 2 and MVC 5](https://docs.microsoft.com/en-us/aspnet/signalr/overview/getting-started/tutorial-getting-started-with-signalr-and-mvc) ? 

We would use RP (Razor Pages) in lieu of MVC. What other changes? (besides changes to the product)

scottaddie:
Fixed by https://github.com/aspnet/Docs/pull/5540

PR #5540, opened 2018-02-23
Get Started SignalR Core

rachelappel:
[review URL](https://review.docs.microsoft.com/en-us/aspnet/core/signalr/get-started-signalr-core?branch=pr-en-us-5540)
#5391


rachelappel:
@anurse Would you review?

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+# Tutorial: Get Started with SignalR for ASP.NET Core
+
+By [Rachel Appel](https://twitter.com/rachelappel)
+
+This tutorial teaches the basics of building a real-time app using SignalR for ASP.NET Core. 
+
+This tutorial demonstrates the following SignalR development tasks:
+
+> [!div class="checklist"]
+> * Add the SignalR library to an ASP.NET web application.
```
scottaddie: ~~ASP.NET web application~~
ASP.NET Core web app

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+> * Create a SignalR hub to push content to clients.
+> * Modify the `Startup` class to configure the application.
```
scottaddie: application --> app

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+
+  ![New Project dialog in Visual Studio](get-started-signalr-core/_static/signalr-new-project-dialog.png)
+
+1. In the next step, choose **Web Application** to use a lightweight project template for using Razor Pages. Then select **Ok**.
+
+  ![New Project dialog in Visual Studio](get-started-signalr-core/_static/signalr-new-project-choose-type.png)
+
+1. Add the `Microsoft.AspNetCore.SignalR` NuGet package to the ASP.NET Core project. You can do this by selecting **Tools** > **NuGet Package Manager** > **Package Manager Console**. In the console, issue  the following command:
+
+  ```text
```
scottaddie: Change the "text" code fence to "powershell".

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+
+1. Use the **File** > **New Project** menu option and choose **ASP.NET Core Web Application**.
+
+  ![New Project dialog in Visual Studio](get-started-signalr-core/_static/signalr-new-project-dialog.png)
+
+1. In the next step, choose **Web Application** to use a lightweight project template for using Razor Pages. Then select **Ok**.
+
+  ![New Project dialog in Visual Studio](get-started-signalr-core/_static/signalr-new-project-choose-type.png)
+
+1. Add the `Microsoft.AspNetCore.SignalR` NuGet package to the ASP.NET Core project. You can do this by selecting **Tools** > **NuGet Package Manager** > **Package Manager Console**. In the console, issue  the following command:
```
scottaddie: - Convert this 1st occurrence of the package name to a link that goes here: https://www.nuget.org/packages/Microsoft.AspNetCore.SignalR
- Remove the extra space after "issue".

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Configure the project to use SignalR
+
+The SignalR server must be configured so that it knows to pass requests to SignalR.
+
+1. To configure a SignalR project, modify the `ConfigureServices` method of the application's `Startup` class by inserting a call to `services.AddSignalR()`.
+
+  This adds SignalR as part of the [ASP.NET Core middleware](xref:fundamentals/middleware/index) pipeline.
+
+1. In the `Startup` class you must also configure SignalR routes by calling `UseSignalR`. In the call to `UseSignalR` is where you setup the SignalR hub mappings.
```
scottaddie: Add a comma after "class".

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+
+  [!code-csharp[Startup](get-started-signalr-core/sample/Startup.cs?highlight=22,40-43)]
+
+## Create the SignalR client code
+
+1. Open and replace the content in `\Pages\Index.cshtml` with the following code:
```
scottaddie: Italicize Pages\Index.cshtml instead of code fencing it

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+
+  [!code-html[Index](get-started-signalr-core/sample/Index.cshtml)]
+
+  The preceding HTML displays name and message fields, and a submit button. Notice the script references at the bottom: a reference to SignalR, and a reference to *chat.js*, the custom SignalR client code.
+
+1. Add a JavaScript file to the *\wwwroot\js\* folder named `chat.js` and add the following code to it:
```
scottaddie: Italicize chat.js instead of code fencing it

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+
+  ```text
+  Install-Package Microsoft.AspNetCore.SignalR -Version 1.0.0-alpha2-final
+  ```
+
+  This command installs the `Microsoft.AspNetCore.SignalR` NuGet package that contains libraries SignalR server-side code.
+
+1. Install the JavaScript client library by using npm. `npm install` creates a *\node_modules\* folder and sub-folders for the libraries in the same location as where you run the npm command.
+
+  ```
```
scottaddie: - Add "console" after the 3 back ticks.
- I recommend specifying "--save" in the install command, so that the dependency is saved in *package.json*.

aspnetcore/toc.md
```diff
@@ -212,6 +212,9 @@
 # [Mobile](xref:mobile/index)
 ## [Create backend services for native mobile apps](xref:mobile/native-mobile-backend)
 
+# [SignalR](xref:signalr/index)
+## [Get Started](xref:signalr/get-started-signalr-core)
```
scottaddie: Get Started --> Get started

aspnetcore/signalr/index.md
```diff
@@ -0,0 +1,15 @@
...
+ms.author: rachelap
+ms.date: 02/25/2018
+ms.prod: asp.net-core
+ms.technology: aspnet
+ms.topic: article
+uid: signalr/index
+---
+# SignalR
+
+* [Get Started with SignalR for ASP.NET Core](get-started-signalr-core.md)
```
scottaddie: Started --> started

aspnetcore/signalr/get-started-signalr-core/sample/Startup.cs
```diff
@@ -0,0 +1,53 @@
...
+            else
+            {
+                app.UseExceptionHandler("/Error");
+            }
+
+            app.UseStaticFiles();
+
+            app.UseSignalR(routes =>
+            {
+                routes.MapHub<ChatHub>("chathub");
```
davidfowl: Make the url chat

aspnetcore/signalr/get-started-signalr-core/sample/chat.js
```diff
@@ -0,0 +1,17 @@
+﻿var connection = new signalR.HubConnection('chathub');
+connection.start().catch(err => appendLine(err, 'red'));
+
+connection.on('Send', (timestamp, user, message) => {
+    var listItem = document.createElement('li');
+    listItem.innerHTML = timestamp + ' <b>' + user + '</b>: ' + message;
```
davidfowl: We should html encode the content that's coming from arguments to prevent XSS

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+
+1. Copy the *signalr-client-1.0.0-alpha2-final.js* for SignalR from *\node_modules\@aspnet\signalr-client\dist\browser* to the *wwwroot\lib* folder in your project.
+
+## Create the SignalR Hub
+
+A hub is a class that represents a high-level pipeline that allows the client and server to call methods on each other directly.
+
+1. Add a class to the project by choosing **File** > **New** > **File** and selecting **Visual C# Class**. 
+
+1. Inherit from `Microsoft.AspNetCore.SignalR.Hub`. The `Hub` class contains properties and events for managing connections and groups, as well as sending and receiving data.
```
Rick-Anderson: I'd delete these two step and tell them to copy the following code.

after the code

The preceding code inherits ....
Give them all the details you did above.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+
+1. Create the `Send` method that sends a message to all connected chat clients. Notice it returns a `Task`, because SignalR is asynchronous and scales better.
+
+  [!code-csharp[Startup](get-started-signalr-core/sample/ChatHub.cs?range=7-14)]
+
+## Configure the project to use SignalR
+
+The SignalR server must be configured so that it knows to pass requests to SignalR.
+
+1. To configure a SignalR project, modify the `ConfigureServices` method of the application's `Startup` class by inserting a call to `services.AddSignalR()`.
```
Rick-Anderson: ~ To configure a SignalR project, ~

aspnetcore/signalr/get-started-signalr-core.md
Rick-Anderson: ~because SignalR is asynchronous and scales better.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+
+A hub is a class that represents a high-level pipeline that allows the client and server to call methods on each other directly.
+
+1. Add a class to the project by choosing **File** > **New** > **File** and selecting **Visual C# Class**. 
+
+1. Inherit from `Microsoft.AspNetCore.SignalR.Hub`. The `Hub` class contains properties and events for managing connections and groups, as well as sending and receiving data.
+
+1. Create the `Send` method that sends a message to all connected chat clients. Notice it returns a `Task`, because SignalR is asynchronous and scales better.
+
+  [!code-csharp[Startup](get-started-signalr-core/sample/ChatHub.cs?range=7-14)]
```
Rick-Anderson: Startup?
Show the entire file so they get includes.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Configure the project to use SignalR
+
+The SignalR server must be configured so that it knows to pass requests to SignalR.
```
Rick-Anderson: That doesn't make sense. So it know to pass requests to the SignalR clients?  SignalR by itself is meaningless.

aspnetcore/signalr/get-started-signalr-core.md
Rick-Anderson: We never append methods with `()`
It should be something like 
register the signalR service with the ASP.NET Core middleware pipeline. in `ConfigureServices`

aspnetcore/signalr/get-started-signalr-core.md
Rick-Anderson: Something like
Configure SignalR routes by `UseSignalR`.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+
+1. To configure a SignalR project, modify the `ConfigureServices` method of the application's `Startup` class by inserting a call to `services.AddSignalR()`.
+
+  This adds SignalR as part of the [ASP.NET Core middleware](xref:fundamentals/middleware/index) pipeline.
+
+1. In the `Startup` class you must also configure SignalR routes by calling `UseSignalR`. In the call to `UseSignalR` is where you setup the SignalR hub mappings.
+
+  [!code-csharp[Startup](get-started-signalr-core/sample/Startup.cs?highlight=22,40-43)]
+
```
Rick-Anderson: The `UseSignalR` call in the preceding code sets up hub mappings.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+## Create the SignalR client code
```
Rick-Anderson: ## Create the SignalR client ~code~

aspnetcore/signalr/get-started-signalr-core.md
Rick-Anderson: ~Open and~

Rick-Anderson: reformat markup so there is no horiz scroll bar using the widest window with no left or right chrome
image

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+
+1. Open and replace the content in `\Pages\Index.cshtml` with the following code:
+
+  [!code-html[Index](get-started-signalr-core/sample/Index.cshtml)]
+
+  The preceding HTML displays name and message fields, and a submit button. Notice the script references at the bottom: a reference to SignalR, and a reference to *chat.js*, the custom SignalR client code.
```
Rick-Anderson: Screaming for a list
The preceding markup:

aspnetcore/signalr/get-started-signalr-core.md
Rick-Anderson: Create *wwwroot/js/chat.js* with the following code:

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+
+1. Add a JavaScript file to the *\wwwroot\js\* folder named `chat.js` and add the following code to it:
+
+  [!code-html[Index](get-started-signalr-core/sample/chat.js)]
+
+## Verify the code works
+
+1. Select **Debug** > **Start without debugging** to launch a browser and load the website locally.
```
Rick-Anderson: Run the app.
That works xplat and this is not a super beginner tutorial.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+
+1. Enter a name and message, and click the **Send** button. You should see the name and message displayed in the page.
```
Rick-Anderson: ~You should see~ the name and message is displayed in the page.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+ms.technology: dotnet-signalr
+ms.prod: aspnet-core
+ms.custom: mvc
+---
+# Tutorial: Get Started with SignalR for ASP.NET Core
+
+By [Rachel Appel](https://twitter.com/rachelappel)
+
+This tutorial teaches the basics of building a real-time app using SignalR for ASP.NET Core. 
+
```
Rick-Anderson: Somewhere near the top show the completed app.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+> * Create a SignalR hub to push content to clients.
+> * Modify the `Startup` class to configure the application.
+> * Use the SignalR JavaScript library to send messages and display updates from the hub.
+
+# Prerequisites
+
+Install the following:
+
+* [.NET Core 2.0.0 SDK](https://www.microsoft.com/net/core) or later.
+* [Visual Studio 2017](https://www.visualstudio.com/downloads/) version 15.6 or later with the ASP.NET and web development workload.
```
Rick-Anderson: We should get rid of VS or use tabs. You shouldn't need VS to do this.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,107 @@
...
+
+1. Add the `Microsoft.AspNetCore.SignalR` NuGet package to the ASP.NET Core project. You can do this by selecting **Tools** > **NuGet Package Manager** > **Package Manager Console**. In the console, issue  the following command:
+
+  ```text
+  Install-Package Microsoft.AspNetCore.SignalR -Version 1.0.0-alpha2-final
+  ```
+
+  This command installs the `Microsoft.AspNetCore.SignalR` NuGet package that contains libraries SignalR server-side code.
+
+1. Install the JavaScript client library by using npm. `npm install` creates a *\node_modules\* folder and sub-folders for the libraries in the same location as where you run the npm command.
```
Rick-Anderson: NPM is not installed with the prerequisites you list.

Rick-Anderson:
@rachelappel  need to show SignalR server. The current app could be done as windows desktop app. How about at least showing two browsers chatting.

aspnetcore/signalr/get-started-signalr-core.md
scottaddie: It should be installed with the **ASP.NET and web development** workload. On Windows, you'd see it here: *C:\Program Files (x86)\Microsoft Visual Studio\Preview\Ent-15.6\Web\External\npm.cmd*

aspnetcore/signalr/get-started-signalr-core.md
scottaddie: If you remove VS from the list, then you'll need to add npm.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,110 @@
...
+
+1. Add the `Microsoft.AspNetCore.SignalR` NuGet package to the ASP.NET Core project. Select **Tools** > **NuGet Package Manager** > **Package Manager Console**. In the console, issue the following command:
+
+  ```powershell
+  Install-Package Microsoft.AspNetCore.SignalR –IncludePrerelease
+  ```
+
+  The preceeding command installs the [Microsoft.AspNetCore.SignalR](https://www.nuget.org/packages/Microsoft.AspNetCore.SignalR) NuGet package that contains libraries SignalR server-side code.
+
+1. Install the JavaScript client library by using npm. `npm install` creates a *\node_modules\* folder and sub-folders for the libraries in the same location as where you run the npm command.
```
scottaddie: Remove the 2 backslashes, so that the *node_modules* folder name appears italicized.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,110 @@
...
+
+  ```console
+   npm install @aspnet/signalr-client --save
+  ```
+
+1. Copy the *signalr-client-1.0.0-alpha2-final.js* for SignalR from *\node_modules\@aspnet\signalr-client\dist\browser* to the *wwwroot\lib* folder in your project.
```
scottaddie: Add another backslash before "@aspnet"

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,110 @@
...
+
+1. Replace the content in *\Pages\Index.cshtml* with the following code:
+
+  [!code-html[Index](get-started-signalr-core/sample/Index.cshtml)]
+
+  The preceding HTML displays name and message fields, and a submit button. Notice the script references at the bottom: a reference to SignalR and *chat.js*.
+
+1. Add a JavaScript file to the *\wwwroot\js\* folder named *chat.js* and add the following code to it:
+
+  [!code-html[Index](get-started-signalr-core/sample/chat.js)]
```
scottaddie: Use javascript as the snippet type

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Create the SignalR client code
+
+1. Replace the content in *\Pages\Index.cshtml* with the following code:
+
+  [!code-html[Index](get-started-signalr-core/sample/Index.cshtml)]
+
+  The preceding HTML displays name and message fields, and a submit button. Notice the script references at the bottom: a reference to SignalR and *chat.js*.
+
+1. Add a JavaScript file to the *\wwwroot\js\* folder named *chat.js* and add the following code to it:
```
scottaddie: Remove the backslashes on either end of the folder path.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,110 @@
...
+
+  This adds SignalR as part of the [ASP.NET Core middleware](xref:fundamentals/middleware/index) pipeline.
+
+1. In the `Startup` class, you must also configure SignalR routes by calling `UseSignalR`. In the call to `UseSignalR` is where you setup the SignalR hub mappings.
+
+  [!code-csharp[Startup](get-started-signalr-core/sample/Startup.cs?highlight=22,40-43)]
+
+## Create the SignalR client code
+
+1. Replace the content in *\Pages\Index.cshtml* with the following code:
```
scottaddie: Remove the leading backslash.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,110 @@
...
+By [Rachel Appel](https://twitter.com/rachelappel)
+
+This tutorial teaches the basics of building a real-time app using SignalR for ASP.NET Core. 
+
+  ![Solution](get-started-signalr-core/_static/signalr-get-started-finished.png)
+
+This tutorial demonstrates the following SignalR development tasks:
+
+> [!div class="checklist"]
+> * Add the SignalR library to an ASP.NET web app.
```
anurse: ASP.NET _Core_ ?

aspnetcore/signalr/get-started-signalr-core.md
anurse: We should keep this as cross-platform as we can. I don't mind if the tutorial focuses on the VS scenario but perhaps if we mention how to do things on non-Windows (`dotnet new razor ...` instead of `New Project...`, etc.).

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,110 @@
...
+
+  The preceding HTML displays name and message fields, and a submit button. Notice the script references at the bottom: a reference to SignalR and *chat.js*.
+
+1. Add a JavaScript file to the *\wwwroot\js\* folder named *chat.js* and add the following code to it:
+
+  [!code-html[Index](get-started-signalr-core/sample/chat.js)]
+
+## Run the app
+
+1. Select **Debug** > **Start without debugging** to launch a browser and load the website locally.
```
anurse: Let's instruct the user to open two browser windows (they can use the same browser, just two separate windows) and chat between them. That's the demo that really shows off the ✨magic ✨ of SignalR ;)

aspnetcore/signalr/get-started-signalr-core/sample/Startup.cs
```diff
@@ -0,0 +1,53 @@
...
+            else
+            {
+                app.UseExceptionHandler("/Error");
+            }
+
+            app.UseStaticFiles();
+
+            app.UseSignalR(routes =>
+            {
+                routes.MapHub<ChatHub>("chat");
```
anurse: More than once, I have accidentally created a `Chat.cshtml` razor page and a `chat` Hub (and I **really** should know better 😝). I've also seen customers do it. Of course, when that happens, one of them doesn't work (which one depends on the order of SignalR and MVC in `Configure`) because they are both mapped to the same route.

Can we start a precedent of mapping hubs to `/hubs/[HubName]` in our docs to ensure we keep Hubs in their own area of the URL space? (Thoughts @davidfowl ?).

aspnetcore/signalr/get-started-signalr-core/sample/chat.js
```diff
@@ -0,0 +1,25 @@
+﻿var connection = new signalR.HubConnection('chathub');
+connection.start().catch(err => showErr(err));
+
+connection.on('Send', (timestamp, user, message) => {
```
anurse: Put the call to `.on` before the call to `.start`. Otherwise you could technically miss invocations that occur immediately after the connection is started but before you've bound the handler. That's the main reason we have a `.start` method rather than just starting the connection in the constructor ;)

aspnetcore/signalr/get-started-signalr-core/sample/chat.js
```diff
@@ -0,0 +1,25 @@
...
+    var msg = document.getElementById('message').value;
+    var usr = document.getElementById('user').value;
+
+    connection.invoke('Send', usr, msg).catch(err => showErr(err));
+    event.preventDefault();
+});
+
+function showErr(msg) {
+    var listItem = document.createElement('li');
+    listItem.setAttribute("font-color", "red");
```
anurse: Use `style` (`.setAttribute("style", "color: red")`). I don't think `font-color` works... but I may be wrong.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,110 @@
...
+> * Add the SignalR library to an ASP.NET web app.
+> * Create a SignalR hub to push content to clients.
+> * Modify the `Startup` class to configure the app.
+> * Use the SignalR JavaScript library to send messages and display updates from the hub.
+
+# Prerequisites
+
+Install the following:
+
+* [.NET Core 2.0.0 SDK](https://www.microsoft.com/net/core) or later.
```
anurse: [Preview 1 update] requires `.NET Core 2.1.0 Preview 1 SDK`

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,110 @@
...
+* [Visual Studio 2017](https://www.visualstudio.com/downloads/) version 15.6 or later with the ASP.NET and web development workload.
+
+## Create an ASP.NET Core project that uses SignalR
+
+1. Use the **File** > **New Project** menu option and choose **ASP.NET Core Web Application**.
+
+  ![New Project dialog in Visual Studio](get-started-signalr-core/_static/signalr-new-project-dialog.png)
```
anurse: [Preview 1] Make sure to tell the user to create an ASP.NET Core **2.1** project

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,110 @@
...
+  ```powershell
+  Install-Package Microsoft.AspNetCore.SignalR –IncludePrerelease
+  ```
+
+  The preceeding command installs the [Microsoft.AspNetCore.SignalR](https://www.nuget.org/packages/Microsoft.AspNetCore.SignalR) NuGet package that contains libraries SignalR server-side code.
+
+1. Install the JavaScript client library by using npm. `npm install` creates a *\node_modules\* folder and sub-folders for the libraries in the same location as where you run the npm command.
+
+  ```console
+   npm install @aspnet/signalr-client --save
```
anurse: [Preview 1] The package is `@aspnet/signalr` (i.e ~~`-client`~~) in Preview 1

aspnetcore/signalr/get-started-signalr-core.md
anurse: [Preview 1] Script is now just `signalr.js` (same path, but with `signalr-client` changed to `signalr`)

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,110 @@
...
+  ![New Project dialog in Visual Studio](get-started-signalr-core/_static/signalr-new-project-dialog.png)
+
+1. Select **Web Application** which creates a project using Razor Pages. Then select **Ok**.
+
+  ![New Project dialog in Visual Studio](get-started-signalr-core/_static/signalr-new-project-choose-type.png)
+
+1. Add the `Microsoft.AspNetCore.SignalR` NuGet package to the ASP.NET Core project. Select **Tools** > **NuGet Package Manager** > **Package Manager Console**. In the console, issue the following command:
+
+  ```powershell
+  Install-Package Microsoft.AspNetCore.SignalR –IncludePrerelease
```
anurse: [Preview 1] Not needed in Preview 1. SignalR is part of the ASP.NET Core metapackages in 2.1 (and thus already present in the template)

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,110 @@
+---
+uid: signalr/get-started-signalr-core
+title: Get Started with SignalR on ASP.NET Core
+author: rachelappel
+ms.author: rachelap
+description: In this tutorial, you will create a SignalR on ASP.NET Core application.
```
scottaddie: There seems to be a word missing after "a SignalR".

rachelappel:
@anurse Ok changes are in. I think I got everything. Thumbs up?

Rick-Anderson:
@rachelappel  still need to make this xplat. You don't need to support VS for mac on this version (but create an issue for that). We don't want to give the impression VS is a prereq for SignalR core.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
```
Rick-Anderson: ~will~ no future tense

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+By [Rachel Appel](https://twitter.com/rachelappel)
+
+This tutorial teaches the basics of building a real-time app using SignalR for ASP.NET Core.
+
+  ![Solution](get-started-signalr-core/_static/signalr-get-started-finished.png)
+
+This tutorial demonstrates the following SignalR development tasks:
+
+> [!div class="checklist"]
+> * Create a SignalR on ASP.NET Core web app.
```
Rick-Anderson: SignalR  what? client?

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+> * Create a SignalR hub to push content to clients.
+> * Modify the `Startup` class to configure the app.
```
Rick-Anderson: Either delete this step or tell what you're doing
Add SignalR to the service container and ...

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+> * Use the SignalR JavaScript library to send messages and display updates from the hub.
+
+# Prerequisites
+
+Install the following:
```
Rick-Anderson: add VS and CLI tabs (approved tabs)

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+* [.NET Core 2.1.0 Preview 1 SDK](https://www.microsoft.com/net/core) or later.
+* [Visual Studio 2017](https://www.visualstudio.com/downloads/) version 15.6 or later with the ASP.NET and web development workload.
+
+## Create an ASP.NET Core project that uses SignalR
+
+1. Use the **File** > **New Project** menu option and choose **ASP.NET Core Web Application**.
+
+  ![New Project dialog in Visual Studio](get-started-signalr-core/_static/signalr-new-project-dialog.png)
+
+1. Select **Web Application** which creates a project using Razor Pages. Then select **Ok**. Be sure that **ASP.NET Core 2.1** is selected from the framework selector.
```
Rick-Anderson: @anurse  can we add something like. The Razor Pages app is used to create a basic client app and host the server. I'd like to tell they why we're using RP (because it's easy) but give them an idea of what they could do with other clients.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+
+  ![New Project dialog in Visual Studio](get-started-signalr-core/_static/signalr-new-project-choose-type.png)
+
+The libraries that contain SignalR server-side code are included in the project template. You must include the client side JavaScript file.
+
+1. Install the JavaScript client library by using npm. `npm install` creates a *node_modules* folder and sub-folders for the libraries in the same location as where you run the npm command.
```
Rick-Anderson: `Install the JavaScript client library with [npm](link).` Tell them to open a console window and where to open it.

aspnetcore/signalr/get-started-signalr-core.md
Rick-Anderson: Need to tell them to name it SignalRChat so the ns matches.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+> * Use the SignalR JavaScript library to send messages and display updates from the hub.
+
+# Prerequisites
+
+Install the following:
+
+* [.NET Core 2.1.0 Preview 1 SDK](https://www.microsoft.com/net/core) or later.
+* [Visual Studio 2017](https://www.visualstudio.com/downloads/) version 15.6 or later with the ASP.NET and web development workload.
+
+## Create an ASP.NET Core project that uses SignalR
```
Rick-Anderson: How about hosts SignalR client and server?

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+  ![Solution](get-started-signalr-core/_static/signalr-get-started-finished.png)
+
+This tutorial demonstrates the following SignalR development tasks:
+
+> [!div class="checklist"]
+> * Create a SignalR on ASP.NET Core web app.
+> * Create a SignalR hub to push content to clients.
+> * Modify the `Startup` class to configure the app.
+> * Use the SignalR JavaScript library to send messages and display updates from the hub.
+
```
Rick-Anderson: Add in download link and instructions to download.

rachelappel:
@Rick-Anderson Yep - The xplat is going in through a separate PR as an update to the article. I'll make an issue, forgot to do that earlier with the push.

Rick-Anderson:
@rachelappel  don't merge until you get approval from me and answer my questions or list them in new issues.

aspnetcore/signalr/get-started-signalr-core.md
scottaddie: I think this should say something like "Create an ASP.NET Core SignalR app".

rachelappel: "Create a SignalR on ASP.NET Core web app" is the thing development task they will do.   
Do you mean to add something like "containing server and client libraries"? The question isn't clear.
For consistency - None of the other tutorials describe what's in a proj template in the "development tasks" checklist.  The steps are supposed to say what you'll accomplish, not describe what is in the product.  What is in the proj template is covered during the tutorial anyway.

aspnetcore/signalr/get-started-signalr-core.md
rachelappel: Tabs are going in the xplat PR

aspnetcore/signalr/get-started-signalr-core.md
Rick-Anderson: >Create a SignalR on ASP.NET Core web app.

Doesn't make any sense. A SignalR what?
The template is secondary to the tutorial.  

> "containing server and client libraries

Yes, or to host the SignalR server and client. It's more than the client libs, you supply the client app.

Rick-Anderson: >Create an ASP.NET Core SignalR app

But there's really no such thing. We're creating a client and server app  together.  We should try to make that clear - client and server sides of SignalR.

anurse: How about just: "Create an ASP.NET Core web app with a SignalR Hub".

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,108 @@
...
+> [!div class="checklist"]
+> * Create a SignalR on ASP.NET Core web app.
+> * Create a SignalR hub to push content to clients.
+> * Use the SignalR JavaScript library to send messages and display updates from the hub.
+
+# Prerequisites
+
+Install the following:
+
+* [.NET Core 2.1.0 Preview 1 SDK](https://www.microsoft.com/net/core) or later.
```
anurse: This link is for 2.0.0 RTM, it should be updated to `https://www.microsoft.com/net/download/dotnet-core/sdk-2.1.300-preview1`

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,108 @@
...
+1. Select **Web Application** which creates a project using Razor Pages. Then select **Ok**. Be sure that **ASP.NET Core 2.1** is selected from the framework selector.
+
+  ![New Project dialog in Visual Studio](get-started-signalr-core/_static/signalr-new-project-choose-type.png)
+
+The libraries that contain SignalR server-side code are included in the project template. You must include the client side JavaScript file.
+
+1. Install the JavaScript client library with [npm](https://www.npmjs.com/). Open a command window at the root of your project and run `npm install`. `npm install` creates a *node_modules* folder and sub-folders for the libraries in the same location as where you run the npm command.
+
+  ```console
+   npm install @aspnet/signalr-client --save
```
anurse: The module ID is now `@aspnet/signalr`, also `--save` is the default, so you can drop it from the invocation.

aspnetcore/signalr/get-started-signalr-core/sample/Hubs/ChatHub.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Microsoft.AspNetCore.SignalR;
+using System;
+using System.Threading.Tasks;
+
+namespace SignalRChat.Hubs
+{
+    public class ChatHub : Hub
+    {
+       public Task Send(string user, string message)
```
anurse: indentation is off ;)

aspnetcore/signalr/get-started-signalr-core/sample/NuGet.config
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+    <packageSources>
+        <clear />
+        <add key="AspNetVNext" value="https://dotnet.myget.org/F/aspnetcore-dev/api/v3/index.json" />
```
anurse: Not needed now that preview1 is out! 🎉

anurse: The whole config file can be removed, since only NuGet.org is needed.

aspnetcore/signalr/get-started-signalr-core/sample/Startup.cs
```diff
@@ -0,0 +1,53 @@
...
+            else
+            {
+                app.UseExceptionHandler("/Error");
+            }
+
+            app.UseStaticFiles();
+
+            app.UseSignalR(routes =>
+            {
+                routes.MapHub<ChatHub>("ChatHub");
```
anurse: Use `/ChatHub` (leading slash is required in Preview 1)

aspnetcore/signalr/get-started-signalr-core/sample/wwwroot/js/chat.js
```diff
@@ -0,0 +1,26 @@
+﻿var connection = new signalR.HubConnection('chathub');
```
anurse: You're using arrow functions above (`() => {`) which is awesome. All the browsers which support arrow functions also support `let` (for mutable variables) and `const` for immutable variables. Let's update our scripts to use best practices. So if a variable is modified, use `let` instead of `var` and if it isn't modified use `const`.

anurse: All the variables in this script are immutable (you don't replace the value) so replace all the `var`s with `const`

Rick-Anderson:
@rachelappel  I installed the latest SDK cli build 20180218-02-1397458
and the latest VS 
Microsoft Visual Studio Enterprise 2017 Int Preview
Version 15.6.0 Preview 7.0 [27426.0.d15.6]
VisualStudio.15.IntPreview/15.6.0-pre.7.0+27426.0.d15.6

npm and node.exe are in C:\Program Files (x86)\Microsoft Visual Studio\Preview\Enterprise\Web\External
but from the console I get
C:\Users\riande\source\1\SignalRChat\SignalRChat>npm install @aspnet/signalr-client --save
'npm' is not recognized as an internal or external command,

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,108 @@
+---
+uid: signalr/get-started-signalr-core
+title: Get Started with SignalR on ASP.NET Core
+author: rachelappel
+ms.author: rachelap
+description: In this tutorial, you create a SignalR on ASP.NET Core application.
```
Rick-Anderson: a real-time app using SignalR for ASP.NET Core.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,108 @@
...
+1. Use the **File** > **New Project** menu option and choose **ASP.NET Core Web Application**.
+1. Name the project `SignalRChat`
+
+  ![New Project dialog in Visual Studio](get-started-signalr-core/_static/signalr-new-project-dialog.png)
+
+1. Select **Web Application** which creates a project using Razor Pages. Then select **Ok**. Be sure that **ASP.NET Core 2.1** is selected from the framework selector.
+
+  ![New Project dialog in Visual Studio](get-started-signalr-core/_static/signalr-new-project-choose-type.png)
+
+The libraries that contain SignalR server-side code are included in the project template. You must include the client side JavaScript file.
```
Rick-Anderson: @scottaddie  and Acrolynx prefer client-side

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,108 @@
...
+
+1. Install the JavaScript client library with [npm](https://www.npmjs.com/). Open a command window at the root of your project and run `npm install`. `npm install` creates a *node_modules* folder and sub-folders for the libraries in the same location as where you run the npm command.
```
Rick-Anderson: subfolders

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Configure the project to use SignalR
+
+The SignalR server must be configured so that it knows to pass requests to SignalR.
+
+1. To configure a SignalR project, modify the `ConfigureServices` method of the application's `Startup` class by inserting a call to `services.AddSignalR()`.
+
+  This adds SignalR as part of the [ASP.NET Core middleware](xref:fundamentals/middleware/index) pipeline.
+
+1. In the `Startup` class, you must also configure SignalR routes by calling `UseSignalR`. In the call to `UseSignalR` is where you setup the SignalR hub mappings.
```
Rick-Anderson: @anurse  can you here?
Do you config routes our setup hub mappings. Why do you need two sentences? I think it reads better something like
Configure SignalR routes and hub mappings by calling `UseSignalR`:
No need to tell them about Startup again. also is a banned word.

aspnetcore/signalr/get-started-signalr-core/sample/Hubs/ChatHub.cs
```diff
@@ -0,0 +1,15 @@
...
+        {
+            string timestamp = DateTime.Now.ToShortTimeString();            
+            return Clients.All.InvokeAsync("Send", timestamp, user, message);
```
tugberkugurlu: Should this be `SendAsync`? I get intellisense for SendAsync for `1.0.0-preview1-final`:

image

anurse: Indeed. It is SendAsync in preview 1. I missed that. Thanks @tugberkugurlu

aspnetcore/signalr/get-started-signalr-core.md
anurse: I agree, it doesn't need to specified twice.

anurse: Also, let's refer to them as "hub routes" (since we use the same routing system as MVC, which refers to them as routes.

anurse: I think just "Configure routes to your hubs using `UseSignalR`". "Hub Mappings" are not really a thing. A Hub has a route that defines the URLs that map to it.

anurse:
> npm and node.exe are in C:\Program Files (x86)\Microsoft Visual Studio\Preview\Enterprise\Web\External

It's possible we'll need to find a VS way to install the package. I don't think npm is put on the `PATH` by VS.

scottaddie:
@spboyer See the comment from @anurse ☝️ . This is related to your email from earlier this week. It should be easier to get going with npm in VS.

rachelappel:
@anurse After checking, correct. No path in VS. I have the Node tools installed but that doesn't me access to npm in Visual Studio, and I thought it was supposed to be part of the install. It then seemed easier to just go command line. Do you have ideas/suggestions or is it supposed to work in VS and mine just isn't working? Otherwise, there are some add-ins for node or pkg mgmt. Add copy in the doc for installing and using one of them?

anurse:
@rachelappel At the moment, NPM is the only way to get the package. If VS doesn't come with it, perhaps we can just add Node as a prerequisite? It's not ideal, but it'll work

aspnetcore/signalr/get-started-signalr-core/sample/wwwroot/js/chat.js
```diff
@@ -0,0 +1,26 @@
+﻿const connection = new signalR.HubConnection('chathub');
```
anurse: For consistency, let's make this `/chatHub` (leading slash)

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+> * Create a SignalR hub to push content to clients.
+> * Use the SignalR JavaScript library to send messages and display updates from the hub.
+
+# Prerequisites
+
+Install the following software:
+
+* [.NET Core 2.1.0 Preview 1 SDK](https://www.microsoft.com/net/download/dotnet-core/sdk-2.1.300-preview1) or later
+* [Visual Studio 2017](https://www.visualstudio.com/downloads/) version 15.6 or later with the ASP.NET and web development workload
+* [npm](https://www.npmjs.com/get-npm)
```
scottaddie: This still doesn't seem accurate to me. I'm almost 100% confident that the ASP.NET workload in VS 2017 15.6 adds npm; however, npm doesn't get added to the path. @jodavis or @justcla, can one of you please confirm?

jodavis: That's correct, we install a sandboxed version of node and npm so we can provide some built-in tools without requiring a full install of Node.js. However, we don't add it to the path generally because it's a somewhat stripped down version and it gets out of date fast. You're better off installing Node.js properly if you want to use it for things.

Installing Node.js should add it to the path if you select the option, and that should make it available in VS, assuming you restarted VS after it was added to the path. If VS was launched from a command prompt, then that also needs to be restarted after Node.js is added to the path.

rachelappel:
@anurse Do you spot anything else or does this look good? Shipit squirrel?

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
+---
+uid: signalr/get-started-signalr-core
+title: Get started with SignalR on ASP.NET Core
+author: rachelappel
+ms.author: rachelap
+description: In this tutorial, you create an app using SignalR for ASP.NET Core.
+manager: wpickett
+ms.date: 02/23/2018
+ms.topic: tutorial
+ms.technology: dotnet-signalr
```
scottaddie: Shouldn't this value be "aspnet"?

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+---
+# Tutorial: Get started with SignalR for ASP.NET Core
+
+By [Rachel Appel](https://twitter.com/rachelappel)
+
+This tutorial teaches the basics of building a real-time app using SignalR for ASP.NET Core.
+
+   ![Solution](get-started-signalr-core/_static/signalr-get-started-finished.png)
+
+[View or download the sample](https://github.com/aspnet/Docs/tree/master/aspnetcore/signalr/get-started-signalr-core/sample/). [View download instructions](/tutorials/#how-to-download-a-sample).
```
scottaddie: Let's make this consistent with how it's used elsewhere. I suggest the following:
```
[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/signalr/get-started-signalr-core/sample/) ([how to download](xref:tutorials/index#how-to-download-a-sample))
```

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+
+This tutorial demonstrates the following SignalR development tasks:
+
+> [!div class="checklist"]
+> * Create an ASP.NET Core web app with a SignalR Hub.
```
scottaddie: Hub --> hub

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+> * Create a SignalR hub to push content to clients.
+> * Use the SignalR JavaScript library to send messages and display updates from the hub.
+
+# Prerequisites
+
+Install the following software:
+
+* [.NET Core 2.1.0 Preview 1 SDK](https://www.microsoft.com/net/download/dotnet-core/sdk-2.1.300-preview1) or later
+* [Visual Studio 2017](https://www.visualstudio.com/downloads/) version 15.6 or later with the ASP.NET and web development workload
```
scottaddie: ASP.NET and web development --> **ASP.NET and web development**

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+* [npm](https://www.npmjs.com/get-npm)
+
+## Create an ASP.NET Core project that hosts SignalR client and server
+
+1. Use the **File** > **New Project** menu option and choose **ASP.NET Core Web Application**. Name the project `SignalRChat`
```
scottaddie: Missing a period at the end of this 2nd sentence.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+
+  ![New Project dialog in Visual Studio](get-started-signalr-core/_static/signalr-new-project-dialog.png)
+
+2. Select **Web Application** to create a project using Razor Pages. Then select **Ok**. Be sure that **ASP.NET Core 2.1** is selected from the framework selector.
```
scottaddie: Should we also indicate that this works when targeting either .NET Framework or .NET Core?

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+
+  ![New Project dialog in Visual Studio](get-started-signalr-core/_static/signalr-new-project-choose-type.png)
+
+  The libraries that host SignalR server-side code are included in the project template. Install the client-side JavaScript separately with [npm](https://www.npmjs.com/).
+
+  ```console
+   npm install @aspnet/signalr
+  ```
+
+3. Copy the *signalr.js* for SignalR from *node_modules\\@aspnet\signalr\dist\browser* to the *wwwroot\lib* folder in your project.
```
scottaddie: Missing the word "file" before "for SignalR"?

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+
+1. Configure routes to your hubs using `UseSignalR`.
+
+  [!code-csharp[Startup](get-started-signalr-core/sample/Startup.cs?highlight=22,40-43)]
+
+## Create the SignalR client code
+
+1. Replace the content in *Pages\Index.cshtml* with the following code:
+
+  [!code-html[Index](get-started-signalr-core/sample/Pages/Index.cshtml)]
```
scottaddie: I suggest using `code-cshtml` instead. This makes it clearer to the reader that it's not plain old HTML.

aspnetcore/signalr/index.md
scottaddie: Just a reminder to update this to sentence case

aspnetcore/toc.md
scottaddie: Just a reminder to update this to sentence case

aspnetcore/signalr/get-started-signalr-core.md
scottaddie: @rachelappel Since we're telling the user to install npm, I think we should also have them modify the **Tools** > **Options** > **Projects and Solutions** > **Web Package Management** > **External Web Tools** setting in VS. Specifically, move `$(PATH)` immediately above `$(VSINSTALLDIR)\Web\External`. This would give their fresh install of npm precedence over what ships with VS.

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+This tutorial teaches the basics of building a real-time app using SignalR for ASP.NET Core.
+
+   ![Solution](get-started-signalr-core/_static/signalr-get-started-finished.png)
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/signalr/get-started-signalr-core/sample/) ([how to download](xref:tutorials/index#how-to-download-a-sample))
+
+This tutorial demonstrates the following SignalR development tasks:
+
+> [!div class="checklist"]
+> * Create an ASP.NET Core web app with a SignalR hub.
```
anurse: Now that I'm reading the whole list, I think the "SignalR hub" part becomes redundant with the second item. Maybe just:

* Create an ASP.NET Core Web App
* Create a SignalR Hub to push content to clients
* Use the SignalR JavaScript library to send messages and display updates from the hub.

aspnetcore/signalr/get-started-signalr-core/sample/wwwroot/js/chat.js
```diff
@@ -0,0 +1,26 @@
...
+    const usr = document.getElementById('user').value;
+
+    connection.invoke('Send', usr, msg).catch(err => showErr(err));
+    event.preventDefault();
+});
+
+function showErr(msg) {
+    const listItem = document.createElement('li');
+    listItem.setAttribute("style", "color: red");
+    listItem.innerHTML = msg;
```
anurse: Hrm, let's use `innerText` here. `innerHTML` could technically be used to inject HTML. Technically the error content comes from SignalR, which should be trustworthy, but it's better to use `innerText` because it should be treated as plain text. Also, we should use `.toString`.

So this would be

```javascript
listItem.innerText = msg.toString();
```

aspnetcore/signalr/get-started-signalr-core.md
```diff
@@ -0,0 +1,106 @@
...
+1. Open another browser instance (any browser) and paste the URL in the address bar.
+
+1. Choose either browser, enter a name and message, and click the **Send** button. The name and message are displayed on both pages instantly.
+
+  ![Solution](get-started-signalr-core/_static/signalr-get-started-finished.png)
+
+## Next Steps
+
+
+[Publish to Azure](xref:tutorials/publish-to-azure-webapp-using-vs)
```
anurse: There are a few extra steps needed when publishing a SignalR app to Azure App Service. You need to manually enable WebSockets and ARR Affinity. Should we remove this link until we can have a proper "Publish a SignalR App to Azure" article?

rachelappel: @anurse  Good idea, I forgot about the Websockets thing. I'll remove until we have something better to direct the reader to.

PR #5547, opened 2018-02-24
Drop [Main] from snippet references

guardrex:
`[Main]` was apparently copied early from DocFX markdown doc examples and habitually persisted. Since they aren't usually accurate or helpful, I just did a quick find-and-replace to knock them out.

I scanned all of these, and they look fine ... *but get the eye drops out again!* 👁


scottaddie:
This PR is sponsored by Visine eye drops 😄

ovation22:
This removes the link to the code within GitHub. Not sure if this was intended/acceptable.

guardrex:
@ovation22

> This removes the link to the code within GitHub.

Not sure what u mean ... can you elaborate?

ovation22:
When navigating the Docs/MD in GitHub the [Main] appears as a link to the code. Removing Main removes the link when navigating the MD in GitHub. This isn't a concern if navigating the Docs from the main site, but removes the ability to easily navigate to the code in question with the link removed. This causes a slight issue as the code isn't displayed when viewing the MD in GitHub, but not an issue when viewing the Docs from the main site.

GitHub:
https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/dependency-injection.md
[!code-csharp] <-- used to have a link

GitHub History:
https://github.com/aspnet/Docs/blob/a4aa4ebb98e5788bf78254e4f4a1b518a1f2454f/aspnetcore/fundamentals/dependency-injection.md
[!code-csharpMain] <-- link to code

Docs:
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection

Not sure it's an issue, was discussed, or even noticed before.

guardrex:
@ovation22 Thanks for explaining. We anticipate that the vast majority of doc consumers will browse the built pages. The lack of snippet rendering (and now linking) isn't the only issue: All of the relative links, all of the API links, and at least some of the `xref` doc cross-links are also broken. There's no handy TOC. The unstyled markdown is rough on the :eye: :smile:. It's just never going to be a good general experience.

@scottaddie @Rick-Anderson I'm not too concerned about it; but since it's your show and you might be, an OPS solution might be possible before we go with something like `[LINK]`, `[See snippet]`, or something along those lines. I mean if you have a concern about it ... personally, I wouldn't sweat it.

PR #5582, opened 2018-03-02
Add lockout feature content

### LINKED ISSUE #1195: No docs on Identity "Lockout" feature
The docs at http://docs.asp.net/en/latest/security/authentication/index.html cover most aspects of Identity, but they don't mention the account lockout feature.

It's not easy to discover how to enable this. Even if you discover the `Lockout` property on the `AddIdentity` callback in `Startup.cs`, that's not enough - you also have to modify `AccountController`'s `Login` method to pass `lockoutOnFailure: true`.

guardrex:
Fixes #1195

### DIFF
--- a/aspnetcore/security/authentication/2fa.md
+++ b/aspnetcore/security/authentication/2fa.md
@@ -137,7 +137,7 @@ If you don't get a text message, see twilio log page.
 
 ## Account lockout for protecting against brute force attacks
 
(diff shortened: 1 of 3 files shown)


guardrex:
@Rick-Anderson @scottaddie @rachelappel There was a minor hiccup getting this to build at first.

It threw a :boom: ...

> Error happened in build.entrypoint.ps1. Please build your target branch first and then reopen this pull request to do another round of verification. If the issue still happens, open a ticket in http://SiteHelp and include the following error message for our team to troubleshoot: Method invocation failed because [System.Collections.Hashtable] does not contain a method named 'Kill'.

... then it built ok after closing/re-opening. 😅

guardrex:
@SteveSandersonMS I think ur presenting Blazor right now :smile: ... see u there!

@HaoK ... are u free to take a look?

HaoK:
Looks fine, but maybe its worth mentioning that any successful login will reset the failed count.  So basically it takes a number of consecutive unsuccessful logins to trigger lockout, as opposed to an aggregate over time.

guardrex:
@HaoK Made an addition to the text ...

> Any successful login resets the failed access attempts count and resets the clock.

scottaddie:
@HaoK Please review once more.

guardrex:
We're advised not to use "sign in" as a noun. @scottaddie, would like to just go for it, or would u prefer I refactor that to avoid the noun.

guardrex:
Frankly, this direction ...

> Avoid using as a noun or adjective (sign-in, sign-out). Instead, talk about what the customer needs to do.

... doesn't work for technical guidance. A "sign-in" is a thing that happens. If we force ourselves to say something like "authenticated," it feels less meaningful and clear. I feel the reader thinks that they are "signing in" not "authenticating themselves."

Stand-by tho ... I want to hyphenate it.

PR #5587, opened 2018-03-02
Iterating the PendingRequirements collection and calling Succeed on a…

rianjs:
Iterating the `PendingRequirements` collection and calling `Succeed()` on an element modifies the underlying collection which triggers an `InvalidOperationException`. This behavior isn't going to be changed: https://github.com/aspnet/Security/issues/1375

### DIFF
--- a/aspnetcore/security/authorization/policies/samples/PoliciesAuthApp1/Services/Handlers/PermissionHandler.cs
+++ b/aspnetcore/security/authorization/policies/samples/PoliciesAuthApp1/Services/Handlers/PermissionHandler.cs
@@ -10,7 +10,9 @@
     {
         public Task HandleAsync(AuthorizationHandlerContext context)
         {
-            foreach (var requirement in context.PendingRequirements)
+            var pendingRequirements = context.PendingRequirements.ToList();
+
+            foreach (var requirement in pendingRequirements)
             {
                 if (requirement is ReadPermission)
                 {


rianjs:
The offending docs page with the broken example is here: https://docs.microsoft.com/en-us/aspnet/core/security/authorization/policies

scottaddie:
@HaoK Please review. Is using the `Requirements` property instead of `PendingRequirements` still the recommended workaround? Is there anything planned for 2.1 to address this?

HaoK:
To be even safer, maybe you should copy the pending requirements via ToList() or something

HaoK:
No you shouldn't use Requirements instead of PendingRequirements as those will give you ones that are already succeeded, I would just do `PendingRequirements.ToList()`

scottaddie:
@HaoK Are you suggesting something like the following?

```csharp
var requirements = context.PendingRequirements.ToList();

foreach (var requirement in requirements)
{
}
```

HaoK:
Exactly

scottaddie:
@rianjs Please update your PR accordingly, and then we'll get this merged. Thank you for your help!

rianjs:
I considered that solution before I made the docs change, but decided the `Requirements` enumeration was better since it avoided the memory allocation + GC overhead on a fairly hot path.

> No you shouldn't use Requirements instead of PendingRequirements as those will give you ones that are already succeeded, I would just do PendingRequirements.ToList()

This is empirically not the case. I set a breakpoint in the `foreach` during testing, and the enumeration contained requirements that had not yet been checked and marked successful.

I'll make the change in the docs, but I don't feel good about it.

HaoK:
Requirements gives you all of them, Pending gives you only the set that hasn't been succeeded. For the purposes of docs, its clearer to use Pending

scottaddie:
@rianjs Thank you again for your help with this!

PR #5602, opened 2018-03-06
add ms.topic metadata value

v-savila:
for task 1199273 per Sarah Barrett


v-savila:
#sign-off from apexprod

PR #5629, opened 2018-03-08
Update sample2.cs

sambres:
Missing semicolon

When you copy paste this code without the simcolon, it results as a compilation error.
I just added the semicolon to be able to copy/paste the code without it to cause a compilation error.


scottaddie:
@sambres Very nice catch! Thank you for taking the time to contribute.

PR #5638, opened 2018-03-08
Changed route order

ridercz:
The current route order (0) causes that generating links (ie. using tag helpers) does not work. Proposing fix after consultation with @pranavkm.


scottaddie:
@ridercz Thank you very much for addressing this!

Issue #5649, opened 2018-03-10
Surface these GREAT docs and surface landing page from RP tutorial


guardrex:
Discussed in docs sessions at the Summit: The progression of moving from `dot.net` to the docs makes no mention of how great the docs are. Only the navbar link to **Docs** is present. There is no language to suggest that .NET in general and ASP.NET Core specifically offer better support documentation than other web frameworks.

1. [.NET](https://www.dot.net) `dot.net`/`https://www.microsoft.com/net` :point_right: *Web: web apps and services* link ...
1. [Web and Server Applications](https://www.microsoft.com/net/learn/apps/web) :point_right: *Get started with ASP.NET* button ...
1. [Get Started with ASP.NET](https://www.asp.net/get-started) :point_right: Last step *Razor Pages* link ...
1. [Introduction to Razor Pages in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/mvc/razor-pages/?tabs=visual-studio)

Location 1 (above) might include a section (below *One Platform*):

> ### Documentation and Support
>
> Whether you're new to software development or an experienced professional, .NET is fully supported by extensive and helpful documentation, Channel 9 videos, forums, and chat rooms.

Location 2 (above) might include a brief mention along the lines of:

> .NET is supported by helpful documentation, Channel 9 videos, forums, and chat rooms.

Location 3 (above):

Seems ok given that the *Learn more* link lands on https://www.asp.net/learn, which in turn links to ASP.NET and ASP.NET Core doc sets.

I suppose the idea is to encourage the visitor (following the link chain, a **new to .NET** visitor) to get right into building a RP app. However, I'm not so sure that's the only suggestion the text should make. I mean that dumps someone brand new to .NET, possibly new even to software dev or web dev, right into a rather meaty tutorial. I wonder if the last point in Location 3 should do something more like ...

> ### That’s it!
> 
> Next up, continue building your first ASP.NET Core app with Razor Pages or get a general introduction to ASP.NET Core. (link *ASP.NET Core* to https://docs.microsoft.com/aspnet/core/).

Location 4 (above):

One thing I can do is add a quick remark to the top of the tutorial to make it a bit easier for someone trying to reach an overview to get out to the ASP.NET Core landing page.

guardrex:
@scottaddie @Rick-Anderson I did what I could here. Do you want to advance this along for possible changes to microsoft.com pages? If not, you can close this after #5652 merges.

shirhatti:
[...]

PR #5652, opened 2018-03-11
Surface the intro topic in the RP tutorial

guardrex:
Addresses #5649 

This topic is the landing page for a reader arriving along the chain of links starting with `https://dot.net` (`https://www.microsoft.com/net`). In addition to the current link allowing a reader to get out to the *Get started with RP* page, I suggest we provide an additional circuit breaker that will allow the reader to get out to the doc set's main landing page.


guardrex:
I see a few more "Getting"s ... I'll square them away shortly :point_right: https://github.com/aspnet/Docs/issues/5653

Issue #4655, opened 2017-10-29
Mention the opt-in support for binding props OnGet in RP topic?


guardrex:
In the [RP topic](https://docs.microsoft.com/en-us/aspnet/core/mvc/razor-pages), non-binding props on GET requests is mentioned in passing ...

> Razor Pages, by default, bind properties only with non-GET verbs.

It could add the opt-in property requirement for binding on GET requests:

```csharp
[BindProperty(SupportsGet=true)]
```

... and then mention that there are security considerations for doing so.

The default behavior was confusing for a dev who needed prop binding on GET: https://youtu.be/p7iHB9V-KVU?t=54m27s (Standup 9/5)

Rick-Anderson:
@guardrex  closing. We don't want to encourage bad behavior. Dev's who are smart enough to use this feature correctly will figure out how to do it.

PR #5658, opened 2018-03-12
Add note explaining Razor Pages property binding on GET

scottaddie:
Fixes https://github.com/aspnet/Docs/issues/5657

### DIFF
--- a/aspnetcore/mvc/razor-pages/index.md
+++ b/aspnetcore/mvc/razor-pages/index.md
@@ -149,6 +149,8 @@ The `Customer` property uses `[BindProperty]` attribute to opt in to model bindi
 Razor Pages, by default, bind properties only with non-GET verbs. Binding to properties can reduce the amount of code you have to write. Binding reduces code by using the same property to render form fields (`<input asp-for="Customer.Name" />`) and accept the input.
 
 > [!NOTE]
+> For security reasons, you must opt in to binding GET request data to page model properties. Proceed with caution when trusting user input and mapping it to properties.
+>
 > To bind a property on GET requests, set the `[BindProperty]` attribute's `SupportsGet` property to `true`:
 > `[BindProperty(SupportsGet = true)]`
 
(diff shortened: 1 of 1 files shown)


guardrex:
@Rick-Anderson previously shot this one down. https://github.com/aspnet/Docs/issues/4655 :gun: lol

scottaddie:
@guardrex In conversations with both @rynowak and @isaac2004 on Friday, it was recommended that we add this. In the past, I've also seen customers ask Damian how to do this. There's a need for it. If not in this doc, we should put it somewhere else.

guardrex:
Preach'in to the choir my friend. I opened #4655 because I wanted to do it.

scottaddie:
I can see both sides too, but it's a problem if customers are going to a PM for the answer.

isaac2004:
Or bothering people on the docs team :)

rynowak:
This absolutely needs to be part of the docs for the cases where folks want to need to use it. This is not a bad practice or something we tell users to avoid.

Rick-Anderson:
@rynowak  can we add something like:
Binding on `Get` is generally not required.

It would be nice to provide some background on why Get is excluded by default. Are there case where blindly adding binding on get is problematic? There must be a good reason it was excluded.

For example, you might need to add binding to `Get` when ...

Rick-Anderson:
@scottaddie  play the video: [ `@DamianEdwards explains risk`](https://youtu.be/p7iHB9V-KVU?t=3251) and add those caveats.

scottaddie:
@rynowak @Rick-Anderson Please take another look.

aspnetcore/mvc/razor-pages/index.md
```diff
@@ -148,6 +148,12 @@ The `Customer` property uses `[BindProperty]` attribute to opt in to model bindi
 
 Razor Pages, by default, bind properties only with non-GET verbs. Binding to properties can reduce the amount of code you have to write. Binding reduces code by using the same property to render form fields (`<input asp-for="Customer.Name" />`) and accept the input.
 
+> [!NOTE]
+> For security reasons, you must opt in to binding GET request data to page model properties. Proceed with caution when trusting user input and mapping it to properties.
```
Rick-Anderson: >Proceed with caution

That's not very helpful. How about something along the lines.

Verify user input.

Rick-Anderson: Should you add. You might need to bind GET to support searching - or something like that. Gives motivation.

PR #5663, opened 2018-03-12
Update index.md

Rick-Anderson:
@adamtuliper @MarkMichaelis

### DIFF
--- a/aspnetcore/fundamentals/configuration/index.md
+++ b/aspnetcore/fundamentals/configuration/index.md
@@ -23,8 +23,8 @@ There are configuration providers for:
 * Command-line arguments
 * Environment variables
 * In-memory .NET objects
-* An encrypted user store
-* [Azure Key Vault](xref:security/key-vault-configuration)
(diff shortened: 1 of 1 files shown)


aspnetcore/fundamentals/configuration/index.md
```diff
@@ -23,8 +23,8 @@ There are configuration providers for:
 * Command-line arguments
 * Environment variables
 * In-memory .NET objects
-* An encrypted user store
-* [Azure Key Vault](xref:security/key-vault-configuration)
+* The unencrypted [Secret Manager](xref:security/app-secrets) storage.
```
guardrex: They have that 3-ish word cutoff when they talk about using periods or not. If you use them as you have (and it looks like you should), then all of the list items will get them.

> If one or more list elements are complete sentences, use a period after every element, even if a list element contains three or fewer words.

guardrex:
I just had the one comment about the periods. Otherwise, LGTM.

PR #5664, opened 2018-03-13
SignalR Example: add method to ChatHub broadcast

caleblloyd:
I followed this example today and was getting this error:

screenshot from 2018-03-12 20-59-53

I realized that this example omits the method name `Send`, and it is passing `timestamp` as the method name instead (when that should be the first argument)

Here is the function deceleration, it requires method name:

```
    /// <summary>
    /// Invokes a method on the connection(s) represented by the <see cref="T:Microsoft.AspNetCore.SignalR.IClientProxy" /> instance.
    /// Does not wait for a response from the receiver.
    /// </summary>
    /// <param name="clientProxy">The <see cref="T:Microsoft.AspNetCore.SignalR.IClientProxy" /></param>
    /// <param name="method">name of the method to invoke</param>
    /// <param name="arg1">The first argument</param>
    /// <param name="arg2">The second argument</param>
    /// <returns>A task that represents when the data has been sent to the client.</returns>
    public static Task SendAsync(this IClientProxy clientProxy, string method, object arg1, object arg2)
```

After I added the method name `Send`, the example started working

### DIFF
--- a/aspnetcore/signalr/get-started-signalr-core/sample/Hubs/ChatHub.cs
+++ b/aspnetcore/signalr/get-started-signalr-core/sample/Hubs/ChatHub.cs
@@ -6,10 +6,10 @@ namespace SignalRChat.Hubs
 {
     public class ChatHub : Hub
     {
-        public Task Send(string user, string message)
+        public Task SendMessage(string user, string message)
         {
             string timestamp = DateTime.Now.ToShortTimeString();
-            return Clients.All.SendAsync(timestamp, user, message);
+            return Clients.All.SendAsync("ReceiveMessage", timestamp, user, message);
         }
     }
 }
--- a/aspnetcore/signalr/get-started-signalr-core/sample/wwwroot/js/chat.js
+++ b/aspnetcore/signalr/get-started-signalr-core/sample/wwwroot/js/chat.js
@@ -1,6 +1,6 @@
 ﻿const connection = new signalR.HubConnection('/chathub');
 
-connection.on('Send', (timestamp, user, message) => {
+connection.on('ReceiveMessage', (timestamp, user, message) => {
     const encodedUser = user;
     const encodedMsg = message;
     const listItem = document.createElement('li');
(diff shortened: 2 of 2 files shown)


aspnetcore/signalr/get-started-signalr-core/sample/Hubs/ChatHub.cs
```diff
@@ -9,7 +9,7 @@ public class ChatHub : Hub
         public Task Send(string user, string message)
         {
             string timestamp = DateTime.Now.ToShortTimeString();
-            return Clients.All.SendAsync(timestamp, user, message);
+            return Clients.All.SendAsync("Send", timestamp, user, message);
```
scottaddie: We should change this string to `nameof(Send)`

caleblloyd:
@scottaddie made the change to `nameof(Send)` and tested it out locally, works

aspnetcore/signalr/get-started-signalr-core/sample/Hubs/ChatHub.cs
BrennanConroy: `nameof(Send)` implies that the server function is the same as the client function. It should be `"Send"`

scottaddie: In that case, would it be a better idea to use a client function name other than "Send" in this example? This would prove that the names don't have to match.

aspnetcore/signalr/get-started-signalr-core/sample/Hubs/ChatHub.cs
```diff
@@ -9,7 +9,7 @@ public class ChatHub : Hub
         public Task Send(string user, string message)
         {
             string timestamp = DateTime.Now.ToShortTimeString();
-            return Clients.All.SendAsync(timestamp, user, message);
+            return Clients.All.SendAsync(nameof(Send), timestamp, user, message);
```
anurse: This should not be `nameof`. The name expected here is the name of the **client** method, which is not guaranteed to be the same as the server method. We shouldn't set the assumption that they will be the same. This is why I keep advocating for these samples to use "Receive" as the name on the client but @davidfowl doesn't like that for some reason :)

scottaddie: I second the idea of using a different name. As you can see, I was under the impression the names were to be the same.

caleblloyd:
@anurse @scottaddie how about this:

Method invoked on client is called `ReceiveMessge`
Method called on server is called `SendMessage`

I've made this change

rachelappel:
Thank you very much @caleblloyd, your PR has been merged!

PR #5666, opened 2018-03-13
Remove LogoutPath from cookie options in Identity topic

### LINKED ISSUE #5615: Add note on LogOff behavior to Intro to Ident sample
See: https://github.com/aspnet/templating/issues/88

guardrex:
Fixes #5615 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/security/authentication/identity?branch=pr-en-us-5666)

The engineers decided not to change the template ([2.1 ref source](https://github.com/aspnet/templating/blob/dev/src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Controllers/AccountController.cs#L104) and [comment \#5615](https://github.com/aspnet/templating/issues/88#issuecomment-370969013)). In effect, the template is set up so that the `LogOff` method performs a hard `RedirectToAction` to the Home controller `Index` method. Therefore, setting the `LogoutPath` option no-ops, which might be unexpected behavior given that the doc is showing how to set it.

I added the note on this to the code, and I think it should stay with the sample (and future samples for this topic). When a dev pulls down the code, they'll always get the note. I think if we have this information in the topic, it will be lost by the time the dev gets into configuring the sample app.

### DIFF
--- a/aspnetcore/security/authentication/identity.md
+++ b/aspnetcore/security/authentication/identity.md
@@ -56,7 +56,7 @@ In this topic, you'll learn how to use ASP.NET Core Identity to add functionalit
 
     # [ASP.NET Core 2.x](#tab/aspnetcore2x)
     
-    [!code-csharp[](identity/sample/src/ASPNETv2-IdentityDemo/Startup.cs?name=snippet_configureservices&highlight=7-9,11-28,30-39)]
+    [!code-csharp[](identity/sample/src/ASPNETv2-IdentityDemo/Startup.cs?name=snippet_configureservices&highlight=7-9,11-28,30-42)]
     
     These services are made available to the application through [dependency injection](xref:fundamentals/dependency-injection).
     
@@ -66,7 +66,7 @@ In this topic, you'll learn how to use ASP.NET Core Identity to add functionalit
     
     # [ASP.NET Core 1.x](#tab/aspnetcore1x)
     
-    [!code-csharp[](identity/sample/src/ASPNET-IdentityDemo/Startup.cs?name=snippet_configureservices&highlight=7-9,13-34)]
+    [!code-csharp[](identity/sample/src/ASPNET-IdentityDemo/Startup.cs?name=snippet_configureservices&highlight=7-9,13-33)]
     
     These services are made available to the application through [dependency injection](xref:fundamentals/dependency-injection).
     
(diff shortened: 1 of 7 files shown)


guardrex:
@HaoK I still have a concern about doing this.

Isn't the bottom line here that `LogoutPath` set in the cookie options really just isn't used any longer in a modern MVC/RP app?

The `LogoutPath` was originally intended to be used with a different logoff flow, where the user is sent to the path provided in `LogoutPath` with a `returnUrl`, then the controller at the `LogoutPath` would sign them out, and then the app would honor the `returnUrl` that was provided. 

Nowadays, it seems like the login partial sends them directly to the `Logoff` method, they get signed out, then they are redirected directly a location without ever sweating the `returnUrl`. I mean if the `returnUrl` has kind'a dropped out of the picture in the modern logoff flow, then I kind'a wonder why we're showing setting it at all in the cookie options.

Has my reasoning gone bad somewhere? :smile: lol

HaoK:
No you are right, we don't really use return url anymore, so you can just remove the docs talking about it if you want...its pretty confusing and esoteric anyways :)

guardrex:
> just remove ... its pretty confusing and esoteric anyways

:+1: Yes, indeed. Please let's do that. *I swear I'm the poster child for Azure B2C!* :smile: lol

Regarding the other ones:

* options.LoginPath
* options.AccessDeniedPath

The behavior for login also looks hard-wired these days: You try to access a page that you aren't authenticated to access. You're redirected to the login page (and the `returnUrl` is added). After login ...

```csharp
public async Task<IActionResult> OnPostAsync(string returnUrl = null)
...
return LocalRedirect(Url.GetLocalUrl(returnUrl));
```
... it sends you back to the page you tried to access. However, is the `options.LoginPath` doesn't look like it's playing a role here either. The `returnUrl` wouldn't be added by this path setting, and the `LocalRedirect` is what's kicking the user back to where they came from.

Can we remove the line for `options.LoginPath` as well?

Lastly, we have `options.AccessDeniedPath`.

This endpoint is completely disconnected from the app in the template. Nothing seems to direct the user to this endpoint. What do we do (if anything) about this one?

HaoK:
Authorization controls the AccessDenied behavior, so if there's an AuthZ policy that fails but they still have a cookie, access denied path should be where you end up.


I'm fine removing the login path stuff too that stuff is super confusing anyways, so if we aren't using it/relying on it anymore, maybe we can get rid of it in 3.0

guardrex:
Cool. Thanks for your help. I'm going to double-check all behaviors and **especially keep an :eye: on status code changes** during redirects to confirm behaviors are as expected. For example ...

> The LoginPath property informs the middleware that it should change an outgoing 401 Unauthorized status code into a 302 redirection onto the given login path. 

I just want to make sure that not only the flow is correct but that the status codes are correct when these redirects occur.

If it turns out that these paths must be correctly set in order to control status codes properly, then the topic should surface that information. Thus far, all I've looked at is the flow ... time for me to assess the status codes.

I'll let you know how it turns out. Thanks again! :smile:

HaoK:
Sounds good, yeah some of those comments might be old now, I don't believe the auth middleware look at status codes at all anymore, they should only be rely on the appropriate verbs being called to do things (challenge/forbid)

guardrex:
@HaoK Good news ...

* `options.LoginPath`
* `options.AccessDeniedPath`

... work as advertised ... they're fine.

It's `options.LogoutPath` that no-ops. Setting it to a non-sense value ...

```csharp
options.LogoutPath = "/Rat/Burgers";
```

:rat: 🍔

... has no effect on the behavior or the status code ...

capture

I did note a number of changes for 2.1. When the Identity topic is addressed for 2.1, it would be wise to take another look at the behavior and make sure it continues to no-op.

For now, I think we can just remove the line from the samples here, and I've updated the PR to reflect that approach. This is only for the Identity topic and our Identity template. All of the `Logout` methods, regardless of template configuration (`OrganizationalAuth`, `IndividualB2CAuth`, `IndividualAuth`) use a hard `return RedirectToAction(nameof(HomeController.Index), "Home");` in 2.0.

HaoK:
Yeah logout path is the outlier as it was only used to control which url were allowed to do redirects I think

PR #5669, opened 2018-03-13
Enhance ClaimsPrincipal/Identity content

### LINKED ISSUE #1190: Enhance ClaimsPrincipal/Identity content in CookieAuth w/o Identity topic
In https://docs.asp.net/en/latest/security/authentication/cookie.html#creating-an-identity-cookie, nothing describes how to create the necessary ClaimsPrincipal. There are a few tricks to deal with there and some docs would help. For example:
- Must set `AuthenticationType` on the ClaimsIdentity
- What claims should I add? Why?

/cc @blowdart @Tratcher

guardrex:
Fixes #1190 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/security/authentication/cookie?branch=pr-en-us-5669)

Mostly addressed by the significant updates on https://github.com/aspnet/Docs/pull/5307.

* I add a role claim to the sample code.
* I add two related topics.

As for ...

> What claims should I add? Why?

Shouldn't that be left to [Claims-Based Authorization](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/claims)? I link it now in this topic.

### DIFF
--- a/aspnetcore/security/authentication/cookie.md
+++ b/aspnetcore/security/authentication/cookie.md
@@ -421,3 +421,5 @@ await HttpContext.Authentication.SignInAsync(
 
 * [Auth 2.0 Changes / Migration Announcement](https://github.com/aspnet/Announcements/issues/262)
 * [Limiting identity by scheme](xref:security/authorization/limitingidentitybyscheme)
+* [Claims-Based Authorization](xref:security/authorization/claims)
+* [Policy-based role checks](xref:security/authorization/roles#policy-based-role-checks)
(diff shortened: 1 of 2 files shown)


anurse:
Not sure I have the context on why this change is needed. It looks fine to me, but I think someone like @Tratcher should take a look.

guardrex:
@anurse In your [original issue \#1190](https://github.com/aspnet/Docs/issues/1190#issue-153299751), you asked about showing how to set a role claim, so I added that in along with the two name claims. Since the ClaimsPrincipal/Identity was added on the earlier PR (I think that covers your other ask) and a reader might want to know more, I added the two links down to the additional resources.

anurse:
Thanks for the context @guardrex ! LGTM

Issue #5682, opened 2018-03-13
Create Include announcing SignalR Core vs SignalR legacy content


rachelappel:
Some kind of message stating which type of content the doc is.

PR #5692, opened 2018-03-14
Include with notice about where versioned docs are for SignalR

rachelappel:
Fixes https://github.com/aspnet/Docs/issues/5682

### DIFF
--- a/aspnetcore/includes/signalr-version-notice.md
+++ b/aspnetcore/includes/signalr-version-notice.md
@@ -1,2 +1,2 @@
 > [!NOTE]
-> This document details the latest version of ASP.NET Core SignalR. Looking to get started with an earlier version of SignalR? See the [SignalR 1.x documentation](xref:aspnet/signalr/).
+> This document details the latest version of ASP.NET Core SignalR. See the [SignalR 1.x documentation](../../aspnet/signalr/) for the ASP.NET Framework version.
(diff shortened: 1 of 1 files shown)


rachelappel:
@Rick-Anderson  This is just the include. I'll open a PR to actually include it in files once this is merged. That way it will avoid merge issues.

aspnetcore/includes/signalr-version-notice.md
```diff
@@ -0,0 +1,2 @@
+> [!NOTE]
+> This document details the latest version of ASP.NET Core SignalR. Looking to get started with an earlier version of SignalR? See the [SignalR 1.x documentation](xref:aspnet/signalr/).
```
Rick-Anderson: The PU prefers the more ref style.

This document details the latest version of ASP.NET Core SignalR. ~Looking to get started with an earlier version of SignalR?~ See the [SignalR 1.x documentation](xref:aspnet/signalr/) for the ASP.NET Framework version.

rachelappel: Ok. I took it from here https://docs.microsoft.com/en-us/aspnet/core/getting-started
I'll add an issue to change it there too then.

Rick-Anderson:
I reopened #5682 until you add the include. Make sure it goes in both topics.

aspnetcore/includes/signalr-version-notice.md
```diff
@@ -0,0 +1,2 @@
+> [!NOTE]
+> This document details the latest version of ASP.NET Core SignalR. See the [SignalR 1.x documentation](../../aspnet/signalr/) for the ASP.NET Framework version.
```
Rick-Anderson: Don't use relative links. I think the (aspnet/aspnet/signalr) is what you want. @guardrex  is that right?

guardrex: Yes ... that will land here :point_right: https://docs.microsoft.com/en-us/aspnet/signalr/

PR #5696, opened 2018-03-15
replace Post with Put

pkellner:
Since post is an insert, it does not make sense to pass in an "id" to a post, but it does to a put.

  Fixes #Issue_Number


scottaddie:
@pkellner Nice catch! Thank you for fixing this.

PR #5710, opened 2018-03-15
Hosting bundle installer + SDK/runtime links update

### LINKED ISSUE #5688: Hosting bundle installer link is pulling down an old installer
@joeloff I thought the new link we're using for the hosting bundle installer would auto roll-forward to the latest hosting bundle installer. However, it's still pulling down the one with 2.0.5 (I assume there's a new one because the 2.0.6 runtime is out).

https://aka.ms/dotnetcore-2-windowshosting

... downloads `DotNetCore.2.0.5-WindowsHosting.exe`.

Also, I see they revamped the .NET Downloads page and the hosting bundle installer isn't available there any longer. https://www.microsoft.com/net/download/all Do you know if it's available anywhere outside of the docs now?

cc/ @shirhatti

guardrex:
Fixes #5688 

There are some general SDK/runtime link updates here, but the main thrust is to solve our hosting bundle installer link problems **permanently**. The recurring pattern on the hosting bundle link updates is ...

1. The new hosting bundle installer is released.
1. We don't know, so we don't update the links in the docs to coincide with the release.
1. We get up to a dozen devs contacting us (me! :smile: lol) saying that the links have gone bad.
1. Then, we have trouble figuring out what to do to harden the links so that they don't break in the future (e.g., going with `https://aka.ms/dotnetcore-2-windowshosting` on the last attempt). These attempts to harden the links have failed.

Let's try another approach: Let's provide instructions to get to the latest stable link, which is the one that they'll put up and maintain. We just need to get the reader there. See if the instruction I added on the PR makes sense. If this works (i.e., they don't change the titles/locations of pages, page headings, or installer link text), we won't need to deal with this again. 😅

See the `aspnetcore/host-and-deploy/iis/index.md` diff, and here's the [Internal Review Topic for Hosting on IIS](https://review.docs.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/index?branch=pr-en-us-5710#install-the-net-core-windows-server-hosting-bundle).


guardrex:
A consensus might be forming! :smile:

guardrex:
The 'ayes' have it 5-0-1 ... the motion is carried.

Next measure ... Docs Bill #145245 ... :beer: *Free Beer for Docs Authors* :beer: ... All in favor signify by *smiling*. The smiles have it ... the motion is carried!

Thanks everyone for looking. I think this will cut us a little slack on missing the hosting bundle installer releases.

PR #5715, opened 2018-03-15
Update blazor-preview-notice.md

Rick-Anderson:


guardrex:
@Rick-Anderson What if we call out the Blazor repo, too? See my "Update" commit.

guardrex:
@Rick-Anderson ... and probably better if we break that out. See: "Break it out"

PR #5722, opened 2018-03-17
Add link to App Service troubleshooting topic

guardrex:
Per request by @Rick-Anderson.


guardrex:
@Rick-Anderson Did it again on this PR ...

capture

... and the error is similar to the first time. It claims a bad link ...


Warning | Unspecified | aspnetcore/SignalR/index.md | Invalid file link:(~/SignalR/get-started-signalr-core.md). | 15 | 03/17/2018 04:53:14
-- | -- | -- | -- | -- | --

... but that isn't correct for `master` at this time. It seems like the build system is using an older doc set, where perhaps that link was broken and it's no longer the case.

Issue #5238, opened 2018-01-22
Enhance cookie sharing topic for common Identity dB


guardrex:
~Blocked by: https://github.com/aspnet/AuthSamples/pull/45~

aspnetcore/security/data-protection/compatibility/cookie-sharing.md
https://docs.microsoft.com/aspnet/core/security/data-protection/compatibility/cookie-sharing

The three cookie-sharing projects in the cookie sharing topic sample are under revision in the AuthSamples repo. ~The refactor aims to have all three projects use a common Identity dB. For the cookie sharing topic, the dev will only need to create the user in one project to sign into all three apps. For an alternative topic, the projects will be an example of sharing an Identity dB generally across apps (Core-to-Core and Core-to-NET).~

* ~Drop the current sample from the topic.~
* ~Cross-repo link the AuthSamples sample into the topic (See the *External code snippets* section of *Step 3. Authoring* > *Authoring Content* > *Code snippets* of the OPS docs).~
* ~Update the instructions/content in the topic (and README, if needed) to reflect a single-dB sample.~

We're going to comment in the text (and possibly even update the sample) on sharing a common Identity dB across the apps.

Tratcher:
A consistent issue I see people copying from the current samples is that they're always creating local data protection providers rather than configuring the central service.
https://stackoverflow.com/a/49329480/2588374
This sample would be better:
https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/configuration/overview?tabs=aspnetcore2x#persistkeystofilesystem

guardrex:
It's documented in the topic and in the sample, but it's a quick one-liner in the topic. I'll add a paragraph to highlight it.

guardrex:
@HaoK I've worked up to this issue now.

We were discussing the possibility of adding a section that describes the use of a common Identity dB across the cookie-sharing apps. What did you have in mind for this section? Alternatively, we could let this issue go because this isn't a developer ask.

HaoK:
Yeah we can just let it go for now if noone has asked about it to date.

PR #5723, opened 2018-03-17
Update cookie sharing topic

guardrex:
Addresses #5238 

Based on @Tratcher remark at https://github.com/aspnet/Docs/issues/5238#issuecomment-373896559:

> A consistent issue I see people copying from the current samples is that they're always creating local data protection providers rather than configuring the central service.
https://stackoverflow.com/a/49329480/2588374
> This sample would be better:
https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/configuration/overview?tabs=aspnetcore2x#persistkeystofilesystem

This PR takes the existing one-liner in the topic, which is easy to overlook when reading the topic, and fleshes it out with an added paragraph that also surfaces `PersistKeysToFileSystem`.


aspnetcore/security/cookie-sharing.md
```diff
@@ -45,6 +45,8 @@ In the `ConfigureServices` method, use the [ConfigureApplicationCookie](/dotnet/
 
 [!code-csharp[](cookie-sharing/sample/CookieAuthWithIdentity.Core/Startup.cs?name=snippet1)]
 
+Data protection keys must be shared among apps. In the sample apps, `GetKeyRingFolderPath` returns the common key storage location. In ASP.NET Core apps, an alternative approach to supplying the key location to [DataProtectionProvider.Create(DirectoryInfo)](/dotnet/api/microsoft.aspnetcore.dataprotection.dataprotectionprovider.create#Microsoft_AspNetCore_DataProtection_DataProtectionProvider_Create_System_IO_DirectoryInfo_) is to configure the data protection service with the [PersistKeysToFileSystem](/dotnet/api/microsoft.aspnetcore.dataprotection.dataprotectionbuilderextensions.persistkeystofilesystem) method. For more information, see [Configuring Data Protection: PersistKeysToFileSystem](xref:security/data-protection/configuration/overview#persistkeystofilesystem).
```
Tratcher: Please directly include the sample code:
```
{
    services.AddDataProtection()
        .PersistKeysToFileSystem(new DirectoryInfo(@"\\server\share\directory\"));
}
```

Also, let's move this above  ConfigureApplicationCookie and make it the primary recommendation. In core setting DP directly on ConfigureApplicationCookie is the alternative.

guardrex:
@Tratcher Need a tip now ... this change broke the cookie sharing. Logging into one app or the other logs the user *out* of the other app.

guardrex:
`options.DataProtectionProvider` is null unless it's created by `DataProtectionProvider.Create` or supplied via DI.

If I use `DataProtectionProvider.Create`, then it takes the `DirectoryInfo` of the key storage location, so we're back to square one ... no point in using `PersistKeysToFileSystem`.

If I try to DI it into `Startup`, `public Startup(IDataProtectionProvider dataProtection)`, it claims that it can't resolve it, so I don't get a chance to use `options.DataProtectionProvider = _dp;`.

U sure about this? @HaoK reviewed these apps, and he didn't say anything at the time.

Tratcher:
You don't need to DI into Startup, CookieAuth takes care of that itself:
https://github.com/aspnet/Security/blob/d24d4fa99f46822a3b9b87702507e0dc914a0126/src/Microsoft.AspNetCore.Authentication.Cookies/PostConfigureCookieAuthenticationOptions.cs#L17

You only need
```
 services.AddDataProtection()
        .PersistKeysToFileSystem(new DirectoryInfo(@"\\server\share\directory\"));
```

guardrex:
I did, and it didn't work. Logging into one app logs the user out of the other app.

Using the original samples (without `PersistKeysToFileSystem` and with `DataProtectionProvider.Create`), everything works.

I'll have to come back to this later. I'm working on :zap: **_Blazor!_** :zap: stuff this weekend 😀. I'll take another look at this on Monday.

HaoK:
Yeah this seems better, basically configure the central data protection to use shared keys instead of individually configuring instances....  Once you get it working of course :)

guardrex:
> Once you get it working of course

No 🎲🎲.

Look at the samples (the two Core ones) and tell me what to change.

My first test that didn't work was between the two Core apps. Logging into one logs the user out of the other. Latest change was to use `Identity.Application` in the `CookieAuthWithIdentity.Core` sample. It didn't work with that change, and it didn't work when removing `options.DataProtectionProvider` completely.

guardrex:
I'll try `DataProtectionProvider.Create()` with the `applicationName` overload and see what that does.

guardrex:
First two are working now. I'll see if I can get the netfx one to play nicely with the other two now.

guardrex:
Arrrrrrgh! That netfx project is in :fire: **_NUGET HELL!_** :fire: I'll have to come back to this later.

Good news is that the two Core projects are back to sharing cookies with `PersistKeysToFileSystem`.

guardrex:
I just setup a new netfx app to clear my woes.

I don't see (yet) how to set the app name, which must be the same as the other apps in order to DP to permit use of the same keys.

```csharp
app.UseCookieAuthentication(new CookieAuthenticationOptions
{
    AuthenticationType = "Identity.Application",
    CookieName = ".AspNet.SharedCookie",
    LoginPath = new PathString("/Account/Login"),
    Provider = new CookieAuthenticationProvider
    {
        OnValidateIdentity =
            SecurityStampValidator
                .OnValidateIdentity<ApplicationUserManager, ApplicationUser>(
                    validateInterval: TimeSpan.FromMinutes(30),
                    regenerateIdentity: (manager, user) =>
                        user.GenerateUserIdentityAsync(manager))
    },
    TicketDataFormat = new AspNetTicketDataFormat(
        new DataProtectorShim(
            DataProtectionProvider.Create(GetKeyRingDirInfo())
            .CreateProtector(
                "Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationMiddleware",
                "Identity.Application",
                "v2"))),
    CookieManager = new ChunkingCookieManager()
});
```
The overloads for `DataProtectionProvider.Create` take either an app name --OR-- a key storage location but not both.

guardrex:
Rubber :duck: says ...
```csharp
DataProtectionProvider.Create(GetKeyRingDirInfo(), 
    (builder) => { builder.SetApplicationName("SharedCookieApp"); })
```

guardrex:
kk ... try it now ...

https://review.docs.microsoft.com/en-us/aspnet/core/security/cookie-sharing?branch=pr-en-us-5723

Rick-Anderson:
@Tratcher  or @HaoK can one of you take a quick look at @guardrex  last update so I can merge?

aspnetcore/security/cookie-sharing.md
```diff
@@ -83,6 +105,20 @@ When using cookies directly:
 
 [!code-csharp[](cookie-sharing/sample/CookieAuth.Core/Startup.cs?name=snippet1)]
 
+Data protection keys and the app name must be shared among apps. In the sample apps, `GetKeyRingDirInfo` returns the common key storage location to the [PersistKeysToFileSystem](/dotnet/api/microsoft.aspnetcore.dataprotection.dataprotectionbuilderextensions.persistkeystofilesystem) method. Use [SetApplicationName](/dotnet/api/microsoft.aspnetcore.dataprotection.dataprotectionbuilderextensions.setapplicationname) to configure a common shared app name (`SharedCookieApp` in the sample). For more information, see [Configuring Data Protection](xref:security/data-protection/configuration/overview). 
+
+In ASP.NET Core apps, an alternative approach is not to use `PersistKeysToFileSystem` but to supply the key location to [DataProtectionProvider.Create(DirectoryInfo)](/dotnet/api/microsoft.aspnetcore.dataprotection.dataprotectionprovider.create#Microsoft_AspNetCore_DataProtection_DataProtectionProvider_Create_System_IO_DirectoryInfo_):
+
+```csharp
+services.AddAuthentication(CookieAuthenticationDefaults.AuthenticationScheme)
```
Tratcher: You're using inconsistent auth schemes between samples. Use "Identity.Application" everywhere and put a note at the top that then need to match so you selected "Identity.Application" because it's the value used by ConfigureApplicationCookie.

Tratcher: Do we need to mirror this example for ConfigureExternalCookie?

guardrex:
thanks @Tratcher 

I'm going to make one more tiny change. In the existing note about "Identity.Application" (at the top where it talks about what the examples show), I'd like to add a line that explains how it must match across the app. I'll have it in a few minutes.

guardrex:
@Tratcher How about from this ...

> * `Identity.Application` is used as the authentication scheme.

... to this ...

> * `Identity.Application` is used as the authentication scheme. Whatever scheme is used, it must be used consistently *within each app* either as the default scheme or by explicitly setting it. The scheme doesn't need to be consistent across the shared cookie apps.

Tratcher:
No, it's worse than that. The scheme is used as one of the seeds into data protection so it needs to be consistent across applications.

guardrex:
Nice job Rex! ... No wonder all of my credit cards numbers are freely available on the open Internet!!  :weary:

guardrex:
How about this one?

> * `Identity.Application` is used as the authentication scheme. Whatever scheme is used, it must be used consistently *within and across* the shared cookie apps either as the default scheme or by explicitly setting it. The scheme is used to secure data in the data protection stack. Because the apps must share data protection keys in order to decrypt the authentication ticket within a shared cookie, a common scheme must be used across the apps.

Didn't want to use the word "seed" in it. I feel like I would need to explain more if I did, so I went with "used to secure data" instead.

Tratcher:
> The scheme is used to secure data in the data protection stack. Because the apps must share data protection keys in order to decrypt the authentication ticket within a shared cookie, a common scheme must be used across the apps.

how about
> The scheme is used as one of the inputs when encrypting and decrypting cookies so a consistent value must be used across applications.

guardrex:
mmmm ... I'm not too hot about *used as one of the inputs*.

Let's try this ...

> The scheme is used when encrypting and decrypting cookies, so a consistent scheme must be used across apps.

There was merit tho to explaining it all out the way that I did. It tells the reader more explicitly how the cookie sharing is tied to the DP stack. However, I like short, clear explanations, too. Let's go with the short one. If a reader ever wants more, I think they'll give us that feedback.

Tratcher:
:shipit:

Issue #5735, opened 2018-03-18
Link updates


guardrex:
Coming up on that time again. I have some good software that will find the offensive links (301/302/401/403/404).

However, I might whip up a 😵 *mad Rex script* 😵 that will update 301/302 links ✨ auto-magically :sparkles:.

... actually ... I think I have that somewhere on the hard drive! :smile:

Rick-Anderson:
@tdykstra 	@scottaddie 	what's the right tools to look for bad links?

guardrex:
Not that you asked *me*, but I recommend http://home.snafu.de/tilman/xenulink.html ... I'm I 😵? ... but it works *InSaNeLy* well. Also note that the problem with script-kitty approaches on the repo is that because the content validator thing throws errors instead of warnings, we can't script massive sets of changes on PRs now. *Just say'in!* ... but perhaps you (admins) can merge anyway. That's an avenue because at least in the 301 case, the links can be updated via a script a heck of a lot faster than by a *lowly human*.

tdykstra:
> tools to look for bad links?

CATS -- https://contentqacats.azurewebsites.net/

Rick-Anderson:
We should probably run CATS a couple every 6 months - at least.
Last CATS output had about 50% false positives - not too bad.

PR #5736, opened 2018-03-18
Get CDN content links on HTTPS protocol

### LINKED ISSUE #5734: Get CDN content links on HTTPS protocol
* `http://ajax.aspnetcdn.com` :point_right: `https://ajax.aspnetcdn.com`
* `http://ajax.microsoft.com` :point_right: `https://ajax.aspnetcdn.com`

guardrex:
Fixes #5734

* `http://ajax.aspnetcdn.com` :point_right: `https://ajax.aspnetcdn.com`
* `http://ajax.microsoft.com` :point_right: `https://ajax.aspnetcdn.com`

Changes made with find-and-replace, so there's no need to visually inspect every file.

We'll need to eventually do a link check again. I opened #5735 for that, and I'll get to that after 2.0 and Blazor are tended to.


damienbod:
It would also make sense to add the integrity parameter to the script tag. If you want I could do this

```
<script src="https://ajax.aspnetcdn.com/ajax/jquery/jquery-1.9.0.js"></script>
```

to 

```
<script src="https://ajax.aspnetcdn.com/ajax/jquery/jquery-2.2.0.min.js"
 asp-fallback-src="~/lib/jquery/dist/jquery.min.js"
 asp-fallback-test="window.jQuery" 
 crossorigin="anonymous"
 integrity="sha384-K+ctZQ+LL8q6tP7I94W+qzQsfRV2a+AfHIi9k8z8l9ggpc8X+Ytst4yBo/hH+8Fk">
</script>
```

guardrex:
@damienbod Perhaps for crossorigin and integrity but not for the fallback. We'll need some discussion (e.g., algorithm strength, prob sha384), and we should highlight the possible pain point here: If the SRI hashes aren't known for the CDN, they should be produced/checked from the source files on that CDN (i.e., I don't trust them being copied over from the issuer or some other CDN without checking). If you're interested in taking it on, please open an issue, and let's talk about it.

PR #5745, opened 2018-03-20
Title, description, and H1 updates (4)

### LINKED ISSUE #3278: Add technology names to doc titles and h1's to improve discoverability in search
Task: 
Improve search
Add ASP.NET technology names to doc titles and H1's where feasible across the entire doc set. Both ASP.NET Core and ASP.NET.

Background:
This seems trivial but it makes a difference. Where we have corrected this so far we see around a 40% increase in clicks from search with indications that the community is getting to where they need to be rather than having to visit pages and read details to determine if the topic is applicable to their need.

Here is an example: https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/routing
Page title (metadata): You can do the following either across a content set via a file in the repo, or for each article individually as it appears was done so far.
`Change: title: Routing to Controller Actions`
`To:         title: Routing to Controller Actions - ASP.NET Core MVC`
  [Note hyphen in place of text: You don’t need to add a complete phrase.]
H1:
`Change: # Routing to Controller Actions in ASP.NET Core MVC`
`To:         # Routing Controller Actions in ASP.NET Core MVC`

Verify the rich snippet results in search engines on a few as you go to make sure we are getting the results.

### LINKED ISSUE #5280: Tracking: Metadata, title, description, and H1 updates
Tracking:

* [x] Metadata entry sequence (https://github.com/aspnet/Docs/pull/5296)
* [ ]  Too long/short titles/descriptions/H1s, missing descriptions, and technology names in titles/H1s (#3278)

Metadata sequence has become a bit messy over time. I'd like to run one of my 😵 mad scripts 😵 on the repo to get a consistent order on the entries.

Sequence:

1. Title
2. Everything else alphabetized
3. UID

Descriptions should be between 115-145 characters including spaces. I'd like to run another script to check for and repair too long/short descriptions (and provide one when missing).

### LINKED ISSUE #5534: Add "ASP.NET Core" to titles and H1's
For example, [Model Binding](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/model-binding) should be "Model binding in ASP.NET Core".

Use SkyEye to start at the top, but update all articles.

guardrex:
Fixes #5280 
Fixes #5534 
Fixes #3278

Last 72 to match H1s with titles and a few last fixups

🎈 *This concludes our metadata updates!* :tada:


guardrex:
These aren't title changes that introduce gerunds ... they're H1 changes to match existing titles. The gerunds were already present in the titles.

I agree: I'll do another pass for this PR and get the gerunds out.

guardrex:
Good one @scottaddie ... they were so unnecessary in most topic titles.

Left a few tho ... things like "logging" and "databinding" are proper names.

I would like to ask our :guitar: *localization rockstar* :guitar:, @Sara-Nicolini, about another set of changes we probably should make. Sara, should we also update the topic filenames (which are part of the URLs)?

For example ...

`aspnetcore/client-side/using-grunt.md` :point_right: `aspnetcore/client-side/use-grunt.md`

... thus ...

`https://docs.microsoft.com/aspnet/core/client-side/using-grunt`

👇 

`https://docs.microsoft.com/aspnet/core/client-side/use-grunt`

... and if 'yes,' do we need to send your team a list of those filename changes *before* we merge them?

aspnetcore/includes/webApi/next.md
```diff
@@ -1,7 +1,7 @@
 ## Next steps
 
 * [ASP.NET Core Web API Help Pages using Swagger](xref:tutorials/web-api-help-pages-using-swagger)
-* [Routing to Controller Actions](xref:mvc/controllers/routing)
+* [Routing to controller actions](xref:mvc/controllers/routing)
```
Rick-Anderson: Reverted this. I think this is an exception.

guardrex: Very good ... thus far, it looks like the exceptions are:

* logging
* databinding
* routing
* hosting
* rewriting
* caching
* scripting
* setting
* using (but "use" works in most cases, so we prob should use "use" where possible)

I added this list to my old skool :cow2: beef :cow2: issue for easy future reference. https://github.com/aspnet/Docs/issues/2617

PR #5749, opened 2018-03-20
Update highlight on dependency-injection.md

Jaxelr:
Currently the highlight is mapping the operation of GetAveragePriority on the incorrect line:

image

The correction is to highlight the previous line.

*** Files no issue since its a minor correction, apologies.


scottaddie:
@Jaxelr Thank you for taking the time to fix this! Your change will go live later today.

PR #5755, opened 2018-03-20
Update TodoController2.cs

### LINKED ISSUE #5711: Source code for Visual Studio Mac and EF Example Confusing
## General
The example source code for this article:

https://docs.microsoft.com/en-us/aspnet/core/tutorials/first-mvc-app-mac/start-mvc

has a commented out controller2 method that looks like it is used to test adding an record to the inmemory database.  Then, you can see the data using the controller not commented out.

I doubt someone new to asp.net will figure that out. I think it would be better to have the test for 0 records in the same controller that displays that data. You've already got the context there so it would just add one line (like initData()) to populate the data if it does not exist already.
IMHO of course.

Rick-Anderson:
Fixes #5711


Rick-Anderson:
cc @pkellner

PR #5760, opened 2018-03-21
includes/signalr-version Include to drive traffic from SignalR to SignalR Core

rachelappel:
fixes https://github.com/aspnet/Docs/issues/5700


rachelappel:
For some reason this sucked in a bunch of commits that don't belong there. Closing.

Issue #5700, opened 2018-03-15
Create include to drive traffic from SignalR to SignalR Core


Rick-Anderson:
See https://docs.microsoft.com/en-us/aspnet/mvc/overview/getting-started/introduction/index as an example.

Obviously mention the core version is in preview.

rachelappel:
The whole doc? And isn't it redundant if there is already an include stating where the new docs are? This issue needs to be more clear.

Rick-Anderson:
https://docs.microsoft.com/en-us/aspnet/signalr/overview/getting-started/introduction-to-signalr needs something.

Probably remove **Questions and comments**

PR #5761, opened 2018-03-21
includes/signalr-updated-version Create include to drive traffic from SignalR to SignalR Core #5700

rachelappel:
fixes https://github.com/aspnet/Docs/issues/5700

### DIFF
--- a/aspnet/includes/signalr-updated-version.md
+++ b/aspnet/includes/signalr-updated-version.md
@@ -1,0 +1,1 @@
+The [latest version of SignalR documentation](https://docs.microsoft.com/aspnet/core/signalr/introduction) for .NET Core is now available.
--- a/aspnet/includes/signalr-updated-version.md
+++ b/aspnet/includes/signalr-updated-version.md
@@ -1,1 +1,1 @@
-The [latest version of SignalR documentation](https://docs.microsoft.com/aspnet/core/signalr/introduction) for .NET Core is now available.
+`The preview version of SignalR for ASP.NET Core documentation is [here](https://docs.microsoft.com/aspnet/core/signalr/introduction).`


rachelappel:
The existing sample (below) violates our SEO guidelines by using [Here] as a link. Additionally it should not have to be stated that there are improvements in our later versions of products.  So this notice takes the simple road: "The latest version of SignalR documentation for .NET Core is now available." as that is all that is necessary. 

 "An updated version of this tutorial is available [here](https://docs.microsoft.com/aspnet/core/tutorials/first-mvc-app/start-mvc) using the latest version of [Visual Studio.](https://www.visualstudio.com) The new tutorial uses [ASP.NET Core MVC](https://docs.microsoft.com/aspnet/core/mvc/), which provides **many** improvements over this tutorial."

Rick-Anderson:
@rachelappel  SEO is not an issue here. There is plenty of SignalR in this content for SEO. In general not using here in a link is good advice. Repeating nearly the same title is clumsy and doesn't improve SEO.

What I don't like is there is no suggestion to go to the new content. The link is secondary, there is no action to go.

The latest version is not not now available, it's in preview.

@danroth27  how do you want to encourage folks to consider SignalR Core when they land on one of the SignalR .NET Framework tutorials. Given this is an include, we can update it frequently/easily as we go from preview to RTW.

I would like to mention some of the benefits of SignalR Core for new development over the .NET Framework versions. If there's no benefit, why go to the Core version.

I added the following to my MVC 5 tutorial, and it's been effective at driving traffic to the Core version:

 "An updated version of this tutorial is available [here](https://docs.microsoft.com/aspnet/core/tutorials/first-mvc-app/start-mvc) using the latest version of [Visual Studio.](https://www.visualstudio.com) The new tutorial uses [ASP.NET Core MVC](https://docs.microsoft.com/aspnet/core/mvc/), which provides **many** improvements over this tutorial."

All of that's true. The tutorial has many improvement besides those directly related to Core. Core has lots of benefits.

cc @scottaddie

rachelappel:
Someone write up some text and I'll put it in there. @wadepickett said SEO is always something to consider, and that [here] link is a bad SEO practice. Since any suggestions for included text by me has been rejected 100% of the time, I think other people should decide what the text should be from now on and just have me do the actual includes instead anymore.

Rick-Anderson:
@rachelappel  How about something like 
The preview version of SignalR for .NET Core documentation is [here](https://docs.microsoft.com/aspnet/core/signalr/introduction).

After it RTW's, we can push harder for folks to use the Core version.

scottaddie:
@Rick-Anderson Nit: change ".NET Core" to "ASP.NET Core" in your suggested text. Without that change, we're implying SignalR doesn't work when targeting .NET Framework.

PR #5775, opened 2018-03-22

harmoniemand:
dependency injection documentation updated with a small warning on scoped dependencies


guardrex:
Quick grammatical observation ...

> it forces the service to behave like a singleton

It's importance implies that it's an essential clause. I recommend promoting it in the sentence. Just a suggestion ...

> Don't inject via constructor injection because it forces the service to behave like a singleton.

Also tiny nit ... it's missing a space after the angle bracket:

```
> [!WARNING]
```

pranavkm:
@pakrym could you have a look?

pakrym:
LGTM

scottaddie:
@harmoniemand Thank you very much for adding this warning!

Issue #5764, opened 2018-03-22
ASP.NET Core Blazor (landing page) topic


guardrex:
Title: ASP.NET Core Blazor
UID: client-side/blazor/index
URL: https://docs.microsoft.com/aspnet/core/client-side/blazor/

* TOC
* Define & describe
* Link out .NET site Get started content
* aspnet/Docs issues
* Filing bugs/suggestions in Blazor repo
* Link Blazor repo release notes

guardrex:
Moved to https://github.com/aspnet/Blazor.Docs/issues/11

PR #5782, opened 2018-03-22
Blazor landing page topic

guardrex:
Addresses #5764 

Notes:
* Relies on .NET site for feature list and quick start.
* Prefer the tech list on line 20 inline (not in an unordered list) because I don't want to push the TOC down on the page any more than necessary.
* Will return to this with future commits as other topics are merged to update the descriptions.


guardrex:
We'll formally review this one when it's in final form. More commits coming on it.

PR #5785, opened 2018-03-23
Update azure-ad-b2c-webapi.md

Gutek:
Adding information about **Scope** url - how it is constructed. as this *api* part was causing a lot of issues in our dev team.

[Internal Review Page](https://review.docs.microsoft.com/en-us/aspnet/core/security/authentication/azure-ad-b2c-webapi?branch=pr-en-us-5785#obtain-a-bearer-token)


scottaddie:
@Gutek Thank you very much for bringing this to our attention and for taking the time to address it. I can definitely understand how this would be confusing.

PR #5823, opened 2018-03-29
Added database check.

Aaron-KSOC:
In the most up to date version of .Net Core running the code as it was constantly tried to recreate the database which errored it out.

I would have simply left this as a comment on the page but commenting is apparently being overhauled.  I've never done a pull request on GitHub so if I've done something bad here please let me know.

### DIFF
--- a/aspnetcore/tutorials/razor-pages/razor-pages-start/sample/RazorPagesMovie/Program.cs
+++ b/aspnetcore/tutorials/razor-pages/razor-pages-start/sample/RazorPagesMovie/Program.cs
@@ -6,6 +6,8 @@ using Microsoft.Extensions.Logging;
 using RazorPagesMovie.Models;
 using System;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Storage;
 
 namespace RazorPagesMovie
 {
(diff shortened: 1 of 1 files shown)


guardrex:
Would be nice tho to combine the two checks. There's already one check in `Initialize` ...

https://github.com/aspnet/Docs/blob/master/aspnetcore/tutorials/razor-pages/razor-pages-start/sample/RazorPagesMovie/Models/SeedData.cs#L20

Aaron-KSOC:
That's true but unfortunately it was passing that check and then failing when it tried to recreate the database anyway.  I'm not sure why...

Rick-Anderson:
@bricelam  per [Appling migrations in production](https://docs.microsoft.com/en-us/aspnet/core/data/ef-rp/migrations#appling-migrations-in-production)

We recommend production apps should not call Database.Migrate at application startup. ... (more on the topic)

But we've found adding it to our sample lowers friction of customers running the sample app.

Having said that, is the the pattern we should follow?

bricelam:
I'll defer to @ajcvickers and @divega

Rick-Anderson:
per [Appling migrations in production](https://docs.microsoft.com/en-us/aspnet/core/data/ef-rp/migrations#appling-migrations-in-production)

We recommend production apps should not call `Database.Migrate` at application startup. ... (more on the topic)

But we've found adding `Migrate`  to our sample lowers friction of customers running the sample app.

 @ajcvickers or @divega can you check this 2 line change?

ajcvickers:
@Rick-Anderson There should never be a need for an exists check before calling Migrate, unless the database was not created using Migrations, but in that case Migrate should not be there at all.

bricelam:
Yeah, I don't think it helps anything. The original description is interesting--that certainly shouldn't happen.

PR #5827, opened 2018-03-29
Update first two lines of Options pattern topic

guardrex:
Reader feedback from @Baharestani ...

> The first line is referring to something called "option classes" which doesn't make any sense to a reader who is not familiar with Options Pattern.

The use of "options class" now first appears on line 29 where it's defined.

Thank you @Baharestani for pointing this out. :rocket:


Rick-Anderson:
Thanks @guardrex  and  @baharestani
It reads much better now.

PR #5851, opened 2018-04-02
await added

rachelappel:
fixes https://github.com/aspnet/Docs/issues/5850

### DIFF
--- a/aspnetcore/signalr/hubs/sample/ChatHub.cs
+++ b/aspnetcore/signalr/hubs/sample/ChatHub.cs
@@ -25,7 +25,7 @@ namespace SignalRChat
 
         public override Task OnConnectedAsync()
         {
-            Groups.AddAsync(Context.User.Identity.Name, "SignalR Users");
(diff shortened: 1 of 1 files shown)


aspnetcore/signalr/hubs/sample/ChatHub.cs
```diff
@@ -25,7 +25,7 @@ public Task SendMessageToGroups(string user, string message)
 
         public override Task OnConnectedAsync()
```
BrennanConroy: This now needs to be `public override async Task ...`

aspnetcore/signalr/hubs/sample/ChatHub.cs
```diff
@@ -25,7 +25,7 @@ public Task SendMessageToGroups(string user, string message)
...
         {
-            Groups.AddAsync(Context.User.Identity.Name, "SignalR Users");
+            await Groups.AddAsync(Context.User.Identity.Name, "SignalR Users");
             return base.OnConnectedAsync();
```
BrennanConroy: change `return` to `await`

PR #5852, opened 2018-04-02
Update partial-tag-helper.md

Jaxelr:
Corrected language syntax of ProductModel from cshtml to csharp.


scottaddie:
@Jaxelr Nice catch! Thank you for correcting this.

PR #5861, opened 2018-04-03
SignalR JavaScript client

rachelappel:
fixes https://github.com/aspnet/Docs/issues/5646 
[Int Review](https://review.docs.microsoft.com/en-us/aspnet/core/signalr/javascript-client?branch=pr-en-us-5861)


aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,108 @@
...
+
+While the preceding code sample creates a connection to a hub named `ChatHub`, it doesn't start it. To start the connection, call the `connection.start` method.
+
+```javascript
+connection.start().catch(err => logError(err));
+```
+
+### Specify the connection's transport type
+
+Transports define how servers and clients communicate. To set the transport type when you create the connection by setting the `signalR.TransportType` to *WebSockets*, *LongPolling*, or *ServerSentEvents*. The default, recommended transport is *WebSockets*.
```
Rick-Anderson: This has a few problem. Can you show more code? `transportType` is not shown, I think you have the range wrong?

Something like, the following code sets the trans type to WebSockets

Then below tell them the types they can set.

I think I'd combine this with the previous snippet.  better to show code and explain it.

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,108 @@
...
+
+[!code-javascript[Specify transport](javascript-client/sample/chat.js?range=1-3)]
+
+### Cross-origin connections
+
+Typically, browsers load connections from the same domain as the requested page. However, there are occasions when a connection to another domain is required.
+
+For better security, [cross-origin connections](xref:security/cors) are disabled by default. If you need to create a cross-origin request, configure and enable it in the `Startup` class.
```
Rick-Anderson: To prevent a malicious site from reading sensitive data from another site, [cross-origin connections](xref:security/cors) are disabled by default.  To allow  a cross-origin request, configure and enable it in the `Startup` class.

Rick-Anderson: When "you" helps make the concept more clear, use it. In many cases, it's better to remove  "you".

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,108 @@
...
+
+[!code-csharp[Cross-origin connections](javascript-client/sample/startup.cs?highlight=22-29,39-40)]
+
+## Call hub methods from client
+
+JavaScript clients can call public methods on hubs by issuing `connection.invoke`. The `invoke` method accepts two arguments:
```
Rick-Anderson: JavaScript clients ~can~ call public methods on hubs with `connection.invoke`.

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Install the SignalR client package
+
+The SignalR JavaScript client library is delivered as an [npm](https://www.npmjs.com/) package. If you're using Visual Studio, call `npm install` from the command line in any folder. For Visual Studio Code, call the command from the **Integrated Terminal**.
+
+  ```console
+   npm install @aspnet/signalr
+  ```
+
+Npm installs the package contents in the *node_modules\\@aspnet\signalr\dist\browser* folder. Copy the *signalr.js* or its minified version to a folder in your project.
```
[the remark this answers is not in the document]
Rick-Anderson: Need to add 
`[View or download sample code](https://github.com/aspnet/Docs/tree/live/aspnetcore/tutorials/razor-pages/razor-pages-start/sample/RazorPagesMovie) ([how to download](xref:tutorials/index#how-to-download-a-sample))`

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,108 @@
...
+connection.start().catch(err => logError(err));
+```
+
+### Specify the connection's transport type
+
+Transports define how servers and clients communicate. To set the transport type when you create the connection by setting the `signalR.TransportType` to *WebSockets*, *LongPolling*, or *ServerSentEvents*. The default, recommended transport is *WebSockets*.
+
+[!code-javascript[Specify transport](javascript-client/sample/chat.js?range=1-3)]
+
+### Cross-origin connections
```
Rick-Anderson: Add 
The complete JavaScript:
(add js file ref here)

aspnetcore/signalr/javascript-client.md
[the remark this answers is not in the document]
rachelappel: @stimms Trying to keep it so that it's SignalR only, rather than info on the bundling/minification. Will add more explicity copy path.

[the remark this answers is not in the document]
Rick-Anderson: I agree with @rachelappel, need to keep it simple.
@anurse  should we have a comment that using *bundleconfig.json* is an alternative?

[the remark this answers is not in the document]
stimms: Yeah that's fair: limit the scope to just what we are documenting

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,108 @@
...
+
+# ASP.NET Core SignalR JavaScript client
+
+By [Rachel Appel](http://twitter.com/rachelappel)
+
+The ASP.NET Core SignalR JavaScript client library enables developers to call server-side hub code.
+
+## Install the SignalR client package
+
+The SignalR JavaScript client library is delivered as an [npm](https://www.npmjs.com/) package. If you're using Visual Studio, call `npm install` from the command line in any folder. For Visual Studio Code, call the command from the **Integrated Terminal**.
```
anurse: I think "**run** `npm install`" and "**run** the command from the Integrated Terminal" are clearer

aspnetcore/signalr/javascript-client.md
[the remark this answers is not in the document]
anurse: We should definitely keep this simple. If there's an existing article on managing client-side dependencies using Bundling, we could link to that.

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,108 @@
...
+
+To use the SignalR JavaScript client, a script reference is required. Since the reference is an HTML `<script>` element, it can go in any kind of client-side page, such as a [Razor Page](xref:mvc/razor-pages/index?tabs=visual-studio), [MVC view](xref:mvc/views/overview), or HTML page.
+
+```html
+<script src="~/scripts/signalr.js"></script>
+```
+
+## Connect to a hub
+
+To connect to a SignalR hub, create a new instance of a connection with the route as its argument. The route maps to a hub with the same name. However, the hub's name can be mixed case, such as `ChatHub`.
```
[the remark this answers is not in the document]
anurse: There's no need to reference the hub name here. The only parameter to the constructor is a **URL** to the hub, which should match what was registered in the `MapHub` call on the server.

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,108 @@
...
+
+```javascript
+const connection = new signalR.HubConnection('/chathub');
```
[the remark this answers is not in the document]
anurse: This should match the path used in `MapHub`, which is `/chathub` in this case.

aspnetcore/signalr/javascript-client.md
anurse: To be clear, the default, recommended pattern is to **not provide the transport type at all**. It is not required and by default we will automatically choose an appropriate transport.

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,108 @@
...
+
+For better security, [cross-origin connections](xref:security/cors) are disabled by default. If you need to create a cross-origin request, configure and enable it in the `Startup` class.
+
+[!code-csharp[Cross-origin connections](javascript-client/sample/startup.cs?highlight=22-29,39-40)]
+
+## Call hub methods from client
+
+JavaScript clients can call public methods on hubs by issuing `connection.invoke`. The `invoke` method accepts two arguments:
+
+* The name of the hub's method. In the following example, the hub name is `SendMessage`.
```
anurse: "The name of the hub method"

We generally just use the compound noun "hub method" to refer to methods on hubs, rather than using a possessive like "hub's method"

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,108 @@
...
+* Any arguments defined in the hub's method. In the following example, the argument name is `message`.
```
anurse: Ditto, "hub method"

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,108 @@
...
+
+[!code-javascript[Call hub methods](javascript-client/sample/chat.js?range=13-16)]
+
+## Call client methods from hub
+
+To receive messages from the hub, define a method in the client's `connection.on` method. The `connection.on` method requires the following information:
+
+* The name of the JavaScript client's method. In the following example, the method name is `ReceiveMessage`.
```
anurse: "The name of the client method. ..."

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,108 @@
...
+* Arguments the hub passes to the method. In the following example, the argument value is `message`.
+* The method's code.
```
anurse: There are only two parameters to the `.on` method: The name of the method, and a JavaScript function to implement the method. I wouldn't describe the arguments and code as separate parameters, since they aren't parameters.

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,108 @@
...
+[!code-javascript[Receive calls from hub](javascript-client/sample/chat.js?range=6-9)]
+
+The preceding code in `connection.on` runs when  server-side code calls it using the `SendAsync` method.
+
+[!code-javascript[Call client-side](javascript-client/sample/chathub.cs?range=8-11)]
+
+SignalR determines which client method to call by matching the method name and arguments defined in `SendAsync` and `connection.on`.
+
+> [!NOTE]
+> As a best practice, call `connection.start` after `connection.on` so the code is properly loaded.
```
anurse: "... so your handlers are registered before any messages are received"

aspnetcore/signalr/javascript-client/sample/chat.js
```diff
@@ -0,0 +1,20 @@
+﻿const transportType = signalR.TransportType.WebSockets;
+const logger = new signalR.ConsoleLogger(signalR.LogLevel.Information);
+const connection = new signalR.HubConnection('/chathub', 
+                   { transport: transportType, logger: logger });
```
anurse: As I said above, drop the `transport` option. We don't recommend it unless you are a more advanced user.

aspnetcore/signalr/javascript-client.md
rachelappel: Do you want it to list just the two actual arguments? I was hoping that saying "information" rather than "arguments" -- "The `connection.on` method requires the following information:" made it clear it wasn't just arguments and everything needed for the method.

rachelappel:
@anurse Any more change requests before merging?

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Connect to a hub
+
+The following code creates a connection. The hub's name is case insensitive.
+
+```javascript
+const connection = new signalR.HubConnection('/chathub');
+```
+
+While the preceding code sample creates a connection to a hub named `ChatHub`, it doesn't start it. To start the connection, call the `connection.start` method.
```
Rick-Anderson: These two steps need to be combined into on as I previously suggested.

rachelappel:
@Rick-Anderson This isn't a tutorial. Those steps are only supposed to go in tutorials.

Rick-Anderson:
@rachelappel  we still want a a sentence or something telling them how to connect this sample code and get it working. By itself it doesn't do anything.

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Error handling and logging
+
+Chain a `catch` method to the end of the `connection.start` method to handle client-side errors. Use `console.error` to output errors to the browser's console.
+
+[!code-javascript[Error handling](javascript-client/sample/chat.js?range=9)]
+
+Setup client-side log tracing by passing a logger and type of event to log when the connection is made. Available log levels are as follows:
+
+* `signalR.LogLevel.Information` : Status messages without errors.
```
anurse: > Status messages without errors.

Not quite accurate, if you specify a log level, we'll log messages at that level **or above**. The highest level is `Error`, and the lowest is `Trace` (which is much more detailed and not generally useful for users). So if you set the log level to `Information`, you'll get `Information`, `Warning` and `Error` messages logged. If you set it to `Warning`, you'll just get `Warning` and `Error`, if you set it to `Trace`, you get every dang thing ;).

aspnetcore/signalr/javascript-client/sample/Startup.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
+        {
+            app.UseSignalR(routes =>
+            {
+                routes.MapHub<ChatHub>("/chathub");
+            });
+            app.UseMvc();
+            app.UseCors(x => x.AllowAnyOrigin().AllowAnyMethod()
```
anurse: I don't think you need to specify the policy in both places, but I'm not as familiar with the nuances of Cors configuration.

aspnetcore/signalr/javascript-client/sample/chat.js
```diff
@@ -0,0 +1,14 @@
+﻿const logger = new signalR.ConsoleLogger(signalR.LogLevel.Information);
+const connection = new signalR.HubConnection('/chathub', logger: logger });
```
anurse: There's a shorthand for using the console logger at a particular level. If you want `ConsoleLogger` but at a specific level, you can just write the following (no need to construct a `ConsoleLogger`)

```
const connection = new signalR.HubConnection('/chathub', logger: signalR.LogLevel.Information)
```

rachelappel:
@Rick-Anderson I added the entire code sample so then they only need the download instructions already 
there.

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,102 @@
...
+
+By [Rachel Appel](http://twitter.com/rachelappel)
+
+The ASP.NET Core SignalR JavaScript client library enables developers to call server-side hub code.
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/live/aspnetcore/signalr/javascript-client/sample) ([how to download](xref:tutorials/index#how-to-download-a-sample))
+
+## Install the SignalR client package
+
+The SignalR JavaScript client library is delivered as an [npm](https://www.npmjs.com/) package. If you're using Visual Studio, run `npm install` from the command line in any folder. For Visual Studio Code, run the command from the **Integrated Terminal**.
```
scottaddie: - Since most Visual Studio users don't like leaving the IDE, have them use the Package Manager Console window.
- The *package.json* file needs to exist before `npm install` will succeed. Add the following command before that one: `npm init -y`.

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,102 @@
...
+
+  ```console
+   npm install @aspnet/signalr
+  ```
+
+Npm installs the package contents in the *node_modules\\@aspnet\signalr\dist\browser* folder. Create a new folder named *signalr* under the *wwwroot\\lib* folder. Copy the *signalr.js* to the *wwwroot\lib\signalr* folder.
```
scottaddie: Add the word "file" after *signalr.js*

aspnetcore/signalr/javascript-client.md
```diff
@@ -0,0 +1,102 @@
...
+## Call client methods from hub
+
+To receive messages from the hub, define a method using the `connection.on` method.
+
+* The name of the JavaScript client method. In the following example, the method name is `ReceiveMessage`.
+* Arguments the hub passes to the method. In the following example, the argument value is `message`.
+
+[!code-javascript[Receive calls from hub](javascript-client/sample/wwwroot/js/chat.js?range=4-9)]
+
+The preceding code in `connection.on` runs when  server-side code calls it using the `SendAsync` method.
```
scottaddie: Remove the extra space before "server-side"

PR #5873, opened 2018-04-04
Update IIS Modules topic

### LINKED ISSUE #5844: "Minimum module configuration" for IIS + ANCM caused confusing for perf test
https://github.com/aspnet/Docs/blob/master/aspnetcore/host-and-deploy/iis/modules.md

Looks like the "Minimum module configuration" in this doc caused a lot of confusion among users - while it is the right minimum requirement to make ASP.NET core app work, people think it's the fastest configuration and use it in all sorts of perf tests:

aspnet/IISIntegration#747
aspnet/Performance#186

I think it might worth to make some clarifications in this doc.

guardrex:
Fixes #5844 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/modules?branch=pr-en-us-5873)

* Adds notes on two modules that aren't strictly speaking "required" for an ASP.NET Core app to run in reverse proxy with IIS but whose perf is affected when the module isn't active.
* Refactors the managed modules table to remove a column (and adds a line of text).
* Renames a column (and adds a line of text) for the native modules column.
* Corrects a mistake with a module name.
* Remove the config code to disable "default" IIS modules because that's too variable really to be effective given how variable module config will be server-to-server and organization-to-organization.

Thanks to @bangbingsyb :rocket: for highlighting the problems with the minimum modules section.

### DIFF
--- a/aspnetcore/host-and-deploy/iis/modules.md
+++ b/aspnetcore/host-and-deploy/iis/modules.md
@@ -59,7 +59,7 @@ The table indicates native IIS modules that are functional on reverse proxy requ
 
 ## Managed modules
 
-Managed modules are *not* functional with hosted ASP.NET Core apps. ASP.NET Core offers middleware alternatives in several cases.
+Managed modules are *not* functional with hosted ASP.NET Core apps when the app pool's .NET CLR version is set to **No Managed Code**. ASP.NET Core offers middleware alternatives in several cases.
 
 | Module                  | ASP.NET Core Option |
 | ----------------------- | ------------------- |
--- a/aspnetcore/host-and-deploy/iis/modules.md
+++ b/aspnetcore/host-and-deploy/iis/modules.md
@@ -5,7 +5,7 @@ description: Discover active and inactive IIS modules for ASP.NET Core apps and 
 manager: wpickett
 ms.author: riande
 ms.custom: mvc
-ms.date: 02/15/2018
+ms.date: 04/04/2018
 ms.prod: aspnet-core
 ms.technology: aspnet
 ms.topic: article
@@ -19,8 +19,10 @@ ASP.NET Core apps are hosted by IIS in a reverse proxy configuration. Some of th
 
 ## Native modules
 
-| Module | .NET Core Active | ASP.NET Core Option |
-| ------ | :--------------: | ------------------- |
+The table indicates native IIS modules that are functional on reverse proxy requests to ASP.NET Core apps.
+
+| Module | Functional with ASP.NET Core apps | ASP.NET Core Option |
+| ------ | :-------------------------------: | ------------------- |
 | **Anonymous Authentication**<br>`AnonymousAuthenticationModule` | Yes | |
 | **Basic Authentication**<br>`BasicAuthenticationModule` | Yes | |
 | **Client Certification Mapping Authentication**<br>`CertificateMappingAuthenticationModule` | Yes | |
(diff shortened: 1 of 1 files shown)


aspnetcore/host-and-deploy/iis/modules.md
```diff
@@ -151,6 +139,10 @@ The only modules required to run an ASP.NET Core app are the Anonymous Authentic
 
 ![IIS Manager open to Modules with the minimum module configuration shown](modules/_static/modules.png)
 
+The URI Caching Module (`UriCacheModule`) allows IIS to cache website configuration at the URL level. Without this module, IIS must read and parse configuration on every request, even when the same URL is repeatedly requested. Parsing the configuration every request results in a significant performance penalty. *Although the URI Caching Module isn't strictly required for a hosted ASP.NET Core app to run, we recommend that the URI Caching Module be enabled for all ASP.NET Core deployments.*
```
bangbingsyb: Many thanks for your update. The additional notes here looks great and accurate to me.

pan-wang:
For managed modules sections, the statement "Managed modules are not functional with hosted ASP.NET Core apps." is incorrect. Theoretically, user still can have managed module enabled and be part of their application, e.g., the OutputCache a,d authentication. May be we should change to "not recommended"

aspnetcore/host-and-deploy/iis/modules.md
```diff
@@ -57,21 +59,23 @@ ASP.NET Core apps are hosted by IIS in a reverse proxy configuration. Some of th
...
-| OutputCache             | No               | [Response Caching Middleware](xref:performance/caching/middleware) |
-| Profile                 | No               | |
-| RoleManager             | No               | |
-| ScriptModule-4.0        | No               | |
-| Session                 | No               | [Session Middleware](xref:fundamentals/app-state) |
-| UrlAuthorization        | No               | |
-| UrlMappingsModule       | No               | [URL Rewriting Middleware](xref:fundamentals/url-rewriting) |
-| UrlRoutingModule-4.0    | No               | [ASP.NET Core Identity](xref:security/authentication/identity) |
-| WindowsAuthentication   | No               | |
+Managed modules are *not* functional with hosted ASP.NET Core apps. ASP.NET Core offers middleware alternatives in several cases.
```
bangbingsyb: Technically these managed modules are functional with the ASP.NET Core apps, although the middleware provides alternatives for a subset :)

guardrex:
Our guidance is to set No Managed Code, so wouldn't they be dysfunctional with that setting in place?

guardrex:
@pan-wang I added a qualifier to the remark. Is it still inaccurate because the .NET CLR version only affects the ASP.NET parts of the module (and the IIS parts still work)?

If it is still inaccurate, I agree that we should just say that using managed modules isn't recommended. If we were to say that "some of these work ... some don't," then we're going to need to specify exactly which ones will work and under what conditions.

In trying to understand how this happened, I conclude that we only had a quick engineering look and not really a formal review and sign-off on https://github.com/aspnet/Docs/pull/2609. In the future, I'll try to prevent that from happening again.

guardrex:
ping @pan-wang

pan-wang:
Yes, our recommendation is not configure IIS application pool to "No Managed Code" for better performance. In such case, managed module will not be loaded and thus will not be functional.

guardrex:
@pan-wang I just want to make sure I'm reading you correctly ...

> our recommendation **is not configure IIS application pool to "No Managed Code"**

... *is not configure*? Did you mean to type "**is to configure**"?

> I am OK with current version of the doc.

... *current version of the doc*? ... Do you mean "**current version of the PR**"?

Sorry for my misunderstanding, but you didn't mark the PR approved using the GH Review System, and I can't tell if you're rejecting or approving this PR.

pan-wang:
My fault. I typed on my cellphone and made some typos.  Yes,  your understanding is correct.  Configure apppool with no managed code and  approval for the pr.

PR #5875, opened 2018-04-04
Update SeedData.cs

sdavis3:
Several references to Movie should be plural Movies based on the previous exercise. It is a DbSet<Movie> Movies that is declared in MovieContext.cs, thus Movies is required.


Rick-Anderson:
@sdavis3  did you try this? It's actually `public DbSet<Movie> Movie { get; set; }`

PR #5883, opened 2018-04-05
@function description change

guardrex:
@Rick-Anderson I used this in a Blazor topic, but `@SteveSandersonMS` recommended that it just say "C# code block".


guardrex:
Also note that it exploded in the email for the build (but it seems ok here on the PR) ...

capture

PR #5895, opened 2018-04-06
Note about task runner explorer option

user135711:
The gulp file has to be in the root directory or in the directory with package.json not in any subdirectory.


scottaddie:
@user135711 Thank you for your contribution!

PR #5901, opened 2018-04-07
authorization requires an authentication mechanism

### LINKED ISSUE #5870: Unclear documentation for Authorization Policy/Authentication
Reposted from: https://github.com/aspnet/Security/issues/1715

I've attempted to follow this page: https://docs.microsoft.com/en-us/aspnet/core/security/authorization/policies#using-a-func-to-fulfill-a-policy

I performed the following steps (a simplified contrived example):
* I created a new Web API project `dotnet new webapi`
* On the `Get` method of the values controller I added an attribute: `[Authorize(Policy = "AlwaysDeny")]`
* In `Startup.cs` I added the following to `ConfigureServices`:
```
      services.AddAuthorization(options =>
      {
        options.AddPolicy("AlwaysDeny", policy =>
        {
          policy.RequireAssertion(context => false);
        });
      });
```
When I get `http://localhost:5000/api/values` I would have expected a 401 response

I instead get a 500 the following exception:  `System.InvalidOperationException: No authenticationScheme was specified, and there was no DefaultChallengeScheme found.`

The documentation does not indicate any requirement for authentication. I do not want to use JWT token or Cookies.

Rick-Anderson:
@dave-hillier  please review.

Fixes #5870

### DIFF
--- a/aspnetcore/security/authorization/introduction.md
+++ b/aspnetcore/security/authorization/introduction.md
@@ -16,7 +16,7 @@ uid: security/authorization/introduction
 
 Authorization refers to the process that determines what a user is able to do. For example, an administrative user is allowed to create a document library, add documents, edit documents, and delete them. A non-administrative user working with the library is only authorized to read the documents.
 
-Authorization is orthogonal and independent from authentication, which is the process of ascertaining who a user is. Authentication may create one or more identities for the current user. However, authorization requires an authentication mechanism.
+Authorization is orthogonal and independent from authentication. Authentication is the process of ascertaining who a user is. Authentication may create one or more identities for the current user. However, authorization requires an authentication mechanism.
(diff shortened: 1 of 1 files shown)


blowdart:
@dave-hillier do you think it needs to be a reminder on all the other pages as well, in case people don't even read the introduction?

dave-hillier:
I assume people can land anywhere. Not sure which other pages you mean so
I’ll take a look later.

I think a big part of my issue was that the examples on the page looked
complete, so perhaps they can be made to look more like code snippets

Rick-Anderson:
@dave-hillier  I'll monitor these doc's and see if I should an an include file to each doc saying authentication is required. thanks for pointing this out. I've never tried authz without auth.

PR #5904, opened 2018-04-09
Fix typo

LindaLawton:
Fixing typo  The other two examples have the attribute called 

>[IdentityBasicAuthentication] // Enable Basic authentication for this controller.

This section will also need to use IdentityBasicAuthentication

IdentityBasicAuthentication


scottaddie:
@LindaLawton We appreciate the contribution; however, the sample is correct as it stands.

When instantiated directly (to apply as a global filter), the "Attribute" suffix is required. When used to decorate a class member (e.g., `[AttributeName]`), the "Attribute" suffix can be omitted.

LindaLawton:
Are you sure I just ran through it today and I gives a error that it can't be found until you fix the name.

PR #5926, opened 2018-04-10
Change folder and description of the anchor.

fesnarriaga:
I am proposing this change because the document below references this document.
https://docs.microsoft.com/en-us/aspnet/core/tutorials/razor-pages/page
If another tutorial makes an Schedule Page and points to this page, please do not approve this change.

### DIFF
--- a/aspnetcore/tutorials/razor-pages/razor-pages-start/sample/RazorPagesMovie/Pages/_Layout.cshtml
+++ b/aspnetcore/tutorials/razor-pages/razor-pages-start/sample/RazorPagesMovie/Pages/_Layout.cshtml
@@ -31,7 +31,7 @@
             <div class="navbar-collapse collapse">
                 <ul class="nav navbar-nav">
                     <li><a asp-page="/Index">Home</a></li>
-                    <li><a asp-page="/Schedules/Index">Schedules</a></li>
+                    <li><a asp-page="/Movies/Index">Movies</a></li>
                     <li><a asp-page="/About">About</a></li>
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@guardrex  we shouldn't mix up schedules in the main RP code - should we? Will this break your add on tutorial?

Upload files should be a standalone tutorial. It's OK that it builds on the Movies tutorial. This is a long term fix, not to worry about now.

guardrex:
The downloadable sample is the *completely functioning* sample that has all of the steps in the tutorial applied; it's the final finished app. Therefore, I thought the internal links link everywhere they would link in the final app. It's true that we don't instruct in the File Uploads step to add this link (perhaps we should), but it makes it convenient to reach the page when the sample is downloaded.

As for this specific change, I'd need to take a closer look. I can look in 20 minutes ... I'm knee-deep in another thing at the moment. 😅

guardrex:
Ok ... I had a peek. I think the fix is to add this link in the steps of the File Uploads tutorial step and leave the link as it is in this layout page (given that this sample is the final, built sample app).

Perhaps, the Movie page step should also have an instruction to tell the reader to add a link to Movies (if it doesn't do so already). IIRC, the **RpMovie** links to the Movie page, so it might not be necessary. It might be necessary tho to tell the reader to put the **RpMovie** link into the app as they're building it.

I opened #5931 to address this.

Thanks @fesnarriaga for pointing this out. I recommend closing here, and I'll fix it on #5931.

fesnarriaga:
@guardrex, glad I could help.

PR #5970, opened 2018-04-13
Address location of Razor syntax doc

scottaddie:
Fixes https://github.com/aspnet/Docs/issues/5969

### DIFF
--- a/aspnetcore/toc.md
+++ b/aspnetcore/toc.md
@@ -144,7 +144,6 @@
 
 # [MVC](mvc/overview.md)
 ## [Razor Pages](xref:mvc/razor-pages/index)
-### [Razor syntax](mvc/views/razor.md)
 ### [Route and app convention features](xref:mvc/razor-pages/razor-pages-convention-features)
 ### [Filter methods for Razor Pages](xref:mvc/razor-pages/filter)
 ### [Razor SDK](xref:mvc/razor-pages/sdk)


guardrex:
I'm wondering what a reader will perceive about it only being under `Views`.

scottaddie:
@guardrex Where else would it go? I think it's more confusing to put the doc in multiple locations. That's a sign that the doc's focus isn't narrow enough.

guardrex:
I ask because the technology applies to both app frameworks. Can it sit out under `MVC`? ... or would that cause a new problem?

scottaddie:
@guardrex I suggest we move Razor Pages out to the same level as MVC. Then we can put the **Razor syntax** doc back under it. Thoughts?

scottaddie:
@guardrex I appreciate the feedback. It's useful to get another perspective.

Web API is technically considered MVC, yet we got approval from Dan to move that content out into its own TOC node. Most customers I talk to consider Razor Pages to be its own app model (though we know it isn't). As a customer, I'd want to see Razor Pages at that top level. As an insider, I'd prefer to see the Razor Pages node within MVC. See the dilemma?

Rick-Anderson:
I did alot of test on the [internal review](https://review.docs.microsoft.com/en-us/aspnet/core/mvc/overview?view=aspnetcore-2.0&branch=pr-en-us-5970) and I can't find anything wrong. For sure it's a huge improvement.

Rick-Anderson:
> @Rick-Anderson Are you okay with my proposal? ☝️ 

Good idea. Let's do that in another PR. I'd like to merge this now and make it live.

PR #5981, opened 2018-04-14
fixed globbing pattern.

user135711:
Good information. The old globbing pattern (ignoring the right parenthesis typo) would match every input so I fixed it. A third way I thought to do it would be: `/**/!(*.!(css))` which prevents min because the *. before the parenthesis will match only the first . and not the last `.`. Negating twice gets the .css extension back. I decided for the other way because this will also match .css in the middle of words.

### DIFF
--- a/aspnetcore/client-side/bundling-and-minification.md
+++ b/aspnetcore/client-side/bundling-and-minification.md
@@ -200,7 +200,7 @@ To minify *custom.css* and bundle it with *site.css* into a *site.min.css* file,
 > Alternatively, the following globbing pattern could be used:
 >
 > ```json
-> "inputFiles": ["wwwroot/**/*(*.!(min).css)"]
+> "inputFiles": ["wwwroot/**/@(*!(min).css)"]
 > ```
 >
 > This globbing pattern matches all CSS files and excludes the minified file pattern.
--- a/aspnetcore/client-side/bundling-and-minification.md
+++ b/aspnetcore/client-side/bundling-and-minification.md
@@ -200,7 +200,7 @@ To minify *custom.css* and bundle it with *site.css* into a *site.min.css* file,
 > Alternatively, the following globbing pattern could be used:
 >
 > ```json
-> "inputFiles": ["wwwroot/**/*(*.css|!(*.min.css)"]
+> "inputFiles": ["wwwroot/**/*(*.!(min).css)"]
 > ```
 >
 > This globbing pattern matches all CSS files and excludes the minified file pattern.


guardrex:
I haven't been using BundlerMinifier, as I've been more favorable to straight Gulp (where I've never had a problem). I haven't used BundlerMinifier much ... just doc samples. I tested various negate patterns in VS with BundlerMinifier, and I had trouble with them.

My test setup was ...

*testwwwroot/site1.css*:

```css
body {
    padding-top: 50px;
    padding-bottom: 20px;
}
```

*testwwwroot/site1.min.css*:

```css
.someclass {
    padding-right:150px;
}
```

*bundleconfig.json*:

```json
{
  "outputFileName": "wwwroot/css/siteoutput.min.css",
  // An array of relative input file paths. Globbing patterns supported
  "inputFiles": [
    "testwwwroot/**/@(*!(min).css)"
  ]
},
```

... and if it were to work, I should get only the css from *site1.css* in the *siteoutput.min.css* file. Instead, I get no *siteoutput.min.css* file.

When I change the pattern to drop the match-once logic ...

```json
"inputFiles": [
  "testwwwroot/**/*!(min).css"
]
```

... I get both in *siteoutput.min.css* ...

```css
body{padding-top:50px;padding-bottom:20px}.someclass{padding-right:150px}
```

... then I tried a half-dozen other approaches and same thing. I either get no *siteoutput.min.css* file, or the file contains the CSS from both input files.

Can you cats see where I'm going wrong here?

[On the bright side, it looks like this whole technology is going away on a 2.1 patch release ... we're going to LibMan.]

user135711:
@guardrex I am getting the same results. On a generalized globtester I can get double negative globs style globs to work [1](http://www.globtester.com/#p=eJxT1C8tLtLX0tKvVtTSSy4u1tHKzcwDMWoBZH8IKA%3D%3D&r=eJy1kM0KwjAQhF8l5KQgXetBoqZ5l%2FxRC%2FmRbG2pT2%2BkBb1UtOhlTzPfzA43TUe0k4gVpQKumMDE2kJn0wDSy5stfBMKjcghSwV%2F0XvZ6vNkclFLB6oJ0Mf%2BrZzjRQaC7eBsRXV0MR1JqtWK7TakZCyfA1uf5rDGZPKDILhK4stOZoHn07ojemz256XK7T7HTmd%2BquU7%2FfLnZ%2FwdgwHOgA%3D%3D&) [2](http://www.globtester.com/#p=eJxT1C8tLtLX0tKvVtTSSy4u1tHKzcwDMWoBZH8IKA%3D%3D&r=eJy1kM0KwjAQhF8l5KQgXetBoqZ5l%2FxRC%2FmRbG2pT2%2BkBb1UtOhlTzPfzA43TUe0k4gVpQKumMDE2kJn0wDSy5stfBMKjcghSwV%2F0XvZ6vNkclFLB6oJ0Mf%2BrZzjRQaC7eBsRXV0MR1JqtWK7TakZCyfA1uf5rDGZPKDILhK4stOZoHn07ojemz256XK7T7HTmd%2BquU7%2FfLnZ%2FwdgwHOgA%3D%3D&) but either everything or nothing matches in Visual studio. This may have been a non issue since the output min file never gets reappended so it doesn't matter unless you use two different bundler/minifiers.

user135711:
@guardrex I'm closing my pull request as I personally don't have a solution.

PR #5983, opened 2018-04-15
npm test directory wrong

user135711:
Running the command in the project root complains that no package.json can be found. You have to run the command from the ClientApp directory where the package.json is located also indicated in the instructions on the splash page "The ClientApp subdirectory is a standard Angular CLI application. If you open a command prompt in that directory, you can run any ng command (e.g., ng test), or use npm to install extra packages into it."


scottaddie:
@user135711 Thank you for your help!

PR #5988, opened 2018-04-16
Task continuation correction

user135711:
I think the example should be written using `task.ContinueWith` as the original implies the more calls could be made after the function already returned.


scottaddie:
@user135711 Thank you for your help!

PR #5990, opened 2018-04-16
fixed infinite loop error

user135711:
The original would loop infinitely if the file doesn't exist. (bug).

 Now, private void InvokeChanged(IHostingEnvironment env) should invoke computenhash() in a try catch... I could add that if desired

### DIFF
--- a/aspnetcore/fundamentals/primitives/change-tokens/sample/Utilities/Utilities.cs
+++ b/aspnetcore/fundamentals/primitives/change-tokens/sample/Utilities/Utilities.cs
@@ -23,7 +23,8 @@ namespace ChangeTokenSample.Utilities
                             return System.Security.Cryptography.SHA1.Create().ComputeHash(fs);
                         }
                     }
-                    else {
+                    else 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
LGTM
@scottaddie  can you review this.

guardrex:
@user135711 @scottaddie [FileNotFoundException](https://docs.microsoft.com/dotnet/api/system.io.filenotfoundexception) is an [IOException](https://docs.microsoft.com/dotnet/api/system.io.ioexception), so I don't think an endless loop would ensue for that scenario.

guardrex:
Just had a chance to take a closer look. Yes, thank you @user135711! :rocket:

guardrex:
@user135711 

> Now, private void InvokeChanged(IHostingEnvironment env) should invoke computenhash() in a try catch... I could add that if desired

Yes. If you'd like to submit that, please do. We appreciate the assistance! 🏃🏃🏃🏃🏃🏃🏃😅

PR #6000, opened 2018-04-17
Update sample2.xml

jasonruz:
Fix case of ClientID attribute in web.config example.


scottaddie:
@jasonruz Nice catch! Thank you for fixing this.

PR #6003, opened 2018-04-17
[WIP] Add Web API controller definition doc

scottaddie:
Addresses https://github.com/aspnet/Docs/issues/5401

This is a new doc explaining the various ways to create a Web API controller. The ASP.NET Core 2.1 `[ApiController]` attribute is covered in it.

### DIFF
--- a/aspnetcore/web-api/define-controller.md
+++ b/aspnetcore/web-api/define-controller.md
@@ -1,0 +1,117 @@
+---
+title: Define a controller in ASP.NET Core Web API
+author: scottaddie
+description: Learn about the options for defining a controller class in an ASP.NET Core Web API and when it's most appropriate to use each.
+manager: wpickett
+ms.author: scaddie
+ms.custom: mvc
+ms.date: 04/17/2018
+ms.prod: aspnet-core
+ms.technology: aspnet
+ms.topic: article
+uid: web-api/define-controller
+---
+# Define a controller in ASP.NET Core Web API
+
+By [Scott Addie](https://github.com/scottaddie)
+
+::: moniker range=">= aspnetcore-2.1"
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/web-api/define-controller/samples) ([how to download](xref:tutorials/index#how-to-download-a-sample))
+::: moniker-end
+
+ASP.NET Core offers the following options for creating a Web API controller:
+
+::: moniker range="<= aspnetcore-2.0"
+* [Derive class from Controller](#derive-class-from-controller)
+* [Derive class from ControllerBase](#derive-class-from-controllerbase)
+::: moniker-end
+::: moniker range=">= aspnetcore-2.1"
+* [Derive class from Controller](#derive-class-from-controller)
+* [Derive class from ControllerBase](#derive-class-from-controllerbase)
+* [Decorate class with ApiControllerAttribute](#decorate-class-with-apicontrollerattribute)
+::: moniker-end
+
+This document explains when it's most appropriate to use each option.
+
+## Derive class from Controller
+
+Inherit from the [Controller](/dotnet/api/microsoft.aspnetcore.mvc.controller) class when your controller needs to support MVC views in addition to Web API actions. For example:
+
+```csharp
+[Route("api/[controller]")]
+public class ProductsController : Controller
+{
+}
+```
+
+## Derive class from ControllerBase
+
+Inherit from the [ControllerBase](/dotnet/api/microsoft.aspnetcore.mvc.controllerbase) class when your controller doesn't need to support MVC views. For example:
+
+```csharp
+[Route("api/[controller]")]
+public class ProductsController : ControllerBase
+{
+}
+```
+
+::: moniker range=">= aspnetcore-2.1"
+## Decorate class with ApiControllerAttribute
+
+ASP.NET Core 2.1 introduces the `[ApiController]` attribute to denote a Web API controller class. For example:
+
+[!code-csharp[](../web-api/define-controller/samples/WebApiSample.Api/Controllers/ProductsController.cs?name=snippet_ControllerSignature&highlight=2)]
+
+The following sections describe convenience features added by the attribute.
+
+### Automatic HTTP 400 responses
+
+Validation errors automatically trigger an HTTP 400 response. Code such as the following becomes unnecessary:
+
+```csharp
+if (!ModelState.IsValid)
+{
+    return BadRequest(ModelState);
+}
+```
+
+This default behavior is disabled with the following code in *Startup.ConfigureServices*:
+
+[!code-csharp[](../web-api/define-controller/samples/WebApiSample.Api/Startup.cs?name=snippet_ConfigureApiBehaviorOptions&highlight=5)]
+
+### Binding source parameter inference
+
+Inference rules are applied for the default data sources of action parameters. The binding source attributes behave as follows:
+
+* [[FromBody]](/dotnet/api/microsoft.aspnetcore.mvc.frombodyattribute) is inferred for complex type parameters. An exception to this rule is any complex, built-in type with a special meaning, such as [IFormCollection](/dotnet/api/microsoft.aspnetcore.http.iformcollection) and [CancellationToken](/dotnet/api/system.threading.cancellationtoken). The binding source inference code ignores those special types. If you decide to explicitly apply the `[FromBody]` attribute, multiple occurrences of it in the same action results in an exception.
+* [[FromRoute]](/dotnet/api/microsoft.aspnetcore.mvc.fromrouteattribute) is inferred for any action parameter name matching a parameter in the route template. When multiple routes match an action parameter, any route value is considered `[FromRoute]`.
+* [[FromQuery]](/dotnet/api/microsoft.aspnetcore.mvc.fromqueryattribute) is inferred for anything else.
+
+For example, the `[FromBody]` attribute is implied for the `Product` parameter:
+
+[!code-csharp[](../web-api/define-controller/samples/WebApiSample.Api/Controllers/ProductsController.cs?name=snippet_CreateAsync)]
+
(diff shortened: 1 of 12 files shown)


aspnetcore/web-api/define-controller.md
```diff
@@ -0,0 +1,113 @@
+---
+title: Define a controller in ASP.NET Core Web API
```
rynowak: I have a bone to pick with this branding @DamianEdwards @danroth27  are we still using the Web API branding? If I'm wrong about this I'll just go away 😆

aspnetcore/web-api/define-controller.md
```diff
@@ -0,0 +1,113 @@
...
+::: moniker-end
+::: moniker range=">= aspnetcore-2.1"
+* [Derive class from Controller](#derive-class-from-controller)
+* [Derive class from ControllerBase](#derive-class-from-controllerbase)
+* [Annotate class with ApiControllerAttribute](#annotate-class-with-apicontrollerattribute)
+::: moniker-end
+
+This document explains when it's most appropriate to use each option.
+
+## Derive class from Controller
```
rynowak: The document seems focused on APIs, I think it's good to do a comparison, but explain that `Controller` is recommended for HTML and Razor.

To most users this won't make a big difference, but it's possible in the future we'd ship a remixed framework that doesn't include `Controller` by default.

aspnetcore/web-api/define-controller.md
```diff
@@ -0,0 +1,113 @@
...
+A benefit of deriving from `ControllerBase` instead of `Controller` is that IntelliSense displays only Web API-specific members.
+
+::: moniker range=">= aspnetcore-2.1"
+## Annotate class with ApiControllerAttribute
+
+ASP.NET Core 2.1 introduces the `[ApiController]` attribute to denote a Web API controller class. For example:
+
+[!code-csharp[](../web-api/define-controller/samples/WebApiSample.Api/Controllers/ProductsController.cs?name=snippet_ControllerSignature&highlight=2)]
+
+This attribute is commonly coupled with either `ControllerBase` or `Controller` to gain access to useful methods and properties. `ControllerBase` provides access to methods such as [CreatedAtAction](/dotnet/api/microsoft.aspnetcore.mvc.controllerbase.createdataction) and [File](/dotnet/api/microsoft.aspnetcore.mvc.controllerbase.file). `Controller` provides access to methods such as [Json](/dotnet/api/microsoft.aspnetcore.mvc.controller.json) and [View](/dotnet/api/microsoft.aspnetcore.mvc.controller.view).
```
rynowak: You can also create your own base class and slap `[ApiController]` on it. There's nothing magic about `Controller` or `ControllerBase`.

aspnetcore/web-api/define-controller.md
danroth27: We shouldn't use "ASP.NET Core Web API". Suggested title: "Building web APIs with ASP.NET Core"

aspnetcore/web-api/define-controller.md
```diff
@@ -0,0 +1,113 @@
...
+
+[!code-csharp[](../web-api/define-controller/samples/WebApiSample.Api.Pre21/Controllers/PetsController.cs?range=46-49)]
+
+This default behavior is disabled with the following code in *Startup.ConfigureServices*:
+
+[!code-csharp[](../web-api/define-controller/samples/WebApiSample.Api/Startup.cs?name=snippet_ConfigureApiBehaviorOptions&highlight=5)]
+
+### Binding source parameter inference
+
+Inference rules are applied for the default data sources of action parameters. The binding source attributes behave as follows:
```
rynowak: I think before digging into the details it would be good to show an example and explain the common patterns. Most APIs will bind a parameter (or two from the route) - usually an int or string, and a parameter or two from the query string - usually a simple type, and usually a  complex type from the body.

It would be good to show how the `[ApiController]` attribute cleans up the code by inferring the details of the idiomatic approach for binding in APIs

aspnetcore/web-api/define-controller.md
```diff
@@ -0,0 +1,113 @@
...
+
+* **[[FromBody]](/dotnet/api/microsoft.aspnetcore.mvc.frombodyattribute)** is inferred for complex type parameters. An exception to this rule is any complex, built-in type with a special meaning, such as [IFormCollection](/dotnet/api/microsoft.aspnetcore.http.iformcollection) and [CancellationToken](/dotnet/api/system.threading.cancellationtoken). The binding source inference code ignores those special types. When an action has more than one parameter explicitly specified (via `[FromBody]`) or inferred as bound from the request body, an exception is thrown. For example, the following action signatures cause an exception:
+
+[!code-csharp[](../web-api/define-controller/samples/WebApiSample.Api/Controllers/TestController.cs?name=snippet_ActionsCausingExceptions)]
+
+* **[[FromForm]](/dotnet/api/microsoft.aspnetcore.mvc.fromformattribute)** is inferred for action parameters of type [IFormFile](/dotnet/api/microsoft.aspnetcore.http.iformfile) and [IFormFileCollection](/dotnet/api/microsoft.aspnetcore.http.iformfilecollection).
```
rynowak: It would be good to add that `[FromForm]` won't be inferred for any simple types. If you want to bind form data in an api controller, you should tell us about it because it's not super common

aspnetcore/web-api/define-controller.md
Rick-Anderson: > You can also create your own base class and slap `[ApiController]` on it. There's nothing magic about `Controller` or `ControllerBase`.

Can you suggest why you'd want to do that?

aspnetcore/web-api/define-controller/samples/WebApiSample.Api.Pre21/Controllers/OrdersController.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            _repository = repository;
+        }
+
+        public IActionResult Index() => View();
+
+        [HttpGet("{id}")]
+        [ProducesResponseType(typeof(Order), 200)]
+        [ProducesResponseType(404)]
+        public IActionResult GetById(int id)
```
rynowak: This should use `ActionResult<Order>`

rynowak: That allows you to remove the `[ProducesResponseType]` attribute when it's a 200.

aspnetcore/web-api/define-controller.md
rynowak: Sure, lots of folks like to do one of two things:
1. Give us lots of specific feedback about specific action-result-creating methods they want to see or change
2. Create an `ApiController` **base class** because they like that idea

Users can solve both of these problems for themselves. They can create their own base class that either inherits from `ControllerBase`, slap `[ApiController]` on it and they are good to go.

aspnetcore/web-api/define-controller/samples/WebApiSample.Api.Pre21/Controllers/OrdersController.cs
scottaddie: This was done intentionally to show how things are accomplished prior to 2.1.

scottaddie:
@rynowak Please take another look.

aspnetcore/web-api/define-controller/samples/WebApiSample.Api.Pre21/Controllers/OrdersController.cs
scottaddie: This is a 2.0 version of the sample app, which is why it doesn't use ActionResult<T>.

aspnetcore/web-api/index.md
```diff
@@ -1,22 +1,135 @@
...
+* [Derive class from Controller](#derive-class-from-controller)
+* [Derive class from ControllerBase](#derive-class-from-controllerbase)
+* [Annotate class with ApiControllerAttribute](#annotate-class-with-apicontrollerattribute)
+::: moniker-end
+
+This document explains when it's most appropriate to use each option.
+
+## Derive class from Controller
+
+Inherit from the [Controller](/dotnet/api/microsoft.aspnetcore.mvc.controller) class in a controller that needs to support HTML and Razor in addition to web API actions. Examples that require `Controller` inheritance include returning MVC views or [invoking view components](xref:mvc/views/view-components#invoking-a-view-component-directly-from-a-controller).
```
rynowak: >in a controller that needs to support HTML and Razor in addition to web API actions. 

I don't really know what that's a topic. You certainly could write a controller that serves both APIs and views, but I don't think we'd recommend that users do this on purpose.

aspnetcore/web-api/index.md
```diff
@@ -1,22 +1,135 @@
...
+
+::: moniker range=">= aspnetcore-2.1"
+## Annotate class with ApiControllerAttribute
+
+ASP.NET Core 2.1 introduces the `[ApiController]` attribute to denote a web API controller class. For example:
+
+[!code-csharp[](../web-api/define-controller/samples/WebApiSample.Api/Controllers/ProductsController.cs?name=snippet_ControllerSignature&highlight=2)]
+
+This attribute is commonly coupled with either `ControllerBase` or `Controller` to gain access to useful methods and properties. `ControllerBase` provides access to methods such as [CreatedAtAction](/dotnet/api/microsoft.aspnetcore.mvc.controllerbase.createdataction) and [File](/dotnet/api/microsoft.aspnetcore.mvc.controllerbase.file). `Controller` provides access to methods such as [Json](/dotnet/api/microsoft.aspnetcore.mvc.controller.json) and [View](/dotnet/api/microsoft.aspnetcore.mvc.controller.view). Another approach is to create a custom base controller class annotated with the `[ApiController]` attribute.
+
```
rynowak: 👍

aspnetcore/web-api/index.md
rynowak: Regardless of what it used to derive from, I don't understand why we would discuss it as `a controller that needs to support HTML and Razor in addition to web API actions` - this isn't something we've ever really told people to do. It's explicitly recommended or disrecommended, which is why I'm suprised to see it getting coverage here.

scottaddie: @danroth27 Thoughts?

danroth27: :+1: to @rynowak's comment. I don't think we need to discuss deriving from `Controller` in this topic or anything views related.

PR #6012, opened 2018-04-18
Utilities.cs infinite loop

user135711:
Same infinite loop error in 2nd example. Also, three links titled reference source linking to core 2.0.1 are now dead. The first on the page works.


guardrex:
@user135711 ... at least I'm consistent! :smile: Thanks again for the fix.

I'll patch those 2.0.1 references right now. I'm going to see if I can eliminate them. We really don't want links that are hard to maintain. I probably put them in there because this API was new, and I probably felt that early readers of the topic would like to take a look. That's no longer the case, so I'll see about removing them. Thanks for the heads-up @user135711. :rocket:

PR #6025, opened 2018-04-18
Update nginx.conf

zokiad:
Performance issue. Add nodelay to avoid nginx to wait to be forward
https://www.nginx.com/blog/rate-limiting-nginx/


scottaddie:
@zokiad Thank you for your contribution!

PR #6026, opened 2018-04-19
Update jQuery Example per 6004

### LINKED ISSUE #6004: WebAPI web api "Call the Web API with jQuery" doesn't work
[**Call the Web API with jQuery**](https://docs.microsoft.com/en-us/aspnet/core/tutorials/first-web-api?hubRefSrc=email&utm_source=lfemail&utm_medium=email&utm_campaign=lfnotification&view=aspnetcore-2.1#call-the-web-api-with-jquery)

In this section, an HTML page is added that uses jQuery to call the Web API.

Per customers:

This is a great tutorial, up until the last part. It would be helpful if there were instructions on WHERE and HOW to add the HTML page. Also, the scripts on the given page don't work, they're apparently partially converted from what looks like maybe a product catalog app? Please either remove this section or beef it up so it actually works.
Edit
@Chris_Tavares I'm having the same problem with the HTML page.  I've got the Postman stuff running but can't figure out the HTML.

@isaac2004  any chance you could fix this? Otherwise I think I better delete it until we can get it fixed.
cc @scottaddie

isaac2004:
Fixes #6004

@scottaddie 

[Internal Review Page](https://review.docs.microsoft.com/en-us/aspnet/core/tutorials/first-web-api?view=aspnetcore-2.1&branch=pr-en-us-6026#call-the-web-api-with-jquery)


Rick-Anderson:
@isaac2004  thanks

PR #6045, opened 2018-04-21
Update live with current master

scottaddie:


user135711:
When does this go live? All the links are broken for me now with group1-dest **added** not removed to almost every TOC.

scottaddie:
@user135711 We apologize for the inconvenience. Our engineers are working on it. As you've pointed out, removing "group1-dest/" from the URL is the workaround.

/cc: @dend - I opened an LSI last night for this problem.

PR #6054, opened 2018-04-23
Update index.md

Jaxelr:
Updated myMiddleware index.md to use the code syntax highlight.

Noticed that on the asp.net page one code block wasnt highlighting (line 240). 

image

Noticed that the other blocks were using the keyword csharp instead of c# (example of line 122)

image


Rick-Anderson:
@Jaxelr  once again, thanks!

PR #6077, opened 2018-04-25
Typo on required configuration switch mappings

psimoneau22:
Typo on required configuration switch mappings


scottaddie:
@psimoneau22 Thank you for catching this mistake and correcting it!

PR #6086, opened 2018-04-25
Document improved support for integration testing

### LINKED ISSUE #4652: Document improved support for in memory end to end testing of MVC apps
We need to cover:
* Setting up a project for testing an MVC application in memory using TestServer and the new Microsoft.AspNetCore.Mvc.Testing package.
* How to customize the IWebHostBuilder instance from the app for testing.
    * How to override configured services on the application for testing. (e.g the database).
* How to configure where to find views and static files.

### LINKED ISSUE #6048: RP Integration testing content updates
* Remove integration testing content from RP testing topic
* Add API topic links to new topic

### LINKED ISSUE #6467: Drop 2.0 content in integration test topic
*Integration tests in ASP.NET Core*
https://docs.microsoft.com/aspnet/core/test/integration-tests
aspnetcore\test\integration-tests.md

Drop 2.0 content and moniker-ranges. Discussed on #6086.

guardrex:
Fixes #4652 
Fixes #6467
Fixes #6048 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-1.0&branch=pr-en-us-6086)

### Notes

* Makes a UE pass on the original topic.
* In our quest to move away from gerunds, changing *testing* to *tests*.
* This isn't the *only* way to organize the content, but it is one decent approach. The outline is below.
* I show (and include in the sample app) *per-test class config* with `WithWebHostBuilder` **--AND--** *custom* `WebApplicationFactory`. I think they should both be shown. If we leave either out, then I think we're bound to upset one group of devs or the other.
* We use xUnit ... the topic is xUnit-ish ... the sample uses xUnit. We may need to add more on the use of other test frameworks. It's a management call (i.e., @Rick-Anderson) to say if we need more (or how far to take that). Let's assess this on review.
* I'm a bit concerned that we don't show alternate configuration (e.g., app settings) here. It's not a big deal to add it. We have the test web host config covered well here; so in theory for any sort of custom test web host config the dev needs, they'll be able to work it out from here. However, I have a sneaky feeling that we'll get feedback along the lines of: "How do I use a different app settings file?" Let's assess this on review as well.
* Not much to say about RP vs. MVC in this context. What I had to say ... well ... *I said IT!* :smile: lol Let's see if that covers the bases. If not, we'll add more content that goes beyond test class naming.

### Draft topic outline

* Integration tests in ASP.NET Core
  * Introduction to integration tests (general info on the basics)
  * Integration tests in ASP.NET Core (more focused specifically on ASP.NET Core testing)
  * Customize the web host and create a client
    * Per-test class configuration
    * Custom WebApplicationFactory implementations
  * Client options
  * How the test infrastructure infers the app content root path
  * Disable shadow copying (xUnit only)
  * Integration tests sample
    * Message app (SUT) organization
    * Test app organization
    * Integration tests of the SUT
  * Additional resources
  
btw - @scottaddie I guess it's ok that the build is throwing warnings for the duplicate UID's (original topic as <= 2.0 and new topic for >= 2.1) ... that's ok?


guardrex:
> You'll need to change the uid of the new file to something like test/integration-tests. 

... but the new file is already `test/integration-tests`. That's what both topics are set to.

If you're saying they must be different, that's confusing: If they should have different UID's, then how does it know that the two files are related (to swap in one or the other's content) if they don't share the UID? I mean they already don't share the file name. There's nothing else to link them if both the UID and the file names are different.

guardrex:
@javiercn I marked this **WIP** as we work out the versioning angle. Please don't let that hold you up on reviewing the PR tho. We're just going to move a few files/folders around and change some metadata to get this pair of topics hosted properly. The content of the new topic isn't at stake with these changes, so you can look at this whenever you want.

guardrex:
@scottaddie Checking in with you: Any new info yet on some sort of "Version by File" approach for this (this is the PR where there's an old topic that should load for <=2.0 and a new topic that should load for >=2.1)?

If it looks like we'll need two topics in the TOC (for a while anyway), I'm fairly certain I can set that up for the PR with some careful titling, cross-linking, and TOC entries.

guardrex:
@javiercn I was the caller who asked on the Standup radio show 📻 for you to clarify "integration testing" versus "functional testing."

I agree with the original author of the topic that these are "integration tests" ... but yes, indeed ... they are of the E2E variety of integration tests. As you said, some purists insist that user components in the test process (e.g., browser, network) can't be factored out of a true "functional test." After all, browsers and networks might not work the same way as their mocks/replacements. I think (guessing) that's why the original author of the topic provided this content the way that he did.

However, we might still want to address the difference between integration testing and functional testing in the topic (or someplace else) ... **_or not_** lol. It's beyond my pay grade to decide! :smile: lol

cc/ @danroth27

Rick-Anderson:
@scottaddie  Do you have a solution for @guardrex  question 
> Any new info yet on some sort of "Version by File" approach f

For now perhaps make two SxS topics in the TOC for these two topics. When 2.1 RTWs delete the 2.0 topic.

guardrex:
We can also set this up that way now (drop the 2.0 topic in this PR) and sit on this ... mark this with "[HOLD]" in the title and give it a "blocked" label. Merge the PR when 2.1 lands. That would be the most efficient approach.

guardrex:
Actually since you're saying that we can drop the 2.0 stuff at 2.1 RTM, then let's go ahead with one topic and use large moniker-range blocks to hold everything. The only reason that wasn't an option before is because we were talking about maintaining the 2.0 content and moniker-range blocks can't be nested. Now that I know we're going to drop that old content, we can do this in one topic.

Marking WIP for just a bit until I can set it up.

guardrex:
From @javiercn ...

1. I would show how to setup a basic test before I go into how to customize the Web Application Factory.
1. The recommendation is to use CreateClient instead of CreateDefaultClient (I know the name is unfortunate). The reason is that CreateClient does cookies and redirects by default.
1. I would start by showing how to customize the factory by extending WebApplicationFactory. The pattern is that you typically extend web application factory to customize setup for the whole set of integration tests or for the tests in a class (when you have several) and use WithWebHostBuilder for one off situations.
1. Disabling shadow copying is required in general, not just for xUnit. (It's a framework dependent setting)
1. If you need to showcase how to test things that include HTML. Can you show it how to do it in the same way I did for my final demo on live.asp.net? Be explicit about anglesharp being a 3rd party library and not part of ASP.NET Core. You can reuse your code and sample, just make sure you follow the same pattern and use CreateClient instead of CreateDefaultClient (which will handle cookies automatically for you). https://github.com/javiercn/testingdemos/blob/master/test/SampleApplication.Functionaltests/BasicTests.cs#L76

guardrex:
@javiercn Let's discuss a few points ...

1. The problem with having another section that shows an even more basic test (i.e., a test without the dB context ceremony, for example) is that it breaks the outline. The outline for this node goes like this ...

[Note: The example is the same, so it's easy to compare the approaches SxS.]

* Customize the web host and create a client (define and describe; list the two approaches)
  * Approach 1 with example
  * Approach 2 with example

If we interject another (simpler) example, it's going to break that node and confuse the SxS layout. Either the new example will be part of Approach 1 or Approach 2, but then we'll have two examples for one and not the other. It will be an outlier in a SxS comparison.

I think whether the per-test class config approach or the `WebApplicationFactory` config approach appears first that devs can see how the basic set up works.

2. :+1: Gotcha ... I'll update it on the next commit, where I'm also going to sort out our content location problem.

3. Are you asking me to reverse the coverage of `WebApplicationFactory` and per-test class configuration so that the `WebApplicationFactory` approach is pushed as the main approach? If so, I can also modify the text in order to convey that the `WebApplicationFactory` approach is the primary configuration approach.

4. I don't follow. Is this inaccurate? ...

> xUnit shadow copying causes the tests to execute in a different folder than the output folder. For tests to work properly, shadow copying must be disabled.

I'm also confused because you said, "required in general, not just for xUnit," but shadow copying is disabled via an *xunit.runner.json* file ... that isn't xUnit-specific?

5. Let's discuss on a separate issue, which will also allow it to be worked separately. This topic provides enough content to configure the test app for basic integration testing. I like what you're suggesting, but I'd prefer someone work it as a follow-up.

6. I imagine auth testing is a complex advanced subject and not suitable for this introductory-level topic. Since that would probably be an advanced topic, I think it should be opened in a new issue.

scottaddie:
@Rick-Anderson @guardrex I haven't had time to look at the versioning issue yet. I plan to set aside some time tomorrow morning to do so.

guardrex:
@scottaddie We're good now. I didn't know we could drop the 2.0 content at 2.1 RTM. What I just did was put everything in the *one* topic with moniker-ranges. When 2.1 hits, we can strip out the 2.0 bits and the moniker-ranges. I think we're good here. I'll open an issue to make sure that work is done.

javiercn:
> I'm also confused because you said, "required in general, not just for xUnit," but shadow copying is disabled via an xunit.runner.json file ... that isn't xUnit-specific?

Shadow copying is a framework feature.
xunit.runner.json is the way to disable it for xUnit. MSTest and other testing frameworks, either don't use it or have a different mechanism to disable them.

guardrex:
@javiercn I addressed item 2.

RE: Item 4 ...

> MSTest and other testing frameworks, either don't use it or have a different mechanism to disable them.

My current understanding (perhaps incorrect) is that other frameworks don't require disabling shadow copying. I need additional guidance on what should be said in this section. What do you want this section on disabling shadow copying to say?

Item 1 is still up for discussion. I'm concerned about breaking the outline with an additional sample there.

Item 3 Do you want me to move up and emphasize the `WebApplicationFactory` config approach and down-play the per-text class config approach?

Note that I've rolled the content together now in one topic :point_right: [Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-1.0&branch=pr-en-us-6086)

javiercn:
> Let's discuss on a separate issue, which will also allow it to be worked separately. This topic provides enough content to configure the test app for basic integration testing. I like what you're suggesting, but I'd prefer someone work it as a follow-up.

Handling the cookie and using regex to get the Antiforgery token from the html response is a pit of failure (i'm talking from experience) and is not the path we want our customers to follow. You can add anglesharp to the testing project. Copy these couple of helpers verbatim https://github.com/javiercn/testingdemos/tree/master/test/SampleApplication.Functionaltests/Helpers
and use the client produced by CreateClient to implement that test without having to worry about all the  HTML/HTTP specific details.

> I imagine auth testing is a complex advanced subject and not suitable for this introductory-level topic. Since that would probably be an advanced topic, I think it should be opened in a new issue.

The only thing I'm asking here is to have a point specifically for routing and another one specifically for authorization. The tests can be as simple as
https://github.com/javiercn/testingdemos/blob/master/test/SampleApplication.Functionaltests/BasicTests.cs#L44
https://github.com/javiercn/testingdemos/blob/master/test/SampleApplication.Functionaltests/BasicTests.cs#L58

With the goal of being explicit about the scenarios where these type of tests are a much better option than unit tests.

> Are you asking me to reverse the coverage of WebApplicationFactory and per-test class configuration so that the WebApplicationFactory approach is pushed as the main approach? If so, I can also modify the text in order to convey that the WebApplicationFactory approach is the primary configuration approach.

I'm saying use WebApplicationFactory as a starting point (no customization, runs in the same way as when you F5 in Development).
When you want to "productionize" your tests and need to adapt things for them:
* Extend WebApplicationFactory to apply changes that you want for all your tests.
* Use this same approach if all the tests in a class require some additional customization.
When the customization is unique to a single test, use WithWebHostBuilder()

guardrex:
`@danroth27` said to make sure that there's no indication that AngleSharp is supported,  recommended, or required.

danroth27:
> @danroth27 said to make sure that there's no indication that AngleSharp is supported, recommended, or required.

Yes, but we are also fine to use AngleSharp as an example approach. I would prefer to show AngleSharp as an example approach instead of showing using regex's.

guardrex:
I am ... doing it now.

guardrex:
Thus far ...

https://github.com/guardrex/IntegrationTestingSample/tree/master/tests/RazorPagesProject.Tests

* There's a button to delete a single message that I trigger by looking at its `formaction` value. We might decide to `id` those **Delete** message buttons and call for one of them by `id`.
* I don't show additional `_client` config in this batch of tests with `WithWebHostBuilder` yet. I plan to see if I can work out a good example tomorrow.
* No Identity in the SUT yet. I'll add that tomorrow and an Identity test point.

guardrex:
@javiercn Having a problem with my custom `WebApplicationFactory`. Goes like this ...

```csharp
public class CustomWebApplicationFactory<TStartup> : WebApplicationFactory<RazorPagesProject.Startup>
{
    {
        builder.ConfigureServices(services =>
        {
            var serviceProvider = new ServiceCollection()
                .AddEntityFrameworkInMemoryDatabase()
                .BuildServiceProvider();

            services.AddDbContext<ApplicationDbContext>(options => 
                {
                    options.UseInMemoryDatabase("InMemoryDbForTests");
                    options.UseInternalServiceProvider(serviceProvider);
                });

            var sp = services.BuildServiceProvider();

            using (var scope = sp.CreateScope())
            {
                var scopedServices = scope.ServiceProvider;
                var db = scopedServices.GetRequiredService<ApplicationDbContext>();
                var logger = scopedServices
                    .GetRequiredService<ILogger<CustomWebApplicationFactory<TStartup>>>();

                db.Database.EnsureCreated();

                try
                {
                    Utilities.InitializeDbForTests(db);
                }
                catch (Exception ex)
                {
                    logger.LogError(ex, $"An error occurred seeding the " +
                        "database with test messages. Error: {ex.Message}");
                }
            }
        });
    }
}
```

This works (dB is seeded correctly) ... 

```csharp
public IndexPageTests(CustomWebApplicationFactory<RazorPagesProject.Startup> factory)
{
    Client = factory.CreateDefaultClient();
}
```

Calling `CreateClient` doesn't work (`InitializeDbForTests` runs but the dB is **not** seeded) ...

```csharp
public IndexPageTests(CustomWebApplicationFactory<RazorPagesProject.Startup> factory)
{
    Client = factory.CreateClient(new WebApplicationFactoryClientOptions
    {
        AllowAutoRedirect = false
    });
}
```

Can u see where I'm going wrong getting the dB seeded with `CreateClient`?

Couple of other quick ?:

In test classes, why is `HttpClient` always set up as a public prop and not a private read-only field?

For string comparisons with the `Location` header, why is `response.Headers.Location.ToString()` always favored over `response.Headers.Location.AbsoluteUri`?

javiercn:
> In test classes, why is HttpClient always set up as a public prop and not a private read-only field?

It's a matter of style. I usually just `ctrl+.` those things from within the class constructor.



> Same question on factories. You want the WebApplicationFactory<RazorPagesProject.Startup> as a public prop in test classes

Same as for 1.



> For string comparisons with the Location header, why is response.Headers.Location.ToString() always favored over response.Headers.Location.AbsoluteUri

I didn't specifically paid attention to that, but I believe the location can be relative, so not sure if AbsoluteUri will work.

javiercn:
> @javiercn Having a problem with my custom WebApplicationFactory. Goes like this ...

Were you able to solve this? There's no difference between calling Createclient or CreateDefaultClient regarding how the server gets created/initialized. Could it be by how you initialize the Database? This is how I've done in other projects.

https://github.com/aspnet/Identity/blob/dev/test/Service.FunctionalTests/Infrastructure/EntityFrameworkSeedReferenceData.cs#L14

https://github.com/aspnet/Identity/blob/dev/test/Identity.FunctionalTests/Infrastructure/FunctionalTestsServiceCollectionExtensions.cs#L20

guardrex:
> I didn't specifically paid attention to that, but I believe the location can be relative, so not sure if AbsoluteUri will work.

`AbsoluteUri` feels safer (more specific), since the `Location` header value could be relative or absolute and that property would (should 🤞) get set appropriately. Perhaps, `ToString` does the same thing ... I just don't trust it as much.

WRT fields vs. props ... These aren't exposed outside of these classes, so my inclination would be to make them fields. I can make them properties if you like. I thought perhaps something in the testing infrastructure may require it. Since that's not the case, you can decide. If you don't care either way, then I'll make them fields.

> This is how I've done in other projects.

Cool. Thx. I'll take a look at those projects and see what's going wrong here. It's kind'a fishy tho that `CreateDefaultClient` seeds it but `CreateClient` doesn't. I'll troubleshoot it a bit more and see what's what.

javiercn:
> WRT fields vs. props ... These aren't exposed outside of these classes, so my inclination would be to make them fields. I can make them properties if you like. I thought perhaps something in the testing infrastructure may require it. Since that's not the case, you can decide. If you don't care either way, then I'll make them fields.

I'm fine with them being `private readonly fields`

javiercn:
> AbsoluteUri feels safer (more specific), since the Location header value could be relative or absolute and that property would (should 🤞) get set appropriately. Perhaps, ToString does the same thing ... I just don't trust it as much

Seems like AbsoluteUri will fail if the URI is relative. I would use ToString or new Uri(expectedUri) for the comparison.
https://github.com/dotnet/corefx/blob/master/src/System.Private.Uri/src/System/Uri.cs#L686

I usually ToString it because it's the simplest way to obtain a string representation for the Uri

guardrex:
The 302 redirect for an unauthorized user attempting to reach the Secure Page of the app provides a `Location` header value of `https://localhost:5001/Identity/Account/Login?ReturnUrl=%2FSecurePage`. Since it's absolute and only compared up to "Login," it just so happened to be safe either way.

I see what you mean about `AbsoluteUri`. I studied it a bit more ...

`Location: /Contact`
* `ToString`: `/Contact`
* `OriginalString`: `/Contact`
* `AbsoluteUri`: Exception

`Location: https://localhost:5001/Contact`
* `ToString`: `https://localhost:5001/Contact`
* `OriginalString`: `https://localhost:5001/Contact`
* `AbsoluteUri`: `https://localhost:5001/Contact`

`Location: https://localhost:5001/Contact%20Us`
* `ToString`: `https://localhost:5001/Contact Us`
* `OriginalString`: `https://localhost:5001/Contact%20Us`
* `AbsoluteUri`: `https://localhost:5001/Contact%20Us`

I think this explains why Steve used `OriginalString` in the sample. I guess he wanted the *rawest* **safe** version of the `Location` header value for comparisons.

guardrex:
The difference seems to be `HandleCookies` ...

When `HandleCookies = true`, the dB is **not** seeded.
When `HandleCookies = false`, the dB **is** seeded.

... which explains why `CreateDefaultClient` works and `CreateClient` doesn't work (when handling cookies).

I can't inspect this and divine any bearing on the dB context: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Testing/Handlers/CookieContainerHandler.cs

The same seeding procedure works with `WithWebHostBuilder` in the test class, but that's not what I'm trying to do. I'd like the dB seeded in the `CustomWebApplicationFactory`.

btw - Tried it with the `IStartupFilter` approach you linked ... same behavior.

The latest sample app is here :point_right: https://github.com/guardrex/IntegrationTestingSample

javiercn:
@guardrex I'll clone the repo and take a look at it.

javiercn:
@guardrex It's this lines.

https://github.com/guardrex/IntegrationTestingSample/blob/7cab119ec5bfc47963c6843349a5af2a74764750/tests/RazorPagesProject.Tests/IntegrationTests/IndexPageTests.cs#L43-L46.

You don't have to handle the request verification token. the helper does that for you.

javiercn:
@guardrex Also the test doesn't run properly when running among other tests, but runs ok in isolation, so I believe you might be experiencing a test isolation problem.

guardrex:
I realize now that this issue came up with the unit testing sample. In that situation, I supply a `DbContextOptions` instance that's based on a new service provider and then use it for each unit test ...

```csharp
public static DbContextOptions<AppDbContext> TestingDbContextOptions()
{
    // Create a new service provider to create a new in-memory database.
    var serviceProvider = new ServiceCollection()
        .AddEntityFrameworkInMemoryDatabase()
        .BuildServiceProvider();

    // Create a new options instance using an in-memory database and 
    // IServiceProvider that the context should resolve all of its 
    // services from.
    var builder = new DbContextOptionsBuilder<AppDbContext>()
        .UseInMemoryDatabase("InMemoryDb")
        .UseInternalServiceProvider(serviceProvider);

    return builder.Options;
}

...

using (var db = new AppDbContext(Utilities.TestingDbContextOptions()))
{
    // Use the db here in the unit test.
}
```

That was straight-forward for unit testing. Thus far, we have the one factory and one client for the entire group of tests.

A possible approach for this one test is to seed the dB with a new record in the Arrange step using a new client and with a call to `WithWebHostBuilder` ... then delete the record in the Act step and confirm it for Assert.

@javiercn How do you want to play this? What advice do you want to give devs about database seeding and configuring reliable tests that don't step on each other (atomic tests)?

javiercn:
@guardrex Perform the appropiate setup on each test as part of the arrange. For example, if you are deleting an item, you first create the item as part of your arrange step. That's what we do in the tests in identity.

Another approach is to use the seed data capabilities of EF (it's a 2.1 feature)

javiercn:
image

This is what I was seeing.

guardrex:
> This is what I was seeing.

Yes, I believe this is just being caused by the "delete all" test ... it's hosing the data for this test that deletes one record.

> Perform the appropriate setup on each test as part of the arrange. 

:+1: I'll do that next for this test. ... and I'll make sure this point is covered in the topic when I swing back over to update it after the sample is happy.

guardrex:
All tests passing now. Let me know if you see anything else you would like to change.

The latest updates that show how the "delete one" test uses it's own client and seeding are here :point_right: https://github.com/guardrex/IntegrationTestingSample/blob/master/tests/RazorPagesProject.Tests/IntegrationTests/IndexPageTests.cs#L54-L101

Wednesday, I'll flip back over to the topic and update with all the new bits.

javiercn:
@guardrex Can you place a link to the updated review topic? I'm trying to read the completed doc.

guardrex:
@javiercn Oh ... haven't reached that point yet. That will be tomorrow morning. All we have at this point is the updated sample. I'll get this done tomorrow and ping u back ... probably tomorrow afternoon.

guardrex:
This is going to take another day ... there are a *LOT* of updates 😅. Even if I finish this evening, I need to sleep on the text and make an edit pass in the morning. Look for a ping on Thursday.

guardrex:
@javiercn Still a bit on the *raw* side, since I was making quite a few content changes today. I'll pass it over again Friday for more polish. In the meantime tho, the main updates are there.

* Note: `IClassFixture` isn't in the API browser for some reason: https://docs.microsoft.com/en-us/dotnet/api/?term=IClassFixture
* Walking a fine line here on helper method details ... trying to strike the right balance between not-too-much and not-too-little content.
* Follows your suggested sequence:
   1. Basic tests (OOB `WebApplicationFactory` without customization)
   1. `CustomWebApplicationFactory` class used in a test class
   1. `WithWebHostBuilder` example for further customization
* Added your suggestions for basic tests of:
   1. Routing
   1. Secure page
   1. HTML work, but the GH user bit is only explained lightly. Topic is getting really *L .. O .. N .. G*. I hope this will be enough, but we can expand if you like.
* Updated the shadow copying bit to make it more generic (but it retains the specifics for xUnit).

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-1.0&branch=pr-en-us-6086)

javiercn:
> IClassFixture

This is an xUnit feature

javiercn:
@guardrex I can't see the internal review topic. @Rick-Anderson can you help here?

guardrex:
@javiercn So this link is failing ...

https://review.docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-1.0&branch=pr-en-us-6086

guardrex:
@Sara-Nicolini FYI: This PR moves the "testing" folder topics to a new "test" folder in the TOC.

Rick-Anderson:
> @guardrex I can't see the internal review topic. @Rick-Anderson can you help here?

@javiercn  can you have a couple of your colleagues try the [link](https://review.docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-1.0&branch=pr-en-us-6086)?

guardrex:
@Rick-Anderson When you get to this one, the main topic is `test/integration-tests.md`; so when you flip to the diff (and after everything loads), search for "tests.md" ... that should scroll it right up for review.

guardrex:
That is the main thing. If you wait for the diff to load and do a browser search for "tests.md," it should scroll up into view.

capture

Rick-Anderson:
@guardrex  I'll do a more through review in a few days.

PR #6089, opened 2018-04-26
Update index.md

harleydk:
Added further detail to the BuildWebHost() function, to clarify the requirement of having both the UseKestrel() and UseIISIntegration() statements. Would've helped me out, maybe others.


guardrex:
Those are already added by `CreateDefaultBuilder` ...

https://github.com/aspnet/MetaPackages/blob/dev/src/Microsoft.AspNetCore/WebHost.cs#L153

https://github.com/aspnet/MetaPackages/blob/dev/src/Microsoft.AspNetCore/WebHost.cs#L206

If you're having trouble with app config/startup, devs on Slack can probably help. The aspnetcore Slack Team room self-signup is here :point_right: http://tattoocoder.com/aspnet-slack-sign-up/

PR #6094, opened 2018-04-26
Change jquery onready to IIFE

user135711:
This fixes errors that work here but may cause problems if copied into other projects. The function is called by jquery on document ready which will fail if the validator has already run which also runs on ready. There is also a mistake in passing the jquery parameter.


scottaddie:
@user135711 Thank you for cleaning up this JS! I converted to the shortened form of jQuery's `document.ready` function, which is what this sample was intended to use.

PR #6102, opened 2018-04-27
New kestrel configuration & related APIs

### LINKED ISSUE #5056: New kestrel configuration & related APIs
Kestrel now has first class support for loading endpoints via config. See https://github.com/aspnet/KestrelHttpServer/pull/2186

### LINKED ISSUE #6129: Document Kestrel support for SNI
> Server Name Indication (SNI) can be used to allow hosting multiple domains on the same IP address and port. It does this by sending the expected host name in the TLS handshake so that the server can provide the correct certificate. Kestrel now supports this via the ServerCertificateSelector callback. This is invoked once per connection to allow you to inspect the host name and select the most appropriate certificate.

> SNI support requires running on netcoreapp2.1. On netcoreapp2.0 and net461 the callback will be invoked but the name will always be null. The name will also be null if the client did not provide this optional parameter.

```csharp
WebHost.CreateDefaultBuilder()
    .UseKestrel((context, options) =>
    {
        options.ListenAnyIP(5005, listenOptions =>
        {
            listenOptions.UseHttps(httpsOptions =>
            {
                var localhostCert = CertificateLoader.LoadFromStoreCert("localhost", "My", StoreLocation.CurrentUser, allowInvalid: true);
                var exampleCert = CertificateLoader.LoadFromStoreCert("example.com", "My", StoreLocation.CurrentUser, allowInvalid: true);
                var subExampleCert = CertificateLoader.LoadFromStoreCert("sub.example.com", "My", StoreLocation.CurrentUser, allowInvalid: true);
                var certs = new Dictionary(StringComparer.OrdinalIgnoreCase);
                certs["localhost"] = localhostCert;
                certs["example.com"] = exampleCert;
                certs["sub.example.com"] = subExampleCert;

                httpsOptions.ServerCertificateSelector = (features, name) =>
                {
                    if (name != null && certs.TryGetValue(name, out var cert))
                    {
                        return cert;
                    }

                    return exampleCert;
                };
            });
        });
    });
```

guardrex:
Fixes #5056 
Fixes #6129

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel?view=aspnetcore-1.0&branch=pr-en-us-6102)

Much of this was derived from https://github.com/aspnet/KestrelHttpServer/pull/2186. However, I made a big pass on the whole topic, too, in an attempt to polish it up (e.g., API topic links, xref links, code fencing, lists, grammar+style, and sample touch-ups).


aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,443 @@ uid: fundamentals/servers/kestrel
...
+> If not using a reverse proxy with host filtering enabled, [host filtering](#host-filtering) must be enabled.
 
 ## How to use Kestrel in ASP.NET Core apps
 
 #### [ASP.NET Core 2.x](#tab/aspnetcore2x/)
+
 The [Microsoft.AspNetCore.Server.Kestrel](https://www.nuget.org/packages/Microsoft.AspNetCore.Server.Kestrel/) package is included in the [Microsoft.AspNetCore.All metapackage](xref:fundamentals/metapackage).
 
-ASP.NET Core project templates use Kestrel by default. In *Program.cs*, the template code calls `CreateDefaultBuilder`, which calls [UseKestrel](https://docs.microsoft.com/aspnet/core/api/microsoft.aspnetcore.hosting.webhostbuilderkestrelextensions#Microsoft_AspNetCore_Hosting_WebHostBuilderKestrelExtensions_UseKestrel_Microsoft_AspNetCore_Hosting_IWebHostBuilder_) behind the scenes.
+ASP.NET Core project templates use Kestrel by default. In *Program.cs*, the template code calls [CreateDefaultBuilder](/dotnet/api/microsoft.aspnetcore.webhost.createdefaultbuilder), which calls [UseKestrel](/dotnet/api/microsoft.aspnetcore.hosting.webhostbuilderkestrelextensions.usekestrel) behind the scenes.
```
Tratcher: Why even show an example that calls UseKestrel in this scenario?

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,443 @@ uid: fundamentals/servers/kestrel
...
+* `--urls` command-line argument
+* `urls` host configuration key
+* `ASPNETCORE_URLS` environment variable
 
-You can configure endpoints by calling the `UseUrls` method or using the `urls` command-line argument or the ASPNETCORE_URLS environment variable. These methods are useful if you want your code to work with servers other than Kestrel. However, be aware of these limitations:
+These methods are useful for making code work with servers other than Kestrel. However, be aware of these limitations:
 
-* You can't use SSL with these methods.
-* If you use both the `Listen` method and `UseUrls`, the `Listen` endpoints override the `UseUrls` endpoints.
+* SSL can't be used with these approaches.
```
Tratcher: Except... SSL/HTTPS now works with these scenarios if you provide a default certificate through one of the new configuration mechanisms (e.g. config file, ConfgiureHttpsDefaults, etc).

aspnetcore/fundamentals/servers/kestrel.md
guardrex: At this point, it doesn't look like (Tom?) was trying to show that ... I think he's going for an example of the ...

> ASP.NET Core project templates use Kestrel by default ... `CreateDefaultBuilder` ... calls `UseKestrel`

... bit. It looks like he's just making reference to the `CreateDefaultBuilder` line at this point.

Then, it looks like he wanted to show the options config, either to show the `ListenOptions` and/or a cert being used ... or perhaps just for some context early in the topic.

I'm just guessing, of course.

Would you like me to strike the `UseKestrel`?

I should mention that when 2.1 ships that this topic is going to get a *green dino samples treatment*: https://github.com/aspnet/Docs/issues/5495. For now, I'm just trying to work with what we have as best I can. We aren't going to have 2.0 and 2.1 samples ... only 2.x ... and 2.x = 2.0 now and 2.x = 2.1 at 2.1 RTM.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -197,7 +197,7 @@ ASP.NET Core 2.1 and later project templates configure apps to run on HTTPS by d
 
 Call [Listen](/dotnet/api/microsoft.aspnetcore.server.kestrel.core.kestrelserveroptions.listen) or [ListenUnixSocket](/dotnet/api/microsoft.aspnetcore.server.kestrel.core.kestrelserveroptions.listenunixsocket) methods on [KestrelServerOptions](/dotnet/api/microsoft.aspnetcore.server.kestrel.core.kestrelserveroptions) to configure URL prefixes and ports for Kestrel.
 
-`UseUrls`, the `--urls` command-line argument, `urls` host configuration key, and the `ASPNETCORE_URLS` environment variable also work but have the limitations noted later in this section (a default certificate must be available for HTTPS endpoint configuration).
+`UseUrls`, the `--urls` command-line argument, `urls` host configuration key, and the `ASPNETCORE_URLS` environment variable also work but have a limitations noted later in this section (a default certificate must be available for HTTPS endpoint configuration).
```
halter73: Nit: limitation instead of limitations

guardrex: It will work out to "the limitations" (plural), so I'm reverting. There will be two (unless we change that). I reacted to Chris's remark on the last commit. It's down at lines 407-408.

aspnetcore/fundamentals/servers/kestrel.md
Tratcher: Removing UseKestrel would be more clear here, showing that the app can run without it.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,443 @@ uid: fundamentals/servers/kestrel
...
 
-[!code-csharp[](kestrel/sample2/Program.cs?name=snippet_UnixSocket)]
+[!code-csharp[](kestrel/samples/2.x/Program.cs?name=snippet_UnixSocket)]
 
 **Port 0**
 
-If you specify port number 0, Kestrel dynamically binds to an available port. The following example shows how to determine which port Kestrel actually bound to at runtime:
+When the port number is `0` is specified, Kestrel dynamically binds to an available port. The following example shows how to determine which port Kestrel actually bound at runtime:
+
+[!code-csharp[](kestrel/samples/2.x/Startup.cs?name=snippet_Configure&highlight=3-4,17-20)]
```
Tratcher: This sample is pretty ironic because you wouldn't be able to connect to the site without knowing the port. The real port will be reported in the console output "Now listening on:". That might be better to show here.

guardrex:
> Something seems wrong with your review link, it's not showing most of the new content like the json config files.

I think it's the conceptual versioning. I don't have access to the internal review topics, and they don't build locally correctly. @Rick-Anderson and/or @scottaddie will need to help with layout if versioning isn't working well. I'm only aware of bad lists and bad code blocks (e.g., https://github.com/aspnet/Docs/issues/6052) ... not sure about content not showing at all.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,443 @@ uid: fundamentals/servers/kestrel
...
+
+*No configuration*
+
+Kestrel listens on `http://localhost:5000` and `https://localhost:5001` (if a default cert is available).
+
+Specify URLs using the:
+
+* `ASPNETCORE_URLS` environment variable.
+* `--urls` command-line argument.
+* `urls` host configuration key.
```
Tratcher: it's important to distinguish that this must come from the host configuration, not the app configuration. There's an FAQ where people try to set Urls in appsettings.json and it doesn't work because appsettings.json isn't read until after the host is initialized. https://github.com/aspnet/MetaPackages/issues/247.  Kestrel's new config schema doesn't have this limitation, it reads directly from the config much later in the startup process.

aspnetcore/fundamentals/servers/kestrel.md
guardrex: The port is never set to `0` anyway in the sample. I'll add a demo of it and the console window output.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,443 @@ uid: fundamentals/servers/kestrel
...
+* `UseUrls` extension method.
+
+For more information, see [Server URLs](xref:fundamentals/hosting#server-urls) and [Overriding configuration](xref:fundamentals/hosting#overriding-configuration).
+
+The value provided using these approaches can be one or more HTTP and HTTPS endpoints (HTTPS if a default cert is available). Configure the value as a semicolon-separated list (for example, `"Urls": "http://localhost:8000;http://localhost:8001"`).
+
+*Replace the default certificate from configuration*
+
+Call `serverOptions.Configure(context.Configuration.GetSection("Kestrel"))` to specify the configuration section. A default HTTPS app settings configuration schema is available for Kestrel. Configure multiple endpoints, including the URLs and the certificates to use, either from a file on disk or from a certificate store.
```
Tratcher: This Configure call is included by default in WebHost.CreateDefaultBuilder.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,443 @@ uid: fundamentals/servers/kestrel
...
+
+In the following *appSettings.json* example:
+
+* Set **AllowInvalid** to `true` to permit the use of invalid certificates (for example, self-signed certificates).
+* **HttpsDefaultCert** indicates use of the cert defined under **Certificates** > **Default** or the development certificate.
```
Tratcher: This sounds like the name HttpsDefaultCert is special. Any https endpoint that does not specify a certificate will fall back to the the default specified the the Certificates section or elsewhere in code.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,443 @@ uid: fundamentals/servers/kestrel
...
+      "Default": {
+        "Path": "<path to .pfx file>",
+        "Password": "<certificate password>"
+      }
+    }
+  }
+}
+```
+
+An alternative to using **Path** and **Password** for the **Certificates** > **Default** certificate specification is to specify the certificate using the certificate store fields:
```
Tratcher: or for any certificate node.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,443 @@ uid: fundamentals/servers/kestrel
...
+* The `Certificate` section is optional. If the `Certificate` section isn't specified, the defaults defined in earlier scenarios are used. If no defaults are available, the server throws an exception and fails to start.
+* The `Certificate` section supports both **Path**&ndash;**Password** and **Subject**&ndash;**Store** certificates.
+* Any number of endpoints may be defined in this way so long as they don't cause port conflicts.
+* `serverOptions.Configure(context.Configuration.GetSection("Kestrel"))` returns a `KestrelConfigurationLoader` with an `.Endpoint(string name, options => { })` method that can be used to supplement a configured endpoint's settings:
+
+  ```csharp
+  serverOptions.Configure(context.Configuration.GetSection("Kestrel"))
+      .Endpoint("HTTPS", opt =>
+      {
+          opt.ListenOptions.Protocols = HttpProtocols.Http1;
```
Tratcher: We removed the Protocols property for 2.1.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,443 @@ uid: fundamentals/servers/kestrel
...
+      });
+  ```
+
+* The configuration section for each endpoint is a available on the options in the `Endpoint` method so that custom settings may be read.
+* Multiple configurations may be loaded by calling `serverOptions.Configure(context.Configuration.GetSection("Kestrel"))` again with another section. Only the last configuration is used unless `Load` is explicitly called on prior instances. The metapackage doesn't call `Load` so that its default configuration section may be replaced.
+* `KestrelConfigurationLoader` mirrors the `Listen` family of APIs from `KestrelServerOptions` as `Endpoint` overloads, so code and config endpoints may be configured in the same place. These overloads don't use names and only consume default settings from configuration.
+
+*Change the defaults in code*
+
+`ConfigureEndpointDefaults` and `ConfigureHttpsDefaults` can be used to change default settings for `ListenOptions` and `HttpsConnectionAdapterOptions`, including overriding the default certificate specified in the prior scenario.
```
Tratcher: They should be called before any endpoints are configured.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,443 @@ uid: fundamentals/servers/kestrel
...
+
+```csharp
+options.ConfigureEndpointDefaults(opt =>
+{
+    opt.Protocols = HttpProtocols.Http1;
```
Tratcher: Remove Protocols, you can set nagle instead.

Rick-Anderson:
@Tratcher  Do you have version 2.1 selected?
image

Tratcher:
Ah, I'd missed that. It defaulted to 1.0. Thanks Rick.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,445 @@ uid: fundamentals/servers/kestrel
...
+Call `serverOptions.Configure(context.Configuration.GetSection("Kestrel"))` to specify the configuration section. A default HTTPS app settings configuration schema is available for Kestrel. Configure multiple endpoints, including the URLs and the certificates to use, either from a file on disk or from a certificate store.
+
+In the following *appSettings.json* example:
+
+* Set **AllowInvalid** to `true` to permit the use of invalid certificates (for example, self-signed certificates).
+* **HttpsDefaultCert** indicates use of the cert defined under **Certificates** > **Default** or the development certificate.
+
+```json
+{
+"Kestrel": {
```
Tratcher: For some reason the rendered version flattens this JSON.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,445 @@ uid: fundamentals/servers/kestrel
...
+* The `Url` parameter is required for each endpoint. The format for this parameter is the same as the top-level `Urls` configuration parameter except that it's limited to a single value.
+* These endpoints replace those defined in the top-level `Urls` configuration rather than adding to them. Endpoints defined in code via `Listen` are cumulative with the endpoints defined in the configuration section.
+* The `Certificate` section is optional. If the `Certificate` section isn't specified, the defaults defined in earlier scenarios are used. If no defaults are available, the server throws an exception and fails to start.
+* The `Certificate` section supports both **Path**&ndash;**Password** and **Subject**&ndash;**Store** certificates.
+* Any number of endpoints may be defined in this way so long as they don't cause port conflicts.
+* `serverOptions.Configure(context.Configuration.GetSection("Kestrel"))` returns a `KestrelConfigurationLoader` with an `.Endpoint(string name, options => { })` method that can be used to supplement a configured endpoint's settings:
+
+  ```csharp
+  serverOptions.Configure(context.Configuration.GetSection("Kestrel"))
```
Tratcher: You can also directly access KestrelServerOptions.ConfigurationLoader to keep iterating on the existing loader, such as the one provided by WebHost.CreatedDeafaultBuilder.

aspnetcore/fundamentals/servers/kestrel.md
guardrex: I'm going to address this under line 175 because that's where the 2.0 mention of the `urls` config key is covered.

aspnetcore/fundamentals/servers/kestrel.md
guardrex: > For some reason the rendered version flattens this JSON.

See: https://github.com/aspnet/Docs/issues/6052

Tratcher:
By all means.

guardrex:
@blowdart We're adding a bit for Kestrel SNI support. The passage goes like this ...

*Kestrel support for SNI*

> \[Server Name Indication (SNI)](https://docs.microsoft.com/windows-server/security/tls/transport-layer-security-protocol#BKMK_SNI) can be used to host multiple domains on the same IP address and port. For SNI to function, the client sends the host name for the secure session to the server during the TLS handshake so that the server can provide the correct certificate. The client uses the furnished certificate for encrypted communication with the server during the secure session that follows the TLS handshake.

> Kestrel supports SNI via the `ServerCertificateSelector` callback. The callback is invoked once per connection to allow the app to inspect the host name and select the appropriate certificate.

> SNI support requires running on target framework `netcoreapp2.1`. On `netcoreapp2.0` and `net461`, the callback is invoked but the `name` is always `null`. The `name` is also `null` if the client doesn't provide the host name parameter in the TLS handshake.

... with the code sample from the blog post ( https://blogs.msdn.microsoft.com/webdev/2018/04/12/asp-net-core-2-1-0-preview2-now-available/ ). Corrections/suggestions? ... and is that a decent link for "Server Name Indication (SNI)"?

blowdart:
SNI link - https://tools.ietf.org/html/rfc6066#section-3

And that sample seems fine.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -341,30 +355,75 @@ Schema notes:
...
+                    allowInvalid: true);
+                var subExampleCert = CertificateLoader.LoadFromStoreCert(
+                    "sub.example.com", "My", StoreLocation.CurrentUser, 
+                    allowInvalid: true);
+                var certs = new Dictionary(StringComparer.OrdinalIgnoreCase);
+                certs["localhost"] = localhostCert;
+                certs["example.com"] = exampleCert;
+                certs["sub.example.com"] = subExampleCert;
+
+                httpsOptions.ServerCertificateSelector = (features, name) =>
```
Tratcher: FYI: the features type was changed to a context object in rc1.
https://github.com/aspnet/KestrelHttpServer/blob/2f2a0047a1715db6d3b82f4c74288f63aaf1fdd7/src/Kestrel.Core/HttpsConnectionAdapterOptions.cs#L50

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,504 @@ uid: fundamentals/servers/kestrel
...
-- Maximum request body size
-- Minimum request body data rate
+The Kestrel web server has constraint configuration options that are especially useful in Internet-facing deployments. A few important limits that can be customized:
 
-You set these constraints and others in the `Limits` property of the [KestrelServerOptions](https://github.com/aspnet/KestrelHttpServer/blob/rel/2.0.0/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerOptions.cs) class. The `Limits` property holds an instance of the [KestrelServerLimits](https://github.com/aspnet/KestrelHttpServer/blob/rel/2.0.0/src/Microsoft.AspNetCore.Server.Kestrel.Core/KestrelServerLimits.cs) class. 
+* Maximum client connections
+* Maximum request body size
+* Minimum request body data rate
+
+Set these and other constraints on the `Limits` property of the [KestrelServerOptions](/dotnet/api/microsoft.aspnetcore.server.kestrel.core.kestrelserveroptions) class. The `Limits` property holds an instance of the [KestrelServerLimits](/dotnet/api/microsoft.aspnetcore.server.kestrel.core.kestrelserverlimits) class.
```
scottaddie: In the first sentence, remove the code-fencing on Limits and link to its API ref:
/dotnet/api/microsoft.aspnetcore.server.kestrel.core.kestrelserveroptions.limits

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,504 @@ uid: fundamentals/servers/kestrel
...
-The recommended way to override the limit in an ASP.NET Core MVC app is to use the [RequestSizeLimit](https://github.com/aspnet/Mvc/blob/rel/2.0.0/src/Microsoft.AspNetCore.Mvc.Core/RequestSizeLimitAttribute.cs) attribute on an action method:
+The recommended approach to override the limit in an ASP.NET Core MVC app is to use the [RequestSizeLimit](/dotnet/api/microsoft.aspnetcore.mvc.requestsizelimitattribute) attribute on an action method:
 
 ```csharp
 [RequestSizeLimit(100000000)]
 public IActionResult MyActionMethod()
 ```
 
-Here's an example that shows how to configure the constraint for the entire application, every request:
+Here's an example that shows how to configure the constraint for the entire app, every request:
```
scottaddie: Should the ", every request" portion of this sentence be removed? Or maybe that should say ", on every request"?

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,504 @@ uid: fundamentals/servers/kestrel
...
+
+For more information, see [Server URLs](xref:fundamentals/hosting#server-urls) and [Overriding configuration](xref:fundamentals/hosting#overriding-configuration).
+
+The value provided using these approaches can be one or more HTTP and HTTPS endpoints (HTTPS if a default cert is available). Configure the value as a semicolon-separated list (for example, `"Urls": "http://localhost:8000;http://localhost:8001"`).
+
+*Replace the default certificate from configuration*
+
+[WebHost.CreateDefaultBuilder](/dotnet/api/microsoft.aspnetcore.webhost.createdefaultbuilder) calls `serverOptions.Configure(context.Configuration.GetSection("Kestrel"))` by default to load Kestrel configuration. A default HTTPS app settings configuration schema is available for Kestrel. Configure multiple endpoints, including the URLs and the certificates to use, either from a file on disk or from a certificate store.
+
+In the following *appSettings.json* example:
```
scottaddie: *appSettings.json* --> *appsettings.json*

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,504 @@ uid: fundamentals/servers/kestrel
...
-You can listen on a Unix socket for improved performance with Nginx, as shown in this example:
+Listen on a Unix socket with [ListenUnixSocket](/dotnet/api/microsoft.aspnetcore.server.kestrel.core.kestrelserveroptions.listenunixsocket) for improved performance with Nginx, as shown in this example:
 
-[!code-csharp[](kestrel/sample2/Program.cs?name=snippet_UnixSocket)]
+[!code-csharp[](kestrel/samples/2.x/Program.cs?name=snippet_UnixSocket)]
 
 **Port 0**
 
-If you specify port number 0, Kestrel dynamically binds to an available port. The following example shows how to determine which port Kestrel actually bound to at runtime:
+When the port number is `0` is specified, Kestrel dynamically binds to an available port. The following example shows how to determine which port Kestrel actually bound at runtime:
```
scottaddie: When the port number ~~is~~ `0` is specified

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,504 @@ uid: fundamentals/servers/kestrel
...
-* You can't use SSL with these methods.
-* If you use both the `Listen` method and `UseUrls`, the `Listen` endpoints override the `UseUrls` endpoints.
+* SSL can't be used with these approaches unless a default certificate is provided in the HTTPS endpoint configuration (for example, using `KestrelServerOptions` configuration or a configuration file as shown earlier in this topic).
+* When both the `Listen` and `UseUrls` approaches are used simultaneously, the `Listen` endpoints override the `UseUrls` endpoints.
 
-**Endpoint configuration for IIS**
+**IIS endpoint configuration**
 
-If you use IIS, the URL bindings for IIS override any bindings that you set by calling either `Listen` or `UseUrls`. For more information, see [Introduction to ASP.NET Core Module](aspnet-core-module.md).
+When using IIS, the URL bindings for IIS override bindings set by either `Listen` or `UseUrls`. For more information, see the [ASP.NET Core Module](xref:fundamentals/servers/aspnet-core-module) topic.
```
scottaddie: Should the word "are" appear before "set by either"?

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,220 +15,504 @@ uid: fundamentals/servers/kestrel
...
+      .Endpoint("HTTPS", opt =>
+      {
+          opt.HttpsOptions.SslProtocols = SslProtocols.Tls12;
+      });
+  ```
+
+  You can also directly access `KestrelServerOptions.ConfigurationLoader` to keep iterating on the existing loader, such as the one provided by `WebHost.CreatedDeafaultBuilder`.
+
+* The configuration section for each endpoint is a available on the options in the `Endpoint` method so that custom settings may be read.
+* Multiple configurations may be loaded by calling `serverOptions.Configure(context.Configuration.GetSection("Kestrel"))` again with another section. Only the last configuration is used unless `Load` is explicitly called on prior instances. The metapackage doesn't call `Load` so that its default configuration section may be replaced.
```
scottaddie: Add a comma after "Only the last configuration is used".

aspnetcore/fundamentals/servers/kestrel.md
guardrex: Let's drop "entire," too.

> Here's an example that shows how to configure the constraint for the app on every request:

guardrex:
ty @Tratcher and @scottaddie 

@halter73 About to merge this one ... do you have any additional corrections/suggestions?

PR #6139, opened 2018-04-30
few missed lines

dengere:
closeButton missed to define. 
and tr td missed add.


scottaddie:
@dengere Thank you for correcting these problems. We really appreciate you taking the time to do so.

PR #6142, opened 2018-04-30
Add AllowedHosts

### LINKED ISSUE #5793: Add AllowedHosts to proxy config topic
https://github.com/aspnet/BasicMiddleware/blob/dev/src/Microsoft.AspNetCore.HttpOverrides/ForwardedHeadersOptions.cs#L70-L84

guardrex:
Fixes #5793 

@Tratcher @scottaddie In the new table, you only need to look at the first row for `AllowedHosts`. Nothing else in the table changes on this PR.

@Tratcher By "normalized form" in the [API comments](https://github.com/aspnet/BasicMiddleware/blob/dev/src/Microsoft.AspNetCore.HttpOverrides/ForwardedHeadersOptions.cs#L81), it means "conventional" format, correct? ... because the spec doesn't refer to "normalized" ... it only speaks to "conventional" formats.

Also, leading zeros are fine in IPv6 "conventional" addresses ... they're ok here as well (i.e., Kestrel fills them in perhaps)?


aspnetcore/host-and-deploy/proxy-load-balancer.md
```diff
@@ -118,6 +119,23 @@ services.Configure<ForwardedHeadersOptions>(options =>
...
+::: moniker-end
+::: moniker range=">= aspnetcore-2.1"
+| Option | Description |
+| ------ | ----------- |
+| `AllowedHosts` | Restricts hosts by the `X-Forwarded-Host` header to the values provided. Port numbers must be excluded. If the list is empty, all hosts are allowed. A top level wildcard `*` allows all non-empty hosts. Subdomain wildcards are permitted but don't match the root (naked) domain [for example, `*.contoso.com` matches the subdomain `foo.contoso.com` but not the root (naked) domain `contoso.com`]. Unicode host names are allowed but are converted to [Punycode](https://tools.ietf.org/html/rfc3492) for matching. [IPv6 addresses](https://tools.ietf.org/html/rfc4291) must include bounding brackets and be in [conventional form](https://tools.ietf.org/html/rfc4291#section-2.2) (for example, `[ABCD:EF01:2345:6789:ABCD:EF01:2345:6789]`). Failure to restrict the allowed hosts may allow an attacker to spoof links generated by the service.<br><br>The default value is an empty [IList\<string>](/dotnet/api/system.collections.generic.ilist-1). |
```
scottaddie: - top level --> top-level
- Delete the square braces around the following, and make it a sentence:
> [for example, `*.contoso.com` matches the subdomain `foo.contoso.com` but not the root (naked) domain `contoso.com`]

guardrex:
Since the bracketed part of that item is part of one requirement, let's try an in-cell list. Let me know if it breaks, since I can't see it.

https://review.docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-1.0&branch=pr-en-us-6142

guardrex:
I see a new DocFX is out tho ... Version 2.35.2 ... I'll give that try and see if more bits are working locally.

scottaddie:
Here's how it displays:
image

Can you also remove the code fencing from AllowedHosts?

Tratcher:
Naked?

aspnetcore/host-and-deploy/proxy-load-balancer.md
blowdart: It's one of the terms. Bare or apex are the others, but they're not as common.

guardrex:
@scottaddie U sure ... if it's a property, I thought we were fencing them. Well ... truly ... I've been using API references (links) without styling, but that's when it's possible to do so. I can't with the 2.1 stuff. I don't think that matters here, so just confirm that you want it removed given that it is a property.

@Tratcher ...

> Values are compared using ordinal-ignore-case. IPv6 addresses are not special cased to check for logical equality between different formats and no canonicalization will be performed for you.

In which case, would you say it's inaccurate for the API comments to remark that "normalized form" is ok? ... I mean I guess it's just unclear what that means in context of the spec. Anyway ... On the next commit, I'll get an updated line in there for that along the lines of what you just said.

YES! **"naked!"** lol I hate for it to sound so seedy ... many devs know the "root" domain as the "naked domain" down in SoCal. Probably a bunch of whacky devs down there who like the word. I'll change it to "root domain" ... I think our "naked"-domain peeps will know that it's the same thing. :smile:

scottaddie:
@guardrex Normally, I'd have you keep the code-fencing. In this case, all the other properties aren't fenced. For the sake of consistency, and to avoid extra work, it's best to remove.

PR #6184, opened 2018-05-02
Update libuv references in Kestrel docs

### LINKED ISSUE #5695: Update libuv references in Kestrel docs
https://github.com/aspnet/Announcements/issues/296

Per the announcement, managed sockets are replacing libuv in Kestrel as of ASP.NET Core 2.1. Update the docs to mention this and that libuv is used in versions prior to that.

Affected docs:
- [x] https://docs.microsoft.com/aspnet/core/fundamentals/servers/
- [x] https://docs.microsoft.com/aspnet/core/fundamentals/servers/kestrel

guardrex:
Fixes #5695


aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -75,6 +82,27 @@ ASP.NET Core project templates use Kestrel by default. In *Program.cs*, the temp
 
 [!code-csharp[](kestrel/samples/2.x/Program.cs?name=snippet_DefaultBuilder&highlight=7)]
 
+::: moniker range=">= aspnetcore-2.1"
+**Default transport changes from Libuv to managed sockets (ASP.NET Core 2.1 or later)**
+
+With the release of ASP.NET Core 2.1, Kestrel's default transport is no longer based on Libuv but instead based on managed sockets. This is a breaking change for ASP.NET Core 2.0 apps that call [WebHostBuilderLibuvExtensions.UseLibuv](/dotnet/api/microsoft.aspnetcore.hosting.webhostbuilderlibuvextensions.uselibuv) and depend on either of the following packages:
+
+* [Microsoft.AspNetCore.Server.Kestrel](https://www.nuget.org/packages/Microsoft.AspNetCore.Server.Kestrel/)
+* [Microsoft.AspNetCore.App](https://www.nuget.org/packages/Microsoft.AspNetCore.App/)
```
natemcmaster: Since Microsoft.AspNetCore.App is new in 2.1, I might change the phrasing. This is only a breaking change for those using the Microsoft.AspNetCore.Server.Kestrel package directly without references to any metapackages. 

Users upgrading from 2.0 to 2.1 shouldn't break unless they also take action to move from .All to .App. But in that case, they will break in a bunch of other ways too since .All doesn't have Redis, Sqlite, etc.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -75,6 +82,27 @@ ASP.NET Core project templates use Kestrel by default. In *Program.cs*, the temp
...
+
+```xml
+<PackageReference Include="Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv" 
+                  Version="2.1.0" />
+```
+
+> [!NOTE]
+> The [Microsoft.AspNetCore.All](xref:fundamentals/metapackage) metapackage doesn't directly reference [Libuv](https://www.nuget.org/packages/Libuv/), but it does directly reference [Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv](https://www.nuget.org/packages/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/). The Libuv transport package is transitively referenced when the `Microsoft.AspNetCore.All` package is referenced. No action is required by the developer when the `Microsoft.AspNetCore.All` metapackage is used and `UseLibuv` is called in the app. Note that the ASP.NET Core 2.1 or later templates reference the [Microsoft.AspNetCore.App](https://www.nuget.org/packages/Microsoft.AspNetCore.App/) metapackage by default. The use of the `Microsoft.AspNetCore.All` metapackage isn't recommended for ASP.NET Core 2.1 or later apps.
```
natemcmaster: This is pretty wordy. I think your main point here is that the previous step is unnecessary in some cases, right? If so, consider shortening to

> No additional reference is required when using the Microsoft.AspNetCore.All metapackage, which will pull in Libuv transitively. For more details on this metapackage, see {cref}

aspnetcore/fundamentals/servers/kestrel.md
natemcmaster: Oops, I meant *.App* doesn't have Redis, Sqlite.

aspnetcore/fundamentals/servers/kestrel.md
guardrex: > main point here

It's there to explain why the `All` package doesn't require the dev to take action. I think we should be careful not to get too concise when providing technical details. It might be "wordy," but it's easy to read and understand because it doesn't try to say too much in a handful of words. However, I think we can drop the first sentence with a modification to the second sentence that will shorten this. Let's see how it looks on the next commit.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -75,6 +82,24 @@ ASP.NET Core project templates use Kestrel by default. In *Program.cs*, the temp
 
 [!code-csharp[](kestrel/samples/2.x/Program.cs?name=snippet_DefaultBuilder&highlight=7)]
 
+::: moniker range=">= aspnetcore-2.1"
+**Default transport changes from Libuv to managed sockets (ASP.NET Core 2.1 or later)**
```
natemcmaster: Reviewing this in context of the rest of the document now...

I think this new block of text belongs lower in the doc. Transport configuration deserves its own section, maybe after "Endpoint configuration" and before "URL prefixes". The breaking change from Libuv to Sockets should be one of the things mentioned in that section, but I don't expect this breaking change to affect many people. We're hoping that in most cases no action will be required. 

(I'm not expecting you to write this, but) it would also be good to have the transport configuration section explain why you might want to use sockets vs libuv. That can be punted for now, but it's worth addressing because changing the transport off default isn't a common scenario. We should be able to justify why it's ever needed.

guardrex:
@natemcmaster Sounds good. I moved the section down on the last commit.

> it would also be good to have the transport configuration section explain why you might want to use sockets vs libuv

Can you or one of the engineers provide the content? ... even if only a list of pros/cons of each or a single list of why sockets are better. It would be good for us to get enough coverage in place for now to close this issue and not open a new one (if possible).

guardrex:
Are the motivations for sockets over Libuv you laid out in https://github.com/aspnet/KestrelHttpServer/issues/2360#issue-302421622 applicable for the community?

Motivations:

* broad platform support. Wherever managed sockets are supported, Kestrel should work. Doesn't require recompiling libuv to bring up a new OS.
* source-build. (https://github.com/dotnet/source-build). In current form, Kestrel packages can't be produced in a source-build environment
* ecosystem. Helps to drive improvement to the .NET sockets implementation, which benefits anyone else using managed sockets.

If so, I can draft some text that we can work forward.

halter73:
@guardrex Those are the primary motivations to default to the socket transport. I don't think we should call too much attention to configuring Kestrel transports in the docs though. I expect that the difference between transports won't be noticeable to the vast majority of developers.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -489,6 +496,28 @@ When using IIS, the URL bindings for IIS override bindings set by `UseUrls`. For
...
+
+For ASP.NET Core 2.1 or later projects that use the `Microsoft.AspNetCore.App` metapackage and require the use of Libuv (the app calls [WebHostBuilderLibuvExtensions.UseLibuv](/dotnet/api/microsoft.aspnetcore.hosting.webhostbuilderlibuvextensions.uselibuv)), add a dependency for the [Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv](https://www.nuget.org/packages/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/) package to the app's project file:
+
+```xml
+<PackageReference Include="Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv" 
+                  Version="2.1.0" />
+```
+
+No action is required when the [Microsoft.AspNetCore.All](xref:fundamentals/metapackage) metapackage is used and `UseLibuv` is called because the Libuv transport package is transitively referenced by the `Microsoft.AspNetCore.All` metapackage. Note that the ASP.NET Core 2.1 or later templates reference the [Microsoft.AspNetCore.App](https://www.nuget.org/packages/Microsoft.AspNetCore.App/) metapackage by default. The use of the `Microsoft.AspNetCore.All` metapackage isn't recommended for ASP.NET Core 2.1 or later apps.
```
halter73: I didn't realize that Microsoft.AspNetCore.All was deprecated. Maybe we shouldn't mention it at all and just tell people to reference Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv directly.

guardrex: I believe they're phrasing it more along the lines of "not recommended" and "subject to future removal" at this time. That's why I slanted the language in this way (i.e., not use the word "depreciated" there). *they = `@DamianEdwards` in Standup remarks.

natemcmaster:
I would leave out the reasons for changing the default. IMO the only important reason document transport config at all is for compatibility. If users upgrade to 2.1 but rely on some libuv-only features, they will need to switch their transport to Libuv. At the moment, we're only aware of a few, rarely used features of libuv that are not available in sockets. +1 to what @halter73 said - we expect transport configuration to be an uncommon option and 99% of users will be fine with the default.

aspnetcore/fundamentals/servers/kestrel.md
natemcmaster: This is a good point. I'm not sure this is a recommendation we actually need to make yet. But even if we do, this seems like information that belongs on the metapackage doc, not here.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -15,7 +15,14 @@ uid: fundamentals/servers/kestrel
 
 By [Tom Dykstra](https://github.com/tdykstra), [Chris Ross](https://github.com/Tratcher), and [Stephen Halter](https://twitter.com/halter73)
 
+::: moniker range="<= aspnetcore-2.0"
 Kestrel is a cross-platform [web server for ASP.NET Core](xref:fundamentals/servers/index) based on [libuv](https://github.com/libuv/libuv), a cross-platform asynchronous I/O library. Kestrel is the web server that's included by default in ASP.NET Core project templates.
+::: moniker-end
+::: moniker range=">= aspnetcore-2.1"
+Kestrel is a cross-platform [web server for ASP.NET Core](xref:fundamentals/servers/index) based on managed sockets (`Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets`), a cross-platform asynchronous I/O library. In versions of ASP.NET Core prior to 2.1, Kestrel is based on [libuv](https://github.com/libuv/libuv).
```
halter73: I think we should leave discussion of libuv and managed sockets transports out of the intro. Mentioning libuv before wasn't so bad because it was short, but now it seems overly technical for the intro paragraph. This should be irrelevant to the vast majority of ASP.NET developers.

guardrex: ok ... I'll address that on the next commit.

guardrex:
RE: The new section

Are we all on the same page?

* Leave this in it's new section
* Don't sweat more content than what I have

aspnetcore/fundamentals/servers/index.md
```diff
@@ -19,7 +19,12 @@ An ASP.NET Core app runs with an in-process HTTP server implementation. The serv
 
 ASP.NET Core ships two server implementations:
 
+::: moniker range="<= aspnetcore-2.0"
 * [Kestrel](xref:fundamentals/servers/kestrel) is a cross-platform HTTP server based on [libuv](https://github.com/libuv/libuv), a cross-platform asynchronous I/O library.
+::: moniker-end
+::: moniker range=">= aspnetcore-2.1"
+* [Kestrel](xref:fundamentals/servers/kestrel) is a cross-platform HTTP server based on managed sockets (`Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets`), a cross-platform asynchronous I/O library. In versions of ASP.NET Core prior to 2.1, Kestrel is based on [libuv](https://github.com/libuv/libuv).
```
halter73: I also don't think discussion of the transport belongs here. If we must include it, I wouldn't call `Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets` a cross-platform asynchronous I/O library. That's what libuv is, but the sockets transport is more limited. It's just a socket library that implements Kestrel's transport abstraction.

halter73:
@guardrex I think the content you've added already is enough. I would just keep the old intro's more-or-less with the libuv stuff removed.

aspnetcore/fundamentals/servers/index.md
guardrex: Sure, I can address that on the next commit.

aspnetcore/fundamentals/servers/kestrel.md
halter73: I don't know how detailed we want this paragraph to be, but I think something like "Kestrel is a cross-platform web server for ASP.NET Core that's included by default in ASP.NET Core project templates" is enough.

guardrex:
Ok ... I'll have a new commit up shortly.

> But even if we do, this seems like information that belongs on the metapackage doc, not here.

I can't assume that a reader will figure that out if they get stuck. We might be able to place it elsewhere and link to it (note tho that the 2.1 `App` topic hasn't even been written yet), but it isn't a burdensome amount of content here. It's only one short paragraph. In the future ... down the road a year ... I think it should definitely be moved. For now tho, I think we should keep this here.

Devs are likely going to hit this topic when their app's Kestrel bits fail (and we don't have a "Troubleshoot Kestrel" topic). In the metapackages and migrations topics, it will be a buried. Please remember that clicking the "This doc isn't helpful" button gets people canned ... not YOU! lol ... but green dinosaurs land in the unemployment line! 😅 😀

guardrex:
@Rick-Anderson Would you take a look and see if u feel I should cut/move content? ... particularly the last paragraph.

Rick-Anderson:
@guardrex  I'd go with what @halter73  suggests.

guardrex:
ty @Rick-Anderson 

@halter73 I think the last commit addresses your feedback ... let me know if not, and I'll square it away.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -489,6 +489,24 @@ When using IIS, the URL bindings for IIS override bindings set by `UseUrls`. For
...
+
+* [Microsoft.AspNetCore.Server.Kestrel](https://www.nuget.org/packages/Microsoft.AspNetCore.Server.Kestrel/) (direct package reference)
+* [Microsoft.AspNetCore.App](https://www.nuget.org/packages/Microsoft.AspNetCore.App/)
+
+For ASP.NET Core 2.1 or later projects that use the `Microsoft.AspNetCore.App` metapackage and require the use of Libuv (the app calls [WebHostBuilderLibuvExtensions.UseLibuv](/dotnet/api/microsoft.aspnetcore.hosting.webhostbuilderlibuvextensions.uselibuv)), add a dependency for the [Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv](https://www.nuget.org/packages/Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv/) package to the app's project file:
+
+```xml
+<PackageReference Include="Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv" 
+                  Version="2.1.0" />
+```
```
halter73: One more thing. Can we have a one-line snippet like this, but for the actual call to UseLibuv()?

aspnetcore/fundamentals/servers/index.md
guardrex: Yes, I think we should remove it. If I try to add that sentiment, it becomes perhaps a bit too complicated ...

> [Kestrel](xref:fundamentals/servers/kestrel) is a cross-platform HTTP server based on managed sockets (`Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets`), a cross-platform asynchronous I/O library. In versions of ASP.NET Core prior to 2.1, Kestrel is based on [libuv](https://github.com/libuv/libuv), which remains a viable alternative for 2.1 or later apps.

We can just go with this here ...

> > [Kestrel](xref:fundamentals/servers/kestrel) is a cross-platform HTTP server based on managed sockets (`Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets`), a cross-platform asynchronous I/O library.

guardrex:
@halter73 

> one-line snippet

How about a bit more context? ... I've learned the hard way from readers that they really don't like one liners all that much ... they get lost and don't know where to place the line. See the last commit.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -525,6 +525,23 @@ For ASP.NET Core 2.1 or later projects that use the `Microsoft.AspNetCore.App` m
...
+            var host = new WebHostBuilder()
+                .UseKestrel()
+                .UseLibuv()
+                .UseContentRoot(Directory.GetCurrentDirectory())
+                .UseStartup<Startup>()
+                .Build();
+
+            host.Run();
+        }
+    }
```
halter73: That's perfect 👍

aspnetcore/fundamentals/servers/index.md
```diff
@@ -20,14 +20,13 @@ An ASP.NET Core app runs with an in-process HTTP server implementation. The serv
 ASP.NET Core ships two server implementations:
 
 * [Kestrel](xref:fundamentals/servers/kestrel) is a cross-platform HTTP server based on [libuv](https://github.com/libuv/libuv), a cross-platform asynchronous I/O library.
```
halter73: Is this a 1.x section? If not, I'd just drop everything after and including "based on".  Maybe you could mention it's the default server, but that's covered almost immediately after this already.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -487,6 +490,61 @@ When using IIS, the URL bindings for IIS override bindings set by `UseUrls`. For
...
+        public static void Main(string[] args)
+        {
+            CreateWebHostBuilder(args).Build().Run();
+        }
+
+        public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
+            WebHost.CreateDefaultBuilder(args)
+                .UseLibuv()
+                .UseStartup<Startup>();
+    }
```
halter73: This looks great. I think the first version is enough since that's what most people see in their templates these days.

guardrex:
ty @halter73 ... the first one was just exhaustion taking over. Updates made.

aspnetcore/fundamentals/hosting.md
```diff
@@ -94,23 +95,23 @@ This setting controls the capture of startup errors.
...
+#### [ASP.NET Core 1.x](#tab/aspnetcore1x)
 
 ```csharp
 var host = new WebHostBuilder()
     .CaptureStartupErrors(true)
     ...
 ```
 
----
+* * *
```
scottaddie: The tab specification doc calls for 3 consecutive dashes. Let's stick with that.

aspnetcore/fundamentals/hosting.md
```diff
@@ -124,23 +125,23 @@ This setting determines where ASP.NET Core begins searching for content files, s
...
+#### [ASP.NET Core 1.x](#tab/aspnetcore1x)
 
 ```csharp
 var host = new WebHostBuilder()
     .UseContentRoot("c:\\mywebsite")
     ...
 ```
 
----
+* * *
```
scottaddie: Convert to 3 dashes.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -487,6 +492,44 @@ When using IIS, the URL bindings for IIS override bindings set by `UseUrls`. For
...
+    ```
+
+No action is required when the [Microsoft.AspNetCore.All](xref:fundamentals/metapackage) metapackage is used and `UseLibuv` is called because the Libuv transport package is transitively referenced by the `Microsoft.AspNetCore.All` metapackage. Note that the ASP.NET Core 2.1 or later templates reference the [Microsoft.AspNetCore.App](https://www.nuget.org/packages/Microsoft.AspNetCore.App/) metapackage by default. The use of the `Microsoft.AspNetCore.All` metapackage isn't recommended for ASP.NET Core 2.1 or later apps.
```
scottaddie: Since this section only displays for ASP.NET Core 2.1+, we shouldn't reference the older ".All" metapackage. That will really slim down this paragraph.

aspnetcore/fundamentals/hosting.md
guardrex: I just saw these start to change somewhere ... I thought they were going for a new format ... along with the four hash tabs.

Where did I see this coming in? I'll research and see where I got this.

guardrex: I think it was this one ... https://github.com/aspnet/Docs/pull/5878

guardrex: These were changed in a lot of spots (but not everywhere).

aspnetcore/fundamentals/servers/kestrel.md
guardrex: I think it would be best in a migration topic, but I don't see TO-2.1 content in the migration area.

This para has taken a lot of heat. :smile: lol I'd prefer to cut it at 3.0; but since everyone else dislikes it, I'll cut it.

PR #6187, opened 2018-05-02
Correct Code Sample Method Reference

AdrianJSClark:
The code sample and explaining text appear to be out of sync. This is a correction of the reference made to a function name in the surrounding text of the sample.


scottaddie:
@AdrianJSClark Thank you for correcting this mistake!

PR #6223, opened 2018-05-04
How to setup a developer certificate for Docker

Rick-Anderson:
@spboyer this surface the instructions until we have time to move them to the doc.


spboyer:
Why add a link to the issue, just add the content that is in the issue where the link is

PR #6240, opened 2018-05-07
wrong api link

user135711:
The extension method in the sample app is different from the linked api page. I was trying to figure something unrelated out, and got confused. Strangely, the same thing occurs in another above link (https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.options.iconfigureoptions-1) which is the wrong api but I didn't fix it because the autogenerated api doesn't have the correct extension method where it should [be] (https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.optionsservicecollectionextensions.configure?view=aspnetcore-2.0&viewFallbackFrom=aspnetcore-2.1#Microsoft_Extensions_DependencyInjection_OptionsServiceCollectionExtensions_Configure__1_Microsoft_Extensions_DependencyInjection_IServiceCollection_System_Action___0__)

### DIFF
--- a/aspnetcore/fundamentals/configuration/options.md
+++ b/aspnetcore/fundamentals/configuration/options.md
@@ -172,7 +172,7 @@ Named options support with [IConfigureNamedOptions](/dotnet/api/microsoft.extens
 
 *Requires ASP.NET Core 2.0 or later.*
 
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks for flagging these @user135711 ... I'll review shortly.

guardrex:
> The extension method in the sample app is different from the linked api page. 

Yes, thank you @user135711 for the update.

Although this merged 🏃 **_FAST_** 🏃 :smile:, I'll take a look at the other instance you mention.

PR #6266, opened 2018-05-08
missing db vscode api

user135711:
Db wasn't set up in the new version.


scottaddie:
Thank you for adding this, @user135711. It's very much appreciated.

PR #6279, opened 2018-05-08
Escape environment values service files

### LINKED ISSUE #6273: [Proposal] Enviornment variables that contains escape characters
## Proposal

[Connection string in Linux service file configuration as environment variable](https://github.com/aspnet/Configuration/issues/791) I raised an issue where the `Environment=ConnectionStrings__Default={value}` wasn't handled correctly if entered as is. For me, not all to familiar with escape values in Ubuntu services files, it wasn't obvious to use `systemd-escape`.

I thinks this should be added to the service files presented in `linux-nginx.md` and `linux-apache.md` in **Host and Deploy**.

Just a small information ui element just before the service file example that says

> Note. Some values, example SQL connection strings, need to be propertly escaped for the configuration providers to properly read the environment variables. On Ubuntu use `systemd-escape "{connectionString}"` to generate a working value to use in configuration file.

joacar:
Fixes #6273

### DIFF
--- a/aspnetcore/fundamentals/configuration/options.md
+++ b/aspnetcore/fundamentals/configuration/options.md
@@ -24,6 +24,8 @@ The options pattern uses options classes to represent groups of related settings
 
 ## Basic options configuration
 
+Options configuration need to be registered with dependency injection by calling the `AddOptions()` method on an instance of type `IServiceCollection`, typically `services.AddOptions()`. The method `AddOptions()` is in the namespace `Microsoft.Extensions.DependencyInjection`.
+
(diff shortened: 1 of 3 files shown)


joacar:
Gaah the commit [a834032](https://github.com/aspnet/Docs/pull/6279/commits/a834032d460079403b17873035f197ede04d3c15) shouldn't be here! Does it suffice just to revert it dirty, i.e. remove the added part as new commit? Gues this will be squashed and merged anyhow leaving no trail if it.

guardrex:
@joacar Yes, correct. We squash and merge, so that's fine.

guardrex:
Thanks for setting this up @joacar. I've requested a review from engineering.

PR #6282, opened 2018-05-08
Code ref incorrect, use line 38 instead => services.AddSignalR();

damienbod:
Ref to the wrong line in the code example, required is

line 38 => services.AddSignalR();

https://github.com/aspnet/Docs/blob/master/aspnetcore/signalr/hubs.md

page

https://docs.microsoft.com/en-us/aspnet/core/signalr/hubs?view=aspnetcore-2.1

@rachelappel


rachelappel:
Thanks @damienbod! Good catch!

damienbod:
thanks, sorry that I missed the other code ref fixes :)

rachelappel:
@damienbod No worries. Doing great!

Issue #5331, opened 2018-02-01
2.0 update of File Uploads doc


scottaddie:
16K PV/m 61% CSAT

https://docs.microsoft.com/aspnet/core/mvc/models/file-uploads

- [ ] Provide 2.0 code samples (address https://github.com/aspnet/Docs/pull/5217)
- [ ] UE edit (improve CSAT and use MVC style of writing)
- [ ] Address https://github.com/aspnet/Docs/issues/3989
- [x] Address https://github.com/aspnet/Docs/issues/4777
- [x] Address https://github.com/aspnet/Docs/issues/4964
- [ ] Address #6247
- [ ] Address #6349
- [ ] Address #6927
- [ ] Address #7702
- [x] Address #7742
- [ ] Address #8039
- [ ] Address #8100
- [x] Address #8298
- [ ] Address #8797
- [x] Address #9116
- [ ] Address #9327
- [ ] See #9473 for a possible enhancement opportunity.
- [x] Address #10161
- [ ] Address #11898 &ndash; Perhaps just a remark to process files in the background (i.e., [Background tasks with hosted services](https://docs.microsoft.com/aspnet/core/fundamentals/host/hosted-services)).

Will fix #6609, #6610, #6611

See #6962 for a Q&A on the 65535 file limit of `GetTempFileName` mentioned in the topic and in a code comment. This subject has just come up again on #10161.

MacGenius:
First, I'm new to visual studio enterprise. I'm trying to build the RazorPagesMovie app in five minutes but I'm already stuck at "add a model - scaffold the movie model". I'm not sure where to run "dotnet aspnet-codegenerator razorpage -m Movie -dc MovieContext -udl -outDir Pages\Movies --referenceScriptLibraries" I have run this in my pc cmd console, in the Package Manager Console, and in Visual Studio code Terminal, the VSC terminal actually ran this command. But I am unable to successfully test the app  in the step "Test the app
Run the app and append /Movies to the URL in the browser (http://localhost:port/movies)." This gives me a 404 error.  I did notice the "if" in the Movie class example posted here but the tutorial makes no mention of this line...again I am new to Visual Studio I apologize if this is basic stuff, but Ive used netbeans and eclipse, and android studio successfully. Any suggestions for a rookie to VSE so I can complete this 5 minute app?
![doh](https://user-images.githubusercontent.com/29048190/40341641-8372b0f6-5d43-11e8-840f-d0705da3d801.PNG)
[...]

PR #6344, opened 2018-05-10
Migrate FileUpload sample to ASP.NET Core 2x

teamkiller7112:
Migrate FileUpload sample to ASP.NET Core 2.0.

Disable request size limit for StreamingController by adding DisableRequestSizeLimit attribute.


user135711:
General feedback for the sample, not the pull request: I think the sample should actually save the uploaded image to a public folder or even better a database and then display it in the view in ProfileController.cs Line63 mainly because I want to learn how to do something similar with pdfs and books. I checked out this branch and it ran fine on my machine. Here's how I fixed the errors 2 weeks ago for [comparison](https://github.com/user135711/Docs/tree/FileUploadMigration/aspnetcore/mvc/models/file-uploads/sample/FileUploadSample).

aspnetcore/mvc/models/file-uploads/sample/FileUploadSample/FileUploadSample.csproj
```diff
@@ -0,0 +1,41 @@
...
+    <OutputType>Exe</OutputType>
+    <PackageId>FileUploadSample</PackageId>
+  </PropertyGroup>
+  <ItemGroup>
+    <None Update="wwwroot\**\*;Views\**\*;Areas\**\Views">
+      <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
+    </None>
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.2.1" />
```
Rick-Anderson: Use Microsoft.AspNetCore.App  see https://docs.microsoft.com/en-us/aspnet/core/fundamentals/metapackage-app?view=aspnetcore-2.1

guardrex:
btw - This is due for a major overhaul on https://github.com/aspnet/Docs/issues/5331.

guardrex:
😅 Finally have a chance to take a closer look here. We're going not going to be able to accept this because this makes a set of updates to the 1.0 sample that seek to convert it directly to 2.1 and we must maintain a 1.1 version of the sample with the topic. A PR that seeks to provide a 2.1 sample for the topic will require two versions of the sample and associated topic updates to render their code snippets. There are also a number of changes that are required beyond the updates here in order to make this conform to our 2.1 template.

This PR is linked now on the tracking issue (#5331) for the topic+sample, so these changes will be considered when that issue is worked. Thanks @teamkiller7112 for providing this.

Issue #6299, opened 2018-05-09
What is the "CacheKeys"?


rybomar:
It is hard to understand what is the "CacheKeys". Of course the search engine on GitHub helped me to find the answer but examples should to be more precisely (shuld be able to compile the code).

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 13f5f8f9-18dd-9c12-9ed1-fcc6b44ba7fd
* Version Independent ID: 640b4f36-5886-7b69-1c95-cfc8574a02fd
* Content: [Cache in-memory in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/memory?view=aspnetcore-2.1)
* Content Source: [aspnetcore/performance/caching/memory.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/performance/caching/memory.md)
* Service: **asp.net-core**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

guardrex:
@rybomar Thanks for commenting. I'll let @Rick-Anderson respond to the "CacheKeys" part of your question. I just wanted to mention that the old sample that's there is due for an upgrade after 2.1 releases. We're tracking it on a samples update issue: https://github.com/aspnet/Docs/issues/5495

Rick-Anderson:
fixed in #7792

Issue #5495, opened 2018-02-17
Sample Update Tracking Issue - complete list


guardrex:
**Last Updated**: 4/18/21

Doesn't include migration topic samples.

| | Current<br>Version | Topic + Notes |
| :---: | :---: | --- |
| 🟢 | 5.0, 3.1 | Blazor Samples |
| 🟢 | 2.1 | client-side/libman/index.md |
| 🔴 | 1.1 | client-side/spa-services.md |
| 🔴 | 2.0 | common/samples/WebApplication1 |
| 🔴 | 2.0 | common/samples/WebApplication1DotNetCore2.0App |
| 🔴 | 5.0, 2.2 | data\ef-mvc\intro.md |
| 🔴 | 5.0, 3.0, 2.1, 2.0 | data\ef-rp\intro.md |
| 🟢 | 3.1 | entity-framework-6.md |
| 🔴 | 3.0, 2.2 | fundamentals/metapackage-app.md |
| 🔴 | 3.1, 2.2 | fundamentals\app-state.md |
| 🔴 | 3.1, 2.2 | fundamentals\change-tokens.md |
| 🔴 | 3.1, 2.2 | fundamentals\configuration\index.md |
| 🔴 | 3.1, 2.2 | fundamentals\configuration\options.md |
| 🔴 | 3.1, 2.2 | fundamentals\dependency-injection.md |
| 🔴 | 3.1, 2.2 | fundamentals\environments.md |
| 🔴 | 5.0, 2.2 | fundamentals\error-handling.md |
| 🔴 | 3.1, 2.2 | fundamentals\file-providers.md |
| 🔴 | 2.2 | fundamentals\host\generic-host.md |
| 🔴 | 3.1, 2.2 | fundamentals\host\hosted-services.md |
| 🔴 | 3.0, 2.2, 2.1, net2.0 | fundamentals\host\platform-specific-configuration.md |
| 🔴 | 3.1, 2.2 | fundamentals\http-requests.md |
| 🟢 | 3.1, 2.1 | fundamentals\localization.md |
| 🔴 | 3.1, 3.0, 2.2 | fundamentals\logging\index.md |
| 🔴 | 3.1, 2.2 | fundamentals\logging\loggermessage.md |
| 🔴 | 2.0 | fundamentals/metapackage/samples/Metapackage.All.Example.csproj |
| 🔴 | 3.1, 2.2 | fundamentals\middleware\extensibility.md |
[...]

PR #6355, opened 2018-05-11
[WIP] Fix property assignment in primitives example

user135711:
Fixed a bug where CurrentState is assigned the wrong value on each invocation followed by being assigned the correct value in index. Also fixed an np completeness inefficiency by passing a lamda instead of a method group since I was editing the line, and used a non primitive state object so that the value is passed by reference to reflect the current state. For np explanation see
<a href="https://blogs.msdn.microsoft.com/ericlippert/2007/03/26/lambda-expressions-vs-anonymous-methods-part-four/">here</a>and
<a href="https://blogs.msdn.microsoft.com/ericlippert/2007/03/28/lambda-expressions-vs-anonymous-methods-part-five/">here</a>


user135711:
I didn't realize that github would squash any commit into my repository into any pull requests so this commit now has an unrelated commit included which makes webcache compilable. This fixes #6299.

guardrex:
Hello @user135711! Thanks for submitting this.

Would you mind removing the changes for the WebCache sample from this? We wouldn't want both samples updated on the same PR, and we can't take a PR for the WebCache sample at this time. For one thing, we can't upgrade it to 2.1 until 2.1 RTM. Also, it's due for a significant overhaul when it is upgraded that will go beyond just making it run under 2.1. We're tracking the work for it on #5495.

guardrex:
RE: The ChangeTokens updates

The changes proposed require matching changes to the topic's text for the `ConfigurationMonitor` ctor and the `InvokedChange` callback on the same PR so that the live sample doesn't become out-of-sync with the topic when the PR is merged.

https://docs.microsoft.com/aspnet/core/fundamentals/primitives/change-tokens#monitoring-configuration-changes-as-a-service

user135711:
@guardrex That's a good point, thanks...

guardrex:
If you prefer that this just be taken into account for the Change Tokens 2.1 update, I recommend just opening an issue referencing this, then I'll add a note to the tracking issue for it.

guardrex:
Alternatively, feel free to go ahead with a PR on the Change Tokens topic with these changes now, and I'll ping engineering to take a look.

user135711:
@guardrex Shrug, if you're doing it anyway, I'll just let you update them at once. I do have some suggestions for the change tokens update that I'll post directly to the issue so you'll consider including it and save me some time if you do it.

guardrex:
ok ... sounds good.

PR #6373, opened 2018-05-11
Update user secrets doc to 2.x

scottaddie:
Fixes https://github.com/aspnet/Docs/issues/6370
Fixes https://github.com/aspnet/Docs/issues/6186
Fixes https://github.com/aspnet/Docs/issues/4889

**Internal Review Pages**
* [ASP.NET Core 2.1 version](https://review.docs.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-2.1&branch=pr-en-us-6373&tabs=visual-studio)
* [ASP.NET Core 1.1 version](https://review.docs.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-1.1&branch=pr-en-us-6373&tabs=visual-studio)


aspnetcore/security/app-secrets.md
```diff
@@ -1,125 +1,250 @@
...
 
-## Installing the Secret Manager tool
+## How the Secret Manager tool works
 
-# [Visual Studio](#tab/visual-studio/)
+The Secret Manager tool abstracts away the implementation details, such as where and how the values are stored. You can use the tool without knowing these implementation details. The values are stored in a [JSON](https://json.org/) configuration file in a system-protected user profile folder on the local machine:
 
-Right-click the project in Solution Explorer, and select **Edit \<project_name\>.csproj** from the context menu. 
-Add the highlighted line to the *.csproj* file, and save to restore the associated NuGet package:
+* Windows: `%APPDATA%\Microsoft\UserSecrets\<user_secrets_id>\secrets.json`
```
natemcmaster: Originally, we wanted to provide tooling to work with this file. Using JSON to store the secrets was meant to be temporary until we could encrypt the secrets at rest. We *might* some day actually do that, which is why I wanted to express my reservations about documenting too much what goes on under the hood.

cc @glennc @danroth27 - thoughts on this? Originally, we said we wanted this file location to be an internal implementation detail that no one should ever see. But, two years in, I think enough people have discovered this that we could never change this without making a big breaking change.

aspnetcore/security/app-secrets.md
```diff
@@ -1,125 +1,250 @@
...
+* Linux & macOS: `~/.microsoft/usersecrets/<user_secrets_id>/secrets.json`
 
-[!code-xml[](app-secrets/sample/UserSecrets/UserSecrets-before.csproj?highlight=10)]
+In the preceding file paths, replace `<user_secrets_id>` with the `UserSecretsId` value specified in the *.csproj* file.
 
-Right-click the project in Solution Explorer again, and select **Manage User Secrets** from the context menu. This gesture adds a new `UserSecretsId` node within a `PropertyGroup` of the *.csproj* file, as highlighted in the following sample:
+Don't write code that depends on the location or format of data saved with the Secret Manager tool. These implementation details may change. For example, the secret values aren't encrypted, but could be in the future.
 
-[!code-xml[](app-secrets/sample/UserSecrets/UserSecrets-after.csproj?highlight=4)]
+## Install the Secret Manager tool
```
natemcmaster: This doesn't apply to 2.1 projects. Can we remove this section completely in the 2.1 version of the docs? Even though you mention this, I think the code sample gets more visibility here. Using DotNetCLiToolReference in 2.1 will lead to a warning.

image


https://review.docs.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-2.1&branch=pr-en-us-6373&tabs=visual-studio

aspnetcore/security/app-secrets.md
```diff
@@ -1,125 +1,250 @@
...
+  list    Lists all the application secrets
+  remove  Removes the specified user secret
+  set     Sets the user secret to the specified value
+
+Use "dotnet user-secrets [command] --help" for more information about a command.
 ```
 
-# [Visual Studio Code](#tab/visual-studio-code/)
+> [!NOTE]
+> You must be in the same directory as the *.csproj* file to run tools defined in the *.csproj* file's `DotNetCliToolReference` elements.
```
natemcmaster: (as you already know) this only applies to 1.1. In 2.1, you can do `dotnet user-secrets --project ./path/to/app.csproj`.

aspnetcore/security/app-secrets.md
```diff
@@ -1,125 +1,250 @@
...
+```console
+dotnet user-secrets set "Movies:ServiceApiKey" "12345" --project "C:\apps\WebApp1\src\WebApp1"
+```
+
+## Set multiple secrets
+
+A batch of secrets can be set by piping JSON to the `set` command. In the following example, the *secrets.json* file's contents are piped to the `set` command:
+
+```console
+type .\secrets.json | dotnet user-secrets set
```
natemcmaster: Maybe using `input.json` would make it clearer that you're not talking about the $HOME/.microsoft/usersecrets/id/secrets.json file.

Also, you might also mention this only works on Windows. Use `cat inputs.json` for macOS/Linux

aspnetcore/security/app-secrets.md
```diff
@@ -1,125 +1,250 @@
...
-Test the Secret Manager tool by running the following command:
+::: moniker range="<= aspnetcore-1.1"
+[!code-xml[](app-secrets/samples/1.1/UserSecrets/UserSecrets.csproj?name=snippet_PropertyGroup&highlight=3)]
+::: moniker-end
+::: moniker range=">= aspnetcore-2.0"
+[!code-xml[](app-secrets/samples/2.1/UserSecrets/UserSecrets.csproj?name=snippet_PropertyGroup&highlight=3)]
+::: moniker-end
+
+> [!TIP]
+> In Visual Studio, right-click the project in Solution Explorer, and select **Manage User Secrets** from the context menu. This gesture adds a `UserSecretsId` element, populated with a GUID, to the *.csproj* file. Visual Studio opens a *secrets.json* file in the text editor. Replace the contents of *secrets.json* with the key-value pairs to be stored. For example:
```
natemcmaster: Just a heads up - one of the things users often complain about is that the command-line tool does not preserve nesting in the secrets.json file when making changes. This is for good technical reasons that we don't intend to fix, especially because we hope to replace this JSON editing experience with a UI in a future version of Visual Studio.

PR #6396, opened 2018-05-12
fix null ptr in routing

user135711:
Some of the controllers don't have a namespace so this allows the project to run.


Rick-Anderson:
@user135711  thanks!

PR #6399, opened 2018-05-14
Updated connection string code

ehighbaugh:
Update connection string code for current version of Entity Framework with LocalDb; prior code no longer works.

No reference number.


scottaddie:
I've validated that the connection string provided in the tutorial works properly. The connection string in your proposed change doesn't work for me.

Please make sure you're using the latest release of VS 2017, which is version 15.7.1 at the time of writing. If the tutorial still gives you trouble, please let us know.

PR #6409, opened 2018-05-14
CC56109: Extra spaces to avoid loc issues

carolinacmoravia:
Hello, @Rick-Anderson,

This proposed file change comes from https://github.com/aspnet/Docs.ja-jp/pull/29/files. 

Many thanks in advance.

  Fixes #Issue_Number


guardrex:
Shouldn't the repair in this case be inserting a blank markdown line? The two-spaces will render a bare return (`<br>`), which we do use but just not in this scenario.

scottaddie:
@carolinacmoravia Thank you!

PR #6416, opened 2018-05-14
Minor changes in Get Started guide for SignalR

shahedc:
1. Following the instructions as is, the code app.UseBrowserLink() doesn’t compile in the Startup class. I had to add the BrowserLink package via NuGet to get the project to compile. (This package is included in the .csproj file in the Github sample, so only the doc needs to be updated for this.)

2. The “signalr.js” should be in a "signalr" subfolder, not directly in the "lib" folder, as it says in the docs. “Copy the signalr.js file from node_modules\@aspnet\signalr\dist\browser to the lib folder in your project.”

### DIFF
--- a/aspnetcore/signalr/get-started.md
+++ b/aspnetcore/signalr/get-started.md
@@ -126,9 +126,9 @@ The SignalR server must be configured so that it knows to pass requests to Signa
 
    `services.AddSignalR` adds SignalR as part of the [middleware](xref:fundamentals/middleware/index) pipeline.
 
-2. Before you can use the code sample shown in the Startup.cs file, you will also need to add the BrowserLink package via Nuget. Simply right-click on your project in Solution Explorer, and click Manage NuGet Packages to browse for and install the package named "Microsoft.VisualStudio.Web.BrowserLink".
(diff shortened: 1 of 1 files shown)


rachelappel:
Closing/Reopening because it appears to not be building.

Rick-Anderson:
@rachelappel  Why do we need  BrowserLInk in this sample. Can't we remove it. Would it make the tutorial simpler if we removed it?

Rick-Anderson:
@rachelappel  @shahedc  please review my changes. I'd rather remove browserlink from startup and simplify the tutorial - if that makes it simpler.  
Why aren't you using M.A.All or M.A.App which includes BL.

rachelappel:
@Rick-Anderson There have been too many complaints and it's actually become easier to just have them put this one in. We don't *need* it for SignalR but the project bombs without it and it has been more confusing to people to remove it than to just add the package. Frankly it should be in the template and I expect that will happen at some point so this will be moot soon enough anyway.

shahedc:
Just checked out the above comments, and I'll all for any action that resolves build errors and/or also clarifies the instructions for the reader. Thanks, all!

rachelappel:
Fixed here https://github.com/aspnet/Docs/pull/6532

PR #6417, opened 2018-05-14
Explain how to get bootstrap.cs for local use

shahedc:
Downloading the github sample doesn’t include the bootstrap CSS. The layout CSHTML expects a local copy of the bootstrap CSS file when in “Development”, so I think the docs should be updated to mention this. I had to add the CSS file manually in my project to resolve this.

Location of _Layout.cshtml: /Docs/aspnetcore/signalr/get-started/sample/Pages/_Layout.cshtml


rachelappel:
@shahedc  We should just add in all the bootstrap items, to avoid explaining things that aren't SignalR in the doc.

Rick-Anderson:
@rachelappel  I updated the .gitignore so it's trivial to update the sample to include wwwroot/lib

see Update SignalR samples to include wwwroot/lib #6433

PR #6421, opened 2018-05-15
Capitalize search string 'Ghost' in URL segment

wooleyra:
Since the searching method in SQLite is case-sensative, the example should pass the search string in the correct case.


Rick-Anderson:
@wooleyra  Thanks. Case not needed but it looks nicer.

PR #6446, opened 2018-05-16
Clarifying TempData cookie encryption

bgribaudo:
Current wording suggests the possibilty that `Base64UrlTextEncoder` is both encoding and encrypting `TempData`'s cookie data. Clarifying that encryption is handled by a different (and more appropriate) class.


blowdart:
It looks correct :)

scottaddie:
@bgribaudo Thank you for your contribution.

PR #6504, opened 2018-05-20
Add extension overload note to change the display name

### LINKED ISSUE #6503: Change the default name wsfederation
I have posed my query on stackoverflow. How do i change the default name produced "wsfederation" to a name of my choice after enabling adfs (on premise) as shown above. The link https://stackoverflow.com/questions/50431964/asp-net-core-authetication-with-adfs-change-the-button-name-from-wsfederation

---
#### Document Details

* ID: 59d6d442-8d46-3f9e-e75a-ad9d900fc117
* Version Independent ID: d07c4058-452f-cdac-c59d-8f09e70a43f3
* Content: [Authenticate users with WS-Federation in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/ws-federation?view=aspnetcore-2.0)
* Content Source: [aspnetcore/security/authentication/ws-federation.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/security/authentication/ws-federation.md)
* Service: **unspecified**
* Product: **asp.net-core**
* GitHub Login: @chlowell

guardrex:
Fixes #6503 

Thanks @jgathairu! ⚽️

Thanks @poke! 🏎 ... and your note on changing the UI feels a bit more workaround-ish, so perhaps it's best to only officially document the overload.


poke:
Note that the overload exists for all providers and it makes sense for others to change it there too (e.g. OIDC and OAuth). So you might want to adjust the other docs too if you think this is worth mentioning (imho, if we started documenting every overload, the documentation would be hard).

As for my “workaround” (which wasn't meant as such), in real projects I don't really expect the default Identity UI to be used that much. Modifications are likely, including placement and generation of login links.

guardrex:
> overload exists for all providers

Yes, I'll do that now.

> if we started documenting every overload, the documentation would be hard

Another way to handle this is just to say that there are overloads that *do things like set the display name* (i.e., just use display name as an example and don't detail a specific overload). That's probably a better approach.

> Modifications are likely, including placement and generation of login links.

I'm presuming that it's a secondary approach because they went to the trouble of providing the overload that can set the display name. Seems like the simplest approach if that's all one wishes to do.

guardrex:
@poke :eyes: the last commit. How's that look?

poke:
> Subsequent calls to it have the potential of overriding any previously configured AuthenticationOptions properties.

Note that subsequent calls to `AddXYScheme` with the same scheme name will raise an exception. You cannot reconfigure already registered schemes.

> because they went to the trouble of providing the overload that can set the display name

They probably did that to provide a way to specify all of the two properties that are set [when adding a scheme](https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationSchemeBuilder.cs) that cannot be adjusted later on.

Otherwise, the change looks fine to me.

guardrex:
A few other fix-ups added, such as API ref links.

poke:
Sorry, I think I misunderstood the original text there. That’s what I get from reviewing pull requests on my phone..

There may be multiple calls to `AddAuthentication` but that will update the authentication configuration. Only the `AddXY` calls on the AuthenticationBuilder may only be done once per scheme.

What about the following text?

> The `AddAuthentication(String)` overload sets the default scheme. The `AddAuthentication(Action<AuthenticationOptions>)` overload allows configuring authentication options, which can be used to set up default authentication schemes for different purpses. Subsequented calls to `AddAuthentication` allow updating the options.
>
> `AuthenticationBuilder` extension methods that register an authentication scheme may only be called once per authentication scheme. There exist various overloads that allow configuring the scheme properties, scheme name and display name.

guardrex:
@poke Try the last commit ... it's a hybrid of your new suggestions and some things that I liked about the old language, if I what I had wasn't a big pile of 🐴 💩 😄.

scottaddie:
Do these changes look good to you, @blowdart?

blowdart:
Sure

PR #6513, opened 2018-05-21
Fix format

nopara73:
`[Edit by @rick-anderson]`
added:

**Note:** Linux has a case-sensitive file system. Setting ASPNETCORE_ENVIRONMENT to "Production" results in searching for the configuration file *appsettings.Production.json*, not *appsettings.production.json*.


shirhatti:
That shouldn't be the case. What version are you observing this in? Are you using `CreateDefaultBuilder()`?
cc @Tratcher

nopara73:
@shirhatti to be clear, look ad the diff, I simply added two spaces to separate the notes, my comment above seems to be an autogenerated message, I wasn't even aware of it.

Issue #78, opened 2015-04-17
change request for "Understand the Fundamental Concepts of Web Programming"


Rick-Anderson:
The title need to change to something like "ASP.NET 5 project structure overview" - article is not about Web programming.
The following sections need more details on when/why/how I would use the feature. These details can come from links to other articles. 
Expand on  commands section (what can’t it do)
Expand on (for aspnetcore50 only - aspnet50 will include the full framework )need detailed explanation on how to choose )
Expand  The exclude section is used to identify files and folders that should be excluded from builds. Likewise, bundleExclude - why would you exclude, at least a pointer to bundling and why exclude that.
Expand The scripts section—explain postrestore/prepare and maybe some links. 
The scripts section is used to specify when certain build automation scripts should run.
What does certain me?
running such scripts before and after certain events
Shouldn’t that be specific events?  Explain the events listed  and perhaps list other useful events.

allowing build artifacts to be placed in a sibling folder
What is the sibling folder? .vs, artifacts?

>> You’ll notice that each of the items listed in the figure above correspond to dependencies listed in bower.json

Can we use code in lieu of an image? This has already changed and code is easier to change than images.Code is searchable. 

Can you explain the value key in the dependencies (that is, the version) and perhaps wild cards? What’s the meaning of  exportsOverride?   I think we’re better off replacing the image with code as long as we have decent pretty print, if not, stick with the image.
What are the child elements of each dependency - both the key and the key values including wildcarding.

Grunt/guplfile: again, replace image with code and explain elements.

>>configured NuGet package source (specified under Options – NuGet Package Manager – Package Sources).
I think that should be (Tools > NuGet Package Manager > Package Manager Settings > Package Sources).
(Editor note: Use > to navigate UI)
>>Accessing configuration data from your application is best done by injecting the IConfiguration interface into your Controller

For important references (IConfiguration), can we make that a link to the GIT source? for example https://github.com/aspnet/Configuration/blob/dev/src/Microsoft.Framework.ConfigurationModel.Interfaces/IConfiguration.cs
>>First, add the entry to project.config.
Use code, not image
Second use of IConfiguration format it as code.

>>services.AddSingleton(_ => Configuration);
Any reason you're using _ in lieu of more common 'c'?

ardalis:
It's very hard to respond to feedback given in this fashion - in the future use line-item comments on a PR, or group feedback into separate issues, or issue your own PR which can include line-by-line discussion of individual feedback items.

[...]

PR #6557, opened 2018-05-23
CC56654: Changing Link text

gosali:
Hello, @riande,

This proposed file change comes from https://github.com/aspnet/Docs.ko-kr/pull/62/files?utf8=%E2%9C%93&diff=split&w=1 . 
Line #78

Many thanks in advance.

  Fixes #Issue_Number


Rick-Anderson:
@gosali  this is the 2nd PR from loc where the customer takes a working link in the format we require and puts in the the full URL - something we don't want to do. Is there a reason they want to change the link? The link works for me. If the link is broken in ko-kr we should figure out why.

image

PR #6597, opened 2018-05-25
update SignalR getting started sample to RC1

anurse:
We missed a file in the earlier update! 😄 

Also, the ES5 version seemed to be out of date and was missing the `.start` call.


anurse:
Also removed the `wwwroot/lib/signalr/` folder since it's easy for it to get out of date, and none of the other samples have it checked in.

scottaddie:
@anurse Thank you for updating this.

PR #6620, opened 2018-05-26
Windows authorization migration

user135711:
Makes the Windows Authorization 2.1 sample compilable and adds a troubleshooting step.


guardrex:
Some of our topics just go along the lines of "new up a RP template app" (use the Windows auth switch in this case). The more samples we put in the repo, the longer the repo builds take, so I think it's a good idea to limit the use of samples where possible.

RE: Having a working sample ... We default to Razor Pages template-based sample apps. Not everywhere, and this might be one of those cases where something else is done. The *Program1x.cs* bits can probably just be inlined (or in a snapshot folder).

I defer to @Rick-Anderson and @scottaddie to make the call.

user135711:
@guardrex ok, I was curious to see the output, but it's such as simple app that it doesn't need to be compilable to be useful I think...

PR #6622, opened 2018-05-26
notes about HttpPost routing

jeremysimmons:
aspnet/Docs/#6447 add notes about conflict between `[HttpPost]` and `[EnableCors]`

  Fixes aspnet/Docs#6447

### DIFF
--- a/aspnetcore/security/cors.md
+++ b/aspnetcore/security/cors.md
@@ -88,7 +88,7 @@ To specify the CORS policy for a specific controller add the `[EnableCors]` attr
 [!code-csharp[](cors/sample/CorsMVC/Controllers/ValuesController.cs?name=EnableOnController)]
 
 > [!NOTE]
-> Actions with a route defined by `[HttpGet]`, `[HttpPost]`, `[HttpHead]`, `[HttpPut]`, and `[HttpDelete]` don't function as expected with `[EnableCors]` due to the way these attributes create routes that restrict the method specified in the name of the attribute. `[EnableCors]` must be used along with `[Route]` on the action, controller, or both. For more information on action constraints, see [Understanding IActionConstraint](xref:mvc/controllers/routing#understanding-iactionconstraint). 
+> Actions that use the `[HttpGet]` attribute, and similar `[Http-VERB]` attributes, can result in unexpected behavior when using `[EnableCors]`. When a CORS preflight is required, these routes may not be matched and result in a *404 Not Found*. See [Understanding IActionConstraint](xref:mvc/controllers/routing#understanding-iactionconstraint). Instead of using `[Http-VERB]` attributes, use `[Route]` with `[EnableCors]`.
 
 It's also possible to define routes using [HttpMethodRouteConstraint](/dotnet/api/microsoft.aspnetcore.routing.constraints.httpmethodrouteconstraint) when [defining non-attribute routes](xref:fundamentals/routing) to enforce the desired regular HTTP method and the OPTIONS HTTP method. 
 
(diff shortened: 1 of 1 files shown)


guardrex:
@jeremysimmons Thanks for working further on this. I've pinged for engineering to take a look. You may need to touch-up any inadvertent changes to meaning on my style update.

Just in passing tho, this remark ...

> due to the way these attributes create routes that restrict the method specified in the name of the attribute

... seems a bit cryptic. Can it be explained out in a few sentences what that means?

aspnetcore/security/cors.md
```diff
@@ -87,6 +87,11 @@ To specify the CORS policy for a specific controller add the `[EnableCors]` attr
 
 [!code-csharp[](cors/sample/CorsMVC/Controllers/ValuesController.cs?name=EnableOnController)]
 
+> [!NOTE]
+> Actions with a route defined by `[HttpGet]`, `[HttpPost]`, `[HttpHead]`, `[HttpPut]`, and `[HttpDelete]` don't function as expected with `[EnableCors]` due to the way these attributes create routes that restrict the method specified in the name of the attribute. `[EnableCors]` must be used along with `[Route]` on the action, controller, or both. For more information on action constraints, see [Understanding IActionConstraint](xref:mvc/controllers/routing#understanding-iactionconstraint).
```
Rick-Anderson: That sentence is way too long, you'll need to split it up.

> don't function as expected

can result in unexpected behavior.

jeremysimmons: @Rick-Anderson edits, per your request.

Original

> Actions with a route defined by `[HttpGet]`, `[HttpPost]`, `[HttpHead]`, `[HttpPut]`, and `[HttpDelete]` don't function as expected with `[EnableCors]` due to the way these attributes create routes that restrict the method specified in the name of the attribute. `[EnableCors]` must be used along with `[Route]` on the action, controller, or both. For more information on action constraints, see [Understanding IActionConstraint](xref:mvc/controllers/routing#understanding-iactionconstraint).

Update

Actions that use `[HttpGet]` attribute and similar `[Http-VERB]` attributes can result in unexpected behavior when using `[EnableCors]`. When a CORS preflight is required, these routes may not be matched, and result in a 404 Not Found. See [Understanding IActionConstraint](xref:mvc/controllers/routing#understanding-iactionconstraint). `[EnableCors]` can be used along with `[Route]`, instead of `[Http-VERB]` attributes. 

@guardrex 
> ... seems a bit cryptic. Can it be explained out in a few sentences what that means?

I had an action which used `[HttpPut]`. The method requires a Body with Content-Type application/json. That combination triggers a CORS preflight by the client. The server responded with 404 Not Found, because the `[Http-VERB]` attributes add a constraint for that verb, which causes the OPTIONS CORS preflight request to fail. 
I consider this behavior to fall outside the 'pit of success'. 
I believe that the docs advertise the following things, which are currently in conflict.
1. `[Http-VERB]` attributes as the preferred way to make Routes that are verb-specific
2. Adding `[EnableCors]` to existing Controllers/Actions will "Just work" (because there are no guard rails or warning signs)

This is solved  by either
- Updating the documentation with guard rails/warning signs (current task)
- Making [EnableCors] do more work, by registering a Route that allows the OPTIONS verb, in addition to the ones defined by `[Http-VERB]`
- Advertising a work-around which is a combination of `[Http-VERB]` and `[EnableCors]`  ( I mentioned a hypothetical class `HttpPostWithCorsAttribute` in aspnet/Docs#6447)
- some other clever fix.

guardrex: Thanks for explaining. The new text looks good.

Here's a cut-'n-paste that conforms to our grammar and style guidelines ...

> Actions that use the \`[HttpGet]\` attribute, and similar \`[Http-VERB]\` attributes, can result in unexpected behavior when using \`[EnableCors]\`. When a CORS preflight is required, these routes may not be matched and result in a \*404 Not Found\*. See \[Understanding IActionConstraint](xref:mvc/controllers/routing#understanding-iactionconstraint). Instead of using \`[Http-VERB]\` attributes, use \`[Route]\` with \`[EnableCors]\`.

guardrex: I'll go ahead and toss that in, and please feel free to apply additional updates.

Rick-Anderson:
@rynowak  2 minute review pls.

rynowak:
I realize that I'm late to the conversation, but I don't understand where is coming from. Lots of folks are using attribute routing and CORS successfully. 

The guidance provided in this PR is the opposite guidance we would give. I wouldn't recommend that anyone attempt to use `HttpMethodRouteContraint` for anything.

PR #6659, opened 2018-05-29
Add clarification to scenario in anti-request-forgery.md

mattbrooks2010:
Clarify that there are alternative scenarios a malicious site could implement, in addition to the example of a basic CSRF attack. The original content conveys the impression that the user is required to interact with the malicious site, which is presumably almost never the case when the user has JavaScript enabled.


Rick-Anderson:
@mattbrooks2010  Looks like a nice update.

blowdart:
LGTM

guardrex:
Thanks again @mattbrooks2010! 🏎

PR #6725, opened 2018-05-31
Merge to live 5-30 Part

Rick-Anderson:


guardrex:
@Rick-Anderson @scottaddie I think the problem is that I need to squash the commits on those moniker range PRs. The first one is probably blowing up before the second one can repair the damage. I'll see if I can squash them, and that might let this build.

PR #6730, opened 2018-05-31
Update gdpr.md

just-peteking:
I believe you mean, "EU" and not "UE".


scottaddie:
@just-peteking Nice catch! Thank you for the PR.

PR #6738, opened 2018-05-31
Revert "Repair Enforcing SSL topic moniker range"

guardrex:
Reverts aspnet/Docs#6712


guardrex:
@scottaddie Is it normal for a reversion at this point to add these other commits? It only built one commit, and the files changed is showing the correct number, three. That gives me some confidence. I didn't expect to see the commits external to this one; however, it might be reverting those prior to re-applying them.

My plan here is to revert this, revert https://github.com/aspnet/Docs/pull/6694, create a new, clean, valid PR for the changes.

guardrex:
Nevermind ... I pulled the commit, and my assumption about this being the cause was probably incorrect.

PR #6766, opened 2018-05-31
Move env var config down in Generic Host sample

### LINKED ISSUE #6765: Move env var config down in Generic Host sample
More closely matches `CreateDefaultBuilder` for the Web Host, where settings files go first, then env vars, followed by command line args.

guardrex:
Fixes #6765 
Cross-ref: https://github.com/aspnet/Hosting/issues/1440#issuecomment-393698716

More closely matches `CreateDefaultBuilder` for the Web Host, where settings files go first, then env vars, followed by command line args.

cc/ @chrisckc


guardrex:
Me, neither! We have some silly doc authors around here I'll tell ya! 😀

PR #6770, opened 2018-06-01
Update toc.md

Rick-Anderson:
@scottaddie  why don't we drop the qs `?view=aspnetcore-2.1` so it picks up the version from our TOC?


scottaddie:
@Rick-Anderson Without the querystring, "All APIs" is selected on the API ref landing page.
image

I pinged Dan about this yesterday. Eilon and Den were working to update our API ref to RTM. It's still on RC1. Should we hold off on this PR until that's done?

PR #6775, opened 2018-06-01
Fix link to metapackage

breyed:
The link in the note that was supposed to be to the new metapackage was instead to the same page that contained the note.


scottaddie:
@breyed Thank you for catching this and fixing it. Much appreciated.

PR #6803, opened 2018-06-02
2.1 Sample Updates (Round 5)

guardrex:
Addresses #5495 

Three more topics+samps ...

* fundamentals\servers\kestrel.md
* mvc\razor-pages\razor-pages-conventions.md
* performance\response-compression.md

The last one required a bit of snippet work. Might want to eyeball the built version: [Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/performance/response-compression?view=aspnetcore-1.1&branch=pr-en-us-6803).


aspnetcore/performance/response-compression.md
```diff
@@ -75,21 +77,26 @@ You can explore the features of the Response Compression Middleware with the [sa
 
 ## Package
 
-To include the middleware in your project, add a reference to the [`Microsoft.AspNetCore.ResponseCompression`](https://www.nuget.org/packages/Microsoft.AspNetCore.ResponseCompression/) package or use the [`Microsoft.AspNetCore.All`](https://www.nuget.org/packages/Microsoft.AspNetCore.All/) package. This feature is available for apps that target ASP.NET Core 1.1 or later.
+To include the middleware in your project, add a reference to the [Microsoft.AspNetCore.ResponseCompression](https://www.nuget.org/packages/Microsoft.AspNetCore.ResponseCompression/) package or use the [Microsoft.AspNetCore.App metapackage](xref:fundamentals/metapackage-app). This feature is available for apps that target ASP.NET Core 1.1 or later.
```
scottaddie: Since this doc is available for ASP.NET Core 1.1+, we need to provide more guidance in this paragraph. If targeting .NET Framework, install the Microsoft.AspNetCore.ResponseCompression package. If targeting .NET Core 2.0, install the .All metapackage. If targeting .NET Core 2.1, install the .App metapackage.

guardrex: We weren't saying "If targeting .NET Framework" that before because there were cases even in Core when the metapackage wasn't desired by devs. It was always supported to just use the package directly. Let me try some moniker-range blocks to see if I can address this.

scottaddie: I guess that makes sense. Thanks for the explanation.

PR #6826, opened 2018-06-02
Update route.md

davidich:
Remove invalid route example: '/files/myFile.' as the router doesn't match it.

[documentation page](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-2.1)


scottaddie:
Thank you for your contribution, @davidich.

Issue #5910, opened 2018-04-09
signalr/groups Manage Groups in ASP.NET Core SignalR


rachelappel:
Outline:

•	Add and remove users
(use User.Identity rather than connection id)
•	Message members of group
•	Store group membership in SQL Table
•	Store group membership in Azure Table
•	Verify group membership when reconnecting

anurse:
Here are the notes from the Groups and Users issue I filed and closed as a dupe of this.

* Groups
  * Use Groups to manage collections of connections you want to be able to broadcast to
  * It's not possible to enumerate or count the members of a group (because each server only knows about it's own connections, not those on other servers)
  * Group membership is not preserved when you reconnect
  * A connection can be a part of multiple groups
  * Groups are the recommended way for one connection to send to another connection (or connections)
  * Group names are case-sensitive
* Users
  * Use Users to manage all the connections associated with a single logical user
  * Adding a custom `IUserIdProvider` to DI allows you to change the "Identifier" used by SignalR (used in `Clients.User(...)` calls
  * User identifiers are case-sensitive.

PR #6861, opened 2018-06-04
 SignalR/Groups #5910

BrennanConroy:
#5910 
[Int Review Link](https://review.docs.microsoft.com/en-us/aspnet/core/signalr/groups?view=aspnetcore-2.1&branch=pr-en-us-6861)

This one needs a little work, soliciting early feedback. Might be worth mentioning the `Users`, `Groups`, `GroupExcept`, and `OthersInGroup` methods.


aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,77 @@
...
+ms.author: 
+ms.custom: mvc
+ms.date: 06/04/2018
+ms.prod: aspnet-core
+ms.technology: aspnet
+ms.topic: article
+uid: signalr/groups
+---
+
+# Manage users and groups in SignalR
```
anurse: nit: Manag**ing**

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,77 @@
...
+
```
anurse: Should have some intro blurb about what users and groups are.

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,77 @@
...
+## Users?
+
+A user is one or more connections associated with an identifier. By default SignalR will use the `ClaimTypes.NameIdentifier` from the `ClaimsPrincipal` associated with the connection. There can be multiple connections for a single user. For example you can be connected on your desktop, your phone, and your laptop. If you send to the user all three connections will receive the message.
```
anurse: I would avoid treating a "user" as a concrete object here. So instead of "A user is one or more connections..." something like "SignalR allows you to send messages to all connections associated with a specific User. By default, SignalR will use the (blah blah claims) associated with the connection as the user identifier. Multiple connections can be associated with the same user. For example, ..."

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,77 @@
...
+
+You can send a message to a specific user by passing the user identifier to the `User(...)` function in your hub method as shown in the following example. One important aspect to note is that the user identifier is case-sensitive.
```
anurse: Link to API docs! Check if there's a way to do that "semantically" (i.e. other than just grabbing a URL like: https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.signalr.hubcallerclients.user?view=aspnetcore-2.1)

anurse: ~~One important aspect to note is that~~ **Note:** The user identifier is case-sensitive

Also I think there's a way to have a note call-out in the special markdown syntax docs uses.

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+The user identifier can be customized by creating your own `IUserIdProvider`.
+
+```csharp
+public class CustomUserIdProvider : IUserIdProvider
+{
+    public virtual string GetUserId(HubConnectionContext connection)
+    {
+        // Be careful when using ClaimTypes.Name.
```
anurse: Use a different claim type in the example then. Maybe Email?

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,77 @@
...
+        // If your app allows non-unique names then the identifier will apply to everyone that uses that name.
+        return connection.User?.FindFirst(ClaimTypes.Name)?.Value;
+    }
+}
+```
+
+And injecting it into the Apps DI container.
```
anurse: Strike this

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,77 @@
...
+You can send a message to a specific user by passing the user identifier to the `User(...)` function in your hub method as shown in the following example. One important aspect to note is that the user identifier is case-sensitive.
+
+```csharp
+public Task SendPrivateMessage(string user, string message)
+{
+    return Clients.User(user).SendAsync("ReceiveMessage", message);
+}
+```
+
+The user identifier can be customized by creating your own `IUserIdProvider`.
```
anurse: The user identifier can be customized by creating your own `IUserIdProvider`, and registering it in `ConfigureServices`

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,77 @@
...
+        // If your app allows non-unique names then the identifier will apply to everyone that uses that name.
+        return connection.User?.FindFirst(ClaimTypes.Name)?.Value;
+    }
+}
+```
+
+And injecting it into the Apps DI container.
+
+```csharp
+services.AddSingleton(typeof(IUserIdProvider), typeof(CustomUserIdProvider));
```
anurse: Expand this snippet out to include the context and use a comment to illustrate omitted bits. Also use generics.

```csharp
public void ConfigureService(IServiceCollection services)
{
    // ... other services ...
    services.AddSingleton<IUserIdProvider, CustomUserIdProvider>();
}
```

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+## Groups?
+
+A group is a collection of connections that can be broadcast to. Groups are the recommended way to send to a connection or multiple connections because the groups are managed by the application. A connection can be subscribed to multiple groups, one example would be for a chat app to have a group per chat room so only connections interested in certain rooms would be a part of those groups. Group management can be done in the hub via the `AddToGroupAsync` and `RemoveFromGroupAsync` methods.
```
anurse: "broadcast to" isn't really a verb we should use outside the team. It's a bit esoteric IMO. 

A groups is collection of connections associated with a name. Messages can be sent to all connections in a group. Groups are the recommended way ... A connection can be a member of multiple groups. This makes groups ideal for something like a chat application, where each room can be represented as a Group. Connections can be added to or removed from groups via the `AddToGroupAsync` and `RemoveFromGroupAsync` methods.

(also, API links!)

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,77 @@
...
+
+public async Task RemoveFromGroup(string groupName)
+{
+    await Groups.RemoveFromGroupAsync(Context.ConnectionId, groupName);
+
+    await Clients.Group(groupName).SendAsync("Send", $"{Context.ConnectionId} has left the group {groupName}.");
+}
+```
+
+Group membership is not preserved when a connection reconnects, the connection will need to resubscribe or the aplication will need to keep track of each connections subscriptions. It is also not possible to count the number of members per group as the app can be scaled out and wont know about the other servers subscription information. Just like user identifiers, group names are case-sensitive.
```
anurse: Group membership is not preserved when a connection reconnects, the connection will need to ~~resubscribe or the aplication will need to keep track of each connections subscriptions.~~ **rejoin the group when it is re-established**.

Also "scaled out" isn't a verb ;). Perhaps "It is not possible to enumerate the members of a group, since this information isn't available if the application is scaled to multiple servers"

aspnetcore/signalr/groups.md
rachelappel: Can't use Managing. It's a gerund and docs guidelines say no gerunds.

aspnetcore/signalr/groups.md
rachelappel: It's:

> [!NOTE]
> note text note text note text

aspnetcore/signalr/groups.md
anurse: I sit corrected 😄

BrennanConroy:
Should 'user' and 'group' be capitalized in most cases?

anurse:
I was debating that internally myself while reviewing this. I lean towards **not** capitalizing it I think, as it's not a proper noun.

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,64 @@
...
+
+By [Brennan Conroy](https://github.com/BrennanConroy)
+
+Users and groups is how SignalR exposes connection management.
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/signalr/groups/sample/) [(how to download)](xref:tutorials/index#how-to-download-a-sample)
+
+## What are users?
+
+SignalR allows you to send messages to all connections associated with a specific user. By default SignalR will use the `ClaimTypes.NameIdentifier` from the `ClaimsPrincipal` associated with the connection as the user identifier. Multiple connections can be associated with the same user. For example, you can be connected on your desktop, your phone, and your laptop. If you send to the user all three connections will receive the message.
```
rachelappel: will use -> uses
If you send to the user all three connections will receive the message. -> If a message is sent to the user all three connections will receive the message.

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,64 @@
...
+
+You can send a message to a specific user by passing the user identifier to the `User(...)` function in your hub method as shown in the following example.
```
rachelappel: You can send a message to a specific user by passing the user identifier to the User(...) function in your hub method as shown in the following example. ->
Send a message to a specific user by passing the user identifier to the `User` function in your hub method as shown in the following example:

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,64 @@
...
+> The user identifier is case-sensitive.
+
+```csharp
+public Task SendPrivateMessage(string user, string message)
+{
+    return Clients.User(user).SendAsync("ReceiveMessage", message);
+}
+```
+
+The user identifier can be customized by creating your own `IUserIdProvider`, and registering it in `ConfigureServices`.
```
rachelappel: your own -> an

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,64 @@
...
+
+[!code-csharp[UserIdProvider](groups/sample/customuseridprovider.cs?range=4-10)]
+
+[!code-csharp[Configure service](groups/sample/startup.cs?range=21-22,39-42)]
+
+> [!NOTE]
+> AddSignalR() must be called before registering your custom SignalR services.
```
rachelappel: AddSignalR() -> `AddSignalR`

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,64 @@
...
+
+## What are groups?
+
+A group is a collection of connections associated with a name. Messages can be sent to all connections in a group. Groups are the recommended way to send to a connection or multiple connections because the groups are managed by the application. A connection can be a member of multiple groups. This makes groups ideal for something like a chat application, where each room can be represented as a group. Connections can be added to or removed from groups via the `AddToGroupAsync` and `RemoveFromGroupAsync` methods.
+
+[!code-csharp[Hub methods](groups/sample/hubs/chathub.cs?range=15-27)]
+
+Group membership is not preserved when a connection reconnects, the connection will need to rejoin then group when it is re-established. It is not possible to count the members of a group, since this information is not available if the application is scaled to multiple servers.
```
rachelappel: Group membership is not preserved when a connection reconnects, the connection will need to rejoin then group when it is re-established. It is not possible to count the members of a group, since this information is not available if the application is scaled to multiple servers.

Group membership isn't preserved when a connection reconnects. The connection needs to rejoin then group when it's re-established. It's not possible to count the members of a group, since this information is not available if the application is scaled to multiple servers.

BrennanConroy:
Ready for final review

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,64 @@
...
+[!code-csharp[UserIdProvider](groups/sample/customuseridprovider.cs?range=4-10)]
+
+[!code-csharp[Configure service](groups/sample/startup.cs?range=21-22,39-42)]
+
+> [!NOTE]
+> AddSignalR must be called before registering your custom SignalR services.
+
+## What are groups?
+
+A group is a collection of connections associated with a name. Messages can be sent to all connections in a group. Groups are the recommended way to send to a connection or multiple connections because the groups are managed by the application. A connection can be a member of multiple groups. This makes groups ideal for something like a chat application, where each room can be represented as a group. Connections can be added to or removed from groups via the `AddToGroupAsync` and `RemoveFromGroupAsync` methods.
```
DylanDmitri: > Groups are the recommended way to send to a connection or multiple connections because the groups are managed by the application. 

> Groups are the recommended way of sending messages, even to individual users. This is because groups, being managed by the application, are more <*robust and modular*>.

DylanDmitri: but eplace the <*robust and modular*> with something more accurate

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,64 @@
...
+ms.technology: aspnet
+ms.topic: article
+uid: signalr/groups
+---
+
+# Manage users and groups in SignalR
+
+By [Brennan Conroy](https://github.com/BrennanConroy)
+
+Users and groups is how SignalR exposes connection management.
```
mikaelm12: nit: "This sounds a little weird to me. Maybe "is how" ->  "are how"

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,64 @@
...
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/signalr/groups/sample/) [(how to download)](xref:tutorials/index#how-to-download-a-sample)
+
+## What are users?
+
+SignalR allows you to send messages to all connections associated with a specific user. By default SignalR uses the `ClaimTypes.NameIdentifier` from the `ClaimsPrincipal` associated with the connection as the user identifier. Multiple connections can be associated with the same user. For example, you can be connected on your desktop, your phone, and your laptop. If a message is sent to the user all three connections will receive the message.
```
mikaelm12: > Multiple connections can be associated

I think this is good. Maybe emphasize that there will be multiple connection ids for a single user in this scenario

mikaelm12: And that sending via a connection ID might result in a single user getting a message in place and not others

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Manage users and groups in SignalR
+
+By [Brennan Conroy](https://github.com/BrennanConroy)
+
+Users and groups is how SignalR exposes connection management.
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/signalr/groups/sample/) [(how to download)](xref:tutorials/index#how-to-download-a-sample)
+
+## What are users?
```
mikaelm12: nit: Maybe make this "SignalR Users", "Users in SignalR" etc. Just a though.
Just sounds odd to me as "What are users"

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,64 @@
...
+The user identifier can be customized by creating an `IUserIdProvider`, and registering it in `ConfigureServices`.
+
+[!code-csharp[UserIdProvider](groups/sample/customuseridprovider.cs?range=4-10)]
+
+[!code-csharp[Configure service](groups/sample/startup.cs?range=21-22,39-42)]
+
+> [!NOTE]
+> AddSignalR must be called before registering your custom SignalR services.
+
+## What are groups?
```
mikaelm12: See comment above about Users. This one doesn't seem as odd to me. But, if you think it sounds good, I'm fine with keeping it as well

aspnetcore/signalr/groups/sample/Hubs/ChatHub.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Microsoft.AspNetCore.SignalR;
+using System;
```
mikaelm12: nit: Ordering 😄

aspnetcore/signalr/groups.md
anurse: SignalR allows messages to be sent to named groups of connections, as well as to all connections associated with a specific user.

aspnetcore/signalr/groups.md
anurse: What even are users?

I like "Users in SignalR"

aspnetcore/signalr/groups.md
anurse: SignalR allows you to send messages to all connections associated with a specific user. By default SignalR uses the `ClaimTypes.NameIdentifier` from the `ClaimsPrincipal` associated with the connection as the user identifier. For example, you can be connected on your desktop, your phone, and your laptop. If a message is sent to the user all three connections will receive the message.

->

A single user can have multiple connections to a SignalR server. For example, a user could be connected on their desktop as well as their phone. Each device has a separate SignalR connection, but they are all associated with the same user. If a message is sent to the user, all of the connections associated with that user will receive the message.

aspnetcore/signalr/groups.md
anurse: Part of that *<robust and modular>* is that Connection IDs are only valid on the server that generated them, so they should never be used from the client or other servers. Groups are recommended even for single connections because of this.

aspnetcore/signalr/groups.md
```diff
@@ -0,0 +1,64 @@
...
+ms.technology: aspnet
+ms.topic: article
+uid: signalr/groups
+---
+
+# Manage users and groups in SignalR
+
+By [Brennan Conroy](https://github.com/BrennanConroy)
+
+SignalR allows messages to be sent to named groups of connections, as well as to all connections associated with a specific user.
```
mikaelm12: Super nit: In this overview you mention Groups then Users, but the order of the sections you have is Users then Groups

rachelappel:
LGTM

PR #6878, opened 2018-06-05
C19396: Missing hyperlink due to incorrect redenring

carolinacmoravia:
Hello, @Rick-Anderson,

Localization team has reported source content issue that causes localized version to have broken format compared to en-us version.  
From the context of this source string, we can assume that writer did not intend this example URL to be rendered as hyperlink, and that text "App Namespace" within brackets {} is expected to be translated.

We found out that OPS is using a Markdig flavor called "Markdig (DocFX)" which does render URL surrounded by space to be rendered as hyperlink. 

To avoid this, suggested fix to source string is add an escape char (\) to part of the URL.

Many thanks in advance.

  Fixes #Issue_Number

### DIFF
--- a/aspnet/mvc/overview/security/create-an-aspnet-mvc-5-app-with-facebook-and-google-oauth2-and-openid-sign-on.md
+++ b/aspnet/mvc/overview/security/create-an-aspnet-mvc-5-app-with-facebook-and-google-oauth2-and-openid-sign-on.md
@@ -169,7 +169,7 @@ For Facebook OAuth2 authentication, you need to copy to your project some settin
     ![Create new app](create-an-aspnet-mvc-5-app-with-facebook-and-google-oauth2-and-openid-sign-on/_static/image22.png)
 4. Enter an **App Name** and **Category**, then click **Create App**.
 
-    This must be unique across Facebook. The <strong>App Namespace</strong> is the part of the URL that your App will use to access the Facebook application for authentication (for example, https://apps.facebook.com/{App Namespace}). If you don't specify an <strong>App Namespace</strong>, the <strong>App ID</strong> will be used for the URL. The <strong>App ID</strong> is a long system-generated number that you will see in the next step.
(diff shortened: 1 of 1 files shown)


guardrex:
Hi @carolinacmoravia ... On our repo, the fix is to code-fence the URL. We normally do this ...

```
(for example, `https://apps.facebook.com/{App Namespace}`).
```

v-jacard:
Hi @guardrex, cc @Rick-Anderson, @carolinacmoravia,
We also considered using inline code syntax for the URL, but in that case, the text "App Namespace" within brackets {} will be locked from localization, so translator will not be able to translate that text. 
So if you would like the text "App Namespace" within brackets {} to be translated, then we would proposed to use the escape character within the URL, so it would not be rendered as hyperlink and still be exposed to localization.

guardrex:
I see. We've been code-fencing URLs for quite some time, but perhaps we need a special rule for these URLs that have segments we would like to localize.

On a related note, I wonder about segments of code-fenced commands. I was just working with a few in another topic, where I needed to indicate "service name" and "service path" placeholders, so I composed it the way we normally do ...

> \`sc create <SERVICE_NAME> binpath= "<SERVICE_PATH>"\`

`sc create <SERVICE_NAME> binpath= "<SERVICE_PATH>"`

If we wanted to localize the parts the reader would replace, we'd drop out of and into code-fencing ...

> \`sc create <\`SERVICE NAME\`> binpath= "<\`SERVICE PATH\`>"\`

... which would render as ...

`sc create <`SERVICE NAME`> binpath= "<`SERVICE PATH`>"`

Looks kind'a funky, which begs the question if it should be ...

> \`sc create <\`\<code>SERVICE NAME\</code>\`> binpath= "<\`\<code>SERVICE PATH\</code>\`>"\`


`sc create <`<code>SERVICE NAME</code>`> binpath= "<`<code>SERVICE PATH</code>`>"`

🤦‍♂️ ... Even if localization would run on those placeholders, **they'll never agree to that!** :smile: lol

I have a side-question about this pertaining to commands on localized systems: Our friends in Mexico for example ... do they type "sc create ..." or do they type something like "sc crear ..."? Are the command arguments localized?

v-jacard:
@guardrex, @Rick-Anderson, cc @carolinacmoravia, @ishasahni1 for visibility,
Command arguments within code block are usually not localized.
It would not look nice to try break up code block to localize parts within code block, so I would recommend to leave code blocks as they are.

Issue #6538, opened 2018-05-22
Expand scope of the Enforce HTTPS article


shirhatti:
Our current Enforce HTTPS article isn't sufficient for folks looking to enforce HTTPS on their deployment. While all the content lives in our docs, we don't do a good enough job of cross-linking to relevant content.

Here's my proposed additions to the TOC

- [x] Setting up HTTPS
  - [x] Cross-link to existing server specific guidance on enabling HTTPS
  - [x] Cross-link to the proxy guidance especially around validating forwarded headers **Addressed by #6644.**
  - ALLOWED_HOSTS and HTTPS_PORT need to be set
- [x] Enforcing HTTPS in Azure App Service
  - [x] Antares has a portal experience where you can check a box to enable HTTPS which is what we should guide people towards
- [x] Enforcing HTTPS behind other proxies
- [x] Surface IIS hosting issues in the IIS topic pertaining to use of the middleware when the site isn't configured for HTTPS in IIS. For example, see https://github.com/aspnet/Home/issues/3176.

Rick-Anderson:
Here's my proposed additions to the ~TOC~ article layout

@shirhatti  can you supply the content for **Enforcing HTTPS behind other proxies**? If so, I can prioritize this.

cc @guardrex  - who could probably knock off the link addition pretty quickly.

guardrex:
Sure, but can "pretty quickly" be after my 2.1 issues are resolved? I can do it asap after that.

Rick-Anderson:
@guardrex  I meant to say, after he's done with 2.1, I don't think it will take him long to add those links.

guardrex:
*From @Praveen-Rai on #9052 ...*

This has been missing link in our Hosting section.

Few references :
aspnet/KestrelHttpServer#2902
https://stackoverflow.com/questions/52227372/asp-net-core-application-service-only-listening-to-port-5000-on-ubuntu
https://www.digitalocean.com/community/tutorials/how-to-secure-nginx-with-let-s-encrypt-on-ubuntu-16-04

@guardrex I think it's time to add SSL configuration to our documentation. Let'sEncrypt should be a good CA for SSL certificates for our documentation.

Also, below pages will need the details on the same :
https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/linux-apache?view=aspnetcore-2.1&tabs=aspnetcore2x
[...]

PR #6886, opened 2018-06-05
add essential step in NGINX tutorial

user135711:


guardrex:
Thanks @user135711 for mentioning it.

This reminds me that I'll need to address the same scenario for full IIS hosting.

We'll probably look at how this is phrased. It probably should go along the lines of ...

> When using a \`dotnet new\` template and not \[configuring Nginx for HTTPS](#configure-ssl), disable the \[HTTPS Redirection Middleware](xref:security/enforcing-ssl) (\`app.UseHttpsRedirection();\`) in \`Startup.Configure\`. Either remove the line or comment it out. If the middleware is active but Nginx can't forward the request because it isn't configured for HTTPS, the web app won't be reachable after the redirect to the secure endpoint.

... something like that.

user135711:
@guardrex The second sentence (If the middleware is active) needs to be clearer because non experts like me need to understand it. By Nginx isn't configured to use https you mean that the server block passes to the local app with the directive `proxy_pass         http://localhost:5000/;` which has it's own redirect to https://localhost:5000 which will then return to NGINX (this step is what makes it a reverse proxy) which then will use the default proxy_redirect command, which reverses the location header with the unwritten default command `proxy_redirect http://localhost:5001 /;` It takes the second argument as from the location and in the server block location appended to the $host_proxy and exchanges it to be the original $host. Since an asymmetric reverse step comes from a secure host, you could still do it but you need to use `proxy_redirect https://localhost:5001 /;`

guardrex:
Indeed ... it is best if you and @shirhatti work out the language, and I'll just handle the grammar+style guidelines part.

A bit unrelated to your issue, but I was just looking at how the middleware attempts to determine the secure port. Apparently when IIS doesn't have a secure port for the app, the ANCM forwards that fact via the `IServerAddressesFeature`'s binding address. Without a secure port, the middleware no-ops.

user135711:
How does this version sound?
> When using a `dotnet new` template and not [configuring Nginx for HTTPS](#configure-ssl), disable the [HTTPS Redirection Middleware](xref:security/enforcing-ssl) (`app.UseHttpsRedirection();`) in `Startup.Configure`. Either remove the line or comment it out. If the middleware is active but Nginx  forwards the request to an insecure localhost, the web app won't be reachable after the redirect to the secure endpoint because the reverse redirection to the public address won't occur. Only host production apps which are not on the same server or internal network as Nginx using [Nginx configured for HTTPS](#configure-ssl).

guardrex:
Sounds good to me ... let's see what `@shirhatti` says. You can update the PR with that text if you like.

shirhatti:
I'm not a fan of this.
If you **choose** to disable HTTPS of your own volition, I'd rather you start out with `dotnet new webapp --no-https` rather than going down this approach

shirhatti:
That being said, we still need to document everything it entails to run behind a proxy (#6538)

guardrex:
@shirhatti Does the Nginx and the middleware configuration handle the secure port in a similar fashion to IIS? ...

1. No secure port is configured for the app in Nginx
1. `IServerAddressesFeature` addresses don't have a bound address with an HTTPS scheme
1. The middleware no-ops

guardrex:
@user135711 I'm going to close on @shirhatti's advice. Thanks anyway for opening the PR and the discussion.

PR #6893, opened 2018-06-06
missing capture group in regex

user135711:
The URI captured was not re-added in the regex replacement. See [example](https://wiki.apache.org/httpd/RewriteHTTPToHTTPS). I also tested this in Apache to confirm.


shirhatti:
@user135711 Good catch! Thanks for the contribution.

PR #6895, opened 2018-06-06
Update razor-pa

malhaar-p:
> change from webapp to web, as there is no template with name in dotnet core version 2.1.2


scottaddie:
@malhaar-p The "webapp" alias does work. You'll see the "razor" short name in the CLI, but this is an alias for that project template. Please run `dotnet --version` to confirm that you have the 2.1 RTM SDK installed.

PR #6901, opened 2018-06-06
Use AddHostedService for hosted service registration in background tasks topic

### LINKED ISSUE #6900: Use AddHostedService for hosted service registration
In the background tasks topic and sample, swap `AddHostedService` for the singleton registration of hosted services. Note this is a 2.1-only scenario.

guardrex:
Fixes #6900

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services?view=aspnetcore-2.0&branch=pr-en-us-6901)

### DIFF
--- a/aspnetcore/fundamentals/host/hosted-services/samples-snapshot/2.x/BackgroundTasksSample-WebHost/Startup.cs
+++ b/aspnetcore/fundamentals/host/hosted-services/samples-snapshot/2.x/BackgroundTasksSample-WebHost/Startup.cs
@@ -1,8 +1,6 @@
 using Microsoft.AspNetCore.Builder;
 using Microsoft.Extensions.DependencyInjection;
 using BackgroundTasksSample.Services;
-using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Hosting;
 
 namespace BackgroundTasksSample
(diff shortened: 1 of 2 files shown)


aspnetcore/fundamentals/host/hosted-services/samples-snapshot/2.x/BackgroundTasksSample-WebHost/Startup.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace BackgroundTasksSample
+{
+    public class Startup
+    {
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddMvc();
+
+            #region snippet1
```
davidfowl: Why aren't we using AddHostedService in here?

guardrex: It's a 2.1 thing, right? The "samples-snapshot" provides the doc with the 2.0 approach when the 2.0 version of the doc is loaded.

davidfowl: Ah ok yes this is 2.1

PR #6929, opened 2018-06-07
Wrong reference to IFilter instead of IFilterMetadata

Ricciolo:


guardrex:
Thanks for this update @Ricciolo. I've requested engineering approval.

Issue #6806, opened 2018-06-02
SignalR: Information about streaming messages


JesperTreetop:
There's no information about how streaming messages work - how to subscribe, how to unsubscribe and so on.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 3cf283c9-5cd2-a9ff-b8c8-372d8a04356d
* Version Independent ID: 729fddcf-24e4-b563-4952-a8045b231868
* Content: [ASP.NET Core SignalR JavaScript client](https://docs.microsoft.com/en-us/aspnet/core/signalr/javascript-client?view=aspnetcore-2.1)
* Content Source: [aspnetcore/signalr/javascript-client.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/signalr/javascript-client.md)
* Product: **aspnet-core**
* GitHub Login: @rachelappel
* Microsoft Alias: **rachelap**

Rick-Anderson:
@anurse please comment/prioritize

anurse:
Yeah, this would be a good thing to prioritize. I'd suggest P2 depending on how much feedback we get. We're still lacking docs for the core cross-cutting scenarios so I think those still need to be the focus right now. The team would probably need to help produce a sample and some copy (in outline form) for this.

dvdmrk:
I've tried literally everything @anurse, you are my only hope. 
These are the errors I'm getting based off of an angular build using the latest vs code, node, core, and signalr. I've built a pretty generic chat application, and I'm not sure where to go from here.
Angular is running in the development mode. Call enableProdMode() to enable the production mode.
/home/echo/negotiate:1 Failed to load resource: net::ERR_CONNECTION_REFUSED
Utils.js:151 Warning: Error from HTTP request. 0: 
ConsoleLogger.log @ Utils.js:151
Utils.js:148 Error: Failed to complete negotiation with the server: Error
ConsoleLogger.log @ Utils.js:148
Utils.js:148 Error: Failed to start the connection: Error
ConsoleLogger.log @ Utils.js:148
core.js:1448 ERROR Error: Uncaught (in promise): Error
Error
    at new HttpError (Errors.js:16)
    at XMLHttpRequest.xhr.onerror [as __zone_symbol__ON_PROPERTYerror] (HttpClient.js:80)
    at XMLHttpRequest.wrapFn (zone.js:1166)
    at ZoneDelegate.invokeTask (zone.js:421)
[...]

PR #6931, opened 2018-06-07
signalr/streaming #6806

BrennanConroy:
https://github.com/aspnet/Docs/issues/6806

[Int Review link](https://review.docs.microsoft.com/en-us/aspnet/core/signalr/streaming?view=aspnetcore-2.1&branch=pr-en-us-6931)


aspnetcore/signalr/streaming.md
```diff
@@ -0,0 +1,63 @@
...
+ms.technology: aspnet
+ms.topic: article
+uid: signalr/streaming
+---
+
+# Use streaming in ASP.NET Core SignalR
+
+By [Brennan Conroy](https://github.com/BrennanConroy)
+
+Streaming is a way for clients to subscribe to a stream of data from the server on demand.
```
rachelappel: This sentence works better up here in the intro area:

Streaming is useful for scenarios when you don't have all the data to send to the client yet, but will be getting it over time and would like to update the client whenever new data is available.

aspnetcore/signalr/streaming.md
```diff
@@ -0,0 +1,63 @@
...
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/live/aspnetcore/signalr/streaming/sample) ([how to download](xref:tutorials/index#how-to-download-a-sample))
+
+## Set up the hub
+
+A hub method automatically becomes a streaming hub method when it returns a variant of `ChannelReader`. For example, return values of `ChannelReader<int>` and `Task<ChannelReader<string>>` are streaming methods. Streaming is useful for scenarios when you don't have all the data to send to the client yet, but will be getting it over time and would like to update the client whenever new data is available. Below is a sample that shows the basics of streaming data to the client. Everytime data is written to the `ChannelReader` the client will be notified of the new data and at the end the `ChannelReader` is completed to tell the client that the stream is closed.
```
rachelappel: will be getting -> get 
would like -> want
the client will be notified - >the client is notified
Everytime data is written to the `ChannelReader` the client will be notified of the new data and at the end the `ChannelReader` is completed to tell the client that the stream is closed. 
->
Everytime data is written to the `ChannelReader` the client is notified of the new data. At the end, the `ChannelReader` is completed to tell the client the stream is closed.

aspnetcore/signalr/streaming.md
```diff
@@ -0,0 +1,63 @@
...
+
+[!code-javascript[Streaming hub method](streaming/sample/hubs/streamhub.cs?range=10-29)]
+
+## .NET client
+
+`StreamAsChannelAsync` on the `HubConnection` calls streaming methods on the hub. Pass the hub method name and any arguments defined in the hub method to `StreamAsChannelAsync`. Set the generic in `StreamAsChannelAsync` based on the type returned from the streaming method. A `ChannelReader<T>` is returned from the stream invocation and represents the stream on the client. To read data, a common pattern is to loop over `WaitToReadAsync` and call `TryRead` when data is available. The loop will end when the stream has been closed by the server or the cancellation token passed to `StreamAsChannelAsync` is canceled.
```
rachelappel: method name and any arguments  -> method name, and arguments

rachelappel: invocation and represents -> invocation, and represents

rachelappel: server or the cancellation -> server, or the cancellation

aspnetcore/signalr/streaming.md
```diff
@@ -0,0 +1,63 @@
...
+}
+
+Console.WriteLine("Streaming completed");
+```
+
+## JavaScript client
+
+JavaScript clients call streaming methods on hubs by using `connection.stream`. The `stream` method accepts two arguments:
+
+* The name of the hub method. In the following example, the hub name is `Counter`.
```
rachelappel: In the following example, the hub name is Counter -> I think this should be "In the following example, the hub method name is Counter" because you mean the method name, correct? If not, ignore.

aspnetcore/signalr/streaming.md
```diff
@@ -0,0 +1,63 @@
...
+* Any arguments defined in the hub method. In the following example, the arguments are, a count for the number of stream items to receive and the delay between stream items.
```
rachelappel: Any arguments -> Arguments
the arguments are, -> the arguments are:

rachelappel: receive and the delay -> receive, and the delay

aspnetcore/signalr/streaming.md
```diff
@@ -0,0 +1,63 @@
...
+
+`connection.stream` returns an `IStreamResult` which has a method `subscribe` on it. Pass an `IStreamSubscriber` to `subscribe` and set the `next`, `error`, and `complete` callbacks to get notifications from the `stream` invocation.
```
rachelappel: IStreamResult which has a method subscribe on it -> `IStreamResult`, with its `subscribe` method. 
or 
`IStreamResult` which contains a `subscribe` method.

aspnetcore/signalr/streaming.md
```diff
@@ -0,0 +1,65 @@
...
+ms.technology: aspnet
+ms.topic: article
+uid: signalr/streaming
+---
+
+# Use streaming in ASP.NET Core SignalR
+
+By [Brennan Conroy](https://github.com/BrennanConroy)
+
+Streaming is useful for scenarios when you don't have all the data to send to the client yet, but will get it over time and want to update the client whenever new data is available.
```
anurse: Let's start by explaining what "streaming" means here :).

> ASP.NET Core SignalR supports streaming return values of server methods. This is useful for scenarios where fragments of data will come in over time. When a return value is streamed to the client, each fragment is sent to the client as soon as it is available, rather than waiting for all the data to become available.

Also, I want to feel out this blurb to see how people think. It's a useful analogy if you're familiar with these C# concepts:

----

A streaming method is the SignalR equivalent of a C# iterator (using `yield` syntax). When calling methods locally, if the method returns an `IList<T>`, all the items must be fetched and put into the list before a return value can be produced. However, if the method returns `IEnumerable<T>` and uses the `yield` keyword in C#, items are made available to the caller right away. Consider the following example program:

```csharp
using System;
using System.Collections.Generic;
using System.Threading;

public class Program {
    public static IList<int> BlockingCounter()
    {
        var items = new List<int>();
        for (var i = 0; i < 3; i++)
        {
            Thread.Sleep(1000);
            items.Add(i);
        }
        return items;
    }

    public static IEnumerable<int> YieldingCounter()
    {
        for (var i = 0; i < 3; i++)
        {
            Thread.Sleep(1000);
            yield return i;
        }
    }

    public static void Main()
    {
        Console.WriteLine(DateTime.Now);
        foreach (var item in BlockingCounter())
        {
            Console.WriteLine($"Blocking: {item}");
            Console.WriteLine(DateTime.Now);
        }

        foreach (var item in YieldingCounter())
        {
            Console.WriteLine($"Yielding: {item}");
            Console.WriteLine(DateTime.Now);
        }
    }
}
```

When this program runs, the output will be something like this:

```
6/9/2018 8:57:47 AM
Blocking: 0
6/9/2018 8:57:50 AM
Blocking: 1
6/9/2018 8:57:50 AM
Blocking: 2
6/9/2018 8:57:50 AM
Yielding: 0
6/9/2018 8:57:51 AM
Yielding: 1
6/9/2018 8:57:52 AM
Yielding: 2
6/9/2018 8:57:53 AM
```

For the `Blocking` variant, there is a delay of 3 seconds before the first item is written, and then all three items are returned at once. For the `Yielding` variant, the first item is returned 1 second after calling the method and then each item comes a second after the previous item.

This is similar to how streaming works in SignalR. In SignalR, **even if you use `yield`**, the server will wait for all the items to be returned before sending anything to the client, but when you use the streaming pattern, you can return items as soon as they become available.

aspnetcore/signalr/streaming.md
```diff
@@ -0,0 +1,65 @@
...
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/live/aspnetcore/signalr/streaming/sample) ([how to download](xref:tutorials/index#how-to-download-a-sample))
+
+## Set up the hub
+
+A hub method automatically becomes a streaming hub method when it returns a variant of `ChannelReader`. For example, return values of `ChannelReader<int>` and `Task<ChannelReader<string>>` are streaming methods. Below is a sample that shows the basics of streaming data to the client. Everytime data is written to the `ChannelReader` the client is notified of the new data. At the end, the `ChannelReader` is completed to tell the client the stream is closed.
```
anurse: when it returns a ~~variant of `ChannelReader`~~  **`ChannelReader<T>` or a `Task<ChanneReader<T>>`**

anurse: ~~Everytime data~~ **Whenever an object** is written to the `ChannelReader` ~~the client is notified of the new data~~ **that object is immediately sent to the client**.

aspnetcore/signalr/streaming.md
```diff
@@ -0,0 +1,65 @@
...
+
+[!code-javascript[Streaming hub method](streaming/sample/hubs/streamhub.cs?range=10-29)]
+
+## .NET client
+
+`StreamAsChannelAsync` on the `HubConnection` calls streaming methods on the hub. Pass the hub method name, and arguments defined in the hub method to `StreamAsChannelAsync`. Set the generic in `StreamAsChannelAsync` based on the type returned from the streaming method. A `ChannelReader<T>` is returned from the stream invocation, and represents the stream on the client. To read data, a common pattern is to loop over `WaitToReadAsync` and call `TryRead` when data is available. The loop will end when the stream has been closed by the server, or the cancellation token passed to `StreamAsChannelAsync` is canceled.
```
anurse: `StreamAsChannelAsync` on the `HubConnection` calls streaming methods on the hub.

->

The `StreamAsChannelAsync` method on `HubConnection` is used to invoke a streaming method.

anurse: Set the generic in `StreamAsChannelAsync` based on the type returned from the streaming method.

->

The generic parameter to `StreamAsChannelAsync` specifies the type of objects returned by the streaming method.

aspnetcore/signalr/streaming.md
```diff
@@ -0,0 +1,65 @@
...
+
+```csharp
+var channel = await hubConnection.StreamAsChannelAsync<int>("Counter", 10, 500, CancellationToken.None);
+while (await channel.WaitToReadAsync())
+{
+    while (channel.TryRead(out var count))
```
anurse: Comment above this line: Read all the data that's currently available synchronously, before we wait for more data to arrive.

aspnetcore/signalr/streaming/sample/Hubs/StreamHub.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.SignalR;
+
+namespace SignalRChat.Hubs
+{
+    public class StreamHub : Hub
+    {
+        public ChannelReader<int> Counter(int count, int delay)
+        {
+            var channel = Channel.CreateUnbounded<int>();
```
anurse: I think I prefer this pattern (comment included), even though I know that's not what our current sample does. It makes it clearer that you're expected to give the Writer away to some other process to do the writing.

```csharp
public ChannelReader<int> Counter(int count, int delay)
{
    var channel = Channel.CreateUnbounded<int>();

    // We don't want to await WriteItems, otherwise we'd end up waiting 
    // for all the items to be written before returning the channel back to
    // the client.
    _ = WriteItems(channel.Writer, count, delay);

    return channel.Reader;
}

private async Task WriteItems(ChannelWriter<int> writer, int count, int delay)
{
    for (var i = 0; i < count; i++)
    {
        await channel.Writer.WriteAsync(i);
        await Task.Delay(delay);
    }

    channel.Writer.TryComplete();
}
```

aspnetcore/signalr/streaming/sample/wwwroot/js/stream.js
```diff
@@ -0,0 +1,49 @@
...
+    });
+};
+const connection = new signalR.HubConnectionBuilder()
+    .withUrl("/streamHub")
+    .build();
+document.getElementById("streamButton").addEventListener("click", (event) => __awaiter(this, void 0, void 0, function* () {
+    try {
+        connection.stream("Counter", 10, 500)
+            .subscribe({
+            next: (item) => {
```
anurse: Indent these in one.

aspnetcore/signalr/streaming.md
```diff
@@ -0,0 +1,65 @@
...
+## JavaScript client
+
+JavaScript clients call streaming methods on hubs by using `connection.stream`. The `stream` method accepts two arguments:
+
+* The name of the hub method. In the following example, the hub method name is `Counter`.
+* Arguments defined in the hub method. In the following example, the arguments are: a count for the number of stream items to receive, and the delay between stream items.
+
+`connection.stream` returns an `IStreamResult` which contains a `subscribe` method. Pass an `IStreamSubscriber` to `subscribe` and set the `next`, `error`, and `complete` callbacks to get notifications from the `stream` invocation.
+
+[!code-javascript[Streaming javascript](streaming/sample/wwwroot/js/stream.js?range=16-32)]
```
anurse: You're one line short :).

aspnetcore/signalr/streaming.md
```diff
@@ -0,0 +1,68 @@
...
+ms.technology: aspnet
+ms.topic: article
+uid: signalr/streaming
+---
+
+# Use streaming in ASP.NET Core SignalR
+
+By [Brennan Conroy](https://github.com/BrennanConroy)
+
+ASP.NET Core SignalR supports streaming return values of server methods. This is useful for scenarios where fragments of data will come in over time. When a return value is streamed to the client, each fragment is sent to the client as soon as it is available, rather than waiting for all the data to become available.
```
DylanDmitri: One use case if for when "fragments come in over time" (like telemetry). The other main use case is for breaking up large downloads so they can be interleaved with other messages. If you try and download a 5gb file in one message, no other messages can be sent during that time period. 

also would change
> each fragment is sent as soon as it **becomes** available

aspnetcore/signalr/streaming.md
```diff
@@ -0,0 +1,68 @@
...
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/live/aspnetcore/signalr/streaming/sample) ([how to download](xref:tutorials/index#how-to-download-a-sample))
```
DylanDmitri: url seems broken ??

aspnetcore/signalr/streaming.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Set up the hub
+
+A hub method automatically becomes a streaming hub method when it returns a `ChannelReader<T>` or a `Task<ChannelReader<T>>`. Below is a sample that shows the basics of streaming data to the client. Whenever an object is written to the `ChannelReader` that object is immediately sent to the client. At the end, the `ChannelReader` is completed to tell the client the stream is closed.
+
```
DylanDmitri: Maybe add something saying "It's important that the calls to `WriteAsync` happen in a background thread. The HubMethod should return the `ChannelReader` quickly, so that the dispatcher is free to process other messages. Once the `ChannelReader` is returned, the background thread can send whatever stream responses it wants at its own pace."

DylanDmitri: Andrew explains this same idea in the comment on the example.

BrennanConroy:
> Most of the urls linking in code seem broken?

Some of them won't work until the doc is published to the site

PR #6932, opened 2018-06-07
Update razor-pages-start.md

bcisnero:
Changed the console command for the razor page creation with the right command.

  Fixes #Issue_Number


guardrex:
Closing because the command is correct for 2.1. They created an alias for `razor` ... the alias is `webapp`. Make sure you have the 2.1 SDK installed. If you're still developing with 2.0, then *yes*, continue to use `razor`. When the topic is flipped over to "2.0" (upper left of the page), you should see the command revert back to `razor`.

Issue #6896, opened 2018-06-06
Split dynamic loading section into separate document


davidfowl:
This document describes how to use IHostingStartup in a completely dynamic way but fails to describe the HostingStartup assembly attribute that can point to types in existing referenced libraries. https://github.com/aspnet/Hosting/blob/8377d226f1e6e1a97dabdb6769a845eeccc829ed/samples/SampleStartups/StartupInjection.cs#L8

We should talk to about the IHostingStartup feature, what it does and then describe the various ways to consume it starting from the very simple attribute to the dynamic additional deps way.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 7a759e56-4ba4-c464-ae74-e2705df72811
* Version Independent ID: 7a4ee224-ce61-ac3f-6b60-88c7a5a24f2c
* Content: [Enhance an app from an external assembly in ASP.NET Core with IHostingStartup](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/platform-specific-configuration?view=aspnetcore-2.1)
* Content Source: [aspnetcore/fundamentals/configuration/platform-specific-configuration.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/configuration/platform-specific-configuration.md)
* Product: **asp.net-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

davidfowl:
Also how did this topic end up under Configuration? It's a hosting feature and has nothing to do with configuration. We should move it as well.

guardrex:
Move it to the *Host* node? https://docs.microsoft.com/aspnet/core/fundamentals/host

guardrex:
Also need to describe the new exclusion machinery. https://github.com/aspnet/Hosting/pull/1243

PR #6969, opened 2018-06-09
Order of IHostingStartup Configure calls

### LINKED ISSUE #6951: What will be the order of middleware execution?
Middleware in configure method is executed in sequence. If I implement a middleware using IHostingStartup abstraction and use it in my project, when will this get executed? What if I want to include this only in development and after exception handler is executed? 

Here is the configure in your example:
```
        {
            if (env.IsDevelopment())
            {
                app.UseDeveloperExceptionPage();
            }
            else
            {
                app.UseHsts();
            }

            app.UseHttpsRedirection();
            app.UseStaticFiles();
            app.UseCookiePolicy();
            app.UseMvc();
        }
```

And here is the configure implementation of the StartupDiagnosticsHostingStartup 
```
public Action<IApplicationBuilder> Configure(Action<IApplicationBuilder> next)
        {
            return app =>
            {
                // Use a terminal middleware that branches on a request for
                // /services. The middleware uses a factory to obtain the services
                // registered for the app and outputs them in a webpage.
                app.Map("/services", builder => builder.Run(async ctx =>
                {
                    // Make sure this only runs in the Development environment.
```

---
#### Document Details

* ID: 7a759e56-4ba4-c464-ae74-e2705df72811
* Version Independent ID: 7a4ee224-ce61-ac3f-6b60-88c7a5a24f2c
* Content: [Enhance an app from an external assembly in ASP.NET Core with IHostingStartup](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/platform-specific-configuration?view=aspnetcore-2.1)
* Content Source: [aspnetcore/fundamentals/configuration/platform-specific-configuration.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/configuration/platform-specific-configuration.md)
* Product: **asp.net-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #6951 

This doesn't address *Split dynamic loading section into separate document* #6896 and the new exclusion bits provided by https://github.com/aspnet/Hosting/pull/1243. I'll hit those on #6896.

### DIFF
--- a/aspnetcore/fundamentals/configuration/platform-specific-configuration.md
+++ b/aspnetcore/fundamentals/configuration/platform-specific-configuration.md
@@ -51,7 +51,7 @@ A [HostingStartup](/dotnet/api/microsoft.aspnetcore.hosting.hostingstartupattrib
 
 [!code-csharp[](platform-specific-configuration/snapshot_sample/StartupEnhancement.cs?name=snippet1)]
 
-A class implements `IHostingStartup`. The class's [Configure](/dotnet/api/microsoft.aspnetcore.hosting.ihostingstartup.configure) method uses an [IWebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.iwebhostbuilder) to add enhancements to an app:
+A class implements `IHostingStartup`. The class's [Configure](/dotnet/api/microsoft.aspnetcore.hosting.ihostingstartup.configure) method uses an [IWebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.iwebhostbuilder) to add enhancements to an app. `IHostingStartup.Configure` is called before `Startup.Configure` in user code, which allows user code to overwrite any configruation provided by the hosting startup assembly.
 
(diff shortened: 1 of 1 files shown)


guardrex:
ping @davidfowl

davidfowl:
Wait, I just read this. Nobody should *ever* call Configure in an IHostingStartup. It overrides the normal one. Those do not compose.

guardrex:
@davidfowl Easy to do, but now *I'm confused!* lol

This is about the order of `Configure` executions ... not calling them.

The first blurb is saying that the `IHostingStartup.Configure` is called before `Startup.Configure` in user code (in the app where the hosting startup assembly is added).

The second blurb is saying that when multiple hosting startup assembles are present, their `Configure` methods are executed in the order that the assemblies are listed in `DOTNET_ADDITIONAL_DEPS`.

guardrex:
Would it be better as ...

> `IHostingStartup.Configure` is called by the runtime before `Startup.Configure` in user code ...

davidfowl:
Ahhh my bad. yes that's fine. I was thinking about WebHostBuilder.Configure.

PR #6977, opened 2018-06-10
Update intro.md for views->pages and removing mention of auth

pkellner:
I just tested and it does not matter whether the project in vs2017 is created with or without authentication regarding the _viewimports file. Seems odd it mentions that making the reader think that there is a difference or why would it be mentioned.  Second ,and not sure if this should be addressed because it might be the tip of the iceberg, the files are no longer in views/ but are now in pages/

  Fixes #Issue_Number


Rick-Anderson:
@pkellner  Thanks.

PR #6989, opened 2018-06-11
Fixes ResponseCaching package name and link

rodrigoff:
Changes the corresponding package for the `UseResponseCaching` middleware, which was incorrectly linking to the `Microsoft.AspNetCore.ResponseCompression` package, instead of the correct `Microsoft.AspNetCore.ResponseCaching`.


scottaddie:
@rodrigoff Nice catch! Thanks for taking the time to fix it.

PR #6996, opened 2018-06-12
Update Index.cshtml.cs

Erwin537:
Update custom fields on the user.

### DIFF
--- a/aspnetcore/security/authentication/add-user-data/sample/Areas/Identity/Pages/Account/Manage/Index.cshtml.cs
+++ b/aspnetcore/security/authentication/add-user-data/sample/Areas/Identity/Pages/Account/Manage/Index.cshtml.cs
@@ -131,7 +131,9 @@ namespace WebApp1.Areas.Identity.Pages.Account.Manage
                     throw new InvalidOperationException($"Unexpected error occurred setting phone number for user with ID '{userId}'.");
                 }
             }
-
+            
+            await _userManager.UpdateAsync(user);
+            
             await _signInManager.RefreshSignInAsync(user);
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@Erwin537  Thanks for fixing this. Name was being updated, but not DOB.

Erwin537:
@Rick-Anderson is it possible for me to also propose a change on the documentation page for people following the docs without downloading the sample? Or is that something only you can do?

Rick-Anderson:
@Erwin537  what did you have in mind?

Rick-Anderson:
You can always propose a change.

Erwin537:
@Rick-Anderson in the section addressing the changes that should be done to the register page that line of code should be added and highlighted as well I guess. (https://docs.microsoft.com/en-us/aspnet/core/security/authentication/add-user-data?view=aspnetcore-2.1&tabs=visual-studio#update-the-accountregistercshtml-page)
--
Okay, Typing this reply I found out how it works, it references the files from github and in querystrings it says which lines to highlight. Not sure how to do that though, as the referenced file is not yet updated. Should I have included those changes in this PR?

Rick-Anderson:
@Erwin537  I can fix that and you can review my PR. Send me a screen shot of what needs updating. Is it just the code you added?
I wonder how name was updated without your code.

Rick-Anderson:
this - right?
image

Erwin537:
@Rick-Anderson yes, perfect 👍 
I just tested using "Name" (I used other properties to test) but Name doesn't save for me without that one line of code either. Does your browser maybe do some autofill which made you think the property was saved?

PR #7016, opened 2018-06-13
Update sample13.cs

petyusa:
I was reading the WebAPI 2 tutorial and found a mistake in this sample. The GetBookDetail action gets a parameter called id, and in the query, this id is checked against the AuthorId, which I think is wrong, it should be checked against the BookId property.


Rick-Anderson:
@petyusa  Nice catch. It's correct in a previous snippet.

PR #7028, opened 2018-06-13
L25324: Corrupted links to snipets

v-maudel:
Hello, @scottaddie ,

Many thanks in advance.


guardrex:
This is an incorrect format for moniker ranges. However, we did settle on one line space above and below them, so I'll add a commit to update the layout of these.

PR #7057, opened 2018-06-14
301 to 307

### LINKED ISSUE #7053: Add a warning about 301 Redirects during development
The example code for AddHttpsRedirection uses a 301 redirect.

    services.AddHttpsRedirection(options =>
    {
        options.RedirectStatusCode = StatusCodes.Status301MovedPermanently;
        options.HttpsPort = 5001;
    });            

After running this once, the browser will cache that 301. During development, this means changes to this redirection code after that will not be picked up, because the browser will just continue to redirect to 5001. You will have to manually clear the cache after each change.

There should be a note or warning to use 302 status codes during testing (or make use of IHostingEnvironment).

In the next section of this document, there is already a warning against using `UseHsts` in development for comparable reasons, as it will be cached by the browser. It seems desirable to me that a comparable heads-up be given for `AddHttpsRedirection`.

Full disclosure: I was actually stumped by this for over 15 minutes before I realized it was the browser and not some leftover build or IIS config fragment. :sweat_smile:

---
#### Document Details

* Content: [Enforce HTTPS in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?view=aspnetcore-2.1&tabs=visual-studio)
* Product: **asp.net-core**
* Microsoft Alias: **riande**

Rick-Anderson:
Fixes #7053
@norachuga please review
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?view=aspnetcore-2.1&branch=pr-en-us-7057&tabs=visual-studio)

### DIFF
--- a/aspnetcore/security/enforcing-ssl.md
+++ b/aspnetcore/security/enforcing-ssl.md
@@ -42,7 +42,7 @@ The following code calls [AddHttpsRedirection](/dotnet/api/microsoft.aspnetcore.
 
 The preceding highlighted code:
 
-* Sets [HttpsRedirectionOptions.RedirectStatusCode](/dotnet/api/microsoft.aspnetcore.httpspolicy.httpsredirectionoptions.redirectstatuscode) to `Status307TemporaryRedirect`. Production apps may want to use `Status301MovedPermanently`.
+* Sets [HttpsRedirectionOptions.RedirectStatusCode](/dotnet/api/microsoft.aspnetcore.httpspolicy.httpsredirectionoptions.redirectstatuscode) to `Status307TemporaryRedirect`. Production apps should call [UseHsts](#hsts).
 * Sets the HTTPS port to 5001.
 
 The following mechanisms set the port automatically:
--- a/aspnetcore/security/enforcing-ssl.md
+++ b/aspnetcore/security/enforcing-ssl.md
@@ -42,8 +42,8 @@ The following code calls [AddHttpsRedirection](/dotnet/api/microsoft.aspnetcore.
 
 The preceding highlighted code:
 
-* Sets [HttpsRedirectionOptions.RedirectStatusCode](/dotnet/api/microsoft.aspnetcore.httpspolicy.httpsredirectionoptions.redirectstatuscode) to `Status307TemporaryRedirect`. Production apps should call [UseHsts](#hsts).
-* Sets the HTTPS port to 5001.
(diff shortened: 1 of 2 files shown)


aspnetcore/security/enforcing-ssl.md
```diff
@@ -42,7 +42,7 @@ The following code calls [AddHttpsRedirection](/dotnet/api/microsoft.aspnetcore.
 
 The preceding highlighted code:
 
-* Sets [HttpsRedirectionOptions.RedirectStatusCode](/dotnet/api/microsoft.aspnetcore.httpspolicy.httpsredirectionoptions.redirectstatuscode).
+* Sets [HttpsRedirectionOptions.RedirectStatusCode](/dotnet/api/microsoft.aspnetcore.httpspolicy.httpsredirectionoptions.redirectstatuscode) to `Status307TemporaryRedirect`. Production apps may want to use `Status301MovedPermanently`.
```
guardrex: That's the default. Which is fine. I've done this before ... shown the code with the defaults ... but I always call it out. Also, might want to use the corollary for 307, which is 308.

> \* Sets \[HttpsRedirectionOptions.RedirectStatusCode](/dotnet/api/microsoft.aspnetcore.httpspolicy.httpsredirectionoptions.redirectstatuscode) to \`Status307TemporaryRedirect\`, which is the default value. Production apps may want to use \`Status308PermanentRedirect\`.

Note that @Tratcher just said that the 307 is fine for development and production ... the HSTS ` Strict-Transport-Security` will enforce HTTPS, so there's no need to change this for a production app.

Final concern: If it says 'production app may ...', then I wonder if the reader wonders, 'ok ... how do I get the dev version to do one thing and the production version to do another?' ... they possibly need a cross-link to the multiple environments topic.

Tratcher: Ditch the 301, refer to HSTS for production.

aspnetcore/security/enforcing-ssl.md
```diff
@@ -42,7 +42,7 @@ The following code calls [AddHttpsRedirection](/dotnet/api/microsoft.aspnetcore.
 
 The preceding highlighted code:
 
-* Sets [HttpsRedirectionOptions.RedirectStatusCode](/dotnet/api/microsoft.aspnetcore.httpspolicy.httpsredirectionoptions.redirectstatuscode) to `Status307TemporaryRedirect`. Production apps may want to use `Status301MovedPermanently`.
+* Sets [HttpsRedirectionOptions.RedirectStatusCode](/dotnet/api/microsoft.aspnetcore.httpspolicy.httpsredirectionoptions.redirectstatuscode) to `Status307TemporaryRedirect`. Production apps should call [UseHsts](#hsts).
```
guardrex: I still suggest calling that 307 out as the default.

PR #7065, opened 2018-06-15
Update secure-data.md

### LINKED ISSUE #7009: Are there any examples using the Secret Manager in VS 2017?
I've tried running the final2 project and I get an error seeding the database. I believe the password is not being set properly as the error says "ArugmentNullException: Value cannot be null." I've tried following the instructions for Secrets Manager at https://docs.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-2.1&tabs=windows, but I can't seem to get it to work. Any help or direction is appreciated.

Rick-Anderson:
Fixes #7009

@rweekes4

### DIFF
--- a/aspnetcore/security/authorization/secure-data.md
+++ b/aspnetcore/security/authorization/secure-data.md
@@ -135,7 +135,7 @@ The `SeedData` class creates two accounts: administrator and manager. Use the [S
 dotnet user-secrets set SeedUserPW <PW>
 ```
 
-If you don't use a strong password, you'll get an exception when `SeedData.Initialize` is called.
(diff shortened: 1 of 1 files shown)


aspnetcore/security/authorization/secure-data.md
```diff
@@ -135,6 +135,8 @@ The `SeedData` class creates two accounts: administrator and manager. Use the [S
 dotnet user-secrets set SeedUserPW <PW>
 ```
 
+If you don't use a strong password, you'll get an exception when `SeedData.Initialize` is called.
```
guardrex: To avoid the "will" of "you'll," it could go "... password, an exception is thrown at runtime when ..."

PR #7094, opened 2018-06-16
Update configuration samples

guardrex:
Addresses #5495 

All samples confirmed running :+1: with these updates.

### DIFF
--- a/aspnetcore/fundamentals/configuration/index/sample/CustomConfigurationProvider/CustomConfigurationProvider.csproj
+++ b/aspnetcore/fundamentals/configuration/index/sample/CustomConfigurationProvider/CustomConfigurationProvider.csproj
@@ -1,16 +1,16 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
     <OutputType>Exe</OutputType>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="2.0.0" />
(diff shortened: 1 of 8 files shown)


aspnetcore/fundamentals/configuration/index/sample/CustomConfigurationProvider/CustomConfigurationProvider.csproj
```diff
@@ -1,16 +1,16 @@
...
     <OutputType>Exe</OutputType>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="2.0.0" />
-    <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="2.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="2.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="2.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="2.0.0" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="2.1.0" />
```
Rick-Anderson: If you use M.A.App then you don't have a version that needs updating next time. Just a thought, I'm not requiring that.

guardrex: I didn't want to mess with them much because we're probably going to need to perform a config shake-up at some point. The whole topic is based on `ConfigurationBuilder`s with console apps, which means that the whole topic doesn't match what devs see in web project templates. This is a fine topic for Configuration API and `ConfigurationBuilder`, but I think we need a sister topic that pertains explicitly to configuration of a RP app. Somehow, we need to figure out how to do it with related info in ...
* App startup
* Multiple environments
* Web Host configuration (and later Generic Host)

😅

PR #7173, opened 2018-06-20
Remove Unnecessary Using

raiytu4:
Remove Unnecessary Using


Rick-Anderson:
@raiytu4  thanks for your contribution. We've got lots of files with unnecessary using and we typically don't remove them outside of a doc sweep. Thanks again.

PR #7223, opened 2018-06-21
Move INCLUDES file

guardrex:


Rick-Anderson:
@guardrex  OK to S & M?
image

guardrex:
😅 What a pain. Yes, it's finally in order. Those dot dot *DOT* links ... dot dot dot ... slash ... are finally ok.

This should be fine when the repos are split. I moved the include file and the links to it are good now.

PR #7242, opened 2018-06-22
In ASP.Net core 2.0, the route for this controller should be api/[con…

thegrims:
…troller] rather than api/[todo]


scottaddie:
@thegrims The text is simply explaining what ASP.NET Core is doing with that `[Route]` attribute. It's not an instruction to the reader to replace anything.

thegrims:
@scottaddie My bad, thanks for the clarification.

PR #7247, opened 2018-06-22
Update AccessDenied.cshtml

jackfperryjr:
It was just missing the `@` in the `<h1>` tag for the title.


scottaddie:
@jackfperryjr Thank you for fixing this! As an FYI, this was introduced by a project template issue that has since been addressed. See https://github.com/aspnet/templating/pull/40.

jackfperryjr:
@scottaddie No sweat. I was just browsing and it caught my eye!

Issue #7248, opened 2018-06-22
Explain scaffold generated files.


BrittanyDilts:
After the section on the [Update-Database] step, a sentence is included and makes no sense because haven't defined a "DbContext" and there is no created "Models/MovieContext.cs" file. IIRC, this was done in the 2.0 tut, but if you're following step-by-step along, this is absolutely confounding because we haven't done anything involving "DbContext" or a "MovieContext.cs" file in the 2.1 tutorial.

Quoted text:
The schema is based on the model specified in the [DbContext] (In the Models/MovieContext.cs file). 

Edit - List the generated files. The CRUD files are explained in the next part. Explain `RazorPagesMovieContext`

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 6719f08e-3bd7-dc1a-71df-f2ef9fbca9d8
* Version Independent ID: 7096fdb3-612e-9e00-bd0b-8ea4886a09ce
* Content: [Add a model to a Razor Pages app in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/tutorials/razor-pages/model?view=aspnetcore-2.1)
* Content Source: [aspnetcore/tutorials/razor-pages/model.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/tutorials/razor-pages/model.md)
* Product: **aspnet-core**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
In 2.1 the scaffolder create the DbContext.

BrittanyDilts:
Yes, I noticed that - we set the DbContext to "RazorPagesMovieContext". I guess my point was there was no previous reference to what a DbContext is, that we've already created one, and that there's a referenced file path that doesn't exist ("Models/MovieContext.cs"). Unless this tutorial isn't meant to be used by n00bs, it was just a note of confusing text. I mean, I've been  crash-coursed into RazorPages, have been up to my ears in learning about it and it still threw me for a loop.

Rick-Anderson:
Would it be more clear if I wrote
The schema is based on the model specified in the `RazorPagesMovieContext` (In the Models/MovieContext.cs file).

Rick-Anderson:
What it needs is before **Perform initial migration** it needs a list of files generated. It should explain the DbContext. That would make it easier to follow
You're right, that's the wrong place.

[...]

PR #7249, opened 2018-06-22
Update model.md

Rick-Anderson:
@BrittanyDilts  please review. Short term fix until I can address #7248


BrittanyDilts:
That's great. Reading through it is not confusing at all if you are paying attention to what you're doing in the tutorial (and not paying attention while going through a tutorial is an ID-10t error and you can't really fix those).

PR #7269, opened 2018-06-25
Update adding-model.md

DanielSchmitt94:
On Website the "Add a model to an ASP.NET Core MVC app"-text is shown twice.


guardrex:
Thanks @DanielSchmitt94.

I see a few more to yank in the other xplat tutorials, so I'll add them to this PR.

PR #7272, opened 2018-06-25
C26453: escaping URL to display it as text

v-maudel:
Hello, @MikeWasson  ,
"Example URL at line 24 is displayed as a link, it must be escaped in order to be displayed as text"

Please review and merge the proposed file change to fix to target versions. If you make related fix in another PR then share your PR number so we can confirm and close this PR.
Many thanks in advance.


v-maudel:
Thanks @guardrex ,

PR #7273, opened 2018-06-25
C26450: URL format modified

v-maudel:
Hello, @rick-anderson ,
"URL is not well formatted if this is intended to be shown as a link"

Please review and merge the proposed file change to fix to target versions. If you make related fix in another PR then share your PR number so we can confirm and close this PR.
Many thanks in advance.


v-maudel:
Thanks @guardrex

PR #7276, opened 2018-06-25
Changes "logs" to be lower-case everywhere

arglgruml:
The directory structure table and text mentions a lower case *logs* directory. 
After the XML code snippet the spelling changes to *Logs* (upper case).
Changed every mention of the *logs* directory to be lower case.


arglgruml:
@guardrex  That was a coinflip - evidently I chose the wrong side. :)

guardrex:
It's a tough one ... the default *web.config* will always have it lowercase, so complete consistency is something that we'll never reach on this one.

Thanks again for calling it out.

PR #7287, opened 2018-06-25
Simple grammar fix

zunderscore:
Removed an extra verb so that sentence now reads "The `Genre` field is no longer nullable" instead of "The `Genre` field is are no longer nullable"


Rick-Anderson:
@zunderscore  Thanks!

PR #7312, opened 2018-06-26
SignalR AuthN/AuthZ and Security Concepts

anurse:
Docs on Authentication, Authorization and some notes about security considerations. A lot of the things from [the original ASP.NET Signal doc](https://docs.microsoft.com/en-us/aspnet/signalr/overview/security/introduction-to-security) aren't really relevant any more.

Review Links:
* [Auth/AuthZ](https://review.docs.microsoft.com/en-us/aspnet/core/signalr/authn-and-authz?view=aspnetcore-2.1&branch=anurse%2Fsignalr-security)
* [Security](https://review.docs.microsoft.com/en-us/aspnet/core/signalr/security?view=aspnetcore-2.1&branch=anurse%2Fsignalr-security)

TODO:
* [x] Build a full sample for auth and add the snippets in.
* [x] Add docs on buffer management, back pressure and maximum message sizes in security doc.


anurse:
:up: :date: added the sample and some docs on Windows Auth. Please review! 😄

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
+---
+title: Authentication and Authorization in SignalR
+author: rachelappel
+description: Learn how to use authentication and authorization in ASP.NET Core SignalR.
+monikerRange: '>= aspnetcore-2.1'
+ms.author: rachelap
+ms.custom: mvc
+ms.date: 05/01/2018
```
BrennanConroy: fix date

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+uid: signalr/authn-and-authz
+---
+
+# Authentication and authorization in SignalR
+
+By [Andrew Stanton-Nurse](https://twitter.com/anurse)
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/signalr/authn-and-authz/sample/ ) [(how to download)](xref:tutorials/index#how-to-download-a-sample)
```
BrennanConroy: nit: space at end of url

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Authenticate users connecting to a SignalR Hub
+
+SignalR can be used with [ASP.NET Core Authentication](xref:security/authentication/index) to associate a User with each connection. In a Hub, authentication data can be accessed from the [`HubConnectionContext.User`](/dotnet/api/microsoft.aspnetcore.signalr.hubconnectioncontext.user?view=aspnetcore-2.1) property. Once a connection is authenticated, it is possible to invoke methods on all connections associated with a specific user. See [Manage users and groups in SignalR](xref:signalr/groups) for more information.
```
BrennanConroy: "Once a connection is authenticated, it is possible to invoke methods on all connections associated with a specific user."
What exactly does this sentence mean?

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,36 @@
...
+* HTTP methods `GET` and `POST` must be allowed.
+* Credentials must be enabled, even when you are not using authentication.
+
+### Access Token Logging
+
+Due to limitations in browser APIs, when using WebSockets or EventSource, the access token must be specified in the query string. This is generally as secure as using the standard `Authorization` header, however many web servers log the URL for each request, including the query string. This means the access token may be included in logs. Consider reviewing your web server's logging settings to avoid logging this information.
+
+### Exceptions
+
+Exception messages are generally considered sensitive data that should not be revealed to a client. By default SignalR will not send the details of an exception thrown by a Hub method to the client, the client will receive a generic message indicating an error occurred. You can override this behavior by setting the `EnableDetailedErrors` setting.
```
BrennanConroy: link to configuration doc for `EnableDetailedErrors`?

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Security Considerations in ASP.NET Core SignalR
+author: rachelappel
+description: Learn how to use authentication and authorization in ASP.NET Core SignalR.
+monikerRange: '>= aspnetcore-2.1'
+ms.author: rachelap
+ms.custom: mvc
+ms.date: 05/01/2018
```
BrennanConroy: date

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Cross-Origin Resource Sharing
+
+[Cross-origin resource sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) can be used to allow cross-origin SignalR connections in the browser. If your JavaScript code is hosted on a different domain name from your SignalR application, you will have to enable the [ASP.NET Core CORS Middleware](xref:security/cors) in order to allow the connection. You should generally only allow cross-origin requests from domains you control. For example, if your site is hosted on `http://www.example.com` and your SignalR application is hosted on `http://signalr.example.com`, you should configure CORS in your SignalR application to only allow the origin `www.example.com`.
+
+For more information on configuring CORS, see [the documentation on ASP.NET Core CORS](xref:security/cors). SignalR requires the following CORS policies in order to operate correctly:
+
+* The policy must allow the specific origins you expect or allow any origin (not recommended).
+* HTTP methods `GET` and `POST` must be allowed.
+* Credentials must be enabled, even when you are not using authentication.
```
BrennanConroy: code snippet showing example of AllowCredentials, Get and Post enabled and allow example.com origin?

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Cookie Authentication
+
+In a simple web application, cookie authentication allows your existing user credentials to automatically flow to SignalR connections. When using the browser client, no additional configuration is needed. If the user is logged in to your application, the SignalR connection will automatically inherit this authentication.
```
rachelappel: simple -> is that necessary? and not true for more complex apps?
application -> app
will automatically inherit -> automatically inherits

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+
+We do not recommend using Cookie authentication unless you only need to authenticate users from the browser client. When using the .NET Client, the `Cookies` property can be configured in the `.WithUrl` call in order to provide a cookie. However, this requires that you provide an API to exchange authentication data for a Cookie.
```
rachelappel: Cookie -> cookie

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Bearer Token Authentication
+
+When using the .NET Client, or when your SignalR Hubs are located on a different server from your web application, we recommend using bearer token authentication. In this authentication scheme, you configure the SignalR client with an access token to send to the server. The server validates this token and uses the data within it to identify the user. The details of bearer token authentication are beyond the scope of this document, but there are some things to note when using this form of authentication in SignalR. On the server, bearer token authentication is configured using the [JWT Bearer middleware](/dotnet/api/microsoft.extensions.dependencyinjection.jwtbearerextensions.addjwtbearer?view=aspnetcore-2.1).
```
rachelappel: Hubs -> hubs
we recommend using bearer token authentication -> bearer token authentication is the recommended approach. (or scheme)

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+
+In the JavaScript client, the token can be provided using the `accessTokenFactory` option.
+
+[!code-javascript[Configure Access Token](authn-and-authz/sample/wwwroot/js/chat.ts?range=63-65)]
+
+In the .NET client, there is a simlar `AccessTokenProvider` property that can be used to configure this:
```
rachelappel: this: -> the token:

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+
+```csharp
+var connection = new HubConnectionBuilder()
+    .WithUrl("https://example.com/myhub", options => options.AccessTokenProvider = () => _myAccessToken)
+    .Build();
+```
+
+> [!NOTE]
+> This function is called before **every** HTTP request made by SignalR. This means that if you need to renew the token in order to keep the connection active (because it may expire during the connection) you can do so from within this function and SignalR will automatically start using the new token.
```
rachelappel: This means that if you need to renew the token in order to keep the connection active (because it may expire during the connection) you can do so from within this function and SignalR will automatically start using the new token.
->
If you need to renew the token in order to keep the connection active (because it may expire during the connection), do so from within this function. SignalR will automatically start using the new token.

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+
+In standard Web APIs, bearer tokens are sent via an HTTP header. However, due to security limitations imposed by browsers, the browser client is not able to set these headers when using the WebSockets or Server-Sent Events transports. When using those transports, the token is transmitted as a query string parameter. In order to support this on the server, additional configuration is required:
```
rachelappel: via -> through or in 
is not -> isn't

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+
+[!code-csharp[Configure Server to accept access token from Query String](authn-and-authz/sample/Startup.cs?range=33-34,42-80,90)]
+
+### Windows Authentication
+
+If you have configured [Windows Authentication](xref:security/authentication/windowsauth) in your application, SignalR can use that authentication to secure Hubs. However, in order to send messages to individual users, you need to add a custom User ID provider, because the Windows Authentication system does not provide the "Name Identifier" claim that SignalR uses to determine the user name. To do this, add a new class that implements `IUserIdProvider` and retrieve one of the claims from the user to use as the identifier. For example, to use the "Name" claim (which is the Windows username in the form `[Domain]\[Username]`), create the following class:
```
rachelappel: application->app
Hubs->hubs
However, in order to send messages to individual users, you need to add a custom User ID provider, because the Windows Authentication system does not provide the "Name Identifier" claim that SignalR uses to determine the user name.
->
However, In order to send messages to individual users, you need to add a custom user ID provider. This is because the Windows authentication system doesn't provide the "Name Identifier" claim that SignalR uses to determine the user name.

aspnetcore/signalr/authn-and-authz/sample/Hubs/ChatHub.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.AspNetCore.SignalR;
+
+namespace SignalRAuthenticationSample.Hubs
+{
+    [Authorize]
+    public class ChatHub: Hub
+    {
+        public override async Task OnConnectedAsync()
+        {
+            await Clients.All.SendAsync("ReceiveSystemMessage", $"{Context.User.Identity.Name} joined.");
```
BrennanConroy: `Context.UserIdentifier` otherwise you aren't using the custom UserIdProvider

aspnetcore/signalr/authn-and-authz.md
rachelappel: In a Hub -> In a hub

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+    // ... other services ...
+
+    services.AddSignalR();
+    services.AddSingleton<IUserIdProvider, NameUserIdProvider>();
+}
+```
+
+## Authorize users to access Hubs and Hub methods
+
+By default, all methods in a Hub can be called by an unauthenticated user. In order to require authentication, apply the [`Authorize`](/dotnet/api/microsoft.aspnetcore.authorization.authorizeattribute?view=aspnetcore-2.1) attribute to the Hub:
```
rachelappel: in a Hub -> in a hub
to the Hub: -> to the hub:

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,36 @@
...
+author: rachelappel
+description: Learn how to use authentication and authorization in ASP.NET Core SignalR.
+monikerRange: '>= aspnetcore-2.1'
+ms.author: rachelap
+ms.custom: mvc
+ms.date: 05/01/2018
+uid: signalr/security
+---
+
+# Security Concepts in ASP.NET Core SignalR
```
BrennanConroy: Should we talk about the buffer settings and the memory concerns if it's changed?

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,36 @@
...
+
+By [Andrew Stanton-Nurse](https://twitter.com/anurse)
+
+## Overview
+
+SignalR provides a number of security protections by default, but it is important to understand their limitations.
+
+### Cross-Origin Resource Sharing
+
+[Cross-origin resource sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) can be used to allow cross-origin SignalR connections in the browser. If your JavaScript code is hosted on a different domain name from your SignalR application, you will have to enable the [ASP.NET Core CORS Middleware](xref:security/cors) in order to allow the connection. You should generally only allow cross-origin requests from domains you control. For example, if your site is hosted on `http://www.example.com` and your SignalR application is hosted on `http://signalr.example.com`, you should configure CORS in your SignalR application to only allow the origin `www.example.com`.
```
rachelappel: [ASP.NET Core CORS Middleware] -> [ASP.NET Core CORS middleware]

rachelappel: You should generally only allow cross-origin requests from domains you control. ->
In general, allow cross-origin requests only from domains you control.

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,36 @@
...
+
+For more information on configuring CORS, see [the documentation on ASP.NET Core CORS](xref:security/cors). SignalR requires the following CORS policies in order to operate correctly:
+
+* The policy must allow the specific origins you expect or allow any origin (not recommended).
```
rachelappel: expect or -> expect, or

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,36 @@
...
+* HTTP methods `GET` and `POST` must be allowed.
+* Credentials must be enabled, even when you are not using authentication.
+
+### Access Token Logging
+
+Due to limitations in browser APIs, when using WebSockets or EventSource, the access token must be specified in the query string. This is generally as secure as using the standard `Authorization` header, however many web servers log the URL for each request, including the query string. This means the access token may be included in logs. Consider reviewing your web server's logging settings to avoid logging this information.
```
rachelappel: Should there be a link to EventSource? This link? https://developer.mozilla.org/en-US/docs/Web/API/EventSource

rachelappel: your web server's -> the web server's

aspnetcore/signalr/security.md
rachelappel: By default SignalR will not send the details of an exception thrown by a Hub method to the client, the client will receive a generic message indicating an error occurred.
->
By default, SignalR doesn't send the details of an exception thrown by a hub method to the client. Instead, the client receives a generic message indicating an error occurred.

aspnetcore/signalr/authn-and-authz.md
anurse: I'm trying to say that once you apply the `Authorize` attribute you can use `Clients.User(...)`. But yeah, it needs some clearing up.

anurse: but it sounds like those two aren't as coupled as I thought

aspnetcore/signalr/authn-and-authz.md
anurse: I'll clarify what is meant by simple. We generally do not recommend using cookies in anything that may have multiple clients (i.e. not just the browser)

aspnetcore/signalr/authn-and-authz.md
mikaelm12: Might be worth adding in a note that multiple connections can be associated with a single user

aspnetcore/signalr/authn-and-authz.md
mikaelm12: Maybe point out that you can't require auth on all hubs like you used to be able to in ASP.NET SignalR

aspnetcore/signalr/security.md
anurse: Yep, hadn't been published when I wrote this part :). Good catch.

aspnetcore/signalr/security.md
anurse: YAS

aspnetcore/signalr/authn-and-authz.md
anurse: Ah man, I had it in my head that "via" was a word to avoid and I still threw one in here 😄 . My mind must just work via "via" all the time 😛

aspnetcore/signalr/security.md
anurse: I don't think it's necessary, but I will change it to be "Server-Sent Events" which is the term we've used elsewhere in the docs.

aspnetcore/signalr/authn-and-authz.md
anurse: I don't think we need to call out missing things like this here. Plus, we are considering bringing it back anyway :)

anurse:
:up: :date: I think I got all the feedback except @BrennanConroy 's comment on adding a section about buffer management, etc. I'll work on that and ping this thread when there's an update.

aspnetcore/signalr/authn-and-authz/sample/Startup.cs
```diff
@@ -0,0 +1,118 @@
...
+                    // to read the access token from the query string when a WebSocket or Server-Sent Events request comes in.
+                    options.Events = new JwtBearerEvents
+                    {
+                        OnMessageReceived = context =>
+                        {
+                            var accessToken = context.Request.Query["access_token"];
+
+                            // If the request is for our hub...
+                            if (!string.IsNullOrEmpty(accessToken) &&
+                                (context.HttpContext.Request.Path.StartsWithSegments("/hubs/chat")))
```
davidfowl: Is this what we expect people to do? Check the path?

aspnetcore/signalr/authn-and-authz.md
rachelappel: That one took me a while too.

aspnetcore/signalr/authn-and-authz/sample/Startup.cs
anurse: Our examples use `IsWebSocketRequest` but there's a problem with that. If you have the following:

```csharp
app.UseAuthentication();
app.UseSignalR();
```

Then `IsWebSocketRequest` won't be set when your event is fired because it's the WebSocket middleware that does that and SignalR adds the WebSocket middleware, so it ends up **after** authentication. This can be fixed by adding `UseWebSockets` above `UseAuthentication` but that seemed harder to explain and uglier.

anurse:
:up: :date: added the buffer management section

aspnetcore/signalr/authn-and-authz/sample/Hubs/ChatHub.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public override async Task OnDisconnectedAsync(Exception exception)
+        {
+            await Clients.All.SendAsync("ReceiveSystemMessage", $"{Context.UserIdentifier} left.");
+            await base.OnDisconnectedAsync(exception);
+        }
+
+        public async Task SendToUser(string user, string message)
+        {
+            await Clients.User(user).SendAsync("ReceiveDirectMessage", $"{Context.UserIdentifier}: {message}");
```
BrennanConroy: super nit: return

aspnetcore/signalr/authn-and-authz/sample/wwwroot/js/chat.ts
```diff
@@ -0,0 +1,147 @@
...
+        } finally {
+            // Update rendering with any final state.
+            this.render();
+        }
+    }
+
+    public async directMessageFormSubmitted(evt: Event) {
+        try {
+            evt.preventDefault();
+            await this.connection.invoke("SendToUser", toUserInput.value, directMessageInput.value);
```
BrennanConroy: nit: send

aspnetcore/signalr/authn-and-authz/sample/Startup.cs
```diff
@@ -0,0 +1,118 @@
...
+            services.AddDbContext<ApplicationDbContext>(options =>
+                options.UseSqlServer(Configuration.GetConnectionString("DefaultConnection")));
+
+            services.AddIdentity<ApplicationUser, IdentityRole>()
+                .AddEntityFrameworkStores<ApplicationDbContext>()
+                .AddDefaultTokenProviders();
+
+            services.AddAuthentication(options =>
+                {
+                    // Identity made Cookie authentication the default, we want JWT Bearer Auth to be the default.
```
Rick-Anderson: Code looks close, spit up comments
image

scottaddie: The official guidance is to not exceed 85 characters per line.

rachelappel: "spit up comments" :joy: :joy: :joy: :joy:

scottaddie: @rachelappel The hack-a-loogie variety. 😄

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+
+# Authentication and authorization in SignalR
+
+By [Andrew Stanton-Nurse](https://twitter.com/anurse)
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/signalr/authn-and-authz/sample/) [(how to download)](xref:tutorials/index#how-to-download-a-sample)
+
+## Authenticate users connecting to a SignalR hub
+
+SignalR can be used with [ASP.NET Core Authentication](xref:security/authentication/index) to associate a User with each connection. In a hub, authentication data can be accessed from the [`HubConnectionContext.User`](/dotnet/api/microsoft.aspnetcore.signalr.hubconnectioncontext.user?view=aspnetcore-2.1) property. Authentication also allows the hub to call methods on all connections associated with a user (See [Manage users and groups in SignalR](xref:signalr/groups) for more information). Multiple connections may be associated with a single user.
```
scottaddie: Remove the query string from the API ref link.

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Cookie Authentication
+
+In a browser-based app, cookie authentication allows your existing user credentials to automatically flow to SignalR connections. When using the browser client, no additional configuration is needed. If the user is logged in to your app, the SignalR connection automatically inherits this authentication.
+
+We do not recommend using Cookie authentication unless you only need to authenticate users from the browser client. When using the .NET Client, the `Cookies` property can be configured in the `.WithUrl` call in order to provide a cookie. However, this requires that you provide an API to exchange authentication data for a cookie.
```
scottaddie: - do not --> don't
- link the ".NET Client" text to the associated doc (xref:signalr/dotnet-client)

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Bearer Token Authentication
+
+When using the .NET Client, or when your SignalR hubs are located on a different server from your web app, bearer token authentication is the recommended approach. In this authentication scheme, you configure the SignalR client with an access token to send to the server. The server validates this token and uses the data within it to identify the user. The details of bearer token authentication are beyond the scope of this document, but there are some things to note when using this form of authentication in SignalR. On the server, bearer token authentication is configured using the [JWT Bearer middleware](/dotnet/api/microsoft.extensions.dependencyinjection.jwtbearerextensions.addjwtbearer?view=aspnetcore-2.1).
```
scottaddie: Remove the query string from API ref link.

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+
+In the JavaScript client, the token can be provided using the `accessTokenFactory` option.
+
+[!code-javascript[Configure Access Token](authn-and-authz/sample/wwwroot/js/chat.ts?range=63-65)]
```
scottaddie: code-javascript --> code-typescript

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+```csharp
+var connection = new HubConnectionBuilder()
+    .WithUrl("https://example.com/myhub", options => options.AccessTokenProvider = () => _myAccessToken)
+    .Build();
+```
+
+> [!NOTE]
+> This function is called before **every** HTTP request made by SignalR. If you need to renew the token in order to keep the connection active (because it may expire during the connection), do so from within this function and return the updated token.
+
+In standard Web APIs, bearer tokens are sent in an HTTP header. However, due to security limitations imposed by browsers, the browser client isn't able to set these headers when using the WebSockets or Server-Sent Events transports. When using those transports, the token is transmitted as a query string parameter. In order to support this on the server, additional configuration is required:
```
scottaddie: isn't able --> is unable

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+    }
+}
+```
+
+Instead of `ClaimTypes.Name`, you can use any value from the `User` (such as the Windows SID identifier, etc.).
+
+> [!NOTE]
+> The value you choose must be unique among all the users in your system. Otherwise, a message intended for one user could end up going to a different user.
+
+Then, register this component in your `ConfigureServices` method **after** the call to `.AddSignalR`
```
scottaddie: `ConfigureServices` --> `Startup.ConfigureServices`

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+    // ... other services ...
+
+    services.AddSignalR();
+    services.AddSingleton<IUserIdProvider, NameUserIdProvider>();
+}
+```
+
+## Authorize users to access hubs and hub methods
+
+By default, all methods in a hub can be called by an unauthenticated user. In order to require authentication, apply the [`Authorize`](/dotnet/api/microsoft.aspnetcore.authorization.authorizeattribute?view=aspnetcore-2.1) attribute to the hub:
```
scottaddie: Remove the query string from the API ref link.

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+
+[!code-csharp[Restrict a hub to only authorized users](authn-and-authz/sample/Hubs/ChatHub.cs?range=8-10,32)]
+
+You can use the constructor arguments and properties of the `Authorize` attribute to restrict access to only users matching specific [authorization policies](xref:security/authorization/policies). For example, if you have a custom authorization policy called `MyAuthorizationPolicy` you can ensure that only users matching that policy can access the hub using the following code:
```
scottaddie: `Authorize` --> `[Authorize]`

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+
+```csharp
+[Authorize("MyAuthorizationPolicy")]
+public class ChatHub: Hub
+{
+}
+```
+
+Individual hub methods can have the `Authorize` attribute applied as well. If the current user does not match the policy applied to the method, an error will be returned to the caller:
```
scottaddie: - `Authorize` --> `[Authorize]`
- does not --> doesn't
- will be --> is

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,114 @@
...
+
+In the JavaScript client, the token can be provided using the `accessTokenFactory` option.
+
+[!code-javascript[Configure Access Token](authn-and-authz/sample/wwwroot/js/chat.ts?range=63-65)]
+
+In the .NET client, there is a simlar `AccessTokenProvider` property that can be used to configure the token:
+
+```csharp
+var connection = new HubConnectionBuilder()
+    .WithUrl("https://example.com/myhub", options => options.AccessTokenProvider = () => _myAccessToken)
```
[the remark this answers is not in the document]
anurse: > With a snippet you can spit the code up and compile to make sure you have compilable code.

As I asked in a different issue, I'm not clear on the policy around this. These code snippets are not part of a sample and therefore it isn't possible to compile them. I could build a bunch more samples, but that would dramatically delay this doc. I don't think there's value in just putting snippets into a C# file with no associated sample as that doesn't actually protect against syntax/semantic errors.

I can easily just throw these in to `.cs` files and link them, but I don't see the value without a full sample that can be compiled.

I'll go through and split up the comments for sure. Is there a documented column-count that makes things appropriate for mobile devices?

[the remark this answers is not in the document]
anurse: I see from @scottaddie in a different comment that 85 characters is the limit. I'll update the snippets and sample to break lines over that limit.

aspnetcore/signalr/authn-and-authz/sample/Hubs/ChatHub.cs
anurse: I actually don't agree for samples. For our code, performance is crucial, but for samples I think it's easier if the user can simply add their own code around our sample without worrying about having to switch from `return`ing a task to `await`ing` it.

BrennanConroy: 😢 but lots of users don't know there is a difference and we're teaching them bad practices...
Although, probably better to accidentally allocate a state machine than to not await a `SendAsync` which is more common.

aspnetcore/signalr/authn-and-authz.md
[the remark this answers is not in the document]
Rick-Anderson: >  85 characters is the limit.

That's a guideline we try to follow. Template generated code can be excluded and  markup that isn't easily split up JSON/XML

aspnetcore/signalr/authn-and-authz/sample/Hubs/ChatHub.cs
Rick-Anderson: My preference is to use best practices, unless it's too complicated. If we go with return task we need a side bar of  `return a task vs. await`

This sample returns a task. A production app might gain perf benefits from async.

The [intro into Razor Pages](https://docs.microsoft.com/en-us/aspnet/core/tutorials/razor-pages/page?view=aspnetcore-2.1) is all async

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,119 @@
+---
+title: Authentication and Authorization in SignalR
+author: rachelappel
+description: Learn how to use authentication and authorization in ASP.NET Core SignalR.
+monikerRange: '>= aspnetcore-2.1'
+ms.author: rachelap
```
Rick-Anderson: anurse

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,119 @@
...
+
+# Authentication and authorization in SignalR
+
+By [Andrew Stanton-Nurse](https://twitter.com/anurse)
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/signalr/authn-and-authz/sample/) [(how to download)](xref:tutorials/index#how-to-download-a-sample)
+
+## Authenticate users connecting to a SignalR hub
+
+SignalR can be used with [ASP.NET Core Authentication](xref:security/authentication/index) to associate a User with each connection. In a hub, authentication data can be accessed from the [`HubConnectionContext.User`](/dotnet/api/microsoft.aspnetcore.signalr.hubconnectioncontext.user) property. Authentication also allows the hub to call methods on all connections associated with a user (See [Manage users and groups in SignalR](xref:signalr/groups) for more information). Multiple connections may be associated with a single user.
```
Rick-Anderson: Authentication ~also~ allows
avoid also

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,119 @@
...
+
+Cookie authentication is not recommended unless the app only needs to authenticate users from the browser client. When using the [.NET Client](xref:signalr/dotnet-client), the `Cookies` property can be configured in the `.WithUrl` call in order to provide a cookie. However, using cookie authentication from the .NET Client requires the app to provide an API to exchange authentication data for a cookie.
+
+### Bearer token authentication
+
+Bearer token authentication is the recommended approach when using clients other than the browser client. In this approach, the client provides an access token that the server validates and uses to identify the user. The details of bearer token authentication are beyond the scope of this document, but there are some things to note when using this form of authentication in SignalR. On the server, bearer token authentication is configured using the [JWT Bearer middleware](/dotnet/api/microsoft.extensions.dependencyinjection.jwtbearerextensions.addjwtbearer).
```
Rick-Anderson: avoid this
In this approach, 
In the bearer token approach,

Find a way to split this sentence

>The details of bearer token authentication are beyond the scope of this document, but there are some things to note when using this form of authentication in SignalR.

The details of bearer token authentication are beyond the scope of this document.  There are some things to note when using bearer token authentication  with SignalR.

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,119 @@
...
+    { 
+        options.AccessTokenProvider = () => _myAccessToken;
+    })
+    .Build();
+```
+
+> [!NOTE]
+> The access token function you provide is called before **every** HTTP request made by SignalR. If you need to renew the token in order to keep the connection active (because it may expire during the connection), do so from within this function and return the updated token.
+
+In standard Web APIs, bearer tokens are sent in an HTTP header. However, due to security limitations imposed by browsers, the browser client is unable to set these headers when using the WebSockets or Server-Sent Events transports. When using those transports, the token is transmitted as a query string parameter. In order to support this on the server, additional configuration is required:
```
Rick-Anderson: Can we shorten some of these ROs?

 However the browser client is unable to set bearer token headers when using the WebSockets or Server-Sent Events transports. The browser can't set set bearer token headers with WebSockets or Server-Sent Events transports for security reasons.

I don't like my rewrite but it's easier to read and MT.

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,119 @@
+---
+title: Authentication and Authorization in SignalR
```
scottaddie: Authorization --> authorization

aspnetcore/signalr/authn-and-authz.md
scottaddie: User --> user

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Cookie authentication
+
+In a browser-based app, cookie authentication allows your existing user credentials to automatically flow to SignalR connections. When using the browser client, no additional configuration is needed. If the user is logged in to your app, the SignalR connection automatically inherits this authentication.
+
+Cookie authentication is not recommended unless the app only needs to authenticate users from the browser client. When using the [.NET Client](xref:signalr/dotnet-client), the `Cookies` property can be configured in the `.WithUrl` call in order to provide a cookie. However, using cookie authentication from the .NET Client requires the app to provide an API to exchange authentication data for a cookie.
```
scottaddie: is not --> isn't

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Bearer token authentication
+
+Bearer token authentication is the recommended approach when using clients other than the browser client. In this approach, the client provides an access token that the server validates and uses to identify the user. The details of bearer token authentication are beyond the scope of this document, but there are some things to note when using this form of authentication in SignalR. On the server, bearer token authentication is configured using the [JWT Bearer middleware](/dotnet/api/microsoft.extensions.dependencyinjection.jwtbearerextensions.addjwtbearer).
+
+In the JavaScript client, the token can be provided using the [`accessTokenFactory`](xref:signalr/configuration#configure-bearer-authentication) option.
```
scottaddie: Remove the code-fencing from accessTokenFactory, since it's a link.

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,119 @@
...
+
+[!code-typescript[Configure Access Token](authn-and-authz/sample/wwwroot/js/chat.ts?range=63-65)]
+
+In the .NET client, there is a similar [`AccessTokenProvider`](xref:signalr/configuration#configure-bearer-authentication) property that can be used to configure the token:
```
scottaddie: Remove code-fencing from AccessTokenProvider, since it's a link.

aspnetcore/signalr/authn-and-authz.md
scottaddie: Web APIs --> web APIs

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,119 @@
...
+
+In standard Web APIs, bearer tokens are sent in an HTTP header. However, due to security limitations imposed by browsers, the browser client is unable to set these headers when using the WebSockets or Server-Sent Events transports. When using those transports, the token is transmitted as a query string parameter. In order to support this on the server, additional configuration is required:
+
+[!code-csharp[Configure Server to accept access token from Query String](authn-and-authz/sample/Startup.cs?range=33-34,42-80,90)]
+
+### Windows authentication
+
+If [Windows authentication](xref:security/authentication/windowsauth) is configured in your app, SignalR can use that identity to secure hubs. However, in order to send messages to individual users, you need to add a custom User ID provider. This is because the Windows authentication system does not provide the "Name Identifier" claim that SignalR uses to determine the user name.
```
scottaddie: does not --> doesn't

aspnetcore/signalr/authn-and-authz/sample/wwwroot/js/chat.js
```diff
@@ -0,0 +1,142 @@
+var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
```
scottaddie: Add the generated *.js files to a *.gitignore file.

aspnetcore/signalr/index.md
```diff
@@ -7,6 +7,7 @@ ms.author: rachelap
 ms.date: 06/21/2018
 uid: signalr/index
 ---
+
```
scottaddie: This file looks outdated. Is a rebase needed?

aspnetcore/toc.md
```diff
@@ -250,6 +250,8 @@
 ## [HubContext](xref:signalr/hubcontext)
 ## [Users and groups](xref:signalr/groups)
 ## [Configuration](xref:signalr/configuration)
+## [Authentication and Authorization](xref:signalr/authn-and-authz)
```
scottaddie: This file looks outdated too. Did we decide to create a new "Security" ToC node within "SignalR"?

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,70 @@
...
+
+There are separate limits for incoming and outgoing messages, both can be configured on the [`HttpConnectionDispatcherOptions`](xref:signalr/configuration#configure-server-options) object configured in `MapHub`:
+
+* `ApplicationMaxBufferSize` represents the maximum number of bytes from the client that the server will buffer. If the client attempts to send a message larger than this limit, the connection may be closed.
+* `TransportMaxBufferSize` represents the maximum number of bytes the server can send. If the server attempts to send a message (includes return values from hub methods) larger than this limit, an exception will be thrown.
+
+Setting the limit to `0` will disable the limit entirely. However, this should be done with extreme caution. Removing the limit allows a client to send a message of any size and the server will buffer it. This could be used by a malicious client to cause excess memory to be allocated, which could dramatically reduce the number of concurrent connections your app can support.
```
scottaddie: - will disable --> disables
- will buffer --> buffers

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,70 @@
...
+
+There are separate limits for incoming and outgoing messages, both can be configured on the [`HttpConnectionDispatcherOptions`](xref:signalr/configuration#configure-server-options) object configured in `MapHub`:
+
+* `ApplicationMaxBufferSize` represents the maximum number of bytes from the client that the server will buffer. If the client attempts to send a message larger than this limit, the connection may be closed.
```
scottaddie: will buffer --> buffers

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Buffer management
+
+SignalR uses per-connection buffers in order to manage incoming and outgoing messages. By default, SignalR limits these buffers to 32KB. This means the largest possible message a client or server can send is 32KB. This also means the maximum amount of memory consumed by a connection for messages is 32KB. If you know your messages will always be smaller than this limit, you can reduce this size to prevent a client from being able to send a larger message and force the server to allocate memory to accept it. Similarly, if you know messages can be larger than this limit, you can increase it. However, be aware that increasing this limit means that the client will be able to cause the server to allocate additional memory and may reduce the number of concurrent connections your app can handle.
```
scottaddie: - will always be --> is always
- will be able to --> can

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,70 @@
...
+}
+```
+
+### Access Token Logging
+
+Due to limitations in browser APIs, when using WebSockets or Server-Sent Events, the access token must be specified in the query string. This is generally as secure as using the standard `Authorization` header, however many web servers log the URL for each request, including the query string. This means the access token may be included in logs. Consider reviewing the web server's logging settings to avoid logging this information.
+
+### Exceptions
+
+Exception messages are generally considered sensitive data that should not be revealed to a client. By default SignalR doesn't send the details of an exception thrown by a hub method to the client. Instead, the client receives a generic message indicating an error occurred. You can override this behavior by setting the [`EnableDetailedErrors`](xref:signalr/configuration#configure-server-options) setting.
```
scottaddie: - should not --> shouldn't
- Add a comma after "By default"

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,70 @@
...
+
+    // ... other middleware ...
+
+    app.UseSignalR();
+
+    // ... other middleware ...
+}
+```
+
+### Access Token Logging
```
scottaddie: Access Token Logging --> Access token logging

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,70 @@
...
+[Cross-origin resource sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) can be used to allow cross-origin SignalR connections in the browser. If your JavaScript code is hosted on a different domain name from your SignalR app, you will have to enable the [ASP.NET Core CORS middleware](xref:security/cors) in order to allow the connection. In general, allow cross-origin requests only from domains you control. For example, if your site is hosted on `http://www.example.com` and your SignalR app is hosted on `http://signalr.example.com`, you should configure CORS in your SignalR app to only allow the origin `www.example.com`.
+
+For more information on configuring CORS, see [the documentation on ASP.NET Core CORS](xref:security/cors). SignalR requires the following CORS policies in order to operate correctly:
+
+* The policy must allow the specific origins you expect, or allow any origin (not recommended).
+* HTTP methods `GET` and `POST` must be allowed.
+* Credentials must be enabled, even when you are not using authentication.
+
+For example, the following CORS policy will allow a SignalR browser client hosted on `http://example.com` to access your SignalR app:
```
scottaddie: will allow --> allows

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Cross-Origin Resource Sharing
+
+[Cross-origin resource sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) can be used to allow cross-origin SignalR connections in the browser. If your JavaScript code is hosted on a different domain name from your SignalR app, you will have to enable the [ASP.NET Core CORS middleware](xref:security/cors) in order to allow the connection. In general, allow cross-origin requests only from domains you control. For example, if your site is hosted on `http://www.example.com` and your SignalR app is hosted on `http://signalr.example.com`, you should configure CORS in your SignalR app to only allow the origin `www.example.com`.
+
+For more information on configuring CORS, see [the documentation on ASP.NET Core CORS](xref:security/cors). SignalR requires the following CORS policies in order to operate correctly:
+
+* The policy must allow the specific origins you expect, or allow any origin (not recommended).
+* HTTP methods `GET` and `POST` must be allowed.
+* Credentials must be enabled, even when you are not using authentication.
```
scottaddie: are not --> aren't

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,70 @@
...
+
+By [Andrew Stanton-Nurse](https://twitter.com/anurse)
+
+## Overview
+
+SignalR provides a number of security protections by default, but it is important to understand their limitations.
+
+### Cross-Origin Resource Sharing
+
+[Cross-origin resource sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) can be used to allow cross-origin SignalR connections in the browser. If your JavaScript code is hosted on a different domain name from your SignalR app, you will have to enable the [ASP.NET Core CORS middleware](xref:security/cors) in order to allow the connection. In general, allow cross-origin requests only from domains you control. For example, if your site is hosted on `http://www.example.com` and your SignalR app is hosted on `http://signalr.example.com`, you should configure CORS in your SignalR app to only allow the origin `www.example.com`.
```
scottaddie: - you will --> you
- Add " (CORS)" after the "Cross-origin resource sharing" link in the 1st sentence

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,70 @@
...
+uid: signalr/security
+---
+
+# Security Concepts in ASP.NET Core SignalR
+
+By [Andrew Stanton-Nurse](https://twitter.com/anurse)
+
+## Overview
+
+SignalR provides a number of security protections by default, but it is important to understand their limitations.
```
scottaddie: Split this into 2 sentences. I suggest the following:
"SignalR provides a number of security protections by default. It's important to understand their limitations."

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,70 @@
...
+author: rachelappel
+description: Learn how to use authentication and authorization in ASP.NET Core SignalR.
+monikerRange: '>= aspnetcore-2.1'
+ms.author: rachelap
+ms.custom: mvc
+ms.date: 06/29/2018
+uid: signalr/security
+---
+
+# Security Concepts in ASP.NET Core SignalR
```
scottaddie: Do you prefer the word "concepts" or "considerations" here? The title metadata above should match.

aspnetcore/signalr/security.md
```diff
@@ -0,0 +1,70 @@
+---
+title: Security Considerations in ASP.NET Core SignalR
```
scottaddie: Considerations --> considerations

aspnetcore/signalr/authn-and-authz/sample/wwwroot/js/chat.js
anurse: Yeah, I can do that here. Generally we've tried to ensure there were JavaScript versions of the samples but that's difficult in this case since it's a much more complicated sample.

aspnetcore/toc.md
anurse: I haven't rebased since the ToC reorganization. I'll do that.

aspnetcore/signalr/security.md
anurse: Derp. Yeah I like considerations, I'll make them consistent.

aspnetcore/toc.md
```diff
@@ -256,6 +256,8 @@
 ### [.NET client](xref:signalr/dotnet-client)
 ### [WebPack and TypeScript](xref:tutorials/signalr-typescript-webpack)
 ## [Configuration](xref:signalr/configuration)
+## [Authentication and authorization](xref:signalr/authn-and-authz)
+## [Security considerations](xref:signalr/security)
```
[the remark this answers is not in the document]
scottaddie: I like the idea of the "Security" node. What if you just shorten to ToC text to "Considerations"?

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,119 @@
+---
+title: Authentication and authorization in SignalR
```
scottaddie: SignalR --> ASP.NET Core SignalR

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -0,0 +1,119 @@
...
+author: rachelappel
+description: Learn how to use authentication and authorization in ASP.NET Core SignalR.
+monikerRange: '>= aspnetcore-2.1'
+ms.author: anurse
+ms.custom: mvc
+ms.date: 06/29/2018
+uid: signalr/authn-and-authz
+---
+
+# Authentication and authorization in SignalR
```
scottaddie: SignalR --> ASP.NET Core SignalR

Rick-Anderson:
@anurse  do you need any help getting this approved?

mikaelm12:
Andrew is out for the next week. It looks like there are only minor changes left I'm sure he wouldn't mind someone just knocking those out and merging this though 😄

mikaelm12:
Thanks for finishing this up @Rick-Anderson!

Rick-Anderson:
@mikaelm12  it's live [here](https://docs.microsoft.com/en-us/aspnet/core/signalr/authn-and-authz?view=aspnetcore-2.1)

PR #7329, opened 2018-06-27
Update customize_identity_model.md

Rick-Anderson:


Rick-Anderson:
@scottaddie  it's consistent with the filename. I much prefer _ to - in filenames, but we've started down the wrong path on that one.

PR #7331, opened 2018-06-27
L25985: Adding space between !INCLUDE and []

v-maudel:
Hello, @rick-anderson,
"INCLUDE syntax is incorrect. There should be a space before title bracket."

Many thanks in advance.

### DIFF
--- a/aspnetcore/getting-started/index.md
+++ b/aspnetcore/getting-started/index.md
@@ -11,7 +11,7 @@ uid: getting-started
 
 ::: moniker range=">= aspnetcore-2.1"
 
-1. Install the [!INCLUDE[](~/includes/2.1-SDK.md)].
+1. Install the [!INCLUDE [](~/includes/2.1-SDK.md)].
 
 2. Create an ASP.NET Core project. Open a command shell and enter the following command:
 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
cc @rachelappel  @guardrex  @scottaddie  "INCLUDE syntax is incorrect. There should be a space before title bracket."

guardrex:
Is it for localization purposes? According to DocFX, it's ok ...

https://dotnet.github.io/docfx/spec/docfx_flavored_markdown.html?tabs=tabid-1%2Ctabid-a#block

Their example:

```
[!include[<title>](<filepath>)]
```

v-maudel:
Hi @guardrex , yes, this issue was reported by localization team. 
In the link you provide it seems correct to use it as it, although according to the Markdown reference for OPS, it should have a space between !INCLUDE statement and title. 
https://review.docs.microsoft.com/en-us/help/contribute/markdown-reference?branch=master#included-markdown-files

Maybe the issue comes because it is written in uppercase and without the space instead in lowercase without the space?

Please, let us know if it doesn't apply for this file. 

CC: @Rick-Anderson 

Regards.

v-maudel:
Thanks for your investigation @guardrex ,
It seems both syntax were used for the include statement and they display correctly in the live page.

PR #7347, opened 2018-06-28
Fixed invalid variable name in generic host snippet

roketworks:
Snippet `ConfigureAppConfiguration` in the generic host fundamental page was referencing an invalid variable name in the context of the scope. Updated to use the correct variable name.


scottaddie:
@roketworks Nice catch! Thank you for the PR.

PR #7374, opened 2018-06-29
Enhance service creation content

### LINKED ISSUE #7373: Hosting in a .Net Core 2.1 Windows Service on Windows Server 2016
I've followed the instructions here to the letter and my .Net Core 2.1 Windows Service works perfectly when I create and start the service on my Windows 10 Pro machine.

The problem arises when I deploy on to our Windows Server 2016 Datacenter OS. It creates the service as expected, but when I go to start the service I immediately get the following error:

The service did not respond to the start or control request in a timely fashion.

To emphasize, the service instantly fails to start with the above error.

If I try to run the exe itself I get the following error:

Error:
An assembly specified in the application dependencies manifest (TestService.deps.json) was not found:
package: 'Microsoft.AspNetCore.Hosting.WindowsServices', version: '2.1.1'
path: 'lib/netstandard2.0/Microsoft.AspNetCore.Hosting.WindowsServices.dll'

I suspect the two error are interlinked somehow.

I'm fairly convinced that the code is not being hit in Program.Main as I've added logging in the first line of code to be ran but nothing happens. I've tried adding the ServicesPipeTimeout registry entry to extend the time given for the service to start but I still get an instant fail. I've ensured that the server has the required versions of .Net Core installed and they are:

Microsoft. NET Core 2.1.1 - Windows Server Hosting - v2.1.13007.0
Microsoft .NET Core Runtime - 2.1.1 (x64) - 2.1.1.26606
I've also ensured the IIS is enabled on the system.

I've been struggling with this for 2 days and cannot find the answers to resolve these issues anywhere.

Can this document be updated with instructions on how to deploy to the server to ensure all the pre-requisites are covered so that these issues don't occur?

I'd really like an answer to this, thanks in advance :)

---
#### Document Details

* Content: [Host ASP.NET Core in a Windows Service](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/windows-service?view=aspnetcore-2.1)
* Content Source: [aspnetcore/host-and-deploy/windows-service.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/host-and-deploy/windows-service.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex

guardrex:
Fixes #7373 

@scottaddie In this case, I'm making an exception to our general rule about not specifying the TF explicitly because I want it to be utterly clear what the path will look like. For the example prereqs, I've added a bullet to say that the project is published for `netcoreapp2.1`. [EDIT] The TF will be updated at each sample update, so I'm not worried about it for the future of the topic.

Also, I want to reinforce the `binPath` argument space requirement, so I repeat the following line on purpose for this new example ...

> **Make sure the space is present between the `binPath=` argument and its value.**

It's an easily overlooked instruction that turns into a big 'ole :bomb: if it isn't seen by the reader.

Thanks to @mmaguirehub for reporting! :rocket:

### DIFF
--- a/aspnetcore/host-and-deploy/windows-service.md
+++ b/aspnetcore/host-and-deploy/windows-service.md
@@ -61,7 +61,7 @@ The following minimum changes are required to set up an existing ASP.NET Core pr
    For the sample app and command that follows, the service is:
 
    * Named **MyService**.
-   * Published to *c:\\svc* folder.
+   * Published to the *c:\\svc* folder.
    * Has an app executable named *AspNetCoreService.exe*.
 
    Open a command shell with administrative privileges and run the following command:
@@ -76,7 +76,7 @@ The following minimum changes are required to set up an existing ASP.NET Core pr
    
    * Named **MyService**.
    * Created from a project that sets a target framework to `netcoreapp2.1`.
-   * Published to *c:\\my_services\\AspNetCoreService\\bin\\Release\\netcoreapp2.1\\publish* folder.
+   * Published to the *c:\\my_services\\AspNetCoreService\\bin\\Release\\netcoreapp2.1\\publish* folder.
    * Has an app executable named *AspNetCoreService.exe*.
 
    Open a command shell with administrative privileges and run the following command:
--- a/aspnetcore/host-and-deploy/windows-service.md
+++ b/aspnetcore/host-and-deploy/windows-service.md
@@ -48,32 +48,22 @@ The following minimum changes are required to set up an existing ASP.NET Core pr
 
      ::: moniker-end
 
-1. Publish the app to a folder. Use [dotnet publish](/dotnet/articles/core/tools/dotnet-publish) or a [Visual Studio publish profile](xref:host-and-deploy/visual-studio-publish-profiles) that publishes to a folder.
+1. Publish the app. Use [dotnet publish](/dotnet/articles/core/tools/dotnet-publish) or a [Visual Studio publish profile](xref:host-and-deploy/visual-studio-publish-profiles).
 
    To publish the sample app from the command line, run the following command in a console window from the project folder:
 
    ```console
-   dotnet publish --configuration Release --output c:\svc
+   dotnet publish --configuration Release
    ```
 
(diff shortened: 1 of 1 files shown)


aspnetcore/host-and-deploy/windows-service.md
```diff
@@ -61,14 +61,29 @@ The following minimum changes are required to set up an existing ASP.NET Core pr
...
+   * Published to the *c:\\svc* folder.
+   * Represented by an app executable named *AspNetCoreService.exe*.
 
    Open a command shell with administrative privileges and run the following command:
 
    ```console
    sc create MyService binPath= "c:\svc\aspnetcoreservice.exe"
    ```
+   
+   **Make sure the space is present between the `binPath=` argument and its value.**
```
scottaddie: Since this is an important detail, I suggest putting it in an IMPORTANT widget to make it pop. If I were scanning the doc, I may miss this.

guardrex: There's a risk that readers read past them. I wish we had some hard data on it. 

I'll make that update when I get back, and something else has occurred to me. The use of the `--output` arg goes back to the original version of the topic. I'd rather only show one publish action here and just tell the reader briefly that they can use that arg and link over to it. I'm not feel'in two examples here upon reflection.

guardrex:
@scottaddie Let's see how that looks :point_up:. I'm leaning toward this version that doesn't repeat information. I still call out the `--output` option in case the reader would like to publish to a different folder.

aspnetcore/host-and-deploy/windows-service.md
```diff
@@ -48,29 +48,40 @@ The following minimum changes are required to set up an existing ASP.NET Core pr
...
+   ```console
+   sc create <SERVICE_NAME> binPath= "<PATH_TO_SERVICE_EXECUTABLE>"
+   ```
+
+   For a service published in the project folder, use the path to the *publish* folder to create the service. In the following example, the service is:
 
    * Named **MyService**.
-   * Published to *c:\\svc* folder.
-   * Has an app executable named *AspNetCoreService.exe*.
+   * Created from a project that sets the target framework to `netcoreapp2.1`.
```
scottaddie: Does the `netcoreapp2.1` have to be used? Is there a way to future-proof this text?

guardrex:
Before, I had it remarking explicitly for the sample app, which is `netcoreapp2.1`. Now that I've removed that, let's use our usual rule to go with a placeholder.

PR #7376, opened 2018-06-29
Path missing "Shared" folder

juricav:
The tutorial page has a typo in the path.

### DIFF
--- a/aspnetcore/includes/RP/page1.md
+++ b/aspnetcore/includes/RP/page1.md
@@ -84,7 +84,7 @@ The `PageModel` base class has a `ViewData` dictionary property that can be used
 
 ::: moniker range="= aspnetcore-2.0"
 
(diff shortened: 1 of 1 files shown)


scottaddie:
The ASP.NET Core 2.0 project template didn't use a *Pages/Shared* folder. The *_Layout.cshtml* file lived in the root of *Pages*. I do, however, see what would've led the reader to believe otherwise. I'll keep this issue open to address that problem.

Rick-Anderson:
@scottaddie  you need to select 2.1 to see the 2.1 layout. What's the problem? This PR has the layout file in Shared, which is wrong for 2.0

scottaddie:
@Rick-Anderson Please see the linked PR. Looks like the 2.1 include was being used when 2.0 was selected.

juricav:
I may have taken the wrong file.
But the 2.1 layout is selected and there's not Shared folder.
https://docs.microsoft.com/en-us/aspnet/core/tutorials/razor-pages/page?view=aspnetcore-2.1#update-the-layout

PR #7390, opened 2018-06-30
updated the way the builder was used for UseSolutionRelativeContentRoot

mozts2005:
just moved to so to ensure it will run.


guardrex:
Hi @mozts2005 ... thanks for offering; but when `ConfigureWebHost` is called, the path will be set with the statement in its current position.

Issue #6312, opened 2018-05-09
Expand DI documentation with additional extension method overloads


stevejgordon:
I've recently been exploring and documenting some more "advanced" DI registration examples on my blog. I've had a few occasions where I've found that people are not aware of the additional method overloads they can use when calling AddSingleton, AddTransient etc.

I briefly chatted with @scottaddie about this gap and as agreed I'm raising an issue to discuss this further.

The current documentation I've found is in the [fundamentals section here](https://docs.microsoft.com/en-gb/aspnet/core/fundamentals/dependency-injection).

I'm opening this issue to discuss if any of the following scenarios should be considered for inclusion...

- [ASP.NET Core Dependency Injection – How to Register Generic Types](https://www.stevejgordon.co.uk/asp-net-core-dependency-injection-how-to-register-generic-types)
- [ASP.NET Core Dependency Injection – Registering Multiple Implementations of an Interface](https://www.stevejgordon.co.uk/asp-net-core-dependency-injection-registering-multiple-implementations-interface)
- [ASP.NET Core Dependency Injection – Registering Implementations Using Delegates](https://www.stevejgordon.co.uk/asp-net-core-dependency-injection-registering-implementations-using-delegates)

I accept there is a balance between showing all of the API vs making it too overwhelming for less experienced developers. I'm not sure if these would serve best as a section nearer the end of that current document, or in their own more advanced doc?

My experience to date (which may not extend to the wider .NET audience) is that people are not naturally discovering these capabilities when exploring the API directly.

If this is agreed for inclusion I'm happy to assist by reworking some of my examples and content for the docs.

guardrex:
IMHO ... Do this in an advanced topic. The current topic is already fairly long and loaded with info.

> overwhelming for less experienced developers

Exactly ... in a separate topic it would be easy to call out early that these are advanced DI registration concepts. Makes it clear to the new dev or new-to-Core dev that they don't strictly need to consume this content just to be successful with ASP.NET Core. It might not be so easy to make that distinction if this turns out to be a (large?) section in the current topic.

scottaddie:
@danroth27 Thoughts on what type of content you'd like Steve to contribute?

Rick-Anderson:
@stevejgordon you've been approved by non other than `@davidfowl`

stevejgordon:
[...]

PR #7412, opened 2018-07-02
Dependency injection topic and sample updates

### LINKED ISSUE #6331: Dependency Injection - Behaviour on constructor overload selection needs update.
The document said that constructor overloads are supported only when no more than one overloaded constructor can be fulfilled, but this is not exactly correct as of .NET Core 2.0 (I am not really sure when it started).

Here is the current text:

> Constructor injection requires that only one applicable constructor exist. Constructor overloads are supported, but **only one overload can exist whose arguments can all be fulfilled by dependency injection**. If more than one exists, your app will throw an InvalidOperationException
 
https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/dependency-injection.md

However, the latest version of DI allow us to have multiple constructor overloads which their set of parameters are subset/superset of each other and fulfil the largest parameter set possible. For example, we can have

class Foo {
    Foo(IBar bar) { }
    Foo(IBar bar, IOptional optionalService) { }
}

If `IOptional` is registered, then both overloads can be fulfilled hence it should throw an exception if the document were accurate. However, the current version of DI will not throw exception but will choose second overload because its parameter set is the superset of all other applicable overloads.

The reason other than making this accurate is that it can be used to inject optional service, which I believe is useful for many developers.

### LINKED ISSUE #7208: Expand thread safety section with respect to singleton factory
The thread-safety section talks about services:

> ### Thread safety
>
> Singleton services need to be thread safe. If a singleton service has a dependency on a transient service, the transient service may also need to be thread safe depending how it's used by the singleton.

However, it would be help to add a word or two about the thread-safety of a _singleton factory_ when using something like [`AddSingleton<TService>(IServiceCollection, Func<IServiceProvider,TService>)`][AddSingleton], as in:

```c#
services.AddSingleton<IFoo>(sp =>
{
    // initialization code...
    return new Foo(/* ... */);
});
```

One would imagine that the factory will be called lazily, when the service is requested for the first time. What happens, though, if multiple threads request the singleton service simultaneously that in turn incurs its initialization?

1. Does the initialization code run once and run by one thread alone, like a static type constructor?
2. Does the factory need to be _idempotent_ and so may be called more than once, by more than one thread? The assumption here is that one of those will win (rest discarded) but will effectively return the same result that's used for all subsequent requests.

I'm guessing it's the first case but it would help to clarify in that section. It avoids:

- people being paranoid and adding locks in their factory code.
- asking Luke to [use The Source](http://wiki.c2.com/?UseTheSourceLuke). 😉 

[AddSingleton]: https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.servicecollectionserviceextensions.addsingleton?view=aspnetcore-2.1#Microsoft_Extensions_DependencyInjection_ServiceCollectionServiceExtensions_AddSingleton__1_Microsoft_Extensions_DependencyInjection_IServiceCollection_System_Func_System_IServiceProvider___0__

---
#### Document Details

* Content: [Dependency injection in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.1#recommendations)
* Content Source: [aspnetcore/fundamentals/dependency-injection.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/dependency-injection.md)
* Product: **aspnet-core**
* GitHub Login: @ardalis
* Microsoft Alias: **riande**

guardrex:
Addresses #5495 
Fixes #7208 
Fixes #6331
Addresses #7700

[Dependency injection topic](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.1&branch=pr-en-us-7412)
[Repository pattern topic](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/repository-pattern?view=aspnetcore-2.1&branch=pr-en-us-7412)

* Upgrade 1.0 MVC sample to 1.1.
* Shiny new 2.1 RP sample. :tada:
* The usual suspects :suspect: ... grammar, style, API links, tabs out, etc.
* Replace the repository pattern example of DI with a simpler example. Move the repository pattern to its own topic. Repository pattern is overkill in the DI topic, and we aren't using it in this doc set per the guidance of `@danroth27`. However, it's a great pattern, so I hate to lose the content. It composes well into it's own topic with its own little sample app cross-linked back and forth to the DI topic. The DI example I use here is easy to grok and makes it simple to ........
* Employ the :zap: **_Wasson Gambit_**:tm: :zap: ... i.e., take the time and space to explain DI with a clear example showing why hardcoded dependencies are bad before explaining why DI is better. Tip of the 🎩 to @MikeWasson for https://docs.microsoft.com/aspnet/web-api/overview/advanced/dependency-injection#what-is-dependency-injection. I looked at quite a few DI explanations around the Net, and I think your approach is **THE BEST**. You may be surprised how many DI explanations were apparently written for 🔪 *Halloween readings to frighten young children!* 🎃
* I retain the "Operations" services example from the current topic, but I modify the explanations a bit to (hopefully) make it a bit clearer what it's showing, especially the `OperationService`.


guardrex:
@Rick-Anderson @scottaddie @rachelappel I think it best if you look this over prior to engineering. I'm taking this in a new direction with a major overhaul to the introduction and how DI is explained. It's probably best for you to say if you like it ... or issue a round of :skull: *Rex death threats* :skull: :smile: ... before engineering takes a look.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -3,139 +3,202 @@ title: Dependency injection in ASP.NET Core
...
+ms.date: 07/02/2018
 uid: fundamentals/dependency-injection
 ---
 # Dependency injection in ASP.NET Core
 
-<a name="fundamentals-dependency-injection"></a>
+By [Steve Smith](https://ardalis.com/), [Scott Addie](https://scottaddie.com), and [Luke Latham](https://github.com/guardrex)
 
-By [Steve Smith](https://ardalis.com/) and [Scott Addie](https://scottaddie.com)
+ASP.NET Core supports the dependency injection (DI) software design pattern, which is a technique for achieving [Inversion of Control (IoC)](http://deviq.com/inversion-of-control/) between classes and their dependencies.
```
scottaddie: http --> https

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -3,139 +3,202 @@ title: Dependency injection in ASP.NET Core
...
+[!code-csharp[](dependency-injection/samples/1.x/DependencyInjectionSample/Services/MyDependency.cs?name=snippet1)]
 
-[!code-csharp[](../fundamentals/dependency-injection/sample/DependencyInjectionSample/Models/CharacterRepository.cs?highlight=9,11,12,13,14)]
+::: moniker-end
 
-Note that `CharacterRepository` requests an `ApplicationDbContext` in its constructor. It's not unusual for dependency injection to be used in a chained fashion like this, with each requested dependency in turn requesting its own dependencies. The container is responsible for resolving all of the dependencies in the graph and returning the fully resolved service.
+`MyDependency` requests an [ILogger&lt;TCategoryName&gt;](/dotnet/api/microsoft.extensions.logging.ilogger-1) in its constructor. It's not unusual to use dependency injection in a chained fashion with each requested dependency in turn requesting its own dependencies. The container resolves the dependencies in the graph and returns the fully resolved service. The collective set of dependencies that must be resolved is typically referred to as a *dependency tree*, *dependency graph*, or *object graph*.
```
scottaddie: Add a comma after "in a chained fashion".

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -3,139 +3,202 @@ title: Dependency injection in ASP.NET Core
...
+
+`IMyDependency` and `ILogger<TCategoryName>` must be registered in the service container. `IMyDependency` is registered in `Startup.ConfigureServices`. `ILogger<TCategoryName>` is registered by the logging abstractions infrastructure, so it's a [framework-provided-service](#framework-provided-services) registered by default by the framework.
```
scottaddie: Remove the dash between "provided" and "service".

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -3,139 +3,202 @@ title: Dependency injection in ASP.NET Core
...
+If a service's constructor requires a primitive, such as a `string`, this can be injected by using [configuration](xref:fundamentals/configuration/index) and the [options pattern](xref:fundamentals/configuration/options).
 
-[!code-csharp[](dependency-injection/sample/DependencyInjectionSample/Startup.cs?highlight=3-5,11&range=16-32)]
+An instance of the service is requested via the constructor of a class where the service is used and assigned to a private field. The field is used to access the service as necessary throughout the class.
 
-Entity Framework contexts should be added to the services container using the `Scoped` lifetime. This is taken care of automatically if you use the helper methods as shown above. Repositories that will make use of Entity Framework should use the same lifetime.
+In the following example class from the sample app, the `IMyDependency` instance is requested and used to call the service's `WriteMessage` method:
```
scottaddie: example class --> class

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -146,45 +209,166 @@ Transient lifetime services are created each time they're requested. This lifeti
...
+> When using a scoped service in a middleware, inject the service into the `Invoke` or `InvokeAsync` method. Don't inject via constructor injection because it forces the service to behave like a singleton. For more information, see [Middleware](xref:fundamentals/middleware/index).
 
 **Singleton**
 
-Singleton lifetime services are created the first time they're requested (or when `ConfigureServices` is run if you specify an instance there) and then every subsequent request will use the same instance. If your application requires singleton behavior, allowing the services container to manage the service's lifetime is recommended instead of implementing the singleton design pattern and managing your object's lifetime in the class yourself.
+Singleton lifetime services are created the first time they're requested (or when `ConfigureServices` is run if an instance is specified there) and then every subsequent request uses the same instance. If the app requires singleton behavior, allowing the service container to manage the service's lifetime is recommended. Don't implement the singleton design pattern and provide user code to manage the object's lifetime in the class.
```
scottaddie: Add a comma before "if an instance is".

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -146,45 +209,166 @@ Transient lifetime services are created each time they're requested. This lifeti
...
+
+Constructor injection requires a *public* constructor. Otherwise, the app throws an [InvalidOperationException](/dotnet/api/system.invalidoperationexception):
+
+> A suitable constructor for type '&lt;TYPE&gt;' couldn't be located. Ensure the type is concrete and services are registered for all parameters of a public constructor.
+
+Constructor injection requires that only one applicable constructor exist. Constructor overloads are supported, but only one overload can exist whose arguments can all be fulfilled by dependency injection. If more than one exists, the app throws an `InvalidOperationException`:
+
+> Multiple constructors accepting all given argument types have been found in type '&lt;TYPE&gt;'. There should only be one applicable constructor.
+
+Constructors can accept arguments that are not provided by dependency injection, but these must support default values.
```
scottaddie: are not --> aren't

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -146,45 +209,166 @@ Transient lifetime services are created each time they're requested. This lifeti
...
+Singleton: 01271bc1-9e31-48e7-8f7c-7261b040ded9  
+Instance: 00000000-0000-0000-0000-000000000000
+
+Observe which of the `OperationId` values vary within a request and between requests:
+
+* *Transient* objects are always different. Note that the transient `OperationId` value for both the first and second requests are different for both `OperationService` operations and across requests. A new instance is provided to each service and request.
+* *Scoped* objects are the same within a request but different across requests.
+* *Singleton* objects are the same for every object and every request regardless of whether an `Operation` instance is provided in `ConfigureServices`.
 
 ## Resolve a scoped service within the application scope
```
scottaddie: application --> app

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -230,131 +414,129 @@ For more information, see [Scope validation in the Web Host topic](xref:fundamen
...
+Design services to use dependency injection to obtain their dependencies. Avoid the use of stateful static method calls (a bad practice known as [static cling](http://deviq.com/static-cling/)) and the direct instantiation of dependent classes within services. Direct instantiation "glues" the code to a particular implementation. It may help to remember the phrase, [New is Glue](https://ardalis.com/new-is-glue), when choosing whether to instantiate a type or to request it via dependency injection. By following the [SOLID Principles of Object Oriented Design](http://deviq.com/solid/), app classes naturally tend to be small, well-factored, and easily tested.
```
scottaddie: - Add a comma between "stateful" and "static".
- http --> https

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -230,131 +414,129 @@ For more information, see [Scope validation in the Web Host topic](xref:fundamen
...
 
-With regards to data access specifically, you can inject the `DbContext` into your controllers (assuming you've added EF to the services container in `ConfigureServices`). Some developers prefer to use a repository interface to the database rather than injecting the `DbContext` directly. Using an interface to encapsulate the data access logic in one place can minimize how many places you will have to change when your database changes.
+If a class seems to have too many injected dependencies, this is generally a sign that the class has too many responsibilities and is probably violating the [Single Responsibility Principle (SRP)](http://deviq.com/single-responsibility-principle/). Attempt to refactor the class by moving some of its responsibilities into a new class. Keep in mind that Razor Pages page model classes and MVC controller classes should be focused on UI concerns. Business rules and data access implementation details should be kept in classes appropriate to these [separate concerns](http://deviq.com/separation-of-concerns/).
```
scottaddie: - probably violating --> violating
- http --> https
- should be focused on --> should focus on

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -230,131 +414,129 @@ For more information, see [Scope validation in the Web Host topic](xref:fundamen
...
+    // The container doesn't create the following instances, so it doesn't dispose of
+    // the instances automatically:
     services.AddSingleton<Service3>(new Service3());
     services.AddSingleton(new Service3());
 }
 ```
 
 > [!NOTE]
-> In version 1.0, the container called dispose on *all* `IDisposable` objects, including those it didn't create.
+> In ASP.NET Core 1.0, the container calls dispose on *all* `IDisposable` objects, including those it didn't create.
```
scottaddie: Display this note only when the 1.0 version of the docset is selected.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -230,131 +414,129 @@ For more information, see [Scope validation in the Web Host topic](xref:fundamen
...
+    }
+    ```
 
-> [!NOTE]
-> When using a third-party DI container, you must change `ConfigureServices` so that it returns `IServiceProvider` instead of `void`.
+    > [!NOTE]
+    > When using a third-party dependency injection container, change `Startup.ConfigureServices` so that it returns `IServiceProvider` instead of `void`.
 
-Finally, configure Autofac as normal in `DefaultModule`:
+3. Configure Autofac normally in `DefaultModule`:
```
scottaddie: I'm confused by the word "normally" in this context. Is it needed?

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -230,131 +414,129 @@ For more information, see [Scope validation in the Web Host topic](xref:fundamen
...
+        {
+            builder.RegisterType<CharacterRepository>().As<ICharacterRepository>();
+        }
     }
-}
-```
+    ```
 
-At runtime, Autofac will be used to resolve types and inject dependencies. [Learn more about using Autofac and ASP.NET Core](http://docs.autofac.org/en/latest/integration/aspnetcore.html).
+At runtime, Autofac is used to resolve types and inject dependencies. To learn more about using Autofac with ASP.NET Core, see the [Autofac documentation](http://docs.autofac.org/en/latest/integration/aspnetcore.html).
```
scottaddie: http --> https

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -230,131 +414,129 @@ For more information, see [Scope validation in the Web Host topic](xref:fundamen
...
 * [Dependency injection into views](xref:mvc/views/dependency-injection)
 * [Dependency injection into controllers](xref:mvc/controllers/dependency-injection)
 * [Dependency injection in requirement handlers](xref:security/authorization/dependencyinjection)
+* [Repository pattern](xref:fundamentals/repository-pattern)
 * [Application Startup](xref:fundamentals/startup)
 * [Test and debug](xref:test/index)
 * [Factory-based middleware activation](xref:fundamentals/middleware/extensibility)
 * [Writing Clean Code in ASP.NET Core with Dependency Injection (MSDN)](https://msdn.microsoft.com/magazine/mt703433.aspx)
 * [Container-Managed Application Design, Prelude: Where does the Container Belong?](https://blogs.msdn.microsoft.com/nblumhardt/2008/12/26/container-managed-application-design-prelude-where-does-the-container-belong/)
 * [Explicit Dependencies Principle](http://deviq.com/explicit-dependencies-principle/)
```
scottaddie: http --> https

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -230,131 +414,129 @@ For more information, see [Scope validation in the Web Host topic](xref:fundamen
...
 
-Dependency injection is an *alternative* to static/global object access patterns. You may not be able to realize the benefits of DI if you mix it with static object access.
+Dependency injection is an *alternative* to static/global object access patterns. You may not be able to realize the benefits of dependency injection if you mix it with static object access.
 
 ## Additional resources
 
 * [Dependency injection into views](xref:mvc/views/dependency-injection)
 * [Dependency injection into controllers](xref:mvc/controllers/dependency-injection)
 * [Dependency injection in requirement handlers](xref:security/authorization/dependencyinjection)
+* [Repository pattern](xref:fundamentals/repository-pattern)
```
scottaddie: Prefer the <xref:uid> pattern for these additional resources links.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -3,139 +3,202 @@ title: Dependency injection in ASP.NET Core
...
+> Each `services.Add<ServiceName>` extension method adds (and potentially configures) services. For example, `services.AddMvc()` adds the services Razor Pages and MVC require. We recommended that apps follow this convention. Place extension methods in the [Microsoft.Extensions.DependencyInjection](/dotnet/api/microsoft.extensions.dependencyinjection) namespace to encapsulate groups of service registrations.
 
-In this case, both `ICharacterRepository` and in turn `ApplicationDbContext` must be registered with the services container in `ConfigureServices` in `Startup`. `ApplicationDbContext` is configured with the call to the extension method `AddDbContext<T>`. The following code shows the registration of the `CharacterRepository` type.
+If a service's constructor requires a primitive, such as a `string`, this can be injected by using [configuration](xref:fundamentals/configuration/index) and the [options pattern](xref:fundamentals/configuration/options).
```
Rick-Anderson: this can be injected 
What is this? 
The primative can be .. or The string can be

guardrex:
Thx @scottaddie ... what do you think of the overall direction taken by the PR?

aspnetcore/fundamentals/dependency-injection.md
guardrex: 😕 They do. What do you mean?

aspnetcore/fundamentals/dependency-injection.md
Rick-Anderson: In the following example ~class from the sample app~

aspnetcore/fundamentals/dependency-injection.md
guardrex: No comma between prepositional phrases. However, let's break the sentence here. That's really better in this case for localization anyway.

> It's not unusual to use dependency injection in a chained fashion. Each requested dependency in turn requests its own dependencies.

aspnetcore/fundamentals/dependency-injection.md
guardrex: > I'm confused by the word "normally" in this context. Is it needed?

I think it just fell under "conversational tone." I'm removing it now.

aspnetcore/fundamentals/dependency-injection.md
Rick-Anderson: Singleton lifetime services are created:

* The first time

Or some other way to keep sentence under 20 words.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -3,139 +3,202 @@ title: Dependency injection in ASP.NET Core
...
+```
 
-Services that have dependencies should register them in the container. If a service's constructor requires a primitive, such as a `string`, this can be injected by using [configuration](xref:fundamentals/configuration/index) and the [options pattern](xref:fundamentals/configuration/options).
+For more information, see the [ServiceCollection Class](/dotnet/api/microsoft.extensions.dependencyinjection.servicecollection) in the API documentation.
 
-## Service lifetimes and registration options
+## Service lifetimes
 
-ASP.NET services can be configured with the following lifetimes:
+It's important to choose an appropriate lifetime for each registered service. ASP.NET Core services can be configured with the following lifetimes:
```
Rick-Anderson: Services must be registered with the  appropriate lifetime. Now explain why. You can't just say important without explaining why. This is a good question to get PU help on.

guardrex:
> From VS the output is a needle in the haystack. Is there a way to remove the default debug output so you can see these messages? If you add that to the readme or the doc. If you add it to the readme mention in the doc the readme shows how to view the messages.

Yes, there is a way to filter ... but if you're referring to the inline code block in the intro where I show `Console.WriteLine`, that's for display-only purposes in the topic. The sample app doesn't implement it.

I don't want to show a dependency that calls for another dependency at this point in the topic. I do want to swap out the `Console.WriteLine` for the `ILogger` when I show how DI works. This is why I left the empty constructor in the inlined code block ... I add the `ILogger` there when DI is shown. I want that first dependency class example to look a lot like the upcoming version of it for the DI explanation.

aspnetcore/fundamentals/dependency-injection.md
guardrex: I agree. That text is a holdover from Line 103 of the original topic ...

> The `AddTransient` method is used to map abstract types to concrete services that are instantiated separately for every object that requires it. This is known as the service's *lifetime*, and additional lifetime options are described below. It's important to choose an appropriate lifetime for each of the services you register. Should a new instance of the service be provided to each class that requests it? Should one instance be used throughout a given web request? Or should a single instance be used for the lifetime of the application?

AFAIK one biggie is reducing resource demands. I need PU to help if the reasons go beyond that. I'll ask if they don't remark on it.

aspnetcore/fundamentals/dependency-injection.md
scottaddie: Sorry, my comment didn't display as expected. I meant use `<xref:mvc/views/dependency-injection>` for example. The link text is applied for you.

guardrex:
@Rick-Anderson I'll let you call the right engineering cat (or cats) when the time comes.

aspnetcore/fundamentals/dependency-injection.md
scottaddie: Also explain what's meant by *lifetime*.

guardrex:
btw @Rick-Anderson I looked that up just in case we need it. An example goes back to the first version of the `LoggerMessage` topic. This will do the trick if we ever need to filter console logging to get rid of the spew ...

```csharp
.ConfigureLogging((hostingContext, logging) =>
{
    logging.ClearProviders();
    logging.AddConfiguration(hostingContext.Configuration.GetSection("Logging"));
    logging.AddFilter("Microsoft.EntityFrameworkCore.Update", LogLevel.None);
    logging.AddFilter("Microsoft.EntityFrameworkCore.Infrastructure", LogLevel.None);
    logging.AddFilter("Microsoft.AspNetCore.Mvc.RedirectToRouteResult", LogLevel.None);
    logging.AddFilter("Microsoft.AspNetCore.Mvc.RazorPages.Internal.PageActionInvoker", LogLevel.None);
    logging.AddFilter("Microsoft.AspNetCore.StaticFiles.StaticFileMiddleware", LogLevel.None);
    logging.AddConsole();
})
```

aspnetcore/fundamentals/dependency-injection.md
guardrex: IIRC the decision that was made back in Tom days was to split the "ASP.NET Core" out by specifying the shorter manual titles. Looks like a change of direction was settled upon since then. Are we doing this for all cross-links or just for *Additional resources* section links?

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -230,131 +414,133 @@ For more information, see [Scope validation in the Web Host topic](xref:fundamen
...
+::: moniker-end
 
-The built-in services container is meant to serve the basic needs of the framework and most consumer applications built on it. However, developers can replace the built-in container with their preferred container. The `ConfigureServices` method typically returns `void`, but if its signature is changed to return `IServiceProvider`, a different container can be configured and returned. There are many IOC containers available for .NET. In this example, the [Autofac](https://autofac.org/) package is used.
+## Default service container replacement
 
-First, install the appropriate container package(s):
+The built-in service container is meant to serve the basic needs of the framework and most consumer apps built on it. However, developers can replace the built-in container with their preferred container. The `Startup.ConfigureServices` method typically returns `void`. If the method's signature is changed to return an [IServiceProvider](/dotnet/api/system.iserviceprovider), a different container can be configured and returned. There are many IoC containers available for .NET. In the following example, the [Autofac](https://autofac.org/) container is used:
```
scottaddie: Confirm with engineering, but I seem to remember something related to this wasn't possible until ASP.NET Core 1.1. As noted in the Autofac docs, "ASP.NET Core 1.1 introduced the ability to have strongly typed container configuration." Also, I have a personal 2.0 app using Autofac with a return type of `void` in `Startup.ConfigureServices`.

guardrex: ok ... will do.

guardrex:
Open subjects for PU comment:

- [ ] [Who/what/when on ConfigureServices returning IServiceProvider instead of void to enable the use of 3rd party IoC containers](https://github.com/aspnet/Docs/pull/7412#discussion_r199604596)
- [ ] [Describe why setting appropriate lifetimes is important](https://github.com/aspnet/Docs/pull/7412#discussion_r199592392) and [What is meant by *lifetime*?](https://github.com/aspnet/Docs/pull/7412#discussion_r199596537)

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -3,139 +3,202 @@ title: Dependency injection in ASP.NET Core
...
+        return View();
+    }
 }
 ```
 
-## Using framework-provided services
+::: moniker-end
 
-The `ConfigureServices` method in the `Startup` class is responsible for defining the services the application will use, including platform features like Entity Framework Core and ASP.NET Core MVC. Initially, the `IServiceCollection` provided to `ConfigureServices` has the following services defined (depending on [how the host was configured](xref:fundamentals/host/index)):
+The class creates and directly depends on the `MyDependency` instance. However, it's a bad practice to hardcode the dependency in this way for the following reasons:
```
Rick-Anderson: ~However, it's a bad practice to hardcode the dependency in this way for the following reasons:~

We should avoid bad. Hardcode is jargon and should be avoided. Maybe something like

Code dependencies (such as the previous example) are problematic and should be avoided for the following reasons:

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -146,54 +209,180 @@ Transient lifetime services are created each time they're requested. This lifeti
...
+> It's dangerous to resolve a scoped service from a singleton. It may cause the service to have incorrect state when processing subsequent requests.
+
+### Constructor injection behavior
+
+Services can be resolved by two mechanisms:
+
+* `IServiceProvider`
+* [ActivatorUtilities](/dotnet/api/microsoft.extensions.dependencyinjection.activatorutilities)
+
+When services are resolved by `IServiceProvider` or `ActivatorUtilities`, constructor injection requires a *public* constructor. Otherwise, the app throws an [InvalidOperationException](/dotnet/api/system.invalidoperationexception):
```
Rick-Anderson: ~Otherwise, the app throws an [InvalidOperationException](/dotnet/api/system.invalidoperationexception):~


Delete all that stuff, TMI. It bloats the article. Fantastic error message, no need to add all those details.
+
+> A suitable constructor for type '&lt;TYPE&gt;' couldn't be located. Ensure the type is concrete and services are registered for all parameters of a public constructor.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -146,54 +209,180 @@ Transient lifetime services are created each time they're requested. This lifeti
...
+
+> A suitable constructor for type '&lt;TYPE&gt;' couldn't be located. Ensure the type is concrete and services are registered for all parameters of a public constructor.
+
+When services are resolved by `ActivatorUtilities`, constructor injection requires that only one applicable constructor exist. Constructor overloads are supported, but only one overload can exist whose arguments can all be fulfilled by dependency injection. If more than one exists, the app throws an `InvalidOperationException`:
+
+> Multiple constructors accepting all given argument types have been found in type '&lt;TYPE&gt;'. There should only be one applicable constructor.
+
+Constructors can accept arguments that aren't provided by dependency injection, but these must support default values.
+
+The following code throws an `InvalidOperationException` because the runtime is unable to resolve the `item` string:
```
Rick-Anderson: I think you should remove all that code, just give them the rule. The valid  and invalid code is nonsense. There's no reason to do that. The rule is all you need.

Unless you have reasonable code, don't show it.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -146,54 +209,180 @@ Transient lifetime services are created each time they're requested. This lifeti
...
+}
+```
+
+## Entity Framework contexts
+
+Entity Framework contexts should be added to the service container using the scoped lifetime. This is handled automatically with a call to the [AddDbContext](/dotnet/api/microsoft.extensions.dependencyinjection.entityframeworkservicecollectionextensions.adddbcontext) method when registering the database context. Services that use the database context should also use the scoped lifetime.
+
+## Lifetime and registration options
+
+To demonstrate the difference between the lifetime and registration options, consider the following interfaces that represent tasks as an operation with a unique identifier, `OperationId`. Depending on how the lifetime of a service is configured, the container provides either the same or a different instance of the service to the requesting class.
```
Rick-Anderson: > Depending on how the lifetime of a service is configured, the container provides either the same or a different instance of the service to the requesting class.

Can you say
The  `OperationId` depends on the service lifetime.  X, Y, Z return a new instance of  `OperationId`. M, X returns ...

guardrex: Idk about this ... the statement is accurate and the extensions are both obvious and previously covered in the *Service lifetimes* section just above this section.

I would like to clarify it to ...

> Depending on how the lifetime of an operations service is configured for the following interfaces, the container provides either the same or a different instance of the service when requested by a class:

... which makes it a hair clearer that the following interfaces will be used to create services shortly and that service lifetime registrations will dictate receiving a new or existing instance. The naming of the interfaces is obvious on which one goes with which lifetime registration.

Rick-Anderson:
@scottaddie @guardrex  for some reason I can't find the following markup From the review page. I can see it in the review URL and when I edit the content.

## Scope validation

When the app is running in the Development environment on ASP.NET Core 2.0 or later, the default service provider performs checks to verify that:

--- end of stuff I can't find

Question for @scottaddie 
Should we change this to

:::moniker 2.0 >
When the app is running in the Development environment ~on ASP.NET Core 2.0 or later~, the default service provider performs checks to verify that:

scottaddie:
@Rick-Anderson The **Scope validation** section is there in the preview page. You'll need to click the icon by the line numbers to see it:
image

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -146,54 +209,180 @@ Transient lifetime services are created each time they're requested. This lifeti
...
+::: moniker-end
+
+::: moniker range="<= aspnetcore-2.0"
+
+[!code-csharp[](dependency-injection/samples/1.x/DependencyInjectionSample/Models/Operation.cs?name=snippet1)]
+
+::: moniker-end
 
-Services can be registered with the container in several ways. We have already seen how to register a service implementation with a given type by specifying the concrete type to use. In addition, a factory can be specified, which will then be used to create the instance on demand. The third approach is to directly specify the instance of the type to use, in which case the container will never attempt to create an instance (nor will it dispose of the instance).
+An `OperationService` is registered that depends on each of the other `Operation` types. `OperationService` makes it clear if transient services are created when requested or merely mirroring the `OperationsId` of the `IOperationTransient` service:
```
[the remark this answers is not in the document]
guardrex: > Where does it make it clear. I don't see that. What does or merely mirroring the OperationsId of the IOperationTransient service: mean?

I'll come back to this one in a bit. It's going to take a few minutes to work it out.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -146,54 +209,180 @@ Transient lifetime services are created each time they're requested. This lifeti
...
+Singleton: 01271bc1-9e31-48e7-8f7c-7261b040ded9  
+Instance: 00000000-0000-0000-0000-000000000000
+
+Observe which of the `OperationId` values vary within a request and between requests:
+
+* *Transient* objects are always different. Note that the transient `OperationId` value for both the first and second requests are different for both `OperationService` operations and across requests. A new instance is provided to each service and request.
+* *Scoped* objects are the same within a request but different across requests.
+* *Singleton* objects are the same for every object and every request regardless of whether an `Operation` instance is provided in `ConfigureServices`.
+
+## Resolve a scoped service within the app scope
```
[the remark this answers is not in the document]
guardrex: > Heading is too abstract. 

Fixed.

> import a code snippet from ... Get a DB context instance from the dependency injection container ...

We already have one tho. This is about more than just dB contexts. I don't want a reader to get too married to the idea that that's all this is for. Let me try a version that links to that section as a further example. If it isn't well received, I can swap it out completely on a future commit.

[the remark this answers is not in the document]
guardrex: Oh ... wait ... I don't think we can do that at all. Seeding is entering a new world order here with 2.1 ... we have https://github.com/aspnet/Docs/issues/7342 to work it over.

aspnetcore/fundamentals/dependency-injection.md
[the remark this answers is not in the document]
guardrex: > Where does it make it clear. I don't see that. What does or merely mirroring the OperationsId of the IOperationTransient service: mean?

I did a reboot on this section. We'll have to see tho ... is it an improvement :smile: :tada: or a white line nightmare :boy: :gun:.

guardrex:
@davidfowl This is an important set of updates for the DI topic. Do you want to 🔪 carve me up 🦃?

guardrex:
@pakrym, `@davidfowl` recommended you TAL at this, too.

This is a fairly heavy rewrite to simplify the *Programming 101* explanation of DI. If it's ultimately determined that this topic should launch right into DI with something like EF/RazorPagesMovie as the example, then I propose we move the intro that I have here (based on the excellent approach for new devs taken by @MikeWasson) into an index topic under a new DI node, then have this DI topic (as a more advanced reference topic) live under the DI node.

The rest of this is mostly sample updates and guidelines adherence.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -3,139 +3,202 @@ title: Dependency injection in ASP.NET Core
...
+> Each `services.Add<ServiceName>` extension method adds (and potentially configures) services. For example, `services.AddMvc()` adds the services Razor Pages and MVC require. We recommended that apps follow this convention. Place extension methods in the [Microsoft.Extensions.DependencyInjection](/dotnet/api/microsoft.extensions.dependencyinjection) namespace to encapsulate groups of service registrations.
 
-In this case, both `ICharacterRepository` and in turn `ApplicationDbContext` must be registered with the services container in `ConfigureServices` in `Startup`. `ApplicationDbContext` is configured with the call to the extension method `AddDbContext<T>`. The following code shows the registration of the `CharacterRepository` type.
+If a service's constructor requires a primitive, such as a `string`, the primitive can be injected by using [configuration](xref:fundamentals/configuration/index) and the [options pattern](xref:fundamentals/configuration/options).
```
pakrym: I would rephrase it somehow or add a sample. It's not entirely clear what should I do with the string parameter.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -146,54 +209,160 @@ Transient lifetime services are created each time they're requested. This lifeti
...
+Services can be resolved by two mechanisms:
+
+* `IServiceProvider`
+* [ActivatorUtilities](/dotnet/api/microsoft.extensions.dependencyinjection.activatorutilities)
+
+Constructors can accept arguments that aren't provided by dependency injection, but the arguments must assign default values.
+
+When services are resolved by `IServiceProvider` or `ActivatorUtilities`, constructor injection requires a *public* constructor.
+
+When services are resolved by `ActivatorUtilities`, constructor injection requires that only one applicable constructor exist. Constructor overloads are supported, but only one overload can exist whose arguments can all be fulfilled by dependency injection.
```
pakrym: Needs more info where and why ActivatorUtilities is used

aspnetcore/fundamentals/dependency-injection.md
guardrex: `ConfigureContainer` looks like it was 1.1, but returning `IServiceProvider` from `ConfigureServices` was 1.0, so I think we're ok as far as this passage is concerned. https://github.com/aspnet/Hosting/issues/91

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -146,54 +209,160 @@ Transient lifetime services are created each time they're requested. This lifeti
...
+
+::: moniker-end
+
+::: moniker range="<= aspnetcore-2.0"
+
+[!code-csharp[](dependency-injection/samples/1.x/DependencyInjectionSample/Models/Operation.cs?name=snippet1)]
+
+::: moniker-end
+
+An `OperationService` is registered that depends on each of the other `Operation` types. When `OperationService` is requested via dependency injection, it receives either a new instance of each service or an existing instance based on how the dependent service is registered.
```
pakrym: based on the lifetime of dependent service

aspnetcore/fundamentals/dependency-injection.md
guardrex: @pakrym I see the API remarks and a few code comments here and there (e.g., Hosting repo for `StartupLoader.LoadMethods`), which don't seem quite adequate to just drop in here. I'd prefer if you or someone from engineering write a paragraph for this or even just supply a handful of sentences that I can wordsmith into a paragraph.

There's probably a need for an advanced DI topic that goes beyond the basics laid out in this topic. I'm aware of `@stevejgordon`'s [Expand DI documentation with additional extension method overloads (aspnet/Docs \#6312)](https://github.com/aspnet/Docs/issues/6312), but it feels to me like we need something between this topic and those. I propose we open a new issue for it tho ... we're swamped at the moment.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -217,144 +386,153 @@ public static void Main(string[] args)
...
 
-* DI is for objects that have complex dependencies. Controllers, services, adapters, and repositories are all examples of objects that might be added to DI.
-
-* Avoid storing data and configuration directly in DI. For example, a user's shopping cart shouldn't typically be added to the services container. Configuration should use the [options pattern](xref:fundamentals/configuration/options). Similarly, avoid "data holder" objects that only exist to allow access to some other object. It's better to request the actual item needed via DI, if possible.
+* Avoid storing data and configuration directly in the service container. For example, a user's shopping cart shouldn't typically be added to the service container. Configuration should use the [options pattern](xref:fundamentals/configuration/options). Similarly, avoid "data holder" objects that only exist to allow access to some other object. It's better to request the actual item via dependency injection, if possible.
 
 * Avoid static access to services.
 
-* Avoid service location in your application code.
+* Avoid service location in app code.
```
pakrym: using service locator pattern

guardrex:
@pakrym See if the examples I added are sane ...

> * Avoid static access to services (for example, \[IApplicationBuilder.ApplicationServices](/dotnet/api/microsoft.aspnetcore.builder.iapplicationbuilder.applicationservices)).
>
> * Avoid using the service locator pattern (for example, \[IServiceProvider.GetService](/dotnet/api/system.iserviceprovider.getservice)).
>
> * Avoid static access to `HttpContext`.

Is an example of the last one `RequestServices`?

pakrym:
They are not entirely correct, a bit better example of "static access to HttpContext" is using `IHttpContextAccessor`. While being a service it's implemented as AsyncLocal and has a lot of similarities with static HttpContext.

Using `IApplicationBuilder.ApplicationServices` on the other hand is fine when appropriate.

guardrex:
@pakrym How about making it more specific for `ApplicationServices`?

> * Avoid static access to services (for example, statically-typing \[IApplicationBuilder.ApplicationServices](/dotnet/api/microsoft.aspnetcore.builder.iapplicationbuilder.applicationservices) for use elsewhere).

... or do you have an example ... something devs commonly do that's *bad*&mdash;*bad*&mdash;*not good*.

guardrex:
@pakrym I modified your description of `ActivatorUtilities` for the topic on the last commit.

guardrex:
Thanks @pakrym!

@davidfowl ... do you still want to take a look?

guardrex:
@Rick-Anderson @scottaddie Anything else on this one?

PR #7416, opened 2018-07-02
Add missing links to Bootstrap 4 CDN resources

peterblazejewicz:
This add missing links to resources already hosted on CDN for Bootstrap 4, but missing in docs.
The resources linked can be used standalone by developers without using main `bootstrap.css` resource:
- to just reset the default browsers styling and built upon this own solution
- to just use the grid layout from Bootstrap without a need to use the fully fledged components.

Reference:
http://getbootstrap.com/docs/4.1/getting-started/contents/

Thanks!


scottaddie:
Thank you, @peterblazejewicz

PR #7423, opened 2018-07-02
[SignalR] UserIdentifier

BrennanConroy:
I realized we don't mention how to see the user's identifier anywhere. Adding a small blurb for it.


scottaddie:
Thank you, @BrennanConroy

PR #7426, opened 2018-07-03
move away from Directory.GetCurrentDirectory()

Gwunhar:
As I discovered when I took this code to use in a console app yesterday...Directory.GetCurrentDirectory() failed to provide the path for the appsettings.json when the console app was launched in a method other than double clicking (for instance, pasting the full path in the Explorer path bar, or launching via CMD when the directory was set somewhere other than the EXE's directory).

Would it be better to use AppDomain.CurrentDomain.BaseDirectory which is more likely to accurately find the EXE's actual directory, and thus the appsettings.json, on any given launch method?

  Fixes #Issue_Number

### DIFF
--- a/aspnetcore/fundamentals/configuration/index/sample/ConfigJson/Program.cs
+++ b/aspnetcore/fundamentals/configuration/index/sample/ConfigJson/Program.cs
@@ -11,7 +11,7 @@ public class Program
     public static void Main(string[] args = null)
     {
         var builder = new ConfigurationBuilder()
-            .SetBasePath(Directory.GetCurrentDirectory())
+            .SetBasePath(AppDomain.CurrentDomain.BaseDirectory)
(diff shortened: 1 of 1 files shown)


Tratcher:
This is premature, even the product code uses GetCurrentDirectory:
https://github.com/aspnet/MetaPackages/blob/622bdeec84fe502e441f296a9af10bc896848167/src/Microsoft.AspNetCore/WebHost.cs#L156

I recall the most common issue was when you didn't have an exe, you were using dotnet.exe myapp.dll.

guardrex:
Thanks for contributing @Gwunhar. We'll keep an eye on engineering to see if they change this in the future.

Gwunhar:
Ah, ok, that makes sense. Thanks for the time!

PR #7436, opened 2018-07-03
Update localization.md

### LINKED ISSUE #7434: Document usage of RootNamespaceAttribute
## Requests for new Topics

We should have a short section or paragraph describing what the RootNamespaceAttribute is, when you would use it, and how to include it in your project.

### Proposed Contents
When the root namespace of an assembly is different than the assembly name localization does not work by default due to the way resources are searched for within the assembly, and the fact that `RootNamespace` is a build-time value which is not available to the executing process. If your `RootNamespace` is different from your `AssemblyName` then you should include the following in your `AssemblyInfo.cs` (with values replaced with your actual values):
```
using System.Reflection;
using Microsoft.Extensions.Localization;

[assembly: ResourceLocation("ResourceFolderName")]
[assembly: RootNamespace("YourRootNamespace")]
```

This should allow for successful resolution of resx files.

Rick-Anderson:
Fixes #7434

### DIFF
--- a/aspnetcore/fundamentals/localization.md
+++ b/aspnetcore/fundamentals/localization.md
@@ -154,6 +154,27 @@ Resource files using `@inject IViewLocalizer` in Razor views follow a similar pa
 
 If you don't use the `ResourcesPath` option, the *.resx* file for a view would be located in the same folder as the view.
 
+### RootNamespaceAttribute 
+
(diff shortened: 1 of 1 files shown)


hishamco:
Thanks @Rick-Anderson

PR #7444, opened 2018-07-03
Updated database name template to 2.x

alfredmyers:


Rick-Anderson:
@alfredmyers  thanks for the fix. I don't think it ever was a date string.

alfredmyers:
@Rick-Anderson The 'Date string' component is part of the database name for the ASP.NET MVC 5 project template (is that how it's called?) - not ASP.NET Core.

Issue #7377, opened 2018-06-29
Can't use "Options pattern" with generic host


andriialbumentormate:
Can't understand how to grab data using  "Options pattern" from appsettings.json file.
I have next code (part of Main method):

```
var environment = EnvironmentName.Development;

            var config = new ConfigurationBuilder()
                .AddJsonFile("appsettings.json", optional: true)
                .AddJsonFile(
                        $"appsettings.{environment}.json",
                        optional: true)
                .AddCommandLine(args)
                .AddEnvironmentVariables()
                .Build();


            var host = new HostBuilder()
                .ConfigureAppConfiguration(configApp => {
                    configApp.AddConfiguration(config);
                })
                .ConfigureLogging((hostContext, configLogging) =>
                {
                    configLogging.AddConsole();
                    configLogging.AddDebug();
                })
                .ConfigureServices((hostContext, services) => {
                    services.AddLogging();
                    services.AddHostedService<SomeService>();
                    services.AddSingleton<ISomeHelperService, SomeHelperService>();
                    services.Configure<MyOptions>(config);
                })
                .UseEnvironment(environment)
                .Build();
```

My app settings file contains
```
{
  "Logging": {
    "LogLevel": {
      "Default": "Warning"
    }
  },
  "AllowedHosts": "*",
  "option1": "value1_from_json",
  "option2": -1
}
```

Options class
```
public class MyOptions
    {
        public string Option1 { get; set; }
        public int Option2 { get; set; }
    }
```
SomeService looks like

```
internal class SomeService : IHostedService
    {
        private readonly ILogger _logger;
        private readonly MyOptions _myOptions;
        private readonly ISomeHelperService _someHelperService;

        public SynchronizationService(ISomeHelperService someHelperService, ILogger<SomeService> logger, IOptions<MyOptions> optionsAccessor)
        {
            _someHelperService = someHelperService;
            _logger = logger;
            _myOptions = optionsAccessor.Value;
        }

        public Task StartAsync(CancellationToken cancellationToken)
        {
            return Task.CompletedTask;
        }

        public Task StopAsync(CancellationToken cancellationToken)
        {
            return Task.CompletedTask;
        }
    }
```
When I run app _myOptions' fields have default values, but have to hold values from settings file.
What can be the reason of this issue and how to solve it ?

Thanks a lot for any answer !

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

[...]

PR #7472, opened 2018-07-05
Move settings files update

guardrex:
Addresses #7377 

I guessed it best to align this with the way that Sdk.Web does it with the difference being that we'll go with `CopyToOutputDirectory` instead of `CopyToPublishDirectory`.


aspnetcore/fundamentals/host/generic-host.md
```diff
@@ -122,6 +122,14 @@ Example app configuration using `ConfigureAppConfiguration`:
...
 
+To move settings files to the output directory, specify the settings files as [MSBuild project items](/visualstudio/msbuild/common-msbuild-project-items) in the project file. The sample app moves its JSON app settings files with the following **&lt;Content:&gt;** item:
+
+```xml
+<ItemGroup>
+  <Content Include="**\*.json" CopyToOutputDirectory="PreserveNewest" />
```
Tratcher: How many settings files are there? This might be more discoverable if you enumerated them.

guardrex: In the case of the configuration sample, just one. For this sample, there are all three.

PR #7505, opened 2018-07-06
Correct broken link to source code sample

marklnz:
As per the "contributing" instructions, because this is a simple change, there is no issue raised for this change. 

There is a broken link in this file (to a source code repo in GitHub). I've corrected the link, pointing it to what is (I hope) the correct location.


scottaddie:
@marklnz Thank you for fixing this!

PR #7512, opened 2018-07-07
Windows Service topic updates

### LINKED ISSUE #5578: UE pass on Host in a Windows Service doc
https://docs.microsoft.com/aspnet/core/host-and-deploy/windows-service

- [x] UE edit
- [x] ~Update template name referenced in https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/windows-service?tabs=aspnetcore2x#prerequisites~ No longer relevant.
- [x] Update sample for 2.1 (`CreateWebHostBuilder` approach).
- [x] Remove `UseApplicationInsights` calls.
- [x] ~Display necessary `using` statements in code snippets.~ They're in the sample app.
- [x] ~Address customer comments~ Except for the "namespace" remark, they're not actionable, and the remark on namespaces is covered by the sample app.

guardrex:
Fixes #5578 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/host-and-deploy/windows-service?view=aspnetcore-1.0&branch=pr-en-us-7512)

These changes convert the sample app over to a 2.1-flavored `CreateWebHostBuilder` approach.

@javiercn For testing infrastructure purposes, is it ok if the method signature for `CreateWebHostBuilder` is changed from ...

```csharp
CreateWebHostBuilder(string[])
```

... to ...

```csharp
CreateWebHostBuilder(string[], bool)
```

... I'd like to add a parameter, but I don't want to break the potential for testing by doing so. btw- I do have a workaround if you say that the signature can't change.


javiercn:
Nope. It needs to be CreateWebHostBuilder(string [] args)

guardrex:
Thanks @javiercn ... I'll go with my workaround that preserves the signature then.

guardrex:
@javiercn On the last commit, I preserve the sig of `CreateWebHostBuilder` by using a static field ...

```csharp
private static bool _isService = true;

public static void Main(string[] args)
{
    if (Debugger.IsAttached || args.Contains("--console"))
    {
        _isService = false;
    }

    var host = CreateWebHostBuilder(args).Build();

    if (_isService)
    {
        host.RunAsService();
    }
    else
    {
        host.Run();
    }
}

public static IWebHostBuilder CreateWebHostBuilder(string[] args)
{
    var pathToContentRoot = Directory.GetCurrentDirectory();

    if (_isService)
    {
        var pathToExe = Process.GetCurrentProcess().MainModule.FileName;
        pathToContentRoot = Path.GetDirectoryName(pathToExe);
    }

    var webHostArgs = args.Where(arg => arg != "--console").ToArray();

    WebHost.CreateDefaultBuilder(webHostArgs)
        .UseContentRoot(pathToContentRoot)
        .UseStartup<Startup>();
}
```

Tratcher:
Does UseContentRoot need to be in CreateWebHostBuilder? What if you moved that to Main?

guardrex:
Since `UseSolutionRelativeContentRoot` is a thing, perhaps so. Either @javiercn will have to say :+1:, or I'm going to have to do some testing that I don't really have time to do. 🏃😅 Even if we do it tho, we're breaking an implementation pattern of calling it in `CreateWebHostBuilder`.

guardrex:
> we're breaking an implementation pattern

Given it's a WS tho ... maybe that's better. This is a one-off style of Core app.

javiercn:
```csharp
public static void Main(string[] args)
{
    var isService = !(Debugger.IsAttached || args.Contains("--console"));
    var builder = CreateWebHostBuilder(args.Where(arg => arg != "--console").ToArray());

    if (isService)
    {
        var pathToExe = Process.GetCurrentProcess().MainModule.FileName;
        var pathToContentRoot = Path.GetDirectoryName(pathToExe);
        builder.UseContentRoot(pathToContentRoot);
    }

    var host = builder.Build();

    if (isService)
    {
        host.RunAsService();
    }
    else
    {
        host.Run();
    }
}

public static IWebHostBuilder CreateWebHostBuilder(string[] args)
{
    return WebHost.CreateDefaultBuilder(webHostArgs)
        .UseStartup<Startup>();
}
```

guardrex:
Nice one. I'll take out the `return` then on `CreateWebHostBuilder`.

PR #7540, opened 2018-07-10
Remove instructions about using the implicit PackageReference feature for AspNetCore.App/All

natemcmaster:
cref aspnet/Home#3292

Per customer feedback, we have decided to restore the Version attribute on the PackageReference for .App/.All in our new project templates. Users are still struggling with the implicit package version if they have not updated to the 2.1.301 SDK, so let's change the docs to show 2.1.1 as the recommended minimum version.


Rick-Anderson:
@natemcmaster  changed to "reference does **not** guarantee that version of the shared framework will be used." From " will be chosen". If that's OK, I'll S&M and merge to live so it's live.

natemcmaster:
Thanks for the fixups! :shipit:

PR #7560, opened 2018-07-11
Update TodoController.cs

jdzhang1221:
dear,
I think the code snippets on lines 6 and 14 are not neat and tidy, and the removal will be cleaner.
Will this look better?


scottaddie:
The regions are required to import the appropriate code snippets into the doc. If this controller wasn't used for a doc, I would agree with you on this change. Thank you for the PR, anyway.

PR #7574, opened 2018-07-11
Update RazorPagesMovieContext.cs

Rick-Anderson:
@guardrex  anything else I need to delete to fix RP intro? You can clean up after I merge this.


guardrex:
I'm doing it now. Best to wait .... could create a merge :boom: if we both make changes.

PR #7576, opened 2018-07-11
Move RP tutorial file uploads into its own topic

### LINKED ISSUE #7575: Move RP tutorial file uploads topic into its own topic
See: #7570

guardrex:
Fixes #7575 

[Internal Review Link](https://review.docs.microsoft.com/en-us/aspnet/core/razor-pages/upload-files?view=aspnetcore-2.0&branch=pr-en-us-7576)

Given the speed 🏃 that this was done, it's best if we all look it over.

I actually don't inline here ... I do move copies of the projects over to draw the code snippets from the original samples.

We could make this a proper, dedicated, linked sample at some point in the future. I could chop out all of the RazorPagesMovies bits and focus it solely on the file uploads bits.

I **don't link** to the code I moved over here, so we don't need to sweat devs trying to run these samples over here without plumbing out the dB and doing all of the tutorial migration steps. All I say is that this topic builds on the sample app completed by the tutorial.


guardrex:
@Rick-Anderson We can post-look :eyes: if @scottaddie is away. I just can't see the built topics from here.

I'll put the internal review links in now.

PR #7606, opened 2018-07-13
Fix links pointing to dev branch to point to master branch

Rutix:
Found some more links pointing to non-existing dev branches. Also found a link in docfx.json which was non-existent.


Rick-Anderson:
@Rutix  thanks again. cc/ @guardrex

PR #7640, opened 2018-07-16
Update Hosted Service topic: Refactor Queued Hosted Service to use BackgroundService

### LINKED ISSUE #7638: Convert sample to use BackgroundService
Convert the *BackgroundTasksSample-WebHost* sample over to [BackgroundService](https://docs.microsoft.com/dotnet/api/microsoft.extensions.hosting.backgroundservice).

* Requires commensurate changes to the topic's text.
* Probably need to snapshot 2.0 code, since `BackgroundService` is new in 2.1.

See: https://github.com/aspnet/Docs/pull/7397

granthoff1107:
Fixes #7638
Updating Queued Hosted Service to use Background Service.

Re-purposed text for description from https://docs.microsoft.com/en-us/dotnet/standard/microservices-architecture/multi-container-microservice-net-applications/background-tasks-with-ihostedservice

Modified the BackgroundService Link to use xref (I'm under the assumption that just linking the Class is enough based on the IHosted Service implementation from other docs)


guardrex:
@Tratcher I'll touch-up the text after your engineering review/approval.

aspnetcore/fundamentals/host/hosted-services/samples-snapshot/2.x/BackgroundTasksSample-WebHost/QueuedHostedService.cs
```diff
@@ -0,0 +1,65 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+namespace BackgroundTasksSample.Services
+{
+    #region snippet1
+    public class QueuedHostedService : IHostedService
```
Tratcher: Why are there two copies of QueuedHostedService? For 2.0 and 2.1? The folder structure does not make that obvious.

guardrex: I wish we didn't need to break tradition with this 2.0 and 2.1 coverage. What do you think about dropping the 2.0 coverage entirely? We generally decided not to do it. I was hoping we could craft this without full 2.0 and 2.1 samples, hence my interest in just linking the 2.0 code via snippets. We probably can do it this way if I just identify that the sample app is a 2.1 sample in the topic where the sample is linked.

Tratcher: October... https://blogs.msdn.microsoft.com/dotnet/2018/06/20/net-core-2-0-will-reach-end-of-life-on-september-1-2018/
How you want to juggle it in the meantime I'll leave to you and Rick.

guardrex: I vote for 🔪 it now. We already decided internally to *disfavor* 2.0-specific code after 2.1 came out. My understanding is that this decision was based on the difficulty in maintaining the topics+samples. It's just too burdensome to us to maintain 1.x, 2.0, and 2.1 topics/samples.

Let's ask ... that way I won't get into trouble. I'll be next for the 🔪 if I'm not careful! :smile:

@Rick-Anderson TL;DR Do you want to maintain a full, working 2.0 version of the BackgroundTasksSample-WebHost sample because the proposed changes here are only available for 2.1, or do you want to stick with what we decided earlier and cut 2.0 loose for 2.1 scenarios (i.e., drop the 2.0 approach from the topic+sample here)?

Rick-Anderson:
@guardrex  I've  been providing a link to a PDF for the 2.0 version and dropping it. See https://docs.microsoft.com/en-us/aspnet/core/data/ef-rp/intro?view=aspnetcore-2.0&tabs=visual-studio

guardrex:
@granthoff1107 Sorry ... change of plans. Let's drop the 2.0 bits after all. We don't normally leave 2.0 code in the topics for 2.1, so we'll link the PDF as @Rick-Anderson suggests.

Do this ...

1. Delete the snapshot file at from this branch ...

   aspnetcore/fundamentals/host/hosted-services/samples-snapshot/2.x/BackgroundTasksSample-WebHost/QueuedHostedService.cs

1. Remove the "moniker range" and "moniker-end" *lines* for 2.1. Leave the text in the topic ... we'll show the text and your 2.1 bits to all readers.

1. Totally *remove* the 2.0 moniker range block (including its content) ...

```
::: moniker range="= aspnetcore-2.0"

[!code-csharp[](hosted-services/samples-snapshot/2.x/BackgroundTasksSample-WebHost/Services/QueuedHostedService.cs?name=snippet1&highlight=30-31,35)]

::: moniker-end
```

I'll take care of the rest. I'll make some minor guidelines adherence updates after your next commit.

granthoff1107:
I've updated so we only support 2.1

guardrex:
Thanks @granthoff1107 ... I'll make a handful of guidelines adherence updates this evening, and I'll get the PDF link in the right spot.

guardrex:
I'd like to preserve the logging and use a `cancellationToken` over a `stoppingToken`. I made these updates on the last commit, but are there any objections?

```csharp
protected async override Task ExecuteAsync(
    CancellationToken cancellationToken)
{
    _logger.LogInformation("Queued Hosted Service is starting.");

    while (!cancellationToken.IsCancellationRequested)
    {
        var workItem = await TaskQueue.DequeueAsync(cancellationToken);

        try
        {
            await workItem(cancellationToken);
        }
        catch (Exception ex)
        {
            _logger.LogError(ex, 
                $"Error occurred executing {nameof(workItem)}.");
        }
    }

    _logger.LogInformation("Queued Hosted Service is stopping.");
}
```

guardrex:
Ok ... seems like there are no objections to those updates.

I'll put the final touches on text updates (and get the PDF link into the topic) by the end of the day.

guardrex:
There's no PDF version for 2.0 AFAICT using the same link format as the EF-RP example, so I'm going to go ahead and merge this based on our decision to doc the latest minor release for the current major release. I'll drop a patch in later to conform the rest of the topic to this approach.

guardrex:
Thanks @granthoff1107 for your work on this ... *greatly appreciated!!!* 🥇

PR #7663, opened 2018-07-17
Remove confusing verbiage from .All metapackage title metadata

scottaddie:
Remove the "and later" portion of the title from the metadata, since the .App metapackage should be used instead.

@Rick-Anderson Should this doc only display for 2.0? I was just talking to a customer who was confused about which metapackage to use.


Rick-Anderson:
@scottaddie  no, it's still available.
image

PR #7669, opened 2018-07-17
Changed Bootswatch URL to point to Bootswatch 3.

ameyer117:
It is well known that if you use a Bootstrap 4 template on an ASP.NET template it will break the layout. This is a convenience change to prevent confusion as it will send people to the correct versioned URL as it was back when this article was written.

  Fixes #Issue_Number


scottaddie:
@ameyer117 Thank you for your contribution!

PR #7670, opened 2018-07-17
SignalR - .NET Client Windows Auth

anurse:
https://github.com/aspnet/Docs/issues/7586 revealed a gap in the new authentication docs. Added a small blurb about configuring clients for Windows Auth.


scottaddie:
Thank you, @anurse!

PR #7676, opened 2018-07-18
Corrected broken URLs to point to updated content.

ameyer117:
The "Start Free Trial" and "Subscribe Now" links pointed to a 404 page at Pluralsight. I have updated the URLs with destinations that make sense to the average user. The free trial URL remains focused on Scott Allen's ASP.NET MVC 5 course for promotional purposes.

**If Microsoft has updated affiliated links, that might be a better option for the future.**

### DIFF
--- a/aspnet/mvc/pluralsight.md
+++ b/aspnet/mvc/pluralsight.md
@@ -22,14 +22,13 @@ MVC Video Training from Pluralsight
 - [SignalR](https://pluralsight.com/training/Player?author=scott-allen&name=aspdotnet-mvc5-fundamentals-m7-signalr&mode=live&clip=0&course=aspdotnet-mvc5-fundamentals)
 - [Web Developer Tools and Visual Studio 2013](https://pluralsight.com/training/Player?author=scott-allen&name=aspdotnet-mvc5-fundamentals-m8-visualstudio&mode=live&clip=0&course=aspdotnet-mvc5-fundamentals)
 
-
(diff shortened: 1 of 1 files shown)


ameyer117:
The "free" part was there before I even started. :)  Maybe you would want to edit/remove the word "Free" 
As for the other link - I'm simply just linking them to their default pricing page as you suggested?

guardrex:
> The "free" part was there before I even started. 

Yes, that's why I said "pre-existing" in my remarks.

*Yes!* on the link ... I agree with you that that's the best place to send cats.

The inclusion of pricing info and "free" language is an editorial decision for @Rick-Anderson and @scottaddie to work out. Our guidelines dissuade the practice. I'm just a little worker 🐝 around here, so I'm just making a passing suggestion on it. They'll let you know if they want to make changes to that content.

ameyer117:
All good. Sorry missed the pre-existing part. Not a very good excuse on my part, but I was reading your reply on my phone while waiting for coffee! lol

Rick-Anderson:
> I recommend just linking to 3rd party product/service description/pricing pages and allowing the 3rd party to handle their pitch and pricing descriptions. 

+99 - Lets do that.

guardrex:
@ameyer117 I've moved on from morning ☕️ to an afternoon 🍺. I'm sure the quality of my next two PR's will **go UP!** :smile:

guardrex:
Thanks @ameyer117 ... I'll get this in as soon as the build completes.

PR #7678, opened 2018-07-18
Hosting startup topic updates

### LINKED ISSUE #6896: Split dynamic loading section into separate document

guardrex:
Fixes #6896 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/host/platform-specific-configuration?view=aspnetcore-2.0&branch=pr-en-us-7678)

* Surfaces the new content in the *same topic*. I think this pulls it off and avoids 👹 **_Topic Creep_**:tm: 👹 .
* Sample expanded to demo a hosting startup activated from a class lib. The class lib adds some config values via `AddInMemoryCollection`.
* Includes a description of the new (2.1) exclusion machinery:
  - Hosting Startup Exclude Assemblies host configuration setting.
  - `ASPNETCORE_HOSTINGSTARTUPEXCLUDEASSEMBLIES`
* Moves topic to *host* in the TOC.


aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -1,67 +1,108 @@
...
+
+[!code-csharp[](platform-specific-configuration/samples/2.x/HostingStartupSample/HostingStartupApp/Startup.cs?name=snippet1&highlight=1)]
+
+The app's Index page reads the configuration values for the two keys set by the class library's hosting startup assembly. The configuration values for these keys are displayed when the page is rendered.
+
+*HostingStartupApp/Pages/Index.cshtml.cs*:
+
+[!code-csharp[](platform-specific-configuration/samples/2.x/HostingStartupSample/HostingStartupApp/Pages/Index.cshtml.cs?name=snippet1&highlight=11-12,19-20)]
 
 ## Discover loaded hosting startup assemblies
```
Tratcher: Related: The entry assembly or the assembly containing your Startup class is scanned by default for HostingStartup attributes. https://github.com/aspnet/Hosting/blob/864caa20b43ab33415795bd4f6c6d3a4365c8bfe/src/Microsoft.AspNetCore.Hosting/Internal/WebHostOptions.cs#L36-L37

aspnetcore/fundamentals/host/platform-specific-configuration/samples/2.x/HostingStartupSample/HostingStartupApp/Pages/Index.cshtml.cs
```diff
@@ -16,12 +16,16 @@ public IndexModel(IConfiguration config)
         }
 
         public string[] LoadedHostingStartupAssemblies { get; private set; }
+        public string ServiceKey_Development { get; private set; }
+        public string ServiceKey_Production { get; private set; }
 
         public void OnGet()
         {
             LoadedHostingStartupAssemblies =
```
Tratcher: Why isn't all of this done in the constructor?

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -1,67 +1,108 @@
...
 
 To discover hosting startup assemblies loaded by the app or by libraries, enable logging and check the application logs. Errors that occur when loading assemblies are logged. Loaded hosting startup assemblies are logged at the Debug level, and all errors are logged.
 
 The sample app reads the [HostingStartupAssembliesKey](/dotnet/api/microsoft.aspnetcore.hosting.webhostdefaults.hostingstartupassemblieskey) into a `string` array and displays the result in the app's Index page:
```
Tratcher: I'm not sure why we have this sample, this is not something we expect apps to do in code. Also, after aspnet/Hosting#1243 this list isn't very accurate. Also see my prior comment about your entry assembly being added to the list by default. Describe the config keys but remove the LoadedHostingStartupAssemblies code sample.

aspnetcore/fundamentals/host/platform-specific-configuration/samples/2.x/HostingStartupSample/HostingStartupApp/Pages/Index.cshtml.cs
guardrex: Fine by me. I'll change it on the next commit.

aspnetcore/fundamentals/host/platform-specific-configuration.md
guardrex: It was just for demo purposes. I'll remove it on the next commit.

guardrex: What is "this list?"

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -1,67 +1,106 @@
...
+
+The [sample app](https://github.com/aspnet/Docs/tree/master/aspnetcore/fundamentals/host/platform-specific-configuration/samples/2.x/HostingStartupSample) includes a Razor Pages app, *HostingStartupApp*, and a class library, *HostingStartupLib*. The class library contains a hosting startup class, `ServiceKeyInjection`, which implements `IHostingStartup`.
+
+`ServiceKeyInjection` adds a pair of service strings to the app's configuration using the in-memory configuration provider ([AddInMemoryCollection](/dotnet/api/microsoft.extensions.configuration.memoryconfigurationbuilderextensions.addinmemorycollection)).
+
+*HostingStartupSample/HostingStartupLib/ServiceKeyInjection.cs*:
+
+[!code-csharp[](platform-specific-configuration/samples/2.x/HostingStartupSample/HostingStartupLib/ServiceKeyInjection.cs?name=snippet1)]
+
+The app's `Startup` class file specifies a [HostingStartup](/dotnet/api/microsoft.aspnetcore.hosting.hostingstartupattribute) attribute for the class library's `ServiceKeyInjection` class.
```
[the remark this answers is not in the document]
guardrex: > fails to describe the HostingStartup assembly attribute that can point to types in existing referenced libraries

I knew "referenced library" could've meant that the lib isn't in the same solution. Perhaps, I misunderstood @davidfowl's "very simple attribute" remark and his link to the repo sample to mean that the app *would provide the direct reference from the Startup class* ... that's what the linked repo sample shows.

> unreferenced assembly and discovered via environment variable.

.... so .....

1. Have the lib as a standalone project
1. Compile the standalone lib
1. bin-deploy the lib in the app
1. Set the env var (`ASPNETCORE_HOSTINGSTARTUPASSEMBLIES`)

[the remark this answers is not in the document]
guardrex: Ok ... sounds good. I'll work on this tomorrow, and let's look again next week.

guardrex:
@Tratcher I updated the topic+sample along the lines of your feedback. It has a standalone class lib scenario and a NuGet package scenario now. I go with the NuGet scenario first, as I think that might be more popular.

The paint is still wet on these updates, so I'll make anther pass Monday morning.

Two items came up along the way:

1. I distinguish *Activation from a class library* from *Activation from an assembly* (the "additionalDeps" scenario). I don't think it's clear/correct in its present form. I guess a console app without an entry point is an "assembly." What do we call the "assembly" (console app without an entry point) that's activating the hosting startup in the "additonalDeps" scenario? I looked up some info and found that in the olden days there were *library assemblies* and *process assemblies*. Do we have a third type of "assembly" these days?
1. I removed "dynamic" in the description of the "additionalDeps" scenario because we need to further explain what "dynamic" means. Dynamic in what sense? How is "dynamic" contrasted with the NuGet package and class lib scenarios ... are they "static" in some sense?

guardrex:
@Tratcher I'm sure it needs more work, but take a :eyes: now. There were a lot of updates this round.

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -0,0 +1,289 @@
...
+
+The compiled class library is bin-deployed to the app, and the class library's assembly name is listed in the `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES` environment variable. The environment variable is a semicolon-delimited list of assemblies that contain a `HostingStartup` attribute that identifies a hosting startup enhancement to load.
+
+The class library's assembly name can also be provided using the [Hosting Startup Assemblies](xref:fundamentals/host/web-host#hosting-startup-assemblies) host configuration setting.
+
+When multiple hosting startup assembles are present, their [Configure](/dotnet/api/microsoft.aspnetcore.hosting.ihostingstartup.configure) methods are executed in the order that the assemblies are listed.
+
+## Assembly activation
+
+This section describes how to activate a hosting startup from a bin-deployed or runtime store-deployed assembly.
```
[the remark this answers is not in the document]
Tratcher: This will only work for .NET Core, not .NET Framework. The direct dependency approach works on both.

aspnetcore/fundamentals/host/index.md
```diff
@@ -15,3 +15,9 @@ uid: fundamentals/host/index
 * [Generic Host](xref:fundamentals/host/generic-host) (ASP.NET Core 2.1 or later) &ndash; Suitable for hosting non-web apps (for example, apps that run background tasks). In a future release, the Generic Host will be suitable for hosting any kind of app, including web apps. The Generic Host will eventually replace the Web Host.
 
 For hosting ASP.NET Core *web apps*, developers should use the Web Host based on [IWebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.iwebhostbuilder). For hosting *non-web apps*, developers should use the Generic Host based on [HostBuilder](/dotnet/api/microsoft.extensions.hosting.hostbuilder).
+
+<xref:fundamentals/host/hosted-services>  
+Learn how to implement background tasks with hosted services in ASP.NET Core.
+
+<xref:fundamentals/configuration/platform-specific-configuration>
```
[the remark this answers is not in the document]
guardrex: `@Rick-Anderson` issued a ruling that UIDs are not allowed to change when topics are moved.

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -0,0 +1,289 @@
...
+
+An [IHostingStartup](/dotnet/api/microsoft.aspnetcore.hosting.ihostingstartup) (hosting startup) implementation adds enhancements to an app at startup from an external assembly from outside of the app's `Startup` class. For example, an external tooling library can use a hosting startup implementation to provide additional configuration providers or services to an app. `IHostingStartup` *is available in ASP.NET Core 2.0 or later.*
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/fundamentals/host/platform-specific-configuration/samples/) ([how to download](xref:tutorials/index#how-to-download-a-sample))
+
+## Discover loaded hosting startup assemblies
+
+To discover hosting startup assemblies loaded by the app or by libraries, enable logging and check the application logs. Errors that occur when loading assemblies are logged. Loaded hosting startup assemblies are logged at the Debug level, and all errors are logged.
+
+## Disable automatic loading of hosting startup assemblies
```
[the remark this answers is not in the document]
guardrex: These sections originally ended up here due to your feedback at https://github.com/aspnet/Docs/pull/4967#pullrequestreview-83003242 ...

> Let's move these new sections to the very top right after the opening paragraph, they are applicable to the most users. Comparatively few users implement this feature.

I'm not sure it matters. All it's really saying is list them in this env var and they go away. `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES` isn't directly relevant to the goal of this section, is it?

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -0,0 +1,289 @@
...
+
+There are three approaches that you can take to set up and activate a hosting startup assembly:
+
+* [NuGet package](#nuget-package-activation)
+* [Class library](#class-library-activation)
+* [Assembly](#assembly-activation)
+
+Each approach is described in the following sections.
+
+## NuGet package activation
```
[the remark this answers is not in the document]
guardrex: > are not treated specially for this feature

I understand that from the engineering perspective (i.e., delivery is a separate issue), but we're dealing with a slightly complicated scenario overall with some subtle differences between the gross implementation approaches (*gross* = including delivery).

It's easier to organize the topic content for clarity on the differences by going with separate (but similar) sections due the differences in procedure. This leaves the reader simply to ask, "How do I want to supply this?" Then, they read the applicable section. Otherwise if they aren't sure, they can read them in a more SxS fashion and easily compare and contrast the two.

If we roll NuGet into the class lib section (and remove the parts of the sample that pertain to the NuGet approach ... probably water down the text to basic 'use NuGet to deliver it' statements), I think the gross implementation will become *harder to understand.*

It's your call ... I'll cut it if you want, but I don't recommend it. If you want to proceed with cutting/moving, do you want it out of the sample, too?

Just remember as u make your final call, I didn't write this for you ... I wrote it for *ME*. :smile: lol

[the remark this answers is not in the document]
Tratcher: If you keep it then there are some parts that need to be fixed up. E.g. you say to set the environment variable to packageid, but the env var name always needs to use the assembly name which may not match the packageid.

aspnetcore/fundamentals/host/platform-specific-configuration.md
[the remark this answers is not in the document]
Tratcher: I just though it was weird you said how to turn the feature off before you said how to turn it on.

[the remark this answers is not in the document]
guardrex: When I first added it, it was way down at the bottom just above the last section on the sample app.

https://github.com/aspnet/Docs/pull/4967/commits/76db1ec5b6a10b35c49eacd65796773cdb92e3e5

If you want me to move the disabling part back down there, let me know. We can tack-on a sentence to the discovery section that says ...

> To disable hosting startup enhancements, see \<bookmark>.

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -0,0 +1,289 @@
...
+
+When multiple hosting startup assembles are present, their [Configure](/dotnet/api/microsoft.aspnetcore.hosting.ihostingstartup.configure) methods are executed in the order that the assemblies are listed.
+
+## Assembly activation
+
+This section describes how to activate a hosting startup from a bin-deployed or runtime store-deployed assembly.
+
+### Create the assembly
+
+An `IHostingStartup` enhancement is deployed as an assembly based on a console app without an entry point. The assembly references the [Microsoft.AspNetCore.Hosting.Abstractions](https://www.nuget.org/packages/Microsoft.AspNetCore.Hosting.Abstractions/) package:
```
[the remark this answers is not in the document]
guardrex: I think the wrapper project and getting the right deps file came about from `@pakrym`'s comment here :point_right: https://github.com/aspnet/Docs/issues/4438#issuecomment-349041919 (referring to answering me on 4 and 7).

What's the deal with the hosting startup on ...

* Compile-time reference present: No deps file needed (no additional deps specification)
* No compile-time reference: Deps file needed (additional deps specification required)

If you can help me understand that, I think a few words in the topic on it would help readers understand the requirement in the no-compile-time reference scenario.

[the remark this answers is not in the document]
Tratcher: Core will only load assemblies listed in a deps file. When you have a compile time reference then the dependent assembly and all of its own dependencies end up in the app deps file.  When you don't have a compile time reference then you need to provide a supplementary deps file with all of that assemblies dependencies.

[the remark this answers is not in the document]
guardrex: Under the class lib approach if the class lib's assembly is placed into the runtime store, is an explicit compile-time reference required in the consuming app's project file?

I have it bin-deployed in the text and sample app with a reference in the sample app's project file ..

```xml
<ItemGroup>
  <Reference Include=".\bin\Debug\netcoreapp2.1\HostingStartupLibrary.dll">
    <HintPath>.\bin\Debug\netcoreapp2.1\HostingStartupLibrary.dll</HintPath>
    <SpecificVersion>False</SpecificVersion> 
  </Reference>
</ItemGroup>
```

I'm just wondering in the runtime store case what the compile-time reference looks like. We may need to add it.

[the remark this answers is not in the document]
Tratcher: No, it doesn't require a compile time reference, just an additional deps. That's how we ended up using it for AppInsights and AppServices. Go look at C:\Program Files\dotnet\additionalDeps\Microsoft.AspNetCore.AzureAppServices.HostingStartup.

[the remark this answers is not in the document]
guardrex: What's the practical difference between:

* A class lib in the runtime store with additional deps
* A console app without an entry point in the runtime store with addtional deps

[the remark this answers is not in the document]
guardrex: > No, it doesn't require a compile time reference, just an additional deps. 

This has taken on new complexity: The explanation for additional deps under the "assembly activation" approach (console app w/o entry point) now has to be duplicated under class lib with runtime store deployment.

I'm going to have to mull this over a bit. I'll sleep on it and see what I can come up with in the morning.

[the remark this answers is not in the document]
Tratcher: There's a lot of overlapping parts between the scenarios. What if we start by explaining the smaller components and then pull them together in the two main scenarios at the end? And those scenarios build on eachother. E.g. the no-compile-time-reference scenario is an expansion on top of the compile time reference scenario that uses all the same mechanics and adds the additional deps.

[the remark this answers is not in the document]
guardrex: Yes, that sounds good (devil might be in the details tho). I'll take a good night sleep + up early + plenty of ☕️ and see how it goes in the morning.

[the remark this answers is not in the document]
guardrex: I need to know this tho -------

What's the practical difference between:

* A class lib in the runtime store with additional deps
* A console app without an entry point in the runtime store with addtional deps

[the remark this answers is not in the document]
Tratcher: @pakrym ?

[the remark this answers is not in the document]
guardrex: Apparently if the **console app w/o entry point assembly** is bin-deployed with a compile-time ref, there's no need to deal with the deps file (just like in the console lib assembly case).

### Assembly type (delivery options)

* Class lib in a local assembly (runtime store, bin, or NuGet)
* Console app w/o entry point (runtime store, bin, or NuGet - can it be packed?)

### Delivery options

* runtime store without compile-time ref and provide deps
* bin with compile-time ref
* NuGet package with compile-time ref

### Outline

🥁 roll plz ...

[This outline is in the order of presentation.]

* Class lib (only how to create the assembly here)
* Console app w/o entry point (only how to create the assembly here)
* Activation
  - List in `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES`
  - Runtime store
    * Where to place assembly
    * Provide deps instructions
  - NuGet - I still prefer to keep NuGet separate from class lib and bin-deployment. With this outline, there should be minimal duplication of text among the sections.
    * Link to NuGet packing instructions
    * No deps work required
    * Deployment
      - NuGet package via nuget.org (https://docs.microsoft.com/dotnet/core/deploying/creating-nuget-packages and https://docs.microsoft.com/nuget/create-packages/publish-a-package)
      - NuGet package in the runtime store (https://docs.microsoft.com/dotnet/core/deploying/runtime-store)
  - bin with compile-time ref
    * Where to place assembly
    * No deps work required
* Sample app - Unlike the topic today, completely separate the sample description and steps out of the text. Only include in the sample the most likely deployment approaches.

The pieces are just about all here, so I'm not sorry about the work that's been done thus far. However, let's settle on an outline (maybe that one :point_up:) before I continue.

### Posers, questions, queries

1. What's the practical difference between **class lib** and **console app w/o entry point** approaches? Since it looks like they can both delivered with any delivery approach here, it comes down to the practical difference between them. Which one makes sense for given scenarios?
1. Are **class lib+NuGet package** and **console app w/o entry point+runtime store** the most common anticipated assembly+delivery approaches? If so, the sample only needs to show these two.
1. Between **NuGet package on nuget.org** and **NuGet package in the runtime store**, shall we just link out, or shall we link out with guidance on pros+cons?
1. Can we say something about **installer options**? ... making delivery dead-on simple in multi-machine environments without using NuGet? ... or do we just lean in the direction of **_you should always use NuGet!_**

Tratcher:
Something about the organization is still bugging me. The ordering seems strange and ASPNETCORE_HOSTINGSTARTUPASSEMBLIES is duplicated several times. If I'm trying to author one of these what progression would I need?

How's this for a TOC?
- Intro to IHostingStartup
- HostingStartup attribute
- Activation via ASPNETCORE_PREVENTHOSTINGSTARTUP
- Disable automatic loading
- Discover loaded hosting startup assemblies
- Referencing the assembly
  - Overview: Core vs Framework, deps, gac, etc..
  - Project dependency, class library, bin deployed
  - Nuget dependency, class library, bin deployed
  - Runtime store, class library
    - Deps generation, DOTNET_ADDITIONAL_DEPS

guardrex:
I didn't do the outline work that I show in https://github.com/aspnet/Docs/pull/7678#discussion_r205943890. I'll take another look at what I have there and what you posted and see what the differences are.

Check my *Activation* node in that outline tho ... I have the env var moved up very early. However, I do have creating the class lib or console app w/o entry point before that.

Before I do anything tho, I probably need to hear the answers to these questions ...

1. What's the practical difference between **class lib** and **console app w/o entry point** approaches? Since it looks like they can both delivered with any delivery approach here, it comes down to the practical difference between them. Which one makes sense for given scenarios?
1. Are **class lib+NuGet package** and **console app w/o entry point+runtime store** the most common anticipated assembly+delivery approaches? If so, the sample only needs to show these two.
1. Between **NuGet package on nuget.org** and **NuGet package in the runtime store**, shall we just link out, or shall we link out with guidance on pros+cons?
1. Can we say something about **installer options**? ... making delivery dead-on simple in multi-machine environments without using NuGet? ... or do we just lean in the direction of **_you should always use NuGet!_**

cc: @pakrym

guardrex:
I looked this over in comparison ...

- Intro to IHostingStartup
- HostingStartup attribute
- Activation via ASPNETCORE_PREVENTHOSTINGSTARTUP
- Disable automatic loading
- Discover loaded hosting startup assemblies
- Referencing the assembly
  - Overview: Core vs Framework, deps, gac, etc..
  - Project dependency, class library, bin deployed
  - Nuget dependency, class library, bin deployed
  - Runtime store, class library
    - Deps generation, DOTNET_ADDITIONAL_DEPS

I didn't include a complete outline before, so a couple of those would've been left in-place.

* Your outline doesn't show creating the assembly (especially in relation to "referencing" or "activating" it).
* Idk on the deltas between "Core vs Framework" or additional content on "deps" and "gac" ... Thus far, I'm just trying to doc what's there today for Core. If bits are missing, I need more guidance.
* You have "disable" and "discover" in the middle ... between activation and referencing, but I think those two either need to go first or last.
* Sequence: On face value, it seems backwards ... "activation" :point_right: "referencing" ... it seems like one would *reference something* before they *activate it*, so I don't understand that sequence from your outline. These could also be interpreted as the *same thing*. By "referencing" the assembly, one is also "activating" it at runtime. Anywho ... all of this probably just boils down to your outline doesn't have enough detail for me to make sense of it.
* As for your 'Referencing the assembly' part, that matches what I have (but in a different sequence).

Broadly speaking, my outline does this ...

* Create assembly
* Reference/activate assembly

Breaking down *Create*, it's either ...

* Class lib
* Console app w/o entry point

(I need to hear more on the scenarios where they apply in the real world so I know what to tell readers.)

Breaking down *Activation*, it's ...

* List in the env var
* Placement/deployment/delivery: Specific instructions depending on delivery choice (in order of best/most common to worst/least common):
  - Runtime store
  - NuGet
  - bin deployment

Therefore, I devised it this way (and I'll add those discovery/disable bits this time) ...

* Introduction
* Discover hosting startups
* Disable automatic loading of hosting startups
  - All
  - Specific assemblies
* Class lib (only how to create the assembly here)
* Console app w/o entry point (only how to create the assembly here)
* Activation
  - List in `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES`
  - Runtime store
    * Where to place assembly
    * Provide deps instructions
  - NuGet
    * Link to NuGet packing instructions (https://docs.microsoft.com/dotnet/core/deploying/creating-nuget-packages)
    * No deps work required
    * Deployment
      - NuGet package via nuget.org (https://docs.microsoft.com/nuget/create-packages/publish-a-package)
      - NuGet package in the runtime store (https://docs.microsoft.com/dotnet/core/deploying/runtime-store)
  - bin with compile-time ref
    * Where to place assembly
    * No deps work required
* Sample app - Separate the sample description and steps out of the text above. Only include in the sample the most likely deployment approaches.

... but even if that's a good work starting point, I need the answers to those posers that I listed earlier :point_up: before I proceed.

Tratcher:
I've asked @pakrym for a second opinion on organization and to answer some of your technical questions.

guardrex:
I'm giving that outline :point_up: a shot on the next commit. If it turns out to be a *big steam'in pile of RexPoop* :poop:, it's easy enough to roll a commit back. I should have it done by the end of the day.

guardrex:
@Tratcher It can use another grammar+style pass, but see if the last commit (that follows that new outline :point_up:) is headed in the right direction.

I didn't make updates to the sample yet. Let's hear back from Pavel before I dig into those.

guardrex:
@Tratcher Gave it a pass this morning. It seems fine to me ... I didn't get 😵 outside of the questions I asked earlier. I ended up moving the section on the attribute to the top of the topic, but I left the "discovery" and "disable" sections immediately under it. I left the sample code as it was until I receive more feedback.

pakrym:
>    I distinguish Activation from a class library from Activation from an assembly (the "additionalDeps" scenario). I don't think it's clear/correct in its present form. I guess a console app without an entry point is an "assembly." What do we call the "assembly" (console app without an entry point) that's activating the hosting startup in the "additonalDeps" scenario? I looked up some info and found that in the olden days there were library assemblies and process assemblies. Do we have a third type of "assembly" these days?

I think there is a bit of confusion about why exactly we have a class library and console app. Let me clarify a bit.

What usually happens when we build hosting startups is first we create an implementation library (the one the has IHostingStartup) and target appropriate TFM in it (netstandard/netcoreapp) and treat it as a normal package.

Problem with a class library is that we can't get additional deps file out of it (because deps files are runable application assets) and you can't generate runtime store for it either (`dotnet store` needs a runable project that targets shared runtime to generate store). That's where console application comes into play, we create a dummy app that references our hosting startup package and targets some shared runtime [1] and then publish it to get additional deps file (nice consequence of having an app is that all shared framework stuff is trimmed from deps file automatically). We also call `dotnet store` on the same app project and get runtime store generated.

The last thing we need to do is to remove dummy app entry from `.deps.json` file because we won't be shipping it and runtime would fail to start if it's not found.

[1] - shared runtime you target should be the same as the framework name you put into `additionalDeps` subdirectory.

>    I removed "dynamic" in the description of the "additionalDeps" scenario because we need to further explain what "dynamic" means. Dynamic in what sense? How is "dynamic" contrasted with the NuGet package and class lib scenarios ... are they "static" in some sense?

Dynamic here means that you expand application dependency closure in runtime dynamically vs build time (at build time you do it by referencing packages).

>    Rick has suggested that this topic have a section listing all of the OOB hosting startups. I can find them across the various repos with an aspnet-wide search for "IHostingStartup." However if the list is short and you know it off the top of your head, can you provide the list? If not, no worries ... I'll search them out.

Two largest ones are:

1. https://github.com/aspnet/AzureIntegration/tree/master/src/Microsoft.AspNetCore.AzureAppServices.HostingStartup for AppServices integration

2. ApplicationInsights - it's source is not public

guardrex:
@pakrym Thanks for explaining. Yes, you were correct to clarify the dynamic approach. I can enhance the topic's dynamic content with the information you provided.

You don't address the *static* (compile-time referencing) approach for a class lib/NuGet package.

I'm attempting to follow `@davidfowl`'s guidance in his remark on the issue at https://github.com/aspnet/Docs/issues/6896#issue-329724017 (repeated here) ...

> This document describes how to use IHostingStartup in a completely dynamic way but fails to describe the HostingStartup assembly attribute that can point to types in existing referenced libraries. https://github.com/aspnet/Hosting/blob/8377d226f1e6e1a97dabdb6769a845eeccc829ed/samples/SampleStartups/StartupInjection.cs#L8
>
> We should talk to about the IHostingStartup feature, what it does and then describe the various ways to consume it starting from the very simple attribute to the dynamic additional deps way.

Focusing on ...

> assembly attribute that can point to types in existing referenced libraries

... and ...

> describe the various ways to consume it starting from the very simple attribute to the dynamic additional deps way

Do the approaches on this PR meet those goals?

RE: The practical diff between static and dynamic ...

It looks like the practical diff **_is just that_** ... The dynamic approach doesn't require the compile-time ref. The static approach does. That's it (it seems). If so, that's easy enough to remark upon in the topic.

For the sample updates on this PR (to match the approaches described in the topic), I have:

* A compile-time ref to a class lib.
* A compile-time ref to a NuGet package.
* The dynamic, no-compile-time ref console app approach.

Is that good?

guardrex:
Latest version is creating additional questions/concerns ...

1. If the runtime store delivery is only associated with the dynamic approach (console app w/o entry point), then this outline is splitting this content in an unnatural way. My first outline/original commit didn't have this split, so it might be better to revert. I'll paste a new draft outline below for discussion.
1. @pakrym The explanation you gave (dynamic approach) doesn't mirror the instructions that we provided in the topic when the topic was first written. Either there's something wrong with the instructions/approach in the topic, the cheese has moved, or the framework is performing some steps and the developer is performing other steps and they're commingled. For example ...
   1. There's not obviously an *implementation library* (ordinary package reference) with the dynamic approach shown in the topic ... `IHostingStartup` impl is just a class in the console app.
   1. "We" create a dummy app? Who is "we"? ... the dev or the framework? Isn't the console app here created by the developer? It doesn't make sense to me right now given the steps that I currently have that there's *another console app* involved in the process.
   1. `dotnet store` isn't being used with the sample demo in the topic. The process is manual. The deps file is modified manually (by a PS script in the demo). The compiled assembly and deps files are manually placed (via MSBuild target in the demo). The assembly is placed in the runtime store when the dev (target in this case) moves it, so idk if you're saying the framework calls `dotnet store` later ... at activation ??? ... or has the cheese moved on the entire approach here since the topic was first written? 

If the original topic (and my current instructions are incorrect), I should've pinged you back in December. https://github.com/aspnet/Docs/pull/4967 However, `@davidfowl` didn't say in his issue that there was anything cosmically wrong with how the dynamic approach is explained in the topic today. He seems to want the class lib (bin or NuGet delivery) approach explained via a compile-time reference. https://github.com/aspnet/Docs/issues/6896

Hypothetical outline reversion to recombine runtime package store delivery with the dynamic approach:

* Introduction
* HostingStartup attribute
* Discover hosting startups
* Disable automatic loading of hosting startups
  - All
  - Specific assemblies
* Approaches summary
  - Compile-time reference to assembly
  - Dynamic approach w/o compile-time reference to assembly
* Compile-time reference to assembly
  - Class lib
  - Compile-time reference activation
    - NuGet
    - bin
* Dynamic approach w/o compile-time reference to assembly
  - Console app w/o entry point
  - Runtime store activation
    * Where to place assembly and deps file
    * deps file reference in env var
* List in `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES` (applies to all approaches and activation strategies)
* Sample app

pakrym:
>There's not obviously an implementation library (ordinary package reference) with the dynamic approach shown in the topic ... IHostingStartup impl is just a class in the console app.

Depends on the requirement. If the only requirement is to produce output that would work with additionalDeps/runtime store having IHostingStartup in console app is enough. Sometimes having HostingStartup Nuget package is required and aconsole app would produce an unusable package. So having package and app covers all the scenarios.


   > "Who is "we"?

The dev. Everything I talked about was from the point of view of HostingStartup package author.

>    dotnet store isn't being used with the sample demo in the topic. 

`dotnet store` is required if hosting startup has more dependencies then just the shared framework because they have to be collected too. It would also put hosting startup dll itself into the correct folder structure.

> so idk if you're saying the framework calls dotnet store later ... at activation ??? ... or has the cheese moved on the entire approach here since the topic was first written?

No, dev has to call `dotnet store` somewhere in the build pipeline. `dotnet store` when ran on a `.csproj ` collects it's dependencies and places them into a correct folder structure that later can be dropped into runtime store location.

guardrex:
@pakrym I'm back! ... Last week, I was buried in Configuration topic updates (+ I was battling the *FLU* 🚑).

So ... the process I worked out before didn't rely upon a `dotnet store` step, and the topic needs to go that way for a hosting startup that doesn't get consumed with a compile-time reference.

What I had before (what's there now) is manually moving the assembly to the runtime store. I can see now that that's not good ... and not good for additional deps that aren't in the shared framework. Therefore, I hope that I'm merely converting a step to cover that aspect ... going from manual placement by the dev to having the dev execute the `dotnet store` command.

I'm still a bit concerned about your phrase "dotnet store ... places them into a correct folder structure **that later can be dropped into runtime store location**" ... since I thought that executing the `dotnet store` command should take care of the **dropping** bit, too. I'll find out soon enough ... I'm about to work it thru here and see what happens with the sample app.

I'll work on that now and ping you back when I have the topic+sample updated.

Long-range ... after I get the topic+sample right ... we should still discuss the possibility of coverage for hosting startup **delivery**. How do devs supply these no-compile-time-reference hosting startups ... an installer? If so, shouldn't the topic broach how that's done?

guardrex:
@pakrym Ok ... On the last commit, I swapped in a `dotnet store`-based process for the prior manual-based process. It works. I can't say if there are any "gotchas" that we need to cover.

pakrym:
>  This command places the hosting startup assembly and other dependencies that aren't part of the shared framework in the user profile's runtime store at: 

While it does place things in the right locations most of the time your goal is not to do it once on one machine but have some kind of redistributable to put on multiple machines.

Imagine you were writing a diagnostic extension hosting startup, you wouldn't want to carry hosting startup sources to every machine where apps would run. Rather you would prefer a combination of additionalDeps.json and store that you can drop into your dotnet directory

guardrex:
> some kind of redistributable to put on multiple machines.

Hence the question ...

> Can we say something about installer options? ... making delivery dead-on simple in multi-machine environments without using NuGet?

> Long-range ... after I get the topic+sample right ... we should still discuss the possibility of coverage for hosting startup delivery. How do devs supply these no-compile-time-reference hosting startups ... an installer? If so, shouldn't the topic broach how that's done?

I have no personal knowledge of installers (being a web dev), so I'll need engineering guidance on what approach to describe to readers.

pakrym:
I don't think we need to show installer here.

Collecting files into a single directory ready to be xcopied and cmd file that uses `setx` for additional deps environment variable should be enough.

pakrym:
Ideal output of HS sample would be a folder that contains:

```
additinalDeps/...
store/....
install.cmd - that sets DOTNET_ADDITIONAL_DEPS to point to additinalDeps folder
```

guardrex:
@pakrym That looks good. I'll set that up tomorrow and ping u back.

guardrex:
@pakrym I'm going to do this with PS ... not CMD ... due to the need for xplat.

guardrex:
@pakrym I'm thinking about the following for the topic/sample. I'd add this section to the topic and create the infrastructure in the sample app to do what's described ...

**Deployment**

To facilitate the distribution of the hosting startup in multi-machine environments, a target in the project file performs the following actions:

1. Creates a *deployment* folder that can be taken machine-to-machine for deployment of the hosting startup.
1. Creates a *deployment/store/&lt;RUNTIME_IDENTIFIER&gt;* folder to hold the hosting startup assembly for each runtime identifier (RID).
1. Creates a *deployment/manifest* folder to hold the hosting startup's project file for use as a manifest when placing the hosting startup's assembly in the runtime store on the machine.
1. Creates a *deployment/additionalDeps* folder to hold the hosting startup assembly's dependencies file.
1. Includes a *deployment/hosting_startup_install.ps1* PowerShell script that performs the following when run on a machine where you want to deploy the hosting startup:
   1. Specifies the hosting startup assembly's name in the `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES` environment variable. If the environment variable exists, the name is added to the list.
   1. Runs the `dotnet store` command to place the hosting startup assembly in the machine's runtime store for each runtime identifier (RID) in the hosting startup's project file.
   1. Places the hosting startup's dependencies file in the machine's *additionalDeps/&lt;HOSTING_STARTUP_ASSEMBLY_NAME&gt;* folder.
   1. Sets the `DOTNET_ADDITIONAL_DEPS` environment variable to point to the hosting startup assembly's folder inside the *additionalDeps* folder.

... and it isn't even quite that "simple" ...

* There's x86 vs. x64 to consider.
* There's taking the hosting startup's version into account.

I believe that I can pull this off. However, the topic tells the reader now that these are the steps that one should perform. It would be nice to automate the process, but I wonder if I'm (as a low-level docs dino) really the right one to do this ... shouldn't automating the deployment of hosting startups be an engineering goal?

pakrym:
> shouldn't automating the deployment of hosting startups be an engineering goal

It is for our hosting startups and we have it. It's just not reusable outside of aspnetcore builds. Building tooling for hosting startups was never a goal because the amount of people who would use it is very small.

The minimum we can do is generate additionalDeps and publish store into subfolder and tell people what has to be done to get it distributed.

guardrex:
@pakrym U cool with the latest updates :point_up:?

Rick-Anderson:
@pakrym Can you review the last and hopefully final updates?

pakrym:
Looks good. Awesome work @guardrex !

guardrex:
Thanks @pakrym ... *that means a lot to me coming from you.*

You cats are going to turn me into a halfway decent developer one day. I have the best professors anyone could ask for. Until then ... **You build it ... I'll tell 'um about it!**

aspnetcore/fundamentals/host/index.md
```diff
@@ -4,7 +4,7 @@ author: guardrex
 description: Learn about the ASP.NET Core Web Host and .NET Generic Host, which are responsible for app startup and lifetime management.
 ms.author: riande
 ms.custom: mvc
-ms.date: 05/16/2018
+ms.date: 07/27/2018
```
scottaddie: 07 --> 08

aspnetcore/fundamentals/host/index.md
```diff
@@ -15,3 +15,9 @@ uid: fundamentals/host/index
 * [Generic Host](xref:fundamentals/host/generic-host) (ASP.NET Core 2.1 or later) &ndash; Suitable for hosting non-web apps (for example, apps that run background tasks). In a future release, the Generic Host will be suitable for hosting any kind of app, including web apps. The Generic Host will eventually replace the Web Host.
 
 For hosting ASP.NET Core *web apps*, developers should use the Web Host based on [IWebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.iwebhostbuilder). For hosting *non-web apps*, developers should use the Generic Host based on [HostBuilder](/dotnet/api/microsoft.extensions.hosting.hostbuilder).
+
+<xref:fundamentals/host/hosted-services>  
+Learn how to implement background tasks with hosted services in ASP.NET Core.
+
+<xref:fundamentals/configuration/platform-specific-configuration>  
+Discover how to enhance an ASP.NET Core app from a referenced  or unreferenced assembly using an `IHostingStartup` implementation.
```
scottaddie: - Remove the extra space after "referenced".
- Link to the IHostingStartup API ref

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -0,0 +1,320 @@
...
+To discover loaded hosting startup assemblies, enable logging and check the app's logs. Errors that occur when loading assemblies are logged. Loaded hosting startup assemblies are logged at the Debug level, and all errors are logged.
+
+## Disable automatic loading of hosting startup assemblies
+
+::: moniker range=">= aspnetcore-2.1"
+
+To disable automatic loading of hosting startup assemblies, use one of the following approaches:
+
+* To prevent all hosting startup assemblies from loading, set one of the following to `true` or `1`:
+  - [Prevent Hosting Startup](xref:fundamentals/host/web-host#prevent-hosting-startup) host configuration setting.
```
scottaddie: Convert dashes to asterisks

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -0,0 +1,320 @@
...
+
+### Console app without an entry point
+
+*This approach is only available for .NET Core apps, not .NET 4.x.*
+
+A dynamic hosting startup enhancement that doesn't require a compile-time reference for activation can be provided in a console app without an entry point. The app contains a `HostingStartup` attribute. To create a dynamic hosting startup:
+
+1. An implementation library is created from the class that contains the `IHostingStartup` implementation. The implementation library is treated as a normal package.
+1. A console app without an entry point references the implementation library package. A console app is used because:
+   - A dependencies file is a runnable app asset, so a library can't furnish a dependencies file.
```
scottaddie: Use asterisks instead of dashes

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -0,0 +1,320 @@
...
+After the hosting startup is built, the hosting startup's project file serves as the manifest file for the [dotnet store](/dotnet/core/tools/dotnet-store) command.
+
+```console
+dotnet store --manifest <PROJECT_FILE> --runtime <RUNTIME_IDENTIFIER>
+```
+
+This command places the hosting startup assembly and other dependencies that aren't part of the shared framework in the user profile's runtime store at:
+
+```
+<DRIVE>\Users\<USER>\.dotnet\store\x64\<TARGET_FRAMEWORK_MONIKER>\<ENHANCEMENT_ASSEMBLY_NAME>\<ENHANCEMENT_VERSION>\lib\<TARGET_FRAMEWORK_MONIKER>\
```
scottaddie: This path looks specific to Windows. What about macOS and Linux?

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -0,0 +1,320 @@
...
+```
+
+If you desire to place the assembly and dependencies for global use, add the `-o|--output` switch to the `dotnet store` command with the following path:
+
+```
+<DRIVE>\Program Files\dotnet\store\x64\<TARGET_FRAMEWORK_MONIKER>\<ENHANCEMENT_ASSEMBLY_NAME>\<ENHANCEMENT_VERSION>\lib\<TARGET_FRAMEWORK_MONIKER>\
```
scottaddie: This is Windows specific again. I recommend using OS tabs.

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -0,0 +1,320 @@
...
+
+* The enhanced app's project file makes an assembly reference to the hosting startup (a compile-time reference). With the compile-time reference in place, the hosting startup assembly and all of its dependencies are incorporated into the app's dependency file (*\*.deps.json*).
+* The hosting startup's dependencies file is made available to the enhanced app as described in the [Runtime store](#runtime-store) section (without a compile-time reference).
+
+## Sample code
+
+The [sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/fundamentals/host/platform-specific-configuration/samples/) ([how to download](xref:tutorials/index#how-to-download-a-sample)) demonstrates three hosting startup implementation scenarios:
+
+* Two hosting startup assemblies (class libraries) set a pair of in-memory configuration key-value pairs each:
+  - NuGet package (*HostingStartupPackage*)
```
scottaddie: Convert dashes to asterisks

aspnetcore/fundamentals/host/platform-specific-configuration/samples/2.x/HostingStartupApp/HostingStartupApp.csproj
```diff
@@ -0,0 +1,36 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.App" Version="2.1.2" />
```
scottaddie: Upgrade to 2.1.3

aspnetcore/fundamentals/host/platform-specific-configuration/samples/2.x/HostingStartupApp/HostingStartupApp.csproj
```diff
@@ -0,0 +1,36 @@
...
+    The following PropertyGroup sets the package restore sources for the app.
+
+    The HostingStartupPackage bin/Debug folder is added.
+    
+    This permits the app to restore the HostingStartupPackage from its 
+    project's published output. There is no need to publish the package to 
+    nuget.org.
+  -->
+  <PropertyGroup>
+    <RestoreSources>$(RestoreSources);https://api.nuget.org/v3/index.json;../HostingStartupPackage/bin/Debug</RestoreSources>
```
scottaddie: Isn't this typically defined in a _NuGet.config_ file? What's the advantage of putting it here?

aspnetcore/fundamentals/host/platform-specific-configuration/samples/2.x/HostingStartupPackage/HostingStartupPackage.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageId>HostingStartupPackage</PackageId>
+    <Version>1.0.0</Version>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore" Version="2.1.2" />
```
scottaddie: Upgrade to 2.1.3

guardrex:
@scottaddie @pakrym These look correct based on my iMac research.

For the user profile versions ...

```
/Users/<USER>/.dotnet/store/x64/...
```

For the machine-wide versions ...

```
/usr/local/share/dotnet/store/x64/...
```

I also updated the Windows ones to use env vars ...

```
%USERPROFILE%\.dotnet\store\x64\...
%PROGRAMFILES%\dotnet\store\x64\...
```

aspnetcore/fundamentals/host/platform-specific-configuration/samples/2.x/HostingStartupApp/HostingStartupApp.csproj
dasMulli: Two things here:
1. pointing to a bin/Debug output as restore source is a bit hacky but I don't know the context here. (what if it fails, what if NuGet decides that you haven't changed any feeds/references and does a no-op (incremental) restore etc.)

2. There's also `RestoreAdditionalProjectSources` for project-specific sources. But now idea how well a "manage package for solution" dialog works for this.

The property-based approach is nice if you don't want to explain NuGet.config and the mechanism behind and it is easy to set in `Directory.Build.props` for all projects, even for local paths by using `$(MSBuildThisFileDirectory)local\nupkgs` from it. Also, you can make use of standard MSBuild conditions to alter the restore sources which is why dotnet/aspnet moved to them to be able to use different sources for different types of builds which would require editing NuGet.config for every build otherwise.

guardrex: @dasMulli 

True ... it's a bit of a hack. It has the following benefits/goals in this context:

1. It allows the dev to create and use the *HostingStartupPackage* NuGet package in the *HostingStartupApp* by just building that *HostingStartupPackage* NuGet package project and then building/running the *HostingStartupApp*.
1. They don't need to move anything (i.e., they don't need to move the NuGet package out of the *HostingStartupPackage* project). They don't need to take any additional steps in *HostingStartupApp* to find and consume the package.
1. They don't need to put the package on nuget.org. **_That's the really important goal here._**

> you haven't changed any feeds/references and does a no-op (incremental) restore

I thought that if the NuGet package is rebuilt that NuGet will pick up on that (without a package version change). Does NuGet go purely by the package's version on incremental restore? That is a problem if that's the case. I would like to have this set up so that the dev can make a change to the NuGet package project and have it picked up on restore/run of the *HostingStartupApp*.

> RestoreAdditionalProjectSources 

Can the *additional project source* be a NuGet package that is then consumed by the app? If I can use `RestoreAdditionalProjectSources` pointed to the NuGet package project, that would be great ... but is that going to work for consuming a NuGet package? If so, do I then drop the `<RestoreSources>` property?

dasMulli: @guardrex AFAIK you'd have to do a `dotnet nuget locals all --clear` if you change package contents without changing the version anyway.
That's why you should always use project (P2P) references during development or use a local "global packages cache" to nuke in between full rebuilds (CI runs).

dasMulli: Why not use a project reference?

guardrex: Thanks @dasMulli ... I pushed a commit to make sure devs know that if they change the package project to clear caches. That looks good.

I'm going to leave the `<RestoreSources>` in the app as is. Unless @pakrym says 'don't do that,' it seems like a very clean, easy-to-use, easy-to-grok setup for the demo. I'm just trying to show how the hosting startup consumer takes *a real package*, and I don't want the dev to have to put a hosting startup package on nuget.org or need to sweat a *Nuget.config* file with the source (which I think buries how the source is being set).

Your advice on the caches should go well :+1:. They just need to do that each time they make changes to the hosting startup package. 🤞

https://github.com/aspnet/Docs/pull/7678/commits/37562f03d2d3d4a5c1d8ace2bae44ee5af38918c

guardrex:
@scottaddie Unless there's anything else, I think we're in good shape here. The paths for Mac/Linux seem good based on my iMac research here at the house. I'm sure the community will come for my head 🔪 :boy: if I made a *boo boo*, but my iMac is saying these are good. (We'll blame my iMac if they're wrong. 🙈)

guardrex:
Thanks again to @dasMulli! :rocket: :rocket: :rocket: ... the tip for clearing local caches is critical for the dev who wants to play with the NuGet package approach and get their hosting startup package updated in the app. That was a 🎸 *rockstar suggestion* 🎷.

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -0,0 +1,451 @@
...
+
+The app's Index page reads and renders the configuration values for the two keys set by the package's hosting startup assembly:
+
+*HostingStartupApp/Pages/Index.cshtml.cs*:
+
+[!code-csharp[](platform-specific-configuration/samples/2.x/HostingStartupApp/Pages/Index.cshtml.cs?name=snippet1&highlight=7-8,13-14)]
+
+### Console app without an entry point
+
+*This approach is only available for .NET Core apps, not .NET 4.x.*
```
scottaddie: .NET 4.x.* --> .NET Framework

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -0,0 +1,451 @@
...
+
+When multiple hosting startup assembles are present, their [Configure](/dotnet/api/microsoft.aspnetcore.hosting.ihostingstartup.configure) methods are executed in the order that the assemblies are listed.
+
+## Activation
+
+Three options for hosting startup activation are:
+
+* [Runtime store](#runtime-store) &ndash; Activation doesn't require a compile-time reference for activation. The sample app places the hosting startup assembly and dependencies files into a folder, *deployment*, to facilitate deployment of the hosting startup in a multimachine environment. The *deployment* folder also includes a PowerShell script that creates or modifies environment variables on the deployment system to enable the hosting startup.
+* Compile-time reference required for activation
+  - [NuGet package](#nuget-package)
```
scottaddie: Can you use asterisks here instead of dashes?

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -0,0 +1,451 @@
...
+### Project bin folder
+
+A hosting startup enhancement can be provided by a *bin*-deployed assembly in the enhanced app. The hosting startup types provided by the assembly are made available to the app using either of the following approaches:
+
+* The enhanced app's project file makes an assembly reference to the hosting startup (a compile-time reference). With the compile-time reference in place, the hosting startup assembly and all of its dependencies are incorporated into the app's dependency file (*\*.deps.json*).
+* The hosting startup's dependencies file is made available to the enhanced app as described in the [Runtime store](#runtime-store) section (without a compile-time reference).
+
+## Sample code
+
+The [sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/fundamentals/host/platform-specific-configuration/samples/) ([how to download](xref:tutorials/index#how-to-download-a-sample)) demonstrates three hosting startup implementation scenarios:
```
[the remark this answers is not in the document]
guardrex: This is similar to the problem you pointed about above: There are ... for lack of a better expression ... three options that call under two approaches. *ei ei ei*. I think the word "three" can be removed to solve it most easily.

aspnetcore/fundamentals/host/platform-specific-configuration/samples/2.x/HostingStartupApp/HostingStartupApp.csproj
```diff
@@ -0,0 +1,36 @@
...
+  </PropertyGroup>
+
+  <!--
+    The following ItemGroup permits the app to load the HostingStartupLibrary
+    from the app's bin folder.
+  -->
+  <ItemGroup>
+    <Reference Include=".\bin\Debug\netcoreapp2.1\HostingStartupLibrary.dll">
```
scottaddie: @dasMulli,
It does seem hacky to hardcode certain pieces of this path. Should they be converted to variables? For example:
- Debug --> $(Configuration)
- netcoreapp2.1 --> $(TargetFramework)
- HostingStartupLibrary.dll --> $(AssemblyName)

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -0,0 +1,451 @@
...
+   * Moves the *StartupDiagnostics.deps.json* file to the user profile's *additionalDeps* folder.
+1. Execute the `dotnet store` command at a command prompt in the hosting startup's directory to store the assembly and its dependencies in the user profile's runtime store:
+
+   ```console
+   dotnet store --manifest StartupDiagnostics.csproj --runtime win7-x64
+   ```
+   For Windows, the command uses the `win7-x64` runtime identifier (RID). When providing the hosting startup for a different runtime, substitute the correct RID.
+1. Set the environment variables:
+   * Add the assembly name of *StartupDiagnostics* to the `ASPNETCORE_HOSTINGSTARTUPASSEMBLIES` environment variable.
+   * On Windows, set the `DOTNET_ADDITIONAL_DEPS` environment variable to `%UserProfile%\.dotnet\x64\additionalDeps\StartupDiagnostics\`. On macOS/Linux, set the `DOTNET_ADDITIONAL_DEPS` environment variable to `/Users/<USER>/.dotnet/x64/additionalDeps/StartupDiagnostics/`, where `<USER>` is the user profile that contains the hosting startup.
```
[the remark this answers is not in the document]
guardrex: I fixed the command by showing `<RID>`. Therefore, there's only one sentence left specific to non-Windows ... this one. Let's see on the next commit. It might not be worth it to copy the entire set of steps into three tabs only to accommodate the single sentence.

aspnetcore/fundamentals/host/platform-specific-configuration/samples/2.x/HostingStartupApp/HostingStartupApp.csproj
guardrex: I think this is a great idea. We do that elsewhere, and we probably should do it here, too.

I'll use different names. They refer to characteristics of another project (i.e., the hosting startup package project). I'll make it clear in their names on the next commit.

dasMulli: why is the reference here in the first place? shouldn't it be coming from the `HostingStartupPackage` package reference?

guardrex: idk on this now ....... this is taking something very simple and making it *very messy*. It takes three vars for both to define them. It makes it harder to see what it is .... just a path. I'd rather be super clear on the components of the path than have a bunch of vars stuck in here.

dasMulli: So.. one is a package reference and the other a `Reference` instead of a project-to-project reference..

I'd argue that the `Reference` should really be a `ProjectReference` instead, so it is built, up-to-date and you get the correct path.

dasMulli: btw i'd prefer `<Reference Include="path/to/some.dll" />` instead of the hint path approach.

guardrex: Yes on the first one. The package is a different demo from the library. These are two different  compile-time approaches.

On the second point ... The topic's demos include the *deployment strategy*. It's similar to our earlier discussion on why I'm working with package sources and not a project reference to the hosting startup package project. Same thing here: AFAICT from what the engineers have delivered for this scenario, the hosting startup is **_deployed_** as a DLL. Therefore, this doesn't use a project ref. It references the DLL directly (and this will change for any given organization/environment) ... or the DLL will be dropped directly into the consuming project.

I think what's missing here is a little more guidance in the project file. Let me see what I can do on the next commit with the explanation. I'll check the topic, too, to see what I need to do there.

dasMulli: Ah yeah so in theory it could also be a `<Content Include="../startup.dll" CopyToOutputDirectory="PreserveNewest" />`.
(however, MSBuild also considers `@(Content)` for assembly resolution which is kind of weird).

guardrex: Sorry about that @scottaddie ... I just noticed that you are correct about the instructions and location for the hosting startup library. The instructions are correct ... I didn't recall that I have the dev deploy the assembly to the consuming project. These scenarios aren't exactly easy to track. I'll add project file language to reinforce the concept, and I'll add the vars you recommended.

☕️ ....... *one more cup please!*

dasMulli: >Debug --> $(Configuration)
>netcoreapp2.1 --> $(TargetFramework)
>HostingStartupLibrary.dll --> $(AssemblyName)

All of these only apply when the referenced project is built under the exact same parameters.
* `$(Configuration)` -> Probably okay if they are built in the same solution and users didn't mess with solution configurations.
* `$(TargetFramework)` -> as long as the library isn't a `netstandard*`...
* `$(AssemblyName)` -> only if the referenced project's `.csproj` file has the exact same name as the currently built project.. don't think that will happen

guardrex: > All of these only apply when the referenced project is built under the exact same parameters.

It's just for the path tho (within the consuming project; I mean the configuration and the target framework vars), so the path to the DLL should be correct.

I need to think this over a little more. Everything we've discussed has merit, but this is on a slippery slope to being so ✨ _magical_ ✨ that it clouds the basic concepts.

dasMulli: there's also..
```
<ProjectReference Include="../some/other.csproj" ReferenceOutputAssembly="false" />
```

guardrex: I'm fleshing out some details in the app's project file to explain what's going on.

I may also mirror some or all of the new remarks in the topic itself.

It will be on the next commit.

guardrex:
Ok ... here's what I came up with.

I provide additional notes to hopefully make it clear what's going on in the project file.

I don't want those notes in the topic. These are sample demo implementation details. They're a bit beyond the scope of the hosting startup basic concepts. The topic states plainly what the options are and links out where needed (e.g., NuGet packages).

I add a couple of lines where the deployment options are described explaining the circumstances where the two main compile-time approaches apply. In the first case, it would be a NuGet package published to nuget.org. In the lib case, it would be a compiled assembly moved to some location where an app can take a compile-time ref on it's assembly.

Let's deal with any confusion that arises in the community as we get feedback. I'm sure they'll let me know if things turn sour.

PR #7693, opened 2018-07-19
Fix samples link for signalr webpack doc

BrennanConroy:


scottaddie:
Thank you for catching that, @BrennanConroy!

PR #7701, opened 2018-07-20
Updated link to follow-up tutorial

KirkMunro:
The current follow-up tutorial link takes users to an MVC tutorial based on ASP.NET Core 2.0. This tutorial is based on Razor Pages and ASP.NET Core 2.1. I think it makes more sense for the recommended follow-up tutorial to continue building on the knowledge learned here about Razor Pages, so I updated the link to take users to [Razor Pages with Entity Framework Core in ASP.NET Core - Tutorial 1 of 8](https://docs.microsoft.com/en-us/aspnet/core/data/ef-rp/intro) instead.


Rick-Anderson:
@KirkMunro  thanks!

PR #7710, opened 2018-07-20
Clarify handshake options in SignalR

anurse:
See https://github.com/aspnet/Docs/issues/7691

### DIFF
--- a/aspnetcore/signalr/configuration.md
+++ b/aspnetcore/signalr/configuration.md
@@ -56,7 +56,7 @@ The following table describes options for configuring SignalR hubs:
 
 | Option | Description |
 | ------ | ----------- |
-| `HandshakeTimeout` | If the client doesn't send an initial handshake message within this time interval, the connection is closed. |
+| `HandshakeTimeout` | If the client doesn't send an initial handshake message within this time interval, the connection is closed. This is an advanced setting that should only be modified if the client or network are particularly slow, causing handshake timeout errors. For more detail on the Handshake process, see the [SignalR Hub Protocol Specification](https://github.com/aspnet/SignalR/blob/master/specs/HubProtocol.md) |
 | `KeepAliveInterval` | If the server hasn't sent a message within this interval, a ping message is sent automatically to keep the connection open. |
 | `SupportedProtocols` | Protocols supported by this hub. By default, all protocols registered on the server are allowed, but protocols can be removed from this list to disable specific protocols for individual hubs. |
 | `EnableDetailedErrors` | If `true`, detailed exception messages are returned to clients when an exception is thrown in a Hub method. The default is `false`, as these exception messages can contain sensitive information. |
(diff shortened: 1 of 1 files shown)


aspnetcore/signalr/configuration.md
```diff
@@ -56,7 +56,7 @@ The following table describes options for configuring SignalR hubs:
 
 | Option | Description |
 | ------ | ----------- |
-| `HandshakeTimeout` | If the client doesn't send an initial handshake message within this time interval, the connection is closed. |
+| `HandshakeTimeout` | If the client doesn't send an initial handshake message within this time interval, the connection is closed. This is an advanced setting that should only be modified if the client or network are particularly slow, causing handshake timeout errors. For more detail on the Handshake process, see the [SignalR Hub Protocol Specification](https://github.com/aspnet/SignalR/blob/master/specs/HubProtocol.md) |
```
guardrex: If "causing handshake timeout errors" is a (suggested) requirement for setting the option, drop the comma before "causing." If that phrase is just an example of one thing that might accompany a slow client/network such that the option might be set under other slow client/network conditions, the comma is :+1:.

Either way, it needs a period at the end ... after the *SignalR Hub Protocol Specification* link.

aspnetcore/signalr/configuration.md
```diff
@@ -211,7 +211,7 @@ Additional options for configuring timeout and keep-alive behavior are available
 | .NET Option | JavaScript Option | Description |
 | ----------- | ----------------- | ----------- |
 | `ServerTimeout` | `serverTimeoutInMilliseconds` | Timeout for server activity. If the server hasn't sent any message in this interval, the client considers the server disconnected and trigger the `Closed` event (`onclose` in JavaScript). |
-| `HandshakeTimeout` | Not configurable | Timeout for initial server handshake. If the server doesn't send a handshake response in this interval, the client cancels the handshake and trigger the `Closed` event (`onclose` in JavaScript). |
+| `HandshakeTimeout` | Not configurable | Timeout for initial server handshake. If the server doesn't send a handshake response in this interval, the client cancels the handshake and trigger the `Closed` event (`onclose` in JavaScript). This is an advanced setting that should only be modified if the server or network are particularly slow, causing handshake timeout errors. For more detail on the Handshake process, see the [SignalR Hub Protocol Specification](https://github.com/aspnet/SignalR/blob/master/specs/HubProtocol.md) |
```
guardrex: * If the comma before "causing" is dropped above, drop it here, too.
* Needs a period at the end (after the *SignalR Hub Protocol Specification* link).

btw- There's a subject-verb agreement problem here that predates your PR ...

> the client cancels the handshake and trigger

... it should be "triggers"

anurse:
I clarified one additional statement, as per comments on the original issue.

aspnetcore/signalr/configuration.md
anurse: I did a bit of a hack job at trying to clarify it. The setting should only be changed if you are seeing handshake timeouts **and** those timeouts are due to a slow network (i.e. you expect delays in transmitting data).

To be perfectly honest, this isn't a setting I'd ever expect a user to need to configure, it's there because we're always hesitant to have hard-coded timeouts. The defaults are just fine in 99.999% of cases.

anurse:
Derp, forgot the verb change you mentioned, thanks for grabbing that :). I'm good!

PR #7712, opened 2018-07-20
Highlight one more line

RyanTaite:
This line originally generated is `public IList<Instructor> Instructor { get;set; }` and gets changed to `public InstructorIndexData Instructor { get; set; }` so it's worth highlighting, I feel


scottaddie:
@RyanTaite Thank you for fixing this!

Issue #6230, opened 2018-05-05
HTML.ActionLink not working


thomasiverson:
It appears there is an issue with the Html.ActionLink on Pages/Instructors/Index.cshtml - line 87). It doesn't seem to be rendering the link. 

@Html.ActionLink("Select", "OnGetAsync", 
                                  new { courseID = item.CourseID })

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: d0026f69-7e5d-0d3e-5586-16ecc0656ff0
* Version Independent ID: 3cb264da-2ff1-1256-572f-aa4ebca6e255
* Content: [Razor Pages with EF Core in ASP.NET Core - Read Related Data - 6 of 8](https://docs.microsoft.com/en-us/aspnet/core/data/ef-rp/read-related-data?view=aspnetcore-2.0#explicit-loading)
* Content Source: [aspnetcore/data/ef-rp/read-related-data.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/data/ef-rp/read-related-data.md)
* Service: **unspecified**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
@thomasiverson  there is no such `Select` page. I think it should be the Courses details page.

Rick-Anderson:
Here's the correct code from the MVC version
`                    @Html.ActionLink("Select", "Index", new { courseID = item.CourseID })`

zedOdin:
@thomasiverson I have the same problem.
I think there are unusual security error, because I created another DataBase with other tables/content and it's working. But, when I create a new Project with Individual User Accounts with the same new code for DataBase+Seed, Html.ActionLink doesn't work.
Did you solved your problem?
@Rick-Anderson, I'm missing something about Security?, i'm a newbie and sorry for my English.

Rick-Anderson:
@zedOdin It wouldn't be a security issue. For `Html.ActionLink`  to generate a URL, the action must exist and the PageModel must match.

ksnzau:
I've got the same issue, in c#,  _public async Task OnGetAsync(int? id, int? courseID)_  courseID value is null, never passed through.

Rick-Anderson:
@ksnzau let us know if you figure out what you did wrong. The sample download works.

thomasiverson:
@ksnzau, @zedOdin, CourseID is null for me too.  
[...]

PR #7717, opened 2018-07-20
Update Select HTML.ActionLink to use asp tag helpers

RyanTaite:
`<td>@Html.ActionLink("Select", "OnGetAsync", new { courseId = item.CourseID })</td>` throws an error saying:

> Cannot resolve action 'OnGetAsync'

I also fixed a spacing issue

<hr />

I see a now closed issue: #6230 that seems to also have trouble with this link, but does not mention this specific error.

The only similarity I have with that issue is that I also created the project with[ Individual User Accounts as the security](https://github.com/aspnet/Docs/issues/6230#issuecomment-390510927), though as Rick-Anderson [mentioned ](https://github.com/aspnet/Docs/issues/6230#issuecomment-390786825) it's likely not the problem.


Rick-Anderson:
@RyanTaite  much better, thanks.

PR #7719, opened 2018-07-21
Removed out of date text

KirkMunro:
The text removed references an old issue before EF Core 2.1 was released. Now that EF Core 2.1 is released, this text is no longer current.


scottaddie:
@KirkMunro You are correct. Thank you for finding this and fixing it!

PR #7727, opened 2018-07-21
Removed EF Core 2.1 note

KirkMunro:
EF Core 2.1 is out and already supports data seeding, so this note is no longer relevant.


scottaddie:
@KirkMunro Another nice improvement to the doc. Thank you again!

PR #7784, opened 2018-07-24
Update complex-data-model.md

GoFightNguyen:
Removed duplication in the Update the Enrollment entity header.


scottaddie:
@GoFightNguyen Nice catch! Thank you for taking the time to fix it.

PR #7786, opened 2018-07-24
I25974: Finxing spacing/tabulation to avoid localization issues

v-maudel:
Hello, @rick-anderson,
"It seems some spaces/Tabulations make some content as part of non-translatable content"

Many thanks in advance.

### DIFF
--- a/aspnetcore/getting-started/index.md
+++ b/aspnetcore/getting-started/index.md
@@ -29,11 +29,11 @@ uid: getting-started
     dotnet dev-certs https --trust
     ```
 
-    The preceding command displays the following dialog:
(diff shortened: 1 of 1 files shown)


guardrex:
> It seems some spaces/Tabulations make some content as part of non-translatable content

If that's the case, there's probably something amiss in the build system. We have some number of other topics that use four space indentation in lists. If we take this, we would need to update those topics, and there's a question on why code blocks aren't indented but the text is indented.

v-maudel:
Thanks @Rick-Anderson 
@guardrex , it happen in the Github editor as well. the strings I modified were in grey color, as they were part of code or something like that.

PR #7787, opened 2018-07-24
C28229: Escaping URL that must not be shown as link

v-maudel:
Hello, @pfletcher ,
"Escaping URL that must not be shown as link"

Many thanks in advance.


scottaddie:
Thank you, @v-maudel!

v-maudel:
Thanks to you @scottaddie .

PR #1386, opened 2016-06-14
Update model-binding.rst

SyniaN:
typo should be "as long as **the** keys....", instead of "as long as **they** keys"

PR #7788, opened 2018-07-24
Update windows-service.md

eyal7773:
I offer to insert an instruction to validate the the AspNetCore project is configured as "Windows Application", becuase if it configured as "Console Application" - so when the developer will try to start service with "sc start....", he will got an error : "the service did not respond to the start or control request in a timely fashion "

 Fixes #1386

Thanks
Eyal

### DIFF
--- a/aspnetcore/host-and-deploy/windows-service.md
+++ b/aspnetcore/host-and-deploy/windows-service.md
@@ -76,7 +76,7 @@ The following minimum changes are required to set up an existing ASP.NET Core pr
 
      ::: moniker-end
      
-     * Configure the project as "Windows Application". To do so, In Visual Studio, right click on the project > Properties > on "Application" tab > set "Output type" field to "Windows Application".
+     * Configure the project as a "Windows Application".  In Visual Studio, right click on the project > Properties > Application tab > set "Output type" field to "Windows Application".
 
 1. Publish the app. Use [dotnet publish](/dotnet/articles/core/tools/dotnet-publish) or a [Visual Studio publish profile](xref:host-and-deploy/visual-studio-publish-profiles).
 
--- a/aspnetcore/host-and-deploy/windows-service.md
+++ b/aspnetcore/host-and-deploy/windows-service.md
@@ -75,6 +75,8 @@ The following minimum changes are required to set up an existing ASP.NET Core pr
      [!code-csharp[](windows-service/samples_snapshot/1.x/AspNetCoreService/Program.cs?name=ServiceOnly&highlight=3-4,8,13)]
 
      ::: moniker-end
+     
+     * Configure the project as "Windows Application". To do so, In Visual Studio, right click on the project > Properties > on "Application" tab > set "Output type" field to "Windows Application".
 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@eyal7773  do you know the related issue, 1386 is not correct.

guardrex:
Thanks @eyal7773 ... I'll take a look at this by the end of the day.

guardrex:
Sorry @eyal7773 ... I didn't get to this yesterday. I'll try again today. 🏃😅

guardrex:
Hello @eyal7773 ... I finally just got free to take a look. I think the problem is that you either aren't publishing in VS or you aren't using the published output from a VS folder publish.

Leave the output type set to "Console Application" for your test here:

1. Select **Build** > **Publish &lt;YourAppName&gt;**.
1. Select the **FolderProfile**.
1. Select the **Publish** button.

It will publish and output the path ...

```
Web App was published successfully
file:///C:/_apps/WindowsServiceSample/bin/Release/netcoreapp2.1/win7-x64/publish/
```

Use that path for your `sc create` and `sc start` commands. It should work. Let me know how it goes.

For the topic, I think where it says ...

> 1. Publish the app. Use \[dotnet publish](/dotnet/articles/core/tools/dotnet-publish) or a \[Visual Studio publish profile](xref:host-and-deploy/visual-studio-publish-profiles).

... we should make it say ...

> 1. Publish the app. Use \[dotnet publish](/dotnet/articles/core/tools/dotnet-publish) or a \[Visual Studio publish profile](xref:host-and-deploy/visual-studio-publish-profiles). When using a Visual Studio, select the \*\*FolderProfile\*\*.

That's :point_up: a cut-and-paste for the line if you'd like to swap the change you made earlier for this one, then I can merge your PR.

After that, it should all :sparkles: **_Just Work_**:tm: :sparkles:. A few steps later, the reader will see that the path is to the *publish* folder ... same situation as when not using VS (e.g., VSC or command-line).

guardrex:
That's interesting, because when you set that in the property page in VS (before you published from a command prompt) what it's doing is setting the `OutputType` to `WinExe` in the project file ...

```xml
<OutputType>WinExe</OutputType>
```

... and that's not the Windows Service configuration guidance that we received from the engineer.

You're sure you're using the path to the *publish* folder when you create the service? It really should work with an `OutputType` of `Exe`, which is what's used when `OutputType` isn't specified in the project file. There might be something else going on that makes it work in your case, but we'll need to check with the engineer to explore that further.

First, set the `OutputType` to `Exe` (if present in the project file; if not present, you don't need to set it), then publish (`dotnet publish -c Release`), then create the service (confirm ur using the path that includes the *publish* directory), and try to start it again.

```console
sc create MyService binPath= "c:\my_services\AspNetCoreService\bin\Release\netcoreapp2.1\win7-x64\publish\AspNetCoreService.exe"
```

... and I need to make one more change. The RID is missing from the line in the topic, so I'll add that at the same time as making the update above.

guardrex:
@eyal7773 Would you mind reviewing and signing the CLA :point_up: if you agree with terms. We can't accept the PR until the CLA is signed.

guardrex:
Moved to #7863

PR #7800, opened 2018-07-24
Enhance DI remarks in Startup topic

### LINKED ISSUE #6783: Dependency injection with convenience methods
Using the convenience methods, how can I access services such as ILoggerFactory provided by the hosting service? Using Startup, they are available via dependency injection in the constructor and Configure. What is the equivalent for the convenience methods?

---
#### Document Details

* Content: [Application startup in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/startup?view=aspnetcore-2.1)
* Content Source: [aspnetcore/fundamentals/startup.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/startup.md)
* Service: **unspecified**
* Product: **asp.net-core**
* GitHub Login: @ardalis

guardrex:
Fixes #6783

### DIFF
--- a/aspnetcore/fundamentals/startup/snapshot_sample/Startup2.cs
+++ b/aspnetcore/fundamentals/startup/snapshot_sample/Startup2.cs
@@ -2,29 +2,31 @@ public class Startup
 {
     private readonly IHostingEnvironment _env;
     private readonly IConfiguration _config;
-    private readonly ILogger<Startup> _logger;
+    private readonly ILoggerFactory _loggerFactory;
 
     public Startup(IHostingEnvironment env, IConfiguration config, 
-        ILogger<Startup> logger)
+        ILoggerFactory loggerFactory)
     {
         _env = env;
         _config = config;
-        _logger = logger;
+        _loggerFactory = loggerFactory;
     }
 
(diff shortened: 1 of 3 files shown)


aspnetcore/fundamentals/startup/snapshot_sample/Startup2.cs
```diff
@@ -1,27 +1,35 @@
 public class Startup
 {
-    public Startup(IHostingEnvironment env, IConfiguration config)
+    private readonly IHostingEnvironment _env;
+    private readonly IConfiguration _config;
+    private readonly ILogger<Startup> _logger;
+
+    public Startup(IHostingEnvironment env, IConfiguration config, 
+        ILogger<Startup> logger)
```
Tratcher: The bug was about ILoggerFactory, not ILogger.

guardrex:
@Tratcher Do you think it will be enough for the service locator pattern used with `Configure` in the convenience method to merely show the example in code, or do you think it requires a line or two under the example. My concern here is that if a dev is reading our docs start-to-finish, this topic appears fairly early in their readings. However if you feel that's easy to grok, then we can leave it as it is.

PR #7812, opened 2018-07-25
Enhance text for first Enforce HTTPS example

### LINKED ISSUE #7811: Primary example should address most common use case http to https port 443
The example provided does not address the most common use case of having a redirect from http to https (on port 443). Feel free to provide a second example for edge cases.

---
#### Document Details

* Content: [Enforce HTTPS in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?view=aspnetcore-2.1&tabs=visual-studio#require-https)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

guardrex:
Fixes #7811

### DIFF
--- a/aspnetcore/security/enforcing-ssl.md
+++ b/aspnetcore/security/enforcing-ssl.md
@@ -32,6 +32,11 @@ The following code calls `UseHttpsRedirection` in the `Startup` class:
 
 [!code-csharp[](enforcing-ssl/sample/Startup.cs?name=snippet1&highlight=13)]
 
+The preceding highlighted code:
+
+* Uses the default [HttpsRedirectionOptions.RedirectStatusCode](/dotnet/api/microsoft.aspnetcore.httpspolicy.httpsredirectionoptions.redirectstatuscode) (`Status307TemporaryRedirect`). Production apps should call [UseHsts](#hsts).
+* Uses the default [HttpsRedirectionOptions.HttpsPort](/dotnet/api/microsoft.aspnetcore.httpspolicy.httpsredirectionoptions.httpsport) (443).
+
 The following code calls [AddHttpsRedirection](/dotnet/api/microsoft.aspnetcore.builder.httpsredirectionservicesextensions.addhttpsredirection) to configure middleware options:
 
 [!code-csharp[](enforcing-ssl/sample/Startup.cs?name=snippet2&highlight=14-99)]


barry-schneider:
I do not feel this solution fixes #7811 which was about having the primary example address the most common use case

```
{
    if (env.IsDevelopment())
    {
        app.UseDeveloperExceptionPage();
    }
    else
    {
        app.UseHsts();
    }

    app.UseHttpsRedirection();
    app.UseStaticFiles();
    app.UseCookiePolicy();

    app.UseMvc();
}

{
    services.AddMvc();

    services.AddHsts(options =>
    {
        options.Preload = true;
        options.IncludeSubDomains = true;
        options.MaxAge = TimeSpan.FromDays(60);
        options.ExcludedHosts.Add("example.com");
        options.ExcludedHosts.Add("www.example.com");
    });

    services.AddHttpsRedirection();              
}
```

barry-schneider:
I think you might need to set both app.UseHttpsRedirection() and services.AddHttpsRedirection(); to get redirect to work. This is what I had to do. Please verify with the middleware developer or test yourself. Thanks

guardrex:
@barry-schneider, adding the service to the container isn't required.

I'll note it in an update for the topic.

barry-schneider:
I disagree that adding the service to the container isn't required. Please test yourself. In VS 2017, I just created asp.net core 2.1 web application with configured for https checked. When I run locally I am redirected to https but when I deploy to azure and browse site I am not redirected to https.

image

image

image

After adding to startup.cs file and publishing, the site redirects

```
   services.AddHttpsRedirection(options =>
            {
                options.HttpsPort = 443;
            });
```
image

barry-schneider:
you need to add the options because addthhpsredirect requires an option paramter

image

guardrex:
Yep ... totally forgot that I pushed this back to https://github.com/aspnet/Docs/issues/6538.

https://github.com/aspnet/Home/issues/3176#issuecomment-395496161

That PR is still open ... I'll add it shortly. Thanks for the reminder. We have a lot going on these days. 🏃

guardrex:
There's already a *cryptic* note in the topic about this ...

> \[!NOTE]
> When an app is run behind a reverse proxy (for example, IIS, IIS Express), `IServerAddressesFeature` isn't available. The port must be manually configured. When the port isn't set, requests aren't redirected.

... note it doesn't say "Azure App Service."

I'll see if I can surface that concept more plainly in the open text where the port config is described. I'll also refactor (probably remove) the "443 is set as the default" language.

guardrex:
This is interesting ...

> Antares has a portal experience where you can check a box to enable HTTPS which is what we should guide people towards

@shirhatti Do you prefer that we *NOT* guide devs towards always setting the port? If so, I'll try to surface the Azure App Service scenario separately, and we'll leave the examples as they are with some minor text updates.

PR #7824, opened 2018-07-25
[WIP] Further enhancement to Enforce HTTPS topic

guardrex:
Addresses #7811 

Let's tell readers explicitly that they don't need to add the middleware to the service container unless they need to mod the default options.


guardrex:
I've been looking over the options, and I think @barry-schneider is right. :+1: The simplest approach that will keep most devs in a good state is not try and grok the fine print in order to get the most common desired result. The best way to go is to just show people what to do.

The topic (mostly) contains the information on what's going on in the reverse proxy scenario, but the examples show the common approach using a non-standard port (5001); so upon casual inspection, it looks like one would only configure the port if they need a non-standard port set.

I'm going to close this to make some more extensive changes than just patching the text. Let's change the sample and use a snapshot to show the non-standard port being set.

PR #7836, opened 2018-07-26
Update free Azure account link

BethMassi:
We should point .NET developers to the specific signup page for them.


scottaddie:
@CamSoper Can you please make sure this link is used in the Azure Lifecycle eBook?

PR #7844, opened 2018-07-27
Document accessing HttpContext via Razor views

scottaddie:
Fixes https://github.com/aspnet/Docs/issues/7832


davidfowl:
We need to tweak this doc. It's not how you'd access the HttpContext in a razor view. It's already there and there's no point using the IHttpContextAccessor.

scottaddie:
@davidfowl No problem. Just let me know what you'd like to see, and I'll get right on it.

PR #7879, opened 2018-07-30
WIP: Identify the package that has UseHsts

tdykstra:
@Rick-Anderson This uses a code comment to identify the package needed, as we do elsewhere for identifying `using` statements. It could go in the doc text instead.  I didn't specify the using statement because it's the same one already in use for the other UseXXX methods.

### DIFF
--- a/aspnetcore/security/enforcing-ssl/sample/Startup.cs
+++ b/aspnetcore/security/enforcing-ssl/sample/Startup.cs
@@ -48,6 +48,7 @@ namespace WebHTTPS
             else
             {
                 app.UseExceptionHandler("/Error");
+                // Requires NuGet package Microsoft.AspNetCore.HttpsPolicy
                 app.UseHsts();
             }
 
--- a/aspnetcore/security/enforcing-ssl/sample/Startup.cs
+++ b/aspnetcore/security/enforcing-ssl/sample/Startup.cs
@@ -48,7 +48,6 @@ namespace WebHTTPS
             else
             {
                 app.UseExceptionHandler("/Error");
-                // Requires NuGet package Microsoft.AspNetCore.HttpsPolicy
                 app.UseHsts();
             }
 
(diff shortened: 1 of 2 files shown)


aspnetcore/security/enforcing-ssl/sample/Startup.cs
```diff
@@ -48,6 +48,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
             else
             {
                 app.UseExceptionHandler("/Error");
+                // Requires NuGet package Microsoft.AspNetCore.HttpsPolicy
```
Rick-Anderson: That package is available via `<PackageReference Include="Microsoft.AspNetCore.App" />`
So why call it out? You could add a similar comment to every line.

tdykstra:
I missed that somehow this weekend.  It wasn't resolving UseHsts until I added that package.  I'll change this to WIP and see if I can figure out what happened.

aspnetcore/security/enforcing-ssl/sample/Startup.cs
scottaddie: For projects targeting .NET Framework, knowledge of this package name would be necessary. We should call that out somewhere in the doc.

Rick-Anderson: > For projects targeting .NET Framework, knowledge of this package name would be necessary. We should call that out somewhere in the doc.

We could do that for every call in every sample. @danroth27  do we want to start calling out NuGet packages for APIs? 

A better approach is to have /dotnet/api list the package. Currently it shows the DLL which usually has the same name format as the package - but not always.

[UseHsts](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.hstsbuilderextensions.usehsts?view=aspnetcore-2.1#Microsoft_AspNetCore_Builder_HstsBuilderExtensions_UseHsts_Microsoft_AspNetCore_Builder_IApplicationBuilder_)  Assembly:
Microsoft.AspNetCore.HttpsPolicy.dll

cc @dend  do we have plans for API Browser to list package?

tdykstra: If I were a developer reading this doc who needed a package name, I'd rather have it conveniently alongside the thing I need it for.  But I agree that including the package in the API ref would adequately address the issue.  Should I close this PR and open an API issue?

If it's going to be a long time before the API feature is added we should reconsider adding package names in the docs in places like this.  It's a really bad docs experience when you need that and it could/should have been included but isn't there and takes a lot of work to find.

danroth27: I think it would be cleaner to have our API refs identified the package for the API instead of having to mention the package IDs throughout our docs.

PR #7882, opened 2018-07-30
Update webapp-alias-notice.md

nickwalkmsft:
Historical context in https://github.com/aspnet/Docs/issues/6937 & https://github.com/aspnet/Docs/pull/6939.

The current text specifies "2.1", which wasn't specific enough (I had 2.1.200 installed and wondered why it wasn't working). This generalizes things, moves the most important part to the front and changes the download link to the .NET Core download home page instead of the version list.

If you want to keep the link pointed at the version list, I suggest https://www.microsoft.com/net/download/dotnet-core/2.1. The current link points directly to a specific version that is already out of date.


scottaddie:
@nickwalkmsft Thank you for bringing this to our attention. You're absolutely right that the text wasn't precise enough. I've added the minimum SDK version number, as some readers may not be in a position to install the latest SDK (due to internal controls).

PR #7906, opened 2018-07-31
Add default values to SignalR config doc

anurse:
Had a question about default values for the config values and I figured it would be faster to answer by updating the docs ;).

Not committed to the formatting. I thought a new column could work well but I could also do this as sentences in the Description column (some of the options already had this, I think I moved them all)

[Internal Review Page](https://review.docs.microsoft.com/en-us/aspnet/core/signalr/configuration?view=aspnetcore-2.1&branch=pr-en-us-7906)


scottaddie:
Thanks for adding this, @anurse!

PR #7934, opened 2018-08-03
Remove duplicate paragrahs

dkthedon:


guardrex:
Hello @dkthedon ... Those aren't duplicated paragraphs. They show different content depending on which version of the doc is loaded =2.1, =2.0, <2.0.

PR #7944, opened 2018-08-03
Azure Lifecycle Guide work

CamSoper:
Figuring out formatting

[Internal Review Page](https://review.docs.microsoft.com/en-us/aspnet/core/azure/devops/?view=aspnetcore-2.1&branch=pr-en-us-7944)


CamSoper:
@scottaddie There was a method to my madness in that the PDF link at the bottom of the TOC is per-TOC. My thought was that the PDF for this section would be self-contained that way. Then again, there's a strong argument for keeping the context of the ASP.NET Core docset.

PR #7966, opened 2018-08-06
improve doc and sample for https redirect

### LINKED ISSUE #7811: Primary example should address most common use case http to https port 443

barry-schneider:
Fixes #7811

Updated the sample to reflect the most common option settings. Provided other options in the comment in the code.

Update MD file to make more readable.

### DIFF
--- a/aspnetcore/security/enforcing-ssl.md
+++ b/aspnetcore/security/enforcing-ssl.md
@@ -38,24 +38,10 @@ The following code calls `UseHttpsRedirection` in the `Startup` class:
 
 The following code calls [AddHttpsRedirection](/dotnet/api/microsoft.aspnetcore.builder.httpsredirectionservicesextensions.addhttpsredirection) to configure middleware options:
 
-[!code-csharp[](enforcing-ssl/sample/Startup.cs?name=snippet2&highlight=14-99)]
-
-The preceding highlighted code:
-
-* Sets [HttpsRedirectionOptions.RedirectStatusCode](/dotnet/api/microsoft.aspnetcore.httpspolicy.httpsredirectionoptions.redirectstatuscode).
-* Sets the HTTPS port to 5001.
-
-The following mechanisms set the port automatically:
-
-* The middleware can discover the ports via [IServerAddressesFeature](/dotnet/api/microsoft.aspnetcore.hosting.server.features.iserveraddressesfeature) when the following conditions apply:
-  - Kestrel or HTTP.sys is used directly with HTTPS endpoints (also applies to running the app with Visual Studio Code's debugger).
-  - Only **one HTTPS port** is used by the app.
-* Visual Studio is used:
-  - IIS Express has HTTPS enabled.
-  - *launchSettings.json* sets the `sslPort` for IIS Express.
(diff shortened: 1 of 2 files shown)


Rick-Anderson:
@barry-schneider  It looks like you created the PR from a stale branch. Have you reviewed the **Files changed** tab?

guardrex:
@Rick-Anderson In addition, we need to discuss this further with @shirhatti (and possibly `@Tratcher`) before we proceed. We'll need a more nuanced approach to clarifying the scenarios.

https://github.com/aspnet/Docs/pull/7812#issuecomment-407937125

barry-schneider:
I am going to pull the latest into my repo and redo the pull request with my changes.

guardrex:
@barry-schneider I recommend holding off until we get feedback from the engineers. @shirhatti can comment on the situation with Azure Apps ... for example ... 'Does the Azure switch remove the need to set the port manually?' He said earlier that he prefers we tell devs to use the Azure switch and not set the port explicitly.

Also, the content you're cutting here is accurate in the local environment. We need to find out how the engineers want to advise devs in the non-local environment. I suspect we need to split out the instruction by the hosting environment, which would mean placing the current content into a subsection, not cutting it from the topic.

@Tratcher, would u like to advise us as well?

Tratcher:
Yes the Azure switch would be preferred over using AddHttpsRedirection, that redirect takes place before the request reaches the application. However you still want AddHttpsRedirection for Development and other non-Azure environments.

Setting HttpsPort to 443 is fine. Consider only setting it in Production. The VS dev tools set it for you in Development. If you've already set the Azure switch then this will have no effect.

307 is the default redirect status code and I don't expect users will need to change it. You can remove it from the examples.

I agree removing IServerAddressesFeature is overkill.

barry-schneider:
My main object was to have the documentation reflect the most common use case, which is having http be redirected to https using port 443. As a developer, I followed the current documentation and was not working on my Azure web site. I suggested we start with that example. Then provide additional examples for redirect response and setting the ports, that specifies the order of priority.

What is the .net core convention for the order of priority of applying configuration setting? 

- code
- appsetting
- environment variable
- other

Does [TryGetHttpsPort](https://github.com/aspnet/BasicMiddleware/blob/d8a90ce88fb8c5d26a6604963e1dbc50c12c0dae/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs#L106) function in [HttpsRedirectionMiddleware]
(https://github.com/aspnet/BasicMiddleware/blob/master/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs) confirm to the convention?

guardrex:
That's fine. The content will (probably) end up in a couple of subsections ...

* Azure Apps (no port set; use the portal)
* Local and IIS (set the port to 443 in the Production environment only; leave the existing content for IServerAddressesFeature)

PR #7987, opened 2018-08-08
C32681: Changing position of Italic Marker (asterisk)

v-maudel:
Hello, @scottcate,
"There's an extra asterisk that's breaking the Note code. In case the italics must be conserved, we suggest to start them on line 53."
Many thanks in advance.


scottaddie:
Thank you, @v-maudel!

v-maudel:
Thanks to you @scottaddie

PR #8006, opened 2018-08-09
Small change in documentation to call server method from client

CorneHogerheijde:


scottaddie:
@CorneHogerheijde Thank you for fixing this. Your change will go live later today.

PR #8023, opened 2018-08-10
Paramater naming convention fix

JohnCOsborne:
The parameter should be camelCase.


scottaddie:
@JohnCOsborne Thank you for catching this. That should indeed use camel casing. Your change will go live within the next 24 hours.

PR #8031, opened 2018-08-10
Azure DevOps book cover

CamSoper:


scottaddie:
@CamSoper These changes look good, but why are so many commits shown?

CamSoper:
Because I merged from master before I PR'd. Not sure why that would cause it, but I've seen it do it before.

Sent from Nine<http://www.9folders.com/>

scottaddie:
@CamSoper My recommendation is to remove the Word document. That will quickly become out of sync with the Markdown. Once that's done, this is ready to merge.

PR #8113, opened 2018-08-16
Spell correction

mr5z:


scottaddie:
@mr5z Good catch! Thank you for taking the time to fix it too. Your change will go live later today.

PR #4308, opened 2017-09-16
Continue to update unit testing and debugging MD to use VS 2017

eriawan:
This PR is a part of WIP of task list in #3474 to update some ASP.NET Classic related pages to use Visual Studio 2017, as agreed in proposal #3452 (closed).

This PR contains update of multiple pages but related (under the section of ASP.NET Web API 2, subsection of Testing and Debugging:

- https://github.com/aspnet/Docs/blob/master/aspnet/web-api/overview/testing-and-debugging/unit-testing-with-aspnet-web-api.md
- https://github.com/aspnet/Docs/blob/master/aspnet/web-api/overview/testing-and-debugging/unit-testing-controllers-in-web-api.md
- https://docs.microsoft.com/en-us/aspnet/web-api/overview/testing-and-debugging/mocking-entity-framework-when-unit-testing-aspnet-web-api-2

The updates were done in those pages in this PR, because all three pages are closely related.
The images have had been updated accordingly, especially on the images that show nuget installation and Test Explorer.

The sample code has been updated as well, and it's been uploaded to MSDN code gallery in this link: https://code.msdn.microsoft.com/Unit-Testing-with-ASPNET-1374bc11

Please review, thank you in advance.

PR #8115, opened 2018-08-16
synchronize link to updated sample code that use VS 2017

eriawan:
Synchronize the link to sample code at the top description to be the same link to sample code that use VS 2017.

* Refers to #4308 because looks like the link from the top description isn't the same as the "Download the complete code" link. The top description still points to the old VS 2013 sample code, therefore I update the link to be the same link as I have submitted before 🙂 

cc @scottaddie


scottaddie:
Many thanks for updating this, @eriawan! Your changes will go live later today.

PR #8124, opened 2018-08-16
fix the highlighting for AddUserSecrets

### LINKED ISSUE #8123: code for AddUserSecret snippet - highlight is not correct lines
## General

The highlight on the AddUserSecret is off a bit. This fixes that on https://docs.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-2.1&tabs=windows#access-a-secret

cicorias:
Docs: Code hightlighted in snippet is NOT the code that the section is
referring to. Prior 5-8 is not the AddUserSecret call - line 12 of the snippet is

Fixes #8123


scottaddie:
@cicorias Thank you for fixing that!

PR #8148, opened 2018-08-17
SignalR: Add table to help identify version

anurse:
Adds a section to the ASP.NET Core SignalR Differences document to help identify the version you are using. This can be a helpful resource when customers aren't sure which version is which.

We're also adding a link to this doc (through an `aka.ms` forwarder) to some new version detection logic we're adding to the client (see https://github.com/aspnet/SignalR/pull/2820). I'm not 100% confident in putting a link in an exception message but I think it really needs more explanation than an exception message can provide. We're using a forwarder though so we can re-target it if things move around.

Also cleared up some references to make sure they use the more canonical "ASP.NET Core SignalR" and "ASP.NET SignalR" phrases

[Internal Review Link](https://review.docs.microsoft.com/en-us/aspnet/core/signalr/version-differences?view=aspnetcore-2.1&branch=pr-en-us-8148)

### DIFF
--- a/aspnetcore/signalr/version-differences.md
+++ b/aspnetcore/signalr/version-differences.md
@@ -12,7 +12,7 @@ uid: signalr/version-differences
 
 ASP.NET Core SignalR is not compatible with clients or servers for ASP.NET SignalR. This article details the features which have been removed or changed in the ASP.NET Core SignalR.
 
-## How to identify which version of SignalR you are using:
+## How to identify which version of SignalR you are using
 
 |                      | ASP.NET SignalR | ASP.NET Core SignalR |
 | -------------------- | --------------- | -------------------- |
--- a/aspnetcore/signalr/version-differences.md
+++ b/aspnetcore/signalr/version-differences.md
@@ -53,7 +53,7 @@ app.UseSignalR(routes =>
 
 ### Sticky sessions now required
 
-Because of how scale-out worked in ASP.NET SignalR, clients could reconnect and send messages to any server in the farm. Due to changes to the scale-out model, as well as not supporting reconnects, this is no longer supported. Once the client connects to the server it must interact with the same server for the duration of the connection.
+Because of how scale-out worked in ASP.NET SignalR, clients could reconnect and send messages to any server in the farm. Due to changes to the scale-out model, as well as not supporting reconnects, this is no longer supported. Once the client connects to the server, it must interact with the same server for the duration of the connection.
 
 ### Single hub per connection
 
--- a/aspnetcore/signalr/version-differences.md
+++ b/aspnetcore/signalr/version-differences.md
@@ -8,9 +8,19 @@ ms.date: 06/30/2018
 uid: signalr/version-differences
 ---
 
-# Differences between SignalR and ASP.NET Core SignalR
+# Differences between ASP.NET SignalR and ASP.NET Core SignalR
 
-ASP.NET Core SignalR is not compatible with clients or servers for ASP.NET 4.x SignalR. This article details the features which have been removed or changed in the ASP.NET Core SignalR.
+ASP.NET Core SignalR is not compatible with clients or servers for ASP.NET SignalR. This article details the features which have been removed or changed in the ASP.NET Core SignalR.
+
+## How to identify which version of SignalR you are using:
+
+|                      | ASP.NET SignalR | ASP.NET Core SignalR |
+| -------------------- | --------------- | -------------------- |
+| Server NuGet Package | [Microsoft.AspNet.SignalR](https://www.nuget.org/packages/Microsoft.AspNet.SignalR/) | [Microsoft.AspNetCore.App](https://www.nuget.org/packages/Microsoft.AspNetCore.App/) |
+| Client NuGet Packages | [Microsoft.AspNet.SignalR.Client](https://www.nuget.org/packages/Microsoft.AspNet.SignalR.Client/) [Microsoft.AspNet.SignalR.JS](https://www.nuget.org/packages/Microsoft.AspNet.SignalR.JS/) | [Microsoft.AspNetCore.SignalR.Client](https://www.nuget.org/packages/Microsoft.AspNetCore.SignalR.Client/) |
+| Client NPM Package | [signalr](https://www.npmjs.com/package/signalr) | [@aspnet/signalr](https://www.npmjs.com/package/@aspnet/signalr) |
+| Server App Type | ASP.NET (System.Web) or OWIN Self-Host | ASP.NET Core |
+| Supported Server Platforms | .NET Framework 4.5 | .NET Framework 4.6.1<br/>.NET Core 2.1 |
 
 ## Feature differences
 
@@ -24,12 +34,12 @@ ASP.NET Core SignalR supports JSON, as well as a new binary protocol based on [M
 
 ## Differences on the server
 
-The SignalR server-side libraries are included in the `Microsoft.AspNetCore.App` package that is part of the **ASP.NET Core Web Application** template for both Razor and MVC projects.
+The ASP.NET Core SignalR server-side libraries are included in the `Microsoft.AspNetCore.App` package that is part of the **ASP.NET Core Web Application** template for both Razor and MVC projects.
 
-SignalR is an ASP.NET Core middleware, so it must be configured by calling `AddSignalR` in `Startup.ConfigureServices`.
+ASP.NET Core SignalR is an ASP.NET Core middleware, so it must be configured by calling `AddSignalR` in `Startup.ConfigureServices`.
 
 ```csharp
-services.AddSignalR();
+services.AddSignalR()
 ```
 
 To configure routing, map routes to hubs inside the `UseSignalR` method call in the `Startup.Configure` method.
@@ -43,7 +53,7 @@ app.UseSignalR(routes =>
 
 ### Sticky sessions now required
 
-Because of how scale-out worked in the previous versions of SignalR, clients could reconnect and send messages to any server in the farm. Due to changes to the scale-out model, as well as not supporting reconnects, this is no longer supported. Now, once the client connects to the server it needs to interact with the same server for the duration of the connection.
+Because of how scale-out worked in ASP.NET SignalR, clients could reconnect and send messages to any server in the farm. Due to changes to the scale-out model, as well as not supporting reconnects, this is no longer supported. Now, once the client connects to the server it needs to interact with the same server for the duration of the connection.
 
 ### Single hub per connection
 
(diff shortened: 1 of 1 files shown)


aspnetcore/signalr/version-differences.md
```diff
@@ -8,9 +8,19 @@ ms.date: 06/30/2018
...
+
+## How to identify which version of SignalR you are using:
+
+|                      | ASP.NET SignalR | ASP.NET Core SignalR |
+| -------------------- | --------------- | -------------------- |
+| Server NuGet Package | [Microsoft.AspNet.SignalR](https://www.nuget.org/packages/Microsoft.AspNet.SignalR/) | [Microsoft.AspNetCore.App](https://www.nuget.org/packages/Microsoft.AspNetCore.App/) |
+| Client NuGet Packages | [Microsoft.AspNet.SignalR.Client](https://www.nuget.org/packages/Microsoft.AspNet.SignalR.Client/) [Microsoft.AspNet.SignalR.JS](https://www.nuget.org/packages/Microsoft.AspNet.SignalR.JS/) | [Microsoft.AspNetCore.SignalR.Client](https://www.nuget.org/packages/Microsoft.AspNetCore.SignalR.Client/) |
+| Client NPM Package | [signalr](https://www.npmjs.com/package/signalr) | [@aspnet/signalr](https://www.npmjs.com/package/@aspnet/signalr) |
+| Server App Type | ASP.NET (System.Web) or OWIN Self-Host | ASP.NET Core |
+| Supported Server Platforms | .NET Framework 4.5 | .NET Framework 4.6.1<br/>.NET Core 2.1 |
```
anurse: The `<br/>` here seems a bit sketchy but it works. I'd like to manually break the line here instead of risking weird flow issues.

guardrex: That's fine ... we do it fairly often to control how cells are rendered. However, I recommend `<br>` as we aren't XHTML.

aspnetcore/signalr/version-differences.md
```diff
@@ -8,9 +8,19 @@ ms.date: 06/30/2018
 uid: signalr/version-differences
 ---
 
-# Differences between SignalR and ASP.NET Core SignalR
+# Differences between ASP.NET SignalR and ASP.NET Core SignalR
 
-ASP.NET Core SignalR is not compatible with clients or servers for ASP.NET 4.x SignalR. This article details the features which have been removed or changed in the ASP.NET Core SignalR.
+ASP.NET Core SignalR is not compatible with clients or servers for ASP.NET SignalR. This article details the features which have been removed or changed in the ASP.NET Core SignalR.
+
+## How to identify which version of SignalR you are using
```
guardrex: Recomend shortening ... "How to identify the SignalR version"

aspnetcore/signalr/version-differences.md
```diff
@@ -61,7 +71,7 @@ The ability to pass arbitrary state between clients and the hub (often called Hu
 
 ### TypeScript
 
-The ASP.NET Core version of SignalR is written in [TypeScript](https://www.typescriptlang.org/). You can write in JavaScript or TypeScript when using the [JavaScript client](xref:signalr/javascript-client).
+The ASP.NET Core SignalR is written in [TypeScript](https://www.typescriptlang.org/). You can write in JavaScript or TypeScript when using the [JavaScript client](xref:signalr/javascript-client).
```
BrennanConroy: "The ASP.NET Core SignalR **client**"

aspnetcore/signalr/version-differences.md
```diff
@@ -24,12 +34,12 @@ ASP.NET Core SignalR supports JSON, as well as a new binary protocol based on [M
 
 ## Differences on the server
 
-The SignalR server-side libraries are included in the `Microsoft.AspNetCore.App` package that is part of the **ASP.NET Core Web Application** template for both Razor and MVC projects.
+The ASP.NET Core SignalR server-side libraries are included in the `Microsoft.AspNetCore.App` package that is part of the **ASP.NET Core Web Application** template for both Razor and MVC projects.
```
guardrex: If this is the first mention of the package, link it like this ...

> [Microsoft.AspNetCore.App metapackage](xref:fundamentals/metapackage-app)

Contraction ... "that is" ... "that's"

aspnetcore/signalr/version-differences.md
```diff
@@ -24,12 +34,12 @@ ASP.NET Core SignalR supports JSON, as well as a new binary protocol based on [M
...
 
-SignalR is an ASP.NET Core middleware, so it must be configured by calling `AddSignalR` in `Startup.ConfigureServices`.
+ASP.NET Core SignalR is an ASP.NET Core middleware, so it must be configured by calling `AddSignalR` in `Startup.ConfigureServices`.
```
guardrex: If we have an API broswer link for "AddSignalR," I recommend linking it.

aspnetcore/signalr/version-differences.md
```diff
@@ -43,15 +53,15 @@ app.UseSignalR(routes =>
 
 ### Sticky sessions now required
 
-Because of how scale-out worked in the previous versions of SignalR, clients could reconnect and send messages to any server in the farm. Due to changes to the scale-out model, as well as not supporting reconnects, this is no longer supported. Now, once the client connects to the server it needs to interact with the same server for the duration of the connection.
+Because of how scale-out worked in ASP.NET SignalR, clients could reconnect and send messages to any server in the farm. Due to changes to the scale-out model, as well as not supporting reconnects, this is no longer supported. Now, once the client connects to the server it needs to interact with the same server for the duration of the connection.
```
guardrex: Comma splice (out of position) ... You can use "Now" or not if you prefer (the style manual slightly dissuades it), but structure either of these ways (comma after "server") ...

* Now once the client connects to the server, it needs to interact ...
* Once the client connects to the server, it needs to interact ...

Recommend "requires" over "needs," since "needs" personifies inanimate objects.

aspnetcore/signalr/version-differences.md
guardrex: I think you can drop the article "The" here, too.

aspnetcore/signalr/version-differences.md
anurse: `<you><got><it><buddy>`

aspnetcore/signalr/version-differences.md
```diff
@@ -8,9 +8,19 @@ ms.date: 06/30/2018
 uid: signalr/version-differences
 ---
 
-# Differences between SignalR and ASP.NET Core SignalR
+# Differences between ASP.NET SignalR and ASP.NET Core SignalR
 
-ASP.NET Core SignalR is not compatible with clients or servers for ASP.NET 4.x SignalR. This article details the features which have been removed or changed in the ASP.NET Core SignalR.
+ASP.NET Core SignalR is not compatible with clients or servers for ASP.NET SignalR. This article details the features which have been removed or changed in the ASP.NET Core SignalR.
```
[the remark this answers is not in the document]
anurse: I keep going in to "formal mode" when writing docs and that seems to mean avoiding contractions. I'll get it one of these days :)

aspnetcore/signalr/version-differences.md
anurse: Yep. We didn't at the time this was original authored but I'll add it now.

aspnetcore/signalr/version-differences.md
anurse: "requires" is a bit clunky here, going with "it must interact with the same server...", but feel free to smith that further :)

guardrex: Possibly could just drop it ...

> Once the client connects to the server, it interacts with the same server for the duration of the connection.

guardrex: That looks good ...

> ... it must interact with the same server ...

aspnetcore/signalr/version-differences.md
```diff
@@ -96,21 +106,21 @@ connection.on("ReceiveMessage", (user, message) => {
 });
 ```
 
-After creating the client method, start the hub connection. Chain a `catch` method to log or handle errors.
+After creating the client method, start the hub connection. Chain a [catch](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch) method to log or handle errors.
```
anurse: This is the resource we link to in the JavaScript docs for core types like Promise, it's the Mozilla Developer Network docs. Not sure if we want to add this kind of link, it was just `code fenced` before. I can go back to that, but I like linking to as many API docs as we can :)

aspnetcore/signalr/version-differences.md
```diff
@@ -24,15 +34,15 @@ ASP.NET Core SignalR supports JSON, as well as a new binary protocol based on [M
 
 ## Differences on the server
 
-The SignalR server-side libraries are included in the `Microsoft.AspNetCore.App` package that is part of the **ASP.NET Core Web Application** template for both Razor and MVC projects.
+The ASP.NET Core SignalR server-side libraries are included in the [Microsoft.AspNetCore.App metapackage](xref:fundamentals/metapackage-app) package that's part of the **ASP.NET Core Web Application** template for both Razor and MVC projects.
 
-SignalR is an ASP.NET Core middleware, so it must be configured by calling `AddSignalR` in `Startup.ConfigureServices`.
+ASP.NET Core SignalR is an ASP.NET Core middleware, so it must be configured by calling [AddSignalR](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.signalrdependencyinjectionextensions.addsignalr?view=aspnetcore-2.1) in `Startup.ConfigureServices`.
```
guardrex: Drop the first part of the link ...

```
https://docs.microsoft.com/en-us
```

guardrex: ... and drop the `?view=aspnetcore-2.1` piece.

guardrex:
@anurse Thanks ... looks good.

Scott may have you change the API links to the new format. The benefit of xref API links is validation, but imo it comes at a slight **_grokability cost_**:tm: which makes them a little harder to get right ... until I get used to them anyway. He'll let you know if he wants those in there.

anurse:
FYI API doc links don't seem to work on the review site, but when you remove `review.` from the domain name and the branch query string value they do seem to work :)

anurse:
> Scott may have you change the API links to the new format.

Gotcha, will wait for his review.

guardrex:
> from the domain name and the branch query string value they do seem to work

Yes. At some point early on it was decided to make them all relative.

Now, we'll be going with xref. Personally, I find it painful here at the start of using them to get the `*` correct for the API references that don't link to an overload ... I'm confused about the properties and fields. It's soooooo much simpler to cut-n-paste from the API broswer address in the address bar. I wish we just had a relative link validator in the build system. That's not going to happen tho. It's going to take some guessing/experimenting with xref API links until I understand their quirky rules.

anurse:
> I wish we just had a relative link validator in the build system

Or better yet, something that compiled absolute URLs into the proper xref :)

PR #8177, opened 2018-08-20
Update app-parts.md

behm:
Changed misspelling of the word "behavior"


scottaddie:
@behm Thank you for taking the time to correct this problem. Your change will go live later today.

PR #5948, opened 2018-04-11
SetCompatibilityVersion

Rick-Anderson:
Fixes #5939 
[Review URL](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/startup?view=aspnetcore-2.1&branch=pr-en-us-5948#setcompatibilityversion)

PR #8196, opened 2018-08-20
Move SetCompatibilityVersion into an MVC subtopic

### LINKED ISSUE #8170: SetCompatibilityVersion in the Application startup topic
@Rick-Anderson @scottaddie Why is `SetCompatibilityVersion` for ASP.NET Core MVC in the *Application startup* topic? ... just to surface it? The content seems much better suited for topics on migration, What's new, release notes, and MVC ... *especially MVC*. `SetCompatibilityVersion` doesn't have anything (directly) to do with app startup.

If it's felt that it's critical to surface it here regardless of its direct relevance, what we could do is leave the extension in the example and have a one-liner mention that links out to the [MVC overview](https://docs.microsoft.com/aspnet/core/mvc/overview), where I think this section should appear. We then link to the section in the *MVC overview* from elsewhere (migration, What's new, release notes, app startup).

guardrex:
Fixes #8170 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/mvc/compatibility-version?view=aspnetcore-2.1&branch=pr-en-us-8196)

* Other than an API link, there are no changes to the original text. Just moving the content and sample app here.
* The new topic preserves the original date of the content from #5948.

Please check [20-21.md (Internal Review Topic)](https://review.docs.microsoft.com/en-us/aspnet/core/migration/20_21?view=aspnetcore-1.0&branch=pr-en-us-8196), as I escaped several underscores in an attempt to make GH rendering good.

### DIFF
--- a/aspnetcore/mvc/compatibility-version/samples/2.x/CompatibilityVersionSample/Startup2.cs
+++ b/aspnetcore/mvc/compatibility-version/samples/2.x/CompatibilityVersionSample/Startup2.cs
@@ -5,7 +5,7 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Formatters;
 using Microsoft.Extensions.DependencyInjection;
 
-namespace WebApp1
+namespace CompatibilityVersionSample
 {
     public class Startup2
     {
@@ -13,18 +13,18 @@ namespace WebApp1
         public void ConfigureServices(IServiceCollection services)
         {
             services.AddMvc()
-             // Include the 2.1 behaviors
-             .SetCompatibilityVersion(CompatibilityVersion.Version_2_1)
-             // Except for the following.
-             .AddMvcOptions(options =>
-             {
-               // Don't combine authorize filters (keep 2.0 behavior).
-               options.AllowCombiningAuthorizeFilters = false;
-               // All exceptions thrown by an IInputFormatter will be treated
-               // as model state errors (keep 2.0 behavior).
-               options.InputFormatterExceptionPolicy =
-                              InputFormatterExceptionPolicy.AllExceptions;
-             });
+                // Include the 2.1 behaviors
+                .SetCompatibilityVersion(CompatibilityVersion.Version_2_1)
+                // Except for the following.
+                .AddMvcOptions(options =>
+                {
+                    // Don't combine authorize filters (keep 2.0 behavior).
+                    options.AllowCombiningAuthorizeFilters = false;
+                    // All exceptions thrown by an IInputFormatter are treated
+                    // as model state errors (keep 2.0 behavior).
+                    options.InputFormatterExceptionPolicy =
+                        InputFormatterExceptionPolicy.AllExceptions;
+                });
         }
         #endregion
 
--- a/aspnetcore/fundamentals/startup.md
+++ b/aspnetcore/fundamentals/startup.md
@@ -58,45 +58,6 @@ For features that require substantial setup, there are `Add[Service]` extension 
 
 [!code-csharp[](../common/samples/WebApplication1/Startup.cs?highlight=4,7,11&start=40&end=55)]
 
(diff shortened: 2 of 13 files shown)


aspnetcore/mvc/compatibility-version/samples/2.x/CompatibilityVersionSample/Startup2.cs
```diff
@@ -5,26 +5,26 @@
 using Microsoft.AspNetCore.Mvc.Formatters;
 using Microsoft.Extensions.DependencyInjection;
 
-namespace WebApp1
+namespace CompatibilityVersionSample
```
Rick-Anderson: Next time don't change namespace. I prefer if most samples use `WebApp` - it makes it a bit easier to share code, generate a sample, etc. If the namespace isn't featured - why not standardize on `WebApp` ?
cc @scottaddie  for your thoughts.

guardrex: All of the samples have namespaces that match the template behavior ... i.e., they match whatever folder the sample is generated in. It did occur to me (a few times actually) that we could always use the same namespace. However, the pattern for upgrading these samples major release-to-major release is to copy the prior release sample (with whatever namespace it uses) and then upgrade the copy for the new release.

The main reason tho that I'm using namespaces that match the sample folder and project file name is the template behavior experience. This is what devs will see when they use the templates with `dotnet new`.

I guess it's also fair to say at this point that for consistency sake, we should stick with what I've been doing. Over half of the samps on the repo have been updated on the scheme I'm using, and it would be a :bear: to go back and update all of the samples to conform to a single namespace name.

guardrex: One more note on this ... I feel we should keep all of the following in sync ...

* The sample's folder name.
* The project file (*csproj*) file name.
* The namespace.

... also because when a sample is pulled down, we probably want readers to be able to distinguish samples locally. It would suck if someone had a handful of our samples locally but couldn't, for example, just look at the folder names (or be inside a folder looking at the project file file name) and not be able to immediately tell one sample from another one.

If we're going to have the sample folder names and project file file names be a meaningful sample name, then it stands to reason that the namespace should match. Again, it's the `dotnet new` behavior anyway.

Rick-Anderson:
No, don't go back and change them. I'm saying it's not worth it to change them in the first place. We recommend using the -o  option which allows any namespace.

Rick-Anderson:
It's not the recommended `dotnet new -o` behavior.

guardrex:
For new projects for readers, sure ... that's cool.

However, it's going to lead to problems if we go that way for samples. Some topics require multiple samples, so their folders can't have the same name. Also, readers who pull down samples will have SxS samples with meaningless sample folder names. Of course, it's actually impossible to have SxS samples with the same folder name.

Anywho ... it's your call. I do need to know what your orders are going forward on all three counts:

* Sample folder name (and how to resolve two sample folders in the same version folder)
* Project file file name (csproj)
* Namespace

Rick-Anderson:
Sample folder name should be descriptive. Don't change the defaults on the project name or namespace.

Rick-Anderson:
I'd rather you spent time working on issues that make a difference than changing folder names. Such work comes across as patting your numbers for no real improvement.

Rick-Anderson:
@guardrex  no more moving topics when the TOC changes without my prior approval. Makes it difficult to track prior commits.

Rick-Anderson:
Never mind, looks like this was taken out of a topic and made into it's own topic.

guardrex:
Yes, that's right.

The only thing we did in that vein was fix up the Blazor stuff. Best because sticking with "Blazor client-side" and "Blazor server-side" would have been very bad in future years here.

PR #8364, opened 2018-08-31
Fix link to metapackage docs

vuchl:
The line break in the markdown seems to break the link to further docs


scottaddie:
Thank you so much for fixing this, @vuchl! This change will go live later today.

PR #8376, opened 2018-08-31
the default model class is `ErrorViewModel`, not `ErrorModel`

johnmwright:
You can see the correct name is already used the the code sample two sentences down

### DIFF
--- a/aspnetcore/fundamentals/error-handling.md
+++ b/aspnetcore/fundamentals/error-handling.md
@@ -60,7 +60,7 @@ Configure an exception handler page to use when the app isn't running in the `De
 
 [!code-csharp[](error-handling/samples/2.x/ErrorHandlingSample/Startup.cs?name=snippet_DevExceptionPage&highlight=11)]
 
-In a Razor Pages app, the [dotnet new](/dotnet/core/tools/dotnet-new) Razor Pages template provides an Error page and `ErrorModel` page model class in the *Pages* folder.
+In a Razor Pages app, the [dotnet new](/dotnet/core/tools/dotnet-new) Razor Pages template provides an Error page and `ErrorViewModel` page model class in the *Pages* folder.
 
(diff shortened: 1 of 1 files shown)


guardrex:
Hello @johnmwright ... This passage refers to a Razor Pages app. It refers to the template code of the `ErrorModel` ...

```csharp
namespace Sample.Pages
{
    public class ErrorModel : PageModel
    {
        public string RequestId { get; set; }
        
        public bool ShowRequestId => !string.IsNullOrEmpty(RequestId);

        [ResponseCache(Duration = 0, Location = ResponseCacheLocation.None, NoStore = true)]
        public void OnGet()
        {
            RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier;
        }
    }
}
```

The example below the next paragraph refers to an MVC app.

guardrex:
btw - It wouldn't be terrible to have that code on display in the text. However, the engineers change the content of the `ErrorModel`. They changed it between 2.0 and 2.1. If we were to include the code, it would require versioning in the topic, and it would be fragile going forward. We'd likely miss a change at some point in the future and end up with a stale code example when the engineers make additional changes.

guardrex:
@Rick-Anderson Do u think we should maintain example code for the `ErrorModel`? I'm not really feel'in it based on the template churn. This page model was changed between 2.0 and 2.1 ... and it's [changing again for 2.2 (click to see the 2.2 version)](https://github.com/aspnet/Templating/blob/master/src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Pages/Error.cshtml.cs). We might be fighting a losing battle if we go down this road. ⚠️

Rick-Anderson:
See my commit, that should solve the problem.

Rick-Anderson:
@scottaddie  another WIP bug. No WIP in commit title or label.

Rick-Anderson:
No WIP link either
image

PR #8388, opened 2018-09-02
IP safelist article

tdykstra:
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/security/ip-safelist?view=aspnetcore-1.0&branch=pr-en-us-8388) right now must be run from an incognito window.

### DIFF
--- a/aspnetcore/security/ip-safelist.md
+++ b/aspnetcore/security/ip-safelist.md
@@ -1,0 +1,174 @@
+---
+title: Remote IP HTTP request safe-list in ASP.NET Core
+author: damienbod
+description: Learn how to write Middleware to validate remote IP addresses
+ms.author: tdykstra
+ms.custom: mvc
+ms.date: 01/31/2018
+uid: security/ip-safelist
+---
+# IP-Safelist Middleware in ASP.NET Core
+
+By [Damien Bowden](https://twitter.com/damien_bod) and [Tom Dykstra](https://github.com/tdykstra)
+ 
+This help article shows how a client safe-list could be implemented using ASP.NET Core middleware checking the Remote IP address of the request. If the client IP is on the safe-list, no restrictions exist.
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/security/ip-safelist/samples/2.x/ClientIpAspNetCore) ([how to download](xref:tutorials/index#how-to-download-a-sample))
+
+The sample illustrates middleware which validtes the remote IPof every client request.
+
+* ASP.NET Core 2.x Middleware which checks the remote IP (IP4, IP6) and validates this uses a safe-list of IPs from the app.settings.
+
+
+The middleware uses an admin white-list parameter from the constructor to compare with the remote IP address from the HttpContext Connection property. This is different to previous versions of .NET. In the example, all GET requests are allowed. If any other request method is used, the remote IP is used to check if it exists in the safe-list. If it does not exist, a 403 is returned.
+
+```csharp
+using System;
+using System.Linq;
+using System.Net;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+
+namespace ClientIpAspNetCore
+{
+    public class AdminSafeListMiddleware
+    {
+        private readonly RequestDelegate _next;
+        private readonly ILogger<AdminSafeListMiddleware> _logger;
+        private readonly string _adminSafeList;
+
+        public AdminSafeListMiddleware(RequestDelegate next, ILogger<AdminSafeListMiddleware> logger, string adminWhiteList)
+        {
+            _adminSafeList = adminWhiteList;
+            _next = next;
+            _logger = logger;
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            if (context.Request.Method != "GET")
+            {
+                var remoteIp = context.Connection.RemoteIpAddress;
+                _logger.LogDebug($"Request from Remote IP address: {remoteIp}");
+
+                string[] ip = _adminSafeList.Split(';');
+
+                var bytes = remoteIp.GetAddressBytes();
+                var badIp = true;
+                foreach (var address in ip)
+                {
+                    var testIp = IPAddress.Parse(address);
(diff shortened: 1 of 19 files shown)


damienbod:
@tdykstra Thanks, looks good

aspnetcore/security/ip-safelist.md
```diff
@@ -0,0 +1,59 @@
...
+ms.author: tdykstra
+ms.custom: mvc
+ms.date: 08/31/2018
+uid: security/ip-safelist
+---
+# Client IP safelist for ASP.NET Core
+
+By [Damien Bowden](https://twitter.com/damien_bod) and [Tom Dykstra](https://github.com/tdykstra)
+ 
+This article shows two ways to implement an IP safelist (also known as a whitelist):
```
Rick-Anderson: as an IP whitelist - when I first read this I thought you were saying  "IP safelist" AKA whitelist - which it's not.

Rick-Anderson: I'd move "By using ASP.NET Core" to this line. That makes the two bullets really stand out - middleware and filter

aspnetcore/security/ip-safelist.md
```diff
@@ -0,0 +1,59 @@
...
+
+* By using ASP.NET Core middleware to check the remote IP address of every request.
+* By using ASP.NET Core action filters to check the remote IP address of requests for specific action methods.
+
+The sample app illustrates both approaches. In each case, a string containing approved client IP addresses is stored in an app setting. The middleware or filter parses the string into a list and  checks if the remote IP is in the list. If not, an HTTP 403 Forbidden status code is returned.
```
Rick-Anderson: >stored in an app setting.

That's an implementation detail they don't need to know about now. A real app would fetch it from a DB or a Web API call. It sounds clumsy removing tho so maybe leave it.

Use VSC DocFX addin to find/remove double spaces. 

an HTTP 403 (Forbidden) status code is returned.

aspnetcore/security/ip-safelist.md
```diff
@@ -0,0 +1,59 @@
...
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/security/ip-safelist/samples/2.x/ClientIpAspNetCore) ([how to download](xref:tutorials/index#how-to-download-a-sample))
+
+## The safelist
+
+The list is configured in the *appsettings.json* file. It's a semicolon-delimited list and can contain IPv4 and IPv6 addresses.
```
Rick-Anderson: Consider/ Maybe remove from app setting deatail above and write this as
The list is contained in the *app.json* file, a real app would use a database or other data store. 
Or leave as is.

aspnetcore/security/ip-safelist/samples/2.x/ClientIpAspNetCore/AdminSafeListMiddleware.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace ClientIpAspNetCore
+{
+    #region snippet_ClassOnly
+    public class AdminSafeListMiddleware
+    {
+        private readonly RequestDelegate _next;
+        private readonly ILogger<AdminSafeListMiddleware> _logger;
+        private readonly string _adminSafeList;
+
+        public AdminSafeListMiddleware(RequestDelegate next, ILogger<AdminSafeListMiddleware> logger, string adminSafeList)
```
Rick-Anderson: Loose horiz scroll bars
image

aspnetcore/security/ip-safelist.md
```diff
@@ -0,0 +1,59 @@
...
+
+The action filter is added to the services container.
+
+[!code-csharp[](ip-safelist/samples/2.x/ClientIpAspNetCore/Startup.cs?name=snippet_ConfigureServices&highlight=3)]
+
+The filter can then be used on a controller or action method.
+
+[!code-csharp[](ip-safelist/samples/2.x/ClientIpAspNetCore/Controllers/ValuesController.cs?name=snippet_Filter&highlight=1)]
+
+In the sample app, the filter is applied to the Get method. So when you test the app by sending a Get request, the attribute is validating the client IP address. When you test by sending any other HTTP method, the middleware is validating the client IP.
```
Rick-Anderson: Why check on GET with this approach? Can you put a form on the page and test on the POST?

Rick-Anderson:
@damienbod can you show how to use [RP filters](https://docs.microsoft.com/en-us/aspnet/core/razor-pages/filter?view=aspnetcore-2.1#implement-a-filter-attribute) for a RP?

aspnetcore/security/ip-safelist/samples/2.x/ClientIpAspNetCore/AdminSafeListMiddleware.cs
scottaddie: Each line of code should be no longer than 85 characters.

tdykstra: @damienbod Let me know if you want to do the breaking up of long code lines or you'd like me to do it.  The length restriction is intended to limit the need for horizontal scroll bars and applies only to code that gets displayed in the article.  Here's the current guidance on this:

>Horizontal scroll bars
>Break up long lines to avoid horizontal scroll bars. Scroll bars on code blocks make code hard to read. They're especially problematic on longer code blocks, where it may be impossible to see the scroll bar and the line you want to read at the same time.

>A good practice for minimizing horizontal scroll bars on code blocks is to break up code lines longer than 85 characters. But keep in mind that the presence or absence of a scroll bar is not the only criterion of readability. If breaking a line before 85 hurts readability in other ways, feel free to go over 85.

Rick-Anderson: @tdykstra  can you fix that?

tdykstra: @damienbod I'll go ahead and do the line breaks.

PR #8422, opened 2018-09-05
App startup topic updates

### LINKED ISSUE #8407: Update Startup Filter section to demo adding it from outside of the Startup class
I found the inclusion of IStartupFilter in this section confusing as there did not seem to be a valuable use case.  See my SO question https://stackoverflow.com/questions/51437607/what-is-the-point-of-istartupfilter-in-asp-net-core/52178476#52178476 .  I think that it is better covered in the section "enhance an app from an external assembly in ASP.NET Core with IHostingStartup" or the doc should make it clear that external assemblies are the main use case.

---
#### Document Details

* Content: [Application startup in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/startup?view=aspnetcore-2.1#feedback)
* Content Source: [aspnetcore/fundamentals/startup.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/startup.md)
* Product: **aspnet-core**
* GitHub Login: @ardalis

guardrex:
Fixes #8407 

* The update of the 2.0 sample here is tracked by #5495.
* Placed the snippet region just around the `CreateDefaultBuilder` call so that the 2.0-era `BuildWebHost` call isn't seen in the topic.

### DIFF
--- a/aspnetcore/fundamentals/startup.md
+++ b/aspnetcore/fundamentals/startup.md
@@ -80,7 +80,7 @@ For more information on how to use `IApplicationBuilder` and the order of middle
 
 [!code-csharp[](startup/snapshot_sample/Program.cs?highlight=18,22)]
 
-## Startup filters
+## Extending Startup with startup filters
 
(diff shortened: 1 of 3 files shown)


aspnetcore/fundamentals/startup.md
```diff
@@ -80,7 +80,7 @@ For more information on how to use `IApplicationBuilder` and the order of middle
 
 [!code-csharp[](startup/snapshot_sample/Program.cs?highlight=18,22)]
 
-## Startup filters
+## Extending Startup with startup filters
```
scottaddie: Extending --> Extend

Issue #8417, opened 2018-09-05
KnownProxies/Networks enhancements


guardrex:
Following discussion on #2384.

Thus far ...

> Note that's an IPv4 address nested in an IPv6 address. Your KnownProxies or KnownNetworks would need to be represented as IPv6 addresses as well.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e548cd16-b06c-5802-a1a7-c627167857ae
* Version Independent ID: c48abb08-bea3-dcff-c067-2edf5dc9b3a6
* Content: [Configure ASP.NET Core to work with proxy servers and load balancers](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-2.1)
* Content Source: [aspnetcore/host-and-deploy/proxy-load-balancer.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/host-and-deploy/proxy-load-balancer.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

PR #8449, opened 2018-09-07
Update IP address format with note

guardrex:
Addresses #8417


guardrex:
@scottaddie I'm on the fence about the &dagger; ... Should I inline this text between the bullet list and the code example?

If we want to de-emphasize the points, let's keep them in footnote format. If we want to emphasize those points, then I'd say let's move those two sentences just above the code example.

scottaddie:
@guardrex Please move the points just above the code example and de-dagger them.

PR #8450, opened 2018-09-07
2018-09-07 push to live 

CamSoper:
Need to get the new cover for the DevOps book out.  @guardrex, looks like you're along for the ride. 😄


guardrex:
Put the top *DOWN*! 🚗 :smile:

PR #8472, opened 2018-09-09
Update dotnet-client.md

Rick-Anderson:
remove double send
Remove `you` when it doesn't add clarity. You aren't sending/receiving, the client is. It's more clear without you.

### DIFF
--- a/aspnetcore/signalr/dotnet-client.md
+++ b/aspnetcore/signalr/dotnet-client.md
@@ -13,7 +13,7 @@ uid: signalr/dotnet-client
 
 By [Rachel Appel](http://twitter.com/rachelappel)
 
-The ASP.NET Core SignalR .NET client can be used by Xamarin, WPF, Windows Forms, Console, and .NET Core apps. Like the [JavaScript client](xref:signalr/javascript-client), the .NET client can send and receive messages to a hub in real time.
+The ASP.NET Core SignalR .NET client library lets you communicate with SignalR hubs from .NET apps.
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


aspnetcore/signalr/dotnet-client.md
```diff
@@ -13,7 +13,7 @@ uid: signalr/dotnet-client
 
 By [Rachel Appel](http://twitter.com/rachelappel)
 
-The ASP.NET Core SignalR .NET client can be used by Xamarin, WPF, Windows Forms, Console, and .NET Core apps. Like the [JavaScript client](xref:signalr/javascript-client), the .NET client enables you to receive and send and receive messages to a hub in real time.
+The ASP.NET Core SignalR .NET client can be used by Xamarin, WPF, Windows Forms, Console, and .NET Core apps. Like the [JavaScript client](xref:signalr/javascript-client), the .NET client can send and receive messages to a hub in real time.
```
tdykstra: Your change is fine, but take a look at the fist sentence -- it's a mix of development platforms, development frameworks, app types, and target frameworks.  Can we just say something like ".NET apps that target .NET Core or .NET Framework 4.6.1 or later"? (Replace 4.6.1 by whatever version is the earliest that works with the .NET client?).  

The second sentence is improved by your change, but I wonder how helpful it is.  Sending and receiving real-time messages is what SignalR is all about, so it seems redundant. And why focus on JS client for comparison when there is also a Java client?

Could we reduce the intro to 

> The ASP.NET Core SignalR .NET client library lets you communicate with SignalR hubs from .NET apps.

Rick-Anderson: Good idea. Can you make that change and I'll review.

Rick-Anderson: Less is more.

PR #8475, opened 2018-09-10
Fixed code sample

davidfowl:
Fixed code to retrieve the IHubContext from middleware


scottaddie:
Thank you for fixing this, @davidfowl!

PR #8486, opened 2018-09-10
Add Context section to signalr/hubs.md

### LINKED ISSUE #7629: Include section about replicating old asp.net signal QueryString functionality
Adding parameters at the end of the `connectionBuilder` `.withUrl("/transportHub?param=foobar")` like this, and accessing it in the Hub `OnConnectedAsync` override like this `Context.GetHttpContext().Request.Query["foobar"]`.

---
#### Document Details

* ID: de8cdc99-48bb-3266-05bb-eda98f59f2e2
* Version Independent ID: 66d252c3-6300-7e28-9aa0-ef2ff66e7a76
* Content: [ASP.NET Core SignalR configuration](https://docs.microsoft.com/en-us/aspnet/core/signalr/configuration?view=aspnetcore-2.1)
* Content Source: [aspnetcore/signalr/configuration.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/signalr/configuration.md)
* Product: **aspnet-core**
* GitHub Login: @rachelappel
* Microsoft Alias: **rachelap**

tdykstra:
Fixes #7629

@BrennanConroy -  added Context object section similar to Clients object section as suggested.

A few questions:
* How do things get into the `Items` collection?
* Any additional useful information we should add to the property and method descriptions?

[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/signalr/hubs?view=aspnetcore-2.1&branch=pr-en-us-8486)

### DIFF
--- a/aspnetcore/signalr/hubs.md
+++ b/aspnetcore/signalr/hubs.md
@@ -43,10 +43,10 @@ Each instance of the `Hub` class has a property named `Context` that contains th
 
 | Property | Description |
 | ------ | ----------- |
-| `ConnectionId` | Gets the unique ID for the connection, assigned by SignalR. There is one connection ID for each connection, and the same connection ID is used by all hubs if there are multiple hubs in an application.|
+| `ConnectionId` | Gets the unique ID for the connection, assigned by SignalR. There is one connection ID for each connection.|
 | `UserIdentifier` | Gets the user identifier. By default, SignalR uses the `ClaimTypes.NameIdentifier` from the `ClaimsPrincipal` associated with the connection as the user identifier. |
 | `User` | Gets the `ClaimsPrincipal`. |
-| `Items` | Gets a key/value collection that can be used to share data within the scope of this connection. |
+| `Items` | Gets a key/value collection that can be used to share data within the scope of this connection. You can store items in this collection and they will persist for that connection across different hub method invocations. |
 | `Features` | Gets the collection of [HTTP features](xref:fundamentals/request-features) available on the connection. |
 | `ConnectionAborted` | Gets a `CancellationToken` that notifies when the connection is aborted. |
 
@@ -54,7 +54,7 @@ Each instance of the `Hub` class has a property named `Context` that contains th
 
 | Method | Description |
 | ------ | ----------- |
-| `GetHttpContext` | Returns the `HttpContext` for the connection, or `null` if the connection is not associated with an HTTP request. For HTTP connections, you can use this method to get information such as HTTP header and query string data. |
+| `GetHttpContext` | Returns the `HttpContext` for the connection, or `null` if the connection is not associated with an HTTP request. For HTTP connections, you can use this method to get information such as HTTP headers and query strings. |
 | `Abort` | Aborts the connection. |
 
 ## The Clients object
--- a/aspnetcore/signalr/hubs.md
+++ b/aspnetcore/signalr/hubs.md
@@ -44,8 +44,8 @@ Each instance of the `Hub` class has a property named `Context` that contains th
 | Property | Description |
 | ------ | ----------- |
 | `ConnectionId` | Gets the unique ID for the connection, assigned by SignalR. There is one connection ID for each connection, and the same connection ID is used by all hubs if there are multiple hubs in an application.|
-| `UserIdentifier` | Gets the user identifier. |
-| `User` | Gets the user (an instance of `ClaimsPrincipal`). |
+| `UserIdentifier` | Gets the user identifier. By default, SignalR uses the `ClaimTypes.NameIdentifier` from the `ClaimsPrincipal` associated with the connection as the user identifier. |
+| `User` | Gets the `ClaimsPrincipal`. |
 | `Items` | Gets a key/value collection that can be used to share data within the scope of this connection. |
 | `Features` | Gets the collection of [HTTP features](xref:fundamentals/request-features) available on the connection. |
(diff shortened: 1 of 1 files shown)


aspnetcore/signalr/hubs.md
```diff
@@ -37,9 +37,29 @@ Create a hub by declaring a class that inherits from `Hub`, and add public metho
 
 You can specify a return type and parameters, including complex types and arrays, as you would in any C# method. SignalR handles the serialization and deserialization of complex objects and arrays in your parameters and return values.
 
+## The Context object
+
+Each instance of the `Hub` class has a property named `Context` that contains the following properties with information about the connection:
+
+| Property | Description |
+| ------ | ----------- |
+| `ConnectionId` | Gets the unique ID for the connection, assigned by SignalR. There is one connection ID for each connection, and the same connection ID is used by all hubs if there are multiple hubs in an application.|
```
BrennanConroy: ~~"and the same connection ID is used by all hubs if there are multiple hubs in an application."~~ Single hub per connection.

aspnetcore/signalr/hubs.md
```diff
@@ -37,9 +37,29 @@ Create a hub by declaring a class that inherits from `Hub`, and add public metho
...
+| `User` | Gets the `ClaimsPrincipal`. |
+| `Items` | Gets a key/value collection that can be used to share data within the scope of this connection. |
+| `Features` | Gets the collection of [HTTP features](xref:fundamentals/request-features) available on the connection. |
+| `ConnectionAborted` | Gets a `CancellationToken` that notifies when the connection is aborted. |
+
+`Hub.Context` also contains the following methods:
+
+| Method | Description |
+| ------ | ----------- |
+| `GetHttpContext` | Returns the `HttpContext` for the connection, or `null` if the connection is not associated with an HTTP request. For HTTP connections, you can use this method to get information such as HTTP header and query string data. |
```
BrennanConroy: "HTTP header**s**"

aspnetcore/signalr/hubs.md
```diff
@@ -37,9 +37,29 @@ Create a hub by declaring a class that inherits from `Hub`, and add public metho
 
 You can specify a return type and parameters, including complex types and arrays, as you would in any C# method. SignalR handles the serialization and deserialization of complex objects and arrays in your parameters and return values.
 
+## The Context object
+
+Each instance of the `Hub` class has a property named `Context` that contains the following properties with information about the connection:
```
Rick-Anderson: I prefer
Each instance of the `Hub` class has a `Context` property with the following properties which contain information about the connection:

I wonder if we even need each instance.

`Hub.Context` has the following properties which contain information about the connection:

or ~which~ containing information ..

aspnetcore/signalr/hubs.md
```diff
@@ -37,9 +37,29 @@ Create a hub by declaring a class that inherits from `Hub`, and add public metho
...
+
+| Property | Description |
+| ------ | ----------- |
+| `ConnectionId` | Gets the unique ID for the connection, assigned by SignalR. There is one connection ID for each connection.|
+| `UserIdentifier` | Gets the user identifier. By default, SignalR uses the `ClaimTypes.NameIdentifier` from the `ClaimsPrincipal` associated with the connection as the user identifier. |
+| `User` | Gets the `ClaimsPrincipal`. |
+| `Items` | Gets a key/value collection that can be used to share data within the scope of this connection. You can store items in this collection and they will persist for that connection across different hub method invocations. |
```
Rick-Anderson: Prefer to remove "you" unless it aids clarity.

Items can be stored in this ...

aspnetcore/signalr/hubs.md
```diff
@@ -37,9 +37,29 @@ Create a hub by declaring a class that inherits from `Hub`, and add public metho
...
+| `Features` | Gets the collection of [HTTP features](xref:fundamentals/request-features) available on the connection. |
```
anurse: Actually this isn't quite correct. It doesn't get the collection of HTTP features, it gets a collection of "connection features" available as part of the SignalR connection. SignalR uses the same basic abstraction but the content of this collection is *different* from `HttpContext.Features`. This is a very advanced feature that I wouldn't expect customers to need to use though.

anurse: I don't have a great idea on how to document this without going in to a whole new level of detail :). Honestly I just want to tell users "don't worry about this one for now" ;P. At some point, as we expand SignalR and move in to non-HTTP spaces, this might become more relevant but for now there's very little a user can do here that they can't already do elsewhere.

1kevgriff:
>Each property or method in the preceding tables returns an object with a `SendAsync` method. The `SendAsync` method allows you to supply the name and parameters of the client method to call.

Would this be a good place to talk about strongly typed hubs as well, or would it be better suited for a different area)?  For example:

```csharp
    public class ChatHub : Hub<IMethods>
    {
        public async Task Message()
        {
            await Clients.All.MethodA();
        }
    }

    public interface IMethods
    {
        Task MethodA();
    }
```

tdykstra:
@1kevgriff This seems like a good place for that, but it's different enough from this issue and PR that I would put it in a new issue and PR.

anurse:
We're actually thinking about trying to encourage the use of `Hub<T>` by default, so it would be awesome if we could get a section in here about it!

1kevgriff:
@anurse I discovered it during a code review with someone else.  I'm definitely happy to write the section on it.

tdykstra:
@1kevgriff I created an issue for this #8505 and you're welcome to write the section!

PR #8497, opened 2018-09-11
WIP: Add section for Closed event in SignalR .NET Client doc

### LINKED ISSUE #8457: Why does the HubConnection.Closed event expect a delegate that returns a Task?
It's not clear why the Closed event (which I was able to use to diagnose a protocol mismatch during connection.StartAsync) requires a delegate that returns a Task, instead of returning void. Am I expected to provide a potential delay before a reconnect occurs, or is this meant to be used for something else? Some documentation on that would be great. :)

---
#### Document Details

* ID: c9bc1072-be2c-94f6-49bf-daa9e97dff2d
* Version Independent ID: 36052173-7061-6d01-8aed-915685cd7cf6
* Content: [ASP.NET Core SignalR .NET Client](https://docs.microsoft.com/en-us/aspnet/core/signalr/dotnet-client?view=aspnetcore-2.1#error-handling-and-logging)
* Content Source: [aspnetcore/signalr/dotnet-client.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/signalr/dotnet-client.md)
* Product: **aspnet-core**
* GitHub Login: @tdykstra

tdykstra:
Fixes #8457

### DIFF
--- a/aspnetcore/signalr/dotnet-client.md
+++ b/aspnetcore/signalr/dotnet-client.md
@@ -36,6 +36,20 @@ To establish a connection, create a `HubConnectionBuilder` and call `Build`. The
 
 [!code-csharp[Build hub connection](dotnet-client/sample/signalrchatclient/MainWindow.xaml.cs?name=snippet_MainWindowClass&highlight=14-16,32)]
 
+### Handle lost connection
+
+You can use the `Closed` event to respond to a lost connection. For example, you might want to automate reconnection. The `Closed` event requires a delegate that returns a `Task`. The reason for that return type is to allow you to run async code without having to use `async void`. If your `Closed` handler doesn't use `await`, you can return `Task.CompletedTask` to satisfy the delegate signature, as in the following example:
+
+```csharp
+connection.Closed += (error) => {
+    // Do your close logic
+    return Task.CompletedTask;
+}
+
+The main reason for the async support is so you can restart the connection. Starting a connection is an async action, as shown in the following example:
+
+[!code-csharp[Use Closed event handler to automate reconnection](dotnet-client/sample/signalrchatclient/MainWindow.xaml.cs?name=snippet_ClosedRestart)]
+
 ## Call hub methods from client
 
 `InvokeAsync` calls methods on the hub. Pass the hub method name and any arguments defined in the hub method to `InvokeAsync`. SignalR is asynchronous, so use `async` and `await` when making the calls.
--- a/aspnetcore/signalr/dotnet-client/sample/SignalRChatClient/MainWindow.xaml.cs
+++ b/aspnetcore/signalr/dotnet-client/sample/SignalRChatClient/MainWindow.xaml.cs
@@ -1,5 +1,6 @@
 ﻿#region snippet_MainWindowClass
 using System;
+using System.Threading.Tasks;
 using System.Windows;
 using Microsoft.AspNetCore.SignalR.Client;
 
@@ -14,7 +15,18 @@ namespace SignalRChatClient
 
             connection = new HubConnectionBuilder()
                 .WithUrl("http://localhost:53353/ChatHub")
-                .Build();        
+                .Build();
+
+            #region snippet_ClosedRestart
+            connection.Closed += async (error) =>
+            {
+                // Consider waiting for some random delay 
+                // to prevent overloading the server, in case the
+                // connection was lost because the server went down.
+                await Task.Delay(3000);
+                await connection.StartAsync();
+            };
+            #endregion
         }
 
         private async void connectButton_Click(object sender, RoutedEventArgs e)


aspnetcore/signalr/dotnet-client.md
```diff
@@ -34,7 +32,24 @@ Install-Package Microsoft.AspNetCore.SignalR.Client
 
 To establish a connection, create a `HubConnectionBuilder` and call `Build`. The hub URL, protocol, transport type, log level, headers, and other options can be configured while building a connection. Configure any required options by inserting any of the `HubConnectionBuilder` methods into `Build`. Start the connection with `StartAsync`.
 
-[!code-csharp[Build hub connection](dotnet-client/sample/signalrchatclient/MainWindow.xaml.cs?name=snippet_MainWindowClass&highlight=14-16,32)]
+[!code-csharp[Build hub connection](dotnet-client/sample/signalrchatclient/MainWindow.xaml.cs?name=snippet_MainWindowClass&highlight=15-17,42)]
+
+### Handle lost connection
+
+You can use the `Closed` event to respond to a lost connection. For example, you might want to automate reconnection.
```
scottaddie: ~~You can use the~~
Use the

Link to the API ref page for the `Closed` event. You can use the following syntax:

```
<xref:Microsoft.AspNetCore.SignalR.Client.HubConnection.Closed>
```

aspnetcore/signalr/dotnet-client.md
```diff
@@ -34,7 +32,24 @@ Install-Package Microsoft.AspNetCore.SignalR.Client
...
+
+The `Closed` event requires a delegate that returns a `Task`. The reason for that return type is to allow you to run async code without having to use `async void`. If your `Closed` handler doesn't use `await`, you can return `Task.CompletedTask` to satisfy the delegate signature, as in the following example:
```
scottaddie: Use the words "you" and "your" sparingly. Here's a suggested rework on the last 2 sentences:

~~The reason for that return type is to allow you to run async code without having to use `async void`. If your `Closed` handler doesn't use `await`, you can return `Task.CompletedTask` to satisfy the delegate signature, as in the following example:~~
The `Task` return type allows async code to run without using `async void`. If the `Closed` handler doesn't use `await`, return `Task.CompletedTask` to satisfy the delegate signature. For example:

aspnetcore/signalr/dotnet-client.md
```diff
@@ -34,7 +32,24 @@ Install-Package Microsoft.AspNetCore.SignalR.Client
 
 To establish a connection, create a `HubConnectionBuilder` and call `Build`. The hub URL, protocol, transport type, log level, headers, and other options can be configured while building a connection. Configure any required options by inserting any of the `HubConnectionBuilder` methods into `Build`. Start the connection with `StartAsync`.
 
-[!code-csharp[Build hub connection](dotnet-client/sample/signalrchatclient/MainWindow.xaml.cs?name=snippet_MainWindowClass&highlight=14-16,32)]
+[!code-csharp[Build hub connection](dotnet-client/sample/signalrchatclient/MainWindow.xaml.cs?name=snippet_MainWindowClass&highlight=15-17,42)]
+
+### Handle lost connection
```
scottaddie: Would this content be more easily discovered if the heading is promoted to an H2?

aspnetcore/signalr/dotnet-client/sample/SignalRChatClient/MainWindow.xaml.cs
```diff
@@ -14,7 +15,18 @@ public MainWindow()
 
             connection = new HubConnectionBuilder()
                 .WithUrl("http://localhost:53353/ChatHub")
-                .Build();        
+                .Build();
+
+            #region snippet_ClosedRestart
+            connection.Closed += async (error) =>
+            {
+                // Consider waiting for some random delay
```
scottaddie: Consider moving this comment into the doc itself for 2 reasons:
1. It will then be localized.
1. Many readers ignore comments in the code.

aspnetcore/signalr/dotnet-client/sample/SignalRChatClient/MainWindow.xaml.cs
```diff
@@ -14,7 +15,15 @@ public MainWindow()
 
             connection = new HubConnectionBuilder()
                 .WithUrl("http://localhost:53353/ChatHub")
-                .Build();        
+                .Build();
+
+            #region snippet_ClosedRestart
+            connection.Closed += async (error) =>
+            {
+                await Task.Delay(3000);
```
anurse: To actually make this "work", we should consider having the same do a truely random delay (i.e. `random.Next(0, 5) * 1000` or something). The purpose of a random delay is to ensure that when the server returns, it doesn't get a barrage of clients reconnecting, so each client needs to pick a **different** delay.

PR #8523, opened 2018-09-13
RP conventions topic+sample updates

### LINKED ISSUE #6173: Update razor-pages-convention-features.md
Three bugs:
1)Navigate to the sample app localhost:5000/TheContactPage
The index page is loaded instead of localhost:5000/Contact when the optional route is left off with the TheContact. I stepped through the call stack starting from the loading of RazorPages from the pages directory and haven't found the point where the index is loaded yet so I don't know if this is a doc or core bug.

2)Navigate to the sample app localhost:5000/TheContactPage/TextValue
Mouse over the links in the top and you will see that every link has been replaced with the TheContactPage appended to every link.

A confusing point in the article: The notes all indicate that the order property determines which route comes first {globalTemplate?} vs {aboutTemplate?}. This wording makes it seem like the order of directories /GlobalRouteValue/OtherPagesRouteValue was determined by this number but stepping through the sourcelink provided .NET CORE source code shows that the order of the route order is determined by the order that options.Conventions.Add are called in the startup. This is because the model.Selectors.Add in the sample adds a new selector attribute route model template on top of the one that already exists. So each route added creates a route plus the old ie 1,2, 1/2, 3, 1/3, 1/2/3, 1/4, 2/4, etc... The example happened to create a conflict by appending templates rather than adding a new template and the selection of the template when one value is provided then gets resolved between {globalTemplate?} vs {aboutTemplate?} using the order. I think you could make this clear just by adding one line at the top note saying the order of routes is determined by the template. You can see an example by switching the call order in startup between snippets 1 and 3 options.Conventions.AddFolderRouteModelConvention("/OtherPages" and options.Conventions.Add(new GlobalTemplatePageRouteModelConvention() and then navigating to localhost:5000/OtherPages/Page1/GlobalRouteValue/OtherPagesRouteValue and observe that the html now displays
"Route data for 'globalTemplate' was provided: OtherPagesRouteValue
Route data for 'otherPagesTemplate' was provided: GlobalRouteValue"
in contradiction with the note stating that the order property decided the route order "The Order property for the AttributeRouteModel is set to 1. This ensures that the template for {globalTemplate?} (set earlier in the topic) is given priority for the first route data value position". Here there was no conflict between equally specific routes and was determined by the calls to the order.

guardrex:
Fixes #6173 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/razor-pages/razor-pages-conventions?view=aspnetcore-2.0&branch=pr-en-us-8523)

@rynowak This follows our discussion on https://github.com/aspnet/Mvc/issues/8385.

Thanks again to @user135711 for surfacing the issues.

### DIFF
--- a/aspnetcore/razor-pages/razor-pages-conventions.md
+++ b/aspnetcore/razor-pages/razor-pages-conventions.md
@@ -81,6 +81,8 @@ Route processing is established by convention:
 
 If possible, avoid depending on an established route processing order. Generally, routing selects the correct route with URL matching. If you must set route `Order` properties to route requests correctly, the app's routing scheme is probably confusing to clients and fragile to maintain. Seek to simplify the app's routing scheme. The sample app requires an explicit route processing order to demonstrate several routing scenarios using a single app. However, you should attempt to avoid the practice of setting route `Order` in production apps.
 
+Razor Pages routing and MVC controller routing share an implementation. Information on route order in the MVC topics is available at [Routing to controller actions: Ordering attribute routes](xref:mvc/controllers/routing#ordering-attribute-routes).
+
 ## Model conventions
 
 Add a delegate for [IPageConvention](/dotnet/api/microsoft.aspnetcore.mvc.applicationmodels.ipageconvention) to add [model conventions](xref:mvc/controllers/application-model#conventions) that apply to Razor Pages.
--- a/aspnetcore/razor-pages/razor-pages-conventions.md
+++ b/aspnetcore/razor-pages/razor-pages-conventions.md
@@ -63,6 +63,24 @@ public void ConfigureServices(IServiceCollection services)
 }
 ```
 
+## Route order
+
+Routes specify an <xref:Microsoft.AspNetCore.Mvc.ApplicationModels.AttributeRouteModel.Order*> for processing (route matching).
+
+| Order            | Behavior |
+| :--------------: | -------- |
+| -1               | The route is processed before other routes are processed. |
+| 0                | Order isn't specified (default value). Not assigning `Order` (`Order = null`) defaults the route `Order` to 0 (zero) for processing. |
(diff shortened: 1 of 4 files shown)


aspnetcore/razor-pages/razor-pages-conventions.md
```diff
@@ -63,6 +63,24 @@ public void ConfigureServices(IServiceCollection services)
...
+| 1, 2, &hellip; n | Specifies the route processing order. |
+
+Route processing is established by convention:
+
+* Routes are processed in sequential order (-1, 0, 1, 2, &hellip; n).
+* When routes have the same `Order`, the most specific route is matched first followed by less specific routes.
+* When routes with the same `Order` and the same number of parameters match a request URL, routes are processed in the order that they're added to the <xref:Microsoft.AspNetCore.Mvc.ApplicationModels.PageConventionCollection>.
+
+If possible, avoid depending on an established route processing order. Generally, routing selects the correct route with URL matching. If you must set route `Order` properties to route requests correctly, the app's routing scheme is probably confusing to clients and fragile to maintain. Seek to simplify the app's routing scheme. The sample app requires an explicit route processing order to demonstrate several routing scenarios using a single app. However, you should attempt to avoid the practice of setting route `Order` in production apps.
```
rynowak: This section all seems pretty good. If there's details about this in the regular routing section it might be helpful to link it. This is a shared implementation with regular routing.

guardrex: Yes, there's one spot in routing where this is discussed, and there is some duplication. Of course, `PageCollectionConvention` isn't discussed over there. I'll take another look at this in a few minutes and see about cross-linking both ways (just in case someone looking for RP content lands over there).

PR #8538, opened 2018-09-14
Nix SignalR hubs mention from ActivatorUtilities remark

### LINKED ISSUE #8537: Cannot explicitly instantiate a SignalR hub
From this article:

"Services can be resolved by two mechanisms:

*    IServiceProvider

*    ActivatorUtilities – Permits object creation without service registration in the dependency injection container. ActivatorUtilities is used with user-facing abstractions, such as Tag Helpers, MVC controllers, SignalR Hubs, and model binders."

The way I understood this was that, if it was necessary, I could explicitely instantiate a SignalR hub via Activator utilities.

So I tried:

    using (var scope = scopeFactory.CreateScope())
    {
        var hub = ActivatorUtilities.CreateInstance<ChatHub>(scope.ServiceProvider);
        // do stuff with hub
    }

Unfortunately, it seems that this yields the hub in an unusable state; hub.clients and other important properties are null, so I can't for example call

    hub.Clients.All.SendAsync("something", args)

Trial and error showed me that:

    using (var scope = scopeFactory.CreateScope())
    {
        var hub = scope.ServiceProvider.GetRequiredService<IHubActivator<ChatHub>>().Create();
        // do stuff with hub
    }

This again leads to creating an unusable hub for the same reasons as above;

    using (var scope = scopeFactory.CreateScope())
    {
        var hubContext = scope.ServiceProvider.GetRequiredService<IHubContext<ChatHub>>();
        // do stuff with hubContext
    }

Now this is usable - I can call hubContext.Clients.All.SendAsync("something", args) and similar stuff and this works - BUT this is not a full hub so I have no access to private methods I have defined in ChatHub and I also don't have access to services injected in ChatHub's constructor - which is inconvenient but can be worked around;

    using (var scope = scopeFactory.CreateScope())
    {
        var hub = ActivatorUtilities.CreateInstance<ChatHub>(scope.ServiceProvider);
        var hubContext = scope.ServiceProvider.GetRequiredService<IHubContext<ChatHub>>();
        hub.Context = hubContext;
        // do stuff with hub
    }

This fails to compile, but the compilation message hints about an explicit conversion, so I try it:

    using (var scope = scopeFactory.CreateScope())
    {
        var hub = ActivatorUtilities.CreateInstance<ChatHub>(scope.ServiceProvider);
        var hubContext = scope.ServiceProvider.GetRequiredService<IHubContext<ChatHub>>();
        hub.Context = (HubCallerContext)hubContext;
        // do stuff with hub
    }

This throws in the line when the conversion is done.

The way I'm getting the quoted paragraph from this article is that I should be able to explicitely instantiate a SignalR hub with ActivatorUtilities.CreateInstance. Is this the case? If so, how exactly? If I might suggest, should the documentation be more precise with regard to this?

---
#### Document Details

* Content: [Dependency injection in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.1#call-services-from-main)
* Content Source: [aspnetcore/fundamentals/dependency-injection.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/dependency-injection.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #8537

### DIFF
--- a/aspnetcore/fundamentals/dependency-injection.md
+++ b/aspnetcore/fundamentals/dependency-injection.md
@@ -237,7 +237,7 @@ Singleton lifetime services are created the first time they're requested (or whe
 Services can be resolved by two mechanisms:
 
 * `IServiceProvider`
-* [ActivatorUtilities](/dotnet/api/microsoft.extensions.dependencyinjection.activatorutilities) &ndash; Permits object creation without service registration in the dependency injection container. `ActivatorUtilities` is used with user-facing abstractions, such as Tag Helpers, MVC controllers, SignalR Hubs, and model binders.
+* [ActivatorUtilities](/dotnet/api/microsoft.extensions.dependencyinjection.activatorutilities) &ndash; Permits object creation without service registration in the dependency injection container. `ActivatorUtilities` is used with user-facing abstractions, such as Tag Helpers, MVC controllers, and model binders.
 
 Constructors can accept arguments that aren't provided by dependency injection, but the arguments must assign default values.
 


guardrex:
@scottaddie This is another one ... here's the engineering comment :point_right: https://github.com/aspnet/Docs/pull/8382 ... I think we can go off of that for this.

PR #8543, opened 2018-09-14
Add new ASP.NET issue templates

scottaddie:
Following in the footsteps of the dotnet/docs repo. Add 2 distinct issue templates: one for doc requests and one for actual doc issues. These templates apply to both ASP.NET 4.x and ASP.NET Core.


.github/ISSUE_TEMPLATE/doc-issue.md
```diff
@@ -0,0 +1,36 @@
...
+- A general support question, consider asking on a support forum site.
+- A site design concern, create an issue at [MicrosoftDocs/Feedback](https://github.com/MicrosoftDocs/Feedback/Issues/new/choose).
+- A problem completing a tutorial, compare your code with the completed sample.
+- A duplicate of an open or closed issue, leave a comment on that issue.
+
+**Issue description**
+
+&lt;include description here&gt;
+
+**Target framework**
```
tdykstra: More than Target Framework is asked for in this section.  Could be named "Environment" or "Software versions"?  I wonder how often it will get filled out and will be helpful.  Perhaps we could plan a review after some months, and if it doesn't get used much, remove the second two sections leaving only "Before you open an issue".

PR #8551, opened 2018-09-15
fixed link not rendered

user135711:
A non standard parenthesis was inserted on the left parenthesis of the link I added likely due to one of the foreign keyboards I use being activated when I entered it.


guardrex:
Thanks @user135711.

@Rick-Anderson Do you want to keep this link? IMO, these ref source links are too hard to maintain release-to-release. Also, this links to live, which is a preview release and might not reflect what devs are actually using. Therefore, I get rid of these whenever I can.

Rick-Anderson:
@guardrex We should avoid them. In this case lets's keep it. It should get folks close to the correct code should it change.

Rick-Anderson:
@user135711  thanks for fixing this.

PR #8578, opened 2018-09-17
Fix compile errors

BrennanConroy:


scottaddie:
Thanks for fixing this, @BrennanConroy!

PR #8590, opened 2018-09-18
Update tracing-in-aspnet-web-api.md

scout208:
Cleaned up first sentence in "How Web API Tracing Works".


scottaddie:
@scout208 Thank you for fixing this. Your change will go live later today.

PR #8595, opened 2018-09-18
Fix SPA warning

Rick-Anderson:
Long term fix via #8596


Rick-Anderson:
@scottaddie I'm going to merge this now. We can plan on updating the doc when we have time.

scottaddie:
@Rick-Anderson Sorry for the delay on this. Can you also please remove the sample download link from https://docs.microsoft.com/en-us/aspnet/core/client-side/spa-services?view=aspnetcore-2.1? If the *package.json* file isn't there, we will have readers filing issues stating that the sample doesn't work.

PR #8597, opened 2018-09-18
CC71738: Escaping components names to avoid broken format in translation pages

v-maudel:
Hello, @rick-anderson,
Description: Some of the "folder names" inside \<\> are not displaying well in localized pages, and also those terms are not blocked for localization.

Many thanks in advance.

### DIFF
--- a/aspnetcore/mvc/views/view-components.md
+++ b/aspnetcore/mvc/views/view-components.md
@@ -69,21 +69,20 @@ A view component defines its logic in an `InvokeAsync` method that returns an `I
 
 The runtime searches for the view in the following paths:
 
-* /Pages/Components/`<view_component_name>`/`<view_name>`
-* /Views/`<controller_name>`/Components/`<view_component_name>`/`<view_name>`
-* /Views/Shared/Components/`<view_component_name>`/`<view_name>`
-<!-- Could you please confirm if the components between < and > needs to be left as it or maybe they should be -->
+* /Pages/Components/{View Component Name}/{View Name}
+* /Views/{Controller Name}/Components/{View Component Name}/{View Name}
+* /Views/Shared/Components/{View Component Name}/{View Name}
 
 The default view name for a view component is *Default*, which means your view file will typically be named *Default.cshtml*. You can specify a different view name when creating the view component result or when calling the `View` method.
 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@guardrex  I think I prefer code-fencing. @v-maudel  thanks for the PR. We'll try to do this in the future.

guardrex:
To be consistent with other paths, are we going to do the whole path?

Rick-Anderson:
@v-maudel  I though "view_component_name" "view_name" should be localized.

What you have is perfect for english. Would the following be better for loc?

/Views/{controller name}/Components/{view component name}/{view name}

v-maudel:
Thanks @guardrex and @Rick-Anderson 
If those names should be translated, maybe we can adopt the `&lt;` and `&gt;` HTML or maybe escaping the greater than and lower than symbols with backlashes.
The thing is for some reason the actual syntax is blocking for translation some strings but not all of them.

Regards,

guardrex:
@v-maudel Are braces a problem for localization ... { } ... ? Will the following localize? ...

> Here is a sentence with a link, /Views/{controller name}/Components/{view component name}/{view name}, that isn't code-fenced.

Next ? ... Any chance that the engineers could provide something equivalent to the code-fenced **_CSS style_** that will still allow localization? Something that produces the same effect today as ...

> Here is a sentence with a link, \<code>/Views/{controller name}/Components/{view component name}/{view name}\</code>, in-text and not code-fenced.

I know we don't want HTML tags. I'm just asking about producing a similar CSS style as code-fenced content that would localize.

Rick-Anderson:
@nokura What's the best way to get "controller name", "view component name", and "view name" localized in the following path

/Views/{controller name}/Components/{view component name}/{view name}

Everything inside of {} should be localized.

v-makoud:
Hi @nokura and @sonja,
can you please look into this?

Best regards

Cc: @Rick-Anderson

v-makoud:
Hi @nokura and @Sonja,
can you please look into this?

Best regards

Cc: @Rick-Anderson

v-maudel:
HI @Rick-Anderson , 
@guardrex approach seems to be effective, could you please take a look and let us know?

Thanks.

Cc: @nokura

guardrex:
I've put in a few PRs with this format, and this approach is looking very good to me.

I'll leave this for @Rick-Anderson, @scottaddie, and @tdykstra to let me know if this is what I should do going forward. Let me know if I should keep going with it.

tdykstra:
@guardrex This approach looks good to me.

Rick-Anderson:
@JasonCard please review what we're doing.

PR #8638, opened 2018-09-20
Update instruction on picking up env var change

### LINKED ISSUE #8635: Guidance around changing environment variables in confusing
&gt; [!IMPORTANT]
&gt; When hosting an app in IIS and adding or changing the `ASPNETCORE_ENVIRONMENT` environment variable, use any one of the following approaches to have the new value picked up by apps:
&gt;
&gt; * Restart an app's app pool.
&gt; * Execute `net stop was /y` followed by `net start w3svc` from a command prompt.
&gt; * Restart the server.

Restarting an AppPool will only suffice if you set the environment variable on the AppPool. If you set a system-wide environment variable, that will not suffice. It's impossible to figure out from the docs are currently phrased.

---
#### Document Details

* ID: ceffa6a2-770c-1cd9-5145-dcd423f9dd22
* Version Independent ID: 1cb5eb30-dd84-e5e1-5826-d74243fabf04
* Content: [Use multiple environments in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-2.1#windows)
* Content Source: [aspnetcore/fundamentals/environments.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/environments.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

guardrex:
Fixes #8635 

Let me know if you'd prefer to see it just changed to something along the lines of ...

> Restart an app's app pool. This approach only works if the environment variable is set on the app pool. This approach doesn't work for a system-wide environment variable.

If we were to go that way, should it address the user env var case? (restart of app pool tied to a specific user account and the env var is set for that user)?

However, I removed the item on this PR because it's easier for a general audience to follow the simpler instructions for `net stop`/`net start` or restarting the server.

### DIFF
--- a/aspnetcore/fundamentals/environments.md
+++ b/aspnetcore/fundamentals/environments.md
@@ -239,7 +239,6 @@ To set the `ASPNETCORE_ENVIRONMENT` environment variable for an app running in a
 > [!IMPORTANT]
 > When hosting an app in IIS and adding or changing the `ASPNETCORE_ENVIRONMENT` environment variable, use any one of the following approaches to have the new value picked up by apps:
 >
-> * Restart an app's app pool.
 > * Execute `net stop was /y` followed by `net start w3svc` from a command prompt.
 > * Restart the server.
 


guardrex:
ping @shirhatti

PR #8685, opened 2018-09-24
Adding Note about the implicit `AddHttpContextAccessor`

sguryev:
`.AddIdenity()` is already providing the `.AddHttpContextAccessor()` registration. I have spent about 30 mins to understand how does it resolve the Accessor with no registration.

### DIFF
--- a/aspnetcore/fundamentals/http-context.md
+++ b/aspnetcore/fundamentals/http-context.md
@@ -89,7 +89,7 @@ public void ConfigureServices(IServiceCollection services)
 ```
 
 > [!NOTE]
-> `.AddIdenity()` is already providing the `.AddHttpContextAccessor()` registration. Sources: https://github.com/aspnet/Identity/blob/9b385180a9abcb264507efc23279f083bfc50520/src/Identity/IdentityServiceCollectionExtensions.cs#L79 . You can add the explicit registration without any problems related to the twice registration or performance. Discussion: https://github.com/aspnet/Hosting/issues/793#issuecomment-224831136
+> <xref:System.Security.Claims.ClaimsPrincipal.AddIdentity*> provides `AddHttpContextAccessor()` registration. Sources: https://github.com/aspnet/Identity/blob/9b385180a9abcb264507efc23279f083bfc50520/src/Identity/IdentityServiceCollectionExtensions.cs#L79. An additional explicit registration can be provided without generating an error or performance penalty. Discussion: https://github.com/aspnet/Hosting/issues/793#issuecomment-224831136
 
 ::: moniker-end
 
--- a/aspnetcore/fundamentals/http-context.md
+++ b/aspnetcore/fundamentals/http-context.md
@@ -89,7 +89,7 @@ public void ConfigureServices(IServiceCollection services)
 ```
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


guardrex:
I suggest streamlining this a bit. It's best if we don't call on issues and code unless absolutely necessary. Can it just say ...

> \> [!NOTE]
> \> \<xref:System.Security.Claims.ClaimsPrincipal.AddIdentity*> provides \`AddHttpContextAccessor\` registration. An additional explicit registration can be provided without generating an error or performance penalty.

sguryev:
I have updated the wording though. Thank you!

guardrex:
The problem for source links is usually that either:

* The link breaks because the API changes (they move things around or the line number changes).
* The link points to a stale version.

The problem with issue links is that we don't monitor those issues release-to-release. When engineering resolves an issue, we often (usually) don't know and don't update a topic. I suppose we could open an issue for the next release to come back and look. However, that would generate a lot of new issues if we did it frequently.

My recommendation is to avoid links to source and avoid links to issues unless absolutely necessary.

I defer to @Rick-Anderson 🎷 and @scottaddie 🎸 on how they'd like to go.

guardrex:
In that case, I think something along the lines of my recommendation would be best here. :point_up: https://github.com/aspnet/Docs/pull/8685#issuecomment-424114690

scottaddie:
@guardrex I like your recommendation, but keep the following things in mind:
* `AddHttpContextAccessor` is a 2.1+ feature.
* The API ref link you provided is incorrect. Should be `<xref:Microsoft.Extensions.DependencyInjection.IdentityServiceCollectionExtensions.AddIdentity*>`.

guardrex:
Thanks ... I fixed the remark.

sguryev:
@scottaddie thank you for the comments. I have added the Note to the 2.1 moniker.
@guardrex done

aspnetcore/fundamentals/http-context.md
```diff
@@ -88,6 +88,9 @@ public void ConfigureServices(IServiceCollection services)
 }
 ```
 
+> [!NOTE]
+> <xref:Microsoft.Extensions.DependencyInjection.IdentityServiceCollectionExtensions.AddIdentity*> provides <xref:Microsoft.Extensions.DependencyInjection.HttpServiceCollectionExtensions.AddHttpContextAccessor*> registration. An additional explicit registration can be added without generating an error or performance penalty.
```
scottaddie: The "provides AddHttpContextAccessor registration" part of the first sentence sounds odd to me. I suggest replacing the following:

```
provides <xref:Microsoft.Extensions.DependencyInjection.HttpServiceCollectionExtensions.AddHttpContextAccessor*> registration.
```

with this:

```
registers <xref:Microsoft.AspNetCore.Http.IHttpContextAccessor>.
```

Thoughts on this, @davidfowl?

davidfowl:
While I understand the reasoning behind wanting to add this, it feels like an implementation detail that could change.

scottaddie:
After some offline discussion with engineering, we've decided to not include this language in the doc. David's explanation above summarizes the reasoning. @sguryev We very much appreciate you submitting this PR, so that we could have that discussion.

sguryev:
It's really confusing to me guys. I have discovered the part of the current implementation which is important to the developers apart from babysitters who use mouse only to create applications. Startup.cs file becomes shorter because of new extensions methods which hide the logic inside. I want to know what does each of them do to make sure that I have full control of the process. It's annoing that I have to search through forums, GitHub and decompiled sources to get this knowledge. I have faced the issue with changing the hosted services registration from singleton to transient. And this hidden registration is one more sample. I think you have to give me respect to guys who wants to understand the core processes and not only use boilerplate templates.

PR #8720, opened 2018-09-26
Separate IFilterFactory from TypeFilterAttribute

royshea:
Add separate section using h3 title to separate out the IFilterFactory description from that of the TypeFilterAttribute when discussing dependency injection.


scottaddie:
@royshea Nice catch! Thank you for your PR. This change will go live later today.

PR #8724, opened 2018-09-27
Use the correct type as a parameter for the generic

Maxim-Kornilov:
Docs/aspnetcore/fundamentals/host/generic-host/samples-snapshot/2.x/GenericHostSample/


guardrex:
Thanks for the fix @Maxim-Kornilov! :rocket:

PR #2996, opened 2017-03-17
Update overview.md

ChrisSfanos:
Adding jQuery 3.2.0 to the CDN

PR #8729, opened 2018-09-27
Add available clients to SignalR supported platforms

tdykstra:
Fixes #2996

@anurse -- what versions of Node and Java do the SignalR clients support?


Rick-Anderson:
@anurse @BrennanConroy -- what versions of Node and Java do the SignalR clients support?

tdykstra:
@anurse Can you confirm that the versions Brennan cited are correct, and what do you think of moving "Supported platforms" out of Server concepts in the ToC?
toc

anurse:
> what do you think of moving "Supported platforms" out of Server concepts in the ToC?

💯 🎆 😄 👍 

Sounds good to me.

PR #8734, opened 2018-09-27
Use forward slashes for destination path

eogas:
I don't have a way of testing this on Windows atm, but most of the other examples use forward slashes, so I assume this is correct on all platforms.  Backslashes fail on MacOS, creating a directory called "wwwrootlibsignalr".


scottaddie:
Tested this on Windows. Thank you for correcting this, @eogas!

mrcaron:
Way to go @eogas !

PR #8743, opened 2018-09-28
Added using for clarity

daveabrock:
Visual Studio does not recognize the query without using the `Microsoft.Linq` namespace, which isn't specified in the document.

Interestingly, without it - you get an error saying `Could not find an implementation of the query pattern for source type 'DbSet<Movie>.' 'Select' not found.` You can't even have VS easily import the namespace using `Ctrl` + `.` or similar methods.


Rick-Anderson:
Thanks @daveabrock

PR #8757, opened 2018-09-29
Use $http_host instead of $host

mne1991:
Use $http_host instead of $host to not temper the port in case of using non standard ports.


guardrex:
Hello @mne1991 ... We use `$host` to ward off header spoofing. It was a change we made back in March. See https://github.com/aspnet/Docs/issues/5707 and the linked content there. If you would like to discuss it further with the engineers, you can open a new issue on the [aspnet/Home issues](https://github.com/aspnet/Home/issues).

PR #8769, opened 2018-10-01
correct a term

obayit:
The word implement can mislead the reader into thinking that he have to implement the interface,
an alternative is the word require, which is used when talking about dependency injection

  Fixes #Issue_Number

### DIFF
--- a/aspnetcore/fundamentals/logging/index.md
+++ b/aspnetcore/fundamentals/logging/index.md
@@ -19,7 +19,7 @@ For information on stdout logging when hosting with IIS, see <xref:host-and-depl
 
 ## How to create logs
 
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks @obayit 

"Ask" personifies the framework a bit much imo. I like "obtain" (and remove "object") ...

> To create logs, obtain an `ILogger<TCategoryName>` from the [dependency injection](#) container:

IIRC, we use "obtain" elsewhere in similar circumstances.

scottaddie:
@guardrex I'm in agreement that we should avoid "ask" in this case. Your suggested text looks good to me.

obayit:
Thanks for the correction @guardrex

PR #8786, opened 2018-10-01
Code fence <receiver> uri component

MaxRamult:
The description of the webhook uri contains the following:
```
"The component is the name of the receiver, for example github or slack"
```
It should read:
```
 "The <receiver> component is the name of the receiver, for example github or slack"
```
This PR code fences &lt;receiver&gt; so it displays correctly.

### DIFF
--- a/aspnet/webhooks/receiving/receivers.md
+++ b/aspnet/webhooks/receiving/receivers.md
@@ -28,7 +28,7 @@ https://<host>/api/webhooks/incoming/<receiver>/{id}
 
 For security reasons, many WebHook receivers require that the URI is an *https* URI and in some cases it must also contain an additional query parameter which is used to enforce that only the intended party can send WebHooks to the URI above.
 
-The <em>&lt;receiver&gt;</em> component is the name of the receiver, for example <em>github</em> or <em>slack</em>.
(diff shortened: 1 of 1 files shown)


aspnet/webhooks/receiving/receivers.md
```diff
@@ -28,7 +28,7 @@ https://<host>/api/webhooks/incoming/<receiver>/{id}
 
 For security reasons, many WebHook receivers require that the URI is an *https* URI and in some cases it must also contain an additional query parameter which is used to enforce that only the intended party can send WebHooks to the URI above.
 
-The <em><receiver></em> component is the name of the receiver, for example <em>github</em> or <em>slack</em>.
+The <em>&lt;receiver&gt;</em> component is the name of the receiver, for example <em>github</em> or <em>slack</em>.
```
scottaddie: Let's go with the following syntax instead:

```
`<receiver>`
```

We avoid HTML elements whenever possible. Also, code-fencing this text eliminates the need to encode the symbols.

scottaddie:
Many thanks for the fix, @MaxRamult! Also, congrats on your 1st contribution to our Docs repo. Your change will go live within the next 24 hours.

PR #8788, opened 2018-10-01
Fixed hyperlink so user routed to the correct page to create a new issue

RyanLilla:
Noticed that in the template for creating a Doc Issue, the link navigating you to the page to choose which type of issue applies is not valid and returns a 404 error instead. this is due to the fact that in the url itself, "Issues" should not be capitalized -- Changing to lowercase fixes the problem.

**Live:** https://github.com/MicrosoftDocs/Feedback/Issues/new/choose
**PR Change:** https://github.com/MicrosoftDocs/Feedback/issues/new/choose


Rick-Anderson:
@RyanLilla 	 thanks. That's really bad on GitHubs part to have case sensitive URLs.

PR #8800, opened 2018-10-02
mention HTML sanitization as deprecated approach

explunit:
Include some content from https://github.com/aspnet/Docs/issues/7962#issuecomment-410779470 that would be helpful to those who search for this topic.


aspnetcore/security/cross-site-scripting.md
```diff
@@ -219,4 +219,4 @@ The general accepted practice is that encoding takes place at the point of outpu
...
+Validation can be a useful tool in limiting XSS attacks. For example, a numeric string containing only the characters 0-9 won't trigger an XSS attack. Validation becomes more complicated should you wish to accept HTML in user input. Parsing HTML input is difficult, if not impossible. Previous versions of Anti-XSS libraries included a sanitization feature for stripping "unsafe" elements from HTML input. This approach is no longer recommended, and future libraries won't include this feature. Markdown, coupled with a parser that strips embedded HTML, is a safer option for accepting rich input. Never rely on validation alone. Always encode untrusted input before output, no matter what validation or sanitization has been performed.
```
blowdart: There are no future versions planned, so I'd remove any hint that some might be forthcoming. In fact I'd remove all reference to AntiXSS

"Validation can be a useful tool in limiting XSS attacks. For example, a numeric string containing only the characters 0-9 won't trigger an XSS attack. Validation becomes more complicated should you wish to accept HTML in user input. Parsing HTML input is difficult, if not impossible.  Markdown, coupled with a parser that strips embedded HTML, is a safer option for accepting rich input. Never rely on validation alone. Always encode untrusted input before output, no matter what validation or sanitization has been performed."

blowdart:
AntiXSS was never support on Core, and I feel mentioning it will just confuse all the more I'm afraid.

Issue #8323, opened 2018-08-29
Localization sample project has resources in wrong folder


Eilon:
Original issue is here: https://github.com/aspnet/Home/issues/3472

> FIrst of all i have to tell that seems like there's an errors in sample localization project: https://github.com/aspnet/Docs/tree/master/aspnetcore/fundamentals/localization/sample/Localization, when trying convention resolving i found that InfoController localization not working until you place resources in Resources/StarterWeb/Controllers/…. Folder where it searching them for.

(Note: I have not verified this issue, I'm just copying it from the original issue.)

cc @Unders0n

guardrex:
The sample is due for an update, but we don't have an ETA for that at this time. I've cross-referenced this issue on the sample update tracking issue (#5495) to make sure that this is addressed when the sample is updated.

hishamco:
@Rick-Anderson this need to be closed

Unders0n:
Sorry, i will post a sample in the coming days

guardrex:
@hishamco, are you saying that the work is either already planned on another issue or that the work is already complete?

hishamco:
@guardrex It's already done in commit https://github.com/aspnet/Docs/commit/29c39626cd2bf929507a4eed1512fe8fff6d888a

guardrex:
Oh ... I see what happened. You wrote ...

> This address #8219 #8323, 

... but GH only auto-closes these issues when the PR says "Fixes" ...

```
Fixes #8219
Fixes #8323
```

hishamco:
Perhaps I didn't mention that in the PR title itself, really i'm not sure how the auto closing is happen :smile:

guardrex:
Do it this way in the PRs comment ...

```
Fixes #8219
Fixes #8323
```

... and the issues will auto-close when the PR merges.

guardrex:
If you want the issue to say open, use "Addresses" instead ...

```
Fixes #8219
Addresses #8323
```

8219 will close ... 8323 will remain open.

hishamco:
I see, thanks for the new information that I know today, I will do it for my next PR today

thanks

guardrex:
I just added to my :cow2: **_Where's the BEEF?!_**:tm: :cow2: issue https://github.com/aspnet/Docs/issues/2617. I'm sort of keeping that up-to-date. We might do something to surface some of it on https://github.com/aspnet/Docs/issues/8814.

Issue #8219, opened 2018-08-22
Wrong Localization sample


ckams:
Seems a little [buggy sample](https://github.com/aspnet/Docs/tree/master/aspnetcore/fundamentals/localization/sample/Localization), Resources/ViewModels/ must be inside Resources/StarterWeb to get Localization for Email, Password, etc…

hishamco:
@Rick-Anderson this need to be closed

PR #8860, opened 2018-10-05
Fix localization sample namespace

hishamco:
This address #8219 #8323, @Rick-Anderson seems the original code is copied from the entropy repository which is make the app buggy, to avoid such issues I fixed the namespace and everything now it works as expected


Rick-Anderson:
@hishamco 	much appreciated.

PR #8869, opened 2018-10-05
Clarifications around ConfigureAll and named options

andrewlock:
`ConfigureAll` configures _all_ instances, not just explicitly named instances. The "Note" section sort explains this later by mentioning that that "All options are named instances". But I think it makes the behaviour clearer to remove the "named" qualifier in this case?


andrewlock:
I agree -`IPostConfigureOptions` has its own section, why not poor old `ConfigureAll`? 😉 

I think

> \#\# Configure all options (ConfigureAll method)

... is maybe best? Or ...

> \#\# Configure all options  with the ConfigureAll method

guardrex:
> \#\# Configure all options with the ConfigureAll method

... seems nice.

andrewlock:
Cool, added the section header

PR #8884, opened 2018-10-07
Fixed all the checklists

phanirithvij2000:
The syntax used for checklists in markdown doesn't seem to be correct.


phanirithvij2000:
The preview in GitHub isn't like that.

guardrex:
Yes, that's correct. We intend for readers to access the built docs, so we don't use GH preview behavior/rendering for content decisions.

It was discussed a while back when someone asked why we don't use the code notes feature for code content (`[!code-csharp[we could use this for a note](PATH)]`) because they surf the docs in GH and the note becomes a clickable link to the code.

It was decided that it's too burdensome to maintain the content for two sets of rules. The vast majority of readers use the built docs.

PR #8918, opened 2018-10-09
Configuration enhancements

### LINKED ISSUE #8795: JSON FileProvider does not mention new method .ConfigureAppConfiguration() on the WebHostBuilder.
When trying to add my own custom JSON file to the configuration using ASP .NETCore 2.1 I had to use .ConfigureAppConfiguration() instead of  .AddJsonFile() as the latter wouldn't add the configuration to the provider list.  Couldn't find that documented here

---
#### Document Details

* Content: [Configuration in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-2.1#json-configuration-provider)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #8795 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/index?view=aspnetcore-1.0&branch=pr-en-us-8918)

Although the topic does call out `ConfigureAppConfiguration` early in the coverage (for 2.1), it didn't show provider examples using it down through the topic.

It was relying too heavily on devs *remembering* what they read and **not showing up from an external link directly to a provider section**.

This PR seeks to enhance the examples in every provider section.

I can't see the internal review topic, so plz put an :eye: on that and confirm the versioning is all :+1:.

### DIFF
--- a/aspnetcore/fundamentals/configuration/index.md
+++ b/aspnetcore/fundamentals/configuration/index.md
@@ -4,7 +4,7 @@ author: guardrex
 description: Learn how to use the Configuration API to configure an ASP.NET Core app.
 ms.author: riande
 ms.custom: mvc
-ms.date: 08/13/2018
+ms.date: 10/09/2018
 uid: fundamentals/configuration/index
 ---
 # Configuration in ASP.NET Core
@@ -185,7 +185,7 @@ Configuration sources are read in the order that their configuration providers a
 
 A typical sequence of configuration providers is:
 
-1. Files (*appsettings.json*, *appsettings.&lt;Environment&gt;.json*, where `<Environment>` is the app's current hosting environment)
+1. Files (*appsettings.json*, *appsettings.{Environment}.json*, where `<Environment>` is the app's current hosting environment)
 1. [User secrets (Secret Manager)](xref:security/app-secrets) (in the Development environment only)
 1. Environment variables
 1. Command-line arguments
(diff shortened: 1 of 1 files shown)


guardrex:
Linking up to the earlier section was an option. However, we have a bunch of incoming links from other topics that go directly to a provider section here, and I think we'll have more inbound links in the future. This seems like a winning approach ... they can't miss the config now.

aspnetcore/fundamentals/configuration/index.md
```diff
@@ -185,7 +185,7 @@ Configuration sources are read in the order that their configuration providers a
 
 A typical sequence of configuration providers is:
 
-1. Files (*appsettings.json*, *appsettings.&lt;Environment&gt;.json*, where `<Environment>` is the app's current hosting environment)
+1. Files (*appsettings.json*, *appsettings.{Environment}.json*, where `<Environment>` is the app's current hosting environment)
```
scottaddie: - `<Environment>` --> `{Environment}`
- List the Azure Key Vault provider and link to our doc. That provider takes precedence over Secret Manager, since it's a more secure provider.

aspnetcore/fundamentals/configuration/index.md
```diff
@@ -336,7 +401,41 @@ Switch mappings dictionary key rules:
...
+        WebHost.CreateDefaultBuilder()
+            .ConfigureAppConfiguration((hostingContext, config) =>
+            {
+                config.AddCommandLine(args, _switchMappings)
+            })
+            .UseStartup<Startup>();
+}
+```
+
+As shown in the preceding example, the call to `CreateDefaultBuilder` shouldn't pass arguments when switch mappings are used. `CreateDefaultBuilder` method's `AddCommandLine` call doesn't include mapped switches, and there's no way to pass the switch mapping dictionary to `CreateDefaultBuilder`. If the arguments include a mapped switch and are passed to `CreateDefaultBuilder`, its `AddCommandLine` provider fails to initialize with a <xref:System.FormatException>. The solution is not to pass the arguments to `CreateDefaultBuilder` but instead to allow the `ConfigurationBuilder` method's `AddCommandLine` method to process both the arguments and the switch mapping dictionary.
```
scottaddie: is not --> isn't

aspnetcore/fundamentals/configuration/index.md
guardrex: > \<Environment> --> {Environment}

*DAH!* 🙈 Find and replace ... it's great ... NOT smart ... but great.

PR #8927, opened 2018-10-09
Fix localization resources links

hishamco:
This PR address:
- Fix Localization.StarterWeb project link
- Change "Resource Files in Visual Studio" link because it seems talking about C++ specific resources

/cc @Rick-Anderson


hishamco:
Thanks again @guardrex 😗

PR #8957, opened 2018-10-10
Added missing navigation expressions

ViciousIXIL:
Changed

```csharp
WithOne()
```

to

```csharp
WithOne(e => e.User)
```

The documentation is not consistent, fixed these inconsistencies.


guardrex:
Hello @ViciousIXIL ... Please review and sign the CLA :point_up: if you agree with its terms.

PR #8959, opened 2018-10-10
Health Checks topic

### LINKED ISSUE #7734: Health Checks scenario coverage
> This feature is a way to trivially setup an endpoint in your application that can be used to check the health of your application, potentially taking into account things like memory consumption or database availability as well as just "can I accept a HTTP request".

https://github.com/aspnet/Announcements/issues/307

guardrex:
Fixes #7734 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-2.2&branch=pr-en-us-8959)

@glennc Would u like to co-author on this one?


aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,362 @@
+---
+title: Health checks in ASP.NET Core
+author: guardrex
+description: Learn how to set up health checks for ASP.NET Core infrastructure, such as apps and databases.
+monikerRange: '>= aspnetcore-2.2'
+ms.author: riande
+ms.custom: mvc
```
rynowak: mvc???

aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,362 @@
...
+ASP.NET Core offers Health Check Middleware and libraries for reporting the health of app infrastructure components.
+
+Health checks are exposed by an app as HTTP endpoints. Health check endpoints can be configured for a variety of real-time monitoring scenarios:
+
+* Health probes can be used by container orchestrators and load balancers to check an app's status. For example, a container orchestrator may respond to a failing health check by halting a rolling deployment or restarting a container. A load balancer might react to an unhealthy app by routing traffic away from the failing instance to a healthy instance.
+* Use of memory, disk, and other physical server resources can be monitored for healthy status.
+* Health checks can test an app's dependencies, such as databases and external service endpoints, to confirm availability and normal functioning.
+
+> [!WARNING]
+> Health Check Middleware doesn't disable caching of health check responses. This scenario is planned for a future release.
```
rynowak: This was addressed in preview3 - you can remove this warning 👍

aspnetcore/host-and-deploy/health-checks.md
guardrex: :smile: I asked that, too ...... *mvc* == "Minimum Viable Content" (I think)

rynowak: I'm going to start using that 👍

guardrex: I shouldn't say this, but APEX does some *very interesting things*. :smile: lol

aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,359 @@
...
+}
+```
+
+Add the `ExampleHealthCheck` with `AddCheck` and call `UseHealthChecks` in the processing pipeline with the endpoint URL:
+
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
+    services.AddHealthChecks()
+        .AddCheck(new ExampleHealthCheck);
```
rynowak: This example should be different - it should be `AddCheck<ExampleHealthCheck>()` because it metnions DI. If you construct the health check and pass it it the build it won't get DI services.

aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,359 @@
...
+        // variable to true.
+        var healthCheckResultHealthy = true;
+
+        // Return a HealthCheckResult value depending on the outcome of the
+        // check(s). Other HealthCheckResult outcomes not shown in this
+        // example are Degraded and Failed.
+        if (healthCheckResultHealthy)
+        {
+            return Task.FromResult(
+                HealthCheckResult.Healthy("The check indicates a healthy state."));
```
rynowak: This is a little out of date. We've pivoted on a few things since preview 3.

Here are the two important things to understand.
1. The application developer can configure the health check name and severity
2. The health check author can return pass/fail + a description + optional key value pairs


------

Here's a dump of some guidance I sent to the author of BeatPulse - some of it is for developing a **library** of health checks. Incorporate this as you see fit. 

> Some feedback on health check implementations:


> In general you don't need to log inside a health check. We do boilerplate logging for success and failure, so a log statement like "health check XYZ has failed" is redundant with that the developer gets by default. If you do have detailed things you want to report, that's what the description is for - which we also include in logs. I really struggle to think of a case where you'd need to log inside a health check implementation. Let me know if you have one, maybe I'm wrong 😂


> I would also suggest against put multiple checks inside a single implementation, such as ... where it loops through a list of URLs. If you just treat each URL as a separate check then will have already have a good logging and reporting experience. Let me know if you think there's a good case for this. It really seems to me like as an app developer I would want to see each URL that I'm checking pass/fail independently.

> Some feedback on AddXyzCheck methods:

> Your 'AddXyzCheck' methods accept the 'name' and 'failureSeverity' as parameters. It's of course cool and good to have default values for these as well - but it's intended that the application developer has control over these. This is important because the names need to be unique. So if you have multiple storage accounts, they can have their own check/logging/reporting/severity/tags. The same goes for tags, I'd suggest not providing any by default, and allow the developer to use as they desire.


> You can also use named options if it's a good fit for your scenario.  If you end up having any health checks that need DI and need access to data passed in by the caller its a good solution.


> You can see an example of all of this feedback via the EF health check I recently added to our repo: https://github.com/aspnet/Diagnostics/blob/release/2.2/src/Microsoft.Extensions.Diagnostics.HealthChecks.EntityFrameworkCore/DependencyInjection/EntityFrameworkCoreHealthChecksBuilderExtensions.cs

aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,359 @@
...
+
+### Docker example
+
+[Docker](xref:host-and-deploy/docker/index) offers a built-in `HEALTHCHECK` directive that can be used to check the status of an app that uses the basic health check configuration:
+
+```
+HEALTHCHECK CMD curl --fail http://localhost:5000/health || exit
+```
+
+## Create health checks
```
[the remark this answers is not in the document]
guardrex: I needed to explain the basic layout of a health check *before* I get to options. The first version of this topic (before I submitted the PR) didn't have this section and it was hard to grok the options without first introducing the overall concept/layout.

We could sit on the whole PR, but I'd rather merge it and then come back with a follow-up issue/PR.

aspnetcore/host-and-deploy/health-checks.md
rynowak: I'd suggest looking over these again, as they changed significantly https://github.com/aspnet/Diagnostics/blob/release/2.2/src/Microsoft.AspNetCore.Diagnostics.HealthChecks/Builder/HealthCheckApplicationBuilderExtensions.cs

aspnetcore/host-and-deploy/health-checks.md
guardrex: > This is a little out of date.

I think you meant 'since preview2', right? Preview 3 isn't out yet AFAIK.

So, we should sit on this after all?? ... I should pull down Preview 3 when it lands and update this topic with the latest bits/approach.

aspnetcore/host-and-deploy/health-checks.md
guardrex: > as they changed significantly

Sure thing ... I was following Glen's lead based on the blog post, which is probably *old* given the latest bits.

aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,359 @@
...
+
+Call Health Check Middleware in the app processing pipeline in `Startup.Configure`. In the sample app, the health check endpoints are created at `/health/ready` for the readiness check and `/health/live` for the liveness check. The liveness check filters out the `SlowDependencyHealthCheck` by returning `false` in the `HealthCheckOptions.Predicate` (for more information, see [Filter health checks](#filter-health-checks)):
+
+[!code-csharp[](health-checks/samples/2.x/HealthChecksSample/LivenessProbeStartup.cs?name=snippet_Configure&highlight=3,5-9)]
+
+To run the readiness/liveness configuration scenario using the sample app, execute the following command from the project's folder in a command shell:
+
+```console
+dotnet run --scenario liveness
+```
```
rynowak: This doesn't really change in preview3, except now we have tags. I'd show an example that uses tags for `liveness` and `readiness` to filter.

aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,359 @@
...
+
+The `WriteResponse` method formats the `CompositeHealthCheckResult` into a JSON object and yields JSON output for the health check response:
+
+[!code-csharp[](health-checks/samples/2.x/HealthChecksSample/CustomWriterStartup.cs?name=snippet_WriteResponse)]
+
+To run the metric-based probe with custom response writer output using the sample app, execute the following command from the project's folder in a command shell:
+
+```console
+dotnet run --scenario writer
+```
```
rynowak: This is another thing that the new BeatPulse libraries will provide. Our guidance will be to use those when available.

aspnetcore/host-and-deploy/health-checks.md
guardrex: > no more planned changes for preview3

Cool beans ... I'll get preview3 when it lands and make the updates and ping you back when it's ready for another :eye:. We'll have time to wrap it up before RTM.

aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,359 @@
...
+}
+```
+
+Register health check services with `AddHealthChecks` in `Startup.ConfigureServices`. The call to `UseHealthChecks` specifies the management port (*ManagementPortStartup.cs*):
+
+[!code-csharp[](health-checks/samples/2.x/HealthChecksSample/ManagementPortStartup.cs?name=snippet1&highlight=12,18)]
+
+> [!NOTE]
+> You can avoid creating the *launchSettings.json* file in the sample app by setting the URLs and management port explicitly in code. In *Program.cs* where the `WebHostBuilder` is created, add a call to `UseUrls` and provide the app's normal response endpoint and the management port endpoint. In *ManagementPortStartup.cs* where `UseHealthChecks` is called, specify the management port explicitly.
+>
```
rynowak: 👍  lovely

aspnetcore/host-and-deploy/index.md
```diff
@@ -40,36 +40,46 @@ An ASP.NET Core app is a console app that must be started when a server boots an
...
 
 ## Host in a web farm
 
 For information on configuration for hosting ASP.NET Core apps in a web farm environment (for example, deployment of multiple instances of your app for scalability), see <xref:host-and-deploy/web-farm>.
 
+::: moniker range=">= aspnetcore-2.2"
+
+## Perform health checks
+
+Use Health Check Middleware to perform health checks on an app and its dependencies. For more information, see <xref:host-and-deploy/health-checks>.
```
rynowak: Is it worth mentioning that health checks are an easy way for an external monitoring service or container system to check the status of your app? It's not something you would setup until you figure out how you will be monitoring the app.

guardrex:
@rynowak I marked it WIP and will revisit when preview3 lands. I'll ping u back after the updates are complete. Thanks for your help. :+1: :rocket:

aspnetcore/fundamentals/middleware/index.md
```diff
@@ -231,6 +231,7 @@ ASP.NET Core ships with the following middleware components. The *Order* column
 | [CORS](xref:security/cors) | Configures Cross-Origin Resource Sharing. | Before components that use CORS. |
 | [Diagnostics](xref:fundamentals/error-handling) | Configures diagnostics. | Before components that generate errors. |
 | [Forwarded Headers](/dotnet/api/microsoft.aspnetcore.builder.forwardedheadersextensions) | Forwards proxied headers onto the current request. | Before components that consume the updated fields. Examples: scheme, host, client IP, method. |
+| [Health Check](xref:host-and-deploy/health-checks) | Checks the health of ASP.NET Core infrastructure, such as database availability. | Terminal if a request matches a health check endpoint. |
```
glennc: 'Checks the health of ASP.NET Core infrastructure` makes me think that it's checking up on things about ASP.NET Core. "Checking the health of your ASP.NET Core application, such as by checking database availability" or something like that. Not sure exactly.

aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,359 @@
...
+
+## Basic health probe
+
+For many apps, a basic health probe configuration that reports the app's availability to process requests (*liveness*) is sufficient to discover the status of the app.
+
+The basic configuration registers health check services and calls the Health Check Middleware to respond at a URL endpoint with a health response. By default, no specific health checks are registered to test any particular dependency or subsystem. The app is considered healthy if it's capable of responding at the health endpoint URL. The default response writer writes the status (`HealthCheckStatus`) as `text/plain` content back to the client. The response is a *200 Ok* status code, and the content is either healthy (`HealthCheckResult.Healthy`) or unhealthy (`HealthCheckResult.Unhealthy`).
```
glennc: The part about the status code here seems miss-leading. Implying you always get a 200 regardless of the health status, since you talk about the content being unhealthy.

guardrex:
@rynowak I'm starting the work on reacting to Preview 3 updates.

RE: BeatPulse

> FYI we're working with the author of BeatPulse to port his health checks to our framework. We hope that will be done by RTM.
>
> This somewhat devalues guidance for writing checks yourself, because our primary guidance will be to use his libraries.

What is recommended regarding the use of BeatPulse versus a dev writing checks? This ...

https://github.com/Xabaril/BeatPulse

... looks parallel to Health Checks (at least on the surface, idk what's under the covers), so I'm not sure what I'm documenting about BeatPulse here.

Is the basic idea to say early in the topic, 'just use BeatPulse' (link out to BeatPulse docs? show an example of using BeatPulse?) ... then pick up with health checks in the subsequent sections of the topic?

guardrex:
Note to self: https://blogs.msdn.microsoft.com/webdev/2018/10/17/asp-net-core-2-2-0-preview3-now-available/#aspnetcoremodule

guardrex:
> our guidance would be to use these checks for things that they support

Are their bits going to be ready by 2.2 RTM such that our doc can describe them?

If 'yes,' what should I show? Should I just replicate their basic documentation example in our docs, or should I entirely punt the reader over to their docs?

... presumably in an early section of this topic that calls out BeatPulse as the *way to go, if possible*.

unaizorrilla:
Hi @guardrex,

Currently all checks and UI are preview3 ready and when ASP.NET 2.2 go RTM this also go RTM.

Currently I,m working on docs and samples and next week all should be done.

Feedback and contributions are welcome.

Thanks

guardrex:
Thanks for the update @unaizorrilla. I'll hear back from @rynowak on what we should show/describe + if the sample app should also use BeatPulse, then I can proceed with the updates on my end.

aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,359 @@
...
+            HealthCheckResult.Healthy("Foo is OK!")));
+}
+```
+
+## Health check options
+
+`HealthCheckOptions` provide an opportunity to customize health check behavior:
+
+* [Filter health checks](#filter-health-checks)
+* [Customize the HTTP status code](#customize-the-http-status-code)
```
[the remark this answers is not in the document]
guardrex: > suppress the anti-caching headers

The language logic of `SuppressCacheHeaders` seems harder to grok than needed. Did you reject options that pertain to the result of setting the headers and not what the property does to the headers themselves ...

* `SuppressCachedResponses`, `PreventCachedResponses`, `SuppressCaching` or `PreventCaching`
   * `true` adds the cache-suppression headers.
   * `false` leaves caching headers alone.
* `AllowCachedResponses` or `CacheResponses`
   * `true` leaves caching headers alone.
   * `false` adds the cache-suppression headers.

Sure, I'll put what `true`/`false` mean as it stands; but when I read the phrase "suppress cache headers," it's not immediately evident what's going to happen when I set the value one way or ther other. Actually when I see the property name, I think it's going to do the exact opposite of what it does ... I think it's going to suppress whatever caching headers the **_app already has set_** and apply a set that will stop caching.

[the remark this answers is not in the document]
rynowak: Yeah I think that's fair feedback. I'll log an issue and we'll consider changing it. In general any boolean options are phrased the way they are so that it makes sense when `false` is the default.

I think the best option name I've heard so far is `AllowCachedResponses` or `AllowCachingResponses`.

guardrex:
Update: I've dropped the BeatPulse SQL Server code into the app, and it's running fine ...

capture

rynowak:
LMK if/when this is ready for review.

guardrex:
> LMK if/when this is ready for review.

@rynowak Will do. I'm still working BeatPulse into the coverage (*db probe* and *liveness/readiness* areas of the topic+sample app; it's going well thus far 👍), and I still need to address your remarks on tag updates in Preview3.

I'll work on it today and ping u back either later today or tomorrow morning.

guardrex:
@rynowak The engineering sample for readiness delays a task inside the health check ...

```
public SlowDependencyHealthCheck()
{
    _task = Task.Delay(15 * 1000);
}
```

... but I'd like the health checks sample app to show a more realistic scenario.

Is a more realistic scenario to have a Hosted Service perform some background tasks at startup and the app shouldn't report 'ready' until those background tasks are finished?

If so, I have an implementation that pulls it off, but it seems a little hacky.

Let's address the question first (before we get into my **_RexHacks_**:tm: 🙈). What's a more realistic scenario for liveness/readiness than the engineering sample? Would it be a Hosted Service with a long-running backround task that should finish before the app reports 'ready' in its health check?

rynowak:
Yeah I agree. A better example would be like, reading some files from disk into a cache via a hosted service. The examples that we write in product repos are always fairly constrained in what kinds of things we can include.

guardrex:
@rynowak .... ok ... but just remember that **I warned you** 🙈😀 lol ...

A `StartupHostedServiceHealthCheck` has a property to keep tabs on when a startup Hosted Service is done ...

```csharp
public class StartupHostedServiceHealthCheck : IHealthCheck
{
    public string Name => "slow_dependency_check";

    public bool StartupTaskCompleted { get; set; } = false;

    public Task<HealthCheckResult> CheckHealthAsync(
        HealthCheckContext context, 
        CancellationToken cancellationToken = default(CancellationToken))
    {
        if (StartupTaskCompleted)
        {
            return Task.FromResult(
                HealthCheckResult.Passed("The startup task is finished."));
        }

        return Task.FromResult(
            HealthCheckResult.Failed("The startup task is still running."));
    }
}
```

The Hosted Service is registered as usual (and the health check is registered because the hosted service is going to need to set that property) ...

```csharp
services.AddHostedService<StartupHostedService>();
services.AddSingleton<StartupHostedServiceHealthCheck>();
```

The simulated long-running task in the Hosted Service sets that property when it's done (`_delaySeconds` is 15 seconds) ...

```csharp
public Task StartAsync(CancellationToken cancellationToken)
{
    _logger.LogInformation($"Startup Background Service is starting.");

    Task.Delay(_delaySeconds * 1000)
        .ContinueWith(_ => { _startupHostedServiceHealthCheck.StartupTaskCompleted = true; });

    return Task.CompletedTask;
}
```

😀 It ✨ **_Just Works_**:tm: ✨, but Chris and I never discussed how to keep tabs on a Hosted Service from outside of the service. What would a real developer do?

guardrex:
@rynowak Two items remain to figure out ...

1. The plumbing required to get a hosted service health check wired up *properly* to the hosted service. Although what I did works and is easy to understand, I don't trust it.

2. I'm still a bit confused on the role of BeatPulse here.

   > I think in general our guidance would be to use these checks for things that they support. I think that's what most users would prefer anyway

    You want the topic to encourage devs away from using the supported scenarios in the framework in favor of the unsupported scenarios&dagger; in BeatPulse? (&dagger;Unsupported by Microsoft.)
   * Which BeatPulse scenarios do you want to show in this topic (outside of SQL Server, which I've already implemented)?
   * Are BeatPulse scenarios going to be covered in the topic and in the sample app, or should they only be covered in the topic?
   * Do you want the BeatPulse scenarios SxS with the framework scenarios, or do you want them to completely replace the framework scenarios?

guardrex:
@unaizorrilla When Core 2.2 is RTM, I'll come in and perform a quick update to the sample to get the BeatPulse library code out of the sample app and activate a package reference for `BeatPulse.SqlServer` in the project file.

When you release the `BeatPulse.SqlServer` package, will it be a preview package or an ordinary release? Do you know what the package version will be?

unaizorrilla:
Hi @guardrex

I’m a litle bit confused, when you talk about BeatPulse.SqlServer is HealthChecks.SqlServer right? BeatPulse project  port for HealthChecks use this package names .

My idea is align versions with aspnetcore disgnostics package after RTM is released, ie 2.2.0 right?

Unai

guardrex:
Yes ... sorry about that ... `HealthChecks.SqlServer` :+1:

Correct, they will release 2.2.0.

unaizorrilla:
👍👍

rynowak:
I'm not ignoring you BTW - I've been out of the office 😲 . will look at this likely on Saturday

guardrex:
No worries ... no rush. *Plenty* to keep me busy over here. 🏃🏃🏃🏃😅 lol

aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,527 @@
...
+      # after pod startup, before applying health checking
+      initialDelaySeconds: 30
+      timeoutSeconds: 1
+    ports:
+      - containerPort: 80
+```
+
+## Metric-based probe with a custom response writer
+
+The sample app demonstrates a memory health check with a custom response writer.
```
rynowak: Note that beatpulse also has a disk space check: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/blob/master/src/HealthChecks.System/DiskStorageHealthCheck.cs

aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,527 @@
...
+```
+
+## Health Check Publisher
+
+When an `IHealthCheckPublisher` is added to the service container, the health check system periodically executes your health checks and calls `PublishAsync` with the result. This is useful in a push-based health monitoring system scenario that expects each process to call the monitoring system periodically in order to determine health.
+
+The `IHealthCheckPublisher` interface has a single method:
+
+```csharp
+Task PublishAsync(HealthReport report, CancellationToken cancellationToken);
```
rynowak: I think it's also worth mentioning beatpulse here as they have publishers for a few systems. AI is here - more to come https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/tree/master/src/HealthChecks.Publisher.ApplicationInsights

aspnetcore/host-and-deploy/health-checks/samples/2.x/HealthChecks.SqlServer/DependencyInjection/SqlServerHealthCheckBuilderExtensions.cs
```diff
@@ -0,0 +1,33 @@
+﻿using HealthChecks.SqlServer;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using System.Collections.Generic;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class SqlServerHealthCheckBuilderExtensions
```
rynowak: I think it's worth having a section on writing a health check that's distributed as a library. The stuff thats in this class is basically our guidance. 

1. Write a health check as a standalone class - you can rely on DI, type activation, or named options to access configuration data
2. Write an extension method with parameters like the below to wire it up

aspnetcore/host-and-deploy/health-checks/samples/2.x/HealthChecksSample/Services/StartupHostedService.cs
```diff
@@ -0,0 +1,46 @@
...
+        private readonly StartupHostedServiceHealthCheck _startupHostedServiceHealthCheck;
+
+        public StartupHostedService(ILogger<StartupHostedService> logger, 
+            StartupHostedServiceHealthCheck startupHostedServiceHealthCheck)
+        {
+            _logger = logger;
+            _startupHostedServiceHealthCheck = startupHostedServiceHealthCheck;
+        }
+
+        public Task StartAsync(CancellationToken cancellationToken)
```
rynowak: I think it's worth a comment in here that this code exists for sample purposes, and to **simulate** the effect of a long-running startup job.

aspnetcore/host-and-deploy/health-checks/samples/2.x/HealthChecks.SqlServer/DependencyInjection/SqlServerHealthCheckBuilderExtensions.cs
guardrex: Sorry ... I should've explained ... this is only here because I don't have a BeatPulse package. I took their code and just jammed it into the sample temporarily. None of the code under `HealthChecks.SqlServer` will be in the final sample ... just a package reference to BeatPulse.

aspnetcore/host-and-deploy/health-checks/samples/2.x/HealthChecksSample/Services/StartupHostedService.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            _logger.LogInformation($"Startup Background Service is starting.");
+
+            Task.Delay(_delaySeconds * 1000)
+                .ContinueWith(_ => 
+                    { _startupHostedServiceHealthCheck.StartupTaskCompleted = true; });
```
rynowak: You could also do something like:
```
Task.Run(async () =>
{
    // This is intended to simulate the effect of a long-running startup task.
    await Task.Delay(_delaySeconds * 1000);

    _startupHostedServiceHealthCheck.StartupTaskCompleted = true;
});
```

This is a little more idiomatic.

guardrex: humm ... I thought that doing it that way would block the whole app from starting. Maybe not ... I trust *you* a LOT more than I trust *silly green dinosaurs*. lol

guardrex:
Ok ... so my Item 1 is generally ok (with your fixes :point_up:).

What about Item 2 ...

https://github.com/aspnet/Docs/pull/8959#issuecomment-432238516

Thus far, BeatPulse will be the sample for SqlServer.

Do you want BeatPulse to totally take over for the disk space check for the metric health check.

Or, are we just mentioning in passing the BeatPulse stuff?

aspnetcore/host-and-deploy/health-checks/samples/2.x/HealthChecksSample/Services/StartupHostedService.cs
rynowak: yeah :) it returns a task, but doesn't block itself. https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.run?view=netframework-4.7.2#System_Threading_Tasks_Task_Run_System_Func_System_Threading_Tasks_Task__

guardrex:
@rynowak For the best user experience, this topic requires the BeatPulse RTM package to be published to NuGet (for the SQL Server check, which is the one here that depends on BeatPulse). They probably require 2.2 RTM. As it stands, I'll mark this WIP until we get that package. You can review the latest updates (keep **_RexHacks_**:tm: out of the docs!) sooner or later ........ later or sooner ....... lol .......

martha

> Oh, don't 'ee mind him, my lovelies. **_Sooner or later_** he'll be sorry. **_Later or sooner_**, he'll regret. Well, I'll be going now, Professor. You can tell Doctor Fendelman I'll come back when the others is gone. I don't hold with the likes of 'ee. There isn't a dog born that as attack me, boy. They got more sense than most people.

&ndash; Martha Tyler ([Daphne Heard](https://en.wikipedia.org/wiki/Daphne_Heard)) [Dr. Who: Image of the Fendahl](http://www.bbc.co.uk/doctorwho/classic/episodeguide/imageoffendahl/detail.shtml) &copy;1977 BBC

guardrex:
@rynowak This is making me rich 🤑 or crazy 😵 ..... *probably both*. 😀 lol

guardrex:
@unaizorrilla Hello --- quick ? ...

Will you be releasing a preview package of `HealthChecks.SqlServer` on NuGet?

If so, do you know when you will release a preview package?

I'd like to tie it into the sample app and drop the class lib code that I've embedded whenever you have such a package available on NuGet. No giant rush ... just wondering about a preview package that I might use. If there will be no preview package from you, that's ok ... I'll just wait until RTM and you have published your package to finalize the topic+sample app.

guardrex:
No problem @unaizorrilla ... and no rush. Was just wondering if you planned to do something like that by the end of the year.

unaizorrilla:
Hi @guardrex 

Build and publishing packages are done automatically on appveyor, but I've needed change the packages names because are already used on nuget for other projects ( from 9 days ago :-( ). I changed the package names to AspNetCore.HealthChecks.XXX -> where XXX is SqlServer, Redis etc 

**Could you update docs to rename HealthChecks.SqlServer package name to AspNetCore.HealthChecks.SqlServer?**

guardrex:
@Rick-Anderson @scottaddie @tdykstra Pain point here with that content validation system. It's choking this PR on the PR template.

Would you like me to add an H1 to the PR template?

> Annotations
>
> Check warning on line 1 in .github/PULL_REQUEST_TEMPLATE.md
>
> @docs-content-validation docs-content-validation / Commit Validation
>
> .github/PULL_REQUEST_TEMPLATE.md#L1
>
> H1 is required. Use a single hash (#) followed by a space to create your top-level heading. See http://aka.ms/MarkdownHeadings for more information.

guardrex:
@rynowak Updates using 2.2.0 packages are in. Sample app seems to be running well. I think it's ready for your final checks to see if I have the post-preview3 reversions correct.

btw - Do I have the default status code assignments correct, or how do you want to set these now?

```csharp
ResultStatusCodes =
{
    [HealthCheckStatus.Healthy] = StatusCodes.Status200OK,
    [HealthCheckStatus.Degraded] = StatusCodes.Status200OK,
    [HealthCheckStatus.Unhealthy] = StatusCodes.Status503ServiceUnavailable
}
```

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-2.2&branch=pr-en-us-8959)

Rick-Anderson:
[IETF](https://www.ietf.org/assignments/http-status-codes/http-status-codes.xml) needs to add more status codes so you can different `HealthCheckStatus.Healthy` from `Degraded`

guardrex:
@scottaddie This should only need a final quick pass by @rynowak to confirm that my changes for the post-preview3 updates are good. I think we can go ahead with our review.

guardrex:
@scottaddie Don't sweat the review. We're going to go ahead and merge this at EOD. I'm going to UE pass it again now, and we can take another look at it later for anything I don't catch on my pass today.

@rynowak If you can look at the updates, :+1:. If not, we can catch them asap after publication. I'm only concerned about the last commit, which touches the bits for the post-preview3 updates. The sample runs with those updates, so I have some confidence that we're ok.

aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,614 @@
...
+* Use of memory, disk, and other physical server resources can be monitored for healthy status.
+* Health checks can test an app's dependencies, such as databases and external service endpoints, to confirm availability and normal functioning.
+
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/host-and-deploy/health-checks/samples) ([how to download](xref:tutorials/index#how-to-download-a-sample))
+
+The sample app includes examples of the scenarios described in this topic. To run the sample app for a given scenario, use the [dotnet run](/dotnet/core/tools/dotnet-run) command from the project's folder in a command shell. See the sample app's *README.md* file and the scenario descriptions in this topic for details on how to use the sample app.
+
+## Prerequisites
+
+Health checks are usually used with an external monitoring service or container orchestrator to check the status of an app. Before building a production health check app, decide on which monitoring system to use. The monitoring system dictates what types of health checks to create and how to configure their endpoints.
```
rynowak: > building a production health check app

This could be misleading... I'd rather see something here like "before adding health checks to your app..."

aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,614 @@
...
+
+## Basic health probe
+
+For many apps, a basic health probe configuration that reports the app's availability to process requests (*liveness*) is sufficient to discover the status of the app.
+
+The basic configuration registers health check services and calls the Health Check Middleware to respond at a URL endpoint with a health response. By default, no specific health checks are registered to test any particular dependency or subsystem. The app is considered healthy if it's capable of responding at the health endpoint URL. The default response writer writes the status (`HealthCheckStatus`) as a plaintext response back to the client, indicating either a `HealthCheckResult.Healthy` or `HealthCheckResult.Unhealthy` status.
+
+Register health check services with `AddHealthChecks` in `Startup.ConfigureServices`. Call Health Check Middleware in the app processing pipeline in `Startup.Configure`.
```
rynowak: The statement "Call Health Check Middleware" is really odd.

You don't call a middleware, they **get called** by requests

guardrex: This might be a slightly broader problem. We do often say 'add to the middleware processing pipeline,' but we may have similar language to what I have here in other topics. I've made a note to investigate that separately. I'll update this on the next commit.

aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,614 @@
...
+
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
+    services.AddHealthChecks()
+        .AddCheck<ExampleHealthCheck>("example_health_check");
+}
+```
+
+An `AddCheck` overload provides the opportunity to set the failure status (`HealthStatus`) that should be reported when the health check reports a failure. If the failure status is set to `null` (default), `HealthStatus.Unhealthy` is reported.
```
rynowak: Seeing these two examples next to each other is a little confusing. The failure status that's passed in is a recommendation that the check **should** return it. This is a feature for library authors, and requires the library to honor it.. This should be called out somwehre with a note.

aspnetcore/host-and-deploy/health-checks/samples/2.x/HealthChecksSample/DBHealthStartup.cs
```diff
@@ -0,0 +1,52 @@
+﻿using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using HealthChecks.SqlServer;
```
unaizorrilla: AddSqlServer extension method is on Microsoft.Extensions.DependencyInjection, this namespace entry is not necessary

aspnetcore/host-and-deploy/health-checks.md
guardrex: Thanks @unaizorrilla ... I'll get it on the next commit.

guardrex:
@rynowak I *think* I follow you on your `AddCheck` overload example comment. I strengthen the existing remark and call out how it might be helpful for lib authors. Is the following correct?

> The `AddCheck` overload shown in the following example sets the failure status (`HealthStatus`) to report when the health check reports a failure. If the failure status is set to `null` (default), `HealthStatus.Unhealthy` is reported. This overload is a useful scenario for library authors, where the failure status indicated by the library is enforced by the app when a health check failure occurs.

aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,614 @@
...
+To run the database probe scenario using the sample app, execute the following command from the project's folder in a command shell:
+
+```console
+dotnet run --scenario db
+```
+
+> [!NOTE]
+> [BeatPulse](https://github.com/Xabaril/BeatPulse) isn't maintained or supported by Microsoft.
+
+## Database context
```
rynowak: /cc @divega 

I'm not sure that `Database context` is a term that we use. I think it would be more straightforward to call this `DbContext` health checks for `Entity Framework Core DbContext`..

aspnetcore/host-and-deploy/health-checks.md
```diff
@@ -0,0 +1,614 @@
...
+           try
+           {
+               // Health check logic
+               //
+               // data1 and data2 are used in the method to
+               // run the probe's health check logic.
+
+               return HealthCheckResult.Healthy();
+           }
+           catch (Exception ex)
```
rynowak: We wouldn't recommend that a health check handle all exceptions by default. We'll do this for you.

If you do expect the code you're calling inside the check to throw exceptions, then handle it and return `context.Registration.FailureStatus`.

Either way, you should indictate a failure by returning `context.Registration.FailureStatus` from a library. This allows users of your library to configure the severity of the failure.

aspnetcore/host-and-deploy/health-checks.md
guardrex: Nit: We can't use inline code styles in headings.

The rest of the headings at this level use the word "probe," so we could go with that ...

> Entity Framework Core DbContext probe

guardrex:
@Rick-Anderson Apparently, content validation won't permit the opening H1 (title) to be in an INCLUDE. I guess I'll need to fix it on this PR or else this show is go'in nowhere fast. I'll move those H1s from the INCLUDES into the topics that call for them.

>Check warning on line 12 in aspnetcore/tutorials/razor-pages-mac/page.md
> @docs-content-validation docs-content-validation / Commit Validation
> aspnetcore/tutorials/razor-pages-mac/page.md#L12
> H1 is required. Use a single hash (#) followed by a space to create your top-level heading. See http://aka.ms/MarkdownHeadings for more information.

> Check warning on line 11 in aspnetcore/tutorials/razor-pages-vsc/page.md
> @docs-content-validation docs-content-validation / Commit Validation
> aspnetcore/tutorials/razor-pages-vsc/page.md#L11
> H1 is required. Use a single hash (#) followed by a space to create your top-level heading. See http://aka.ms/MarkdownHeadings for more information.

Rick-Anderson:
@guardrex 	ignore those bogus errors.

guardrex:
@rynowak See the commit before the last. Brace yourself for my 😀 *RexHacks*:tm: 😀.

... and ignore the bad build :x: on that. We're having a little "fun" with the validation this evening.

rynowak:
> This overload is a useful scenario for library authors, where the failure status indicated by the library is enforced by the app when a health check failure occurs

This only works if the health check implementation honors the setting. Nothing is magic.

guardrex:
I'll make it say ...

> ... where the failure status indicated by the library is enforced by the app when a health check failure occurs if the health check implementation honors the setting.

aspnetcore/host-and-deploy/health-checks.md
guardrex: Just to finish off the point :gun: ... "call middleware," i.e., "*call a middleware extension method when setting up the request processing pipeline*" actually **_isn't_** used anywhere in the docs. *Good!* 😅

PR #8960, opened 2018-10-10
update ASP.NET SignalR docs to show observing send errors

anurse:
Related: https://github.com/SignalR/SignalR/issues/4144

FYI: This is for **ASP.NET** SignalR, *not* ASP.NET Core SignalR. (The concepts are relative to both but the newer docs are better at showing the right pattern ;))

When SignalR fails to send a message (possibly due to an error in the backplane), the exception bubbles all the way up to the Client proxy (e.g. `Clients.All.foo()`). However, all our docs tell users to simply invoke these methods without `await`ing the task or otherwise observing it. This means that these errors can end up triggering the [`TaskScheduler.UnobservedTaskException`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskscheduler.unobservedtaskexception?view=netframework-4.7.2) event.

This updates our docs on the Hubs API to show `await`ing the Client proxy invocation.

[Internal Review Version](https://review.docs.microsoft.com/en-us/aspnet/signalr/overview/guide-to-the-api/hubs-api-guide-server?branch=pr-en-us-8960)

### DIFF
--- a/aspnet/signalr/overview/guide-to-the-api/hubs-api-guide-server.md
+++ b/aspnet/signalr/overview/guide-to-the-api/hubs-api-guide-server.md
@@ -569,10 +569,9 @@ In VB.NET or in a strongly-typed hub, the caller state object can't be accessed 
 
 ## How to handle errors in the Hub class
 
-To handle errors that occur in your Hub class methods, use one or more of the following methods:
+To handle errors that occur in your Hub class methods, first ensure you "observe" any exceptions from async operations (such as invoking client methods) by using `await`. Then use one or more of the following methods:
 
 - Wrap your method code in try-catch blocks and log the exception object. For debugging purposes you can send the exception to the client, but for security reasons sending detailed information to clients in production is not recommended.
-- Use `await` when invoking client methods to ensure that any exceptions that occur while trying to send the message are observed and can be caught in a try-catch block.
 - Create a Hubs pipeline module that handles the [OnIncomingError](https://msdn.microsoft.com/library/microsoft.aspnet.signalr.hubs.hubpipelinemodule.onincomingerror(v=vs.111).aspx) method. The following example shows a pipeline module that logs errors, followed by code in Startup.cs that injects the module into the Hubs pipeline.
 
     [!code-csharp[Main](hubs-api-guide-server/samples/sample61.cs)]
--- a/aspnet/signalr/overview/guide-to-the-api/hubs-api-guide-server.md
+++ b/aspnet/signalr/overview/guide-to-the-api/hubs-api-guide-server.md
@@ -298,6 +298,9 @@ To call client methods from the server, use the `Clients` property in a method i
 
 [!code-csharp[Main](hubs-api-guide-server/samples/sample23.cs?highlight=5)]
 
+> [!NOTE]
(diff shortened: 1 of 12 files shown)


aspnet/signalr/overview/guide-to-the-api/hubs-api-guide-server.md
```diff
@@ -298,6 +298,9 @@ To call client methods from the server, use the `Clients` property in a method i
 
 [!code-csharp[Main](hubs-api-guide-server/samples/sample23.cs?highlight=5)]
 
+> [!NOTE]
+> Invoking a client method is an asynchronous operation and returns a `Task`. Use `await` to ensure you wait for the message to be sent. If an error occurs while sending the message, `await` will also allow you to catch that exception using a try-catch block.
```
mikaelm12: You could maybe link to the appropriate API docs here. Just a thought.

aspnet/signalr/overview/guide-to-the-api/hubs-api-guide-server.md
```diff
@@ -569,6 +572,7 @@ In VB.NET or in a strongly-typed hub, the caller state object can't be accessed
 To handle errors that occur in your Hub class methods, use one or more of the following methods:
 
 - Wrap your method code in try-catch blocks and log the exception object. For debugging purposes you can send the exception to the client, but for security reasons sending detailed information to clients in production is not recommended.
+- Use `await` when invoking client methods to ensure that any exceptions that occur while trying to send the message are observed and can be caught in a try-catch block.
```
tdykstra: The lead-in to the bullets says "use one or more" of these methods, which implies that any one of them is optional. If only the try-catch method depends on use of await, you could incorporate the await text into that bullet instead of making it a separate one.  If await is always required for all of these methods, it might be better to take it out of the bullet list and say that all methods depend on it.

aspnet/signalr/overview/guide-to-the-api/hubs-api-guide-server.md
anurse: Yeah, I actually did some searching to try and find appropriate docs on observing exceptions and didn't really like what I found.

anurse: https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/exception-handling-task-parallel-library - Doesn't really cover `async`/`await`

https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskscheduler.unobservedtaskexception?view=netframework-4.7.2 - Very light on narrative descriptions

https://blogs.msdn.microsoft.com/pfxteam/2011/09/28/task-exception-handling-in-net-4-5/ - I trust the source a lot but it's not really a "doc"

Rick-Anderson: > Yeah, I actually did some searching to try and find appropriate docs on observing exceptions and didn't really like what I found.

@BillWagner 	can you suggest a link?

Invoking a client method is an asynchronous operation and returns a `Task`. Use `await` :
* To ensure the message is sent without error. 
* To enable catching and handling errors.

I think most folks know  you'd need a try/catch.

anurse:
@Rick-Anderson sent a PR to this branch. Once that's in, can we get this merged?

PR #8976, opened 2018-10-11
Update filters.md

### LINKED ISSUE #8958: Clarify threading/lifetime behaviours of TypeFilterAttribute and ServiceFilterAttribute
We should add the relevant information to address this users' feedback: https://github.com/aspnet/Mvc/issues/8579

When using `TypeFilterAttribute` - setting `IsReusable` is a hint that the filter instance *may* be reused outside of the request scope it was created within. The framework provides no guarantees that a single instance of the filter will be created. Avoid using `IsReusable` when using a filter that depends on services with a lifetime other than singleton.

When using `ServiceFilterAttribute` -  setting `IsReusable` is a hint that the filter instance *may* be reused outside of the request scope it was created within. The framework provides no guarantees that a single instance of the filter will be created or the filter will not be re-requested from the DI container at some later point. Avoid using `IsReusable` when using a filter that depends on services with a lifetime other than singleton.

If you need to ensure that a single instance of a filter is created, you can implement `IFilterFactory` yourself and provide whatever policy meets your requirements.

Rick-Anderson:
Fixes #8958


aspnetcore/mvc/controllers/filters.md
```diff
@@ -199,12 +199,14 @@ If your filters have dependencies that you need to access from DI, there are sev
 
 ### ServiceFilterAttribute
 
-A `ServiceFilter` retrieves an instance of the filter from DI. You add the filter to the container in `ConfigureServices`, and reference it in a `ServiceFilter` attribute
+A `ServiceFilter` retrieves an instance of the filter from DI. Add the `ServiceFilter` to the container in `ConfigureServices`, and reference it in a `ServiceFilter` attribute:
```
rynowak: > add the `ServiceFilter`  

I think is text is misleading. You register the filter **implementation type** in DI. 

Also, should this say `ServiceFilterAttribute` in all of the places it says `ServiceFilter`? See L208

aspnetcore/mvc/controllers/filters.md
```diff
@@ -199,12 +199,14 @@ If your filters have dependencies that you need to access from DI, there are sev
...
 
 [!code-csharp[](./filters/sample/src/FiltersSample/Startup.cs?name=snippet_ConfigureServices&highlight=11)]
 
 [!code-csharp[](../../mvc/controllers/filters/sample/src/FiltersSample/Controllers/HomeController.cs?name=snippet_ServiceFilter&highlight=1)]
 
+When using `ServiceFilterAttribute`,  setting `IsReusable` is a hint that the filter instance *may* be reused outside of the request scope it was created within. The framework provides no guarantees that a single instance of the filter will be created or the filter will not be re-requested from the DI container at some later point. Avoid using `IsReusable` when using a filter that depends on services with a lifetime other than singleton.
```
rynowak: 👍

PR #9006, opened 2018-10-13
attribute-routing-in-web-api-2, RouteOrder=>Order

athamala:
Correct property name for route ordering in RouteAttribute is Order, not RouteOrder. Replace incorrect RouteOrder property name with correct name Order

source references:
https://github.com/aspnet/Mvc/blob/master/src/Microsoft.AspNetCore.Mvc.Core/RouteAttribute.cs#L40
https://github.com/ASP-NET-MVC/aspnetwebstack/blob/master/src/System.Web.Http/RouteAttribute.cs#L40


scottaddie:
Thank you for fixing this, @athamala! Your changes will go live within the next 24 hours.

PR #9022, opened 2018-10-15
Typo in version-differences

agriffard:
Adding comma

  Fixes #Issue_Number


guardrex:
@tdykstra Would put an :eye: on the update and see what you think? "Now" seems temporal, but the doc seems to generally distinguish ASP.NET SignalR vs. ASP.NET Core SignalR.

If you prefer the way you wrote it, let me know, and I'll revert.

I'm also concerned on the side about conjunctive adverbs + subordinate phrases. It's not an issue if we continue to shorten introductory phrases (i.e., remove "Now" here). Anyway ... that's a different discussion. Removing "Now" resolves the problem (if it is a problem :smile:).

PR #9036, opened 2018-10-15
Update crud.md with fixed typo

RyanLilla:
Found a typo in the `FindAsync` section where "they" was used instead of "that".

Currently: "Can have perf benefits in some situations, but **they** rarely happens for typical web apps."


Rick-Anderson:
@RyanLilla 	 Thanks!

PR #9039, opened 2018-10-15
Update sort-filter-page.md with proper path names

RyanLilla:
Sections in the document where "**Student**/Index.cshtml" was used instead of the correct, plural form like so: **Students**/Index.cshtml


Rick-Anderson:
@RyanLilla 	thanks!

PR #9042, opened 2018-10-16
Link to MVC and Razor page route conventions

JamesNK:


Rick-Anderson:
@JamesNK 	thanks. I edited your updates to remove future tense and replace long sentences with bullet list. Sorry I messed up on commit so the commit diff doesn't help.

PR #9045, opened 2018-10-16
Uncommented EnsureCreated() in DbInitializer.cs

RyanLilla:
In [DbInitializer.cs](https://github.com/aspnet/Docs/blob/master/aspnetcore/data/ef-rp/intro/samples/cu21/Data/DbInitializer.cs) the line of code for `context.Database.EnsureCreated();` is currently commented out. 

This doesn't seem to be an issue if simply copy-pasting the code samples throughout the tutorial, but the reader could potentially run into an error if the database is ever dropped as there would not be a check for an existing database. 

Also, further in the tutorial (see [here](https://docs.microsoft.com/en-us/aspnet/core/data/ef-rp/migrations?view=aspnetcore-2.1&tabs=visual-studio#create-an-initial-migration-and-update-the-db)), the reader is then asked to remove this line as they would have setup the use of migrations. This is a bit confusing as the line would have been commented out thus far anyway, so deleting the comment seems trivial.

### DIFF
--- a/aspnetcore/data/ef-rp/intro/samples/cu21/Data/DbInitializer.cs
+++ b/aspnetcore/data/ef-rp/intro/samples/cu21/Data/DbInitializer.cs
@@ -12,7 +12,7 @@ namespace ContosoUniversity.Data
     {
         public static void Initialize(SchoolContext context)
         {
-            // context.Database.EnsureCreated();
+            context.Database.EnsureCreated();
 
             // Look for any students.
             if (context.Student.Any())


Rick-Anderson:
@RyanLilla 	I'd prefer to leave it commented out so the code works later in the tutorial when you're supposed to comment it out. We usually don't have code for each stage of the tutorial - but we've got some at https://github.com/aspnet/Docs/tree/master/aspnetcore/data/ef-rp/intro/samples/21StageSnapShots

Any changes would need to be made in the stage snapshots.

Rick-Anderson:
@tdykstra 	I think we should try to have stage snapshots for the first 4 stages when we update to 2.2
If they become too much work to update we can delete them.

PR #9048, opened 2018-10-16
Updated help reference

daveabrock:
I was reviewing an issue - to me, it looks like the VS Code help reference was a bit out of place. Having it up front right under the prerequisites might be a better spot.


Rick-Anderson:
@daveabrock 	thanks!

PR #9056, opened 2018-10-16
Add information on connection token vs other types

### LINKED ISSUE #9026: SignalR Security Docs - Clarify cookie vs querystring token
Hi all, I'm willing to submit a PR for this to update the docs if I can get a clarified understanding.

# Issue description

I'm in the process of working with our client's security team on mitgiating some findings of their automated security tool, which reported a finding because of a "session token reported in URL".

The [Introduction to SignalR Security doc](https://github.com/aspnet/Docs/blob/master/aspnet/signalr/overview/security/introduction-to-security.md) states:

> SignalR passes the connection token as a query string value, instead of as a cookie. Storing the connection token in a cookie is unsafe because the browser can inadvertently forward the connection token when malicious code is encountered. Also, passing the connection token in the query string prevents the connection token from persisting beyond the current connection. Therefore, a malicious user cannot make a request under another user's authentication credentials.

The security team responded with this point, which I think is a fair one:

> If malicious code is encountered it would be trivial (and much easier) to simply extract the token out of the URL. Further - with the token in the URL it will show up in logs and in referer headers (as described in the report).  There are protections you can employ on cookies to prevent malicious access and from it being sent where it shouldn't be (httponly and path flags). 

This point seems reasonable, so I wonder if there's something that I'm missing here conceptually. Is it that the token isn't truly a session token? How does SignalR protect against someone impersonating a user account and re-using a token? Is the user verification more robust than I understand? 

# Software versions

Check the .NET target framework(s) being used, and include the version number(s).

- [x] .NET Framework 4.x (seems independent of minor version number)

SeanKilleen:
Fixes #9026. Based on @anurse's comments in the discussion thread, which I attempted to extract / simplify.

Updates language to disambiguate connection token vs session / auth tokens.


SeanKilleen:
@anurse may want to look at this to see if I captured his comments in a reasonable way.

aspnet/signalr/overview/security/introduction-to-security.md
```diff
@@ -80,6 +80,13 @@ For each request, the server validates the contents of the token to ensure that
...
+
+The connection token is not a "session" token here, as it provides no authentication claim of it's own. Taking a user A's connection token and replaying it in a request authenticated as user B (or an unauthenticated request) will fail, because the user identity of the request and the identity stored in the token will not match.
```
anurse: I suggest a bit of a rephrase of the first sentence:

Since the connection token provides no authentication claim of it's own, it is not considered a "session" or "authentication" token.

aspnet/signalr/overview/security/introduction-to-security.md
```diff
@@ -80,6 +80,13 @@ For each request, the server validates the contents of the token to ensure that
...
 
+### Note: Connection Tokens vs Other Types of Tokens
+Connection tokens are occasionally flagged by security tools because they appear to potentially be session tokens or authentication tokens, which would pose a risk if exposed.
+
+In SignalR, the The Connection Token is not an authentication token. It is used to confirm that the user making this request is the same one that created the connection. This is needed because ASP.NET SignalR allows connections to move between servers. The connection token contains the "connection ID" and the user name of the user that created the connection, and is encrypted with a key known only to the servers. The server that initialized the connection checks the user ID provided by your authentication mechanism and embeds it in this token so that if a request is made to a different server, it can verify that the connection ID the client presents is valid and was created on behalf of this user. However the token itself makes no claim that the request was made by that user, only that the connection ID contained within the token is associated with that user.
```
anurse: the The Connection Token

->

the connection token

--

extra "the", and connection token isn't a proper noun here, so we don't need Title Case.

anurse: This is needed because ASP.NET SignalR allows connections to move between servers.

->

The connection token is necessary because ASP.NET SignalR allows connections to move between servers.

anurse: > The connection token contains the "connection ID" and the user name of the user that created the connection, and is encrypted with a key known only to the servers. The server that initialized the connection checks the user ID provided by your authentication mechanism and embeds it in this token so that if a request is made to a different server, it can verify that the connection ID the client presents is valid and was created on behalf of this user.

I think we can actually simplify this a little:

The token associates the connection with a particular user but doesn't assert the identity of the user making the request. For a SignalR request to be properly authenticated, it must have some other token (Cookie, Bearer Token, etc.) that asserts the identity of the user.

SeanKilleen:
@anurse great feedback, thanks; will incorporate those suggestions. And for the record, I thought the comment very closely mapped to what a doc could be; the rough draft was great! 😄

aspnet/signalr/overview/security/introduction-to-security.md
```diff
@@ -80,6 +80,13 @@ For each request, the server validates the contents of the token to ensure that
 
 Because the connection id is part of the verification process, you should not reveal one user's connection id to other users or store the value on the client, such as in a cookie.
 
+### Note: Connection Tokens vs Other Types of Tokens
```
tdykstra: A couple of thoughts on the heading:
* Our style guide prescribes sentence casing for headings -- "... vs. other types of tokens"
* Consider making it H4 as it logically fits under the preceding Connection token heading.
* Consider dropping "Note: "

aspnet/signalr/overview/security/introduction-to-security.md
tdykstra: Could drop "In SignalR, " at start of paragraph

"This is needed because ASP.NET Core SignalR ..." (add "Core")

etc. is not allowed by our style guide -- one possible way to rephrase: "... some other token that asserts the identity of the user, such as a cookie or bearer token."

aspnet/signalr/overview/security/introduction-to-security.md
tdykstra: "... of its own ..." (no apostrophe)

"Taking user A's ... " (no "a")

aspnet/signalr/overview/security/introduction-to-security.md
anurse: This isn't a doc for ASP.NET Core SignalR, so don't add "Core"! :)

SeanKilleen:
@anurse @tdykstra I believe I've incorporated requested feedback & tweaks. 👍

aspnet/signalr/overview/security/introduction-to-security.md
```diff
@@ -80,6 +80,13 @@ For each request, the server validates the contents of the token to ensure that
 
 Because the connection id is part of the verification process, you should not reveal one user's connection id to other users or store the value on the client, such as in a cookie.
 
+#### Connection Tokens vs other types of tokens
```
scottaddie: - Change the heading to "Connection tokens vs. other token types"
- Add a line break after the heading

aspnet/signalr/overview/security/introduction-to-security.md
```diff
@@ -80,6 +80,13 @@ For each request, the server validates the contents of the token to ensure that
...
+Connection tokens are occasionally flagged by security tools because they appear to potentially be session tokens or authentication tokens, which would pose a risk if exposed.
```
scottaddie: - ~~potentially~~
- would pose --> poses

aspnet/signalr/overview/security/introduction-to-security.md
```diff
@@ -80,6 +80,13 @@ For each request, the server validates the contents of the token to ensure that
...
+
+SignalR's connection token is not an authentication token. It is used to confirm that the user making this request is the same one that created the connection. The connection token is necessary because ASP.NET SignalR allows connections to move between servers. The token associates the connection with a particular user but doesn't assert the identity of the user making the request. For a SignalR request to be properly authenticated, it must have some other token that asserts the identity of the user, such as a cookie or bearer token. However, the connection token itself makes no claim that the request was made by that user, only that the connection ID contained within the token is associated with that user.
```
scottaddie: - is not --> isn't
- it is --> it's
- ASP.NET SignalR --> ASP.NET Core SignalR

aspnet/signalr/overview/security/introduction-to-security.md
```diff
@@ -80,6 +80,13 @@ For each request, the server validates the contents of the token to ensure that
...
+
+Since the connection token provides no authentication claim of its own, it is not considered a "session" or "authentication" token. Taking a given user's connection token and replaying it in a request authenticated as a different user (or an unauthenticated request) will fail, because the user identity of the request and the identity stored in the token will not match.
```
scottaddie: - is not --> isn't
- will not --> won't

SeanKilleen:
@scottaddie I appreciate the input but I have to say that those really seem like style choices. Are there actual style rules around those somewhere? I happen to be fine with making them but I'd be surprised if those changes needed to hold up merging this PR.

Some questions below:

> Change the heading to "Connection tokens vs. other token types". Also make it a level 3 heading instead of level 4.

It previously was a level 3 heading and I changed it to a level 4 because I was asked to do so, as it was considered to be a sub-section of the larger point. Is there an MS tie-breaker somewhere around? :)

> ASP.NET SignalR --> ASP.NET Core SignalR

I was also asked specifically not to do this, as this document is not for Core SignalR. See [@anurse's comment](https://github.com/aspnet/Docs/pull/9056#discussion_r225995494) and [@tdykstra's confirmation](https://github.com/aspnet/Docs/pull/9056#discussion_r226003473).

scottaddie:
@SeanKilleen You're correct that these are stylistic concerns. Thank you for asking about it. We're supposed to abide by the [official style guide](https://docs.microsoft.com/style-guide/welcome/) whenever possible. Here are the official guidelines for a couple of the changes I recommended:

* "vs." in headings: https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/v/versus-vs
* contractions: https://docs.microsoft.com/en-us/style-guide/word-choice/use-contractions

I missed the comments from Andrew and Tom. Apologies for that. Please disregard those 2 comments.

SeanKilleen:
@scottaddie thanks for pointing out the style guide sections to me! Seriously I appreciate it as I wasn't able to comb the whole thing. I'll make the corrections and ping you when complete. :+1:

SeanKilleen:
@scottaddie I believe I've addressed the comments; please give it your 👀 when you're able. Thanks!

scottaddie:
@SeanKilleen Thank you for your patience and for all your great work on this PR!

PR #9057, opened 2018-10-16
Fix typo in ValidationSummary

agriffard:


scottaddie:
Thank you for fixing this, @agriffard!

PR #9060, opened 2018-10-16
Testing on different environments

### LINKED ISSUE #8917: Testing on different environment
Hi,
Nicely mentioned on how to test locally. However, if I need automation tests to run in different environments, can it be done ? My concern is that I am referring the API project to my integration library and then running it locally. Will it work once I deploy my API to an environment and run integration tests against it ? 
If so, how can I implement that ?

---
#### Document Details

* Content: [Integration tests in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-2.1)
* Content Source: [aspnetcore/test/integration-tests.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/test/integration-tests.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #8917 

cc: @seshenoy

### DIFF
--- a/aspnetcore/test/integration-tests.md
+++ b/aspnetcore/test/integration-tests.md
@@ -204,17 +204,11 @@ If the SUT requires special testing configuration when tests are executed:
     When configuring an `IWebHost` in *Program.cs*:
     
     ```csharp
-    public static IHostingEnvironment Environment { get; set; }
-    
     public static IWebHostBuilder BuildWebHostBuilder(string[] args) =>
         WebHost.CreateDefaultBuilder(args)
-            .ConfigureAppConfiguration((hostingContext, config) =>
+            .ConfigureServices((hostContext, services) =>
             {
-                Environment = hostingContext.HostingEnvironment;
-            })
-            .ConfigureServices(services =>
-            {
-                if (Environment.IsStaging())
+                if (hostContext.HostingEnvironment.IsStaging())
                 {
                     // Code runs in the Staging environment
                 }
--- a/aspnetcore/test/integration-tests.md
+++ b/aspnetcore/test/integration-tests.md
@@ -5,7 +5,7 @@ description: Learn how integration tests ensure that an app's components functio
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 05/30/2018
+ms.date: 10/16/2018
 uid: test/integration-tests
 ---
 # Integration tests in ASP.NET Core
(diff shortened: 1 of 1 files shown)


aspnetcore/test/integration-tests.md
```diff
@@ -184,6 +184,71 @@ Because another test in the `IndexPageTests` class performs an operation that de
...
+### Set the environment
+
+If the SUT requires special testing configuration when tests are executed:
+
+* Provide test environment-specific configuration using one or both of the following approaches:
+  * Provide conditional configuration in the SUT's `Startup.ConfigureServices`/`Startup.Configure`:
+  
+    When configuring an `IWebHostBuilder` in *Startup.cs*:
+    
+    ```csharp
```
javiercn: Where is this code snippet, in configure services?

guardrex: Yes ... and given `IHostingEnvironment` examples we have are mostly shot into `Configure`, I'm not sure this is the best way to demo this. Should I go in the direction of a `Configure`-based injection and change this to ...

```csharp
{
    if (env.IsStaging())
    {
        // Code runs in the Staging environment
    }
}
```

javiercn:
Can you provide an internal review topic? It's really challenging to review these changes in isolation

guardrex:
Yes, sorry about that @javiercn ...

https://review.docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-2.1&branch=pr-en-us-9060

guardrex:
@javiercn It was the first thing that came to mind: The app loads whatever services and pipeline config it requires for whatever environment it finds itself in depending on its environment (dev, staging, test, prod) ... in the normal way ... in the way that we clearly document in the *App startup* and *Multiple environments* topics, then the dev sets the environment for testing using the env var or by setting it in code in the test project the way that I show. That's super-easy to understand and comports well with what we current document about running apps in various environments.

> The pattern mentioned above (calling plain ConfigureServices and using services.TryAdd in your startup code is the recommended approach to follow when you want to register services that you might need to override for testing).

Even now just reading that, it's not super clear to me, so I'll think it over this weekend.

javiercn:
@guardrex Also for environment specific setups you can do Configure<<Environment>>Services in your startup and setup the environment in the WebHostBuilder. Those two approaches are better than relying on ConfigureTestServices

guardrex:
WRT `Configure{Environment}Services` ... we do document it, but I'm concerned about showing devs too many ways of handling environments (i.e., not being opinionated enough). The current thrust of doc samples and content is to use `if` logic. Actually, I have a better idea for this section now that we've discussed it. Let's just punt the reader (via link) over to the *Multiple environments* topic for that bit. We don't need it here. We just need to point to what we currently have.

WRT controlling the environment for tests, are you just saying to strip out the piece that explains how to set it explicitly in `ConfigureWebHost` and `WithWebHostBuilder`? ... just tell the reader to set the environment on the testing server with the env var and leave it at that? If so, that's easy. Then if they prefer to use `ConfigureWebHost` and/or `WithWebHostBuilder` approaches, they're welcome to discover them on their own (i.e., IntelliSense).

guardrex:
@javiercn How about this (on the last commit) ...

> \#\#\# Set the environment
>
> If the SUT requires special testing configuration when tests are executed:
>
> \* Specify the environment on the test system (for example, set the \`ASPNETCORE_ENVIRONMENT\` environment variable to a value that reflects a staging or testing environment, such as \`Staging\` or \`Test\`).
> \* Provide code in \[Startup](xref:fundamentals/startup) that loads services and configures the request pipeline based on its environment.
>
> For more information, see \<xref:fundamentals/environments>.

javiercn:
> Specify the environment on the test system (for example, set the `ASPNETCORE_ENVIRONMENT` environment variable to a value that reflects a staging or testing environment, such as `Staging` or `Test`).

Why would you do this instead of just calling webhostbuilder.UseEnvironment?

guardrex:
> Why would you do this instead of just calling webhostbuilder.UseEnvironment?

Wouldn't calling `UseEnvironment` be hard-coding the environment into the app? ... unless you meant in `CustomWebApplicationFactory` or `WithWebHostBuilder` ... that's what I did on the 1st commit (but I had the impression that you didn't want to mention those approaches).

Suppose that I configure the app based on its environment in `ConfigureServices` and `Configure` based on `Development`, `Staging` (testing), and `Production`. The app reacts to whatever environment it's in. All I have to do is set the environment in the testing environment (i.e., set the `ASPNETCORE_ENVIRONMENT` env var and call `CreateDefaultBuilder`). That seems super-simple to understand.

Alternatively, I could have set the environment for the app based on what I showed in the first commit ... `builder.UseEnvironment("Staging")` (with `CustomWebApplicationFactory`) or `builder.UseSetting("Environment", "Staging");` (`WithWebHostBuilder`). I tested this approach, and it worked.

guardrex:
@javiercn Should I close this PR? I have other issues to address on the topic, so we can move on to the next one if there's no need to address the environment-related issue (#8917).

PR #9063, opened 2018-10-16
Moved section earlier in this tutorial

### LINKED ISSUE #7728: Warning provided too late to do anything about it
At the end of this part of the tutorial, you talk about an optional step that can be used to fix a foreign key constraint with legacy data. That section includes the following warning:

&gt; Don't make these code changes if you completed the previous section and updated the database.

Unfortunately, by the time you get here, if you're following along all of the steps in the tutorial, you have already updated the database, and so there's not really any way for you to proceed with this optional content. Perhaps you could move this section earlier in the tutorial, or call out the optional section higher up in the tutorial, and use anchor links to take learners to the optional section, and then back to where they should be once they have completed that section.

---
#### Document Details

* ID: 634dfb2c-2626-9ec5-35fe-2bd2f27c5e23
* Version Independent ID: 7a77b058-f005-8d68-ae46-e487c597f18e
* Content: [Razor Pages with EF Core in ASP.NET Core - Data Model - 5 of 8](https://docs.microsoft.com/en-us/aspnet/core/data/ef-rp/complex-data-model?view=aspnetcore-2.1&tabs=visual-studio)
* Content Source: [aspnetcore/data/ef-rp/complex-data-model.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/data/ef-rp/complex-data-model.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

daveabrock:
A link to the FK section is actually provided, but it is easy to gloss over. Moving the section up in the document prevents users reading it too late to do anything about it, as said in the issue.

Fixes #7728

### DIFF
--- a/aspnetcore/data/ef-rp/complex-data-model.md
+++ b/aspnetcore/data/ef-rp/complex-data-model.md
@@ -571,45 +571,17 @@ The ALTER TABLE statement conflicted with the FOREIGN KEY constraint "FK_dbo.Cou
 database "ContosoUniversity", table "dbo.Department", column 'DepartmentID'.
 ```
 
-When migrations are run with existing data, there may be FK constraints that are not satisfied with the existing data. For this tutorial, a new DB is created, so there are no FK constraint violations. 
+When migrations are run with existing data, there may be FK constraints that are not satisfied with the existing data. For this tutorial, a new DB is created, so there are no FK constraint violations. See [Apply the migration to the existing database](#applyexisting) for instructions on how to fix the FK violations on the current DB.
 
-For instructions on how to fix the FK violations on the current DB, perform the following steps.
+## Apply the migration
 
-### Optional: Fix foreign key restraints with legacy data
+Now that you have an existing database, you need to think about how to apply future changes to it. For this, you have two approaches:
+* [Drop and re-create the database](#drop)
+* [Apply the migration to the existing database](#applyexisting). While this method more complex and time-consuming, it's the preferred approach for real-world, production environments. 
 
-When migrations are run with existing data, there may be FK constraints that are not satisfied with the existing data. With production data, steps must be taken to migrate the existing data. This section provides an example of fixing FK constraint violations. Don't make these code changes without a backup. Don't make these code changes if you completed the previous section and updated the database.
+<a name="drop"></a>
 
-The *{timestamp}_ComplexDataModel.cs* file contains the following code:
-
-[!code-csharp[](intro/samples/cu/Migrations/20171027005808_ComplexDataModel.cs?name=snippet_DepartmentID)]
-
-The preceding code adds a non-nullable `DepartmentID` FK to the `Course` table. The DB from the previous tutorial contains rows in `Course`, so that table cannot be updated by migrations.
-
-To make the `ComplexDataModel` migration work with existing data:
-
-* Change the code to give the new column (`DepartmentID`) a default value.
-* Create a fake department named "Temp" to act as the default department.
-
-#### Fix the foreign key constraints
-
-Update the `ComplexDataModel` classes `Up` method:
-
-* Open the *{timestamp}_ComplexDataModel.cs* file.
-* Comment out the line of code that adds the `DepartmentID` column to the `Course` table.
-
-[!code-csharp[](intro/samples/cu/Migrations/20171027005808_ComplexDataModel.cs?name=snippet_CommentOut&highlight=9-13)]
-
-Add the following highlighted code. The new code goes after the `.CreateTable( name: "Department"` block:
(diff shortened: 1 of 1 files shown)


daveabrock:
Thanks, @tdykstra! Yes, I agree this one is tough to get right - so thanks for the suggestions. I do like the better approach, which seems to provide more detail and context. I should be able to crank that out tonight.

daveabrock:
Pushed updates, @tdykstra - let me know!

aspnetcore/data/ef-rp/complex-data-model.md
```diff
@@ -571,10 +571,17 @@ The ALTER TABLE statement conflicted with the FOREIGN KEY constraint "FK_dbo.Cou
...
 ```
 
-When migrations are run with existing data, there may be FK constraints that are not satisfied with the exiting data. For this tutorial, a new DB is created, so there are no FK constraint violations. See
-[Fixing foreign key constraints with legacy data](#fk) for instructions on how to fix the FK violations on the current DB.
+When migrations are run with existing data, there may be FK constraints that are not satisfied with the existing data. For this tutorial, a new DB is created, so there are no FK constraint violations. See [Apply the migration to the existing database](#applyexisting) for instructions on how to fix the FK violations on the current DB.
 
-### Drop and update the database
+## Apply the migration
+
+Now that you have an existing database, you need to think about how to apply future changes to it. For this, you have two approaches:
```
tdykstra: > For this, you have two approaches:

Might be a little more accurate to say 

> This tutorial shows two approaches:

aspnetcore/data/ef-rp/complex-data-model.md
```diff
@@ -571,10 +571,17 @@ The ALTER TABLE statement conflicted with the FOREIGN KEY constraint "FK_dbo.Cou
...
+* [Drop and re-create the database](#drop)
+* [Apply the migration to the existing database](#applyexisting). While this method more complex and time-consuming, it's the preferred approach for real-world, production environments.
```
tdykstra: Consider making it clearer that in this tutorial the choices are mutually exclusive:

This is an optional section of the tutorial. You can do the drop and re-create steps and skip this section. If you do want to follow the steps in this section, don't do the drop and re-create steps.

missing "is" before "more complex"

aspnetcore/data/ef-rp/complex-data-model.md
```diff
@@ -571,10 +571,17 @@ The ALTER TABLE statement conflicted with the FOREIGN KEY constraint "FK_dbo.Cou
 database "ContosoUniversity", table "dbo.Department", column 'DepartmentID'.
 ```
 
-When migrations are run with existing data, there may be FK constraints that are not satisfied with the exiting data. For this tutorial, a new DB is created, so there are no FK constraint violations. See
-[Fixing foreign key constraints with legacy data](#fk) for instructions on how to fix the FK violations on the current DB.
+When migrations are run with existing data, there may be FK constraints that are not satisfied with the existing data. For this tutorial, a new DB is created, so there are no FK constraint violations. See [Apply the migration to the existing database](#applyexisting) for instructions on how to fix the FK violations on the current DB.
```
tdykstra: This explanation is given at the start of the "Apply migration to existing database" section, and in the immediately following section you explain the options, so I think we can remove it from here.

aspnetcore/data/ef-rp/complex-data-model.md
```diff
@@ -618,13 +625,11 @@ Examine the **CourseAssignment** table:
...
-<a name="fk"></a>
-
-## Fixing foreign key constraints with legacy data
+<a name="applyexisting"></a>
 
-This section is optional.
+### Apply the migration to the existing database
 
-When migrations are run with existing data, there may be FK constraints that are not satisfied with the exiting data. With production data, steps must be taken to migrate the existing data. This section provides an example of fixing FK constraint violations. Don't make these code changes without a backup. Don't make these code changes if you completed the previous section and updated the database.
+When migrations are run with existing data, there may be FK constraints that are not satisfied with the existing data. With production data, steps must be taken to migrate the existing data. This section provides an example of fixing FK constraint violations. Don't make these code changes without a backup. Don't make these code changes if you completed the previous section and updated the database.
```
tdykstra: Here I'd start off with an explanation that it's optional in case they missed it earlier, something like

> This section is optional. These steps work only if you skipped the preceding **Drop and re-create the database** section.

daveabrock:
Thanks for the suggestions, @tdykstra. All makes sense. Let me know what you think of the latest push.

aspnetcore/data/ef-rp/complex-data-model.md
tdykstra: On re-reading this I have too many "do"s in a row; feel free to reword, e.g., by eliminating "do" before "want" or some other rewording.

PR #9080, opened 2018-10-17
Documentation on WebSocket Origin validation

anurse:
Added some notes on WebSocket Origin validation to ASP.NET Core SignalR documentation.

[Internal Review Link](https://review.docs.microsoft.com/en-us/aspnet/core/signalr/security?view=aspnetcore-2.1&branch=pr-en-us-9080)

### DIFF
--- a/aspnetcore/signalr/security.md
+++ b/aspnetcore/signalr/security.md
@@ -103,7 +103,7 @@ public void Configure(IApplicationBuilder app)
 ```
 
 > [!NOTE]
-> The `Origin` header is completely controlled by the client and should never be used as an authentication mechanism.
+> The `Origin` header is completely controlled by the client and, like the `Referer` header, can be faked. These headers should never be used as an authentication mechanism.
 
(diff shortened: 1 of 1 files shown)


aspnetcore/signalr/security.md
```diff
@@ -53,6 +53,58 @@ public void Configure(IApplicationBuilder app)
 > [!NOTE]
 > SignalR is not compatible with the built-in CORS feature in Azure App Service.
 
+### WebSocket Origin Restriction
```
BrennanConroy: Shouldn't this section be on the [CORS](https://docs.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-2.1) page and we just reference it in the SignalR docs?

BrennanConroy: >It's not CORS! :)

Fine, but people assuming it is will be confused. Maybe CORS doc should reference this when it's in WebSockets

PR #9088, opened 2018-10-17
UE edit of Security considerations in ASP.NET Core SignalR

Rick-Anderson:
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/signalr/security?view=aspnetcore-2.1&branch=pr-en-us-9088)

- Kill long sentences. Long sentences are difficult for native speakers and impossible for MT (machine translation)
- Format code to prevent horizontal scroll bar on tablets 85 characters wide. You can simulate tablets by viewing in a browser of the maximum width that doesn't have left or right panes (TOC's).
- Prune dead wood. Avoid unnecessary words that don't add meaning to the text.
- Avoid "you" except when it make the statement more clear. 

  - For example, the following CORS policy allows a SignalR browser client hosted on `http://example.com` to access **your** SignalR app:
  -For example, the following CORS policy allows a SignalR browser client hosted on `http://example.com` to access the SignalR app hosted on `http://signalr.example.com`:
- Avoid Wikipedia links
- Code snippets

Explicit sign-off required on my security rewording below:
- [ ]  Allow cross-origin requests only from domains you trust or control. 

The following are slightly too long but I couldn't find an elegant way to shorten them (not that my other trimming was elegant):

- Applications should be configured to validate these headers to ensure that only WebSockets coming from the expected origins are allowed.

### DIFF
--- a/aspnetcore/signalr/security.md
+++ b/aspnetcore/signalr/security.md
@@ -63,12 +63,12 @@ Exception messages are generally considered sensitive data that shouldn't be rev
 
 ## Buffer management
 
-SignalR uses per-connection buffers to manage incoming and outgoing messages. By default, SignalR limits these buffers to 32 KB. The largest message a client or server can send is 32 KB. The maximum memory consumed by a connection for messages is 32 KB. If your messages are always smaller than 32 K, you can reduce the limit, which:
+SignalR uses per-connection buffers to manage incoming and outgoing messages. By default, SignalR limits these buffers to 32 KB. The largest message a client or server can send is 32 KB. The maximum memory consumed by a connection for messages is 32 KB. If your messages are always smaller than 32 KB, you can reduce the limit, which:
 
 * Prevents a client from being able to send a larger message.
 * The server will never need to allocate large buffers to accept messages.
 
-If your messages are larger than 32 K, you can increase the limit. Increasing this limit means:
+If your messages are larger than 32 KB, you can increase the limit. Increasing this limit means:
 
 * The client can cause the server to allocate large memory buffers.
 * Server allocation of large buffers may reduce the number of concurrent connections.
--- a/aspnetcore/signalr/security.md
+++ b/aspnetcore/signalr/security.md
@@ -13,7 +13,7 @@ uid: signalr/security
 
 By [Andrew Stanton-Nurse](https://twitter.com/anurse)
 
-SignalR provides a number of security protections by default. It's important to understand how to configure these protections.
+This article provides information on securing SignalR.
 
 ## Cross-origin resource sharing
 
@@ -24,36 +24,15 @@ SignalR provides a number of security protections by default. It's important to 
 
 CORS should be configured in the SignalR app to only allow the origin `www.example.com`.
 
-For more information on configuring CORS, see [Enable Cross-Origin Requests (CORS)](xref:security/cors). SignalR requires the following CORS policies in order to operate correctly:
+For more information on configuring CORS, see [Enable Cross-Origin Requests (CORS)](xref:security/cors). SignalR requires the following CORS policies:
 
-* The policy must allow the specific origins you expect, or allow any origin (not recommended).
+* Allow the specific expected origins. Allowing any origin is possible but is **not** secure or recommended.
 * HTTP methods `GET` and `POST` must be allowed.
-* Credentials must be enabled, even when you aren't using authentication.
+* Credentials must be enabled, even when authentication is not used.
 
-For example, the following CORS policy allows a SignalR browser client hosted on `http://example.com` to access your SignalR app:
+For example, the following CORS policy allows a SignalR browser client hosted on `http://example.com` to access the SignalR app hosted on `http://signalr.example.com`:
 
 [!code-csharp[Main](security/sample/Startup.cs?name=snippet1)]
-
-```csharp
-public void Configure(IApplicationBuilder app)
-{
-    // ... other middleware ...
-
-    // Make sure the CORS middleware is ahead of SignalR.
-    app.UseCors(builder => {
-        builder.WithOrigins("http://example.com")
-            .AllowAnyHeader()
-            .WithMethods("GET", "POST")
-            .AllowCredentials();
-    });
-
-    // ... other middleware ...
-
-    app.UseSignalR();
-
-    // ... other middleware ...
-}
-```
 
 > [!NOTE]
 > SignalR is not compatible with the built-in CORS feature in Azure App Service.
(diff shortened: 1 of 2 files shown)


aspnetcore/signalr/security.md
```diff
@@ -5,121 +5,78 @@ description: Learn how to use authentication and authorization in ASP.NET Core S
...
-* `TransportMaxBufferSize` represents the maximum number of bytes the server can send. If the server attempts to send a message (includes return values from hub methods) larger than this limit, an exception will be thrown.
+* `TransportMaxBufferSize` represents the maximum number of bytes the server can send. If the server attempts to send a message (including return values from hub methods) larger than this limit, an exception will be thrown.
 
-Setting the limit to `0` disables the limit entirely. However, this should be done with extreme caution. Removing the limit allows a client to send a message of any size. This could be used by a malicious client to cause excess memory to be allocated, which could dramatically reduce the number of concurrent connections your app can support.
+Setting the limit to `0` disables the limit. However, disabling the limit exposes the app malicious large message creation. Removing the limit allows a client to send a message of any size. Malicious clients sending large
```
BrennanConroy: ~~However, disabling the limit exposes the app malicious large message creation.~~
Don't think the sentence is useful since we describe it immediately after

aspnetcore/signalr/security/sample/Startup.cs
```diff
@@ -0,0 +1,108 @@
...
+            app.Use((context, next) =>
+            {
+                // Check for a WebSocket request.
+                if (string.Equals(context.Request.Headers["Upgrade"], "websocket"))
+                {
+                    var origin = context.Request.Headers["Origin"];
+
+                    // If there is no origin header, or if the origin header doesn't match 
+                    // an allowed value:
+                    if (string.IsNullOrEmpty(origin) && !_allowedOrigins.Contains(origin))
```
BrennanConroy:
```suggestion
                    if (!string.IsNullOrEmpty(origin) && !_allowedOrigins.Contains(origin))
```
@anurse The sample was slightly wrong

aspnetcore/signalr/security.md
```diff
@@ -5,121 +5,78 @@ description: Learn how to use authentication and authorization in ASP.NET Core S
...
-* HTTP methods `GET` and `POST` must be allowed.
-* Credentials must be enabled, even when you aren't using authentication.
-
-For example, the following CORS policy allows a SignalR browser client hosted on `http://example.com` to access your SignalR app:
-
-```csharp
-public void Configure(IApplicationBuilder app)
-{
-    // ... other middleware ...
+For more information on configuring CORS, see [Enable Cross-Origin Requests (CORS)](xref:security/cors). SignalR requires the following CORS policies:
```
BrennanConroy: Maybe bold 'requires'

aspnetcore/signalr/security/sample/Startup.cs
```diff
@@ -0,0 +1,108 @@
...
+            // Validate Origin header on WebSocket requests to prevent unexpected cross-site 
+            // WebSocket requests.
+            app.Use((context, next) =>
+            {
+                // Check for a WebSocket request.
+                if (string.Equals(context.Request.Headers["Upgrade"], "websocket"))
+                {
+                    var origin = context.Request.Headers["Origin"];
+
+                    // If there is no origin header, or if the origin header doesn't match
```
BrennanConroy:
```suggestion
                    // If there is an origin header, and the origin header doesn't match 
```

aspnetcore/signalr/security/sample/Startup.cs
anurse: No, the original content was correct. We want to send a `400` when there is no `Origin` header as well.

anurse:
```suggestion
                    if (string.IsNullOrEmpty(origin) || !_allowedOrigins.Contains(origin))
```

BrennanConroy: If the origin header is null or empty then it could have been the C# or Java client connecting so we should allow them to connect

aspnetcore/signalr/security/sample/Startup.cs
anurse: The comment is correct here. We return `400` if there is no `Origin` header, or if the `Origin` header doesn't match the allowed values. The logic below was inaccurate but once the `if` is fixed, the original comment is correct.

aspnetcore/signalr/security/sample/Startup.cs
anurse: Fair point.

Rick-Anderson:
I took @BrennanConroy 	commit, not @anurse 	- hope that's correct.
I rewrote 
_In general, allow cross-origin requests only from domains you control._
as
_Allow cross-origin requests only from domains you trust or control._
can someone approve that?

Let me know if it's merge ready.

BrennanConroy:
If we take the code change (which I think we agreed on) then we should also take the comment change

aspnetcore/signalr/security.md
```diff
@@ -5,121 +5,77 @@ description: Learn how to use authentication and authorization in ASP.NET Core S
...
 
-When using WebSockets or Server-Sent Events, the browser client sends the access token in the query string. This is generally as secure as using the standard `Authorization` header, however many web servers log the URL for each request, including the query string. This means the access token may be included in logs. Consider reviewing the web server's logging settings to avoid logging this information.
+## Buffer management
 
-### Exceptions
+SignalR uses per-connection buffers to manage incoming and outgoing messages. By default, SignalR limits these buffers to 32 KB. The largest message a client or server can send is 32 KB. The maximum memory consumed by a connection for messages is 32 KB. If your messages are always smaller than 32 K, you can reduce the limit, which:
```
BrennanConroy: 32 K -> 32 KB

aspnetcore/signalr/security.md
```diff
@@ -5,121 +5,77 @@ description: Learn how to use authentication and authorization in ASP.NET Core S
...
 
-Exception messages are generally considered sensitive data that shouldn't be revealed to a client. By default, SignalR doesn't send the details of an exception thrown by a hub method to the client. Instead, the client receives a generic message indicating an error occurred. You can override this behavior by setting the [`EnableDetailedErrors`](xref:signalr/configuration#configure-server-options) setting.
+* Prevents a client from being able to send a larger message.
+* The server will never need to allocate large buffers to accept messages.
 
-### Buffer management
+If your messages are larger than 32 K, you can increase the limit. Increasing this limit means:
```
BrennanConroy: 32 K -> 32 KB

PR #9092, opened 2018-10-18
Update aspnetcore/security/authorization/razor-pages-authorization/sa…

CodefridgeVan:
`AuthenticateUser()` now uses the passed in email parameter, instead of the Input class property.

When checking the user entered email address, the `AuthenticateUser()` method was using the bound property `Input`, which is incorrect, the method must use the `email` parameter that is passed into the method.

PS: This is my first pull request, so is the above enough information?


scottaddie:
Thank you for your contribution, @CodefridgeVan! Your change will go live within the next 24 hours.

PR #9100, opened 2018-10-18
Update README.md

phaneendrababu:


scottaddie:
Thank you for the PR, but we prefer to explain this in the actual docs.

PR #9103, opened 2018-10-18
Description line in the wrong spot

guardrex:
Quick patch ... Idk how this got past three of us! 🙈


scottaddie:
@guardrex Thank you for catching this.

PR #9124, opened 2018-10-19
self-closing TH

### LINKED ISSUE #6075: Add docs for the semantics of self-closing for our built-in tag helpers.
It's a common mistake to try and use the built-in tag helpers like select as self-closing tags, which will not work. 

The semantics of self-closing tags are discussed briefly here: https://github.com/aspnet/Docs/blob/1832723bd1557feb789e7a96ef80d7c4a2a9436b/aspnetcore/mvc/views/tag-helpers/authoring.md#setattribute-and-setcontent

We should add this detail to docs like: https://docs.microsoft.com/en-us/aspnet/core/mvc/views/working-with-forms?view=aspnetcore-2.1#the-select-tag-helper

Users who haven't authored a tag helper before won't find the docs that explain this.

Example:
https://developercommunity.visualstudio.com/content/problem/238965/tag-helpers-dont-work-when-i-use-self-closing-tags.html

Rick-Anderson:
Fixes  #6075

### DIFF
--- a/aspnetcore/mvc/views/tag-helpers/intro.md
+++ b/aspnetcore/mvc/views/tag-helpers/intro.md
@@ -126,6 +126,10 @@ In the code image below, the Tag Helper prefix is set to `th:`, so only those el
 
 The same hierarchy rules that apply to `@addTagHelper` also apply to `@tagHelperPrefix`.
 
+## Self closing Tag Helpers
+
+In general, Tag Helpers can't be used as self closing tags. Self-closing a Tag Helper results in a self-closing tag in the rendered output. For more information see [this note](xref:mvc/views/tag-helpers/authoring#self-closing) in [Authoring Tag Helpers](xref:mvc/views/tag-helpers/authoring).
+
 ## IntelliSense support for Tag Helpers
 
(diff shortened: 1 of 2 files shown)


aspnetcore/mvc/views/tag-helpers/intro.md
```diff
@@ -126,6 +126,10 @@ In the code image below, the Tag Helper prefix is set to `th:`, so only those el
 
 The same hierarchy rules that apply to `@addTagHelper` also apply to `@tagHelperPrefix`.
 
+## Self-closing Tag Helpers
+
+In general, Tag Helpers can't be used as self-closing tags. Self-closing a Tag Helper results in a self-closing tag in the rendered output. For more information, see [this note](xref:mvc/views/tag-helpers/authoring#self-closing) in [Authoring Tag Helpers](xref:mvc/views/tag-helpers/authoring).
```
rynowak: I think it's more appropriate to say that you should use care when using a tag helper as a self-closing tag. Some tag helpers are designed to be self-closing tags, others are not, and it will suppress their output.

PR #9128, opened 2018-10-19
Added note for working with virtual directories

rschiefer:
I spent a day or more researching this so that I could use Swagger-UI in my ASP.NET Core API which is used with a reverse-proxy and virtual directory.  In the end found Swagger-UI requires relative URLs to start with a period.  Related to https://github.com/swagger-api/swagger-ui/issues/4107

Hoping this documentation note will help others.

  Fixes #Issue_Number


guardrex:
Hello @rschiefer ... I'm going to make a quick style pass and then call for review. Thanks for submitting this.

guardrex:
Let's discuss this for a sec ...

We shouldn't say "virtual directory" because that's overloaded for IIS's vdir scenario, which isn't supported. Also, can it be shortened? ... Can the TIP just say ...

> \> [!TIP]
> \> If the Swagger assets are in a sub-directory, set the Swagger endpoint to a relative path using the `./` prefix (for example, `./swagger/v1/swagger.json`).

Follow-up question: Will it also resolve the path correctly in this scenario if provided like this: `swagger/v1/swagger.json`?

rschiefer:
That makes sense.  I can simulate the virtual path issue by setting the following in the launchsettings file under iisSettings.

      "applicationUrl": "http://localhost:56943/SignatureRequest/v0.0.8/",

scottaddie:
@rschiefer Thank you for your patience and contribution. Your changes will go live within the next 24 hours.

rschiefer:
Awesome, thanks!

PR #9159, opened 2018-10-23
fix template option error

lovelangy:
There is no such  template option named 'webapp'. we should use 'mvc' or 'razor' instead.

### DIFF
--- a/aspnetcore/security/authentication/identity.md
+++ b/aspnetcore/security/authentication/identity.md
@@ -47,7 +47,7 @@ Create an ASP.NET Core Web Application project with Individual User Accounts.
 # [.NET Core CLI](#tab/netcore-cli)
 
 ```cli
-dotnet new webapp --auth Individual -o WebApp1
+dotnet new razor --auth Individual -o WebApp1
 ```
 
 ---


scottaddie:
@lovelangy The "razor" alias doesn't appear in the CLI output; however, it is valid. See https://github.com/aspnet/Templating/blob/1dfe06323f84e4c25356e6befe8b052cdf0c3707/src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/.template.config/template.json#L17. The first alias in the `shortName` array is the alias that displays in the CLI output.

lovelangy:
@scottaddie  oh, i see.  'webapp' is introduced in .net core 2.2.   It doesn't work on .net core 2.1.

lovelangy:
@scottaddie   I am afraid that is not the case.  

 This is what I have done test on my laptop:

1. Install .net core, 2.2.100-preview3-009430,   i can run "dotnet new webapp --auth Individual -o WebApp1" successfully.
2. Uninstall   .net core, 2.2.100-preview3-009430, verify the version via "dotnet --version".
3. it says "2.1.202", which means it is .net core 2.1
4. run "dotnet new webapp --auth Individual -o WebApp1" again.
5. it returns "No templates matched the input template name: webapp." 

Also, Based this doc,  webapp alias is not in .net core 2.1.
[dotnet new command](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-new?tabs=netcore21)

scottaddie:
@lovelangy Here's the ASP.NET Core 2.1 RTM templates release branch showing that both `webapp` and `razor` were supported in the Razor Pages project template:
https://github.com/aspnet/Templating/blob/release/2.1/src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/.template.config/template.json#L15

.NET Core SDK 2.1.300 was the RTM version of .NET Core 2.1. I see you're using .NET Core SDK 2.1.202, which would've installed the ASP.NET Core 2.0 runtime (2.0.9 to be exact). See https://www.microsoft.com/net/download/dotnet-core/2.0.

The doc you linked to needs to be updated. I'll take care of that right now. Thank you for pointing this out!

PR #9163, opened 2018-10-23
Update routing.md

amadosoft:
i think it should be "whether" not "where"


scottaddie:
Thank you for your contribution, @amadosoft!

PR #9175, opened 2018-10-23
fix error handling if BinderModelName is null

BenjaminAbt:
this sample code uses a fallback if no name of the model to bind is set

```cs
            var modelName = bindingContext.BinderModelName;
            if (string.IsNullOrEmpty(modelName))
            {
                modelName = "authorId";
            }
```

if no binder name is passed and the parse of int fails too, it results in an ArgumentException because `bindingContext.BinderModelName` is null

### DIFF
--- a/aspnetcore/mvc/advanced/custom-model-binding/sample/CustomModelBindingSample/Binders/AuthorEntityBinder.cs
+++ b/aspnetcore/mvc/advanced/custom-model-binding/sample/CustomModelBindingSample/Binders/AuthorEntityBinder.cs
@@ -56,7 +56,7 @@ namespace CustomModelBindingSample.Binders
             {
                 // Non-integer arguments result in model state errors
                 bindingContext.ModelState.TryAddModelError(
-                                        bindingContext.ModelName,
+                                        modelName,
                                         "Author Id must be an integer.");
                 return Task.CompletedTask;
             }


scottaddie:
@BenjaminAbt If I'm understanding correctly, wouldn't it be sufficient to simply do the following?

```csharp
var modelName = bindingContext?.BinderModelName;
```

BenjaminAbt:
This PR is made via edit on https://docs.microsoft.com/en-us/aspnet/core/mvc/advanced/custom-model-binding?view=aspnetcore-2.1#custom-model-binder-sample

The sample `Custom model binder sample` uses a fallback if no model binder name is given via Action attribute. The null-conditional operator would not help in case of the fallback

```cs
  var modelName = bindingContext.BinderModelName;
            if (string.IsNullOrEmpty(modelName))
            {
                modelName = "authorId";
            }
```

What happens:

Case 1:
`public IActionResult GetById([ModelBinder(Name = "author")]Author author)`

then `bindingContext.BinderModelName` is "author"

Case 2:
`public IActionResult GetById(Author author)`

then `bindingContext.BinderModelName` is `null`

Additional:

if the given value is no valid `int` then
```cs
 bindingContext.ModelState.TryAddModelError(
                                    bindingContext.ModelName,
                                    "Author Id must be an integer.");
```
would be reached and fails in case 2, because `bindingContext.ModelName` is null.

scottaddie:
@BenjaminAbt Thank you for the detailed explanation.

Thoughts on this one, @dougbu?

dougbu:
The in-box model binders don't directly use `BinderModelName` much at all. (`BodyModelBinder` and `FormFileModelBinder` are the exceptions and they don't use `IValueProvider`s.) The reason is `BinderModelName` does not contain the full path to a C# property when binding nested models. That is, this sample will produce unexpected results if the `Author` class is used as a property in another model.

Recommend changing the model binder to use `bindingContext.ModelName` everywhere. That property is never `null` and defaults to the name of the parameter or property when binding top-level models. Deeper down, it will always contain the path to a C# property. And, in either case, `ModelName` reflects any `ModelBinderAttribute.Name` settings.

BenjaminAbt:
Thanks for your answer, but I think this wont help here too.

This is the Action I test:
```cs
 [HttpGet("{id}")]
public async Task<ActionResult<AuthorApiViewModel>> Get(Author author)
```

image

`bindingContext.ModelName` is not null, but empty.
So it wont help in case of this fall back sample.

It would fix the issue that we dont try to add `null` as key; but it would not cover the fallback of `id`

dougbu:
@BenjaminAbt are you talking about the following action?
``` c#
[HttpGet("get/{authorId}")]
public IActionResult Get(Author author)
{
    // ...
}
```
I ask because `binderContext.ModelName` is only empty when the `IValueProvider`s have nothing that matches the top-level parameter or property name. In the above case, there's `"authorId"` data but nothing for `"author"`.

The right way to fix this is to use `[ModelBinder(Name = "authorId")]` or to rename the parameter to `authorId`. Either informs MVC of the correct `ModelName`. Trying to workaround this later, in the model binder, is a hack and not what we'd recommend.

BenjaminAbt:
This means in this sample:

```cs
[HttpGet("{author}", Name = "GetAuthorById")]
public async Task<ActionResult<AuthorApiViewModel>> Get(Author author)
```
`binderContext.ModelName` should be `author` ?

In my case it's still empty.
image

I tried several ways; only with the attribute `ModelName` is not empty.

```cs
[HttpGet("{authorId}", Name = "GetAuthorById")]
public async Task<ActionResult<AuthorApiViewModel>> Get([ModelBinder(Name = "authorId")]Author author)
```

image

BenjaminAbt:
The earlier screenshots were from a private code I cannot share.

But please take a look to this one: https://github.com/BenjaminAbt/AspNetCoreModelBinderDemo
Same behavior:

image

The naming of parameter in route and action also affects a reported bug issue of me: https://github.com/aspnet/Mvc/issues/8636

dougbu:
Your repro project is completely different from the sample we're discussing in this issue. `[ApiController]` changes many things, including overriding `BinderModelName` when the parameter type is complex and the binding source is inferred to be `BindingSource.Path`.

Get rid of `[ApiController]` (and add checks of `ModelState.IsValid` because `ModelStateInvalidFilter` won't be added automatically) or be explicit about the name using `[ModelBinder(...)]`.

BenjaminAbt:
So maybe a hint in the documentation about issues with `ApiController` would be nice....

dougbu:
I wouldn't say "issues with `ApiController`" at all. That attribute brings in a set of relatively simple conventions and, if you need to do something more complex, either don't use those conventions or be explicit.

dougbu:
> It makes no sense if a new introduced attribute removes a default behavior and there is no documentation hint for that.

I haven't read all of the words in this repo about `[ApiController]`. If there isn't a list of the conventions that attribute introduces, I agree there should be.

/cc @danroth27 @rynowak

dougbu:
@scottaddie please close this PR

But, using whatever the current process is, please open an issue about moving from `BinderModelName` (which can be `null` and which applies only to top-level model binding) to `ModelName`.

dougbu:
@scottaddie have you filed an issue about changing the example to use `ModelName` instead of `BinderModelName`?

scottaddie:
Thank you for the reminder, @dougbu! This work is being tracked here: https://github.com/aspnet/Docs/issues/9476.

PR #9177, opened 2018-10-23
Update TOC to have the .NET Client API reference.

mikaelm12:
Adding a line to the table of contents. We need a reference to the .NET Client API reference link here.

### DIFF
--- a/aspnetcore/toc.md
+++ b/aspnetcore/toc.md
@@ -219,7 +219,7 @@
 ### [Publish to Azure](xref:signalr/publish-to-azure-web-app)
 ## [Clients](xref:signalr/clients)
 ### [.NET client](xref:signalr/dotnet-client)
-### [.NET API reference](/dotnet/api/microsoft.aspnetcore.signalr.client?view=aspnetcore-2.1)
+### [.NET API reference](/dotnet/api/microsoft.aspnetcore.signalr.client)
 ### [Java client](xref:signalr/java-client)
 ### [Java API reference](/java/api/com.microsoft.aspnet.signalr?view=aspnet-signalr-java)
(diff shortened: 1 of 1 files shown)


aspnetcore/toc.md
```diff
@@ -219,6 +219,7 @@
 ### [Publish to Azure](xref:signalr/publish-to-azure-web-app)
 ## [Clients](xref:signalr/clients)
 ### [.NET client](xref:signalr/dotnet-client)
+### [.NET API reference](/dotnet/api/microsoft.aspnetcore.signalr.client?view=aspnetcore-2.1)
```
tdykstra: Is it necessary to hard-code 2.1 here?

Would an xref link work -- `(xref:Microsoft.AspNetCore.Signalr.Client)`

mikaelm12: Hard coding to 2.1 isn't necessary.

 >Would an xref link work

I was just following the Java and JavaScript API reference example. They don't use xref so I assumed it shouldn't be used here as well.

tdykstra: I just checked with https://xref.docs.microsoft.com/query?uid=Microsoft.AspNetCore.SignalR.Client and the xref link should work.

The Java link should also work as an xref, as https://xref.docs.microsoft.com/query?uid=com.microsoft.aspnet.signalr checks out too.

JavaScript API ref works differently.

Updating them now isn't necessary, it's a direction we want to go in the future.  Info about the xref links is in an [internal guide doc](https://review.docs.microsoft.com/en-us/help/contribute/contribute-how-to-write-links?branch=master).

mikaelm12:
cc @anurse and @BrennanConroy just fyi

PR #4532, opened 2017-10-10
Allow lib folder contents for samples

guardrex:
Due to the changes coming along for https://github.com/aspnet/templating/issues/48, samples will get *lib* folder static assets that Bower would have normally provided when the app is built/published.

Side Note: We'll also include bundled+minified styles and scripts that `BundlerMinifier.Core` would have produced for the app.

Therefore, new samples and existing sample updates should contain all of the static assets they need to run (*lib* folder and *.min.* CSS and JS files) and Bower+`BunderMinifier.Core` shouldn't be part of the app (no configuration files for these features and no targets and package+tooling references in the project file for these features).

cc/ @Rick-Anderson

PR #9188, opened 2018-10-24
Add wwwroot to .gitignore

### LINKED ISSUE #9184: Update .gitignore

Rick-Anderson:
Fixes #9184

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 # The 5 approved on boarding samples that need wwwroot/lib/ can add a local .gitignore that allows wwwroot/lib/.
-wwwroot/lib/
+# with 2.1 templates and excluding .min & .map -> 27 files and < 1 KB
+# 3.0 templates should be even less
+# wwwroot/lib/
 # When wwwroot/lib/ is commented out, excluse .min and .map files
 *.min.css
 *.min.js
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,10 @@
 # The 5 approved on boarding samples that need wwwroot can add a local .gitignore that allows wwwroot.
 wwwroot/
+# When wwwroot is commented out, excluse .min and .map files
+*.min.css
+*.min.js
+*.map
+
 _build/
 _site/
 Properties/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
+# The 5 approved on boarding samples that need wwwroot can add a local .gitignore that allows wwwroot.
```
tdykstra: The change in https://github.com/aspnet/Docs/pull/4532 removed ignore lines for wwwroot/lib, not all of wwwroot.  Are you sure it won't create problems to ignore all of wwwroot?  Don't some projects put project-specific files in there on occasion?

Rick-Anderson:
[This version of .gitignore](https://github.com/aspnet/Docs/commit/29d43601f87a883f0beade5b7231097ed9c4a15f) had wwwroot

Rick-Anderson:
@danroth27 	can we merge this? We've had wwwroot listed in *.gitignore* in the past. If it's turns out to be a problem we'll update it.

danroth27:
@Rick-Anderson No, I wouldn't. This *.gitignore* file change seems way too aggressive. I'm sure we will have some sample apps that need content *wwwroot* or will want to include JavaScript or CSS files. I would instead focus our efforts on simplifying the existing samples

Rick-Anderson:
Last pitch. It seems aggressive but it's actually not. We've had *wwwroot* in for extended periods with no problems. We've almost always had `wwwroot/lib/` in *.gitignore*.

For the 95% of samples that don't need *wwwroot*, it's alot of work to remove  *wwwroot* and it's typically not done. It's impossible to police keeping out unneeded *wwwroot* assets without *.gitignore*.

For the 5% that need *wwwroot* , it's trivial to add.

Should we go back to what we had for most of our history? see #4532

```
**/sample/**/wwwroot/lib/
**/samples/**/wwwroot/lib/
```

I'm OK with `wwwroot/lib/` - that's 97% of `wwwroot`.

Keep in mind samples that currently contain *wwwroot* assets don't loose them when *wwwroot* is added back and the project is updated.

danroth27:
> Should we go back to what we had for most of our history? see #4532

I think that's certainly better than excluding all static files support from every sample.

PR #9213, opened 2018-10-25
Update web app template

akhilreddy18:
There is no template such as: webapp. The available templates for web app are web, mvc and razor.

  Fixes #Issue_Number

### DIFF
--- a/aspnetcore/razor-pages/index.md
+++ b/aspnetcore/razor-pages/index.md
@@ -49,7 +49,7 @@ Open the generated *.csproj* file from Visual Studio for Mac.
 
 ::: moniker range=">= aspnetcore-2.1"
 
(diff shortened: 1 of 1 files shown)


scottaddie:
@akhilreddy18 Thank you for the PR; however, `webapp` is a valid alias. This is documented here: https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-new?tabs=netcore21#arguments. The alias was introduced in ASP.NET Core 2.1. The `dotnet new` CLI experience only displays the first alias in the `shortName` alias. See https://github.com/aspnet/Templating/blob/release/2.1/src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/.template.config/template.json#L15.

PR #9215, opened 2018-10-25
Changed to reflect default json file

daveabrock:
While going through [this topic](https://docs.microsoft.com/en-us/aspnet/core/tutorials/razor-pages-vsc/model?view=aspnetcore-2.1), I noticed that the *appsettings.json* file that is shipped with the template is different now. I updated it. With the change, the highlighted lines do not change - a nice surprise.


scottaddie:
Thank you for fixing this, Dave!

PR #9220, opened 2018-10-25
Update TOC to reflect new SignalR Java client package name

mikaelm12:
The current link 404s because we've updated the package name from `com.microsoft.aspnet.signalr` to `com.microsoft.signalr`


mikaelm12:
@scottaddie Can I make that a separate change? I did this change in the browser which only allows editing a single file? Thanks

scottaddie:
@mikaelm12 That works for me! Thanks for your help.

mikaelm12:
Yeah! Thanks for your help in getting this all set up!

PR #9223, opened 2018-10-25
Implicit versioning of the All package

### LINKED ISSUE #9219: Clarify that with 2.1 you should not specify an explicit version when using the .All package
We currently document that starting with 2.1 that you should not specify an explicit package version when using .App and we recommend using .App instead of .All, but for customers that are still using .All we should also recommend that you should not specify an explicit version just like with .App.

guardrex:
Fixes #9219 

@JunTaoLuo Dan calls out the implicit versioning "for 2.1 SDK or later;" however, we don't usually version the doc content by SDK release. We might need to say something for this update in the passage that I'm adding to the All topic (*aspnetcore/fundamentals/metapackage.md*) about which SDK version this applies to. I don't know if that's something that you would like to do or what version to call out.

I also remove the version from the samples. We also assume that devs are using the latest SDK with our samples.

### DIFF
--- a/aspnetcore/client-side/bundling-and-minification/samples/BuildBundlerMinifierApp/BuildBundlerMinifierApp.csproj
+++ b/aspnetcore/client-side/bundling-and-minification/samples/BuildBundlerMinifierApp/BuildBundlerMinifierApp.csproj
@@ -4,7 +4,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="BuildBundlerMinifier" Version="2.6.362" />
-    <PackageReference Include="Microsoft.AspNetCore.All" />
+    <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.9" />
   </ItemGroup>
   <ItemGroup>
     <DotNetCliToolReference Include="BundlerMinifier.Core" Version="2.6.362" />
--- a/aspnetcore/client-side/bundling-and-minification/samples/BuildBundlerMinifierApp/BuildBundlerMinifierApp.csproj
+++ b/aspnetcore/client-side/bundling-and-minification/samples/BuildBundlerMinifierApp/BuildBundlerMinifierApp.csproj
@@ -4,7 +4,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="BuildBundlerMinifier" Version="2.6.362" />
-    <PackageReference Include="Microsoft.AspNetCore.All" Version="2.1.4" />
+    <PackageReference Include="Microsoft.AspNetCore.All" />
   </ItemGroup>
   <ItemGroup>
     <DotNetCliToolReference Include="BundlerMinifier.Core" Version="2.6.362" />
--- a/aspnetcore/fundamentals/metapackage.md
+++ b/aspnetcore/fundamentals/metapackage.md
@@ -37,7 +37,7 @@ The following *.csproj* file references the `Microsoft.AspNetCore.All` metapacka
 
 ## Implicit versioning
 
-In ASP.NET Core 2.x, you can specify the `Microsoft.AspNetCore.All` package reference without a version. When the version isn't specified, an implicit version is specified by the SDK (`Microsoft.NET.Sdk.Web`). We recommend relying on the implicit version specified by the SDK and not explicitly setting the version number on the package reference. If you have questions about this approach, leave a GitHub comment at the [Discussion for the Microsoft.AspNetCore.App implicit version](https://github.com/aspnet/Docs/issues/6430).
+In ASP.NET Core 2.1 or later, you can specify the `Microsoft.AspNetCore.All` package reference without a version. When the version isn't specified, an implicit version is specified by the SDK (`Microsoft.NET.Sdk.Web`). We recommend relying on the implicit version specified by the SDK and not explicitly setting the version number on the package reference. If you have questions about this approach, leave a GitHub comment at the [Discussion for the Microsoft.AspNetCore.App implicit version](https://github.com/aspnet/Docs/issues/6430).
 
 The implicit version is set to `major.minor.0` for portable apps. The shared framework roll-forward mechanism runs the app on the latest compatible version among the installed shared frameworks. To guarantee the same version is used in development, test, and production, ensure the same version of the shared framework is installed in all environments. For self-contained apps, the implicit version number is set to the `major.minor.patch` of the shared framework bundled in the installed SDK.
 
--- a/aspnetcore/fundamentals/metapackage.md
+++ b/aspnetcore/fundamentals/metapackage.md
@@ -31,13 +31,17 @@ You can use the package trimming process to remove packages that you don't use. 
 
 The following *.csproj* file references the `Microsoft.AspNetCore.All` metapackage for ASP.NET Core:
 
-[!code-xml[](metapackage/samples/Metapackage.All.Example.csproj?highlight=6)]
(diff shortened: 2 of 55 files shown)


guardrex:
Also should note that I didn't check that warning message to confirm that it says "All" instead of "App" ... you know ... if the dev tries to use the `Microsoft.NET.Sdk` SDK instead of `Microsoft.NET.Sdk.Web`.

aspnetcore/client-side/bundling-and-minification/samples/BuildBundlerMinifierApp/BuildBundlerMinifierApp.csproj
```diff
@@ -4,7 +4,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="BuildBundlerMinifier" Version="2.6.362" />
-    <PackageReference Include="Microsoft.AspNetCore.All" Version="2.1.4" />
+    <PackageReference Include="Microsoft.AspNetCore.All" />
```
JunTaoLuo: In most of these cases, we should convert to using the Microsoft.AspnetCore.App metapackage instead of .All. This is the default in our 2.1 templates.

guardrex: Yes, these are all un-converted 2.0 sample apps. They're tracked on #5495.

aspnetcore/fundamentals/metapackage.md
```diff
@@ -34,6 +33,18 @@ The following *.csproj* file references the `Microsoft.AspNetCore.All` metapacka
 
 [!code-xml[](metapackage/samples/Metapackage.All.Example.csproj?highlight=6)]
 
+The preceding markup represents a typical ASP.NET Core 2.0 template. It doesn't specify a version number for the `Microsoft.AspNetCore.All` package reference. When the version isn't specified, an [implicit](https://github.com/dotnet/core/blob/master/release-notes/1.0/sdk/1.0-rc3-implicit-package-refs.md) version is specified by the SDK (`Microsoft.NET.Sdk.Web`). We recommend relying on the implicit version specified by the SDK and not explicitly setting the version number on the package reference. If you have questions about this approach, leave a GitHub comment at the [Discussion for the Microsoft.AspNetCore.App implicit version](https://github.com/aspnet/Docs/issues/6430).
```
JunTaoLuo: I think we need 2.0 => 2.1. In 2.0, we expect explicit versions. Also I wouldn't link the dotnet release notes since the implicit reference to Microsoft.NETCore.App works quite a bit differently than the semi-implicit reference to Microsoft.AspNetCore.App.

aspnetcore/fundamentals/metapackage.md
```diff
@@ -34,6 +33,18 @@ The following *.csproj* file references the `Microsoft.AspNetCore.All` metapacka
...
+
+The implicit version is set to `major.minor.0` for portable apps. The shared framework roll-forward mechanism runs the app on the latest compatible version among the installed shared frameworks. To guarantee the same version is used in development, test, and production, ensure the same version of the shared framework is installed in all environments. For self-contained apps, the implicit version number is set to the `major.minor.patch` of the shared framework bundled in the installed SDK.
+
+Specifying a version number on the `Microsoft.AspNetCore.All` package reference does **not** guarantee that version of the shared framework is chosen. For example, suppose version "2.0.0" is specified, but "2.0.1" is installed. In that case, the app will use "2.0.1". Although not recommended, you can disable roll forward (patch and/or minor). For more information regarding dotnet host roll-forward and how to configure its behavior, see [dotnet host roll forward](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/roll-forward-on-no-candidate-fx.md).
```
JunTaoLuo: Let's use 2.1.1 and 2.1.3 as examples instead since 2.1 is the first time we have this behaviour. This would also bring the docs to parity with what's in the metapackage-app.md file.

aspnetcore/fundamentals/metapackage.md
JunTaoLuo: What I mean is that if we are adding a section regarding implicit versions, then it should be clear that we are talking about 2.1. There is no implicit versioning in 2.0. In fact I realized the sample apps are all mismatched right now with a netcoreapp2.0 tfm but using 2.1.* referenced.

guardrex: Yes, I see what you mean ... `2.0.9` would have been a better choice. :cry:

JunTaoLuo: Let me clarify a little here. In 2.0, all versions were explicit and the recommended way of targeting aspnetcore was through an explicit package reference to Microsoft.AspNetCore.All. In 2.1, the recommended way of targeting aspnetcore is through a version-less reference to Microsoft.AspNetCore.App. However, we still support references to Microsoft.AspNetCore.All in 2.1, though if you continue to use it, the recommendation would be to remove the version number and rely on the implicit version mechanisms.

guardrex: Looks like I should do the following:

* Restore the `Version` in the 2.0 `All`-based sample app project files but use `2.0.9`.
* Leave the current explanation alone and leave the example `netcoreapp2.0` with `All` and a version.
* Add a separate section here for the info on implicit versioning. Make it clear that it's for apps that target 2.1. Perhaps, include our usual recommendation (found elsewhere here, too) to use the `App` package. This section won't show actual code (i.e., `netcoreapp2.1` with `All`).

JunTaoLuo: Sounds good

guardrex:
@JunTaoLuo Ok ... try it now. I didn't add the warning, as we have it clearly spelled out at the top and bottom of the topic.

aspnetcore/fundamentals/metapackage.md
```diff
@@ -32,13 +31,31 @@ You can use the package trimming process to remove packages that you don't use.
...
 The following *.csproj* file references the `Microsoft.AspNetCore.All` metapackage for ASP.NET Core:
 
-[!code-xml[](metapackage/samples/Metapackage.All.Example.csproj?highlight=6)]
+[!code-xml[](metapackage/samples/Metapackage.All.Example.csproj?highlight=8)]
+
+::: moniker range=">= aspnetcore-2.1"
+
+## Implicit versioning
+
+In ASP.NET Core 2.x, you can specify the `Microsoft.AspNetCore.All` package reference without a version. When the version isn't specified, an implicit version is specified by the SDK (`Microsoft.NET.Sdk.Web`). We recommend relying on the implicit version specified by the SDK and not explicitly setting the version number on the package reference. If you have questions about this approach, leave a GitHub comment at the [Discussion for the Microsoft.AspNetCore.App implicit version](https://github.com/aspnet/Docs/issues/6430).
```
JunTaoLuo: 2.x => 2.1 and higher. I know it's guarded by the moniker range but seeing 2.x is still rather confusing.

guardrex: I was slightly concerned about saying "or later" in the context of 3.0 and our not changing this (by accident) when 3.0 arrives and perhaps `All` becomes unsupported. I think I can't put 3.0 on the moniker range at this time because the version probably hasn't been added yet to the build system ... it will probably :boom: if I do that. I'll say "or later" here, but brace yourself ... we don't have a stellar track record when we don't harden something on the spot for future changes. 🙈

PR #9241, opened 2018-10-26
Update working-with-forms.md

### LINKED ISSUE #3863: Views/Working With Forms has a confusing (potentially wrong) statement about using for vs foreach
A statement on the [Working With Forms](https://docs.microsoft.com/en-us/aspnet/core/mvc/views/working-with-forms) has been brought up to me, and it got me a bit confused. 

It states (it's the only reference to `foreach` on the whole page):
> Always use `for` (and not `foreach`) to iterate over a list. Evaluating an indexer in a LINQ expression can be expensive and should be minimized.

Now this is interesting for two reasons.

First of all - wouldn't it be the other way around? If you want to iterate through the whole list, wouldn't foreach use an IEnumerator, which should be more efficient, when going through all the elements in a collection in order?

Secondly.. If evaluating indexers are expensive, shouldn't we **avoid** using a for loop?

It seems like this is either badly worded, and just adds confusion.. Or it's just plain wrong, in which case it needs amending :)

Thoughts?

Rick-Anderson:
Fixes #3863

### DIFF
--- a/aspnetcore/mvc/views/working-with-forms.md
+++ b/aspnetcore/mvc/views/working-with-forms.md
@@ -239,8 +239,7 @@ The *Views/Shared/EditorTemplates/ToDoItem.cshtml* template:
 
 [!code-HTML[](working-with-forms/sample/final/Views/Shared/EditorTemplates/ToDoItem.cshtml)]
 
->[!NOTE]
->Always use `for` (and *not* `foreach`) to iterate over a list. Evaluating an indexer in a LINQ expression can be expensive and should be minimized.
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@NTaylorMullen 	can you [edit in browser](https://github.com/aspnet/Docs/edit/Rick-Anderson-patch-7/aspnetcore/mvc/views/working-with-forms.md?pr=%2Faspnet%2FDocs%2Fpull%2F9241) and fix the text how you like it?

Or paste it here.

PR #9245, opened 2018-10-26
Update Moves Styling of Tutorial Site to a site.css file.

zez9787:
Minor change to move styling from inside of index.cshtml into its own site.css file.


guardrex:
Hello @zez9787 ... thanks for contributing. We can't take this because the tutorial has the reader add the content of this file to the app as a cut-and-paste from the topic. If the styles are split out into another file, then the topic will require another step (and the new *site.css* file must be shown). It's easier for a user to just cut-and-paste the one file instead of two.

PR #9248, opened 2018-10-26
Update docker-aspnetmvc.md

Lampei:
The current windows container releases appear to have fixed the "localhost" issue that was previously mentioned in this readme.  I've updated the article and have made a separate pull request for the referenced ps1 script to also remove the `docker inspect`


spboyer:
@Lampei Thanks for the update, and congrats on your **first commit**

PR #9259, opened 2018-10-26
convert toc for aspnet folder to yaml

tdykstra:
[Internal review link](https://review.docs.microsoft.com/en-us/aspnet/overview?branch=pr-en-us-9259
)


Rick-Anderson:
@tdykstra 	missing config builders
image

tdykstra:
Added it, though it looks out of place all by itself like that.  Unfortunately the ToC doesn't have an appropriate place for it.  I was tempted to add a Fundamentals group node, but we're not likely to have anything else to put in it, so that would look strange too with just one entry.

PR #9261, opened 2018-10-27
Use comment in pull request template

SeanKilleen:
Since GitHub supports comments in pull request templates, I thought this might be a nice way to provide the text without causing anyone to have to delete it afterward. We used it on another project of mine with success, so contributing the idea back.


Rick-Anderson:
@SeanKilleen that's a really nice improvement.

cc 	@rpetrusha - you'll want this for sure.
cc @tdykstra 	@guardrex 	@scottaddie

SeanKilleen:
@guardrex great idea, and I think you're right, it would. Will make the change now.

SeanKilleen:
@guardrex since this was merged, see #9265. Thanks!

PR #9262, opened 2018-10-27
Update validation.md

yimogit:
Added description of default model validation
by https://github.com/aspnet/AspNetCore/issues/3485

@Rick-Anderson 	edit: 
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-1.0&branch=pr-en-us-9262#handling-model-state-errors)

### DIFF
--- a/aspnetcore/mvc/models/validation.md
+++ b/aspnetcore/mvc/models/validation.md
@@ -84,6 +84,16 @@ Model validation occurs prior to each controller action being invoked, and it's 
 
 Some apps will choose to follow a standard convention for dealing with model validation errors, in which case a filter may be an appropriate place to implement such a policy. You should test how your actions behave with valid and invalid model states.
 
+Because the internal model validator is implemented by default, so the previous custom validator will not work. 
+This line can be added to the ConfigureServices method of the Startup class to disable the default behavior:   
+
+``` csharp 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@dougbu please review aspnet/AspNetCore#3485. The snippet is not in the right place and the wording isn't correct. Please suggest proper wording.

dougbu:
@Rick-Anderson I'm not sure of the proper way to word this. But, I agree the current wording is misleading.

Custom validations are still supported -- no change in 2.1. Further, the `SuppressModelStateInvalidFilter` setting has no effect on validations. Instead, it ensures requests which contain invalid data (i.e. `ModelState.IsValid` is false) automatically result in a 400 response and never reach your action.

The other important point here is that the filter doing this work isn't applied unless `[ApiController]` is associated with the controller containing the action.

Rick-Anderson:
@dougbu is there anything that needs to be documented on this?

dougbu:
> is there anything that needs to be documented on this?

Yes, should document the *actual* case where `SuppressModelStateInvalidFilter` makes sense and the tie between that property and `[ApiController]`. Suppressing this filter is useful if your action can recover from a model state error or wishes to customize its response. (Again, custom validators are identical to other validators for these scenarios.)

PR #9263, opened 2018-10-27
Additional cookie/bearer token auth info

### LINKED ISSUE #9147: Bearer Token Authentication
Please could you expand on the recommendation that bearer token authentication is used rather than cookie based authentication in a non-browser scenario. We currently have a C# WPF desktop application using cookie authentication and are trying to establish the benefits and drawbacks of migrating to bearer token authentication on our usage of SignalR.

---
#### Document Details

* ID: d0da2144-3730-f8ef-1d16-a89f11fd6198
* Version Independent ID: 6d8f5bc8-8f5c-ba13-b7f2-0366eac46f81
* Content: [Authentication and authorization in ASP.NET Core SignalR](https://docs.microsoft.com/en-us/aspnet/core/signalr/authn-and-authz?view=aspnetcore-2.1)
* Content Source: [aspnetcore/signalr/authn-and-authz.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/signalr/authn-and-authz.md)
* Product: **aspnet-core**
* GitHub Login: @tdykstra
* Microsoft Alias: **anurse**

tdykstra:
Fixes #9147

### DIFF
--- a/aspnetcore/signalr/authn-and-authz.md
+++ b/aspnetcore/signalr/authn-and-authz.md
@@ -23,11 +23,13 @@ SignalR can be used with [ASP.NET Core Authentication](xref:security/authenticat
 
 In a browser-based app, cookie authentication allows your existing user credentials to automatically flow to SignalR connections. When using the browser client, no additional configuration is needed. If the user is logged in to your app, the SignalR connection automatically inherits this authentication.
 
-Cookie authentication isn't recommended unless the app only needs to authenticate users from the browser client. When using the [.NET Client](xref:signalr/dotnet-client), the `Cookies` property can be configured in the `.WithUrl` call in order to provide a cookie. However, using cookie authentication from the .NET Client requires the app to provide an API to exchange authentication data for a cookie.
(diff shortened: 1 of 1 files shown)


aspnetcore/signalr/authn-and-authz.md
```diff
@@ -23,11 +23,13 @@ SignalR can be used with [ASP.NET Core Authentication](xref:security/authenticat
...
 
 ### Bearer token authentication
 
-Bearer token authentication is the recommended approach when using clients other than the browser client. In this approach, the client provides an access token that the server validates and uses to identify the user. The details of bearer token authentication are beyond the scope of this document. On the server, bearer token authentication is configured using the [JWT Bearer middleware](/dotnet/api/microsoft.extensions.dependencyinjection.jwtbearerextensions.addjwtbearer).
+The client can provide an access token without using a cookie. The server validates the token and uses it to identify the user. This validation is done only when the connection is established. During the life of the connection, the server doesn't automatically revalidate to check for token revocation.
```
anurse: Possible reword?

```suggestion
The client can provide an access token instead of using a cookie. The server validates the token and uses it to identify the user. This validation is done only when the connection is established. During the life of the connection, the server doesn't automatically revalidate to check for token revocation.
```

Also, technically the connection is re-authenticated regularly when using Long Polling. But I'm not sure it's worth calling that out.

PR #9281, opened 2018-10-29
app-stat.md -  added needed commas, corrected grammar.

binarygalwalkin:
No issue # to reference.

### DIFF
--- a/aspnetcore/fundamentals/app-state.md
+++ b/aspnetcore/fundamentals/app-state.md
@@ -34,15 +34,15 @@ State can be stored using several approaches. Each approach is described later i
 
 Cookies store data across requests. Because cookies are sent with every request, their size should be kept to a minimum. Ideally, only an identifier should be stored in a cookie with the data stored by the app. Most browsers restrict cookie size to 4096 bytes. Only a limited number of cookies are available for each domain.
 
-Because cookies are subject to tampering, they must be validated by the app. Cookies can be deleted by users and expire on clients. However, cookies are generally the most durable form of data persistence on the client.
+Because cookies are subject to tampering, they must be validated by the app. Cookies can be deleted by users, and expire on clients. However, cookies are generally the most durable form of data persistence on the client.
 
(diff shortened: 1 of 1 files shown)


binarygalwalkin:
@guardrex yes, those (2) lines struck me as funny.  I wanted to say something along those lines "this setting" - did not want to be too presumptuous. This is my first contribution to the project.

If you feel it noteworthy to go over, I am game.

guardrex:
Sure thing ...

37 - Comma splice (compound verb)
39 - Comma splice or unnecessary: A splice **_if_** we take "is" as auxiliary and "identified" and "authenticated" as intransitive verbs. If "is" is a state verb, then perhaps one could argue this the other way. *Where's an English professor when you need one??!!* :smile: The other thing about this sentence is that it's short and there's no confusion when leaving out commas. There's another way to go, too, where the "[and] not authenticated" phrase is set off with commas on both sides making it nonessential. Going back to the *it's short-no confusion* concept anyway, I just felt the comma isn't needed.
45 - Subordinate conjunction phrases don't get commas. They're considered essential, and that's certainly true on Line 45. We sometimes fight about this with a compound introductory phrase (e.g., a prep phrase + a subordinate conjunction phrase ... 👊🚑😄 lol) ... but we'll probably simplify those cases into one phrase to avoid the debate.
67 - Unnecessary definite article with the adjective
84 - We use "session" (or "session state") to identify this subsystem. It's not necessary to say "each" session in this context. It might be a little nicer as "session state," and that would match the section heading.

Last two are a little strange. We're generally trying to remove "note" in these cases. For the change you put in, it shifts the subject from an assumed "you" to "this," and I'm trying to avoid pronouns as much as possible. I'm not sure how the whole team feels about it. I think the regular use of pronouns a slippery slope to unfocused technical writing ... they can be downright confusing if not carefully used. I think it's best to just slap a good noun in there in most cases.

Congrats on your first contribution! :tada:

We **_DO_** debate grammar and style from time-to-time, and I think it's healthy. Usually, we reach agreement. The [MS Style Manual](https://docs.microsoft.com/style-guide/welcome/) is fairly hefty, and the Chicago Manual of Style is over 1,000 pages! *Yikes!* lol

PR #9285, opened 2018-10-29
C47282: Broken links

v-makoud:
Hello, @Rick-Anderson,
Description of the source issue: There are two links in source code next to each other, causing they appear broken on LOC site.
Proposed fix: Only one link should be in the () brackets, but neither of them is working. This will at least fix the issue that is causing a code to be shown in LOC sites
Many thanks in advance.


guardrex:
@Rick-Anderson Does this one look correct for this link? ...

https://code.msdn.microsoft.com/MVC5-Demo-with-Entity-c6bc81df

PR #9286, opened 2018-10-29
C47285: Adding "\" before "@"

v-makoud:
Hello, @rick-anderson,
Description of the source issue: Missing "\" is causing creation of link on ENU site.
Many thanks in advance.


guardrex:
Moved to https://github.com/aspnet/Docs/pull/9291

PR #9295, opened 2018-10-29
Update Razor Pages

agriffard:
Add 'as' in 'Mark page properties **as** required' to be clearer and avoid mistakes when translated.


scottaddie:
Thank you for clarifying this header, @agriffard!

PR #9301, opened 2018-10-29
Enforce HTTPS updates

### LINKED ISSUE #6538: Expand scope of the Enforce HTTPS article

guardrex:
Fixes #6538 

@shirhatti I think&dagger; this will wrap the updates.

&dagger;The changes span several recent PRs. I think your list is covered.

I suggest that after we get this one in that some of us take a look and see if we can spot anything MIA. Let's open one or more new issues to address anything we find lacking.

### DIFF
--- a/aspnetcore/host-and-deploy/azure-apps/index.md
+++ b/aspnetcore/host-and-deploy/azure-apps/index.md
@@ -202,6 +202,13 @@ When deploying a self-contained app:
 
 The [Docker Hub](https://hub.docker.com/r/microsoft/aspnetcore/) contains the latest preview Docker images. The images can be used as a base image. Use the image and deploy to Web Apps for Containers normally.
 
+## Protocol settings (HTTPS)
+
+Secure protocol bindings allow you specify which certificate to use when responding to requests to a specific hostname over HTTPS. Binding requires a valid private certificate (*.pfx*) issued for the specific hostname. For more information, see the following topics:
+
+* [Tutorial: Bind an existing custom SSL certificate to Azure Web Apps](/azure/app-service/app-service-web-tutorial-custom-ssl)
+* [How To Configure TLS Mutual Authentication for Web App](/azure/app-service/app-service-web-configure-tls-mutual-auth)
+
 ## Additional resources
 
 * [Web Apps overview (5-minute overview video)](/azure/app-service/app-service-web-overview)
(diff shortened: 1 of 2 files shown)


aspnetcore/host-and-deploy/azure-apps/index.md
```diff
@@ -202,6 +202,13 @@ When deploying a self-contained app:
 
 The [Docker Hub](https://hub.docker.com/r/microsoft/aspnetcore/) contains the latest preview Docker images. The images can be used as a base image. Use the image and deploy to Web Apps for Containers normally.
 
+## Protocol settings (HTTPS)
+
+Secure protocol bindings allow you specify which certificate to use when responding to requests to a specific hostname over HTTPS. Binding requires a valid private certificate (*.pfx*) issued for the specific hostname. For more information, see the following topics:
+
+* [Tutorial: Bind an existing custom SSL certificate to Azure Web Apps](/azure/app-service/app-service-web-tutorial-custom-ssl)
+* [How To Configure TLS Mutual Authentication for Web App](/azure/app-service/app-service-web-configure-tls-mutual-auth)
```
shirhatti: Why this link? I understand TLS is a pre-req for mTLS but it does more to distract than help in this article

guardrex: They put this link at the end of their **SSL bindings** description.

addbinding

How about this one ... https://docs.microsoft.com/azure/app-service/web-sites-purchase-ssl-web-site ... or shall I just remove the 2nd link.

shirhatti: My vote is to remove the link.

PR #9350, opened 2018-10-31
Fix path for _Layout.cshtml

KexyBiscuit:
Some of the links was fixed before. This should be the last one.


Rick-Anderson:
@KexyBiscuit 	thanks for fixing this. The includes make it messy. We'll be removing them in the next month.

PR #9377, opened 2018-11-01
Explain relationship between model validation and [ApiController]

scottaddie:
Fixes https://github.com/aspnet/Docs/issues/9307

**Internal Review Pages**
* [Web API doc](https://review.docs.microsoft.com/en-us/aspnet/core/web-api/?branch=pr-en-us-9377&view=aspnetcore-2.1#automatic-http-400-responses)
* [Model validation doc](https://review.docs.microsoft.com/en-us/aspnet/core/mvc/models/validation?branch=pr-en-us-9377&view=aspnetcore-2.1#handle-model-state-errors)

### DIFF
--- a/aspnetcore/mvc/models/validation.md
+++ b/aspnetcore/mvc/models/validation.md
@@ -4,7 +4,7 @@ author: tdykstra
 description: Learn about model validation in ASP.NET Core MVC.
 ms.author: riande
 ms.custom: mvc
-ms.date: 10/24/2018
+ms.date: 11/01/2018
 uid: mvc/models/validation
 ---
 # Model validation in ASP.NET Core MVC
@@ -27,10 +27,10 @@ Validation attributes are a way to configure model validation so it's similar co
 
 Validation attributes are specified at the property level: 
 
-```csharp 
-[Required] 
+```csharp
+[Required]
 public string MyProperty { get; set; } 
-``` 
+```
 
 Below is an annotated `Movie` model from an app that stores information about movies and TV shows. Most of the properties are required and several string properties have length requirements. Additionally, there's a numeric range restriction in place for the `Price` property from 0 to $999.99, along with a custom validation attribute.
 
@@ -76,13 +76,19 @@ Client-side validation requires a value for a form field that corresponds to a m
 
 Model state represents validation errors in submitted HTML form values.
 
-MVC will continue validating fields until reaches the maximum number of errors (200 by default). You can configure this number by inserting the following code into the `ConfigureServices` method in the *Startup.cs* file:
+MVC will continue validating fields until it reaches the maximum number of errors (200 by default). You can configure this number with the following code in `Startup.ConfigureServices`:
 
 [!code-csharp[](validation/sample/Startup.cs?range=27)]
 
-## Handling Model State Errors
+## Handle Model State errors
 
-Model validation occurs prior to each controller action being invoked, and it's the action method's responsibility to inspect `ModelState.IsValid` and react appropriately. In many cases, the appropriate reaction is to return an error response, ideally detailing the reason why model validation failed.
+Model validation occurs before the execution of a controller action. It's the action's responsibility to inspect `ModelState.IsValid` and react appropriately. In many cases, the appropriate reaction is to return an error response, ideally detailing the reason why model validation failed.
+
+::: moniker range=">= aspnetcore-2.1"
+
+For ASP.NET Core 2.1 or later web API controllers using the `[ApiController]` attribute, model validation occurs automatically. For more information, see [Automatic HTTP 400 responses](xref:web-api/index#automatic-http-400-responses).
+
+::: moniker-end
 
 Some apps will choose to follow a standard convention for dealing with model validation errors, in which case a filter may be an appropriate place to implement such a policy. You should test how your actions behave with valid and invalid model states.
 
--- a/aspnetcore/mvc/models/validation.md
+++ b/aspnetcore/mvc/models/validation.md
@@ -4,7 +4,7 @@ author: tdykstra
 description: Learn about model validation in ASP.NET Core MVC.
 ms.author: riande
 ms.custom: mvc
-ms.date: 11/01/2018
+ms.date: 11/06/2018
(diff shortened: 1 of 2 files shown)


aspnetcore/web-api/index.md
```diff
@@ -71,11 +71,14 @@ The following sections describe convenience features added by the attribute.
...
-Validation errors automatically trigger an HTTP 400 response. The following code becomes unnecessary in your actions:
+Model validation errors automatically trigger an HTTP 400 response. Consequently, the following code becomes unnecessary in your actions:
 
 [!code-csharp[](define-controller/samples/WebApiSample.Api.Pre21/Controllers/PetsController.cs?name=snippet_ModelStateIsValidCheck)]
 
-Use <xref:Microsoft.AspNetCore.Mvc.ApiBehaviorOptions.InvalidModelStateResponseFactory> to customize the output of the resulting response.
+Disabling this behavior is useful when your action:
+
+* Can recover from a model validation error.
+* Should return a customized response using <xref:Microsoft.AspNetCore.Mvc.ApiBehaviorOptions.InvalidModelStateResponseFactory>.
```
dougbu: This bullet is now backwards. The `InvalidModelStateResponseFactory` is used when `[ApiController]` applies or the `ModelStateInvalidFilter` has been manually configured. The previous sentence was correct.

But, in addition, users may wish to remove `[ApiController]` or set `ApiBehaviorOptions.SuppressModelStateInvalidFilter` to `true` if their action can recover from a model validation error.

aspnetcore/mvc/models/validation.md
```diff
@@ -76,13 +76,19 @@ Client-side validation requires a value for a form field that corresponds to a m
...
 [!code-csharp[](validation/sample/Startup.cs?range=27)]
 
-## Handling Model State Errors
+## Handle Model State errors
+
+Model validation occurs before the execution of a controller action. It's the action's responsibility to inspect `ModelState.IsValid` and react appropriately. In many cases, the appropriate reaction is to return an error response, ideally detailing the reason why model validation failed.
+
+::: moniker range=">= aspnetcore-2.1"
+
+For ASP.NET Core 2.1 or later web API controllers using the `[ApiController]` attribute, model validation occurs automatically. For more information, see [Automatic HTTP 400 responses](xref:web-api/index#automatic-http-400-responses).
```
dougbu: No "model validation _always_ occurs automatically". The difference with `[ApiController]` is the automatic 400 response containing details of issues when `ModelState.IsValid == false`.

aspnetcore/mvc/models/validation.md
```diff
@@ -76,13 +76,19 @@ Client-side validation requires a value for a form field that corresponds to a m
 
 Model state represents validation errors in submitted HTML form values.
 
-MVC will continue validating fields until reaches the maximum number of errors (200 by default). You can configure this number by inserting the following code into the `ConfigureServices` method in the *Startup.cs* file:
+MVC will continue validating fields until it reaches the maximum number of errors (200 by default). You can configure this number with the following code in `Startup.ConfigureServices`:
```
dougbu: MVC has an additional option which will stop validation before it's done: `MvcOptions.MaxValidationDepth` constrains the visitor to recursing through the model only to the specified depth.  The default value is `200` when running with `CompatibilityVersion.Version_2_2` or later and `null` i.e. unconstrained when running with an earlier `CompatibilityVersion`.

dougbu:
I find these two pages difficult to read. Information is misleading or unclear in a few places. For example,
> MVC model binding, which isn't concerned with validation and validation attributes, rejects a form field submission containing a missing value or whitespace for a non-nullable type. In the absence of a BindRequired attribute on the target property, model binding ignores missing data for non-nullable types, where the form field is absent from the incoming form data.

The two sentences in the above paragraph describe two different default behaviours. Later paragraphs better describe how `[Required]` is handled client- and server-side as well as how `[BindRequired]` is handled server-side. But the first sentence above starts off on the wrong foot.

1. By default, non-nullable values are treated as if the parameter or property had a `[Required]` attribute
2. `[Required]` is enforced client- and server-side
  - Server-side this attribute has the default .NET semantics i.e. non-nullable values _cannot_ be invalid according to the attribute
  - Client-side validation (if Unobtrusive validation is enabled e.g. through including `_ValidationScriptsPartial.cshtml` in a page or view) adds an additional semantic that even non-nullable value types must have a value (a case that cannot be detected server-side)
3. `[BindRequired]` controls whether MVC model binding requires submitted data for a parameter or property
  - Unlike `[Required]`, this attribute is enforced server-side only and before validation
  - Unlike `[Required]`, if `[BindRequired]` is applied, the model binding system will detect when information is not submitted for a non-nullable value type and add a `ModelState` error

So,
> MVC model binding, which isn't concerned with validation and validation attributes, rejects a form field submission containing a missing value or whitespace for a non-nullable type. 

This is _not_ correct w.r.t. model binding. A writer may have been confused between client-side (Unobtrusive) validation and model binding.

In addition, client-side validation treats whitespace as valid even if `[Required]` applies to a `string` parameter or property. See https://jqueryvalidation.org/required-method/

> where the form field is absent from the incoming form data.

This bit is confusing. Barring some very strange JavaScript, a form field _cannot_ be absent from the form data. And, model binding accepts values from many sources e.g. the query string and route values. `[BindRequired]` validations usually fail when a form is missing a field or the user goes to an unexpected URI e.g. one lacking a required route value.

dougbu:
Should also mention that the wording sometimes fails to distinguish new options from things that are enabled with certain `CompatibiltyVersion`s.

Not sure I caught any examples, but I have a vague feeling that a couple of "with 2.1" cases that are controlled using the `CompatibiltyVersion` don't mention the associated option. `CompatibiltyVersion`s _never_ lack an override in one of the options classes e.g. `MvcOptions`.

scottaddie:
@dougbu I really appreciate the great feedback. I've opened issue https://github.com/aspnet/Docs/issues/9481 to address it. I prefer to keep this PR scoped to the relationship between [ApiController] and model binding. With that in mind, how do the latest changes look?

PR #9435, opened 2018-11-05
Update index.md

amadosoft:
"ArrayExamples" doesn't exist, should be "ArrayExample"


amadosoft:
you're welcome @guardrex 😃

PR #9456, opened 2018-11-05
Update options for 2.2

HaoK:
Fixes https://github.com/aspnet/Docs/issues/9443
Fixes https://github.com/aspnet/Docs/issues/9232


HaoK:
@guardrex can you review?

guardrex:
Sure thing @HaoK ... to save time, should I just edit any grammar/style things directly, or would you prefer comments?

HaoK:
Nah just edit away, I noticed my wordsmithing skills were...not great compared to the rest of the doc, hopefully you can help fix that :)

guardrex:
@HaoK 

> was introduced in 2.1 to 

> has been a source of friction and prior to 2.1

> Starting in 2.1

> is something we are still thinking about for a future release

We were just chatting about this on another issue: We try to doc the current scenario for the current version and not make reference to prior versions, plans for the future, and reasons for the changes. We just let the versioning of the content take care of it. We're up to ~400 markdown files now, and this approach helps.

Are u ok if I revise without that language? ... then you can tell me what a *goober* I am on review? ... that's our normal pattern, isn't it? 😀 lol

HaoK:
Sounds good @guardrex

guardrex:
If the language that calls out explicit versions is ok, yes ... merge now.

If I should go thru and try to eliminate language that refers to specific versions, then I can do that tomorrow (Fri, 11/9) ... hold for a day.

Rick-Anderson:
Tomorrow is fine.

HaoK:
Reads much better thanks for polishing it up @guardrex !  Looks good to me

PR #9462, opened 2018-11-05
added .net client discussion

### LINKED ISSUE #9460: [SignalR] SignalR .NET Clients page doesn't call out "client environments"
In [this](https://docs.microsoft.com/en-us/aspnet/core/signalr/supported-platforms) article, we cover the details of the .NET client for SignalR, but we don't call out any "client" scenarios, like Xamarin or desktop client apps, and in fact mainly call out "server" environments. 

We should call out some of the client environments, like Xamarin, where the SignalR .NET Client could be used to provide users with real-time updates when server-side "things" happen.

bradygaster:
fixes #9460, by limiting the "server side" language in the .net client segment and by adding a brief discussion on Xamarin, which links off to the issue we created to help Xamarin developers get going.

### DIFF
--- a/aspnetcore/signalr/supported-platforms.md
+++ b/aspnetcore/signalr/supported-platforms.md
@@ -29,7 +29,7 @@ The [JavaScript client](https://www.npmjs.com/package/@aspnet/signalr) runs on N
  
 ## .NET client
 
-The [.NET client](https://www.nuget.org/packages/Microsoft.AspNetCore.SignalR/) runs on any server platform supported by ASP.NET Core.
+The [.NET client](https://www.nuget.org/packages/Microsoft.AspNetCore.SignalR/) can be used by developers building apps using .NET Core. [Xamarin developers can use SignalR](https://github.com/aspnet/Announcements/issues/305) for building applications for Android using Xamarin.Android 8.4.0.1 and later, and for iOS using Xamarin.Android 11.14.0.4 and later. 
 
 When the server runs IIS, the WebSockets transport requires IIS 8.0 or higher, on Windows Server 2012 or higher. Other transports are supported on all platforms.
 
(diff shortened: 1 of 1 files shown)


aspnetcore/signalr/supported-platforms.md
```diff
@@ -29,7 +29,7 @@ The [JavaScript client](https://www.npmjs.com/package/@aspnet/signalr) runs on N
  
 ## .NET client
 
-The [.NET client](https://www.nuget.org/packages/Microsoft.AspNetCore.SignalR/) runs on any server platform supported by ASP.NET Core.
+The [.NET client](https://www.nuget.org/packages/Microsoft.AspNetCore.SignalR/) can be used by developers using .NET Core. [Xamarin developers can use SignalR](https://github.com/aspnet/Announcements/issues/305) for building applications for Android using Xamarin.Android 8.4.0.1 and later, and for iOS using Xamarin.Android 11.14.0.4 and later.
```
scottaddie: What about ASP.NET Core projects targeting .NET Framework? I thought this Microsoft.AspNetCore.SignalR package worked there too.

tdykstra: You could retain the original "runs on any platform supported by .NET Core" by just dropping "server" and "ASP" before .NET.  Then put "For example" in front of your Xamarin sentence.  Then I don't think you need to mention explicitly the ability of ASP.NET Core to target Framework, as that platform would be Windows, which is covered by "any platform supported by .NET Core".

sfarmer-MesaAZ: Shouldn't that read as "and for iOS using **Xamarin.iOS** 11.14.0.4 and later" ?

bradygaster: Absolutely. Thanks for the catch @sfarmer-MesaAZ

scottaddie:
Thank you for the PR, @bradygaster!

Issue #9441, opened 2018-11-05
ANCM In-Process documentation


jkotalik:
Overall, ANCM In-Process documentation still needs a lot of work. Topics missing include
- Documentation on new module (AspNetCoreModuleV2) and associated changes
- Getting started with ANCM Inproc
- New ways of gathering information (Logs and debug logs, environment variable activation, etc.)
- Diagnostics (new startup pages) => #9452 resolves this
- Managed server configuration options
- IServerVariableFeature

guardrex:
@jkotalik Are you looking at the 2.2 version of the ANCM ref doc?

https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/aspnet-core-module?view=aspnetcore-2.2

jkotalik:
Yep. 

> Documentation on new module (AspNetCoreModuleV2) and associated changes

This refers to how ANCM is now split into two different dlls and architectural changes associated.

> Getting started with ANCM Inproc

@shirhatti will be taking care of this.

> New ways of gathering information (Logs and debug logs, environment variable activation, etc.)

I think this is covered here: https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/aspnet-core-module?view=aspnetcore-2.2#enhanced-diagnostic-logs, but want to do another pass through.

> Diagnostics (new startup pages)

We added a few new startup pages in 2.2-preview3. @shirhatti we need to make sure the hyperlinks on the startup pages redirect to something 😄

guardrex:
> This refers to how ANCM is now split into two different dlls and architectural changes associated.

> Getting started with ANCM Inproc

Just want to mention in passing that we doc the current approach for the current version. We try not to use 'new in 2.2 you can'- and 'you used to do it this way'-types of remarks in reference docs. Of course, the blog posts make such remarks. Just want to make sure we're on the same page ... it's 💰 **_costly_** 💰 to make changes then go back in later and refactor the changes.

> new startup pages in 2.2-preview3

:+1: Very good ... it's the first I'm hearing about it.

guardrex:
@shirhatti Note that we want to defer configuration details to the ANCM reference topic as much as possible.

We're lightly discussing it on https://github.com/aspnet/Docs/issues/9416 ... e.g., imo there's far (**_FAR_** :smile: lol) too much config now for Kestrel to be in a "fundamentals"-type of topic. The best thing imo for fundamentals is to refer back to the ref topic in *Host and deploy* for the juicy details.

[...]

PR #9463, opened 2018-11-05
Update IIS index.md

jkotalik:
More for #9441 with a bit of refactoring.

- IISServerOptions (inproc) are unique from IISOptions (out of proc)
- Move HTTP/2 topic lower in the document (flow felt off by having it before how to enable IISIntegration)
- Remove stop/starting app pool in favor of dropping app_offline (@shirhatti for thoughts)
- Generalize addressing AspNetCore behind IIS (remove reverse proxy stuff).

### DIFF
--- a/aspnetcore/host-and-deploy/iis/index.md
+++ b/aspnetcore/host-and-deploy/iis/index.md
@@ -9,7 +9,7 @@ uid: host-and-deploy/iis/index
 ---
 # Host ASP.NET Core on Windows with IIS
 
-By [Luke Latham](https://github.com/guardrex)
+By [Luke Latham](https://github.com/guardrex) and [Justin Kotalik](https://github.com/jkotalik)
 
 [Install the .NET Core Hosting Bundle](#install-the-net-core-hosting-bundle)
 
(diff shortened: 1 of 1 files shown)


guardrex:
> Remove stop/starting app pool in favor of dropping app_offline

I was under the impression that some devs were still having issues with the app_offline drop but that the PS command to stop the app pool *always* worked. I just mean that no one ever said that they ran the command and it failed; whereas, there's an issue where devs express some on-going problems with the file drop approach.

If the PS command is pulled, then it doesn't surface (perhaps) a sure-fire way to go if all else fails. However, it could be one of *ye olde* **_RexHacks_**:tm: that the docs really shouldn't show. :smile: If both approaches could appear here, that's another option.

jkotalik:
In my opinion, if app_offline doesn't work locally, we shouldn't suggest it to users. I know users have had difficulties in the past with it, and if people still have issues, we can add it back.

guardrex:
Fair enough. On Slack they would say ...

> ✨ `Stop-WebAppPool` ✨ Rex! You're a damn bonifide rockstar GENIUS!

It only happened a few times (or once lol).

jkotalik:
I can keep it, but I'll put it as a last resort and mention why it isn't a great idea 😄

aspnetcore/host-and-deploy/iis/index.md
```diff
@@ -161,6 +124,24 @@ For more information on hosting, see [Host in ASP.NET Core](xref:fundamentals/ho
 
 ### IIS options
 
+**In-process hosting model**
```
guardrex: OH ... one more thing here. This new content should be versioned for 2.2 or later.

aspnetcore/host-and-deploy/iis/index.md
```diff
@@ -524,6 +500,43 @@ ICACLS C:\sites\MyWebApp /grant "IIS AppPool\DefaultAppPool":F
...
 
+## HTTP/2 support
+
+::: moniker range=">= aspnetcore-2.2"
+
+[HTTP/2](https://httpwg.org/specs/rfc7540.html) is supported with ASP.NET Core in the following IIS deployment scenarios:
+
+* In-process
+  * Windows Server 2016/Windows 10 or later; IIS 10 or later
+  * Target framework: .NET Core 2.2 or later
```
shirhatti: Redundant since we're already in the 2.2 section of the doc

aspnetcore/host-and-deploy/iis/index.md
```diff
@@ -524,6 +500,43 @@ ICACLS C:\sites\MyWebApp /grant "IIS AppPool\DefaultAppPool":F
...
+  * TLS 1.2 or later connection
+* Out-of-process
+  * Windows Server 2016/Windows 10 or later; IIS 10 or later
+  * Public-facing edge server connections use HTTP/2, but the reverse proxy connection to the [Kestrel server](xref:fundamentals/servers/kestrel) uses HTTP/1.1.
+  * Target framework: Not applicable to out-of-process deployments, since the HTTP/2 connection is handled entirely by IIS.
```
shirhatti: With above comment, we can drop this as well

aspnetcore/host-and-deploy/iis/index.md
```diff
@@ -524,6 +500,43 @@ ICACLS C:\sites\MyWebApp /grant "IIS AppPool\DefaultAppPool":F
...
+  * TLS 1.2 or later connection
+
+For an in-process deployment when an HTTP/2 connection is established, [HttpRequest.Protocol](xref:Microsoft.AspNetCore.Http.HttpRequest.Protocol*) reports `HTTP/2`. For an out-of-process deployment when an HTTP/2 connection is established, [HttpRequest.Protocol](xref:Microsoft.AspNetCore.Http.HttpRequest.Protocol*) reports `HTTP/1.1`.
+
+For more information on the in-process and out-of-process hosting models, see the <xref:fundamentals/servers/aspnet-core-module> topic and the <xref:host-and-deploy/aspnet-core-module>.
```
tdykstra: see the x article
or 
see x

either way is ok but it would be best to be consistent

aspnetcore/host-and-deploy/iis/index.md
```diff
@@ -161,6 +124,24 @@ For more information on hosting, see [Host in ASP.NET Core](xref:fundamentals/ho
...
+```csharp
+services.Configure<IISServerOptions>(options => 
+{
+    options.ForwardWindowsAuthentication = false;
+});
+```
+
+| Option                         | Default | Setting |
+| ------------------------------ | :-----: | ------- |
+| `AutomaticAuthentication`      | `true`  | If `true`, IIS Server sets the `HttpContext.User` authenticated by [Windows Authentication](xref:security/authentication/windowsauth). If `false`, the server only provides an identity for `HttpContext.User` and responds to challenges when explicitly requested by the `AuthenticationScheme`. Windows Authentication must be enabled in IIS for `AutomaticAuthentication` to function. For more information, see the [Windows Authentication](xref:security/authentication/windowsauth) topic. |
```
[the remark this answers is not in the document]
jkotalik: I actually forgot that ForwardWindowsAuthentication isn't a public option. Changing example to use AutomaticAuthentication.

aspnetcore/host-and-deploy/iis/index.md
```diff
@@ -24,62 +24,31 @@ The following operating systems are supported:
...
-## Application configuration
-
-### Enable the IISIntegration components
-
 ::: moniker range=">= aspnetcore-2.2"
 
 **In-process hosting model**
 
-A typical *Program.cs* calls <xref:Microsoft.AspNetCore.WebHost.CreateDefaultBuilder*> to begin setting up a host. `CreateDefaultBuilder` calls the `UseIIS` method to boot the [CoreCLR](/dotnet/standard/glossary#coreclr) and host the app inside of the IIS worker process (`w3wp.exe`). Performance tests indicate that hosting a .NET Core app in-process delivers higher request throughput compared to hosting the app out-of-process and proxying requests to [Kestrel](xref:fundamentals/servers/kestrel).
+`CreateDefaultBuilder` calls the `UseIIS` method to boot the [CoreCLR](/dotnet/standard/glossary#coreclr) and host the app inside of the IIS worker process (`w3wp.exe` or `iisexpress.exe`). Performance tests indicate that hosting a .NET Core app in-process delivers significantly higher request throughput compared to hosting the app out-of-process and proxying requests to [Kestrel](xref:fundamentals/servers/kestrel).
```
scottaddie: Nit: Italicize these 2 file names instead of code-fencing them.

guardrex: I didn't remark because these are described as "worker process[es]." Could go either way: Is the file **_the_** worker process itself? Maybe so, maybe not.

Side-note: You're welcome to keep "significantly," but it adds no real value because it isn't specific. It's kind of like the ban on words like "simple" and "easy" here ... they don't add much more than marketing-speak.

aspnetcore/host-and-deploy/iis/index.md
guardrex: That's me ... trying to phrase it to roll off the 👅 a bit better. The titles are ...

* ASP.NET Core Module
* ASP.NET Core Module configuration reference

It would read like this in its present form ...

> For more information on the in-process and out-of-process hosting models, see the [ASP.NET Core Module](#) topic and the [ASP.NET Core Module configuration reference](#).

... but for consistency the other way it will be ...

> For more information on the in-process and out-of-process hosting models, see [ASP.NET Core Module](#) and [ASP.NET Core Module configuration reference](#).

Your call, but I like my version better. :smile: lol

aspnetcore/host-and-deploy/iis/index.md
```diff
@@ -488,6 +458,12 @@ ASP.NET Core apps are configured using other configuration providers. For more i
 
 When hosting multiple websites on a server, we recommend isolating the apps from each other by running each app in its own app pool. The IIS **Add Website** dialog defaults to this configuration. When **Site name** is provided, the text is automatically transferred to the **Application pool** textbox. A new app pool is created using the site name when the site is added.
 
+::: moniker range=">= aspnetcore-2.2"
+
+When using in-process hosting, application are required to run in separate app pools.
```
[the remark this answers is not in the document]
guardrex: ... and we use "app[s]" everywhere, so I recommend "apps" here.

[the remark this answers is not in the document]
tdykstra: Worth remembering, we don't use app everywhere -- we use application when it sounds better in context.  It's up to the writer to decide when that is the case.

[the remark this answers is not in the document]
guardrex: I was told by @Rick-Anderson that unless it's part of a product/service official name *to use it everywhere*. I don't think it's been an issue. I mean I don't recall a case where it sounded better as "application." As you say, the writer's :ear: might hear different things.

aspnetcore/host-and-deploy/iis/index.md
tdykstra: I prefer the second way you present it.  The first way is inconsistent in the way "the" goes with "reference" but "reference" is in the link text, whereas "topic" is outside of the link text.

guardrex: We were ... afaik ... always *only* linking the title. Every other word outside of the title hasn't been part of the link, except for the metapackage topic link for `App` and `All` given that those could link to the nuget.org package, so we have the word "metapackage" in the link text.

I do 99% agree and almost always .... 99% :smile: of the time ... link with "For more information, see \[link text](LINK)." ... done and done. It was just that in this case it seemed to roll off the tongue a little better.

Rick-Anderson:
@guardrex 	can you make a commit with these edit changes? @jkotalik 	can review the commit and you can back out anything he doesn't like. Alternatively you could create a new PR to merge into this branch.

jkotalik:
@Rick-Anderson I addressed most of the feedback already, I'll just continue to update this PR.

guardrex:
> I'll just continue to update this PR.

:+1:

We still should discuss removing "reverse proxy" language: There are ~97 references across the docs. The replacement language is to say something along the lines of ... **_the ANCM allows an app to run behind IIS_** ... so we can mostly (or maybe even entirely) remove "reverse proxy" from the other topics if that's a goal.

jkotalik:
> We still should discuss removing "reverse proxy" language: There are ~97 references across the docs. The replacement language is to say something along the lines of ... the ANCM allows an app to run behind IIS ... so we can mostly (or maybe even entirely) remove "reverse proxy" from the other topics if that's a goal.

Separate issue/PR?

aspnetcore/host-and-deploy/iis/index.md
```diff
@@ -486,7 +454,22 @@ ASP.NET Core apps are configured using other configuration providers. For more i
 
 ## Application Pools
 
-When hosting multiple websites on a server, we recommend isolating the apps from each other by running each app in its own app pool. The IIS **Add Website** dialog defaults to this configuration. When **Site name** is provided, the text is automatically transferred to the **Application pool** textbox. A new app pool is created using the site name when the site is added.
+::: moniker range=">= aspnetcore-2.2"
+
```
guardrex: Now that I look at the break-out on paper, it seems strange (and *I wrote IT.* 🙈).

Would this be better? ...

```
App pool isolation is determined by the hosting model:

* In-process hosting &ndash; Apps are required to run in separate app pools.
* Out-of-process hosting &ndash; We recommend isolating the apps from each other by running each app in its own app pool.
```

guardrex:
I'm only ever really unhappy when the :beer: runs out. *Cheers!*

Thanks again @jkotalik for your help with the topics.

scottaddie:
Thank you for your help, @jkotalik!

PR #9493, opened 2018-11-07
Docs: Mentioned that the Controller name replacement is by convention

danielabbatt:
Making it a bit clearer that the [controller] name replacement of MyNameController is by naming convention.


danielabbatt:
Reason for suggesting this was that during a training session and there was some confusion/interest in how the framework knew that it was MyNameController that was matched on. I explained it was by naming convention.

Rick-Anderson:
@danielabbatt 	nice improvement, thanks.

PR #9517, opened 2018-11-08
Update options.md

amadosoft:
i think it's "ConfigureServices" not "Configure".


amadosoft:
yeah. that happens. lol

guardrex:
This might come as a slight surprise to many readers, but usually **_THREE_** of us look these over before they go live. 🙈🙉🙊 ...... *bad* things happen to *good* people! :cry: 😄 :cry: 😄 :cry: 😄

PR #9560, opened 2018-11-12
Update Login.cshtml.cs to use method parameter

VincentPestana:
AuthenticateUser now checks the passed in parameter, instead of the class Input variable.


VincentPestana:
@guardrex Thanks to you and the team for the great examples, I would be lost without some.

PR #9557, opened 2018-11-12
Update path to `signalr.js`

dmorrison42:
The sample code for the ASP.NET Core SignalR does not work due to a path issue.

It appears that it should refer to [signalr.js](https://github.com/aspnet/Docs/blob/master/aspnetcore/tutorials/signalr/sample/wwwroot/lib/signalr/signalr.js), but this file has moved a few times in the past so I'm not sure if you want to fix this at the reference or the source.

Thanks!

PR #9564, opened 2018-11-13
Fixes #9557 The sample code for the ASP.NET Core SignalR does not work due to a path issue.

dmorrison42:
This addresses the issue #9557 that I was working on with @scottaddie.

Unfortunately, I experienced severe GitHub workflow issues, and the pull request was closed.


dmorrison42:
@scottaddie, all set!

I am experiencing some strange behavior if I do `dotnet run` vs F5, but I believe that it is a separate issue. I think it has something to do with referencing minified files that get built somewhere else.

image

Assuming you agree that is either user error, or at least a separate issue, I'm good to merge.

scottaddie:
@dmorrison42 I'm not seeing that behavior when I run `dotnet run`. I suspect your `ASPNETCORE_ENVIRONMENT` variable is set to something other than `Development`. Because of that, the *_Layout.cshtml* file is attempting to load the minified files, which likely don't exist in your project.

scottaddie:
@dmorrison42 Thank you for your patience and help with fixing this issue. Your changes will go live within the next 24 hours.

PR #9567, opened 2018-11-13
Version the 1.1 PDF of the Kestrel and Web Host topics

### LINKED ISSUE #9565: Use version moniker
For the 1.1 version of this topic, download Kestrel web server implementation in ASP.NET Core (version 1.1, PDF).

That's not a great lead for 99.9% of customers. 

---
#### Document Details

* ID: f22dcddb-0c25-be1a-9a02-9e71dce86bdb
* Version Independent ID: 3e0e12c2-e544-5655-e001-d2ec180b11a8
* Content: [Kestrel web server implementation in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel?view=aspnetcore-2.1)
* Content Source: [aspnetcore/fundamentals/servers/kestrel.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/servers/kestrel.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex

guardrex:
Fixes #9565

### DIFF
--- a/aspnetcore/fundamentals/host/web-host.md
+++ b/aspnetcore/fundamentals/host/web-host.md
@@ -11,7 +11,6 @@ uid: fundamentals/host/web-host
 
 By [Luke Latham](https://github.com/guardrex)
 
-
 ::: moniker range="<= aspnetcore-1.1"
 
 For the 1.1 version of this topic, download [ASP.NET Core Web Host (version 1.1, PDF)](https://webpifeed.blob.core.windows.net/webpifeed/Partners/Web-Host_1.1.pdf).


guardrex:
Cool ... no probs. There's one other ... I'll patch that here. brb

guardrex:
Let me know if you prefer to make that <= 1.1.

Now that we know this approach works, we could even have an = 2.0 remark that says something to effect that 2.0 has reached EOL and to see the 2.1 version (link it).

guardrex:
One more quick note: I was right about the warning, but it isn't on the build itself ... it's limited to the email that is generated about the build. It comes back with ...

> Line : [Warning] No intersection between zone and file level monikers. The result of zone level range string `< aspnetcore-2.1` is , while file level monikers is aspnetcore-2.1,aspnetcore-2.2.

... and that's on the build details here :point_right: https://opbuildstorageprod.blob.core.windows.net/report/2018%5C11%5C13%5C6a0d29ae-bb37-09f1-c55a-4cfa862c6784%5CPullRequest%5C201811130117356690-9567%5Cworkflow_report.html?sv=2015-02-21&sr=b&sig=x44U6fuz9rJFloUp6f8mOo%2FWv0Ms7tFKWzzEbgBFkFc%3D&st=2018-11-13T01%3A19%3A12Z&se=2018-12-14T01%3A24%3A12Z&sp=r

... so there is some concern about whatever this ends up rendering correctly. I'll put an internal topic link here so that someone can check ...

[Internal Review Topic (Kestrel)](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel?view=aspnetcore-2.1&branch=pr-en-us-9567)

Rick-Anderson:
>Now that we know this approach works, we could even have an = 2.0 remark that says something to effect that 2.0 has reached EOL and to see the 2.1 version (link it).

That sounds like a great plan. Create an include and use a regex to put that in every file. That would really pat your numbers.

@scottaddie 	@tdykstra 	 is that over-kill?

guardrex:
Right now (based on this PR), someone could load the 2.0 version and receive 2.1 content without any indication that the topic is really a 2.1 topic.

If the build system isn't modified to deal with this scenario ... and if we can't merge with the warning and have it work that way ... and if we don't provision some text at the top for 2.0 ... 

... I don't think what we have in the PR right now is going to work out. One of those approaches (or something else) is needed.

tdykstra:
>  Create an include and use a regex to put that in every file. That would really pat your numbers.
> is that over-kill?

I think so

Rick-Anderson:
@guardrex 	create the include and use it where you need it.

guardrex:
> Create the include and use it where you need it.

I can change the link text on these, but how do I handle the link itself? ... it will change for every include. For example ...

> For the 1.1 version of this topic, download \[URL Rewriting Middleware in ASP.NET Core (version 1.1, PDF)](https://webpifeed.blob.core.windows.net/webpifeed/Partners/URL_Rewriting_1.1.pdf).

Rick-Anderson:
Never mind. We'll figure this out later.

PR #9569, opened 2018-11-13
Fixing the wording of a sentence

stevejgordon:
This keeps the pluralisation consistent within the sentence and in my opinion, reads better.


guardrex:
Thanks @stevejgordon ... I think ur right ... and I take a stab at simplifying the sentences a bit.

PR #9602, opened 2018-11-15
Update title of button in HTML code for clarity

alistairmcmillan:
Minor change but it seems a little confusing to call the button "Edit" when you use it to "Save" or "Submit" your changed todo item.


scottaddie:
@alistairmcmillan Thank you for your PR! I agree that this change makes things much clearer.

PR #9645, opened 2018-11-20
Riande/perf best practices2

### LINKED ISSUE #9526: Performance best practices
@mjrousos to provide first draft.

Rick-Anderson:
Adds a document listing common performance pitfalls in ASP.NET Core apps and how to avoid them.

Fixes #9526
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/performance/performance-best-practices?view=aspnetcore-2.1&branch=pr-en-us-9645)

### DIFF
--- a/aspnetcore/performance/performance-best-practices.md
+++ b/aspnetcore/performance/performance-best-practices.md
@@ -43,11 +43,11 @@ A profiler like [PerfView](https://github.com/Microsoft/perfview) can be used to
 ## Minimize large object allocations
 
 <!-- TODO review Bill - replaced original .NET language below with .NET Core since this targets .NET Core -->
-The [.NET Core garbage collector](/dotnet/standard/garbage-collection/) manages allocation and release of memory automatically in ASP.NET Core apps. Automatic garbage collection generally means that developers don't need to worry about how or when memory is freed. However, cleaning up unreferenced objects takes CPU time, so developers should minimize allocating objects in [hot code paths](#hot). Garbage collection is especially expensive on large objects (> 85K bytes). Large objects are stored on the [large object heap](/dotnet/standard/garbage-collection/large-object-heap) and require a full (generation 2) garbage collection to clean up. Unlike generation 0 and generation 1 collections, a generation 2 collection requires app execution to be temporarily suspended. Frequent allocation and de-allocation of large objects can cause inconsistent performance.
+The [.NET Core garbage collector](/dotnet/standard/garbage-collection/) manages allocation and release of memory automatically in ASP.NET Core apps. Automatic garbage collection generally means that developers don't need to worry about how or when memory is freed. However, cleaning up unreferenced objects takes CPU time, so developers should minimize allocating objects in [hot code paths](#hot). Garbage collection is especially expensive on large objects (> 85 K bytes). Large objects are stored on the [large object heap](/dotnet/standard/garbage-collection/large-object-heap) and require a full (generation 2) garbage collection to clean up. Unlike generation 0 and generation 1 collections, a generation 2 collection requires app execution to be temporarily suspended. Frequent allocation and de-allocation of large objects can cause inconsistent performance.
 
 Recommendations:
 
-* **Do** consider caching large objects that are frequently used so that they don't need to be reallocated each time they're needed.
+* **Do** consider caching large objects that are frequently used. Caching large objects prevents expensive allocations.
 * **Do** pool buffers by using an [`ArrayPool<T>`](/dotnet/api/system.buffers.arraypool-1) to store large arrays.
 * **Do not** allocate many, short-lived large objects on [hot code paths](#hot).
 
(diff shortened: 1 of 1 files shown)


aspnetcore/performance/performance-best-practices.md
```diff
@@ -0,0 +1,132 @@
...
+Recommendations:
+
+* **Do** use ASP.NET Core's [built-in support](xref:client-side/bundling-and-minification) for bundling and minifying client assets.
+* **Do** consider other third-party tools like [Gulp](uid:client-side/bundling-and-minification#consume-bundleconfigjson-from-gulp) or [Webpack](https://webpack.js.org/) for more complex client asset management.
+
+## Use the latest ASP.NET Core release
+
+Each new release of ASP.NET includes performance improvements. Optimizations in .NET Core and ASP.NET Core mean that newer versions will outperform older versions. For example, .NET Core 2.1 added support for compiled regular expressions and benefitted from [`Span<T>`](https://msdn.microsoft.com/magazine/mt814808.aspx). ASP.NET Core 2.2 added support for HTTP/2. If performance is a priority, consider upgrading to the most current version of ASP.NET Core.
+
+<!-- TODO review link and taking advantage of new [performance features](#TBD)
```
mjrousos: I think there's value in having section of our docs that groups performance features (things like caching, response compression, etc.). This could just link to that section.

aspnetcore/performance/performance-best-practices.md
```diff
@@ -0,0 +1,132 @@
...
+Maybe skip this TBD link as each version will have perf improvements -->
+
+<!-- TODO review Do we want to say:
+## Minimize exceptions
+
+Exceptions should be rare. Avoid using exceptions as a way to control normal program flow. If possible, include logic in the app code to detect and handle conditions that would cause an exception. For example, log problems rather than throwing an exception.
```
mjrousos: Good idea. I think this would be a good point to add.

Rick-Anderson:
@mjrousos 	 do you want to add anything from `@natemcmaster` issue https://github.com/aspnet/EntityFramework.Docs/issues/212

aspnetcore/performance/performance-best-practices.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Optimize Data Access
+
+<!-- TODO review by EF folks -->
+
+Interactions with a data store or other remote services are often the slowest parts of an ASP.NET Core app. It's important to make sure that data is read and written efficiently. In addition to ensuring data access is asynchronous, some best practices include:
+
+* **Do** consider caching frequently accessed data retrieved from a database or remote service if it is acceptable for the data to be slightly out-of-date. Depending on the scenario, you might use a [MemoryCache](xref:performance/caching/memory) or a [DistributedCache](xref:performance/caching/distributed). For more information, see [Cache responses in ASP.NET Core](xref:performance/caching/index).
+* Prefer *chunky* over *chatty* database interactions. A *chatty* interaction requires many calls to the data store. A *chunky* interaction requires fewer network calls. The goal is to retrieve all the data that will be needed in a single call rather than  several calls.
+* **Do** use [no-tracking queries](/ef/core/querying/tracking) in Entity Framework Core when accessing read-only data.
```
divega: I think about this more "when accessing data for read-only purposes". The data itself isn't read-only by nature.

aspnetcore/performance/performance-best-practices.md
```diff
@@ -0,0 +1,129 @@
...
+* **Do** filter and aggregate LINQ queries (with `.Where`, `.Select`, or `.Sum` statements, for example) so that the filtering is done by the database.
+* **Do not** retrieve more data than is necessary. Write queries to return just the data that is necessary for the current HTTP request.
```
divega: I would make this the number one recommendation. It may appear obvious but in my experience many applications miss it.

aspnetcore/performance/performance-best-practices.md
```diff
@@ -0,0 +1,129 @@
...
+
+* **Do** consider caching frequently accessed data retrieved from a database or remote service if it is acceptable for the data to be slightly out-of-date. Depending on the scenario, you might use a [MemoryCache](xref:performance/caching/memory) or a [DistributedCache](xref:performance/caching/distributed). For more information, see [Cache responses in ASP.NET Core](xref:performance/caching/index).
+* Prefer *chunky* over *chatty* database interactions. A *chatty* interaction requires many calls to the data store. A *chunky* interaction requires fewer network calls. The goal is to retrieve all the data that will be needed in a single call rather than  several calls.
+* **Do** use [no-tracking queries](/ef/core/querying/tracking) in Entity Framework Core when accessing read-only data.
+* **Do** filter and aggregate LINQ queries (with `.Where`, `.Select`, or `.Sum` statements, for example) so that the filtering is done by the database.
```
divega: You can choose to make it a child item here, or its own thing, but I would have something on the lines of "Do consider that EF Core can resolve some query operators on the client which may lead to inefficient query execution in some cases" and have a pointer to https://docs.microsoft.comef/core/querying/client-eval#client-evaluation-performance-issues.

aspnetcore/performance/performance-best-practices.md
```diff
@@ -0,0 +1,129 @@
...
+For more information, see [Garbage Collection and Performance](/dotnet/standard/garbage-collection/performance).
+
+## Optimize Data Access
+
+<!-- TODO review by EF folks -->
+
+Interactions with a data store or other remote services are often the slowest parts of an ASP.NET Core app. It's important to make sure that data is read and written efficiently. In addition to ensuring data access is asynchronous, some best practices include:
+
+* **Do** consider caching frequently accessed data retrieved from a database or remote service if it is acceptable for the data to be slightly out-of-date. Depending on the scenario, you might use a [MemoryCache](xref:performance/caching/memory) or a [DistributedCache](xref:performance/caching/distributed). For more information, see [Cache responses in ASP.NET Core](xref:performance/caching/index).
+* Prefer *chunky* over *chatty* database interactions. A *chatty* interaction requires many calls to the data store. A *chunky* interaction requires fewer network calls. The goal is to retrieve all the data that will be needed in a single call rather than  several calls.
```
divega: Another key way often used to talk about this is to minimize network roundtrips.

aspnetcore/performance/performance-best-practices.md
divega: Another issue related to inefficient client eval is N+1 queries caused by projecting collections in query results. A good pointer for how to avoid this is https://docs.microsoft.com/ef/core/what-is-new/ef-core-2.1#optimization-of-correlated-subqueries.

aspnetcore/performance/performance-best-practices.md
```diff
@@ -0,0 +1,129 @@
...
+* **Do** use [no-tracking queries](/ef/core/querying/tracking) in Entity Framework Core when accessing read-only data.
+* **Do** filter and aggregate LINQ queries (with `.Where`, `.Select`, or `.Sum` statements, for example) so that the filtering is done by the database.
+* **Do not** retrieve more data than is necessary. Write queries to return just the data that is necessary for the current HTTP request.
+
+Query issues can be detected by reviewing time spent accessing data with [Application Insights](/azure/application-insights/app-insights-overview) or with profiling tools. Most databases also make statistics available concerning frequently executed queries.
+
```
divega: For customers that want to extract the last bit of performance we can also point to explicitly compiled queries https://docs.microsoft.com/ef/core/what-is-new/ef-core-2.0#explicitly-compiled-queries and DbContext pooling https://docs.microsoft.com/en-us/ef/core/what-is-new/ef-core-2.0#dbcontext-pooling. But we should recommend to customers to measure the actual impact these features could have in their applications, because they have limitations and compiled queries make the application code more complicated.

divega:
> @divega please review the short EF section. It would be great if EF core had a perf doc we could point to. aspnet/EntityFramework.Docs#212

@Rick-Anderson agreed. I think it wouldn't look much different from what is in here actually.

ajcvickers:
@divega Looks okay. These are high-level recommendations that may not be super-easy to act on, but I don't disagree with them.

aspnetcore/performance/performance-best-practices.md
mjrousos: Good point. +1

aspnetcore/performance/performance-best-practices.md
mjrousos: Agreed. This would be a useful resource to link to.

aspnetcore/performance/performance-best-practices.md
```diff
@@ -0,0 +1,136 @@
...
+## Use the latest ASP.NET Core release
+
+Each new release of ASP.NET includes performance improvements. Optimizations in .NET Core and ASP.NET Core mean that newer versions will outperform older versions. For example, .NET Core 2.1 added support for compiled regular expressions and benefitted from [`Span<T>`](https://msdn.microsoft.com/magazine/mt814808.aspx). ASP.NET Core 2.2 added support for HTTP/2. If performance is a priority, consider upgrading to the most current version of ASP.NET Core.
+
+<!-- TODO review link and taking advantage of new [performance features](#TBD)
+Maybe skip this TBD link as each version will have perf improvements -->
+
+## Minimize exceptions
+
+Exceptions should be rare. Avoid using exceptions as a way to control normal program flow. If possible, include logic in the app code to detect and handle conditions that would cause an exception. For example, log problems rather than throwing an exception.
```
mjrousos: This is a good addition. It doesn't follow the do/do-not pattern of many of the other recommendations, though. The 'Use the latest ASP.NET Core release' one doesn't either but it feels different somehow. It may be worthwhile moving this recommendation about the 'use the latest ASP.NET Core' one and re-formatting it to match the pattern of the others:

## Minimize exceptions

Exceptions should be rare. Throwing and catching exceptions is slow relative to other code flow patterns. Because of this, exceptions should not be used to control normal program flow.

Recommendations:

* **Do not** use throwing or catching exceptions as a means of normal program flow, especially in hot code paths.
* **Do** include logic in the app to detect and handle conditions that would cause an exception.
* **Do** throw or catch exceptions which may be related to unusual or unexpected conditions.

App diagnostic tools (like Application Insights) can help to identify common exceptions in an application which may affect performance.

aspnetcore/performance/performance-best-practices.md
```diff
@@ -0,0 +1,136 @@
...
+* What percentage of the processor time is spent in garbage collection.
+* How many garbage collections are generation 0, 1, and 2.
+
+For more information, see [Garbage Collection and Performance](/dotnet/standard/garbage-collection/performance).
+
+## Optimize Data Access
+
+Interactions with a data store or other remote services are often the slowest parts of an ASP.NET Core app. It's important to make sure that data is read and written efficiently. In addition to ensuring data access is asynchronous, some best practices include:
+
+* **Do not** retrieve more data than is necessary. Write queries to return just the data that is necessary for the current HTTP request.
```
mjrousos: Nit-pick - I just noticed that the other do/do-not sections begin with the line "Recommendations:" and this one does not.

PR #9654, opened 2018-11-20
Clarify use of async in strongly typed Hub

scottaddie:
Fixes https://github.com/aspnet/Docs/issues/9647

Thanks for @f2calv for the suggestion!

### DIFF
--- a/aspnetcore/signalr/hubs.md
+++ b/aspnetcore/signalr/hubs.md
@@ -121,7 +121,7 @@ This interface can be used to refactor the preceding `ChatHub` example.
 
 Using `Hub<IChatClient>` enables compile-time checking of the client methods. This prevents issues caused by using magic strings, since `Hub<T>` can only provide access to the methods defined in the interface.
 
-Using a strongly typed `Hub<T>` disables the ability to use `SendAsync`.
+Using a strongly typed `Hub<T>` disables the ability to use `SendAsync`. Any methods defined in the interface can still be defined as asynchronous.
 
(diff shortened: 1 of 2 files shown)


aspnetcore/signalr/hubs.md
```diff
@@ -121,7 +121,7 @@ This interface can be used to refactor the preceding `ChatHub` example.
 
 Using `Hub<IChatClient>` enables compile-time checking of the client methods. This prevents issues caused by using magic strings, since `Hub<T>` can only provide access to the methods defined in the interface.
 
-Using a strongly typed `Hub<T>` disables the ability to use `SendAsync`.
+Using a strongly typed `Hub<T>` disables the ability to use `SendAsync`. Any methods defined in the interface can still be defined as asynchronous.
```
anurse:
```suggestion
Using a strongly typed `Hub<T>` disables the ability to use `SendAsync`. Any methods defined on the interface can still be defined as asynchronous.
```

In fact, they *should* be `Task` returning. Though since it's an interface, you don't use `async`, so they should look like this:

```csharp
public interface IClient
{
    Task ClientMethod();
}
```

(Note: We don't strip the `Async` suffix, so unless your client method is defined with `.on('MyMethodAsync')` you should not use `MyMethodAsync` as a name)

scottaddie: Awesome info! Thank you for that.

scottaddie:
@anurse Please take another look.

PR #9684, opened 2018-11-23
Emphasize IOptionsMonitor in Options topic + updates

### LINKED ISSUE #9662: Emphasize IOptionsMonitor as the recommended choice
@DamianEdwards brought up a good point recently that our options docs don't make it clear enough that `IOptionsMonitor` is the recommended choice going forward, since its effectively everything IOptions is but more...

@guardrex can you help me add this guidance?

I think maybe we just add a section that says something like this:

## Which options interface to use?

`IOptionsMonitor` is generally recommended as the interface to use because it supports everything `IOptions` does, in addition to supporting named options, reloadable configuration, change notifications, as well as the ability to selectively invalidate instances via the `IOptionsMonitorCache`.  (Note: these should be links to appropriate doc sections)

`IOptionsSnapshot` is a good alternative for scenarios where options should be recomputed on every request.

`IOptions` is really the only choice when using existing frameworks/libraries that are already using `IOptions`

### LINKED ISSUE #9663: Why should we use IOptions<T> when we can inject the configuration object directly.
Hello, by using IOptions&lt;T&gt; we are taking a dependency on microsoft.extensions.options in every library. This page does not go into details about why we can not inject the configuration object T directly into constructor? Even if we inject IOptions&lt;T&gt; we are still grabbing the options.value in the constructor and assigning to a local variable. Can you explain explicit reasons why we should use IOptions&lt;T&gt; ? 

---
#### Document Details

* Content: [Options pattern in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-2.1)
* Content Source: [aspnetcore/fundamentals/configuration/options.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/configuration/options.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #9663 
Fixes #9662 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-1.0&branch=pr-en-us-9684)

* Emphasizes `IOptionsMonitor`; downplays `IOptions`
* Sample to 2.2
* UE pass
* xref API links


aspnetcore/fundamentals/configuration/options.md
```diff
@@ -38,29 +44,68 @@ Add a package reference to the [Microsoft.Extensions.Options.ConfigurationExtens
...
+
+* Change notifications
+* [Reloadable configuration](#reload-configuration-data-with-ioptionssnapshot)
+
+::: moniker-end
+
+::: moniker range="< aspnetcore-1.1"
+
+* Change notifications
+
```
HaoK: This should also mention Named options maybe?

HaoK: And maybe also something to the effect of granular/selective cache invalidation via the IOptionsMonitorCache

HaoK:
Actually looks like this isn't right, OptionsMonitor is a 2.0 thing, and it looks like the rename hit 1.1 and 1.0 as well

HaoK:
I don't think the docs for 1.0 and 1.1 should be updated at all, this should only target 2.0+

aspnetcore/fundamentals/configuration/options.md
HaoK: Never mind this is 1.0/1.x I see now

HaoK:
@ajcvickers as fyi, we are updating the docs to push IOptionsMonitor instead

guardrex:
Yes, I was going by the API browser as a guide to which interfaces/scenarios were made available for given versions ...

https://docs.microsoft.com/dotnet/api/microsoft.extensions.options.ioptionsmonitor-1

... at the bottom, where **Applies to** is listed on these.

You're saying that the information is incorrect there. I'll react on the update commit ...

`IOptionsMonitor` is 2.0+.

I guess that's the only one that didn't pan out.

guardrex:
> yes this is confusing

lol :smile: ... and this type of thing *could get a contractor into trouble* :skull:. lol :smile:

The content is (seemingly) pretty good. I just need to fix up the versioning, so the update shouldn't take long. I hope to have it done by the EOD.

guardrex:
@HaoK I have the feeling that we need to dump the <2.1 coverage (i.e., take the 1.1 topic into a PDF until next June and set the topic up for 2.1 or later. Working with these versioning ranges given the re-implementation work is problematic (e.g., `IOptionsMonitor` changes at 2.0). We've been taking this approach on a few other topics with similar maintenance issues. Is it ok with you if this topic goes 2.1 or later?

HaoK:
Sure I imagine the most important versions for documentation are the LTS and current versions, so 2.1 and later seems a reasonable place to be.

guardrex:
:tada: *Wonderful!* ... I :heart: 🔪 those versioning ranges out.

I've sent the PDF off to be posted. As soon as that's online, I'll take this off of WIP. You can look now tho to see if there are any other changes to make.

HaoK:
One minor thing, Post-configuration is supported by everything, its not specific to `IOptionsMonitor`, its common to all 3 interfaces, just like validation, its something that the `IOptionsFactory` does automatically which is used by all 3.

guardrex:
@HaoK kk ... the link is live ...

https://webpifeed.blob.core.windows.net/webpifeed/Partners/Options_1.1.pdf

I moved the post-config remark out of the list into its own paragraph. Anything else before I call for docs team review?

guardrex:
@HaoK I think I have that post-configure bit under control. I'll ask Scott to help me put the final touches on this now. Let me know if you have anything else to address.

HaoK:
Cool looks good to me!

guardrex:
@scottaddie Note/reminder on the PDF situation: The 1.1 PDF notice should only appear for <=1.1; but with metadata versioning of >=2.1, it ends in a :boom: :cry:. Therefore, I think the plan is to leave the metadata versioning off until the PDF can be dropped next June, then the topic can get proper versioning. *I think that's the plan.* Let me know if we're going another way. btw- I'm *only* doing this on topics where the versioning is *horri-terri-nasti-bad* 😝. I prefer to leave the 1.x versioning in place, but it's a giant PITA on some of these topics at this point.

aspnetcore/fundamentals/configuration/options.md
```diff
@@ -4,63 +4,74 @@ author: guardrex
...
+* The [Interface Segregation Principle (ISP) or Encapsulation](/dotnet/standard/modern-web-apps-azure-architecture/architectural-principles#encapsulation) &ndash; Scenarios (classes) that depend on configuration settings depend only on the configuration settings that they use.
+* [Separation of Concerns](/dotnet/standard/modern-web-apps-azure-architecture/architectural-principles#separation-of-concerns) &ndash; Settings for different parts of the app aren't dependent or coupled to one another.
 
-[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/fundamentals/configuration/options/sample) ([how to download](xref:index#how-to-download-a-sample)) This article is easier to follow with the sample app.
+Options also provide a mechanism to validate configuration data. For more information, see the [Options validation](#options-validation) section.
 
-## Prerequisites
+[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/fundamentals/configuration/options/samples) ([how to download](xref:index#how-to-download-a-sample)) This article is easier to follow with the sample app.
```
scottaddie:
```suggestion
[View or download sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/fundamentals/configuration/options/samples) ([how to download](xref:index#how-to-download-a-sample))
```

aspnetcore/fundamentals/configuration/options.md
```diff
@@ -246,7 +237,7 @@ named_options_2: option1 = named_options_2_value1_from_action, option2 = 5
 
 ## Configure all options with the ConfigureAll method
 
-Configure all options instances with the [OptionsServiceCollectionExtensions.ConfigureAll](/dotnet/api/microsoft.extensions.dependencyinjection.optionsservicecollectionextensions.configureall) method. The following code configures `Option1` for all configuration instances with a common value. Add the following code manually to the `ConfigureServices` method:
+Configure all options instances with the <xref:Microsoft.Extensions.DependencyInjection.OptionsServiceCollectionExtensions.ConfigureAll*> method. The following code configures `Option1` for all configuration instances with a common value. Add the following code manually to the `ConfigureServices` method:
```
scottaddie:
```suggestion
Configure all options instances with the <xref:Microsoft.Extensions.DependencyInjection.OptionsServiceCollectionExtensions.ConfigureAll*> method. The following code configures `Option1` for all configuration instances with a common value. Add the following code manually to the `Startup.ConfigureServices` method:
```

aspnetcore/fundamentals/configuration/options/samples/2.x/OptionsSample/SampleApp.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
```
scottaddie: The 2.2 web app project templates also add the following element to this ItemGroup:

```xml
<AspNetCoreHostingModel>InProcess</AspNetCoreHostingModel>
```

aspnetcore/fundamentals/configuration/options.md
guardrex: @scottaddie Are you trying to get me :boy: :gun: shot!!! :smile: lol

@Rick-Anderson That was your add ... https://github.com/aspnet/Docs/pull/4797#discussion_r153640204 ... u ok with taking it off?

scottaddie: I suggested it because that statement seems to be implied when a download link is included.

aspnetcore/fundamentals/configuration/options/samples/2.x/OptionsSample/SampleApp.csproj
guardrex: Yes, indeed. The template opts-in ... NOT the SDK. I'm still getting used to the new bits.

aspnetcore/fundamentals/configuration/options.md
guardrex: Don't look at me ... it wasn't my idea. lol

# This change approved by Scott Addie

For the record ....... so they offer the blindfold and final meal to the right doc author.

# 😀

PR #9720, opened 2018-11-28
Fix typo in signalR snippets/index2.ts

sdcb:
Fix typo in signalR snippets/index2.ts


scottaddie:
Nice catch, @sdcb! Thank you for taking the time to create a PR.

PR #9751, opened 2018-11-30
Create net-core-prereqs-all-2.1.md

Rick-Anderson:
We should use this for all 2.2 tutorials - as appropriate. We can flesh this out. Nice consistency.

We could even add `## Prerequisites` to the top.


Rick-Anderson:
I don't think we need this so I'll close.

PR #9798, opened 2018-12-04
Quick patch

guardrex:
Follow-up to #9793 

I didn't see the suggestion clearly on my phone at the 💪 gym 💪 ... didn't have my 👓 with me. Originally, I was just going to touch the capital "T" here, but I think it reads a little better using our normal pattern (prep phrase to open).


guardrex:
My little nits may have put the build system into a bad mood here. I'll try a single commit on a new branch for this.

PR #9839, opened 2018-12-05
C53605: Escaping mock URL

v-makoud:
Hello, @Rick-Anderson,
Description: contoso.com is a mock URL and should not be linked. The proposed fix for this is either write the URL within code box, meaning between grave accents (`) or using backslashes.
Many thanks in advance.


Rick-Anderson:
@v-makoud 	thanks. We prefer `code-fencing`, but given it was using **bold formatting** your change was fine. Thanks for this improvement.

PR #9855, opened 2018-12-05
Enhance package/namespace coverage

### LINKED ISSUE #9843: Namespaces Are Missing
Most of the examples require importing namespaces which are not listed in this documentation.

---
#### Document Details

* ID: 7184a308-a254-9350-a5cd-5bced1f369ae
* Version Independent ID: 726e3bf1-f367-d733-8933-bccc04da0e16
* Content: [Logging in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-2.2#create-logs)
* Content Source: [aspnetcore/fundamentals/logging/index.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/logging/index.md)
* Product: **aspnet-core**
* GitHub Login: @tdykstra

guardrex:
Fixes #9843

@tdykstra I think if packages are moved up + we add the API links to the interfaces that we'll cover this. Since pretty much all of the code samps are sample app snippets, it isn't a trivial matter to add `using` namespaces.

I also rejected the approach of adding the API ref (e.g., `(<xref:Microsoft.Extensions.Logging>)`) to to the members throughout ... it disturbs the flow of the doc too much imo to do that. I think this is our best bet.


guardrex:
Yes, but we probably can't add `using` statements when 95%-ish of the code examples are snippets that draw in segments of code well below where the `using` statements appear in the sample app. We could add xref namespaces all over the place throughout the topic, but that breaks the flow of the text quite a bit. I can flesh out the interfaces in this section (perhaps rename it to *Packages and namespaces*), but that's not what we typically do in *Packages* sections when they appear early in the topic.

What's the alternative? What else can we do?

tdykstra:
For specific instances where omitting the namespace is problematic, we can add code comments in the snippet that mention the needed using.  I'm more inclined to invest that effort in instances where it's hard to know what you're missing because IntelliSense can't provide it.  I don't think we have resources or it's high enough priority to do everywhere.

PR #9930, opened 2018-12-10
Update title of button in HTML code for clarity

alistairmcmillan:
Minor change but it seems a little confusing to call the button "Edit" when you use it to "Save" your changed todo item.

Slight sense of deja vu. #9602


scottaddie:
Looks like this detail was missed when the 2.2 sample app was created. Thank you for fixing it, @alistairmcmillan!

PR #9988, opened 2018-12-12
Update Program.cs

pakrym:
This should do it for https://github.com/aspnet/Docs/issues/9774

`CreateDefaultWebHostBuilder` will call `UseContentRoot(Directory.GetCurrentDirectory())` and pick up the change.

### DIFF
--- a/aspnetcore/host-and-deploy/windows-service/samples/2.x/AspNetCoreService/Program.cs
+++ b/aspnetcore/host-and-deploy/windows-service/samples/2.x/AspNetCoreService/Program.cs
@@ -15,15 +15,16 @@ namespace SampleApp
         public static void Main(string[] args)
         {
             var isService = !(Debugger.IsAttached || args.Contains("--console"));
-            var builder = CreateWebHostBuilder(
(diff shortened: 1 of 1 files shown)


guardrex:
We can't only make changes to the sample ... the topic explains how these bits work, so there are textual changes needed, too.

guardrex:
If you want, we can merge this with my performing a quick patch for the topic to immediately follow. Sound good?

pakrym:
Sounds good.

sertunc:
Hello; when i change logic like that. service broken ? can you help me ?

if (isService)
            {
                var pathToExe = Process.GetCurrentProcess().MainModule.FileName;
                var pathToContentRoot = Path.GetDirectoryName(pathToExe);
                Directory.SetCurrentDirectory(pathToContentRoot);

                host.RunAsService();
            }
            else
            {
                host.Run();
            }

Tratcher:
Comments on closed issues are not tracked, please open a new issue with the details for your scenario.

@sertunc Calling SetCurrentDirectory after CreateWebHostBuilder is too late, the application has already read the current directory.

Issue #10006, opened 2018-12-14
Wrong folder name in " (In the Models/RazorPagesMovieContext.cs file) "


Ligren:
Should be " (In the Data/RazorPagesMovieContext.cs file) "

`[Edit by @Rick-Anderson ]`

## Executive summary:

VSC and VSmac put the context in the Models folder. I did that because the scaffolder uses Models as the namespace, but puts the code in the Data folder. A fix needs to reconcile the VSC/VS mac/VS versions. See aspnet/Scaffolding#822  for the root problem

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 6719f08e-3bd7-dc1a-71df-f2ef9fbca9d8
* Version Independent ID: 7096fdb3-612e-9e00-bd0b-8ea4886a09ce
* Content: [Add a model to a Razor Pages app in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/tutorials/razor-pages/model?view=aspnetcore-2.2&tabs=visual-studio#pmc)
* Content Source: [aspnetcore/tutorials/razor-pages/model.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/tutorials/razor-pages/model.md)
* Product: **aspnet-core**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
It's in Models for VSC/VSmac. I need to move it to Data.

PR #10007, opened 2018-12-14
wrong name of folder, issue #10006

Ligren:
Should be Data, not Models.


scottaddie:
@Rick-Anderson I'm able to reproduce the problem in VS. This change looks good from that perspective. I'll let you be the final approver on this PR.

Rick-Anderson:
It's not that simple. The scaffolder puts it in the data folder but uses the Model namespace. The VSC and VS for mac put it in the models folder. See https://github.com/aspnet/Scaffolding/issues/822

PR #10014, opened 2018-12-14
Update model.md

ThiagoCM:
Fixing reference for the Context file after the Initial Migration call.


scottaddie:
@ThiagoCM Thank you for the PR; however, this problem was already addressed in https://github.com/aspnet/Docs/pull/10007. The fix will go live within the next 24 hours.

Issue #8622, opened 2018-09-19
Update for RP


Rick-Anderson:
Needs rewrite, update, UE pass.
Doc needs alot of work. RP is a minor part of that work.
16K PV, 56% CSAT

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0156ffc3-47ac-d5da-fbc2-9ff5018b071b
* Version Independent ID: 615db388-a244-41f7-21b5-a4085d9ed479
* Content: [Model validation in ASP.NET Core MVC](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-2.1#custom-validation)
* Content Source: [aspnetcore/mvc/models/validation.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/mvc/models/validation.md)
* Product: **aspnet-core**
* GitHub Login: @tdykstra
* Microsoft Alias: **riande**

daveabrock:
I can take this RP update as well.

Rick-Anderson:
Sounds great.

henrikrxn:
Sample code still uses 2.0

daveabrock:
Thanks @henrikrxn. This is on my radar in the next few weeks and will be using fresh 2.2 examples.

henrikrxn:
@daveabrock Would you like some help updating the Github example?

But I have no idea which branches to make the PRs for 2.1 and 2.2 repectively

Rick-Anderson:
> But I have no idea which branches to make the PRs for 2.1 and 2.2 respectively

@henrikrxn 	update all to 2.2 and use master.

daveabrock:
@Rick-Anderson 	Is it okay to still reference `MVC`? To note a few examples from the document:

- "This is a great feature of MVC..."
- "MVC will try to bind request data..."
- "In MVC, simple types are..."

Both Model-View-Controller and Razor Pages run on MVC, but saying MVC might confuse some customers who think it only relates to Model-View-Controller.

Should these references just be changed to something like "in ASP.NET Core..."?

Rick-Anderson:
@danroth27 	@DamianEdwards 	@rynowak @tdykstra 	@scottaddie 	
To avoid the ambiguity between ASP.NET Core MVC and MVC with controllers and view vs Razor Pages - in this article on model binding (and similar articles) where we've written MVC will, In MVC, ....
[...]

PR #10021, opened 2018-12-15
ASP.NET Core MVC: Updating validation sample to 2.2

henrikrxn:
Addresses #8622

- Target 2.2
- Switching from using deprecated logging setup


henrikrxn:
Have tested that the sample works

But still need to test that sample works in the same way as before update

henrikrxn:
Have tested before and after changes and the functionality works in the same way

PR #10031, opened 2018-12-17
CC88813: Changing path to the file

v-makoud:
Hello, @rick-anderson,
This proposed file change comes from https://github.com/aspnet/Docs.pt-br/pull/97 .
Many thanks in advance.

### DIFF
--- a/aspnetcore/tutorials/razor-pages/model.md
+++ b/aspnetcore/tutorials/razor-pages/model.md
@@ -196,7 +196,7 @@ Update-Database
 ---  
 <!-- End of VS tabs -->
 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
Thanks for the PR. The fix is more complicated than that as it needs to work on the VSC and VS Mac tabs. See  #10006
cc @scottaddie 	@tdykstra 	incase we get more of these before I can fix it.

v-maudel:
Thanks @Rick-Anderson, now we saw you could fixed it here: https://github.com/aspnet/Docs/pull/10038

PR #10041, opened 2018-12-18
C53599: Adding break line

v-makoud:
Hello, @rick-anderson,
Translator has reported possible source content issue.
Please review and merge the suggested proposed file change to avoid this error. If you make related fix in another PR, then share your PR number, so we can confirm and close this PR.
Many thanks in advance.


scottaddie:
Thank you, @v-makoud!

PR #10044, opened 2018-12-18
Fixes misspelling in order to help when searching the codebase

witheej:


scottaddie:
Nice catch, @witheej! Thank you for the fix. Your change will go live within the next 24 hours.

PR #10047, opened 2018-12-18
Replace reference to ASP.NET Core 2.1 with 2.2

JeremySkinner:
This page in the docs states that `Microsoft.Extensions.Options.DataAnnotations` is available as part of the metapackage in Core 2.1, when actually it's only available in 2.2.


scottaddie:
Thank you for fixing this, @JeremySkinner. Your change will go live within the next 24 hours.

PR #10049, opened 2018-12-18
Update Startup.cs

cr1mp:
It's necessary for the AuthorizeAttribute(roles) to work properly.


Rick-Anderson:
@cr1mp 	that's not used in this sample.

cc @joeaudette

PR #10075, opened 2018-12-20
Note about HttpClient instance in Performance

felipeleusin:
We should not promote bad patterns regarding HttpClient instances in documentation, specially in the performance docs.

This might not be the ideal way, but I think it's better than nothing.


Rick-Anderson:
@felipeleusin 	see #10088

PR #10082, opened 2018-12-21
Updating the document.`

Arkasian:
The cause: It will become outdated in March next year, but there is a simple way to go around and keep log in via Google functionality in Core apps.


guardrex:
Hello @Arkasian ... Thanks for submitting this, but we're going to tackle it in a more comprehensive way (for example, with a sample app update) on https://github.com/aspnet/Docs/issues/10079.

PR #10194, opened 2019-01-04
Controller action return types fix

### LINKED ISSUE #6776: Asynchronous Action example checks model state, even though invalid model state/BadRequest response occurs by default
see the asynch example on this page. It uses ModelState.IsValid as an example. But actually with an [ApiController] the model state is checked and BadRequest returned by default, which means this code would never be hit. We should update this to something more relevant (use the NotFound example from above?)

---
#### Document Details

* ID: fa03501b-5e71-4c18-0269-39babbf9d93c
* Version Independent ID: b51bc104-a6e2-b21f-8362-45cee22e8c22
* Content: [Controller action return types in ASP.NET Core Web API](https://docs.microsoft.com/en-us/aspnet/core/web-api/action-return-types?view=aspnetcore-2.1#asynchronous-action-1)
* Content Source: [aspnetcore/web-api/action-return-types.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/web-api/action-return-types.md)
* Service: **unspecified**
* Product: **aspnet-core**
* GitHub Login: @scottaddie

Rick-Anderson:
Fixes #6776

[Internal Review Page](https://review.docs.microsoft.com/en-us/aspnet/core/web-api/action-return-types?view=aspnetcore-2.2&branch=pr-en-us-10194#iactionresult-type)


Rick-Anderson:
@fiksen99 	`NotFound` doesn't make sense for a create task. How about these changes?

scottaddie:
@Rick-Anderson Please see my PR for suggested changes. I annotated the `Description` property with `[Required]` since not providing the property in the request would result in an exception.

PR #10223, opened 2019-01-05
We don't have `View` object

mrmowji:
`View` is a `Controller` method which generates a `ViewResult` object.


Rick-Anderson:
@mrmowji 	thanks for catching/fixing this. I'll bet the ASP.NET Core version has the same error.

mrmowji:
@Rick-Anderson My pleasure. I'll pull the same request for .Net Core version with our modifications.

PR #10257, opened 2019-01-07
Updated IISServerOptions to IISOptions

m-wilczynski:
Changes `IISServerOptions` which is not available to `IISOptions` in docs: https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/?view=aspnetcore-2.2.

### DIFF
--- a/aspnetcore/host-and-deploy/iis/index.md
+++ b/aspnetcore/host-and-deploy/iis/index.md
@@ -142,10 +142,10 @@ For more information on hosting, see [Host in ASP.NET Core](xref:fundamentals/ho
 
 **In-process hosting model**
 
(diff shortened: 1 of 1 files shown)


guardrex:
Hello @m-wilczynski ... Thanks but see :point_right: https://github.com/aspnet/Docs/pull/9463#issue-228523235 ... `IISServerOptions` is correct for in-proc. What do you mean by "not available"?

m-wilczynski:
Alright my bad, I can see it now.
I'm atm developing ASP.NET Core 2.1 application, while for docs it's toggled for current == 2.2, where In-Proc was introduced (and where IISServerOptions becomes valid).

PR #10308, opened 2019-01-10
Update Highlighted new code to add new filed

ahdbk:
Update highlighted line for new code to add a new filed.


scottaddie:
Thank you for fixing this, @ahdbk! Your change will go live within the next 24 hours.

PR #10322, opened 2019-01-10
Create cdnjqueryvalidate1190.md

ChrisSfanos:
Adding a new jQuery Validate release, so I copied the 1.17 page and created this one.  AssetID tag wasn't updated, as it seems to be a generated GUID


guardrex:
@ChrisSfanos It's generated manually, so you can make one and place it there.

PR #10325, opened 2019-01-10
Live

mmcgowa3:


Rick-Anderson:
@mmcgowa3 	we don't allow PR's off live. Go to https://github.com/aspnet/Docs/tree/master/aspnet/mvc/overview/older-versions/using-the-html5-and-jquery-ui-datepicker-popup-calendar-with-aspnet-mvc and click on the right file. YOu can edit in browser there and create a PR.

Thanks

PR #10373, opened 2019-01-14
Updated values of CommandName

cuperman007:
Updated values of CommandName to show the valid values that can be used, and to clearly indicate that these are strings.


scottaddie:
@cuperman007 Thank you for your contribution. These changes will go live within the next 24 hours.

PR #10391, opened 2019-01-15
Update routing.md

### LINKED ISSUE #8197: Add documentation for complex segments in routing
We don't really have much documentation for how complex segments are processed in routing.

Ex: `foo{x}bar`. 

Complex segments are processed by matching up literals from right to left in a non-greedy way.  The best description is probably here: https://github.com/aspnet/Routing/blob/9cea167cfac36cf034dbb780e3f783114ef94780/src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternMatcher.cs#L296

Ex: https://github.com/aspnet/Routing/issues/733

Rick-Anderson:
Fixes #8197


aspnetcore/fundamentals/routing.md
```diff
@@ -731,3 +731,7 @@ routes.MapRoute("blog_route", "blog/{*slug}",
 ```
 
 Link generation only generates a link for this route when the matching values for `controller` and `action` are provided.
+
+## Complex segments
+
+Complex segments (for example [Route("/x{token}y")]) are processed by matching up literals from right to left in a non-greedy way. For detailed instructions, see the [source code](https://github.com/aspnet/Routing/blob/9cea167cfac36cf034dbb780e3f783114ef94780/src/Microsoft.AspNetCore.Routing/Patterns/RoutePatternMatcher.cs#L289).
```
serpent5: The URL should probably be https://github.com/aspnet/AspNetCore/blob/release/2.2/src/Http/Routing/src/Patterns/RoutePatternMatcher.cs#L293 as the aspnet/Routing repo has been archived.

JamesNK:
My concern is the linked code is no longer used to match patterns with endpoint routing.

Rick-Anderson:
@JamesNK Can you provide the correct link to the code?

JamesNK:
The new code is part of building a state machine. It is so abstracted from matching that I doubt it is of any value to someone trying to understand matching complex segments - https://github.com/aspnet/AspNetCore/blob/91514c9af7e0f4c44029b51f05a01c6fe4c96e4c/src/Http/Routing/src/Matching/DfaMatcherBuilder.cs#L227-L244

Rick-Anderson:
@JamesNK that makes it clumsy to document. See my latest revision. Suggestions welcome. cc @tdykstra 	@scottaddie

aspnetcore/fundamentals/routing.md
```diff
@@ -731,3 +731,7 @@ routes.MapRoute("blog_route", "blog/{*slug}",
 ```
 
 Link generation only generates a link for this route when the matching values for `controller` and `action` are provided.
+
+## Complex segments
+
+Complex segments (for example `[Route("/x{token}y")]`) are processed by matching up literals from right to left in a non-greedy way. See the [previous source code](https://github.com/aspnet/AspNetCore/blob/release/2.2/src/Http/Routing/src/Patterns/RoutePatternMatcher.cs#L293) for how complex segments are matched. While that code is no longer used by ASP.NET Core for complex segment matching, it provides a good match to the current algorithm. The [current code](https://github.com/aspnet/AspNetCore/blob/91514c9af7e0f4c44029b51f05a01c6fe4c96e4c/src/Http/Routing/src/Matching/DfaMatcherBuilder.cs#L227-L244) is too abstracted from matching to be useful for understanding complex segment matching.
```
tdykstra: A few thoughts 
* I suspect few readers will understand "non-greedy".
* There's an extended comment in the linked source code that you could bring into the text here with minor changes to explain the process, rather than linking to the source code.
* The source code links seem to be reversed -- why is release/2.2 "previous" and a direct link to a commit "current"?
* It shouldn't be necessary to mention and point to code that isn't helpful.

PR #10407, opened 2019-01-15
Typo

julieteg:
Fixing retreive to retrieve


scottaddie:
@julieteg Thank you for taking the time to fix this. Your change will go live within the next 24 hours.

PR #10423, opened 2019-01-16
Fixes #8567 How to configure CRA to point to dev ASP.NET Core server for API calls

### LINKED ISSUE #8567: How to configure CRA to point to dev ASP.NET Core server for API calls?
I realize this is a React/JavaScript question more than an ASP.NET Core question, but it would be nice if in the "Run the CRA server independently" section there was a note about how to point your API calls to the appropriate localhost/port for development. Else, the weather forecast page will attempt to fetch from localhost:3000/... and fail.

---
#### Document Details

* ID: 261c9875-3db6-4d30-f9d6-922a9334ecb2
* Version Independent ID: b8f8f7c4-9c90-869d-9fd7-a887bc28f262
* Content: [Use the React project template with ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/client-side/spa/react?view=aspnetcore-2.1&tabs=visual-studio)
* Content Source: [aspnetcore/client-side/spa/react.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/client-side/spa/react.md)
* Product: **aspnet-core**
* GitHub Login: @SteveSandersonMS

VirtuaBoza:
Fixes #8567

### DIFF
--- a/aspnetcore/client-side/spa/react.md
+++ b/aspnetcore/client-side/spa/react.md
@@ -91,22 +91,25 @@ The project is configured to start its own instance of the CRA development serve
 
 There's a drawback to this default setup. Each time you modify your C# code and your ASP.NET Core app needs to restart, the CRA server restarts. A few seconds are required to start back up. If you're making frequent C# code edits and don't want to wait for the CRA server to restart, run the CRA server externally, independently of the ASP.NET Core process. To do so:
 
-1. In a command prompt, switch to the *ClientApp* subdirectory, and launch the CRA development server:
+1. Add a *.env* file to the *ClientApp* subdirectory with the following setting:
+
+    ```
+    BROWSER=none
+    ```
+    
+    This will prevent your web browser from opening when starting the CRA server externally.
+
+2. In a command prompt, switch to the *ClientApp* subdirectory, and launch the CRA development server:
 
     ```console
     cd ClientApp
     npm start
     ```
 
-2. Modify your ASP.NET Core app to use the external CRA server instance instead of launching one of its own. In your *Startup* class, replace the `spa.UseReactDevelopmentServer` invocation with the following:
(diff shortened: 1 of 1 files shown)


SteveSandersonMS:
Thanks for contributing this, @VirtuaBoza.

However, I don't know why this would be needed. As long as your browser is pointing to your ASP.NET Core application (and *not* to the underlying CRA server), then it should already be using the correct URLs.

Is the issue that you're trying to browse directly to the CRA server instead of using the site served by `UseProxyToSpaDevelopmentServer`?

VirtuaBoza:
Thanks for the look, Steve. It appears that the issue (which I wrote) I’m trying to fix is the result of a misunderstanding of the documentation. Following step 1 (npm start) would launch a browser pointed to :3000. If you use this without my change, fetches will fail.  I think you’re saying is that developers are not meant to use that browser tab, but instead they’re meant to close it, start the asp.net application and VS launch a browser session which will proxy requests to the :3000 server. 

Do I understand this correctly? 

Do you think that many other developers might have the same misunderstanding and start trying to use the browser session launched by npm start? If so, I could try to change this branch to make those instructions more clear.

VirtuaBoza:
@SteveSandersonMS I was just playing around with this to figure out just that. Adding an .env file with BROWSER=none will prevent CRA from launching a browser window on start. I'd like to scratch the changes I made about the CRA proxy and add this suggestion to step 1. I'll update this branch when I get a chance today.

VirtuaBoza:
@SteveSandersonMS Previous change removed and instruction to prevent CRA from launching the broswer added with e6fd4c7.

Rick-Anderson:
@SteveSandersonMS 	looks like this is ready for a quick review. @VirtuaBoza 	added the short `Add a *.env* file to the *ClientApp* subdirectory with the following setting:` instruction.

SteveSandersonMS:
Wouldn't you say that *everyone* should have that `.env` file there? And if so, perhaps this should be baked into the project template, rather than being something we document for people to do themselves.

Rick-Anderson:
https://github.com/dotnet/templating/issues/1752

PR #10447, opened 2019-01-17
get value from StringSegment object as string

hanoitown:


Rick-Anderson:
@hanoitown 	
1>MultipartRequestHelper.cs(14,52,14,57): error CS1061: 'string' does not contain a definition for 'Value' and no accessible extension method 'Value' accepting a first argument of type 'string' could be found (are you missing a using directive or an assembly reference?)

PR #10488, opened 2019-01-18
Specify what is being used for application discrimination

pakrym:
As part of https://github.com/aspnet/AspNetCore/issues/2512

### DIFF
--- a/aspnetcore/security/data-protection/configuration/overview.md
+++ b/aspnetcore/security/data-protection/configuration/overview.md
@@ -129,7 +129,7 @@ public void ConfigureServices(IServiceCollection services)
 
 ## SetApplicationName
 
-By default, the Data Protection system isolates apps from one another, even if they're sharing the same physical key repository. This prevents the apps from understanding each other's protected payloads.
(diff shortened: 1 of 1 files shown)


aspnetcore/security/data-protection/configuration/overview.md
```diff
@@ -129,7 +129,7 @@ public void ConfigureServices(IServiceCollection services)
 
 ## SetApplicationName
 
-By default, the Data Protection system isolates apps from one another, even if they're sharing the same physical key repository. This prevents the apps from understanding each other's protected payloads.
+By default, the Data Protection system isolates apps from one another based on their content root path, even if they're sharing the same physical key repository. This prevents the apps from understanding each other's protected payloads.
```
guardrex: Since the apps don't share the same content root path, make "path" plural ...

```suggestion
By default, the Data Protection system isolates apps from one another based on their content root paths, even if they're sharing the same physical key repository. This prevents the apps from understanding each other's protected payloads.
```

pakrym:
Thank you @guardrex!

PR #10507, opened 2019-01-19
Mongo Docs code fixes

### LINKED ISSUE #10311: Bug in Update method
When trying to update a book with an HTTP PUT using e.g. URL http://localhost:54978/api/books/5c377ee3abdf6607f491228c 
and JSON body 
 {
        "id": "5c377ee3abdf6607f491228c",
        "bookName": "Design Patterns UPDATED",
        "price": 111.93,
        "category": "Computers",
        "author": "Ralph Johnson"
    }

you'll get a conversion error on the Id in the book parameter since it cannot automatically be converted to an ObjectId. However, omitting the id row in the JSON body still does not work since the ID is never set in Book object in the BookService Update method before the ReplaceOne method is run in the database. You'll get a MongoWriteException because the ObjectId is 0. Just setting the Id on the Book object solves the problem.

The other way to solve it would be be able to keep the Book Id as a string in the JSON body and for the Id string to be converted to an ObjectId type somehow by the framework. Is that possible to achieve plugging in a transformer or something in the MVC request pipeline? Otherwise I guess you always would have to get rid of the ObjectId on the Book objects before sending it to the server in various update scenarios (in a more advanced scenario). Kind of a bummer...

---
#### Document Details

* ID: 5240866d-d0a1-f0dd-7b2c-203e1d69bc4b
* Version Independent ID: 83a5cb3b-23bc-4f0c-cd1c-1d1ffeb1131a
* Content: [Build web APIs with ASP.NET Core and MongoDB](https://docs.microsoft.com/en-us/aspnet/core/tutorials/first-mongo-app?view=aspnetcore-2.2&tabs=visual-studio)
* Content Source: [aspnetcore/tutorials/first-mongo-app.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/tutorials/first-mongo-app.md)
* Product: **aspnet-core**
* GitHub Login: @prkhandelwal

prkhandelwal:
Fixes #10311

### DIFF
--- a/aspnetcore/tutorials/first-mongo-app/sample/BooksApi/Models/Book.cs
+++ b/aspnetcore/tutorials/first-mongo-app/sample/BooksApi/Models/Book.cs
@@ -5,7 +5,9 @@ namespace BooksApi.Models
 {
     public class Book
     {
-        public ObjectId Id { get; set; }
+        [BsonId]
+        [BsonRepresentation(BsonType.ObjectId)]
+        public string Id { get; set; }
 
         [BsonElement("Name")]
         public string BookName { get; set; }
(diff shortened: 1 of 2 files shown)


scottaddie:
@prkhandelwal Thanks much for the PR to fix the problem!

PR #10509, opened 2019-01-19
[DO NOT MERGE] Testing xrefs

guardrex:
## **DO NOT MERGE!**

Working on a little tool to help with xref-formatted links. Testing a few link formats here.

### DIFF
--- a/aspnetcore/index.md
+++ b/aspnetcore/index.md
@@ -9,9 +9,37 @@ uid: index
 ---
 # Introduction to ASP.NET Core
 
+## Class
+
 * <xref:Microsoft.AspNetCore.Server.Kestrel.KestrelServerOptions>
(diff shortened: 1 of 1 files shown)


guardrex:
@Rick-Anderson 

Do these links render correctly? When clicked, do they go to the right place:

https://review.docs.microsoft.com/en-us/aspnet/core/index?branch=pr-en-us-10509

* Class goes to the class
* CTOR goes to the ctor
* Methods go to the method (but not the overload)
* Property goes to the property

Rick-Anderson:
BTW - when I'm doing a test build, it's safer to copy an MD file, change the UID, and add test markup. If someone accidently merges it - no problem, it's not in the TOC and we don't have to urgently remove it.

PR #10538, opened 2019-01-22
deploybutton.png was coming from http

calway:
Changed http to https to make the entire page render as "secure".


scottaddie:
@calway Thank you for fixing this! Your change will go live within the next 24 hours.

PR #10605, opened 2019-01-24
Improve documentation for HttpClient Factory

### LINKED ISSUE #9306: Improve documentation for HttpClient Factory related to Message Handlers and sharing state
See: https://github.com/aspnet/HttpClientFactory/issues/198

This is a topic that has come up a number of times and I think we should improve the docs. A few points below that need to be documented with more clarity.

------

The HttpClient Factory manages the lifetime of message handlers. You do not have control over the lifetimes of the handlers that are created by the factory. 

In 2.1 the services that your handlers depend upon must not be scoped. This will blow up when the DI scope ends and those services get disposes. If the services you depend upon in a handler are Disposable, you will likely have problems as those services could be disposed before the handler goes out of scope.

In 2.2 the HttpClient Factory creates a separate DI scope used to create handlers. You are free to depend upon services of any scope in your handlers, and you will not run into issues with services being disposed while your handler is still in use. 

-----

If you are in ASP.NET Core need to share per-request state with message handlers from your application, there are three ways to do that:

1. Pass data into the handler using `HttpRequestMessage.Properties`
2. Use `IHttpContextAccessor` to access the current request
3. Create your own AsyncLocal storage to pass your data

guardrex:
Fixes #9306 

[Internal Review Topic (link will be added when the build completes)](#)

Location for ...

> The HttpClient Factory manages the lifetime of message handlers. You do not have control over the lifetimes of the handlers that are created by the factory.

... seemed to fit better into the *HttpClient and lifetime management* section. I put the rest into the *Outgoing request middleware* section.

Also, updating links here to xref.

### DIFF
--- a/aspnetcore/fundamentals/http-requests.md
+++ b/aspnetcore/fundamentals/http-requests.md
@@ -5,14 +5,14 @@ description: Learn about using the IHttpClientFactory interface to manage logica
 monikerRange: '>= aspnetcore-2.1'
 ms.author: scaddie
 ms.custom: mvc
-ms.date: 08/07/2018
+ms.date: 01/24/2019
 uid: fundamentals/http-requests
 ---
 # Initiate HTTP requests
 
 By [Glenn Condron](https://github.com/glennc), [Ryan Nowak](https://github.com/rynowak), and [Steve Gordon](https://github.com/stevejgordon)
 
-An [IHttpClientFactory](/dotnet/api/system.net.http.ihttpclientfactory) can be registered and used to configure and create [HttpClient](/dotnet/api/system.net.http.httpclient) instances in an app. It offers the following benefits:
+An <xref:System.Net.Http.IHttpClientFactory> can be registered and used to configure and create <xref:System.Net.Http.HttpClient> instances in an app. It offers the following benefits:
 
 * Provides a central location for naming and configuring logical `HttpClient` instances. For example, a *github* client can be registered and configured to access GitHub. A default client can be registered for other purposes.
 * Codifies the concept of outgoing middleware via delegating handlers in `HttpClient` and provides extensions for Polly-based middleware to take advantage of that.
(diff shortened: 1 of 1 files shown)


aspnetcore/fundamentals/http-requests.md
```diff
@@ -151,22 +151,42 @@ public class ValuesController : ControllerBase
...
+
+In the preceding code, the `ValidateHeaderHandler` is registered with DI. The `IHttpClientFactory` creates a separate DI scope for each handler. Handlers are free to depend upon services of any scope. Services that handlers depend upon are disposed when the handler is disposed.
+
+Once registered, <xref:Microsoft.Extensions.DependencyInjection.HttpClientBuilderExtensions.AddHttpMessageHandler*> can be called, passing in the type for the handler.
+
+::: moniker-end
+
+::: moniker range="< aspnetcore-2.2"
+
+In the preceding code, the `ValidateHeaderHandler` is registered with DI. The handler **must** be registered in DI as a transient service, never scoped. If the hander were to be registered as a scoped service and the services that the handler depends upon are disposable, the handler's services could be disposed before the handler goes out of scope causing the handler to fail.
```
scottaddie:
```suggestion
In the preceding code, the `ValidateHeaderHandler` is registered with DI. The handler **must** be registered in DI as a transient service, never scoped. If the handler were to be registered as a scoped service and the services that the handler depends upon are disposable, the handler's services could be disposed before the handler goes out of scope, causing the handler to fail.
```

aspnetcore/fundamentals/http-requests.md
```diff
@@ -151,22 +151,42 @@ public class ValuesController : ControllerBase
...
+
+Once registered, <xref:Microsoft.Extensions.DependencyInjection.HttpClientBuilderExtensions.AddHttpMessageHandler*> can be called, passing in the type for the handler.
```
scottaddie:
```suggestion
Once registered, <xref:Microsoft.Extensions.DependencyInjection.HttpClientBuilderExtensions.AddHttpMessageHandler*> can be called, passing in the handler type.
```

aspnetcore/fundamentals/http-requests.md
```diff
@@ -42,11 +42,11 @@ The `IHttpClientFactory` can be registered by calling the `AddHttpClient` extens
...
+Once registered, code can accept an `IHttpClientFactory` anywhere services can be injected with [dependency injection (DI)](xref:fundamentals/dependency-injection). The `IHttpClientFactory` can be used to create a `HttpClient` instance:
 
 [!code-csharp[](http-requests/samples/2.x/HttpClientFactorySample/Pages/BasicUsage.cshtml.cs?name=snippet1&highlight=9-12,21)]
 
-Using `IHttpClientFactory` in this fashion is a great way to refactor an existing app. It has no impact on the way `HttpClient` is used. In places where `HttpClient` instances are currently created, replace those occurrences with a call to [CreateClient](/dotnet/api/system.net.http.ihttpclientfactory.createclient).
+Using `IHttpClientFactory` in this fashion is a great way to refactor an existing app. It has no impact on the way `HttpClient` is used. In places where `HttpClient` instances are currently created, replace those occurrences with a call to <xref:System.Net.Http.IHttpClientFactory.CreateClient*>.
```
Rick-Anderson: Using `IHttpClientFactory` in this fashion is a good way to refactor an existing app. It has no impact on the way `HttpClient` is used. In places where `HttpClient` instances are currently created, replace those occurrences with a call to <xref:System.Net.Http.IHttpClientFactory.CreateClient*>.

PR #10632, opened 2019-01-26
Added user code to sample

### LINKED ISSUE #7072: User not in sample
The code is not in the sample. 

public Task SendPrivateMessage(string user, string message)
{
    return Clients.User(user).SendAsync("ReceiveMessage", message);
}

Also, I couldn't get User to work. What's the 'user' in "string user" above? connection.User?.FindFirst(ClaimTypes.Email)?.Value; I got null.

--edit--
I removed user identifier customization (which in the sample was Claim.Email). The default ClaimTypes.NameIdentifier is working

---
#### Document Details

* ID: c8d77892-d376-2f23-2634-f5d34d962a1e
* Version Independent ID: 92ed6330-b49f-1620-6beb-8556a78ac7da
* Content: [Manage users and groups in SignalR](https://docs.microsoft.com/en-us/aspnet/core/signalr/groups?view=aspnetcore-2.1)
* Content Source: [aspnetcore/signalr/groups.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/signalr/groups.md)
* Product: **aspnet-core**
* GitHub Login: @rachelappel
* Microsoft Alias: **rachelap**

bradygaster:
fixes #7072


bradygaster:
@BrennanConroy the actual comment was: 

> The issue is sample code on page doesn't work (with ClaimTypes.Email). The sample code in github also doesn't include this. **This comment is just as recommendation to either update the sample that works with ClaimTypes.Email, or leave the customization and stick with default which is ClaimTypes.NameIdentifier, or remove this misleading sample.**

So what I did was, add the code to the sample and make sure it was the default. This code is now in the sample. Now, there's no call *to that code* in the sample, so I don't know if that was the goal - if so that's a much bigger set of work items with the doc. It seemed like the reader was mainly interested in getting the code out of "just in the doc" and into the sample.

bradygaster:
@BrennanConroy i've removed the `ClaimTypes.Email` code fragment from the article. I think we should add a discussion for the `ClaimTypes` and custom provider on the [auth article](https://review.docs.microsoft.com/en-us/aspnet/core/signalr/authn-and-authz). If you like this idea, approve this PR and i'll make the changes over there, too. Thanks.

aspnetcore/signalr/groups.md
```diff
@@ -26,17 +26,10 @@ Send a message to a specific user by passing the user identifier to the `User` f
...
 
-```csharp
-public Task SendPrivateMessage(string user, string message)
-{
-    return Clients.User(user).SendAsync("ReceiveMessage", message);
-}
-```
+[!code-csharp[Configure service](groups/sample/hubs/chathub.cs?range=29-32)]
 
 The user identifier can be customized by creating an `IUserIdProvider`, and registering it in `ConfigureServices`.
```
BrennanConroy: If we do that I think this line and the one below should be removed too otherwise it looks odd.

bradygaster: SGTM will do

bradygaster:
absolutely will @scottaddie - turns out there was a merge issue with this, so let me know if you need me to fix anything prior to merge.

PR #10633, opened 2019-01-26
Update navigation address in Contributing topic

guardrex:


CONTRIBUTING.md
```diff
@@ -114,30 +114,34 @@ DocFX requires:
 
 * Download and unzip *docfx.zip* from [DocFX releases](https://github.com/dotnet/docfx/releases).
 * Add DocFX to your PATH.
-* In a command-line window, navigate to the appropriate folder that contains the *docfx.json* file (*aspnet* for ASP.NET content or *aspnetcore* for ASP.NET Core content) and run the following command:
+* In a Command Prompt window, navigate to the folder that contains the *docfx.json* file (*aspnet* for ASP.NET content or *aspnetcore* for ASP.NET Core content) and run the following command:
```
scottaddie:
```suggestion
* In a command shell, navigate to the folder that contains the *docfx.json* file (*aspnet* for ASP.NET content or *aspnetcore* for ASP.NET Core content) and run the following command:
```

guardrex: I just read over that guidance when I turned this in. The style manual downplays the "shell" lingo, but I think that's because the guidelines are still mostly MS-focused. The manual states that "Command Prompt window" with capitalization is the right way to describe that window. This section of the topic only pertains to Windows, so I went with it. Do you still want to go "shell"?

https://docs.microsoft.com/style-guide/a-z-word-list-term-collections/c/command-prompt

guardrex: Sounds good to me as "command shell." Do we *really* think that anyone doesn't know what that means? Besides, it isn't like the manual prevents "shell" in the docs. They just don't pay much attention to it. It wasn't *a thing* in the olden days of MS yesteryear. 😀 lol

PR #10639, opened 2019-01-27
Edited middleware purpose

dtekchandani:
It was confusing how middleware can execute after next component, so removed next to illustrate it is for current component.


guardrex:
Hello @dtekchandani ... It's correct in its current form. The middleware can perform work both *before* and *after* the next middleware component. See: https://docs.microsoft.com/aspnet/core/fundamentals/middleware#create-a-middleware-pipeline-with-iapplicationbuilder

PR #10687, opened 2019-01-29
NetFX CORS OPTIONS guidance update

### LINKED ISSUE #5522: The requested resource does not support http method 'OPTIONS'
_From @ryancole on February 14, 2018 21:22_

Hi folks,

I realize that this question seems to come up very, very frequently. I assure you that I've gone through each one and tried many of the suggested solutions. One solution did seem to get me to a "working" state, but it does not seem like it addressed the actual underlying issue, but instead somehow worked around it. So, let me just explain my scenario and what I've tried and maybe we can figure this out and come up with a solution to whatever the true underlying issue here is.

I'm writing an AspNet WebApi application that will be served up by IIS. My application is currently using convention based routing, as opposed to attribute routing. My application needs to handle CORS pre-flight requests (OPTIONS). I've installed the required Microsoft.AspNet.WebApi.Cors package, from Nuget. Below are the versions of the AspNet packages that I'm using ...

```
<package id="Microsoft.AspNet.Cors" version="5.2.4" targetFramework="net461" />
<package id="Microsoft.AspNet.WebApi" version="5.2.4" targetFramework="net461" />
<package id="Microsoft.AspNet.WebApi.Client" version="5.2.4" targetFramework="net461" />
<package id="Microsoft.AspNet.WebApi.Core" version="5.2.4" targetFramework="net461" />
<package id="Microsoft.AspNet.WebApi.Cors" version="5.2.4" targetFramework="net461" />
<package id="Microsoft.AspNet.WebApi.WebHost" version="5.2.4" targetFramework="net461" />
```

Based on my reading, the Microsoft.AspNet.WebApi.Cors package is supposed to add support for handling CORS pre-flight requests. This means I won't have to add manual OPTIONS action methods, etc - the nuget package brings that for me. So, I've configured the package so that it enables CORS for all available controllers, based on a commonly referenced blog post. To my understanding, this should be all that is needed for my application to begin responding to OPTIONS requests. Below is my configuration for completeness sake - it shows that I also configure some other things ...

```
public static class WebApiConfig
{
    public static void Register(HttpConfiguration config)
    {
        var jsonMapping = new RequestHeaderMapping("Accept", "text/html", System.StringComparison.OrdinalIgnoreCase, true, "application/json");
        var corsAttribute = new EnableCorsAttribute("*", "*", "*");

        config.Formatters.JsonFormatter.MediaTypeMappings.Add(jsonMapping);

        config.Services.Replace(typeof(IFilterProvider), new UnityFilterProvider(UnityConfig.Container));

        config.EnableCors(corsAttribute);

        config.MapHttpAttributeRoutes();

        config.Routes.MapHttpRoute(
            name: "DefaultApi",
            routeTemplate: "{controller}/{id}",
            defaults: new { id = RouteParameter.Optional }
        );
    }
}
```

Lastly, I've seen some reference to some items in the web.config regarding the OPTIONSVerbHandler. Mine is left at whatever the default value is, which is shown below ...

```
  <system.webServer>
    <handlers>
      <remove name="ExtensionlessUrlHandler-Integrated-4.0" />
      <remove name="OPTIONSVerbHandler" />
      <remove name="TRACEVerbHandler" />
      <add name="ExtensionlessUrlHandler-Integrated-4.0" path="*." verb="*" type="System.Web.Handlers.TransferRequestHandler" preCondition="integratedMode,runtimeVersionv4.0" />
    </handlers>
  </system.webServer>
```

Finally, with all of these items in place, it appears that my application still fails to handle OPTIONS requests. Using postman, I simply send an OPTIONS request to any of my endpoints and regardless of endpoint I receive the following 405 response ...

```
Allow →POST
Cache-Control →no-cache
Content-Length →76
Content-Type →application/json; charset=utf-8
Date →Wed, 14 Feb 2018 20:50:14 GMT
Expires →-1
Pragma →no-cache
Server →Microsoft-IIS/10.0
X-AspNet-Version →4.0.30319
X-Powered-By →ASP.NET
X-SourceFiles →=?UTF-8?B?QzpcVXNlcnNccmNvbGVcRG9jdW1lbnRzXFZpc3VhbCBTdHVkaW8gMjAxN1xQcm9qZWN0c1xQb3J0YWwuUkVTVFxQb3J0YWwuUkVTVFxhY2NvdW50XGF1dGhlbnRpY2F0ZQ==?=

{"Message":"The requested resource does not support http method 'OPTIONS'."}
```

Here's some of the things I've tried to fix this ...

1. I tried removing the OPTIONSVerbHandler from web.config, which seemed to make no change.
2. I tried explicitly putting EnableCorsAttribute on individual action methods and controllers, which seemed to make no change.
3. I explicitly put HttpOptionsAttribute on an action method, which seemed to fix the issue. This seems like the incorrect approach, though, as that attribute is for saying that action method is for handling OPTIONS requests, which is not at all what I want. I thought Microsoft.AspNet.WebApi.Cors package brings its own OPTIONS handler.

So, after spending the past two days googling, trying various things, sleeping it on and trying again ... I'm unable to figure this out without a bit of help - I'm stuck.

Does anybody know what could be causing this issue? It doesn't seem like my Microsoft.AspNet.WebApi.Cors package is correctly wired up, or doing anything. The response has no CORS headers in it. It looks like AspNet doesn't think it has any OPTIONS handler in place, despite me enabling it in my configuration.

Is this a bug in this Cors package?

Any help is appreciated! Thanks!

_Copied from original issue: aspnet/AspNetWebStack#128_

guardrex:
Fixes #5522 

@dougbu 

> Put another way, my "anywhere else" was IIS, nginx, et cetera.

... referring to allowing preflight requests to reach the app.

This shouldn't be an issue (in theory) with the `aspNetCore` handler on the ASP.NET Core side AFAIK. I remark in the netfx topics here on this issue. I include content on using external tools (Fiddler, Postman) with preflight requests.

### DIFF
--- a/aspnet/web-api/overview/security/enabling-cross-origin-requests-in-web-api.md
+++ b/aspnet/web-api/overview/security/enabling-cross-origin-requests-in-web-api.md
@@ -4,7 +4,7 @@ title: "Enabling Cross-Origin Requests in ASP.NET Web API 2 | Microsoft Docs"
 author: MikeWasson
 description: "Shows how to support Cross-Origin Resource Sharing (CORS) in ASP.NET Web API."
 ms.author: riande
-ms.date: 10/10/2018
+ms.date: 01/29/2019
 ms.assetid: 9b265a5a-6a70-4a82-adce-2d7c56ae8bdd
 msc.legacyurl: /web-api/overview/security/enabling-cross-origin-requests-in-web-api
 msc.type: authoredcontent
@@ -84,7 +84,7 @@ When you click the "Try It" button, an AJAX request is submitted to the WebServi
 !['Try it' error in browser](enabling-cross-origin-requests-in-web-api/_static/image7.png)
 
 > [!NOTE]
-> If you watch the HTTP traffic in a tool like [Fiddler](http://www.telerik.com/fiddler), you'll see that the browser does send the GET request, and the request succeeds, but the AJAX call returns an error. It's important to understand that same-origin policy does not prevent the browser from *sending* the request. Instead, it prevents the application from seeing the *response*.
+> If you watch the HTTP traffic in a tool like [Fiddler](https://www.telerik.com/fiddler), you'll see that the browser does send the GET request, and the request succeeds, but the AJAX call returns an error. It's important to understand that same-origin policy does not prevent the browser from *sending* the request. Instead, it prevents the application from seeing the *response*.
 
 ![Fiddler web debugger showing web requests](enabling-cross-origin-requests-in-web-api/_static/image8.png)
 
@@ -158,6 +158,22 @@ Here is an example response, assuming that the server allows the request:
 [!code-console[Main](enabling-cross-origin-requests-in-web-api/samples/sample9.cmd?highlight=6-7)]
 
 The response includes an Access-Control-Allow-Methods header that lists the allowed methods, and optionally an Access-Control-Allow-Headers header, which lists the allowed headers. If the preflight request succeeds, the browser sends the actual request, as described earlier.
+
+Tools commonly used to test endpoints with preflight OPTIONS requests (for example, [Fiddler](https://www.telerik.com/fiddler) and [Postman](https://www.getpostman.com/)) don't send the required OPTIONS headers by default. Confirm that the `Access-Control-Request-Method` and `Access-Control-Request-Headers` headers are sent with the request and that OPTIONS headers reach the app through IIS.
+
+To configure IIS to allow an ASP.NET app to receive and handle OPTION requests, add the following configuration to the app's *web.config* file in the `<system.webServer><handlers>` section:
+
+```xml
+<system.webServer>
+  <handlers>
+    <remove name="ExtensionlessUrlHandler-Integrated-4.0" />
+    <remove name="OPTIONSVerbHandler" />
+    <add name="ExtensionlessUrlHandler-Integrated-4.0" path="*." verb="*" type="System.Web.Handlers.TransferRequestHandler" preCondition="integratedMode,runtimeVersionv4.0" />
+  </handlers>
+</system.webServer>
+```
+
+The removal of `OPTIONSVerbHandler` prevents IIS from handling OPTIONS requests. The replacement of `ExtensionlessUrlHandler-Integrated-4.0` allows OPTIONS requests to reach the app because the default module registration only allows GET, HEAD, POST, and DEBUG requests with extensionless URLs.
 
 ## Scope Rules for [EnableCors]
 


aspnet/web-api/overview/security/enabling-cross-origin-requests-in-web-api.md
```diff
@@ -159,6 +159,22 @@ Here is an example response, assuming that the server allows the request:
...
+To configure IIS to allow an ASP.NET app to receive and handle OPTION requests, add the following configuration to the app's *web.config* file in the `<system.webServer><handlers>` section:
+
+```xml
+<system.webServer>
+  <handlers>
+    <remove name="ExtensionlessUrlHandler-Integrated-4.0" />
+    <remove name="OPTIONSVerbHandler" />
+    <add name="ExtensionlessUrlHandler-Integrated-4.0" path="*." verb="*" type="System.Web.Handlers.TransferRequestHandler" preCondition="integratedMode,runtimeVersionv4.0" />
+  </handlers>
+</system.webServer>
```
dougbu: This looks right to me but I'm not the right person to confirm. @shirhatti could you have a look please?

guardrex:
> is using IIS with ASP.NET Core covered elsewhere

AFAIK, the headers will be passed by the `aspNetCore` handler on IIS. @shirhatti will confirm. It seems unlikely that the authors of [ASP.NET CORS topic](https://docs.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-2.2) would've missed that if it wasn't the case (*guessing*).

Yes on Linux ... this might be an issue ... along the lines of ...

```
proxy_set_header   X-Forwarded-For $proxy_add_x_forwarded_for;
```

... to get the `X-Forwarded-For` header passed thru on Linux.

Both of our Linux topics (and the ASP.NET CORS topic) don't address this for Linux ...

* [Linux/Apache](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/linux-apache?view=aspnetcore-2.2)
* [Linux/Nginx](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/linux-nginx?view=aspnetcore-2.2))

If @shirhatti confirms that we need to call out passing the headers thru, I'll add some details in those topics on this. I think we'll end up with a short mention in both places and cross-links both ways.

shirhatti:
> Both of our Linux topics (and the ASP.NET CORS topic) don't address this for Linux ...

Am I missing something? We do mention it https://github.com/aspnet/Docs/blame/master/aspnetcore/host-and-deploy/linux-nginx.md#L151

guardrex:
@shirhatti I was only using `X-Forwarded-For` as an example of steps that must be taken to forward information to the the app. I looked around, and sources seem to indicate that Nginx/Apache won't touch CORS headers. If that's correct, we're good.

WRT IIS, I'll add a cross-link from the IIS topic on my next commit.

PR #10688, opened 2019-01-29
Upgraded view components sample projects to 2.2 instead of 2.0 and renewed seeding

### LINKED ISSUE #10657: Upgrade view-components samples to 2.2 instead of 2.0
@Rick-Anderson 	edit: 11 KPV, 88% CSAT.

1. I suggest to upgrade the [samples of view components](https://github.com/aspnet/Docs/tree/master/aspnetcore/mvc/views/view-components/sample) to 2.2 instead of 2.0.

2. I suggest to use the following 2.1 clean way to seed data instead of using a separate SeedData.cs file which also requires additional advanced code to be added to Program.cs:

  ```csharp
protected override void OnModelCreating(ModelBuilder modelBuilder)
{
      for (int i = 0; i < 9; i++)
            {
                modelBuilder.Entity<TodoItem>().HasData(
                   new TodoItem
                   {
                       IsDone = i % 3 == 0,
                       Name = "Task " + (i + 1),
                       Priority = i % 5 + 1
                   });   
            }
 }
```

**Note**
Using the above method of seeding does not need lots of work as the DB is in-memory and the seeding process was not referenced in the documentation.

---
#### Document Details

* ID: 9d2902f4-b6fd-ba1c-6762-d7c26f5eb4b9
* Version Independent ID: b5715d72-5f10-bb93-1051-7f06dfa9cde8
* Content: [View components in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/mvc/views/view-components?view=aspnetcore-2.2)
* Content Source: [aspnetcore/mvc/views/view-components.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/mvc/views/view-components.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

### PR DESCRIPTION
**Pull request by shadinamrouti**
Fixes #10657

### CODE REVIEW

**Comment by guardrex** [comment]
The Error page is static, so that's ok.

The scripts and styles in *Shared/_Layout.cshtml* ... you could update those, but I *gravely doubt* that it will matter all that much.

**Comment by guardrex** [comment]
@shadinamrouti, there's no migration required for `OnModelCreating` to seed the dB, correct?

**Comment by shadinamrouti** [comment]
@guardrex, I suggest to merge my PR because it is an identical working mirror of the current samples with the only differences being:
1. Targeting the 2.2 libraries.
2. Using the OnModelCreating for seeding.

Regarding your comments:
**I. Additional files**: 

The original samples do not contain any of the following (it seems on purpose to concentrate on the View Component topic):
- Views: The _layout, the _ViewImports, the Error View, navigation
- Settings: appsettings.json, HTS, cookie policy
- Static files: www root, site.css, site.js and the libraries

_Notes_:
- None of the above files is mandatory to have a working sample. 
- Avoiding the above at this time is recommended to make the samples compatible with the documentation on the page.
- Anyway, if you recommend adding the above, I can manage this in a subsequent PR. What do you think?

**II. OnModelCreating() to seed in-memory DB's**

Yes @guardrex , no migration is required because the used database is in-memory and thus migration is not possible in this context, however, to invoke OnModelCreating() an additional line is added in the constructor of the ToDo Controller which is:

```csharp
public ToDoController(ToDoContext context)
{
            _ToDoContext = context;

            //This line is used to invoke OnModelCreating() for in-memory databases 
            _ToDoContext.Database.EnsureCreated();
}
```

For more info about EnsureCreated() see this related issue:
https://github.com/aspnet/EntityFrameworkCore/issues/11666

guardrex:
> Targeting the 2.2 libraries.

We don't retain the prior patch release sample; so yes, we'd drop 2.1 for 2.2. :+1:

> if you recommend adding the above

I would just update the scripts to match the 2.2 template to take advantage of any security/bug fixes. I gravely doubt that it's a big deal in a sample like this. I wouldn't do anything else for this one.

> no migration is required

### :tada: :beers: :heart: :+1:

There's a seeding techninque that does require a migration, and that's not the most fun imo. I want to pull down a sample, fire it up, and have it ✨ *Just Work!*:tm: ✨ ... no ceremony ... no distractions.

Although it's no biggie to do a migration, I just like the idea of *not doing one* ... *LazyRex* 😀

### Approval

Given the seeding approach here, I defer to @Rick-Anderson and the team to sign off on this.

Rick-Anderson:
@shadinamrouti 	nice work.

PR #10697, opened 2019-01-30
Fix casing of configuration property

patriksvensson:
Encountered a typo while browsing the migration guide from ASP.NET Core 2.1 to 2.2.


guardrex:
@pakrym In your original content submission, you had it this way ... not as a property. We frequently use a property in our examples, but was this a case where it's assumed the dev built some `configuration` just above the line in question?

pakrym:
If Configuration as a property is more commonly used in samples this change is fine. It didn't assume anything, just that you get `IConfiguration` object from somewhere.

guardrex:
Thank @pakrym ... and thank you @patriksvensson for updating the topic! :rocket:

PR #10709, opened 2019-01-30
Update CORS doc

### LINKED ISSUE #9705: Replace ranges with snippet tags
Replace ranges with snippet tags and correct highlighting.

---
#### Document Details

* ID: 59bca5db-1196-2a8b-a0d8-4e89b2382288
* Version Independent ID: d3f332c4-4c60-039e-0e1e-bc6b11831f87
* Content: [Enable Cross-Origin Requests (CORS) in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-2.1#set-the-allowed-origins)
* Content Source: [aspnetcore/security/cors.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/security/cors.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

### LINKED ISSUE #10623: CORS instructions don't work
I think the author didn't test their code. Per the author, we need to do at least 2 things:
1. Register CORS services:
```
{
    services.AddCors();
}

```
2. Enable CORS, for example with middle ware:
```
{
    app.UseCors(builder =>
       builder.WithOrigins("http://example.com"));
}

```
However, the above instruction and code never works for me.

This works for me:
1. AddCors
```
        {
            services.AddCors(options =>
            {
                options.AddPolicy("ScormStorageCors",
                    builder => builder.AllowAnyOrigin()
                    .AllowAnyHeader()
                    .AllowAnyMethod()
                    .AllowCredentials());
            });

            services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Version_2_1);
        }
```

2. Use Cors in Configure
```

{
    app.UseCors("ScormStorageCors"); // This will apply cors to all controllers or methods
    // if you want apply cors to some controllers or methods, there are two ways:
    // 1. remove this app.UseCors and declare cors as step #3 below
    // 2. keep app.UseCors, but apply [DisableCors] to controllers or methods

}
```

3. Add policy to controller or method
```

    [Route("api/[controller]")]
    [ApiController]
    [EnableCors("ScormStorageCors")]
    public class ValuesController : ControllerBase
    {
     //..... code here
    }
```

Also, if you are using Postman tool to test, make sure you add "Origin" header to the request, the value can be the URL of the the current API (e.g. http://localhost:15349) or "*". Without this, the response will not return "Access-Control-Allow-Origin" header.

Hope this helps.

@Rick-Anderson 	edit: 24.4 K PV, 69% CSAT

---
#### Document Details

* ID: 59bca5db-1196-2a8b-a0d8-4e89b2382288
* Version Independent ID: d3f332c4-4c60-039e-0e1e-bc6b11831f87
* Content: [Enable Cross-Origin Requests (CORS) in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-2.1)
* Content Source: [aspnetcore/security/cors.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/security/cors.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

### LINKED ISSUE #10839: Please update middleware order to specify where UseCors should be placed.  Would also help to include all other middleware as well.
UseCors is missing from the order information link below. 

Please add UseCors order information.

Reference
------------- 
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/index?view=aspnetcore-2.2#order

The following Startup.Configure method adds middleware components for common app scenarios:

Exception/error handling
HTTP Strict Transport Security Protocol
HTTPS redirection
Static file server
Cookie policy enforcement
Authentication
Session
MVC

---
#### Document Details

* ID: 59bca5db-1196-2a8b-a0d8-4e89b2382288
* Version Independent ID: d3f332c4-4c60-039e-0e1e-bc6b11831f87
* Content: [Enable Cross-Origin Requests (CORS) in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-2.2)
* Content Source: [aspnetcore/security/cors.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/security/cors.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

Rick-Anderson:
Fixes #10623, Fixes #9705, Fixes #10839

[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-2.2&branch=pr-en-us-10709)


Rick-Anderson:
@blowdart 	please assign someone to review this doc - perhaps @javiercn

danroth27:
@pranavkm Could you please take a look at this?

javiercn:
Here is my feedback as a whole

// Shows UseCors with CorsPolicyBuilder.
    app.UseCors(builder =>
    {
    builder.WithOrigins("http://example.com",
                        "http://www.contoso.com");
    });

The above sample is an advanced case. It's more common to configure CORS through MVC attributes or through 
services.AddCors and then have the call be app.UseCors() (There's a default CORS policy if I remember well
and you can configure named policies)

Razor Page action method -> Unless this has changed, this statement is false. Have you tried it out?

model/action method, and CORS is enabled in middleware, the attribute takes precedence. -> Are you sure about this?
I would have swore that it applies both. We usually recommend people to use one or the other AFAIK.

$.ajax({
  type: 'get',
  url: 'https://www.example.com/home',
  xhrFields: {
    withCredentials: true
}

This is very 2000s. I would simply use the fetch API which is the modern way to do XHR requests from the browser.

#########################

How CORS works
This section describes what happens in a CORS request at the level of the HTTP messages. It's important to understand how CORS works so that the CORS policy can be configured correctly and debugged when unexpected behaviors occur.

The CORS specification introduces several new HTTP headers that enable cross-origin requests. If a browser supports CORS, it sets these headers automatically for cross-origin requests. Custom JavaScript code isn't required to enable CORS.

The following is an example of a cross-origin request. The Origin header provides the domain of the site that's making the request:


Copy
GET https://myservice.azurewebsites.net/api/test HTTP/1.1
Referer: https://myclient.azurewebsites.net/
Accept: */*
Accept-Language: en-US
Origin: https://myclient.azurewebsites.net
Accept-Encoding: gzip, deflate
User-Agent: Mozilla/5.0 (compatible; MSIE 10.0; Windows NT 6.2; WOW64; Trident/6.0)
Host: myservice.azurewebsites.net
If the server allows the request, it sets the Access-Control-Allow-Origin header in the response. The value of this header either matches the Origin header from the request or is the wildcard value "*", meaning that any origin is allowed:


Copy
HTTP/1.1 200 OK
Cache-Control: no-cache
Pragma: no-cache
Content-Type: text/plain; charset=utf-8
Access-Control-Allow-Origin: https://myclient.azurewebsites.net
Date: Wed, 20 May 2015 06:27:30 GMT
Content-Length: 12

Test message
If the response doesn't include the Access-Control-Allow-Origin header, the cross-origin request fails. Specifically, the browser disallows the request. Even if the server returns a successful response, the browser doesn't make the response available to the client app.

#########################

I would reword this section to make emphasis on a couple of things
* CORS is NOT a security feature.
  * For example, an evil actor could XSS your site and do a Cross-site request to their CORS enabled site to steal information.
* Your API is not safer by allowing CORS.
  * It's up to the client to enforce CORS, the server is going to continue executing the request no matter what, so for example
    a .NET HTTP Client can keep calling your API. It's completely up to the browser to enforce it on the client and prevent the page
    from accessing the response.
* It's a way for a given server to allow browsers to do something (cross-origin xhr request) that otherwise would be forbidden.
  * Browsers couldn't tipically do cross-origin requests, that's why they resorted to techniques like jsonp wich receives the response
    as the contents of a script that get injected into the page (scripts are allowed to be loaded cross-origin).

scottaddie:
@Rick-Anderson Looks like *Index2.cshtml* in the sample app is never used. You can probably delete that too. I was going to in my PR, but I forgot.

Rick-Anderson:
That's the actual snippet with the following change:

        `fetch('https://<web app>.azurewebsites.net/api/values/1')`

I copy/pasted your code.

Rick-Anderson:
@pranavkm please review. I think I address all of @javiercn 	comments.

Rick-Anderson:
@javiercn 	@pranavkm 	can one of you finish this for the Doc-a-thon?

aspnetcore/security/cors/sample/Cors/WebAPI/Program.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        public static void Main(string[] args)
+        {
+            CreateWebHostBuilder(args).Build().Run();
+        }
+
+        public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
+            WebHost.CreateDefaultBuilder(args)
+                .UseStartup<StartupTest>();
+                //.UseStartup<StartupTestMultiPolicy>();
```
pranavkm: Remove?

Rick-Anderson:
@javiercn @pranavkm can one of you review this?

aspnetcore/security/cors/sample/Cors/ClientApp/Pages/Index.cshtml
```diff
@@ -0,0 +1,47 @@
...
+</div>
+
+<div>
+    <h3>Test results:</h3>
+    <span id='result'></span>
+</div>
+
+<div>
+    <input type="button" value="Test Widget 1"
+           onclick="requestVal('https://webapi20190128063025.azurewebsites.net/api/widget/1')" />
```
pranavkm: Should the url be cleaned up?

aspnetcore/security/cors/sample/Cors/ClientApp/Pages/Index2.cshtml
```diff
@@ -0,0 +1,26 @@
+﻿@page
```
pranavkm: Is this file checked in by accident? The file name seems suspect and you usually don't share models.

aspnetcore/security/cors/sample/Cors/WebAPI/Program.cs
Rick-Anderson: I have multiple StartUp classes and I uncomment the one I want to test.

aspnetcore/security/cors/sample/Cors/ClientApp/Pages/Index2.cshtml
Rick-Anderson: Index2 is used imported into the docs. It's uses

`onclick="requestVal('https://<web app>.azurewebsites.net/api/values')" />`

Which is not a valid URL. Index 1 has the code I test - which I cleaned up the URL as you suggest - now that I have delete the Azure resource.

`onclick="requestVal('https://webapi123.azurewebsites.net/api/widget/1')" />`

Rick-Anderson:
MOved to #11141 as I picked up some unwanted commits.

PR #10733, opened 2019-02-01
Changes to RP node of TOC

tdykstra:
Fixed a parent node that was linked to a URL.

@Rick-Anderson Also changed name of RP Intro article node from Overview to Introduction, thinking that may better reflect the extent of the content in this article (Overview implies something more superficial). Feel free to change it back if you don't think the change will have the desired effect.

### DIFF
--- a/aspnetcore/toc.yml
+++ b/aspnetcore/toc.yml
@@ -90,11 +90,12 @@
   items: 
     - name: Razor Pages
       items: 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
Intro is better than overview but not idea. I don't think we can come up with anything better than that. the most accurate is "Intro for good developers".

PR #10744, opened 2019-02-01
Clarify that it is _.NET Core_ File Providers

jzabroski:
I originally thought that File Providers were an ASP.NET Core feature, as I was trying to figure out why Razor Pages depended on ASP.NET Core at all.  My use case is to understand how to run Razor Engine through a Console Application that sends e-mails, without all the heavy dependencies of ASP.NET.

### DIFF
--- a/aspnetcore/fundamentals/file-providers.md
+++ b/aspnetcore/fundamentals/file-providers.md
@@ -11,7 +11,7 @@ uid: fundamentals/file-providers
 
 By [Steve Smith](https://ardalis.com/) and [Luke Latham](https://github.com/guardrex)
 
-ASP.NET Core abstracts file system access through the use of File Providers. File Providers are used throughout the ASP.NET Core framework:
+ASP.NET Core abstracts file system access through the use of .NET Core File Providers. File Providers are used throughout the ASP.NET Core framework:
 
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks @jzabroski ... Let's see what engineering thinks.

@natemcmaster File Providers are in [Microsoft.Extensions.FileProviders](https://github.com/aspnet/Extensions/tree/master/src/FileProviders); therefore, I wouldn't think we would want to take this.

jzabroski:
@guardrex Hmm... good catch.  I thought Microsoft.Extensions namespace would've been .NET Core.  I would have expected if it was an ASP.NET extension it would be under Microsoft.AspNet.Extensions...

At any rate, it seems as though this is a bit of a strange problem we're trying to solve on our team, and a lot of people seem to have the same problem as us. - Our current workaround is to pivot to use RazorLight open source engine, but it has _serious_ bugs: https://github.com/toddams/RazorLight/issues/134

jzabroski:
Interesting can of worms my junior engineer opened! :)

I like the idea of https://docs.microsoft.com/en-us/dotnet/standard/io/ referring back to these extensions, but it's also confusing the ASP.NET team owns this.  Rather than put it under other libraries, if you do proceed forward, I think it might be helpful to break it out into a section on Testing IO. In the past I've used libraries on Nuget like https://www.nuget.org/packages/Simulated.FileSystem/0.5.1

PR #10750, opened 2019-02-01
Remove incorrect logging content

tdykstra:
Removing recently added examples showing logging code in startup rather than main.

@pakrym Are there other logging changes from 2.1 to 2.2 that we should be covering, or is this all that needs to be done?


pakrym:
We've obsolete a bunch of members in logging in 2.2 that's why I did this migration guide.

While we don't recommend putting stuff into `Configure` or use old filtering since 2.0 customers might have it from 1.x days and it would start showing obsolete warnings in 2.2.

I can rewrite examples to say Main instead of ConfigureServices.

PR #10775, opened 2019-02-02
Razor Components: 'introduction' over 'index' landing page

guardrex:
As I was working the redirects in Blazor.Docs, I noticed that the "index" landing page for the RC Overview doesn't work well in links. It results in ...

* **Overview:** https://docs.microsoft.com/aspnet/core/razor-components/
* **Components:** https://docs.microsoft.com/aspnet/core/razor-components/components

This would compose better for links as ...

* https://docs.microsoft.com/aspnet/core/razor-components/introduction

We do this in several spots, including most recently for SignalR docs (but note the kind'a strange "SignalR overview" link text in the TOC) ...

* https://docs.microsoft.com/aspnet/core/signalr/introduction
* *aspnetcore/signalr/introduction.md*

An alternative is to match the TOC heading of "Overview" [but this wouldn't match other places in the repo that have used the 'introduction' approach (e.g., SignalR, Data Protection, Authorization)].

* https://docs.microsoft.com/aspnet/core/razor-components/overview

How do you want to play it?

1. Go with the "introduction" approach (on this PR).
1. Go with the "overview" approach. This makes sense if APEX has embraced "Overview" as the link text. However, it seems like these other "introduction" topics (SignalR, DP, and authn) should embrace the **_"Overview" World_** in their file names/links if that's the case.
1. Close and keep what we have. :-1:

### DIFF
--- a/aspnetcore/razor-components/introduction.md
+++ b/aspnetcore/razor-components/introduction.md
@@ -6,7 +6,7 @@ monikerRange: '>= aspnetcore-3.0'
 ms.author: riande
 ms.custom: mvc
 ms.date: 01/29/2019
-uid: razor-components/index
+uid: razor-components/introduction
 ---
 # Introduction to Razor Components
 
--- a/aspnetcore/toc.yml
+++ b/aspnetcore/toc.yml
@@ -166,7 +166,7 @@
         - name: Unit test
           uid: mvc/controllers/testing
     - name: Razor Components
-      uid: razor-components/index
+      uid: razor-components/introduction
     - name: Session and app state
       uid: fundamentals/app-state
     - name: Tag Helpers
@@ -438,7 +438,7 @@
     - name: Razor Components
       items: 
         - name: Overview
-          uid: razor-components/index
+          uid: razor-components/introduction
         - name: FAQ
           uid: razor-components/faq
         - name: Get started


tdykstra:
Some considerations on the TOC/file names for the intro/overview:
* "Overview" is the standard MVC (content model) node name
* We used that in most places, but where the referenced doc gave more depth than you would expect from an "overview" we named it "introduction"
* In some cases the use of "introduction" was probably an oversight.
* File name doesn't have to match TOC here
* We do need either an index.md or a redirect so the /aspnet/core/razor-components URL goes to the right place

So what you have is fine except for this last point -- if you keep what you have I think you need a redirect.

guardrex:
> "Overview" is the standard MVC (content model) node name

Sounds good ... but I noticed that SignalR doesn't follow the convention. I did retain "Overview" here in this case because I guessed it was a convention for the node name.

> We used that in most places, but where the referenced doc gave more depth than you would expect from an "overview" we named it "introduction"

I see. :+1:

> File name doesn't have to match TOC here

Yes ... and it doesn't here. I left "Overview" in the TOC but used "introduction" in the file name. "overview" is good, too.

> We do need either an index.md or a redirect so the /aspnet/core/razor-components URL goes to the right place

Little 😕 on this point ... you imply that it's a kind of "cover-your-fanny" redirect rule. The link wouldn't exist anywhere on the repo. The link from the TOC would be directly to the "introduction.md" or "overview.md" file, so no link to `https://docs.microsoft.com/aspnet/core/razor-components/` would exist in the doc set. However if this is to cover backing out or losing the last segment, yes, I understand. :+1:

This PR is no more than a suggestion: Do we want this to be either ....

* https://docs.microsoft.com/aspnet/core/razor-components/introduction
* https://docs.microsoft.com/aspnet/core/razor-components/overview

... over ...

* https://docs.microsoft.com/aspnet/core/razor-components/

If so, which one? If yes, I'll add a redirect. If no, then let's close this ... it's not cosmically important. I just generally like *at-a-glace-where-the-freak'in-heck-am-I*-types of links. 😀 An interesting side-conclusion tho is that when APEX decided this, they thought that it **_was NOT important_** to have an ending segment that said "overview" or "introduction." Truly on those grounds, we can just close this. This is the way they're going, and I think consistency is *very important*.

tdykstra:
So let's go ahead and stick with the standard pattern -- file is named index.md, TOC node is Overview, URL is https://docs.microsoft.com/aspnet/core/razor-components/

PR #10803, opened 2019-02-05
IAlwaysRunResultFilter/IAsyncAlwaysRunResultFilter

### LINKED ISSUE #5419: Document IAlwaysRunResultFilter and IAsyncAlwaysRunResultFilter
New filters have been added during 2.1-preview1.
The following documentation page should be updated to include those types: https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/filters

guardrex:
Fixes #5419

I'm going by the API comments on the exceptions for these filters running (e.g., authz filter, exception filter). The issue attached to the engineering PR made it sound like there would be **_no exceptions_** for these running, but the API comments seem to indicate otherwise.

@filipw Is it ok if I **_steal_** your example for this? 🚓

### DIFF
--- a/aspnetcore/mvc/controllers/filters.md
+++ b/aspnetcore/mvc/controllers/filters.md
@@ -4,7 +4,7 @@ author: ardalis
 description: Learn how filters work and how to use them in ASP.NET Core MVC.
 ms.author: riande
 ms.custom: mvc
-ms.date: 02/05/2019
+ms.date: 02/08/2019
 uid: mvc/controllers/filters
 ---
 # Filters in ASP.NET Core
@@ -381,18 +381,20 @@ Because these filters implement <xref:Microsoft.AspNetCore.Mvc.Filters.IResultFi
 * An authorization filter that prevents execution of the action.
 * An exception filter that handles an exception.
 
-For example, the following filter always runs and returns a <xref:Microsoft.AspNetCore.Mvc.NotFoundResult> if a `null` <xref:Microsoft.AspNetCore.Mvc.ObjectResult> is returned from anywhere in the request processing pipeline:
+For example, the following filter always runs and sets an action result (<xref:Microsoft.AspNetCore.Mvc.ObjectResult>) with a *422 Unprocessable Entity* status code when content negotiation fails:
 
 ```csharp
-public class NotFoundAlwaysRunFilterAttribute : Attribute, 
-    IAlwaysRunResultFilter
+public class UnprocessableResultFilter : Attribute, IAlwaysRunResultFilter
 {
     public void OnResultExecuting(ResultExecutingContext context)
     {
-        if (context.Result is ObjectResult objResult && 
-            objResult.Value == null)
+        if (context.Result is StatusCodeResult statusCodeResult &&
+            statusCodeResult.StatusCode == 415)
         {
-            context.Result = new NotFoundResult();
+            context.Result = new ObjectResult("Can't process this!")
+            {
+                StatusCode = 422,
+            };
         }
     }
 
--- a/aspnetcore/mvc/controllers/filters.md
+++ b/aspnetcore/mvc/controllers/filters.md
@@ -374,12 +374,7 @@ The framework provides an abstract `ResultFilterAttribute` that you can subclass
 
 ### IAlwaysRunResultFilter and IAsyncAlwaysRunResultFilter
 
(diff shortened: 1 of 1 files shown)


guardrex:
@pranavkm Let's consider changing the example because I haven't heard back from Filip. Do you have something akin to what's there now that would be good for this?

pranavkm:
The filter here might be an useful example: https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/test/WebSites/BasicWebSite/Filters/UnprocessableResultFilter.cs. You can use it to set an action result when conneg fails

filipw:
no, please use the existing one, I missed this thread 😅

aspnetcore/mvc/controllers/filters.md
```diff
@@ -381,18 +381,20 @@ Because these filters implement <xref:Microsoft.AspNetCore.Mvc.Filters.IResultFi
...
 {
     public void OnResultExecuting(ResultExecutingContext context)
     {
-        if (context.Result is ObjectResult objResult && 
-            objResult.Value == null)
+        if (context.Result is StatusCodeResult statusCodeResult &&
+            statusCodeResult.StatusCode == 415)
         {
-            context.Result = new NotFoundResult();
+            context.Result = new ObjectResult("Can't process this!")
```
pranavkm: Maybe a different message: 😄

guardrex:
*Tooooo late* :smile: lol Thanks anyway tho. btw- Nice post on your blog :+1:

@pranavkm Try it now :eyes:

aspnetcore/mvc/controllers/filters.md
guardrex: Not bad ...... if it were up to me, it would be even whacker ...

```csharp
new ObjectResult("Are you nuts?! I'm not processing this crap!")
```

guardrex: Seriously tho ... we need to keep that short b/c of the width limit on our code blocks. This is nice and **_narrow_** for the doc.

pranavkm: Sure

aspnetcore/mvc/controllers/filters.md
```diff
@@ -367,6 +372,38 @@ For an `IAsyncResultFilter` a call to `await next` on the `ResultExecutionDelega
 
 The framework provides an abstract `ResultFilterAttribute` that you can subclass. The [AddHeaderAttribute](#add-header-attribute) class shown earlier is an example of a result filter attribute.
 
+### IAlwaysRunResultFilter and IAsyncAlwaysRunResultFilter
+
+The <xref:Microsoft.AspNetCore.Mvc.Filters.IAlwaysRunResultFilter> and <xref:Microsoft.AspNetCore.Mvc.Filters.IAsyncAlwaysRunResultFilter> interfaces declare an <xref:Microsoft.AspNetCore.Mvc.Filters.IResultFilter> implementation that runs for all action results, even if the response is short-circuited in the request processing pipeline.
```
Rick-Anderson: Here you say always run, even when short circuited, but then below you say except. Can you add the exception cases here? No pun intended.

guardrex:
@Rick-Anderson See if that works.

Rick-Anderson:
Maybe add something to point out It's not short-circuited by any other filter?

guardrex:
It says ...

... applied to an action result unless an `IExceptionFilter` or `IAuthorizationFilter` applies and short-circuits the response.

... seems ok given that it says "unless."

guardrex:
It's a little hard to word that concept ... **it works ... unless** ... :smile:. I suggest this *unless* version for now. We can work it again if we get an issue from a reader saying that they can't figure out what it means.

Rick-Anderson:
@guardrex 	yes, it's really hard. I almost like your previous one better.

guardrex:
I think u were on to something there. It did say ... *this is true &ndash; unless this is the case* ... **across two paragraphs**. Now, it says the same thing but in the same paragraph. I think it's better to say it in the same paragraph.

I think a reader will let us know if that's hard to figure out or if their app's behavior doesn't match expectations.

I have a mild concern that those aren't the *only* cases where short-circuiting behavior will prevent these `IResultFilter`s from working. I wonder how one creates a filter that can short-circuit **and break the `IResultFilter`**. `IExceptionFilter` and `IAuthorizationFilter` are doing something to break this always-activate `IAlwaysRunResultFilter`/`IAsyncAlwaysRunResultFilter` behavior ... we don't say what does the trick on that point.

If you think we need to make another pass, I'll open another issue for perhaps Sprint 150 (**later** 😅) to take another look at this.

PR #10837, opened 2019-02-07
Windows Service topic enhancements

### LINKED ISSUE #10836: Windows Service topic enhancements
Thanks @pherbel for remarking on how `sc create` **_does NOT_** set the local security policy for the user to log on as a service. I'll update for that and a few other items here.

* Call out executing `net user` commands from an admin command shell.
* Call out default password expiry of six weeks.
* Possibly call out (or just link) `wmic UserAccount where Name='username' set PasswordExpires=False` to set the password expiry to `never`.
* Use Local Security Policy console to grant the user the `Log on as a service` privilege. Turns out that `sc create` doesn't do it, and I'm not sure how my test account gained the privilege earlier.
* Cross-links:
  * Net user: https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-R2-and-2012/cc771865(v=ws.11)
  * Log on as a service: https://docs.microsoft.com/windows/security/threat-protection/security-policy-settings/log-on-as-a-service
  * User Rights Assignment: https://docs.microsoft.com/windows/security/threat-protection/security-policy-settings/user-rights-assignment
  * Configure security policy settings: https://docs.microsoft.com/windows/security/threat-protection/security-policy-settings/how-to-configure-security-policy-settings

---
#### Document Details

* Content: [Host ASP.NET Core in a Windows Service](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/windows-service?view=aspnetcore-2.2)
* Content Source: [aspnetcore/host-and-deploy/windows-service.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/host-and-deploy/windows-service.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex

guardrex:
Fixes #10836

* Call out use of admin command shell when running commands.
* Explicitly grant `Log on as a service` privilege to the user account.
* Call out MSA as an alternative approach with AD. Confirm that I chose a good link for that scenario.
* Call out default password expiry of six weeks.

### DIFF
--- a/aspnetcore/host-and-deploy/windows-service.md
+++ b/aspnetcore/host-and-deploy/windows-service.md
@@ -5,7 +5,7 @@ description: Learn how to host an ASP.NET Core app in a Windows Service.
 monikerRange: '>= aspnetcore-2.1'
 ms.author: tdykstra
 ms.custom: mvc
-ms.date: 01/22/2019
+ms.date: 02/07/2019
 uid: host-and-deploy/windows-service
 ---
 # Host ASP.NET Core in a Windows Service
@@ -141,11 +141,13 @@ dotnet publish --configuration Release --runtime win7-x64 --output c:\svc
 
 ### Create a user account
 
-Create a user account for the service using the `net user` command:
+Create a user account for the service using the `net user` command from an administrative command shell:
 
 ```console
 net user {USER ACCOUNT} {PASSWORD} /add
 ```
+
+The default password expiration is six weeks.
 
 For the sample app, create a user account with the name `ServiceUser` and a password. In the following command, replace `{PASSWORD}` with a [strong password](/windows/security/threat-protection/security-policy-settings/password-must-meet-complexity-requirements).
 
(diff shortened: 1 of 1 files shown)


guardrex:
> I'd call using an MSA as the recommended way

It never came up before for this topic. If that's what we're going to do, then I'll open a new issue. Since this PR fixes a known 🐞 WRT `Log on as a service`, I recommend we go ahead and get this in now.

Who will make the call on this going with an MSA-based approach?

If we do, my *very brief look* seemed to indicate that Win10 needs some kind of install (MMC bits?) or something ... is that correct? If so, the push-back would be that the topic should maintain its current approach for Win10 and add a whole section for MSA. However if Win10 is good-to-go with MSA OOB, then sure, we could drop the current approach in favor of MSA. Either way tho, engineering should make the call on exactly what to cover ... *not a small green dinosaur*. :smile:

shirhatti:
Agreed. Let's get this in first.

I'm ambivalent on the MSA issue and willing to let that slide as it's not that important.

guardrex:
@blowdart

guardrex:
@blowdart Yes, that was the original ask on https://github.com/aspnet/Docs/issues/10833. I'll re-open that, but I can't do it in the near future. I have 3.0 breath'in down my neck.

### 😨 🔥 🐲

guardrex:
@Rick-Anderson

PR #10872, opened 2019-02-10
Typo

kai6424:
Fixed output folder path.


kai6424:
Since I am using bash, I should do it with `For macOS and Linux`.

At the command prompt, it ran successfully according to the `For Windows`.

PR #10879, opened 2019-02-11
cache objects as key-value pairs,cache extensions

anirudhsamudrala:


Rick-Anderson:
@anirudhsamudrala 	thanks for the PR

PR #10889, opened 2019-02-11
Clarify IHostingStartup delegate calls sequence

### LINKED ISSUE #10785: IHostingStartup Overrides User Code Instead of Other Way Around
From @aodl ...

Relates to https://github.com/aspnet/Docs/issues/6951

Does this statement from MSDN documentation still hold true, since configuration building has been moved out of the Startup class (since core 2.0)?

> IHostingStartup.Configure in the hosting startup assembly is called by the runtime before Startup.Configure in user code, which allows user code to overwrite any configuration provided by the hosting startup assembly.

From this^ I was of the impression that `IHostingStartup` code would be overridable by user code in terms of web host configuration.

The problem I'm having is that although `IHostingStartup.Congifure` gets called early on as suggested, the actual lambda's that I'm passing into things like `.AddConfigureAppConfiguration` don't get called until after the ones defined in the actual app (so the `IHostingStartup` in the external assembly actually overrides the app's configuration instead of the other way around). Is this expected to be the case?

---
#### Document Details

* ID: 613ad64e-bc62-edf7-2e86-42210de718f4
* Version Independent ID: aabbbc32-3237-5ef9-c422-892dc6884a22
* Content: [Use hosting startup assemblies in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/platform-specific-configuration?view=aspnetcore-2.2#class-library)
* Content Source: [aspnetcore/fundamentals/host/platform-specific-configuration.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/host/platform-specific-configuration.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #10785

### DIFF
--- a/aspnetcore/fundamentals/host/platform-specific-configuration.md
+++ b/aspnetcore/fundamentals/host/platform-specific-configuration.md
@@ -5,7 +5,7 @@ description: Discover how to enhance an ASP.NET Core app from an external assemb
 monikerRange: '>= aspnetcore-2.0'
 ms.author: riande
 ms.custom: "mvc, seodec18"
-ms.date: 11/22/2018
+ms.date: 02/11/2019
 uid: fundamentals/configuration/platform-specific-configuration
 ---
 # Use hosting startup assemblies in ASP.NET Core
@@ -76,7 +76,7 @@ The [sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/fundamen
 * Contains a hosting startup class, `ServiceKeyInjection`, which implements `IHostingStartup`. `ServiceKeyInjection` adds a pair of service strings to the app's configuration using the in-memory configuration provider ([AddInMemoryCollection](/dotnet/api/microsoft.extensions.configuration.memoryconfigurationbuilderextensions.addinmemorycollection)).
 * Includes a `HostingStartup` attribute that identifies the hosting startup's namespace and class.
 
-The `ServiceKeyInjection` class's [Configure](/dotnet/api/microsoft.aspnetcore.hosting.ihostingstartup.configure) method uses an [IWebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.iwebhostbuilder) to add enhancements to an app. `IHostingStartup.Configure` in the hosting startup assembly is called by the runtime before `Startup.Configure` in user code, which allows user code to overwrite any configuration provided by the hosting startup assembly.
+The `ServiceKeyInjection` class's [Configure](/dotnet/api/microsoft.aspnetcore.hosting.ihostingstartup.configure) method uses an [IWebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.iwebhostbuilder) to add enhancements to an app. `IHostingStartup.Configure` in the hosting startup assembly is called by the runtime before `Startup.Configure` in user code. However, delegates registered by `IHostingStartup` (for example, when calling `ConfigureAppConfiguration`) run after user code. To ensure that configuration code that relies on callbacks is executed before user code, call `UseConfiguration` to apply the configuration to the app.
 
 *HostingStartupLibrary/ServiceKeyInjection.cs*:
 
--- a/aspnetcore/fundamentals/host/platform-specific-configuration.md
+++ b/aspnetcore/fundamentals/host/platform-specific-configuration.md
@@ -5,7 +5,7 @@ description: Discover how to enhance an ASP.NET Core app from an external assemb
 monikerRange: '>= aspnetcore-2.0'
 ms.author: riande
 ms.custom: "mvc, seodec18"
-ms.date: 02/11/2019
+ms.date: 02/14/2019
 uid: fundamentals/configuration/platform-specific-configuration
 ---
 # Use hosting startup assemblies in ASP.NET Core
@@ -76,7 +76,7 @@ The [sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/fundamen
 * Contains a hosting startup class, `ServiceKeyInjection`, which implements `IHostingStartup`. `ServiceKeyInjection` adds a pair of service strings to the app's configuration using the in-memory configuration provider ([AddInMemoryCollection](/dotnet/api/microsoft.extensions.configuration.memoryconfigurationbuilderextensions.addinmemorycollection)).
 * Includes a `HostingStartup` attribute that identifies the hosting startup's namespace and class.
 
-The `ServiceKeyInjection` class's [Configure](/dotnet/api/microsoft.aspnetcore.hosting.ihostingstartup.configure) method uses an [IWebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.iwebhostbuilder) to add enhancements to an app. `IHostingStartup.Configure` in the hosting startup assembly is called by the runtime before `Startup.Configure` in user code. However, delegates registered by `IHostingStartup` (for example, when calling `ConfigureAppConfiguration`) run after user code. To ensure that configuration code that relies on callbacks is executed before user code, call `UseConfiguration` to apply the configuration to the app.
+The `ServiceKeyInjection` class's [Configure](/dotnet/api/microsoft.aspnetcore.hosting.ihostingstartup.configure) method uses an [IWebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.iwebhostbuilder) to add enhancements to an app.
 
 *HostingStartupLibrary/ServiceKeyInjection.cs*:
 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@pakrym 	45 second review
image

pakrym:
I don't think the last sentence if phrased correctly. I'm going to let @Tratcher comment on this since he was involved in the original issue.

guardrex:
I think the part that says ...

> To ensure that configuration code that relies on callbacks is executed before user code, call `UseConfiguration` to apply the configuration to the app.

... should say ...

> To ensure that configuration code is executed before user code, call `UseConfiguration` to apply the configuration to the app.

The "relies on callbacks" mangles it a bit.

... and should it add a clear purpose at the end? ...

> If you apply configuration with `UseConfiguration`, the user can override the configuration in their app.

Of course, shoot me a line (or a few) and I'll wordsmith the content into the doc.

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -76,7 +76,7 @@ The [sample code](https://github.com/aspnet/Docs/tree/master/aspnetcore/fundamen
...
 
-The `ServiceKeyInjection` class's [Configure](/dotnet/api/microsoft.aspnetcore.hosting.ihostingstartup.configure) method uses an [IWebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.iwebhostbuilder) to add enhancements to an app. `IHostingStartup.Configure` in the hosting startup assembly is called by the runtime before `Startup.Configure` in user code, which allows user code to overwrite any configuration provided by the hosting startup assembly.
+The `ServiceKeyInjection` class's [Configure](/dotnet/api/microsoft.aspnetcore.hosting.ihostingstartup.configure) method uses an [IWebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.iwebhostbuilder) to add enhancements to an app. `IHostingStartup.Configure` in the hosting startup assembly is called by the runtime before `Startup.Configure` in user code. However, delegates registered by `IHostingStartup` (for example, when calling `ConfigureAppConfiguration`) run after user code. To ensure that configuration code that relies on callbacks is executed before user code, call `UseConfiguration` to apply the configuration to the app.
```
Tratcher: I think the really confusing thing here is that it's comparing `IHostingStartup.Configure(IWebHostBuilder)` to `Startup.Configure(IApplicationBuilder)`. The two aren't comparable. I'm not sure why the comment "which allows user code to overwrite any configuration provided by the hosting startup assembly" was ever there, Startup.Configure can't override anything on the IWebHostBuilder. Maybe this whole paragraph should be removed.

guardrex: > not sure why the comment ... was ever there

I can research that, but I have a feeling that it will just come back as my blowing a comment based on an engineering issue/PR comment that review didn't catch.

I think "any configuration" was supposed to mean "app configuration" (configuration-provider-based configuration), and then it looks like the remark is **wrong** anyway based on the issue discussion (delegates from `IHostingStartup` go **after** those from the app).

It will probably help me to repro what the OP said was happening. I usually write a better bit when I see it with my own :eyes:.

I'll try again on Thursday morning. :coffee:

Tratcher: The config ordering is a real issue, but that has nothing to do with Startup.Configure. If you want IHostingStartup to add config before the application's ConfigureAppConfiguration then you need to call UseConfiguration which executes immediately rather than as a delayed callback.

guardrex:
@Tratcher How about something like this?

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -141,6 +141,45 @@ When building an `IHostingStartup` project, the dependencies file (*\*.deps.json
 
 Only part of the file is shown. The assembly name in the example is `StartupEnhancement`.
 
+## Configuration provided by the hosting startup
+
+Provide configuration to the app using <xref:Microsoft.AspNetCore.Hosting.WebHostBuilder.ConfigureAppConfiguration*> to load the configuration after the app's <xref:Microsoft.AspNetCore.Hosting.WebHostBuilder.ConfigureAppConfiguration*> delegates execute. The app can't override configuration values provided by the hosting startup using this approach.
```
Tratcher:
```suggestion
Provide configuration to the app using <xref:Microsoft.AspNetCore.Hosting.WebHostBuilder.ConfigureAppConfiguration*> to load the configuration after the app's <xref:Microsoft.AspNetCore.Hosting.WebHostBuilder.ConfigureAppConfiguration*> delegates execute. Hosting startup configuration will take priority over the app's configuration using this approach.
```

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -141,6 +141,45 @@ When building an `IHostingStartup` project, the dependencies file (*\*.deps.json
...
+
+Provide configuration to the app using <xref:Microsoft.AspNetCore.Hosting.HostingAbstractionsWebHostBuilderExtensions.UseConfiguration*> to load the configuration before the app's <xref:Microsoft.AspNetCore.Hosting.WebHostBuilder.ConfigureAppConfiguration*> delegates execute. The app can change the configuration values provided by the hosting startup using this approach.
```
Tratcher:
```suggestion
Provide configuration to the app using <xref:Microsoft.AspNetCore.Hosting.HostingAbstractionsWebHostBuilderExtensions.UseConfiguration*> to load the configuration before the app's <xref:Microsoft.AspNetCore.Hosting.WebHostBuilder.ConfigureAppConfiguration*> delegates execute. The app's  configuration values will take priority over those provided by the hosting startup using this approach.
```

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -141,6 +141,45 @@ When building an `IHostingStartup` project, the dependencies file (*\*.deps.json
 
 Only part of the file is shown. The assembly name in the example is `StartupEnhancement`.
 
+## Configuration provided by the hosting startup
```
Tratcher: "There are two approaches to configuration depending on if you want the app's config or the hosting startup's config to take precedence.   1. ..."

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -141,6 +141,45 @@ When building an `IHostingStartup` project, the dependencies file (*\*.deps.json
...
+    public void Configure(IWebHostBuilder builder)
+    {
+        Dictionary<string, string> dict;
+
+        builder.ConfigureAppConfiguration(config =>
+        {
+            dict = new Dictionary<string, string>
+            {
+                {"ConfigurationKey1", 
+                    "From IHostingStartup: Can't be overridden by the app."},
```
Tratcher:
```suggestion
                    "From IHostingStartup: Higher priority than the app's config."},
```

aspnetcore/fundamentals/host/platform-specific-configuration.md
```diff
@@ -141,6 +141,45 @@ When building an `IHostingStartup` project, the dependencies file (*\*.deps.json
...
+            };
+
+            config.AddInMemoryCollection(dict);
+        });
+
+        dict = new Dictionary<string, string>
+        {
+            {"ConfigurationKey2", 
+                "From IHostingStartup: The app can override this value."},
```
Tratcher:
```suggestion
                "From IHostingStartup: Lower priority than the app's config."},
```

guardrex:
I made your changes, but I disagree with this direction ("precedence"/"priority") **_IF AND ONLY IF_** there is no way for the app to override in the `ConfigureAppConfiguration` approach. What I had made it super-clear when overriding can occur. Use of "precedence" and "priority" makes it sound like it will default to using what the hosting startup sets but that there is **some way** that the app can still override the hosting startup.

Tratcher:
Config is a very layered system where each provider is layered on top of the prior one and may mask some of it's values. Overriding happens all the time. `ConfigureAppConfiguration` can be called N times and contributes to the layering. `UseConfiguration` is just getting involved at an earlier stage so the later ones layer on top. That's a little more obvious in the generic host were `UseConfiguration` is replaced by `ConfigureHostConfiguration`.  `ConfigureHostConfiguration` is also callback based and can be called multiple times with the callbacks executed in order. All `ConfigureHostConfiguration` callbacks are executed before `ConfigureAppConfiguration` callbacks.

IHostingStartup can insert itself at one of two places in the config stack but the normal config layering still applies. Consider if you have two IHostingStartup's adding config, the config layering would be the order in which the IHostingStartup's ran (assuming they used the same insertion mechanism). I think precedence/priority better reflect the layering of this non-binary system.

guardrex:
Ok, but it still seems like an app today (2.x; Web Host) can't override config set by a hosting startup setting config in `ConfigureAppConfiguration`.

This is *vastly improved* over what was there before. I think the code sample 🎸 *ROCKS!* 🎷 for this because it makes it so much clearer how to structure it. If you're good, approve and I'll merge.

WRT Generic Host, we'll get into that shortly. AFAIK, we'll be hitting it **_BIG_** for Preview 3.

PR #10936, opened 2019-02-14
Update areas.md

Rick-Anderson:


wadepickett:
I also checked on seo for fun, and the keywords look good already.

wadepickett:
...also on the SEO check: I should add, there was a huge number of folks looking for "how to" which is not used, but found the doc anyway.  However that is a  design policy issue to address across the doc set.

PR #10960, opened 2019-02-15
Fixed broken link to MvcOptions source file

MarkSFrancis:
Fixed the link to MvcOptions in Compatibility Version documentation
[Link to current docs](https://docs.microsoft.com/en-us/aspnet/core/mvc/compatibility-version?view=aspnetcore-2.2)

### DIFF
--- a/aspnetcore/mvc/compatibility-version.md
+++ b/aspnetcore/mvc/compatibility-version.md
@@ -5,7 +5,7 @@ description: Discover how the Startup class in ASP.NET Core configures services 
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 12/10/2018
+ms.date: 02/15/2019
 uid: mvc/compatibility-version
 ---
 # Compatibility version for ASP.NET Core MVC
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks for sending this in @MarkSFrancis.

This predates your update, but we shouldn't link to `master` for source. There can be a lot of churn for the **_next release_**. 

AFAIK, you can go with `release/2.2` branch now. The change is just to swap `release/2.2` for `master` in these links.

Example:

```
https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/Mvc.Core/src/MvcOptions.cs
```

... to ...

```
https://github.com/aspnet/AspNetCore/blob/release/2.2/src/Mvc/Mvc.Core/src/MvcOptions.cs
```

MarkSFrancis:
That makes perfect sense - I'll make these changes now. The only reason I didn't was because as far as I could tell - the documentation is version ambiguous for this page. Is there a way I can specify from the documentation source that this file only applies to 2.2? If not, then I'll just set the links to the branch for 2.2, but it may fall out of date as version 3 onwards are released, as the docs will still link to 2.2.

guardrex:
Unfortunately, no. However, these could be referenced (probably should he) in the API browser ... not source. If they go the source route, then they'll get updated in the future on doc updates to this topic.

MarkSFrancis:
I've changed the links for
`AllowCombiningAuthorizeFilters`
and
`InputFormatterExceptionPolicy`
to point to the docs, instead of the source code.

I've also changed the link to `MvcOptions` to use the `release/2.2` branch instead of `master`

Issue #8874, opened 2018-10-06
How to perform simple authorization in code?


ackava:
I have already looked into `IAuthorizationService`, but I can't figure out AuthorizationPolicy, I have setup simple cookie based custom authentication.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 4c3930e5-8a4f-8bdf-7167-9aed8afc4253
* Version Independent ID: cea5942d-fb01-f5bf-3b63-25873d5b79cf
* Content: [Simple authorization in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/simple?view=aspnetcore-2.1)
* Content Source: [aspnetcore/security/authorization/simple.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/security/authorization/simple.md)
* Product: **aspnet-core**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

guardrex:
Hello @ackava ... It's not currently linked into the Policies topic, but there is a sample at ...

https://github.com/aspnet/Docs/tree/master/aspnetcore/security/authorization/policies/samples/PoliciesAuthApp1

It's an older 1.0-era sample (due for update on #5495), but it should still compile and run.

See if you can get the sample app running. If you need additional guidance, we recommend a support forum, such as Stack Overflow, or a support chat, such as Slack or Gitter. There are only a handful of doc authors on this repo, so we don't have the staffing to field general support requests. We do, however, definitely handle doc issues. If you run into a problem with the Policies doc or the Simple Auth doc, please do let us know about that here.

guardrex:
Your remarks are one of the first reports (probably the only report on this repo) that pushes back on the idea that devs can just go to SO and get a good, fast answer.

I've added this issue to the sample tracking issue. Whenever the docs team gets to the sample (+topic) update, they'll see this and possibly be able to work with you to address what's missing from the topic.

Since you're not getting assistance on SO, you can try the [aspnet/Home issues](https://github.com/aspnet/AspNetCore/issues). The engineers are super-busy, so make sure that you lay out the full scenario clearly in your opening post. If it's a specific code-related problem, it's best if you can put up a repro project on GH and link that into the issue so that the engineers can see what you're trying to do.

[...]

PR #10976, opened 2019-02-15
Upgrade policy sample from 1.1 to 2.2.

### LINKED ISSUE #10932: Upgrade policy sample from 1.1 to 2.2
Upgrade policy sample from 1.1 to 2.2

Changed `GetExternalAuthenticationSchemes` to `GetExternalAuthenticationSchemesAsync`
but `Where` needs work.

```
                                               BuildingEntryRequirement requirement)
{
    if (context.User.HasClaim(c => c.Type == ClaimTypes.BadgeId &&
                                   c.Issuer == "http://microsoftsecurity"))
    {
        context.Succeed(requirement);
    }

    //TODO: Use the following if targeting a version of
    //.NET Framework older than 4.6:
    //      return Task.FromResult(0);
    return Task.CompletedTask;
}
```
 
`ClaimTypes` does not contain a definition for `TemporaryBadgeId`

---
#### Document Details

* Content: [Policy-based authorization in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/policies?view=aspnetcore-2.1)
* Content Source: [aspnetcore/security/authorization/policies.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/security/authorization/policies.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

serpent5:
Fixes #10932.

@Rick-Anderson 	edit: [Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/security/authorization/policies?branch=pr-en-us-10976)

A few things to note:

* As the whole sample project was rebuilt from a 2.2 scaffold, I made multiple commits to make it easier to compare changes.
* I reorganised `System.*` namespaces to come first in the files that were copied over to make things consistent with the scaffolded files.
* As the first comment is labeled, I used `dotnet new webapp -au Individual -uld` to create the base project.
* I copied the GUID for user-secrets and the connection-string from the old project.
* I updated the `policies.md` file to reference the `AlcoholPurchase` RP that replaced the original controller and to show the correct ranges from the new code.
* I went with magic strings to resolve the `ClaimTypes.BadgeId` and `ClaimTypes.TemporaryBadgeId`. I concluded that this was fine as the issuer and policy names are already magic strings and this is just a sample project.


guardrex:
It looks like the topic+sample was touched at some point because the topic has links to the sample now.

Anyway, you might want to look over https://github.com/aspnet/Docs/issues/8874 to see if anything can/should be done. It might be that prior changes have resolved that issue and that we just never closed it out.

serpent5:
@guardrex I didn't know there was a sample-tracker issue here (but now I do :smile:). It looks like the OP from the issue you linked wants an example of how to use `IAuthorizationService` - I can add something to the sample project with a corresponding section on the docs page if that would be helpful. Let me know if it would be helpful and if so, also let me know whether I'd be better off waiting for @blowdart to review this pull-request first or whether I should add it to what's already here.

guardrex:
Let's ask because @Rick-Anderson closed https://github.com/aspnet/Docs/issues/8874 ... Are you interested in the request at #8874, or let that one go?

Rick-Anderson:
@blowdart 	do you want @serpent5 	to add  an example of how to use `IAuthorizationService` with policies

Rick-Anderson:
@blowdart 	do you want @serpent5 	to add  an example of how to use `IAuthorizationService` with policies?

Rick-Anderson:
@HaoK 	can you review this or do you need @blowdart 	?

blowdart:
If you can give me an internal preview URL I'll try to take a look tomorrow evening.

Rick-Anderson:
https://review.docs.microsoft.com/en-us/aspnet/core/security/authorization/policies?view=aspnetcore-2.2&branch=pr-en-us-10976

blowdart:
Ah it's razor pages.

So this presents a problem to me; we're assuming right up front everyone is using razor pages. Which isn't true. For example,

using Microsoft.AspNetCore.Authorization;
using Microsoft.AspNetCore.Mvc.RazorPages;

[Authorize(Policy = "AtLeast21")]
public class AlcoholPurchaseModel : PageModel
{
}

PageModel is razor pages. People will still use MVC. I believe this needs to present controller as well, alongside this

Another slight hiccup is

_Each handler is added to the services collection by invoking services.AddSingleton<IAuthorizationHandler, YourHandlerClass>();._

Not exactly, the Add* method called needs to be appropriate to the handler. If you take a dependency on something like a repository then Singleton isn't what you want :)

serpent5:
@blowdart @Rick-Anderson In terms of the RP vs MVC concern here, I see a couple of options:

1. Replace `AlcoholPurchaseModel` with `AlcoholPurchaseController`, as it was in the 1.1 version of the example.
2. Add `AlcoholPurchaseController` as it was before, but with the understanding that it won't have any effect within the sample project itself as the RP version will "win" for routing.
3. Add a new controller, but something other than `AlcoholPurchaseController` as the sample project could be confusing with two versions of the same endpoint.

Should I go with one of those or is there a better option? I'm also not sure how best to include both examples in the documentation...

serpent5:
@blowdart @Rick-Anderson Please advise. I'm happy to recreate the project as MVC if that's preferred.

Rick-Anderson:
@serpent5 	if it's not too much work I'd create a 2nd/independent sample for MVC. I'd include them both.

serpent5:
I intend to create a new PR that first converts the sample from 1.1 to 2.2 and then adds an additional sample that uses RP.

blowdart:
Yea, both is best unfortunately.

Issue #10987, opened 2019-02-17
    app.UseStaticFiles(); before  app.UseResponseCaching(); - is it right?


shapeh:
Under Conditions section it says:
"Terminal middleware, such as Static File Middleware, must not process the response prior to the Response Caching Middleware."

In the Configuration you do this:

   ...
   app.UseHttpsRedirection();
    app.UseStaticFiles();   // static comes before caching?
    app.UseCookiePolicy();

    app.UseResponseCaching(); // caching. 

Shouldn't UseResponseCaching come before UseStaticFiles - or am I misreading the Conditions?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 10cda0f9-e2c7-015d-4fbf-d361f6c67d9f
* Version Independent ID: 13154131-c2b6-2c72-fb0e-1d83bca4e21a
* Content: [Response Caching Middleware in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/middleware?view=aspnetcore-2.2)
* Content Source: [aspnetcore/performance/caching/middleware.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/performance/caching/middleware.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
The "conditions" are just saying what has to be true in order for the Response Caching Middleware to work. It's not saying that any particular middleware should or shouldn't be allowed to cache, including the Static File Middleware. I think the best thing to do is just remove that example. I'll just make it say what is obviously true ...

> Terminal middleware must not process the response prior to the Response Caching Middleware.

... and leave it at that. Thanks for commenting on this.

shapeh:
Okay, I am not sure I get your response correctly - I am unsure if the order matter?

Is this order correct?
```
app.UseStaticFiles();
app.UseResponseCaching();
```

Or this one? - or it does not matter?
```
app.UseResponseCaching();
app.UseStaticFiles();
```

guardrex:
When a *terminal middleware* (for example, Static File Middleware) processes a request, it terminates the middleware processing pipeline and no middleware that comes after it will execute.

In order for Response Caching Middleware to work, it must process the request.

There's no "correct" order, per se. It just depends on whether or not you want responses from terminal middleware to be cachable.

### Response Caching Middleware *after* Static File Middleware

[...]

Issue #8706, opened 2018-09-26
Confirm middleware order notes appear in individual topics


guardrex:
From a discussion on Slack: Check middleware scenario topics and confirm that the order notes from ...

https://docs.microsoft.com/aspnet/core/fundamentals/middleware#built-in-middleware

... are present in the individual topics.

Rick-Anderson:
@guardrex 	should we close this or promote to P1 + P2

guardrex:
Let's close. I have a ton of higher priority work to do. This was a 'good idea if we have time' thing to check. I don't have time in the foreseeable future to address this.

PR #10990, opened 2019-02-17
"Terminal middleware" enhancements

guardrex:
**Apologies for the length of these remarks :point_down:, but there be 🐉 here.**

Addresses #10987 

@shapeh and I deliberated a bit further: Turns out that we aren't covering "terminal middleware" very well. The last PR didn't go far enough.

"Terminal middleware" only appears in the *Built-in middleware* table of the Middleware topic and assumes that the reader will define it correctly.

* I've elaborated on it in the Middleware topic ... both early where "short-circuiting" is described and later where it's used in the table.
* The Response Caching topic now has a spot that we can link directly for *terminal middleware* (and then links to the whole topic as 'for more info, see'). I link to the table because the caching topic is saying that 'a terminal middleware' must not process the request first and the table indicates which middlewares can act as a terminal middleware. Since that condition of caching now links to both the table and the topic as a whole, I think we're covered.
* The paragraph at Line 31 of the Middleware topic is a bit disorganized. I update it here so that the concepts pertaining to short-circuiting/terminal middleware are in their own paragraph. Also ...
* :dragon: I move the paragraph on terminal middleware down in the section because I'd like to make a point about earlier middleware that still processes its post-`next.Invoke` code *after* the terminal middleware runs. The only way to do that is to move this paragraph down below where `next` is described.
* :dragon: The current text implies (via the `HasStarted` check remark) that *completely terminating request processing* (i.e., including any post-`next.Invoke` code in *prior* middleware) isn't possible. Just making sure the doc is correct on that point. We could add a `HasStarted` example to the warning where a prior middleware checks before it processes any post-`next.Invoke` code. Let me know if you'd like to see that drafted in here.

Thanks @shapeh! :rocket:

### DIFF
--- a/aspnetcore/fundamentals/middleware/index.md
+++ b/aspnetcore/fundamentals/middleware/index.md
@@ -4,7 +4,7 @@ author: rick-anderson
 description: Learn about ASP.NET Core middleware and the request pipeline.
 ms.author: riande
 ms.custom: mvc
-ms.date: 10/10/2018
+ms.date: 02/17/2019
 uid: fundamentals/middleware/index
 ---
 # ASP.NET Core Middleware
@@ -18,7 +18,7 @@ Middleware is software that's assembled into an app pipeline to handle requests 
 
 Request delegates are used to build the request pipeline. The request delegates handle each HTTP request.
 
-Request delegates are configured using <xref:Microsoft.AspNetCore.Builder.RunExtensions.Run*>, <xref:Microsoft.AspNetCore.Builder.MapExtensions.Map*>, and <xref:Microsoft.AspNetCore.Builder.UseExtensions.Use*> extension methods. An individual request delegate can be specified in-line as an anonymous method (called in-line middleware), or it can be defined in a reusable class. These reusable classes and in-line anonymous methods are *middleware*, also called *middleware components*. Each middleware component in the request pipeline is responsible for invoking the next component in the pipeline or short-circuiting the pipeline.
+Request delegates are configured using <xref:Microsoft.AspNetCore.Builder.RunExtensions.Run*>, <xref:Microsoft.AspNetCore.Builder.MapExtensions.Map*>, and <xref:Microsoft.AspNetCore.Builder.UseExtensions.Use*> extension methods. An individual request delegate can be specified in-line as an anonymous method (called in-line middleware), or it can be defined in a reusable class. These reusable classes and in-line anonymous methods are *middleware*, also called *middleware components*. Each middleware component in the request pipeline is responsible for invoking the next component in the pipeline or short-circuiting the pipeline. When a middleware short-circuits, it's called a *terminal middleware* because it prevents further middleware from processing the request.
 
 <xref:migration/http-modules> explains the difference between request pipelines in ASP.NET Core and ASP.NET 4.x and provides more middleware samples.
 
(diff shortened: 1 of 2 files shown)


guardrex:
Wait to see what Chris thinks, then we can go forward with it. *Watch out for little 🐉 here.*

aspnetcore/fundamentals/middleware/index.md
```diff
@@ -40,6 +40,8 @@ Chain multiple request delegates together with <xref:Microsoft.AspNetCore.Builde
 
 [!code-csharp[](index/snapshot/Chain/Startup.cs?name=snippet1)]
 
+When a delegate doesn't pass a request to the next delegate, it's called *short-circuiting the request pipeline*. Short-circuiting is often desirable because it avoids unnecessary work. For example, [Static File Middleware](xref:fundamentals/static-files) can act as a *terminal middleware* by returning a request for a static file and short-circuiting the rest of the pipeline. Middleware added to the pipeline before the middleware that terminates further processing still processes code after their `next.Invoke` statements. However, see the following warning about attempting to write to a response that has already been sent.
```
Tratcher:
```suggestion
When a delegate doesn't pass a request to the next delegate, it's called *short-circuiting the request pipeline*. Short-circuiting is often desirable because it avoids unnecessary work. For example, [Static File Middleware](xref:fundamentals/static-files) can act as a *terminal middleware* by processing a request for a static file and short-circuiting the rest of the pipeline. Middleware added to the pipeline before the middleware that terminates further processing still processes code after their `next.Invoke` statements. However, see the following warning about attempting to write to a response that has already been sent.
```

aspnetcore/performance/caching/middleware.md
```diff
@@ -132,7 +132,7 @@ When testing and troubleshooting caching behavior, a browser may set request hea
 
 * The request must result in a server response with a 200 (OK) status code.
 * The request method must be GET or HEAD.
-* Terminal middleware must not process the response prior to the Response Caching Middleware.
+* [Terminal middleware](xref:fundamentals/middleware/index#built-in-middleware) must not process the response before Response Caching Middleware in the middleware request processing pipeline. For more information, see <xref:fundamentals/middleware/index>.
```
Tratcher: I'm not sure why this comment is here. This isn't specific to response caching, this is how ordering works with any middleware.

guardrex: This isn't a general informational list. It's in the [Response Caching Topic (here)](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/middleware?view=aspnetcore-2.2#conditions-for-caching). This is a list of conditions that must be true for caching to occur. This line is just calling it out for this middleware's ordering. Actually, this was the line that confused the reader and generated the whole PR.

Tratcher: Would they be less confused if we removed the line?

guardrex: **_I say nae_** ... if a user of this middleware places it after something else ... some middleware of their own perhaps that shoots a response, it won't work. That was the question. The reader is asking for clarification on the ordering of this middleware.

This is also part of a *larger docs concern* that I have based on the number of questions that we've had on middleware ordering. We closed [Confirm middleware order notes appear in individual topics \#8706](https://github.com/aspnet/Docs/issues/8706) due to a lack of staffing (or cloning ... since David refused to give me his cloning tech on the grounds that *one Rex is more than enough!* 😝).

This is a list of **conditions** for the middleware to work, and **_it's order is a condition of it working_**.

What's your objection to this line? It's clear now and links to the right info.

Tratcher: Let's use the same language from the linked doc. "The middleware must be placed before components that require compression." It's simpler and positive rather than negative.

guardrex: ok

guardrex: > The middleware must be placed before components that require compression.

Let's say *where* and flesh out *before*. I'm scared of using "components," so let's say "middleware scenarios" (given that we can't use the word "features" any longer).

> In `Startup.Configure`, Response Caching Middleware must be placed before middleware scenarios that require compression. For more information, see \<xref:fundamentals/middleware/index>.

Tratcher: Scenarios?
"before middleware that require compression"?

guardrex: Yeah ... we're limited given that "features" is a *banned word* now. lol

Stand-by ........................

guardrex: The subject-verb agreement *sounds* :ear: a hair funky given that "middleware" is plural here but sounds/assumed singular when read. No worries. I'm sure that readers will survive. :smile:

guardrex:
ok ...

> [Error] An exception occurred during the build process. Please check the raw log and retry building your content again. If the issue persists, please open a ticket in http://SiteHelp and include the URL of publishing report as well as below information: One or more errors occurred.

guardrex:
Ty @Tratcher ......

### *Sir Chris the 🐉 Slayer*

If you'll be around for the Summit, the :beers: are on me .... er .... MS. lol See you there.

PR #11013, opened 2019-02-19
IIS shared config update

### LINKED ISSUE #10913: Update section on using ASP.NET Core when IIS shared config is enabled
See PR here - https://github.com/aspnet/AspNetCore/pull/6498#issue-243121827

We added a flag to the hosting bundle that allows you to override the shared config block.
While it is not recommended as the default, you can use it if you can guarantee the installer won't fail because it doesn't have permissions to access appHost.config. In the default case, shared config is usually pointing a network-location and the installer will fail. If you happen to use shared config to just move appHost.config to a different location on the same machine you can successfully use this flag. This seems to be common pattern with folks using Puppet/Ansible/etc

https://github.com/aspnet/Docs/blame/master/aspnetcore/host-and-deploy/aspnet-core-module.md#L495

guardrex:
Fixes #10913

### DIFF
--- a/aspnetcore/host-and-deploy/aspnet-core-module.md
+++ b/aspnetcore/host-and-deploy/aspnet-core-module.md
@@ -494,7 +494,7 @@ A pairing token is used to guarantee that the requests received by Kestrel were 
 
 ## ASP.NET Core Module with an IIS Shared Configuration
 
-The ASP.NET Core Module installer runs with the privileges of the **SYSTEM** account. Because the local system account doesn't have modify permission for the share path used by the IIS Shared Configuration, the installer throws an access denied error when attempting to configure the module settings in the *applicationHost.config* file on the share.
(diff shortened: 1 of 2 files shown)


aspnetcore/host-and-deploy/aspnet-core-module.md
```diff
@@ -494,13 +494,36 @@ A pairing token is used to guarantee that the requests received by Kestrel were
 
 ## ASP.NET Core Module with an IIS Shared Configuration
 
-The ASP.NET Core Module installer runs with the privileges of the **SYSTEM** account. Because the local system account doesn't have modify permission for the share path used by the IIS Shared Configuration, the installer hits an access denied error when attempting to configure the module settings in *applicationHost.config* on the share. When using an IIS Shared Configuration, follow these steps:
+The ASP.NET Core Module installer runs with the privileges of the **SYSTEM** account. Because the local system account doesn't have modify permission for the share path used by the IIS Shared Configuration, the installer throws an access denied error when attempting to configure the module settings in the *applicationHost.config* file on the share.
```
shirhatti: Installer runs as `TrustedInstaller`, not `System`

PR #11016, opened 2019-02-20
Add RenderTreeBuilder for built components

neven10:
There wasn't any official example so i added this. I know you plan to add higher level API for this anyway.

Fixes https://github.com/aspnet/Docs/issues/10767


guardrex:
@neven10 Work in two warning statements in a single warning block for this PR. Change my suggested language if you work up improved phrasing. I'd place this immediately under the opening line (Line 767) so that the reader will receive it as soon as they understand what this section is about.

```
> [!WARNING]
> Dynamic components are an advanced scenario and not recommended for general development. Components generated dynamically can bypass normal safety checks on tag structure. Inadvertently creating a faulty tag that becomes rendered in the browser can result in a security vulnerability.
>
> When creating components dynamically, **always** hardcode the sequence number. Do **not** generate the number using any sort of calculation or counter.
```

@blowdart Would you like to supply advice or text on the wording of this warning?

neven10:
> @neven10 Work in two warning statements in a single warning block for this PR. Change my suggested language if you work up improved phrasing. I'd place this immediately under the opening line (Line 767) so that the reader will receive it as soon as they understand what this section is about.
> 
> This is a cut-'n-paste (at least to get u started) ...
> 
> > > [!WARNING]
> > > Dynamic components are an advanced scenario and not recommended for general development. Components generated dynamically can bypass normal safety checks on tag structure. Inadvertently creating a faulty tag that becomes rendered in the browser can result in a security vulnerability.
> > >
> > > When creating components dynamically, **always** hardcode the sequence number. Do **not** generate the number using any sort of calculation or counter.
> 


> > > [!WARNING]
> > > Dynamic components should be used only for advanced scenarios and it's not recommended for general development. As best practise you should always use Parent/Child components along with setting component parameters.
Dynamically generated components require hard coding the sequence number of components and attributes. This can can bypass normal safety checks on tag structure and may lead to creation of faulty tag which , when rendered, may result in security vulnerability.
This feature will be changed in the future.


Is this ok? Tbh i don't even know which advanced scenarios this covers. I think this is the only way to have non rendered component. "Should render" doesn't help in this scenario , because component is always initially rendered.

guardrex:
I prefer what I posted for several reasons [e.g., state verbs, sentence agreement ('require hard coding' followed by 'can bypass' in the next sentence), general phrasing, style+grammar]. Let's go with I sent unless @blowdart has an objection.

> i don't even know which advanced scenarios this covers. I think this is the only way to have non rendered component.

It's saying that 'dynamic components is an advanced scenario.' The warning addresses the entire use of this approach.

> "Should render" doesn't help in this scenario , because component is always initially rendered.

That's not part of the text that I supplied. What are u referring to?

guardrex:
@neven10 ... see if I butchered the example. It's probably best to state the naming/file names when leading into the code.

guardrex:
@danroth27 ... ready for a look.

aspnetcore/razor-components/components.md
```diff
@@ -761,3 +761,54 @@ The time is 10/04/2018 01:26:52.
 
 Your pet's name is Rex.
 ```
+
+## Dynamic components
+
+Dynamic components can be used to dynamically generate content. `Microsoft.AspNetCore.Components.RenderTree` provides methods for manipulating components and elements.
+
+> [!WARNING]
+> Dynamic components are an advanced scenario and aren't recommended for general use. Dynamically generated components can bypass tag structure safety checks. Inadvertently rendering a faulty tag in the browser can result in a security vulnerability.
```
danroth27: What is the security vulnerability and safety check being referred to here?

aspnetcore/razor-components/components.md
```diff
@@ -761,3 +761,54 @@ The time is 10/04/2018 01:26:52.
...
+>
+> When creating components dynamically, **always** hardcode the sequence number. Do **not** generate the sequence number using a calculation or counter.
```
danroth27: What is the motivation for this guidance?

aspnetcore/razor-components/components.md
guardrex: Discussed with Steve while there: He suggested this exact phrasing ("an advanced scenario" ... from my notes of the meeting). I inquired about a malformed tag escaping into the rendered output that might lead to a security vulnerability, and he said it's possible for that to happen.

aspnetcore/razor-components/components.md
guardrex: Same thing: Steve said to call this out ... he made it sound rather important. He implied that if the sequence number is calculated or based on a counter that it can result in malformed content.

aspnetcore/razor-components/components.md
SteveSandersonMS: I wouldn’t refer to a security issue unless there is a specific known one. I guess I was unclear, but my intent was to say you can just cause undefined behaviour if your tags don’t match.

guardrex: Change to a NOTE then? ... and soften the lingo?

aspnetcore/razor-components/components.md
SteveSandersonMS: The reason is that the diff algorithm relies on the sequence numbers corresponding to distinct lines of code, not distinct call invocations. If developers get this wrong, it won’t lead to anything being visibly  malformed, but will mean the diff produces a suboptimal output (as in, perf is worse). 

It’s problematic precisely because developers don’t realise they have caused a perf problem by trying to be too smart.

guardrex: Thx ... I'll patch it up now.

aspnetcore/razor-components/components.md
SteveSandersonMS: Up to you - I just suggest not conflating this with security at all, as that would be misleading.

aspnetcore/razor-components/components.md
danroth27: Given that the logic for dynamically generated rendering code may be arbitrarily complicated it's not clear to me that the sequence numbers can always be static and hard coded. For example, what if I'm adding N components where N is read from a database? It feels like there is a constraint on sequence numbers here that is not being clearly articulated. @SteveSandersonMS?

SteveSandersonMS:
I’d prefer not to use the term “dynamic components” for this. It implies that other components are static, which they are not. 

Can we use the term “manual RenderTreeBuilder logic” instead? I know that sounds more advanced, which is good because that matches the reality.

aspnetcore/razor-components/components.md
SteveSandersonMS: Sequence numbers are source code line numbers. Never generate them dynamically.

In your example, the N is irrelevant. It would never be an aspect of sequence numbers.

aspnetcore/razor-components/components.md
danroth27: I agree that dynamic components is an advanced scenario. But the security warning doesn't provide enough information to know that I am avoiding the unspecified vulnerability. What does it mean to render a "faulty tag" and what is the resulting security vulnerability? How do I know that I haven't rendered a faulty tag? How do I know if I've mitigated the vulnerability? What "safety check" do I need to replicate?

guardrex:
@SteveSandersonMS @danroth27 Synthesizing the remarks with some softening of the language ...

> \> \[!NOTE]
> \> Use of \`RenderTreeBuilder\` to create components is an advanced scenario. Malformed components (for example, an unclosed markup tag) can cause undefined behavior.
> \>
> \> When creating components with \`RenderTreeBuilder\`, hardcode the sequence number. Using a calculation or counter to generate the sequence number can lead to poor performance.

aspnetcore/razor-components/components.md
danroth27: So my for loop should look like this (where the sequence numbers never change)?:

```
int n = ReadCountFromDb();
for (int i = 0; i < n; i++) {
        builder.OpenComponent(0, typeof(PetDetails));
        builder.AddAttribute(1, "PetDetails", "Someone's best friend");              
        builder.CloseComponent();
}
```

SteveSandersonMS: Exactly!

danroth27: OK - I think this deserves more than just a note. We should add some text to this section that describes what the sequence numbers are and how they are used to motivate any guidance on how to specify them correctly.

guardrex: I'll draft some text (and move it out of the NOTE into the text) on the next commit within the hour.

guardrex:
@danroth27 How's it looking on the latest commit?

aspnetcore/razor-components/components.md
```diff
@@ -761,3 +761,61 @@ The time is 10/04/2018 01:26:52.
...
 ```
+
+## Manual RenderTreeBuilder logic
+
+`Microsoft.AspNetCore.Components.RenderTree` provides methods for manipulating components and elements, including building components manually in C# code.
+
+> [!NOTE]
+> Use of `RenderTreeBuilder` to create components is an advanced scenario. A malformed component (for example, an unclosed markup tag) can result in undefined behavior.
+
+When calling `RenderTreeBuilder` methods, sequence numbers are source code line numbers. The Razor Components difference algorithm relies on the sequence numbers corresponding to distinct lines of code, not distinct call invocations. When creating a component with `RenderTreeBuilder`, hardcode the sequence number. Using a calculation or counter to generate the sequence number can lead to poor performance.
```
danroth27: I think it would read better if this paragraph was moved to the end of the paragraph before the example that shows the sequence numbers

guardrex: I think I'll need to add something for the Pet Details component, too. It's somewhat out-of-context anyway ... not clear what purpose it serves.

aspnetcore/razor-components/components.md
```diff
@@ -761,3 +761,59 @@ The time is 10/04/2018 01:26:52.
...
+<p>@PetDetailsQuote<p>
+
+@functions
+{
+    [Parameter]
+    string PetDetailsQuote { get; set; }
+}
+```
+
+In the following example, the loop in the `CreateComponent` method generates three Pet Details components. When calling `RenderTreeBuilder` methods to create the components (`OpenComponent` and `AddAttribute`), sequence numbers are source code line numbers. The Razor Components difference algorithm relies on the sequence numbers corresponding to distinct lines of code, not distinct call invocations. When creating a component with `RenderTreeBuilder` methods, hardcode the arguments for sequence numbers. Using a calculation or counter to generate the sequence number can lead to poor performance.
```
Rick-Anderson: > Using a calculation or counter to generate the sequence number can lead to poor performance.

What about bolding that? Super good info.

guardrex:
@neven10 😅 Thanks again.

Lupusa87:
I am sorry but I don't agree here in lot of things.
It will be hard to explain but I will try.

I think we should not hide and put behind this great approach to build components dynamically.
Building dynamic components is strength of blazor which makes it powerful tech.

There is a question? Can we do everything what we need using both approaches, with razor pages and with buildrendertree?
I think no and name **dynamic** is relative here.

In razor pages we can use if, switch to adapt UI on parameters/state but it is still **static** at this point, when in code behind we render completely UI as we need and it can be 100% depended on code logic.

For example my [SvgHelper](https://github.com/Lupusa87/BlazorSvgHelper/blob/master/README.md) which is separated blazor lib (published on nuget) is receiving builder as parameter and adding tags based on what developer did with c# svg classes.
Can you do same in razor pages? No.

This approach is dynamic because everything is generated based on provided logic and methods (internal or external) can return pieces of UI and together build whole component body.

@rynowak Had idea to make same for razor pages which will be great but is not available yet.
https://twitter.com/aVerySpicyBoi/status/1104110920102375424

If someone likes razor way and thinks it is enough and finds it simpler, no problem, they can do it as they like.
But it does not give too much coding sugar and doing rendering in c# can be preferred way to many developers, especially when doing blazor native components, blazor libs, tools and bit complex apps.

I can agree that this is bit advanced scenario but not too much and it should be considered and recommended as general development style. 

There is nothing hard and highly advanced to use buildrendertree even for entry level developer, for bit experienced devs it should be trivial thing.

Warning that there can be unclosed tag is not relevant because same is still problem in razor page because compiler is not detecting it and any other html syntax errors.
Unclosed tag or any similar issue we detect and solve when build/test our apps and can not be reason not to chose buildrendertree way.

Hard code sequence numbers is not possible because as said already in c# class we can realize complex logic based on parameters/state and generate huge UI so it is impossible to hard code sequence, if it is problem it should be solved with another way rather than telling developers don't do this.
Dynamic nature makes no sense hard code recommendation.

Using razor pages should be recommended in very simple cases, leading way should be c# way.
It should be well documented, explained and supported with examples.

Blazor is going to be standard way to build web applications at least for .net developers and it should not concentrate on toys like razor page, most developers will skip it in most apps which are bit complex than counter sample, we expect in future to see large and wide range web apps ported or build in blazor, this naturally requires buildrendertree.

People with serious intentions to blazor should see how powerful mechanism do they have in their hands.

Components are basic concept for blazor and option to build them dynamically and unlimited is what we should tell everyone and make it mainstream.

Here you are people whom I hard imagine to teach something but in this case I can't not to protect better approach and not try to push it forward.

I am sorry if my comment is confusing or not clear, it is hard to do better because of my limited English.

Please consider that I am facing this situations each day and I share my thoughts with you as one blazor dev, you can ask another devs too.
People who work close to components will have fast answer on this issue.

Andrzej-W:
I have to say that @Lupusa87 is 100% right. If you want to create complex and reusable component it is sometimes impossible to do it in standard way (in .razor file). Even if it is possible it is often easier to do it in code. As @Lupusa87 said it have to be well documented and that is the only problem.

Nobody likes something like this in the doc:
> When creating components dynamically, always hardcode the sequence number. Do not generate the number using any sort of calculation or counter.

The question immediately arises: what is wrong with sequence++ statement?

SteveSandersonMS:
Sequence numbers are what the diff algorithm used to produce minimal diffs quickly. It can see where loops and branches are in your code. 

If you do seq++, you are providing literally no useful information. We still make it work, but the efficiency will be nowhere near as good. This is one of the great advantages of writing .razor files - the compiler deals with this for you automatically.

guardrex:
Let's move this ... we don't work on closed issues/PRs.

Sequence numbers are what the diff algorithm used to produce minimal diffs quickly. It can see where loops and branches are in your code.

https://github.com/aspnet/Docs/issues/11684

PR #11050, opened 2019-02-22
Update toc.yml

### LINKED ISSUE #6277: Remove version querystring from API Reference TOC link
The **API Reference** link in the TOC is hard-coded to use "?view=aspnetcore-2.0". We shouldn't assume the user always wants to view v2.0 of the API ref.

Rick-Anderson:
Fixes #6277

### DIFF
--- a/aspnetcore/toc.yml
+++ b/aspnetcore/toc.yml
@@ -876,6 +876,6 @@
     - name: Logging (not ASP.NET Core)
       uid: migration/logging-nonaspnetcore
 - name: API reference
-  href: /dotnet/api/?view=aspnetcore-2.2
+  href: /dotnet/api/
 - name: Contribute
   href: https://github.com/aspnet/Docs/blob/master/CONTRIBUTING.md


aspnetcore/toc.yml
```diff
@@ -876,6 +876,6 @@
     - name: Logging (not ASP.NET Core)
       uid: migration/logging-nonaspnetcore
 - name: API reference
-  href: /dotnet/api/?view=aspnetcore-2.2
+  href: /dotnet/api/
```
scottaddie: This URL will take the reader to the generic .NET API browser page: https://docs.microsoft.com/dotnet/api/. Is that the desired behavior? As far as I can tell, there's no way to link to an ASP.NET Core API ref landing page.

As an alternative, what if **API reference** were a top-level ToC node? Under it, we could include a link to each specific version's API ref landing page. It might look something like the following:

- API reference
  - ASP.NET Core 3.0 Preview 2
  - ASP.NET Core 2.2
  - ASP.NET Core 2.1
  - ASP.NET Core 2.0
  - ASP.NET Core 1.1
  - ASP.NET Core 1.0

Rick-Anderson: I forgot that. Let's just manually update to the latest released version.

PR #11057, opened 2019-02-22
Fix ambiguity in last paragraph regarding usage of statics

mahic:
The wording in the last paragraph is ambiguous - it addresses IHttpContextAccessor interface rather than the static HttpContextAccessor helper. For the reader, it might be difficult to understand that usage of the static helper - and not the interface - is discouraged.


pakrym:
> the static property on the `HttpContextAccessor` class

Which property are you referring to?

mahic:
@pakrym The `HttpContext?.User` property (the same property that the rest of the last paragraph is referring to, which is stated as the last element in the list). Perhaps the wording rather should refer to the `HttpContextAccessor` class instead of the property. Either way would be more clear.

aspnetcore/migration/claimsprincipal-current.md
```diff
@@ -50,4 +50,4 @@ There are several options for retrieving the current authenticated user's `Claim
   * Get an instance of `IHttpContextAccessor` during startup and store it in a static variable. The instance is made available to code that was previously retrieving the current user from a static property.
   * Retrieve the current user's `ClaimsPrincipal` using `HttpContextAccessor.HttpContext?.User`. If this code is used outside of the context of an HTTP request, the `HttpContext` is null.
 
-The final option, using `IHttpContextAccessor`, is contrary to ASP.NET Core principles (preferring injected dependencies to static dependencies). Plan to eventually remove the dependency on the static `IHttpContextAccessor` helper. It can be a useful bridge, though, when migrating large existing ASP.NET apps that were previously using `ClaimsPrincipal.Current`.
+The final option, using the static property on the `HttpContextAccessor` class, is contrary to ASP.NET Core principles (preferring injected dependencies to static dependencies). Plan to eventually remove the dependency on the static `HttpContextAccessor` helper. It can be a useful bridge, though, when migrating large existing ASP.NET apps that were previously using `ClaimsPrincipal.Current`.
```
mjrousos: I think this is still a little ambiguous. The thing we're recommending against here is storing an instance of `IHttpContextAccessor` in a static (whether it's in a helper class or not is less relevant). How about something like: 

```suggestion
The final option, using an `IHttpContextAccessor` instance stored in a static variable, is contrary to ASP.NET Core principles (preferring injected dependencies to static dependencies). Plan to eventually retrieve `IHttpContextAccessor` or `HttpContext` instances from dependency injection, instead. A static helper can be a useful bridge, though, when migrating large existing ASP.NET apps that were previously using `ClaimsPrincipal.Current`.
```

mahic: I understand the motivation behind the recommendation. But since the framework provides the `HttpContextAccessor.HttpContext` static, that will be the main source of use of a static accessor for most people, and at least for me, giving an example makes it really obvious. Perhaps the following wording: 

> The final option, using an `IHttpContextAccessor` instance stored in a static variable, is contrary to ASP.NET Core principles (preferring injected dependencies to static dependencies). Plan to eventually remove the use of all static accessors such as `HttpContextAccessor.HttpContext` and retrieve `IHttpContextAccessor` or `HttpContext` instances from dependency injection instead. A static helper can be a useful bridge, though, when migrating large existing ASP.NET apps that were previously using `ClaimsPrincipal.Current`.

And since you are already encouraging people to use DI, what would make it painfully obvious, is to add an `[Obsolete]` on the `HttpContextAccessor` helper for v3. Is that something you have considered?

pakrym:
cc @Tratcher

aspnetcore/migration/claimsprincipal-current.md
Tratcher: @mahic HttpContextAccessor is not static, nor is HttpContextAccessor.HttpContext. What we discourage is storing IHttpContextAccessor into your own static field. AspNetCore does not provide any static field containing an IHttpContextAccessor. @mjrousos's description is more accurate.

@mjrousos HttpContext cannot be retrieved directly from DI, only via IHttpContextAccessor.

mjrousos: @Tratcher Good catch about HttpContext not coming from DI.

aspnetcore/migration/claimsprincipal-current.md
```diff
@@ -50,4 +50,4 @@ There are several options for retrieving the current authenticated user's `Claim
...
   * Retrieve the current user's `ClaimsPrincipal` using `HttpContextAccessor.HttpContext?.User`. If this code is used outside of the context of an HTTP request, the `HttpContext` is null.
 
-The final option, using `IHttpContextAccessor`, is contrary to ASP.NET Core principles (preferring injected dependencies to static dependencies). Plan to eventually remove the dependency on the static `IHttpContextAccessor` helper. It can be a useful bridge, though, when migrating large existing ASP.NET apps that were previously using `ClaimsPrincipal.Current`.
+The final option, using an `IHttpContextAccessor` instance stored in a static variable, is contrary to ASP.NET Core principles (preferring injected dependencies to static dependencies). Plan to eventually retrieve `IHttpContextAccessor` or `HttpContext` instances from dependency injection instead. A static helper can be a useful bridge, though, when migrating large existing ASP.NET apps that were previously using `ClaimsPrincipal.Current`.
```
Tratcher:
```suggestion
The final option, using an `IHttpContextAccessor` instance stored in a static variable, is contrary to ASP.NET Core principles (preferring injected dependencies to static dependencies). Plan to eventually retrieve `IHttpContextAccessor` instances from dependency injection instead. A static helper can be a useful bridge, though, when migrating large existing ASP.NET apps that were previously using `ClaimsPrincipal.Current`.
```

scottaddie:
Thank you for your patience and for opening this PR to start the discussion, @mahic!

PR #11091, opened 2019-02-25
Integration testing non-Development environments

### LINKED ISSUE #11083: Development environment by design?!
WTF?! Nowhere this insane default is documented. 

```CSharp
protected virtual IWebHostBuilder CreateWebHostBuilder()
{
    var builder = WebHostBuilderFactory.CreateFromTypesAssemblyEntryPoint&lt;TEntryPoint&gt;(Array.Empty&lt;string&gt;());
     if (builder == null)
     {
         throw new InvalidOperationException(Resources.FormatMissingCreateWebHostBuilderMethod(
                    nameof(IWebHostBuilder),
                    typeof(TEntryPoint).Assembly.EntryPoint.DeclaringType.FullName,
                    typeof(WebApplicationFactory&lt;TEntryPoint&gt;).Name,
                    nameof(CreateWebHostBuilder)));
      }
      else
      {
          return builder.UseEnvironment("Development");
      }
}
```
It looks like sometimes decisions are made intentionally to make developers suffer. By default regular aspnet core application runs under `Production` environment, unless other value is set explicitly.

This is what YOUR [docs](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-2.2) state:

>If ASPNETCORE_ENVIRONMENT isn't set, it defaults to Production.

---
#### Document Details

* Content: [Integration tests in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-2.2#basic-tests-with-the-default-webapplicationfactory)
* Content Source: [aspnetcore/test/integration-tests.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/test/integration-tests.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #11083 
Addresses #8917

We worked on this earlier, but it didn't result in changes. Let's try again based on @voroninp's issue asking for a clarification. I updated (and hopefully improved) the wording.

### DIFF
--- a/aspnetcore/test/integration-tests.md
+++ b/aspnetcore/test/integration-tests.md
@@ -5,7 +5,7 @@ description: Learn how integration tests ensure that an app's components functio
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 01/11/2019
+ms.date: 02/25/2019
 uid: test/integration-tests
 ---
 # Integration tests in ASP.NET Core
@@ -103,6 +103,15 @@ These prerequisites can be seen in the [sample app](https://github.com/aspnet/Do
 * [xunit](https://www.nuget.org/packages/xunit/)
 * [xunit.runner.visualstudio](https://www.nuget.org/packages/xunit.runner.visualstudio/)
 * [AngleSharp](https://www.nuget.org/packages/AngleSharp/)
+
+## SUT environment
+
+If the SUT's [environment](xref:fundamentals/environments) isn't set, the environment defaults to Development. If the SUT requires a different environment when tests are executed (for example, Staging, Testing, Production):
+
+* Specify the environment on the test system (for example, set the `ASPNETCORE_ENVIRONMENT` environment variable to a value that reflects a staging or testing environment, such as `Staging` or `Testing`).
+* Provide code in the SUT's [Startup](xref:fundamentals/startup) that loads services and configures the request pipeline based on its environment.
+
+For more information, see <xref:fundamentals/environments>.
 
 ## Basic tests with the default WebApplicationFactory
 
--- a/aspnetcore/test/integration-tests.md
+++ b/aspnetcore/test/integration-tests.md
@@ -106,12 +106,7 @@ These prerequisites can be seen in the [sample app](https://github.com/aspnet/Do
 
 ## SUT environment
 
-If the SUT's [environment](xref:fundamentals/environments) isn't set, the environment defaults to Development. If the SUT requires a different environment when tests are executed (for example, Staging, Testing, Production):
-
-* Specify the environment on the test system (for example, set the `ASPNETCORE_ENVIRONMENT` environment variable to a value that reflects a staging or testing environment, such as `Staging` or `Testing`).
-* Provide code in the SUT's [Startup](xref:fundamentals/startup) that loads services and configures the request pipeline based on its environment.
-
-For more information, see <xref:fundamentals/environments>.
+If the SUT's [environment](xref:fundamentals/environments) isn't set, the environment defaults to Development.
 
 ## Basic tests with the default WebApplicationFactory
 


voroninp:
I solved it a bit differently, though =)

```CSharp
internal sealed class WebAppFactory<TStartup> : WebApplicationFactory<TStartup> where TStartup: class
{
    {
        var location = Assembly.GetCallingAssembly().Location;

        builder.UseEnvironment("Testing");
        builder.ConfigureAppConfiguration(cfg => cfg.AddJsonFile("appsettings.Testing.json"));
    }
}
```

guardrex:
There are a few ways to do it. They're described in issue and the PR that we closed earlier.

* https://github.com/aspnet/Docs/issues/8917
* https://github.com/aspnet/Docs/pull/9060 (the early commits had them IIRC ... later commits whittled it down to about what's on this PR ... i.e., the simplest approach perhaps)

guardrex:
I reviewed what we discussed on the earlier issue/PR and ran a test here. Setting the env var doesn't work. The options are ... 

* Using a [CustomWebApplicationFactory](#customize-webapplicationfactory), call `UseEnvironment` on the `IWebHostBuilder`:

    ```csharp
    {
        builder.UseEnvironment("Staging")
        ...
    ```

* When calling [WithWebHostBuilder](#customize-the-client-with-withwebhostbuilder), call `UseSetting` on the `Environment` key:
    
    ```csharp
    var client = _factory.WithWebHostBuilder(builder =>
    {
        builder.UseSetting("Environment", "Staging");
        ...
    ```

Since those weren't approved before, I'll change the text to only comment on the default.

guardrex:
@Rick-Anderson I don't think we need engineering review on this now. It's stating a simple fact that the SUT's environment is Development when not set.

voroninp:
> It's stating a simple fact

Not that simple, if it requires some time to understand how to change the environment. It's even begs the question why 'Dev'? Why not 'Test'? 
Guys, it is too implicit. =( Explicitly requiring environment name when instantiating `WebApplicationFactory` would be a much better option, or maybe a constructor having a parameter with  default value equal to `Development`.

At least add the comment about default environment in xmldoc of `WebApplicationFactory` constructor, please.

MikeBeaton:
It's not completely obvious why the `CustomWebApplicationFactory` version [above](https://github.com/aspnet/AspNetCore.Docs/pull/11091#issuecomment-467262297) is able to work using `builder.UseEnvironment("Staging")` whereas the `WithWebHostBuilder` version apparently requires  `builder.UseSetting("Environment", "Staging")` instead, when both operate on an `IWebHostBuilder` instance.

The reason is that you need to add `using Microsoft.AspNetCore.Hosting` to access the extension method `builder.UseEnvironment` (but also, you *have* to add `using Microsoft.AspNetCore.Hosting` to explictly access `IWebHostBuilder`, as you are doing in the first version but not the second).

I agree with @voroninp that it would be very useful to add a line to this documentation saying *how* to change the environment, since it's definitely not obvious. E.g If one was stumbling around using code-completion to try to find a relevant settings, one might well guess that `WithWebHostBuilder` is likely to require some fairly heavyweight set up to get it off the ground and move on, whereas apparently it doesn't require anything; and even if one guessed right that this was the place to look , one still wouldn't stumble across the `UseEnvironment` extension method because one likely wouldn't have the relevant using `statement`.

It seems like it would be best to document the `UseEnvironment` method as opposed to the `UseSetting` method, with a note that it is an extension method which is only available when `using Microsoft.AspNetCore.Hosting`.

guardrex:
@MikeBeaton Yes ... #12303 tracks this. Engineering will take a look and determine what they'd like to do.

I'm going to lock this one down. Please post further comments on #12303. This issue is cross-linked into that issue, so all of the comments here thus far are visible when that comes up for work via the cross-link.

PR #11096, opened 2019-02-25
Updating index page to use correct terminology consistently

timheuer:
Quickstart links refer to ASP.NET Core and .NET Core...we should be consistent to what it is (ASP.NET Core)

### DIFF
--- a/aspnetcore/host-and-deploy/azure-apps/index.md
+++ b/aspnetcore/host-and-deploy/azure-apps/index.md
@@ -5,7 +5,7 @@ description: This article contains links to Azure host and deploy resources.
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
 ms.custom: mvc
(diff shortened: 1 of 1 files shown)


timheuer:
@guardrex But this didn't take my suggestion?  The Linux one still has just ".NET Core" isntead of "ASP.NET Core"?

guardrex:
That's the title of the topic. Open an issue at the bottom of that topic with their **Content feedback** button and see if they'll change it, then we can change it over here.

https://docs.microsoft.com/en-us/azure/app-service/containers/quickstart-dotnetcore

guardrex:
I've marked this PR here, so I won't lose this. I'll keep an :eye: on it. If you want me to open that issue with them, let me know.

timheuer:
Related: https://github.com/MicrosoftDocs/azure-docs/issues/25709

PR #11097, opened 2019-02-25
DI -> Dependency Injection

pavanagrawal123:
I'm not sure if using an abbreviation in a first time, tutorial doc is a good idea...


guardrex:
It's defined on Line 178 ...

> In ASP.NET Core, services such as the DB context must be registered with the \[dependency injection (DI)](xref:fundamentals/dependency-injection) container. The container provides the service to controllers.

pavanagrawal123:
@guardrex Whoops missed that! Also noticed this doc still uses [ApiController]... AFAIK, we're supposed to use [Controller] now?

guardrex:
I don't see anything in Scott's text that raises concern in general ... https://docs.microsoft.com/aspnet/core/web-api/#annotation-with-apicontroller-attribute

pavanagrawal123:
Ahh I was using an older version of ASP.Net, sorry for the noob questions, and thank you for the prompt responses @guardrex  :)

PR #11108, opened 2019-02-26
Advice on setting IsPersistent for Cookies.

### LINKED ISSUE #10145: You can set an absolute expiration time with ExpiresUtc
According to the docs: 
"You can set an absolute expiration time with `ExpiresUtc`. You must also set `IsPersistent`; otherwise, `ExpiresUtc` is ignored and a single-session cookie is created." 
However, I am running the sample app code (on .Net Core 2.1), and when I set ExpiresUtc=DateTimeOffset.UtcNow.AddSeconds(10)  and IsPersistent=false,
 the user is logged out after 10 seconds of inactivity. So ExpiresUtc is not ignored despite the fact that IsPersistent is not set.  This actually makes sense, because I would like to have control on the length of the session of the user when the cookie is not persistent, but it seems to contradict the docs. Can I count on this behavior in production code?

---
#### Document Details

* ID: d2229fc6-f8c8-952e-e567-b1d5a463055a
* Version Independent ID: b18a015b-914d-f014-f711-120d208904d9
* Content: [Use cookie authentication without ASP.NET Core Identity](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/cookie?view=aspnetcore-2.2)
* Content Source: [aspnetcore/security/authentication/cookie.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/security/authentication/cookie.md#absolute-cookie-expiration)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

serpent5:
Fixes #10145.

### DIFF
--- a/aspnetcore/security/authentication/cookie.md
+++ b/aspnetcore/security/authentication/cookie.md
@@ -398,7 +398,7 @@ The [AuthenticationProperties](/dotnet/api/microsoft.aspnetcore.http.authenticat
 
 ## Absolute cookie expiration
 
-You can set an absolute expiration time with `ExpiresUtc`. You must also set `IsPersistent`; otherwise, `ExpiresUtc` is ignored and a single-session cookie is created. When `ExpiresUtc` is set on `SignInAsync`, it overrides the value of the `ExpireTimeSpan` option of `CookieAuthenticationOptions`, if set.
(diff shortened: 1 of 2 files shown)


Rick-Anderson:
@Tratcher 	or @haok please review

PR #11116, opened 2019-02-26
Add suggested extensions file for aspnet/Docs repo

scottaddie:
Fixes https://github.com/aspnet/Docs/issues/11115

Altered the *.gitignore* file to allow check-in of the *extensions.json* file. Suggests the Docs Authoring Pack extension for VS Code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,8 @@ Project_Readme.html
 *.user
 *.userosscache
 *.sln.docstates
-*.vscode/
+.vscode/*
+!.vscode/extensions.json
 # User-specific files (MonoDevelop/Xamarin Studio)
 *.userprefs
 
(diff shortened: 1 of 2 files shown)


guardrex:
@scottaddie I think this rule ...

```
.vscode/*
```

... is problematic. It's now trying to load *ALL* of my *tasks.json* and *launch.json* files on my local system into the repo for all of the samples that I've worked directly in the repo.

Was this on purpose? ... I think given the occasional churn in VSC files, I'm concerned that this could end in tears down the road. It also might break in some cases ... some of my files may have custom config for running samples here that might break users on their own systems.

guardrex:
Also, devs not using VSC might be confused.

We also don't load solution files and other VS files and folders.

PR #11121, opened 2019-02-26
Fixed Directive name does not render in column

EdCharbeneau:
The @ symbol was causing the Directive column values to not render. Added `\` to have the values render properly on the final output.


scottaddie:
Thanks for fixing this, @EdCharbeneau! The changes will go live within the next hour.

Issue #10392, opened 2019-01-15
Add doc section:Enable using a Cookie Session Store from DI


Rick-Anderson:
From https://github.com/aspnet/Security/pull/1719

`@HaoK` wrote:
We should have an explicit section recommending this pattern of how to configure options instances with things from DI in this doc.
```
public class UseSomethingFromDI : IConfigureNamedOptions/IConfigureOptions<YourOptions> {
       public CookieConfig(ISomethingFromDI thing) => _thing = thing;
       public void Configure([string name,] YourOptions o) { o.Thing = thing; }
   }
   services.AddTransient<IConfigureOptions<YourOptions>, UseSomethingFromDI>();
```

We can also demonstrate a new in 2.1 API ConfigureOptions here as well maybe:

`services.ConfigureOptions<UseSomethingFromDI>` instead of `services.AddTransient<IConfigureOptions<YourOptions>, UseSomethingFromDI>();`, this uses reflection to basically do the same thing and also lets UseSomethingFromDI configure multiple option types if desired.

Moved from #5922

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ed39cb0e-aa85-8bba-ffb4-859b56a60723
* Version Independent ID: dc7e3449-650b-d117-5eac-69e5ccb250cb
* Content: [Options pattern in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-2.2)
* Content Source: [aspnetcore/fundamentals/configuration/options.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/configuration/options.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixed by #11128

PR #11128, opened 2019-02-26
Using services to configure options

### LINKED ISSUE #10553: Add a doc on how to use services from DI when configuring options
The current [Options](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options) topic is mostly of reference of everything you can do with options. We need a doc that is more task oriented that helps users figure out how to access DI services when configuring options for various components.

@Rick-Anderson  edit:
27 K PV, 50% CSAT

---
#### Document Details

* Content: [Options pattern in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-3.0#feedback)
* Content Source: [aspnetcore/fundamentals/configuration/options.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/fundamentals/configuration/options.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

rynowak:
Improves the exising docs for this area and changes the title so it's
easier to find.

### DIFF
--- a/aspnetcore/fundamentals/configuration/options.md
+++ b/aspnetcore/fundamentals/configuration/options.md
@@ -268,9 +268,11 @@ services.AddOptions<MyOptions>("optionalName")
     .Configure(o => o.Property = "named");
 ```
 
-## Configure&lt;TOptions, TDep1, ... TDep4&gt; method
+## Using DI Services to configure options
 
-Using services from DI to configure options by implementing `IConfigure[Named]Options` in a boilerplate manner is verbose. Overloads for `ConfigureOptions` on `OptionsBuilder<TOptions>` allow you to use up to five services to configure options:
+You can access other services from dependency injection while configuring options in two ways. 
+
+The simplest way is to pass a configuration delegate to `Configure` on `OptionsBuilder<TOptions>`. `OptionsBuilder<TOptions>` provides overloads of `Configure` that allow you to use up to five services to configure options:
 
 ```csharp
 services.AddOptions<MyOptions>("optionalName")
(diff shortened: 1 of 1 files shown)


guardrex:
@rynowak Does this also resolve https://github.com/aspnet/Docs/issues/10392?

Also, just to confirm it from the title: Does this fully resolve #10553? It seems like the ask goes a bit further to reorganize/refocus the whole topic.

rynowak:
#10392 seems to recommend that you write a class to configure options. I don't think we'd recommend that over the simpler options.  If you think it's really important to have an example, I can add one since I already have a few. 

#10553  says

> We need a doc that is more task oriented that **helps users figure out how to access DI services when configuring options for various components.**

I think it does this now.

Rewriting this doc is out of scope for what I'm prepared to do. I'd be happy to see this merged and leave the work item open if that what you'd prefer.

guardrex:
>  If you think it's really important to have an example

Let's ask @HaoK ... Should we keep (and work) #10392 separately, take @rynowak up on his offer of an example (probably as a new section in the topic), or close that issue?

> if that what you'd prefer

I think that I agree ... in a biased kind of way given that I wrote it under @HaoK's :eye:. I just want to make sure that @danroth27 is 😀. If @danroth27 is calling for a different organization/approach/examples, then I'd like to take a stab at resolving his concerns separately from this PR. Let's merge, and I'll follow-up with further work on #10553 later.

HaoK:
This looks fine, the class was needed before 2.1 when we added the Configure overloads that take up to 5 services

guardrex:
Cool ...

> was needed before

I'm working to 🏃 catch-up 🏃 on older-ish issues before 3.0 hits. Sorry for any delays on our side. 😅

For now, let's leave this as ...

Addresses #10553 

If he doesn't respond here, I'll take up #10553 with @danroth27 shortly.

Stand-by ... I'll UE and *grammarize*:tm: this now and get it in.

guardrex:
@rynowak I re-organized a bit placing the reasoning for going with the first approach in a new paragraph. See if I botched anything on the update.

PR #11154, opened 2019-02-27
Update policyschemes.md

Rick-Anderson:
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/security/authentication/policyschemes?view=aspnetcore-1.0&branch=pr-en-us-11154)


Rick-Anderson:
@HaoK 	we format code using the widest browser window that doesn't have left or right TOC. Code is formatted to not have a horizontal scroll bar. That works best on tablets.

PR #11155, opened 2019-02-27
Add more ANCM startup failures

### LINKED ISSUE #10725: Add more ANCM startup failure troubleshooting docs
There are two cases that we are not covering:

1. Having two preview site extensions installed would cause 500 error
image
2. Publishing standalone with bitness mismatch between app and AppServices siet would cause the same 500 error

guardrex:
Fixes #10725 

> Confirm that the app's **Platform** in **Application Settings** matches the bitness of the app.

~I still haven't figured out how to recover a 32-bit app set to 64-bit in the portal and then set back to 32-bit. It doesn't seem to recover on its own by merely flipping the switch in the portal back to 32-bit and restarting the app. I show the errors from logging back over on the issue for this scenario.~

~It does recover if the x86 + x64 site extensions are present and then the x64 site extension is deleted (or the site extension is just added and then deleted and added back) ... of course, following a restart of the site.~

Checked with a fresh service this morning, and I can't repro the failure behavior. I think there is **_some combo_** of adding and subtracting site extensions and flipping the 32/64 platform setting that can lead to an unstable service, but I don't know exactly how to repro it. I was performing many service updates to characterize the behaviors for this PR **_using the same service instance_** that a normal dev under normal circumstances wouldn't be doing. Let's move on. I'm sure if their is a prob, we (or the Azure folks) will hear about it from users.

### DIFF
--- a/aspnetcore/host-and-deploy/azure-iis-errors-reference.md
+++ b/aspnetcore/host-and-deploy/azure-iis-errors-reference.md
@@ -4,7 +4,7 @@ author: guardrex
 description: Obtain troubleshooting advice for common errors when hosting ASP.NET Core apps on Azure Apps Service and IIS.
 ms.author: riande
 ms.custom: mvc
-ms.date: 02/21/2019
+ms.date: 02/26/2019
 uid: host-and-deploy/azure-iis-errors-reference
 ---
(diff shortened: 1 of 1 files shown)


guardrex:
@jkotalik I made a few updates ... hopefully improvements. It's ready for your :eye:.

Issue #11118, opened 2019-02-26
RedisCache links should go to the newer Caching.StackExchangeRedis package


marekr:
The page still links to the older RedisCache package that is effectively deprecated.
There's the newer Microsoft.Extensions.Caching.StackExchangeRedis package that uses the 2.x versions of stack exchange redis and is also built for 2.2 (and I'm guessing later 3.0)

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0d01025d-86b2-bb2d-86c1-ef0695bc98e3
* Version Independent ID: 5b9a6572-3671-0d54-b2ef-83d04f8b3e96
* Content: [Distributed caching in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/distributed?view=aspnetcore-2.2#distributed-redis-cache)
* Content Source: [aspnetcore/performance/caching/distributed.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/performance/caching/distributed.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Thanks for the heads-up @marekr ... I'll take a look this afternoon.

PR #11160, opened 2019-02-27
Update Redis section of dist caching

guardrex:
Addresses #11118 

Thanks @marekkar

@natemcmaster Need a tip .... I couldn't get API Browser links for: 

* `AddStackExchangeRedisCache`
* `RedisCache` (in the `Microsoft.Extensions.Caching.StackExchangeRedis` namespace)

They live for 2.2 ...

https://github.com/aspnet/Extensions/blob/release/2.2/src/Caching/StackExchangeRedis/src/StackExchangeRedisCacheServiceCollectionExtensions.cs#L23

https://github.com/aspnet/Extensions/blob/release/2.2/src/Caching/StackExchangeRedis/src/RedisCache.cs#L13

... and here's the 2.2 package ...

https://www.nuget.org/packages/Microsoft.Extensions.Caching.StackExchangeRedis/2.2.0

... but no :heart: from the API Browser ...

capture2

Is that something that someone needs to know?

I've code-fenced them into the topic, so we can proceed here without the API pages if that's what we need to do for now.

### DIFF
--- a/aspnetcore/performance/caching/distributed.md
+++ b/aspnetcore/performance/caching/distributed.md
@@ -134,8 +134,23 @@ The sample app implements <xref:Microsoft.Extensions.Caching.SqlServer.SqlServer
 
 ### Distributed Redis Cache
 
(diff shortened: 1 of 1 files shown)


natemcmaster:
cc @JunTaoLuo @muratg

I'm on parental leave so won't be able to review this PR.

>

JunTaoLuo:
Hmm that's odd, let me see why the API docs are missing.

guardrex:
@Rick-Anderson If ur ok with just code fencing these two references (`RedisCache`, `AddStackExchangeRedisCache`) for now, we can go ahead and merge this and then update the API links on the next doc update.

PR #11165, opened 2019-02-27
Fix glob pattern in .gitignore

scottaddie:
Related to https://github.com/aspnet/Docs/pull/11116. Restores the original *.vscode* pattern.


scottaddie:
@guardrex Thank you for reporting the problem!

PR #11214, opened 2019-02-28
Jkotalik/request response docs

tdykstra:
I moved some things around and the diff is pretty confused.  

Also a question -- if this is about how to read and write, should it have write examples also?


jkotalik:
> Also a question -- if this is about how to read and write, should it have write examples also?

We can maybe give more information about how to Write with pipes, but I don't think we should have another full blown sample/example.

PR #11267, opened 2019-03-04
Enhance Configuration topic DI and startup remarks

### LINKED ISSUE #11265: Config retrieval examples
From the GetValue section and onward, the examples use a "config" object. However, there is no guidance or examples on how to actually get a config object. It would be helpful to show a complete example.

It is not clear if configs are available in the ConfigureServices() context when is called after CreateDefaultBuilder(). 
Or if it is necessary to wait for Startup to access the IConfiguration object.

---
#### Document Details

* Content: [Configuration in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-2.2#configureappconfiguration)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #11265 

Thanks @gaziqbal.

### DIFF
--- a/aspnetcore/fundamentals/configuration/index.md
+++ b/aspnetcore/fundamentals/configuration/index.md
@@ -1326,10 +1326,29 @@ var host = new WebHostBuilder()
 
 [ConfigurationBinder.GetValue&lt;T&gt;](xref:Microsoft.Extensions.Configuration.ConfigurationBinder.GetValue*) extracts a value from configuration with a specified key and converts it to the specified type. An overload permits you to provide a default value if the key isn't found.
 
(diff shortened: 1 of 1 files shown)


guardrex:
Yeah, I agree with you @gaziqbal ... stand-by ... I'll repeat that. It takes *a long time* to read past all of the config providers. The reader may lose track of how to get the values out by the time they get to `GetValue`. The first example there can repeat it. Commit on the way ..........

guardrex:
@gaziqbal It's a repeat, but it isn't an exact repeat given that this time it stores the value in a page model property. It's a reminder after reading a 🚢 *sea of text* 🚢 on providers how the DI works. I think after this point, the reader should be ok with the rest of the examples. Do you concur?

guardrex:
I don't repeat anything from providers. The reader was just exposed to all of them, and there are plenty of examples in those sections. I'm more concerned about them getting lost on getting the values out.

gaziqbal:
Thank you!

PR #11277, opened 2019-03-05
Change AddIdentity to AddDefaultIdentity

StevenTCramer:
Given the reference is to AddDefaultIdentity the text should be the same.


Rick-Anderson:
@StevenTCramer 	 thanks!

PR #11293, opened 2019-03-06
Duplicate ToC  nodes for tutorial series

### LINKED ISSUE #10207: Duplicate tutorial series nodes in ToC
Currently a tutorial series has only its overview article in the Tutorials section of the ToC, while the individual tutorials are only under the subject area node (MVC, Web API, Real time, etc.). Per the treejack ToC test, people often look in the Tutorials area for specific tasks, but they won't find any if we only have the overview article there.

tdykstra:
Fixes #10207

[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/tutorials/razor-pages/razor-pages-start?view=aspnetcore-2.2&branch=pr-en-us-11293&tabs=visual-studio)

@Rick-Anderson @scottaddie @wadepickett I've implemented this for one series -- all tutorials in the RP Intro series are duplicated under the Tutorials top-level node and the subject area top-level node (Web apps).  When you click on a link in the ToC, only the node you clicked on is highlighted, but when you get to one of these tutorials some other way, like clicking Next or Previous buttons, the same tutorial in both places in the ToC is highlighted.  Since you'd normally be using Next/Previous in a series, most of the time you'd see two nodes highlighted, which is not ideal.  Is it worth putting up with this behavior to make the individual tutorials more discoverable in the ToC?

### DIFF
--- a/aspnetcore/toc.yml
+++ b/aspnetcore/toc.yml
@@ -142,8 +142,8 @@
               uid: data/ef-mvc/inheritance
             - name: Advanced topics
               uid: data/ef-mvc/advanced
-  - name: Fundamentals
-    items: 
+- name: Fundamentals
+  items: 
     - name: Overview
       uid: fundamentals/index
     - name: The Startup class
--- a/aspnetcore/toc.yml
+++ b/aspnetcore/toc.yml
@@ -46,7 +46,7 @@
               uid: tutorials/razor-pages/model
             - name: Scaffolding
               uid: tutorials/razor-pages/page
-            - name: Work with a DB
+            - name: Work with a database
               uid: tutorials/razor-pages/sql
             - name: Update the pages
               uid: tutorials/razor-pages/da1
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
I think it's worth it, but not by much. I could go with either approach.

aspnetcore/toc.yml
```diff
@@ -37,7 +37,25 @@
...
+          items: 
+            - name: Overview
+              uid: tutorials/razor-pages/index
+            - name: Get started
+              uid: tutorials/razor-pages/razor-pages-start
+            - name: Add a model
+              uid: tutorials/razor-pages/model
+            - name: Scaffolding
+              uid: tutorials/razor-pages/page
+            - name: Work with a DB
```
scottaddie:
```suggestion
            - name: Work with a database
```

tdykstra:
+1/2 from Rick and +1 from Scott = go -ahead. @RickAndMSFT @scottaddie I extended this to the other three series -- could one of you do a quick sanity check to make sure everything is in order?

aspnetcore/toc.yml
```diff
@@ -53,15 +93,55 @@
...
+        - name: EF Core with MVC, existing database
           displayName: tutorial
           href: /ef/core/get-started/aspnetcore/existing-db
-        - name: EF Core with MVC, new DB
+        - name: EF Core with MVC, new database
           displayName: tutorial
           href: /ef/core/get-started/aspnetcore/new-db
-        - name: EF Core with MVC, long tutorial
-          uid: data/ef-mvc/index
+        - name: EF Core with MVC, 10 tutorials
```
scottaddie: Shouldn't this read "long tutorial" instead of "10 tutorials"?

tdykstra: That's what it was originally, I changed it in both places thinking "10 tutorials" would better indicate what to expect than "long tutorial"

Rick-Anderson:
`@RickAndMSFT` never gets checked. Use @Rick-Anderson

PR #11307, opened 2019-03-07
Update first-mongo-app.md

JasonGT:
Within **Create the ASP.NET Core web API project**, change from **ASP.NET Core 2.1** to  **ASP.NET Core 2.2**.


scottaddie:
Thank you for taking the time to fix this, @JasonGT!

PR #11309, opened 2019-03-07
Course -> Courses

imba-tjd:
Fix https://github.com/aspnet/Docs/issues/11299


Rick-Anderson:
@imba-tjd  I don't think we can take this change. That change can have a major problematic impact. We'll look at this next time we  refresh the docs. cc @tdykstra

PR #11365, opened 2019-03-11
Configuration topic+sample update

### LINKED ISSUE #11361: Overriding appsettings.json from environment variables with prefix
From my understanding, for "environment variables prefixed with `ASPNETCORE_`" I should be able to add config variables. This seems to work fine when the variable is not already set in e.g. `appsettings.json`. However if I try to override an already existing variable it does not work and the value of the variable from `appsettings.json` stays that way. However, if I remove the prefix from the environment variable it overrides it just fine. Is this the expected behavior? And if so, I think this should be made clear in the doc.

**More details:**

```
// appsettings.json
"ConnectionStrings" {
    "MyConnection": "Server=1.1.1.1;..."
}
```

```
// In VS2017, I set environment variables in project configuration:
ASPNETCORE_ConnectionStrings__MyConnection="Server=9.9.9.9;..."
```

Behavior:

```
// Expected true but is false
Configuration.GetConnectionString("MyConnection") == "Server=9.9.9.9;..."

// Expected false but is true
Configuration.GetConnectionString("MyConnection") == "Server=1.1.1.1;..."
```

If I instead set the following environment variables, then the above expectations hold:
```
// Note: prefix is removed
ConnectionStrings__MyConnection="Server=9.9.9.9;..."
```

```
// Expected true and is true
Configuration.GetConnectionString("MyConnection") == "Server=9.9.9.9;..."

// Expected false and is false
Configuration.GetConnectionString("MyConnection") == "Server=1.1.1.1;..."
```

---
#### Document Details

* Content: [Configuration in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-2.1#environment-variables-configuration-provider)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #11361 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/index?view=aspnetcore-2.1&branch=pr-en-us-11365)

* :hocho: the <2.1 bits
* Sample to 2.2
* Drop static assets
* Focus item: Let's clarify the order of config providers with `CreateDefaultBuilder`.

cc: @Xerillio


guardrex:
@scottaddie can u review?

Xerillio:
Sorry for the late response but I think it explains pretty well how the prefix is used now, thanks 👍

PR #11368, opened 2019-03-12
Add notes on using clients with Azure SignalR Service serverless mode

anthonychu:
Add note that SignalR clients cannot call hub methods if used with SignalR Service in serverless mode.


scottaddie:
Thanks much for the PR, @anthonychu!

PR #11381, opened 2019-03-13
C73564: Adding backslash to non-clickable URLs

v-makoud:
Hello @rick-anderson,
Many thanks in advance.


Rick-Anderson:
@v-makoud thanks for this. See my changes. URLs should be \`code fenced\`

Rick-Anderson:
@v-makoud sorry for the delay. Do you need this PR merged into live?

Issue #10717, opened 2019-01-31
Blazor Topic Reviews


guardrex:
| Pre2 | Pre3 | Pre4 | Pre5 | Pre6 | Pre7 | Topic |
| :---: | :---: | :---: | :---: | :---: | :---: | --- |
| #10800 | #11388 | #12106 | &mdash; | #13070 |  | Overview |
| :x: | :x: | :x: | &mdash; |  #13070 |  | Supported platforms |
| #10873 | &mdash; | #12111 | #12289 | #13070 |  | Get started |
| &mdash; | #11721 | #12165 | &mdash; | #13070 |  | Hosting models |
| #10789 | #11390 | #12112 | &mdash; | #13070 |  | Tutorials: Build your first app |
| #10934 | #11391 | #12114 | &mdash; | #13075 |  | Components |
| :x: | :x: | :x: | &mdash; | #13083 |  | Forms and validation |
| #10971 | #11394 | #12132 | #12289 | #13083 |  | Component libraries |
| #10971 | #11395 | #12149 | &mdash; | #13083 |  | Layouts |
| #11012 | &mdash; | #12149 | &mdash; | #13083 |  | Dependency injection |
| #11012 | #11396 | #12149 | #12289 | #13083 |  | Routing |
| &mdash; | #11314<br>#11675 | #12165 | &mdash; | #13083 |  | JavaScript interop |
| :x: | :x: | :x: | :x: | &mdash; |  | Security: Security and Identity |
| :x: | :x: | :x: | :x: | :x: | :x: | Handle errors |
| :x: | :x: | :x: | :x: | :x: | :x: | State management |
| #11329 | &mdash; | &mdash; | &mdash; | &mdash; |  | Debug |
| :x: | :x: | :x: | :x: | &mdash; |  | Call a web API |
| &mdash; | #11723 | #12149 | &mdash; | &mdash; |  | Host/deploy: Overview |
| :x: | :x: | &mdash; | &mdash; | #13083 |  | Host/deploy: Client-side |
| :x: | :x: | #12149 | &mdash; | &mdash; |  | Host/deploy: Server-side |
| &mdash; | #11367 | #12149 | &mdash; | #13083 |  | Host/deploy: Configure the Linker |

guardrex:
@danroth27 I'm moving to the Components (_razor-components/components.md_) topic this afternoon. To confirm our game plan for this one:

* Keep the current Blazor sample app.
* Create a new RC sample app.
[...]

PR #11394, opened 2019-03-13
Razor Components: Component libs Preview3 updates

guardrex:
Addresses #10717


aspnetcore/razor-components/class-libraries.md
```diff
@@ -73,4 +93,4 @@ Upload the package to NuGet using the [dotnet nuget publish](/dotnet/core/tools/
 dotnet nuget publish
 ```
 
-Any included static resources are included in the NuGet package. Library consumers automatically receive scripts and stylesheets, so consumers aren't required to manually install the resources.
+When using the `blazorlib` template, any included static resources are included in the NuGet package. Library consumers automatically receive scripts and stylesheets, so consumers aren't required to manually install the resources.
```
tdykstra: I think you could drop the first "included"

PR #11407, opened 2019-03-14
Fix typo

Potapy4:
Just rename from `.Net` to `.NET`


scottaddie:
Thank you for fixing these, @Potapy4! Your changes will go live within the next 24 hours.

PR #11439, opened 2019-03-16
Missing highlighted code+remark

Ponant:
I highlighted the `services.AddTransient` for the custom token provider.

NB: There is also an empty code description below the line `The complete method is shown with the changed line highlighted`, which I did not correct since I do not know to what it refers to.


Rick-Anderson:
Thanks @Ponant

PR #11445, opened 2019-03-17
chore: Add Markdownlint config

nschonni:
Added the `default:true` and then turned off all the existing rules that had failures

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,0 +1,59 @@
+{
+    "default": true,
+
+    "blanks-around-fences": true,
+    "blanks-around-headers": true,
+    "fenced-code-language": true,
+    "first-header-h1": {
+        "level": 1
+    },
+    "header-increment": true,
+    "header-start-left": true,
+    "header-style": {
+        "style": "atx"
+    },
+    "hr-style": {
+        "style": "---"
+    },
+    "line-length": false,
+    "list-indent": true,
+    "list-marker-space": true,
+    "no-alt-text": true,
+    "no-bare-urls": true,
+    "no-blanks-blockquote": true,
+    "no-duplicate-header": true,
+    "no-empty-links": true,
+    "no-hard-tabs": true,
+    "no-inline-html": {
+        "allowed_elements": [
+            "a",
+            "br",
+            "p"
+        ]
+    },
+    "no-missing-space-atx": true,
+    "no-multiple-blanks": true,
+    "no-multiple-space-atx": true,
+    "no-multiple-space-blockquote": true,
+    "no-reversed-links": true,
+    "no-space-in-code": true,
+    "no-space-in-emphasis": true,
+    "no-space-in-links": true,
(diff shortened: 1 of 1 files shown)


nschonni:
@Rick-Anderson if the Core content is still staying here, I think this one still applies

nschonni:
Rebased since #11582 landed

scottaddie:
@nschonni I've started an internal conversation to decide which rules should be disabled via this config file.

nschonni:
No problem, this wasn't a recommendation on what should actually be disabled, it was just so that VS Code flags any new issues while people edit the files. I've just been slowly fixing the failures separately.
https://github.com/MicrosoftDocs/PowerShell-Docs/blob/staging/.markdownlint.json looks like they tried to match it up to a styleguide, but I'm not sure how it matches up.
Also, a few of these were disabled because of the things poping up with the checked in sample projects dependencies (jQuery Validation, Bootstrap, etc..) README files

.markdownlint.json
```diff
@@ -0,0 +1,36 @@
+{
```
scottaddie: Let's use the following as a starting point for the config file:

```json
{
    "default": true,

    "blanks-around-fences": true,
    "blanks-around-headers": true,
    "fenced-code-language": true,
    "first-header-h1": {
        "level": 1
    },
    "header-increment": true,
    "header-start-left": true,
    "header-style": {
        "style": "atx"
    },
    "hr-style": {
        "style": "---"
    },
    "line-length": false,
    "list-indent": true,
    "list-marker-space": true,
    "no-alt-text": true,
    "no-bare-urls": true,
    "no-blanks-blockquote": true,
    "no-duplicate-header": true,
    "no-empty-links": true,
    "no-hard-tabs": true,
    "no-inline-html": {
        "allowed_elements": [
            "a",
            "br",
            "p"
        ]
    },
    "no-missing-space-atx": true,
    "no-multiple-blanks": true,
    "no-multiple-space-atx": true,
    "no-multiple-space-blockquote": true,
    "no-reversed-links": true,
    "no-space-in-code": true,
    "no-space-in-emphasis": true,
    "no-space-in-links": true,
    "no-trailing-punctuation": true,
    "no-trailing-spaces": false,
    "ol-prefix": {
        "style": "one_or_ordered"
    },
    "proper-names": {
        "code_blocks": false,
        "names": [
            ".NET",
            "ASP.NET",
            "JavaScript",
            "NuGet",
            "PowerShell"
        ]
    },
    "single-h1": true,
    "ul-start-left": true
}
```

nschonni: I've added you config and updated the title to reflect that these rules aren't all currently passing

nschonni: Here is the current warnings (exlucding the sample's third party files) https://gist.github.com/nschonni/9ec9b1c00fb2fa98f39f9687079e13a8

scottaddie:
@nschonni Thank you for your help with this!

PR #11489, opened 2019-03-17
Update the Server name

phuctv95:
I've tested, the Server name "(localdb)\Projects13" lead to an error, the right value is "(localdb)\Projects**V**13".


Rick-Anderson:
@phuctv95 thanks for this fix. Can you make the change in the new repo https://github.com/aspnet/AspNetDocs ?

PR #11491, opened 2019-03-17
Update accconfirm.md

gitjafar:
added missing ending quote for aspnet-codegenerator command


scottaddie:
@gitjafar Thank you for the fix!

PR #11583, opened 2019-03-19
fix: MD038/no-space-in-code

nschonni:
- Replace ` with %60 in xref links
- Remove spaces inside code span elements

### DIFF
--- a/aspnetcore/fundamentals/configuration/options.md
+++ b/aspnetcore/fundamentals/configuration/options.md
@@ -32,22 +32,22 @@ Reference the [Microsoft.AspNetCore.App metapackage](xref:fundamentals/metapacka
 
 ## Options interfaces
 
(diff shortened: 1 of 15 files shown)


Rick-Anderson:
@scottaddie  @tdykstra @guardrex  do we want to replace Replace ` with %60 in xref links

guardrex:
@nschonni I see that this PR produced no build warnings/errors, so it should be ok, but ...

The build system is rendering good API links with backticks, which are probably in the `ms.asset` values.

What's the problem with backticks?
.

tdykstra:
> do we want to replace Replace ` with %60 in xref links

>What's the problem with backticks?

If %60 works I'd say it's preferable, since backticks get confused with inline code format backticks.

guardrex:
If u go that way, I'll get the XREF Gen app updated this weekend.

scottaddie:
%60 appears to work. I'm in favor of this change.

scottaddie:
@guardrex Go ahead with updating your app.

PR #11628, opened 2019-03-21
Updated for consistency

EdCharbeneau:
The step explaining this code shows usage of `Count(predicate)` instead of `Where(predicate).Count()`.

See step 14/15 of the given example.


scottaddie:
Nice catch, @EdCharbeneau! Thank you for fixing it too.

PR #11634, opened 2019-03-21
Update name of Blazor extension

SpaceShot:
Extension now shipped as Blazor in the marketplace and in the UI.  Reduce confusion by getting rid of the old name.


scottaddie:
Thank you for fixing this, @SpaceShot!

Issue #11575, opened 2019-03-19
Activation from a runtime store-deployed assembly


a14907:
It is realy hard to follow in this section! what stop 1 and step 2 mean!

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 613ad64e-bc62-edf7-2e86-42210de718f4
* Version Independent ID: aabbbc32-3237-5ef9-c422-892dc6884a22
* Content: [Use hosting startup assemblies in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/platform-specific-configuration?view=aspnetcore-2.2#feedback)
* Content Source: [aspnetcore/fundamentals/host/platform-specific-configuration.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/host/platform-specific-configuration.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Hello @a14907 ... Sorry for the delay getting back to you. I'm away at a conference, so communication is slow until I get back on Saturday.

Thanks for letting me know about that broken step. We made BIG changes to the demonstration with the sample about a week ago, and we failed to update this step (Step 2).

Until I get a chance to fix it ...

1. Execute the build.ps1 script in the RuntimeStore folder.
1. Run the deploy.ps1 script in the 'deployment' folder.

Fortunately, Pavel included code comments in the scripts to explain what's happening, and it matches what the topic describes earlier. 

I'll probably put in a quick PR today to patch that step, but I'd also like to put in a PR when I get back that explains those script comments in the doc. Doc content is localized, but code comments aren't.

PR #11646, opened 2019-03-22
Patch step in hosting startup topic

guardrex:
Addresses #11575

Will take a further look :eyes: when I get back. This at least addresses the immediate concern.


guardrex:
If this requires a patch, could u do it @scottaddie? ... This phone-built PR was hand-to-hand combat.

guardrex:
There's a warning ... but it's not related to this PR.

PR #11726, opened 2019-03-28
Corrected registration example of middleware

Greeng0:
Not sure if this is the right file to update as the documentation page is pointing to samples/2.0, but this file seems more recently maintained.

According to https://github.com/RicoSuter/NSwag/issues/869#issuecomment-390503562, app.UseSwagger() is included in app.UseSwaggerUi3(), so I removed the former.

https://github.com/aspnet/Docs/edit/master/aspnetcore/tutorials/getting-started-with-NSwag.md will probably need to be modified as well, so an empty line is not highlighted?


Greeng0:
After actually trying out the code, it seems that both app.UseSwagger() and app.UseSwaggerUi3() are required, so I cancelled this pull request.

Issue #11758, opened 2019-03-31
ConnectionStrings example would be handy


dustout:
When setting up the environment variables it would be nice if an example of a default connection string was provided, as the __ fix took quite a bit of time to find, and I think this will be a common issue. 
https://stackoverflow.com/questions/40094998/asp-net-core-environment-variable-colon-in-linux

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: dcdd2c35-bf2b-2673-d150-303a8aaa54d2
* Version Independent ID: 815fe034-c9fc-aee3-a8b5-f2e860b30302
* Content: [Host ASP.NET Core on Linux with Nginx](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/linux-nginx?view=aspnetcore-2.2#feedback)
* Content Source: [aspnetcore/host-and-deploy/linux-nginx.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/host-and-deploy/linux-nginx.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-hostdeploy**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

PR #11761, opened 2019-03-31
Mirror env vars separator remarks from Nginx topic

guardrex:
Addresses #11758 

Mirroring updates made on #11759 in the Linux-Apache topic.

If we need to make further updates to #11759 based on the review here, I'll add them to this PR.


guardrex:
Only two spots. I hope we can do this **_without_** an INCLUDE. This pair of topics sit SxS, and I have them under control. No worries on keeping them in-sync. Truly if we went with INCLUDES for these two topics, we would need *several of them* for this pair. That's why I don't want them. Too much pain when working on this pair to have several INCLUDES stuck in there hiding the content from me when I edit them.

PR #11777, opened 2019-04-01
Reset buffer before receiving socket messages

### LINKED ISSUE #11776: Missing buffer reset in socket example
In [Send and receive messages](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/websockets?view=aspnetcore-2.2#send-and-receive-messages) section, the featured example has an code issue.

If you don't clear the buffer before `ReceiveAsync` method inside the while loop, you can end up with som invalid data.

### Example:
- First received message is: `Hello, my name is Foo`
- Second received message is: `Hi there`

After this second message, the server will try to handle the following message:
```
Hi therey name is Foo 
```
Wich is the first (and longer) message overwritten at the begining by the second message because the buffer was not reset.
I was able to address this issue with the following code:

```
var buffer = new byte[1024 * 4];
var result = await webSocket.ReceiveAsync(new ArraySegment<byte>(buffer), CancellationToken.None);
[...]
while (!result.CloseStatus.HasValue)
{
    [...]
    buffer = new byte[1024 * 4]; // Clean the buffer in order to have consistent messages
    result = await webSocket.ReceiveAsync(new ArraySegment<byte>(buffer), CancellationToken.None);
}
[...]
await webSocket.CloseAsync(result.CloseStatus.Value, result.CloseStatusDescription, CancellationToken.None);
```

Hope this helps =)

---
#### Document Details

* Content: [WebSockets support in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/websockets?view=aspnetcore-2.2#feedback)
* Content Source: [aspnetcore/fundamentals/websockets.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/websockets.md)
* Product: **aspnet-core**

thiagosgarcia:
Fixes #11776 

When the buffer is not clean, messages can overlap one on another. This fixes this issue by resetting the buffer all over again before awaiting for new messages from the client.

### DIFF
--- a/aspnetcore/fundamentals/websockets/samples/1.x/WebSocketsSample/Startup.cs
+++ b/aspnetcore/fundamentals/websockets/samples/1.x/WebSocketsSample/Startup.cs
@@ -81,6 +81,7 @@ namespace EchoApp
             {
                 await webSocket.SendAsync(new ArraySegment<byte>(buffer, 0, result.Count), result.MessageType, result.EndOfMessage, CancellationToken.None);
 
+                Array.Clear(buffer, 0, result.Count);
                 result = await webSocket.ReceiveAsync(new ArraySegment<byte>(buffer), CancellationToken.None);
             }
             await webSocket.CloseAsync(result.CloseStatus.Value, result.CloseStatusDescription, CancellationToken.None);
--- a/aspnetcore/fundamentals/websockets/samples/2.x/WebSocketsSample/Startup.cs
+++ b/aspnetcore/fundamentals/websockets/samples/2.x/WebSocketsSample/Startup.cs
@@ -103,6 +103,7 @@ namespace EchoApp
             {
                 await webSocket.SendAsync(new ArraySegment<byte>(buffer, 0, result.Count), result.MessageType, result.EndOfMessage, CancellationToken.None);
 
+                Array.Clear(buffer, 0, result.Count);
                 result = await webSocket.ReceiveAsync(new ArraySegment<byte>(buffer), CancellationToken.None);
             }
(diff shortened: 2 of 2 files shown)


aspnetcore/fundamentals/websockets/samples/1.x/WebSocketsSample/Startup.cs
```diff
@@ -81,6 +81,7 @@ private async Task Echo(HttpContext context, WebSocket webSocket)
             {
                 await webSocket.SendAsync(new ArraySegment<byte>(buffer, 0, result.Count), result.MessageType, result.EndOfMessage, CancellationToken.None);
 
+                buffer = new byte[1024 * 4];
```
BrennanConroy: Use `Array.Clear` so you don't need to allocate a whole new buffer every loop

thiagosgarcia: That is, indeed, much better.
Thanks for pointing this out.

    Array.Clear(buffer, 0, result.Count);

davidfowl:
Isn’t this only a problem if you look at more bytes than were read?

thiagosgarcia:
@davidfowl you're right.
I've just noticed this code uses `new ArraySegment<byte>(buffer, 0, result.Count)` to read data. In my code I changed that in my code, in order to use with UTF8Enconding and JsonConvert class. 
This actually does not apply for this example. I'm gonna close the issue I opened and retrieve this PR.

Sorry about that!

PR #11784, opened 2019-04-02
fix: MD032/blanks-around-lists

nschonni:
Lists should be surrounded by blank lines

### DIFF
--- a/aspnetcore/data/ef-mvc/advanced.md
+++ b/aspnetcore/data/ef-mvc/advanced.md
@@ -16,6 +16,7 @@ In the previous tutorial, you implemented table-per-hierarchy inheritance. This 
 In this tutorial, you:
 
 > [!div class="checklist"]
+>
 > * Perform raw SQL queries
 > * Call a query to return entities
 > * Call a query to return other types
@@ -261,6 +262,7 @@ For information about other topics related to ASP.NET Core MVC, such as authenti
 In this tutorial, you:
 
 > [!div class="checklist"]
+>
 > * Performed raw SQL queries
 > * Called a query to return entities
 > * Called a query to return other types
--- a/aspnetcore/data/ef-mvc/complex-data-model.md
+++ b/aspnetcore/data/ef-mvc/complex-data-model.md
@@ -20,6 +20,7 @@ When you're finished, the entity classes will make up the completed data model t
 In this tutorial, you:
 
 > [!div class="checklist"]
+>
 > * Customize the Data model
 > * Make changes to Student entity
 > * Create Instructor entity
@@ -511,6 +512,7 @@ Right-click the **CourseAssignment** table and select **View Data** to verify th
 In this tutorial, you:
 
 > [!div class="checklist"]
+>
 > * Customized the Data model
 > * Made changes to Student entity
 > * Created Instructor entity
--- a/aspnetcore/data/ef-mvc/concurrency.md
+++ b/aspnetcore/data/ef-mvc/concurrency.md
@@ -22,6 +22,7 @@ You'll create web pages that work with the Department entity and handle concurre
 In this tutorial, you:
 
 > [!div class="checklist"]
+>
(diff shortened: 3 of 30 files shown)


guardrex:
Thanks @nschonni. WRT to the checklists, there might be a delta on the general guidance. @scottaddie will prob know.

scottaddie:
@guardrex It renders correctly with that change and addresses the Markdownlint warning. :shipit:

guardrex:
Excellent ... Thx. I'll review and merge this when I get home from the gym. :muscle:

tdykstra:
@nschonni Thanks for all your work on making our Markdown syntax more consistent -- much appreciated, but in this case the rule calls for blank lines where they don't really fit, specifically in the checklists that guardrex pointed out.

guardrex:
> specifically in the checklists that guardrex pointed out.

... but @scottaddie said 'ok' on those. So, we won't have the line in the check box list case? ... for my future reference.

tdykstra:
> So, we won't have the [extra] line in the check box list case?

That's correct, they can stay as they are.

scottaddie:
@guardrex For context, Rick, Tom, and myself chatted about this offline.

guardrex:
@nschonni If you'd like to resubmit the PR sans the check box list changes, we can get it merged.

guardrex:
As a quick side note: I've found that the rule for code doesn't work so well in lists. They seem to render better without the post-code block line. Example:

```
* Item 1

  '''csharp
  some code
  '''
* Item 2
* Item 3
```

Line before but no line after and before the next bullet.

This may have changed tho. My test and examination of the rendered output was *ages ago in ancient DocFX history before the dawn of time* ... i.e. ... **_last year_**. :smile:

nschonni:
> If you'd like to resubmit the PR sans the checklist changes, we can get it merged.

Looks like the rule is disabled now, so i don't think resubmitting makes sense now



> I've found that the rule for code doesn't work so well in lists.

Think this is a comment about #11799

guardrex:
In the case of within a NOTE/WARNING/etc., that's fine.

I'm referring to code in lists (the other changes that you made on #11799). It works better AFAIK without the second blank line (the one below) ... but only because it results in a lot of vertical margin (or padding) in the rendered list.

That's up to them. If they want the line above and below code blocks in lists, it's cool. I just need to know. Fortunately, this is one of those very, *VERY* tiny issues around here. I'm sure we'll eventually get consistent on it as we keep passing over topics with updates each release.

@RickAndMSFT @scottaddie @tdykstra @wadepickett Let me know how you want to play that aspect, and I'll continue with my current approach or change as I go forward.

guardrex:
> Looks like the rule is disabled now, so i don't think resubmitting makes sense now

I don't quite follow u there. You can resubmit this PR without the checklists changes. IIRC, the rest of your PR was fine. For example, we do want the blank line between a sentence that ends in a colon and the list.

tdykstra:
> Let me know how you want to play that aspect, 

I would say keep doing it the way you know renders well on the site. We need to have a team discussion to reach consensus on how to handle markdownlint rules, especially when they are problematic in some scenarios.

guardrex:
Sounds good.

I can't see the internal previews or else it would be easy to confirm if what I saw last year is still the case.

Comparing ...

```
* Item 1

  '''csharp
  some code
  '''
* Item 2
* Item 3
```

... to ...

```
* Item 1

  '''csharp
  some code
  '''

* Item 2
* Item 3
```

... and look at the margin/padding between Item 2 and Item 3. I guess *close margin of a normal list* is what you want. If so and if the 2nd example can pull it off now, cool. If you see extra (undesireable) vertical space between Item 2 and 3 in the 2nd example, then nothing has changed over last year, and I recommend we continue to use the first example that I show.

I'd be happy to check, but I can't see the internal review topics. I don't like the idea of messing with live topics to test the scenarios.

PR #11796, opened 2019-04-03
SEO April: include ASP.NET Core 3.0 in title

wadepickett:
SEO April Effort: This topic is for ASP.NET Core 3.0 and a top keyword being used in search included "3".  It is not immediately clear the topic is for ASP.NET Core 3.0 in search or even once you look at the topic.  Changing to see if the bounce rate improves.  Following up on bounce ratings next month for this as part of SEO hackadoc.  No related issue for this PR.

### DIFF
--- a/aspnetcore/razor-components/index.md
+++ b/aspnetcore/razor-components/index.md
@@ -1,10 +1,10 @@
 ---
-title: Introduction to Razor Components
+title: Introduction to Razor Components in ASP.NET Core 3.0
 author: guardrex
 description: Explore ASP.NET Core Razor Components, a way to build interactive client-side web UI with .NET in an ASP.NET Core app.
 monikerRange: '>= aspnetcore-3.0'
 ms.author: riande
-ms.custom: mvc
+ms.custom: seoapril2019
 ms.date: 03/27/2019
 uid: razor-components/index
 ---


guardrex:
Including a version in the title of a topic that spans versions via the version selector may cause problems:

* At 3.1 when the user selects 3.0, the title will still say 3.1 ... could be confusing to the visitor.
* We have to maintain this every patch release ... could be missed ... *someone probably will one day*. 🙈 
* What about the caching of the title on external systems (search engines being the worst case scenario)? It could cache with a 3.0 title but load 3.1. Again, confusing ... and perhaps even a little embarrassing.
* I suspect the topic will position well with Google ... it's look'in good today :+1: ...
  Capture

tdykstra:
+1 to @guardrex comments -- 
* We don't usually put version number in the title -- we'd have to change the title with every release. If we started doing that with this one we should be consistent and do it everywhere, but that would be confusing for customers and a lot of work to maintain.
* What's more relevant to Blazor is not its 3.0 status but its preview status -- if the goal is bounce rate reduction you'd probably get more by adding (Preview) than by adding 3.0.

wadepickett:
@guardrex and @tdykstra for the feedback.  Greatly appreciated. I did goof this up.  I will remove the version from the title.  The version is included in the text body, but far enough down it is not seen right away.  Tom pointed out it could also be that folks are figuring out late after searching and landing on it that it is preview and bouncing out.  I will move the version communicated later in the doc and move it to the intro paragraph which will show up in search results so it is clear up front.  It will have to be changed later of course, but it is also a high traffic topic, there is some confusion over it, and  worth some extra clarification and a later update.

PR #11798, opened 2019-04-03
Mention IMemoryCache lifetime in DI container

### LINKED ISSUE #11797: Mention IMemoryCache lifetime
The documentation does not mention that IMemoryCache lifetime is singleton in the DI container.

---
#### Document Details

* Content: [Cache in-memory in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/memory?view=aspnetcore-2.2)
* Content Source: [aspnetcore/performance/caching/memory.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/performance/caching/memory.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

natenho:
Fixes #11797

### DIFF
--- a/aspnetcore/performance/caching/memory.md
+++ b/aspnetcore/performance/caching/memory.md
@@ -59,6 +59,8 @@ Request the `IMemoryCache` instance in the constructor:
 
 [!code-csharp[](memory/sample/WebCache/Controllers/HomeController.cs?name=snippet_ctor)]
 
+`IMemoryCache` has a *Singleton* lifetime.
+
 ::: moniker range="< aspnetcore-2.0"
 
 `IMemoryCache` requires NuGet package [Microsoft.Extensions.Caching.Memory](https://www.nuget.org/packages/Microsoft.Extensions.Caching.Memory/).


Rick-Anderson:
That's what it has to be, I'm not sure this should be documented. I'll let @JunTaoLuo  decide.

natenho:
I understand, it looks obvious at first sight, but  what if someone is
looking for a scoped memorycache instance to be used along a request and
then get stuck figuring out the lifetime of the object they will receive?

JunTaoLuo:
This isn't quite accurate. The IMemoryCache is just an abstraction and doesn't imply any particular lifetime restrictions.

What you probably mean is that the [`AddMemoryCache`](https://github.com/aspnet/AspNetCore.Docs/blob/e7810f6b611b8eed75e17fddea6be2dbba73c4ed/aspnetcore/performance/caching/memory/sample/WebCache/Startup.cs#L9) method adds a singleton service: https://github.com/aspnet/Extensions/blob/c19b309cd8839c4e572a7fc10ae873b4c7ee1f93/src/Caching/Memory/src/MemoryCacheServiceCollectionExtensions.cs#L30.

Given it's the method we want to clarify the documentation for, I'd say the XML docs should be updated instead.

natenho:
> This isn't quite accurate. The IMemoryCache is just an abstraction and doesn't imply any particular lifetime restrictions.
> 
> What you probably mean is that the [`AddMemoryCache`](https://github.com/aspnet/AspNetCore.Docs/blob/e7810f6b611b8eed75e17fddea6be2dbba73c4ed/aspnetcore/performance/caching/memory/sample/WebCache/Startup.cs#L9) method adds a singleton service: https://github.com/aspnet/Extensions/blob/c19b309cd8839c4e572a7fc10ae873b4c7ee1f93/src/Caching/Memory/src/MemoryCacheServiceCollectionExtensions.cs#L30.
> 
> Given it's the method we want to clarify the documentation for, I'd say the XML docs should be updated instead.

I understand it's an abstraction, but the documentation is specifying the usage of the [`AddMemoryCache`] so adding the information the behavior would be more accurately explained without the need of looking the source code.

JunTaoLuo:
I suppose it wouldn't be a bad idea to add this to the doc if we were to change the wording to something like "AddMemoryCache adds a IMemoryCache with a Singleton lifetime to the service container" to the docs. XML docs feel like the more natural place for this but the duplication isn't bad. Up to @Rick-Anderson to decide.

PR #11799, opened 2019-04-03
fix: MD031/blanks-around-fences

nschonni:
Fenced code blocks should be surrounded by blank lines

### DIFF
--- a/aspnetcore/data/ef-mvc/migrations.md
+++ b/aspnetcore/data/ef-mvc/migrations.md
@@ -49,6 +49,7 @@ This change sets up the project so that the first migration will create a new da
 > ```console
 > dotnet ef database drop
 > ```
+>
 > The following section explains how to run CLI commands.
 
 ## Create an initial migration
--- a/aspnetcore/host-and-deploy/azure-apps/index.md
+++ b/aspnetcore/host-and-deploy/azure-apps/index.md
@@ -162,6 +162,7 @@ When the operation completes, the latest .NET Core preview is installed. Verify 
 > ```powershell
 > Test-Path D:\home\SiteExtensions\AspNetCoreRuntime.{X.Y}.x64\
 > ```
+>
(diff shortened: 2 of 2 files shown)


guardrex:
Under discussion on #11784.

I go with ...

```
* Item 1

  '''csharp
  some code
  '''
* Item 2
* Item 3
```

... to keep vertical margin (or padding) from being added to *all list items*. However, that should be checked because my check on the rendering difference was done last year. DocFX/build sys may be doing something different these days (i.e., **not** adding a lot of extra vertical space to all list items).

nschonni:
I don't think it affects the virtical margin, since looking at the code it renders out to the following on docsmsft
```
<li><p>Replace <code>UseIdentity</code> with <code>UseAuthentication</code> in the <code>Configure</code> method:</p>
<div class="codeHeader" id="code-try-2" data-bi-name="code-header"><span class="language">C#</span>
			<button class="action" data-bi-name="copy">
				<span class="docon docon-edit-copy" role="presentation"></span>
				<span>Copy</span>
			</button></div><pre><code class="lang-csharp" data-author-content="app.UseAuthentication();
">app.UseAuthentication();
</code></pre>
</li>
```
The GitHub render is also similar. EX: closing `<pre>` and then the closeing `</li>` in both cases

guardrex:
You can't see the classes tho, and one would need to look at both cases (line + no line) with markup+styles. It's actually quicker to just perform an internal preview test both ways ... that will settle it.

We're all a bit 🏃🏃🏃😅 at the moment tho with priority work (e.g., VS2019-related doc updates). They may want to revisit this later.

I recommend sitting on this for a bit until we can find out what's what. They'll let us know what they want to do.

nschonni:
@guardrex not worries, I've rebased out the conflict. The remaining code fence ones still have a bug till the next VSCode and CLI version of Markdownlint is released.

PR #11814, opened 2019-04-03
Fixed title.  Moved version clarification to intro

wadepickett:
Part of April SEO sweep.  High traffic topic with relatively high bounce rates.  Fixed title problem I created in the first place.  Moved the version clarfication that is mid page to the top intro.  This topic will be part of a lit of topics that are followed up on next month for SEO fixes to verify if that helped.


wadepickett:
@tdykstra, does this work better?

PR #11853, opened 2019-04-05
Adjust .NET Standard paragraph

thomasclaudiushuber:
A platform that supports .NET Standard 2.0 supports all previous versions: 1.0, 1.1, 1.2,1.3,1.4,1.5 and 1.6. So the original statement ".NET Standard 2.0 or higher is supported" is misleading, as Blazor won't support .NET Standard 2.1 without adjustments to the platform.

So my suggestion is to just change it to "Blazor supports .NET Standard 2.0".

In addition libraries can not only be shared between browser and server side, but between different platforms.

### DIFF
--- a/aspnetcore/client-side/spa/blazor/index.md
+++ b/aspnetcore/client-side/spa/blazor/index.md
@@ -86,7 +86,7 @@ For apps that require third-party JavaScript libraries and browser APIs, Blazor 
 
 ## Code sharing and .NET Standard
 
-Apps can reference and use existing [.NET Standard](/dotnet/standard/net-standard) libraries. .NET Standard is a formal specification of .NET APIs that are common across .NET implementations. Blazor implements .NET Standard 2.0. APIs that aren't applicable inside a web browser (for example, accessing the file system, opening a socket, threading, and other features) throw <xref:System.PlatformNotSupportedException>. .NET Standard 2.0 class libraries can be shared across different .NET platforms, like Blazor, .NET Framework, .NET Core, Xamarin, Mono, and Unity.
+Apps can reference and use existing [.NET Standard](/dotnet/standard/net-standard) libraries. .NET Standard is a formal specification of .NET APIs that are common across .NET implementations. Blazor implements .NET Standard 2.0. APIs that aren't applicable inside a web browser (for example, accessing the file system, opening a socket, threading, and other features) throw <xref:System.PlatformNotSupportedException>. .NET Standard class libraries can be shared across different .NET platforms, like Blazor, .NET Framework, .NET Core, Xamarin, Mono, and Unity.
 
 ## Optimization
 
--- a/aspnetcore/client-side/spa/blazor/index.md
+++ b/aspnetcore/client-side/spa/blazor/index.md
@@ -86,7 +86,7 @@ For apps that require third-party JavaScript libraries and browser APIs, Blazor 
 
 ## Code sharing and .NET Standard
 
-Apps can reference and use existing [.NET Standard](/dotnet/standard/net-standard) libraries. .NET Standard is a formal specification of .NET APIs that are common across .NET implementations. Blazor supports .NET Standard 2.0. APIs that aren't applicable inside a web browser (for example, accessing the file system, opening a socket, threading, and other features) throw <xref:System.PlatformNotSupportedException>. .NET Standard 2.0 class libraries can be shared across different .NET platforms, like Blazor, .NET Framework, .NET Core, Xamarin, Mono, and Unity.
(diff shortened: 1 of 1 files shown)


guardrex:
> .NET Standard 2.0 or higher is supported

Side note: "or higher" ... it should read "or later" btw. A tiny goof on my part in editing the content provided by engineering.

Usually, this language just means "[A target framework] of ..." However if engineering agrees that it's misleading in this case, then a change might be in order.

The "or higher" bit is perhaps only a point-in-time situation during preview. I wouldn't be surprised if "or later" was appropriate by the time it reaches RTM. We prefer "or later" wherever possible to harden the topic for future releases.

> .NET Standard class libraries

It just means as opposed to PCLs. I doubt that the version is necessary (or even a good idea given that we like to avoid versions where possible and based on my prior remark).

thomasclaudiushuber:
I'm sorry, but I think "or later" is not better than "or higher". :) 

With .NET Standard it's the opposite than with a platform. 

Normally you say: This app runs on .NET Framework 2.0 or later/higher
But you can't say that for .NET Standard, as it is not a platform, but a spec/contract: Blazor supports .NET Standard 2.0 or later/higher doesn't work. Instead you should say it supports .NET Standard 2.0. Everyone familiar with .NET Standard will know that 2.0 is a superset of the previous standards and that you can use it on different platforms like .NET Framework, .NET Core, Xamarin etc.

> The "or higher" bit is perhaps only a point-in-time situation during preview

No, definitely not. But it could be that the final release of Blazer will support .NET Standard 2.1, but it does this not by default just because it supports .NET Standard 2.0 right now.

>  We prefer "or later" wherever possible to harden the topic for future releases.

That's good for case where it makes sense. But as mentioned above, this works for frameworks, but not for specifications. You can't say "or later" for .NET Standard.

> I doubt that the version is necessary

Mentioning version 2.0 adds a lot of value, as a developer can see then on which platform they can use the library: https://github.com/dotnet/standard/blob/master/docs/versions.md

@terrajobst Maybe it makes sense to include Blazor or Mono WASM in the version table under https://github.com/dotnet/standard/blob/master/docs/versions.md, even though it's an experimental platform right now?

guardrex:
The phrase "or later" is required by our styles. We don't use "or higher," so that has to be updated.

U misunderstood me ... The language in these cases generally refers to the TFM specifically, not the platform(s) under it. TFMs release in order over time with a number that gets larger each time; so in that context, the language is ok. 

I'm describing the "general" case. In this case, most of your suggestions make sense ... U said "misleading" ... and that might be true. 

Engineering will let us know what they would like to do.

thomasclaudiushuber:
That sounds like a good plan Luke.

Thanks,
Thomas

terrajobst:
I agree with @thomasclaudiushuber. A platform implementing .NET Standard should never be described as "<platform> supports .NET Standard <version> or later" as that's factually wrong. The correct way would "<platform> supports .NET Standard <version> or earlier" but nobody can parse that. So I suggest "Blazor implements .NET Standard 2.0". That's inline with the guidance @mairaw has developed.

guardrex:
@thomasclaudiushuber ... do you want to react to that, and we don't need to sweat the "or later/higher" bit.

WRT "or later," "supports," and "version" with .NET Standard, I'll perform a search of our docs for similar language. We normally only mean to refer to the TFM (i.e., *targets* a given implementation) with "or later;" so in (hopefully) all other spots where .NET Standard is mentioned, we're good. However, I won't know for sure until I put an :eye: on the search results.

thomasclaudiushuber:
@guardrex Yep, I've just pushed another commit to change it to "Blazor implements .NET Standard 2.0"

aspnetcore/client-side/spa/blazor/index.md
```diff
@@ -86,7 +86,7 @@ For apps that require third-party JavaScript libraries and browser APIs, Blazor
 
 ## Code sharing and .NET Standard
 
-Apps can reference and use existing [.NET Standard](/dotnet/standard/net-standard) libraries. .NET Standard is a formal specification of .NET APIs that are common across .NET implementations. Blazor supports .NET Standard 2.0. APIs that aren't applicable inside a web browser (for example, accessing the file system, opening a socket, threading, and other features) throw <xref:System.PlatformNotSupportedException>. .NET Standard 2.0 class libraries can be shared across different .NET platforms, like Blazor, .NET Framework, .NET Core, Xamarin, Mono, and Unity.
+Apps can reference and use existing [.NET Standard](/dotnet/standard/net-standard) libraries. .NET Standard is a formal specification of .NET APIs that are common across .NET implementations. Blazor implements .NET Standard 2.0. APIs that aren't applicable inside a web browser (for example, accessing the file system, opening a socket, threading, and other features) throw <xref:System.PlatformNotSupportedException>. .NET Standard 2.0 class libraries can be shared across different .NET platforms, like Blazor, .NET Framework, .NET Core, Xamarin, Mono, and Unity.
```
guardrex: Thanks @thomasclaudiushuber 

WRT ...

> .NET Standard 2.0 class libraries ...

I still prefer the versionless approach. ".NET Standard" in this context is only here AFAIK to distinguish them from PCLs.

Let's let @danroth27's look and see if that's correct.

We have a hard time on this repo keeping up with explicit versions release-to-release, so we generally prefer *not* to use them whenever possible. 🏃🏃🏃😅

thomasclaudiushuber: I think it's totally fine to omit the version here. I added it, because .NET Standard 2.1 is not supported on .NET Framework. But I agree, one version per paragraph might be enough. :)

PR #11854, opened 2019-04-05
Paragraph duplication

Ponant:


tdykstra:
@guardrex hold off on merging this, I need to take another look, I think the issue may be just the comparison operator.

guardrex:
If u need me to put a final :eye: on it, I'll look when I get back from the :gym:. Hard to look on this crappy phone. lol

tdykstra:
@guardrex No hurry 😄

guardrex:
@tdykstra The topic isn't versioned but the 1st moniker range is for <2.1 with Razor Pages ... the whole topic should pertain to all versions (including 1.1&dagger;). It still needs a touch of work.

&dagger;We've been killing off the 1.1 coverage, so really the whole topic can go 2.1+ now ... then changes throughout will probably end up removing all (or most) versioning.

Would you like me to make a pass on this PR to do that? ... take the whole topic to 2.1+ and update all moniker ranges in the process?

Side note on a *super-nit* ... this is probably from while you were away in Azure docs land: We came to always placing the moniker ranges in descending order by version. If we don't update the whole topic to 2.1+ here and keep the moniker ranges, I recommend we stick with the pattern and go (in this case) ...

```
::: moniker range=">= aspnetcore-2.0"

... content that includes RP lingo ...

::: moniker-end

::: moniker range="< aspnetcore-2.0"

... content that only refers to controllers ...

::: moniker-end
```

Let me know if you'd like me to make that 2.1+ pass.

tdykstra:
I don't want to version the article metadata as 2.1+ because then it drops out of the ToC when 2.0 or 1.1 is selected.  But I'm OK with eliminating earlier than 2.1 moniker blocks.  

Meanwhile, this PR fixes the problem @Ponant noticed -- thanks @Ponant !

I'll create another issue and PR for 2.1+ifying the text.

guardrex:
> it drops out of the ToC when 2.0 or 1.1 is selected

By-design tho ... RA already gave an approval for that approach.

Whatever the case with that ... idc either way and it doesn't really matter either way until 1.1 goes away.

> PR fixes the problem @Ponant noticed

Dup para, yes.

> I'll create another issue and PR for 2.1+ifying the text.

... you mean to resolve the RP showing for 1.1 with that. Sounds good. :+1:

PR #11870, opened 2019-04-05
Describe how authorization handlers all run re success, failure and when unauthenticated.

### LINKED ISSUE #11609: Document the fact that custom AuthorizationHandlers will still execute even if *authentication* fails
When using requirement-based policies, it seems that all custom AuthorizationHandlers will still execute even if token validation fails.  (For example expired token.)

**Why does this matter?**
Some AuthorizationHandlers could perform some relatively time-intensive/expensive tasks.  If _authentication_ fails, there is a high chance users wouldn't want their custom AuthorizationHandlers to start evaluating requirements.

**Rough draft**
"Note: If token authentication fails (due to expired token, invalid cryptographic signature, etc), you may want to exit the `AuthorizationHandler.HandleAsync()` early.  You can do this by checking `context.Principal.Identity.IsAuthenticated`:"
```
        public Task HandleAsync(AuthorizationHandlerContext context)
        {
            if (!context.User.Identity.IsAuthenticated)
            {
                return Task.CompletedTask;
            }

            // Perform expensive authorization
        }
```

---
#### Document Details

* Content: [Policy-based authorization in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/policies?view=aspnetcore-2.2#feedback)
* Content Source: [aspnetcore/security/authorization/policies.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/security/authorization/policies.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

serpent5:
Fixes #11609.

* I built a sample project to test all of this and discovered that the explanation re all handlers being called even when a handler fails also applies to when a handler succeeds, so I updated the relevant section. I fully appreciate that this wasn't part of the linked issue so I'm more than happy to revert that if needed.
* I didn't include an example of how to check whether the user is authenticated as I concluded that this is a little out of scope for the topic.
* I updated the date on the topic itself as it was dating back to 2017.

### DIFF
--- a/aspnetcore/security/authorization/policies.md
+++ b/aspnetcore/security/authorization/policies.md
@@ -4,7 +4,7 @@ author: rick-anderson
 description: Learn how to create and use authorization policy handlers for enforcing authorization requirements in an ASP.NET Core app.
 ms.author: riande
 ms.custom: mvc
-ms.date: 11/21/2017
+ms.date: 04/05/2019
 uid: security/authorization/policies
 ---
 # Policy-based authorization in ASP.NET Core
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@serpent5  much appreciated.

PR #11894, opened 2019-04-08
Translation into Chinese

yilezhu:
Translation into Chinese


guardrex:
Hello @yilezhu ... We have a localization team that takes care of localizing the docs. The English version is the base doc, so this must remain in English.

yilezhu:
thanks

PR #11904, opened 2019-04-09
Object pool/ra

### LINKED ISSUE #10477: Document ObjectPool
ObjectPool exists in Microsoft.Extensions.ObjectPool but does not appear to be documented in terms of when and how it should be used.

It would be good to provide guidance and when using ObjectPool may be appropriate.

Describe the use of classes deriving from PooledObjectPolicy<T>, the creation of a pool from the provider, getting and returning items to the pool.

FYI - @rynowak and @scottaddie

Rick-Anderson:
Fixes #10477
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/performance/objectpool?view=aspnetcore-2.2&branch=pr-en-us-11904)


Rick-Anderson:
@stevejgordon  please review.

Rick-Anderson:
@guardrex  `ObjectPool<T>` AKA <xref:Microsoft.Extensions.ObjectPool.ObjectPool%601> has build error 	1 invalid cross reference(s) "<xref:Microsoft.Extensions.ObjectPool.ObjectPool*>".   
So does `Microsoft.Extensions.ObjectPool.ObjectPool<T>.Return(T)`

guardrex:
It looks right tho according to the API Browser. Here's the tag ...

```html
<meta name="ms.assetid" content="Microsoft.Extensions.ObjectPool.ObjectPool`1">
```

... but I'm not sure why "%601" should have resulted in "*" in the bad cross-ref notice. That's unusual/unexpected. Could give it a shot with the backtick and see if it likes like (although we said we wouldn't do that due to the messed up DocFX code style rendering) ...

```
<xref:Microsoft.Extensions.ObjectPool.ObjectPool`1>
```

The other one comes up as ...

```html
<meta name="ms.assetid" content="Microsoft.Extensions.ObjectPool.ObjectPool`1.Return*">
```

... so it might be happy with either ...

```
<xref:Microsoft.Extensions.ObjectPool.ObjectPool%601.Return*>
```

or

```
<xref:Microsoft.Extensions.ObjectPool.ObjectPool`1.Return*>
```

I can try to make some additional updates. Keep in mind tho that the XREF Gen thing is probably never going to be 100% accurate. I had to hack up a rules-based engine to deal with what the API Browser's web API returns. Since so many doc teams across the company would use something like this, I think the best bet is to request that that team produce something like this for us (and everyone else) that would be more accurate than what I can do.

Rick-Anderson:
@guardrex  the following works
<xref:Microsoft.Extensions.ObjectPool.ObjectPool`1> to get an object and <xref:Microsoft.Extensions.ObjectPool.ObjectPool`1.Return*>

guardrex:
We could go back to that format, but it screws up the markdown when editing.

WRT the one ending `Return` ... that's perhaps going to remain an edge case. There will prob always be a handful of oddball ones that the tool doesn't address.

I still pitch that this type of tool would be good for *all docs teams* to have access to. They need to build that ... not good for the company to be using a bunch of 🙈 *RexHacks* 🙈.

aspnetcore/performance/ObjectPool.md
```diff
@@ -0,0 +1,66 @@
...
+
+- Expensive to allocate/initialize.
+- Represent some limited resource.
+- Used predictably and frequently.
+
+For example, the ASP.NET Core framework uses the object pool in some places to reuse <xref:System.Text.StringBuilder> instances. `StringBuilder` allocates and manages its own buffers to hold character data. ASP.NET Core regularly uses `StringBuilder` to implement features, and reusing them provides a performance benefit.
+
+Object pooling doesn't always improve performance:
+
+- Unless the initialization cost of an object is hight, it's usually slower to get the object from the pool.
```
tdykstra:
```suggestion
- Unless the initialization cost of an object is high, it's usually slower to get the object from the pool.
```

stevejgordon:
@Rick-Anderson LGTM

PR #11944, opened 2019-04-11
Update AKV DP provider content

### LINKED ISSUE #11943: Confirm AKV permission required
Looks the app requires `unwrap key` permission.

https://github.com/aspnet/AspNetCore.Docs/issues/11941#issuecomment-482182673

This makes sense because that's exactly what it has to do ... unwrap the key!

I'll repro it here and then adjust the text as necessary.

cc: @brobichaud

---
#### Document Details

* ID: b63ab082-ad30-b00d-ecd1-ca88eee1659e
* Version Independent ID: 69818122-e47a-3fed-ea37-8009f66e2a5c
* Content: [Configure ASP.NET Core Data Protection](https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/configuration/overview?view=aspnetcore-2.2#protectkeyswithazurekeyvault)
* Content Source: [aspnetcore/security/data-protection/configuration/overview.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/security/data-protection/configuration/overview.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

guardrex:
Fixes #11943 

@pakrym That was a blown update earlier. I *assumed* it was just **Get** and **List**. Looks like it's actually **Unwrap Key** and **Wrap Key** permissions. I also add that the dev should create the key in the first place ... if they don't, it all ends in tears :boom::cry:.

Do we want one of the Azure KV cats to look at this? This PR is based on local testing, but I still might not be covering all of the bases.

Special thanks to @brobichaud :rocket: for reminding that **_assumptions_** can be a real pit of failure. 🙈


pakrym:
I missed that it was for DP, not configuration provider.

guardrex:
Yes ... I incorrectly assumed that they would be the same; but upon reflection, yeah ... this now makes perfect sense given what it's doing.

pakrym:
https://github.com/aspnet/AspNetCore/blob/master/src/DataProtection/AzureKeyVault/src/IKeyVaultWrappingClient.cs#L9

Confirmed. We only use wrap and unwrap in DP.

guardrex:
Cool ... thx again @brobichaud.

Now ... if you gentlemen will excuse me ........ I'm going to check out the *Help Wanted* ads for a position in **_typewriter maintenance_**. :smile:

PR #11950, opened 2019-04-11
Additional Claims topic+sample update

### LINKED ISSUE #5106: Document the claims actions
_Moved from https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/210._

While a sample can be found in the [aspnet/Security repository](https://github.com/aspnet/Security/blob/26d27d871b7992022c082dc207e3d126e1d9d278/samples/CookieSample/Startup.cs), there's currently no formal documentation for the ["claims actions" APIs offered by the OAuth-based providers](https://github.com/aspnet/Security/tree/26d27d871b7992022c082dc207e3d126e1d9d278/src/Microsoft.AspNetCore.Authentication.OAuth/Claims) in ASP.NET Core 2.x.

/cc @Tratcher @guardrex

### LINKED ISSUE #12084: Deprecated "plus.login" Scope
The `https://www.googleapis.com/auth/plus.login` scope has been deprecated and so the examples should be probably be updated accordingly.

---
#### Document Details

* Content: [Persist additional claims and tokens from external providers in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/social/additional-claims?view=aspnetcore-2.2#establish-the-authentication-scope)
* Content Source: [aspnetcore/security/authentication/social/additional-claims.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/security/authentication/social/additional-claims.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #5106 
Fixes #12084

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/security/authentication/social/additional-claims?view=aspnetcore-2.1&branch=pr-en-us-11950)

* Sample update to 2.2
  * ~Change scope to `userinfo.profile` from depreciated `plus.login`.~ Remove scope ... it's added by the framework. Thanks @serpent5! :rocket:
  * Changing over to `givenname` ~and `surname`~, locale, and picture because Google provides `gender` via a scope to a `Person`; whereas, the user's given name ~and surname~, locale, and picture are available directly from the `userinfo.profile` scope. It's simpler to stick with something that they provide via `userinfo.profile`.
  * Moving the instructions out of the topic and into the *README.md* of the sample.
* Topic gets some minor coverage added from the issue (new *Creating and adding claims* and *Removal of claim actions and claims* sections). If you'd like more, please provide additional guidance. What else from: https://github.com/aspnet/AspNetCore/tree/master/src/Security/Authentication/OAuth/src
* Fixed the IDE instructions in the social landing topic.


guardrex:
ping @HaoK

Tratcher:
@PinpointTownes

aspnetcore/security/authentication/social/additional-claims/samples/2.x/ClaimsSample/Startup.cs
```diff
@@ -30,26 +32,30 @@ public void ConfigureServices(IServiceCollection services)
...
+                // Register with User Secrets using:
+                // dotnet user-secrets set "Authentication:Google:ClientId" "{Client ID}"
+
+                // Provide the Google Client Secret
+                options.ClientSecret = "{Client Secret}";
+                // Register with User Secrets using:
+                // dotnet user-secrets set "Authentication:Google:ClientSecret" "{Client Secret}"
+
+                options.Scope.Add("https://www.googleapis.com/auth/userinfo.profile");
+                options.ClaimActions.MapJsonKey(ClaimTypes.GivenName, "givenname");
```
PinpointTownes: Interesting. Doesn't the Google handler already have claim actions to extract those 2 claims?

guardrex: Thanks ... it does. I'd like to use something else.

I'll be free shortly to address this. I'll ping u back in a bit.

PinpointTownes: Actually, it seems the claim names are slightly different: https://github.com/aspnet/AspNetCore/blob/28d44af91a249c9229ba1bbe543898bdbbe4a871/src/Security/Authentication/Google/src/GoogleOptions.cs#L30-L31

Does Google return both claims or was the custom claim action registered in the sample basically ineffective? 😕

guardrex: I need to research this further (and the latest bits more or less match the 2.2 release of that source ... just the Google Plus stuff if a bit different). Commenting out my `MapJsonKey` lines, I still get the names that I used (and not due to an old cookie) ...

```
User Claims
...
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname
Luke
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/surname
Latham
...
```

I'll be free in a bit to dig deeper.

guardrex: For one thing, I didn't set this up correctly ... but it doesn't work anyway yet when I change it to what I think it should be ...

[MapJsonKey](https://docs.microsoft.com/dotnet/api/microsoft.aspnetcore.authentication.claimactioncollectionmapextensions.mapjsonkey) should take care of creating the claim for me.

> Select a top level value from the json user data with the given key name and add it as a Claim. This no-ops if the key is not found or the value is empty.
> 
> ```csharp
> public static void MapJsonKey (
>     this Microsoft.AspNetCore.Authentication.OAuth.Claims.ClaimActionCollection collection, 
>     string claimType, 
>     string jsonKey);
> ```
>
> **claimType** The value to use for Claim.Type when creating a Claim.
>
> **jsonKey** The top level key to look for in the json user data.

Looking at the 2.2 bits :point_right: https://github.com/aspnet/AspNetCore/blob/release/2.2/src/Security/Authentication/Google/src/GoogleOptions.cs

... makes sense. I see ...

```csharp
ClaimActions.MapJsonKey(ClaimTypes.Name, "name");
ClaimActions.MapJsonKey(ClaimTypes.GivenName, "given_name");
ClaimActions.MapJsonKey(ClaimTypes.Surname, "family_name");
ClaimActions.MapJsonKey("urn:google:profile", "link");
ClaimActions.MapJsonKey(ClaimTypes.Email, "email");
```

If I do this, I would expect to see a `Upn` claim that has the value of `name` ...

```csharp
options.ClaimActions.MapJsonKey(ClaimTypes.Upn, "name");
```

That doesn't work. I'm not getting a claim.

I'll need to come back to this over the weekend and see what's what. Marking WIP for now.

guardrex:
**Improvements!** ... but open questions, too.

Updates:

* Let's keep the given (first name) bit. I made a small mistake in that I didn't create the given name claim for the app user thinking that it would happen for me auto-magically ... no ... no ... of course not. That's resolved.
* I found a couple that we can use from profile data: The user's `picture` and `locale` are readily available. I even add them to the homepage when the claims exist so that the reader can see :eyes: the values after they authenticate.

## Open questions/concerns:

### Engineering sample differences from the topic sample

This sample app was derived from an engineering sample ages ago. There are deltas here between the topic's sample app and the engineering sample. They diverged at some point. The [current engineering sample](https://github.com/aspnet/AspNetCore/blob/release/2.2/src/Security/Authentication/samples/SocialSample/Startup.cs) ...

* Performs an `.AddOAuth("Google-AccessToken", "Google AccessToken only", o =>` bit that I'm not familiar with.
* Calls `AuthorizationEndpoint` describing it as a "hack" to always get a refresh token.
* Sets up an `OAuthEvents` for an `OnRemoteFailure`, which the sample app doesn't do.
* Gets the Google image from a subkey [`o.ClaimActions.MapJsonSubKey("urn:google:image", "image", "url");` (i.e., `image`)], but is that really a thing? I found the user's Google image from `picture`. I didn't see the subkey that the engineering sample is trying to get ... well ... maybe I didn't look hard enough at the response.
* For some reason, it removes given name (`o.ClaimActions.Remove(ClaimTypes.GivenName);`), and it's not apparent why.

:question: here is: Do we need to match any of this in the topic's sample app? If so, which bits?

### Linking the engineering sample

The engineering sample also has Twitter, MS, and GH. It's probably *overkill* for the topic sample. I'd like to link it. However, we don't want to maintain the release branch link to the current engineering sample. The docs team is small and 🏃😅. I propose to link to `master` and **_note_** that the dev should look at the release branch that matches the version they're working with. :+1: That would be a lot better for us doc :cat2::cat2::cat2:.

### Failure to get birthday

I wired up the sample app to create a `birthdays` claim, but I never received the data from Google (it worked in their dev tools but not for the app). I think it has to do with the way that I registered the app in the first place. Should I continue down that road, or is what the sample is showing enough?

Tratcher:
Don't worry too much about the engineering sample, it tweaks a lot of knobs just to see what happens.

> Performs an .AddOAuth("Google-AccessToken", "Google AccessToken only", o => bit that I'm not familiar with.

This shows how to use just the base class to do a minimal OAuth flow, no user info. No need to doc it, nobody has shown an interest in that scenario.

> Calls AuthorizationEndpoint describing it as a "hack" to always get a refresh token.

Tacking `prompt=consent` onto AuthorizationEndpoint is an easy way to modify the flow without hooking any events. This parameter forces a fresh consent prompt which is needed before google will give you a refresh token. The whole refresh token scenario is pretty complicated and varies a lot between providers. Some day we'll make it a priority.

> Sets up an OAuthEvents for an OnRemoteFailure, which the sample app doesn't do.

It seemed like a good idea at the time. Don't worry about it for now.

> Gets the Google image from a subkey [o.ClaimActions.MapJsonSubKey("urn:google:image", "image", "url"); (i.e., image)], but is that really a thing? I found the user's Google image from picture. I didn't see the subkey that the engineering sample is trying to get ... well ... maybe I didn't look hard enough at the response.

This might be outdated data from the G+ days. I'd have to try it.

> For some reason, it removes given name (o.ClaimActions.Remove(ClaimTypes.GivenName);), and it's not apparent why.

That was to show how to remove a default claim map and filter out claims that you don't want to store long term. This is a good thing to cover. The more claims you keep the larger your cookie gets, and big cookies can cause a number of problems depending on the browser.

> Linking the engineering sample

Perhaps a note that there are more complex samples available with the source code, Yes, link them to master and note they should check the branch relevant to their release.

guardrex:
I try a NOTE and a link to the engineering sample on the last commit.

Looks like the rest of it we don't need to sweat on this pass.

We can't merge this for a week, so there's no rush. However, I do need at least one :shipit: from engineering to pull the trigger later if everything looks good here.

btw ---- We don't doc anything on `SessionStore`/`ITicketStore`. There are blog posts out there ... some dated (e.g., broken links to old engineering code and pre-Auth overhaul at 2.0). Anyway, we don't cover it other than a passing mention and the API Browser. If that needs to change at some point, let me know, and I'll open an issue.

guardrex:
@Tratcher Does my last commit address remaining points?

aspnetcore/security/authentication/social/additional-claims/samples/2.x/ClaimsSample/Startup.cs
```diff
@@ -30,27 +27,31 @@ public void ConfigureServices(IServiceCollection services)
...
+                // Register with User Secrets using:
+                // dotnet user-secrets set "Authentication:Google:ClientSecret" "{Client Secret}"
+
+                options.ClaimActions.MapJsonKey("urn:google:picture", "picture", "url");
+                options.ClaimActions.MapJsonKey("urn:google:locale", "locale", "string");
                 options.SaveTokens = true;
+
                 options.Events.OnCreatingTicket = ctx =>
                 {
                     List<AuthenticationToken> tokens = ctx.Properties.GetTokens()
```
Tratcher: This example shouldn't be casting. Calling ToList() would be better.

aspnetcore/security/authentication/social/additional-claims.md
```diff
@@ -42,98 +42,108 @@ Specify the list of permissions to retrieve from the provider by specifying the
...
 
-The sample app saves the access token in:
-
-* `OnPostConfirmationAsync` &ndash; Executes for new user registration.
-* `OnGetCallbackAsync` &ndash; Executes when a previously registered user signs into the app.
-
-*Account/ExternalLogin.cshtml.cs*:
-
-[!code-csharp[](additional-claims/samples/2.x/AdditionalClaimsSample/Pages/Account/ExternalLogin.cshtml.cs?name=snippet_OnPostConfirmationAsync&highlight=34-35)]
+The sample app saves the access token in `OnPostConfirmationAsync` (new user registration) and `OnGetCallbackAsync` (previously registered user) in *Account/ExternalLogin.cshtml.cs*:
```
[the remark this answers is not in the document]
guardrex: AFAICT (and I might have the link ref wrong since there are a few of these), this is built into the template.

https://github.com/aspnet/AspNetCore/blob/4e271ac124591052d4fe989e0ea464e8127aaded/src/Identity/UI/src/Areas/Identity/Pages/V4/Account/ExternalLogin.cshtml.cs#L145

aspnetcore/security/authentication/social/additional-claims/samples/2.x/ClaimsSample/Startup.cs
guardrex: From the engineering sample ...

https://github.com/aspnet/AspNetCore/blob/658b37d2bd3b0da2c07e25b53fa5c97ce8d656d3/src/Security/samples/Identity.ExternalClaims/Startup.cs#L49

@HaoK?

guardrex:
Anything else? ..... *Going once* ...... *Going twice* ...........

guardrex:
@Rick-Anderson This one just needs a final look.

aspnetcore/security/authentication/social/additional-claims.md
```diff
@@ -42,98 +42,105 @@ Specify the list of permissions to retrieve from the provider by specifying the
...
 
-In the sample app, `OnPostConfirmationAsync` (*Account/ExternalLogin.cshtml.cs*) establishes a <xref:System.Security.Claims.ClaimTypes.Gender> claim for the signed in `ApplicationUser`:
+[!code-csharp[](additional-claims/samples/2.x/ClaimsSample/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs?name=snippet_OnPostConfirmationAsync&highlight=35-51)]
 
-[!code-csharp[](additional-claims/samples/2.x/AdditionalClaimsSample/Pages/Account/ExternalLogin.cshtml.cs?name=snippet_OnPostConfirmationAsync&highlight=30-31)]
+> [!NOTE]
+> By default, claims are stored in the authentication cookie. Unless the app assigns a custom <xref:Microsoft.AspNetCore.Authentication.Cookies.ITicketStore> to the Cookie Authentication Middleware's <xref:Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationOptions.SessionStore> to mitigate issues with serializing large identities into cookies, limit the number of user claims held by the app to only those required for request processing. The cookie can grow in size to a point where the app fails to process requests properly because the browser detects one or more cookie headers that are too long or the overall size of the request is too large.
```
Rick-Anderson: Can you rewrite this to get sentence length under control?
I'd remove the **NOTE** BLING.

Auth cookies can grow too large and browsers will reject them? Not clear how it fails from your RO sentence. But I'd start out with the salient  point, Auth cookies can grow to large. To mitigate the large cookie problem:

* assigns a custom <xref:Microsoft.AspNetCore.Authentication.Cookies.ITicketStore> to the Cookie Authentication Middleware's <xref:Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationOptions.SessionStore> 
etc.

That note is hard to read and impossible to MT.

guardrex:
@Rick-Anderson Last commit look good?

Rick-Anderson:
@guardrex  nice work!

PR #11955, opened 2019-04-12
C80832: Adding two spaces at the end of a line

v-makoud:


v-makoud:
#sign-off

PR #11972, opened 2019-04-14
Missing class name

drahuja6:
Missing class name found.


Rick-Anderson:
This PR looks correct to me.  cc @tdykstra 
How did it compile the way it is now?

tdykstra:
@Rick-Anderson It doesn't compile with or without the change due to NuGet version conflicts. However, this error was caused by #9850 which intended to delete AsNoTracking but accidentally deleted Contact along with it.

PR #11975, opened 2019-04-15
chore: Update allowed HTML elements

nschonni:
li and ul are sometimes needed in Markdown tables


tdykstra:
Let's wait a bit on this, while I check if HTML lists within MD tables is considered acceptable for docs.

guardrex:
It's the only way btw ... tried MD first and it ended in :cry: tears. This was asked-and-answered earlier: HTML is *dissuaded* but not outlawed. In those cases where I used lists, they worked out well ... look good ... organized that content well.

tdykstra:
Turns out we have a lot longer list of whitelisted HTML tags, and we should make MarkdownLint match it:

The following HTML elements, organized by category, are whitelisted:

Type | Elements  
---------|---------
Headings     |  h1, h2, h3, h4, h5, h6
Prose     |  p, div, blockquote
Formatted     |   pre
Inline     |     b, i, strong, em, tt, code, ins, del, sup, sub, kbd, samp, q, var
Lists     |   ol, ul, li, dl, dt, dd
Tables     | table, thead, tbody, tfoot, tr, td, th
Breaks     |   br, hr

guardrex:
**_Sweet!_** ... only when needed, of course. MD is nice but a bit too constraining sometimes imo.

nschonni:
The alternate would be to turn on the rule and use `<!-- markdownlint-disable MD033 -->` in the areas/files that it's required

tdykstra:
I would vote for either disabling the rule entirely or adding the whitelisted elements.

scottaddie:
I vote for adding the whitelisted elements.

nschonni:
> I would vote for either disabling the rule entirely or adding the whitelisted elements.

This is one of the rules were it's not a good idea to disable it. EX:
`List<T>` renders to  "List<T>" and swallows the `<T>` without escaping

tdykstra:
I used the other PR only for disabling selected rules, so I"ll add a commit to this PR for the rest of the allowed HTML elements.

tdykstra:
@nschonni Thanks for getting the work on this rule started and for all your contributions to making our Markdown more consistent!

PR #11982, opened 2019-04-15
Turn off some markdownlint rules

tdykstra:
Per team discussion on 4/12, turn off selected rules that we anticipate would be more nuisance than helpful.


scottaddie:
@tdykstra Can you also apply these changes to the config in the ASP.NET 4.x docs repo?

PR #12017, opened 2019-04-17
Update link

guardrex:
The link here is throwing.

@bradygaster I don't quite see where this one is at in https://docs.microsoft.com/javascript/api/%40aspnet/signalr/?view=signalr-js-latest ... if you point me to the right one, I'll take care of the link (I'll attempt to XREF link it in here).

### DIFF
--- a/aspnetcore/signalr/javascript-client.md
+++ b/aspnetcore/signalr/javascript-client.md
@@ -113,7 +113,7 @@ const connection = new signalR.HubConnectionBuilder()
 
 Without any parameters, `withAutomaticReconnect()` configures the client to wait 0, 2, 10, and 30 seconds respectively before trying each reconnect attempt, stopping after four failed attempts.
 
-Before starting any reconnect attempts, the `HubConnection` will transition to the `Reconnecting` <xref:Microsoft.AspNetCore.SignalR.Client.HubConnectionState> state and fire its `onreconnecting` callbacks instead of transitioning to the `Disconnected` state and triggering its `onclose` callbacks like a `HubConnection` without automatic reconnect configured. This provides an opportunity to warn users that the connection has been lost and to disable UI elements.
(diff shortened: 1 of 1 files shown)


aspnetcore/signalr/javascript-client.md
```diff
@@ -113,7 +113,7 @@ const connection = new signalR.HubConnectionBuilder()
 
 Without any parameters, `withAutomaticReconnect()` configures the client to wait 0, 2, 10, and 30 seconds respectively before trying each reconnect attempt, stopping after four failed attempts.
 
-Before starting any reconnect attempts, the `HubConnection` will transition to the `Reconnecting` [HubConnectionState](javascript/api/%40aspnet/signalr/hubconnectionstate) state and fire its `onreconnecting` callbacks instead of transitioning to the `Disconnected` state and triggering its `onclose` callbacks like a `HubConnection` without automatic reconnect configured. This provides an opportunity to warn users that the connection has been lost and to disable UI elements.
+Before starting any reconnect attempts, the `HubConnection` will transition to the `Reconnecting` <xref:Microsoft.AspNetCore.SignalR.Client.HubConnectionState> state and fire its `onreconnecting` callbacks instead of transitioning to the `Disconnected` state and triggering its `onclose` callbacks like a `HubConnection` without automatic reconnect configured. This provides an opportunity to warn users that the connection has been lost and to disable UI elements.
```
anurse: Is this the right syntax for referencing **JavaScript** API docs? That may be why the 🧀 appears to has moved :).

Also, I think this was done as part of the reconnect changes in preview 4. Perhaps the API docs aren't up to date? I don't see the `HubConnectionState` enum anywhere in the latest JS docs.

guardrex: No ... it's not correct. I'll work on getting an XREF link in there once I see what we're linking.

> reconnect changes in preview 4

That explains it.

Why don't I just code fence it for now (no link), and that will allow us to clean up the build warnning.

guardrex:
I code fenced it for now. If the format is ok, we can fix this with a proper API link when it becomes available.

PR #12027, opened 2019-04-17
Using Microsoft.AspNetCore.Http.StatusCodes instead of numbers is much readable

tahazayed:
using Microsoft.AspNetCore.Http.StatusCodes instead of numbers is much readable


scottaddie:
Thank you for your contribution, @tahazayed! Your changes will go live within the next 24 hours.

PR #12028, opened 2019-04-17
Add web API MS Learn module to ToC

scottaddie:
Note: I'm trying to mirror the ToC structure under the **Tutorials** node. We'll have a **Data access** child node once the EF Core Learn module publishes. There will also be a **Web apps** child node when Wade's Razor Pages Learn module publishes.

### DIFF
--- a/aspnetcore/toc.yml
+++ b/aspnetcore/toc.yml
@@ -20,6 +20,13 @@
       uid: aspnetcore-2.0
     - name: What's new in 1.1
       uid: aspnetcore-1.1
+- name: Tutorials (interactive)
+  items:
+    - name: Web API apps
+      items:
+        - name: Create a web API
+          displayName: tutorial
+          href: /learn/modules/build-web-api-net-core/
 - name: Tutorials
   items:
     - name: Web apps
--- a/aspnetcore/toc.yml
+++ b/aspnetcore/toc.yml
@@ -20,13 +20,6 @@
       uid: aspnetcore-2.0
     - name: What's new in 1.1
       uid: aspnetcore-1.1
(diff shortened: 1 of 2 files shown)


aspnetcore/toc.yml
```diff
@@ -20,6 +20,13 @@
       uid: aspnetcore-2.0
     - name: What's new in 1.1
       uid: aspnetcore-1.1
+- name: Tutorials (interactive)
+  items:
+    - name: Web API apps
+      items:
+        - name: Create a web API
+          displayName: tutorial
```
scottaddie: @tdykstra Is this the correct value for `displayName`? I wasn't sure if we needed to differentiate Learn modules from MVC-style tutorials in some way.

tdykstra: A couple of thoughts:
* Consider putting "Microsoft Learn" in the parens instead of interactive.  The site is branded that way, it's in the URL, and it's clearer that it's taking you out of this doc set.
* Given that it is an exit from this doc set and should still be seen as an alternative rather than the primary offering, I would be inclined to put the MS Learn Tutorials node after the main Tutorials node rather than before it.
* "tutorial" for displayName sems fine.  if you go with (Microsoft Learn), you could make displayName "tutorial, interactive"  All displayName does is expand the number of search terms that cause the node to show up when something is entered in the filter box.
* Rather than have an extra level, I'd put the link to the tutorial on the "Web API apps" node. When we get another Web API MS Learn tutorial we can split them out to another level.

scottaddie:
@tdykstra Please take another look.

scottaddie:
@tdykstra Thank you for your feedback.

PR #12034, opened 2019-04-18
Update 22-to-30.md

Rick-Anderson:


guardrex:
I should make one more pass this morning. I'll do it first thing when I get back from the 💪.

PR #12035, opened 2019-04-18
Improve discoverability of secrets management docs

scottaddie:
Moves the Secret Manager and Azure Key Vault config provider docs under a new **Secrets management** ToC node. This change should make it easier to locate those docs.

### DIFF
--- a/aspnetcore/toc.yml
+++ b/aspnetcore/toc.yml
@@ -939,9 +939,10 @@
     - name: Secrets management
       items:
         - name: Protect secrets in development
-          displayName: password
+          displayName: password, connection string, API key
           uid: security/app-secrets
         - name: Azure Key Vault Configuration Provider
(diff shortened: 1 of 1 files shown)


aspnetcore/toc.yml
```diff
@@ -936,11 +936,13 @@
...
               uid: security/data-protection/compatibility/replacing-machinekey
-    - name: Protect secrets in development
-      displayName: password
-      uid: security/app-secrets
-    - name: Azure Key Vault Configuration Provider
-      uid: security/key-vault-configuration
+    - name: Secrets management
+      items:
+        - name: Protect secrets in development
+          displayName: password
```
tdykstra: consider adding "connection string", plus any other terms you can think of for the kinds of secrets being managed.  Then put the same displayName text on the Key Vault node.

PR #12054, opened 2019-04-19
To disable validation in startup for Razor Pages

Ponant:
See https://github.com/aspnet/AspNetCore/issues/7576


Ponant:
Ignore this as it seems there is some ongoing bug

PR #12055, opened 2019-04-19
Disable client side validation in Razor Pages

Ponant:
Related to the creation of Startup3.cs PR.
See https://github.com/aspnet/AspNetCore/issues/7576


Ponant:
Ignore this as it seems there is some ongoing bug

PR #12067, opened 2019-04-19
Fixing grammar issue

DominicJ2:
Several spots say `Razor Pages is` when it should be `are`

### DIFF
--- a/aspnetcore/razor-pages/index.md
+++ b/aspnetcore/razor-pages/index.md
@@ -11,7 +11,7 @@ uid: razor-pages/index
 
 By [Rick Anderson](https://twitter.com/RickAndMSFT) and [Ryan Nowak](https://github.com/rynowak)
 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@DominicJ2 I believe is **is** correct
cc @tdykstra  who's the grammar expert. Razor Pages is not plural

replace Razor Pages with MVC in the sentences

tdykstra:
Yes, it's being used as the name of a framework in these contacts, so "is" is correct.  In some contexts that's not the case, for example, "an app may have multiple Razor pages" where "pages" is lower case and refers to the actual pages.

Singular verbs with plural-appearing product names is common -- Azure Web Apps is another example. See how [this Azure DevOps page ](https://azure.microsoft.com/en-us/solutions/architecture/azure-devops-continuous-integration-and-continuous-deployment-for-azure-web-apps/)uses "is" after Web Apps.

PR #12070, opened 2019-04-19
fix: MD023/heading-start-left/header-start-left

nschonni:


dlld:
salut

aspnetcore/blazor/get-started.md
```diff
@@ -24,8 +24,9 @@ In a few steps, get started with Blazor:
 
 1. Follow the guidance for your choice of tooling:
 
+   <!-- markdownlint-disable MD023 -->
```
guardrex: Hi @nschonni ... Please revert the changes to this file. I don't want the markdownlint disable rule in here because they clutter the content and will confuse community contributors. We have enough cryptic rules and guidelines on the repo as it is.

@scottaddie Should the linter issue here be raised elsewhere? I think markdownlint disable rules sprinkled in all over the place is going to end in tears :cry:. IMHO, either the linter should be fixed or ignored.

aspnetcore/signalr/streaming.md
```diff
@@ -52,7 +52,7 @@ Streaming hub methods can return `IAsyncEnumerable<T>` in addition to `ChannelRe
 
 ::: moniker-end
 
-The following sample shows the basics of streaming data to the client using Channels. Whenever an object is written to the <xref:System.Threading.Channels.ChannelWriter`1>, the object is immediately sent to the client. At the end, the `ChannelWriter` is completed to tell the client the stream is closed.
+The following sample shows the basics of streaming data to the client using Channels. Whenever an object is written to the <xref:System.Threading.Channels.ChannelWriter%601>, the object is immediately sent to the client. At the end, the `ChannelWriter` is completed to tell the client the stream is closed.
```
guardrex: IIRC this was done on purpose when the doc was written. I think the `%60` broke. I'll check the build tho. Maybe 2nd time is the charm! 🍀:smile:

guardrex: *Oh, cool!* It passed this time. Leave it ... we were going to do that anyway until we hit a snag on the original PR (either it was this one, or I might be thinking of another one where we tried this and it :boom: on us).

aspnetcore/blazor/get-started.md
guardrex: Another factor on this topic, too ... I'm about to merge a bunch of changes to it. It's best if I pick up this work on that PR. Otherwise, one or the other of us is going to create a :boom:.

scottaddie: I'm in agreement that we don't want to disable rules inline. All rules configuration should be handled in the config file.

nschonni: I rebased out this change.
I do think it's better to suppress rules inline when there are only a few exceptions throughout the codebase

PR #12106, opened 2019-04-22
Blazor Overview updates

guardrex:
Addresses #10717 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/index?view=aspnetcore-3.0&branch=pr-en-us-12106)

* Some of these can be taken as *i-dears* (but I think I'm right :smile: lol).
* I found the original example to be a bit of a head-scratcher, so I replace that here with something a bit more drawn-out and clear: We probably want to show just how HTML- and C#-ish the code is in components, including nesting composition. This covers a surprising number of bases in an example even a bit simpler than what we do in the *Components* topic.
* Per our earlier updates, I'm moving client-side coverage ahead of server-side coverage. I think we should do this everywhere, and I'll confirm as I go that we are.


aspnetcore/blazor/index.md
```diff
@@ -4,120 +4,142 @@ author: guardrex
...
 uid: blazor/index
 ---
 # Introduction to Blazor
 
 By [Daniel Roth](https://github.com/danroth27) and [Luke Latham](https://github.com/guardrex)
 
-Welcome to Blazor!
+*Welcome to Blazor!*
 
 Build interactive client-side web UI with .NET:
```
danroth27:
```suggestion
Blazor is a framework for building interactive client-side web UI with .NET. 

With Blazor you can:
```

aspnetcore/blazor/index.md
```diff
@@ -4,120 +4,142 @@ author: guardrex
...
 
-* Build rich interactive UIs using C# instead of JavaScript.
+* Create rich interactive UIs using C# instead of JavaScript.
 * Share server-side and client-side app logic written with .NET.
 * Render the UI as HTML and CSS for wide browser support, including mobile browsers.
 
-Blazor supports core scenarios required by most apps:
+Blazor supports app development scenarios required by most apps:
```
danroth27: This Blazor feature list feels a bit awkward here. Some of these features are self-explanatory, but many of them are not.

I think it's important that this first section covers: 1. What is Blazor? and 2. Why should I care?

I think we have (1) covered, but I think we could do a better job with (2). We have some text that talks about the advantage of using Blazor in the client-side section below. How about we move it up here in place of this feature list?

aspnetcore/blazor/index.md
```diff
@@ -4,120 +4,142 @@ author: guardrex
...
 * Routing
 * Dependency injection
 * Layouts
 * Templates
 * Cascading values
+* Parameters
 
 ## Components
 
 A *component* in Blazor is an element of UI, such as a page, dialog, or data entry form. Components handle user events and define flexible UI rendering logic. Components can be nested and reused.
```
danroth27:
```suggestion
Blazor apps are based on *components*. A component in Blazor is an element of UI, such as a page, dialog, or data entry form. Components handle user events and define flexible UI rendering logic. Components can be nested and reused.
```

aspnetcore/blazor/index.md
```diff
@@ -4,120 +4,142 @@ author: guardrex
...
 
-* Handles sending UI events from the browser to the server.
-* Applies UI updates sent by the server back to the browser after running the components.
+The dialog is rendered when the parent (*Index.razor*) is accessed in a browser:
 
-The connection used by Blazor server-side to communicate with the browser is also used to handle JavaScript interop calls.
+![Dialog component rendered in the browser](index/_static/dialog.png)
 
-![Blazor server-side runs .NET code on the server and interacts with the Document Object Model on the client over a SignalR connection](index/_static/blazor-server-side.png)
+When this component is used in the app, [Visual Studio](https://visualstudio.microsoft.com/vs/) IntelliSense speeds development with syntax and parameter completion.
```
danroth27:
```suggestion
When this component is used in the app, IntelliSense in [Visual Studio](https://visualstudio.microsoft.com/vs/) and [Visual Studio Code](https://code.visualstudio.com) speeds development with syntax and parameter completion.
```

aspnetcore/blazor/index.md
```diff
@@ -4,120 +4,142 @@ author: guardrex
...
-For more information, see <xref:blazor/hosting-models#server-side>.
+Components render into an in-memory representation of the browser DOM called a *render tree* that's used to update the UI in a flexible and efficient way.
 
 ## Blazor client-side
 
-Blazor client-side is a single-page app framework for building interactive client-side Web apps with .NET. Blazor client-side uses open web standards without plugins or code transpilation. Blazor client-side works in all modern web browsers, including mobile browsers.
+Blazor client-side is a single-page app framework for building interactive client-side web apps with .NET. Blazor client-side uses open web standards without plugins or code transpilation and works in all modern web browsers, including mobile browsers.
 
-Using .NET in the browser for client-side web development offers many advantages:
+Using .NET in the browser for client-side web development offers the following advantages:
```
danroth27: This is the list I think we should move into the opening section

PR #12120, opened 2019-04-23
sample uses new beatpulse, so updated

dzmitry-lahoda:
fixed doc according sample and new lib used in sample


guardrex:
@dzmitry-lahoda Thanks ... BeatPulse is listed right at the top of that repo. There are a few more references that I updated on the last commit. I figure that the simplest thing to do is just use the one name, `AspNetCore.Diagnostics.HealthChecks` everywhere. Check my updates and see if you agree with the changes.

guardrex:
I put the "BeatPulse" naming and link back, but I only added it to the NOTE where it talks about the lib not being supported by Microsoft. That way, it keeps their branding intact here.

guardrex:
@unaizorrilla Would you like to review this PR?

guardrex:
No response, so let's go ahead with this. Thanks @dzmitry-lahoda! :rocket:

unaizorrilla:
Hi @guardrex

I’m sorry but I’m on different timezone!!

I read the PR and look good!

guardrex:
Sorry about that ... I should've known. If you think it needs more work, let me know.

PR #12160, opened 2019-04-25
Updated app configuration notes

vvimjam:
Updated notes regarding app configuration extensions & their relative nuget packages which are not part of the standard Microsoft.Extensions.Configuration package.

### DIFF
--- a/aspnetcore/fundamentals/host/generic-host.md
+++ b/aspnetcore/fundamentals/host/generic-host.md
@@ -5,7 +5,7 @@ description: Learn about ASP.NET Core's Generic Host, which is responsible for a
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 03/31/2019
+ms.date: 04/25/2019
 uid: fundamentals/host/generic-host
 ---
 # .NET Generic Host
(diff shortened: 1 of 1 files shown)


guardrex:
I think it's fine to mention this. This should be cross-linked to the Configuration topic, too. I'll make some updates to the text, including some grammatical touches.

... and not necessarily that the Configuration topic does a super great job at calling out the packages. It does so via the sample app, and we have been placing direct package references into the text lately across the doc set. That work is on-going. I'll open an issue to address it in the Configuration topic when I get some free time.

btw- It's best to open an issue to discuss changes like this before creating a PR because the we usually need to discuss/plan the coverage a bit. We generally only take PRs without issues if it's for a patch (e.g., a spelling error, a small coding error, etc.).

vvimjam:
@guardrex Thank you for the feedback, I will look into the api browser. My bad i read the Contribute to the ASP.NET Core documentation and thought since this was just a small edit this would not fall under the complex submission or a new topic which requires opening an issue like you said. Guess i should reread it.

guardrex:
Not so much "new topic" ... just a more complicated change than a spelling error or a small coding bug. No worries ... this is a small update (and a good idea btw :+1:).

I'll get on the Configuration topic sometime in May. We've been working over time to better surface packages and namespaces, and that doc definitely needs some help in that regard.

PR #12178, opened 2019-04-26
C86844: Adding two spaces after line 173 to avoid display issues on localized pages

v-maudel:
Description:  missing spaces break formatting
  
Many thanks in advance.
cc: @rick-anderson,


v-maudel:
Thanks @Rick-Anderson

PR #12192, opened 2019-04-29
Update from microsoft/* to mcr.microsoft.com/*.

olegburov:
Switch .NET Core Images from the legacy location at Docker Hub Registry (`microsoft/*`) to a new location at Microsoft Container Registry (`mcr.microsoft.com/*`).


olegburov:
@guardrex Sounds good 👍

Issue #9025, opened 2018-10-15
Getting Error in SeedData.cs file.


Srusti-Thakkar:
Hello,

I have implemented your code step by step. But I am getting null user in below code, but as per expectation  It should return User at the line " var user = await userManager.FindByIdAsync(uid);" because user is already created. What is the issue? Any thing missed by me? I am able to create user and it return Id successfully.

```
private static async Task<IdentityResult> EnsureRole(IServiceProvider serviceProvider,
                                                                      string uid, string role)
        {
            IdentityResult IR = null;
            var roleManager = serviceProvider.GetService<RoleManager<IdentityRole>>();

            if (roleManager == null)
            {
                throw new Exception("roleManager null");
            }

            if (!await roleManager.RoleExistsAsync(role))
            {
                IR = await roleManager.CreateAsync(new IdentityRole(role));
            }

            var userManager = serviceProvider.GetService<UserManager<IdentityUser>>();

            var user = await userManager.FindByIdAsync(uid);

            IR = await userManager.AddToRoleAsync(user, role);

            return IR;

        }
```

#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b66f8ffa-699c-8b11-0cab-805f5953805f
* Version Independent ID: 7c151ed5-fd78-84fd-a08a-88912990c4ca
* Content: [Create an ASP.NET Core app with user data protected by authorization](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/secure-data?view=aspnetcore-2.1)
* Content Source: [aspnetcore/security/authorization/secure-data.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/security/authorization/secure-data.md)
* Product: **aspnet-core**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
It's almost always a weak password.

Srusti-Thakkar:
@Rick-Anderson . Thank you for the help. I have changed password and it's working as expected.

sturlath:
This tripped me up also. Why not add 
```csharp
if(user == null)
{
	throw new Exception("The password is probably not strong enough!");
}
```

Should I do a PR?

Rick-Anderson:
@sturlath  a PR would be great.

PR #12196, opened 2019-04-29
Update SeedData.cs

sturlath:
Give the user of the example explicit information (in the form of an exception) if the password used to create the user was not strong enough.  #9025


Rick-Anderson:
@sturlath  thanks!

PR #12206, opened 2019-04-30
Odata-expand

FIVIL:
This PR contains examples of  OData `$expand`.

/cc @Rick-Anderson 

My other fork ran into some problems with this new branch so it kept including all those other commits to my PR, I created a new fork it should work fine and only include my commits


guardrex:
Hello @FIVIL ... without performing a full review, two quick observations ...

* Several of your files don't end on a new line. Although it doesn't really hurt anything (and a number of *wwwroot* assets don't do it), we provide them where we can to stop GH from complaining.

  Capture

* WRT the grades assigned: I know that you didn't mean anything by it, but I suggest assigning all of the grades to the student records in the range of A-C (not D or F). We're a bit sensitive about negative connotations inadvertently applied to ethnicity, race, culture, and religion. It's best to go as neutral as possible. Besides, I used to teach, and I hate to see anyone get an 'F' in a course. :cry:

FIVIL:
@guardrex 
Tnx for your feedback,
ill make some changes

PR #12219, opened 2019-04-30
Add .NET universal header

rowanmiller:
This will put the same header on the ASP.NET Core docs that is used on the [.NET website](https://dotnet.microsoft.com/), [main .NET docs](https://docs.microsoft.com/en-us/dotnet/), and [EF docs](https://docs.microsoft.com/en-us/ef/).

Here are examples of this setting being used in the other doc sets:
* Main .NET docs https://github.com/dotnet/docs/blob/master/docfx.json#L108
* EF docs https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/docfx.json#L81


rowanmiller:
Corresponding PR on ASP.NET 4.x docs https://github.com/aspnet/AspNetDocs/pull/157

scottaddie:
Thanks, @rowanmiller! We'll merge this on Monday, when the change freeze has been removed.

PR #12225, opened 2019-05-01
Fix language specification in one code sample

MiKom:
There was a typo "chstml" instead of "cshtml" that caused the code section to not do proper syntax coloring


scottaddie:
Thank you for the PR, @MiKom! We'll merge and publish this change on Monday, when our change freeze has been removed.

guardrex:
Thanks from me, too, @MiKom ... that was my goof. My Microsoft Ergonomic Keyboard must be trying to get back at me for not using all of the *fancy function keys* that it offers! :smile:

guardrex:
We can send this to staging now.

PR #12249, opened 2019-05-02
v3 preview 4 update to core vs 4.x compare

wadepickett:
Updated this comparison doc to account for ASP.NET Core 3.0 Preview 4.    The compare doc uses an include called benefits.md which was also updated.  Minor updates overall.

Current published version of this doc:
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/choose-aspnet-framework?view=aspnetcore-3.0

### DIFF
--- a/aspnetcore/fundamentals/choose-aspnet-framework.md
+++ b/aspnetcore/fundamentals/choose-aspnet-framework.md
@@ -4,7 +4,7 @@ author: rick-anderson
 description: Explains ASP.NET Core vs. ASP.NET 4.x and how to choose between them.
 ms.author: riande
 ms.custom: "mvc, seodec18"
-ms.date: 09/11/2018
+ms.date: 05/02/2019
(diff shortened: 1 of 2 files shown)


wadepickett:
Blazor or Blazor (Preview):  Checking with the PM, thanks Tom.

wadepickett:
@danroth27, at this point in time should we include (Preview) after "Blazor"?

danroth27:
Blazor only appears in the 3.0 preview doc set, so I think preview is implied and doesn't need to be stated in prose.

wadepickett:
@tdykstra, PM says "Blazor" is the most appropriate here.  Good catch though, thanks for asking.

wadepickett:
@tdykstra, I thought this PR went through.  Looking back I realized it is still showing build/license checks did not complete for some reason.  Also no option to merge.  What am I missing?  Thanks for looking.

PR #12257, opened 2019-05-03
Update MovieGenreViewModel to contain only properties in stead of fields

LocalJoost:
It's odd to have two PUBLIC fields and two public properties. Those should all be properties I think


scottaddie:
Thank you for the contribution, @LocalJoost! We're currently in a content freeze. This change will be merged on Monday.

PR #12277, opened 2019-05-04
Use correct gRPC client project name

randy-r:
When adding gRPC packages the project to add them to should be GrpcGreeterClient.csproj.


Rick-Anderson:
@randy-r  thanks

PR #12282, opened 2019-05-04
Update debug.md

tronsoft:
In my case the shortcut Shift+Alt+D does not work. I am using Chrome Version 74.0.3729.131 (Official Build) (64-bit). Maybe the shortcut changed?


tronsoft:
Sorry, the shift + windows key + d shortcut also works but is not the right one. So this pull request can be closed.

PR #12344, opened 2019-05-09
File uploads topic

### LINKED ISSUE #10214: File Upoads - Master issue
image

Moved from #5331

https://docs.microsoft.com/aspnet/core/mvc/models/file-uploads

- [x] Provide 2.0 code samples (address #5217)
- [x] Address #3989
- [ ] Address #4777 **MOVED TO https://github.com/aspnet/AspNetCore.Docs/issues/12285**
- [x] Address #4964
- [x] Address #6247
- [x] Address #6349
- [x] Address #6927 
- [x] Address #7702
- [x] Address #7742
- [x] Address #8039
- [x] Address #8100
- [x] Address #8298
- [x] Address #8797
- [x] Address #9116
- [x] Address #9327
- [x] Address #9473
- [x] Address #10161
- [ ] Address #10537 **wontfix** ... sample+topic are going RP
- [x] Address #11898
- [x] Address #11949
- [ ] Address #12130 &ndash; Won't fix because we're not going to cover reading from `Request.Form` directly on the updates. If more devs ask about the approach, we can reconsider it later.
- [x] Address #12204
- [x] Address #6609
- [x] Address #6610
- [x] Address #6962
- [x] Address #11822

---
#### Document Details

* Version Independent ID: c6787f0f-3e9a-6efb-0e6e-5ccb957d3f4a
* Content: [File uploads in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/file-uploads?view=aspnetcore-2.1)
* Content Source: [aspnetcore/mvc/models/file-uploads.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/mvc/models/file-uploads.md)
* Product: **aspnet-core**
* GitHub Login: @ardalis
* Microsoft Alias: **riande**

guardrex:
Fixes #10214

* [Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/mvc/models/file-uploads?view=aspnetcore-2.2&branch=pr-en-us-12344)
* [Sample app (guardrex repo)](https://github.com/guardrex/FileUploadsSampleApp)

@blowdart The new security bits *_are quite extensive_* 🐄🔔 compared to the current topic. There's a strong security lead-in section and a validation section further down the topic that explains the sample app approaches. If you look at the sample in my repo (linked above), you can see the bits. Feel free to comment directly on that sample app in my repo if that's easier, and I'll roll your recommendations into the sample here. I recommend it because this PR has *_A LOT_* of file churn.


guardrex:
@Rutix No rush; but when you get a minute, can you take a look at https://github.com/aspnet/AspNetCore.Docs/pull/12344/commits/f5e2b641e2fb999305a4b0e3cda548ee75eefd89 and see how that looks?

Rutix:
@guardrex Looks good. I was pondering if we need to make a mention of `DisableRequestSizeLimit` but considering security ect. I think we only want to encourage use of `RequestSizeLimit`. I started reviewing the PR but it's kinda hard because of the file churn 😅

guardrex:
`DisableRequestSizeLimit` ... yes, let's see if @blowdart wants that covered.

Another one of great concern is removal of the IIS Request Filtering Module. On the engineering issue, John said ...

> There is a limit with the Request Filter in IIS which has a limit of UInt32.Max (~4GB) but this filter can be removed.

The module also deals with URLs ... segments/file extensions, query strings, hides private locations, and limits verbs. We probably **don't** want to say 'remove it if you need to.' That's dangerous.

blowdart:
**I think we only want to encourage use of RequestSizeLimit.**

Yes please :)

Same with not encouraging removing the request filter.

guardrex:
@blowdart Ur still digging out from Build? We could proceed with the docs review if you want to circle around later for the engineering review.

blowdart:
Is there a preview, because  584 changed files is frankly impossible to review. I don't even know where to start

aspnetcore/mvc/models/file-uploads/samples/2.x/SampleApp/Utilities/FileHelpers.cs
```diff
@@ -0,0 +1,268 @@
...
+            if (property != null)
+            {
+                if (property.GetCustomAttribute(typeof(DisplayAttribute)) is
+                    DisplayAttribute displayAttribute)
+                {
+                    fieldDisplayName = $"{displayAttribute.Name} ";
+                }
+            }
+
+            // Don't trust the file name sent by the client. Use Linq to
```
blowdart: STOP DOING THIS.

Don't attempt to sanitize, use a random filename. *ALWAYS*

guardrex: I'll change it in both PRs.

aspnetcore/mvc/models/file-uploads/samples/2.x/SampleApp/Utilities/FileHelpers.cs
```diff
@@ -0,0 +1,268 @@
...
+            // HTML encoded. Another option is to use
+            // Path.GetRandomFileName to generate a safe random
+            // file name.
+            var invalidFileNameChars = Path.GetInvalidFileNameChars();
+            var trustedFileName = WebUtility.HtmlEncode(
+                    invalidFileNameChars.Aggregate(
+                        formFile.FileName, (current, c) =>
+                            current.Replace(c, '_')));
+
+            // Check the file length. This check doesn't catch files that only have
```
blowdart: What are you trying to do here? This has become more than a sample and started to veer into security sensitive guidance, which feels to me to be very out of scope for docs.

aspnetcore/mvc/models/file-uploads/samples/2.x/SampleApp/Utilities/FileHelpers.cs
```diff
@@ -0,0 +1,268 @@
...
+                modelState.AddModelError("File",
+                    "The upload failed. Please contact the Help Desk " +
+                    $" for support. Error: {ex.HResult}");
+                // Log the exception
+            }
+
+            return new byte[0];
+        }
+
+        private static bool IsValidFileExtensionAndSignature(string fileName, Stream data, string[] permittedExtensions)
```
blowdart: What are you trying to do here? This has become more than a sample and started to veer into security sensitive guidance, which feels to me to be very out of scope for docs.

guardrex:
@blowdart 

Yes ... it's in the opening comment, and you can look at the sample in my repo if that's easier.

blowdart:
Ah yes. What is the purpose here? I started look at code and you've gone beyond a sample, into attempting file type checks and other things. This seems dangerous, because, well, it's a sample, it's going to have bypasses, be out of date and so on. While you have a caveat that you're not virus scanning you're presenting things like file extension validation as solution, which they're not. Each one of those functions needs highlighting they're samples, not full solutions Would you feel happy someone cut and pasted those and thinking they're secure?

As for IFormFile, that needs someone from MVC to review. I'm concerned that they don't have anything for this and you need helpers for what should be a common scenario.

aspnetcore/mvc/models/file-uploads/samples/2.x/SampleApp/Utilities/FileHelpers.cs
guardrex: We provided several of these guardrails when we put up the RP file uploads topic on https://github.com/aspnet/AspNetCore.Docs/pull/4265. You didn't object, so I assumed that you'd be ok with these approaches over here. Actually, I thought that you'd love the attention to security here. Sorry ... I should've asked you on the issue. The outline for these changes was approved, but only by the docs team.

Do you want me to strip all of the security of out the topic and sample app?

Equally, strip out the bit that I added to the other PR on custom model binding?

Follow-up on the other PR ... it has the dev uploading directly to the app's *wwwroot* folder. Do you want to leave that, too? If so, then that other PR (and the issue) can simply be closed.

guardrex:
@blowdart I understand, but I was looking at it from the opposite angle. I thought that if we had a demonstration of guardrails that match the advice that the topic lists, then it would be better than nothing at all. It also provided an opportunity to show dev how to trigger `ModelState` errors during validation checks that would go back to the UI for display, either via model binding or via JSON. My (admittedly *wild guess*) is that devs will cut-and-paste file upload code with little or no security measures at all. That, to me, seems *more dangerous*, but I yield to you, of course, on that point.

I wish I would've pinged you earlier, but you seemed ok with the guardrails approach on the RP topic. I don't think the docs team thought differently, as they approved the outline for these updates.

The next commit will ...

* Strip out all of the security.
* Use a generated file name (but we probably should sanitize and keep the original for human-readable displays).

On the other PR (custom model binding) ...

* Same deal with the file name.
* Do you want to leave that topic uploading to *wwwroot*? If so, I can just close the issue+PR. If not, I'll leave that bit in place that uploads the files to a dedicated location.

I'll take care of all of this today ... and truly since I'm stripping out just a couple of classes ... this won't take long to fix.

blowdart:
I think there's a compromise, cut it back. So, for example, check for JPGS, with extension and file signature, but not anything else. So it demonstrates methods without looking like a fully operational death star, err, sample.

blowdart:
Oh missed the upload directory, no, outside of wwwroot, into an uploads directory or some such.

guardrex:
Gotcha .... Ok, cool. That sounds great.

Rutix:
I think the compromise is a good idea. We already found out earlier that the signature checking didn't match all files correctly and I see the point of view that it goes beyond a sample.

guardrex:
I'll consult with @blowdart shortly. I have some serious concerns about what we show devs here. The only drop from current coverage is file length, which is trivial to check.

Stand-by ........ I'm work'in now 🏃 ...... will post shortly on the work.

aspnetcore/mvc/models/file-uploads/samples/2.x/SampleApp/Utilities/MultipartRequestHelper.cs
```diff
@@ -2,15 +2,16 @@
 using System.IO;
 using Microsoft.Net.Http.Headers;
 
-namespace FileUploadSample
+namespace SampleApp.Utilities
 {
     public static class MultipartRequestHelper
     {
         // Content-Type: multipart/form-data; boundary="----WebKitFormBoundarymx2fSWqWSd0OxQqq"
         // The spec says 70 characters is a reasonable limit.
```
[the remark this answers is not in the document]
guardrex: Yes ... I'll roll this into the commit. Thanks @Rutix.

guardrex:
@blowdart Made progress today. Sorry for the delay ... I had a surgery this afternoon, and it felt like they used an 🐘 tranquilizer on me!

A couple of follow up questions:

* Are we still keeping the file name **_for display purposes only_** and only in the database storage case (small files that are saved to a database field). If so, do you think that HTML-encoding the file name prior to storing it in the dB is enough (i.e., removing invalid file name chars is pretty much irrelevant if the file name is HTML encoded)?

  An alternative to using the file name from the file upload **_at all_** is to accept the human-readable *for-display-only* file name from a form field ... something they type in and submit ... with normal client- and server-side form validation. That would completely drop use of the submitted file upload file name from the sample app. I can go that way if you like.
* Turns out that the only check beyond file extension/file signature (that code that I modified from the [Azure folks security topic](https://docs.microsoft.com/azure/security/azure-security-threat-modeling-tool-input-validation#controls-users)) is file length. It's just a basic check. Can I leave that check in here? However if you say 'yes, ok,' I'm still subject to the charge of creating a 🎃  *Rex Franken App*:tm: 👻 ... even if my intentions are good.

I think I've resolved your key concerns on file name + I'll react to your responses on my questions. I am using the generated *very safe* file names from `Path.GetRandomFileName` now for all physically-stored files. It looks good.

I'll get the next commit up after I hear back from you.

blowdart:
Yes you could keep the name for display purposes, but you store it as is, and you encode it on output.
And yea, length check works.

However, 

TAKE A DAY OR TWO AND SIT ON YOUR DRUGGED UP BUTT. 

I don't want to see anything more on this till Monday.

guardrex:
lol .... I'm ok.

The doc said, 'Sitting quietly and writing docs is ok ... **_BUT_** ... going to the 💪 gym? *No, sir!* Don't you dare raise your blood pressure like that until next Monday.'

guardrex:
@blowdart Looking good from here:

Last few commits:

* All physical storage based on randomly-generated file name `Path.GetRandomFileName`. Also, `Path.GetTempFileName` is retained in the topic to surface that approach (to temp folder with full system generated path).
* dB storage stores the untrusted file name and HTML-encodes it for display. It's covered in both code comments and in the topic section on file name security. The dB field and model prop are super clear as to what's being held ... the name is `AppFile.UntrustedName`.

When these security aspects are all good, I'll ask Ryan if he (or who) can review.

blowdart:
Works for me. Let Ryan free :D

guardrex:
> Works for me.

_WOOT!_ My attorney thanks u in advance for that. The jury will be appeased ... my acquittal is assured! :smile:

guardrex:
Sry @ryanbrandenburg .... GH AI didn't read my mind on that ping.

@rynowak Do you have 🕐 to look? If not, can you suggest an engineer to review this?

guardrex:
@pranavkm Hi -- Who would be best to review the file upload updates here? ... and note that due to the file churn here, I put the sample up on my repo for easy access https://github.com/guardrex/FileUploadsSampleApp and review.

btw --- quick ?

It looks like the buffering limit changed at some point from 64 KB to **30 KB** ... is that correct (the topic still says 64, so I'll need to update that if it's actually 30)?

* https://github.com/aspnet/AspNetCore/blob/92cae6faab9cd414eccf77eab687c32c7a6a0f1b/src/Http/Http/src/Internal/BufferingHelper.cs#L12
* https://github.com/aspnet/AspNetCore/blob/093df67c06297c20edb422fe6d3a555008e152a9/src/Mvc/Mvc.Formatters.Xml/src/XmlDataContractSerializerInputFormatter.cs#L25

Rutix:
@guardrex that one has always been 30 KB that's the default in case you don't specify the memoryThreshold. As far as I know in the FormFeature the default form options values are passed. Which is 64KB for the memoryThreshold

guardrex:
Thanks @Rutix ... I see. Cool. No changes needed then.

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,654 @@
...
 
-## Uploading large files with streaming
+### Upload large files with streaming
 
-If the size or frequency of file uploads is causing resource problems for the app, consider streaming the file upload rather than buffering it in its entirety, as the model binding approach shown above does. While using `IFormFile` and model binding is a much simpler solution, streaming requires a number of steps to implement properly.
+The following example demonstrates using JavaScript to stream a file to a controller action. The file's antiforgery token is generated using a custom filter attribute and passed to the client HTTP headers instead of in the request body. Because the action method processes the uploaded data directly, model binding is disabled by another custom filter. Within the action, the form's contents are read using a `MultipartReader`, which reads each individual `MultipartSection`s, processing the file or storing the contents as appropriate. Once all of the sections are read, the action performs its own model binding.
```
Rutix: The filter only disables form model binding not all binding. It is still possible to info passed in the url to parameters.
```suggestion
The following example demonstrates using JavaScript to stream a file to a controller action. The file's antiforgery token is generated using a custom filter attribute and passed to the client HTTP headers instead of in the request body. Because the action method processes the uploaded data directly, form model binding is disabled by another custom filter. Within the action, the form's contents are read using a `MultipartReader`, which reads each individual `MultipartSection`s, processing the file or storing the contents as appropriate. Once all of the sections are read, the action performs its own model binding.
```

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,654 @@
...
+
+The `DisableFormValueModelBindingAttribute` is used to disable model binding:
+
+[!code-csharp[](file-uploads/samples/2.x/SampleApp/Filters/ModelBinding.cs?name=snippet_DisableFormValueModelBindingAttribute)]
+
+`GenerateAntiforgeryTokenCookieAttribute` and `DisableFormValueModelBindingAttribute` are applied as filters to the page application models of `/StreamedSingleFileUploadDb` and `/StreamedSingleFileUploadPhysical` in `Startup.ConfigureServices` using [Razor Pages conventions](xref:razor-pages/razor-pages-conventions):
+
+[!code-csharp[](file-uploads/samples/2.x/SampleApp/Startup.cs?name=snippet_AddMvc&highlight=8-11,17-20)]
+
+Since model binding is disabled, the action method doesn't accept parameters. The action method works directly with the `Request` property. A `MultipartReader` is used to read each section. Key/value data is stored in a `KeyValueAccumulator`. After all of the sections are read, the contents of the `KeyValueAccumulator` are used to bind the form data to a model type.
```
Rutix: See comment above. Might want to make clear only the form parameters wont bind.

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,654 @@
...
+>
+> The examples provided don't take into account security considerations. See the [Security considerations](#security-considerations) and [Validation](#validation) sections and the sample app for information and examples of file upload validation.
+
+When uploading files using model binding and <xref:Microsoft.AspNetCore.Http.IFormFile>, the action method can accept:
+
+* A single <xref:Microsoft.AspNetCore.Http.IFormFile>.
+* An <xref:System.Collections.IEnumerable>\<<xref:Microsoft.AspNetCore.Http.IFormFile>> or [List](xref:System.Collections.Generic.List`1)\<<xref:Microsoft.AspNetCore.Http.IFormFile>> representing several files. 
 
-![File upload form](file-uploads/_static/upload-form.png)
+The following example:
```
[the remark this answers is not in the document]
guardrex: I think this is ok. This is fairly close to the sample app's *Buffered Multiple File Upload - Physical* ... showing a trusted file name, a loop over form files, and `CopyToAsync`. The sample app's code has a bit too much going on (i.e., `FileHelpers` + `ModelState`) for a quick, focused in-text example. However, I'll add a sentence that says something to the effect of 'for another example, see the sample app.'

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,654 @@
...
+
+## File upload scenarios
+
+Two general approaches for uploading files are buffering and streaming.
+
+**Buffering**
+
+The entire file is read into an <xref:Microsoft.AspNetCore.Http.IFormFile>, which is a C# representation of the file used to process or save the file.
+
+Any single buffered file exceeding 64 KB is moved from memory to a temp file on disk.
```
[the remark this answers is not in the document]
guardrex: Management generally doesn't like them and has ordered that they mostly not be used. However, I think you're correct here for that reason ... and another important reason: It can lead to deadly resource issues on the server and the app doesn't report anything ... the server just blows up in your face. I'll put it back into a NOTE.

guardrex:
Thanks @Rutix ... see if the last commit covers your feedback.

Rutix:
LGTM :)

guardrex:
@Rick-Anderson This one wasn't reached (yet). Is it still coming up for PU review?

guardrex:
@pranavkm Who would be best to review this one? It's been here a while (May 8), and the PU label didn't attract an engineer to look on the recent round of PU reviews.

To facilitate review of the PR, which has a lot of file churn, the sample is in my repo :point_right: https://github.com/guardrex/FileUploadsSampleApp

Then, the only file to :eyes: here is the *file-uploads.md* topic file on the diff (2nd file down).

@blowdart Has already looked and :+1: on the security aspects.

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/mvc/models/file-uploads?view=aspnetcore-2.2&branch=pr-en-us-12344)

Rick-Anderson:
@pranavkm  can you review  https://github.com/guardrex/FileUploadsSampleApp

guardrex:
@scottaddie I fixed up some merge conflicts here, and I noticed some strange warnings ...

https://opbuildstorageprod.blob.core.windows.net/report/2019%5C7%5C15%5Cff1b9f40-00b1-57d9-58da-03fc9ac53777%5CPullRequest%5C201907152038158450-12344%5Cworkflow_report.html?sv=2016-05-31&sr=b&sig=JUIjl0oe67nNZYES3B2YBqXSb56yk5oD8n%2FoXFcsaQk%3D&st=2019-07-15T20%3A35%3A00Z&se=2019-08-15T20%3A40%3A00Z&sp=r

guardrex:
Well ... not too strange. 06/31/2017 **isn't a REAL date!** :smile: That month ended on the 30th.

Rick-Anderson:
@guardrex  This branch has conflicts that must be resolved

guardrex:
@Rick-Anderson Fixed

Rutix:
@guardrex any idea when this is being merged into master :D? I also think we might have to look at the 3.0 changes or are you going to do that seperately?

guardrex:
*Definitely separately* :smile: This major overhaul is hard enough to review as it is. I'll do the 3.0 work immediately after this goes in.

For the engineering review, I was told that they might look at this after 3.0 GA ... yesterday.

guardrex:
@mkArtakMSFT I'm offering **_10 cases of :beer:_** (!!!) to any engineer there who will review this PR. :smile: lol

Seriously tho ... the readers are still suffering this horrid topic. This PR fixes a *TON* of community feedback. See :point_right: https://github.com/aspnet/AspNetCore.Docs/issues/10214#issuecomment-491382346.

We already have Barry's sign-off on the security aspects. We need an engineer to review the sample app. I put it up over here :point_right: https://github.com/guardrex/FileUploadsSampleApp for easy inspection given all of the churn on this PR. We'll take care of the textual review on the docs side.

pranavkm:
@guardrex I can have a look at this tomorrow. Does that sound good?

guardrex:
Thank you @pranavkm! That's great.

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,656 @@
...
+## File upload scenarios
+
+Two general approaches for uploading files are buffering and streaming.
+
+**Buffering**
+
+The entire file is read into an <xref:Microsoft.AspNetCore.Http.IFormFile>, which is a C# representation of the file used to process or save the file.
+
+> [!NOTE]
+> Any single buffered file exceeding 64 KB is moved from memory to a temp file on disk.
```
pranavkm: Should we call out that the temp file is opaque to the user and entirely managed by ASP.NET Core? Also is this useful for users to know? Were users concerned about buffering large files in memory?

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,656 @@
...
 </form>
+
+@section Scripts {
+    @{await Html.RenderPartialAsync("_ValidationScriptsPartial");}
+
+    <script>
+        "use strict";
+
+        function AJAXSubmit (oFormElement) {
+          var oReq = new XMLHttpRequest();
```
pranavkm: Should this use [fetch](https://caniuse.com/#search=fetch)? It's likely that if you needed to support a browser that did not support it, you're going to use a polyfill library \ jQuery.

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,656 @@
...
+            'Result: ' + this.status + ' ' + this.statusText;
+          };
+          oReq.open("post", oFormElement.action);
+          oReq.send(new FormData(oFormElement));
+        }
+    </script>
+}
+```
+
+The `<form>` element attribute `asp-antiforgery` is set to `true` so that the form receives an antiforgery request token. The token isn't automatically provided when a form's `action` is set explicitly. Setting the attribute to `true` overrides the default behavior. When the token is present in the form, it's automatically sent with the POST to the server to validate the request. For more information, see <xref:security/anti-request-forgery>.
```
pranavkm: You could use `asp-page="/ThisPage"` in the form or `var url = "@Url.Page()";` in JS to avoid explaining about antiforgery

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,656 @@
...
+
+> [!WARNING]
+> Don't rely on or trust the `FileName` property of <xref:Microsoft.AspNetCore.Http.IFormFile> without validation. The `FileName` property should only be used for display purposes and only after HTML encoding the value.
+>
+> The examples provided don't take into account security considerations. See the [Security considerations](#security-considerations) and [Validation](#validation) sections and the sample app for information and examples of file upload validation.
+
+When uploading files using model binding and <xref:Microsoft.AspNetCore.Http.IFormFile>, the action method can accept:
+
+* A single <xref:Microsoft.AspNetCore.Http.IFormFile>.
+* An <xref:System.Collections.IEnumerable>\<<xref:Microsoft.AspNetCore.Http.IFormFile>> or [List](xref:System.Collections.Generic.List`1)\<<xref:Microsoft.AspNetCore.Http.IFormFile>> representing several files.
```
pranavkm: Also an `IFormFileCollection`

This should also call out that MVC will match form files by name when binding i.e. the `name` in `<input type="file" name="formFile">` needs to match the parameter \ property being bound just like everything else.

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,656 @@
...
+    // Full path to file, including the file name. You can also use
+    // Path.GetRandomFileName to generate a file name without a path
+    // if the app supplies the path (for example, via configuration).
+    var filePath = Path.GetTempFileName();
+
+    foreach (var formFile in files)
+    {
+        if (formFile.Length > 0)
+        {
+            using (var stream = new FileStream(filePath, FileMode.Create))
```
pranavkm: `File.Create(filePath)`?

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,656 @@
...
-    string FileName { get; }
-    Stream OpenReadStream();
-    void CopyTo(Stream target);
-    Task CopyToAsync(Stream target, CancellationToken cancellationToken = null);
+    long size = files.Sum(f => f.Length);
+
+    // Full path to file, including the file name. You can also use
+    // Path.GetRandomFileName to generate a file name without a path
+    // if the app supplies the path (for example, via configuration).
+    var filePath = Path.GetTempFileName();
```
pranavkm: Should this say `Path.Combine(MyAppsUploadFileFolder, Path.GetRandomFileName())`? Maybe it will have the user think about a value for for the destination before they copy it in their app?

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,656 @@
...
+Files uploaded using the <xref:Microsoft.AspNetCore.Http.IFormFile> technique are buffered in memory or on disk on the server before processing. Inside the action method, the <xref:Microsoft.AspNetCore.Http.IFormFile> contents are accessible as a <xref:System.IO.Stream>. In addition to the local file system, files can be saved to a network share or to a file storage service, such as [Azure Blob storage](/azure/visual-studio/vs-storage-aspnet5-getting-started-blobs).
 
-[!INCLUDE [GetTempFileName](../../includes/GetTempFileName.md)]
+For another example that loops over multiple file uploads and uses safe file names and <xref:System.IO.FileStream.CopyToAsync*>, see the sample app.
 
-[!code-csharp[](file-uploads/sample/FileUploadSample/Controllers/UploadFilesController.cs?name=snippet1)]
+> [!WARNING]
+> [Path.GetTempFileName](xref:System.IO.Path.GetTempFileName*) throws an <xref:System.IO.IOException> if more than 65,535 files are created without deleting previous temporary files. The limit of 65,535 files is a per-server limit. For more information on this limit on Windows OS, see the remarks in the following topics:
```
pranavkm: Also avoids this warning 👍

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,656 @@
...
-// POST: /Account/Register
-[HttpPost]
-[AllowAnonymous]
-[ValidateAntiForgeryToken]
-public async Task<IActionResult> Register(RegisterViewModel model)
+public async Task<IActionResult> OnPostUploadAsync()
 {
-    ViewData["ReturnUrl"] = returnUrl;
-    if  (ModelState.IsValid)
+    using (var memoryStream = new MemoryStream())
```
pranavkm: `using var memoryStream = new MemoryStream()` for 💯 points.

pranavkm: Also, should this one have a line that validates file size before doing the `ToArray()`? Even in the ordinary case it's a DOS vector since the server is allocating 2x the memory for every 1x file the client uploads.

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,656 @@
...
-Copy the `IFormFile` to a stream and save it to the byte array:
+When the form is POSTed to the server, copy the <xref:Microsoft.AspNetCore.Http.IFormFile> to a stream and save it as a byte array in the database:
 
 ```csharp
-// POST: /Account/Register
-[HttpPost]
-[AllowAnonymous]
-[ValidateAntiForgeryToken]
-public async Task<IActionResult> Register(RegisterViewModel model)
+public async Task<IActionResult> OnPostUploadAsync()
```
pranavkm: This action doesn't return an `IActionResult`.

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,656 @@
...
+
+The `DisableFormValueModelBindingAttribute` is used to disable model binding:
+
+[!code-csharp[](file-uploads/samples/2.x/SampleApp/Filters/ModelBinding.cs?name=snippet_DisableFormValueModelBindingAttribute)]
+
+`GenerateAntiforgeryTokenCookieAttribute` and `DisableFormValueModelBindingAttribute` are applied as filters to the page application models of `/StreamedSingleFileUploadDb` and `/StreamedSingleFileUploadPhysical` in `Startup.ConfigureServices` using [Razor Pages conventions](xref:razor-pages/razor-pages-conventions):
+
+[!code-csharp[](file-uploads/samples/2.x/SampleApp/Startup.cs?name=snippet_AddMvc&highlight=8-11,17-20)]
+
+Since model binding is disabled, the action method doesn't accept parameters, and form parameters don't bind. The action method works directly with the `Request` property. A `MultipartReader` is used to read each section. Key/value data is stored in a `KeyValueAccumulator`. After the multipart sections are read, the contents of the `KeyValueAccumulator` are used to bind the form data to a model type.
```
pranavkm: Since model binding ~~is disabled~~ does not read the form, parameters that are bound from the form do not bind.

pranavkm: Query, route, header, etc should still work

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,656 @@
...
 
-The initial action loads the form and saves an antiforgery token in a cookie (via the `GenerateAntiforgeryTokenCookieForAjax` attribute):
+```csharp
+private Dictionary<string, byte[]> _fileSignature = 
+    new Dictionary<string, byte[]>
+{
+    { ".docx", new byte[] { 0x50, 0x4B, 0x03, 0x04 } },
+};
+
+using (var reader = new BinaryReader(uploadedFileData))
```
pranavkm: You could use a `StreamReader`, no need for `BinaryReader`. 

```C#
using var reader = new StreamReader();
var headerBytes = reader.ReadBytes(signatures.Values.Max(m => m.Length));

foreach (var (key, value) in signatures)
{
   if (headerBytes.Slice(0, value.Length).SequenceEquals(value))
   {
      // Success!
   }
}
```

pranavkm: ^ dont know if this compiles

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,656 @@
...
+
+```cshtml
+@using System.Net
+
+@foreach (var file in Model.DatabaseFiles) {
+    // Don't display the untrusted file name in the UI. HTML-encode the value.
+    var fileName = WebUtility.HtmlEncode(file.UntrustedName);
+    <tr>
+        <td>
+            @Html.DisplayFor(modelItem => fileName)
```
pranavkm: `@file.UntrustedName` should suffice. That would HTML encode it.

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,656 @@
...
-The `DisableFormValueModelBinding` attribute, shown below, is used to disable model binding for the `Upload` action method.
+* When using an `<input>` element, the `name` attribute is set to the value `battlePlans`:
 
-[!code-csharp[](file-uploads/sample/FileUploadSample/Filters/DisableFormValueModelBindingAttribute.cs?name=snippet1)]
+  ```html
+  <input type="file" name="battlePlans" multiple>
+  ```
 
-Since model binding is disabled, the `Upload` action method doesn't accept parameters. It works directly with the `Request` property of `ControllerBase`. A `MultipartReader` is used to read each section. The file is saved with a GUID filename and the key/value data is stored in a `KeyValueAccumulator`. Once all sections have been read, the contents of the `KeyValueAccumulator` are used to bind the form data to a model type.
+* When using `FormData` in JavaScript, the name is set to the value `battlePlans`:
```
pranavkm: 👍

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,656 @@
...
+                            // Set the limit to 256 MB
+                            MultipartBodyLengthLimit = 268435456
+                        });
+        })
+    .SetCompatibilityVersion(CompatibilityVersion.Version_2_2);
 ```
-HTTP 404.13 - Not Found
-The request filtering module is configured to deny a request that exceeds the request content length.
+
+In a Razor pages app or an MVC app, apply the filter to the page handler class or action method:
```
pranavkm: to the page ~~handler class~~ model

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,656 @@
...
+            // Handle requests up to 50 MB
+            options.Limits.MaxRequestBodySize = 52428800;
+        });
+```
+
+::: moniker-end
+
+<xref:Microsoft.AspNetCore.Mvc.RequestSizeLimitAttribute> is used to set the [MaxRequestBodySize](xref:fundamentals/servers/kestrel#maximum-request-body-size) for a single page or action.
+
+In a Razor Pages app, apply the filter with a [convention](xref:razor-pages/razor-pages-conventions) in `Startup.ConfigureServices`:
```
pranavkm: Super untested, but in 3.0, you should be able to do

`@attribute [RequestSizeLimitAttribute(52428800)]` in the cshtml

aspnetcore/mvc/models/file-uploads/samples/2.x/SampleApp/Controllers/StreamingController.cs
```diff
@@ -0,0 +1,325 @@
...
+                // Log error
+
+                return BadRequest(ModelState);
+            }
+
+            var boundary = MultipartRequestHelper.GetBoundary(
+                MediaTypeHeaderValue.Parse(Request.ContentType),
+                _defaultFormOptions.MultipartBoundaryLengthLimit);
+            var reader = new MultipartReader(boundary, HttpContext.Request.Body);
+            var section = await reader.ReadNextSectionAsync();
```
pranavkm: This code makes me sad. c`est la vie

aspnetcore/mvc/models/file-uploads/samples/2.x/SampleApp/Pages/BufferedDoubleFileUploadDb.cshtml.cs
```diff
@@ -0,0 +1,114 @@
...
+            // attribute violations.
+            if (!ModelState.IsValid)
+            {
+                Result = "Please correct the form.";
+
+                return Page();
+            }
+
+            var formFiles = new List<IFormFile>() 
+                {
```
pranavkm: Weird indentation

aspnetcore/mvc/models/file-uploads/samples/2.x/SampleApp/Pages/Index.cshtml
```diff
@@ -0,0 +1,109 @@
...
+        <tbody>
+            @foreach (var file in Model.DatabaseFiles) {
+                // Don't display the untrusted file name in the UI. HTML-encode the value.
+                var fileName = WebUtility.HtmlEncode(file.UntrustedName);
+                <tr>
+                    <td>
+                        <a asp-page="./DeleteDbFile" asp-route-id="@file.Id">Delete</a>
+                    </td>
+                    <td>
+                        <b>@Html.DisplayFor(modelItem => fileName)</b><br>
```
pranavkm: `@fileName`

aspnetcore/mvc/models/file-uploads/samples/2.x/SampleApp/Utilities/FileHelpers.cs
```diff
@@ -0,0 +1,261 @@
...
+            { ".gif", new byte[] { 0x47, 0x49, 0x46, 0x38 } },
+            { ".jpeg", new byte[] { 0xFF, 0xD8, 0xFF, 0xE0, 0xFF, 0xD8, 0xFF, 0xE2, 0xFF, 0xD8, 0xFF, 0xE3 } },
+            { ".jpg", new byte[] { 0xFF, 0xD8, 0xFF, 0xE0, 0xFF, 0xD8, 0xFF, 0xE1, 0xFF, 0xD8, 0xFF, 0xE8 } },
+            { ".png", new byte[] { 0x89, 0x50, 0x4E, 0x47, 0x0D, 0x0A, 0x1A, 0x0A } },
+            { ".pdf", new byte[] { 0x25, 0x50, 0x44, 0x46 } },
+            { ".ppt", new byte[] { 0xD0, 0xCF, 0x11, 0xE0, 0xA1, 0xB1, 0x1A, 0xE1 } },
+            { ".pptx", new byte[] { 0x50, 0x4B, 0x03, 0x04 } },
+            { ".xls", new byte[] { 0xD0, 0xCF, 0x11, 0xE0, 0xA1, 0xB1, 0x1A, 0xE1 } },
+            { ".xlsx", new byte[] { 0x50, 0x4B, 0x03, 0x04 } },
+            { ".zip", new byte[] { 0x50, 0x4B, 0x03, 0x04, 0x50, 0x4B, 0x4C, 0x49, 0x54, 0x55, 0x50, 0x4B, 0x53, 0x70, 0x58, 0x50, 0x4B, 0x05, 0x06, 0x50, 0x4B, 0x07, 0x08, 0x57, 0x69, 0x6E, 0x5A, 0x69, 0x70 } },
```
pranavkm: I think these are several possible values for zip files. I don't think the header is this large.

aspnetcore/mvc/models/file-uploads/samples/2.x/SampleApp/Utilities/FileHelpers.cs
```diff
@@ -0,0 +1,261 @@
...
+            ModelStateDictionary modelState, string[] permittedExtensions, 
+            long sizeLimit)
+        {
+            var fieldDisplayName = string.Empty;
+
+            // Use reflection to obtain the display name for the model
+            // property associated with this IFormFile. If a display
+            // name isn't found, error messages simply won't show
+            // a display name.
+            MemberInfo property =
```
pranavkm: 😕

pranavkm: Can the user pass this in?

aspnetcore/mvc/models/file-uploads.md
guardrex: Generally, I'd say 'yes' to that. It makes one aware, for example, that an app working with files just a hair over 64KB will incur an immediate perf penalty over one that always handles files <64 KB.

I'll move it down in the section a paragraph to de-emphasize it a bit.

aspnetcore/mvc/models/file-uploads.md
guardrex: Yes, let's go that way, but I'm going to add this to the 3.0 updates for next week.

aspnetcore/mvc/models/file-uploads.md
guardrex: > `using var memoryStream = new MemoryStream()`

That's a C# 8.0 thing, right? Doesn't it break in a 2.2 app without setting the lang version up?

aspnetcore/mvc/models/file-uploads.md
guardrex: Can u `ReadBytes` on a `StreamReader`? IntelliSense is hating on it. Is that another C# 8.0 thing? Maybe I'm missing a namespace?

Your version defeats one of the goals. The current code makes sure that not only it matches *a signature in the dict* but that it matches the *one specific signature for that file extension*. I think that's important, but a dev would be welcome to go another way with it.

aspnetcore/mvc/models/file-uploads.md
guardrex: I'll note this on the 3.0 update issue.

aspnetcore/mvc/models/file-uploads/samples/2.x/SampleApp/Utilities/FileHelpers.cs
guardrex: That's what the Azure engineer recommended.

https://docs.microsoft.com/en-us/azure/security/develop/threat-modeling-tool-input-validation#example-8

Do we want to track that engineer down and see if they can shed more light on it?

aspnetcore/mvc/models/file-uploads/samples/2.x/SampleApp/Utilities/FileHelpers.cs
guardrex: AFAIK, this is the only way to get the display name for the model property ✨ *auto-magically* ✨. Do you think that reflection is really that bad to use/show? It would simplify this quite a bit to just have them pass it in. However, it would duplicate what they have for the model property's display name. This ended up in here because I wondered if I could just have my display names appear in **one place**. When I change it there, it changes it everywhere for every purpose. I don't mind the reflection; but if you prefer dropping it, then that's cool. I like *simple/straightforward stuff*, too.

aspnetcore/mvc/models/file-uploads/samples/2.x/SampleApp/Pages/Index.cshtml
guardrex: Can I do this? ... and drop the encoding on Line 40? ...

```cshtml
<td>
    <b>@file.UntrustedName</b><br>
    @Html.DisplayFor(modelItem => file.Note)
</td>
```

aspnetcore/mvc/models/file-uploads.md
Rutix: @pranavkm To answer the question "Is it useful for the user to know?" Yes it is. As an user we have had problems with this in the past because we did not know this happens. For example we gracefully wanted to handle the error flow when there is a virus but if you do not know it writes to disk the virusscanner will react and throw it away.

aspnetcore/mvc/models/file-uploads.md
guardrex: Thinking back to what I was *trying* to do here (and failing to compose well 🙈) is more of a pure HTML/JS static page-y approach for old skoolers/special circumstances/those learning .NET and used to other ways. Let me try another version and update that text on the next commit. What I have here doesn't make sense. ............. and also I'll get the fetch into it. I changed my mind about holding off on that until 3.0 updates. Best to do it now.

guardrex: ..... well .... nah ..... doesn't make sense to talk about antiforgery in that context here if the "solution" is `asp-`-anything to make it work. Rubber :duck: says to strip most of this, place fetch here, and update the sample to use fetch. The sample works without `asp-antiforgery` set to `true` **_anyway_** in its AJAX example.

We'll see what happens down the road. If one or more devs get into trouble where they don't get an antiforgery token and need to know about `asp-antiforgery`, then we can bring some form of this back.

guardrex: One more note here tho: I don't get the cookie for the AJAX in this example, and that might explain why the sample works and this in-text example requires the `asp-antiforgery` setting.

guardrex:
@pranavkm There are at least four things up there :point_up: to discuss further.

... and ...

> I would much rather users understand the documentation and build up something that works best for them rather than use a thing that looks like a prescriptive \ general purpose solution for files. Do we absolutely need it?

... *"understand the documentation and build up something that works best for them"* ...

I don't quite follow u there ... probably need to explain that out a bit more to me.

What's here (what's in the sample app) are just examples of how things can be accomplished in various scenarios. A dev is welcome to change or strip out whatever they don't want. I wish to provide more (and *safe*) implementation guidance than the ho-hum file uploads topic out there.

The helper itself (the class) is here because *I most definitely want to show them* how to:

* Do a simple, single file upload with one control to dB or physical.
* Do a multiple files upload with one control to dB or physical.
* Do a multiple files upload with multiple controls in one form to dB or physical.
* Keep those cool streaming examples in there ... *separately* from those basic cases.

... and, of course, we'd want to just have one class for any sort of security, limits, streaming, and other cool things that we want to do across those examples.

It's **_dead simple_** this way. There's no guesswork or head-scratching ... 💀 *NO death threats* 💀 to the topic author. :smile: ....... well .... *vastly fewer than we've been getting from readers!*

One alternative is to describe/show little or nothing about many of the scenarios and leave the dev to fend for themselves with blog posts, support forums, and chat ... the quality of which are all over the place with really, really bad advice and samples all over the Net. Sure, we could drop a lot of the security, validation, limits examples, tie-in with model binding display names, etc. ... indeed pretty much 🔪 the whole file helper thing), but I don't recommend it:&dagger; ... if that's what you're suggesting.

*don't recommend it*:&dagger; &ndash; **Just a recommendation.** I'll kill all of it (or whatever parts you don't want) if you decide to go that way. :ear:

aspnetcore/mvc/models/file-uploads.md
guardrex:
```csharp
var signatures = _fileSignature[ext];
var headerBytes = reader.ReadBytes(signatures.Max(m => m.Length));

return signatures.Any(signature => 
    headerBytes.Take(signature.Length).ToArray().SequenceEqual(signature));
```

aspnetcore/mvc/models/file-uploads/samples/2.x/SampleApp/Utilities/FileHelpers.cs
Rutix: @guardrex @pranavkm there are different version on which you can check. But as far as I know the most common is PKZIP 1 which is `0x50 0x4B 0x03 0x04`. You also have then indication of what type zip it is. Eg. `0x50 0x4B 0x05 0x06` is empty archive and `0x50 0x4B 0x07 0x08` multi volume archive file.  :)

guardrex: With the new bits on this morning's commit, I put all of the ZIP sigs in there from the File Signatures Database. See how that implementation looks with the fancy one-liner check.

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -430,28 +430,33 @@ if (string.IsNullOrEmpty(ext) || !permittedExtensions.Contains(ext))
 
 ### File signature validation
 
-A file's signature is determined by the first few bytes that make up the file. These bytes can be used to indicate if the extension matches the content of the file. The sample app checks file signatures for common file types. In the following example, the file signature for a Microsoft Word document is checked against the file:
+A file's signature is determined by the first few bytes that make up the file. These bytes can be used to indicate if the extension matches the content of the file. The sample app checks file signatures for for a few common file types. In the following example, the file signature for a JPEG image is checked against the file:
```
Rutix: I think you mean `for a few common file types.` and not `for for a few common file types.`

guardrex: oooooops. 🙈 I'll fix that on the next commit. Thx.

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -430,28 +430,33 @@ if (string.IsNullOrEmpty(ext) || !permittedExtensions.Contains(ext))
...
-    }
+    var signatures = _fileSignature[ext];
+    var headerBytes = reader.ReadBytes(signatures.Max(m => m.Length));
+    
+    return signatures.Any(signature => 
+        headerBytes.Take(signature.Length).SequenceEqual(signature));
 }
 ```
 
+To obtain additional file signatures, see the [File Signatures Database](https://www.filesignatures.net/).
```
Rutix: I am not sure if you can treat this as a the main source. Maybe add "or the official specification of the file type" which would make it more complete.

Rutix:
@pranavkm @guardrex I am pretty sure the `using var memoryStream = new MemoryStream()` won't work. I just tried this on my development environment where I don't have 3.0 and it won't swallow it. Like @guardrex says C#8 added using declarations which we don't have in .NET Core 2.2

aspnetcore/mvc/models/file-uploads.md
guardrex: This was caused by a common problem working on docs: When using a wide window and a line breaks across the double-word, they're *very easy* to miss!

guardrex:
That was my experience, and I don't want to set the language version up in the project file. However, I've added it to the 3.0 update issue, so we can get that into the 3.0 sample next week.

aspnetcore/mvc/models/file-uploads.md
pranavkm: You're right. You could use the raw stream to read it, not `StreamReader`, but that's a bit more involved.

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -1,195 +1,717 @@
...
-[!INCLUDE [GetTempFileName](../../includes/GetTempFileName.md)]
+```csharp
+public class BufferedSingleFileUploadPhysicalModel : PageModel
+{
+    private readonly long _fileSizeLimit;
 
-[!code-csharp[](file-uploads/sample/FileUploadSample/Controllers/StreamingController.cs?name=snippet1)]
+    public BufferedSingleFileUploadPhysicalModel(IConfiguration config)
+    {
+        _fileSizeLimit = config.GetValue<long>("FileSizeLimit");\
```
pranavkm: incorrect slash at the end of this line

guardrex:
> but the helper would get used because it's more prescriptive \ easy to use

True, but that's the only way to demo it. I suppose ur saying to have topic sections that address these aspects and show little bits of example code in the topic but don't actually roll all of that into the sample because they'll just cut-and-paste it into their app.

guardrex:
@Rick-Anderson Hold off on your review until Monday. I'd like to make one more edit pass on the text. I'll do that either today (Sunday) or first thing Monday morning. I'll ping you with a new review request when ready.

guardrex:
@Rick-Anderson Alrighty ... look'in good. It's ready for your **_final :eyes:, updates, and merge_**. :tada:

There was already one small 3.0 update made to the PR on an earlier commit, but don't sweat 3.0 on this PR ... the main 3.0 update for this topic+sample is coming up this week on https://github.com/aspnet/AspNetCore.Docs/issues/13344.

guardrex:
btw @pranavkm ... As I was making that final pass and considering what u said about devs dumping the `FileHelpers` class into an app, I added a bit about that in the *Validation* section to the existing warning. It now goes ...

> [!WARNING]
> The validation processing methods demonstrated in the sample app don't scan the content of uploaded files. In most production scenarios, a virus/malware scanner API is used on the file before making the file available to users or other systems.
>
> Although the topic sample provides a working example of validation techniques, don't implement the `FileHelpers` class in a production app unless you:
>
> * Fully understand the implementation.
> * Modify the implementation as appropriate for the app's environment and specifications.
>
> **Never indiscriminately implement security code in an app without addressing these requirements.**

However if you reach a point where you just don't like that in the sample app, then I'm returning for 3.0 updates this week. I can 🔪 it out of the sample and create independent sections in the topic under *Validation* with merely code snippet examples.

What I hope isn't done is a wholesale 🔪 of all validation/security concept and example coverage ... let's at least go for topic sections. The hacking and data breaches are **so bad** these days. I'm just disgusted with it, so I hope we do whatever we can here.

guardrex:
# **:tada: _WOOT!_ 🕺💃**

Thanks @Rutix and @ardalis!

I'm going to work on the 3.0 updates for this soon.

Rutix:
Wooooo! Nice! Thank you too @guardrex! I will follow the 3.0 changes and if I can contribute I will.

guardrex:
I'll ping u on the PR.

PR #12375, opened 2019-05-10
Add Blazor RenderTreeBuilder engineering content

### LINKED ISSUE #11684: RenderTreeBuilder coverage feedback
@Lupusa87 commented on https://github.com/aspnet/Docs/pull/11016 ...

I am sorry but I don't agree here in lot of things.
It will be hard to explain but I will try.

I think we should not hide and put behind this great approach to build components dynamically.
Building dynamic components is strength of blazor which makes it powerful tech.

There is a question? Can we do everything what we need using both approaches, with razor pages and with buildrendertree?
I think no and name **dynamic** is relevant here.

In razor pages we can use if, switch to adapt UI on parameters/state but it is still **static** at this point, when in code behind we render completely UI as we need and it can be 100% depended on code logic.

For example my [SvgHelper](https://github.com/Lupusa87/BlazorSvgHelper/blob/master/README.md) which is separated blazor lib (published on nuget) is receiving builder as parameter and adding tags based on what developer did with c# svg classes.
Can you do same in razor pages? No.

This approach is dynamic because everything is generated based on provided logic and methods (internal or external) can return pieces of UI and together build whole component body.

@rynowak Had idea to make same for razor pages which will be great but is not available yet.
https://twitter.com/aVerySpicyBoi/status/1104110920102375424

If someone likes razor way and thinks it is enough and finds it simpler, no problem, they can do it as they like.
But it does not give too much coding sugar and doing rendering in c# can be preferred way to many developers, especially when doing blazor native components, blazor libs, tools and bit complex apps.

I can agree that this is bit advanced scenario but not too much and it should be considered and recommended as general development style. 

There is nothing hard and highly advanced to use buildrendertree even for entry level developer, for bit experienced devs it should be trivial thing.

Warning that there can be unclosed tag is not relevant because same is still problem in razor page because compiler is not detecting it and any other html syntax errors.
Unclosed tag or any similar issue we detect and solve when build/test our apps and can not be reason not to chose buildrendertree way.

Hard code sequence numbers is not possible because as said already in c# class we can realize complex logic based on parameters/state and generate huge UI so it is impossible to hard code sequence, if it is problem it should be solved with another way rather than telling developers don't do this.
Dynamic nature makes no sense hard code recommendation.

Using razor pages should be recommended in very simple cases, leading way should be c# way.
It should be well documented, explained and supported with examples.

Blazor is going to be standard way to build web applications at least for .net developers and it should not concentrate on toys like razor page, most developers will skip it in most apps which are bit complex than counter sample, we expect in future to see large and wide range web apps ported or build in blazor, this naturally requires buildrendertree.

People with serious intentions to blazor should see how powerful mechanism do they have in their hands.

Components are basic concept for blazor and option to build them dynamically and unlimited is what we should tell everyone and make it mainstream.

Here you are people whom I hard imagine to teach something but in this case I can't not to protect better approach and not try to push it forward.

I am sorry if my comment is confusing or not clear, it is hard to do better because of my limited English.

Please consider that I am facing this situations each day and I share my thoughts with you as one blazor dev, you can ask another devs too.
People who work close to components will have fast answer on this issue.

@Andrzej-W commented ...

I have to say that @Lupusa87 is 100% right. If you want to create complex and reusable component it is sometimes impossible to do it in standard way (in .razor file). Even if it is possible it is often easier to do it in code. As @Lupusa87 said it have to be well documented and that is the only problem.

Nobody likes something like this in the doc:

> When creating components dynamically, always hardcode the sequence number. Do not generate the number using any sort of calculation or counter.

The question immediately arises: what is wrong with sequence++ statement?

---
#### Document Details

* ID: 11a64802-62dc-40e0-034f-fcee55f31d4e
* Version Independent ID: 038d7f90-aa71-a35c-cee9-c8543403cfd7
* Content: [Create and use Razor Components](https://docs.microsoft.com/en-us/aspnet/core/razor-components/components?view=aspnetcore-3.0)
* Content Source: [aspnetcore/razor-components/components.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/razor-components/components.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #11684


aspnetcore/blazor/components.md
```diff
@@ -889,7 +889,7 @@ Consider the following Pet Details component, which can be manually built into a
...
+In the following example, the loop in the `CreateComponent` method generates three Pet Details components. When calling `RenderTreeBuilder` methods to create the components (`OpenComponent` and `AddAttribute`), sequence numbers are source code line numbers. The Blazor difference algorithm relies on the sequence numbers corresponding to distinct lines of code, not distinct call invocations. When creating a component with `RenderTreeBuilder` methods, hardcode the arguments for sequence numbers. **Using a calculation or counter to generate the sequence number can lead to poor performance.** For more information, see [Why sequence numbers should relate to code line numbers, not execution order (SteveSandersonMS/sequence-number.md Gist)](https://gist.github.com/SteveSandersonMS/ec232992c2446ab9a0059dd0fbc5d0c3).
```
danroth27: I'm really against having links to external blog posts/gists/etc as as substitute for documenting our framework features. Can we please just move the relevant parts of Steve's gist into our official docs?

guardrex: Sure ... I'll see if I can condense his content a bit and move it over here.

guardrex:
@danroth27 Edited but not shortened. Steve's content is clear and well written. I just cleaned it up a bit.

aspnetcore/blazor/components.md
```diff
@@ -923,3 +923,93 @@ In the following example, the loop in the `CreateComponent` method generates thr
     }
 }
 ```
+
+### Sequence numbers relate to code line numbers and not execution order
+
+Unlike `.jsx` files, `.razor` files are always compiled. This is potentially a great advantage for `.razor` because the compile step can be used to inject information that improve app performance at runtime.
```
danroth27: Let's not talk about .jsx files - we shouldn't assume folks know React.

guardrex:
Hi @danroth27 ... I killed that bit about React. Does the PR require more work?

aspnetcore/blazor/components.md
```diff
@@ -923,3 +923,93 @@ In the following example, the loop in the `CreateComponent` method generates thr
...
+
+This is a trivial example. In more realistic cases with complex and deeply nested structures, and especially with loops, the performance cost is more severe. Instead of immediately identifying which loop blocks or branches have been inserted or removed, the diff algorithm has to recurse deeply into the render trees and usually build far longer edit scripts because it is misinformed about how the old and new structures relate to each other.
+
+#### Guidance and conclusions
+
+* App performance suffers if sequence numbers are generated dynamically.
+* The framework can't create its own sequence numbers automatically at runtime because the necessary information doesn't exist unless it's captured at compile time.
+* Don't write long blocks of manually-implemented `RenderTreeBuilder` logic. Prefer `.razor` files and allow the compiler deal sequence numbers. If you're unable to avoid manual `RenderTreeBuilder` logic, split long blocks of code into smaller pieces wrapped in `OpenRegion`/`CloseRegion` calls. Each region has its own separate space of sequence numbers, so you can restart from zero (or any other arbitrary number) inside each region.
```
```suggestion
* Don't write long blocks of manually-implemented `RenderTreeBuilder` logic. Prefer `.razor` files and allow the compiler to deal with the sequence numbers. If you're unable to avoid manual `RenderTreeBuilder` logic, split long blocks of code into smaller pieces wrapped in `OpenRegion`/`CloseRegion` calls. Each region has its own separate space of sequence numbers, so you can restart from zero (or any other arbitrary number) inside each region.
```

Issue #12091, opened 2019-04-21
Google et al without Identity


serpent5:
I wrote an [answer](https://stackoverflow.com/a/53705039/2630078) on Stack Overflow at the end of last year that explained how to use Google authentication without ASP.NET Core Identity. Is this general topic something that would be useful as its own page in the docs?

I expect this has come up before and perhaps has been dismissed, but I'm happy to give a more detailed proposal for the page if there's any interest.

Rick-Anderson:
@blowdart  this sound like a good topic. Do you agree?

blowdart:
It's fine, I do it in my conference labs. It's just one of those things that falls out of date every three months as google or facebook change their application management UI yet again, so beware there will be a maintenance cost.

serpent5:
I've been giving this a bit of thought and have a few ideas that I'd like to get some feedback on before considering the finer details of the page I suggested. I imagine some of this has already been considered and is very deliberate, so it'd be great to learn a bit more around that too. Here we go:

* Add an overview page that describes authentication along with links to some of the standout sections. This would be similar to the authorisation overview page.
* Add information about authentication schemes along with details about challenge, etc. This could be its own page or perhaps be part of the overview suggested above. I haven't seen anywhere that covers this already.
* Group all of the Identity-specific information under an Identity topic of some sort. There's a mix of Identity and non-Identity links all at the same level and a separate topic would allow the docs to lead with the explanation of authentication suggested above without going directly into Identity. 

I understand that the last suggestion would have a big impact on the structure of the docs, so I completely appreciate that it might not be practical even if it were desired.

serpent5:
@Rick-Anderson

Rick-Anderson:
@serpent5  go ahead with the PR.  Should it ever get too far out of date we'll delete the article.

PR #12385, opened 2019-05-11
Google et al without Identity + sample.

serpent5:
Initial work for #12091.

I've created a sample app that demonstrates login and logout using Google + Cookies without Identity. I've also created an almost-empty docs page that just contains links to the relevant source snippets that I intend to discuss within the page.

I'm submitting this early as a draft PR so that I can get feedback on two things:

1. The location and naming of the docs page.
2. The sample app. I don't expect to add anything else to this so I'd appreciate eyes on whether it's a good basis for the docs page.

I intend for the docs page to focus on Google specifically but to acknowledge and link to the other providers accordingly.


guardrex:
A couple of passing remarks:

* Idk if either alternative was considered ...
  * Use a modified version of the Additional Claims sample for this. (However, note that a big PR is modifying that one now on https://github.com/aspnet/AspNetCore.Docs/pull/11950.)
  * No sample ... just call out the changes that are required in one or a few snapshot files. If this is possible, it would be great to not have *one more sample* to update every release. We can't keep up as it is. :cry:
* Nit: We *try* to make GH happy about ending files on a blank line. See the *Privacy.cshtml.cs* file ... add a return to the last line. It may have even come from the template that way. I'll check the engineering repo. If they need a PR for it, I'll put that in with them.
* Since this subject matter pertains to social login, I think @tdykstra will want it in `aspnetcore/security/authentication/social/`, and the UID would get the *social* segment. Because of that, I would recommend that having the word "social" again in the URL isn't required. The last segment can be *without-identity* (or something else that doesn't include "social"). He'll come on and let u know what he wants to do on these points.
* Name the sample app *SampleApp*. We put them into a *samples* (plural) folder and then inside of a *2.x* folder for this version. That way , we can add a *3.x* folder with another *SampleApp* folder later. The samples link in the intro of the topic will go like this (but hold changes tho until u hear from Tom).

  > \[View or download sample code](https://github.com/aspnet/AspNetCore.Docs/tree/master/aspnetcore/security/authentication/social/without-identity/samples) (\[how to download](xref:index#how-to-download-a-sample))

  ... and the full path to a file would be ...

  > aspnetcore/security/authentication/social/without-identity/samples/2.x/SampleApp/SomeFile.xxxx

  Ellipses in the TOC? I'm surprised. Still tho ... my half-penny's worth ... something more like this would be better ...

  ```
  - name: Social authentication without Identity
    uid: security/authentication/social/without-identity
  ```

* For most samples we're drop static assets to slim the repo, but these authn/authz ones really suffer without them. We've been leaving the static assets (CSS, JS) in these authn/authz samps, so you're good there. :+1: However, you can drop the cookie policy bits from the sample. Not needed; we're dropping it from samples now; and IIRC, engineering is dropping it in the future anyway (although, they might keep it for auth templates ... not sure).

serpent5:
Thanks for the detailed feedback @guardrex - this is exactly why I decided to get visibility on this as early as possible. I’ll wait for the others to respond accordingly before I look into making changes, as you’ve suggested. It seems like I might already be looking at a do-over so perhaps I’m being a bit too ambitious with contributing on this one... 😟

guardrex:
I hope they say (and u think it can be done) that we don't need a full sample ... just snapshots. I'm still trying to get back to our 2.0 and 2.1 samples (and I think there are even 1.x samples! *Yikes!*). 🏃😅 .... this is one case where *The more the merrier!* definitely **_does NOT apply_**. :smile: lol

guardrex:
I sent in a patch PR on the line endings. That will speed us up over here on docs. I always stop and add them as I go with a new sample. https://github.com/aspnet/AspNetCore/pull/10192

Rick-Anderson:
>For most samples we're drop static assets to slim the repo,

Just the minified and a few others that aren't used in local dev. That's automatic via *.gitignore*

guardrex:
I think the scenario where things get ugly is working in development. There's no fallback for the non-min versions of static assets.

Rick-Anderson:
>I think the scenario where things get ugly is working in development. There's no fallback for the non-min versions of static assets.

Just use *.gitignore*. That's our standard. The debug assets are trivial.

>My approach totally removes them (in non-auth sample cases). My samps always look marvelous! 😄 .... hopefully, they run ok, too! lol

Don't do that. Just depend on *.gitignore*

guardrex:
> The debug assets are trivial ... Just depend on .gitignore

They aren't. In dev, static assets won't be present on the dev's machine for a downloaded sample. The sample will look great here (on my machine) but when the dev downloads it from the repo (without non-min static assets) it will look like crap when they run it in development.

If you're saying to force the static assets into the repo with .gitignore, ok, but that defeats the goal of having slim samples.

https://github.com/aspnet/AspNetCore.Docs/issues/9186 is still open ... recommend that we get that figured out if there's still a misunderstanding on my part. I'm just going by what I thought we worked out there ... see DR's remark at https://github.com/aspnet/AspNetCore.Docs/issues/9186#issuecomment-432827970.

serpent5:
@guardrex Are there any samples I can help bring up to 2.2 to try and lighten the load?

guardrex:
> Are there any samples I can help bring up to 2.2 to try and lighten the load?

I defer to management on that.

Rick-Anderson:
@serpent5 take a look at the issues and let me know what you'd like to do. It doesn't matter if the issue is currently assigned to someone else.

serpent5:
What are the next steps for this?

Rick-Anderson:
@serpent5 sorry. I'll start reviewing it today.

serpent5:
That's fine. No need to apologise at all. I'm just checking in here as the comments went a bit off topic and I also wanted to make sure you know I'm still keen to try and get this done.

Rick-Anderson:
@serpent5  what are the steps to get this working? This is what I tried:

`dotnet new webapp -o WebApp1 -au Individual -uld`

Remove `<partial name="_LoginPartial" />` from the layout file.

Update code to what you have. I run the code and click on the Privacy link, but a break point on `OnGetLogin` is never hit.

If we do this, I think we'd need to hook up the UI so you could sign-in/sign-out

serpent5:
You don't want to create the project with `-au Individual -uld` as that sets up Identity, which I expect is conflicting with the `AddAuthentication` stuff I've added. `OnGetLogin` is a replacement for what would normally be the "Account/Login" route in MVC/RP and is triggered within the sample I created by clicking on a login link that I added to the navbar to mimic what it would look like with Identity.

If you run the sample app as is, it should all work. If you click "Privacy", it won't trigger the `OnGetLogin` page but it will send you straight over to Google. `OnGetLogin` and `OnPostLogoutAsync` are handlers that get triggered from the navbar in the sample, which I added in an attempt for consistency with the Identity approach. You can see this in the `_LoginPartial` I added and modified in the sample.

I'd intended for the docs to be a guide that called out the sections most relevant to the auth stuff, but to just mention that the navbar in the sample app is linked to those page handlers. I'm happy to change the approach to a step-by-step guide that starts with an empty app and talks through the start to finish process. Alternatively, we can ditch the page handlers and just detail adding the auth stuff and applying the `[Authorize]` attribute. This could be based on the webapp template or could be more basic, starting from the empty template.

Rick-Anderson:
I like your approach.  I think it's best to keep it as simple as possible.

Let me add some instructions and we'll see what the security folks say.

Rick-Anderson:
Moved to #12444

Rick-Anderson:
@serpent5 Using postman, I post to https://localhost:44369/Index to logout but I get back 400 bad request.

When I'm not sign-ed in. I have a break-point (BP) on
```
public IActionResult OnGetLogin()
{
    return Challenge(new AuthenticationProperties
    {
        RedirectUri = Url.Page("/Index")
    });
}
```
I redirect too google to approve, but I never hit the above BP.

serpent5:
@Rick-Anderson If you're using Postman, it's likely 400ing due to the antiforgery token being absent. The endpoint would also need the `Logout` handler to invoke the logout code, i.e. `https://localhost:44369/Index?handler=Logout`. Otherwise, it's just going to render the view without first logging out.

`OnGetLogin` was only there for the case where we have a "Sign in" button, in order to issue a challenge explicitly. If we don't have a button/link for signing in, we don't need to include a `OnGetLogin` handler.

Thanks for your patience here. I had intended to flesh out the docs page after the sample was reviewed, but things have gone in a different direction to what I originally anticipated, so I realise that what I've provided isn't very clear now.

Rick-Anderson:
@serpent5  thanks. I'll remove the  `OnGetLogin` handler.  I think all we need is a mechanism to logout.

PR #12391, opened 2019-05-13
Add Anchor Link for Implicit Versioning

spottedmahn:
> See below for more information

Which below info?  Let's make it explicit 😊


spottedmahn:
oh, nice, good idea right back atcha!

https://docs.microsoft.com/en-us/contribute/how-to-write-links#link-text

>Do not use "click here." It's bad for search engine optimization, and doesn't adequately describe the target.

PR #12392, opened 2019-05-13
Blazor _Host.cshtml coverage + base path touchup

### LINKED ISSUE #12188: Enhance _Host.cshtml coverage
It's not clear to me how one can use plugins like AddMe or Disqus with Blazor pages. Note that with regard to "Inside the &lt;head&gt; element of wwwroot/index.html, provide a function that uses TextDecoder to decode a passed array", with the recent version of Blazor and .net core 3 which I downloaded 3 days ago, there isn't an index.html in wwwroot. Instead, one seems to need to put any javascript that one wants to call into _Host.cshtml, so perhaps the documentation needs to be updated? I've tried to put javascript code there for Disqus and then call it in OnAfterRender in a Razor page, but without managing to get it to run without errors so far.  So any suggestions would be welcome!

---
#### Document Details

* Content: [Blazor JavaScript interop](https://docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interop?view=aspnetcore-3.0#feedback)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #12188 

* Adds *\_Host.cshtml* coverage.
* Touch the app base path coverage in Blazor client-side host and deploy topic.

Thanks @patrickjlee! :rocket:

btw- I would've asked here about contacting Disqus about what we did on #12188 to get an example of a Disqus implementation in Blazor apps, but I think we'll need Blazor at RTM to approach them with the example. Maybe it's something we could do later this year.

### DIFF
--- a/aspnetcore/blazor/hosting-models.md
+++ b/aspnetcore/blazor/hosting-models.md
@@ -5,7 +5,7 @@ description: Understand client-side and server-side Blazor hosting models.
 monikerRange: '>= aspnetcore-3.0'
 ms.author: riande
 ms.custom: mvc
-ms.date: 05/10/2019
+ms.date: 05/13/2019
 uid: blazor/hosting-models
 ---
 # Blazor hosting models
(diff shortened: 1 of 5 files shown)


guardrex:
ping @danroth27

PR #12426, opened 2019-05-15
File uploads in Custom Model Binding topic

### LINKED ISSUE #12292: Potential Path traversal issue in the sample
The docs have the following line of code:
```
Path.Combine(_env.ContentRootPath, "wwwroot/images/upload", filename);
```
Which might be vulnerable to path traversal (e.g. when filename is `../../../` or something like that. Please correct me if I'm missing something, I'm not an expert.
As people (including myself) tend to copy-paste code from docs - maybe worth adding a note here? Look like fixing the issue requires some regex (see [this post](https://blog.mindedsecurity.com/2018/10/how-to-prevent-path-traversal-in-net.html)) and will make the example a lot harder to read.

---
#### Document Details

* ID: cfb00531-4ac6-8bf6-9bd8-1eed6446397f
* Version Independent ID: ed418d09-6c33-699a-0d48-9aabc99457be
* Content: [Custom Model Binding in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/mvc/advanced/custom-model-binding?view=aspnetcore-2.2)
* Content Source: [aspnetcore/mvc/advanced/custom-model-binding.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/mvc/advanced/custom-model-binding.md)
* Product: **aspnet-core**
* GitHub Login: @ardalis
* Microsoft Alias: **riande**

guardrex:
Fixes #12292

* Mirrors the approach used in the [File Uploads PR (aspnet/AspNetCore.Docs #12344)](https://github.com/aspnet/AspNetCore.Docs/pull/12344).
* Specifically:
  * Have the example save the file elsewhere, not to *wwwroot*.
  * ~Cleanse the filename with `Path.GetInvalidFileNameChars` and Linq.~ Use `Path.GetRandomFileName` to get a fully trusted random file name.
* The topic makes no direct reference to any of this, so the topic doesn't require changes. Line highlights in the topic for the `regions` here don't require an update either.

Thanks @omerlh! :rocket:

### DIFF
--- a/aspnetcore/mvc/advanced/custom-model-binding/sample/CustomModelBindingSample/Controllers/ImageController.cs
+++ b/aspnetcore/mvc/advanced/custom-model-binding/sample/CustomModelBindingSample/Controllers/ImageController.cs
@@ -25,13 +25,12 @@ namespace CustomModelBindingSample.Controllers
         [HttpPost]
         public void Post(byte[] file, string filename)
         {
-            // Don't trust the file name sent by the client. Use Linq to
-            // remove invalid characters. Another option is to use
+            // Don't trust the file name sent by the client. Use
             // Path.GetRandomFileName to generate a safe random
-            // file name.
-            var invalidFileNameChars = Path.GetInvalidFileNameChars();
-            var trustedFileName = invalidFileNameChars.Aggregate(
-                filename, (current, c) => current.Replace(c, '_'));
+            // file name. _targetFilePath receives a value
+            // from configuration (the appsettings.json file in
+            // the sample app).
+            var trustedFileName = Path.GetRandomFileName();
             var filePath = Path.Combine(_targetFilePath, trustedFileName);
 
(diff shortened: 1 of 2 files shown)


blowdart:
Don't attempt to sanitize, use a random filename which you acknowledge is possible.

guardrex:
@blowdart Try it now.

I recommend an update pass on the whole topic at some point, and the sample for this topic is already scheduled for a 2.2 update on the [sample update tracking issue (aspnet/AspNetCore #5495)](https://github.com/aspnet/AspNetCore.Docs/issues/5495). For one thing, I prefer these remarks (the code comments) in the topic and not in the code for localization. @Rick-Anderson @tdykstra If you're ok with it, I'll make a full pass on the topic when the sample is updated for 2.2; otherwise, let me know if you only prefer a sample 2.2 update only; in which case, I'll just work the sample to 2.2 and leave the topic as-is.

PR #12444, opened 2019-05-16
social/without/identity

### LINKED ISSUE #12091: Google et al without Identity

Rick-Anderson:
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/security/authentication/social/social-without-identity?view=aspnetcore-1.0&branch=pr-en-us-12444)

Fixes #12091

@serpent5  I brought your commit over here

@guardrex  this is how you do minimal files


Rick-Anderson:
@serpent5 are you happy with my changes?

serpent5:
On the linked issue, we agreed that `OnGetLogin` could be removed, but that's still there in this one. Just in case it helps, I've created a branch and a couple of commits over on my GitHub - the [second commit](https://github.com/serpent5/aspnetcore-Docs/commit/dc47882ab2ff398fc747c1b114cd3184cf370679) shows just the changes from a pure `webapp` template.

What are the next steps for the page itself? I understand that we'll want to get the security folks to review the sample code and general approach, but do you plan to add more to the page itself?

Rick-Anderson:
@serpent5  how do I get your commits into this PR?

serpent5:
@Rick-Anderson Those commits assume a full sample app, but it looks like we moved away from that after Luke's comments on the original PR I started. Do you want to go back to having a full sample app? I could create a PR into your branch, if that would actually work whilst your branch is a PR itself.

I could also just create a fresh PR with the full sample app (minus the cookie policy stuff as noted by Luke) and the docs page you've got here, which could take over from this one. In case you do want to go with that, would you want to go with the location suggested by Luke in the comments of the original PR? This is what I did in the commits I linked above for reference. I'm more than happy to take responsibility for updating the sample app when 3.0 etc are released too, if that's something you guys would be happy with.

You're in charge here, so I'm happy to help out with whatever approach you want to go with. :smile:

Rick-Anderson:
@serpent5  I think the minimum code we have the better.  Folks who do this are expected to know about security. We don't want a complete walk-through tutorial.

What's the advantage of a full app? We don't want to have custom code that needs review and to be maintained.

I think what we have will get folks going in the right direction.

@blowdart check out [Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/security/authentication/social/social-without-identity?view=aspnetcore-1.0&branch=pr-en-us-12444) and let me know what else you need.

blowdart:
The code is fine, but it lacks explanation of why in the depth that using cookie authentication without identity does.

Is OnGetLogin() really needed? In MVC demos I never had to use a login controller with social auth. But I guess if we're having logout, it'd match, but that implementation is always going to bounce you to the root, instead of using the RedirectURI as you would in real life.

serpent5:
@blowdart I added `OnGetLogin` in a different PR that was pulled into this one. I was trying to make the sample app look like it would if it were using the Identity templates/UI, with a signin/signout section in the navbar and the name claim's value if the user is signed in. Since this isn't going to be a full sample, it makes sense to me that we can drop both `OnGetLogin` and `OnPostLogoutAsync` and just focus on the `AddAuthentication()` stuff. If you get chance, would you mind reviewing the comment I made [here](https://github.com/aspnet/AspNetCore.Docs/issues/12091#issuecomment-487312569) a while back re documenting some of the higher-level authn stuff?

@Rick-Anderson Understood. I'll wait until we're agreed on what happens to `OnGetLogin` and `OnPostLogoutAsync` etc, but I can suggest any relevant code changes here (I finally figured out how to do that). I've also got some ideas around introductory text etc than I can also suggest if needed.

serpent5:
@blowdart What are your thoughts on the above?

serpent5:
Is there anything I can do to help progress this?

Rick-Anderson:
@serpent5  can you make a stab at:

*explanation of why in the depth that using cookie authentication without identity does.*

Let's remove `OnGetLogin`

PR #12462, opened 2019-05-19
Model binding doc updates

### LINKED ISSUE #10502: Model binding doc updates
Moved from #1080 by `@danroth27`
 	
Something like "Model state is a record of what data was bound to your model and any errors with the bound data". Include a link to the validation doc.

- [ ] eliminate MVC-controller-views-specific language (include Razor Pages)
- [ ] explain [ bind] attribute
- [ ] Model binding doc should define what model state is
- [ ] Example of when binding fails - and how to fix it (pass a subset of model, and post back)
- [ ] UE edit - remove market speak (vast majority, works great)
- [ ] Provide a simple sample where it fails and show how to create a simple binder. For example if you post a subset of a model it won't create a model.

From #1680 

It would be nice to table out the list of formats that we support in model binding. For example, for `Dictionary<string, string>` following formats work which is kind of difficult for an end user to figure out:

_Model_: `Dictionary<string, string> states`
1. states[wa]=washington&states[va]=virginia&states[mi]=michigan
2. states[0].Key=wa&states[0].Value=washington&states[1].Key=va&states[1].Value=virginia

 `@danroth27` wrote:
@dougbu is our model binding expert. He did the reviews on the [model binding doc](https://docs.asp.net/en/latest/mvc/models/model-binding.html) that is currently published. It might be enough just to get some time with him to get his thoughts on the specific issues this bug is tracking.

---
#### Document Details

* Content: [Model Binding in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/model-binding?view=aspnetcore-2.2)
* Content Source: [aspnetcore/mvc/models/model-binding.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/mvc/models/model-binding.md)
* Product: **aspnet-core**
* GitHub Login: @tdykstra

### LINKED ISSUE #12190: Please Clarify this paragraph
Model binding looks for the pattern parameter_name.property_name to bind values to properties. If it doesn't find matching values of this form, it will attempt to bind using just the property name. For those types such as Collection types, model binding looks for matches to parameter_name[index] or just [index]. Model binding treats Dictionary types similarly, asking for parameter_name[key] or just [key], as long as the keys are simple types. Keys that are supported match the field names HTML and tag helpers generated for the same model type. This enables round-tripping values so that the form fields remain filled with the user's input for their convenience, for example, when bound data from a create or edit didn't pass validation.

---
#### Document Details

* Content: [Model Binding in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/model-binding?view=aspnetcore-2.2#feedback)
* Content Source: [aspnetcore/mvc/models/model-binding.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/mvc/models/model-binding.md)
* Product: **aspnet-core**
* GitHub Login: @tdykstra

tdykstra:
Fixes #10502 
Fixes #12190
[Internal staging link](https://review.docs.microsoft.com/en-us/aspnet/core/mvc/models/model-binding?branch=pr-en-us-12462)


wadepickett:
Currently reviewing the docs and code...

aspnetcore/mvc/models/model-binding.md
```diff
@@ -1,146 +1,434 @@
...
+
+And the app receives a request with this URL:
+
+```
+http://contoso.com/api/pets/2?DogsOnly=true
+```
+
+Model binding goes though the following steps after the routing system selects the action method:
+
+* Finds the first parameter of `GetByID`, an integer named `id`.
```
pranavkm: This is super nice

aspnetcore/mvc/models/model-binding.md
```diff
@@ -1,146 +1,434 @@
...
+
+### [BindProperties] attribute
+
+Available in ASP.NET Core 2.1 and later.  Can be applied to a controller or `PageModel` class to tell model binding to target all public properties of the class:
+
+[!code-csharp[](model-binding/samples/2.x/Pages/Instructors/Create.cshtml.cs?name=snippet_BindProperties&highlight=1-2)]
+
+### Model binding for HTTP GET requests
+
+By default, the `[BindProperty]` and `[BindProperties]` attributes have no effect for HTTP GET requests. Typically, all you need for a GET request is a record ID parameter. The record ID is used to look up the item in the database. Therefore, there is no need to bind a property that holds an instance of the model. In scenarios where you do want properties bound to data from GET requests, set the `SupportsGet` property to `true`:
```
pranavkm: >  attributes have no effect for HTTP GET requests

Should this instead say "do not bind properties for HTTP GET requests"?

aspnetcore/mvc/models/model-binding.md
```diff
@@ -1,146 +1,434 @@
...
+
+1. Form fields 
+1. The request body (For [controllers that have the [ApiController] attribute](xref:web-api/index#binding-source-parameter-inference).)
+1. Route data
+1. Query string parameters
+1. Uploaded files 
+
+For each target parameter or property, the sources are scanned in the order indicated in this list. There are a few exceptions:
+
+* Form fields and request body data are used only for properties of complex types.
```
pranavkm: Model binding should be fine reading simple types from forms. If you're seeing otherwise, this might be a bug.

aspnetcore/mvc/models/model-binding.md
```diff
@@ -1,146 +1,434 @@
...
+* `[FromQuery]` attribute
+* `[FromRoute]` attribute
+* `[FromForm]` attribute
+* `[FromBody]` attribute
 
-Note: The strings in the URL route are not case sensitive.
+### [FromHeader] attribute
 
-MVC will try to bind request data to the action parameters by name. MVC will look for values for each parameter using the parameter name and the names of its public settable properties. In the above example, the only action parameter is named `id`, which MVC binds to the value with the same name in the route values. In addition to route values MVC will bind data from various parts of the request and it does so in a set order. Below is a list of the data sources in the order that model binding looks through them:
+Gets values from HTTP headers. For some headers, you can't make a C# parameter or property name that matches the header name because the header name has a hyphen in it.  In these cases, provide the header name to the attribute constructor:
```
pranavkm: > For some headers, you can't make a C# parameter or property name that matches the header name because the header name has a hyphen in it.  In these cases, provide the header name to the attribute constructor:

Should we make this a general purpose comment for the `From*` attribute list above? It's just as valid for query string parameters as it is for headers. Something along the lines of:

> All of these attributes optionally accept a model name value in the constructor. This is provided in case the property name does not match the value in the Request or if the value in the Request cannot be represented as a C# property (for instance, a header name might have a hypen in it).

aspnetcore/mvc/models/model-binding.md
```diff
@@ -1,146 +1,434 @@
...
 
-1. `Form values`: These are form values that go in the HTTP request using the POST method. (including jQuery POST requests).
+[!code-csharp[](model-binding/samples/2.x/Pages/Instructors/Index.cshtml.cs?name=snippet_FromHeader)]
 
-2. `Route values`: The set of route values provided by [Routing](xref:fundamentals/routing)
+### [FromQuery] attribute
 
-3. `Query strings`: The query string part of the URI.
+Gets values from the query string. The attribute must be added to model properties individually (not to the model class).
```
pranavkm: > The attribute must be added to model properties individually (not to the model class).

This shouldn't be unique to `FromQuery`, no?

aspnetcore/mvc/models/model-binding.md
```diff
@@ -1,146 +1,434 @@
...
+* Nullable simple types are set to `null`.
+* Non-nullable value types are set to `default(T)`. For example, a parameter `int id` is set to 0.
+* For complex Types, model binding creates an instance by using the default constructor, without setting properties.
+* Arrays are set to `Array.Empty<T>()`, except that `byte[]` arrays are set to `null`.
 
-## Customize model binding behavior with attributes
+ If model state should be invalidated when nothing is found in form fields for a model property, use the [[BindRequired] attribute](#bindrequired-attribute).
 
-MVC contains several attributes that you can use to direct its default model binding behavior to a different source. For example, you can specify whether binding is required for a property, or if it should never happen at all by using the `[BindRequired]` or `[BindNever]` attributes. Alternatively, you can override the default data source, and specify the model binder's data source. Below is a list of model binding attributes:
+Note that this behavior applies to model binding from posted form data, not to JSON data in a request body. Request body data is handled by [input formatters](#input-formatters).
```
pranavkm: not to ~~JSON data~~ JSON or XML data in a request body.

aspnetcore/mvc/models/model-binding.md
```diff
@@ -1,146 +1,434 @@
...
 
 ```csharp
-services.AddMvc().AddMvcOptions(options =>
-    options.ModelMetadataDetailsProviders.Add(
-        new ExcludeBindingMetadataProvider(typeof(System.Version))));
+public IActionResult OnPost(int? id, Instructor instructorToUpdate)
 ```
 
-To disable validation on properties of a certain type, add a [SuppressChildValidationMetadataProvider](/dotnet/api/microsoft.aspnetcore.mvc.modelbinding.suppresschildvalidationmetadataprovider) in `Startup.ConfigureServices`. For example, to disable validation on properties of type `System.Guid`:
+Model binding starts by looking through the sources for the key `instructorToUpdate.ID`. If that isn't found, it looks for `ID` without a prefix. The process is repeated for each `Instructor` property.
```
pranavkm: Model binding starts by looking for the presence of any value in the request with a key that starts with the prefix `instructorToUpdate`. If no values are found, model binding proceeds to bind the model without the `instructorToUpdate` prefix. 

~~The process is repeated for each `Instructor` property.~~

aspnetcore/mvc/models/model-binding.md
```diff
@@ -1,146 +1,434 @@
...
-{
-    services.AddMvc()
-        .AddXmlSerializerFormatters();
-   }
+[HttpPost]
+public IActionResult OnPost([Bind("LastName,FirstMidName,HireDate")] Instructor instructor)
 ```
 
-Code in the *Startup.cs* file contains a `ConfigureServices` method with a `services` argument you can use to build up services for your ASP.NET Core app. In the sample, we are adding an XML formatter as a service that MVC will provide for this app. The `options` argument passed into the `AddMvc` method allows you to add and manage filters, formatters, and other system options from MVC upon app startup. Then apply the `Consumes` attribute to controller classes or action methods to work with the format you want.
+The `[Bind]` attribute can be used to protect against overposting in *create* scenarios. It doesn't work well in edit scenarios because excluded properties are set to null or a default value instead of being left unchanged. For defense against overposting, view models are recommended rather than the `[Bind]` attribute. For more information, see [Security note about overposting](xref:data/ef-mvc/crud#security-note-about-overposting).
```
pranavkm: 👍

aspnetcore/mvc/models/model-binding.md
```diff
@@ -1,146 +1,434 @@
...
+* For all of the preceding example formats, model binding passes a dictionary of two items to the `selectedCourses` parameter:
+
+  * selectedCourses["1050"]="Chemistry"
+  * selectedCourses["2000"]="Economics"
+
+## Special data types
+
+There are some special data types that model binding can handle.
+
+### IFormFile
```
pranavkm: `IFormFile` and `IFormFileCollection`

aspnetcore/mvc/models/model-binding.md
```diff
@@ -1,146 +1,434 @@
...
+
+Used to cancel activity in asynchronous controllers.
+
+### FormCollection
+
+Used to retrieve all the values from posted form data.
+
+## Input formatters
+
+Data in the request body can be in JSON, XML, or some other format. To parse this data, model binding uses an *input formatter* that is configured to handle a particular content type. By default, ASP.NET Core includes <xref:Microsoft.AspNetCore.Mvc.Formatters.JsonInputFormatter> and <xref:Microsoft.AspNetCore.Mvc.Formatters.JsonPatchInputFormatter> classes for handling JSON data. You can add other formatters for other content types.
```
pranavkm: > By default, ASP.NET Core includes ...

This is not true in 3.0. We have a different set of formatter types. Could we instead say 

> By default, ASP.NET Core includes JSON based input formatters for handling JSON data.

PR #12477, opened 2019-05-20
Blazor call web API topic with sample

### LINKED ISSUE #10743: Call web APIs from Blazor topic
From @danroth27 at [Add topic on calling web APIs from Blazor](https://github.com/aspnet/Blazor.Docs/issues/71) ...

- `HttpClient`
- JSON helpers
- Setting fetch arguments for the request (see https://github.com/aspnet/Blazor/pull/399#issuecomment-377193365)
- Making cross origin request

guardrex:
Fixes #10743

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/call-web-api?view=aspnetcore-3.0&branch=pr-en-us-12477)

* Not crazy about the title, but we can work on that.
* The sample app component (*Pages/CallWebAPI.razor*) is built in concert with our [web API tutorial](https://docs.microsoft.com/aspnet/core/tutorials/first-web-api). The examples in the topic and the sample app code are set up for a happy path by merely running the web API tutorial app and then running the code in the topic or the Razor component in sample against that service.
* I'm mostly concerned about the *HttpClient and HttpRequestMessage with Fetch API request options* section. It's the most challenging bit to organize and explain ... *lot's 'o concepts* there ... and there are probably *more concepts* to add.
* This 1st draft is *client-side-y* in nature. `There is no registered service of type 'System.Net.Http.HttpClient'.` ... I see what you mean on the engineering issue. Will take ur advice ... `IHttpClientFactory` for server-side.

### DIFF
--- a/aspnetcore/blazor/call-web-api.md
+++ b/aspnetcore/blazor/call-web-api.md
@@ -1,0 +1,230 @@
+---
+title: Call a web API
+author: guardrex
+description: Learn how to call a web API from a Blazor app using JSON helpers, including making cross-origin resource sharing (CORS) requests.
+monikerRange: '>= aspnetcore-3.0'
+ms.author: riande
+ms.custom: mvc
+ms.date: 05/20/2019
+uid: blazor/call-web-api
+---
+# Call a web API
+
+By [Luke Latham](https://github.com/guardrex)
+
+Blazor apps call web API services using [HttpClient](xref:fundamentals/http-requests). Compose requests using Blazor JSON helpers or with <xref:System.Net.Http.HttpRequestMessage>, which can include JavaScript [Fetch API](https://developer.mozilla.org/docs/Web/API/Fetch_API) request options.
+
+[View or download sample code](https://github.com/aspnet/AspNetCore.Docs/tree/master/aspnetcore/blazor/common/samples/) ([how to download](xref:index#how-to-download-a-sample))
+
+See the following components in the sample app:
+
+* Call Web API (*Pages/CallWebAPI.razor*)
+* HTTP Request Tester (*Components/HTTPRequestTester.razor*)
+
+## HttpClient and JSON helpers
+
+Use [HttpClient](xref:fundamentals/http-requests) and JSON helpers to call a web API service endpoint in a Razor component. Blazor client-side uses the [Fetch API](https://developer.mozilla.org/docs/Web/API/Fetch_API), while Blazor server-side uses <xref:System.Net.Http.HttpClient?displayProperty=fullName>.
+
+Include an `@using` statement for <xref:System.Threading.Tasks> for asynchronous web API calls and inject an `HttpClient` instance:
+
+```cshtml
+@using System.Threading.Tasks
+@inject HttpClient Http
+```
+
+In the following examples, a Todo web API service processes create, read, update, and delete (CRUD) operations. The examples are based on a `TodoItem` class that stores the:
+
+* ID (`Id`, `long`) &ndash; Unique ID of the item.
+* Name (`Name`, `string`) &ndash; Name of the item.
+* Status (`IsComplete`, `bool`) &ndash; Indication if the Todo item is finished.
+
+```csharp
+private class TodoItem
+{
+    public long Id { get; set; }
+    public string Name { get; set; }
+    public bool IsComplete { get; set; }
+}
+```
+
+JSON helper methods send requests to a URI (a web API in the following examples) and process the response:
+
+* `GetJsonAsync` &ndash; Sends a GET request and parses the JSON response body to create an object.
+
+  In the following code, the `_todoItems` are displayed by the component. The `GetTodoItems` method is triggered when the component is finished rendering ([OnInitAsync](xref:blazor/components#lifecycle-methods)). See the sample app for a complete example.
+
+  ```cshtml
+  @using Microsoft.AspNetCore.Blazor.Http
+  @inject HttpClient Http
+
+  @functions {
+      private const string ServiceEndpoint = "https://localhost:10000/api/todo";
+      private TodoItem[] _todoItems;
+
+      protected override async Task OnInitAsync() => 
+          await GetTodoItems();
+
+      private async Task GetTodoItems() => 
+          _todoItems = await Http.GetJsonAsync<TodoItem[]>(ServiceEndpoint);
+  }
+  ```
+
+* `PostJsonAsync` &ndash; Sends a POST request, including JSON-encoded content, and parses the JSON response body to create an object.
+
+  In the following code, `_newItemName` is provided by a bound element of the component. The `AddItem` method is triggered by selecting a `<button>` element. See the sample app for a complete example.
+
+  ```cshtml
+  @using Microsoft.AspNetCore.Blazor.Http
+  @inject HttpClient Http
+
+  <input type="text" bind="@_newItemName" placeholder="New Todo Item" />
+  <button onclick="@(async () => await AddItem())">Add</button>
+
+  @functions {
+      private const string ServiceEndpoint = "https://localhost:10000/api/todo";
+      private string _newItemName;
+
+      private async Task AddItem()
+      {
+          var addItem = new TodoItem { Name = _newItemName, IsComplete = false };
+          await Http.PostJsonAsync(ServiceEndpoint, addItem);
+      }
+  }
+  ```
+
+* `PutJsonAsync` &ndash; Sends a PUT request, including JSON-encoded content.
+
+  In the following code, `_editItem` values (`Id`, `Name`, `IsCompleted`) are provided by bound elements of the component. The `SaveItem` method is triggered by selecting the Save `<button>` element. See the sample app for a complete example.
+
+  ```cshtml
+  @using Microsoft.AspNetCore.Blazor.Http
+  @inject HttpClient Http
+
+  <input type="text" bind="@_editItem.Id" />
+  <input type="checkbox" bind="@_editItem.IsComplete" />
+  <input type="text" bind="@_editItem.Name" />
+  <button onclick="@(async () => await SaveItem())">Save</button>
+
+  @functions {
+      private const string ServiceEndpoint = "https://localhost:10000/api/todo";
+      private TodoItem _editItem = new TodoItem();
+
+      private async Task SaveItem()
+      {
+          await Http.PutJsonAsync(
+              Path.Combine(ServiceEndpoint, _editItem.Id.ToString()), _editItem);
+      }
+  }
+  ```
+
+<xref:System.Net.Http> includes additional extension methods for sending HTTP requests and receiving HTTP responses. [HttpClient.DeleteAsync](xref:System.Net.Http.HttpClient.DeleteAsync*) is used to send a DELETE request to a web API.
+
+In the following code, the Delete `<button>` element supplies the `id` when it's selected in the UI. See the sample app for a complete example.
+
+```cshtml
+@using Microsoft.AspNetCore.Blazor.Http
+@inject HttpClient Http
+
+<input type="text" bind="@_id" />
+<button onclick="@(async () => await DeleteItem())">Delete</button>
+
+@functions {
+    private const string ServiceEndpoint = "https://localhost:10000/api/todo";
+    private long _id;
+
+    private async Task DeleteItem()
+    {
+        await Http.DeleteAsync(Path.Combine(ServiceEndpoint, _id.ToString()));
+    }
+}
+```
+
+## HttpClient and HttpRequestMessage with Fetch API request options
+
+Use [HttpClient](xref:fundamentals/http-requests) and <xref:System.Net.Http.HttpRequestMessage> to supply request options to the underlying JavaScript [Fetch API](https://developer.mozilla.org/docs/Web/API/Fetch_API).
+
+In the following example:
+
+* JSON serialization and deserialization must be handled by user code (*not shown*).
+* The `credentials` property is set to any of the following values:
+
+  * `FetchCredentialsOption.Include` ("include") &ndash; Advises the browser to send credentials (such as cookies or HTTP auth headers) even for cross-origin requests. Only allowed when the CORS Middleware policy in the web API is configured to <xref:Microsoft.AspNetCore.Cors.Infrastructure.CorsPolicyBuilder.AllowCredentials*>.
+  * `FetchCredentialsOption.Omit` ("omit") &ndash; Advises the browser never to send credentials (such as cookies or HTTP auth headers).
+  * `FetchCredentialsOption.SameOrigin` ("same-origin") &ndash; Advises the browser to send credentials (such as cookies or HTTP auth headers) only if the target URL is on the same origin as the calling application.
+
+```cshtml
+@using System.Net.Http.Headers
+@using Microsoft.AspNetCore.Blazor.Http
+@inject HttpClient Http
+
+@functions {
+    private async Task PostRequest()
+    {
+        Http.DefaultRequestHeaders.Authorization =
+            new AuthenticationHeaderValue("Bearer", "{OAUTH TOKEN}");
+
+        var requestMessage = new HttpRequestMessage()
+        {
+            Method = new HttpMethod("POST"),
+            RequestUri = new Uri("https://localhost:10000/api/todo"),
+            Content = 
+                new StringContent(
+                    @"{""name"":""A New Todo Item"",""isComplete"":false}")
+        };
+
+        requestMessage.Content.Headers.ContentType = 
+            new System.Net.Http.Headers.MediaTypeHeaderValue(
+                "application/json");
+
+        requestMessage.Content.Headers.TryAddWithoutValidation(
+            "x-custom-header", "value");
+        
+        requestMessage.Properties[WebAssemblyHttpMessageHandler.FetchArgs] = new
+        { 
+            credentials = FetchCredentialsOption.Include
+        };
+
+        var response = await Http.SendAsync(requestMessage);
+        var responseStatusCode = response.StatusCode;
+        var responseBody = await response.Content.ReadAsStringAsync();
+    }
+}
+```
+
+For more information on Fetch API options, see [MDN web docs: Window​OrWorker​Global​Scope​.fetch():Parameters](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Parameters).
+
+When sending credentials (authorization cookies/headers) on CORS requests, allow the `Authorization` header when creating the CORS policy for CORS middleware. The following policy includes configuration for:
+
+* Request origins (`http://localhost:5000`, `https://localhost:5001`).
+* Any method (verb).
+* `Content-Type` and `Authorization` headers. To allow a custom header (for example, `x-custom-header`), list the header when calling <xref:Microsoft.AspNetCore.Cors.Infrastructure.CorsPolicyBuilder.WithHeaders*>.
+* Credentials set by client-side JavaScript code (`credentials` property set to `include`).
+
+```csharp
+app.UseCors(policy => 
+    policy.WithOrigins("http://localhost:5000", "https://localhost:5001")
+    .AllowAnyMethod()
+    .WithHeaders(HeaderNames.ContentType, HeaderNames.Authorization, "x-custom-header")
+    .AllowCredentials());
+```
+
+For more information, see <xref:security/cors> and the sample app's HTTP Request Tester component (*Components/HTTPRequestTester.razor*).
+
+## Cross-origin resource sharing (CORS)
+
+For more information on making cross-origin resource sharing (CORS) requests, see <xref:security/cors>. The sample app demonstrates CORS. See the Call Web API component (*Pages/CallWebAPI.razor*).
+
+## HTTP Request Tester
+
+The sample app includes an HTTP Request Tester component (*Components/HTTPRequestTester.razor*). The component is included in the Call Web API component:
+
+```cshtml
+<HTTPRequestTester />
+```
+
+Use the component to test request-responses against web API service endpoints.
+
+## Additional resources
+
+* <xref:fundamentals/http-requests>
+* [Cross Origin Resource Sharing (CORS) at W3C](https://www.w3.org/TR/cors/)
--- a/aspnetcore/blazor/call-web-api.md
+++ b/aspnetcore/blazor/call-web-api.md
@@ -1,14 +1,14 @@
 ---
-title: Call a web API
+title: Call a web API from ASP.NET Core Blazor
 author: guardrex
 description: Learn how to call a web API from a Blazor app using JSON helpers, including making cross-origin resource sharing (CORS) requests.
 monikerRange: '>= aspnetcore-3.0'
 ms.author: riande
 ms.custom: mvc
-ms.date: 05/20/2019
+ms.date: 06/14/2019
 uid: blazor/call-web-api
 ---
-# Call a web API
+# Call a web API from ASP.NET Core Blazor
 
 By [Luke Latham](https://github.com/guardrex)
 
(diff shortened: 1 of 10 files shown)


guardrex:
I'm double-posting that over here for my ease-of-access (*LazyRex*:tm:).

If you can guide me on the untrusted cert situation ... how to handle it better than this ... *along with any **gotchas!** ... then I could roll this into the topic (but not the sample app given that we only have a client-side common sample for our topics).

```xml
<ItemGroup>
  <PackageReference Include="System.Net.Http" Version="4.3.4" />
</ItemGroup>
```

```csharp
services.AddScoped(s =>
{
    var uriHelper = s.GetRequiredService<IUriHelper>();
    return new HttpClient
    {
        BaseAddress = new Uri(uriHelper.GetBaseUri()),
    };
});
```

```cshtml
@using System.Net.Http

@functions {
    private async Task MakeHttpClientRequest()
    {
        using (var httpClientHandler = new HttpClientHandler())
        {
            httpClientHandler.ServerCertificateCustomValidationCallback = 
                (message, cert, chain, errors) => true;
    
            using (var client = new HttpClient(httpClientHandler))
            {
                // Use client
            }
        }
    }
}
```

guardrex:
@danroth27 Shall we proceed with this? The client-side here looks good ... works great ... based on the web API sample provided by our existing [ASP.NET Core web API sample app topic](https://docs.microsoft.com/aspnet/core/tutorials/first-web-api?view=aspnetcore-2.2&tabs=visual-studio).

The problem is server-side, which I'm aware is in-design. We could either:

* Offer a workaround :point_up: (last comment) with some engineering help to make it good (e.g., get a little into the weeds on how to validate a cert there with that code).
* Go ahead with the client-side-only topic here. Add a NOTE at the top about server-side being in-design, and cross-link the note to [*Making HTTP requests from Blazor apps* (aspnet/AspNetCore #10397)](https://github.com/aspnet/AspNetCore/issues/10397). Open an issue for follow-up on the server-side later.

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,230 @@
...
+    private async Task DeleteItem()
+    {
+        await Http.DeleteAsync(Path.Combine(ServiceEndpoint, _id.ToString()));
+    }
+}
+```
+
+## HttpClient and HttpRequestMessage with Fetch API request options
+
+Use [HttpClient](xref:fundamentals/http-requests) and <xref:System.Net.Http.HttpRequestMessage> to supply request options to the underlying JavaScript [Fetch API](https://developer.mozilla.org/docs/Web/API/Fetch_API).
```
danroth27: Most of this stuff is just the normal way you use `HttpClient`, but the `FetchArgs` stuff is special.

```suggestion
Use [HttpClient](xref:fundamentals/http-requests) and <xref:System.Net.Http.HttpRequestMessage> to customize requests. 

When running on WebAssembly, supply request options to the underlying JavaScript [Fetch API](https://developer.mozilla.org/docs/Web/API/Fetch_API) using the `WebAssemblyHttpMessageHandler.FetchArgs` property on the request. As shown in the following example, the `credentials` property is set to any of the following values:
```

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,230 @@
...
+
+In the following example:
```
danroth27: Remove

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,230 @@
...
+
+* JSON serialization and deserialization must be handled by user code (*not shown*).
```
danroth27: Remove. You might not even be using JSON.

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,230 @@
...
+* The `credentials` property is set to any of the following values:
```
danroth27: Remove. Added in suggestion above.

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,230 @@
...
+app.UseCors(policy => 
+    policy.WithOrigins("http://localhost:5000", "https://localhost:5001")
+    .AllowAnyMethod()
+    .WithHeaders(HeaderNames.ContentType, HeaderNames.Authorization, "x-custom-header")
+    .AllowCredentials());
+```
+
+For more information, see <xref:security/cors> and the sample app's HTTP Request Tester component (*Components/HTTPRequestTester.razor*).
+
+## Cross-origin resource sharing (CORS)
```
danroth27: This section is pretty empty and there was already some CORs content in the section above about dealing with credentials. Maybe this section should move earlier in the doc along with a mention of the browser same origin policy and why you need the *server* to support CORS if you want to make requests to any other origin.

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,230 @@
...
+
+For more information on making cross-origin resource sharing (CORS) requests, see <xref:security/cors>. The sample app demonstrates CORS. See the Call Web API component (*Pages/CallWebAPI.razor*).
+
+## HTTP Request Tester
```
danroth27: I think we can remove this section. We already point folks to the sample app at the beginning.

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,230 @@
...
+        };
+
+        var response = await Http.SendAsync(requestMessage);
+        var responseStatusCode = response.StatusCode;
+        var responseBody = await response.Content.ReadAsStringAsync();
+    }
+}
+```
+
+For more information on Fetch API options, see [MDN web docs: Window​OrWorker​Global​Scope​.fetch():Parameters](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Parameters).
```
danroth27: Can you set any of these parameters using `FetchArgs`? If yes, we should say that. @SteveSandersonMS

aspnetcore/blazor/common/samples/3.x/BlazorSample/Components/HTTPRequestTester.razor
```diff
@@ -0,0 +1,178 @@
+@using System.Net
```
danroth27: Update sample code to preview6 and apply changes suggested for the snippets above.

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,230 @@
...
+See the following components in the sample app:
+
+* Call Web API (*Pages/CallWebAPI.razor*)
+* HTTP Request Tester (*Components/HTTPRequestTester.razor*)
+
+## HttpClient and JSON helpers
+
+Use [HttpClient](xref:fundamentals/http-requests) and JSON helpers to call a web API service endpoint in a Razor component. Blazor client-side uses the [Fetch API](https://developer.mozilla.org/docs/Web/API/Fetch_API), while Blazor server-side uses <xref:System.Net.Http.HttpClient?displayProperty=fullName>.
+
+Include an `@using` statement for <xref:System.Threading.Tasks> for asynchronous web API calls and inject an `HttpClient` instance:
```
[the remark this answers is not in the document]
guardrex: It doesn't seem like I need `System.Net.Http`. I don't need it in the sample app's component for this experience.

I just have ...

```cshtml
@page "/CallWebAPI"
@using System.IO
@using System.Linq
@using System.Threading.Tasks
@inject HttpClient Http
```

I had `System.Threading.Tasks` for the `await`s. However, they have access to the sample. It's just that we've been dinged **_so many times_** by readers for not calling out namespaces. In the following examples, `await` is used. I'll remove it, but I have a feel'in that readers will get out the pitchforks and torches again.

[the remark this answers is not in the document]
guardrex: ... and I called out the Blazor client- and server- side situation in the comments on the PR.

I'll try a NOTE at the top wrt Blazor server-side being in-design and link to the engineering issue.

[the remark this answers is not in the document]
guardrex: I deleted a remark here. I'll note my question in a comment when I shoot the commit in shortly.

dazinator:
> btw ... just curious @dazinator ... is this about what you ended up doing

Yes I registered my singleton httpclient with DI, and configured it in the factory delegate.

However.. my eyes have just been opened to the `IHttpClientFactory` way: https://github.com/aspnet/AspNetCore/issues/11440#issuecomment-504531334

So today I am now doing this:

```
  services.AddTransient<WebAssemblyHttpMessageHandler>(); // also register other custom handlers you see below..

 services.AddHttpClient<IMyService, MyService>()
                    .ConfigurePrimaryHttpMessageHandler<WebAssemblyHttpMessageHandler>()
                    .AddHttpMessageHandler<JwtAccessTokenHeaderHandler>()
                    .AddHttpMessageHandler<CsrfHeaderHandler>();    

```

I don't directly inject `HttpClient` into my components any longer, I inject a typed service as above - like IMyService which takes a configured HttpClient instance in it's constructor. I guess i should also remove the default HttpClient registration that blazor adds to make sure an unconfigured instance cannot be accidetnally injected into a component by someone who doesn't know any better.

guardrex:
@danroth27 Update pass made.

* I'm still a little off perhaps on server-side. I was under the impression from the in-design issue that it wouldn't be good at this time to add a remark like ...

  > In Blazor server-side apps, the \<xref:System.Net.Http.HttpClient> implementation is provided by .NET Core and can be used when making web API requests in server-side code.

  ... as that makes it seem like the `HttpClient` service is available for server-side ... however ... see if the NOTE and text that I updated works for the server-side scenario. I might not be too far off what we need to do.

  Also, we have two approaches as workaround now. We have what I show at https://github.com/aspnet/AspNetCore.Docs/pull/12477#issuecomment-494190766, and we have what @dazinator is doing at https://github.com/aspnet/AspNetCore.Docs/pull/12477#issuecomment-504541902. Either or both could be written in as a workaround, but https://github.com/aspnet/AspNetCore/issues/10397 is marked for Pre8 ... so that's not a very long wait if we just skip it. If a reader were to open an issue here, I'd just point them to this PR. IIRC, I noted this PR on your in-design engineering issue, too.

* It looks like `@using System.Net.Http` isn't required if there's an `@page` directive. Does that make sense? ... I addressed it on the last commit. We might need to work that text a little more.

* Finally, there's the ? that you pinged Steve on. I have no i-dear wrt the other Fetch API things that can be set. I'm all :ear: :ear: :ear: for what we should say there.

danroth27:
>  see if the NOTE and text that I updated works for the server-side scenario.

I'm not a fan of the note 😜. This topic isn't just for client-side Blazor. Also, the statement about how the `HttpClient` is implemented (using `fetch` or .NET Core's implementation) is separate from whether it is available as an injectable service. I think all we need to say is 1. `HttpClient` in client-side Blazor is based on `fetch`, while in server-side Blazor it's the normal .NET Core implementation, and 2. `HttpClient is available as a preconfigured service in client-side Blazor but not server-side Blazor. You should ignore the design issue for now.

The thing to realize here is that you don't typically want to make HTTP requests from a server-side Blazor app back to the same server. Instead you should abstract away this concern using your own service that has different implementations depending on where the app is host. For example, you would define an `IWeatherForecast` service and configure an implementation on the client that uses HttpClient to call a Web API and on the server just access the backing data store (probably a DB) directly. See https://github.com/danroth27/BlazorDualMode for an example of this setup.

> It looks like @using System.Net.Http isn't required if there's an @page directive. Does that make sense?

Nope :smile:. Blazor doesn't include a using statement for `System.Net.Http` by default. You have to add it. In the templates this is done in the root *_Imports.razor* file.

> Finally, there's the ? that you pinged Steve on. I have no i-dear wrt the other Fetch API things that can be set. 

Let's just ignore this for now. We can add more details later if we're missing something.

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,216 @@
...
+Blazor apps call web API services using [HttpClient](xref:fundamentals/http-requests). Compose requests using Blazor JSON helpers or with <xref:System.Net.Http.HttpRequestMessage>, which can include JavaScript [Fetch API](https://developer.mozilla.org/docs/Web/API/Fetch_API) request options.
+
+[View or download sample code](https://github.com/aspnet/AspNetCore.Docs/tree/master/aspnetcore/blazor/common/samples/) ([how to download](xref:index#how-to-download-a-sample))
+
+See the following components in the sample app:
+
+* Call Web API (*Pages/CallWebAPI.razor*)
+* HTTP Request Tester (*Components/HTTPRequestTester.razor*)
+
+> [!NOTE]
```
danroth27: I think we should remove this note. The topic isn't actually client-side Blazor specific, and we already in the text below address that HttpClient is only setup as a service client-side.

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,216 @@
...
+
+## HttpClient and JSON helpers
+
+Use [HttpClient](xref:fundamentals/http-requests) and JSON helpers to call a web API service endpoint from a Blazor app. In Blazor client-side apps, `HttpClient` is implemented using the browser [Fetch API](https://developer.mozilla.org/docs/Web/API/Fetch_API) and is subject to its limitations, including enforcement of the same origin policy. In Blazor server-side apps, the <xref:System.Net.Http.HttpClient> implementation is provided by .NET Core and can be used when making web API requests in server-side code.
+
+In Blazor client-side apps, `HttpClient` is available as a service. The client's base address is set to the originating server's address. Inject an `HttpClient` instance using the `@inject` directive. If the component isn't routable and thus doesn't include the `@page` directive, import <xref:System.Net.Http> with an `@using` directive:
```
danroth27:
```suggestion
In Blazor client-side apps, `HttpClient` is available as a preconfigured service for making requests back to the origin server. The client's base address is set to the originating server's address. Inject an `HttpClient` instance using the `@inject` directive:
```

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,216 @@
...
+@inject HttpClient Http
+```
+
+In the following examples, a Todo web API service processes create, read, update, and delete (CRUD) operations. The examples are based on a `TodoItem` class that stores the:
+
+* ID (`Id`, `long`) &ndash; Unique ID of the item.
+* Name (`Name`, `string`) &ndash; Name of the item.
+* Status (`IsComplete`, `bool`) &ndash; Indication if the Todo item is finished.
+
+```csharp
```
danroth27: cshtml -> razor? Do we support Razor snippets yet?

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,216 @@
...
+
+* `PutJsonAsync` &ndash; Sends a PUT request, including JSON-encoded content.
+
+  In the following code, `_editItem` values (`Id`, `Name`, `IsCompleted`) are provided by bound elements of the component. The `SaveItem` method is triggered by selecting the Save `<button>` element. See the sample app for a complete example.
+
+  ```cshtml
+  @using System.Net.Http
+  @inject HttpClient Http
+
+  <input @bind="_editItem.Id" />
```
danroth27: Remove? It's a bit unusually I think for the identity of an entity to be user editable.

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,216 @@
...
+
+The sample app demonstrates the use of CORS in the Call Web API component (*Pages/CallWebAPI.razor*).
+
+For more information, see <xref:security/cors>.
+
+## HttpClient and HttpRequestMessage with Fetch API request options
+
+Use [HttpClient](xref:fundamentals/http-requests) and <xref:System.Net.Http.HttpRequestMessage> to customize requests.
+
+In the following example:
```
danroth27: Remove?

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,216 @@
...
+
+Browser security prevents a web page from making requests to a different domain than the one that served the web page. This restriction is called the *same-origin policy*. The same-origin policy prevents a malicious site from reading sensitive data from another site. Sometimes, you might want to allow other sites make cross-origin resource sharing (CORS) requests to your app.
+
+The sample app demonstrates the use of CORS in the Call Web API component (*Pages/CallWebAPI.razor*).
+
+For more information, see <xref:security/cors>.
+
+## HttpClient and HttpRequestMessage with Fetch API request options
+
+Use [HttpClient](xref:fundamentals/http-requests) and <xref:System.Net.Http.HttpRequestMessage> to customize requests.
```
danroth27: Maybe expand on this a little bit?

```suggestion
Use [HttpClient](xref:fundamentals/http-requests) and <xref:System.Net.Http.HttpRequestMessage> to customize requests. For example, you can specify the request URI, HTTP method, and any desired request headers.
```

aspnetcore/blazor/call-web-api.md
guardrex: Not yet. The request is in for it.

aspnetcore/blazor/call-web-api.md
guardrex: It was just a simplification over what the sample app does. I'll make this more like the sample app on the next commit.

guardrex:
You see where this all went wrong, *right?* :smile: ...

> `HttpClient is available as a preconfigured service in client-side Blazor but not server-side Blazor.

That was the hang-up. What I was doing is making it all about **_injected `HttpClient`_**.

I see what you want now, but it seems like if we're going to wait on the design for server-side (not mention it) and not show a workaround (e.g., `HttpClientHandler`), we should have an example of using the .NET Core implementation (just a class used by the component, right?) in the server-side scenario.

danroth27:
> I see what you want now, but it seems like if we're going to wait on the design for server-side (not mention it) and not show a workaround (e.g., HttpClientHandler), we should have an example of using the .NET Core implementation (just a class used by the component, right?) in the server-side scenario.

Sorry about my confusion, but what workaround are you referring to? How is `HttpClientHandler` relevant? On the server the story for making HTTP requests is no different than it is for any other ASP.NET Core app. If you want a DI based solution on the server you can use `IHttpClientFactory` and friends. So what class would you show?

guardrex:
In order to use `HttpClient`, you said that devs were 'jumping thru hoops.' I didn't know what devs were doing, and someone mentioned `HttpClientHandler` like I showed at https://github.com/aspnet/AspNetCore.Docs/pull/12477#issuecomment-494190766. I gave it a shot, and it does work (you know ... it gets u some `var client = new HttpClient(httpClientHandler)` action in the RC).

> just a class used by the component

Nevermind that ... I thought for sec that we wouldn't use the `IHttpClientFactory` directly in an RC but put it into a separate class. I see now that *yes* ... that's exactly what you want ... inject it into the RC. Gotcha.

So, the topic will go with `IHttpClientFactory`, and we can cross-link that content for almost all of the coverage. Perhaps, a quick example *might be nice*. Let me see how that composes. I hope that it would only take a short code block. If it looks unnecessary, then we can cut it.

danroth27:
> In order to use HttpClient, you said that devs were 'jumping thru hoops.'

I think what this may be referring to is that some folks have been adding an HttpClient as a service in their server-side Blazor apps to create some symmetry with client-side Blazor. But most of the time I don't think that's what you really want, because then in your server-side Blazor app you end up making HTTP requests back to your own server. Instead, if you want to use the same component code independent of where the app is hosted, you should create a service that abstracts away how the data is requested.

In Blazor server-side apps your making HTTP requests to third party service and the story is no different than it is for existing ASP.NET Core apps, so we should just say that and point to https://docs.microsoft.com/aspnet/core/fundamentals/http-requests. The `HttpClient` service setup in Blazor client-side apps is specifically setup for making requests back to the server of origin.

guardrex:
@danroth27 Recommend one more quick :eye: on this to confirm that the feedback updates are correct.

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,217 @@
...
+[View or download sample code](https://github.com/aspnet/AspNetCore.Docs/tree/master/aspnetcore/blazor/common/samples/) ([how to download](xref:index#how-to-download-a-sample))
+
+For Blazor client-side examples, see the following components in the sample app:
+
+* Call Web API (*Pages/CallWebAPI.razor*)
+* HTTP Request Tester (*Components/HTTPRequestTester.razor*)
+
+## HttpClient and JSON helpers
+
+In Blazor client-side apps, [HttpClient](xref:fundamentals/http-requests) is available as a preconfigured service for making requests back to the origin server. `HttpClient` and JSON helpers are also used to call third party web API service endpoints. `HttpClient` is implemented using the browser [Fetch API](https://developer.mozilla.org/docs/Web/API/Fetch_API) and is subject to its limitations, including enforcement of the same origin policy.
```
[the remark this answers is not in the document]
guardrex: I looked up "third party," and it wasn't hyphenated. AHD was offline, so I had to use whatever Google threw up there from `define:third party`. Do you still want that hyphen in there?

guardrex:
Thx @scottaddie :rocket:

WRT the constants situation: It breaks :boom: in a non-page component (the HTTP Request Tester component) (and Steve has it that way in the engineering repo version of that component), but it's :+1: in a page component (the Call Web API component). Therefore, I made it a constant only in the page component. Idk if that should be reported in engineering, but Dan is :ear: here and will let us know.

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,217 @@
...
+        var response = await Http.SendAsync(requestMessage);
+        var responseStatusCode = response.StatusCode;
+        var responseBody = await response.Content.ReadAsStringAsync();
+    }
+}
+```
+
+For more information on Fetch API options, see [MDN web docs: WindowOrWorkerGlobalScope.fetch():Parameters](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Parameters).
+
+When sending credentials (authorization cookies/headers) on CORS requests, allow the `Authorization` header when creating the CORS policy for CORS middleware. The following policy includes configuration for:
```
danroth27:
```suggestion
When sending credentials (authorization cookies/headers) on CORS requests, the `Authorization` header must be allowed by the CORS policy. The following policy includes configuration for:
```

aspnetcore/blazor/call-web-api.md
```diff
@@ -0,0 +1,217 @@
...
+    private long _id;
+
+    private async Task DeleteItem() =>
+        await Http.DeleteAsync($"api/todo/{_id}");
+}
+```
+
+## Cross-origin resource sharing (CORS)
+
+Browser security prevents a web page from making requests to a different domain than the one that served the web page. This restriction is called the *same-origin policy*. The same-origin policy prevents a malicious site from reading sensitive data from another site. Sometimes, you might want to allow other sites to make cross-origin resource sharing (CORS) requests to your app.
```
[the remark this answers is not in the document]
guardrex: Should we point the link to our own topic?

https://docs.microsoft.com/aspnet/core/security/cors

```
xref:security/cors
```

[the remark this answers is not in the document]
danroth27: The point of my comment here is that often you are trying to call a third-party HTTP endpoint that you don't have any control over and might not be implemented with .NET (:shock:).

guardrex:
@scottaddie Do u want to run a final :eye: over the updates?

danroth27:
:shipit:

Ozkr16:
> > In order to use HttpClient, you said that devs were 'jumping thru hoops.'
> 
> I think what this may be referring to is that some folks have been adding an HttpClient as a service in their server-side Blazor apps to create some symmetry with client-side Blazor. But most of the time I don't think that's what you really want, because then in your server-side Blazor app you end up making HTTP requests back to your own server. Instead, if you want to use the same component code independent of where the app is hosted, you should create a service that abstracts away how the data is requested.
> 
> In Blazor server-side apps your making HTTP requests to third party service and the story is no different than it is for existing ASP.NET Core apps, so we should just say that and point to https://docs.microsoft.com/aspnet/core/fundamentals/http-requests. The `HttpClient` service setup in Blazor client-side apps is specifically setup for making requests back to the server of origin.

Hey @danroth27, this is an extremely useful insight. When reading thought the documentation I was unable to find something like this. For people with an ASP.Net Core app development background (specially those of us who deal with microservices architectures), the Blazor Server HTTP Client situation can be confusing. Thanks for clearing this out.

PR #11881, opened 2019-04-07
Unclosed void elements in topics

guardrex:
Fixes #11404 

Cross-ref: https://github.com/aspnet/Mvc/pull/2956

I'm not hitting the doc sample apps here ... that would require hundreds of updated lines that I don't think we should do for 2.x apps. This just hits the topics.

@NTaylorMullen, do u want me to go over the templates and submit a PR to engineering to revise the templates for void elements? It will resolve the situation for upcoming 3.0 doc samples ... we haven't started on those yet. [EDIT] Or, let me know if you just want an issue for it to let engineering handle that work.

cc: @ygoe

PR #12484, opened 2019-05-21
Close void elements in Razor components

### LINKED ISSUE #12481: Found a malformed 'input' tag helper. Tag helpers must have a start and end tag or be self closing
The input tags must be self-closed

&lt;input bind="@StartDate" format-value="yyyy-MM-dd" /&gt;

---
#### Document Details

* Content: [Create and use Razor components](https://docs.microsoft.com/en-in/aspnet/core/blazor/components?view=aspnetcore-3.0#feedback)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #12481 

Engineering resolved unclosed void elements on https://github.com/aspnet/Mvc/pull/2956.

Docs have moved in that direction (for example, [Unclosed void elements in topics (aspnet/AspNetCore.Docs #11881](https://github.com/aspnet/AspNetCore.Docs/pull/11881)).

Is this a point-in-time problem for Blazor, or will they always require closure?

btw - There are two spots where `<base>` tags are still unclosed. I think those don't suffer from this issue; but if I'm incorrect, I can add those two on another commit here before we merge this. :ear:

### DIFF
--- a/aspnetcore/blazor/components.md
+++ b/aspnetcore/blazor/components.md
@@ -5,7 +5,7 @@ description: Learn how to create and use Razor components, including how to bind
 monikerRange: '>= aspnetcore-3.0'
 ms.author: riande
 ms.custom: mvc
-ms.date: 05/10/2019
+ms.date: 05/21/2019
 uid: blazor/components
 ---
 # Create and use Razor components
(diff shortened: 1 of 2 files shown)


danroth27:
I chatted with @NTaylorMullen @ajaybhargavb about this and we think this is a product issue. If the HTML spec allows for no end tag then we should too. Filled https://github.com/aspnet/AspNetCore/issues/10498.

But for the time being we can update the docs.

guardrex:
I'll open a follow-up docs issue tracking on that engineering issue and put it in the HOLD column on the project.

PR #12488, opened 2019-05-21
Add Blazor supported platforms topic

### LINKED ISSUE #12480: Document Browser support matrix for Blazor

guardrex:
Fixes #12480

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/supported-platforms?view=aspnetcore-3.0&branch=pr-en-us-12488)

### DIFF
--- a/aspnetcore/blazor/supported-platforms.md
+++ b/aspnetcore/blazor/supported-platforms.md
@@ -1,0 +1,27 @@
+---
+title: ASP.NET Core Blazor supported platforms
+author: guardrex
+description: Learn about the supported platforms for ASP.NET Core Blazor.
+monikerRange: '>= aspnetcore-3.0'
+ms.author: riande
+ms.custom: mvc
+ms.date: 05/21/2019
+uid: blazor/supported-platforms
+---
+# ASP.NET Core Blazor supported platforms
+
+By [Luke Latham](https://github.com/guardrex)
+
+## Server system requirements
+
+[Blazor server-side](xref:blazor/hosting-models#server-side) supports any server platform that ASP.NET Core supports.
+
+## Browser requirements
+
+| Browser                          | Version |
+| -------------------------------- | :-----: |
+| Microsoft Edge                   | current |
+| Mozilla Firefox                  | current |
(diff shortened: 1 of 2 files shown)


aspnetcore/blazor/supported-platforms.md
```diff
@@ -0,0 +1,27 @@
...
+ms.author: riande
+ms.custom: mvc
+ms.date: 05/21/2019
+uid: blazor/supported-platforms
+---
+# ASP.NET Core Blazor supported platforms
+
+By [Luke Latham](https://github.com/guardrex)
+
+## Server system requirements
```
danroth27: I know this section was copied from the SignalR supported platforms doc, but I think it's self-evident and can be removed. Blazor is part of ASP.NET Core, so it should go without saying that it works wherever ASP.NET Core works.

aspnetcore/blazor/supported-platforms.md
```diff
@@ -0,0 +1,27 @@
...
+
+[Blazor server-side](xref:blazor/hosting-models#server-side) supports any server platform that ASP.NET Core supports.
+
+## Browser requirements
```
danroth27: This browser matrix is actually specific to server-side Blazor.

For client-side Blazor we don't support IE at all, because it doesn't support WebAssembly.

aspnetcore/blazor/supported-platforms.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Blazor server-side
+
+| Browser                          | Version |
+| -------------------------------- | :-----: |
+| Microsoft Edge                   | Current |
+| Mozilla Firefox                  | Current |
+| Google Chrome, including Android | Current |
+| Safari, including iOS            | Current |
+| Microsoft Internet Explorer      | 11      |
```
danroth27: Let's put a asterix  next to IE11 and add a note that IE11 will require adding additional polyfills.

guardrex:
@danroth27 The advice is lacking in the sense that it doesn't list *explicitly* the polyfills required for IE11. Is there no such list that we can provide here?

PR #12526, opened 2019-05-24
Configuration namespaces and line lengths

### LINKED ISSUE #12162: Improve package/namespace visibility in the Configuration topic
* Call out and link the packages for each provider.
* Specify the required namespaces in each bit of example code with code comments.
* Call out the packages provided by the `App` package ... or at least cross-link to the `App` package so that readers can find out which ones come in. I like cross-linking because it lightens the maintenance load when packages churn in the metapackage, so I'll try to go that way with it.

A PR by @vvimjam (https://github.com/aspnet/AspNetCore.Docs/pull/12160) reminds us on the need to continue improving namespace and package visibility. Although the API Browser is helpful, we can save readers time. Also, not everyone is aware of using the API Browser to get the info. Thanks @vvimjam! :rocket:

---
#### Document Details

* Content: [Configuration in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-2.2#memory-configuration-provider)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #12162 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/index?view=aspnetcore-2.1&branch=pr-en-us-12526)

* Turns out that the topic already does a good job on package refs. Minor language fix for the `App` package at the top.
* Add the namespace call outs.
* Shorten long code lines for display.

### DIFF
--- a/aspnetcore/fundamentals/configuration/index.md
+++ b/aspnetcore/fundamentals/configuration/index.md
@@ -5,7 +5,7 @@ description: Learn how to use the Configuration API to configure an ASP.NET Core
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 03/11/2019
+ms.date: 05/24/2019
 uid: fundamentals/configuration/index
 ---
 # Configuration in ASP.NET Core
@@ -22,11 +22,11 @@ App configuration in ASP.NET Core is based on key-value pairs established by *co
 * In-memory .NET objects
 * Settings files
 
+Configuration packages for common configuration provider scenarios are included in the [Microsoft.AspNetCore.App metapackage](xref:fundamentals/metapackage-app).
+
 The *options pattern* is an extension of the configuration concepts described in this topic. Options uses classes to represent groups of related settings. For more information on using the options pattern, see <xref:fundamentals/configuration/options>.
 
 [View or download sample code](https://github.com/aspnet/AspNetCore.Docs/tree/master/aspnetcore/fundamentals/configuration/index/samples) ([how to download](xref:index#how-to-download-a-sample))
-
-These three packages are included in the [Microsoft.AspNetCore.App metapackage](xref:fundamentals/metapackage-app).
 
 ## Host vs. app configuration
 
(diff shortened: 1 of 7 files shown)


aspnetcore/fundamentals/configuration/index.md
```diff
@@ -190,6 +190,8 @@ Call <xref:Microsoft.Extensions.Hosting.HostBuilder.ConfigureAppConfiguration*>
 `AddCommandLine` has already been called by `CreateDefaultBuilder`. If you need to provide app configuration and still be able to override that configuration with command-line arguments, call the app's additional providers in <xref:Microsoft.Extensions.Hosting.HostBuilder.ConfigureAppConfiguration*> and call `AddCommandLine` last.
 
 ```csharp
+// using Microsoft.Extensions.Configuration;
```
tdykstra: Why is this commented out?  When it has to be because it's out of place in the source code in order to be included in a snippet, there's no choice but to comment out, but I'm not sure it makes sense when it's unnecessary because it's inline code in the Markdown file.

aspnetcore/fundamentals/configuration/index.md
```diff
@@ -1023,6 +1072,8 @@ In the sample app, a POCO class is available to hold the bound configuration dat
 The configuration data is bound to the object:
 
 ```csharp
+// using Microsoft.Extensions.Configuration;
```
tdykstra: This is repeated throughout -- you might consider including it only the first one or two times and omitting it from the rest.

guardrex:
@tdykstra It's a long topic, so I think readers may hit a section from the sidebar TOC or a cross-link that lands anywhere here. Then, they don't have that experience of having seen it near the top.

WRT commenting out ... *yes* ...

* It has to be or the IDE whines about it ... 'unnecessary namespace'.
* It's also out-of-place, and I kind'a didn't want new devs or new-to-Core devs to think that this is where these `using` statements go.

I really don't prefer to do it this way. I don't think that the docs should have these *anywhere*, but that's not really the call that's been made by the team (perhaps when you were in AzureLand). We have samples, and the devs have the API Browser. At some point ... new or not ... from another framework or not ... *devs gotta learn how to be devs*. You just have to look crap up when you're a programmer imo. I mean the sample is **_right there_** :eyes:. They just have to look at it.

So ... anyway ... what would you prefer to do? I don't have strong feeling either way. I did put these everywhere because we have no idea where these folks are coming from ... a cross-link that lands them right smack in the middle or sidebar TOC jumping down as soon as the topic loads.

tdykstra:
My preference would be to omit the obvious ones and include only those that would be difficult to look up or figure out if you didn't know them.

guardrex:
I agree ... and I think that none of them are ever really difficult to look up when the sample is right there. *LazyDevs*:tm: :smile:

I'll update after lunch and ping back.

tdykstra:
If they're all in the sample, I'd omit them, and include usings only in beginner tutorials.

guardrex:
@tdykstra kk ... this is mostly covering line lengths now + the bit at the top on the `App` package + `using` in use.

PR #12559, opened 2019-05-28
Clarify location and name of the MvcMovieContext

biagidp:
The build for the migrations fails when following the original instructions to place the MvcMovieContext.cs file in the Models directory. Further down the in the documentation it specifies that this file is expected in the Data directory.


Rick-Anderson:
@biagidp you need to open an issue for this first and get it approved. I believe the scaffolder invoked from VS works differently than following the instructions for VSC/VS Mac.

This is too big of a change to not undergo testing.
We've got literally millions of folks reporting success with what we have.

PR #12560, opened 2019-05-28
fix: DbContext class path error

mrnixe:


Rick-Anderson:
@mrnixe thanks for your contribution.

PR #12378, opened 2019-05-10
Add note about keeping the middleware pipeline running for a WebSocket

anurse:
Related to https://github.com/aspnet/AspNetCore/issues/2854

I think I got the API doc xrefs right but we'll see :).

[Internal Review Version](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/websockets?view=aspnetcore-1.1&branch=pr-en-us-12378)

PR #12620, opened 2019-05-31
Add WebSocket lifetime info

### LINKED ISSUE #5466: Add guidance for WebSocket lifetime and the middleware pipeline
See https://github.com/aspnet/Home/issues/2462

We see customers having a frequent issue with WebSockets, where they do something like this:

```csharp
app.Use(async (context, next) => {
    var socket = await context.WebSockets.AcceptWebSocketAsync();
    BackgroundSocketProcessor.AddSocket(socket); 
});
```

`BackgroundSocketProcessor` is some service or component in their app that adds the socket to a collection. A background thread, or other process (other requests, etc) are then accessing those sockets and sending data. This is a common pattern and pretty much how SignalR works, so it's not entirely wrong :). However, there's a problem with this as written. After adding the socket to the background processor queue, the middleware function **returns**. This completes the request and causes Kestrel to conclude the Response and close the stream. As a result, when the background processor tries to send/receive on the socket, it throws an `ObjectDisposedException`. The message is obscure (though it will be clarified soon: https://github.com/aspnet/KestrelHttpServer/issues/2299 :tada: and https://github.com/aspnet/Home/issues/2854) and difficult to understand. We should include documentation in our WebSockets topic about how to identify this problem and deal with it.

Suggested outline.

* New Subtopic on https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/websockets.md titled "WebSocket Lifetime" (does not need to be a new page, just a subheading should be sufficient)
    * The WebSocket received from `AcceptWebSocketAsync` lives only as long as the Middleware pipeline for that request is still running. Once the middleware pipeline ends, the socket becomes broken.
    * Suggest providing some sample exception messages, to help searchability when the exception occurs :). Easy to repro by just having a background thread try to send to one of these sockets after the middleware pipeline ends (a timer would do it).
    * If you are going to use the WebSocket from outside the request context (in a background thread or from another request) you need to ensure that the original WebSocket request does not complete until you are finished with the socket. A sample pattern for this is given below (I suggest expanding it to something more functional in the actual doc, I'm just putting a rough sketch in here right now)
    * This applies to MVC controller actions too. If you accept a socket in the action, you need to make sure you wait for the socket to complete before returning from the Action method
    * As always, **never** use `Task.Wait()` or similar blocking calls to wait for the socket to complete, it will cause serious threading issues in the application, **always** use `await`.

```csharp
app.Use(async (context, next) => {
    var socket = await context.WebSockets.AcceptWebSocketAsync();
    var socketFinishedTcs = new TaskCompletionSource<object>();

    // The background processor should call socketFinishedTcs.TrySetResult(null) when it has finished with the socket
    BackgroundSocketProcessor.AddSocket(socket, socketFinishedTcs); 

    // Wait for the BackgroundSocketProcessor to finish with the socket before concluding the request.
    await socketFinishedTcs.Task;
});
```

tdykstra:
Fixes #5466 

The bulk of #5466 was done in #12378, but this PR adds a few notes from #5466 that weren't included in the earlier PR. It also reorganizes a bit to promote H3s to H2s and removes 1.1 sections.

[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/websockets?branch=pr-en-us-12620)

### DIFF
--- a/aspnetcore/fundamentals/websockets.md
+++ b/aspnetcore/fundamentals/websockets.md
@@ -14,7 +14,9 @@ By [Tom Dykstra](https://github.com/tdykstra) and [Andrew Stanton-Nurse](https:/
 
 This article explains how to get started with WebSockets in ASP.NET Core. [WebSocket](https://wikipedia.org/wiki/WebSocket) ([RFC 6455](https://tools.ietf.org/html/rfc6455)) is a protocol that enables two-way persistent communication channels over TCP connections. It's used in apps that benefit from fast, real-time communication, such as chat, dashboard, and game apps.
 
-[View or download sample code](https://github.com/aspnet/AspNetCore.Docs/tree/master/aspnetcore/fundamentals/websockets/samples) ([how to download](xref:index#how-to-download-a-sample)). See the [Next steps](#next-steps) section for more information.
+[ASP.NET Core SignalR](xref:signalr/introduction) is a library that simplifies adding real-time web functionality to apps. It uses WebSockets whenever possible.
+
+[View or download sample code](https://github.com/aspnet/AspNetCore.Docs/tree/master/aspnetcore/fundamentals/websockets/samples) ([how to download](xref:index#how-to-download-a-sample)). [How to run](#sample-app).
 
 ## Prerequisites
 
@@ -85,29 +87,7 @@ System.Net.WebSockets.WebSocketException (0x80004005): The remote party closed t
 Object name: 'HttpResponseStream'.
 ```
 
-If you're using a background service to write data to a WebSocket, make sure you keep the middleware pipeline running. Do this by using a <xref:System.Threading.Tasks.TaskCompletionSource%601>. Pass the `TaskCompletionSource` to your background service and have it call <xref:System.Threading.Tasks.TaskCompletionSource%601.TrySetResult%2A> when you finish with the WebSocket. Then `await` the <xref:System.Threading.Tasks.TaskCompletionSource%601.Task> property during the request.
-
-## Send and receive messages
-
-The `AcceptWebSocketAsync` method upgrades the TCP connection to a WebSocket connection and provides a [WebSocket](/dotnet/core/api/system.net.websockets.websocket) object. Use the `WebSocket` object to send and receive messages.
-
-The code shown earlier that accepts the WebSocket request passes the `WebSocket` object to an `Echo` method. The code receives a message and immediately sends back the same message. Messages are sent and received in a loop until the client closes the connection:
-
-[!code-csharp[](websockets/samples/2.x/WebSocketsSample/Startup.cs?name=Echo)]
-
-When accepting the WebSocket connection before beginning the loop, the middleware pipeline ends. Upon closing the socket, the pipeline unwinds. That is, the request stops moving forward in the pipeline when the WebSocket is accepted. When the loop is finished and the socket is closed, the request proceeds back up the pipeline.
-
-## WebSocket lifetime
-
-The WebSocket that is received from `AcceptWebSocketAsync` lives only as long as the middleware pipeline for the request is still running. When the middleware pipeline ends, the socket is closed. If your code continues to try to send and receive messages, the following `WebSocketException` exception is thrown:
-
-> The remote party closed the WebSocket connection without completing the close handshake.
-
-The exception has an `ObjectDisposedException` inner exception with the following message:
-
-> Cannot write to the response body, the response has completed. Object name: 'HttpResponseStream'.
-
-To avoid this error, don't let the request pipeline end before you're finished using the WebSocket. If you use a WebSocket in a background thread, use code like the following example:
+If you're using a background service to write data to a WebSocket, make sure you keep the middleware pipeline running. Do this by using a <xref:System.Threading.Tasks.TaskCompletionSource%601>. Pass the `TaskCompletionSource` to your background service and have it call <xref:System.Threading.Tasks.TaskCompletionSource%601.TrySetResult%2A> when you finish with the WebSocket. Then `await` the <xref:System.Threading.Tasks.TaskCompletionSource%601.Task> property during the request. For example:
 
 ```csharp
 app.Use(async (context, next) => {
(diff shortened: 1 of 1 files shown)


aspnetcore/fundamentals/websockets.md
```diff
@@ -148,13 +139,42 @@ When accepting the WebSocket connection before beginning the loop, the middlewar
 
 ::: moniker range=">= aspnetcore-2.2"
 
-### Handle client disconnects
+## WebSocket lifetime
+
+The WebSocket that is received from `AcceptWebSocketAsync` lives only as long as the middleware pipeline for the request is still running. When the middleware pipeline ends, the socket is closed. If your code continues to try to send and receive messages, a `WebSocketException` exception is raised with the following message:
```
Rick-Anderson: The WebSocket that is received from `AcceptWebSocketAsync` lives only as long as the middleware pipeline for the request is still running. When the middleware pipeline ends, the socket is closed.  The following `WebSocketException` exception is thrown when attempting to send and receive messages after the socket is closed:

aspnetcore/fundamentals/websockets.md
```diff
@@ -148,13 +139,42 @@ When accepting the WebSocket connection before beginning the loop, the middlewar
...
+
+> The remote party closed the WebSocket connection without completing the close handshake.
+
+The exception has an `ObjectDisposedException` inner exception with the following message:
+
+> Cannot write to the response body, the response has completed. Object name: 'HttpResponseStream'.
+
+To avoid this error, don't let the request pipeline end before you're finished using the WebSocket. If you use a WebSocket from outside the request context (in a background thread or from another request) use code like the following example to make sure the request pipeline doesn't end prematurely:
```
Rick-Anderson: > If you use a WebSocket from outside the request context (in a background thread or from another request) use code like the following example to make sure the request pipeline doesn't end prematurely:

That's too long. Shorten if possible. Maybe

When using a WebSocket from outside the request context:

* For example, in a background thread or from another request.
* Use code like the following to make sure the request pipeline doesn't end prematurely:

aspnetcore/fundamentals/websockets.md
```diff
@@ -148,13 +139,42 @@ When accepting the WebSocket connection before beginning the loop, the middlewar
...
+    var socket = await context.WebSockets.AcceptWebSocketAsync();
+    var socketFinishedTcs = new TaskCompletionSource<object>();
+
+    BackgroundSocketProcessor.AddSocket(socket, socketFinishedTcs); 
+
+    await socketFinishedTcs.Task;
+});
+```
+
+The `await socketFinishedTcs.Task` statement keeps the the request open until the background processor calls `socketFinishedTcs.TrySetResult(null)` when it finishes with the socket.
```
Rick-Anderson: The `await socketFinishedTcs.Task` statement keeps the the request open until the background processor calls `socketFinishedTcs.TrySetResult(null)` ~when it finishes with the socket~.

Rick-Anderson: The `await socketFinishedTcs.Task` statement keeps the the request open until the background processor calls `socketFinishedTcs.TrySetResult(null)`.

aspnetcore/fundamentals/websockets.md
```diff
@@ -148,13 +139,42 @@ When accepting the WebSocket connection before beginning the loop, the middlewar
...
+
+The same WebSocket closed error can happen if you return too soon from an MVC controller action method. If you accept a socket in the action method, make sure you wait for the code that uses the socket to complete before you return from the action method.
```
Rick-Anderson: Doesn't this apply to Razor methods too? If so, change to action method.

The same WebSocket closed error can happen if you return too soon from an ~MVC controller~ action method. If you accept a socket in the action method, make sure you wait for the code that uses the socket to complete before you return from the action method.

Rick-Anderson: The preceding `WebSocket` closed exception can happen by returning from an action method before code using the socket has completed. If the socket is accepted in an action method, wait for the code that uses the socket to complete before calling return.

aspnetcore/fundamentals/websockets.md
```diff
@@ -37,20 +37,11 @@ This article explains how to get started with WebSockets in ASP.NET Core. [WebSo
 
 * For supported browsers, see https://caniuse.com/#feat=websockets.
 
-## When to use WebSockets
+## NuGet package
```
Rick-Anderson: This is removed because SignalR is preferred? If so, why don't we say that in the intro?

Rick-Anderson: Replace all MVC action with action to include RP.

tdykstra: When to use is removed because it's already covered adequately in the intro, and omitting it lets me promote all the how-to section H3s to H2s.

Rick-Anderson: Consider adding 
SignalR is generally preferred over WebSockets
in the intro.

Rick-Anderson: That statement would need PU approval.

anurse: I think that's OK. Might be worth a small amount more detail there though, something like "For most applications, SignalR is generally preferred over raw WebSockets. SignalR provides transport fallback for environments where WebSockets is not available. It also provides a simple "remote procedure call" app model."

I don't have a clean phrasing for it but I'd also make it clear that SignalR should be mostly as efficient as raw WebSockets in almost all environments.

aspnetcore/fundamentals/websockets.md
```diff
@@ -37,35 +39,16 @@ This article explains how to get started with WebSockets in ASP.NET Core. [WebSo
...
-[ASP.NET Core SignalR](xref:signalr/introduction) is a library that simplifies adding real-time web functionality to apps. It uses WebSockets whenever possible.
-
-## How to use WebSockets
+## NuGet package
 
-* Install the [Microsoft.AspNetCore.WebSockets](https://www.nuget.org/packages/Microsoft.AspNetCore.WebSockets/) package.
-* Configure the middleware.
-* Accept WebSocket requests.
-* Send and receive messages.
+Install the [Microsoft.AspNetCore.WebSockets](https://www.nuget.org/packages/Microsoft.AspNetCore.WebSockets/) package.
```
anurse: WebSockets is in the shared framework, so it's automatically available in all ASP.NET Core 2.1+ apps. I'd remove this section entirely. Or at least make it 2.0 or lower.

aspnetcore/fundamentals/websockets.md
```diff
@@ -124,37 +87,41 @@ System.Net.WebSockets.WebSocketException (0x80004005): The remote party closed t
...
+    BackgroundSocketProcessor.AddSocket(socket, socketFinishedTcs); 
 
-The code shown earlier that accepts the WebSocket request passes the `WebSocket` object to an `Echo` method. The code receives a message and immediately sends back the same message. Messages are sent and received in a loop until the client closes the connection:
+    await socketFinishedTcs.Task;
+});
+```
+The WebSocket closed exception can also happen if you return too soon from an action method. If you accept a socket in an action method, wait for the code that uses the socket to complete before returning from the action method.
 
-::: moniker range=">= aspnetcore-2.0"
+Never use `Task.Wait()` or similar blocking calls to wait for the socket to complete, as that can cause serious threading issues. Always use `await`.
```
anurse:
```suggestion
Never use `Task.Wait()`, `Task.Result`, or similar blocking calls to wait for the socket to complete, as that can cause serious threading issues. Always use `await`.
```

anurse: Seems worth calling out `Result` explicitly too as it's just as commonly seen in incorrect user code.

PR #12622, opened 2019-05-31
'detonate' a package? Really?

ChuckByram:
There is no open issue that this PR addresses. But it seems like bad form to have `detonate` as a package Action for a Microsoft sample.


guardrex:
That goes back to **_ancient ASP.NET Core times_** ...... in other words ... a couple of years ago. :smile:

https://github.com/aspnet/AspNetCore.Docs/blob/3b583593fbfa3d144c708ed30c4ce5778d7d0ac7/aspnetcore/fundamentals/routing/samples/1.x/RoutingSample/Startup.cs#L32

I'll mark this for the Product Unit to take a look at. I suspect that it came from an ancient sample that they generated, and they may know why that's there.

danroth27:
I have no idea how this got into the docs, but we should absolutely remove this.

PR #12645, opened 2019-06-03
typo: Double word "is"

nschonni:


guardrex:
@nschonni ... Can we really have enough *"is"* and *"the"* tho?!! :smile: lol

Actually, I can tell you now after a few years working on these how this happens. This isn't the first time, and it won't be the last: When working on a topic ... even proofreading it several times ... even when reviewers look ... if a line breaks right where the first word occurrance appears at the end of one line and the next occurrance of the word appears at the start of the next line .. :eye: :boom: .. **_the human eye can read right past it it._** :smile:

Thanks for spotting these and getting them *OUT*! :rocket:

PR #12647, opened 2019-06-03
fix: MD033/no-inline-html

nschonni:
Markdownlint changed the HTML regex and now flags some new elements.
These elements don't get swallowed like the others.

### DIFF
--- a/aspnetcore/tutorials/first-mongo-app.md
+++ b/aspnetcore/tutorials/first-mongo-app.md
@@ -264,7 +264,7 @@ The `BookService` class uses the following `MongoDB.Driver` members to perform C
 
     [!code-csharp[](first-mongo-app/sample/BooksApi/Services/BookService.cs?name=snippet_BookServiceConstructor&highlight=3)]
 
-* [IMongoDatabase](https://api.mongodb.com/csharp/current/html/T_MongoDB_Driver_IMongoDatabase.htm) &ndash; Represents the Mongo database for performing operations. This tutorial uses the generic [GetCollection<TDocument>(collection)](https://api.mongodb.com/csharp/current/html/M_MongoDB_Driver_IMongoDatabase_GetCollection__1.htm) method on the interface to gain access to data in a specific collection. Perform CRUD operations against the collection after this method is called. In the `GetCollection<TDocument>(collection)` method call:
+* [IMongoDatabase](https://api.mongodb.com/csharp/current/html/T_MongoDB_Driver_IMongoDatabase.htm) &ndash; Represents the Mongo database for performing operations. This tutorial uses the generic [GetCollection\<TDocument>(collection)](https://api.mongodb.com/csharp/current/html/M_MongoDB_Driver_IMongoDatabase_GetCollection__1.htm) method on the interface to gain access to data in a specific collection. Perform CRUD operations against the collection after this method is called. In the `GetCollection<TDocument>(collection)` method call:
   * `collection` represents the collection name.
   * `TDocument` represents the CLR object type stored in the collection.
 


guardrex:
By convention, we're escaping non-code-fenced < characters. We're not escaping > ... unless something changed. Let's ask Scott to take a look and confirm.

nschonni:
Right now, the part you undid wasn't escaping the `<`, it was escaping the `\`. I just did it at the end instead of a third stacked backtick. I agree that mostly the front angle bracket is what is escaped though.

guardrex:
lol ..... ooooooops! ☕️ ... definitely missing a cup of ☕️.

guardrex:
This one looks ok ...

https://docs.microsoft.com/aspnet/core/host-and-deploy/iis/#application-pool-identity

I'll check the other one next.

guardrex:
This looks ok, too ...

https://docs.microsoft.com/aspnet/core/mvc/views/razor#inspect-the-razor-c-class-generated-for-a-view

Which change did I make tho that you think is going to break? (and keep in mind that I *still* haven't had another cup of :coffee: yet :smile:).

guardrex:
I see :eyes: what you mean ...

```
1. Enter **IIS AppPool\\<app_pool_name>** in **Enter the object names to select** area. Select the 
**Check Names** button. For the *DefaultAppPool* check the names using **IIS 
AppPool\DefaultAppPool**. When the **Check Names** button is selected, a value of 
**DefaultAppPool** is indicated in the object names area. It isn't possible to enter the app 
pool name directly into the object names area. Use the **IIS AppPool\\<app_pool_name>** 
format when checking for the object name.
```

The double is making the single backslash possible.

I'll revert that and then do the triple given that we're escaping them.

guardrex:
ok .... I think that will work.

Let's have that ☕️ now before I break something else this morning. :smile: Thanks for your patience. Scott will let me know if anything has changed WRT our repo conventions for escapes.

aspnetcore/host-and-deploy/iis/index.md
```diff
@@ -518,7 +518,7 @@ If the IIS worker process requires elevated access to the app, modify the Access
 
 1. Select the **Locations** button and make sure the system is selected.
 
-1. Enter **IIS AppPool\\<app_pool_name>** in **Enter the object names to select** area. Select the **Check Names** button. For the *DefaultAppPool* check the names using **IIS AppPool\DefaultAppPool**. When the **Check Names** button is selected, a value of **DefaultAppPool** is indicated in the object names area. It isn't possible to enter the app pool name directly into the object names area. Use the **IIS AppPool\\<app_pool_name>** format when checking for the object name.
+1. Enter **IIS AppPool\\\<app_pool_name>** in **Enter the object names to select** area. Select the **Check Names** button. For the *DefaultAppPool* check the names using **IIS AppPool\DefaultAppPool**. When the **Check Names** button is selected, a value of **DefaultAppPool** is indicated in the object names area. It isn't possible to enter the app pool name directly into the object names area. Use the **IIS AppPool\\\<app_pool_name>** format when checking for the object name.
```
tdykstra: Why are three backslashes needed here? The original with two seems to render correctly as one.

guardrex: We were (last I heard ... it might be a 🧀-moved scenario for me) that we were escaping the < char, too. Therefore, the first `\\` yields a `\` and the `\<` yields a `<`. Net result ... `\<`.

guardrex: ... and I agree on what's going on with the rendering. It looks like that first `\` is really coming out as `\`. Therefore, I assume that `\` is fine to get a `\` based on the rendered behavior ... but didn't we have problems with that at one point? I thought that unless it went in as `\\` it didn't work. Perhaps, there's an odd scenario here where if the `\\<` in just this one case works out but other `\` don't ... maybe some kind of parsing algorithm thing allows it to work.

guardrex: How does it look based on this PR?

https://review.docs.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/index?view=aspnetcore-2.1&branch=pr-en-us-12647#application-pool-identity

tdykstra: It renders correctly in both production and staging, which doesn't make sense to me, so I guess we could go either way ... Maybe the rendering engine is getting smarter about not swallowing unescaped  `<`  in which case some or all of the other changes in this PR may not be needed.

guardrex: Yes, I'm mostly just going by convention these days. I think Scott knows more about it. We talked about this ages ago. Check out the Example 224 here ...

https://spec.commonmark.org/0.15/#backslash-escapes

The only one that I wasn't doing was the `#` when it appears mid-sentence. I think it would only be a problem if it starts a line and we wanted it to show up ... and that never has happened AFAIK.

... and check out the next one ... Example 225 ... it seems to indicate to me if I'm reading that correctly that the correct way to go **_IS_** `\\\<`.

tdykstra: Not exactly - that just confirms that triple backslash will work, not that we need to do it.  Escaping is an option that is available to make things render the way we intend them to.  IF they render correctly without the escaping, we're just adding noise to the text.  I'd like to understand why the escaping isn't needed in these instances, though.  Does the rendering engine now swallow only valid HTML elements and everything else no longer needs escaping?

guardrex: > not that we need to do it

I think that we should escape all backslashes. In other cases (earlier ... months and months ago when discussed prior) I think we had to escape backslashes in cases where they had a non-alphanumeric char behind them.

I agree that the `\\<` working properly is *very strange* ...... 👽? lol I actually would expect that to render as a backslash followed by **_a tag_** that doesn't show up.

Another side note here is that discussions led to using the brace `{` with text instead instead of angle brackets and underscores. The thinking there is that we get our paths localized. So ... it was like ...

* `/\<A_PLACEHOLDER_WITH_UNDERSCORES>` :x: *Bad*
* `/{A PLACEHOLDER WITHOUT UNDERSCORES}` ✔️ *Good*

... when not code-fenced, it should (in theory 🤞) localize well.

tdykstra: If the markdownlint rule is reporting false positives, another option is to repurpose this PR to make it disable the rule instead.  I'm not sure the benefit of this rule outweighs the nuisance factor, as any time we use inline HTML we'll find out quickly enough if the rendering engine has a problem with it.

scottaddie: I'm in agreement with Tom.

guardrex: @scottaddie ... but WRT the backslash: Are there conventions/rules/scenarios to escape and not to escape, or was it decided back then that we're always going to escape it?

nschonni: Looks like there is a new release of Markdownlint that changed the regex https://github.com/DavidAnson/markdownlint/issues/188
I'll wait till the CLI library is updated and re-run this

nschonni:
New version of markdownlint is out. Most of the changes went away except for a single instance that was missing the escape

PR #12708, opened 2019-06-05
Add TestServer doc note about non-essential cookies

scottaddie:
Fixes https://github.com/aspnet/AspNetCore.Docs/issues/12688

**Summary of changes**
- Clarify how consent is provided in a template-generated project (by clicking the **Accept** button).
- Clarify what consent is provided (cookie storage).
- Move GDPR doc's non-essential cookies **Tempdata provider and session state cookies are not essential** section within the **Essential cookies** section. This makes linking easier from the Integration Tests doc.
- Replaced occurrences of *folder* with *directory* in the Integration Tests doc. Both terms were being used.

### DIFF
--- a/aspnetcore/security/gdpr.md
+++ b/aspnetcore/security/gdpr.md
@@ -29,8 +29,8 @@ The [sample app](https://github.com/aspnet/AspNetCore.Docs/tree/live/aspnetcore/
 Razor Pages and MVC projects created with the project templates include the following GDPR support:
 
 * [CookiePolicyOptions](/dotnet/api/microsoft.aspnetcore.builder.cookiepolicyoptions) and [UseCookiePolicy](/dotnet/api/microsoft.aspnetcore.builder.cookiepolicyappbuilderextensions.usecookiepolicy) are set in the `Startup` class.
-* The *_CookieConsentPartial.cshtml* [partial view](xref:mvc/views/tag-helpers/builtin-th/partial-tag-helper). An **Accept** button is included in this file. When the user clicks the **Accept** button, consent to store cookies is provided.
-* The *Pages/Privacy.cshtml* page or *Views/Home/Privacy.cshtml* view provides a page to detail your site's privacy policy. The *_CookieConsentPartial.cshtml* file generates a link to the Privacy page.
+* The *\_CookieConsentPartial.cshtml* [partial view](xref:mvc/views/tag-helpers/builtin-th/partial-tag-helper). An **Accept** button is included in this file. When the user clicks the **Accept** button, consent to store cookies is provided.
+* The *Pages/Privacy.cshtml* page or *Views/Home/Privacy.cshtml* view provides a page to detail your site's privacy policy. The *\_CookieConsentPartial.cshtml* file generates a link to the Privacy page.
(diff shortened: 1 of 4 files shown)


aspnetcore/security/gdpr.md
```diff
@@ -57,15 +57,15 @@ This partial:
...
 
-If consent has not been given, only cookies marked essential are sent to the browser. The following code makes a cookie essential:
+If consent to store cookies hasn't been provided, only cookies marked essential are sent to the browser. The following code makes a cookie essential:
 
 [!code-csharp[Main](gdpr/sample/RP/Pages/Cookie.cshtml.cs?name=snippet1&highlight=5)]
 
 <a name="tempdata"></a>
 
-## Tempdata provider and session state cookies are not essential
+### TempData provider and session state cookies are not essential
```
guardrex: AFAIK we're using contractions in headings. It was just styles that we were avoiding, but the 🧀 may have moved on me.

```suggestion
### TempData provider and session state cookies aren't essential
```

Side-note: There are several unescaped underscores north of this causing the emphasis in this view. Recommend fixing those.

aspnetcore/test/integration-tests.md
```diff
@@ -78,7 +78,7 @@ Infrastructure components, such as the test web host and in-memory test server (
 
 The `Microsoft.AspNetCore.Mvc.Testing` package handles the following tasks:
 
-* Copies the dependencies file (*\*.deps*) from the SUT into the test project's *bin* folder.
+* Copies the dependencies file (*\*.deps*) from the SUT into the test project's *bin* directory.
```
guardrex: Was there a 🧀 move on me? "Folder" was welcome from the start in docs. Was it determined that "directory" is more xplat friendly? If so, that makes sense. I'll start going that way.

scottaddie: The official guidance is:

"Use directory only in content for developers when you need to refer to the structure of the file system or to match the API."

Since this is developer content, and we're talking about file system structure, I used *directory*.

guardrex: They may have changed it ... my copy is old. My old version has another line there before it says that. It says this leading into that .....

> Most of the time, use folder instead. 

There are *many* examples in the official guidance of them using "folder."

e.g ....

> 5.5.22	child folder
> Don't use. Use subfolder, subdirectory, folder, or the name of the folder.

> 5.6.21	destination
> Use a more precise term, such as website or folder, to describe an end point, such as the location reached when a customer clicks a hyperlink, the folder to which a file is copied or moved, or the document in which a linked or embedded object is stored. Use destination as a modifier only if necessary for clarity.

The reason that I said it might be better for xplat is that I recalled this remark ...

> 5.8.20	folder, folder icon
> In general, try not to talk about UI. Instead, talk about what the customer should do. 
In content about Windows, use folder to refer to a container for files and other folders. Use directory only in content for developers when you need to refer to the structure of the file system or to match the API. For MS-DOS, use directory. 

... it's the little bit there about "In content about Windows."

This is a break from what we've been doing. Let me know if it's a new repo rule ... "folder" is all over the place to the tune of **_>700 instances of the word_** across the docs.

PR #12728, opened 2019-06-06
Disable auto-start of OmniSharp VS Code extension

scottaddie:
Fixes https://github.com/aspnet/AspNetCore.Docs/issues/12713

Prevent auto-start of the OmniSharp VS Code extension when the root workspace folder (_AspNetCore.Docs_) is opened. Continue to allow auto-start of the extension when the _aspnetcore_ folder is opened to support working with the sample apps.


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "omnisharp.autoStart": false
+}
```
guardrex: LF here irrelevant really, but *red things* scare me. :smile: lol

PR #12780, opened 2019-06-10
Remove anchor link from using-data-protection

spottedmahn:
The anchor doesn't seem to add any value except for hiding the title of the page 😜.  The link is essentially to the top of the document.


scottaddie:
Great catch! Thank you for the PR, @spottedmahn.

PR #12799, opened 2019-06-11
C95509: Escaping "|"

v-makoud:


guardrex:
@v-makoud Can you provide the link to the live topic that has bad rendering?

Because that vertical bar is inline code, it's rendering ok here: https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-2.2#forwarded-headers-middleware-options

If we do need to apply a fix, I recommend a different approach that changes the character outside of our build system (`&vert` with `<code>` tags):

> \<code>ForwardedHeaders.XForwardedFor \&vert; ForwardedHeaders.XForwardedProto\</code>

v-makoud:
#sign-off

guardrex:
Do you have the link to the localized topic that was reported by the reader?

guardrex:
@v-makoud Let's not do this unless we take a look at the localized topic and see how it breaks there. This change shouldn't be necessary ... *in theory*. :smile: lol

Let me know if you get the link to the localized topic so that we can take a closer look. If we need to do something, we can re-open this PR and continue.

PR #12806, opened 2019-06-11
Elaborate how view files are selected

mrmowji:
When no name parameter is provided to a `View()` method, a view file with the same name of the caller action will be used, not an `Index.cshtml` view file by default.


Rick-Anderson:
@mrmowji  thanks!

PR #12823, opened 2019-06-12
Workaround designer issue in VS

anurse:
Because it inherits from `ServiceBase` (via `WebHostService`), VS treats this `CustomWebHostService` class like a "Component" and tries to open a design surface for it. This causes problems because the designer can't load properly over this type (it has a constructor). We are fixing `WebHostService` in 3.0 (and it's no longer recommended anyway, with Worker Service), but the 2.x sample should be updated to lead people down the happy path.

Addresses aspnet/AspNetCore#11114 for 2.1/2.2.


anurse:
I did wonder why this was surfacing now. Thanks for the clarification!

PR #12859, opened 2019-06-13
fix: MD038/no-space-in-code

nschonni:
Spaces inside code span elements


Rick-Anderson:
We're not accepting PR's that only fix MD warnings. Warnings like these can be fixed next time the doc is updated.

cc @tdykstra @nschonni  @guardrex  @scottaddie @wadepickett

PR #12872, opened 2019-06-13
Blazor prerendering + reconnect

### LINKED ISSUE #11836: [Preview 5] Prerendering with reconnect
We should add docs for the prerendering + reconnect feature.
* Describe how it works.
* Define the limitations.
* Talk about scale-out.

guardrex:
Fixes #11836 

[Internal Review Topic (Component rendering modes section)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/components?view=aspnetcore-3.0&branch=pr-en-us-12872#component-rendering-modes)


javiercn:
The problem is that it is not clear that you can also opt to render server side components in a “detached” fashion. 

I like introducing terminology because it standardizes around the capabilities of the scenario vs the actual technical implementation

javiercn:
See for example how google does it in their docs which I found enlightening at the time https://developers.google.com/web/updates/2019/02/rendering-on-the-web#csr

danroth27:
> The problem is that it is not clear that you can also opt to render server side components in a “detached” fashion.

I assume you're referring to statically rendering a component from a page or view (see what I just did there? :smiley:). That would be covered in the topic on "Using components with Razor Pages and MVC Views".

javiercn:
@danroth27 I leave it to your best judgement then 😄

guardrex:
@danroth27 Do you want me to use your outline :point_up: for the content? ... in the Components topic where this PR places the content?

danroth27:
@guardrex I'd like to create two new completely new top level topics:

- Prerendering
- Using components with Pages and Views

The content in this PR would then be covered in those new topics, but in the context of those specific scenarios and without new jargon.

Do you want me to draft up the content/outline for these topics? Or do you want to just give it a shot?

guardrex:
🤔 ... I think it's better if you draft it. If I do it, it will probably require a lot of PR churn in review. If you do it, then I just need to focus on the *grammarishy whatnots*:tm:. :smile:

If we go that way and you provide the content, let's put it back over on the issue at https://github.com/aspnet/AspNetCore.Docs/issues/11836 and close this.

guardrex:
... and note that I'll take care of cross-links for the current prerendering content. There's such content in the Hosting Models and JS Interop topics.

guardrex:
Closing ... I work the content provided on the main issue into a new PR.

PR #12874, opened 2019-06-14
 Update dotnet-client.md to discuss auto reconnect

halter73:
This auto-reconnect sections of the dotnet-client doc closely follows the structure of the same sections in javascript-client.md

### DIFF
--- a/aspnetcore/signalr/dotnet-client.md
+++ b/aspnetcore/signalr/dotnet-client.md
@@ -33,6 +33,160 @@ To establish a connection, create a `HubConnectionBuilder` and call `Build`. The
 
 ## Handle lost connection
 
+::: moniker range=">= aspnetcore-3.0"
+
+### Automatically reconnect
+
+The <xref:Microsoft.AspNetCore.SignalR.Client.HubConnection> can be configured to automatically reconnect using the `WithAutomaticReconnect` method on the <xref:Microsoft.AspNetCore.SignalR.Client.HubConnectionBuilder>. It won't automatically reconnect by default.
+
+```csharp
+HubConnection connection= new HubConnectionBuilder()
+    .WithUrl(new Uri("http://127.0.0.1:5000/chatHub"))
+    .WithAutomaticReconnect()
+    .Build();;
+```
+
+Without any parameters, `WithAutomaticReconnect()` configures the client to wait 0, 2, 10, and 30 seconds respectively before trying each reconnect attempt, stopping after four failed attempts.
+
+Before starting any reconnect attempts, the `HubConnection` will transition to the `HubConnectionState.Reconnecting` state and fire the `Reconnecting` event instead of transitioning to the `Disconnected` state and triggering the `Closed` callback like a `HubConnection` without automatic reconnect configured. This provides an opportunity to warn users that the connection has been lost and to disable UI elements. Non-interactive apps can start queuing or dropping messages instead.
+
+```csharp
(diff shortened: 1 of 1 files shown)


aspnetcore/signalr/dotnet-client.md
```diff
@@ -33,6 +33,160 @@ To establish a connection, create a `HubConnectionBuilder` and call `Build`. The
...
+```
+
+If the client doesn't successfully reconnect within its first four attempts, the `HubConnection` will transition to the `Disconnected` state and fire the <xref:Microsoft.AspNetCore.SignalR.Client.HubConnection.Closed> event. This provides an opportunity to attempt to restart the connection manually or inform users the connection has been permanently lost.
+
+```csharp
+connection.Closed += error =>
+{
+    Debug.Assert(connection.State == HubConnectionState.Disconnected);
+
+    // Notify users the connection has been closed, manually try to restart the connection, etc...
```
tdykstra: The Style guide advises against using Latin abbreviations.  In this instance it could just be omitted and the code comment parallels the preceding text

```suggestion
    // Notify users the connection has been closed or manually try to restart the connection.
```

aspnetcore/signalr/dotnet-client.md
```diff
@@ -33,6 +33,160 @@ To establish a connection, create a `HubConnectionBuilder` and call `Build`. The
...
+    // .WithAutomaticReconnect(new[] { TimeSpan.Zero, TimeSpan.FromSeconds(2), TimeSpan.FromSeconds(10), TimeSpan.FromSeconds(30) }) yields the default behavior.
+```
+
+The preceding example configures the `HubConnection` to start attempting reconnects immediately after the connection is lost. This is also true for the default configuration.
+
+If the first reconnect attempt fails, the second reconnect attempt will also start immediately instead of waiting 2 seconds like it would in the default configuration.
+
+If the second reconnect attempt fails, the third reconnect attempt will start in 10 seconds which is again like the default configuration.
+
+The custom behavior then diverges again from the default behavior by stopping after the third reconnect attempt failure instead of trying one more reconnect attempt in another 30 seconds like it would in the default configuration.
```
tdykstra: Consider breaking up long sentence.

```suggestion
The custom behavior then diverges again from the default behavior by stopping after the third reconnect attempt failure. In the default configuration there would be one more reconnect attempt in another 30 seconds.
```

aspnetcore/signalr/dotnet-client.md
```diff
@@ -33,6 +33,160 @@ To establish a connection, create a `HubConnectionBuilder` and call `Build`. The
...
+        }
+        else
+        {
+            // If we've been reconnecting for more than 60 seconds so far, stop reconnecting.
+            return null;
+        }
+    }
+}
+
+HubConnection connection= new HubConnectionBuilder()
```
tdykstra: Consider making the following snippet a separate code block.  And should it be C# rather than JavaScript?

halter73:
> Our grammar and style tool Acrolinx advises us to avoid future tense when feasible

Noted. I'll try to avoid the future tense in the future.

Since this isn't a requirement, I will leave it as is for now, though I'm perfectly fine if someone wants to fix it. The JS version of this doc also uses future tense, so maybe those could be fixed together if someone finds it worthwhile.

If I tried to rush a fix, I'd probably end up mixing tenses more than the current version.

PR #12756, opened 2019-06-08
Update blazor docs to reflect directive attributes changes

ajaybhargavb:
Fixes https://github.com/aspnet/AspNetCore.Docs/issues/12578

More context about directive attributes [here](https://github.com/aspnet/AspNetCore/issues/6364)

PR #12893, opened 2019-06-16
Components topic code example updates

### LINKED ISSUE #12892: Components topic code example updates
* `_italicsCheck` and `CurrentValue` are used without a leading `@` in binding examples.
* Question for engineering can be asked on the review about inconsistent syntax for binding.

---
#### Document Details

* Content: [Create and use Razor components](https://docs.microsoft.com/en-us/aspnet/core/blazor/components?view=aspnetcore-3.0)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #12892

**Skip these remarks and access the diff. My confusion was cleared in the discussion.**

**This PR has the final set of updates that we need.**

RE: [Update blazor docs to reflect directive attributes changes (aspnet/AspNetCore.Docs \#12756)](https://github.com/aspnet/AspNetCore.Docs/pull/12756)

1. WRT the temporary (Pre6) need to place an `@` on a field/property value of an `@bind` directive. I looked at [Implement directive attribute feature for components (aspnet/AspNetCore \#6364)](https://github.com/aspnet/AspNetCore/issues/6364), but none of the description discusses the scenario.

   For example in the [Format strings section](https://docs.microsoft.com/aspnet/core/blazor/components?view=aspnetcore-3.0#data-binding) of the Components topic, updates didn't add the `@` to the value for `StartDate` ...

   ```cshtml
   <input @bind="StartDate" @bind:format="yyyy-MM-dd" />
   ```

   Same thing in the `ParentYear` examples (`@bind-{property}` syntax) ...

   ```cshtml
   <ChildComponent @bind-Year="ParentYear" />
   ```

    ... and should devs apply the `@`-value temporary rule to `@bind-value:event`/`bind-{property}:event`? It was left on the updates as ...

   ```cshtml
   <ChildComponent @bind-Year="ParentYear" @bind-Year:event="YearChanged" />
   ```

1. What's the difference between the following, if any:

   ```cshtml
   <input @bind="@CurrentValue" />
   ```

   ```cshtml
   <input @bind-value="@CurrentValue" @bind-value:event="onchange" />
   ```

   Is the former just a shorthand way of setting up `onchange` binding? If they are the same, it isn't considered better to have *Just One Consistent Way*:tm: :smile: to do it (the latter approach, which comports with other `on{event}` binding)?

### DIFF
--- a/aspnetcore/blazor/components.md
+++ b/aspnetcore/blazor/components.md
@@ -5,7 +5,7 @@ description: Learn how to create and use Razor components, including how to bind
 monikerRange: '>= aspnetcore-3.0'
 ms.author: riande
 ms.custom: mvc
-ms.date: 06/12/2019
+ms.date: 06/16/2019
 uid: blazor/components
 ---
 # Create and use Razor components
@@ -123,14 +123,14 @@ Data binding to both components and DOM elements is accomplished with the `@bind
 
 ```cshtml
 <input type="checkbox" class="form-check-input" id="italicsCheck" 
-    @bind="_italicsCheck" />
+    @bind="@_italicsCheck" />
 ```
 
 When the check box is selected and cleared, the property's value is updated to `true` and `false`, respectively.
 
 The check box is updated in the UI only when the component is rendered, not in response to changing the property's value. Since components render themselves after event handler code executes, property updates are usually reflected in the UI immediately.
 
-Using `@bind` with a `CurrentValue` property (`<input @bind="CurrentValue" />`) is essentially equivalent to the following:
+Using `@bind` with a `CurrentValue` property (`<input @bind="@CurrentValue" />`) is essentially equivalent to the following:
 
 ```cshtml
 <input value="@CurrentValue" 
@@ -142,7 +142,7 @@ When the component is rendered, the `value` of the input element comes from the 
 In addition to `onchange`, the property can be bound using other events like `oninput` by adding a `@bind` attribute with an `event` parameter:
 
 ```cshtml
-<input type="text" @bind-value="@CurrentValue" @bind-value:event="oninput" />
+<input @bind-value="@CurrentValue" @bind-value:event="oninput" />
 ```
 
 Unlike `onchange`, `oninput` fires for every character that is input into the text box.
--- a/aspnetcore/blazor/components.md
+++ b/aspnetcore/blazor/components.md
@@ -145,7 +145,7 @@ In addition to handling `onchange` events with `@bind` syntax, a property or fie
 <input @bind-value="CurrentValue" @bind-value:event="oninput" />
 ```
 
-Unlike `onchange`, `oninput` fires for every character changed in the text box.
+Unlike `onchange`, which fires when the element loses focus, `oninput` fires when the value of the text box changes.
 
 **Format strings**
 
--- a/aspnetcore/blazor/components.md
+++ b/aspnetcore/blazor/components.md
@@ -123,14 +123,14 @@ Data binding to both components and DOM elements is accomplished with the `@bind
 
 ```cshtml
 <input type="checkbox" class="form-check-input" id="italicsCheck" 
-    @bind="@_italicsCheck" />
+    @bind="_italicsCheck" />
 ```
 
 When the check box is selected and cleared, the property's value is updated to `true` and `false`, respectively.
 
 The check box is updated in the UI only when the component is rendered, not in response to changing the property's value. Since components render themselves after event handler code executes, property updates are usually reflected in the UI immediately.
 
-Using `@bind` with a `CurrentValue` property (`<input @bind="@CurrentValue" />`) is essentially equivalent to the following:
+Using `@bind` with a `CurrentValue` property (`<input @bind="CurrentValue" />`) is essentially equivalent to the following:
 
 ```cshtml
 <input value="@CurrentValue" 
--- a/aspnetcore/blazor/components.md
+++ b/aspnetcore/blazor/components.md
@@ -139,13 +139,13 @@ Using `@bind` with a `CurrentValue` property (`<input @bind="CurrentValue" />`) 
 
 When the component is rendered, the `value` of the input element comes from the `CurrentValue` property. When the user types in the text box, the `onchange` event is fired and the `CurrentValue` property is set to the changed value. In reality, the code generation is a little more complex because `@bind` handles a few cases where type conversions are performed. In principle, `@bind` associates the current value of an expression with a `value` attribute and handles changes using the registered handler.
 
-In addition to `onchange`, the property can be bound using other events like `oninput` by adding a `@bind` attribute with an `event` parameter:
+In addition to handling `onchange` events with `@bind` syntax, a property or field can be bound using other events by specifying an `@bind-value` attribute with an `event` parameter. The following example binds the `CurrentValue` property for the `oninput` event:
 
 ```cshtml
 <input @bind-value="CurrentValue" @bind-value:event="oninput" />
 ```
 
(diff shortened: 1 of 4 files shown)


ajaybhargavb:
@guardrex, looks like you're missing a detail about this. https://github.com/aspnet/AspNetCore.Docs/issues/12766 refers to `@` not being needed for `@onclick`, `@onfocus` etc. `@bind` doesn't have any bug and doesn't have to start with `@`.

ajaybhargavb:
> What's the difference between the following, if any:
<input @bind="@CurrentValue" />
<input @bind-value="@CurrentValue" @bind-value:event="onchange" />
Is the former just a shorthand way of setting up onchange binding? If they are the same, it isn't considered better to have Just One Consistent Way™️ 😄 to do it (the latter approach, which comports with other on{event} binding)?

There is no difference in behavior between the two in your example. The first one is just a short version. The second format is useful if you want to specify an event other than `onchange` like `oninput`. Like in the code example that you edited. `<input @bind-value="@CurrentValue" @bind-value:event="oninput" />`

aspnetcore/blazor/components.md
```diff
@@ -142,7 +142,7 @@ When the component is rendered, the `value` of the input element comes from the
 In addition to `onchange`, the property can be bound using other events like `oninput` by adding a `@bind` attribute with an `event` parameter:
 
 ```cshtml
-<input type="text" @bind-value="@CurrentValue" @bind-value:event="oninput" />
+<input @bind-value="@CurrentValue" @bind-value:event="oninput" />
```
ajaybhargavb: Why was `type="text"` removed?

guardrex: I thought everyone was allowing the default `type` to take care of it (e.g., Dan did it that way in the current ToDo List tutorial; Steve in the Blazing Pizza workshop). We're leaning that way, too, with the upcoming changes to the *Build your first* experience. I was just trying to make everything consistent. Is there a reason to keep it?

guardrex:
> you're missing a detail about this

Thanks ..... yes .... *it all makes sense now*. :smile: ...... i.e., I got 😵 on it.

aspnetcore/blazor/components.md
ajaybhargavb: I see. Okay

guardrex:
WRT my point number 2 .... do you think we should call that out in the topic? I could draft some text here for you to look at.

guardrex:
@ajaybhargavb *"it all makes sense"* ... famous last words for my tombstone. :smile: *RIP*

I went back to the original PR #12756.

At https://github.com/aspnet/AspNetCore.Docs/pull/12756/files#diff-5cf16565c6014e97ab5fd792aba7b93dR24 ... it preserves the `@` on `_italicsCheck`.

However, at https://github.com/aspnet/AspNetCore.Docs/pull/12756/files#diff-aec5790e12a05d20a3cda92a708f6679R123, it removes the `@`.

It looks like what we need to do here is take that `@` off of the `_italicsCheck` in the sample app's *HeadingComponent.razor*.

guardrex:
Also note that the `@` is on `CurrentValue` for `@bind-value` here in that PR ... https://github.com/aspnet/AspNetCore.Docs/pull/12756/files#diff-aec5790e12a05d20a3cda92a708f6679R142 ...

```cshtml
<input type="text" @bind-value="@CurrentValue" @bind-value:event="oninput" />
```

... and my original change on this PR was to make the text match (but for `@bind`, which seems incorrect) ...

```
Using `@bind` with a `CurrentValue` property (`<input @bind="@CurrentValue" />`) is essentially 
equivalent to the following:
```

So, I assume that there's a difference between `@bind` and `@bind-value` in these contexts.

ajaybhargavb:
> So, I assume that there's a difference between @bind and @bind-value in these contexts.

Nope. Both `@bind` and `@bind-value` should be expecting a C# value (meaning doesn't have to start with `@`). Please correct it if I missed anything in my original PR

guardrex:
ok ... I'll shape this PR up and ping u back.

aspnetcore/blazor/components.md
```diff
@@ -139,13 +139,13 @@ Using `@bind` with a `CurrentValue` property (`<input @bind="CurrentValue" />`)
...
-In addition to `onchange`, the property can be bound using other events like `oninput` by adding a `@bind` attribute with an `event` parameter:
+In addition to handling `onchange` events with `@bind` syntax, a property or field can be bound using other events by specifying an `@bind-value` attribute with an `event` parameter. The following example binds the `CurrentValue` property for the `oninput` event:
 
 ```cshtml
-<input type="text" @bind-value="@CurrentValue" @bind-value:event="oninput" />
+<input @bind-value="CurrentValue" @bind-value:event="oninput" />
 ```
 
-Unlike `onchange`, `oninput` fires for every character that is input into the text box.
+Unlike `onchange`, `oninput` fires for every character changed in the text box.
```
ajaybhargavb: Not sure about this change. I feel like `changed` is a bit too ambiguous for what it is trying to say.

aspnetcore/blazor/components.md
```diff
@@ -139,13 +139,13 @@ Using `@bind` with a `CurrentValue` property (`<input @bind="CurrentValue" />`)
 
 When the component is rendered, the `value` of the input element comes from the `CurrentValue` property. When the user types in the text box, the `onchange` event is fired and the `CurrentValue` property is set to the changed value. In reality, the code generation is a little more complex because `@bind` handles a few cases where type conversions are performed. In principle, `@bind` associates the current value of an expression with a `value` attribute and handles changes using the registered handler.
 
-In addition to `onchange`, the property can be bound using other events like `oninput` by adding a `@bind` attribute with an `event` parameter:
+In addition to handling `onchange` events with `@bind` syntax, a property or field can be bound using other events by specifying an `@bind-value` attribute with an `event` parameter. The following example binds the `CurrentValue` property for the `oninput` event:
```
ajaybhargavb: 👍

guardrex:
@ajaybhargavb kk ... that should get it. 🤞

aspnetcore/blazor/components.md
guardrex: But doesn't it fire for every change (e.g., a delete of a character isn't an "input" of text).

guardrex: Or is it literally **_only char inputs_** (typed) that it fires. I haven't worked with `oninput` in ages and ages. 👴

ajaybhargavb: No, you are right. But after reading that people might think, "But what does `onchange` do then? 🤔 ". Maybe if it explains both what onchange and oninput does, it might be less confusing.

guardrex: Looked it up ... yeah ... "Fired at controls when the user changes the value" ... so not just for chars "input" but chars changed. Should we reword and make it say something like ...

> fires when the value of the text box changes.

guardrex: > Maybe if it explains both what onchange and oninput does, it might be less confusing.

*Yes, indeed!* I must be getting smarter because I reached the same conclusion. :smile: See my last commit.

aspnetcore/blazor/components.md
```diff
@@ -145,7 +145,7 @@ In addition to handling `onchange` events with `@bind` syntax, a property or fie
 <input @bind-value="CurrentValue" @bind-value:event="oninput" />
 ```
 
-Unlike `onchange`, `oninput` fires for every character changed in the text box.
+Unlike `onchange`, which fires when the element loses focus, `oninput` fires when the value of the text box changes.
```
ajaybhargavb: Looks great!

PR #12901, opened 2019-06-17
Updates within PrefixKeyVaultSecretManager

### LINKED ISSUE #12900: Updates within PrefixKeyVaultSecretManager
In the doc page https://docs.microsoft.com/en-us/aspnet/core/security/key-vault-configuration?view=aspnetcore-2.2#use-a-key-name-prefix improve sample for `PrefixKeyVaultSecretManager` and inherit from `DefaultKeyVaultSecretManager` and use the predefined methods

    public class PrefixKeyVaultSecretManager : DefaultKeyVaultSecretManager
    {
        private readonly string _prefix;

        public PrefixKeyVaultSecretManager(string prefix)
        {
            _prefix = $"{prefix}-";
        }

        public override bool Load(SecretItem secret)
        {
            return secret.Identifier.Name.StartsWith(_prefix);
        }

        public override string GetKey(SecretBundle secret)
        {
            return base.GetKey(secret).Substring(_prefix.Length);
        }
    }
```

michaelmairegger:
Fixes #12900

### DIFF
--- a/aspnetcore/security/key-vault-configuration/sample_snapshot/Startup.cs
+++ b/aspnetcore/security/key-vault-configuration/sample_snapshot/Startup.cs
@@ -33,7 +33,7 @@ namespace SampleApp
     }
 
     #region snippet1
-    public class PrefixKeyVaultSecretManager : IKeyVaultSecretManager
+    public class PrefixKeyVaultSecretManager : DefaultKeyVaultSecretManager
     {
         private readonly string _prefix;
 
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks @michaelmairegger, but I'd prefer not to have this rest on the default implementation because I'd like readers to see the broadest possible approach for supplying an `IKeyVaultSecretManager`. The default impl assumes double-dashes are colons in stored secrets, which is a common case and what the topic recommends, but I don't want to assume that for everyone.

@anurse is the expert tho (and my co-author on the topic) ... @anurse, is it ok to continue with our current approach?

anurse:
I don't have a problem with using `DefaultKeyVaultSecretManager` as a base class, but this change also has the effect of forbidding the use of `--` in a prefix. Prior to this change, we stripped the prefix **before** doing the subsitution, but now we do the substitution first. Since the substitution will change the length of the string (`--` is changed to `:`) it would mess up the substring later.

Do I expect people to use `--` in their prefixes? No, but I've seen many things I didn't expect :). We'd either need to doc (and *validate*) that `--` is not allowed in a prefix, or leave this code as-is. Since I don't see a huge value in swapping to use the 1-line base class implementation here, I don't think this is a change we'd want to take.

guardrex:
Thanks @michaelmairegger tho for sending this. It's great to take a look at and consider alternatives for what we show.

PR #12907, opened 2019-06-17
Changed 'This partial:' to 'Above partial view:'

gurmeetsinghdke:
'Above partial view:' seems more appropriate than 'This partial'.


gurmeetsinghdke:
Changed 'This partial:' heading to 'Above partial view:' which seems more appropriate.

gurmeetsinghdke:
Linen no. 62,  changed This partial to Above partial view.

guardrex:
We can't use a directional word, such as "above," in the docs according to the style guidelines. We'd need to use a word like "preceding" there. However, I don't this is necessary. This whole section only deals with one partial view ... the one just shown before this list. Thanks for the suggestion ... we're always considering alternatives for what the docs show and explain.

Issue #10721, opened 2019-01-31
Could not load file or assembly


kleinepannekoek:
Hi! I'm trying to create a Scaffolding Item in a solution that already has multiple Scaffolding Items in it (so it should be possible, you'd think). After clicking "add", I get the following error

could not load file or assembly microsoft.entityframeworkcore,version2.0.3.0, culture=neutral, publickeytoken=adb9793829ddae60
the located assemblys manifest definition does not match the assembly reference.
exception from hresult 0x80131040

Does this mean the git-project I cloned was written in another version of VS? What can I do?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1d975a83-1550-b284-d0d8-7fe2ea206ff5
* Version Independent ID: 48e81b42-7c2b-afbf-71e6-d4f442a4bd1f
* Content: [Add a model to an ASP.NET Core MVC app](https://docs.microsoft.com/en-us/aspnet/core/tutorials/first-mvc-app/adding-model?view=aspnetcore-2.2&tabs=visual-studio)
* Content Source: [aspnetcore/tutorials/first-mvc-app/adding-model.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/tutorials/first-mvc-app/adding-model.md)
* Product: **aspnet-core**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
You can add multiple scaffolded projects. The error doesn't have anything to do with scaffolding. Can you build the project? 2.0 is an old version. Do you have the 2.0 SDK installed?

kleinepannekoek:
I had 2.2 installed first, but later I installed the later version 3.0. Both give the same message.

Rick-Anderson:
Yes, it's looking for the 2.0 version. 
Can you build the project?

Rick-Anderson:
Or update to 2.2

kleinepannekoek:
Yes I can build the project that I cloned, I just can't add a scaffolded item (and possibly oher items) and other versions of .NET CORE don't seem to solve the problem - it does seem to have something to do with Entity Framework but I dont know how to solve it. Here is some additional information that I get

Attempting to figure out the EntityFramework metadata for the model and DbContext: 'AP_Monitoring'
[...]

PR #12911, opened 2019-06-17
RCL static assets update

### LINKED ISSUE #6838: New doc:  use static assets that are part of a RCL project
I created a Razor Class Library this weekend. I was able to encapsulate Razor Views and associated Models + a Controller in a Razor Class Library and accompanying NuGet package. However, I would also like to encapsulate a collection of files under a Razor Class Library "wwwroot" folder that would be copied over to the "wwwroot" folder of the referencing ASP.NET Core web application upon package installation. Is this possible?

---
#### Document Details

* ID: d92ffa7c-fd65-7dc3-194e-6a67a8a7f176
* Version Independent ID: 2df11a37-d64e-6353-bd92-0b7729070c3e
* Content: [Reusable Razor UI in class libraries with ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/mvc/razor-pages/ui-class?view=aspnetcore-2.1&tabs=visual-studio)
* Content Source: [aspnetcore/mvc/razor-pages/ui-class.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/mvc/razor-pages/ui-class.md)
* Product: **asp.net-core**
* Microsoft Alias: **riande**

### LINKED ISSUE #12884: What can use static assets from RCLs?
This page says:
"In ASP.NET Core 3.0 Preview, Razor class libraries aren't compatible with Blazor client-side apps."
and
"In ASP.NET Core 3.0 Preview, Blazor server-side apps can't consume static assets from either a Razor class library (razorclasslib) or a Blazor library (blazorlib)."
but later
"ASP.NET Core allows creating RCLs that include static assets that are available to a consuming app."

But ... if client-side Blazor can't use RCLs and server-side Blazor can't consume static assets from an RCL... what exactly can use them?

---
#### Document Details

* ID: 37b02665-c8ab-bd91-8057-3b07b4e238d3
* Version Independent ID: e4f176f5-0b94-d532-53d5-cffce437961a
* Content: [Razor components class libraries](https://docs.microsoft.com/en-us/aspnet/core/blazor/class-libraries?view=aspnetcore-3.0&tabs=visual-studio)
* Content Source: [aspnetcore/blazor/class-libraries.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/class-libraries.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #12884
Fixes #6838

* Updates the Component Library topic: Thx @thatnerdyguy ... I pull the remarks that state RCL can't have static assets.
* Tea leaves indicate that `blazorlib` will be around until Pre9 looking at [Remove the Blazor Class Library template (aspnet/AspNetCore \#10721)](https://github.com/aspnet/AspNetCore/issues/10721). btw - The goal of getting client-side on RCL (either tracking on that engineering issue or another one) isn't on the [Blazor Roadmap (aspnet/AspNetCore \#8177)](https://github.com/aspnet/AspNetCore/issues/8177), which doesn't list Pre9.
* **Web apps** RCL topic moved up in the TOC to just under the *Razor syntax* topic.
* **Web apps** RCL topic receives the static assets content. The content is cross-linked from the **Blazor** Razor components class library topic.
* Escape pesky underscores in the RCL topic.


guardrex:
Ignore the content validation system ... it's just pull'in our chain there (false positive on tabs).

guardrex:
@tdykstra kk ... I'm *BACK!* 🏃 🏃 🏃 

Is this where you want the **Web apps** RCL topic? ... immediately under the *Razor syntax* topic?

tdykstra:
>  immediately under the Razor syntax topic?

Sounds good

mikes-gh:
Can I point out

https://github.com/aspnet/AspNetCore/issues/11174

I don't think its documented that if your RCL is named e.g. `Razor.Class.Lib`  the path to the static content will be `_content/razorclasslib `  (no dots lowercase)

guardrex:
@mikes-gh Thx ... I'll wait to hear from Dan on this. :question: Will we ...

* Have a warning about the behavior (phrased similarly to Javier's remarks perhaps).
* Have more of a *recommendation* ... 'We recommend naming libraries {this way} because {reasons}.'
* Not address it given that they consider it a bug to be fixed by the Pre8 release. Of course, we can always back text out of the topic at Pre8 release and go ahead with some coverage. After all, you hit this, and others are going to hit it between now and Pre8.

For reference ... @mikes-gh remarks ...

> Struggled with this as didn't know I had to replace the dotted project name with lowercase without dot.
> AFAIK dots are allowed in the path part of a URL.
> Also case is ignored in URLs
>
>Whats the reason for this
>
> Unlikely but if I had 2 projects of same name but only differing by dot this would break.
>
> Project.Test
> ProjectTest

Javier's reply ...

> dots are allowed, but normally cause issues, with many proxies and redirection rules, because people write simple rules like . to refer to files and that causes problems.
>
> Urls are case-sensitive. Many static files implementations simply rely on the file system casing, that's why in some cases they can seem case insensitive, and you can choose to interpret them that way, but in HTTP they are case sensitive, if not, try to set a cookie path to /PATH and visit the site at /path and you'll see how the browser doesn't send the cookie.

aspnetcore/razor-pages/ui-class.md
```diff
@@ -221,3 +221,28 @@ Suppose *RazorUIClassLib/Pages/Shared* contains two partial files: *_Header.csht
...
+
+Razor class libraries (RCL) frequently require companion static assets that can be referenced by the consuming app of the RCL. ASP.NET Core allows creating RCLs that include static assets that are available to a consuming app.
+
+To include companion assets as part of a Razor class library, create a *wwwroot* folder in the class library and include any required files in that folder.
+
+When packing a Razor class library, all companion assets in the *wwwroot* folder are included in the package automatically and are made available to apps referencing the package.
+
+### Consume content from a referenced Razor class library
+
+The files included in the *wwwroot* folder of the Razor class library are exposed to the consuming app under the prefix `_content/{LIBRARY NAME}/`. The consuming app references these assets via `<script>`, `<style>`, `<img>`, and other HTML tags.
```
danroth27: We should also say here the the consuming project needs to have static files support enabled: `app.UseStaticFiles()`

danroth27: Also I think we should go ahead and doc that "." characters are removed from the library name as pointed out by @mikes-gh.

guardrex:
> I don't understand why the underscores need to be escaped, but I trust that they do.

They don't usually cause a problem in built topics. However, GH diffs don't follow the CommonMark rules correctly, and the emphasis is rendered incorrectly making it more difficult to spot where emphasis has been missed or applied incorrectly.

Rick-Anderson:
> I don't understand why the underscores need to be escaped, but I trust that they do.

Let's not escape underscores, it makes the MD harder to read.

cc @tdykstra @scottaddie  @wadepickett

Rick-Anderson:
@guardrex  feel free to open a bug with GitHub on that.

guardrex:
@tdykstra Ignore the content validation on the tabs warning/error: It's a false positive.

aspnetcore/blazor/class-libraries.md
```diff
@@ -5,14 +5,14 @@ description: Discover how components can be included in Blazor apps from an exte
...
-ms.date: 06/14/2019
+ms.date: 06/24/2019
 uid: blazor/class-libraries
 ---
 # ASP.NET Core Razor components class libraries
 
 By [Simon Timms](https://github.com/stimms)
 
-Components can be shared in Razor class libraries across projects. Components can be included from:
+Components can be shared in a [Razor class library (RCL)](xref:razor-pages/ui-class) across projects. A *Razor components class library* can be included from:
```
tdykstra: Is a RCL the same thing as a RCCL?  If so, just to avoid possibility of confusion might be best to settle on one name.

aspnetcore/razor-pages/ui-class.md
```diff
@@ -221,3 +221,30 @@ Suppose *RazorUIClassLib/Pages/Shared* contains two partial files: *_Header.csht
   <partial name="_Footer">
 </body>
 ```
+
+## Create a Razor class library with static assets
+
+Razor class libraries (RCL) frequently require companion static assets that can be referenced by the consuming app of the RCL. ASP.NET Core allows creating RCLs that include static assets that are available to a consuming app.
```
tdykstra: The plural is RCL in one place and RCLs in another place.  Recommend standardize on RCLs for plural.

 ASP.NET Core allows creating RCLs that include static assets -->  RCLs can include static assets

aspnetcore/razor-pages/ui-class.md
```diff
@@ -221,3 +221,30 @@ Suppose *RazorUIClassLib/Pages/Shared* contains two partial files: *_Header.csht
...
+
+To include companion assets as part of a Razor class library, create a *wwwroot* folder in the class library and include any required files in that folder.
+
+When packing a Razor class library, all companion assets in the *wwwroot* folder are included in the package automatically and are made available to apps referencing the package.
```
tdykstra: packing --> packaging (?)

aspnetcore/razor-pages/ui-class.md
```diff
@@ -221,3 +221,30 @@ Suppose *RazorUIClassLib/Pages/Shared* contains two partial files: *_Header.csht
...
+
+### Consume content from a referenced Razor class library
+
+The files included in the *wwwroot* folder of the Razor class library are exposed to the consuming app under the prefix `_content/{LIBRARY NAME}/`. `{LIBRARY NAME}` is the library project name converted to lowercase with periods (`.`) removed. For example, a library named *Razor.Class.Lib* results in a path to static content at `_content/razorclasslib`.
```
tdykstra:
```suggestion
The files included in the *wwwroot* folder of the Razor class library are exposed to the consuming app under the prefix `_content/{LIBRARY NAME}/`. `{LIBRARY NAME}` is the library project name converted to lowercase with periods (`.`) removed. For example, a library named *Razor.Class.Lib* results in a path to static content at `_content/razorclasslib/`.
```

aspnetcore/razor-pages/ui-class.md
```diff
@@ -221,3 +221,30 @@ Suppose *RazorUIClassLib/Pages/Shared* contains two partial files: *_Header.csht
...
+
+The consuming app references static assets provided by the library with `<script>`, `<style>`, `<img>`, and other HTML tags. The consuming app must have static file support enabled. For more information, see <xref:fundamentals/static-files>.
```
tdykstra:
```suggestion
The consuming app references static assets provided by the library with `<script>`, `<style>`, `<img>`, and other HTML tags. The consuming app must have [static file support](xref:fundamentals/static-files) enabled.
```

aspnetcore/razor-pages/ui-class.md
```diff
@@ -221,3 +221,30 @@ Suppose *RazorUIClassLib/Pages/Shared* contains two partial files: *_Header.csht
...
+
+### Multi-project development flow
+
+When the app runs:
+
+* The assets stay in their original folders.
+* Any change within the class library *wwwroot* folder is reflected in the app without rebuilding.
```
tdykstra: "Without rebuilding" is ambiguous, could refer to the app or the class library.  Here I am guessing the reference is to building the app, but in the next paragraph it sounds like the reference is to building the library.

aspnetcore/blazor/class-libraries.md
guardrex: @danroth27 Drop the *RCCL*? If so, the title should probably go to something like ...

#### Provide ASP.NET Core Razor components in Razor class libraries

aspnetcore/razor-pages/ui-class.md
guardrex: Let's go with RCL. It's defined at the top of the topic and then used in many spots prior to adding the new content on this PR.

aspnetcore/razor-pages/ui-class.md
guardrex: "packing" I think ... the NuGet cats seem to lean that way, and our .NET Core topics, too. Example: https://docs.microsoft.com/dotnet/core/tools/dotnet-pack

guardrex:
To avoid confusion (*if the following is accurate* ping @danroth27 :eyes:), it can go in this direction (but the list and the para below are a hair bit redundant) ...

### Multi-project development flow

When the consuming app runs:

* The assets in the RCL stay in their original folders.
* Any change within the RCL's *wwwroot* folder is reflected in the consuming app without rebuilding the consuming app.

When the RCL is built, a manifest is produced that describes the static web asset locations. The consuming app reads the manifest at runtime to consume the assets from referenced projects and packages.

Rick-Anderson:
@guardrex  I don't find it redundant. 

> When the RCL is built, a manifest is produced

But a manifest only has address to assets. What if you add a new asset?

> Any change within the RCL's wwwroot folder is reflected in the consuming app without rebuilding the consuming app.

So it picks up the new asset. Is the manifest rebuilt?

guardrex:
We need DR ... that content was *dropped* here without further review (by me) because it came from the other PR where the engineers (and another author) worked on it.

> What if you add a new asset?

AFAIK, add the asset, rebuild the RCL, and the consuming app gets it at runtime without a rebuild.

> So it picks up the new asset. Is the manifest rebuilt?

Yes, AFAIK ... when the RCL is rebuilt.

aspnetcore/blazor/class-libraries.md
danroth27: A Razor class library is a class library that is setup for compiling Razor files (MVC Views, Razor Pages, Components, etc.). I'm fine with referring to Razor class libraries that contain components as Razor component class libraries. Always capitalizing "Razor Class Library" feels a bit excessive to me. I though the trend was towards lower case words?

guardrex: > I thought the trend was towards lower case words?

Just wrt using sentence case in titles and headings. Not with regard to proper nouns that are multiple words. ..... the Eiffel tower? Dan roth? Doctor who!!!! :smile: lol

Now that I re-examine the topic, Rick has it as a proper noun (capitalized), so this should go the same way. It matches the choice to capitalize by the VS devs/managers in the VS UI.

It's a bit of a grey area ... "Razor" in "Razor class library" is an adjective to describe a run-the-mill "class library" that does Razor-y things. That seems accurate. However, a "Razor Class Library" implies that this is a specially named aspect of the technology ... specifically for the purposes that you listed.

Anywho ... hardly matters from a presentation point-of-view. This will finish off the process of using RCL everywhere.

... and another lovely side note (do they ever *END*?! 🛑) ... "Razor Component" is a named, specific thing/aspect of the framework. I think that should be capitalized (i.e., not "Razor component"). I think that's a change that I've been slowly massaging docs into.

danroth27: I've been using "Razor component" instead of "Razor Component" in part to avoid confusion with the old "ASP.NET Core Razor Components" name. But it's also consistent with what other SPA frameworks do. For comparison, would you say "React component" or "React Component"? "Angular component" or "Angular Component"? The React and Angular docs both don't capitalize the "c" in component. Similarly, I don't think we should be capitalizing "class library" except maybe in the case where you are referring to the name of a specific project template. I.e the Razor Class Library template is used to create a Razor class library.

guardrex: Ok ... sounds good. I'll make the updates on the next commit.

guardrex:
Updates on the last commit for the pair of topics (other to be updated as I sweep the Blazor docs shortly on the tracking issue) ...

* "Razor class library" (except for the one spot where it first mentions the template, "Razor Class Library template")
* "Razor component"
* RCL used everywhere after the first mention. The Razor TOC area doc (*ui-class.md*) already mostly did this.

@danroth27 Is this bit correct? (@Rick-Anderson, I added a bit more in from your feedback) ...

### Multi-project development flow

When the consuming app runs:

* The assets in the RCL stay in their original folders. The assets aren't moved to the consuming app.
* Any change within the RCL's *wwwroot* folder is reflected in the consuming app after the RCL is rebuilt and without rebuilding the consuming app.

When the RCL is built, a manifest is produced that describes the static web asset locations. The consuming app reads the manifest at runtime to consume the assets from referenced projects and packages. When a new asset is added to an RCL, the RCL must be rebuilt to update its manifest before a consuming app can access the new asset.

danroth27:
> @danroth27 Is this bit correct?

Yup, I think so.

guardrex:
@Rick-Anderson That's your authored topic (the Razor class lib one).

Do you want :eyes: this before we :rocket: *GO! for throttle UP!* :rocket:

aspnetcore/blazor/components.md
```diff
@@ -49,7 +49,7 @@ Component members can then be used as part of the component's rendering logic us
...
 
-Components are ordinary C# classes and can be placed anywhere within a project. Components that produce webpages usually reside in the *Pages* folder. Non-page components are frequently placed into the *Shared* folder or a custom folder added to the project. To use a custom folder, either add the custom folder's namespace to the parent component or to the app's *\_Imports.razor* file. For example, the following namespace makes components in a *Components* folder available when the app's root namespace is `WebApplication`:
+Components are ordinary C# classes and can be placed anywhere within a project. Components that produce webpages usually reside in the *Pages* folder. Non-page components are frequently placed into the *Shared* folder or a custom folder added to the project. To use a custom folder, either add the custom folder's namespace to the parent component or to the app's *_Imports.razor* file. For example, the following namespace makes components in a *Components* folder available when the app's root namespace is `WebApplication`:
```
Rick-Anderson: Components are ordinary C# classes and can be placed anywhere within a project. Components that produce webpages usually reside in the *Pages* folder. Non-page components are frequently placed in the *Shared* folder or a custom folder added to the project. To use a custom folder, either add the custom folder's namespace to the parent component or to the app's *_Imports.razor* file. For example, the following namespace makes components in a *Components* folder available when the app's root namespace is `WebApplication`:

nacoleman:
Would I be correct in thinking that these assets cant then be used my webpack as they don't exist to the consuming app before runtime?

guardrex:
@nacoleman Yes WRT the consuming app. If webpack is an option as a pre-build step in the RCL, that should work.

javiercn:
@nacoleman All the assets are available as MSBuild items for you to use however you see fit, you can use MSBuild extensibility to find the assets and do whatever you see fit with them. Including invoking webpack to produce a bundle.

PR #12972, opened 2019-06-20
Disallow synchronous IO

### LINKED ISSUE #11148: Disallow synchronous IO
We should update the docs to reflect the fact that we disabled sync IO across all our servers.
Let's add conceptual content in the servers doc about and also covers how to disable us blocking synchronous IO

guardrex:
Fixes #11148

### DIFF
--- a/aspnetcore/fundamentals/servers/kestrel.md
+++ b/aspnetcore/fundamentals/servers/kestrel.md
@@ -5,7 +5,7 @@ description: Learn about Kestrel, the cross-platform web server for ASP.NET Core
 monikerRange: '>= aspnetcore-2.1'
 ms.author: tdykstra
 ms.custom: mvc
-ms.date: 06/18/2019
+ms.date: 06/20/2019
 uid: fundamentals/servers/kestrel
 ---
 # Kestrel web server implementation in ASP.NET Core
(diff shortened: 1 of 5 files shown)


aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -453,6 +453,44 @@ The default value is 96 KB (98,304).
 
 ::: moniker-end
 
+### Synchronous IO
```
Tratcher: Needs a warning: A large number of blocking synchronous IO operations can lead to thread pool starvation and make your application unresponsive. Only enable this if you're using a library that does not support async IO.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -453,6 +453,47 @@ The default value is 96 KB (98,304).
...
+::: moniker-end
+
+::: moniker range="< aspnetcore-3.0"
+
+<xref:Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.AllowSynchronousIO> controls whether synchronous IO is allowed for the request and response. The  default value is `true`.
+
+::: moniker-end
+
+> [!WARNING]
+> A large number of blocking synchronous IO operations can lead to thread pool starvation, which makes the app unresponsive. Only enable `AllowSynchronousIO` if using a library that doesn't support asynchronous IO.
```
Rick-Anderson: A large number of blocking synchronous IO operations can lead to thread pool starvation, which makes the app unresponsive. Only enable `AllowSynchronousIO` with a library that doesn't support asynchronous IO.

Rick-Anderson: A large number of blocking synchronous IO operations can lead to thread pool starvation, which makes the app unresponsive. Only enable `AllowSynchronousIO` when using a library that doesn't support asynchronous IO.

PR #12973, opened 2019-06-20
Remove PreserveCompilationContext requirement

pranavkm:
The RuntimeCompilation package does this on user's behalf. We had a bug in one of the earlier preview releases that required this, but it should be fixed now.


pranavkm:
Updated

scottaddie:
Thank you, @pranavkm!

PR #12975, opened 2019-06-21
Add HTTP REPL doc

### LINKED ISSUE #12507: Create doc for HTTP REPL tool
Work with Ahmed to create the doc. We need to develop content that:
o	Covers tool acquisition.
o	Guide: what the different commands are.
o	Use cases (how to use it with a couple of examples).
o	Configuration (so that when a user performs an HTTP post, it automatically opens up their editor of choice to edit the JSON to post).
o	Output: What the user should expect as out from the tool and what it means.

metoly:
Initial version of the HTTPRepl documentation.

Fixes #12507

[Internal review page](https://review.docs.microsoft.com/en-us/aspnet/core/web-api/http-repl?view=aspnetcore-3.0&branch=pr-en-us-12975)


aspnetcore/web-api/http-repl-tool.md
```diff
@@ -0,0 +1,224 @@
...
+description: Learn how to use the HTTP REPL .NET Core Global Tool to browse and test an ASP.NET Core web API.
+monikerRange: '>= aspnetcore-3.0'
+ms.author: scaddie
+ms.custom: mvc
+ms.date: 06/21/2019
+uid: web-api/http-repl-tool
+---
+# Use the HTTP REPL tool
+
+The HTTP REPL is a cross-platform tool that's supported everywhere .NET Core is supported.
```
scottaddie: Some more detail is needed. I suggest adding another sentence or two explaining the tool's purpose. Maybe also mention that it's a command-line alternative to tools like Fiddler and Postman.

aspnetcore/web-api/http-repl-tool.md
```diff
@@ -0,0 +1,224 @@
...
+    "id": 2,
+    "name": "Scott Guthrie"
+  }
+]
+
+
+http://localhost:5000/People~
+```
+
+### HTTP POST requests
```
scottaddie: They're common enough in the wild that a customer will inevitably ask for examples of DELETE and PUT. Should there be two additional sections to cover these HTTP verbs?

aspnetcore/web-api/http-repl-tool.md
```diff
@@ -0,0 +1,224 @@
...
+```console
+http://localhost:5000~ post
+```
+
+Issuing a `post` command opens an editor with the JSON template for you to enter the data to post. To use your preferred editor, use the command `pref set editor.command.default '<EXECUTABLE>'`, where `<EXECUTABLE>` is the full path to the editor's executable file. For example, `pref set editor.command.default 'C:\Program Files\Microsoft VS Code\Code.exe'`.
+
+## Additional resources
+
+* [REST API requests](https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#74-supported-methods)
+* [HTTPRepl GitHub repository](https://github.com/aspnet/AspLabs)
```
scottaddie:
```suggestion
* [HTTP REPL GitHub repository](https://github.com/aspnet/AspLabs)
```

aspnetcore/web-api/http-repl-tool.md
metoly: I didn't notice that earlier. That the description was not applicable. I added more details. I don't want to mention postman or fiddler as I don't want to get into the communalities and the differences. This might be better suited for a blog post.

aspnetcore/web-api/http-repl-tool.md
metoly: Added DELETE and PUT.

aspnetcore/web-api/http-repl-tool.md
```diff
@@ -0,0 +1,338 @@
...
+## Prerequisites
+
+* [.NET Core SDK 3.0.100](https://dotnet.microsoft.com/download/dotnet-core/3.0) or later
+
+## Installation
+
+To install the HTTP REPL:
+
+```console
+dotnet tool install -g dotnet-httprepl
```
bradygaster: Do we need to include the version number prior to it landing in the non-pre-release state? could be helpful prior to release.

aspnetcore/web-api/http-repl-tool.md
```diff
@@ -0,0 +1,338 @@
...
+
+GET            Issues a GET request.
+POST           Issues a POST request.
+PUT            Issues a PUT request.
+DELETE         Issues a DELETE request.
+PATCH          Issues a PATCH request.
+HEAD           Issues a HEAD request.
+OPTIONS        Issues an OPTIONS request.
+
+set header     Sets or clears a header for all requests. e.g. `set header content-type application/json`
```
bradygaster: Isn't `application/json` in there by default? Might be good to call that out.

aspnetcore/web-api/http-repl-tool.md
```diff
@@ -0,0 +1,338 @@
...
+
+### HTTP POST requests
+
+To issue an HTTP POST request, run the `post` command:
+
+```console
+http://localhost:5000~ post
+```
+
+Issuing a `post` command opens an editor with the JSON template for you to enter the data to post. To use your preferred editor, use the command `pref set editor.command.default '<EXECUTABLE>'`, where `<EXECUTABLE>` is the full path to the editor's executable file. For example, `pref set editor.command.default 'C:\Program Files\Microsoft VS Code\Code.exe'`.
```
bradygaster: Might want to call out how this would be done for a Mac/Linux user as well. For example, a Mac user using VS Code would route their command like this: `/Applications/Visual Studio Code.app/Contents/Resources/app/bin/code`

aspnetcore/toc.yml
```diff
@@ -425,6 +425,8 @@
       uid: web-api/advanced/analyzers
     - name: Conventions
       uid: web-api/advanced/conventions
+    - name: Use the HTTP REPL tool
```
bradygaster: Not sure I like this title. How about **Testing APIs with dotnet httprepl**? @scottaddie ?

aspnetcore/web-api/http-repl-tool.md
```diff
@@ -0,0 +1,338 @@
...
+title: Use the HTTP REPL tool
+author: scottaddie
+description: Learn how to use the HTTP REPL .NET Core Global Tool to browse and test an ASP.NET Core web API.
+monikerRange: '>= aspnetcore-3.0'
+ms.author: scaddie
+ms.custom: mvc
+ms.date: 06/21/2019
+uid: web-api/http-repl-tool
+---
+# Use the HTTP REPL tool
```
bradygaster: See previous comment - not entirely sure the title does it justice.

aspnetcore/toc.yml
scottaddie: I like your proposed title better. I'll incorporate this change.

aspnetcore/web-api/http-repl.md
```diff
@@ -0,0 +1,597 @@
...
+https://localhost:5001/people~
+```
+
+## Customize the HTTP REPL
+
+The HTTP REPL's default [colors](#set-the-default-colors) and [editor](#set-the-default-editor) can be customized. The HTTP REPL preferences are persisted across the current session and are honored in future sessions. Once modified, the preferences are stored in the following file:
+
+# [Linux / macOS](#tab/linux+macos)
+
+*%HOME%\.httpreplprefs*
```
tlmii: Looks like the backslashes are getting stripped (probably escaping the period) on the two paths.

aspnetcore/web-api/http-repl.md
```diff
@@ -0,0 +1,597 @@
...
+
+# [Windows](#tab/windows)
+
+*%USERPROFILE%\.httpreplprefs*
+
+---
+
+### Set the default colors
+
+To customize the default HTTP REPL tool coloring, run the `pref get` command. For example:
```
tlmii: It's a little awkward phrasing to say to customize it, run the get command. Should probably say something like "First, run the `pref get` command to see what the existing settings are"; or maybe even have another subheading of "View the color preferences" first before setting them? Not sure what the best route is.

aspnetcore/web-api/http-repl.md
```diff
@@ -0,0 +1,597 @@
...
+
+*%HOME%\.httpreplprefs*
+
+# [Windows](#tab/windows)
+
+*%USERPROFILE%\.httpreplprefs*
+
+---
+
+### Set the default colors
```
tlmii: Maybe title this "Set color preferences" instead? There are defaults baked in and this is really about overriding them.

aspnetcore/web-api/http-repl.md
```diff
@@ -0,0 +1,597 @@
...
+        "id": 2,
+        "name": "Scott Hanselman"
+      }
+    ]
+    
+    
+    https://localhost:5001/people~
+    ```
+
+### HTTP POST requests
```
tlmii: We should note that there are two other ways to do the HTTP requests that require bodies (POST, PUT, PATCH) - you can do it inline:
```
https://localhost:5001/people~ post --content "{ 'id': 0, 'name': 'Scott Addie'}"
```
and you can request it be pulled out of an existing file
```
https://localhost:5001/people~ post --file "c:\whatever.json"
```
Further, if you don't want to send a body (which is rare, so that's why it's not the default) you can do
```
https://localhost:5001/people~ post --no-body
```
If none of those three arguments are specified, that's when the tool will pop up the default editor

tlmii:
I'm guessing you were waiting for the Run command tests to go in before writing the documentation on that, since we discussed that offline a bit. Just putting this comment out there as a placeholder that we should definitely document that command since it can be pretty useful.

tlmii:
I think the only other commands not documented right now are `set diag` and`config`, both of which are related and which are not well documented in the code either. I think they can probably fall in that same bucket to wait for the tests for them so its clear what should be documented.

aspnetcore/web-api/http-repl.md
scottaddie: Awesome idea to add a "View..." section. I'll probably go that route.

aspnetcore/web-api/http-repl.md
```diff
@@ -0,0 +1,811 @@
...
+
+### Arguments
+
+`PARAMETER`
+
+The route parameter, if any, expected by the associated controller action method.
+
+### Options
+
+* `c|--content`
```
tlmii: missing the dash before the c (same in other commands)

tlmii:
Do you think there's any value in discussing the default colors and the way they sort of "cascade"?

See code [here](https://github.com/aspnet/HttpRepl/blob/master/src/Microsoft.HttpRepl/Preferences/JsonConfig.cs), [here](https://github.com/aspnet/HttpRepl/blob/master/src/Microsoft.HttpRepl/Preferences/RequestOrResponseConfig.cs), [here](https://github.com/aspnet/HttpRepl/blob/master/src/Microsoft.HttpRepl/Preferences/RequestConfig.cs) and [here](https://github.com/aspnet/HttpRepl/blob/master/src/Microsoft.HttpRepl/Preferences/ResponseConfig.cs). 

But for example,

In Json output, if colors.json.name doesn't have a value, it uses colors.json.string. But if colors.json.string doesn't have a value, it uses colors.json.literal. But if colors.json.literal doesn't have a value, it uses colors.json. And if that doesn't have a value, it ends up with AllowedColors.None, which leaves the coloring at the normal text color.

I'm guessing most people won't change the coloring, and even fewer will probably care about that depth of customization. But it's something that was non-obvious at first, so I wanted to throw it out there to see what you thought.

scottaddie:
@tlmii Since you already figured it out, I'll document it. Thank you for pointing it out.

aspnetcore/web-api/http-repl.md
```diff
@@ -0,0 +1,875 @@
...
+
+Response colorization is currently supported for JSON only. To customize the default HTTP REPL tool coloring, locate the key corresponding to the color to be changed. For instructions on how to find the keys, see the [View the settings](#view-the-settings) section. For example, change the `colors.json` key value from `Green` to `White` as follows:
+
+```console
+https://localhost:5001/people~ pref set colors.json White
+```
+
+Only the [allowed colors](https://github.com/aspnet/HttpRepl/blob/01d5c3c3373e98fe566ff5ef8a17c571de880293/src/Microsoft.Repl/ConsoleHandling/AllowedColors.cs) may be used. Subsequent HTTP requests display output with the new coloring.
+
+If `colors.json.name` doesn't have a value, `colors.json.string` is used. If `colors.json.string` doesn't have a value, `colors.json.literal` is used. If `colors.json.literal` doesn't have a value, `colors.json` is used. If `colors.json` doesn't have a value, the command shell's default text color (`AllowedColors.None`) is used.
```
tlmii: Since this is just an example of the cascading effect of the colors, should we note that somehow? Perhaps preface this section with something like (definitely needs wordsmithing): 

> The default values for the color settings will fall back to more generic settings if not specified. For instance, if `colors.json.name` ...

PR #13015, opened 2019-06-26
Fixed typo in secpool.msc command

The-Smallest:
Fixed typo in secpool.msc command

### DIFF
--- a/aspnetcore/host-and-deploy/windows-service.md
+++ b/aspnetcore/host-and-deploy/windows-service.md
@@ -205,7 +205,7 @@ An alternative approach to managing users when using Active Directory is to use 
 
 To establish *Log on as a service* rights for a service user account:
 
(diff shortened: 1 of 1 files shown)


scottaddie:
@The-Smallest Thank you for correcting this mistake!

The-Smallest:
Always welcome.
There is one more issue with this article:
There is no such cmdlet *Remove-Service* in Powershell, so I had to use `sc delete $name`.
Didn't have time to investigate and not sure what to do with this knowledge (

guardrex:
@The-Smallest `Remove-Service` is PS6 ... https://docs.microsoft.com/powershell/module/microsoft.powershell.management/remove-service?view=powershell-6, which is called out in the prereqs ... https://docs.microsoft.com/aspnet/core/host-and-deploy/windows-service#prerequisites

PR #13028, opened 2019-06-27
Update java doc

mikaelm12:
We've added some features in 3.0 so we need to update the limitations section on the Java client.


scottaddie:
Thank you, @mikaelm12!

PR #12996, opened 2019-06-24
[WIP] Social without Identity - Updates

serpent5:
@Rick-Anderson

Still a work-in-progress for now, but I thought it might be useful to create a draft PR to show where I'm at. The main thing that's outstanding is the explanation of the `DefaultScheme`, `DefaultChallengeScheme`, etc (I know, I know, that's the main part :stuck_out_tongue_closed_eyes:). I'll address that in the next commit, which I'm hoping to get to in the next day or two.

Feedback on what I have so far is, of course, very much appreciated. I've borrowed content from other topics and tried to follow conventions I've found there as best as I can.

PR #13032, opened 2019-06-27
Updates to Social without Identity.

### LINKED ISSUE #12091: Google et al without Identity

serpent5:
Fixes #12091
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/security/authentication/social/social-without-identity?view=aspnetcore-1.0&branch=pr-en-us-13032)

Copy of #12996 as suggested by @tdykstra.


tdykstra:
@serpent5 Same error, I'll open a ticket to investigate.

serpent5:
Thanks @tdykstra. I hope it's not just that I've done the same wrong thing in my workflow both times. At least this time all the checks passed, so that's better at least.

tdykstra:
Internal ticket number [101626](https://ceapex.visualstudio.com/SRE/_workitems?id=101626)

tdykstra:
The error seems to have gone away.

serpent5:
@Rick-Anderson I think I'm done with the first attempt at this. I never quite realised just how much time could be spent working on something like this to end up with only a few lines at the end. You'd think I'd know all about that being a programmer but this is different... :sweat_smile:

Rick-Anderson:
@serpent5 can you check [my commit](https://github.com/aspnet/AspNetCore.Docs/pull/13032/commits/ae8b5f136ff3b6b842bcbf165c6bb8a24a8aff86)?

serpent5:
Yeah, LGTM. I'd tried to get that xref change working myself but it looks like I was missing a `*` that I had no idea about.

Rick-Anderson:
@serpent5  try https://xrefgen.azurewebsites.net/?handler=GetSearchResults

blowdart:
"The DefaultScheme is used as a fallback default scheme. For example, if DefaultChallengeScheme is not set, calls to HttpContext.ChallengeAsync fallback to use the DefaultScheme." feels a little clumsy wording wise. It's not a fall back, it's the default. Switch to be more positive? Something like

"The DefaultScheme sets the scheme to be used when not overridden by setting any of the other options. For example setting the DefaultChallengeScheme to "Example" would mean HttpContext.ChallengeAsync calls the challenge code provided by the Example scheme, but leaves every other action with the default scheme".

That's not great either, but you get my point hopefully.

serpent5:
> That's not great either, but you get my point hopefully.

Yeah, I completely understand. I had a hard time with that part and tried numerous approaches before I landed on that. I tried to go with the terminology from the [api docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.authentication.authenticationoptions.defaultscheme?view=aspnetcore-2.2) for it that says:

_Used as the fallback default scheme for all the other defaults._

But that didn't quite fit as it made it seem like `ChallengeAsync` had special treatment when referring to "other defaults" in the context of this page. I'll give it some more thought and see if I can come up with something better that's more along the lines of your suggestion. @Rick-Anderson Do you have any suggestions for the wording here?

Rick-Anderson:
@serpent5  feel free to be verbose. I think you'll need an explanation, then perhaps some follow up to really clarify any ambiguity.

serpent5:
@Rick-Anderson @blowdart I hope that you'll agree that reads a little better than it did...

Rick-Anderson:
@serpent5  thanks for your help!
You might want to look at my last 2 commits.

serpent5:
@Rick-Anderson No problem. More than happy to help. :smile:

serpent5:
@Rick-Anderson I just had a look at the latest changes. It looks good except I think switching from "Applying simple authorization" to "Applying basic authorization" *might* be a little confusing. Could this be confused with [basic authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#Basic_authentication_scheme)?. I might be overthinking it... I went with "simple" as it's the term used for [this page](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/simple?view=aspnetcore-2.2).

Rick-Anderson:
Good point. Simple is a banned word and what's simple to the author often isn't simple to the reader.  In this case simple might be better, unless you can think of a better synonym.

serpent5:
I think I'd go with "simple" to match the page I mentioned, or maybe even just drop it and go with "Applying authorization"?

PR #13058, opened 2019-07-01
Drop remarks on setting the content root

### LINKED ISSUE #13057: Possible mistake in content root path setting example
If we use code sample from example, line ``builder.UseSolutionRelativeContentRoot("<SOLUTION-RELATIVE-PATH>");`` won't be called:
```
public class CustomWebApplicationFactory<TStartup>
     : WebApplicationFactory<RazorPagesProject.Startup>
{
     {
         builder.ConfigureServices(services =>
         {
             //This code is not called
             builder.UseSolutionRelativeContentRoot("<SOLUTION-RELATIVE-PATH>");
             ...
         });
     }
}
```
But if we write this line directly in overriden method, this will override content root path correctly:

```
public class CustomWebApplicationFactory<TStartup>
     : WebApplicationFactory<RazorPagesProject.Startup>
{
     {
        //Now this works
        builder.UseSolutionRelativeContentRoot("<SOLUTION-RELATIVE-PATH>");
     }
}
```

---
#### Document Details

* Content: [Integration tests in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-2.2#how-the-test-infrastructure-infers-the-app-content-root-path)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #13057

### DIFF
--- a/aspnetcore/test/integration-tests.md
+++ b/aspnetcore/test/integration-tests.md
@@ -5,7 +5,7 @@ description: Learn how integration tests ensure that an app's components functio
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 06/05/2019
+ms.date: 07/01/2019
 uid: test/integration-tests
 ---
 # Integration tests in ASP.NET Core
@@ -295,12 +295,9 @@ Call the [UseSolutionRelativeContentRoot](/dotnet/api/microsoft.aspnetcore.testh
    {
        protected override void ConfigureWebHost(IWebHostBuilder builder)
        {
-           builder.ConfigureServices(services =>
-           {
-               builder.UseSolutionRelativeContentRoot("<SOLUTION-RELATIVE-PATH>");
+           builder.UseSolutionRelativeContentRoot("<SOLUTION-RELATIVE-PATH>");
 
-               ...
-           });
+           ...
        }
    }
    ```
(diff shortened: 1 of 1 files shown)


guardrex:
@javiercn This might be a bug fix. It will only take 10 seconds to confirm with you.

aspnetcore/test/integration-tests.md
```diff
@@ -295,12 +295,9 @@ Call the [UseSolutionRelativeContentRoot](/dotnet/api/microsoft.aspnetcore.testh
    {
        protected override void ConfigureWebHost(IWebHostBuilder builder)
        {
-           builder.ConfigureServices(services =>
-           {
-               builder.UseSolutionRelativeContentRoot("<SOLUTION-RELATIVE-PATH>");
+           builder.UseSolutionRelativeContentRoot("<SOLUTION-RELATIVE-PATH>");
```
javiercn: I'm confused, why are recommending this in the first place? The WAF base class should take care of this automatically.

guardrex: We say this today ...

> In most cases, it isn't necessary to explicitly set the app content root, as the search logic usually finds the correct content root at runtime. In special scenarios where the content root isn't found using the built-in search algorithm, the app content root can be specified explicitly or by using custom logic. To set the app content root in those scenarios, call the `UseSolutionRelativeContentRoot` extension method ...

Would you like to drop all of that + the code example?

javiercn: I think so.

guardrex: Thanks @javiercn ... I hope to be free-er-ish later today, and I'll get this squared away.

guardrex:
@Rick-Anderson Approval is here :point_right: https://github.com/aspnet/AspNetCore.Docs/pull/13058#discussion_r310690947

PR #13064, opened 2019-07-01
Fix configuration across versions

BrennanConroy:
`ClientTimeoutInterval` was added in 2.2 so should not be present in 2.1


scottaddie:
Thank you, @BrennanConroy!

PR #13107, opened 2019-07-03
Auth handlers with SignalR

### LINKED ISSUE #13050: Document new HubInvocationContext in authorization handlers
In 3.0, SignalR now passes a resource for [authorization handlers](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/resourcebased).

The change is [here](https://github.com/aspnet/AspNetCore/pull/11070) for reference when writing/updating the docs.

We should show how to write an auth handler and explain what resource we are passing in.

BrennanConroy:
Fixes #13050

Bike-shed and come up with a better arbitrary example!

[Internal review link](https://review.docs.microsoft.com/aspnet/core/signalr/authn-and-authz?view=aspnetcore-3.0&branch=brecon%2Fauth)

### DIFF
--- a/aspnetcore/signalr/authn-and-authz.md
+++ b/aspnetcore/signalr/authn-and-authz.md
@@ -203,7 +203,7 @@ public class HubMethodAuthorizationHandler :
 
 public class HubMethodRequirement : IAuthorizationRequirement { }
 
-public ChatHub : Hub
+public class ChatHub : Hub
 {
     [Authorize("NamedMethod")]
     public void Steve()
--- a/aspnetcore/signalr/authn-and-authz.md
+++ b/aspnetcore/signalr/authn-and-authz.md
@@ -5,7 +5,7 @@ description: Learn how to use authentication and authorization in ASP.NET Core S
 monikerRange: '>= aspnetcore-2.1'
 ms.author: bradyg
 ms.custom: mvc
-ms.date: 05/09/2019
+ms.date: 07/15/2019
 uid: signalr/authn-and-authz
 ---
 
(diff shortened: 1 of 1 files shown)


aspnetcore/signalr/authn-and-authz.md
```diff
@@ -176,6 +176,62 @@ public class ChatHub: Hub
 }
 ```
 
+
+### Use authorization handlers to customize hub method authorization
+
+SignalR passes a custom resource to authorization handlers when a hub method requires authorization. The resource is of type `HubInvocationContext` and contains `HubCallerContext`, the name of the hub method, and the arguments to the hub method for use in custom authorization logic.
```
anurse:
```suggestion
SignalR provides a custom resource to authorization handlers when a hub method requires authorization. The resource an instance of `HubInvocationContext`. The `HubInvocationContext` includes the `HubCallerContext`, the name of the hub method being invoked, and the arguments to the hub method.
```

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -176,6 +176,62 @@ public class ChatHub: Hub
...
+public class ChatHub : Hub
+{
+    [Authorize("NamedMethod")]
+    public void Steve()
+    {
+    }
+}
+```
+
+Setup the policy and register the handler in DI.
```
anurse:
```suggestion
Set up the policy and register the handler in DI.
```

"Setup" is a noun, such as "it's a setup!". "Set up" is the verb phrase you're looking for ;).

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -176,6 +176,62 @@ public class ChatHub: Hub
...
+        if (context.User?.FindFirst(ClaimTypes.NameIdentifier)?.Value == resource.HubMethodName)
+        {
+            context.Succeed(requirement);
+        }
+
+        return Task.CompletedTask;
+    }
+}
+
+public class HubMethodRequirement : IAuthorizationRequirement { }
```
halter73: This seems really weird. Should we be using OperationAuthorizationRequirement?

halter73: https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.authorization.iauthorizationrequirement?view=aspnetcore-2.2

BrennanConroy:
cc @HaoK 
Have any input on a better scenario to show off here? And also avoid an empty requirement class.

anurse:
@bradygaster is using "user must have an `@microsoft.com` email address to call this method" example in another context. Maybe that's good?

bradygaster:
thanks for the vote @anurse - i could retrofit @BrennanConroy's doc with the sample i'm going to put into the blog if y'all think that'd be good. Here's the M&P of it: 

Hub and Custom Classes:

```csharp
public class ChatHub : Hub
{
    [Authorize("DomainRestricted")]
    public void SendMessage()
    {
    }
}

public class DomainRestrictedRequirement : IAuthorizationRequirement { }

public class DomainRestrictedAuthorizationHandler : AuthorizationHandler<DomainRestrictedRequirement, HubInvocationContext>
{
        DomainRestrictedRequirement requirement, 
        HubInvocationContext resource)
    {
        if(context.User != null && 
            context.User.Identity != null && 
            context.User.Identity.Name.EndsWith("@microsoft.com"))
        {
            context.Succeed(requirement);
        }
            
        return Task.CompletedTask;
    }
}
```

Startup.cs:

```csharp
services
    .AddAuthorization(options =>
    {
        options.AddPolicy("DomainRestricted", policy =>
        {
            policy.Requirements.Add(new DomainRestrictedRequirement());
        });
    });

services.AddSingleton<IAuthorizationHandler, DomainRestrictedAuthorizationHandler>();
```

My question, though, is - what does a developer do if they want to have a second (or Nth) `IAuthorizationRequirement` and `AuthorizationHandler` implementation to use on other Hub methods? From our example, it looks like one needs to wire up the handler as a Singleton, so it seems like the implication is that I'd need to write on handler for N requirements. What's a dev to do when they need multiple requirements? I know we'll be asked this.

BrennanConroy:
That's a fine sample, however it doesn't make use of the `HubInvocationContext resource` at all which is what this PR is trying to highlight.

As it stands that code would have worked fine before in 2.1 and 2.2 (assuming you tweaked the code slightly to use the non-strongly-typed resource based handler)

anurse:
Ah yeah, if we could make it conditional on a particular argument being passed that would be most appropriate. Even just being conditional on the method name (though that's a little contrived since if you wanted to apply this to specific methods you could just use a policy in your `AuthorizeAttribute`)

HaoK:
You can put state into the requirement if you want too, for example you could do something like this where the requirement is also the handler, and then store a white list of allowed methods to call:

```
public class HubMethodRequirement : AuthorizationHandler<HubMethodRequirement, HubInvocationContext>, IAuthorizationRequirement
{
   private IEnumerable<string> _whitelist;
   public HubMethodRequirement(IENumerable<string> whitelist) => _whitelist = whitelist;

                                                   HubMethodRequirement requirement,
                                                   HubInvocationContext resource)
    {
        if (_whiteList.Contains(resource.HubMethodName) && context.User?.FindFirst(ClaimTypes.NameIdentifier)?.Value == resource.HubMethodName)
        {
            context.Succeed(requirement);
        }

        return Task.CompletedTask;
    }
}
```

bradygaster:
i like that idea @HaoK - i'll tweak the example i worked up to be more inline with that approach and highlight the `HubInvocationContext resource` too, as @BrennanConroy make a good point that this is the crux of the highlight.

bradygaster:
i've updated the code in the above comment i made to reflect the great feedback provided by @BrennanConroy and @HaoK. went with a method approach rather than a whitelist, thinking one might have a database of user/hub method rights. Still a little contrived, but also a little more realistic to show how to control access to methods.

bradygaster:
so last ping to @BrennanConroy @anurse - you okay with me using the code above for the blog? if so i could also update the PR to include it. Thanks to @HaoK for the answer to my multiple-requirements question, too.

BrennanConroy:
Sounds good!

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -176,6 +176,62 @@ public class ChatHub: Hub
...
+    services.AddSingleton<IAuthorizationHandler, HubMethodAuthorizationHandler>();
+}
+```
+
+::: moniker-end
+
 ## Additional resources
 
 * [Bearer Token Authentication in ASP.NET Core](https://blogs.msdn.microsoft.com/webdev/2016/10/27/bearer-token-authentication-in-asp-net-core/)
+* [Resource based Authorization](xref:security/authorization/resourcebased)
```
scottaddie:
```suggestion
* [Resource-based Authorization](xref:security/authorization/resourcebased)
```

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -176,6 +176,62 @@ public class ChatHub: Hub
...
+public class ChatHub : Hub
+{
+    [Authorize("NamedMethod")]
+    public void Steve()
+    {
+    }
+}
+```
+
+Set up the policy and register the handler in DI.
```
scottaddie:
```suggestion
Set up the policy and register the handler in DI:
```

PR #13135, opened 2019-07-05
update the commands to install and update

DrMoray:
the old lines don't run


scottaddie:
Thank you for fixing this, @DrMoray!

PR #13209, opened 2019-07-09
Enhance service registration scenario coverage

### LINKED ISSUE #12734: Enhance service registration scenario coverage
The topic isn't specific/clear on the different service registration scenarios (`Add{LIFETIME}` extensions). Seems like a section should be here called "Service registration scenarios" (or similar) that explains the service extensions in more detail than what is found via the code remarks that appear in the API Browser.

For the reader feedback, see @MarkStega's comments on https://github.com/aspnet/AspNetCore.Docs/issues/10772.

---
#### Document Details

* Content: [Dependency injection in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.2)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #12734

[Internal Review Topic (*Service registration methods* section direct link)](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.1&branch=pr-en-us-13209#service-registration-methods)

@anurse We had an ask to clarify the service registration overloads. Do you have a couple of minutes to take a look at this new section? It's short and won't take long to :eyes:.

1. Is it sane thus far? 🙈
1. What's it missing? 🐄:bell:
1. Should it explicitly recommend the *concrete type that implements an interface* approach? 👩‍⚖

### DIFF
--- a/aspnetcore/fundamentals/dependency-injection.md
+++ b/aspnetcore/fundamentals/dependency-injection.md
@@ -63,7 +63,7 @@ The class creates and directly depends on the `MyDependency` instance. Code depe
 Dependency injection addresses these problems through:
 
 * The use of an interface to abstract the dependency implementation.
-* Registration of the dependency in a service container. ASP.NET Core provides a built-in service container, [IServiceProvider](/dotnet/api/system.iserviceprovider). Services are registered in the app's `Startup.ConfigureServices` method.
+* Registration of the dependency in a service container. ASP.NET Core provides a built-in service container, <xref:System.IServiceProvider>. Services are registered in the app's `Startup.ConfigureServices` method.
 * *Injection* of the service into the constructor of the class where it's used. The framework takes on the responsibility of creating an instance of the dependency and disposing of it when it's no longer needed.
 
 In the [sample app](https://github.com/aspnet/AspNetCore.Docs/tree/master/aspnetcore/fundamentals/dependency-injection/samples), the `IMyDependency` interface defines a method that the service provides to the app:
@@ -74,7 +74,7 @@ This interface is implemented by a concrete type, `MyDependency`:
 
 [!code-csharp[](dependency-injection/samples/2.x/DependencyInjectionSample/Services/MyDependency.cs?name=snippet1)]
 
-`MyDependency` requests an [ILogger&lt;TCategoryName&gt;](/dotnet/api/microsoft.extensions.logging.ilogger-1) in its constructor. It's not unusual to use dependency injection in a chained fashion. Each requested dependency in turn requests its own dependencies. The container resolves the dependencies in the graph and returns the fully resolved service. The collective set of dependencies that must be resolved is typically referred to as a *dependency tree*, *dependency graph*, or *object graph*.
+`MyDependency` requests an <xref:Microsoft.Extensions.Logging.ILogger`1> in its constructor. It's not unusual to use dependency injection in a chained fashion. Each requested dependency in turn requests its own dependencies. The container resolves the dependencies in the graph and returns the fully resolved service. The collective set of dependencies that must be resolved is typically referred to as a *dependency tree*, *dependency graph*, or *object graph*.
 
 `IMyDependency` and `ILogger<TCategoryName>` must be registered in the service container. `IMyDependency` is registered in `Startup.ConfigureServices`. `ILogger<TCategoryName>` is registered by the logging abstractions infrastructure, so it's a [framework-provided service](#framework-provided-services) registered by default by the framework.
 
@@ -89,7 +89,7 @@ In the sample app, the `IMyDependency` service is registered with the concrete t
 [!code-csharp[](dependency-injection/samples/2.x/DependencyInjectionSample/Startup.cs?name=snippet1&highlight=5)]
 
 > [!NOTE]
-> Each `services.Add{SERVICE_NAME}` extension method adds (and potentially configures) services. For example, `services.AddMvc()` adds the services Razor Pages and MVC require. We recommended that apps follow this convention. Place extension methods in the [Microsoft.Extensions.DependencyInjection](/dotnet/api/microsoft.extensions.dependencyinjection) namespace to encapsulate groups of service registrations.
+> Each `services.Add{SERVICE_NAME}` extension method adds (and potentially configures) services. For example, `services.AddMvc()` adds the services Razor Pages and MVC require. We recommended that apps follow this convention. Place extension methods in the <xref:Microsoft.Extensions.DependencyInjection?displayProperty=fullName> namespace to encapsulate groups of service registrations.
 
 If the service's constructor requires a [built in type](/dotnet/csharp/language-reference/keywords/built-in-types-table), such as a `string`, the type can be injected by using [configuration](xref:fundamentals/configuration/index) or the [options pattern](xref:fundamentals/configuration/options):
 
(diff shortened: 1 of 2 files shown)


anurse:
Looks good to me. I wonder if it's easiest to just build a table like:

| Method | Lifetime | Disposed automatically? |
| - | - | - |
| `AddSingleton<IMyService, MyService>()` | Singleton | ✔️ |
| `AddSingleton<IMyService>(() => new MyService())` | Singleton | ✔️ |
| `AddSingleton<MyService>()` | Singleton | ✔️ |
| `AddSingleton<IMyService>(new MyService())` | Singleton | ❌ |
| `AddSingleton(new MyService())` | Singleton | ❌ |

(Repeat for Transient and Scoped)

It might also be good to mention the `TryAdd` and `TryAddEnumerable` methods.

* `TryAdd{LIFETIME}` methods only add the service if there isn't already an instance of it in the container

```csharp
services.AddSingleton<IFoo, OverrideFoo>();
services.TryAddSingleton<IFoo, DefaultFoo>(); // No-op! Already an IFoo registered
```

* `TryAddEnumerable(ServiceDescriptor)` methods only add the service if there isn't already an instance of **this specific implementing type** in the container. This one doesn't have all the overloads, it only has the version that takes `ServiceDescriptor` directly (which is what all the `TryAdd{LIFETIME}` methods proxy to anyway). *This is definitely a edge-case scenario*. Generally it's used by library authors.

```csharp
services.TryAddEnumerable(ServiceDescriptor.Singleton<IFrobProvider, DatabaseFrobProvider>());
services.TryAddEnumerable(ServiceDescriptor.Singleton<IFrobProvider, BlobFrobProvider>());
services.TryAddEnumerable(ServiceDescriptor.Singleton<IFrobProvider, DatabaseFrobProvider>()); // No-op! There is already a DatabaseFrobProvider registered for IFrobProvider
```

The idea of `TryAddEnumerable` is to support the scenarios where you're registering multiple services in the container and you plan to resolve them via `IEnumerable<IFrobProvider>` but when registering you want to only add an instance if one hasn't already been added. This is used a lot in things like SignalR where we register the `JsonHubProtocol` in `services.AddSignalR()`, but if you've already registered it, we don't want two copies!

guardrex:
Thanks for looking. :eyes:

> easiest to just build a table

I *think* 🤔 that it's best to make the lifetime generic to avoid duplication, and the details on disposal are surfaced clearly down in the linked, dedicated section on disposal. A table makes it harder to display the X-by-Y scenarios ... the *other whys* that a given overload might be chosen besides automatic disposal (testing and parameters/arguments).

After the next commit, we'll see what the consensus is on docs review for the layout. Whatever we do, it will certainly be helpful over the current missing content.

*Yes, indeed!* on `TryX` bits. I'll get those in there. *Thanks for that!* I'll 👮 *blatantly steal* 🚓 as much of your language as I can. :smile:

guardrex:
@anurse @Rick-Anderson 

Take a look at these options for the content display:

[Internal Review Topic (*Service registration methods* section direct link)](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.1&branch=pr-en-us-13209#service-registration-methods)

Two content layouts are shown: **Option A: 2-level list** and **Option B: Table**.

I like them both ... I think either would be fine. What do you prefer there?

@anurse You won't need to look again after one more feedback round. The rest of the updates are just for things like API links, style updates, etc. ... things that we'll take care of on docs team review. Technically tho, anything wrong thus far with the bits that I added?

Rick-Anderson:
I much prefer the table because I can quickly compare  disposal/testing/pass-args
That seems like great info.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -174,6 +174,85 @@ Singleton lifetime services are created the first time they're requested (or whe
...
+| ------ | :-----------------------------: | :---------------------------------: | :---------------------------------: |
+| `Add{LIFETIME}<{INTERFACE}, {TYPE}>()`<br>Example:<br>`services.AddScoped<IMyDep, MyDep>();` | Yes | Yes | No |
+| `Add{LIFETIME}<{INTERFACE}>(sp => new {TYPE})`<br>Examples:<br>`services.AddScoped<IMyDep>(sp => new MyDep());`<br>`services.AddScoped<IMyDep>(sp => new MyDep("A string!"));` | Yes | Yes | Yes |
+| `Add{LIFETIME}<{TYPE}>()`<br>Example:<br>`services.AddScoped<MyDep>();` | Yes | No | No |
+| `Add{LIFETIME}<{INTERFACE}>(new {TYPE})`<br>Examples:<br>`services.AddScoped<IMyDep>(new MyDep());`<br>`services.AddScoped<IMyDep>(new MyDep("A string!"));` | No | Yes | Yes |
+| `Add{LIFETIME}(new {TYPE})`<br>Examples:<br>`services.AddScoped(new MyDep());`<br>`services.AddScoped(new MyDep("A string!"));` | No | No | Yes |
+
+For more information on type disposal, see the [Disposal of services](#disposal-of-services) section. For more information on mocking types for testing, see <xref:test/integration-tests#inject-mock-services>.
+
+`TryAdd{LIFETIME}` methods only register the service if there isn't already a registered instance of the service.
```
anurse:
```suggestion
`TryAdd{LIFETIME}` methods only register the service if there isn't already an implementation registered for the service.
```

I want to try and make it clear that TryAdd only checks the `{INTERFACE}` value. For example:

```csharp
services.TryAddSingleton<IFrob, DefaultFrob>();
services.TryAddSingleton<IFrob, AlternateFrob>(); // No-op, even though the implementation type is different
```

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -150,36 +150,78 @@ public void ConfigureServices(IServiceCollection services)
...
+Singleton lifetime services (<xref:Microsoft.AspNet.OData.Builder.ODataModelBuilder.AddSingleton*>) are created the first time they're requested (or when `Startup.ConfigureServices` is run and an instance is specified with the service registration). Every subsequent request uses the same instance. If the app requires singleton behavior, allowing the service container to manage the service's lifetime is recommended. Don't implement the singleton design pattern and provide user code to manage the object's lifetime in the class.
 
 > [!WARNING]
 > It's dangerous to resolve a scoped service from a singleton. It may cause the service to have incorrect state when processing subsequent requests.
 
+## Service registration methods
+
+Each service registration extension method offers overloads that are useful in specific scenarios.
+
+| Method | Automatic<br>object<br>disposal | Mocks for<br>testing | Pass args |
```
anurse:
```suggestion
| Method | Automatic<br>object<br>disposal | Alternate or multiple implementations | Pass args |
```

Maybe?

Or maybe just drop this column completely? The reasons on why you'd want to use an interface vs a concrete type are numerous and may be beyond the scope of this document (or this section at least). Mocking is one of many reasons.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -150,36 +150,78 @@ public void ConfigureServices(IServiceCollection services)
...
+| ------ | :-----------------------------: | :------------------: | :-------: |
+| `Add{LIFETIME}<{INTERFACE}, {TYPE}>()`<br>Example:<br>`services.AddScoped<IMyDep, MyDep>();` | Yes | Yes | No |
```
anurse:
```suggestion
| `Add{LIFETIME}<{SERVICE}, {IMPLEMENTATION}>()`<br>Example:<br>`services.AddScoped<IMyDep, MyDep>();` | Yes | Yes | No |
```

Unfortunately this is used throughout the table so it'll take more than just accepting this suggestion :).

We avoid using "interface" for this in our APIs because it isn't required to be an interface. The service can also be identified by an `abstract` (or even non-`abstract`) class. All that is required by DI is that the *implementation type* must be assignable to a variable/field/etc. of the *service type* (i.e. it implements the interface, or inherits from the class).

Most other uses of "interface" are in examples that actually choose to use an interface which is fine. I'm mostly trying to avoid using that in places that refere to the general concept of a DI service, rather than a specific example. I'll do a quick pass and see if I can find other places where it would be worth updating our language.

And if this is a little too in-the-weeds for this doc, that's fine too, we can leave it.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -63,7 +63,7 @@ The class creates and directly depends on the `MyDependency` instance. Code depe
 Dependency injection addresses these problems through:
 
 * The use of an interface to abstract the dependency implementation.
```
anurse: Perhaps:

```suggestion
* The use of an interface or base class to abstract the dependency implementation.
```

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -150,36 +150,78 @@ public void ConfigureServices(IServiceCollection services)
...
+```
+
+For more information, see:
+
+* <xref:Microsoft.Extensions.DependencyInjection.Extensions.ServiceCollectionDescriptorExtensions.TryAdd*>
+* <xref:Microsoft.Extensions.DependencyInjection.Extensions.ServiceCollectionDescriptorExtensions.TryAddTransient*>
+* <xref:Microsoft.Extensions.DependencyInjection.Extensions.ServiceCollectionDescriptorExtensions.TryAddScoped*>
+* <xref:Microsoft.Extensions.DependencyInjection.Extensions.ServiceCollectionDescriptorExtensions.TryAddSingleton*>
+
+[TryAddEnumerable(ServiceDescriptor)](xref:Microsoft.Extensions.DependencyInjection.Extensions.ServiceCollectionDescriptorExtensions.TryAddEnumerable*) methods only register the service if there isn't already a registered instance of a specific implementing type. Multiple services are resolved via `IEnumerable<{INTERFACE}>`; but when registering services, the the developer only wants to add an instance if one of the same type hasn't already been added. Generally, this method is used by library authors to avoid registering two copies of an instance in the container.
```
anurse:
```suggestion
[TryAddEnumerable(ServiceDescriptor)](xref:Microsoft.Extensions.DependencyInjection.Extensions.ServiceCollectionDescriptorExtensions.TryAddEnumerable*) methods only register the service if there isn't already an implementation **of the same type**. Multiple services are resolved via `IEnumerable<{SERVICE}>`; but when registering services, the the developer only wants to add an instance if one of the same type hasn't already been added. Generally, this method is used by library authors to avoid registering two copies of an instance in the container.
```

The thing I'm trying to clarify here is that **both** the service type and implementation type have to match for `TryAddEnumerable` to no-op. I may be fumbling for words a bit 😄. Also, maybe too esoteric to worry about here unless confused customers come knocking :)

For example:

```csharp
public interface IFoo {}
public interface IBar {}

public class FooBar: IFoo, IBar {}

// Registers an instance of FooBar
services.TryAddEnumerable(ServiceDescriptor.Singleton<IFoo, FooBar>());
// Registers a SECOND instance of FooBar with it's own lifetime
services.TryAddEnumerable(ServiceDescriptor.Singleton<IBar, FooBar>());
// No-ops because an (IFoo, FooBar) pair already exists.
services.TryAddEnumerable(ServiceDescriptor.Singleton<IFoo, FooBar>());
```

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -150,36 +150,78 @@ public void ConfigureServices(IServiceCollection services)
...
+| `Add{LIFETIME}<{INTERFACE}>(sp => new {TYPE})`<br>Examples:<br>`services.AddScoped<IMyDep>(sp => new MyDep());`<br>`services.AddScoped<IMyDep>(sp => new MyDep("A string!"));` | Yes | Yes | Yes |
+| `Add{LIFETIME}<{TYPE}>()`<br>Example:<br>`services.AddScoped<MyDep>();` | Yes | No | No |
+| `Add{LIFETIME}<{INTERFACE}>(new {TYPE})`<br>Examples:<br>`services.AddScoped<IMyDep>(new MyDep());`<br>`services.AddScoped<IMyDep>(new MyDep("A string!"));` | No | Yes | Yes |
+| `Add{LIFETIME}(new {TYPE})`<br>Examples:<br>`services.AddScoped(new MyDep());`<br>`services.AddScoped(new MyDep("A string!"));` | No | No | Yes |
+
+For more information on type disposal, see the [Disposal of services](#disposal-of-services) section. For more information on mocking types for testing, see <xref:test/integration-tests#inject-mock-services>.
+
+`TryAdd{LIFETIME}` methods only register the service if there isn't already an implementation registered for the service.
+
+In the following example, the first line registers `MyDependency` for `IMyDependency`. The second line no-ops because `IMyDependency` already has a registered type:
```
anurse: I know I used it in my comment but is "no-ops" a phrase we're comfortable with? I know that sometimes jargon terms can be problematic, but perhaps this is common enough. Another option would be

```suggestion
In the following example, the first line registers `MyDependency` for `IMyDependency`. The second line has no effect because `IMyDependency` already has a registered implementation:
```

(with a similar change to the code comment below)

anurse: We should also clarify what we mean by "registered type". Plus we use "... implementation registered for the service" above. My above suggestion in re: "no-ops" incorporates that, but if you don't want to take that change here's one without the 'no-ops' change ;)

```suggestion
In the following example, the first line registers `MyDependency` for `IMyDependency`. The second line no-ops because `IMyDependency` already has a registered implementation:
```

aspnetcore/fundamentals/dependency-injection.md
anurse: Even if we don't take that full change, we should change to `IEnumerable<{SERVICE}>` as per my comments above about prefering "service" over "implementation"

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -150,36 +150,78 @@ public void ConfigureServices(IServiceCollection services)
...
+* <xref:Microsoft.Extensions.DependencyInjection.Extensions.ServiceCollectionDescriptorExtensions.TryAddTransient*>
+* <xref:Microsoft.Extensions.DependencyInjection.Extensions.ServiceCollectionDescriptorExtensions.TryAddScoped*>
+* <xref:Microsoft.Extensions.DependencyInjection.Extensions.ServiceCollectionDescriptorExtensions.TryAddSingleton*>
+
+[TryAddEnumerable(ServiceDescriptor)](xref:Microsoft.Extensions.DependencyInjection.Extensions.ServiceCollectionDescriptorExtensions.TryAddEnumerable*) methods only register the service if there isn't already a registered instance of a specific implementing type. Multiple services are resolved via `IEnumerable<{INTERFACE}>`; but when registering services, the the developer only wants to add an instance if one of the same type hasn't already been added. Generally, this method is used by library authors to avoid registering two copies of an instance in the container.
+
+In the following example, the first line registers `MyDependency` for `IMyDependency`. The second line registers `DifferentDependency`. The third line no-ops because `IMyDependency` already has a registered type of `MyDependency`:
```
anurse: Same comment re: "no-ops"

guardrex:
I think I was able to get all of those in there.

I wonder if "Alternate or multiple implementations" can just be "Multiple implementations"? ... because if there are multiple impl, then they probably are *alternates* ... otherwise, what would be the point.

anurse:
> I wonder if "Alternate or multiple implementations" can just be "Multiple implementations"

Sounds fine :)

Issue #13219, opened 2019-07-10
Add Vegeta link to load test third-party tools list.


apecnascimento:
Vegeta is a versatile HTTP load testing tool, and deserve one place in this list.

scottaddie:
Fixed by https://github.com/aspnet/AspNetCore.Docs/pull/13220

PR #13220, opened 2019-07-10
Add Vegeta to third party tools list.  #13219

apecnascimento:
- Add Vegeta to third party tools list.


scottaddie:
Thank you for adding this tool to the list, @apecnascimento!

PR #13241, opened 2019-07-11
Fix http to https in CATS report 6-21-19

v-thepet:
Per User Story https://mseng.visualstudio.com/TechnicalContent/_queries/edit/1562764


v-thepet:
@tdykstra - Here are the http to https fixes for the CATS report. Sorry for the size of the PR but it's all that one fix. Spot checked on staging and all is good. Thanks.

PR #13270, opened 2019-07-12
C101893: Fixing table data to avoid localization issues

v-maudel:
Description:  Missing dash is preventing content to show properly.


v-maudel:
Thanks to you @guardrex

PR #13274, opened 2019-07-13
Update the choice text for a Blazor server-side experience.

britebit:
This text now matches the dialogue box in Visual Studio 2019 16.2.0 Preview 3.

### DIFF
--- a/aspnetcore/blazor/get-started.md
+++ b/aspnetcore/blazor/get-started.md
@@ -38,7 +38,7 @@ Get started with Blazor:
 
    6\. In the **Create a new ASP.NET Core Web Application** dialog, confirm that **.NET Core** and **ASP.NET Core 3.0** are selected.
 
-   7\. For a Blazor client-side experience, choose the **Blazor (client-side)** template. For a Blazor server-side experience, choose the **Blazor (server-side)** template. Select **Create**. For information on the two Blazor hosting models, server-side and client-side, see <xref:blazor/hosting-models>.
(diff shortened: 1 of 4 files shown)


guardrex:
Hello @britebit ... The current text is correct. The server-side experience should be called **Blazor (server-side)**. The "hosted" template is a client-side Blazor app hosted by an ASP.NET Core app.

Capture

britebit:
Ahh .. my bad. I made an assumption about which option was server-side.  I think the docs still need updating. I downloaded the VS 2019 preview today and saw the options below. I think we need to change `Blazor (server-side)` to `Blazor Server App`. Would you agree? If so, I'll raise another PR.

image

guardrex:
Yes, that's great. As you already surmised, I'm on 16.1.6 (not Preview). Let's do that right here and for `master` as you did given that we instruct the use of VS Preview for all Blazor apps at this time.

1. Search the entire repo for the text "Blazor (server-side)" ... five results across four files. One of them is in commented-out text for MS4Mac.
1. Replace those five instances with the new text.
1. Commit the updates back to this PR.

guardrex:
If you agree with the CLA, sign the agreement :point_up:. We can't accept the PR without it.

britebit:
Is this the one, or do I need to do something else?

> [CLA assistant check](https://cla.dotnetfoundation.org/aspnet/AspNetCore.Docs?pullRequest=13274)
> All CLA requirements met.

PR #13279, opened 2019-07-13
Update filters.md

### LINKED ISSUE #13239: Order doesn't seem to apply to ExceptionFilter
Not sure if this is a particularity or a bug, but when I specify the order of the exception filters, they do not run in that order. They don't seem to be affected if I change their physical order (the order in which they were added to the filter collection) either.

``` csharp
services.AddMvc(options =>
{
	options.Filters.AddService(typeof(PetmApplicationExceptionFilter), 2);
	options.Filters.AddService(typeof(ConfigurationExceptionFilter), 1);
})
```

---
#### Document Details

* ID: a70a955f-dd5a-c46b-dfe8-21521e619320
* Version Independent ID: ebb299e0-6c1c-7077-689d-ac711352e418
* Content: [Filters in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/filters?view=aspnetcore-2.2)
* Content Source: [aspnetcore/mvc/controllers/filters.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/mvc/controllers/filters.md)
* Product: **aspnet-core**
* GitHub Login: @ardalis
* Microsoft Alias: **riande**

Rick-Anderson:
Fixes #13239

### DIFF
--- a/aspnetcore/mvc/controllers/filters.md
+++ b/aspnetcore/mvc/controllers/filters.md
@@ -393,6 +393,7 @@ The `OnActionExecuted` method runs after the action method:
 
 Exception filters:
 
+* 
 * Implement <xref:Microsoft.AspNetCore.Mvc.Filters.IExceptionFilter> or <xref:Microsoft.AspNetCore.Mvc.Filters.IAsyncExceptionFilter>. 
 * Can be used to implement common error handling policies.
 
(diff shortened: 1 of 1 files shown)


aspnetcore/mvc/controllers/filters.md
```diff
@@ -403,6 +404,7 @@ The following sample exception filter uses a custom error view to display detail
 Exception filters:
 
 * Don't have before and after events.
+* Can't override the order of execution with the `Order` property.
```
dougbu: The `Order` value _does_ apply to `ExceptionFilter`s that implement `IOrderedFilter`. Unless we have a repro of a case where that isn't the case, this PR and #13239 should be closed. I'm dubious such a repro is possible.

PR #13283, opened 2019-07-14
Minor tweaks to "IP Safelist".

serpent5:
I was reading through [Client IP safelist for ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/ip-safelist?view=aspnetcore-2.2) (uid: security/ip-safelist) and I noticed some minor errors:

* 3 approaches are listed and then the text explains how *both* approaches are in the sample.
* A line highlight for the second snippet is off.
* One of the code examples uses an enum for a 403, whereas the others use 401 and they use it directly.

You'll see that I collapsed the code that needed to be highlighted onto 1 line, as this works out to be exactly 80 characters, which I've read is the max for docs pages.

---

I think the page could do with some more work, which I'm happy to raise as an issue if this isn't already on the radar. Of course, I'd also be happy to PR something for that too, once I've had a look at a couple of other PRs I've volunteered for. :sweat_smile:

### DIFF
--- a/aspnetcore/security/ip-safelist.md
+++ b/aspnetcore/security/ip-safelist.md
@@ -17,7 +17,7 @@ This article shows three ways to implement an IP safelist (also known as a white
 * Action filters to check the remote IP address of requests for specific controllers or action methods.
 * Razor Pages filters to check the remote IP address of requests for Razor pages.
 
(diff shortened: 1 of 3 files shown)


serpent5:
@Rick-Anderson Would you prefer it if I open an issue and PR against that next time I do something like this?

Rick-Anderson:
> @Rick-Anderson Would you prefer it if I open an issue and PR against that next time I do something like this?

Yes, that's our normal work flow.

serpent5:
> Thanks for these corrections! I made one suggestion.

Thanks @tdykstra. That’s definitely easier to maintain.

tdykstra:
Fixes #13285

PR #13358, opened 2019-07-17
Expand built-in filters tasks description

WindOfMind:
Expand built-in filters handling tasks description in order to make it consistent with, for example, https://docs.microsoft.com/en-us/aspnet/mvc/overview/older-versions-1/controllers-and-routing/understanding-action-filters-cs

### DIFF
--- a/aspnetcore/mvc/controllers/filters.md
+++ b/aspnetcore/mvc/controllers/filters.md
@@ -4,7 +4,7 @@ author: ardalis
 description: Learn how filters work and how to use them in ASP.NET Core.
 ms.author: riande
 ms.custom: mvc
-ms.date: 05/08/2019
+ms.date: 07/17/2019
 uid: mvc/controllers/filters
 ---
 # Filters in ASP.NET Core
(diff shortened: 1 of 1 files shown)


serpent5:
@guardrex Correct me if I'm wrong, but there isn't a *built-in* filter for handling errors in ASP.NET Core. There's [`HandleError`](https://docs.microsoft.com/en-us/dotnet/api/system.web.mvc.handleerrorattribute?view=aspnet-mvc-5.2) in ASP.NET 4 but that didn't make it across.

guardrex:
Is this too "custom" to be considered "built-in"? It's probably ok, but let me know if you disagree, and we'll get more :eyes: on this.

https://docs.microsoft.com/aspnet/core/mvc/controllers/filters#exception-filters

serpent5:
That's just an example of a custom exception filter, but it doesn't exist "out of the box", which is what I interpret "built-in" to mean. For example, there are built-in authorisation filters (e.g. `AuthorizeFilter`) and built-in response caching filters (e.g. `ResponseCacheFilter`), but there's no implementation of `IExceptionFilter` (or the async version) in the code base.

I'm more than happy to go with you on this, but I wanted to raise it for consideration. I'm not gonna be offended if you say thanks but no thanks :)

guardrex:
Sure ... it's abstract. You're the main author, so let's go with your recommendation.

serpent5:
I wouldn't say "main author", but I guess the page does read that way. Honestly, I'm still chuffed to be on the page at all. Thanks for going with my suggestion and apologies for creating more work for you. :disappointed:

guardrex:
We try to take PRs because it encourages folks to contribute more updates to the docs. I suspect @WindOfMind understands where we're coming from here. It's *almost built-in* ... er ... *sort-of built-in*. :smile:

Since the whole topic pertains to filters generally, an alternative is to change the lead-in line to ...

> Built-in and custom filters handle tasks such as:

... and then that added line would be fine. However, it probably isn't quite worth the trouble.

serpent5:
Sure. I'll be much more mindful of this contribution encouragement problem in future. I really didn't mean to be discouraging. I completely understand where @WindOfMind is coming from and I :clap: the desire for consistency.

PR #13374, opened 2019-07-17
Updating help output to match change in product

tlmii:
Making a quick change to the docs to match a change in the product.


Rick-Anderson:
@tlmii  thanks!

PR #13376, opened 2019-07-18
Remove curly braces from proto rpcs

JamesNK:
This is the most common syntax for declaring proto rpcs


JamesNK:
One sec, a new proto file to update...

PR #13407, opened 2019-07-19
Blazor Handle Errors topic

### LINKED ISSUE #13288: Doc proposal for error handling in server-side Blazor
As part of https://github.com/aspnet/AspNetCore/issues/11794, I've written up some guidance around error handling scenarios in server-side Blazor. This information is now in a gist at https://gist.github.com/SteveSandersonMS/9451f3b5497ce2b5ad16b0d07ad73539

Just like with https://github.com/aspnet/AspNetCore.Docs/issues/13259, will it be possible to convert this into a doc page? Please feel free to edit however you wish! cc @guardrex 

There are a couple of bits in the text where it would have been nice to link to other docs (for example, about circuit handlers or prerendering), but I couldn't spot any suitable docs to link to. This might indicate that we're missing docs on those other topics. Obviously that's a separate matter, but I'm just mentioning it now in case it's something you wanted to track.

guardrex:
Fixes #13288 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/handle-errors?view=aspnetcore-3.0&branch=pr-en-us-13407)

* Goes with "handle" over the participle ("handling"). More *verb-y-ish*:tm: (action based) this way. We gave this treatment to the ASP.NET Core "error handling" topic a while back as far as is permitted (e.g., title), not for its UID or URL. Here, we can start with the better option.

### DIFF
--- a/aspnetcore/blazor/handle-errors.md
+++ b/aspnetcore/blazor/handle-errors.md
@@ -5,7 +5,7 @@ description: Discover how ASP.NET Core Blazor how Blazor manages unhandled excep
 monikerRange: '>= aspnetcore-3.0'
 ms.author: riande
 ms.custom: mvc
-ms.date: 08/03/2019
+ms.date: 08/06/2019
 uid: blazor/handle-errors
 ---
 # Handle errors in ASP.NET Core Blazor apps
@@ -25,13 +25,20 @@ If a user opens the app in multiple browser tabs, they have multiple independent
 
 Blazor treats most unhandled exceptions as fatal to the circuit where they occur. If a circuit is terminated due to an unhandled exception, the user can only continue to interact with the app by reloading the page to create a new circuit. Circuits outside of the one that's terminated, which are circuits for other users or other browser tabs, aren't affected. This scenario is similar to a desktop app that crashes&mdash;the crashed app must be restarted, but other apps aren't affected.
 
-The reason for terminating a circuit for most unhandled exceptions is that an unhandled exception often leaves the circuit in an undefined state. If the framework continues processing a component's code after an unhandled exception and regardless of the circuit's state, the app's normal operation can't be guaranteed. Even security vulnerabilities in the app may result.
+A circuit is terminated when an unhandled exception occurs for the following reasons:
+
+* An unhandled exception often leaves the circuit in an undefined state.
+* The app's normal operation can't be guaranteed after an unhandled exception.
+* Security vulnerabilities may appear in the app if the circuit continues.
 
 ## Manage unhandled exceptions in developer code
 
-If you want users to be able to continue using an app after an error occurs (for example, when loading or saving data), develop the app with error handling logic. Later sections of this article describe potential sources of unhandled exceptions.
+For an app to continue after an error, the app must have error handling logic. Later sections of this article describe potential sources of unhandled exceptions.
 
-Be careful not to disclose sensitive information to end users. Under normal conditions in a production app, don't render framework exception messages or stack traces in the UI, as doing so may help a malicious user discover weaknesses in an app that can compromise the security of the app, server, or network.
+In production, don't render framework exception messages or stack traces in the UI. Rendering exception messages or stack traces could:
+
+* Disclose sensitive information to end users.
+* Help a malicious user discover weaknesses in an app that can compromise the security of the app, server, or network.
 
 ## Log errors with a persistent provider
 
@@ -39,11 +46,11 @@ If an unhandled exception occurs, the exception is logged to <xref:Microsoft.Ext
 
 During development, Blazor usually sends the full details of exceptions to the browser's console to aid in debugging. In production, detailed errors in the browser's console are disabled by default, which means that errors aren't sent to clients but the exception's full details are still logged server-side. For more information, see <xref:fundamentals/error-handling>.
 
-You must decide which incidents to log and the level of severity of logged incidents. Note that hostile users might be able to trigger errors deliberately (for example, by supplying an unknown `ProductId` in the URL of a component that displays product details), so you shouldn't necessarily treat all errors as high-severity incidents with logging.
+You must decide which incidents to log and the level of severity of logged incidents. Hostile users might be able to trigger errors deliberately. For example, don't log an incident from an error where an unknown `ProductId` is supplied in the URL of a component that displays product details. Not all errors should be treated as high-severity incidents for logging.
 
 ## Places where errors may occur
 
-Framework and app code may trigger unhandled exceptions in any of the following locations, which are described in the following sections of this article:
+Framework and app code may trigger unhandled exceptions in any of the following locations:
 
 * [Component instantiation](#component-instantiation)
 * [Lifecycle methods](#lifecycle-methods)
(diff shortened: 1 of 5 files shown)


aspnetcore/blazor/handle-errors.md
```diff
@@ -0,0 +1,214 @@
...
+ms.author: riande
+ms.custom: mvc
+ms.date: 07/20/2019
+uid: blazor/handle-errors
+---
+# Handle errors in ASP.NET Core Blazor apps
+
+By [Steve Sanderson](https://github.com/SteveSandersonMS) and [Luke Latham](https://github.com/guardrex)
+
+This article describes how Blazor manages unhandled exceptions and how to develop apps to detect and handle errors.
```
SteveSandersonMS:
```suggestion
This article describes how Blazor manages unhandled exceptions and how to develop apps that detect and handle errors.
```

Suggested phrasing is because people won't build an app whose primary purpose is to detect and handle errors. Rather, that's one of the secondary concerns of app building.

aspnetcore/blazor/handle-errors.md
```diff
@@ -0,0 +1,214 @@
...
+The reason for terminating a circuit for most unhandled exceptions is that an unhandled exception often leaves the circuit in an undefined state. If the framework continues processing a component's code after an unhandled exception and regardless of the circuit's state, the app's normal operation can't be guaranteed. Even security vulnerabilities in the app may result.
+
+## Manage unhandled exceptions in developer code
+
+If you want users to be able to continue using an app after an error occurs (for example, when loading or saving data), develop the app with error handling logic. C# [try-catch](/dotnet/csharp/language-reference/keywords/try-catch) statements are useful in error handling scenarios and are a suggested approach in later sections of this article that describe potential sources of unhandled exceptions.
+
+The structure of a `try-catch` statement includes two *blocks* of code:
+
+* The `try` block contains code that may throw an exception.
+* The `catch` block contains code that executes when an exception occurs in the `try` block.
```
SteveSandersonMS: TBH I'm surprised we would introduce aspects of basic C# syntax in this doc, but up to you to judge that!

aspnetcore/blazor/handle-errors.md
```diff
@@ -0,0 +1,214 @@
...
+{
+    ...
+}
+catch (Exception ex)
+{
+    ...
+}
+```
+
+In many C# development scenarios, trapping *all exceptions* with <xref:System.Exception?displayProperty=fullName> isn't considered a *best practice*. Usually, a more specific exception is trapped because:
```
SteveSandersonMS: I don't think we should be talking about general exception handling practices in this doc. It's not specific to Blazor at all. 

Plus I don't really even agree with this advice. In C# nobody is very disciplined about what types of exceptions get thrown from where (including the BCL), so very often the only sane way to handle errors is to catch `Exception`. There are always going to be unknown unknowns.

aspnetcore/blazor/handle-errors.md
```diff
@@ -0,0 +1,214 @@
...
+
+* You should only catch those exceptions that you know how to recover from.
+* Specific exceptions can be handled in a fine-grained manner:
+  * Specific method calls and property settings can be made by the app for a specific error condition.
+  * Specific error messages can be logged and shown to the user.
+
+In Blazor exception scenarios, however, we recommend trapping all exceptions in components where unhandled exceptions may occur. Trapping all exceptions provides an opportunity to:
```
SteveSandersonMS: I think it's very strange to make this distinction as if it's different for Blazor. The same advice applies to basically all .NET programming.

aspnetcore/blazor/handle-errors.md
```diff
@@ -0,0 +1,214 @@
...
+
+* Update the rendered component UI with a general exception message for the user.
+* Log a general exception for debugging.
+
+This approach usually avoids a circuit failure and allows a user to continue interacting with a component.
+
+`try-catch` statements can be nested inside of other `try-catch` statements. When nested statements are used, more specific exceptions can be caught before the most general type of exception is trapped with `System.Exception`.
```
SteveSandersonMS: Again, super weird to talk about this here

aspnetcore/blazor/handle-errors/samples_snapshot/3.x/product-details.razor
```diff
@@ -0,0 +1,40 @@
...
+    {
+        try
+        {
+            loadFailed = false;
+            details = await ProductRepository.GetProductByIdAsync(ProductId);
+        }
+        catch (Exception ex)
+        {
+            loadFailed = true;
+            Logger.LogWarning(ex, "Failed to load product {PRODUCT_ID}", ProductId);
```
SteveSandersonMS: Is `PRODUCT_ID` consistent with the standard naming convention for structured logging in our docs?

Maybe it is; I don't know. In the product sources we'd use a name like `ProductID` instead.

aspnetcore/blazor/handle-errors.md
guardrex: Primarily only to address what I'm about to describe below. Stand-by .......

aspnetcore/blazor/handle-errors.md
guardrex: That's what we doc in official guidance for `try-catch`. @danroth27, what's best? ...

* Ignore what the `try-catch` doc says and just roll with the cross-linking. (I feel like someone is going to call us out on it.)
* Change the `try-catch` doc to say that it's ok to trap `System.Exception`. Proceed with only the cross-link. If so, I'll submit a PR over there and delete this content.
* Do what I do here. Say that 'in this scenario, it's ok.' Cross-link it, too.

aspnetcore/blazor/handle-errors.md
guardrex: For newbies ... but *sure thing*. I'm not super-excited about this coverage.

aspnetcore/blazor/handle-errors/samples_snapshot/3.x/product-details.razor
guardrex: Yes ... consistent. We uppercase template placeholders in other docs.

guardrex:
Thanks @SteveSandersonMS ... do you have any feedback on these ❓ ...

1. The gist doesn't address how to manage exceptions in the *component instantiation* error scenario. How can we guide devs there?
1. Consider providing an example for `try-catch` with `Html.RenderComponentAsync` (*Prerendering* section). I can *dream* one up, but it will probably be more of a *nightmare* 🎃 compared to what you'll suggest.
1. In the *Recursive rendering* section ...

   > To avoid infinite recursion patterns, ensure that recursive rendering code contains suitable stopping conditions or enforces invariants (provide code that imposes a strict specification on recursive sections of code to prevent such code from executing infinitely).

   I added the parenthetical bit at the end because "enforces invariants" alone might not be dead clear in this context (e.g., for newbies). However, isn't "enforcing invariants" basically the same thing as "suitable stopping conditiions" in this context? If so, can the sentence just read ...

   > To avoid infinite recursion patterns, ensure that recursive rendering code contains suitable stopping conditions.

SteveSandersonMS:
> The gist doesn't address how to manage exceptions in the component instantiation error scenario. How can we guide devs there?

People can also use `try`/`catch` in constructors if they have some particular reason to believe some code may throw but it's safe to continue somehow.

> Consider providing an example for try-catch with Html.RenderComponentAsync

It's literally just to put try/catch in the markup:

```razor
@try
{    
  @Html.RenderComponentAsync<MyComponent>()
}
catch (Exception ex)
{
  <div>Oops</div>
}
```

However I'm not keen on emphasising this as an option, because people aren't realistically going to log their exceptions well or otherwise reason about why they are getting an exception. It would be better to put the error handling into `MyComponent` instead, rather than wrapping it around the `Html.RenderComponentAsync` call.

> To avoid infinite recursion patterns, ensure that recursive rendering code contains suitable stopping conditions.

That sounds perfect to me!

guardrex:
Thanks ... I'm going to get @danroth27 feedback on both topics ... fix them both up ... ping for a final :eyes:.

aspnetcore/blazor/handle-errors/samples_snapshot/3.x/product-details.razor
guardrex: *Ooops!* ... I researched this. We started with uppercase template placeholders (the framework code at the time did both), then we changed it in logging topics to go with Pascal case.

Hard to remember a two-minute convo from **_two years ago!_** :smile: lol

I'll revert this on the updates.

aspnetcore/blazor/handle-errors.md
```diff
@@ -0,0 +1,180 @@
...
+
+The reason for terminating a circuit for most unhandled exceptions is that an unhandled exception often leaves the circuit in an undefined state. If the framework continues processing a component's code after an unhandled exception and regardless of the circuit's state, the app's normal operation can't be guaranteed. Even security vulnerabilities in the app may result.
+
+## Manage unhandled exceptions in developer code
+
+If you want users to be able to continue using an app after an error occurs (for example, when loading or saving data), develop the app with error handling logic. C# [try-catch](/dotnet/csharp/language-reference/keywords/try-catch) statements are useful in error handling scenarios and are a suggested approach in later sections of this article that describe potential sources of unhandled exceptions.
```
danroth27: I think we can assume the reader knows what a try-catch is.

```suggestion
If you want users to be able to continue using an app after an error occurs (for example, when loading or saving data), develop the app with error handling logic. Later sections of this article describe potential sources of unhandled exceptions.
```

aspnetcore/blazor/handle-errors.md
```diff
@@ -0,0 +1,180 @@
...
+
+Be careful not to disclose sensitive information to end users. Under normal conditions in a production app, don't render framework exception messages or stack traces in the UI, as doing so may help a malicious user discover weaknesses in an app that can compromise the security of the app, server, or network.
+
+## Log errors with a persistent provider
+
+If an unhandled exception occurs, the exception is logged to <xref:Microsoft.Extensions.Logging.ILogger> instances configured in the service container. By default, template-based ASP.NET Core Blazor apps log to console output with the Console Logging Provider. Consider logging to a more permanent location with a provider that manages log size and log rotation. For more information, see <xref:fundamentals/logging/index>.
```
danroth27:
```suggestion
If an unhandled exception occurs, the exception is logged to <xref:Microsoft.Extensions.Logging.ILogger> instances configured in the service container. By default, Blazor apps log to console output with the Console Logging Provider. Consider logging to a more permanent location with a provider that manages log size and log rotation. For more information, see <xref:fundamentals/logging/index>.
```

Rick-Anderson:
@guardrex working on this. Too many long sentence violations so I'm doing a PR.

guardrex:
It was converted from engineering, so that's not very surprising.

Rick-Anderson:
Waiting for approval to S&M #13677 before we S&M this.

PR #13466, opened 2019-07-24
Update get-started.md

harleydk:
Removed reference to the blazor marketplace extension. When I go to that marketplace-page, I'm informed the extension is now obsolete.


guardrex:
Hello @harleydk ... Thanks, but we already have this change on https://github.com/aspnet/AspNetCore.Docs/pull/13460, which will be merging this morning.

PR #13471, opened 2019-07-24
Update DI example

### LINKED ISSUE #13450: Usage of options.CurrentValue
If in the example about avoiding service locator pattern MyClass is registered as a singleton, the CurrentValue returned in the constructor will stale as soon as the options monitor updates options. In order to avoid that you need to retrieve CurrentValue in the method where options are used.

---
#### Document Details

* Content: [Dependency injection in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.2#feedback)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #13450

* In this example (instructing the reader to avoid the service locator pattern), `MyClass` isn't a registered service ... it's just a class example ... so it's fine in its present form. However, @moikot correctly calls out that *IF* the pattern shown in the example were to be adopted by a singleton service that it won't work and probably isn't the best example.
* Let's inject `IOptionsMonitor<MyOptions>`, then pull the `CurrentValue` in the `MyMethod`. That will always work if this pattern were to become part of a singleton.
* This whole topic is coming up for a 3.0 update in a few weeks. I'll comb thru it (and take a look at the Options topic when it comes up) to see if any other little *gotchas* like this might be lurking.

Thanks @moikot for calling this out. 🎸

### DIFF
--- a/aspnetcore/fundamentals/dependency-injection.md
+++ b/aspnetcore/fundamentals/dependency-injection.md
@@ -5,7 +5,7 @@ description: Learn how ASP.NET Core implements dependency injection and how to u
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 07/09/2019
+ms.date: 07/24/2019
 uid: fundamentals/dependency-injection
 ---
 # Dependency injection in ASP.NET Core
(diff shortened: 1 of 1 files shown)


guardrex:
@HaoK, we also can go with a full service-based pair of examples here. Use `IMyDependency` and `MyDependency`. That might be overkill tho. The pattern won't really change if we do that, so I'm not sure if it really adds anything to what these examples are trying to demo.

HaoK:
This update is good, this is the general recommended pattern

PR #13522, opened 2019-07-26
Add some hints on how to use tag helpers and areas together

ivarne:
The standard `asp-*` tag-helpers did not work when used from a view inside an area.

Could probably be written better, but this would at least save me some time, because I'd probably move `_ViewImports.cshtml` togheter with `_ViewStart.cshtml` when I started using Areas.

For current users it is good that `ctrl-f` `area` on [tag heplpers](https://docs.microsoft.com/en-us/aspnet/core/mvc/views/tag-helpers/built-in/?view=aspnetcore-3.0) would point directly to hints on what to do.

### DIFF
--- a/aspnetcore/mvc/controllers/areas.md
+++ b/aspnetcore/mvc/controllers/areas.md
@@ -3,7 +3,7 @@ title: Areas in ASP.NET Core
 author: rick-anderson
 description: Learn how Areas are an ASP.NET MVC feature used to organize related functionality into a group as a separate namespace (for routing) and folder structure (for views).
 ms.author: riande
-ms.date: 05/10/2019
+ms.date: 08/07/2019
 uid: mvc/controllers/areas
 ---
 # Areas in ASP.NET Core
@@ -116,7 +116,7 @@ To share a common layout for the entire app, move the *_ViewStart.cshtml* to the
 
 ### _ViewImports.cshtml
 
-In its standard location */Views/_ViewImports.cshtml* does not apply to areas. To use common [tag helpers](xref:mvc/views/tag-helpers/intro), `@using` or `@inject` in your area, ensure a proper *_ViewImports.cshtml* file [applies to your area views](xref:/mvc/views/layout#importing-shared-directives). If you want the same behavior in all your views, just move */Views/_ViewImports.cshtml* to application root.
+In its standard location, */Views/_ViewImports.cshtml* doesn't apply to areas. To use common [Tag Helpers](xref:mvc/views/tag-helpers/intro), `@using`, or `@inject` in your area, ensure a proper *_ViewImports.cshtml* file [applies to your area views](xref:/mvc/views/layout#importing-shared-directives). If you want the same behavior in all your views, move */Views/_ViewImports.cshtml* to the application root.
 
 <a name="rename"></a>
 
--- a/aspnetcore/mvc/controllers/areas.md
+++ b/aspnetcore/mvc/controllers/areas.md
@@ -114,6 +114,10 @@ For more information, see [Routing to controller actions](xref:mvc/controllers/r
 
 To share a common layout for the entire app, move the *_ViewStart.cshtml* to the application root folder.
 
+### _ViewImports.cshtml
+
+In its standard location */Views/_ViewImports.cshtml* does not apply to areas. To use common [tag helpers](xref:mvc/views/tag-helpers/intro), `@using` or `@inject` in your area, ensure a proper *_ViewImports.cshtml* file [applies to your area views](xref:/mvc/views/layout#importing-shared-directives). If you want the same behavior in all your views, just move */Views/_ViewImports.cshtml* to application root.
+
 <a name="rename"></a>
 
 ### Change default area folder where views are stored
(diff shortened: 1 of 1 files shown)


aspnetcore/mvc/controllers/areas.md
```diff
@@ -110,9 +110,9 @@ When the area or controller is not specified, routing depends on the *ambient* v
 
 For more information, see [Routing to controller actions](xref:mvc/controllers/routing).
 
-### Shared layout for Areas using the _ViewStart.cshtml file
+### Shared layout and tag helpers for Areas using the _ViewStart.cshtml and _ViewImports.cshtml file
```
NTaylorMullen: I would just add another section instead of mutating this one.

aspnetcore/mvc/views/tag-helpers/intro.md
```diff
@@ -56,7 +56,7 @@ The `asp-for` attribute is made available by the `For` property in the [LabelTag
 
 ## Managing Tag Helper scope
 
-Tag Helpers scope is controlled by a combination of `@addTagHelper`, `@removeTagHelper`, and the "!" opt-out character.
+Tag Helpers scope is controlled by a combination of `@addTagHelper`, `@removeTagHelper`, and the "!" opt-out character. If the standard `asp-` tag helpers don't work in an [area](xref:mvc/controllers/areas), you probably need to move *Views/_ViewImports.cshtml* to the root of your project, in order for `@addTagHelper` to affect area views.
```
NTaylorMullen: Ehh, I wouldn't have this here. The section in areas is sufficient if a user is working with areas.

ivarne: I kind of agree, but this is where I looked when i had the problem. If I expand the explanation in the area docs to make it even more obvious that you need to consider moving `_ViewImports.cshtml` it will probably be sufficient.

ivarne:
@NTaylorMullen I think I have addressed your comments.

aspnetcore/mvc/controllers/areas.md
```diff
@@ -114,6 +114,10 @@ For more information, see [Routing to controller actions](xref:mvc/controllers/r
 
 To share a common layout for the entire app, move the *_ViewStart.cshtml* to the application root folder.
 
+### Shared tag helpers using the _ViewImports.cshtml file
+
+To share a common configuration of [tag helpers](xref:mvc/views/tag-helpers/intro) (eg: for `<a asp-action="Index">` to be converted to correct links), move the *_ViewImports.cshtml* to the application root folder.
```
NTaylorMullen:
```suggestion
To ensure [tag helpers](xref:mvc/views/tag-helpers/intro) apply in your area ensure a proper *_ViewImports.cshtml* file exists at the application root or in your areas folder.
```

aspnetcore/mvc/controllers/areas.md
```diff
@@ -114,6 +114,10 @@ For more information, see [Routing to controller actions](xref:mvc/controllers/r
 
 To share a common layout for the entire app, move the *_ViewStart.cshtml* to the application root folder.
 
+### Shared tag helpers using the _ViewImports.cshtml file
```
NTaylorMullen:
```suggestion
### Tag helpers in areas
```

aspnetcore/mvc/controllers/areas.md
ivarne: Why not suggest to move the file that the template generates (as we do for layout)?

I don't have any data to back it up, but it seems more common to share *_ViewImports.cshtml* than to have separate for each area in an application.

NTaylorMullen: > Why not suggest to move the file that the template generates (as we do for layout)?

Ehh, _ViewImports can be used for many different reasons and for projects that have lots of TagHelpers developers typically want to separate out where certain TagHelpers are applicable.  In simple applications it's very common to only have one _ViewImports shared globally however in larger applications you'll see more _ViewImports interspersed throughout.

ivarne: I'm probably just more focused on helping others not spend hours to figure out why `<a asp-action="List">List</a>` doesn't work by default in an area, and suggest a quick fix. If you want to use ViewImports more heavily, you probably know their hirarctic nature and can adjust accordingly.

How about

 ### *_ViewImports.cshtml*
In its standard location, */Views/_ViewImports.cshtml* does not apply to areas. To use eg [tag helpers](xref:mvc/views/tag-helpers/intro), `@using` or `@inject` in your area, ensure a proper *_ViewImports.cshtml* file [apply to your area views](xref:/mvc/views/layout#importing-shared-directives). If you want the same behaviour in all your views, just move */Views/_ViewImports.cshtml* to application root.

NTaylorMullen: That sounds fine. I'd just make `apply` plural

ivarne: Ok. I changed `eg tag helpers, ...` to `common tag helpers, ....` as well.

scottaddie:
@ivarne Our team can take care of the metadata (`ms.`) changes for your PRs. Thank you for your help with this one! Your change will go live within the next 24 hours.

PR #13523, opened 2019-07-26
Document public [Parameter] expectation in 3.0.0-preview8.

### LINKED ISSUE #13453: [Preview 8] Add doc for Blazor [Parameter] usage
In 3.0.0-preview8 we added an analyzer to suggest users have public parameters. We should document this expectation and all of the details that reside in it.

NTaylorMullen:
- A lot of the parameter documentation was already present (such as not setting parameters outside of their declaring class).
- Updated all `[Parameter]` locations to utilize `public` properties with `public` setters.
- Updated all the locations I could find where we were suggesting users to utilize non-public parameters.

Fixes #13453

### DIFF
--- a/aspnetcore/blazor/common/samples/3.x/BlazorSample/Components/ChildComponent.razor
+++ b/aspnetcore/blazor/common/samples/3.x/BlazorSample/Components/ChildComponent.razor
@@ -9,12 +9,12 @@
 
 @code {
     [Parameter]
-    private string Title { get; set; }
+    public string Title { get; set; }
 
     [Parameter]
-    private RenderFragment ChildContent { get; set; }
+    public RenderFragment ChildContent { get; set; }
 
     [Parameter]
-    private EventCallback<UIMouseEventArgs> OnClick { get; set; }
+    public EventCallback<UIMouseEventArgs> OnClick { get; set; }
 }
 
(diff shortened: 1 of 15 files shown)


NTaylorMullen:
Also note: A lot of the sample projects I updated are still preview7. That being said there is little harm in updating them to also show `public` since having `non-public` parameters in the past wasn't a hard requirement.

guardrex:
@NTaylorMullen Could you resubmit this for the `release/3.0-preview8` branch?

NTaylorMullen:
> @NTaylorMullen Could you resubmit this for the `release/3.0-preview8` branch?
> 
> [EDIT] Unless this is ok to publish live now (Preview 7). I can't tell from the title versus your last comment if this PR would be ok to go live with now or not.

Eh probably not great to publish for preview7, i'll re-do it on preview8

guardrex:
Ok, thanks. Yeah ... we can review+merge now to that branch.

NTaylorMullen:
🆙 📅

guardrex:
@NTaylorMullen It hasn't been necessary to expose properties publicly ... but I see "expectation" and "an analyzer to suggest users have public parameters." We don't document design decisions generally, so this is just a personal curiosity: Why did it change when in many instances it isn't necessary to expose properties publicly (e.g., child component scenarios)?

NTaylorMullen:
> Why did it change when in many instances it isn't necessary to expose properties publicly (e.g., child component scenarios)?

We have to jump through hoops in order to properly get metadata or properly discover parameters for private properties and in some situations we just can't discover either (ref assemblies are built without private metadata included). Metadata is important because it's what drives the documentation and having members on a type (in the ref assembly case) drives what a component is capable of.

PR #13527, opened 2019-07-27
Update health-checks.md

dthompsonza:
This proposal attempts to simplify the sample code in the `ResponseWriter` delegate to be a little easier to read and remove the numerous new JProperty and JObject invocations. The `StringEnumConverter` is passed into the serializer to ensure the `HealthStatus` enum's are rendered by their names and not their values (Healthy or Unhealthy vs. 2 or 0).
Thank you.

### DIFF
--- a/aspnetcore/host-and-deploy/health-checks.md
+++ b/aspnetcore/host-and-deploy/health-checks.md
@@ -263,16 +263,15 @@ private static Task WriteResponse(HttpContext httpContext,
 {
     httpContext.Response.ContentType = "application/json";
 
-    var json = new JObject(
-        new JProperty("status", result.Status.ToString()),
-        new JProperty("results", new JObject(result.Entries.Select(pair =>
-            new JProperty(pair.Key, new JObject(
-                new JProperty("status", pair.Value.Status.ToString()),
-                new JProperty("description", pair.Value.Description),
-                new JProperty("data", new JObject(pair.Value.Data.Select(
-                    p => new JProperty(p.Key, p.Value))))))))));
-    return httpContext.Response.WriteAsync(
-        json.ToString(Formatting.Indented));
+    var jsonString = JsonConvert.SerializeObject(result, Formatting.Indented,
+        new JsonSerializerSettings
+        {
+            Converters = new List<JsonConverter>
+            {
+                new Newtonsoft.Json.Converters.StringEnumConverter(),
+            }
+        });
+    return httpContext.Response.WriteAsync(jsonString);
 }
 ```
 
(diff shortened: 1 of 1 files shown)


guardrex:
Hello @dthompsonza ... That bit was taken from an engineering sample, and I'm not sure if matching a particular schema is a specification or if they wanted to show the more manual approach just as an example.

The topic+sample follow the engineering example at ...

https://github.com/aspnet/AspNetCore/blob/release/2.2/src/Middleware/HealthChecks/test/testassets/HealthChecksSample/CustomWriterStartup.cs#L44-L56

Current topic sample produces ...

```json
{
  "status": "Healthy",
  "results": {
    "memory": {
      "status": "Healthy",
      "description": "Reports degraded status if allocated bytes >= 1073741824 bytes.",
      "data": {
        "AllocatedBytes": 1153168,
        "Gen0Collections": 0,
        "Gen1Collections": 0,
        "Gen2Collections": 0
      }
    }
  }
}
```

Suggested update on this PR produces ...

```json
{
  "Entries": {
    "memory": {
      "Data": {
        "AllocatedBytes": 1153168,
        "Gen0Collections": 0,
        "Gen1Collections": 0,
        "Gen2Collections": 0
      },
      "Description": "Reports degraded status if allocated bytes >= 1073741824 bytes.",
      "Duration": "00:00:00.0014867",
      "Exception": null,
      "Status": "Healthy"
    }
  },
  "Status": "Healthy",
  "TotalDuration": "00:00:00.0089104"
}
```

Stand-by ... I'll ask if we should go this way. 

If they like it, then I'll request that you also change the topic's sample app at ...

https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/host-and-deploy/health-checks/samples/2.x/HealthChecksSample/CustomWriterStartup.cs#L55-L65

... and add the `System.Collections.Generic` namespace on Line 1.

They may like this in the engineering sample as well on a PR over there.

rynowak:
The complexity of the sample is intentional. The point is that **you** the developer can choose what information you want to return from your health endpoints. We want to make it clear that *we're not providing a recommendation to just call JsonConvert on the object we give you*.

Sorry if this seems arbitrary, but I'd rather this stay as is.

dthompsonza:
> The complexity of the sample is intentional. The point is that **you** the developer can choose what information you want to return from your health endpoints. We want to make it clear that _we're not providing a recommendation to just call JsonConvert on the object we give you_.
> 
> Sorry if this seems arbitrary, but I'd rather this stay as is.

Hey thank you for taking the time to consider. I understand the reasoning and accept the result.

Thanks so much for the quick response.

Cheers

Issue #13612, opened 2019-08-02
Error if using the "copy" function for the code under heading "Add validation rules to the movie model"


WinterDragon76:
public int ID { get; set; }

should be 

public int Id { get; set; }

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 443a7c53-cc16-7ed5-5950-46a6f195f19c
* Version Independent ID: 12fe9329-2cef-4056-4986-18ed7ae5c397
* Content: [Add validation to an ASP.NET Core MVC app](https://docs.microsoft.com/en-us/aspnet/core/tutorials/first-mvc-app/validation?view=aspnetcore-2.2)
* Content Source: [aspnetcore/tutorials/first-mvc-app/validation.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/tutorials/first-mvc-app/validation.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-tutorials**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
We consistently use upper case `ID`

PR #13616, opened 2019-08-02
Fix: Wrong property naming on Movie

Der-Kraken:
There is an compiler error because the property "ID" on model "Movie" was written different than the references.


Rick-Anderson:
@Der-Kraken  sorry I didn't see this.  See #13622
and #13612

Rick-Anderson:
cc @LouisT123

PR #13676, opened 2019-08-05
changed localhost:xxxx to localhost:{PORT}

LouisT123:
improved readability


LouisT123:
@guardrex @Rick-Anderson 
Can you review this PR and merge it? 
/cc: @anangaur

guardrex:
@LouisT123 No need to ping us ... we see check several times a day. Pinging hits us with even more email load, which slows us down. Ping if the PR sits for a few days with no response ... that means that something really did fall off the radar. :smile:

LouisT123:
> @LouisT123 No need to ping us ... we see check several times a day. Pinging hits us with even more email load, which slows us down. Ping if the PR sits for a few days with no response ... that means that something really did fall off the radar. 😄

Noted, sorry about that.

PR #13677, opened 2019-08-05
Update handle-errors.md

Rick-Anderson:
Fantastic info but way too long to MT (Machine translate).

Avoid safe unless you can define/tell them how to be safe.

After @guardrex  review, we'll bring in the rest of the gang.

Once approved, S&M into #13407

### DIFF
--- a/aspnetcore/blazor/handle-errors.md
+++ b/aspnetcore/blazor/handle-errors.md
@@ -119,7 +119,7 @@ Client-side code can trigger invocations of C# code. Event handler code might th
 
 If an event handler throws an unhandled exception, the exception is fatal to the circuit. If the app calls code that could fail for external reasons (for example, a database query fails):
 
-* The app should Trap the exceptions using [try-catch](/dotnet/csharp/language-reference/keywords/try-catch).
+* The app should trap the exceptions using [try-catch](/dotnet/csharp/language-reference/keywords/try-catch).
 * Handle and log the error.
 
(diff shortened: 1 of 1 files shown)


guardrex:
I'll look first thing on Tuesday morning. The content was already approved by engineering, so I think we can merge after this.

btw - How are you deciding which sentences to break? Has the localization team provided guidance on sentence length or stated that the use of all (or most) non-essential clauses should be avoided?

The problem with breaking sentences to this degree is that it slows down reading. It literally takes longer to read the whole topic. It's plainly a less efficient approach to transferring information.

Rick-Anderson:
25 words max.

> The problem with breaking sentences to this degree is that it slows down reading

True, but it increases the speed of comprehension for most folks, and certainly for folks not familiar with the topic.  Our internal tool flagged all of these and  more.

Rick-Anderson:
Yes we can merge once you verify I didn't change the meaning.

guardrex:
I see. Thanks ..... that's the figure that I didn't remember. I think you did state that figure one day in *ancient .NET Core times* 👴 lol. I just now remembered an old convo about it, but that was *years* ago IIRC.

> Our internal tool flagged all of these and more.

That's not surprising. I've seen about *a bazillion* sentences that wouldn't qualify anywhere near 25 words.

guardrex:
I'm on a 😵 *brain fry* 😵 now, but I'll hit this first thing tomorrow morning.

PR #13692, opened 2019-08-06
xxxx to {PORT}

LouisT123:
Should be the last one for port url syntax change
for [Examine the Details and Delete methods of an ASP.NET Core app](https://docs.microsoft.com/en-us/aspnet/core/tutorials/first-mvc-app/details?view=aspnetcore-2.2)


guardrex:
> should be the last one

Oh, ok ... sounds good. I thought that there might be more. Thanks for fixing these up.

PR #13724, opened 2019-08-08
Blazor localization

### LINKED ISSUE #13436: [Preview 8] Blazor localization
# Summary

To localize a server-side Blazor application use the localization middleware (https://docs.microsoft.com/en-us/aspnet/core/fundamentals/localization?view=aspnetcore-2.2#localization-middleware) to select the appropriate culture for each user. You can you use any scheme that you want to select the culture, see the localization middleware documentation for examples.

## Cookies

As as special consideration for server-size Blazor we recommend setting the culture in a cookie as part of your host page. This will ensure that the websocket connection can propagate the culture correctly. Localization schemes based on the URL path or query string may not be able to work with websockets, so using the cookie as a fallback is useful.

Here's an example of how to set the current culture in a cookie that can be read by the localization middleware.

```C#
    public class HostModel : PageModel
    {
        public void OnGet()
        {
            HttpContext.Response.Cookies.Append(
                    CookieRequestCultureProvider.DefaultCookieName,
                    CookieRequestCultureProvider.MakeCookieValue(
                        new RequestCulture(
                            CultureInfo.CurrentCulture,
                            CultureInfo.CurrentUICulture)));
        }
    }
```

Here's an explanation of how this works:
1. The browser sends an initial HTTP request to the application
2. The culture is assigned by the localization middleware
3. The `_Host.cshtml` page persists the culture in a cookie as part of the response
4. The browser opens a web socket to being an interactive server-side Blazor session
5. The localization middleware reads the cookie and assign the culture
6. The server-side Blazor session begins with the correct culture

## Providing UI to choose the culture

To provide UI to allow users of your application to select the culture, we recommend using a redirect-based approach. This means that a new server-side Blazor session will start, and you should persist any state you want to retain, similar to what happens when a user logs in.

In order to make this work, you will need an HTTP endpoint on the server to change the users stored locale, and perform the redirect. Here's an example you can customize:

```C#
    [Route("[controller]/[action]")]
    public class CultureController : Controller
    {
        public IActionResult SetCulture(string culture, string redirectUri)
        {
            if (culture != null)
            {
                HttpContext.Response.Cookies.Append(
                    CookieRequestCultureProvider.DefaultCookieName,
                    CookieRequestCultureProvider.MakeCookieValue(new RequestCulture(culture)));
            }

            return LocalRedirect(redirectUri);
        }
    }
```

This controller will accept the culture and a redirect URI in the query string, and will update the users' stored culture cookie before redirecting them. Note: It's very important to use the `LocalRedirect` action result here for security reasons. 

Here's a simplified sample of how to perform this redirection flow from a component:

```razor
@inject IUriHelper UriHelper

<h3>Select your language</h3>
<select @onchange="OnSelected" id="culture-selector">
    <option>Select...</option>
    <option value="en-US">English</option>
    <option value="fr-FR">Français</option>
</select>
@code {
    double textNumber;

    void OnSelected(UIChangeEventArgs e)
    {
        var culture = (string)e.Value;
        var uri = new Uri(UriHelper.GetAbsoluteUri()).GetComponents(UriComponents.PathAndQuery, UriFormat.Unescaped);
        var query = $"?culture={Uri.EscapeDataString(culture)}&redirectUri={Uri.EscapeDataString(uri)}";
        UriHelper.NavigateTo("/Culture/SetCulture" + query, forceLoad: true);
    }
}
```

## Using localization features from Blazor

Inside Blazor applications you can use existing localization and globalization features such as .NET's resources system, or culture-specific formatting of numbers and dates. Blazor's `@bind` functionality performs globalization based on the current culture.

A limited set of ASP.NET Core's localization features are supported in Blazor at the time of writing. `IStringLocalizer<>` can be used with Blazor. `IHtmlLocalizer<>`, `IViewLocalizer<>` and Data Annotations localization are ASP.NET Core MVC features and cannot be used with Blazor. https://docs.microsoft.com/en-us/aspnet/core/fundamentals/localization?view=aspnetcore-2.2#make-the-apps-content-localizable

guardrex:
Fixes #13436

[Internal Review Topic (Localization section direct link)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/components?view=aspnetcore-3.0&branch=pr-en-us-13724#localization)

* **For the `release/3.0-preview8` staging branch.**
* Watch me carefully here ... I might have :boom: *broken something* :boom:.
* Placed at the bottom, but it is cross-linked both ways to the `@bind` section, which is where the globalization bits live.
* You suggested possibly bringing some bits over from ASP.NET Core docs on the subject. Did you have anything specific in mind? I cross-link as you suggest ... will that cover it?

### DIFF
--- a/aspnetcore/blazor/components.md
+++ b/aspnetcore/blazor/components.md
@@ -247,6 +247,8 @@ The following field types have specific formatting requirements and aren't curre
 
 `@bind` supports the `@bind:culture` parameter to provide a <xref:System.Globalization.CultureInfo?displayProperty=fullName> for parsing and formatting a value. Specifying a culture isn't recommended when using the `date` and `number` field types. `date` and `number` have built-in Blazor support that provides the required culture.
 
+For information on how to set the user's culture, see the [Localization](#localization) section.
+
 **Format strings**
 
 Data binding works with <xref:System.DateTime> format strings using [@bind:format](xref:mvc/views/razor#bind). Other format expressions, such as currency or number formats, aren't available at this time.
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/components.md
```diff
@@ -1223,3 +1225,121 @@ This is a trivial example. In more realistic cases with complex and deeply neste
...
+For more information and examples, see <xref:fundamentals/localization>.
+
+### Cookies
+
+A localization culture cookie can persist the user's culture. The cookie is created by the `OnGet` method of the app's host page (*Pages/Host.cshtml.cs*). The Localization Middleware reads the cookie on subsequent requests to set the user's culture. 
+
+Use of a cookie ensures that the WebSocket connection can correctly propagate the culture. If localization schemes are based on the URL path or query string, the scheme might not be able to work with WebSockets, thus fail to persist the culture. Therefore, use of a localization culture cookie is the recommended approach.
+
+The following example shows how to set the current culture in a cookie that can be read by the Localization Middleware. Create a *Pages/Host.cshtml.cs* file with the following contents in the Blazor server-side app:
+
```
rynowak: One detail that I'd like to see called out stronger here:

You can use any technique you like to assign a culture, as long as you persist it in a cookie as well. So if you have a localization scheme that you already like for server-side ASP.NET Core, you can still do that! Just also set the cookie and you're good to go.

aspnetcore/blazor/components.md
```diff
@@ -1223,3 +1225,121 @@ This is a trivial example. In more realistic cases with complex and deeply neste
...
+                    new RequestCulture(culture)));
+        }
+
+        return LocalRedirect(redirectUri);
+    }
+}
+```
+
+> [!WARNING]
+> Use the `LocalRedirect` action result for security reasons.
```
rynowak: Do we have docs about open redirects and why they are bad anywhere? https://cwe.mitre.org/data/definitions/601.html

I don't want to bloat this doc with open redirect info, but I'm wondering if we have something good to link to.

guardrex:
Thanks ... *yes* we do have open redirect attack coverage. I linked it on the last commit in that warning.

PR #13756, opened 2019-08-10
Update asp-all-route-data

deisterhold:
When using `asp-all-route-data`, if it is specified after other ` asp-route-myvar` attributes they will not be used to generate the route.

```
@{
    var queryString = new Dictionary<string, string>()
    {
        {"returnUrl", @Context.Request.GetEncodedUrl()}
    };
}
```
The following works:
`asp-all-route-data="@queryString" asp-action="Edit" asp-route-id="@Model.Id"`

The following does NOT work:
`asp-action="Edit" asp-route-id="@Model.Id" asp-all-route-data="@queryString"`


Rick-Anderson:
Needs edit after approval.

NTaylorMullen:
I'm not convinced this is a mainline enough scenario to warrant adding it to the docs.

Rick-Anderson:
@NTaylorMullen  thanks. It's very discoverable  on this page from the issue. We  work on the title.

PR #13783, opened 2019-08-12
Removed images that are not referenced anywhere.

IEvangelist:
Hi @scottaddie,

I ran the [DocFx.Repository.Sweeper](https://github.com/IEvangelist/DocFX.Repository.Sweeper) over this repo. It found and deleted, 363 images -- 15.2 MB.


guardrex:
Thanks @IEvangelist ... **_but_** ...

@scottaddie Recommend that we **don't** take this .... for example, the SVG font file in the Blazor sample provides icons and isn't markup-linked into the app.

IEvangelist:
Hi @guardrex,

No worries... I'm curious though, how could a tool programmatically determine if those files should be excluded from deletion? Are they part of a sample directory or what?

guardrex:
Usually, topics consume image assets in the *_static* folder at the same path as the topic, so one approach is to look at this from the other side: Any images in the *_static* folder that aren't in a topic's markup links should be safe to remove.

IEvangelist:
> Usually, topics consume image assets in the __static_ folder at the same path as the topic, so one approach is to look at this from the other side: Any images in the __static_ folder that aren't in a topic's markup links should be safe to remove.

That is actually how the tool works. It creates file tokens and parses all markdown files, discovering image references and even other markdown file references along the way. As part of this discovery it finds the corresponding files on disc and builds out a token-reference lookup mechanism. This is how it knows which files are not being referenced anywhere else.

In other words, an instance of a file token (md file) contains images that it refers to and relies on, and those references can exist in the __static_ folder (or anywhere else within the doc set). The same is true for YAML/YML files as well, they are tokenized in the same manner - where both markdown and image file references exist.

> Any images in the __static_ folder that aren't in a topic's markup links should be safe to remove.

That is what the tool discovered here, unless I'm missing something.

guardrex:
https://github.com/aspnet/AspNetCore.Docs/tree/master/aspnetcore/blazor/common/samples/3.x/BlazorSample/wwwroot/css/open-iconic/font/fonts/open-iconic.svg

... isn't in the *_static* folder. It's in the *samples* folder ... sometimes it's named *sample* ... *usually*  :smile: ... I wouldn't be surprised if naming isn't 100% consistent.

Rick-Anderson:
We need to be careful about this. If you look on the file tab, most of the deleted files are from the *wwwroot* folder. Those files aren't used in the doc generation, but they **are** used in the sample app.

Rick-Anderson:
cc @tdykstra

Rick-Anderson:
@IEvangelist  thanks for the effort. I may look at this commit and figure out files that can safely be deleted.

IEvangelist:
> @IEvangelist thanks for the effort. I may look at this commit and figure out files that can safely be deleted.

I'm baking in the new rules into the tool and will have another PR soon.

Rick-Anderson:
Now if you could remove those binaries from our history I'd put you in for a major award.

PR #13797, opened 2019-08-13
"Blazor App" → Blazor Server App

leotsarev:
In latestest preview, it called "Blazor Server App"


guardrex:
Thanks @leotsarev ... but we're handling the change via a staging branch.

guardrex:
Also note @leotsarev that the top-level name will be **Blazor App** and then the other templates will appear below that for server-side, client-side (not hosted), client-side (hosted by ASP.NET Core). We'll have an update for the docs going in very shortly after Preview 8 releases.

PR #13802, opened 2019-08-13
Updated prereq file for VS for Mac

LouisT123:
New instructions in vs for mac will need version 8.0


Rick-Anderson:
@LouisT123  that's not correct. You've changed the include in the  `::: moniker range=">= aspnetcore-3.0"` portion 

If the include file has the wrong version, update the include file.

PR #13803, opened 2019-08-13
Removed images that are not referenced anywhere.

IEvangelist:
Applied whitelisting rules, ignored "wwwroot", "sample" and "snippet" directories.

image


IEvangelist:
New PR from new sweep, here's the old PR : https://github.com/aspnet/AspNetCore.Docs/pull/13803

PR #13823, opened 2019-08-14
add cookie path

noctwolf:
It is recommended to add this line, compatible with applications deployed in IIS

### DIFF
--- a/aspnetcore/security/cookie-sharing.md
+++ b/aspnetcore/security/cookie-sharing.md
@@ -45,6 +45,7 @@ services.AddDataProtection()
 
 services.ConfigureApplicationCookie(options => {
     options.Cookie.Name = ".AspNet.SharedCookie";
+    options.Cookie.Path = "/";
 });
 ```
 


guardrex:
@Tratcher Isn't this the default if it isn't set?

https://github.com/aspnet/AspNetCore/blob/master/src/Http/Http.Abstractions/src/CookieBuilder.cs#L103

Tratcher:
Yes. Why are you setting it to the default value?

noctwolf:
The instantiated here is the RequestPathBaseCookieBuilder class.
https://github.com/aspnet/AspNetCore/blob/master/src/Security/Authentication/Cookies/src/CookieAuthenticationOptions.cs#L15
Changed Path in the Build method of the RequestPathBaseCookieBuilder class
https://github.com/aspnet/AspNetCore/blob/master/src/Security/Authentication/Core/src/RequestPathBaseCookieBuilder.cs#L21-L35

noctwolf:
@Rick-Anderson @guardrex @Tratcher 
For the first time people using SSO, this is a trap

Tratcher:
Ah, I forgot about those custom builders. Yes, this is a scenario that we should discuss and document but changing it in just this code sample isn't going to help. Please file a doc bug so we can make sure the SSO scenario is covered.

noctwolf:
The file modified in this PR is already cookie-sharing.md
What should I do?

guardrex:
Cookie options are first discussed here ...

https://docs.microsoft.com/aspnet/core/security/authentication/identity-configuration#cookie-settings

@noctwolf File an issue using the **This Page** feedback form at the bottom of that topic.

PR #13827, opened 2019-08-14
Language change in cookie sharing topic

guardrex:


guardrex:
I made it nice and short. The lead-in to this already says 'for the following examples.'

Issue #13848, opened 2019-08-15
SameSiteMode  requires using Microsoft.AspNetCore.Http;


mochr:
A using directive (using Microsoft.AspNetCore.Http;) has to be added to enable the SameSiteMode-enum. This should be added as a separate bullet point after the paragraph: "To enable the default cookie consent feature like that found in the ASP.NET Core 2.2 templates in an ASP.NET Core 3.0 template generated app:".

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 28d0490e-205f-20a5-ace5-904f4c8ad78f
* Version Independent ID: 4a2c8d95-fdd2-fdb1-b61d-802d0a8539c3
* Content: [General Data Protection Regulation (GDPR) support in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/gdpr?view=aspnetcore-3.0#feedback)
* Content Source: [aspnetcore/security/gdpr.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/security/gdpr.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-security**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
When updating this, move #8544 to **Additional resources**

Rick-Anderson:
We'll fix this in the next doc update. We typically don't add all the includes in advanced docs.

Issue #8544, opened 2018-09-14
Blog on revoking cookie consent after consent has been accepted


joeaudette:
http://www.joeaudette.com/blog/2018/08/28/gdpr---adding-a-revoke-consent-button-in-aspnet-core

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 28d0490e-205f-20a5-ace5-904f4c8ad78f
* Version Independent ID: 4a2c8d95-fdd2-fdb1-b61d-802d0a8539c3
* Content: [General Data Protection Regulation (GDPR) support in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/gdpr?view=aspnetcore-2.1)
* Content Source: [aspnetcore/security/gdpr.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/security/gdpr.md)
* Product: **aspnet-core**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

mochr:
Will be closed by #13849.

wadepickett:
@Rick-Anderson, should this one be closed?  Not clear to me if #13849 that was merged closes this issue or not.

Elvis254:
Hello @Rick-Anderson,
The link: [http://www.joeaudette.com/blog/2018/08/28/gdpr---adding-a-revoke-consent-button-in-aspnet-core](http://www.joeaudette.com/blog/2018/08/28/gdpr---adding-a-revoke-consent-button-in-aspnet-core) isn't available.
Elvis.

PR #13849, opened 2019-08-15
Adding a hint to include a missing using directive

mochr:
Closes issue #13848


mochr:
Added a commit to close issue #8544 in the same document.

Rick-Anderson:
@mochr  much appreciated!
I removed code snippet with include, I don't think that's necessary.

PR #13860, opened 2019-08-16
Automatic backing field support update for Preview 9

### LINKED ISSUE #13854: [Preview 9] ElementReference and backing fields
* Remove `@ref:suppressField` in examples.
* Update text for [Removing automatic backing field support for @ref in 3.0.0-preview9](https://github.com/aspnet/Announcements/issues/381).

guardrex:
Fixes #13854

* Update current text+examples for **Preview 9**. This PR goes to the `release/3.0-preview9` branch.
* Drop some text that I held on to (it was commented out) from the release and patch at Preview 8.
* This seeks to follow the guidelines in the [Removing automatic backing field support for @ref in 3.0.0-preview9](https://github.com/aspnet/Announcements/issues/381) announcement.

cc: @rynowak @danroth27

### DIFF
--- a/aspnetcore/blazor/components.md
+++ b/aspnetcore/blazor/components.md
@@ -517,10 +517,9 @@ Component references provide a way to reference a component instance so that you
 
 * Add an [@ref](xref:mvc/views/razor#ref) attribute to the child component.
 * Define a field with the same type as the child component.
-* Provide the `@ref:suppressField` parameter, which suppresses backing field generation. For more information, see [Removing automatic backing field support for @ref in 3.0.0-preview9](https://github.com/aspnet/Announcements/issues/381).
 
 ```cshtml
-<MyLoginDialog @ref="loginDialog" @ref:suppressField ... />
+<MyLoginDialog @ref="loginDialog" ... />
 
 @code {
     private MyLoginDialog loginDialog;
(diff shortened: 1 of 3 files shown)


aspnetcore/blazor/components.md
```diff
@@ -537,30 +536,6 @@ When the component is rendered, the `loginDialog` field is populated with the `M
...
-    }
-}
-```
-
-When the component is rendered, the generated `myLoginDialog` field is populated with the `LoginDialog` component instance. You can then invoke .NET methods on the component instance.
-
-In some cases, a backing field is required. For example, declare a backing field when referencing generic components. To suppress backing field generation, specify the `@ref:suppressField` parameter.
-
-> [!IMPORTANT]
-> The generated `myLoginDialog` variable is only populated after the component is rendered and its output includes the `LoginDialog` element. Until that point, there's nothing to reference. To manipulate components references after the component has finished rendering, use the `OnAfterRenderAsync` or `OnAfterRender` methods.
```
rynowak: It looks like there's some stuff here that's getting removed that isn't related to `suppressField`. Is that intentional?

guardrex: Yes ... this text was held just in case I needed it. It was convenient to just comment it out for a few days. It can go now. 🔪

PR #13880, opened 2019-08-16
Update startup.md

Rick-Anderson:
@guardrex  see [this comment](https://github.com/aspnet/AspNetCore.Docs/issues/8407#issuecomment-418798089)

Startup filters are definitely an augmentation to Startup rather than an integral piece of it. They pair well with IHostingStartup as a means of injection into Startup, but that's not the only use case.

Is that worth adding or should we close this PR?

### DIFF
--- a/aspnetcore/fundamentals/startup.md
+++ b/aspnetcore/fundamentals/startup.md
@@ -190,6 +190,8 @@ To configure services and the request processing pipeline without using a `Start
 
 ## Extend Startup with startup filters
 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
The only problem is this begs, what other use cases? I don't really want to bother `@Tratcher` for that.

guardrex:
The first sentence (out of context of the readers question on that issue) doesn't seem to convey any useful information here (in that form anyway). The second sentence (the example) seems like it would work better at the end of the existing first paragraph (Line 195 on this PR; LIne 193 on the live topic).

PR #13882, opened 2019-08-17
Update controller name

colmprunty:
I ran a new create command and there is no `SampleDataController`, the `Authorize` header is in `WeatherForecastController`.


scottaddie:
Thank you for fixing this, @colmprunty! Your change will go live within the next 24 hours.

PR #13883, opened 2019-08-17
Fixes inconsistency in the attribute splatting example

pekspro:
This fixes a minor inconsistency in the attribute splatting example.

### DIFF
--- a/aspnetcore/blazor/components.md
+++ b/aspnetcore/blazor/components.md
@@ -148,7 +148,7 @@ In the following example, the first `<input>` element (`id="useIndividualParams"
     public string Placeholder { get; set; } = "Input placeholder text";
 
     [Parameter]
-    public string Required { get; set; } = "true";
+    public string Required { get; set; } = "required";
 
     [Parameter]
     public string Size { get; set; } = "50";
@@ -159,7 +159,7 @@ In the following example, the first `<input>` element (`id="useIndividualParams"
         {
             { "maxlength", "10" },
             { "placeholder", "Input placeholder text" },
-            { "required", "true" },
+            { "required", "required" },
             { "size", "50" }
         };
 }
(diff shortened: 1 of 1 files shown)


guardrex:
Hi @pekspro ... How do feel about doing it this way? ...

```
<input id="useIndividualParams"
       maxlength="@Maxlength"
       placeholder="@Placeholder"
       @Required
       size="@Size" />
```

Side note regarding XML and closing: They either *were* or *are* working on the self-closing void element thing. I'd need to research that further to see where they are with it. However, I think we can go with the hybrid here if you agree on floating the `required` attribute in the element.

guardrex:
mmmm 🤔 ... `InputAttributes` dictionary ... I think we can do ...

```csharp
new Dictionary<string, object>()
        {
            { "maxlength", "10" },
            { "placeholder", "Input placeholder text" },
            { "required", "required" },
            { "size", "50" }
        };
```

... in this scenario and get that behavior.

pekspro:
@guardrex, I think you are right, I’ve just updated the PR. I’ve just noticed when I read the document that the two samples generated different output and tried to fix that. Obviously I picked the wrong one… :-)

(BTW, I’m impressed by the insanely fast feedback on this PR. I hope you have more fun things to do on a weekend :-)

guardrex:
Right ... this looks good. The spec is https://www.w3.org/TR/html50/forms.html#the-required-attribute, and boolean attribute is at https://www.w3.org/TR/html50/infrastructure.html#boolean-attribute ...

> The values "true" and "false" are not allowed on boolean attributes. To represent a false value, the attribute has to be omitted altogether.

Thanks for fixing this! :rocket:

> I hope you have more fun things to do on a weekend

There's nothing **_more fun_** to do on a weekend than ASP.NET Core! :smile:

Issue #10481, opened 2019-01-18
Update Web API Mongo for 3.0


tdykstra:
Update the tutorial for ASP.NET Core 3.0.
---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 5240866d-d0a1-f0dd-7b2c-203e1d69bc4b
* Version Independent ID: 83a5cb3b-23bc-4f0c-cd1c-1d1ffeb1131a
* Content: [Build web APIs with ASP.NET Core and MongoDB](https://docs.microsoft.com/en-us/aspnet/core/tutorials/first-mongo-app?view=aspnetcore-2.1&tabs=visual-studio#next-steps)
* Content Source: [aspnetcore/tutorials/first-mongo-app.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/tutorials/first-mongo-app.md)
* Product: **aspnet-core**
* GitHub Login: @prkhandelwal
* Microsoft Alias: **scaddie**

xtianus79:
What are the primary changes going to be for this update?  Is the driver affected in a meaningful way?  The reason I ask is because the mongo university team has since removed the windows m201W course and is from what I understand working on a new one.

ricnef2121:
An unhandled exception occurred while processing the request.
FormatException: Cannot deserialize a 'String' from BsonType 'Double'.
MongoDB.Bson.Serialization.Serializers.StringSerializer.DeserializeValue(BsonDeserializationContext context, BsonDeserializationArgs args)
FormatException: An error occurred while deserializing the Price property of class BooksApi.Models.Book: Cannot deserialize a 'String' from BsonType 'Double'.
MongoDB.Bson.Serialization.BsonClassMapSerializer<TClass>.DeserializeMemberValue(BsonDeserializationContext context, BsonMemberMap memberMap)
Stack Query Cookies Headers 
FormatException: Cannot deserialize a 'String' from BsonType 'Double'.
MongoDB.Bson.Serialization.Serializers.StringSerializer.DeserializeValue(BsonDeserializationContext context, BsonDeserializationArgs args)
MongoDB.Bson.Serialization.Serializers.SealedClassSerializerBase<TValue>.Deserialize(BsonDeserializationContext context, BsonDeserializationArgs args)
[...]

PR #13892, opened 2019-08-18
First mongo app 3.0

spboyer:
* Update & add .NET Core 3.0 version of  [tutorials/first-mongo-app](https://docs.microsoft.com/en-us/aspnet/core/tutorials/first-mongo-app?view=aspnetcore-2.2)
* Add `/sample/BooksApi-3.0` source code folder inclusive if README file for https://docs.microsoft.com/samples/browse/


guardrex:
@scottaddie If you want this to follow our sample conventions but are too busy to update it, I can take care of it.

spboyer:
@guardrex can you send me details on conventions?

guardrex:
* We've been naming the folders `2.x` and `3.x` with the sample folder within those. The sample folder itself is named *samples* and the snapshots folder is named *samples_snapshot*.

  We're trying to use "*SampleApp*" as the name of all (new anyway) sample apps. Since you have a pre-existing sample here, you could just name the sample app's folder *SampleApp* and leave the sample's project name alone (i.e., leave the *.csproj* file name and namespaces within the app as they are).

  I have concerns about the approach because:

  * Often, a dev performs a cut-'n-paste that doesn't grab the namespace.
  * Even if they do, the namespace is easy enough for the dev to change.
  * If the samples have a dedicated project names (different folder names, different namespaces), it's easier to identify and flip in and out of them when you have a few sample projects open in multiple IDE windows. When they're all called "SampleApp," I find it confusing as to what I have open where ... what exactly it is that I'm looking at.

  Anyway, the guidance is to go with *SampleApp* for both the folder name and the project name (+namespaces), especially for new topics with new sample apps.

  ```
  first-mongo-app/samples/2.x/SampleApp/...
  first-mongo-app/samples_snapshot/2.x/SampleApp/...
  first-mongo-app/samples/3.x/SampleApp/...
  first-mongo-app/samples_snapshot/3.x/SampleApp/...
  ```

* At each moniker block instance, place the >=3.0 entry (most recent) first and the <3.0 entry second.
* Use the versioning blocks **_for every code link_**, even if the code doesn't change across the samples. Therefore, don't use a `2.x` sample file link within a `3.x` moniker block. This makes it a little easier to drop the 2.x sample and mod the 3.x sample independently later. 

  As an alternative on this point: You can version the **_whole topic_** in two giant moniker blocks, one for >=3.0 and one for <3.0. This only works if there are no pre-existing moniker blocks, as nesting them isn't permitted and will :boom: if attempted.

  We've been taking this alternate approach more frequently lately. Here's a recent example: https://github.com/aspnet/AspNetCore.Docs/pull/13785/files

  There are some concerns here, too:

  * Having to split giant blocks to place third version updates into the topic would be time-consuming ... or the topic ends up with **_three_** giant blocks.
  * This approach duplicates all of the text, so many changes will need to be made in two spots (*or more!* ... possibly **_three_** spots later).
  * Readers submitting a PR may only mod one block and not the other block, which will require the docs team member to mod the other one.

  Anyway ... I'm sure if it doesn't work out that this approach will be dropped/modified later. This approach does shine in one way tho: It comports well with the upcoming version-by-file scenario for the docs. It will make it easy to set that up when the time comes for those topics that must cover a lot of engineering churn.

spboyer:
The folder structure makes sense, I can make those changes in this same PR. Would that satisfy?

guardrex:
Sure ... or I can do it if you're 🏃😅. I've done so many of these at this point that I can just about do them in my sleep. :smile: If you want me to take care of it, I recommend going ahead with Scott on the review and S&M. I'll do a quick PR immediately thereafter.

spboyer:
@gaurdrex - have at it, as long as you do it in this branch. I’d like to pull it later to see the changes.

Thanks!

guardrex:
That *readme.md* in the sample app is unusual ... looks like an old topic version or something.

spboyer:
It is needed to add the to the samples browser. https://docs.microsoft.com/samples/browse/

The metadata provides the needed info for the webhook CI process.

guardrex:
I see. I'm not familiar with those bits (yet anyway).

spboyer:
@scottaddie to finalize

spboyer:
Closes #10481

PR #13898, opened 2019-08-18
Blazor client-side host and deploy updates

### LINKED ISSUE #13715: Publish a Blazor app using standalone method in a virtual path (problems)
I followed the official documentation (https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/blazor/client-side?view=aspnetcore-3.0), but the steps for publishing a blazor client side app are unclear.

I would like to publish to IIS.

The modification in the base tag (`<base href="/">`) has a bug.

The suggested modifications to webconfig (parent) and webconfig (app folder) do not work, and in my case the tags to change in parent web.config didn't even exist.

First try, I created a virtual folder, turned the folder into an application, wanted to access `http://site/app`, ran the `dotnet publish -c release` command, and copied the dist folder (`\bin\release\netstandard2.0\publish\app\dist`) from the release to the application's virtual folder. Unsuccessfully. Routes do not work and the application always routes urls with reference to the root and not the virtual folder.

Second try, I modified the base <base href> tag by pointing to the virtual folder (`<base href="/App/"`). First problem, a lot of errors, after a lot of debugging I realized that the tag like this (`<base href="/app/">`), without the capitalization, worked, I imagine it's a bug. The paths to the pages continued without referencing the application folder .

Third try, I copied the whole folder of the publication (`\bin\release\netstandard2.0\publish\`), as the second try, also ran the application, with the same capitalization problem (corrected before), and same problem of paths not including the virtual folder (not corrected). 

What is the correct approach? Copy entire publish folder? Or just the dist folder? Or just static content and webconfig? The documentation doesn't make that clear either, just copying the dist folder causes a lot of headaches with the application paths. The most straightforward approach was to copy all the contents of the publish folder, and the most correct, static content (assembly folder) and webconfig with the rewrites of the routes.

Still, the application loads, but the page paths do not include the virtual folder.

Publishing directly to the root works perfectly, but using a virtual folder is quite problematic, and even worse in IIS, the solution for now has been to directly reference the virtual folder in the application page paths if you want to use a virtual directory.

More detailed documentation of ways of publishing client-side apps would be of great help.

guardrex:
Fixes  #13715 

* [Internal Review Topic (Index)](https://review.docs.microsoft.com/en-us/aspnet/core/host-and-deploy/blazor/index?view=aspnetcore-3.0&branch=pr-en-us-13898)
* [Internal Review Topic (Client-side)](https://review.docs.microsoft.com/en-us/aspnet/core/host-and-deploy/blazor/client-side?view=aspnetcore-3.0&branch=pr-en-us-13898)
* [Internal Review Topic (Server-side)](https://review.docs.microsoft.com/en-us/aspnet/core/host-and-deploy/blazor/server-side?view=aspnetcore-3.0&branch=pr-en-us-13898)

Two sets of updates here for the Blazor client-side host and deploy topic:

1. **Layout Updates** &ndash; In the client-side topic, I'm moving the less important content (host config values and *configuration of the linker*) **_down_** to the bottom of the topic. This way, the topic starts and focuses on the most important concepts. I'm also improving section cross-links and placing IIS sub-app coverage under IIS. The *App Base Path* section can come out to the *index* (landing page) of this node because it applies to both hosting models.

1. **Dropping "Virtual"** &ndash; The word "virtual" is overloaded between speaking to **paths** and speaking to **IIS** (i.e., **IIS vdirs**). I'm taking a shot at dropping the word "virtual" in favor of "relative URL," which *I think* serves the same purpose and avoids confusion over what "virtual" means, especially in the context of IIS.

NOTE: If there's a broken bookmark in the Cookie Sharing topic on the build here, it was fixed on https://github.com/aspnet/AspNetCore.Docs/pull/14160.


aspnetcore/host-and-deploy/blazor/server-side.md
```diff
@@ -5,7 +5,7 @@ description: Learn how to host and deploy a Blazor server-side app using ASP.NET
 monikerRange: '>= aspnetcore-3.0'
 ms.author: riande
 ms.custom: mvc
-ms.date: 06/11/2019
+ms.date: 09/05/2019
 uid: host-and-deploy/blazor/server-side
 ---
 # Host and deploy Blazor server-side
```
danroth27: The Blazor server-side topic seems surprisingly sparse. I think we need to flesh out this topic. Doesn't block this PR though.

aspnetcore/host-and-deploy/blazor/index.md
```diff
@@ -38,6 +38,45 @@ A Blazor client-side app is published to the */bin/Release/{TARGET FRAMEWORK}/pu
...
+
+```console
+dotnet run --pathbase=/CoolApp
+```
+
+The app responds locally at `http://localhost:port/CoolApp`.
+
 ## Deployment
 
 For deployment guidance, see the following topics:
```
danroth27: The "Blazor serverless hosting with Azure Storage" section below seems like it belongs in the Blazor client-side host and deploy topic as it is specific to Blazor WebAssembly apps.

guardrex:
@scottaddie This mostly just moves content. There's one section that I'd like to hear your opinion on ... scroll down to the ` aspnetcore/host-and-deploy/blazor/index.md` file. See that section on the *App base path* ... I refactored that quite a bit trying to clarify it. What do u think on that section?

The rest of this you can probably skip unless you want to review all of it. **_ALL_** of the Blazor docs are about to be reviewed by engineering ... before going GA ... so we'll catch my goofs across all of the docs later. Therefore, don't sweat the rest of this if ur busy.

aspnetcore/host-and-deploy/blazor/client-side.md
```diff
@@ -231,7 +137,7 @@ For more information on troubleshooting deployments to IIS, see <xref:test/troub
 
 ### Azure Storage
 
-Azure Storage static file hosting allows serverless Blazor app hosting. Custom domain names, the Azure Content Delivery Network (CDN), and HTTPS are supported.
+[Azure Storage](https://azure.microsoft.com/services/storage/) static file hosting allows serverless Blazor app hosting. Custom domain names, the Azure Content Delivery Network (CDN), and HTTPS are supported.
```
scottaddie: Let's keep the reader in docs.

```suggestion
[Azure Storage](/azure/storage/) static file hosting allows serverless Blazor app hosting. Custom domain names, the Azure Content Delivery Network (CDN), and HTTPS are supported.
```

PR #13911, opened 2019-08-19
Use static web assets scenario in all environments

### LINKED ISSUE #13864: Document how to use static web assets feature in all environments
see: https://github.com/aspnet/AspNetCore/issues/13190

We should make sure the docs include the difference between how this feature behaves in production vs development, and how to override this.

guardrex:
Fixes #13864

[Internal Review Topic (Create an RCL with static assets section)](https://review.docs.microsoft.com/en-us/aspnet/core/razor-pages/ui-class?view=aspnetcore-2.1&branch=pr-en-us-13911#create-an-rcl-with-static-assets)

This content is versioned >=3.0, so it's ok to shoot this to `master`.

Ignore the italics on the diff&mdash;it's a false positive here. A decision was made by management not to sweat CommonMark problems in GH diffs, so we don't escape those characters any longer.

### DIFF
--- a/aspnetcore/razor-pages/ui-class.md
+++ b/aspnetcore/razor-pages/ui-class.md
@@ -4,7 +4,7 @@ author: Rick-Anderson
 description: Explains how to create reusable Razor UI using partial views in a class library in ASP.NET Core.
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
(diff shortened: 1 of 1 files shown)


guardrex:
@javiercn I only called for Ryan because he opened the original issue. Would you like to review instead?

PR #13956, opened 2019-08-22
Removed broken link

ohelmer:


scottaddie:
Thank you for the fix, @ohelmer! This change will go live within the next 24 hours.

PR #13971, opened 2019-08-22
Routing to file names

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/12122

[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/routing?view=aspnetcore-3.0&branch=pr-en-us-13971#routing-with-urls-that-contain-dots)

### DIFF
--- a/aspnetcore/blazor/routing.md
+++ b/aspnetcore/blazor/routing.md
@@ -105,7 +105,7 @@ In Blazor server-side apps, the default route in *_Host.cshtml* is `/` (`@page "
 @page "/{**path}"
 ```
 
-The "/{**path}" template includes:
+The `"/{**path}"` template includes:
 
 * Double-asterisk *catch-all* syntax (`**`) to capture the path across multiple folder boundaries without encoding forward slashes (`/`).
 * A `path` route parameter name.
--- a/aspnetcore/blazor/routing.md
+++ b/aspnetcore/blazor/routing.md
@@ -99,7 +99,7 @@ The route constraints shown in the following table are available. For the route 
 
 ### Routing with URLs that contain dots
 
-In Blazor server-side apps, the default route in *_Host.cshtml* is `/` (`@page "/"`). A route that contains a dot (`.`) isn't matched by the default route because such a route appears to request a file&mdash;the app returns a *404 - Not Found* response for a static file that doesn't exist. To use routes that contain a dot, configure *_Host.cshtml* with the following route template:
+In Blazor server-side apps, the default route in *_Host.cshtml* is `/` (`@page "/"`). A request URL that contains a dot (`.`) isn't matched by the default route because the URL appears to request a file&mdash;the app returns a *404 - Not Found* response for a static file that doesn't exist. To use routes that contain a dot, configure *_Host.cshtml* with the following route template:
 
 ```cshtml
 @page "/{**path}"
--- a/aspnetcore/blazor/routing.md
+++ b/aspnetcore/blazor/routing.md
@@ -5,7 +5,7 @@ description: Learn how to route requests in apps and about the NavLink component
 monikerRange: '>= aspnetcore-3.0'
 ms.author: riande
 ms.custom: mvc
-ms.date: 08/22/2019
+ms.date: 08/23/2019
 uid: blazor/routing
 ---
 # ASP.NET Core Blazor routing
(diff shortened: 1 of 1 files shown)


guardrex:
imo, this needs a little more work. I can touch it up after I hear back ...

Is this route template going to be the default for Pre9? ... or is this literally supposed to say that if they want this to work, they *must* set this route template up themselves going forward?

pranavkm:
>  or is this literally supposed to say that if they want this to work, they must set this route template up themselves going forward?

Yup. Users need to do this in their apps if they want to do this. It breaks 404 for missing static files which is why we don't want it on by default.

pranavkm:
Also, there's a separate issue tracking this for client side apps: https://github.com/aspnet/AspNetCore/issues/13247

guardrex:
kk ... see if that commit is *sane* 😵 ... I'm trying to make it dead simple to grok.

Wonder if it should say a little more about what `{**path}` specifically does? ... 'captures the whole path, including dots'?

guardrex:
How about we add something like the following at the end (after the example code) ...

> The `"/{**path}"` template includes:
>
> * A double-asterisk (`**`) to capture multiple directory boundaries.
> * `path` to capture the last segment of the URL as a path (not a file).

pranavkm:
Sure if you'd like to elaborate. Think there's some documentation about catch-all parameters elsewhere too if you'd like to point to it.

pranavkm:
Like this: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-2.2#create-routes

guardrex:
I forgot that was the route parameter name. See how that looks.

aspnetcore/blazor/routing.md
```diff
@@ -105,6 +105,13 @@ In Blazor server-side apps, the default route in *_Host.cshtml* is `/` (`@page "
 @page "/{**path}"
 ```
 
+The "/{**path}" template includes:
```
pranavkm: The `"/{**path}"` template includes:

aspnetcore/blazor/routing.md
```diff
@@ -97,6 +97,21 @@ The route constraints shown in the following table are available. For the route
 > [!WARNING]
 > Route constraints that verify the URL and are converted to a CLR type (such as `int` or `DateTime`) always use the invariant culture. These constraints assume that the URL is non-localizable.
 
+### Routing with URLs that contain dots
+
+In Blazor server-side apps, the default route in *_Host.cshtml* is `/` (`@page "/"`). A request URL that contains a dot (`.`) isn't matched by the default route because the URL appears to request a file&mdash;the app returns a *404 - Not Found* response for a static file that doesn't exist. To use routes that contain a dot, configure *_Host.cshtml* with the following route template:
```
Rick-Anderson: `&mdash;` doesn't MT.  Accept my following comment.

Rick-Anderson: In Blazor server-side apps, the default route in *_Host.cshtml* is `/` (`@page "/"`). A request URL that contains a dot (`.`) isn't matched by the default route because the URL appears to request a file. Blazor apps return a *404 - Not Found* response for a static file that doesn't exist. To use routes that contain a dot, configure *_Host.cshtml* with the following route template:

guardrex: I didn't know that ... is that new info from localization? `&mdash;` has been in use for years here. There are **_87_** of them all throughout the docs (~55 files).

Ping @scottaddie, @tdykstra, and @wadepickett to let them know, too.

How do you want to handle them? Eighty-seven makes it challenging to build into a single PR ... find-'n-replace won't be effective in rewriting sentences to get them out.

tdykstra: Removing dashes is not a candidate for a script, each instance would require thoughtful rewriting.  I would leave the ones in place that are already there and just be judicious about using them in the future. I have not seen a statement from loc against dashes.

guardrex: I still have quite a few 3.0 topic updates coming up. I'll nix 'um on those as I go. Over time, it should be possible to get them all out of the repo.

tdykstra: Wait a minute, some of those are mine, you can leave those in place. 😄

guardrex: I was referring to my assigned 3.0 update topics: The ones that I wrote were the vast majority of that group. If I come to one of your topics, I'll leave them be.

Rick-Anderson:
@guardrex stop adding mdash - we'll remove the others as we update the topics.

guardrex:
That's what I said I would do. @tdykstra is going a different way tho; so if I come across them in his topics, I'll leave them be.

Rick-Anderson:
@guardrex  can I S&M this?

guardrex:
@Rick-Anderson Ryan hasn't reviewed it yet.

pranavkm:
Let's get this in 👍

PR #13974, opened 2019-08-22
Add Blazor docs on externally invoking component methods to update state.

### LINKED ISSUE #13935: Document InvokeAsync() of a BaseComponent
In some cases the use of `InvokeAsync(Action workItem)` on a component is recommended, I read it somewhere in a PR from Steve for doing background work. However what are the best practises to use this function and when do you use it in the first place?

---
#### Documentdetails

⚠ *Dit gedeelte niet bewerken. Het is vereist om problemen te koppelen tussen docs.microsoft.com en GitHub.*

* Content: [Create and use ASP.NET Core Razor components](https://docs.microsoft.com/nl-nl/aspnet/core/blazor/components?view=aspnetcore-3.0#cascading-values-and-parameters)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

NTaylorMullen:
- Added a simple example and explanation of how you can update state externally and what `InvokeAsync` is used for and why.

Fixes #13935


guardrex:
A few style notes for the update: No gerunds (*-ing words) in titles and section headings unless the word is a technology name (e.g., "logging," "hosting," etc.), and our doc style rarely uses prompts (questions). I recommend *imperative tone*. See if my update broke the meaning tho. 💥🙈

guardrex:
*Excellent!* :+1:

guardrex:
Does the ...

```csharp
// Can be called from anywhere
```

... code comment go with `Notify` or with the whole `Update` method? If `Update`, I recommend moving it up above the method.

NTaylorMullen:
Goes with the Update, i'll move it up

guardrex:
Merge it whenever ur ready (idk if you want further engineering :eyes: on it).

NTaylorMullen:
Ya gonna wait for @rynowak / @pranavkm or @SteveSandersonMS to approve given that it's a new section in the sacred component doc 😄

guardrex:
*Yes!* A Microsoft keyboard must be sacrificed to appease the :pray: *server gods* :pray: before merging.

Rick-Anderson:
@rynowak / @pranavkm or @SteveSandersonMS  short review please.

aspnetcore/blazor/components.md
```diff
@@ -541,6 +541,56 @@ While capturing component references use a similar syntax to [capturing element
...
+    {
+        Notifier.Notify += OnNotify;
+    }
+
+    public async void OnNotify(string key, int value)
+    {
+        lastNotification = (key, value);
+
+        await InvokeAsync(StateHasChanged);
+    }
```
SteveSandersonMS: Possible improvement:

```cs
    async Task OnNotifyAsync()
    {
        return InvokeAsync(() =>
        {
            lastNotification = (key, value);
            StateHasChanged();
        });
    }
```

This is better because:

 * It moves the error handling responsibility upstream. We're still ignoring any errors (like you had `async void` before), but now if you *wanted* to know about them from the caller you could.
 * It avoids potential thread safety issues. If instead of `lastNotification = (key, value)` you had a large block of complex logic, you might not realise there could be multiple concurrent invocations, given that most component logic assumes one-concurrent-execution within the sync context. In most cases it's better to run your update logic on the sync context to avoid needing to reason about different thread safety considerations.

If you do make this change, you'll need to change the caller because the signature is different. Maybe it's also possible to have the caller handle exceptions in some sensible way but I'm not sure what (just log them?).

SteveSandersonMS: Not sure what you mean by "a lot of complexity". I'd suggest updating the code to match what I posted, but not necessarily also adding all the explanations I've given (I only posted that as justification for us).

Even if we're not explaining all the reasons for why our code samples are as they are, it's still good for them to be robust where possible so people can safely copy/paste them, right?

SteveSandersonMS: Oh perhaps you were referring to the thoughts about error handling. I don't think it's necessary to put that in these samples really. We'd just need to change the calling code to handle the fact that it's invoking a method called `OnNotifyAsync` instead of `OnNotify`.

mkArtakMSFT:
@NTaylorMullen do you expect to make further changes here?

guardrex:
@danroth27 This is one for the staging branch. Not sure it's status. :point_up:

aspnetcore/blazor/components.md
danroth27: I'm in favor of the changes @SteveSandersonMS is proposing here.

pranavkm: Made a couple of small tweaks to the text.

guardrex:
@danroth27 Look good to :shipit:? This is the last PR for the staging branch, and we can go live right after this goes in and I rebase.

Ricardoaraujo00:
Hi there.
Are there any developments about an explanation on how to update a Blazor page with received data, for example, from a gRPC service?
I tried with a thread safe sigleton, not thread safe, static class, and the UI only updates when i navigate to another page on the application and come back. Then the info is displayed. I want it to be displayed as soon as it is received, like having a bind. 

With the static classe atempt i have this in it:
```
public static class InstanceTransporterClass
    {
        static string _queue;
        public static string queue {
            get =>_queue;
            set
            {
                _queue = value;
                NotifyDataChanged();
            }
        }

        public static event Action OnChange;

        private static void NotifyDataChanged() => OnChange?.Invoke(StateHasChanged);
    }
```
But the invoke method throws an error:
`System.InvalidOperationException: The current thread is not associated with the Dispatcher. Use InvokeAsync() to switch execution to the Dispatcher when triggering rendering or component state.`

I dont remember if with the singleton i had this error but the UI didn't updated either...

Any ideas on how this can be solved?

SteveSandersonMS:
> But the invoke method throws an error:
System.InvalidOperationException: The current thread is not associated with the Dispatcher. Use InvokeAsync() to switch execution to the Dispatcher when triggering rendering or component state.

The error message is trying to describe the solution. You do need to wrap your `StateHasChanged` call inside a call to `InvokeAsync` on the component, e.g.:

```
// Code in your component that subscribes to an external event
myService.OnSomeEvent += (sender, eventArgs) => {
    InvokeAsync(StateHasChanged);
};
```

PR #13984, opened 2019-08-23
Fix and match tab groups in GRPC topic

guardrex:
[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/tutorials/grpc/grpc-start?view=aspnetcore-3.0&branch=pr-en-us-13984)

Quick patch for a couple of nits and tab groups. A warning cropped up this morning on another PR ...

> | invalid-tab-group | aspnetcore/tutorials/grpc/grpc-start.md | Tab group with different tab id set. |

I think it was from a missing paren on https://github.com/aspnet/AspNetCore.Docs/pull/13954.

### DIFF
--- a/aspnetcore/tutorials/grpc/grpc-start.md
+++ b/aspnetcore/tutorials/grpc/grpc-start.md
@@ -131,7 +131,7 @@ info: Microsoft.Hosting.Lifetime[0]
 
 ## Create the gRPC client in a .NET console app
 
-## [Visual Studio](#tab/visual-studio)
+# [Visual Studio](#tab/visual-studio)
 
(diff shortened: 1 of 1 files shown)


guardrex:
ok ... think that got it. There was also an inconsistent tab heading level.

Note that this topic isn't internally consistent on tab heading levels. I thought that we were always using 1st level headings for tabs (`#`) and ignoring the linter, which complains about the heading levels.

Anyway ... I left that alone, but I made sure that the tabs in each group use the same heading levels. If you want all tabs on 1st level, let me know, and I'll add a commit here to fix that.

PR #13986, opened 2019-08-23
Validate nested sub-properties

pranavkm:

### DIFF
--- a/aspnetcore/blazor/forms-validation.md
+++ b/aspnetcore/blazor/forms-validation.md
@@ -5,7 +5,7 @@ description: Learn how to use forms and field validation scenarios in Blazor.
 monikerRange: '>= aspnetcore-3.0'
 ms.author: riande
 ms.custom: mvc
-ms.date: 08/13/2019
+ms.date: 08/23/2019
 uid: blazor/forms-validation
 ---
 # ASP.NET Core Blazor forms and validation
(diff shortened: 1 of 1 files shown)


guardrex:
@pranavkm Where does `HomeAddress` come from?

pranavkm:
Fixed

Rick-Anderson:
@SteveSandersonMS one minute review please.

SteveSandersonMS:
I'm not sure this approach is quite right. The `IValidatableObject` technique is going to end up registering error messages with the wrong keys. I think @pranavkm also saw this happening.

Making this work cleanly involves some logic changes inside `<DataAnnotationsValidator>` to make it recognize the concept of nested objects. I posted a sample of this here: https://gist.github.com/SteveSandersonMS/86a0bd3be0826f57af1ae2d3bd5fa95c. However that is a bit complicated to have in docs! Our intention is to bake this functionality into the framework in 3.1 (probably - cc @danroth27 @mkArtakMSFT).

As for what we should document for 3.0, I'm not sure.

@pranavkm What problems occur due to the incorrect dictionary keys? Is it good enough that we can tell people to do this in 3.0 then change the advice in 3.1?

SteveSandersonMS:
I don't think we should suggest an approach that leads to buggy behavior (e.g., duplicated validation error messages). I'd prefer to hold off on giving any advice until we have a reasonable story for it.

pranavkm:
@guardrex the sample was merged. All yours

guardrex:
@pranavkm Tough to describe! :smile: See if the update is sane. If so, :shipit:.

pranavkm:
Let's get this merged in 👍

Issue #13944, opened 2019-08-21
Question: Expiration allmost never happens on it's own in the background


Rick-Anderson:
a

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 13f5f8f9-18dd-9c12-9ed1-fcc6b44ba7fd
* Version Independent ID: 640b4f36-5886-7b69-1c95-cfc8574a02fd
* Content: [Cache in-memory in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/memory?view=aspnetcore-3.0)
* Content Source: [aspnetcore/performance/caching/memory.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/performance/caching/memory.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-performance**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
@serpent5 can you review the following 3.0 version sections:

* [MemoryCache.Compact](https://docs.microsoft.com/en-gb/aspnet/core/performance/caching/memory?view=aspnetcore-3.0#memorycachecompact)
* the section starting with *A cached item set with sliding expiration only is at risk for becoming stale because there is no bound to its expiration.s*

serpent5:
> * [MemoryCache.Compact](https://docs.microsoft.com/en-gb/aspnet/core/performance/caching/memory?view=aspnetcore-3.0#memorycachecompact)

This LGTM. 

> * the section starting with _A cached item set with sliding expiration only is at risk for becoming stale because there is no bound to its expiration.s_

I have a few suggestions for the wording in this section. Shall I PR something for that?

Rick-Anderson:
Yes, please

serpent5:
Ok, will do. I should be able to look tomorrow evening at this and the outstanding Razor compilation issue we’ve discussed.

ChrisRichner:
Sorry, did you mention me by accident?

Rick-Anderson:
@ChrisRichner  no, I was hoping you could look at [Additional Notes](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/memory?view=aspnetcore-3.0#additional-notes) because that came from your issue.

ChrisRichner:
Right, long time ago :)

Looks good to me 👍

PR #13996, opened 2019-08-23
Slight rework of combining Sliding and Abs IMemoryCache expirations.

serpent5:
@Rick-Anderson 

This one's for #13944, but it's not a "fix", just some suggestions. I know it goes without saying, but feel free to just take parts that you like or just throw it all away. :smirk_cat:


Rick-Anderson:
Much better, Thanks!

PR #14049, opened 2019-08-27
Add VS for Mac Tab (restored)

LouisT123:
Restoring current state of last closed pr, will work on VS4Mac invalid-tab-group conflict


sayedihashimi:
@Rick-Anderson If i understand correct we will need to add VSMac tabs to all 8 tutorial pages in one go. That may take some time, but I'll see when I can get to work on that.

PR #14054, opened 2019-08-28
Update AccountController.cs

allanchua101:
- Fixed comment to indicate code execution instead of using the word "we".


allanchua101:
Thanks @guardrex 👍

PR #14084, opened 2019-08-29
Update request-response.md

jkotalik:
Fixes for 3.0 api changes.

### DIFF
--- a/aspnetcore/fundamentals/middleware/request-response.md
+++ b/aspnetcore/fundamentals/middleware/request-response.md
@@ -70,7 +70,7 @@ Both `Body` and `BodyReader/BodyWriter` properties are available for `HttpReques
 
 ## StartAsync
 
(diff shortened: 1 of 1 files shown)


guardrex:
@jkotalik ❓ ... The whole topic is versioned 3.0, so we generally try not to say anything about the version in such cases as a future-proofing effort (i.e., when ASP.NET Core reaches 4.x). Is it ok if I put a commit in here to take out the "3.0" language?

jkotalik:
Sure.

guardrex:
@jkotalik See if the changes on the commit are 😵 *sane* 😵 ... *you know how I like to 💥 things from time-to-time.*

There's a remark in here ...

> In 3.0-preview3, you must call `StartAsync` before using `HttpRequest.BodyPipe`. In future releases, the sequence will be a recommendation.

It's a bit awkward calling out a specific preview like that. Can I rephrase that somehow? Is the "future release" **_here today_** so that I can rephrase it to just say the recommendation?

guardrex:
btw just so you know ... the primary change set grammar+style-wise includes:

* Dropping most of the 2nd person.
* Dropping future tense.
* Use asterisks for list markers.
* Use contractions.
* Break a couple of long sentences (when they get up into the 25 word length).

This could use some more API xrefs ... but I'm too 🏃 right now to put them in. Those can wait for a future update by one of the doc authors.

jkotalik:
> It's a bit awkward calling out a specific preview like that. Can I rephrase that somehow? Is the "future release" here today so that I can rephrase it to just say the recommendation (i.e., drop the "future release" lingo)?

Go ahead and drop the entire line.

PR #14091, opened 2019-08-30
3.0 updates to web API action return types doc

scottaddie:
Fixes https://github.com/aspnet/AspNetCore.Docs/issues/12650
Fixes https://github.com/aspnet/AspNetCore.Docs/issues/13876
Fixes https://github.com/aspnet/AspNetCore.Docs/issues/14093

**Summary of changes**
- Removed the 2 tips that distract from the doc's focus
- Added a working 3.0 sample project
- Conducted an edit pass on the entire doc
- Added an example of returning an async stream (3.0+)

[Internal Review Page](https://review.docs.microsoft.com/aspnet/core/web-api/action-return-types?view=aspnetcore-3.0&branch=pr-en-us-14091)


scottaddie:
@BillWagner Please take a look at the async stream code in this PR.

aspnetcore/web-api/action-return-types/samples/3x/WebApiSample.Api.30/Controllers/ProductsController.cs
```diff
@@ -0,0 +1,98 @@
...
+        public async IAsyncEnumerable<Product> GetByPageChunk(
+            int pageNumber, 
+            int pageSize)
+        {
+            int firstRecord = pageNumber * pageSize;
+            int stopRecord = firstRecord + pageSize;
+
+            while (firstRecord < stopRecord)
+            {
+                var products = _repository.GetProductsByPage(pageNumber, pageSize);
```
BillWagner: Since the 2nd arg is `pageSize`, I think this would always call `GetProductsByPage` exactly once. If `pageSize` was hard coded to `10`, as you mention in the text, this would behave as you described.

aspnetcore/web-api/action-return-types/samples/3x/WebApiSample.Api.30/Controllers/ProductsController.cs
```diff
@@ -0,0 +1,98 @@
...
+
+                await foreach (var product in products)
+                {
+                    yield return product;
+                }
+
+                firstRecord += pageSize;
```
BillWagner: A similar change to the arg here would be needed depending on the change from my first comment.

aspnetcore/web-api/action-return-types/samples/3x/WebApiSample.Api.30/Controllers/ProductsController.cs
scottaddie: I think I see the source of the confusion. I need to clarify in my explanation of the example. The `pageSize` parameter comes from the route parameter (10 in this case). The `pageNumber` parameter comes from the route parameter too (1 in this case).

BillWagner: I really like the final sample. This is a perfect scenario for AsyncEnumerables.

aspnetcore/web-api/action-return-types.md
```diff
@@ -36,60 +36,99 @@ This document explains when it's most appropriate to use each return type.
...
+When known conditions need to be accounted for in an action, multiple return paths are introduced. In such a case, it's common to mix an <xref:Microsoft.AspNetCore.Mvc.ActionResult> return type with the primitive or complex return type. Either [IActionResult](#iactionresult-type) or [ActionResult\<T>](#actionresultt-type) are necessary to accommodate this type of action.
+
+::: moniker range=">= aspnetcore-3.0"
+
+### Asynchronous stream
+
+Memory consumption on the server becomes a performance consideration when returning large data sets. Imagine an action that calls <xref:System.Linq.Enumerable.ToList*> on a LINQ query. The query returns an entire product catalog, consisting of hundreds of thousands of products. Each product record is stored in server memory. An asynchronous stream can relieve the burden on the server.
```
rynowak: This is not true sadly. We don't stream individual records into the response.

What this feature does is make using thing that return `IAsyncEnumerable<>` as efficient as `IEnumerable<>`.

aspnetcore/web-api/action-return-types.md
```diff
@@ -36,60 +36,99 @@ This document explains when it's most appropriate to use each return type.
...
+
+[!code-csharp[](../web-api/action-return-types/samples/3x/WebApiSample.Api.30/Controllers/ProductsController.cs?name=snippet_GetNPages)]
+
+The `for` loop results in three iterations. Iterations one and two each retrieve 10 product records. Iteration three retrieves four product records. The `yield return product;` statement executes 24 times for the lifetime of the entire request. Product records are returned individually until all 24 products have been sent to the client. The client can begin its processing upon receiving the first product record. The synchronous form of this action would have blocked the client until all 24 records had been retrieved from the server.
```
rynowak: Yep, none of these things are true.

aspnetcore/web-api/action-return-types.md
rynowak: The problem that we solved:

- If you returned an `IAsyncEnumerable<>` in 2.2 we would iterated it **synchronously** - which results in blocking calls and potential thread-pool starvation. The work we did in 3.0 makes this naive code just as efficent as using `IEnumerable<>`.

We can't implement streaming for real until the JSON serializer has it.

aspnetcore/web-api/action-return-types.md
```diff
@@ -36,60 +36,105 @@ This document explains when it's most appropriate to use each return type.
...
+When known conditions need to be accounted for in an action, multiple return paths are introduced. In such a case, it's common to mix an <xref:Microsoft.AspNetCore.Mvc.ActionResult> return type with the primitive or complex return type. Either [IActionResult](#iactionresult-type) or [ActionResult\<T>](#actionresultt-type) are necessary to accommodate this type of action.
+
+::: moniker range=">= aspnetcore-3.0"
+
+### Return IAsyncEnumerable\<T>
+
+In ASP.NET Core 3.0 or later, a web API action can return <xref:System.Collections.Generic.IAsyncEnumerable%601>. The `IAsyncEnumerable<T>` support addresses the common mistake of returning an `IEnumerable<T>` type from an action. Most of .NET Standard's [strongly typed collection types](/dotnet/api/system.collections.generic) implement `IEnumerable<T>`. A common example of such a collection type is <System.Collections.Generic.List%601>. An action that returns an `IEnumerable<T>` type results in blocking because the serializer synchronously iterates the collection.
```
rynowak: This isn't really an accurate description. It's not a bug to return `IEnumerable<>`. 

The problem is when you return an `IEnumerable<>` that **requires async work**.

aspnetcore/web-api/action-return-types/samples/3x/WebApiSample.Api.30/Controllers/ProductsController.cs
```diff
@@ -0,0 +1,101 @@
...
+        #region snippet_GetNRecordsAsync
+        [HttpGet("page/{pageSize:int:min(1)}")]
+        public async IAsyncEnumerable<Product> GetNRecordsAsync(int pageSize)
+        {
+            var products = _repository.GetProductsByPageAsync(1, pageSize);
+            
+            await foreach (var product in products)
+            {
+                yield return product;
+            }
```
rynowak: There's no value in doing this, you can just return `products`.

scottaddie:
@rynowak Please take another look.

aspnetcore/web-api/action-return-types.md
```diff
@@ -36,60 +36,97 @@ This document explains when it's most appropriate to use each return type.
...
+When known conditions need to be accounted for in an action, multiple return paths are introduced. In such a case, it's common to mix an <xref:Microsoft.AspNetCore.Mvc.ActionResult> return type with the primitive or complex return type. Either [IActionResult](#iactionresult-type) or [ActionResult\<T>](#actionresultt-type) are necessary to accommodate this type of action.
+
+::: moniker range=">= aspnetcore-3.0"
+
+### Return IAsyncEnumerable\<T>
+
+In ASP.NET Core 2.2 or earlier versions, returning <xref:System.Collections.Generic.IAsyncEnumerable%601> from a web API action results in synchronous iteration by the serializer. The result is the blocking of calls and a potential for thread pool starvation. In ASP.NET Core 3.0 or later, returning `IAsyncEnumerable<T>` from a web API action is as efficient as returning <xref:System.Collections.Generic.IEnumerable%601>.
```
rynowak: Note that this behavior is based on the underlying value that's returned, not the declared typed of the method. I recommend that users declare the return type as `IAsyncEnumerable<>` where possible since that's the guarantee you want.

aspnetcore/web-api/action-return-types/samples/3x/WebApiSample.DataAccess/Repositories/ProductsRepository.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            var products = _context.Products
+                                   .Skip(pageSize * (pageNumber - 1))
+                                   .Take(pageSize)
+                                   .AsAsyncEnumerable();
+
+            await foreach (var product in products)
+            {
+                yield return product;
+            }
```
rynowak: This isn't needed (L72/74) - it has no effect.

aspnetcore/web-api/action-return-types/samples/3x/WebApiSample.DataAccess/Repositories/ProductsRepository.cs
```diff
@@ -0,0 +1,92 @@
...
+                        Description = "A best-selling book covering the fundamentals of C#"
+                    });
+                _context.SaveChanges();
+            }
+        }
+
+        public List<Product> GetProducts() =>
+            _context.Products.OrderBy(p => p.Name).ToList();
+
+        public IEnumerable<Product> GetProductsByPage(
```
rynowak: I would expect that this method returns a type that implements **both** IAsyncEnumerable<> and IEnumerable<>.

You're having trouble demonstrating the problem because we fixed the problem 😆 .

If you want a good sample of this, you have to put a foreach loop in both this and `GetProductsByPage` that does something in the loop.

scottaddie: > I would expect that this method returns a type that implements both IAsyncEnumerable<> and IEnumerable<>.

@rynowak Do you have an example of such a collection type? Or would this be a custom collection type I create that implements both interfaces?

scottaddie:
Adding @stephentoub for the `IEnumerable<T>` / `IAsyncEnumerable<T>` discussion in https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Api.Analyzers. Thoughts?

aspnetcore/web-api/action-return-types/samples/3x/WebApiSample.DataAccess/Repositories/ProductsRepository.cs
stephentoub: @rynowak, can you clarify your expectation?  I expect it'll be super rare for a type to implement both interfaces.  The design of having an `IEnumerable<T> Foo` and `IAsyncEnumerable<T> FooAsync` is much more in line with what I'd expect to see.

rynowak: https://github.com/aspnet/EntityFrameworkCore/blob/24b9aa1d2e14fe2e737255ede9b2a7a623fcf2af/src/EFCore/DbSet.cs

**LOTS** of types in EF.

We did this feature to solve a very specific problem.

```C#
public IEnumerable<Project> GetOnSaleProducts()
{
    return db.Products.Where(p => p.IsOnSale);
}
```

In previous versions of MVC - this would **SYNCHRONOUSLY** enumerate the results of that query while serializing the data.

It's totally non-obvious to a user that this is doing blocking waits on the database. To write this code correctly in 2.2, you would write:

```C#
public IEnumerable<Project> GetOnSaleProducts()
{
    return db.Products.Where(p => p.IsOnSale).ToListAsync();
}
```

---

In 3.0 you can write the naive code - and we will buffer the results before handing it off to the serializer:

```C#
public IEnumerable<Project> GetOnSaleProducts()
{
    return db.Products.Where(p => p.IsOnSale);
}
```

This works because:
- MVC looks at the concrete type of the value returned by your action, and buffers it if it's `IAsyncEnumerable<>`
- The **declared** return types of EF's query stuff are `IQueryable<>` and related types, but the implementations of `IQueryable<>` all also implement `IAsyncEnumerable<>`.

This kind of design approach is necessary because `IAsyncEnumerable<>` was adding to existing APIs.

---

In 5.0 when S.T.J supports `IAsyncEnumerable<>` we'll remove the buffering and users can have end-to-end streaming.

rynowak: @scottaddie  - if you want a good example of this:

```c#
public IEnumerable<Product> GetOnSaleProducts()
{
    var products = repository.GetProducts();
    foreach (var product in products)
    {
        if (product.IsOnSale)
        {
             yield return product;
        }
    }
}

public async IAsyncEnumerable<Product> GetOnSaleProducts()
{
    var products = repository.GetProductsAsync();
    await foreach (var product in products)
    {
        if (product.IsOnSale)
        {
             yield return product;
        }
    }
}
```

rynowak: When you add some user code - the example becomes more useful

stephentoub: > if you want a good example of this

Exactly.  This example doesn't have one method that returns something that implements both `IEnumerable<T>` and `IAsyncEnumerable<T>`; it has two methods, one returning `IEnumerable<T>` and one returning `IAsyncEnumerable<T>`.  Maybe I misunderstood your comment "I would expect that this method returns a type that implements both IAsyncEnumerable<> and IEnumerable<>."?

scottaddie: I think we both misunderstood that comment, @stephentoub. Shout out to @rynowak for all of the help. I learned a lot of useful stuff through the work on this PR. I'm confident the community will benefit from this collaboration.

scottaddie:
@rynowak Ready for another look.

scottaddie:
@rynowak Bump.

aspnetcore/web-api/action-return-types/samples/3x/WebApiSample.DataAccess/Repositories/ProductsRepository.cs
rynowak: Perhaps my statements were a little confusing. 😁  

This feature helps two cases:
- EF's return types that implement both APIs
- User-code that returns IAsyncEnumerable

I'm only separating these two because in the former case, it's totally non-obvious to a user that they were working with IAsyncEnumerable because it's not part of the declared type (IQueryable).

aspnetcore/web-api/action-return-types.md
```diff
@@ -36,60 +36,121 @@ This document explains when it's most appropriate to use each return type.
...
+### Return IEnumerable\<T> or IAsyncEnumerable\<T>
+
+In ASP.NET Core 2.2 and earlier, returning <xref:System.Collections.Generic.IAsyncEnumerable%601> from an action results in synchronous collection iteration by the serializer. The result is the blocking of calls and a potential for thread pool starvation. To illustrate, imagine that Entity Framework (EF) Core is being used for the web API's data access needs. The following action's return type is synchronously enumerated during serialization:
+
+```csharp
+public IEnumerable<Product> GetOnSaleProducts() =>
+    _context.Products.Where(p => p.IsOnSale);
+```
+
+To avoid synchronous enumeration and blocking waits on the database in ASP.NET Core 2.2 and earlier, invoke `ToListAsync`:
```
rynowak: 👍

aspnetcore/web-api/action-return-types.md
```diff
@@ -36,60 +36,121 @@ This document explains when it's most appropriate to use each return type.
...
+* Becomes as efficient as returning <xref:System.Collections.Generic.IEnumerable%601>.
+
+ASP.NET Core 3.0 and later buffers the result of the following action before providing it to the serializer:
+
+```csharp
+public IEnumerable<Product> GetOnSaleProducts() =>
+    _context.Products.Where(p => p.IsOnSale);
+```
+
+Consider declaring the action signature's return type as `IAsyncEnumerable<T>` to guarantee the asynchronous iteration. Ultimately, the iteration mode is based on the underlying concrete type being returned. MVC automatically buffers any concrete type that implements `IAsyncEnumerable<T>`. EF Core queries commonly return `IQueryable<T>`. Types that implement `IQueryable<T>` also implement `IAsyncEnumerable<T>`.
```
rynowak: > Types that implement `IQueryable<T>` also implement `IAsyncEnumerable<T>`.

This is not a true fact. This is try of EF's types. I think some amount of details could be left out here.

PR #14106, opened 2019-09-02
Improve cascaded values/params content

### LINKED ISSUE #14102: Suggestion/Improvment
Currently:
`Binding with a string name value is relevant if you have multiple cascading values of the same type and need to differentiate them within the same subtree.`

better (imo):
If you have multiple cascading values of the same type, which have to be differentiated within the same subtree, a `Name` attribute with a unique value should be applied to each `CascadingValue` and `CascadingParameter` in order to distinguish each of them.
See example below:

---
#### Document Details

* Content: [Create and use ASP.NET Core Razor components](https://docs.microsoft.com/en-us/aspnet/core/blazor/components?view=aspnetcore-3.0#cascading-values-and-parameters)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #14102

* Move coverage to the bottom of the section because the current content breaks right into the middle of the example without naming.
* Flesh out the code example with additional detail. For example, include both the ancestor and descendant components.

Thanks @Legends! :rocket:

### DIFF
--- a/aspnetcore/blazor/components.md
+++ b/aspnetcore/blazor/components.md
@@ -5,7 +5,7 @@ description: Learn how to create and use Razor components, including how to bind
 monikerRange: '>= aspnetcore-3.0'
 ms.author: riande
 ms.custom: mvc
-ms.date: 08/13/2019
+ms.date: 09/02/2019
 uid: blazor/components
 ---
 # Create and use ASP.NET Core Razor components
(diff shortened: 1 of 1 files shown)


guardrex:
@Legends

* If it's permissible to show both a prop and a field, I like it. An alternative tho is to call out in text that both work. Let's see what Dan says to do.
* WRT SxS or nested, depends on what engineering meant by "within the same subtree.". I'll update that PR to the nested presentation if that turns out to be what they meant (*if GH will let me!* 🙈 ... it usually doesn't stop me from committing to my own PRs via the interface ... *usually!* lol).

Legends:
> @Legends
> 
> * If it's permissible to show both a prop and a field, I like it. An alternative tho is to call out in the text that both work. Let's see what Dan says to do.
> * WRT SxS or nested, depends on what engineering meant by "within the same subtree." I'll update that PR to the nested presentation if that turns out to be what they meant (_if GH will let me!_ 🙈 ... it usually doesn't stop me from committing to my own PRs via the interface ... _usually!_ lol).

1.) Fields, Props, Props with Attribute, everything is legit.
2.) I guess it has to be nested, otherwise it would be in another side branch.

No hurries :-)

guardrex:
Good to be able to commit to **_my own PR_** ... *this horrible hurricane over here is messing everything up.* :smile:

guardrex:
Thanks @Legends! 🏍

PR #14121, opened 2019-09-03
Blazor EventHandlers and BindAttributes moving

### LINKED ISSUE #13832: [Preview 9] EventHandlers and BindAttributes moving to `M.A.Components.Web` namespace
This is an impactful enough change that it needs its own docs information in preview 9. This is a summary for now, and I'll write detailed guidance later.

---

## Summary

In 3.0-preview9 and later, the set of bindings and event handlers available for HTML elements are defined in the `Microsoft.AspNetCore.Components.Web` assembly and based on the `Microsoft.AspNetCore.Components.Web` namespace.

### Blazor WebAssembly and Libraries

When migrating an application to preview 9 or later, make sure to add a reference to the `Microsoft.AspNetCore.Components.Web` nuget package. Next, open the application's root `_Imports.razor` and add `@using Microsoft.AspNetCore.Components.Web`.

### Blazor Server-Side

When migrating an application to preview 9 or later, open the application's root `_Imports.razor` and add `@using Microsoft.AspNetCore.Components.Web`.

note: Server-Side Blazor applications get a reference to `Microsoft.AspNetCore.Components.Web` implicitly without any additional package references.

## Troubleshooting guidance

With the correct references and `@using`, event handlers like `@onclick` and `@bind` should be bold font, and colorized as shown below when using Visual Studio.

image

If `@bind` or `@onclick` is colorized as a normal HTML attribute, like the below, then the `@using` is missing.

image

You may also see errors like:
```txt
CS0169	The field 'Index.text' is never used
CS0428	Cannot convert method group 'Submit' to non-delegate type 'object'. Did you intend to invoke the method?
```

Try adding `@using Microsoft.AspNetCore.Components.Web` to the file. If you now see an error like the following, this means that the reference to the `Microsoft.AspNetCore.Components.Web` package is missing from the project.
```txt
CS0234	The type or namespace name 'Web' does not exist in the namespace 'Microsoft.AspNetCore.Components' (are you missing an assembly reference?)
```

If adding the using fixed the problem, consider moving to the application's root `_Imports.razor` so it will apply to all files.

guardrex:
Fixes #13832

**This is for master.**

### DIFF
--- a/aspnetcore/migration/22-to-30.md
+++ b/aspnetcore/migration/22-to-30.md
@@ -633,8 +633,6 @@ When migrating an app to ASP.NET Core Preview 9 or later, open the app's root *_
 
 ### Troubleshoot
 
-**Syntax highlighting in Visual Studio**
-
 If `@bind` or `@on...` attributes are colorized as a normal HTML attributes (shown in the following image), the `@using` statement is missing:
 
 ![Index.razor component CSHTML markup in Visual Studio showing HTML colorized unemphasized font for bind and onclick attributes.](https://user-images.githubusercontent.com/1430011/64197856-f1679300-ce3b-11e9-83b7-71ffa6d68fec.png)
--- a/aspnetcore/migration/22-to-30.md
+++ b/aspnetcore/migration/22-to-30.md
@@ -654,4 +654,4 @@ Try adding `@using Microsoft.AspNetCore.Components.Web` to the file. If you now 
 CS0234 The type or namespace name 'Web' does not exist in the namespace 'Microsoft.AspNetCore.Components' (are you missing an assembly reference?)
 ```
 
-If adding the `@using` statement fixes the problem, consider moving to the app's root *_Imports.razor* so that it applies to all files.
+If adding the `@using` statement fixes the problem, consider moving the statement to the app's root *_Imports.razor* so that it applies to all files.
--- a/aspnetcore/migration/22-to-30.md
+++ b/aspnetcore/migration/22-to-30.md
@@ -633,7 +633,7 @@ When migrating an app to ASP.NET Core Preview 9 or later, open the app's root *_
 
 ### Troubleshoot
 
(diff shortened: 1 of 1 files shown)


aspnetcore/migration/22-to-30.md
```diff
@@ -607,3 +607,57 @@ If you're using features of `Newtonsoft.Json` that aren't supported in `System.T
...
+
+![Index.razor component CSHTML markup in Visual Studio showing Razor colorized bold font for bind and onclick attributes.](https://user-images.githubusercontent.com/1430011/64197727-a3528f80-ce3b-11e9-8cc4-fe34a858eb49.png)
+
+**Missing Microsoft.AspNetCore.Components.Web package reference**
+
+The following error appears if the reference to the `Microsoft.AspNetCore.Components.Web` package is missing from the project:
+
+```
+CS0169 The field 'Index.text' is never used
+CS0428 Cannot convert method group 'Submit' to non-delegate type 'object'. Did you intend to invoke the method?
```
rynowak: This isn't really right. These errors can occur due to either issue. Depending on the exact code you wrote, you may or may not see these issues.

aspnetcore/migration/22-to-30.md
```diff
@@ -607,3 +607,57 @@ If you're using features of `Newtonsoft.Json` that aren't supported in `System.T
...
+```
+
+**The @using statement isn't applied to all project files**
+
+If the `@using` statement isn't applied to all files, the following error appears:
+
+```
+CS0234 The type or namespace name 'Web' does not exist in the namespace 'Microsoft.AspNetCore.Components' (are you missing an assembly reference?)
```
rynowak: This error happens when you're missing the assembly reference.

guardrex:
I use your exact format now ... should've done that the first time around.

guardrex:
ok ... ready :eyes: :ear:

danroth27:
This is preview to preview migration content. I don't think this is the right doc for this content. This doc is for migrating from 2.2 to 3.0. Since I've got this content covered in the blog post I think we can just close this PR.

PR #14161, opened 2019-09-05
Ensure VS Code doesn't trim significant whitespace

halter73:
"files.trimTrailingWhitespace" already defaults to false in VS Code, but some people like me might have this configured to true in their user setting.

See https://github.com/aspnet/AspNetCore.Docs/pull/13844/files#r321423599

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,6 @@
 {
-    "omnisharp.autoStart": false
+    "omnisharp.autoStart": false,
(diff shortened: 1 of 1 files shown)


guardrex:
@scottaddie Will need to approve this. I'm just a little green dinosaur in a big MS world. :smile:

scottaddie:
Thanks for bringing this to our attention and fixing, @halter73

PR #14200, opened 2019-09-06
Link fix related to 14108

wadepickett:
Changing accidental contentRootPath link to Webroot link on Generic Host doc.

This fix should have been in https://github.com/aspnet/AspNetCore.Docs/pull/14108#discussion-diff-321499728R267     Discovered just as it was merged however so fixing here.

Fixes issue https://github.com/aspnet/AspNetCore.Docs/issues/10559


wadepickett:
Thanks Tom!

PR #14208, opened 2019-09-07
Call out the URL for Windows Service sample app

### LINKED ISSUE #14193: port configuration
Hi team, 
I moved this from here: https://github.com/aspnet/AspNetCore.Docs.de-de/issues/51
The contributor is making a suggestion to the original documentation

@taori commented 10 hours ago
In case someone hosts a windows application it would probably be best if this sample also contained the necessary snippets to configure that so the service which is running can be navigated to through a browser.

adding the following to appsettings.Production.json to this made it work:

```
"Kestrel": {
	"EndPoints": {
		"Http": {
			"Url": "http://localhost:4300"
		}
	}
},


```

---
#### Document Details

* Content: [Host ASP.NET Core in a Windows Service](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/windows-service?view=aspnetcore-2.2&tabs=visual-studio#feedback)
* Content Source: [aspnetcore/host-and-deploy/windows-service.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/host-and-deploy/windows-service.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #14193

* Calls out the default URL (http://localhost:5000) in the sample apps' *README.md* files.
* Slight touch to *appsettings.json* to mirror what the `worker` template app does.
* I suppose that `AllowedHosts` in the settings file is ok given that this is a web app service and Host Filtering Middleware would work here if added. Let me know if you'd like that 🔪 out of here.

### DIFF
--- a/aspnetcore/host-and-deploy/windows-service/samples/2.x/AspNetCoreService/README.md
+++ b/aspnetcore/host-and-deploy/windows-service/samples/2.x/AspNetCoreService/README.md
@@ -1,3 +1,5 @@
-# ASP.NET Core Windows Service Sample (Framework-dependent Deployment)
+# ASP.NET Core Windows Service Sample
 
 This sample shows how to host an ASP.NET Core app as a Windows Service without using IIS. This sample demonstrates the scenario described in [Host ASP.NET Core in a Windows Service](https://docs.microsoft.com/aspnet/core/host-and-deploy/windows-service).
+
+After establishing and starting the service, access the app in a browser at `http://localhost:5000/`.
--- a/aspnetcore/host-and-deploy/windows-service.md
+++ b/aspnetcore/host-and-deploy/windows-service.md
@@ -5,7 +5,7 @@ description: Learn how to host an ASP.NET Core app in a Windows Service.
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 06/03/2019
+ms.date: 09/09/2019
 uid: host-and-deploy/windows-service
 ---
 # Host ASP.NET Core in a Windows Service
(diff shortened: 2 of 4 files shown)


aspnetcore/host-and-deploy/windows-service/samples/2.x/AspNetCoreService/README.md
```diff
@@ -1,3 +1,5 @@
-# ASP.NET Core Windows Service Sample (Framework-dependent Deployment)
+# ASP.NET Core Windows Service Sample
 
 This sample shows how to host an ASP.NET Core app as a Windows Service without using IIS. This sample demonstrates the scenario described in [Host ASP.NET Core in a Windows Service](https://docs.microsoft.com/aspnet/core/host-and-deploy/windows-service).
+
+After establishing and starting the service, access the app in a browser at `http://localhost:5000/`.
```
Tratcher: #14193 asked to show how to configure a port, not about the default. The default exists for dev purposes, published apps should always configure a url/port.

guardrex: I *almost* read it that way, but he shows "localhost" in the ask and ...

> so the service which is running can be navigated to through a browser.

... implies that he had exactly that problem ... couldn't get it running on the default and seemed to be wondering if it would run at all without an explicit config. He didn't elaborate, and I may have guessed wrong.

WRT an enhancement to the topic for this and since endpoint config is a Kestrel thing, shouldn't we just cross-link to the Kestrel topic endpoint config section and explicitly recommend setting a URL/port as opposed to showing an opinionated example here? I'd like to avoid content duplication and just get the reader to the right spot to see all of the knobs.

Tratcher: Showing even the simplified ASPNETCORE_URLS example would be a good start. And yes linking to Kestrel and HttpSys for advanced config would help too.

guardrex:
@Tratcher "Showing" ... don't we just tell the reader to 'set it' ... there isn't much to show unless you want to show the console command, which isn't the only way to go.

We had a section on HTTPS that was headed in the right direction. I've tried to enhance that, including the links.

aspnetcore/host-and-deploy/windows-service/samples/2.x/AspNetCoreService/README.md
taori: @Tratcher guessed my intentions correctly. Changing localhost to an IP made it available publicly too, perhaps a small sample + reference for kestrel configuration is the safest way here. then again this was already merged so  i guess it is done anyway.

guardrex: @taori ... access the Kestrel endpoint config content. There are **_many knobs_** for those options.

taori: @guardrex Yes i know. But it's a guide to make a standalone service. It only makes sense to reference the contents of most common concerns a developer might have when reading that article, which in this case among other things is "where do i configure my ip so my service can be talked to". As you can see i did find the info myself after some research - just trying to ease the use the documentation for the next developer really.

PR #14215, opened 2019-09-08
Blazor chained bind

### LINKED ISSUE #14144: Add sample showing "chained bind"
This is a small code sample that shows a solution to cases like this: https://github.com/aspnet/AspNetCore/issues/13408

The typical use case for this is when you want to write a component that represents an input element, but adds some additional functionality like styling, or processing the values that come out of it.

We call this *chained-bind* because there are multiple levels of `@bind` happening:
- Your component binds to an `<input>` element
- Another component binds to your component

Here's a working sample:

```razor
@* Index.razor *@
<PasswordField bind-Value="@password" />
@code {
    string password;
}
```

```razor
@* PasswordField.razor
<input
       type="@(ShowPassword ? "text" : "password")"
       required
       value="@Password"
       @oninput="@OnPasswordChanged" />
<a @onclick="ToggleShowPassword">
    Show password
</a>

@code {
    private string password;

    [Parameter]
    public string Password
    {
        get
        {
            return this.password ?? "";
        }
        set
        {
            if (this.password != value)
            {
                this.password = value;
            }
        }
    }

    [Parameter]
    public EventCallback<string> PasswordChanged { get; set; }

    bool ShowPassword { get; set; }

    private void ToggleShowPassword()
    {
        this.ShowPassword = !this.ShowPassword;
    }

    private Task OnPasswordChanged(ChangeEventArgs e)
    {
        Password = (string)e.Value;
        return PasswordChanged.InvokeAsync(Password);
    }
}
```

guardrex:
Fixes #14144

[Internal Review Topic (direct link to the new *Chained bind* section)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/components?view=aspnetcore-3.0&branch=pr-en-us-14215#chained-bind)

The *chained bind concept* and *checking values with the property's setter concept* seem to me to be *two different things*. One doesn't need to perform any checking if they're just binding to an element and binding across nested components. Therefore, what I do here is cover chained bind first. Get that basic idea out of the way. Then, I expand upon chained bind with a check for spaces in the password value in a second example. This 🤠 *Texas Two-step* 🤠 makes sense to me because it avoids complexity when explaining the basic concept while not losing the great example of checking values simultaneously in the child component.

I made a few (I hope simplifying) changes. Does anything fall into the *RexHacks*:tm: 🤦‍♂ category?

@epsitec, do you want to :eyes: this and comment on it? Also, I was curious about what sort of check(s) you performed on your password?

Note-to-self after chatting with @Legends: When this makes its way into the sample app (later), we might want to put a button in the parent that changes the value of the `password` field (just use a hardcoded value in a method) to show the two-way binding aspect. It could even be as simple as a **Clear password** button that does a `password = string.Empty()` in the parent.

### DIFF
--- a/aspnetcore/blazor/components.md
+++ b/aspnetcore/blazor/components.md
@@ -511,6 +511,129 @@ Use `EventCallback` and `EventCallback<T>` for event handling and binding compon
 
 Prefer the strongly typed `EventCallback<T>` over `EventCallback`. `EventCallback<T>` provides better error feedback to users of the component. Similar to other UI event handlers, specifying the event parameter is optional. Use `EventCallback` when there's no value passed to the callback.
 
+## Chained bind
+
+A common scenario when creating nested components is the desire to bind a property:
+
+* To a page element in a component.
+* Across nested components.
+
+This scenario is called a *chained bind* because multiple levels of binding occur simultaneously.
+
+The following `PasswordField` component (*PasswordField.razor*):
+
+* Binds an `<input>` element to a `Password` property.
+* Exposes the `Password` property to a parent component with an [EventCallback](#eventcallback).
+
+```cshtml
+Password: 
+
+<input @oninput="OnPasswordChanged" 
+       required 
+       type="@(showPassword ? "text" : "password")" 
+       value="@Password" />
+
+<button class="btn btn-primary" @onclick="ToggleShowPassword">
+    Show password
+</button>
+
+@code {
+    private bool showPassword;
+
+    [Parameter]
+    public string Password { get; set; }
+
+    [Parameter]
+    public EventCallback<string> PasswordChanged { get; set; }
+
+    private Task OnPasswordChanged(ChangeEventArgs e)
+    {
+        Password = e.Value.ToString();
+
+        return PasswordChanged.InvokeAsync(Password);
+    }
+
+    private void ToggleShowPassword()
+    {
+        showPassword = !showPassword;
+    }
+}
+```
+
+The `PasswordField` component is used in another component:
+
+```cshtml
+<PasswordField @bind-Password="password" />
+
+@code {
+    private string password;
+}
+```
+
+If you need to perform checks or trap errors on the password in the preceding example:
+
+* Create a backing field for `Password` (`password` in the following example code).
+* Perform the checks or trap errors in the `Password` setter.
+
+The following example instantly provides feedback to the user if a space is provided anywhere in the password's value:
+
+```cshtml
+Password: 
+
+<input @oninput="OnPasswordChanged" 
+       required 
+       type="@(showPassword ? "text" : "password")" 
+       value="@Password" />
+
+<button class="btn btn-primary" @onclick="ToggleShowPassword">
+    Show password
+</button>
+
+<span class="text-danger">@validationMessage</span>
+
+@code {
+    private bool showPassword;
+    private string password;
+    private string validationMessage;
+
+    [Parameter]
+    public string Password
+    {
+        get { return password ?? string.Empty; }
+        set
+        {
+            if (password != value)
+            {
+                if (value.Contains(' '))
+                {
+                    validationMessage = "Spaces not allowed!";
+                }
+                else
+                {
+                    password = value;
+                    validationMessage = string.Empty;
+                }
+            }
+        }
+    }
+
+    [Parameter]
+    public EventCallback<string> PasswordChanged { get; set; }
+
+    private Task OnPasswordChanged(ChangeEventArgs e)
+    {
+        Password = e.Value.ToString();
+
+        return PasswordChanged.InvokeAsync(Password);
+    }
+
+    private void ToggleShowPassword()
+    {
+        showPassword = !showPassword;
+    }
+}
+```
+
 ## Capture references to components
 
 Component references provide a way to reference a component instance so that you can issue commands to that instance, such as `Show` or `Reset`. To capture a component reference:
--- a/aspnetcore/blazor/components.md
+++ b/aspnetcore/blazor/components.md
@@ -513,17 +513,14 @@ Prefer the strongly typed `EventCallback<T>` over `EventCallback`. `EventCallbac
 
 ## Chained bind
 
-A common scenario when creating nested components is the desire to bind a property:
+A common scenario is chaining a data-bound parameter to a page element in the component's output. This scenario is called a *chained bind* because multiple levels of binding occur simultaneously.
 
-* To a page element in a component.
-* Across nested components.
-
-This scenario is called a *chained bind* because multiple levels of binding occur simultaneously.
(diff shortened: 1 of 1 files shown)


epsitec:
Your split example makes complete sense and would have been really helpful at the time when I was writing the code and upgrading through the various previews.

Originally, I had a custom setter, because I was raising _changed_ events from there. I did not implement any form of validation in the password field, but that would be a good scenario, in case the password field has to implement its own intelligence for that.

By the way, do you have any guidance as to how you recommend that we implement form validation with Blazor?

guardrex:
> implement form validation with Blazor

https://docs.microsoft.com/aspnet/core/blazor/forms-validation

aspnetcore/blazor/components.md
```diff
@@ -511,6 +511,129 @@ Use `EventCallback` and `EventCallback<T>` for event handling and binding compon
 
 Prefer the strongly typed `EventCallback<T>` over `EventCallback`. `EventCallback<T>` provides better error feedback to users of the component. Similar to other UI event handlers, specifying the event parameter is optional. Use `EventCallback` when there's no value passed to the callback.
 
+## Chained bind
+
+A common scenario when creating nested components is the desire to bind a property:
+
+* To a page element in a component.
+* Across nested components.
```
danroth27: I'm not really sure what "nested components" means here.

guardrex: Components that use other components ... i.e., a graph of two or more components.

This is trying to say that a component's property is bound to one of its elements and to a field/property of a parent component.

aspnetcore/blazor/components.md
```diff
@@ -511,6 +511,129 @@ Use `EventCallback` and `EventCallback<T>` for event handling and binding compon
...
+
+This scenario is called a *chained bind* because multiple levels of binding occur simultaneously.
+
+The following `PasswordField` component (*PasswordField.razor*):
+
+* Binds an `<input>` element to a `Password` property.
```
danroth27: I think this merits some explanatory text to point out that you don't use the standard `@bind` syntax in this case. You need to instead go level lower to specify the event handler and value yourself.

guardrex: Can it just say 'sets the value' then? That would remove the word "bind" from it.

aspnetcore/blazor/components.md
danroth27: How about something like this?: "A common scenario when authoring components is to support chaining a data bound parameter to a page element in the component's output."

aspnetcore/blazor/components.md
guardrex: > * Sets the value of an `<input>` element to a `Password` property.

danroth27: It both sets the value and listens to the desired event.

guardrex: Is the "listens" part the 2nd bullet tho?

danroth27: The important thing is to say that you can't use `@bind` when implementing a chained bind.

guardrex: Ok.

Since this two-bullet sequence is only trying to describe what the code in the following example does, I'll try adding it a bit earlier ... something like ...

> Chained bind can't be implemented with `@bind` syntax. Chained bind is created using the property with an event handler.

danroth27: Yeah, "Chained bind can't be implemented with `@bind` syntax *on the page element*...". The consume of the component can of course use `@bind` with the parameter.

aspnetcore/blazor/components.md
```diff
@@ -511,6 +511,126 @@ Use `EventCallback` and `EventCallback<T>` for event handling and binding compon
...
+
+```cshtml
+<PasswordField @bind-Password="password" />
+
+@code {
+    private string password;
+}
+```
+
+If you need to perform checks or trap errors on the password in the preceding example:
```
Rick-Anderson: To perform checks or trap errors on the password in the preceding example:

aspnetcore/blazor/components.md
```diff
@@ -511,6 +511,126 @@ Use `EventCallback` and `EventCallback<T>` for event handling and binding compon
...
+
+* Create a backing field for `Password` (`password` in the following example code).
+* Perform the checks or trap errors in the `Password` setter.
+
+The following example instantly provides feedback to the user if a space is provided anywhere in the password's value:
```
Rick-Anderson: The following example provides  immediate feedback to the user if a space is used in the password's value:

aspnetcore/blazor/components.md
```diff
@@ -511,6 +511,126 @@ Use `EventCallback` and `EventCallback<T>` for event handling and binding compon
...
+A common scenario is chaining a data-bound parameter to a page element in the component's output. This scenario is called a *chained bind* because multiple levels of binding occur simultaneously.
+
+Chained bind can't be implemented with `@bind` syntax in the page's element. Chained bind is created using the property with an event handler. A parent component can use `@bind` syntax with the parameter.
+
+The following `PasswordField` component (*PasswordField.razor*):
+
+* Sets an `<input>` element's value to a `Password` property.
+* Exposes changes of the `Password` property to a parent component with an [EventCallback](#eventcallback).
+
+```cshtml
```
Rick-Anderson: Hopefully this will eventually be migrated to code file.
i can do that post GA

aspnetcore/blazor/components.md
```diff
@@ -511,6 +511,126 @@ Use `EventCallback` and `EventCallback<T>` for event handling and binding compon
 
 Prefer the strongly typed `EventCallback<T>` over `EventCallback`. `EventCallback<T>` provides better error feedback to users of the component. Similar to other UI event handlers, specifying the event parameter is optional. Use `EventCallback` when there's no value passed to the callback.
 
+## Chained bind
+
+A common scenario is chaining a data-bound parameter to a page element in the component's output. This scenario is called a *chained bind* because multiple levels of binding occur simultaneously.
+
+Chained bind can't be implemented with `@bind` syntax in the page's element. Chained bind is created using the property with an event handler. A parent component can use `@bind` syntax with the parameter.
```
```suggestion
A chained bind can't be implemented with `@bind` syntax in the page's element. The event handler and value must be specified separately. A parent component, however, can use `@bind` syntax with the component's parameter.
```

aspnetcore/blazor/components.md
guardrex: I added a bullet point for this to the existing issue WRT what the sample will cover and what it won't. There was no point to doing the extra work for snippets at this time with so much churn and without engineering identifying which examples they want in the sample app and which ones they just want in-text. I used snippets when I needed highlighting. Otherwise, it wasn't worth the cost during preview. 🏃

https://github.com/aspnet/AspNetCore.Docs/issues/14216

PR #14264, opened 2019-09-11
Use RedirectToPage after SignOutAsync as per current Identity UI.

### LINKED ISSUE #14255: Redirection after SignOutAsync is required, not to be avoided
> Don't redirect after calling `SignOutAsync` or the user will **not** be signed out.

The suggestion here is incorrect. It shouldn't be prefixed with "don't", as redirection is what actually forces the user to be signed out. The code above this line for `LogoutModel` should also be changed to end with `return RedirectToPage()` instead of `return Page()` to match, which is what the current [2.2 version](https://github.com/aspnet/AspNetCore/blob/release/2.2/src/Identity/UI/src/Areas/Identity/Pages/V4/Account/Logout.cshtml.cs#L47) of the Identity UI and scaffolding produce.

---
#### Document Details

* ID: 94bfd895-d2af-390b-0801-428b36a27ff4
* Version Independent ID: 0e0d870e-e49a-e16c-c259-68d58e701b8f
* Content: [Introduction to Identity on ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/identity?view=aspnetcore-2.2&tabs=visual-studio)
* Content Source: [aspnetcore/security/authentication/identity.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/security/authentication/identity.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

serpent5:
Fixes #14255.

The comment that comes in with the code update explains the need for a redirect, so I elected to remove the outdated line from the docs page.

### DIFF
--- a/aspnetcore/security/authentication/identity.md
+++ b/aspnetcore/security/authentication/identity.md
@@ -207,7 +207,7 @@ The **Log out** link invokes the `LogoutModel.OnPost` action.
 
 [!code-csharp[](identity/sample/WebApp1/Areas/Identity/Pages/Account/Logout.cshtml.cs)]
 
-[SignOutAsync](/dotnet/api/microsoft.aspnetcore.identity.signinmanager-1.signoutasync#Microsoft_AspNetCore_Identity_SignInManager_1_SignOutAsync) clears the user's claims stored in a cookie. Don't redirect after calling `SignOutAsync` or the user will **not** be signed out.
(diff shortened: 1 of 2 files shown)


Rick-Anderson:
@HaoK  I'm confident this PR is correct. It's a 30 second review when you have time. I'm merging for now, will fix later if we're wrong.

PR #14287, opened 2019-09-12
Unparsable values when databinding

### LINKED ISSUE #13014: Add documentation for parsing during binding
Currently, the [docs for Blazor's `@bind`](https://docs.microsoft.com/en-us/aspnet/core/blazor/components?view=aspnetcore-3.0#data-binding) don't mention anything about what we do when you bind to an `int`, `decimal`, `DateTime`, or similar and the user enters an unparseable value.

The **old** behavior was that we would ignore the user's input until they type something valid, at which point we'd apply it to the bound field/property.

The **new** behavior is that we automatically *revert* any invalid inputs. For example, if you have a textbox bound to an `int` and the current value is `123`, then the user edits it to `123.45` and tabs out, we will now reset the textbox to `123`. This has the benefit that the user can now see their input wasn't accepted, and they can see what value we are actually holding internally.

### Customizing how and when parsing rules are enforced

This logic applies when the corresponding bind event occurs. By default bind applies on `onchange`, but you can use `@bind-value:event` or similar to set a different event. So for example if you use `@bind-value:event="oninput"`, then the reversion would occur after any keystroke that introduced an unparseable value. Usually that's a good thing (in the example previously, it would stop the user from typing a `.` character by immediately removing it, so they would be forced to understand that only whole numbers may be entered). However there are some cases where it might not be what you want, such as if you want users to be able to clear the textbox completely. In that case we'd immediately reset it to `0`, which could confuse someone who was trying to delete the `0` to enter some other digit. If this behavior doesn't match what you want then alternatives include:

 * Don't use `@bind-value:event="oninput"`, but rather use the default event which is `onchange`. That way the user can type invalid values and we won't interfere until they tab out.
 * Or, instead of binding to an `int`, bind to something looser such as a `int?` or a `string`, and put in your own custom logic to decide what to do with invalid entries.
 * Or, instead of using `<input>`, use `<InputNumber>`/`<InputDate>`/etc., as these have more powerful built-in behavior around invalid inputs. They allow the user to type in bad data and register validation errors on the associated `EditContext` so you can display such errors to the user without also interfering with typing.

guardrex:
Fixes #13014

### DIFF
--- a/aspnetcore/blazor/components.md
+++ b/aspnetcore/blazor/components.md
@@ -224,6 +224,26 @@ In addition to handling `onchange` events with `@bind` syntax, a property or fie
 
 Unlike `onchange`, which fires when the element loses focus, `oninput` fires when the value of the text box changes.
 
+**Unparsable values**
+
+When a user provides an unparsable value to a databound element, the value is automatically reverted by default for the `onchange` event (`@bind="{PROPERTY OR FIELD}"`).
+
+Consider the following scenario:
+
+* An `<input>` element is bound to an `int` type.
+* The current value of the element is `123`.
+* The user updates the value of the element to `123.45` and changes the element focus.
+
+In the preceding scenario, the element's value is reverted to `123`. When the value `123.45` is rejected in favor of the original value of `123`, the user understands that their value wasn't accepted.
+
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/components.md
```diff
@@ -224,6 +224,34 @@ In addition to handling `onchange` events with `@bind` syntax, a property or fie
 
 Unlike `onchange`, which fires when the element loses focus, `oninput` fires when the value of the text box changes.
 
+**Unparsable values**
+
+When a user provides an unparsable value to a databound element, the unparsable value is automatically reverted by default for the `onchange` event (`@bind="{PROPERTY OR FIELD}"`).
```
```suggestion
When a user provides an unparsable value to a databound element, the unparsable value is automatically reverted to its previous value when the bind event is triggered.
```

guardrex:
> It's growing very organically right now

Yes, indeed. It's almost a **_26 Volume Encyclopedia_**. :smile:

PR #14250, opened 2019-09-11
What's new in 3.0

Rick-Anderson:
Fixes #14240

[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-3.0?view=aspnetcore-3.0&branch=pr-en-us-14250)

Requires sign-off by the following:

- [x]  @bradygaster  SignalR 
- [x]  @anurse 
- [ ]  @glennc 
- [ ]  @shirhatti 
- [x]  @blowdart 
- [x] @mkArtakMSFT 
- [ ]  ~@Pilchie~
- [ ]  @danroth27

Issue #14240, opened 2019-09-10
No "What's new in 3.0" page


fiksen99:
3.0 is in preview, but there's no page indicating what's new to help me decide if I want to upgrade. I think the feature set is now locked for 3.0 and the changes are now focussed on quality, correct? If so, this means we can begin adding documentation for the 3.0 features.

---
#### Document details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: dd495160-090c-065e-5b98-53b112db0cea
* Version Independent ID: 46699e55-2685-95f8-4b54-dcb66d630000
* Content: [What's new in ASP.NET Core 2.2](https://docs.microsoft.com/en-gb/aspnet/core/release-notes/aspnetcore-2.2?view=aspnetcore-3.0)
* Content Source: [aspnetcore/release-notes/aspnetcore-2.2.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/release-notes/aspnetcore-2.2.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-releasenotes**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
@danroth27 @scottaddie  I'll start "What's new in 3.0" from https://devblogs.microsoft.com/aspnet/ 
cc @guardrex

Rick-Anderson:
- gRPC
-

PR #14369, opened 2019-09-17
added core 3.0 signalr

bradygaster:
included all the stuff for signalr 3core 3.0. cc @BrennanConroy @anurse @halter73 @RickAndMSFT

### DIFF
--- a/aspnetcore/release-notes/aspnetcore-3.0.md
+++ b/aspnetcore/release-notes/aspnetcore-3.0.md
@@ -88,7 +88,7 @@ Options for certificate authentication include the ability to:
 
 A default user principal is constructed from the certificate properties. The user principal contains an event that enables supplementing or replacing the principal. For more information, see <xref:security/authentication/certauth>.
 
-[Windows Authentication](/windows-server/security/windows-authentication/windows-authentication-overview) has been extended onto Linux and macOS. In previous versions, Windows Authentication authentication was limited to [IIS](xref:host-and-deploy/iis/index) and [HttpSys](xref:fundamentals/servers/httpsys). In ASP.NET Core 3.0, [Kestrel](xref:fundamentals/servers/kestrel) has the ability to use Negotiate, [Kerberos](/windows-server/security/kerberos/kerberos-authentication-overview), and [NTLM on Windows](/windows-server/security/kerberos/ntlm-overview), Linux, and macOS for Windows domain joined hosts. Kestrel support of these authentication schemes is provide by the [Microsoft.AspNetCore.Authentication.Negotiate nuget](https://www.nuget.org/packages/Microsoft.AspNetCore.Authentication.Negotiate) package. As with the other authentication services, configure authentication app wide, then configure the service:
+[Windows Authentication](/windows-server/security/windows-authentication/windows-authentication-overview) has been extended onto Linux and macOS. In previous versions, Windows Authentication authentication was limited to [IIS](xref:host-and-deploy/iis/index) and [HttpSys](xref:fundamentals/servers/httpsys). In ASP.NET Core 3.0, [Kestrel](xref:fundamentals/servers/kestrel) has the ability to use Negotiate, [Kerberos](/windows-server/security/kerberos/kerberos-authentication-overview), and [NTLM on Windows](/windows-server/security/kerberos/ntlm-overview), Linux, and macOS for Windows domain joined hosts. Kestrel support of these authentication schemes is provided by the [Microsoft.AspNetCore.Authentication.Negotiate nuget](https://www.nuget.org/packages/Microsoft.AspNetCore.Authentication.Negotiate) package. As with the other authentication services, configure authentication app wide, then configure the service:
 
 ```csharp
 public void ConfigureServices(IServiceCollection services)
@@ -134,9 +134,9 @@ For more information, see <xref:blazor/index>.
 
 ## SignalR
 
-The SignalR JavaScript client has changed from being `@aspnet/signalr` to `@microsoft/signalr`. To react to this change, you will need to change your references in `package.json` files, require statements, and ECMAScript import statements. 
+See [Update SignalR code](xref:migration/22-to-30#update-signalr-code) for migration instructions.
 
-In the JavaScript and .NET Clients for SignalR, we've added support for automatic reconnection via the `withAutomaticReconnect()` method on the `HubConnectionBuilder`. By default, the client will try to reconnect immediately and after 2, 10, and 30 seconds. Enlisting in automatic reconnect is opt-in, but simple via this new method.
+In the JavaScript and .NET Clients for SignalR, support was added for automatic reconnection. By default, the client tries to reconnect immediately and retry after 2, 10, and 30 seconds if necessary. If the client successfully reconnects, it receives a new connection ID. Automatic reconnect is opt-in:
 
 ```javascript
 const connection = new signalR.HubConnectionBuilder()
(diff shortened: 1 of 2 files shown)


aspnetcore/release-notes/aspnetcore-3.0.md
```diff
@@ -134,7 +134,184 @@ For more information, see <xref:blazor/index>.
 
 ## SignalR
 
-@bradygaster  to provide
+The SignalR JavaScript client has changed from being `@aspnet/signalr` to `@microsoft/signalr`. To react to this change, you will need to change your references in `package.json` files, require statements, and ECMAScript import statements. 
+
+In the JavaScript and .NET Clients for SignalR, we've added support for automatic reconnection via the `withAutomaticReconnect()` method on the `HubConnectionBuilder`. By default, the client will try to reconnect immediately and after 2, 10, and 30 seconds. Enlisting in automatic reconnect is opt-in, but simple via this new method.
```
halter73: It might be nice to call out that this will give you a new ConnectionId which is different from the OG SignalR clients. Maybe something like the following would be good.

```suggestion
In the JavaScript and .NET Clients for SignalR, we've added support for automatic reconnection via the `withAutomaticReconnect()` method on the `HubConnectionBuilder`. By default, the client will try to reconnect immediately and retry after 2, 10, and 30 seconds if necessary. If the client successfully reconnects, it will be given a new connection ID. Enlisting in automatic reconnect is opt-in, but simple via this new method.
```

Rick-Anderson: @halter73 I'll migrate your suggestion over to #14371

aspnetcore/release-notes/aspnetcore-3.0.md
```diff
@@ -134,7 +134,184 @@ For more information, see <xref:blazor/index>.
...
+    IAuthorizationRequirement
+{
+    protected override Task HandleRequirementAsync(AuthorizationHandlerContext context,
+        DomainRestrictedRequirement requirement,
+        HubInvocationContext resource)
+    {
+        if (IsUserAllowedToDoThis(resource.HubMethodName, context.User.Identity.Name) &&
+            context.User != null &&
+            context.User.Identity != null &&
+            context.User.Identity.Name.EndsWith("@jabbr.net", StringComparison.OrdinalIgnoreCase))
```
halter73: All this null checking seems unnecessary since `context.User.Identity.Name` is already passed to IsUserAllowedToDoThis above. IsUserAllowedToDoThis should verify the name ends with "@jabbr.net" regardless.

BTW

```c#
            context.User != null &&
            context.User.Identity != null &&
            context.User.Identity.Name.EndsWith("@jabbr.net", StringComparison.OrdinalIgnoreCase)
```

and 

```c#
            context?.User?.Identity?.Name?.EndsWith("@jabbr.net", StringComparison.OrdinalIgnoreCase) == true
```

are equivalent expressions, but the latter is probably a little too cute for a doc sample.

halter73: The best approach is probably to add exit early if no user name is set. Ex:

```
if (context.User?.Identity?.Name == null)
{
    return Task.CompletedTask;
}
```

Then you can remove the other null checking.

Rick-Anderson: @halter73 I'll stop working on #14371 if you can move over to #14371. That way I can accept your suggestion commits without copy/paste.

aspnetcore/release-notes/aspnetcore-3.0.md
```diff
@@ -134,7 +134,184 @@ For more information, see <xref:blazor/index>.
...
+    private bool IsUserAllowedToDoThis(string hubMethodName,
+        string currentUsername)
+    {
+        return !(currentUsername.Equals("bob42@jabbr.net", StringComparison.OrdinalIgnoreCase) &&
+            hubMethodName.Equals("banUser", StringComparison.OrdinalIgnoreCase));
+    }
+}
+```
+
+Now, individual Hub methods can be decorated with the name of the policy the code will need to check at run-time. As clients attempt to call individual Hub methods, the `DomainRestrictedRequirement` handler will run and control access to the methods. Based on the way the `DomainRestrictedRequirement` controls access, all logged-in users should be able to call the `SendMessage` method, only users who’ve logged in with a `@jabbr.net` email address will be able to view users’ histories, and – with the exception of `bob42@jabbr.net` – will be able to ban users from the chat room.
```
halter73: I think only having one user who cannot ban other users is unusual/unexpected making this sample somewhat harder to follow.

aspnetcore/release-notes/aspnetcore-3.0.md
```diff
@@ -134,7 +134,184 @@ For more information, see <xref:blazor/index>.
...
+{
+    // read from and process stream items
+    while (await streamContent.WaitToReadAsync(Context.ConnectionAborted))
+    {
+        while (streamContent.TryRead(out var content))
+        {
+            // process content
+        }
+    }
+}
```
halter73: I would use the new IAsyncEnumerable sample here, since it's simpler than the ChannelReader version:

https://github.com/aspnet/AspNetCore.Docs/blob/4aa859cb075f3b1b250aeccfcbcabd5e810b974e/aspnetcore/signalr/streaming.md#L88-L106

Rick-Anderson: @halter73  can you make that change in a commit on #14371 ?

aspnetcore/release-notes/aspnetcore-3.0.md
```diff
@@ -134,7 +134,184 @@ For more information, see <xref:blazor/index>.
...
+```
+
+Clients would use the SignalR `Subject` (or an RxJS Subject) as an argument to the `streamContent` parameter of the Hub method above.
```
halter73: Should we mention that with the C# client you can just pass an `IAsyncEnumerable` or `ChannerReader` as an argument?

Rick-Anderson:
@halter73 if any of these comments didn't make it into #14371  you can make commits directly against #14250

aspnetcore/release-notes/aspnetcore-3.0.md
```diff
@@ -134,7 +134,229 @@ For more information, see <xref:blazor/index>.
...
+app.UseRouting(routes =>
+{
+    routes.MapRazorPages();
+    routes.MapHub<ChatHub>("hubs/chat");
+});
+```
+
+ASP.NET Core 3.0 SignalR added:
+
+* Streaming support, which enables streaming return values from server-side methods. This is useful for when fragments of data will come in over a period of time.
```
BrennanConroy: This is not new in 3.0...

Rick-Anderson: > This is not new in 3.0...

@BrennanConroy  I'll remove that from #14240. This has been merged into #14240 so leave comments in #14240.  I'll remove the following from `ASP.NET Core 3.0 SignalR added:` to
`## Generic Host`

That is, the remaining SignalR content.

bradygaster: ack. i globbed server-to-client in there by accident. @Rick-Anderson - you got this change or need me to hop on it?

aspnetcore/release-notes/aspnetcore-3.0.md
bradygaster: yes - can you make this or want @Rick-Anderson or i to wordsmith it?

halter73: I already did this in https://github.com/aspnet/AspNetCore.Docs/pull/14371

PR #14370, opened 2019-09-17
Update 22-to-30.md

Rick-Anderson:
An edit from @bradygaster PR  #14369


Rick-Anderson:
@halter73  if this is not OK I'll revert.

PR #14401, opened 2019-09-19
SignalR Json instructions

Rick-Anderson:
@anurse your comments belong in the migration guide, not what's new. I did put a pointer to these instructions in the what's new. You'll see that soon :)


BrennanConroy:
https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/migration/22-to-30.md#switch-to-newtonsoftjson 😕

anurse:
I still think we should include a link to this part of the guide in the SignalR section of the what's new doc. It **is** what's new, and understanding it is pretty critical to user success.

anurse:
I agree that a detailed description isn't needed in the what's new doc though.

anurse:
I see we do have some linking content. Never mind then! Thanks!

PR #14425, opened 2019-09-19
Add ANCM updates

shirhatti:


Rick-Anderson:
@shirhatti check my commit to see what I changed.
Thanks for doing this.
@guardrex is this documented? If so, provide me a link.

guardrex:
@Rick-Anderson Our docs always said to use the Windows Event Log. There are two sections in the troubleshooting topic that address it, one section each for Azure and IIS. If you want a cross-link, you could use the whole topic `<xref:test/troubleshoot-azure-iis>`; however, the output is just enhanced, so there's nothing specific to link to.

PR #14483, opened 2019-09-23
Update index.md

Rick-Anderson:


wadepickett:
Thanks for the fix.  Just saw this Monday morning.

wadepickett:
I will use the include going forward.  I didn't realize we had one we were to use for downloads for 3.0.  Thanks.

PR #14487, opened 2019-09-23
Health Checks ReadinessPublisher update

### LINKED ISSUE #14459: How to understand the ReadinessPublisher sample
From @ThomasArdal on https://github.com/aspnet/AspNetCore.Docs/issues/14458 ...

I'm trying to understand the `ReadinessPublisher` class in the health checks 2.x and 3.x samples. I have two main questions for things that don't make sense. At least not in my head.

### Why is `Entries` there?

There's a list of health results named `Entries`. The publisher keeps pushing to this list:

```csharp
Entries.Add((report, cancellationToken));
```

But the results are new read or logged by anyone. What is the purpose of this list?

### Who sets `Exception`?

There's a property named `Exception` on the publisher. No-one ever sets this property to any value. Is that done automatically somehow? The publisher checks if the exception is not null and then throw it:

```csharp
if (Exception != null)
{
    throw Exception;
}
```

I cannot see when this if will ever be true?

---
#### Document Details

* ID: c0cbb094-29e9-740a-9fd6-db372641d753
* Version Independent ID: 52d8efc9-a336-71d9-6d90-98173cbf6bf4
* Content: [Health checks in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-2.2)
* Content Source: [aspnetcore/host-and-deploy/health-checks.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/host-and-deploy/health-checks.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #14459

Ignore the build warning ... that link was fixed on another PR.

Thanks @ThomasArdal

### DIFF
--- a/aspnetcore/host-and-deploy/health-checks/samples/2.x/HealthChecksSample/ReadinessPublisher.cs
+++ b/aspnetcore/host-and-deploy/health-checks/samples/2.x/HealthChecksSample/ReadinessPublisher.cs
@@ -26,12 +26,12 @@ namespace SampleApp
         {
             if (report.Status == HealthStatus.Healthy)
             {
-                _logger.LogInformation("{timestamp} Readiness Probe Status: {result}", 
+                _logger.LogInformation("{Timestamp} Readiness Probe Status: {Result}", 
                     DateTime.UtcNow, report.Status);
             }
             else
             {
-                _logger.LogError("{timestamp} Readiness Probe Status: {result}", 
+                _logger.LogError("{Timestamp} Readiness Probe Status: {Result}", 
                     DateTime.UtcNow, report.Status);
             }
 
--- a/aspnetcore/host-and-deploy/health-checks/samples/2.x/HealthChecksSample/ReadinessPublisher.cs
+++ b/aspnetcore/host-and-deploy/health-checks/samples/2.x/HealthChecksSample/ReadinessPublisher.cs
@@ -21,12 +21,12 @@ namespace SampleApp
         {
             if (report.Status == HealthStatus.Healthy)
             {
-                _logger.LogInformation("{TIMESTAMP} Readiness Probe Status: {RESULT}", 
+                _logger.LogInformation("{timestamp} Readiness Probe Status: {result}", 
                     DateTime.UtcNow, report.Status);
             }
             else
             {
-                _logger.LogError("{TIMESTAMP} Readiness Probe Status: {RESULT}", 
+                _logger.LogError("{timestamp} Readiness Probe Status: {result}", 
                     DateTime.UtcNow, report.Status);
             }
 
--- a/aspnetcore/host-and-deploy/health-checks/samples/2.x/HealthChecksSample/ReadinessPublisher.cs
+++ b/aspnetcore/host-and-deploy/health-checks/samples/2.x/HealthChecksSample/ReadinessPublisher.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Diagnostics.HealthChecks;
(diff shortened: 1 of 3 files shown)


aspnetcore/host-and-deploy/health-checks/samples/2.x/HealthChecksSample/ReadinessPublisher.cs
```diff
@@ -17,24 +16,18 @@ public ReadinessPublisher(ILogger<ReadinessPublisher> logger)
...
         {
-            Entries.Add((report, cancellationToken));
-
-            _logger.LogInformation("{TIMESTAMP} Readiness Probe Status: {RESULT}", 
-                DateTime.UtcNow, report.Status);
-
-            if (Exception != null)
+            if (report.Status == HealthStatus.Healthy)
+            {
+                _logger.LogInformation("{TIMESTAMP} Readiness Probe Status: {RESULT}",
```
rynowak: I'm really not a fan of this kind of sample... **because** we already have logging for health check results. This isn't something a user should ever have to write.

guardrex: Then perhaps let's just have it ... as you say ... do nothing here. We can just show the skeleton of it. I'll try it on the next commit so that we can see how it looks.

guardrex: 🤔 ..... idk about this ...... How about this ... we can add this line right into the sample code for the `ReadinessPublisher`. We already say this in the `LivenessProbeStartup` code where this publisher is registered ...

> The readiness check might do a set of more expensive or time-consuming checks to determine if all other resources are responding.

guardrex:
I added this above the method ...

```csharp
// The following example is for demonstration purposes only. Health Checks 
// Middleware already logs health checks results. A real-world readiness 
// check in a production app might perform a set of more expensive or 
// time-consuming checks to determine if other resources are responding 
// properly.
public Task PublishAsync(HealthReport report, 
    CancellationToken cancellationToken)
{
    ...
```

aspnetcore/host-and-deploy/health-checks/samples/2.x/HealthChecksSample/ReadinessPublisher.cs
```diff
@@ -17,24 +16,18 @@ public ReadinessPublisher(ILogger<ReadinessPublisher> logger)
...
-            if (Exception != null)
+            if (report.Status == HealthStatus.Healthy)
+            {
+                _logger.LogInformation("{TIMESTAMP} Readiness Probe Status: {RESULT}", 
+                    DateTime.UtcNow, report.Status);
+            }
+            else
             {
-                throw Exception;
+                _logger.LogError("{TIMESTAMP} Readiness Probe Status: {RESULT}",
```
rynowak: we don't normally show screaming caps for log value names - I don't want users to get the impression this is a convention we are recommending.

guardrex: We were ... perhaps in ancient times around here ... capitalizing placeholders in strings. It was asked/discussed back when it was started. I'll research it and report back later. You want this as `{timestamp}`?

guardrex: btw - this will be easy to fix everywhere. I've already made a note of it. I can RegEx the repo and change all of them on one PR. I'll open an issue and square that away at a later date.

guardrex: ok ... I **_didn't_** remember the convo correctly. It was Pascal case when talking to `anurse` about it ...

https://github.com/aspnet/AspNetCore.Docs/pull/4697#discussion_r148855863

I'll fix that up on the next commit. I'll address this everywhere in docs that it might exist on a new issue and another PR later.

PR #14499, opened 2019-09-23
Add Blazor samples link to ToC

scottaddie:


guardrex:
... and btw ... there is only *one* Blazor sample. The tutorial is made up of snapshots. I *greatly preferred* it that way, and I think that has paid off during preview ... vastly easier on us (*me* :smile:).

scottaddie:
Let's stick with the engineering samples link for now.

guardrex:
Very good ..... the common sample is linked directly in the topics. I'm sure that the devs/readers will see it.

PR #14535, opened 2019-09-24
Update 'SDK' to 'Sdk' in project file examples

guardrex:
Per templates, fact that Pascal case has always been used, and guidance in https://docs.microsoft.com/visualstudio/msbuild/sdk-element-msbuild


guardrex:
I'm going to go ahead with this because I have a small change in the Configuration topic that I want to get live asap. Ping me if this one needs more thought/work.

PR #14542, opened 2019-09-24
Update typo in logging doc

anurse:
Noticed this type while on the doc and it seemed incorrect, so I fixed it up :).


Rick-Anderson:
@anurse  BTW, @tdykstra  move over to the .NET team

PR #14580, opened 2019-09-25
Fix options.ReceiveMaxMessageSize example

jtattermusch:
the size is in bytes

### DIFF
--- a/aspnetcore/grpc/migration.md
+++ b/aspnetcore/grpc/migration.md
@@ -43,7 +43,7 @@ However, a service implementation with a singleton lifetime is no longer able to
 
 In C-core-based apps, settings such as `grpc.max_receive_message_length` and `grpc.max_send_message_length` are configured with `ChannelOption` when [constructing the Server instance](https://grpc.io/grpc/csharp/api/Grpc.Core.Server.html#Grpc_Core_Server__ctor_System_Collections_Generic_IEnumerable_Grpc_Core_ChannelOption__).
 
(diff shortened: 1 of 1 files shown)


jtattermusch:
@JunTaoLuo please take a look

aspnetcore/grpc/migration.md
```diff
@@ -50,7 +50,7 @@ public void ConfigureServices(IServiceCollection services)
 {
     services.AddGrpc(options =>
     {
-        options.ReceiveMaxMessageSize = 16384; // 16 MB
+        options.ReceiveMaxMessageSize = 16 * 1024 * 1024; // 16 MB
```
guardrex: Since this changes to a non-default message size as a demonstration, I think a better fix is to just remove the `// 16 MB` in favor of `// Changes the default 16 MB limit`.

jtattermusch: AFAIK the default is 4MB.   So it's just that the comment and the actual value don't match.

guardrex: Yes, indeed ... so it is.

jtattermusch: Feel free to change to comment to  `// Changes the default limit to 16MB` or sth like that.

guardrex: It's cool. We link in-text better anyway. Those remarks get translated into other languages. Code comments are stuck in English forever.

guardrex:
Thanks @jtattermusch ... I'll shoot this in as soon the build finishes.

JunTaoLuo:
posthoc :shipit:

PR #14598, opened 2019-09-25
Update SignalR Client Features matrix

anurse:
A few updates:

* Added transports
* Added Azure SignalR support - every client supports it, but we do get questions like "can I use the service with Java?" so it's helpful to confirm that in the matrix, and technically the client has to do some (very minimal) work to support it. I put this at the top, because it's so awesome 😉, and that messed up the visual diff a bit (sorry!).
* Added note about issue tracking automatic reconnect

### DIFF
--- a/aspnetcore/signalr/client-features.md
+++ b/aspnetcore/signalr/client-features.md
@@ -14,7 +14,7 @@ uid: signalr/client-features
 
 The table below shows the features and support for the clients that offer real-time support.
 
-| Feature | .NET Core | JavaScript | Java |
+| Feature | .NET | JavaScript | Java |
 | ---- | :-: | :-: | :-: |
 | Azure SignalR Service Support |✔|✔|✔|
 | [Server-to-client Streaming](xref:signalr/streaming)          |✔|✔|✔|
@@ -23,6 +23,8 @@ The table below shows the features and support for the clients that offer real-t
 | WebSockets Transport |✔|✔|✔|
 | Server-Sent Events Transport |✔|✔| |
 | Long Polling Transport |✔|✔|✔|
+| JSON Hub Protocol |✔|✔|✔|
+| MessagePack Hub Protocol |✔|✔| |
 
 Support for automatic reconnect in the Java client is tracked in [our issue tracker](https://github.com/aspnet/AspNetCore/issues/8711).
 
--- a/aspnetcore/signalr/client-features.md
+++ b/aspnetcore/signalr/client-features.md
@@ -16,9 +16,15 @@ The table below shows the features and support for the clients that offer real-t
 
 | Feature | .NET Core | JavaScript | Java |
 | ---- | :-: | :-: | :-: |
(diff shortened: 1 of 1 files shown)


aspnetcore/signalr/client-features.md
```diff
@@ -16,9 +16,15 @@ The table below shows the features and support for the clients that offer real-t
 
 | Feature | .NET Core | JavaScript | Java |
 | ---- | :-: | :-: | :-: |
+| Azure SignalR Service Support |✔|✔|✔|
 | [Server-to-client Streaming](xref:signalr/streaming)          |✔|✔|✔|
 | [Client-to-server Streaming](xref:signalr/streaming)          |✔|✔|✔|
 | Automatic Reconnection ([.NET](/aspnet/core/signalr/dotnet-client?view=aspnetcore-3.0&tabs=visual-studio#handle-lost-connection), [JavaScript](/aspnet/core/signalr/javascript-client?view=aspnetcore-3.0#reconnect-clients))          |✔|✔| |
+| WebSockets Transport |✔|✔|✔|
+| Server-Sent Events Transport |✔|✔| |
```
anurse: FYI, we have no plans to support SSE in Java (and are exploring ways to phase it out of all clients), so I didn't reference an issue for it in the list below.

aspnetcore/signalr/client-features.md
```diff
@@ -16,9 +16,15 @@ The table below shows the features and support for the clients that offer real-t
...
+| Long Polling Transport |✔|✔|✔|
+
+* Support for automatic reconnect in the Java client is tracked in [our issue tracker](https://github.com/aspnet/AspNetCore/issues/8711).
```
anurse:
```suggestion
Support for automatic reconnect in the Java client is tracked in [our issue tracker](https://github.com/aspnet/AspNetCore/issues/8711).
```

aspnetcore/signalr/client-features.md
```diff
@@ -14,11 +14,19 @@ uid: signalr/client-features
 
 The table below shows the features and support for the clients that offer real-time support.
 
-| Feature | .NET Core | JavaScript | Java |
+| Feature | .NET | JavaScript | Java |
```
anurse: Also, the client is for .NET Standard and supports .NET Framework, Xamarin, .NET Core, Unity, etc. So calling it ".NET Core" isn't really appropriate. Changed to `.NET`

davidfowl:
Don't we need versions in here?

halter73:
The document has `monikerRange: '>= aspnetcore-3.0'` in its metadata.

davidfowl:
That's not useful for client versions. For example, when we introduce streaming in the java client, or reconnect in the java client. ASP.NET Core 3.0 is insufficient.

anurse:
Perhaps if the matrix includes a minimum version number instead of just a check box? I can sketch that out.

anurse:
https://github.com/aspnet/AspNetCore.Docs/pull/14744

PR #14668, opened 2019-09-27
C125519: Reverting change from &colon; to : to avoid issues in localized bookmarks

v-maudel:
[EDIT by guardrex to show encoded value of the colon]

Description: `&colon;` should be changed for `:`

### DIFF
--- a/aspnetcore/mvc/views/razor.md
+++ b/aspnetcore/mvc/views/razor.md
@@ -221,7 +221,7 @@ The `<text>` tag is useful to control whitespace when rendering content:
 * Only the content between the `<text>` tag is rendered.
 * No whitespace before or after the `<text>` tag appears in the HTML output.
 
-### Explicit line transition with \@&colon;
+### Explicit line transition with \@:
 
 To render the rest of an entire line as HTML inside a code block, use the `@:` syntax:
 


v-maudel:
Hi @rick-anderson, can you please help by merging this PR?
Thanks in advance.

guardrex:
@Rick-Anderson This was done specifically to make bookmarks work better. Recommend not merging this just yet. Recommend allowing me investigate this over the weekend and get back to you. Besides, this isn't the only place we have this, so we'll need more updates than this to resolve it if we end up reverting earlier changes.

guardrex:
@v-maudel Can you provide the cross-ref to the original community issue that was reported to you? I'd like to see what they say was failing for them in the linking.

v-maudel:
Hi @guardrex , this is the request I received [Loc Issue 125519](https://ceapex.visualstudio.com/CEINTL/_workitems/edit/125519)
This came without extended description but I will check with the team.

guardrex:
Thanks ... when you get the page that they were looking at, could you just paste the loc version (non-English) link in here? I have a feeling that we'll need to revert all of them, but I'd like to know more about how the bookmark broke for them. I'll take a closer look at this over the weekend ... I'm off today. 🏖 :smile:

v-maudel:
@guardrex I've checked it by myself and found the bookmark link is not well created for localized live page: https://docs.microsoft.com/es-es/aspnet/core/mvc/views/razor?view=aspnetcore-3.0#explicit-razor-expressions
image

v-maudel:
As additional I've received the comment from the content team:
When the original text is processed by transformer it is converted to:
image

this doesn't match with bookmark calls written in other files causing publishing issues.
After changing the text as proposed it is transformed as:
image


Fixing the issue as the bookmarks links will match.

guardrex:
:point_up: WRT your last remark ... yes ... but that's not what we're going for. I'll explain ......

The bookmark itself is great ... the destination is :boom:.

What I was trying to do (and I don't know yet if this only applies to the colon) was to get both the bookmark link and the bookmark destination (the `id` value of the heading) to compose better. We don't really want hanging dashes or double-dashes in the bookmarks (`#word-word-` ☹️ or `#word-word--word` ☹️).

I would like to research this a bit more this weekend. CommonMark might have a nice solution for these characters. Ideally, we want this in the current case ...

* Bookmark in the source link: `#explicit-line-transition-with-colon`
* Destination section header rendered as: `<h3 id="explicit-line-transition-with-colon">` ... but ... **but** ... we want to code the header in CommonMark (`### Explicit line transition with @:` or `### Explicit line transition with :`), not use an actual `<h3>` in the topic. One obvious way is just to say "colon" in the header. Eliminating the characters and using their names would solve this, too.

I'll get back to u soon. I need to take a closer look. Either way tho, I'll either merge this PR and fix the others on a new PR or create a new PR with a proper fix this weekend.

guardrex:
btw --- when I refer to what's breaking, I'm saying the English version. I think the loc version is the reverse of that. Here's what happens on the English version ...

Capture1

guardrex:
Moved to https://github.com/aspnet/AspNetCore.Docs/pull/14682.

PR #14693, opened 2019-09-29
[FIX] Fix spacing problems

kahveci:
There are problems with spacing in a few different places as follows:

- Redundant white spaces in the line 226, 281, and 320.

- No space in the line 402 between the words "...in its" and "`id` parameter."

This will fix all.


guardrex:
@kahveci If you agree with the terms, please sign the CLA :point_up:. If you don't, we can't accept the PR.

PR #14716, opened 2019-09-30
Update startup.md

### LINKED ISSUE #14694: ` IStartupFilter`
` IStartupFilter` still doesn't make sense to me.
You can control the order of middleware execution via registration order and you can position the code to execute before or after the `next` statement.

So why use ` IStartupFilter` at all. 

---
#### Document Details

* Content: [App startup in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/startup?view=aspnetcore-3.0#configure-services-without-startup)
* Content Source: [aspnetcore/fundamentals/startup.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/fundamentals/startup.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

Rick-Anderson:
Fixes #14694
@Legends please review

### DIFF
--- a/aspnetcore/fundamentals/startup.md
+++ b/aspnetcore/fundamentals/startup.md
@@ -186,7 +186,10 @@ To configure services and the request processing pipeline without using a `Start
 
 ## Extend Startup with startup filters
 
-Use <xref:Microsoft.AspNetCore.Hosting.IStartupFilter> to configure middleware at the beginning or end of an app's [Configure](#the-configure-method) middleware pipeline. `IStartupFilter` is used to create a pipeline of `Configure` methods. [IStartupFilter.Configure](xref:Microsoft.AspNetCore.Hosting.IStartupFilter.Configure*) can set a middleware to run before or after middleware added by libraries.
+Use <xref:Microsoft.AspNetCore.Hosting.IStartupFilter>:
+
+* To configure middleware at the beginning or end of an app's [Configure](#the-configure-method) middleware pipeline without an explicit call to `Use{Middleware}`. `IStartupFilter` is used by ASP.NET Core to add defaults to the beginning of the pipeline without having to make the app author explicitly register the default middleware. `IStartupFilter` allows a different component call `Use{Middleware}` on behalf of the app author.
+* To create a pipeline of `Configure` methods. [IStartupFilter.Configure](xref:Microsoft.AspNetCore.Hosting.IStartupFilter.Configure*) can set a middleware to run before or after middleware added by libraries.
 
 `IStartupFilter` implements <xref:Microsoft.AspNetCore.Hosting.StartupBase.Configure*>, which receives and returns an `Action<IApplicationBuilder>`. An <xref:Microsoft.AspNetCore.Builder.IApplicationBuilder> defines a class to configure an app's request pipeline. For more information, see [Create a middleware pipeline with IApplicationBuilder](xref:fundamentals/middleware/index#create-a-middleware-pipeline-with-iapplicationbuilder).
 


Rick-Anderson:
@serpent5 can you help me word this?

Legends:
The wording is not the problem, but let's recap David's statements:

> when you want to inject middleware at the beginning or end of the pipeline without an explicit call to Use* in the pipeline.

Ok, you don't have an explicit call to `Use*` here, but therefore you have a DI registration for the `IStartupfilter` which registers the middleware, so nothing gained here, instead of a `use` call you have a DI registration.

serpent5:
I think the wording here is fine too. It seems to me like the problem is in the value proposition of startup filters. To address that, it might be helpful to look at a real-world example...

From the [docs](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host?view=aspnetcore-3.0#settings-for-web-apps), setting up the Generic Host to work as a web app looks like this:

```cs
Host.CreateDefaultBuilder(args)
    .ConfigureWebHostDefaults(webBuilder =>
    {
        ...
    });
````

When `ConfigureWebHostDefaults` runs, it wants to add [HostFilteringMiddleware](https://github.com/aspnet/AspNetCore/blob/master/src/Middleware/HostFiltering/src/HostFilteringMiddleware.cs#L21) to the pipeline. Instead of relying on the app's author adding a call to `UseHostFiltering` in e.g. `Startup.Configure`, it adds an implementation of `IStartupFilter` ([`HostFilteringStartupFilter`](https://github.com/aspnet/AspNetCore/blob/master/src/DefaultBuilder/src/HostFilteringStartupFilter.cs)) to the DI container. This means that it gets to:

1. Ensure that the middleware gets added.
2. Decide whether to run the middleware at the beginning or the end of the pipeline.

Rick-Anderson:
@serpent5  should I close this PR and you can PR with your better explanation?

serpent5:
@Rick-Anderson I like the changes you've made here - I've just expounded on those. I'd like to know whether @Legends feels that the additional information provided above sufficiently addresses the points raised and whether or not it would be helpful for said information to go into the docs in some form.

Legends:
> When `ConfigureWebHostDefaults` runs, it wants to add `HostFilteringMiddleware  `to the pipeline. Instead of relying on the app's author adding a call to `UseHostFiltering `in e.g. `Startup.Configure`, it adds an implementation of `IStartupFilter` (`HostFilteringStartupFilter `) to the DI container. This means that it gets to:
> 
> Ensure that the middleware gets added.
> Decide whether to run the middleware at the beginning or the end of the pipeline.

So `ConfigureWebHostDefaults` is an extension method of the ASP.NET Core framework which adds default settings for the host, the app author doesn't have to bother with anything here.

In order to get a better understanding of `IStartupFilter` I registered three different middlewares by registering their respective `IStartupFilter` implementation to the DI container.
All three filters call `UseMiddleware<>` before `next(builder);` and all three middlewares were called in the order their respective Filters were registered to the DI container.

If I add a call to `UseMiddleware<>` in one of the `IStartupFilter`s after `next(builder);`, its `Invoke` method does not get executed at all.
I guess the request hits a terminal middleware before getting to it.

All in all, I don't see how you put middleware to the beginning or the and of an app's middleware pipeline using `IStartupFilter`s.

I guess my point of view is different or I am missing a logical use case for `IStartupFilter`s currently.

But lets close this, as I don't need this right now and I don't want to bother you any longer.
Thanks for trying to help.

serpent5:
> I guess the request hits a terminal middleware before getting to it.

Yeah, I expect that is what's happening in your example. Create a new Web Application with MVC or Razor Pages and add an implementation of `IStartupFilter` that runs some middleware _after_ `next(builder);`. Have the middleware write a basic response (`context.Response.WriteAsync("Terminal Middleware");` should be enough). Load up the app in a browser and it'll give you the standard Index action/page. Change the URL to something that should return a 404 and you should see your "Terminal Middleware" output as plain-text.

> I guess my point of view is different or I am missing a logical use case for `IStartupFilters` currently.

I think it's more of an advanced use-case, more for library authors to be able to influence the middleware pipeline without requiring the app author to do anything other than call a convenience method (e.g. `IHostBuilder.UseMyAwesomeLibrary`). 

> Thanks for trying to help.

You're welcome.

Legends:
@serpent5 I just re-read the "The Startup Class" section and David's comments in the issue I have opened. It's really just how he explained, but I thought there must be something more complicated here :D ;-)

Again, thanks for taking your time.

PR #14722, opened 2019-09-30
Fix broken link in Blazor Components Event handling section

scottaddie:


guardrex:
If `EventArgs` isn't available via API, plz point that link to ...

https://github.com/aspnet/AspNetCore/tree/release/3.0/src/Components/Web/src/Web

guardrex:
... that :boom: every bleed'in release, but it will be fine now that the bits are on their `master`.

guardrex:
I just checked API. Not there yet.

guardrex:
Nevermind ... let's just go with the link at the bottom.

Issue #14731, opened 2019-09-30
Incorrect example in the Additional Fields section?


pjdennis:
It looks like the first example in the [Additional fields](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-3.0#additional-fields) section is incorrect - shouldn't FirstName reference AdditionalFields = nameof(LastName) and vice versa?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0156ffc3-47ac-d5da-fbc2-9ff5018b071b
* Version Independent ID: 615db388-a244-41f7-21b5-a4085d9ed479
* Content: [Model validation in ASP.NET Core MVC](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-3.0#feedback)
* Content Source: [aspnetcore/mvc/models/validation.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/mvc/models/validation.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-mvc**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
[Additional fields](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-3.0#additional-fields)

Rick-Anderson:
@serpent5 please review

serpent5:
Yeah, @pjdennis is correct.

serpent5:
@Rick-Anderson I think this can be closed.

Rick-Anderson:
@serpent5  I thought this was fixed but [GitHub history](https://github.com/aspnet/AspNetCore.Docs/commits/master/aspnetcore/mvc/models/validation.md) doesn't show the PR.

serpent5:
It was just a change to the sample code in https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/mvc/models/validation/sample/Models/User.cs.

Rick-Anderson:
That's it, thanks!

PR #14732, opened 2019-09-30
Fix in AdditionalFields property sample

olcay:
The AdditionalFields property of the [Remote] attribute is for the other fields that you want to relate but in the sample code, the field is referring to itself as an additional field.


serpent5:
This fixes #14731.

Rick-Anderson:
@serpent5  can I merge this?

serpent5:
@Rick-Anderson Yeah. I checked this same fix myself in the sample project and it's good.

Rick-Anderson:
@olcay  thanks for fixing this!

PR #14759, opened 2019-10-02
components reconnect status failed is renamed

MasihAkbari:
`components-reconnect-failed` is renamed to `components-reconnect-rejected` so just updating this

### DIFF
--- a/aspnetcore/blazor/hosting-models.md
+++ b/aspnetcore/blazor/hosting-models.md
@@ -131,7 +131,7 @@ When the client detects that the connection has been lost, a default UI is displ
 
 * `components-reconnect-show` &ndash; Show the UI to indicate the connection was lost and the client is attempting to reconnect.
 * `components-reconnect-hide` &ndash; The client has an active connection, hide the UI.
-* `components-reconnect-failed` &ndash; Reconnection failed. To attempt reconnection again, call `window.Blazor.reconnect()`.
+* `components-reconnect-rejected` &ndash; Reconnection rejected. To attempt reconnection again, call `window.Blazor.reconnect()`.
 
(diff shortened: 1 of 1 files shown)


guardrex:
Hello @MasihAkbari ... Looks like it should be an **_add_** to the list ...

https://github.com/aspnet/AspNetCore/blob/release/3.0/src/Components/Web.JS/src/Platform/Circuits/UserSpecifiedDisplay.ts

MasihAkbari:
> Hello @MasihAkbari ... Looks like it should be an **_add_** to the list ...
> 
> https://github.com/aspnet/AspNetCore/blob/release/3.0/src/Components/Web.JS/src/Platform/Circuits/UserSpecifiedDisplay.ts
> 
> I updated it. Thanks for sending this in.

I think you should add more info there, so what is the difference between `failed` and `rejected` states?

guardrex:
> what is the difference between failed and rejected states?

Great idea.

Ryan will tell us, and I'll update to use his language.

rynowak:
>what is the difference between failed and rejected states?

Failed -> We were not able to connect, probably due to network failure
Rejected -> Server was reachable but refused the connection, your state on the server is gone

You'll get the rejected state when you have a crash in the circuit (server-side code) or when you've been disconnected long enough for the server to drop your state. It means that the instances of components you were talking to are gone.

MasihAkbari:
great thanks, one more thing though, in `components-reconnect-rejected` state is calling `window.Blazor.reconnect()` do anything now that state on the server is gone? isn't it better to say call `location.reload();` to reload the tap?

rynowak:
Yes, I think @MasihAkbari is correct. You won't be able to reconnect from the rejected state.

PR #14762, opened 2019-10-02
Clarify that Visual C++ 2015 Redistributable needs Update 3

jawn:
While the current link is correct, this clarifies the specific version.
This page states the Update 3 is required: https://docs.microsoft.com/en-us/dotnet/core/windows-prerequisites?tabs=netcore30

### DIFF
--- a/aspnetcore/tutorials/publish-to-iis.md
+++ b/aspnetcore/tutorials/publish-to-iis.md
@@ -37,7 +37,7 @@ This tutorial covers the following subjects:
 
 ## Install the .NET Core Hosting Bundle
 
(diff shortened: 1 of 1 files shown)


guardrex:
@jawn I agree on the link text, but there's a potential difference between what .NET Core requires and what the hosting bundle requires. @shirhatti will confirm if this is the best approach going forward (even if this requirement still exists).

shirhatti:
@jkotalik Can you confirm that this is the case?

jkotalik:
I'm not sure at all, but if .NET 3.0 requires Update 3, it's probably a good idea to mention it anyways.

guardrex:
Let's leave it. The hosting bundle installer wanted it. If the machine didn't have an Internet connection and it wasn't installed manually, the dev would :boom:.

Thanks @jawn! :rocket:

PR #14779, opened 2019-10-03
Change title for consistency

donhuvy:
Change title for consistency with https://docs.microsoft.com/en-us/aspnet/core/blazor/index?view=aspnetcore-3.0


guardrex:
Hello @donhuvy ... we want to leave the technology name there. It goes into the `<title>` and meta title tags. **_However_**, I'm going to flip your change and add the technology name to the H1. Thanks for calling this out.

PR #14802, opened 2019-10-03
Update components.md

eugeneniemand:
Should be “a” not “an”


guardrex:
**An** "at" `code` block. Across the docs we say "at" for `@` when reading the text.

Issue #14253, opened 2019-09-11
Execution Error in AspNetCore.Docs\aspnetcore\mvc\controllers\testing\samples\2.x\TestingControllersSample


Gosha24:
{ description here }

# Software versions

 .NET Core 2.2
I downloaded and run sample AspNetCore.Docs\aspnetcore\mvc\controllers\testing\samples\2.x\TestingControllersSample
I got exception
An error occurred while starting the application.
InvalidOperationException: Cannot resolve scoped service 'TestingControllersSample.Core.Interfaces.IBrainstormSessionRepository' from root provider.
Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteValidator.ValidateResolution(Type serviceType, IServiceScope scope, IServiceScope rootScope)

InvalidOperationException: Cannot resolve scoped service 'TestingControllersSample.Core.Interfaces.IBrainstormSessionRepository' from root provider.
Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteValidator.ValidateResolution(Type serviceType, IServiceScope scope, IServiceScope rootScope)
Microsoft.Extensions.DependencyInjection.ServiceProvider.Microsoft.Extensions.DependencyInjection.ServiceLookup.IServiceProviderEngineCallback.OnResolve(Type serviceType, IServiceScope scope)
Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngine.GetService(Type serviceType, ServiceProviderEngineScope serviceProviderEngineScope)
Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngine.GetService(Type serviceType)
Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(Type serviceType)
Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetService<T>(IServiceProvider provider)
TestingControllersSample.Startup.Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory) in Startup.cs
+
                var repository = app.ApplicationServices.GetService<IBrainstormSessionRepository>();
Microsoft.AspNetCore.Hosting.ConventionBasedStartup.Configure(IApplicationBuilder app)
Microsoft.AspNetCore.Mvc.Internal.MiddlewareFilterBuilderStartupFilter+<>c__DisplayClass0_0.<Configure>g__MiddlewareFilterBuilder|0(IApplicationBuilder builder)
[...]

PR #14844, opened 2019-10-05
Blazor async work guidance

### LINKED ISSUE #14260: Provide guidance about how to do async work in Blazor
Add a statement in both the Components article and the Razor Syntax article that clarifies that unlike .cshtml files, you can't do async work as part of the rendering of the component. Instead, async work needs to happen as part of one of the async component lifecycle events.

guardrex:
Fixes #14260

[Internal Review Topic (links to *Lifecycle methods* section)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/components?view=aspnetcore-3.0&branch=pr-en-us-14844#lifecycle-methods)

* The issue ask (#14260) includes addressing this in the Razor syntax doc, but this doesn't seem to fit there because it's not really "Razor syntax." It seems to be an instruction for component authoring/behavior.
* There's an issue to create a [new Blazor component lifecycle doc page (aspnet/AspNetNet.Docs \#14253)](https://github.com/aspnet/AspNetCore.Docs/issues/14353), so I think what makes sense today is to add explicit NOTES to the *Lifecycle methods* section to clarify that these methods are where async is done. The guidance can be fleshed out by engineering when the new topic is written.

### DIFF
--- a/aspnetcore/blazor/components.md
+++ b/aspnetcore/blazor/components.md
@@ -5,7 +5,7 @@ description: Learn how to create and use Razor components, including how to bind
 monikerRange: '>= aspnetcore-3.0'
 ms.author: riande
 ms.custom: mvc
-ms.date: 09/30/2019
+ms.date: 10/05/2019
 uid: blazor/components
 ---
 # Create and use ASP.NET Core Razor components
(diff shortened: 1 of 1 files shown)


guardrex:
ping @danroth27

guardrex:
Looked like DR was 🏖 or 🎙, so thanks for helping us get these in.

PR #14866, opened 2019-10-06
More detail on migrating Runtime Compilation.

### LINKED ISSUE #14686: 2.2 -> 3.0 Migration docs should clearly explain view compilation change
We need to bump up the visibility of the instructions for migrating runtime compilation.

See: https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-2.2&tabs=visual-studio#opt-in-to-runtime-compilation

This is what the text says:

> In 3.0, runtime compilation is an opt-in scenario. To enable runtime compilation, see Razor file compilation in ASP.NET Core.

This isn't a clear explanation - what is runtime compilation? That's a technical term that lots of our users won't understand. This linked doc isn't much clearer on what the impact will be on most users migrating code.

I'm looking for an explanation that is at the level of "In previous releases we would compile your view code while the app is running. Now this **only** happens as part of the build by default. If you want to compile view code - in response to changes while running - or if you want to compile view code that wasn't present during the build - then you need to go follow these instructions ...."

Then we should have a clear before/after of code samples.

I want to be able to answer questions like this: https://github.com/aspnet/AspNetCore/issues/14544 by pointing people at the link, but it's not good enough.

serpent5:
Fixes #14686.

### DIFF
--- a/aspnetcore/migration/22-to-30.md
+++ b/aspnetcore/migration/22-to-30.md
@@ -4,7 +4,7 @@ author: rick-anderson
 description: Learn how to migrate an ASP.NET Core 2.2 project to ASP.NET Core 3.0.
 ms.author: riande
 ms.custom: mvc
-ms.date: 10/03/2019
+ms.date: 10/06/2019
 uid: migration/22-to-30
 ---
 # Migrate from ASP.NET Core 2.2 to 3.0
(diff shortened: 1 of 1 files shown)


aspnetcore/migration/22-to-30.md
```diff
@@ -456,6 +456,29 @@ public void ConfigureServices(IServiceCollection services)
 }
 ```
 
+## Opt in to runtime compilation of Razor views
```
serpent5: I pulled the whole section up to sit directly after **MVC service registration** as more of a logical grouping. I realise that this wasn't requested and may have been given a lot of thought to end up where it was, so do let me know and I'll put it back.

aspnetcore/migration/22-to-30.md
```diff
@@ -456,6 +456,29 @@ public void ConfigureServices(IServiceCollection services)
...
+    services.AddMvc();
+}
+```
+
+In ASP.NET Core 3.0, runtime compilation of Razor views is an opt-in scenario. To enable the same support as found in ASP.NET Core 2.2, add a reference to the [Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation](https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation/) NuGet package and call `AddRazorRuntimeCompilation`:
+
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
+    services.AddMvc()
```
serpent5: I'm not entirely sure whether this should be `AddMvc`. It seems like the guidance is to use `AddControllersWithViews` et al, but that muddies things a bit here as the "after" code here would have to be something like this:

```cs
services.AddControllersWithViews().AddRazorRuntimeCompilation();
services.AddRazorPages();
```

-or-

```cs
services.AddControllersWithViews();
services.AddRazorPages().AddRazorRuntimeCompilation();
```

The page does highlight that `AddMvc` is still valid in 3.0, but I can't decide whether that's something you guys want to discourage with the new extensions.

aspnetcore/migration/22-to-30.md
```diff
@@ -456,6 +456,29 @@ public void ConfigureServices(IServiceCollection services)
...
+        .AddRazorRuntimeCompilation();
+}
+```
+
+For more information, see <xref:mvc/views/view-compilation>.
```
serpent5: The linked page doesn't really say much more than what's here now, but it still seems to me like the link should stick around.

serpent5:
I just noticed that @pranavkm has tackled this in #14861. I was too late... :)

scottaddie:
Closing in favor of Pranav's PR.

PR #14875, opened 2019-10-07
Fix the correct path for signalr js library

amgdy:
In the tutorials it mentioned that you download signalr.js into js folder but in razor page it reference it from lib folder.


amgdy:
@bradygaster I agree that there is no "correct path" but if someone was following on the steps on the docs and solution won't work.

scottaddie:
We appreciate the PR, @amgdy! Closing this one, as the issue was fixed over in https://github.com/aspnet/AspNetCore.Docs/pull/14956.

PR #14975, opened 2019-10-10
Add attribute namespace

centreboard:
RequiredAttribute etc comes from the System.ComponentModel.DataAnnotations namespace.
Previously implied coming from Microsoft.Extensions.DependencyInjection


Rick-Anderson:
@centreboard  thanks!

PR #15019, opened 2019-10-11
remove custom JWT stuff from the SignalR auth doc

anurse:
Fixes https://github.com/aspnet/AspNetCore.Docs/issues/15017

The JWT configuration in the sample is insecure. To simplify things, I've removed all the JWT issuance code entirely from the sample and replaced it with placeholders explaining where to inject your JWT acquisition/validation code and how to get the token to SignalR.

Stand by for internal review link.

cc @leastprivilege

### DIFF
--- a/aspnetcore/signalr/authn-and-authz/sample/Startup.cs
+++ b/aspnetcore/signalr/authn-and-authz/sample/Startup.cs
@@ -49,21 +49,6 @@ namespace SignalRAuthenticationSample
                 })
                 .AddJwtBearer(options =>
                 {
-                    // Configure JWT Bearer Auth to expect our security key
-                    options.TokenValidationParameters =
-                        new TokenValidationParameters
-                        {
-                            LifetimeValidator = (before, expires, token, param) =>
-                            {
-                                return expires > DateTime.UtcNow;
-                            },
-                            ValidateAudience = false,
-                            ValidateIssuer = false,
-                            ValidateActor = false,
-                            ValidateLifetime = true,
-                            IssuerSigningKey = SecurityKey
-                        };
-
                     // We have to hook the OnMessageReceived event in order to
                     // allow the JWT authentication handler to read the access
                     // token from the query string when a WebSocket or 
--- a/aspnetcore/signalr/authn-and-authz/sample/wwwroot/js/chat.ts
+++ b/aspnetcore/signalr/authn-and-authz/sample/wwwroot/js/chat.ts
@@ -15,7 +15,6 @@ function showIf(condition: any, ifTrue: HTMLElement, ifFalse?: HTMLElement) {
     }
 }
 
-const loginForm = dom<HTMLFormElement>("loginForm");
 const chatDiv = dom<HTMLDivElement>("chatDiv");
 const errorDiv = dom<HTMLDivElement>("errorDiv");
 const logoutButton = dom<HTMLButtonElement>("logoutButton");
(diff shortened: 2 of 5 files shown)


anurse:
Need to adjust some snippet references so they show up right, stand by.

anurse:
Snippet refs should be fixed now.

aspnetcore/signalr/authn-and-authz/sample/wwwroot/js/chat.ts
```diff
@@ -34,27 +33,18 @@ class ViewModel {
     public connection: signalR.HubConnection;
     public connectionStarted: boolean;
 
-    public async loginFormSubmitted(evt: Event) {
+    // TODO: Bind this method to a UI event that is triggered when the user is logging in
```
halter73: Did you mean to leave this as a TODO?

halter73: Nvm, you clearly did.

anurse: Yep :). It's unfortunate that the sample isn't runnable, but when you run it it gives you a pretty clear message to go look at these TODOs and hook things up.

Rick-Anderson:
@halter73  can I merge this or given it's security related do we need @blowdart  to bless it?

aspnetcore/signalr/authn-and-authz/sample/Startup.cs
```diff
@@ -49,21 +49,6 @@ public void ConfigureServices(IServiceCollection services)
...
-                        {
-                            LifetimeValidator = (before, expires, token, param) =>
-                            {
-                                return expires > DateTime.UtcNow;
-                            },
-                            ValidateAudience = false,
-                            ValidateIssuer = false,
-                            ValidateActor = false,
-                            ValidateLifetime = true,
-                            IssuerSigningKey = SecurityKey
```
blowdart: I think the signing key line needs to stay, the validators can go.

anurse:
I'm confused, I've tried to remove all custom JWT configuration entirely and leave placeholders for users to fill in. What's the better process here? Should I keep something like:

```csharp
var issuerKey = /* TODO: Get the security key that will be used to validate the token issuer */
options.TokenValidationParameters = new TokenValidationParameters
    {
        IssuerSigningKey = issuerKey
    }
```

blowdart:
That works; maybe expand it

```
var issuerKey = /* TODO: Get the security key that will be used to validate the token issuer */
options.TokenValidationParameters = new TokenValidationParameters
    {
        // TODO: Add other validation parameters to match your authentication properties.
        IssuerSigningKey = issuerKey
    }
```

anurse:
What about

```csharp
options.TokenValidationParameters = new TokenValidationParameters()
    {
        // TODO: Add validation parameters to match your authentication properties.
    }
```

Or is `IssueSigningKey` always going to be needed?

blowdart:
If you're not setting it anywhere then no, that works

anurse:
Talked with @Tratcher, what about this for the entire JwtBearer config

```csharp
.AddJwtBearer(options =>
{
    // Configure the Authority to the expected value for your authentication provider
    // This ensures the token is appropriately validated
    options.Authority = /* TODO: Insert Authority URL here */;

    // We have to hook the OnMessageReceived event in order to
    // allow the JWT authentication handler to read the access
    // token from the query string when a WebSocket or 
    // Server-Sent Events request comes in.
    options.Events = new JwtBearerEvents
    {
        OnMessageReceived = context =>
        {
            var accessToken = context.Request.Query["access_token"];

            // If the request is for our hub...
            var path = context.HttpContext.Request.Path;
            if (!string.IsNullOrEmpty(accessToken) &&
                (path.StartsWithSegments("/hubs/chat")))
            {
                // Read the token out of the query string
                context.Token = accessToken;
            }
            return Task.CompletedTask;
        }
    };
});
```

(Yes, I know the query string thing makes us all cry, let's move on ;))

blowdart:
Noooooooooooooooooooooooooooooooooooooooooooooooooooooooo bad query string. But yea, sure. Can we link to the note in the docs about querystrings within the same code?

anurse:
> Noooooooooooooooooooooooooooooooooooooooooooooooooooooooo bad query string

Yes yes, I know.


> Can we link to the note in the docs about querystrings within the same code?

You got it dude!

anurse:
Updated comments to link to the detailed section on the access token (rather than just the small blurb). Also to use `Audience` config instead of `TokenValidationParameters`.

scottaddie:
Thank you for the PR, @anurse!

PR #15072, opened 2019-10-14
Remove curly braces from GUID in user-secrets

leighghunt:
Curly braces in `UserSecretsId` seem to trigger an msbuild bug (see https://github.com/Microsoft/msbuild/issues/1357) - I just experienced this in a test dotnet 2.2 project on Azure App Service. Removing curly braces to avoid encouraging people falling into this trap :-)


guardrex:
Hello @leighghunt ... that's by-design. We use the braces as a placeholder, and the instruction indicates that the entire `{GUID}` should be replaced with the GUID. We'll keep an :ear: out for more feedback like this tho. If it trips up a lot of devs, we'll definitely do something different.

PR #15096, opened 2019-10-14
Integration Testing topic+sample dB replacement

### LINKED ISSUE #14354: How to swap dB guidance in Integration Testing topic
React to https://github.com/aspnet/AspNetCore.Docs/pull/14265#issuecomment-532171607

---
#### Document Details

* Content: [Integration tests in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-2.2)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #14354

* I tested this with both an in-memory dB and SQL Server dB.
* Hoping that we can go with something simpler than the lower-level approach taken by the engineering sample (i.e., can't we just `services.AddDbContext<T>(...)` in the normal way after killing the existing service registration as I show here?).
* Interestingly, I didn't need to keep the `UseInternalServiceProvider`. I have no real confidence tho in this move outside of it ✨ *Just Works*:tm: ✨ in testing here. Do you want to ping an EF Core :cat2: to make sure that this approach is sound?

### DIFF
--- a/aspnetcore/test/integration-tests.md
+++ b/aspnetcore/test/integration-tests.md
@@ -5,7 +5,7 @@ description: Learn how integration tests ensure that an app's components functio
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 10/07/2019
+ms.date: 10/14/2019
 uid: test/integration-tests
 ---
 # Integration tests in ASP.NET Core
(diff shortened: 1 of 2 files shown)


javiercn:
You can ping @bricelam or @ajcvickers to validate the approach we are using, but this is what we do in Identity and I believe we reviewed it at the time.

There's nothing EF specific here other than replacing services in the container. When we did this for Identity I remember looking up how EF did the registration to achieve the desired result.

Looks good to me

guardrex:
The only thing that concerned me was that at least one of the approaches (it might be the seeding approach in EF Core docs) required **migrations**.

Having the user be distracted by migrations in topics that have nothing at all to do with EF Core seems like a poor UE to me. (Also harder for us to maintain in in the doc.)

The approach used by this topic+sample and several others on the repo works well ... **_no migrations required._** The dev just pulls down the sample and runs it. :tada:🎈:smile: 

If we need to tell devs somewhere along the way, *don't do this ... use migrations for seeding* and cross-link, sure thing :+1: ... I hope that's the direction we take. Plz don't add migrations unless you absolutely have to.

guardrex:
@scottaddie If you're *up* 🌞, I'd like to get this in asap.

javiercn:
> The only thing that concerned me was that at least one of the approaches (it might be the seeding approach in EF Core docs) required **migrations**.
> 
> Having the user be distracted by migrations in topics that have nothing at all to do with EF Core seems like a poor UE to me. (Also harder for us to maintain in in the doc.)
> 
> The approach used by this topic+sample and several others on the repo works well ... **_no migrations required._** The dev just pulls down the sample and runs it. 🎉🎈😄
> 
> If we need to tell devs somewhere along the way, _don't do this ... use migrations for seeding_ and cross-link, sure thing 👍 ... I hope that's the direction we take. Plz don't add migrations unless you absolutely have to.

As far as I remember this approach doesn't require migrations. I believe the Db is created on a per-waf basis. Seeding is something you can do on the test itself by retrieving the DbContext and performing the seed in place.

Is there something I'm missing here?

guardrex:
No, that's fine. :+1: We're good. That's what we do.

The app also seeds. I think seeding *can* use migrations there. I'm just speaking generally on the subject for apps. I should have clarified that I wasn't speaking about the test project.

The reason for app seeding is so that the dev can see seeding of the app's dB is different from the test seeding. The messages in testing start with something like `TEST MESSAGE:` to call it out clearly.

guardrex:
@scottaddie I'm off to the 🏋 gym 💪 ... if you get a chance to :eyes:, this should cue up for `master` on https://github.com/aspnet/AspNetCore.Docs/pull/15100.

guardrex:
@Rick-Anderson This is the priority PR that we discussed. It just needs your quick :eye:. Please merge and take live or let me know on your review to take live.

PR #15143, opened 2019-10-16
Blazor JS Interop example enhancements

### LINKED ISSUE #15140: JS function for "setElementValue" (on <Detect when a Blazor app is prerendering> section)
Hey,
I know this is not so tricky ;) but just for the sake of the documentation :), no JavaScript function has provided for "setElementValue" so it leads to "Nothing" on the &lt;Detect when a Blazor app is prerendering&gt; section, and raising an error for the next part.

wish the best.
 

---
#### Document Details

* Content: [ASP.NET Core Blazor JavaScript interop](https://docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interop?view=aspnetcore-3.0#detect-when-a-blazor-app-is-prerendering)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #15140

**Ok ... ready now for review. I made another round of small updates this morning (Thur, 10/17).**

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interop?view=aspnetcore-3.0&branch=pr-en-us-15143)

The ask is to improve the examples for JS Interop:

* Show the JS functions involved even if they aren't germane to the focus of the section's content because:
  * Devs **_DO_** like to cut-'n-paste code bits into sample apps to play with the bits. Providing these is low cost with plenty of upside for dev efficiency.
  * It also allows us to reinforce an `InvokeAsync`/`InvokeVoidAsync` concept: JS functions return/don't return values, respectively.
* I'm also going work on consistency in JS function naming: These updates go with the camel case everywhere.

Not quite 100% there tho on this one (*javascript-interop\samples_snapshot\inject-abstraction.razor*) ...

```cshtml
@inject IJSRuntime JSRuntime

@code {
    protected override void OnInitialized()
    {
        StocksService.OnStockTickerUpdated += stockUpdate =>
        {
            JSRuntime.InvokeVoidAsync("handleTickerChanged",
                stockUpdate.symbol, stockUpdate.price);
        };
    }
}
```

What's the story there on **_not_** awaiting the call to `InvokeVoidAsync`? Should it await, then `async stockUpdate`? ... or is this ok in its present form?

Thanks @aminm1364 🏍 and @Legends 🎸


aspnetcore/includes/blazor-prerendering.md
```diff
@@ -11,17 +11,25 @@ To delay JavaScript interop calls until after the connection with the browser is
...
         }
     }
 }
 ```
 
+For the preceding example code, provide a `setElementValue` JavaScript function inside the `<head>` element of *wwwroot/index.html* (Blazor WebAssembly) or *Pages/_Host.cshtml* (Blazor Server). The function is called with `IJSRuntime.InvokeVoidAsync` and doesn't return a value:
+
+```html
+<script>
+  window.setElementValue = (elem, val) => elem.value = val;
```
danroth27: I'm worried that this is showing an anti-pattern. By manipulating the DOM directly in a way that Blazor is unaware of, you run the risk of messing up Blazor's diffing algorithm. I better example might be initializing some sort of JS component, but I don't know what a simple example might look like. @pranavkm @javiercn @SteveSandersonMS any ideas?

guardrex: We can always add a note that this is for demo purposes only. We've done that a few times with good outcomes AFAIK. I recommend an explicit remark in the topic and a code comment to that effect.

guardrex:
Plz don't forget about my ask in the opening comment on async in that scenario. :point_up:

aspnetcore/includes/blazor-prerendering.md
danroth27: Yeah, I would suggest two things: 1. change the DOM element from an input to a static div (so folks don't try to bind to it or do other Blazor-y things), and 2. Add a note warning folks that modifying the DOM directly with JS can interfere with Blazor's change tracking.

danroth27:
> Plz don't forget about my ask in the opening comment on async in that scenario. ☝️

I reviewed that example with @pranavkm and we think the call to `InvokeVoidAsync` in the event handler is fine as is.

guardrex:
@danroth27 Examples updated to use `<div>` seems nice ... and the WARNING notes are added in both spots.

guardrex:
@danroth27 Wait for Javier to look or go ahead and merge this?

PR #15193, opened 2019-10-18
Configuration language enhancement

### LINKED ISSUE #15182: Note that Configuration.GetValue<T> doesn't support scenarios where T is an array
Example `appsettings.json`:

```json
{
  "MySettings": {
    "Domains": [
      "https://www.example.com",
      "https://www.github.com"
    ]
  }
}
```

Attempting to read this setting in the following way does not work:

```csharp
var myDomains = Configuration.GetValue<string[]>("MySettings:Domains"); // myDomains is null
```

The documentation should be updated to reflect this, as it's not obvious that this will not work.

---
#### Document Details

* Content: [Configuration in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-3.0#json-configuration-provider)
* Content Source: [aspnetcore/fundamentals/configuration/index.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/fundamentals/configuration/index.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #15182

Indicate to the reader that `GetValue` is for single values with noncollection types.

Thanks @IanKemp! :rocket:

### DIFF
--- a/aspnetcore/fundamentals/configuration/index.md
+++ b/aspnetcore/fundamentals/configuration/index.md
@@ -767,7 +767,7 @@ The dictionary is used with a call to `AddInMemoryCollection` to provide the con
 
 ## GetValue
 
-[ConfigurationBinder.GetValue\<T>](xref:Microsoft.Extensions.Configuration.ConfigurationBinder.GetValue*) extracts a single value from configuration with a specified key and converts it to the specified noncollection type. An overload accepts a default value if the key isn't found.
+[ConfigurationBinder.GetValue\<T>](xref:Microsoft.Extensions.Configuration.ConfigurationBinder.GetValue*) extracts a single value from configuration with a specified key and converts it to the specified noncollection type. An overload accepts a default value.
 
 The following example:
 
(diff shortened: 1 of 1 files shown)


aspnetcore/fundamentals/configuration/index.md
```diff
@@ -767,7 +767,7 @@ The dictionary is used with a call to `AddInMemoryCollection` to provide the con
 
 ## GetValue
 
-[ConfigurationBinder.GetValue\<T>](xref:Microsoft.Extensions.Configuration.ConfigurationBinder.GetValue*) extracts a value from configuration with a specified key and converts it to the specified type. An overload permits you to provide a default value if the key isn't found.
+[ConfigurationBinder.GetValue\<T>](xref:Microsoft.Extensions.Configuration.ConfigurationBinder.GetValue*) extracts a single value from configuration with a specified key and converts it to the specified noncollection type. An overload accepts a default value.
```
Rick-Anderson: [ConfigurationBinder.GetValue\<T>](xref:Microsoft.Extensions.Configuration.ConfigurationBinder.GetValue*) extracts a single value from configuration with a specified key and converts it to the specified noncollection type. An overload accepts a default value if the key isn't found.

Rick-Anderson:
Don't you want to keep `if the key isn't found.`

guardrex:
No ... a default value is what *IS* used in all of these scenarios. The example that immediately follows this line calls it out explicitly. It's so common now in programming (and with the explicit example) that we don't need it.

### **_Less is more._** 

### :smile:

PR #15213, opened 2019-10-19
Challenge vs Forbid for authz webapp

### LINKED ISSUE #15174: Rewrite DetailsModel model to use Forbid
See #6418

---
#### Document Details

* ID: b66f8ffa-699c-8b11-0cab-805f5953805f
* Version Independent ID: 7c151ed5-fd78-84fd-a08a-88912990c4ca
* Content: [Create an ASP.NET Core app with user data protected by authorization](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/secure-data?view=aspnetcore-3.0#update-details)
* Content Source: [aspnetcore/security/authorization/secure-data.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/security/authorization/secure-data.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

serpent5:
Fixes #15174.

cc @Rick-Anderson 

@Rick-Anderson edit
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/security/authorization/secure-data?view=aspnetcore-3.0&branch=pr-en-us-15213#update-the-createmodel)
[Differences between Challenge vs Forbid](https://review.docs.microsoft.com/en-us/aspnet/core/security/authorization/secure-data?view=aspnetcore-3.0&branch=pr-en-us-15213#differences-between-challenge-vs-forbid)


aspnetcore/security/authorization/secure-data/samples/final3/Pages/Contacts/Details2.cshtml
```diff
@@ -0,0 +1,61 @@
...
+            @Html.DisplayNameFor(model => model.Contact.Status)
+        </dt>
+        <dd>
+            @Html.DisplayFor(model => model.Contact.Status)
+        </dd>
+    </dl>
+</div>
+
+<div>
+    <a asp-page="./Index">Back to List</a>
```
serpent5: I removed all the editing features for this to keep it simpler.

aspnetcore/security/authorization/secure-data/samples/final3/Pages/Contacts/Details2.cshtml.cs
```diff
@@ -0,0 +1,56 @@
...
+
+            if (!isAuthorized
+                && currentUserId != Contact.OwnerID
+                && Contact.Status != ContactStatus.Approved)
+            {
+                return Forbid();
+            }
+
+            return Page();
+        }
```
serpent5: The sample code from the linked issue was re "documents" rather than "contacts", so I've copied `Details` from this sample and updated it to `[AllowAnonymous]` and then added the `Challenge` and `Forbid` distinction using the authz logic that was there before.

aspnetcore/security/authorization/secure-data/samples/final3/Startup.cs
```diff
@@ -42,17 +41,11 @@ public void ConfigureServices(IServiceCollection services)
                 .AddEntityFrameworkStores<ApplicationDbContext>();
             #endregion
 
-            services.AddRazorPages();
-
-            services.AddControllers(config =>
```
serpent5: `[AllowAnonymous]` didn't work for the new `Details2` page, which brought me here to see what was going on. The sample app doesn't have any controllers, so I think it's better to drop this in favour of using the RP convention shown. That also fixes the `[AllowAnonymous]` thing.

aspnetcore/security/authorization/secure-data/samples/final3/Pages/Contacts/Details2.cshtml.cs
serpent5: This works for demo purposes, but I think it's a bit unusual that we have authz handlers for everything except reads, which means the authz logic is all inline here. I'd expect additional handlers to take care of the role checks and the approved status, which would be more inline with the rest of the app. That would also allow the following in the page:

```cs
if (!User.Identity.IsAuthenticated)
{
    return Challenge();
}

var authorizationResult = await AuthorizationService
    .AuthorizeAsync(User, Contact, ContactOpertations.Read);

if (!authorizationResult.Succeeded)
{
    return Forbid();
}

return Page();
```

serpent5:
@Rick-Anderson I've marked this one as a draft as I'm not sure how to proceed. It's complete in terms of switching from `Challenge` to `Forbid` and adding `Details2`, but the `Details2` implementation is different from the linked issue as I've noted above. What we've got here might be enough, but it opened up a question re inline authz logic that might need addressing (or not).

aspnetcore/security/authorization/secure-data/samples/final3/Pages/Contacts/Details2.cshtml.cs
serpent5: The sample code from the linked issue was re "documents" rather than "contacts", so I've copied `Details` from this sample and updated it to `[AllowAnonymous]` and then added the `Challenge` and `Forbid` distinction using the authz logic that was there before.

serpent5: This works for demo purposes, but I think it's a bit unusual that we have authz handlers for everything except reads, which means the authz logic is all inline here. I'd expect additional handlers to take care of the role checks and the approved status, which would be more inline with the rest of the app. That would also allow the following in the page:

```cs
if (!User.Identity.IsAuthenticated)
{
    return Challenge();
}

var authorizationResult = await AuthorizationService
    .AuthorizeAsync(User, Contact, ContactOpertations.Read);

if (!authorizationResult.Succeeded)
{
    return Forbid();
}

return Page();
```

aspnetcore/security/authorization/secure-data/samples/final3/Pages/Contacts/Details2.cshtml
```diff
@@ -0,0 +1,61 @@
+@page "{id:int}"
```
serpent5: This is just a copy of `Details`, with editing functionality removed. We're just demoing the `OnGet` handler so there's no need for the `OnPost` and calls to it.

serpent5:
I've been thinking about this a bit more. Here's a _suggestion_ for how we proceed:

1. We go with this PR as-is, with the (maybe) somewhat clunky inline authz for `Details` and `Details2`.
2. I create another PR that follows on from this, with a proposal for how to move the inline authz out in to the handlers. This would require a change to one of the existing handlers and an additional handler for allowing read access to all approved contacts.

Rick-Anderson:
@Tratcher  please review at [Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/security/authorization/secure-data?view=aspnetcore-3.0&branch=pr-en-us-15213#update-the-createmodel)

To see changes in code, on ../files tab, search for `final3`

> I've been thinking about this a bit more. Here's a _suggestion_ for how we proceed:

> 1. We go with this PR as-is, with the (maybe) somewhat clunky inline authz for `Details` and `Details2`.
> 2. I create another PR that follows on from this, with a proposal for how to move the inline authz out in to the handlers. This would require a change to one of the existing handlers and an additional handler for allowing read access to all approved contacts.

My feelings on `with a proposal for how to move the inline authz out in to the handlers. `:

* `@blowdart` has approved this approach, and I'd rather not require another security audit unless it's required.
* The current approach is **relatively** simple, debugable,  and customers can  follow.

Rick-Anderson:
@serpent5  do you want to add something like the following?

## Differences between Challenge vs Forbid 

This app sets the default policy to `[require authenticated users](#require-authenticated-users)`. The following code allows anonymous users. Anonymous users are allowed to  show the differences between Challenge vs Forbid.

code here

In the preceding code:

* when the user is **not** authenticated, `ChallengeResult` is returned. When `ChallengeResult` is returned, the user is redirected to the sign-in page.
* when the user is authenticated, but not authorized, `ForbidResult` is returned. `ForbidResult` ...

serpent5:
> My feelings on `with a proposal for how to move the inline authz out in to the handlers.`:

Understood. It was just a suggestion that was specific to tidying up `Details` and `Details2`, but if you're happy with that the way it is, then sure, it's all good.

 > do you want to add something like the following?

Done. I copied it mostly as-is and completed the `ForbidResult` line. I dunno if we'll need to add a bit more explanation re challenge/forbid, especially as they work differently for e.g JwtBearer, but I won't do anything with that unless instructed.

aspnetcore/security/authorization/secure-data/samples/final3/Pages/Contacts/Create.cshtml.cs
```diff
@@ -54,7 +54,7 @@ public async Task<IActionResult> OnPostAsync()
                                                         ContactOperations.Create);
             if (!isAuthorized.Succeeded)
             {
-                return new ChallengeResult();
+                return Forbid();
```
Tratcher: This one's ambiguous, we don't see if the user is authenticated because that check happens inside authorize. This should be split so that `if (!User.Identity.IsAuthenticated)` it calls Challenge, otherwise it calls Forbid. I already see something similar happening in the other pages.

serpent5: I didn't think it was possible to get into this action at all if the user wasn't authenticated, due to the `AuthorizeFilter` that's applied in `Startup`. Have I misunderstood something?

Tratcher: This one? 
https://github.com/aspnet/AspNetCore.Docs/blob/6b3a072f7899b892b2f91d8e279f3f79257aee07/aspnetcore/security/authorization/secure-data/samples/final3/Startup.cs#L51-L54
You're right. Disregard.

Rick-Anderson:
@Tratcher that's why we added `[allowAnon]` to  `Details2` - to show Challenge and Forbid together.
Anything else before you approve?

PR #15217, opened 2019-10-20
Update Customer.cs model

alex-code4okc:
The example Model code in section "Write a basic form" uses a required string length of 10 for validation, however the model Data Annotation in Model code for section "Validation" uses a required length of 100 which does not matching the rest of the examples. The "Write a basic form" section model code is correct (requiring a string length of 10). However the "Validation" section is incorrect (requiring a string length of 100). Model code seems to be coming from two different locations, "Write a basic form" section using: index/3.0sample/RazorPagesContacts/Models/Customer.cs, and "Validation" section using "index/sample/RazorPagesContacts/Data/Customer.cs" Either update "Validation" model code to require string length of 10, or update the code reference to use the same one as "Write a basic form" section model code.

### DIFF
--- a/aspnetcore/razor-pages/index/sample/RazorPagesContacts/Data/Customer.cs
+++ b/aspnetcore/razor-pages/index/sample/RazorPagesContacts/Data/Customer.cs
@@ -6,7 +6,7 @@ namespace RazorPagesContacts.Data
     {
         public int Id { get; set; }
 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
We should ref the *3.0sample/* code, not the previous version.

alex-code4okc:
Thanks, I'll make a new pull request.

PR #15224, opened 2019-10-21
Use `Microsoft.NET.Sdk.Worker` project SDK on .NET Core 3.0 Windows Service sample

caioproiete:
Update the sample code linked to [Host ASP.NET Core in a Windows Service](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/windows-service?view=aspnetcore-3.0) to use the new `Microsoft.NET.Sdk.Worker` project SDK introduced with .NET Core 3.0

### DIFF
--- a/aspnetcore/host-and-deploy/windows-service/samples/3.x/AspNetCoreService/SampleApp.csproj
+++ b/aspnetcore/host-and-deploy/windows-service/samples/3.x/AspNetCoreService/SampleApp.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+﻿<Project Sdk="Microsoft.NET.Sdk.Worker">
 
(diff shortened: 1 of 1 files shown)


guardrex:
🤔 Hummmmmmm .... the updates were done back in May, which was during preview, and engineering signed off at the time on those updates. I'm not sure on the timing of those updates and any other changes that were going on in the framework at that time. Looking at the template today, I'm not sure if there are additional changes ... *probably are*. The slightly unusual scenario here is that this worker is a **_web app_** with an interface that provides the ability to control the service. I don't recall if there was some requirement at the time that led me to keep the Web SDK here in terms of publishing a web app.

I'm going to move this to an issue and work it more holistically, including a round of testing. Thanks for opening this.

Moved to https://github.com/aspnet/AspNetCore.Docs/issues/15230.

PR #15237, opened 2019-10-21
Update windows auth docs re: NTLM on Kestrel

anurse:
Docs update suggested in https://github.com/aspnet/AspNetCore/issues/14951

Need @Tratcher to confirm that I've interpreted the thread right :).

### DIFF
--- a/aspnetcore/security/authentication/windowsauth.md
+++ b/aspnetcore/security/authentication/windowsauth.md
@@ -5,7 +5,7 @@ description: Learn how to configure Windows Authentication in ASP.NET Core for I
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
 ms.custom: "mvc, seodec18"
-ms.date: 09/03/2019
+ms.date: 10/21/2019
 uid: security/authentication/windowsauth
 ---
 # Configure Windows Authentication in ASP.NET Core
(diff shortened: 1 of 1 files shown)


aspnetcore/security/authentication/windowsauth.md
```diff
@@ -145,7 +145,7 @@ Use **either** of the following approaches:
 
 ## Kestrel
 
-The [Microsoft.AspNetCore.Authentication.Negotiate](https://www.nuget.org/packages/Microsoft.AspNetCore.Authentication.Negotiate) NuGet package can be used with [Kestrel](xref:fundamentals/servers/kestrel) to support Windows Authentication using Negotiate, Kerberos, and NTLM on Windows, Linux, and macOS.
+The [Microsoft.AspNetCore.Authentication.Negotiate](https://www.nuget.org/packages/Microsoft.AspNetCore.Authentication.Negotiate) NuGet package can be used with [Kestrel](xref:fundamentals/servers/kestrel) to support Windows Authentication using Negotiate and Kerberos on Windows, Linux, and macOS. Authentication using NTLM is supported on Windows only.
```
Tratcher: The point in aspnet/AspNetCore#14951 was that NTLM was not directly supported on any platform, only via negotiate.

anurse:
Update: Per @Tratcher 's comments and brief offline discussion, I just dropped NTLM from the doc entirely.

guardrex:
:shipit::shipit::shipit::shipit::shipit::shipit::shipit:

tornie2:
As I understand Kestrel could actually authenticate NTLM
But there is a problem with sending the challange / response

Would it be possible to build our own with middleware like this:

`public class ValidateAuthentication : IMiddleware
{
    public async Task InvokeAsync(HttpContext context, RequestDelegate next)
    {
        if (context.Request.Method == "OPTIONS")
        {
            await next(context);
            return;
        }

        if (context.User.Identity != null && context.User.Identity.IsAuthenticated)
        {
            await next(context);
        }
        else
        {
            context.Response.StatusCode = 401;
            context.Response.Headers["Proxy-Authenticate"] = "NTLM";
            context.Response.Headers["WWW-Authenticate"] = "NTLM";
            context.Response.Headers["Access-Control-Allow-Origin"] = context.Request.Headers["Origin"];
        }
    }
}`

guardrex:
@tornie2 ... We don't work closed issues/PR. Please open a new issue using the **This page** feedback button and form at the bottom of the **English-US** topic. Use of the **This page** feedback form adds metadata to your GitHub issue that cross-links the topic and automatically pings the author.

Capture

Tratcher:
@tornie2 That would not be enough. The implementation only looks for auth handshakes that start with "Negotiate". See
https://github.com/dotnet/aspnetcore/blob/d40eea990edd1ac7205a983cbe4d13a71fd4ad68/src/Security/Authentication/Negotiate/src/NegotiateHandler.cs#L101

tornie2:
@Tratcher  Thanks for the tip. First tests indicate, it could actually work :-)

PR #15244, opened 2019-10-21
Additional Blazor attribute splatting guidance

### LINKED ISSUE #15082: [Blazor] Add a section to explain how splatting override works
The splatting section is missing docs and an example about how the override behavior works for splatting. Specifically it needs to point out that the order in which you put `@attributes` with regards to other attributes on the element matters and that the logic is that the most right attribute wins.

For example:
Call site:
```
<MyComponent extra="10" />
```

MyComponent
```
<div @attributes="AdditionalAttributes" extra="5" />

[Parameter(CaptureUnmatchedValues = true)]
public IDictionary<string,object> AdditionalValues {get; set; }
```

In the case above the div will always be extra="5" as the attribute is set to the right of `@attributes`

If we revert the order
MyComponent
```
<div extra="5" @attributes="AdditionalAttributes" />

[Parameter(CaptureUnmatchedValues = true)]
public IDictionary<string,object> AdditionalValues {get; set; }
```

the rendered div will contain extra=10 when passed through an additional attribute.

guardrex:
Fixes #15082

Although it doesn't change between the two examples, I repeat the `Parent` component in the 2nd example. I don't want the reader to get lost when eyeballing it.


guardrex:
@javiercn This is also probably a quick review/approve/comment one. I think Dan is 🏃😅 right now.

PR #15247, opened 2019-10-21
Run Blazor without debugging

danroth27:
F5 is problematic with Blazor WebAssembly apps. Instead let's recommend Ctrl+F5


guardrex:
@Rick-Anderson Are the `<kbd>` ok? ... we don't normally use the tags, but they do look nice when they render.

guardrex:
Confirmed in the style manual ... they want bold keys.

PR #15272, opened 2019-10-23
Blazor prerendering update

### LINKED ISSUE #15270: Incorrect section
&gt; The client reconnects to the server with the same state that was used to prerender the app. If the app's state is still in memory, the component state isn't rerendered after the SignalR connection is established.

This paragraph is obsolete. We don't do "stateful" prerendering anymore, we connect from the client to the server and produce another render after prerendering.

Could we get this fixed?

https://github.com/aspnet/AspNetCore/issues/15266

---
#### Document Details

* Content: [ASP.NET Core Blazor hosting models](https://docs.microsoft.com/en-us/aspnet/core/blazor/hosting-models?view=aspnetcore-3.0#stateful-reconnection-after-prerendering)
* Content Source: [aspnetcore/blazor/hosting-models.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/hosting-models.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

### LINKED ISSUE #15271: [Blazor] Provide guidance about prerendering and flashing UI
When the app prerenders it waits until the first meaningful paint to produce the HTML.
When the browser connects to the client it produces a render as fast as possible, without waiting for the content to be retrieved from async sources like the database.
This results in a visible flash in the UI where the prerendered data disappears and the component renders something else until the data comes back and the first meaningful client-side paint occurs.

We should provide customers with guidance on how to design/build their applications so that they can avoid this flashing behavior.

guardrex:
Fixes #15270
Fixes #15271

We probably need to work on the language here. I'm :ear: on how to improve it.


guardrex:
@javiercn Dan might be 🏃 ...... can u approve/provide feedback?

aspnetcore/blazor/hosting-models.md
```diff
@@ -161,8 +161,67 @@ Blazor Server apps are set up by default to prerender the UI on the server befor
 
 Rendering server components from a static HTML page isn't supported.
 
+::: moniker range=">= aspnetcore-3.1"
```
danroth27: This content should apply equally to ASP.NET Core 3.0.

aspnetcore/blazor/hosting-models.md
```diff
@@ -161,8 +161,67 @@ Blazor Server apps are set up by default to prerender the UI on the server befor
...
+
+When `RenderMode` is `ServerPrerendered`, the renderer waits until the client has completed the intial render of static HTML before rendering the final component. If a [lifecycle method](xref:blazor/components#lifecycle-methods) for initializing the component (`OnInitialized{Async}`) is present, the method is executed *twice*. The first render is made before the final data is received. The second render occurs after the final data is available. This can result in a noticeable change in the data displayed in the UI when the component is finally rendered.
```
danroth27:
```suggestion
When `RenderMode` is `ServerPrerendered`, the component is initially rendered statically as part of the page. Once the browser establish a connection back to the server, the component is rendered *again* and the component is now interactive. If a [lifecycle method](xref:blazor/components#lifecycle-methods) for initializing the component (`OnInitialized{Async}`) is present, the method is executed *twice*: once when the component is prerendered statically, and once after the server connection has been established. This can result in a noticeable change in the data displayed in the UI when the component is finally rendered.
```

guardrex:
I versioned it for 3.0, but that also drops this line out of the coverage ...

> The client reconnects to the server with the same state that was used to prerender the app. If the app's state is still in memory, the component state isn't rerendered after the SignalR connection is established.

... is that line still relevant in the context of what we're adding? Should I simply add that back at the end?

danroth27:
That line should be removed, because it refers to an older implementation that we never shipped in a stable release. What you have in the PR looks correct.

guardrex:
Ping @javiercn ... we can't merge this one until you sign off.

PR #15294, opened 2019-10-24
Remove extra parentheses

imba-tjd:
From https://github.com/aspnet/AspNetCore.Docs.zh-cn/pull/199


Rick-Anderson:
Thanks for fixing this. The fix should be live in 24-48 hours.

PR #15326, opened 2019-10-26
Fix TypeScript target snippet in Razor UI docs

peterblazejewicz:
This fixes rendering of the configuration for TypeScript integration.

Thanks!

For context:
https://docs.microsoft.com/en-us/aspnet/core/razor-pages/ui-class?view=aspnetcore-3.0&tabs=visual-studio#feedback

image


peterblazejewicz:
Thanks! Was not aware about the date in the pragma!

PR #15354, opened 2019-10-28
add Interceptors in doc in Default service container replacement

EngRajabi:
add Interceptors in doc in Default service container replacement

### DIFF
--- a/aspnetcore/fundamentals/dependency-injection.md
+++ b/aspnetcore/fundamentals/dependency-injection.md
@@ -603,6 +603,7 @@ public void ConfigureServices(IServiceCollection services)
 The built-in service container is designed to serve the needs of the framework and most consumer apps. We recommend using the built-in container unless you need a specific feature that the built-in container doesn't support, such as:
 
 * Property injection
+* Interceptors
 * Injection based on name
 * Child containers
 * Custom lifetime management


guardrex:
@anurse ... those sound :ear: pretty cool ... **_Interceptors!_** 🚀 :smile: lol

Seriously tho ... best to check with you while we're here. Are there any changed that you would like to make to ...

> The built-in service container is designed to serve the needs of the framework and most consumer apps. We recommend using the built-in container unless you need a specific feature that the built-in container doesn't support, such as:
>
> * Property injection
> * Interceptors
> * Injection based on name
> * Child containers
> * Custom lifetime management
> * `Func<T>` support for lazy initialization

anurse:
Since this is a list of things **other** containers provide that ours does not, it's technically an unknowable set :). I guess if another DI container has something called Interceptors it's worth putting in?

guardrex:
@EngRajabi ... What is an 🚀 **_Interceptor_** :rocket: ...... besides a type of spacecraft used to engage rapidly advancing enemy ships? :smile: lol ........... (you gotta have some fun around here lol)

Anyway ... just curious what it is? Do you have a link?

guardrex:
@EngRajabi Since we're not really sure what that is and this is isn't going to ever be a complete list of every possible scenario in 3rd party containers that ASP.NET Core's container offers, I think we should limit the list to limit the length of the topic. Thanks tho for submitting it, and we look forward to hearing more feedback from you.

Issue #15370, opened 2019-10-29
Add an example for providing a pre-built configuration to Host


UncleDave:
As far as I can tell there is currently no example which shows how to:

1. Clear out the default configuration sources added by `CreateDefaultBuilder`.
2. Add a pre-built `IConfiguration` via `AddConfiguration`.

As raised in [this StackOverflow question](https://stackoverflow.com/questions/58530942/how-to-read-configuration-settings-before-initializing-a-host-in-asp-net-core/58594026).

If this is something that would be beneficial to the docs I'd be happy to open a PR.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0eb4d762-3910-2c57-8741-3306c15537b6
* Version Independent ID: ba2254a0-ea8e-9b28-e94b-edd0fc58c043
* Content: [Configuration in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-3.0#feedback)
* Content Source: [aspnetcore/fundamentals/configuration/index.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/fundamentals/configuration/index.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-fundamentals**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Hello @UncleDave ... Thanks for the comment, but we're not leaning in that direction (i.e., showing *every* way to handle configuration. We are being a little bit *opinionated* about what's shown, especially for the goal of not overloading new devs and devs new to ASP.NET Core. Let's take those separately ...

### Clear providers

Devs can just clear them. Otherwise, devs build their own host and just don't call `CreateDefaultBuilder`. Of course, the dev can just ignore them (i.e., if they aren't using settings files ... it just no-ops on settings files; not using command-line args ... again ... no-ops; not using Secret Manager ... no-ops; etc.). These are fairly straightforward approaches.

```csharp
.ConfigureAppConfiguration((hostingContext, config) =>
{
    config.Sources.Clear();
    // New config providers here
})
```

[...]

PR #15376, opened 2019-10-29
Add an example for clearing default configuration sources

UncleDave:
Addresses #15370

Added an example that shows how to clear the default configuration sources provided by `Host.CreateDefaultBuilder` to the _Default configuration_ section.

### DIFF
--- a/aspnetcore/fundamentals/configuration/index.md
+++ b/aspnetcore/fundamentals/configuration/index.md
@@ -5,7 +5,7 @@ description: Learn how to use the Configuration API to configure an ASP.NET Core
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 10/24/2019
+ms.date: 10/29/2019
 uid: fundamentals/configuration/index
 ---
 # Configuration in ASP.NET Core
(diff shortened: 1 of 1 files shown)


aspnetcore/fundamentals/configuration/index.md
```diff
@@ -75,11 +75,13 @@ The following applies to apps using the [Generic Host](xref:fundamentals/host/ge
 To remove the providers added by `CreateDefaultBuilder`, call `Sources.Clear()` first:
 
 ```csharp
-.ConfigureAppConfiguration((hostingContext, config) =>
-{
-    config.Sources.Clear();
-    // Add providers here
-})
+Host.CreateDefaultBuilder(args)
```
UncleDave: I considered that it's a bit context-less in this section :D Whereas the context was clearer down in the _ConfigureAppConfiguration_ section.

UncleDave:
@guardrex Other than the difference of `Host` vs `WebHost` is this not just as relevant to _<3.0_?

guardrex:
Yes, `ConfigureAppConfiguration` was a thing back in 2.1, so we're good there. I only did that as a prelude to perhaps having `WebHost.CreateDefaultBuilder` in the <3.0 piece in the *Default configuration* section, but I agree with u ... this is better the way that you had it and it's better down here. Sorry for the churn.

UncleDave:
All good, I think we needed to move it up in order to realise it was better off down below with the rest of the _ConfigureAppConfiguration_ stuff.

guardrex:
These docs are *complicated!* :smile: ..... I'm a small green dinosaur in a **BIG** framework here. :smile:

guardrex:
The main thing in my head right now is the "flow" concept. I think that requires more work in a general way ... it has to be a little less *implied* and a little more *explicit* about the **_flow_** of configuration. I'll take a look at cross-links, too. I'll consider floating the manual host builder with a built config, but I also have that feel'in that engineering isn't going to like that too much. We'll see. I'll ping u on that PR when it goes up. Thanks again for the discussion and your help with the docs.

PR #15377, opened 2019-10-29
Fix API link

guardrex:
The build system choked on the xref version.


guardrex:
That's better! :+1: :tada:

PR #15409, opened 2019-10-31
Update CustomValidationAttributeAdapterProvider.cs - Use is and cast operator

Marcel0024:
Use is operator to check and cast attribute


scottaddie:
Thank you for the contribution, @Marcel0024!

PR #15444, opened 2019-11-01
Damienbod/aspnetcore/security/authorization/roles.md update 3.0

damienbod:
Update the role.md to .NET Core 3.0 and maintain the 2.2 code snippets

https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/security/authorization/roles.md

@scottaddie @Rick-Anderson


Rick-Anderson:
@HaoK minor 3.0 update, can you approve?

aspnetcore/security/authorization/roles/samples/3_0/Startup.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        // This method gets called by the runtime. Use this method to add services to the container.
+        #region snippet
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddDbContext<ApplicationDbContext>(options =>
+                options.UseSqlServer(
+					Configuration.GetConnectionString("DefaultConnection")));
+            services.AddIdentity<IdentityUser, IdentityRole>()
+                .AddRoles<IdentityRole>()
```
serpent5: I don't think `AddRoles` is needed here as this is already handled using `AddIdentity`, but I do see that the section is about adding this call. Should this just use `AddDefaultIdentity` too, as in the 2.2 version?:

```csharp
services.AddDefaultIdentity<IdentityUser>()
    .AddRoles<IdentityRole>()
    .AddEntityFrameworkStores<ApplicationDbContext>();
```

aspnetcore/security/authorization/roles/samples/3_0/Startup.cs
```diff
@@ -0,0 +1,69 @@
...
+            app.UseHttpsRedirection();
+            app.UseStaticFiles();
+
+            app.UseRouting();
+
+            app.UseAuthorization();
+
+            app.UseEndpoints(endpoints =>
+            {
+                endpoints.MapControllerRoute(
```
serpent5: Could this just be `endpoints.MapDefaultControllerRoute();`?

aspnetcore/security/authorization/roles/samples/3_0/Startup.cs
HaoK: `AddIdentityCore` perhaps?  altho then that requires calling add authentication.  Or maybe remove the call to `AddDefaultUI` instead

damienbod: I updated this to the fix from @serpent5 , Is this ok?

aspnetcore/security/authorization/roles/samples/3_0/Startup.cs
damienbod: fixed

Rick-Anderson:
@HaoK  quick review when you have time, no hurry.

PR #15462, opened 2019-11-02
Remove link the SMS 2FA, no longer recommended

damienbod:
Remove link the SMS 2FA, SMS 2FA is no longer recommended

https://docs.microsoft.com/en-us/aspnet/core/security/authentication/individual?view=aspnetcore-3.0


Rick-Anderson:
@damienbod  thanks

PR #15476, opened 2019-11-03
Derive status from allocated < options.Threshold

lol768:
Entirely possible I've missed something obvious, but when trying this sample as-is I noticed it seemed to always report the failure state (degraded) even when the memory threshold hadn't been hit.


guardrex:
That's strange. When I upgrade these, I make an exact copy of the previous version, so this should just be a copy of the 2.x sample at ...

https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/host-and-deploy/health-checks/samples/2.x/HealthChecksSample/MemoryHealthCheck.cs

... I'll research it and see how this might have happened.

guardrex:
Good :eye: @lol768 ... it was a blown update on review at ...

https://github.com/aspnet/AspNetCore.Docs/pull/13946#discussion_r318727260

My bad! 🤦‍♂ Thanks for sending this in.

PR #15489, opened 2019-11-04
Respnose caching enhancements

### LINKED ISSUE #14643: Allowed location value is wrong
Location and Duration
To enable caching, Duration must be set to a positive value and Location must be either Any (the default) or Client. In this case, the Cache-Control header is set to the location value followed by the max-age of the response.

But caching doesn't work with Location set to Client. This page https://docs.microsoft.com/en-us/aspnet/core/performance/caching/middleware?view=aspnetcore-2.2 under *Conditions for caching* also says the following: "Cache-Control header parameters must be valid, and the response must be marked public and not marked private"

---
#### Document Details

* ID: b6728e59-93ec-891e-0416-94819296ad72
* Version Independent ID: 5d581acc-c651-2112-844a-6b85dc5a99a6
* Content: [Response caching in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/response?view=aspnetcore-2.2#feedback)
* Content Source: [aspnetcore/performance/caching/response.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/performance/caching/response.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

guardrex:
Fixes #14643

* I think it's low cost to flesh it out using your language.
* A little more is done here to make sure that it's clear that clients and proxies are *supposed to follow the spec*.
* ... because I see a misspelled word in a doc that I wrote, the 👽 invasion is about to begin! *That's the embedded signal they use in preparation for an attack!* ... I'll head that off by fixing it. :smile:

### DIFF
--- a/aspnetcore/performance/caching/middleware.md
+++ b/aspnetcore/performance/caching/middleware.md
@@ -20,7 +20,7 @@ This article explains how to configure Response Caching Middleware in an ASP.NET
 
 ::: moniker range=">= aspnetcore-3.0"
 
-Response Caching Middleware is made available by the [Microsoft.AspNetCore.ResponseCaching](https://www.nuget.org/packages/Microsoft.AspNetCore.ResponseCaching/) package, which is implicitly added to ASP.NET Core apps.
+Response Caching Middleware is implicitly available for ASP.NET Core apps via the shared framework.
 
 In `Startup.ConfigureServices`, add the Response Caching Middleware to the service collection:
 
--- a/aspnetcore/performance/caching/middleware.md
+++ b/aspnetcore/performance/caching/middleware.md
@@ -5,7 +5,7 @@ description: Learn how to configure and use Response Caching Middleware in ASP.N
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 08/09/2019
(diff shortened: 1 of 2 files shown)


aspnetcore/performance/caching/middleware.md
```diff
@@ -20,7 +20,7 @@ This article explains how to configure Response Caching Middleware in an ASP.NET
 
 ::: moniker range=">= aspnetcore-3.0"
 
-Response Caching Middleware is made avaiable by the [Microsoft.AspNetCore.ResponseCaching](https://www.nuget.org/packages/Microsoft.AspNetCore.ResponseCaching/) package, which is implicitly added to ASP.NET Core apps.
+Response Caching Middleware is made available by the [Microsoft.AspNetCore.ResponseCaching](https://www.nuget.org/packages/Microsoft.AspNetCore.ResponseCaching/) package, which is implicitly added to ASP.NET Core apps.
```
anurse: I'd rephrase this in the 3.0+ moniker range. The package doesn't exist anymore (notice there's no 3.0+ versions), it's part of the `AspNetCore.App` framework. Maybe just:

```suggestion
The Response Caching middleware is available in all ASP.NET Core apps.
```

Issue #15490, opened 2019-11-04
Blazor Server auth properties


guardrex:
Revert the API endpoint/tokens content when engineering resolves auth properties in `GetAuthenticationStateAsync`.

Engineering issue: https://github.com/aspnet/AspNetCore/issues/16831

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: c9942b0a-a7fb-0d34-9201-af0963affa9e
* Version Independent ID: 19770710-5225-a1ea-e082-65da6ba4a49a
* Content: [ASP.NET Core Blazor authentication and authorization](https://docs.microsoft.com/en-us/aspnet/core/security/blazor/?view=aspnetcore-3.0&tabs=visual-studio)
* Content Source: [aspnetcore/security/blazor/index.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/security/blazor/index.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-security**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
This one has been resolved by a few PRs along the way since it was opened back in November.

PR #15491, opened 2019-11-04
Additional claims enhancements

### LINKED ISSUE #15445: Claims added to the user on first registration
Updates will be worked for the linked topic and https://docs.microsoft.com/aspnet/core/security/blazor.

* Claims are added to the user on first registration, not on signin.
* Cross-link this at the Blazor auth topic, but note that an API endpoint for tokens will be required in that scenario.

See:

* https://github.com/aspnet/AspNetCore.Docs/issues/15325#issuecomment-548928837
* https://github.com/aspnet/AspNetCore.Docs/issues/15160

---
#### Document Details

* Content: [Persist additional claims and tokens from external providers in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/social/additional-claims?view=aspnetcore-3.0)
* Content Source: [aspnetcore/security/authentication/social/additional-claims.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/security/authentication/social/additional-claims.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #15445

Opened **Revert guidance on auth properties** (#15490) to cover reverting the auth prop content when the engineering is resolved.


aspnetcore/security/authentication/social/additional-claims.md
```diff
@@ -78,6 +78,8 @@ If a large amount of user data is required for processing user requests:
 * Limit the number and size of user claims for request processing to only what the app requires.
 * Use a custom <xref:Microsoft.AspNetCore.Authentication.Cookies.ITicketStore> for the Cookie Authentication Middleware's <xref:Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationOptions.SessionStore> to store identity across requests. Preserve large quantities of identity information on the server while only sending a small session identifier key to the client.
 
+Claims are added to the user on first registration, not on sign in. If additional claims are enabled in an app after a user registers to use the app, delete the user and have them register again.
```
Tratcher:
```suggestion
Claims are added to the user on first registration, not on sign in. If additional claims are enabled in an app after a user registers to use the app they won't be available.
```
Deleting and recreating a user is only viable in a dev environment. @HaoK how do you update external user claims at signin?

aspnetcore/security/blazor/index.md
```diff
@@ -115,6 +115,12 @@ Add a package reference for [Microsoft.AspNetCore.Components.Authorization](http
 
 Implementation of a custom `AuthenticationStateProvider` service for Blazor WebAssembly apps is covered in the following sections.
 
+### Additional claims and tokens from external providers
+
+To store additional claims from external providers, see <xref:security/authentication/social/additional-claims>.
+
+Currently, `GetAuthenticationStateAsync` hasn't been extended to include authentication properties. Therefore, the guidance in the *Save the access token* and *How to add additional custom tokens* sections of the [Additional claims](xref:security/authentication/social/additional-claims#save-the-access-token) topic can't be followed for Blazor apps. Until the framework supports authentication token access in Blazor apps, you can add an API endpoint to the app that returns the tokens for the current user.
```
Tratcher: Hold off on this, Javier had some feedback on https://github.com/aspnet/AspNetCore/issues/16831#issuecomment-549941535

aspnetcore/security/authentication/social/additional-claims.md
HaoK: You can probably just call ReplaceClaim on the user for claims stored in the identity database https://github.com/aspnet/AspNetCore/blob/master/src/Identity/Extensions.Core/src/UserManager.cs#L1107

HaoK: you would have to also call RefreshSign in to force a new cookie

guardrex: We can work on the text directly in the suggestion :point_up: until it's correct and then add it.

aspnetcore/security/authentication/social/additional-claims.md
```diff
@@ -218,6 +220,8 @@ If a large amount of user data is required for processing user requests:
...
```
guardrex:
```suggestion
Claims are added to the user on first registration, not on sign in. If additional claims are enabled in an app after a user registers to use the app:

* Call [UserManager.ReplaceClaimAsync](xref:Microsoft.AspNetCore.Identity.UserManager*) on the user for claims stored in the identity database.
* Call [SignInManager.RefreshSignInAsync](xref:Microsoft.AspNetCore.Identity.SignInManager*) to force the generation of a new authentication cookie.
```

aspnetcore/security/authentication/social/additional-claims.md
Tratcher: It needs to say where to call these.

guardrex: @HaoK I checked our docs hoping that I could cross-link to content explaining the pattern. AFAICT, we don't have general guidance for this scenario.

The silent re-sign in with `RefreshSignInAsync` isn't explicitly covered, but that's easy to grok from samples where a user updates their profile, changes their password, etc.

Replacing the claim:

1. Isn't "adding" performed with `AddClaimAsync`? The scenario that I envision thus far is that there is a new claim that the dev wants to take from their social auth provider.
1. What would you like to show for adding/changing claims? ... a `IUserClaimsPrincipalFactory` that runs on users?
1. What if the dev prefers instead a utility approach that they can run across all of the users at once? ... also something that I don't think docs show (but I admit that I'm using search to look for things and not reading :eyes: every doc we have).

guardrex: @HaoK ... I need additional guidance ...

Paraphrased ...

> You can call `ReplaceClaimAsync` on the user for claims stored in the identity database followed by `RefreshSignInAsync`.

I'm not sure who "you" is in this context or why it's a "replacement" when the claim is new for the app.

I envision a scenario where users are already registered. The dev wants to add a new claim provided by the social auth provider to their app. I wondered if the dev should create an `IUserClaimsPrincipalFactory` that calls `AddClaimAsync` if the claim isn't already present.

It's best to ask before going down that road with a test app because if it's the wrong approach, it will save some 🕐 ... thus 💰.

HaoK: So the intent of refresh sign in, is that you specify what user you wish the regenerate the claims for, it doesn't have any safeguard to enforce that you are refreshing the user you originally generated the cookie from.  Basically it will recreate the claims for that user and call sign in with them but also will attempt to preserve the original metadata stored in the cookie (which preserves things like remember me)

See https://github.com/aspnet/AspNetCore/blob/master/src/Identity/Core/src/SignInManager.cs#L169

HaoK: You refers to the developer who wants to update a claim for the user, the app could on subsequent external logins call replace claim with the updated value of a claim.  Does that make sense?  It shouldn't be necessary for them to have a custom principal factory for this.

guardrex:
This is going to require a new section in the topic because a couple of sentences aren't going to cover it well. I'll circle back around to the issue with a new PR either before or after the 🎁 holidays ☃️ depending on workload.

Issue #12285, opened 2019-05-04
File Uploads topic enhancements


guardrex:
It won't be possible to address one of the issues in the updates for https://github.com/aspnet/AspNetCore.Docs/issues/10214. 

The following opportunities remain:

* There's no example for handling file Chunking.
* There's no example for Byte Range when downloading.
* Provide help/examples/tips for resuming large file uploads when users were in remote areas with flaky Internet. Ditching streaming large files and going with chunking instead. Differences; pros and cons; terminology explained; common pitfalls.
* RE: Chunked transfer encoding: HTTP/2 apparently doesn't support it ... it goes to DATA frames. Should DATA frames coverage be explained/demoed?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 9a5cfa38-0e6c-4346-dafc-84ea5c513fcf
* Version Independent ID: c6787f0f-3e9a-6efb-0e6e-5ccb957d3f4a
* Content: [File uploads in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/file-uploads?view=aspnetcore-2.2#feedback)
* Content Source: [aspnetcore/mvc/models/file-uploads.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/mvc/models/file-uploads.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-mvc**
* GitHub Login: @ardalis
* Microsoft Alias: **riande**

nemtajo:
I am trying to make File Upload For Large Files work in my ASP.NET Core 3.0 Preview project. I followed the tutorial and I am running my project on localhost at the moment. After clicking on Upload button on the razor page, I get:

The SPA default page middleware could not return the default page '/index.html' because it was not found, and no other middleware handled the request.

Maybe the default VS template doesn't have all the necessary configuration in Startup.cs.
I saw that many developers reported this error before:
https://github.com/aspnet/JavaScriptServices/issues/1514
https://github.com/aspnet/AspNetCore/issues/5223
https://github.com/dotnet/core/issues/1803

It would be great if we the examples from github were updated, to have views with .razor extension and ASP.NET Core 3.0 structure:
WebApp.Client
WebApp.Server
WebApp.Shared

[...]

PR #15493, opened 2019-11-04
Stream was too long

### LINKED ISSUE #15337: Stream was too long - Stream a file with AJAX to a controller endpoint (physical)
Trying to upload a file (2,7GB), but I get an `Stream was too long` exception caused by the code below:

```
using (var memoryStream = new MemoryStream())
                {
                    await section.Body.CopyToAsync(memoryStream);
```
Which of course copies the whole file into a in-memory stream.

**Which setting determines if a stream is too long?**

**Overall memory consumption:**
Before upload-start I have ~10GB of free RAM available. SO memory is not a problem.

image

**Memory consumption IISExpress.exe:**

image

---
#### Document Details

* Version Independent ID: c6787f0f-3e9a-6efb-0e6e-5ccb957d3f4a
* Content: [Upload files in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/file-uploads?view=aspnetcore-3.0#server-and-app-configuration)
* Content Source: [aspnetcore/mvc/models/file-uploads.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/mvc/models/file-uploads.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #15337

Thanks @Legends. Given the workload, I don't recommend that we address the sample/topic directly at this time (except I spotted one that we can easily drop). Also, I note that we have chunking on the [File Uploads Tracking issue (aspnet/AspNetCore.Docs \#12285)](https://github.com/aspnet/AspNetCore.Docs/issues/12285) for future work. For now, let's call it out and see if additional community feedback comes in.

Pranav, can we do better than just saying, "use an alternative approach that doesn't rely upon a single `MemoryStream` for holding an uploaded file's content"? Can we offer more specific advice, or do you feel that merely calling this out is fine for the time being?

### DIFF
--- a/aspnetcore/mvc/models/file-uploads.md
+++ b/aspnetcore/mvc/models/file-uploads.md
@@ -5,7 +5,7 @@ description: How to use model binding and streaming to upload files in ASP.NET C
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 10/31/2019
+ms.date: 11/04/2019
 uid: mvc/models/file-uploads
 ---
 # Upload files in ASP.NET Core
@@ -735,6 +735,10 @@ A connection error and a reset server connection probably indicates that the upl
 
 If the controller is accepting uploaded files using <xref:Microsoft.AspNetCore.Http.IFormFile> but the value is `null`, confirm that the HTML form is specifying an `enctype` value of `multipart/form-data`. If this attribute isn't set on the `<form>` element, the file upload doesn't occur and any bound <xref:Microsoft.AspNetCore.Http.IFormFile> arguments are `null`. Also confirm that the [upload naming in form data matches the app's naming](#match-name-attribute-value-to-parameter-name-of-post-method).
 
+### Stream was too long
+
+The examples in this topic rely upon <xref:System.IO.MemoryStream> to hold the uploaded file's content. The size limit of a `MemoryStream` is `int.MaxValue`. If the app's file upload scenario requires holding file content larger than `int.MaxValue`, use an alternative approach that doesn't rely upon a single `MemoryStream` for holding an uploaded file's content.
+
 ::: moniker-end
 
 ::: moniker range="< aspnetcore-3.0"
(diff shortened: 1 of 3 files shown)


aspnetcore/mvc/models/file-uploads.md
```diff
@@ -735,6 +735,10 @@ A connection error and a reset server connection probably indicates that the upl
 
 If the controller is accepting uploaded files using <xref:Microsoft.AspNetCore.Http.IFormFile> but the value is `null`, confirm that the HTML form is specifying an `enctype` value of `multipart/form-data`. If this attribute isn't set on the `<form>` element, the file upload doesn't occur and any bound <xref:Microsoft.AspNetCore.Http.IFormFile> arguments are `null`. Also confirm that the [upload naming in form data matches the app's naming](#match-name-attribute-value-to-parameter-name-of-post-method).
 
+### Stream was too long
```
Legends: LGTM

guardrex:
Ping @pranavkm ... it's a quick one 🏃.

aspnetcore/mvc/models/file-uploads.md
```diff
@@ -735,6 +735,10 @@ A connection error and a reset server connection probably indicates that the upl
...
+
+The examples in this topic rely upon <xref:System.IO.MemoryStream> to hold the uploaded file's content. The size limit of a `MemoryStream` is `int.MaxValue`. If the app's file upload scenario requires holding file content larger than `int.MaxValue`, use an alternative approach that doesn't rely upon a single `MemoryStream` for holding an uploaded file's content.
```
pranavkm: That's a really large amount of content to be held in memory. We should say use an alternate approach if you know file sizes are larger than a much more reasonable value such as 50 MB or so.

guardrex:
@pranavkm I changed the recommendation there to "larger than 50 MB," but it keeps the fact that the actual cutoff is `int.MaxValue` given that it's an accurate remark (based on my research).

PR #15502, opened 2019-11-05
Update certauth.md

vit100:
Correction for method name where `AddAuthentication` should be added.


scottaddie:
Thank you for the PR, @vit100!

PR #15529, opened 2019-11-05
Update certauth.md

vit100:
Changed sample code in method BuildWebHost(). Replaced WebHost to Host as in versions of ASP.NET Core earlier than 3.0, the Web Host is used for HTTP workloads. The Web Host is no longer recommended for web apps and remains available only for backward compatibility.

### DIFF
--- a/aspnetcore/security/authentication/certauth.md
+++ b/aspnetcore/security/authentication/certauth.md
@@ -182,7 +182,7 @@ In *Program.cs*, configure Kestrel as follows:
 
 ```csharp
 public static IWebHost BuildWebHost(string[] args) =>
-    WebHost.CreateDefaultBuilder(args)
+    Host.CreateDefaultBuilder(args)
         .UseStartup<Startup>()
         .ConfigureKestrel(options =>
         {


Rick-Anderson:
@vit100  we need to maintain the previous version. You can use the following

```
// 3.0 code
::: moniker-end

// Original code here
::: moniker-end
```

vit100:
@Rick-Anderson This particular file is targeting ver 3 and higher only,  see line 5.
>monikerRange: '>= aspnetcore-3.0'

Should we change it this line as well to support versions before v3?

Rick-Anderson:
@vit100  I missed that - thanks.

Rick-Anderson:
@vit100  thanks for the PR

serpent5:
I think this needs a bit more work. The sample code wouldn't compile with this change:

1. `BuildWebHost` is declared to return `IWebHost` but returns `IHost`.
2. `UseStartup` and `ConfigureKestrel` don't exist for `IHostBuilder`. These calls need to go into a `ConfigureWebHostDefaults` callback.

serpent5:
Looks like #15534 has addressed the 2nd point above. I'll add the 1st point as a review comment to that PR.

Rick-Anderson:
@serpent5  can you PR any changes needed?

PR #15534, opened 2019-11-05
Update certauth.md

vit100:
Updated sample code for host configuration for v.3


aspnetcore/security/authentication/certauth.md
```diff
@@ -183,13 +183,13 @@ In *Program.cs*, configure Kestrel as follows:
 ```csharp
 public static IWebHost BuildWebHost(string[] args) =>
```
serpent5: This will have to return `IHost` now instead of `IWebHost`.

vit100: sec.. good catch. 
sec.

vit100: fixed.

PR #15548, opened 2019-11-06
Preivew 2 CLI

EdCharbeneau:
Updated CLI command for Preview 2 template


guardrex:
Thanks @EdCharbeneau, but we'd like to keep the short option there. *The shorter the CLI command the better.* We're trying not to spawn horizontal scrollbars in docs wherever possible. Also, the short options are in use all of the repo ... we use them all the time.

EdCharbeneau:
The version number is still incorrect. I'll submit another PR.

guardrex:
Probably because the last update hasn't gone live yet. I'll check. brb

guardrex:
Here's the update.

https://github.com/aspnet/AspNetCore.Docs/pull/15527/files

PR #15557, opened 2019-11-06
Fix spelling mistake

loic-sharma:


scottaddie:
Thank you for the PR, @loic-sharma!

PR #15560, opened 2019-11-06
Fixed version number for CLI install

EdCharbeneau:
CLI cmd should install Preview 2


guardrex:
It's already been updated. Should already be live. Might need to clear your browser cache.

https://github.com/aspnet/AspNetCore.Docs/pull/15527/files

PR #15598, opened 2019-11-08
fixing text to match code changes in certauth.md

damienbod:
Fix incorrect text to match code. Thumbprint is used to validate certificate, and not subject or issuer. fix from @blowdart


blowdart:
:) Someone pointed out stringtobytearray is a helper function you've written and they're confused. So might address that as well?

https://github.com/aspnet/AspNetCore.Docs/issues/15601

guardrex:
@damienbod, address #15601 if you like on a new PR for more GH issue credit. :tada:

PR #15599, opened 2019-11-08
Adding 3.x to the tabs display authorization-views.md

damienbod:
update aspnetcore/security/authorization/views.md to display the 3.x in the tab


guardrex:
I think that was done that way because it predates our versioning. I'm not even aware if those are approved tabs, but perhaps they are.

Do it this way .......

1. Add a metadata entry on Line 5 of ...

   ```
   monikerRange: '>= aspnetcore-2.1'
   ```

1. Drop the 1.x bits and the tab control. Just flow that 3.x/2.x content directly in the text (no controls and no versioning). The whole topic will be versioned for >=2.1 with the `monikerRange`, so it won't be necessary to version that 3.x/2.x content any longer.

1. btw -- Update the `ms.date` of the topic to today.

guardrex:
I've got a few minutes ... I'll take care of it. Thanks for the PR.

Issue #15601, opened 2019-11-08
What is StringToByteArray?


yshahin:
StringToByteArray is not a C# function, it is not defined in the docs.
It should be replaced with `Convert.FromBase64String(headerValue)`

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 35b7960d-6de5-78ee-bc45-d7218239bbc8
* Version Independent ID: d2db534b-198a-dc99-5b5a-0525582d39ac
* Content: [Configure certificate authentication in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/certauth?view=aspnetcore-3.0#configure-your-host-to-require-certificates)
* Content Source: [aspnetcore/security/authentication/certauth.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/security/authentication/certauth.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-security**
* GitHub Login: @blowdart
* Microsoft Alias: **bdorrans**

blowdart:
@Damienbod

damienbod:
It's actually required.  Will add it.

´´´
private static byte[] StringToByteArray(string hex)
        {
            int NumberChars = hex.Length;
            byte[] bytes = new byte[NumberChars / 2];
            for (int i = 0; i < NumberChars; i += 2)
                bytes[i / 2] = Convert.ToByte(hex.Substring(i, 2), 16);
            return bytes;
        }
´´´

Convert.FromBase64String(headerValue) would not work as it's not a base 64 string.

damienbod:
Added a PR for this.

guardrex:
Fixed by #15608

vit100:
something wrong with StringToByArray()
For example, this is my sample certificate which Azure set in header:
[...]

PR #15608, opened 2019-11-08
Adding missing static StringToByteArray method to certauth.md

damienbod:
fix for https://github.com/aspnet/AspNetCore.Docs/issues/15601


guardrex:
@damienbod ... Use the exact text ...

```
Fixes #{ISSUE NUMBER}
```

... `Fixes #15601` in this case.

... so that it will auto-close the issue when the PR merges.

Thanks again for sending these PRs in. 🎸:beers:

PR #15613, opened 2019-11-08
Adding no-loc metadata

JasonCard:
Adding no-loc metadata to prevent the term "SignalR" from being over-localized.

### DIFF
--- a/aspnetcore/signalr/scale.md
+++ b/aspnetcore/signalr/scale.md
@@ -7,6 +7,7 @@ ms.author: bradyg
 ms.custom: mvc
 ms.date: 11/28/2018
 uid: signalr/scale
+no-loc: [SignalR]
 ---
 
 # ASP.NET Core SignalR hosting and scaling


aspnetcore/signalr/scale.md
```diff
@@ -7,6 +7,7 @@ ms.author: bradyg
 ms.custom: mvc
 ms.date: 11/28/2018
 uid: signalr/scale
+no-loc: [SignalR]
```
anurse: Is there a way to do this more broadly? Across this whole repo, for example? The term "SignalR" will **always** be used to refer to a product so it should never be translated.

JasonCard: Unfortunately, there isn't a way to block a term from localization across the repo. We are working to improve the Machine Translation engine, but its trained on a quarterly basis, so adding this no-loc metadata per article is currently the best solution.

scottaddie: @JasonCard I was under the impression these "no-loc" terms could be defined in *docfx.json*. See the following example: https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/docfx.json#L57

guardrex: We added it on https://github.com/aspnet/AspNetCore.Docs/pull/15239. The word "Blazor" was showing up as "éblouissant" in French. Now, it's all :smile: ...

https://docs.microsoft.com/fr-fr/aspnet/core/blazor/?view=aspnetcore-3.0

guardrex: Oh ... no ... it didn't work for "Blazor" in French ...

https://docs.microsoft.com/fr-fr/aspnet/core/blazor/?view=aspnetcore-3.0

guardrex: In some places it seems to have worked, in others not.

JasonCard: Thank you for checking. As you also found, the "no-loc" extension is not supported as global metadata in docfx.json. The "no-loc" markdown extension is only supported by localization platform (OpenLoc) which transforms source md files into localizable format (xliff). OpenLoc does not reference the docfx.json file, so that's why the "no-loc" metadata is needed in each file. The idea to support this global metadata in docfx.json file was considered, but OPS team decided not to support it, so I'd suggest to remove it from there as well.

guardrex: Thanks @JasonCard ... I'll update and take care of that removal from *docfx.json* this weekend on #15614.

PR #15688, opened 2019-11-13
Expand 2.2->3.0 migration guidance

mjrousos:
I recently helped a customer migrate an ASP.NET Core 2.1 app to 3.0. Most of the breaks we ran into were documented, but a few were not. 

This PR adds some additional text around a few areas that were unclear to the customer based on current docs (changes to custom authorization handlers based on endpoint routing differences, URL.Link changes, and that AllowSynchronousIO settings must be made per-server (or per-request)).


Rick-Anderson:
@mjrousos  much appreciated!

PR #15691, opened 2019-11-14
Update Blazor linker docs to clarify that we always link on build

danroth27:


guardrex:
Ping @SteveSandersonMS

PR #15703, opened 2019-11-15
Mention SignalR Core 3.0's auto reconnect support

halter73:


aspnetcore/signalr/version-differences.md
```diff
@@ -28,7 +28,31 @@ ASP.NET Core SignalR isn't compatible with clients or servers for ASP.NET Signal
...
+    .withUrl("/chatHub")
+    .withAutomaticReconnect()
+    .build();
+```
+
+::: moniker-end
+
+::: moniker range="< aspnetcore-3.0"
+
+Automatic reconnects aren't supported in ASP.NET Core SignalR prior to ASP.NET Core 3.0. If the client is disconnected, the user must explicitly start a new connection if they want to reconnect. In ASP.NET SignalR, SignalR attempts to reconnect to the server if the connection is dropped.
```
scottaddie:
```suggestion
Automatic reconnects aren't supported in ASP.NET Core SignalR prior to ASP.NET Core 3.0. If the client is disconnected, the user must explicitly start a new connection to reconnect. In ASP.NET SignalR, SignalR attempts to reconnect to the server if the connection is dropped.
```

Rick-Anderson:
@scottaddie  check my commits. There's no such thing as drops by default.

PR #15736, opened 2019-11-18
update readme removing wrong text

ganchito55:
Remove misplaced words in the code example


scottaddie:
Thank you for fixing this, @ganchito55!

PR #15741, opened 2019-11-18
Add guidance for using ASP.NET Core APIs in libraries

scottaddie:
Fixes https://github.com/aspnet/AspNetCore.Docs/issues/14309
Fixes https://github.com/aspnet/AspNetCore.Docs/issues/15303

Add a doc focused on guidance for using ASP.NET Core APIs in a class library.

[Internal review page](https://review.docs.microsoft.com/aspnet/core/fundamentals/target-aspnetcore?branch=pr-en-us-15741)


scottaddie:
Pinging @davidfowl & @DamianEdwards for a review.

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,223 @@
...
+
+For example, synchronous reads and writes on HTTP request and response streams are disabled by default in ASP.NET Core 3.0. ASP.NET Core 2.2 supports the synchronous behavior by default. Consider a library in which synchronous reads and writes should be enabled in Kestrel. The library should enclose the code to enable synchronous features in the appropriate preprocessor directive. For example:
+
+```csharp
+public void ConfigureServices(IServiceCollection services)
+{
+    // code omitted for brevity
+
+#if NETCOREAPP3_0
+    services.Configure<KestrelServerOptions>(options =>
```
davidfowl: Libraries should never do this:
- It's too global
- It's kestrel specific

Instead, they should:
- Avoid sync IO
- Buffer the input or output (need code sample)
- AllowSyncIO where the IO is happening (e.g. https://github.com/aspnet/Benchmarks/blob/21e0050f6c14e0248e02b14395212fce5ae6890b/src/Benchmarks/Middleware/JsonMiddleware.cs#L46-L50)

scottaddie:
@davidfowl This is ready for another look.

scottaddie:
@davidfowl @DamianEdwards I'm hoping to get some more feedback before the holiday break.

scottaddie:
@twsouthwick You mentioned you're doing some work in this area. I'd love to hear your feedback as well.

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+If the project targets .NET Core 3.0, it requires:
+
+* An `AddRazorSupportForMvc` MSBuild property set to `true`.
+* A `<FrameworkReference>` element for the shared framework.
+
+For example:
+
+[!code-xml[](target-aspnetcore/samples/single-tfm/netcoreapp3.0-razor-views-pages-library.csproj)]
+
+If the project targets .NET Standard instead, it requires a [Microsoft.AspNetCore.Mvc](https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc) package reference. The `Microsoft.AspNetCore.Mvc` package moved into the shared framework and is therefore no longer published. For example:
```
twsouthwick: This recommendation references the 2.2 version of the app. It may be nice to have a bit of discussion about compatibility concerns when running on 3.x+

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+---
+# Use ASP.NET Core APIs in a class library
+
+By [Scott Addie](https://github.com/scottaddie)
+
+This document provides guidance for using ASP.NET Core APIs in a class library. For all other library guidance, see [Open-source library guidance](/dotnet/standard/library-guidance/).
+
+## Use the ASP.NET Core shared framework
+
+With the release of .NET Core 3.0, many ASP.NET Core packages are no longer published to NuGet. Instead, the packages are included in the `Microsoft.AspNetCore.App` shared framework, which is installed with the .NET Core SDK. The shared framework was previously distributed as a NuGet package. For a list of packages no longer being published, see [Remove obsolete package references](xref:migration/22-to-30#remove-obsolete-package-references).
```
DamianEdwards: "With the release of .NET Core 3.0, many ASP.NET Core **assemblies** are no longer published to NuGet **as packages**. Instead, the **assemblies** are included in the `Microsoft.AspNetCore.App` shared framework, which is installed with the .NET Core SDK **and runtime installers**. "

Just remove the following line as it's more nuanced than that and it doesn't really add anything in this case: "The shared framework was previously distributed as a NuGet package."

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+
+As of .NET Core 3.0, projects using the `Microsoft.NET.Sdk.Web` MSBuild SDK implicitly reference the shared framework. Projects using the `Microsoft.NET.Sdk` or `Microsoft.NET.Sdk.Razor` SDK must target ASP.NET Core to use ASP.NET Core APIs in the shared framework.
```
DamianEdwards: Replace "target" with "reference" I think. It's more consistent and will make the line further down less confusing too:
"... SDK must **reference** ASP.NET Core to use ASP.NET Core APIs in the shared framework."

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+
+To target ASP.NET Core, add the following `<FrameworkReference>` element to your project file:
```
DamianEdwards: "To **reference** ASP.NET Core, ..."

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+
+[!code-xml[](target-aspnetcore/samples/single-tfm/netcoreapp3.0-basic-library.csproj?highlight=8)]
+
+Targeting ASP.NET Core in this manner is only supported for projects targeting .NET Core 3.0.
```
DamianEdwards: "**Referencing** ASP.NET Core in this manner is only supported for projects targeting .NET Core 3.0."

aspnetcore/fundamentals/target-aspnetcore.md
DamianEdwards: I agree. While multi-targeting is discussed later on, this paragraph kinda forces the issue, as I don't expect most readers to know why or when they'd target `netcoreapp` or `netstandard` when including UI components in ASP.NET Core libraries.

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+
+If the project targets .NET Standard instead, it requires a [Microsoft.AspNetCore.Mvc](https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc) package reference. The `Microsoft.AspNetCore.Mvc` package moved into the shared framework and is therefore no longer published. For example:
+
+[!code-xml[](target-aspnetcore/samples/single-tfm/netstandard2.0-razor-views-pages-library.csproj?highlight=8)]
+
+### Razor components
+
+A project that includes [Razor components](xref:blazor/components) must use the [Microsoft.NET.Sdk.Razor SDK](xref:razor-pages/sdk).
+
+To support Razor component consumption from [Blazor Server](xref:blazor/hosting-models#blazor-server) only:
```
DamianEdwards: Maybe "Blazor Server apps" or "Blazor Server projects" instead of just "Blazor Server"?

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+
+* Target .NET Core 3.0.
+* Add a `<FrameworkReference>` element for the shared framework.
+
+For example:
+
+[!code-xml[](target-aspnetcore/samples/single-tfm/netcoreapp3.0-razor-components-library.csproj)]
+
+To support Razor component consumption from both Blazor Server and [Blazor WebAssembly](xref:blazor/hosting-models#blazor-webassembly):
```
DamianEdwards: Same as above.

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+
+* Target .NET Standard 2.0.
```
DamianEdwards: I believe Blazor WebAssembly is moving to .NET Standard 2.1, but best to check timelines with @danroth27

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+
+For more information on libraries containing Razor components, see [ASP.NET Core Razor components class libraries](xref:blazor/class-libraries).
+
+### Tag Helpers
+
+A project that includes [Tag Helpers](xref:mvc/views/tag-helpers/intro) should use the `Microsoft.NET.Sdk` SDK. If targeting .NET Core 3.0, add a `<FrameworkReference>` element for the shared framework. For example:
+
+[!code-xml[](target-aspnetcore/samples/single-tfm/netcoreapp3.0-basic-library.csproj)]
+
+If targeting .NET Standard, add a package reference to [Microsoft.AspNetCore.Mvc.Razor](https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Razor). The `Microsoft.AspNetCore.Mvc.Razor` package moved into the shared framework and is therefore no longer published. For example:
```
DamianEdwards: The statement "The `Microsoft.AspNetCore.Mvc.Razor` package moved into the shared framework and is therefore no longer published." could be confusing here. What is the thinking behind mentioning it again? Maybe whenever we say "If targeting .NET Standard," we should extend it to clarify the *why*, e.g. "If targeting .NET Standard (in order to provide support for versions earlier than ASP.NET Core 3.0).."

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+
+[!code-xml[](target-aspnetcore/samples/single-tfm/netstandard2.0-view-components-library.csproj)]
+
+## Support multiple ASP.NET Core versions
+
+Multi-targeting is required to author a library that supports multiple variants of ASP.NET Core. Consider a scenario in which a Tag Helpers library must support the following ASP.NET Core variants:
+
+* ASP.NET Core 2.1 targeting .NET Framework 4.6.1
+* ASP.NET Core 2.x targeting .NET Core 2.x
+* ASP.NET Core 3.0 targeting .NET Core 3.0
```
DamianEdwards: 3.0 is EOL soon. Change to "3.x"

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+
+The following project file supports these variants via the `TargetFrameworks` property:
+
+[!code-xml[](target-aspnetcore/samples/multi-tfm/recommended-tag-helpers-library.csproj)]
+
+With the preceding project file:
+
+* The `Markdig` package is added for all consumers.
+* A reference to [Microsoft.AspNetCore.Mvc.Razor](https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Razor) is added for consumers targeting .NET Framework 4.6.1 or .NET Core 2.x. Version 2.1.0 of the package works with ASP.NET Core 2.2 because of backwards compatibility.
```
DamianEdwards: "... .NET Framework 4.6.1 **or higher**, or .NET Core 2.x."

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+* The shared framework is referenced for consumers targeting .NET Core 3.0. The `Microsoft.AspNetCore.Mvc.Razor` package is included in the shared framework.
+
+Alternatively, .NET Standard 2.0 could be targeted instead of targeting both .NET Core 2.1 and .NET Framework 4.6.1:
+
+[!code-xml[](target-aspnetcore/samples/multi-tfm/alternative-tag-helpers-library.csproj?highlight=4)]
+
+With the preceding project file, .NET Core 2.x and .NET Framework 4.6.1 projects are supported because both target frameworks implement .NET Standard 2.0. Since this library only contains Tag Helpers, it's more efficient to target the specific platforms on which ASP.NET Core runs: .NET Core and .NET Framework. Tag Helpers can't be used by other .NET Standard 2.0-compliant target frameworks such as Unity, UWP, and Xamarin. If your library needs to call platform-specific APIs, target specific .NET implementations instead of .NET Standard. For more information, see [Multi-targeting](/dotnet/standard/library-guidance/cross-platform-targeting#multi-targeting).
```
DamianEdwards: Replace "efficient" with "straightforward".

Also, we generally recommend that class libraries directly target .NET Framework to support it, rather than .NET Standard, due to the issues of using .NET Standard targeted packages on .NET Framework (this is covered in the general library guidance too I believe).

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Use an API that hasn't changed
+
+Imagine a scenario in which you're upgrading a middleware library from .NET Core 2.2 to 3.0. The ASP.NET Core middleware APIs being used in the library haven't changed between ASP.NET Core 2.2 and 3.0. To continue supporting the middleware library in .NET Core 3.0, take the following steps:
+
+* Follow the [standard library guidance](/dotnet/standard/library-guidance/).
+* Add a package reference for the API's NuGet package if the assembly doesn't exist in the shared framework.
```
DamianEdwards: We should talk about what **version** of the package to reference whenever we say this, or have a section that can be referred to that does.

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+}
+```
+
+The following multi-targeted project file supports this Tag Helper scenario:
+
+[!code-xml[](target-aspnetcore/samples/multi-tfm/recommended-tag-helpers-library.csproj)]
+
+## Use an API removed from shared framework
+
+To use an ASP.NET Core assembly that was removed from the shared framework, add the appropriate package reference. For example, to add the web API client:
```
DamianEdwards: Can we link to the list of removed assemblies again here?

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+
+To target ASP.NET Core, add the following `<FrameworkReference>` element to your project file:
+
+[!code-xml[](target-aspnetcore/samples/single-tfm/netcoreapp3.0-basic-library.csproj?highlight=8)]
+
+Targeting ASP.NET Core in this manner is only supported for projects targeting .NET Core 3.0.
+
+## Include UI components
+
+The following sections outline recommendations for libraries that include UI components. This guidance assumes the library won't multi-target. For guidance on supporting multiple ASP.NET Core versions, see [Support multiple ASP.NET Core versions](#support-multiple-aspnet-core-versions).
```
rynowak: I think it would be great to pick a different term than `UI Components` given that we have a **Razor Components (Blazor)** and **View Components (MVC)**. It would behoove us to avoid a 3rd meaning of components.

Maybe decouple this into two sections? Include MVC extensibility + Include Blazor extensibility?

rynowak: Another suggestion here, I don't think `assumes the library won't multi-target` is very clear, It might be better to say `this section does not discuss multitargeting to support multiple versions of MVC. Read the multitargeting section to....`

aspnetcore/fundamentals/target-aspnetcore.md
rynowak: The guidance for Blazor should position targeting `netstandard2.0` as the best and most prominent option. Library authors should target both WASM and Server by default and only target a specific hosting environment if absolutely necessary.

aspnetcore/fundamentals/target-aspnetcore.md
rynowak: NS2.1 is current previews, so that's fine.

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+
+[!code-xml[](target-aspnetcore/samples/single-tfm/netcoreapp3.0-razor-components-library.csproj)]
+
+To support Razor component consumption from both Blazor Server and [Blazor WebAssembly](xref:blazor/hosting-models#blazor-webassembly):
+
+* Target .NET Standard 2.0.
+* Add the following package references:
+  * [Microsoft.AspNetCore.Components](https://www.nuget.org/packages/Microsoft.AspNetCore.Components)
+  * [Microsoft.AspNetCore.Components.Web](https://www.nuget.org/packages/Microsoft.AspNetCore.Components.Web)
+* Set the `RazorLangVersion` property to `3.0`. `3.0` is the default value for .NET Core 3.0.
```
rynowak: This section should be the default guidance, targeting only server or client is a special case.

This should also recommend using the RCL template - we literally have a template for this.

aspnetcore/fundamentals/target-aspnetcore/samples/middleware.cs
```diff
@@ -0,0 +1,57 @@
...
+
+namespace Benchmarks.Middleware
+{
+    public class JsonMiddleware
+    {
+        private static readonly PathString _path = new PathString(Scenarios.GetPath(s => s.Json));
+        private static readonly UTF8Encoding _encoding = new UTF8Encoding(false);
+        private const int _bufferSize = 27;
+#if !NETCOREAPP3_0 && !NETCOREAPP5_0
+        private static readonly JsonSerializer _json = new JsonSerializer();
```
rynowak: It seems really bizzare to me that this example is using the TFM to choose with JSON serializer to use. That might be what our benchmarking code does, but it makes for confusion when the example does something unnecessary.

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+title: Use ASP.NET Core APIs in a class library
+author: scottaddie
+description: Learn how to use ASP.NET Core APIs in a class library.
+ms.author: scaddie
+ms.custom: mvc
+ms.date: 12/03/2019
+no-loc: [Blazor]
+uid: fundamentals/target-aspnetcore
+---
+# Use ASP.NET Core APIs in a class library
```
rynowak: I feel like one thing that's missing here - a mention of our support lifecyle, and links to where library authors can find the current data about what's in support. 

Whenever I talk to community library authors I make it a point to bring this us. Most don't know that they are targeting release of ASP.NET Core that we've dropped support for - they are used to the old world where stuff lives forever.

Obviously we should boss people around, but making an effort to target all of the ASP.NET Core releases that are currently in support, and it's ok to ignore the ones that aren't seems like an OK suggestion.

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,213 @@
...
+
+By [Scott Addie](https://github.com/scottaddie)
```
rynowak: Another thing that seems missing-ish - how should a user keep tabs on breaking change announcments and new preview info?

I feel like there should be some basic call to action about the fact that we ship previews, and library authors can test against them as features are being developed.

aspnetcore/fundamentals/target-aspnetcore.md
danroth27: Yeah, we moved to NET Standard 2.1 with the last Blazor WebAssembly preview.

aspnetcore/fundamentals/target-aspnetcore.md
scottaddie: Agreed. In this context, I think customers associate the word "target" with a TFM.

aspnetcore/fundamentals/target-aspnetcore.md
scottaddie: Do you have a specific example in mind that I could show in this section?

scottaddie:
@DamianEdwards It's ready for another look. Also, see my question above.

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,280 @@
...
+
+This document provides guidance for using ASP.NET Core APIs in a class library. For all other library guidance, see [Open-source library guidance](/dotnet/standard/library-guidance/).
+
+## Determine which ASP.NET Core versions to support
+
+ASP.NET Core adheres to the [.NET Core support policy](https://dotnet.microsoft.com/platform/support/policy/dotnet-core). Consult the support policy when determining which ASP.NET Core versions to support in a library. A library should make an effort to:
+
+* Support all ASP.NET Core versions classified with *Long-Term Support* (LTS).
+* Ignore non-LTS ASP.NET Core versions, such as those classified with *End of Life* (EOL).
```
rynowak: This statement might be a little more restrictive than what I had in mind.

Possible rephrase: `Don't feel obligated to support ASP.NET Core versions classified as End of Life`

My thoughts:
 - We shouldn't tell people that it's **wrong** to support old versions. It's not really our problem if they want to spend their own time and effort on it. 
- I want library authors to be aware that EOL is a thing, find the relevant info, and then they can consider whether they want to support those versions.
- We definitely don't want to discourage library authors from targeting non-LTS releases, like 3.0. Library updates become adoption blockers for people to these releases really fast.

aspnetcore/fundamentals/target-aspnetcore.md
```diff
@@ -0,0 +1,280 @@
...
+
+To reference ASP.NET Core, add the following `<FrameworkReference>` element to your project file:
+
+[!code-xml[](target-aspnetcore/samples/single-tfm/netcoreapp3.0-basic-library.csproj?highlight=8)]
+
+Referencing ASP.NET Core in this manner is only supported for projects targeting .NET Core 3.x.
+
+## Include Blazor extensibility
+
+Blazor supports WebAssembly (WASM) and Server [hosting models](xref:blazor/hosting-models). Unless there's a specific reason not to, a [Razor components](xref:blazor/components) library should support both hosting models. A Razor components library must use the [Microsoft.NET.Sdk.Razor SDK](xref:razor-pages/sdk).
```
rynowak: 👍

aspnetcore/fundamentals/target-aspnetcore.md
DamianEdwards: And now I think about it more, his might just be **too complicated** to give general guidance on 😕  Let's ignore my original comment for now.

aspnetcore/fundamentals/target-aspnetcore.md
DamianEdwards: Agree with this.

aspnetcore/fundamentals/target-aspnetcore.md
scottaddie: Sounds good. If you change your mind, let me know. I'm sure we'll get some good feedback from the community as well. This doc would be worth sharing in the next community standup.

scottaddie:
@rynowak Ready for another look. Thank you soooo much for all of your feedback.

PR #15772, opened 2019-11-19
Clarify log streaming level setting

guardrex:
Fixes  #3279

There's a sweet spot for a quick call out that will resolve this for readers. It can be done right where the **Level** is set for this provider.


guardrex:
SMH ...... "Subject Matter Hackster" is probably more apt.

*I'm still trying to learn how all this stuff works!* 🤦‍♂:smile:

TehWardy:
If I understand this right this is a documentation change only?
I believe there is still a bug here as per my feedback here ...
   https://github.com/Azure/Azure-Functions/issues/2388
 
Hopefully my additional investigation here helps show a reproducible example of "unexpected" behavior.

PR #15776, opened 2019-11-19
Fix error message in UpdateItem function

eslam-fakhry:
Fix error message in UpdateItem function in web-api-javascript tutorial


scottaddie:
Thank you for fixing this, @eslam-fakhry!

Issue #11860, opened 2019-04-05
@section and RenderSection()


cincir01:
@section Scripts {
    @RenderSection("Scripts", required: false)
}

as far as I understand "@RenderSection" needs to be in Layout and @section needs to be in Razor Content Page. In this example they are together. Am I missing anything?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b29734d0-2d2b-3b82-0456-785cc7f361fd
* Version Independent ID: ab07129c-a4dd-44de-45f9-4b49eff0b54c
* Content: [Layout in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/mvc/views/layout?view=aspnetcore-2.2)
* Content Source: [aspnetcore/mvc/views/layout.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/mvc/views/layout.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-mvc**
* GitHub Login: @ardalis
* Microsoft Alias: **riande**

yogeshdotnet:
As you applied @RenderSection("Scripts", required: false)
 then this section may or may not be to add in razor view or pages and depends on your need, but if we apply required: true in @RenderSection method then it will be mandatory to apply on every page where this layout page is used.

cincir01:
That is clear and it is not what I am asking. If you go to Sections part of the document (https://docs.microsoft.com/en-us/aspnet/core/mvc/views/layout?view=aspnetcore-2.2#sections) , the first example is what is confusing to me.

cysun:
I think this is an error in the documentation that should be fixed. Having @RenderSection() inside @section doesn't seem to cause build errors, but as an example in the documentation it is at least confusing if not wrong.

kosta-arnorsky:
> as an example in the documentation it is at least confusing if not wrong
It's not wrong and might be even useful in certain cases, but it's defiantly confusing, especially in an introductory article, because it has **both** RenderSection and section definition.
As I said, it's not wrong, but may create a fallacious impression that RenderSection can be used only in a section.

Rick-Anderson:
@kosta-arnorsky  thanks for fixing. I can't think of good use case for what we had.

Rick-Anderson:
Fixed by #15804

PR #15804, opened 2019-11-21
Fixes issue #11860 - confusing RenderSection example

kosta-arnorsky:
As noted in the issue #11860, the RenderSection example is confusing and may create a fallacious impression that RenderSection can be used only inside other section definition. It's overloaded with information that is unnecessary for RenderSection understanding.
Also more appropriate tilde (~) notation/operator is used for resources paths.


Rick-Anderson:
@kosta-arnorsky  thanks!

PR #15810, opened 2019-11-21
Inherit from Controller

vanillajonathan:
Controllers using `ActionResult` must inherit from the `Controller` class.


scottaddie:
Thank you for fixing the code sample, @vanillajonathan!

PR #15818, opened 2019-11-21
Adding another scenario for dotnet process launch failures if the logging is in correctly enabled

fasigpt:

### DIFF
--- a/aspnetcore/test/troubleshoot-azure-iis.md
+++ b/aspnetcore/test/troubleshoot-azure-iis.md
@@ -151,7 +151,7 @@ Enable logging the application process' stdout messages
 Attach a debugger to the application process and inspect
 ```
 
-The best practise is to enable logging to a path outside your application folder, example \\?\%home%\LogFiles\stdout . 
+The best practise is to enable logging to a path outside your application folder, example \\\?\%home%\LogFiles\stdout . 
 
 ### 500.32 ANCM Failed to Load dll
 
@@ -255,7 +255,7 @@ Check the system event log for error messages
 Enable logging the application process' stdout messages
 Attach a debugger to the application process and inspect
 ```
-The best practise is to enable logging to a path outside your application folder, example \\?\%home%\LogFiles\stdout . 
+The best practise is to enable logging to a path outside your application folder, example \\\?\%home%\LogFiles\stdout . 
 
 ## Troubleshoot on Azure App Service
 
--- a/aspnetcore/test/troubleshoot-azure-iis.md
+++ b/aspnetcore/test/troubleshoot-azure-iis.md
@@ -137,6 +137,20 @@ The error message lists all the installed .NET Core versions and the version req
 
 When running in development (the `ASPNETCORE_ENVIRONMENT` environment variable is set to `Development`), the specific error is written to the HTTP response. The cause of a process startup failure is also found in the Application Event Log.
 
+This error can also occur if the asp.net core app is using in-process hosting model and aspnetcore module logging is enabled through the stdoutLogFile switch to log the files to WWWROOT folder (or the application content folder). The best practise is to enable logging to a path outside your application folder, example \\?\%home%\LogFiles\stdout . 
+
+For Azure App Service, you might see an Exception as follows
+
+'''
+HTTP Error 500.31 - ANCM Failed to Find Native Dependencies
+Common solutions to this issue:
+The specified version of Microsoft.NetCore.App or Microsoft.AspNetCore.App was not found.
+Troubleshooting steps:
+Check the system event log for error messages
+Enable logging the application process' stdout messages
+Attach a debugger to the application process and inspect
+'''
+
 ### 500.32 ANCM Failed to Load dll
 
 The worker process fails. The app doesn't start.
--- a/aspnetcore/test/troubleshoot-azure-iis.md
+++ b/aspnetcore/test/troubleshoot-azure-iis.md
@@ -137,11 +137,11 @@ The error message lists all the installed .NET Core versions and the version req
 
 When running in development (the `ASPNETCORE_ENVIRONMENT` environment variable is set to `Development`), the specific error is written to the HTTP response. The cause of a process startup failure is also found in the Application Event Log.
 
-This error can also occur if the asp.net core app is using in-process hosting model and aspnetcore module logging is enabled through the stdoutLogFile switch to log the files to WWWROOT folder (or the application content folder). The best practise is to enable logging to a path outside your application folder, example \\?\%home%\LogFiles\stdout . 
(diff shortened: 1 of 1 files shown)


fasigpt:
@jkotalik please review and approve

aspnetcore/test/troubleshoot-azure-iis.md
```diff
@@ -137,6 +137,22 @@ The error message lists all the installed .NET Core versions and the version req
 
 When running in development (the `ASPNETCORE_ENVIRONMENT` environment variable is set to `Development`), the specific error is written to the HTTP response. The cause of a process startup failure is also found in the Application Event Log.
 
+This error can also occur if the asp.net core app is using in-process hosting model, and aspnetcore module logging is enabled through the stdoutLogFile, to log the stdout files to WWWROOT folder (or the application content folder).
```
jkotalik: I think this can be condensed a bit to say "This error can also occur if there is an error with configuration, for example, if there is an invalid path to a log folder." The information below is redundant IMO.

Also, can you give more details about the wwwroot issue? I don't necessarily believe wwwroot is an issue; I thought it was only with an invalid path.

jkotalik:
@Rick-Anderson @guardrex can you do the word smithing?

aspnetcore/test/troubleshoot-azure-iis.md
guardrex: Why would it choke on writing to any folder in the IIS physical app path?

@fasigpt Can you provide the exact steps to repro this (with a valid path)?

guardrex: Confirmed ... I can't repro this behavior with IIS local or Azure Apps. For example ...

Capture

fasigpt:
Here is the repro
Create a .NET core 3.0 project. Deploy to Azure App Service. Modify the web.config as below (enable logging and change the path to .\stdout)

The exception can vary based on the hosting Model

**In Process:**
=============
We get 500.31 and here is how the web.config looks like

<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <location path="." inheritInChildApplications="false">
    <system.webServer>
      <handlers>
        <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModuleV2" resourceType="Unspecified" />
      </handlers>
      <aspNetCore processPath="dotnet" arguments=".\netcore2.dll" stdoutLogEnabled="true" stdoutLogFile=".\stdout" hostingModel="InProcess" />
    </system.webServer>
  </location>
</configuration>

**out Process:**
=============
We get 502.5 and here is how the web.config looks like

<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <location path="." inheritInChildApplications="false">
    <system.webServer>
      <handlers>
        <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModuleV2" resourceType="Unspecified" />
      </handlers>
      <aspNetCore processPath="dotnet" arguments=".\netcore2.dll" stdoutLogEnabled="true" stdoutLogFile=".\stdout" />
    </system.webServer>
  </location>
</configuration>

guardrex:
I still can't repro it ... I'm getting the log here ...

Capture

... with ...

```xml
<aspNetCore processPath="dotnet" arguments=".\rp22.dll" 
    stdoutLogEnabled="true" stdoutLogFile=".\stdout" hostingModel="InProcess" />
```

@jkotalik Do you want @fasigpt to move this to the engineering repo for investigation?

guardrex:
@fasigpt Have you tried running that app without the stdout log enabled just to make sure that the app does run ok?

fasigpt:
@guardrex Yes, the app runs fine without the logging enabled.

guardrex:
Ok ... let's let @jkotalik come back on and determine if this should go to engineering for further investigation. I can't repro it, so it seems like something in the framework/ANCM might be to blame.

guardrex:
... and @fasigpt ... one more ❓ ... can it write the logs using the path recommended ...

```
stdoutLogFile="\\?\%home%\LogFiles\stdout"
```

... out to the existing `LogFiles` folder.

fasigpt:
@guardrex Just to be clear in all cases the logs are always created. Its just that when we create logs under WWWROOT that's when the app fails with 502.5 error. 

yes, stdoutLogFile="\\?\%home%\LogFiles\stdout" generates log files and since the log is outside wwwroot the app runs fine without errors

guardrex:
Ok ... I understand. Thanks for clarifying.

I was able to do that here (create logs in wwwroot), so this will likely need to go to the product unit for investigation.

fasigpt:
@guardrex may i know who can help here? I can setup a call with them to show the repro

guardrex:
Wait until @jkotalik comes back on ... I *think* he's going to ask you to move this to the engineering repo for further investigation given that I can't repro the behavior.

guardrex:
Btw ... even if we end up documenting something about paths for stdout logging, it can’t be for wwwroot ... that’s really kinda scary that logs with very sensitive info would be served openly to anyone. Anyway tho ... let’s see how it plays out. My wild guess is that there’s some other factor at play here.

fasigpt:
@jkotalik  how do we proceed here ? you can just create an asp.net core 3.0 project and easily reproduce the issue.

guardrex:
At this point @fasigpt, I suspect 🦃 *Thanksgiving Break* 🌽 vacation. We'll probably have to wait until at least next week ... perhaps two weeks. Idk when he'll be back on.

Rick-Anderson:
> how do we proceed here ? you can just create an asp.net core 3.0 project and easily reproduce the issue.

@guardrex  can't repo.

guardrex:
Correct here, too. I can't repro this behavior.

Rick-Anderson:
> Its just that when we create logs under WWWROOT that's when the app fails with 502.5 error.

That's not supported or recommended.

guardrex:
Yes, ... **_not recommended_** at all 😨. Do you want to close on that basis? ... that seems reasonable.

guardrex:
@fasigpt ... We're going to close this because writing logs for public serving from the *wwwroot* folder is really scary for us. I can't even repro your behavior here ... I was able to write to that folder in Azure Apps. If you want to press on with this, consult with devs on SO or chats.

fasigpt:
@Rick-Anderson Many customers wouldn't know that they should not write to wwwroot. Can this be documented somewhere. we got few support cases wherein the app failed and customer had no idea that was because of logging to a wrong folder.

guardrex:
> not write to wwwroot

You mean stdout logs? We're not going to discuss that at all. The logs contain sensitive information. Certainly, people know that sensitive logs shouldn't be exposed publicly, which is exactly what's going to happen if they write them to *wwwroot*.

Besides, I can't repro the behavior, and we can't document something that we can't see breaking.

This isn't actionable right now, even for general files (not stdout logs) being written to the *wwwroot* folder. As far as the app is concerned, that's just another regular folder.

Issue #14353, opened 2019-09-17
Improve Blazor component lifecycle page


SteveSandersonMS:
[EDIT by guardrex to add DR's notes from the 1st PR to this opening content list]

[EDIT by guardrex to update on work done and check off work items that have been completed since this issue was opened]

[EDIT by guardrex to remove plan to create a new lifecycle topic ... that's done]

This is a suggestion that it's time to address a whole bunch of *"how to do..."* question issues that we've been collecting in response to developers' questions:

 * [13473 - Improve lifecycle docs](https://github.com/aspnet/AspNetCore/issues/13473)
 * [12773 - Clarify lifecycle events](https://github.com/aspnet/AspNetCore.Docs/issues/12773)
 * [12896 - Wrong documentation for OnParametersSetAsync/OnParametersSet lifecycle methods](https://github.com/aspnet/AspNetCore.Docs/issues/12896)
 * [11741 - Ability for services to detect when JSRuntime is available and safe to invoke](https://github.com/aspnet/AspNetCore/issues/11741)
 * [13584 - Document "cancelable background work" pattern for components](https://github.com/aspnet/AspNetCore/issues/13584)
 * [14260 - Provide guidance about how to do async work in Blazor](https://github.com/aspnet/AspNetCore.Docs/issues/14260)
 * ... and probably many more. I was just searching for "lifecycle" and "background".

Expand content to include:

 * A detailed description of what happens in what order, including:
    * How async tasks affect this
    * When (re)rendering is interleaved into this
    * How things are ordered among parent/child components
    * What kinds of work you're expected to do in each of the methods
 * Relationship with `Dispose`
    * Example: Pattern for cancelling background work
 * Relationship with JS interop
    * Example: Initializing a JS library by passing an `ElementReference`

### From DR on the PR to create the initial *Lifecycle* topic

Instead of focusing purely on the lifecycle events I think we want this article to explain more broadly how components work, like how and when the render. We should walk the user through a day in the life a component in a Blazor app. The lifecycle events are then just specific points that you can plugin your specific behaviors. 

Things that I think should be covered:
- What happens during the initial render of the app
  - We probably need to address prerendering as part of this
- When components get instantiated
- When services get injected from DI
- What happens when a parent/root component gets rendered
- When do components render after the initial render
  - In response to UI events and explicit calls to StateHasChanged
- How long do components life for? When do they get disposed?

WRT the short *State changes* (`StateHasChanged`) section ...

[...]

PR #15841, opened 2019-11-23
Blazor lifecycle topic

guardrex:
Addresses #14353

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/lifecycle?view=aspnetcore-3.0&branch=pr-en-us-15841)

* Split out lifecycle content from *Components* into a new *Lifecycle* topic.
* Revise and extend the content.
* Add API doc cross-links.
* Add lifecycle cross-links to other docs.
* I put `StateHasChanged` in here. It's in the API with `ComponentBase` and the lifecycle methods.

cc: @mkArtakMSFT This is the PR that we discussed: Recommend that we get this reviewed+merged, then JCN/SS can follow this up with a PR to mod the *Lifecycle* topic content as desired.


aspnetcore/blazor/lifecycle.md
```diff
@@ -0,0 +1,173 @@
...
+
+<xref:Microsoft.AspNetCore.Components.ParameterView> contains the entire set of parameter values each time `SetParametersAsync` is called.
+
+The default implementation of `SetParametersAsync` sets the value of each property decorated with the `[Parameter]` or `[CascadingParameter]` attribute that has a corresponding value in the `ParameterView`. Parameters that don't have a corresponding value in `ParameterView` are left unchanged.
+
+If `base.SetParametersAync` isn't invoked, the custom code can interpret the incoming parameters value in any way required. For example, there's no requirement to assign the incoming parameters to the properties on the class.
+
+### After parameters are set
+
+<xref:Microsoft.AspNetCore.Components.ComponentBase.OnParametersSetAsync*> and <xref:Microsoft.AspNetCore.Components.ComponentBase.OnParametersSet*> are called when a component has received parameters from its parent and the values are assigned to properties. These methods are executed after component initialization and each time the parent component is rendered:
```
danroth27: The child component might not get rendered if the parameter values haven't changed.

```suggestion
<xref:Microsoft.AspNetCore.Components.ComponentBase.OnParametersSetAsync*> and <xref:Microsoft.AspNetCore.Components.ComponentBase.OnParametersSet*> are called when a component has received parameters from its parent and the values are assigned to properties. These methods are executed after component initialization and each time new parameter values are specified:
```

aspnetcore/blazor/lifecycle.md
```diff
@@ -0,0 +1,173 @@
...
+no-loc: [Blazor]
+uid: blazor/lifecycle
+---
+# ASP.NET Core Blazor lifecycle
+
+By [Luke Latham](https://github.com/guardrex) and [Daniel Roth](https://github.com/danroth27)
+
+The Blazor framework includes synchronous and asynchronous lifecycle methods. Override lifecycle methods to perform additional operations on components during component initialization and rendering.
+
+### Component initialization methods
```
danroth27: Is it deliberate that these aren't top level headings? Currently these headings don't show up in the summary view.

guardrex: Ah :eye: thanks ... I accidentally killed a heading for these when I was deleting a paragraph: There should be a *Lifecycle methods* H2 here to cover these H3s. I'll add that H2 back.

aspnetcore/blazor/lifecycle.md
```diff
@@ -0,0 +1,173 @@
...
+
+    return renderUI;
+}
+```
+
+`ShouldRender` is called each time the component is rendered.
+
+Even if `ShouldRender` is overridden, the component is always initially rendered.
+
+## State changes
```
danroth27: I think this section would fit in better with a broader description of how and when components get rendered. Right now this section on StateHasChanged feels a bit out of context.

guardrex: ... another option is to move it to the *State Management* topic. It's used in an example over there but not discussed.

aspnetcore/blazor/lifecycle.md
```diff
@@ -0,0 +1,173 @@
...
+ms.custom: mvc
+ms.date: 11/23/2019
+no-loc: [Blazor]
+uid: blazor/lifecycle
+---
+# ASP.NET Core Blazor lifecycle
+
+By [Luke Latham](https://github.com/guardrex) and [Daniel Roth](https://github.com/danroth27)
+
+The Blazor framework includes synchronous and asynchronous lifecycle methods. Override lifecycle methods to perform additional operations on components during component initialization and rendering.
```
danroth27: Instead of focusing purely on the lifecycle events I think we want this article to explain more broadly how components work, like how and when the render. We should walk the user through a day in the life a component in a Blazor app. The lifecycle events are then just specific points that you can plugin your specific behaviors. 

Things that I think should be covered:
- What happens during the initial render of the app
  - We probably need to address prerendering as part of this
- When components get instantiated
- When services get injected from DI
- What happens when a parent/root component gets rendered
- When do components render after the initial render
  - In response to UI events and explicit calls to StateHasChanged
- How long do components life for? When do they get disposed?

guardrex: ... and cross-ref to Steve's original concepts, which include additional ideas for this topic ...

https://github.com/aspnet/AspNetCore.Docs/issues/14353

... and note that the idea here (this PR) was just to set up the new topic and then Steve would come in (or one of the other engineers) and modify the content from there.

danroth27: Ah ok, that's cool then.

guardrex:
I'll wait until @SteveSandersonMS signs off to merge this, and I'll move your notes over to the issue he opened.

aspnetcore/blazor/lifecycle.md
```diff
@@ -0,0 +1,175 @@
...
+}
+```
+
+### After component render
+
+<xref:Microsoft.AspNetCore.Components.ComponentBase.OnAfterRenderAsync*> and <xref:Microsoft.AspNetCore.Components.ComponentBase.OnAfterRender*> are called after a component has finished rendering. Element and component references are populated at this point. Use this stage to perform additional initialization steps using the rendered content, such as activating third-party JavaScript libraries that operate on the rendered DOM elements.
+
+The `firstRender` parameter for `OnAfterRenderAsync` and `OnAfterRender`:
+
+* Is set to `true` the first time that the component instance is invoked.
```
SteveSandersonMS:
```suggestion
* Is set to `true` the first time that the component instance is rendered.
```

aspnetcore/blazor/lifecycle.md
```diff
@@ -0,0 +1,175 @@
...
+* Ensures that initialization work is only performed once.
```
SteveSandersonMS:
```suggestion
* Can be used to ensure that initialization work is only performed once.
```

aspnetcore/blazor/lifecycle.md
```diff
@@ -0,0 +1,175 @@
...
+    {
+        await ...
+    }
+}
+```
+
+> [!NOTE]
+> Asynchronous work immediately after rendering must occur during the `OnAfterRenderAsync` lifecycle event.
+>
+> A component doesn't rerender after the completion of any returned <xref:System.Threading.Tasks.Task>. Infinite render loops are avoided.
```
SteveSandersonMS: I know you have a rule that things have to be phrased like this, but it's nearly impossible to understand. It's so ambiguous.

```suggestion
> Even if you return a <xref:System.Threading.Tasks.Task> from `OnAfterRenderAsync`, the framework will not schedule a further render cycle for your component once that task completes. This is to avoid an infinite render loop. It's different from the other lifecycle methods, which do schedule a further render cycle once the returned task completes.
```

aspnetcore/blazor/lifecycle.md
```diff
@@ -0,0 +1,175 @@
...
+protected override void OnAfterRender(bool firstRender)
+{
+    if (firstRender)
+    {
+        ...
+    }
+}
+```
+
+`OnAfterRender` *isn't called when prerendering on the server.*
```
SteveSandersonMS:
```suggestion
`OnAfterRender` and `OnAfterRenderAsync` *are not called when prerendering on the server.*
```

aspnetcore/blazor/lifecycle.md
guardrex: It's cool ... I'll just change the future tense.

guardrex:
@SteveSandersonMS Cool ... thanks. This only "addresses" your issue. There are a couple of remaining items in your list on the issue, and DR has further goals for this content that I added to your issue.

PR #15846, opened 2019-11-24
Surface ValueTask in Blazor JS interop

### LINKED ISSUE #14739: Blazor JSInterop example return a ValueTask<T>
From @simonpbond ... [EDIT by @guardrex for formatting]

> The JSInterop - Blazor documentation should return a `ValueTask<T>`?

> https://docs.microsoft.com/aspnet/core/blazor/javascript-interop?view=aspnetcore-3.0#capture-references-to-elements

> To use an extension method to focus an element, create a static extension method that receives the `IJSRuntime` instance:

```csharp
public static Task Focus(this ElementReference elementRef, IJSRuntime jsRuntime)
{
    return jsRuntime.InvokeAsync<object>(
        "exampleJsFunctions.focusElement", elementRef);
}
```

> I may be wrong, however I assume it should now be:

```csharp
public static ValueTask<T> Focus<T>(this ElementReference elementRef, IJSRuntime jsRuntime)
{
    return jsRuntime.InvokeAsync<T>(
        "exampleJsFunctions.focusElement", elementRef);
}
```

---
#### Document Details

* Content: [ASP.NET Core Blazor JavaScript interop](https://docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interop?view=aspnetcore-3.0)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes  #14739

[Internal Review Topic (*Capture references to elements* section)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interop?view=aspnetcore-3.0&branch=pr-en-us-15846#capture-references-to-elements)

Part 🐞, part ✨:

* 🐞: Fix the in-text focus example taking it from `return jsRuntime.InvokeAsync<object>` to `await jsRuntime.InvokeVoidAsync`.

* ✨: We've already made updates on a prior PR to use `ValueTask<T>` in the sample app, so the updates here only focus on the topic text+examples.

  Updates the `TickerChanged` example to use `ValueTask<string>`.

  Let's **_NOT_** return a `ValueTask` with the focus-an-element example. That wouldn't be correct (or make sense), but let's also not drop that focus example, which I think is pretty good. Therefore, let's add a `ValueTask<T>`-extension-method-with-`ElementReference` example to the topic.

* **Open ❓**: In our prior update (mostly focused on the sample app), we went with `ValueTask<object>`/`InvokeAsync<object>` ...

  ```csharp
  public ValueTask<object> CallHelloHelperSayHello(string name)
  {
      // sayHello is implemented in wwwroot/exampleJsInterop.js
      return _jsRuntime.InvokeAsync<object>(
          "exampleJsFunctions.sayHello",
          DotNetObjectReference.Create(new HelloHelper(name)));
  }
  ```

  ... and we didn't provide guidance in the topic on the *type*. Is there anything specific that we can add about the `T` in `ValueTask<T>`/`InvokeAsync<T>`? if `T` isn't used directly in dev code, things get murky given the weakly-typed nature of JS.

* For the *not-faint-of-*:heart:, we can cross-link the engineering JS interop sample in an *Additional Resources* section.

* Not sweating the API doc cross-links yet. _That's **coming soon** to Blazor docs near you!_


aspnetcore/blazor/javascript-interop.md
```diff
@@ -173,27 +173,42 @@ window.exampleJsFunctions = {
 }
 ```
 
-Use `IJSRuntime.InvokeAsync<T>` and call `exampleJsFunctions.focusElement` with an `ElementReference` to focus an element:
+For a method that doesn't return a value, such as focusing an element, use `IJSRuntime.InvokeVoidAsync` with the preceding JavaScript function and the `ElementReference`:
```
danroth27:
```suggestion
To call a JavaScript function that doesn't return a value, use `IJSRuntime.InvokeVoidAsync`. The following code sets the focus on the username input by calling the preceding JavaScript function with the captured `ElementReference`:
```

guardrex:
@danroth27 Do you want to address the "open question" :point_up: in the opening remarks?

danroth27:
> Do you want to address the "open question" ☝️ in the opening remarks?

I *think* the `T` is supposed to be the value returned by the call to the JS function. In the case of the call to `sayHello`, the JS function calls a .NET method and returns the result of that. The .NET method returns a `string`, so I would have thought the T in this case should be `string`. @pranavkm could you please confirm?

guardrex:
Experimentation did bear that out (return a number 1 but go with a `string` type and :boom:), but `object` works in that *SayHello* example.

I think the original engineering example I followed for the *SayHello* used `object` in a similar scenario where a `string` was returned. I'm sure that I *stole* that.

pranavkm:
The object you return is a JSON object (https://www.json.org/json-en.html) which is fairly strict about it's conversion rules. Usually `T` would be the .NET type that best maps to the JSON type that is being returned. This makes it easier \ correct to do things like use it: 

```C#
string person = await PromptMessage("What is your name?");
helloText = $"Hello ${person}";
...
```

Reading it as an `object` type would result in the value being an instance of [`JsonElement`](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement?view=netcore-3.0). You'll generally have a harder time using it compared to a POCO \ value type.

guardrex:
@danroth27 The updates for `T` on the last commit should cover it. The examples are string-based, so they use the `string` type now. A line is added where `T` is initially discussed ...

> `T` should match the .NET type that best maps to the JSON type returned.

guardrex:
@danroth27 I reacted to @pranavkm's feedback, so I'm going to unblock Blazor sample updates this afternoon by merging this now. If you see anything sketchy in my last remark :point_up:, I'll :ear: here and react immediately with a patch PR.

PR #15890, opened 2019-11-27
Edit BackgroundProcessing method name

msynk:


guardrex:
Thanks @msynk ... but we're going to keep the name we have. It was approved by engineering, and there's more work that would need to be done to make a change like this ... the other sample and the topic text (probably) would require matching updates. We're 🏃😅 and can't take on the extra work for such a small change. If you want to open an issue and explain why this is important, we can discuss it further (use the **This Page** feedback button/form at the bottom of the topic). Thanks anyway tho.

PR #15891, opened 2019-11-27
Edit BackgroundProcessing method name

msynk:


guardrex:
I see u placed the text update here and that the 2.x sample doesn't require such a change.

I don't follow the reasoning for the change tho. Please open an issue to discuss this if you want to proceed. Use the **This Page** feedback button/form at [the bottom of the topic](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services). If the new name is better, I'll re-open the PRs, cross-link them to the issue, and approve.

PR #15928, opened 2019-11-30
Fix ASP.NET Core version for Content-Language header

hishamco:
@Rick-Anderson this is a reminder to merge this when ASP.NET Core 3.1 is shipped. A little bit history on this, I created a PR in localization long time ago, I supposed to be merged in 3.0, but the branch has been changed before merge this

After that I wrote a blog post about Localization 3.0, since then I got some feedback that this not working on 3.0, then I realize something goes wrong, until Ryan confirms

For that I pushed this PR, also I will change my blog post


Rick-Anderson:
Fixes #15931

PR #15433, opened 2019-11-01
Blazor Components Lifecycle - OnParametersSet

hakenr:
`OnParametersSet[Async]` is not "executed each time the component is rendered". If you just handle an event inside the component (such as button onclick), the component gets re-rendered without calling the `OnParametersSet[Async]` (`SetParametersAsync`).
Try this sample page:
```cshtml
@page "/handleevent"

<div>
	<button @onclick="HandleClick">Handle event</button>
	@something
</div>

@code
{
	private string something = String.Empty;

	private void HandleClick()
	{
		something = "something";
	}
}
```

PR #15630, opened 2019-11-09
Priority 0/1 Blazor Updates

guardrex:
Fixes #15603
Fixes #13619
Fixes #12896
Fixes #14884
Fixes #15679
Fixes #15765 

*... sorry for so many!* :smile: ... the bright side is that this one knocks out a lot of work without merge conflicts 💥.

* Recommended pattern for direct DOM manipulation in Blazor
* Add DOM events to Event Arg Types table
* Wrong documentation for OnParametersSetAsync/OnParametersSet lifecycle methods
* Document control of event propagation and default actions
* Document Blazor's MonoLinkerI18NAssemblies build option
* Remaining Component Tag Helper updates for 3.1.

PR #15956, opened 2019-12-02
Blazor Components Lifecycle updates

### LINKED ISSUE #16099: Enhance lifecycle methods from API remarks
See: https://github.com/aspnet/AspNetCore.Docs/issues/12773#issue-453929033

---
#### Document Details

* ID: 4bae853d-80e3-ef81-69eb-e8276eb853fe
* Version Independent ID: 01216f14-b45c-73b0-9f1f-300e7ddd788f
* Content: [ASP.NET Core Blazor lifecycle](https://docs.microsoft.com/en-us/aspnet/core/blazor/lifecycle?view=aspnetcore-3.1)
* Content Source: [aspnetcore/blazor/lifecycle.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/lifecycle.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #16099
Addresses #16270
Addresses #16273

DR: This one needs a final :eye: following changes that were made *after* you signed off.

Moved from #15433.

Steve, you were requested to :eyes: by DR on this one. He signed off on #15433.


andoband:
I see that @guardrex merged updated language for this a few weeks ago with #15630...

> `OnParametersSetAsync` and `OnParametersSet` are called when a component has received parameters from its parent and the values are assigned to properties. These methods are executed after component initialization and each time the parent component is rendered:  

...which seems more correct than the proposed wording in this request.  Mainly it's important to point out that these methods are called when the _parent_ is rendered, not necessarily when the component is rendered.  For example, if a component has a button and handles it's click event to change some component-local state (state that the parent doesn't know about) then the `OnParameterSet[Async]` methods do not get called.

Also, I think the proposed change incorrectly implies that these methods are only called when parameters have changed, but the behavior is more nuanced than that.  The methods are always called after initialization (even if there are no parameters).  Then, before subsequent renders of the parent, if the `RenderTreeDiffBuilder` determines there's a change to a parameter or can't determine if there was a change then the methods are called again.  The `RenderTreeDiffBuilder` can determine when simple parameters change but not complex type parameters.  If one of the parameters has a complex type then the methods are called before every render, even if the parameters haven't changed.  See `RenderTreeDiffBuilder` code and comment: https://github.com/aspnet/AspNetCore/blob/0a79d34f62a2e153d7104a88b487b75ef8860bb7/src/Components/Components/src/RenderTree/RenderTreeDiffBuilder.cs#L512

Sorry for the wordy post for a simple change request, but I spent a lot of time trying to understand what was happening when implementing custom controls in my own Blazor app, and I'd love to see much better docs for all the lifecycle events.

guardrex:
### Item 1

> these methods are called when the parent is rendered

For ...

> \* After component initialization.

We could go with something like ...

> \* After component initialization when its parent component is rendered.

### Item 2

> these methods are only called when parameters have changed, but the behavior is more nuanced than that

For ...

> \* When a component has received changed parameters from its parent and the values are assigned to properties.

We could go with something like ...

> \* When a component has assigned values to properties from changes to simple parameters by its parent or when complex type parameters are used.

... which could alternatively compose as a 2nd level list (probably better) ...

> \* When a component has assigned values to properties:
> &nbsp;&nbsp;\* From changes to simple parameters by its parent.
> &nbsp;&nbsp;\* When complex type parameters are used.

guardrex:
@SteveSandersonMS Need your sign off ... everything is on the diff, and it's short (quick review!).

aspnetcore/blazor/lifecycle.md
```diff
@@ -19,29 +19,29 @@ The Blazor framework includes synchronous and asynchronous lifecycle methods. Ov
...
-protected override void OnInitialized()
+protected override async Task OnInitializedAsync()
 {
-    ...
+    await ...
 }
 ```
 
+> [!NOTE]
+> Asynchronous work during component initialization must occur during the `OnInitializedAsync` lifecycle event.
```
SteveSandersonMS: I don't completely understand the purpose of this note. I'd say that it's a *common pattern* to do this, but it isn't mandatory.

For example, you could trigger async initialization work during `SetParametersAsync` or `OnParametersSetAsync` or in an extreme case, even in the component's constructor (though in that case you have to fire-and-forget the operation, since constructors aren't async).

In other words, I'd suggest dropping this note as it's a bit misleading. As long as we've been clear that this is a common and recommended pattern, there's no need to use the word `must` about it.

aspnetcore/blazor/lifecycle.md
```diff
@@ -63,7 +63,12 @@ If `base.SetParametersAync` isn't invoked, the custom code can interpret the inc
 
 ### After parameters are set
 
-<xref:Microsoft.AspNetCore.Components.ComponentBase.OnParametersSetAsync*> and <xref:Microsoft.AspNetCore.Components.ComponentBase.OnParametersSet*> are called when a component has received parameters from its parent and the values are assigned to properties. These methods are executed after component initialization and each time new parameter values are specified:
+<xref:Microsoft.AspNetCore.Components.ComponentBase.OnParametersSetAsync*> and <xref:Microsoft.AspNetCore.Components.ComponentBase.OnParametersSet*> are called:
+
+* After component initialization when its parent component is rendered.
+* When a component has assigned values to properties:
+  * From *changes* to simple parameters by its parent.
+  * When complex-type parameters *are used*, even when complex-type parameter values don't change.
```
SteveSandersonMS: For this section, I find the phrasing a bit confusing. I'd say something like:

`OnParametersSetAsync` and `OnParametersSet` are called:

 * When the component is being initialized and has received its first set of parameters from its parent component
 * When the parent component re-renders and supplies a possibly-changed set of parameters
    * As a special case, if the parameters are all of known primitive immutable types, the framework will skip re-rendering the child if their values are unchanged. 
    * If there are any complex-typed parameters, the framework cannot know whether the values have mutated internally so it must always treat the parameter set as changed.

andoband: > - When the component is being initialized and has received its first set of parameters from its parent component

Could you tweak this a bit to indicate they are called after OnInitialized?

aspnetcore/blazor/lifecycle.md
guardrex: I think the purpose was **_in relation to using the non-async `OnInitialized`_**, but that really is *very obvious*. I'll drop it.

aspnetcore/blazor/lifecycle.md
guardrex: I'll remove "being," so it will read ...

> When a component is initialized and ...

These sections are in order. I *think* it will be clear. We'll see. We'll get feedback :ear: if there's confusion on the order.

guardrex:
This isn't working out ...

It doesn't make sense to me to create a condition of the methods being called with "a possibly-changed set of parameters" followed by conditions *that don't say clearly when the methods are called*.

I'll float another draft ... a hybrid ... shortly.

guardrex:
Ok ... that seems to maintain the further explanation and keep it focused on conditions.

PR #15984, opened 2019-12-03
Add note on RegularExpression attribute validation

jasonleenaylor:
This note will help inform developers of the limitations discussed in https://github.com/dotnet/corefx/issues/42487


aspnetcore/mvc/models/validation.md
```diff
@@ -419,6 +419,8 @@ Built-in validation attributes include:
 * `[Url]`: Validates that the property has a URL format.
 * `[Remote]`: Validates input on the client by calling an action method on the server. See [`[Remote]` attribute](#remote-attribute) for details about this attribute's behavior.
 
+The `RegularExpression` attribute uses [ECMAScript](/dotnet/standard/base-types/regular-expression-options#ecmascript-matching-behavior) options for matching on client side validation. For more informaion, see [this GitHub issue](https://github.com/dotnet/corefx/issues/42487).
```
anurse: A couple notes:

1. Might be worth making this a [`[!NOTE]` callout](https://dotnet.github.io/docfx/spec/docfx_flavored_markdown.html#note-warningtipimportant)
2. I'd rephrase this a bit. It implies the "RegularExpression attribute" is doing something, when it's really just a piece of data. Better to rephrase around something like "When using `[RegularExpression]` validation in an ASP.NET or ASP.NET Core application ...".
3. Seems appropriate to make clear *when and why* this happens (when: doing client-side validation, why: because the validation rules are translated to JavaScript)

Rick-Anderson: @anurse we avoid chrome/bling/note when possible.

anurse: Sure, then not a `[!NOTE]` but the rest still applies.

Rick-Anderson: @anurse  I applied the rest. Do you want to review again or can I S&M.

aspnetcore/mvc/models/validation.md
```diff
@@ -419,6 +419,8 @@ Built-in validation attributes include:
 * `[Url]`: Validates that the property has a URL format.
 * `[Remote]`: Validates input on the client by calling an action method on the server. See [`[Remote]` attribute](#remote-attribute) for details about this attribute's behavior.
 
+When using the `[RegularExpression]` validation in an ASP.NET Core app, [ECMAScript](/dotnet/standard/base-types/regular-expression-options#ecmascript-matching-behavior) options are used for matching on client side validation. On client side validation, the validation rules are translated to JavaScript. For more information, see [this GitHub issue](https://github.com/dotnet/corefx/issues/42487).
```
anurse: Had a thought this was a little clearer. We should refer to `[RegularExpression]` as an attribute, and we're already in a doc about ASP.NET Core (I didn't realize that initially for some reason :)).

```suggestion
When using the `[RegularExpression]` attribute with client-side validation, the regex is executed in JavaScript on the client. This means [ECMAScript](/dotnet/standard/base-types/regular-expression-options#ecmascript-matching-behavior) matching behavior will be used. For more information, see [this GitHub issue](https://github.com/dotnet/corefx/issues/42487).
```

PR #15989, opened 2019-12-03
Update appsettings-sample.json

BlancHeart:
fix the comment


guardrex:
Hello @BlancHeart ... Thanks for sending this, but we stick fairly close to standards (*fairly close* :smile:). Comments aren't allowed in JSON files. If you want to submit a PR for the topic that goes with this that calls out this info, that would be great.

PR #15990, opened 2019-12-03
Update appsettings-sample.json

BlancHeart:
Remove the comment in JSON


guardrex:
*Oh* ... was there one there. 🤦‍♂ Yeah ... this isn't something that I feel we should be showing. There's a JSON5 thing I think that allows comments in JSON, but we're not going with that.

If you want to submit a PR with topic changes that would call out this and what you had on the other PR, it would be *welcome*.

guardrex:
Thanks @BlancHeart! :rocket: :rocket:

PR #16048, opened 2019-12-06
Fix logging output

Marcel-B:
UserId was not logged correctly.


guardrex:
Hello @Marcel-B ... the template approach for logging is correct as far as we know. Open an issue from the bottom of the topic for this change ... use the **This Page** feedback button/form at the bottom of the topic to describe the behavior so that we can discuss this further. We might need to have engineering look into it, so we may either ping them on your issue or have you open it on their repo for investigation.

PR #16049, opened 2019-12-06
Fix logging output

Marcel-B:
UserId was not logged correctly.


guardrex:
See :point_right: https://github.com/aspnet/AspNetCore.Docs/pull/16048#issuecomment-562622571

PR #16092, opened 2019-12-09
Add global auth filter + policy migration example

pranavkm:
@Rick-Anderson  edit: Moved to #16129

PR #16129, opened 2019-12-10
Merge into Pranav's PR

Rick-Anderson:
Merge to #16092 when approved.

[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-1.0&branch=pr-en-us-16129&tabs=visual-studio#authorization-for-specific-endpoints)

External review:

* [Download 1.zip](https://www.dropbox.com/sh/yy994551cje0kap/AAB0tRnm7gze2kYrSYPSFHIda?dl=0)
* Unzip 1.zip file

### DIFF
--- a/aspnetcore/migration/22-to-30.md
+++ b/aspnetcore/migration/22-to-30.md
@@ -731,7 +731,7 @@ Authorization can also be configured for specific classes of endpoints. The foll
 
 [!code-csharp[](22-to-30/samples/Startup.cs?name=snippet&highlight=8-9,21-22,26-27,53-54)]
 
-Policies can also be customized. The `DefaultPolicy` is configured to require authentication and a specific scope:
+Policies can also be customized. The `DefaultPolicy` is configured to require authentication:
 
 [!code-csharp[](22-to-30/samples/Startup2.cs?name=snippet&highlight=21-26,52)]
 
--- a/aspnetcore/migration/22-to-30.md
+++ b/aspnetcore/migration/22-to-30.md
@@ -727,6 +727,9 @@ public class HomeController : ControllerBase
 
 Authorization can also be configured for specific classes of endpoints. The following code is an example of converting an MVC app that configured a global `AuthorizeFilter` to an app with a specific policy to use `RequireAuthorization`:
 
+ [!code-csharp[](22-to-30/samples/Startup.cs?name=snippet)]
+
+
 ```csharp
 public void ConfigureServices(IServiceCollection services)
 {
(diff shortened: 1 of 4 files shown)


Rick-Anderson:
@serpent5  please review.

serpent5:
As a more general comment, there's some really useful information here re authz that's not specific to the migration path. I wouldn't expect to come to this particular page to learn out about things like `DefaultPolicy`, `FallbackPolicy`, etc.

I wonder if it would be good to move some of this information out to the authz pages under "Security and Identity" and just cover the migration from the global filter here.

Rick-Anderson:
> As a more general comment, there's some really useful information here re authz that's not specific to the migration path. I wouldn't expect to come to this particular page to learn out about things like `DefaultPolicy`, `FallbackPolicy`, etc.

> I wonder if it would be good to move some of this information out to the authz pages under "Security and Identity" and just cover the migration from the global filter here. Clearly, if that's deemed to be a good idea, it wouldn't be part of this work.
That's a great idea. Do you want to open an issue for that. We can merge this in the mean time then move it over.

aspnetcore/migration/22-to-30/samples/Startup.cs
```diff
@@ -0,0 +1,81 @@
...
+                     options => options.SignIn.RequireConfirmedAccount = true)
+                .AddEntityFrameworkStores<ApplicationDbContext>();
+
+            // Pre 3.0:
+            // services.AddMvc(options => options.Filters.Add(new AuthorizeFilter(...));
+
+            services.AddControllersWithViews();
+            services.AddRazorPages();
+            services.AddAuthorization(o => o.AddPolicy(_RequireAuthenticatedUserPolicy,
+                            builder => builder.RequireAuthenticatedUser()));
```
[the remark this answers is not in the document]
pranavkm: I wanted to show a use of `RequireAuthorization`. `RequireAuthenticatedUser` was easy to set up, but feel free to add some more to this if you think that helps

aspnetcore/migration/22-to-30.md
```diff
@@ -688,7 +688,7 @@ public void Configure(IApplicationBuilder app)
     });
 }
 
-public class HomeController : ControllerBase
+public class HomeController : Controller
```
pranavkm: Why change this?

Rick-Anderson: > Why change `ControllerBase` to `Controller`

That's what the templates generate so I assume that's what most customers use.

PR #16153, opened 2019-12-12
Update security.md

### LINKED ISSUE #16003: AllowCredentials and no authentication
Why do the credentials need to be enabled even if authentication is not used?

---
#### Document Details

* ID: 8af11946-ea05-3be2-6fa1-c4c01aa76d8c
* Version Independent ID: 58e87eb0-240b-ab00-c5f4-32a0ab2e0028
* Content: [Security considerations in ASP.NET Core SignalR](https://docs.microsoft.com/en-us/aspnet/core/signalr/security?view=aspnetcore-3.1#feedback)
* Content Source: [aspnetcore/signalr/security.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/signalr/security.md)
* Product: **aspnet-core**
* GitHub Login: @bradygaster
* Microsoft Alias: **anurse**

Rick-Anderson:
Fixes #16003

### DIFF
--- a/aspnetcore/signalr/security.md
+++ b/aspnetcore/signalr/security.md
@@ -28,9 +28,7 @@ For more information on configuring CORS, see [Enable Cross-Origin Requests (COR
 
 * Allow the specific expected origins. Allowing any origin is possible but is **not** secure or recommended.
 * HTTP methods `GET` and `POST` must be allowed.
-* Credentials must be enabled in order for cookie-based sticky sessions to work correctly:
-  * Even when authentication is not used.
-  * Because there is no customization in the browser client to not use credentials.
+* Credentials must be allowed in order for cookie-based sticky sessions to work correctly. They must be enabled even when authentication is not used.
 
 <!--
 ::: moniker range=">= aspnetcore-5.0"  // Moniker here just to make sure this doesn't get missed in the 5.0 version update.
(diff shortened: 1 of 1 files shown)


aspnetcore/signalr/security.md
```diff
@@ -28,7 +28,17 @@ For more information on configuring CORS, see [Enable Cross-Origin Requests (COR
 
 * Allow the specific expected origins. Allowing any origin is possible but is **not** secure or recommended.
 * HTTP methods `GET` and `POST` must be allowed.
-* Credentials must be enabled, even when authentication is not used.
+* Credentials must be enabled:
+  * Even when authentication is not used.
+  * Because there is no customization in the browser client to not use credentials.
```
BrennanConroy: I would not mention this at all, and it doesn't indicate the real reasons credentials are enabled.

BrennanConroy:
cc @anurse to explain why credentials are enabled while I'm out.

aspnetcore/signalr/security.md
anurse: Yeah, this isn't correct. There **is** customization in the browser to do this (and we've even decided to expose that at some point moving forward.

I'd perhaps rephrase this whole bullet point (line 31-33) as something like (feel free to continue playing with style):

```
* Credentials must be allowed in order for cookie-based sticky sessions to work correctly. They must be enabled even when authentication is not used.
```

PR #16164, opened 2019-12-12
Inconsistent use of @microsoft/signalr vs @aspnet/signalr

igor-toporet:
This page section suggests installing `@microsoft/signalr` but down below the `@aspnet/signalr` is imported in the code snippet which leads to TypeScript code that does not compile.

https://docs.microsoft.com/en-us/aspnet/core/tutorials/signalr-typescript-webpack?view=aspnetcore-2.2&tabs=visual-studio#enable-client-and-server-communication-1


scottaddie:
@igor-toporet Thank you for bringing this problem to our attention. I've opened another PR to fix the problem. We maintain a 2.x version and a 3.x version of the tutorial. For ASP.NET Core 3.x, you'd want to install the newer `@microsoft/signalr` package.

PR #16170, opened 2019-12-12
3.1 Sample Updates

guardrex:
Addresses #5495

* The *low-hanging fruit* ones that can be updated without too much concern that they'll :boom: on us without testing.
* A handful of others will be put up on later PR(s) because they're complicated enough that they warrant a test run when updating to 3.1.


guardrex:
*From you to the :pray: *server gods :ear:* :pray:.

Issue #15473, opened 2019-11-03
Is this up to date with ASP.NET Core 3.0?


lol768:
It suggests we:

> Add a package reference to the Microsoft.AspNetCore.Diagnostics.HealthChecks package. To perform health checks using Entity Framework Core, add a package reference to the Microsoft.Extensions.Diagnostics.HealthChecks.EntityFrameworkCore package.

But the sample repo's [project file](https://github.com/aspnet/AspNetCore.Docs/blob/f0e7d8e39f902c184da54a04c57f74dcf48162d5/aspnetcore/host-and-deploy/health-checks/samples/3.x/HealthChecksSample/SampleApp.csproj#L8-L13) does no such thing? There is no `Microsoft.AspNetCore.Diagnostics.HealthChecks` reference

https://github.com/aspnet/AspNetCore.Docs/blob/f0e7d8e39f902c184da54a04c57f74dcf48162d5/aspnetcore/host-and-deploy/health-checks/samples/3.x/HealthChecksSample/SampleApp.csproj#L8-L13

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: c0cbb094-29e9-740a-9fd6-db372641d753
* Version Independent ID: 52d8efc9-a336-71d9-6d90-98173cbf6bf4
* Content: [Health checks in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-3.0)
* Content Source: [aspnetcore/host-and-deploy/health-checks.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/host-and-deploy/health-checks.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-hostdeploy**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Typo ... it should be `Microsoft.Extensions.Diagnostics.HealthChecks`.

I'll patch it. Thanks for calling it out.

guardrex:
Well .... it **_was_** a thing ...

https://www.nuget.org/packages/Microsoft.AspNetCore.Diagnostics.HealthChecks

... possibly 2.x. I'll take a closer look.

guardrex:
[...]

PR #16204, opened 2019-12-15
Health Checks sample 3.1 updates

guardrex:
Addresses #5495 
Addresses #15473

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-2.2&branch=pr-en-us-16204)

* 3.1 sample app confirmed running.
* Let's put the `System.Text.Json` approach in here but also retain the existing `Newtonsoft.Json` approach. A directive in the file makes it easy for the dev to switch back and forth.
* WRT the `System.Text.Json` approach: I couldn't get `WriteNumber` to work with a directly casted object (`(Decimal)item.Value`). I ended up performing the cast on a separate line with (`Convert.ToDecimal`) to make it happy.

cc: @lol768

### DIFF
--- a/aspnetcore/host-and-deploy/health-checks/samples/3.x/HealthChecksSample/CustomWriterStartup.cs
+++ b/aspnetcore/host-and-deploy/health-checks/samples/3.x/HealthChecksSample/CustomWriterStartup.cs
@@ -1,8 +1,13 @@
-﻿using System.Linq;
+﻿#define SYSTEM_TEXT_JSON
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.Json;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Diagnostics.HealthChecks;
-using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Diagnostics.HealthChecks;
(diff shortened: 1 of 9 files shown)


lol768:
LGTM! Many thanks for this, @guardrex

aspnetcore/host-and-deploy/health-checks/samples/3.x/HealthChecksSample/CustomWriterStartup.cs
```diff
@@ -49,11 +54,53 @@ public void Configure(IApplicationBuilder app)
...
         }
 
-        #region snippet_WriteResponse
-        private static Task WriteResponse(HttpContext httpContext, 
-            HealthReport result)
+#if SYSTEM_TEXT_JSON
+        #region snippet_WriteResponse_SystemTextJson
+        private static Task WriteResponse(HttpContext context, HealthReport result)
+        {
+            context.Response.ContentType = "application/json";
```
rynowak: Ideally this would `application/json; charset=utf-8` (both samples)

aspnetcore/host-and-deploy/health-checks/samples/3.x/HealthChecksSample/CustomWriterStartup.cs
```diff
@@ -49,11 +54,53 @@ public void Configure(IApplicationBuilder app)
...
+                    foreach (var entry in result.Entries)
+                    {
+                        writer.WriteStartObject(entry.Key);
+                        writer.WriteString("status", entry.Value.Status.ToString());
+                        writer.WriteString("description", entry.Value.Description);
+                        writer.WriteStartObject("data");
+                        foreach (var item in entry.Value.Data)
+                        {
+                            var value = Convert.ToDecimal(item.Value);
+                            writer.WriteNumber(item.Key, value);
```
rynowak: This isn't right. The reason why you're having a problem, with this is that there's no guarantees about what kind of object `item.Value` will be.

The right incantation would be:

```C#
writer.WritePropertyName(item.Key);
JsonSerializer.Serialize(writer, item.Value, item.Value?.GetType() ?? typeof(object));
```

guardrex: I understand what ur saying, but I (perhaps incorrectly) assumed that this example wasn't meant to cover a delegate *in general*. I thought that **_this_** delegate for **_this_** health check wouldn't ever have anything in `entry.Value` that wouldn't be castable into a decimal (given the `HealthCheckResult` from the `MemoryHealthCheck`). Therefore, I wasn't concerned (*until now* lol). I understand that you provided code that won't break in general, and I understand that it's better to show it this way. Perhaps, it's never safe to assume anything about an object's type in a scenario like this. I'll keep that in mind going forward.

rynowak: >Perhaps, it's never safe to assume anything about an object's type in a scenario like this. I'll keep that in mind going forward.

Cool, I think you're on track now 👍

Issue #15705, opened 2019-11-15
Add a section that shows how to customize the model bound using [FromBody] and System.Text.Json


javiercn:
This is common enough that it should have an explanatory section.

Use this code-snippet as a starting point
```csharp
    [ApiController]
    [Route("[controller]")]
    public class CustomJsonController : ControllerBase
    {
        [HttpPost]
        public IActionResult Post([FromBody] DataModel model) => Ok(model);

    }

    public sealed class DataModel
    {
        [Required]
        public ObjectId ObjectId { get; set; }
    }

    [JsonConverter(typeof(ObjectIdConverter))]
    public struct ObjectId
    {
        public ObjectId(int id) => Id = id;

        public int Id { get; }
    }

    internal class ObjectIdConverter : JsonConverter<ObjectId>
    {
        public override ObjectId Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options) => new ObjectId(JsonSerializer.Deserialize<int>(ref reader, options));

        public override void Write(Utf8JsonWriter writer, ObjectId value, JsonSerializerOptions options) => writer.WriteNumberValue(value.Id);
    }
```
---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e24f48f8-6c4d-ecda-6ce6-a7826d396e99
* Version Independent ID: 9c4af557-8d06-4c6c-ad2a-0dd234547563
* Content: [Model Binding in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/model-binding?view=aspnetcore-3.0#frombody-attribute)
* Content Source: [aspnetcore/mvc/models/model-binding.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/mvc/models/model-binding.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-mvc**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Legends:
@javiercn Can you also please elaborate on [section manual-model-binding](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/model-binding?view=aspnetcore-3.0#manual-model-binding) please.

Rick-Anderson:
@serpent5  this looks a like an interesting issue, let me know if you're interested at some point.

javiercn:
[...]

PR #16232, opened 2019-12-17
Update model-binding.md

Rick-Anderson:
Fixes  #15745

[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/mvc/models/model-binding?view=aspnetcore-3.1&branch=pr-en-us-16232#prevent-over-posting-with-tryupdatemodelasync)

@serpent5  please review.

PR #16240, opened 2019-12-17
Add 3.0 version of Model Binding sample.

### LINKED ISSUE #15745: Add link to TryUpdateModelAsync
See #15705

add link to https://docs.microsoft.com/en-us/aspnet/core/data/ef-rp/crud?view=aspnetcore-3.0#tryupdatemodelasync

---
#### Document Details

* Content: [Model Binding in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/model-binding?view=aspnetcore-3.0#frombody-attribute)
* Content Source: [aspnetcore/mvc/models/model-binding.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/mvc/models/model-binding.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

serpent5:
Addresses #15705.
Fixes #15745.

As noted in the commit messages, the 1st commit was a copy/paste of the 2.x sample with **no changes**. The other commits show the actual changes made.

/cc @Rick-Anderson 

@Rick-Anderson  edit:
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/mvc/models/model-binding?view=aspnetcore-3.1&branch=pr-en-us-16240)


aspnetcore/mvc/models/model-binding.md
```diff
@@ -857,8 +857,8 @@ To make the ASP.NET Core route value provider and query string value provider un
 * Replace the [culture value](https://github.com/aspnet/AspNetCore/blob/e625fe29b049c60242e8048b4ea743cca65aa7b5/src/Mvc/Mvc.Core/src/ModelBinding/QueryStringValueProviderFactory.cs#L30) passed to the value provider constructor with [CultureInfo.CurrentCulture](xref:System.Globalization.CultureInfo.CurrentCulture)
 * Replace the default value provider factory in MVC options with your new one:
 
-[!code-csharp[](model-binding/samples_snapshot/2.x/Startup.cs?name=snippet)]
-[!code-csharp[](model-binding/samples_snapshot/2.x/Startup.cs?name=snippet1)]
+[!code-csharp[](model-binding/samples_snapshot/Startup.cs?name=snippet)]
+[!code-csharp[](model-binding/samples_snapshot/Startup.cs?name=snippet1)]
```
serpent5: This code is for 3.0, but it's shown in both versions. I removed `2.x` from the folder structure to emphasise that.

Should we?:

1. Create a 2.2 version of this code and reference that accordingly.
2. Remove this section from the 2.2 version.
3. Do nothing.

guardrex: I think it depends on the differences for what's being shown. If churn occurred (as is usually the case for 2.x and 3.x in *Startup.cs*), a good approach is to create a *3.x* folder in the snapshots folder to hold the 3.x *Startup.cs* in parallel to the 2.x version of the file. That way, the future versions (5.x, 6.x ...) can follow suit, and the topic's links are easy to manage over time (i.e., **_saves MS 💰_**). Surely tho, it's overkill when there haven't been any changes between 2.x and 3.x and there won't likely be changes in the future. I almost always just dump those right into the root of the *samples_snapshots* folder.

guardrex: I just took a look at the diff. I don't think this is the best way. Version the sample code from each sample, even if there aren't changes. Also note that the PU has stated that they really dislike the *full topic versioning* approach. It makes topic updates *very hard for them to review*. I've also noted that it's a pain point to mod every section of the ones that duplicate the whole topic up to three times.

```

3.x sample link here

::: moniker-end

2.x sample link here

::: moniker-end
```

guardrex: ... but I'll add that it has been a *necessary evil* at times. Some of the topics *wildly flipped* in and out of versioning blocks ... those had to use *full topic versioning*. For example, Kestrel. *ei ei ei* ... it was a 🚂💥 before we changed it over to *full topic versioning*.

serpent5: Thanks @guardrex. That's very valuable feedback. I try to make decisions based on things I see elsewhere in existing samples, recent PRs, etc, but I'll admit that can be problematic. It's not always clear what the current thinking is and naturally different samples do differing things.

If you have the time, can you turn this into a task-list for me? You can dial the nit level up to _crazy_ too. I'm not sure what to do with this now.

guardrex: I’m at the gym now 💪. When I get back to the house, I’ll see if I can find the original issue where the sample layout decisions were made. You can look at mine btw on topics that I wrote. They’re highly consistent with that guidance. Ping u back in a bit.

guardrex: Btw just for fun on the subject of consistency and guidelines ... it’s been a long-standing concern of mine. See 👉 https://github.com/aspnet/AspNetCore.Docs/issues/2617

Rick-Anderson: I typically make a 3.x copy of the code. If there are no changes, just use the 2.x code and don't change the path.

guardrex: I *think* I've tracked down the very first discussion about sample layouts ... it seems to start with my remark here and follows with a discussion ...

https://github.com/aspnet/AspNetCore.Docs/pull/3896#issuecomment-322605262

I noticed that the PR's that I put in just after that ... several sample updates in late August of 2017 ... were the first to use it.

@serpent5, of course, follow Rick's advice. I'd prefer to stick with what I've been doing. I think it's the fastest 🕐, least confusing 😵, and easiest ❤️ way to go ... i.e., I think it's the lowest cost **_$$💰$$_** approach in the long run.

I'll try to use the *whole topic versioning* only when really desperately needed given my own hardships with it and the complaints from engineers about it.

I'll make a quick pass on my old 🐮 **_BEEF!_** 🐄 issue just in case I need to call it out again. It will only take a few minutes to fix up.

guardrex: I'll mention one other change tho since then. Back in those days, I think we felt that samples should named (the project folder should be named) with something specific to the sample (e.g., *ConfigurationSample* or *ChangeTokensSample*). I liked that (still do) because when I have multiple samples open, it's easier to find them in tabs and tooltips in IDEs.

Later tho ... after a bunch of samples were named that way, the guidance shifted to naming all sample apps *SampleApp*. IIRC, the reasoning had to do with cut-'n-paste, but I don't think ever felt too good about that ... devs are going to have to change namespaces no matter what.

It's only really a problem in those rare cases where you have more than one sample at each version ... for example a console app sample and an ASP.NET Core sample in each of the 2.x and 3.x folders. Otherwise, they would be called something like *SampleApp1* and *SampleApp2*, which is kind'a funky. Anyway ... that's *very rare*.

Just calling that out for you @serpent5: The guidance is to call every sample app (project folder) *SampleApp*. You don't have to change them; but if you were creating a **_new_** topic+sample(s) from scratch, that's the best way to go.

Rick-Anderson: I've never heard the SampApp advice. I frequently use WebApp1 or something related to the sample - WebConfiguration

guardrex: You told me to do that. :smile: lol

guardrex: I can look it up if you want ... but it had to do with namespaces and cut-'n-paste IIRC. I think you said it would be easier for devs to consume our code bits in their own apps. It was a long time ago that it was discussed. I've been making all **_new_** samples use *SampleApp*.

Rick-Anderson:
> Also note that the PU has stated that they really dislike the full topic versioning approach. It makes topic updates very hard for them to review. 

That's when you make the jump from mixed versions to full versions. The new approach is do a PR **before** the updates to make the content full version ready. That's what we've been doing for months now

```
WARNING, this topic has not been updated to 3.1
Full copy of the 2.x doc here
::: moniker-end

2.x content here

::: moniker-end

```

Then merge the preceding PR. Now update the 3.x moniker section and you get a clean diff of what's changed from 2.x to 3.0
That's what the folks I've been working with prefer. 

> I've also noted that it's a pain point to mod every section of the ones that duplicate the whole topic up to three times.

Why would you need to do that? You never touch the older sections unless it's significant bug. The older moniker sections are essentially archived.

guardrex:
That's cool, but it doesn't provide the normal "this topic isn't available for your version" banner ...

https://github.com/aspnet/AspNetCore.Docs/issues/13887

 If we do the metadata version, we get that nice banner. Maybe both should be done ... at least only for the *Fundamentals* node topics, which seem to be many of the most important topics on the repo.

I would've used the technique that you described, but you know how I am on sample/topic updates. I'm **_CRAZY_** 🏃🏃🏃🏃🏃 **_FAST!_** ..... so it wasn't something that I could make much use of. I generally have my topics+samples updated within a week of a release.

guardrex:
> Why would you need to do that? You never touch the older sections unless it's significant bug. 

That hasn't been the case for me ... I'm *very often* having to dig into earlier version content to make matching updates. Perhaps, it's just a temporary thing.

Rick-Anderson:
> That's cool, but it doesn't provide the normal "this topic isn't available for your version" banner ...

That's OK as it's only short term. You do this just before the PR to update the topic. Really clean diff and that's what's important.

guardrex:
The really good news is that *Fundamentals* node churn seems *low*, and I think it will stay that way. I find myself just doing straight sample updates. Other text that pertains to the next release, I've been able to update with versioning in the topic without sample updates. It's been good. Very happy with the 3.1 release ... that was *very smooth*.

guardrex:
I don't think that there are any significant deltas for @serpent5 on this PR given that you're doing it this way on the 2.x vs. 3.x links ... i.e., leaving a single 2.x link in place in what is 3.x content.

You can let him know if you prefer the *full topic versioning*. I did do it ... don't get me wrong ... I put a bunch of topics on that scheme. I'm just hitting some problems ... the churn is affecting engineering reviews + forcing me to have to update more content that I would like to have updated. Consequently, that scheme for these topics has put a dent into 💰 *company profitability* 💰. Hopefully, these problems will go away, as you say, after the first pass on a topic to set it up.

I'd like to keep the Blazor stuff on the ... er ... *2017 System:tm:* :smile: lol ... it seems easiest for me given the churn right now looking out toward 3.2 and 5.0.

serpent5:
> i.e., leaving a single 2.x link in place in what is 3.x content.

If this is referring to the part I called out, it’s the other way around. The topic shows the referenced sample code in both 2.x and 3.x, but it’s using the new 3.0 MVC service registration approach.

Rick-Anderson:
>  but it’s using the new 3.0 MVC service registration approach.

That will need it's own snippet obviously.

serpent5:
I’ll sort that out later tonight. Is there anything else I should change/do?

Rick-Anderson:
>I’ll sort that out later tonight. Is there anything else I should change/do?

Looks good.  You might add the changes I proposed in #16232

serpent5:
Yeah, sure, I’ll do that. I was going to review your PR later anyway so I’ll pull the changes in.

aspnetcore/mvc/models/model-binding.md
```diff
@@ -81,7 +81,7 @@ By default, properties are not bound for HTTP GET requests. Typically, all you n
 Model binding includes <xref:Microsoft.AspNetCore.Mvc.ControllerBase.TryUpdateModelAsync*>, which uses value providers to get data from the form body, query string, and route data. `TryUpdateModelAsync` is typically:
 
 * Used with Razor Pages and MVC apps using controllers and views.
-* Is generally not used with Web API. Web API uses [Input formatters](#input-formatters) to deserialize the request body into an object.
+* Not used with Web API. Web API uses [Input formatters](#input-formatters) to deserialize the request body into an object.
```
serpent5: Tweaked this 'cos we've already got "is typically" leading into the bullets.

serpent5:
@Rick-Anderson I'm all done.

Rick-Anderson:
@serpent5  looks good. Did you want to consider #15705 in another PR?

serpent5:
> Did you want to consider #15705 in another PR?

Yeah. That's next. I figured this approach would allow Javier to review only the requested change without a big sample update getting in the way. I'll work on it tomorrow.

Rick-Anderson:
> Yeah. That's next. I figured this approach would allow Javier to review only the requested change without a big sample update getting in the way. 

Exactly, that's how we should do all our PR's. A clean diff with only what they need to review. cc @guardrex

Issue #16118, opened 2019-12-10
JSRuntime.Current no longer exists


LaughingJohn:
See this sentence:
Multiple user requests are processed by the Blazor Server app. Don't call JSRuntime.Current in a component to invoke JavaScript functions.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 49807900-0d56-e23c-1aef-d79099caeffb
* Version Independent ID: feff434f-912c-0985-d575-4a23a629057d
* Content: [ASP.NET Core Blazor JavaScript interop](https://docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interop?view=aspnetcore-3.1#feedback)
* Content Source: [aspnetcore/blazor/javascript-interop.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/javascript-interop.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Hello @LaughingJohn ...

🤔 It was reviewed by engineering for the 3.0 release (not sure if it changed for 3.1). I'll put this up for triage so that engineering can respond. Stand-by tho if it takes a while to get an answer ... we're **_very_** close to the holidays 🎁⛄️, so they might not remark on it until January.

mkArtakMSFT:
@guardrex JSRuntime.Current is indeed gone. The JSRuntime services are available only through DI now. This is true for both 3.0 and 3.1.

guardrex:
Fixed by https://github.com/aspnet/AspNetCore.Docs/pull/16246.

PR #16227, opened 2019-12-16
Drop JS interop guidance remark

guardrex:
Fixes #16118

Dropping the remark on calling `JSRuntime.Current` leaves an out-of-context, nonsensical remark (1st sentence), so I remove the entire bullet point here.

Thanks @LaughingJohn! 🌴

PR #16246, opened 2019-12-17
Update javascript-interop.md

pranavkm:


guardrex:
This seems to fix #16118 on #16227, so I'll close that PR.

Issue #16241, opened 2019-12-17
UseAuthorization in Built-in middleware


taixes:
In the Built-in middleware section there is no description for the Authorization built-in middleware (Core 3.0). Shouldn't it be added?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 9898482c-fa54-7ce2-1716-d14fbe03990d
* Version Independent ID: 3926c686-0ccc-f9fe-ca52-0134612a1623
* Content: [ASP.NET Core Middleware](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/?view=aspnetcore-3.1)
* Content Source: [aspnetcore/fundamentals/middleware/index.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/fundamentals/middleware/index.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-fundamentals**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

PR #16251, opened 2019-12-17
refactor middleware monikers

Rick-Anderson:
Just a refactor to make #16241 easier for PU to review

### DIFF
--- a/aspnetcore/fundamentals/middleware/index.md
+++ b/aspnetcore/fundamentals/middleware/index.md
@@ -10,6 +10,8 @@ uid: fundamentals/middleware/index
 ---
 # ASP.NET Core Middleware
 
+::: moniker range=">= aspnetcore-3.0"
+
 By [Rick Anderson](https://twitter.com/RickAndMSFT) and [Steve Smith](https://ardalis.com/)
 
 Middleware is software that's assembled into an app pipeline to handle requests and responses. Each component:
@@ -58,8 +60,6 @@ When a delegate doesn't pass a request to the next delegate, it's called *short-
 The order that middleware components are added in the `Startup.Configure` method defines the order in which the middleware components are invoked on requests and the reverse order for the response. The order is **critical** for security, performance, and functionality.
 
 The following `Startup.Configure` method adds security related middleware components in the recommended order:
-
-::: moniker range=">= aspnetcore-3.0"
 
 [!code-csharp[](index/snapshot/StartupAll3.cs?name=snippet)]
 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
cc @guardrex  we refactor before addressing PR's so diff is better

guardrex:
... and if the content changes at ... say ... 3.2? Do you plan to duplicate that whole block again for a few lines, or do you want to cut in 3.2 content into the top block (keeping in mind that these versioning blocks can't be nested)?

Rick-Anderson:
case by case. Really no problem with a full dup.  We need to maintain 3.1 as it's LTS. We need to keep 3.2 until probably 5.1

guardrex:
I'm just perhaps overly worried about it ... thus far, it seems like it's taking me longer to manage them 💸. Just a 'feel'in' tho ... I don't have hard figures. Thus far, concerns are ...

* Engineering freak-outs looking at whole new blocks. I think ur suggesting to avoid this with a PR before the content updates that sets up *whole topic versioning*. Then, they look at the next PR that has the actual content updates and don't see a *wall of green*.
* Having to edit the same content in multiple spots. You said it shouldn't happen too much ... I hope ur right. Thus far, I've been having to do it a little too much. We're headed for three blocks: <3.0 (2.1 goes deep into 2021), 3.0-3.2 block (if we're lucky one block), and then 5.0 lands (maybe late 2020). If ur right, the older content will be so stable that it won't require much attention.
* Nesting isn't allowed, and we might cut in blocks with the *whole topic* approach, which defeats the purpose of the approach and could get a little messy. AFAICT, cut-ins were minimal for 3.1. Hopefully, that trend will continue for 3.2.

For Blazor topics, I've had good luck thus far with just plain versioning blocks ... the old way ... because there's just 3.0 and 3.1 at this point. There aren't enough deltas to sweat, and there's only one pair of sample apps at 3.1 (*3.x* folder). That batch doesn't use the *whole topic* approach yet because there hasn't been a major version release. The 5.0 release (maybe November, 2020) it will need to be done to this batch of topics.

Rick-Anderson:
> ... the old way ... because there's just 3.0 and 3.1 at this point.

Why are you maintaining 3.0? No one is using 3.0. We have approval to update to 3.1 without keeping 3.0.

> Then, they look at the next PR that has the actual content updates and don't see a wall of green.

They actually see far less than the old way. You can roll in all the changes they don't need to see.

guardrex:
It's 3.x actually (with 3.1 cut-ins). I just meant that the content hasn't changed since 3.0, so the topics version all of the content with a metadata version.

Rick-Anderson:
Please don't use 3.1 monikers cut in's. That makes it harder to maintain should you ever leave. Just update to 3.1 and don't do anything to maintain a 3.0 version.

guardrex:
Wait ... what? You're not suggesting setting the metadata version to 3.1 are you? 3.0 is supported until next March. Or, are you saying to go ahead and whole topic version them now for 3.0 and 3.1? That can be done ... but ei ei ei that's a *crazy amount* of duplicated content. We're still changing text quite frequently that would apply to both versions.

guardrex:
> Just update to 3.1 and don't do anything to maintain a 3.0 version.

Yeah ... I don't quite follow you. Which way do you want to go? ...

1. Set metadata version to 3.1 and drop all 3.0 content now. (But again, 3.0 is supported until March.)
2. Use *whole topic versioning* now. A block for 3.0 and a block for all 3.1. In March, kill the 3.0 and update the topic's metadata version to 3.1.

guardrex:
> Please don't use 3.1 monikers cut in's. 

btw -- No choice on that. Almost all of the 3.1 content was merged **before** 3.1 GA. There was no way to drop 3.0 at that time, of course.

Rick-Anderson:
No need to maintain 3.0, there's zero demand for it.

I'd wait until you update the article, don't do anything now. Next time you touch the article I'd say Set metadata version to 3.1 and drop all 3.0-only content now.

It doesn't hurt to have 3.0, but there's also zero benefit.

guardrex:
Ok, cool.

Do you want *whole topic versioning* closer to 3.2 (3.1 and 3.2 blocks), or is that better for closer to 5.0 (3.1, 3.2, and 5.0 blocks)?

Rick-Anderson:
Let's see how much changes for 3.2
I wish they'd go from 3.1 to 5.0

guardrex:
ok ... sounds good.

PR #16307, opened 2019-12-24
Add missing space to code sample

ATimmeh33:
Would work without the space, but all the other samples have 1, so why not?


guardrex:
Hello @ATimmeh33 ... Thanks, but the code is correct. There shouldn't be space after the colon in this line because there's no data to show after the colon in this case. It just needs a line break there. The headers appear on different lines after this line is written.

ATimmeh33:
I stand corrected, my bad.

PR #16343, opened 2019-12-28
Add assembly name as a parameter in JS call to 'invokeMethodAsync'

GregTrevellick:


GregTrevellick:
Hmmm...I'm not so sure about this PR now:

My change is for a JS call to an instance method, and since the instance of the class itself is passed to the JS function, the JS may not need to know the assembly name as it has the instance itself.

Well, critique the PR and merge or reject as appropriate - either way I'll earn something :-)

guardrex:
Correct ... passed by reference in `dotnetHelper` (`HelloHelper` wrapped as a `DotNetObjectReference`) ... it *knows* :wink: the assembly. Whereas for `DotNet.invokeMethod` or `DotNet.invokeMethodAsync` (static method `ReturnArrayAsync` in the sample app), the assembly name must be provided.

PR #16351, opened 2019-12-31
Adding a note for a common configuration error in cert Auth

damienbod:
A lot of cert auth errors are caused by incorrect DNS. Just added a note to the docs for this


scottaddie:
Thank you for the PR, @damienbod!

damienbod:
Thanks for the fixes

PR #16353, opened 2019-12-31
Adding the default values for the certificate authentication options

damienbod:
adding some defualt values for the certificate authentication options


scottaddie:
Thanks again for your help, @damienbod!

PR #16354, opened 2019-12-31
Adding HttpClient examples to send the certificate (cert_auth.md)

### LINKED ISSUE #15974: Do not set X-ARR-ClientCert header on HttpClient
The example code on [this](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/certauth?view=aspnetcore-3.0#implement-an-httpclient-using-a-certificate) section seems to imply that you should configure a `HttpClient` to send along a client certificate for mutual TLS by including the client certificate as part of the _X-ARR-ClientCert_ header:
```
request.Headers.Add("X-ARR-ClientCert", cert.GetRawCertDataString());
```

 This seems wrong since the appropriate way to do so is setting the certificate on the `HttpClient`'s `HttpClientHandler`:
```
httpClientHandler.ClientCertificates.Add(certificate);
````

The confusion appears to come from the fact that an Azure web frontend server will forward the supplied client certificate to the the web app/api backend server in the `X-ARR-ClientCert` request header. The frontend web server however only does this after it verified that the client is in possession of the private key belonging to the certificate as part of the [Certificate Verify](https://tools.ietf.org/html/rfc5246#section-7.4.8) phase of the TLS negotiation. 

It seems to me that by sending the _X-ARR-ClientCert_ as part of the initial `HttpClient` request you run the risk that on an improperly configured system you can trick the backend web server from thinking a proper client certificate was presented by only sending the (public) client certificate, since no proof of possession check is performed. This might therefore cause authentication/authoriation bugs.

---
#### Document Details

* ID: 35b7960d-6de5-78ee-bc45-d7218239bbc8
* Version Independent ID: d2db534b-198a-dc99-5b5a-0525582d39ac
* Content: [Configure certificate authentication in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/certauth?view=aspnetcore-3.0#implement-an-httpclient-using-a-certificate)
* Content Source: [aspnetcore/security/authentication/certauth.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/security/authentication/certauth.md)
* Product: **aspnet-core**
* GitHub Login: @blowdart
* Microsoft Alias: **bdorrans**

### PR DESCRIPTION
**Pull request by damienbod**
Added HttpClient examples, showing different ways to send the certificate in a HttpClient request
Fixes #15974

### DIFF
--- a/aspnetcore/security/authentication/certauth.md
+++ b/aspnetcore/security/authentication/certauth.md
@@ -221,20 +221,49 @@ See the [host and deploy documentation](xref:host-and-deploy/proxy-load-balancer
 
 ### Use certificate authentication in Azure Web Apps
 
+No forwarding configuration is required for Azure. This is already setup in the certificate forwarding middleware.
+
+### Use certificate authentication in custom web proxies
+
 The `AddCertificateForwarding` method is used to specify:
 
 * The client header name.
 * How the certificate is to be loaded (using the `HeaderConverter` property).
 
-In Azure Web Apps, the certificate is passed as a custom request header named `X-ARR-ClientCert`. To use it, configure certificate forwarding in `Startup.ConfigureServices`:
+In custom web proxies, the certificate is passed as a custom request header for example `X-SSL-CERT`. To use it, configure certificate forwarding in `Startup.ConfigureServices`:
 
 ```csharp
 public void ConfigureServices(IServiceCollection services)
 {
     services.AddCertificateForwarding(options =>
     {
-        options.CertificateHeader = "X-ARR-ClientCert";
+        options.CertificateHeader = "X-SSL-CERT";
+        options.HeaderConverter = (headerValue) =>
+        {
+            X509Certificate2 clientCertificate = null;
+	    
+            if(!string.IsNullOrWhiteSpace(headerValue))
+            {
+                byte[] bytes = StringToByteArray(headerValue);
+                clientCertificate = new X509Certificate2(bytes);
+            }
+
+            return clientCertificate;
+        };
     });
+}
+
+private static byte[] StringToByteArray(string hex)
+{
+    int NumberChars = hex.Length;
+    byte[] bytes = new byte[NumberChars / 2];
+
+    for (int i = 0; i < NumberChars; i += 2)
+    {
+        bytes[i / 2] = Convert.ToByte(hex.Substring(i, 2), 16);
+    }
+
+    return bytes;
 }
 ```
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by damienbod** [comment]
https://github.com/aspnet/AspNetCore.Docs/issues/15974

**Comment by guardrex** [comment]
btw @damienbod ... If the PR "fixes" the issue, use `Fixes #{ISSUE NUMBER}` so that when the PR merges the issue will automatically close.

**Comment by Rick-Anderson** [comment]
@damienbod  like this
Fixes #15974

File: aspnetcore/security/authentication/certauth.md [hunk 1]
```diff
@@ -311,7 +311,85 @@ namespace AspNetCoreCertificateAuthApi
...
+        var responseContent = await response.Content.ReadAsStringAsync();
+        var data = JsonDocument.Parse(responseContent);
+        return data;
+    }
+ 
+    throw new ApplicationException($"Status code: {response.StatusCode}, Error: {response.ReasonPhrase}");
+}
+```
+
+#### Implement an HttpClient using a certificate and the X-ARR-ClientCert HTTP Header
```
etortec: I'd suggest to totally remove the now obsolete `Implement an HttpClient using a certificate and the X-ARR-ClientCert HTTP Header`, or at least to add a warning that the example code in that section can solely be used for API testing purposes and is not how you should implement the API client (as explained in #15974).

damienbod: @etortec @blowdart I removed this and updated the forwarding handling for Azure as this is an Azure example

damienbod:
Fixes #16621

damienbod:
@blowdart Are we good here now? 

I removed the forwarding for Azure. And added a section for custom web proxies which uses certificate forwarding like before. Thanks for your help.

blowdart:
Yea, as long as you've called out the middleware still needs to be present. Just doesn't have to be configured.

damienbod:
@blowdart Added the note

damienbod:
Hi @blowdart Would it be possible to review this?

Thanks Damien

damienbod:
@Rick-Anderson @scottaddie 

Anything we need to do to get this completed?

Greetings Damien

jim-brown-fst:
> Yea, as long as you've called out the middleware still needs to be present. Just doesn't have to be configured.

When you say the middleware needs to be present, does that mean this won't work with Azure Functions? I can't add the middleware, but is there a workaround I can use without the middleware to get Azure to forward my certificate in the header?

PR #16365, opened 2020-01-01
Leveraging string interpolation

MaherJendoubi:
String interpolation provides a more readable and convenient syntax.


guardrex:
Hello @MaherJendoubi ... templates are used on purpose and recommended for all logging messages because they provide semantic/structured logging. See :point_right: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-3.1

PR #16404, opened 2020-01-05
Add gRPC versioning article

JamesNK:
~@Rick-Anderson  edit: Current work at #16417~ Merged

Fixes https://github.com/grpc/grpc-dotnet/issues/604

Internal preview URL: https://review.docs.microsoft.com/en-us/aspnet/core/grpc/versioning?view=aspnetcore-3.0&branch=pr-en-us-16404

### DIFF
--- a/aspnetcore/grpc/versioning.md
+++ b/aspnetcore/grpc/versioning.md
@@ -43,11 +43,12 @@ These changes are non-breaking at a gRPC protocol level, and .NET binary level.
 
 ### Binary breaking changes
 
-The following changes are non-breaking at a gRPC protocol level, but the client needs to be updated if it upgrades to the latest *.proto* contract or client .NET assembly.
+The following changes are non-breaking at a gRPC protocol level, but the client needs to be updated if it upgrades to the latest *.proto* contract or client .NET assembly. Binary compatibility is important if you plan to publish a gRPC library to NuGet.
 
 - **Removing a field** - Values from a removed field are deserialized to a message's [unknown fields](https://developers.google.com/protocol-buffers/docs/proto3#unknowns). This isn't a gRPC protocol breaking change, but the client needs to be updated if it upgrades to the latest contract. It is important that a removed field number isn't accidentally reused in the future. One way to make sure this doesn't happen is to specify deleted field numbers and names on the message using Protobuf's [`reserved`](https://developers.google.com/protocol-buffers/docs/proto3#reserved) keyword.
 - **Renaming a field** - Field names are only used in generated code. The field number is used to identify fields on the network. The client will need to be updated if it upgrades to the latest contract.
 - **Renaming a message** - Message names are not sent on the network so this isn't a gRPC protocol breaking change, but the client will need to be updated if it upgrades to the latest contract.
+- **Changing csharp_namespace** - Changing `csharp_namespace` will change the namespace of generated .NET types. This isn't a gRPC protocol breaking change, but the client needs to be updated if it upgrades to the latest contract.
 
 ### Breaking changes
 
--- a/aspnetcore/grpc/versioning.md
+++ b/aspnetcore/grpc/versioning.md
@@ -29,7 +29,7 @@ Making non-breaking changes to a service has a number of benefits:
 This content focuses on whether changes are **breaking at a gRPC protocol and .NET binary compatibility level**. When making changes, consider whether older clients can logically continue working. For example, adding a new field to a request message:
 
 * Is not a protocol breaking change.
-* Setting an error condition if the new field is not set makes it a breaking change for old clients.
+* Returning an error status on the server if the new field is not set makes it a breaking change for old clients.
 
 ### Non-breaking changes
 
@@ -37,9 +37,9 @@ These changes are non-breaking at a gRPC protocol level, and .NET binary level.
 
 - **Adding a new service**
 - **Adding a new method to a service**
-- **Adding a field to a request message** - Fields added to a request message are deserialized with the [default value](https://developers.google.com/protocol-buffers/docs/proto3#default) on the server when not set.
+- **Adding a field to a request message** - Fields added to a request message are deserialized with the [default value](https://developers.google.com/protocol-buffers/docs/proto3#default) on the server when not set. To be non-breaking the service will need to succeed when it is not set by older clients.
 - **Adding a field to a response message** - Fields added to a response message are deserialized into the message's [unknown fields](https://developers.google.com/protocol-buffers/docs/proto3#unknowns) collection on the client.
-- **Adding a value to an enum** - Enums are serialized as a numeric value. New enum values are deserialized on the client to the enum value without an enum name.
+- **Adding a value to an enum** - Enums are serialized as a numeric value. New enum values are deserialized on the client to the enum value without an enum name. To be non-breaking older clients will need to run correctly when they receive and unexcepted value.
 
 ### Binary breaking changes
 
@@ -84,10 +84,14 @@ app.UseEndpoints(endpoints =>
 });
 ```
 
-Including a [SemVer major version number](https://semver.org/) in the package name gives you the opportunity to publish a *v2* version of your service with breaking changes, while continuing to support older clients who call the *v1* version. Once clients have updated to use the *v2* service you can choose to remove the old version.
+Including a version number in the package name gives you the opportunity to publish a *v2* version of your service with breaking changes, while continuing to support older clients who call the *v1* version. Once clients have updated to use the *v2* service you can choose to remove the old version. When versioning your services:
 
-To avoid duplication, consider moving business logic from the service implementations to a centralized location that can be reused by the old and new implementations:
+- Avoid breaking changes if reasonable.
+- Do not update the version number unless making breaking changes.
+- Do update the version number when you make breaking changes.
+
+Publishing multiple versions of a service duplicates it. To reduce duplication, consider moving business logic from the service implementations to a centralized location that can be reused by the old and new implementations:
 
 [!code-csharp[](versioning/sample/GreeterServiceV1.cs?highlight=10,19)]
 
-Services and messages generated from different packages are **different .NET types**. Moving business logic to a centralized location requires mapping messages to common types.
+Services and messages generated with package names are **different .NET types**. Moving business logic to a centralized location requires mapping messages to common types.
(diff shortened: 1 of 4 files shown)


JamesNK:
@Rick-Anderson I committed some changes without realizing you're making edits. Ignore my recent changes 😄

aspnetcore/grpc/versioning.md
```diff
@@ -0,0 +1,93 @@
...
+* Is not a protocol breaking change.
+* Setting an error condition if the new field is not set makes it a breaking change for old clients.
+
+### Non-breaking changes
+
+These changes are non-breaking at a gRPC protocol level, and .NET binary level.
+
+- **Adding a new service**
+- **Adding a new method to a service**
+- **Adding a field to a request message** - Fields added to a request message are deserialized with the [default value](https://developers.google.com/protocol-buffers/docs/proto3#default) on the server when not set.
```
rynowak: should this say `adding an optional field...`?

aspnetcore/grpc/versioning.md
```diff
@@ -0,0 +1,93 @@
...
+- **Adding a field to a response message** - Fields added to a response message are deserialized into the message's [unknown fields](https://developers.google.com/protocol-buffers/docs/proto3#unknowns) collection on the client.
+- **Adding a value to an enum** - Enums are serialized as a numeric value. New enum values are deserialized on the client to the enum value without an enum name.
```
rynowak: I feel like this needs a little bit more elucidation - it's correct that this is not a protocol level break - but it probably represents a behavior change that the client is not prepared for.

aspnetcore/grpc/versioning.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Breaking changes
+
+These are protocol and binary breaking changes.
+
+- **Changing a field data type** - Changing a field's data type to an [incompatible type](https://developers.google.com/protocol-buffers/docs/proto3#updating) will cause errors when deserializing the message. Even if the new data type is compatible, it is likely the client will need to be updated to support the new type if it upgrades to the latest contract.
+- **Changing a field number** - The field number is used to identify fields on the network.
+- **Renaming a package, service or method** - gRPC uses the package name, service name and method name to build the URL. The client gets an *UNIMPLEMENTED* status from the server.
+- **Removing a service or method** - The client gets an *UNIMPLEMENTED* status from the server when calling the removed method.
+
```
rynowak: What would you think about adding some additional stuff here about breaking changes that aren't protocol or codegen changes.
- Expanding the allowed range of a return-value (adding an enum value)
- Adding additional validate rules

aspnetcore/grpc/versioning.md
```diff
@@ -0,0 +1,93 @@
...
+These changes are non-breaking at a gRPC protocol level, and .NET binary level.
+
+- **Adding a new service**
+- **Adding a new method to a service**
+- **Adding a field to a request message** - Fields added to a request message are deserialized with the [default value](https://developers.google.com/protocol-buffers/docs/proto3#default) on the server when not set.
+- **Adding a field to a response message** - Fields added to a response message are deserialized into the message's [unknown fields](https://developers.google.com/protocol-buffers/docs/proto3#unknowns) collection on the client.
+- **Adding a value to an enum** - Enums are serialized as a numeric value. New enum values are deserialized on the client to the enum value without an enum name.
+
+### Binary breaking changes
```
rynowak: I find the term "binary breaking change" a little confusing here because it's either *source-breaking* or *source-breaking-and-binary-breaking* considering whether you consume the `.proto` or you consume a generated .NET package.

I don't have a better suggestion.

aspnetcore/grpc/versioning.md
```diff
@@ -0,0 +1,93 @@
...
+{
+    // Implements greet.v1.Greeter
+    endpoints.MapGrpcService<GreeterServiceV1>();
+
+    // Implements greet.v2.Greeter
+    endpoints.MapGrpcService<GreeterServiceV2>();
+});
+```
+
+Including a [SemVer major version number](https://semver.org/) in the package name gives you the opportunity to publish a *v2* version of your service with breaking changes, while continuing to support older clients who call the *v1* version. Once clients have updated to use the *v2* service you can choose to remove the old version.
```
rynowak: Rather than just saying *do semver* - I'd suggest explicitly including the guidance to:
- Avoid breaking changes if reasonable
- Do not update the version unless it contains breaking changes
- Do update the version when you make breaking changes

aspnetcore/grpc/versioning.md
JamesNK: Originally I had "source/binary breaking change" everywhere but I simplified it to binary because:
1. If you're updating proto file then updating source code to match probably isn't a big deal. And if it then you can choose not to do it.
2. I think binary is more important consideration because it covers distributing gRPC clients in NuGet packages, and blindly pushing out a breaking change in a NuGet package is the bigger danger here.

aspnetcore/grpc/versioning.md
JamesNK: All gRPC fields are optional. I know what you mean though 😄

aspnetcore/grpc/versioning.md
JamesNK: I think that is a never ending subject. At the top of this section I say that you need to determine whether older clients can logically still run.

aspnetcore/grpc/versioning.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Non-breaking changes
+
+These changes are non-breaking at a gRPC protocol level, and .NET binary level.
+
+- **Adding a new service**
+- **Adding a new method to a service**
+- **Adding a field to a request message** - Fields added to a request message are deserialized with the [default value](https://developers.google.com/protocol-buffers/docs/proto3#default) on the server when not set. To be non-breaking the service will need to succeed when it is not set by older clients.
+- **Adding a field to a response message** - Fields added to a response message are deserialized into the message's [unknown fields](https://developers.google.com/protocol-buffers/docs/proto3#unknowns) collection on the client.
+- **Adding a value to an enum** - Enums are serialized as a numeric value. New enum values are deserialized on the client to the enum value without an enum name. To be non-breaking older clients will need to run correctly when they receive and unexcepted value.
```
JunTaoLuo: I suppose we are only highlighting the most prevalent scenarios right? 

For example, though adding a value to an enum is non-breaking, we didn't cover removing enum values. Also, there are more esoteric cases, such as options like `csharp_namespace`.

aspnetcore/grpc/versioning.md
```diff
@@ -0,0 +1,97 @@
...
+{
+    // Implements greet.v1.Greeter
+    endpoints.MapGrpcService<GreeterServiceV1>();
+
+    // Implements greet.v2.Greeter
+    endpoints.MapGrpcService<GreeterServiceV2>();
+});
+```
+
+Including a version number in the package name gives you the opportunity to publish a *v2* version of your service with breaking changes, while continuing to support older clients who call the *v1* version. Once clients have updated to use the *v2* service you can choose to remove the old version. When versioning your services:
```
JunTaoLuo: I wonder if we should do this in our templates. @shirhatti

aspnetcore/grpc/versioning.md
JamesNK: It's not an exhaustive list.

`csharp_namespace` is a good one though. I'll include that.

aspnetcore/grpc/versioning.md
```diff
@@ -0,0 +1,98 @@
...
+
+### Binary breaking changes
+
+The following changes are non-breaking at a gRPC protocol level, but the client needs to be updated if it upgrades to the latest *.proto* contract or client .NET assembly. Binary compatibility is important if you plan to publish a gRPC library to NuGet.
+
+- **Removing a field** - Values from a removed field are deserialized to a message's [unknown fields](https://developers.google.com/protocol-buffers/docs/proto3#unknowns). This isn't a gRPC protocol breaking change, but the client needs to be updated if it upgrades to the latest contract. It is important that a removed field number isn't accidentally reused in the future. One way to make sure this doesn't happen is to specify deleted field numbers and names on the message using Protobuf's [`reserved`](https://developers.google.com/protocol-buffers/docs/proto3#reserved) keyword.
+- **Renaming a field** - Field names are only used in generated code. The field number is used to identify fields on the network. The client will need to be updated if it upgrades to the latest contract.
```
jtattermusch: renaming a field is discouraged. If JSON formatting or reflection is used, things can potentially break - basically, it's a thin ice. 
same for renaming messages (this can also affect the Any type).

Rick-Anderson:
@jtattermusch  this merged yesterday. Feel free to open an issue on the doc. you can create an issue from the doc page? That will connect the doc to the issue. Go to the bottom of the page and select **This page** .

image

chwarr:
gRPC (but not yet grpc-dotnet) supports arbitrary encoding schemes for method payloads. What was merged here is almost entirely about using Protocol Buffers with gRPC. How should this distinction be communicated?

JamesNK:
The detail on renaming a field is the main one that is specific to Protobuf. I'll expand the description to mention that it is breaking for JSON.

I think going into detail for every serialization technology is not realistic. If you go off the standard path then you'll have to figure it out on your own.

chwarr:
gRPC really is just method routing, headers, trailers, status, message framing, and message cardinality. Thus, wouldn't "gRPC versioning" only talk about things like adding/removing service methods, changing cardinality, and adding & removing headers? If the wording were changed to "gRPC+ProtoBuf versioning" then all the talk about field changes are in scope. Or, those could just be turned into a pointer to a general ProtoBuf versioning document.

Talking about "gRPC+ProtoBuf" avoids needing to go into the detail for every serialization technology while remaining true to the layers that are right behind the nice-to-program-against abstraction. 

Talking just about gRPC and pointing to ProtoBuf will allow people to compose the right answer themselves when they aren't using ProtoBuf, but hurts those who are using the default.

Thus, my suggestion would be to rename & scrub this document to talk about "gRPC Protocol Buffer services" instead of just bare "gRPC services" (or however the style guide recommends combining gRPC with a given serialization technology.)

JamesNK:
The gRPC HTTP/2 protocol is just the envolope spec + framing. The wider gRPC ecosystem includes the contact file (proto) which has services + messages + data types. This doc talks about versioning in the context of the concepts from the proto contract file.

And yes you can have gRPC without the proto contract file, but the vast majority of people will use it, and if you're not then you should know that a lot of advice that applies to gRPC probably doesn't apply to your custom solution.

chwarr:
Thanks. That helps me understand better how you see things.

jtattermusch:
@JamesNK  FTR this is the talk I mentioned https://www.youtube.com/watch?v=F2WYEFLTKEw

PR #16417, opened 2020-01-06
Update versioning.md

Rick-Anderson:
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/grpc/versioning?view=aspnetcore-3.1&branch=pr-en-us-16417)

- On the wire -> On the network. Wire doesn't MT (Machine translate). Wire is better if you understand the jargon in english. Perhaps HTTP2 framing would be better.
- This doc has nothing to do with ASP.NET Core's gRPC. This is about gRPC. I don't see this at https://grpc.io/docs/guides/ - that's the right place for this doc.
- remove bling and use **bold** for emphasis.
- long sentences don't MT and are difficult to parse for native speakers.

### DIFF
--- a/aspnetcore/grpc/versioning.md
+++ b/aspnetcore/grpc/versioning.md
@@ -66,7 +66,7 @@ gRPC supports an optional [`package`](https://developers.google.com/protocol-buf
 
 [!code-protobuf[](versioning/sample/greet.v1.proto?highlight=3)]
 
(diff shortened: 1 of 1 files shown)


JamesNK:
> This doc has nothing to do with ASP.NET Core's gRPC. This is about gRPC. I don't see this at grpc.io/docs/guides - that's the right place for this doc.

The .NET binary compatibility is specific to .NET, and the code sample.

Issue #353, opened 2015-07-30
Update Diagnostics to beta6


danroth27:

danroth27:
b05b4df3e7f23b99226f5f192c16ede008822b0f

PR #16421, opened 2020-01-06
Service config execution order breaking change

### LINKED ISSUE #16000: Incorrect guidance on ConfigureServices execution order
In the section "Customize WebApplicationFactory", it says:

&gt; The SUT's database context is registered in its Startup.ConfigureServices method. The test app's builder.ConfigureServices callback is executed **after** the app's Startup.ConfigureServices code is executed.

This is incorrect. If you follow the guide with breakpoints, you can see that it's the other way around and the test app's `builder.ConfigureServices` is run **before** the real Startup class's method. It's a quite dangerous guidance as well since the "real" services could end up being used for tests rather than the test versions.

---
#### Document Details

* Content: [Integration tests in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-3.1#feedback)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #16000

Call out that the execution order was a breaking change for 3.0. There is no announcement to link for this, per se. All we would have to link is the Generic Host announcement, so we may as well just cross-link the Generic Host topic here instead.

Thanks @SaebAmini! 🎸

### DIFF
--- a/aspnetcore/test/integration-tests.md
+++ b/aspnetcore/test/integration-tests.md
@@ -5,7 +5,7 @@ description: Learn how integration tests ensure that an app's components functio
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 10/28/2019
+ms.date: 01/06/2019
 uid: test/integration-tests
 ---
 # Integration tests in ASP.NET Core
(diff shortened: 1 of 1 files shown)


guardrex:
We could get into the weeds by cross-linking [Generic Host restricts Startup constructor injection #353](https://github.com/aspnet/Announcements/issues/353), but that's kind'a nasty ... it doesn't really directly say what's going on with the testing scenario. Barring further feedback/confusion, we can just say that the breaking change has to do with the changeover to the Generic Host and move on. I'll keep an :ear: open for more remarks on this subject.

PR #16440, opened 2020-01-07
change component devlang to cshtml

scottaddie:
The Component Tag Helper will be used within a CSHTML file. Updating the devlang from `razor` to `cshtml`.


guardrex:
Perhaps, that was done on purpose ... not sure ... haven't seen that construct before.

PR #16442, opened 2020-01-07
fix missing header in pre-3.0 version of signalr tutorial

anurse:
Noticed this while doing a check over some SignalR docs :). If you go to https://docs.microsoft.com/en-us/aspnet/core/tutorials/signalr-typescript-webpack and toggle the version selector to a version below 3.0, the `Prerequisites` header just disappears :). It wasn't duplicated when the doc was duplicated across monikers.


Rick-Anderson:
@anurse  can I S&M and merge to live?

anurse:
Go for it

PR #16464, opened 2020-01-08
minor update to add instruction to add dotnet ef tools

sayedihashimi:
Adding some missing instructions for scaffolding with Visual Studio for Mac.


Rick-Anderson:
Thanks for the fix

PR #16511, opened 2020-01-12
Update StartupDevProd.cs

yhippa:
updated class name to work correctly with examples


scottaddie:
Thank you for bringing this to our attention. I'm going to close this PR in favor of https://github.com/aspnet/AspNetCore.Docs/pull/16532. This is part of a larger problem that I'm addressing over there. It's tricky because both the MVC and Razor Pages tutorials share include files.

PR #16520, opened 2020-01-13
aspnetcore update index migration link

khancyr:
In Migration from the .NET Framework the link to ASP core migration was broken


scottaddie:
Thank you for the PR, @khancyr! The change will go live within the next 24 hours.

PR #16521, opened 2020-01-13
Removed preview channel note

Reisclef:
There is no need to switch to the preview channel on Mac as VS for Mac uses 8.4 as of 8th Jan 2020.


scottaddie:
Thank you for fixing this, @Reisclef! Your change will go live within the next 24 hours.

PR #16623, opened 2020-01-20
C150960: Adding missing backticks at end of code

v-maudel:
Description of the source issue: Lack of a backticks at the end of code-block is preventing following content to be localized


scottaddie:
Thank you for the PR, @v-maudel!

v-maudel:
Thanks to you @scottaddie

PR #16640, opened 2020-01-21
Update links to MessagePack-CSharp

AArnott:
One of the anchor links was out of date. But really both links needed updating because they point people at the 2.1 documentation when for SignalR only the 1.x version is relevant. So I've updated the links to explicitly point at the v1.8 branch of the docs which fixes the broken link and points people at the docs relevant to the version they're using.


Rick-Anderson:
@AArnott  thanks!

PR #16658, opened 2020-01-22
L166242: removing extra comma

v-maudel:
Description: There is an extra comma which is causing the JSON file to have an invalid format


v-maudel:
HI @tdykstra, can you please approve and merge this PR?
Thanks in advance.

v-maudel:
Thanks @guardrex

PR #16668, opened 2020-01-22
Update BlazorHandleErrors

### LINKED ISSUE #16544: Clarify for Server vs. WebAssembly
I was thinking, as I'm reading through all the docs here :)...
and it would be so much easier to to read it if you could just have separate documentation for Server and WebAssembly.
I know there are often remarks to the side about it, but there is no clear way just to filter out info about Server, if I'm only interested in WebAssembly.

Like in this article we start with "Blazor Server is a stateful framework" - does that mean these is nothing here about WebAssembly error handling? I have to search for "WebAssembly" or scroll thought the whole thing to get an idea.

So my take on how to solve it:
- a tab at the top of the page to select one or the other, and when selected you only see relevant information - a different version of the article.

---
#### Document Details

* ID: ce1440e7-febc-3fbc-09d7-7f47b740fd3b
* Version Independent ID: 31071595-7c71-b6f7-3856-74f78c5c7cdf
* Content: [Handle errors in ASP.NET Core Blazor apps](https://docs.microsoft.com/en-us/aspnet/core/blazor/handle-errors?view=aspnetcore-3.0&source=docs#feedback)
* Content Source: [aspnetcore/blazor/handle-errors.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/handle-errors.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #16544

WRT the strange PR title: Because the new *What's New* content is automated to capture good PRs for readers to find out about, titles of PRs that we don't want to appear should use a `Update {ONE WORD}` PR title. This PR falls into that bucket ... just clarifying the bits of the topic that apply to Blazor Server.

Thanks @RedDeathGitHub! ⛄️

### DIFF
--- a/aspnetcore/blazor/handle-errors.md
+++ b/aspnetcore/blazor/handle-errors.md
@@ -5,7 +5,7 @@ description: Discover how ASP.NET Core Blazor how Blazor manages unhandled excep
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 12/18/2019
+ms.date: 01/22/2020
 no-loc: [Blazor, SignalR]
 uid: blazor/handle-errors
 ---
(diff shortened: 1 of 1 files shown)


RedDeathGitHub:
👍

PR #16750, opened 2020-01-28
Merge to live 28 Jan

Rick-Anderson:


guardrex:
@Rick-Anderson Idk what's what with that sync problem. We should take that version upgrade live if possible.

PR #16783, opened 2020-01-30
Update WhatsNew2019-09

guardrex:
Should we do this when there's a bad link/redirect? I don't like seeing them in my local DocFX build, and this will solve that problem if that's ok.


scottaddie:
I'm okay with this approach.

guardrex:
Redirects aren't a huge deal, but I generally run with the 'why waste bandwidth' cats. In this case, DocFX gives me a nasty hate message about it.

PR #16799, opened 2020-01-31
SignalR Docs update in response to CVE-2020-5234 in MessagePack library

anurse:
In response to [CVE-2020-5234](https://github.com/neuecc/MessagePack-CSharp/security/advisories/GHSA-7q36-4xx7-xcxf), we are updating our docs to recommend setting a new static property in their applications when using MessagePack.

cc @AArnott 

`@Rick-Anderson  edit:` [Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/signalr/messagepackhubprotocol?view=aspnetcore-2.1&branch=pr-en-us-16799#configure-messagepack-on-the-server)

### DIFF
--- a/aspnetcore/signalr/messagepackhubprotocol.md
+++ b/aspnetcore/signalr/messagepackhubprotocol.md
@@ -45,7 +45,7 @@ services.AddSignalR()
 ```
 
 > [!WARNING]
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
> But do you want to add any suggestion that folks consider other changes that may be necessary to secure their application by reviewing the advisory itself?

@AArnott  I'll add that unless I get objections.

Rick-Anderson:
@anurse  can you take a quick look at my latest commit?

Rick-Anderson:
@anurse  @BrennanConroy  does it have to be `1.9.x`? What about `2.1.x`

BrennanConroy:
2.1.x isn't applicable because our library references 1.9.x and is not compatible with 2.1.x.

PR #16820, opened 2020-02-02
Replace double articles

RobJohnston:
This replaces all occurrences of "the the" or "a a".


RobJohnston:
Sure thing @guardrex, but I should say that I already checked and it only exists in one place there and was fixed in https://github.com/jquery-validation/jquery-validation/commit/1e2a9ea682d935d370edcc0cba92166e5ba40d79#diff-8a0f59179e7555ee15605d1abd45cdec

Issue #16553, opened 2020-01-14
It's not yet clear with which old browser versions (even past years) we could encounter incompatibilities.


idchlife:
I know this was discussed already, but I can't see clear picture with browser versions support, which makes using (or starting using) Blazor kinda risky.

How would we know that Blazor server would work for example if some people use like kinda outdated version of Firefix/Chrome due to them sitting in Windows 7 and not knowing how to update their browser? (happened several time (not with Blazor but overall in some projects), some people don't event understand that they should update their browser, and sometimes their corporate laptops are just the way they are)

I mean it would be very helpful info like "Blazor Server: Firefox 67+ (previous need polyfills, add them when needed)"

Maybe even basic info about blazor server, like "core blazor is transpiled/not transpiled so your browsers should have at least es6 support" or something like this.

Reference to pull request with discussion.

https://github.com/aspnet/AspNetCore.Docs/pull/14842

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: c7957ff6-5b09-31dd-17dd-2bf3660a8ebd
* Version Independent ID: b03c3cec-a3de-d563-d0f8-45fabfb39758
* Content: [ASP.NET Core Blazor supported platforms](https://docs.microsoft.com/en-us/aspnet/core/blazor/supported-platforms?view=aspnetcore-3.1)
* Content Source: [aspnetcore/blazor/supported-platforms.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/supported-platforms.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Hello @idchlife ... Management usually triages these on Fridays, so stand-by for them to take a followup look at this issue, either this Friday or the next.

mkArtakMSFT:
This will require some discussions on our end to come up with a more specific browser compatibility version list.

guardrex:
Let's close this as not workable.

One of the issues is that SignalR only supports the latest browser. Since Blazor Server is based on SignalR, Blazor Server can't have a minimum version.

[...]

PR #16823, opened 2020-02-03
Updated the supported versions matrix

mkArtakMSFT:
Time passes by and the `Current` version changes too. We can't simply say "current" as the only supported version.
I've used relatively recent version numbers for the browsers

- [Google Chrome Version History](https://en.wikipedia.org/wiki/Google_Chrome_version_history)
- [Firefox Version History](https://en.wikipedia.org/wiki/Firefox_version_history)
- [Safari Version History](https://en.wikipedia.org/wiki/Safari_version_history)

Fixes https://github.com/aspnet/AspNetCore.Docs/issues/16553

### DIFF
--- a/aspnetcore/blazor/supported-platforms.md
+++ b/aspnetcore/blazor/supported-platforms.md
@@ -5,7 +5,7 @@ description: Learn about the supported platforms for ASP.NET Core Blazor.
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 12/18/2019
+ms.date: 02/03/2019
 no-loc: [Blazor, SignalR]
 uid: blazor/supported-platforms
 ---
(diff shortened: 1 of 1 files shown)


guardrex:
Cross-reference to the last time *someone* :smile: tried this. lol

https://github.com/aspnet/AspNetCore.Docs/pull/14842

guardrex:
Also, should be a fixes for https://github.com/aspnet/AspNetCore.Docs/issues/16553?

anurse:
When @danroth27 and I chatted previously we had landed on the fact that SignalR will officially support the "current" version and not make statements about specific minimum versions, and thus so should Blazor Server (at least). If you want to change that for Blazor though I'm not worried, but for SignalR we didn't want to increase the scale of our support matrix to deal with historical versions.

If we want to clarify things like which versions include the minimum features that Blazor/SignalR require, then perhaps it would be better to rephrase the doc. But this chart is about *supported* versions and we've generally tried to avoid making specific browser version statements when we talk about *supported* versions and instead strongly encourage the use of the latest version.

guardrex:
Given that :point_up: if you want the versions of the browsers at the time of 3.0 release, they're on that old PR :point_right: https://github.com/aspnet/AspNetCore.Docs/pull/14842/files. I used major version only and my MS Edge version might be wrong based on this PR (this PR says version "80"-something), but I took it from Edge (Settings) ...

Capture

I didn't address Blazor WebAssembly given that it's in preview. *Current* seemed to make that most sense until release.

> SignalR we didn't want to increase the scale of our support matrix to deal with historical versions.

> rephrase the doc

We can qualify this "support" by saying exactly that:

* This applies to the Blazor framework, not SignalR, which always only pertains to *current versions* of browsers.
* Don't say "supported version" ... say **_"suggested minimum version."_**

mkArtakMSFT:
I'm planning to come up with a proposal regarding how to phrase this. I had some conversations with @danroth27 and @anurse related to this. Will try to summarize that soon.

scottaddie:
@mkArtakMSFT Any updates to provide on the proposal?

guardrex:
We still have the issue, #16553, on `HOLD` on the Blazor project. With this PR cross-linked, we won't lose it. Let's take this down for now. The issue will jump to `Done`, but I'll reset it to `HOLD` in a sec.

PR #16854, opened 2020-02-04
Clarify log level and default for Windows Services

### LINKED ISSUE #16843: Event log
This page explains that the UseWindowsService() extension method enables logging to the event log. But, when I debug the app, I get zero entries in the event log. When I run the app as a service (using local system - so it has the necessary admin rights to create a new event source), nothing is logged to the event source either. So, what am I missing here?

---
#### Document Details

* Content: [Host ASP.NET Core in a Windows Service](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/windows-service?view=aspnetcore-3.1&tabs=visual-studio#feedback)
* Content Source: [aspnetcore/host-and-deploy/windows-service.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/host-and-deploy/windows-service.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #16843

CR, would have handled this one with only docs review, but need to check something with you:

From the issue comment (pertaining to the Event Logging added automatically for a Windows Service app) ...

> the default event log source config is to log only messages with warning or higher

... but we've documented the log level as being set by `Logging:LogLevel:Default`, which should be *Information* or higher via a template-based app.

Is the guidance on this PR correct, or is the level actually defaulting to *Warning* or higher? If *Warning* or higher, where's that coming from?

### DIFF
--- a/aspnetcore/host-and-deploy/windows-service.md
+++ b/aspnetcore/host-and-deploy/windows-service.md
@@ -47,7 +47,7 @@ The app requires a package reference for [Microsoft.Extensions.Hosting.WindowsSe
 * Enables logging to the event log:
   * The application name is used as the default source name.
   * The default log level is *Warning* or higher for an app based on an ASP.NET Core template that calls `CreateDefaultBuilder` to build the host.
-  * Override the default log level with the `Logging:LogLevel:Default` key in *appsettings.json*/*appsettings.{Environment}.json*.
+  * Override the default log level with the `Logging:LogLevel:Default` key in *appsettings.json*/*appsettings.{Environment}.json* or other configuration provider.
   * Only administrators can create new event sources. When an event source can't be created using the application name, a warning is logged to the *Application* source and event logs are disabled.
 
 In `CreateHostBuilder` of *Program.cs*:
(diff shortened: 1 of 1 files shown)


Tratcher:
> Is the guidance on this PR correct, or is the level actually defaulting to _Warning_ or higher? If _Warning_ or higher, where's that coming from?

I don't see anything that would restrict it to warning.

Tratcher:
Ah, I found it:
https://github.com/dotnet/extensions/blob/ef5d488c054c3aa850c29c10d851fcdc8a004eed/src/Hosting/Hosting/src/Host.cs#L97-L103

guardrex:
Ah ... I see. Ok. I'll need to fix this up.

guardrex:
> this lets the defaults be overridden by the configuration.

Is the key going to be correct tho? We currently say to set the default key ...

```
Logging:LogLevel:Default
```

guardrex:
The updates that I made assume that the key remains the same to override the default.

Tratcher:
Since it's a provider specific filter it might require a provider specific config to override.
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-3.1#log-filtering

guardrex:
I would prefer to document it that way. So give them a key of ❓ ...

```
Logging:Console:LogLevel:Microsoft.Extensions.Logging.EventLog.EventLogLoggerProvider
```

Tratcher:
Try: `Logging:EventLog:LogLevel:Default"?

guardrex:
Ok ... will do ...... tomorrow tho. I'm toast for the day. Will mark this on hold until I can confirm here. 

Thanks again for your help @Tratcher! 🍻

PR #16869, opened 2020-02-05
Update Disposal section and versioning in DI topic

### LINKED ISSUE #16773: Disposing of service instances not created by the container
As stated in the [Disposal of Services](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-3.1#disposal-of-services) section:

```csharp
    // The container doesn't create the following instances, so it doesn't dispose of
    // the instances automatically:
    services.AddSingleton<Service3>(new Service3());
    services.AddSingleton(new Service3());
```

Where is the correct place to dispose of service instances not created by the container such as these?

---
#### Document Details

* Content: [Dependency injection in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-3.1#feedback)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #16773

We're going to convert this topic to *whole topic versioning* at the same time that I address the *Disposal of services* section. Since the diff is hard to review, I'll place the entire section in this comment for review. The updated lingo is :point_down: ...

---

### Disposal of services

The container calls \<xref:System.IDisposable.Dispose*> for the \<xref:System.IDisposable> types it creates. If an instance is added to the container by user code, it isn't disposed automatically.

In the following example, the services are created by the service container and disposed automatically:

```csharp
public class Service1 : IDisposable {}
public class Service2 : IDisposable {}

public interface IService3 {}
public class Service3 : IService3, IDisposable {}

{
    services.AddScoped<Service1>();
    services.AddSingleton<Service2>();
    services.AddSingleton<IService3>(sp => new Service3());
}
```

In the following example:

* The service instances aren't created by the service container.
* The intended service lifetimes aren't known by the framework.
* The framework doesn't dispose of the services automatically.
* If the services aren't explicitly disposed in developer code, they persist until the app shuts down.

```csharp
public class Service1 : IDisposable {}
public class Service2 : IDisposable {}

{
    services.AddSingleton<Service1>(new Service1());
    services.AddSingleton(new Service2());
}
```

---

Thanks @antmdvs! :rocket:


guardrex:
Ping @anurse for review. It's a quick one, but don't look at the diff (we have whole-topic versioning in progress here, too). Check out my opening comment, where I show the new text for disposal. :point_up:

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -575,26 +415,38 @@ If a class seems to have too many injected dependencies, this is generally a sig
...
+In the following example:
 
-    // The container doesn't create the following instances, so it doesn't dispose of
-    // the instances automatically:
-    services.AddSingleton<Service3>(new Service3());
-    services.AddSingleton(new Service3());
+* The service instances aren't created by the service container.
+* The intended service lifetimes aren't known by the framework.
+* The framework doesn't dispose of the services automatically.
+* If the services aren't explicitly disposed in developer code, they persist until the app shuts down.
```
antmdvs: Is there technically any benefit to using the [Lock pattern ](https://github.com/aspnet/AspNetCore.Docs/issues/16773#issuecomment-580091374 )for explicit disposal on shutdown if possible or is the end result effectively the same without it?

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -692,3 +544,538 @@ DI is an *alternative* to static/global object access patterns. You may not be a
...
+      public void MyMethod()
+      {
+          var option = _optionsMonitor.CurrentValue.Option;
+
+          ...
+      }
+  }
+  ```
+
+* Another service locator variation to avoid is injecting a factory that resolves dependencies at runtime. Both of these practices mix [Inversion of Control](/dotnet/standard/modern-web-apps-azure-architecture/architectural-principles#dependency-inversion) strategies.
```
antmdvs: Do you specifically mean a factory that **uses a service locator** to resolve dependencies? Otherwise, this could be misconstrued with the factory form given in the example above and may seem contradictory to the guidance in the Disposal section:

```services.AddSingleton<IService3>(sp => new Service3());```

guardrex: I don't think it's a significant problem following what was just shown with "Another service locator variation ...", but this sentence could say ...

> Another service locator variation to avoid is injecting a factory that resolves dependencies at runtime using \<xref:System.IServiceProvider.GetService*>.

antmdvs: This does add clarity IMO. 👍

aspnetcore/fundamentals/dependency-injection.md
guardrex: I can go with any of (based on @anurse thoughts on that question) ...

* Cross-link the Lock blog.
* Just *say* to use `ApplicationStopping` ... let's call this option 👮 *Stealing Lock!*:tm: 👮‍♀. (cc: @andrewlock)
* Do nothing.

guardrex:
@mkArtakMSFT Can you look at this (just need to review the revised content that I pasted into the opening comment :point_up:) or assign someone available to look ... I don't want to randomly ping *busy* 🏃 folks.

There's also a quick ? on how far to go with the guidance on the explicit disposal at shutdown point ... we need to go one of three ways there. :point_up:

mkArtakMSFT:
This looks reasonable to me. @javiercn can you please have a look at this too? Thanks!

aspnetcore/fundamentals/dependency-injection.md
binki: I think saying “If the services aren't explicitly disposed in developer code, they persist until the app shuts down.” is confusing/misleading. Anything that is not `Dispose`()d prior to shutdown will not somehow automatically be `Dispose()`d during shutdown. Also, `Dispose()`ing something doesn’t cause it to cease to exist (i.e., things persist after they are `Dispose()`d).

For the sake of code correctness and discipline, it would be nice to know how to register a handler for when the container is shut down. If I manually create an instance, then I would like to be able to be informed that the container is being shutdown so I can know the container no longer needs my instance (and, if that was the last thing needing the instance, that would be the time to call `Dispose()` on it).

guardrex: I wouldn't phrase that concern in exactly that way. It doesn't say that the services are disposed at shutdown, it merely says that the services *persist until the app shuts down*, which isn't the same thing. It flows from the prior bullet that the framework does nothing on its own. It could say ...

> If the services aren't explicitly disposed in developer code, they persist until the app's process shuts down.

WRT ...

> how to register a handler for when the container is shut down

See https://github.com/aspnet/AspNetCore.Docs/issues/16773, including *Lock's Gambit*:tm:. On the notion that we'd cover something explicitly here (or cross-link Lock's blog post), note that Nurse or Nelson haven't called for that (*yet*). Let's see what they say on review. I'd like their guidance before making further changes.

binki: > I wouldn't phrase that concern in exactly that way. It doesn't say that the services are disposed at shutdown, it merely says that the services _persist until the app shuts down_, which isn't the same thing. It flows from the prior bullet that the framework does nothing on its own. It could say ...
> 
> > If the services aren't explicitly disposed in developer code, they persist until the app's process shuts down.

Adding “process” doesn’t change things. In fact, it makes it less exact—especially if the process lives longer than the server which could be shut down, allowing the instance to be garbage collected (an cease to “persist”—without necessarily ever being `Dispose()`d) prior to the process exiting.

Also, the other thing I don’t like about that is how it read in English. The language of “If x, then y” in English implies “If not x, then not y”. So you’re saying that `Dispose()`ing something causes it to cease to exist (to be fully unreferenced and garbage collected). I’d rather not have the language of the docs suggest that ;-).

guardrex:
@javiercn (or @anurse) ... a quick one to review. You only need to look at the content my opening comment :point_up: (don't look at the diff ... we're moving content around that obscures the critical content changes).

For services created by the dev (not the container), do you want to say something more detailed than ...

> If the services aren't explicitly disposed in developer code, they persist until the app shuts down.

... about exactly when/how services and their resources are released (garbage collected)?

Finally, do you want to show or cross-link the [pattern that Lock shows](https://github.com/dotnet/AspNetCore.Docs/issues/16773#issuecomment-580091374) in one of his blog posts?

guardrex:
@mkArtakMSFT Wait longer, or should we go ahead here?

There's still a bit of an open question on ...

> If the services aren't explicitly disposed in developer code, they persist until the app shuts down.

... about exactly when/how services and their resources are released (garbage collected)? It's not necessarily blocking tho because the remark isn't particularly specific. We can update the remark on a future PR when we're less busy. 🏃

guardrex:
@mkArtakMSFT Let's try and unblock this from merging because it's holding up full-topic versioning for the DI topic. If another author touches the topic before this merges, :boom:.

The only remaining questions are ...

1. For services created by the dev (not the container), do we want to say something more detailed than ...

   > If the services aren't explicitly disposed in developer code, they persist until the app shuts down.

   ... about exactly when/how services and their resources are released (garbage collected)?

1. Do we want to show or cross-link the [pattern that Lock shows](https://github.com/dotnet/AspNetCore.Docs/issues/16773#issuecomment-580091374) in one of his blog posts?

mkArtakMSFT:
@guardrex , for 1. I'm ok with that phrasing for now.
For 2. we can have some soft working around it not to sound like "This is the pattern you should follow", rather "This blog shows different ways you can dispose of objects".

guardrex:
@Rick-Anderson ok ... 😅 ... ready for whole topic versioning now on this one. Best to :eyes: the build before merging in case I 💥 something ...

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.1&branch=pr-en-us-16869)

antmdvs:
Appreciate your diligence on this, @guardrex!

PR #16908, opened 2020-02-07
Whole-topic versioning

### LINKED ISSUE #16907: Whole-topic Versioning

guardrex:
Fixes  #16907

Versioning for the remaining fifteen that I'm `ms.author` on. No warnings/errors on the build report.

Break out the *EYE DROPS!* 👁

### DIFF
--- a/aspnetcore/fundamentals/configuration/index.md
+++ b/aspnetcore/fundamentals/configuration/index.md
@@ -5,12 +5,14 @@ description: Learn how to use the Configuration API to configure an ASP.NET Core
 monikerRange: '>= aspnetcore-2.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 01/29/2020
+ms.date: 02/07/2020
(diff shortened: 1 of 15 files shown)


Rick-Anderson:
@wadepickett  can you review this builds? Get two browsers side by side. One browser the review code, the other on the live site. Compare the versions too.

wadepickett:
Thanks! Breaking out the eye drops...  :)

wadepickett:
I did a visual check plus a text diff for all 16 x every version.  All looks good. Congrats! I could not find an unintended difference.   The only differences I did see were intended and checked out, such as:
-  A bug fix in fundamental/servers for 
`options.AllowSynchronousIO = true;`
- A repeated line edited out in iis/index
- Sample download moved below the intro so the intro is first as it should be for caching/middlewhere.
- The providers section of performance/response-compression changed so the intro rather than the code is first as it should be.

PR #16957, opened 2020-02-11
NCache SignalR backplane for asp.net core

mohsinnasir:
Mentioning NCache as opensource thirdparty asp.net core signalR backplane
Added ncache-backplance image explaining the functionality , usage and benefits of ncache signalR

### DIFF
--- a/aspnetcore/signalr/scale.md
+++ b/aspnetcore/signalr/scale.md
@@ -90,14 +90,6 @@ The Azure SignalR Service advantages noted earlier are disadvantages for the Red
 * A SignalR app must scale out based on number of clients even if few messages are being sent.
 * A SignalR app uses significantly more connection resources than a web app without SignalR.
 
-## NCache backplane
-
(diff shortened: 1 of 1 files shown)


davidfowl:
I’m not sure this should be featured like this. There are a bunch of 3rd party backplanes that should fit as well. I’d put those in another section and list a bunch of them (orleans etc). We do this for logging and DI today

BrennanConroy:
We generally don't write articles about 3rd party libraries, we can include them in a list of other providers like we do for logging: https://docs.microsoft.com/aspnet/core/fundamentals/logging/?view=aspnetcore-3.1#third-party-logging-providers

Rick-Anderson:
@mohsinnasir can you update this PR with a list in a format similar to [Third-party logging providers](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-3.1#third-party-logging-providers) ?

mohsinnasir:
@Rick-Anderson I have updated the PR and incorporated the changes as you asked.

PR #16982, opened 2020-02-12
Replace generic List return type with IList

scottaddie:
Fixes https://github.com/aspnet/AspNetCore.Docs/issues/14761

Return `IList<T>` instead, per https://docs.microsoft.com/visualstudio/code-quality/ca1002?view=vs-2019.


rynowak:
Framework guidelines are for the framework.

scottaddie:
@rynowak Is this rule still a good idea for .NET Core?

rynowak:
Rules like this are not relevant for controllers.

This is guidance for creating a maintainable public API surface that will be used by other libraries, so that you can evolve it without binary breaking changes. None of these things are concerns with an MVC action method.

PR #16988, opened 2020-02-12
Update new-field.md

reidhaegele:
If you do not change the binding then editing the rating will not work.


Rick-Anderson:
@reidhaegele  thanks!

PR #16994, opened 2020-02-13
Add missing services for authorization

geersjake:
Without these two lines program throws:

> System.InvalidOperationException: Unable to resolve service for type 'Microsoft.Extensions.Options.IOptions`1[Microsoft.AspNetCore.Authorization.AuthorizationOptions]' while attempting to activate 'Microsoft.AspNetCore.Authorization.DefaultAuthorizationPolicyProvider'

### DIFF
--- a/aspnetcore/security/blazor/index.md
+++ b/aspnetcore/security/blazor/index.md
@@ -196,6 +196,8 @@ public class Program
     public static async Task Main(string[] args)
     {
         var builder = WebAssemblyHostBuilder.CreateDefault(args);
+        builder.Services.AddOptions();
+        builder.Services.AddAuthorizationCore();
         builder.Services.AddScoped<AuthenticationStateProvider, 
             CustomAuthStateProvider>();
         builder.RootComponents.Add<App>("app");


guardrex:
@geersjake Clarify ...

> program

Your program, or the sample code as shown? ... because I ran the code successfully without options and only had to add auth (`AddAuthentication`) because I was testing with social auth providers. If this sample code runs as-is without these additional services, then it is up the dev to configure services as-needed for their own app's requirements.

geersjake:
I created a fresh project using the WebAssembly template (ASP.NET Core hosted) and then added the various code snippets specified in this guide. It is a very bare bones implementation with nothing extra added. Both the `AddOptions` and `AddAuthorizationCore` were necessary in addition to the template + document code to get the project running. Just my experience though, perhaps I erred.

guardrex:
I retested in the base case, and that went well as described. I recalled running that scenario before, and recalled it going well.

As far as I can tell, I trusted the cascading code and didn't test it locally. I was probably 🏃😅 at the time and felt that it *should* be ok as provided. 🙈

This will probably become valid guidance, but it should be covered a little differently ... it should go in the text in the section that pertains to *auth state as a cascading parameter*.

I'm going to close this and open an issue to address it today. I'll ping you on the issue and the PR. Thanks for calling this out.

PR #17017, opened 2020-02-14
Update the first paragraph for clarity

rhires:
Make it more clear what the scaffolding does and is for, and why it's necessary. Also use a bit more plain language.


Rick-Anderson:
@rhires  thanks for your PR but I prefer the original.

PR #17021, opened 2020-02-15
Add a note about including the DependencyInjection Extension

irlcatgirl:
This merge request adds a note to users about adding the DependencyInjection Extension to program.cs. I'm not sure exactly how to word this but open to suggestions on it.


scottaddie:
Thank you for the PR, @irlcatgirl. The `using` statement is already provided for you in the `Program` class code. You simply need to copy and paste that code into your text editor. It can replace what already exists in the *Program.cs* file.

PR #17045, opened 2020-02-18
Update global application example of AutoValidateAntiForgeryTokenAttribute

martinmladenov:
I replaced `services.AddMvc` with `services.AddControllersWithViews` as the way to apply filters globally was changed with the release of ASP.NET Core 3


Rick-Anderson:
@martinmladenov  thanks for this. Check my commit. We need to maintain the 2.1 version.

martinmladenov:
@Rick-Anderson Thank you. Keeping the 2.1 example is certainly a good idea.

PR #17049, opened 2020-02-19
Update samesite.md

### LINKED ISSUE #17033: Test with Chrome: add instructions for disabling temporary workaround?
The article correctly explains that Chrome 78+ applies a mitigation workaround, *even* if the two flags `#same-site-by-default-cookies` and `#cookies-without-same-site-must-be-secure` are force-enabled. This, ironically, makes Chrome 80 more useless than Chrome 76/77 for testing whether e.g. an OpenID authentication flow breaks under the new rules that would theoretically be enforced starting with Chrome 80.

There is, however, a **hidden** Chrome flag that can "really, really, truly" enforce the new behaviour and prevent the workaround from being applied. The flag is documented under the heading *Nov 21, 2019* in [this Chromium blog post](https://www.chromium.org/updates/same-site). The flag is not shown in `chrome://flags`, it has to be set using the command-line argument `--enable-features=SameSiteDefaultChecksMethodRigorously`. Also, when set, there seems to be no way to tell whether it's actually active within the Chrome session. It does, however, finally break our authentication flow and expose our cookie policy problem, as expected, and identical to the behaviour in Firefox.

This might be a worthwhile addition to the Chrome section of the article, as it should be a simpler method than installing an older version of Chrome. Because of the ramshackle way Google has implemented this option, it sadly always has to come with the warning that there is no guarantee the flag has actually been correctly activated. I don't know if that is enough of a problem to speak against mentioning it in the article – I can just say that for me personally, it would have saved me a lot of time and confusion if I had known it beforehand.

The worst part is that Google/Chromium don't proactively warn that the mitigation is being triggered, and that the two visible flags are **no longer** sufficient for testing the new behaviour. There is still a lot of writing from the pre-Chrome-78 days floating around the web which implies that they are sufficient. I was wondering why Firefox 72 with `network.cookie.sameSite.laxByDefault` and `network.cookie.sameSite.noneRequiresSecure` broke our authentication flow while Chrome 80 with the two above-mentioned flags still worked, until I finally figured out that this hard-coded workaround was being triggered. Hence my suggestion that the ASP.NET Core article on the topic should maybe at least refer to this flag.

---
#### Document Details

* ID: 1945dedd-e2b8-37c1-2676-e473498bcb30
* Version Independent ID: 631281cb-65db-3196-1a6a-da9b3f2e3fed
* Content: [Work with SameSite cookies in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/samesite?view=aspnetcore-3.1#feedback)
* Content Source: [aspnetcore/security/samesite.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/security/samesite.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

Rick-Anderson:
Fixes #17033

### DIFF
--- a/aspnetcore/security/samesite.md
+++ b/aspnetcore/security/samesite.md
@@ -138,6 +138,8 @@ Google does not make older chrome versions available. Follow the instructions at
 * [Chromium 76 Win64](https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Win_x64/664998/)
 * [Chromium 74 Win64](https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Win_x64/638880/)
 
+Starting in Canary version `80.0.3975.0`, the Lax+POST temporary mitigation can be disabled for testing purposes using the new flag `--enable-features=SameSiteDefaultChecksMethodRigorously` to allow testing of sites and services in the eventual end state of the feature where the mitigation has been removed. For more information, see The Chromium Projects [SameSite Updates](https://www.chromium.org/updates/same-site)
+
 ### Test with Safari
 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@blowdart  do we want to add
Starting in Canary version `80.0.3975.0`, the Lax+POST temporary mitigation can be disabled for testing purposes using the new flag `--enable-features=SameSiteDefaultChecksMethodRigorously` to allow testing of sites and services

blowdart:
That'd be great

PR #17064, opened 2020-02-19
Expand `OwningComponentBase` docs

### LINKED ISSUE #14316: Need clarification on OwningComponentBase usage
There's a section on `OwningComponentBase<T>` but I'm a little confused about it. It states that there is a `ScopedServices` property of type `IServiceProvider` that is scoped to the component's creation. However, it also has a generic type parameter, and based on the example provided, I assume also a `Service` property of type `T`, which presumably is automatically injected? Is that correct?

Does `OwningComponentBase` only support a single service type to be injected or will it also let you inject multiple service types in addition to the `Service` property which is presumably automatically injected (is it automatically injected?)?

Is there a non-generic version of `OwningComponentBase`?

Is there a way to create scopes that last less than the lifetime of the component? For example, let's say I have a blazor server side app, and I want to have a `Save` method in my component that saves a record but stays on the page. Each time the `Save` method is executed I want a new instance of my `DbContext` to be created and injected (roughly equivalent to calling a REST service, where the server-side `DbContext` is scoped to that request). How is that accomplished?

Edit: I found some answers to the above questions (though I'll leave this issue open as a suggestion for clarification in the docs):

So, first thing I found out is that, contrary to what appears to be implied by that last section on `OwningComponentBase`, services injected into the component appear to already be scoped to that component's lifetime automatically. So if you have a transient service injected into the component, navigate away and back to the component, the service will be constructed each time the component is instantiated (though not if the URL changes but is mapped to the same component, as expected).

If you add `@inherits OwningComponentBase<TServiceType>` and use the `Service` property, it will be scoped the same way as if you injected it normally. Not sure what the purpose is, in this case. Additionally, contrary to the documentation, there is no `ScopedServices` property on the generic version of `OwningComponentBase`.

There is a non-generic version of `OwningComponentBase` which *does* have a `ScopedServices` property. This property can be used to inject services on demand, and when those services are transient they are constructed at the time they are requested - exactly what I wanted when injecting transient `DbContext` instances or transient services that inject transient `DbContext` instances.

Example:
Given this service definition (which should probably implement `IDisposable`):
```CSharp
    public interface IMyService
    {
        Task<bool> SaveUsingDbContext(string value);
    }

    public class MyService : IMyService
    {
        public MyService() => Console.WriteLine("MyService constructed!");
        public Task<bool> SaveUsingDbContext(string value)
        {
            Console.WriteLine($"TODO: Save value '{value}' to database...");
            return Task.FromResult(true);
        }
    }
```

And this component:

```
@page "/counter"
@inherits OwningComponentBase

<h1>Counter</h1>

<input @bind-value="Value" />

<button class="btn btn-primary" @onclick="Save">Save</button>

@code {
    string Value { get; set; } = "";

    /// <Summary>
    /// The value is saved using async IMyService injected on demand.
    /// Since IMyService is a transient service, it (and any services
    /// it depends on, such as a DbContext) will be created within this
    /// method.
    /// </Summary>
    void Save()
    {
        var myService = ScopedServices.GetService(typeof(IMyService)) as IMyService;
        myService.SaveUsingDbContext(Value);
    }
}

```

This is the result when you click the "Save" button:
image

---
#### Document Details

* ID: ce9457c7-13d4-1520-790c-a18187532ff4
* Version Independent ID: 43f87dbb-2f36-c91e-084c-c28d93da923f
* Content: [ASP.NET Core Blazor dependency injection](https://docs.microsoft.com/en-us/aspnet/core/blazor/dependency-injection?view=aspnetcore-3.0&viewFallbackFrom=aspnetcore-2.2#feedback)
* Content Source: [aspnetcore/blazor/dependency-injection.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/dependency-injection.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

mjrousos:
This beefs up `OwningComponentBase` docs in *blazor/dependency-injection*, explaining the uses of `OwningComponentBase` and `OwningComponentBase<T>`, as well as commenting on the specific case of `DbContext` usage, which is a common case for DI woes in Blazor.

Fixes #14316

### DIFF
--- a/aspnetcore/blazor/dependency-injection.md
+++ b/aspnetcore/blazor/dependency-injection.md
@@ -199,11 +199,12 @@ There are two versions of the `OwningComponentBase` type:
 
     ```razor
     @page "/preferences"
+    @using Microsoft.Extensions.DependencyInjection
     @inherits OwningComponentBase
 
     <h1>User (@UserService.Name)</h1>
     <ul>
-        @foreach (var setting in SettingService)
+        @foreach (var setting in SettingService.GetSettings())
         {
             <li>@setting.SettingName: @setting.SettingValue</li>
         }
@@ -211,9 +212,8 @@ There are two versions of the `OwningComponentBase` type:
 
     @code {
         private IUserService UserService { get; set; }
-
-        private IPlayerService SettingService { get; set; }
-
+        private ISettingService SettingService { get; set; }
+        
         protected override void OnInitialized()
         {
             UserService = ScopedServices.GetRequiredService<IUserService>();
--- a/aspnetcore/blazor/dependency-injection.md
+++ b/aspnetcore/blazor/dependency-injection.md
@@ -219,7 +219,7 @@ Two versions of the `OwningComponentBase` type are available:
   @code {
       private IUserService UserService { get; set; }
       private ISettingService SettingService { get; set; }
-        
+
       protected override void OnInitialized()
       {
           UserService = ScopedServices.GetRequiredService<IUserService>();
@@ -233,7 +233,7 @@ Two versions of the `OwningComponentBase` type are available:
   ```razor
   @page "/users"
   @attribute [Authorize]
-  @inherits OwningComponentBase<Data.ApplicationDbContext>
+  @inherits OwningComponentBase<AppDbContext>
 
   <h1>Users (@Service.Users.Count())</h1>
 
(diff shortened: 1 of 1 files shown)


guardrex:
Cool @mjrousos ... I'll take a look tomorrow (Thursday).

guardrex:
@mjrousos I'm ready to take a look. Do you need more time with it?

When I go over it, it's faster if I make the nit edits directly, only focusing on grammar+style matters. I'll submit regular feedback comments for anything else.

mjrousos:
@guardrex It's ready for review. Thanks!

guardrex:
@mjrousos Ok ... the key nits were ...

* Remove future tense.
* Shorten sentences for better machine translation, including making use of lists, simplifying sentence structure, and dropping some words and phrases that aren't required.
* Dropped some pronouns here and there because they're less specific than using a solid noun.
* A few comma nits updated.
* If you think it's ok, it's best (*for me* lol) to not have an unpaired 3rd level heading. I *think* 🤔 that last section can be a 2nd level section here.

Check the wording updates to confirm all is well here. If you have any questions about why I nitted something, let me know. If I modified or removed an important word/phrase, I'll revert/update it when I hear back from you.

guardrex:
I'm concerned that there isn't enough here to get the implementation right ...

* Create the `DbContext` directly using `DbContextOptions<TContext>` as an argument, which can be retrieved from DI and is thread safe. 
* Register the `DbContext` in the service container with a transient lifetime:
  * When registering the context, use `ServiceLifetime.Transient`.
  * Request a new instance for each logical EF operation.

Can we provide examples here to tide us over until I write up the EF Core with Blazor topic?

### First bullet example

Something like? .......

```razor
@page "/example"
@inject DbContextOptions<AppDbContext> DbContextOptions

<ul>
    @foreach (var item in _data)
    {
        <li>@item</li>
    }
</ul>

<button @onclick="LoadData">Load Data</button>

@code {
    private List<string> _data;

    private async Task LoadData()
    {
        _data = await GetAsync();
        StateHasChanged();
    }

    public async Task<List<string>> GetAsync()
    {
        using (var context = new AppDbContext(DbContextOptions))
        {
            return await context.StringData.ToListAsync();
        }
    }
}
```

### Second bullet example

```csharp
services.AddDbContext<AppDbContext>(options =>
    options.UseSqlServer("{CONNECTION STRING}"),
        contextLifetime:ServiceLifetime.Transient,
        optionsLifetime:ServiceLifetime.Transient);
```

mjrousos:
Yep, I'll add some samples. I didn't want to make that part *too* long since there's other 'EF Core on Blazor' stuff coming later, but I few more samples for the short-term. I'll take a look at your updates and add more EF samples now.

mjrousos:
I've updated the EF samples to be more detailed. 

@bricelam, can you please take a look at the last part of this updated doc and make sure the EF recommendations are correct? @guardrex will be adding a more in-depth "Blazor + EF" doc in the future but, for now, I've included some quick tips on avoiding issues with DbContexts in the Blazor DI docs since this topic comes up a lot with customer.

guardrex:
What I *think* we can ultimately do is make the EF Core section here the basis of the new topic. You'll be the author, but I can take the hits 👊 (i.e., the pings).

I took a peek at [Javier's example](https://github.com/dotnet/aspnetcore/issues/16745#issuecomment-549505605).  He goes the `OwningComponentBase<T>` route client-side in a Blazor Hosted app. I'm not keen on adding *another sample* 😅 to the repo. I'll try to figure out a way to add it as version-agnostic snapshots. *I'll try!*

Let's proceed here with this. The EF Core+Blazor topic will compose smoothly if we manage this process in two steps. I'll take care of the next step after this goes in, and I'll ping you for review.

bricelam:
I'll defer to @ajcvickers

mjrousos:
Just checking in here. @guardrex, is anything else needed before this one gets merged?

guardrex:
@mjrousos Delayed while the team works on Blazor WebAssembly framework tasks. No ETA yet, but they are aware of the PRs awaiting final review.

guardrex:
Let's go ahead with this one. I think we've had enough :eyes: on it for now. We can take another look if the community reacts and when we get to a formal EF Core+Blazor doc.

PR #17079, opened 2020-02-21
Add responsive support to Blazor samples

kiyoaki:


guardrex:
Thanks @kiyoaki, but we're going to leave these matching the Blazor templates. If you want to float this suggestion to the engineering team for the templates, open an issue with them here :point_right: https://github.com/dotnet/aspnetcore/issues.

PR #17108, opened 2020-02-24
Improve Blazor binding guidance

### LINKED ISSUE #17040: Surface Blazor @bind/@bind:event binding syntax
To change the binding event this page shows:

     @bind-value:event="oninput"

but it doesn't work. From the talk of Steve at https://www.youtube.com/watch?v=Khn7sDUSEJM I've realized that it should be corrected into:

     @bind:event="oninput"

---
#### Document Details

* ID: 8b1b6ad8-13ba-0195-8829-ade3b1b0b9cb
* Version Independent ID: 6609ea68-a918-f3b9-8c9c-004a4436e350
* Content: [ASP.NET Core Blazor data binding](https://docs.microsoft.com/en-us/aspnet/core/blazor/data-binding?view=aspnetcore-3.1#feedback)
* Content Source: [aspnetcore/blazor/data-binding.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/data-binding.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #17040

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/data-binding?view=aspnetcore-3.1&branch=pr-en-us-17108)

Thanks @curia-damiano! 🎷

### DIFF
--- a/aspnetcore/blazor/data-binding.md
+++ b/aspnetcore/blazor/data-binding.md
@@ -41,7 +41,7 @@ Using `@bind` with the `CurrentValue` property (`<input @bind="CurrentValue" />`
 
 When the component is rendered, the `value` of the input element comes from the `CurrentValue` property. When the user types in the text box and changes element focus, the `onchange` event is fired and the `CurrentValue` property is set to the changed value. In reality, the code generation is more complex because `@bind` handles cases where type conversions are performed. In principle, `@bind` associates the current value of an expression with a `value` attribute and handles changes using the registered handler.
 
-In addition to handling `onchange` events with `@bind` syntax, a property or field can be bound for other events by also including an `@bind:event` attribute with an `event` parameter. The following example binds the `CurrentValue` property for the `oninput` event:
+Bind a property or field on other events by also including an `@bind:event` attribute with an `event` parameter. The following example binds the `CurrentValue` property on the `oninput` event:
 
 ```razor
 <input @bind="CurrentValue" @bind:event="oninput" />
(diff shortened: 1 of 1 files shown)


guardrex:
@scottaddie Let's do this one. These are fairly straightforward updates to improve the binding content. Everything has been tested locally.

aspnetcore/blazor/data-binding.md
```diff
@@ -41,10 +41,10 @@ Using `@bind` with the `CurrentValue` property (`<input @bind="CurrentValue" />`
 
 When the component is rendered, the `value` of the input element comes from the `CurrentValue` property. When the user types in the text box and changes element focus, the `onchange` event is fired and the `CurrentValue` property is set to the changed value. In reality, the code generation is more complex because `@bind` handles cases where type conversions are performed. In principle, `@bind` associates the current value of an expression with a `value` attribute and handles changes using the registered handler.
 
-In addition to handling `onchange` events with `@bind` syntax, a property or field can be bound using other events by specifying an [`@bind-value`](xref:mvc/views/razor#bind) attribute with an `event` parameter ([`@bind-value:event`](xref:mvc/views/razor#bind)). The following example binds the `CurrentValue` property for the `oninput` event:
+In addition to handling `onchange` events with `@bind` syntax, a property or field can be bound for other events by also including an `@bind:event` attribute with an `event` parameter. The following example binds the `CurrentValue` property for the `oninput` event:
```
mkArtakMSFT: not sure whether this should be `for` or `on`?
```suggestion
In addition to handling `onchange` events with `@bind` syntax, a property or field can be bound on other events by also including an `@bind:event` attribute with an `event` parameter. The following example binds the `CurrentValue` property on the `oninput` event:
```

aspnetcore/blazor/data-binding.md
```diff
@@ -78,9 +91,9 @@ Consider the following scenario:
...
+By default, binding applies to the element's `onchange` event (`@bind="{PROPERTY OR FIELD}"`). Use `@bind="{PROPERTY OR FIELD}" @bind:event={EVENT}` to set a different event. For the `oninput` event (`@bind:event="oninput"`), the reversion occurs after any keystroke that introduces an unparsable value. When targeting the `oninput` event with an `int`-bound type, a user is prevented from typing a `.` character. A `.` character is immediately removed, so the user receives immediate feedback that only whole numbers are permitted. There are scenarios where reverting the value on the `oninput` event isn't ideal, such as when the user should be allowed to clear an unparsable `<input>` value. Alternatives include:
```
mkArtakMSFT:
```suggestion
By default, binding applies to the element's `onchange` event (`@bind="{PROPERTY OR FIELD}"`). Use `@bind="{PROPERTY OR FIELD}" @bind:event={EVENT}` to trigger binding on a different event. For the `oninput` event (`@bind:event="oninput"`), the reversion occurs after any keystroke that introduces an unparsable value. When targeting the `oninput` event with an `int`-bound type, a user is prevented from typing a `.` character. A `.` character is immediately removed, so the user receives immediate feedback that only whole numbers are permitted. There are scenarios where reverting the value on the `oninput` event isn't ideal, such as when the user should be allowed to clear an unparsable `<input>` value. Alternatives include:
```

aspnetcore/blazor/data-binding.md
guardrex: I went with "for" here for a couple of reasons, but I don't think either is a bad choice. One was to imply **_purpose_**, which "for" does well. "On" seems like a good choice, too, because it implies location/time. Also in the 2nd case, we end up with "on" "onevent," which sounded a little repetitive.

Let's try another version of this that might be a further improvement of the first sentence. I'll try *imperative tone* (an order to the reader) on the next commit. I'm cool with "on" for the second sentence. I like that better this morning! :smile:

Issue #15412, opened 2019-10-31
Code sample references nonexistent property ApplyCurrentCultureToResponseHeaders


grovesNL:
The following code sample appears on this page:

```csharp
app.UseRequestLocalization(new RequestLocalizationOptions
{
    ApplyCurrentCultureToResponseHeaders = true
});
```

But the `ApplyCurrentCultureToResponseHeaders` property described doesn't seem to exist on the `RequestLocalizationOptions` class, so this code sample fails to build with ASP.NET Core 3.0. The documentation for [`RequestLocalizationOptions`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.requestlocalizationoptions?view=aspnetcore-3.0) also doesn't reference mention this property.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: d1833676-ff1b-c587-cc9c-81afffa8ef68
* Version Independent ID: ad989395-3f49-4d55-521e-cf6fe56281ee
* Content: [Globalization and localization in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/localization?view=aspnetcore-3.0#feedback)
* Content Source: [aspnetcore/fundamentals/localization.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/fundamentals/localization.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-fundamentals**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
@hishamco please review

hishamco:
> But the ApplyCurrentCultureToResponseHeaders property described doesn't seem to exist on the RequestLocalizationOptions class

https://github.com/aspnet/AspNetCore/blob/master/src/Middleware/Localization/src/RequestLocalizationOptions.cs#L90

Rick-Anderson:
@grovesNL  you're probably missing a package. See https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-3.0&tabs=visual-studio#add-package-references-for-removed-assemblies

grovesNL:
Thanks for the quick response and link to the property on `master`. I think this problem still exists in the 3.0 release though.

This happens when I create a new project with `dotnet new web -f netcoreapp3.0` and make that single change in Startup.cs.

[...]

PR #17156, opened 2020-02-28
Only show property when it exists

ryanbrandenburg:
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/15412

The property was created in 5.0. so let's only show it there.

CC @Rick-Anderson

### DIFF
--- a/aspnetcore/fundamentals/localization.md
+++ b/aspnetcore/fundamentals/localization.md
@@ -272,7 +272,7 @@ The [Accept-Language header](https://www.w3.org/International/questions/qa-accep
 
 6. Tap the language, then tap **Move Up**.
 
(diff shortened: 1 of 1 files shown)


ryanbrandenburg:
Seems this is failing because there are no monikers which match >3.1, because there's no 5.0 moniker. @Rick-Anderson, thoughts?

Rick-Anderson:
@ryanbrandenburg  that's correct. We should probably comment it out until we have a 5.0 pre-release.

cc @scottaddie

scottaddie:
The build is fixed now. I added the `aspnetcore-5.0` moniker for us to begin using. For now, that version won't display in the docs version picker.

PR #17161, opened 2020-02-29
Add Blazor label to Blazor security topic issues

guardrex:
I started the Blazor security topic work this morning and remembered that the "Blazor" label isn't applied to the ones that open from that neck of the woods. Is it ok to add that label to issues that come from those topics? We're about to have a whole slew of new reference topics there, which I think will generate a number of new issues in the future.

**Additional**: ~I added the upcoming Blazor Components TH topic to this in advance of that topic being set up. It will happen fairly soon-ish. That ok? If not, I can add it later when that topic goes up. :ear:~ **HELD** until that topic is created.

### DIFF
--- a/.ghal.rules.json
+++ b/.ghal.rules.json
@@ -36,7 +36,6 @@
           "(?i).*master\/aspnetcore\/tutorials\/grpc*.*": {
             "labels-add": "gRPC"
           },
-          
           "(?i).*master\/aspnetcore\/tutorials\/signalr*.*": {
             "labels-add": "SignalR"
           }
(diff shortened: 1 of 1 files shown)


.ghal.rules.json
```diff
@@ -18,6 +18,12 @@
           "(?i).*master\/aspnetcore\/host-and-deploy\/blazor.*": {
             "labels-add": "Blazor"
           },
+          "(?i).*master\/aspnetcore\/mvc\/views\/tag-helpers\/built-in\/component.md": {
```
scottaddie: Let's hold off on the TH entry until the topic goes up. The other entry you added looks good.

guardrex:
My issue is marked to take care of the upcoming TH topic entry when the time comes.

PR #17172, opened 2020-03-01
Use FindAsync in Overposting sample.

### LINKED ISSUE #17171: Use SingleOrDefaultAsync() Method?
Nice tutorial, however I've spotted a minor issue: in the GetTodoItem(long id) action of the TodoItemsController (when updating this to use the DTO), the LINQ query is using the .SingleAsync method. This will throw an exception if a request asks for an id which does not exist and returns a server 500 code (the code checking for null after the query is never reached). I think the query should be using the SingleOrDefaultAsync method instead, then the null-checking code will work and return either the item or 404 code?

---
#### Document Details

* ID: 583ae3e7-5915-f22b-e54c-6c100cf2a7da
* Version Independent ID: 95f2549e-29f9-e31a-b4d0-432b971643ff
* Content: [Tutorial: Create a web API with ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/tutorials/first-web-api?view=aspnetcore-3.1&tabs=visual-studio#feedback)
* Content Source: [aspnetcore/tutorials/first-web-api.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/tutorials/first-web-api.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

serpent5:
Fixes #17171.

Looking at this closer, I think `FindAsync` is a better option for both simplicity and consistency.

### DIFF
--- a/aspnetcore/tutorials/first-web-api/samples/3.0/TodoApiDTO/Controllers/TodoItemsController.cs
+++ b/aspnetcore/tutorials/first-web-api/samples/3.0/TodoApiDTO/Controllers/TodoItemsController.cs
@@ -31,17 +31,14 @@ namespace TodoApi.Controllers
         [HttpGet("{id}")]
         public async Task<ActionResult<TodoItemDTO>> GetTodoItem(long id)
         {
-            var todoItemDTO = await _context.TodoItems
-                .Where(x => x.Id == id)
-                .Select(x => ItemToDTO(x))
-                .SingleAsync();
+            var todoItem = await _context.TodoItems.FindAsync(id);
 
-            if (todoItemDTO == null)
+            if (todoItem == null)
             {
                 return NotFound();
             }
 
-            return todoItemDTO;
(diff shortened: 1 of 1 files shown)


scottaddie:
Thank you for another awesome PR!

Rick-Anderson:
From the [EF folks](https://docs.microsoft.com/en-us/aspnet/core/data/ef-rp/crud?view=aspnetcore-3.1#update-the-edit-page)

When you don't have to include related data, `FindAsync `is more efficient.

serpent5:
> When you don't have to include related data, `FindAsync `is more efficient.

Oh nice, thanks. I didn't see that explicit comment, but I came to the same conclusion based on the explanations [a bit higher up](https://docs.microsoft.com/en-us/aspnet/core/data/ef-rp/crud?view=aspnetcore-3.1#ways-to-read-one-entity).

PR #15897, opened 2019-11-27
Call out header for Linux/Nginx+Blazor Server

guardrex:
Fixes #12107

This is untested (by me), and idk if a similar config applies to Apache. I can't find *any* Linux guidance in the SignalR topics.

It looks like the primary guidance here is for the `Connection` header because the rest of the suggestion on the issue matches what we have in the Linux Nginx hosting coverage. Over there, we show ...

```
proxy_pass         http://localhost:5000;
proxy_http_version 1.1;
proxy_set_header   Upgrade $http_upgrade;
proxy_set_header   Connection keep-alive;
proxy_set_header   Host $host;
proxy_cache_bypass $http_upgrade;
proxy_set_header   X-Forwarded-For $proxy_add_x_forwarded_for;
proxy_set_header   X-Forwarded-Proto $scheme;
```

... so on this PR, I just call out the `Connection` header being set to `$http_connection`.

Issue #11403, opened 2019-03-14
Missing SignalR config?


LucaGabi:
Is there nothing to configure for SignalR to work behing a ngnix reverse proxi ?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: dcdd2c35-bf2b-2673-d150-303a8aaa54d2
* Version Independent ID: 815fe034-c9fc-aee3-a8b5-f2e860b30302
* Content: [Host ASP.NET Core on Linux with Nginx](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/linux-nginx?view=aspnetcore-2.2)
* Content Source: [aspnetcore/host-and-deploy/linux-nginx.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/host-and-deploy/linux-nginx.md)
* Product: **aspnet-core**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

dustout:
I ran into the same issue. Adding the below location to nginx where /dataHub is the name of your hub should do the trick. 

```
location /dataHub {
        proxy_pass http://localhost:5000;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection $http_connection;
        #proxy_set_header Connection "upgrade";
        proxy_set_header Host $host;
        proxy_cache_bypass $http_upgrade;
    }
```

davidfowl:
cc @anurse @bradygaster this seems important

BrennanConroy:
Similar to/dupe of https://github.com/aspnet/Docs/issues/10827

guardrex:
Note to the content author: We have some movement on this subject for SignalR+WebSockets on Linux at https://github.com/dotnet/AspNetCore.Docs/pull/17181, but that's only a partial solution because it only directly addresses the subject in the Blazor topic.

That's a community generated PR, so it might be best to wait until that goes in and then perform some followup work. After that one is merged, I recommend moving the content to the Linux topics (per this issue) and cross-link to it from SignalR and Blazor topics.

Rick-Anderson:
Moved to Master issue: Host ASP.NET Core on Linux with Nginx #17735

guardrex:
[...]

PR #17181, opened 2020-03-02
proxy_set_header Connection "Upgrade"

timbze:
Updating to match nginx documentation (also linked in this page) which works. https://www.nginx.com/blog/websocket-nginx/

### DIFF
--- a/aspnetcore/host-and-deploy/blazor/server.md
+++ b/aspnetcore/host-and-deploy/blazor/server.md
@@ -5,7 +5,7 @@ description: Learn how to host and deploy a Blazor Server app using ASP.NET Core
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 02/15/2020
+ms.date: 03/03/2020
 no-loc: [Blazor, SignalR]
 uid: host-and-deploy/blazor/server
 ---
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks @timbze ... Hold a sec while I confer with our ⚔️ *Sergent at Code* 🏰 ...

@anurse When we added this on https://github.com/dotnet/AspNetCore.Docs/pull/15897#issuecomment-575697593, we didn't do the map bit. Was that a proper bit for us to have added at the time?

```
map $http_upgrade $connection_upgrade {
    default upgrade;
    '' close;
}
```

timbze:
I needed that line for SignalR to work, but now I'm running into this issue: https://github.com/dotnet/aspnetcore/issues/17081

timbze:
I ended up getting it working properly after using these lines:
`proxy_set_header Upgrade $http_upgrade;`
`proxy_set_header Connection $http_connection;`

And not this at all `proxy_set_header Connection "Upgrade";`

guardrex:
The whole scenario is spelled out here :point_right: http://nginx.org/en/docs/http/websocket.html.

That's probably a good resource to add via a cross-link. It looks like if we're going to go with `$http_connection` that we should have the `map` entry.

I'll make an update to this PR with what I think would be good and then we'll see on review if it's sane.

guardrex:
@timbze, can you confirm that the guidance there works in your scenario?

Also, I'm taking the example content from our Nginx topic, so I'm now showing the full set of headers. We'll have to see if that's the right thing to show.

**Also**, can you sign the CLA if you agree with its terms. We can't merge this PR without that.

guardrex:
FYI @scottaddie ... This PR looks like it could've been merged without the CLA being signed. It only asked for it when I shot in a commit :point_up:. If I hadn't done that, I think we'd have had a .NET Foundation process violation for accepting content. cc: @jongalloway

timbze:
@guardrex Yes it works for me with the example you have added now. And the other headers is exactly what I was using already too.

guardrex:
**_Cool Corebeans!_**:tm: :+1: ... so we just need this reviewed by one of the engineers that works more with Linux than I do, which is very rarely. :smile:

guardrex:
## Note to self

Work #11403 immediately after this merges.

anurse:
> @anurse When we added this on [#15897 (comment)](https://github.com/dotnet/AspNetCore.Docs/pull/15897#issuecomment-575697593), we didn't do the map bit. Was that a proper bit for us to have added at the time?

I don't remember that being there when I used that post as a reference for earlier WebSocket configuration docs, but it seems reasonable to add it now.

guardrex:
I'll merge now. I think we're good here. I'll tend to the additional work on this subject asap.

PR #17122, opened 2020-02-26
App State to 3.1

Rick-Anderson:
Fixes #13328

[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/app-state?view=aspnetcore-3.1&branch=pr-en-us-17122)

Changes:

* Update to 3.1, mostly `UseEndpoints` type changes.
* Removed references to application state since ASP.NET Core has never had application state like ASP.NET did.
* Add links to memory caching to store application wide data.

Engineering team review required for the following:

Use the following sample code to serialize objects:

```
public static class SessionExtensions
{
    public static void Set<T>(this ISession session, string key, T value)
    {
        session.SetString(key, JsonSerializer.Serialize(value));
    }

    public static T Get<T>(this ISession session, string key)
    {
        var value = session.GetString(key);
        return value == null ? default : JsonSerializer.Deserialize<T>(value);
    }
}
```

PR #17193, opened 2020-03-04
Topic author updates

guardrex:
I don't see any reason to keep the bylines ... they aren't really useful ... especially for someone getting close to retirement! *Woot!* 👴

### DIFF
--- a/aspnetcore/fundamentals/app-state.md
+++ b/aspnetcore/fundamentals/app-state.md
@@ -12,7 +12,7 @@ uid: fundamentals/app-state
 
 ::: moniker range=">= aspnetcore-3.0"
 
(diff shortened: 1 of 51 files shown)


aspnetcore/fundamentals/app-state.md
```diff
@@ -12,7 +12,7 @@ uid: fundamentals/app-state
```
Rick-Anderson: I'll fix this in #17122
back this out now so I don't get a merge conflict.

guardrex: oh, ok ... will do.

guardrex:
... and I left the Configuration topic alone. You can drop me from the author byline on that one, too.

PR #17258, opened 2020-03-10
Update index.md

Rick-Anderson:
See https://stackoverflow.com/questions/59822749/register-authentication-schemes-based-on-tenant-in-asp-net-core-3-1

### DIFF
--- a/aspnetcore/security/authentication/index.md
+++ b/aspnetcore/security/authentication/index.md
@@ -107,7 +107,12 @@ See the following links for differences between challenge and forbid:
 
 ## Authentication providers per tenant
 
-Orchard Core is an open-source modular and multi-tenant app framework built with ASP.NET Core, and a content management system (CMS) built on top of that app framework. See the [Orchard Core](https://github.com/OrchardCMS/OrchardCore) source for an example of authentication providers per tenant.
+Orchard Core is:
+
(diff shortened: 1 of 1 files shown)


aspnetcore/security/authentication/index.md
```diff
@@ -105,6 +105,15 @@ See the following links for differences between challenge and forbid:
 * [Challenge and forbid with an operational resource handler](xref:security/authorization/resourcebased#challenge-and-forbid-with-an-operational-resource-handler).
 * [Differences between challenge and forbid](xref:security/authorization/secure-data#challenge).
 
+## Authentication providers per tenant
+
+Orchard Core is:
```
mkArtakMSFT: Given that `Orchard Core` is not part of the ASP.NET Core framework, I would like this section to be a bit more verbose. Here is a suggestion. Feel free to use it or tweak it as necessary:

```suggestion
ASP.NET Core framework does not have a built-in solution for multi-tenant authentication.
While it's certainly possible for customers to write one, using the built-in features, we recommend customers to look into [Orchard Core](https://www.orchardcore.net/) for this purpose.

Orchard Core is:
```

Rick-Anderson:
@serpent5  can you vote [this](https://stackoverflow.com/a/60716345/502537) up given it's the official response?

PR #17295, opened 2020-03-13
Blazor options for 3rd party login providers

### LINKED ISSUE #17288: Add a section to describe options for hosted applications and third party login providers
Here is the content an the link to the Identity docs on how to do this https://docs.microsoft.com/en-us/aspnet/core/security/authentication/social/additional-claims?view=aspnetcore-3.1

### Content
When authenticating and authorizing a hosted Blazor app with a third party provider, there are several options available for authenticating the user. Which one you choose depends on your specific scenario:

**Do you want to authenticate a user to talk to a third party API?**
* You need to authenticate the user with a client-side oauth flow against the third-party API provider.
    ``` builder.services.AddOidcAuthentication(options =&gt; { ... })`
* The server does not play a role here.
* You can't protect APIs on the server in this way.
* You can only call protected APIs for the third party resource.

**Do you want to allow users to log-in with a third party provider and call protected APIs on your host server and the third party API resources?**
* You need to configure Identity with a third party login provider, get the tokens you need and store them.
* When a user logs-in this way, Identity will collect any access tokens and refresh tokens as part of the authentication process. At that point there are a couple of ways to make API calls to the 3rd party API:
  * Create an endpoint in your API and use the access token generated by the server to retrieve the access token to talk to the third party API resource from Identity. While this is doable, this is something that we do not recommend. The reason for this is that you are treating the access token as if it was emitted for a public client (in oauth terms, an app that doens't have a client secret because it can't be trusted to store secrets safely) when it might have been produced for a confidential client (a client that has a secret and that is assumed to be able to safely store secrets).
    * The implication of this is that that access token might have been granted additional scopes to perform sensitive operations based on the fact that the server considered it was emitting it for a more trusted client.
    * A similar example would be refresh tokens, you wouln't give them to a client you don't trust as that would give the client unlimited access (unless other restrictions are put into place)
  * Make API calls from your the client to server API and from there, retrieve the access token for the third party API resource and issue whatever call is necessary.
    * While this involves an extra hop through your server to talk to the 3rd party APIs, it ultimately results in a better experience:
      * The server can store refresh tokens and ensure the app doesn't loose access to the third party resources.
      * You don't leak any access token from the server that might contain more sensitive permissions.

---
#### Document Details

* ID: f34f7534-347a-5752-3127-fd774abea530
* Version Independent ID: af2c6c8a-d398-0202-4c28-cd5e15ce5219
* Content: [Secure ASP.NET Core Blazor WebAssembly](https://docs.microsoft.com/en-us/aspnet/core/security/blazor/webassembly/?view=aspnetcore-3.1)
* Content Source: [aspnetcore/security/blazor/webassembly/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/security/blazor/webassembly/index.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #17288

[Internal Review Topic (links directly to section)](https://review.docs.microsoft.com/en-us/aspnet/core/security/blazor/webassembly/index?view=aspnetcore-3.1&branch=pr-en-us-17295#options-for-hosted-apps-and-third-party-login-providers)

Check me here. :eyes: In an attempt to clarify some of the complex sentences, such as ...

> Create an endpoint in your API and use the access token generated by the server to retrieve the access token to talk to the third party API resource from Identity. 

... to ...

> Use the access token generated on the server to retrieve the third-party access token from a server API endpoint. From there, use the third-party access token to call third-party API resources directly from Identity on the client.

... *I may have gone right off the rails.* 🚂:boom:

### DIFF
--- a/aspnetcore/security/blazor/webassembly/index.md
+++ b/aspnetcore/security/blazor/webassembly/index.md
@@ -5,7 +5,7 @@ description: Learn how to secure Blazor WebAssemlby apps as Single Page Applicat
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 03/09/2020
+ms.date: 03/12/2020
 no-loc: [Blazor, SignalR]
 uid: security/blazor/webassembly/index
 ---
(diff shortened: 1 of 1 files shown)


guardrex:
@mkArtakMSFT I think Javier is 🏃 right now. I had to reword the technical guidance in a few spots because it was a bit confusing on token sources, sinks, and purposes. I think I have the flow right, but there's a chance that I mistranslated something.

The original content for comparison is on https://github.com/dotnet/AspNetCore.Docs/issues/17288.

PR #17335, opened 2020-03-16
Blazor Razor syntax remark improvements

### LINKED ISSUE #17119: No explanation of how @ prior to a parameter name affects the way the value is passed
[EDIT by guardrex to fix the encoding]

It seems like doing something like `<a @onclick="x" />` somehow does double binding against the member `x` (or one way binding if `x` is not writable). And that `<a onclick="@x" />` does one-way passing of the member.

Where is this explained? When should I use one style versus the other?

It seems that `@parameter="«expression»"` is used to pass a non-string value. But `parameter="@expression"` is really just a super-simple variant of `parameter="some string @withOptionalInterpolatedValues"`.

It’d be nice if the documentation was clear on this ;-).

---
#### Document Details

* Content: [Create and use ASP.NET Core Razor components](https://docs.microsoft.com/en-us/aspnet/core/blazor/components?view=aspnetcore-3.1#parameters)
* Content Source: [aspnetcore/blazor/components.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/components.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #17119

* Nits on how we lead into binding and event handling.
* There's an out-of-place remark in the *Components* topic. I move it to the *Binding* topic.
* I kill off 🔪 the "to both components and DOM elements" phrase in the opening sentence of the *Binding* topic. There are **only** components and DOM elements under discussion, so it's not really explaining anything. Also, the concept is well covered down in the topic.
* We've been instructed to use the word *scenario* over *feature*. I don't fully understand it, but that's the word.

### DIFF
--- a/aspnetcore/blazor/components.md
+++ b/aspnetcore/blazor/components.md
@@ -80,8 +80,6 @@ For information on setting an app's base path, see <xref:host-and-deploy/blazor/
 
 Components can include other components by declaring them using HTML element syntax. The markup for using a component looks like an HTML tag where the name of the tag is the component type.
 
-Attribute binding is case sensitive. For example, `@bind` is valid, and `@Bind` is invalid.
-
 The following markup in *Index.razor* renders a `HeadingComponent` instance:
 
 ```razor
--- a/aspnetcore/blazor/data-binding.md
+++ b/aspnetcore/blazor/data-binding.md
@@ -5,7 +5,7 @@ description: Learn about data binding scenarios for components and DOM elements 
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 02/24/2020
+ms.date: 03/16/2020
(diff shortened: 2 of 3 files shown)


guardrex:
That came down via docs management quite a while ago, and there's been a low-level background effort to factor out "feature" for quite a while now. Current counts put use at about 50-50 these days ...

* feature &ndash; 707
* scenario &ndash; 617

@Rick-Anderson can provide more color on that guidance. We spoke about it in detail when it came down, and the guidance was (and still is AFAIK) to mostly avoid "feature" and speak about "scenarios." I even went as far as to ask if I should avoid using the word in issues, and the answer was *yes* generally.

Rick-Anderson:
> avoid "feature" and speak about "scenarios." 

That's no longer stressed. Feature is fine.

PR #17353, opened 2020-03-18
Add the Blazor WebAssembly version to INCLUDES

### LINKED ISSUE #17349: Dozen of blazor dotnet command doesn't work if we are using a latest listed Microsoft.AspNetCore.Blazor.Templates package
All of the authentication commands (standalone) in docs doesn't work if we don't specifically follow get started page of the docs in which it is stated that 3.2.0-preview2.20160.5 should be used. 
However, it doesn't make sense to go through all the documentation each and every time you want to update your project so please list this version on nuget.org. You'll save people a lot of time.

guardrex:
Fixes #17349

Going ahead with that plan ... Given that the version is applied to all topics by two INCLUDES files, this is an easy way to signal the latest package version across the topics. It will be easy/fast to update each preview release.

Thanks @tomidix! 🌴


guardrex:
Artak, I'm going to go ahead on this one. Just adding the version number here. I'll keep up with it as we go preview-to-preview. These INCLUDES will be dropped from the repo soon as well.

PR #17359, opened 2020-03-18
fix #17358 by calling 'start()' in the JS sample

### LINKED ISSUE #17358: SignalR JS client example doesn't work. (solution within)
_From @TW8B on Wednesday, May 8, 2019 8:33:14 PM_

aspnetcore/signalr/javascript-client/sample/wwwroot/js/chat.js requires an iife around start to start. 

The sample appears to be based off the [JS SignalR tutorial](https://docs.microsoft.com/en-gb/aspnet/core/tutorials/signalr?view=aspnetcore-2.2&tabs=visual-studio) or vice versa, but with substantive differences around things like use or not of async. 

I'm happy to submit a pull request once I'm on a machine that isn't locked down.

```typescript
(async function start() {
    try {
        await connection.start();
        console.log("connected");
    } catch (err) {
        console.log(err);
        setTimeout(() => start(), 5000);
    }
})();
```

It was the only change required to make the thing run. There were some other issues with things like bootstrap being missing, but I didn't dig into that. Again, happy to have a poke.

I'm using VS 2019 latest.

Cheers

Edit: I could make the tutorial match or not match the sample, perhaps in a separate issue?

_Copied from original issue: dotnet/core#2695_

anurse:
Turns out the JS version of the sample doesn't actually work because we don't call the `async function start()` we create.

Fixes https://github.com/dotnet/AspNetCore.Docs/issues/17358 (originally reported in dotnet/aspnetcore).

### DIFF
--- a/aspnetcore/signalr/javascript-client/sample/wwwroot/js/chat.js
+++ b/aspnetcore/signalr/javascript-client/sample/wwwroot/js/chat.js
@@ -39,6 +39,9 @@ connection.onclose(async () => {
     await start();
 });
 
+// Start the connection.
+start();
+
 /* this is here to show an alternative to start, with a then
 connection.start().then(function () {
     console.log("connected");


Rick-Anderson:
@anurse  no need to pull in all of the doc reviewers on this.  
The change is not showing up in https://review.docs.microsoft.com/en-us/aspnet/core/signalr/javascript-client?view=aspnetcore-3.1&branch=pr-en-us-17359

where should the change be in the docs?

anurse:
@Rick-Anderson it's in the sample, this code doesn't appear in the narrative doc at all AFAIK.

anurse:
> @anurse no need to pull in all of the doc reviewers on this.

I tried to focus on people who had reviewed SignalR folks before. Let me know if there's a better set to tag in these PRs.

Rick-Anderson:
@anurse 
> Let me know if there's a better set to tag in these PRs.

If it's just code like us, we're not the right folks.
Usually it involves docs, in that case tag one of us.

The SignalR docs could use a lot of editing work, something we hope to get to eventually. Please tag us for doc changes/updates, but not code changes.

anurse:
Fair enough. Merging as we've had signoff from the product team.

PR #17361, opened 2020-03-19
Changed ASP.NET Core version "2.1" to "3.1"

alavifazel:
Changed a mistakenly written ASP.NET Core version "2.1" to "3.1" inside a paragraph.

### DIFF
--- a/aspnetcore/security/authentication/scaffold-identity.md
+++ b/aspnetcore/security/authentication/scaffold-identity.md
@@ -403,7 +403,7 @@ Delete the *Pages/Shared* folder and the files in that folder.
 
 To maintain full control of the Identity UI, run the Identity scaffolder and select **Override all files**.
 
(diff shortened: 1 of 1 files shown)


alavifazel:
Or it should be changed to something like "2.1 or above".

alavifazel:
Its corresponding code snippet (CompatibilityVersion.Version_2_1) should also be changed.

guardrex:
@alavifazel ... Check your document version selector in the upper left-hand corner of your screen. This PR is changing the version in the <3.0 version of the topic (see Line 264). I think you're just looking at the wrong version of the topic in your browser.

guardrex:
Oh ... I better let @Rick-Anderson look at this ... I see a 2.1 reference on Line 162. There might be something for that line instead. Stand-by @alavifazel ... he'll be on in a bit to look at this.

PR #17376, opened 2020-03-19
Update Blazor script sources

### LINKED ISSUE #17367: Unable to run with Microsoft Account
First, thank you for detailed guide. Unfortunately I couldn't get this to work with Microsoft (AAD) Authentication, both in my own app and in the sample application that is created by the "dotnet new blazorwasm -au Individual" command. There appeared to be several issues: 

1st: The sample application created by "dotnet -au" uses a different Javascript file. Your guide references to "_content/Microsoft.Authentication.WebAssembly.Msal/AuthenticationService.js" which belongs to the Microsoft-Account guide and won't work here. The sample application uses "_content/Microsoft.AspNetCore.Components.WebAssembly.Authentication/AuthenticationService.js" instead. 

2nd: In the sample application the value for "Authority" is set to "https://login.microsoftonline.com/". I believe this isn't a valid endpoint (at least I got a 404 with that). The url should end with something like "/common" or "/common/v2.0". 

3rd: After changing the url to https://login.microsoftonline.com/common/v2.0 I could click on "Log in" and got a promt to sign in to my Microsoft Account. It also redirected me back to my application which then stuck at "Completing your log in". I could see that it got an application token in the url, but it was unable to parse it, because the url was of the format "/login-callback#id_token=". After replacing the # with a ? the URL got parsed, but the script ended with an error somewhere deep inside the usermanager-Javascript saying that it could not find a valid state (despite one being shown in the debugger). I tried to debug and look into the Code on GitHub but couldn't find the source of this error. 

I gave up on this approach and repeated the steps from the "Standalone with Microsoft Accounts" instead (which worked in the end). 
Could you please tell me whether and what I did wrong? Or is this not intended to work with Microsoft Authentication (despite the url in the sample)?

---
#### Document Details

* ID: 55d96ce7-2ead-ebc2-973e-e319082937b0
* Version Independent ID: 44ee9a17-4e34-6463-66e8-f1eb212845a4
* Content: [Secure an ASP.NET Core Blazor WebAssembly standalone app with the Authentication library](https://docs.microsoft.com/en-us/aspnet/core/security/blazor/webassembly/standalone-with-authentication-library?view=aspnetcore-3.1)
* Content Source: [aspnetcore/security/blazor/webassembly/standalone-with-authentication-library.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/security/blazor/webassembly/standalone-with-authentication-library.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #17367

* This is *one way* that we can try and get AAD/B2C devs to skip this and go directly to the AAD/B2C topics in this node.
* The other set of updates here is to use two INCLUDES files for the *index.html* page in order to show the two different scripts in use across the node.


guardrex:
This includes a bug fix, so I'm going to go ahead and take it live.

PR #17425, opened 2020-03-23
Add an XML linker config file to a library

### LINKED ISSUE #17369: Document how to add an XML linker config file to a library
@guardrex Currently we have [docs](https://docs.microsoft.com/aspnet/core/host-and-deploy/blazor/configure-linker) on how to specify an XML linker config file for a Blazor app, but we don't document how to provide one for a library, like a component library. I've put together a sample that shows how to do this: https://github.com/danroth27/ComponentLibWithXmlLinkerConfig. Basically you embed the XML file into the library as an embedded resource with the correct name the linker will pick it up. We should document this pattern. Maybe we should elevate the Configure the Linker doc to be a top level Blazor topic and add it there? Or if you have a better location, that's fine.

Eventually, this content on embedding an XML link config in a library should probably live in the .NET docs - @samsp-msft is working on that. So it would be nice to keep this content mostly generic so that once it move to the .NET docs we can then juts link off to that content.

guardrex:
Fixes #17369

[Internal Review Topic (links to section)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/class-libraries?view=aspnetcore-3.1&branch=pr-en-us-17425#add-an-xml-linker-configuration-file-to-a-library)

In passing, I'm moving the *Create a Razor components class library with static assets* section up because I want the *Build, pack, and ship to NuGet* section to appear last.

### DIFF
--- a/aspnetcore/blazor/class-libraries.md
+++ b/aspnetcore/blazor/class-libraries.md
@@ -110,37 +110,7 @@ Include the `@using MyComponentLib1` directive in the top-level *_Import.razor* 
 
 An RCL can include static assets. The static assets are available to any app that consumes the library. For more information, see <xref:razor-pages/ui-class#create-an-rcl-with-static-assets>.
 
-## Add an XML linker configuration file to a library
 
-Embed the XML file into the library as an embedded resource.
-
-For example, create a *LinkerConfig.xml* file in the library that targets the `System.Linq.Queryable` type in the `System.Core` assembly:
-
-```xml
-<?xml version="1.0" encoding="utf-8" ?>
-<linker>
-  <assembly fullname="System.Core">
-    <type fullname="System.Linq.Queryable" preserve="all" />
-  </assembly>
-</linker>
-```
-
-For information on the IL linker format, see [Link xml file examples (mono/linker GitHub repository)](https://github.com/mono/linker#link-xml-file-examples).
-
-The library's project file has:
-
-* A package reference for [System.Linq.Dynamic.Core](https://www.nuget.org/packages/System.Linq.Dynamic.Core/).
-* The *LinkerConfig.xml* file specified as an embedded resource.
-
-```xml
-<ItemGroup>
-  <EmbeddedResource Include="LinkerConfig.xml">
-    <LogicalName>$(MSBuildProjectName).xml</LogicalName>
-  </EmbeddedResource>
-</ItemGroup>
-```
-
-For more information, see <xref:host-and-deploy/blazor/configure-linker>.
 
 ## Build, pack, and ship to NuGet
 
@@ -155,3 +125,4 @@ Upload the package to NuGet using the [dotnet nuget push](/dotnet/core/tools/dot
 ## Additional resources
 
 * <xref:razor-pages/ui-class>
+* [Add an XML linker configuration file to a library](xref:host-and-deploy/blazor/configure-linker#add-an-xml-linker-configuration-file-to-a-library)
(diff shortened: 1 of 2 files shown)


aspnetcore/blazor/class-libraries.md
```diff
@@ -106,6 +106,42 @@ Welcome to your new app.
 
 Include the `@using MyComponentLib1` directive in the top-level *_Import.razor* file to make the library's components available to an entire project. Add the directive to an *_Import.razor* file at any level to apply the namespace to a single page or set of pages within a folder.
 
+## Create a Razor components class library with static assets
+
+An RCL can include static assets. The static assets are available to any app that consumes the library. For more information, see <xref:razor-pages/ui-class#create-an-rcl-with-static-assets>.
+
+## Add an XML linker configuration file to a library
```
danroth27: This content looks good, but it feels a bit out of context because all of the content explaining what the linker does and why a config file might be necessary lives in the "Configure the linker" doc. I know we discussed putting this content here in the component libraries doc, but technically this content isn't specific to component libraries. It applies to *any* library that you might be using in a Blazor WebAssembly app. I think we should move this section to the end of the "Configure the linker" doc. 

Also, the "Configure the linker" doc points to https://github.com/mono/linker/blob/master/src/linker/README.md#syntax-of-xml-descriptor for the syntax of the XML file, but there is a much more extensive syntax description that I think we should link to instead here: https://github.com/mono/linker#link-xml-file-examples.

aspnetcore/blazor/class-libraries.md
```diff
@@ -106,6 +106,42 @@ Welcome to your new app.
...
+
+Embed the XML file into the library as an embedded resource.
```
danroth27:
```suggestion
Embed the XML file into the library as an embedded resource that has the same name as the assembly.
```

aspnetcore/blazor/class-libraries.md
```diff
@@ -106,6 +106,42 @@ Welcome to your new app.
...
+    <type fullname="System.Linq.Queryable" preserve="all" />
+  </assembly>
+</linker>
+```
+
+For information on the IL linker format, see [Link xml file examples (mono/linker GitHub repository)](https://github.com/mono/linker#link-xml-file-examples).
+
+The library's project file has:
+
+* A package reference for [System.Linq.Dynamic.Core](https://www.nuget.org/packages/System.Linq.Dynamic.Core/).
```
danroth27: Actually, can we not use System.Linq.Dyanmic.Core as an example? The community is working with the owner of that library to include the XML linker config in that library, so that consumers of the library no longer need to.

guardrex:
@danroth27 ... ate dinner 🍗 ... 2nd wind! ... *Yes!* This composes much better now.

aspnetcore/host-and-deploy/blazor/configure-linker.md
```diff
@@ -78,7 +78,21 @@ Control linking on a per-assembly basis by providing an XML configuration file a
 </linker>
 ```
 
-For more information, see [IL Linker: Syntax of xml descriptor](https://github.com/mono/linker/blob/master/src/linker/README.md#syntax-of-xml-descriptor).
+For more information, see [Link xml file examples (mono/linker GitHub repository)](https://github.com/mono/linker#link-xml-file-examples).
+
+## Add an XML linker configuration file to a library
+
+Embed an XML linker configuration file into a library as an embedded resource in the library's project file.
```
danroth27:
```suggestion
To configure the linker for a specific library, add an XML linker configuration file into the library as an embedded resource. The embedded resource must have the same name as the assembly.
```

PR #17440, opened 2020-03-24
Blazor WASM configuration

### LINKED ISSUE #17419: Configuration for Blazor-wasm
Blazor WebAssembly 3.2-preview3 adds support for configuration. Blazor will populate config from `wwwroot/appsettings.json` and `wwwroot/appsettings.${HostingEnvironment}.json` from the Blazor WebAssembly (client) app. 

* In a hosted application, the value of `HostingEnvironment` is the same as the server app's value. See <xref:aspnet/core/fundamentals/environments>
* `HostingEnvironment` is always `Development` when using the Blazor Development server. In non-ASP.NET Core hosts, this value would be `Production` by default.

Warning: Configuration in a Blazor WebAssembly application is visible to users and should not be used to store application secrets or credentials.

guardrex:
Fixes #17419

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/hosting-model-configuration?view=aspnetcore-3.1&branch=pr-en-us-17440)

### DIFF
--- a/aspnetcore/blazor/hosting-model-configuration.md
+++ b/aspnetcore/blazor/hosting-model-configuration.md
@@ -24,9 +24,9 @@ As of the ASP.NET Core 3.2 Preview 3 release, Blazor WebAssembly supports config
 * *wwwroot/appsettings.json*
 * *wwwroot/appsettings.{ENVIRONMENT}.json*
 
-In a Blazor Hosted app, the value of `HostingEnvironment` is the same as the server app's value.
+In a Blazor Hosted app, the [runtime environment](xref:fundamentals/environments) is the same as the server app's value.
 
-`HostingEnvironment` defaults to the Development environment when using the Blazor development server. In non-ASP.NET Core hosts, the Production environment is used by default. For more information, see <xref:fundamentals/environments>.
+When running the app locally, the environment defaults to Development. When the app is published, the environment defaults to Production. For more information, including how to configure the environment, see <xref:fundamentals/environments>.
 
 > [!WARNING]
 > Configuration in a Blazor WebAssembly app is visible to users. **Don't store app secrets or credentials in configuration.**
--- a/aspnetcore/blazor/hosting-model-configuration.md
+++ b/aspnetcore/blazor/hosting-model-configuration.md
@@ -5,7 +5,7 @@ description: Learn about Blazor hosting model configuration, including how to in
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 02/12/2020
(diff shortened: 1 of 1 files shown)


pranavkm:
@guardrex could we document a special case for PWA - The config JSON file will be cached for offline use. Note that with PWAs specifically, since the configs are cached, you can only edit it when making a new deployment. There's no point editing it between deployments because users will already have a cached version of the file which would continue to be used.

guardrex:
@pranavkm Need a clarification on that: PWAs update in the background and will pull down the new file **_when?_** ...

* When merely the updated config file alone is updated on the server (i.e., not the whole deployment).
* When the whole app is redeployed with the updated config file.

pranavkm:
> When the whole app is redeployed with the updated config file.

would be the right answer. There's no mechanism to update it in PWAs if the config file is independently updated.

guardrex:
I see the *service worker assets manifest* doesn't get updated until the app is recompiled.

aspnetcore/blazor/hosting-model-configuration.md
```diff
@@ -17,11 +17,26 @@ By [Daniel Roth](https://github.com/danroth27)
...
 
-<!-- For future use:
-
 ## Blazor WebAssembly
 
--->
+As of the ASP.NET Core 3.2 Preview 3 release, Blazor WebAssembly supports configuration from:
+
+* *wwwroot/appsettings.json*
+* *wwwroot/appsettings.{ENVIRONMENT}.json*
```
pranavkm: Should this say `HostingEnvironment` instead of `Environment`? It's referred to as `HostingEnvironment` every where else that follows.

guardrex: Yes ... I was making this match the way that we normally use the placeholder for this file.

I still need to hear back on the question that I ask in the opening comment and then I can fix it all up.

guardrex:
@pranavkm Try it now. My concern about "Blazor development server" is just a nit, but it sounds like it might be *something new/different* based on all of our other docs.

aspnetcore/blazor/hosting-model-configuration.md
```diff
@@ -17,11 +17,26 @@ By [Daniel Roth](https://github.com/danroth27)
...
+
+In a Blazor Hosted app, the [runtime environment](xref:fundamentals/environments) is the same as the server app's value.
+
+When running the app locally, the environment defaults to Development. When the app is published, the environment defaults to Production. For more information, including how to configure the environment, see <xref:fundamentals/environments>.
```
pranavkm: Much better.

PR #17457, opened 2020-03-25
Switch + and - signs for PackageReference

khalidabuhakmeh:
I think you meant it the other way right?


scottaddie:
Nice catch, @khalidabuhakmeh! I clearly needed coffee when I wrote that.

khalidabuhakmeh:
Woot 🎉🎉🎉👍

PR #17469, opened 2020-03-26
Update sdk.md

pranavkm:


guardrex:
@pranavkm ❓ ... If this is the recommended alternative and since this content is versioned, would it better to **_replace_** the current entry with the new property completely.

btw - either way - we can remove the version bit from that line, since this content is versioned >=3.0.

pranavkm:
It's a bit weird, because the property if a feature of the .NET Sdk (which does not have any documentation). `CopyRefAssembliesToPublishDirectory` is the property RazorSDK defines and understands, and it aliases it to `PreserveCompilationReferences` when present.

guardrex:
Perhaps the best way to go for a while is to have **_two entries_** in the table, where `CopyRefAssembliesToPublishDirectory` recommends to use `PreserveCompilationReferences` for >=3.0.

pranavkm:
Better?

guardrex:
**_Yikes!!!!_** 😨 ... that will :boom: *badly break* :boom: the table. :smile: lol

I'll fix it up.

guardrex:
Ok 😅 ..... it's happy now.

pranavkm:
🚢 🇮🇹

guardrex:
:smile: ..... it's a bit fragile .... a table like that could bring the docs build system to its knees. lol

PR #17477, opened 2020-03-26
Fix Typo

Vuurwarm:
"Tun" should be "Run"


guardrex:
I just took the change live, too. It should hit the live site in a few minutes.

If I didn't do that, *I'd be hearing about that one all evening!* lol

PR #17499, opened 2020-03-28
Update app-component.md

IIStevowII:
Added using statement that was missing.

### DIFF
--- a/aspnetcore/includes/blazor-security/app-component.md
+++ b/aspnetcore/includes/blazor-security/app-component.md
@@ -5,6 +5,7 @@ The `App` component (*App.razor*) is similar to the `App` component found in Bla
 * The `RedirectToLogin` component manages redirecting unauthorized users to the login page.
 
 ```razor
+@using Microsoft.AspNetCore.Components.Authorization
 <CascadingAuthenticationState>
     <Router AppAssembly="@typeof(Program).Assembly">
         <Found Context="routeData">


guardrex:
Hello @IIStevowII ... This INCLUDE file is only used in the standalone and hosted Blazor WebAssembly security topics, and the Blazor WebAssembly template doesn't include it ...

https://github.com/dotnet/aspnetcore/blob/blazor-wasm/src/ProjectTemplates/ComponentsWebAssembly.ProjectTemplates/content/ComponentsWebAssembly-CSharp/Client/App.razor

Were you thinking of Blazor Server when you sent this in?

IIStevowII:
Hi @guardrex, I don't think so. I was following the guide for a Blazor project, containing a Client and Server project. That's a hosted Blazor WebAssembly right? I added the code to my App.razor file and it complained about not understanding CascadingAuthenticationState, that's why I added the using.

guardrex:
> added the code

*Ah!* :smile: ... so you didn't use the template. Normally, I guess engineering would like it to be in the *_Imports.razor* file ...

https://github.com/dotnet/aspnetcore/blob/blazor-wasm/src/ProjectTemplates/ComponentsWebAssembly.ProjectTemplates/content/ComponentsWebAssembly-CSharp/Client/_Imports.razor#L4

I **missed** 🙈 mentioning this when setting these topics up.

What I'd like to do is not change this file ... but show the *_Imports.razor* file in the topic.

I opened https://github.com/dotnet/AspNetCore.Docs/issues/17501 to address it. Thanks for sending this in.

IIStevowII:
Ah! Didn't knew that. I used the template when creating a new project from Visual Studio and went from there. Totally missed the import file. Including them as snippets would be great, thanks!

PR #17505, opened 2020-03-29
Update Startup2.cs

garyray-k:
Being more explicit with the naming of these variables allows for readers of the code to easily digest what the variable actually is and what it's purpose may be.

Related
#17504

PR #17504, opened 2020-03-29
Update Startup2.cs

garyray-k:
Being more explicit with the naming of these variables allows for readers of the code to easily digest what the variable actually is and what it's purpose may be.

PR #17506, opened 2020-03-29
Update Startup2.cs

garyray-k:
Being more explicit with the naming of these variables allows for readers of the code to easily digest what the variable actually is and what it's purpose may be.

Related
#17505 
#17504


garyray-k:
So I went down this rabbit hole and did a regex search for any lamda expression with single letter variables. I plan to update all of these with a succinct and more descriptive name. 

My goal is to make the docs friendlier to new devs that may not have a grasp on lambda yet. A single letter may be descriptive enough with the surrounding context but I argue that this can lead to confusion and wasted brain cycles to determine a possible type for the variable in the lambda. This is especially so when you don't understand what a lambda is doing.

I'm also finding that sometimes the single variable letter doesn't even match the overall type name, which is further confusing.

All this to ask, would this change be considered if I go through and update all the single letters to meaningful names? I have begun but my simple regex has now uncovered ~1200 instances to be chagned.
image

I'm open to counter points/arguments.

Rick-Anderson:
Waiting for single PR

PR #17511, opened 2020-03-29
Upgrade the info about exposing attribute routes to Warning

brgrz:
Makes it more obvious you have to do something special, e.g. call MapControllers() to expose attribute only routes on regular (non-API) controllers. If you just comment out the default conventional route and add attribute routes to controllers, it won't work. And it is not immediately obvious why it doesn't work. This change makes those two lines more prominent.


scottaddie:
@brgrz Thank you for the PR!

PR #17544, opened 2020-03-31
[fix] Removed double using

rdeveen:
Fix a typo.


scottaddie:
Thank you for the fix, @rdeveen!

PR #17559, opened 2020-04-01
Update breadcrumb toc.yml to point to new hub page

wadepickett:
[Internal Review - see breadcrumb at top](https://review.docs.microsoft.com/en-us/aspnet/core/introduction-to-aspnet-core?view=aspnetcore-3.1&branch=pr-en-us-17559)

Updating the breadcrumb/toc.yml so it points to the new hub page instead of the old one.

- I will remove the old unused hub page in a separate PR once this last reference to it is updated.
- The ASP.NET 4.x breadcrumb that point to the old hub page will also have to be changed in a separate PR.

### DIFF
--- a/aspnetcore/breadcrumb/toc.yml
+++ b/aspnetcore/breadcrumb/toc.yml
@@ -4,7 +4,7 @@
   items:
   - name: ASP.NET
     tocHref: /aspnet/
-    topicHref: /aspnet/index
+    topicHref: /aspnet/core/index
     items:
     - name: ASP.NET Core
       tocHref: /aspnet/core/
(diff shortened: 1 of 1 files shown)


scottaddie:
I think the breadcrumb trail should go *Docs* > *ASP.NET Core*. It doesn't make much sense to me to have *Docs* > *ASP.NET* > *ASP.NET Core* if both *ASP.NET* and *ASP.NET Core* point to the same page. Thoughts, @wadepickett & @Rick-Anderson?

Rick-Anderson:
@scottaddie  agreed. @wadepickett  can you do another PR and fix that?

wadepickett:
Sorry Scott, I merged a minute before your reply.     Given .NET 5, would we instead go to this?

Docs > .NET > ASP.NET Core


I agree on getting rid of the extra "ASP.NET".  That never made sense to me.

scottaddie:
@wadepickett No worries. Sorry I couldn't look at this sooner. I think your proposal could work well.

PR #17571, opened 2020-04-01
Rendering components in RP/MVC updates

### LINKED ISSUE #17210: Class instance as parameter
[**EDIT by guardrex** Note to self ... see :point_right: https://github.com/dotnet/aspnetcore/issues/19605]

A class instance (that has public properties that are set by a default constructor) as a parameter causes the razor component to not be displayed.

Can the following quote be confirmed?

> The parameter type must be JSON serializable, which typically means that the type must have a default constructor and settable properties.

I checked if it was JSON serializable with the following, and it serializes it as expected. I also included the call to the `<component />`.

Code
```
var myObject = new MyClass(5, "test");
var serialized = Newtonsoft.Json.JsonConvert.SerializeObject(myObject);
<component type="typeof(MyComponent)" render-mode="Server" param-MyObject="myObject" />
```

MyClass
```
public class MyClass
{
	public int MyInt { get; set; }
	public string MyString { get; set; }

	public MyClass(int myInt, string myString)
	{
		MyInt = myInt;
		MyString = myString;
	}
}
```

MyComponent
```
@MyObject.MyInt, @MyObject.MyString

@code
{
	[Parameter]
	public MyClass MyObject { get; set; }
}
```

---
#### Document Details

* ID: fcac856a-7cdc-41fb-b7ea-7c5583916019
* Version Independent ID: b6f9f287-407a-6036-10b0-4bc13cfade17
* Content: [Integrate ASP.NET Core Razor components into Razor Pages and MVC apps](https://docs.microsoft.com/en-us/aspnet/core/blazor/integrate-components?view=aspnetcore-3.1)
* Content Source: [aspnetcore/blazor/integrate-components.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/integrate-components.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #17210

[Internal Review Topic (Component TH topic)](https://review.docs.microsoft.com/en-us/aspnet/core/mvc/views/tag-helpers/built-in/component-tag-helper?view=aspnetcore-1.0&branch=pr-en-us-17571)

Surfaces the guidance that we discussed on the issues. Thanks @mrlife! 🎷 ... and sorry for the delay, but I've been 🏃😅 lately.

WRT my problem on the issue: My problem was merely that I wasn't following the integration prereqs that we established earlier for the *Integrate components* topic. I was merely missing the *blazor.server.js* script. 👦🔫 Once I figured that out and dropped the script in there, it was :tada:. I've updated the text here to help others avoid my mistake.

 One other thing to ask about here ... all of the `render-mode` guidance seems to be specifically angled toward the `App` component. Here's what we're saying today ...

<hr>

`RenderMode` configures whether the component:

* Is prerendered into the page.
* Is rendered as static HTML on the page or if it includes the necessary information to bootstrap a Blazor app from the user agent.

| Render Mode | Description |
| ----------- | ----------- |
| `ServerPrerendered` | Renders the component into static HTML and includes a marker for a Blazor Server app. When the user-agent starts, this marker is used to bootstrap a Blazor app. |
| `Server` | Renders a marker for a Blazor Server app. Output from the component isn't included. When the user-agent starts, this marker is used to bootstrap a Blazor app. |
| `Static` | Renders the component into static HTML. |

<hr>

... and you can see that I took that right from the ref source. However, that seems awfully `App` component-specific (e.g., "marker for a Blazor Server app", "bootstrap a Blazor app"). If that's correct language, ok ... we're good. If those descriptions should state things more generally for **_components_**, then help me craft some lingo for it ... I'll update those descriptions on this PR.

### DIFF
--- a/aspnetcore/blazor/integrate-components.md
+++ b/aspnetcore/blazor/integrate-components.md
@@ -5,7 +5,7 @@ description: Learn about data binding scenarios for components and DOM elements 
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 03/17/2020
+ms.date: 04/01/2020
 no-loc: [Blazor, SignalR]
 uid: blazor/integrate-components
 ---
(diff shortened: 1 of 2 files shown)


guardrex:
@mkArtakMSFT ... you didn't remark on my ask about `render-mode` guidance, which seems to be specifically angled toward the `App` component. If the lingo is fine, then we can go ahead. If not tho, I'd like some engineering guidance on how to make it more generic for *components*.

PR #17671, opened 2020-04-08
[Preview 4] Logging configuration in Blazor WASM

### LINKED ISSUE #17665: [Preview 4] Document Logging configuration in Blazor WASM preview4
This is a request to document the changes introduced in https://github.com/dotnet/aspnetcore/pull/20432.

With these changes, developers can now configure logging on their Blazor WASM applications by accessing the `Logging` property on the `WebAssemblyHostBuilder`. For example:

```csharp
using Microsoft.Extensions.Logging;

namespace BasicTestApp
{
    public class Program
    {
        public static async Task Main(string[] args)
        {
            var builder = WebAssemblyHostBuilder.CreateDefault(args);

            builder.Logging.Services.AddSingleton<ILoggerProvider, PrependMessageLoggerProvider>(s =>
                new PrependMessageLoggerProvider("Custom logger", s.GetService<IJSRuntime>()));
            builder.Logging.SetMinimumLevel(LogLevel.Debug);
            builder.Logging.AddProvider(new SomeLoggingProvider());

            var host = builder.Build();
            await host.RunAsync();
        }
    }
}

```

The `builder.Logging` property is of type `ILoggingBuilder` so all extension methods available on `ILoggingBuilder` are also available on `builder.Logging`.

guardrex:
Fixes #17665

[Internal Review Topic (links to section)](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/index?view=aspnetcore-2.1&branch=pr-en-us-17671#create-logs-in-blazor-webassembly)

❓ ...

1. Setting the logging level and adding a simple provider will likely be the most common use cases in the example code. Are you 👍 with my placing those two lines first?
1. `PrependMessageLoggerProvider` seems like an advanced scenario. If it appears, it seems like it should be explained (and probably shown). I noticed that you removed the commented description of it at https://github.com/dotnet/aspnetcore/pull/20432/files#diff-f86c9ef39022e31d7bac9d1aedb571cbL9. Do you want to ...
   * Only say what it is here? Is the description that you cut ok, or should we say something else about it? I'm concerned about this choice because it isn't going to help much in a fundamentals topic.
   * Say what it is, and show it? I need the description. Is it ok to use the code from the **BasicTestApp**? I'm concerned about this choice because it's going to take up a lot of space, and this seems like an advanced scenario. If we go this route, perhaps this bit should be over in our [Blazor Advanced Scenarios topic](https://docs.microsoft.com/aspnet/core/blazor/advanced-scenarios) with a cross-link from here (and the WASM hosting model config topic) to there.
   * **_Kill it!_** 🔪
1. Did I miss any other required namespaces that aren't in the template for the code example? (Devs regularly express their 😠 when we don't show them.)
1. WRT `IConfiguration` (https://github.com/dotnet/aspnetcore/issues/20231), I guess we don't need to say anything ... it's about to become 👍 **_A Thing!:tm:_** 👍 for Preview 5.


captainsafia:
> Setting the logging level and adding a simple provider will likely be the most common use cases in the example code. Are you 👍 with my placing those two lines first?

Yes. That seems sensible to me.

> PrependMessageLoggerProvider seems like an advanced scenario. If it appears, it seems like it should be explained (and probably shown). I noticed that you removed the commented description of it at

It's probably safe to remove it. I included it in the example in the original issue to show users the variety of ways in which they can configure providers (via both the `AddProvider` extension method and directly by modifying the `Services` object). To keep the docs simple, we can axe it for now.

> Did I miss any other required namespaces that aren't in the template for the code example? (Devs regularly express their 😠 when we don't show them.)

We'll probably want to include the following for completeness.

```
using System;
using Microsoft.AspNetCore.Components.WebAssembly.Hosting;
using Microsoft.AspNetCore.Components.WebAssembly.Services;
using Microsoft.Extensions.Logging;
```

Note that I have removed `Microsoft.JSInterop` because it won't really be needed if we remove the `PrependMessageProvider` per the note above.

> WRT IConfiguration (dotnet/aspnetcore#20231), I guess we don't need to say anything ... it's about to become 👍 A Thing!™️ 👍 for Preview 5.

Yes, I'll open another issue with documentation details for that once it is in.

Thanks!

guardrex:
Very good ... thanks.

> We'll probably want to include the following for completeness.

Our repo guidelines ... *non-verbal and assumed in this case* :smile: (there are a lot of them 🙈) ... are only to show the `usings` for the namespaces involved in the example at hand. We don't show the additional ones that the template provides because they take up too much doc space for little benefit. The complaint from readers is almost never about the existing namespaces. They sharpen their 🦅 *beaks and claws* 🦅 for us when they go to drop the code into their app, and :boom: ... they need the namespace(s) that go with the code example.

I get the `Microsoft.Extensions.Logging` need, but what about `Microsoft.AspNetCore.Components.WebAssembly.Services`? That one isn't provided by the template (p3 template anyway, not sure about p4) ... will that one be required for ...

```csharp
builder.Logging.SetMinimumLevel(LogLevel.Debug);
builder.Logging.AddProvider(new CustomLoggingProvider());
```

aspnetcore/fundamentals/logging/index.md
```diff
@@ -287,6 +287,25 @@ public class Program
...
+### Create logs in Blazor WebAssembly
+
+Configure logging in Blazor WebAssembly apps with the `WebAssemblyHostBuilder.Logging` property in `Program.Main`:
+
+```csharp
+using Microsoft.Extensions.Logging;
+
+...
+
+builder.Logging.SetMinimumLevel(LogLevel.Debug);
```
captainsafia, line 301: Perhaps we should show the line `var builder = WebAssemblyHostBuilder.CreateDefault(args);` to make it clear that `builder` has to be `WebAssemblyHostBuilder` type.

guardrex:
**_Done!_** :+1:

I'll take care of merging this on Preview 4 release day. There will be several of these, and they'll all go in together right after the DR blog post goes live.

guardrex:
@mkArtakMSFT All good on this one?

PR #17384, opened 2020-03-20
C177636 Escaping Asterisks

V-vafune:
Localization team has reported source content issue that causes localized version to have broken/different format compared to en-us version.   
Description: Asterisks (*) are breaking right rendering on LOC pages.
Affected Lines 55 and 56

PR #17693, opened 2020-04-09
C177636 : removing extra hyphen to avoid broken display of table

v-maudel:
Description: Extra hyphen is broken display on table for this page https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/directory-structure?view=aspnetcore-2.2


v-maudel:
Hi @Rick-Anderson , It seems our previous [fix](https://github.com/dotnet/AspNetCore.Docs/pull/17384) didn't work. Could you please help us with this one? For aspnetcore > 2.2 the hyphen I removed is not present. 
Regards

guardrex:
The "fix" ...

> ... .EXTENSION} \*.exe* extension on ...

... seems like it would break the rendering because escaping the first asterisk isn't going to result in emphasis. Perhaps since this one topic (the table perhaps) is causing so much trouble, the asterisks here should be removed in favor of just text extensions ...


> ... .EXTENSION} .exe extension on ...

... or code ...

> ... .EXTENSION} \`.exe\` extension on ...

cc: @scottaddie

Rick-Anderson:
@guardrex @scottaddie  this is rendering correctly. I'm going to S&M and push to live unless I hear an objection.

guardrex:
The asterisks aren't right ... a quick edit could just make the extensions plain text to resolve it.

guardrex:
Interesting how they are rendered correctly. Seems strange tho that it works that way.

Rick-Anderson:
@guardrex can you edit/fix it. I'd go with 

`... .EXTENSION} .exe extension on ...`

The simplest approach.

guardrex:
Ok ... I think that licked it.

Rick-Anderson:
@guardrex  nice work

v-maudel:
Thanks @Rick-Anderson, @guardrex

PR #17694, opened 2020-04-09
Fix missing @ before typeof(Startup)

inouiw:
Fixes an error in a code snipped of the docs under: https://docs.microsoft.com/en-us/aspnet/core/blazor/layouts?view=aspnetcore-5.0#default-layout


scottaddie:
@inouiw Thank you again for your help.

PR #17695, opened 2020-04-09
Fix missing @ before typeof(Startup)

inouiw:
Fixes an error in the second layout code example of the docs under: https://docs.microsoft.com/en-us/aspnet/core/blazor/layouts?view=aspnetcore-5.0#default-layout


scottaddie:
@inouiw This is probably a good change, as it mirrors the project template's file. If you're curious, it is valid to omit the leading `@` sign in this location. Thank you for the PR!

PR #17734, opened 2020-04-10
Add MinimumProtocolVersion

BrennanConroy:
`MinimumProtocolVersion` was added in 3.1 but never added to the docs.

Had to copy the whole 3.0 section because we don't do monikers for subsections anymore. The only change was adding MinimumProtocolVersion to the table.


Rick-Anderson:
@BrennanConroy  it's OK to use monikers for subsections if you don't have more than a few pairs. But if you're doing 6 or more moniker pairs,  it's better to duplicate the entire version.

PR #17778, opened 2020-04-13
Added a new External resources entry

akovac35:
Please consider accepting this proposal. Logging in .NET Core and ASP.NET core seems quite well laid out until you get into details required for enterprise applications, and then good samples are quite useful.

### DIFF
--- a/aspnetcore/fundamentals/logging/index.md
+++ b/aspnetcore/fundamentals/logging/index.md
@@ -1180,14 +1180,9 @@ Using a third-party framework is similar to using one of the built-in providers:
 
 For more information, see each provider's documentation. Third-party logging providers aren't supported by Microsoft.
 
-## Third-party logging samples
-
-Third-party logging samples that demonstrate logging in .NET Core and ASP.NET Core:
-
-* [Logging.Samples](https://github.com/akovac35/Logging.Samples)
-
-Third-party logging samples aren't supported by Microsoft.
-
 ## Additional resources
 
 * <xref:fundamentals/logging/loggermessage>
+* [Logging.Samples](https://github.com/akovac35/Logging.Samples) - Samples for .NET Core and ASP.NET Core logging using various logger frameworks and Microsoft.Extensions.Logging.ILogger interface.
+
+Third-party resources aren't supported by Microsoft.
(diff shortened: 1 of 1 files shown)


aspnetcore/fundamentals/logging/index.md
```diff
@@ -1180,6 +1180,14 @@ Using a third-party framework is similar to using one of the built-in providers:
 
 For more information, see each provider's documentation. Third-party logging providers aren't supported by Microsoft.
 
+## Third-party logging samples
```
Rick-Anderson, line 1183: Not saying we'll accept this PR, buf if you do, you'd need to move this to `## Additional resources` and make it one line.

Rick-Anderson:
@scottaddie  @guardrex

aspnetcore/fundamentals/logging/index.md
guardrex: Not sure this meets DR's cutoff for *value*. AFAIK IIRC, the official provider docs have Core samples and link to additional external resources. Also, we haven't looked at this, and *someone* should look. Official docs probably just can't link anything sent in without review.

@akovac35, this would be good for the Community Standup and https://github.com/quozd/awesome-dotnet.

akovac35: Thanks for reviewing.

I have updated the proposal as suggested by @Rick-Anderson. @guardrex: thanks for suggestions. The official docs have samples which cover only the very basics. The samples I prepared cover much, much more and that for the Microsoft.Extensions.Logging.ILogger interface in various scenarios.

Because of using third party resources as logger providers (NLog, Serilog) I could not merge them with your samples, so a new repository was created. I think the value provided for site users exceed reasonable thresholds, I really covered many details.

guardrex:
Thanks @akovac35 for offering this, but I just don't feel *safe enough* about it to blindly place the cross-link into the doc. The logging provider official docs have Core samples, and devs can find additional Core-based implementations for those providers out on the Net. We would need to review your content because we can't just blindly cross-link resources in here ... we'd be **_in a world of trouble with management_** :boy::gun: if we did that and something :boom: badly for someone. We just don't have the staffing to look at this time. We also have no way of knowing if the content will be maintained ... updated for each release of every 3rd party provider in your samples.

> I could not merge them with your samples

We do use 3rd party bits in our samples, and there wouldn't be a technical barrier to doing so here. It's just that we *try not to do that* because we are challenged to maintain the 3rd party bits as they release new versions. Again, we don't have the staffing to provide that level of support for 3rd party. Therefore, we want the 3rd parties to document their own stuff and provide their own samples.

Thanks for working on this and offering it. I think that your samples would make a great callout for the Community Standup and for Awesome .NET, and I hope you'll pursue those channels.

PR #17805, opened 2020-04-14
Remove [ViewData] example using @Model with a controller

### LINKED ISSUE #17733: Getting runtime error while displaying ViewData attribute value using @Model on view
I am using _ViewData_ attribute to display its value on _View_ using _Model_.

Here is the code from my controller file:

` public class HomeController : Controller
    {
        [ViewData]
        public string Title { get; set; }

        public ViewResult Index()
        {
            Title = "Home page title";

            return View();
        }
    }`

Now as per the [documentation ](https://docs.microsoft.com/en-us/aspnet/core/mvc/views/overview?view=aspnetcore-3.1), there are two ways to display this data on view file.
1. We can display this data like below
`  <h1>@ViewData["Title"]</h1>`

Using this approach data gets displayed on view.

2. We can display this data using Model and it gives an error:
` <h1>@Model.Title</h1>`

Here is the error:

> RuntimeBinderException: Cannot perform runtime binding on a null reference

And as per my understanding, this error is correct as we are not passing data as a Model.
second,  suppose I am already using strongly type binding on view and along with this I want to use ViewData attribute. Now if I display the ViewData value using @ Model.Property then I will get compile-time error.

I find a similar issue [here](https://github.com/dotnet/AspNetCore.Docs/issues/7257). This issue is closed without any valid answer. I believe this is a wrong explanation on the document. Kindly clarify. 

## Software versions

Asp.Net Core MVC 3.1

---
#### Document Details

* ID: d2e47828-6da8-2323-053f-dcbf7475b1b6
* Version Independent ID: 62d4c2be-4233-64e1-2a3d-8a5e98d9d941
* Content: [Views in ASP.NET Core MVC](https://docs.microsoft.com/en-us/aspnet/core/mvc/views/overview?view=aspnetcore-2.2)
* Content Source: [aspnetcore/mvc/views/overview.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/mvc/views/overview.md)
* Product: **aspnet-core**
* GitHub Login: @ardalis
* Microsoft Alias: **riande**

serpent5:
Fixes #17733.

### DIFF
--- a/aspnetcore/mvc/views/overview.md
+++ b/aspnetcore/mvc/views/overview.md
@@ -266,12 +266,6 @@ public class HomeController : Controller
 }
 ```
 
-In the About view, access the `Title` property as a model property:
-
-```cshtml
-<h1>@Model.Title</h1>
-```
-
 In the layout, the title is read from the ViewData dictionary:
 
 ```cshtml


Rick-Anderson:
@serpent5  thanks for the fix.

PR #17829, opened 2020-04-15
Add some notes accessing controller action descriptor from auth handlers

pranavkm:


serpent5:
Related: #15218.

Of course, that can be closed, but it links to a couple of issues that would be resolved when this one merges.

PR #17861, opened 2020-04-16
[Preview 4] Update Blazor HttpClient coverage

### LINKED ISSUE #17858: [Preview 4] Update HttpClient coverage
https://devblogs.microsoft.com/aspnet/blazor-webassembly-3-2-0-preview-4-release-now-available/

---
#### Document Details

* Content: [Call a web API from ASP.NET Core Blazor](https://docs.microsoft.com/en-us/aspnet/core/blazor/call-web-api?view=aspnetcore-3.1)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #17858

### DIFF
--- a/aspnetcore/blazor/call-web-api.md
+++ b/aspnetcore/blazor/call-web-api.md
@@ -32,7 +32,7 @@ Reference the [System.Net.Http.Json](https://www.nuget.org/packages/System.Net.H
 
 To use a stable API, use the [Microsoft.AspNet.WebApi.Client](https://www.nuget.org/packages/Microsoft.AspNet.WebApi.Client/) package, which uses [Newtonsoft.Json](https://www.nuget.org/packages/Newtonsoft.Json/)/[Json.NET](https://www.newtonsoft.com/json/help/html/Introduction.htm). Using the stable API in `Microsoft.AspNet.WebApi.Client` doesn't provide the JSON helpers described in this topic, which are unique to the `System.Net.Http.Json` package.
 
-## Add HttpClient service
+## Add the HttpClient service
 
 In `Program.Main`, add an `HttpClient` service:
 
--- a/aspnetcore/blazor/call-web-api.md
+++ b/aspnetcore/blazor/call-web-api.md
@@ -5,7 +5,7 @@ description: Learn how to call a web API from a Blazor app using JSON helpers, i
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 01/22/2020
+ms.date: 04/16/2020
 no-loc: [Blazor, SignalR]
 uid: blazor/call-web-api
 ---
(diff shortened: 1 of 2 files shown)


aspnetcore/blazor/call-web-api.md
```diff
@@ -28,9 +28,21 @@ See the following components in the *BlazorWebAssemblySample* sample app:
...
+Reference the [System.Net.Http.Json](https://www.nuget.org/packages/System.Net.Http.Json/) NuGet package in the project file.
 
-To use a stable API, use the [Microsoft.AspNet.WebApi.Client](https://www.nuget.org/packages/Microsoft.AspNet.WebApi.Client/) package, which uses [Newtonsoft.Json](https://www.nuget.org/packages/Newtonsoft.Json/)/[Json.NET](https://www.newtonsoft.com/json/help/html/Introduction.htm). Using the stable API in `Microsoft.AspNet.WebApi.Client` doesn't provide the JSON helpers described in this topic, which are unique to the experimental `Microsoft.AspNetCore.Blazor.HttpClient` package.
+To use a stable API, use the [Microsoft.AspNet.WebApi.Client](https://www.nuget.org/packages/Microsoft.AspNet.WebApi.Client/) package, which uses [Newtonsoft.Json](https://www.nuget.org/packages/Newtonsoft.Json/)/[Json.NET](https://www.newtonsoft.com/json/help/html/Introduction.htm). Using the stable API in `Microsoft.AspNet.WebApi.Client` doesn't provide the JSON helpers described in this topic, which are unique to the `System.Net.Http.Json` package.
```
mkArtakMSFT, line 33: Should we really say this? By the time Blazor goes to GA, this API will also be GA.

@danroth27 thoughts?

guardrex: We can yank it at GA tho ... I think that was the original idea.

mkArtakMSFT: If users are using preview bits, they already are using non-GA bits. So I think that is fine. But would like to leave it to @danroth27 to choose.

danroth27: Blazor Server is shipped and stable, so I guess it's reasonable that a Blazor Server user would prefer to use Microsoft.AspNet.WebApi.Client until System.Net.Http.Json has shipped.

guardrex: Although we orient readers for Blazor Server a bit in the text, we open this topic by saying ...

> The guidance in this topic only pertains to Blazor WebAssembly apps.

... this remark sort'a kind'a is angled to just the WASM case.

guardrex: Related question ... should we go ahead and make the title of this topic ...

> Call a web API from ASP.NET Core Blazor WebAssembly

mkArtakMSFT: I agree with @guardrex and I also read this as a `WASM`-only doc.

danroth27: I guess having this doc be WebAssembly specific makes sense. Calling a web API from a Blazor Server app should be no different than calling an API from any ASP.NET Core app. So, if this doc is WebAssembly specific, I think we can drop any mention of the Microsoft.AspNet.WebApi.Client and just tell people to use System.Net.Http.Json, which is what we recommend.

guardrex: Broader point related to our offline discussion: Blazor is rather difficult to cleanly document the way that we feel compelled to cover it with minimal duplicated content. It has a bit of a split personality disorder:

* Blazor WASM-only content
* Blazor Server and WASM content
* Blazor Server-only content

... with Blazor Server being halfway (or perhaps more than halfway) covered in the ASP.NET Core docs because it's very much is an ASP.NET Core app with Razor components and SignalR.

An alternative to the whole scheme here (and devs have asked for this) is a **_completely separate full coverage for WASM and Server_**. That's all docs in two versions, one for WASM and one for Server ... even if content is duplicated ... and even if the content duplicates ASP.NET Core documentation. That would clear up a lot of reader confusion and angst ... *however* ... at a rather 💰 *high cost* 💰, of course.

mkArtakMSFT: That's a much broader decision and it can't happen now. Let's scope this change down to Blazor WASM.

mkArtakMSFT: Didn't want to appear rude above, @guardrex. I respect your opinion, but that decision is too big to discuss here.

guardrex: I'll make those updates now.

PR #17870, opened 2020-04-17
Blazor WebAssembly doesn't require server-side identity UI

bricelam:
cc @javiercn


bricelam:
Well, I guess it does, but it's added by IdentityServer. (I'm learning)

PR #17899, opened 2020-04-18
Update intro.md

HamidMolareza:
Add additional resource about **EF Core keys**.
Beginners may think that `ID` must be in **capital letters**. (for auto detect the primary key)


Rick-Anderson:
Thanks

PR #17996, opened 2020-04-24
Remove extraneous text in header

captainsafia:
Noticed there was some extraneous text in the header while looking at the docs.


captainsafia:
Definitely aliens.

PR #18005, opened 2020-04-25
add link to Components Lifecycle

behroozbc:
when i read docs about component i what know life cycle of component but i dont find

i dont create issue for this 😅


guardrex:
Thanks @behroozbc, but we don't generally do this because we would end up cross-linking virtually all of the topics to each other. We rely heavily on the sidebar TOC (table of contents) for readers to find other topics.

**_However ..._** Let's do this ... let's move the *Lifecycle* topic **_UP_** in the sidebar TOC ...

Capture

I'll open an issue and take care of that. Thanks for calling this out.

PR #18050, opened 2020-04-29
Fix xref build warnings

scottaddie:


guardrex:
Something went screwy on my end .. I looked right at it and didn't see it on the branch that I was working with. Thanks for the fix.

PR #18085, opened 2020-05-01
fix misspelling

ankona:
Fixes minor misspelling of Brotli compression algorithm name.


Rick-Anderson:
@ankona  thanks for the fix.

PR #18116, opened 2020-05-03
Enabling 32-bit App Pool for 32-bit in-proc

### LINKED ISSUE #18115: Important note
"Deploy a 32-bit app with a 32-bit (x86) .NET Core SDK unless the app..."
For this to work, one has to enable the following: Application Pools - Enable 32-Bit Applications.

I spent a lot of time figuring this out.

---
#### Document Details

* ID: ec484002-18c1-6a0d-00de-820d56f038d3
* Version Independent ID: 7903ce53-2cbb-f083-7479-c8abc30211a0
* Content: [Host ASP.NET Core on Windows with IIS](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/?view=aspnetcore-3.1#feedback)
* Content Source: [aspnetcore/host-and-deploy/iis/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/host-and-deploy/iis/index.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

guardrex:
Fixes #18115

Pre-existing in the topic, we qualify the 64-bit case by saying to disable 32-bit when the app is both SCD and in-proc.

For the 32-bit case ...

> For a 32-bit (x86) [self-contained deployment](/dotnet/core/deploying/#self-contained-deployments-scd) published with a 32-bit SDK that uses the [in-process hosting model](#in-process-hosting-model), enable the Application Pool for 32-bit.

Should the 32-bit guidance only say for in-proc (i.e,. drop the SCD phrase)?


guardrex:
@shirhatti ... I think Justin is unavailable. Can you take a look at this? It's a quick review.

I'm concerned about why we originally said that this only applies to SCD ... I don't recall why that's there.

aspnetcore/host-and-deploy/iis/index.md
```diff
@@ -315,11 +317,15 @@ When deploying apps to servers with [Web Deploy](/iis/install/installing-publish
...
+
+   * For a 64-bit (x64) [self-contained deployment](/dotnet/core/deploying/#self-contained-deployments-scd) that uses the [in-process hosting model](#in-process-hosting-model), disable the app pool for 32-bit (x86) processes. In IIS Manager, navigate to **Application Pools** in the **Connections** sidebar. Select the app's Application Pool. In the **Actions** sidebar, select **Advanced Settings**. Set **Enable 32-Bit Applications** to `False`. 
+
+   The default bitness for new sites can be configured in IIS Manager. In the **Connections** sidebar, select **Application Pools**. In the **Actions** sidebar, select **Set Application Pool Defaults**. Locate **Enable 32-Bit Applications** and set the value to `False`. This setting doesn't affect apps deployed for [out-of-process hosting](xref:host-and-deploy/aspnet-core-module#out-of-process-hosting-model).
```
jkotalik, line 328: I don't see why you need this paragraph.

guardrex: I removed it.

It was there merely to call out that they can set it as a convenience for future sites that they set up, but they'll figure that out if they want that behavior.

jkotalik:
I think we said it only applies to SCD because when you do a framework dependent deployment, the deployed app doesn't have any bitness requirements, meaning it could run in either 32 or 64 bit as long as the x64 and x86 runtimes are installed.

PR #18127, opened 2020-05-04
Blazor WASM config file+other providers updates

### LINKED ISSUE #18088: Clarify wording of default appsettings.json loading.
he File Configuration Provider for app settings files by default:
wwwroot/appsettings.json
wwwroot/appsettings.{ENVIRONMENT}.json
Other configuration providers registered by the app.

When I read this it left just enough uncertainty if it was saying those files would be picked up automatically or if I needed to add them like I would in other project types and they were "supported" in that addjsonfile would work.

Maybe if it said something like "The File Configuration Provider will automatically look for and load appsettings.json or appsettings.{environment}.json located in the wwwroot folder"

see
https://github.com/dotnet/aspnetcore/issues/21391

---
#### Document Details

* ID: a015ab57-3e5c-f5c8-338c-d491c3e9e3fb
* Version Independent ID: 951c8cfc-6aae-2e15-a085-66071d80f6a4
* Content: [ASP.NET Core Blazor hosting model configuration](https://docs.microsoft.com/en-us/aspnet/core/blazor/hosting-model-configuration?view=aspnetcore-3.1#feedback)
* Content Source: [aspnetcore/blazor/hosting-model-configuration.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/hosting-model-configuration.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #18088

Thanks @paul-wade 🎸 and @conficient 🎷!

This fixes an obvious blunder where I had paragraphs at first (`<p>`) for example output then switched to list items (`<li>`) but didn't get all of the tags updated. I was rushing and didn't test. The IDE would have issued an instant 💀 *death threat* 💀 for that. Anyway ...

Although I wasn't asked to show loading a non-default app settings file, 🤔 I think we have to have it. Devs are very likely going to want to have other config files and want them to ✨ *Just Work!*:tm: ✨. Without being able to just toss a File Configuration Provider in there, they'll be stuck and wondering how to do it. Therefore, I tack on a brief example at the end of the section on this PR. Thanks again @conficient ... nice bit there.

Another problem here that's **_unresolved_** on the PR is that we're saying, 'Just go over to the ASP.NET Core *Fundamentals* > *Configuration Provider* topic and use anything you like.' Well ... **_Nope!_** :smile: ... That's just not true. Two options:

* We can remove the mention of using general config providers, at least for the time being. We need a lower-priority follow-up issue to track dealing with this problem later.
* We can actually say which ones work OOB/as-expected and which ones either don't work or work with special configuration.

How shall we proceed?


guardrex:
I think temporarily that we should tone down this idea about sending readers to that topic ... I mean until we can say which ones work OOB with WASM. I'll open an issue to address it further, but I'm going to make it a P2 (Medium) given all of the work I have on my plate for the next few weeks (e.g., security ... let's see .......... *security* ........ and *uh ... oh, yeah!* I almost forgot .... **_SECURITY!_** :smile:).

> see if I can cook something up to extract files from non-wwwroot directorie (if that's possible).

If you come up with anything, I'll have an :eye: out for your issue.

guardrex:
Ok, I opened https://github.com/dotnet/AspNetCore.Docs/issues/18134 to work on it later.

Issue #16467, opened 2020-01-08
Retire duplicate Create a gRPC client and server in ASP.NET Core 3.0 using Visual Studio


Rick-Anderson:
[Sample hosted: Create a gRPC client and server in ASP.NET Core 3.0 using VS only (no VSC or VS Mac)](https://docs.microsoft.com/en-us/samples/aspnet/AspNetCore.Docs/create-grpc-client/)

is a complete duplicate of [Tutorial: Create a gRPC client and server in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/tutorials/grpc/grpc-start?view=aspnetcore-3.1&tabs=visual-studio) (8 K PV/month) with the following exceptions.  The sample hosted version :

* Has instructions that no long work. 
* It's version 3.0
* Has only 325 PV/month

Recommend we retire this version - **OR** remove all the instructions and link to [Tutorial: Create a gRPC client and server in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/tutorials/grpc/grpc-start?view=aspnetcore-3.1&tabs=visual-studio) 

That way folks would have an easy download mechanism and updated instructions.

I'm guessing we can't move the VS, VSC, VSmac version to https://docs.microsoft.com/samples/ because it doesn't support tabs.

@JamesNK  please advise.

JamesNK:
I don't want to move the tutorial to samples.

Whatever you want to do with the samples copy is fine with me.

Rick-Anderson:
@wadepickett can you replace this tutorial with a link to our sample. Update the download  code to the 3.1 version from our repo.

JamesNK:
Don't you mean the opposite: replace the sample content with a link to the tutorial?

Rick-Anderson:
@JamesNK  Thanks, corrected text.

wadepickett:
@Rick-Anderson, will do, thanks!

PR #18169, opened 2020-05-07
Update sample readme for Create gRPC client & server

wadepickett:
For Internal Review: I am not sure how to get a review build for sample readme.md files which lacks structure used by the review build process.

Partially addresses issue #16467

Updated the readme.md to point to the sample tutorial rather than duplicate the tutorial in the readme itself which became quickly out of date.


wadepickett:
@Rick-Anderson, I'm not sure how to get a review build and URI for this sample readme.md.
I would think it would generate something like: https://review.docs.microsoft.com/en-us/aspnet/core/tutorials/grpc/readme?view=aspnetcore-3.1&branch=pr-en-us-18169&tabs=visual-studio

Rick-Anderson:
You can add the readme content to one of your repos like I did [here](https://github.com/Rick-Anderson/Stop-Watch/blob/master/README.md)

I don't see any value to the readme, I think you can delete it. All the info is in a tutorial, right?

wadepickett:
The only value the readme provides, is a way for the sample to be discoverable through the sample browser system.  The readme is required by  the sample system for that to happen.

wadepickett:
For now, I will update the readme so it at least points to the right tutorial until I can get it removed from the samples system.

wadepickett:
@Rick-Anderson, sorry, I forgot I didn't have patches to squash and squashed when merging.  I will try to remember not to do that.

wadepickett:
Fixes #16467

PR #18178, opened 2020-05-07
Fix custom middleware limitations

### LINKED ISSUE #18177: Be clear about custom middleware limitations
The documentation provides a middleware (quoted below). However, it should be noted that it is over-ridden by the contents of a ResponseCacheAttribute on the action or controller.
If the ResponseCacheAttribute doesn't use VaryByHeader, that is understood to mean that for this action, no varying by header is wanted.
The only way to set defaults when using ResponseCacheAttribute is via [cache profiles](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/response?view=aspnetcore-3.1#cache-profiles).
This is partly relevant to the [documentation for ResponseCacheAttribute](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/response), but also relevant here. If someone is trying to set a default ResponseCacheAttribute for a controller and adds another on an action, because only attributes are being dealt with, it'll be more clear that one ResponseCacheAttribute is cancelling another.
Here, though, I was very confused as to why the cache response wasn't varying by header for a bit.
```c#
{
    context.Response.GetTypedHeaders().CacheControl = 
        new Microsoft.Net.Http.Headers.CacheControlHeaderValue()
        {
            Public = true,
            MaxAge = TimeSpan.FromSeconds(10)
        };
    context.Response.Headers[Microsoft.Net.Http.Headers.HeaderNames.Vary] = 
        new string[] { "Accept-Encoding" };

    await next();
});
```

---
#### Document Details

* ID: 10cda0f9-e2c7-015d-4fbf-d361f6c67d9f
* Version Independent ID: 13154131-c2b6-2c72-fb0e-1d83bca4e21a
* Content: [Response Caching Middleware in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/middleware?view=aspnetcore-3.1#feedback)
* Content Source: [aspnetcore/performance/caching/middleware.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/performance/caching/middleware.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

yringler:
Fixes #18177

### DIFF
--- a/aspnetcore/performance/caching/middleware.md
+++ b/aspnetcore/performance/caching/middleware.md
@@ -31,12 +31,14 @@ Configure the app to use the middleware with the <xref:Microsoft.AspNetCore.Buil
 
 [!code-csharp[](middleware/samples/3.x/ResponseCachingMiddleware/Startup.cs?name=snippet2&highlight=16)]
 
-The sample app adds headers to control caching on subsequent requests. Note that this only has affect if there is no [`[ResponseCache]`](xref:Microsoft.AspNetCore.Mvc.ResponseCacheAttribute) in use for the request. If there is, it will overwrite any cache related headers:
+The sample app adds headers to control caching on subsequent requests:
 
 * [Cache-Control](https://tools.ietf.org/html/rfc7234#section-5.2) &ndash; Caches cacheable responses for up to 10 seconds.
 * [Vary](https://tools.ietf.org/html/rfc7231#section-7.1.4) &ndash; Configures the middleware to serve a cached response only if the [Accept-Encoding](https://tools.ietf.org/html/rfc7231#section-5.3.4) header of subsequent requests matches that of the original request.
 
 [!code-csharp[](middleware/samples_snippets/3.x/AddHeaders.cs)]
+
+The preceding middeware is not run when a controller, action, or Razor Page has a [[ResponseCache]](xref:Microsoft.AspNetCore.Mvc.ResponseCacheAttribute) attribute.
 
 Response Caching Middleware only caches server responses that result in a 200 (OK) status code. Any other responses, including [error pages](xref:fundamentals/error-handling), are ignored by the middleware.
 
@@ -171,12 +173,14 @@ Configure the app to use the middleware with the <xref:Microsoft.AspNetCore.Buil
 
 [!code-csharp[](middleware/samples/2.x/ResponseCachingMiddleware/Startup.cs?name=snippet2&highlight=14)]
 
-The sample app adds headers to control caching on subsequent requests. Note that this only has affect if there is no [`[ResponseCache]`](xref:Microsoft.AspNetCore.Mvc.ResponseCacheAttribute) in use for the request. If there is, it will overwrite any cache related headers:
+The sample app adds headers to control caching on subsequent requests:
 
 * [Cache-Control](https://tools.ietf.org/html/rfc7234#section-5.2) &ndash; Caches cacheable responses for up to 10 seconds.
 * [Vary](https://tools.ietf.org/html/rfc7231#section-7.1.4) &ndash; Configures the middleware to serve a cached response only if the [Accept-Encoding](https://tools.ietf.org/html/rfc7231#section-5.3.4) header of subsequent requests matches that of the original request.
 
 [!code-csharp[](middleware/samples_snippets/2.x/AddHeaders.cs)]
+
+The preceding middeware is not run when a controller, action, or Razor Page has a [[ResponseCache]](xref:Microsoft.AspNetCore.Mvc.ResponseCacheAttribute) attribute.
 
 Response Caching Middleware only caches server responses that result in a 200 (OK) status code. Any other responses, including [error pages](xref:fundamentals/error-handling), are ignored by the middleware.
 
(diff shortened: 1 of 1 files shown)


aspnetcore/performance/caching/middleware.md
```diff
@@ -171,7 +171,7 @@ Configure the app to use the middleware with the <xref:Microsoft.AspNetCore.Buil
 
 [!code-csharp[](middleware/samples/2.x/ResponseCachingMiddleware/Startup.cs?name=snippet2&highlight=14)]
 
-The sample app adds headers to control caching on subsequent requests:
+The sample app adds headers to control caching on subsequent requests. Note that this only has affect if there is no [`[ResponseCache]`](xref:Microsoft.AspNetCore.Mvc.ResponseCacheAttribute) in user for the request. If there is, it will overwrite any cache related headers:
```
Rick-Anderson, line 174: Doesn't this apply to 3.x too? You've added this to the 2.x section of the doc.

yringler: Yes, I'm using 3.0. How is version delineated in the markdown file? Should this change be made in two places?

yringler: Ok, I see how that's done.
I haven't tested in version 2, but I would guess that this applies over there also.
So for now I'll copy it to v3 also.

yringler: Done

Rick-Anderson: Correct. Looks like you figured it out.
```
::: moniker-end

```

Rick-Anderson:
@yringler please review my changes. We should probably add something about [Vary](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/response?view=aspnetcore-3.1#vary)

Rick-Anderson:
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/performance/caching/middleware?view=aspnetcore-2.1&branch=pr-en-us-18178)

Rick-Anderson:
@yringler please review my changes.

aspnetcore/performance/caching/middleware.md
```diff
@@ -178,6 +180,10 @@ The sample app adds headers to control caching on subsequent requests:
 
 [!code-csharp[](middleware/samples_snippets/2.x/AddHeaders.cs)]
 
+The preceding headers are not written to the response and overriden when a controller, action, or Razor Page:
+
+* Has a [[ResponseCache]](xref:Microsoft.AspNetCore.Mvc.ResponseCacheAttribute) attribute and the `[ResponseCache]` attribute doesn't contain [VaryByHeader](/aspnet/core/performance/caching/response#vary).
```
yringler, line 185: If the attribute does contain VaryByHeader, it will still override the custom middleware.

I think the point you're trying to get across is that _even if_ it doesn't set VaryByHeader, it will override the vary header. Which is an important point to make, because people (like myself) may expect that if the attribute doesn't set the VaryByHeader property, it won't effect the Vary header.

(I suppose the reason for this behaviour is because a default value of null is understood to mean you don't want a vary header? I would think that a default value means you don't want the attribute to change the vary header... Perhaps an issue should be opened on the aspnetcore repository?)

Maybe change "and the `[ResponseCache]` attribute" to "even if the `[ResponseCache]` attribute"

Rick-Anderson: @yringler Thanks.
Take another look.

Rick-Anderson:
@JunTaoLuo  two minute review please.

aspnetcore/performance/caching/middleware.md
```diff
@@ -178,6 +180,10 @@ The sample app adds headers to control caching on subsequent requests:
 
 [!code-csharp[](middleware/samples_snippets/2.x/AddHeaders.cs)]
 
+The preceding headers are not written to the response and are overriden when a controller, action, or Razor Page:
+
+* Has a [[ResponseCache]](xref:Microsoft.AspNetCore.Mvc.ResponseCacheAttribute) attribute. Setting the `[ResponseCache]` attribute to contain [VaryByHeader](/aspnet/core/performance/caching/response#vary) doesn't change this behavior.
```
JunTaoLuo, line 185: I'm a little confused by what you mean by `doesn't change this behavior.` It seem contradictory.

yringler: I'm hesitant to mix in once this has passed to internal Microsoft review, but in lieu of knowing for sure, I pushed an update to the branch.

Now the bullet point reads
> Has a [[ResponseCache]](xref:Microsoft.AspNetCore.Mvc.ResponseCacheAttribute) attribute. This applies even if a property isn't set. For example, omitting the [VaryByHeader] property (/aspnet/core/performance/caching/response#vary) will cause the corresponding header to be removed from the response.

yringler: Also, this is only in ::: moniker range="< aspnetcore-3.0", but should also be in the moniker range=">= aspnetcore-3.0"

I haven't made the change yet, because it'll just be annoying updating 2 places untill the text passes review.

(Probably that was the reason it was omitted until now.)

Rick-Anderson:
@yringler  thanks for the PR!

yringler:
You're very welcome, thrilled that I was able to help out!
I noticed that we only edited the 2.x section, and forgot to update the 3.x section (`<!-- make same change here once approved -->`).

I opened up a new PR, #18298, which adds the text to the 3.x section (and fixes a spelling mistake)

PR #18187, opened 2020-05-07
Surface handling errors in Blazor web API topic

### LINKED ISSUE #17634: Dealing with other ActionResult values
It would be nice to have information about dealing with Http requests that return values other than the expected serialized object. The default MVC core controller templates return all results in wrapped in an ActionResult object, including "NotFound()" or "BadRequest()" results, but as a beginner looking at this page I have no idea how to deal with any result other than the expected serialized object, and I also don't know where to look to find out more.

---
#### Document Details

* Content: [Call a web API from ASP.NET Core Blazor](https://docs.microsoft.com/en-us/aspnet/core/blazor/call-web-api?view=aspnetcore-3.1#feedback)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #17634

[Internal Review Topic (links to section)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/call-web-api?view=aspnetcore-3.1&branch=pr-en-us-18187#handle-errors)

At this point, the draft language is for discussion. I used a vanilla example ... the exact scenario from a Hosted app with defaults and a bad URI. Would you like to do something else?

### DIFF
--- a/aspnetcore/blazor/call-web-api.md
+++ b/aspnetcore/blazor/call-web-api.md
@@ -5,7 +5,7 @@ description: Learn how to call a web API from a Blazor WebAssembly app using JSO
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 05/04/2020
+ms.date: 05/07/2020
 no-loc: [Blazor, "Identity", "Let's Encrypt", Razor, SignalR]
 uid: blazor/call-web-api
 ---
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/call-web-api.md
```diff
@@ -170,6 +170,36 @@ In the following code, the Delete `<button>` element calls the `DeleteItem` meth
 }
 ```
 
+## Handle errors
+
+When errors occur while interacting with a web API, they can be handled by developer code. For example, `GetFromJsonAsync` expects JSON from the server API, a `Content-Type` of `application/json`. If the response isn't JSON, content validation throws a <xref:System.NotSupportedException>.
```
mkArtakMSFT, line 175:
```suggestion
When errors occur while interacting with a web API, they can be handled by developer code. For example, `GetFromJsonAsync` expects JSON response from the server API, with a `Content-Type` of `application/json`. If the response isn't in JSON format, content validation throws a <xref:System.NotSupportedException>.
```

aspnetcore/blazor/call-web-api.md
```diff
@@ -170,6 +170,36 @@ In the following code, the Delete `<button>` element calls the `DeleteItem` meth
...
+
+In the following example, the URI endpoint for the weather forecast data request is misspelled. The URI should be to `WeatherForecast` but appears in the call as `WeatherForcast` (missing "e"). The call expects JSON to be returned, but the server returns HTML for an unhandled exception on the server with a `Content-Type` of `text/html`. The call throws a <xref:System.NotSupportedException> when the response content is validated as non-JSON. The exception is caught in the `catch` block, where it could be logged or result in a friendly error message to the user.
```
mkArtakMSFT, line 177: I think it's worth to add a light mention about why html will be returned (server will not find path, will redirect to response page middleware, which will try to respond with an html from that page).

aspnetcore/blazor/call-web-api.md
```diff
@@ -170,6 +170,36 @@ In the following code, the Delete `<button>` element calls the `DeleteItem` meth
...
+
+```csharp
+protected override async Task OnInitializedAsync()
+{
+    try
+    {
+        forecasts = await Http.GetFromJsonAsync<WeatherForecast[]>(
+            "WeatherForcast");
+    }
+    catch (AccessTokenNotAvailableException exception)
```
mkArtakMSFT, line 187: I think the first catch block is irrelevant in this particular example and should be removed.

aspnetcore/blazor/call-web-api.md
```diff
@@ -170,6 +170,36 @@ In the following code, the Delete `<button>` element calls the `DeleteItem` meth
...
+    {
+        exception.Redirect();
+    }
+    catch (NotSupportedException exception)
+    {
+        ...
```
mkArtakMSFT, line 193:
```suggestion
        // custom logic goes here for handling the exception
```

guardrex: We can't put text in due to localization, but I can call that out in the text itself. I'll do that on the next commit.

PR #18216, opened 2020-05-08
fixed link and made button title same as on page

dzmitry-lahoda:


Rick-Anderson:
@dzmitry-lahoda  thanks!

PR #18129, opened 2020-05-04
Rename Blazor WASM file extensions

guardrex:
Fixes #18087

[Internal Review Topic (links to section)](https://review.docs.microsoft.com/en-us/aspnet/core/host-and-deploy/blazor/webassembly?view=aspnetcore-3.1&branch=pr-en-us-18129#rename-dll-files-to-bypass-proxies-firewalls-and-antivirus-software)

Thanks @stavroskasidis for your tip at https://github.com/dotnet/aspnetcore/issues/5477#issuecomment-623167309. :rocket:

PR #18241, opened 2020-05-11
Blazor ext in service worker assets manifest

### LINKED ISSUE #18236: Blazor extensions in service worker assets manifest
https://github.com/dotnet/aspnetcore/issues/21489#issuecomment-626662536

---
#### Document Details

* ID: 89b4d6a1-1436-7ba8-a5c5-8dcb6a955bb0
* Version Independent ID: 4a6a13b4-fe5c-a9fb-5d7d-33d5aa0fbac3
* Content: [Host and deploy ASP.NET Core Blazor WebAssembly](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/blazor/webassembly?view=aspnetcore-3.1#change-the-filename-extension-of-dll-files)
* Content Source: [aspnetcore/host-and-deploy/blazor/webassembly.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/host-and-deploy/blazor/webassembly.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #18236

### DIFF
--- a/aspnetcore/host-and-deploy/blazor/webassembly.md
+++ b/aspnetcore/host-and-deploy/blazor/webassembly.md
@@ -5,7 +5,7 @@ description: Learn how to host and deploy a Blazor app using ASP.NET Core, Conte
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 05/07/2020
+ms.date: 05/11/2020
 no-loc: [Blazor, "Identity", "Let's Encrypt", Razor, SignalR]
 uid: host-and-deploy/blazor/webassembly
 ---
@@ -416,11 +416,23 @@ dir .\_framework\_bin | rename-item -NewName { $_.name -replace ".dll\b",".bin" 
 ((Get-Content .\_framework\blazor.boot.json -Raw) -replace '.dll"','.bin"') | Set-Content .\_framework\blazor.boot.json
 ```
 
+If service worker assets are also in use on Windows, add the following command:
+
+```powershell
+((Get-Content .\service-worker-assets.js -Raw) -replace '.dll"','.bin"') | Set-Content .\service-worker-assets.js
+```
+
 On Linux or macOS:
 
 ```console
 for f in _framework/_bin/*; do mv "$f" "`echo $f | sed -e 's/\.dll\b/.bin/g'`"; done
 sed -i 's/\.dll"/.bin"/g' _framework/blazor.boot.json
+```
+
+If service worker assets are also in use on Linux or macOS, add the following command:
+
+```console
+sed -i 's/\.dll"/.bin"/g' service-worker-assets.js
 ```
    
 To use a different file extension than *.bin*, replace *.bin* in the preceding commands.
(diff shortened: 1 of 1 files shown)


aspnetcore/host-and-deploy/blazor/webassembly.md
```diff
@@ -416,12 +416,24 @@ dir .\_framework\_bin | rename-item -NewName { $_.name -replace ".dll\b",".bin"
 ((Get-Content .\_framework\blazor.boot.json -Raw) -replace '.dll"','.bin"') | Set-Content .\_framework\blazor.boot.json
 ```
 
+If service worker assets are also in use on Windows, add the following command:
```
mkArtakMSFT, line 419:
```suggestion
If service worker assets are also in use, add the following command:

On Windows:
```

mkArtakMSFT, line 419: This `on Windows` confused me. Wondering whether we should have `On Windows` as a subsection - so all the content is visually under it.

```suggestion
If service worker assets are also in use, add the following command:
```

aspnetcore/host-and-deploy/blazor/webassembly.md
```diff
@@ -416,12 +416,24 @@ dir .\_framework\_bin | rename-item -NewName { $_.name -replace ".dll\b",".bin"
...
+```
+
 On Linux or macOS:
 
 ```console
 for f in _framework/_bin/*; do mv "$f" "`echo $f | sed -e 's/\.dll\b/.bin/g'`"; done
 sed -i 's/\.dll"/.bin"/g' _framework/blazor.boot.json
 ```
+
+If service worker assets are also in use on Linux or macOS, add the following command:
```
mkArtakMSFT, line 432:
```suggestion
If service worker assets are also in use, add the following command:
```

guardrex:
> Can you also add a small paragraph there, indicating why this may be needed? 

The language came from Steve when we wrote this section.

The initial paragraph that I proposed was ...

> Some environments block the download of *.dll* files that are normally part of a Blazor app's published deployment. To address this issue, you can rename *.dll* files to use a different file extension and update the *blazor.boot.json* file.

... but Steve said on review that ...

> Rather than phrase it like this, I think we should use phrasing more like the following:
>
> > Changing the filename extension of .dll files
> >
> > In case you have any reason to change the filename extensions for the .dll files in your application, you can follow the procedure below.
>
> That is, I don't think we should imply this is something people should do "just in case". We only have a handful of confirmed cases where people had reason to do this, so we don't want to falsely give the impression that everyone should be doing it. I definitely don't think we should stir up fear about it by talking about firewalls and antivirus.
>
> If people think they have reasons to do this, I'd like them to be unbiased and tell us their reasons, rather than being prejudiced by suggestions about why to do it.

From: https://github.com/dotnet/AspNetCore.Docs/pull/18129#discussion_r419677647

In particular, his remark ...

**_I definitely don't think we should stir up fear about it by talking about firewalls and antivirus._**

... led me to go with something close to his suggestion.

PR #18267, opened 2020-05-12
Add helpful tip to novices

vthg2themax:
Add a helpful tip to novices just starting out.


guardrex:
Thanks @vthg2themax ... I'll take a look within 24 hours. I'm buried in code right now. *Heeelp!* :smile:

guardrex:
@vthg2themax ... It's best to open this for the product unit. It sounds more like a regression than something that we should document because the default profile in *launchSettings.json* lists the HTTPS endpoint. Open an issue for engineering at ...

PR #18296, opened 2020-05-14
Cleanup SignalR Windows Auth with browsers

BrennanConroy:
cc @davidfowl 

Maybe we should check all the browsers listed in https://docs.microsoft.com/en-us/aspnet/core/signalr/supported-platforms?view=aspnetcore-3.1#javascript-client and put their status here?

### DIFF
--- a/aspnetcore/signalr/authn-and-authz.md
+++ b/aspnetcore/signalr/authn-and-authz.md
@@ -153,7 +153,7 @@ var connection = new HubConnectionBuilder()
     .Build();
 ```
 
-Windows Authentication is only supported by the browser client when using Microsoft Internet Explorer or Microsoft Edge.
+Windows Authentication is not supported in all browsers. And in some browsers, such as Chrome and Safari, WebSockets with Windows Authentication doesn't work and the client will try to fallback to other transports.
 
 ### Use claims to customize identity handling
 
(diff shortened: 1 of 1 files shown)


davidfowl:
Yep

Rick-Anderson:
@BrennanConroy is that better or worse?

aspnetcore/signalr/authn-and-authz.md
```diff
@@ -153,7 +153,10 @@ var connection = new HubConnectionBuilder()
     .Build();
 ```
 
-Windows Authentication is only supported by the browser client when using Microsoft Internet Explorer or Microsoft Edge.
+Windows authentication is supported in Internet Explorer and Microsoft Edge, but not in all browsers. For example, in Chrome and Safari, attempting to use Windows authentication and WebSockets:
+
+* Fails.
```
BrennanConroy, line 158: Maybe it would be better as a sentence than a list. "Windows auth with WebSockets fails and the client will then attempt to fall back to other transports which might work better with Windows auth." 🤷‍♀️

Rick-Anderson: @BrennanConroy  how about now?

BrennanConroy: Yeah looks good! Thanks :)

PR #18298, opened 2020-05-14
add to 3.x section , fix spelling

yringler:
Completes #18177.

We forgot to add the text to the netcore 3.x section.
This PR updates that section, and fixes a spelling mistake.

PR #18299, opened 2020-05-14
fix markdown url

yringler:
Related to #18298 
Some markdown was fixed in one version, but not the other.
This updates the other one.
(`[VaryByHeader](/aspnet/core/performance/caching/response#vary) property`)


Rick-Anderson:
Thanks

PR #18317, opened 2020-05-15
Simplify the using statement

IliyanAng:
Given this example is for .net core 3.x and C# 8 is supported, we're taking advantage of the simplified version of the using statement.


IliyanAng:
Ah shucks, I made a typo in the commit subject line and content.

PR #18369, opened 2020-05-18
prevent over-localization in link text

JasonCard:
The NuGet package name within link text is getting over-localized, so suggest to make the link text as inline code to prevent this.


guardrex:
It appears about seven more times across the docs. I'll put in a quick PR after this to patch those up as well.

Issue #16341, opened 2019-12-28
Integration tests MASTER issue


Rick-Anderson:
.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 77889d43-fedb-f8b4-8fea-379ec946555c
* Version Independent ID: f499b131-8b14-7e10-1728-543fdd5a5656
* Content: [Integration tests in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-3.0#customize-webapplicationfactory)
* Content Source: [aspnetcore/test/integration-tests.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/test/integration-tests.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-test**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
There's a project for these (most on the project are for Integration Testing) at https://github.com/aspnet/AspNetCore.Docs/projects/31. Would you prefer that I add the rest of the issues to this issue and close the project and remaining issues?

Rick-Anderson:
We need an issue to track issues. It's nice to have them all in one place.

guardrex:
That's the purpose of a GH project, and that's where they've all been kept. Do you want me to add the others here and close those issues?

Rick-Anderson:
The PU is not closing issues and adding them to projects. I'm tracking by issues. Project is fine too, but we need this master issue.

guardrex:
The project was for Javier.

Unfortunately, this project never had a regular triage. It takes *a long time* 🙈 for them to get triaged due to engineer workload (*I guess*).

I'll move the rest, close those issues, and shut that project down. Javier can look at the list here when he wants to see what's what. I doubt that he cares, but we'll find out when he sees this at some point.

Rick-Anderson:
@guardrex  BTW, the master issue approach is working pretty well. I'm reviewing them all now. If the PU wants projects, use them too.

michal-ciechan:
For anyone struggling with this after using these Docs, if you use `builder.ConfigureTestServices(Action<IServiceProvider> services)` extension method, they will run after the WebHost  `Startup.ConfigureServices` file is run!.

e.g.

```csharp
public class CustomWebApplicationFactory<TStartup>
[...]

PR #18391, opened 2020-05-19
Adding ConfigureTestServices

### LINKED ISSUE #18286: Customize WebApplicationFactory clarification on ConfigureServices
"The test app's builder.ConfigureServices callback is executed after the app's Startup.ConfigureServices code is executed. The execution order is a breaking change for the Generic Host with the release of ASP.NET Core 3.0. To use a different database for the tests than the app's database, the app's database context must be replaced in builder.ConfigureServices"

This is valid ONLY if the host is an IHost using IHostBuilder. 
For IWebHost ConfigureServices in ConfigureWebHost will run BEFORE the Startups ConfigureServices, while using IHost it will run AFTER. 
For IWebHost usage a workaround is to instead use builder.ConfigureTestServices. 

I would like the docs to clarify this. When reading the above I assumed that the behaviour would be the same for IWebHost.

For me migrating from ASP.CORE 2.2 to 3.1, and still using IWebHost, following the documentation broke the tests (I'd removed a routines specific for 2.2 for performing configuration after Startup was run). 
This was not mentioned in the migration guide either. 

---
#### Document Details

* Content: [Integration tests in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-3.1#customize-webapplicationfactory)
* Content Source: [aspnetcore/test/integration-tests.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/test/integration-tests.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

michal-ciechan:
Fixes #18286 

This Issue is now part of #16341

`builder.ConfigureServices` (At least in AspNetCore 3.1) no longer is run after the `Startup.ConfigureServices`. Updating the documentation to reflect this, and mentioning of `ConfigureTestServices` extension method.


aspnetcore/test/integration-tests.md
```diff
@@ -515,7 +515,7 @@ By default, non-essential cookies aren't preserved across requests when the [GDP
...
 
-1. Inherit from `WebApplicationFactory` and override [ConfigureWebHost](/dotnet/api/microsoft.aspnetcore.mvc.testing.webapplicationfactory-1.configurewebhost). The [IWebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.iwebhostbuilder) allows the configuration of the service collection with [ConfigureServices](/dotnet/api/microsoft.aspnetcore.hosting.istartup.configureservices):
+1. Inherit from `WebApplicationFactory` and override [ConfigureWebHost](/dotnet/api/microsoft.aspnetcore.mvc.testing.webapplicationfactory-1.configurewebhost). The [IWebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.iwebhostbuilder) allows the configuration of the service collection with [ConfigureServices](/dotnet/api/microsoft.aspnetcore.hosting.istartup.configureservices) which is executed before the app's `Startup.ConfigureServices`. The [IWebHostBuilder](/dotnet/api/microsoft.aspnetcore.hosting.iwebhostbuilder) allows overriding and modiying registered services in the service collection with [ConfigureTestServices](/dotnet/api/microsoft.aspnetcore.testhost.webhostbuilderextensions.configuretestservices):
```
michal-ciechan, line 523: I guess this is correct considering the example and context currently uses `Web Host` rather than `Generic Host`?

michal-ciechan: I can see the sample at [CustomWebApplicationFactory](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/test/integration-tests/samples/3.x/IntegrationTestsSample/tests/RazorPagesProject.Tests/CustomWebApplicationFactory.cs) uses `Generic Host` builder in the SUT, but `ConfigureWebHost` in the test. Is that correct as I can see the `ConfigureHost(IHostBuilder builder)` virtual method in `WebApplicationFactory` which I would assume is for `Generic Host`

guardrex: Yes, this only appears in the 2.x part of the doc, so this should be ok.

guardrex: ... and the sample code below this is 2.x.

If you're saying that the 2.x code is using `ConfigureHost` (or vice-versa the 3.x code is using `ConfigureWebHost`), then there might be a problem here. AFAIK, this is all ok given that ...

* IIRC, we have proper sign-offs from engineering on this content.
* There haven't been any comments coming in that the code examples/text are wrong between 2.x and 3.x.

I'm 🏃😅 right now on high priority issues, so I'd prefer not to look unless you think there is a problem. If so, let me know ... but I'll need to schedule this for next week.

michal-ciechan: Sorry for my ignorance and this stupid question, but could you expand on "this only appear in the 2.x part of the doc"?

Reason I am asking is because I can see `ConfigureWebHost` in the 3.x samples, and in the doc @docs.microsoft.com I have ASP.NET Core 3.1 selected.

I think I am missing something.

Either way, if you say it's ok I'll take your word for it 👍😃

guardrex: The way these docs work with the version selector on the live site is that there are these markdown lines that set which text is shown. You can see them sprinkled around in the docs ...

```

3.x content here

::: moniker-end

2.x content here

::: moniker-end
```

When a reader selects the version on the live site, they see the text in the matching version block ("moniker range").

... and sample cross-links in there (usually) have a `2.x` segment for 2.x and `3.x` for 3.x.

As long as the code is right for that version of the text, it should be ok. We occasionally goof up and show 3.x code in 2.x text or vice-versa, but the community is pretty quick on issuing a good 💀 *death threat* 💀 if we don't fix it. :smile:

michal-ciechan: Awesome thanks for the clarification! Is it worth adding this info to the Contribution Guidelines linked form the root README? (I am assuming the same syntax/logic is used across all docs in this repo?

guardrex: It applies to all of Microsoft docs everywhere. I'm not sure how they've decided to handle it company-wide/.NET Foundation-wide.

I've asked about adding *a lot* of content to the local guidelines (e.g., https://github.com/dotnet/AspNetCore.Docs/issues/2617), but there hasn't been time and/or consensus among the team to do so.

PR #18455, opened 2020-05-24
Changed wording

David-Okeke:
Changed the instruction "Set **Two pane view**" to "Toggle **Two pane view**"


Rick-Anderson:
Toggle won't work if it's currently on 2 pane view. Thanks for the suggestion.

PR #18473, opened 2020-05-25
Update model.md

Takahiro901:
[environment]
Visual Studio for Mac
ASP.NET Core 3.1.2
  
When I studied this tutorial, I saw the view like below screenshot. This is different from the screenshot in the document. 
  
image
  
In my opinion, this procedure is required when Visual Studio for Mac is used.


Rick-Anderson:
That's not correct.

PR #18485, opened 2020-05-26
Update get-started.md

jongio:
- I had an older version of the .NET CLI 3.1.x and it couldn't find the wasm template. I upgraded to 3.1.300 and it worked.
- Added instructions for enabling the debug preview in VS Code.


guardrex:
Hello @jongio ...

> Open VS Code Settings (File -> Preferences -> Settings). Search for "debug javascript use preview" and check the "Use the new in-preview JavaScript debugger for Node.js and Chrome" checkbox.

I don't see that setting. Did that just come down with 1.45.1? If not, could you post a screenshot here?

PR #18501, opened 2020-05-27
(Blazor Wasm) Fixes duplication of sections in security docs, broken links and typos

chrissainty:
The Name and role claim with API authorization and Profile Service sections on the Blazor WebAssembly > hosted with identity server security docs were duplicated. This PR removes that duplication. 

It also fixes a couple of broken links to the `Authorize` attribute docs and a small typo.


guardrex:
Thanks @chrissainty ...

* That dup of content likely occurred when Javier fixed the factory for the various values that can come down in the `role` claim. I obviously didn't build locally, and I can't see the internal build. My bad. Thanks for fixing it.
* I don't care for the original layout: It's too difficult to grok the two Server app approaches for establishing the role claim (API auth options or Profile Service). I made some additional updates here that I hope clarifies it.
* WRT the link: Yesterday, there was a massive (500 line) update PR for API doc crosslinks. This one was a victim of large-scale cut-'n-paste and snuck past eyeballing the PR diff. I don't want to fix it like this tho: The goal of the links is to link to both the topic and the API docs. All that it needs is for a bracket to be removed.
* WRT "username"/"user name": @scottaddie The manual is using "username" (one word), but we're using "user name" (two words) everywhere (AFAICT from a quick search). Did the guidance change at some point?

guardrex:
Given that the doc is in rough shape at the moment, I'm going to 🚓 **_Emergency Merge_**:tm: 🚒 this now. I'll :eyes: **_LOOK_** :eyes: this time at the live topic.

@scottaddie, let me know on the "username" vs. "user name" situation. I'm about to kill the en dash layout elements in favor of *ye olde colon-space syntax*, and that's a good patch PR to update our use of "user name" to "username" if that's what we're going to do. :ear:

chrissainty:
@guardrex That all makes sense, I agree, the name `CustomUserFactory` is a better choice and fits in with existing examples. 

I agree about the layout, the reason I spotted these issues was because i was brushing up on the newer approached for Blazor Wasm; it took me a few passes to realise that API auth options and profile service were alternate approaches for handling role claims.

PR #18507, opened 2020-05-27
MIME types for Blazor Azure Storage deployments

### LINKED ISSUE #18483: Standalone deployment with Azure Storage: problems with mime types and compression
Hi,

When deploying my Blazor WebAssembly app on Azure Storage a lot of mime types are not correct. For example I had to change the dotnet.wasm to 'application/wasm' to get it working. also for the .js files I also had to change the mime type to 'application/javascript'. 

Additionally nothing was compressed because the content encoding was also missing on the dll files so my application was huge. I changed it by hand the first time which was very cumbersome and later I found a script.

This issue is nowhere mentioned in the documentation. I don't know if this should be mentioned here or in the 'Static website hosting in Azure Storage' Docs but I think this should be mentioned somewhere.

---
#### Document Details

* ID: 89b4d6a1-1436-7ba8-a5c5-8dcb6a955bb0
* Version Independent ID: 4a6a13b4-fe5c-a9fb-5d7d-33d5aa0fbac3
* Content: [Host and deploy ASP.NET Core Blazor WebAssembly](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/blazor/webassembly?view=aspnetcore-3.1)
* Content Source: [aspnetcore/host-and-deploy/blazor/webassembly.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/host-and-deploy/blazor/webassembly.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #18483

[Internal Review Topic (links to *Azure Storage* section)](https://review.docs.microsoft.com/en-us/aspnet/core/host-and-deploy/blazor/webassembly?view=aspnetcore-3.1&branch=pr-en-us-18507#azure-storage)

Thanks @daviddejaeger! :rocket:


guardrex:
@mkArtakMSFT I'm a bit suspicious about this. My understanding is that the `Content-Type` for compressed content should be the `Content-Type` of the original file, and that's what the browser will see when the file is decompressed before WASM kicks in.

In the uncompressed case, @daviddejaeger indicated that the site fails if the MIME types aren't correct. Therefore, I made the content on this PR address setting the MIME type (`Content-Type`) correctly either before or after deployment. The best way being to configure tooling based on the file's extension.

In the compressed case, the app seems to run ok out of Azure Storage (this scenario I tested myself with Gzip). However, the MIME types are **not** correct for the files in the compressed case. For example, *dotnet.wasm.gz* isn't `application/wasm` after decompression.

Soooo ... why is this a problem in the uncompressed case if in the compressed case they're wrong and the app still runs. I mean the browser decompresses a compressed *dotnet.wasm.gz*, for example, into *dotnet.wasm* with the wrong MIME type and the app runs.

I try a repro of what @daviddejaeger reported for the uncompressed case.

guardrex:
@mkArtakMSFT I ran an uncompressed test here. The only file that caused a problem was the weather data JSON file, and I could simply set the `Content-Type` header in Cloudberry's tool (now called MP360). Since the error message is obvious when the site doesn't run, I think the content here can be minimized. I'll do it on the next commit and merge.

PR #18518, opened 2020-05-27
Blazor WASM auth config binding

### LINKED ISSUE #18516: Blazor webassembly configuration binding update in version 3.2.0
The instructions under Program.cs no longer work in the latest build; the webassembly configuration binding system has been changed significantly 

                builder.Configuration.Bind("AzureAdB2C", options.ProviderOptions.Authentication);

Useful related links:

* https://github.com/dotnet/aspnetcore/issues/20231

* https://github.com/dotnet/aspnetcore/pull/20647

Also: https://stackoverflow.com/q/62052986/12900015
---
#### Document Details

* ID: 7d682d4c-86b5-c47e-89a3-3684b2235320
* Version Independent ID: a80e115c-466a-288d-a918-9ef7514f5add
* Content: [Secure an ASP.NET Core Blazor WebAssembly hosted app with Azure Active Directory B2C](https://docs.microsoft.com/en-us/aspnet/core/security/blazor/webassembly/hosted-with-azure-active-directory-b2c?view=aspnetcore-3.1)
* Content Source: [aspnetcore/security/blazor/webassembly/hosted-with-azure-active-directory-b2c.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/security/blazor/webassembly/hosted-with-azure-active-directory-b2c.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #18516

Per the update at https://docs.microsoft.com/en-us/aspnet/core/blazor/hosting-model-configuration?view=aspnetcore-3.1#authentication-configuration.

btw @captainsafia, let's change the OIDC example in the configuration content. Javier told a dev recently not to take that route following the *Standalone with Auth Lib* topic but instead to go with the dedicated Azure AD topic and MSAL.

Thanks @RedBlue1776! :rocket:

### DIFF
--- a/aspnetcore/blazor/hosting-model-configuration.md
+++ b/aspnetcore/blazor/hosting-model-configuration.md
@@ -213,9 +213,9 @@ builder.Configuration.AddJsonStream(stream);
 
 ```json
 {
-  "AzureAD": {
-    "Authority": "https://login.microsoftonline.com/",
-    "ClientId": "aeaebf0f-d416-4d92-a08f-e1d5b51fc494"
+  "Local": {
+    "Authority": "{AUTHORITY}",
+    "ClientId": "{CLIENT ID}"
   }
 }
 ```
@@ -224,7 +224,7 @@ builder.Configuration.AddJsonStream(stream);
 
 ```csharp
 builder.Services.AddOidcAuthentication(options =>
-    builder.Configuration.Bind("AzureAD", options);
+    builder.Configuration.Bind("Local", options);
 ```
 
 #### Logging configuration
--- a/aspnetcore/security/blazor/webassembly/azure-active-directory-groups-and-roles.md
+++ b/aspnetcore/security/blazor/webassembly/azure-active-directory-groups-and-roles.md
@@ -110,8 +110,7 @@ Register the factory in `Program.Main` (*Program.cs*) of the standalone app or C
 builder.Services.AddMsalAuthentication<RemoteAuthenticationState, 
     CustomUserAccount>(options =>
 {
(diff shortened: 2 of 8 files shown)


captainsafia:
@javiercn Did something change here? I though the authentication values passed in the mapped to `options.ProviderOptions.Authentication` not `options` directly?

guardrex:
Actually, this doesn't seem to work in practice. My test apps ran with the current content for RC and GA, but this seems to break things.

guardrex:
When we did the config updates tho, we put in ...

```csharp
builder.Services.AddOidcAuthentication(options =>
    builder.Configuration.Bind("AzureAD", options);
```

... which I need to change to a `Local` example ... I'll take care of that tonight. **_BUT_** ...

It seems now like that wouldn't be right. We have in the *Standalone with Auth Lib* topic with the following pattern ...

```csharp
builder.Services.AddOidcAuthentication(options =>
{
    builder.Configuration.Bind("Local", options.ProviderOptions);
});
```

guardrex:
Yeah ... this isn't right @RedBlue1776 ... it doesn't run in practice, nor does it run without the `builder.Configuration.Bind` line. We can discuss further tho here or back on the issue if there are changes.

One thing that I do want to do is take the config example from an Azure AD approach with the Auth Lib to a "Local" config entry approach, since we want devs to use the Azure AD topic and MSAL for Azure AD.

guardrex:
btw ...

> The Bind method is no longer available in version 3.2

... add the namespace ...

```csharp
using Microsoft.Extensions.Configuration;
```

PR #18519, opened 2020-05-27
Patch Blazor config example

guardrex:
Addresses #18516

Let's make this example based on "Local" and a placeholder for authority and client id. We don't want to suggest Azure AD with `AddOidcAuthentication` here.

Also, check me on the statement ... `options.ProviderOptions`, not just `options`. That matches the *Standalone with Auth Lib* topic at ...

https://docs.microsoft.com/en-us/aspnet/core/security/blazor/webassembly/standalone-with-authentication-library?view=aspnetcore-3.1#authentication-service-support

... where we show ...

```csharp
builder.Services.AddOidcAuthentication(options =>
{
    builder.Configuration.Bind("Local", options.ProviderOptions);
});
```

I never tested this scenario because I don't have a non-Azure, non-Identity Server OIDC provider readily available.

This seems correct [a test app loaded config correctly (`@Config["Local:ClientId"]`)], so I'm going to go ahead and merge it. Hit me up :ear: if I need to patch it.

### DIFF
--- a/aspnetcore/blazor/hosting-model-configuration.md
+++ b/aspnetcore/blazor/hosting-model-configuration.md
@@ -5,7 +5,7 @@ description: Learn about Blazor hosting model configuration, including how to in
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 05/19/2020
+ms.date: 05/27/2020
 no-loc: [Blazor, "Identity", "Let's Encrypt", Razor, SignalR]
 uid: blazor/hosting-model-configuration
 ---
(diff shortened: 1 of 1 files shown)


captainsafia:
That looks correct.

The [Program.cs template](https://github.com/dotnet/aspnetcore/blob/9b98da360a3051a4d75a716048960840fb1b9b6d/src/ProjectTemplates/ComponentsWebAssembly.ProjectTemplates/content/ComponentsWebAssembly-CSharp/Client/Program.cs) is a good source of truth on this since I validated it against all auth providers.

PR #18533, opened 2020-05-28
Updated security > data-protection > configuration > overview.md to e…

markmcgookin:
…xplain when using blob storage to store keys and azure keyvault to protect them. They key store blob xml file must already exist before running. This isn't clear from any of the other examples or documentation.

Fixes issue 10299 from the main aspcore repo (not adding # as the issue isn't with this documentation, but a change to the documentation could solve the issue from occurring in the first place)


HaoK:
Barry beat me to it, I was going to say there are new packages coming too :)

pakrym:
New packages are going to be out GA at the end of this week so I would advise to include references to them instead.

scottaddie:
@pakrym Can you please provide the package names?

pakrym:
Azure.Extensions.AspNetCore.DataProtection.Blobs and Azure.Extensions.AspNetCore.DataProtection.Keys

scottaddie:
@markmcgookin Can you please update the PR to use these 2 packages?

markmcgookin:
> @markmcgookin Can you please update the PR to use these 2 packages?

I might be missing something here @pakrym .... I created a prototype webapi that shows the original issue (you need to run the application without the keyvault config line once to create the xml document) or else you get a URI exception. If I add those new packages to the solution, I can still use the same code, and see the same error. Do I have to use some specific overload of the constructor to take advantage of these new extensions? How are they supposed to solve the issue?

Rick-Anderson:
@pakrym  quick question please.

pakrym:
I think this PR is fine. Changing the doc to point to new packages is a bigger and slightly unrelated issue.

PR #18571, opened 2020-05-29
Add .dat to sample web.config

danroth27:
Blazor WebAssembly apps have a .dat file for the timezone data, but it's missing from the sample web.config.


guardrex:
I'll take this live now.

Issue #16287, opened 2019-12-20
Lifecycle diagram and coverage


LeeW32:
[EDIT by guardrex to expand issue based on the engineering docs issue] 

Might be good to include a 'Component Lifecycle' diagram like the Vue Team have done. I've always found the Microsoft docs in general lacking in visual cues. This would be a perfect place to start!

Also address: [Add or improve docs for lifecycle methods (dotnet/aspnetcore #13473)](https://github.com/dotnet/aspnetcore/issues/13473)

**From the original [Improve Blazor component lifecycle page (aspnet/AspNetCore.Docs #14353)](https://github.com/aspnet/AspNetCore.Docs/issues/14353) issue (_closed now_)** ...

Expand content to include:

* A detailed description of what happens in what order, including:
    * How async tasks affect this
    * When (re)rendering is interleaved into this
    * How things are ordered among parent/child components
    * What kinds of work you're expected to do in each of the methods
 * Relationship with `Dispose`
    * Example: Pattern for cancelling background work
 * Relationship with JS interop
    * Example: Initializing a JS library by passing an `ElementReference`

### From DR on the PR to create the initial *Lifecycle* topic

Instead of focusing purely on the lifecycle events I think we want this article to explain more broadly how components work, like how and when the render. We should walk the user through a day in the life a component in a Blazor app. The lifecycle events are then just specific points that you can plugin your specific behaviors. 

Things that I think should be covered:
- What happens during the initial render of the app
  - We probably need to address prerendering as part of this
- When components get instantiated
- When services get injected from DI
- What happens when a parent/root component gets rendered
- When do components render after the initial render
  - In response to UI events and explicit calls to StateHasChanged
- How long do components life for? When do they get disposed?

WRT the short *State changes* (`StateHasChanged`) section ...

I think this section would fit in better with a broader description of how and when components get rendered. Right now this section on StateHasChanged feels a bit out of context.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 4bae853d-80e3-ef81-69eb-e8276eb853fe
* Version Independent ID: 01216f14-b45c-73b0-9f1f-300e7ddd788f
* Content: [ASP.NET Core Blazor lifecycle](https://docs.microsoft.com/en-us/aspnet/core/blazor/lifecycle?view=aspnetcore-3.1#feedback)
[...]

PR #18610, opened 2020-06-01
Blazor lifecycle ordering clarifications

guardrex:
Addresses #16287

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/lifecycle?view=aspnetcore-3.1&branch=pr-en-us-18610)

I clarify lifecycle method ordering in two ways:

* I re-order the sections by *lifecycle method calling order*. To do so, I move the *Before parameters are set* section above the *Component initialization methods* section.
* I reword the explanations on when methods are called to clarify the order in each section.

Do you still want to produce a *lifecycle diagram* (a graphic) with additional goodness and/or address issue asks in other ways, e.g., "broadly how components work"?

If this is as far as you want to go right now, then I'll merge and close the issue. Otherwise, I'll leave that issue as a P1 assigned to you.

### DIFF
--- a/aspnetcore/blazor/lifecycle.md
+++ b/aspnetcore/blazor/lifecycle.md
@@ -5,7 +5,7 @@ description: Learn how to use Razor component lifecycle methods in ASP.NET Core 
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 05/07/2020
+ms.date: 06/01/2020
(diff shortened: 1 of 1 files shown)


guardrex:
Ping @danroth27

guardrex:
@mkArtakMSFT It's best if we put this in now. We can address the remainder of the issue if there's more work to do later.

Issue #18647, opened 2020-06-03
HUB pages needs core


Rick-Anderson:
ASP.NET Core is always ASP.NET Core and never ASP.NET without Core.

https://docs.microsoft.com/en-us/aspnet/core/

Also fix https://docs.microsoft.com/ or create another issue for that

Rick-Anderson:
![image](https://user-images.githubusercontent.com/3605364/83677851-80930a00-a578-11ea-89cf-6bc383e981c4.png)

Rick-Anderson:
Fix the following:
https://docs.microsoft.com/aspnet/overview
![image](https://user-images.githubusercontent.com/3605364/83678045-c2bc4b80-a578-11ea-8bb8-774c80ff9f5e.png)

wadepickett:
Fixed ASP.NET docs toc portion of this issue in that repo with: https://github.com/dotnet/AspNetDocs/issues/399

PR #18649, opened 2020-06-03
Hub page title and description update

wadepickett:
[Internal Review](https://review.docs.microsoft.com/en-us/aspnet/core/?view=aspnetcore-1.0&branch=pr-en-us-18649)  

Partial fix for: #18647 

* Changed title to "ASP.NET Core documentation"
*  Added link to ASP.NET 4.x content in description.

### DIFF
--- a/aspnetcore/index.yml
+++ b/aspnetcore/index.yml
@@ -1,12 +1,12 @@
 ### YamlMime:Hub
 
 title: ASP.NET Core documentation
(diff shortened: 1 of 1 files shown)


wadepickett:
@Rick-Anderson.  The Hub page design does not allow us to put links in the summary text.  The only option for the ASP.NET Documentation link I have is to create a card for it.    I dropped one in, see if you like it there.  If I make it too prominent new folks will just see "documentation" if they are looking for docs and head there I think without realizing they are leaving ASP.NET Core Docs.  Does this example seem good?

Issue #18269, opened 2020-05-12
Avoid Popup


morganc9:
Hi, I'm playing with Azure B2C to see if I can use it for my company.
The tutorial is quite clear and works perfectly! Thanks a lot.
The only thing is when you click on Login in, it's opening another window (popup).
Is there a mode when we can host the page directly into our page?
I thought that the "RemoteAuthenticatorView" would make it but it's not the case. 

Regarding the doc, it would be nice to have a schema of interaction/checks between azure B2C/the blazor client & the server to show how it is secured.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 7d682d4c-86b5-c47e-89a3-3684b2235320
* Version Independent ID: a80e115c-466a-288d-a918-9ef7514f5add
* Content: [Secure an ASP.NET Core Blazor WebAssembly hosted app with Azure Active Directory B2C](https://docs.microsoft.com/en-us/aspnet/core/security/blazor/webassembly/hosted-with-azure-active-directory-b2c?view=aspnetcore-3.1#feedback)
* Content Source: [aspnetcore/security/blazor/webassembly/hosted-with-azure-active-directory-b2c.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/security/blazor/webassembly/hosted-with-azure-active-directory-b2c.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-security**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
> The tutorial is quite clear and works perfectly!

Thank you ❤️ for saying that @morganc9. It's still a work-in-progress. I'll be making further clarifications in the topic later this week.

> Is there a mode when we can host the page directly into our page?

The flows for B2C involve UI on the Azure side by default. You can [customize the UIs that Azure presents](https://docs.microsoft.com/en-us/azure/active-directory-b2c/customize-ui-overview). It might be possible using MSAL, REST API, and/or Graph API to create an app-side user flow, but I've never investigated doing that myself. It would defeat one of the goals of B2C in that Azure's handling of the UI maintenance is big ⌚ and 💰 savings for a lot of companies. Anyway, you'd have to research further in the Azure docs. It's not something that we'll be addressing in our doc set here.

[...]

PR #18652, opened 2020-06-03
Update Blazor WASM security state container

### LINKED ISSUE #17365: Inaccurate information
Hi, some issues I found when trying to follow this page:
1st: Could you please add the required using directive or at least some assembly/namespace information for the StateContainer? I'm using the latest Blazor WebAssembly template (3.2.0-preview2.20160.5), but I can't inject it because the compiler doesn't know that class (neither does Visual Studio). 

2nd: The fragment for the /login-callback route is called CompletingLoggingIn, not CompletingLogIn. 

---
#### Document Details

* ID: f2cdb52c-843f-5944-429d-11ea821c1e22
* Version Independent ID: 9c4f99ef-f070-b527-4efc-96a04798d9ef
* Content: [ASP.NET Core Blazor WebAssembly additional security scenarios](https://docs.microsoft.com/en-us/aspnet/core/security/blazor/webassembly/additional-scenarios?view=aspnetcore-3.1)
* Content Source: [aspnetcore/security/blazor/webassembly/additional-scenarios.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/security/blazor/webassembly/additional-scenarios.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #17365

[Internal Review Topic (links to ** section)]()

Thanks XXXXXXX :rocket: ... and sorry for the 🐌 *looooooong* 🐌 delay. I've been 🏃😅 to get caught up on issues and making progress now that 3.2 is out.

Two remaining problems here tho ...

* For whatever reason, when I go to sign in to AAD with this app now ...
  * An empty sign-in popup appears as tho the browser is going to maintain the app in the background window. (See screenshot below.)
  * **_BUT_** ... the browser (in the window under the popup) redirects to AAD for the login.
  * I close the empty popup window. The app works fine, but this empty popup window is clearly a bad artifact.
* This code/example works for successful sign in, but it doesn't work for successful sign out in spite of `OnLogOutSucceeded="RestoreState"`. Any ideas on what I need to do?
* I'll get this code into snippets shortly ... I still need to work out which bits will be in the sample app and which can just be snapshots.

### DIFF
--- a/aspnetcore/security/blazor/webassembly/additional-scenarios.md
+++ b/aspnetcore/security/blazor/webassembly/additional-scenarios.md
@@ -5,7 +5,7 @@ description: Learn how to configure Blazor WebAssembly for additional security s
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 06/01/2020
+ms.date: 06/03/2020
 no-loc: [Blazor, "Identity", "Let's Encrypt", Razor, SignalR]
 uid: security/blazor/webassembly/additional-scenarios
 ---
@@ -511,9 +511,62 @@ The following example shows how to:
 
 ## Save app state before an authentication operation
 
-During an authentication operation, there are cases where you want to save the app state before the browser is redirected to the IP. This can be the case when you are using something like a state container and you want to restore the state after the authentication succeeds. You can use a custom authentication state object to preserve app-specific state or a reference to it and restore that state once the authentication operation successfully completes.
+During an authentication operation, there are cases where you want to save the app state before the browser is redirected to the IP. This can be the case when you are using a state container and you want to restore the state after the authentication succeeds. You can use a custom authentication state object to preserve app-specific state or a reference to it and restore that state once the authentication operation successfully completes. The following example demonstrates the approach.
 
-`Authentication` component (*Pages/Authentication.razor*):
+A state container class is created in the app with properties to hold the app's state values. In the following example, the container is used to maintain the counter value of the default template's `Counter` component. Methods for serializing and deserializing the container are based on <xref:System.Text.Json>.
+
+```csharp
+using System.Text.Json;
+
+public class StateContainer
+{
+    public int CounterValue { get; set; }
+
+    public string ToStore()
+    {
+        return JsonSerializer.Serialize(this);
+    }
+
+    public void FromStore(string stored)
+    {
+        var storedState = JsonSerializer.Deserialize<StateContainer>(stored);
+        CounterValue = storedState.CounterValue;
+    }
+}
+```
+
+The `Counter` component uses the container to maintain the `currentCount` value outside of the component:
+
+```razor
+@page "/counter"
+@inject StateContainer State
+
+<h1>Counter</h1>
+
+<p>Current count: @currentCount</p>
+
+<button class="btn btn-primary" @onclick="IncrementCount">Click me</button>
+
+@code {
+    private int currentCount = 0;
+
+    protected override void OnInitialized()
+    {
+        if (State.CounterValue > 0)
+        {
+            currentCount = State.CounterValue;
+        }
+    }
+
+    private void IncrementCount()
+    {
+        currentCount++;
+        State.CounterValue = currentCount;
+    }
+}
+```
+
+The `Authentication` component (*Pages/Authentication.razor*) saves and restores the app's state using local session storage with the `StateContainer` serialization and deserialization methods:
 
 ```razor
 @page "/authentication/{action}"
@@ -555,6 +608,24 @@ During an authentication operation, there are cases where you want to save the a
     public ApplicationAuthenticationState AuthenticationState { get; set; } = 
         new ApplicationAuthenticationState();
 }
+```
+
+This example uses Azure Active Directory (AAD) for authentication. In `Program.Main` (*Program.cs*):
+
+* The `ApplicationAuthenticationState` is configured as the Microsoft Autentication Library (MSAL) `RemoteAuthenticationState` type.
+* The state container is registered in the service container.
+
+```csharp
+using static StandAADRC.Pages.Authentication;
+
+...
+
+builder.Services.AddMsalAuthentication<ApplicationAuthenticationState>(options =>
+{
+    builder.Configuration.Bind("AzureAd", options.ProviderOptions.Authentication);
+});
+
+builder.Services.AddSingleton<StateContainer>();
 ```
 
 ## Customize app routes


guardrex:
I tried to tap the actions explicitly in `OnInitializedAsync` to see if I could get state values to reload ...

```csharp
if (RemoteAuthenticationActions.IsAction(
    RemoteAuthenticationActions.LogInFailed, Action) ||
    RemoteAuthenticationActions.IsAction(
    RemoteAuthenticationActions.LogOutSucceeded, Action) ||
    RemoteAuthenticationActions.IsAction(
    RemoteAuthenticationActions.LogOutFailed, Action))
{
    await RestoreState(AuthenticationState);
}
```

... but no ❤️ from the app ... same behavior. I only get state back on sign in. Every other action, including sign out (`OnLogOutSucceeded="RestoreState"`) doesn't restore the state.

guardrex:
... and WRT that popup ... it's the reverse of ...

https://github.com/dotnet/AspNetCore.Docs/issues/18269

... what's known about the popup being triggered or not being triggered?

guardrex:
@javiercn ... the live guidance is broken. What's here *mostly* fixes it. See my opening remarks.

javiercn:
> * An empty sign-in popup appears as tho the browser is going to maintain the app in the background window. (See screenshot below.)

This is the first time I've seen this.

> * This code/example works for successful sign in, but it doesn't work for successful sign out in spite of `OnLogOutSucceeded="RestoreState"`. Any ideas on what I need to do?

Not sure what it means by "It doesn't work" can you be more specific? Does the logout flow complete successfully?

javiercn:
> ```cs
> ```cs
> if (RemoteAuthenticationActions.IsAction(
>     RemoteAuthenticationActions.LogInFailed, Action) ||
>     RemoteAuthenticationActions.IsAction(
>     RemoteAuthenticationActions.LogOutSucceeded, Action) ||
>     RemoteAuthenticationActions.IsAction(
>     RemoteAuthenticationActions.LogOutFailed, Action))
> {
>     await RestoreState(AuthenticationState);
> }
> ```
> 
> 
> ... but no ❤️ from the app ... same behavior. I only get state back on sign in. Every other action, including sign out (OnLogOutSucceeded="RestoreState") doesn't restore the state.
> Let me know if you'd like me to put the whole test app up on GH.
> ```

Not sure what you are trying to do here, but the way to do this is inside RestoreState to check if you are on the LoginCallback or the LogoutCallback and restore the state there.

For other scenarios it doesn't matter because you either logged in silently or through a pop-up so you never left the app.

guardrex:
"doesn't work" = "state isn't restored"

Consider the `Counter` component counter value as I show on the PR ...

* It recalls the state on log in. :+1:
* It will not restore state after logging out or any other auth state change, which seems especially strange given that you provided `OnLogOutSucceeded`.

Here's the process and when I expect it to restore state ...
  * Count up a few times with the `Counter` component while the user isn't logged in.
  * Log in.
  * Navigate back to the `Counter` component. The state is known, and the last value is shown. :+1:
  * Count up a few more times.
  * Log out.
  * Navigate back to the `Counter` component. **_The count is reset back to 0 (zero)._** :x: However, the value is still in local storage AFAIK. `RestoreState` didn't run.

Same thing happens if you fail to authenticate and return to the app: The log in "failed" action doesn't seem to trigger `RestoreState` (with the code in my followup comment :point_up:). The only auth state change that results in re-establishing state thus far is the "succeeded" log in case.

That why I was trying it with `RemoteAuthenticationActions.IsAction`. I just wanted to see if that approach would be effective.

What does the code need to do to restore state for ...

* Log in failed
* Log out succeeded
* Log out failed

... given the state container approach on the PR.

guardrex:
> inside RestoreState to check if you are on the LoginCallback or the LogoutCallback and restore the state there

That seems like a strange design given that `RestoreState` is provided to something called "OnLogInSucceeded" ... I would expect to have to assign it to "OnLogInFailed," "OnLogOutSucceeded" and "OnLogOutFailed". I mean if the dev gets fine-grained control by assigning *different methods* for each. In my case, I'd be fine just assigning `RestoreState` to all of them.

guardrex:
> For other scenarios it doesn't matter because you either logged in silently or through a pop-up so you never left the app.

... but you did leave the app for ...

* Login that fails at the provider.
* Logout that succeeds or fails at the provider.

javiercn:
> ... but you did leave the app for ...

You didn't if that happens inside an iframe or in a popup. You app is still open in the main browser window and all your state is kept in memory, why restore what never got destroyed?

guardrex:
I'm not doing it. Like I show in the screenshot ... the app's browser window is redirecting and the popup is blank.

I've only tested in one browser (Firefox) and only in an inprivate window. I'll try again outside of inprivate mode, and I'll try a few browsers. However, the app works fine normally with inprivate/incognito mode ... popups appear and work as expected.

If it turns out that inprivate/incognito breaks this aspect of Blazor apps, then I'll need to call that out here. I'll get back to you.

javiercn:
> If it turns out that inprivate/incognito breaks this aspect of Blazor apps, then I'll need to call that out here. I'll get back to you.

I remember something like this in the past, don't remember if it was with firefox or with safari

guardrex:
What I think has gone wrong with my state restore is that I'm checking `AuthenticationState.Id` in `RestoreState` and using the `Id` to key the state data. Returning from the provider to the app on a failed sign in or sign out, I suspect that there **_is no Id_**, so state isn't restored. I'll work on this and see if that's right and what I can do about it.

Anyway ... wrt the strange popup window behavior .......

Taking the app off of inprivate/incognito didn't change the behavior.

Sign in with Chrome ...

Capture

Sign in with Edge ...

Capture1

Using that popup to sign in with Edge ...

Capture2

... it gets stuck there. The main window logs in normally tho.

The app runs normally if I revert it back over to the stock `Authentication` component from the template.

It's true as you say that with popup behavior and the app remaining in the background browser window that state is fine.

javiercn:
@guardrex the popup + the backwindow redirecting is weird, I'm not sure if its caused by a misconfiguration in your AAD settings.

I haven't seen it in the apps I use as a sample.

guardrex:
It works fine if I revert it back to the default (template) `Authorization` component. I'll try a fresh app with a fresh Azure registration in the morning.

I thought perhaps that the reason I wasn't getting the logout behavior that I wanted (even in this broken app) was that I wasn't saving it again prior to logout. So I tried both ...

```csharp
if (RemoteAuthenticationActions.IsAction(
    RemoteAuthenticationActions.LogOut, Action))
{
    await JS.InvokeVoidAsync("sessionStorage.setItem",
        AuthenticationState.Id, State.ToStore());
}
```

... and ...

```csharp
if (RemoteAuthenticationActions.IsAction(
    RemoteAuthenticationActions.LogOut, Action))
{
    AuthenticationState.Id = Guid.NewGuid().ToString();

    await JS.InvokeVoidAsync("sessionStorage.setItem",
        AuthenticationState.Id, State.ToStore());
}
```

In the first case, the Id is null. In the second case, it saves with a new Id but it doesn't restore the state using the new Id after returning from the Azure redirect ...

Capture

Anyway ... I understand that I need to get it working properly before checking on logout state behavior. I'll pick back up with it in the morning.

guardrex:
Ah ... mini-breakthrough on that logout behavior. I merely forgot to reactivate my `OnLogOutSucceeded="RestoreState"` in the `<RemoteAuthenticatorViewCore>` element.

After fixing that and dropping in .......

```csharp
if (RemoteAuthenticationActions.IsAction(
    RemoteAuthenticationActions.LogOut, Action))
{
    AuthenticationState.Id = Guid.NewGuid().ToString();

    await JS.InvokeVoidAsync("sessionStorage.setItem",
        AuthenticationState.Id, State.ToStore());
}
```

... (and the new Id is required because it's `null` otherwise) ... even with this app's silly popup/redirect behavior, **_it does restore state after logout._** :tada:

I'm still not clear on if the following is **_required_** when I get the popup/redirect situation sorted out. Thus far, it looks like it would need `LogIn` **_AND_** `LogOut` to keep the state across both gestures ...

```csharp
if (RemoteAuthenticationActions.IsAction(RemoteAuthenticationActions.LogIn,
    Action) ||
    RemoteAuthenticationActions.IsAction(RemoteAuthenticationActions.LogOut,
    Action))
{
    AuthenticationState.Id = Guid.NewGuid().ToString();

    await JS.InvokeVoidAsync("sessionStorage.setItem",
    AuthenticationState.Id, State.ToStore());
}
```

... I'll find out when I get the app's popup/redirect behavior fixed.

guardrex:
I made updates along the lines of what works here ...

* Including both ...
  * Candidate checks for `RemoteAuthenticationActions.LogIn` and `RemoteAuthenticationActions.LogOut`  to save state.
  * `OnLogInSucceeded="RestoreState"` and `OnLogOutSucceeded="RestoreState"` for restoring state.
* It's optional I guess, but I also add a line to `RestoreState` to remove the local storage entry after state is restored. *Now, that's a tidy app!*
* I use dead-simple naming for methods and vars so that it's clear what everything is.

The last thing to resolve is the popup/redirect behavior. Everything else is working perfectly. I'll take another 🔪 at it in the morning.

PR #18659, opened 2020-06-04
Fix link to localization sample

sibbl:
As the whole document refers to the sample code, the invalid URL should be fixed.

As the document targets .NET Core 3 or newer, I've changed the URL to the 3.x sample code instead of the parent directory including samples for .NET Core 2 as well.


guardrex:
Thanks @sibbl! :rocket:

cc: @hishamco

PR #18678, opened 2020-06-05
Admonish Blazor iframe sign in

### LINKED ISSUE #18548: x-frame-options security error with some identity providers
A number of identity providers won't work with the iFrame-based b2c login. For example, Google and Amazon produce these errors:

Refused to display 'https://accounts.google.com/signin/oauth?client_id=173429469116-dfb4asbe0mr3e3tghhpv0d5jbdfnd7c9.apps.googleusercontent.com&redirect_uri=https://wickedbyteapps.b2clogin.com/wickedbyteapps.onmicrosoft.com/oauth2/authresp&response_type=code&scope=email+profile&state=StateProperties%3...' in a frame because it set 'X-Frame-Options' to 'deny'.

Refused to display 'https://na.account.amazon.com/ap/oa?arb=5c72c689-4587-4f7c-9379-4a44626219f9' in a frame because it set 'X-Frame-Options' to 'sameorigin'.

Is there a workaround for this?

---
#### Document Details

* ID: 7d682d4c-86b5-c47e-89a3-3684b2235320
* Version Independent ID: a80e115c-466a-288d-a918-9ef7514f5add
* Content: [Secure an ASP.NET Core Blazor WebAssembly hosted app with Azure Active Directory B2C](https://docs.microsoft.com/en-us/aspnet/core/security/blazor/webassembly/hosted-with-azure-active-directory-b2c?view=aspnetcore-3.1&source=docs)
* Content Source: [aspnetcore/security/blazor/webassembly/hosted-with-azure-active-directory-b2c.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/security/blazor/webassembly/hosted-with-azure-active-directory-b2c.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #18548

Thanks @marshallr12! :rocket:

Javier ... not sure about the wording on this one. I say configure the behavior with the IP because there isn't a way to configure it on the app side AFAIK.

Engineering discussion is at :point_right: https://github.com/dotnet/aspnetcore/issues/22343

### DIFF
--- a/aspnetcore/security/blazor/index.md
+++ b/aspnetcore/security/blazor/index.md
@@ -5,7 +5,7 @@ description: Learn about Blazor authentication and authorization scenarios.
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 05/19/2020
+ms.date: 06/05/2020
 no-loc: [Blazor, "Identity", "Let's Encrypt", Razor, SignalR]
 uid: security/blazor/index
 ---
@@ -49,6 +49,10 @@ For more information on creating apps and configuration, see <xref:security/blaz
 Blazor Server apps operate over a real-time connection that's created using SignalR. [Authentication in SignalR-based apps](xref:signalr/authn-and-authz) is handled when the connection is established. Authentication can be based on a cookie or some other bearer token.
 
 For more information on creating apps and configuration, see <xref:security/blazor/server/index>.
+
+## Avoid sign in with an iframe
+
+Sign in with an iframe isn't recommended due to security and browser compatibility concerns. Configure the app with the identity provider to use the redirect or popup approaches to sign in users.
 
 ## AuthenticationStateProvider service
 


javiercn:
Err, this is not correct to the best of my understanding. We should not have this in the docs. Using an iframe to sign-in is used within the OAuth and OPEN ID Connect specs and its perfectly fine. `X-Frame-Options` is not a standard and is legacy, it is recommended to use CSP with a proper frame ancestors value to constrain the iframe to allowed domains, but even if that is not the case, there is no security concern here.

guardrex:
@blowdart said ...

> Given that safari has now irrevocably broken iframe login in the name of privacy we shouldn't be recommending it any more in docs, but rather switch to using the popups method.

Reference: https://github.com/dotnet/aspnetcore/issues/22343#issuecomment-639117413

guardrex:
I'm going to move my note on the engineering issue over here. Even if we went forward with the guidance on this PR, it isn't (apparently) actionable guidance because control over the approach (popup or redirect) is either something that I don't know how to configure or it's something that can't be configured. Also, I've hit some strange behavior recently with https://github.com/dotnet/AspNetCore.Docs/pull/18652.

Here are my remarks ...

IIRC, I read in MSAL docs that one can configure the type of login (popup or redirect) by setting `LoginMode` to `Popup` or `Redirect` in the AD config settings. However, I understand that such a setting might pertain to using MSAL.NET directly. I tried setting it to `Redirect` in a test app (based on our Azure AD guidance) that was using the popup approach, but it had no effect. I assumed that it was hardwired into the framework at this point ... no dev setting was mentioned when we were putting the docs up.

Recently, strange popup/redirect behavior appeared while I was working on restoring state with `RemoteAuthenticatorViewCore` ... fixing up our coverage and fleshing it out a bit more ...

**Update Blazor WASM security state container**
https://github.com/dotnet/AspNetCore.Docs/pull/18652

The app (standalone WASM with AAD) is now a bit scatterbrained about which approach to use. It's throwing up an empty popup, and the underlying browser window is using the redirect approach. I show a screenshot on that PR of what's happening here. I close the popup, and the app signs in users and behaves normally using the redirect approach. This behavior only comes about due to the changes on that PR with `RemoteAuthenticatorViewCore`.

I need guidance to proceed ...

* If there's a fix for the behavior, let me know ... I'll write it into the updates on that PR.
* If it's something that we need to call out in text (e.g., this is a temporary known issue), I'll add a NOTE on it there.
* If that PR can't proceed without engineering work, then let me know if I need to create an engineering issue. In that case tho, shouldn't we remove that section from the topic and not support restoring state at this time? I don't think we can put up broken guidance **_and_** not even have a NOTE about it.

blowdart:
@javiercn See https://leastprivilege.com/2020/03/31/spas-are-dead/ for how Safari is breaking SPAs

javiercn:
@blowdart it is not clear to me what breaks in those scenarios. It is somewhat clear that it will break silent sign-in, but I'm not confident on how it will break silent refresh. 

I don't think we have an alternative now on how to handle these changes if they in fact, end up breaking OAuth scenarios like silent refresh. We'll have to see how things evolve in the future, but giving this guidance at this time is not adequate IMO. Specially given the lack of a proper alternative as BFF doesn't cover the same set of scenarios (no support for standalone apps) and I don't think we want to recommend using refresh tokens in the browser even with all the proper mitigations in place.

blowdart:
Fair. You own blazor auth, so it's your guidance and decision here :)

marshallr12:
Since one of the "Hosted with AAD B2C" scenario already creates a BFF after following the tutorial [here](https://docs.microsoft.com/en-us/aspnet/core/security/blazor/webassembly/hosted-with-azure-active-directory-b2c?view=aspnetcore-3.1) is it possible to to enable "dotnet new blazorwasm" to create an app that uses the SameSite cookie approach?

javiercn:
> Since one of the "Hosted with AAD B2C" scenario already creates a BFF after following the tutorial [here](https://docs.microsoft.com/en-us/aspnet/core/security/blazor/webassembly/hosted-with-azure-active-directory-b2c?view=aspnetcore-3.1) is it possible to to enable "dotnet new blazorwasm" to create an app that uses the SameSite cookie approach?

That is not a BFF we don't stamp a cookie out of the result of an OAuth interaction and the entire OAuth dance happens between the Browser and Azure AD B2C. The server is never involved.

Also relying on SameSite=true relies on you having full control of what browsers your users use to access your app, which we can't predict in general and ultimately exposes you to CSRF attacks. While the number of Browsers that don't support SameSite cookies is going down quickly, we are still not at a point where we want to base our security implementations on it.

marshallr12:
What I meant by BFF in the "Hosted with AAD B2C" scenario is that it involves the webassembly front end tied to a backend web API that provides the Weather Forecast data. The WeatherForecastController already has an [Authorize] attribute. SameSite cookie support currently appears to be similar to webassembly support except for the Safari bug, at least based on looking at [caniuse.com](https://caniuse.com/#feat=same-site-cookie-attribute).

javiercn:
> SameSite cookie support currently appears to be similar to webassembly support except for the Safari bug, at least based on looking at [caniuse.com](https://caniuse.com/#feat=same-site-cookie-attribute).

Yes, but that approach prevents you from using other things in your app like MVC or Razor Pages without either complicating the security design a lot or forcing you to implement CSRF protection in all your API endpoints.

The difference here is that the APIs always use a token for Authorization, the moment you switch to a cookie, you are exposing yourself to CSRF from browsers that don't support SameSite and you can't predict in general whether a user will be able to login on a site without a browser that supports samesite cookies.

marshallr12:
I see. You guys have a great product - I started with classic ASP, and IMO this is the best technology coming out of Microsoft since .NET 1.0 .

sravimohan:
Has anything happened with with.   I am still getting the below from the blazor wasm app.
`Refused to display 'https://xxx.auth.ap-southeast-2.amazoncognito.com/' in a frame because it set 'X-Frame-Options' to 'deny'.`

guardrex:
@sravimohan ... We don't work on closed issues/PRs. Also, that's already been addressed here: Microsoft isn't the problem here. The specs are being followed. You'd need to take that up with Amazon support.

PR #18693, opened 2020-06-06
Add example solution to large request body section

thomasdane:
The current content for `Avoid reading large request bodies or response bodies into memory` describes the problem but does not provide suggestions for mitigation. This commit adds a link to other documentation that can help solve the large request body issue.

### DIFF
--- a/aspnetcore/performance/performance-best-practices.md
+++ b/aspnetcore/performance/performance-best-practices.md
@@ -212,6 +212,8 @@ Naively storing a large request or response body into a single `byte[]` or `stri
 * May result in quickly running out of space in the LOH.
 * May cause performance issues for the app because of full GCs running.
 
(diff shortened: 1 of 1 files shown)


guardrex:
Hello @thomasdane ... That's not a solution to the problem. I would guess something more like a custom streaming approach. Also, that's a bit too specific for this doc. This is a general problem, not just pertaining to file uploads.

Ping @mjrousos ... and note Mike that in our [upload large files scenario](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/file-uploads?view=aspnetcore-3.1#upload-large-files-with-streaming), I think the guidance violates the guidance you provide here. 🙈 ... We probably need an issue to address that in the topic if so, but idk who would have time to work it these days.

Issue #18222, opened 2020-05-09
Incorrect explanation of Genre's regular expression validation.


CommanderMark:
The explanation for the validation attributes mentions that it ensures Genre "Must only use letters."

The RegEx however shows `^[A-Z]+[a-zA-Z""'\s-]*$`, which indicates that every character that isn't the first one can include double/single quotes, white space and a dash in addition to letters.

---
#### Document details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 39b6d0f3-9bc7-d091-0d45-9fc1e2e7aac8
* Version Independent ID: 20e7eb53-215b-09ab-205d-d20177221278
* Content: [Add validation to an ASP.NET Core Razor Page](https://docs.microsoft.com/en-ca/aspnet/core/tutorials/razor-pages/validation?view=aspnetcore-3.1&tabs=visual-studio#feedback)
* Content Source: [aspnetcore/tutorials/razor-pages/validation.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/tutorials/razor-pages/validation.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-tutorials**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
Yes, that should be
`   [RegularExpression(@"^[A-Z]+[a-zA-Z]*$")]`

This also needs to be fixed in https://docs.microsoft.com/en-ca/aspnet/core/tutorials/first-mvc-app/validation

PR #18699, opened 2020-06-08
Genre regular expression

dharmatech:
Update Genre regular expression to match the intention as described in the tutorial text.

[This page](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/tutorials/razor-pages/validation.md) says that Genre:

* Must only use letters
* The first letter is required to be uppercase. White space, numbers, and special characters are not allowed.

The original regular expression used:

`@"^[A-Z]+[a-zA-Z""'\s-]*$"`

says to allow double quotes, whitespace, and dashes after the first letter. This contradicts "must only use letters".


scottaddie:
@dharmatech Thank you for bringing this issue to our attention for opening a PR to fix it.

serpent5:
@scottaddie This fixes the report in #18222, but Rick also calls out another place that needs fixing. Shall I PR that second topic so the issue can be closed?

scottaddie:
@serpent5 Yes, please. Thank you for catching that.

PR #18700, opened 2020-06-08
adding-view.md: add missing words

dharmatech:
Before:

> The default view file has the same name as the method (`Index`), so in the */Views/HelloWorld/Index.cshtml* is used.

After:

> The default view file has the same name as the method (`Index`), so the view template in */Views/HelloWorld/Index.cshtml* is used.


scottaddie:
Thank you for the PR, @dharmatech! Your change will go live within the next 24 hours.

PR #18759, opened 2020-06-11
SignalR naming consistency

### LINKED ISSUE #18758: SignalR consistency
Some topics use `/myhub`, while other topics use `/chatHub`/`chathub`. Let's go with "chathub" (lowercase) everywhere.

Also, there might be one or more `var client =` for building new hub connections. Let's go with the more common `var connection =`.

guardrex:
Fixes #18758

* Went lowercase in URL endpoints everywhere (`chathub`).
* There was just one `var client =` to update.


guardrex:
I had a filter on my search that didn't capture them.

I'll commit additional updates now.

guardrex:
I think that licks 'um.

PR #18798, opened 2020-06-13
Drop configure AuthorizationMessageHandler guidance

### LINKED ISSUE #18778: WASM: The inner handler has not been assigned - Configure AuthorizationMessageHandler
When trying to use the code in the section #configure-authorizationmessagehandler, I receive the 

> error: Microsoft.AspNetCore.Components.WebAssembly.Rendering.WebAssemblyRenderer[100]
>       Unhandled exception rendering component: The inner handler has not been assigned.

In a vanilla Blazor WASM app (from the template) have the following in Program.cs:

```
	builder.Services.AddTransient(sp =>
			{
				return new HttpClient(sp.GetRequiredService<AuthorizationMessageHandler>()
						.ConfigureHandler(
								authorizedUrls: new[] { $"https://api.mysite.com" },
								scopes: new[] { "scope1", "scope2" }))

				{
					BaseAddress = new Uri($"https://api.mysite.com")
				};
			});
```
I then call Http.GetFromJsonAsync and the above error occurs.

If I configure a named client (i.e. use AddHttpClient rather than AddTransient) otherwise identically, I don't get the error.

Is there something missing in the docs, or is it something I'm doing wrong?

---
#### Document details

* ID: f2cdb52c-843f-5944-429d-11ea821c1e22
* Version Independent ID: 9c4f99ef-f070-b527-4efc-96a04798d9ef
* Content: [ASP.NET Core Blazor WebAssembly additional security scenarios](https://docs.microsoft.com/en-gb/aspnet/core/security/blazor/webassembly/additional-scenarios?view=aspnetcore-3.1#configure-authorizationmessagehandler)
* Content Source: [aspnetcore/security/blazor/webassembly/additional-scenarios.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/security/blazor/webassembly/additional-scenarios.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #18778

The original example came in from the Preview 5 blog post at https://devblogs.microsoft.com/aspnet/blazor-webassembly-3-2-0-preview-5-release-now-available/. It's at the end of the *Attach tokens to outgoing requests* section, except that we changed the registration from `AddSingleton` to `AddTransient` later. IIRC, I tested it back in those days, and it worked back then.

> That code snippet is wrong. Likely because you are not using HttpClientFactory. In that case you need to create an HttpClientHandler instance and assign it as the innerhandler of the AuthorizationMessageHandler manually.

☝️ Didn't work for me ... no doubt due to my not implementing it exactly right. The error is that the `AuthorizationMessageHandler` isn't configured ... but of course that was the goal of my code. Clearly, I botched it.

In any case, you said earlier that the **_custom AuthorizationMessageHandler class_** is the preferred approach. I tested that, and it works great. :tada: Therefore, I punt 🏈 here and drop the errant section. If you want to keep that section, give me the code example to make it work.

btw ... The change on Line 56 is a passing nit. We need to start code fencing ALL C# for localization. The MT is translating these. *Yikes!*

Thanks @paulguz! :rocket: ... sorry we didn't get this sooner. You're the first one to report a problem with that example/approach.

### DIFF
--- a/aspnetcore/security/blazor/webassembly/additional-scenarios.md
+++ b/aspnetcore/security/blazor/webassembly/additional-scenarios.md
@@ -5,7 +5,7 @@ description: Learn how to configure Blazor WebAssembly for additional security s
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 06/10/2020
+ms.date: 06/13/2020
 no-loc: [Blazor, "Identity", "Let's Encrypt", Razor, SignalR]
 uid: security/blazor/webassembly/additional-scenarios
 ---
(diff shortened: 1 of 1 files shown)


guardrex:
Tried again this morning ... no 🎲🎲.

The custom `AuthorizationMessageHandler` class works well.

Side ❓ tho: For the section on using a custom `AuthorizationMessageHandler`, should we add a note for MSAL users (`AddMsalAuthentication`) that one doesn't need to specify the `ProviderOptions.DefaultAccessTokenScopes`?

guardrex:
This PR will be re-opened after the ToC updates are complete.

Issue #18829, opened 2020-06-17
Improve Blazor built-in component content and ToC


guardrex:
Blazor ToC links of the built-in components are matched (selected) in the sidebar for cross-links and search landing pages. This isn't desirable behavior, so I'll add bookmarks to the links in hopes that they won't be matched (selected).

PR #18831, opened 2020-06-17
Blazor ToC updates

### LINKED ISSUE #17169: Further split Blazor Razor Components
Tracking issue for splitting long topics: https://github.com/dotnet/AspNetCore.Docs/issues/17167

I-dears ...

* Reduce the length of the *Use components* example.
* Split out *Cascading values and parameters* section into its own topic.

@mkArtakMSFT, did you spot additional splits or *document diet* 🍰 strategies? Is this list ok thus far?

---
#### Document Details

* Content: [Create and use ASP.NET Core Razor components](https://docs.microsoft.com/en-us/aspnet/core/blazor/components?view=aspnetcore-3.1)
* Content Source: [aspnetcore/blazor/components.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/components.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

### LINKED ISSUE #18064: Missing overview of built-in components
It is hard to find information on built-in components like InputText and InputDate

---
#### Document Details

* Content: [Create and use ASP.NET Core Razor components](https://docs.microsoft.com/en-us/aspnet/core/blazor/components?view=aspnetcore-3.1&source=docs#feedback)
* Content Source: [aspnetcore/blazor/components.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/components.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

### LINKED ISSUE #18217: Have a new article dedicated to listing the differences in Blazor Client and Server
Here is a proposed outline:

## What is Blazor

[High level description about single component model]

## Hosting models

[Lightly describe the two hosting model - Client and Server and reference to a separate article we have detailing each one separately]

## Differences between the two hosting models

[Small ToC for the below topics, listing differences between Blazor WebAssembly and Server-Side Blazor]

[Each of the following topics is going to have two subsections for Blazor WASM and Blazor Server, detailing the way to apply default configuration for each feature with a link to more detailed configuration in a separate article if there is one.
In the beginning of each, however, there will be a small text summarizing the conceptual difference]

guardrex:
Fixes #18217
Fixes #18064
Fixes #17169

* We want to adopt the SignalR model for topic physical locations (i.e., SignalR topics in the *signalr* folder): We want all Blazor topics (except for the tutorials) in the *blazor* folder.
* We want to mirror the *Fundamentals* approach of the doc set at large and collect the app-level fundamentals topics in a node before the *Components* topics.
* Components-related topics go into the *Components* **node** with current *Components* topic content landing in an **overview** topic there.
* On the new *Built-in components* ToC node, I only want to link to **_the topic_** at this time. This gets the ball rolling. I have additional ideas on how to improve the navigation/content, but I'd like to address it further on a new issue that I opened: [Improve Blazor built-in component content and ToC (dotnet/AspNetCore.Docs #18829)](https://github.com/dotnet/AspNetCore.Docs/issues/18829).
* Good success here on **_NOT_** using H2 headings for WASM and Server within topics because that's all that will show in the sidebar sections list when we do that. So far, so good. I'll keep working on these types of within-topic layout improvements when we get to UE work for each topic later.
* Artak, I worked components content layout forward per our earlier discussion, but I recommend that we not deeply focus on it for review here unless something looks really off. Let's take it up on the UE review of the topic shortly, when I'll be able to focus exclusively on the subject.
* Scott, "/index" crept into the redirection file `redirect_url` entries, which seems to result in an extra redirect. I pull that segment throughout the file to fix it ... *if I'm right.* 😄 If you want to see the network fetches, let me know ... I can paste them into a comment here.


guardrex:
lol ... that ☝️ was unexpected. *I was sure that I fouled something up somewhere.* 😁

Thanks for looking it over. I'll call for Artak now.

scottaddie:
@guardrex Are there plans to move the **Hosting models** doc to the **Host and deploy** ToC node?

aspnetcore/blazor/components/lifecycle.md
```diff
@@ -156,7 +156,7 @@ protected override bool ShouldRender()
 
 Even if <xref:Microsoft.AspNetCore.Components.ComponentBase.ShouldRender%2A> is overridden, the component is always initially rendered.
 
-For more information, see <xref:performance/blazor/webassembly-best-practices#avoid-unnecessary-component-renders>.
+For more information, see <xref:blazor/webassembly-performance-best-practices#avoid-unnecessary-component-renders>.
```
mkArtakMSFT, line 159: I think this also should be part of fundamentals

guardrex:
Two things come to mind for keeping the topic in its current location. These are my definitions and notions, not necessarily those of the product unit ...

* The word "host" is overloaded, flexible, and difficult to pin down. However, I think about "host" in this context is a model for how Razor components are executed, client or server. Then, there's a different "host" associated where the app lives (server) and is distributed (network). For example, a WASM app is executed on the client, but the app still has to come from a "host" (server) somewhere. I admit that each model's content between the *Hosting models* topic and their landing topics in the *Host and deploy* node has no hard separation. The WASM and Server topics in the *Host and deploy* node aren't much more than extending concepts and implementation details for the two models. But, I'll say *why a split at all* in a sec&dagger;.
* The hosting model choice is foundational to the app. It's like the first decision for one to make before they lay the first line of code down; whereas for other apps, you know that one often only needs a vague notion ... or even no notion of hosting (server/network) in the base case. Topics that follow will refer to the hosting models, so the reader has to have some notion of them as they read down the ToC. In the future tho, this might change for Blazor. We have an open issue to address making Razor components work in the *Server* project of a hosted Blazor app **_and_** an open issue for making a kind of *dual-boot thingy* that can fire right up with minimal changes (or via configuration) in either WASM or Server models. When it reaches that point ... if it does reach that point, then perhaps there wouldn't be such a strong need to explain hosting model concepts so early in the ToC.

Those two sets of ideas seem to boil down to having the *Hosting models* topic early in the reading.

&dagger;WRT that split ... where to make it ... answering the question if a given concept should be addressed in the *Hosting models* topic or a different topic further down the ToC: I want to keep the *Hosting models* topic light and loose, introductory with just the basics. I'm explicitly avoiding a lot of implementation concepts and examples. In fact, I've moved content out of this topic to lighten it up. I just want the reader to:

* Get a basic familiarity such that in the following topics the reader understands what the models are.
* Be able to make an informed choice of model based on the product unit's list of pros and cons.

guardrex:
@mkArtakMSFT Sure ... I'll wait for DR to respond and affirm that and do it first thing in the morning if so. I'm already in 🛏️😪💤.

guardrex:
Side note WRT the perf topic ...

"Performance" hasn't been considered "fundamental" set of concepts thus far in ASP.NET Core docs. They're kept in their own node in the main ToC.

I'm not making a suggestion either way ... just mentioning that it would be a break with what's been done in the past on the repo.

guardrex:
I'll open a followup issue to address the perf topic. I'll merge this now to unblock the next work item ... code fencing to avoid over localization.

PR #18849, opened 2020-06-18
Add a note about output formatter buffering

pranavkm:

### DIFF
--- a/aspnetcore/migration/22-to-30.md
+++ b/aspnetcore/migration/22-to-30.md
@@ -428,6 +428,16 @@ If you have trouble with <xref:System.IO.TextWriter> implementations or other st
 
 For more information, see [[Announcement] AllowSynchronousIO disabled in all servers (dotnet/AspNetCore #7644)](https://github.com/dotnet/AspNetCore/issues/7644).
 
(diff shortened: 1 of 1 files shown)


zvolkov:
This is nice, but does not mention change of behavior as far as response completion. If this were me, I would add a paragraph saying "When buffering is enabled, Content-Length header is set by the output formatter even if it wasn't set by your code. When this happens, the client will consider the request completed as soon as all bytes are received and before your middleware finishes e.g. saving the session state. For more details see the following GitHub issue:"

pranavkm:
> saving the session state, because the client may now send the next request before the previous request is done updating the state.

This seems problematic and we wouldn't recommend doing this to our users. I'm happy to document the Content-Length since it's something users have asked for.

zvolkov:
That piece was just an example of the problem, not a recommendation.

Rick-Anderson:
Branched to #18853, work on that until it merges into this branch.

Rick-Anderson:
@pranavkm  can I S&M (squash and merge)?

pranavkm:
:shipit:

PR #18909, opened 2020-06-22
updating formatting and fixing grammatical error

raydixon:
Adding bold to `Note:` prefix and correcting punctuation in paragraph about the `Where` method returning a collection and using `Single`.


raydixon:
Thanks, @Rick-Anderson!

PR #18917, opened 2020-06-23
[HOLD] Added link to a bug which prevents PWA update

MichaelSL:
There is a bug in PWA manifest generator which prevents PWA cache update. Added a link to an issues for explanation and possible workaround.


guardrex:
We prefer not to document short-term framework problems that will be patched quickly unless they're serious and widespread. @mkArtakMSFT, will the next patch release with the fix for this land soon?

MichaelSL:
Adding more 'keywords' to the issue description may be the alternative for now. I was able to find the issue on github only after going through the target's source.

guardrex:
**Update** ... Although the patch release date isn't published publicly ahead of time, I do know the date and can say that it won't be long 🕙. We'll let the patch take care of this.

PR #18950, opened 2020-06-25
Updated Home to Index

snehansh:
Updated Home to Index since Index.razor is created.


guardrex:
Pertains to the running app. This page is called "Home" in the app's sidebar nav menu.

PR #18955, opened 2020-06-25
`foreach` variable can be captured.

nikonthethird:
Ever since C# 5.0, the `foreach` variable can be captured in a lambda, and it works. [See the breaking changes section from the VS magazine back then.](https://visualstudiomagazine.com/articles/2012/11/01/more-than-just-async.aspx)

I tested it in Blazor, and `foreach` does not break like `for` does.


guardrex:
Thanks @nikonthethird! :rocket:

PR #18981, opened 2020-06-26
Blazor Server EF Core guidance and sample

### LINKED ISSUE #16558: [Blazor] Blazor + EF documentation topic
Can we make an article about using EF with serverside Blazor in the same way we have one for Razor pages?

I wrote a comment and a sample on the pattern to follow here.

The main points are:
* Showcase how to do CRUD
* Use OwningComponentBase to isolate the usage of the context to the component it is being used in.
* Explain that if you use the context directly, the context is shared among components, which might cause multiple queries to start at the same time, and that is something that Entity Framework doesn't support.

https://github.com/dotnet/aspnetcore/issues/16745#issuecomment-549505605

JeremyLikness:
[EDIT by guardrex to add the 'fixes']

Fixes #16558

This PR provides a sample application and guidance for using EF Core in Blazor Server apps.


JeremyLikness:
Pinging for review ...

aspnetcore/blazor/blazor-server-ef-core.md
```diff
@@ -0,0 +1,120 @@
...
+* Use a flag to prevent multiple concurrent operations:
+
+    ```csharp
+    if (Loading)
+    {
+        return;
+    }
+    Loading = true;
+    // operations go here
+    Loading = false;
```
javiercn, line 46: this should go in a try..finally

aspnetcore/blazor/blazor-server-ef-core.md
```diff
@@ -0,0 +1,120 @@
...
+
+[!code-razor[](./common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/EditContact.razor?range=5-7)]
+
+The sample app includes a helper method for disposal of the current context and ensures it is disposed when the component is:
+
+[!code-csharp[](./common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/EditContact.razor?range=172-191)]
+
+Finally, `OnInitializedAsync` is overridden to create a new context. In the sample app, it loads the contact in the same method. The `DisposeContext` call ensures any previous context is disposed when the component is reused.
+
+[!code-csharp[](./common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/EditContact.razor?range=85-98)]
```
javiercn, line 88: This https://github.com/JeremyLikness/AspNetCore.Docs/blob/jrl-blazor-server-ef-core/aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/EditContact.razor#L92 looks wrong, on initalizedasync only gets called once per component. Implementing dispose should be enough

javiercn: If you want to dispose the previous context when the component instance gets reused you are looking for OnParametersSetAsync

aspnetcore/blazor/blazor-server-ef-core.md
JeremyLikness: Done

aspnetcore/blazor/blazor-server-ef-core.md
JeremyLikness: Makes sense. I updated the changes and simplified the control.

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Data/DbContextFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// <summary>
+        /// Service provider.
+        /// </summary>
+        private readonly IServiceProvider _provider;
+
+        /// <summary>
+        /// Inject the service provider.
+        /// </summary>
+        /// <param name="provider">The <see cref="IServiceProvider"/> instance.</param>
+        public DbContextFactory(IServiceProvider provider)
```
javiercn, line 22: This should create its own SF scope internally and dispose it at the end? Otherwise the IServiceProvider will hold a reference to each created DbContext until it goes away

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Data/FactoryExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// </summary>
+        /// <typeparam name="TContext">The instance of <see cref="DbContext"/> to register.</typeparam>
+        /// <param name="collection">The instance of the <see cref="IServiceCollection"/>.</param>
+        /// <param name="optionsAction">Optional access to the <see cref="DbContextOptions{TContext}"/>.</param>
+        /// <param name="contextAndOptionsLifetime">Set the <see cref="ServiceLifetime"/> of the factory and options.</param>
+        /// <returns>The registered <see cref="IServiceCollection"/>.</returns>
+        public static IServiceCollection AddDbContextFactory<TContext>(
+            this IServiceCollection collection,
+            Action<DbContextOptionsBuilder> optionsAction = null,
+            ServiceLifetime contextAndOptionsLifetime = ServiceLifetime.Singleton)
```
javiercn, line 24: I'm not sure the factory being a singleton will work. Unless it is able to create its own scope and dispose it at the right time, creating Disposable objects through the DI container will be problematic here.

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Data/FactoryExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        /// <summary>
+        /// Gets the options for a specific <seealso cref="TContext"/>.
+        /// </summary>
+        /// <param name="action">Option configuration action.</param>
+        /// <param name="sp">The scoped <see cref="IServiceProvider"/>.</param>
+        /// <returns>The newly configured <see cref="DbContextOptions{TContext}"/>.</returns>
+        private static DbContextOptions<TContext> GetOptions<TContext>(
+            Action<DbContextOptionsBuilder> action,
+                IServiceProvider sp = null) where TContext : DbContext
```
javiercn, line 50: nit: formatting

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Shared/ContactForm.razor
```diff
@@ -0,0 +1,142 @@
...
+        <DataAnnotationsValidator />
+        <ValidationSummary />
+        <h2>🆔 Identification</h2>
+        <p>
+            <label>
+                First Name:
+                <InputText @bind-Value="Contact.FirstName" />
+                <ConcurrencyField Model="@Contact"
+                                  DbModel="@DbContact"
+                                  Property="obj => obj?.FirstName" />
```
javiercn, line 19: Can you explain a bit what this is for?

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/ViewContact.razor
```diff
@@ -0,0 +1,196 @@
...
+
+    /// <summary>
+    /// Set to true when delete is requested.
+    /// </summary>
+    public bool ShowConfirmation = false;
+
+    /// <summary>
+    /// Set to true when delete is successful
+    /// </summary>
+    public bool Deleted = false;
```
javiercn, line 102: This looks like the state on the component, we don't recommend making this public

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/AddContact.razor
```diff
@@ -0,0 +1,113 @@
...
+        if (!success)
+        {
+            Success = false;
+            Error = false;
+            return;
+        }
+
+        Busy = true;
+
+        using var context = DbFactory.CreateDbContext();
```
javiercn, line 83: This will be problematic because the SP will hold on to the context after disposal. The factory needs to create its own scope and have a way of getting rid of the context.

I think it might be best to change the factory to have a Create/Release method pair and handle the disposal inside Release (disposing of the internal scope and the DbContext) or create a Disposable result time inside the factory that you new up in order to return the context and the scope in a way that the scope and associated context get disposed when the result type gets disposed.

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Data/DbContextFactory.cs
JeremyLikness: I'm not creating the DbContext through the service provider, that is just used to resolve the context options. It is using ActivatorUtilities which returns a new copy that isn't tracked by the service provider, as far as I understand.

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Data/FactoryExtensions.cs
JeremyLikness: I'm not sure I understand. It is using activator utilities to create a new instance and return it, so there are no references to the created contexts in either the factory or the SP ... am I missing something?

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Shared/ContactForm.razor
JeremyLikness: For concurrency resolution. If the field was changed by another user, it will show a warning box with the changed value. That lets the current user decide if they want to revert changes or force an update.

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Data/FactoryExtensions.cs
JeremyLikness: Will fix

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/AddContact.razor
JeremyLikness: See previous comments.

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/ViewContact.razor
JeremyLikness: Got it ... will review all components and update as needed.

JeremyLikness: @javiercn it is state but it manipulates the UI, wouldn't it make sense to keep public for testing? Otherwise should I make methods private too?

javiercn: @JeremyLikness If its public, it can be manipulated from outside the component, which will has the potential to break the component rendering model in the sense that you might expect a change in the property to trigger a re-render. We normally discourage such behaviors since a component should be responsible to decide when it needs to render itself.

Exposing something like that gives consumers the impression that they can take a `@ref` to the component and modify the state to cause a change, which is not the case.

The general recommendation is that components are explicit about what API they expose to other components.

JeremyLikness: Addressed in latest commit.

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Data/DbContextFactory.cs
JeremyLikness: @javiercn to test this I changed the factory to hold a collection of `WeakReference` to the db context. I then generated 50+ contexts and forced garbage collection and they all were deallocated. I also ran profiler and I do not show any lingering references. I can share the code if you like, just looking at what is remaining to finish this. Thanks!

guardrex:
@JeremyLikness ... Recommend hitting the **Resolve conversation** buttons ☝️ for the ones that are done. After any remaining concerns are resolved and closed, I'll take a look at the text for the grammar+style nits (we have a lot of little repo rules here).

javiercn:
@danroth27 @SteveSandersonMS Can you also help review this?

JeremyLikness:
Any updates?

guardrex:
I'm ready when they are. After engineering signoff, I'll make quick grammar+style pass, ping u to check those updates, then we can get this merged.

javiercn:
I've pinged @danroth27 on this and he's given a thumbs up, so I imagine he's going to help here.

aspnetcore/blazor/blazor-server-ef-core.md
```diff
@@ -0,0 +1,126 @@
...
+
+The factory is injected into components and used to create new instances. For example, in `Pages/Index.razor`:
+
+[!code-csharp[](./common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/Index.razor?range=199-212)]
+
+### Scope to the component lifetime
+
+There are two approaches to scope a context to the component lifetime. If the context is registered with DI, you can use the special `OwningComponentBase` base class to create a control scope. Learn more about the class [here](https://docs.microsoft.com/dotnet/api/microsoft.aspnetcore.components.owningcomponentbase-1).
+
+The second approach is to use the factory to create a context and track it for the lifetime of the component. First, implement `IDisposable` and inject the factory as shown in `Pages/EditContact.razor`.
```
danroth27, line 84: How do we want users to choose between these two approaches? Do we recommend one over the other? If yes, then should we just cover the approach that we recommend? I'm just trying to avoid confusing the user with choices that they might not know how to make. I'm not sure myself which pattern is preferred.

aspnetcore/blazor/blazor-server-ef-core.md
```diff
@@ -0,0 +1,126 @@
...
+
+[!code-csharp[](./common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/EditContact.razor?range=181-184)]
+
+Finally, `OnInitializedAsync` is overridden to create a new context. In the sample app, it loads the contact in the same method.
+
+[!code-csharp[](./common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/EditContact.razor?range=89-104)]
+
+## The sample app
+
+The sample application was built as a reference for Blazor Server apps the use EF Core. It includes a grid with sorting and filtering, delete, add, and update operations. It demonstrates use of EF Core to handle optimistic concurrency.
```
danroth27, line 98: > It includes a grid 

@SteveSandersonMS Since a bunch of folks will look at this sample, do we need to make sure this grid is aligned with the component authoring performance guidance you've been working on?

aspnetcore/blazor/blazor-server-ef-core.md
```diff
@@ -0,0 +1,126 @@
...
+
+[View or download sample code](https://github.com/dotnet/AspNetCore.Docs/tree/master/aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample) ([how to download](xref:index#how-to-download-a-sample))
+
+By default, the sample uses [SQL Server Express LocalDB](https://docs.microsoft.com/sql/database-engine/configure-windows/sql-server-express-localdb). You can override the connection string by updating the connection string in `appsettings.json`:
```
danroth27, line 102: Maybe we should switch this to SQLite so its xplat?

aspnetcore/blazor/blazor-server-ef-core.md
```diff
@@ -0,0 +1,126 @@
...
+
+The sample also configures database logging to show the SQL queries that are generated. This is configured in `appsettings.Development.json`. 
+
+[!code-json[](./common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/appsettings.Development.json?highlight=8)] 
+
+The `Startup.cs` class uses the factory extension to register the context and set the options. 
+
+[!code-csharp[](./common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Startup.cs?highlight=28-32)]
+
+To make it easier to populate the sample database, a call to ensure it is created and populate random sample contacts is made in `Program.cs`.
```
danroth27, line 114: I would remove this last section about database initialization sample code. You already have a nice comment in the sample pointing out that the database initialization code is just for demonstration purposes. I think calling this out in the doc actually draws more attention to it and creates more opportunity for folks to be confused by it.

aspnetcore/blazor/blazor-server-ef-core.md
```diff
@@ -0,0 +1,126 @@
...
+}
+```
+
+Another solution is to use a factory. EF Core 5.0 (preview) provides a built-in factory for creating new contexts. Because the sample app is based on the latest stable versions and not the preview, it implements its own factory in `Data/DbContextFactory.cs`. 
+
+[!code-csharp[](./common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Data/DbContextFactory.cs)]
+
+The factory is injected into components and used to create new instances. For example, in `Pages/Index.razor`:
+
+[!code-csharp[](./common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/Index.razor?range=199-212)]
```
danroth27, line 78: Should we show how to configure the factory as a service and how to inject it here? I know you can just look at the sample app, but it seems convenient to point out all of the specific code involved.

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/App.razor
```diff
@@ -0,0 +1,11 @@
+﻿
```
danroth27, line 1: Extra newline

aspnetcore/blazor/blazor-server-ef-core.md
```diff
@@ -0,0 +1,126 @@
...
+@code {
+    public Task LoadAsync(int id)
+    {
+        using var context = new MyContext(Options);
+        return context.MyEntities.SingleOrDefaultAsync(e => e.Id == id);
+    }
+}
+```
+
+Another solution is to use a factory. EF Core 5.0 (preview) provides a built-in factory for creating new contexts. Because the sample app is based on the latest stable versions and not the preview, it implements its own factory in `Data/DbContextFactory.cs`.
```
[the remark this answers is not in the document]
guardrex: Use the moniker range sections for each version of the topic. You can set them in immediately under the author byline ...

```

... 5.0+ version of the topic ...

::: moniker-end

::: moniker range="< aspnetcore-5.0"

... 3.1 version of the topic ...

::: moniker-end
```

WRT to code snippets: If you create a separate 5.0 sample (as opposed to just using code blocks in the document), continue the pattern we have in place for folder naming. We'll need a `common/samples/5.x` folder anyway for when I create the other 5.0 sample apps later. Don't sweat the release versions of the packages for 5.0 later. I'll take care of them when the time comes.

aspnetcore/blazor/blazor-server-ef-core.md
```diff
@@ -0,0 +1,126 @@
...
+
+[!code-csharp[](./common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Data/DbContextFactory.cs)]
+
+The factory is injected into components and used to create new instances. For example, in `Pages/Index.razor`:
+
+[!code-csharp[](./common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/Index.razor?range=199-212)]
+
+### Scope to the component lifetime
+
+There are two approaches to scope a context to the component lifetime. If the context is registered with DI, you can use the special `OwningComponentBase` base class to create a control scope. Learn more about the class [here](https://docs.microsoft.com/dotnet/api/microsoft.aspnetcore.components.owningcomponentbase-1).
```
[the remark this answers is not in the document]
guardrex:
```
<xref:Microsoft.AspNetCore.Components.OwningComponentBase>
```

aspnetcore/blazor/blazor-server-ef-core.md
guardrex: Note tho that code comments aren't *localized*. We've been moving virtually all code comments to in-text remarks for that reason.

JeremyLikness:
Ok @danroth27 a few changes, ready for review. I hope we're close!

- simplified guidance to recommend the factory
- changed the project to use SQLite so it can run anywhere
- created a 5.0 sample
- changed the text to be version specific

Jeremy

aspnetcore/blazor/blazor-server-ef-core.md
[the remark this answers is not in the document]
JeremyLikness: This is done.

aspnetcore/blazor/blazor-server-ef-core.md
```diff
@@ -0,0 +1,126 @@
...
+    finally 
+    {
+        Loading = false;
+    }
+    ```
+* For longer-lived operations that take advantage of EF Core's [change tracking](https://docs.microsoft.com/ef/core/querying/tracking) and/or [concurrency control](https://docs.microsoft.com/ef/core/saving/concurrency), scope the context to the lifetime of the component.
+
+### Create new DbContext instances
+
+The fastest way to create a new `DbContext` instance is by using `new` to create a new instance. There are several scenarios that may require additional dependencies to be resolved. For example, you may wish to use [DbContextOptions](https://docs.microsoft.com/ef/core/miscellaneous/configuring-dbcontext#configuring-dbcontextoptions) to configure the context. One approach is to register the options in DI and inject them into the component, like this:
```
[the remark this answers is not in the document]
JeremyLikness: Done.

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/App.razor
JeremyLikness: fixed

aspnetcore/blazor/blazor-server-ef-core.md
JeremyLikness: I included that earlier in the document.

aspnetcore/blazor/blazor-server-ef-core.md
JeremyLikness: I left it out for now.

aspnetcore/blazor/blazor-server-ef-core.md
JeremyLikness: Done.

aspnetcore/blazor/blazor-server-ef-core.md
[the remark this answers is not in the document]
JeremyLikness: I removed the reference altogether to avoid too many choices.

aspnetcore/blazor/blazor-server-ef-core.md
```diff
@@ -0,0 +1,126 @@
...
+    {
+        Loading = true;
+        // operations go here
+    }
+    finally 
+    {
+        Loading = false;
+    }
+    ```
+* For longer-lived operations that take advantage of EF Core's [change tracking](https://docs.microsoft.com/ef/core/querying/tracking) and/or [concurrency control](https://docs.microsoft.com/ef/core/saving/concurrency), scope the context to the lifetime of the component.
```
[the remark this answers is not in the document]
JeremyLikness: Done.

Rick-Anderson:
@guardrex  who should I ping to finish this?

guardrex:
@danroth27 (for approval of the last round of updates) and @SteveSandersonMS (to [answer the question](https://github.com/dotnet/AspNetCore.Docs/pull/18981#discussion_r463753743) ☝️)

I'll make my pass on it, then @JeremyLikness can make sure that I didn't break anything.

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Shared/GridWrapper.razor
```diff
@@ -0,0 +1,29 @@
+﻿<CascadingValue Value="this">
```
SteveSandersonMS, line 1: The following is important for perf:

```suggestion
﻿<CascadingValue Value="this" IsFixed="true">
```

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/Index.razor
```diff
@@ -0,0 +1,242 @@
...
+                <div class="col-12 alert alert-warning">
+                    No contacts found.
+                </div>
+            </div>
+        }
+        @if (Contacts != null)
+        {
+            @foreach (var contact in Contacts)
+            {
+                <ContactRow CurrentContact="@contact"
```
SteveSandersonMS, line 86: It's very valuable to use `@key` when rendering components inside a loop and the set can change over time. I'd recommend that here.

aspnetcore/blazor/common/samples/5.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/Index.razor
```diff
@@ -0,0 +1,242 @@
...
+        </div>
+        <div class="row">&nbsp;</div>
+        <div class="row">
+            <div class="col-6">
+                Page @Filters.PageHelper.Page of @Filters.PageHelper.PageCount:
+                showing @Filters.PageHelper.PageItems of
+                @Filters.PageHelper.TotalItemCount items.
+                <a disabled="@(Filters.Loading || !Filters.PageHelper.HasPrev)"
+                   class="btn btn-primary @IsDisabled(Filters.PageHelper.HasPrev)"
+                   href="/@Filters.PageHelper.PrevPage">
```
SteveSandersonMS, line 28: Why the leading slash on the URLs like this? Does it work OK without that? I expect so since it's resolving relative to the base href. If so it's better not to put on the leading slash, as that would break any case where this was deployed to a non-root URL.

aspnetcore/blazor/common/samples/5.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/Index.razor
```diff
@@ -0,0 +1,242 @@
...
+@using Microsoft.EntityFrameworkCore
+
+@inject IContactFilters Filters
+@inject IDbContextFactory<ContactContext> DbFactory
+@inject NavigationManager Nav
+@inject GridQueryAdapter QueryAdapter
+
+<h1>Contacts</h1>
+<GridWrapper @ref="Wrapper"
+             FilterChanged="async () => await ReloadAsync()"
```
SteveSandersonMS, line 13: Can this be simplified?

```suggestion
             FilterChanged="ReloadAsync"
```

aspnetcore/blazor/blazor-server-ef-core.md
SteveSandersonMS: Since this is inherently Blazor Server-only, it's likely that perf will look good regardless. However in case people copy these patterns for WebAssembly I've posted a couple of suggestions that will help perf.

aspnetcore/blazor/blazor-server-ef-core.md
```diff
@@ -0,0 +1,150 @@
...
+
+:::moniker range=">= aspnetcore-5.0"
+The recommended solution to create a new `DbContext` with dependencies is to use a factory. EF Core 5.0 provides a built-in factory for creating new contexts.
+:::moniker-end
+:::moniker range="< aspnetcore-5.0"
+The recommended solution to create a new `DbContext` with dependencies is to use a factory. The sample app implements its own factory in `Data/DbContextFactory.cs`. 
+
+[!code-csharp[](./common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Data/DbContextFactory.cs)]
+:::moniker-end
+
```
danroth27, line 83: Since the factory is introduced in the second paragraph, I think the bit about registering the factory should be moved afterwards.

```suggestion
:::moniker range=">= aspnetcore-5.0"
The recommended solution to create a new `DbContext` with dependencies is to use a factory. EF Core 5.0 provides a built-in factory for creating new contexts.
:::moniker-end
:::moniker range="< aspnetcore-5.0"
The recommended solution to create a new `DbContext` with dependencies is to use a factory. The sample app implements its own factory in `Data/DbContextFactory.cs`. 

[!code-csharp[](./common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Data/DbContextFactory.cs)]
:::moniker-end

Here is an example that configures SQLite and enables data logging. The code uses an extension method to configure the database factory for dependency injection and provide the default options:

:::moniker range=">= aspnetcore-5.0"
[!code-csharp[](./common/samples/5.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Startup.cs?range=29-31)]
:::moniker-end
:::moniker range="< aspnetcore-5.0"
[!code-csharp[](./common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Startup.cs?range=29-31)]
:::moniker-end

```

aspnetcore/blazor/blazor-server-ef-core.md
```diff
@@ -0,0 +1,150 @@
...
+:::moniker range="< aspnetcore-5.0"
+[!code-csharp[](./common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/EditContact.razor?range=89-104)]
+:::moniker-end
+:::moniker range=">= aspnetcore-5.0"
+[!code-csharp[](./common/samples/5.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/EditContact.razor?range=89-104)]
+:::moniker-end
+
+## The sample app
+
+The sample application was built as a reference for Blazor Server apps the use EF Core. It includes a grid with sorting and filtering, delete, add, and update operations. It demonstrates use of EF Core to handle optimistic concurrency.
```
danroth27, line 125:
```suggestion
The sample application was built as a reference for Blazor Server apps that use EF Core. It includes a grid with sorting and filtering, delete, add, and update operations. It demonstrates use of EF Core to handle optimistic concurrency.
```

aspnetcore/blazor/blazor-server-ef-core.md
JeremyLikness: Thanks for the review!

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Shared/GridWrapper.razor
JeremyLikness: Done.

aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/Index.razor
JeremyLikness: Great suggestion. Done.

aspnetcore/blazor/common/samples/5.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/Index.razor
JeremyLikness: I changed these to avoid leading slash and tested - appears to work fine.

aspnetcore/blazor/common/samples/5.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/Index.razor
JeremyLikness: Simplified this throughout the codebase.

guardrex:
@JeremyLikness I'll get on this Friday morning for a quick UE pass ... grammar nits and bits. I'll ping you for a final look before merging tomorrow.

guardrex:
@JeremyLikness ...

* We use "whole-topic versioning" by convention, so I made that update here. You'll see two versions of the whole topic now on the PR. This approach can be a bit of a pain at times when a change needs to be made in multiple spots, but it's dead easy to make mods for a single version, add a new version, and drop versions when a framework version goes out of support.
* We don't place (many) remarks in samples due to a localization constraint ... code comments don't get translated (yet). I'll open a follow-up issue to deal with that later.
* Some of these updates are **_merely suggestions_**, but it's so much faster for me to float them on the commit than to place a gazillion GH suggestions on the PR. You can reverse something that you feel was better as long as it doesn't break one of the repo conventions and bring down the wrath of the :pray: *doc server gods* :pray: on us. :smile:
* We usually link and describe the sample app earlier. Readers may pull it down immediately and then refer to it as they read the topic. I moved it up. I left it in a section has you had it.
* I'll work [Drop EF Core content from the Blazor DI topic (dotnet/AspNetCore.Docs #19051)](https://github.com/dotnet/AspNetCore.Docs/issues/19051) as soon as we merge this.

Finally, there are other review requests without response. My inclination is to go ahead and get this merged. Do you want to re-ping others and wait for them to respond? :ear:

I won't merge this until I've heard back that you're ready to go.

JeremyLikness:
@guardrex I've addressed all of the comments I'm aware of. If there are additional I say let's get this published and we can update it as needed. Thanks!

guardrex:
I agree. This one needs to be surfaced asap. I'll shoot it in now, merge to live, and inspect it on the live site just to make sure that it's all good. I'll work that other issue sometime today.

PR #19012, opened 2020-06-30
gRPC-Web with bidirectional streaming on IIS/Azure

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/963

### DIFF
--- a/aspnetcore/grpc/browser.md
+++ b/aspnetcore/grpc/browser.md
@@ -69,7 +69,10 @@ The preceding code:
 
 ### gRPC-Web and streaming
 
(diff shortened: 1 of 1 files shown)


guardrex:
@JamesNK ... Were the following intended?

> Traditional gRPC over HTTP/2 supports streaming in all directions. gRPC-Web offers limited support for streaming:

Should "streaming in all directions" be "bidirectional streaming"?

> gRPC-Web browser clients don't support calling client streaming and bidirection streaming methods.

Should "bidirection" be "bidirectional"?

JamesNK:
> Should "bidirection" be "bidirectional"?

Yes

PR #19047, opened 2020-07-02
Update Warning section for in Enable Cors

xboxeer:
according to this github issue reply https://github.com/dotnet/aspnetcore/issues/23004#issuecomment-647755359, I think it is better to warn our developer that usecors need to be placed between UseRouter and UseAuthentication. Typically customer will directly goes to this cors page of asp.net core docs and find it should be placed before UseResponseCache, but the UseRouter and UseAuthentication sequence is as same import as UseResponseCache, so it would be better to add them in warining as well

### DIFF
--- a/aspnetcore/security/cors.md
+++ b/aspnetcore/security/cors.md
@@ -54,7 +54,7 @@ There are three ways to enable CORS:
 Using the [[EnableCors]](#attr) attribute with a named policy provides the finest control in limiting endpoints that support CORS.
 
 > [!WARNING]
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
In the same doc we have:
The call to UseCors must be placed after UseRouting, but before UseAuthorization. For more information, see Middleware order.

The change you made is to a temporary bug work around. That comment will go when the bug is fixed.

Rick-Anderson:
@xboxeer  I would have taken this PR after submitting my own commit, but you declined that option.  It should just have appended 

`UseCors` currently must go before `UseResponseCaching` due to [this bug](https://github.com/dotnet/aspnetcore/issues/23218).

PR #19056, opened 2020-07-02
Clean up the Static Files sample

serpent5:
cc @Rick-Anderson


aspnetcore/fundamentals/static-files/samples/3.x/StaticFilesSample/Controllers/HomeController.cs
```diff
@@ -28,6 +18,17 @@ public IActionResult Privacy()
...
+        #region snippet_BannerImage
+        [Authorize]
+        public IActionResult BannerImage()
+        {
+            var filePath = Path.Combine(
+                Directory.GetCurrentDirectory(), "MyStaticFiles", "images", "banner1.svg");
+
+            return PhysicalFile(filePath, "image/svg+xml");
+        }
+        #endregion
```
serpent5, line 30: The 3.x version was referencing the 1.x version of the code here. There's no notable difference between the versions, but adding this here means the 3.x stuff is more self-contained.

serpent5:
The <xref:...> changes are blind, but I used the epic xrefgen, so 🤞 it's all fine. It got me thinking it would be pretty awesome if a VS Code extension could consume the xrefgen as an API and make things a bit easier there.

aspnetcore/fundamentals/static-files/samples/3.x/StaticFilesSample/Controllers/HomeController.cs
Rick-Anderson: Per `@Tratcher ` at 
Yeah, `WebRootPath `and `ContentRootPath `are better than `GetCurrentDirectory`, but they're better because the app can configure them, not because of any xpat issue with `GetCurrentDirectory`. `ContentRootPath `is set to `GetCurrentDirectory `by default.

Can we do the same here? Also, change it to *rose.jpg* 
I'm using so we can test.

serpent5: Done. The `.min` files don't exist so I fixed those up too. There's no authz configuration in the sample, but I verified that it works by commenting out `[Authorize]`.

Rick-Anderson:
I reviewed the build and it looks good. Can I S&M?

serpent5:
Yeah, I'm finished with my changes.

PR #19156, opened 2020-07-09
Sample code of resolve services from ConfigureService method

xboxeer:
We always have customer asking how to resolve a service in ConfigureSerivice method or related question. Our docs currently does not covers this part, only a section of how to resolve service in main. Sometimes these question will be redirected to this stackoverflow post https://stackoverflow.com/questions/31863981/how-to-resolve-instance-inside-configureservices-in-asp-net-core .  But the sample code in that thread misses a using and is binding to the context of the original post, so I did some optimize to the sample code and hope this can be added to our documents

### DIFF
--- a/aspnetcore/fundamentals/dependency-injection.md
+++ b/aspnetcore/fundamentals/dependency-injection.md
@@ -382,6 +382,34 @@ public class Program
             });
 }
 ```
+## Resolve services in ConfigureService method
+
+In some cases you may need to resolve a service in ConfigureService method, for example, reading a setting from database then decide which service to register. Typically, in asp.net core, the most well known approach to resolve a service is constructor injection. However, in this stage, this approach is not available. So, we can create a service provider in ConfigureService, then resolve the service that we want by utilizing that service provider.
+
+```csharp
+public void ConfigureService(IServiceCollection services)
+{
+    // Configure the services
+    services.AddTransient<ISettingService, DBSettingService>();    
+
+    // Build an intermediate service provider
+    using (var serviceProvider = services.BuildServiceProvider())
+    {
+        // Resolve the services from the service provider
+        var settingService = serviceProvider.GetRequiredService<ISettingService>();
+        // Use your services to decide if FooService and BarService should be injected
+        if(settingService.EnableFoo())
+        {
+            services.AddScoped(IFooService,FooService);
+        }
+        if(settingService.EnableBar())
+        {
+            services.AddScoped(IBarService,BarService);
+        }
+    };
+
+}
+```
 
 ## Scope validation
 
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/AspNetCore.Docs?pullRequest=19156) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/AspNetCore.Docs?pullRequest=19156) before we can accept your contribution.<br/><br/>:x: xboxeer [sign now](https://cla.dotnetfoundation.org/dotnet/AspNetCore.Docs?pullRequest=19156)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/AspNetCore.Docs?pullRequest=19156) it.</sub>

scottaddie:
@mkArtakMSFT Can someone on your team please review this PR?

serpent5:
Can I help with this one? I know that's not nearly the same as someone from the team, but I know those guys are _very_ busy.

scottaddie:
@serpent5 Yes, please.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -382,6 +382,34 @@ public class Program
...
+        var settingService = serviceProvider.GetRequiredService<ISettingService>();
+        // Use your services to decide if FooService and BarService should be injected
+        if(settingService.EnableFoo())
+        {
+            services.AddScoped(IFooService,FooService);
+        }
+        if(settingService.EnableBar())
+        {
+            services.AddScoped(IBarService,BarService);
+        }
```
serpent5, line 408: I'd like to try and help suggest an alternative approach for this, but I'm a bit confused by the intent here. If `IFooService` is registered only if `EnableFoo` returns `true`, any services, controllers, etc that depend on `IFooService` are going to see runtime exceptions, because `IFooService` won't have been registered at all. This is similar for `EnableBar` and `IBarService`.

xboxeer: @serpent5 Very appericate for your comments, I understand your concern that register different services based on different setting may causes runtime exceptions if a service that not registered is called. And causing multiple singleton instance never comes to my mind before, thanks for point me out through the links. 
This user case may not be best approprate. I can name another senario, here is a sample [thread](https://stackoverflow.com/questions/62050031/using-dependency-injection-to-configure-cookieauthenticationoptions-in-aspnetcor), when configuring CookieAuthentication, developers want to get some settings from database, and that requires a service that registered in DI.
But anyway, if this is a not recommanded approach, it is better that our docs can cover it or we can have a proper solution to get DI instance in configservice method

davidfowl:
Yes this isn't something that should be in the docs. Creating a second container can create torn singletons and cause references to object graphs across multiple containers. 

People shouldn't be copy and pasting this from the docs and if you can figure it out and manage the downsides then that's great.

aspnetcore/fundamentals/dependency-injection.md
serpent5: Authentication uses the options pattern, which has built-in support for DI and is covered in the [docs](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-3.1#use-di-services-to-configure-options). Here's an example of how this might look:

```csharp
services.AddAuthentication(CookieAuthenticationDefaults.AuthenticationScheme)
    .AddCookie();

services.AddOptions<CookieAuthenticationOptions>(CookieAuthenticationDefaults.AuthenticationScheme)
    .Configure<IFooService>((options, fooService) =>
    {
        // Use fooService to configure options.
        // ...
    });
```

In this authentication case, the name of the options being configured is the name of the authentication scheme.

scottaddie:
Based on David's response, I'm closing this PR as something we don't want included in the docs.

xboxeer:
> Yes this isn't something that should be in the docs. Creating a second container can create torn singletons and cause references to object graphs across multiple containers.
> 
> People shouldn't be copy and pasting this from the docs and if you can figure it out and manage the downsides then that's great.

Make sense, since this is a not good practice in DI, how abt we add a this in [DI Recommendations](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-3.1#recommendations) to advise develop not doing the same? I can help raise another PR for this.

davidfowl:
Yes that sounds like it's good for an anti-patterns example

PR #19165, opened 2020-07-09
Update routing.md

mrlife:
Added note on limitation of looping over the NavLink component


guardrex:
@mrlife ... Thanks, but this will likely require more context. Also, we don't want to reference engineering issues in the docs except for temporary scenarios that will be resolved at some future point. I opened https://github.com/dotnet/AspNetCore.Docs/issues/19166 to address it.

PR #19133, opened 2020-07-07
Sweep up a few typos

serpent5:
Fixes #18988.
Fixes #19096.

PR #19167, opened 2020-07-09
Fix up dev-langs (cross-repo)

serpent5:
Following on from my previous PR, #19133, I decided to build and run a script that ran against the repo to check if there were any other devlangs that weren't right.

I don't necessarily expect this to be merged, but I thought it might be a useful exercise to show where the changes are needed.

cc @scottaddie @Rick-Anderson


serpent5:
For reference, here's the map I used:

```
.csproj = xml
.cs = csharp
.cshtml = cshtml
.razor = razor
.json = json
.js = javascript
.min.js = javascript
.css = css
.ts = typescript
.html = html
.xml = xml
.ini = ini
.pubxml = xml
.proto = protobuf
.yml = yaml
```

scottaddie:
@serpent5 Thank you for fixing these!

PR #19207, opened 2020-07-15
Fix link

guardrex:
This change saves one 301 redirect.


serpent5:
I'll leave these alone in future. 🤭

PR #19213, opened 2020-07-15
Update Blazor debug for hosted config and Edge

### LINKED ISSUE #19204: [VSCode] Add in details on how to debug blazorwasm when project structure is nested
Would love to have details on how to enable debugging when the project structure is not the default in VSCode

For Example:
/
/.vs
/src
/src/Blazor.App
/src/Blazor.App/Client
/src/Blazor.App/Server
/src/Blazor.App/Shared

The default setup will not work for a blazorwasm hosted app
#19204 https://github.com/dotnet/aspnetcore/issues/21166#issuecomment-632933868
---
#### Document Details

* ID: ec2f61c7-9f51-6735-e328-3c7b495d70c3
* Version Independent ID: 4d06e565-282d-6387-aa90-cb2c2986d35a
* Content: [Debug ASP.NET Core Blazor WebAssembly](https://docs.microsoft.com/en-us/aspnet/core/blazor/debug?view=aspnetcore-3.1)
* Content Source: [aspnetcore/blazor/debug.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/debug.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #19204

I don't think we need to say more than what the config options explain. The bits work OOB as described (with the touch-ups on this PR).

### DIFF
--- a/aspnetcore/blazor/debug.md
+++ b/aspnetcore/blazor/debug.md
@@ -5,7 +5,7 @@ description: Learn how to debug Blazor apps.
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 07/06/2020
+ms.date: 07/15/2020
 no-loc: [Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
 uid: blazor/debug
 ---
@@ -35,8 +35,8 @@ We will continue to improve the debugging experience in upcoming releases.
 
 Debugging requires either of the following browsers:
 
+* Google Chrome (version 70 or later) (default)
 * Microsoft Edge (version 80 or later)
-* Google Chrome (version 70 or later)
 
 ## Enable debugging for Visual Studio and Visual Studio Code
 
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/debug.md
```diff
@@ -185,18 +185,24 @@ The following launch configuration options are supported for the `blazorwasm` de
 }
 ```
 
-#### Launch and debug a hosted Blazor WebAssembly app
+#### Launch and debug a hosted Blazor WebAssembly app with Microsoft Edge
+
+Browser configuration defaults to Google Chrome. When using Microsoft Edge for debugging, set `browser` to `edge`. To use Google Chrome, don't set `browser` or set it to `chrome`.
```
captainsafia, line 190:
```suggestion
Browser configuration defaults to Google Chrome. When using Microsoft Edge for debugging, set `browser` to `edge`. To use Google Chrome, unset the `browser` property in the configuration.
```

guardrex: It's not config that the dev would unset if it weren't there in the first place. Also, "unset" (or even "don't set" in the original come to think of it), makes it sound like an empty string is a valid value, which it isn't.

```suggestion
Browser configuration defaults to Google Chrome. When using Microsoft Edge for debugging, set `browser` to `edge`. To use Google Chrome, either don't set the `browser` option or set the option's value to `chrome`.
```

Kronos11:
This is good, and maybe I'm doing something wrong, but with the above structure I have the following launch.json and it's not attaching debugging correctly:`{
      "type": "blazorwasm",
      "request": "launch",
      "name": "Launch and Debug Hosted Blazor WebAssembly App",
      "program": "${workspaceFolder}/src/App/Server/bin/Debug/netcoreapp3.1/App.Server.dll",
      "cwd": "${workspaceFolder}/src/App/Server",
      "trace": "verbose",
      "hosted": true,
      "browser" : "edge"
    }`

guardrex:
We ........ and when I say *"we"* I mean @captainsafia :smile: lol ...... might be able to work it out for your scenario; but as you suggest, the topic can only show the base case.

If it can't be resolved quickly here ... i.e., requires investigation/testing/etc. ... then I recommend opening an engineering issue.

PR #19231, opened 2020-07-16
Blazor scoped service registrations

### LINKED ISSUE #18359: [3.1.6] Incorrect HTTPClient registration
On this page and probably in a few other places (for example https://docs.microsoft.com/en-us/aspnet/core/security/blazor/webassembly/additional-scenarios?view=aspnetcore-3.1) HTTPClient is registered as `Transient` which is wrong
```c#
builder.Services.AddTransient(sp => 
    new HttpClient
    {
        BaseAddress = new Uri(builder.HostEnvironment.BaseAddress)
    });
```
See this https://github.com/dotnet/aspnetcore/issues/21655
It looks that project template issue will be addressed in 5.0 but maybe you can make corrections in the doc earlier.

---
#### Document Details

* Content: [Call a web API from ASP.NET Core Blazor WebAssembly](https://docs.microsoft.com/en-us/aspnet/core/blazor/call-web-api?view=aspnetcore-3.1)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #18359

Now that 3.1.6 is out, this updates service registrations from transient to scoped ...

* `HttpClient`
* `IHttpClientFactory`
* `CustomAuthorizationMessageHandler` (`AuthorizationMessageHandler`)

### DIFF
--- a/aspnetcore/blazor/common/samples/3.x/BlazorWebAssemblySample/Program.cs
+++ b/aspnetcore/blazor/common/samples/3.x/BlazorWebAssemblySample/Program.cs
@@ -15,7 +15,7 @@ namespace BlazorSample
             var builder = WebAssemblyHostBuilder.CreateDefault(args);
             builder.RootComponents.Add<App>("app");
 
-            builder.Services.AddSingleton(new HttpClient { BaseAddress = new Uri(builder.HostEnvironment.BaseAddress) });
+            builder.Services.AddScoped(new HttpClient { BaseAddress = new Uri(builder.HostEnvironment.BaseAddress) });
 
             await builder.Build().RunAsync();
         }
--- a/aspnetcore/blazor/fundamentals/configuration.md
+++ b/aspnetcore/blazor/fundamentals/configuration.md
@@ -126,7 +126,7 @@ var client = new HttpClient()
     BaseAddress = new Uri(builder.HostEnvironment.BaseAddress)
 };
 
-builder.Services.AddTransient(sp => client);
+builder.Services.AddScoped(sp => client);
 
 using var response = await client.GetAsync("cars.json");
 using var stream = await response.Content.ReadAsStreamAsync();
--- a/aspnetcore/tutorials/signalr-blazor-webassembly/samples/3.x/BlazorSignalRApp/Client/Program.cs
+++ b/aspnetcore/tutorials/signalr-blazor-webassembly/samples/3.x/BlazorSignalRApp/Client/Program.cs
@@ -15,7 +15,7 @@ namespace BlazorSignalRApp.Client
             var builder = WebAssemblyHostBuilder.CreateDefault(args);
             builder.RootComponents.Add<App>("app");
 
-            builder.Services.AddSingleton(new HttpClient { BaseAddress = new Uri(builder.HostEnvironment.BaseAddress) });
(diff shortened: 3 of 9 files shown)


aspnetcore/blazor/fundamentals/configuration.md
```diff
@@ -126,7 +126,7 @@ var client = new HttpClient()
     BaseAddress = new Uri(builder.HostEnvironment.BaseAddress)
 };
 
-builder.Services.AddTransient(sp => client);
+builder.Services.AddScoped(sp => client);
```
serpent5, line 129: This one still seems strange to me.

I'll bet I'm just missing something, and I understand that scopes are a little different in Blazor WASM, but the `client` instance is essentially a singleton in this case, yet is being exposed as though it's being created per-scope. In the case where the owning scope was disposed, `client` would also be disposed and continue to be given back for other scopes.

I know it _works_,  but I just wanted to mention this in passing.

aspnetcore/blazor/common/samples/3.x/BlazorWebAssemblySample/Program.cs
```diff
@@ -15,7 +15,7 @@ public static async Task Main(string[] args)
             var builder = WebAssemblyHostBuilder.CreateDefault(args);
             builder.RootComponents.Add<App>("app");
 
-            builder.Services.AddSingleton(new HttpClient { BaseAddress = new Uri(builder.HostEnvironment.BaseAddress) });
+            builder.Services.AddScoped(new HttpClient { BaseAddress = new Uri(builder.HostEnvironment.BaseAddress) });
```
serpent5, line 18: @guardrex This change means the code will no longer compile because `AddScoped` can't be given a singleton instance.

aspnetcore/tutorials/signalr-blazor-webassembly/samples/3.x/BlazorSignalRApp/Client/Program.cs
```diff
@@ -15,7 +15,7 @@ public static async Task Main(string[] args)
             var builder = WebAssemblyHostBuilder.CreateDefault(args);
             builder.RootComponents.Add<App>("app");
 
-            builder.Services.AddSingleton(new HttpClient { BaseAddress = new Uri(builder.HostEnvironment.BaseAddress) });
+            builder.Services.AddScoped(new HttpClient { BaseAddress = new Uri(builder.HostEnvironment.BaseAddress) });
```
serpent5, line 18: @guardrex Same here.

aspnetcore/blazor/common/samples/3.x/BlazorWebAssemblySample/Program.cs
javiercn: That's true, I overlooked this. It has to be `builder.Services.AddScoped<HttpClient>(sp => new HttpClient...`

guardrex: Find-'n-replace ......... *It's GrEaT!* :smile: ........... **_Until it isn't!_** 🙈

I'll patch these now and take the update live immediately.

Issue #18869, opened 2020-06-19
Add common anti-patterns samples and outcomes


davidfowl:
There are a set of patterns that can cause bad behavior when using the DI container (I reference some of them in my talk here https://youtu.be/x-C-CNBVTaY?t=1628 code samples here 
 https://github.com/davidfowl/NDCLondon2017/blob/417394eaa2c764eb9ed85b64a1dde0574068eea2/DependencyInjectionDemos/Program.cs)

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0d78f540-81fd-1ba9-cb90-591e4ec27d9b
* Version Independent ID: 2251f557-dd3a-ec35-2558-6243337db009
* Content: [Dependency injection in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-3.1#design-services-for-dependency-injection)
* Content Source: [aspnetcore/fundamentals/dependency-injection.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/fundamentals/dependency-injection.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-fundamentals**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
Moved to https://github.com/dotnet/docs/issues/21229

PR #19271, opened 2020-07-21
DI rewrite

### LINKED ISSUE #18480: Add DI sample to options pattern
And anything else that needs to be fixed

cc @serpent5 

---
#### Document Details

* Content: [Options pattern in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-3.1&branch=pr-en-us-18335)
* Content Source: [aspnetcore/fundamentals/configuration/options.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/fundamentals/configuration/options.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

### LINKED ISSUE #19422: Total rewrite of DI article
---
#### Document Details

* Content: [Dependency injection in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-3.1#design-services-for-dependency-injection)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

Rick-Anderson:
Fixes #19422  and some of  #18869
Fixes #18480

To review the doc build 
- download https://www.dropbox.com/sh/yy994551cje0kap/AAB0tRnm7gze2kYrSYPSFHIda?dl=0
- unzip
- view HTML file in FF (Fire Fox). It best to view the doc with the widest window that has no left or right TOC (Table  of contents)


aspnetcore/fundamentals/dependency-injection/samples/3.x/DependencyInjectionSample2/Models/Operation.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace DependencyInjectionSample.Models
+{
+    #region snippet1
+    public class Operation : IOperationTransient, 
+        IOperationScoped, 
+        IOperationSingleton, 
+        IOperationSingletonInstance
+    {
+        public Operation() : this(DateTime.Now.ToString("hh:mm:ss.fff"))
```
Rick-Anderson, line 12: @serpent5  I changed this in my proposed version to use a time string so you can easily see what's updated when you refresh the screen. Forced to use `DoWork` so we get different times. Impossible for the human eye to compare GUIDs but time is easy.

Rick-Anderson: @dougbu  do you approve of me changing from GUID to time string? It makes it easy to see what has changed. The down side is you end up with `DoWork`, but I can explain that.

aspnetcore/fundamentals/dependency-injection/samples/3.x/DependencyInjectionSample/Models/Operation.cs
```diff
@@ -4,21 +4,32 @@
...
         IOperationSingletonInstance
     {
-        public Operation() : this(Guid.NewGuid())
+        public Operation() : this(DateTime.Now.ToString("hh:mm:ss.fff"))
         {
+            DoWork();
         }
 
-        public Operation(Guid id)
+        public Operation(string id)
```
Rick-Anderson, line 17: @dougbu  here is the change from GUID to string. This makes it easy to see what has changed and what stays the same.

aspnetcore/fundamentals/dependency-injection/samples/3.x/DependencyInjectionSample/Models/Operation.cs
```diff
@@ -4,21 +4,32 @@
...
         {
             OperationId = id;
         }
 
-        public Guid OperationId { get; private set; }
```
Rick-Anderson, line 21: @dougbu @serpent5  I propose we move all these operations to the view. That's more realistic to show scope has the same value per request but transient doesn't. I could even add a middleware. 

Should I add a middleware? That seems more realistic.

dougbu:
@Rick-Anderson please reach out to someone else to answer your questions e.g. @davidfowl or @Tratcher. I'm not expecting to have time to help.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -199,12 +209,17 @@ In apps that process requests, scoped services are disposed at the end of the re
...
+Singleton lifetime services (<xref:Microsoft.Extensions.DependencyInjection.ServiceCollectionServiceExtensions.AddSingleton*>) are created either:
 
-In apps that process requests, singleton services are disposed when the <xref:Microsoft.Extensions.DependencyInjection.ServiceProvider> is disposed at app shutdown.
+* The first time they're requested.
+* When `Startup.ConfigureServices` is run and an instance is specified with the service registration.
+<!-- REVIEW what is user code? -->
+Every subsequent request uses the same instance. If the app requires singleton behavior, allowing the service container to manage the service's lifetime is recommended. Don't implement the singleton design pattern and provide user code to manage the object's lifetime in the class.
```
serpent5, line 217: I think _user code_ here means implementing the Singleton pattern manually. e.g. using one of [Jon Skeet's patterns](https://csharpindepth.com/articles/singleton).

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -421,7 +437,7 @@ Best practices are to:
 If a class seems to have too many injected dependencies, this is generally a sign that the class has too many responsibilities and is violating the [Single Responsibility Principle (SRP)](/dotnet/standard/modern-web-apps-azure-architecture/architectural-principles#single-responsibility). Attempt to refactor the class by moving some of its responsibilities into a new class. Keep in mind that Razor Pages page model classes and MVC controller classes should focus on UI concerns. Business rules and data access implementation details should be kept in classes appropriate to these [separate concerns](/dotnet/standard/modern-web-apps-azure-architecture/architectural-principles#separation-of-concerns).
 
 ### Disposal of services
-
+<!-- REVIEW: define user code.  -->
```
serpent5, line 440: I think this means something like this:

```csharp
var mySingleton = new MySingleton();
services.AddSingleton(mySingleton);
```
`mySingleton` isn't disposed by the container because _we_ provided the instance directly.

However, these two approaches result in automatic disposal:

```csharp
services.AddSingleton<MySingleton>();
services.AddSingleton(() => new MySingleton()); // Factory.
```

This one's more interesting:

```csharp
var mySingleton = new MySingleton();
services.AddSingleton(() => mySingleton);
```

This also causes automatic disposal, because the container assumes you created an instance on its behalf. I don't think you should ever really do this, though. It'd be dangerous if you tried to use `mySingleton` after the container was disposed.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -466,7 +482,9 @@ public void ConfigureServices(IServiceCollection services)
 **Scenario**
 
 The app requires an <xref:System.IDisposable> instance with a transient lifetime for either of the following scenarios:
-
+<!-- Review need to define root scope. Is that the same as the root container that's only disposed when the app shuts down?
```
serpent5, line 485: This is how I understand it.

aspnetcore/fundamentals/dependency-injection/samples/3.x/DependencyInjectionSample2/Models/Operation.cs
serpent5: I think I prefer the GUID approach for this because it avoids the `DoWork` problem and I expect developers are used to GUIDs. Here's an example of some formatting changes that _might_ look a bit better with GUIDs:

image

aspnetcore/fundamentals/dependency-injection/samples/3.x/DependencyInjectionSample/Models/Operation.cs
serpent5: I'm not sure I understand the suggestion here. Can you explain it a bit more, please?

Rick-Anderson: The code behind calles each service twice. The `Operation` code has the same 3 services. I think it would be  better to show the output in the logger, not a view. Move the  `Operation` code  to middleware. It's just not realistic to call the same services twice each in a RP code behind. I was thinking it would be slightly better to move that to the RP view page, but I think moving it to middleware or a filter is better.

aspnetcore/fundamentals/dependency-injection/samples/3.x/DependencyInjectionSample2/Models/Operation.cs
Rick-Anderson: How about using a 3 digit random number and moving the **Operations service** to middleware - not showing this in the view, showing it in the logger.

Rick-Anderson: Or I could stick to GUID

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -84,15 +84,20 @@ This interface is implemented by a concrete type, `MyDependency`:
 The container resolves `ILogger<TCategoryName>` by taking advantage of [(generic) open types](/dotnet/csharp/language-reference/language-specification/types#open-and-closed-types), eliminating the need to register every [(generic) constructed type](/dotnet/csharp/language-reference/language-specification/types#constructed-types):
 
 ```csharp
+// This code provided by the framework.
 services.AddSingleton(typeof(ILogger<>), typeof(Logger<>));
```
davidfowl, line 88: Maybe we should remove this as an example since it's never relevant in practice

Rick-Anderson: Will do.
BTW, this is not ready for review but comments like that are welcome.

aspnetcore/fundamentals/dependency-injection.md
halter73: I think "user code" in this case is referring code that might incorrectly manually dispose a resolved singleton.

Services should never be disposed by any code that resolved the service from a container. If a type or factory is registered as a singleton, the container will dispose the singleton for you.

If an object instance is registered as a singleton, the code that created that instance is responsible for the lifetime, but the instance still must outlive the container. A reference to the singleton instance should be kept outside of the container. The singleton should then be disposed via this external reference after the container/host has already been disposed.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -158,69 +179,85 @@ The `Startup.ConfigureServices` method is responsible for defining the services
...
 
 Scoped lifetime services (<xref:Microsoft.Extensions.DependencyInjection.ServiceCollectionServiceExtensions.AddScoped*>) are created once per client request (connection).
 
+When using Entity Framework Core, the [AddDbContext](/dotnet/api/microsoft.extensions.dependencyinjection.entityframeworkservicecollectionextensions.adddbcontext) extension method registers `DbContext` types with a scoped lifetime by default.
+
 In apps that process requests, scoped services are disposed at the end of the request.
 
-> [!WARNING]
-> When using a scoped service in a middleware, inject the service into the `Invoke` or `InvokeAsync` method. Don't inject via [constructor injection](xref:mvc/controllers/dependency-injection#constructor-injection) because it forces the service to behave like a singleton. For more information, see <xref:fundamentals/middleware/write#per-request-middleware-dependencies>.
+Use scoped services in a middleware with one of the following approaches:
```
Rick-Anderson, line 215: @rahulpnath please review this change. I'm rewriting the entire doc and should be finished in a couple days. WHen I'm done, it would be great if you could review the updates. I can post a link to the new doc build.
I dropped the warning because this is not a trap you can fall into. The runtime throws and exception if you try to do this.

rahulpnath: Looks good and better now!

aspnetcore/fundamentals/dependency-injection/samples/3.x/DIsample2/DIsample2/Services/Service1.cs
```diff
@@ -0,0 +1,85 @@
...
+
+namespace DIsample2.Services
+{
+    public class Service1 : IDisposable
+    {
+        private bool disposedValue;
+
+        protected virtual void Dispose(bool disposing)
+        {
+            Console.WriteLine("Service1.Dispose");
```
Rick-Anderson, line 14: @serpent5  take a look at the section in the doc [Disposal of services](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-3.1&branch=pr-en-us-19271#disposal-of-services)

I've implemented this in aspnetcore/fundamentals/dependency-injection/samples/3.x/DIsample2 but  `Dispose` is never called.

For the singleton case, who cares if   `Dispose` is never called? At app shutdown, all the memory allocated by the process is released.

serpent5: > I've implemented this in aspnetcore/fundamentals/dependency-injection/samples/3.x/DIsample2 but `Dispose` is never called.

`Dispose` isn't called because no instances of `Service1` are created in that sample. You need to take a dependency on `Service1` for the DI container to both create and dispose of `Service1`. e.g.:

```csharp
public IndexModel(ILogger<IndexModel> logger, Service1 service1)
{
    ...
}
````

**Comment by Rick-Anderson** [comment]
@serpent5  please review when you have time, no hurry

To review the doc build 
- download https://www.dropbox.com/sh/yy994551cje0kap/AAB0tRnm7gze2kYrSYPSFHIda?dl=0
- unzip
- view HTML file in FF (Fire Fox). It best to view the doc with the widest window that has no left or right TOC (Table  of contents)

cc @rahulpnath if you'd like to review this.

File: aspnetcore/fundamentals/dependency-injection.md [hunk 10]
```diff
@@ -77,50 +77,56 @@ This interface is implemented by a concrete type, `MyDependency`:
 
 [!code-csharp[](dependency-injection/samples/3.x/DependencyInjectionSample/Services/MyDependency.cs?name=snippet1)]
 
-`MyDependency` requests an <xref:Microsoft.Extensions.Logging.ILogger`1> in its constructor. It's not unusual to use dependency injection in a chained fashion. Each requested dependency in turn requests its own dependencies. The container resolves the dependencies in the graph and returns the fully resolved service. The collective set of dependencies that must be resolved is typically referred to as a *dependency tree*, *dependency graph*, or *object graph*.
+In the sample app, the `IMyDependency` service is registered with the concrete type `MyDependency`. The registration service lifetime is <xref:Microsoft.Extensions.DependencyInjection.ServiceCollectionServiceExtensions.AddScoped*>, the lifetime of a single request. [Service lifetimes](#service-lifetimes) are described later in this topic.
```
rahulpnath, line 80: @Rick-Anderson Curious as to why this is AddScoped?
Since the dependency was explicitly created in the example above, feels it is more to be a Transient (Just from the context of this example). I might be missing something here.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -133,94 +139,89 @@ public void Configure(IApplicationBuilder app, IOptions<MyOptions> options)
...
+
+In apps that process requests, singleton services are disposed when the <xref:Microsoft.Extensions.DependencyInjection.ServiceProvider> is disposed on application shutdown. Because memory is not released until the app is shut down, memory use with a singleton must be considered.
 
 > [!WARNING]
-> It's dangerous to resolve a scoped service from a singleton. It may cause the service to have incorrect state when processing subsequent requests.
+> Do ***not*** resolve a scoped service from a singleton. It may cause the service to have incorrect state when processing subsequent requests. It's fine to resolve a singleton service from a scoped or transient service.
 
 ## Service registration methods
 
 Service registration extension methods offer overloads that are useful in specific scenarios.
```
rahulpnath, line 211: Should it be mentioned here that IServiceCollection is a collection of ServiceDescriptor and an example of adding it like that explicitly (as below)? And state the extension methods makes it easier to add registrations.
`
var item = new ServiceDescriptor(typeof(IMyDependency), context = >new MyDependency(),   ServiceLifetime.Transient);

services.Add(item);
`

aspnetcore/fundamentals/dependency-injection.md
Rick-Anderson: We could change it to transient but I don't see much advantage. The sample is used to explain how to use dependency injection. It's not a good sample because as we show later, the logger is already in the DI container. I'd like to have a simple sample to demonstrate DI  where the code required DI.

@serpent5  should we change the sample to transient?

aspnetcore/fundamentals/dependency-injection.md
Rick-Anderson: @rahulpnath  great suggestion. Check the updated doc build.

rahulpnath: lgtm :)

rahulpnath:
@Rick-Anderson Just a general observation, I am not sure what the guidelines around it are.
For the code samples, there is a mix of wrap/indent in different places
image


Not sure if this needs to be consistent. Ignore if not relevant 😀

Rick-Anderson:
@rahulpnath  it driven by what looks good when it's rendered in a browser with the widest window with no left or right TOC.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -133,94 +139,93 @@ public void Configure(IApplicationBuilder app, IOptions<MyOptions> options)
...
+Singleton lifetime services (<xref:Microsoft.Extensions.DependencyInjection.ServiceCollectionServiceExtensions.AddSingleton*>) are created either:
 
-In apps that process requests, singleton services are disposed when the <xref:Microsoft.Extensions.DependencyInjection.ServiceProvider> is disposed at app shutdown.
+* The first time they're requested.
+* When `Startup.ConfigureServices` is run and an instance is specified with the service registration.
+
+Every subsequent request uses the same instance. If the app requires singleton behavior, allowing the service container to manage the service's lifetime is recommended. Don't implement the singleton design pattern and provide code to dispose of the singleton. Services should never be disposed by code that resolved the service from a container. If a type or factory is registered as a singleton, the container will dispose the singleton automatically.
```
serpent5, line 200: Option 1:

```suggestion
Every subsequent request uses the same instance. If the app requires singleton behavior, allow the service container to manage the service's lifetime. Don't implement the singleton design pattern and provide code to dispose of the singleton. Services should never be disposed by code that resolved the service from a container. If a type or factory is registered as a singleton, the container will dispose the singleton automatically.
```

serpent5, line 200: Option 2:

```suggestion
Every subsequent request uses the same instance. If the app requires singleton behavior, allow the service container to manage the service's lifetime. Don't implement the singleton design pattern to manage the lifetime of the service. Services should never be disposed by code that resolved the service from a container. If a type or factory is registered as a singleton, the container will dispose the singleton automatically.
```

This is a separate suggestion because the _user code_ thing has been discussed previously and I don't want to bundle it all together. It also includes the preceding suggestion.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -133,94 +139,93 @@ public void Configure(IApplicationBuilder app, IOptions<MyOptions> options)
...
 > [!WARNING]
-> It's dangerous to resolve a scoped service from a singleton. It may cause the service to have incorrect state when processing subsequent requests.
+> Do ***not*** resolve a scoped service from a singleton. It may cause the service to have incorrect state when processing subsequent requests. It's fine to resolve a singleton service from a scoped or transient service.
 
 ## Service registration methods
 
+`IServiceCollection` is a collection of <xref:Microsoft.Extensions.DependencyInjection.ServiceDescriptor>. The following code shows how to add a service with a constructor:
+
+[!code-csharp[](dependency-injection/samples/3.x/DependencyInjectionSample/Startup5.cs?name=snippet)]
+
```
serpent5, line 214: I don't think the section should lead with this, because I expect it's rare that this approach is used. It's useful to know this is how it works behind the scenes, but I think I would:

- Move it to the end of the _Service registration methods_ section and highlight that the `AddXXX` methods are wrappers around this approach.
- Perhaps even just show an example of how e.g. `AddTransient` works, which demonstrates concept without showing it in app code.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -237,7 +242,7 @@ For more information, see:
...
 
-[TryAddEnumerable(ServiceDescriptor)](xref:Microsoft.Extensions.DependencyInjection.Extensions.ServiceCollectionDescriptorExtensions.TryAddEnumerable*) methods only register the service if there isn't already an implementation *of the same type*. Multiple services are resolved via `IEnumerable<{SERVICE}>`. When registering services, the developer only wants to add an instance if one of the same type hasn't already been added. Generally, this method is used by library authors to avoid registering two copies of an instance in the container.
+[TryAddEnumerable(ServiceDescriptor)](xref:Microsoft.Extensions.DependencyInjection.Extensions.ServiceCollectionDescriptorExtensions.TryAddEnumerable*) methods register the service if there isn't already an implementation *of the same type*. Multiple services are resolved via `IEnumerable<{SERVICE}>`. When registering services, the developer should add an instance if one of the same type hasn't already been added. Generally, `TryAddEnumerable` is used by library authors to avoid registering two copies of an instance in the container.
```
serpent5, line 245:
```suggestion
[TryAddEnumerable(ServiceDescriptor)](xref:Microsoft.Extensions.DependencyInjection.Extensions.ServiceCollectionDescriptorExtensions.TryAddEnumerable*) methods register the service if there isn't already an implementation *of the same type*. Multiple services are resolved via `IEnumerable<{SERVICE}>`. When registering services, the developer should add an instance if one of the same type hasn't already been added. Generally, library authors use `TryAddEnumerable` to avoid registering multiple copies of an implementation in the container.
```

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -249,16 +254,20 @@ public class MyDep : IMyDep1, IMyDep2 {}
 
 services.TryAddEnumerable(ServiceDescriptor.Singleton<IMyDep1, MyDep>());
 services.TryAddEnumerable(ServiceDescriptor.Singleton<IMyDep2, MyDep>());
-// Two registrations of MyDep for IMyDep1 is avoided by the following line:
 services.TryAddEnumerable(ServiceDescriptor.Singleton<IMyDep1, MyDep>());
 ```
 
+Service registration is generally not order dependent except when registering multiple implementations of an interface.
```
serpent5, line 260:
```suggestion
Service registration is generally order independent except when registering multiple implementations of the same type.
```

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -249,16 +254,20 @@ public class MyDep : IMyDep1, IMyDep2 {}
...
+
 ### Constructor injection behavior
 
 Services can be resolved by two mechanisms:
 
 * <xref:System.IServiceProvider>
-* <xref:Microsoft.Extensions.DependencyInjection.ActivatorUtilities>: Permits object creation without service registration in the dependency injection container. `ActivatorUtilities` is used with user-facing abstractions, such as Tag Helpers, MVC controllers, and model binders.
+* <xref:Microsoft.Extensions.DependencyInjection.ActivatorUtilities>:
+  * Permits object creation without service registration in the dependency injection container.
```
serpent5, line 268:
```suggestion
  * Creates objects without service registration in the dependency injection container.
```

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -249,16 +254,20 @@ public class MyDep : IMyDep1, IMyDep2 {}
...
+  <!-- Review: what is user-facing abstractions? Frame work features or library features? -->
```
serpent5, line 269: It looks like this means framework features.

e.g. A controller is an abstraction provided by the framework that users work with closely. Users don't usually register their controllers with DI, but those controllers can have dependencies that come from DI.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -272,14 +281,15 @@ Entity Framework contexts are usually added to the service container using the [
 
 ## Lifetime and registration options
 
-To demonstrate the difference between the lifetime and registration options, consider the following interfaces that represent tasks as an operation with a unique identifier, `OperationId`. Depending on how the lifetime of an operations service is configured for the following interfaces, the container provides either the same or a different instance of the service when requested by a class:
+To demonstrate the difference between the lifetime and registration options, consider the following interfaces that represent tasks as an operation with an identifier, `OperationId`. Depending on how the lifetime of an operations service is configured for the following interfaces, the container provides either the same or a different instance of the service when requested by a class:
```
serpent5, line 284:
```suggestion
To demonstrate the difference between the lifetime and registration options, consider the following interfaces that represent tasks as an operation with an identifier, `OperationId`. Depending on how the lifetime of an operation's service is configured for the following interfaces, the container provides either the same or a different instance of the service when requested by a class:
```

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -288,90 +298,63 @@ An `OperationService` is registered that depends on each of the other `Operation
...
-Instance: 00000000-0000-0000-0000-000000000000
-
-`OperationService` operations:
+To reduce the logging output, set "Logging:LogLevel:Microsoft:Error" in the *appsettings.Development.json* file:
 
-Transient: c4cbacb8-36a2-436d-81c8-8c1b78808aaf  
-Scoped: 31e820c5-4834-4d22-83fc-a60118acb9f4  
-Singleton: 01271bc1-9e31-48e7-8f7c-7261b040ded9  
-Instance: 00000000-0000-0000-0000-000000000000
+[!code-json[](dependency-injection/samples/3.x/DependencyInjectionSample/appsettings.Development.json)]
```
serpent5, line 327:
```suggestion
[!code-json[](dependency-injection/samples/3.x/DependencyInjectionSample/appsettings.Development.json&highlight=7)]
```

aspnetcore/fundamentals/configuration/index.md
```diff
@@ -83,6 +83,8 @@ Using the [default](#default) configuration, the *appsettings.json* and *appsett
 
 See [JSON configuration provider](#jcp) in this document for information on adding additional JSON configuration files.
 
+[!INCLUDE[](~/includes/combine-di.md)]
```
serpent5, line 86: To me, this doesn't fit here. The sample code shows configuration and options, but it's really about the extension method wrappers.

aspnetcore/includes/combine-di.md
```diff
@@ -0,0 +1,17 @@
+<a name="csc"></a>
+
+## Combining service collection
```
serpent5, line 3: It might be just me, but I read this as something that's going to cover how to take two `IServiceCollection` instances and combine them. I think it's a useful section, but I think it'd fit better inline in the DI topic, at the end of the _Register additional services with extension methods_ section. I also think that section could move down a bit, perhaps after _Service registration methods_.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -418,29 +410,31 @@ Best practices are to:
...
+If a class seems to have too many injected dependencies, that's generally a sign that the class has too many responsibilities and is violating the [Single Responsibility Principle (SRP)](/dotnet/standard/modern-web-apps-azure-architecture/architectural-principles#single-responsibility). Attempt to refactor the class by moving some of its responsibilities into a new class. Keep in mind that Razor Pages page model classes and MVC controller classes should focus on UI concerns.
 
 ### Disposal of services
 
-The container calls <xref:System.IDisposable.Dispose*> for the <xref:System.IDisposable> types it creates. If an instance is added to the container by user code, it isn't disposed automatically.
+The container calls <xref:System.IDisposable.Dispose*> for the <xref:System.IDisposable> types it creates. Services should never be disposed by any code that resolved the service from a container. If a type or factory is registered as a singleton, the container will dispose the singleton.
```
serpent5, line 417:
```suggestion
The container calls <xref:System.IDisposable.Dispose*> for the <xref:System.IDisposable> types it creates. Services should never be disposed by any code that resolved the service from a container. If a type or factory is registered as a singleton, the container disposes the singleton.
```

aspnetcore/fundamentals/configuration/index.md
Rick-Anderson: My thinking was if you're creating a bunch of config services you'd like to know how to clean up startup and move config services.

Rick-Anderson:
@halter73  3 minute question. Look at [Services not created by the service container](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-3.1&branch=pr-en-us-19271#services-not-created-by-the-service-container)

Since you can only do this incorrect approach with Singletons, who cares? Singletons are disposed when the app shuts down. But so are are any other components when the app shuts down. If that's right, should we keep this section? AFIAK, the apps behaves the same when I use the bad approach or the good approach that preceds this.

halter73:
The "Services not created by the service container" section looks good to me. What's this "incorrect approach" you're referring to?

Rick-Anderson:
@halter73  THE code with the red X through it. All the reason not to use that code seem bogus. Singletons have the same lifetime, that of the app.

image

halter73:
I see. I don't know how I overlooked the big red X. I agree that there isn't any fundamental problem with registering instances as singletons. There's an AddSingleton overload that takes an instance for a reason. It makes sense in some scenarios.

I think all the bullet points make sense, and it's worth mentioning that the instances won't be disposed automatically with the container, but I don't think we need to warn against ever registering instances.

PR #19276, opened 2020-07-21
 [Preview 8] [READY] Blazor HTML <head> tag control

### LINKED ISSUE #19265: Add documentation on influencing the HTML head from Blazor

A new preview8 feature enables influencing the HTML `head` element from Blazor.

When rendered, the `Title`, `Link`, and `Meta` components add or update data in the HTML head.

Usage example:

```
@using Microsoft.AspNetCore.Components.Web.Extensions

<Title Value="New document title" />
<Link rel="stylesheet" href="styles.css" />
<Meta name="description" content="Influencing the head from a Blazor component" />
```

Here are a few notes about the feature:
* Server-side prerendering is supported.
* Only the `Value` parameter is valid in `Title`. For `Meta` and `Link`, all provided parameters will be reflected in the DOM.
* When there are multiple `Title` components being rendered, the title of the page will reflect that of the `Title` that was most recently rendered.
* Even if there are multiple `Meta` components being rendered with identical attributes, there will be exactly one `meta` tag per `Meta` component (in other words, two `Meta` components can't refer to the same `meta` tag). This also applies to `Link` components.
* Changes to parameters of existing `Meta` components will be reflected by its `meta` tag (also applies to `Link`).
* Disposing a `Meta` component removes its `meta` tag from the DOM (also applies to `Link`).

Let me know if you would like additional information.

---
#### Document Details

* Content: [Create and use ASP.NET Core Razor components](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-3.1)
* Content Source: [aspnetcore/blazor/components/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/components/index.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #19265

[Internal Review Topic (links to section)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/additional-scenarios?view=aspnetcore-3.1&branch=pr-en-us-19276#influence-html-head-tag-elements)

@MackinnonBuck, check carefully my translation of your "characteristics" notes. It's very easy for me to make a mistake when translating engineering notes. For example ...

> Only the `Value` parameter is valid in `Title`. For `Meta` and `Link`, all provided parameters will be reflected in the DOM.

... could mean a few different things depending on interpretation, so we might need to do a little more work here if I guessed wrong.

Sorry for the *update noise* 🙉 ... kept seeing little nits to touch.

I'll put this on draft until we get to Preview 8 and merge at that time. It can be reviewed now.

### DIFF
--- a/aspnetcore/blazor/fundamentals/additional-scenarios.md
+++ b/aspnetcore/blazor/fundamentals/additional-scenarios.md
@@ -11,7 +11,7 @@ uid: blazor/fundamentals/additional-scenarios
 ---
 # ASP.NET Core Blazor hosting model configuration
 
-By [Daniel Roth](https://github.com/danroth27) and [Luke Latham](https://github.com/guardrex)
+By [Daniel Roth](https://github.com/danroth27), [Mackinnon Buck](https://github.com/MackinnonBuck), and [Luke Latham](https://github.com/guardrex)
 
 This article covers hosting model configuration.
 
@@ -236,7 +236,7 @@ When rendered, the `Title`, `Link`, and `Meta` components add or update data in 
 
 In the preceding example, placeholders for `{TITLE}`, `{FILE NAME}`, and `{DESCRIPTION}` are sting values, Razor variables, or Razor expressions.
 
-The follow characteristics apply:
+The following characteristics apply:
 
 * Server-side prerendering is supported.
 * The `Value` parameter is only valid with the `Title` component. For `Meta` and `Link` components, all provided parameters are reflected in the rendered HTML tags.
--- a/aspnetcore/blazor/fundamentals/additional-scenarios.md
+++ b/aspnetcore/blazor/fundamentals/additional-scenarios.md
@@ -242,7 +242,7 @@ The following characteristics apply:
 * The `Value` parameter is the only valid parameter for the `Title` component.
 * HTML attributes provided to the `Meta` and `Link` components are captured in [additional attributes](xref:blazor/components/index#attribute-splatting-and-arbitrary-parameters) and passed through to the rendered HTML tag.
 * For multiple `Title` components, the title of the page reflects the `Value` of the last `Title` component rendered.
-* If multiple `Meta` or `Link` components are included with identical attributes, there is exactly one HTML tag rendered per `Meta` or `Link` component. Two `Meta` or `Link` components can't refer to the same rendered HTML tag.
+* If multiple `Meta` or `Link` components are included with identical attributes, there's exactly one HTML tag rendered per `Meta` or `Link` component. Two `Meta` or `Link` components can't refer to the same rendered HTML tag.
 * Changes to the parameters of existing `Meta` or `Link` components are reflected in their rendered HTML tags.
 * Disposing a `Meta` or `Link` component removes its rendered HTML tag.
 
--- a/aspnetcore/blazor/fundamentals/additional-scenarios.md
+++ b/aspnetcore/blazor/fundamentals/additional-scenarios.md
@@ -246,10 +246,10 @@ The following characteristics apply:
 * Changes to the parameters of existing `Meta` or `Link` components are reflected in their rendered HTML tags.
 * When the `Link` or `Meta` components are no longer rendered and thus disposed by the framework, their rendered HTML tags are removed.
 
-When one of the framework components is used in a child component, the rendered HTML tag influences any other child component of the parent component as long as the child component containing the framework component is rendered. The distinction between using the one of these framework components in a child component and placing a an HTML tag in `wwwroot/index.html` or `Pages/_Host.cshtml` is that:
+When one of the framework components is used in a child component, the rendered HTML tag influences any other child component of the parent component as long as the child component containing the framework component is rendered. The distinction between using the one of these framework components in a child component and placing a an HTML tag in `wwwroot/index.html` or `Pages/_Host.cshtml` is that a framework component's rendered HTML tag:
 
-* An HTML tag created from one of the framework components can be modified by application state. A hard-coded HTML tag can't be modified by applcation state.
-* An HTML tag created from one of the framework components is removed from the HTML `<head>` when the parent component is no longer rendered.
+* Can be modified by application state. A hard-coded HTML tag can't be modified by application state.
+* Is removed from the HTML `<head>` when the parent component is no longer rendered.
 
 ::: moniker-end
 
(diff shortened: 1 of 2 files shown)


aspnetcore/blazor/fundamentals/additional-scenarios.md
```diff
@@ -218,6 +218,36 @@ Blazor.defaultReconnectionHandler._reconnectionDisplay =
...
+<Meta name="description" content="{DESCRIPTION}" />
+```
+
+In the preceding example, placeholders for `{TITLE}`, `{FILE NAME}`, and `{DESCRIPTION}` are sting values, Razor variables, or Razor expressions.
+
+The following characteristics apply:
+
+* Server-side prerendering is supported.
+* The `Value` parameter is only valid for the `Title` component.
+* All provided attributes are reflected in the rendered HTML tags.
```
MackinnonBuck, line 243: This only applies to the `Meta` and `Link` components:

```suggestion
* All parameters provided to the `Meta` and `Link` components will be reflected in the rendered HTML tags.
```

aspnetcore/blazor/fundamentals/additional-scenarios.md
```diff
@@ -218,6 +218,36 @@ Blazor.defaultReconnectionHandler._reconnectionDisplay =
...
+* For multiple `Title` components, the title of the page reflects the `Value` of the last `Title` component in the Razor markup.
```
MackinnonBuck, line 244: This is true for most cases, but if any `Title` is re-rendered, it will change the page title regardless of its location in the markup (e.g. if a `Value` parameter changes, that change will be applied to the site).

```suggestion
* For multiple `Title` components, the title of the page reflects the `Value` of the last `Title` component to be rendered.
```

Maybe there's a better way to word this than what I've suggested?

aspnetcore/blazor/fundamentals/additional-scenarios.md
```diff
@@ -218,6 +218,36 @@ Blazor.defaultReconnectionHandler._reconnectionDisplay =
...
+* If there are multiple `Meta` or `Link` components rendered with identical attributes, there is exactly one tag per component rendered. Two `Meta` or `Link` components can't refer to the same rendered HTML tag.
```
MackinnonBuck, line 245: This property also applies when `Meta` or `Link` components _aren't_ rendered with identical attributes.

```suggestion
* Even if there are multiple `Meta` or `Link` components rendered with identical attributes, there is exactly one tag per component rendered. Two `Meta` or `Link` components can't refer to the same rendered HTML tag.
```

guardrex:
Thanks ... your updates looks good. We'll get more feedback when Artak looks :eyes:.

I might .... errrr .... *probably* 🤔 .... need to take a look at how the docs generally name passed-through HTML attributes (not actually component parameters) for components. Have I called them "attributes" or "parameters"? It's a separate issue. I'll take a look at some point soon-ish.

MackinnonBuck:
@guardrex All "parameters" for `Meta` and `Link` are received as additional attributes and passed through to their respective DOM elements - if you see it fit to call them "attributes", I think that sounds great 😄

guardrex:
That's what I thought was happening.

I imagine that it could be a bit confusing to newbies coming in from HTML contexts because they see lowercase name-value pairs that match HTML attribute name-value pairs that they're familiar with and that they would call "attributes," but the topic (here and I think a few others) calls them "parameters" because this is a **component** and under-the-covers they are parameters before they finally render out as true HTML tag attributes.

mmm 🤔 .... let me think about it a bit more. There are a few spots that have the same scenario, and I want to look at how we describe component parameters in this context. Whatever we do here, I might open an issue to address it elsewhere separately.

aspnetcore/blazor/fundamentals/additional-scenarios.md
```diff
@@ -218,6 +218,36 @@ Blazor.defaultReconnectionHandler._reconnectionDisplay =
...
+::: moniker range=">= aspnetcore-5.0"
+
+## Influence HTML `<head>` tag elements
+
+*This section applies to Blazor WebAssembly and Blazor Server.*
+
+When rendered, the `Title`, `Link`, and `Meta` components add or update data in the HTML `<head>` tag elements:
+
+```razor
+@using Microsoft.AspNetCore.Components.Web.Extensions
```
MackinnonBuck, line 230: This was just updated to `Microsoft.AspNetCore.Components.Web.Extensions.Head` (https://github.com/dotnet/aspnetcore/pull/23833 is now merged, so further changes like this won't occur 😃).

```suggestion
@using Microsoft.AspNetCore.Components.Web.Extensions.Head
```

aspnetcore/blazor/fundamentals/additional-scenarios.md
```diff
@@ -218,6 +218,36 @@ Blazor.defaultReconnectionHandler._reconnectionDisplay =
...
+
+```razor
+@using Microsoft.AspNetCore.Components.Web.Extensions.Head
+
+<Title Value="{TITLE}" />
+<Link href="{URL}" rel="stylesheet" />
+<Meta content="{DESCRIPTION}" name="description" />
+```
+
+In the preceding example, placeholders for `{TITLE}`, `{URL}`, and `{DESCRIPTION}` are sting values, Razor variables, or Razor expressions.
```
mrlife, line 237: `sting` -> `string` on line 237

(not sure I can directly edit this)

guardrex: I'll fix it. You can't edit directly ... it's a bit of a pain, but you have to branch from the PR's branch and then submit a PR back to it. It's faster 🏃 if I just edit directly. Thanks for calling it out.

guardrex:
@MackinnonBuck ... I try an improvement to clarify "component" in ...

> Even if there are multiple `Meta` or `Link` components rendered with identical attributes, there is exactly one tag per component rendered. Two `Meta` or `Link` components can't refer to the same rendered HTML tag.

... to ...

> If multiple `Meta` or `Link` components are included with identical attributes, there's exactly one HTML tag rendered per `Meta` or `Link` component. Two `Meta` or `Link` components can't refer to the same rendered HTML tag.

Does that make sense?

Also, for ...

> Disposing a `Meta` or `Link` component removes its rendered HTML tag.

... do you mean with an `@ref` so that the scenario is to implement `IDisposable` and then call `Dispose` on the component's reference in the `Dispose` method? If so, I'll add an example along those lines.

MackinnonBuck:
@guardrex I think your clarification looks great.

Regarding "dispose", I was more saying when the component is naturally disposed (i.e. no longer being rendered), its HTML tag disappears with it. No extra code is required to make this work.

guardrex:
Ah, I see what you mean.

I'll make a little touch to it along the lines of what you said. I won't mention the explicit `Dispose`. That seems like an *ultra edge case* even if it is correct.

PR #19333, opened 2020-07-27
[Preview 8] [READY] Blazor RCL stylesheet and Link component

### LINKED ISSUE #19317: Mention linking the stylesheet
This section is missing a mention that the stylesheet for the example component, the one provided by the RCL template, must be linked via the `wwwroot/index.html` file. Also, I think the section should use the exact example component that the RCL project template provides.

Include the new `Link` component for 5.0+. Use this issue to update any spot where adding a link to the `<head>` is described.

---
#### Document Details

* ID: d2b709f5-e04e-fee3-7eea-1364893ab997
* Version Independent ID: a92c88d3-6ee8-1e69-f4c3-49eb82909746
* Content: [ASP.NET Core Razor components class libraries](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/class-libraries?view=aspnetcore-3.1&tabs=visual-studio#consume-a-library-component)
* Content Source: [aspnetcore/blazor/components/class-libraries.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/components/class-libraries.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #19317

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/components/class-libraries?view=aspnetcore-3.1&branch=pr-en-us-19333)

* Adds the `Link` component to the Blazor RCL content.
* Explains how to get the stylesheet applied to components: 5.0+ with `Link` component or `<link>` tag and <5.0 with a `<link>` tag.
* Better naming: I've been leaning away from `My`-named things these days, so I name the lib `ComponentLibrary`. Also, I go with the project template-generated example component in the RCL named `Component1.razor`.
* I went ahead and cross-linked the `Link` component. There's no build warning on the premature link here, but the coverage won't be there until [[Preview 8] Blazor HTML <head> tag control (dotnet/AspNetCore.Docs &num;19276)](https://github.com/dotnet/AspNetCore.Docs/pull/19276) goes in.

### DIFF
--- a/aspnetcore/blazor/components/class-libraries.md
+++ b/aspnetcore/blazor/components/class-libraries.md
@@ -5,7 +5,7 @@ description: Discover how components can be included in Blazor apps from an exte
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 03/23/2020
+ms.date: 07/27/2020
 no-loc: [Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
 uid: blazor/components/class-libraries
 ---
@@ -28,7 +28,7 @@ Just as components are regular .NET types, components provided by an RCL are nor
 1. Create a new project.
 1. Select **Razor Class Library**. Select **Next**.
 1. In the **Create a new Razor class library** dialog, select **Create**.
-1. Provide a project name in the **Project name** field or accept the default project name. The examples in this topic use the project name `MyComponentLib1`. Select **Create**.
+1. Provide a project name in the **Project name** field or accept the default project name. The examples in this topic use the project name `ComponentLibrary`. Select **Create**.
 1. Add the RCL to a solution:
    1. Right-click the solution. Select **Add** > **Existing Project**.
    1. Navigate to the RCL's project file.
@@ -48,10 +48,10 @@ Just as components are regular .NET types, components provided by an RCL are nor
 
 # [.NET Core CLI](#tab/netcore-cli)
 
-1. Use the **Razor Class Library** template (`razorclasslib`) with the [`dotnet new`](/dotnet/core/tools/dotnet-new) command in a command shell. In the following example, an RCL is created named `MyComponentLib1`. The folder that holds `MyComponentLib1` is created automatically when the command is executed:
+1. Use the **Razor Class Library** template (`razorclasslib`) with the [`dotnet new`](/dotnet/core/tools/dotnet-new) command in a command shell. In the following example, an RCL is created named `ComponentLibrary`. The folder that holds `ComponentLibrary` is created automatically when the command is executed:
 
    ```dotnetcli
-   dotnet new razorclasslib -o MyComponentLib1
+   dotnet new razorclasslib -o ComponentLibrary
    ```
 
    > [!NOTE]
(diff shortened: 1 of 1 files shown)


guardrex:
@MackinnonBuck ... Is this still on your radar for review? ... no rush ... just checking. There's a little bit of `Link` component goodness on it, so I thought you might take a look at the whole thing while checking that bit out.

guardrex:
@MackinnonBuck ... try it now. I like your language there.

I think I need to look at the other `Link` content now. Don't we need these remarks there, too? Seems so.

MackinnonBuck:
I like the update, @guardrex!

I think these two bullets in the existing `<head>` component content just about cover what I've explained:
> * Changes to the parameters of existing `Meta` or `Link` components are reflected in their rendered HTML tags.
> * When the `Link` or `Meta` components are no longer rendered and thus disposed by the framework, their rendered HTML tags are removed.

I suppose the second bullet more accurately describes the actual behavior than my comment in this PR saying that the HTML tag "will be removed from the HTML head when its parent component stops being rendered", since it could also be removed if you were conditionally rendering a `<Link>`, and the `<Link>` stopped being rendered.

However, if you think it's worth rewording the `<head>` component content, I don't have any problems with that 😃 

One question I have about the changes in this PR - Is there really a reason to use `<Link>` in this example instead of just using a static `<link>` in the `<head>`? In this specific case, using `<Link>` doesn't seem to bring too much additional benefit, and using `<link>` is technically slightly more performant since there would be no additional code running...

Was the goal behind this change to bring more attention to users the existence of `<Link>` rather than show an example of its typical usage? IMO, linking stylesheets using `<Link>` wouldn't be a very common use case (maybe a more typical example would be changing the webpage icon depending on whether there are pending notifications, etc.).

Let me know what you think!

guardrex:
> I think these two bullets in the existing <head> component content just about cover what I've explained:

It's not as plain as explaining it out. The added text makes those points clearer, especially nice here so that the reader doesn't have to click over there to read the full explanation. I think devs need to be aware that if they RCL an asset in like this that other children of parents where they use the RCL's component ... children of the app that is consuming the RCL ... are also going to get the asset ... as you say ... if the RCL's component is rendered. I think that's a good point here.

> worth rewording the <head> component content

I added content to that other PR. It leaves the bullets as they are but also adds this additional explanation under the bullet list. That text probably needs more work. I'll take a look and edit it tomorrow (Thursday) morning.

> Is there really a reason to use <Link> in this example instead of just using a static <link> in the <head>? 

The stylesheet is in the RCL isn't automatically applied. Since the RCL has this component and its own stylesheet, it seemed like a nice way to compose an RCL that wouldn't require the dev to do anything other than add the component in the project consuming the RCL. Then, they're done ... no messing about with the `<head>` in the app. However, I do show linking the stylesheet both ways in case their intention is to provide the RCL's stylesheet everywhere.

guardrex:
btw - This is how I figured this out. I was working with RCLs in multiple client app scenarios. See the images in ...

https://github.com/dotnet/AspNetCore.Docs/pull/19315

... that `JeepImage.Component1` example was what gave me the idea. It would be nice to have the `Link` component in the RCL's `Jeep` component and be done with it.

MackinnonBuck:
@guardrex Ah, thanks for the clarifications! In that case, I think the changes look good to me 😃

guardrex:
@MackinnonBuck and @captainsafia ... Artak has pulled the trigger on the batch of Pre8 doc PRs. They're going in right now ... in an hour after lunch 🍔🥛 ... and in advance of the actual release of Pre8. Let me know if you have a concern about it. The content will only be visible when the reader selects to view the preview docs.

PR #19375, opened 2020-07-30
Create markdownlint.yml

Youssef1313:


Youssef1313:
This is **NOT** ready to merge yet. There are [many lint violations](https://github.com/Youssef1313/AspNetCore.Docs/actions/runs/188807910).

@tdykstra, Is there any specific rules you want me to disable?

tdykstra:
>any specific rules you want me to disable

All rules should be disabled by default and we should only enable rules that the team has discussed and agreed on.

Youssef1313:
This GitHub action should be dependent on [.markdownlint.json](https://github.com/dotnet/AspNetCore.Docs/blob/master/.markdownlint.json) which already exists in the repo, but there are [violations](https://github.com/Youssef1313/AspNetCore.Docs/runs/928419780?check_suite_focus=true).

The most obvious rules that may be disabled are "MD040/fenced-code-language Fenced code blocks should have a language specified", and "MD004/ul-style Unordered list style [Expected: asterisk; Actual: dash]".
But I think there will remain many violations. "MD044/proper-names Proper names should have the correct capitalization" is among of them.

@tdykstra, What do you think?

tdykstra:
MD040 - probably OK because we will before long have a build validation that will do the same thing.
MD004 - I thought we had decided earlier not to enable this one.
MD044 - this seems potentially problematic, as names are used in different contexts that may call for different capitalization

Youssef1313:
> MD004 - I thought we had decided earlier not to enable this one.

It looks like it's currently enabled. I'll disable it.

Is adding `"ul-style": false` the correct way to disable it? I'm unsure.

tdykstra:
It's disabled here:  https://github.com/dotnet/docs/blob/master/.markdownlint.json

Sorry, I didn't notice you were talking about the ASP.NET Core repo.  For that repo I'll defer to @Rick-Anderson

Rick-Anderson:
We're not ready to add these rules. cc @scottaddie @guardrex

PR #19376, opened 2020-07-30
Blazor WASM security outgoing request coverage

### LINKED ISSUE #19352: Get "401 Unauthorized" when running the Server project and the Client project on different ports
Hallo!

Thank you for a good sample:
https://docs.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/hosted-with-azure-active-directory-b2c?view=aspnetcore-3.1

I get "401 Unauthorized" when running the Server project and the Client project on different ports to simulate having the Client Blazor App and the Server API on different domains.

I assume that the AAD B2C Authentication tokens are no longer sent from the Client project to the Server project since I cannot find any "Authentication Header" in the Header of the Request sent to the server?

What do I need to change to make this work?

Am I missing a setting or similar to allow the AAD B2C Authentication tokens to be sent to another port/domain?

I have written an in detail question on Stack Overflow with what I have changed:
https://stackoverflow.com/questions/63124615/blazor-webassembly-azure-ad-b2c-connect-to-webapi-on-other-port-domain-does-not

---
#### Document Details

* ID: 520b9821-ecef-20af-a601-ea36c4297d88
* Version Independent ID: da9c8ac9-4d7a-4b40-56a2-a4bfe7083567
* Content: [Secure an ASP.NET Core Blazor WebAssembly hosted app with Azure Active Directory B2C](https://docs.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/hosted-with-azure-active-directory-b2c?view=aspnetcore-3.1)
* Content Source: [aspnetcore/blazor/security/webassembly/hosted-with-azure-active-directory-b2c.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/security/webassembly/hosted-with-azure-active-directory-b2c.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #19352

@TNCodes ... are you able to review this? See if this would have made it clear that you would need to:

* Configure a custom `AuthorizationMessageHandler`.
* Configure CORS.

... for the scenario that we were discussing on the issue.

Summary of the updates on this PR ...

* Move the `BaseAddressAuthorizationMessageHander` coverage **UP**. It's more of a foundational concept to cover before speaking to dev configuration using a custom `AuthorizationMessageHandler` class or custom `AuthorizationMessageHandler` configuration.
* Make it utterly clear that the recommendation is to use a custom `AuthorizationMessageHandler`.
* Clarify the conditions on when the access token is added relative to the base URI of the client and the server apps.
* Call out the need for server-side CORS config and clarify that a hosted app based on the project template doesn't require CORS config. Also state that non-shared base path clients must have CORS config.

Notes ...

* There are existing cross-links in place from the WASM security topics, so no changes required there.
* Let's avoid pitching the idea that the client app of a hosted Blazor solution should be split out, as that would defeat the purpose of such an app. Devs should go with the standalone guidance in that client scenario. That guidance is cross-linked to this doc, too, so devs should find it ... *in theory* 🤞.
* These concepts are so thorny to explain that I plan to sleep on it 💤🛌 and edit once more on Friday (7/31) before merging.


TNCodes:
Hallo!

Thanks for the updates!

I am busy with other things right now, but I will review this later today.

Mvh,
Thomas Nyström

TNCodes:
Hallo!

I have now reviewed the updates you suggest.

I like the idea of using a Custom AuthorizationMessageHandler and agree that it should be the preferred option. Thankyou for pointing it out!

I suggest simply adding a bullet for "Using BaseAddressAuthorizationMessageHandler" in the "Attach tokens to outgoing requests" section and since this will probably be the least used option, then it should be the last option.

I also think that we should have a separate section that explains in detail of how "BaseAddressAuthorizationMessageHandler " works and when to use it instead of mixing it in the "Configure AuthorizationMessageHandler" section.

Mixing the text about "BaseAddressAuthorizationMessageHandler" in the text about "Configure AuthorizationMessageHandler" section is rather confusing since they are two different options.

guardrex:
Thanks @TNCodes ... Thanks for the feedback. It looks like I'll be working on this further over the weekend or Monday morning. I'll ping u back.

guardrex:
@TNCodes ... I made additional significant updates to this today.

I don't agree (*YET*) with splitting out `BaseAddressAuthorizationMessageHandler` into its own section (and having such a section last) because today I don't know how many devs are using an OOB hosted Blazor solution versus how many devs are using separate server APIs. No doubt, the external API pattern is going to be very common, but there probably will be many specifications out there where an OOB hosted Blazor solution will work well.

Since `BaseAddressAuthorizationMessageHandler` is a form of `AuthorizationMessageHandler`, it is rolled in at the top right after `AuthorizationMessageHandler` is introduced. I agree on not leaving it down in the customize `AuthorizationMessageHandler` section.

> how "BaseAddressAuthorizationMessageHandler " works and when to use it

That's on the PR, and the text is clear to use it for scenarios where the requests are within the app's base URI. I made that whole sentence **bold**.

I think devs will understand these concepts without additional text, but we'll continue to take feedback and see if more work is needed. This is *tricky* to explain because the word "base" is used with different meaning (i.e., different values) depending on the context (hosted Blazor OOB vs. non-hosted Blazor).

I'm going to need to sleep on this 💤🛌 one more night. When I make this many changes, it's best to take a little break from it and come back the next morning and edit again ☕👀 .

PR #19401, opened 2020-08-03
Add anti-patterns example in DI recommendations

xboxeer:
According to https://github.com/dotnet/AspNetCore.Docs/pull/19156, we'd like to add a new anti-patterns to DI recommendations in DI docs. The reason is in Stackoverflow, some community member is suggesting to call BuildServiceProvider to resolve service instance in ConfigureService method(https://stackoverflow.com/questions/31863981/how-to-resolve-instance-inside-configureservices-in-asp-net-core), which will result to couple issues as described in this PR. So it is better to add this as an anti-patterns in recommendations and provide an alternative way to address developers requirement.

### DIFF
--- a/aspnetcore/fundamentals/dependency-injection.md
+++ b/aspnetcore/fundamentals/dependency-injection.md
@@ -574,7 +574,7 @@ The factory method of single service, such as the second argument to [AddSinglet
 
 * Avoid call services.BuildServiceProvider() method in ConfigureServices. This usually happens when developer want to resolve service in ConfigureServices phase. Some code that a developer will use like below.
 
-**Incorrect:**
+**❌Incorrect:**
 
 ```csharp
 public void ConfigureService(IServiceCollection services)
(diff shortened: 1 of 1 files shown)


scottaddie:
@davidfowl Please review. I'll do an edit pass if you're okay with this content.

davidfowl:
This is good.

scottaddie:
@davidfowl That's a good idea. We have an example at https://docs.microsoft.com/dotnet/standard/library-guidance/#types-of-recommendations. I'll incorporate that in my edit pass.

Rick-Anderson:
@scottaddie I'm dong a major update of this doc in #19271 - so I'll take care of this.

scottaddie:
@Rick-Anderson Merge this PR when you're ready.

Rick-Anderson:
I think I'll do an image with a red X in the bad code. Can't copy/paste and it's obvious you shouldn't use the code.

Rick-Anderson:
@scottaddie  see [this link](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-3.1&branch=pr-en-us-19271#recommendations)

Can't copy image and has red X through the code.

xboxeer:
Seems we can directly use the ❌ in plan text, can we add it before Incorrect, and ✔️ before correct?

xboxeer:
@Rick-Anderson I added ✔️and ❌in this PR, would you please help check? My understanding is that they can be displayed in preview so should be displayed in our docs site as well. Correct me if I there is any issue

Rick-Anderson:
Moved to #19468

PR #19455, opened 2020-08-08
Update adding-controller.md

heesunlee9:
MVC Controller Class -> Controller Class.

I made an app following this page today. 
I guess MVC Controller Class is from an older version of Visual Studio for mac.

Have a nice day!


Rick-Anderson:
Thanks!

PR #19468, opened 2020-08-11
Anti pattern di/ra/xboxeer

Rick-Anderson:
@xboxeer move your commits here. I'm merge this when the DI rewrite merges. Thanks for your help.

[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-3.1&branch=pr-en-us-19468#ASP0000)


Rick-Anderson:
@xboxeer  here is what the build looks like. 

image

The bad code is an image, so you can't copy/paste it.

xboxeer:
@Rick-Anderson Thank you for moving this commits here. I think we can add a description to the sample code so that developer can understanding what we are doing in the sample. My suggestion would be like this:

For example, when you want to dynamically config your LoginPath for CookieAuthentication, you may want to get the LoginPath from database through a services you just registered in ConfigureServices
->Incorrect code here

Another thoughts is that, when Visual Studio is telling the code is not recommended(The ASP0000 error), can we add a link in the error message to this document section? From a developer's point of view, vs is telling me that this is not recommended, but not telling me what is the correct way to do so is very confusing. Although people may search internet about this ASP0000 error but directly adding a link to this scenario can save them a lot of time.

PR #19472, opened 2020-08-11
Fix typo

ChristopheVanassche:
"currentFilter" should be using capital "C"


scottaddie:
Thank you for fixing this, @ChristopheVanassche!

PR #19483, opened 2020-08-12
File content type provider in Blazor Server

### LINKED ISSUE #19340: Missing critical information regarding UseStaticFiles with server-side
[EDIT by guardrex to replace the scraped content link and remove the metadata]

In order to use a font, I had to extend static files to serve woff/ttf.

```cs
            var provider = new FileExtensionContentTypeProvider();
            provider.Mappings[".ttf"] = "font/ttf";
            provider.Mappings[".woff"] = "font/woff";
            provider.Mappings[".woff2"] = "font/woff2";
            app.UseStaticFiles(new StaticFileOptions
            {
                ContentTypeProvider = provider
            });
```

Turns out that this breaks server-side blazor. After flailing around without a clue as to why my site was dead, other than a cryptic console error about not being able to load 'a resource' (without actually logging what resource that is), I found [this post](https://github.com/dotnet/aspnetcore/issues/9588) that explains why.

There is a critical dependence on using the default UseStaticFiles() in Startup.cs, otherwise you break the serving of _framework.blazor.cs, and it is **absolutetely imperative** that you add an additional call to
```cs
app.UseStaticFiles();
```

The documentation (I started with this page) is missing this vital fact.  The doc for UseStaticFiles(StaticFileOptions) makes no mention of this, either.

guardrex:
Fixes #19340

Per engineering remarks :point_right: https://github.com/dotnet/aspnetcore/issues/9588#issuecomment-505465169

Thanks @Paul-Dempsey! 🎸 ... Do you want to review this to confirm that it would have surfaced the guidance for you?

### DIFF
--- a/aspnetcore/blazor/fundamentals/additional-scenarios.md
+++ b/aspnetcore/blazor/fundamentals/additional-scenarios.md
@@ -5,7 +5,7 @@ description: Learn about additional scenarios for ASP.NET Core Blazor hosting mo
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 07/10/2020
+ms.date: 08/12/2020
 no-loc: [cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
 uid: blazor/fundamentals/additional-scenarios
 ---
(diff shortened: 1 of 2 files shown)


Paul-Dempsey:
This content addresses my issue. Hard to see exactly how it appears in context, so I'm wondering if I would have found it, but this a definite improvement.

guardrex:
Thanks ... Right now, it will be a bit buried in the Blazor node ... *Fundamentals* > *Additional Resources* topic. It will be there for a bit until perhaps we have a Blazor static files dedicated topic one day. It's also cross-linked from the main Static files topic.

I'll keep an :ear: open for more feedback. If we get asked again .... and again ... about it, then it's not surfaced well enough and I'll see what else I can do. For example, more feedback might be the thing that prompts us to actually have a dedicated topic in the Blazor node.

PR #19550, opened 2020-08-18
clarify provider and category event level

Rick-Anderson:


noahfalk:
cc @dotnet/dotnet-diag

PR #19575, opened 2020-08-19
trust linux cert anchor

Rick-Anderson:
@danroth27  can I change the target URL: 
https://docs.microsoft.com/aspnet/core/security/enforcing-ssl                  <-- Current
https://docs.microsoft.com/aspnet/core/security/enforcing-ssl#ssl-linux   <-- Proposed


javiercn:
There are a few things that need to line up, and that might mean this is after 5.0 RTM. The current text is fine, we can also say something like export the certificate to a file on disk with `dotnet dev-certs https -ep ./https.pem --format pem` and follow your distribution guidelines to trust the certificate.

For .NET Core to trust the certificate on Linux we are waiting for a fix that is in openssl 1.1.1h, so that will be a requisite.

PR #19578, opened 2020-08-19
Add ASP.NET Identity to no loc

Rick-Anderson:


guardrex:
There's an extra space.

IIRC, I thought @JasonCard said that since we were already quoting multi-word strings to go ahead and keep doing that.

guardrex:
Quote goes inside the comma ..... and now you have two extra spaces.

guardrex:
Down to one extra space.

This just isn't your day for metadata updates. :smile:

PR #19624, opened 2020-08-23
gRPC performance best practices

JamesNK:
Internal preview: https://review.docs.microsoft.com/en-us/aspnet/core/grpc/performance?view=aspnetcore-3.0&branch=pr-en-us-19624

Collection of best practices for good performance with gRPC.

### DIFF
--- a/aspnetcore/grpc/performance.md
+++ b/aspnetcore/grpc/performance.md
@@ -1,0 +1,112 @@
+---
+title: Performance best practices
+author: jamesnk
+description: Learn the best practices for building high-performance gRPC services.
+monikerRange: '>= aspnetcore-3.0'
+ms.author: jamesnk
+ms.date: 08/23/2020
+no-loc: ["ASP.NET Core Identity", cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
+uid: grpc/performance
+---
+# Performance best practices
+
+By [James Newton-King](https://twitter.com/jamesnk)
+
+gRPC is designed for high-performance services. This document explains how to get the best performance possible from gRPC.
+
+## Reuse channel
+
+A gRPC channel should be reused when making gRPC calls. Reusing a channel allows calls to be multiplexed through an existing HTTP/2 connection.
+
+If a new channel is created for each gRPC call then the amount of time it takes to complete can increase significantly. Each call will require multiple network round-trips between the client and the server to create an HTTP/2 connection:
+
+1. Opening a socket
+2. Establishing TCP connection
+3. Negotiating TLS
+4. Starting HTTP/2 connection
+5. Making the gRPC call
+
+Channels are safe to share and reuse between gRPC calls:
+
+* gRPC clients are created with channels. gRPC clients are lightweight objects and don't need to be cached or reused.
+* Multiple gRPC clients can be created from a channel, including different types of clients.
+* A channel and clients created from the channel can safely be used by multiple threads.
+* Clients created from the channel can make multiple simultaneous calls.
+
+::: moniker range=">= aspnetcore-5.0"
+
+## Keep alive pings
+
+Keep alive pings can be used to keep HTTP/2 connections alive during periods of inactivity. Having an existing HTTP/2 connection ready when an app resumes activity allows for the initial gRPC calls to be made quickly, without a delay caused by the connection being reestablished.
+
+Keep alive pings are configured on <xref:System.Net.Http.SocketsHttpHandler>:
+
+```csharp
+var handler = new SocketsHttpHandler
+{
+    PooledConnectionIdleTimeout = Timeout.InfiniteTimeSpan,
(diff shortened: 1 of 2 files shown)


aspnetcore/grpc/performance.md
```diff
@@ -0,0 +1,100 @@
...
+
+By [James Newton-King](https://twitter.com/jamesnk)
+
+gRPC is designed for high-performance services. This document explains how to get the best performance possible from gRPC.
+
+## Reuse channel
+
+A gRPC channel should be reused when making gRPC calls. Reusing a channel allows calls to be multiplexed through an existing HTTP/2 connection.
+
+If a new channel is created for each gRPC call then the amount of time it takes to complete can increase significantly. Each call will require multiple network round-trips between the client and the server to create a HTTP/2 connection:
```
Rick-Anderson, line 21:
```suggestion
If a new channel is created for each gRPC call then the amount of time it takes to complete can increase significantly. Each call will require multiple network round-trips between the client and the server to create an HTTP/2 connection:
```

aspnetcore/grpc/performance.md
```diff
@@ -0,0 +1,112 @@
...
+1. Opening a socket
+2. Establishing TCP connection
+3. Negotiating TLS
+4. Starting HTTP/2 connection
+5. Making the gRPC call
+
+Channels are safe to share and reuse between gRPC calls:
+
+* gRPC clients are created with channels. gRPC clients are lightweight objects and don't need to be cached or reused.
+* Multiple gRPC clients can be created from a channel, including different types of clients.
```
JunTaoLuo, line 32: Should we mention the max streams per connection limit? We've had several issues filed when that limit is reached so we might want to at least mention it here.

JamesNK:
@Rick-Anderson I've made a lot of changes since you looked at the PR. Could you please give the doc's grammar/wording another review

aspnetcore/grpc/performance.md
```diff
@@ -0,0 +1,150 @@
...
+    EnableMultipleHttp2Connections = true
+};
+
+var channel = GrpcChannel.ForAddress("https://localhost:5001", new GrpcChannelOptions
+{
+    HttpHandler = handler
+});
+```
+
+The preceding code configures a channel that sends a keep alive ping to the server every 60 seconds during periods of inactivity. The ping will ensure the server and any proxies in use won't close the connection because of inactivity.
```
Rick-Anderson, line 96:
```suggestion
The preceding code configures a channel that sends a keep alive ping to the server every 60 seconds during periods of inactivity. The ping ensures the server and any proxies in use won't close the connection because of inactivity.
```

aspnetcore/grpc/performance.md
```diff
@@ -0,0 +1,150 @@
...
+
+HTTP/2 connections typically have a limit on the number of [maximum concurrent streams (active HTTP requests)](https://http2.github.io/http2-spec/#rfc.section.5.1.2) on a connection at one time. By default, most servers set this limit to 100 concurrent streams.
+
+A gRPC channel uses a single HTTP/2 connection, and concurrent calls are multiplexed on that connection. When the number of active calls reaches the connection stream limit, additional calls will be queued in the client. Queued calls will wait for active calls to complete before they are sent. Applications with very high load, or long running streaming gRPC calls, could see performance issues caused by calls queuing because of this limit.
+
+::: moniker range=">= aspnetcore-5.0"
+
+.NET 5 introduces the `SocketsHttpHandler.EnableMultipleHttp2Connections` property. When set to true, additional HTTP/2 connections are created by a channel when the concurrent stream limit is reached. When a `GrpcChannel` is created its internal `SocketsHttpHandler` is automatically configured to create additional HTTP/2 connections. If an app configures its own handler, consider setting `EnableMultipleHttp2Connections` to true:
```
Rick-Anderson, line 44:
```suggestion
.NET 5 introduces the `SocketsHttpHandler.EnableMultipleHttp2Connections` property. When set to `true`, additional HTTP/2 connections are created by a channel when the concurrent stream limit is reached. When a `GrpcChannel` is created its internal `SocketsHttpHandler` is automatically configured to create additional HTTP/2 connections. If an app configures its own handler, consider setting `EnableMultipleHttp2Connections` to `true`:
```

aspnetcore/grpc/performance.md
```diff
@@ -0,0 +1,150 @@
...
+    }
+});
+```
+
+::: moniker-end
+
+There are a couple of workarounds for .NET Core 3.1 apps:
+
+* Create separate gRPC channels for areas of the app with high load. For example, the `Logger` gRPC service might have very high load. Use a separate channel to create the `LoggerClient` in the app.
+* Use a pool of gRPC channels. A simple way to do this could be to create a list of gRPC channels. `Random` is used to pick a channel from the list each time a gRPC channel is needed. This would randomly distribute calls over multiple connections.
```
Rick-Anderson, line 63:
```suggestion
* Use a pool of gRPC channels, for example,  create a list of gRPC channels. `Random` is used to pick a channel from the list each time a gRPC channel is needed. Using `Random` randomly distributes calls over multiple connections.
```

PR #19626, opened 2020-08-24
gRPC and inter-process communication

JamesNK:
Internal preview: https://review.docs.microsoft.com/en-us/aspnet/core/grpc/interprocess?view=aspnetcore-5.0&branch=pr-en-us-19626

Note: Uses features new in .NET 5 RC1. Don't merge until RC1 is released.

### DIFF
--- a/aspnetcore/grpc/interprocess.md
+++ b/aspnetcore/grpc/interprocess.md
@@ -1,0 +1,86 @@
+---
+title: Use gRPC with inter-process communication
+author: jamesnk
+description: Learn how to make inter-process gRPC calls over custom transports.
+monikerRange: '>= aspnetcore-5.0'
+ms.author: jamesnk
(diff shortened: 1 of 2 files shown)


Rick-Anderson:
@JamesNK  if possible can you target this to the RC1 branch?

BTW, RC1 is available at https://github.com/dotnet/installer/blob/master/README.md#installers-and-binaries, so maybe we could put in a note that this requires RC1, which is planned to be released mid September.

Rick-Anderson:
@JamesNK  I think it's OK to merge this once you're ready because we're already adding RC1 content. With your monike, the article won't be visible by default. RC1 is available now, just not the released RC1.

JamesNK:
I've added a note that `ConnectionFactory` is new in RC1. It can be removed once RC1 is generally available.

PR #19637, opened 2020-08-25
replace case-insensitive with not case sensitive

Rick-Anderson:
@JasonCard  given how common this is across docs.ms.com, it might be best to fix the translator?


JasonCard:
I see, yes, I'll follow up with the loc vendor to fix the German translation.

PR #19623, opened 2020-08-23
Add Protobuf guide for .NET

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/1020

Preview - https://review.docs.microsoft.com/en-us/aspnet/core/grpc/protobuf?view=aspnetcore-3.0&branch=pr-en-us-19623

gRPC users are often asking questions about defining Protobuf types. This page summarizes the content from https://docs.microsoft.com/en-us/dotnet/architecture/grpc-for-wcf-developers/protocol-buffers

PR #19638, opened 2020-08-25
Add Create gRPC services doc

JamesNK:
Add a new doc: Create gRPC services

In the TOC I have grouped this with "gRPC services using C#". When the [Protobuf guide](https://github.com/dotnet/AspNetCore.Docs/pull/19623) is merged it will also be in this TOC group

### DIFF
--- a/aspnetcore/grpc/services.md
+++ b/aspnetcore/grpc/services.md
@@ -206,7 +206,7 @@ public override async Task<SimpleResponse> StreamingFromClient(
 A bi-directional streaming method starts *without* the method receiving a message. The `requestStream` parameter is used to read messages from the client. The method can choose to send messages with `responseStream.WriteAsync`. A bi-directional streaming call is complete when the the method returns.
 
 ```csharp
-public override async Task ReadySetGo(IAsyncStreamReader<SimpleRequest> requestStream,
+public override async Task StreamingBothWays(IAsyncStreamReader<SimpleRequest> requestStream,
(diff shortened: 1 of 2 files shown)


aspnetcore/grpc/services.md
```diff
@@ -0,0 +1,236 @@
...
+    return new SimpleResponse();
+}
+```
+
+### Bi-directional streaming method
+
+A bi-directional streaming method starts *without* the method receiving a message. The `requestStream` parameter is used to read messages from the client. The method can choose to send messages with `responseStream.WriteAsync`. A bi-directional streaming call is complete when the the method returns.
+
+```csharp
+public override async Task ReadySetGo(IAsyncStreamReader<SimpleRequest> requestStream,
```
JunTaoLuo, line 209: This example is a little simplistic right? We should probably mention that bi-directional stream often involves dispatching the reading of the request stream like the real [RacerService](https://github.com/grpc/grpc-dotnet/blob/master/examples/Racer/Server/Services/RacerService.cs#L36). It might be a little advanced for this doc but we should at least mention it. The current sample only illustrate a ping-pong pattern which doesn't fully illustrate the capabilities of bi-directional streams.

JamesNK:
Ready to merge after a grammar/wording review by docs team.

Rick-Anderson:
Edits moved to #19669

JamesNK:
Thanks!

PR #19641, opened 2020-08-25
Blazor forms display name support

### LINKED ISSUE #19633: [RC1] Add documentation on DisplayName for validation messages
We now support setting a `DisplayName` attribute on InputNumber, InputDate, and InputSelect components. So for example, I can have a form that looks like this.

```
@page "/form"

<EditForm Model="@this" OnValidSubmit="@HandleValidSubmit">
    <div class="input-group mb-3">
        <div class="input-group-prepend">
            <span class="input-group-text" id="birth-date">Date of Birth</span>
         </div>
         <InputDate Class="form-control" Id="DateOfBirth" @bind-Value="@BirthDate" DisplayName="birthday"/>
          <div class="invalid-feedback d-block">
              <ValidationMessage For="@(() => BirthDate)" />
          </div>
    </div>
</EditForm>

@code {
    public DateTime BirthDate { get; set; }

    public void HandleValidSubmit()
    {
        Console.WriteLine("foo");
    }
}
```

If the date field is invalid, then the validation message will read:

> The birthday must be a date.

Wheras previously, it would show this.

> The BirthDate must be a date.

We probably want to add this under the "Validation Support" heading.

---
#### Document Details

* ID: eef47457-c8cf-74a4-b884-c1fe5c2ef52d
* Version Independent ID: 09a02d10-2834-467d-d8c2-3fd19ceb3863
* Content: [ASP.NET Core Blazor forms and validation](https://docs.microsoft.com/en-us/aspnet/core/blazor/forms-validation?view=aspnetcore-3.1)
* Content Source: [aspnetcore/blazor/forms-validation.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/forms-validation.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #19633

It might be handy to foreshadow the proper section by adding a column to the built-in table. I think 🤔 that will be fine; but if it seems distracting, I'll kill that and we can just keep the section. :ear:

### DIFF
--- a/aspnetcore/blazor/forms-validation.md
+++ b/aspnetcore/blazor/forms-validation.md
@@ -62,7 +62,7 @@ In the preceding example:
 
 ## Built-in forms components
 
-A set of built-in input components are available to receive and validate user input. Inputs are validated when they're changed and when a form is submitted. Available input components are shown in the following table. For built-in components that support the `DisplayName` parameter, the display name assigned to the component appears in validation messages.
+A set of built-in components are available to receive and validate user input. Inputs are validated when they're changed and when a form is submitted. Available input components are shown in the following table. For built-in components that support the `DisplayName` parameter, the display name assigned to the component appears in validation messages.
 
 ::: moniker range=">= aspnetcore-5.0"
 
--- a/aspnetcore/blazor/forms-validation.md
+++ b/aspnetcore/blazor/forms-validation.md
@@ -270,7 +270,7 @@ The following built-in components support display names with the `DisplayName` p
 In the following `InputDate` component example:
 
 * The display name (`DisplayName`) is set to "birthday" (`birthday`).
-* The component is bound to the `DateTime` type.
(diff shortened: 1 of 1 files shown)


guardrex:
@captainsafia Earlier, we went ahead before Pre8 landed and merged Pre8 content. Do you/Artak want to continue doing that for RC1? It seems kind'a scary putting up RC1 bits when a dev would have to be building the framework as you go to get these toys.

captainsafia:
> @captainsafia Earlier, we went ahead before Pre8 landed and merged Pre8 content. Do you/Artak want to continue doing that for RC1? It seems kind'a scary putting up RC1 bits when a dev would have to be building the framework as you go to get these toys.

I'll let @mkArtakMSFT chime in but I would hold off on merging RC1 content since we've drafted a substantial amount of new docs.

PR #19642, opened 2020-08-25
Blazor lazy loading renamed assemblies

### LINKED ISSUE #19628: [preview8] Add docs on interplay between lazy-loading and assembly renames
The `LazyLoadAssemblyService` works by interfacing with the ResourceLoader on the client. This ResourceLoader relies on the assembly names that are defined in the blazor.boot.json file.

When users rename assemblies following our guidelines, the assembly names they use in the code in their OnNavigateAsync methods and the assembly names in the blazor.boot.json file maybe out of sync.

To rectify this, users can:

- Check to see if the app is running in a production environment when determining what assembly names to use
- Store the renamed assembly names in a separate file and read from that file to determine what assembly name to use in both the LazyLoadAssemblyService and the rename

We might want to document this in the lazy-load docs and cross-link from the docs on renaming from deplyoment.

guardrex:
Fixes #19628

### DIFF
--- a/aspnetcore/blazor/host-and-deploy/webassembly.md
+++ b/aspnetcore/blazor/host-and-deploy/webassembly.md
@@ -5,7 +5,7 @@ description: Learn how to host and deploy a Blazor app using ASP.NET Core, Conte
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 08/03/2020
+ms.date: 08/25/2020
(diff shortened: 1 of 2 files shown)


guardrex:
I was just asking about that on the other PR. Artak said to go ahead and merge *immediately* those earlier Pre8 PRs before Pre8 landed. I was just asking if we're going to keep doing that. If so, we can just merge this now. If someone asks here, I guess I can say ... it will be for RC1.

guardrex:
Yes ... I'll split depending on Artak's response. If he says 'merge now' like he did last time, we're good. I'll just have to tell anyone who writes in 'where is this?' that it will arrive in a few weeks.

PR #19655, opened 2020-08-26
use JavaScript to sign out when browser is closed

### LINKED ISSUE #19404: How to sign the user out when browser tab or window is closed?
I implemented this ideas expressed in this documentation, but I have run into a problem with the user remaining logged in even after the close the browser tab. If more than one browser window is open, and they close the window with my app, they are also still logged in. They are only logged out if they close all the browser windows completely. 
There is no mention of this in the article, and there should be. A solution would also be beneficial for all, I feel. How does one code their app to sign out users if the user closes the browser tab or window?
Thanks for your time!

---
#### Document Details

* ID: d2229fc6-f8c8-952e-e567-b1d5a463055a
* Version Independent ID: b18a015b-914d-f014-f711-120d208904d9
* Content: [Use cookie authentication without ASP.NET Core Identity](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/cookie?view=aspnetcore-3.1)
* Content Source: [aspnetcore/security/authentication/cookie.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/security/authentication/cookie.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

Rick-Anderson:
Fixes #19404

### DIFF
--- a/aspnetcore/security/authentication/cookie.md
+++ b/aspnetcore/security/authentication/cookie.md
@@ -101,6 +101,8 @@ To sign out the current user and delete their cookie, call <xref:Microsoft.AspNe
 
 If `CookieAuthenticationDefaults.AuthenticationScheme` (or "Cookies") isn't used as the scheme (for example, "ContosoCookie"), supply the scheme used when configuring the authentication provider. Otherwise, the default scheme is used.
 
+The server has no control of the clients browser. If the user closes the browser or tab, the server cannot sign out the user. To implement signing out the user when the browser is closed, you must detect that with JavaScript. Search for "How to Detect Browser Window Tab Close Event?".
+
 ## React to back-end changes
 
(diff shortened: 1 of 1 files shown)


Tratcher:
Rick, did you see the users edit? They couldn't make this work.

> EDIT: Account to this poster, Javascript cannot detect when a window or tab is going to be closed, which halts my signing the user out when they close a tab. I sort of wish I had known this was a problem before implementing cookie authentication....

Rick-Anderson:
@Tratcher  should I remove 
**To implement signing out the user when the browser is closed, you must detect that with JavaScript. Search for "How to Detect Browser Window Tab Close Event?".**

And keep the rest?

Tratcher:
How about replacing the whole thing with "When the browser closes it will automatically delete session based cookies (non-persistent cookies), but no cookies are cleared when an individual tab is closed. The server is not notified of these events."

PR #19669, opened 2020-08-27
Update services.md

Rick-Anderson:
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/grpc/services?view=aspnetcore-3.0&branch=pr-en-us-19669)

minor nits only. Change `Simple` to `Basic`. We try to avoid simple.

### DIFF
--- a/aspnetcore/grpc/services.md
+++ b/aspnetcore/grpc/services.md
@@ -156,7 +156,7 @@ public override async Task StreamingFromServer(BasicRequest request,
 }
 ```
 
-The client has no way to send additional messages or data once the server streaming method has started. Some streaming methods are designed to run forever. In this situation, a client can cancel the call when they no longer need it. When cancellation happens the client sends a signal to the server and the [ServerCallContext.CancellationToken](xref:System.Threading.CancellationToken) is raised. The `CancellationToken` token should be used on the server with async methods so that:
+The client has no way to send additional messages or data once the server streaming method has started. Some streaming methods are designed to run forever. For continuous streaming methods, a client can cancel the call when it's no longer need it. When cancellation happens the client sends a signal to the server and the [ServerCallContext.CancellationToken](xref:System.Threading.CancellationToken) is raised. The `CancellationToken` token should be used on the server with async methods so that:
 
 * Any asynchronous work is canceled together with the streaming call.
 * The method exits quickly.
--- a/aspnetcore/grpc/services.md
+++ b/aspnetcore/grpc/services.md
@@ -123,7 +123,8 @@ Each call type has a different method signature. Overriding generated methods fr
 A unary method gets the request message as a parameter, and returns the response. A unary call is complete when the response is returned.
 
 ```csharp
-public override Task<BasicResponse> UnaryCall(BasicRequest request, ServerCallContext context)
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
I could move `ServerCallContext context)` to the next line to get rid of the horizontal scroll bar. Let me know if you'd like that change.

image

aspnetcore/grpc/services.md
```diff
@@ -103,16 +103,16 @@ syntax = "proto3";
 
 service ExampleService {
   // Unary
-  rpc UnaryCall (SimpleRequest) returns (SimpleResponse);
+  rpc UnaryCall (BasicRequest) returns (BasicResponse);
```
JamesNK, line 106: BasicRequest -> ExampleRequest
BasicResponse -> ExampleResponse

aspnetcore/grpc/services.md
```diff
@@ -145,96 +145,104 @@ message SimpleRequest {
...
+        await responseStream.WriteAsync(new BasicResponse());
         await Task.Delay(TimeSpan.FromSeconds(1));
     }
 }
 ```
 
-The client has no way to send additional messages or data once the server streaming method has started. Some streaming methods are designed to run forever. In this situation a client can cancel the call when they no longer need it. When cancellation happens the client sends a signal to the server and the `ServerCallContext.CancellationToken` <xref:System.Threading.CancellationToken> is raised. The token should be used on the server with async methods used so that any asynchronous work is canceled together with the streaming call, and the method exits quickly.
+The client has no way to send additional messages or data once the server streaming method has started. Some streaming methods are designed to run forever. For continuous streaming methods, a client can cancel the call when it's no longer need it. When cancellation happens the client sends a signal to the server and the [ServerCallContext.CancellationToken](xref:System.Threading.CancellationToken) is raised. The `CancellationToken` token should be used on the server with async methods so that:
```
JamesNK, line 159:
```suggestion
The client has no way to send additional messages or data once the server streaming method has started. Some streaming methods are designed to run forever. For continuous streaming methods, a client can cancel the call when it's no longer needed. When cancellation happens the client sends a signal to the server and the [ServerCallContext.CancellationToken](xref:System.Threading.CancellationToken) is raised. The `CancellationToken` token should be used on the server with async methods so that:
```

JamesNK (approved):

aspnetcore/grpc/services.md
```diff
@@ -123,17 +123,17 @@ Each call type has a different method signature. Overriding generated methods fr
 A unary method gets the request message as a parameter, and returns the response. A unary call is complete when the response is returned.
 
 ```csharp
-public override Task<SimpleResponse> UnaryCall(SimpleRequest request, ServerCallContext context)
+public override Task<BasicResponse> UnaryCall(BasicRequest request, ServerCallContext context)
```
JamesNK, line 126:
```suggestion
public override Task<BasicResponse> UnaryCall(BasicRequest request,
    ServerCallContext context)
```

aspnetcore/grpc/services.md
Rick-Anderson: Oops, I missed this. I'll change all of them from Basic to Example in your PR

PR #19683, opened 2020-08-27
Update CoursesController.cs

Raredall:
重定向代码位置错误，应该在保存更改后重定向


Rick-Anderson:
> The redirect code is in the wrong location, it should be redirected after saving the changes

Because it's inside the block, it doesn't change the code flow. I agree it might be better where you have it but I'll let the author @tdykstra  decide.

PR #19691, opened 2020-08-28
Add Create JSON Web APIs from gRPC doc

JamesNK:
Discuss gRPC HTTP API project.


JamesNK:
@shirhatti

shirhatti:
I'd like to see a mention that unlike gRPC (or gRPC-web), with gRPC HTTP API  there is no experience for generation strongly-typed clients. Maybe a sample HTTP Request/Response will help clear that up?

It wasn't clear to me if this the only gRPC HTTP API or there are more changes. Regardless, it isn't important enough to block merging.

PR #19700, opened 2020-08-28
Use GetTestClient() instead of GetTestServer().CreateClient()

### LINKED ISSUE #19645: Use GetTestClient instead of GetTestServer().CreateClient()?
I was playing around with some of the stuff in this topic today, and I noticed the following:

```csharp
host.GetTestServer().CreateClient()
```

-can be replaced with-:

```csharp
host.GetTestClient()
```

It's only a minor point, but is it worth updating the example to use this approach?

---
#### Document Details

* ID: 9421feee-5207-e379-00fb-847cc880014c
* Version Independent ID: 8415adee-793e-2ba7-b97b-672b493ba42b
* Content: [Test ASP.NET Core middleware](https://docs.microsoft.com/en-us/aspnet/core/test/middleware?view=aspnetcore-3.1)
* Content Source: [aspnetcore/test/middleware.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/test/middleware.md)
* Product: **aspnet-core**
* GitHub Login: @Tratcher
* Microsoft Alias: **riande**

serpent5:
Fixes #19645.

I'm not sure that the original highlights are quite right. i.e. There's a highlight for the _CreateClient + Get_, then just the _CreateClient_, then the _Assert_. I'd say the first should be the (new) _CreateTestClient + Get_ line (there's no Assert), then the next should be the _Assert.NotEqual_ line, and then the _Assert.Equal_ line. Does that make sense?

cc @guardrex

### DIFF
--- a/aspnetcore/test/middleware.md
+++ b/aspnetcore/test/middleware.md
@@ -43,6 +43,7 @@ In the test project, create a test:
 [!code-csharp[](middleware/samples_snapshot/3.x/setup.cs?highlight=4-18)]
 
 ## Send requests with HttpClient
+
 Send a request using <xref:System.Net.Http.HttpClient>:
 
 [!code-csharp[](middleware/samples_snapshot/3.x/request.cs?highlight=20)]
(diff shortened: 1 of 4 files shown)


guardrex:
What about Line 49?

https://docs.microsoft.com/aspnet/core/test/middleware#send-requests-with-httpclient

> Does that make sense?

They (the sample code) may have changed at some point without the highlights being updated in the topic. *It wouldn't be the first time.* 🙈😄

guardrex:
Also, let's version the whole topic for 3.1 in the metadata ...

```
```

.... under the `description` entry.

serpent5:
I was so close to asking about that!

I'll add a commit with the line highlights to just show what I was trying to explain. I don't think I did a good job of that. 😊

guardrex:
If you were saying that they don't look right **_today_**, that's not really surprising. Make whatever fix makes sense.

guardrex:
Looks good ... all set on this one?

serpent5:
Yep. All done. 🥳

serpent5:
Thanks for the help, and for trying to wrap your head around my poor explanations. 😆

PR #19701, opened 2020-08-28
Remove point re injecting transients into singletons/scopeds

### LINKED ISSUE #19682: Scope validation section should be updated
There are 3 points for scope validation:
- Scoped services aren't resolved from the root service provider.
- Scoped services aren't injected into singletons.
- Transient services aren't injected into singletons or scoped services.

But the last one is not true, transient service can be injected to singletons or scoped service (https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/web-host?view=aspnetcore-3.1#scope-validation)

---
#### Document Details

* Content: [Dependency injection in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-3.1)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

serpent5:
Fixes #19682.

cc @Rick-Anderson

### DIFF
--- a/aspnetcore/fundamentals/dependency-injection.md
+++ b/aspnetcore/fundamentals/dependency-injection.md
@@ -331,7 +331,6 @@ When the app runs in the [Development environment](xref:fundamentals/environment
 
 * Scoped services aren't resolved from the root service provider.
 * Scoped services aren't injected into singletons.
-* Transient services aren't injected into singletons or scoped services.
 
 The root service provider is created when <xref:Microsoft.Extensions.DependencyInjection.ServiceCollectionContainerBuilderExtensions.BuildServiceProvider%2A> is called. The root service provider's lifetime corresponds to the app's lifetime when the provider starts with the app and is disposed when the app shuts down.
 


Rick-Anderson:
Doesn't injecting transient into singleton make the transient service singleton?

serpent5:
> 
> 
> Doesn't injecting transient into singleton make the transient service singleton?

Yep, it does - that's why I thought it was valid. I tried it all out, and the scope validation doesn't check for it. I found an answer on SO that helped clear it up (I think): https://stackoverflow.com/a/56291583.

serpent5:
There's this bullet in the recommendations section of this topic:

> Disposable transient services are captured by the container for disposal. This can turn into a memory leak if resolved from the top level container.

That suggests that it's _valid_, but I guess it's saying that disposables should be treated with caution. There's a whole explanation about using factories for transient disposables, so it ties together. Ultimately, the issue here was that the topic claimed the scope-validation logic checked for this, but it doesn't.

PR #19753, opened 2020-09-01
Remove inter-process communication doc

JamesNK:
ConnectionFactory is being removed from .NET 5 - https://github.com/dotnet/runtime/pull/41648

IPC will have to wait for .NET 6, and the doc on it needs to be removed.

I'm removing the doc rather than changing the version to `>= net6` because the API might be different when it is re-added.


guardrex:
> \>= net6

The other factor on that is that these identifiers are manually created, and I don't think that one exists yet.

Issue #19760, opened 2020-09-02
Add detail about string and ByteString


jskeet:
It might be worth talking a bit about `string` and `ByteString` properties in more detail:

- The default to empty values, and can't be set to null references
- There are "wrapper" types for them, just like numeric primitives, which generate properties with the same types, but then they *can* be set to null references

You might also want to describe `ByteString` itself just for a bit of context.

Finally, the example of using "Add" for a collection may look a bit odd without the motivation, which is to be able to use it in object initializers, e.g.

```csharp
Person person = new Person
{
    Roles = { roles.Where(role => role.SomeCondition) }
}
```

It looks a little odd to start with, but it's super-helpful.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: fc570f1c-bb62-88b8-fbfc-ca4e436e8091
* Version Independent ID: 4ce52659-cd0f-0172-2f8c-e58b8497eadc
* Content: [Create Protobuf messages for .NET apps](https://docs.microsoft.com/en-us/aspnet/core/grpc/protobuf?view=aspnetcore-3.1)
* Content Source: [aspnetcore/grpc/protobuf.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/grpc/protobuf.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-grpc**
* GitHub Login: @JamesNK
* Microsoft Alias: **jamesnk**

PR #19769, opened 2020-09-03
Protobuf scalar default values and ByteString detail

JamesNK:
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/19760

@jskeet

### DIFF
--- a/aspnetcore/grpc/protobuf.md
+++ b/aspnetcore/grpc/protobuf.md
@@ -69,6 +69,10 @@ Protobuf supports a range of native scalar value types. The following table list
 | `string`      | `string`     |
 | `bytes`       | `ByteString` |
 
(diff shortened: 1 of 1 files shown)


davidfowl:
Performance warning: this copy copies byte[] how many times?

JamesNK:
`ByteString.FromBytes` and `BytesString.ToByteArray` are an allocation plus copy.

For reading values, Memory/Span properties exist. For creating values there is discussion about adding an unsafe API - https://github.com/protocolbuffers/protobuf/pull/7645

If it does get added I don't think it is appropriate to talk about in this doc. The performance guide would be a better place.

jskeet (approved):
LGTM - thanks James.

PR #19775, opened 2020-09-03
Update iauthorizationpolicyprovider.md

christianmgreen:
The example on line 79 did not match the name of the AuthorizeAttribute ("MinimumAgeAuthorizeAttribute") on line 49.
More specifically, it was missing the "Attribute" suffix.


Rick-Anderson:
The suffix `AuthorizeAttribute` can be dropped when it's used in an attribute like `[MinimumAgeAuthorize(10)]`

`[MinimumAgeAuthorize(` is used 3 times in the topic.

PR #19778, opened 2020-09-04
Enhance Blazor start content

### LINKED ISSUE #15373: Rename and expand the `Configure the SignalR client for Blazor Server apps` to cover more cases
https://docs.microsoft.com/en-us/aspnet/core/blazor/hosting-models?view=aspnetcore-3.0#configure-the-signalr-client-for-blazor-server-apps

* Rename the section above to "Manually starting the Blazor application"
* Keep the information in the current section.
* Move the "Configure the SignalR client for Blazor Server apps" to a sub-section.
* Add a section/snippet to initialize Blazor when the document is ready.
* Add a section/snippet to demonstrate you can chain to the Promise resulting from Blazor.start({...}) to perform some other tasks, like JS interop initialization, etc.

---
#### Document Details

* Content: [ASP.NET Core Blazor hosting models](https://docs.microsoft.com/en-us/aspnet/core/blazor/hosting-models?view=aspnetcore-3.0#configure-the-signalr-client-for-blazor-server-apps)
* Content Source: [aspnetcore/blazor/hosting-models.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/hosting-models.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #15373

[Internal Review Topic (links to section)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/additional-scenarios?view=aspnetcore-3.1&branch=pr-en-us-19778#manual-start-of-the-blazor-app)

😕 ... I've run into a point of confusion that's even blocking this PR ...

Our content implies (and I was under the assumption) that when `autostart="false"` is placed on the script that calling `Blazor.start()` is absolutely required to get the app going. Testing here seems to indicate that that's not so. When I set the `_Host.cshtml` page up with **_only_** the following ...

```cshtml
<script autostart="false" src="_framework/blazor.server.js"></script>
```

and no call to `Blazor.start()`, the **_app still starts and runs normally_**, which I don't expect to happen.

What am I missing? ..... I'll adjust the PR accordingly to head off the (**_my_** :smile:) confusion.

### DIFF
--- a/aspnetcore/blazor/fundamentals/additional-scenarios.md
+++ b/aspnetcore/blazor/fundamentals/additional-scenarios.md
@@ -121,7 +121,7 @@ Configure the manual start of a Blazor Server app's [SignalR circuit](xref:blazo
 * Add an `autostart="false"` attribute to the `<script>` tag for the `blazor.server.js` script.
 * Place a script that calls `Blazor.start` after the `blazor.server.js` script's tag and inside the closing `</body>` tag.
 
-When `autostart` is disabled, any aspect of the app that doesn't depend on the circuit works normally. For example, client-side routing is operational. However, any aspect that depends on the circuit isn't operational until `Blazor.start` is called. For example, component methods fail to execute when called while the circuit is disconnected.
+When `autostart` is disabled, any aspect of the app that doesn't depend on the circuit works normally. For example, client-side routing is operational. However, any aspect that depends on the circuit isn't operational until `Blazor.start` is called. For example, component methods fail to execute while the circuit is disconnected.
 
 ### Initialize Blazor when the document is ready
 
--- a/aspnetcore/blazor/fundamentals/additional-scenarios.md
+++ b/aspnetcore/blazor/fundamentals/additional-scenarios.md
@@ -116,16 +116,20 @@ Rendering server components from a static HTML page isn't supported.
 
 *This section applies to Blazor Server.*
 
-Configure the manual start of a Blazor Server app in the `Pages/_Host.cshtml` file:
+Configure the manual start of a Blazor Server app's [SignalR circuit](xref:blazor/hosting-models#circuits) in the `Pages/_Host.cshtml` file:
 
 * Add an `autostart="false"` attribute to the `<script>` tag for the `blazor.server.js` script.
-* Place a script that calls `Blazor.start` after the `_framework/blazor.server.js` script and inside the closing `</body>` tag.
+* Place a script that calls `Blazor.start` after the `blazor.server.js` script's tag and inside the closing `</body>` tag.
+
+When `autostart` is disabled, any aspect of the app that doesn't depend on the circuit works normally. For example, client-side routing is operational. However, any aspect that depends on the circuit isn't operational until `Blazor.start` is called. For example, component methods fail to execute when called while the circuit is disconnected.
 
 ### Initialize Blazor when the document is ready
 
 To initialize the Blazor app when the document is ready:
 
 ```cshtml
+<body>
+
     ...
 
     <script autostart="false" src="_framework/blazor.server.js"></script>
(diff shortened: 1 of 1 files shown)


captainsafia:
@guardrex Content here looks good.

I wasn't able to reproduce the issue you cited. My repro steps are as follows:

1. Create new project via `dotnet new blazorserver -o AutostartTest`
2. Modify `Pages/_Host.cshtml` to set `autostart='false'`.
3. Observe that the circuit isn't initialized.

This was on `5.0.100-rc.1.20431.3`. 

Is there a chance that you have a `Blazor.start` invocation somewhere else in the code?

guardrex:
"Circuit" ... I expected to see the warning bar that a circuit hasn't been established. Turns out that there's no indication ... just everything that would depend on the circuit no-ops. Therefore, I'd like to look again at the text here. It should probably say somehow explicitly that "start" in `Blazor.start` and `autostart` refers to 'starting the circuit' and perhaps that only functionality that depends on the circuit is impacted by an disabled circuit ... something like that ... I'll work on the wording. I'll update it tomorrow (Tuesday) morning and ping u back.

guardrex:
@captainsafia ... Here's the latest opening remarks of that section ...

> Configure the manual start of a Blazor Server app's \[SignalR circuit](xref:blazor/hosting-models#circuits) in the \`Pages/_Host.cshtml\` file:
>
> * Add an \`autostart="false"\` attribute to the \`\<script>\` tag for the \`blazor.server.js\` script.
> * Place a script that calls \`Blazor.start\` after the \`blazor.server.js\` script's tag and inside the closing \`\</body>\` tag.
>
> When \`autostart\` is disabled, any aspect of the app that doesn't depend on the circuit works normally. For example, client-side routing is operational. However, any aspect that depends on the circuit isn't operational until \`Blazor.start\` is called. For example, component methods fail to execute when called while the circuit is disconnected.

guardrex:
@captainsafia ... does this look good :point_up:?

aspnetcore/blazor/fundamentals/additional-scenarios.md
```diff
@@ -190,21 +235,22 @@ To adjust the reconnection retry count and interval:
 </body>
 ```
 
-### Hide or replace the reconnection display
+## Hide or replace the reconnection display
```
captainsafia, line 238: We recently added a feature that allowed a user to customize the delay before the reconnection display would show up. We might want to document this here as well. Users can modify this by setting the `transition-delay` property in the CSS for the modal element.

```
'transition: visibility 0s linear 500ms'
```

The default is above which delays the display by 500ms second. Users can modify this value to delay it.

aspnetcore/blazor/fundamentals/additional-scenarios.md
```diff
@@ -112,20 +112,62 @@ Blazor Server apps are set up by default to prerender the UI on the server befor
...
 
-Configure the SignalR client used by Blazor Server apps in the `Pages/_Host.cshtml` file. Place a script that calls `Blazor.start` after the `_framework/blazor.server.js` script and inside the `</body>` tag.
+Configure the manual start of a Blazor Server app's [SignalR circuit](xref:blazor/hosting-models#circuits) in the `Pages/_Host.cshtml` file:
 
-### Logging
+* Add an `autostart="false"` attribute to the `<script>` tag for the `blazor.server.js` script.
+* Place a script that calls `Blazor.start` after the `blazor.server.js` script's tag and inside the closing `</body>` tag.
 
-To configure SignalR client logging:
+When `autostart` is disabled, any aspect of the app that doesn't depend on the circuit works normally. For example, client-side routing is operational. However, any aspect that depends on the circuit isn't operational until `Blazor.start` is called. For example, component methods fail to execute while the circuit is disconnected.
```
captainsafia, line 124: I like the idea of using an example here but I think it doesn't convey the severity of not having an initialized Blazor circuit enough. Maybe replace the example with a warning that application behavior will be unpredictable if the circuit is not trusted.

captainsafia (commented):
Nice additions! Left some feedback inline.

guardrex:
@captainsafia ... Confirm the CSS ... I'm confused by Steve's PR and what you just said. Is it ......

```css
#components-reconnect-modal {
    transition: visibility 0s linear 500ms;
}
```

PR #19831, opened 2020-09-12
fix duplicate headers

Rick-Anderson:

### DIFF
--- a/aspnetcore/data/ef-rp/complex-data-model.md.bak
+++ b/aspnetcore/data/ef-rp/complex-data-model.md.bak
@@ -26,7 +26,7 @@ The completed data model is shown in the following illustration:
 
 ![Entity diagram](complex-data-model/_static/diagram.png)
 
-## The Student entity 3.1
+## The Student entity
 
 ![Student entity](complex-data-model/_static/student-entity.png)
 
@@ -43,11 +43,11 @@ The preceding code adds a `FullName` property and adds the following attributes 
 * `[Required]`
 * `[Display]`
 
-### The FullName calculated property 3.1
+### The FullName calculated property
 
 `FullName` is a calculated property that returns a value that's created by concatenating two other properties. `FullName` can't be set, so it has only a get accessor. No `FullName` column is created in the database.
 
-### The DataType attribute 3.1
+### The DataType attribute
 
 ```csharp
 [DataType(DataType.Date)]
@@ -62,7 +62,7 @@ The [DataType](/dotnet/api/system.componentmodel.dataannotations.datatypeattribu
 
 The `DataType` attribute emits HTML 5 `data-` (pronounced data dash) attributes. The `DataType` attributes don't provide validation.
 
-### The DisplayFormat attribute 3.1
+### The DisplayFormat attribute
 
 ```csharp
 [DisplayFormat(DataFormatString = "{0:yyyy-MM-dd}", ApplyFormatInEditMode = true)]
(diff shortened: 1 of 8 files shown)


guardrex:
No other options? ... *version-by-file* ... wasn't it coming for us as an option? It would solve this problem if we had it, right?

Rick-Anderson:
> No other options? ... *version-by-file* ... wasn't it coming for us as an option? It would solve this problem if we had it, right?

It's easy to remove these when we get version by file. That's going to be alot more work that this approach. I think once we get version by file, we can use that approach when we do a major update of a topic.

guardrex:
How does this work with bookmarks to heading ids? I know you prefer the more stable but not-HTML5 anchors, but there are a lot of header id-based bookmarks around.

Rick-Anderson:
> How does this work with bookmarks to heading ids? I know you prefer the more stable but not-HTML5 anchors, but there are a lot of header id-based bookmarks around.

Obviously they're not working for the non released version. But their not working now with all the duplicates. Important ones can be fixed.

guardrex:
I feel kind'a lucky that for Blazor topics because either *whole topics* are for 5.0 (no dup headers) or only small bits of topics require 5.0 versioning of text/new sections (also no dups). The only one that's an issue is the Blazor EF Core topic. I'll tend to that on Monday along these lines or see if it's possible to just juggle the content around the monikers and headings that he has in there such that those few headings aren't duplicated.

I ❤️ the concept of *whole file versioning*. I'd enjoy using it I think. 🤔

Rick-Anderson:
@scottaddie  what's the status of whole file versioning? How much work is the conversion?

wadepickett:
@Rick-Anderson, if the fix until we have whole doc versioning is adding an ID.  Did we decide what the ID's should be for sure yet?   I am wondering if there is an ID scheme that would be helpful for identifying later such as reflecting the content version it is in, or perhaps a sequential number that starts with the first copy.  So it is the 2nd or 3rd duplicate.   Just a thought, perhaps that generates more work than it saves.  

My vote on priority: The sooner we can do whole file versioning the better: for archiving, for lost time dealing with lengthy inline "spaghetti-versioning" topics on a daily basis, as well as just to keep from endlessly running afoul of a validation/publishing system that expects it.

Rick-Anderson:
ID is exactly what's generated by default + version number. It's not "spaghetti-versioning", it's explicit. Change the heading text and the link stays the same.

guardrex:
Rick and I chatted about it offline. Just so you, Wade and Scott, know what we were discussing ...

When I mentioned "bookmarks to heading ids," I was referring to how markdown headers are rendered.

So ...

```
## Something
```

... produces ...

```html
<h2 id="something">Something</h2>
```

... and links with ...

```
<xref:UID#something>
```

... break if the heading goes to ...

```
## Something 3.1
```

However, an option similar to using old skool anchors would be to just drop the markdown headings in favor of HTML heading tags and set the `id` explicitly with a permanent value, which would leave the text free to change without breaking links in the future in an HTML5-friendly way ...

```html
<h2 id="permanent-id">Something</h2>
```

The only thing lost is the fancy markdown syntax (`##`).

Anyway ... a moot approach (probably 🤞) if we get *version by file* .... I'm ❤️ to give it a shot.

Rick-Anderson:
Per `@tdykstra`, they are going to fix this in the platform tools OR something similar. Holding off for now.

Issue #17491, opened 2020-03-27
Anchors for .NET Generic Host all have a -1 suffix


serpent5:
I'm in the process of updating the Fundamentals Overview (#13877), where I'm looking at updating some of the links that currently target the Web Host topic to target this topic. I noticed that the anchors in this topic all have a -1 suffix, which doesn't seem right. It also means that some of the internal links in this topic don't work. For example, in the [IHostEnvironment](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host?view=aspnetcore-3.1#ihostenvironment-1) section, 4 out of the 5 links are internal links, and they're all broken. I've checked a couple of other topics, which seem fine.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: c5f1eeee-d500-82bf-0d0a-b4e0d5d5c175
* Version Independent ID: 2373a3df-6ea3-50c7-d63d-61c51d572511
* Content: [.NET Generic Host](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host?view=aspnetcore-3.1)
* Content Source: [aspnetcore/fundamentals/host/generic-host.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/fundamentals/host/generic-host.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-fundamentals**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
The `-1` suffix from from repeating  the same `## H2's` with the monikered versions. Because of duplicate H2's, one bookmark stays as predicted, and the other has The `-1` suffix 
I believe the first version of the H2 stays as predicted, and the 2nd has The `-1` suffix  appended. That's why I always put the new version on top.

I'm surprised we haven't had any complaints on broken links.

I'm not sure if file versioning will fix that problem. CC @scottaddie

serpent5:
Right, I see. There's a >= 5.0 version of this topic, so that gets e.g. `#webroot`, the >=3.0 version gets e.g. `#webroot-1`, and the < 3.0 version gets e.g. `#webroot-2`.

[...]

PR #19847, opened 2020-09-14
Drop markdown headings to solve dup headings

### LINKED ISSUE #19846: Fix Blazor-EF Core dup headings
---
#### Document Details

* ID: 4a101e4d-85a2-8013-0f62-40e13e45b3c6
* Version Independent ID: 66a0a920-ebd8-9b1d-c8c5-953ebb766de0
* Content: [ASP.NET Core Blazor Server with Entity Framework Core (EFCore)](https://docs.microsoft.com/en-us/aspnet/core/blazor/blazor-server-ef-core?view=aspnetcore-3.1)
* Content Source: [aspnetcore/blazor/blazor-server-ef-core.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/blazor-server-ef-core.md)
* Product: **aspnet-core**
* GitHub Login: @JeremyLikness
* Microsoft Alias: **jeliknes**

guardrex:
Fixes  #19846

This is **_temporary_** until we can lay hands on **_version-by-file_**. At that point, I'll revert back to markdown-style headings.

cc: @Rick-Anderson @scottaddie @wadepickett ... The HTML tag headings approach bypasses the build system, but I did need to apply different tag ids. This approach provides stable text across sections without a visible version to readers. I'll just do this temporarily until hopefully 🤞 we get *version-by-file* and then revert back to plain markdown headings. 

This is the only Blazor topic with this problem right now, so it's trivial to revert later. I'll see how Blazor WASM security topics play out when I work the content for RC1. It might not be that bad. There might not be many changes, so whole-topic versioning might not be required. If so, I don't expect new dup heading issues.

### DIFF
--- a/aspnetcore/blazor/blazor-server-ef-core.md
+++ b/aspnetcore/blazor/blazor-server-ef-core.md
@@ -20,7 +20,7 @@ Blazor Server is a stateful app framework. The app maintains an ongoing connecti
 > [!NOTE]
 > This article addresses EF Core in Blazor Server apps. Blazor WebAssembly apps run in a WebAssembly sandbox that prevents most direct database connections. Running EF Core in Blazor WebAssembly is beyond the scope of this article.
 
-<h2 id="sample-app">Sample app</h2>
+<h2 id="sample-app-5x">Sample app</h2>
 
 The sample app was built as a reference for Blazor Server apps that use EF Core. The sample app includes a grid with sorting and filtering, delete, add, and update operations. The sample demonstrates use of EF Core to handle optimistic concurrency.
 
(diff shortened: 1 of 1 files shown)


serpent5:
@guardrex Is this the same problem I saw earlier in the year in #17491? Can you point me to more information re how _version-by-file_ will work (if you get it)?

guardrex:
Scott will let us know ... track on ...

https://github.com/dotnet/AspNetCore.Docs/pull/19831#issuecomment-691372178

Yes ... there are several pain points with *whole topic versioning*. We think *version by file* will fix the dup headings problem, the cross-link problem, and the dup `alt` text problem. It probably will be nicer for just performing general updates when you know what version you need to apply a change to. We'll just open the **_file_** for that version instead of having to search for content under a moniker range within a multi-version file.

guardrex:
We kind'a knew from the start that there would be some problems ... it's **very clear** that moniker ranges were never designed to be used for whole topic versioning. I think it's clear at this point that they either want us to version little bits of content (paragraphs, code snippets, etc.) or use version by file. However, massive churn across versions, especially minor releases, makes the topic insane to maintain flipping in and out of versions. The Kestrel topic was probably the worst back when it had like **four versions** and was flipping in and out of different moniker ranges every few paragraphs. 😨 It cleaned up very well when we went to whole-topic, but now all of these new problems have come up.

serpent5:
Ok, understood. Thanks for the information - I've subscribed to the linked issue to see how things progress.

PR #19917, opened 2020-09-17
Move Blazor built-in components down in ToC

guardrex:
Scott ...

I'm still not getting the best behavior out of the ToC with the built-in component links embedded near the top of the *Components* node. The problem is still that when a reader selects one they end up with a highlighted ToC entry of the component in the *Built-in components* list and not the highlighted ToC entry of **_the topic where that link sent them_**.

I think what's happening is that it's matching the first `href`/`uid` (URL) it comes to down the ToC. I think that by moving the whole node to the bottom of the Blazor ToC that it will start matching, thus highlighting, the **_topic_** where the component is described.

If you think I'm right, let's give this PR shot. If you have a better plan to achieve the behavior I seek, I'm all :ear::ear::ear::ear::ear:.

**UPDATE**: What about setting `topicHref`?

> If topicHref is set for this TOC Item, it will be considered as the href of the expanded TOC Item.

... that sounds good if I'm interpreting it correctly as matching my goal (i.e., set it to the landing topic). 🤔 Depends on what they mean by "it" ... pronouns ... they're great ... *until they aren't*. :smile:

### DIFF
--- a/aspnetcore/toc.yml
+++ b/aspnetcore/toc.yml
@@ -325,48 +325,6 @@
           items:
             - name: Overview
               uid: blazor/components/index
-            - name: Built-in components
-              items:
-                - name: App
-                  href: blazor/templates.md#blazor-project-structure
-                - name: Authentication
-                  href: blazor/security/webassembly/index.md#authentication-component
-                - name: AuthorizeView
-                  href: blazor/security/index.md#authorizeview-component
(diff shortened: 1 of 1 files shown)


scottaddie:
@guardrex `topicHref` sounds promising. I recommend trying that.

PR #19968, opened 2020-09-22
Add a note about model binding record types

pranavkm:

### DIFF
--- a/aspnetcore/mvc/models/model-binding.md
+++ b/aspnetcore/mvc/models/model-binding.md
@@ -381,10 +381,12 @@ For `Dictionary` targets, model binding looks for matches to *parameter_name* or
 
 <a name="glob"></a>
 ::: moniker range="> aspnetcore-5.0"
-## Constructor binding and record types
-Model binding requires that complex types have a parameterless constructor. Both System.Text.Json and Newtonsoft.Json-based input formatters support deserialization of classes that do not have a parameterless constructor. 
 
-C# 9 introduces record types which are a great way to succintly represent data over the wire. MVC adds support for model binding and validating record types with a single constructor.
+## Constructor binding and record types
+
+Model binding requires that complex types have a parameterless constructor. Both `System.Text.Json` and `Newtonsoft.Json` based input formatters support deserialization of classes that do not have a parameterless constructor. 
+
+C# 9 introduces record types which are a great way to succintly represent data over the network. MVC adds support for model binding and validating record types with a single constructor.
 
 ```C#
 public record Person([Required] string Name, [Range(0, 150)] int Age);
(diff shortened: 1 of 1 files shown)


Rick-Anderson (approved):
Slight edit. Thanks for doing this.

Rick-Anderson:
@pranavkm  what about Razor Pages?

pranavkm:
It should be identical, no changes there.

Rick-Anderson:
@pranavkm  see my last [commit](https://github.com/dotnet/AspNetCore.Docs/pull/19968/commits/20a08170db3325f57df1270cda8598233b50efc7)

pranavkm:
Looks great. I have some additional write up about what works and does not that was created as part of the PR: https://github.com/dotnet/aspnetcore/pull/23976. Pasting it here because that PR takes forever to load. In case you think there's anything useful for you to add to the docs:

-------------------------------

A summary of what works and what does not:

## We allow binding to and validating record types.
```
public record Person([Required] string Name, [Range(0, 100)] int Age);
```
For this to work, we require that the type

* Is a record type,
* has exactly one public constructor
* every parameter has a property with the same name (no difference in casing) and type


### We do not bind to regular POCOs that do not have parameterless constructors.
e.g., this will result in an exception saying that the type must have a parameterless constructor.

```
public class Person(string Name)
When a record type has more than one constructor, binding will fail stating this is unsupported:
public record Person([Required] string Name, [Range(0, 100)] int Age)
{
   public Person(string Name) : this (Name, 0);
}
```

### Record types with manually authored constructors that look like primary constructors will work:

```
public record Person
{
   public Person([Required] string Name, [Range(0, 100)] int Age) => (this.Name, this.Age) = (Name, Age);

   public string Name { get; set; }
   public int Age { get; set; }
}
```

### For record types, validation and binding metadata on parameters is used. Any metadata on properties is ignored.

```
public record Person (string Name, int Age)
{
   [BindProperty(Name = "SomeName")] // This does not get used
   [Required] // This does not get used
   public string Name { get; init; }
}
```

### Validation uses metadata on the parameter but uses the property to read the value. In the ordinary case with primary constructors, the two would be identical. However, there are ways to defeat it:
```
public record Person([Required] string Name)
{
   private readonly string _name;
   public Name { get; init => _name = value ?? string.Empty; } // Now this property is never null. However this object could have been constructed as `new Person(null);`
}
```

### TryUpdateModel does not update parameters on a record type
```
public record Person(string Name)
{
   public int Age { get; set; }
}

var person = new Person("initial-name");
TryUpdateModel(person, ...);
```

In this case, MVC will not attempt to bind Name again. However, Age is allowed to be updated

Rick-Anderson:
@pranavkm  fantastic. I'll take care of it. I created #19969 for tracking.

PR #19972, opened 2020-09-23
Update linux-nginx.md

tunggnu:
Modify the /etc/nginx/sites-available/default file to redirect all HTTP requests to HTTPS properly

### DIFF
--- a/aspnetcore/host-and-deploy/linux-nginx.md
+++ b/aspnetcore/host-and-deploy/linux-nginx.md
@@ -386,7 +386,7 @@ Edit the */etc/nginx/nginx.conf* configuration file. The example contains both `
 > [!NOTE]
 > Blazor WebAssembly apps require a larger `burst` parameter value to accommodate the larger number of requests made by an app. For more information, see <xref:blazor/host-and-deploy/webassembly#nginx>.
 
-Update the `/etc/nginx/sites-available/default` file to redirect HTTP requests to HTTPS. The following example redirects requests for `example.com`:
+Update the `/etc/nginx/sites-available/default` file to redirect HTTP requests to HTTPS. The following example redirects requests for `example.com`/`*.example.com`:
 
 ```nginx
 server {
--- a/aspnetcore/host-and-deploy/linux-nginx.md
+++ b/aspnetcore/host-and-deploy/linux-nginx.md
@@ -386,6 +386,17 @@ Edit the */etc/nginx/nginx.conf* configuration file. The example contains both `
 > [!NOTE]
 > Blazor WebAssembly apps require a larger `burst` parameter value to accommodate the larger number of requests made by an app. For more information, see <xref:blazor/host-and-deploy/webassembly#nginx>.
 
+You now have configured HTTPS for `example.com`. Then edit the */etc/nginx/sites-available/default* file again to redirect all HTTP requests to HTTPS:
+
+```nginx
+server {
+    listen        80;
+    server_name   example.com *.example.com;
+    add_header    Strict-Transport-Security max-age=15768000;
+    return        301 https://example.com$request_uri;
+}
+```
+
 #### Secure Nginx from clickjacking
 
 [Clickjacking](https://blog.qualys.com/securitylabs/2015/10/20/clickjacking-a-common-implementation-mistake-that-can-put-your-websites-in-danger), also known as a *UI redress attack*, is a malicious attack where a website visitor is tricked into clicking a link or button on a different page than they're currently visiting. Use `X-FRAME-OPTIONS` to secure the site.
--- a/aspnetcore/host-and-deploy/linux-nginx.md
+++ b/aspnetcore/host-and-deploy/linux-nginx.md
@@ -386,7 +386,7 @@ Edit the */etc/nginx/nginx.conf* configuration file. The example contains both `
 > [!NOTE]
 > Blazor WebAssembly apps require a larger `burst` parameter value to accommodate the larger number of requests made by an app. For more information, see <xref:blazor/host-and-deploy/webassembly#nginx>.
 
-You now have configured HTTPS for `example.com`. Then edit the */etc/nginx/sites-available/default* file again to redirect all HTTP requests to HTTPS:
+Update the `/etc/nginx/sites-available/default` file to redirect HTTP requests to HTTPS. The following example redirects requests for `example.com`:
 
 ```nginx
 server {


tunggnu:
We have to update the */etc/nginx/sites-available/default* file after configuring HTTPS. If not, all HTTP requests are still captured and processed by the server block in the */etc/nginx/sites-available/default* file. The following server block in the *linux-nginx/nginx.conf* file is unable to capture HTTP requests as expected because it doesn't have a proper server_name directive. It captures only HTTP requests with an empty HOST header.

```
server {
    listen     *:80;
    add_header Strict-Transport-Security max-age=15768000;
    return     301 https://$host$request_uri;
}
```

guardrex (approved):

guardrex:
Thanks @tunggnu ... We'll have Sourabh look this over and then we can see about getting it merged.

shirhatti:
Any HSTS guidance that sets max-age to a value greater than zero or setting up a 301 in production needs a big bold warning of how it impacts clients.

shirhatti:
EDIT 2: I see this content is already part of the doc. See
https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/host-and-deploy/linux-nginx/nginx.conf#L14-L18

guardrex:
... but @shirhatti ...

> The following server block in the `linux-nginx/nginx.conf` file is unable to capture HTTP requests as expected because it doesn't have a proper server_name directive. It captures only HTTP requests with an empty HOST header.

... should that block contain ...

```
server_name               example.com;
```

... to match the other `server` entry?

shirhatti:
> ... should that block contain ...

```
server_name example.com;
```

Yes! It should contain the `server_name` directive. When not specificed, nginx uses the empty name as the server name and that will not match

> There isn't such a WARNING in the doc. Do you have text in mind that I can drop in?

Copying some verbiage from https://docs.microsoft.com/aspnet/core/security/enforcing-ssl

> We recommend using temporary redirects rather than permanent redirects. Link caching can cause unstable behavior in development environments.

> For production environments that are implementing HTTPS for the first time, set the initial `max-age` to a small value. Set the value from hours to no more than a single day in case you need to revert the HTTPS infrastructure to HTTP. After you're confident in the sustainability of the HTTPS configuration, increase the HSTS max-age value; a commonly used value is one year.

guardrex:
Thanks @shirhatti ... My hours are burned, so I'll take care of this first thing Tuesday morning.

guardrex:
@tunggnu ... do you want to modify this further, or do you want me to take care of the additional updates?

* We'll be able to avoid a new section.
* Let's update the existing example (`AspNetCore.Docs/aspnetcore/host-and-deploy/linux-nginx/nginx.conf`).
* Add a WARNING based on ...

   > We recommend using temporary redirects rather than permanent redirects. Link caching can cause unstable behavior in development environments.
  >
  > For production environments that are implementing HTTPS for the first time, set the initial max-age to a small value. Set the value from hours to no more than a single day in case you need to revert the HTTPS infrastructure to HTTP. After you're confident in the sustainability of the HTTPS configuration, increase the HSTS max-age value; a commonly used value is one year.

  ... but I'll need to do a touch of grammar work on that.

tunggnu:
Hi all,

The main purpose of this pull request is to properly redirect all HTTP requests to HTTPS because I thought the current configuration was unable to redirect HTTP requests. In fact, technically, it could. Why?

The tutorial said:

> Edit the */etc/nginx/nginx.conf* configuration file. The example contains both http and server sections in one configuration file.

The instruction above is confusing, especially for those who are familiar with Ubuntu's convention for Nginx. This convention includes:

* The `/etc/nginx/nginx.conf` file contains general settings. Also, the last part of it includes the settings from other files in other folders, which looks like this:

      ##
      # Virtual Host Configs
      ##

      include /etc/nginx/conf.d/*.conf;
      include /etc/nginx/sites-enabled/*;
* The `/etc/nginx/sites-available` folder contains configuration files corresponding to their websites.
* The `/etc/nginx/sites-enabled` folder contains symlinks to the actual configuration files in the `/etc/nginx/sites-available` folder.

That's why when readers only want to use HTTP, they only need to edit the `/etc/nginx/sites-available/default` file without touching the `/etc/nginx/nginx.conf` file, then Nginx can still know the new contents of the `/etc/nginx/sites-available/default` file.

When I read the quote above, I thought that I had to **copy** all the new contents of the example into my existing `/etc/nginx/nginx.conf` file, rather than **replace** all the old contents with the new ones. That means `include /etc/nginx/sites-enabled/*;` will be preserved. Thus Nginx will still read the contents of the `/etc/nginx/sites-available/default` file and I will have two HTTP server blocks colliding with each other.

But it seems that the quote above implies that the reader must replace all the old contents of the `/etc/nginx/nginx.conf` file with the example's contents. If so, Nginx will only read the contents of the `/etc/nginx/nginx.conf` file and ignore all other files, including the `/etc/nginx/sites-available/default` file. Consequently, the HTTP server block in the `/etc/nginx/nginx.conf` file is the only server block that handles HTTP requests, so it will catch all HTTP requests even if it is missing a `server_name`. If that is what the above quote really means, then it is necessary to add a note to the reader something like: "You need to **replace** the existing content of the `/etc/nginx/nginx.conf` file with the example's contents.". And you can decline this pull request.

Anyway, I see some other minor issues that can be improved:

* Inconsistency in the declaration of the `server_name` directive. In the `/etc/nginx/sites-available/default` file you declared `server_name example.com *.example.com;`, but in the `/etc/nginx/nginx.conf` file you declared `server_name example.com;`.
* The redundancy in the `listen` directive. You should replace `listen *:80;` with  `listen 80;`, and `listen *:443 ssl;` with `listen 443 ssl;`
* Does not adhere to the **separation of concerns** principle. You created the `/etc/nginx/sites-available/default` file to contain HTTP configuration, but included HTTPS configuration settings in the `/etc/nginx/nginx.conf` file, which should only contain the general settings. In my opinion, the ideal solution is to move all the specific settings out of the `/etc/nginx/nginx.conf` file. They can be placed inside a new file or merged into the `/etc/nginx/sites-available/default` file.
* About HSTS: you don't need to include `add_header Strict-Transport-Security max-age=15768000;` in the HTTP server block, because according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security), the Strict-Transport-Security header is ignored by the browser when your site is accessed using HTTP.
* About redirection status code: the first sentence of this tutorial is "This guide explains setting up a **production-ready** ASP.NET Core environment on an Ubuntu 16.04 server." So, the configuration should NOT focus on development environments. The redirection status code should be `301` and you can add a comment for development environments like this:
  > For development environments, we recommend using temporary redirects (302) rather than permanent redirects (301). Link caching can cause unstable behavior in development environments.

@guardrex I'm not a native English speaker, so please help me to take care of the additional updates.

guardrex:
Thanks for providing that feedback, @tunggnu.

@shirhatti ... I can work on this, but not at the moment. I'm 🏃😅 on 5.0 Blazor issues over the next few weeks. Do you want to provide feedback on the updates that @tunggnu suggested and have me work on this, or do you want to take another course of action?

guardrex:
I have some time this morning to try and address this. Because I can't edit more than just the one file modified by this PR here without submitting a PR to your branch, I'm just going to open a new PR. Also, I opened https://github.com/dotnet/AspNetCore.Docs/issues/20359 to track the work.

Thanks @tunggnu ... I'll ping you on the PR.

Issue #10827, opened 2019-02-06
Document reverse proxy usage with SignalR


BrennanConroy:
We should write a doc that has example reverse proxy configurations and common pitfalls that can occur.

One example is that SSE with nginx likely wont work by default because the request is buffered.
Nginx configuration can turn buffering off via `proxy_buffering off;` but you might only want to turn it off for SignalR specific requests via:
```
server {
    location /hubroute {
        proxy_buffering off;
    }
}
```
Another way to fix that issue is to send `X-Accel-Buffering: no` with your Http requests, so we could mention how to do that also.

WebSockets is another option that can sometimes need configuration in your proxies.

bradygaster:
I like the idea. You feel it should go into the "hosting and scaling" or "server concepts" TOC item? I think the former makes sense if we make a new article. 

Or, is it small enough that it'd make for a good section in an existing article?

BrennanConroy:
Hosting and scaling sounds right, however, if when we write it it gets too large, we can make a new article.

davidfowl:
Pretty sure there are more issues, like configuration for sticky-sessions as well

BrennanConroy:
Info dump

```nginx
http {
  map $http_upgrade $connection_upgrade {
    default Upgrade;
    '' close;
  }

  server {
    server_name backend1.example.com;
    listen 80 default_server;

    root /path/to/wwwroot;
    
    location /hubroute {
      proxy_pass http://localhost:5000;

      # Configure WebSockets
      proxy_set_header Upgrade $http_upgrade;
      proxy_set_header Connection $connection_upgrade;
      proxy_cache_bypass $http_upgrade;

      # Configure ServerSentEvents
      proxy_buffering off;

      # Configure LongPolling
      proxy_read_timeout 100s;

      proxy_set_header Host $host;
    }
  }
}
```

If you have multiple servers configured you will need to enable sticky sessions so clients wont get random 404's by hitting a different backend:
```nginx
http {
  upstream backend {
    server backend1.example.com;
    server backend2.example.com;
    sticky cookie srv_id expires=max domain=.example.com path=/ httponly;
  }
}
```

halter73:
> ```
>  map $http_upgrade $connection_upgrade {
>    default Upgrade;
>    '' close;
>  }
> ```

> ```
> proxy_set_header Connection $connection_upgrade;
> ```

[...]

PR #19983, opened 2020-09-23
Improve SignalR Nginx settings

BrennanConroy:
Progress on https://github.com/dotnet/AspNetCore.Docs/issues/10827

### DIFF
--- a/aspnetcore/signalr/scale.md
+++ b/aspnetcore/signalr/scale.md
@@ -108,10 +108,10 @@ The following contains the minimum required settings to enable WebSockets, Serve
 
 ```nginx
 http {
-  map $http_upgrade $connection_upgrade {
-    default Upgrade;
-    '' keep-alive;
-  }
+  map $http_connection $connection_upgrade {
+  "~*Upgrade" Upgrade;
+  default keep-alive;
+}
 
   server {
     listen 80;
--- a/aspnetcore/signalr/scale.md
+++ b/aspnetcore/signalr/scale.md
@@ -130,7 +130,7 @@ http {
       # Configuration for ServerSentEvents
       proxy_buffering off;
 
-      # Configuration for LongPolling
+      # Configuration for LongPolling or if your KeepAliveInterval is longer than 60 seconds
       proxy_read_timeout 100s;
 
       proxy_set_header Host $host;
--- a/aspnetcore/signalr/scale.md
+++ b/aspnetcore/signalr/scale.md
@@ -117,8 +117,6 @@ http {
     listen 80;
     server_name example.com *.example.com;
 
-    root /path/to/wwwroot;
-
     # Configure the SignalR Endpoint
     location /hubroute {
       # App server url
@@ -147,6 +145,8 @@ When multiple backend servers are used, sticky sessions must be added to prevent
 
 The following is added in addition to the previous configuration. `backend` in the following example is the name of the group of servers:
 
+With Nginx Open Source, use `ip_hash` to route connections to a server based on the clients IP address.
+
 ```nginx
 http {
   upstream backend {
@@ -155,11 +155,22 @@ http {
     # App server 2
     server http://localhost:5002;
 
-    # for Nginx Plus
-    sticky cookie srv_id expires=max domain=.example.com path=/ httponly;
-
-    # for Nginx Open Source
     ip_hash;
+  }
+}
+```
+
+With Nginx Plus, use `sticky` to add a cookie to requests and pin them to a server.
+
+```nginx
+http {
+  upstream backend {
+    # App server 1
+    server http://localhost:5000;
+    # App server 2
+    server http://localhost:5002;
+
+    sticky cookie srv_id expires=max domain=.example.com path=/ httponly;
   }
 }
 ```
@@ -167,6 +178,8 @@ http {
 Finally, change `proxy_pass http://localhost:5000` in the `server` section to `proxy_pass http://backend`.
 
 For more information on WebSockets over Nginx, see [NGINX as a WebSocket Proxy](https://www.nginx.com/blog/websocket-nginx).
+
+For more information on load balancing and sticky sessions, see [NGINX load balancing](https://docs.nginx.com/nginx/admin-guide/load-balancer/http-load-balancer/).
 
 ## Third-party SignalR backplane providers
 
--- a/aspnetcore/signalr/scale.md
+++ b/aspnetcore/signalr/scale.md
@@ -127,7 +127,7 @@ http {
       # Configuration for WebSockets
       proxy_set_header Upgrade $http_upgrade;
       proxy_set_header Connection $connection_upgrade;
-      proxy_cache_bypass $http_upgrade;
+      proxy_cache off;
 
       # Configuration for ServerSentEvents
       proxy_buffering off;
@@ -143,13 +143,12 @@ http {
 }
 ```
 
-When multiple backend servers are used sticky sessions must be added so SignalR connections do not switch servers when connecting.
-There are multiple ways to add sticky sessions in Nginx, we show two below depending on what you have available to you.
-The following is added in addition to the previous configuration:
+When multiple backend servers are used, sticky sessions must be added to prevent SignalR connections from switching servers when connecting. There are multiple ways to add sticky sessions in Nginx. Two approaches are shown below depending on what you have available.
+
+The following is added in addition to the previous configuration. `backend` in the following example is the name of the group of servers:
 
 ```nginx
 http {
-  # 'backend' is the name of our group of servers
   upstream backend {
     # App server 1
     server http://localhost:5000;
@@ -165,9 +164,9 @@ http {
 }
 ```
 
-Finally, change `proxy_pass http://localhost:5000` in the `server {` section to `proxy_pass http://backend`.
+Finally, change `proxy_pass http://localhost:5000` in the `server` section to `proxy_pass http://backend`.
 
-For more information on WebSockets over Nginx, see [NGINX as a WebSocket Proxy](https://www.nginx.com/blog/websocket-nginx/).
+For more information on WebSockets over Nginx, see [NGINX as a WebSocket Proxy](https://www.nginx.com/blog/websocket-nginx).
 
 ## Third-party SignalR backplane providers
 
--- a/aspnetcore/signalr/scale.md
+++ b/aspnetcore/signalr/scale.md
@@ -109,8 +109,8 @@ The following contains the minimum required settings to enable WebSockets, Serve
 ```nginx
 http {
   map $http_connection $connection_upgrade {
-  "~*Upgrade" Upgrade;
-  default keep-alive;
+    "~*Upgrade" $http_connection;
+    default keep-alive;
 }
 
   server {
--- a/aspnetcore/signalr/scale.md
+++ b/aspnetcore/signalr/scale.md
@@ -143,9 +143,9 @@ http {
 
 When multiple backend servers are used, sticky sessions must be added to prevent SignalR connections from switching servers when connecting. There are multiple ways to add sticky sessions in Nginx. Two approaches are shown below depending on what you have available.
 
-The following is added in addition to the previous configuration. `backend` in the following example is the name of the group of servers:
+The following is added in addition to the previous configuration. In the following examples, `backend` is the name of the group of servers.
 
-With Nginx Open Source, use `ip_hash` to route connections to a server based on the clients IP address.
+With [Nginx Open Source](https://nginx.org/en/), use `ip_hash` to route connections to a server based on the client's IP address:
 
 ```nginx
 http {
@@ -160,7 +160,7 @@ http {
 }
 ```
 
-With Nginx Plus, use `sticky` to add a cookie to requests and pin them to a server.
+With [Nginx Plus](https://www.nginx.com/products/nginx), use `sticky` to add a cookie to requests and pin the user's requests to a server:
 
 ```nginx
 http {
--- a/aspnetcore/signalr/scale.md
+++ b/aspnetcore/signalr/scale.md
@@ -110,7 +110,7 @@ The following contains the minimum required settings to enable WebSockets, Serve
 http {
   map $http_upgrade $connection_upgrade {
     default Upgrade;
-    '' close;
+    '' keep-alive;
   }
 
   server {
--- a/aspnetcore/signalr/scale.md
+++ b/aspnetcore/signalr/scale.md
@@ -181,6 +181,9 @@ For more information on WebSockets over Nginx, see [NGINX as a WebSocket Proxy](
 
 For more information on load balancing and sticky sessions, see [NGINX load balancing](https://docs.nginx.com/nginx/admin-guide/load-balancer/http-load-balancer/).
 
+For more information about ASP.NET Core with Nginx see the following article:
+* <xref:host-and-deploy/linux-nginx>
+
 ## Third-party SignalR backplane providers
 
 * [NCache](https://www.alachisoft.com/ncache/asp-net-core-signalr.html)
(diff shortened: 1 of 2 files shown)


aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,70 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
+The following contains the minimum required settings to enable WebSockets, ServerSentEvents, and LongPolling for SignalR:
 
 ```nginx
-proxy_set_header Upgrade $http_upgrade;
-proxy_set_header Connection $connection_upgrade;
+http {
+  map $http_upgrade $connection_upgrade {
+    default Upgrade;
+    '' close;
+  }
```
halter73, line 114: What's going on here?

BrennanConroy: Like I was saying on the issue, "Connection: close" is applied by default in Nginx, and "Connection: keep-alive" should not be forwarded by proxies as it is a hop-by-hop header. We can have another discussion about that behavior, but I think for this PR this is a big improvement over what we had before.

aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,70 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
+
+      # Configuration for WebSockets
+      proxy_set_header Upgrade $http_upgrade;
+      proxy_set_header Connection $connection_upgrade;
+      proxy_cache_bypass $http_upgrade;
+
+      # Configuration for ServerSentEvents
+      proxy_buffering off;
+
+      # Configuration for LongPolling
```
halter73, line 135: It looks like this is needed for WebSockets if the KeepAliveInterval is large enough http://nginx.org/en/docs/http/websocket.html

aspnetcore/signalr/scale.md
halter73: Now I remember. I have the same question [I had on the issue](https://github.com/dotnet/AspNetCore.Docs/issues/10827#issuecomment-631177400).

> Does this mean that every request to /hubroute will always contain either a "Connection: upgrade" or a "Connection: close" header? We wouldn't want to disable keep-alive for long polling.

Sure, "Connection" is a hop-by-hop header, but I want to make sure the connection between nginx and the signalr server isn't closed after every request.

aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,70 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
+
+    # Configure the SignalR Endpoint
+    location /hubroute {
+      # App server url
+      proxy_pass http://localhost:5000;
+
+      # Configuration for WebSockets
+      proxy_set_header Upgrade $http_upgrade;
+      proxy_set_header Connection $connection_upgrade;
+      proxy_cache_bypass $http_upgrade;
```
halter73, line 130: If we worried about the proxy caching requests, shouldn't we be worried about every request, not just WebSocket requests?

```suggestion
      proxy_cache off;
```

aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,70 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
+  map $http_upgrade $connection_upgrade {
+    default Upgrade;
+    '' close;
+  }
+
+  server {
+    listen 80;
+    server_name example.com *.example.com;
+
+    root /path/to/wwwroot;
```
halter73, line 120: Why bother with this in the signalr doc? What if wwwroot is on another machine? I expect static files will be proxied in many cases.

aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,70 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
+      proxy_set_header Host $host;
+      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+      proxy_set_header X-Forwarded-Proto $scheme;
+    }
+  }
+}
 ```
 
-For more information, see [NGINX as a WebSocket Proxy](https://www.nginx.com/blog/websocket-nginx/).
+When multiple backend servers are used sticky sessions must be added so SignalR connections do not switch servers when connecting.
```
guardrex, line 146: I think you meant to combine sentences into a paragraph ...

```suggestion
When multiple backend servers are used, sticky sessions must be added to prevent SignalR connections from switching servers when connecting. There are multiple ways to add sticky sessions in Nginx. Two approaches are shown below depending on what you have available.

```

aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,70 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
+There are multiple ways to add sticky sessions in Nginx, we show two below depending on what you have available to you.
```
guardrex, line 147:
```suggestion
```

aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,70 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
+    # for Nginx Plus
+    sticky cookie srv_id expires=max domain=.example.com path=/ httponly;
+
+    # for Nginx Open Source
+    ip_hash;
+  }
+}
+```
+
+Finally, change `proxy_pass http://localhost:5000` in the `server {` section to `proxy_pass http://backend`.
```
guardrex, line 168:
```suggestion
Finally, change `proxy_pass http://localhost:5000` in the `server` section to `proxy_pass http://backend`.
```

aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,70 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
+
+For more information on WebSockets over Nginx, see [NGINX as a WebSocket Proxy](https://www.nginx.com/blog/websocket-nginx/).
```
guardrex, line 170: Turns out for this link, they don't redirect, so we can go slash-less on it ...

```suggestion
For more information on WebSockets over Nginx, see [NGINX as a WebSocket Proxy](https://www.nginx.com/blog/websocket-nginx).
```

aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,70 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
+      proxy_set_header X-Forwarded-Proto $scheme;
+    }
+  }
+}
 ```
 
-For more information, see [NGINX as a WebSocket Proxy](https://www.nginx.com/blog/websocket-nginx/).
+When multiple backend servers are used sticky sessions must be added so SignalR connections do not switch servers when connecting.
+There are multiple ways to add sticky sessions in Nginx, we show two below depending on what you have available to you.
+The following is added in addition to the previous configuration:
```
guardrex, line 148: Recommend moving as much of the unlocalized language into the text as possible.

```suggestion
The following is added in addition to the previous configuration. `backend` in the following example is the name of the group of servers:
```

aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,70 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
+
+```nginx
+http {
+  # 'backend' is the name of our group of servers
```
guardrex, line 152:
```suggestion
```

guardrex (changes requested):
WRT "Nginx Plus" and "Nginx Open Source" ... If those aren't really meant to be used together and given our general desire to place text in the doc where it will MT, I recommend splitting into two examples ... one for each. The lead-in line(s) can then be in text and translate. If you want that but don't want to deal with the setup, I can edit directly for you to sign-off on. :ear:

aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,82 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
+      proxy_set_header X-Forwarded-Proto $scheme;
+    }
+  }
+}
 ```
 
-For more information, see [NGINX as a WebSocket Proxy](https://www.nginx.com/blog/websocket-nginx/).
+When multiple backend servers are used, sticky sessions must be added to prevent SignalR connections from switching servers when connecting. There are multiple ways to add sticky sessions in Nginx. Two approaches are shown below depending on what you have available.
+
+The following is added in addition to the previous configuration. `backend` in the following example is the name of the group of servers:
```
guardrex, line 146:
```suggestion
The following is added in addition to the previous configuration. In the following examples, `backend` is the name of the group of servers.
```

*"Ooops!*" on this one. The conditional is better up-front.

aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,82 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
+
+With Nginx Open Source, use `ip_hash` to route connections to a server based on the clients IP address.
```
guardrex, line 148:
```suggestion
With [Nginx Open Source](https://nginx.org/en/), use `ip_hash` to route connections to a server based on the client's IP address:
```

aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,82 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
+    server http://localhost:5000;
+    # App server 2
+    server http://localhost:5002;
+
+    ip_hash;
+  }
+}
+```
+
+With Nginx Plus, use `sticky` to add a cookie to requests and pin them to a server.
```
guardrex, line 163:
```suggestion
With [Nginx Plus](https://www.nginx.com/products/nginx), use `sticky` to add a cookie to requests and pin the user's requests to a server:
```

guardrex (approved):
Just some nit suggestions.

aspnetcore/signalr/scale.md
guardrex: btw - the "en" localization is required for their doc to load. **_Shame on them!_** :smile: lol

guardrex:
@BrennanConroy ... I made a few final nit suggestions :point_up:. Outside of those, was there anything else you wanted to do on this one before merging it?

BrennanConroy:
> Outside of those, was there anything else you wanted to do on this one before merging it?

Yeah, need to figure out what to do about keep-alive

guardrex:
Ok ... I'll leave it to you to merge when ready.

aspnetcore/signalr/scale.md
BrennanConroy: Is it safe to make all connections keep alive?

halter73: I don't see why not. Most modern HTTP proxies and clients do. It can be problem with stateful NTLM connections, but we're not telling people how to configure NTLM.

@Tratcher

Tratcher: Yes, enable keep-alive by default between nginx and kestrel.

aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,82 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
-Set the proxy's `Connection` and `Upgrade` headers to the following for SignalR WebSockets:
+The following contains the minimum required settings to enable WebSockets, ServerSentEvents, and LongPolling for SignalR:
 
 ```nginx
-proxy_set_header Upgrade $http_upgrade;
-proxy_set_header Connection $connection_upgrade;
+http {
+  map $http_upgrade $connection_upgrade {
+    default Upgrade;
+    '' close;
```
BrennanConroy, line 113:
```suggestion
    '' keep-alive;
```

Tratcher:
How different is this guidance from our general nginx guidance? Should we be improving that instead and linking to it from here with signalr specific notes?

BrennanConroy:
> How different is this guidance from our general nginx guidance? Should we be improving that instead and linking to it from here with signalr specific notes?

I prefer cross-referencing the docs.

aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,85 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
-Set the proxy's `Connection` and `Upgrade` headers to the following for SignalR WebSockets:
+The following contains the minimum required settings to enable WebSockets, ServerSentEvents, and LongPolling for SignalR:
 
 ```nginx
-proxy_set_header Upgrade $http_upgrade;
-proxy_set_header Connection $connection_upgrade;
+http {
+  map $http_upgrade $connection_upgrade {
+    default Upgrade;
+    '' keep-alive;
```
halter73, line 113: What's the reason for basing the Connection header on the incoming Upgrade header? Maybe the following which bases it on the incoming Connection header would make more sense. It seems more straightforward not to look at a different header.

```
map $http_connection $connection_upgrade {
  Upgrade Upgrade;
  default keep-alive;
}
```

BrennanConroy: Doesn't work if the request has a multi-value Connection header.

BrennanConroy: Looks like it might work if I do `"~*Upgrade" Upgrade;`

halter73: Given that, I'm fine with the way you have it now or the old way.

halter73 (approved):

aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,85 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
 
-Set the proxy's `Connection` and `Upgrade` headers to the following for SignalR WebSockets:
+The following contains the minimum required settings to enable WebSockets, ServerSentEvents, and LongPolling for SignalR:
 
 ```nginx
-proxy_set_header Upgrade $http_upgrade;
-proxy_set_header Connection $connection_upgrade;
+http {
+  map $http_connection $connection_upgrade {
+  "~*Upgrade" Upgrade;
```
halter73, line 112: Out of curiosity, does the following work?

```suggestion
  "~*Upgrade" $http_connection;
```

halter73:
We should wait for @Tratcher's approval before merging though.

aspnetcore/signalr/scale.md
Tratcher: What does `"~*Upgrade" Upgrade;` mean?

aspnetcore/signalr/scale.md
```diff
@@ -104,14 +104,85 @@ The preceding conditions make it likely to hit the 10 connection limit on a clie
...
+  default keep-alive;
```
Tratcher, line 113: Comparing to the other nginx doc, why do `default keep-alive;` here and `proxy_http_version 1.1; proxy_set_header   Connection keep-alive;` there?

Tratcher:
Don't block on me, but I would like to see consistent guidance across the docs for common settings like keep-alive.

aspnetcore/signalr/scale.md
BrennanConroy: Not sure what you mean, this is how you set the Connection header to Upgrade for websockets, and keep-alive for other connections. The other doc doesn't handle websockets.

aspnetcore/signalr/scale.md
BrennanConroy: Case-insensitive check for "Upgrade" being in the request header. For example, Firefox sets "keep-alive, Upgrade" which wouldn't match normally.

aspnetcore/signalr/scale.md
Tratcher: Can you not have a generic nginx config that works for normal requests and websockets?

Tratcher: If our general nginx guidance doesn't work for websockets we at least need to call that out more clearly there. There's a weird callout there for blazor server apps, but not for general websockets.

aspnetcore/signalr/scale.md
BrennanConroy: It does

aspnetcore/signalr/scale.md
BrennanConroy, line 113:
```suggestion
    default keep-alive;
```

Issue #15408, opened 2019-10-31
"Complete the Channel in the catch and outside the catch"?


Daniel15:
This article has a note:

> Wrap logic in a try ... catch. Complete the Channel in the catch and outside the catch to make sure the hub method invocation is completed properly.

This is worded pretty clumsily and doesn't match the code snippet underneath (which just completes the channel under the try-catch, not "in the catch"). 

I think what it's *trying* to say is that you should complete the channel either in a `finally` block or underneath a try-catch block, such that it's always called, even if an exception is thrown.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: a02b4e7d-971a-670b-02eb-89c5456ec8f2
* Version Independent ID: 5f2f089e-310d-83a1-9c56-9cb441b9914a
* Content: [Use streaming in ASP.NET Core SignalR](https://docs.microsoft.com/en-us/aspnet/core/signalr/streaming?view=aspnetcore-3.0)
* Content Source: [aspnetcore/signalr/streaming.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/signalr/streaming.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-signalr**
* GitHub Login: @bradygaster
* Microsoft Alias: **bradyg**

bradygaster:
@BrennanConroy feels like it'd benefit this doc for us to change the sample code to highlight the `IAsyncEnumerable` approach versus Channel. Yay/Nay?

BrennanConroy:
> it'd benefit this doc for us to change the sample code to highlight the `IAsyncEnumerable`

The bit above the Channels part shows `IAsyncEnumerable`, we just need to update the comment for the Channels section.

PR #19984, opened 2020-09-23
Fix streaming wording and example

BrennanConroy:
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/15408

### DIFF
--- a/aspnetcore/signalr/streaming.md
+++ b/aspnetcore/signalr/streaming.md
@@ -58,7 +58,7 @@ The following sample shows the basics of streaming data to the client using Chan
 > [!NOTE]
 > Write to the `ChannelWriter<T>` on a background thread and return the `ChannelReader` as soon as possible. Other hub invocations are blocked until a `ChannelReader` is returned.
 >
-> Wrap logic in a `try ... catch`. Complete the `Channel` in the `catch` and outside the `catch` to make sure the hub method invocation is completed properly.
+> Wrap logic in a `try ... catch`. Capture any exceptions inside the `catch` if you want to flow the error and complete the `Channel` in a `finally` to make sure the hub method invocation is completed properly.
 
 ::: moniker range=">= aspnetcore-3.0"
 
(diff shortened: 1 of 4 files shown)


halter73 (approved):

aspnetcore/signalr/streaming.md
```diff
@@ -58,7 +58,7 @@ The following sample shows the basics of streaming data to the client using Chan
 > [!NOTE]
 > Write to the `ChannelWriter<T>` on a background thread and return the `ChannelReader` as soon as possible. Other hub invocations are blocked until a `ChannelReader` is returned.
 >
-> Wrap logic in a `try ... catch`. Complete the `Channel` in the `catch` and outside the `catch` to make sure the hub method invocation is completed properly.
+> Wrap logic in a `try ... catch`. Capture any exceptions inside the `catch` if you want to flow the error and complete the `Channel` in a `finally` to make sure the hub method invocation is completed properly.
```
guardrex, line 61: Recommend singular "exception" and "error" all the way thru. It might be a bit challenging to grok with the condition in the middle. See if the following works ...

```suggestion
> Wrap logic in a [`try ... catch` statement](/dotnet/csharp/language-reference/keywords/try-catch). If you want to flow an error and complete the `Channel` in the hub method invocation properly, capture the exception inside the `catch` block and write it in a [`finally` block](/dotnet/csharp/language-reference/keywords/try-catch-finally).
```

guardrex (approved):
Just a recommendation on that sentence that might simplify its structure a bit.

aspnetcore/signalr/streaming.md
BrennanConroy: Tried my hand at the wording, the emphasis should be on the complete in the finally block.

guardrex: LGTM, except possibly the singular/plural errors/exceptions.

> If you want to flow any errors, capture them inside the `catch` block and write them in the finally block.

... or ...

> If you want to flow an error, capture it inside the `catch` block and write it in the finally block.

Something like either of those to either make all of the nouns singular or plural.

... and btw for a **_crazy Rex super nit_** :smile: ... pronouns cause a number of problems in technical writing, so I use them carefully ... I could even say that I use them 'rarely.' In this case, the subject is really clear, so I think they work well [error(s) ... "them" or "it" later in the same sentence].

guardrex (approved):

guardrex:
Merge when ready.

Issue #19982, opened 2020-09-23
Ideas for restructuring ANCM/IIS docs


jkotalik:
We were recently chatting in the ASP.NET Core team about how some of our docs for Servers/Host and Deploy are a bit long and have many topics that don't closely relate to each other. For example, the IIS overview doc takes 76 minutes to read and is just a collection of different topics that aren't strictly related to each other: https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/?view=aspnetcore-3.1. To a new user, seeing a 76 minute document is quite intimidating if they are just trying to get started with IIS or trying to find a specific issue.

My current thought is to break down these large grouping documents into more subsections. Here is a list of topics that I was thinking of breaking things into.

- Getting started
- Publishing
- Logging and diagnostics
- Troubleshooting lookup (similar to https://docs.microsoft.com/en-us/aspnet/core/test/troubleshoot-azure-iis?view=aspnetcore-3.1)
- Troubleshooting guide (More of a walkthrough guide like https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/azure-iis-errors-reference?view=aspnetcore-3.1)
- Hosting model - Inprocess
- Hosting model - Out of process
- Hosting bundle
- Application warmup
- app_offline.htm
- web.config
- Advanced
- Handler settings/configuration

Again, this wouldn't be changing the content, just the organization. 
@guardrex @Rick-Anderson @shirhatti Thoughts?

guardrex:
I agree. It started at a reasonable length; but years later, it has become a BeAsT 👹.

Do you propose to set the PR up, or do you want to provide the breakouts and have me (or someone else) set it up. I'll be free-er-ish in a couple of weeks right after Blazor WASM security docs are updated.

jkotalik:
I can do the basic layout and a bit of polish, I started specing it out locally to see what topics I wanted to highlight. I'll make a PR and we can decide how to party on afterwards.

guardrex:
Cool ... We won't need a redirect entry for this AFAIK because the overview will remain with whatever isn't stripped out.

jkotalik:
I was hoping to just target 5.0 with this change? That way 3.1 links will all still work.

guardrex:
It can be done at the risk of creating more maintenance overhead until the LTM for 3.x drops. Two copies mean two updates for at least some issues.

On the other hand, this content has been remarkably stable over the years. IIS doesn't change much, fortunately.

[...]

PR #19987, opened 2020-09-23
Restructure for IIS docs

jkotalik:
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/19982

None of the docs are finished yet, but contains the general structure I'm aiming for.


guardrex:
Dashes in the file names won't matter nearly as much as the content for SEO, but the recommendation is for dashes and these match our usual pattern ...

  * appinit.md :point_right: app-initialization.md (and trying to avoid an abbreviation)
  * appoffline.md :point-right: app-offline.md
  * dataprotection.md :point-right: data-protection.md
  * gettingstarted.md :point-right: get-started.md (and dropping the gerund "getting" in favor of "get")
  * hostingbundle.md :point_right: hosting-bundle.md
  * iisconfiguration.md :point_right: iis-configuration.md
  * iisspecific.md :point_right: "Specific" in this name isn't ... well ... *specific* :smile:. Everything in this node is supposed to be specific to IIS. Is there another file name that would be better for this content? "Advanced" (advanced.md) is already a concept, so are these "miscellaneous" (miscellaneous.md) IIS items?
  * inprocesshosting.md :point_right: in-process-hosting.md
  * logginganddiagnostics.md :point_right: logging-and-diagnostics.md
  * outofprocesshosting.md :point_right: out-of-process-hosting.md
  * webconfig.md :point_right: web-config.md

I ordered the original sections based on my usual general-to-specific and step-by-step writing guidelines. When you reach the stage of setting up the ToC for these later, you should be able to at least start with ordering of the topics on the basis of the order of the sections in the current topic then tweak it further from there.

Fortunately, there are only 24 cross-links to the *index.md* topic. Of those, only ten link to a specific section. I recommend checking all 24. Hopefully, only ten or fewer links will require an update.

jkotalik:
@guardrex thanks for the feedback. 

Question for you, I noticed that we have a doc called [Publishing To IIS](https://docs.microsoft.com/en-us/aspnet/core/tutorials/publish-to-iis?view=aspnetcore-3.1&tabs=visual-studio) which shares a lot of information in main doc about publishing in the [overview doc](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/?view=aspnetcore-3.1#create-the-iis-site). One of these is more of a brief overview while the other is more indepth. I think having a simple doc makes sense for getting started, however I don't want to lose some of the info here as well.

guardrex:
Yes, that tutorial came directly from reader feedback. Many weren't happy with the BOOK that we had here on the subject ... even getting lost in the forest of configuration. Therefore, we added this base case IIS tutorial to get up and running with a lot of defaults. I recommend keeping it. It gets very little negative feedback/questions.

jkotalik:
Do you recommend keeping both tutorials? My other thought was to remove the more verbose tutorial and have that in subsection for advance material (like web deploy details, process bitness, etc).

guardrex:
IIRC, the main doc that has the section was probably left in place because there were some differences ... I thought that there might be a tiny bit more detail than in the newer Pub to IIS tutorial, where I tried to say as little as possible for the base case.

If u think that section can go away leaving the tutorial, that seems fine. I suggest comparing carefully tho before chopping it out in case my hunch is right that it has a tiny bit more detail that should either be added to the Pub to IIS tutorial (less favored) or remain somewhere in other docs of this node (more favored) ... as u just suggested. Keeping that tutorial as light as possible is probably a good goal.

Rick-Anderson:
FYI  from the build report
image

guardrex:
@jkotalik ... 😕 ... Don't we want to leave the versioning as-is? 2.1 is supported until August 21, 2021.

jkotalik:
@guardrex a few things.
1. I think it's going to be best to leave all the new doc changes as exclusively 5.0, as I'd rather not break links to 2.1 and 3.1.
2. The only files that are 5.0+ are new files I just introduced. All other ones will remain the same.

guardrex:
Ok ... I thought you were going to pull the content from the overview. I guess not ... if that existing overview content (that's showing up in the new 5.0 topics) will just be versioned <5.0 in the overview topic, then I think 🤔 this will go well.

There would be a concern about maintaining multiple copies of the same content across docs. It's more 💰 *costly* 💰 to do it this way **_if docs change_**; however, these docs have been so stable over the last two years that I think this approach will be fine and **_not_** be a major 🕐 and 💲 drag. *The :pray: bean counter gods :pray: will ❤️ me for keeping such things in mind!* :smile: lol

guardrex:
btw -- wrt the breaking link point ... that's going to happen anyway. When this group of docs drops 3.1 (in the distant future), existing links to the prior content in the existing overview **_will_** need to be updated. I think it's about ten links, so it's no huge deal. If these changes were being done without the duplication, such a future change would be automatically covered.

jkotalik:
Making good progress: https://review.docs.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/aspnet-core-module?view=aspnetcore-5.0&branch=jkotalik%2FupdateTroubleshootingGuide

jkotalik:
I think this is ready to pass of to the doc team; content and structure are there now.

Few things that are left over:
- Figuring out if we should delete the get-started doc in the iis folder or not.
- Glue with referencing docs from other docs. Now that there are less topics in the same page, I think there could be some missed chances to references docs from other docs.

guardrex:
Cool ... I'll take a look first thing tomorrow morning.

jkotalik:
🥂

guardrex:
*ei ei ei* ... I'm at the DMV 🚗 ... *and you know how that goes!* 😖

Give me a few hours. I'll review as soon as I get back.

guardrex:
I'm back! ... *and I was only at the DMV for two hours.* 😩 Glad that's over with.

I'll make an edit pass directly, then I have a comment about the file locations to post to you.

guardrex (approved):
Ok ..... *whew* 😅 ... that was **_a lot_** of content to go thru. Most of my updates were for localization. We're code-fencing a lot more these days (e.g., paths, file and folder names, API, etc.) to avoid MT.

WRT file locations ... just letting you know as an FYI ... the ANCM topics were placed in the `host-and-deploy` folder because they apply to **both** IIS and Azure topic sets. Therefore, they  were placed up a level from those nodes. I don't think there are any changes for the PR. Just letting you know why that was the case.

WRT image links ... I can't see the internal build topics, so someone should :eye: them to make sure static image links are ok. I think they'll be fine.

You have a few 'make a bullet list here' remarks in there.

I'll approve now and probably take one more quick look tomorrow (Thursday) morning.

guardrex:
... and check the build report. There are build errors to address.

guardrex:
kk ... Performed another round of localization updates. There's even **_more_** that could be done for older versions, but let's not take those updates any further. The 2.x content will go away next year. I did run thru and try to get all of the XREF asterisks out in all versions in favor of `%2A`.

guardrex:
ok ... truly ....... :smile: ... there's **_a lot_** of content here. I'm sure nits will be missed. Except for the build report items, it should be ok&dagger;.

&dagger;*'It should be ok.'* &mdash; Put that on my tombstone plz! ⚰️

guardrex:
Looks like the ANCM topic and the old tutorial (and any others that were moved or deleted) will need redirection entries ...

https://github.com/dotnet/AspNetCore.Docs/blob/master/.openpublishing.redirection.json

... and existing redirect entries (if there are any) in the file can be updated to prevent multiple redirects from ancient links.

guardrex:
... and there are still build report errors. :boom:

Rick-Anderson:
Line 963: [Warning] Invalid file link: 'index/_static/ANCM-502_5.png'.

guardrex:
That was my bad if it was breaking. MT will do things like take "web.config" and try to translate the word "web" leaving it a broken file name. You did the right thing.

jkotalik:
@guardrex @Rick-Anderson I think this is in an OK state. As this is such a large change, what are the next steps for getting this merged?

guardrex:
I still don't see redirect file entries here, and I think two or more are needed.

jkotalik:
I don't know what those are or how to fix them 😅

guardrex:
Did one topic disappear and at least one more totally move (i.e., it's folder location changed, so it's URL also changed)?

We need redirect changes so that the old topics don't 404 but redirect to elsewhere (or the new topic URL) ...

https://github.com/dotnet/AspNetCore.Docs/blob/master/.openpublishing.redirection.json

guardrex:
oh ... one more thing on that ... there are probably earlier redirects already in the file that point to what are now the old locations of topics based on this PR. Therefore, it's not merely a matter of adding two (or more entries). The file has to be searched and updated for any earlier redirects that were sending traffic to the prior locations, too.

jkotalik:
I don't believe any files should have moved; I only added files and changed content in existing ones.

If you can point me to the ones that need redirecting, I can fix them. But not sure which ones 😃

guardrex:
oh ... ok ... false alarm then. IIRC, it was some earlier commit that I thought I saw something be deleted. The current PR seems good.

We can check the live topics as soon as this goes in to make sure that everything composed well. The build report seems ok.

@Rick-Anderson ... do u want to take another look? If not, I think it's good to go.

jkotalik:
Yeah, I'm going to continue to follow up with the restructure with *new* docs.

PR #20000, opened 2020-09-24
Added missing step causing compilation issue

cpoDesign:
Fixed guide to remove compilation issue caused by incorrect namespace in greet.proto.


cpoDesign:
JunTaoLuo @guardrex Is there anything else you need me to do? from my perspective, it is completed

guardrex:
We just need John to approve this and then we can get it merged. Stand-by for a bit longer ... the product unit 🐈 is *very busy* these days with the upcoming .NET 5 release, so there will be short delays for doc reviews during this period. 🏃🏃🏃🏃🏃😅

JunTaoLuo (approved):
Changes look good. We updated our server templates to include adjusting the namespace but we never updated the docs for creating the client.

PR #20016, opened 2020-09-26
Fix moniker range + nits

guardrex:
cc: @pranavkm ... It looks like you were going for **>=** 5.0 (to include 5.0) instead of > 5.0. If the change on that here is incorrect, I'm :ear: and will change it back to > 5.0.


guardrex:
@Rick-Anderson ... I went ahead and fixed this moniker range situation because I had a spelling error in a Blazor doc that I wanted to fix and go live with. The PR for the live branch gave me a 💀 *death threat* 💀 about the moniker range in this topic. I thought it best to just go ahead and not risk this topic breaking on the live site.

Rick-Anderson:
@guardrex  THanks!

PR #20032, opened 2020-09-28
[.NET 5 RC1] Add content on customizing client-side disconnect

captainsafia:
Fixes https://github.com/dotnet/aspnetcore/issues/26383.


guardrex:
I just noticed that both this PR's addition and the existing `<head>` modification feature should be 5.0+.

PR #20037, opened 2020-09-29
Fix Localization external link

metapone:


guardrex (approved):
Thanks @metapone! :rocket: ... our favorite type of cross-link is the "xref" style link. They're the most stable/safest (i.e., not magic string URLs) because the rendered cross-link is built by the build system. We're still working through all of the updates across the whole repo to change URL-style links to XREF-style links. It's an on-going multi-year project! ⛰️⛏️😅

Issue #20039, opened 2020-09-29
Missing link to source code


ChrisKlug:
The page contains a lot of references to sample code, and includes comments like "Code removed for brevity", but I can't find a link to the actual code. This makes it really hard to grasp the whole picture as the codes is very limited and shows only the exact things talked about.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ebdbdab0-c474-3fdd-76aa-ade905e76079
* Version Independent ID: 761ea2cc-7e05-4ae0-0875-44061a163621
* Content: [Multi-factor authentication in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/mfa?view=aspnetcore-3.1)
* Content Source: [aspnetcore/security/authentication/mfa.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/security/authentication/mfa.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-security**
* GitHub Login: @damienbod
* Microsoft Alias: **riande**

damienbod:
@ChrisKlug @Rick-Anderson 

I could add src code projects for this if you want. Most of the code was built using snippets from this repo:

https://github.com/damienbod/AspNetCoreHybridFlowWithApi

Greetings Damien

ChrisKlug:
I found that repo after Googling it, but the problem is that it is not linked anywhere in the page. This makes it really hard to find, and makes the code hard to understand to be honest.

guardrex:
The problem with moving the samples into the repo is keeping them up-to-date. Given that you, @damienbod, are the topic author and keep that repo up-to-date, I recommend cross-linking that repo as an additional resource for the topic. If you ever stopped keeping that repo up-to-date in the future, then I recommend that they come into the repo.

guardrex:
In addition to a title, you can also add some parenthetical info. For example something like ...

> \* [Authentication samples (damienbod/AspNetCoreHybridFlowWithApi GitHub repository)](https://github.com/damienbod/AspNetCoreHybridFlowWithApi): Code examples in this article are demonstrated in the sample apps.

damienbod:
I 'll do a PR then. @guardrex  What's the best way to link it? A note at the top and a link?

@ChrisKlug Would you find this good, would this help?

Greetings Damien

ChrisKlug:
[...]

PR #20043, opened 2020-09-29
 Add a link to the src at the top MFA doc

damienbod:
fixes https://github.com/dotnet/AspNetCore.Docs/issues/20039

### DIFF
--- a/aspnetcore/security/authentication/mfa.md
+++ b/aspnetcore/security/authentication/mfa.md
@@ -13,6 +13,8 @@ uid: security/authentication/mfa
 
 By [Damien Bowden](https://github.com/damienbod)
 
+[View or download sample code (damienbod/AspNetCoreHybridFlowWithApi GitHub repository)](https://github.com/damienbod/AspNetCoreHybridFlowWithApi)
+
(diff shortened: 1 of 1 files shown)


guardrex:
In the issue, we discussed *where* to place the link. Given that @damienbod is the topic author and keeps the repo up-to-date, I suggested that moving all of these samples into the repo might not be best. It creates *even more overhead* to maintain on this repo (and more topic overhead for @damienbod). I suggested moving the samples to the repo **_if/when_** @damienbod would ever decide in the future to stop keeping that repo up-to-date.

WRT to the link(s), the options are ...

* At the top only (calling out the external destination in the link text) ... This is what's on the PR. This is best for visibility, but it (probably) breaks the pattern of only having "official" samples cross-linked at the top.
* At the bottom only as the first link of *Additional resources*  (I suggested this approach ... bottom only), like ...

  > \* \[Authentication samples (damienbod/AspNetCoreHybridFlowWithApi GitHub repository)](https://github.com/damienbod/AspNetCoreHybridFlowWithApi): Code examples in this article are demonstrated in the sample apps.
* Both locations

Rick-Anderson (approved):

Issue #19286, opened 2020-07-22
Blazor project tracking 2021


guardrex:
This issue will ...

* Maintain a set of doc ideas for Blazor topics that we either don't want to (or can't) address in the near-term or are just ideas for consideration that don't warrant an issue at this time.
* Track work on topic UE (user experience) passes (i.e., topic overhauls).

### Doc ideas

* [ ] Confirm with the PU that runtime relinking is a **publish-and-Release config** gesture. Building only produced 51.5 MB. Publishing in `Debug` was 9.3 MB without relevant console compiler output. Publishing in `Release` was 6.6 with console compiler output that looked like runtime relinking was taking place.
* [ ] When https://github.com/dotnet/docs/issues/27395 is resolved and Mono/WASM MSBuild props are covered, update the cross-links from the targets file to the new official doc. Links can be located on the PR that added them at https://github.com/dotnet/AspNetCore.Docs/pull/24493.
* [ ] It seems like everyplace that we have hosted WASM, we should call out that the project must be run from the **`Server`** project. The errors for not doing so can be quite cryptic 😵. For example, the File Uploads topic example throws a "JSON token" error when the solution is run from the client app. It's not at all clear that it was simply run from the wrong project.
* [x] The pivots for the *Handle Errors* topic were dropped, then the content was reorganized on https://github.com/dotnet/AspNetCore.Docs/pull/24195 for its final layout. However, the *Prerendering* section is currently only for Blazor Server apps. That section should be made to generically cover both WASM and Server prerendering. https://github.com/dotnet/AspNetCore.Docs/pull/24211
* [x] For `NavLink` components in examples: It will harden docs against past and future deltas to merely mention adding example components to the `NavMenu` and not actually show the markup, which seems to change quite a bit. #24130
* [x] A number of spots show project file XML markup for adding packages to apps. There are also Package Manager and CLI commands. I think mentioning the package would be better and having an INCLUDE for the options: edit the project file, use the Package Manager, or use the CLI add package command. I should be able to filter the locations for these updates with `<PackageReference`. https://github.com/dotnet/AspNetCore.Docs/pull/24212
[...]

PR #20046, opened 2020-09-29
Blazor file uploads topic updates

### LINKED ISSUE #20035: Blazor InputFile

The InputFile component allows reading one or more files selected by a user to be upload. The `OnChange` event handler is invoked when a file selection is performed by the user. 

`FileChangeEventArgs.GetMultipleFiles` allows reading multiple files. Specify the maximum number of files you expect to read here., This prevents a malicious user from uploading a larger number of files than the app expects. `FileChangeEventArgs.File` allows reading the first and only file if the file upload does not support multiple files. 

`IBrowserFile` returns metadata [exposed by the browser](https://developer.mozilla.org/en-US/docs/Web/API/File#Instance_properties) as properties. This metadata can be useful to preliminary validation. It's important to understand and apply the the security consideration described in https://docs.microsoft.com/en-us/aspnet/core/mvc/models/file-uploads here when reading these files.

To begin reading a file, use `ReadAsync`. By default `ReadAsync` only allows reading a file smaller than 512kb in size. This limit is present to prevent developers from accidentally reading large files in to memory. Specify a reasonable approximation for the maximum expected file size if larger files need to be supported. Avoid reading the incoming file stream is not read to memory (copied to a `MemoryStream`, read as a byte array etc) as this can result in performance and security issues, especially in Blazor Server. Instead consider copying to an external store such as a blob or a file on disk.

In Blazor Server, file upload is performed by sending chunks of data over the SignalR connection. `<xref:Microsoft.AspNetCore.Components.Forms.RemoteBrowserFileStreamOptions />` allows configuring some of the implementation that drives the file upload.

Sample app demonstrating file upload + validation + progress bar: https://github.com/pranavkm/FileUpload/blob/master/Pages/Index.razor

---
#### Document Details

* ID: eef47457-c8cf-74a4-b884-c1fe5c2ef52d
* Version Independent ID: 09a02d10-2834-467d-d8c2-3fd19ceb3863
* Content: [ASP.NET Core Blazor forms and validation](https://docs.microsoft.com/en-us/aspnet/core/blazor/forms-validation?view=aspnetcore-5.0)
* Content Source: [aspnetcore/blazor/forms-validation.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/forms-validation.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #20035

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/file-uploads?view=aspnetcore-5.0&branch=pr-en-us-20046)

* Initial content of the [current live topic](https://docs.microsoft.com/aspnet/core/blazor/file-uploads?view=aspnetcore-5.0) is based on DR's blog post ... it turned out to be a good start.
* Recommend that we keep the bare bones example in the topic and include your example **_and the framework's example_** as cross-linked sample code (this is all on this PR). This gives the reader immediate access to examples of various approaches/features without loading the topic down with code.
* Check the code carefully tho. I set this up quickly without local testing 🏃 so that I could get on to other issues that need to be addressed. I may have 💥😢 something. I will check this again when I get to this topic on the UE passes on https://github.com/dotnet/AspNetCore.Docs/issues/19286.
* Cross-links are here between the file uploads topic and the file validation topic. I think they'll be visible either way, but see if you agree on that point.
* I placed file uploads into its own new topic and not into the forms validation topic in the first place because this component works outside of a form and it seems like there's enough 🍖 here to warrant its own topic. It's visible in the ToC immediately after the forms validation topic. I think this is best, but let's discuss and change it if not. :ear:

Capture

### DIFF
--- a/aspnetcore/blazor/file-uploads.md
+++ b/aspnetcore/blazor/file-uploads.md
@@ -5,21 +5,35 @@ description: Learn how to upload files in Blazor with the InputFile component.
 monikerRange: '>= aspnetcore-5.0'
 ms.author: riande
 ms.custom: mvc
-ms.date: 09/17/2020
+ms.date: 09/29/2020
 no-loc: ["ASP.NET Core Identity", cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
 uid: blazor/file-uploads
 ---
 # ASP.NET Core Blazor file uploads
 
-By [Daniel Roth](https://github.com/danroth27)
+By [Daniel Roth](https://github.com/danroth27) and [Pranav Krishnamoorthy](https://github.com/pranavkm)
 
-Use the `InputFile` component to read browser file data into .NET code, including for file uploads. The `InputFile` component renders as an HTML input of type `file`.
+[View or download sample code](https://github.com/dotnet/AspNetCore.Docs/tree/master/aspnetcore/blazor/file-uploads/samples/) ([how to download](xref:index#how-to-download-a-sample))
 
-By default, the user selects single files. Add the `multiple` attribute to permit the user to upload multiple files at once. When one or more files is selected by the user, the `InputFile` component fires an `OnChange` event and passes in an `InputFileChangeEventArgs` that provides access to the selected file list and details about each file.
+Use the `InputFile` component to read browser file data into .NET code, including for file uploads.
+
+> [!WARNING]
+> Always follow file upload security best practices. For more information, see <xref:mvc/models/file-uploads#security-considerations>.
+
+## `InputFile` component
+
+The `InputFile` component renders as an HTML input of type `file`.
+
+By default, the user selects single files. Add the `multiple` attribute to permit the user to upload multiple files at once. When one or more files is selected by the user, the `InputFile` component fires an `OnChange` event and passes in an `FileChangeEventArgs` that provides access to the selected file list and details about each file.
+
+To read data from a user-selected file:
+
+* Call `OpenReadStream` on the file and read from the returned stream. For more information, see the [File streams](#file-streams) section.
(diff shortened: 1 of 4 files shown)


aspnetcore/blazor/file-uploads.md
```diff
@@ -45,12 +59,12 @@ The following example demonstrates multiple image file upload in a component:
 @code {
     IList<string> imageDataUrls = new List<string>();
 
-    private async Task OnInputFileChange(InputFileChangeEventArgs e)
+    private async Task OnInputFileChange(FileChangeEventArgs e)
```
pranavkm, line 62:
```suggestion
    private async Task OnInputFileChange(InputFileChangeEventArgs e)
```

aspnetcore/blazor/file-uploads.md
```diff
@@ -64,4 +78,21 @@ The following example demonstrates multiple image file upload in a component:
...
+
+## File streams
+
+In a Blazor WebAssembly app, the data is streamed directly into the .NET code within the browser.
+
+In a Blazor Server app, the file data is streamed over the SignalR connection into .NET code on the server as the file is read from the stream. In the following table, `Forms.RemoteBrowserFileStreamOptions` allows configuring file upload characteristics for Blazor Server.
+
+| Blazor Server Property | Description | Default |
+| ---------------------- | ----------- | ------- |
+| `MaxBufferSize` | Maximum internal buffer size for unread data sent over a SignalR circuit. | 20,480 KB (20 KB) |
```
pranavkm, line 91: This looks nice. I thought this option type didn't warrant a whole lot of explanation because it's kinda difficult to figure out what tweaking some of these means to a user's code. It's hard to tell why one would change some of these unless users came up with very specific issues. The other bit is that adding the defaults would mean we would have to remember to update these if the code is updated. How do you feel about dropping that?

pranavkm (approved):
Looks great. Feel free to keep file upload in it's own section if you think that's best

aspnetcore/blazor/file-uploads.md
guardrex: Ok ... I'll drop it. They'll eventually have a cross-link here to the API doc, and I'll get that in there after GA.

guardrex:
Temporarily, I link to the engineering `master` branch for `RemoteBrowserFileStreamOptions` because we don't have an API doc yet. I'll re-point this after GA to the API doc.

PR #20059, opened 2020-09-30
Server Reset/Trailers docs

jkotalik:
Fixes https://github.com/dotnet/aspnetcore/issues/26255

Need some feedback for HTTP.SYS @Tratcher. Also probably need some glue to fit these two together better.


guardrex (approved):
We're trying to get all of the code comments out that we can into the text for localization.

Check me on that commit ...... I may have made a 💥:cry: on the updates.

Also ...... is this supposed to be versioned at all?

aspnetcore/fundamentals/servers/httpsys.md
```diff
@@ -256,6 +256,48 @@ In Visual Studio, the default launch profile is for IIS Express. To run the proj
...
+* TLS 1.2 or later connection
+
+### Trailers
+
+HTTP Trailers are similar to HTTP Headers, except they are sent after the response body is sent. For HTTP.sys, only HTTP/2 response trailers are supported.
+
+```csharp
+if (httpContext.Response.SupportsTrailers())
+{
+    httpContext.Response.DeclareTrailer("trailername");
```
JamesNK, line 275: This isn't necessary with HTTP/2.

aspnetcore/fundamentals/servers/httpsys.md
```diff
@@ -256,6 +256,48 @@ In Visual Studio, the default launch profile is for IIS Express. To run the proj
...
+
+In the preceding example code:
+
+* `SupportsTrailers` ensures that trailers are supported for the response.
+* `DeclareTrailer` adds the given trailer name to the `Trailer` response header. This must occur before the response headers are sent.
+* `AppendTrailer` appends the trailer.
+
+### Reset
+
+Reset allows for a consumer to reset a HTTP/2 request and abort it:
```
JamesNK, line 292:
```suggestion
Reset allows for the server to reset a HTTP/2 request with a specified error code. A reset request is considered aborted.
```

aspnetcore/fundamentals/servers/httpsys.md
```diff
@@ -256,6 +256,48 @@ In Visual Studio, the default launch profile is for IIS Express. To run the proj
...
+
+```csharp
+var resetFeature = httpContext.Features.Get<IHttpResetFeature>();
+resetFeature.Reset(errorCode: 1111);
```
JamesNK, line 296:
```suggestion
resetFeature.Reset(errorCode: 2);
```

aspnetcore/fundamentals/servers/httpsys.md
```diff
@@ -256,6 +256,48 @@ In Visual Studio, the default launch profile is for IIS Express. To run the proj
...
+```
+
+`Reset` in the preceding code example specifies the custom error code.
```
JamesNK, line 299:
```suggestion
`Reset` in the preceding code example specifies the `INTERNAL_ERROR` error code. For more information about HTTP/2 error codes, visit the [HTTP/2 specification error code section](https://tools.ietf.org/html/rfc7540#page-50).
```

aspnetcore/fundamentals/servers/httpsys.md
```diff
@@ -256,6 +256,48 @@ In Visual Studio, the default launch profile is for IIS Express. To run the proj
...
+    httpContext.Response.WriteAsync("Hello world");
+
+    httpContext.Response.AppendTrailer("trailername", "TrailerValue");
+}
+```
+
+In the preceding example code:
+
+* `SupportsTrailers` ensures that trailers are supported for the response.
+* `DeclareTrailer` adds the given trailer name to the `Trailer` response header. This must occur before the response headers are sent.
```
JamesNK, line 287: Should be removed. This is for trailers with chunked encoding, not HTTP/2.

aspnetcore/host-and-deploy/iis/iis-http2.md
```diff
@@ -0,0 +1,71 @@
...
+  * Public-facing edge server connections use HTTP/2, but the reverse proxy connection to the [Kestrel server](xref:fundamentals/servers/kestrel) uses HTTP/1.1.
+  * TLS 1.2 or later connection
+
+For an in-process deployment when an HTTP/2 connection is established, [`HttpRequest.Protocol`](xref:Microsoft.AspNetCore.Http.HttpRequest.Protocol*) reports `HTTP/2`. For an out-of-process deployment when an HTTP/2 connection is established, [`HttpRequest.Protocol`](xref:Microsoft.AspNetCore.Http.HttpRequest.Protocol*) reports `HTTP/1.1`.
+
+For more information on the in-process and out-of-process hosting models, see <xref:host-and-deploy/aspnet-core-module>.
+
+HTTP/2 is enabled by default. Connections fall back to HTTP/1.1 if an HTTP/2 connection isn't established. For more information on HTTP/2 configuration with IIS deployments, see [HTTP/2 on IIS](/iis/get-started/whats-new-in-iis-10/http2-on-iis).
+
+## Advanced HTTP/2 features to support gRPC
```
JamesNK, line 32: Mirror changes from other page here.

JamesNK:
@Tratcher @halter73 The sections on trailers and HTTP/2 reset is generic to all 3 servers. Is there an appropriate place to centralize it, and link from Kestrel, IIS and HttpSys docs?

aspnetcore/fundamentals/servers/httpsys.md
```diff
@@ -256,6 +256,45 @@ In Visual Studio, the default launch profile is for IIS Express. To run the proj
 
 For apps hosted by HTTP.sys that interact with requests from the Internet or a corporate network, additional configuration might be required when hosting behind proxy servers and load balancers. For more information, see [Configure ASP.NET Core to work with proxy servers and load balancers](xref:host-and-deploy/proxy-load-balancer).
 
+## Advanced HTTP/2 features to support gRPC
+
+Additional HTTP/2 features in HTTP.sys support gRPC, including support for response trailers and sending reset frames.
+
+Requirements to run gRPC with HTTP.SYS:
+
+* Windows 10 Version 2004, OS Build 20300.1000 or later
```
Tratcher, line 265: This has worked with HttpSys a lot longer than it's worked in IIS. I think it's 20H1 (2004), who's public build number is 19041.508.

aspnetcore/host-and-deploy/iis/iis-http2.md
```diff
@@ -0,0 +1,70 @@
...
+
+Additional HTTP/2 features in IIS support gRPC, including support for response trailers and sending reset frames.
+
+Requirements to run gRPC on IIS:
+
+* Must use in-process hosting.
+* Windows 10 Version 2004, OS Build 20300.1000 or later
```
Tratcher, line 39: 2004 is misleading here, that value doesn't get updated on insiders builds. Just list the OS Build number until there's an official release. Call out that they'll need to use Windows Insiders builds to get it.

jkotalik:
I don't think there is an appropriate place today. Should I file an issue to centralize common things between servers? Otherwise, I don't think the organization work until we do the refactor for both Kestrel and HttpSys.

Tratcher (commented):
> @Tratcher @halter73 The sections on trailers and HTTP/2 reset is generic to all 3 servers. Is there an appropriate place to centralize it, and link from Kestrel, IIS and HttpSys docs?

I don't think it even belongs in the servers section, these are just advanced HttpContext APIs. I don't think we have a concept doc section for HttpContext, but we should.

aspnetcore/fundamentals/servers/httpsys.md
jkotalik: Oops, was going to tag you in this for the build number.

aspnetcore/fundamentals/servers/httpsys.md
```diff
@@ -256,6 +256,45 @@ In Visual Studio, the default launch profile is for IIS Express. To run the proj
...
+    httpContext.Response.WriteAsync("Hello world");
+
+    httpContext.Response.AppendTrailer("trailername", "TrailerValue");
+}
+```
+
+In the preceding example code:
+
+* `SupportsTrailers` ensures that trailers are supported for the response.
+* `AppendTrailer` appends the trailer.
```
Tratcher, line 285: There's also DeclareTrailer to be called before starting the response body.

aspnetcore/host-and-deploy/iis/iis-http2.md
```diff
@@ -0,0 +1,70 @@
...
+* Out-of-process
+  * Windows Server 2016/Windows 10 or later; IIS 10 or later
+  * Public-facing edge server connections use HTTP/2, but the reverse proxy connection to the [Kestrel server](xref:fundamentals/servers/kestrel) uses HTTP/1.1.
+  * TLS 1.2 or later connection
+
+For an in-process deployment when an HTTP/2 connection is established, [`HttpRequest.Protocol`](xref:Microsoft.AspNetCore.Http.HttpRequest.Protocol*) reports `HTTP/2`. For an out-of-process deployment when an HTTP/2 connection is established, [`HttpRequest.Protocol`](xref:Microsoft.AspNetCore.Http.HttpRequest.Protocol*) reports `HTTP/1.1`.
+
+For more information on the in-process and out-of-process hosting models, see <xref:host-and-deploy/aspnet-core-module>.
+
+HTTP/2 is enabled by default. Connections fall back to HTTP/1.1 if an HTTP/2 connection isn't established. For more information on HTTP/2 configuration with IIS deployments, see [HTTP/2 on IIS](/iis/get-started/whats-new-in-iis-10/http2-on-iis).
```
Tratcher, line 30:
```suggestion
HTTP/2 is enabled by default for HTTPS/TLS connections. Connections fall back to HTTP/1.1 if an HTTP/2 connection isn't established. For more information on HTTP/2 configuration with IIS deployments, see [HTTP/2 on IIS](/iis/get-started/whats-new-in-iis-10/http2-on-iis).
```

JamesNK:
I've created an issue to make a HttpContext related fundamentals doc page - https://github.com/dotnet/aspnetcore/issues/26509

Leave the docs on trailers and reset where they are for this PR. They can be moved later when the new docs page is created.

JamesNK (approved):

aspnetcore/toc.yml
```diff
@@ -933,6 +933,9 @@
         - name: Transform web.config
           displayName: deploy, publish
           uid: host-and-deploy/iis/transform-webconfig
+        - name: HTTP/2
+          displayName: deploy, publish
+          uid: host-and-deploy/iis/iis-http2
```
JamesNK, line 938: Eventually IIS will support HTTP/3. Should this page have a more generic ID so that content could be added.

`host-and-deploy/iis/iis-protocols`?

aspnetcore/fundamentals/servers/httpsys.md
JamesNK: Not for HTTP/2 trailers.

aspnetcore/host-and-deploy/iis/iis-http2.md
```diff
@@ -0,0 +1,70 @@
...
+
+Additional HTTP/2 features in IIS support gRPC, including support for response trailers and sending reset frames.
+
+Requirements to run gRPC on IIS:
+
+* Must use in-process hosting.
+* Windows 10, OS Build 20300.1000 or later (may require being on Windows Insider Builds).
+* TLS 1.2 or later connection
+
+### Trailers
```
halter73, line 42: Until we resolve #26509, should we add this to kestrel.md?

@guardrex Is there a way to include content from one markdown file into another to reduce duplication?

JamesNK: Yes you can reference another file.

Example:
* https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/includes/gRPCazure.md
* https://github.com/dotnet/AspNetCore.Docs/blame/23bd8d205726bf80904800f6e98a98fee8ea0bfa/aspnetcore/grpc/index.md#L29

halter73: We should do that that for this. Even if it's temporary, it reduces the risk of the sections becoming out of sync.

guardrex: Place the content into an INCLUDE file and place it into an `iis` folder (node) `includes` folder.

`host-and-deploy/iis/includes/trailers.md` ... with the content.

Then ...

```
[!INCLUDE[](includes/trailers.md)]
```

... in the text of a topic in the `iis` folder.

aspnetcore/fundamentals/servers/httpsys.md
Tratcher: ? I thought it was still recommended to declare what trailers should be expected.

JamesNK: Not that I've seen. No gRPC implementation does that with trailers.

Is that in a spec?

jkotalik: Why is it recommend to declare trailers for HTTP/2?

aspnetcore/host-and-deploy/iis/iis-http2.md
jkotalik: What should I do about DeclareTrailers between these docs? Should I just go above and beyond and describe why it's needed for HTTP/1.1?

aspnetcore/fundamentals/servers/httpsys.md
Tratcher: Yes the HTTP/1.1 spec discusses Trailer with chunked because that was the only format that supported trailers at the time. https://tools.ietf.org/html/rfc7230#section-4.4 

>    When a message includes a message body encoded with the chunked
   transfer coding and the sender desires to send metadata in the form
   of trailer fields at the end of the message, the sender SHOULD
   generate a Trailer header field before the message body to indicate
   which fields will be present in the trailers.  This allows the
   recipient to prepare for receipt of that metadata before it starts
   processing the body, which is useful if the message is being streamed
   and the recipient wishes to confirm an integrity check on the fly.

Http/2 changed the wire format but it didn't change the semantics of most headers. I don't see anything in the HTTP/2 spec that negates the recommendation to set a Trailer header, and it's still used in the HTTP/2 samples.

```
     HTTP/1.1 200 OK                  HEADERS
     Content-Type: image/jpeg   ==>     - END_STREAM
     Transfer-Encoding: chunked         + END_HEADERS
     Trailer: Foo                         :status = 200
                                          content-length = 123
     123                                  content-type = image/jpeg
     {binary data}                        trailer = Foo
     0
     Foo: bar                         DATA
                                        - END_STREAM
                                      {binary data}

                                      HEADERS
                                        + END_STREAM
                                        + END_HEADERS
                                          foo = bar
```

jkotalik: Okay, bringing back declare trailers.

JamesNK: Ok.

Do you still want to mention `trailer` even though HTTP/2 clients don't use it? We should at least say that it is optional, and generally not required by HTTP/2 clients.

Tratcher: It was never required by HTTP/1.1 clients either, it was only a polite warning of things to come. I don't see how it's any different in HTTP/2.

aspnetcore/includes/trailers.md
```diff
@@ -0,0 +1,19 @@
...
+    httpContext.Response.WriteAsync("Hello world");
+
+    httpContext.Response.AppendTrailer("trailername", "TrailerValue");
+}
+```
+
+In the preceding example code:
+
+* `SupportsTrailers` ensures that trailers are supported for the response.
+* `DeclareTrailer` adds the given trailer name to the `Trailer` response header. This must occur before the response headers are sent.
```
JamesNK, line 18:
```suggestion
* `DeclareTrailer` adds the given trailer name to the `Trailer` response header. Declaring a response's trailers in the header is optional, but recommended. `DeclareTrailer` must be called before the response headers are sent.	
```

JamesNK (approved):

jkotalik:
@Rick-Anderson want to do an extra pass and do the honors to merge?

PR #20091, opened 2020-10-02
Add docs about using JavaScript libraries that render UI

SteveSandersonMS:
Fixes https://github.com/dotnet/aspnetcore/issues/22974


guardrex:
We need to check carefully ... I started editing before your 2nd and 3rd commits.

aspnetcore/blazor/call-javascript-from-dotnet.md
```diff
@@ -514,11 +514,13 @@ public async ValueTask<string> Prompt(string message)
...
 
-Fortunately it is actually quite simple to embed externally-generated UI within your Blazor component UI reliably. The recommended technique is to have your `.razor` code produce an empty element. As far as Blazor's diffing system is concerned, the element will always be empty, so it doesn't need to recurse into it and will simply leave its contents alone. This makes it safe to populate that element with arbitrary externally-managed content. Here is an example:
+Fortunately, it's straightforward to embed externally-generated UI within a Blazor component UI reliably. The recommended technique is to have the component's code (`.razor` file) produce an empty element. As far as Blazor's diffing system is concerned, the element is always empty, so the component isn't required to recurse into the element and  leave its contents alone. This makes it safe to populate the element with arbitrary externally-managed content.
```
SteveSandersonMS, line 521: One part of this edit looks erroneous:

```
so the component isn't required to recurse into the element and  leave its contents alone
```

This has the opposite meaning than intended. We do want it to leave the contents alone, but this says "it isn't required to leave its contents alone".

```suggestion
Fortunately, it's straightforward to embed externally-generated UI within a Blazor component UI reliably. The recommended technique is to have the component's code (`.razor` file) produce an empty element. As far as Blazor's diffing system is concerned, the element is always empty, so the renderer does not recurse into the element and instead leaves its contents alone. This makes it safe to populate the element with arbitrary externally-managed content.
```

aspnetcore/blazor/call-javascript-from-dotnet.md
```diff
@@ -579,42 +582,46 @@ As a more detailed example, consider the following component that renders an int
 
 The corresponding JavaScript module, which should be placed at `wwwroot/mapComponent.js`, is as follows:
 
-```js
+```javascript
 import 'https://api.mapbox.com/mapbox-gl-js/v1.12.0/mapbox-gl.js';
 
-// TO MAKE THE MAP APPEAR YOU MUST ADD YOUR ACCESS TOKEN FROM https://account.mapbox.com
-mapboxgl.accessToken = '<your access token here>';
+mapboxgl.accessToken = '{ACCESS TOKEN}';
```
SteveSandersonMS, line 588: I don't think you should remove the comment (`// TO MAKE THE MAP APPEAR...`). This is taken directly from https://docs.mapbox.com/mapbox-gl-js/api/.

The information about where to get an API token (https://account.mapbox.com) is essential for anyone who wants to make this code run correctly.

aspnetcore/blazor/call-javascript-from-dotnet.md
```diff
@@ -579,42 +582,46 @@ As a more detailed example, consider the following component that renders an int
...
 }
 
 export function setMapCenter(map, latitude, longitude) {
-    map.setCenter([longitude, latitude]);
+  map.setCenter([longitude, latitude]);
 }
 ```
 
-Finally, to produce correct styling, you should also add the following stylesheet tag to the host HTML page (`index.html` or `_Host.cshtml`):
+The placeholder `{ACCESS TOKEN}` in the preceding code is a valid access token from https://account.mapbox.com.
```
SteveSandersonMS, line 604: Not sure about this phrasing! It sounds to me like "You can use the string {ACCESS TOKEN} as an access token", but you can't. Could we emphasise something like "You need to replace the string {ACCESS TOKEN} with a valid access token that you can get from ..."

aspnetcore/blazor/call-javascript-from-dotnet.md
guardrex: It's below ... we're trying to get text out of code examples. See if the paragraph that follows covers it.

How about both? I'll add it back. I just want to make sure that we get a localized version of that guidance in the text.

I'll do it on the next commit.

SteveSandersonMS: Ah OK, thanks. Both sounds good!

guardrex: Done! :+1:

guardrex (approved):
Looks good ... ready to go?

SteveSandersonMS:
Wow @guardrex, you are *very* quick to response 😄

SteveSandersonMS:
Yes, ready as far as I'm concerned.

PR #20093, opened 2020-10-02
Fix sequence number in the component sample

meziantou:
I think sequence number should not be duplicated in the component


guardrex:
Thanks @meziantou ... The code originally came from the engineering repo. That's the way that they had it (before @captainsafia just removed it). That's why it was placed here ... they just took it out of their repo.

https://github.com/aspnet/samples/pull/63/files#diff-d385e94d1ae0c1570ac60bedd09037eaL37

pranavkm:
@guardrex how do we feel about pointing people at the current (5.0) implementation instead? 

> We don't have an implementation in 3.1, but you could build a component using the same techniques as <link to sources>.

It's much more robust compared to the implementation here.

guardrex:
Sure ... are you saying to drop it totally for 3.x? If so, we don't need to show the guts of the 5.x version of the component. The docs for the 5.x version are already here. We can just chop all of the 3.x coverage out and version this topic for 5.x.

... but we aren't going to drop the 3.x version, then we'll have to carry this code here. It's not available to cross-link from the product unit repo any longer.

pranavkm:
> Do you want to drop the 3.x coverage?

Let's do that. It would be less confusing if users accidentally end up in the 3.1 section or stick this in their app and do not update it when migrating to 5.0.

guardrex (approved):
Thanks @meziantou! :rocket:

Issue #19503, opened 2020-08-14
[RC1] Microsoft Identity Platform 2.0 for Blazor


guardrex:
**Update blazor templates with Microsoft.Identity.Web APIs**
https://github.com/dotnet/aspnetcore/pull/24268

**Remove accepted scope check for non-Identity.Web scenarios**
https://github.com/dotnet/aspnetcore/pull/24894

**Add Microsoft.Identity.Web migration guide to 3.1->5.0 documentation**
https://github.com/dotnet/AspNetCore.Docs/issues/19810

**Update Azure Active Directory documentation**
https://github.com/dotnet/AspNetCore.Docs/issues/19809

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: c572618c-8b98-95af-6a75-1f8e413071db
* Version Independent ID: 8670b314-460d-3194-c53b-9a44f39c6b2a
* Content: [Secure ASP.NET Core Blazor WebAssembly](https://docs.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/?view=aspnetcore-3.1)
* Content Source: [aspnetcore/blazor/security/webassembly/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/security/webassembly/index.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
@captainsafia ... I have a quick point of personal confusion here as I get into the updates. I decided to start with the small footprint coverage of Blazor Server docs. These will be quick to update, and I should have a PR up tomorrow morning.

According to [John's cross-link](https://github.com/dotnet/AspNetCore.Docs/issues/19810) to the AAD docs ...

**Quickstart: Add sign-in with Microsoft to an ASP.NET Core web app**
https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-v2-aspnet-core-webapp

[...]

PR #20095, opened 2020-10-02
Tweak Blazor authentication docs for .NET 5

captainsafia:
@guardrex Opening this as draft since I'm not totally sure what the strategy is for 3.1/5.0 docs. It looks like we might just be able to distinguish the different behavior with callouts.

### DIFF
--- a/aspnetcore/blazor/security/webassembly/hosted-with-azure-active-directory-b2c.md
+++ b/aspnetcore/blazor/security/webassembly/hosted-with-azure-active-directory-b2c.md
@@ -2,7 +2,7 @@
 title: Secure an ASP.NET Core Blazor WebAssembly hosted app with Azure Active Directory B2C
 author: guardrex
 description: Learn how to secure an ASP.NET Core Blazor WebAssembly hosted app with Azure Active Directory B2C.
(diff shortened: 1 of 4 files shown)


guardrex:
I'm tracking and working on this on https://github.com/dotnet/AspNetCore.Docs/issues/19503, and there's a lot more to update than just the platform config.

Was there an internal discussion that I didn't get wind of on the updates? My understanding was that I would perform all of the updates for all Blazor authn/z docs in one shot on one PR ... probably next week.

Also, we'll need to keep 3.x coverage unless DR decided to drop it. Did he decide to drop it? ... it would be the first that I've heard of it.

captainsafia:
Ah, no, I saw the issue but wasn't sure what the timeline for changes was and was reacting to some issues folks have opened in the aspnetcore repo as well as some stuff discovered during the CTI meeting. This PR just addresses the Azure AD stuff. I'll close this PR and stay tuned for your work.

guardrex:
Ok ... cool. I plan to hit that Monday morning now that the two other security PRs are merged. I hope it will only take a day or two. I'll ping u on the PR or on that issue if I run into problems.

PR #20119, opened 2020-10-06
Blazor Server auth topic updates for 5.0

guardrex:
Addresses #19503

Internal Review Topics ...

* [Blazor Server security *Overview*](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/security/server/index?view=aspnetcore-3.1&branch=pr-en-us-20119)
* [Blazor Server security *Additional Resources*](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/security/server/additional-scenarios?view=aspnetcore-3.1&branch=pr-en-us-20119)

Notes ...

* Whole topic versioning ... **_BUT_** ... let's try my fancy new hardcoded `<h2>` headings with `id`s to avoid the duplicate heading build warnings and preserve our bookmark anchors in the topic doing so in an HTML5-friendly way. It's probably the best of both worlds until we get *version-by-file*, when I'll revert these back to markdown H2s.
* Let's perform full topic UE passes while we're here, and let's keep an :eye: on my naming of *things*. Do my class and member names look good? :ear:
* For the `WeatherForecastService` [that Javier provided](https://github.com/javiercn/blazor-server-aad-sample/blob/master/BlazorServerAuthWithAzureActiveDirectory/Data/WeatherForecastService.cs#L17), can't I just make a field out of the `HttpClient` as shown on this PR? Like me, Javier isn't known for doing *random things*, so I'm concerned about making the change. I'll revert back to his property approach if I'm about to break :boom: something here. :ear:
* We should be able to just drop the last three sections for the 5.0 coverage. All of that stuff goes away as a concern AFAICT.
* I place those Azure Quickstart doc cross-links in the *Overview* topic. If you think of anything else that would be good to cross-link, I'm :ear: on review.

### DIFF
--- a/aspnetcore/blazor/security/server/additional-scenarios.md
+++ b/aspnetcore/blazor/security/server/additional-scenarios.md
@@ -38,16 +38,6 @@ services.Configure<OpenIdConnectOptions>(AzureADDefaults.OpenIdScheme, options =
 
 Optionally, additional scopes are added with `options.Scope.Add("{SCOPE}");`, where the placeholder `{SCOPE}` is the additional scope to add.
 
-Define a class to pass in the initial app state with the access and refresh tokens:
-
-```csharp
-public class InitialApplicationState
-{
-    public string AccessToken { get; set; }
-    public string RefreshToken { get; set; }
-}
-```
-
 Define a **scoped** token provider service that can be used within the Blazor app to resolve the tokens from [dependency injection (DI)](xref:blazor/fundamentals/dependency-injection):
 
 ```csharp
@@ -66,6 +56,16 @@ In `Startup.ConfigureServices`, add services for:
 ```csharp
 services.AddHttpClient();
 services.AddScoped<TokenProvider>();
+```
+
+Define a class to pass in the initial app state with the access and refresh tokens:
+
+```csharp
+public class InitialApplicationState
+{
+    public string AccessToken { get; set; }
+    public string RefreshToken { get; set; }
+}
 ```
 
 In the `_Host.cshtml` file, create and instance of `InitialApplicationState` and pass it as a parameter to the app:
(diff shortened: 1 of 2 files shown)


aspnetcore/blazor/security/server/additional-scenarios.md
```diff
@@ -157,7 +309,7 @@ endpoints.MapBlazorHub().RequireAuthorization(
 
 ## Use OpenID Connect (OIDC) v2.0 endpoints
 
-The authentication library and Blazor templates use OpenID Connect (OIDC) v1.0 endpoints. To use a v2.0 endpoint, configure the <xref:Microsoft.AspNetCore.Builder.OpenIdConnectOptions.Authority?displayProperty=nameWithType> option in the <xref:Microsoft.AspNetCore.Builder.OpenIdConnectOptions>:
+The authentication library and Blazor templates use OpenID Connect (OIDC) v1.0 endpoints in versions of ASP.NET Core prior to 5.0. To use a v2.0 endpoint with versions of ASP.NET Core prior to 5.0, configure the <xref:Microsoft.AspNetCore.Builder.OpenIdConnectOptions.Authority?displayProperty=nameWithType> option in the <xref:Microsoft.AspNetCore.Builder.OpenIdConnectOptions>:
```
captainsafia, line 312:
```suggestion
In versions of ASP.NET Core prior to 5.0, the authentication library and Blazor templates use OpenID Connect (OIDC) v1.0 endpoints. To use a v2.0 endpoint with versions of ASP.NET Core prior to 5.0, configure the <xref:Microsoft.AspNetCore.Builder.OpenIdConnectOptions.Authority?displayProperty=nameWithType> option in the <xref:Microsoft.AspNetCore.Builder.OpenIdConnectOptions>:
```

aspnetcore/blazor/security/server/additional-scenarios.md
```diff
@@ -5,36 +5,189 @@ description: Learn how to configure Blazor Server for additional security scenar
...
+    options.ResponseType = OpenIdConnectResponseType.Code;
     options.SaveTokens = true;
 
     options.Scope.Add("offline_access");
-    options.Scope.Add("{SCOPE}");
-    options.Resource = "{RESOURCE}";
 });
 ```
 
+Optionally, additional scopes are added with `options.Scope.Add("{SCOPE}");`, where the placeholder `{SCOPE}` is the additional scope to add.
```
captainsafia, line 39: Do we need a note about `Resource` here?

aspnetcore/blazor/security/server/additional-scenarios.md
```diff
@@ -5,36 +5,189 @@ description: Learn how to configure Blazor Server for additional security scenar
...
+
+Define a class to pass in the initial app state with the access and refresh tokens:
```
captainsafia, line 41: I would move this lower to where we talk about setting up the initial params in the _Host.cshtml file.

captainsafia (commented):
Took a look at this. @javiercn or @HaoK might have more feedback on the auth-specific stuff.

aspnetcore/blazor/security/server/additional-scenarios.md
guardrex: We don't ... it's an Identity v1.0 concept. @javiercn's guidance was to remove it when using v2.0 endpoints.

HaoK (approved):

Issue #20063, opened 2020-10-01
need a new page: "Troubleshooting connection issues"


Rick-Anderson:
per @BrennanConroy  in #20056

I think we need a new page: "Troubleshooting connection issues" and contain more general information about possible errors.
These errors aren't necessarily specific to the javascript client.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 3cf283c9-5cd2-a9ff-b8c8-372d8a04356d
* Version Independent ID: 729fddcf-24e4-b563-4952-a8045b231868
* Content: [ASP.NET Core SignalR JavaScript client](https://docs.microsoft.com/en-us/aspnet/core/signalr/javascript-client?view=aspnetcore-3.1)
* Content Source: [aspnetcore/signalr/javascript-client.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/signalr/javascript-client.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-signalr**
* GitHub Login: @bradygaster
* Microsoft Alias: **bradyg**

BrennanConroy:
Done via https://github.com/dotnet/AspNetCore.Docs/pull/20125

PR #20125, opened 2020-10-06
Add troubleshooting page for SignalR

BrennanConroy:
https://github.com/dotnet/AspNetCore.Docs/issues/20063

Needs some cleanup.

This is definitely missing some common scenarios. Don't need to get them all in this first pass.

@Rick-Anderson  edit
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/signalr/troubleshoot?view=aspnetcore-3.1&branch=pr-en-us-20125)

### DIFF
--- a/aspnetcore/signalr/javascript-client.md
+++ b/aspnetcore/signalr/javascript-client.md
@@ -262,46 +262,6 @@ The following code demonstrates a typical manual reconnection approach:
 
 A real-world implementation would use an exponential back-off or retry a specified number of times before giving up.
 
-## Troubleshoot WebSocket handshake errors
-
-This section provides help on the *"Error during WebSocket handshake"* exception that occurs when trying to establish connection to ASP.NET Core SignalR hub.
-
-### Response code 400 or 503
-
-For the following error:
-
-```log
-WebSocket connection to 'wss://xxx/HubName' failed: Error during WebSocket handshake: Unexpected response code: 400
-
-Error: Failed to start the connection: Error: There was an error with the transport.
-```
-
-The error is usually caused by client only use WebSockets transport but WebSockets protocol is not enabled on server.
-
-### Response code 307
-
-```log
-WebSocket connection to 'ws://xxx/HubName' failed: Error during WebSocket handshake: Unexpected response code: 307
-```
-
-Frequently this occurs when the SignalR hub server:
-
-* Listens to and responds over both HTTP and HTTPS.
-* Is configured to enforce HTTPS by calling `UseHttpsRedirection` in `Startup`, or enforces HTTPS via URL rewrite rule.
-
-This error can be caused by specifying the HTTP URL on client side using `.withUrl("http://xxx/HubName")`. The fix for this case is modifying the code to use an HTTPS URL.
-
-### Response code 404
-
-```log
-WebSocket connection to 'wss://xxx/HubName' failed: Error during WebSocket handshake: Unexpected response code: 404
-```
-
-If the app works on localhost, but returns this error after publishing to IIS server:
-
-* Verify the ASP.NET Core SignalR app is hosted as an IIS sub-application.
-* Don't set URL with the sub-app's pathbase on SignalR JavaScript client side `.withUrl("/SubAppName/HubName")`.
-
 ## Additional resources
 
 * [JavaScript API reference](/javascript/api/?view=signalr-js-latest&preserve-view=true )
@@ -312,6 +272,7 @@ If the app works on localhost, but returns this error after publishing to IIS se
 * [Publish to Azure](xref:signalr/publish-to-azure-web-app)
 * [Cross-Origin Requests (CORS)](xref:security/cors)
 * [Azure SignalR Service serverless documentation](/azure/azure-signalr/signalr-concept-serverless-development-config)
+* [Troubleshoot connection errors](xref:signalr/troubleshoot)
 
 ::: moniker-end
 
--- a/aspnetcore/signalr/troubleshoot.md
+++ b/aspnetcore/signalr/troubleshoot.md
@@ -1,0 +1,88 @@
+---
+title: ASP.NET Core SignalR connection troubleshooting
+author: 
+description: ASP.NET Core SignalR connection troubleshooting.
+monikerRange: '>= aspnetcore-2.1'
+ms.author: 
+ms.custom: mvc
+ms.date: 04/08/2020
+no-loc: ["ASP.NET Core Identity", cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
+uid: signalr/troubleshoot
+---
+## Troubleshoot connection errors
+
+This section provides help with errors that can occur when trying to establish a connection to a ASP.NET Core SignalR hub.
+
+### Response code 404
+
+When using WebSockets and `skipNegotiation = true`
+```log
+WebSocket connection to 'wss://xxx/HubName' failed: Error during WebSocket handshake: Unexpected response code: 404
+```
+
+* If the connection takes too long to send a request to the server after the negotiate the server will delete the ID and return a 404 if the connection uses that ID
+* If you have multiple servers and aren't using sticky sessions, the connection can start on one server and then switch to another server that doesn't know about it
+* Double check that the client is connecting to the correct endpoint. e.g. Server is hosted at http://127.0.0.1:5000/hub/myHub and client is trying to connect to http://127.0.0.1:5000/myHub.
+
+### Response code 400 or 503
+
+For the following error:
+
+```log
+WebSocket connection to 'wss://xxx/HubName' failed: Error during WebSocket handshake: Unexpected response code: 400
+
+Error: Failed to start the connection: Error: There was an error with the transport.
+```
+
+The error is usually caused by client only use WebSockets transport but WebSockets protocol is not enabled on server.
+
+### Response code 307
+
+When using WebSockets and `skipNegotiation = true`
+```log
+WebSocket connection to 'ws://xxx/HubName' failed: Error during WebSocket handshake: Unexpected response code: 307
+```
+
+This can also happen during the negotiate request.
+
+Common cause:
+* App is configured to enforce HTTPS by calling `UseHttpsRedirection` in `Startup`, or enforces HTTPS via URL rewrite rule.
+
+Possible solution:
+* Change the URL on the client side from "http" to "https". `.withUrl("https://xxx/HubName")`
+
+### Response code 405
+
+Http status code 405 - Method Not Allowed
+
+* The app doesn't have [CORS](xref:signalr/security#cross-origin-resource-sharing) enabled
+
+### Response code 0
+
+Http status code 0 - Usually a [CORS](xref:signalr/security#cross-origin-resource-sharing) issue, no status code is given
+
+```log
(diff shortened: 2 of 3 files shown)


aspnetcore/signalr/javascript-client.md
```diff
@@ -262,46 +262,6 @@ The following code demonstrates a typical manual reconnection approach:
 
 A real-world implementation would use an exponential back-off or retry a specified number of times before giving up.
 
-## Troubleshoot WebSocket handshake errors
-
-This section provides help on the *"Error during WebSocket handshake"* exception that occurs when trying to establish connection to ASP.NET Core SignalR hub.
```
bradygaster, line 267: Do we need quotes AND stars? One or the other seems like enough.

aspnetcore/signalr/troubleshoot.md
```diff
@@ -0,0 +1,88 @@
...
+This section provides help with errors that can occur when trying to establish a connection to a ASP.NET Core SignalR hub.
+
+### Response code 404
+
+When using WebSockets and `skipNegotiation = true`
+```log
+WebSocket connection to 'wss://xxx/HubName' failed: Error during WebSocket handshake: Unexpected response code: 404
+```
+
+* If the connection takes too long to send a request to the server after the negotiate the server will delete the ID and return a 404 if the connection uses that ID
```
bradygaster, line 23: 👍

bradygaster (approved):

aspnetcore/signalr/troubleshoot.md
Rick-Anderson, line 23: This is difficult for native speakers to read and impossible to MT (machine translate)

```suggestion
* If the connection takes too long to send a request to the server after the negotiate, the server:

  * Deletes the ID.
  * Returns a 404 if the connection uses that ID.
```

Rick-Anderson: Does it only return 404 if the connection uses that ID? Maybe the following is better:
```suggestion
* If the connection uses the ID and takes too long to send a request to the server after the negotiate, the server:

  * Deletes the ID.
  * Returns a 404.
```

aspnetcore/signalr/troubleshoot.md
```diff
@@ -0,0 +1,88 @@
...
+* If you have multiple servers and aren't using sticky sessions, the connection can start on one server and then switch to another server that doesn't know about it
```
Rick-Anderson, line 24:
```suggestion
* When using multiple servers without sticky sessions, the connection can start on one server and then switch to another server. The other server is not aware of the previous connection.
```

aspnetcore/signalr/troubleshoot.md
```diff
@@ -0,0 +1,88 @@
...
+* Double check that the client is connecting to the correct endpoint. e.g. Server is hosted at http://127.0.0.1:5000/hub/myHub and client is trying to connect to http://127.0.0.1:5000/myHub.
```
Rick-Anderson, line 25:
```suggestion
* Verify the client is connecting to the correct endpoint. For example, the server is hosted at `http://127.0.0.1:5000/hub/myHub` and client is trying to connect to `http://127.0.0.1:5000/myHub`.
```

aspnetcore/signalr/troubleshoot.md
```diff
@@ -0,0 +1,88 @@
...
+
+For the following error:
+
+```log
+WebSocket connection to 'wss://xxx/HubName' failed: Error during WebSocket handshake: Unexpected response code: 400
+
+Error: Failed to start the connection: Error: There was an error with the transport.
+```
+
+The error is usually caused by client only use WebSockets transport but WebSockets protocol is not enabled on server.
```
Rick-Anderson, line 37:
```suggestion
This error is usually caused by client using only WebSockets transport but the WebSockets protocol is not enabled on the server.
```

aspnetcore/signalr/troubleshoot.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Response code 307
+
+When using WebSockets and `skipNegotiation = true`
+```log
+WebSocket connection to 'ws://xxx/HubName' failed: Error during WebSocket handshake: Unexpected response code: 307
+```
+
+This can also happen during the negotiate request.
```
Rick-Anderson, line 46:
```suggestion
This error can also happen during the negotiate request.
```

Rick-Anderson (changes requested):
@FredHan001 FYI, this should hlep.

Rick-Anderson:
@FredHan001   FYI

Rick-Anderson:
@BrennanConroy  let me know what's need to get this merged. Great info, thanks.

BrennanConroy:
I guess it needs author to be filled in to get rid of the warnings

Rick-Anderson:
image

aspnetcore/signalr/javascript-client.md
```diff
@@ -262,46 +262,6 @@ The following code demonstrates a typical manual reconnection approach:
 
 A real-world implementation would use an exponential back-off or retry a specified number of times before giving up.
 
-## Troubleshoot WebSocket handshake errors
```
halter73, line 265: Is this going to break incoming links? If so, we should leave it here if `< aspnetcore-5.0'.

aspnetcore/signalr/troubleshoot.md
```diff
@@ -0,0 +1,91 @@
+---
+title: ASP.NET Core SignalR connection troubleshooting
+author: 
+description: ASP.NET Core SignalR connection troubleshooting.
+monikerRange: '>= aspnetcore-2.1'
+ms.author:
```
halter73, line 6: @Rick-Anderson I see your screenshot with the suggested pending changes to add @bradygaster as an author to this doc. Did you mean to submit a new review with those suggestion?

aspnetcore/signalr/javascript-client.md
BrennanConroy: This was added a couple days ago, so no

aspnetcore/signalr/troubleshoot.md
bradygaster: same question.

aspnetcore/signalr/troubleshoot.md
```diff
@@ -0,0 +1,91 @@
...
+
+When using WebSockets and `skipNegotiation = true`
+```log
+WebSocket connection to 'wss://xxx/HubName' failed: Error during WebSocket handshake: Unexpected response code: 404
+```
+
+* If the connection uses the ID and takes too long to send a request to the server after the negotiate, the server:
+
+  * Deletes the ID.
+  * Returns a 404.
```
halter73, line 26: This is probably the least common cause of a 404, don't you think? I would probably order the bullet points with bad path first, no sticky session second and this last.

aspnetcore/signalr/troubleshoot.md
```diff
@@ -0,0 +1,91 @@
...
+This error is usually caused by client using only WebSockets transport but the WebSockets protocol is not enabled on the server.
+
+### Response code 307
+
+When using WebSockets and `skipNegotiation = true`
+```log
+WebSocket connection to 'ws://xxx/HubName' failed: Error during WebSocket handshake: Unexpected response code: 307
+```
+
+This error can also happen during the negotiate request.
```
halter73, line 49: Do any of the HTTP clients and/or WebSocket clients SignalR uses not follow 307 redirects? I just searched the aspnetcore repo for "SignalR 307" (without quotes) and came up with no reports of an issue like that.

halter73: Found it: https://stackoverflow.com/questions/63437471/websocket-connection-to-ws-localhost5000-notificationhub-failed-error-duri. I'm still not sure that this could "happen during the negotiate request." Do you have evidence of that?

BrennanConroy: Yes, I point out one possible cause below

halter73: I don't see it. Why would a 307 break the negotiate request itself?

BrennanConroy: Because we don't follow redirects

halter73: Oh. Why? Because we assume it will break WebSockets?

Rick-Anderson:
> @Rick-Anderson I see your screenshot with the suggested pending changes to add @bradygaster as an author to this doc. Did you mean to submit a new review with those suggestion?

No, I made a bunch of suggestions, most were accepted. We can't merge this without the author meta-data. These are my suggestions, if you agree, I'll add them.

aspnetcore/signalr/troubleshoot.md
BrennanConroy: Hmm, I'm not sure. I thought it was because auth handlers often redirect instead of returning 401, but we set a header to try and avoid that...

Rick-Anderson:
The following needs to be added:

Q: Customer would like to know on the server the reason of the disconnection. However the exception parameter of the `OnDisconnectedAsync `seems to be always null:
        `public override async Task OnDisconnectedAsync(Exception exception)`

A: I suspect you're right that there was some sort of network issue. If the connection closes between SignalR messages, the server may interpret that as an unexceptional client disconnect. I filed https://github.com/dotnet/aspnetcore/issues/26701 to track the issue of not passing an Exception to Hub.OnDisconnectedAsync in this case.

BrennanConroy:
> The following needs to be added:
> 
> Q: Customer would like to know on the server the reason of the disconnection. However the exception parameter of the `OnDisconnectedAsync `seems to be always null:
> `public override async Task OnDisconnectedAsync(Exception exception)`
> 
> A: I suspect you're right that there was some sort of network issue. If the connection closes between SignalR messages, the server may interpret that as an unexceptional client disconnect. I filed [dotnet/aspnetcore#26701](https://github.com/dotnet/aspnetcore/issues/26701) to track the issue of not passing an Exception to Hub.OnDisconnectedAsync in this case.

I don't know what kind of wording we want in the doc for something like this.

aspnetcore/signalr/troubleshoot.md
BrennanConroy: https://github.com/dotnet/aspnetcore/issues/27068

bradygaster:
Whilst not very friendly, I think the statement

> If the connection closes between SignalR messages, the server may interpret that as an unexceptional client disconnect.

Sums it up concisely. That said, i'd love some wordsmithery by @Rick-Anderson

Rick-Anderson:
> If the connection closes between SignalR messages, the server may interpret that as an unexceptional client disconnect.

That's clear to me.

BrennanConroy:
> > If the connection closes between SignalR messages, the server may interpret that as an unexceptional client disconnect.
> 
> That's clear to me.

Sure it's clear, but it doesn't mean anything. If we rephrase it a little it's equivalent to: "If the client closes the connection, it could be a graceful or ungraceful disconnect."

Rick-Anderson:
@BrennanConroy  I see your point. How about:

Transient network failures may close the connection between SignalR messages. The server may interpret the closed connection as an unexceptional client disconnect.

Rick-Anderson:
@BrennanConroy  added 
For more information, see [this GitHub issue](https://github.com/dotnet/aspnetcore/issues/26701).

What I have is pretty incomplete, but the issue fleshes it out and we might get feedback on the issue.

Rick-Anderson (approved):

Rick-Anderson:
@BrennanConroy  let me know when I can merge this.

BrennanConroy:
Lets wait till middle of the day tomorrow to give folks some time to take a look

bradygaster:
LGTM

BrennanConroy:
Ok, think it's good to go

PR #20147, opened 2020-10-08
Blazor WASM security updates for 5.0

guardrex:
Addresses #19503

Internal Review Topics ...

* [Hosted AAD](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/hosted-with-azure-active-directory?view=aspnetcore-3.1&branch=pr-en-us-20147)
* [Hosted AAD B2C](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/hosted-with-azure-active-directory-b2c?view=aspnetcore-3.1&branch=pr-en-us-20147)

Alright ... so ... the good news is that everything is working here :+1:. The bad news is that ...

### Scopes ... scopes ... and SCOPES!

The scopes scenario between `api://`-based and `https://{TENANT}.onmicrosoft.com/`-based App Id URIs is resolved now with Jean-Marc's assistance. The difference is based on the tenant's domain: If it's verified, you get full control and `api://`-based. I'm updating the PR to react to this new understanding.

### WRT additional scopes (openid, offline_access, User.Read)

I'm dropping the **_hosted_** INCLUDES from yesterday for these hosted scenarios. Testing indicates that the app doesn't :boom: and the access token doesn't :boom: if the dev just goes with the solution config guidance and ends up with a client app that only specifies the API scope as the default access token scope. There's no need or reason to do anything with `openid`, `offline_access`, or `User.Read` for the access token scope AFAICT. Correct me if I'm wrong about that.

### B2C broken config from CLI command

Another little side note on a broken scenario with the hosted B2C scenario ...

When creating the solution from the template with the CLI command, one still has to keep an :eye: out for a duplicated scheme and host in the App ID URI. Even with the 5.x template, the command still yielded ...

```csharp
options.ProviderOptions.DefaultAccessTokenScopes
    .Add("https://XXXXX.onmicrosoft.com/https://XXXXX.onmicrosoft.com/eaee236e-faf8-442e-b472-7ff311d5ae76/API.Access");
```

... so it had to be manually changed to ...

```csharp
options.ProviderOptions.DefaultAccessTokenScopes
    .Add("https://XXXXX.onmicrosoft.com/eaee236e-faf8-442e-b472-7ff311d5ae76/API.Access");
```

This is an on-going problem from 3.x. Idk if you want to discuss it further. The topic does call it out as a *gotcha* to keep an :eye: out for.


aspnetcore/includes/blazor-security/azure-scope.md
```diff
@@ -1,14 +1,67 @@
+::: moniker range=">= aspnetcore-5.0"
+
+> [!NOTE]
+> If the app throws an unhandled exception when it receives a *401 Unauthorized* response from the API, inspecting the token error may reveal that the access token's audience isn't correct. This scenario is most likely to occur when the App ID URI of your *`Server`* app isn't `api://{SERVER API APP CLIENT ID OR CUSTOM VALUE}` but instead is in the format `https://{TENANT}.onmicrosoft.com/{SERVER API APP CLIENT ID OR CUSTOM VALUE}`. If that's the case, the default access token scope in `Program.Main` (`Program.cs`) of the *`Client`* app appears similar to the following after the solution is created from the Blazor Hosted template:
```
captainsafia, line 4: The difference between these two URIs depends on if the application is configured as B2C or AD. I would clarify that and make sure that people match the type of auth they are using with the scope that is provided.

Also, maybe add a note about how the templates will take care of setting the correct scope?

guardrex: > make sure that people match the type of auth they are using with the scope that is provided.

Implies that the initial setup and configuration guidance is incorrect, but Javier couldn't guide me when we set the topic up as to what's wrong with what it says. 

Take the Azure portal aspect ... It doesn't permit me to set the App ID URI outside of the server app client ID segment. That's never been explained ... do you know why my portal won't allow me to use an `api://`-based URI?

guardrex: @jmprieur ... Can I trouble u for a few minutes to look at something? If you have time, read my remarks in the opening comment under **Scopes ... scopes ... and SCOPES!** :point_up:. I noticed that you wrote ...

[What if the App ID URI of your application is not api://{ClientID}](https://github.com/AzureAD/microsoft-identity-web/wiki/Web-APIs#what-if-the-app-id-uri-of-your-application-is-not-apiclientid)

... and that's exactly what I need to make my hosted Blazor WASM with AAD work.

You said ...

> This is the default when your register your application with the application registration portal. 

... but it seems **more restrictive** than just a default. My portal won't allow me to have an `api://`-based App ID URI at all. It defaults to the format ...

```
https://{TENANT}.onmicrosoft.com/{SERVER API APP CLIENT ID OR CUSTOM VALUE}
```

... and will only allow me to change the segment after the `https://{TENANT}.onmicrosoft.com/` part. Here's what I see ...

Capture

Do you know why some of us, myself and a number of community members, can't change the App ID URI in the portal and why it defaults to that format?

WRT to the coverage here ... we have it. :+1: I assume in the coverage that the reader will receive the `api://`-based value and won't require additional config (i.e., `Authority` setting). **_BUT_** ... it is covered in a NOTE that will steer devs in the right direction if they have a portal like mine. I'm basically handling this the same way that you did with that section that you wrote.

jmprieur: @guardrex : I'd think that this is because the tenant in which you are working has a verified domain. Do you want us to take this offline to verify?

guardrex: @jmprieur ... :tada: **_BINGO!_** :tada: ... That's it. A virtual case of :beer: to you for that clarification. When I switched to my default directory, the publisher domain is verified, sooooooo ......

Capture

Capture1

Now that I understand this, I can clarify it in the doc for our readers. I'll probably cross-link [How to: Configure an application's publisher domain](https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-configure-publisher-domain) as an additional resource.

guardrex:
@captainsafia ... I just hit a 🧠🔥 *brain fry* 🧠🔥 for the day wrapping up these draft updates. In order not to have broken preview docs over the weekend, I'm going to go ahead and merge this now and take it live. I'll go back over these updates on Monday morning and possibly make further updates. The next major PR is probably Tuesday (Wednesday at the latest) and will focus on the AAD groups and roles topic.

PR #20148, opened 2020-10-08
Add QueryString to returnUrl in SelectLanguage

ScholliYT:
This allows users to stay on a page which needs to have a QueryString value when switching language from the dropdown menu

### DIFF
--- a/aspnetcore/fundamentals/localization/sample/3.x/Localization/Views/Shared/_SelectLanguagePartial.cshtml
+++ b/aspnetcore/fundamentals/localization/sample/3.x/Localization/Views/Shared/_SelectLanguagePartial.cshtml
@@ -13,6 +13,7 @@
         .Select(c => new SelectListItem { Value = c.Name, Text = c.DisplayName })
         .ToList();
     var returnUrl = string.IsNullOrEmpty(Context.Request.Path) ? "~/" : $"~{Context.Request.Path.Value}";
+    returnUrl += string.IsNullOrEmpty(Context.Request.QueryString.ToUriComponent()) ? "" : Context.Request.QueryString.ToUriComponent();
 }
 
 <div title="@Localizer["Request culture provider:"] @requestCulture?.Provider?.GetType().Name">


guardrex:
@hishamco

hishamco:
AFAIK the language switcher is using cookie not querystring and it should works, let me check the code now ..

hishamco:
https://github.com/dotnet/AspNetCore.Docs/blob/6d22c8de3137b976174f3b1b86f056337cbd3c9b/aspnetcore/fundamentals/localization/sample/3.x/Localization/Controllers/HomeController.cs#L40-L44

It uses a cookie to track the current culture, I'm not sure why you want to use the query string here?!! IMHO no need for this change

Thanks

guardrex:
Thanks @hishamco ... let's close based on that.

@ScholliYT ... If you're working with a different approach, we recommend the usual support channels. I'm sure you're familiar with SO, but the chats below can be great for discussing scenarios ...

* [Stack Overflow](https://stackoverflow.com/questions)
* [ASP.NET Core Slack](http://tattoocoder.com/aspnet-slack-sign-up/)
* [ASP.NET Gitter](https://gitter.im/aspnet/Home)

hishamco:
@guardrex You are welcome ;)

@ScholliYT if you have any question or concern about localization feel free to ask me

ScholliYT:
Thanks for taking a look @hishamco 

In my application I have a page like `app/editcompany?companyid=123` where there is a Combobox for selecting the language in the footer. When the user switches selected language the page will automatically reload with the new language. But it will load `app/editcompany` instead of `app/editcompany?companyid=123`.

hishamco:
Aha in this case you should tweak the url before the page is reloaded

ScholliYT:
Can you elaborate on this? @hishamco

hishamco:
```
var returnUrl = string.IsNullOrEmpty(Context.Request.Path) ? "~/" : $"~{Context.Request.Path.Value}";
```

Simply add your queystring into the `returnUrl` if it's not there

ScholliYT:
Well thats what I did here right? 
```cs
returnUrl += string.IsNullOrEmpty(Context.Request.QueryString.ToUriComponent()) ? "" : Context.Request.QueryString.ToUriComponent();
```

hishamco:
Seems right but I did deal too much with `ToUriComponent()`, just check if the `Request.Query` contains your key then add it to the `returnUrl`

PR #20164, opened 2020-10-10
Update the old integration package to new in key-vault-configuration.md

v-xuto:
Task [76906](https://dev.azure.com/jongio/docindex/_workitems/edit/76906/): [Article] Azure Key Vault Configuration Provider in ASP.NET Core.

### DIFF
--- a/aspnetcore/security/key-vault-configuration.md
+++ b/aspnetcore/security/key-vault-configuration.md
@@ -24,7 +24,7 @@ This document explains how to use the [Microsoft Azure Key Vault](https://azure.
 
 ## Packages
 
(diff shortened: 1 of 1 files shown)


guardrex:
Hello @v-xuto (and cc: @pakrym) ... I asked about versioning once before for Azure SDK updates. IIRC, these updates are good for >=3.0.

It's safest tho for me to ask again given that the API browser still isn't specific on versioning. It only says "Azure SDK for .NET ... Latest Preview") ...

https://docs.microsoft.com/en-us/dotnet/api/azure.extensions.aspnetcore.configuration.secrets.keyvaultsecretmanager?view=azure-dotnet

Will the new package and API work as far back as ASP.NET Core 3.0? If so, we're good without versioning these changes further.

**NOTE**: There's a sample app for this topic, so one of us should update that to go with these updates.

https://github.com/dotnet/AspNetCore.Docs/tree/master/aspnetcore/security/key-vault-configuration/samples/3.x/SampleApp

PR #20171, opened 2020-10-12
Added missing reference to MSBuild

gurmeetsinghdke:
Hyper link to MSBuild was missing so added that for easy redirection.


guardrex (approved):
Great idea @gurmeetsinghdke! :rocket:

PR #20173, opened 2020-10-12
Added quotes around parameters

bartosz-jarmuz:
Without the quotes the command won't work


guardrex (approved):
Thanks @bartosz-jarmuz! :rocket: ... The last version of PS was more flexible with it, and the PS docs used this exact format for these commands. Now, this is the *right thing to do*. :+1:

PR #20184, opened 2020-10-13
Remove unused using in Blazor RedirectToLogin component

scottsauber:
I found this `using` to be unused when implementing this component.


guardrex:
@scottsauber ... Thanks, but I'm going to leave it because it's part of the template ...

https://github.com/dotnet/aspnetcore/blob/648c15dbe90fb8c113f7c6b4adeb40d9e10494f6/src/ProjectTemplates/Web.ProjectTemplates/content/ComponentsWebAssembly-CSharp/Client/Shared/RedirectToLogin.razor#L2

If you want to open an issue for the product unit to remove it on their side, you can do that. If they agree, we could remove it over here for a future version of the docs.

scottsauber:
Good point - thanks @guardrex!

PR #20194, opened 2020-10-14
Update new-field.md

tamilselvan719:
Added the missing DB migration command (Visual Code SQL lite section) to make it easy for anyone new


Rick-Anderson:
> dotnet ef migrations add InitialCreate

You can't do that unless you deleted the migration. update looks at the DB state and runs all the migrations needed.

Rick-Anderson:
@tamilselvan719 check my suggested change. cc @tdykstra

tdykstra:
>  check my suggested change. 

Works if there's only been one migration, alternative that works regardless of number of migrations is to delete the Migrations folder.

Rick-Anderson:
@tamilselvan719  thanks for the PR, much appreciated. Keep them coming.

PR #20204, opened 2020-10-15
Update services.md -- wrong method name in example code

cantSleepNow:


JamesNK (approved):

JamesNK:
Good catch. Thanks for the fix.

PR #20247, opened 2020-10-20
Demonstrate SignalR JS header option

halter73:
I also added `transport` to the JS options table for consistency considering we also do that for the `accessTokenFactory` despite already demonstrating how to use it in an above section. I almost did the same for `logger` but decided against it considering `configureLogging()` is the right way to configure the `logger`.

@BrennanConroy

### DIFF
--- a/aspnetcore/signalr/configuration.md
+++ b/aspnetcore/signalr/configuration.md
@@ -366,6 +366,7 @@ var connection = new HubConnectionBuilder()
     .WithUrl("https://example.com/chathub", options => {
         options.Headers["Foo"] = "Bar";
         options.SkipNegotiation = true;
+        options.Transports = HttpTransportType.WebSockets;
         options.Cookies.Add(new Cookie(/* ... */);
         options.ClientCertificates.Add(/* ... */);
     })
@@ -379,7 +380,6 @@ let connection = new signalR.HubConnectionBuilder()
     .withUrl("/chathub", {
         // "Foo: Bar" will not be sent with WebSockets or Server-Sent Events requests
         headers: { "Foo": "Bar" },
-        skipNegotiation: true,
         transport: signalR.HttpTransportType.LongPolling 
     })
     .build();
--- a/aspnetcore/signalr/configuration.md
+++ b/aspnetcore/signalr/configuration.md
@@ -380,7 +380,7 @@ let connection = new signalR.HubConnectionBuilder()
         // "Foo: Bar" will not be sent with WebSockets or Server-Sent Events requests
         headers: { "Foo": "Bar" },
         skipNegotiation: true,
-        transport: signalR.HttpTransportType.WebSockets
+        transport: signalR.HttpTransportType.LongPolling 
     })
     .build();
 ```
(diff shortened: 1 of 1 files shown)


aspnetcore/signalr/configuration.md
```diff
@@ -343,6 +343,7 @@ Additional options can be configured in the `WithUrl` (`withUrl` in JavaScript)
 | JavaScript Option | Default Value | Description |
 | ----------------- | ------------- | ----------- |
 | `accessTokenFactory` | `null` | A function returning a string that is provided as a Bearer authentication token in HTTP requests. |
+| `transport` | `null` | An <xref:Microsoft.AspNetCore.Http.Connections.HttpTransportType> value specifying the transport to use for the connection. |
```
BrennanConroy, line 346: nit: this can be in all versions of the table. 2.1+

BrennanConroy (approved):

aspnetcore/signalr/configuration.md
```diff
@@ -375,8 +377,10 @@ In the JavaScript Client, these options can be provided in a JavaScript object p
 ```javascript
 let connection = new signalR.HubConnectionBuilder()
     .withUrl("/chathub", {
+        // "Foo: Bar" will not be sent with WebSockets or Server-Sent Events requests
+        headers: { "Foo": "Bar" },
         skipNegotiation: true,
-        transport: signalR.HttpTransportType.WebSockets
+        transport: signalR.HttpTransportType.LongPolling
```
halter73, line 383: I switched this from WebSockets to LongPolling. I want the header configuration to have an impact, and since skipNegotiation is true, WebSockets and SSE are out.

BrennanConroy: Well now the example will just throw, `skipNegotiation` only works with WebSockets. In fact it only works when you explicitly specify `transport: signalR.HttpTransportType.WebSockets`

aspnetcore/signalr/configuration.md
```diff
@@ -1808,6 +1815,7 @@ Additional options can be configured in the `WithUrl` (`withUrl` in JavaScript)
 | JavaScript Option | Default Value | Description |
 | ----------------- | ------------- | ----------- |
 | `accessTokenFactory` | `null` | A function returning a string that is provided as a Bearer authentication token in HTTP requests. |
+| `transport` | `null` | An <xref:Microsoft.AspNetCore.Http.Connections.HttpTransportType> value specifying the transport to use for the connection. |
```
halter73, line 1818: @BrennanConroy This is for `::: moniker range="< aspnetcore-2.2"`, so it goes back before 2.1, but this has been on IHttpConnectionOptions longer than anything else in this table.

https://github.com/dotnet/aspnetcore/commit/d169b96d2d794d52ae4c613df0854e2525314500#diff-5fb15f68b85f362298c3477405ec237742d208c5eaca152615a4c814a672f6b1R4-R7

BrennanConroy: Ok? I feel like I'm missing something. Is there a problem?

aspnetcore/signalr/configuration.md
halter73: Good catch. So do we remove headers or skipNegotiation from this sample?

aspnetcore/signalr/configuration.md
halter73: You asked for me to add this for "2.1+". I don't want to have to add a new moniker range. Sounds like we're in agreement that I don't have to.

aspnetcore/signalr/configuration.md
BrennanConroy: I vote for skipNegotiation to be removed. Also, should be removed from the .NET one above too, unless you add WebSockets.

aspnetcore/signalr/configuration.md
BrennanConroy: These docs don't exist for <2.1 so we're fine :)

guardrex (approved):
Merge when ready. 🚀 ... good luck to OSIRIS-REx tonight! 🍀🤞

PR #20265, opened 2020-10-22
Clarified "Test" link in TOC

daveabrock:
The `Test` link in the `Blazor` section for the .NET 5 preview docs was misleading to me. At first, I wasn't sure if it was put there accidentally or what—looks like it links to the wonderful section on testing Blazor components. I think changing it to something like `Test components` is more clear.


guardrex (approved):
Agreed @daveabrock :rocket: ... thanks for sending this in.

PR #20269, opened 2020-10-22
IdS with WebSockets and Server-Sent Events

### LINKED ISSUE #17751: SignalR Core Auth - Token in Query String section does not address AddIdentityServerJwt()
The Token in Query String section addresses the case where `AddJwtBearer()` is called directly:

```
 services.AddAuthentication()
        .AddJwtBearer(options =&gt;//options from documentation);
```

It does not cover the Identity Server case where `AddIdentityServerJwt()` is used instead, which is the default for Blazor WebAssembly apps scaffolded with authentication.

```
 services.AddAuthentication()
        .AddIdentityServerJwt();
```

In this case, calling `AddJwtBearer` after `AddIdentityServerJwt` does not work:

```
// this does not work
services.AddAuthentication()
                .AddIdentityServerJwt()
                .AddJwtBearer(options =&gt;//options from documentation);
```

---
#### Document Details

* ID: d0da2144-3730-f8ef-1d16-a89f11fd6198
* Version Independent ID: 6d8f5bc8-8f5c-ba13-b7f2-0366eac46f81
* Content: [Authentication and authorization in ASP.NET Core SignalR](https://docs.microsoft.com/en-us/aspnet/core/signalr/authn-and-authz?view=aspnetcore-3.1#feedback)
* Content Source: [aspnetcore/signalr/authn-and-authz.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/signalr/authn-and-authz.md)
* Product: **aspnet-core**
* GitHub Login: @bradygaster
* Microsoft Alias: **bradyg**

guardrex:
Fixes  #17751

[Internal Review Topic (links to section)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/hosted-with-identity-server?view=aspnetcore-3.1&branch=pr-en-us-20269#websockets-and-server-sent-events)

Brennan ... I'm winging this one even more than usual from the context on the issue. I need help with it ...

* Is this the best spot?
* Are these remarks sufficient/correct?
* Is Caleb's code good to provide?

### DIFF
--- a/aspnetcore/blazor/security/webassembly/hosted-with-identity-server.md
+++ b/aspnetcore/blazor/security/webassembly/hosted-with-identity-server.md
@@ -550,6 +550,50 @@ If troubleshooting a certificate loading problem, execute the following command 
 Get-ChildItem -path Cert:\CurrentUser\My -Recurse | Format-List DnsNameList, Subject, Thumbprint, EnhancedKeyUsageList
 ```
 
+## WebSockets and Server-Sent Events
+
+When the bearer token is transmitted as a query string parameter with WebSockets and Server-Sent Events, additional configuration is required.
+
+Add a <xref:Microsoft.Extensions.Options.PostConfigureOptions%601> service to the *`Server`* project:
+
+```csharp
+using Microsoft.AspNetCore.Authentication.JwtBearer;
+using Microsoft.Extensions.Options;
+
+public class ConfigureJwtBearerOptions : IPostConfigureOptions<JwtBearerOptions>
+{
+    public void PostConfigure(string name, JwtBearerOptions options)
+    {
+        var originalOnMessageReceived = options.Events.OnMessageReceived;
+
+        options.Events.OnMessageReceived = async context =>
+        {
+            await originalOnMessageReceived(context);
+                
+            if (string.IsNullOrEmpty(context.Token))
+            {
+                var accessToken = context.Request.Query["access_token"];
+                var path = context.HttpContext.Request.Path;
+                
+                if (!string.IsNullOrEmpty(accessToken) && 
+                    path.StartsWithSegments("/hubs"))
+                {
+                    context.Token = accessToken;
+                }
+            }
+        };
+    }
+}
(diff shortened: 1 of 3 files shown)


BrennanConroy:
I don't think this belongs in Blazor docs as it's specific to using SignalR directly. I would probably put it in the signalr/authn-and-authz doc.

guardrex:
I'll move it.

Does it warrant a cross-link from the Blazor IdS topic because Caleb said ...

> It does not cover the Identity Server case where `AddIdentityServerJwt()` is used instead, which is the default for Blazor WebAssembly apps scaffolded with authentication.

... implying that this was for a Blazor scenario in his case.

BrennanConroy:
> implying that this was for a Blazor scenario in his case

Right, they were using SignalR in Blazor WASM, and you'll notice that they commented on the SignalR docs because it's a SignalR scenario. The only reason Blazor is mentioned is because the scaffolding for a Blazor app adds IdentityServerJwt by default. Anyone using IdentityServerJwt with SignalR outside of a Blazor app will have a similar issue.

I *could* see a possible cross-reference in the Blazor-Wasm with SignalR tutorial as a potential "next steps, adding auth" type of thing.

guardrex:
Seems like it goes in the same section. I try to avoid unpaired child sections.

Does this compose well now?

BrennanConroy:
Should we have subsections for this to make it more obvious that this section is for IdentityServer and the other one is for the built in Jwt auth? Otherwise, looks good

guardrex:
See if the section naming on the last commit works.

Yes, I think *two* is good. I just didn't want to add one (unpaired) heading. I try to follow old skool outlining rules and pair sections (and bullets) when possible.

BrennanConroy:
Another nit: the section about the clients should probably go above the first subsection now?
From "In the JavaScript client, the token can be provided..." to "To support this on the server, additional configuration is required:"

BrennanConroy (approved):

PR #20280, opened 2020-10-23
Add more IEnumerable<TService> documentation

### LINKED ISSUE #18942: Document resolving IEnumerable<TService> when multiple registrations are provided
Given multiple registrations in **Startup.cs**

```csharp
{
    services.AddTransient<IService, ServiceA>();
    services.AddTransient<IService, ServiceB>();
    services.AddTransient<IService, ServiceC>();
}
```

You should be able to depend on an `IEnumerable<TService>` in the depending service.

```csharp
public Foo(IEnumerable<IService> services)
{
    this.services = services;
}
```

This does not appear to be documented yet, although, strangely, the opposite behavior is documented as the `TryAdd*` methods.

---
#### Document Details

* Content: [Dependency injection in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-3.1#service-registration-methods)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

halter73:
Fixes #18942

### DIFF
--- a/aspnetcore/fundamentals/dependency-injection.md
+++ b/aspnetcore/fundamentals/dependency-injection.md
@@ -223,7 +223,8 @@ services.TryAddSingleton<IMyDependency, DifferentDependency>();
 
 public class MyService
 {
-    public MyService(IMyDependency myDependency, IEnumberable<IMyDependency> myDependencies)
+    public MyService(IMyDependency myDependency, 
+        IEnumberable<IMyDependency> myDependencies)
     {
         Trace.Assert(myDependency is MyDependency);
         Trace.Assert(myDependencies.Single() is MyDependency);
@@ -231,17 +232,18 @@ public class MyService
 }
 ```
 
-`DifferentDependency` replaces `MyDependency` if the call to `TryAddSingleton` is changed to a second call to `AddSingleton` unless multiple `IMyDependency` instances are resolved via `IEnumerable<IMyDependency>`. In that case, the second call to `AddSingleton` adds `DifferentDependency` after `MyDependency` in the `IEnumerable<IMyDependency>`.
+`DifferentDependency` replaces `MyDependency` if the call to `TryAddSingleton` is changed to a second call to `AddSingleton`, unless multiple `IMyDependency` instances are resolved via `IEnumerable<IMyDependency>`. In that case, the second call to `AddSingleton` adds `DifferentDependency` after `MyDependency` in the `IEnumerable<IMyDependency>`.
+
+In the following example, the second call to `AddSingleton` overrides the previous one or adds to the previous one when multiple services are resolved via `IEnumerable<IMyDependency>`:
 
 ```csharp
 services.AddSingleton<IMyDependency, MyDependency>();
-// The following line overrides the previous one or adds to the previous one when
-// multiple services are resolved via IEnumerable<IMyDependency>:
 services.AddSingleton<IMyDependency, DifferentDependency>();
 
 public class MyService
 {
-    public MyService(IMyDependency myDependency, IEnumberable<IMyDependency> myDependencies)
+    public MyService(IMyDependency myDependency, 
+       IEnumberable<IMyDependency> myDependencies)
     {
         Debug.Assert(myDependency is DifferentDependency);
 
@@ -807,7 +809,8 @@ services.TryAddSingleton<IMyDependency, DifferentDependency>();
 
 public class MyService
 {
-    public MyService(IMyDependency myDependency, IEnumberable<IMyDependency> myDependencies)
+    public MyService(IMyDependency myDependency, 
+        IEnumberable<IMyDependency> myDependencies)
     {
         Trace.Assert(myDependency is MyDependency);
         Trace.Assert(myDependencies.Single() is MyDependency);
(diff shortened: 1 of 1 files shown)


mkArtakMSFT (approved):
Thanks @halter73

guardrex:
@halter73 ... I split out the code comments so that they will be localized via MT. Check my commit to make sure that I didn't break anything. Otherwise, merge whenever ur ready.

guardrex (approved):

Rick-Anderson (approved):

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -220,6 +220,38 @@ In the following example, the call to `AddSingleton` registers `MyDependency` as
...
+    {
+        Trace.Assert(myDependency is MyDependency);
+        Trace.Assert(myDependencies.Single() is MyDependency);
+    }
+}
+```
+
+`DifferentDependency` replaces `MyDependency` if the call to `TryAddSingleton` is changed to a second call to `AddSingleton`, unless multiple `IMyDependency` instances are resolved via `IEnumerable<IMyDependency>`. In that case, the second call to `AddSingleton` adds `DifferentDependency` after `MyDependency` in the `IEnumerable<IMyDependency>`.
+
+In the following example, the second call to `AddSingleton` overrides the previous one or adds to the previous one when multiple services are resolved via `IEnumerable<IMyDependency>`:
```
BrennanConroy, line 237:
```suggestion
In the following example, the second call to `AddSingleton` overrides the previous one when resolved as `IMyDependency` and adds to the previous one when multiple services are resolved via `IEnumerable<IMyDependency>`:
```

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -220,6 +220,38 @@ In the following example, the call to `AddSingleton` registers `MyDependency` as
...
+    public MyService(IMyDependency myDependency, 
+        IEnumberable<IMyDependency> myDependencies)
+    {
+        Trace.Assert(myDependency is MyDependency);
+        Trace.Assert(myDependencies.Single() is MyDependency);
+    }
+}
+```
+
+`DifferentDependency` replaces `MyDependency` if the call to `TryAddSingleton` is changed to a second call to `AddSingleton`, unless multiple `IMyDependency` instances are resolved via `IEnumerable<IMyDependency>`. In that case, the second call to `AddSingleton` adds `DifferentDependency` after `MyDependency` in the `IEnumerable<IMyDependency>`.
```
BrennanConroy, line 235: Is this bit needed if we immediately show the example below? I'm asking because that sentence is non-trivial to read.

aspnetcore/fundamentals/dependency-injection.md
halter73: What if I replace all of the prose above with the following? It's quite the long sentence but is probably still easier to grok than what I had before.

```suggestion
In the following example, the call to `TryAddSingleton` is changed to a second call to `AddSingleton` which overrides the previous one when resolved as `IMyDependency` and adds to the previous one when multiple services are resolved via `IEnumerable<IMyDependency>`:
```

aspnetcore/fundamentals/dependency-injection.md
halter73:
```suggestion
```

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -220,6 +220,38 @@ In the following example, the call to `AddSingleton` registers `MyDependency` as
 services.AddSingleton<IMyDependency, MyDependency>();
```
Tratcher, line 220: This doc should explain what happens when you register multiple services for an interface before it tries to describe how the TryAdd* overloads are different.

E.g.:
- Add{Lifetime} registers a service
- Calling Add{Lifetime} again registers another instance of that service. The last one registered will be used in most situations.
- Resolving `IEnumerable<{MyService}>` will return all services registered for that type in the order they were registered.
- TryAdd{Lifetime} avoids adding a service if there is already something registered for that interface
- TryAddEnumerable...

halter73: I was thinking the same thing. I went for the smaller change initially, but I'll try reordering and see if it reads better.

Tratcher: The `Multiple implementations` column above isn't explicitly explained and could be worked into this description.

halter73: I don't entirely understand the "Multiple implementations" column. Anything other than the TryAdd methods can add multiple implementations of a service. It would be a bit weird to do so if you aren't specifying different service and implementation types, but nothing's stopping you.

halter73: I guess I'm probably confusing "implementations" and "instances".

halter73: Do we add a "Multiple instances" column that's distinct from the "Multiple implementations" column? I feel that just confuses things further. I don't really want to touch that with this PR b/c then we start getting into rewrite territory.

Tratcher: Filing follow-up issues works too

halter73: I fixed the ordering and added a sentence about the distinction between "implementations" and instances." I'm not sure exactly how you want done with the table. Can you please file the issue if you still think we need one?

Tratcher (approved):

BrennanConroy (approved):

PR #20291, opened 2020-10-23
Referred to ModelState.IsValid as a property

maxalmonte14:
In the _Processing the POST Request_ section we have a paragraph telling us that "The `ModelState.IsValid` method verifies...", I thought it must be a mistake because as we can see in the [documentation ](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.modelbinding.modelstatedictionary.isvalid?view=aspnetcore-3.1#Microsoft_AspNetCore_Mvc_ModelBinding_ModelStateDictionary_IsValid) `ModelState.IsValid` is not a method but a property.


Rick-Anderson:
@maxalmonte14  Thanks!

maxalmonte14:
> @maxalmonte14 Thanks!

Pleased to help!

PR #20293, opened 2020-10-24
Correct Label of Supported Account Type Option

AdrianJSClark:
The "Supported account types" field has changed the text of the option to the value above. This change updates the text to match the value in the Azure Portal "Register an application" screen.


guardrex (approved):
Thanks @AdrianJSClark! :rocket:

PR #20301, opened 2020-10-26
add using to sockets sample

### LINKED ISSUE #19788: Calling Dispose after AcceptWebSocketAsync()
Hi,
I've been investigating high cpu usage and memory leak on WebSockets recently. It turned out that I was missing a `Dispose()` after accepting WebSocket.

Basically, before fix:
```
WebSocket webSocket = await context.WebSockets.AcceptWebSocketAsync();
... other code
```
After fix
```
using (WebSocket webSocket = await context.WebSockets.AcceptWebSocketAsync()) {
   ... other code
}
```

It turns out that the `Dispose()` fixes the issue in case the connection is abnormally closed without WebSockets close handshake. For normal closures, such leaks does not occur.

I think it should be mentioned in the documentation and in code samples/snippets that it should be disposed - right now it's not clear right away because the function call (AcceptWebSocketAsync) returns an object that receiver (my code) is supposed to call `Dispose` on. Also, current documentation somehow suggests that objects related to websocket connection are cleaned up after exiting pipeline (by saying that it should only be used before exiting pipeline/middleware) - at least that's the thought I had after reading and looking at the example without `Dispose`.

To justify putting `Dispose` in the documentation here are aspnet source codes that use `Dispose`:
https://github.com/dotnet/aspnetcore/blob/4e44025a52e4b73aa17e09a8041b0e166e0c5ce0/src/Middleware/WebSockets/test/ConformanceTests/AutobahnTestApp/Startup.cs#L25
https://github.com/dotnet/aspnetcore/blob/1867750ec997911a64a0a02c0eaf82cc8a2ba4ae/src/SignalR/common/Http.Connections/src/Internal/Transports/WebSocketsServerTransport.cs#L54

Note: I didn't have resources at the moment to test it but I think this fix (https://github.com/dotnet/corefx/commit/c5688531605218f58b30b377f48e74de4d6c83d3) might cause the leaks to never be there in the first place but it is not included in runtime version I'm working on (if I see correctly it's only included in release/3.0 and later). Nevertheless `Dispose` behaviour of object returned from `AcceptWebSocketAsync` should be mentioned in my opinion.

---
#### Document Details

* Content: [WebSockets support in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/websockets?view=aspnetcore-3.1)
* Content Source: [aspnetcore/fundamentals/websockets.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/fundamentals/websockets.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

Rick-Anderson:
Fixes #19788

[Internal review URL](#19788)

### DIFF
--- a/aspnetcore/fundamentals/websockets/samples/2.x/WebSocketsSample/Startup2.cs
+++ b/aspnetcore/fundamentals/websockets/samples/2.x/WebSocketsSample/Startup2.cs
@@ -40,7 +40,8 @@ namespace EchoApp
             }
 
             #region AcceptWebSocket
-            app.Use(async (context, next) => {
+            app.Use(async (context, next) =>
+            {
                 using (WebSocket webSocket = await context.WebSockets.AcceptWebSocketAsync())
                 {
                     var socketFinishedTcs = new TaskCompletionSource<object>();
--- a/aspnetcore/fundamentals/websockets/samples/2.x/WebSocketsSample/Startup2.cs
+++ b/aspnetcore/fundamentals/websockets/samples/2.x/WebSocketsSample/Startup2.cs
@@ -1,0 +1,65 @@
+﻿#define UseOptions // or NoOptions or UseOptionsAO
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.WebSockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Console;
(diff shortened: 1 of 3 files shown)


aspnetcore/fundamentals/websockets/samples/2.x/WebSocketsSample/Startup2.cs
```diff
@@ -0,0 +1,65 @@
...
+        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
+        {
+            if (env.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
+
+            #region AcceptWebSocket
+            app.Use(async (context, next) => {
```
BrennanConroy, line 43:
```suggestion
            app.Use(async (context, next) =>
            {
```

BrennanConroy (approved):

PR #20338, opened 2020-10-28
Update document date

jongalloway:
Updated document date
Changed "click" to "select" as better wording for accessibility (from team review)


guardrex (approved):

guardrex:
Ah ... I see a nit. That line needs a period. I'll shoot in a PR to fix that.

Issue #20349, opened 2020-10-29
Firefox does not trust self-signed certificates (dot net core development)


jaoord:
Even after cleaning old certificates and creating new certificates and restarting the browser, Firefox will not trust the localhost development certificate:

![image](https://user-images.githubusercontent.com/6464609/97552089-1a4dce80-19d4-11eb-9d54-9a794a078945.png)

How can we run a localhost dotnet development project in Firefox?

---
#### Documentdetails

⚠ *Dit gedeelte niet bewerken. Het is vereist om problemen te koppelen tussen docs.microsoft.com en GitHub.*

* ID: c7ddfb2f-5dea-5e31-2a27-02a4dc16104c
* Version Independent ID: 12ca0b9a-2638-a4ce-c45f-d4579c83dafb
* Content: [Enforce HTTPS in ASP.NET Core](https://docs.microsoft.com/nl-nl/aspnet/core/security/enforcing-ssl?view=aspnetcore-3.1&tabs=visual-studio)
* Content Source: [aspnetcore/security/enforcing-ssl.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/security/enforcing-ssl.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-security**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
Please transfer this to https://github.com/dotnet/AspNetCore.Docs/issues

Rick-Anderson:
@jaoord  see https://github.com/dotnet/AspNetCore.Docs/pull/20356

guardrex:
@jaoord ... I use FF, including 5.0 preview, and I'm not seeing any cert errors.

We probably can't provide product support here for that, so you can open an issue for the product unit to see if they want to investigate the cause. Before doing that tho, I recommend that you search the Net, particularly [SO](https://stackoverflow.com/), first just to make sure that someone hasn't already called it out and figured out the cause. Then ... open an issue here ...

https://github.com/dotnet/aspnetcore/issues

If you do find out what's causing it and think that docs need to mention a *gotcha*, then you can post back on this issue (or open a new one) with what you find out.

@Rick-Anderson I recommend that closing this for now until more information is available.

jaoord:
As mentioned in the other issue (#20356) the solution is to go to **about:config** in Firefox and set **security.enterprise_roots.enabled** to **true**.

[...]

PR #20356, opened 2020-10-29
FF doesn't trust dev cert

Rick-Anderson:
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/20349

### DIFF
--- a/aspnetcore/security/enforcing-ssl.md
+++ b/aspnetcore/security/enforcing-ssl.md
@@ -126,7 +126,7 @@ Specify the HTTPS port using any of the following approaches:
 > [!NOTE]
 > When an app is run in a reverse proxy configuration, <xref:Microsoft.AspNetCore.Hosting.Server.Features.IServerAddressesFeature> isn't available. Set the port using one of the other approaches described in this section.
 
-### Edge deployments 
+### Edge server deployments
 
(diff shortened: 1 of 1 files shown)


blowdart:
Seems odd, there should be an mechanism to import and trust in Firefox rather than ruling it out altogether.

guardrex:
I use FF, and I'm not having the trust issue that @jaoord indicated. I recommend more analysis on what might be causing that.

jaoord:
> 
> 
> I use FF, and I'm not having the trust issue that @jaoord indicated. I recommend more analysis on what might be causing that.

Interesting, because I also thought it used to work before. Are you on windows?

jaoord:
> 
> 
> I use FF, and I'm not having the trust issue that @jaoord indicated. I recommend more analysis on what might be causing that.

Interesting, because I also thought it used to work before. Are you on windows?

jaoord:
> I use FF, and I'm not having the trust issue that @jaoord indicated. I recommend more analysis on what might be causing that.

Thanks. I thought I was getting delusional because I thought it used to work before at my computer as well. I dug a little deeper and found this: [Firefox no longer trusts my internal certificate authority used for internal sites on our domain.](https://support.mozilla.org/en-US/questions/1175296#answer-1006092)

According to the accepted answer there you need to set **security.enterprise_roots.enabled** to **true** (in about:config).
Did that and verified that I can now visit my localhost dotnet core website over https.

PR #20366, opened 2020-10-30
Update hubcontext.md

Rick-Anderson:
Fixes https://github.com/dotnet/AspNetCore.Docs.tr-tr/issues/19
cc @guardrex @scottaddie @wadepickett  @BrennanConroy

### DIFF
--- a/aspnetcore/signalr/hubcontext.md
+++ b/aspnetcore/signalr/hubcontext.md
@@ -6,7 +6,7 @@ monikerRange: '>= aspnetcore-2.1'
 ms.author: bradyg
 ms.custom: mvc
 ms.date: 11/12/2019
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@JasonCard  this is a surprising over-translation

The style guide prevents using styling types in headings, maybe that rule should be abandoned.

guardrex:
Indeed ... Given our earlier convos with u, @JasonCard, I did style a few section headings that I knew would break if I didn't. 😈

JasonCard:
The no-loc metadata should fix this over-localization issue without having the inline code style in headings, but I don't have any concerns with using the code style in headings. FYI, I found a new bug was introduced to the no-loc feature where the terms protected by no-loc metadata are displayed like ":::no-loc(SignalR):::". We're currently following up on this bug.

guardrex:
> displayed like ":::no-loc(SignalR):::". We're currently following up on this bug.

There was a reader remark on that this morning. I'll let that reader know that it's being looked at when I get back from being OOF.

PR #20448, opened 2020-11-08
Fix typo "The works because..."

maxalmonte14:
Hi ASP .NET Core docs crew.

I think I found a little typo in this document; I'm assuming that what the original author wanted to say was "_This_ works..." but other options like "_It_ works..." could be also valid in this context.

Please let me know if this really is a typo.

Thanks in advance, have a great rest of the day.


guardrex (approved):
You got it @maxalmonte14! :rocket: ... thanks for sending in the fix.

PR #20468, opened 2020-11-10
Update HttpRepl docs for 5.0 changes

tlmii:
Updates the HttpRepl documentation for the changes we've made in 5.0.
Additionally, adds a new HTTP REPL node in the table of contents with two children - the original docs and the new telemetry doc.
The telemetry page is taken directly from https://github.com/dotnet/HttpRepl/blob/master/TELEMETRY.md

@scottaddie I didn't add any of the metadata to the top of the telemetry page since I wasn't sure if that was autogenerated or not.

### DIFF
--- a/aspnetcore/includes/http-repl/standard-options.md
+++ b/aspnetcore/includes/http-repl/standard-options.md
@@ -9,10 +9,6 @@
   * `{header}={value}`
   * `{header}:{value}`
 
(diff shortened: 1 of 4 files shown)


bradygaster (approved):

scottaddie (approved):
Thank you for your help, @tlmii!

Issue #20475, opened 2020-11-10
Update Blazor sample apps to 5.0


guardrex:
Include change from `app` to `#app` in `Program.Main` of WASM apps ...

```csharp
builder.RootComponents.Add<App>("#app");
```

... check the prerender+integrate topic for this change, too.

PR #20483, opened 2020-11-10
Updating Program.cs examples to "#app"

crgrieve:
Hello,

I noticed in latest Blazor WebAssembly sample code, the Program.cs now uses ("#app") rather than ("app"), so thought I would update to make consistent 😊

Thanks,
Carole

### DIFF
--- a/aspnetcore/blazor/common/samples/3.x/BlazorWebAssemblySample/Program.cs
+++ b/aspnetcore/blazor/common/samples/3.x/BlazorWebAssemblySample/Program.cs
@@ -13,7 +13,7 @@ namespace BlazorSample
         public static async Task Main(string[] args)
         {
             var builder = WebAssemblyHostBuilder.CreateDefault(args);
-            builder.RootComponents.Add<App>("app");
+            builder.RootComponents.Add<App>("#app");
 
             builder.Services.AddScoped(sp => new HttpClient { BaseAddress = new Uri(builder.HostEnvironment.BaseAddress) });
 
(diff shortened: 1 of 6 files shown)


guardrex:
This wouldn't surprise me. I think I did all of my testing with **_new_** 5.0 apps, and I didn't :ear: about this one from engineering.

I'll get back to you @crgrieve in a bit. We might need to version the changes before we can take action.

guardrex:
Confirmed @crgrieve ... This is a 5.0 change, but I'll need to version it in the sample app and in the sample code. It requires the use of versioning tags in the topics, and I might try to change some of the examples/samples to avoid showing the line in order to avoid having to version the code. I'm sorry we won't be able to accept this given the extra work required, but thanks for calling it out.

I'll take care of this tomorrow morning on https://github.com/dotnet/AspNetCore.Docs/issues/20475.

crgrieve:
No worries, thanks @guardrex. Glad my understanding was correct and hopefully a help 😊

Well done on the launch of 5.0 🎉

Thanks,
Carole

guardrex:
For sure ... I only knew about it briefly for the new *Prerendering and Integration* topic ...

https://docs.microsoft.com/en-us/aspnet/core/blazor/components/prerendering-and-integration?view=aspnetcore-5.0&pivots=webassembly#configuration

... until you showed it here. I think all of my test apps were generated new, and that `#` is something that my :eye: didn't catch. Thanks again for sending this in.

PR #20513, opened 2020-11-11
Update appsettingsSQLite.json

flaviomarcilio:
Update **ConnectionStrings** in appsettings.json file to correspond to **GetConnectionString()** in the ConfigureServices method in the Startup.cs file.


Rick-Anderson:
@flaviomarcilio  I think you're right. @wadepickett  please verify as you made the same change in the RP version.

wadepickett (approved):

wadepickett:
Thanks @flaviomarcilio!

@Rick, the connection string miss match was one of issues I corrected in the Razor Pages series.  It did originally have that same issue.  Good to check though, thanks.

PR #20521, opened 2020-11-12
Fix Typo

CreativeEdge007:
"ap" Should Be "App"


guardrex (approved):
Thanks @CreativeEdge007! :rocket:

PR #20531, opened 2020-11-12
Highlight is in the wrong place

intrepion:
The highlighted line for DTO code should be on the new Secret field

right now it is on Name


guardrex (approved):
Thanks @intrepion! :rocket: ... Yes, this happened because for the 5.0 version of the topic it was decided to include the namespace in the snippet, so it resulted in the highlight two lines earlier. Thanks for the fix!

PR #20532, opened 2020-11-12
Change 'razor.css' to '.razor.css'

northerngeek:
Small (probably incredibly pedantic!) change - the docs always refer to '.razor' files, so it would make sense for the 'razor.css' files to be written as '.razor.css' so idiots like me don't save a file as 'Indexrazor.css'. It would be consistent with [other .Net docs](https://docs.microsoft.com/en-us/aspnet/core/tutorials/razor-pages/razor-pages-start#pages-folder).


guardrex (approved):
Yes, thank you @northerngeek! :rocket: This topic is **_brand new_** for 5.0, so I'm not surprised about a few quirks like this. Thanks for the fix!

PR #20589, opened 2020-11-15
Dependency is a class rather than an object

wuyuansushen:


wuyuansushen:
I have misunderstood, sorry!

PR #20597, opened 2020-11-15
Modify Scoped lifetime description

wuyuansushen:


davidfowl:
I would re-define scoped services to explain that they are like singletons within a scope.An HTTP request, and outgoing HTTP client request, a SignalR hub invocation  are examples of unique scopes

wuyuansushen:
@davidfowl I think description about *scoped* in the docs have confused many learner like me........Please improve it.

PR #20626, opened 2020-11-17
Make install and update command more consistent

hitochan777:
* Fix typo in package name `Dotnet-aspnet-codegenerator`
* Use `-g` instead of `--global` to be consistent with `update` command


Rick-Anderson (approved):

Rick-Anderson:
Thanks!

PR #20628, opened 2020-11-17
Corrected error in text

erikly:
The text says `id`, while the code consistently uses a `class`. I haven't tested, but I'm guessing the code is correct?


guardrex:
Yes, thank you @erikly :eye:. I'll get this merged and live **_immediately!_** 🏃

PR #20633, opened 2020-11-17
Merge to live - 11/17

guardrex:


guardrex:
FYI @wadepickett ... This one generated some duplicate alt text suggestions. Perhaps, they were pre-existing. I noticed that the build system doesn't always flag them until a topic is touched with an update, so maybe that's what happened here.

PR #20651, opened 2020-11-18
Move Blazor pass tokens guidance

### LINKED ISSUE #20649: [Blazor][Server] Docs about how to avoid reliance on HTTP Context need to be in a more prominent place
Despite the fact that we have docs [here](https://docs.microsoft.com/en-us/aspnet/core/blazor/security/server/additional-scenarios?view=aspnetcore-5.0#pass-tokens-to-a-blazor-server-app), we keep getting issues about this.

I believe the problem is the topic is too specific to auth, and I would like us to have a separate topic in a different area that is more general. Something like:
Differences between Blazor Server apps and MVC apps, things to account for when developing blazor server apps, ASP.NET Core features you should avoid when developing blazor server apps or something like that.

And provide a more general guidance on things like HttpContext in addition to specific solutions on how to tackle the problem.

/cc: @danroth27

guardrex:
Fixes #20649

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/blazor-server-pass-httpcontext-tokens?view=aspnetcore-3.1&branch=pr-en-us-20651)

* This should catch the reader's :eye: for `HttpContext`. Is the title (**Pass HttpContext tokens to ASP.NET Core Blazor Server apps**) too specific? It doesn't allow for adding more *differences between Blazor Server and MVC/RP apps, things to avoid, or things to take into account* coverage later.
* ToC text is **Pass HttpContext tokens to Blazor Server**. I don't think it can appear in the ToC earlier than this (after the *Components* node). Everything prior is *fundamental* coverage.


javiercn:
Maybe the issue details weren't clear. 

I want to avoid talking about tokens specifically and have a topic about integration between ASP.NET Core and Blazor Server, and as part of that. The problem we have is that people end up using HttpContextAccessor because they don't find the current guidance, because is very specific to tokens.

My suggestion is that we create a new section about integration between ASP.NET Core and Blazor Server and we talk about these things in there. (Things from ASP.NET Core that people shouldn't use in their Blazor server apps and alternatives).

Using the sample with tokens for this is unnecessary complicated and distracts from the main key learnings that we want people to get from here, which are what things they should avoid and how to solve the problem in a different way.

So my proposal is to have a new topic on the index about integration between ASP.NET Core and Blazor server and discuss these topics there.

I also appreciate your eagerness to do the changes, but I want to make sure we don't waste your time, so I would prefer if we wait until other folks like @danroth27  had a chance to look at this.

PR #20666, opened 2020-11-18
First-mvc series: Moving snapshot code to its own snapshot directory

### LINKED ISSUE #20665: First-MVC Series: New-Field cshtml example should use MovieDateRatingDAmult.cs model and included only when defined
The cshtml files for view Views/Movies/IndexGenreRating.cshtml needs to use the model https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/tutorials/first-mvc-app/start-mvc/sample/MvcMovie3/Models/MovieDateRatingDAmult.cs.  It also needs to be set up to use #if to be included only when defined.

wadepickett:
[Internal Review: Section Add a Rating Property to the Movie Model ](https://review.docs.microsoft.com/en-us/aspnet/core/tutorials/first-mvc-app/new-field?view=aspnetcore-5.0&branch=pr-en-us-20666&tabs=visual-studio#add-a-rating-property-to-the-movie-model)

Fixes #20665 

Moving Views/Movies/IndexGenreRating.cshtml out of the 3.x and 5 sample downloads and into their own snapshot directory. 

The 3.x and 5 samples represent an earlier state of the application in the series and will not compile after download with IndexGenreRating.cshtml example in them.


aspnetcore/tutorials/first-mvc-app/new-field.md
```diff
@@ -58,13 +58,13 @@ Edit the */Views/Movies/Index.cshtml* file and add a `Rating` field:
 
 ::: moniker range=">= aspnetcore-3.0 < aspnetcore-5.0"
 
-[!code-cshtml[](~/tutorials/first-mvc-app/start-mvc/sample/MvcMovie3/Views/Movies/IndexGenreRating.cshtml?highlight=16,38&range=24-63)]
+[!code-cshtml[](~/tutorials/first-mvc-app/start-mvc/SnapshotNewField3/Views/Movies/IndexGenreRating.cshtml?highlight=16,38&range=24-63)]
```
Rick-Anderson, line 61: > The 3.x and 5 samples represent an earlier state of the application in the series and will not compile after download with IndexGenreRating.cshtml example in them.

That's because https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/tutorials/first-mvc-app/start-mvc/sample/MvcMovie3/Models/Movie.cs is defined. That's the wrong file. It should be  https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/tutorials/first-mvc-app/start-mvc/sample/MvcMovie3/Models/MovieDateRatingDAmult.cs

The right way to fix this is define the correct movie model for the completed applications.

Rick-Anderson (changes requested):

Rick-Anderson:
@wadepickett  this page needs to work with the completed download. It won't if it's moved outside the project.

wadepickett:
@Rick-Anderson, yes thanks.  I was trying to guess your original intent for how that is set up since the needed file didn't exist.  You are right, the model existed, but the related view didn't.  I'll change the issue description to match better and use the view naming convention you already have for the model.

Rick-Anderson:
> @wadepickett this page needs to work with the completed download. It won't if it's moved outside the project.

That is, the Razor page you moved needs to work with the completed download.

wadepickett:
Rick-Anderson,   I'm on the same page here, except I was not getting what that you were considering the completed sample sample to be.  There was only a "2.2 completed sample". Is there any reason you didn't provide a 3.x completed sample at the time?   Looks like I should create both a 3.x and 5 completed sample, and that is where the new-field.md referenced razor page view should live?    I guess I can add the razor page view to the 2.2-complete you are using, reference that for now, then create the other 3 and 5 completed samples as part of the larger  rework after that.

Rick-Anderson:
All you need to do to fix this problem is define the correct model, which I have identified. Then the download builds.

Rick-Anderson:
@wadepickett  https://github.com/dotnet/AspNetCore.Docs/tree/master/aspnetcore/tutorials/first-mvc-app/start-mvc/sample/MvcMovie3 is the completed 3.1 sample.

Rick-Anderson:
@wadepickett  call me if you have any problems and I'll help.

aspnetcore/tutorials/first-mvc-app/new-field.md
```diff
@@ -58,13 +58,13 @@ Edit the */Views/Movies/Index.cshtml* file and add a `Rating` field:
...
 
 ::: moniker-end
 
 ::: moniker range=">= aspnetcore-5.0"
 
-[!code-cshtml[](~/tutorials/first-mvc-app/start-mvc/sample/MvcMovie5/Views/Movies/IndexGenreRating.cshtml?highlight=16,38&range=24-63)]
+[!code-cshtml[](~/tutorials/first-mvc-app/start-mvc/SnapshotNewField5/Views/Movies/IndexGenreRating.cshtml?highlight=16,38&range=24-63)]
```
Rick-Anderson, line 67: Has this file changed between 3.1 and 5? I don't think the scaffolder changed. If that's true, you don't need a moniker here.

wadepickett:
Closing, since this was the wrong approach for adding the missing view that represented one portion of the tutorial, thanks Rick.  I will fix this in a new PR when I get back from vacation.  Understood that we are using #if directive to exclude/include the code from the compiled app as needed from the completed sample.

PR #20680, opened 2020-11-19
Fixed a typo in routing.md

sahan91:
Changed "the app to use **other** controllers and pages without..." to "the app to use **other** controllers and pages without...".


guardrex (approved):
Thanks @radical! :rocket:

PR #20690, opened 2020-11-20
Be more explicit about component lifecycle

philippdolder:
Be more explicit about the rendering that happens as soon as the first call is awaited. With the current description it sounds like it will only render after awaiting the `OnInitializedAsync` method.


guardrex:
I think 🤔 that the update preserves the concept of your suggestion but has the advantage of rolling it into the existing sentence. I'll ping Steve for an :eye: and advice/approval.

SteveSandersonMS:
It's fine with me - I don't mind whether you merge this change or not :)

I agree the updated text is correct, but then I also think the original text states the same information (in C# in general, the return point is the first `await` - this isn't specific to Blazor).

guardrex:
The **_"and then"_** language seems like it's really the problem here. Let's just remove the "then" part of the original text (on the last commit) and take additional feedback. We can try again if more readers indicate confusion.

Issue #20696, opened 2020-11-20
Update Blazor prerender with auth for 5.0


guardrex:
... or merely find out why there's a regression here. It might not be a 5.0-related problem.

cc: @christianweyer ... I probably won't reach this until after 🦃 *T-bird Day holiday* 🦃 ..... I'll get to this as soon as I can.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 53f3209c-2414-9b35-6245-383c9b7c6338
* Version Independent ID: cc8dfcd6-8d93-5aa3-c381-2b613e88cdbe
* Content: [ASP.NET Core Blazor WebAssembly additional security scenarios](https://docs.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/additional-scenarios?view=aspnetcore-5.0#support-prerendering-with-authentication)
* Content Source: [aspnetcore/blazor/security/webassembly/additional-scenarios.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/security/webassembly/additional-scenarios.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
@ChristianWeyer ... Is the non-working behavior **an error** :boom: or **no-op** in nature?

ChristianWeyer:
@guardrex I tested it with 5.0 only.
See the demo video here:

![BlazorDemo](https://user-images.githubusercontent.com/888718/99828435-3b30bc00-2b5b-11eb-9875-b158d6c978cb.gif)

When not doing prerendering, everything works fine and as expected.

I followed the docs here:
https://docs.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/additional-scenarios?view=aspnetcore-5.0#support-prerendering-with-authentication

guardrex:
Thanks for that @ChristianWeyer ... Given that you're getting an error, there wouldn't be anything wrong with you opening a companion product unit issue and showing that ... and put the error message out in the issue comment, which is probably faster for them to consume than watching the GIF.

https://github.com/dotnet/aspnetcore/issues

Leave this issue open and add a `cc: @guardrex` and a cross-link to this docs issue to your opening comment over there. They'll be probably faster in reaching the issue and responding and in a better position to troubleshoot it.

[...]

PR #20698, opened 2020-11-20
Update Blazor prerender with auth for 5.0

guardrex:
Addresses #20696

Thanks @ChristianWeyer ... I'll leave this as a DRAFT for now. We'll start with this and then see if it requires more work. Your issue will remain open until it's fully resolved.


guardrex:
@ChristianWeyer ... I'll go ahead and merge this now to correct the tag. This merely addresses your original issue, so it will remain open until they investigate and report back and see if/what we need to adjust the doc further.

PR #20701, opened 2020-11-20
Update FileUpload.razor

nza-dk:
person needs to be initialized before constructing EditContext, or an ArgumentNullException is thrown.


guardrex (approved):
Thanks @radical! :rocket: ... It was a typo from the original sample provided by the product unit and missed on review. My bad.

guardrex (approved):

PR #20714, opened 2020-11-21
Fixed Typo in Blazor WebAssembly performance best practices docs

Bigtalljosh:


guardrex (approved):
Thanks @Bigtalljosh! :rocket: ... This happens sometimes because in an editor with a wide window if the first word of the pair is at the end of one line and the second word is on the other side of the screen at the beginning of the next line, it seems like the :eye: doesn't catch it sometimes and these double-words sneak into the docs 😈.

Issue #16119, opened 2019-12-10
StateHasChanged Explained (vs base.InvokeAsync(StateHasChanged)


kevingates:
[EDIT by guardrex to add the metadata. We'll get an update from the product unit on this after 5.0 reaches GA. Everyone is 🏃😅 on 5.0 at the moment (9/24).] 

The differences between `StateHasChanged()` and `base.InvokeAsync(StateHasChanged);` is never clearly explained.

A topic on when a developer **should invoke** state has changed (via either method) as well has when is best to use each option would be helpful.

I understand that one handles things asynchronously, but are there times when it is better to have on method over the other? And what is being changed automatically vs us needing to call the method manually?

### Location:
Blazor --> Web Apps --> Components / State Management / "New Item"

### Abstract:
This topic will cover how State is managed in a Component, when it automatically updates, when a developer may want to manually update, and what options are available to the developer for updating the state.

### Potential Outline
1.0 - Definitions: Defining State in Blazor Components.
2.0 - How State is Updated
  2.1 - Automatic Updates
  2.2 - Manual Update
    2.2.1 - Example Scenarios when you will want to manually update.
3.0 - Manual Update Options
  3.1 - `StateHasUpdated();` Explained
    3.1.1 - Examples
  3.2 - `base.InvokeAsync(StateHasChanged);` Explained
    3.2.1 - Examples

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: a970fb72-3dd0-d173-585d-2ccdd7923f81
* Version Independent ID: 4871f4f7-9cf6-ac7a-0f21-c65d34bd1119
* Content: [ASP.NET Core Blazor lifecycle](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?view=aspnetcore-3.1#state-changes)
* Content Source: [aspnetcore/blazor/components/lifecycle.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/components/lifecycle.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Hello @kevingates ...

It's covered *lightly* in a spots where it's relevant to make the call, typically for a render when Blazor isn't aware that it should update the UI. For example:

[...]

PR #20727, opened 2020-11-22
Update event-handling.md

akovac35:
The intro provided an async example which in my opinion required some further elaboration regarding the usage of StateHasChanged().

### DIFF
--- a/aspnetcore/blazor/components/event-handling.md
+++ b/aspnetcore/blazor/components/event-handling.md
@@ -66,30 +66,6 @@ In the following example, `UpdateHeading` is called asynchronously when the butt
 }
 ```
 
-It is possible to add further intermediate UI updates with the help of `StateHasChanged`:
-
-```razor
-<h1>@Heading</h1>
-
-<button class="btn btn-primary" @onclick="UpdateHeading">
-    Update heading
-</button>
-
-@code {
-    private string Heading = "Initial heading";
-    
-    private async Task UpdateHeading(MouseEventArgs e)
-    {
-        Heading = "Started updating heading ...";
-	await ...
-	Heading = "Intermediate heading update ..."
-	StateHasChanged(); // Required only to display the intermediate heading update
-	await ...	
-	Heading = "Updated heading";
-    }
-}
-```
-
 ## Event argument types
 
 For some events, event argument types are permitted. Specifying an event parameter in an event method definition is optional and only necessary if the event type is used in the method. In the following example, the `MouseEventArgs` event argument is used in the `ShowMessage` method to set message text:
--- a/aspnetcore/blazor/components/event-handling.md
+++ b/aspnetcore/blazor/components/event-handling.md
@@ -45,17 +45,47 @@ The following code calls the `CheckChanged` method when the check box is changed
 
 Event handlers can also be asynchronous and return a <xref:System.Threading.Tasks.Task>. There's no need to manually call [StateHasChanged](xref:blazor/components/lifecycle#state-changes). Exceptions are logged when they occur.
 
-In the following example, `UpdateHeading` is called asynchronously when the button is selected:
+In the following example, `UpdateHeading` is called asynchronously when the button is clicked; the `Heading` variable will be changed from "Initial heading" through "Started updating heading" to "Updated heading":
 
 ```razor
+<h1>@Heading</h1>
+
 <button class="btn btn-primary" @onclick="UpdateHeading">
     Update heading
 </button>
 
 @code {
+    private string Heading = "Initial heading";
+    
     private async Task UpdateHeading(MouseEventArgs e)
     {
-        await ...
+        Heading = "Started updating heading ...";
+	await ...
+	Heading = "Updated heading";
+    }
+}
+```
+
+It is possible to add further intermediate UI updates with the help of `StateHasChanged`:
+
+```razor
+<h1>@Heading</h1>
+
+<button class="btn btn-primary" @onclick="UpdateHeading">
+    Update heading
+</button>
+
+@code {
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks @akovac35, but I don't agree with the changes. This is introductory content on event handling, and `StateHasChanged` isn't required to make the point.

I **_DO_** think more work has to be done in how `StateHasChanged` is described and demonstrated (I think we have an open issue for it.). It comes up in **_spots_** across the docs, but the section dedicated to it isn't fleshed out ...

https://docs.microsoft.com/aspnet/core/blazor/components/lifecycle?view=aspnetcore-5.0#state-changes

I think the example you're pitching here would go a lot better there. `StateHasChanged` is relevant for event handlers and mentioned just above this point ... i.e. ... the topic states that under typical conditions the dev won't need to call it. However, the current paragraph ...

> Event handlers can also be asynchronous and return a \<xref:System.Threading.Tasks.Task>. There's no need to manually call \[StateHasChanged](xref:blazor/components/lifecycle#state-changes). Exceptions are logged when they occur.

... *might be able to say* (unless product unit engineers have a problem with it) ...

> Event handlers can also be asynchronous and return a \<xref:System.Threading.Tasks.Task>. There's no need to manually call \[StateHasChanged](xref:blazor/components/lifecycle#state-changes) unless the event handler is required to explicitly update UI. Exceptions are logged when they occur.

... something like that. Would you like to give that a try and then create an example for the section at ...

https://docs.microsoft.com/aspnet/core/blazor/components/lifecycle?view=aspnetcore-5.0#state-changes

... for this PR?

akovac35:
Let me explain how I got here: it all started with the need for a loading indicator ... But there was nothing concrete regarding when the SHC needs to be called and when it may be redundant. So I made an example page for testing this out. It turned out the functionality is pretty clear but the description is hard to find. Should I just paste the example code here or give it a try with the page you recommended? BTW I think the first part of the header sample should stay, it is the second one with the second await which is the obvious misfit.

So for the page you recommended I would add an explanation regarding when the SHC is not needed and when it may be needed in the form of a few samples. Would that be ok?

guardrex:
> of a few samples

One sample probably. We'd need to discuss why more than one is needed and perhaps get engineering involved.

The general guidance is to call it when you need to force a UI render; otherwise, let the framework call it as needed. We kind'a say that thus far and only call it out in docs when that is the case with an example. That's why I was saying that it comes up in spots here and there.

It's not required for an async handler ... unless you do something like you proposed in your handler ... but that's another obvious case where the dev is going to need to trigger it on their own. It's already mentioned in that one line just above this text/example that you added, and the `StateHasChanged` section is already cross-linked there, so I think expanding that one line is sufficient here.

In the Lifecycle topic where the section resides ... that's where the coverage is weak. I recommend placing your example there. You can can cross-link back to the event handler topic section if you want to for the example's lead-in paragraph ... if you plan to use this async event handler scenario as the `StateHasChanged` example.

guardrex:
> nothing concrete regarding when the SHC needs to be called and when it may be redundant

That's the critical point ... the product unit engineers have mostly (at least) implied that devs should let the framework do it ... saves on unnecessary processing, of course ... but if you don't get the UI render behavior that you need relying on the framework, then add it.

> description is hard to find

Additional UE work on all of the topics will help with that. However as I said, just a few lines above this there is an existing cross-link. It just doesn't link to much of anything ...

https://docs.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?view=aspnetcore-5.0#state-changes

* No example there.
* Nothing saying 'the framework calls it automatically' ... 'extra, unnecessary calls lead to poor performance' ... but 'call it when you need to explicit trigger a render'.

guardrex:
You don't need to do this ..... I'll get to it ..... *eventually* 😅 ........

https://github.com/dotnet/AspNetCore.Docs/issues/16119

... but that's probably the issue I was thinking about that pertains to this. That will also (eventually) go in this section in the Lifecycle topic.

akovac35:
I suggest you accept the latest update as now it only updates the one example and it fits well in the page. As for all we have been discussing, I will open a separate "issue" so it can serve as a discussion etc.

guardrex:
Sorry @akovac35 ... I just don't agree with the changes. The change on the last commit has no context or purpose.

I'll cross-link this PR to my tracking issue for the Lifecycle topic and possibly work something like your original example into the `StateHasChanged` section on #16119 or the UE tracking issue updates. That's the direction that we would like to go.

Thanks again for discussing the topic and pitching the updates. Don't be dissuaded from submitting further **_issues_** for things that you would like to work on ... note that we prefer to discuss significant changes in an issue before reaching a PR. That way we can avoid community work that ultimately we decide not to accept.

Thanks again! :rocket:

guardrex:
@akovac35 ... Just wanted to follow up and let you know that your issue was addressed.

I'm getting into updates for the *Lifecycle* topic now on my [UE ("user experience") passes](https://github.com/dotnet/AspNetCore.Docs/issues/19286). The current section on "state changes" is still just a passing mention of `StateHasChanged` with some cross-links ... 

https://docs.microsoft.com/aspnet/core/blazor/components/lifecycle#state-changes

... **_BUT_** there's a new topic pertaining to this subject, which is cross-linked there ...

**ASP.NET Core Blazor component rendering**
https://docs.microsoft.com/aspnet/core/blazor/components/rendering

It includes example code and scenarios for when to call `StateHasChanged`.

akovac35:
@guardrex Hi Luke, a quick scan of the main two links and the outgoing connections seems to be pretty detailed & covers what devs will need from day to day, so good job and ++ for .NET and Blazor.

If I spot anything which could be added, I will open a separate issue / suggestions.

PR #20747, opened 2020-11-24
Update and improve Blazor unmarshalled JS

### LINKED ISSUE #20598: Unmarshalled JavaScript Isolation Code sample is incorrect
Unmarshalled JavaScript Isolation Code sample is incorrect
```cs
var unmarshalledRuntime = (IJSUnmarshalledRuntime)js;
var jsUnmarshalledReference = unmarshalledRuntime
    .InvokeUnmarshalled<IJSUnmarshalledObjectReference>("unmarshalledInstance");

string helloWorldString = jsUnmarshalledReference.InvokeUnmarshalled<string, string>(
    "helloWorld");
```

Error
```
There is no argument given that corresponds to the required formal parameter 'arg0' of 'IJSUnmarshalledObjectReference.InvokeUnmarshalled<T0, TResult>(string, T0)'
```
incorrect line 👇
```cs
string helloWorldString = jsUnmarshalledReference.InvokeUnmarshalled<string, string>(
    "helloWorld");
```

---
#### Document Details

* ID: aada0049-99a8-2f91-5d60-6970474e2371
* Version Independent ID: c29fb1a8-028f-6efd-f1bc-e889fb362311
* Content: [Call JavaScript functions from .NET methods in ASP.NET Core Blazor](https://docs.microsoft.com/en-us/aspnet/core/blazor/call-javascript-from-dotnet?view=aspnetcore-5.0#blazor-javascript-isolation-and-object-references)
* Content Source: [aspnetcore/blazor/call-javascript-from-dotnet.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/call-javascript-from-dotnet.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #20598

Thanks @sps014! :rocket:

There are big holes in my understanding here (*not surprising* 😄), so I need help.

* The [current example in the live topic](https://docs.microsoft.com/aspnet/core/blazor/call-javascript-from-dotnet#blazor-javascript-isolation-and-object-references) is broken and throws `There is no argument given that corresponds to the required formal parameter 'arg0'`.
* The new example on the PR is based on framework test code from [here](https://github.com/dotnet/aspnetcore/blob/master/src/Components/test/testassets/BasicTestApp/InteropComponent.razor#L198-L204) and [here](https://github.com/dotnet/aspnetcore/blob/master/src/Components/test/testassets/BasicTestApp/wwwroot/js/jsinteroptests.js#L222-L242) with [this](https://github.com/dotnet/aspnetcore/blob/master/src/Components/test/testassets/BasicTestApp/InteropTest/InteropStruct.cs).

Questions ...

1. WRT `int[]` and `string` between .NET and JS not matching ... https://github.com/dotnet/aspnetcore/issues/21492#issuecomment-637201926 ... is what I say on the PR (last line of the new section) correct? ... or how can that be made correct? Passing back a `string` did fail in my testing. Passing back an `int` or a `bool` worked fine.
1. WRT these struct field `FieldOffset` values ... any guidance to add to the PR regarding the values? I have no knowledge on this subject. I cross-link the `struct` C# topic earlier. Is that enough for advanced devs working with advanced code to grok what to do?
1. The framework example has a JS `dispose`, which I include here. Should that be included? Anything to say about it in the text?

### DIFF
--- a/aspnetcore/blazor/call-javascript-from-dotnet.md
+++ b/aspnetcore/blazor/call-javascript-from-dotnet.md
@@ -526,28 +526,6 @@ public async ValueTask<string> Prompt(string message)
 
 `IJSInProcessObjectReference` represents a reference to a JavaScript object whose functions can be invoked synchronously.
 
-`IJSUnmarshalledObjectReference` represents a reference to an JavaScript object whose functions can be invoked without the overhead of serializing .NET data. This can be used in Blazor WebAssembly when performance is crucial:
-
-```javascript
-window.unmarshalledInstance = {
-  helloWorld: function (personNamePointer) {
-    const personName = Blazor.platform.readStringField(value, 0);
-    return `Hello ${personName}`;
-  }
-};
-```
-
-```csharp
-var unmarshalledRuntime = (IJSUnmarshalledRuntime)js;
-var jsUnmarshalledReference = unmarshalledRuntime
-    .InvokeUnmarshalled<IJSUnmarshalledObjectReference>("unmarshalledInstance");
-
-string helloWorldString = jsUnmarshalledReference.InvokeUnmarshalled<string, string>(
-    "helloWorld");
-```
-
-In the preceding example, the <xref:Microsoft.JSInterop.IJSRuntime> service is injected into the class and assigned to `js` (not shown).
-
 ## Use of JavaScript libraries that render UI (DOM elements)
 
 Sometimes you may wish to use JavaScript libraries that produce visible user interface elements within the browser DOM. At first glance, this might seem difficult because Blazor's diffing system relies on having control over the tree of DOM elements and runs into errors if some external code mutates the DOM tree and invalidates its mechanism for applying diffs. This isn't a Blazor-specific limitation. The same challenge occurs with any diff-based UI framework.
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/call-javascript-from-dotnet.md
```diff
@@ -691,6 +669,92 @@ Consider the following guidance when developing code that transfers a large amou
...
+        [FieldOffset(0)]
+        public string Name;
+
+        [FieldOffset(8)]
+        public int Year;
+    }
+}
+```
+
+Integer array types (`int[]` in C#) and strings (`string` in C#) don't match between .NET and JavaScript and require additional code to pass back to .NET from unmarshalled JavaScript functions.
```
pranavkm, line 756: The `string` one is fairly easy. You could document it with the caveat that the JS APIs are subject to breaking change:

```js
function unmarshalledFunctionReturningString() {
    BINDING.js_string_to_mono_string("Hello");
}
```

which can be read as `jsUnmarshalledReference.InvokeUnmarshalled<string>`. Arrays are a bit more complicated. Some numeric arrays can be converted using `js_typed_array_to_array` (https://github.com/dotnet/aspnetcore/blob/648c15dbe90fb8c113f7c6b4adeb40d9e10494f6/src/Components/Web.JS/src/Platform/WebAssemblyConfigLoader.ts#L15) but it must be a [typed array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays). You could do other data types (such as string arrays), but it's much less straightforward: 

https://github.com/dotnet/aspnetcore/blob/e8dade6436c7359920d726084e4fd80cb54a29e5/src/Components/Web.JS/src/Platform/Mono/MonoPlatform.ts#L382-L386

That one is an array of bytes, but it could just as well be an array of strings. On the receiving end, I think you're allowed to read arrays as `object[]`

guardrex:
@pranavkm ... Updates made. I'm making some phrasing guesses based on your language. It probably needs a bit more work. ⚒️

Also, I like the way this goes with two examples ... one for simple, direct conversion (boolean) and one for the special binding case (string). See if you agree on direction and composition.

aspnetcore/blazor/call-javascript-from-dotnet.md
```diff
@@ -691,6 +669,149 @@ Consider the following guidance when developing code that transfers a large amou
...
+            const year = Blazor.platform.readInt32Field(fields, 8);
+
+            return name === "Brigadier Alistair Gordon Lethbridge-Stewart" &&
+                year === 1968;
+        },
+        unmarshalledFunctionReturnString: function (fields) {
+            const name = Blazor.platform.readStringField(fields, 0);
+            const year = Blazor.platform.readInt32Field(fields, 8);
+
+            return BINDING.js_string_to_mono_string(`Hello, ${name} (${year})!`);
```
pranavkm, line 706: Should we call out that this function is very likely to be broken in future releases? We definitely want to scrub the term `mono` from this APIs. It's also possible that we remove the need to do this entirely (the same way it works for bool values).

aspnetcore/blazor/call-javascript-from-dotnet.md
```diff
@@ -691,6 +669,149 @@ Consider the following guidance when developing code that transfers a large amou
...
+        [FieldOffset(0)]
+        public string Name;
+
+        [FieldOffset(8)]
+        public int Year;
+    }
+}
+```
+
+If an `IJSUnmarshalledRuntime` instance isn't disposed in C# code, it can be disposed in JavaScript. The following `dispose` function disposes the object reference when called from JavaScript:
```
pranavkm, line 797:
```suggestion
If an `IJSUnmarshalledObjectReference` instance isn't disposed in C# code, it can be disposed in JavaScript. The following `dispose` function disposes the object reference when called from JavaScript:
```

aspnetcore/blazor/call-javascript-from-dotnet.md
```diff
@@ -691,6 +669,149 @@ Consider the following guidance when developing code that transfers a large amou
...
+        dispose: function () {
+            DotNet.disposeJSObjectReference(this);
+        },
+
+        ...
+    };
+}
+```
+
+Array types can be converted from JavaScript objects into .NET objects using `js_typed_array_to_array`, but the JavaScript array must be a typed array. Arrays from JavaScript can be read in C# code as a .NET object array (`object[]`). For examples, search the [dotnet/aspnetcore reference source](https://github.com/dotnet/aspnetcore/search?q=js_typed_array_to_array).
```
pranavkm, line 811: I hesitate to point to our code, but could we link them to our source files at a specific commit? These function names are very likely to change in 6.0 at which point these searches would not yield any results.

pranavkm (approved):

guardrex:
@pranavkm I see what you mean about the links to source. I removed them. If we link to a specific release, devs will use it directly and will still end up in a broken state 💥. It's best to just have these warnings about potential changes and then :ear: for their feedback if they don't catch them (and hopefully let us know with a docs issue 🤞).

See if the last commit WARNINGS are ok for these types of changes.

aspnetcore/blazor/call-javascript-from-dotnet.md
```diff
@@ -709,6 +709,13 @@ window.returnJSObjectReference = () => {
 }
 ```
 
+> [!WARNING]
```
pranavkm, line 712: 👍

aspnetcore/blazor/call-javascript-from-dotnet.md
```diff
@@ -709,6 +709,13 @@ window.returnJSObjectReference = () => {
...
+> The `js_string_to_mono_string` function name, behavior, and existence is subject to change in a future release of .NET. For example:
+>
+> * The `mono` segment of the function name might be removed.
```
pranavkm, line 715:
```suggestion
> * The function is likely to be renamed
```

aspnetcore/blazor/call-javascript-from-dotnet.md
```diff
@@ -709,6 +709,13 @@ window.returnJSObjectReference = () => {
...
+> * The behavior of the function may change.
```
pranavkm, line 716:
```suggestion
```

It's unlikely to happen

pranavkm (approved):

PR #20768, opened 2020-11-25
Delete orphan images

gewarren:
The build is clean after deleting these 209 images.


guardrex (changes requested):
Hello @gewarren ... This can't be merged in its present form. It's deleting images that are part of (and used by) sample apps.

gewarren:
Okay, I'll just abandon it. It looks like most have them have "sample" in their path.

PR #20776, opened 2020-11-26
update css-isolation.md

pmdot:
written the same word twice was removed


guardrex (approved):
Thanks @pmdot! :rocket:

PR #20782, opened 2020-11-27
Update testing.md

JulioNobre:
Replacemente for broken link


guardrex (approved):
Thanks @JulioNobre! :rocket: ... btw -- We use relative links and drop the localization segment.

PR #20790, opened 2020-11-28
Strengthen the instruction and update button name

guardrex:
Can we 💪 **_strengthen_** 💪 the guidance on using the **This page** feedback button and form?

Even if we don't tho, let's at least take a look at the following ...

Although "**Content feedback** button" is for the "Feedback" section and does pertain to the "content," it really isn't the "**Content feedback** button." Using "**This page** button" is clear imo.

Just for list consistency, I put asterisks in for dashes below.

### DIFF
--- a/.github/ISSUE_TEMPLATE/doc-issue.md
+++ b/.github/ISSUE_TEMPLATE/doc-issue.md
@@ -5,15 +5,23 @@ about: Create an issue to help us improve
 
 # Before you open an issue
 
-If the issue is with an ASP.NET Core document, open the issue with the **Content feedback** button at the bottom of that document page.
+If the issue is with an ASP.NET Core document:
+
+1. Do **not** open a new issue using this form.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/doc-issue.md
```diff
@@ -23,9 +31,9 @@ If the issue is:
 
 Check the .NET target framework(s) being used, and include the version number(s).
 
-- [ ] .NET Core
-- [ ] .NET Framework
-- [ ] .NET Standard
+* [ ] .NET Core
```
scottaddie, line 34: While you're in here, maybe we should add another item for .NET 5.0+.

scottaddie:
@guardrex I'm okay with your proposed changes. See my suggestion above.

scottaddie (approved):

PR #20819, opened 2020-12-01
app.UseResponseCompression(); to middleware ordering

Rick-Anderson:

### DIFF
--- a/aspnetcore/fundamentals/middleware/index.md
+++ b/aspnetcore/fundamentals/middleware/index.md
@@ -73,7 +73,7 @@ The **Endpoint** middleware in the preceding diagram executes the filter pipelin
 
 The order that middleware components are added in the `Startup.Configure` method defines the order in which the middleware components are invoked on requests and the reverse order for the response. The order is **critical** for security, performance, and functionality.
 
-The following `Startup.Configure` method adds security-related middleware components in the recommended order:
+The following `Startup.Configure` method adds security-related middleware components in the typical recommended order:
 
 [!code-csharp[](index/snapshot/StartupAll3.cs?name=snippet)]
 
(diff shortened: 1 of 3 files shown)


aspnetcore/fundamentals/middleware/index/snapshot/StartupAll3.cs
```diff
@@ -53,6 +53,7 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
             app.UseAuthentication();
             app.UseAuthorization();
             // app.UseSession();
+            // app.UseResponseCompression();
```
BrennanConroy, line 56: Might want to compress before caching? @Tratcher

Tratcher: This is fine. Caching and compression ordering are scenario specific, there's multiple valid orderings. Yes, you can compress and then cache, but then you might need to store multiple compressed representations. I've also seen these combine with static files to allow you to cache compressed static files.
UseResponseCaching
UseResponseCompression
UseStaticFiles

BrennanConroy (approved):

aspnetcore/fundamentals/middleware/index.md
```diff
@@ -84,6 +84,23 @@ In the preceding code:
...
   * `UseCors` currently must go before `UseResponseCaching` due to [this bug](https://github.com/dotnet/aspnetcore/issues/23218).
 
+In some scenarios, middleware will have different ordering. For example, caching and compression ordering is scenario specific, and there's multiple valid orderings. For example:
+
+```csharp
+app.UseResponseCompression();
+app.UseResponseCaching();
+```
+
+With the preceding code, you might need to store multiple cached representations.
```
Tratcher, line 94:
```suggestion
```csharp
app.UseResponseCaching();
app.UseResponseCompression();
```

With the preceding code, you could save CPU by caching the compressed response, but you might end up caching multiple representations of a resource using different compression algorithms such as gzip or brotli.
```

Tratcher (approved):

PR #20850, opened 2020-12-03
Add example with set extension methods to memory cache usage

### LINKED ISSUE #20849: Improved MemoryCache Usage
The first example of using the MemoryCache creates an extra ``MemoryCacheEntryOptions`` object,just to set the sliding expiry time. Unfortunately this example is a bit misleading, because it is unclear if ``MemoryCacheEntryOptions`` can be reused as it is not marked thread-safe. The bottom line is, that this being the first example, it will lead to developers creating that object just to set extra values even though there are better options available (as you know, everyone just looks for the first copyable code piece ;)). I will create a PR which adds a code sample with Set and absolute expiry without the ``MemoryCacheEntryOptions``  in it. Imho it would be better to change the first example, not to use sliding but this would require rather large changes to the narrative of this page or just use the ``CreateEntry`` api, but that api is way too clunky to be used in normal circumstances.

This is a direct result of https://github.com/dotnet/runtime/issues/45436

---
#### Document Details

* Content: [Cache in-memory in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/memory?view=aspnetcore-5.0)
* Content Source: [aspnetcore/performance/caching/memory.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/performance/caching/memory.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

Tornhoof:
This adds an example without creating the `MemoryCacheEntryOptions` object.
Fixes #20849

### DIFF
--- a/aspnetcore/performance/caching/memory.md
+++ b/aspnetcore/performance/caching/memory.md
@@ -67,7 +67,7 @@ The current time and the cached time are displayed:
 
 [!code-cshtml[](memory/3.0sample/WebCacheSample/Views/Home/Cache.cshtml)]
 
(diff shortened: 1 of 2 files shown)


Rick-Anderson (approved):
Looks good. The link is correct/tested. This is how it renders:
image

adamsitnik (approved):
LGTM, thank you @Tornhoof !

Rick-Anderson:
@Tornhoof  thanks for this important PR, much appreciated.

PR #20856, opened 2020-12-03
Blazor AAD groups and roles topic 5.0

### LINKED ISSUE #17683: Blazor WASM with AAD groups and roles
In progress.

---
#### Document Details

* ID: 217bc7a2-94db-e23b-9e74-7ad48b884462
* Version Independent ID: 071763aa-0173-af10-9634-9c50a4d5868b
* Content: [ASP.NET Core Blazor WebAssembly with Azure Active Directory groups and roles](https://docs.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/azure-active-directory-groups-and-roles?view=aspnetcore-5.0)
* Content Source: [aspnetcore/blazor/security/webassembly/azure-active-directory-groups-and-roles.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/security/webassembly/azure-active-directory-groups-and-roles.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #17683

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/azure-active-directory-groups-and-roles?view=aspnetcore-5.0&branch=pr-en-us-20856)

* Drops the prior Identity Platform v1.0 coverage (e.g., ADAL). Mirrors what Azure docs have done, and this simplifies the forward maintenance of our coverage **_a lot_**. I kept a PDF of the current coverage. Do you you think we should provide and cross-link the PDF to 3.1/Identity v1.0 coverage? I *vastly prefer* doing that than trying to keep the content in the topic. :ear:
* Adopts latest guidance from RBAC Azure engineers: Use **Role Template IDs** for AAD Administrator Roles (`wids` claims) because `groups` claims only contain role **entity IDs** that can't be trusted for stability under Identity v2.0.
* Uses existing approaches for Graph API (Javier's bits in the [Graph API](https://docs.microsoft.com/aspnet/core/blazor/security/webassembly/graph-api) topic) and a custom user account with a custom user based on `RemoteUserAccount`. ... and btw ... I *highly recommend* that we keep the mobile phone example here because I know that devs will wonder how to integrate the Graph API call. By showing it, we head off doc issues and get readers on the right approach.
* The `MemberOf` call is required for security groups here because the `groups` claim includes **entity IDs** for AAD roles **_that we do NOT want to create claims out of_** and there's no way to distinguish GUIDs in the claim (their type) between (good) security group GUIDs and (wrong) role GUIDs unless a separate Graph API call is made. With the result of a Graph API call, we can check `entry.ODataType` for `"#microsoft.graph.group"` to ensure that we're creating **good claims** for the user that are only security groups. This also avoids the claims overage (i.e., `hasGroups`) scenario.
* I tested with more than five AAD Admin Roles, and Azure never tossed anything like a `hasGroups` my way for `wids`, and the Azure docs don't discuss such a thing even existing. `wids` seems to be limited by claim count. For JSON, the limit is perhaps 200 roles (*guessing* from Azure docs).
* We're still waiting to hear from the RBAC folks on the *mysterious* `wids` claim with a GUID of `b79fbf4d-3ef9-4689-8143-76b194e85509`. That's under investigation. I don't mention it right now, but I might need to add a NOTE to the topic at some point. It's confusing why it shows up because it's not a Role Template ID but it appears with them and doesn't match anything in an Azure portal app registration, groups, roles, or otherwise.
* I name claims client-side because I'm creating them and thus it's easy & nice to do so. Server-side, I just go with existing names from Azure. I'm :ear: for changes on the approach ... naming ... etc.

Courtesy ping @blowdart on another one of my attempts to **_NOT bring down the Internet_** with a security-based Blazor PR. 🙈

cc: @jmprieur @abhijeetsinha @VinceSmith


guardrex:
@mkArtakMSFT ... I'm not getting any ❤️ on this one. This is our 5.0/Identity v2.0 coverage, so we're getting late with it. Shall we go ahead and merge? We can take reader feedback on the approaches outlined and see how it goes.

danroth27:
@jmprieur Can you take a look at this doc? Or is there someone else on the Identity team that could help out with a review?

guardrex:
@danroth27 @mkArtakMSFT ... This one is in *limbo*. Do you want to keep waiting? We can merge and then take feedback from the community. I'll jump on any reported problems/concerns as they come in.

Either way, I'll ask @abhijeetsinha offline if there's any news on the *mysterious* wids claim that shows up but doesn't seem to represent any group/role in AAD. It was under investigation by AAD 🐈🐈 back when this was put up.

guardrex:
**UpDaTe!** @danroth27 @mkArtakMSFT .... **_Now, we're cook'in!_** 🍖 I'm in discussion with stakeholders in AAD. They've already explained some of the *mystery* 😄 behind these GUIDs. They're going to get back to me with more info shortly. I'm going to move this back to DRAFT status until I get this final piece worked out with them.

endintiers:
Hi guardrex,
I like your stuff on AAD Groups and Roles, but there are no clear instructions on App Roles, just a reference to doing the same as you would for groups and AD Roles, but that is not correct.
When using AddMsalAuthentication on an Application with "groupMembershipClaims": "SecurityGroup",  the App Roles associated with this User (and User's Groups) are already provided as a Claim (Type == "roles") when the user is Authenticated.
So there is no need to talk to the Server or Graph API, the information already exists - although we have to transform the 'roles' claim to multiple 'role' claims (if that is the type we are using) - so we still need to implement a custom AccountClaimsPrincipalFactory.
The bit that wasn't documented is that you have to use
                .AddAccountClaimsPrincipalFactory<CustomUserFactory>();
rather than
                .AddAccountClaimsPrincipalFactory<RemoteAuthenticationState, CustomUserAccount, CustomUserFactory>();
because there is no need for remote state - the information is already available in a Claim.
Took me several hours of "mashing keyboard" to work that out.
Thanks, Chris

guardrex:
> but there are no clear instructions on App Roles, just a reference to doing the same as you would for groups and AD Roles

"App Roles" are called "user-defined roles" here. They're read into individual claims from the `Roles` property (fed by `[JsonPropertyName("roles")]`) from the token ...

```csharp
foreach (var role in account.Roles)
{
    userIdentity.AddClaim(new Claim("userDefinedRole", role));
}
```

This approach was sent over by Javier when the first version of the topic was written and remains in use for the guidance on this PR. The rest of the bits are for AAD Administrator roles (via `wids` claims) and groups (from Graph SDK/API).

But *Yes!* 🤔 WRT "app roles" versus "user-defined roles." I think the naming should be changed here to match the Azure docs and app manifest. Good idea. I might play with "user-defined app roles," but I'm not sure the "user-defined" bit is necessary. The hard part is the word "roles" ... "App Roles" shouldn't be confused with built-in "AAD Administrator Roles." I think what I'll do is capitalize as "App Roles" to formally name the feature. It's a specific, named thing, so it can be a formal noun.

Groups can be obtained from `groups` claims, but:

* They're interspersed with AAD Admin Role "entity Ids" that shouldn't be used at all going forward in favor of "role template IDs".
* Groups claims run out of space quickly, and you lose it in favor of a `hasGroups` overage claim.

With Graph SDK/API, one can:

* Avoid AAD Admin Role entity IDs.
* Avoid the low limit of groups in `groups` claims (i.e., `hasGroups`).
* Obtain additional info about users, which is a common requirement.

However ... *Yes!* You can work directly with `roles` claims and avoid Javier's approach; and when an app doesn't need Graph SDK/API, you don't have to use it.

We won't be able to cover all of the permutations, and devs will need to adapt the code to their needs. We'll provide **_one approach_** suitable for App Roles, AAD Admin Roles, and security groups, and devs can take it from there. What we need to do here imo is pull together scattered bits of information, often without clear examples, into one coherent approach for these requirements with examples. It's easy for a dev to then cut out the bits that they don't need.

guardrex:
I've updated the PR ... 'user-defined roles' 👉 'App Roles' ... and I like it very much. Thanks for the feedback @endintiers! 🎷

mkArtakMSFT (approved):
Thank you, @guardrex!.

Let's merge this PR now. We can address any feedback either from partners or the community in separate PRs.

guardrex:
# 🍻 *Booyeah!*

Happy to get this one wrapped up (for now).

I'll reach this topic for another look :eyes: on the [UE tracking pass](https://github.com/dotnet/AspNetCore.Docs/issues/19286) at some point this year. All of the examples will be retested at that time as well.

PR #20874, opened 2020-12-05
Updates to Routing

poornimanayar:
Improved the Routing docs


guardrex (approved):
Thanks @poornimanayar! :rocket:

Let's check with @javiercn on the timing ("when the app boots") and see if there's anything else that might be helpful to convey in this paragraph.

PR #20885, opened 2020-12-07
Change MyLoginDialog to CustomLoginDialog

mahyar1386:
That part should use `CustomLoginDialog` instead of `MyLoginDialog`, But since the rest of the section is using `MyLoginDialog` it's kind of confusing. It would be better to change the `CustomLoginDialog` Type in lines 442 & 445 to `MyLoginDialog` for the sake of integrity.

### DIFF
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -451,7 +451,7 @@ Component references provide a way to reference a component instance so that you
 }
 ```
 
-When the component is rendered, the `loginDialog` field is populated with the `MyLoginDialog` child component instance. You can then invoke .NET methods on the component instance.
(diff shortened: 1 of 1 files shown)


guardrex (approved):
Thanks @mahyar1386 ... *Yes!* ... The example was a "My"-based name originally, and I've been moving code away from "My"-based naming across all of the Blazor docs. I missed this reference when I changed the code. Thanks for resolving it.

guardrex:
I'm working down all of the topics now with general updates on https://github.com/dotnet/AspNetCore.Docs/issues/19286. I'll catch other "My"-based naming here and change them all (along with many other improvements) when I reach this topic shortly.

PR #20905, opened 2020-12-09
First MVC - Set MovieDateRatingDA model as defined

### LINKED ISSUE #20665: First-MVC Series: New-Field cshtml example should use MovieDateRatingDAmult.cs model and included only when defined

wadepickett:
[Internal Review](https://review.docs.microsoft.com/en-us/aspnet/core/tutorials/first-mvc-app/new-field?view=aspnetcore-5.0&branch=pr-en-us-20905&tabs=visual-studio#add-a-rating-property-to-the-movie-model)

Fixes #20665

Summary of what changed:

- Models\MovieDateRatingDA was set as defined so it compiles by default along with View\Movies\IndexGenreRating
- Models\Movie received a #define that is commented off so it is not compiled by default.
- Updated referring .md tutorial files snippet regions and line numbers adjusted to match.


aspnetcore/tutorials/first-mvc-app/start-mvc/sample/MvcMovie3/Views/Movies/IndexGenreRating.cshtml
```diff
@@ -1,4 +1,5 @@
-﻿@model MvcMovie.Models.MovieGenreViewModel
+﻿#if COMBINED
```
Rick-Anderson, line 1: Why are you adding this?

wadepickett: I added it to make it clear that it is only used if its matching model using `COMBINED` is currently defined.  Otherwise it is hard to see what models are intended to work with what views.    Technically it is not needed, but it was intended to add clarity since we have several in there.   If you think it adds confusion rather than helping I'll take it out.

Rick-Anderson: Remove it.

Rick-Anderson:
@wadepickett  I think you need to run migrations to add the new model. Also test the download, I think you'll need to update the layout to bring in the the CSS to render.

Rick-Anderson:
@wadepickett  how is the coming?

wadepickett:
I moved on to another topic until you get back from vacation.   The way the files are organized in this sample do not appear to all be on the same plan.  I ended up needing to change several to make it all work out and need to meet with you briefly to walk through it.  Looks like you are back from vacation?  If so I'll send out an invite.  Thanks.

Rick-Anderson:
> I think you need to run migrations to add the new model. Also test the download, I think you'll need to update the layout to bring in the the CSS to render.

What more needs to be done than that?

You could just start working on updating the entire series like you did with RP.

Rick-Anderson:
@wadepickett  recommend closing this and just updating the entire series.

wadepickett:
@Rick-Anderson (Back from vacation as of today, I had to take much of November and December off to use up required vacation hours).

I'm fine with closing and fixing this issue as part of the larger collection of issues in the series.  I do need a way to break the work up a bit for the whole series. It gets a bit crazy for me as I to get through review cycles in one large PR for all tutorials.  

**Question:** Is it OK if I do one PR for the 5.0 sample and one for each Chapter? After all are approved they  they go live at same time.  That would make it muuuuch more approachable for me.

For sample structure.   When we met last you indicated the 5.0 sample should reflect the finished app achieved after all tutorials in the series have been completed.  That sounds good to me, going with that.  The pervious 3.1 version didn't follow that.  3.1 contained some of the steps, not all (which is why the issue tied to this PR came up), and 2.2 reflected the completed series.

For the completed 5.x sample, unlike the previous version, in between states of code will use use an added naming convention that indicates that indicates that in between step.    eg.  Views\Movies\Index.cshtml  reflects the series completed sample.     Views\Movies\IndexGenreRating.cshtml reflects a previous state for a particular step.tutorial.  Same for Models\SeedData.cs and Models\Movie.cs etc.      

**Question:** Does that structure for the 5.0 sample and its in-between states sound good?

Rick-Anderson:
The 3.x version didn't  have a complete code version.  But you should follow how MD files are written when you update. It makes each update much faster.

The main thing to note, if any file is going to be update in future versions, you need to make a copy of that file. That's done for you now, just follow the naming conventions.

> Question: Is it OK if I do one PR for the 5.0 sample and one for each Chapter? 

How about 2 per PR?

Rick-Anderson:
Closing as this will be done with the total clean-up work. When can I expect the first PR?

Issue #19113, opened 2020-07-06
Add section for loading PEM keys


javiercn:
[EDIT by guardrex to add the metadata]

In 5.0 we added support for loading certificates in PEM format and keys in PEM format. We added a new property `KeyPath` in the certificate config to indicate the path to a PEM key file in PKCS8 format (the PKCS8 format bit is important).

Essentially you can pass a certificate and a key separately and the key can be optionally protected with a password specified in the existing password. So for example

```
{
  "Path": "cert.pem",
  "KeyPath": "cert.key"
}
```

```
{
  "Path": "cert.pem",
  "KeyPath": "cert.key",
  "Password": "<<some-password>>"
}
```

```
{
  "Path": "cert.crt",
  "KeyPath": "cert.key",
  "Password": "<<some-password>>"
}
```

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: f22dcddb-0c25-be1a-9a02-9e71dce86bdb
* Version Independent ID: 3e0e12c2-e544-5655-e001-d2ec180b11a8
* Content: [Kestrel web server implementation in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel?view=aspnetcore-3.1)
* Content Source: [aspnetcore/fundamentals/servers/kestrel.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/fundamentals/servers/kestrel.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-fundamentals**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

PR #20913, opened 2020-12-10
Add PEM config info to Kestrel doc

JamesNK:
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/19113
Fixes https://github.com/dotnet/aspnetcore/issues/26240

### DIFF
--- a/aspnetcore/fundamentals/servers/kestrel.md
+++ b/aspnetcore/fundamentals/servers/kestrel.md
@@ -546,7 +546,8 @@ In the following *appsettings.json* example:
       "Https": {
         "Url": "https://*:5004",
         "Certificate": {
-          "Path": "<path to .pfx file>",
+          "Path": "<path to .pem file>",
+          "KeyPath": "<path to .key file>",
           "Password": "<certificate password>"
         }
       }
@@ -561,7 +562,24 @@ In the following *appsettings.json* example:
 }
 ```
 
-An alternative to using **Path** and **Password** for any certificate node is to specify the certificate using certificate store fields. For example, the **Certificates** > **Default** certificate can be specified as:
+Schema notes:
+
+* Endpoints names are case-insensitive. For example, `HTTPS` and `Https` are valid.
+* The `Url` parameter is required for each endpoint. The format for this parameter is the same as the top-level `Urls` configuration parameter except that it's limited to a single value.
+* These endpoints replace those defined in the top-level `Urls` configuration rather than adding to them. Endpoints defined in code via `Listen` are cumulative with the endpoints defined in the configuration section.
+* The `Certificate` section is optional. If the `Certificate` section isn't specified, the defaults defined in earlier scenarios are used. If no defaults are available, the server throws an exception and fails to start.
+* The `Certificate` section supports **Path**&ndash;**Password**, **Path**&ndash;**KeyPath**&dash;**Password**, and **Subject**&ndash;**Store** certificates.
+* Any number of endpoints may be defined in this way so long as they don't cause port conflicts.
+
+##### Certificate sources
+
+Certificate nodes can be configured to load certificates from a number of sources:
+
+* **Path** and **Password** to load *.pfx* files.
+* **Path**, **KeyPath** and **Password** to load *.pem* and *.key* files.
+* **Subject** and **Store** to load from the certificate store.
+
+For example, the **Certificates** > **Default** certificate can be specified as:
 
 ```json
 "Default": {
@@ -572,15 +590,9 @@ An alternative to using **Path** and **Password** for any certificate node is to
 }
 ```
 
-Schema notes:
+##### ConfigurationLoader
 
-* Endpoints names are case-insensitive. For example, `HTTPS` and `Https` are valid.
-* The `Url` parameter is required for each endpoint. The format for this parameter is the same as the top-level `Urls` configuration parameter except that it's limited to a single value.
-* These endpoints replace those defined in the top-level `Urls` configuration rather than adding to them. Endpoints defined in code via `Listen` are cumulative with the endpoints defined in the configuration section.
-* The `Certificate` section is optional. If the `Certificate` section isn't specified, the defaults defined in earlier scenarios are used. If no defaults are available, the server throws an exception and fails to start.
-* The `Certificate` section supports both **Path**&ndash;**Password** and **Subject**&ndash;**Store** certificates.
-* Any number of endpoints may be defined in this way so long as they don't cause port conflicts.
-* `options.Configure(context.Configuration.GetSection("{SECTION}"))` returns a `KestrelConfigurationLoader` with an `.Endpoint(string name, listenOptions => { })` method that can be used to supplement a configured endpoint's settings:
+`options.Configure(context.Configuration.GetSection("{SECTION}"))` returns a `KestrelConfigurationLoader` with an `.Endpoint(string name, listenOptions => { })` method that can be used to supplement a configured endpoint's settings:
 
 ```csharp
 webBuilder.UseKestrel((context, serverOptions) =>
--- a/aspnetcore/fundamentals/servers/kestrel.md
+++ b/aspnetcore/fundamentals/servers/kestrel.md
@@ -546,7 +546,7 @@ In the following *appsettings.json* example:
       "Https": {
         "Url": "https://*:5004",
         "Certificate": {
-          "Path": "<path to .pem file>",
+          "Path": "<path to .pem or .crt file>",
           "KeyPath": "<path to .key file>",
           "Password": "<certificate password>"
         }
@@ -579,7 +579,7 @@ Schema notes:
 Certificate nodes can be configured to load certificates from a number of sources:
 
 * **Path** and **Password** to load *.pfx* files.
-* **Path**, **KeyPath** and **Password** to load *.pem* and *.key* files.
+* **Path**, **KeyPath** and **Password** to load *.pem*/*.crt* and *.key* files.
 * **Subject** and **Store** to load from the certificate store.
 
 For example, the **Certificates** > **Default** certificate can be specified as:
--- a/aspnetcore/fundamentals/servers/kestrel.md
+++ b/aspnetcore/fundamentals/servers/kestrel.md
@@ -568,7 +568,10 @@ Schema notes:
 * The `Url` parameter is required for each endpoint. The format for this parameter is the same as the top-level `Urls` configuration parameter except that it's limited to a single value.
 * These endpoints replace those defined in the top-level `Urls` configuration rather than adding to them. Endpoints defined in code via `Listen` are cumulative with the endpoints defined in the configuration section.
 * The `Certificate` section is optional. If the `Certificate` section isn't specified, the defaults defined in earlier scenarios are used. If no defaults are available, the server throws an exception and fails to start.
-* The `Certificate` section supports **Path**&ndash;**Password**, **Path**&ndash;**KeyPath**&dash;**Password**, and **Subject**&ndash;**Store** certificates.
+* The `Certificate` section supports multiple certificate sources:
+  * **Path**&ndash;**Password**
+  * **Path**&ndash;**KeyPath**&dash;**Password**
+  * **Subject**&ndash;**Store**
 * Any number of endpoints may be defined in this way so long as they don't cause port conflicts.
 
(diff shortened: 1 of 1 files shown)


JamesNK:
My change currently adds PEM to just .NET Core 3 docs. Should be changed to just .NET 5 docs. I don't know the right way to do that in this file. There seems like A LOT of duplication. I think this file needs to be refactored to support versioning without some much duplication.

halter73:
Review link: https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel?view=aspnetcore-2.2&branch=pr-en-us-20913

Does something like this fix the moniker ranges for now? https://gist.github.com/halter73/464ad810feab0c3857e1b49252fdb8b9 I'm not sure how to test the output without submitting a PR.

I completely agree that kestrel.md is in severe need of refactoring. I hope to fix this properly and remove most of the unnecessary duplication when I update the docs to include other .NET 5 Kestrel config changes like SNI. If you want to do some deduping this as part of this PR, I'd be super thankful though.

guardrex:
FYI on the duplication: **_It's by-design._** We call it "whole-topic versioning." It was required back in the 2.x days, when there were three supported versions of Kestrel content under wildly differing framework feature sets. The versioning tags (moniker ranges) were being flipped into and out of every couple of lines or paragraphs ... it was 😵 **_insanely hard to maintain_** 😵. We had to do it back then. However, that was *then* ... and this is *now* ... so hopefully you'll be successful. I'm just letting you know that it had to be that way back then. We had no choice on it ... the doc was a 👹 **_BeAsT_** 👹 that ate authors for lunch! 😨

JamesNK:
What should happen here to support these .NET 5 specific properties? More duplicate?

guardrex:
Yes ... an entire duplicate of the doc (the 3.x version) becomes the new 5.x doc/version with added 5.x sample(s), if needed. This new moniker-range versioned content goes at the top above the existing 3.x content.

At some point (perhaps now), the 2.x block of content and any 2.x samples that exist get the *chop* 🔪. There are **_two_** maintenance benefits of *whole topic versioning* ... first was what I was saying about not flipping in and out of versions ... this is the second benefit ... we can just chop all of a version's content in a few quick edits.

You see the *major downside* ... multiple updates to content are required when something changes across versions.

In the future if we get *version by file*, we'd use the same process as we use for *whole topic versioning*, except the 3.x content would be duplicated to a **_new 5.x file_**, then the new file would be updated. Scott said he'd look further into *version by file* for the repo in 2021.

I hope we get *version by file* because we've been hitting section header build warnings and cross-linking problems with our *whole topic versioning* approach. The build system flakes out over duplicate headers in a markdown file, and cross-links to content (bookmarks) can break. *Whole topic versioning* isn't what doc system managers and engineers envisioned or currently support for moniker range versioning. I've avoided problems in two Blazor docs by using explicit `id`s on explicit heading tags. For an example, see the headers in ...

https://github.com/dotnet/AspNetCore.Docs/edit/master/aspnetcore/blazor/blazor-server-ef-core.md

... compare **Line 23** to **Line 154**.

I don't care for having to do it, but it works. No build warnings that way. I avoid broken bookmarks by not linking directly to sections. In a few spots where needed, I tell the reader, '... see the \*{SECTION HEADING TITLE}\* of the \<xref:{UID}> article.'

IIRC, @Rick-Anderson has another approach. Anyway ... with *version by file*, there would be no dup heading problem. It's designed to solve our *flipping in and out of versions* and our *:hocho: content when a version drops* scenarios.

I'll cc: @scottaddie here to correct any mistakes in my remarks and/or advise further.

Rick-Anderson:
> What should happen here to support these .NET 5 specific properties? More duplicate?

I prefer to not duplicate ***IF you can keep the moniker pairs to 6 or under***. That's not a hard number either. Once you get lots of moniker pairs, it becomes too fragile and error prone. At that point, resort to our old duplicate versioning:

```
// copy of 3.x version, updated to 5.x
::: moniker-end

// 3.1 version
::: moniker-end
```

cc @wadepickett  @scottaddie

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -618,7 +633,7 @@ webBuilder.ConfigureKestrel(serverOptions =>
 });
 ```
 
-*Kestrel support for SNI*
+#### Kestrel support for SNI
 
 [Server Name Indication (SNI)](https://tools.ietf.org/html/rfc6066#section-3) can be used to host multiple domains on the same IP address and port. For SNI to function, the client sends the host name for the secure session to the server during the TLS handshake so that the server can provide the correct certificate. The client uses the furnished certificate for encrypted communication with the server during the secure session that follows the TLS handshake.
```
blowdart, line 638: Provided rather than furnished? Easier to understand

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -561,7 +562,27 @@ In the following *appsettings.json* example:
...
+
+* Endpoints names are case-insensitive. For example, `HTTPS` and `Https` are valid.
+* The `Url` parameter is required for each endpoint. The format for this parameter is the same as the top-level `Urls` configuration parameter except that it's limited to a single value.
+* These endpoints replace those defined in the top-level `Urls` configuration rather than adding to them. Endpoints defined in code via `Listen` are cumulative with the endpoints defined in the configuration section.
+* The `Certificate` section is optional. If the `Certificate` section isn't specified, the defaults defined in earlier scenarios are used. If no defaults are available, the server throws an exception and fails to start.
+* The `Certificate` section supports multiple certificate sources:
+  * **Path**&ndash;**Password**
+  * **Path**&ndash;**KeyPath**&dash;**Password**
+  * **Subject**&ndash;**Store**
+* Any number of endpoints may be defined in this way so long as they don't cause port conflicts.
```
blowdart, line 575: port conflicts needs explaining.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -561,7 +562,27 @@ In the following *appsettings.json* example:
 }
 ```
 
-An alternative to using **Path** and **Password** for any certificate node is to specify the certificate using certificate store fields. For example, the **Certificates** > **Default** certificate can be specified as:
+Schema notes:
+
+* Endpoints names are case-insensitive. For example, `HTTPS` and `Https` are valid.
```
blowdart, line 569: Needs an example. I know what it means, but would all readers?

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -561,7 +562,27 @@ In the following *appsettings.json* example:
 }
 ```
 
-An alternative to using **Path** and **Password** for any certificate node is to specify the certificate using certificate store fields. For example, the **Certificates** > **Default** certificate can be specified as:
+Schema notes:
+
+* Endpoints names are case-insensitive. For example, `HTTPS` and `Https` are valid.
```
blowdart, line 568: Can you link back to the Urls parameter so people can easily see the format? If not, duplicate the format here.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -561,7 +562,27 @@ In the following *appsettings.json* example:
...
+* These endpoints replace those defined in the top-level `Urls` configuration rather than adding to them. Endpoints defined in code via `Listen` are cumulative with the endpoints defined in the configuration section.
+* The `Certificate` section is optional. If the `Certificate` section isn't specified, the defaults defined in earlier scenarios are used. If no defaults are available, the server throws an exception and fails to start.
```
blowdart, line 570: What earlier scenarios? Specify them or link to them. Also what exception?

blowdart (changes requested):
Some small nit picks.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -561,7 +562,27 @@ In the following *appsettings.json* example:
 }
 ```
 
-An alternative to using **Path** and **Password** for any certificate node is to specify the certificate using certificate store fields. For example, the **Certificates** > **Default** certificate can be specified as:
+Schema notes:
+
+* Endpoints names are case-insensitive. For example, `HTTPS` and `Https` are valid.
```
Rick-Anderson, line 567: HTTPS and http are valid too.

```suggestion
* Endpoints names are case-insensitive. For example, `HTTPS` and `Https` are equivalent.
```

aspnetcore/fundamentals/servers/kestrel.md
Rick-Anderson: Just checking, is it literally `Urls` or is it the plural of `Url`? This needs to be a literal and you can't make literals plural by tacking on a `s`

aspnetcore/fundamentals/servers/kestrel.md
Rick-Anderson: What is an earlier scenario? The style guide prefers previous, but you need to be specific, such as

 If the `Certificate` section isn't specified, the defaults, defined previously, are used.

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -572,15 +593,9 @@ An alternative to using **Path** and **Password** for any certificate node is to
...
-* These endpoints replace those defined in the top-level `Urls` configuration rather than adding to them. Endpoints defined in code via `Listen` are cumulative with the endpoints defined in the configuration section.
-* The `Certificate` section is optional. If the `Certificate` section isn't specified, the defaults defined in earlier scenarios are used. If no defaults are available, the server throws an exception and fails to start.
-* The `Certificate` section supports both **Path**&ndash;**Password** and **Subject**&ndash;**Store** certificates.
-* Any number of endpoints may be defined in this way so long as they don't cause port conflicts.
-* `options.Configure(context.Configuration.GetSection("{SECTION}"))` returns a `KestrelConfigurationLoader` with an `.Endpoint(string name, listenOptions => { })` method that can be used to supplement a configured endpoint's settings:
+`options.Configure(context.Configuration.GetSection("{SECTION}"))` returns a `KestrelConfigurationLoader` with an `.Endpoint(string name, listenOptions => { })` method that can be used to supplement a configured endpoint's settings:
```
Rick-Anderson, line 598:
```suggestion
`options.Configure(context.Configuration.GetSection("{SECTION}"))` returns a <xref:Microsoft.AspNetCore.Server.Kestrel.KestrelConfigurationLoader> with an `.Endpoint(string name, listenOptions => { })` method that can be used to supplement a configured endpoint's settings:
```

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -1570,9 +1585,32 @@ Schema notes:
...
+  "Subject": "<subject; required>",
+  "Store": "<cert store; required>",
+  "Location": "<location; defaults to CurrentUser>",
+  "AllowInvalid": "<true or false; defaults to false>"
+}
+```
+
+##### ConfigurationLoader
+
+`options.Configure(context.Configuration.GetSection("{SECTION}"))` returns a `KestrelConfigurationLoader` with an `.Endpoint(string name, listenOptions => { })` method that can be used to supplement a configured endpoint's settings:
```
Rick-Anderson, line 1613:
```suggestion
`options.Configure(context.Configuration.GetSection("{SECTION}"))` returns a <xref:Microsoft.AspNetCore.Server.Kestrel.KestrelConfigurationLoader> with an `.Endpoint(string name, listenOptions => { })` method that can be used to supplement a configured endpoint's settings:
```

aspnetcore/fundamentals/servers/kestrel.md
```diff
@@ -2434,9 +2472,32 @@ Schema notes:
...
```
Rick-Anderson, line 2500:
```suggestion
`options.Configure(context.Configuration.GetSection("{SECTION}"))` returns a <xref:Microsoft.AspNetCore.Server.Kestrel.KestrelConfigurationLoader> with an `.Endpoint(string name, listenOptions => { })` method that can be used to supplement a configured endpoint's settings:
```

Rick-Anderson (approved):

JamesNK:
I'm going to wait until https://github.com/dotnet/AspNetCore.Docs/pull/21091 is merged and rebase before making anymore changes.

PR #21029, opened 2020-12-23
Clarify Virtualize component item source

### LINKED ISSUE #21026: Clarify why setting both Items and ItemsProvider is not allowed for Blazor Virtualize component
At the moment, the page calls out the following:

```Don't attempt to use an items provider and assign a collection to Items for the same Virtualize component.```

An obvious question which comes up is `why? what will happen?`. We should maybe turn the above sentence into a warning block and also add content regarding the behavior in this case (exception?).

---
#### Document Details

* ID: 0f4ab8b4-4d2e-9e1b-793e-dbb17f6d694d
* Version Independent ID: 26ed2425-63ec-e206-edcb-e51f17edb844
* Content: [ASP.NET Core Blazor component virtualization](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/virtualization?view=aspnetcore-5.0)
* Content Source: [aspnetcore/blazor/components/virtualization.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/components/virtualization.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #21026

We usually use warnings for security concerns, writing unstable code or code that produces strange/unexpected results, unsupported approaches that might break, and practices that might fall victim to a future breaking change.

This one seems less severe given that there's a clear exception when the params are set, but I place the guidance into its own paragraph to highlight it a bit more than just tacking it on at the end of the current paragraph.

cc: @mkArtakMSFT

### DIFF
--- a/aspnetcore/blazor/components/virtualization.md
+++ b/aspnetcore/blazor/components/virtualization.md
@@ -74,7 +74,7 @@ If you don't want to load all of the items into memory, you can specify an items
 
 The items provider receives an `ItemsProviderRequest`, which specifies the required number of items starting at a specific start index. The items provider then retrieves the requested items from a database or other service and returns them as an `ItemsProviderResult<TItem>` along with a count of the total items. The items provider can choose to retrieve the items with each request or cache them so that they're readily available.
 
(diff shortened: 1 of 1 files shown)


guardrex:
... and I just applied bold to "**one item source**" to make that stand out a bit more.

mkArtakMSFT:
Thanks @guardrex

PR #21033, opened 2020-12-23
Fixes #21032 - Updates outdated SignalR package reference in custom logging example

### LINKED ISSUE #21032: SignalR Custom Logging Example uses outdated package
I noticed the example for building a custom logging in TypeScript references the outdated SignalR NPM package.

```typescript
import { ILogger, LogLevel, HubConnectionBuilder } from "@aspnet/signalr";
```

I'd like to correct the example so it points the correct `@microsoft/signalr` package with a PR.

---
#### Document Details

* ID: cef32cc5-6a1f-25c9-0d89-a984a3f68851
* Version Independent ID: 3e302985-8504-d333-3222-ab8da8902da2
* Content: [Logging and diagnostics in ASP.NET Core SignalR](https://docs.microsoft.com/en-us/aspnet/core/signalr/diagnostics?view=aspnetcore-5.0#server-side-logging)
* Content Source: [aspnetcore/signalr/diagnostics.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/signalr/diagnostics.md)
* Product: **aspnet-core**
* GitHub Login: @anurse
* Microsoft Alias: **anurse**

1kevgriff:
Fixes #21032


guardrex:
@BrennanConroy ... If this is ok, shouldn't we version it? If so, which version(s) should show the change?

1kevgriff:
@guardrex I was thinking about that as well.  I believe the switch was after 3.0 - so the old example would be for 2.x and the new example would be 3.0 and greater.

guardrex:
We're starting to drop 2.x coverage these days; so if that's the case (>=3.1 ... note that 3.0 is already out of support as of last March), then the update on the PR is fine. We'll make one more change tho to set the versioning on the whole topic from 2.1 to 3.1. If that's what we'll do, then the change is from ...

```
monikerRange: '>= aspnetcore-2.1'
```

... to ...

```
```

... in the topic's metadata on Line 5.

Hold off for a sec tho until @BrennanConroy advises us on the versioning for this.

BrennanConroy:
> We're starting to drop 2.x coverage these days

2.1 is still supported, and on full framework will basically be supported forever. So not sure why we are dropping coverage there.

2.1 should reference `@aspnet/signalr` (although the client is backwards compatible so `@microsoft/signalr` will work with a 2.1 server)
And 3.1+ should reference `@microsoft/signalr`.

guardrex:
> So not sure why we are dropping coverage there.

For efficiency given that readers are mostly off of 2.x. If the update requires jumping through hoops (as it does here) to add or maintain the coverage, we've been dropping 2.x. This has been going on for a little while now. We're only strictly required to cover the current release via MS guidelines. In this case, we can keep this covered for 2.x. However ... indeed ... there be **_hoops_** here ...

@1kevgriff, you need to modify the PR to version the content, which unfortunately can take longer for us to explain than for one of us to just perform on a new PR. If you want that explanation, one of us can provide it. Alternatively, you'd close this PR and open an issue from the bottom of the associated topic. One of us would take the issue and work up the PR.

1kevgriff:
@guardrex I'm pretty sure I can get this versioned correctly.  Give me a few minutes and I'll update the PR for review.

1kevgriff:
@guardrex PR updated.  How's that?

guardrex (approved):
Great @1kevgriff ... thanks for taking care of it.

@BrennanConroy ... Swap Nurse out for someone else?

guardrex:
Thanks again @1kevgriff ... I'll merge to live now, and these updates should hit the live site within 15 minutes.

Issue #18198, opened 2020-05-08
[Blazor][Webassembly] CORS sample


javiercn:
This is an area of confusion for people, so it will be nice to have a sample we can point people to that showcases how to achieve this scenario.

This is a tracking issue for us to do the work here to provide such sample.

guardrex:
We sort of **_do_** have this. It's in the common sample app and there's a brief section. Perhaps, we'll need to expand the section. However, keep in mind that CORS is a very **_known_** thing with a set of standard behaviors that the cross-linked content in our ASP.NET Core CORS docs covers in detail.

https://docs.microsoft.com/en-us/aspnet/core/blazor/call-web-api?view=aspnetcore-3.1#cross-origin-resource-sharing-cors

javiercn:
There are a bunch fo Blazor specific things here, like configuring the HttpClient instance and is painful in general to setup CORS, the CORS doc describes several ways to do it and I think we should have one specific straightforward way to do it with Blazor.

I'll elaborate on this once I get time to look into the area.

PR #21043, opened 2020-12-25
Update Readme.MD to add correct startup file for CORS

corneliusroemer:
Helps with #18198

The sample app contains two files that contain a `Configure` method:
- `Startup.cs`
- `StartupJavaScript.cs`

Hitherto, the Readme said:
>Add the following CORS middleware configuration to the web API's `Startup.Configure` method:</p>

But that is incorrect. The configuration needs to be added to `StartupJavaScript.Configure`. It took me ages to figure it out, trusting the Readme to be correct - when it wasn't.

There's no harm in making this change. Yes, people can eventually try it out themselves, but there's no harm in adding the correct file reference.

### DIFF
--- a/aspnetcore/blazor/common/samples/3.x/BlazorWebAssemblySample/README.md
+++ b/aspnetcore/blazor/common/samples/3.x/BlazorWebAssemblySample/README.md
@@ -6,7 +6,7 @@ This sample illustrates the use of Blazor scenarios described in the Blazor docu
 
 The web API example requires a running web API based on the sample app for the <a href="https://docs.microsoft.com/aspnet/core/tutorials/first-web-api">Create a web API with ASP.NET Core</a> topic, which by default uses the same HTTPS port (5001) as the Blazor sample app. To use both apps on the same machine at the same time, change the port of the web API (for example, use port 10000). The sample app makes requests to the web API at `https://localhost:10000/api/TodoItems`. If a different web API address is used, update the `ServiceEndpoint` constant value in the Razor component's `@code` block.</p>
 
-The sample app makes a <a href="https://docs.microsoft.com/aspnet/core/security/cors">cross-origin resource sharing (CORS)</a> request from `http://localhost:5000` or `https://localhost:5001` to the web API. Credentials (authorization cookies/headers) are permitted. Add the following CORS middleware configuration to the web API's `Startup.Configure` method:</p>
+The sample app makes a <a href="https://docs.microsoft.com/aspnet/core/security/cors">cross-origin resource sharing (CORS)</a> request from `http://localhost:5000` or `https://localhost:5001` to the web API. Credentials (authorization cookies/headers) are permitted. Add the following CORS middleware configuration to the web API's `StartupJavaScript.Configure` method:</p>
 
 ```csharp
 app.UseCors(policy => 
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/AspNetCore.Docs?pullRequest=21043) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/AspNetCore.Docs?pullRequest=21043) before we can accept your contribution.<br/><br/>:x: corneliusroemer [sign now](https://cla.dotnetfoundation.org/dotnet/AspNetCore.Docs?pullRequest=21043)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/AspNetCore.Docs?pullRequest=21043) it.</sub>

corneliusroemer:
Two other things that could be added to the Readme to help smooth out cors problems for any newcomers are:
- `app.UseCors` needs to come before `app.UseEndpoints`
- For the syntax to work, one needs `using Microsoft.Net.Http.Headers;` otherwise `HeaderNames` is not recognized.

guardrex:
Hello @corneliusroemer ... When it says "to the web API's Startup.Configure method," it's referring to an ASP.NET Core app web API with CORS Middleware cross-linked to :point_right:

* https://docs.microsoft.com/aspnet/core/tutorials/first-web-api
* https://docs.microsoft.com/aspnet/core/security/cors

... not to the JS tutorial code at ...

https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/tutorials/first-web-api/samples/3.0/TodoApi/StartupJavaScript.cs

... which I'm not familiar with. We have multiple authors working on the doc set, and I don't get over to those regions very often. I'm not sure where that's linked into the docs. I thought that maybe the 5.x version dropped the `StartupJavaScript` in favor of just `Startup` ... or actually .... now that I think about it 🤔 ... `StartupJavaScript` is merely perhaps a SxS startup file showing a different startup config for a JS client **_in the same app_**, which would be required to compose a sample app with multiple startup classes for a tutorial. Anyway ... whatever the case with that tutorial/sample ... that's not really the normal layout of an ASP.NET Core app, which would just have a `Startup` class and file. Therefore, the README is correct in the general case, and we don't really want to make this change.

Location of CORS Middleware is called out in the CORS topic and further cross-links to the *Middleware Order* section of the *Middleware* topic at :point_right: https://docs.microsoft.com/aspnet/core/fundamentals/middleware/#middleware-order

Good call on `HeaderNames`, so I'll add that to my tracking issue for doc updates and include it when I reach this topic ... hopefully soon.

guardrex:
Confirmed ... They dropped that SxS double-startup file approach for 5.0 and later ...

https://github.com/dotnet/AspNetCore.Docs/tree/master/aspnetcore/tutorials/first-web-api/samples/5.x/TodoApi

WRT their use of the two startup files for 3.x, I suppose we can see what Javier wants to do with this in 2021. He might decide to drop the whole approach of using their web API example. He might say ... as he hinted in the issue ... that a simpler example is what's needed here. We could end up going with a hosted Blazor sample app that uses CORS in the `Server` app's web API or WASM SxS with an MVC web API app + CORS ... all kept over here in Blazor sample code. I'm not sure which way he'll want to go with it. I think what happened is that a lot of devs opened up a lot of CORS-Blazor issues for the product unit troubling them to some degree with support requests. We'll likely discuss it further on that issue that he opened. Thanks again for commenting on it. Hopefully, we'll address the whole affair in 2021.

PR #21069, opened 2020-12-29
Update IE support for Core SignalR

### LINKED ISSUE #21068: ASP.NET Core SignalR client support for Internet Explorer unclear
The "Differences between SignalR and ASP.NET Core SignalR" page under "Internet Explorer support" states that IE 11 is supported.

However, the "suppported platforms" link goes to https://docs.microsoft.com/en-us/aspnet/core/signalr/supported-platforms?view=aspnetcore-3.1#javascript-client which doesn't mention Internet Explorer at all - which implies no versions of Internet Explorer are supported.

Which interpretation is right?

---
#### Document Details

* ID: 04f4bdfd-dfac-d6ae-4bd9-57439d300a5d
* Version Independent ID: 1f37f1ba-3e43-e726-f231-52acc6d26516
* Content: [Differences between SignalR and ASP.NET Core SignalR](https://docs.microsoft.com/en-us/aspnet/core/signalr/version-differences?preserve-view=true&view=aspnetcore-3.1)
* Content Source: [aspnetcore/signalr/version-differences.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/signalr/version-differences.md)
* Product: **aspnet-core**
* GitHub Login: @bradygaster
* Microsoft Alias: **bradyg**

guardrex:
Fixes #21068

Thanks @chrisseed! :rocket:

Brennan ... I don't version this because we've always treated browser support as **_current only_** (unversioned in the *Supported Platforms* doc).

### DIFF
--- a/aspnetcore/signalr/version-differences.md
+++ b/aspnetcore/signalr/version-differences.md
@@ -188,7 +188,7 @@ The dependency on jQuery has been removed, however projects can still use jQuery
 ### Internet Explorer support
 
 ASP.NET Core SignalR doesn't support Microsoft Internet Explorer, whereas ASP.NET SignalR supports Microsoft Internet Explorer 8 or later.
-More info on browser support can be found in <xref:signalr/supported-platforms#javascript-client>.
+For more information, see <xref:signalr/supported-platforms#javascript-client>.
 
 ### JavaScript client method syntax
 
--- a/aspnetcore/signalr/version-differences.md
+++ b/aspnetcore/signalr/version-differences.md
@@ -187,8 +187,8 @@ The dependency on jQuery has been removed, however projects can still use jQuery
 
 ### Internet Explorer support
 
-ASP.NET Core SignalR supports Microsoft Internet Explorer 11 or later, whereas ASP.NET SignalR supports Microsoft Internet Explorer 8 or later.
-More info on browser support can be found at [supported platforms](xref:signalr/supported-platforms#javascript-client).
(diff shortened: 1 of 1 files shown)


chrisseed:
That was impressively quick @guardrex !

guardrex:
The *rotten family 😼* wakes me up daily at 4am! ..... to be fed or let out or whatever. Even on ⛄ vacation 🎁 like this, there's nothing else to do in these early AM hours.

aspnetcore/signalr/version-differences.md
```diff
@@ -187,8 +187,8 @@ The dependency on jQuery has been removed, however projects can still use jQuery
 
 ### Internet Explorer support
 
-ASP.NET Core SignalR supports Microsoft Internet Explorer 11 or later, whereas ASP.NET SignalR supports Microsoft Internet Explorer 8 or later.
-More info on browser support can be found at [supported platforms](xref:signalr/supported-platforms#javascript-client).
+ASP.NET Core SignalR doesn't support Microsoft Internet Explorer, whereas ASP.NET SignalR supports Microsoft Internet Explorer 8 or later.
+For more information, see <xref:signalr/supported-platforms#javascript-client>.
```
halter73, line 191: Should we update this doc to explicitly call out that IE is unsupported?

halter73 (approved):

aspnetcore/signalr/version-differences.md
guardrex: That's what it says :point_up:, so you mean in some other way. How do you mean?

PR #21085, opened 2020-12-31
updated scope to match instructions

shehabafawzy:
Added new [] to the scope to match the instructions.


guardrex:
The braces of the placeholder aren't C# braces, and a C# string array using the `new` keyword is `new[] { ... }`, so the current instruction is accurate. However, that reasoning breaks in the next example for the extension method ...

```csharp
IServiceCollection AddGraphClient(IServiceCollection, string[])
```

I'll make a few edits to fix it up. Thanks for calling this out.

guardrex (approved):
Ok ... that seems clear and consistent across those two scope array configurations. Thanks again.

PR #21090, opened 2020-12-31
 Add docs for MessagePack

wtgodbe:
[Internal Review](https://review.docs.microsoft.com/en-us/aspnet/core/signalr/messagepackhubprotocol?view=aspnetcore-5.0&branch=pr-en-us-21090#java-client)

Resolves https://github.com/dotnet/aspnetcore/issues/26248

### DIFF
--- a/aspnetcore/signalr/java-client.md
+++ b/aspnetcore/signalr/java-client.md
@@ -93,9 +93,16 @@ HubConnection hubConnection = HubConnectionBuilder.create("YOUR HUB URL HERE")
 
 ## Known limitations
 
-::: moniker range=">= aspnetcore-3.0"
+::: moniker range=">= aspnetcore-5.0"
 
 * Transport fallback and the Server Sent Events transport aren't supported.
+
+::: moniker-end
+
+::: moniker range=">= aspnetcore-3.0" < aspnetcore-5.0"
+
+* Transport fallback and the Server Sent Events transport aren't supported.
+* Only the JSON protocol is supported.
 
 ::: moniker-end
 
--- a/aspnetcore/signalr/java-client.md
+++ b/aspnetcore/signalr/java-client.md
@@ -91,6 +91,32 @@ HubConnection hubConnection = HubConnectionBuilder.create("YOUR HUB URL HERE")
     })).build();
 ```
 
+::: moniker range=">= aspnetcore-5.0"
+
+### Passing Class information in Java
+
+When calling the `on`, `invoke`, or `stream` methods of `HubConnection` in the Java client, users should pass a `Type` object rather than a `Class<?>` object to describe each `Object` passed to the method. A `Type` can be acquired using the provided `TypeReference` class. For example, using a custom classes named `Foo` and `Bar`, the following code gets each `Type`:
+
+```java
+Type fooType = new TypeReference<Foo>() { }).getType();
+Type barType = new TypeReference<Bar>() { }).getType();
+```
+
+When calling one of these methods with one or more object types, use the generics syntax when invoking the method. For example, when registering an `on` handler for a method named `func`, which takes as arguments a `Foo` object and a `Bar` object, use the following code to set an action to print the two objects:
+
+```java
+hubConnection.<Foo, Bar>on("func", (param1, param2) ->{
+    System.out.println(param1);
+    System.out.println(param2);
+}, fooType, barType);
+```
+
+The reason that this convention is necessary is, due to type erasure in Java, we can not retrieve complete information about complex types with the `Object.getClass` method. For example, calling `getClass` on an `ArrayList<String>` would not return `Class<ArrayList<String>>`, but rather `Class<ArrayList>`, which does not give the deserializer enough information to correctly deserialize an incoming message. The same is true for custom objects.
+
+The exception to this rule is primitives such as `int`. When using `int`, use the built-in `int.class`.
+
+::: moniker-end
+
 ## Known limitations
 
 ::: moniker range=">= aspnetcore-5.0"
--- a/aspnetcore/signalr/java-client.md
+++ b/aspnetcore/signalr/java-client.md
@@ -124,7 +124,7 @@ This convention is necessary because we can not retrieve complete information ab
 
 ::: moniker-end
 
-::: moniker range=">= aspnetcore-3.0" < aspnetcore-5.0"
+::: moniker range=">= aspnetcore-3.0 < aspnetcore-5.0"
 
 * Transport fallback and the Server Sent Events transport aren't supported.
 * Only the JSON protocol is supported.
--- a/aspnetcore/signalr/java-client.md
+++ b/aspnetcore/signalr/java-client.md
@@ -111,7 +111,7 @@ hubConnection.<Foo, Bar>on("func", (param1, param2) ->{
 }, fooType, barType);
 ```
 
-The reason that this convention is necessary is, due to type erasure in Java, we can not retrieve complete information about complex types with the `Object.getClass` method. For example, calling `getClass` on an `ArrayList<String>` would not return `Class<ArrayList<String>>`, but rather `Class<ArrayList>`, which does not give the deserializer enough information to correctly deserialize an incoming message. The same is true for custom objects.
+This convention is necessary because we can not retrieve complete information about complex types with the `Object.getClass` method due to type erasure in Java. For example, calling `getClass` on an `ArrayList<String>` would not return `Class<ArrayList<String>>`, but rather `Class<ArrayList>`, which does not give the deserializer enough information to correctly deserialize an incoming message. The same is true for custom objects.
 
 The exception to this rule is primitives such as `int`. When using `int`, use the built-in `int.class`.
 
--- a/aspnetcore/signalr/java-client.md
+++ b/aspnetcore/signalr/java-client.md
@@ -95,7 +95,6 @@ HubConnection hubConnection = HubConnectionBuilder.create("YOUR HUB URL HERE")
 
 ::: moniker range=">= aspnetcore-3.0"
 
-* Only the JSON protocol is supported.
 * Transport fallback and the Server Sent Events transport aren't supported.
 
 ::: moniker-end
--- a/aspnetcore/signalr/java-client.md
+++ b/aspnetcore/signalr/java-client.md
@@ -95,25 +95,24 @@ HubConnection hubConnection = HubConnectionBuilder.create("YOUR HUB URL HERE")
 
 ### Passing Class information in Java
 
-When calling the `on`, `invoke`, or `stream` methods of `HubConnection` in the Java client, users should pass a `Type` object rather than a `Class<?>` object to describe each `Object` passed to the method. A `Type` can be acquired using the provided `TypeReference` class. For example, using a custom classes named `Foo` and `Bar`, the following code gets each `Type`:
+When calling the `on`, `invoke`, or `stream` methods of `HubConnection` in the Java client, users should pass a `Type` object rather than a `Class<?>` object to describe any generic `Object` passed to the method. A `Type` can be acquired using the provided `TypeReference` class. For example, using a custom generic class named `Foo<T>`, the following code gets the `Type`:
 
 ```java
-Type fooType = new TypeReference<Foo>() { }).getType();
-Type barType = new TypeReference<Bar>() { }).getType();
+Type fooType = new TypeReference<Foo<String>>() { }).getType();
 ```
 
-When calling one of these methods with one or more object types, use the generics syntax when invoking the method. For example, when registering an `on` handler for a method named `func`, which takes as arguments a `Foo` object and a `Bar` object, use the following code to set an action to print the two objects:
+For non-generics, such as primitives or other non-parameterized types like `String`, you can simply use the built-in `.class`.
+
+When calling one of these methods with one or more object types, use the generics syntax when invoking the method. For example, when registering an `on` handler for a method named `func`, which takes as arguments a String and a `Foo<String>` object, use the following code to set an action to print the arguments:
 
 ```java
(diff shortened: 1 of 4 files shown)


aspnetcore/signalr/messagepackhubprotocol.md
```diff
@@ -104,6 +104,26 @@ const connection = new signalR.HubConnectionBuilder()
 > [!NOTE]
 > At this time, there are no configuration options for the MessagePack protocol on the JavaScript client.
 
+### Java client
```
wtgodbe, line 107: Does anyone know why this doc repeats itself 3 times?

aspnetcore/signalr/messagepackhubprotocol.md
```diff
@@ -104,6 +104,26 @@ const connection = new signalR.HubConnectionBuilder()
...
+
+To enable MessagePack in the Java, first install the `com.microsoft.signalr.messagepack` package. If using Gradle, add the following line to the `dependencies` section of your *build.gradle* file:
+
+```gradle
+implementation 'com.microsoft.signalr.messagepack:signalr-messagepack:5.0.0'
+```
+
+If using Maven, add the following lines inside the `<dependencies>` element of your *pom.xml* file:
+
+[!code-xml[pom.xml dependency element messagePack](java-client/sample/pom.xml?name=snippet_dependencyElement_messagePack)]
```
wtgodbe, line 117: This didn't work, anybody know what the right syntax is?

aspnetcore/signalr/messagepackhubprotocol.md
BrennanConroy: Yeah, different versions. Each one is between a `::: moniker range=">= aspnetcore-5.0"` and `::: moniker-end` and has different content based on which version you're referencing

wtgodbe: Ah, didn't notice that. I'll remove the new stuff from the pre-5.0 sections

aspnetcore/signalr/java-client.md
```diff
@@ -95,7 +95,6 @@ HubConnection hubConnection = HubConnectionBuilder.create("YOUR HUB URL HERE")
 
 ::: moniker range=">= aspnetcore-3.0"
 
-* Only the JSON protocol is supported.
```
BrennanConroy, line 98: Should add a 5.0 moniker and remove it there and keep it for 3.1

aspnetcore/signalr/messagepackhubprotocol.md
```diff
@@ -172,6 +192,30 @@ InvalidDataException: Error binding arguments. Make sure that the types of the p
 
 For more information on this limitation, see GitHub issue [aspnet/SignalR#2937](https://github.com/aspnet/SignalR/issues/2937).
 
+### Passing Class information in Java
```
BrennanConroy, line 195: This isn't specific to MessagePack is it?

aspnetcore/signalr/messagepackhubprotocol.md
```diff
@@ -104,6 +104,26 @@ const connection = new signalR.HubConnectionBuilder()
 > [!NOTE]
 > At this time, there are no configuration options for the MessagePack protocol on the JavaScript client.
 
+### Java client
+
+To enable MessagePack in the Java, first install the `com.microsoft.signalr.messagepack` package. If using Gradle, add the following line to the `dependencies` section of your *build.gradle* file:
```
Rick-Anderson, line 109: Something like:

```suggestion
To enable MessagePack with Java, install the `com.microsoft.signalr.messagepack` package. When using Gradle, add the following line to the `dependencies` section of the *build.gradle* file:
```

aspnetcore/signalr/messagepackhubprotocol.md
```diff
@@ -104,6 +104,26 @@ const connection = new signalR.HubConnectionBuilder()
...
+
+```gradle
+implementation 'com.microsoft.signalr.messagepack:signalr-messagepack:5.0.0'
+```
+
+If using Maven, add the following lines inside the `<dependencies>` element of your *pom.xml* file:
```
Rick-Anderson, line 115:
```suggestion
When using Maven, add the following lines inside the `<dependencies>` element of the *pom.xml* file:
```

aspnetcore/signalr/messagepackhubprotocol.md
```diff
@@ -104,6 +104,26 @@ const connection = new signalR.HubConnectionBuilder()
...
+
+[!code-xml[pom.xml dependency element messagePack](java-client/sample/pom.xml?name=snippet_dependencyElement_messagePack)]
+
+Then call `withHubProtocol(new MessagePackHubProtocol())` on `HubConnectionBuilder`.
```
Rick-Anderson, line 119:
```suggestion
Call `withHubProtocol(new MessagePackHubProtocol())` on `HubConnectionBuilder`.
```

aspnetcore/signalr/messagepackhubprotocol.md
```diff
@@ -172,6 +192,30 @@ InvalidDataException: Error binding arguments. Make sure that the types of the p
 
 For more information on this limitation, see GitHub issue [aspnet/SignalR#2937](https://github.com/aspnet/SignalR/issues/2937).
 
+### Passing Class information in Java
+
+When calling the `on()`, `invoke()`, or `stream()` methods of `HubConnection` in the Java client, MessagePack users should pass a `Type` object rather than a `Class<?` object to describe the `Object`(s) they are passing to the method. A `Type` can be acquired using the provided `TypeReference` class. For example, if you have a custom class named `Foo` and another named `Bar`, you can get their `Type`s with the following code:
```
Rick-Anderson, line 197:
```suggestion
When calling the `on`, `invoke`, or `stream` methods of `HubConnection` in the Java client, MessagePack users should pass a `Type` object rather than a `Class<?` object to describe each `Object` passed to the method. A `Type` can be acquired using the provided `TypeReference` class. For example, using a custom classes named `Foo` and `Bar`, the following code gets each `Type`:
```

aspnetcore/signalr/messagepackhubprotocol.md
```diff
@@ -172,6 +192,30 @@ InvalidDataException: Error binding arguments. Make sure that the types of the p
...
+
+```java
+Type fooType = new TypeReference<Foo>() { }).getType();
+Type barType = new TypeReference<Bar>() { }).getType();
+```
+
+Furthermore, when calling one of these methods with one or more object types, you should use the generics syntax when invoking the method. For example, if registering an `on` handler for a method named `func`, which takes as arguments a `Foo` object and a `Bar` object, you could use the following code to set an action to print the two objects:
```
Rick-Anderson, line 204:
```suggestion
When calling one of these methods with one or more object types, use the generics syntax when invoking the method. For example, when registering an `on` handler for a method named `func`, which takes as arguments a `Foo` object and a `Bar` object, use the following code to set an action to print the two objects:
```

aspnetcore/signalr/messagepackhubprotocol.md
```diff
@@ -172,6 +192,30 @@ InvalidDataException: Error binding arguments. Make sure that the types of the p
...
+
+```java
+hubConnection.<Foo, Bar>on("func", (param1, param2) ->{
+    System.out.println(param1);
+    System.out.println(param2);
+}, fooType, barType);
+```
+
+The exception to this rule is primitives such as `int`, for which you should use the built-in `int.class`.
```
Rick-Anderson, line 213:
```suggestion
The exception to this rule is primitives such as `int`. When using `int`, use the built-in `int.class`.
```

aspnetcore/signalr/messagepackhubprotocol.md
```diff
@@ -172,6 +192,30 @@ InvalidDataException: Error binding arguments. Make sure that the types of the p
...
+
+### Chars and Strings in Java
+
+In the java client, `char` objects will be serialized as one-character `String`s. This is in contrast with the C# and JavaScript client, which serialize them as `short`s. The MessagePack spec itself does not define behavior for `char`s, so it is up to the library author to determine how to serialize them. The difference in behavior between our clients is a result of the libraries we used for our implementations.
```
Rick-Anderson, line 217: The following doesn't localize
```
`String`s
`short`s
`char`s
```

So rewrite as strings, or series of one character string types, or something like that.

Rick-Anderson (changes requested):

aspnetcore/signalr/messagepackhubprotocol.md
halter73: Looks like it worked to me: https://review.docs.microsoft.com/en-us/aspnet/core/signalr/messagepackhubprotocol?view=aspnetcore-5.0&branch=pr-en-us-21090#java-client

blakehavard (approved):

aspnetcore/signalr/messagepackhubprotocol.md
wtgodbe: No, but it's only required in MessagePack, JSON still works without it. I could move this to java-client.md and call out the differences there.

halter73: How does JSON manage to work without it given the type erasure? It's also probably worth pointing out type erasure is the reason a TypeReference is necessary.

wtgodbe: Oh, you're right, I was remembering wrong - JSON had the same issue with more complex types but we hadn't noticed before. I'll move this segment & expand it now.

wtgodbe: https://github.com/dotnet/AspNetCore.Docs/pull/21090/commits/fd51aba3760da7845281f792159238d121ee3741

aspnetcore/signalr/java-client.md
```diff
@@ -91,12 +91,44 @@ HubConnection hubConnection = HubConnectionBuilder.create("YOUR HUB URL HERE")
...
+::: moniker range=">= aspnetcore-5.0"
+
+### Passing Class information in Java
+
+When calling the `on`, `invoke`, or `stream` methods of `HubConnection` in the Java client, users should pass a `Type` object rather than a `Class<?>` object to describe each `Object` passed to the method. A `Type` can be acquired using the provided `TypeReference` class. For example, using a custom classes named `Foo` and `Bar`, the following code gets each `Type`:
+
+```java
+Type fooType = new TypeReference<Foo>() { }).getType();
+Type barType = new TypeReference<Bar>() { }).getType();
+```
```
halter73, line 103: This only necessary for generic types, right? I think `Bar.class` would work fine.

aspnetcore/signalr/java-client.md
```diff
@@ -91,12 +91,44 @@ HubConnection hubConnection = HubConnectionBuilder.create("YOUR HUB URL HERE")
...
+When calling one of these methods with one or more object types, use the generics syntax when invoking the method. For example, when registering an `on` handler for a method named `func`, which takes as arguments a `Foo` object and a `Bar` object, use the following code to set an action to print the two objects:
+
+```java
+hubConnection.<Foo, Bar>on("func", (param1, param2) ->{
+    System.out.println(param1);
+    System.out.println(param2);
+}, fooType, barType);
+```
+
+The reason that this convention is necessary is, due to type erasure in Java, we can not retrieve complete information about complex types with the `Object.getClass` method. For example, calling `getClass` on an `ArrayList<String>` would not return `Class<ArrayList<String>>`, but rather `Class<ArrayList>`, which does not give the deserializer enough information to correctly deserialize an incoming message. The same is true for custom objects.
```
halter73, line 114: Nit:

```suggestion
This convention is necessary because we can not retrieve complete information about complex types with the `Object.getClass` method due to type erasure in Java. For example, calling `getClass` on an `ArrayList<String>` would not return `Class<ArrayList<String>>`, but rather `Class<ArrayList>`, which does not give the deserializer enough information to correctly deserialize an incoming message. The same is true for custom objects.
```

aspnetcore/signalr/java-client.md
wtgodbe: Oops, yes, you're right

halter73 (approved):

halter73:
Once https://docs.microsoft.com/en-us/java/api/com.microsoft.signalr?view=aspnet-signalr-java is updated for 5.0, we should link to the TypeReference docs from here.

halter73:
@BrennanConroy @Rick-Anderson What do we have to do to update the SignalR Java client API docs for 5.0?

aspnetcore/signalr/java-client.md
```diff
@@ -91,12 +91,43 @@ HubConnection hubConnection = HubConnectionBuilder.create("YOUR HUB URL HERE")
...
 ## Known limitations
 
-::: moniker range=">= aspnetcore-3.0"
+::: moniker range=">= aspnetcore-5.0"
+
+* Transport fallback and the Server Sent Events transport aren't supported.
+
+::: moniker-end
+
+::: moniker range=">= aspnetcore-3.0" < aspnetcore-5.0"
```
wtgodbe, line 127: @Rick-Anderson the build is complaining about invalid characters here, did I do something wrong with the moniker sections?

halter73:
```suggestion
::: moniker range=">= aspnetcore-3.0 < aspnetcore-5.0"
```

scottaddie: There's an extra double quote after "3.0". Removing that should fix the error.

BrennanConroy (approved):

Rick-Anderson (approved):

Rick-Anderson:
> @BrennanConroy @Rick-Anderson What do we have to do to update the SignalR Java client API docs for 5.0?

@scottaddie ?

PR #21104, opened 2021-01-04
Fix typo in simple.md

ericmutta:
The following sentence:

> **Moved** shared content into one or more partial views...

Should read:

> **Move** shared content into one or more partial views...


guardrex (approved):
Thanks @ericmutta! :rocket:

ericmutta:
Thanks @guardrex for following up! :rocket:

PR #21107, opened 2021-01-04
Fix Add URL example

davidhenley:
`Add URL` example was using `Add File` and is now correct


scottaddie (approved):

scottaddie:
Thank you for the fix, @davidhenley!

PR #21132, opened 2021-01-06
Update windowsauth.md

### LINKED ISSUE #21020: Impersonation: RunImpersonated should work for async workloads
The docs on impersonation currently say this:

> `RunImpersonated` doesn't support asynchronous operations and shouldn't be used for complex scenarios. For example, wrapping entire requests or middleware chains isn't supported or recommended.

However, that doesn’t seem to be the case, going by what happened in dotnet/runtime#24009. The async overloads there do nothing than forward `Func<Task<T1>>` as a `Func<T2>` to the `RunImpersonated` so that the return type is `Task<T1>` itself. This also seems to be supported by [the docs on `RunImpersonated`](https://docs.microsoft.com/en-us/dotnet/api/system.security.principal.windowsidentity.runimpersonated?view=net-5.0):

> This method may be used reliably with the async/await pattern, unlike `Impersonate`. In an async method, the generic overload of this method may be used with an async delegate argument so that the resulting task may be awaited.

So maybe warning in the documentation is just a remnant from previous docs that used `Impersonate` there?

Anyway, in the runtime issue, there is also explicit mention by @davidfowl of [wrapping the entire middleware pipeline in a `RunImpersonated`](https://github.com/dotnet/runtime/issues/24009#issuecomment-544196325):

```csharp
{
    await WindowsIdentity.RunImpersonated(someToken, () => next());
});
```

So I would expect that this scenario is indeed fully supported.

---
#### Document Details

* ID: 401ad9a0-9e51-80a2-5846-82e9790d7257
* Version Independent ID: fbc36c26-9992-1f4c-66d3-02f898ee7ec4
* Content: [Configure Windows Authentication in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/windowsauth?view=aspnetcore-5.0&tabs=visual-studio#impersonation)
* Content Source: [aspnetcore/security/authentication/windowsauth.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/security/authentication/windowsauth.md)
* Product: **aspnet-core**
* GitHub Login: @scottaddie
* Microsoft Alias: **riande**

Rick-Anderson:
Fixes #21020

@poke please review.

### DIFF
--- a/aspnetcore/security/authentication/windowsauth.md
+++ b/aspnetcore/security/authentication/windowsauth.md
@@ -286,7 +286,7 @@ ASP.NET Core doesn't implement impersonation. Apps run with the app's identity f
 
 [!code-csharp[](windowsauth/sample_snapshot/Startup.cs?highlight=10-19)]
 
-`RunImpersonated` doesn't support asynchronous operations and shouldn't be used for complex scenarios. For example, wrapping entire requests or middleware chains isn't supported or recommended.
+<xref:System.Security.Principal.WindowsIdentity.RunImpersonatedAsync%2A> is also available.
 
 ::: moniker range=">= aspnetcore-3.0"
 
--- a/aspnetcore/security/authentication/windowsauth.md
+++ b/aspnetcore/security/authentication/windowsauth.md
@@ -282,11 +282,9 @@ When both Windows Authentication and anonymous access are enabled, use the `[Aut
 
 ## Impersonation
 
-ASP.NET Core doesn't implement impersonation. Apps run with the app's identity for all requests, using app pool or process identity. If the app should perform an action on behalf of a user, use [WindowsIdentity.RunImpersonated](xref:System.Security.Principal.WindowsIdentity.RunImpersonated*) in a [terminal inline middleware](xref:fundamentals/middleware/index#create-a-middleware-pipeline-with-iapplicationbuilder) in `Startup.Configure`. Run a single action in this context and then close the context.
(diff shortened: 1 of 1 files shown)


poke (approved):
That’s simple enough, I guess.
Maybe expand it to _“[…] is also available to impersonate asynchronous operations, including the execution of middleware”_ to make it clear that this is explicitly a supported scenario?

Rick-Anderson:
@poke  take another look. Re-request review is not working
image

poke:
Hm, now you let me notice that _“terminal inline middleware”_ part. 😅  As David said in that linked runtime issue, you *can* run the whole middleware pipeline in an impersonated action. So the middleware isn’t be required to be terminal in order to use impersonation.

Maybe just?

> If the app should perform an action on behalf of a user, use WindowsIdentity.RunImpersonated or RunImpersonatedAsync in a middleware in Startup.Configure. _[end-of-paragraph]_

(I would drop the part about closing the context since that sounds as if this would be something to use with a `using` while you actually pass a function to `RunImpersonated`)

Btw. how does the `<xref:…>` render? I’m asking because the `RunImpersonated` uses a Markdown link with a custom link text.

blowdart:
_you can run the whole middleware pipeline in an impersonated action_

Whether you should or not is another matter. I'd caveat this with dire security warnings if you are going to doc it.

Rick-Anderson:
@poke if you want that change, why don't you PR this and get @blowdart  to author the warning.

poke:
@blowdart
Agreed, it’s probably not a good idea unless you have very special needs (and know what you’re doing).
I just would like the docs not to rule out this possibility which is why I opened the issue in the first place (because someone on Stack Overflow asked how they could do it when the docs said that this doesn’t work).

I’m fine with leaving it out of the docs, to avoid advertise for doing it. Let’s just skip the “terminal” part then :slightly_smiling_face:

blowdart:
Leaving it out works for me :)

PR #21133, opened 2021-01-06
update Note part

FredHan001:
[`HubCallerContext` class](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.signalr.hubcallercontext?view=aspnetcore-5.0) does not have a property named **Others**

### DIFF
--- a/aspnetcore/signalr/hubcontext.md
+++ b/aspnetcore/signalr/hubcontext.md
@@ -53,7 +53,7 @@ app.Use(async (context, next) =>
 ```
 
 > [!NOTE]
-> When hub methods are called from outside of the `Hub` class, there's no caller associated with the invocation. Therefore, there's no access to the `ConnectionId`, `Caller`, and `Others` properties.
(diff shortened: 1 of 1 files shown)


aspnetcore/signalr/hubcontext.md
```diff
@@ -53,7 +53,7 @@ app.Use(async (context, next) =>
 ```
 
 > [!NOTE]
-> When hub methods are called from outside of the `Hub` class, there's no caller associated with the invocation. Therefore, there's no access to the `ConnectionId`, `Caller`, and `Others` properties.
+> When hub methods are called from outside of the `Hub` class, there's no caller associated with the invocation. Therefore, there's no access to the `ConnectionId`, `Caller` and other properties.
```
BrennanConroy, line 56: The doc is talking about how inside the `Hub` you have access to `Clients.Others`, and here it says you do **not** have access to the `Others` property when you're outside the `Hub`.

https://docs.microsoft.com/dotnet/api/microsoft.aspnetcore.signalr.ihubcallerclients

FredHan001: Hi BrennanConroy,  got it, thanks for pointing that.

PR #21168, opened 2021-01-09
Update cookie-sharing.md

### LINKED ISSUE #21163: Sharing Cookies among ASP.NET apps does not work with a session store
The instructions at https://docs.microsoft.com/en-us/aspnet/core/security/cookie-sharing?view=aspnetcore-3.1#share-authentication-cookies-with-aspnet-core-identity don't work when you use an `ITicketStore` (`CookieAuthenticationOptions.SessionStore`).

The reason is that the session id claim set by Owin has the type string `Microsoft.Owin.Security.Cookies-SessionId` while ASP.Net Core expects `Microsoft.AspNetCore.Authentication.Cookies-SessionId`. The result is that the ASP.Net core can't find a session key to use. See https://github.com/dotnet/aspnetcore/blob/master/src/Security/Authentication/Cookies/src/CookieAuthenticationHandler.cs#L151.

I implemented a workaround that can be found at https://stackoverflow.com/a/65630992/587277.

---
#### Document Details

* ID: 99307104-c0d7-3e82-f0b2-af61dcfe072a
* Version Independent ID: 315c02cb-54c4-e386-da85-6e4311ca10c2
* Content: [Share authentication cookies among ASP.NET apps](https://docs.microsoft.com/en-us/aspnet/core/security/cookie-sharing?view=aspnetcore-3.1#share-authentication-cookies-with-aspnet-core-identity)
* Content Source: [aspnetcore/security/cookie-sharing.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/security/cookie-sharing.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

Rick-Anderson:
Fixes #21163

### DIFF
--- a/aspnetcore/security/cookie-sharing.md
+++ b/aspnetcore/security/cookie-sharing.md
@@ -49,7 +49,7 @@ services.ConfigureApplicationCookie(options => {
 });
 ```
 
-**Note:** The preceding instructions don't work with `ITicketStore` (`CookieAuthenticationOptions.SessionStore`). This approach doesn't work because  the session id claim set by Owin has the type string `Microsoft.Owin.Security.Cookies-SessionId` while ASP.Net Core expects `Microsoft.AspNetCore.Authentication.Cookies-SessionId`. For more information, see [this GitHub issue](https://github.com/dotnet/AspNetCore.Docs/issues/21163).
+**Note:** The preceding instructions don't work with `ITicketStore` (`CookieAuthenticationOptions.SessionStore`).  For more information, see [this GitHub issue](https://github.com/dotnet/AspNetCore.Docs/issues/21163).
 
 ## Share authentication cookies without ASP.NET Core Identity
 
--- a/aspnetcore/security/cookie-sharing.md
+++ b/aspnetcore/security/cookie-sharing.md
@@ -49,6 +49,8 @@ services.ConfigureApplicationCookie(options => {
 });
 ```
 
+**Note:** The preceding instructions don't work with `ITicketStore` (`CookieAuthenticationOptions.SessionStore`). This approach doesn't work because  the session id claim set by Owin has the type string `Microsoft.Owin.Security.Cookies-SessionId` while ASP.Net Core expects `Microsoft.AspNetCore.Authentication.Cookies-SessionId`. For more information, see [this GitHub issue](https://github.com/dotnet/AspNetCore.Docs/issues/21163).
+
 ## Share authentication cookies without ASP.NET Core Identity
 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@avendel you write in your SO thread:

> This code should be added to the ASP.Net Framework project. 

Shouldn't that be 
This code should be added to the ASP.NET **_Core_** project. 

Also, at the very top of the post, I'm going to replace 

> Have .NET Core 2.0 application which uses a Data Protection Provider

With

Have ***ASP.NET*** Core app which uses a Data Protection Provider

Rick-Anderson:
@Tratcher  I thought this was too niche/advanced to mention, but the [SO post](https://stackoverflow.com/questions/48510143/) has alot of activity.

avendel:
> Shouldn't that be
> This code should be added to the ASP.NET **_Core_** project.

The workaround from my SO answer adds the ASP.NET Core session id claim during serialization of the authentication ticket.

```csharp
public string Protect(AuthenticationTicket data) {
    var sessionClaim = data.Identity.FindFirst(OwinSessionIdClaim);
    if (sessionClaim != null) {
        data.Identity.AddClaim(new Claim(AspNetCoreSessionIdClaim, sessionClaim.Value));
    }
    return this.dataFormat.Protect(data);
}
```

So this code should run in the ASP.NET Framework OWIN project to make sure the generated cookie contains a session id claim known to the ASP.NET Core project.

One would have to make sure to always add both to get it working in the opposite circumstance where the cookie is generated in the ASP.NET Core project. I will update my SO answer to clarify.

aspnetcore/security/cookie-sharing.md
```diff
@@ -49,6 +49,8 @@ services.ConfigureApplicationCookie(options => {
 });
 ```
 
+**Note:** The preceding instructions don't work with `ITicketStore` (`CookieAuthenticationOptions.SessionStore`). This approach doesn't work because  the session id claim set by Owin has the type string `Microsoft.Owin.Security.Cookies-SessionId` while ASP.Net Core expects `Microsoft.AspNetCore.Authentication.Cookies-SessionId`. For more information, see [this GitHub issue](https://github.com/dotnet/AspNetCore.Docs/issues/21163).
```
Tratcher, line 52:
```suggestion
**Note:** The preceding instructions don't work with `ITicketStore` (`CookieAuthenticationOptions.SessionStore`).  For more information, see [this GitHub issue](https://github.com/dotnet/AspNetCore.Docs/issues/21163).
```

If we're not going to provide a solution here then we don't need the implementation details.

Tratcher (approved):

PR #21190, opened 2021-01-11
Convert Blazor WASM into a PWA

### LINKED ISSUE #20859: Document how to convert a Blazor WebAssembly app into a PWA
We document how to create a PWA from scratch. We should add content on how to turn an existing Blazor app into a PWA.

---
#### Document Details

* ID: 7c8c235e-1ce4-046f-e848-294096361922
* Version Independent ID: 6240f626-97d4-ace0-5e34-20e1c3592f86
* Content: [Build Progressive Web Applications with ASP.NET Core Blazor WebAssembly](https://docs.microsoft.com/en-us/aspnet/core/blazor/progressive-web-app?view=aspnetcore-5.0&tabs=visual-studio)
* Content Source: [aspnetcore/blazor/progressive-web-app.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/progressive-web-app.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes  #20859

[Internal Review Topic (links to section)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/progressive-web-app?view=aspnetcore-3.1&branch=pr-en-us-21190#convert-an-existing-blazor-webassembly-app-into-a-pwa)

I've successfully used the reference assets cross-linking approach you see here in the past. By linking over to the `wwwroot` folder assets, we'll avoid maintaining copies of these assets in the docs repo that tend to go stale in future releases. It does require the reader to select the right branch over there (and the added little NOTE here about the Blazor folder churn between 3.1 and 5.0), but I haven't seen readers having trouble following similar guidance in the past.

**_Did I miss anything?_** This looks complete, but PWAs are one of things that I haven't had much time to play with.

### DIFF
--- a/aspnetcore/blazor/progressive-web-app.md
+++ b/aspnetcore/blazor/progressive-web-app.md
@@ -5,7 +5,7 @@ description: Learn how to build a Blazor-based Progressive Web Application (PWA)
 monikerRange: '>= aspnetcore-3.1'
 ms.author: riande
 ms.custom: mvc
-ms.date: 06/10/2020
+ms.date: 01/11/2020
 no-loc: [appsettings.json, "ASP.NET Core Identity", cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
 uid: blazor/progressive-web-app
 ---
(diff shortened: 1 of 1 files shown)


guardrex:
@captainsafia ... The test app that I converted *seems to be running ok* 🤞 based on the guidance added here. I'll go ahead with this. If you :eyes: and spot anything, I'll touch up these bits.

aspnetcore/blazor/progressive-web-app.md
```diff
@@ -52,6 +52,74 @@ dotnet new blazorwasm -o MyNewProject --pwa
...
+  ```xml
+  <ItemGroup>
+    <ServiceWorker Include="wwwroot\service-worker.js" 
+      PublishedContent="wwwroot\service-worker.published.js" />
+  </ItemGroup>
+  ```
+
+::: moniker range=">= aspnetcore-5.0"
+
+Navigate to the ASP.NET Core GitHub repository at the following URL, which links to 5.0 release reference source and assets. If you aren't converting an app for the 5.0 release, select the release that you're working with from the **Switch branches or tags** drop-down list that applies to your app.
```
captainsafia, line 80: Maybe instead of referring people to the source code we can ask them to use the `dotnet new` interface and copy the assets from there?

Seems better to me than having people dig through GH.

guardrex: I don't think we've ever taken that approach, but we could offer it.

> dig through GH

They don't need to traverse any folders. They just need to select the branch. A certain amount of developer training is probably ok. It looks like GH is something now that all .NET devs working above hobbyist level will need to know a little bit about it. This technique has been used successfully for a while. It doesn't generate doc issues indicating confused developers.

I'll modify the passage to provide both approaches.

Issue #20175, opened 2020-10-12
Multiple OIDC provider support


billyjim:
[EDIT by guardrex to update the metadata]

It would be very helpful to show how to provide OIDC authentication for several providers (e.g., Google, Facebook, Microsoft, etc.) for Blazor WASM: how to add a service for each provider and how to override the authentication login to allow users to choose from a selection of providers and send an authentication request to the selected provider.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 985b834f-534a-aba6-80a0-d8845adbfc25
* Version Independent ID: d2b5a556-82a4-e939-be68-2357a3d4408f
* Content: [Secure an ASP.NET Core Blazor WebAssembly standalone app with the Authentication library](https://docs.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/standalone-with-authentication-library?view=aspnetcore-3.1&tabs=visual-studio)
* Content Source: [aspnetcore/blazor/security/webassembly/standalone-with-authentication-library.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/security/webassembly/standalone-with-authentication-library.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
**UPDATE (11/21)**

Hello @billyjim ... I was just responding to another reader with an update on this. Now that .NET 5 is out :tada: and we have the **_new_** redirect approach for Identity Provider sign in :tada:, I think this will be triaged for work. The pop-up approach with external providers was apparently causing a lot of problems, so I believe we wouldn't have been able to base docs on it.

I'm not sure if management will prefer a reference topic with focused explanations on the minimum configuration or if they'll want a full-blown tutorial. My suggestion is to go with reference material because tutorials take a lot more time to create and maintain.

There's one negative aspect tho to providing coverage. We're getting close to the ⛄ *holidays* 🎁, and I have a lot of other work to do on existing topics and then I leave for an extended vacation 🏖️. Therefore, I think (*guessing*) that this issue will be addressed early next year. We'll see how it plays out.

billyjim:
Thanks for getting in touch, Luke.
I'll check it out and await developments.

Bill

guardrex:
[...]

PR #21209, opened 2021-01-13
Improve WASM security coverage for external IPs

guardrex:
Addresses #20175

cc: @billyjim ... This won't resolve your ask. However, the Blazor framework and existing examples (that I can find anyway) don't provide API or a recommended approach for a standalone WASM app with multiple IPs. Today, the developer can ...

* Write the code for a custom auth provider.
* Perhaps, find a third-party lib that will do it (but I didn't see one at Awesome Blazor).
* Go with a hosted solution and Identity Server, with IdS configured for the desired IPs.
* Go with Blazor Server and configure as an ASP.NET Core app for the desired IPs.

This PR tho can do a couple of quick things to improve the coverage ...

* In the *Auth Lib* topic, call out that the IP must be OIDC compliant and cross-link the *Overview* topic where that aspect is described.
* Show an example ... I use Google here, which was based on [Javier's example](https://github.com/javiercn/BlazorGoogleAuthSample/blob/master/GoogleAuth/Program.cs) and just worked for me locally.

Unless they let us know otherwise, your ask is probably a **_won't fix_** for docs at this time on the basis of:

* The framework not providing anything in-box.
* I think that the team is working hard 🚧⛏️ on 6.0 bits and too busy to knock up a sample.

Javier ... I didn't see an open Blazor issue for multiple IPs or anything on the 6.0 roadmap for it. Let me know if you think there's anything we should say about multiple IPs with a standalone WASM app.

### DIFF
--- a/aspnetcore/blazor/security/webassembly/standalone-with-authentication-library.md
+++ b/aspnetcore/blazor/security/webassembly/standalone-with-authentication-library.md
@@ -106,7 +106,7 @@ Google OAuth 2.0 OIDC example:
 }
 ```
 
-The redirect URI (`https://localhost:5001/authentication/login-callback`) is registered in the [Google APIs console](https://console.developers.google.com/apis/dashboard) in **Credentials** > **{NAME}** > **Authorized redirect URIs**, where `{NAME}` is the app's client name in the **OAuth 2.0 Client IDs** app list.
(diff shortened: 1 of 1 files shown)


billyjim:
Thanks, Luke.
I realized that the best way to handle this is through a 3rd party
authentication app. There are a few that offer the functionality I was
looking for.
Bill

javiercn (approved):

PR #21223, opened 2021-01-14
Fix mismatch between code and explanation in routing.md

ericmutta:
The section [`Specifying attribute route optional parameters, default values, and constraints`](https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/routing?view=aspnetcore-5.0#specifying-attribute-route-optional-parameters-default-values-and-constraints) contains source code with `[HttpPost("product14/{id:int}")]` however the explanatory text that follows, says the code uses `[HttpPost("product/{id:int}")]` which is incorrect (`product` should read `product14`). I have changed the explanatory text, the example of a matching URL (`/product/3` should read `/product14/3`) and the name of controller class (`ProductsController.ShowProduct` should read `Products14Controller.ShowProduct`).

PS: routing.md is a MASSIVE document, it could really use decomposition into multiple smaller docs.

### DIFF
--- a/aspnetcore/mvc/controllers/routing.md
+++ b/aspnetcore/mvc/controllers/routing.md
@@ -553,7 +553,7 @@ Attribute routes support the same inline syntax as conventional routes to specif
 
 [!code-csharp[](routing/samples/3.x/main/Controllers/ProductsController.cs?name=snippet8&highlight=3)]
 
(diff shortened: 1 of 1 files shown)


scottaddie (approved):

scottaddie:
Thank you for the PR, @ericmutta!

PR #21235, opened 2021-01-16
Add PEM config info to Kestrel doc

JamesNK:
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/19113
Fixes https://github.com/dotnet/aspnetcore/issues/26240

Replaces https://github.com/dotnet/AspNetCore.Docs/pull/20913

***EDIT by @Rick-Anderson***
**[Internal Review URL](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/endpoints?view=aspnetcore-5.0&branch=pr-en-us-21235)**

### DIFF
--- a/aspnetcore/fundamentals/servers/kestrel/endpoints.md
+++ b/aspnetcore/fundamentals/servers/kestrel/endpoints.md
@@ -185,6 +185,7 @@ In the following *appsettings.json* example:
         "Url": "https://*:5004",
         "Certificate": {
           "Path": "<path to .pfx file>",
+          "KeyPath": "<path to .key file>",
           "Password": "<certificate password>"
         }
       }
@@ -199,7 +200,27 @@ In the following *appsettings.json* example:
 }
 ```
 
-An alternative to using `Path` and `Password` for any certificate node is to specify the certificate using certificate store fields. For example, the `Certificates` > `Default` certificate can be specified as:
+Schema notes:
+
+* Endpoints names are case-insensitive. For example, `HTTPS` and `Https` are valid.
+* The `Url` parameter is required for each endpoint. The format for this parameter is the same as the top-level `Urls` configuration parameter except that it's limited to a single value.
+* These endpoints replace those defined in the top-level `Urls` configuration rather than adding to them. Endpoints defined in code via `Listen` are cumulative with the endpoints defined in the configuration section.
+* The `Certificate` section is optional. If the `Certificate` section isn't specified, the defaults defined in earlier scenarios are used. If no defaults are available, the server throws an exception and fails to start.
+* The `Certificate` section supports multiple certificate sources:
+  * **Path**&ndash;**Password**
+  * **Path**&ndash;**KeyPath**&dash;**Password**
+  * **Subject**&ndash;**Store**
+* Any number of endpoints may be defined in this way so long as they don't cause port conflicts.
+
+#### Certificate sources
+
+Certificate nodes can be configured to load certificates from a number of sources:
+
+* **Path** and **Password** to load *.pfx* files.
+* **Path**, **KeyPath** and **Password** to load *.pem*/*.crt* and *.key* files.
+* **Subject** and **Store** to load from the certificate store.
+
+For example, the **Certificates** > **Default** certificate can be specified as:
 
 ```json
 "Default": {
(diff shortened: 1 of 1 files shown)


Tratcher (approved):

aspnetcore/fundamentals/servers/kestrel/endpoints.md
```diff
@@ -199,7 +200,27 @@ In the following *appsettings.json* example:
 }
 ```
 
-An alternative to using `Path` and `Password` for any certificate node is to specify the certificate using certificate store fields. For example, the `Certificates` > `Default` certificate can be specified as:
+Schema notes:
+
+* Endpoints names are case-insensitive. For example, `HTTPS` and `Https` are valid.
+* The `Certificate` section is optional. If the `Certificate` section isn't specified, the defaults defined in **Certificates** > **Default** are used. If no defaults are available, the server throws an exception and fails to start.
```
halter73, line 208: It might be worth pointing out it will fallback to a dev certificate if available.

aspnetcore/fundamentals/servers/kestrel/endpoints.md
```diff
@@ -199,7 +200,27 @@ In the following *appsettings.json* example:
 }
 ```
 
-An alternative to using `Path` and `Password` for any certificate node is to specify the certificate using certificate store fields. For example, the `Certificates` > `Default` certificate can be specified as:
+Schema notes:
+
+* Endpoints names are case-insensitive. For example, `HTTPS` and `Https` are valid.
```
halter73, line 205: Maybe link to https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-5.0#configuration-keys-and-values which states "Configuration keys: Are case-insensitive. For example, ConnectionString and connectionstring are treated as equivalent keys."

halter73 (approved):

aspnetcore/fundamentals/servers/kestrel/endpoints.md
```diff
@@ -199,7 +200,24 @@ In the following *appsettings.json* example:
 }
 ```
 
-An alternative to using `Path` and `Password` for any certificate node is to specify the certificate using certificate store fields. For example, the `Certificates` > `Default` certificate can be specified as:
+Schema notes:
+
+* Endpoints names are [case-insensitive](xref:fundamentals/configuration#configuration-keys-and-values). For example, `HTTPS` and `Https` are valid.
```
Rick-Anderson, line 205:
```suggestion
* Endpoints names are [case-insensitive](xref:fundamentals/configuration/index#configuration-keys-and-values). For example, `HTTPS` and `Https` are equivalent.
```

aspnetcore/fundamentals/servers/kestrel/endpoints.md
```diff
@@ -199,7 +200,24 @@ In the following *appsettings.json* example:
...
+* The `Url` parameter is required for each endpoint. The format for this parameter is the same as the top-level `Urls` configuration parameter except that it's limited to a single value.
+* These endpoints replace those defined in the top-level `Urls` configuration rather than adding to them. Endpoints defined in code via `Listen` are cumulative with the endpoints defined in the configuration section.
```
Rick-Anderson, line 207:
```suggestion
* These endpoints replace those defined in the top-level `Url` configuration rather than adding to them. Endpoints defined in code via `Listen` are cumulative with the endpoints defined in the configuration section.
```
Is `Urls` the plural of `Url`? That's typically written as `Url`s which is bad, but not as bad as `Urls`.

aspnetcore/fundamentals/servers/kestrel/endpoints.md
```diff
@@ -199,7 +200,24 @@ In the following *appsettings.json* example:
...
+* The `Certificate` section is optional. If the `Certificate` section isn't specified, the defaults defined in **Certificates** > **Default** are used. If no defaults are available then the development certificate is used. If there are no defaults and the development certificate isn't present then the server throws an exception and fails to start.
```
Rick-Anderson, line 208:
```suggestion
* The `Certificate` section is optional. If the `Certificate` section isn't specified, the defaults defined in `Certificates:Default` are used. If no defaults are available, the development certificate is used. If there are no defaults and the development certificate isn't present, the server throws an exception and fails to start.
```

Rick-Anderson: `**Certificates** > **Default**` would have been localized.

aspnetcore/fundamentals/servers/kestrel/endpoints.md
```diff
@@ -199,7 +200,24 @@ In the following *appsettings.json* example:
...
+* The `Certificate` section supports multiple [certificate sources](#certificate-sources).
+* Any number of endpoints may be defined in this way so long as they don't cause port conflicts.
```
Rick-Anderson, line 210:
```suggestion
* Any number of endpoints may be defined in [Configuration](xref:fundamentals/configuration/index) as long as they don't cause port conflicts.
```

aspnetcore/fundamentals/servers/kestrel/endpoints.md
```diff
@@ -199,7 +200,24 @@ In the following *appsettings.json* example:
...
+
+#### Certificate sources
+
+Certificate nodes can be configured to load certificates from a number of sources:
+
+* **Path** and **Password** to load *.pfx* files.
+* **Path**, **KeyPath** and **Password** to load *.pem*/*.crt* and *.key* files.
+* **Subject** and **Store** to load from the certificate store.
+
+For example, the **Certificates** > **Default** certificate can be specified as:
```
Rick-Anderson, line 220:
```suggestion
* `Path` and `Password` to load *.pfx* files.
* `Path`, `KeyPath` and `Password` to load *.pem*/*.crt* and *.key* files.
* `Subject` and `Store` to load from the certificate store.

For example, the `Certificates:Default` certificate can be specified as:
```

Rick-Anderson: You don't want them localized, right?

Rick-Anderson (changes requested):

aspnetcore/fundamentals/servers/kestrel/endpoints.md
JamesNK: I believe there is a setting called `Urls` that this refers to.

aspnetcore/fundamentals/servers/kestrel/endpoints.md
JamesNK: I'm going to update this one place, but `**Certificates** > **Default**` is used in many other places in this doc.

aspnetcore/fundamentals/servers/kestrel/endpoints.md
Rick-Anderson: > I believe there is a setting called `Urls` that this refers to.

Can you confirm that?
It's not in the sample file, but that's OK.

Rick-Anderson (approved):

aspnetcore/fundamentals/servers/kestrel/endpoints.md
JamesNK: It refers to the Urls mentioned at the start of the article - https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/endpoints

That part should be split out into its own section so others can link to it.

PR #21249, opened 2021-01-19
Blazor script location

### LINKED ISSUE #21200: Surface location of Blazor scripts
Surface best practice location of Blazor scripts guidance (i.e., inside the closing `</body>` tag). Possibly here and elsewhere. 

It's already covered in the *Prerendering and integration* topic; but since that topic will usually only be consumed by devs supporting components in an RP/MVC app or performing a total to-Blazor conversion, it probably won't catch devs who are just working with scripts.

Wherever the new language/section goes, it probably should be linked from the JS interop topics and anyplace else where we discuss adding scripts to `wwwroot/index.html` and/or `Pages/_Host.cshtml`.

Related: https://github.com/dotnet/aspnetcore/issues/29215

cc: @mrlife

---
#### Document Details

* ID: d31248d7-d2c2-1a17-3173-889a0d9bb6e8
* Version Independent ID: 6f27ff37-6b3a-57f5-353a-5cf5aabba7cf
* Content: [ASP.NET Core Blazor project structure](https://docs.microsoft.com/en-us/aspnet/core/blazor/project-structure?view=aspnetcore-5.0)
* Content Source: [aspnetcore/blazor/project-structure.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/project-structure.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #21200

Thanks @mrlife! 🎸 

* I'll focus on the JS interop use case as an example for this.
* Safia, I'm concerned about how to phrase the scope of the problem on ordering (this says 'when it's important' and leaves it at that), but this might be ok if it hits the main general point and JS interop is a decent example.


aspnetcore/blazor/call-dotnet-from-javascript.md
```diff
@@ -19,6 +19,9 @@ This article covers invoking .NET methods from JavaScript. For information on ho
 
 [View or download sample code](https://github.com/dotnet/AspNetCore.Docs/tree/master/aspnetcore/blazor/common/samples/) ([how to download](xref:index#how-to-download-a-sample))
 
+> [!NOTE]
+> The order that custom JavaScript code is loaded from JavaScript files in the `wwwroot/index.html` file (Blazor WebAssembly) or `Pages/_Host.cshtml` file (Blazor Server) is important in some scenarios, such as JS interop. Add custom JS interop scripts after the `<script>` element for Blazor's script (`blazor.webassembly.js`/`blazor.server.js`). Add scripts before the closing `</body>` tag. For more information, see <xref:blazor/project-structure>.
```
captainsafia, line 23: > Add custom JS interop scripts after the `<script>` element for Blazor's script (`blazor.webassembly.js`/`blazor.server.js`). Add scripts before the closing `</body>` tag.

This confused me a bit when reading. Would it be possible to show a code sample in the note?

```
<body>
   <script src="yourstuff.js"></script>
   <script src="blazor.webassembly.js"></script>
</body>
```

aspnetcore/blazor/project-structure.md
```diff
@@ -62,6 +62,16 @@ The following files and folders are produced from the Blazor WebAssembly templat
 
 ::: moniker-end
 
+> [!NOTE]
+> Custom script files added to the `index.html` file should be added after the `<script>` element for `blazor.webassembly.js` if the order of loading JavaScript code is important to Blazor. All script elements should appear before the closing `</body>` tag. The following example adds a JS interop script (`wwwroot/js/interop.js`):
```
captainsafia, line 66: I like this style a lot more!

captainsafia (approved):
I dunno that we want to cover the script precedence in JavaScript in our docs. It's a bit of a tangential topic and I'm not sure that it will be more helpful than explicitly tell users what the order should be.

aspnetcore/blazor/call-dotnet-from-javascript.md
mrlife: In my experience, it was "Add custom JS interop scripts _before_ the `<script>` element for Blazor's script ..."

guardrex:
> I dunno that we want to cover the script precedence in JavaScript in our docs. It's a bit of a tangential topic and I'm not sure that it will be more helpful than explicitly tell users what the order should be.

> I like this style a lot more!

... and my scoping question (i.e., "'when it's important'") ...

I agree. Let's **_NOT_** get into the weeds on when the order matters because (at least) it won't matter in some scenarios and (at most) we can't tell devs what the exact order is. What we can do is make readers aware of the concept.

I'll revise this.

guardrex:
@captainsafia @mrlife ... This latest commit makes it generic and only calls out that the scripts are placed inside the closing `</body>` tag and which files (`index.html` or `_Host.cshtml`). If this isn't useful, then I recommend that we close as a *won't fix*. Ordering is a known conceptual point for basic JS use scenarios and covered by external JS docs.

We could have a *Fundamentals* > **_JavaScript in Blazor_**-type of topic one day, where could get into the weeds on ordering. It could also cover things like [Javascript caching (dotnet/AspNetCore.Docs #21278)](https://github.com/dotnet/AspNetCore.Docs/issues/21278) scenarios. **_BUT_**, I recommend discussing it further later ... mid-year ... I'm 🏃 for quite a while to get back to UE passes, which are once again being dogged by other issues popping up.

mrlife:
That the Blazor script needs to come after any JS interop calls - is important and is lost in the latest commit. That was the issue I originally had that was the catalyst for this PR, so please consider mentioning that. It's always true, right?

guardrex:
I defer to @captainsafia 🙏 ...

> I dunno that we want to cover the script precedence in JavaScript in our docs. It's a bit of a tangential topic and I'm not sure that it will be more helpful than explicitly tell users what the order should be.

... the "than explicitly tell users what the order should be" bit. Perhaps, I misinterpreted what you meant by that, @captainsafia.

captainsafia:
> That the Blazor script needs to come after any JS interop calls - is important and is lost in the latest commit. That was the issue I originally had that was the catalyst for this PR, so please consider mentioning that. It's always true, right?

If it helps, @guardrex we can include it as an illustrative example. 

> The order that custom JavaScript code is loaded from JavaScript files is important in some scenarios. For example, ensure that JavaScript files with interop methods are included before Blazor framework files.

captainsafia (approved):

guardrex:
That should be ok for now. I'll touch the wording again when I reach these topics for a UE pass in a bit.

PR #21252, opened 2021-01-19
Removed extra `@` to match all other uses

mrlife:


guardrex (approved):
# 😈 

That one got by review, too! :smile:

PR #21256, opened 2021-01-20
Remove unnecessary word in comparison.md

ericmutta:
The following sentence:

> From this file gRPC frameworks will code generate a service base class...

Has the word "code" which is not required and should read:

> From this file gRPC frameworks will generate a service base class...

### DIFF
--- a/aspnetcore/grpc/comparison.md
+++ b/aspnetcore/grpc/comparison.md
@@ -44,7 +44,7 @@ HTTP/2 is not exclusive to gRPC. Many request types, including HTTP APIs with JS
 
 ### Code generation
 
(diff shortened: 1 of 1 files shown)


guardrex (approved):
Thanks @ericmutta ... There were a few more nits to update on this line (and in the whole topic really), so I just touched it up a bit more. Thanks for sending this in.

ericmutta:
Many thanks @guardrex for following up on this and other PRs. And kudos to you and the team for the ASP.NET docs - they are extremely detailed and have been very handy on my quest to acquire the sum of all ASP.NET knowledge :smile: :rocket:

PR #21281, opened 2021-01-22
Flesh out SetParametersAsync example

### LINKED ISSUE #20921: Enhance Blazor Lifecycle topic example for SetParametersAsync
I don't find the documentation about custom route constraint (IRouteConstraint) for a Blazor Page.

`@page "/{name:myconstraint}"`

Where **myconstraint** would be a _custom route constraint_.

---
#### Document Details

* Content: [ASP.NET Core Blazor routing](https://docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-5.0)
* Content Source: [aspnetcore/blazor/fundamentals/routing.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/fundamentals/routing.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #20921

Thanks @dumaisf! :rocket:

Safia, the example thus far on the PR might be a steaming pile of **_Rex dinosaur droppings_** 💩〰️👃😵, so the example might need work or replacement with something you think would be better. Whatever we do, I hope we flesh out the example and make it a complete *cut-'n-paste* experience. As we discussed offline, it's not an example of a workaround for doing anything specific with the parameters, just *how to work with them generally*.

@dumaisf ... Following our chat on the issue, you can interact with the parameters here ... try to parse one (or several) into your custom type(s) ... if they don't parse, do something else besides load the component ... throw ... redirect. That's a *workaround*/*hack* **_idea only_**, and Safia and I didn't discuss it further. You can talk to the community further about it for other ideas ...

* [Stack Overflow (tag: `blazor`)](https://stackoverflow.com/questions/tagged/blazor)
* [ASP.NET Core Slack Team](http://tattoocoder.com/aspnet-slack-sign-up/)
* [Blazor Gitter](https://gitter.im/aspnet/Blazor)

We won't address the custom route constraint part because that's not a thing at this time. We usually only cover what the framework is capable of doing at any time. If the product unit comes up with something for that for a future release, we'll get it into the docs.

### DIFF
--- a/aspnetcore/blazor/components/lifecycle.md
+++ b/aspnetcore/blazor/components/lifecycle.md
@@ -72,16 +72,13 @@ In the following example, <xref:Microsoft.AspNetCore.Components.ParameterView.Tr
 
     public override async Task SetParametersAsync(ParameterView parameters)
     {
-        if (parameters.TryGetValue<string>(nameof(Param), out var value))
+        if (!parameters.TryGetValue<string>(nameof(Param), out var value))
         {
-            if (value is null)
-            {
-                message = $"The value of 'Param' is null.";
-            }
-            else
-            {
-                message = $"The value of 'Param' is {value}.";
-            }
+            message = "The value of 'Param' is null.";
+        }
+        else 
+        {
+            message = $"The value of 'Param' is {value}.";
         }
 
         await base.SetParametersAsync(parameters);
--- a/aspnetcore/blazor/components/lifecycle.md
+++ b/aspnetcore/blazor/components/lifecycle.md
@@ -51,14 +51,41 @@ Developer calls to [`StateHasChanged`](#state-changes) result in a render. For m
 
 ### Before parameters are set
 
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/components/lifecycle.md
```diff
@@ -51,14 +51,41 @@ Developer calls to [`StateHasChanged`](#state-changes) result in a render. For m
...
 
-    await base.SetParametersAsync(parameters);
+    [Parameter]
+    public string Param { get; set; }
+
+    public override async Task SetParametersAsync(ParameterView parameters)
+    {
+        if (parameters.TryGetValue<string>(nameof(Param), out var value))
+        {
+            if (value is null)
```
captainsafia, line 77: Instead of doing this if statement, you can do something like:

```csharp
if (!parameters.TryGetValue<string>(nameof(Param), out var value))
{
  message = $"The value of 'Param' is null.";
}
else 
{
  message = $"The value of 'Param' is {value}.";
}
```

captainsafia (approved):
Looks good -- left a small comment inline.

Issue #20319, opened 2020-10-27
Revert <head> control Blazor guidance


guardrex:
**NOTE**: There is deactivated `Link` component content in the *Class Libraries* topic to update/reactivate for 6.0.

When components for influencing `<head>` elements come back, revert changes made on  #20318.

TOC *Built-in components* entries

```
- name: Link
  href: blazor/fundamentals/additional-scenarios.md#influence-html-head-tag-elements
- name: Meta
  href: blazor/fundamentals/additional-scenarios.md#influence-html-head-tag-elements
- name: Title
  href: blazor/fundamentals/additional-scenarios.md#influence-html-head-tag-elements
```

Release notes ...

```
### Influence the HTML head in Blazor apps

Use the new `Title`, `Link`, and `Meta` components to programmatically set the title of a page and dynamically add links and meta tags to the HTML head in a Blazor app. For more information, see <xref:blazor/fundamentals/additional-scenarios#influence-html-head-tag-elements>.
```

Also, update the following line ...

> \*This section applies to the upcoming ASP.NET Core 5.0 release of Blazor WebAssembly and Blazor Server.\*

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: fc56d254-09c5-993b-2559-76af9a3f97ed
* Version Independent ID: a0bb183f-7e4d-ba39-c94c-cf60e951c7a9
* Content: [ASP.NET Core Blazor hosting model configuration](https://docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/additional-scenarios?view=aspnetcore-5.0)
* Content Source: [aspnetcore/blazor/fundamentals/additional-scenarios.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/fundamentals/additional-scenarios.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
@captainsafia ... Will the work for bringing back `<head>` control in Blazor apps be done for the 6.0 release? I just want to get the labeling right for this docs issue.

captainsafia:
@guardrex It's on the roadmap for 6.0 but prioritized pretty low. I would tentatively put this on-hold until we confirm we are doing this for 6.0.

guardrex:
Note to self:

[...]

PR #21331, opened 2021-01-27
Blazor SignalR guidance re-organization

### LINKED ISSUE #20223: Blazor Server configuration and limits

As per various document i have read , it is mentioned that Blazor Server manages state for each connected client.
How can we do sizing for Blazor Server side, How do we get a number on how many concurrent user can a Blazor Server side support?
What kind of memory does it use for each connected user what does it store on server ?
Can timeout be configured for longer or shorter duration than the default ?

---
#### Document Details

* ID: 7367a651-2b27-0a57-e7d0-95686cc97f1c
* Version Independent ID: 8c769648-728d-72ce-ebf6-2a06e5076aa2
* Content: [ASP.NET Core Blazor configuration](https://docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/configuration?view=aspnetcore-3.1)
* Content Source: [aspnetcore/blazor/fundamentals/configuration.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/fundamentals/configuration.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #20223

[Internal Review Topic (Blazor *Fundamentals* node SignalR guidance)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/signalr?view=aspnetcore-3.1&branch=pr-en-us-21331)

Thanks @nssidhu! :rocket:

Brennan ...

* I think I've successfully avoided significant duplication with what's in the SignalR config topic for SignalR options. The goal for the options coverage in the *Circuit handler options* section is to surface them for `AddServerSideBlazor` and the hub context options for `AddHubOptions` with a couple of *Easy Squeezie*:tm: cut-'n-paste code examples. Although we had coverage in the *Threat Mitigation* security topic, it was rather buried down there and incomplete.

  **The following ❓ has been resolved.**

  ❓ ~For `SupportedProtocols`, the pattern here would be to state what goes in for "..." of `new List<string>() { ... }` that represents the default full set of protocols. However, the description is ...~

  > ~Protocols supported by this hub. By default, all protocols registered on the server are allowed, but protocols can be removed from this list to disable specific protocols for individual hubs.~

  ~... which makes me think that we should split out this one into a separate example that:~

  * ~Lists (in text) what the default protocols are.~
  * ~Shows a code example on how to remove one of them.~

  ~How shall we handle `SupportedProtocols` here?~

  **... that's RESOLVED!** :+1:

* The rest of this is to coalesce our SignalR and Blazor Server hosting model config that deals with the circuit in the new Blazor *Fundamentals* node SignalR guidance article. I think the title, file name (URL), and UID are ok, but we can change those if we come up with something better.

* Static files coverage goes into its own topic, and I think I'll have some additional coverage to coalesce on static files with Blazor later.

* The commented-out bit that's lingering in the new SignalR topic here for components that modify the `<head>` will be moved to another topic later when I reach [Revert <head> control Blazor guidance (dotnet/AspNetCore.Docs #20319)](https://github.com/dotnet/AspNetCore.Docs/issues/20319). That commented-out markdown can stay there for the time being.

btw -- Ignore the first commit naming ("Server") 👇. Most of it focuses on Blazor Server, but this topic is really our general Blazor SignalR coverage (e.g., cross-coverage for SignalR with hosted WASM would also be covered by the *SignalR* topic here).


BrennanConroy:
>     * Lists (in text) what the default protocols are.

Well the only real default is Json, when you use `AddMessagePackProtocol` then you get MessagePack. And `AddServerSideBlazor` adds BlazorPack, and if you remove that, good luck getting Blazor to work 😆 

>     * How shall we handle `SupportedProtocols` here?

I don't think it's a very common scenario to mess with that list. It's probably fine to just not worry about it.

guardrex:
> not worry about it

Should that option be removed from the doc, or should I leave it set to `new List<string>() { ... }`?

BrennanConroy:
Actually, I think it should be removed. I don't think there is any scenario where you want to change it for the Blazor Hub.

BrennanConroy:
SignalR stuff™ looks fine, @pranavkm for Blazor things™

guardrex:
Thanks @BrennanConroy ... I'll stand-by until @pranavkm has time to take a look.

@pranavkm ... There's no giant rush here. This just moves things around to improve how we surface the (mostly Blazor Server) SignalR coverage in the *Fundamentals* node. What's going on is listed in my opening comment :point_up:.

guardrex:
@pranavkm ... Did you want to look at this one 👀. Since this mostly just moves things around and doesn't change content, we can go ahead if you're 🏃😅.

Issue #19797, opened 2020-09-08
Miss a strategy to handle the refresh of tokens


andmig-ilty:
It would be great if this article would also contains a pattern to handle the token expiration and refresh of the access_token.

BTW, I applied the suggested pattern, while using a customized Oauth Provider and still works fine!

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 6f8acbbf-c9d4-7041-c1a7-ec9ce8f0981b
* Version Independent ID: c98be365-408d-7ee6-cb74-14c44d01b0b8
* Content: [ASP.NET Core Blazor Server additional security scenarios](https://docs.microsoft.com/en-us/aspnet/core/blazor/security/server/additional-scenarios?view=aspnetcore-3.1)
* Content Source: [aspnetcore/blazor/security/server/additional-scenarios.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/security/server/additional-scenarios.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Hello @andmig-ilty ... Engineering has refused to show that scenario. They've flatly said that refresh tokens and client-side apps create security risks because there's no way to completely and effectively deal with them. If you search this out over on the engineering repo (try the filter phrase "refresh token" to see if it filters), you'll probably find the answers.

If all of the issues are closed but you'd like to strike up a **_new_** convo with them about some **_new_** approach that might deal with the security aspects, open a **_new_** issue with them (i.e., don't comment on closed issues over there).

Here's the Blazor-filtered engineering issues link ...

https://github.com/dotnet/aspnetcore/issues?q=is%3Aopen+is%3Aissue+label%3Aarea-blazor

andmig-ilty:
As far as I know Blazor Server is a server side application and the secrets are never threated on the client. This is opposite to Blazor WebAssembly where, I'm with you, we should never handle secrets and thus, the oauth implicity flow is then used (no client secret, no refresh_token, etc)

The linked article do refer to Blazor **Server** and for completeness the refresh_token workflow worth to be covered...

guardrex:
[...]

PR #21345, opened 2021-01-28
User claims added or updated

### LINKED ISSUE #15445: Claims added to the user on first registration

### LINKED ISSUE #21560: React to Google docs updates
It looks like Google has updated their docs (and possibly their OAuth UI, too) since this topic was written. The topic calls for the reader to ...

> Navigate to `Integrating Google Sign-In into your web app` and select **Configure a project**.

... but that Google doc at https://developers.google.com/identity/sign-in/web/sign-in doesn't have a **Configure a project** thing.

Looks like it should change to ... "**Create authorization credentials**" ... maybe ... ❓.

@Rick-Anderson ... I'm going to try and fix this on my existing PR at https://github.com/dotnet/AspNetCore.Docs/pull/21345. However, let me know if you want me to leave this out of that PR for work by you or someone else. Please re-assign this if that's the best way to go.

---
#### Document Details

* ID: 6d19b295-f935-3755-4ca4-9d0717b21e75
* Version Independent ID: 747a179e-90b7-de07-e270-82414963048a
* Content: [Google external login setup in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/social/google-logins?view=aspnetcore-5.0)
* Content Source: [aspnetcore/security/authentication/social/google-logins.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/security/authentication/social/google-logins.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

guardrex:
Fixes #15445
Fixes #21560

Let's consider this again. 😄

Hao, we left this off over a year ago on the closed PR https://github.com/dotnet/AspNetCore.Docs/pull/15491.

The scenario is that there's an app in production. New external provider claims are needed. The dev updates the app according to the guidance for adding claims following the topic and on top of existing external claims that were put into the app when it was first built (i.e., just following the same pattern that they used when they built the app). The dev deploys the app to production with the new claims code in place. There are two user scenarios at this point:

* New claims aren't added to users on sign in, they're added on registration.
* New claims aren't updated for users who are currently signed into the app.

From the PR discussion and starting with language that Chris suggested, I've **_drafted_** text to **_start to cover_** the two scenarios, **_but_** ...

* On layout, should we have these two new sections here, or how should it be organized otherwise?
* On the content when I was talking to Chris about it, I merely took the low-hanging fruit and deleted my test user and recreated them. Of course, that works, but it's only useful in Development with test user accounts. On the content pertaining to the `UserManager` and `SigninManager` in a production app, are these the right calls for these two scenarios and are they in the right sections with the right (starting) text?
* WRT **_where/how to call these_**, can you be even more specific than [what you said in your last two remarks](https://github.com/dotnet/AspNetCore.Docs/pull/15491#discussion_r348664977) ... **_exactly where/how_** are these calls made?

btw - If it's best for you to take and resolve the issue with your own PR, let's close this PR and assign the issue to you. I can circle around on your PR for some **_Grammar-i-zation_**:tm: updates if needed.

It's also possible that this is a *won't fix* documentation scenario ...

* These scenarios weren't described or even mentioned in any of the *Identity and Security* topics ... *by-design*? ... are these scenarios covered somewhere that I'm not seeing (e.g., external-to-our-repo MS *Identity* docs and samples)?
* We've gone for year now, and no issues are coming in on these scenarios. No one has remarked on the existing issue or closed cross-linked issues. There are a few :+1:, but not many ... and none of them are recent.

### DIFF
--- a/aspnetcore/security/authentication/social/additional-claims.md
+++ b/aspnetcore/security/authentication/social/additional-claims.md
@@ -110,7 +110,7 @@ For more information, see <xref:Microsoft.AspNetCore.Authentication.OAuth.Claims
 
 ## Add claims to users
 
-Claims are added to the user on first registration, not on sign in. If additional claims are enabled in an app after a user registers to use the app, call [SignInManager.RefreshSignInAsync](xref:Microsoft.AspNetCore.Identity.SignInManager%2A) on a user to force the generation of a new authentication cookie.
+Claims are added to the user on first registration, not on sign in. If additional claims are enabled in an app after a user registers to use the app, call [SignInManager.RefreshSignInAsync](xref:Microsoft.AspNetCore.Identity.SignInManager%601) on a user to force the generation of a new authentication cookie.
 
 In the Development environment working with test user accounts, you can simply delete and recreate the user account.
 
@@ -118,8 +118,8 @@ In the Development environment working with test user accounts, you can simply d
 
 Claims that change after a user signs in aren't automatically updated. If claims change while a user is signed in, call:
 
-* [UserManager.ReplaceClaimAsync](xref:Microsoft.AspNetCore.Identity.UserManager%2A) on the user for claims stored in the identity database.
-* [SignInManager.RefreshSignInAsync](xref:Microsoft.AspNetCore.Identity.SignInManager%2A) on the user to force the generation of a new authentication cookie.
+* [UserManager.ReplaceClaimAsync](xref:Microsoft.AspNetCore.Identity.UserManager%601) on the user for claims stored in the identity database.
+* [SignInManager.RefreshSignInAsync](xref:Microsoft.AspNetCore.Identity.SignInManager%601) on the user to force the generation of a new authentication cookie.
 
 ## Removal of claim actions and claims
 
--- a/aspnetcore/security/authentication/social/additional-claims.md
+++ b/aspnetcore/security/authentication/social/additional-claims.md
@@ -135,12 +135,13 @@ public async Task<IActionResult> OnGetCallbackAsync(
     if (remoteError != null)
     {
         ErrorMessage = $"Error from external provider: {remoteError}";
+
         return RedirectToPage("./Login", new {ReturnUrl = returnUrl });
     }
 
-    var externalLogin = await _signInManager.GetExternalLoginInfoAsync();
+    var info = await _signInManager.GetExternalLoginInfoAsync();
 
-    if (externalLogin == null)
+    if (info == null)
     {
         ErrorMessage = "Error loading external login information.";
         return RedirectToPage("./Login", new { ReturnUrl = returnUrl });
@@ -148,17 +149,18 @@ public async Task<IActionResult> OnGetCallbackAsync(
 
     // Sign in the user with this external login provider if the user already has a 
     // login.
-    var result = await _signInManager.ExternalLoginSignInAsync(externalLogin.LoginProvider, 
-        externalLogin.ProviderKey, isPersistent: false, bypassTwoFactor : true);
+    var result = await _signInManager.ExternalLoginSignInAsync(info.LoginProvider, 
+        info.ProviderKey, isPersistent: false, bypassTwoFactor : true);
+
     if (result.Succeeded)
     {
         _logger.LogInformation("{Name} logged in with {LoginProvider} provider.", 
-            externalLogin.Principal.Identity.Name, externalLogin.LoginProvider);
+            info.Principal.Identity.Name, info.LoginProvider);
 
         if (_claimsToSync.Count > 0)
         {
-            var user = await _userManager.FindByLoginAsync(externalLogin.LoginProvider, 
-                externalLogin.ProviderKey);
+            var user = await _userManager.FindByLoginAsync(info.LoginProvider, 
+                info.ProviderKey);
             var userClaims = await _userManager.GetClaimsAsync(user);
             bool refreshSignIn = false;
 
@@ -167,9 +169,9 @@ public async Task<IActionResult> OnGetCallbackAsync(
                 var userClaim = userClaims
                     .FirstOrDefault(c => c.Type == addedClaim.Key);
 
-                if (externalLogin.Principal.HasClaim(c => c.Type == addedClaim.Key))
+                if (info.Principal.HasClaim(c => c.Type == addedClaim.Key))
                 {
-                    var externalClaim = externalLogin.Principal.FindFirst(addedClaim.Key);
+                    var externalClaim = info.Principal.FindFirst(addedClaim.Key);
 
                     if (userClaim == null)
                     {
@@ -198,9 +200,10 @@ public async Task<IActionResult> OnGetCallbackAsync(
                 await _signInManager.RefreshSignInAsync(user);
             }
         }
-                
+
         return LocalRedirect(returnUrl);
     }
+
     if (result.IsLockedOut)
     {
         return RedirectToPage("./Lockout");
@@ -210,13 +213,13 @@ public async Task<IActionResult> OnGetCallbackAsync(
         // If the user does not have an account, then ask the user to create an 
         // account.
         ReturnUrl = returnUrl;
-        ProviderDisplayName = externalLogin.ProviderDisplayName;
+        ProviderDisplayName = info.ProviderDisplayName;
 
-        if (externalLogin.Principal.HasClaim(c => c.Type == ClaimTypes.Email))
+        if (info.Principal.HasClaim(c => c.Type == ClaimTypes.Email))
         {
             Input = new InputModel
             {
-                Email = externalLogin.Principal.FindFirstValue(ClaimTypes.Email)
+                Email = info.Principal.FindFirstValue(ClaimTypes.Email)
             };
         }
 
--- a/aspnetcore/security/authentication/social/additional-claims.md
+++ b/aspnetcore/security/authentication/social/additional-claims.md
@@ -112,7 +112,112 @@ For more information, see <xref:Microsoft.AspNetCore.Authentication.OAuth.Claims
 
 Claims are copied from external providers to the user database on first registration, not on sign in. If additional claims are enabled in an app after a user registers to use the app, call [SignInManager.RefreshSignInAsync](xref:Microsoft.AspNetCore.Identity.SignInManager%601) on a user to force the generation of a new authentication cookie.
 
-In the Development environment working with test user accounts, you can simply delete and recreate the user account.
+In the Development environment working with test user accounts, you can simply delete and recreate the user account. For production systems, new claims added to the app can be backfilled into user accounts. After [scaffolding the `ExternalLogin` page](xref:security/authentication/scaffold-identity) into the app at `Areas/Pages/Identity/Account/Manage`, add the following code to the `ExternalLoginModel` in the `ExternalLogin.cshtml.cs` file.
+
+Add a dictionary of added claims. Use the dictionary keys to hold the claim types, and use the values to hold a default value. Add the following line to the top of the class. The following example assumes that one claim is added for the user's Google picture with a generic headshot image as the default value:
+
+```csharp
+private readonly IDictionary<string, string> _addedClaims = 
+    new Dictionary<string, string>()
+    {
+        { "urn:google:picture", "https://localhost:5001/headshot.png" },
+    };
+```
+
+Replace the default code of the `OnGetCallbackAsync` method with the following code. The code loops through the claims dictionary. Claims are added (backfilled) or updated for each user. When claims are added or updated, the user sign-in is refreshed using the <xref:Microsoft.AspNetCore.Identity.SignInManager%2A>, preserving the existing authentication properties (`AuthenticationProperties`).
+
+```csharp
+public async Task<IActionResult> OnGetCallbackAsync(
+    string returnUrl = null, string remoteError = null)
+{
+    returnUrl = returnUrl ?? Url.Content("~/");
+    if (remoteError != null)
+    {
+        ErrorMessage = $"Error from external provider: {remoteError}";
+        return RedirectToPage("./Login", new {ReturnUrl = returnUrl });
+    }
+    var info = await _signInManager.GetExternalLoginInfoAsync();
+    if (info == null)
+    {
+        ErrorMessage = "Error loading external login information.";
+        return RedirectToPage("./Login", new { ReturnUrl = returnUrl });
+    }
+
+    // Sign in the user with this external login provider if the user already has a 
+    // login.
+    var result = await _signInManager.ExternalLoginSignInAsync(info.LoginProvider, 
+        info.ProviderKey, isPersistent: false, bypassTwoFactor : true);
+    if (result.Succeeded)
+    {
+        _logger.LogInformation("{Name} logged in with {LoginProvider} provider.", 
+            info.Principal.Identity.Name, info.LoginProvider);
+
+        if (_addedClaims.Count > 0)
+        {
+            var user = await _userManager.FindByLoginAsync(info.LoginProvider, 
+                info.ProviderKey);
+            var userClaims = await _userManager.GetClaimsAsync(user);
+            bool refreshSignIn = false;
+
+            foreach (var addedClaim in _addedClaims)
+            {
+                var userClaim = userClaims
+                    .FirstOrDefault(c => c.Type == addedClaim.Key);
+
+                if (info.Principal.HasClaim(c => c.Type == addedClaim.Key))
+                {
+                    var principalClaim = info.Principal.FindFirst(addedClaim.Key);
+
+                    if (userClaim == null)
+                    {
+                        await _userManager.AddClaimAsync(user, 
+                            new Claim(addedClaim.Key, principalClaim.Value));
+                        refreshSignIn = true;
+                    }
+                    else if (userClaim.Value != principalClaim.Value)
+                    {
+                        await _userManager
+                            .ReplaceClaimAsync(user, userClaim, principalClaim);
+                        refreshSignIn = true;
+                    }
+                }
+                else if (userClaim == null)
+                {
+                    await _userManager.AddClaimAsync(user, new Claim(addedClaim.Key, 
+                        addedClaim.Value));
+                    refreshSignIn = true;
+                }
+            }
+
+            if (refreshSignIn)
+            {
+                await _signInManager.RefreshSignInAsync(user);
+            }
+        }
+                
+        return LocalRedirect(returnUrl);
+    }
+    if (result.IsLockedOut)
+    {
+        return RedirectToPage("./Lockout");
+    }
+    else
+    {
+        // If the user does not have an account, then ask the user to create an 
+        // account.
+        ReturnUrl = returnUrl;
+        ProviderDisplayName = info.ProviderDisplayName;
+        if (info.Principal.HasClaim(c => c.Type == ClaimTypes.Email))
+        {
+            Input = new InputModel
+            {
+                Email = info.Principal.FindFirstValue(ClaimTypes.Email)
+            };
+        }
+        return Page();
+    }
+}
+```
 
 ## Update user claims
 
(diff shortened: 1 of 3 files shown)


HaoK:
I think this PR is good, and the updates look correct to me.  I know we ended up duping a lot of issues and consolidating them as I still have a few of these assigned to me, for example https://github.com/dotnet/AspNetCore.Docs/issues/19797 is the main ask, how do they refresh tokens.

@blowdart for his eyes and thoughts on your questions about topics too.

guardrex:
@blowdart ... The most important ❓ from my 📖 **_volume_** 📖 on the subject :point_up: is ...

> WRT where/how to call these ... exactly where/how are these calls made?

*Security and Identity* docs in this doc set don't address these, so I don't have anything to cross-link here.

blowdart:
That would be better asked of @Tratcher

aspnetcore/security/authentication/social/additional-claims.md
```diff
@@ -105,6 +108,19 @@ Users can define custom actions by deriving from <xref:Microsoft.AspNetCore.Auth
...
+
+In the Development environment working with test user accounts, you can simply delete and recreate the user account.
+
+## Update user claims
+
+Claims that change after a user signs in aren't automatically updated. If claims change while a user is signed in, call:
+
+* [UserManager.ReplaceClaimAsync](xref:Microsoft.AspNetCore.Identity.UserManager%601) on the user for claims stored in the identity database.
+* [SignInManager.RefreshSignInAsync](xref:Microsoft.AspNetCore.Identity.SignInManager%601) on the user to force the generation of a new authentication cookie.
+
```
Tratcher, line 123: This isn't giving much context on where these calls should be put. Do we have sample code to go with it?

aspnetcore/security/authentication/social/additional-claims.md
```diff
@@ -105,6 +108,19 @@ Users can define custom actions by deriving from <xref:Microsoft.AspNetCore.Auth
 
 For more information, see <xref:Microsoft.AspNetCore.Authentication.OAuth.Claims>.
 
+## Add claims to users
+
+Claims are added to the user on first registration, not on sign in. If additional claims are enabled in an app after a user registers to use the app, call [SignInManager.RefreshSignInAsync](xref:Microsoft.AspNetCore.Identity.SignInManager%601) on a user to force the generation of a new authentication cookie.
```
Tratcher, line 113:
```suggestion
Claims are copied from external providers to the user database on first registration, not on sign in. If additional claims are enabled in an app after a user registers to use the app, call [SignInManager.RefreshSignInAsync](xref:Microsoft.AspNetCore.Identity.SignInManager%601) on a user to force the generation of a new authentication cookie.
```

aspnetcore/security/authentication/social/additional-claims.md
guardrex: That's the question that I'm asking :point_up:. We don't cover these in the doc set. AFAIK, there's no sample for this scenario. 

It was discussed at https://github.com/dotnet/AspNetCore.Docs/pull/15491#discussion_r348664977, and that's where we left it. I'd need more guidance on how to address two scenarios that come to mind:

* The dev wants a utility method that they can run on all of the users to add/subtract claim(s) in the Identity dB.
* The devs wants `IUserClaimsPrincipalFactory` (?) to run automatically the next time a user signs in and updates them with the added claim(s).

... and that's if I'm even understanding the problem correctly. When we discussed this back in the day, I added claims ... set up the app ... but nothing was happening for my test user. You told me that the Identity dB needs to be updated (and currently signed in users). At the time, my quick resolution was merely to delete the test user and recreate them, which is fine for dev/test, not for production.

guardrex: > the app could on subsequent external logins call replace claim with the updated value of a claim. ... It shouldn't be necessary for them to have a custom principal factory for this.

But, that's not the same as adding a new claim from the provider for all users, right?

guardrex: I'm going to try again here now. Reminder on the scenario ...

1. App is created with user's stored locally.
2. Follow our guidance add an external social IP with some claims set up for social IP information about users.
3. Run the app ... register a user. Confirm that the user is able to auth and that the user has claims that were set up for the app.
4. The app is then modified ... the app is coded to set a **_new claim_** from Google about users (i.e., *Map user data keys and create claims* is followed to add a new one to the app).
5. Run the app again. Sign the test user in. Initially, the user **_doesn't get this new claim_** because "**_Claims are added to the user on first registration, not on signin._**"

That's where we left it. Thus far, your remarks are still a bit *ungrokable* 😵 to me, but I'll hack around a bit with your remarks in a test app to see if I can figure out *where* these calls go. Again, I think there are two potential solutions ...

* The dev runs a utility method that does ✨ *magic things* ✨ to all registered users in one shot to add the new claim. The next time any user signs in, *Bingo!* ... they have the new claim.
* The next time a user signs in, some ✨ *magic things* ✨ happen automatically for just that user to add the new claim. *Bingo!* ... they have the new claim.

Sorry if I've overly-simplified/overly-complicated the scenario. We never used social IPs at my last employer with .NET (pre-Core, of course), and claims-based Identity (IIRC) was only used for B2C-based apps during my final years there, which was *Identity Dreamy*&trade; :smile: to use. I recall my auth stress level dropped quite a bit using B2C.

guardrex: Rubber 🦆 working through this. Adding a few **_note to self_** cross-links.

This blog post covers what I'm facing ...

https://haacked.com/archive/2019/07/16/external-claims/

... and following the breadcrumbs ...

https://github.com/aspnet/Identity/issues/628

... leads to ...

https://github.com/dotnet/aspnetcore/issues/5771

... which was mentioned again on my issue for this PR ... and then trying to reach this ...

https://github.com/aspnet/AuthSamples/tree/dev/samples/Identity.ExternalClaims

... which is :boom: now but I think I can see here ...

https://github.com/aspnet/AuthSamples/commit/404105bd191f2e973d4befb668f1310d2fd82701

guardrex: Ok ..... rubber 🦆 says we're getting closer ...

```csharp
public async Task<IActionResult> OnGetCallbackAsync(string returnUrl = null, string remoteError = null)
{
    ...

    if (result.Succeeded)
    {
        _logger.LogInformation("{Name} logged in with {LoginProvider} provider.", info.Principal.Identity.Name, info.LoginProvider);

        if (info.Principal.HasClaim(c => c.Type == "urn:google:picture"))
        {
            _logger.LogInformation($"########### {info.Principal.FindFirst("urn:google:picture").Value}");

            var user = await _userManager.FindByLoginAsync(info.LoginProvider, info.ProviderKey);
            var claims = await _userManager.GetClaimsAsync(user);
            var claim = claims.FirstOrDefault(c => c.Type == "urn:google:picture");
            await _userManager.ReplaceClaimAsync(user, claim, info.Principal.FindFirst("urn:google:picture"));

            await _signInManager.RefreshSignInAsync(user);
        }

        return LocalRedirect(returnUrl);
    }
    ...
}
```

guardrex: @HaoK ... If I'm hack'in anything good here, **_A approach_** (not necessarily *THE approach* to show) is to ...

1. Run a migration on the dB and add a claim entry to `AspNetUserClaims` for every user. In my testing use case for adding the user's Google avatar image URL to the app, I can add the following to the dB for each user ...
   * UserId: The UserId (GUID key) for the user in the dB.
   * ClaimType: `urn:google:picture`
   * ClaimValue: `https://localhost:5001/foilhead.jpg` (a dummy value given that if this is left `NULL` in the dB it leads to a :boom: when a user signs into the app)

1. Drop an image for that into the app's `wwwroot` folder just in case a claim with a real Google image URL can't be assigned to the value of the claim.
 
   `wwwroot/foilhead.jpg`:

     foilhead

   *btw - Yeah, it was a hell of a party.* :beers:

1. Use that code in my last comment :point_up:.

   Of course, it also uses the current guidance for [map a key and make a claim](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/social/additional-claims?view=aspnetcore-5.0#map-user-data-keys-and-create-claims) that we have. In my test case ...

   ```csharp
   options.ClaimActions.MapJsonKey("urn:google:picture", "picture", "url");
   ```

   ... and the bits we show for `OnPostConfirmationAsync` ...

   ```csharp
   if (info.Principal.HasClaim(c => c.Type == "urn:google:picture"))
   {
       await _userManager.AddClaimAsync(user, info.Principal.FindFirst("urn:google:picture"));
   }
   ```

This approach seems to work. Is it sane :+1: ... or totally 😵 **_INSANE_** 😵?

HaoK: Right, I would think its more reasonable to 'backfill' update the user's claim on login(option 2) over (periodically) synchronizing claims via a migration/import.

And the code you wrote looks reasonable, one minor thing is maybe you would also want to check if the claim is the same before replacing/refreshing the sign in.  So to summarize what that's doing, is on long, you check to see if any claims need to be updated, and do so then (and refresh the cookie if needed)  There's always this handshake to switch from the external login cookie to the real application cookie, so it doesn't seem unreasonable to look at the claims during that process.  @blowdart @Tratcher thoughts?

guardrex: I think I follow what you're suggesting. I think that I see the next step for the 'backfill' part and avoiding replace/refresh when there's no work to do for a given user that already has it and with a correct value. I'll work it further forward on Wednesday morning.

Tratcher: Refreshing/backfilling claims on login makes sense. Can that be done generically for a set of claims rather than for one claim at a time?

guardrex: Ok ... put an :eye: on this. This covers every scenario AFAICT in testing (but it might lack code economy and/or perf) ...

* Claim is present for the principal and ...

  * The user has no claim at all: Do a `AddClaimAsync` + `RefreshSignInAsync`.
  * The user has a claim and the values of the two match: Do a `ReplaceClaimAsync` + `RefreshSignInAsync`.
  * The user and principal claim values match: 🛑 **Drop out** ... just use what was loaded from the user claim.

* Claim not present in principal and ...

  * The user has a claim stored that they can use: 🛑 **Drop out** ... just use what was loaded from the user claim.
  * Both no principal claim and no existing user claim: Give them a fake headshot claim (`headshot.jpg`) ...

    headshot2

    *I'll let the Baltimore Police Department arrest report and my attorney explain that one!* :smile:

**WRT what we'll eventually show**: If it's possible to generalize and cover a group of claims, that's cool. If so tho, it would be best to provide the full code to me because I need to get back to Blazor issues asap, where I have a *ton of work* ⛰️⛏️ to do. Otherwise, my plan is to show+describe an example of adding one specific claim, probably this one, and leaving it to the dev to take it further.

```csharp
var result = await _signInManager.ExternalLoginSignInAsync(info.LoginProvider, info.ProviderKey, isPersistent: false, bypassTwoFactor : true);
if (result.Succeeded)
{
    _logger.LogInformation("{Name} logged in with {LoginProvider} provider.", info.Principal.Identity.Name, info.LoginProvider);

    var user = await _userManager.FindByLoginAsync(info.LoginProvider, info.ProviderKey);
    var userClaims = await _userManager.GetClaimsAsync(user);
    var userPictureClaim = userClaims.FirstOrDefault(c => c.Type == "urn:google:picture");

    if (info.Principal.HasClaim(c => c.Type == "urn:google:picture"))
    {
        var principalPictureClaim = info.Principal.FindFirst("urn:google:picture");

        if (userPictureClaim == null)
        {
            await _userManager.AddClaimAsync(user, new Claim("urn:google:picture", principalPictureClaim.Value));
            await _signInManager.RefreshSignInAsync(user);
        }
        else if (userPictureClaim.Value != principalPictureClaim.Value)
        {
            await _userManager.ReplaceClaimAsync(user, userPictureClaim, principalPictureClaim);
            await _signInManager.RefreshSignInAsync(user);
        }
    }
    else if (userPictureClaim == null)
    {
        await _userManager.AddClaimAsync(user, new Claim("urn:google:picture", "https://localhost:5001/headshot.jpg"));
        await _signInManager.RefreshSignInAsync(user);
    }

    return LocalRedirect(returnUrl);
}
```

guardrex:
@HaoK @Tratcher ... The latest commits place yesterday's 🙈 *RexHacks*&trade; 🙈 into the section, so you can see the prototype in the diff and internal review topic.

It runs off of an added claims dictionary with default claim values. It seeks to address your, @Tratcher, question about how it *might* be done in a general way for a number of added claims. Idk if there's a fancy, more economical way using framework features and/or with a better algorithm.

[Internal Review Topic (links to section)](https://review.docs.microsoft.com/en-us/aspnet/core/security/authentication/social/additional-claims?view=aspnetcore-2.1&branch=pr-en-us-21345#add-and-update-user-claims)

aspnetcore/security/authentication/social/additional-claims.md
```diff
@@ -105,6 +108,122 @@ Users can define custom actions by deriving from <xref:Microsoft.AspNetCore.Auth
...
+
+Claims are copied from external providers to the user database on first registration, not on sign in. If additional claims are enabled in an app after a user registers to use the app, call [SignInManager.RefreshSignInAsync](xref:Microsoft.AspNetCore.Identity.SignInManager%601) on a user to force the generation of a new authentication cookie.
+
+In the Development environment working with test user accounts, you can simply delete and recreate the user account. For production systems, new claims added to the app can be backfilled into user accounts. After [scaffolding the `ExternalLogin` page](xref:security/authentication/scaffold-identity) into the app at `Areas/Pages/Identity/Account/Manage`, add the following code to the `ExternalLoginModel` in the `ExternalLogin.cshtml.cs` file.
+
+Add a dictionary of added claims. Use the dictionary keys to hold the claim types, and use the values to hold a default value. Add the following line to the top of the class. The following example assumes that one claim is added for the user's Google picture with a generic headshot image as the default value:
+
+```csharp
+private readonly IDictionary<string, string> _addedClaims =
```
Tratcher, line 120: Generalizing from add to add-or-update.
```suggestion
private readonly IDictionary<string, string> _claimsToSync = 
```

aspnetcore/security/authentication/social/additional-claims.md
```diff
@@ -105,6 +108,122 @@ Users can define custom actions by deriving from <xref:Microsoft.AspNetCore.Auth
...
+public async Task<IActionResult> OnGetCallbackAsync(
+    string returnUrl = null, string remoteError = null)
+{
+    returnUrl = returnUrl ?? Url.Content("~/");
+    if (remoteError != null)
+    {
+        ErrorMessage = $"Error from external provider: {remoteError}";
+        return RedirectToPage("./Login", new {ReturnUrl = returnUrl });
+    }
+    var info = await _signInManager.GetExternalLoginInfoAsync();
```
Tratcher, line 139:
```suggestion
    var externalLogin = await _signInManager.GetExternalLoginInfoAsync();
```

aspnetcore/security/authentication/social/additional-claims.md
```diff
@@ -105,6 +108,122 @@ Users can define custom actions by deriving from <xref:Microsoft.AspNetCore.Auth
...
+            bool refreshSignIn = false;
+
+            foreach (var addedClaim in _addedClaims)
+            {
+                var userClaim = userClaims
+                    .FirstOrDefault(c => c.Type == addedClaim.Key);
+
+                if (info.Principal.HasClaim(c => c.Type == addedClaim.Key))
+                {
+                    var principalClaim = info.Principal.FindFirst(addedClaim.Key);
```
Tratcher, line 169:
```suggestion
                    var externalClaim = info.Principal.FindFirst(addedClaim.Key);
```

aspnetcore/security/authentication/social/additional-claims.md
```diff
@@ -105,6 +108,122 @@ Users can define custom actions by deriving from <xref:Microsoft.AspNetCore.Auth
...
+                    else if (userClaim.Value != principalClaim.Value)
+                    {
+                        await _userManager
+                            .ReplaceClaimAsync(user, userClaim, principalClaim);
+                        refreshSignIn = true;
+                    }
+                }
+                else if (userClaim == null)
+                {
+                    await _userManager.AddClaimAsync(user, new Claim(addedClaim.Key,
```
Tratcher, line 186:
```suggestion
                    // Fill with a default value
                    await _userManager.AddClaimAsync(user, new Claim(addedClaim.Key, 
```

Tratcher (commented):
Looks promising to me.

aspnetcore/security/authentication/social/additional-claims.md
```diff
@@ -105,6 +108,122 @@ Users can define custom actions by deriving from <xref:Microsoft.AspNetCore.Auth
...
+            {
+                Email = info.Principal.FindFirstValue(ClaimTypes.Email)
+            };
+        }
+        return Page();
+    }
+}
+```
+
+A similar approach is taken when claims change while a user is signed in but a backfill step isn't required. To update a user's claims, call the following on the user:
```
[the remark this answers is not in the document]
guardrex: But this line is meant to cover the **_not adding_** claim(s) scenario, where only replacing claim(s) and refreshing are required. I've angled this toward the dev refactoring the example to remove the `userClaim == null` pieces. I admit that I'm trying to punt 🏈 out of showing it. I think most devs can work it out easily enough from here. That's just my *guess* tho.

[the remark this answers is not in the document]
Tratcher: *shrug* if the above code works and then they decide to remove part of it they're not using, that seems fine for them to sort out.

aspnetcore/security/authentication/social/additional-claims.md
guardrex: @Tratcher  ... This is template driven, right? ...

https://github.com/dotnet/aspnetcore/blob/main/src/Identity/UI/src/Areas/Identity/Pages/V4/Account/ExternalLogin.cshtml.cs#L137

U sure that we should change it?

HaoK: Yeah I'd match the scaffolded code over clarity

Tratcher: Darn. Maybe we should change the template...

HaoK: Yeah that's fair, feel free to file an issue or PR for that

guardrex: Change it back to `externalLogin`?

If you know that's what you'll go with, I'd rather not come back to this given the workload 🏃😅.

Tratcher: Don't worry about it for now, any change wouldn't happen until 6.0.

HaoK: Yeah this looks great to me,!

guardrex: I'll merge this when the build system comes back online.

PR #21350, opened 2021-01-28
Fix typo in routing.md

ericmutta:
The sentence:

> ...is an example of a *conventional routing*.

Should read:

> is an example of a *conventional route*.


guardrex (approved):
Thanks @ericmutta! :rocket:

PR #21372, opened 2021-01-30
Fix typo in application-model.md

ericmutta:
The sentence:

> ...application model INCLUDE both abstract interfaces and concrete implementation classes...

Should read:

> ...application model INCLUDES both abstract interfaces and concrete implementation classes...


guardrex (approved):
Thanks @ericmutta! ⛵

PR #21373, opened 2021-01-30
Update hosted-with-identity-server.md

vitalragaz:
Removed unnecessary code line that doesn't relate to the description. The Authentication middleware is already explained before.


guardrex (approved):
Thanks @vitalragaz! :rocket:

PR #21374, opened 2021-01-30
Minor (I think) confliction of "authentication" & "authorization" usage

GregTrevellick:
There's a possible (definate?) conflict of text here - user will either see `You're not authorized` or `Authentication in progress` - one would expect both phrases to refer to "authentication" or both phrases to refer to "authorization".

I believe the latter is correct, hence this PR, but there are potentially other similar confusions of "authentication" / "authorization" in this markdown page, which aren't covered by this PR.

### DIFF
--- a/aspnetcore/blazor/security/index.md
+++ b/aspnetcore/blazor/security/index.md
@@ -392,8 +392,8 @@ In the default Blazor Server project template, the `App` component (`App.razor`)
                     <p>You may need to log in as a different user.</p>
                 </NotAuthorized>
                 <Authorizing>
-                    <h1>Authentication in progress</h1>
-                    <p>Only visible while authentication is in progress.</p>
+                    <h1>Authorization in progress</h1>
+                    <p>Only visible while authorization is in progress.</p>
                 </Authorizing>
             </AuthorizeRouteView>
         </Found>


guardrex:
Hello @GregTrevellick ... It's best to open an issue using the **This page** feedback button+form at the bottom of the English-US (`en-us`) topic, but I'll take a look at this and respond tomorrow morning.

GregTrevellick:
Sure, no probs, I did wonder about doing that from the outset funnily enough, will do tomorrow 😃

guardrex:
We usually discuss a potentially more complicated changes to a topic before receiving a PR. Changes that go directly to a PR are typically just patch fixes, such as a misspelled word or an incorrect C# statement. That way for content changes that are more conceptual in nature, the dev doesn't end up doing work that we don't take and ultimately feel like they wasted time trying to help improve the docs. Fortunately in this case, the change is small.

I'll get back to you in the morning on this. ☕

guardrex:
I'm up. ☕ ... and the 🐈 let me sleep in to 5:30am for a change. Gee ... thanks! ... *rotten 😼!*

The language used in the `<Authorizing>` element is currently by-design because the rate limiting step for authorization isn't usually the *authorization* aspect (authenticated or not) ... that's really *FAST* 🏃. It's the *authentication* step for an unauthenticated user that takes long enough that showing something is worthwhile in most cases. It takes mere tiny fractions of a second to compute authz for an authenticated user in most cases. That's why it was written this way. Technically *yes*, they're being authorized, but Steve implies that most users are really waiting on authentication to occur. He's made the markup of the property be *more specific for most users*.

I'm not opposed to the change. The [description of this property](https://docs.microsoft.com/dotnet/api/microsoft.aspnetcore.components.authorization.authorizerouteview.authorizing) is ...

> The content that will be displayed while asynchronous authorization is in progress.

Practically, it's usually going to display during authentication. However, it invites confusion in its present form.

The bullet point above this for "allows the app to specify custom content if" reads ...

> Asynchronous authentication is in progress.

... so that covers authn for the scenario.

I searched the doc set, and this is the only place this property is shown/described. We don't maintain any sample apps that host a version of this component with this property set. Therefore, this is likely the only change that we need.

WRT ...

> there are potentially other similar confusions

Those would have to be explicitly called out until I reach this topic for an overhaul on https://github.com/dotnet/AspNetCore.Docs/issues/19286. I'm working on those passes now, and I hope to reach this topic by mid-year.

GregTrevellick:
Thanks Luke, just checking, is it still worthwhile me raising an issue? Happy to do so, but suspect already covered by the existing wider topic.

guardrex:
> still worthwhile me raising an issue

Only if you think that there are other potential changes that aren't patches.

PR #21375, opened 2021-01-30
Fix WARNING box format in filters.md

ericmutta:
Fix format for the WARNING box with the following sentence::

> Only configure IFilterFactory.IsReusable to return true if the source of the filters...

Hopefully this will fix the broken rendering on the [live site](https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/filters?view=aspnetcore-5.0#ifilterfactory):

image


guardrex (approved):
Thanks @ericmutta! :rocket:

ericmutta:
Thanks @guardrex I see there's no rest for the weary (it's 2am here and I am still up trying to assimilate the sum of all ASP.NET knowledge :grin:)

PR #21388, opened 2021-02-01
Expand nullable reference types

pranavkm:

### DIFF
--- a/aspnetcore/mvc/models/validation.md
+++ b/aspnetcore/mvc/models/validation.md
@@ -80,7 +80,7 @@ To find out which parameters are passed to `String.Format` for a particular attr
 
 ## Non-nullable reference types and [Required] attribute
 
-The validation system treats non-nullable parameters or bound properties as if they had a `[Required]` attribute. By enabling `Nullable` contexts in your application (see https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references#nullable-contexts), MVC will implicitly start validating non-nullable properties or parameters as if they had been attributed with the Required attribute. Consider,
+The validation system treats non-nullable parameters or bound properties as if they had a `[Required]` attribute. By [enabling `Nullable` contexts](/dotnet/csharp/nullable-references#nullable-contexts), MVC implicitly starts validating non-nullable properties or parameters as if they had been attributed with the `[Required]` attribute. Consider the following code:
 
 ```csharp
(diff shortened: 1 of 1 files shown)


aspnetcore/mvc/models/validation.md
```diff
@@ -76,9 +76,29 @@ When applied to a `Name` property, the error message created by the preceding co
 
 To find out which parameters are passed to `String.Format` for a particular attribute's error message, see the [DataAnnotations source code](https://github.com/dotnet/runtime/tree/master/src/libraries/System.ComponentModel.Annotations/src/System/ComponentModel/DataAnnotations).
 
-## [Required] attribute
+::: moniker range=">= aspnetcore-3.0"
+
+## Non-nullable reference types and [Required] attribute
+
+The validation system treats non-nullable parameters or bound properties as if they had a `[Required]` attribute. By enabling `Nullable` contexts in your application (see https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references#nullable-contexts), MVC will implicitly start validating non-nullable properties or parameters as if they had been attributed with the Required attribute. Consider,
```
Rick-Anderson, line 83:
```suggestion
The validation system treats non-nullable parameters or bound properties as if they had a `[Required]` attribute. By [enabling `Nullable` contexts](/dotnet/csharp/nullable-references#nullable-contexts), MVC implicitly starts validating non-nullable properties or parameters as if they had been attributed with the `[Required]` attribute. Consider the following code:
```

aspnetcore/mvc/models/validation.md
```diff
@@ -76,9 +76,29 @@ When applied to a `Name` property, the error message created by the preceding co
...
+
+```csharp
+public class Person
+{
+    public string Name { get; set; }
+}
+```
+
+If the application was built with `<Nullable>enable</Nullable>`, a missing value for `Name` in a JSON or form post would result in a validation error. Use a nullable reference type to allow null or missing values to be specified for the `Name` property:
```
Rick-Anderson, line 92:
```suggestion
If the app was built with `<Nullable>enable</Nullable>`, a missing value for `Name` in a JSON or form post results in a validation error. Use a nullable reference type to allow null or missing values to be specified for the `Name` property:
```

Rick-Anderson (approved):
Avoid future case.

PR #21480, opened 2021-02-08
Configure Trimmer UE pass

guardrex:
Addresses #19286

[Internal Review Topic](#)

* Coalesce the coverage in the Trimmer topic. Only cross-link from the WASM Performance topic.
* There's either missing coverage or a misunderstanding on what the following line means ...

  > The Trimmer is also unable to react to an app's dynamic behavior at runtime.

  Does that only refer to **_reflection_**, or is that referring to **_other dynamic behavior_** that the topic doesn't describe? If it does refer to other dynamic behavior, should the topic list the behavior and provide guidance on dealing with it? If it's just referring to reflection, we can remove that line. :ear:
* My recommendation is to let the IL Trimming topic take care of the implementation details (i.e., this draft allows the external doc cover the `<PublishTrimmed>` property). However, I propose in this draft that we briefly list what the dev will find over there.
* Is there anything else that you'd like to cover?

### DIFF
--- a/aspnetcore/blazor/host-and-deploy/configure-trimmer.md
+++ b/aspnetcore/blazor/host-and-deploy/configure-trimmer.md
@@ -11,13 +11,7 @@ uid: blazor/host-and-deploy/configure-trimmer
 ---
 # Configure the Trimmer for ASP.NET Core Blazor
 
-Blazor WebAssembly performs [Intermediate Language (IL)](/dotnet/standard/managed-code#intermediate-language--execution) trimming to reduce the size of the published output.
-
-By default, trimming occurs when publishing an app in Release configuration using the [`dotnet publish`](/dotnet/core/tools/dotnet-publish) command with the [-c|--configuration](/dotnet/core/tools/dotnet-publish#options) option set to `Release`:
-
-```dotnetcli
-dotnet publish -c Release
-```
+Blazor WebAssembly performs [Intermediate Language (IL)](/dotnet/standard/managed-code#intermediate-language--execution) trimming to reduce the size of the published output. By default, trimming occurs when publishing an app.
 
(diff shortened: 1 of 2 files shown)


aspnetcore/blazor/host-and-deploy/configure-trimmer.md
```diff
@@ -5,28 +5,31 @@ description: Learn how to control the Intermediate Language (IL) Linker (Trimmer
...
 uid: blazor/host-and-deploy/configure-trimmer
 ---
 # Configure the Trimmer for ASP.NET Core Blazor
 
-By [Pranav Krishnamoorthy](https://github.com/pranavkm)
-
 Blazor WebAssembly performs [Intermediate Language (IL)](/dotnet/standard/managed-code#intermediate-language--execution) trimming to reduce the size of the published output.
 
-Trimming an app optimizes for size but may have detrimental effects. Apps that use reflection or related dynamic features may break when trimmed because the trimmer doesn't know about dynamic behavior and can't determine in general which types are required for reflection at runtime. To trim such apps, the trimmer must be informed about any types required by reflection in the code and in packages or frameworks that the app depends on.
+By default, trimming occurs when publishing an app in Release configuration using the [`dotnet publish`](/dotnet/core/tools/dotnet-publish) command with the [-c|--configuration](/dotnet/core/tools/dotnet-publish#options) option set to `Release`:
```
pranavkm, line 16: Trimming occurs during publish for all configurations. The "Release" specific behavior only happened in 3.2

pranavkm (approved):

pranavkm:
> Does that only refer to reflection, or is that referring to other dynamic behavior that the topic doesn't describe?

Generally reflection and reflection-like things. Basically it looks for references to types or methods and code and removes them if it cannot determine they're being used. For example, Blazor uses `Activactor.CreateInstance` by default to new up components and reflection to assign component parameters. Because this is not clear, running the trimmer on an assembly that contains components will remove constructors (including the default one) and property setters which will fail at runtime.

Issue #15627, opened 2019-11-09
Global Blazor and SignalR no-loc


guardrex:
Remove `no-loc: [Blazor, SignalR]` from topics when a solution is available to globally apply no-loc behavior  ...

```json
"no-loc": "[Blazor, SignalR]"
```

If possible, including the "blazor"-named files:

* *blazor.webassembly.js*
* *blazor.server.js*

The files are currently code fenced to avoid localizing the "blazor" part of the file names. Discussion was on https://github.com/aspnet/AspNetCore.Docs/pull/16123.

If including the files isn't possible at the time the global approach is adopted, re-purpose this issue and leave it open for the work.

guardrex:
@JasonCard ... checking in with you on this scenario.

JasonCard:
@guardrex The global solution won't be docfx.json, but we are currently discussing a possible solution that will work with the MT engine. Currently following up and will update when I know more.

guardrex:
@JasonCard Is there any update on this one ... a global approach to setting no-loc words/phrases in one spot for the whole repo? If you think it will take quite a while (or never be a thing), then I may as well close this issue.

JasonCard:
@guardrex I was told that MSR team is planning on a dynamic dictionary feature for the MT engine, but there is no ETA and it may take quite awhile. So I agree, we may want to go ahead and close this issue.

guardrex:
Cool ... thanks. I have a feeling that a non-Blazor issue will be the way to go for this anyway. Now that we have the `no-loc` metadata everywhere, it's more a general thing. The team can tend to it when the time comes.

PR #21495, opened 2021-02-09
Prevent Blazor from being localized

Youssef1313:
Contributes to https://github.com/dotnet/docs/issues/22731

### DIFF
--- a/aspnetcore/blazor/includes/state-container.md
+++ b/aspnetcore/blazor/includes/state-container.md
@@ -1,3 +1,7 @@
+---
+no-loc: [Blazor, Blazor Server, Blazor WebAssembly]
+---
+
 Nested components typically bind data using *chained bind* as described in <xref:blazor/components/data-binding>. Nested and un-nested components can share access to data using a registered in-memory state container. A custom state container class can use an assignable <xref:System.Action> to notify components in different parts of the app of state changes. In the following example:
 
 * A pair of components uses a state container to track a property.
--- a/aspnetcore/blazor/includes/state-container.md
+++ b/aspnetcore/blazor/includes/state-container.md
@@ -1,5 +1,5 @@
 ---
-no-loc: [Blazor, Blazor Server, Blazor WebAssembly]
+no-loc: [Blazor, "Blazor Server", "Blazor WebAssembly"]
 ---
 
 Nested components typically bind data using *chained bind* as described in <xref:blazor/components/data-binding>. Nested and un-nested components can share access to data using a registered in-memory state container. A custom state container class can use an assignable <xref:System.Action> to notify components in different parts of the app of state changes. In the following example:


guardrex:
**_Confirmed_** ...

https://docs.microsoft.com/fr-fr/aspnet/core/blazor/state-management?view=aspnetcore-5.0&pivots=webassembly#in-memory-state-container-service-wasm

Thanks @Youssef1313 ... let me check one thing before we merge this because we may have a more general coverage problem that I'm not fully clear on. Actually, you may know and can respond .......

@JasonCard ... I was about to update the `no-loc` string here to match our normal one ...

```
no-loc: [appsettings.json, "ASP.NET Core Identity", cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
```

... but I have to ask the general question: Should we be doing this for **_all of our INCLUDES_** files? If so, I'll close this PR and open an issue to address it everywhere across the docs.

btw- @JasonCard ... Also, the rendered pivot box at the top isn't honoring the `no-loc` metadata.
"Choose a Blazor hosting model" --> "Choisir un modèle d’hébergement éblouissant" ... and the names in the buttons are incorrect.

guardrex:
@Youssef1313 ... I don't think we'll be using the PR for the change. This is probably going to have to happen **_everywhere_** here. Let me get confirmation from @JasonCard and then I'll proceed from there.

Youssef1313:
@guardrex Feel free to close and open an issue or PR to address this everywhere in the repository.

I'd personally like a feature that allows specifying "no-loc" once in [docfx.json](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/docfx.json). Just like any other metadata being defined globally. For example:

https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/docfx.json#L69-L91

This is currently not supported for no-loc AFAIK. But it will make things much easier to manage.

guardrex:
@Youssef1313 ... I think that they're working on it ...

https://github.com/dotnet/AspNetCore.Docs/issues/15627#issuecomment-590565860

JasonCard:
@Youssef1313 @guardrex 

> ... but I have to ask the general question: Should we be doing this for all of our INCLUDES files? It seems that we should. If so, I'll close this PR and open an issue to address it everywhere across the docs.

Yes, this is currently needed for each file as we don't yet have a global solution to block terms from being localized in all files.

> btw- @JasonCard ... Also, the rendered pivot box at the top isn't honoring the no-loc metadata.
"Choose a Blazor hosting model" --> "Choisir un modèle d’hébergement éblouissant" ... and the names in the buttons are incorrect.

Thank you for pointing this out. I found that [zone-pivot-groups.yml](https://github.com/dotnet/AspNetCore.Docs/blob/live/aspnetcore/zone-pivot-groups.yml) file was being machine translated, but this file should not have been localized in the first place, so I've corrected this and it should be falling back to English shortly.

> I'd personally like a feature that allows specifying "no-loc" once in docfx.json. Just like any other metadata being defined globally.

This idea has been proposed, but was not a feasible solution. As docfx.json file is used for publishing, but not used to determine which text should be localized or not. The "no-loc" metadata is used by our loc platform when extracting localizable text from the source files and it's not feasible for it to reference docfx.json file. 
One pending global solution is a dynamic dictionary feature for our Machine Translation engine, but there is no ETA and it may take quite awhile.

guardrex:
Thanks @JasonCard.

Thanks @Youssef1313 ... I'll open an issue and address this across the INCLUDES.

PR #21506, opened 2021-02-09
Add no-loc metadata to INCLUDES

### LINKED ISSUE #21501: Add no-loc to INCLUDES
Per https://github.com/dotnet/AspNetCore.Docs/pull/21495#issuecomment-776197842, we should have our no-loc string in the INCLUDES files just above the text. I suggest no blank line between the `---` of the metadata and the first line of content as we normally do.

@Rick-Anderson ... I've assigned to myself, and I have some time this afternoon to knock this out. However, feel free to re-assign to yourself or another 🐈.

```
---
no-loc: [appsettings.json, "ASP.NET Core Identity", cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
---
```

Thanks @Youssef1313 🎸 for calling this out.

guardrex:
Fixes #21501

For INCLUDES that had an existing blank line at the top, I assumed that the line is there by-design and left it alone. The other files start with content, so the metadata is added immediately above it (i.e., no blank line between the `---` and the content).

Thanks @Youssef1313! 🎸

### DIFF
--- a/aspnetcore/blazor/includes/prefer-exact-matches.md
+++ b/aspnetcore/blazor/includes/prefer-exact-matches.md
@@ -1,3 +1,6 @@
+---
+no-loc: [appsettings.json, "ASP.NET Core Identity", cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
+---
 ::: moniker range="= aspnetcore-5.0"
 
 > [!NOTE]
--- a/aspnetcore/blazor/includes/prerendering.md
+++ b/aspnetcore/blazor/includes/prerendering.md
@@ -1,3 +1,6 @@
+---
+no-loc: [appsettings.json, "ASP.NET Core Identity", cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
+---
 While a Blazor Server app is prerendering, certain actions, such as calling into JavaScript, aren't possible because a connection with the browser hasn't been established. Components may need to render differently when prerendered.
 
 To delay JavaScript interop calls until after the connection with the browser is established, you can use the [OnAfterRenderAsync component lifecycle event](xref:blazor/components/lifecycle#after-component-render). This event is only called after the app is fully rendered and the client connection is established.
--- a/aspnetcore/blazor/includes/security/additional-scopes-standalone-AAD.md
+++ b/aspnetcore/blazor/includes/security/additional-scopes-standalone-AAD.md
@@ -1,3 +1,6 @@
+---
+no-loc: [appsettings.json, "ASP.NET Core Identity", cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
+---
 Add a <xref:Microsoft.Authentication.WebAssembly.Msal.Models.MsalProviderOptions> for `User.Read` permission with <xref:Microsoft.Authentication.WebAssembly.Msal.Models.MsalProviderOptions.DefaultAccessTokenScopes>:
 
 ```csharp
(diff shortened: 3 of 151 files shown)


JasonCard:
@guardrex I'm sorry, but after checking internally, we don't think that no-loc *metadata* will work in INCLUDES files, because those files usually don’t have any yaml metadata. We could maybe test it with one INCLUDES file to see if works or not to be sure. If it doesn't work, it seems that the inline syntax like *:::no-loc text="String":::* should work, but you would have to search for and apply it to each individual instances... 
In order to make the metadata work, we may have to request a feature for it, but it's not likely to be approved, because we're working on a new loc pipeline/process that would resolve this issue. In the new loc pipeline, we would extract the localizable text from the published HTML rather than the source MD files in GitHub repo, so content from INCLUDES files would be localized based on the published HTML. However, ETA for this new loc process is tentatively end of FY21 Q4.

guardrex:
@JasonCard I'll perform a test on one INCLUDE file and see what happens and report back.

>  but you would have to search for and apply it to each individual instances...

I assume you mean that it goes around the exact text, but would you show an explicit example of that for the following example ...

> ... blah blah blah Blazor Server blah ...

JasonCard:
Yes, that's correct. There's also an example and note on special characters [here](https://review.docs.microsoft.com/en-us/help/contribute/markdown-reference?branch=master#non-localized-strings).

guardrex:
Ok, thanks.

I don't have access, so I can't see endpoints for links like that. Can you paste the note on special chars in here?

JasonCard:
Np, I've copied the entire section on the 'no-loc' feature below (included the part about yaml metadata for your reference):

## Non-localized strings

You can use the custom `no-loc` Markdown extension to identify strings of content that you would like the localization process to ignore.

All strings called out will be case-sensitive; that is, the string must match exactly to be ignored for localization.

To mark an individual string as non-localizable, use this syntax:

```markdown
:::no-loc text="String":::
```

For example, in the following, only the single instance of `Document` will be ignored during the localization process:

```markdown
# Heading 1 of the Document

Markdown content within the :::no-loc text="Document":::.  The are multiple instances of Document, document, and documents.
```

> [!NOTE]
> Use `\` to escape special characters:
> ```markdown
> Lorem :::no-loc text="Find a \"Quotation\""::: Ipsum.
> ```

You can also use metadata in the YAML header to mark all instances of a string within the current Markdown file as non-localizable:

```yml
author: cillroy
no-loc: [Global, Strings, to be, Ignored]
```

> [!NOTE]
> The no-loc metadata is not supported as global metadata in *docfx.json* file. The localization pipeline doesn't read the *docfx.json* file, so the no-loc metadata must be added into each individual source file.

In the following example, both in the metadata `title` and the Markdown header the word `Document` will be ignored during the localization process.

In the metadata `description` and the Markdown main content the word `document` is localized, because it does not start with a capital `D`.

```markdown
---
title: Title of the Document
author: author-name
description: Description for the document
no-loc: [Title, Document]
---
# Heading 1 of the Document

Markdown content within the document.
```

guardrex:
UPDATE ...

The test is underway on https://github.com/dotnet/AspNetCore.Docs/pull/21511. I'll report back when I confirm from the French doc what happened ... *Bonne chance!* 🤞🍀

guardrex:
# :+1:

@JasonCard ... The French version of the topic has arrived, and no-loc metadata in the INCLUDE file worked. :beers:

**Test PR**: https://github.com/dotnet/AspNetCore.Docs/pull/21511
**Live Topic (French localization, links to section)**: https://docs.microsoft.com/fr-fr/aspnet/core/blazor/state-management?view=aspnetcore-5.0&pivots=webassembly#in-memory-state-container-service-wasm

## BEFORE

before

## AFTER

Capture

## Merge?

Do you concur that I can proceed to merge this PR?

cc: @Rick-Anderson @scottaddie @wadepickett

JasonCard:
Great! Yes, thank you for this successful test and I concur :)

JasonCard (approved):

guardrex:
Thanks @JasonCard. 

@Rick-Anderson ... I'll wait until you concur before merging.

Rick-Anderson (approved):

PR #21555, opened 2021-02-15
Component parameter assignment guidance

### LINKED ISSUE #21552: Very hard to interpret content about component parameters
The docs about component parameters include the following:

```
By convention, an attribute value that consists of C# code is assigned to a parameter using Razor's reserved @ symbol:

- Parent field or property: Title="@{FIELD OR PROPERTY}, where the placeholder {FIELD OR PROPERTY} is a C# field or property of the parent component.
- Result of a method: Title="@{METHOD}", where the placeholder {METHOD} is a C# method of the parent component.
- Implicit or explicit expression: Title="@({EXPRESSION})", where the placeholder {EXPRESSION} is a C# expression.
```

I can't really interpret this.

> By convention, an attribute value that consists of C# code is assigned to a parameter using Razor's reserved @ symbol:

The passive voice makes this unintelligible to me. Who assigns what to what, and when do they do it? Am I supposed to do something with an `@` symbol, or is this something the system is doing for me?

> Parent field or property: Title="@{FIELD OR PROPERTY}, where the placeholder {FIELD OR PROPERTY} is a C# field or property of the parent component.

Parent of what? Who is assigning what to what here?

> - Result of a method: Title="@{METHOD}", where the placeholder {METHOD} is a C# method of the parent component.

Method invocation has to end with `()`, but this seems to imply not.

> - Implicit or explicit expression: Title="@({EXPRESSION})", where the placeholder {EXPRESSION} is a C# expression.

Is this different from the cases above? Are the cases above a special case of this one? Why are we talking about three different cases - aren't they all the same?

Hope this doesn't seem too negative! I'm just a bit weirded out when I find things in our docs that I genuinely can't understand.

---
#### Document Details

* Content: [Create and use ASP.NET Core Razor components](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-5.0#parameters)
* Content Source: [aspnetcore/blazor/components/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/components/index.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

### LINKED ISSUE #21553: Suggestion for content about parameters being auto-properties
In the section about Component Parameters, could we add a short section similar to the following? The reason is that we're about to add a new compile-time warning if anyone creates a parameter which *isn't* an auto-property, so it could be good to have docs explaining why they should be auto-properties.

#### Component parameters should be auto-properties

Component parameters should be declared as auto-properties, meaning that they do not contain any custom logic in their getters or setters. For example:

```cs
[Parameter]
public DateTime StartData { get; set; }
```

Developers should not put any custom logic in the `get` or `set` accessor. This is because component parameters are intended purely as a channel for a parent component to flow information to a child component. If, for example, the setter contained logic that caused the parent to be re-rendered, this would lead to an infinite rendering loop.

If you need to transform the received parameter value in any way, then you should leave the parameter property as a pure auto-property to represent the initial raw data, and create some other property or method that supplies the transformed data based on this. You can override `OnParametersSetAsync` if you want to do this each time new data is received.

---
#### Document Details

* Content: [Create and use ASP.NET Core Razor components](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-5.0#parameters)
* Content Source: [aspnetcore/blazor/components/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/components/index.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #21552
Fixes #21553

[Internal Review Topic (links to section)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/components/index?view=aspnetcore-3.1&branch=pr-en-us-21555#component-parameters)

This is a little rough. If you don't get a chance to look today (Monday), I plan to give it another polish pass on Tuesday AM. It will probably improve tomorrow morning.

### General remark on UE progress

I'm addressing sour guidance on the UE passes on https://github.com/dotnet/AspNetCore.Docs/issues/19286. I haven't reached this topic yet. Given the amount of hack'in on these topics over a long period of time, all bets are off on quality until I reach and overhaul them. This one tho ... for sure ... let's immediately address ...

* The main bug of missing the `()` for a method result assignment.
* Other problems in assignment (e.g., Razor expressions).
* Flesh out examples.
* Clarify concepts (e.g., who assigns what to whom).

### Do you want to take the approach drafted here on the PR of a bullet list explaining simple assignment scenarios?

Of course, I'm all :ear: :ear: :ear: (yes, I have three :ear: ... 👽-rex) for advice on missing assignment scenarios and **_gotchas_** 😈 to call out. I'll get them in probably Tuesday morning if I hear back. If you want to cover this a different way, let me know.

### DIFF
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -343,7 +343,7 @@ Assign C# fields, properties, and methods to component parameters as HTML attrib
   }
   ```
   
-* To assign the result of an [explicit C# expression](xref:mvc/views/razor#explicit-razor-expressions) in the parent component to a child component's parameter, surround the expression in parentheses with an `@` symbol prefix. To calculate a date that's one week in the past for assignment to a child's parameter, use the syntax `Title="@((DateTime.Now - TimeSpan.FromDays(7)).ToString())"`, as seen in the following child and parent component:
+* To assign the result of an [explicit C# expression](xref:mvc/views/razor#explicit-razor-expressions) in the parent component to a child component's parameter, surround the expression in parentheses with an `@` symbol prefix. To calculate a date that's one week in the past for assignment to a child's parameter, use the syntax `{PARAMETER}="@((DateTime.Now - TimeSpan.FromDays(7)).ToString())"`, where the `{PARAMETER}` placeholder is the child component's parameter name. In the following example, the child component has a `DateTime` component parameter. The parent component assigns a date value calculated from the current local date with an explicit C# expression.
 
   `Shared/ChildComponent.razor`:
   
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -343,8 +343,10 @@ Assign C# fields, properties, and methods to component parameters as HTML attrib
   }
   ```
   
-* To assign the result of an [explicit C# expression](xref:mvc/views/razor#explicit-razor-expressions) in the parent component to a child component's parameter, surround the expression in parentheses with an `@` symbol prefix. To calculate a date that's one week in the past for assignment to a child's parameter, use the syntax `{PARAMETER (DateTime type)}="@(DateTime.Now - TimeSpan.FromDays(7))"`, where the `{PARAMETER (DateTime type)}` placeholder is the child component's parameter name. In the following example, the child component has a <xref:System.DateTime> component parameter. The parent component assigns a date value calculated from the current local date with an explicit C# expression.
+* To assign the result of an [explicit C# expression](xref:mvc/views/razor#explicit-razor-expressions) in the parent component to a child component's parameter, surround the expression in parentheses with an `@` symbol prefix.
 
+  The following child component has a <xref:System.DateTime> component parameter, `ShowItemsSinceDate`.
+  
   `Shared/ChildComponent.razor`:
   
   ```razor
@@ -364,6 +366,8 @@ Assign C# fields, properties, and methods to component parameters as HTML attrib
       public RenderFragment ChildContent { get; set; }
   }
   ```
+  
+  The following parent component calculates a date with an explicit C# expression that's one week in the past for assignment to the child's `ShowItemsSinceDate` parameter.
   
   `Pages/ParentComponent.razor`:
 
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -270,36 +270,24 @@ In the following example from the sample app, the `ParentComponent` sets the val
 
 Assign C# fields, properties, and methods to component parameters as HTML attribute values using [Razor's reserved `@` symbol](xref:mvc/views/razor#razor-syntax):
 
-* To assign a parent component's field or property value to a child component's parameter, prefix the field or property name with the `@` symbol. To assign a C# field named `title` to the `Title` parameter, use the syntax `Title="@title"`, as seen in the following parent component:
+* To assign a parent component's field, property, or method to a child component's parameter, prefix the field, property, or method name with the `@` symbol. To assign the result of an [implicit C# expression](xref:mvc/views/razor#implicit-razor-expressions) in the parent component to a child component's parameter, prefix the implicit expression with an `@` symbol.
+
+  The following parent component example uses the preceding `ChildComponent` component example and sets its `Title` parameter value to:
+
+  * The value of the `title` field in the first child component.
+  * The result of the `GetTitle` C# method in the second child component.
+  * The current local date in long format with <xref:System.DateTime.ToLongDateString%2A> in the third child component.
+  * The `person` object's `Name` property in the fourth child component.
 
   ```razor
   <ChildComponent Title="@title">
       Title from field.
   </ChildComponent>
   
-  @code {
-      private string title = "Panel Title from Parent";
-  }
-  ```
-
-* To assign the result of a parent component's method to a child component's parameter, prefix the method name with the `@` symbol. To assign the result of a C# method named `GetTitle` to the `Title` parameter, use the syntax `Title="@GetTitle"`, as seen in the following parent component:
-
-  ```razor
   <ChildComponent Title="@GetTitle()">
       Title from method.
   </ChildComponent>
   
-  @code {
-      private string GetTitle()
-      {
-          return "Panel Title from Parent";
-      }
-  }
-  ```
-  
-* To assign the result of an [implicit C# expression](xref:mvc/views/razor#implicit-razor-expressions) in the parent component to a child component's parameter, prefix the implicit expression with an `@` symbol. To assign the current local date in long format with <xref:System.DateTime.ToLongDateString%2A> in the parent component, assign `@DateTime.Now.ToLongDateString()` to the child's parameter. To assign a `person` object's `Name` property, assign `@person.Name`. The following example demonstrates these concepts:
-
-  ```razor
   <ChildComponent Title="@DateTime.Now.ToLongDateString()">
       Title from implicit Razor expression.
   </ChildComponent>
@@ -307,6 +295,21 @@ Assign C# fields, properties, and methods to component parameters as HTML attrib
   <ChildComponent Title="@person.Name">
       Title from implicit Razor expression.
   </ChildComponent>
+  
+  @code {
+      private string title = "Panel Title from Parent";
+      private Person person = new Person();
+      
+      private string GetTitle()
+      {
+          return "Panel Title from Parent";
+      }
+      
+      private class Person
+      {
+          public string Name { get; set; } = "Dr. Who";
+      }
+  }
   ```
   
   Unlike in Razor pages (`.cshtml`), Blazor can't perform asynchronous work in a Razor expression while rendering a component. This is because Blazor is designed for rendering interactive UIs. In an interactive UI, the screen must always display something, so it doesn't make sense to block the rendering flow. Instead, asynchronous work is performed during one of the [asynchronous lifecycle events](xref:blazor/components/lifecycle). After the component is rendered, the UI may be updated. The following Razor syntax is **not** supported:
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/components/index.md
```diff
@@ -268,17 +268,141 @@ In the following example from the sample app, the `ParentComponent` sets the val
...
+
+  ```razor
+  <ChildComponent Title="@DateTime.Now.ToLongDateString()">
+      Title from implicit Razor expression.
+  </ChildComponent>
+  
+  <ChildComponent Title="@person.Name">
+      Title from implicit Razor expression.
+  </ChildComponent>
+  ```
```
SteveSandersonMS, line 310: The above is all absolutely fine and good now, thanks! I do still wonder why we're splitting it into these three cases, since there's nothing in the product code that distinguishes them. As far as the Razor language is concerned, all three of these are simply "using a C# expression", and we don't care whether the expression is a field/property/method-call/complex-expression.

You could even combine all three into a single example, similar to:

----

To assign the result of a C# expression to a child component parameter, use the `@` symbol. As you see in the following example, the expression can be a field, property, method call, or other complex evaluation.

```razor
<ChildComponent Visible="@IsVisible" Title="@ComputeTitle()"  ExpiryDate="@DateTime.Now.AddDays(7)" />
```

aspnetcore/blazor/components/index.md
```diff
@@ -268,17 +268,141 @@ In the following example from the sample app, the `ParentComponent` sets the val
...
+  
+  Unlike in Razor pages (`.cshtml`), Blazor can't perform asynchronous work in a Razor expression while rendering a component. Instead, asynchronous work is performed during one of the [asynchronous component lifecycle events](xref:blazor/components/lifecycle). The following Razor syntax is **not** supported:
```
SteveSandersonMS, line 312: Just to clarify that it's for a good reason, not just a technological limitation:

```suggestion
  Unlike in Razor pages (`.cshtml`), Blazor can't perform asynchronous work in a Razor expression while rendering a component. This is because Blazor is designed for rendering interactive UIs, and in an interactive UI, there must always be something on the screen, so it doesn't make sense to block the rendering flow. Instead, asynchronous work is performed during one of the [asynchronous component lifecycle events](xref:blazor/components/lifecycle), after which the UI may be updated. The following Razor syntax is **not** supported:
```

aspnetcore/blazor/components/index.md
```diff
@@ -268,17 +268,141 @@ In the following example from the sample app, the `ParentComponent` sets the val
...
+  <ChildComponent Title="@title">
+      Title from implicit Razor expression.
+  </ChildComponent>
+  
+  @code {
+      private string title;
+      
+      protected override async Task OnInitializedAsync()
+      {
+          title = await Task.FromResult<string>("Panel Title from Parent");
```
SteveSandersonMS, line 336: The `Task.FromResult` snippet may be confusing because people might think they are supposed to use that in their code for some reason. It might be easier to follow like this:

```suggestion
          title = await ...;
```

... like you have in the Razor snippet above.

aspnetcore/blazor/components/index.md
```diff
@@ -268,17 +268,141 @@ In the following example from the sample app, the `ParentComponent` sets the val
...
+      }
+  }
+  ```
+  
+* To assign the result of an [explicit C# expression](xref:mvc/views/razor#explicit-razor-expressions) in the parent component to a child component's parameter, surround the expression in parentheses with an `@` symbol prefix. To calculate a date that's one week in the past for assignment to a child's parameter, use the syntax `Title="@((DateTime.Now - TimeSpan.FromDays(7)).ToString())"`, as seen in the following parent component:
+
+  ```razor
+  <ChildComponent Title="@((DateTime.Now - TimeSpan.FromDays(7)).ToString())">
```
SteveSandersonMS, line 344: Putting `.ToString()` on this is a bit weird. Suggestion:

```suggestion
  <ChildComponent ShowItemsSinceDate="@(DateTime.Now - TimeSpan.FromDays(7))">
```

aspnetcore/blazor/components/index.md
guardrex: I'm only concerned on the **_gotcha_** 😈 scenarios that work in a Razor page but not a Razor component parameter assignment. IMO, we *might* want to leave those two split out and combine the rest.

I'll work it again Tuesday morning and ping u back.

aspnetcore/blazor/components/index.md
```diff
@@ -268,17 +268,141 @@ In the following example from the sample app, the `ParentComponent` sets the val
...
+      Composed title from method.
+  </ChildComponent>
+  
+  @code {
+      private Product product = new Product();
+
+      private string GetTitle()
+      {
+          return $"SKU-{product.SKU}";
+      }
```
SteveSandersonMS, line 374: Not a big deal, but might be nice:

```suggestion
      private string GetTitle() => $"SKU-{product.SKU}";
```

aspnetcore/blazor/components/index.md
```diff
@@ -268,17 +268,141 @@ In the following example from the sample app, the `ParentComponent` sets the val
...
+#### Component parameters should be auto-properties
+
+Component parameters should be declared as *auto-properties*, meaning that they shouldn't contain custom logic in their getters or setters. For example, the following `StartData` property is an auto-property:
+
+```csharp
+[Parameter]
+public DateTime StartData { get; set; }
+```
+
+Don't place custom logic in the `get` or `set` accessor because component parameters are purely intended for use as a channel for a parent component to flow information to a child component. If a setter of a child component property contains logic that causes rerendering of the parent component, an *infinite rendering loop results*.
```
SteveSandersonMS, line 397: I don't know why *infinite rendering loop results* is in italics here, but will leave that to you :)

aspnetcore/blazor/components/index.md
```diff
@@ -268,17 +268,141 @@ In the following example from the sample app, the `ParentComponent` sets the val
...
+
+If you need to transform a received parameter value:
+
+* Leave the parameter property as a pure auto-property to represent the initial raw data.
```
SteveSandersonMS, line 401:
```suggestion
* Leave the parameter property as a pure auto-property to represent the supplied raw data.
```

SteveSandersonMS:
Thanks for responding so quickly to this, @guardrex!

aspnetcore/blazor/components/index.md
guardrex: Just to emphasize that it will :boom:. I'll remove it.

```suggestion
Don't place custom logic in the `get` or `set` accessor because component parameters are purely intended for use as a channel for a parent component to flow information to a child component. If a setter of a child component property contains logic that causes rerendering of the parent component, an infinite rendering loop results.
```

guardrex:
Sure thing ... I'll wrap it up first thing Tuesday morning and ping u back for final touches.

aspnetcore/blazor/components/index.md
guardrex: btw - This happened because my parameter in the child for this example is marked `string` because I was trying to pass a **_title string_**. I probably should've made it clearer in the text what I was trying to do. It's not really a mistake given what I was doing; but given that you're changing the example here to pass a DT, it's not going to be a factor.

aspnetcore/blazor/components/index.md
guardrex: @SteveSandersonMS ... I'm not so sure about making this change. Here's the deal ... readers are cutting and pasting our code from these topics directly into their own test app to play with the scenarios. Your example changes to combine (and including your prompt to not `ToString` the DT) break that educational experience because they can't use the `ChildComponent` component any longer to perform the testing/hacking/playing with our code. I don't think we should necessarily expand (and complicate) that child component. I'll work with it a bit more and see what I can come up with. I'll at least try combining the simple use cases into a single bullet and example.

guardrex:
@SteveSandersonMS ... Try it now ...

* Combine simple examples into one parent ... **_BUT_** ... let's continue to use the `ChildComponent` example to allow readers to cut-'n-paste and play with the parent component example.
* When we switch over to `ShowItemsSinceDate` for the explicit Razor expression example and drop the use of `Title` and the `ToString`, let's give the reader a new `ChildComponent` with a DT param. Again, **_cut-'n-paste_** is a winner 🥇 with readers.

**WRT component naming**: Using component naming with the word "Component" (e.g., "`ChildComponent` component") is going away. These examples read better as just "`Child` component" in the text with a filename of `Shared/Child.razor`. I'll get these and other similarly-named components squared away on the UE pass coming up shortly for this topic. For now, let's go with what's here.

**WRT UE pass progress**: This topic and many others haven't been reached *yet* on [Blazor project tracking (dotnet/AspNetCore.Docs #19286)](https://github.com/dotnet/AspNetCore.Docs/issues/19286). I understand your frustration ... I'm not happy either with all of the content. However, topics have been hacked to hell over time. Some passages and examples were shot into the docs in mere minutes because that's all the time that I had. I think that's what happened for the bits that we're addressing on this PR. As we all know, I'm no **_Bill Shakespeare of tech docs_** 😆 ... I can't produce the highest quality docs when working so fast 🏃. It's going to take me a little while to address all of the problems. Let's see how things look **_after_** all of the UE passes are complete.

aspnetcore/blazor/components/index.md
```diff
@@ -268,17 +268,169 @@ In the following example from the sample app, the `ParentComponent` sets the val
...
+      }
+      
+      private class Person
+      {
+          public string Name { get; set; } = "Dr. Who";
+      }
+  }
+  ```
+  
+  Unlike in Razor pages (`.cshtml`), Blazor can't perform asynchronous work in a Razor expression while rendering a component. This is because Blazor is designed for rendering interactive UIs. In an interactive UI, the screen must always display something, so it doesn't make sense to block the rendering flow. Instead, asynchronous work is performed during one of the [asynchronous lifecycle events](xref:blazor/components/lifecycle). After the component is rendered, the UI may be updated. The following Razor syntax is **not** supported:
```
SteveSandersonMS, line 317:
```suggestion
  Unlike in Razor pages (`.cshtml`), Blazor can't perform asynchronous work in a Razor expression while rendering a component. This is because Blazor is designed for rendering interactive UIs. In an interactive UI, the screen must always display something, so it doesn't make sense to block the rendering flow. Instead, asynchronous work is performed during one of the [asynchronous lifecycle events](xref:blazor/components/lifecycle). After each asynchronous lifecycle event, the UI may be updated again. The following Razor syntax is **not** supported:
```

aspnetcore/blazor/components/index.md
```diff
@@ -268,17 +268,169 @@ In the following example from the sample app, the `ParentComponent` sets the val
...
+  
+  ```razor
+  <ChildComponent Title="@await ...">
+      ...
+  </ChildComponent>
+  ```
+  
+  The code in the preceding example generates an IntelliSense error in Visual Studio and a *compiler error* if the app is built:
```
SteveSandersonMS, line 325: We support non-VS scenarios too (e.g., VS Code or command line) so best not to refer to VS or IntelliSense here.

```suggestion
  The code in the preceding example generates a *compiler error* if the app is built:
```

aspnetcore/blazor/components/index.md
```diff
@@ -268,17 +268,169 @@ In the following example from the sample app, the `ParentComponent` sets the val
...
+
+  Use of an explicit expression to concatenate text with an expression result for assignment to a parameter is **not** supported. The following example seeks to concatenate the text "SKU-" with a product stock number (`SKU` property, "Stock Keeping Unit") provided by a parent component's `product` object. Although this syntax is supported in a Razor page (`.cshtml`), it isn't valid for assignment to the child's `Title` parameter.
+  
+  ```razor
+  <ChildComponent Title="SKU-@(product.SKU)">
+      Title from composed Razor expression. This doesn't compile.
+  </ChildComponent>
+  ```
+  
+  The code in the preceding example generates an IntelliSense error in Visual Studio and a *compiler error* if the app is built:
```
SteveSandersonMS, line 388:
```suggestion
  The code in the preceding example generates a *compiler error* if the app is built:
```

Again, it's not important that VS gives you a preview of the compiler errors that you're going to get if you build. That's just a VS feature.

SteveSandersonMS (approved):
Thanks @guardrex, this is great!

PR #21562, opened 2021-02-16
Fix IFilterFactory heading level in filters.md

dionrhys:
`IFilterFactory` is unrelated to **Result filters** so this PR moves it up in the heading structure to be its own heading in the TOC.

It might be the case that this section should actually be under **Dependency injection** instead though, and that it's located in the wrong part of the page currently.


Rick-Anderson:
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/mvc/controllers/filters?view=aspnetcore-6.0&branch=pr-en-us-21562)

Rick-Anderson:
Thanks @dionrhys, but I'll bet you can find bigger issues to improve. Much appreciated.

Issue #21615, opened 2021-02-24
RemoteIpAddress


uxoricide:
Hi, 

The `HttpContext.Connection.RemoteIpAddress` is the `client IP` and not - quoting the doc. "The last proxy's IP address, and optionally a port number..."

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e548cd16-b06c-5802-a1a7-c627167857ae
* Version Independent ID: c48abb08-bea3-dcff-c067-2edf5dc9b3a6
* Content: [Configure ASP.NET Core to work with proxy servers and load balancers](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-5.0)
* Content Source: [aspnetcore/host-and-deploy/proxy-load-balancer.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/host-and-deploy/proxy-load-balancer.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-hostdeploy**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
@Tratcher  please review

Tratcher:
Yeah, it's a complicated flow and an execution example would help. There's the start of one here that shows the setup:
https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/host-and-deploy/proxy-load-balancer.md#forwarded-headers-middleware-options
(I'm ignoring the X-Forwarded-For-My-Custom-Header-Name setting...)

Next you'd need to specify the initial values for RemoteIpAddress, Scheme, and Host (`127.0.10.1`, `http`, and `internalhost`). These represent the connection to the most immediate proxy.

Now you need some headers:
X-Forwarded-For: 1.1.1.1, 2.2.2.2, 3.3.3.3
X-Forwarded-Proto: https, http
X-Forwarded-Host: publicdomain.org, middletier.mydomain

Now for the flow:
- The middleware processes entries in sets, right to left.
- The first set is the initial conditions (`127.0.10.1`, `http`, and `internalhost`). `127.0.10.1` is on the `KnownProxies` list so we trust it and continue.
[...]

PR #21633, opened 2021-02-26
Update proxy-load-balancer.md

Rick-Anderson:
Stop gap for https://github.com/dotnet/AspNetCore.Docs/issues/21615 until it can be addressed in the doc.


Tratcher (approved):
The discussion isn't specific to this field, it's the whole section. Can you move that link after the list?

Rick-Anderson:
Hopefully this is the right place.

Tratcher (approved):

PR #21645, opened 2021-02-28
Update protobuf.md

rantri:
Property name should be “Data”, not “Metadata”.


guardrex (approved):
Thanks @rantri ... We'll get this merged shortly.

PR #21648, opened 2021-03-01
Update call-javascript-from-dotnet.md

AnkitSharma-007:
I think the custom JS files should be referenced after the blazor framework JS files as shown in the following code sample. 
- https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/common/samples/5.x/BlazorWebAssemblySample/wwwroot/index.html
- https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/common/samples/5.x/BlazorServerSample/Pages/_Host.cshtml

Please correct me if I am wrong or have missed something.

Thanks

### DIFF
--- a/aspnetcore/blazor/call-javascript-from-dotnet.md
+++ b/aspnetcore/blazor/call-javascript-from-dotnet.md
@@ -18,7 +18,7 @@ This article covers invoking JavaScript functions from .NET. For information on 
 [View or download sample code](https://github.com/dotnet/AspNetCore.Docs/tree/master/aspnetcore/blazor/common/samples/) ([how to download](xref:index#how-to-download-a-sample))
 
 > [!NOTE]
-> Add JS files (`<script>` tags) before the closing `</body>` tag in the `wwwroot/index.html` file (Blazor WebAssembly) or `Pages/_Host.cshtml` file (Blazor Server). Ensure that JS files with JS interop methods are included before Blazor framework JS files.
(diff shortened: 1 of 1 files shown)


guardrex:
Hello @AnkitSharma-007 ... If you use the GH blame feature, you can trace back to the source of the content, where we often discuss complex bits and how the framework is described. This content traces to ...

https://github.com/dotnet/AspNetCore.Docs/pull/21249

Also note that it's best for us not to open a PR for anything other than a minor grammatical change or a simple code example/snippet change. We don't get a ping when a PR is opened out of the blue for a topic, but we get an automatic ping (via the metadata) when an issue is opened, where we can discuss the possible change first.

I'll close this as addressed; but if you feel after reading the PR discussion that there's an error here or a point of confusion, open a new issue from the bottom of the topic with the **This page** feedback form for further discussion and possibly more work.

guardrex:
btw -- Yes, I see what you mean about the files in the sample app. I opened #21652 and will address that right now. Thanks for the heads-up on it.

AnkitSharma-007:
@guardrex 

The ordering of JS file in the code sample and the text in the doc were different and hence confusing for me. I **_assumed_** that the code sample is correct and the doc text is a typo. Therefore, I opened the PR.

However, after referring to the #21249 and #21654, the doc and code sample are in sync. So my doubts are cleared.

Thank you for your prompt response.

guardrex:
Thanks for catching it. I was going toooooo fast 🏃🏃🏃 (*again* 🙈) when we put the NOTE into the topic.

This pair of JS interop topics are coming up on my [UE ("user experience," i.e., topic overhaul) issue](https://github.com/dotnet/AspNetCore.Docs/issues/19286) for many improvements ... and hopefully get any lingering 😈 *little devils* 😈 out of them. Let me know if you spot anything else amiss, and I'll get right on it. :ear:

PR #21656, opened 2021-03-01
Fix invalid URLs in the MVC tutorial

scottaddie:
Fixes some malformed URLs in the "Add controller" step of the MVC tutorial


wadepickett (approved):
Looks good, thanks Scott.

PR #21661, opened 2021-03-03
Update first-web-api.md

### LINKED ISSUE #21647: macOS - no Swagger on  Tutorial: Create a web API
Since this is a "Hello World" kind of situation, more help when things go wrong would be appreciated. I followed the instructions and the /Swagger URL wouldn't load. I'm sure I'm missing some prerequisite, but don't know what yet.

---
#### Document Details

* ID: 583ae3e7-5915-f22b-e54c-6c100cf2a7da
* Version Independent ID: 95f2549e-29f9-e31a-b4d0-432b971643ff
* Content: [Tutorial: Create a web API with ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/tutorials/first-web-api?view=aspnetcore-5.0&tabs=visual-studio)
* Content Source: [aspnetcore/tutorials/first-web-api.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/tutorials/first-web-api.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

---
[Associated WorkItem - 97033](https://dev.azure.com/msft-skilling/Content/_workitems/edit/97033)

Rick-Anderson:
Fixes #21647

### DIFF
--- a/aspnetcore/tutorials/first-web-api.md
+++ b/aspnetcore/tutorials/first-web-api.md
@@ -151,6 +151,8 @@ The Swagger page `/swagger/index.html` is displayed. Select **GET** > **Try it o
 * The response code, body, and headers.
 * A drop down list box with media types and the example value and schema.
 
+If the Swagger page doesn't appear, see [this GitHub issue](https://github.com/dotnet/AspNetCore.Docs/issues/21647).
+
 <!-- Review: Do we care the IE generates several errors. It shows the data, but with  Unrecognized response type; displaying content as text.
 -->
 Swagger is used to generate useful documentation and help pages for web APIs. This tutorial focuses on creating a web API. For more information on Swagger, see <xref:tutorials/web-api-help-pages-using-swagger>.


aspnetcore/tutorials/first-web-api.md
```diff
@@ -151,6 +151,8 @@ The Swagger page `/swagger/index.html` is displayed. Select **GET** > **Try it o
 * The response code, body, and headers.
 * A drop down list box with media types and the example value and schema.
 
+If the Swagger page doesn't appear, see [this GitHub issue](https://github.com/dotnet/AspNetCore.Docs/issues/21647).
```
scottaddie, line 154: Did you mean to link to a different issue?

Rick-Anderson: No, that's the right one.

scottaddie: I think it would be better to include the relevant info from that issue in this doc. Then we keep the reader in docs, where the content is localized.

Rick-Anderson: It's a pretty rare issue. We have 100K PV/month on this and only a couple folks have had that problem. I'm only guessing on what the problem is, there might be other reasons, and the issue is the right place to update those reasons. If we get more than 2 hits every quarter, we can think about moving it into the doc. Right now I don't want to pollute the doc with very rare error information.

Rick-Anderson: Our localization is no better than google translate.

Rick-Anderson: > I fixed it by going to Project Properties -> Debug and setting Hosting Model to "Out of Process".

We can't include that in the doc. Perhaps there's some reason that would fix it, but it doesn't make sense to me. Let's see how much more info we get. Folks that hit this problem can  easily provide feedback in the right place.

PR #21670, opened 2021-03-03
Contact.cs: include using in region

dharmatech:
The following section:

https://docs.microsoft.com/en-us/aspnet/core/security/authorization/secure-data?view=aspnetcore-5.0#create-the-starter-app

shows the following:

image

This will not build without the line:

    using System.ComponentModel.DataAnnotations;

which the user may not know to add. VS and vscode will not add the line for them.

This patch updates the snippet so as to include the using line.

### DIFF
--- a/aspnetcore/security/authorization/secure-data/samples/starter2.1/Models/Contact.cs
+++ b/aspnetcore/security/authorization/secure-data/samples/starter2.1/Models/Contact.cs
@@ -11,6 +11,7 @@ namespace ContactManager.Models
         public string City { get; set; }
         public string State { get; set; }
         public string Zip { get; set; }
+        // using System.ComponentModel.DataAnnotations;
         [DataType(DataType.EmailAddress)]
         public string Email { get; set; }
     }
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@dharmatech  at this stage of the tutorial, we expect developers to know that's needed.
VS add the line when you right click the red squiggly line.

You could add

// requires using ....

and I'll accept that PR

aspnetcore/security/authorization/secure-data/samples/starter2.1/Models/Contact.cs
```diff
@@ -14,5 +14,5 @@ public class Contact
         [DataType(DataType.EmailAddress)]
```
Rick-Anderson, line 14:
```suggestion
        // using System.ComponentModel.DataAnnotations;
        [DataType(DataType.EmailAddress)]
```

aspnetcore/security/authorization/secure-data/samples/starter2.1/Models/Contact.cs
```diff
@@ -1,8 +1,8 @@
-﻿using System.ComponentModel.DataAnnotations;
+#region snippet1
```
Rick-Anderson, line 1: move back

PR #21678, opened 2021-03-04
Blazor Data Binding topic UE pass

guardrex:
Addresses #19286

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/components/data-binding?view=aspnetcore-3.1&branch=pr-en-us-21678)

* There's a new *Custom binding formats* section that flows directly from the error message ...

  > Error: System.ArgumentException: 'bind' does not accept values of type System.Decimal. To read and write this value type, wrap it in a property of type string with suitable getters and setters.

  The section demos the use of `get`/`set` accessors for type binding cases that the built-in bits can't handle directly.

  Is the `DecimalBinding` component example a good 👼 *RexHack*&trade; or is it a 😈 *RexHack*&trade;? I favored it over a `System.Drawing.Color`/color picker example because I think this sort of type/use case is more common.

* I can't grok Ryan's remark in the [engineering discussion (dotnet/aspnetcore #5523)](https://github.com/dotnet/aspnetcore/issues/5523#issuecomment-514848033) that ...

  > We have bespoke support for decimal.

  What did he mean about having [custom] support for decimals? AFAICT, one can't bind a `decimal` type.

* We're still not explaining/showing the use of `CreateBinder`. I have an example if we want to. See the following and let me know. :ear:

  https://github.com/dotnet/aspnetcore/issues/5523#issuecomment-791444003


guardrex:
@captainsafia ... Should I go ahead and merge? I can roll the 🎲🎲 on my `DecimalBinding` component. I think it's ok 🤞🍀. I can take reader feedback on it.

WRT `CreateBinder`, I'll ask Steve about it offline.

PR #21695, opened 2021-03-05
Add information about the location of app.UseWebsockets()

princefishthrower:
See https://github.com/dotnet/aspnetcore/issues/30646 - you won't be able to accept WebSocket connections in a controller if the UseWebSockets() middleware is below the UseEndpoints() middleware. In my opinion, whether or not the root issue is a bug is irrelevant (and I don't know if it is or not yet). This little note would benefit documentation readers. :)

### DIFF
--- a/aspnetcore/fundamentals/websockets.md
+++ b/aspnetcore/fundamentals/websockets.md
@@ -46,7 +46,7 @@ Add the WebSockets middleware in the `Configure` method of the `Startup` class:
 [!code-csharp[](websockets/samples/2.x/WebSocketsSample/Startup.cs?name=UseWebSockets)]
 
 > [!NOTE]
-> If you would like to accept WebSocket requests in a controller, the call to `app.UseWebSockets()` must occur before `app.UseEndpoints()`.
+> If you would like to accept WebSocket requests in a controller, the call to `app.UseWebSockets` must occur before `app.UseEndpoints`.
 
 ::: moniker range="< aspnetcore-2.2"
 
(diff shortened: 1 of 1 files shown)


Tratcher:
The examples should be updated too so they inlclude UseRouting and UseEndpoints. Right now they all handle WebSockets in middleware which doesn't seem that realistic.

@BrennanConroy do any of the SignalR docs mention that you need to call UseWebSockets, and where? I didn't see any.

BrennanConroy:
> do any of the SignalR docs mention that you need to call UseWebSockets

No, because we automatically add it for you.

aspnetcore/fundamentals/websockets.md
```diff
@@ -45,6 +45,9 @@ Add the WebSockets middleware in the `Configure` method of the `Startup` class:
 
 [!code-csharp[](websockets/samples/2.x/WebSocketsSample/Startup.cs?name=UseWebSockets)]
 
+> [!NOTE]
+> If you would like to accept WebSocket requests in a controller, the call to `app.UseWebSockets()` must occur before `app.UseEndpoints()`.
```
Rick-Anderson, line 49:
```suggestion
> If you would like to accept WebSocket requests in a controller, the call to `app.UseWebSockets()` must occur before `app.UseEndpoints`.
```

Rick-Anderson:
> The examples should be updated too so they inlclude `UseRouting `and `UseEndpoints`. Right now they all handle WebSockets in middleware which doesn't seem that realistic.

@Tratcher  I'll handle that in #21701. Can we accept this PR until I can take care of #21701 ?

Tratcher (approved):

PR #21731, opened 2021-03-09
Use correct endpoint to avoid 401

ajbeaven:
The `chat.js` file uses the URL "/chatHub" when connecting to the hub rather than lowercase "/chathub". I was receiving a 404 until changing it to align with the endpoint set in Startup.js.

### DIFF
--- a/aspnetcore/tutorials/signalr/sample-snapshot/3.x/Startup.cs
+++ b/aspnetcore/tutorials/signalr/sample-snapshot/3.x/Startup.cs
@@ -52,7 +52,7 @@ namespace SignalRChat
             app.UseEndpoints(endpoints =>
             {
                 endpoints.MapRazorPages();
-                endpoints.MapHub<ChatHub>("/chathub");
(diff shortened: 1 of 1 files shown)


guardrex:
@BrennanConroy ... Is this a problem? We have the Blazor-SignalR tutorial set up the same way. That hasn't been an issue AFAIK.

ajbeaven:
I'd seen /chatHub used in Startup.cs in a few other tutorials or code around the place (sorry can't find it now to link to), so opted for changing it there. Makes sense to align it with the SignalR and TypeScript tutorials though I suppose.

guardrex:
They're in a flexible state. The Blazor one uses "chathub" all lowercase. The TypeScript one goes with "hub." This one with the mix you pointed out. If Brennan wants to go with "chatHub" here, then I'll change the other two to match if he wants consistency. I'll also search the whole repo for other spots to update (on a separate PR).

guardrex:
So, this is optional but will at least align the casing internally in the doc.

I won't sweat the consistency among docs.

Issue #15848, opened 2019-11-24
HttpClient Factory docs should talk about CookieContainer


rynowak:
See: https://github.com/aspnet/Extensions/issues/872

The HttpClient Factory docs don't mention how an instance of `CookieContainer` will be shared across `HttpClient` instances. We should recommend some options about how to use cookies with the client factory.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 881cef77-11ef-ac11-06d0-0996a1ccfa1b
* Version Independent ID: 3d356334-e672-7e8b-1fef-7944ebee99eb
* Content: [Make HTTP requests using IHttpClientFactory in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-3.0)
* Content Source: [aspnetcore/fundamentals/http-requests.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/fundamentals/http-requests.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-fundamentals**
* GitHub Login: @stevejgordon
* Microsoft Alias: **scaddie**

Rick-Anderson:
@rynowak can you provide a draft?

rynowak:
I sent a PR 😁

ATLSAPI:
@rynowak @Rick-Anderson Thanks for adding this to the docs. We think we are experiencing this issue. In your update to the docs, as an example of how to disable automatic cookie handling, you specified the below:

```csharp
 services.AddHttpClient("configured-disable-automatic-cookies")
	.ConfigurePrimaryHttpMessageHandler(() =>
	{
		return new SocketsHttpClientHandler()
		{
			UseCookies = false,
		};
         });
```

Did you mean `return new SocketsHttpHandler()`? As SocketsHttpClientHandler does not exist on `System.Net.Http`. 

Also, it seems that `SocketsHttpHandler` is only available in NET Core 2.1 so netstandard libraries cant use it? Should this be reflected in the docs? 

I'm happy to raise a new issue/pr with your guidance if that's preferable.

rynowak:
yes that's right! this was mistake. It would be helpful if you could send a PR

ATLSAPI:
[...]

PR #21734, opened 2021-03-10
Update name used in cookie handling snippet

nitinag:
`UseCookies = false` should be the default from a security/privacy perspective, but due to backwards compatibility probably can't be changed now. Thus, it's important that the example here show how to change this for the default instance since that is the most common use case and it was difficult/took some time to find how to configure this for the default instance (since you can't call 'ConfigurePrimaryHttpMessageHandler' on `services.AddHttpClient()`. By updating this example, it helps show you need to use `Options.DefaultName` to make this change for the unnamed/default instance. Related to #15848.


serpent5:
I’d rather keep the example as it was and instead add a more general note/explanation re how to configure the default `HttpClient`. The `Options.DefaultName` "trick" applies more generally than just for `UseCookies`, of course, so it might be helpful to give it direct attention.

I wouldn’t be against the topic including an earlier section on how to configure the default `HttpClient` and then updating this particular example, but I do think it might be confusing to show it without such an explanation.

guardrex:
Thanks @serpent5.

@nitinag ... @serpent5 will open an issue to address this. Thanks for remarking on it.

PR #21777, opened 2021-03-13
Update dotnet-trace instructions to use dotnet-trace ps command

sywhang:
dotnet-trace has a ps command that can be used to determine the PID of all running dotnet processes on the machine, so all these instructions are outdated/unnecessary.


Rick-Anderson:
@sywhang  thanks. This is a big improvement.

Rick-Anderson:
cc @IEvangelist

PR #21576, opened 2021-02-18
Jermy likness pr ef/ra

Rick-Anderson:
Fixes #20763
Fixes #20757
Fixes #19931
Fixes #21720

Internal Review URLS:
* [Intro](https://review.docs.microsoft.com/en-us/aspnet/core/data/ef-rp/intro?view=aspnetcore-5.0&branch=pr-en-us-21576&tabs=visual-studio)

Test

- [ ] links for Use SQLite for development, SQL Server for production

PR #21784, opened 2021-03-13
Update Create.cshtml.cs

Rick-Anderson:
[Original code](https://github.com/dotnet/AspNetCore.Docs/pull/21576/files?file-filters%5B%5D=.cs#diff-5e1d23a451c552729577424eb9e1761e5b0931665584b544161389337b09b230)

Adding the same course in the original. Need to add courses taught by new instructor to `Course.Instructors`

### DIFF
--- a/aspnetcore/data/ef-rp/intro/samples/cu50/Pages/Instructors/Create.cshtml.cs
+++ b/aspnetcore/data/ef-rp/intro/samples/cu50/Pages/Instructors/Create.cshtml.cs
@@ -20,6 +20,13 @@ namespace ContosoUniversity.Pages.Instructors
         public IActionResult OnGet()
         {
             var instructor = new Instructor();
+            Instructor = new Instructor   // For fast testing, remove after accepted.
+            {
+                FirstMidName = "Jon P",
+                LastName = "Smith",
+                HireDate = DateTime.Now
+            };
+
             instructor.Courses = new List<Course>();
 
             // Provides an empty collection for the foreach loop
@@ -66,9 +73,8 @@ namespace ContosoUniversity.Pages.Instructors
                       .FirstOrDefaultAsync(m => m.CourseID == int.Parse(course));
 
                 Course.Instructors.Add(newInstructor);
-                await _context.SaveChangesAsync();
             }
-
+            await _context.SaveChangesAsync();
         }
     }
 }
--- a/aspnetcore/data/ef-rp/intro/samples/cu50/Pages/Instructors/Create.cshtml.cs
+++ b/aspnetcore/data/ef-rp/intro/samples/cu50/Pages/Instructors/Create.cshtml.cs
@@ -12,10 +12,10 @@ namespace ContosoUniversity.Pages.Instructors
     public class CreateModel : InstructorCoursesPageModel
     {
         private readonly ContosoUniversity.Data.SchoolContext _context;
-
         private readonly ILogger<InstructorCoursesPageModel> _logger;
 
-        public CreateModel(SchoolContext context, ILogger<InstructorCoursesPageModel> logger)
+        public CreateModel(SchoolContext context,
+                          ILogger<InstructorCoursesPageModel> logger)
         {
             _context = context;
             _logger = logger;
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@JonPSmith please review. The code now works, but I'm sure you could write it more elegantly.

JonPSmith:
Hi @Rick-Anderson,

Long time not chat! The code will work but it calls `SaveChangesAsync` twice, which isn't a good idea (what happens if the second part fails - you have a new instructor with no courses).

The old code, which adds each course to the `newInstructor.Courses` was better, but still had a problem because it creates new `Course`, which would duplicate the courses. Here is my version (which I have NOT checked)

```c#
public async Task<IActionResult> OnPostAsync(string[] selectedCourses)
{
	if (selectedCourses != null)
	{
		//You already have a Instructor instance in the property, so use that.
		Instructor.Courses = new List<Course>();

		//You add existing courses to the new instructor
		foreach (var course in selectedCourses)
		{
			var foundCourse = _context.Courses.SingleOrDefault(c => c.CourseID == int.Parse(course));
			if (foundCourse != null

					Instructor.Courses.Add(foundCourse);
			//Could return error if course id is wrong
		}
		_context.Instructors.Add(Instructor); //This creates a new instructor, with a list of existing courses
											  //NOTE: this is the new EF Core 5 direct many-to-many 
		await _context.SaveChangesAsync(); //Single call to SaveChangesAsync

		//Good return
	}
       //Bad return - no courses

}
```

This relies on EF Core's cleverness at working out that the `Instructor` should be added while the `Course`s are already in the database and all it has to do as create the (hidden) linking table.

Come back to me if this isn't clear. Also, please unit test it - it looks right but you can't be sure until you test it.

Rick-Anderson:
@JonPSmith  fantastic. Slight tweaks adding logging. I think I'll work on this some more and use a view model to prevent over-posting.

Rick-Anderson:
@JonPSmith thanks for your help. Please review my latest tweeks.

JonPSmith:
Looks good, calling `SaveChangesAsync` once is important. The code makes more sense as the user inputs the instructor's info.

aspnetcore/data/ef-rp/intro/samples/cu50/Pages/Instructors/Create.cshtml.cs
```diff
@@ -32,27 +39,44 @@ public IActionResult OnGet()
...
+                var foundCourse = await _context.Courses.FirstOrDefaultAsync(
+                                                 c => c.CourseID == int.Parse(course));
+                if (foundCourse != null)
+                {
+                    newInstructor.Courses.Add(foundCourse);
+                }
+                else
                 {
-                    newInstructor.Courses.Add(new Course { CourseID = int.Parse(course) });
+                    _logger.LogWarning("Course {course} not found", foundCourse);
```
Rick-Anderson, line 55:
```suggestion
                    _logger.LogWarning("Course {course} not found", course);
```

aspnetcore/data/ef-rp/intro/samples/cu50/Pages/Instructors/Create.cshtml.cs
```diff
@@ -32,27 +39,44 @@ public IActionResult OnGet()
...
             var newInstructor = new Instructor();
-            if (selectedCourses != null)
+            newInstructor.Courses = new List<Course>();
+
+            // Add selected Courses courses to the new instructor.
+            foreach (var course in selectedCourses)
             {
-                newInstructor.Courses = new List<Course>();
-                foreach (var course in selectedCourses)
+                var foundCourse = await _context.Courses.FirstOrDefaultAsync(
```
JeremyLikness, line 47: I would use `Find` here instead of `First`. This way the course won't be loaded again if it's already being tracked. If the courses collection isn't large, you could even preemptively do a `_context.Courses.Load()` prior to the loop.

JeremyLikness (approved):
A minor suggestion. Otherwise LGTM.

JeremyLikness (commented):
@Rick-Anderson, this looks fine to me.

PR #21786, opened 2021-03-14
Fix broken link

loic-sharma:


guardrex (approved):
Thanks @loic-sharma! :rocket:

PR #21788, opened 2021-03-14
Missing await in ReadFromJsonAsync

ndc:
The example line to call ReadFromJsonAsync: `var content = response.Content.ReadFromJsonAsync<WeatherForecast>();`, shouldn't it use await?


guardrex (approved):
Thanks @ndc! :rocket: ... I went back to see if there's a clue how this little 😈 got in here ...

https://github.com/dotnet/AspNetCore.Docs/pull/17910/files#diff-4db4dee20a7f98403c569cc20bb53ea4d33ce4d214dfdd561b4a5dcf8bf9af38

... and it's just an oversight on my part. It wasn't part of a sample app, so it wasn't flagged in a build. Indeed, Rick remarked earlier on the lack of a build when these lines had a lowercase "c" in them for `Content` that had to be corrected on an earlier PR ... even then, my :eye: didn't spot the missing `await`.

We're in the process now of having most code snippets come from doc engine-built sample apps. Goofs like this 🙈 will get picked up by the build in the future. That work in on-going as I work down the [UE ("User Experience," i.e., doc overhauls) issue](https://github.com/dotnet/AspNetCore.Docs/issues/19286).

Thanks again for catching and reporting this. :+1:

PR #21823, opened 2021-03-18
Configure gRPC client factory on .NET Standard 2.0

JamesNK:
How to use gRPC client factory on .NET Standard 2.0 implementations.


Rick-Anderson (approved):

Rick-Anderson (approved):
close open to trigger a build

PR #21828, opened 2021-03-19
Clarify health check behavior in case of failures

mfkl:
This PR aims to clarify health check behavior and implementation guidance since https://github.com/dotnet/aspnetcore/pull/21175.

- Document preference for `context.Registration.FailureStatus` to `HealthCheckResult.Unhealthy` in case of failure. This allows to not overwrite the defined status.

- Document framework behavior in case of a health check throwing exception.

Not sure about my xref:something markdown syntax being perfectly correct..

Any feedback welcome!


guardrex:
Hello @mfkl ... Thanks for sending this in, but we can't accept this in its present form because these product updates are for >=5.0 and the topic covers >=3.0 and <3.0 (i.e., we'll need a full >=5.0 copy of the 3.x content and then only have the updates applied to that versioned content, leaving the 3.x content as it is). It's usually best first to discuss the changes on an issue first for non-trivial updates, where we can provide the versioning guidance for the PR. Fortunately here, the code is in the doc and not cross-linked from the sample app ... it's more painful when there's sample app code to work with.

To adjust this PR ...

* Copy all of the 3.x content into a new >=5.0 section added to the top of the topic.

  The moniker range for the 3.x content (where it has `::: moniker range=">= aspnetcore-3.0"` now) becomes the middle section of content with a moniker range of ...

  ```
  ```

* Only apply your updates to the new >=5.0 versioned content.

If you don't want to make these updates to this PR, close this PR and open a new issue for the topic from the **This Page** feedback button+form at the bottom and mention this closed PR in your issue comment. One of the doc authors will work the issue.

guardrex:
Moved to https://github.com/dotnet/AspNetCore.Docs/issues/21848.

PR #21830, opened 2021-03-19
Fixed typo and added deep link

poornimanayar:
Hi,

I have corrected a typo in the `Getting Started ` for EF MVC. I also think it can be helpful if there can be a deep link to the performance considerations from the MVC page. I have done both in a single PR as they are on the same page. 

Poornima


scottaddie (approved):

scottaddie:
Thank you for the PR, @poornimanayar!

Issue #21848, opened 2021-03-22
Clarify health check behavior in case of failures


guardrex:
Apply https://github.com/dotnet/AspNetCore.Docs/pull/21828 with versioning.

@Rick-Anderson ... I can squeeze this one into my schedule sometime this week, but re-assign if that's best.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: c0cbb094-29e9-740a-9fd6-db372641d753
* Version Independent ID: 52d8efc9-a336-71d9-6d90-98173cbf6bf4
* Content: [Health checks in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-5.0)
* Content Source: [aspnetcore/host-and-deploy/health-checks.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/host-and-deploy/health-checks.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-hostdeploy**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

PR #21853, opened 2021-03-23
Clarify health check behavior in case of failures

mfkl:
Replaces https://github.com/dotnet/AspNetCore.Docs/pull/21828
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/21848

/cc @guardrex


guardrex:
Thanks for trying this, @mfkl, but it's still not quite correct. There's an embedded section of versioning within the >=5.0 content on this PR, which isn't allowed by our build engine. It's probably faster if I set it up from your content. I'm sorry it's so complicated.

PR #21915, opened 2021-03-29
Create breakingChanges6.md

Rick-Anderson:


guardrex:
Don't you want it to be more general in this context? ... I mean if it's going to be used across the whole repo. I only focused on the ANCM/hosting bundle because that pertains to those two topics.

PR #21942, opened 2021-04-01
Updated the ASP.NET Core version

ankitvijay:
Updated the ASP.NET Core version to reflect that it works for ASP.NET Core 2.1 and above versions.


guardrex (approved):
Thanks @ankitvijay! :rocket: ... We use "or later" by convention, so I touched it up. I'll merge this as soon as the build is complete.

PR #21952, opened 2021-04-01
Blazor state container example fix

### LINKED ISSUE #21948: ASP.NET Core Blazor state management code smell
Hi,
The "In-memory state container service" demo code contains a major code-smell --

```cs
    public string Property { get; set; } = "Initial value from StateContainer";

    public event Action OnChange;

    public void SetProperty(string value)
    {
        Property = value;
        NotifyStateChanged();
    }
```

The property `Property` has a public setter but the demo code shows that the property should be set through the `SetProperty` method. It's bad practice to have both a SetPropertyName method, and public property setter which would only set the autoproperty without doing the other bits (like the notifying of the UI which defeats the point!)

A sub-optimal fix would be to make the set private on the property so SetProperty must be called.
Better, the fix would be to scrap the SetProperty, move its implementation into the setter, and have a backstore for the property.

```cs
        private string savedStringValue;

        public string SavedStringValue
        {
            get => savedStringValue;
            set 
            {
                savedStringValue = value;
                NotifyStateChanged();
            }
        }
 ```

Thanks

---
#### Document Details

* ID: ae14b3e3-3271-b919-9e0a-6f317a6b2d2d
* Version Independent ID: e5e1273b-195e-5da1-b4aa-66bbcff1425b
* Content: [ASP.NET Core Blazor state management](https://docs.microsoft.com/en-us/aspnet/core/blazor/state-management?view=aspnetcore-3.1&pivots=server#in-memory-state-container-service-server)
* Content Source: [aspnetcore/blazor/state-management.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/state-management.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #21948

Thanks @kjhf! :rocket: ... Would u like to review this?

### DIFF
--- a/aspnetcore/blazor/includes/state-container.md
+++ b/aspnetcore/blazor/includes/state-container.md
@@ -95,7 +95,7 @@ services.AddScoped<StateContainer>();
     </button>
 </p>
 
-<NestedComponent />
+<Nested />
 
 @code {
     protected override void OnInitialized()
--- a/aspnetcore/blazor/includes/state-container.md
+++ b/aspnetcore/blazor/includes/state-container.md
@@ -4,22 +4,28 @@ no-loc: [appsettings.json, "ASP.NET Core Identity", cookie, Cookie, Blazor, "Bla
 Nested components typically bind data using *chained bind* as described in <xref:blazor/components/data-binding>. Nested and un-nested components can share access to data using a registered in-memory state container. A custom state container class can use an assignable <xref:System.Action> to notify components in different parts of the app of state changes. In the following example:
 
 * A pair of components uses a state container to track a property.
-* The components of the example are nested, but nesting isn't required for this approach to work.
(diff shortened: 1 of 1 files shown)


guardrex:
I'm going to go ahead and get this in as soon as the build is clear. I had the J&J vaccine 💉 yesterday ... mild soreness everywhere kept me up last night. I need to go lie down 🛏️ for a little while. Ping me for a patch if it needs it.

guardrex:
Arrrr ... looks like the build system melted down. :boom:

Rick-Anderson:
That's the fix the internal folks recommended close/reopen

kjhf:
Looks great thank you for your work on this @guardrex :)

guardrex:
Sure thing. Thanks again for pointing me in the right direction.

PR #21964, opened 2021-04-02
Cross-link and enhance scoping in a console app

### LINKED ISSUE #21958: Why do you need scope in console app usage example?
Why do you need scope in console app usage example?

        using (var serviceScope = host.Services.CreateScope())
        {
            var services = serviceScope.ServiceProvider;

            try
            {
                var myService = services.GetRequiredService<IMyService>();

Why not just  `var myService = host.Services.GetRequiredService<IMyService>();` ?

---
#### Document Details

* ID: 881cef77-11ef-ac11-06d0-0996a1ccfa1b
* Version Independent ID: 3d356334-e672-7e8b-1fef-7944ebee99eb
* Content: [Make HTTP requests using IHttpClientFactory in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-5.0)
* Content Source: [aspnetcore/fundamentals/http-requests.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/http-requests.md)
* Product: **aspnet-core**
* GitHub Login: @stevejgordon

guardrex:
Fixes #21958

Thanks @Gladskih! :rocket:

cc: @stevejgordon

### DIFF
--- a/aspnetcore/fundamentals/dependency-injection.md
+++ b/aspnetcore/fundamentals/dependency-injection.md
@@ -809,6 +809,8 @@ public class Program
 }
 ```
 
+It isn't necessary to create a scope for transient services, including for the `ILogger` in the preceding example (see: <xref:fundamentals/logging#create-logs-in-main>). Transients don't resolve inadvertantly as singletons when resolved from the root. Transients are created when requested. If a transient service happens to be disposable, then it's rooted by the container until disposal.
+
 ## Scope validation
 
 When the app is running in the Development environment, the default service provider performs checks to verify that:
--- a/aspnetcore/fundamentals/dependency-injection.md
+++ b/aspnetcore/fundamentals/dependency-injection.md
@@ -809,7 +809,7 @@ public class Program
 }
 ```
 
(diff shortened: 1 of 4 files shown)


guardrex:
@stevejgordon ... Do you have a sec to look at this PR and the context on https://github.com/dotnet/AspNetCore.Docs/issues/21958?

The DI topic can take the current update on this PR because the new phrase pertaining to "lifetimes" is in a sentence directly speaking to scoped services. That seems fine.

However, the topic section that you wrote (the example) in the [HTTP Requests topic](https://docs.microsoft.com/aspnet/core/fundamentals/http-requests#use-ihttpclientfactory-in-a-console-app) seems to contradict what DF🦆 is saying on the issue because the service lifetimes at play are transient ... it seems like @Gladskih suggestion of merely ...

```csharp
myService = host.Services.GetRequiredService<IMyService>();
``` 

... works ... I mean if I'm interpreting everything correctly here. If there's a separate reason to create that scope, it's probably best to call it out there in the HTTP Requests topic. I can add a line if I know what to add. Conversely, I'll refactor the example to drop the scope. :ear:

stevejgordon:
Hey @guardrex! I must admit that I don't remember adding that particular sample, but I do have a bad memory! 😄  You're correct, though, that a scope should not really be needed there. I'd probably favour the removal of that scope block as it simplifies the concepts a reader needs to understand. My only concern would be someone copy/pasting and later resolving something like EF Core from there without realising the scoped lifetime issue. However, that shouldn't be a reason to make that sample more complex, just for that scenario as it is unrelated. Perhaps a link to the main DI topic could be provided so the reader can get more information if they need it.

guardrex:
Thanks @stevejgordon ... I'll make the changes, and I did cross-link the HTTP Requests topic section to the DI topic section on using services in `Program.Main`.

PR #21970, opened 2021-04-03
Use ??= operator instead of "If"  for null check

babula38:
Currently using **if**  for null checking and creating **BindingInfo** object, but we have a operator in C# i.e. ??= which does the same thing with less amount of code.
Reference article   https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/application-model?view=aspnetcore-5.0#conventions


babula38:
Closing as it looks this feature is only compatable with C# 8.0, need to re verify from my end.

PR #21978, opened 2021-04-04
Removing extra '-' when calling aspnet-generator

iamnicoj:
Removing an extra '-' when explaining how to call aspnet-generator. The correct way of calling it doesn't need the '-' between the dotnet and the aspnet-generator option. You can also see this as a way of unify how the asp-generator is being used in other tutorials like the first-mvc-app adding-model page here: https://github.com/iamnicoj/AspNetCore.Docs/blob/main/aspnetcore/tutorials/first-mvc-app/adding-model.md

Other related open issue caused by this extra hyphen: dotnet/Scaffolding#1404

### DIFF
--- a/aspnetcore/tutorials/razor-pages/model.md
+++ b/aspnetcore/tutorials/razor-pages/model.md
@@ -128,13 +128,13 @@ The *appsettings.json* file is updated with the connection string used to connec
 * **For Windows**: Run the following command:
 
   ```dotnetcli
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
I believe  the  current syntax is preferred, although they both work. I'll verify and update https://docs.microsoft.com/en-us/aspnet/core/fundamentals/tools/dotnet-aspnet-codegenerator

cc @wadepickett

Rick-Anderson:
Actually, your syntax has better error handling.

PR #22013, opened 2021-04-09
Changes to the ASP.NET Core Razor SDK

### LINKED ISSUE #21758: Update Razor SDK documentation to react to .NET 6 changes
For .NET 6, we moved Razor compilation from using our own compilation server to leveraging source generators. This will require that we update a few things in the repo.

- We need to update the Razor SDK page to reference the targets that will no longer be written.
- We need to remove references to a Views assembly from .NET 6 Preview 3 and onwards since we are now falling back to a single app assembly.

Rick-Anderson:
Fixes #21758

[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/migration/50-to-60?view=aspnetcore-1.0&branch=pr-en-us-22013&tabs=visual-studio#changes-to-the-aspnet-core-razor-sdk)

### DIFF
--- a/aspnetcore/migration/50-to-60.md
+++ b/aspnetcore/migration/50-to-60.md
@@ -79,18 +79,20 @@ For apps using Docker, update your *Dockerfile* `FROM` statements and scripts. U
 
 ## Changes to the ASP.NET Core Razor SDK
 
-The Razor compiler in .NET 6 leverages the new source generators feature to generated compiled C# files from the Razor views and pages in a document. In previous versions:
+The Razor compiler now leverages the new [source generators feature](https://devblogs.microsoft.com/dotnet/introducing-c-source-generators/) to generated compiled C# files from the Razor views and pages in a document. In previous versions:
 
 * The compilation relied on an individual `RazorGenerate` and `RazorCompile` target to produce the generated code. These targets are no longer valid. In .NET 6, both code generation and compilation are supported by a single call to the compiler.
 * A separate Razor assembly was generated that contained the compiled view types in an application. This behavior has been deprecated and a single assembly is produced that contains both the app types and the generated views.
 
-- Remove all the targets listed under `Targets`. Users can continue to use `RazorComponentGenerateDependsOn` to do things that need to happen before the build runs.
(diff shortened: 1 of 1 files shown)


captainsafia:
The comment I wrote referred to content changes that needed to be made to https://docs.microsoft.com/en-us/aspnet/core/razor-pages/sdk AKA https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/razor-pages/sdk.md.

There aren't any explicit migration steps we need to document from 5.0 to 6.0 WRT to the Razor SDK at the moment.

aspnetcore/migration/50-to-60.md
```diff
@@ -77,6 +77,23 @@ For apps using Docker, update your *Dockerfile* `FROM` statements and scripts. U
 + docker pull mcr.microsoft.com/dotnet/aspnet:6.0
 ```
 
+## Changes to the ASP.NET Core Razor SDK
+
+The Razor compiler now leverages the new [source generators feature](https://devblogs.microsoft.com/dotnet/introducing-c-source-generators/) to generated compiled C# files from the Razor views and pages in a document. In previous versions:
+
+* The compilation relied on an individual `RazorGenerate` and `RazorCompile` target to produce the generated code. These targets are no longer valid. In .NET 6, both code generation and compilation are supported by a single call to the compiler.
```
Rick-Anderson, line 84:
```suggestion
* The compilation relied on an individual `RazorGenerate` and `RazorCompile` target to produce the generated code. These targets are not valid in .NET 6. In .NET 6, both code generation and compilation are supported by a single call to the compiler.
```

Rick-Anderson:
I can easily move the added H2 (`## Changes to the ASP.NET Core Razor SDK`) to the `sdk.md` doc, but I don't understand why that it doesn't belong in the migration doc. It states:

The compilation relied on an individual `RazorGenerate` and `RazorCompile` target to produce the generated code. ***These targets are not valid in .NET 6. In .NET 6, both code generation and compilation are supported by a single call to the compiler.***
* A separate Razor assembly was generated that contained the compiled view types in an application. This behavior has been deprecated and a single assembly is produced that contains both the app types and the generated views.

I don't understand why that's not a change that needs to be called out.

PR #22015, opened 2021-04-09
Update gRPC WinHttpHandler required version

JamesNK:
[Internal Review:  Use gRPC client with .NET Standard 2.0](https://review.docs.microsoft.com/en-us/aspnet/core/grpc/netstandard?view=aspnetcore-5.0&branch=pr-en-us-22015)


wadepickett (approved):
Looks good James, thanks!

PR #22068, opened 2021-04-16
Update iauthorizationpolicyprovider.md

stephanharmse:
1. Change to custom AuthorizationPolicyProvider example:
- Use the more standard and ubiquitous task based async pattern.

2. Change to using BackupPolicyProvider return statement:
- `BackupPolicyProvider.GetPolicyAsync(policyName)` needs to be awaited. It could cause a race condition.

### DIFF
--- a/aspnetcore/security/authorization/iauthorizationpolicyprovider.md
+++ b/aspnetcore/security/authorization/iauthorizationpolicyprovider.md
@@ -100,17 +100,17 @@ internal class MinimumAgePolicyProvider : IAuthorizationPolicyProvider
     // to be embedded in the policy names. This is abstracted away from developers
     // by the more strongly-typed attributes derived from AuthorizeAttribute
     // (like [MinimumAgeAuthorize()] in this sample)
-    public Task<AuthorizationPolicy> GetPolicyAsync(string policyName)
+    public async Task<AuthorizationPolicy> GetPolicyAsync(string policyName)
     {
         if (policyName.StartsWith(POLICY_PREFIX, StringComparison.OrdinalIgnoreCase) &&
             int.TryParse(policyName.Substring(POLICY_PREFIX.Length), out var age))
         {
             var policy = new AuthorizationPolicyBuilder(CookieAuthenticationDefaults.AuthenticationScheme);
             policy.AddRequirements(new MinimumAgeRequirement(age));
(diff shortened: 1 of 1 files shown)


aspnetcore/security/authorization/iauthorizationpolicyprovider.md
```diff
@@ -100,17 +100,17 @@ internal class MinimumAgePolicyProvider : IAuthorizationPolicyProvider
     // to be embedded in the policy names. This is abstracted away from developers
     // by the more strongly-typed attributes derived from AuthorizeAttribute
     // (like [MinimumAgeAuthorize()] in this sample)
-    public Task<AuthorizationPolicy> GetPolicyAsync(string policyName)
```
mjrousos, line 103: `GetPolicyAsync` doesn't actually do anything async, so it doesn't need the async keyword. `Task.FromResult` is useful for scenarios like this where an async interface is being implemented with a synchronous implementation.

It's not a big difference, but making this method async would trigger diagnostic CS1998 and `Task.FromResult` is the recommended solution (https://stackoverflow.com/a/13254787).

aspnetcore/security/authorization/iauthorizationpolicyprovider.md
```diff
@@ -141,7 +141,7 @@ Then, the `GetPolicyAsync` method can be updated to use the `BackupPolicyProvide
 
 ```csharp
 ...
-return BackupPolicyProvider.GetPolicyAsync(policyName);
```
mjrousos, line 144: Since we're returning a task, there's no need to await it. We can return the task directly and our callers can await _that_ task rather than awaiting our own method (which, in turn, awaits `BackupPolicyProvider.GetPolicyAsync`). Plus, avoiding an extra await will result in the compiler generating more efficient code.

mjrousos (changes requested):

mjrousos:
I think these docs are ok as they are currently. There's a bit of weirdness because the docs demonstrate implementing an async interface with a synchronous implementation, but I believe the way they're doing it currently is best.

guardrex:
Thanks @mjrousos. Closing per your remarks.

stephanharmse:
Thank you for your consideration!

PR #22072, opened 2021-04-16
Added DI sample code required for cache to work

alex-jitbit:
The page only mentions this at the very end, however default ASP.NET Core project created by Visual Studio lacks this line.


Rick-Anderson:
That line is not needed in most cases.. Per the doc:

For most apps, IMemoryCache is enabled. For example, calling AddMvc, AddControllersWithViews, AddRazorPages, AddMvcCore().AddRazorViewEngine, and many other Add{Service} methods in ConfigureServices, enables IMemoryCache. For apps that are not calling one of the preceding Add{Service} methods, it may be necessary to call AddMemoryCache in ConfigureServices.

PR #22082, opened 2021-04-18
Use Array.Empty instead of instantiating a new empty array

meziantou:


guardrex (approved):
Thanks @meziantou! :rocket:

Is this also in the 2.x sample? If so, could you change it there, too?

meziantou:
@guardrex I've fixed a few additional locations

guardrex:
Excellent ... thanks.

PR #22085, opened 2021-04-19
Update to broken link

HannesSam:
The link to the "Build a .NET Core and SQL Database web app in Azure App Service" was broken and now points to the new URL where the article is stored.


guardrex (approved):
Thanks @HannesSam! :rocket:

PR #22091, opened 2021-04-19
Add docs on preventing rendering after event handling

captainsafia:
Addresses https://github.com/dotnet/aspnetcore/issues/18919.

### DIFF
--- a/aspnetcore/blazor/webassembly-performance-best-practices.md
+++ b/aspnetcore/blazor/webassembly-performance-best-practices.md
@@ -422,6 +422,8 @@ In the following example, no event handler added to the component triggers a rer
 ```razor
 @page "/handle-click-1"
 @implements IHandleEvent
+@using Microsoft.Extensions.Logging
+@inject ILogger<HandleClick1> Logger
 
 <p>
     Last render DateTime: @dt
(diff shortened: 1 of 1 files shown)


guardrex:
@captainsafia ... I understand why its here, but the subject matter groups well with the *Event Handling* topic. Is this one of those scenarios that really isn't being pushed out there in the open for the majority of use cases, thus it can hide a bit back here as a perf-only scenario for edge case folks dealing with perf trouble?

captainsafia:
>  I understand why its here, but the subject matter groups well with the Event Handling topic. Is this one of those scenarios that really isn't being pushed out there in the open for the majority of use cases, thus it can hide a bit back here as a perf-only scenario for edge case folks dealing with perf trouble?

Yep -- we don't anticipate this is a common scenario and want to put it here for some select scenarios.

guardrex:
@captainsafia ... See if I accidentally *GooberRexified*&trade; 🙈 your section. A few points ...

* I've been making most examples fully working cut-'n-paste examples.
* You had `onclick="@..."` syntax for the buttons' handlers. I changed them to the usual `@onclick="..."` syntax, and they seem to work that way. I'll revert if that's incorrect. :ear:

captainsafia:
@guardrex Thanks for the edit! These changes look good.

> For ReceiverBase, the comment didn't seem complete ...

Actually, I like your approach better. No need to have comments on code users will be copying into their projects.

> You had onclick="@..." syntax for the buttons' handlers. I changed them to the usual @onclick="..." syntax, and they seem to work that way. I'll revert if that's incorrect. 👂

Should still work.

PR #22114, opened 2021-04-23
Use HttpStatusCode enum instead of hardcoded status code

mrnustik:
Replaces hardcoded status code value with the usage of HttpStatusCode enum casted to int.


guardrex:
Thanks @mrnustik.

@Rick-Anderson ... Currently, this approach is taken in one spot ([here](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/security/authorization/customizingauthorizationmiddlewareresponse/sample/AuthorizationMiddlewareResultHandlerSample/MyAuthorizationMiddlewareResultHandler.cs#L24)). Feels like if it were to be done, it should be done everywhere. There are probably a dozen or so cases across the repo.

Rick-Anderson:
@mrnustik  see #22115 in case you'd like to do a PR fixing the rest of them.

Rick-Anderson (approved):

PR #22116, opened 2021-04-23
Replace status code with status code enum

### LINKED ISSUE #22115: Use HttpStatusCode enum instead of hardcoded status code
For example, see #22114
cc @mrnustik

mrnustik:
Fixes #22115.

Created on @Rick-Anderson suggestion in #22114.

I have also replaced all `StatusCodes` references to `HttpStatusCode` so the code is consistent in the documentation. If this is not right just comment and I will revert the last commit and remove only hardcoded status codes.


Rick-Anderson (approved):

Rick-Anderson:
Much appreciated @mrnustik 
Keep up the good work.

PR #22132, opened 2021-04-25
Fix typo and add link

amal-stack:
Changed "concatination" to "concatenation" and added reference link to "global::"


guardrex (approved):
Thanks @amal-stack! :rocket: ... We do have a spell checker that (as they usually do) places a squiggle under a misspelled word ... this rotten devil got by me 😈.

PR #22146, opened 2021-04-27
Use relative path in NavigateTo

beppemarazzi:
In the code snippet provided as example of doing redirection for user selected culture, there was a NavigateTo with absolute path. This doesn't work if the app is behind a reverse proxy in some virtual subfolder.

### DIFF
--- a/aspnetcore/blazor/globalization-localization.md
+++ b/aspnetcore/blazor/globalization-localization.md
@@ -231,7 +231,7 @@ The following component shows an example of how to perform the initial redirecti
         var query = $"?culture={Uri.EscapeDataString(culture)}&" +
             $"redirectUri={Uri.EscapeDataString(uri)}";
 
-        NavigationManager.NavigateTo("/Culture/SetCulture" + query, forceLoad: true);
+        NavigationManager.NavigateTo("Culture/SetCulture" + query, forceLoad: true);
     }
 }
 ```
--- a/aspnetcore/blazor/globalization-localization.md
+++ b/aspnetcore/blazor/globalization-localization.md
@@ -231,11 +231,12 @@ The following component shows an example of how to perform the initial redirecti
         var query = $"?culture={Uri.EscapeDataString(culture)}&" +
             $"redirectUri={Uri.EscapeDataString(uri)}";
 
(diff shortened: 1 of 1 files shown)


guardrex:
@beppemarazzi ... Are you setting the [app base path](https://docs.microsoft.com/aspnet/core/blazor/host-and-deploy/#app-base-path) in this scenario? If so, then it should form the correct `BaseUri` and ultimately the correct path because the URI provided to `NavigateTo` is relative to to the `BaseUri`.

guardrex:
I think instead of this change that a cross-link here will provide general assistance with in this use case. I'll make an update on this PR to cover it.

beppemarazzi:
> @beppemarazzi ... Are you setting the [app base path](https://docs.microsoft.com/aspnet/core/blazor/host-and-deploy/#app-base-path) in this scenario? If so, then it should form the correct `BaseUri` and ultimately the correct path because the URI provided to `NavigateTo` is relative to to the `BaseUri`.

@guardrex Yes, app base path is correctly set. But with the leading '/' the browser navigates to the absolute path ignoring the base path. 
If i enter `Blazor._internal.navigationManager.getBaseURI()` in a browser console i get something like this `"https://myserver/my/virtual/path/app/"` (with a final '/')
In the same console `Blazor._internal.navigationManager.navigateTo("something/else", true)` navigates to `https://myserver/my/virtual/path/app/something/else` while `Blazor._internal.navigationManager.navigateTo("/something/else", true)` navigates to `https://myserver/something/else`

guardrex:
Thanks for that clarification @beppemarazzi.

🤔 ... The API docs seem a bit weak on details. I expected different behavior and guessed wrong that the `BaseUri` would be the root ... that the presence of the leading slash would be from the root using the virtual path. Without the leading slash, I expected the path to form from the path where the component resides.

It doesn't seem like there is any harm in removing the forward slash. I'll put up a PR in a minute to patch that as you originally suggested. I've also made a note on my UE ("User Experience," i.e., total overhaul+updates) tracking issue to take a closer look at this subject when I reach the *Host and Deploy* overview. I think the docs need a careful assessment on how `NavigateTo` is covered and what examples the docs are showing.

beppemarazzi:
Just found this https://github.com/dotnet/aspnetcore/issues/25204...

PR #22186, opened 2021-05-02
Update intro.md

mverbaas:
Updated the text to have the property allign with the models


Rick-Anderson:
I think the original is better being generic.

PR #22220, opened 2021-05-06
Additional HTTP client reference in note section

EdSF:
Add a note referencing Additional Resources about adding a named HTTP Client in context to the topic for a common usage scenario (mixed authenticated/unauthenticated requests).


guardrex:
Thanks @EdSF, but if there is going to be additional cross-linking, I'd like to address it in a more general way across the topics in this node. I'm also not keen at the moment to cross-link individual sections of *Additional Scenarios*. Our intention isn't to burden the individual WASM security topics with to cross-links to *every* additional scenario in *every* topic. The *Additional Scenarios* topic is already cross-linked. I'll take a look at this situation and all of the coverage in this node when I reach it on the tracking issue. It won't get lost because it's cross-linked to the tracking issue.

EdSF:
Cool beans ~

PR #22279, opened 2021-05-12
Blazor startup topic

### LINKED ISSUE #22197: Document Blazor initialization notwithstanding hosting model
Document `Blazor.start` scenarios without regard to hosting model. Create a *Blazor* > *Fundamentals* > *Startup* topic for the content.

* Base the content on the section introduction at https://docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/signalr?view=aspnetcore-5.0&pivots=server#initialize-the-blazor-circuit.
* The current SignalR lead-in can be reduced considerably and cross-linked to the new *Startup* topic for the generic coverage of `Blazor.start`.
* Two sections move from the *SignalR* topic into the *Startup* topic:
  * Init when doc is loaded ... `loadBootResource`​ option to override the default loading behavior.
  * Chain to the promise.
* There will be `Blazor.start` guidance in the *Call JS from .NET* topic (creating/loading scripts via `createElement`). Cross-link the new content to the new *Startup* topic.
* Doc `environment` config (https://github.com/dotnet/aspnetcore/pull/24031) in the *Environments* topic with a cross-link to the new *Startup* doc. *Startup* should be placed before *Environments* in the ToC.
* Document how the `applicationCulture`​ property can be used to specify the culture that the app runs under to override the default that is set by the browser.

Draft metadata:

```
---
title: ASP.NET Core Blazor startup
author: guardrex
description: TBD
ms.author: riande
ms.custom: mvc
ms.date: XX/YY/2021
no-loc: [Kestrel, appsettings.json, "ASP.NET Core Identity", cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
uid: blazor/fundamentals/startup
---
# ASP.NET Core Blazor startup
```

guardrex:
Fixes #22197

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/startup?view=aspnetcore-3.1&branch=pr-en-us-22279)

👨‍🍳 **_Now, we're cook'in!_** 👨‍🍳

* *Loading boot resources* section is combo of what we had in the *Host and deploy* WASM topic + some lingo from DR's blog post.
* Takes care of setting the environment via the property on start, which was a missed scenario for 5.0. 🙈 *Ooops!*
* It has good cross-links both ways for Blazor start scenario coverage, so the subject matter for JS interop, Brotli compression, SignalR, and loc all remain in their topics.


aspnetcore/blazor/fundamentals/environments.md
```diff
@@ -16,9 +16,47 @@ uid: blazor/fundamentals/environments
...
+The following example starts Blazor in the Staging environment:
+
+```cshtml
+<body>
+    ...
+
+    <script src="_framework/blazor.{webassembly|server}.js" autostart="false"></script>
+    <script>
+      Blazor.start({
+        environment: Staging
```
captainsafia, line 45:
```suggestion
        environment: "Staging"
```

aspnetcore/blazor/fundamentals/environments.md
```diff
@@ -16,9 +16,47 @@ uid: blazor/fundamentals/environments
...
+      });
+    </script>
+</body>
+```
+
+Setting the environment using this approach overrides setting the environment with the [`blazor-environment` header](#set-the-environment-via-header).
```
captainsafia, line 51:
```suggestion
Setting the environment using this approach overrides setting the provided when the environment is set via the [`blazor-environment` header](#set-the-environment-via-header).
```

I think this is clearer?

aspnetcore/blazor/fundamentals/startup.md
```diff
@@ -0,0 +1,122 @@
...
+<script>
+  Blazor.start({
+    loadBootResource: function (type, name, defaultUri, integrity) {
+      ...
+
+      var responseContent = ...;
+      var header_value_1 = ...;
+      var header_value_2 = ...;
+
+      return new Response(responseContent, {
```
captainsafia, line 78: This example is a little contrived. Generally, we'd expect the user to do something like issue a fetch call to a CDN or other service here instead of manually creating the response object.

aspnetcore/blazor/fundamentals/environments.md
guardrex: I'll try again ...

```suggestion
Using the `environment` property overrides the environment set by the [`blazor-environment` header](#set-the-environment-via-header).
```

aspnetcore/blazor/fundamentals/startup.md
guardrex: > a little contrived

That's by-design at this point. Idk exactly what to show. This new bit cross-links to the only example that we have at ...

https://docs.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/webassembly?view=aspnetcore-5.0#compression

I'll poke around on the PU repo and Net to see if I can find a real and alternative-to-compression example.

guardrex: *AH, HA!* ... I found DR's blog post for this. I'm going to revise and include all of his text.

guardrex:
@captainsafia ... I found DR's blog post on the boot resource loading **_AND_** that we did get it into the docs at 3.2 GA. It's buried back in the WASM host and deploy content. That's not the best spot any longer now that we have this new *Startup* topic. I've merged the best bits from everything that we have into the new section of the *Startup* topic. Try it now. :ear:

aspnetcore/blazor/fundamentals/startup.md
```diff
@@ -0,0 +1,175 @@
...
+        headers: { 'Custom-Header': 'Custom Value' }
+      });
+    }
+  });
+</script>
+```
+
+The `loadBootResource` function can return `null`/`undefined`, which results in the default loading behavior.
+
+Return a [`Response` promise](https://developer.mozilla.org/docs/Web/API/Response) for content created in JavaScript:
```
captainsafia, line 131: I still don't think this section is helpful and might end up confusing readers since it's not a common pattern to create a `Response` object from scratch in JavaScript.

I would just say something like:

```
The `loadBootResource` function can return `null`/`undefined`, which results in the default loading behavior, or a [Response promise](https://developer.mozilla.org/docs/Web/API/Response)

captainsafia (approved):

aspnetcore/blazor/fundamentals/startup.md
guardrex: Ok ... I'll 🔪 it, but this is what the Brotli *Compression* example does.

guardrex: Perhaps, the link to the Brotli example should be part of that bullet. I'll do it that way. If they want to see it, all they have to do is follow the link.

Issue #20451, opened 2020-11-09
Make it Clear that Azure SignalR Service Removes the Need for a Backplane


richardcox13:
The section "Azure SignalR Service" starts with the sentence "The Azure SignalR Service is a proxy rather than a backplane.". But as documented elsewhere (eg. [here](The Azure SignalR Service is a proxy rather than a backplane.) in the Important block) that when using Azure SignalR Service you do _not_ use a backplane.

A better and less ambiguous way to phrase this would be "The Azure SignalR Service is a proxy and removes the need for a backplane."

Edit: better text from @bradygaster

> The Azure SignalR Service functions as a proxy for your real-time traffic, and doubles as a backplane when your app is scaled out across multiple servers

---`
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 54bf3510-4a8a-468e-c9dd-84eb4514808b
* Version Independent ID: 9b4f1776-1851-5bd3-0cd9-fba69f23aadc
* Content: [ASP.NET Core SignalR production hosting and scaling](https://docs.microsoft.com/en-us/aspnet/core/signalr/scale?view=aspnetcore-2.1#azure-signalr-service)
* Content Source: [aspnetcore/signalr/scale.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/signalr/scale.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-signalr**
* GitHub Login: @bradygaster
* Microsoft Alias: **bradyg**

bradygaster:
Or how about "the service functions as a proxy for your real-time traffic, and doubles as a backplane when your app is scaled out across multiple servers."

Rick-Anderson:
@richardcox13  would you like to make this change? 
At the top right of the article, select the Edit pen icon:

![image](https://user-images.githubusercontent.com/3605364/53368931-11972200-38ee-11e9-86ce-18aaff69c719.png)

Select the pen icon again. Save.

PR #22294, opened 2021-05-13
Update vague text about Azure SignalR service and backplans

BrennanConroy:
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/20451


guardrex (approved):
I used "by" over "in" because the entire diagram seems to display the process, not part of the diagram. If that's incorrect, let me know ... I'll patch it. :ear:

Issue #21459, opened 2021-02-06
Inter-process communication with gRPC. Incomplete code example.


vformanyuk:
The code listed in this page should be updated because it isn't working as is.
In particular, "Server configuration" section should be modified as follows:
`options.ListenUnixSocket(SocketPath);`
need to be replaced with:
* For HTTP. As it turns, setting protocol to HTTP2 explicitly required no only on Mac, but on Windows as well.
`options.ListenUnixSocket(SocketPath, o => { o.Protocols = Microsoft.AspNetCore.Server.Kestrel.Core.HttpProtocols.Http2; });`
* For HTTPS. Thou "Client configuration" section of this page is about HTTP
`options.ListenUnixSocket(SocketPath, o => { o.UseHttps(); });`

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 583bb8dd-b80f-763e-4b84-1b0d2f4e823d
* Version Independent ID: 3d706460-ea6a-59f2-05d7-8a3c9f02e211
* Content: [Inter-process communication with gRPC](https://docs.microsoft.com/en-us/aspnet/core/grpc/interprocess?view=aspnetcore-5.0)
* Content Source: [aspnetcore/grpc/interprocess.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/grpc/interprocess.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-grpc**
* GitHub Login: @JamesNK
* Microsoft Alias: **jamesnk**

Issue #20290, opened 2020-10-23
Following the example sends requests as HTTP/1.1


connorearl:
When following the example as written, my server responds with:
```
info: Grpc.AspNetCore.Server.ServerCallHandler[3]
      Request protocol of 'HTTP/1.1' is not supported.
```

I've fought around with specifiying the HTTP version in an HttpClient Using the SocketHttpHandler with no luck, as seen below. I'm guessing I'm missing something but it's not clear in the documentation.
```
public static GrpcChannel CreateChannel()
{
    var udsEndPoint = new UnixDomainSocketEndPoint(SocketPath);
    var connectionFactory = new UnixDomainSocketConnectionFactory(udsEndPoint);
    var socketsHttpHandler = new SocketsHttpHandler
    {
        ConnectCallback = connectionFactory.ConnectAsync
    };

    var httpClient = new HttpClient(socketsHttpHandler)
    {
        DefaultRequestVersion = new Version("2.0"),
        DefaultVersionPolicy = HttpVersionPolicy.RequestVersionOrHigher
    };

    return GrpcChannel.ForAddress("http://localhost", new GrpcChannelOptions
    {
        HttpClient = httpClient
    });
}
```

Using SDK 5.0.100-rc.2.20479.15

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 583bb8dd-b80f-763e-4b84-1b0d2f4e823d
* Version Independent ID: 3d706460-ea6a-59f2-05d7-8a3c9f02e211
* Content: [Inter-process communication with gRPC](https://docs.microsoft.com/en-us/aspnet/core/grpc/interprocess?view=aspnetcore-5.0)
* Content Source: [aspnetcore/grpc/interprocess.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/grpc/interprocess.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-grpc**
* GitHub Login: @JamesNK
* Microsoft Alias: **jamesnk**

connorearl:
Hmm, after some deeper debuging it seems this is what was missing from the docs:
```
options.ListenUnixSocket(SocketPath, listenOptions => listenOptions.Protocols = HttpProtocols.Http2);
```

PR #22318, opened 2021-05-15
Update gRPC UDS example

JamesNK:
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/21459
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/20290

Added explicit protocol to sample in case it isn't the default


halter73 (approved):
I'm guessing we're not bothering with HTTPS since it's inter-process

JamesNK:
Yeah. And also the simplicity of the sample.

But I'll update it to call out below the sample that HTTPS isn't used and there is the option to enable it.

PR #22331, opened 2021-05-17
refactor: remove always true condition

FrediKats:
image

Result of .ToArray() is always not null.

### DIFF
--- a/aspnetcore/blazor/security/webassembly/hosted-with-identity-server.md
+++ b/aspnetcore/blazor/security/webassembly/hosted-with-identity-server.md
@@ -301,9 +301,9 @@ public class CustomUserFactory
         if (user.Identity.IsAuthenticated)
         {
             var identity = (ClaimsIdentity)user.Identity;
-            var roleClaims = identity.FindAll(identity.RoleClaimType).ToArray();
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks @FrediKats ... but let's go the **_Javier Way!&trade;_**, which I believe is more efficient. I can't explain why the `ToArray` is there really ........ **_WAIT!_** *Yes! Indeed, I do know how that got there!* ........

# 👽 

*They are among us!* :smile:

Cross-reference: https://github.com/javiercn/BlazorAuthRoles/blob/master/Client/RolesClaimsPrincipalFactory.cs#L24-L25

Bibletoon:
Can you please explain what does Javier Way!™ mean. I really can't understand it.

guardrex:
I was just joking that we should adopt Javer's approach to solve this problem with the example. He's one of Microsoft's top engineers, and I try to get his suggestions into the docs when he has an example for a scenario like this. 

I think to `ToArray` is slow and unnecessary. If you disagree with the update that I made,vi can ping Javier to look at this :eyes:.

Bibletoon:
Ok, thanks for the explanation.

PR #22338, opened 2021-05-18
Blazor 'integrity.ps1' script security

guardrex:
Addresses #19286

Courtesy ping for @blowdart 🎯 ... *I'm working with Blazor security subject matter here.* 😨😄 

[Internal Review Topic (links to section, **See NOTE**)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/webassembly?view=aspnetcore-3.1&branch=pr-en-us-22338#troubleshoot-integrity-powershell-script)

* Multiple users have hit this problem where a heuristic scanner throws a false positive.
* Not going to go with an INCLUDE for now. This is the only spot in Blazor docs that we need to cover this at this time. [Note tho that this is a wider problem across the repo, i.e. ... non-Blazor scenarios.]
* Explains that an exception is required for the file, **_BUT_** calls out a strong warning on file security when creating exceptions **_AND_** briefly covers performing a checksum comparison ....... **_and_** calls out that even that doesn't guarantee 100% safety.

### DIFF
--- a/aspnetcore/blazor/host-and-deploy/webassembly.md
+++ b/aspnetcore/blazor/host-and-deploy/webassembly.md
@@ -769,24 +769,28 @@ Placeholders:
 * `{PUBLISH OUTPUT FOLDER}`: The path to the app's `publish` folder or location where the app is published for deployment.
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


mkArtakMSFT (approved):
Thanks @guardrex !

guardrex:
@HaoK ... See if I went too far 😅 with the NOTE and/or phrased it well. I don't want to leave devs hanging, but I don't want to get too deep in the weeds either .... and checksums aren't a guarantee anyway.

PR #22346, opened 2021-05-19
Fix chat.js in SignalR /a

### LINKED ISSUE #22337: SignalR "Get started" tutorial has incorrect HTML encoding logic
In the code for `chat.js` under [this heading](https://docs.microsoft.com/en-us/aspnet/core/tutorials/signalr?view=aspnetcore-5.0&tabs=visual-studio#add-signalr-client-code), we have the following code:

```js
var msg = message.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;");
var encodedMsg = user + " says " + msg;
var li = document.createElement("li");
li.textContent = encodedMsg;
document.getElementById("messagesList").appendChild(li);
```

This is incorrect because:

 * There's no point performing any replacements for `<` or `&` characters if you're going to assign the string to the `textContext` property of a DOM element anyway. The `textContent` is literally text content, not HTML markup, so you can assign any string value to it without creating an XSS concern.
 * Doing these replacements creates misunderstanding, because (1) it suggests you need to in this case, which you don't, and (2) it suggests that in cases where XSS is a concern, the logic presented here would be sufficient, which it would not. Encoding to avoid XSS concerns would require a more sophisticated mechanism to avoid all the kinds of injections that you might find in a list like [this](https://portswigger.net/web-security/cross-site-scripting/cheat-sheet) (note that it depends on exactly where in the DOM you're going to inject the content).
* Doing this replacement means that people can't enter things like `1 < 2` as chat messages as it will be incorrectly rendered as `1 &lt; 2` in the UI.
* For some reason the logic here implies you need to encode the `message` but not the `user` string, but `user` is also free-form text provided by the user

To avoid confusion I suggest this code be replaced with something like:

```js
var li = document.createElement("li");
document.getElementById("messagesList").appendChild(li);

// We can assign user-supplied strings to an element's textContent because this will not be interpreted as markup
// If assigning in any other way, you should be aware of possible script injection concerns
li.textContent = user + " says " + message;
```

---
#### Document Details

* ID: 0dd2991d-c22d-ad0f-80b6-bd1b503db8f3
* Version Independent ID: 5e3fbe1d-c97f-9940-4f0c-067d4b2112b1
* Content: [Get started with ASP.NET Core SignalR](https://docs.microsoft.com/en-us/aspnet/core/tutorials/signalr?view=aspnetcore-5.0&tabs=visual-studio)
* Content Source: [aspnetcore/tutorials/signalr.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/tutorials/signalr.md)
* Product: **aspnet-core**
* GitHub Login: @bradygaster
* Microsoft Alias: **bradyg**

Rick-Anderson:
Fixes #22337
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/tutorials/signalr?view=aspnetcore-5.0&branch=pr-en-us-22346&tabs=visual-studio)

### DIFF
--- a/aspnetcore/tutorials/signalr/sample-snapshot/3.x/chat.js
+++ b/aspnetcore/tutorials/signalr/sample-snapshot/3.x/chat.js
@@ -6,11 +6,12 @@ var connection = new signalR.HubConnectionBuilder().withUrl("/chatHub").build();
 document.getElementById("sendButton").disabled = true;
 
 connection.on("ReceiveMessage", function (user, message) {
-    var msg = message.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;");
-    var encodedMsg = user + " says " + msg;
     var li = document.createElement("li");
-    li.textContent = encodedMsg;
     document.getElementById("messagesList").appendChild(li);
+    // We can assign user-supplied strings to an element's textContent because this won't
+    // be interpreted as markup.
+    // If assigning in any other way, you should be aware of possible script injection concerns.
+    li.textContent = user + " says " + message;
 });
 
 connection.start().then(function () {
--- a/aspnetcore/tutorials/signalr/sample-snapshot/3.x/chat.js
+++ b/aspnetcore/tutorials/signalr/sample-snapshot/3.x/chat.js
@@ -8,10 +8,10 @@ document.getElementById("sendButton").disabled = true;
 connection.on("ReceiveMessage", function (user, message) {
     var li = document.createElement("li");
     document.getElementById("messagesList").appendChild(li);
-    // We can assign user-supplied strings to an element's textContent because this won't
-    // be interpreted as markup.
-    // If assigning in any other way, you should be aware of possible script injection concerns.
-    li.textContent = user + " says " + message;
+    // We can assign user-supplied strings to an element's textContent because it
+    // is not interpreted as markup. If you're assigning in any other way, you 
+    // should be aware of possible script injection concerns.
+    li.textContent = `${user} says ${message}`;
 });
 
 connection.start().then(function () {


IEvangelist (approved):
Looks good to me, :shipit: when you're ready.

aspnetcore/tutorials/signalr/sample-snapshot/3.x/chat.js
```diff
@@ -6,11 +6,12 @@ var connection = new signalR.HubConnectionBuilder().withUrl("/chatHub").build();
...
 connection.on("ReceiveMessage", function (user, message) {
-    var msg = message.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;");
-    var encodedMsg = user + " says " + msg;
     var li = document.createElement("li");
-    li.textContent = encodedMsg;
     document.getElementById("messagesList").appendChild(li);
+    // We can assign user-supplied strings to an element's textContent because this won't
+    // be interpreted as markup.
+    // If assigning in any other way, you should be aware of possible script injection concerns.
+    li.textContent = user + " says " + message;
```
IEvangelist, line 14:
```suggestion
    // We can assign user-supplied strings to an element's textContent because it
    // is not interpreted as markup. If you're assigning in any other way, you 
    // should be aware of possible script injection concerns.
    li.textContent = `${user} says ${message}`;
```

Use JavaScript string interpolation, minor cleanup of comments.

bradygaster:
Looks awesome thanks

bradygaster:
Thanks for the review @Rick-Anderson, and to both of you for the help updating it.

PR #22362, opened 2021-05-21
Fix link

dharmatech:


dharmatech:
The original link:

https://docs.microsoft.com/en-us/ef/core/cli/dotnet#resetting-all-migrations

doesn't seem to lead to a particular section.

I've updated it to the following one:

https://docs.microsoft.com/en-us/ef/core/managing-schemas/migrations/managing?tabs=dotnet-core-cli#resetting-all-migrations

Rick-Anderson:
@dharmatech  thanks!

PR #22371, opened 2021-05-21
Remove outdated instructions

dharmatech:
This section:

https://docs.microsoft.com/en-us/aspnet/core/data/ef-rp/complex-data-model?view=aspnetcore-5.0&tabs=visual-studio-code#create-a-migration

contains the following

image

After running the commands shown, I did not get the SQLite error message shown.

The docs mention `AlterColumnOperation`. Note that the migration file in my case uses `AlterColumn`:

image

Note that column is now called `FirstName`:

image

So the instructions now seem to work without issue on SQLite.

### DIFF
--- a/aspnetcore/data/ef-rp/complex-data-model.md
+++ b/aspnetcore/data/ef-rp/complex-data-model.md
@@ -218,24 +218,6 @@ SqliteException: SQLite Error 1: 'no such column: s.FirstName'.
   dotnet ef database update
   ```
 
-  The database update command displays an error like the following example:
-
-  ```text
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
I just tested this a month ago. Can you move this information to an issue and we'll see if anyone else can reproduce it. It's not blocking if you don't get the error.

Rick-Anderson:
@dharmatech  reverted the delete and commented out that section.

Rick-Anderson:
@dharmatech thanks for the PR.

PR #22391, opened 2021-05-25
fix docker 128 error /9

### LINKED ISSUE #19814: Linux OS with .NET- The configured user limit (128) on the number of inotify instances has been reached
This issue is directly related to and a followup of https://github.com/dotnet/aspnetcore/issues/7531 since that issue has been locked. The same issue occurs now in .NET Core 3.1. Following the previous thread, I was able to specify a different user in my Dockerfile to limit this problem which occurs when putting too many docker containers on the same host that are using the same user UID. Our teams now use a randomized UID to ensure each immutable build has its own ID:

```
RUN useradd --uid $(shuf -i 2000-65000 -n 1) app
USER app
```

This works to limit the issue, but the problem still continues if you intend to run the same instance of that immutable container with 9 or more instances on the same container orchestrator host. For example if I use the following image: "FROM mcr.microsoft.com/dotnet/core/aspnet:3." to run a base default / empty ASP.NET Core application with the following Program.cs (and usage of zero-configuration files):

```
Host.CreateDefaultBuilder(args)
                .ConfigureWebHost(builder =>
                {
                    builder
                        .UseKestrel()
                        .UseStartup<Startup>();
                });
```

Run a default app in the container above, and attach your shell and execute the following to count the number of watchers:
```
apt-get install lsof
lsof | grep inotify
```
.NET Core 3.1 shows me generally 15 or 16, while .NET Core 2.1 default app shows more like 40.
If I am running 8 containers each taking at least 15 instances of the 128 limit - I run out pretty quickly.

General concerns:
1. Scaling is the first concern I have for obvious reasons (depending on my placement strategy). Is the expectation that I should be increasing this limit on my orchestrator hosts to allow for this?
2. With a default execution of ASP.NET Core 3.1 that uses no configuration files, why am I seeing 15 instances? Typically speaking in a container there is very little reason I even want to watch files (especially not configuration files) since its entirely immutable. Is there a way to entirely turn off the inotify watcher instances all together (which would solve this problem entirely)? Or is there an underlying bug in ASP.NET Core here?

***EDIT by @Rick-Anderson***  #28387 mitigates the problem.

---
[Associated WorkItem - 64021](https://dev.azure.com/msft-skilling/Content/_workitems/edit/64021)

Rick-Anderson:
Fixes #19814

[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/host-and-deploy/docker/?view=aspnetcore-1.0&branch=pr-en-us-22391#systemioioexception-the-configured-user-limit-128-on-the-number-of-inotify-instances-has-been-reached)

Per @travisgosselin

> In the ASP.NET Core world driven inside a container, there is next to zero reasons I can think of in which I would want to enforce reloading of that configuration file. Once my container is built and pushed to its container repository with an immutable version it does not change (only environment variables on deploy modify its behavior). In this straightforward scenario, there is very little reason to turn on “reloadOnChange”. Turning it off, lead to slightly less number of occurrences of this problem.

@travisgosselin writes
> The documentation is not great in telling you the default.

Just  above [Combining service collection](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-5.0#combining-service-collection)

Using the default configuration, the appsettings.json and appsettings.Environment.json files are enabled with reloadOnChange: true. Changes made to the appsettings.json and appsettings.Environment.json file after the app starts are read by the JSON configuration provider.

See #22392

### DIFF
--- a/aspnetcore/includes/docker-128/Program.cs
+++ b/aspnetcore/includes/docker-128/Program.cs
@@ -1,0 +1,27 @@
+using Docker128;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Hosting;
+
+#region snippet
+public class Program
+{
+    public static void Main(string[] args)
+    {
+        CreateHostBuilder(args).Build().Run();
+    }
+
+public static IHostBuilder CreateHostBuilder(string[] args) =>
+    Host.CreateDefaultBuilder(args)
+        .ConfigureAppConfiguration((hostingContext, config) =>
+        {
+            config.AddJsonFile("appsettings.json",
+                optional: true,
+                reloadOnChange: false);
+        })
+        .ConfigureWebHostDefaults(webBuilder =>
+        {
+            webBuilder.UseStartup<Startup>();
+        });
+}
+#endregion
--- a/aspnetcore/host-and-deploy/docker/building-net-docker-images.md
+++ b/aspnetcore/host-and-deploy/docker/building-net-docker-images.md
@@ -326,6 +326,7 @@ ENTRYPOINT ["dotnet", "aspnetapp.dll"]
 * [Working with Visual Studio Docker Tools](./visual-studio-tools-for-docker.md)
 * [Debugging with Visual Studio Code](https://code.visualstudio.com/docs/nodejs/debugging-recipes#_debug-nodejs-in-docker-containers)
 * [GC using Docker and small containers](xref:performance/memory#sc)
+* [System.IO.IOException: The configured user limit (128) on the number of inotify instances has been reached](xref:aspnet/core/host-and-deploy/docker#d128)
 
 ## Next steps
 
--- a/aspnetcore/host-and-deploy/docker/index.md
+++ b/aspnetcore/host-and-deploy/docker/index.md
@@ -38,3 +38,5 @@ Additional configuration might be required for apps hosted behind proxy servers 
 
 [GC using Docker and small containers](xref:performance/memory#sc)
 Discusses GC selection with small containers.
+
+[!INCLUDE[](~/includes/docker-128.md)]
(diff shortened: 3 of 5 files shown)


Rick-Anderson:
@BrennanConroy  can you recommend a reviewer?

travisgosselin:
While this addresses the first and simplest change to help reduce the 128 user limit, I think its additionally necessary to address the potential root user usage in the Dockerfile and encourage the usage of other users on their runtime image given scaling implications when deployed to a single host with multiple container instances running (random user option described in https://github.com/dotnet/AspNetCore.Docs/issues/19814). Additionally using a non-root user in your Dockerfile also has security implications on what ports are available by default to expose (i.e. port 80 not accessible by default..).

Rick-Anderson:
@danmoseley  @danmoseley @ericstj  @pranavkm @glennc @spboyer  this is a serious problem Docker folks are frequently hitting that  needs to be documented how to avoid. See #19814

Rick-Anderson:
@danmoseley @danmoseley @ericstj @pranavkm @glennc @spboyer @BrennanConroy  3 minute review  this is a serious problem Docker folks are frequently hitting that needs to be documented how to avoid. See #19814

aspnetcore/includes/docker-128/Program.cs
```diff
@@ -0,0 +1,27 @@
...
+    public static void Main(string[] args)
+    {
+        CreateHostBuilder(args).Build().Run();
+    }
+
+public static IHostBuilder CreateHostBuilder(string[] args) =>
+    Host.CreateDefaultBuilder(args)
+        .ConfigureAppConfiguration((hostingContext, config) =>
+        {
+            config.AddJsonFile("appsettings.json",
```
Tratcher, line 18: This doesn't fix the issue because CreateDefaultBuilder already called `config.AddJsonFile("appsettings.json", optional: true, reloadOnChange: true);`, calling it a second time doesn't turn off the prior calls reload notifications. DOTNET_HOSTBUILDER__RELOADCONFIGONCHANGE=false is the proper fix.

PR #22420, opened 2021-05-27
convinience methods to convenient methods

meenakshirana8587:


guardrex:
Hello @meenakshirana8587 ... The phrase "convenience method(s)" is a term used across docs (26 places) to refer to methods that the framework offers for convenient operations so that the dev doesn't have to create them from scratch. Thanks tho for writing in to ask about it.

PR #22478, opened 2021-06-02
Blazor route constraints with optional params

### LINKED ISSUE #22477: Are optional routes with constraints possible?
From the current page, it is not clear if one can combine both optional routes with constrains... e.g
```
@page "/RouteParameter/{text?}" 
```
 is possible

but can I do?
```
@page "/RouteParameter/{text:bool?}" 
```

---
#### Document Details

* Content: [ASP.NET Core Blazor routing](https://docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-5.0)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #22477

Thanks @ErikApption! :rocket:

Pranav: Optional parameters are covered in the *Route parameters* section, so that's why the bookmark link is set up the way that it is.

### DIFF
--- a/aspnetcore/blazor/fundamentals/routing.md
+++ b/aspnetcore/blazor/fundamentals/routing.md
@@ -191,7 +191,7 @@ The route constraints shown in the following table are available. For the route 
 
 ::: moniker range=">= aspnetcore-5.0"
 
-Route contraints also work with [optional parameters](#optional-parameters):
+Route contraints also work with [optional parameters](#route-parameters):
 
 ```razor
 @page "/user/{Id:int?}"
--- a/aspnetcore/blazor/fundamentals/routing.md
+++ b/aspnetcore/blazor/fundamentals/routing.md
@@ -150,6 +150,9 @@ protected override void OnParametersSet()
 }
 ```
 
+> [!NOTE]
+> Route parameters don't work with query string values. To work with query strings, see the [Query string and parse parameters](#query-string-and-parse-parameters) section.
+
 ## Route constraints
 
 A route constraint enforces type matching on a route segment to a component.
@@ -173,6 +176,9 @@ In the following example, the route to the `User` component only matches if:
 
 ::: moniker-end
 
+> [!NOTE]
+> Route contraints don't work with query string values. To work with query strings, see the [Query string and parse parameters](#query-string-and-parse-parameters) section.
+
 The route constraints shown in the following table are available. For the route constraints that match the invariant culture, see the warning below the table for more information.
 
 | Constraint | Example           | Example Matches                                                                  | Invariant<br>culture<br>matching |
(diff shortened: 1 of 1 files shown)


guardrex:
@pranavkm ... Additional question from @ErikApption ...

We tried to create routing rules with optional parameters with ? 
`@page "/mypage/{id:int}?option1={option1:bool?}"` but this actually generates an error in blazor. I see a section in routing about handling urls - are `?` characters in URLs expected to be handled manually using the whole URL? Or which characters are actually allowed in routing rules?

guardrex:
@ErikApption ... AFAIK, the constraints and optional params won't work in a query string. For that, you'll need to look at ...

https://docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-5.0#query-string-and-parse-parameters

ErikApption:
> FAIK, the constraints and optional params won't work in a query string.

Thanks - that's exactly what I was looking for... any chances to clarify that in the docs?
But I do not totally understand the reasoning behind it - what is the difference between
`@page "/mypage/{id:int}?option1={option1:bool?}"` and `@page "/mypage/{id:int}/{option1:bool?}"` - isn't the second version supposed to be supported?

guardrex:
Yes ... I'll add more to this tomorrow morning.

* The 2nd one :point_up: is in line with what the PU ("product unit") has indicated should work (not a qs).
* The first one might choke because I don't recall them ever saying that the route params would work at all from a qs. They provided the section that I linked for qs and parsing parameters manually.
* On the last bit about supported chars ... that's best for @pranavkm to comment on. After he and you discuss it further here, I'll make additional updates to the PR to cover the new info that he provides.

I'll mark this a draft for now given that we're still working on the updated coverage. Stand-by ... he should be along in a bit ... within a day or two. 🏃😅

guardrex:
@ErikApption ...

* I add three NOTEs in strategic locations that refer the reader to the qs section for working with qs params+values.
* I flesh out the route contraint+optional param example into a full component example that uses two route parameters, where one of them is an optional boolean.

@pranavkm ... I think the only remaining question is ...

> aren't we looking at minor differences in supported characters in routing

AFAIK, the Router uses specific logic for the **_route part_** of the URL that doesn't interact (much) with the qs part of the URL (i.e., it at least doesn't work with route params and optional route params as the new NOTEs indicate). I'll react to anything you want to add/update.

guardrex:
@ErikApption ... I've received word from the PU internally that they might be working on something for release later to address this subject. I don't have further information at this time. You can keep an eye on PU issues and blog posts for further developments as they happen. For now, we'll just go with the updates on this PR to cover your ask for better guidance.

guardrex:
... and btw ... if you search the PU issues, you'll see about halfway down an issue that might be regarding features. Check these out and see what they're working on and discussing ...

https://github.com/dotnet/aspnetcore/issues?q=is%3Aissue+is%3Aopen+query+string

PR #22489, opened 2021-06-03
Blazor interaction with the DOM

guardrex:
Addresses #19286

Thanks @angelotrivelli! :rocket: ... We just discussed the use of Bootstrap JS offline, and the PU ("product unit") confirmed that it's just like any other JS lib/framework. If the JS interop use of Bootstrap interacts with the same DOM objects that Blazor is tracking, bad results can ensue. If the Bootstrap JS is touching something that Blazor isn't tracking, then it's fine.

I'm improving the coverage on this subject ...

* I add a section to the new JS interop *Overview* topic on this.
* I mention Bootstrap JS and jQuery as examples.
* I mention that a few doc examples do violate the general guidance purely for demonstration purposes as part of an example, and I mention that the text calls out when the docs do this.
* I cross-link to the new section from relevant spots.

cc: @mkArtakMSFT ... This takes care of what we discussed. I'll mod further if additional feedback appears in the offline discussion.

### DIFF
--- a/aspnetcore/blazor/javascript-interoperability/index.md
+++ b/aspnetcore/blazor/javascript-interoperability/index.md
@@ -18,6 +18,16 @@ This overview article covers general concepts. Further JS interop guidance is pr
 * <xref:blazor/js-interop/call-javascript-from-dotnet>
 * <xref:blazor/js-interop/call-dotnet-from-javascript>
 
+## Interaction with the Document Object Model (DOM)
+
+Only mutate the Document Object Model (DOM) with JavaScript (JS) when the object doesn't interact with Blazor. Blazor maintains representations of the DOM and interacts directly with DOM objects. If JS interop mutates an element or its content and Blazor attempts to apply diffs to the element, the diffs won't match the DOM and undefined behavior results. Arbitrary behaviors may merely interfere with the presentation of elements or their behavior but may also introduce security risks to the app or server.
+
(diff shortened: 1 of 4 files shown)


aspnetcore/blazor/javascript-interoperability/index.md
```diff
@@ -18,6 +18,16 @@ This overview article covers general concepts. Further JS interop guidance is pr
 * <xref:blazor/js-interop/call-javascript-from-dotnet>
 * <xref:blazor/js-interop/call-dotnet-from-javascript>
 
+## Interaction with the Document Object Model (DOM)
+
+Only mutate the Document Object Model (DOM) with JavaScript (JS) when the object doesn't interact with Blazor. Blazor maintains representations of the DOM and interacts directly with DOM objects. If JS interop mutates an element or its content and Blazor attempts to apply diffs to the element, the diffs won't match the DOM and undefined behavior results. Arbitrary behaviors may merely interfere with the presentation of elements or their behavior but may also introduce security risks to the app or server.
```
pranavkm, line 23: > If JS interop mutates an element or its content and Blazor attempts to apply diffs to the element, the diffs won't match the DOM and undefined behavior results

Perhaps, this should say 

If an element or its content that is rendered by Blazor is modified externally, e.g. using JavaScript including JS Interop, the DOM may no longer match Blazor's internal representation for it and would result in undefined behavior.

PR #22492, opened 2021-06-03
Typo on CurrentCount upper/lower case

ctrl-alt-d:
There was a typo on `CurrentCount` lower case, should be upper case.


guardrex (approved):
Thanks @ctrl-alt-d! :rocket: ... That's a rotten sneaky *gremlin* 😈! :smile:

PR #22502, opened 2021-06-05
fileContent gets disposed prior to upload

kenjohnson03:
Added the StreamContent as ByteArrayContent to allow the fileContent variable to get disposed as expected without impacting the PostAsync call.


guardrex:
Thanks @kenjohnson03 ... We'll need an issue for this one. Moving to :point_right: https://github.com/dotnet/AspNetCore.Docs/issues/22503

PR #22514, opened 2021-06-08
Fix ProfileService registration sample

peterbozso:
Without this, role-based backend API authorization won't work, the API will just return 403.


guardrex:
Hello @peterbozso ... It's already here a little earlier in the *API authorization options* section.

guardrex:
Oh, I see ...... we say *one of the following* approaches, and this is the 2nd approach of the pair.

guardrex (approved):

guardrex:
Just noting for the record that this must have been a delta between 3.2 and 5.0. My 3.2 test app works without this line. This bit of the topic wasn't tested for 5.0 here.

peterbozso:
Thank you @guardrex!

PR #22525, opened 2021-06-10
Blazor Global/Loc UE pass

### LINKED ISSUE #21619: Blazor WebAssembly localization from browser culture requires all globalization data
The first sentence of the localization docs for WebAssembly is

> Blazor WebAssembly apps set the culture using the user's language preference.

For that to work for browsers that are not set to `en-US` (that is, man people outside of the USA), you must still add

```
<PropertyGroup>
  <BlazorWebAssemblyLoadAllGlobalizationData>true</BlazorWebAssemblyLoadAllGlobalizationData>
</PropertyGroup>
```

Which is, admittedly, a few paragraphs down, but I think it makes sense to explicitly add that sentence in the first paragraph. I would personally combine both options in a list, something like this (I have not added the links here for readability):

```
By default, Blazor WebAssembly apps set the culture using the user's language preference. To explicitly configure the culture, set CultureInfo.DefaultThreadCurrentCulture and CultureInfo.DefaultThreadCurrentUICulture in Program.Main.

Both approaches change the culture of the app dynamically, so you need to prepare it for that.

<The rest of the current docs could go here - about the csproj flag, about "hardcoding" a culture in blazor.start(), and so on>
```

Once you scratch your head a little why you see `en-GB`, for example, but the dates are formatted according to the `en-US` format it will become apparent that this is a dynamic change from the app perspective, but if you are working on a large app that you are only now starting to localize you can have a hard time figuring this out.

Frankly, there is quite a lot to unpack in those few paragraphs and I would personally probably divide it into a couple of sections if I were writing the article, maybe something like "invariant culture" (currently the last snippet, but it is the easiest), "static culture" (the one with Blazor.start()) and "dynamic culture" where all the info above would go into that last section. This lets you explain the culture format early on, and go from basic that works for most cases to complex and real globalization.

---
#### Document Details

* Content: [ASP.NET Core Blazor globalization and localization](https://docs.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-5.0#blazor-webassembly)
* Content Source: [aspnetcore/blazor/globalization-localization.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/globalization-localization.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Addresses #19286
Fixes #21619

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-3.1&branch=pr-en-us-22525)

### DIFF
--- a/aspnetcore/blazor/fundamentals/startup.md
+++ b/aspnetcore/blazor/fundamentals/startup.md
@@ -143,7 +143,7 @@ The `loadBootResource` function can also return:
   * [Adjust the reconnection retry count and interval](xref:blazor/fundamentals/signalr#adjust-the-reconnection-retry-count-and-interval)
   * [Hide or replace the reconnection display](xref:blazor/fundamentals/signalr#hide-or-replace-the-reconnection-display)
   * [Disconnect the Blazor circuit from the client](xref:blazor/fundamentals/signalr#disconnect-the-blazor-circuit-from-the-client)
-* [Globalization and localization: Statically set the culture](xref:blazor/globalization-localization#statically-set-the-culture)
+* [Globalization and localization: Statically set the culture with `Blazor.start` (*Blazor WebAssembly only*)](xref:blazor/globalization-localization#statically-set-the-culture)
 * [JS interop: Inject a script after Blazor starts](xref:blazor/js-interop/index#inject-a-script-after-blazor-starts)
 * [Host and deploy: Blazor WebAssembly: Compression](xref:blazor/host-and-deploy/webassembly#compression)
 
--- a/aspnetcore/blazor/fundamentals/startup.md
+++ b/aspnetcore/blazor/fundamentals/startup.md
@@ -143,7 +143,7 @@ The `loadBootResource` function can also return:
   * [Adjust the reconnection retry count and interval](xref:blazor/fundamentals/signalr#adjust-the-reconnection-retry-count-and-interval)
   * [Hide or replace the reconnection display](xref:blazor/fundamentals/signalr#hide-or-replace-the-reconnection-display)
   * [Disconnect the Blazor circuit from the client](xref:blazor/fundamentals/signalr#disconnect-the-blazor-circuit-from-the-client)
-* [Globalization and localization: Set the culture in a Blazor WebAssembly app](xref:blazor/globalization-localization#blazor-webassembly)
+* [Globalization and localization: Statically set the application culture](xref:blazor/globalization-localization#statically-set-the-application-culture)
 * [JS interop: Inject a script after Blazor starts](xref:blazor/js-interop/index#inject-a-script-after-blazor-starts)
 * [Host and deploy: Blazor WebAssembly: Compression](xref:blazor/host-and-deploy/webassembly#compression)
 
--- a/aspnetcore/blazor/fundamentals/startup.md
+++ b/aspnetcore/blazor/fundamentals/startup.md
@@ -143,7 +143,7 @@ The `loadBootResource` function can also return:
   * [Adjust the reconnection retry count and interval](xref:blazor/fundamentals/signalr#adjust-the-reconnection-retry-count-and-interval)
   * [Hide or replace the reconnection display](xref:blazor/fundamentals/signalr#hide-or-replace-the-reconnection-display)
   * [Disconnect the Blazor circuit from the client](xref:blazor/fundamentals/signalr#disconnect-the-blazor-circuit-from-the-client)
-* [Globalization and localization: Statically set the culture with `Blazor.start` (*Blazor WebAssembly only*)](xref:blazor/globalization-localization#statically-set-the-culture)
+* [Globalization and localization: Statically set the culture with `Blazor.start` (*Blazor WebAssembly only*)](xref:blazor/globalization-localization?pivots=webassembly#statically-set-the-culture)
 * [JS interop: Inject a script after Blazor starts](xref:blazor/js-interop/index#inject-a-script-after-blazor-starts)
 * [Host and deploy: Blazor WebAssembly: Compression](xref:blazor/host-and-deploy/webassembly#compression)
 
--- a/aspnetcore/blazor/fundamentals/startup.md
+++ b/aspnetcore/blazor/fundamentals/startup.md
@@ -143,7 +143,7 @@ The `loadBootResource` function can also return:
   * [Adjust the reconnection retry count and interval](xref:blazor/fundamentals/signalr#adjust-the-reconnection-retry-count-and-interval)
   * [Hide or replace the reconnection display](xref:blazor/fundamentals/signalr#hide-or-replace-the-reconnection-display)
   * [Disconnect the Blazor circuit from the client](xref:blazor/fundamentals/signalr#disconnect-the-blazor-circuit-from-the-client)
-* [Globalization and localization: Statically set the application culture](xref:blazor/globalization-localization#statically-set-the-application-culture)
+* [Globalization and localization: Statically set the culture](xref:blazor/globalization-localization#statically-set-the-culture)
 * [JS interop: Inject a script after Blazor starts](xref:blazor/js-interop/index#inject-a-script-after-blazor-starts)
 * [Host and deploy: Blazor WebAssembly: Compression](xref:blazor/host-and-deploy/webassembly#compression)
 
--- a/aspnetcore/blazor/globalization-localization.md
+++ b/aspnetcore/blazor/globalization-localization.md
@@ -30,7 +30,7 @@ For additional general information, see <xref:fundamentals/localization>.
 > [!NOTE]
 > Often, the terms *language* and *culture* are used interchangably when dealing with globalization and localization concepts.
 >
-> In this article, *language* refers to selections made by a user in their browser's settings. The user's selections are submitted in browser requests in the [`Accept-Language` header](https://developer.mozilla.org/docs/Web/HTTP/Headers/Accept-Language). Browser settings usually use the word "language" in the UI.
+> In this article, *language* refers to selections made by a user in their browser's settings. The user's language selections are submitted in browser requests in the [`Accept-Language` header](https://developer.mozilla.org/docs/Web/HTTP/Headers/Accept-Language). Browser settings usually use the word "language" in the UI.
 >
 > *Culture* pertains to members of .NET and Blazor API. For example, a user's request can include the [`Accept-Language` header](https://developer.mozilla.org/docs/Web/HTTP/Headers/Accept-Language) specifying a *language* from the user's perspective, but the app ultimately sets the <xref:System.Globalization.CultureInfo.CurrentCulture> ("**culture**") property from the language that the user requested. API usually uses the word "culture" in its member names.
 
@@ -130,9 +130,7 @@ Add a list item to the navigation menu `<ul>` element in `Shared/NavMenu.razor` 
 
 The [`Accept-Language` header](https://developer.mozilla.org/docs/Web/HTTP/Headers/Accept-Language) is set by the browser and controlled by the user's language preferences in browser settings. In browser settings, a user sets one or more preferred languages in order of preference. The order of preference is used by the browser to set quality values (`q`, 0-1) for each language in the header. The following example specifies United States English, English, and Peruvian Spanish with a preference for United States English or English:
 
-```
-Accept-Language: en-US,en;q=0.9,es-PE;q=0.8
-```
+**Accept-Language**: en-US,en;q=0.9,es-PE;q=0.8
 
 The app's culture is set by matching the first requested language that matches a supported culture of the app.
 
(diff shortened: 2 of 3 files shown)


guardrex:
@hishamco ... Do you want to put an 👁️ on this? I hate to do anything in this area without having your great advice. If ur too busy tho, no worries ... the PU will get in on this and carve 🔪😨 me up. 😄 I'll give you time to look before pinging them.

aspnetcore/blazor/fundamentals/startup.md
```diff
@@ -143,7 +143,7 @@ The `loadBootResource` function can also return:
   * [Adjust the reconnection retry count and interval](xref:blazor/fundamentals/signalr#adjust-the-reconnection-retry-count-and-interval)
   * [Hide or replace the reconnection display](xref:blazor/fundamentals/signalr#hide-or-replace-the-reconnection-display)
   * [Disconnect the Blazor circuit from the client](xref:blazor/fundamentals/signalr#disconnect-the-blazor-circuit-from-the-client)
-* [Globalization and localization: Set the culture in a Blazor WebAssembly app](xref:blazor/globalization-localization#blazor-webassembly)
+* [Globalization and localization: Statically set the culture](xref:blazor/globalization-localization#statically-set-the-culture)
```
hishamco, line 146: Do you mean Blazor Assembly or Blazor Server-Side app?

aspnetcore/blazor/globalization-localization.md
```diff
@@ -40,53 +51,165 @@ The preceding field types:
...
+`Pages/CultureExample1.razor`:
+
+```razor
+@page "/culture-example-1"
+@using System.Globalization
+
+<h1>Culture Example 1</h1>
+
+<p>
+    <b>CurrentCulture</b>: @CultureInfo.CurrentCulture
```
hishamco, line 77: Does `CultureInfo.CurrentUICulture` works too, I remembered there's a limitation in web assembly for related issue, I may search for the tweet that I made long time ago

aspnetcore/blazor/globalization-localization.md
```diff
@@ -40,53 +51,165 @@ The preceding field types:
...
 
-```xml
-<PropertyGroup>
-  <InvariantGlobalization>true</InvariantGlobalization>
-</PropertyGroup>
+```csharp
+using System.Globalization;
+```
+
+Add the culture settings before the line that builds and runs the <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.WebAssemblyHostBuilder> (`await builder.Build().RunAsync();`):
```
hishamco, line 208: This doesn't shows how to call javascript funtion that is used to set the culture

aspnetcore/blazor/globalization-localization.md
```diff
@@ -97,63 +220,271 @@ By default, the Intermediate Language (IL) Linker configuration for Blazor WebAs
...
+```csharp
+services.AddLocalization();
+```
 
-For more information and examples, see <xref:fundamentals/localization>.
+Specify the static culture in `Startup.Configure` (`Startup.cs`) immediately after Routing Middleware is added to the processing pipeline. The following example configures United States English:
 
-#### Cookies
+```csharp
+app.UseRequestLocalization(
```
hishamco, line 238: You can use `app.UseRequestLocalization("en-US")` for simplicity instead

aspnetcore/blazor/globalization-localization.md
```diff
@@ -97,63 +220,271 @@ By default, the Intermediate Language (IL) Linker configuration for Blazor WebAs
...
+                });
+
+-           await builder.Build().RunAsync();
++           builder.Services.AddLocalization();
+
++           var host = builder.Build();
+
++           CultureInfo culture;
++           var js = host.Services.GetRequiredService<IJSRuntime>();
++           var result = await js.InvokeAsync<string>("blazorCulture.get");
```
hishamco, line 359: This is what I mean to refer to in my previous comment about shows how the javascript function is called

hishamco (changes requested):
Great job, I need to have a test link if there's to check the entire changes LIVE

aspnetcore/blazor/globalization-localization.md
guardrex: I didn't test but because the PU focused the original text on `CurrentCulture` only I thought that I'd start with this. Then when Pranav looks, he may say that we need to say and show more about `CurrentUICulture`. I also haven't looked at reference source on this. Perhaps, I should go look.

I also didn't try *only* setting `CurrentUICulture` to see what happens. I figured that since the PU favored the original coverage for `CurrentCulture` that that's what they generally want devs to use.

aspnetcore/blazor/globalization-localization.md
guardrex: Perfect, thanks. I'll get this updated on the next commit.

aspnetcore/blazor/fundamentals/startup.md
guardrex: *Good point!* ..... I only tested Blazor WASM for this type of startup (`Blazor.start`). I should **_definitely_** test the scenario for Blazor Server and confirm that it works. If not, then **Yes, indeed!** ... I need to qualify this here. If this approach (`Blazor.start`) does work for Server, then I need to update over in the glob/loc article because I currently only have it there for WASM.

I'll get this checked/confirmed and updated on the next commit.

guardrex:
Thanks @hishamco ... that's **_COOL_** given my almost total lack of loc experience.

I'll get on your remarks in a few minutes. I'm about to open a PU issue for an inconsistency in globalization between WASM and Server that I don't understand. I'm pinging you on the issue ... you may immediately know why it's happening. Stand-by ... that issue will go up in just a minute or two.

aspnetcore/blazor/fundamentals/startup.md
hishamco: > If this approach (Blazor.start) does work for Server

No need for Server-Side, my comment about what you mean about **Statically set**? Which blazor app type?

aspnetcore/blazor/globalization-localization.md
hishamco: PU stand for?

aspnetcore/blazor/fundamentals/startup.md
guardrex: Both ... The app sets the culture and the user has no opportunity (in the app or via `Accept-Languages`) to change it.

btw -- Indeed, the `Blazor.start` scenario only works for WASM. The glob/loc topic is ok (it's in the WASM pivot). I'm making an adjustment now to the *Startup* topic to call out in the cross-link to this topic that `Blazor.start` for `applicationCulture` is WASM-only.

aspnetcore/blazor/globalization-localization.md
guardrex: "Product Unit" ... Pranav's original content for this topic focuses on `CurrentCulture`, and he never addressed (much, anyway) `CurrentUICulture`. It was just in a couple of his examples. He sets it, but he doesn't say much about it.

aspnetcore/blazor/fundamentals/startup.md
hishamco: > The app sets the culture and the user has no opportunity (in the app or via Accept-Languages) to change it

The culture is set via the app based on the culture provider (Cookie, QueryString, ... etc) it will be change during the app is running when the user change the query string for instance

I'm not sure if I understand you correctly ;)

guardrex: Testing seemed to indicate that the app only supports what loc middleware is configured to support. If I use ...

Blazor WebAssembly ...

```javascript
Blazor.start({
  applicationCulture: 'en-US'
});
```

... or ...

```csharp
CultureInfo.DefaultThreadCurrentCulture = new CultureInfo("en-US");
CultureInfo.DefaultThreadCurrentUICulture = new CultureInfo("en-US");
```

Blazor Server ...

```csharp
app.UseRequestLocalization("en-US");
```

... and then configure Opera with `Accept-Language` with Peruvian Spanish as the top language choice, I get **US English** content. I haven't been able to force any other culture/language. However, I don't understand how loc works at a low level, so can you give an example of the query string approach that would force a different culture/language with that :point_up: type of static culture configuration. I feel like I'm missing some major key point on how the framework works here. The basic question is: *How can the dev set an app up to support a specific culture/language that a user/visitor can't change?* ... that's what I mean here about "statically set the app's culture."

hishamco: > The basic question is: How can the dev set an app up to support a specific culture/language that a user/visitor can't change? ... that's what I mean here about "statically set the app's culture."

Using supported cultures & UI cultures, but this should work for server-side blazor if I'm not wrong

guardrex:
Ok ... I'm learning a little loc. 📚👩‍🏫🏫  

**UPDATE**: Peruvian Spanish **does** use the comma separator and dot decimal. However, there's still a problem here. Blazor Server is correct on the number format, but Blazor WASM is **not correct** (dot separator/comma decimal).

https://github.com/unicode-org/cldr/blob/829d6d3d911dbf032a118c3834737445292dc3e8/common/main/es_PE.xml

We'll hold the PR here a bit longer to see how things play out. We probably won't document a short-term bug that will be fixed, but we might call something out if there will be a long delay on it or if there's something else going wrong here ... something that I missed or failed to understand.

guardrex:
@hishamco ... I slept on the situation with the *inconsistency report* that I made. Although I guessed wrong about Peruvian Spanish 🙈 ... that all of Spanish used one number format ... at least I was correct about the inconsistency WASM vs. Server.

Yes, I think we should wait a sec to hear how the PU plays it out: If it's a quick bug fix for WASM to correct the globalization of certain Spanish cultures, then we can go ahead. They'll probably patch it back to 3.x, too. If it's a long-term item for them to address, then I think a NOTE is warranted in the topic with a cross-link to a potential PU issue. I'll :ear: for what they decide to do and then react here.

Otherwise, I think the topic is ok thus far. I'm sure that the readers will come after me 🔪😨 for anything I may have messed up.

guardrex:
@marin-bratanov ... This PR should help ... **_A LOT_** 😆 ... there's a lot more loc goodness going in on this PR. **_However_**, I think I have a delta from what you reported on #21619. I think u were saying that you need to set `BlazorWebAssemblyLoadAllGlobalizationData` to `true` for the framework (WASM apps) to react to an `Accept-Language` header, but that isn't what I found in testing for that most basic scenario (i.e., the app doesn't try to set the culture, offer a way to set it, or have any localization of resources). I might not have enough of an understanding on this, but when I just set up a Blazor WASM test app and make a request from Opera with Peruvian Spanish as the default (first) language, I get the right globalization without that project file setting. Here's what the test app does ...

Capture

Project file for this app:

```xml
<Project Sdk="Microsoft.NET.Sdk.BlazorWebAssembly">

  <PropertyGroup>
    <TargetFramework>net5.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly" Version="5.0.6" />
    <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly.DevServer" Version="5.0.6" PrivateAssets="all" />
    <PackageReference Include="System.Net.Http.Json" Version="5.0.0" />
  </ItemGroup>

</Project>
```

For setting `BlazorWebAssemblyLoadAllGlobalizationData` to `true`, that only seems to apply for the other scenarios (or a combination of them) ...

* Statically setting the culture in the app.
* The user is offered the ability to set the culture manually (`Microsoft.Extensions.Localization` package use, etc.).
* Localization is in use (i.e., the .NET Resources System is in use).

If you think this isn't correct, let's discuss further with @hishamco. I can repro anything that you're not seeing when the user has merely set the `Accept-Language` header and take it from there. I'm **_very much a NOOB_** on loc because I never had an employer that would pay 💲💰 for it. Their perhaps poor attitude was either 'this is for the US market' or 'everyone should read English' 🤦. Sad, so sad. Anyway ... I'm *learning loc here!!!* :smile: lol

guardrex:
@hishamco ... I'll go over this one more time and merge. I don't want to hold up the whole thing over the (apparent) WASM bug. Thanks again for your help.

marin-bratanov:
The cases I've tested have used the .net resource system, so it is likely that mine was a small subset of the bigger picture. I will try to take a peek at the live page in the next couple of weeks, but I am sure you've tested more extensively than I did.

guardrex:
@marin-bratanov ... Yes, setting `BlazorWebAssemblyLoadAllGlobalizationData` to `true` is now called out for *localization* (.NET Resources System) in the live topic. That was one of the updates, so I think we're good if that's all you were calling out in your issue. Here's the latest section ...

https://docs.microsoft.com/aspnet/core/blazor/globalization-localization?view=aspnetcore-5.0&pivots=webassembly#localization

Earlier sections of the topic pertain to using the `Accept-Language` header value automatically for only *globalization* (numbers, dates, etc.). If that's all the dev is trying to do, then they don't need that setting. The setting kicks in for everything else: an app that sets a static culture, an app that offers users to set the culture and it's maintained by the app, ... and I was saying ... for localization (.NET Resources System).

If you need to leave additional feedback on the updates, open a new issue with the **This page** feedback button and form at the bottom of the topic. Thanks again for your earlier issue.

marin-bratanov:
**The article is great** and explains much better when one needs the various things related to localization. I especially love the wasm/server toggle at the top. I was trying to say that your tests are much better and more inclusive than mine, and what I saw fits perfectly with what you said. I'm sorry for not being clear enough.

PR #22531, opened 2021-06-11
section header should be typeof instaed of nameof

fiyazbinhasan:
section name should be typeof

[Internal review](https://review.docs.microsoft.com/en-us/aspnet/core/mvc/views/view-components?view=aspnetcore-5.0&branch=pr-en-us-22531)


wadepickett:
@fiyazbinhasan, thanks very much for the fix!

PR #22532, opened 2021-06-11
Update jsonpatch.md

gbarrs-at-em:
It took me longer than I care to admit before I worked out I was getting a 400 response to a Patch request because I was serialising the request using  `JsonSerializer.Serialize` rather than `JsonConvert.SerializeObject`.  I hope this will help someone avoid the same issue.


Rick-Anderson (approved):

Rick-Anderson:
@gbarrs-at-em  thanks for the PR

PR #22533, opened 2021-06-12
Update hosted-with-azure-active-directory.md

julioalfredo:
The document, in the client configuration section reference the auto generated code in the program.cs. In this code it specify the template will produce a name based on the App Assembly concatenated with the ".WebAPI" literal. This has changed in recent versions and the reference should be App Assembly concatenated with the ".ServerAPI" literal.


guardrex:
Hello @julioalfredo ... Thanks for sending this in. Due to the requirement for versioning and the possible need to make this update in other WASM security docs, this one will require an issue and more work. Moved to https://github.com/dotnet/AspNetCore.Docs/issues/22534.

PR #22566, opened 2021-06-18
Small grammar fix

BvdLind:


guardrex:
Thanks @BvdLind ... but it's correct in it's current form because when read the "at" (`@`) is spoken.

PR #22594, opened 2021-06-22
Clarify Blazor chained bind example

### LINKED ISSUE #22593: PasswordEntry's Password parameter binding is confusing
The `PasswordEntry` component which is referenced twice on this page seems to be faulty. It's `Password` parameter is never used inside the component's logic, but it's private `password` member is referenced both in the markup and the event handler. Because of this, the parent component is notified of changes, but if the parent component sets the initial value of this parameter (like it does in the `PasswordBinding` example), nothing happens. So currently, it's essentially a one-way binding: from the child to it's parent, but not a two-way one, which it supposed to be.

---
#### Document Details

* ID: b211df59-e1c9-86c0-c904-31fa3c361e6d
* Version Independent ID: 3f912dfc-5ba5-6c6c-ba93-1ae78508a5cc
* Content: [ASP.NET Core Blazor data binding](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/data-binding?view=aspnetcore-5.0#feedback)
* Content Source: [aspnetcore/blazor/components/data-binding.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/data-binding.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #22593

cc: @peterbozso ... **_Baaaaack again!_** 🏃

* Let's see if the additions here to the text explain that this is one-way and that the purpose of the `password` in the parent is to demonstrate an initial value *there* and that it updates from the child.
* I call out the 💀 **_don't write to a component param_** 💀 bit with cross-links to that coverage.
* This links *down* a section for the two-way bit (that includes across a component hierarchy, too).
* The new text on this PR probably needs a little more work ... just a draft at this point. I'll work it a bit more in the morning ☕🍩. Let me know if you have additional feedback here on the PR.

### DIFF
--- a/aspnetcore/blazor/components/data-binding.md
+++ b/aspnetcore/blazor/components/data-binding.md
@@ -244,7 +244,7 @@ The `PasswordEntry` component is used in another component, such as the followin
 
 ::: moniker-end
 
-When the `PasswordBinding` component is initially rendered, the `password` value of `Not set` is displayed in the UI. After initial rendering, the value of `password` reflects changes of the `Password` component parameter value in the `PasswordEntry` component.
+When the `PasswordBinding` component is initially rendered, the `password` value of `Not set` is displayed in the UI. After initial rendering, the value of `password` reflects changes made to the `Password` component parameter value in the `PasswordEntry` component.
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


guardrex:
Let's give this a shot. I'll take more feedback :ear: from readers. For example, there might still be confusion on how to integrate a `BoundValue` property in the `NestedChild` example in this `PasswordEntry` example. If that were shown tho, it would probably be necessary to also explain out how `password` in the `PasswordEntry` component behaves. I also think that if the `BoundValue`'s `get` accessor assigns to `password` that the earlier behaviors that I described can be covered to keep `password` in `PasswordEntry` up-to-date with `Password` ... again, without directly writing to the `Password` parameter. That's not something that we currently show in the docs, but it hasn't been something that devs have asked for AFAIK.

peterbozso:
Thank you @guardrex for the fast and very informative feedback! I think the content of the article is much clearer now. :)

PR #22597, opened 2021-06-23
Blazor object disposal updates

### LINKED ISSUE #22592: Improve a few Blazor object disposal examples
A few object disposal instances can include a null check because `Dispose` can be called on a component during initialization (i.e., `OnInitialized`). There are perhaps a few other disposal nits to address, too.

guardrex:
Fixes #22592

Jeremy, I have you on here for a couple of EFCore-Blazor Server sample app updates (these are mostly null checks with `?` for calling `Dispose` on objects). I'll revert the PR's changes to your sample if incorrect. You don't need to look at everything on the PR ... just your EFCore sample app ... unless you want to :eye: everything. Search the diff with ...

```
BlazorServerDbContextExample
```

... to see the three files.

**UPDATE**: *Spoke with Steve.* We're probably good here so long as ur ok with the EFCore-Blazor Server sample updates.

**UPDATE**: We also just discussed setting disposed objects to `null`. The ordinary cases outside of the EFCore-Blazor sample don't require this pattern. I've left your `null` sets alone for this. I'm sure you had a good reason for doing it.

### DIFF
--- a/aspnetcore/blazor/common/samples/3.x/BlazorSample_Server/Pages/call-js-from-dotnet/CallJsExample3.razor
+++ b/aspnetcore/blazor/common/samples/3.x/BlazorSample_Server/Pages/call-js-from-dotnet/CallJsExample3.razor
@@ -32,5 +32,5 @@
         await jsClass.TickerChanged(stockSymbol, price);
     }
 
-    public void Dispose() => jsClass.Dispose();
+    public void Dispose() => jsClass?.Dispose();
 }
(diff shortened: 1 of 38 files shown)


JeremyLikness (approved):
LGTM. The only reason I set null is to avoid multiple disposals. Probably not needed but old habit, just dispose and null and then you can't dispose again.

guardrex:
Thanks @JeremyLikness ... Yes, that indeed falls into one of Steve's scenarios. My rubber 🦆 kind'a failed me on interpreting the docs, but he set me straight on it. I'll review this PR more time today/tomorrow and merge. Thanks for your help.

PR #22649, opened 2021-07-02
EF sqlite package include: Ensure last package installs

wadepickett:
[Internal Review: First Razor](https://review.docs.microsoft.com/en-us/aspnet/core/tutorials/razor-pages/model?view=aspnetcore-5.0&branch=pr-en-us-22649&tabs=visual-studio)
[Internal Review: First MVC](https://review.docs.microsoft.com/en-us/aspnet/core/tutorials/first-mvc-app/adding-model?view=aspnetcore-5.0&branch=pr-en-us-22649&tabs=visual-studio)

When CLI code in a doc includes multiple command lines, the last command does not execute.  In this case, the include several docs use for installing EF + SQL Server and SQLite has this issue.   

Adding " && \" after each CLI command to tie them together so they all run.

### DIFF
--- a/aspnetcore/includes/add-EF-NuGet-SQLite-CLI-5.md
+++ b/aspnetcore/includes/add-EF-NuGet-SQLite-CLI-5.md
@@ -4,11 +4,11 @@ no-loc: [Home, Privacy, Kestrel, appsettings.json, "ASP.NET Core Identity", cook
 Run the following .NET CLI commands:
 
 ```dotnetcli
-dotnet tool install --global dotnet-ef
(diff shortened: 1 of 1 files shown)


Rick-Anderson (approved):

wadepickett:
This was a draft since I had to test out if this suggestion worked to get around the issue of Powershell not running the last line when a list of commands are copy pasted from the doc.  It didn't.  If you try using " && /" at the end of each command to tie the commands together it still does not work for Powershell which was the problem scenario to address to begin with.  

Closing this draft.

PR #22664, opened 2021-07-06
Fix typo on example Hub method name

VasilisGaitanidis:
Fix typo on example Hub method name in order to fit with sample method name.

public Task SendMessageToGroup(string user, string message)
{
    return Clients.Group("SignalR Users").SendAsync("ReceiveMessage", user, message);
}


guardrex (approved):
Thanks @VasilisGaitanidis! :rocket:

PR #22680, opened 2021-07-08
BadRequest doesn't always work

gray-rabbit:
I think  previous code always return OK
so i fix that issue.

### DIFF
--- a/aspnetcore/mvc/models/file-uploads/samples/5.x/LargeFilesSample/Controllers/FileUploadController.cs
+++ b/aspnetcore/mvc/models/file-uploads/samples/5.x/LargeFilesSample/Controllers/FileUploadController.cs
@@ -72,15 +72,17 @@ namespace LargeFilesSample.Controllers
                     {
                         await section.Body.CopyToAsync(targetStream);
                     }
-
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks @gray-rabbit.

@GadHao ... do you agree with the suggested change?

GadHao:
@gray-rabbit @guardrex 
The comment of the previous code says: *This sample try to get the first file from request and save it, make changes according to your needs in actual use* 

So there will be two problems if you change it to your current code:
1. If you post multiple files to the controller, it will save them all instead of just the first file 
2. When post a multiple keys `form-data` to the controller, it may will return the `BadRequest` even if there are other file contents in the  `form-data`

There's no problem with your code, but it belongs to a specific scenario, not the original intention of this sample

guardrex:
Thanks @GadHao.

@gray-rabbit ... I'll close on that basis. @GadHao created this sample for us. To discuss the sample further, open a new issue using the **This page** feedback button and form at the bottom of the **_English US_** topic and `cc: @GadHao`. If you don't open it from the English-US topic, it will open on the localization repo, not our docs repo.

Capture

Issue #2501, opened 2017-01-12
[Localization] Missing steps for generating and editing resx on Linux and Mac


tanaka-takayoshi:
We need to generate and edit in localization. However [the document](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/localization.md#working-with-resource-files) describes the steps only on the Visual Studio on Windows. Now ASP.NET Core is cross platform so could you describe the supported steps on Linux and Mac OS.

At least, the document should describe the XML schema and the sample XML for resx.

spboyer:
Tooling for VS Code or VS for Mac? @sayedihashimi @madskristensen

madskristensen:
@spboyer I'm not aware of any tooling for those SKUs regarding .resx files. That's not to say that there aren't any

Rick-Anderson:
@ryanbrandenburg  @DamianEdwards  How do Visual Studio for Code folks generate/edit .resx files?

Rick-Anderson:
@tanaka-takayoshi  can you try [How to edit a .resx file on Mac OS ](https://medium.com/@jdassonval/how-to-edit-a-resx-file-on-mac-os-c4dac927145c#.iltk5k9nd)

tanaka-takayoshi:
@Rick-Anderson Thanks. I'm not a Mac user but Linux user. However, it appears working on Linux. I'll try later and report.

tanaka-takayoshi:
@Rick-Anderson I tried it, but there are some problems.

1. POEditor on Linux doesn't have a function to import  resx and add a localization term from scratch. It intends to import localization terms from souce code directly, but it doesn't work for ASP.NET Core codes.

2. http://converter.webtranslateit.com/ has other supported format. I tried CSV. But it can't convert the resource whose key contains space. So, for example, such term can't convert CSV to resx.
https://github.com/aspnet/Entropy/blob/rel/1.1.0/samples/Localization.StarterWeb/Resources/Views.Home.About.fr.resx#L123

Now, as far as https://msdn.microsoft.com/en-us/library/ekyft91f(v=vs.100).aspx is the formal schema for resource file, only supported way on Linux and Mac OSX is editing resx file directly as plain XML file. I'll try to create  VS Code extension for editing resx file or dotnet cli extension for converting to resx from some kind of files (like csv).

Rick-Anderson:
@tanaka-takayoshi  Thanks for the update.

spboyer:
[...]

PR #22691, opened 2021-07-08
Update localization.md

Rick-Anderson:
Mitigate #2501
@hishamco please review


hishamco:
Seems the guys asking about RESX not PO, so we could close this PR and address the some RESX extensions for VS Code

PR #19752, opened 2020-09-01
Update SignalR JavaScript Client to 3.1

serpent5:
Fixes #19606.

[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/signalr/javascript-client?view=aspnetcore-3.1&branch=pr-en-us-19752)

PR #22694, opened 2021-07-09
Use the automatic SignalR reconnect feature

eluchsinger:
> **Warning**: This code snippet is used as a [Sample](https://docs.microsoft.com/en-us/aspnet/core/signalr/javascript-client?view=aspnetcore-5.0#connect-to-a-hub).

I am not sure, if the behavior stays the same after this change. For me it looks like the intent should be the same though.

### DIFF
--- a/aspnetcore/signalr/javascript-client/samples/3.x/SignalRChat/wwwroot/chat.js
+++ b/aspnetcore/signalr/javascript-client/samples/3.x/SignalRChat/wwwroot/chat.js
@@ -2,6 +2,7 @@ document.addEventListener("DOMContentLoaded", () => {
     // <snippet_Connection>
     const connection = new signalR.HubConnectionBuilder()
         .withUrl("/chathub")
+        .withAutomaticReconnect()
         .configureLogging(signalR.LogLevel.Information)
         .build();
     // </snippet_Connection>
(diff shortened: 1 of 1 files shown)


guardrex:
As it stands, this will break the line highlight.

However, it doesn't look correct because there's already a section that deals with automatic reconnect ...

https://docs.microsoft.com/en-us/aspnet/core/signalr/javascript-client?view=aspnetcore-5.0#automatically-reconnect

eluchsinger:
@guardrex that's what looks wrong to me... I am assuming, the line I removed is for some sort of reconnect, but the docs already show how to deal with reconnects correctly.

guardrex:
Indeed, many doc authors have worked on this one, so this may have broken inadvertently with the updates over time. Brennan will let us know what's what when he sees this.

guardrex:
For reference, this bit was set up this way on purpose at (open the diff to see the line) ...

https://github.com/dotnet/AspNetCore.Docs/pull/19752/files#diff-aef409095229bff52581f3578b3cb21c82666335d353efd400eaa5aa1f93ee42R129

BrennanConroy:
Maybe we should keep `connection.onclose` but replace start with a log.

```js
connection.onclose(function (error) {
    if (error) {
        console.log(`Connection closed with error: ${error}`);
    } else {
        console.log("Connection closed.");
    }
});
```

guardrex:
... and that preserves the correct line highlighting in the topic, which focuses on the logging aspect. No further action is required AFAICT. I'll merge this when the build is complete. Thanks @eluchsinger and @BrennanConroy.

PR #22698, opened 2021-07-09
Update http-context.md

Rick-Anderson:
@serpent5  please review

### DIFF
--- a/aspnetcore/fundamentals/http-context.md
+++ b/aspnetcore/fundamentals/http-context.md
@@ -129,7 +129,7 @@ public class UserRepository : IUserRepository
 
 ## HttpContext access from a background thread
 
-`HttpContext` isn't thread-safe. Reading or writing properties of the `HttpContext` outside of processing a request can result in a <xref:System.NullReferenceException>.
+`HttpContext` isn't thread-safe. Reading or writing properties of the `HttpContext` outside of processing a request can result in a <xref:System.NullReferenceException>. In most cases, code should not access <xref:Microsoft.AspNetCore.Http.IHttpContextAccessor.HttpContext?displayProperty=nameWithType> to get data. Rather than accessing `IHttpContextAccessor`, applications should copy the data from the request on the controller or middleware and then pass it to a service.
 
 > [!NOTE]
 > If your app generates sporadic `NullReferenceException` errors, review parts of the code that start background processing or that continue processing after a request completes. Look for mistakes, such as defining a controller method as `async void`.
--- a/aspnetcore/fundamentals/http-context.md
+++ b/aspnetcore/fundamentals/http-context.md
@@ -129,7 +129,7 @@ public class UserRepository : IUserRepository
 
 ## HttpContext access from a background thread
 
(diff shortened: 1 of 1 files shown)


aspnetcore/fundamentals/http-context.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Access HttpContext in ASP.NET Core
 author: coderandhiker
-description: Learn how to access HttpContext in ASP.NET Core.
+description: HttpContext in ASP.NET Core. HttpContext is not thread-safe.
```
Rick-Anderson, line 4:
```suggestion
description: HttpContext in ASP.NET Core. HttpContext is not thread-safe and can throw NullReferenceException.
```

paulomorgado (approved):

aspnetcore/fundamentals/http-context.md
```diff
@@ -138,6 +138,7 @@ To safely perform background work with `HttpContext` data:
 
 * Copy the required data during request processing.
 * Pass the copied data to a background task.
+* Do ***not*** reference `HttpContext` data in parallel tasks. The data needed from the context should be extracted before starting the parallel tasks.
```
serpent5, line 141:
```suggestion
* Do ***not*** reference `HttpContext` data in parallel tasks. Extract the data needed from the context before starting the parallel tasks.
```

aspnetcore/fundamentals/http-context.md
```diff
@@ -129,7 +129,7 @@ public class UserRepository : IUserRepository
 
 ## HttpContext access from a background thread
 
-`HttpContext` isn't thread-safe. Reading or writing properties of the `HttpContext` outside of processing a request can result in a <xref:System.NullReferenceException>.
+`HttpContext` isn't thread-safe. Reading or writing properties of the `HttpContext` outside of processing a request can result in a <xref:System.NullReferenceException>. In most cases, code should not access <xref:Microsoft.AspNetCore.Http.IHttpContextAccessor.HttpContext?displayProperty=nameWithType> to get data. Rather than accessing `IHttpContextAccessor`, applications should copy the data from the request on the controller or middleware and then pass it to a service.
```
serpent5, line 132: I'm not sure if this is the right place for this change. The opening line from this topic suggests using `IHttpContextAccessor` with services, but then this argues against it. I think it's reasonable to reference `IHttpContextAccessor` from a service, but the complication is if said service does background work.

I wonder if this might discourage the reader from using `IHttpContextAccessor` inside services that don't perform background work, which could mean they'll end up copying data and trying to pass it through when it's not necessary to do so.

Rick-Anderson: Agreed.

Rick-Anderson:
```suggestion
`HttpContext` isn't thread-safe. Reading or writing properties of the `HttpContext` outside of processing a request can result in a <xref:System.NullReferenceException>.
```

Rick-Anderson:
@serpent5 please review

serpent5 (approved):
👍

PR #22701, opened 2021-07-09
Fixed unreachable code error in file upload example

dustout:
Max file size check is limited to 500kb without specifying long.MaxValue in file.OpenReadStream

### DIFF
--- a/aspnetcore/blazor/common/samples/5.x/BlazorSample_WebAssembly/Pages/file-uploads/FileUpload2.razor
+++ b/aspnetcore/blazor/common/samples/5.x/BlazorSample_WebAssembly/Pages/file-uploads/FileUpload2.razor
@@ -65,7 +65,7 @@
             if (uploadResults.SingleOrDefault(
                 f => f.FileName == file.Name) is null)
             {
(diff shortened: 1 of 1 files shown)


guardrex:
@dustout ... Thanks for spotting this. I see it now in the API remarks. I agree with the update, but we should go thru an issue for two reasons ... it's not really a "patch" scenario ...

* There are two sample apps with the code. We moved to sample app-based snippets, and there's a sample app for each hosting model.
* More importantly, I think we might want to refactor the whole approach into a try-catch scenario that eliminates the conditional check below on the file size given that `OpenReadStream` will throw if the file size is too large.

Also, I think we can go with the `maxFileSize` for the limit.

I'll open an issue and you can let me know if you want to work on it further. Otherwise, I'll knock it out next week.

dustout:
I totally agree that a try catch would be a better approach, and that maxFileSize for the limit would be better than long.MaxValue. Thank you for fixing that up, and your docs are the best anywhere, I'm a big fan!!!

guardrex:
Thanks @dustout ... I'll take of this first thing Saturday morning. I won't have a pleasant weekend knowing it's in this state. 🙈 lol

PR #22708, opened 2021-07-10
Refactor approach on max file size

### LINKED ISSUE #22703: Refactor approach on max file size
From @dustout on https://github.com/dotnet/AspNetCore.Docs/pull/22701 ...

> Max file size check is limited to 500kb without specifying long.MaxValue in file.OpenReadStream

Remarks ...

 I see it now in the API remarks. I agree with the update, but we should go thru an issue for two reasons ... it's not really a "patch" scenario ...

* There are two sample apps with the code. We moved to sample app-based snippets, and there's a sample app for each hosting model.
* More importantly, I think we might want to refactor the whole approach into a `try`-`catch` scenario that eliminates the conditional check below on the file size given that `OpenReadStream` will throw if the file size is too large.

Also, I think we can go with the `maxFileSize` for the limit.

If you want to work on this further, @dustout, let me know. :ear: ... Otherwise, I'll take care of this early next week. Thanks for reporting that oversight ... **_my bad!_**

**UPDATE**: And there's a third reason that this will need more work: The limitation has to be called out in the topic's text as well. We can't count on devs seeing the API doc ... *I didn't take it into account.* It didn't pop up in testing because I was only working with small test files.

---
#### Document Details

* Content: [ASP.NET Core Blazor file uploads](https://docs.microsoft.com/en-us/aspnet/core/blazor/file-uploads?view=aspnetcore-5.0)
* Content Source: [aspnetcore/blazor/file-uploads.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/file-uploads.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #22703

Thanks @dustout! :rocket: ... Good news is that the topic already had the text to cover it ...

> The `maxAllowedSize` parameter of `OpenReadStream` can be used to specify a larger size if required.

I just didn't implement **_what I wrote in the topic in the sample app!_** 🙈 Did I have too much 🍺???? ... **_Or, not enough!??_** 😆

Thanks again for the catch. I'll merge this to the live doc immediately.

### DIFF
--- a/aspnetcore/blazor/common/samples/5.x/BlazorSample_Server/Pages/file-uploads/FileUpload2.razor
+++ b/aspnetcore/blazor/common/samples/5.x/BlazorSample_Server/Pages/file-uploads/FileUpload2.razor
@@ -66,16 +66,13 @@
             if (uploadResults.SingleOrDefault(
                 f => f.FileName == file.Name) is null)
             {
-                var fileContent = new StreamContent(file.OpenReadStream());
-
-                files.Add(
-                    new()
-                    {
-                        Name = file.Name
-                    });
-
-                if (file.Size < maxFileSize)
+                try
                 {
+                    var fileContent = 
+                        new StreamContent(file.OpenReadStream(maxFileSize));
+
+                    files.Add(new() { Name = file.Name });
+
                     content.Add(
                         content: fileContent,
                         name: "\"files\"",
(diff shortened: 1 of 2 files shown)


dustout:
Looks awesome @guardrex !!!!

PR #22722, opened 2021-07-13

### LINKED ISSUE #22559: Confusing language describing the different lifespans
There seems to be an error in language regarding the lifespans, just above the section ["Call services from main"](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-5.0#call-services-from-main)...

The document states: "Scoped objects are the same for each request but different across each request."

I assume it's instead meant to be: ""Scoped objects are the same for each request but different across each **session**."? (or something similar to that nature)

---
#### Document Details

* Content: [Dependency injection in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-5.0#lifetime-and-registration-options)
* Content Source: [aspnetcore/fundamentals/dependency-injection.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/dependency-injection.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

steveberdy:
## Summary

Fixes ambiguity in description of the lifetime of scoped objects in dependency injection.

Fixes #22559

### DIFF
--- a/aspnetcore/fundamentals/dependency-injection.md
+++ b/aspnetcore/fundamentals/dependency-injection.md
@@ -220,7 +220,7 @@ Scoped services must be resolved in the `InvokeAsync` method:
 The logger output shows:
 
 * *Transient* objects are always different. The transient `OperationId` value is different in the `IndexModel` and in the middleware.
-* *Scoped* objects are the same for each request but different across each request.
+* *Scoped* objects are the same for each request in a session but different across different request sessions.
 * *Singleton* objects are the same for every request.
 
 To reduce the logging output, set "Logging:LogLevel:Microsoft:Error" in the *appsettings.Development.json* file:
--- a/aspnetcore/fundamentals/dependency-injection.md
+++ b/aspnetcore/fundamentals/dependency-injection.md
@@ -220,7 +220,7 @@ Scoped services must be resolved in the `InvokeAsync` method:
 The logger output shows:
 
 * *Transient* objects are always different. The transient `OperationId` value is different in the `IndexModel` and in the middleware.
-* *Scoped* objects are the same for each request but different across requests.
+* *Scoped* objects are the same for a given request but differ across each new request.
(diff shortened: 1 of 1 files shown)


guardrex:
Hi @steveberdy ... I don't recommend the use of the word "session" in this context. It's overloaded given the meaning and use of "[session state](https://docs.microsoft.com/aspnet/core/fundamentals/app-state)". I recommend that if any changes are made that they avoid the use of the word "session."

steveberdy:
I made that change, but I'm not sure if that makes it any less confusing for documentation readers

guardrex:
I don't think it's confusing in either form, but maybe I'm wrong about that. What's confusing about it? On a single request, the object is the same. Across different requests, there are different (new) objects. We've had this language up for *many years* with very few (none until now?) issues opened on it.

steveberdy:
There should be a way to say that scoped objects have the same content for each request, but are different objects for different requests. It's not really that confusing, but it has potential to be confusing. It is fine as is now, but could use more for an abundance of clarity.

steveberdy:
@Rick-Anderson could you provide us with any feedback on this edit before merging?

Rick-Anderson (approved):

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -220,7 +220,7 @@ Scoped services must be resolved in the `InvokeAsync` method:
 The logger output shows:
 
 * *Transient* objects are always different. The transient `OperationId` value is different in the `IndexModel` and in the middleware.
-* *Scoped* objects are the same for each request but different across each request.
+* *Scoped* objects are the same for each request but different across requests.
```
Rick-Anderson, line 223:
```suggestion
* *Scoped* objects are the same for a given request but differ across each new request.
```

PR #22725, opened 2021-07-14
Change "blazorserverside" to "blazorserver"

daveabrock:
There is no `blazorserverside` template. I changed it to the proper name, `blazorserver`.


guardrex (approved):
Ah ... yep ... this looks like a 😈 **_Sneaky Rotten Find-'n-Replace Grimlin!_** 😈 

**UPDATE**: I checked the whole doc set. This was the only little devil. Thanks for the fix!

PR #22726, opened 2021-07-14
fixed MacOS migration command

sisminnmaw:
fixed Mac OS migration command for Add a migration for rating session ( line 119, 120)


Rick-Anderson:
@sisminnmaw thanks for the fix, keep them coming.

Rick-Anderson:
The change won't show up until we merge to live, which should happen in a few days.

PR #22790, opened 2021-07-21
Merge to live - 7/21

guardrex:


Rick-Anderson (approved):

guardrex:
@Rick-Anderson ... That's closer to what PowerShell does (Sean W.) ... he references `toc.yml` for each group ... however ... it's still a bit confusing because he sources it from each group's folder, which doesn't seem to work here locally. We'll have to see if it solves the *pesky sidebar ToC issue* when it goes live. I knew that there would be a few 😈 **_gremlins_** 😈 in this. We're close tho!

PR #22807, opened 2021-07-22
Update DbContextFactory.cs

adamradocz:
I think checking null in the constructor is more elegant, and the code is a little bit simpler.

### DIFF
--- a/aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Data/DbContextFactory.cs
+++ b/aspnetcore/blazor/common/samples/3.x/BlazorServerEFCoreSample/BlazorServerDbContextExample/Data/DbContextFactory.cs
@@ -7,13 +7,16 @@ namespace BlazorServerDbContextExample.Data
     public class DbContextFactory<TContext> 
         : IDbContextFactory<TContext> where TContext : DbContext
     {
-        private readonly IServiceProvider _provider;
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks @adamradocz ... If Jeremy likes it, we'll just need to narrow your code down. OR ... you can do it. Make that so no line is longer than 85 characters to fit the width of our code block display.

guardrex:
Also, I think there's a 5.x sample for this, too, right? If so, can you add the change to that sample app, too?

Finally, we'll need to confirm that the topic isn't highlighting any lines for this bit of code. Our in-topic highlights go by line number; so if a code block changes here in a code file, it can break the highlighting in the topic and highlight the wrong line(s) or highlight no line(s).

I'll help with that last aspect tomorrow. I'm going OOF right now on a major 🧠 **_brain fry_** 🧠😵 for the day. I'll be back on in the morning.

guardrex:
Ok ... made the updates. This isn't part of the 5.x sample, and the topic's text doesn't display anything from this file.

@JeremyLikness ... If you're good with the update, we can proceed to merge.

guardrex (approved):

JeremyLikness (approved):
LGTM

guardrex:
Thanks @adamradocz! :rocket:

adamradocz:
@guardrex I just saw your comment, thanks for updating the PR.

PR #22811, opened 2021-07-22
Fix minor typo

rmourato:
As per title, just happened to be reading through this doc.
Keep up the good work!


guardrex (approved):
Thanks @rmourato! :rocket:

Issue #22831, opened 2021-07-23
Adopt version-by-file approach for Blazor docs


guardrex:

guardrex:
Actually, let's re-open this and work PRs from here. I'm moving my remarks over from the PR ..............

cc: @khoiph1 ... Khoi, I'm going to try this again now that Kirk got me straight on what a JSON array is! 🙈😆 **_DUH!_** I told them ... I just told Dan and Artak that I'm exhausted 😩 and must take a vacation 🏖️🍹 within the next couple of weeks to recharge.

Here's the live English-US doc ...

* https://docs.microsoft.com/en-us/aspnet/core/migration/no-loc-test?view=aspnetcore-6.0

... and then after MT runs, I'll check on ...

* https://docs.microsoft.com/es-es/aspnet/core/migration/no-loc-test?view=aspnetcore-6.0
* https://docs.microsoft.com/de-de/aspnet/core/migration/no-loc-test?view=aspnetcore-6.0
* https://docs.microsoft.com/fr-fr/aspnet/core/migration/no-loc-test?view=aspnetcore-6.0

... and see what happens. If it works this time, I'll change the `docfx.json` to reflect our need for a global **_ARRAY_** 😄 of `no-loc` entries.

Use of `fileMetadata` is fine for the test because we want to scope the test to just the `no-loc-test.md` file. For the full repo tho, we should use `globalMetadata` if this ends up working out. I propose to use the following ...

```json
"globalMetadata": {
  ... existing entries ...
  "no-loc": [ "entry1", "entry2," ... "entryX" ],
  ... existing entries ...
}
```

We also want to keep doc-specific `no-loc` words. When I make the full repo update, I plan to filter those `no-loc` entries and leave them **_with only the doc-specific entries_**. It does lead to a BIG ❓ tho ... **_Are they going to merge cleanly into the final doc?_** If nobody knows if that's going to happen, then I need to run a follow-up test: I'll add our existing `no-loc` metadata entry to my `no-loc-test.md` doc and then see if the test `fileMetadata` entry and the doc's metadata list **_combine_** and prevent localization of the list at the top of the doc **_AND_** prevent loc of my "computer" and "cat" words.

**UPDATE**: Actually, I think we should test `globalMetadata` separately, too. Let's make sure that every aspect of this looks 👀 good before proceeding. There will be two more tests to perform, and I think they can be performed simultaneously ...

[...]

PR #22832, opened 2021-07-23
Blazor doc versioning and revert test docs

guardrex:
Addresses #22831

* Removes versioning test docs
* Moves a few Blazor docs to the new approach

Rick, Wade ... I think even if you sign-off today, I'll wait until Monday morning to merge. I'm beat 😩 ... and this will need full testing after merging. It's also best to merge when fresh in case I need to quickly patch anything.

WRT using a `debug-topic` folder of static assets for the `debug.md` file instead of an expected `debug` folder: We have a `.gitignore` rule that prevents a "`debug`" folder from making it into the repo. Therefore, I tacked on the word `"topic"` to get it in.


Rick-Anderson (approved):

wadepickett (approved):
Looks good, thanks!

guardrex:
btw - I realized this morning that the `.gitignore` I keep at the Blazor root was allowing me to use the `debug` folder name before, so I've moved a copy of that `.gitignore` into each of the versioned `blazor` folders. Now, I can keep my `debug` folder of static assets for the *Debug* topic.

PR #22846, opened 2021-07-25
small text fixes to security-authentication-claims doc

damienbod:
Small text fixes to doc:

https://docs.microsoft.com/en-us/aspnet/core/security/authentication/claims

### DIFF
--- a/aspnetcore/security/authentication/claims.md
+++ b/aspnetcore/security/authentication/claims.md
@@ -23,8 +23,7 @@ This article covers the following areas:
 
 ## Mapping claims using OpenID Connect authentication
 
-The profile claims can be returned in the `id_token` which is returned after a successful authentication. The ASP.NET Core client application just needs to request the profile scope. When using the `id_token` for claims, no extra claims mapping is required.
-
+The profile claims can be returned in the `id_token`, which is returned after a successful authentication. The ASP.NET Core client app only requires the profile scope. When using the `id_token` for claims, no extra claims mapping is required.
 ```csharp
 public void ConfigureServices(IServiceCollection services)
 {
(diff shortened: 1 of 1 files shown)


aspnetcore/security/authentication/claims.md
```diff
@@ -104,7 +104,7 @@ public void ConfigureServices(IServiceCollection services)
 
 ## Claims namespaces, default namespaces
 
-ASP.NET Core adds default namespaces to some known claims which might not be required in your app. You can disable this namespace adding if you would like to use the claims exactly like the Open ID Connect server created. 
+ASP.NET Core adds default namespaces to some known claims which might not be required in your application. You can disable this namespace adding if you would like to use the claims exactly like the Open ID Connect server created.
```
guardrex, line 107: Routinely, we use "app" everywhere, so that bit is ok. However, the "you/your" bits and some of the other phrasing could use a touch, and there's a missing comma for the non-essential clause.

```suggestion
ASP.NET Core adds default namespaces to some known claims, which might not be required in the app. Optionally, disable these added namespaces and use the exact claims that the OpenID Connect server created.
```

... something along those lines.

aspnetcore/security/authentication/claims.md
```diff
@@ -21,9 +21,9 @@ This article covers the following areas:
 * Reset the claims namespaces
 * Customize, extend the claims using <xref:Microsoft.AspNetCore.Authentication.IClaimsTransformation.TransformAsync%2A>
 
-## Mapping claims using Open ID Connect authentication
+## Mapping claims using OpenID Connect authentication
 
-The profile claims can be returned in the `id_token` which is returned after a successful authentication. The ASP.NET Core client app just needs to request the profile scope. When using the `id_token` for claims, no extra claims mapping is required.
+The profile claims can be returned in the `id_token` which is returned after a successful authentication. The ASP.NET Core client application just needs to request the profile scope. When using the `id_token` for claims, no extra claims mapping is required.
```
guardrex, line 26: More nits here. Try ...... (with "app" ... that's ok) ...

```suggestion
The profile claims can be returned in the `id_token`, which is returned after a successful authentication. The ASP.NET Core client app only requires the profile scope. When using the `id_token` for claims, no extra claims mapping is required.
```

aspnetcore/security/authentication/claims.md
damienbod: thanks for the review, updated with you improvements

aspnetcore/security/authentication/claims.md
damienbod: thanks for the review, updated with you improvements

guardrex (approved):
Thanks @damienbod! :rocket:

Rick-Anderson (approved):

PR #22847, opened 2021-07-25
Add Hedging Delay

VahidFarahmandian:
for Hedging policy delay must be equal or greater than zero and without setting the HedgingDelay client will get InvalidOperationException immediately


JamesNK:
The bug is in the validation. I've fixed that in gRPC. The next release is coming soon so I think it is best the docs stay as they are.

PR #22869, opened 2021-07-27
Adding the missing implementation details that caused a casting exception

### LINKED ISSUE #30971: docker image default port changed
### Description

Since .NET 8 docker image default port changed from 80 to 8080, should we mention it in the docker section

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/migration/70-80?view=aspnetcore-8.0&tabs=visual-studio#convert-a-blazor-server-app-into-a-blazor-web-app

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/migration/70-80.md

### Document ID

534ab561-09c5-c33b-4394-0effba34d666

### Article author

Rick-Anderson

andrasdevs:
According to https://github.com/dotnet/aspnetcore/issues/30971, simple inheritance is not enough, it causes an exception. After some testing, it looks that the casting requires the interface to be implemented.

### DIFF
--- a/aspnetcore/blazor/security/index.md
+++ b/aspnetcore/blazor/security/index.md
@@ -116,14 +116,14 @@ For more information on dependency injection (DI) and services, see <xref:blazor
 
 ## Implement a custom AuthenticationStateProvider
 
-If the app requires a custom provider, implement <xref:Microsoft.AspNetCore.Components.Authorization.AuthenticationStateProvider> and override `GetAuthenticationStateAsync`:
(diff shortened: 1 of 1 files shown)


guardrex:
@andrasdevs ... We need to hear from engineering on this on the engineering issue before we adjust the coverage. It looks like they plan to investigate it further. Please add a comment to the engineering issue. I'm tracking on that.

hemant149:
Any progress on issue ?

PR #22882, opened 2021-07-29
Blazor VS new project guidance update

### LINKED ISSUE #22881: Minor issues (bad link, incorrect instructions?)
**Hosting model**: webassembly

1. The (how to download) link does not take you anywhere that explains how to partially clone a git repo, it takes you to an anchor tag which does not appear to exist.
2. There is no new project type called "Blazor App", its blazor webassembly or server
3. After creating the client (as there was no "make both" option in the start), the rest of the tutorial assumes another project named "BlazorWebAssemblySignalRApp.Server" has magically appeared.
4. A quick introduction in as to how the flow/logic will be would be nice, before jumping into the code, so you can see where you're going.

---
#### Document Details

* ID: 097456af-6293-8c24-3991-fc815b94dcbf
* Version Independent ID: 978264ac-18f8-e937-30db-7050e53ac7d1
* Content: [Use ASP.NET Core SignalR with Blazor](https://docs.microsoft.com/en-us/aspnet/core/tutorials/signalr-blazor?view=aspnetcore-5.0&tabs=visual-studio)
* Content Source: [aspnetcore/tutorials/signalr-blazor.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/tutorials/signalr-blazor.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #22881

Thanks @marius00! :rocket:

It's too bad that we need to jettison the explicit instructions (temporarily anyway until I speak with management about this), but we can't version the docs (easily) by VS release. We version docs by **_.NET release_**, and VS is on a different minor release schedule. They just so happened to make these changes in a minor release between .NET major releases.

These instructions could have had lead-in text like 'If you're using VS version 16.9 or earlier, blah blah blah ......... If you're using VS version 16.10 or later, blah blah blah'. However, that's messy. I think it's best for the time being to use a NOTE that says the selections are in different locations during the **Create a new project** process.

I'll take this situation up with management offline today to discuss further and see if they really want the guidance broken out in text by VS release. For now tho, this at least makes the docs accurate (at the expensive of not being as explicit). Thanks again for your report.


guardrex:
**UPDATE @marius00**: Management says that we'll document 16.10. Devs may write in that 16.9 doesn't match; but as I mentioned, we can't maintain docs here by VS release, only .NET Core release. I've opened https://github.com/dotnet/AspNetCore.Docs/issues/22885 to address it Friday morning.

PR #22919, opened 2021-07-31
Fix links

ashotmuradian:
Remove extra spaces in links


guardrex (approved):
Thanks @ashotmuradian! :rocket:

PR #22923, opened 2021-08-02
Fix link anchor

GitHubPang:
Fix the incorrect link anchor in order to point to the ShutdownTimeout section in https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host#shutdowntimeout


guardrex (approved):
Thanks @GitHubPang! :rocket:

PR #22963, opened 2021-08-05
Update Create.cshtml.cs

alarihwertbars:
This change is necessary in case await _context. SaveChangesAsync (); will throw an exception and you will need to return the page back with the previously entered data, so as not to start typing again. If you leave it as it is, then when you try to save the changes, it gives an error that "An unhandled exception occurred while processing the request.
ArgumentNullException: Value cannot be null. (Parameter 'source')"


Rick-Anderson:
@alarihwertbars  good  catch. Do you find that with client side validation disabled and entering invalid data?

alarihwertbars:
> @alarihwertbars хороший улов. Находите ли вы это с отключенной проверкой на стороне клиента и вводом недопустимых данных?

The error appeared due to my error in the code, the data was incorrectly transferred to the database.

PR #22964, opened 2021-08-06

EngincanV:
https://github.com/dotnet/aspnetcore/blob/b7528e5214158fe2d3095e7064e63adbe0c2edd4/src/Http/Http.Abstractions/src/HttpContext.cs#L55

`RequestServices` is not a collection. I understand by saying **collection** it refers to the container, but I think this might create confusion so I think it better be removed.


guardrex (approved):
Thanks @EngincanV! :rocket:

PR #22976, opened 2021-08-08
Update handle-errors.md

HamletHakobyan:
Source code and curl request with corresponding stack trace are not consistent.


HamletHakobyan:
wrong PR

PR #22997, opened 2021-08-11
Use Expression-bodied members

Ausm:


Rick-Anderson:
@serpent5  this looks like a better way.

serpent5:
#

> @serpent5 this looks like a better way.

Yep, agreed. I much prefer this approach.

Rick-Anderson:
@Ausm  thanks for the PR, keep them coming. Note I also updated the current version.

PR #23009, opened 2021-08-11
Broken links sweep

wadepickett:
Contributes to fixing #21751 

This PR fixes all 5.0 to 2.1 version broken links, except for a few where the link is still good but the issue was that the ssl cert expired.

PR #23024, opened 2021-08-13
Blazor link updates

guardrex:
Addresses #23009


wadepickett:
@guardrex, I was already doing a larger scope broken links sweep in #23009, I will continue with that but remove the 6 blazor topics you have here fixed in this PR.  They had moved locations while I had the PR open and so I was having to redo them anyway.  I will remove just these 6.  Thanks!

wadepickett (approved):

PR #23052, opened 2021-08-18
Update http3.md

JamesNK:
Link to preview features spec. There will be better documentation about this in the future, but link here for now.


guardrex (approved):

guardrex:
@JamesNK ... I only made nit updates, but it's best if you look and confirm the updates are 😎 ... especially for the last added sentence. I felt when I read the info that I wouldn't know how to stay abreast of issues with HTTP/3, so I added a sentence with a link to the Announcements repo.

PR #23068, opened 2021-08-19
Add app isolation details from customer feedback

blowdart:
Add isolation clarification driven by a customer misunderstanding who had used key vault to isolate individual keys in the same vault, causing app crashes.


aspnetcore/security/data-protection/configuration/overview.md
```diff
@@ -226,6 +226,8 @@ This isolation mechanism assumes that the apps are not malicious. A malicious ap
 
 If the Data Protection system isn't provided by an ASP.NET Core host (for example, if you instantiate it via the `DataProtectionProvider` concrete type) app isolation is disabled by default. When app isolation is disabled, all apps backed by the same keying material can share payloads as long as they provide the appropriate [purposes](xref:security/data-protection/consumer-apis/purpose-strings). To provide app isolation in this environment, call the [SetApplicationName](#setapplicationname) method on the configuration object and provide a unique name for each app.
 
+Data Protection is developed with the assumption that all applications sharing a key ring can access all items in that key ring, the application unique identifier is used to ioslate application specific keys derived from the key ring provided keys. It does not expect item level permissions, such as those provided by Azure KeyVault to be used to enforce extra isolation, and this will lead to application errors. If you do not want to rely on the built-in application isolation then seperate key store locations should be used and not shared between applications.
```
HaoK, line 229: fix some typos

```suggestion
Data Protection is developed with the assumption that all applications sharing a key ring can access all items in that key ring, the application unique identifier is used to isolate application specific keys derived from the key ring provided keys. It does not expect item level permissions, such as those provided by Azure KeyVault to be used to enforce extra isolation, and this will lead to application errors. If you do not want to rely on the built-in application isolation then separate key store locations should be used and not shared between applications.
```

HaoK (approved):

GrabYourPitchforks:
Was thinking about these comments this morning. The salient points we want to communicate are:

1. If two applications are pointed at the same key repository, this should be done with the intention that the two applications share the same master key material.

2. The application discriminator is used to allow two different applications to share the same master key material but to keep their cryptographic payloads isolated from one another. The docs already draw an analogy between this and multi-tenancy. If you want the apps to be able to read each other's cryptographic payloads, give them the same application discriminator.

3. If an application is compromised (RCE attack), all master key material accessible to that application must also be considered compromised, regardless of its protection-at-rest state. This implies that if two apps are pointed at the same repository, _even if they use different app discriminators_, a compromise of one is functionally equivalent to a compromise of both.

4. If apps need to remain truly _isolated_ from one another, they should use different key repositories. This naturally falls out of the definition of "isolated" (they're not isolated if they both have R+W access to each other's data stores), but maybe it needs to be said explicitly.

I think Barry's doc improvements hit (1) pretty well, but they don't clarify (2) - (4).

aspnetcore/security/data-protection/configuration/overview.md
Rick-Anderson: @Haok suggestion plus a few of my own

```suggestion
Data Protection is developed with the assumption that all applications sharing a key ring can access all items in that key ring. The application unique identifier is used to isolate application specific keys derived from the key ring provided keys. It doesn't expect item level permissions, such as those provided by Azure KeyVault to be used to enforce extra isolation. Attempting item level permissions generates application errors. If you don't want to rely on the built-in application isolation, separate key store locations should be used and not shared between applications.
```

blowdart:
@grabyourpitchforks Well free free to expand :)

Rick-Anderson:
> @GrabYourPitchforks Well free free to expand :)

@blowdart I added them to #23071

blowdart:
Hah, ok I'll let you close this then @Rick-Anderson and delete the branch if it's all in the other

Rick-Anderson:
> Hah, ok I'll let you close this then @Rick-Anderson and delete the branch if it's all in the other

No, just merge my branch into this one.

PR #23071, opened 2021-08-20
Apply Pitch fork guys suggestions

Rick-Anderson:
Applies @GrabYourPitchforks  non-suggestion suggestions to #23068

### DIFF
--- a/aspnetcore/security/data-protection/configuration/overview.md
+++ b/aspnetcore/security/data-protection/configuration/overview.md
@@ -234,7 +234,7 @@ Consider the following for app isolation:
 
 * The application discriminator is used to allow different apps to share the same master key material but to keep their cryptographic payloads distinct from one another. <!-- The docs already draw an analogy between this and multi-tenancy.--> For the apps to be able to read each other's cryptographic payloads, they must have the same application discriminator.
 
-* If an app is compromised (RCE attack), all master key material accessible to that app must also be considered compromised, regardless of its protection-at-rest state. This implies that if two apps are pointed at the same repository, even if they use different app discriminators, a compromise of one is functionally equivalent to a compromise of both.
+* If an app is compromised (for example, by an RCE attack), all master key material accessible to that app must also be considered compromised, regardless of its protection-at-rest state. This implies that if two apps are pointed at the same repository, even if they use different app discriminators, a compromise of one is functionally equivalent to a compromise of both.
 
   This "functionally equivalent to a compromise of both" clause holds even if the two apps use different mechanisms for key protection at rest. Typicallt, this isn't an expected configuration. The protection-at-rest mechanism is intended to provide protection in the event an adversary gains read access to the repository. An adversary who gains write access to the repository (perhaps because they attained code execution permission within an app) can insert malicious keys into storage. The Data Protection system intentionally does not provide protection against an adversary who gains write access to the key repository.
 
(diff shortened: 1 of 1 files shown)


aspnetcore/security/data-protection/configuration/overview.md
```diff
@@ -226,7 +226,19 @@ This isolation mechanism assumes that the apps are not malicious. A malicious ap
...
+
+* The application discriminator is used to allow different apps to share the same master key material but to keep their cryptographic payloads distinct from one another. <!-- The docs already draw an analogy between this and multi-tenancy.--> For the apps to be able to read each other's cryptographic payloads, they must have the same application discriminator.
+
+* If an app is compromised (RCE attack), all master key material accessible to that app must also be considered compromised, regardless of its protection-at-rest state. This implies that if two apps are pointed at the same repository, even if they use different app discriminators, a compromise of one is functionally equivalent to a compromise of both.
```
Rick-Anderson, line 237:
```suggestion
* If an app is compromised (for example, by an RCE attack), all master key material accessible to that app must also be considered compromised, regardless of its protection-at-rest state. This implies that if two apps are pointed at the same repository, even if they use different app discriminators, a compromise of one is functionally equivalent to a compromise of both.
```

blowdart (approved):

aspnetcore/security/data-protection/configuration/overview.md
```diff
@@ -226,8 +226,23 @@ This isolation mechanism assumes that the apps are not malicious. A malicious ap
...
+
+  This "functionally equivalent to a compromise of both" clause holds even if the two apps use different mechanisms for key protection at rest. Typicallt, this isn't an expected configuration. The protection-at-rest mechanism is intended to provide protection in the event an adversary gains read access to the repository. An adversary who gains write access to the repository (perhaps because they attained code execution permission within an app) can insert malicious keys into storage. The Data Protection system intentionally does not provide protection against an adversary who gains write access to the key repository.
+
+* If apps need to remain truly isolated from one another, they should use different key repositories. This naturally falls out of the definition of "isolated". Apps are ***not*** isolated if they all have Read and Write access to each other's data stores.
+=======
```
Rick-Anderson, line 242:
```suggestion
```

aspnetcore/security/data-protection/configuration/overview.md
```diff
@@ -226,8 +226,22 @@ This isolation mechanism assumes that the apps are not malicious. A malicious ap
...
+
+* If apps need to remain truly isolated from one another, they should use different key repositories. This naturally falls out of the definition of "isolated". Apps are ***not*** isolated if they all have Read and Write access to each other's data stores.
 Data Protection is developed with the assumption that all applications sharing a key ring can access all items in that key ring, the application unique identifier is used to isolate application specific keys derived from the key ring provided keys. It does not expect item level permissions, such as those provided by Azure KeyVault to be used to enforce extra isolation, and this will lead to application errors. If you do not want to rely on the built-in application isolation then separate key store locations should be used and not shared between applications.
```
Rick-Anderson, line 242:
```suggestion
Data Protection is developed with the assumption that all applications sharing a key ring can access all items in that key ring, the application unique identifier is used to isolate application specific keys derived from the key ring provided keys. It does not expect item level permissions, such as those provided by Azure KeyVault to be used to enforce extra isolation, and this will lead to application errors. If you do not want to rely on the built-in application isolation then separate key store locations should be used and not shared between applications.

```

Rick-Anderson, line 242:
```suggestion
```

Rick-Anderson:
@lady

Issue #23070, opened 2021-08-20
Indicate that `[Library Name]` = MSBuild $(PackageId) in the default convention


javiercn:
Relevant discussion here
https://github.com/dotnet/aspnetcore/issues/26055#issuecomment-704848519

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 9fd019c7-ceb7-6ef3-26ba-6b44341b6fc5
* Version Independent ID: fd116d00-0dc7-efaf-c714-34d77b345a05
* Content: [Reusable Razor UI in class libraries with ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/razor-pages/ui-class?view=aspnetcore-5.0&tabs=visual-studio)
* Content Source: [aspnetcore/razor-pages/ui-class.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/razor-pages/ui-class.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-razorpages**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

guardrex:
This is already covered ...

https://docs.microsoft.com/en-us/aspnet/core/razor-pages/ui-class?view=aspnetcore-5.0&tabs=visual-studio#consume-content-from-a-referenced-rcl-1

... but I see a number of opportunities to improve it. Many locations just remark on the "library name," so I'll make those explicit about the assembly name and indicate that the `PackageId` is used if specified. I'll include a cross-link for that to NuGet docs.

PR #23091, opened 2021-08-23
Package ID updates

guardrex:
Addresses #23070
Addresses https://github.com/dotnet/aspnetcore/issues/35587

Javier ...

> That's an error on the docs, it's always the package Id that is used.

I think you mean "always" **_for a package_** (and **_if specified_**) ... are those a correct interpretations of your remark?

The docs were too focused on a non-packaged library before I put this PR up and #23070 went in last week.

Let's go over the updates across these two PRs to make sure that we're on the same page. The current approach for docs is ...

* Leave the language: "`{LIBRARY NAME}` placeholder is the library's assembly name"
* Add the following language to it: "or \[NuGet `PackageId` (if specified)](/nuget/create-packages/creating-a-package-msbuild#set-properties)"

... because ...

* There isn't a package produced for a library merely added to a solution (i.e., `<PackageId>` has no influence on static asset paths in this case).
* The `<PackageId>` for packages is used for the library name (and paths) but defaults to the assembly name if not specified.

Is this all ok with you?

### DIFF
--- a/aspnetcore/blazor/components/css-isolation.md
+++ b/aspnetcore/blazor/components/css-isolation.md
@@ -47,7 +47,7 @@ h1 {
 
 ## CSS isolation bundling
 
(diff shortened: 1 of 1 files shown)


guardrex:
Nevermind @javiercn ... I get it now. It's the RCL that receives the `PackageId`.

Issue #23084, opened 2021-08-21
What is the official list of types that Blazor can automatically detect changes in Parameters to avoid re-rendering?


szalapski:
This doc states rerendering will be skipped when "All of the parameter values are of known immutable primitive types, such as int, string, DateTime, and haven't changed since the previous set of parameters were set."

What is the official list of types that Blazor can detect changes in?   I assume double, float, long, short?  All the types that have C# keywords for them?  But then I saw DateTime.  What about enums?  Records?  Any arbitrary struct? Any arbitrary value type?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: a8201d0b-a6c4-7656-a8d9-9ae1fee26d10
* Version Independent ID: 27742f9a-95e5-a2d0-aa54-0ef24cad3d2a
* Content: [ASP.NET Core Blazor component rendering](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/rendering?view=aspnetcore-5.0)
* Content Source: [aspnetcore/blazor/components/rendering.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/rendering.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
@szalapski ... I agree ... and not just for this single location. It should be clarified in the *Components* overview topic where component parameters are first explained. The language is a bit sloppy in perhaps two or three spots around the Blazor docs (`DateTime` struct but in a list of "primitive types"). If the text isn't going to specify types in those locations, it should just say "C# types." However, it should to call out (*somewhere*) types explicitly and/or types not supported (possibly including plans for future support).

*Yes* IIRC for everything on your list, but I'd have to check on `enums` ... I think so tho, and IIRC complex types always trigger rerendering ... the Blazor framework doesn't track if values within a complex type change.

I want to do a 🤠 **_Texas Two-step_** 🕺 on this one. I'll merge a quick PR on Monday just to clean up "primitive type" language in a few spots. Later after 6.0 releases, I'd like to consult with the PU on a permanent fix.

**Notes to self** for when I reach this issue ...

* I wonder if the list of unsupported types is so short that the text can say "C# types" with a &dagger; (dagger) short list that follows to show/explain any caveats.
[...]

PR #23097, opened 2021-08-23
Generalize skip render remark

guardrex:
Addresses #23084

Steve, can I generalize your remark? `DateTime` is a struct, and other unmanaged and reference types (e.g., enums, records) aren't mentioned. If not, do you want to stick with what you have or modify it here?

I'll circle around after 6.0 GA to see if we need further updates (here and elsewhere). This PR is merely a quick fix for this one bit of phrasing.

Thanks @szalapski :rocket: ... This will leave your issue open until I can get back to this subject. Thanks for commenting.

### DIFF
--- a/aspnetcore/blazor/components/rendering.md
+++ b/aspnetcore/blazor/components/rendering.md
@@ -26,7 +26,7 @@ By default, Razor components inherit from the <xref:Microsoft.AspNetCore.Compone
 
 Components inherited from <xref:Microsoft.AspNetCore.Components.ComponentBase> skip rerenders due to parameter updates if either of the following are true:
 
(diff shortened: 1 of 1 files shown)


guardrex:
This wasn't meant to fix the issue, just address one aspect of it.

Actually, I think Steve meant that `DateTime` is *primitive as opposed to complex*, a complex type being borrowed loosely from ADO/EF/database terminology: https://docs.microsoft.com/en-us/ef/ef6/modeling/designer/data-types/complex-types#what-is-a-complex-type. `DateTime`, enums, and records (by themselves) aren't complex.

The PU is busy working on the 6.0 release at the moment. I'll ask them about this as soon as I can on your issue, which should be right after RC2 releases.

PR #23158, opened 2021-08-30
Typo. Using Request in place of Response.

it3xl:
Logical typo. The word "request" is used where the word "response" is expected.


guardrex:
Hello @it3xl ... Thanks, but we define these features broadly from a 'processing requests' viewpoint. The whole topic is written from that perspective.

PR #23184, opened 2021-08-31
Add proper markdown line breaks

MarcusOtter:
These three sections should either have proper markdown line breaks or be converted to headings (or a bullet list?). Currently the "headings" run into the text content without a line break, which looks weird:

bild

Live page: https://docs.microsoft.com/en-US/aspnet/core/mvc/views/tag-helpers/intro?view=aspnetcore-5.0#what-tag-helpers-provide


Rick-Anderson:
Thanks @MarcusOtter  keep the PR's coming.

PR #23191, opened 2021-09-02
Update dotnet-client.md

Yawnder:
Added a clarification on the impact (or lack there of) on how using strong-typed hubs affects the client side.

### DIFF
--- a/aspnetcore/signalr/dotnet-client.md
+++ b/aspnetcore/signalr/dotnet-client.md
@@ -48,9 +48,9 @@ To establish a connection, create a `HubConnectionBuilder` and call `Build`. The
 
 ## Registering handlers on strongly-typed hubs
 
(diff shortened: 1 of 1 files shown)


BrennanConroy:
I don't think this comment having it's own section is particularly useful, if you really want this just add a note in [Call client methods from hub](https://docs.microsoft.com/aspnet/core/signalr/dotnet-client?view=aspnetcore-5.0&tabs=visual-studio#call-client-methods-from-hub)

Yawnder:
Indeed, it maybe doesn't need it's own section. The reason why I added one rather than a note is that in Server Concept/Hubs, there is a section for it, and there was no note anywhere regarding the client side of the same concept.

Feel free to close the PR if it's not helpful.

guardrex:
I already made it a note. @BrennanConroy will provide final signoff (or updates) before we merge this.

Yawnder:
Excellent. Thanks!

PR #23198, opened 2021-09-03
Use expression body

vanillajonathan:
This uses expression body for the properties.


Rick-Anderson:
Looks good except that version is no longer used. Can you update https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/data/ef-rp/intro/samples/cu50/PaginatedList.cs#L22-L36
cc @tdykstra

Rick-Anderson (changes requested):
https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/data/ef-rp/intro/samples/cu50/PaginatedList.cs#L22-L36

vanillajonathan:
Fixed in #23205

PR #23207, opened 2021-09-04
Blazor IdS custom domains and token signing

### LINKED ISSUE #23180: Add small bit that explains how to access certificates on a Linux App Service
On Linux App Services, you cant access the certificate from My\CurrentUser. A solution to this is to point IdentityServer directly to the file path certificates are saved to on the Linux App Service:

```json
{
    "IdentityServer": {
        "Key": {
            "Type": "File",
            "FilePath": "/var/ssl/private/{thumbprintGoesHere}.p12",
            "Password": ""
          }
    }
}
```

Can you add a disclaimer of sorts that highlights this solution? I've had so much trouble finding out that it was this simple to fix, and from google searches, I can see I am not alone in having a hard time making this work with Linux App Services.

---
#### Document Details

* ID: 9638772f-0f1e-c2a9-8a58-ae8c3e37ae28
* Version Independent ID: 7a5b169e-08cb-cdba-4501-88d3da2723bb
* Content: [Secure a hosted ASP.NET Core Blazor WebAssembly app with Identity Server](https://docs.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/hosted-with-identity-server?view=aspnetcore-5.0&tabs=visual-studio)
* Content Source: [aspnetcore/blazor/security/webassembly/hosted-with-identity-server.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/security/webassembly/hosted-with-identity-server.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #23180

@niem94 ... Here's the direction that I'm headed **_thus far_** 🚧⛏️👷. I might ultimately decide not to go this way, but this shows one way to compose coverage for these features.

* Split token signing from custom domains. These are independent concepts.
* Place the new App Service setup in the two new token signing sections (... but I might decide to take all of the new App Service coverage bits out because other docs exist for it in our doc set and in the Azure docs already. That way, all of this IdS coverage can just focus on the token signing and custom domain config.)
* We'll keep the IdS app token signing configuration for Windows as a gesture in the `appsettings.json` file, where it uses the Windows cert store.
* Let's consider **NOT** using the more complicated Azure Key Vault client to load a cert, as I suggested in our discussion on the issue. Given that Barry 🎯 is ok with just grabbing the cert at the path on the server where it's placed by App Service was **_approved_** :+1:, I think we can just do that. However, I don't want to do it the way that you proposed with the word "**_Password_**" appearing in the config file 😨. It's **_conceptually_** something that we're trying to stop devs from considering generally 🐉. Therefore, the approach on the PR is to load it via code with `AddSigningCredential` from the path.
* I *might* need to get a tiny bit of coverage in here for `AddDeveloperSigningCredential`. I'll consider that later. It might be as simple as a cross-link to the API doc to raise awareness on it.

This is merely a DRAFT at this point. I'm sure that it needs work that will require me to walk the steps. Because we're focused on 6.0 docs at the moment and other high priority issues, I'll return to this as I get time here and there to wrap it up. I might be able to do so in the coming days prior to 6.0 RC1 coming out ... that's when I'll get really busy with 6.0 bits again.

In the meantime, you can look this over and see what you think about the approach thus far. If you have time to test the proposed Linux code with `AddSigningCredential` **_without the app settings file config_**, I hope you'll be able to report back that it works. If you don't have time, no worries ... that's what I'll be testing when I get back to this PR.

... and don't worry about any little text problems ... e.g., a misspelled word 🙈, a broken phrase 🙈, a missing period 🙈 ... things like that ... this has all just been typed right in on a first pass. There are bound to be little text problems at this early stage, and I'll get those little devils 😈 out later.

### DIFF
--- a/aspnetcore/blazor/security/webassembly/hosted-with-identity-server.md
+++ b/aspnetcore/blazor/security/webassembly/hosted-with-identity-server.md
@@ -1064,6 +1064,7 @@ The guidance in this section explains:
 1. Create a new Azure Key Vault or use an existing key vault in your Azure subscription.
 
 1. Navigate to Azure App Service in the Azure portal and create a new App Service with the following configuration:
+
    * **Publish** set to `Code`.
    * **Runtime stack** set to the app's runtime.
    * For **Sku and size**, confirm that the App Service tier is `Basic B1` or higher if you also plan to use a custom domain. App Service requires a `Basic B1` or higher service tier to use custom domains.
@@ -1083,6 +1084,7 @@ The guidance in this section explains:
    No configuration changes to the default settings are required for the key vault service.
 
    For testing purposes, an app's local [SQLite](https://www.sqlite.org/index.html) database, which is configured by default by the Blazor template, can be deployed with the app without additional configuration. Configuring a different database for Identity Server in production is beyond the scope of this article. For more information, see the database resources in the following documentation sets:
+
    * [App Service](/azure/app-service/)
    * [Identity Server](https://docs.duendesoftware.com/identityserver/v5/overview/)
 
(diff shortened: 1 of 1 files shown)


guardrex:
@niem94 ... I'll take feedback on the new content that relies on existing documentation for Duende IdS automatic token signing with DP keys in Azure Key Vault and the new content for loading the cert manually on Linux. This topic is scheduled for a full overhaul in 2022, and I'll take another look at all of this in more detail when I reach it. Thanks again for your help.

niem94:
@guardrex Yes, I will add a comment here when I get the time to test it out :-) Happy to have been of help! :-)

PR #23211, opened 2021-09-04
V6 of DI /5

Rick-Anderson:

### DIFF
--- a/aspnetcore/fundamentals/dependency-injection.md
+++ b/aspnetcore/fundamentals/dependency-injection.md
@@ -404,217 +404,155 @@ The following table lists a small sample of these framework-registered services:
 
 ::: moniker-end
 
-::: moniker range="< aspnetcore-3.0"
+::: moniker range="< aspnetcore-6.0"
 
-By [Steve Smith](https://ardalis.com/), [Scott Addie](https://scottaddie.com), and [Brandon Dahler](https://github.com/brandondahler)
+By [Kirk Larkin](https://twitter.com/serpent5), [Steve Smith](https://ardalis.com/), [Scott Addie](https://scottaddie.com), and [Brandon Dahler](https://github.com/brandondahler)
 
 ASP.NET Core supports the dependency injection (DI) software design pattern, which is a technique for achieving [Inversion of Control (IoC)](/dotnet/standard/modern-web-apps-azure-architecture/architectural-principles#dependency-inversion) between classes and their dependencies.
 
 For more information specific to dependency injection within MVC controllers, see <xref:mvc/controllers/dependency-injection>.
 
+For information on using dependency injection in applications other than web apps, see [Dependency injection in .NET](/dotnet/core/extensions/dependency-injection).
+
+For more information on dependency injection of options, see <xref:fundamentals/configuration/options>.
+
+This topic provides information on dependency injection in ASP.NET Core. The primary documentation on using dependency injection is contained in [Dependency injection in .NET](/dotnet/core/extensions/dependency-injection).
+
 [View or download sample code](https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/fundamentals/dependency-injection/samples) ([how to download](xref:index#how-to-download-a-sample))
 
 ## Overview of dependency injection
 
-A *dependency* is any object that another object requires. Examine the following `MyDependency` class with a `WriteMessage` method that other classes in an app depend upon:
+A *dependency* is an object that another object depends on. Examine the following `MyDependency` class with a `WriteMessage` method that other classes depend on:
 
 ```csharp
 public class MyDependency
 {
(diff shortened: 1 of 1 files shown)


aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -109,40 +91,27 @@ In dependency injection terminology, a service:
...
-
-Only the following services can be injected into the `Startup` constructor when using the Generic Host (<xref:Microsoft.Extensions.Hosting.IHostBuilder>):
+## Services injected into Program.c
 
-* <xref:Microsoft.AspNetCore.Hosting.IWebHostEnvironment>
-* <xref:Microsoft.Extensions.Hosting.IHostEnvironment>
-* <xref:Microsoft.Extensions.Configuration.IConfiguration>
-
-Any service registered with the DI container can be injected into the `Startup.Configure` method:
+Services can be injected into the *Program.c* file.
```
Rick-Anderson, line 100: @halter73 What's the correct wording here now that top level statements are in effect?
https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/program-structure/top-level-statements

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -336,8 +295,9 @@ See [Recommendations](/dotnet/core/extensions/dependency-injection-guidelines#re
 * Another service locator variation to avoid is injecting a factory that resolves dependencies at runtime. Both of these practices mix [Inversion of Control](/dotnet/standard/modern-web-apps-azure-architecture/architectural-principles#dependency-inversion) strategies.
 * Avoid static access to `HttpContext` (for example, [IHttpContextAccessor.HttpContext](xref:Microsoft.AspNetCore.Http.IHttpContextAccessor.HttpContext)).
 
+<!--
 <a name="ASP0000"></a>
-* Avoid calls to <xref:Microsoft.Extensions.DependencyInjection.ServiceCollectionContainerBuilderExtensions.BuildServiceProvider%2A> in `ConfigureServices`. Calling `BuildServiceProvider` typically happens when the developer wants to resolve a service in `ConfigureServices`. For example, consider the case where the `LoginPath` is loaded from configuration. Avoid the following approach:
+* Avoid calls to <xref:Microsoft.Extensions.DependencyInjection.ServiceCollectionContainerBuilderExtensions.BuildServiceProvider%2A> in `ConfigureServices`.  For example, consider the case where the `LoginPath` is loaded from configuration. Avoid the following approach:
```
Rick-Anderson, line 298: @halter73  I removed this entire section.

aspnetcore/fundamentals/dependency-injection.md
```diff
@@ -227,16 +194,6 @@ To reduce the logging output, set "Logging:LogLevel:Microsoft:Error" in the *app
 
 [!code-json[](dependency-injection/samples/3.x/DependencyInjectionSample/appsettings.Development.json?highlight=7)]
 
-## Call services from main
```
Rick-Anderson, line 230: @halter73 nice to get rid of this section. I like the new minimal hosting model.

aspnetcore/fundamentals/dependency-injection.md
Rick-Anderson: @halter73 this was an easy migration to minimal host model and everything seems to work. Is there anything with DI that would change other than what I pointed out?

aspnetcore/fundamentals/dependency-injection.md
halter73: Is it just "Calling `BuildServiceProvider` typically happens when the developer wants to resolve a service in `ConfigureServices`." that was removed? I think it's fine removing it.

We could mention that after the `app` is built, you should use `app.Services` instead of `builder.Services.BuildServiceProvider()`, but maybe we shouldn't give people any ideas 😆

aspnetcore/fundamentals/dependency-injection/samples/6.x/WebApp1/Program.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    var services = serviceScope.ServiceProvider;
+
+    var myDependency = services.GetRequiredService<IMyDependency>();
+    myDependency.WriteMessage("Call services from main");
+}
+
+app.MapGet("/", () => "Hello World!");
+
+app.Run();
```
Rick-Anderson, line 17: > Glad you like the new hosting model! However, I think we might want to keep this section with app.Services instead of host.Services to show how to resolve a scoped services for a limited duration.

@halter73  how  is this?

aspnetcore/fundamentals/dependency-injection.md
Rick-Anderson: I added this back and marked the code that demonstrates this.

PR #23263, opened 2021-09-10
Update server.md

heartcut:
Code for "Linux with Nginx" was missing a semi colon after the server_name. Without it nginx gives the error "directive "server_name" is not terminated by ";".


guardrex (approved):
Thanks @heartcut! :rocket:

PR #23307, opened 2021-09-15
Update HomeController.cs

emixxx:
Removal of duplicate semicolon.


guardrex (approved):
Thanks @emixxx! :rocket:

guardrex:
@emixxx ... This bug 🐞 also appears in the other version of the sample app at ...

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/performance/caching/memory/sample/WebCache/Controllers/HomeController.cs#L204

Would you mind adding a commit to fix it there, too? If not, no worries ... I'll do it on a follow-up PR.

PR #23317, opened 2021-09-15
Fixed '@' symbol work definition

pramuka9:
`@` symbol is followed by a Razor reserved keyword, it transitions into Razor-specific markup, otherwise it transitions into HTML. --NOT '...into C#."


Rick-Anderson:
Nice catch!

PR #23330, opened 2021-09-16
Fix broken anchor tag for monickers < 3.0

qJake:


guardrex:
Hello @qJake ... This isn't going be a fix that we can institute everywhere on a case-by-case basis. We'll have a temporary improvement when 6.0 becomes the default doc set, then there will be a permanent fix for the whole doc set in 2022.

PR #23351, opened 2021-09-19
Fix Links to the "Service Lifetime" and "Detect Transient Disposables…

erhanalankus:
Fix Links to the "Service Lifetime" and "Detect Transient Disposables" Sections

The sections have an additional "-1" at the end of their id's, so the internal links which don't have the "-1" didn't work.
This PR adds "-1" to the links, but the better solution may be to remove the "-1" at the end of their id's"
See:
```
<h2 id="detect-transient-disposables-1" class="heading-anchor">
<h3 id="service-lifetime-1" class="heading-anchor">

```


guardrex:
Hello @erhanalankus ... This is a known issue. The current approach of keeping versioned content in the same markdown file is breaking bookmarks because the build system is automatically numbering `id`s of duplicate HTML section headings. This ends up breaking bookmarks for the non-default version of content in the files. The problem will be partially resolved when the 6.0 preview docs become the default loaded doc set at 6.0 GA (November). The problem will be permanently fixed in 2022 when a new content versioning approach called "Version by File" is implemented, where each version of coverage (3.1, 5.0, 6.0, etc.) will be placed into separate files. Section headings won't be duplicated at that point, and all bookmarks for all versions of the content will function normally. We're sorry for the current inconvenience. Chalk it up to *growing pains* for rapidly evolving content.

PR #23354, opened 2021-09-20
Improve [Required] attribute description

### LINKED ISSUE #23034: [Required] on non-nullable reference types has some weird differences from the actual Required attribute.
This is an opinion.

I feel like once you know you will get a `[Required]` attribute on your non-nullable reference types, unless disabled by `services.AddControllers(options => options.SuppressImplicitRequiredAttributeForNonNullableReferenceTypes = true);`, you expect as a developer to get the full package of what `[Required]` has to offer.

Currently, `[Required]` not only checks if a field is null, but it also checks if the type is string - to not be an empty string, or string filled with only white spaces. But, it seems `[Required]` is only applied in the following situations for non-nullable reference types _(Accentuations are mine)_:

> If the app was built with <Nullable>enable</Nullable>, a **missing value** for Name in a JSON or form post results in a validation error. Use a nullable reference type to allow null or missing values to be specified for the Name property: ([source](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-5.0#non-nullable-reference-types-and-required-attribute))

But then when you explicitly define a `null` value to a non-nullable field in a JSON payload, you still get an error, unless "missing value" here refers to strictly missing "any value other than `null`", and not what `[Required]` is supposed to do.

These ambiguities are really confusing:

- Is `[Required]` applied to any non-nullable reference type? If yes: Why is the behavior entirely different from the actual `[Required]` attribute? For example, If you define an empty string to a non-nullable string, no error occurs, and yet if you really had the `[Required]` attribute on a string, you would get an error.

- Is `[Required]` applied only to fields not explicitly given a value? If yes: Why then do we get an error when we explicitly define a `null` value to them?

I hope this made sense.

---
#### Document Details

* ID: 0156ffc3-47ac-d5da-fbc2-9ff5018b071b
* Version Independent ID: 615db388-a244-41f7-21b5-a4085d9ed479
* Content: [Model validation in ASP.NET Core MVC](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-5.0#required-attribute)
* Content Source: [aspnetcore/mvc/models/validation.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/mvc/models/validation.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

SpiritBob:
Fixes #23034 by making it clear to the end-user that the `[Required]` attribute used for non-nullable context is not the default one, but one that only checks for null values.

### DIFF
--- a/aspnetcore/mvc/models/validation.md
+++ b/aspnetcore/mvc/models/validation.md
@@ -466,7 +466,7 @@ To find out which parameters are passed to `String.Format` for a particular attr
 
 ## [Required] attribute
 
-By default, the validation system treats non-nullable parameters or properties as if they had a `[Required]` attribute. [Value types](/dotnet/csharp/language-reference/keywords/value-types) such as `decimal` and `int` are non-nullable.
+By default, the validation system treats non-nullable parameters or properties as if they had a `[Required(AllowEmptyStrings = true)]` attribute. [Value types](/dotnet/csharp/language-reference/keywords/value-types) such as `decimal` and `int` are non-nullable.
 
 ### [Required] validation on the server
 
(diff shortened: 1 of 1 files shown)


guardrex:
Hello @SpiritBob ... You changed the <3.0 section of the doc, so that covers the 2.x-era of docs. As you probably know, 2.x isn't supported. This doc will be updated to remove all of the content for that version, including the change that you're making here. If you don't find a change in the >=3.0 section of the doc, then there's no change to apply to the topic (other than killing the 2.x content).

Rick-Anderson:
@SpiritBob  can you create a PR for the current version? You'll get credit for this PR.
Be sure to check out your name as a contributor about the 6th of next month on the [What's new page](https://docs.microsoft.com/en-us/aspnet/core/whats-new/)

SpiritBob:
> Hello @SpiritBob ... You changed the <3.0 section of the doc, so that covers the 2.x-era of docs. As you probably know, 2.x isn't supported. This doc will be updated to remove all of the content for that version, including the change that you're making here. If you don't find a change in the >=3.0 section of the doc (or add a section to the doc within that moniker range), then there's no change to apply to the topic (other than killing the 2.x content).
> 
> Oh ... one other note ... please name your PRs with a real (short) description of the change.

I'll get to it now, thank you!

SpiritBob:
> @SpiritBob can you create a PR for the current version? You'll get credit for this PR.
> Be sure to check out your name as a contributor about the 6th of next month on the [What's new page](https://docs.microsoft.com/en-us/aspnet/core/whats-new/)

Sorry about that, I've created https://github.com/dotnet/AspNetCore.Docs/pull/23355

Hope all is good this time around! 😅

PR #23355, opened 2021-09-20
Improve [Required] attribute usage

SpiritBob:
Clarifies to the end-user that for non-nullable reference types a `[Required(AllowEmptyStrings = true)]` attribute is used, that only checks for null values.


Rick-Anderson (approved):

Rick-Anderson:
Be sure to check out your name as a contributor about the 6th of next month on the [What's new page](https://docs.microsoft.com/en-us/aspnet/core/whats-new/)

PR #23359, opened 2021-09-20
Cross-link certificate use patterns for App Service

Ted-chiptech:
Fixes #63530

There has some extra setup needed for Azure Web Apps, the validation will fail with the "The system cannot find the file specified" error if they're not supplied.

### DIFF
--- a/aspnetcore/security/authentication/certauth.md
+++ b/aspnetcore/security/authentication/certauth.md
@@ -262,13 +262,12 @@ See the [host and deploy documentation](xref:host-and-deploy/proxy-load-balancer
 
 ### Use certificate authentication in Azure Web Apps
 
-No forwarding configuration is required for Azure. This is already setup in the certificate forwarding middleware.
+No forwarding configuration is required for Azure. Forwarding configuration is set up by the Certificate Forwarding Middleware.
 
 > [!NOTE]
-> This requires that the CertificateForwardingMiddleware is present.
+> Certificate Forwarding Middleware is required for this scenario.
 
-Add "WEBSITE_LOAD_USER_PROFILE=1" to the application setting when hosting on a Windows instance, and the App Service plan needs to be on a Basic or above.
-
+For more information, see [Use a TLS/SSL certificate in your code in Azure App Service (Azure documentation)](/azure/app-service/configure-ssl-certificate-in-code).
 
 ### Use certificate authentication in custom web proxies
 
--- a/aspnetcore/security/authentication/certauth.md
+++ b/aspnetcore/security/authentication/certauth.md
@@ -267,6 +267,9 @@ No forwarding configuration is required for Azure. This is already setup in the 
 > [!NOTE]
 > This requires that the CertificateForwardingMiddleware is present.
 
+Add "WEBSITE_LOAD_USER_PROFILE=1" to the application setting when hosting on a Windows instance, and the App Service plan needs to be on a Basic or above.
+
+
 ### Use certificate authentication in custom web proxies
 
 The `AddCertificateForwarding` method is used to specify:


guardrex:
Hello @Ted-chiptech ... Thanks for sending this in. That issue number for the "Fixes" isn't correct.

@Rick-Anderson ... Indeed, this is fairly close to what we needed for certs and custom domains for [Blazor hosted WASM on Azure Apps with IdS and a custom domain/key signing cert (scroll down to steps 5 and 6)](https://docs.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/hosted-with-identity-server?view=aspnetcore-5.0&tabs=visual-studio#use-a-custom-domain-and-tls-certificate-for-https-protocol-communication-in-azure-app-service), but wouldn't the key setting for app settings be `WEBSITE_LOAD_CERTIFICATES` with the cert thumbprint as the value (separated by commas if there are more than one)?

Ted-chiptech:
@guardrex Thanks for look at this.
Sorry, I just realize that the issue was raised in a different repository(https://github.com/MicrosoftDocs/azure-docs/issues/63530)
Should I remove the Fixes instead?

And I guess my case is a little bit different maybe? I will need to read a certificate from Azure KeyVault, then do a customized validation to check if the inbound certificate's thumbprint matches.

So I have a function hook to OnCertificateValidated event to do the check, and the function contains some code to load the certificate from KeyVault to check if they match, and fail the context if they don't. 

Should I use ```WEBSITE_LOAD_CERTIFICATES ``` to solve this issue and remove my customized validation function? Does it validate the thumbprint and fail the inbound certificate if it doesn't match? Thanks!

Ted-chiptech:
@gabrielaalfaro I tested the ```WEBSITE_LOAD_CERTIFICATES```, looks like it does exactly what I need, It will fail the inbound request if the certificate's thumbprint doesn't match. 👍 

Thanks a lot for pointing it out, I think it could be worth mentioning this in the document too, just in case other people trying to deal with a similar issue.

Ted-chiptech:
Sorry, I was wrong, I test a couple more times, looks like it doesn't validate the inbound certificate thumbprint and it doesn't fail the request if the thumbprint doesn't match the value in the ```WEBSITE_LOAD_CERTIFICATES```, so maybe I did something wrong?

guardrex:
Cool beans. That's what I thought would happen. I didn't test it, but I think it makes sense that it won't load if the thumbprint doesn't match a certificate there.

I'll leave it up to @Rick-Anderson if he wants to edit this a bit to call that key+value out. In the Windows scenario, the cert can be loaded with that approach from the *CurrentUser* > *My* store. In the Linux scenario, the cert can be loaded manually from `/var/ssl/private/{THUMBPRINT}.p12`. How much of these scenarios is covered over here is somewhat debatable, but there's overlap given that ASP.NET Core apps integrate so well with Azure App Service.

Ted-chiptech:
> **UPDATE**: I see. Can't you validate after it loads?
> 
> I'll leave it up to @Rick-Anderson if he wants to edit this a bit to call that key+value out. In the Windows scenario, the cert can be loaded with that approach from the _CurrentUser_ > _My_ store. In the Linux scenario, the cert can be loaded manually from `/var/ssl/private/{THUMBPRINT}.p12`. How much of these scenarios is covered over here is somewhat debatable, but there's overlap given that ASP.NET Core apps integrate so well with Azure App Service.

Yes, I was validate the thumbprint after it loads by hook the method to the OnCertificateValidated event, however, if I don't set the ```WEBSITE_LOAD_USER_PROFILE``` flag, it will throw out "The system cannot find the file specified" error and won't load the certificate. or I could do something wrong, but I think there have people experiencing similar issues too by check the stackoverflow, and the suggestion is to set the ```WEBSITE_LOAD_USER_PROFILE``` flag on windows instance to load the certificate.

guardrex:
Oh, I see ... you aren't using Azure Key Vault to store the certificate. Ok, that makes sense.

Part of the problem here is that there are several scenarios and a few different doc sets.

It's covered here ...

https://docs.microsoft.com/en-us/azure/app-service/configure-ssl-certificate-in-code#load-certificate-from-file

... so perhaps we can just cross-link that whole doc to this section? Perhaps, replace the line that you added with this ...

> For more information, see \[Use a TLS/SSL certificate in your code in Azure App Service](/azure/app-service/configure-ssl-certificate-in-code#load-certificate-from-file).

guardrex:
That Azure doc :point_up: also covers the required service plan level.

guardrex:
Ok ... I went ahead with that update and a few other nit updates. Rick will be on in a bit (or tomorrow). He may revise further and/or approve+merge.

Thanks for commenting on this, @Ted-chiptech! :rocket:

Ted-chiptech:
> Ok ... I went ahead with that update and a few other nit updates. Rick will be on in a bit (or tomorrow). He may revise further and/or approve+merge.
> 
> Thanks for commenting on this, @Ted-chiptech! 🚀

@guardrex Yes, I agree. Thanks a lot for your help! 👍

PR #23379, opened 2021-09-24
Feature/https linux updates

### LINKED ISSUE #23378: HTTPS certificate on Linux needs to trust the user, not root, certificate
In **Trust HTTPS certificate on Linux**, when exporting the certificate, you need to make sure you export the developer certificate for the current, not root, i.e. check the thumbprint you are trusting is the one you want.

To export the current user's developer certificate I had to make sure it was created (`dotnet dev-certs https`), then run `sudo -E dotnet dev-certs https -ep /usr/local/share/ca-certificates/aspnet/https.crt --format PEM`, to use the current user's environment. You can check the current users certificate via `ls ~/.dotnet/corefx/cryptography/x509stores/my`; the filename will be the SHA1 thumbprint, and the certificate you are trusting (`https.crt`) has to match.

To check the thumbprint of the exported certificate: `openssl x509 -noout -fingerprint -sha1 -inform pem -in /usr/local/share/ca-certificates/aspnet/https.crt`

Running `sudo dotnet dev-certs https -ep /usr/local/share/ca-certificates/aspnet/https.crt --format PEM`, without the `-E`, would export root's certificate (generating it if necessary), which had a different thumbprint. You can check this certificate at `sudo ls -la /root/.dotnet/corefx/cryptography/x509stores/my`

If I ran without sudo (just `dotnet dev-certs https -ep ...`), then I didn't have permission; if I ran `sudo -E` without first generating the cert (e.g. straight after `dotnet dev-certs https --clean`), then it failed creating the cert.

---
#### Document Details

* Content: [Enforce HTTPS in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?view=aspnetcore-5.0&tabs=visual-studio)
* Content Source: [aspnetcore/security/enforcing-ssl.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/security/enforcing-ssl.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

sgryphon:
Fixes #23378

Provide consistent instructions for exporting the Linux certificate that both has the needed permissions (sudo), and will use the current user's environment developer certificate (-E), that will work for the common scenario of a non-root user.

Also provide troubleshooting details for Linux, to see the thumbprint of both the Kestrel default certificate and the exported certificate being trusted. If they are not the same, then something went wrong, and the trust won't work.

Note there are many ways to configure/user Linux, e.g. as root, or as a user with permissions to add certificates, so the exact steps needed will vary. The most common way to use Linux as a developer is probably as a non-root user, but with sudo permissions, so instructions are based on that scenario.

The troubleshooting section provides details of where the certificates are found, and how to compare their thumbprints, so that a user can work out what the problem is, or adjust as necessary (e.g. if they need to use the root user certificate).

I also added a slight efficiency improvement to the Firefox instructions to provide a command line that will directly create the needed JSON file.

Note: Changes are separate commits, so you could cherry pick if you only want to include some of them.

### DIFF
--- a/aspnetcore/security/enforcing-ssl.md
+++ b/aspnetcore/security/enforcing-ssl.md
@@ -563,6 +563,28 @@ Close any browser instances open. Open a new browser window to app.
 
 See [HTTPS Error using IIS Express (dotnet/AspNetCore #16892)](https://github.com/dotnet/AspNetCore/issues/16892) for troubleshooting certificate issues with Visual Studio.
 
+### Linux - certificate not trusted
+
+Check that the certificate you are configuring for trust is the user HTTPS developer certificate that will be used by the dotnet Kestrel server.
+
+You can check the current user default HTTPS developer certificate (used by Kestrel) at the following location; the filename will be the SHA1 thumbprint. If you clear this file (via `dotnet dev-certs https --clean`) then it will be regenerated when needed, but with a different thumbprint.
+
+```
+ls -la ~/.dotnet/corefx/cryptography/x509stores/my
+```
+
+Check the thumbprint of the exported certificate (being trusted) matches:
+
+```
+openssl x509 -noout -fingerprint -sha1 -inform pem -in /usr/local/share/ca-certificates/aspnet/https.crt
+```
+
+If it does not match, you either have an old certificate, or may have exported a developer certificate for the root user, and will need to export the correct certificate again. You can check the root user certificate at:
+
+```
+ls -la /root/.dotnet/corefx/cryptography/x509stores/my
+```
+
 ### IIS Express SSL certificate used with Visual Studio
 
 To fix problems with the IIS Express certificate, select **Repair** from the Visual Studio installer. For more information, see [this GitHub issue](https://github.com/dotnet/aspnetcore/issues/16892).
--- a/aspnetcore/security/enforcing-ssl.md
+++ b/aspnetcore/security/enforcing-ssl.md
@@ -411,7 +411,11 @@ Establishing trust is distribution and browser specific. The following sections 
     sudo update-ca-certificates
     ```
 
-This first ensure the current user's developer certificate is created, and then exports it with elevated permissions needed for the `ca-certificates` folder, but using the current user's environment.
+The preceding commands:
+
+* Ensure the current user's developer certificate is created.
+* Exports the certificate with elevated permissions needed for the `ca-certificates` folder, using the current user's environment.
+* Removing the `-E`  flag exports the root user certificate, generating it if necessary. Each newly generated certificate has a different thumbprint. When running as root, `sudo`  and  `-E` are not needed.
 
 Without `-E` it would export the root user certificate (generating it if necessary), which has a different thumbprint. If you are running as root, then you won't need `sudo` (or the `-E`).
 
--- a/aspnetcore/security/enforcing-ssl.md
+++ b/aspnetcore/security/enforcing-ssl.md
@@ -458,17 +458,19 @@ For chromium browsers on Linux:
 
 * Create a JSON file at `/usr/lib/firefox/distribution/policies.json` with the following contents:
 
-  ```json
-  {
-      "policies": {
-          "Certificates": {
-              "Install": [
-                  "/usr/local/share/ca-certificates/aspnet/https.crt"
-              ]
-          }
-      }
-  }
-  ```
+```sh
+cat <<EOF | sudo tee /usr/lib/firefox/distribution/policies.json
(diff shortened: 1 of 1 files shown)


sgryphon:
Tag @Rick-Anderson

aspnetcore/security/enforcing-ssl.md
```diff
@@ -554,6 +563,28 @@ Close any browser instances open. Open a new browser window to app.
 
 See [HTTPS Error using IIS Express (dotnet/AspNetCore #16892)](https://github.com/dotnet/AspNetCore/issues/16892) for troubleshooting certificate issues with Visual Studio.
 
+### Linux - certificate not trusted
```
Rick-Anderson, line 566:
```suggestion
### Linux certificate not trusted
```

aspnetcore/security/enforcing-ssl.md
```diff
@@ -554,6 +563,28 @@ Close any browser instances open. Open a new browser window to app.
...
+
+Check that the certificate you are configuring for trust is the user HTTPS developer certificate that will be used by the dotnet Kestrel server.
```
Rick-Anderson, line 568:
```suggestion
Check that the certificate being configured for trust is the user HTTPS developer certificate that will be used by the Kestrel server.
```

aspnetcore/security/enforcing-ssl.md
```diff
@@ -554,6 +563,28 @@ Close any browser instances open. Open a new browser window to app.
...
+
+You can check the current user default HTTPS developer certificate (used by Kestrel) at the following location; the filename will be the SHA1 thumbprint. If you clear this file (via `dotnet dev-certs https --clean`) then it will be regenerated when needed, but with a different thumbprint.
```
Rick-Anderson, line 570:
```suggestion
Check the current user default HTTPS developer Kestrel certificate at the following location:
```

aspnetcore/security/enforcing-ssl.md
```diff
@@ -554,6 +563,28 @@ Close any browser instances open. Open a new browser window to app.
...
+
+```
+ls -la ~/.dotnet/corefx/cryptography/x509stores/my
+```
+
```
Rick-Anderson, line 575:
```suggestion

The HTTPS developer Kestrel certificate file is the SHA1 thumbprint. When the file is deleted via `dotnet dev-certs https --clean`, it will be regenerated when needed with a different thumbprint.
```

aspnetcore/security/enforcing-ssl.md
```diff
@@ -554,6 +563,28 @@ Close any browser instances open. Open a new browser window to app.
...
+Check the thumbprint of the exported certificate (being trusted) matches:
```
Rick-Anderson, line 576:
```suggestion
Check the thumbprint of the exported certificate matches with the following command:
```

aspnetcore/security/enforcing-ssl.md
```diff
@@ -554,6 +563,28 @@ Close any browser instances open. Open a new browser window to app.
...
+
+```
+openssl x509 -noout -fingerprint -sha1 -inform pem -in /usr/local/share/ca-certificates/aspnet/https.crt
+```
+
+If it does not match, you either have an old certificate, or may have exported a developer certificate for the root user, and will need to export the correct certificate again. You can check the root user certificate at:
```
Rick-Anderson, line 582:
```suggestion
If the certificate doesn't match, it could be one of the following:

* An old certificate.
* An exported a developer certificate for the root user. For this case, export the  certificate.

The root user certificate can be checked at:
```

Rick-Anderson (changes requested):
Wow, this should be super helpful. 
For the suggestions you like, select **Add suggestion to batch**and then use the button to batch commit them.

sgryphon:
I started doing them individually until I realised batch would be better.

I'm glad the draft made sense, and you have been able to improve the style to make it a bit cleaner / fit the voice better (e.g. dot points).

I hope it is useful for anyone else trying to set it up on Linux, especially the troubleshooting because if it isn't working (they don't match), it is not always clear why, so details how to check each of the thumbprints I think is really useful.

Rick-Anderson:
@javiercn  10 minute review, super helpful Linux update.

javiercn (approved):

Rick-Anderson (approved):

Rick-Anderson:
@sgryphon  thanks for this very helpful PR. Sorry it has taken so long. 

PR #23386, opened 2021-09-26
Update code snippets to the correct API - loadbalancing.md

erni27:
Update code snippets to the correct API.

https://github.com/grpc/grpc-dotnet/blob/6920e2abd016401ba2a602fbee45a25872acbea8/src/Grpc.Net.Client/GrpcChannelOptions.cs

### DIFF
--- a/aspnetcore/grpc/loadbalancing.md
+++ b/aspnetcore/grpc/loadbalancing.md
@@ -49,7 +49,7 @@ The resolver is configured using the address a channel is created with. The [URI
 
 A channel doesn't directly call a URI that matches a resolver. Instead, a matching resolver is created and used to resolve the addresses.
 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
> Update code snippets to the current API.

If these are changes from .NET 5 to .NET 6, I need to make a .NET 6 version of this file for you. We need to preserve the .NET 5 version while it's supported.

erni27:
Actually, that’s not an upgrade to .NET 6. At first, I thought it was because of an obselete API.
https://github.com/grpc/grpc-dotnet/blob/6920e2abd016401ba2a602fbee45a25872acbea8/src/Grpc.Net.Client/GrpcChannelOptions.cs
But now everything points to a small omission by the author.

guardrex:
That's right ... this needs the update. `GrpcChannelOptions` ...

>Credentials ... A ChannelCredentials instance.

https://docs.microsoft.com/aspnet/core/grpc/configuration#configure-client-options

guardrex (approved):
Thanks @erni27! :rocket:

PR #23406, opened 2021-09-29
Simplify

vanillajonathan:
The arguments are optional and not normally required


Rick-Anderson:
That's the 3.1 version which I believe requires that code.

vanillajonathan:
The argument is to Swagger, so the requirement would be on the side of Swagger, not .NET 3.1 or ASP.NET 3.1.
The csproj references ` <PackageReference Include="Swashbuckle.AspNetCore" Version="5.0.0-rc2" />`, an older version.
I am not sure how it is in that version, but in the current version of Swagger it is possible to omit the arguments to the `UseSwaggerUI`.

spboyer:
@bradygaster better to review as the PM

bradygaster:
It is true that in later versions that can be omitted so I'm fine with us taking this change. The SB owners have been quite responsive on making the APIs easier to use, so this tweak is ok now.

PR #23428, opened 2021-10-01
Use ISO date syntax in web-config.md

ericmutta:
The section on `stdoutLogFile` shows how a log file saved on `2/5/2018` would be named `stdout_20180205194132_1934.log`. For international readers like me, it's not clear whether `2/5/2018` means 2/May/2018 or 5/Feb/2018. To remove the ambiguity, I have tweaked the example to use the ISO date `2018-02-05` which is how it appears in the log file name as well.


aspnetcore/host-and-deploy/iis/web-config.md
```diff
@@ -107,7 +107,7 @@ For information on IIS sub-application configuration, see <xref:host-and-deploy/
...
+| `stdoutLogFile` | <p>Optional string attribute.</p><p>Specifies the relative or absolute file path for which `stdout` and `stderr` from the process specified in `processPath` are logged. Relative paths are relative to the root of the site. Any path starting with `.` are relative to the site root and all other paths are treated as absolute paths. Any folders provided in the path are created by the module when the log file is created. Using underscore delimiters, a timestamp, process ID, and file extension (`.log`) are added to the last segment of the `stdoutLogFile` path. If `.\logs\stdout` is supplied as a value, an example stdout log is saved as `stdout_20180205194132_1934.log` in the `logs` folder when saved on 2018-02-05 at 19:41:32 with a process ID of 1934.</p> | `aspnetcore-stdout` |
```
guardrex, line 110: I wonder if it could be even more explicit tho ...

```suggestion
| `stdoutLogFile` | <p>Optional string attribute.</p><p>Specifies the relative or absolute file path for which `stdout` and `stderr` from the process specified in `processPath` are logged. Relative paths are relative to the root of the site. Any path starting with `.` are relative to the site root and all other paths are treated as absolute paths. Any folders provided in the path are created by the module when the log file is created. Using underscore delimiters, a timestamp, process ID, and file extension (`.log`) are added to the last segment of the `stdoutLogFile` path. If `.\logs\stdout` is supplied as a value, an example stdout log is saved as `stdout_20180205194132_1934.log` in the `logs` folder when saved on February 5, 2018 at 19:41:32 with a process ID of 1934.</p> | `aspnetcore-stdout` |
```

guardrex (approved):

aspnetcore/host-and-deploy/iis/web-config.md
ericmutta: @guardrex that's even better, zero room for ambiguity left there :+1:

guardrex:
Thanks @ericmutta! :rocket:

ericmutta:
Always a pleasure @guardrex, many thanks for the fast follow up :rocket:

ericmutta:
Hi @guardrex, while on the subject of date syntax, I notice the American format is also used at the very top of each docs page, e.g. for this page here https://docs.microsoft.com/en-us/dotnet/core/run-time-config/#runtimeconfigjson we have:

image

When looking at the raw markdown file here https://github.com/dotnet/docs/blob/main/docs/core/run-time-config/index.md?plain=1  The date appears to be hardcoded in American format next to **ms.date**:

image

Is there an easy way to change the **ms.date** value in all the docs out there so it uses the unambiguous `February 5, 2018` syntax we used for this PR? Or maybe there's a per-user setting on docs.microsoft.com that renders the date localized to the user?

guardrex:
It's by convention and applies to all of MS docs, not just our repo. Float your concern/suggestions on it at the general MS docs site ...

https://github.com/MicrosoftDocs/feedback/issues

guardrex:
Actually, you can add a :+1: to this ... it's already an issue there ...

https://github.com/MicrosoftDocs/feedback/issues/3323

guardrex:
... but there are also closed issues, including one that says it shouldn't be a problem (but it probably still is) ...

https://github.com/MicrosoftDocs/feedback/issues/1326

... and another one that says they never fixed it but may have opened an internal issue ...

https://github.com/MicrosoftDocs/feedback/issues/386

You might want to call all of these out on that first one that I mentioned in my previous comment and see if you can encourage them to address it or close it out firmly as a *won't-fix*. It's not clear really if they're interested, but I see that managers are involved in the discussion on these issues.

ericmutta:
@guardrex many thanks for the pointers to those issues, one of them indicated that you can [scroll to the bottom and change the locale](https://github.com/MicrosoftDocs/feedback/issues/3323#issuecomment-932777847) so at least the dates now render as expected on docs.microsoft.com. Problem solved :rocket:

PR #23433, opened 2021-10-01
v6 Update: gRPC Diagnostics: Moniker versioning only

wadepickett:
[Internal Review: New moniker sections only are the only change.](https://review.docs.microsoft.com/en-us/aspnet/core/grpc/diagnostics?view=aspnetcore-6.0&branch=pr-en-us-23433)

Update gRPC Diagnostics topic moniker version section for .NET 6.0 only, no content updates.   

Then once this is merged, use this v6 monikered version to make v6 updates in the content so it is easier to diff to the original for reviews.

### DIFF
--- a/aspnetcore/grpc/diagnostics.md
+++ b/aspnetcore/grpc/diagnostics.md
@@ -12,6 +12,7 @@ uid: grpc/diagnostics
 
 By [James Newton-King](https://twitter.com/jamesnk)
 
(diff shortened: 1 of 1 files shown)


wadepickett:
Tested new moniker sections, now merging so that update can be used for actual content update to v6, making it easier to diff changes in the content for review.

PR #23452, opened 2021-10-05
Fix modifier declaration order

vanillajonathan:


guardrex (approved):
Thanks @vanillajonathan! :rocket: ... and I see a few more of these, and I'll fix those on a separate PR right now.

PR #23453, opened 2021-10-05
Correct typo in example filepath

marquisdan:
Path should be "\~/js/qr.js" not "\~/js/qs.js"


guardrex (approved):
Thanks @marquisdan! :rocket:

PR #23469, opened 2021-10-06
Fix action name

Bouke:


Rick-Anderson:
@fiyazbinhasan  please review

fiyazbinhasan (commented):
Instead of changing this, you might wanna change the reference snippet in the doc like this,
`[!code-csharp[](samples/TagHelpersBuiltIn/Controllers/SpeakerController.cs?range=18-20)]`

PR #23515, opened 2021-10-12
Fix broken href

vanillajonathan:
It seems that chapter hrefs are suffixed with `-1` for some reason.


guardrex:
Hello @vanillajonathan ... This is a known issue at this time due to the use of our current versioning. Duplicate headings get auto-numbered `id` attributes leading to this, which breaks links after a preview version is added to the doc set (i.e., 6.0). The links will somewhat repair themselves when 6.0 releases. The permanent fix for this is coming next year. **_Version by File_** will place different versions of topics into different markdown files, which will eliminate the duplicate headings and stop this from happening.

PR #1862, opened 2016-09-10
Update sociallogins.rst

jamiewest:
The previous example did not work, I updated the block to reflect what will work under the guide instructions without having to add additional usings.

PR #23668, opened 2021-10-28
Blazor native deps (w/callback coverage)

### LINKED ISSUE #23661: Document how to setup callbacks for native dependencies in Blazor WebAssembly

See https://github.com/dotnet/runtime/issues/60824#issuecomment-953035187.

@lewing @radical @lambdageek 

---
#### Document Details

* ID: be59faf1-5a20-b7e9-19ff-06589023bb15
* Version Independent ID: 250e7458-3c1d-0117-261d-c03ac136f16a
* Content: [Host and deploy ASP.NET Core Blazor WebAssembly](https://docs.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/webassembly?view=aspnetcore-6.0#native-dependencies-support)
* Content Source: [aspnetcore/blazor/host-and-deploy/webassembly.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/host-and-deploy/webassembly.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #23661

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/webassembly-native-dependencies?view=aspnetcore-6.0&branch=pr-en-us-23668)

This guidance requires further work:

* The `WasmBuildNative` property wasn't in the blog post. I *guessed* where/how to describe it.
* If an example makes sense for the callback guidance, place it in an issue comment here for inclusion.
* I can't get the SkiaSharp example to work following this guidance with the GA SDK. AFAICT, this is local to my machine, but I'll need to test on another PC here to confirm. The example and guidance is straight out of the blog post AFAICT ...

  https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-net-6-rc-2/#using-libraries-with-native-dependencies

  The [SkiaSharp repo example](https://github.com/mono/SkiaSharp/tree/main/samples/Basic/Blazor/WebAssembly) isn't an exact match to the scenario because it doesn't use the NuGet package.


mkArtakMSFT:
Thanks @guardrex.
@SteveSandersonMS can you please review this? Thanks!

aspnetcore/blazor/webassembly-native-dependencies.md
```diff
@@ -0,0 +1,159 @@
...
+
+*This article applies to Blazor WebAssembly.*
+
+Blazor WebAssembly apps can use native dependencies built to run on WebAssembly. You can statically link native dependencies into the .NET WebAssembly runtime using the .NET WebAssembly build tools, the same tools used to [ahead-of-time (AOT) compile](xref:blazor/host-and-deploy/webassembly#ahead-of-time-aot-compilation) a Blazor app to WebAssembly and to [relink the runtime to remove unused features](xref:blazor/host-and-deploy/webassembly#runtime-relinking).
+
+## .NET WebAssembly build tools
+
+The .NET WebAssembly build tools are based on [Emscripten](https://emscripten.org/), a compiler toolchain for the web platform. To install the .NET WebAssembly build tools, use either of the following approaches:
+
+* Select the optional component in the Visual Studio installer.
```
SteveSandersonMS, line 22: Could we be more explicit about its name here? Is it called ".NET WebAssembly build tools" in VS?

aspnetcore/blazor/webassembly-native-dependencies.md
```diff
@@ -0,0 +1,159 @@
...
+* Run `dotnet workload install wasm-tools` from an administrative command prompt.
```
SteveSandersonMS, line 23: Do you need to be admin? I thought I did it in the past without elevation.

aspnetcore/blazor/webassembly-native-dependencies.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Use native code
+
+Add a simple native C function to a Blazor WebAssembly app:
+
+1. Create a new Blazor WebAssembly project.
+1. Add the `WasmBuildNative` property to a property group in the app's project file (`.csproj`):
+
+   ```xml
+   <WasmBuildNative>true</WasmBuildNative>
```
SteveSandersonMS, line 44: You don't need this. As long as there's at least one `NativeFileReference`, it will do the right thing anyway.

aspnetcore/blazor/webassembly-native-dependencies.md
guardrex: I'll check on the name. I've removed VS 2019 but haven't installed VS 2022 yet ... I'll do that today.

aspnetcore/blazor/webassembly-native-dependencies.md
```diff
@@ -0,0 +1,159 @@
...
+   ```
+
+   > [!NOTE]
+   > The `WasmBuildNative` property can also be set with the `-p|--property` option passed to the `dotnet run` command:
+   >
+   > ```dotnetcli
+   > dotnet run -p:WasmBuildNative=true
```
SteveSandersonMS, line 51: Also don't need to do this

aspnetcore/blazor/webassembly-native-dependencies.md
```diff
@@ -0,0 +1,159 @@
...
+
+1. Add a `Test.c` file to the project.
+1. Add a C function in `Test.c` for computing factorials:
+
+   ```c
+   int fact(int n)
+   {
+       if (n == 0) return 1;
+       return n * fact(n - 1);
+   }
```
SteveSandersonMS, line 62: I'm surprised you don't need to mark the `fact` function as being exported in some way. As long as this code actually works as given then that's fine though.

aspnetcore/blazor/webassembly-native-dependencies.md
```diff
@@ -0,0 +1,159 @@
...
+
+   <p>@fact(3)</p>
+
+   @code {
+       [DllImport("Test")]
+       static extern int fact(int n);
+   }
+   ```
+
+When you build the app with the .NET WebAssembly build tools installed, the native C code is compiled and linked into the .NET WebAssembly runtime (`dotnet.wasm`). Compiling and linking may take a few minutes. After the app is built, run the app to see the rendered factorial value.
```
SteveSandersonMS, line 86: In the final builds, debug relinking was sped up a lot and should now only take 5-10 seconds. So we could probably drop the statement that it "may take a few minutes".

aspnetcore/blazor/webassembly-native-dependencies.md
```diff
@@ -0,0 +1,159 @@
...
+
+> [!NOTE]
+> During the 6.0 preview release period, you may receive a build error on subsequent builds saying that the output assembly is being used by another process. This is a known issue that will be addressed for the .NET 6 general release. To workaround the issue, rebuild the project a second time.
```
SteveSandersonMS, line 89: I guess this note can be removed now.

aspnetcore/blazor/webassembly-native-dependencies.md
```diff
@@ -0,0 +1,159 @@
...
+Label managed methods that are passed to C++ with the `[UnmanagedCallersOnly]` attribute.
+
+The method marked with the `[UnmanagedCallersOnly]` attribute must be `static`. To call an instance method in a Razor component, pass a `GCHandle` for the instance to C++ and then pass it back to native. Alternatively, use some other method to identify the instance of the component.
+
+The method marked with `[DllImport]` must use a C# 9.0 function pointer rather than a delegate type for the callback argument.
+
+> [!NOTE]
+> For C# function pointer types in `[DllImport]` methods, use `IntPtr` in the method signature on the managed side instead of `delegate *unmanaged<int, void>`. For more information, see [[WASM] callback from native code to .NET: Parsing function pointer types in signatures is not supported (dotnet/runtime #56145)](https://github.com/dotnet/runtime/issues/56145).
+
+## Use libraries
```
SteveSandersonMS, line 102: I think this section title is misleading. It says it's about using libraries, but the actual content is about *creating* libraries. People might mistakenly think that consuming a library like SkiaSharp involves doing stuff with architecture-specific directories, when they don't need to know or care about any of that.

Should the section title be changed to something like "Packaging native dependencies in a NuGet package"?

aspnetcore/blazor/webassembly-native-dependencies.md
```diff
@@ -0,0 +1,159 @@
...
+
+NuGet packages can contain native dependencies for use on WebAssembly. These libraries and their native functionality are then available to any Blazor WebAssembly app. The files for the native dependencies should be built for WebAssembly and packaged in the `browser-wasm` [architecture-specific folder](/nuget/create-packages/supporting-multiple-target-frameworks#architecture-specific-folders). WebAssembly-specific dependencies aren't referenced automatically and must be referenced manually as `NativeFileReference`s. Package authors can choose to add the native references by including a `.props` file in the package with the references.
+
+[SkiaSharp](https://github.com/mono/SkiaSharp) is a cross-platform 2D graphics library for .NET based on the native [Skia graphics library](https://skia.org/), and it now has preview support for Blazor WebAssembly.
+
+> [!WARNING]
+> [SkiaSharp](https://github.com/mono/SkiaSharp) and the [Skia graphics library](https://skia.org/) aren't owned or maintained by Microsoft. The demonstration example in this section is unsupported by Microsoft and shouldn't be used in production.
```
SteveSandersonMS, line 109: > shouldn't be used in production.

That's a strong statement. I think what we mean to say is something like "do this if you want but don't blame us if there's a problem", whereas it sounds like "you should not do this because it's a bad thing to do". We're not actually trying to say there's something bad about the following code, right? We just mean to communicate that we're not officially promising anything about it.

Would it be enough simply to say "... is unsupported by Microsoft"? I think that communicates the intent.

SteveSandersonMS (approved):
I've made a bunch of suggestions but more broadly this looks great. Thanks @guardrex!

aspnetcore/blazor/webassembly-native-dependencies.md
guardrex: That's from @danroth27's blog post, but I agree. I did it here without an admin command shell.

aspnetcore/blazor/webassembly-native-dependencies.md
guardrex: Again, it's from @danroth27 blog post. I've focused on the SkiaSharp example. I'll try this today before merging to make sure that it works.

aspnetcore/blazor/webassembly-native-dependencies.md
guardrex: I yield to @danroth27. We have a few spots where we say that preview/experimental packages are not supported for production use. For example ...

> `Microsoft.AspNetCore.ProtectedBrowserStorage` is an unsupported, experimental package that isn't intended for production use.

> The `Microsoft.AspNetCore.Components.WebAssembly.MultipartBundle` NuGet package and the `CustomPackagedApp` sample app are unsupported, experimental demonstration resources not intended for production use.

... and IIRC, we did something similar for `AspNetCore.Diagnostics.HealthChecks` while it was in preview. My intention is to remove that part of the remark when SkiaSharp goes off of preview.

I'll 👂 @danroth27 for guidance on these examples and update them all on this PR.

aspnetcore/blazor/webassembly-native-dependencies.md
guardrex: Yes, this works as written. 🎉

aspnetcore/blazor/webassembly-native-dependencies.md
guardrex: I'm asking Matthew if he has an ETA on a RTM/GA release at [Is there an ETA on RTM/GA for SkiaSharp.Views.Blazor? (mono/SkiaSharp #1862)](https://github.com/mono/SkiaSharp/issues/1862). I'll hold this PR for a sec to see if he says that the release is imminent. If so, I'll strike the warning about production use and leave our usual 'not supported by MS' remark.

danroth27: Users are free to use whatever features/code they want in production, as long as they've tested that they satisfy their needs. Pretty much all of .NET, stable or not, is provided "as is" as stated in our various licenses. So, we really only need to clarify the level of support and intent to release a stable version.

Suggested language: "Experimental and preview features are provided for the purpose of collecting feedback and are not supported for production use." 

The difference between experimental and preview features is that preview features are expect to eventual ship a stable release, while experimental features may not. Suggested language: "Experimental features are provided for the purpose of exploring feature viability and may not ship a future stable version".

PR #23697, opened 2021-10-31
fixed Json.NET casing

riccikl:
I noticed that Json.NET was not always spelled in the official way, so I changed it to use the correct name according to https://www.newtonsoft.com/json


guardrex (approved):
Thanks @riccikl! :rocket:

Issue #22045, opened 2021-04-13
Blazor 6.0 content tracking issue


guardrex:
### Assignment currently undetermined

* [ ] **Blazor WASM native dependencies support** | [Live](https://docs.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/webassembly?view=aspnetcore-6.0#native-dependencies-support) | [Edit](https://github.com/dotnet/AspNetCore.Docs/edit/main/aspnetcore/blazor/host-and-deploy/webassembly.md) | I hit a little 🐛 while working on this, but it seems to be some kind of Emscripten problem with my ancient 10-year-old PC 🙈.
* [ ] **WebAssembly AoT Compilation with runtime relinking guidance** | [AOT Live](https://docs.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/webassembly?view=aspnetcore-6.0#ahead-of-time-aot-compilation) | [Runtime Relinking Live](https://docs.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/webassembly?view=aspnetcore-6.0#runtime-relinking) | [Edit](https://github.com/dotnet/AspNetCore.Docs/edit/main/aspnetcore/blazor/host-and-deploy/webassembly.md) | Reviewed (but not approved) by DR.
* [ ] **Persist prerendering state** | [Live 1](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/prerendering-and-integration?view=aspnetcore-6.0&pivots=server#preserve-prerendered-state) | [Edit 1](https://github.com/dotnet/AspNetCore.Docs/edit/main/aspnetcore/blazor/components/prerendering-and-integration.md) | [Live (Tag Helper)](https://docs.microsoft.com/en-us/aspnet/core/mvc/views/tag-helpers/built-in/preserve-component-state?view=aspnetcore-6.0&pivots=server) | [Edit 2](https://github.com/dotnet/AspNetCore.Docs/edit/main/aspnetcore/mvc/views/tag-helpers/built-in/preserve-component-state.md) | Approved by DR.
[...]

PR #23708, opened 2021-11-02
Refactor Blazor Streaming Docs

TanayParikh:
@guardrex this is for your consideration. Given how the docs are broken up into _from dotnet_ and _from javascript_, I think it may make sense to split up the streaming sections appropriately as well. I've added cross-links to ensure this remains navigable.


guardrex (approved):
Yes, this was in the notes (in the 6.0 section) on https://github.com/dotnet/AspNetCore.Docs/issues/22045. I wasn't sure when it went up if back-to-back was better or not.

PR #23720, opened 2021-11-02
Blazor Working With Images Docs

TanayParikh:
Fixes: https://github.com/dotnet/aspnetcore/issues/36991
Fixes: https://github.com/dotnet/aspnetcore/issues/37016

### DIFF
--- a/aspnetcore/blazor/file-downloads.md
+++ b/aspnetcore/blazor/file-downloads.md
@@ -52,10 +52,10 @@ For this demonstration, the app creates a 50 KB file of random data from a new b
 
 Create a `DownloadFileFromStream` method that performs the following steps:
 
-* Retrieve the <xref:System.IO.Stream> from `GetFileStream`.
+* Retrieves the <xref:System.IO.Stream> from `GetFileStream`.
 * Specify a file name when file is saved on the user's machine. The following example names the file `log.bin`.
-* Wrap the <xref:System.IO.Stream> in a <xref:Microsoft.JSInterop.DotNetStreamReference>, which allows streaming the file data to the client.
-* Invoke `downloadFileFromStream`, which is a JavaScript function that accepts the data on the client. The `downloadFileFromStream` function is shown later in this article.
+* Wraps the <xref:System.IO.Stream> in a <xref:Microsoft.JSInterop.DotNetStreamReference>, which allows streaming the file data to the client.
+* Invokes `downloadFileFromStream`, which is a JavaScript function that accepts the data on the client. The `downloadFileFromStream` function is shown later in this article.
 
 ```razor
 @code {
--- a/aspnetcore/blazor/file-downloads.md
+++ b/aspnetcore/blazor/file-downloads.md
@@ -50,7 +50,7 @@ For this demonstration, the app creates a 50 KB file of random data from a new b
 }
 ```
 
-Create a `DownloadFileFromStream` method that performs the following steps:
+The following `DownloadFileFromStream` method performs the following steps:
 
 * Retrieves the <xref:System.IO.Stream> from `GetFileStream`.
 * Specify a file name when file is saved on the user's machine. The following example names the file `log.bin`.
--- a/aspnetcore/blazor/images.md
+++ b/aspnetcore/blazor/images.md
@@ -145,7 +145,7 @@ async function setImageUsingStreaming(imageElementId, imageStream) {
 }
 ```
 
-Call [`URL.revokeObjectURL()`](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL) to dispose of the object URL when done working with an image to prevent memory leaks:
+Call [`URL.revokeObjectURL()`](https://developer.mozilla.org/docs/Web/API/URL/revokeObjectURL) to dispose of the object URL when done working with an image to prevent memory leaks:
 
 ```javascript
 URL.revokeObjectURL(url);
--- a/aspnetcore/blazor/images.md
+++ b/aspnetcore/blazor/images.md
@@ -94,6 +94,12 @@ async function setImageUsingStreaming(imageElementId, imageStream) {
 }
 ```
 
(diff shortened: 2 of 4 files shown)


TanayParikh:
@guardrex for your consideration. As always feel free to change / edit as you see fit 😄

aspnetcore/blazor/images.md
```diff
@@ -0,0 +1,195 @@
...
+</button>
+```
+
+Next, we add a C# method that retrieves a <xref:System.IO.Stream> for the image to be displayed. This is where you may dynamically generate an image based on the specific user, or retrieve the image from storage. In this example, we retrieve the GitHub `dotnet` avatar.
+
+```razor
+@code {
+  private Stream GetImageStream()
+  {
+      return await HttpClient.GetStreamAsync("https://avatars.githubusercontent.com/u/9141961");
```
TanayParikh, line 99: Maybe we should use a different image source. ++ this may set a bad example as this could cause CORS related issues. Is there some prior precedent on accessing images elsewhere in the docs?

guardrex: There isn't ... it's a case-by-case basis. In the RCL topic, I favored providing guidance for fully-working examples where I tell the reader explicitly to right-click and download the images that the example uses, then they can use the code directly from the topic (cut-'n-paste) right into an app to play with the scenarios.

https://docs.microsoft.com/en-us/aspnet/core/blazor/components/class-libraries?view=aspnetcore-5.0&tabs=visual-studio#create-an-rcl-with-static-assets

We could do that here, but I'd recommend that I circle around in 2022 to set that up and just try to get this new topic merged for now.

aspnetcore/blazor/images.md
```diff
@@ -0,0 +1,193 @@
...
+
+`ResizeAndDisplayImageUsingStreaming` performs the following steps:
+
+* Accesses the <xref:Microsoft.AspNetCore.Components.Forms.InputFileChangeEventArgs.File> which is an <xref:Microsoft.AspNetCore.Components.Forms.IBrowserFile>.
+* Requests an image file from the specified <xref:Microsoft.AspNetCore.Components.Forms.IBrowserFile>, and resizes it to 250 pixels by 250 pixels.
+* Opens a <xref:System.IO.Stream> to read the `resizedImage`
+* Wrap the `resizedImage` <xref:System.IO.Stream> in a <xref:Microsoft.JSInterop.DotNetStreamReference>, which allows streaming the image data to the client.
+* Invokes `setImageUsingStreaming`, which is a JavaScript function that accepts the data on the client. The `setImageUsingStreaming` function is shown in the prior section.
+
+
```
TanayParikh, line 187:
```suggestion
Note this technique involves round-tripping the image data from the client to the server, and back. In a future version of .NET we may optimize this area to better facilitate image previews.

```

aspnetcore/blazor/file-downloads.md
```diff
@@ -52,10 +52,10 @@ For this demonstration, the app creates a 50 KB file of random data from a new b
 
 Create a `DownloadFileFromStream` method that performs the following steps:
```
guardrex, line 53: Since we're moving from reader instructions to a description of the following code ......

```suggestion
The following `DownloadFileFromStream` method performs the following steps:
```

However, more work is required later elsewhere here. I wouldn't sweat this type of thing too much at this point. All new topics receive 🦖 *The Rex Treatment*&trade; 🦖😄 at some point to clean up/improve the lingo and examples ... surface API doc cross-links ... place code into the snippet samples ... etc. Recommend making this change here but not sweating the rest for now.

guardrex (changes requested):
It looks like from the suggestion that you're still working on it. Should I hold off on an edit pass?

guardrex:
I'll be OOF for a few hours this afternoon. If you're ready for me to proceed :ear:, I'll edit and ping you back this evening. I couldn't tell if you were done because of the `suggestion` up there for Line 187 :point_right:.

TanayParikh:
> I'll be OOF for a few hours this afternoon. If you're ready for me to proceed 👂, I'll edit and ping you back this evening. I couldn't tell if you were done because of the `suggestion` up there for Line 187 ☝️.

Yep all ready, committed the suggestion for 187 😄

guardrex:
Ok, thanks. I'll ping u back later today for the final look and updates.

guardrex:
@TanayParikh ... Let's start with this, but I'll edit it again in the morning.

btw --- You don't say *where* to call ...

```javascript
URL.revokeObjectURL(url);
```

TanayParikh (commented):
This looks great, thanks!

TanayParikh:
> btw --- You don't say/show _where_ to call ...
> 
> ```js
> URL.revokeObjectURL(url);
> ```

There isn't really a set location to call `URL.revokeObjectURL(url)`, it depends on when you're done with the image. For example, in a form, after the user "submits" the data, you may wish to revoke the object URL as it will no longer be needed.

guardrex:
@TanayParikh ... I ran into a few more nittish-new-topic things this morning on the edit, and some of this is just an FYI for future reference:

* Fully working, cut-'n-paste examples are popular with readers. I go ahead here with the first (basic) example and set it up that way. With our built-in **Copy** code button in live docs, it's great. I've added this topic (and the download files topic) to my UE pass tracking issue for more work like this later.
* I added a line about when to call `URL.revokeObjectURL()`, but it's still confusing: We don't say "object URL" or have a `url` var anywhere above this remark, so it's still not clear where/how/on what it should be called. btw ---- and along the lines of a fully-working example ---- do you have a fully working example of this component that you could paste into a comment here? Immediately before the `revokeObjectURL` remarks, I could add it ... "Here's the complete code for the Razor component:" ... and then show the full component that would show `revokeObjectURL` being called. Then, we could have it say, "In the preceding component, `URL.revokeObjectURL()` is called to ..... (proceed with our current text).
* I moved the JS for the last two examples into a lead-in under an H2 heading. Again, this will facilitate (eventually) fully working/cut-'n-paste examples.

guardrex (approved):

TanayParikh:
This looks great, thanks!

Agree copy+paste would help simplify things 😄 

> Do you have these last two components/JS bits in fully working form such that you could paste them into a comment here (or provide a link to a GH repo where you have them)? If not, no worries ... I'll get them set up later when I get back to this topic for more work. I'll ping u on that PR, which very likely wouldn't be until 22H1.

I had working bits aligned with my initial commit, but there's been some churn in the docs so my code isn't up-to-date anymore. I think this should be fine for now ✅

PR #23721, opened 2021-11-02
Update `stopPropagation` Blazor Delegated Events Docs

TanayParikh:
Fixes: https://github.com/dotnet/aspnetcore/issues/30199

### DIFF
--- a/aspnetcore/blazor/components/event-handling.md
+++ b/aspnetcore/blazor/components/event-handling.md
@@ -315,7 +315,9 @@ An expression is also a permitted value of the attribute. In the following examp
 
 ## Stop event propagation
 
-Use the [`@on{DOM EVENT}:stopPropagation`](xref:mvc/views/razor#oneventstoppropagation) directive attribute to stop event propagation, where the `{DOM EVENT}` placeholder is a [Document Object Model (DOM) event](https://developer.mozilla.org/docs/Web/Events).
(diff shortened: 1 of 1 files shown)


guardrex:
@TanayParikh ... Is there a plainer way to describe the reasoning at ...

> This is because Blazor handles event delegation internally, and events must propagate to the root before Blazor can act upon them.

It invites a further question: What's special about the event propagating to the root?

**_Or,_** can it be simplified to just the following? ...

> This is because Blazor handles event delegation internally.

If so, then that's short enough to include in the topic sentence of the paragraph as ...

> The `stopPropagation` directive attribute's effect is limited to the Blazor scope and doesn't extend to the HTML DOM because Blazor handles event delegation internally.

TanayParikh:
> and events must propagate to the root before Blazor can act upon them.

I think the reason why this was specifically called out was because the click events still bubble up to the root, even with the `@stopPropogation` attribute. Users may wonder why these events are bubbling up if we have `@stopPropogation`, and the reason is, they must bubble up to the root before Blazor can act upon them. 


> If so, then that's short enough to include in the topic sentence of the paragraph as ...
> 
> > The `stopPropagation` directive attribute's effect is limited to the Blazor scope and doesn't extend to the HTML DOM because Blazor handles event delegation internally.

This is definitely more succinct, but I wonder if we should still add a note below saying something like:

Note: The `stopPropogation` directive attribute still requires events to bubble up the HTML DOM Hierarchy, as Blazor can only act upon events internally once they propagate to the root.

guardrex:
@TanayParikh ... I made an update to that line on the last commit. See if that's accurate and clear.

TanayParikh (commented):
LGTM!

guardrex (approved):

Issue #23695, opened 2021-10-31
Blazor Server subapp guidance: Can't find the fallback endpoint


cmedianu:
[Enter feedback here]

The instructions for setting up a Blazor Server application under /CoolApp/ are wrong and don't work.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e07aaa4d-4ad4-778f-f74a-60d8f05d9075
* Version Independent ID: c82cf121-831d-75dd-a796-c735aad3b3e9
* Content: [Host and deploy ASP.NET Core Blazor](https://docs.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/?view=aspnetcore-5.0&tabs=visual-studio)
* Content Source: [aspnetcore/blazor/host-and-deploy/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/host-and-deploy/index.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-hostdeploy**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
@cmedianu ... We're unable to provide product support, and your report has no specific information about why/how the guidance is incorrect. Reach out to devs on public support forums for assistance. We recommend the usual places ...

* [Stack Overflow (tag: `blazor`)](https://stackoverflow.com/questions/tagged/blazor)
* [ASP.NET Core Slack Team](http://tattoocoder.com/aspnet-slack-sign-up/)
* [Blazor Gitter](https://gitter.im/aspnet/Blazor)

cmedianu:
@guardrex Thanks Luke. Here's an example of adding the 2 lines from the instructions to a fresh generated project, the application crashes with:

_InvalidOperationException: Cannot find the fallback endpoint specified by route values: { page: /CoolApp/{**path:nonfile}, area: }._

https://dev.azure.com/cmedianu0603/publicProject/_git/BlazorServerPathChangeSoln/commit/c6e135c91ed8f2430d6a6af208bac222500bffeb?refName=refs%2Fheads%2Fmaster

guardrex:
Thanks for the additional info. I'll see if I can repro that behavior on Monday.

cmedianu:
Awesome!

guardrex:
[...]

PR #23735, opened 2021-11-04
Activate new/working MapFallbackToPage coverage

guardrex:
Addresses #23695

**_Draft-ish_** coverage at this point. I'm marking it off on the [tracking item check off list](https://github.com/dotnet/AspNetCore.Docs/issues/19286), but the topic is still up for a UE pass there. This will be improved when it's reached, probably in 22H1.

@cmedianu ... This is just an FYI that I worked out the coverage for this fallback path scenario *that makes sense now* 🙈 for Blazor apps. This use case is for a separate area of components and resources within the app that uses `MapFallbackToPage` with a separate area root page. The use case doesn't pertain to your original ask, but I just wanted to thank you again for the issue. It resulted in several important improvements to the coverage for sub-apps and this use case.

### DIFF
--- a/aspnetcore/blazor/host-and-deploy/index.md
+++ b/aspnetcore/blazor/host-and-deploy/index.md
@@ -131,21 +131,31 @@ dotnet run --pathbase=/CoolApp
 
 The Blazor WebAssembly app responds locally at `http://localhost:port/CoolApp`.
 
-<!-- HOLD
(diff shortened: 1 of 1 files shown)


guardrex:
Note to self (just so I don't lose it later): https://github.com/dotnet/aspnetcore/issues/32333#issuecomment-831159066

Issue #23738, opened 2021-11-04
Blazor Server SignalR message size


guardrex:
@TanayParikh ... This bit I probably missed, since I didn't understand what that work was about.

https://docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-dotnet-from-javascript?view=aspnetcore-6.0#size-limits-on-javascript-interop-calls

If the whole thing is no longer relevant, then I can drop the section. If it requires a re-write, then do you want to do it or paste the revised section into an issue comment here?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 2964d558-cd6d-7421-f7b9-cb939792c2cf
* Version Independent ID: a3517e56-69fc-1ec7-b55a-dbb54d68a807
* Content: [Call .NET methods from JavaScript functions in ASP.NET Core Blazor](https://docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-dotnet-from-javascript?view=aspnetcore-6.0)
* Content Source: [aspnetcore/blazor/javascript-interoperability/call-dotnet-from-javascript.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/javascript-interoperability/call-dotnet-from-javascript.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

TanayParikh:
Hey @guardrex, this is mostly still valid, I'd probably add a cross-link for streaming as the recommended alternative (as it really just does what the rest of the recomendations suggest internally (like chunking and avoiding blocking the UI thread). Will have a PR up shortly 😄

PR #23739, opened 2021-11-04
Update Blazor JSInterop Size Limits Docs

TanayParikh:
Fixes: https://github.com/dotnet/AspNetCore.Docs/issues/23738


guardrex (approved):
Thanks @TanayParikh ... I'm going to preserve the old section in a different INCLUDE for the earlier versions right after this goes in ... I'll take care of it right now on a separate PR.

Issue #22852, opened 2021-07-25
Create What's new in 6.0


Rick-Anderson:
[Enter feedback here]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 6c6a85af-14dd-7102-e2f9-d3850e0e0658
* Version Independent ID: d20b1af2-b6a4-5428-3e79-b2c8f8881446
* Content: [What's new in ASP.NET Core 5.0](https://docs.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-5.0?view=aspnetcore-5.0)
* Content Source: [aspnetcore/release-notes/aspnetcore-5.0.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/release-notes/aspnetcore-5.0.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-releasenotes**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

PR #23775, opened 2021-11-06
Blazor updates for What's New 6.0

guardrex:
Addresses #22852

cc: @Rick-Anderson @serpent5 @wadepickett 

Just a few nits at the moment. I'll have a few more updates on Monday morning for the Error Handling topic. The heading in the topic will change to match what's here, so the link here will ✨ *Just Work!*&trade; ✨ when my Monday morning update goes live. As for the rest of the MIA Blazor stuff, that's still being worked out internally. I'll flesh out the commented-out bits as soon as I have guidance from DR on the final Blazor coverage.

Side-note on the `view` qs param: It only works when non-6.0 docs load if that "preserve view" param is also supplied. I don't recall the exact param at the moment. I think all of the `view=aspnetcore-6.0` links should get it here. It's a confusing UE if we don't add that param to these links 😵.

### DIFF
--- a/aspnetcore/release-notes/aspnetcore-6.0.md
+++ b/aspnetcore/release-notes/aspnetcore-6.0.md
@@ -192,7 +192,7 @@ Binding supports multiple option selection with `<input>` elements. For more inf
 
 ### Head (`<head>`) content control
 
-Razor components can modify the HTML `<head>` element content of a page, including setting the page's title (`<title>` element) and modifying metadata (`<meta>` elements). For more information, see <xref:blazor/components/control-head-content?view=aspnetcore-6.0>.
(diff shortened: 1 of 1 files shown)


guardrex:
Taking this live right now. 🏃

guardrex:
@Rick-Anderson @wadepickett @serpent5 ... The updates for this morning to fix up the Blazor Handle Errors topic are on https://github.com/dotnet/AspNetCore.Docs/pull/23789 going in (and going live) this morning. That concludes the Blazor *What's New* updates at the moment. I'll be receiving further guidance from DR/Artak on the other commented-out entries. I'll contact you offline to coordinate further updates before opening any PR.

PR #23793, opened 2021-11-08
Fix typos

ryandle:
* Missing quote on first code example
* Extra bullet point at end of doc


serpent5:
Thanks @ryandle. That's a very good eye you have there. 🔍

Issue #23771, opened 2021-11-06
Update tutorial images to show LTS


Rick-Anderson:
Update all the tutorial images that currently show .NET 6 (Preview) to the correct LTS text.

wadepickett:
Note to self on an example of this:

![image](https://user-images.githubusercontent.com/10985336/140671244-fa68f98d-9366-44a3-8a04-2852e6a580c6.png)

wadepickett:
Will have these in soon today, updated .NET Hub page at the moment, almost done.

PR #23794, opened 2021-11-08
v6 vs2022 update for prereq include

wadepickett:
[Internal Review](https://review.docs.microsoft.com/en-us/aspnet/core/migration/50-to-60?view=aspnetcore-6.0&branch=pr-en-us-23794&tabs=visual-studio)

Updated the v6 prereq include to point to VS 2022 rather than preview.

Contributes towards #23771.  Doing this item in a separate PR to get it live quickly since it is used in several topics.


wadepickett:
@Rick-Anderson,  keeping this one change in its own PR so it can get out quickly.  Needs your review.

Rick-Anderson (approved):

PR #23808, opened 2021-11-09
Reference the correct code snippet in "Customize IWebHostBuilder" section

hmz777:
Reference the correct code snippet when changing the HTTP server implementation to be HTTP.sys based.


Rick-Anderson:
Nice work @hmz777 
How did you find the right snippet? 

PR #23824, opened 2021-11-09
Comment out hot reload for now

Rick-Anderson:


guardrex (changes requested):
Can't do this. Sourabh links to this until DR/Artak pull the trigger on the Hot Reload coverage move.

PR #23826, opened 2021-11-09
Configure xrefService in docxfx.json

serpent5:
cc @Rick-Anderson, @guardrex, @wadepickett.

This should mean we no longer need to provide the `--xrefService` argument for DocFx. I'm just running it through the build-system to make sure I haven't missed anything. Please let me know if you see any issues before I consider merging.

### DIFF
--- a/aspnetcore/docfx.json
+++ b/aspnetcore/docfx.json
@@ -74,6 +74,9 @@
     "markdownEngineName": "markdig",
     "groups": {
       "group1": { "dest": "group1-dest", "moniker_range": ">= aspnetcore-1.0" }
-    }
+    },
+    "xrefservice": [
+      "https://xref.docs.microsoft.com/query?uid={uid}"
+    ]
   }
 }


Rick-Anderson (approved):
I'm fine but let's get @guardrex  to approve before merging.

guardrex:
I don't know if this change will have side effects 😈. I can't approve because I can't assume any further risk for breaking the docs. I recommend getting a DocFX guru or member of docs management to approve. They can take the 🔥 **_heat_** 🔥 😄 if it all goes sideways.

Rick-Anderson:
> I can't approve because I can't assume any further risk for breaking the docs.

Why don't we test it at night and see if it breaks things. That is, when @serpent5 starts work.

Rick-Anderson:
@serpent5  can you do some testing to make sure we don't need to revert this?

serpent5:
I've received confirmation that this setting doesn't impact the builds here, so we're good there. Unfortunately, this approach is deprecated from DocFX v3, which is in beta right now.

guardrex:
> this approach is deprecated from DocFX v3

Do they have some other way to address it for when v3 goes into use, or will the problem with not finding the XREFs on a local build not be a problem with v3?

serpent5:
I asked that same question, where the response was around how it's a perfect fit for local validation using the VS Code authoring pack. Honestly, I'm not really sure what that means, but I'm trying to find out.

guardrex:
No comment.

# 😆

PR #23845, opened 2021-11-10
*DO NOT MERGE* Test Time-to-Read

serpent5:


serpent5:
@guardrex @Rick-Anderson Are you aware of an issue with the time-to-read that's shown at the top of an article? It seems the time is calculated using *all* moniker ranges, rather than just for the one that's selected.

guardrex:
I'm not aware of that particular problem, but that's always been a goofy number (e.g., when the concepts are complex, one can easily double or **triple** whatever number it provides).

serpent5:
@guardrex Ok, thanks. I added a 3.0-specific version of the topic in this PR and the time went up by 17 minutes, so I think there's an issue specifically with the moniker ranges. I'll try and find out more about it and report back.

guardrex:
It's probably just going by the length of the whole markdown file ...... again ..... super silly imo because code examples can make topics long but not long to read. The whole doc feature is a bit silly 😄.

serpent5:
Yes, I see what you mean here. If it is just going off the .md file, it might not even be including the code that's pulled in from files. I might look into that too, but I guess bigger 🐟 and all that. Anyway, thanks for your input. I won't bother you with this any longer. I won't be offended if you want to unsubscribe 😆 😄

Rick-Anderson:
TTR (time to read) is always way under estimated, anything that makes it longer is an improvement.

PR #23872, opened 2021-11-11
Blazor remove documentation of `_reconnectionDisplay`

TanayParikh:
`Blazor.defaultReconnectionHandler._reconnectionDisplay` is a framework internal API. I believe we may want to avoid officially documenting it as its subject to change without notice.

### DIFF
--- a/aspnetcore/blazor/fundamentals/signalr.md
+++ b/aspnetcore/blazor/fundamentals/signalr.md
@@ -267,39 +267,6 @@ To adjust the reconnection retry count and interval, set the number of retries (
 
 For more information on Blazor startup, see <xref:blazor/fundamentals/startup>.
 
-## Hide or replace the reconnection display (Blazor Server)
-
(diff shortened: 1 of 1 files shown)


guardrex (approved):

guardrex:
Build report looks great, @TanayParikh ... merge when ur ready.

guardrex:
Although, this text says "transition-delay" but the prop is just "transition".

TanayParikh:
> Although, this text says "transition-delay" but the prop is just "transition".

Just copied that text out of the existing section (as it didn't seem directly related to _reconnectionDisplay)

guardrex:
I guess most devs know that the third entry will map to it.

TanayParikh:
> I guess most devs know that the third entry will map to it.

> The transition CSS property is a shorthand property for transition-property, transition-duration, transition-timing-function, and transition-delay.

https://developer.mozilla.org/en-US/docs/Web/CSS/transition

TanayParikh:
Merged, thanks!

PR #23898, opened 2021-11-13
Shim startup/ra/2

Rick-Anderson:
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/migration/50-to-60?view=aspnetcore-5.0&branch=pr-en-us-23898&tabs=visual-studio#wraps)

GitHub diff is badly confused, so look at [Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/migration/50-to-60?view=aspnetcore-5.0&branch=pr-en-us-23898&tabs=visual-studio#wraps) and not the diff code.

Is this too obvious or is it helpful to those who want to migrate piecemeal?


aspnetcore/migration/50-to-60.md
```diff
@@ -146,17 +146,31 @@ The preceding code migrated to the new minimal hosting model:
...
+
 When using a custom dependency injection (DI) container, add the following highlighted code:
 
 [!code-csharp[](~/migration/50-to-60/samples/WebRP31to60/Program.cs?name=snippet_60a&highlight=11-13)]
 
 [!code-csharp[](~/migration/50-to-60/samples/WebRP31to60/Startup.cs?name=snippet_60a&highlight=17-21)]
 
+<a name="wraps"></a>
+
+When using the minimal hosting model, the endpoint routing middleware wraps the entire middleware pipeline, therefore there's no need to have explicit calls to `UseEndpoints` to register routes. `UseRouting` can still be used to specify where route matching happens but `UseRouting` doesn't need to be explicitly called.
```
halter73, line 159: Nit:

```suggestion
When using the minimal hosting model, the endpoint routing middleware wraps the entire middleware pipeline, therefore there's no need to have explicit calls to `UseRouting` or `UseEndpoints` to register routes. `UseRouting` can still be used to specify where route matching happens, but `UseRouting` doesn't need to be explicitly called if routes should be matched at the beginning of the middleware pipeline.
```

I think this makes it more clear where route matching happens by default. This is a little more wordy and possibly redundant given this already indicates " endpoint routing middleware wraps the entire middleware pipeline", so I understand if you don't take this suggestion.

aspnetcore/migration/50-to-60/samples/WebRP31to60/Startup.cs
```diff
@@ -124,5 +124,40 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
+        }
+
+        app.UseHttpsRedirection();
+        app.UseStaticFiles();
+        //app.UseRouting();
+
+        //app.UseEndpoints(endpoints =>
+        //{
+        //    endpoints.MapRazorPages();
+        //});
```
halter73, line 158: I get that this is unnecessary when using `WebApplication` in Program.cs, but why is it commented out here? Isn't highlighting it enough to indicate that this is what's being removed in the following sample?

halter73 (approved):

PR #23960, opened 2021-11-17
added @ to render partial view _ArticleSection in mvc example

vladsaftoiu:
in MVC partial view example, added `@` to actually render the article section partial view


serpent5:
Thanks @vladsaftoiu.

PR #23965, opened 2021-11-17
fix snippet language in dependency-injection

tufteddeer:
Corrects the language of this snippet to C# instead of json


guardrex (approved):
Thanks @tufteddeer! :rocket:

PR #23982, opened 2021-11-19
bool is true or false

yecril71pl:
> Wystąpił wyjątek: CLR/System.InvalidOperationException
> Wystąpił nieobsługiwany wyjątek typu „System.InvalidOperationException” w Microsoft.Extensions.Configuration.Binder.dll: > 'Failed to convert configuration value at 'hostBuilder:reloadConfigOnChange' to type 'System.Boolean'.'
> Znaleziono wewnętrzne wyjątki. Zobacz $exception w oknie zmiennych, aby uzyskać więcej szczegółów.
> 	Najbardziej wewnętrzny wyjątek 	 System.FormatException : String '1' was not recognized as a valid Boolean.


guardrex:
Hello @yecril71pl ... Numbers are accepted in string values for at least some of these (I'll check all of them given the error that you hit for `reloadConfigOnChange`). For example, `detailedErrors` ...

https://github.com/dotnet/aspnetcore/blob/a450cb69b5e4549f5515cdb057a68771f56cefd7/src/Hosting/Hosting/test/WebHostConfigurationsTests.cs#L47-L56

However, doc churn seems to have introduced a problem in the way these acceptable values are described since I worked on this topic a few years ago. `true`/`1` or `false`/`0` are the acceptable values, but updates don't seem to have kept up with that logic/approach in the descriptions. I've opened #23984 to address this. Thanks for pointing it out.

PR #23985, opened 2021-11-19
Update bool value entries for host config

### LINKED ISSUE #23984: Check/update bool value entries for host config
Check and update each entry as needed in host config settings.

For at least some of them, string numbers `1`/`0` are acceptable for `true`/`false`, while others must be provided as `true`/`false` only.

Cross-ref: #23982

---
#### Document Details

* ID: c5f1eeee-d500-82bf-0d0a-b4e0d5d5c175
* Version Independent ID: 2373a3df-6ea3-50c7-d63d-61c51d572511
* Content: [.NET Generic Host in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host?view=aspnetcore-5.0)
* Content Source: [aspnetcore/fundamentals/host/generic-host.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/host/generic-host.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

guardrex:
Fixes #23984

Thanks @yecril71pl! :rocket:

From identifying the `ParseBool`-calling options in `WebHostOptions` and given that `ParseBool` is a `true`/`1` **string** for `true` (**bool**) and `false`/`0`/anything **string** for `false` (**bool**), this PR sets the potential and recommended&dagger; values.

&dagger;Recommended in the sense that a nonsensical/nonstandard value is interpreted as `false`. We don't need to call it out, but ...

```csharp
webBuilder.UseSetting(WebHostDefaults.DetailedErrorsKey, 
    "Spare me the details, man! It's FRIDAY for crying out loud! 🎉🍻");
```

😆 Technically, that's ok. Then, it would reach code review, and you'd have to go find a new dev role. 🙈

* https://github.com/dotnet/aspnetcore/blob/main/src/Hosting/Abstractions/src/WebHostDefaults.cs
* https://github.com/dotnet/aspnetcore/blob/main/src/Hosting/Hosting/src/Internal/WebHostOptions.cs#L21-L29
* https://github.com/dotnet/aspnetcore/blob/main/src/Hosting/Hosting/src/Internal/WebHostUtilities.cs#L11-L15
* There's only a test for one key AFAICT ... I guess just to check that `ParseBool` is working ok. https://github.com/dotnet/aspnetcore/blob/a450cb69b5e4549f5515cdb057a68771f56cefd7/src/Hosting/Hosting/test/WebHostConfigurationsTests.cs#L47-L56

Also, it looks like we missed one back in the day, `SuppressStatusMessagesKey`. It's added here. The 2.x bits use a different format for options. I think we can let that go unchanged here ... and it's going away soon enough.

Kirk, I don't think we need PU review given that these are mostly nit-level updates to the descriptions and the clarity of the framework API. However, if you feel we should ping, I'll call for PU review as well. :ear:

... and btw ... not going for the 6.0 updates here. I need to get back to Blazor-ish things, and the whole topic is going to need more work on that set of updates.

### DIFF
--- a/aspnetcore/fundamentals/host/generic-host.md
+++ b/aspnetcore/fundamentals/host/generic-host.md
@@ -191,7 +191,7 @@ The [IHostEnvironment.ApplicationName](xref:Microsoft.Extensions.Hosting.IHostEn
 **Key**: `applicationName`  
 **Type**: `string`  
 **Default**: The name of the assembly that contains the app's entry point.  
-**Environment variable**: `<PREFIX_>APPLICATIONNAME`
+**Environment variable**: `{PREFIX_}APPLICATIONNAME`
 
 To set this value, use the environment variable. 
 
@@ -202,7 +202,7 @@ The [IHostEnvironment.ContentRootPath](xref:Microsoft.Extensions.Hosting.IHostEn
 **Key**: `contentRoot`  
 **Type**: `string`  
 **Default**: The folder where the app assembly resides.  
-**Environment variable**: `<PREFIX_>CONTENTROOT`
+**Environment variable**: `{PREFIX_}CONTENTROOT`
 
 To set this value, use the environment variable or call `UseContentRoot` on `IHostBuilder`:
 
@@ -224,7 +224,7 @@ The [IHostEnvironment.EnvironmentName](xref:Microsoft.Extensions.Hosting.IHostEn
 **Key**: `environment`  
 **Type**: `string`  
 **Default**: `Production`  
-**Environment variable**: `<PREFIX_>ENVIRONMENT`
+**Environment variable**: `{PREFIX_}ENVIRONMENT`
 
 To set this value, use the environment variable or call `UseEnvironment` on `IHostBuilder`:
 
@@ -246,7 +246,7 @@ If the timeout period expires before all of the hosted services stop, any remain
 **Key**: `shutdownTimeoutSeconds`  
 **Type**: `int`  
 **Default**: 5 seconds  
-**Environment variable**: `<PREFIX_>SHUTDOWNTIMEOUTSECONDS`
+**Environment variable**: `{PREFIX_}SHUTDOWNTIMEOUTSECONDS`
 
 To set this value, use the environment variable or configure `HostOptions`. The following example sets the timeout to 20 seconds:
 
@@ -260,7 +260,7 @@ By [default](xref:fundamentals/configuration/index#default), *appsettings.json* 
 **Type**: `bool` (`true` or `false`)  
 **Default**: `true`  
 **Command-line argument**: `hostBuilder:reloadConfigOnChange`  
-**Environment variable**: `<PREFIX_>hostBuilder:reloadConfigOnChange`
+**Environment variable**: `{PREFIX_}hostBuilder:reloadConfigOnChange`
 
 > [!WARNING]
 > The colon (`:`) separator doesn't work with environment variable hierarchical keys on all platforms. For more information, see [Environment variables](xref:fundamentals/configuration/index#environment-variables).
@@ -288,7 +288,7 @@ When `false`, errors during startup result in the host exiting. When `true`, the
 **Key**: `captureStartupErrors`  
 **Type**: `bool` (`true`/`1` or `false`/`0`)  
 **Default**: Defaults to `false` unless the app runs with Kestrel behind IIS, where the default is `true`.  
-**Environment variable**: `<PREFIX_>CAPTURESTARTUPERRORS`
+**Environment variable**: `{PREFIX_}CAPTURESTARTUPERRORS`
 
 To set this value, use configuration or call `CaptureStartupErrors`:
 
@@ -303,7 +303,7 @@ When enabled, or when the environment is `Development`, the app captures detaile
 **Key**: `detailedErrors`  
 **Type**: `bool` (`true`/`1` or `false`/`0`)  
 **Default**: `false`  
-**Environment variable**: `<PREFIX_>_DETAILEDERRORS`
+**Environment variable**: `{PREFIX_}DETAILEDERRORS`
 
 To set this value, use configuration or call `UseSetting`:
 
@@ -318,7 +318,7 @@ A semicolon-delimited string of hosting startup assemblies to load on startup. A
 **Key**: `hostingStartupAssemblies`  
 **Type**: `string`  
 **Default**: Empty string  
-**Environment variable**: `<PREFIX_>_HOSTINGSTARTUPASSEMBLIES`
+**Environment variable**: `{PREFIX_}HOSTINGSTARTUPASSEMBLIES`
 
 To set this value, use configuration or call `UseSetting`:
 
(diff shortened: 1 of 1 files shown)


aspnetcore/fundamentals/host/generic-host.md
```diff
@@ -301,7 +301,7 @@ webBuilder.CaptureStartupErrors(true);
 When enabled, or when the environment is `Development`, the app captures detailed errors.
 
 **Key**: `detailedErrors`  
-**Type**: `bool` (`true` or `1`)  
+**Type**: `bool` (`true`/`1` or `false`/`0`)  
 **Default**: `false`  
 **Environment variable**: `<PREFIX_>_DETAILEDERRORS`
```
serpent5, line 306: There's a few cases of this `<PREFIX_>_` double `_` in here. Do you want to fix them while you're in here, or shall I patch them up in a separate PR?

serpent5 (commented):
LGTM. I don't think this needs a PU review.

serpent5 (approved):

aspnetcore/fundamentals/host/generic-host.md
guardrex: I'll fix them ... and I think we've (or I've) been settling more on braces for these.

guardrex:
Ok ... try it now (and with braces). I usually say '... the `{PLACEHOLDER}` placeholder is the blah, blah, blah ...', but it's fairly straightforward.

guardrex:
Hold on a sec .... let me call out what it is at the top. brb

guardrex:
Ok ... @serpent5 ... try it now with some callouts for the prefix and with braces in the examples. 👁️👂 ....... 👃😄 ...... I can see that it's going to be a 🦖 *SillyRex Friday*&trade; 🦖 today!

aspnetcore/fundamentals/host/generic-host.md
```diff
@@ -180,7 +180,7 @@ For more information, see [Configuration in ASP.NET Core](xref:fundamentals/conf
 
 ## Settings for all app types
 
-This section lists host settings that apply to both HTTP and non-HTTP workloads. By default, environment variables used to configure these settings can have a `DOTNET_` or `ASPNETCORE_` prefix. For more information, see the [Default builder settings](#default-builder-settings) section.
+This section lists host settings that apply to both HTTP and non-HTTP workloads. By default, environment variables used to configure these settings can have a `DOTNET_` or `ASPNETCORE_` prefix, which appear in the following configuration for the `{PREFIX_}` placeholder. For more information, see the [Default builder settings](#default-builder-settings) section.
```
serpent5, line 183:
```suggestion
This section lists host settings that apply to both HTTP and non-HTTP workloads. By default, environment variables used to configure these settings can have a `DOTNET_` or `ASPNETCORE_` prefix, which appear in the following list of settings as the `{PREFIX_}` placeholder. For more information, see the [Default builder settings](#default-builder-settings) section.
```

aspnetcore/fundamentals/host/generic-host.md
```diff
@@ -257,17 +257,17 @@ To set this value, use the environment variable or configure `HostOptions`. The
...
-**Environment variable**: `<PREFIX_>hostBuilder:reloadConfigOnChange`
+**Environment variable**: `{PREFIX_}hostBuilder:reloadConfigOnChange`
 
 > [!WARNING]
 > The colon (`:`) separator doesn't work with environment variable hierarchical keys on all platforms. For more information, see [Environment variables](xref:fundamentals/configuration/index#environment-variables).
 
 ## Settings for web apps
 
-Some host settings apply only to HTTP workloads. By default, environment variables used to configure these settings can have a `DOTNET_` or `ASPNETCORE_` prefix.
+Some host settings apply only to HTTP workloads. By default, environment variables used to configure these settings can have a `DOTNET_` or `ASPNETCORE_` prefix, which appear in the following configuration for the `{PREFIX_}` placeholder.
```
serpent5, line 270:
```suggestion
Some host settings apply only to HTTP workloads. By default, environment variables used to configure these settings can have a `DOTNET_` or `ASPNETCORE_` prefix, which appear in the following list of settings as the `{PREFIX_}` placeholder.
```

serpent5 (approved):
Yup, looking good. Just a nit suggestion for consideration, but feel free to go as-is.

guardrex:
My only **_minor_** concern is that the prefix isn't as limited as this **_may_** suggest. It can be anything that the dev wants, too. However, that's not the main thrust of the PR's fixes. This PR doesn't limit the value, it just says that the prefix can be one of these values. This topic has been overhauled a few times in the history of ASP.NET Core 👴 ... no doubt it will need that treatment again in a few years, and the doc author might make some further improvements on this aspect.

guardrex:
Yeah ... I just re-read the prefix lingo I think what I had here ... *many years ago in the early days of ASP.NET Core when 🦖🦕 walked the Earth* 😄 ... I think what I had for this business of the prefix was *better* **IMHO** than what's in the topic today. Anyway ... it's not blocking to getting these fixes in. I think that the next doc author should think about that aspect further.

serpent5:
I'm not sure I fully understand what you're saying about the prefixes, but perhaps we can discuss that at a later date. Rick is prioritising the 6.0 update issues, so I think he'll get to opening one for this as it comes to the top of the list.

guardrex:
The dev can have **_any prefix_** that they want IIRC, not just `ASPNETCORE_`/`DOTNET_`. However, the topic doesn't call it out that way or imply it. My 🧠 is fuzzy on the details beyond that.

guardrex:
It might be that ...

```csharp
configHost.AddEnvironmentVariables(prefix: "PREFIX_");
```

... is the way. Put something else for `PREFIX_` there, and it can be anything.

```csharp
configHost.AddEnvironmentVariables(prefix: "SILLYREX_");
```

... and it's stripped off and loads the values. **_However_**, don't quote me on that. It's been several years. The old brain box 🧠 has been adversely effected by too much 🍺😄 to recall all of the juicy details. That setting (and further explanation) might live today over in the Environments topic.

serpent5:
I think you'd have to modify the host builder with your own call to e.g. `AddEnvironmentVariable("MYPREFIX_")` for that to affect the host configuration. The section where you've added coverage of what `{PREFIX_}` is for states that it's "by default" for the specific prefixes, so it might not to be too bad.

guardrex:
Beat you to it, @serpent5. 😆 ........ It's that third cup of ☕ that I just had. I'm *almost fully AWAKE now* 😄.

guardrex:
DAH! This build broke. Let's try again ........................

serpent5:
I guess you could even just call `AddEnvironmentVariables()` without _any prefix_. 🧠 🤯

guardrex:
... right ... and not something (IIRC) that the framework APIs do during a host build/config.

Anyway ... I actually didn't find in Environments. It's over in ...

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-6.0#environment-variables

I think we need a cross-link for that, and I may as well add it here. brb

PR #24007, opened 2021-11-20
A controller is an instantiable public class

yecril71pl:
I understand that from the library POV a non-public class is non-instantiable.  However, when I look at my code, I can instantiate the controller class all right, so I cannot see the problem.  Also, there is no warning from the library when a controller class is non-public, so there is no way to discover the problem at run time.  I hope this small amendment will do no harm.

### DIFF
--- a/aspnetcore/mvc/controllers/actions.md
+++ b/aspnetcore/mvc/controllers/actions.md
@@ -22,7 +22,7 @@ By convention, controller classes:
 * Reside in the project's root-level *Controllers* folder.
 * Inherit from `Microsoft.AspNetCore.Mvc.Controller`.
 
-A controller is an instantiable public class in which at least one of the following conditions is true:
+A controller is an instantiable class, usually [public](/dotnet/csharp/language-reference/keywords/public), in which at least one of the following conditions is true:
 
 * The class name is suffixed with `Controller`.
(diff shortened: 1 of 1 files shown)


guardrex:
I think the problem with this change is that just because a controller isn't marked `public` (which would be strange outside of just making the actions unavailable for some reason in one spot) it doesn't change the definition of what a controller is. By adding the word "public" here, it's conditionally defining the object. If something like this were done, it seems like the language should be more along the lines of '... usually public ...'.

cc: @Rick-Anderson @serpent5

aspnetcore/mvc/controllers/actions.md
```diff
@@ -22,7 +22,7 @@ By convention, controller classes:
 * Reside in the project's root-level *Controllers* folder.
 * Inherit from `Microsoft.AspNetCore.Mvc.Controller`.
 
-A controller is an instantiable class in which at least one of the following conditions is true:
+A controller is an instantiable public class in which at least one of the following conditions is true:
```
guardrex, line 25: Don't accept this suggestion just yet. It's more **_for discussion_** among us at this point .......

```suggestion
A controller is an instantiable class, usually [public](/dotnet/csharp/language-reference/keywords/public), in which at least one of the following conditions is true:
```

PR #24009, opened 2021-11-20
NavigationManager.BaseUri uses Request.PathBase

Bouke:
It was unclear to me how `NavigationManager.BaseUri` was set when pre-rendering, and discovered that it uses `Request.PathBase`. I think this is useful information to provide to users.


guardrex:
Thanks @Bouke ... I'll take a look no later than Monday morning and get back to you.

guardrex:
It's taken directly from the API doc, which is what the PU ("product unit") wrote about it ...

https://docs.microsoft.com/dotnet/api/microsoft.aspnetcore.components.navigationmanager.baseuri

Initialization don't provide further detail. It's probably considered by the PU to be an implementation detail, subject to change for any release.

https://docs.microsoft.com/dotnet/api/microsoft.aspnetcore.components.navigationmanager.initialize

I was under the impression that in some cases, WASM perhaps (but not necessarily prerendered WASM), it comes from `document.baseURI` via JS interop when the Navigation Manager is created.

Let's ask Javier. If he agrees with this (or some alternate version of it), then I'll make a slight grammar+style update and get the improvement in.

aspnetcore/blazor/fundamentals/routing.md
```diff
@@ -216,7 +216,7 @@ Use <xref:Microsoft.AspNetCore.Components.NavigationManager> to manage URIs and
 | Member | Description |
 | ------ | ----------- |
 | <xref:Microsoft.AspNetCore.Components.NavigationManager.Uri> | Gets the current absolute URI. |
-| <xref:Microsoft.AspNetCore.Components.NavigationManager.BaseUri> | Gets the base URI (with a trailing slash) that can be prepended to relative URI paths to produce an absolute URI. Typically, <xref:Microsoft.AspNetCore.Components.NavigationManager.BaseUri> corresponds to the `href` attribute on the document's `<base>` element in `wwwroot/index.html` (Blazor WebAssembly) or `Pages/_Layout.cshtml` (Blazor Server). |
+| <xref:Microsoft.AspNetCore.Components.NavigationManager.BaseUri> | Gets the base URI (with a trailing slash) that can be prepended to relative URI paths to produce an absolute URI. Typically, <xref:Microsoft.AspNetCore.Components.NavigationManager.BaseUri> corresponds to either the `href` attribute on the document's `<base>` element in `wwwroot/index.html` (Blazor WebAssembly) or `Pages/_Layout.cshtml` (Blazor Server), or the `PathBase` property of the `HttpRequest` when pre-rendering. |
```
javiercn, line 219: This is not really true unless you use `~/` in your base tag

guardrex:
I didn't get full clarity on what `PathBase` coorosponds to *exactly* ... just the segments prior any specified by the `<base>` tag's `href` or including any segments specified in `href`. I'll mark this a DRAFT until I test here and see the behavior so that I can compose the guidance clearly.

aspnetcore/blazor/fundamentals/routing.md
Bouke: What is not really true, regarding the addition I made there?

When pre-rendering the NavigationManager is initialized from the current request, no blazor hub / `<base />` involved.  See `StaticComponentRenderer` initializing the `NavigationManager` from the `request` [`GetContextBaseUri(httpContext.Request)`](https://github.com/dotnet/aspnetcore/blob/ae1a6cbe225b99c0bf38b7e31bf60cb653b73a52/src/Mvc/Mvc.ViewFeatures/src/RazorComponents/StaticComponentRenderer.cs#L80). It passes the request to `GetContextBaseUri` which builds the absolute URI using only `Scheme` + `Host` + `PathBase`. 

So the wording should be along the lines of "or the absolute URI to the `PathBase` property of the `HttpRequest` when pre-rendering.".

guardrex: ... and btw while you sort that out, I say again that this added bit pertains to one hosting model (apparently). We need to mention it if it's the case. I'll work on the grammar/style bits once it's clarified.

Bouke:
`PathBase` is the part of the `Path` that was matched when executing the middleware. See also [Branch the middleware pipeline](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/?view=aspnetcore-6.0#branch-the-middleware-pipeline):

> When `Map` is used, the matched path segments are removed from `HttpRequest.Path` and appended to `HttpRequest.PathBase` for each request.

This is a somewhat confusing part of the Blazor routing engine. I'm hosting my blazor hub in a multi-tenant website, with the first path segment being the tenant. E.g. `/{tenant}/_blazor`. So I'm setting `<base href="/@tenant/" />` and hosting the hub using `endpoints.MapBlazorHub("/{tenant}/_blazor");`. This is sufficient to get Blazor Server working and I can specify component `@page /my-page` which makes the `Router` map the path `/{tenant}/my-page` to this component. So far so good. But it fails when pre-rendering:

> Displaying NotFound because path 'some_tenant/my-page' with base URI 'https://localhost:5001/' does not match any component route

To get pre-rendering to work, I somehow need to change the `PathBase`. I cannot use `Map` as the segment is unique (multiple tenants). Suggestions welcome.

guardrex:
> at a non-standard path

Prerendering doesn't fail when the exact path segment is provided to the base uri or `UsePathBase`. You're the first to ask about a multitenant routing scenario with an approach like that (here on the docs repo anyway). There's one issue that I see on the PU repo that might pertain to your ask, @Bouke. Could take a look at the following issue and see if anything there helps ...

https://github.com/dotnet/aspnetcore/issues/15369

In addition to Javier remarking on the scenario ... how to make it work, he may advise if a dedicated section on this subject is warranted.

Bouke:
It's not really related to my multi-tenant setup, but hosting Blazor on a (dynamic) sub-path while also allowing pre-rendering to function.

> UPDATE: Indeed, it seems like we'd need a separate section to address this scenario. 

I agree, some more pointers on this subject would be really helpful. I have a few ideas on how to workaround this, but would be great to have official pointers on how to make this working in a supported way.

guardrex:
Passing (mostly **_NOTE TO SELF_**) regarding location while it's on my mind ...

As long as cross-links are present in relevant locations, devs generally find important content. The *Host and deploy* node *Overview* where the app base path is discussed is likely the best spot for a new section on this subject. We don't want a new section for the (dynamic) sub-path scenario in the *Routing* topic because it isn't a *fundamental* framework routing scenario. It would be such a scenario if we were adopting the approach on this PR to just call out a framework API behavior for routing. It looks like we'll end up with a section to address the entire hosting scenario. We can cross-link to such a new section from the routing topic and from the security node overview. Devs will probably find it. There are two ways to cross-link, and we'll leave that TBD for the moment because we aren't even sure right now if we're going to cover the hosting scenario ...

* Cross-link in the *Additional resources* section. This is generally fine as long as the link text makes the relevancy clear. Usually, the link text has to be set manually because the cross-linked topic's title is insufficient.
* Add a section. This is also a good approach at the cost of lengthening the topic. It allows us to explain concepts or the relevancy of the subject in detail.

guardrex:
**UPDATE**: We'll hear back soon on this ... probably by the end of the week.

pranavkm:
@guardrex do you think we should move the NavigationManager as a separate page? It somewhat feels orthogonal to the routing. It does make sense to talk about BaseUri during pre-rendering at the same time as talking about the property specifically.

guardrex:
That seems like a separate question from the ask at hand. It wasn't flagged on review of the ToC (Steve, Ryan, DR), but I'll set up an issue (depending if its worked separately from the ask here). Another possibility is just to change the topic's title to **_Routing and Navigation_**, and that would probably resolve it, too.

The specific ask here is how to configure an app for **_dynamic sub-app routes_**. How is it done? If it can be done, then will we cover it? If so, then we'll figure out where (e.g., near App base path in *Host and deploy*, here in a possibly re-titled doc, here with the same title, in a new Nav Manager topic that would appear right after this Routing topic in the Fundamentals ToC node)?

guardrex:
@Bouke ... We've just about reached the 🎁 *holiday break* ⛄, and staff is leaving now ... or very soon 🏃😅. We might be picking this discussion back up in early 2022. In the meantime, hit up devs on public support forums. Devs hanging out in these forums might have worked out the scenario ...

* [Stack Overflow (tag: `blazor`)](https://stackoverflow.com/questions/tagged/blazor)
* [ASP.NET Core Slack Team](http://tattoocoder.com/aspnet-slack-sign-up/)
* [Blazor Gitter](https://gitter.im/aspnet/Blazor) 

I'm moving this discussion to an issue that will be triaged on the Blazor.Docs project at https://github.com/dotnet/AspNetCore.Docs/issues/24182.

I'll rename the topic to *Routing and Navigation* for now and create a bullet point on my tracking issue to ask Dan if he'd like to see the Nav Manager content moved into its own topic to appear immediately after this topic.

It's not exactly perhaps what you're asking for, but also take a look at **_Resource Authorization_** ...

https://docs.microsoft.com/aspnet/core/blazor/security/#resource-authorization

... because that approach handles a dynamic path segment that *could be used for a tenant segment*. It's true that it would be part of every route template, but that approach permits authorization policies to act on the segment and action based on the segment value within components. Even if that is viable, we'll still consider covering this scenario via the new issue.

PR #24037, opened 2021-11-23
Update Index.cshtml.cs

cirinatorres:
I think that the warning should be restored after the class instead of being disabled again.


cirinatorres:
@Rick-Anderson @tdykstra not sure if it's right or how to add reviewers for the pull request. Thanks

Rick-Anderson:
@cirinatorres thanks for the PR. 

PR #24050, opened 2021-11-24
Remove slash

vanillajonathan:
The slash shouldn't be in the output, but I don't know if there was any reason it was there.


guardrex (approved):
You're correct, @vanillajonathan. We do that when the text is **not** code fenced to escape the `<`. In this case with the code fencing, we don't want that to render.

guardrex:
Ok ... there were two other spots in versioned coverage. I got 'um 😈 on the commit.

Thanks @vanillajonathan! 🎸

PR #24051, opened 2021-11-24
Fix method name CreateFromPemFile

totpero:

### DIFF
--- a/aspnetcore/security/authentication/certauth.md
+++ b/aspnetcore/security/authentication/certauth.md
@@ -330,7 +330,7 @@ services.AddCertificateForwarding(options =>
         if (!string.IsNullOrWhiteSpace(headerValue))
         {
             string certPem = WebUtility.UrlDecode(headerValue);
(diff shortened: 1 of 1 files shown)


serpent5:
Hi @totpero. Thanks for taking the time to create a PR.

I don't think this should be changed to `CreateFromPemFile`, but it does appear to be missing the second argument to [`CreateFromPem(ReadOnlySpan<Char>, ReadOnlySpan<Char>)`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2.createfrompem?view=net-5.0#System_Security_Cryptography_X509Certificates_X509Certificate2_CreateFromPem_System_ReadOnlySpan_System_Char__System_ReadOnlySpan_System_Char__).

@blowdart Can you confirm that this should stay as `CreateFromPem`? If so, can you provide an updated example that also provides this second argument for .NET 5?

blowdart:
No it shouldn't be changed. This isn't reading from a file at all, but rather the decoded string, so the sample is correct as is.

PR #24075, opened 2021-11-26
Remove/prune .gitignore and .gitattributes

serpent5:
.


Rick-Anderson (approved):

Rick-Anderson:
At one point I was adding .gitignore to override the root version when I wanted to add a file excluded by the root version. For example sometimes you need the `.sln` file. But that's easy to override on the command line.

PR #24100, opened 2021-11-28
Simplify code example

martincostello:
Remove the second parameter as the compiler will do that for you automatically in C# 10.


serpent5:
Thanks @martincostello.

PR #24169, opened 2021-12-02
Update js-collocation.md

rexebin:
fix:  must have `./` to work.


guardrex (approved):
Thanks @rexebin! :rocket: ... and I checked the repo. This is the only spot that requires the update.

PR #24175, opened 2021-12-02
Collections model binding "index" parameter warning

brunolins16:
Fixes dotnet/aspnetcore#28603

Preview URL: https://review.docs.microsoft.com/en-us/aspnet/core/mvc/models/model-binding?view=aspnetcore-3.1&branch=pr-en-us-24175

### DIFF
--- a/aspnetcore/mvc/models/model-binding.md
+++ b/aspnetcore/mvc/models/model-binding.md
@@ -342,7 +342,10 @@ For targets that are collections of simple types, model binding looks for matche
   [a]=1050&[b]=2000&index=a&index=b
   ```
 
-<!-- STUB: Make same change here -->
+ Avoid binding a parameter or a property named `index` or `Index` if it is adjacent to a collection value. Model binding attempts to use `index` as the index for the collection which might result in incorrect binding. For example, consider the following action:
+  
+  ```csharp
+  public IActionResult Post(string index, List<Product> products)
 * The following format is supported only in form data:
 
   ```
--- a/aspnetcore/mvc/models/model-binding.md
+++ b/aspnetcore/mvc/models/model-binding.md
@@ -342,6 +342,7 @@ For targets that are collections of simple types, model binding looks for matche
   [a]=1050&[b]=2000&index=a&index=b
   ```
 
+<!-- STUB: Make same change here -->
 * The following format is supported only in form data:
 
   ```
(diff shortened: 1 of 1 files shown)


aspnetcore/mvc/models/model-binding.md
```diff
@@ -966,6 +966,8 @@ For targets that are collections of simple types, model binding looks for matche
   ```
   [a]=1050&[b]=2000&index=a&index=b
   ```
+  > [!WARNING]
```
pranavkm, line 969: Maybe we could remove the warning and simply state it? It's not really a error case.

aspnetcore/mvc/models/model-binding.md
```diff
@@ -966,6 +966,8 @@ For targets that are collections of simple types, model binding looks for matche
   ```
   [a]=1050&[b]=2000&index=a&index=b
   ```
+  > [!WARNING] 
+  > Since the model binding looks for the keyword `index` from a Form or Query string is strongly recommended to not use it as your action's parameter name to avoid unexpected behaviors.
```
pranavkm, line 970:
```suggestion
  > Avoid binding a parameter or a property named `index` or `Index` if it is adjacent to a collection value. Model binding will attempt to use it as the index for the collection which might result in incorrect binding. For example, consider the following action:
  
  ```csharp
  public IActionResult Post(string index, List<Product> products)
 ```
The `index` parameter will bound to the parameter and will also be used to bind the `product` collection which might not be what is desired. Renaming the parameter or using a model binding attribute to configure the binding would avoid this issue.

**Review by pranavkm — COMMENTED** [review]
@Rick-Anderson could you help phrase this better? Thanks!

File: aspnetcore/mvc/models/model-binding.md (hunk 1 above)
Comment (brunolins16) [reply aspnetcore/mvc/models/model-binding.md:969]: @pranavkm, we could if you prefer but I think that effect of using the index parameter cause enough disturbance, since the list will be populated with an item that could potentially break the user's code.

Comment (Rick-Anderson) [reply aspnetcore/mvc/models/model-binding.md]: Can you check [my commit](https://github.com/dotnet/AspNetCore.Docs/pull/24175/commits/86fed85d8b11e542770b09466ae74d0d681f970c)?

Comment (brunolins16) [reply aspnetcore/mvc/models/model-binding.md]: I like that, it is pretty clear. Thanks both for the help

File: aspnetcore/mvc/models/model-binding.md [hunk 3]
```diff
@@ -342,6 +342,7 @@ For targets that are collections of simple types, model binding looks for matche
   [a]=1050&[b]=2000&index=a&index=b
   ```
 
+<!-- STUB: Make same change here -->
```
Rick-Anderson, line 345:
```suggestion
 Avoid binding a parameter or a property named `index` or `Index` if it is adjacent to a collection value. Model binding attempts to use `index` as the index for the collection which might result in incorrect binding. For example, consider the following action:
  
  ```csharp
  public IActionResult Post(string index, List<Product> products)
  ```
  In the preceding code, the `index` query string parameter binds to the `index` method parameter and also is used to bind the product collection. Renaming the `index` parameter or using a model binding attribute to configure binding avoids this issue:
  ```csharp
  public IActionResult Post(string productIndex, List<Product> products)
  ```
```

Rick-Anderson (approved):

aspnetcore/mvc/models/model-binding.md
```diff
@@ -342,6 +342,10 @@ For targets that are collections of simple types, model binding looks for matche
   [a]=1050&[b]=2000&index=a&index=b
   ```
 
+ Avoid binding a parameter or a property named `index` or `Index` if it is adjacent to a collection value. Model binding attempts to use `index` as the index for the collection which might result in incorrect binding. For example, consider the following action:
+  
+  ```csharp
+  public IActionResult Post(string index, List<Product> products)
```
Rick-Anderson, line 348:
```suggestion
  Avoid binding a parameter or a property named `index` or `Index` if it is adjacent to a collection value. Model binding attempts to use `index` as the index for the collection which might result in incorrect binding. For example, consider the following action:
  
  ```csharp
  public IActionResult Post(string index, List<Product> products)
  ```
  In the preceding code, the `index` query string parameter binds to the `index` method parameter and also is used to bind the product collection. Renaming the `index` parameter or using a model binding attribute to configure binding avoids this issue:
  ```csharp
  public IActionResult Post(string productIndex, List<Product> products)
  ```
```

PR #24192, opened 2021-12-03
Anchor: What if asp-page does not exist

yecril71pl:
Explain that if the [page referenced by an anchor](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.taghelpers.anchortaghelper.page) does not exist, a link to the main page will be generated.  No warning will be produced.


guardrex (approved):
Thanks @yecril71pl! :rocket:

PR #24199, opened 2021-12-06
Virtualization doc updates (keyboard support, guidance on correct use)

SteveSandersonMS:
Fixes https://github.com/dotnet/aspnetcore/issues/29123 and https://github.com/dotnet/aspnetcore/issues/38020

### DIFF
--- a/aspnetcore/blazor/components/virtualization.md
+++ b/aspnetcore/blazor/components/virtualization.md
@@ -175,6 +175,65 @@ By default, the `Virtualize` component measures the rendering size (height) of i
 
 When making changes to items rendered by the `Virtualize` component, call <xref:Microsoft.AspNetCore.Components.ComponentBase.StateHasChanged%2A> to force re-evaluation and rerendering of the component. For more information, see <xref:blazor/components/rendering>.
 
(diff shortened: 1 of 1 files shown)


guardrex (approved):

aspnetcore/blazor/components/virtualization.md
```diff
@@ -175,6 +175,65 @@ By default, the `Virtualize` component measures the rendering size (height) of i
...
+
+```html
+<div style="height:500px; overflow-y:scroll" tabindex="-1">
+    <div style="height:1100px"><!-- Placeholder --></div>
+    <div class="flight-info">Flight 12</div>
+    <div class="flight-info">Flight 13</div>
+    <div class="flight-info">Flight 14</div>
+    <div class="flight-info">Flight 15</div>
+    <div class="flight-info">Flight 16</div>
+    <div style="height:3400px"><!-- Placeholder --></div>
```
Liero, line 218: The name "Placeholder" can be confusing, because it implies that it will use Placeholder template.
I think the correct name is "Spacer", which is how it is called in source code.

PR #24214, opened 2021-12-07
Check for existing PRs before merge-to-live

serpent5:
When the merge-to-live workflow runs, we don't want it to open a new PR if one already exists. This change fails the workflow if a there's an open PR that:

- Targets the `live` branch.
- Is authored by `github-actions`.

This isn't likely to happen, as we won't leave these PRs open for long, but I think it's a worthy addition to protect us against multiple PRs.

### DIFF
--- a/.github/workflows/merge-live.yml
+++ b/.github/workflows/merge-live.yml
@@ -9,6 +9,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - run: |
+          numOpenPullRequests="$(gh pr list --author app/github-actions --search "base:live" --json id --jq ". | length")"
+
+          if [ "$numOpenPullRequests" -ne "0" ]; then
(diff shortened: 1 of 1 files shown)


Rick-Anderson (approved):

Rick-Anderson:
> When the merge-to-live workflow runs, we don't want it to open a new PR if one already exists

I don't think you need special code for that, it's automatic. If I create a PR merge main to live, then try to create another one, it opens up the first opened merge to live PR.

serpent5:
> I don't think you need special code for that, it's automatic.

Ooh, interesting. I'll give it a try and see if that's the case with the auto-creation. Thanks for the pointer.

serpent5:
Yep, it's automatic indeed. Brilliant, thanks.

It writes e helpful error and fails the build, just as I was trying to do in this PR. Here's the failed workflow: https://github.com/dotnet/AspNetCore.Docs/runs/4456055896?check_suite_focus=true.

PR #24236, opened 2021-12-08
Update razor.md

Rick-Anderson:


serpent5:
I'll hold off on this until we find out if there's a better link, as discussed in the linked PR. Although the link here contains some good information, it might be a bit confusing for ASP.NET Core users.

Rick-Anderson:
> I'll hold off on this until we find out if there's a better link, as discussed in the linked PR. Although the link here contains some good information, it might be a bit confusing for ASP.NET Core users.

It's the best link I've found by far and I've found it very helpful. I've updated the text, let me know if that works.

serpent5:
> let me know if that works.

Yes, I think that helps. Now it should be less confusing when the reader sees mention of ASP.NET Web Pages 2/3 and e.g. `Server.MapPath`.

PR #24245, opened 2021-12-09
Update docker-https.md

igblip:
I'm still a bit new to docker, but I think the -p 8001:443 maps external port 8001 to internal port 443, so the internal environment variable setting the HTTPS port to 8001 is incorrect.


serpent5:
Hi @igblip. Thanks for taking the time to open a PR.

In this case, the example is correct. The `ASPNETCORE_HTTPS_PORT` is used, for example, when the server wants to redirect the caller to use the HTTPS port. If the caller requests the HTTP port 8080, the server responds with a redirect to the HTTPS port 8081 - the accessible, external port, rather than 443 - the internal, inaccessible port.

PR #24264, opened 2021-12-10
Document new option behavior for hosting bundle

wtgodbe:


aspnetcore/host-and-deploy/iis/hosting-bundle.md
```diff
@@ -59,6 +59,7 @@ To obtain an earlier version of the installer:
 
 > [!NOTE]
 > For information on IIS Shared Configuration, see [ASP.NET Core Module with IIS Shared Configuration](xref:host-and-deploy/aspnet-core-module#aspnet-core-module-with-an-iis-shared-configuration).
+> When running the Hosting Bundle installer with options set, the value for each option will be saved in the registry. Subsequent installs from the same Major.Minor version band will use the same options, unless another set of options is explicitly passed from the command line. If the first install of the hosting bundle has no options passed, each option will get a default value of `0` written in to the registry.
```
jamshedd, line 62: This looks good. Could we also articulate what writing a default value of 0 implies (option command line option is treated as on or off)?

wtgodbe: Updated

jamshedd (approved):

aspnetcore/host-and-deploy/iis/hosting-bundle.md
```diff
@@ -59,6 +59,7 @@ To obtain an earlier version of the installer:
 
 > [!NOTE]
 > For information on IIS Shared Configuration, see [ASP.NET Core Module with IIS Shared Configuration](xref:host-and-deploy/aspnet-core-module#aspnet-core-module-with-an-iis-shared-configuration).
+> When running the Hosting Bundle installer with options set, the value for each option will be saved in the registry. Subsequent installs from the same Major.Minor version band will use the same options, unless another set of options is explicitly passed from the command line. If the first install of the hosting bundle has no options passed, each option will get a default value of `0` written in to the registry (a value of `0` implies that the option is off, meaning the user is not opting out of the given component).
```
Rick-Anderson, line 62:
```suggestion
> When running the Hosting Bundle installer with options set, the value for each option is saved in the registry. Subsequent installs from the same Major.Minor version band use the same options, unless another set of options is explicitly passed from the command line. If the first install of the hosting bundle has no options passed, each option gets a default value of `0` written in to the registry. A value of `0` implies that the option is off, meaning the user is not opting out of the given component.
```

Rick-Anderson (approved):
Per style guide, Avoid future case and parenthetical expressions.

aspnetcore/host-and-deploy/iis/hosting-bundle.md
```diff
@@ -59,6 +59,7 @@ To obtain an earlier version of the installer:
 
 > [!NOTE]
 > For information on IIS Shared Configuration, see [ASP.NET Core Module with IIS Shared Configuration](xref:host-and-deploy/aspnet-core-module#aspnet-core-module-with-an-iis-shared-configuration).
+> When running the Hosting Bundle installer with options set, the value for each option is saved in the registry. Subsequent installs from the same Major.Minor version band use the same options, unless another set of options is explicitly passed from the command line. If the first install of the hosting bundle has no options passed, each option gets a default value of `0` written in to the registry. A value of `0` implies that the option is off, meaning the user is not opting out of the given component.
```
halter73, line 62: Should this share the same `[!NOTE]` as the IIS Shared Configuration stuff?

Also, it seems like the fragment that `[ASP.NET Core Module with IIS Shared Configuration](xref:host-and-deploy/aspnet-core-module#aspnet-core-module-with-an-iis-shared-configuration)` used to point to is now at `host-and-deploy/iis/advanced#aspnet-core-module-with-an-iis-shared-configuration` which is linked to from the `OPT_NO_SHARED_CONFIG_CHECK=1` bullet point, so I'm not sure that the above note still makes sense.

dougbu: Question on
> When running the Hosting Bundle installer with options set, &hellip;

Aren't settings (`0`s) saved the first time the installer runs _regardless_ of whether overrides are specified❔

wtgodbe: Moved the info to a separate note.

> Aren't settings (0s) saved the first time the installer runs regardless of whether overrides are specified❔

Yes, the first time, the options will always be saved. On subsequent installs, options are only saved if they are set.

dougbu: > Yes, the first time, the options will always be saved. On subsequent installs, options are only saved if they are set.

This isn't clear from "When running the Hosting Bundle installer with options set, &hellip;". But, it's not worth a follow-up PR.

Issue #23688, opened 2021-10-29
.NET 6 update: prefer new Use middleware extension method


Rick-Anderson:
Add a new `app.Use` overload that requires users to pass the `HttpContext` to the `next` function which will save 2 per-request allocations over the previous extension method.

## Customer Impact

Less per-request allocations when using simple middleware.

See https://github.com/dotnet/aspnetcore/pull/31784

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 9898482c-fa54-7ce2-1716-d14fbe03990d
* Version Independent ID: 3926c686-0ccc-f9fe-ca52-0134612a1623
* Content: [ASP.NET Core Middleware](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/?view=aspnetcore-6.0)
* Content Source: [aspnetcore/fundamentals/middleware/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/middleware/index.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-fundamentals**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
@fiyazbinhasan  or @smandia would one of you be able to write a sample app showing the new `app.Use` overload?

fiyazbinhasan:
@Rick-Anderson sure. Will take a look at this soon 🙂

PR #24279, opened 2021-12-12
sample snapshot fort the new app.Use overload

fiyazbinhasan:
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/23688


fiyazbinhasan:
@Rick-Anderson I've added a snapshot file showing the usage of the `Use` overloads and also added some comments which I think could be mentioned in the documentation. Please check 😺

serpent5 (changes requested):
Thanks for this work, @fiyazbinhasan. I can see you've been a huge help here with many great contributions to the docs. ⭐ 

I have a couple of change requests for this one:

- Can we reduce this to just the second snippet, `snippet2`? I don't think we'll link to the first and third snippets in the article, so they're not needed here.
- Can we remove the `CultureInfo` context here, to simplify the example? In keeping with the sample code we have for some of the other examples, could we have something like this?:

    ```csharp
    {
        // Do work that doesn't write to the Response.
        await next.Invoke(context);
        // Do logging or other work that doesn't write to the Response.
    });
    ```

---

Apologies for the triple-post. It looks like I heard a network issue and the retry-logic was quite excited to take over. 😄

serpent5 (changes requested):
[removed duplicate post]

serpent5 (changes requested):
[removed duplicate post]

fiyazbinhasan:
Hi @serpent5. thanks for the review 😸! I've reduced the code snippets to just show the usage of `next(context)`.

Although I do think that this [section](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/?view=aspnetcore-6.0#prefer-appuse-overload-that-requires-passing-the-context-to-next) in the doc need some information like this, (when not using `next` at all)

```
/*
    1. next is not used i.e. it is a terminal middleware
    2. Every other middleware after a terminal middleware is skipped
    1. Use should be replaced with Run or it will throw a compile time error (CS0121: ambiguous)

    app.Use(async (context, next) => {
        await context.Response.WriteAsync(
            $"Hello {CultureInfo.CurrentCulture.DisplayName}");
    });
*/

app.Run(async (context) =>
{
    await context.Response.WriteAsync(
        $"Hello {CultureInfo.CurrentCulture.DisplayName}");
});
```

What do you think?

serpent5:
Thanks @fiyazbinhasan. That was a very quick turnaround. 🏃

I think we should just start by documenting this in the context of `next` being invoked, as the code snippet you've provided shows. If we receive customer reports around *not* calling `next`, that would be the time to add the extra information around using `Run`, etc. We can leave your helpful comments in the snippet, so we can surface it later if we need to.

@Rick-Anderson Let me know if you disagree with the above. We still need to display this sample in the article. Would you like to add anything else in light of this comment from the article?:

> TODO, a minimum, provide a sample usage. Better, use this overload in the 6.0 version

serpent5 (approved):
Ok, let's get this merged and then follow up with another PR, as @Rick-Anderson suggests.

Thanks again @fiyazbinhasan. I'll pick up your other PR tomorrow.

PR #24311, opened 2021-12-14
Update index.md

sanchit154:
The output reference of the shown code is wrong.


guardrex (approved):
Thanks @sanchit154! :rocket:

btw -- It's a string in code, so I code-fence it (and what's in the table) so that it doesn't get automatically localized in the doc.

PR #24329, opened 2021-12-16
update no loc

xh286286:
[Internal Preview URL](https://review.docs.microsoft.com/en-us/aspnet/core/tutorials/first-mvc-app/adding-view?view=aspnetcore-6.0&branch=pr-en-us-24329&tabs=visual-studio#change-the-title-footer-and-menu-link-in-the-layout-file).

### DIFF
--- a/aspnetcore/tutorials/first-mvc-app/adding-view.md
+++ b/aspnetcore/tutorials/first-mvc-app/adding-view.md
@@ -121,7 +121,7 @@ In the preceding markup, the `asp-area=""` [anchor Tag Helper attribute](xref:mv
 
 **Note**: The `Movies` controller hasn't been implemented. At this point, the `Movie App` link isn't functional.
 
-Save the changes and select the **Privacy** link. Notice how the title on the browser tab displays **Privacy Policy - Movie App** instead of **Privacy Policy - MvcMovie**:
+Save the changes and select the **Privacy** link. Notice how the title on the browser tab displays :::no-loc text=" **Privacy Policy - Movie App** " ::: instead of :::no-loc text=" **Privacy Policy - MvcMovie** : " :::
 
 ![Privacy tab](~/tutorials/first-mvc-app/adding-view/_static/privacy50.png)
 
@@ -339,7 +339,7 @@ In the preceding markup, the `asp-area=""` [anchor Tag Helper attribute](xref:mv
 
 **Note**: The `Movies` controller hasn't been implemented. At this point, the `Movie App` link isn't functional.
 
-Save the changes and select the **Privacy** link. Notice how the title on the browser tab displays **Privacy Policy - Movie App** instead of **Privacy Policy - Mvc Movie**:
+Save the changes and select the **Privacy** link. Notice how the title on the browser tab displays :::no-loc text=" **Privacy Policy - Movie App** " ::: instead of :::no-loc text=" **Privacy Policy - MvcMovie** : " :::
 
 ![Privacy tab](~/tutorials/first-mvc-app/adding-view/_static/privacy50.png)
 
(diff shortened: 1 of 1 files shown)


guardrex:
@Rick-Anderson @serpent5 @wadepickett ... This is an interesting approach that we've never used. What I've been doing is code-fencing with bold.

It's been challenging for IDE UI tho: I noticed in some testing a couple of years ago that VS localizes some of the UI but not all of it. Often, the only way to know what shouldn't be localized was to change the language in the IDE and see what it shows. Given the workload, it would be challenging to check everything that way.

Anyway, do you prefer the `{TEXT} :::no-loc text=" {TEXT} " ::: {TEXT}` approach? If so, why are there double spaces (one space outside of the `no-loc` on each end of the `:::` with another space inside of each quote. It seems like the space inside the quotes could be dropped.

serpent5:
> do you prefer the `{TEXT} :::no-loc text=" {TEXT} " ::: {TEXT}` approach?

If both approaches work out the same way, I prefer the `**` approach, just because it's less to type and easier to to not mess it up. I don't understand this all that well, though, but that's my naïve take on things.

> If so, why are there double spaces

The examples in the contributor guide don't have these extra spaces: https://docs.microsoft.com/en-gb/contribute/markdown-reference#non-localized-strings. The internal preview of this change also looks broken, so that might be connected.

guardrex:
Yes ... and the closing colons should be up against the last quote.

That makes sense to me. It is quicker 🏃 to produce and probably quicker 🏃 to maintain. Compare for the server app of a hosted WASM solution ...

* \*\*\`Server\`\*\*
* :::no-loc text="Server":::

Still tho, the IDE problem persists. I don't think we can just go willy nilly along code-fencing *everything* because, for example, the VS IDE **_does_** localize its UI in a lot of spots. The only way to know for sure is to look, and that's going to take a lot of time 💰.

I'll go with whatever you 🐈🐈🐈 decide. 👂

Rick-Anderson:
If it's a string repeated multiple times in one file like this, tack it onto the end of the no-loc meta data. As long as it's at the end, that doesn't impact global meta data updates.
Otherwise Let's stick with code fencing. 

***In this case, it doesn't matter if it's localized.*** It's ok to localize this string, it's just being displayed.

aspnetcore/tutorials/first-mvc-app/adding-view.md
```diff
@@ -121,7 +121,7 @@ In the preceding markup, the `asp-area=""` [anchor Tag Helper attribute](xref:mv
 
 **Note**: The `Movies` controller hasn't been implemented. At this point, the `Movie App` link isn't functional.
 
-Save the changes and select the **Privacy** link. Notice how the title on the browser tab displays **Privacy Policy - Movie App** instead of **Privacy Policy - MvcMovie**:
+Save the changes and select the **Privacy** link. Notice how the title on the browser tab displays :::no-loc text=" **Privacy Policy - Movie App** " ::: instead of :::no-loc text=" **Privacy Policy - MvcMovie** : " :::
```
Rick-Anderson, line 124:
```suggestion
Save the changes and select the **Privacy** link. Notice how the title on the browser tab displays `**Privacy Policy - Movie App**` instead of `**Privacy Policy - MvcMovie**`
```

aspnetcore/tutorials/first-mvc-app/adding-view.md
```diff
@@ -339,7 +339,7 @@ In the preceding markup, the `asp-area=""` [anchor Tag Helper attribute](xref:mv
 
 **Note**: The `Movies` controller hasn't been implemented. At this point, the `Movie App` link isn't functional.
 
-Save the changes and select the **Privacy** link. Notice how the title on the browser tab displays **Privacy Policy - Movie App** instead of **Privacy Policy - Mvc Movie**:
+Save the changes and select the **Privacy** link. Notice how the title on the browser tab displays :::no-loc text=" **Privacy Policy - Movie App** " ::: instead of :::no-loc text=" **Privacy Policy - MvcMovie** : " :::
```
Rick-Anderson, line 342:
```suggestion
Save the changes and select the **Privacy** link. Notice how the title on the browser tab displays `**Privacy Policy - Movie App**` instead of `**Privacy Policy - MvcMovie**`
```

Rick-Anderson (changes requested):
Commit these suggestions.

PR #24357, opened 2021-12-17
Update merge-live to ensure a PR is always available

serpent5:
Instead of running merge-live on a schedule, we could have it open a PR anytime we push to `main`. We can check first to make sure one doesn't exist, so that we don't end up with loads of failed workflow runs.

cc @guardrex

### DIFF
--- a/.github/workflows/merge-live.yml
+++ b/.github/workflows/merge-live.yml
@@ -1,16 +1,24 @@
 name: Merge to Live
+on:
+  push:
+    branches:
+      - main
 permissions:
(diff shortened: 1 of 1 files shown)


Rick-Anderson (approved):
Fantastic.

serpent5:
> How often does it check.

Every time we push a commit to main. We'll still need to manually complete the merge as we've been doing, so we'll still control when the merge completes. Automating that side of things is a job for next year, but we're close to that now.

PR #24390, opened 2021-12-21
Fix a not updated description

Tony-Liou:
Fixes the words to match the example code above this description.


guardrex (approved):
Thanks @Tony-Liou! :rocket:

PR #24393, opened 2021-12-21
Blazor Debugging Launch.json `url` property documentation

TanayParikh:
Fixes: https://github.com/dotnet/aspnetcore/issues/38382

### DIFF
--- a/aspnetcore/blazor/debug.md
+++ b/aspnetcore/blazor/debug.md
@@ -221,7 +221,7 @@ The following launch configuration options are supported for the `blazorwasm` de
 | Option    | Description |
 | --------- | ----------- |
 | `request` | Use `launch` to launch and attach a debugging session to a Blazor WebAssembly app or `attach` to attach a debugging session to an already-running app. |
-| `url`     | The URL to open in the browser when debugging. |
+| `url`     | The URL to open in the browser when debugging. Defaults to `https://localhost:5001`. Note, if your app is running at a different URL an `about:blank` will launch within the browser. |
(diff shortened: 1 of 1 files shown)


guardrex (approved):
@TanayParikh ... Check my grammar/style update to confirm that I didn't shift the meaning.

guardrex:
btw @TanayParikh ... This is confusing. Why is this configurable? If the dev changes it, it **_doesn't_** launch at the right page/tab? ... then are they supposed to navigate to what they set this to manually?

guardrex:
Ok ... maybe I get it. 😄 ... It's just saying that if this doesn't match where the app is configured to load (URL) then they'll get the `about:blank`. They need to match for it to open and debug at the right URL. 🤔

TanayParikh:
> btw @TanayParikh ... This is confusing. Why is this configurable? If the dev changes it, it **_doesn't_** launch at the right page/tab? ... then are they supposed to navigate to what they set this to manually by typing the URL they set for `url` into the browser's address bar?

This `url` is used to launch the browser. The default is localhost:5001. If there's nothing at `localhost:5001` then the browser opens the `about:blank` page. This can occur if the user changed the `Properties/launchSettings.json` to have a different application url. This behavior is documented in the `blazor/tooling` page. 

> Examine the Properties/launchSettings.json file and determine the URL of the app from the applicationUrl property (for example, https://localhost:7268). Note this value for use in the launch.json file.
>
> In the launch configuration of the .vscode/launch.json file:
>
> Set the current working directory (cwd) to the Server project folder.
> Indicate the app's URL with the url property. Use the value recorded earlier from the Properties/launchSettings.json file.

PR #24417, opened 2021-12-26
Apache configuration process update on linux

viremman:
I felt like there is one or two things that can be added to the documentation which i did add.


guardrex (approved):
Thanks @viremman! :rocket:

guardrex:
Sorry for the delay, @viremman. We're having a little trouble with the build system at the moment. No worries tho ... it should clear up in a bit. We'll get this merged as soon as we can. Thanks again for the submission.

Issue #24434, opened 2021-12-28
review and clean up write middleware


Rick-Anderson:
Address comments in  #24348

@fiyazbinhasan can you take this? 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 13e2d968-17a6-1a50-53cd-1d6b9ad0276d
* Version Independent ID: cd22a135-4446-79f5-89db-b741bd44a720
* Content: [Write custom ASP.NET Core middleware](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/write?view=aspnetcore-6.0)
* Content Source: [aspnetcore/fundamentals/middleware/write.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/middleware/write.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-fundamentals**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

fiyazbinhasan:
Taking a look :)

Rick-Anderson:
```
::: moniker range=">= aspnetcore-6.0"

By Fiyaz Hasan, Rick Anderson and Steve Smith
```

Rick-Anderson:
That should be
By Fiyaz Hasan, Rick Anderson***,*** and Steve Smith

fiyazbinhasan:
Just noticed you added my name as well! 😃

PR #24438, opened 2021-12-28
review and clean up write middleware

fiyazbinhasan:
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/24434

@Rick-Anderson, did some rearrangements. I think it will better explain to the user which `Use` overload should be used for performance.

please add your suggestions also :)


Rick-Anderson:
@fiyazbinhasan  much better. Thanks.

Issue #12157, opened 2019-04-25
How to log automatic 400 responses on model validation errors


MNF:
“The ApiController attribute makes model validation errors automatically trigger an HTTP 400 response.”
It is not described, does details of 400 errors logged or not? 
If not, is it possible to configure error details to be logged?
Or I have to SuppressModelStateInvalidFilter = true; and do logging by myself in the block
if (!ModelState.IsValid)
{
     Log (ModelState);
    return BadRequest(ModelState);
}

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b391017f-25e1-d5e1-c4f7-020408ade7b5
* Version Independent ID: 263a7b64-2e42-bbfc-c1d3-61680438e1e4
* Content: [Create web APIs with ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/web-api/?view=aspnetcore-2.1#automatic-http-400-responses)
* Content Source: [aspnetcore/web-api/index.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/web-api/index.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-webapi**
* GitHub Login: @scottaddie
* Microsoft Alias: **scaddie**

serpent5:
There's a section in the [docs](https://docs.microsoft.com/en-us/aspnet/core/web-api/index?view=aspnetcore-2.2#customize-badrequest-response) that explains how to customise the BadRequest response when using the built-in filter. You can set the [`InvalidModelStateResponseFactory`](https://docs.microsoft.com/en-gb/dotnet/api/microsoft.aspnetcore.mvc.apibehavioroptions.invalidmodelstateresponsefactory?view=aspnetcore-2.2#Microsoft_AspNetCore_Mvc_ApiBehaviorOptions_InvalidModelStateResponseFactory) to a custom function that first performs the logging and then returns an appropriate `BadRequestObjectResult`.

@Rick-Anderson I can provide an example of what that would look like here if it would be helpful. Otherwise, is this something that could be taken over to Stack Overflow with a link. What's the best approach for handling something like that?

Rick-Anderson:
@serpent5 an example in this issue would be great.  We can add a link to this issue in the doc and close the issue.

serpent5:
There are a few ways to approach this, but I'll list a couple of ideas to get you started:

[...]

PR #24463, opened 2022-01-02
Add automatic 400 response logging code samples

SpiritBob:
Add automatic 400 response logging code samples, instead of pointing the end-user to our issue tracking website, as proposed in https://github.com/dotnet/AspNetCore.Docs/issues/12157

[Internal Preview URL](https://review.docs.microsoft.com/en-us/aspnet/core/web-api/?view=aspnetcore-6.0&branch=pr-en-us-24463).

### DIFF
--- a/aspnetcore/web-api/index.md
+++ b/aspnetcore/web-api/index.md
@@ -173,33 +173,7 @@ By default, [InvalidModelStateResponseFactory](https://docs.microsoft.com/dotnet
 
 Here is an example of code to be used in `Startup.ConfigureServices` that adds logging and preserves the original behavior:
 
-```csharp
-services.AddControllers().ConfigureApiBehaviorOptions(options =>
-{
-    // If preserving of the original behavior is desired, get a reference to the delegate.
-    var builtInFactory = options.InvalidModelStateResponseFactory;
-
-    options.InvalidModelStateResponseFactory = context =>
-    {
-        // As an example, we will get an instance of ILogger with the category "Startup".
-        var logger = context.HttpContext.RequestServices.GetRequiredService<ILogger<Startup>>();
-
-        // Log accordingly.
-
-        // By using the original delegate, we preserve the default behavior.
-        // Alternatively, you can take full control and simply construct the ValidationProblemDetails object yourself,
-        // or even use a custom object.
-        var result = builtInFactory(context);
-
-        // If accessing the returned ValidationProblemDetails object is required, get a reference to it.
-        var problemDetails = (ValidationProblemDetails)((ObjectResult)result).Value;
-
-        // Modify & Log accordingly.
-
-        return result;
-    };
-});
-```
+[!code-csharp[](index/samples/3.x/Startup.cs?name=snippet_AutomaticBadRequestLogging)]
 
 ### Disable automatic 400 response
 
--- a/aspnetcore/web-api/index.md
+++ b/aspnetcore/web-api/index.md
@@ -167,7 +167,64 @@ To make automatic and custom responses consistent, call the <xref:Microsoft.AspN
 
 ### Log automatic 400 responses
 
-See [How to log automatic 400 responses on model validation errors (dotnet/AspNetCore.Docs#12157)](https://github.com/dotnet/AspNetCore.Docs/issues/12157).
+You can add logging by utilizing [InvalidModelStateResponseFactory](https://docs.microsoft.com/dotnet/api/microsoft.aspnetcore.mvc.apibehavioroptions.invalidmodelstateresponsefactory). By default, [InvalidModelStateResponseFactory](https://docs.microsoft.com/dotnet/api/microsoft.aspnetcore.mvc.apibehavioroptions.invalidmodelstateresponsefactory) uses [ProblemDetailsFactory](https://docs.microsoft.com/dotnet/api/microsoft.aspnetcore.mvc.infrastructure.problemdetailsfactory) to construct an instance of [ValidationProblemDetails](https://docs.microsoft.com/dotnet/api/microsoft.aspnetcore.mvc.validationproblemdetails). Here is an example of code to be used in `Startup.ConfigureServices` that adds logging and preserves the original behavior:
+
+```csharp
+services.AddControllers().ConfigureApiBehaviorOptions(options =>
+{
+    // If preserving of the original behavior is desired, get a reference to the delegate.
+    var builtInFactory = options.InvalidModelStateResponseFactory;
+
+    options.InvalidModelStateResponseFactory = context =>
+    {
+        // As an example, we will create a logger with the full name of the action method as the category.
+        var logger = GetLoggerInstance(context.HttpContext.RequestServices, context.ActionDescriptor.DisplayName);
+
+        // Log accordingly.
+
+        // By using the original delegate, we preserve the default behavior.
+        // Alternatively, you can take full control and simply construct the ValidationProblemDetails object yourself,
+        // or even use a custom object.
+        var result = builtInFactory(context);
+
+        // Example of manually constructing a ValidationProblemDetails object:
+        // var problemDetails = new ValidationProblemDetails(context.ModelState)
+        //    {
+        //        Type = "https://tools.ietf.org/html/rfc7231#section-6.5.1",
+        //        Title = "One or more model validation errors occurred.",
+        //        Status = StatusCodes.Status400BadRequest,
+        //        Detail = "See the errors property for details.",
+        //        Instance = context.HttpContext.Request.Path,
+        //        Extensions =
+        //        {
+        //            ["traceId"] = Activity.Current?.Id ?? context.HttpContext?.TraceIdentifier
+        //        }
+        //    };
+
+        // If accessing the returned ValidationProblemDetails object is required, get a reference to it.
+        var problemDetails = (ValidationProblemDetails)((ObjectResult)result).Value;
+
+        // Modify & Log accordingly.
+
+        return result;
+
+        static ILogger GetLoggerInstance(IServiceProvider provider, string? category)
+        {
+            if (category is null)
+            {
+                // Get an instance of ILogger with the category "Startup".
+                return provider.GetRequiredService<ILogger<Startup>>();
+            }
+            else
+            {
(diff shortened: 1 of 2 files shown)


aspnetcore/web-api/index.md
```diff
@@ -167,7 +167,64 @@ To make automatic and custom responses consistent, call the <xref:Microsoft.AspN
...
+        var logger = GetLoggerInstance(context.HttpContext.RequestServices, context.ActionDescriptor.DisplayName);
+
+        // Log accordingly.
+
+        // By using the original delegate, we preserve the default behavior.
+        // Alternatively, you can take full control and simply construct the ValidationProblemDetails object yourself,
+        // or even use a custom object.
+        var result = builtInFactory(context);
+
+        // Example of manually constructing a ValidationProblemDetails object:
```
SpiritBob, line 190: Is this example of manually constructing a ValidationProblemDetails too much?

aspnetcore/web-api/index.md
```diff
@@ -167,7 +167,64 @@ To make automatic and custom responses consistent, call the <xref:Microsoft.AspN
...
+        // If accessing the returned ValidationProblemDetails object is required, get a reference to it.
+        var problemDetails = (ValidationProblemDetails)((ObjectResult)result).Value;
+
+        // Modify & Log accordingly.
+
+        return result;
+
+        static ILogger GetLoggerInstance(IServiceProvider provider, string? category)
+        {
+            if (category is null)
```
SpiritBob, line 213: Should we use `string.IsNullOrWhitespace` instead here?

aspnetcore/web-api/index.md
```diff
@@ -453,7 +510,43 @@ To make automatic and custom responses consistent, call the <xref:Microsoft.AspN
...
+
+    options.InvalidModelStateResponseFactory = context =>
+    {
+        // As an example, we will create a logger with the full name of the action method as the category.
+        var logger = GetLoggerInstance(context.HttpContext.RequestServices, context.ActionDescriptor.DisplayName);
+
+        // Log accordingly.
+
+        // Use existing functionality - either one defined by us, or in this case the framework's.
+        return builtInFactory(context);
```
SpiritBob, line 531: Should we try and expand this code sample to also show how one can edit the returned object? It seems on ASP Net core 2.2 the object returned is [SerializableError](https://docs.microsoft.com/dotnet/api/microsoft.aspnetcore.mvc.serializableerror)

aspnetcore/web-api/index.md
```diff
@@ -453,7 +510,43 @@ To make automatic and custom responses consistent, call the <xref:Microsoft.AspN
 
 ### Log automatic 400 responses
 
-See [How to log automatic 400 responses on model validation errors (dotnet/AspNetCore.Docs#12157)](https://github.com/dotnet/AspNetCore.Docs/issues/12157).
+You can add logging by utilizing [InvalidModelStateResponseFactory](https://docs.microsoft.com/dotnet/api/microsoft.aspnetcore.mvc.apibehavioroptions.invalidmodelstateresponsefactory). Here is an example of code to be used in `Startup.ConfigureServices` that adds logging on top of **any** already existing functionality that generates the response:
```
SpiritBob, line 513: Should we give a bit more details about the default framework implementation here (that it returns for example `SerializableError` or anything else note-worthy?) I have no idea if other ASP.NET Core versions after 2.2 and prior to 3.0 change this behavior.

SpiritBob:
@serpent5 Could you review this when you have the time? 

Happy New Year! 🎈

aspnetcore/web-api/index.md
serpent5: To simplify this, I suggest we just get an `ILogger<Startup>` and leave out the `ActionDescriptor.DisplayName` approach entirely. Although I do like this approach, I think it just ends up making for a bigger code example than we need. I'd like to just provide the building blocks in this example, which allows the reader to perform their own customisations accordingly.

aspnetcore/web-api/index.md
serpent5: As 2.2 is [no longer supported](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core), we don't need to make changes to this version in the article. You can just revert this version of the section entirely, keeping only the first version that's for 3.0+.

serpent5 (commented):
Thanks for this, @SpiritBob. This look like a good start to me. I've answered your questions as best as I can, so let me know what you think.

Once we've worked through the comments we already have, we should put the code in the [sample project](https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/web-api/index/samples/3.x) so that we can make sure it builds, etc, which also makes it easier to update in the future.

aspnetcore/web-api/index.md
MohammadAsadi1372: https://github.com/MohammadAsadi1372/MohammadAsadi.MasterService

aspnetcore/web-api/index.md
MohammadAsadi1372: https://github.com/MohammadAsadi1372/MohammadAsadi.MasterService

aspnetcore/web-api/index.md
MohammadAsadi1372: https://github.com/MohammadAsadi1372/MohammadAsadi.MasterService

MohammadAsadi1372:
https://github.com/MohammadAsadi1372/MohammadAsadi.MasterService

aspnetcore/web-api/index.md
SpiritBob: I feel like we should avoid linking to the issue, hence why we're giving every bit of information the end-user might require for automatic logging of 400 requests in the document itself, however I'm fine with completely dropping the example for instancing a `ValidationProblemDetails` object. Alternatively, we can find it's place outside of the code example, instead of linking to the issue?

serpent5: Ok, let's just drop it for now then. I'm not sure where else it would live right now, so perhaps it's something we could come back to in the future.

SpiritBob:
> Thanks for this, @SpiritBob. This look like a good start to me. I've answered your questions as best as I can, so let me know what you think.
> 
> Once we've worked through the comments we already have, we should put the code in the [sample project](https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/web-api/index/samples/3.x) so that we can make sure it builds, etc, which also makes it easier to update in the future.

I've added the code sample and did the edits we discussed. Let me know if everything's ready for review!

serpent5:
Thanks @SpiritBob. This looks really good. Let's mark it as ready-for-review, then I'll add a commit on top with a few stylistic changes, etc, which I'll get you to double-check before we merge this.

SpiritBob:
> Thanks @SpiritBob. This looks really good. Let's mark it as ready-for-review, then I'll add a commit on top with a few stylistic changes, etc, which I'll get you to double-check before we merge this.

Sounds perfect!

serpent5:
@SpiritBob Have you tested that your code sample works with the built-in factory? I remember needing to use `PostConfigure` so that the framework would correctly initialise the factory.

SpiritBob:
> @SpiritBob Have you tested that your code sample works with the built-in factory? I remember needing to use `PostConfigure` so that the framework would correctly initialise the factory.

I've based my research upon the current up-to-date source code in:
https://source.dot.net/#Microsoft.AspNetCore.Mvc.Core/DependencyInjection/ApiBehaviorOptionsSetup.cs,b0765e23933b9a31

As you can see, by default the delegate already takes for granted the fact that `ProblemDetailsFactory` exists in the DI container.  (injected in https://source.dot.net/#Microsoft.AspNetCore.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs,257 which is part of `AddControllers`/`AddMvc`)

Perhaps what you were experiencing is the pre 3.0 behavior where everything was different - there was no `ProblemDetailsFactory` back then, and the objects being returned were not even conforming to the standard that `ValidationProblemDetails` meets?

I tested it just in-case, and the lambda is getting hit and returning the expected response.

serpent5:
> Perhaps what you were experiencing is the pre 3.0 behavior where everything was different - there was no `ProblemDetailsFactory` abstraction back then, and the objects being returned were not even conforming to the standard that `ValidationProblemDetails` meets?

Yes, that's what it is. I first wrote the example in the issue for 2.2, so it changed in 3.x. That makes things simpler, which is win-win. 👍

aspnetcore/web-api/index/samples/3.x/Startup.cs
```diff
@@ -48,27 +48,19 @@ public void ConfigureServices(IServiceCollection services)
...
-                        // or even use a custom object.
-                        var result = builtInFactory(context);
-
-                        // If accessing the returned ValidationProblemDetails object is required, get a reference to it.
-                        var problemDetails = (ValidationProblemDetails)((ObjectResult)result).Value;
-
-                        // Modify & Log accordingly.
-
-                        return result;
+                        // Invoke the default behavior, which produces a ValidationProblemDetails response.
```
SpiritBob, line 61: Should we leave it as `which produces a ValidationProblemDetails response`, or change it to `which produces a ObjectResult response containing ValidationProblemDetails`?

serpent5: I was just going to ask you to review my changes...

Just above this, in the previous section, we say:

> With a compatibility version of 2.2 or later, the default response type for an HTTP 400 response is ValidationProblemDetails.

I went with what you're commenting on here for consistency with that.

SpiritBob (commented):
Looks good to me!

serpent5:
Thanks @SpiritBob. I'll get this merged once I've reviewed the final result in our internal preview system. Thanks for being so responsive to my suggestions and for offering up some great opinions. I think what we've ended up with is a good balance - I hope you agree. 😄

SpiritBob:
> Thanks @SpiritBob. I'll get this merged once I've reviewed the final result in our internal preview system. Thanks for being so responsive to my suggestions and for offering up some great opinions. I think what we've ended up with is a good balance - I hope you agree. 😄

Couldn't agree more! Always a pleasure to have you around.

PR #24499, opened 2022-01-04
Update toc.yml

Rick-Anderson:

PR #24501, opened 2022-01-04
Revert "Update toc.yml"

Rick-Anderson:
Reverts dotnet/AspNetCore.Docs#24499


Rick-Anderson:
When it comes to the definition of startup, the research indicates that there is no such word as a startup.

serpent5:
It looks like "start-up" and "startup" are valid too: https://www.merriam-webster.com/dictionary/start-up. "start up" looks odd to me, but that's a very small sample size.

PR #24520, opened 2022-01-06
Fix typo

jbomhold3:
The comment was listed as Microsoft.Http.Extensions but the package is Microsoft.Extensions.Http.


guardrex (approved):
Thanks @jbomhold3! :rocket:

PR #24586, opened 2022-01-12
Add a note about subscriptions triggered thru CascadingValue

pranavkm:
CascadingValue causes components to be re-rendered if the value changes. We don't currently document this.


pranavkm:
@guardrex feel free to rewrite this entirely. I noticed that we didn't document this behavior in our docs and wanted to make sure we mentioned it.

pranavkm:
Thanks!

PR #24603, opened 2022-01-13
Injection into constructor allowed by Factory-based middleware

explorest:
Scoped services can be injected into the middleware's InvokeAsync method (as the previous point mentions) even if the middleware is NOT Factory-based.  Injection into the constructor, on the other hand, throws exception.  So it is the scoped service's injection into constructor -- without throwing exception -- that is allowed by Factory-based middleware. This is also explicitly mentioned in the detailed topic on Factory-based middleware here: 
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/extensibility?view=aspnetcore-6.0#:~:text=IMiddleware%20is%20activated%20per%20client%20request%20(connection)%2C%20so%20scoped%20services%20can%20be%20injected%20into%20the%20middleware%27s%20constructor.


serpent5 (approved):

serpent5:
Thanks for your contribution, @explorest. Be sure to check [What's New](https://docs.microsoft.com/en-us/aspnet/core/whats-new/?view=aspnetcore-6.0) early in February, where you'll see yourself listed as a valued contributor.

PR #24634, opened 2022-01-16
fix wrong application name WebAppAuth with WebApp1

rug:
The wrong application name WebAppAuth is used twice in CLI commands instead of correct name WebApp1, this puts the correct name.


serpent5:
Hi @rug. Thanks for the PR. I'll need to close and re-open this to kick our internal build system.

serpent5 (approved):

PR #24682, opened 2022-01-20
Update ProductsController.cs

Rick-Anderson:
@wadepickett  this is the correct controller.


wadepickett (approved):
Thanks Rick.  I see where I went wrong.  I should have moved 
namespace ProductsCore.Controllers {}
to
namespace ProductsCore.Controllers;

Rick-Anderson:
That and a bunch of unneeded includes

wadepickett:
> That and a bunch of unneeded includes

I was following what was in the previously approved version, but should not have.  I can either leave in the default includes or instruct to remove them so the user understands the change.  I will make sure I always remove.  Thanks.

PR #24683, opened 2022-01-20
File uploads enhancements

### LINKED ISSUE #23645: Review file upload security guidance
Some clarification would be welcome on how incoming file streams may pose a security risk and why it isn't mitigated by default:

> Avoid reading the incoming file stream directly into memory. For example, don't copy file bytes into a MemoryStream or read as a byte array. These approaches can result in performance and security problems, especially in Blazor Server. Instead, consider copying file bytes to an external store, such as a blob or a file on disk. If you need access to a Stream that represents the file's bytes, use IBrowserFile.OpenReadStream.

---
#### Document Details

* Content: [ASP.NET Core Blazor file uploads](https://docs.microsoft.com/en-us/aspnet/core/blazor/file-uploads?view=aspnetcore-5.0&pivots=server)
* Content Source: [aspnetcore/blazor/file-uploads.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/file-uploads.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #23645

Thanks @kanadaj! :rocket: ... This will clarify that line.

... **_ALSO_**, malicious hacksters 😠 have gotten right out of control. I feel compelled to offer an alternative that hardens apps and app servers by removing the app and server from the upload and file storage process entirely. My default notion was straight Azure Storage with JS client lib, but Azure Storage doesn't provide a built-in container quota, only by storage account ... **_Azure Files does!_** Let's consider including the added section on this PR to call out this approach and cross-link to the Azure Files docs for it. Devs will need to work out the implementation on their own using Azure Files/JS client/REST API resources, but at least they'll be aware of it.


mkArtakMSFT:
@TanayParikh can you please review this? Let's sync up after you've done so.

guardrex:
... and @TanayParikh ... reminder that you already approved the stream remarks on the issue. This is held up because of the Azure Files remarks. I don't like pitching cross-tech/replacement tech without management approval, even if it is Azure, which we generally ❤️, of course! 😄

aspnetcore/blazor/file-uploads.md
```diff
@@ -45,7 +45,10 @@ To read data from a user-selected file, call <xref:Microsoft.AspNetCore.Componen
...
+If you need access to a <xref:System.IO.Stream> that represents the file's bytes, use <xref:Microsoft.AspNetCore.Components.Forms.IBrowserFile.OpenReadStream%2A?displayProperty=nameWithType>. Avoid reading the incoming file stream directly into memory all at once. For example, don't copy all of the file's bytes into a <xref:System.IO.MemoryStream> or read the entire stream into a byte array all at once. These approaches can result in performance and security problems, especially for Blazor Server apps. Instead, consider adopting either of the following approaches:
+
+* Copy file bytes to an external store, such as a blob or file on disk.
+* Upload files from the client directly to an external service. For an example, see the [Alternative approach with Azure Files](#alternative-approach-with-azure-files) section.
```
mkArtakMSFT, line 51: These two options seem very similar - as both represent examples of uploading bytes to an external service, rather than to the server. Am I missing something?

aspnetcore/blazor/file-uploads.md
```diff
@@ -515,6 +518,21 @@ In Blazor Server, file data is streamed over the SignalR connection into .NET co
 
 :::zone-end
 
+## Alternative approach with Azure Files
```
mkArtakMSFT, line 521: What is important to call out is the preferred approach for handling file upload scenarios. Specifically, uploading files to a storage service (Azure Blob service, Azure File service, .etc. ...) and letting the server deal with the externally stored file data as needed.
Hence, let's call it out very explicitly here, that this section represents only an example of how that can be done with Azure Files.

aspnetcore/blazor/file-uploads.md
guardrex: Good question. That came from the original text provided by the PU. I took it to mean that the file bytes are copied to an external store **_but still handled directly by the app_**. I perhaps have misinterpreted what this means. In the second case, I was hoping to call out that the app doesn't handle the bytes directly, unless later it fetches them from the external store on its own. Should I need to research the remark further so that we can see which PU member provided it? ... and then ask them what they meant.

aspnetcore/blazor/file-uploads.md
guardrex: The main thing about specifying Azure Files explictly over Azure Storage (Blobs) is that Blob Storage doesn't have by-container quotas. The "quota" on Azure Storage is at the account level, not the container level. That's why I was feeling out Azure Files for this suggestion. Azure doesn't have anything else like this ... that offers this full list ... so that's why I wrote the draft this way.

guardrex: I guess it could say that tho ... it could remark on both generically (Azure Storage) and then for the quota bullet point it could say that Azure Files offers that one, not Azure Blob Storage. That way, the dev doesn't go to the trouble of looking into Azure Storage if they know that they want more fine-grained automatic control of quotas.

aspnetcore/blazor/file-uploads.md
guardrex: 🔍 **_Found it_** 🔍 ... It was @pranavkm-supplied text.

TL;DR @pranavkm ... For the following remark in **BOLD** text in the Blazor File Uploads topic ...

> Avoid reading the incoming file stream directly into memory. For example, don't copy file bytes into a \<xref:System.IO.MemoryStream> or read as a byte array. These approaches can result in performance and security problems, especially in Blazor Server. **Instead, consider copying file bytes to an external store, such as a a blob or a file on disk.**

Cross-reference: https://github.com/dotnet/AspNetCore.Docs/pull/20321/files

Did you mean that the app handles the file bytes directly and sends them to an external store? Or, did you mean that the client uploads file bytes directly to an external store? It's the "copying" word that's obscuring it for me (i.e., "streaming" or "sending" feels like the client is doing it, while "copying" feels like the app is doing it).

We're trying to figure out if your remark is the same or different than my added bullet about perhaps using Azure Files so that clients can use a JS client/REST API to directly send file bytes to an external store without the app handling them at all.

guardrex:
@mkArtakMSFT ... I fixed up that external service concept section. I continue to favor Azure Files because of the way it provides a quota feature **_by share_**. That's 🔥 **_HOT_** 🔥 imo. When we're done with this section (if it requires more work), I'll wait for Pranav to respond :point_up: and react to his feedback before merging this PR.

Do you want an issue to create a new doc for this 'external service' approach for Azure Files? If so, what priority should the issue be given?

aspnetcore/blazor/file-uploads.md
pranavkm: > Or, did you mean that the client uploads file bytes directly to an external store? It's the "copying" word that's obscuring it for me [i.e., "streaming" or "sending" feels like the client is doing it, while "copying" feels like the app (server-side) is doing it].

The latter. The idea is that you don't read arrays of bytes in to memory, because it increases GC pressure, could result in Large-Object-Heap allocations etc. Perhaps we could phrase the bullet points as:

* Upload files from the client directly to an external service. For an example, see the [Alternative approach with Azure Files](#alternative-approach-with-azure-files) section.
* Copy the stream directly to file on disk without reading it in to memory.

guardrex: Thx @pranavkm ... I'll use your exact bullet point language.

mkArtakMSFT (approved):
Thanks everyone!

Issue #24615, opened 2022-01-14
Blazor project tracking 2022


guardrex:
# 2022

For the 2021 idea list, UE pass list, and notes, see [Blazor project tracking 2021 (dotnet/AspNetCore.Docs #19286)](https://github.com/dotnet/AspNetCore.Docs/issues/19286).

## Doc ideas

Not **ALL** of these will be worked. This is an idea list/check list that don't rise to the level of opening an issue at this time.

* [ ] Consider adding a code example for the `MutationObserver` pattern where its discussed.
* [ ] File uploads topic: Fix accessibility of the yes/no icons. Use: `<span aria-hidden="true">XXXXX</span><span class="visually-hidden">XXXXX</span>`
* [ ] Add guidance on ***the reason(s)*** for using partial classes and base classes (inheritance) with pros and cons of each (partial = only splitting, base class = splitting and multiple-inheritance, useful in library-provided API scenarios).
* [x] Looks fixed, so check on the scenario for the Robot example, and then remove a NOTE if there's one there. https://github.com/dotnet/aspnetcore/issues/29342#issuecomment-1317311677 PR: https://github.com/dotnet/AspNetCore.Docs/pull/27716
* [x] Update the new load progress indicator with a NIT and include the recipe for the subtle bar that Steve shows in the audio app. https://github.com/dotnet/AspNetCore.Docs/pull/27543 and https://github.com/dotnet/AspNetCore.Docs/pull/27591
* [ ] Per Steve's .NET Conf talk, shouldn't we briefly cover [`FileSystem` API](https://developer.mozilla.org/en-US/docs/Web/API/FileSystem) use in Blazor apps?
* [ ] Let's run through the [`testassets` pieces in the framework](https://github.com/dotnet/aspnetcore/tree/main/src/Components/test/testassets) to see what might be helpful for devs to see in docs. It's a large task tho because there's so much there. This probably can't be worked in 23H1, possibly for 23Q3 tho.
* [ ] There's great PU-provided memory leak guidance on https://github.com/dotnet/aspnetcore/issues/43221. Take a look and see if/what could go into docs, possibly even just by linking the PU issue to cover it in the near-term.
* [ ] More for '23 than '22, I'm moving over a wrap-up item from the 7.0 tracking issue to this issue for work later: "Open an issue to replace in-text (usually non-working) examples with samples repo-based fully working cut-'n-paste examples. List the sections that require updates. This issue will be targeted for EOY or 23Q1."
[...]

PR #24685, opened 2022-01-20
Injected properties are expected to be available

guardrex:
Addresses #24615

Per https://github.com/dotnet/aspnetcore/pull/39445#discussion_r787781819 ...

Pranav, *keep/update this? ... or close the PR (e.g., it's too weedy ... too rare)?* 👂 

Currently, this is only for the 6.0 or later versions due to NRT enforcement for C# 8.0 or later. If we keep/modify this, let me know if you feel that 3.1/5.0 should have this guidance as well. 👂

### DIFF
--- a/aspnetcore/blazor/fundamentals/dependency-injection.md
+++ b/aspnetcore/blazor/fundamentals/dependency-injection.md
@@ -147,6 +147,7 @@ public class ComponentBase : IComponent
 >
 > For more information, see the following resources:
 >
+> * [Nullable reference types (NRTs) and .NET compiler null-state static analysis](xref:migration/50-to-60#nullable-reference-types-nrts-and-net-compiler-null-state-static-analysis)
 > * [Nullable reference types (C# guide)](/dotnet/csharp/nullable-references)
 > * [default value expressions (C# reference)](/dotnet/csharp/language-reference/operators/default#default-literal)
 > * [! (null-forgiving) operator (C# reference)](/dotnet/csharp/language-reference/operators/null-forgiving)
--- a/aspnetcore/blazor/fundamentals/dependency-injection.md
+++ b/aspnetcore/blazor/fundamentals/dependency-injection.md
@@ -142,7 +142,7 @@ public class ComponentBase : IComponent
 >
 > ```csharp
 > [Inject]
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/fundamentals/dependency-injection.md
```diff
@@ -137,6 +137,21 @@ public class ComponentBase : IComponent
 }
 ```
 
+> [!NOTE]
+> Don't mark injected services as nullable. Instead, assign a default literal with the null-forgiving operator (`default!`). For example:
```
pranavkm, line 141: I guess. One of the alternatives we were considering was to enable [constructor injection](https://github.com/dotnet/aspnetcore/issues/18088) by default in 7.0 which would remove some of this ugliness.

aspnetcore/blazor/fundamentals/dependency-injection.md
```diff
@@ -137,6 +137,21 @@ public class ComponentBase : IComponent
...
+>
+> ```csharp
+> [Inject]
+> private IServiceProvider ServiceProvider { get; set; } = default!;
```
pranavkm, line 145:
```suggestion
> private IMyExampleService ExampleService { get; set; } = default!;
```

pranavkm (approved):

aspnetcore/blazor/fundamentals/dependency-injection.md
jessegood: Sorry for the nitpick, but I think it would be best to clarify as to why you shouldn't mark properties as nullable:

> **Since injected properties are always expected to be available, as a best practice** don't mark injected services as nullable. Instead...

The first time I read this, I thought that there might be some more technical reason as to why you shouldn't mark it as nullable.

guardrex: It's not just a best practice from the remarks made: It's how Blazor is designed to work. Therefore, an imperative remark should be fine.

jessegood: I see, so as a consumer, we shouldn't think of this as being "optional" at all. Thanks for the clarification!

PR #24688, opened 2022-01-20
Update Kestrel Options to 6.0

### LINKED ISSUE #23653: .NET 6 update: Configure options for the ASP.NET Core Kestrel web server
Also add [Emit KestrelServerOptions via EventSource event](https://devblogs.microsoft.com/aspnet/asp-net-core-updates-in-net-6-rc-1/)

---
#### Document Details

* ID: 04754ed9-3508-6f40-fd7f-8e4a46e538af
* Version Independent ID: 5a5a34c5-1c75-ff17-b3d6-762194391319
* Content: [Configure options for the ASP.NET Core Kestrel web server](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/options?view=aspnetcore-6.0)
* Content Source: [aspnetcore/fundamentals/servers/kestrel/options.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/servers/kestrel/options.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

serpent5:
Fixes #23653.

[Internal Preview](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/options?view=aspnetcore-6.0&branch=pr-en-us-24688).

Updated all code samples to 6.0, and moved inline snippets out to code references. As a couple of the defaults we call out don't match the latest defaults in the code, I decided to drop specifying defaults in general. I concluded that these are all shown and accurate in the API references for each property, so they're very easy to find out for the curious reader, but I can add them back in with the latest correct values if that's preferred.

### DIFF
--- a/aspnetcore/fundamentals/servers/kestrel/options.md
+++ b/aspnetcore/fundamentals/servers/kestrel/options.md
@@ -5,7 +5,7 @@ description: Learn about configuring options for Kestrel, the cross-platform web
 monikerRange: '>= aspnetcore-5.0'
 ms.author: riande
 ms.custom: mvc
-ms.date: 05/04/2020
+ms.date: 01/20/2022
 no-loc: [Home, Privacy, Kestrel, appsettings.json, "ASP.NET Core Identity", cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
 uid: fundamentals/servers/kestrel/options
 ---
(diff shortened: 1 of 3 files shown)


serpent5:
@halter73 I've been using the following example for configuring `KestrelServerOptions` values. Is this the correct approach to show?:

```csharp
builder.WebHost.ConfigureKestrel(serverOptions =>
{
    // ...
});
```

I think we can also use something like this:

```csharp
builder.Services.Configure<KestrelServerOptions>(serverOptions =>
{
    // ...
}
```

That's a bit more verbose, but it "hides" `WebHostBuilder` (I'm not saying that's necessarily a good thing). It also doesn't allow the `WebHostBuilderContext` to be passed in, but its properties are available on `builder` in this case. I think I convinced myself that `ConfigureKestrel` is the correct approach, but I just want to double-check that with an expert.

halter73:
I think it's just a matter of preference if you want to use `builder.WebHost.ConfigureKestrel()` or `builder.Services.Configure<KestrelServerOptions>()`. What you have is less verbose but the longer version makes it a little more obvious what's really going on. I'm fine with keeping the shorter version though.

halter73 (approved):

Rick-Anderson:
>  `builder.WebHost.ConfigureKestrel()` or `builder.Services.Configure<KestrelServerOptions>()`. What you have is less verbose but the longer version makes it a little more obvious what's really going on. I'm fine with keeping the shorter version though.

How about writing that.

serpent5:
> How about writing that.

I think that would end up with us showing the two options first and then justifying the one we've gone with. That might be slightly distracting, as the reader's just here to configure the options, and there's no real functional difference between the two approaches. I'd prefer to just stick with what we have, but only if you're OK with that.

Rick-Anderson:
Sure, let's keep what you have. Hopefully we can find this information is someone asks.

Issue #24695, opened 2022-01-21
Drop 2.x content and sample apps


guardrex:
Rick, I have a few minutes this AM ... I'll try a quick-ish pass on the node to see if I can fix up the 2.x aspects.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b391017f-25e1-d5e1-c4f7-020408ade7b5
* Version Independent ID: 263a7b64-2e42-bbfc-c1d3-61680438e1e4
* Content: [Create web APIs with ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/web-api/?view=aspnetcore-3.1)
* Content Source: [aspnetcore/web-api/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/web-api/index.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-webapi**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **scaddie**

PR #24699, opened 2022-01-21
Fix Web-Api C# warning and tweak IEnumerable coverage

serpent5:
Follow-up for #24695.

[Internal Preview](https://review.docs.microsoft.com/en-us/aspnet/core/web-api/action-return-types?view=aspnetcore-6.0&branch=pr-en-us-24699).

cc @guardrex


aspnetcore/web-api/action-return-types.md
```diff
@@ -33,14 +33,7 @@ When multiple return types are possible, it's common to mix an <xref:Microsoft.A
...
-
-```csharp
-public IEnumerable<Product> GetOnSaleProducts() =>
-    _context.Products.Where(p => p.IsOnSale);
-```
-
-Consider declaring the action signature's return type as `IAsyncEnumerable<T>` to guarantee the asynchronous iteration. Ultimately, the iteration mode is based on the underlying concrete type being returned. MVC automatically buffers any concrete type that implements `IAsyncEnumerable<T>`.
+ASP.NET Core buffers the result of actions that return <xref:System.Collections.Generic.IEnumerable%601> before writing them to the response. Consider declaring the action signature's return type as <xref:System.Collections.Generic.IAsyncEnumerable%601> to guarantee asynchronous iteration. Ultimately, the iteration mode is based on the underlying concrete type being returned. MVC automatically buffers any concrete type that implements `IAsyncEnumerable<T>`.
```
serpent5, line 36: Checking this back against the previous version, I think the explanation about blocking and thread starvation was specific to 2.x or earlier. I've pieced this back together to how I think it read for 3.x and later. With this change, I don't think the example code is needed anymore, as it's shown slightly in the next code example anyway.

aspnetcore/web-api/advanced/conventions/sample/ApiConventions.csproj
```diff
@@ -1,11 +1,8 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
+    <IncludeOpenAPIAnalyzers>true</IncludeOpenAPIAnalyzers>
```
serpent5, line 5: This is the fix for the analyzer's NRE and is documented [here](https://docs.microsoft.com/en-us/aspnet/core/web-api/advanced/analyzers?view=aspnetcore-6.0&tabs=visual-studio#reference-the-analyzer-package).

aspnetcore/web-api/action-return-types.md
serpent5, line 36: I _think_ this covers the 3.0+ stuff. The stuff I removed was prefixed with "In ASP.NET Core 2.2 and earlier" before, so I also dropped the sample code as the exact same code sample is shown right after now.

Rick-Anderson (approved):

PR #24707, opened 2022-01-22
Anchor link: what if page does not exist

yecril71pl:
If the referenced page does not exist, a link to the current page is generated.

This is a self-correction; the previous wording was correct only if the current page was in fact the main page.

### DIFF
--- a/aspnetcore/mvc/views/tag-helpers/built-in/anchor-tag-helper.md
+++ b/aspnetcore/mvc/views/tag-helpers/built-in/anchor-tag-helper.md
@@ -284,7 +284,7 @@ The generated HTML:
 <a href="/Attendee?attendeeid=10">View Attendee</a>
 ```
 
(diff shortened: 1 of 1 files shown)


aspnetcore/mvc/views/tag-helpers/built-in/anchor-tag-helper.md
```diff
@@ -284,7 +284,7 @@ The generated HTML:
 <a href="/Attendee?attendeeid=10">View Attendee</a>
 ```
 
-If the referenced page doesn't exist, a link to the main page is generated. No warning is indicated.
+If the referenced page doesn't exist, a link to the current page is generated. No warning is indicated except on debug level.
```
Rick-Anderson, line 287: It used the ambient value.

serpent5:
Thanks @yecril71pl.

PR #24714, opened 2022-01-23
Fix Grammar Mistake

erhanalankus:


guardrex:
Hello @erhanalankus ...  This isn't a grammar mistake. I think the delay in package restore was a common problem generating this error, which is why the guidance basically just says to 'try again.' However, there have been a few other problems over the years in scaffolding. Some of them prompted the issue being moved to the [dotnet/Scaffolding](https://github.com/dotnet/Scaffolding/issues) repo. It will need to be assessed on an issue before action can be taken with the topic.

Please open an issue using the **This page** feedback button and form at the bottom of the **English-US** topic. Use of the **This page** feedback form adds metadata to your GitHub issue that cross-links the topic and automatically pings the author.

Capture

PR #24716, opened 2022-01-23
Clarify requirement for localization

SQL-MisterMagoo:
In Blazor Server, it is not a requirement that the `CurrentCulture` and `CurrentUICulture` are set to the same value for resource based localizations to work, but in Blazor WebAssembly they must be set to the same culture or the `SatelliteResourcesLoader` will not load the resources that Localizers are expected to read.

A quick note to this effect for the Blazor WebAssembly section of this doc could save someone else the trouble of figuring this out.

### DIFF
--- a/aspnetcore/blazor/globalization-localization.md
+++ b/aspnetcore/blazor/globalization-localization.md
@@ -367,8 +367,9 @@ CultureInfo.DefaultThreadCurrentUICulture = culture;
 
 await host.RunAsync();
 ```
-> [!NOTE] 
-> In Blazor WebAssembly, both `DefaultThreadCurrentCulture` and `DefaultThreadCurrentUICulture` must be set to the same culture in order to load and use Localization Resources.
+
+> [!IMPORTANT] 
+> In Blazor WebAssembly, <xref:System.Globalization.CultureInfo.DefaultThreadCurrentCulture> and <xref:System.Globalization.CultureInfo.DefaultThreadCurrentUICulture> must be set to the same culture in order to load and use localization resources.
 
 The following `CultureSelector` component shows how to perform the following actions:
 
@@ -1236,6 +1237,9 @@ CultureInfo.DefaultThreadCurrentUICulture = culture;
 await host.RunAsync();
 ```
 
+> [!IMPORTANT] 
+> In Blazor WebAssembly, <xref:System.Globalization.CultureInfo.DefaultThreadCurrentCulture> and <xref:System.Globalization.CultureInfo.DefaultThreadCurrentUICulture> must be set to the same culture in order to load and use localization resources.
+
 The following `CultureSelector` component shows how to set the user's culture selection into browser local storage via JS interop. The component is placed in the `Shared` folder for use throughout the app.
 
 `Shared/CultureSelector.razor`:
@@ -2014,6 +2018,9 @@ CultureInfo.DefaultThreadCurrentUICulture = culture;
 
 await host.RunAsync();
 ```
+
+> [!IMPORTANT] 
+> In Blazor WebAssembly, <xref:System.Globalization.CultureInfo.DefaultThreadCurrentCulture> and <xref:System.Globalization.CultureInfo.DefaultThreadCurrentUICulture> must be set to the same culture in order to load and use localization resources.
 
 The following `CultureSelector` component shows how to set the user's culture selection into browser local storage via JS interop. The component is placed in the `Shared` folder for use throughout the app.
 
--- a/aspnetcore/blazor/globalization-localization.md
+++ b/aspnetcore/blazor/globalization-localization.md
@@ -368,8 +368,10 @@ CultureInfo.DefaultThreadCurrentUICulture = culture;
 await host.RunAsync();
 ```
 
-> [!IMPORTANT] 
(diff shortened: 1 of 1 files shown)


guardrex:
Please re-open this issue using the **This page** feedback button and form at the bottom of the **English-US** topic. Use of the **This page** feedback form adds metadata to your GitHub issue that cross-links the topic and automatically pings the author.

Capture

guardrex:
Sorry ... I thought this was an issue for a sec. I'm 🏃 on other things at the moment.

guardrex:
@hishamco ... Do you have a sec to take a look at this PR?

guardrex:
I made a few NIT updates ...

* We're on the WASM pivot for this, so we can strike the hosting model qualifier from the beginning of the sentence.
* Let's elevate it from NOTE to IMPORTANT.

Thanks @SQL-MisterMagoo! :rocket:

We'll hold for a bit to see if @hishamco has feedback.

btw -- I'm not opposed to this providing the *exact reason*; so if it is due to ...

> the `SatelliteResourcesLoader` will not load the resources that Localizers are expected to read

... that's cool. I'll add that to this line. However, `SatelliteResourcesLoader` has apparently moved. It's not on `main` over there. We'll need to track that down if we want to place a reference source cross-link in the docs due to its `internal` status.

@SQL-MisterMagoo ... No worries on this, but just for future reference ... It's best to cross-link the engineering discussion. Sometimes, I need to ping the PU, and it's best to see who was working it with you.

Cross-ref: https://github.com/dotnet/aspnetcore/issues/25296

SQL-MisterMagoo:
It's not clear to me if you need me to do anything more on this? Let me know if there is anything, it looks like you found the refs you wanted

guardrex:
Yes, it's all good @SQL-MisterMagoo. Let's see if Hisham likes this. He is **_THE_** (self-described) **_LOC MANIAC_** around here! 😆

guardrex:
@SQL-MisterMagoo @hishamco ... I think this is for the current release ...

https://github.com/dotnet/aspnetcore/blob/main/src/Components/WebAssembly/WebAssembly/src/Hosting/WebAssemblyCultureProvider.cs#L64

If we cross-link, that's what we'll want to use, so the explanation for *why* will be more like ...

> the `WebAssemblyCultureProvider` won't load the resources that Localizers are expected to read

... and for "Localizers" ... I take that to mean `IStringLocalizer` and `IStringLocalizer<T>`, correct?

SQL-MisterMagoo:
Yes, that is all correct.

guardrex:
Ok ... I made some updates to get the ref source in. That way, devs can go look if they want to. I'm not sure if we need that or not. It's not terrible, but it might be too **_weedy_** for the average dev to even mention it.

I'll sleep on it and hear your feedback on that. Otherwise, I think adding the exact API doc links for the localizers makes sense.

hishamco:
Sorry @guardrex we have an internet issue from Thursday in our country, let me check the PR if there's something I can help with

hishamco:
FYI I did a lot of work on Localization with Blazor in Oqtane Framework, seems there's few things to be noted especially in Blazor WASM, so please hold this until I will comment

hishamco:
@SQL-MisterMagoo in ASP.NET Core or Server Blazor, the localization middleware will set the `CurrentCulture` & `CurrentUICulture` based to the culture choosen in the winning culture provider

In Blazor WASM you need to set both `CultureInfo.DefaultThreadCurrentCulture` & `CultureInfo.DefaultThreadCurrentUICulture` 

Is that what you want to say or is there any difference. Also it's useful to point what the scenario that doesn't work with you

SQL-MisterMagoo:
@hishamco The point I think it is vital to make is that there is a difference in implementation between server and webassembly versions of Blazor with regard to the loading of resources.

In Server mode, you can set `CultureInfo.DefaultThreadCurrentCulture` & `CultureInfo.DefaultThreadCurrentUICulture` to different cultures (if you have a reason to do that) and the String Localizer will load the correct resource values (assuming they are present).

Blazor WebAssembly __I believe__ only loads resource files for the `CurrentCulture`, not the `CurrentUICulture` which is not obvious and not currently documented on this page.

So, you must set both `CultureInfo.DefaultThreadCurrentCulture` & `CultureInfo.DefaultThreadCurrentUICulture` to the same culture in Blazor Webassembly to get localized strings to work - which is not obvious currently.

hishamco:
> In Server mode, you can set CultureInfo.DefaultThreadCurrentCulture & CultureInfo.DefaultThreadCurrentUICulture to different cultures (if you have a reason to do that) and the String Localizer will load the correct resource values (assuming they are present).

No need to do that because localization APIs will take care about this unless there are some customization

> So, you must set both CultureInfo.DefaultThreadCurrentCulture & CultureInfo.DefaultThreadCurrentUICulture to the same culture in Blazor Webassembly to get localized strings to work - which is not obvious currently.

I'm agree on this part

SQL-MisterMagoo:
> > In Server mode, you can set CultureInfo.DefaultThreadCurrentCulture & CultureInfo.DefaultThreadCurrentUICulture to different cultures (if you have a reason to do that) and the String Localizer will load the correct resource values (assuming they are present).
> 
> No need to do that because localization APIs will take care about this unless there are some customization
> 
That's why I said "if you have a reason to" - I'm not trying to debate whether you would, just to point out the differences that should be documented (better)

guardrex:
This only touches the WASM pivot of the doc, so we're good.

Thanks for your ongoing help, @hishamco. Thanks for the PR, @SQL-MisterMagoo.

PR #24730, opened 2022-01-25
Logic Error

wuyuansushen:
Natively compressed assets shouldn't be compressed.


serpent5:
Hi @wuyuansushen. I think the original version of this is correct. The original version reads:

> You shouldn't compress natively compressed assets, such as PNG files. If you attempt to further compress a natively compressed response

I think that follows logically, because it states that you shouldn't compress such assets and then explains what happens if you _do_ attempt to process them.

wuyuansushen:
Thanks.

PR #24770, opened 2022-01-27
Param attribute value @-prefix convention

### LINKED ISSUE #24769: Adopt param attribute value prefix `@` convention
---
#### Document Details

* ID: 7e7ba96d-bae0-96ee-e76a-71c6e5f4c23f
* Version Independent ID: 5fe884b8-51e1-1b91-57f8-522ff0b7890c
* Content: [ASP.NET Core Blazor cascading values and parameters](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/cascading-values-and-parameters?view=aspnetcore-6.0)
* Content Source: [aspnetcore/blazor/components/cascading-values-and-parameters.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/cascading-values-and-parameters.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #24769

Years ago, I performed a similar set of updates to make how we assign to parameters consistent in code examples. While I was working on something else today, I noticed that *a few sneaky devils* 😈 *are back!* This PR seeks to fix them up.

Included here is an effort to fix `this` keyword assignments, which I don't think are fundamentally different to anything else. Let me know if I'm mistaken on that. We've had (and I've seen outside of the doc set) any of the following. The last one is consistent with the earlier decision to quote and `@`-prefix these, and that's what's on this PR ...

* `<Something Value=this />`
* `<Something Value="this" />`
* `<Something Value="@this" />`

What about a boolean tho? Razor seems happy with `@true`. Doesn't the **_second one_** 👇 lend itself to consistency in presentation? If so, I need to apply one more update here.

* `<Something IsBlah="true">`
* `<Something IsBlah="@true">`

### DIFF
--- a/aspnetcore/blazor/components/cascading-values-and-parameters.md
+++ b/aspnetcore/blazor/components/cascading-values-and-parameters.md
@@ -139,7 +139,7 @@ Child `Tab` components aren't explicitly passed as parameters to the `TabSet`. I
 
 <!-- Display the tab headers -->
 
-<CascadingValue Value=this>
+<CascadingValue Value="@this">
     <ul class="nav nav-tabs">
         @ChildContent
     </ul>
@@ -358,7 +358,7 @@ Child `Tab` components aren't explicitly passed as parameters to the `TabSet`. I
 
 <!-- Display the tab headers -->
 
-<CascadingValue Value=this>
+<CascadingValue Value="@this">
     <ul class="nav nav-tabs">
         @ChildContent
     </ul>
@@ -577,7 +577,7 @@ Child `Tab` components aren't explicitly passed as parameters to the `TabSet`. I
 
 <!-- Display the tab headers -->
 
-<CascadingValue Value=this>
+<CascadingValue Value="@this">
     <ul class="nav nav-tabs">
         @ChildContent
     </ul>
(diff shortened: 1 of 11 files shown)


pranavkm:
I would think leaving booleans as-is would be preferable. Tooling would show you that a boolean value that is expected so in that  sense, it's less of a clutter to do `value="true"` over `value="@true"`.

guardrex:
Thx @pranavkm ... Ok, I'll leave booleans un-prefixed.

I'm not sure if we should say something about this delta or not. Scroll down to the purple NOTE box at ...

https://docs.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-6.0#component-parameters

When we use `@{directive attribute}` (e.g., `@bind="something"`), we never prefix the "something" unless it's an explicit expression (`@(...)`). For parameter attributes, we always add the `@` to the value (e.g., `<Something Whatever="@something" />` ... this boolean case will be a delta on that. I kind'a feel like we should say something about the delta because readers see things like the following in the docs and get 😖 ...

```razor
<CascadingValue Value="@this" IsFixed="true">
    @ChildContent
</CascadingValue>
```

guardrex:
Guidance approved offline.

PR #24786, opened 2022-01-28
'Position' is a field, not property

explorest:
[ The title is self descriptive , can't think of any more description ]


guardrex (approved):
Thanks @explorest! :rocket:

@serpent5 ... ~Is the capitalization ok here? It's normally a violation of coding guidelines to capitalize a field var.~

OH! ... It's a const ... nevermind.

PR #24794, opened 2022-01-28
Update index.md

kacharyea:


guardrex:
Hello @kacharyea ... We avoid 2nd person in most cases per our guidelines, and this remark isn't necessary given that the text doesn't specify CLI requirements. We rely upon the .NET Core docs in many cases for detailed requirements. For example, https://docs.microsoft.com/dotnet/core/install/.

PR #24798, opened 2022-01-30
Delete duplicated content

simonecavuoti:
In the "ApiController attribute" chapter the first four link are duplicated below, pointing to the same resources (e.g. the sixth is a duplicate of the first and so on. Only the fifth one is not repeated.)


guardrex (approved):
Thanks for the PR, @simonecavuoti.

LGTM, but let's ask Rick to look, too, because it looks like the indentation is incorrect as well.

**UPDATE**: I see ... the indentation is just a strange GH artifact. All of these lines are fully left justified. :+1:

Rick-Anderson (approved):

PR #24849, opened 2022-02-02
Fix docker docs (deep) links

timmkrause:
Fix some (deep) links against docker docs.

One of them is still invalid (`Docker for Windows: What to know before you install`) as it seems that it has been removed in the latest docs.

I don't know how everything is structured (working) on their end regarding the docs but I could find those (old?) resources over here:

https://docs.docker.com.xy2401.com/docker-for-windows/install/#what-to-know-before-you-install


Rick-Anderson (approved):

Rick-Anderson:
Thanks for the fix. Be sure to check out your name as a contributor about the 3rd of next month on the [What's new page](https://docs.microsoft.com/en-us/aspnet/core/whats-new/) in the **Community contributors** section at the bottom.

PR #24866, opened 2022-02-03
Fix asp-for sample (a genuine mistake?)

yecril71pl:
The attribute `asp-for="@Name"` is not what you would normally write.


serpent5:
Thanks @yecril71pl.

PR #24868, opened 2022-02-03
Update RCL control of <head> content guidance

### LINKED ISSUE #24867: Link Component in Blazor
According to this paragraph there should be a `Link` component in Blazor which is distinct from `link` in html. But I get the warning that there is not such component. Does it require to add any package to RCL to have that?

> To provide additional library component styles from stylesheets in the library's wwwroot folder, link the stylesheets using the framework's Link component.

image

---
#### Document Details

* ID: d2b709f5-e04e-fee3-7eea-1364893ab997
* Version Independent ID: a92c88d3-6ee8-1e69-f4c3-49eb82909746
* Content: [Consume ASP.NET Core Razor components from Razor class libraries](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/class-libraries?view=aspnetcore-6.0&tabs=visual-studio)
* Content Source: [aspnetcore/blazor/components/class-libraries.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/class-libraries.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #24867

Thanks @abdollahkahne! 🎷 ... This will take care of the **_rotten little gremlin!_** 😈 

Let me know (perhaps in a new issue because the one you opened will close) ... let me know if you hit a problem with the approach. I do need to check it, but perhaps I don't need to implement it myself if I don't hear back from you. I'll know that the new guidance is ok. I'm going to merge this to the live doc set right after it builds, so this will go live in the article within perhaps 20 or 30 minutes.

### DIFF
--- a/aspnetcore/blazor/components/class-libraries.md
+++ b/aspnetcore/blazor/components/class-libraries.md
@@ -195,7 +195,7 @@ The background image is also included from the RCL project template and resides 
 
 ![Diagonally-striped background image from the RCL project template](~/blazor/components/class-libraries/_static/background.png)
 
-To provide additional library component styles from stylesheets in the library's `wwwroot` folder, link the stylesheets using the framework's `Link` component.
+To provide additional library component styles from stylesheets in the library's `wwwroot` folder, add stylesheet `<link>` tags with a [`HeadContent` component](xref:blazor/components/control-head-content).
 
 The following background image is used in the next example. If you implement the example shown in this section, right-click the image to save it locally.
 
@@ -221,7 +221,9 @@ Add a component to the RCL that uses the `extra-style` class.
 `ExtraStyles.razor` in the `ComponentLibrary` RCL:
 
 ```razor
-<Link href="_content/ComponentLibrary/additionalStyles.css" rel="stylesheet" />
+<HeadContent>
+    <link href="_content/ComponentLibrary/additionalStyles.css" rel="stylesheet" />
+</HeadContent>
 
 <div class="extra-style">
     <p>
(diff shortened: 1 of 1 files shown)


guardrex:
Ok ... merged to live. Now, we'll just wait for it show up.

guardrex:
Something occurs to me tho. The old `Link` component didn't require the app that consumes the RCL to have any special code or provisions. The `HeadContent` component likely requires the consuming app to have a `HeadOutlet` component. Normally (i.e., via a project template), that would be true. However, I don't think the RCL topic can assume that. I think at a minimum I need to call it out. I think at most I need to confer with the PU to see what an RCL author should do to either confirm `HeadOutlet` is present in the consumer, add it themselves via the RCL, or tell them to avoid using it at all. I'll make a note on this on my tracking issue and add further updates to this topic shortly.

guardrex:
Ok @abdollahkahne ... It seems live in the main doc set now. I've made a note to follow up on this quickly. We probably don't want RCL authors to assume that the `HeadOutlet` component (used by the `HeadContent` component) is in the app that consumes the RCL. I'll start with a NOTE and follow-up with the product unit on it shortly. Thanks again for surfacing this doc bug! 😈 

https://docs.microsoft.com/aspnet/core/blazor/components/class-libraries

guardrex:
**UPDATE (2/28)**: We discussed the control of `<head>` content by an RCL. The answer is that class lib authors probably shouldn't have an RCL **_directly control_** `<head>` content. RCL's can use the built-in/OOB behaviors for CSS isolation and bundled styles ... that's fine ... but RCL's shouldn't set `<head>` content for the RCL consumer directly. Therefore, that's the modification that I'll be making to the topic today.

Having said that (and taking the topic's guidance in that direction), it's true that an dev or organization can use `HeadContent` components **_if they want to at their own risk as an advanced scenario_**. They also can (again, **_at their own risk_**), do JS interop with `<head>` content, which is also something that we won't cover.

PR #24905, opened 2022-02-07
No source when property value is empty

yecril71pl:
When a default value for property is non-empty and the value passed by the client is empty, the default value wins.

```C#
[Required]
public string Test { get; set; } = "!";
```

```cshtml
<input asp-for="Test" >
```

### DIFF
--- a/aspnetcore/mvc/models/model-binding.md
+++ b/aspnetcore/mvc/models/model-binding.md
@@ -146,7 +146,7 @@ The preceding code puts the custom value provider after all built-in value provi
 
 ## No source for a model property
 
-By default, a model state error isn't created if no value is found for a model property. The property is set to null or a default value:
+By default, a model state error isn't created if no value is found for a model property or the value is empty. The property is set to null or a default value:
 
 * Nullable simple types are set to `null`.
 * Non-nullable value types are set to `default(T)`. For example, a parameter `int id` is set to 0.


yecril71pl:
> Hi @yecril71pl. This particular section describes model-binding errors, so I don't see the connection between your example and generating a model-state error. The suggested change describes a form of precedence, but doesn't alter how model-binding works, AFAICT. Would you like to provide more information around the motivation for this change?

After reading the documentation, I expected the empty value sent by the browser to replace the default value, which should be preserved only if the parameter in question is not sent at all.  This does not happen.

serpent5:
This isn't the correct place for this change. Consider the following URL:

```
https://localhost:5001/Test?someValue=
```

Here, the value for `someValue` is empty. If I have a model with `[Required]` applied to its `SomeValue` property, a model-binding error _is_ generated using the preceding URL. The change in this PR suggests that an error _isn't_ generated, but it is.

serpent5 (changes requested):
This isn't the correct place for this change. Consider the following URL:

```
https://localhost:5001/Test?someValue=
```

Here, the value for `someValue` is empty. If I have a model with `[Required]` applied to its `SomeValue` property, a model-binding error _is_ generated using the preceding URL. The change in this PR suggests that an error _isn't_ generated, but it is.

yecril71pl:
> This isn't the correct place for this change. Consider the following URL:
> 
> ```
> https://localhost:5001/Test?someValue=
> ```
> 
> Here, the value for `someValue` is empty. If I have a model with `[Required]` applied to its `SomeValue` property, a model-binding error _is_ generated using the preceding URL. The change in this PR suggests that an error _isn't_ generated, but it is.

If `someValue` has a default value of `"!"` and an empty value is passed as a parameter (I experimented with `POST` parameters), the empty value is ignored, the property gets the default value, which is non-`null`, therefore no error is generated.

serpent5:
Ok, perhaps there's a difference between GET and POST, or between URL mapping and forms-handling. Regardless of where the difference is, the current coverage of this suggests that the default value will be used in all situations, but I think we can agree that this isn't true.

yecril71pl:
> Ok, perhaps there's a difference between GET and POST, or between URL mapping and forms-handling. Regardless of where the difference is, the current coverage of this suggests that the default value will be used in all situations, but I think we can agree that this isn't true.

I do not know about all situations, I only wrote what I saw: that the default value overrides an empty value explicitly sent by the client.  Which is unexpected.

serpent5:
Hi @yecril71pl. I'm closing this PR for reasons I've explained above. I agree that the behaviour you describe exists, but this is neither the correct place for the remark nor exhaustive enough in explaining the exact scenario where said behaviour occurs. If you feel strongly about adding an explanation, please do open a new PR with a more appropriate location and detailed explanation.

yecril71pl:
This variety and inconsistency suggests it is a bug that needs to be fixed rather than a feature that needs to be documented.

PR #24915, opened 2022-02-07
Surface two Blazor topics in the ToC

guardrex:
Thanks @serpent5 for catching these.

It seems like these were here earlier and that an update ... moving content, perhaps ... knocked them out.

OH! Wait! NO! I know what ... or **_WHO_** ... did this. **_This is IT!_** It's the coded signal to their operatives!

# 👽

### *The ATTACK PLANNING* 🌎 *is in its final stages!* 🛸

*They are among us!* 😆 

Seriously tho ... These have been **_well cross-linked_** into Blazor docs where needed. The major downside to my missing them is when readers strictly read topics in order top-to-bottom from the ToC. Thanks again, @serpent5 ... *great catch!*

### DIFF
--- a/aspnetcore/toc.yml
+++ b/aspnetcore/toc.yml
@@ -372,6 +372,8 @@ items:
                 uid: blazor/components/event-handling
               - name: Lifecycle
                 uid: blazor/components/lifecycle
+              - name: Virtualization
+                uid: blazor/components/virtualization
               - name: Rendering
                 uid: blazor/components/rendering
               - name: Templated components
@@ -438,6 +440,8 @@ items:
                     uid: blazor/security/webassembly/hosted-with-identity-server
                   - name: Additional scenarios
                     uid: blazor/security/webassembly/additional-scenarios
+                  - name: AAD groups and roles
+                    uid: blazor/security/webassembly/aad-groups-roles
                   - name: Graph API
                     uid: blazor/security/webassembly/graph-api
               - name: Content Security Policy


guardrex:
@serpent5 ... I'd like to add a bit of context here on the history and consensus on these types of problems. I'm not trying to make excuses for silly mistakes, but 🐛🐛🐛 seem to me to be quite unavoidable under the circumstances at anything other than a 🐌 pace.

Looking back over ~5 years here, this happening is **_totally NOT unexpected_**. I think it's somewhat proportional to the accomplishments ... proportional to the amount and speed of work performed, given the complexity of the doc set and the complexity individual topic content and versioning.

The faster I go 🏃😅 and the more issues that I knock out 🥊, the higher the **_absolute number of these errors_** I see occurring. If I go really slow (e.g., careful UE pass/overhaul-type work), I don't get as much total work done while favoring a focus on quality ... fewer of these *sneaky devils* 😈 get into the docs. That tradeoff is obvious to me now after all of these years here.

I think the consensus is that these are going to happen, no one should be put off by them, and they're virtually unavoidable at speed with lots of issues getting closed out quickly. The important thing has been knocking them out *ASAP* 😈🤛. Don't sweat it if you create some of these ... even as silly as this ToC 🐛 is. **_Good grief!_** 🙈😄 *LOL*. We've all had these things happen, and I'm sure that it will continue in spite of our best efforts to avoid them.

Of course ... if I'm wrong about all of that ... **_I'll be un-voluntarily "retired" PDQ by DR!_** ... which will put me in Fowler's neck of the woods ... **_on a sunny beach 🏖️ in Barbados 🇧🇧 with a 🍸 in my hand!_** 😆 Not too shabby. *I should be retired!* 😄 Sounds like a good idea to me!

serpent5:
Yeah, these things happen, and I'm surprised there weren't more across the repo. I'll keep going with things to try and spot stuff like this, as a background effort of some sort.

I fully agree that things like this are unavoidable.

PR #24916, opened 2022-02-07
Delete unused images

serpent5:
We have a fair few unused images in the repo.


Rick-Anderson:
Last time we tried to delete unused files there were too many problems so we gave up.

guardrex:
😄 yeah ... I think that's what I was remembering. I believe that in one attempt very early on many years ago that I screwed up and deleted one or a few and had to put them back!

serpent5:
Yeah, I thought this would be a good thing to do but I quickly realised there's more to it, as you've said. I'd like to come back and revisit this, but it'll have to be another day.

PR #24928, opened 2022-02-08
Update gRPC auth with interceptor scope example

JamesNK:
I haven't written docs before where the example starts off with "imagine a scenario...". Let me know if there is a better way to do this.

### DIFF
--- a/aspnetcore/grpc/authn-and-authz.md
+++ b/aspnetcore/grpc/authn-and-authz.md
@@ -128,10 +128,10 @@ services
     });
 ```
 
-A gRPC interceptor can also be used to configure a bearer token. An advantage to using an interceptor is the client factory can be configured to create a new interceptor for each client. This allows an interceptor to be constructed from DI using scoped and transient services.
+A gRPC interceptor can also be used to configure a bearer token. An advantage to using an interceptor is the client factory can be configured to create a new interceptor for each client. This allows an interceptor to be [constructed from DI using scoped and transient services](/dotnet/core/extensions/dependency-injection#service-lifetimes).
 
 Consider an app that has:
-* A user defined `ITokenProvider` for getting a bearer token which is scoped per-request.
+* A user-defined `ITokenProvider` for getting a bearer token. `ITokenProvider` is registered in DI with a scoped lifetime.
 * gRPC client factory is configured to create clients that are injected into gRPC services and Web API controllers.
 * gRPC calls should use `ITokenProvider` to get a bearer token.
 
--- a/aspnetcore/grpc/authn-and-authz.md
+++ b/aspnetcore/grpc/authn-and-authz.md
@@ -168,7 +168,7 @@ services
 The preceeding code:
 * Defines `AuthInterceptor` which is constructed using the user defined `ITokenProvider`.
 * Registers the `GreeterClient` type with client factory.
-* Configures the `AuthInterceptor` for this client using `InterceptorScope.Client`. A new interceptor is created for each client instance, so when a client is created for a gRPC service or Web API controller the scoped `ITokenProvider` is injected into the interceptor.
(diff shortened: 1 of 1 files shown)


aspnetcore/grpc/authn-and-authz.md
```diff
@@ -128,6 +128,45 @@ services
     });
 ```
 
+A gRPC interceptor can also be used to configure a bearer token. An advantage to using an interceptor is the client factory can be configured to create a new interceptor for each client. This allows an interceptor to be constructed from DI using scoped and transient services.
+
+Imagine an app has a user defined `ITokenProvider` which is scoped per-request, and clients injected into gRPC services and Web API controllers should call it to set the bearer token on all gRPC calls:
```
Rick-Anderson, line 133:
```suggestion
Consider an app that has a user defined `ITokenProvider` which is scoped per-request, and clients injected into gRPC services and Web API controllers should call it to set the bearer token on all gRPC calls:
```

Rick-Anderson (approved):
We use **Consider**  frequently.

aspnetcore/grpc/authn-and-authz.md
```diff
@@ -128,6 +128,48 @@ services
     });
 ```
 
+A gRPC interceptor can also be used to configure a bearer token. An advantage to using an interceptor is the client factory can be configured to create a new interceptor for each client. This allows an interceptor to be constructed from DI using scoped and transient services.
+
+Imagine an app has:
```
JamesNK, line 133:
```suggestion
Consider an app that has:
```

aspnetcore/grpc/authn-and-authz.md
```diff
@@ -128,6 +128,48 @@ services
     });
 ```
 
+A gRPC interceptor can also be used to configure a bearer token. An advantage to using an interceptor is the client factory can be configured to create a new interceptor for each client. This allows an interceptor to be constructed from DI using scoped and transient services.
+
+Consider an app that has:
+* A user defined `ITokenProvider` for getting a bearer token which is scoped per-request.
```
JamesNK, line 134:
```suggestion
* A user defined `ITokenProvider` for getting a bearer token. It is registered in DI with a scoped per-request lifetime.
```

aspnetcore/grpc/authn-and-authz.md
```diff
@@ -128,6 +128,48 @@ services
     });
 ```
 
+A gRPC interceptor can also be used to configure a bearer token. An advantage to using an interceptor is the client factory can be configured to create a new interceptor for each client. This allows an interceptor to be constructed from DI using scoped and transient services.
```
JamesNK, line 131:
```suggestion
A gRPC interceptor can also be used to configure a bearer token. An advantage to using an interceptor is the client factory can be configured to create a new interceptor for each client. This allows an interceptor to be [constructed from DI using scoped and transient services](/dotnet/core/extensions/dependency-injection#service-lifetimes).
```

aspnetcore/grpc/authn-and-authz.md
```diff
@@ -128,6 +128,48 @@ services
...
+    {
+        o.Address = new Uri("https://localhost:5001");
+    })
+    .AddInterceptor<AuthInterceptor>(InterceptorScope.Client);
+```
+
+The preceeding code:
+* Defines `AuthInterceptor` which is constructed using the user defined `ITokenProvider`.
+* Registers the `GreeterClient` type with client factory.
+* Configures the `AuthInterceptor` for this client using `InterceptorScope.Client`. A new interceptor is created for each client instance, so when a client is created for a gRPC service or Web API controller the scoped `ITokenProvider` is injected into the interceptor.
```
JamesNK, line 171:
```suggestion
* Configures the `AuthInterceptor` for this client using `InterceptorScope.Client`. A new interceptor is created for each client instance. When a client is created for a gRPC service or Web API controller, the scoped `ITokenProvider` is injected into the interceptor.
```

PR #24932, opened 2022-02-08
Fix code example

### LINKED ISSUE #24926: suggestion for doc adaptation - Support prerendering with authentication
Hi all,

I was busy with testing blazor prerendering with authentication. I followed a number of docs and I seem to have it working. Although there is an inconsistancy in the docs.

to prepare your code for prerender it is stated to comment out the next line of code when switching from index.htmk to _host.cshtml
`// builder.RootComponents.Add<App>("#app");`

This step seems to be correct. But it goes into conflict when reading the documentation on this page: https://docs.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/additional-scenarios?view=aspnetcore-6.0#support-prerendering-with-authentication More specific in the 5th code block about the _Host file.

As the #app isn't loaded anymore, the example of the mentioned page will stay stuck on "loading" after a login with callback to /authentication/login-callback (identity server default behaviour). This would thus render the login useless and impossible to fullfill (although an F5 seems to do wonders). But an adaptation as below seems to work without need to refresh the page;

```
<div id="app">
    @if (HttpContext.Request.Path.StartsWithSegments("/authentication"))
    {
        <component type="typeof({CLIENT APP ASSEMBLY NAME}.App)"  render-mode="WebAssembly" />
    }
    else
    {
        <component type="typeof({CLIENT APP ASSEMBLY NAME}.App)" 
            render-mode="WebAssemblyPrerendered" />
    }
</div>
```

guardrex:
Fixes #24926 

Thanks @jerrevds! :rocket: ... I'll ping Tanay to look, too, but see if the additional update here makes sense to you. There's a duplication of guidance for the `Pages/_Host.cshtml` page that I'm striking in the process.

Tanay ... This is Javier's original code for the scenario and then updated from https://github.com/dotnet/aspnetcore/issues/37713#issuecomment-961838175 and his following answer to my question there.

However, this looks correct. In the *Prerender and Integrate* topic, we guide adding the Component TH to `Pages/_Host.cshtml` ...

```razor
<component type="typeof(App)" render-mode="WebAssemblyPrerendered" />
```

... in the **_NON-auth case_**. Then, the WASM security doc should just pick up from there.

Cross-reference: **Step 4 of** https://docs.microsoft.com/en-us/aspnet/core/blazor/components/prerendering-and-integration?view=aspnetcore-6.0&pivots=webassembly#prerendering-configuration

In the WASM security doc, I'm dropping the extra bit of guidance for the `_Host.cshtml` file because ... as @jerrevds points out ... the dev should've already taken care of it by following the *Prerender and Integrate* topic guidance (Step 4). What we'll be left with in the WASM security doc is ...

* Factor services to a common location.
* **AVOID** prerendering for relative paths that start with `/authentication` with the code provided (i.e., @jerrevds update to Javier's code).

### DIFF
--- a/aspnetcore/blazor/security/webassembly/additional-scenarios.md
+++ b/aspnetcore/blazor/security/webassembly/additional-scenarios.md
@@ -875,7 +875,8 @@ In the **`Server`** app, create a `Pages` folder if it doesn't exist. Create a `
   <div id="app">
       @if (HttpContext.Request.Path.StartsWithSegments("/authentication"))
       {
-          <text>Loading...</text>
+          <component type="typeof({CLIENT APP ASSEMBLY NAME}.App)" 
+              render-mode="WebAssembly" />
       }
       else
       {
@@ -1926,7 +1927,8 @@ In the server app, create a `Pages` folder if it doesn't exist. Create a `_Host.
   <div id="app">
       @if (HttpContext.Request.Path.StartsWithSegments("/authentication"))
       {
-          <text>Loading...</text>
+          <component type="typeof({CLIENT APP ASSEMBLY NAME}.App)" 
+              render-mode="WebAssembly" />
       }
       else
       {
@@ -2977,7 +2979,8 @@ In the server app, create a `Pages` folder if it doesn't exist. Create a `_Host.
   <app>
       @if (HttpContext.Request.Path.StartsWithSegments("/authentication"))
       {
-          <text>Loading...</text>
+          <component type="typeof({CLIENT APP ASSEMBLY NAME}.App)" 
+              render-mode="WebAssembly" />
       }
       else
       {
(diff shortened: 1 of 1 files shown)


jerrevds:
Hi @guardrex as you asked to verify on my side too, the explanation and proposed adaptation seems valid

guardrex:
Thanks @jerrevds ... now, we'll just get Tanay's feedback/updates/approval and then this can go in. Thanks again for your help. These are some *tricky bits* 😈 to explain 😵😄.

guardrex:
@TanayParikh ... I'm going to proceed with this given that it's a doc 🐛. I'd like to get it live asap.

Let me know if you have a concern about the PR ... a further patch for the content ............. or a 💀 **_death threat_** 💀😆.

PR #24724, opened 2022-01-24
Generic type guidance enhancements

guardrex:
Fixes #23843

[Internal Review Link (links directly to the 'Generic type parameter support' section](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/components/index?view=aspnetcore-6.0&branch=pr-en-us-24724#generic-type-parameter-support)

For max clarity, I think we should demonstrate the cascading generic type concepts with explicit and inferred types separated in a pair of sub-sections. The approach of explicit types is *very relevant* given Steve's remarks on the PU design issue (his *In scope* section). For example, explicit types overrule inferred types, are definitely supported, and might make more sense (or be required) in some scenarios (e.g., cascading values with `[CascadingParameter]`).

Especially important to me is to improve the current guidance that doesn't use fully-working, cut-'n-paste examples.

Another pair of approaches is to either supply different data to child components of the `RenderFragment` or pass the same data to the children via a `CascadingParameterAttribute`.

After a bit of trial-and-error experimentation 😅, this PR seems to show that these various approaches can be laid out cleanly.

## *NOTES*

* Although the latest doc coverage effort prior to this PR was specifically addressing **_inferred_** types, all of the guidance for generic types needed work. This PR is aimed at addressing the whole subject of generic types, not just inferred types.

* There are some code comments in the doc to help with reviewing this. Those code comments will be removed before merging the PR. You only need to review the doc itself ... ` aspnetcore/blazor/components/index.md`.

* The 6.0 PU design issue is here :point_right: [Infer component generic types from ancestor components (dotnet/aspnetcore #29349)](https://github.com/dotnet/aspnetcore/issues/29349)

## ❓ *QUESTIONS* ❓ 

1. The guidance is NOT to supply a type parameter with a generic name, for example `T` or `TItem`. However, I am using `TExample` ... because these are doc examples. Should we change `TExample` here to something more specific and drop the `T` part of the name? ... or because this is just a doc example, is it ok? 👂 

1. It looks like generic array params were addressed for 6.0 on https://github.com/dotnet/aspnetcore/pull/34734. This is [Steve's repro](https://github.com/dotnet/aspnetcore/issues/29105#issuecomment-884041958) of what was breaking ...

   ```csharp
   [Parameter]
   public T[] Items { get; set; }
   ```

   ```razor
   <SomeComponent T="string" Items="@(Array.Empty<string>())"></SomeComponent>
   ```

[...]

PR #24933, opened 2022-02-08
Add File Downloads topic CORS guidance

### LINKED ISSUE #24887: Blazor WASM Download File From URL
The following code opens the file in the browser. It does not download it as the text states.   Is there some way to DownloadFileFromURL directly?

```
<button @onclick="DownloadFileFromURL">
    Download File From URL
</button>

@code {
    private async Task DownloadFileFromURL()
    {
        var fileURL = "{FILE URL}";
        var fileName = "{FILE NAME}";
        await JS.InvokeVoidAsync("triggerFileDownload", fileName, fileURL);
    }
}
```

---
#### Document Details

* ID: e4ad584d-0d9a-da8c-1296-b2d186e437ad
* Version Independent ID: e4ad584d-0d9a-da8c-1296-b2d186e437ad
* Content: [ASP.NET Core Blazor file downloads](https://docs.microsoft.com/en-us/aspnet/core/blazor/file-downloads?view=aspnetcore-6.0)
* Content Source: [aspnetcore/blazor/file-downloads.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/file-downloads.md)
* Product: **aspnet-core**
* GitHub Login: @TanayParikh
* Microsoft Alias: **taparik**

guardrex:
Fixes #24887

Thanks @bisaacson100! :rocket:

Tanay ... I think we'll need to do more than what I currently have on the PR. A couple of things cropped up as I was adding a simple *"watch out for CORS ... and see this"* remarks ...

1. Did this topic (on downloads) inadvertently receive a copy of the *File streams* section in the uploads topic? Even if the remarks should be here, can that first sentence ... perhaps the second sentence, too ... be more specific as to what it's explaining in this context (i.e., downloads)?

   > \#\# File streams
   >
   > In Blazor WebAssembly apps, file data is streamed directly from .NET code into the browser. In Blazor Server apps, file data is streamed over the SignalR connection from .NET code into the browser.

1. I think we need to scope the whole topic better. The topic starts with ...

    > The following example demonstrates how to download a file.

    Is the whole topic for hosted WASM only, or does it apply to Blazor Server, too? The earlier remark (perhaps out of place tho) that data is streamed "over the SignalR connection" is throwing me off because that shouldn't trigger a CORS failure, but the topic is using a browser download gesture ... i.e., it seems to apply to both hosting models. We probably need to say right at the top of the doc what hosting model this topic applies to (hosted WASM and Blazor Server) ... and qualify it further if it's talking about downloading from the app's own static assets or any file anywhere for each. The remark should clarify when CORS (different scheme/origin/port) is a potential problem.

1. Finally ... following up on that last point :point_up: ... we should be careful tossing readers to the main doc set CORS article for dealing with cross-origin file downloads. Depending on scoping of the overall scenario, when will readers find guidance on resolving CORS problems by merely following the guidance in our main CORS doc? When is referring a reader to CORS doc unhelpful? For example, the app/service that hosts the files for download isn't an ASP.NET Core app, so they won't be able to configure CORS using our CORS doc. Those devs must access docs for whatever is serving the files. Is that the only case where our CORS doc isn't going to help? It seems the CORS doc is good for hosted WASM with the **`Server`** app serving the files (or a different ASP.NET Core app serving the files) and Blazor Server apps where the Blazor Server app is hosting the files (or a different ASP.NET Core app is serving the files). Is that correct?

**NOTE TO SELF**: On the next commit, improve the sidebar ToC because the first link is down to the *File streams* section. That section might drop; but whatever the case, I think we need to isolate the example from introductory content (paragraph) that explains what the topic is about.

### DIFF
--- a/aspnetcore/blazor/file-downloads.md
+++ b/aspnetcore/blazor/file-downloads.md
@@ -138,9 +138,17 @@ In the preceding example, replace the placeholders with the following values:
 
 ## Cross-Origin Resource Sharing (CORS)
 
-Without taking further steps to enable cross-origin requests, downloading files from the same origin (same domain) must pass [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/docs/Web/HTTP/CORS) checks made by the browser.
+Without taking further steps to enable cross-origin requests, downloading files won't pass [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/docs/Web/HTTP/CORS) checks made by the browser.
 
-For more information on CORS with ASP.NET Core apps and Microsoft services that host files for download, see <xref:security/cors>. For more information on CORS with non-ASP.NET Core apps and non-Microsoft services, consult the CORS documentation of external framework or service.
+For more information on CORS with ASP.NET Core apps and other Microsoft products and services that host files for download, see the following resources:
+
+* <xref:security/cors>
+* [Using Azure CDN with CORS (Azure documentation)](/azure/cdn/cdn-cors)
+* [Cross-Origin Resource Sharing (CORS) support for Azure Storage (REST documentation)](/rest/api/storageservices/cross-origin-resource-sharing--cors--support-for-the-azure-storage-services)
+* [Core Cloud Services - Set up CORS for your website and storage assets (Learn Module)](/learn/modules/set-up-cors-website-storage/)
+* [IIS CORS module Configuration Reference (IIS documentation)](/iis/extensions/cors-module/cors-module-configuration-reference)
+
+For more information on CORS with non-ASP.NET Core apps and non-Microsoft services, consult the external framework or service CORS documentation.
 
 ## Security considerations
 
--- a/aspnetcore/blazor/file-downloads.md
+++ b/aspnetcore/blazor/file-downloads.md
@@ -11,6 +11,11 @@ uid: blazor/file-downloads
 ---
 # ASP.NET Core Blazor file downloads
 
+This article covers downloading files in Blazor apps. The guidance in this article:
+
+* Addresses downloading files from the app's own static assets or from any other location. When downloading files, [Cross-Origin Resource Sharing (CORS)](#cross-origin-resource-sharing-cors) considerations apply.
+* Applies to any [Blazor hosting model](xref:blazor/hosting-models).
+
 > [!WARNING]
 > Always follow security best practices when allowing users to download files. For more information, see the [Security considerations](#security-considerations) section.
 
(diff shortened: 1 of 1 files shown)


guardrex:
@TanayParikh ... Reminder on this one (in addition to https://github.com/dotnet/AspNetCore.Docs/pull/24724). This is probably the shorter ... *quicker* 🏃 ... one to deal with because it's just a few Q&A things that I can derive coverage from.

TanayParikh:
Thanks @guardrex for putting this together. I've answered the questions below, please let me know if there's anything I can clarify. 

> 1. Did this topic (on downloads) inadvertently receive a copy of the _File streams_ section in the uploads topic?

Yes good catch, this section can just be removed. 


> Is the whole topic for hosted WASM only, or does it apply to Blazor Server, too? The earlier remark (perhaps out of place tho) that data is streamed "over the SignalR connection" is throwing me off because that shouldn't trigger a CORS failure, but the topic is using a browser download gesture ... i.e., it seems to apply to both hosting models. We probably need to say right at the top of the doc what hosting model this topic applies to (hosted WASM and Blazor Server) ...

This docs article is applicable for both hosting models. 


> and qualify it further if it's talking about downloading from the app's own static assets or any file anywhere for each. The remark should clarify when CORS (different scheme/origin/port) is a potential problem.

It's for any file, anywhere for each. However, as you point out, this would have CORS implications depending on the origin of the resource being downloaded. 


> Finally ... following up on that last point ☝️ ... we should be careful tossing readers to the main doc set CORS article for dealing with cross-origin file downloads. Depending on scoping of the overall scenario, when will readers find guidance on resolving CORS problems by merely following the guidance in our main CORS doc? When is referring a reader to CORS doc unhelpful? For example, the app/service that hosts the files for download isn't an ASP.NET Core app, so they won't be able to configure CORS using our CORS doc. Those devs must access docs for whatever is serving the files. Is that the only case where our CORS doc isn't going to help? It seems the CORS doc is good for hosted WASM with the Server app serving the files (or a different ASP.NET Core app serving the files) and Blazor Server apps where the Blazor Server app is hosting the files (or a different ASP.NET Core app is serving the files). Is that correct?

I think this is a correct summary. The main point to highlight is as you said:

> Those devs must access docs for whatever is serving the files

so a WASM hosted/server app serving the file which the developer controls can be updated appropriately using our existing CORS documentation.

guardrex:
Thanks @TanayParikh ... I'll get on this on Thursday and see if I can push it to completion.

guardrex:
@TanayParikh ... Updates made. How does it look now?

TanayParikh (approved):

TanayParikh:
Thanks, looks good to me. :shipit:

PR #24950, opened 2022-02-09
Remove .wasm file extension before redefining it

campersau:
I had to remove the `.wasm` file extension because IIS was complaining about duplicate entries in order to make it work with `br` and `https`.

Others also reported this issue on stackoverflow:
https://stackoverflow.com/a/69888016
https://stackoverflow.com/a/70967738

### DIFF
--- a/aspnetcore/blazor/host-and-deploy/webassembly/_samples/web.config
+++ b/aspnetcore/blazor/host-and-deploy/webassembly/_samples/web.config
@@ -6,6 +6,7 @@
       <remove fileExtension=".json" />
       <remove fileExtension=".woff" />
       <remove fileExtension=".woff2" />
+      <remove fileExtension=".wasm" />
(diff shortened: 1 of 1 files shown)


guardrex (approved):
Thanks @campersau :rocket: ... and note that this whole node of topics (*Host and Deploy*) is due for an overhaul and series of updates per https://github.com/dotnet/AspNetCore.Docs/issues/24615. I've cross-linked the tracking issue to this PR and to the SO posts that you mentioned for when I test and improve the guidance.

I think one of the things that I'll do where it tells readers that the example `web.config` file is a starting point for further customization is to call out our list of recommended public support channels (SO, Slack, Gitter). I'll probably include a tag filter in the SO link (`https://stackoverflow.com/questions/tagged/blazor+iis+compression`).

Thanks for sending this PR.

Issue #24923, opened 2022-02-07
Blazor Hosting Models topic SEO updates


guardrex:
Follow-up discussion for SEO updates after https://github.com/dotnet/AspNetCore.Docs/pull/24878 and https://github.com/dotnet/AspNetCore.Docs/pull/24920. All of the updates are live in the doc set now at ...

https://docs.microsoft.com/aspnet/core/blazor/hosting-models

cc: @v-mepa

> should I use blazor server or blazor wasm

... and similar wording: sometimes with "versus"/"vs" instead of "or" and sometimes with "webassembly" instead of "wasm" in the search string.

The primary expression is placing the page 3rd (at Google). We'd like to eventually reach *First Place* 🥇.

https://www.google.com/search?q=should+i+use+blazor+server+or+blazor+wasm

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 542f38c4-1502-f739-372a-80c86c377491
* Version Independent ID: 72c2a844-ffb2-4799-2c7a-a7a73d3b6833
* Content: [ASP.NET Core Blazor hosting models](https://docs.microsoft.com/en-us/aspnet/core/blazor/hosting-models?view=aspnetcore-6.0)
* Content Source: [aspnetcore/blazor/hosting-models.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/hosting-models.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

v-mepa:
I went through the article and the very first thing we must update is the page title. I looked at the page titles of the other 2 sites which are ranking 1 & 2 on SERP  for  ‘should i use blazor server or blazor wasm’:

Rank 1 Website: [RoundTheCode](https://www.roundthecode.com/dotnet/blazor/blazor-server-vs-blazor-webassembly-four-ways-in-which-they-differ#:~:text=If%20you%20want%20an%20application,be%20looking%20at%20Blazor%20WebAssembly.)
Page Title: _Blazor Server vs. Blazor WebAssembly: Four Ways In Which They Differ_ 

Rank 2 Website: [Programmingwithwolfgang](https://www.programmingwithwolfgang.com/blazor-server-vs-blazor-webassembly/)
[...]

PR #24954, opened 2022-02-09
Update Blazor hosting models topic title

guardrex:
Addresses #24923

@v-mepa ... We can adjust that right now. Is this what you had in mind?

**UPDATE**: Approved on the issue. We'll do this now and perhaps make some additional tweaks in a bit.


v-mepa:
Yep let's update that and see if it makes any difference in rankings.

PR #24978, opened 2022-02-13
server goes off - line not the client

hjoab:
server goes off - line not the client


guardrex (approved):
Thanks @hjoab! :rocket:

PR #24980, opened 2022-02-13
extra unneeded word "is" removed

ptfw:


guardrex (approved):
Thanks @ptfw! :rocket:

PR #24995, opened 2022-02-14
Code-fence references to Program.cs

serpent5:
A global replace-all from `*Program.cs*` to `` `Program.cs` ``, excluding anything in `README.md` files.

cc @guardrex


guardrex:
Are READMEs localized? ... I didn't think that they were part of the build.

serpent5:
I don't think they're part of the build. That's why I didn't bother including them in the change here.

guardrex:
Oh! Geez! 😆 ... I thought that said "including." DUH!

To cut down on caffeine, I started making my morning pot of ☕ with 1/2 caf and 1/2 decaf. **_I think I need to go back to 100% full caf!_** 😄

Rick-Anderson (approved):
Nice consistency change now that we don't have the `Startup` class.

PR #25035, opened 2022-02-16
Update index.md

fscopel:
Remove duplicated paragraph

### DIFF
--- a/aspnetcore/blazor/index.md
+++ b/aspnetcore/blazor/index.md
@@ -25,6 +25,15 @@ Blazor is a framework for building interactive client-side web UI with [.NET](/d
 
 :::moniker-end
 
+:::moniker range="< aspnetcore-6.0"
+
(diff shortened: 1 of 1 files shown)


guardrex:
Hello @fscopel ... This is versioned content and thus this isn't the correct fix. I'll open an issue and address it. Thanks for calling it out.

guardrex:
Actually, I think I can fix it right here on your PR. It looks like I didn't paste the content into the right spot. 🙈

guardrex:
Ok ... that *lick'd IT!* 🐮👅😄 

We just placed that a few days ago. We're adding Blazor Hybrid content to the doc set now, so that last bullet should only appear for the 6.0 version of the topic. I was going tooooooo *FAST* 🏃 there. Thanks for catching it, @fscopel! 🚀

guardrex:
Uh, oh! Looks like we have some *gremlins* 😈 in the build system this morning. As soon as that service gets back online, I'll get this merged. I'll take this one to the live doc set immediately, too. Otherwise, we'll get more issues on it given the popularity of this particular topic.

PR #25039, opened 2022-02-16
Replace remaining /dotnet/api references with xrefs

### LINKED ISSUE #24809: Replace /dotnet/api references with xrefs
I'd like to go through the docs and replace all `/dotnet/api` references with the corresponding `xref` syntax. I think doing this one-by-one, as a background effort, would be most effective. We can link PRs to this issue until we're done.

serpent5:
Fixes #24809.

### DIFF
--- a/aspnetcore/blazor/host-and-deploy/server.md
+++ b/aspnetcore/blazor/host-and-deploy/server.md
@@ -36,7 +36,7 @@ Each circuit uses approximately 250 KB of memory for a minimal *Hello World*-sty
 
 ## SignalR configuration
 
-Blazor Server apps use [ASP.NET Core SignalR](<xref:signalr/introduction>) to communicate with the browser. [SignalR's hosting and scaling conditions](xref:signalr/publish-to-azure-web-app) apply to Blazor Server apps.
+Blazor Server apps use [ASP.NET Core SignalR](xref:signalr/introduction) to communicate with the browser. [SignalR's hosting and scaling conditions](xref:signalr/publish-to-azure-web-app) apply to Blazor Server apps.
 
 Blazor works best when using [WebSockets](xref:fundamentals/websockets) as the SignalR transport due to lower latency, better reliability, and improved [security](xref:signalr/security). [Long Polling](https://github.com/dotnet/aspnetcore/blob/main/src/SignalR/docs/specs/TransportProtocols.md#long-polling-server-to-client-only) is used by SignalR when WebSockets isn't available or when the app is explicitly configured to use Long Polling. When deploying to Azure App Service, configure the app to use WebSockets in the Azure portal settings for the service. For details on configuring the app for Azure App Service, see the [SignalR publishing guidelines](xref:signalr/publish-to-azure-web-app).
 
--- a/aspnetcore/migration/1x-to-2x/identity-2x.md
+++ b/aspnetcore/migration/1x-to-2x/identity-2x.md
@@ -256,7 +256,7 @@ Make the following changes in *Startup.cs*:
 
 ### Setting default authentication schemes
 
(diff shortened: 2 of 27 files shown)


aspnetcore/blazor/host-and-deploy/server.md
```diff
@@ -36,7 +36,7 @@ Each circuit uses approximately 250 KB of memory for a minimal *Hello World*-sty
 
 ## SignalR configuration
 
-Blazor Server apps use [ASP.NET Core SignalR](<xref:signalr/introduction>) to communicate with the browser. [SignalR's hosting and scaling conditions](xref:signalr/publish-to-azure-web-app) apply to Blazor Server apps.
+Blazor Server apps use [ASP.NET Core SignalR](xref:signalr/introduction) to communicate with the browser. [SignalR's hosting and scaling conditions](xref:signalr/publish-to-azure-web-app) apply to Blazor Server apps.
```
serpent5, line 39: @guardrex You might've seen that I've been doing a daily run of fixing up our `/dotnet/api` references across the repo. I've finally finished, so I also added a commit to remove the `<` and `>` around an xref that's got custom text. This is the only place I've touched anything under the Blazor node.

guardrex:
These "-1" bookmarks ... are those going to an earlier version of the coverage. They look like a mistake.

guardrex:
Yeah ... those are wrong. The "-1" should get the chop. I think the doc author had their doc version set to 5.0 when they grabbed the link. You might try that and see ... search the whole repo for "-1)" and see how many of those turn up.

serpent5:
There's just 3, fixed in https://github.com/dotnet/AspNetCore.Docs/pull/25039/commits/90898dcb28e0f923bedb37d6a9a1fd111ffa2853.

aspnetcore/migration/1x-to-2x/identity-2x.md
```diff
@@ -302,8 +302,8 @@ In 2.0 projects, import the `Microsoft.AspNetCore.Authentication` namespace, and
 
 There are two variations of Windows authentication:
 
-* The host only allows authenticated users. This variation isn't affected by the 2.0 changes.
-* The host allows both anonymous and authenticated users. This variation is affected by the 2.0 changes. For example, the app should allow anonymous users at the [IIS](xref:host-and-deploy/iis/index) or [HTTP.sys](xref:fundamentals/servers/httpsys) layer but authorize users at the controller level. In this scenario, set the default scheme in the `Startup.ConfigureServices` method.
+- The host only allows authenticated users. This variation isn't affected by the 2.0 changes.
```
guardrex, line 305: 😕 ... We always use the `*` for lists. Was there an internal change on this? Like I mentioned before, sometimes things change internally, and I get updated on it later.

guardrex (changes requested):

aspnetcore/migration/1x-to-2x/identity-2x.md
serpent5: Thanks for going through this. That whole article uses `-`, so I saw a warning for the use of `*` here. I just made it consistent, but I can go and fix the whole article to use `*`. I probably should've just done that anyway.

guardrex: Oh, ok. Nevermind then. It's perhaps not worth the trouble.

guardrex (approved):

aspnetcore/migration/1x-to-2x/identity-2x.md
serpent5: > Oh, ok. Nevermind then. It's perhaps not worth the trouble.

I'll do it. I just bottled it when I realised I'd have to change the whole file, but I shouldn't have. Thanks for pointing it out.

guardrex: > fighting a bit of a losing battle 

It's really a GIANT doc set at this point. We figured that we'd update things like this as we go primarily on total overhauls. When this topic is finally reached by someone in the distant future, then I think it would be made consistent.

serpent5: There we go, sorted. It was easier than I expected. Once I changed a couple, markdownlint told me that `-` was a bad egg and allowed me to fix all violations in the file with one click.

Issue #24956, opened 2022-02-09
Blazor Hybrid tracking


guardrex:
## Articles (ToC and physical/URL locations)

Topic | Location/filename | Issue | PR | &nbsp;
--- | --- | --- | --- | ---
Static files<sup>1</sup> | `aspnetcore/blazor/hybrid/static-files.md` | https://github.com/dotnet/AspNetCore.Docs/issues/25674 | https://github.com/dotnet/AspNetCore.Docs/pull/25891 | [Live Topic](https://docs.microsoft.com/aspnet/core/blazor/hybrid/static-files)
Routing & Navigation<sup>2</sup> | `aspnetcore/blazor/hybrid/routing-and-navigation.md` | &mdash; | https://github.com/dotnet/AspNetCore.Docs/pull/25130 | [Live Topic](https://docs.microsoft.com/en-us/aspnet/core/blazor/hybrid/routing)
Reuse Razor components<sup>3</sup> | `aspnetcore/blazor/hybrid/reuse-razor-components.md` | https://github.com/dotnet/AspNetCore.Docs/issues/25675 | https://github.com/dotnet/AspNetCore.Docs/pull/25847 | [Live Topic](https://docs.microsoft.com/aspnet/core/blazor/hybrid/reuse-razor-components)
~Integrate with native UI<sup>4</sup>~ **DROPPED&dagger;** | ~`aspnetcore/blazor/hybrid/integrate-with-native-ui.md`~ | https://github.com/dotnet/AspNetCore.Docs/issues/25676 | :x: | :x:
~Access native platform features<sup>5</sup>~ **DROPPED&dagger;** | ~`aspnetcore/blazor/hybrid/access-to-platform-features.md`~ | https://github.com/dotnet/AspNetCore.Docs/issues/25677 | :x: | :x:
~Native look and feel<sup>6</sup>~ **DROPPED&dagger;** | `aspnetcore/blazor/hybrid/native-look-and-feel.md` | https://github.com/dotnet/AspNetCore.Docs/issues/25678 | :x: | :x:
Globalization & Localization<sup>7</sup> | `aspnetcore/blazor/hybrid/globalization-localization.md` | https://github.com/dotnet/AspNetCore.Docs/issues/25679 | &mdash; | &mdash;
[...]

PR #25064, opened 2022-02-18
Add Blazor Hybrid to no-loc metadata

guardrex:
Addresses #24956

We want to make sure that Blazor Hybrid isn't localized, just like Blazor WebAssembly and Blazor Server.

IIRC, we're adding the global changes (via find-'n-replace) **_to the front of the list_**, while we'd add individual terms/phrases for specific topics to the end.


guardrex:
> I'd rather see "Blazor Hybrid" after "Blazor WebAssembly" in the global list

...... ah, **_BUT_** ... what you just said ...

> have a few no-locs that aren't consistent 

... is exactly why when there are hundreds ... like 600+ of these ... adding to the front the **_safest approach_**. It's sure-fire going to work every time. The inconsistencies were bound to show up over time, so we just said that we'd place new global ones on the front .and per-doc ones on the end. It should work if the 😈 stay out of our `no-loc` lists; specifically, duplicates could be a potential problem. True, the order isn't reasonable, but all of the global entries will go away one day, per the following ............

What we really need is a global list in the repo config, and I think that was something Jason said they were thinking about or planning. It's more of an internal docs thing, so I'll leave it to the team if someone wants to ping Khoi Pham and ask about that.

guardrex:
I won't hold u to account for this one. I'll go ahead without review, and they can **_RETIRE ME_** if they want 🏖️🇧🇧🍹😄 if things go sideways.

serpent5:
I think I can override them and do the merge. Shall I?

guardrex:
> I think I can override them and do the merge. Shall I?

That's ok ... I'll fix them. We need them fixed anyway, so I may as well. I'm working on them right now.

> that aren't using the complete "global" list

That's up to the team to address. AFAIK, the Blazor ones are consistent.

serpent5:
> AFAIK, the Blazor ones are consistent.

Yup, that's what I'm seeing too.

I'm more than happy to approve this PR. The comments I made were just higher-level. The approach you've got here makes sense to me. I did try and submit an approval, but it keeps hanging my browser. 🐌

guardrex:
Ok ... that **_lick'd 'um!_** 🐮👅 

**_ONE more thing to do before the 3-day weekend!_** *Woot!*

PR #25066, opened 2022-02-18
Make global "no-loc" tokens consistent across all articles

serpent5:
I've found 18 articles that don't start with the "global" `no-loc` list. The changes here ensure that everything starts with the same list, but still keeps local tokens on the end where needed. We had a few articles with `Index` in the list, but without `Index` in the content, so I've dropped this from said articles.

@Rick-Anderson Feel free to just close this one out. It's informative, at least, but I think it'd be good to make sure the global list is consistent.

cc @guardrex

### DIFF
--- a/aspnetcore/data/entity-framework-6.md
+++ b/aspnetcore/data/entity-framework-6.md
@@ -5,7 +5,7 @@ description: Entity Framework 6.3 and later works with ASP.NET Core 3.1 and late
 ms.author: riande
 ms.custom: mvc
 ms.date: 7/14/2020
-no-loc: ["Blazor Hybrid", Home, Privacy, Kestrel, appsettings.json, "ASP.NET Core Identity", cookie, Cookie, Blazor, "Identity", "Let's Encrypt", Razor, SignalR]
+no-loc: ["Blazor Hybrid", Home, Privacy, Kestrel, appsettings.json, "ASP.NET Core Identity", cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
 uid: data/entity-framework-6
 ---
 # ASP.NET Core and Entity Framework 6
--- a/aspnetcore/fundamentals/dependency-injection.md
+++ b/aspnetcore/fundamentals/dependency-injection.md
@@ -5,7 +5,7 @@ description: Learn how ASP.NET Core implements dependency injection and how to u
 ms.author: riande
 ms.custom: mvc
 ms.date: 11/21/2021
-no-loc: ["Blazor Hybrid", Home, Privacy, Kestrel, appsettings.json, "ASP.NET Core Identity", Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
+no-loc: ["Blazor Hybrid", Home, Privacy, Kestrel, appsettings.json, "ASP.NET Core Identity", cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
 uid: fundamentals/dependency-injection
 ---
 # Dependency injection in ASP.NET Core
--- a/aspnetcore/fundamentals/http-logging/index.md
+++ b/aspnetcore/fundamentals/http-logging/index.md
@@ -6,7 +6,7 @@ monikerRange: '>= aspnetcore-6.0'
 ms.author: jukotali
 ms.custom: mvc
 ms.date: 04/20/2021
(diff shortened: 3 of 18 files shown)


guardrex:
This is cool, @serpent5. 

I want to point out a few things in passing from prior discussions on this ...

* IIRC, Jason said that a global repo-wide configuration (i.e., a single repo config list) was either being considered or was actually planned. That would be **_super amazing_** :+1:. Then, each topic just gets a `no-loc` list for any special terms/phrases for that one article. Idk if you want to ping Khoi to ask what's up with that possible feature.
* Some of these are a bit strange to me to 'no-loc' out. I think a word like "Privacy" is a little scary 😨 given that "Privacy" is a word that might need localization in normal use ... e.g., a normal sentence with "Privacy" as the subject ...

  > Privacy is important to users.

  IMO, these project template-based pieces (Home, Privacy, Index, etc.) should be **_code fenced_**, not in the `no-loc` metadata. I do that throughout the Blazor docs. Anything like "... the `Privacy` component (`Pages/Privacy.razor`) ..." are code-fenced to match template assets.
* Finally ... it seems like "appsettings.json" should be code-fenced per our normal file code-fencing rule and not appear in this list. ✨ **_RegEx Magic!_**&trade; ✨ might capture a first pass list to find-'n-replace (a space ... some text that doesn't start with a backtick ... ".json"), then I'm sure that some follow-up reversions will be needed per things like code blocks.

Rick-Anderson (approved):

Rick-Anderson:
This constant changing of every file in the repo pollutes the history and makes it difficult to find a significant change in the history. It would be great to have one file with the loc. Is there anyway we could use an include or something that doesn't pollute the global history?

guardrex:
IIRC, we've been adding the `no-loc` metadata to each INCLUDE file because the loc engine applies it **_only by file_**, so an INCLUDE doesn't seem like it would apply the list to the whole doc where the INCLUDE appears.

I still don't get why it doesn't work via `docfx.json` per our normal `fileMetadata` entry ...

```json
"fileMetadata": {
  "no-loc": { "**.md": "[ ... ]" },
  ...
}
```

It seems like that would apply it, then the loc engine would run on it downstream. I guess it just doesn't work that way or else Jason would have told us to take this approach.

That ☝️ could be run easily enough in a quick test tho just to see what happens.

Rick-Anderson:
Internal issue https://portal.microsofticm.com/imp/v3/incidents/details/290559630/home

Rick-Anderson (changes requested):
Hold off until we get an answer

serpent5:
> That ☝️ could be run easily enough in a quick test tho just to see what happens.

Just for my own understanding... How would we be able to test this?

guardrex:
1. Create a fake doc that isn't in the ToC (`testdoc.md`) and doesn't have a `no-loc` metadata entry that just has a single paragraph sprinkled with a few common words ... "computer" ... "cat" ... whatever.
1. Add `fileMetadata` for it in `docfx.json` ...

   ```json
   "fileMetadata": {
     "no-loc": { "testdoc.md": "[ computer, cat ]" },
     ...
   }
   ```
1. Given the MT runs on a post-publish schedule, go ahead and take it live.
1. Check back in a few weeks ... did the "computer" and the "cat" words escape MT.

Even if that works tho, HT might not catch it because translators might not be looking at the `docfx.json` file for the string of words to avoid. The whole translation system is opaque to the community AFAIK ... it's ✨😄.

Rick-Anderson:
> Ashwin Manda (v-mandak) edited at 2022-02-23 21:13:35 PST
>
> The metadata in docfx.json seems a solid solution, with that way it puts no-loc words in a central place and take effect for all files affected

@guardrex  can you try that?

Rick-Anderson:
@guardrex  why is this topic not localized?
* https://docs.microsoft.com/zh-cn/aspnet/core/blazor/tutorials/signalr-blazor?#run-the-app-1
* https://docs.microsoft.com/fr/aspnet/core/blazor/tutorials/signalr-blazor?#run-the-app-1
* https://docs.microsoft.com/es/aspnet/core/blazor/tutorials/signalr-blazor?#run-the-app-1

Navigate to other topics and they are localized.

guardrex:
My sense of it is that HT runs on different schedules for different languages. I suppose Khoi would have to provide more info.

> metadata in docfx.json seems a solid solution

Yes, I'll give that a shot with a dummy test doc and report back.

serpent5:
Can I recreate this PR so that we at least have a consistent `no-loc` across the repo?

Rick-Anderson:
> Can I recreate this PR so that we at least have a consistent `no-loc` across the repo?

Let's hold off. Internal issue https://portal.microsofticm.com/imp/v3/incidents/details/290559630/home tracks doing this right

PR #25111, opened 2022-02-23
Adding additional comment about FromServices optionality

### LINKED ISSUE #25106: FromServices could returns null/default when optional
Issue describing feature: https://github.com/dotnet/aspnetcore/issues/39757
PR: https://github.com/dotnet/aspnetcore/pull/39804

---
#### Document Details

* Content: [Model Binding in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/model-binding?view=aspnetcore-6.0)
* Content Source: [aspnetcore/mvc/models/model-binding.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/mvc/models/model-binding.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

brunolins16:
Fix #25106.

### DIFF
--- a/aspnetcore/mvc/models/model-binding.md
+++ b/aspnetcore/mvc/models/model-binding.md
@@ -625,8 +625,12 @@ For more information, see [TryUpdateModelAsync](xref:data/ef-rp/crud#TryUpdateMo
 
 This attribute's name follows the pattern of model binding attributes that specify a data source. But it's not about binding data from a value provider. It gets an instance of a type from the [dependency injection](xref:fundamentals/dependency-injection) container. Its purpose is to provide an alternative to constructor injection for when you need a service only if a particular method is called.
 
-> [!TIP]
-> The binding will fail when obtaining an instance of a type, **not registered**, from the dependency injection container. If the service is expected to be **optional** then the parameter must be marked as nullable, including `?`, or set a default value. In this case, the method is responsible for the required check before the usage.
+If an instance of the type isn't registered in the dependency injection container, the app throws an exception when attempting to bind the parameter. To make the parameter optional, use one of the following approaches:
+
+* Make the parameter nullable.
+* Set a default value for the parameter.
+
+For nullable parameters, ensure that the parameter isn't `null` before accessing it.
 
 ## Additional resources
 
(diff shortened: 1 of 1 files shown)


aspnetcore/mvc/models/model-binding.md
```diff
@@ -625,6 +625,9 @@ For more information, see [TryUpdateModelAsync](xref:data/ef-rp/crud#TryUpdateMo
 
 This attribute's name follows the pattern of model binding attributes that specify a data source. But it's not about binding data from a value provider. It gets an instance of a type from the [dependency injection](xref:fundamentals/dependency-injection) container. Its purpose is to provide an alternative to constructor injection for when you need a service only if a particular method is called.
 
+> [!TIP]
+> The binding will fail when obtaining an instance of a type, **not registered**, from the dependency injection container. If the service is expected to be **optional** then the parameter must be marked as nullable, including `?`, or set a default value. In this case, the method is responsible for the required check before the usage.
```
Rick-Anderson, line 629:
```suggestion
> [!TIP]
> Binding fails when attempting to bind to a type that isn't registered with the dependency injection container or isn't a .NET type. If the service is **optional**,  the parameter must be one of the following:
>* Marked as [nullable](/dotnet/csharp/language-reference/builtin-types/nullable-value-types).
>* Have a default value.
>
> When using optional parameters, the app must check if the parameter is available.
```

This might be closer. @serpent5  can you help out.

serpent5, line 629: How about something like this?:

```suggestion
If an instance of the type isn't registered in the dependency injection container, the app throws an exception when attempting to bind the parameter. To make the parameter optional, use one of the following options:

* Make the parameter nullable.
* Set a default value for the parameter.

For nullable parameters, ensure that the parameter isn't `null` before accessing it.
```

serpent5: I've made an alternative suggestion below. I don't think this needs to be a tip, and I don't think we should link to the nullable docs as they're specific to value-types.

Rick-Anderson (approved):

PR #25130, opened 2022-02-24
Blazor Hybrid routing and nav topic

guardrex:
Addresses #24956

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/hybrid/routing?view=aspnetcore-6.0&branch=pr-en-us-25130)

Notes

* These will be *rough* at first. I'll circle around later for niceties like API doc cross-links (when available), enhanced descriptions, fully-working, cut-'n-paste code examples, etc.
* Although this PR doesn't include routing coverage, I think we'll probably be adding at least a brief remark that component routing (in a `BlazorWebView`) conforms to the coverage in the Blazor *Fundamentals* node topic on the subject with any caveats or *gotchas* 😈 that you want to call out. [*Fundamentals* topic on routing/nav](https://docs.microsoft.com/aspnet/core/blazor/fundamentals/routing) ... **_Does it all apply to Hybrid scenarios (`BlazorWebView`)?_**

PR #25134, opened 2022-02-24
Move namespace remarks

guardrex:
Addresses https://github.com/dotnet/AspNetCore.Docs/pull/25130#discussion_r814367350

I'll do a bit more work on this on the next pass. Eventually, it will say what API requires it.

cc: @TanayParikh


TanayParikh (approved):
Thanks!

PR #25143, opened 2022-02-25
Update hosting-models.md

achingono:


guardrex (approved):
Thanks @achingono! :rocket: ... A sneaky 😈 got me on that one! 😄

PR #25152, opened 2022-02-26
Update anti-request-forgery.md

ptakpiotr:
Listed also Startup.Configure where the MapRazorPages and MapBlazorHub extension methods will be called on app object


Rick-Anderson (approved):
@serpent5  looks right to me.

serpent5:
@Rick-Anderson The change looks fine given the context, but I don't think the original coverage here was correct. I would expect the listed methods to be `AddControllers` and `AddRazorPages`, which are called in `Startup.ConfigureServices`. I don't think that e.g. `MapRazorPages` can add services to DI, as the container has already been built at this stage.

serpent5:
Let's merge this as-is, given that the original coverage was approved by PU and the change looks good based on that.

serpent5:
Thanks for your contribution, @ptakpiotr. Be sure to check [What's New](https://docs.microsoft.com/en-us/aspnet/core/whats-new) early next month, where you'll see yourself listed as a valued contributor.

PR #25223, opened 2022-03-07
Clarify component disposal async+sync

### LINKED ISSUE #24863: Component disposal with IDisposable and IAsyncDisposable - Notes Behavior incorrectly/unclearly
The doc notes:

>If a component implements IDisposable, IAsyncDisposable, or both, the framework calls for unmanaged resource disposal when the component is removed from the UI. Disposal can occur at any time, including during [component initialization](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?view=aspnetcore-6.0#component-initialization-oninitializedasync).

However the behavior in the [render](https://github.com/dotnet/aspnetcore/blob/main/src/Components/Components/src/RenderTree/Renderer.cs#:~:text=disposeComponentState.Component%20is%20IAsyncDisposable) is to only execute one method and to prioritize IAsyncDisposable, meaning you can have both IAsyncDisposable and IDisposable implemented but the framework will not call both.

---
#### Document Details

* ID: a970fb72-3dd0-d173-585d-2ccdd7923f81
* Version Independent ID: 4871f4f7-9cf6-ac7a-0f21-c65d34bd1119
* Content: [ASP.NET Core Razor component lifecycle](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?view=aspnetcore-6.0#component-disposal-with-idisposable-and-iasyncdisposable)
* Content Source: [aspnetcore/blazor/components/lifecycle.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/lifecycle.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #24863

Thanks @jkdba! :rocket: ... Looks like Pranav isn't on the PU team any longer. This should be fine tho given that it comes directly from reference source.

Ref source code comment:

> Components shouldn't need to implement IAsyncDisposable and IDisposable simultaneously,
> but in case they do, we prefer the async overload since we understand the sync overload
> is implemented for more "constrained" scenarios.
> Component authors are responsible for their IAsyncDisposable implementations not taking
> forever.

Cross-reference: https://github.com/dotnet/aspnetcore/blob/main/src/Components/Components/src/RenderTree/Renderer.cs#L1009-L1013

### DIFF
--- a/aspnetcore/blazor/components/lifecycle.md
+++ b/aspnetcore/blazor/components/lifecycle.md
@@ -218,6 +218,10 @@ Although the content in this section focuses on Blazor Server and stateful Signa
 
 If a component implements <xref:System.IDisposable>, <xref:System.IAsyncDisposable>, or both, the framework calls for unmanaged resource disposal when the component is removed from the UI. Disposal can occur at any time, including during [component initialization](#component-initialization-oninitializedasync).
 
+Components shouldn't need to implement <xref:System.IDisposable> and <xref:System.IAsyncDisposable> simultaneously. If both are implemented, the framework only executes the asynchronous overload.
+
+Developer code must ensure that <xref:System.IAsyncDisposable> implementations don't take a long time to complete.
+
 ### Synchronous `IDisposable`
 
 For synchronous disposal tasks, use <xref:System.IDisposable.Dispose%2A?displayProperty=nameWithType>.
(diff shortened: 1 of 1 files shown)


jkdba:
Thanks!

Issue #25243, opened 2022-03-08
Standardize topic titles and TOC for Introduction or Overview


wadepickett:
This issue tracks changes we would make to aspnetcore.docs, but may be linked to a larger standardization for .NET repos.

aspnetcore.docs has topics that either use _Introduction_, _Overview_ or neither in the title that provide a top-level overview or introduction to product and sub-areas. At the same time those topics use one of those options for the TOC entry and the other for the actual topic title.    

It would be best to standardize on one way to refer to these topics.  In the title and the TOC so customers know what to look for and do not second guess when they click on a toc link that end up at a topic with a different title.  We were also seeing heavier bounce (customers arriving and then leaving right away for some reason) on overview type topics that used "Introduction" in their title, which is also used for some .NET tutorial topics across repos. Those same tutorial topics were not getting high bounce even in the case of SignalR where the same title was used for an overview type and a tutorial type topic.

- [ ] Check on current guidance for docs.microsoft.com
- [ ] Verify what seems to be working in the .NET and ASP.NET Core repos, including keywords customers are using in search.
- [ ] If we can't, for some reason, use a standard that is in play across docs.microsoft.com, see if there can be one for development focused doc, and if not that, then the VS, .NET, Azure docs/learn sphere that .NET developers move around in, and if not that, then .NET repos.

If we can't get moving on a broader Standarization, at least get aspnetcore.docs fixed for this.

guardrex:
I have two Blazor "Overviews" that use the word "Introduction," and I can easily/quickly update them to "Overview." I do use "Overview" in the ToC for both ... and everything else for that matter.

Shall I go ahead with that update? It will take like **_2 minutes_** ⏱️🏃 to update.

wadepickett:
I need to still fix:
[Introduction to gRPC on .NET](https://docs.microsoft.com/en-us/aspnet/core/grpc/?view=aspnetcore-6.0)

[Overview of ASP.NET Core authentication](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/?view=aspnetcore-6.0) shows as "Overview" in TOC but we have another overview just below it for "Introduction to Identity".  Change the toc to "Overview of authentication" and "Overview of Identity" and update the titles.

[...]

PR #25244, opened 2022-03-08
Blazor "Introduction" to "Overview"

guardrex:
Addresses #25243

Wade, I use "Overview" everywhere, including all of the ToC entries, except these two spots on the PR. AFAICT, these are the only Blazor spots to touch for your issue.

### DIFF
--- a/aspnetcore/blazor/hybrid/index.md
+++ b/aspnetcore/blazor/hybrid/index.md
@@ -1,5 +1,5 @@
 ---
-title: Introduction to ASP.NET Core Blazor Hybrid
+title: ASP.NET Core Blazor Hybrid overview
 author: guardrex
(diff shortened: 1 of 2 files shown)


wadepickett (approved):
Thanks!  Looks good.  I was going to check the .NET repo first to see what they were doing.  Looking at it now, it is usually Overview but not always.  Looking at VS, it is overview in the TOC but then "About" on the title.

wadepickett:
@Rick-Anderson, you have any preferences on Overview vs Introduction?  I wasn't going to change anymore besides the ones that had bad bounce for the SEO work until looking at more data, but maybe it is a clear choice for now.  See #25243

guardrex:
An alternative is to drop the word, so these two highest-level topics would simply be ...

* ASP.NET Core Blazor
* ASP.NET Core Blazor Hybrid

🎤 *drop* ... the word "overview" doesn't add much here unless u know that it's a common search term.

Issue #23704, opened 2021-11-01
Why render/not render when having/not having ChildContent set?


omuleanu:
for this line:
"The second Expander component's Expanded value isn't reset because no child content is rendered in the second component."

I don't understand why is that when the `Expander.ChildContent` property is set, the component is rendered when calling StateHasChanged on the parent

what's the link between a component having `ChildContent` set and getting rendered on parent's `StateHasChanged`

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 81a32115-33f8-7d06-aaf0-b8ce918b5527
* Version Independent ID: 31ad18df-1904-bbdd-cfed-8d0d0267c10e
* Content: [ASP.NET Core Razor components](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-5.0#overwritten-parameters-1)
* Content Source: [aspnetcore/blazor/components/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/index.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
We can explore fleshing it out further. I'll ping Steve later this month to look and see if he'll supply some additional information. Leave this issue open for now.

guardrex:
**UPDATE**: Sorry for the delay. It wasn't possible to converse with Steve on this prior to the holiday season. We'll take a look at this in 22Q1 when everyone gets back 🏖️.

SteveSandersonMS:
The diffing system re-renders child components if their parameters might have changed.

 * For parameters of known immutable primitive types (strings, bools, ints, and a few others), the diffing system explicitly checks if they have changed and can skip re-rendering this child if none of them have. That's what's happens with your Expander component that only receives a bool.
 * For all other parameter types, the framework can't know if the value has mutated in some way, so it has to re-render the child. For example your Expanded that takes a ChildContent - this is of type RenderFragment which is a delegate that refers to other mutable objects (which logically makes sense - the render fragment's output can change over time), so in this case the framework does re-render that one.

PR #25259, opened 2022-03-09
Update overwritten param examples

### LINKED ISSUE #23715: Expanded property public is only used once privately
in the second `Shared/Expander.razor` (fixed)
the property `Expanded` has a public `get` which is only used once in `OnInitialized`
and it makes no sense to use it after because it won't have the value of the private `expanded`
so we can make it `private get`, because there's no point/intention to use it for any other reason.

---
#### Document Details

* Content: [ASP.NET Core Razor components](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-5.0#overwritten-parameters-1)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #23715

Thanks @omuleanu! 🚀

I haven't forgotten about #23704, but I'll have to get back to that one later. I'd like Steve to answer it, but I still can't ping him on that one at the moment given the workload of high priority issues. It's tracked, and we'll get to it *ASAP* 🏃.

WRT this update: You originally only called out the revised `Expander` (the "good" one) for the update mentioning that it's only used once in `OnInitialized` (as far as parent/children would be concerned). There's no problem applying the update to the "bad"-behaving one, too, so I do that here as well.

Also, I'd like to ask Tanay about this. We don't really sweat private getters. It's when the setter must be restricted that the access modification is particularly relevant. We might not want to make this change at all. I defer to a **_real programmer_** on this point 😄, but I don't see us modifying getter access like this very often, if at all.

# 🇺🇦

### DIFF
--- a/aspnetcore/blazor/samples/3.1/BlazorSample_Server/Shared/index/BadExpander.razor
+++ b/aspnetcore/blazor/samples/3.1/BlazorSample_Server/Shared/index/BadExpander.razor
@@ -11,7 +11,7 @@
 
 @code {
     [Parameter]
-    public bool Expanded { get; set; }
+    public bool Expanded { private get; set; }
 
     [Parameter]
     public RenderFragment ChildContent { get; set; }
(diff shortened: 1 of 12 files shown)


TanayParikh (approved):
Sure, I don't see any harm in making the getter private in this case. 

My concern would be more related to accidentally using the `Expanded` property in your component code, when you should be using the `expanded` field. Granted, the access modifier of the property getter doesn't have any impact on this behavior.

PR #25289, opened 2022-03-11
Update grpc-start.md

krsche:
If the syntax version is not provided, latest Visual Studio will fail on compile with error: 
```
\altsrc\github\grpc\workspace_protoc_windows_x64\third_party\protobuf\src\google\protobuf\compiler\parser.cc:649] No syntax specified for the proto file: Protos/greet.proto. Please use 'syntax = "proto2";' or 'syntax = "proto3";' to specify a syntax version. (Defaulted to proto2 syntax.)
```


krsche:
don't mind this PR, have to read more carefully 😅 (it says copy the proto file over from the server, there this is already set)

PR #25293, opened 2022-03-11
Add initial .editorconfig and extra markdownlint rules

### LINKED ISSUE #7487: Add .editorconfig file to enforce consistency in ASP.NET Core sample apps
See https://github.com/aspnet/Docs/pull/7474#discussion_r200680276

serpent5:
Fixes #7487.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,23 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = crlf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+max_line_length = off
+trim_trailing_whitespace = false
+
+[*.cs]
+indent_size = 4
+dotnet_sort_system_directives_first = true
+
+[*.cshtml]
+indent_size = 4
+
+[*.razor]
+indent_size = 4
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -15,6 +15,9 @@
     "hr-style": {
         "style": "---"
     },
+    "emphasis-style": {
+      "style": "asterisk"
+    },
     "line-length": false,
     "list-indent": true,
     "list-marker-space": true,
@@ -61,5 +64,11 @@
     },
     "required-headers": false,
     "single-h1": false,
-    "ul-start-left": false
+    "strong-style": {
+      "style": "asterisk"
+    },
+    "ul-start-left": false,
+    "ul-style": {
+      "style": "asterisk"
+    }
 }


guardrex:
The default rule (for code) ...

> trim_trailing_whitespace = true

I haven't been favorable to killing a space at the end of a code line when the break is merely present to keep code line lengths from getting too long for docs. I think it's nice for some devs who cut-'n-paste code and then collapse lines. However, this is a good rule in general outside of that, so I suppose that I'd just ignore the warning if this is kept.

Indentation: How about? .....

```
[*.html]
indent_size = 4

[*.css]
indent_size = 4

[*.js]
indent_size = 2

[*.xml]
indent_size = 2

[*.json]
indent_size = 2
```

> max_line_length = off

Can this be turned on for code files and set to 85 chars?

serpent5:
> I haven't been favorable to killing a space at the end of a code line when the break is merely present to keep code line lengths from getting too long for docs.

I don't mind if we just disable this rule, but what does trailing whitespace give us in the scenario where the lines are split for docs presentation?

> Indentation: How about? .....

The default is `indent_size = 2`, so we wouldn't need entries for `.js`, `.xml`, or `.json` in this case. For `.css`, the non-Blazor project templates use `2`, but I see that the Blazor templates and samples use `4`. That might be something for a Blazor-specific override, I guess.

> Can this be turned on for code files and set to 85 chars?

Sure. I think we can add `max_line_width = 85` for the relevant extensions.

guardrex:
> what does trailing whitespace give us

In markdown, two spaces at the end of a line result in DocFX inserting a bare return `<br>`, and we use that in spots where needed.

> we wouldn't need entries for

:+1:

> the non-Blazor project templates use 2

They're apparently adopting it along with the rest of the world. I let them know on their Blazor project template issue and attributed it to you (by name so that you wouldn't get pinged to 💀😄).

Forget about Blazor then ... I recommend two spaces for the future. They'll *very likely* adopt it for Blazor in 7.0+.

serpent5:
> > what does trailing whitespace give us
> 
> In markdown, two spaces at the end of a line result in DocFX inserting a bare return `<br>`, and we use that in spots where needed.

Ahh, OK, gotcha. That's turned off for `.md` files with `trim_trailing_whitespace = false` for the reason you describe.

Rick-Anderson (approved):

guardrex:
> That's turned off for `.md` files with `trim_trailing_whitespace = false` for the reason you describe.

Doesn't cover the C# file case. Do we always want to delete the space at the end of an artificial line break? I speculate that devs will be re-assembling these multiple-line statements back to a single line for many code samples. If that's correct, it's probably helpful to leave the space on the ends of these broken lines.

guardrex:
btw -- Was the decision not to go with the 85 char code line warning? It's not a big deal ... the char count for the line in VSC shows how long they are. It might just shave a hair of time off tho if it's ✨ *automagical* ✨ while one types.

serpent5:
> Was the decision not to go with the 86+ char code line warning?

I didn't decide anything here. I dunno how well the limit would work if there's a few levels of indentation that's got a snippet wrapper, for example. I think I'll try it locally and see how it plays out, before adding it to the repo. I don't mind just adding it though if you'd prefer that.

What do you think? You can even add it to a Blazor-specific `.editorconfig` if you want to commit to it.

guardrex:
Yeah, I see what you mean. Nevermind on that I think. It's just not going to work as well as I hoped, especially in the snippet case. Many (most really AFAIK) of the code examples are snippets in Blazor docs. It would have limited usefulness.

PR #25302, opened 2022-03-13
Fix request-response mixup

erhanalankus:
[Internal Review](https://review.docs.microsoft.com/en-us/aspnet/core/web-api/?view=aspnetcore-6.0&branch=pr-en-us-25302)

It seems like the word "request" was used instead of "response" in these two places.


wadepickett (approved):
Looks good, thanks!

Issue #25155, opened 2022-02-27
.NET 6:  Host ASP.NET Core SignalR in background services


Sethox:
[This page ](https://docs.microsoft.com/en-us/aspnet/core/signalr/background-services?view=aspnetcore-6.0#enable-signalr-in-startup), has not been updated for ASPNetCore 6.

**EDIT by @Rick-Anderson**

Consider adding SLN file. [.gitignore](https://github.com/dotnet/AspNetCore.Docs/blob/main/.gitignore#L15-L17) prevents `.sln` files from being checked in, so follow the instructions to force check in of `.sln` file.

```
# Use git add -f to force override .sln when required. Not needed in most cases.
# git add -f myProj.sln
```

Add readme.md file explaining what each project is for, how to test, etc.

See https://github.com/dotnet/AspNetCore.Docs/pull/25326#issuecomment-1069004348

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1bbb2d1c-4dfe-3faf-d4b0-65ae1146961e
* Version Independent ID: 6b203f18-adfa-3a18-183d-b124325064c3
* Content: [Host ASP.NET Core SignalR in background services](https://docs.microsoft.com/en-us/aspnet/core/signalr/background-services?view=aspnetcore-6.0)
* Content Source: [aspnetcore/signalr/background-services.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/signalr/background-services.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-signalr**
* GitHub Login: @bradygaster
* Microsoft Alias: **bradyg**

PR #25326, opened 2022-03-16
Create Server.sln

Rick-Anderson:
Contributes to #25155

@bradygaster  [.gitignore](https://github.com/dotnet/AspNetCore.Docs/blob/main/.gitignore#L15-L17) prevents `.sln` files from being checked in. Did you intend for [all 4 of these projects](https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/signalr/background-service/samples/3.x) to be in one SLN file?

How do I replace [Clients.ConsoleTwo.Program.Main](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/signalr/background-service/samples/3.x/Clients.ConsoleTwo/Program.cs) in .NET 6?

cc @UncleDave @BrennanConroy 

image

### DIFF
--- a/aspnetcore/signalr/background-service/samples/3.x/Server/Server.sln
+++ b/aspnetcore/signalr/background-service/samples/3.x/Server/Server.sln
@@ -1,0 +1,43 @@
+
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.1.32228.430
+MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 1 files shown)


UncleDave:
@Rick-Anderson I can take #25155 if you like, but I'm not sure what you're asking here?

Rick-Anderson:
> @Rick-Anderson I can take #25155 if you like, but I'm not sure what you're asking here?

That would be great. I fixed all the monikers in #25327 so you won't have any monikers to deal with. Just update aspnetcore/signalr/background-services.md between  Line 21 `:::moniker range=">= aspnetcore-6.0"` and line 83 `:::moniker-end` and add the .NET 6 code.

The sample has 4 projects. If you're using Visual Studio, it's usually easier to have them all in one solution file.

PR #25335, opened 2022-03-16
Update markdown syntax

CGSILVA:
Update markdown syntax to follow the pattern for file names


guardrex (approved):
Thanks @CGSILVA! :rocket:

The build system is having trouble right now. We'll get this merged as soon as we can.

PR #25380, opened 2022-03-22
Expand on string vs object for HttpContext.Items

### LINKED ISSUE #25376: Possible incorrect statment.
In the section "HttpContext.Items", this sentence doesn't seem right ... "For middleware that's only used in a single app, fixed string keys are acceptable. Middleware shared between apps should use unique object keys to avoid key collisions." It seems to me that any middleware needs to use a unique key so that it won't conflict with any other middleware in the same app. What does some other app have to do with collision around this object?

---
#### Document Details

* ID: 76e22150-f1f1-5e4f-0e97-39b1ac089297
* Version Independent ID: 687172c3-b7d8-525d-39dc-658b4a0f526f
* Content: [Session in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/app-state?view=aspnetcore-6.0#httpcontextitems)
* Content Source: [aspnetcore/fundamentals/app-state.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/app-state.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

serpent5:
Fixes #25376.

### DIFF
--- a/aspnetcore/fundamentals/app-state.md
+++ b/aspnetcore/fundamentals/app-state.md
@@ -4,7 +4,7 @@ author: rick-anderson
 description: Discover approaches to preserve session between requests.
 ms.author: riande
 ms.custom: mvc
-ms.date: 11/25/2021
+ms.date: 03/22/2022
 no-loc: ["Blazor Hybrid", Home, Privacy, Kestrel, appsettings.json, "ASP.NET Core Identity", cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
 uid: fundamentals/app-state
 ---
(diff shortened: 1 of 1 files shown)


aspnetcore/fundamentals/app-state.md
```diff
@@ -240,16 +240,14 @@ In the following example, [middleware](xref:fundamentals/middleware/index) adds
 
 [!code-csharp[](app-state/6.0samples/SessionSample/Program.cs?name=snippet_hci)]
 
-For middleware that's only used in a single app, fixed `string` keys are acceptable. Middleware shared between apps should use unique object keys to avoid key collisions. The following example shows how to use a unique object key defined in a middleware class:
+For middleware that's only used in a single app, it's unlikely that using a fixed `string` key would cause a key collision. However, to avoid the possibility of a key collision altogether, an `object` can be used as an item key. This approach is particularly useful for middleware that's shared between apps and also has the advantage of eliminating the use of key strings in the code. The following example shows how to use an `object` key defined in a middleware class:
```
Rick-Anderson, line 243: Much better!!!!

Rick-Anderson (approved):

JamesKiddSmyth:
Thanks for your response!  I'm still not sure why the emphasis seems to be on "shared between apps".  The only thing i can think is that if it is shared by apps, then it is more _likely_ to be included with other middleware that might try to use the same key string.  It might be me, but I think the verbiage used implies that there is some cross-app concern, rather than that the underlying issue is HttpContext itself, specifically an instance per request,  being shared among middleware within the same app.   Sorry if it sounds pedantic, but i want to make sure I'm not missing something more fundamental.  

One other thing, particularly because it includes this section on HttpContect.Items, is there any reason this document doesn't have at least a brief section on using (scoped, probably) DI of objects to share state between middleware?   

Thanks again.

PR #25381, opened 2022-03-22
Enhance example description

### LINKED ISSUE #25330: [Blazor] Does this example have a memory leak?
[EDIT by guardrex ... Just a quick note on the question: This refers to "**_Razor components_**," not Razor Pages pages, which I'm sure is what is meant by the question. I'm just noting it for the future in case a dev finds this issue and isn't sure exactly what this question is about.]

The "/reference-parent-3" surely has a memory leak because it never clears the list? Should it clear it out at the top of the razor page **_[sic]_**?

---
#### Document Details

* Content: [ASP.NET Core Razor components](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-5.0#capture-references-to-components-1)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #25330

Thanks @mrpmorris! 🎸

I'm going to add a remark to the example to call it out. It's a managed list, and it should be GC'ed along with anything else managed by the CLR AFAIK. Let's not get too deep into the weeds here with this example on the whole subject, because ...

* We'd end up on a slippery slope where **_many things_** get a "this is managed, don't sweat disposal"-type of remark. We prefer to go the other way and mostly call it out where object disposal matters.
* The whole subject of object management and GC is covered in great detail by the .NET docs on memory management.
* We call out spots where disposal is important in the section that's dedicated to disposal, [Component disposal with `IDisposable` and `IAsyncDisposable`](https://docs.microsoft.com/aspnet/core/blazor/components/lifecycle#component-disposal-with-idisposable-and-iasyncdisposable).

Still tho, I think this is a good spot to add a brief remark as you suggest. Actually, I'm even going to add a quick cross-link down to that disposal section of this topic.

IIRC, the PU has told devs over the years that *potential* memory leaks time-consuming to investigate; so when someone opens a 'I think I have a memory leak' issue, the PU will close it if the dev can't repro the leak clearly. If this example ... the one that you're calling out for this update ... generates a memory leak, for sure, please do open an issue for the product unit and feel free to CC: me on it so that I can follow the discussion. Just fair warning that you'll probably need clear evidence of the leak for them to proceed.

Thanks again for opening the issue.

### DIFF
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -662,6 +662,7 @@ Use a collection to reference components in a loop. In the following example:
 
 * Components are added to a <xref:System.Collections.Generic.List%601>.
 * A button is created for each component that triggers the corresponding component's `ChildMethod` by its component index in the <xref:System.Collections.Generic.List%601>.
+* There's no need to clear the list of component references in developer code. The <xref:System.Collections.Generic.List%601> is a [managed resource](/dotnet/standard/managed-code) and cleared by the [.NET garbage collector](/dotnet/standard/garbage-collection/) automatically when the framework disposes of the component. 
 
 `Pages/ReferenceParent3.razor` using the preceding `ReferenceChild` component:
 
@@ -2090,6 +2091,7 @@ Use a collection to reference components in a loop. In the following example:
 
 * Components are added to a <xref:System.Collections.Generic.List%601>.
 * A button is created for each component that triggers the corresponding component's `ChildMethod` by its component index in the <xref:System.Collections.Generic.List%601>.
+* There's no need to clear the list of component references in developer code. The <xref:System.Collections.Generic.List%601> is a [managed resource](/dotnet/standard/managed-code) and cleared by the [.NET garbage collector](/dotnet/standard/garbage-collection/) automatically when the framework disposes of the component.
 
 `Pages/ReferenceParent3.razor` using the preceding `ReferenceChild` component:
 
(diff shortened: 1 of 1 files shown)


mrpmorris:
You will definitely hold on to stale references (and therefore leak) if you don't remove items from that list when the element they were rendered for is removed from the source.

The component rendered would need to implement IDisposable and have a Disposed event that the parent could hook into in order to remove it's UI reference from the list.

I think it's a much larger topic than it initially seems to be, so either needs to be fleshed out fully or perhaps not mentioned (and let StackOverflow explain how to do it)?

guardrex:
No, I've emailed Steve, and he'll know what to do with this.

Like I said, there was nothing mentioned on PU issues about this, nor did any PU engineer remark that we were lacking coverage on this aspect when the subject of component disposal came up several times over the years. We documented everything that the PU called for that a dev should be concerned about, and this wasn't one of those aspects.

I merely *interpreted* that this list is managed. If it is managed, then the GC will free it *auto-magically* ✨ ... **_BUT_** also keeping in mind that the app can't be running with the debugger and must be run in the Production environment, per PU remarks ... Javier to exact.

Let's let Steve catch up to us on this. I'm definitely 🛑 **NOT** 🛑 merging this without his approval. *Indeed!* He might say that I misinterpreted prior work and remarks in this area. I won't feel bad about it if I did. Like I said, it never came up. They never asked for further coverage.

Stand-by ... We should get some Steve superhero 🦸 assistance here soon.

guardrex:
I stand corrected, @mrpmorris ... **_sort of_** in that you're correct that building up the list of elements isn't cleared on **_re-rendering_**. However, the whole list is GC'ed as a managed object. Steve asked me to remove the example ... not so much because it doesn't technically work to illustrate the main concept or couldn't be the basis of something that would remove the list entries and thus not demand memory on re-renders without removal of components in the list ... Steve says that this pattern should be generally dissuaded. He wants devs to just use collections of their business objects, not component instances. It mirrors his remarks from one of the issues on this subject at https://github.com/dotnet/aspnetcore/issues/17361#issuecomment-558138782. I'll remove the example now. Thanks for your help with this, @mrpmorris.

mrpmorris:
Yes, I agree. This is a good decision.

PR #25400, opened 2022-03-23
fix text error

ruslan528:


guardrex (approved):
Thanks @ruslan528! :rocket:

PR #25401, opened 2022-03-23
Added instructions for header key Content-Type

tiangco:
Set Content-Type to "application\json" - other wise example returns 415 Unsupported Media Type


serpent5:
Hi @tiangco. Thanks for taking the time to open a Pull Request.

I've just tested this out and found that setting `Content-Type` isn't necessary for this example. After you follow the instructions in the article, Postman automatically sets the `Content-Type` header when you add the JSON payload for the request body.

serpent5:
Did you set the body as `JSON`?:

image

tiangco:
Right! I missed that part.
Thanks for setting this straight.

PR #25404, opened 2022-03-24
Remove commented out bullet

jongalloway:
This is just showing a blank bullet since the content is commented out. Since it's not linkable in the release notes, I suggest just removing the list item.


serpent5:
Thanks @jongalloway.

PR #25437, opened 2022-03-26
Change example code to .NET 6 style

vanillajonathan:


guardrex:
Hello @vanillajonathan ... This would need to be versioned in order for us to accept it because if the topic loads for 5.0 the 6.0 code shown won't be appropriate, nor can we just set the topic to 6.0 or later because we should maintain coverage back to earlier versions. Do you want to update the PR using our versioning? If so, do this ...

Make two FULL copies of the doc content after the H1 heading in `monker range` ... `monkier-end` versioning blocks ...

```
---
title: Multi-factor authentication in ASP.NET Core

... Additional metadata entries here ... leave the metadata as it is ...

uid: security/authentication/mfa
---
# Multi-factor authentication in ASP.NET Core

::: moniker range=">= aspnetcore-6.0"

... Place a copy of the entire document (after the H1) here WITH YOUR 6.0 CODE UPDATES ...

::: moniker-end

::: moniker range="< aspnetcore-6.0"

... Place a copy of all of the existing content (after the H1) here (this will retain the current 3.1/5.0 code) ...

::: moniker-end
```

guardrex (changes requested):
See remarks :point_up:.

vanillajonathan:
Thank you @guardrex, I close this PR and opened the issue #25440.

PR #25438, opened 2022-03-26
New no-loc test

guardrex:
# 🇺🇦

### DIFF
--- a/aspnetcore/docfx.json
+++ b/aspnetcore/docfx.json
@@ -65,6 +65,9 @@
         "**/getting-started/**/**.md": "tutorial",
         "**/tutorials/**/**.md": "tutorial"
       },
+      "no-loc": {
+        "**/migration/no-loc-test.md": [ "computer", "cat" ]
+      },
       "recommendations": {
         "**/tutorials/**/**.md": "false"
       }
--- a/aspnetcore/migration/no-loc-test.md
+++ b/aspnetcore/migration/no-loc-test.md
@@ -1,0 +1,49 @@
+---
+title: Localization test doc
+author: guardrex
+description: This is a localization test doc.
+monikerRange: '>= aspnetcore-3.1'
+ms.author: riande
+ms.custom: mvc
+ms.date: 03/26/2022
+uid: migration/no-loc-test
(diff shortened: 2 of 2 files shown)


guardrex:
cc: @khoiph1 ... Khoi, I'm going to try this again now that Kirk got me straight on what a JSON array is! 🙈😆 **_DUH!_** I told them ... I just told Dan and Artak that I'm exhausted 😩 and must take a vacation 🏖️🍹 within the next couple of weeks to recharge.

Here's the live English-US doc ...

* https://docs.microsoft.com/en-us/aspnet/core/migration/no-loc-test?view=aspnetcore-6.0

... and then after MT runs, I'll check on ...

* https://docs.microsoft.com/es-es/aspnet/core/migration/no-loc-test?view=aspnetcore-6.0
* https://docs.microsoft.com/de-de/aspnet/core/migration/no-loc-test?view=aspnetcore-6.0
* https://docs.microsoft.com/fr-fr/aspnet/core/migration/no-loc-test?view=aspnetcore-6.0

... and see what happens. If it works this time, I'll change the `docfx.json` to reflect our need for a global **_ARRAY_** 😄 of `no-loc` entries.

Use of `fileMetadata` is fine for the test because we want to scope the test to just the `no-loc-test.md` file. For the full repo tho, we should use `globalMetadata` if this ends up working out. I propose to use the following ...

```json
"globalMetadata": {
  ... existing entries ...
  "no-loc": [ "entry1", "entry2," ... "entryX" ],
  ... existing entries ...
}
```

We also want to keep doc-specific `no-loc` words. When I make the full repo update, I plan to filter those `no-loc` entries and leave them **_with only the doc-specific entries_**. It does lead to a BIG ❓ tho ... **_Are they going to merge cleanly into the final doc?_** If nobody knows if that's going to happen, then I need to run a follow-up test: I'll add our existing `no-loc` metadata entry to my `no-loc-test.md` doc and then see if the test `fileMetadata` entry and the doc's metadata list **_combine_** and prevent localization of the list at the top of the doc **_AND_** prevent loc of my "computer" and "cat" words.

PR #25496, opened 2022-04-01
confusing overriding of variables

pkellner:
I see this a lot and I always think it's a bad practice. Not sure "b" is the right choice,but "builder" is definitely one of the few wrong choices.

### DIFF
--- a/aspnetcore/security/cors/6.0sample/Cors/WebAPI/Program.cs
+++ b/aspnetcore/security/cors/6.0sample/Cors/WebAPI/Program.cs
@@ -9,9 +9,9 @@ var builder = WebApplication.CreateBuilder(args);
 builder.Services.AddCors(options =>
 {
     options.AddPolicy(name: MyAllowSpecificOrigins,
-                      builder =>
+                      b =>
                       {
-                          builder.WithOrigins("http://example.com",
+                          b.WithOrigins("http://example.com",
                                               "http://www.contoso.com");
                       });
 });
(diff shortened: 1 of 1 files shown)


serpent5:
We don't use single-letter names in the docs, so I suggest `policy` as something that's consistent with e.g. [the authz policies article](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/policies?view=aspnetcore-6.0).

serpent5 (commented):
We don't use single-letter names in the docs, so I suggest `policy` as something that's consistent with e.g. [the authz policies article](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/policies?view=aspnetcore-6.0).

aspnetcore/security/cors/6.0sample/Cors/WebAPI/Program.cs
```diff
@@ -9,9 +9,9 @@
 builder.Services.AddCors(options =>
 {
     options.AddPolicy(name: MyAllowSpecificOrigins,
-                      builder =>
+                      b =>
```
Rick-Anderson, line 12:
```suggestion
                      policy  =>
```

aspnetcore/security/cors/6.0sample/Cors/WebAPI/Program.cs
```diff
@@ -9,9 +9,9 @@
...
                       {
-                          builder.WithOrigins("http://example.com",
+                          b.WithOrigins("http://example.com",
```
Rick-Anderson, line 14:
```suggestion
                          policy.WithOrigins("http://example.com",
```

Rick-Anderson (changes requested):

serpent5 (approved):

Rick-Anderson (approved):

pkellner:
Wow!  Thank you for fixing this.

Rick-Anderson:
@pkellner  thanks for pointing it out. You get the commit credit.

PR #25506, opened 2022-04-04
What's New in March 2022

serpent5:
The GitHub Workflow failed, due to what appears to be an issue with file URLs being escaped when returned from GitHub. Perhaps this has changed within the last month, but it doesn't appear to have affected the dotnet/docs repo. That uses relative links to articles, whereas we use xrefs, so I expect it could be something to do with that.

Anyway, I fixed up the code to unescape the URLs and it works locally, so here's the content generated from that.

[Internal Review](https://review.docs.microsoft.com/en-us/aspnet/core/whats-new/dotnet-aspnetcore.docs-2022-03-01?view=aspnetcore-6.0&branch=pr-en-us-25506).


Rick-Anderson:
@serpent5 can you update https://github.com/Rick-Anderson/Report/blob/master/kt.md#whats-new-in-docs and outline what you had to do. I'm guessing it's going to keep failing. How hard would it be to fix?

serpent5:
This is fixed in the latest version of the _whatsnew_ tool. I've updated our repo to use the latest version in #25513.

PR #25519, opened 2022-04-05
Corrected Code Execution Reference

mitchelsellers:
The referenced code sample had the same comment both before and after the call to await next.Invoke();

However, you CAN write to the response before you call next.invoke, so I updated the comment to reflect the proper control flow.

I did not create an issue for this, but can if needed.


Rick-Anderson (approved):

Rick-Anderson:
Thanks for the fix. Be sure to check out your name as a contributor about the 3rd of next month on the [What's new page](https://docs.microsoft.com/en-us/aspnet/core/whats-new/) in the **Community contributors** section at the bottom.

Issue #25550, opened 2022-04-07
Inconsistent diff sample for appsettings


baseTwo:
The bullet (below) mentions that |Microsoft.Hosting.Lifetime| should be removed from the appsettings, but the diff example underneath shows erroneously that "Microsoft":"Warning" should be removed too

* `"Microsoft.Hosting.Lifetime": "Information"` was removed from both `appsettings.json` and `appsettings.Development.json`:

```diff
- "Microsoft": "Warning", <---- This should be removed
- "Microsoft.Hosting.Lifetime": "Information"
+ "Microsoft.AspNetCore": "Warning"
```

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 5bec61f9-7298-170d-5e12-abb27cde584e
* Version Independent ID: 27a43e3f-149d-9058-f937-0eda1d6dd4f2
* Content: [Migrate from ASP.NET Core 5.0 to 6.0](https://docs.microsoft.com/en-us/aspnet/core/migration/50-to-60?view=aspnetcore-6.0&tabs=visual-studio)
* Content Source: [aspnetcore/migration/50-to-60.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/migration/50-to-60.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-migration**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
@wadepickett  can you test this? I think the diff is correct but the wording is not completely accurate.

wadepickett:
@Rick-Anderson.  Yes, `"Microsoft.AspNetCore": "Warning"` is the new default in the templates. It replaces the previous defaults of:
`"Microsoft": "Warning",`
`"Microsoft.Hosting.Lifetime": "Information`

Specifying "Microsoft.AspNetCore": "Warning" instead is about namespace level control for log level.  It still allows all the information level messages from the `Microsoft` namespace such as `Microsoft.Entity FrameworkCore`, while keeping `Microsoft.AspNetCore` at a level of "Warning" or higher.

[...]

PR #25551, opened 2022-04-07
Inconsistency fixed when updating appsettings.

baseTwo:
Fix for issue #25550 

The bullet (below) mentions that |Microsoft.Hosting.Lifetime| should be removed from the appsettings, but the diff example underneath shows erroneously that "Microsoft":"Warning" should be removed too

"Microsoft.Hosting.Lifetime": "Information" was removed from both appsettings.json and appsettings.Development.json:
```diff
- "Microsoft": "Warning", <---- This was removed
- "Microsoft.Hosting.Lifetime": "Information"
+ "Microsoft.AspNetCore": "Warning"
```

### DIFF
--- a/aspnetcore/migration/50-to-60.md
+++ b/aspnetcore/migration/50-to-60.md
@@ -117,7 +117,6 @@ There are a few changes to the other files generated for the Web App template:
 * `"Microsoft.Hosting.Lifetime": "Information"` was removed from both `appsettings.json` and `appsettings.Development.json`:
 
 ```diff
-- "Microsoft": "Warning",
 - "Microsoft.Hosting.Lifetime": "Information"
 + "Microsoft.AspNetCore": "Warning"
 ```


serpent5 (changes requested):
Hi @baseTwo. Thanks for taking the time to open a pull request and associated issue.

The 5.0 `appsettings.json` file has this section:

```json
"LogLevel": {
  "Default": "Information",
  "Microsoft": "Warning",
  "Microsoft.Hosting.Lifetime": "Information"
}
```

The 6.0 version looks like this:

```json
"LogLevel": {
  "Default": "Information",
  "Microsoft.AspNetCore": "Warning"
}
```

This shows that the sample in the article is correct, but the explanation doesn't mention that `"Microsoft": "Warning"` was replaced with `"Microsoft.AspNetCore": "Warning"`. I think the best resolution here is to update the text to mention that additional change.

baseTwo:
The only change necessary going from .net 5 to 6 is the replacement of "Microsoft.Hosting.Lifetime" with "Microsoft.AspNetCore"

We found that when removing  "Microsoft":"Warning" , our logs were more noisy. Perhaps it would be better to keep it there in appsettings.json?

serpent5:
This section describes the difference between the 5.0 and 6.0 templates. If we don't remove the `"Microsoft":"Warning"` line, it's an inaccurate description of said changes.

Rick-Anderson:
@baseTwo  let me know if you're interested fixing via mention that `"Microsoft": "Warning"` was replaced with `"Microsoft.AspNetCore": "Warning"`.

baseTwo:
@Rick-Anderson sure I'll fix mid next week. Out of office until then

Issue #25453, opened 2022-03-28
Keeping webviews up-to-date in deployed Blazor Hybrid apps


MackinnonBuck:
We should add some documentation describing what steps can be taken to ensure that deployed Blazor Hybrid apps run with webview versions that meet the security/API requirements of the app. Following is a draft of what such a page might look like:

***

# Keeping webviews up-to-date in deployed Blazor Hybrid apps

By default, the [`BlazorWebView`](https://docs.microsoft.com/en-us/dotnet/maui/user-interface/controls/blazorwebview) control uses the currently-installed platform-specific native webview. Since webviews are periodically updated with support for new APIs and fixes for security issues, it may be necessary to ensure that your app is using a webview version that meets your app's requirements. This could be achieved in the following ways:
* Checking the webview version being used on the current platform and prompting the user to take any necessary steps to update (applicable for all platforms).
* Packaging a fixed-version webview within your app, using it in place of the system's shared webview (applicable only for Windows).

## Platforms

Different platforms have various mechanisms for keeping the installed webview up-to-date and checking the current webview's version.

### Windows (.NET MAUI, WPF, WinForms)

On Windows, the Chromium-based [Edge WebView2](https://docs.microsoft.com/en-us/microsoft-edge/webview2/) is required to run Blazor web apps. By default, the newest installed version of WebView2 (known as the [Evergreen distribution](https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/distribution#details-about-the-fixed-version-runtime-distribution-mode)) is used. If you wish to ship a specific version of WebView2 with your app, you can use the [Fixed Version distribution mode](https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/distribution#details-about-the-fixed-version-runtime-distribution-mode).

See the [WebView2 distribution docs](https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/distribution) for more information about selecting a WebView2 distribution mode and checking the currently-installed WebView2 version.

### Android

[...]

PR #25555, opened 2022-04-07
Blazor Hybrid Security topic

### LINKED ISSUE #25461: Document security recommendations for remote resources in Blazor Hybrid apps
We want document some security recommendations for remote web resources in Blazor Hybrid. We probably want to have these security recommendations in the same page as https://github.com/dotnet/AspNetCore.Docs/issues/25453.

- When iframing external content within a Blazor Hybrid page, we recommend that users leverage sandboxing features to ensure that the iframed content is isolated from the parent page containing the application.
```html
<iframe sandbox src="https://some-external-page.com/foo.html" />
```
- Avoid having the app render untrusted content from a database or other resource (such as user-provided comments) in its web UI and that content is not properly encoded, it could cause malicious code to execute.
- By default, links to URLs outside of the app will be opened using the system's browser and not load within the app. (The user might be able to say that they do want the URL to load in the app, because it is content that they trust, in which case see previous section about remote content). We recommend against overriding this default behavior.

guardrex:
Addresses #25453
Addresses #24956
Fixes #25461

* The title, description, physical folder (URL), and UID reflect the general coverage. This will eventually be the overview of a node of security topics, and it will cover additional security scenarios as needed.
* I commented out the "fixed-version `WebView`" piece until RC2. We'll leave the issue **open** until the RC2 update is made later. Let's review/update the commented-out text now so that I don't need to ping for a follow-up review at RC2.
* The more general security guidance can close the issue that Safia opened AFAICT.
* I'll leave it to you to call for additional reviewers as needed.
* I'm OOF on Friday, but I'll 🛌💤 on it and merge Friday morning anyway if all of the reviews are in. If we're still awaiting reviews, I'll get it merged on or after Monday when all of the reviews are in.

# 🇺🇦

### DIFF
--- a/aspnetcore/blazor/hybrid/security/index.md
+++ b/aspnetcore/blazor/hybrid/security/index.md
@@ -1,0 +1,70 @@
+---
+title: ASP.NET Core Blazor Hybrid authentication and authorization
+author: guardrex
+description: Learn about Blazor Hybrid authentication and authorization scenarios.
+monikerRange: '>= aspnetcore-6.0'
+ms.author: riande
+ms.custom: mvc
+ms.date: 04/07/2022
+no-loc: [".NET MAUI", "Mac Catalyst", "Blazor Hybrid", Home, Privacy, Kestrel, appsettings.json, "ASP.NET Core Identity", cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
+uid: blazor/hybrid/security/index
+---
+# ASP.NET Core Blazor Hybrid authentication and authorization
+
+This article describes ASP.NET Core's support for the configuration and management of security in Blazor Hybrid apps.
+
+## Untrusted and unencoded content
+
+Avoid allowing an app render untrusted and unencoded content from a database or other resource, such as user-provided comments, in its rendered UI. Permitting untrusted, unencoded content to render can cause malicious code to execute.
+
+## External content rendered in an `iframe`
+
+When using an [`iframe`](https://developer.mozilla.org/docs/Web/HTML/Element/iframe) to display external content within a Blazor Hybrid page, we recommend that users leverage sandboxing features to ensure that the content is isolated from the parent page containing the app. In the following example, the [`sandbox` attribute](https://developer.mozilla.org/docs/Web/HTML/Element/iframe) is present for the `<iframe>` tag to apply [sandboxing features](https://developer.mozilla.org/docs/Web/HTML/Element/iframe) to the `foo.html` page:
+
+```html
+<iframe sandbox src="https://contoso.com/foo.html" />
+```
+
+> [!WARNING]
+> The [`sandbox` attribute](https://developer.mozilla.org/docs/Web/HTML/Element/iframe) is ***not*** supported in early browser versions. For more information, see [Can I use: `sandbox`](https://caniuse.com/?search=sandbox).
+
+## Links to external URLs
+
+By default, links to URLs outside of the app are opened using the system's browser, not loaded within the app. We do ***not*** recommend overriding the default behavior.
+
+The user might be able to indicate that they want the URL to load in the app because it's content that they trust. In that case, see the [Untrusted and unencoded content](#untrusted-and-unencoded-content) section.
+
+## Keep `BlazorWebView` (`WebView`) current in deployed apps
+
+By default, the [`BlazorWebView`](/maui/user-interface/controls/blazorwebview) control uses the currently-installed, platform-specific native `WebView`. Since the native `WebView` is periodically updated with support for new APIs and fixes for security issues, it may be necessary to ensure that an app is using a `WebView` version that meets the app's requirements.
+
+To keep the `WebView` current in deployed apps, check the `WebView` version and prompt the user to take any necessary steps to update it.
+
+<!-- HOLD FOR RC2 AND SWAP FOR THE PRIOR SENTENCE
+
+Use one of the following approaches to keep the `WebView` current in deployed apps:
+
+* **On all platforms**: Check the `WebView` version and prompt the user to take any necessary steps to update it.
+* **Only on Windows**: Package a fixed-version `WebView` within the app, using it in place of the system's shared `WebView`.
+
+-->
+
+### Android
+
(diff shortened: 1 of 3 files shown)


MackinnonBuck:
@Eilon FYI

aspnetcore/blazor/hybrid/security/index.md
```diff
@@ -0,0 +1,74 @@
...
+```html
+<iframe sandbox src="https://contoso.com/foo.html" />
+```
+
+> [!WARNING]
+> The [`sandbox` attribute](https://developer.mozilla.org/docs/Web/HTML/Element/iframe) is ***not*** supported in early browser versions. For more information, see [Can I use: `sandbox`](https://caniuse.com/?search=sandbox).
+
+## Links to external URLs
+
+By default, links to URLs outside of the app are opened using the system's browser, not loaded within the app. We do ***not*** recommend overriding the default behavior.
```
MackinnonBuck, line 33: External links may also be opened in apps capable of handling them (other than the browser). Maybe something like:

```suggestion
By default, links to URLs outside of the app are opened in an appropriate external app, not loaded within the WebView. We do ***not*** recommend overriding the default behavior.
```

aspnetcore/blazor/hybrid/security/index.md
```diff
@@ -0,0 +1,74 @@
...
+
+The user might be able to indicate that they want the URL to load in the app because it's content that they trust. In that case, see the [Untrusted and unencoded content](#untrusted-and-unencoded-content) section.
+
+## Keep `BlazorWebView` (`WebView`) current in deployed apps
```
MackinnonBuck, line 37: Hm, I wonder if this should be reworded to:

```suggestion
## Keep the WebView current deployed apps
```

I think readers might see "`BlazorWebView`" and interpret that as meaning our implementations of the `Microsoft.AspNetCore.Components.WebView.*.BlazorWebView` class rather than the WebView technology itself.

MackinnonBuck, line 37: Hm, I wonder if this should be reworded to:

```suggestion
## Keep the WebView current deployed apps
```

I think readers might see "`BlazorWebView`" and interpret that as meaning our implementations of the `Microsoft.AspNetCore.Components.WebView.*.BlazorWebView` class rather than the WebView technology itself.

aspnetcore/blazor/hybrid/security/index.md
```diff
@@ -0,0 +1,74 @@
...
+
+By default, the [`BlazorWebView`](/maui/user-interface/controls/blazorwebview) control uses the currently-installed, platform-specific native `WebView`. Since the native `WebView` is periodically updated with support for new APIs and fixes for security issues, it may be necessary to ensure that an app is using a `WebView` version that meets the app's requirements.
+
+To keep the `WebView` current in deployed apps, check the `WebView` version and prompt the user to take any necessary steps to update it.
```
MackinnonBuck, line 41: I see "WebView" is `monospaced` everywhere in this document - does it need to be? To me, the monospacing implies that it refers to either some code concept or a specific WebView implementation. But most places in this document use "WebView" as a more general concept. Do you think we should switch to making it "WebView" rather than "`WebView`"?

MackinnonBuck, line 41: I see "WebView" is `monospaced` everywhere in this document - does it need to be? To me, the monospacing implies that it refers to either some code concept or a specific WebView implementation. But most places in this document use "WebView" as a more general concept. Do you think we should switch to making it "WebView" rather than "`WebView`"?

aspnetcore/blazor/hybrid/security/index.md
```diff
@@ -0,0 +1,74 @@
...
+
+On Windows, the Chromium-based [Microsoft Edge `WebView2`](/microsoft-edge/webview2/) is required to run Blazor web apps. By default, the newest installed version of `WebView2` (known as the [Evergreen distribution](/microsoft-edge/webview2/concepts/distribution#details-about-the-fixed-version-runtime-distribution-mode)) is used.
+
+<!-- AT RC2, ADD THE FOLLOWING SENTENCE TO THE PRECEDING PARAGRAPH
+
+If you wish to ship a specific version of `WebView2` with the app, use the [Fixed Version distribution mode](/microsoft-edge/webview2/concepts/distribution#details-about-the-fixed-version-runtime-distribution-mode).
+
+-->
+
+For more information on selecting a `WebView2` distribution mode and checking the currently-installed `WebView2` version, see the [`WebView2` distribution docs](/microsoft-edge/webview2/concepts/distribution).
```
MackinnonBuck, line 70:
```suggestion
For more information on checking the currently-installed `WebView2` version, see the [`WebView2` distribution docs](/microsoft-edge/webview2/concepts/distribution).
```

The removed part here is only relevant with the other additions happening at RC2.

MackinnonBuck, line 70:
```suggestion
For more information on checking the currently-installed `WebView2` version, see the [`WebView2` distribution docs](/microsoft-edge/webview2/concepts/distribution).
```

The removed part here is only relevant with the other additions happening at RC2.

aspnetcore/blazor/hybrid/security/index.md
guardrex: It was consistent with various API/UI controls, especially for consistent non-localization. I have seen it as "web view" or "Web View" with a space around the InterWebs. We can have it in text, which sounds like you prefer, as "Web View" (with a space). To avoid loc, I'll place it with our loc-avoidance marker ...

```
:::no-loc text="Web View":::
```

... but I'd like to do it everywhere else at once, so I've opened ...

 Deloc and de-fence "Web View" #25605 

... and will just focus the update for this doc at this time.

MackinnonBuck (approved):
This looks good to me. Any other thoughts @dotnet/aspnet-blazor-eng?

TanayParikh (approved):

aspnetcore/blazor/hybrid/security/index.md
guardrex: Turns out that additional work wasn't needed, so I closed #25605. We only refer to `BlazorWebView` outside of a couple of Edge-specific `WebView2` remarks/cross-links.

guardrex:
@MackinnonBuck ... Did you say we were withholding the commented-out bits for **_RC2_**? My understanding was that we go from RC1 to GA. If we can surface the commented-out bits with RC1, then I suggest I go ahead and surface the content now and hold this just until later today for publication. If there really is an RC2 (and we'll leave the commented-out bits in place), then I'll just wait on your final signal here to merge this.

PR #25602, opened 2022-04-11
Adjust text to recommend Static web apps

IEvangelist:
Tagging @bradygaster and @btardif


guardrex (approved):
Thanks @IEvangelist. I touch up the service naming, too ... "Static Web Apps" is a proper noun, so I fix that along the way here.

btardif:
thanks @IEvangelist 👍🏽

PR #25604, opened 2022-04-11
Blazor debug published, non-production, hosted WASM

### LINKED ISSUE #25512: Enhance with hosted Blazor WASM debug scenario
Per **Debugging Blazor WASM after publish**: https://github.com/dotnet/aspnetcore/issues/41033.

Dev:

> But the problem is that, when I publish it (both with Release or Debug configuration), and when I launch the output executable, Alt+Shift+D throws an error on the browser console:
>
> Cannot start debugging, because the application was not compiled with debugging enabled.

Javier:

> You need to make sure you enable `DebuggerSupport` as well as `CopyOutputSymbolsToPublishDirectory` when you publish

---
#### Document Details

* ID: ec2f61c7-9f51-6735-e328-3c7b495d70c3
* Version Independent ID: 4d06e565-282d-6387-aa90-cb2c2986d35a
* Content: [Debug ASP.NET Core Blazor WebAssembly](https://docs.microsoft.com/en-us/aspnet/core/blazor/debug?view=aspnetcore-6.0&tabs=visual-studio)
* Content Source: [aspnetcore/blazor/debug.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/debug.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #25512

Are the problems that ...

* Debugger support introduces security risk?
* Copying output symbols introduces perf risk (of deployment ... larger app, slower deployment)?

I think it's good if we say why these should be disabled for published, production apps.

### DIFF
--- a/aspnetcore/blazor/debug.md
+++ b/aspnetcore/blazor/debug.md
@@ -220,9 +220,7 @@ To debug a **published**, hosted Blazor WebAssembly app, configure debugger supp
 By default, publishing an app disables the preceding properties by setting them to `false`.
 
 > [!WARNING]
-> Published, hosted Blazor WebAssembly apps should only enable debugging when deploying published assets ***locally***. Do **not*** deploy a published app into production with the `DebuggerSupport` property set to `true`.
->
-> If `CopyOutputSymbolsToPublishDirectory` is set to `true` in a published, production app, there are no security risks, but app size and deployment speed are adversely impacted for no benefit. Therefore, apps deployed into production should ***not** enable copying symbols to the `publish` directory.
+> Published, hosted Blazor WebAssembly apps should only enable debugging and copying output symbols when deploying published assets ***locally***. Do **not*** deploy a published app into production with the `DebuggerSupport` and `CopyOutputSymbolsToPublishDirectory` properties set to `true`.
 
 ## Attach to an existing debugging session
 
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/debug.md
```diff
@@ -210,6 +210,20 @@ For guidance on configuring VS Code assets in the `.vscode` folder and where to
...
+<DebuggerSupport>true</DebuggerSupport>
+<CopyOutputSymbolsToPublishDirectory>true</CopyOutputSymbolsToPublishDirectory>
+```
+
+By default, publishing an app disables the preceding properties by setting them to `false`.
+
+> [!WARNING]
+> Published, hosted Blazor WebAssembly apps should only enable debugging when deploying published assets ***locally***. Do **not*** deploy a published app into production with the `DebuggerSupport` property set to `true`.
+>
+> If `CopyOutputSymbolsToPublishDirectory` is set to `true` in a published, production app, there are no security risks, but app size and deployment speed are adversely impacted for no benefit. Therefore, apps deployed into production should ***not** enable copying symbols to the `publish` directory.
```
captainsafia, line 225: > If `CopyOutputSymbolsToPublishDirectory` is set to `true` in a published, production app, there are no security risks

I dunno if I feel confident making this statement. It feels like we should generally not recommend having symbols on servers for apps, even if they are not served by the Blazor front-end.

guardrex: It's worded to not suggest it, but we've been ultra conservative before in similar scenarios, so I'll update it.

captainsafia (approved):

PR #25620, opened 2022-04-13
Add example how to use the AsyncAuthInterceptor

### LINKED ISSUE #24132: Add async example for Bearer token authentication gRPC client factory
This [paragraph](https://docs.microsoft.com/en-us/aspnet/core/grpc/authn-and-authz?view=aspnetcore-6.0#bearer-token-with-grpc-client-factory) describes how to setup the client factory. It also show how to provide a token. I tried making this method async, but when I do that it doesn't work anymore. Please provide an example how to provide a token using an async method.

---
#### Document Details

* ID: 363b40bb-4e20-e4b3-4456-5eb8fe6b30fb
* Version Independent ID: 8a7e85b0-5c1c-969d-9748-5e5e2dc45cce
* Content: [Authentication and authorization in gRPC for ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/grpc/authn-and-authz?view=aspnetcore-6.0)
* Content Source: [aspnetcore/grpc/authn-and-authz.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/grpc/authn-and-authz.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-grpc**
* GitHub Login: @JamesNK
* Microsoft Alias: **jamesnk**

danm-de:
* added example service registration that uses the AsyncAuthInterceptor to request a bearer token

Fixes [https://github.com/grpc/grpc-dotnet/issues/1682](https://github.com/grpc/grpc-dotnet/issues/1682)
Fixes #24132

### DIFF
--- a/aspnetcore/grpc/authn-and-authz.md
+++ b/aspnetcore/grpc/authn-and-authz.md
@@ -171,7 +171,7 @@ The preceeding code:
 * Registers the `GreeterClient` type with client factory.
 * Configures the `AuthInterceptor` for this client using `InterceptorScope.Client`. A new interceptor is created for each client instance. When a client is created for a gRPC service or Web API controller, the scoped `ITokenProvider` is injected into the interceptor.
 
-##### Bearer token with gRPC client factory using AsyncAuthInterceptor
+##### Bearer token with gRPC client factory using `AsyncAuthInterceptor`
 
 If the bearer token can only be generated using asynchronous method calls, the `AsyncAuthInterceptor` must be used.
 
--- a/aspnetcore/grpc/authn-and-authz.md
+++ b/aspnetcore/grpc/authn-and-authz.md
@@ -199,7 +199,7 @@ services
         {
             using var scope = scopeFactory.CreateScope();
             var tokenProvider = scope.ServiceProvider.GetRequiredService<ITokenProvider>();
-            var token = await tokenProvider.GetToken();
+            var token = await tokenProvider.GetTokenAsync();
             metadata.Add("Authorization", $"Bearer {token}");
         });
         options.Credentials = ChannelCredentials.Create(
--- a/aspnetcore/grpc/authn-and-authz.md
+++ b/aspnetcore/grpc/authn-and-authz.md
@@ -171,6 +171,43 @@ The preceeding code:
 * Registers the `GreeterClient` type with client factory.
 * Configures the `AuthInterceptor` for this client using `InterceptorScope.Client`. A new interceptor is created for each client instance. When a client is created for a gRPC service or Web API controller, the scoped `ITokenProvider` is injected into the interceptor.
 
+##### Bearer token with gRPC client factory using AsyncAuthInterceptor
+
+If the bearer token can only be generated using asynchronous method calls, the `AsyncAuthInterceptor` must be used.
+
+Assuming the token provider is defined as follows and has been registered as scoped service:
+
+```csharp
+public interface ITokenProvider
+{
+    Task<string> GetToken();
+}
+``` 
+
+You have to register the gRPC client using the `AsyncAuthInterceptor`
+
+```csharp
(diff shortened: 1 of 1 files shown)


wadepickett:
@JamesNK, I can move this into a .NET 6.0  gRPC client factory app sample and then pull the relevant code snippets back into the topic if there is a version of this code sample you sign off on.

@danm-de,  Thanks for taking the time and effort to provide a code example!

aspnetcore/grpc/authn-and-authz.md
```diff
@@ -171,6 +171,43 @@ The preceeding code:
 * Registers the `GreeterClient` type with client factory.
 * Configures the `AuthInterceptor` for this client using `InterceptorScope.Client`. A new interceptor is created for each client instance. When a client is created for a gRPC service or Web API controller, the scoped `ITokenProvider` is injected into the interceptor.
 
+##### Bearer token with gRPC client factory using AsyncAuthInterceptor
```
wadepickett, line 174: I'll move the final version of this sample if there is one approved by JamesNK to a full client factory app sample for .NET 6 and then pull in the relevant snippets

wadepickett (changes requested):
See comment.  I will need to move this out to an app sample and link the snippets back in if JamesNK approves

aspnetcore/grpc/authn-and-authz.md
```diff
@@ -171,6 +171,43 @@ The preceeding code:
...
+    .AddGrpcClient<Greeter.GreeterClient>(o =>
+    {
+        o.Address = new Uri("https://localhost:5001");
+    })
+    .ConfigureChannel((sp, options) =>
+    {
+        var scopeFactory = sp.GetRequiredService<IServiceScopeFactory>();
+        var asyncAuthInterceptor = new AsyncAuthInterceptor(async (context, metadata) =>
+        {
+            using var scope = scopeFactory.CreateScope();
```
JamesNK, line 200: Hmm, is creating a new scope here, and registering `ITokenProvider` with a scoped lifetime providing any benefit?

Each call is creating a new scope, and each time a new `ITokenProvider` instance is created. Is that much different from a transient lifetime? I haven't built a full gRPC app using scoped auth, but I would have thought the benefit that you're aiming for is caching the token per-request, so calls for that request share it. Will you get that here?

I wonder if there should be a new extension method for the gRPC client factory that is just for configuring `AsyncAuthInterceptor`. An overload can include an `IServiceProvider` that uses the current request scope.

danm-de: I figured this would be necessary if `ITokenProvider` is registered as 'scoped' - or itself has dependencies registered as `scoped` services. If a background process opens an RPC connection (service-2-service communication) we don't necessarily have a scope - then there would be an exception here, right?

danm-de: The example would be shorter and probably easier to understand if we say

> [...] and has been registered as *singleton* service [...]

wadepickett: @JamesNK, it is not clear to me what state this is in at the moment.  Does the example require a changed from what it is at this point?

JamesNK:
This docs PR has inspired some new features in the gRPC client factory. It should make this much easier. See https://github.com/grpc/grpc-dotnet/pull/1705

Let's pause this PR and then adapt it to the new feature once it's merged.

JamesNK:
@danm-de See https://github.com/dotnet/AspNetCore.Docs/pull/25734 for docs on a new auth feature that's being added to grpc-dotnet. It would replace this PR.

JamesNK:
Thanks for working on this. A new feature and docs - https://github.com/dotnet/AspNetCore.Docs/pull/25734 - replaces this. Closing

PR #25631, opened 2022-04-14
Blazor Hybrid link handling behaviors

### LINKED ISSUE #25432: Improved non-HTTP(S) link handling behavior
https://github.com/dotnet/maui/pull/5313

---
#### Document Details

* ID: 48146c6d-f864-0b56-a2d5-0493a843c7a7
* Version Independent ID: 48146c6d-f864-0b56-a2d5-0493a843c7a7
* Content: [ASP.NET Core Blazor Hybrid routing and navigation](https://docs.microsoft.com/en-us/aspnet/core/blazor/hybrid/routing?view=aspnetcore-6.0)
* Content Source: [aspnetcore/blazor/hybrid/routing.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/hybrid/routing.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #25432

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/hybrid/routing?view=aspnetcore-6.0&branch=pr-en-us-25631)

* There were commits and additional discussion after your remarks at https://github.com/dotnet/maui/pull/5313#issuecomment-1078511479 on the PU PR, and I don't know if your remarks were updated for those followup commits/new design.
* Also, the remarks and discussion are kind'a cryptic in a few spots, such as for permutations for any scheme-based behaviors, static files in the app, and platform/OS behaviors.
* I recommend that you edit the PR directly to get the technical remarks right, then I'll circle around and sprinkle in a few commas and what-nots. You can see where I'm going with this: I suggest laying out the default behaviors first, then describing the dev control that can be taken for URL loading. I feel much more confident in the text/examples for the latter than I do for the initial default behavior section at the top of the doc.
* Also note that I hit a little confusion over the use of URL vs. URI in these contexts, so you'll want to assess that as part of the review to see if/where things went wrong.
* I recommend structuring the bullet list under the line "**_Default URI request routing behavior:_**" to match (roughly, at least) the **_order of applied behaviors_**, if possible. 🤔 I think it makes sense because routing behaviors may change depending on the processing sequence that the framework follows.

### DIFF
--- a/aspnetcore/blazor/hybrid/routing.md
+++ b/aspnetcore/blazor/hybrid/routing.md
@@ -19,20 +19,20 @@ This article explains how to manage request routing and navigation in Blazor Hyb
 
 Default URI request routing behavior:
 
-* A link is *internal* if the host name of the app's origin URI matches the host name of the request URI. When the host names don't match or if the link sets `target="_blank"`, the link is considered *external*.
+* A link is *internal* if the host name and scheme match between the app's origin URI and the request URI. When the host names and schemes don't match or if the link sets `target="_blank"`, the link is considered *external*.
 * If the link is internal, the link is opened in the `BlazorWebView` by the app.
 * If the link is external, the link is opened by an app determined by the device based on the device's registered handler for the link's scheme.
-* For links that appear to request a file because the last segment of the URI uses dot notation (for example, `/file.x`, `/Maryia.Melnyk`, `/image.gif`):
-  * .NET MAUI (WinUI), WPF, and Windows Forms: ?
-  * .NET MAUI (Android, iOS/Mac Catalyst): ?
+* For internal links that appear to request a file because the last segment of the URI uses dot notation (for example, `/file.x`, `/Maryia.Melnyk`, `/image.gif`) but don't point to any static content:
+  * WPF and Windows Forms: The host page content will be returned.
+  * .NET MAUI: A 404 response will be returned.
 
 To change the link handling behavior for links that don't set `target="_blank"`, register the `UrlLoading` event and set the `UrlLoadingEventArgs.UrlLoadingStrategy` property. The `UrlLoadingStrategy` enumeration allows setting link handling behavior to any of the following values:
 
-* `OpenExternally`: Load the URI using an app determined by the device. This is the default strategy for URIs with an external host.
-* `OpenInWebView`: Load the URI within the `BlazorWebView`. This is the default strategy for URIs with a host matching the app origin. ***Don't use this strategy for external links unless you can ensure the destination URI is fully trusted.***
-* `CancelNavigation`: Cancels the current URI loading attempt.
+* `OpenExternally`: Load the URL using an app determined by the device. This is the default strategy for URIs with an external host.
+* `OpenInWebView`: Load the URL within the `BlazorWebView`. This is the default strategy for URLs with a host matching the app origin. ***Don't use this strategy for external links unless you can ensure the destination URI is fully trusted.***
+* `CancelLoad`: Cancels the current URL loading attempt.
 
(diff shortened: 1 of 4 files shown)


aspnetcore/blazor/hybrid/routing.md
```diff
@@ -15,10 +15,27 @@ This article explains how to manage request routing and navigation in Blazor Hyb
...
+[!INCLUDE[](~/blazor/includes/net-maui-release-candidate-notice.md)]
+
+Default URI request routing behavior:
+
+* A link is *internal* if the host name and scheme match between the app's origin URI and the request URI. When the host names and schemes don't match or if the link sets `target="_blank"`, the link is considered *external*.
+* If the link is internal, the link is opened in the `BlazorWebView` by the app.
+* If the link is external, the link is opened by an app determined by the device based on the device's registered handler for the link's scheme.
+* For internal links that appear to request a file because the last segment of the URI uses dot notation (for example, `/file.x`, `/Maryia.Melnyk`, `/image.gif`) but don't point to any static content:
+  * WPF and Windows Forms: The host page content will be returned.
+  * .NET MAUI: A 404 response will be returned.
```
MackinnonBuck, line 27: Since this is an edge case that will be rarely encountered by customers, I almost wonder if it would be better to put this in a "WARNING" box instead? We would warn specifically about using file-like page URIs on MAUI platforms.

aspnetcore/blazor/hybrid/routing.md
```diff
@@ -33,45 +50,42 @@ using Microsoft.AspNetCore.Components.WebView;
...
 
 ```csharp
-blazorWebView.ExternalNavigationStarting += 
-    (sender, externalLinkNavigationEventArgs) =>
+blazorWebView.UrlLoading += 
+    (sender, urlLoadingEventArgs) =>
     {
-        externalLinkNavigationEventArgs.ExternalLinkNavigationPolicy = 
-            ExternalLinkNavigationPolicy.InsecureOpenInWebView;
+        urlLoadingEventArgs.UrlLoadingStrategy = UrlLoadingStrategy.OpenInWebView;
```
MackinnonBuck, line 56: Nit: Since we have a warning about opening external links in the WebView, maybe we should make the examples here do something else? For example, cancel all external navigation requests:

```suggestion
    if (urlLoadingEventArgs.UrlLoadingStrategy == UrlLoadingStrategy.OpenExternally)
            urlLoadingEventArgs.UrlLoadingStrategy = UrlLoadingStrategy.CancelLoad;
```

However, I realize this suggestion doesn't completely relate to the changes in this PR, and it does slightly complicate the example, so it might not be necessary. Just an idea.

guardrex: I leave the choice up to you; but we'll need to explain this in the text if the doc shows it.

aspnetcore/blazor/hybrid/routing.md
guardrex: Fine with me either way ... whatever you think is best.

guardrex: This should be present tense tho, so I'll make that change at the end if you don't change it when making a WARNING.

* The host page content will be returned. 👉 The host page content is returned.
* A 404 response will be returned. 👉 A 404 response is returned.

aspnetcore/blazor/hybrid/routing.md
MackinnonBuck: Let's leave it as-is then to keep the focus on how to subscribe to the event, rather than the logic inside the event handler.

PR #25645, opened 2022-04-16
Update _Layout.cshtml to correct border-radius CSS property

EKanadily:
changed -radius to border-radius (as -radius is not a valid CSS property)

no issue number.


Rick-Anderson:
This is two versions back and not an essential update. Thanks for the PR and I hope you make another contribution.

PR #25646, opened 2022-04-17
Make better wording for beginners

Sebight:
Beginners could associate the word "origin" with the word "domain" and it could be easier in this first CORS example.


guardrex:
Hello @Sebight ... This isn't accurate according to the [spec](https://tools.ietf.org/html/rfc6454), which is summarized by the preceding line. Beginners should access the spec for more information.

PR #25670, opened 2022-04-19
[RC2] Keep WebViews current

### LINKED ISSUE #25453: Keeping webviews up-to-date in deployed Blazor Hybrid apps

guardrex:
Fixes #25453

**_HOLD for RC2 release._**

### DIFF
--- a/aspnetcore/blazor/hybrid/security/index.md
+++ b/aspnetcore/blazor/hybrid/security/index.md
@@ -5,7 +5,7 @@ description: Learn about Blazor Hybrid authentication and authorization scenario
 monikerRange: '>= aspnetcore-6.0'
 ms.author: riande
 ms.custom: mvc
-ms.date: 04/07/2022
+ms.date: 04/19/2022
 no-loc: [".NET MAUI", "Mac Catalyst", "Blazor Hybrid", Home, Privacy, Kestrel, appsettings.json, "ASP.NET Core Identity", cookie, Cookie, Blazor, "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", Razor, SignalR]
 uid: blazor/hybrid/security/index
 ---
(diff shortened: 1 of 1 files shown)


guardrex:
@MackinnonBuck ... RC2 is out. Can I go ahead and merge this now?

PR #25686, opened 2022-04-20
host and app config, appsettings Urls takes precedence over ASPNETCORE /2

### LINKED ISSUE #25626: appsettings Urls takes precedence over ASPNETCORE_URLS in 6.0.4
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Using aspnet 6.0.4 runtime, precedence of appsettings.json `Urls` and `ASPNETCORE_URLS` seems to be different compared to previous versions. Specifying `ASPNETCORE_URLS` environment variable does not override the url provided in appsettings.json `Urls` property. Before 6.0.4, `ASPNETCORE_URLS` would have been used over `Urls` in appsettings.json .

I would assume that the order that these settings are overriden in is the same as for regular settings: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-6.0#default-configuration 

Based on those, environment variables are applied after appsettings.

I am guessing this is related to https://github.com/dotnet/aspnetcore/pull/39836 which meant to fix in-code overriding of any configuration value. Also, while that issue is tagged with 6.0.3, based on comment https://github.com/dotnet/aspnetcore/issues/40614#issuecomment-1067397572 it looks like it was shipped as part of 6.0.4, which is a bit confusing.

### Expected Behavior

`ASPNETCORE_URLS` is used instead of appsettings.json `Urls`.

### Steps To Reproduce

- Create new project `dotnet new webapi`
- Modify `appsettings.json` by adding `"Urls": "http://*:5300/"`
- Set env var `export ASPNETCORE_URLS="http://*:5200/"`
- `dotnet run`

```
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: http://[::]:5300
```

***@Rick-Anderson EDIT following lines***

---
#### Document Details

* Content: [Configuration in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-6.0)
* Content Source: [aspnetcore/fundamentals/configuration/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/configuration/index.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

Rick-Anderson:
Fixes #25626
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-6.0&branch=pr-en-us-25686)

@Hoffs @chertby  please review.

### DIFF
--- a/aspnetcore/fundamentals/configuration/index.md
+++ b/aspnetcore/fundamentals/configuration/index.md
@@ -65,7 +65,7 @@ The following list contains host configuration sources from lowest to highest pr
 1. Command-line arguments
 1. `ASPNETCORE_` prefixed environment variable
 
-When a configuration value is set in host and application configuration, the host configuration is used.
+When a configuration value is set in host and application configuration, the application configuration is used.
 
 See [**Explanation** in this GitHub comment](https://github.com/dotnet/AspNetCore.Docs/issues/25626#issuecomment-1098616664) for an explanation of why in host configuration, `ASPNETCORE_` prefixed environment variables have higher priority than command-line arguments.
 
@@ -73,10 +73,10 @@ See [**Explanation** in this GitHub comment](https://github.com/dotnet/AspNetCor
 
 The following variables are locked in early when initializing the host builders and can't be influenced by application config:
 
-* Application name
-* Environment name, for example `Development`, `Production`, and `Staging`
-* Content root
-* Web root
+* [Application name](xref:fundamentals/minimal-apis#change-the-content-root-application-name-and-environment)
+* [Environment name](xref:fundamentals/environments), for example `Development`, `Production`, and `Staging`
+* [Content root](xref:fundamentals/index#content-root)
+* [Web root](xref:fundamentals/index#web-root)
 * Whether to scan for [hosting startup assemblies](xref:fundamentals/configuration/platform-specific-configuration) and which assemblies to scan for.
 * Variables read by app and library code from [HostBuilderContext.Configuration](xref:Microsoft.Extensions.Hosting.HostBuilderContext.Configuration) in [IHostBuilder.ConfigureAppConfiguration](xref:Microsoft.Extensions.Hosting.IHostBuilder.ConfigureAppConfiguration%2A) callbacks.
 
--- a/aspnetcore/fundamentals/configuration/index.md
+++ b/aspnetcore/fundamentals/configuration/index.md
@@ -40,14 +40,14 @@ ASP.NET Core web apps created with [dotnet new](/dotnet/core/tools/dotnet-new) o
 var builder = WebApplication.CreateBuilder(args);
 ```
 
-[WebApplication.CreateBuilder](xref:Microsoft.AspNetCore.Builder.WebApplication.CreateBuilder%2A) initializes a new instance of the <xref:Microsoft.AspNetCore.Builder.WebApplicationBuilder> class with preconfigured defaults. The initialized `WebApplicationBuilder` (`builder`) provides default configuration for the app in the following order:
+[WebApplication.CreateBuilder](xref:Microsoft.AspNetCore.Builder.WebApplication.CreateBuilder%2A) initializes a new instance of the <xref:Microsoft.AspNetCore.Builder.WebApplicationBuilder> class with preconfigured defaults. The initialized `WebApplicationBuilder` (`builder`) provides default configuration for the app in the following order, from highest to lowest:
 
-1. [ChainedConfigurationProvider](xref:Microsoft.Extensions.Configuration.ChainedConfigurationSource) :  Adds an existing `IConfiguration` as a source. In the default configuration case, adds the [host](#hvac) configuration and setting it as the first source for the *app* configuration.
-1. [appsettings.json](#appsettingsjson) using the [JSON configuration provider](#file-configuration-provider).
-1. `appsettings.{Environment}.json` using the [JSON configuration provider](#file-configuration-provider). For example, `appsettings.Production.json` and `appsettings.Development.json`.
-1. [App secrets](xref:security/app-secrets) when the app runs in the `Development` environment.
-1. Environment variables using the [Environment Variables configuration provider](#evcp).
 1. Command-line arguments using the [Command-line configuration provider](#command-line).
+1. Environment variables using the [Environment Variables configuration provider](#evcp).
+1. [User secrets](xref:security/app-secrets) when the app runs in the `Development` environment.
+1. `appsettings.{Environment}.json` using the [JSON configuration provider](#file-configuration-provider). For example, `appsettings.Production.json` and `appsettings.Development.json`.
+1. [appsettings.json](#appsettingsjson) using the [JSON configuration provider](#file-configuration-provider).
+1. A fallback to the host configuration described in the next section
 
 Configuration providers that are added later override previous key settings. For example, if `MyKey` is set in both `appsettings.json` and the environment, the environment value is used. Using the default configuration providers, the  [Command-line configuration provider](#clcp) overrides all other providers.
 
@@ -59,11 +59,10 @@ The following code displays the enabled configuration providers in the order the
 
 ## Host configuration
 
-The following list contains the default host configuration sources from lowest to highest priority:
+The following list contains the default host configuration sources from highest to lowest priority:
 
-1. `DOTNET_` prefixed environment variables
-1. Command-line arguments
 1. `ASPNETCORE_` prefixed environment variable
+1.  Command-line arguments using the [Command-line configuration provider](#command-line)1. `DOTNET_` prefixed environment variables
 
 When a configuration value is set in host and application configuration, the application configuration is used.
 
--- a/aspnetcore/fundamentals/configuration/index.md
+++ b/aspnetcore/fundamentals/configuration/index.md
@@ -46,7 +46,7 @@ var builder = WebApplication.CreateBuilder(args);
 
 <a name="hi2low"></a>
 
-[WebApplication.CreateBuilder](xref:Microsoft.AspNetCore.Builder.WebApplication.CreateBuilder%2A) initializes a new instance of the <xref:Microsoft.AspNetCore.Builder.WebApplicationBuilder> class with preconfigured defaults. The initialized `WebApplicationBuilder` (`builder`) provides default configuration for the app in the following order, from highest to lowest:
+[WebApplication.CreateBuilder](xref:Microsoft.AspNetCore.Builder.WebApplication.CreateBuilder%2A) initializes a new instance of the <xref:Microsoft.AspNetCore.Builder.WebApplicationBuilder> class with preconfigured defaults. The initialized `WebApplicationBuilder` (`builder`) provides default configuration for the app in the following order, from highest to lowest priority:
 
 1. Command-line arguments using the [Command-line configuration provider](#command-line).
 1. Non-prefixed environment variables using the [Non-prefixed environment variables configuration provider](#evcp).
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@serpent5 > Application config falls back to host config,
 see [this comment](https://github.com/dotnet/AspNetCore.Docs/issues/25626#issuecomment-1098616664)

What does falls back mean? Host config first, then app config? Just getting started. Please review the issue, I'll have question for you Thursday morning.

serpent5:
I think it ultimately means a value will be read from host config if it doesn't exist in app config. App config overrides host config.

aspnetcore/fundamentals/configuration/index.md
```diff
@@ -57,6 +57,35 @@ The following code displays the enabled configuration providers in the order the
...
+The following variables are locked in early when initializing the host builders and can't be influenced by application config:
+
+* Application name
+* Environment name, for example `Development`, `Production`, and `Staging`
+* Content root
+* Web root
+* Whether to scan for [hosting startup assemblies](xref:fundamentals/configuration/platform-specific-configuration) and which assemblies to scan for.
+* Variables read by app and library code from [HostBuilderContext.Configuration](xref:Microsoft.Extensions.Hosting.HostBuilderContext.Configuration) in [IHostBuilder.ConfigureAppConfiguration](xref:Microsoft.Extensions.Hosting.IHostBuilder.ConfigureAppConfiguration%2A) callbacks.
+
+`URLS` is notably absent from the previous list of host variables. `URLS` is read later from application config where non-prefixed environment variables override `appsettings.json`. `ASPNETCORE_` prefixed and `DOTNET_` prefixed environment variables set in application configuration are overridden by those set in host configuration.
```
Hoffs, line 83: I feel like this introduces a contradiction with statement above: `When a configuration value is set in host and application configuration, the host configuration is used.`

Following that statement, my understanding is that `ASPNETCORE_URLS` has highest priority and if `Urls` is provided in application configuration, then `host configuration is used`. But then this section says:
> `ASPNETCORE_` prefixed and `DOTNET_` prefixed environment variables set in application configuration are overridden by those set in host configuration

Which states the opposite. That it is not `host configuration is used`, but rather application configuration is used.

Or is this purely for `Host variables`? But in that case this statement makes me wonder why mention `application configuration` at all if it can't influence host configuration at all:

> The following variables are locked in early when initializing the host builders and can't be influenced by application config:

Hoffs: I re-read the statement, and I suppose it states that variables that would be set in application configuration will not be used when host configuration contains those same variables, as application configuration is overridden by host configuration. Which ultimately leads back to overall meaning that host configuration takes precedence over application configuration. Combining statement above that host configuration is used over application configuration and that `ASPNETCORE_ prefixed environment variable` has highest priority, it would make me believe that `ASPNETCORE_URLS` is the highest priority variable.

Rick-Anderson:
@halter73 please review

aspnetcore/fundamentals/configuration/index.md
halter73: Here's what I wrote in the comment:

> `URLS` is notably absent from this list of host variables. URLS is read later from application config where non-prefixed environment variables override appsettings.json, but `ASPNETCORE_`-prefixed and `DOTNET_`-prefixed environment variables get overridden.

The wording I had here wasn't clear, but I meant to indicate that `URLS` gets overridden by application config, not hosting config, because it is not a "host variable".

I think a lot of settings like `URLS` and `shutdownTimeoutSeconds` feel like they should be "host variables", but they are not by this definition. What if we refer to everything above like content root and environment name as "host bootstrap settings" instead of "host variables"? This would indicate that only settings needed in the very early stages of building the host like content root and environment name are included but not all host-related settings.

How about this?

```suggestion
Every other host setting is read from application config instead of host config.

`URLS` is one of the many common host settings that is not a bootstrap setting. Like every other host setting not in the previous list, `URLS` is read later from application config. Host config is a fallback for application config, so host config can be used to set `URLS`, but it will be overridden by any configuration source in application config like `appsettings.json`.
```

aspnetcore/fundamentals/configuration/index.md
```diff
@@ -26,13 +26,13 @@ Configuration in ASP.NET Core is performed using one or more [configuration prov
...
-This topic provides information on configuration in ASP.NET Core. For information on using configuration in console apps, see [.NET Configuration](/dotnet/core/extensions/configuration).
+ASP.NET Core apps configure and launch a *host*. The host is responsible for app startup and lifetime management. The ASP.NET Core templates create a <xref:Microsoft.AspNetCore.Builder.WebApplicationBuilder> which contains the host. While some configuration can be done in both the host and the application configuration providers, generally, only configuration that is necessary for the host should be done in host configuration.
 
-[View or download sample code](https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/fundamentals/configuration/index/samples) ([how to download](xref:index#how-to-download-a-sample))
+This article provides information on configuration in ASP.NET Core. For information on using configuration in console apps, see [.NET Configuration](/dotnet/core/extensions/configuration).
 
 <a name="default"></a>
 
-## Default configuration
+## Default application configuration
```
halter73, line 35: I think the structure should be as follows.

```
## Default configuration
### Application configuration
### Host configuration
### Host bootstrap settings
```

I also think that the section discussing default application config should be reordered to go from highest to lowest priority along with the same list for default host configuration sources. As ordered right now, the highest priority source for application config (Command-line arguments) is in the middle of the combined list of application and host sources which makes no sense. Application config falls back to host config, so I want to be able to skim down the list of sources without the order resetting in the middle.

The following should be deleted:

> 1. [ChainedConfigurationProvider](xref:Microsoft.Extensions.Configuration.ChainedConfigurationSource) : Adds an existing `IConfiguration` as a source. In the default configuration case, adds the [host](#hvac) configuration and setting it as the first source for the *app* configuration.

It's super confusing. Hardly anyone should care about the mechanism that application config falls back to hosting config (which isn't entirely accurate anymore anyway for `WebApplicationBuilder`). I think it similar to the `user secrets` where the actual configuration provider used to make this fallback work (if any) is unimportant.

The last (lowest-priority) element of the list of default application sources should just read:

> 1. A fallback to the host configuration described in the next section

Can we also update `[App secrets](xref:security/app-secrets)` to `[user-secrets](xref:security/app-secrets)` and `Environment variables` to `Non-prefixed environment variables` similar to in my comment?

aspnetcore/fundamentals/configuration/index.md
```diff
@@ -57,6 +57,35 @@ The following code displays the enabled configuration providers in the order the
 
 [!code-csharp[](index/samples/6.x/ConfigSample/Pages/Index2.cshtml.cs?name=snippet)]
 
+## Host configuration
+
+The following list contains host configuration sources from lowest to highest priority:
```
halter73, line 62:
```suggestion
The following list contains the default host configuration sources from lowest to highest priority:
```

aspnetcore/fundamentals/configuration/index.md
```diff
@@ -57,6 +57,35 @@ The following code displays the enabled configuration providers in the order the
...
+
+1. `DOTNET_` prefixed environment variables
+1. Command-line arguments
+1. `ASPNETCORE_` prefixed environment variable
```
halter73, line 66: Can we be consistent about providing links to `#evcp` and `#command-line` like we do for default application config sources. I want to make the list of sources look as much like a single list as possible while still making it clear that host bootstrap settings are only influenced by host configuration.

aspnetcore/fundamentals/configuration/index.md
```diff
@@ -26,37 +26,79 @@ Configuration in ASP.NET Core is performed using one or more [configuration prov
...
+
+The remaining sections in this article refer to application configuration.
+
+## Application configuration providers
 
 The following code displays the enabled configuration providers in the order they were added:
 
 [!code-csharp[](index/samples/6.x/ConfigSample/Pages/Index2.cshtml.cs?name=snippet)]
 
+The preceding [highest to lowest default configuration list](#hi2low) shows the providers in the opposite order they are added to template generated application. For example, the [JSON configuration provider](#file-configuration-provider) is added before the [Command-line configuration provider](#command-line).
```
halter73, line 100:
```suggestion
The preceding [list of highest to lowest priority default configuration sources](#hi2low) shows the providers in the opposite order they are added to template generated application. For example, the [JSON configuration provider](#file-configuration-provider) is added before the [Command-line configuration provider](#command-line).
```

aspnetcore/fundamentals/configuration/index.md
```diff
@@ -116,7 +158,12 @@ For more information on storing passwords or other sensitive data:
 
 <a name="evcp"></a>
 
-## Environment variables
+## Non-prefixed environment variables
+
+Non-prefixed environment variables are environment variables other than those prefixed by `ASPNETCORE_` or `DOTNET_`. For example, the ASP.NET Core web application templates set `"ASPNETCORE_ENVIRONMENT": "Development"` in `launchSettings.json`. For more information on `ASPNETCORE_` and `DOTNET_` environment variables, see:
+
+* <xref:fundamentals/environments>
+* [`DOTNET_` environment variables](/dotnet/core/tools/dotnet-environment-variables)
```
halter73, line 170: `<xref:fundamentals/environments>` looks to be specifically about prod/dev/staging environments. While this can be configured using `DOTNET_` or `ASPNETCORE_` environment variables, it's not about environment variables themselves.

```suggestion
* [List of highest to lowest priority default configuration sources](#hi2low) including non-prefixed, `ASPNETCORE_`-prefixed and `DOTNETCORE_`-prefixed environment variables.
* [`DOTNET_` environment variables](/dotnet/core/tools/dotnet-environment-variables) used outside of [Microsoft.Extensions.Hosting](xref:Microsoft.Extensions.Hosting).
```

halter73 (approved):
This looks really good. @Tratcher said he might also take a look.

aspnetcore/fundamentals/configuration/index.md
```diff
@@ -26,37 +26,83 @@ Configuration in ASP.NET Core is performed using one or more [configuration prov
...
+<a name="host"></a>
 
-For more information on `CreateBuilder`, see [Default builder settings](xref:fundamentals/host/generic-host#default-builder-settings).
+### Default host configuration sources
+
+The following list contains the default host configuration sources from highest to lowest priority:
+
+1. [`ASPNETCORE_` prefixed environment variable](xref:fundamentals/environments)
+1.  Command-line arguments using the [Command-line configuration provider](#command-line)
+1. [`DOTNET_` prefixed environment variables](/dotnet/core/tools/dotnet-environment-variables)
```
halter73, line 66: It looks like my last comment here is outdated. I'll redo my suggestion for reasons explained in https://github.com/dotnet/AspNetCore.Docs/pull/25686/files#r868567158. Basically, these link to 

```suggestion
1. `ASPNETCORE_`-prefixed environment variables using the [Environment variables configuration provider](xref:Microsoft.Extensions.Configuration.EnvironmentVariables.EnvironmentVariablesConfigurationProvider).
1.  Command-line arguments using the [Command-line configuration provider](#command-line)
1. `DOTNET_`-prefixed environment variables using the [Environment variables configuration provider](xref:Microsoft.Extensions.Configuration.EnvironmentVariables.EnvironmentVariablesConfigurationProvider).
```

If we don't like linking directly to `EnvironmentVariablesConfigurationProvider` here, we could have `#evcp` discuss all environment variables, not just non-prefixed environment variables. It's already been updated to at least mention the prefixed environment variables.

Tratcher:
:shipit:

PR #25693, opened 2022-04-20
Fix broken link

henry-js:
Previously referenced .md file is no longer in this repository

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@ You need a basic understanding of [Git and GitHub.com](https://guides.github.com
 * If your PR has the label 'cla-required' assigned, [complete the Contribution License Agreement (CLA)](https://cla.dotnetfoundation.org/).
 * Respond to PR feedback.
 
-For an example where this process led to publication of a new article, see [Issue 1477](https://github.com/dotnet/docs/issues/1477) and [Pull Request 18955](https://github.com/dotnet/docs/pull/18955) in the .NET Docs repository. The new article is [Use code coverage for unit testing](/dotnet/core/testing/unit-testing-code-coverage).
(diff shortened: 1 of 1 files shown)


guardrex:
Hello @henry-js ... Thanks for pointing this out, but this won't be the right fix. We don't point to markdown files. Let me research it out and see what we can do here. I'll edit the PR with whatever I come up with.

serpent5:
@guardrex I _just_ made an update to fix this, then saw your comment and assignment.

guardrex:
I don't see why you're calling this one out, @henry-js. It's live at ...

https://docs.microsoft.com/dotnet/core/testing/unit-testing-code-coverage

guardrex:
HA ... nevermind ... I see we're in a markdown file. 😄 ..... **_MORE_** ☕ **_PLZ!_** ... I'm not fully awake yet.

guardrex:
@serpent5 ... Is it ok now? I just saw that we'll need the full URL to that topic.

serpent5:
@guardrex I'll sort it.

guardrex:
Yeah ... it depends if the contributing page shows up at the site. We have markdown files that don't ... that aren't reachable at docs.microsoft.com ... then we need the full URL. If they are reachable at docs.microsoft.com, then the relative URL is cool.

serpent5:
Yeah, it caught me out too. I fixed it by completely undoing the change... 🤦... Should be OK now.

guardrex (approved):

serpent5:
Thanks @henry-js.

guardrex:
Yeah ... thanks again, @henry-js! 🚀 We don't normally "FIX" a **_"FIX"_** for a **_community-contributed "FIX"_** ... but it does happen from time-to-time. 😄 The whole docs system is rather complex. There are at least *HUNDREDS* ... several hundred, probably ... little rules and *NITs* and 😈 **_gotchas_** 😈 around here. The links are one of them because it depends on where they show up and where they're going.

... add on to that a less-than-ideal amount of ☕☕☕☕ for things to get a little sketchy in the morning! 😆

PR #25704, opened 2022-04-21
Update nuget package names

D63H:
In sample code blocks, Microsoft.Http.Extensions has been replaced with Microsoft.Extensions.Http

I am still new to GitHub and creating pull requests, so I am sorry if I did something wrong


guardrex:
Thanks ... I'll get to this ASAP ... I'm **_buried_** at the moment. ⛰️⛏️😅

guardrex (approved):
Thanks @D63H! 🚀

... and this is the only spot the whole doc set. I researched it back to see how it came in this way, and it looks like it was just a typo from a community-supplied PR that was trying to let devs know which namespace to add.

PR #25731, opened 2022-04-25
Changed end-of-sentence punctuation from : to .

aleksvujic:
In the documentation about [Microsoft authentication](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/social/microsoft-logins?view=aspnetcore-6.0), there is a wrong punctuation at the end of the following sentence:

`You are now logged in using your Microsoft credentials:`

It should be:

`You are now logged in using your Microsoft credentials.`


guardrex (approved):
Thanks @aleksvujic! :rocket:

Issue #23737, opened 2021-11-04
Incomplete list of re-render triggers?


mrlife:
I am seeing re-rendering of a child component when none of the listed triggers have occurred.

If the list of triggers is missing something, can it be added? ...and if it isn't, hopefully the team will respond in the linked ticket.

ref: https://github.com/dotnet/aspnetcore/issues/38021

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: a8201d0b-a6c4-7656-a8d9-9ae1fee26d10
* Version Independent ID: 27742f9a-95e5-a2d0-aa54-0ef24cad3d2a
* Content: [ASP.NET Core Blazor component rendering](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/rendering?view=aspnetcore-6.0#rendering-conventions-for-componentbase)
* Content Source: [aspnetcore/blazor/components/rendering.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/rendering.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
It will take a while to get to this given the high priority issues right now. If you can set `ShouldRender` in the child without foul effects, that might be best (as a hack 🙈😄) for now until the PU can respond. They're all 🏃 like 😵 for the 6.0 release ... then will no doubt get slammed immediately after release with a ton of 6.0-related issues from devs adopting the new bits ... then no doubt be so 🧠 *brain fried* 🍳 that we won't hear from them until January.

guardrex:
I took a look at Artak's response to your engineering issue, and it seems that the behavior is by-design and that the docs describe the *general* behavior that they're happy with at this time. For example, [Rendering conventions for ComponentBase](https://docs.microsoft.com/aspnet/core/blazor/components/rendering#rendering-conventions-for-componentbase), aspects of rerendering mentioned in the *Lifecycle* topic, and notes on [Overwritten parameters in child components](https://docs.microsoft.com/aspnet/core/blazor/components/#overwritten-parameters). Of course, you have access to the reference source. The approach that the PU ("product unit") recommends for explicitly controlling rendering is via `ShouldRender`.

Side note: I can't download/open ZIP files here, so I can't see the app that you mentioned in the PU issue. Best to place projects for inspection at GH.

[...]

Issue #13610, opened 2019-08-01
Blazor Server App Short Name Changed in 3.0.0-preview7.19365.7


Just2Sour:
Getting Started step #3 uses the command dotnet new blazorserverside -o WebApplication1 for .NET Core CLI; however, the short name in this preview version is blazorserver.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: c07e57dc-526f-9479-4695-de7d8f06eb9a
* Version Independent ID: 8226ed72-69b2-3c93-a931-218bd11cf9c8
* Content: [Get started with ASP.NET Core Blazor](https://docs.microsoft.com/en-us/aspnet/core/blazor/get-started?view=aspnetcore-3.0&tabs=netcore-cli#feedback)
* Content Source: [aspnetcore/blazor/get-started.md](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/blazor/get-started.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Hello @Just2Sour ... That will be for Pre8. We have it on https://github.com/aspnet/AspNetCore.Docs/pull/13536.

PR #25740, opened 2022-04-26
Clarify change detection

### LINKED ISSUE #23084: What is the official list of types that Blazor can automatically detect changes in Parameters to avoid re-rendering?

guardrex:
Fixes #23084 

Thanks @szalapski! 🚀 ... Sorry for the delay. We tend to work through the backlog in H1 of each year following a hectic end-of-year release cycle and the holidays.

For the *Binding* topic, I'm adding a cross-link to the *Rendering* topic, where the subject is covered. I think it could go either way ... it could be an INCLUDES file shared in both spots; however, this subject is bit more in-the-weeds ... more implementation-ish ... than perhaps binding needs to cover directly. For most devs in most scenarios, the change detection should ✨ **_Just Work!_**&trade; ✨. The *Rendering* topic comes up fairly fast in the reading list down the ToC, so they'll get there PDQ anyway. Let's try this. If more feedback comes in on the *Binding* topic asking for more detail earlier, then I'll make further mods.

Thanks again, @szalapski. Although the coverage is rather .... uh .... **_vague_** 😆 ... I hope this answers your question(s) and gets you to the right part of the reference source to see the framework's magic bits that make it happen.

### DIFF
--- a/aspnetcore/blazor/components/data-binding.md
+++ b/aspnetcore/blazor/components/data-binding.md
@@ -262,6 +262,7 @@ For an alternative approach suited to sharing data in memory and across componen
 
 ## Additional resources
 
+* [Parameter change detection and additional guidance on Razor component rendering](xref:blazor/components/rendering)
 * <xref:blazor/forms-validation>
 * [Binding to radio buttons in a form](xref:blazor/forms-validation#radio-buttons)
 * [Binding `InputSelect` options to C# object `null` values](xref:blazor/forms-validation#binding-inputselect-options-to-c-object-null-values)
@@ -457,6 +458,7 @@ For an alternative approach suited to sharing data in memory and across componen
 
 ## Additional resources
 
+* [Parameter change detection and additional guidance on Razor component rendering](xref:blazor/components/rendering)
 * <xref:blazor/forms-validation>
 * [Binding to radio buttons in a form](xref:blazor/forms-validation#radio-buttons)
 * [Binding `InputSelect` options to C# object `null` values](xref:blazor/forms-validation#binding-inputselect-options-to-c-object-null-values)
@@ -652,6 +654,7 @@ For an alternative approach suited to sharing data in memory and across componen
 
 ## Additional resources
 
+* [Parameter change detection and additional guidance on Razor component rendering](xref:blazor/components/rendering)
 * <xref:blazor/forms-validation>
 * [Binding to radio buttons in a form](xref:blazor/forms-validation#radio-buttons)
 * [Binding `InputSelect` options to C# object `null` values](xref:blazor/forms-validation#binding-inputselect-options-to-c-object-null-values)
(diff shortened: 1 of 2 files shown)


szalapski:
Thanks for these changes.  

>"For most devs in most scenarios, the change detection should ✨ Just Work!™ ✨. The Rendering topic comes up fairly fast in the reading list down the ToC, so they'll get there PDQ anyway."

I find this is perhaps too sweeping a statement and I doubt it to an extent.  I have [outlined the problem](http://www.szalapski.com/2022/03/is-your-blazor-site-rendering-thousands.html) and [my solution](http://www.szalapski.com/2022/03/avoiding-unnecessary-rendering-in-blazor.html) in these blog posts.  IF you have time to read them and think about them, I'd love to get your feedback and thoughts.

guardrex:
I don't agree based on doc feedback. Blazor is three years old, and this subject came up [only one other time](https://github.com/dotnet/AspNetCore.Docs/issues/23737). There was nothing to change for that issue based on the engineering remarks at the time. Idk tho for the PU more generally. They open issues for me when they get hit with the same question over and over. I think for most devs in most cases, they're getting the behavior for change detection that they want and/or expect OOB. Besides, the update now has the link to reference source, so they can see exactly what the framework is doing.

What you might do if you feel the framework isn't behaving well in at least some cases is open an issue with them to discuss it. If you do, please add ...

```
cc: @guardrex https://github.com/dotnet/AspNetCore.Docs/issues/23084
```

... to the bottom of your opening comment so that I follow along. I might re-open the original issue or create a new docs issue for doc work depending on what they say.

szalapski:
I think #13610 is the existing issue; I'll give more feedback there.

As for the doc, I appreciate the change, but I wish more examples of the types in question could be listed.  Just `int`, `string`, and `DateTime` really isn't enough.

szalapski:
Hmm, I am stuck.  https://github.com/dotnet/aspnetcore/issues/13610 is the existing issue but it doesn't allow comments in favor of https://github.com/dotnet/aspnetcore/issues/22432 which is much too broad and also closed.  How can I ask to reopen 13610 ?

szalapski:
> "Great, I  would like to add a patch PR to this PR for that API doc....UPDATE: Done!"

Where can I see this change?

guardrex:
It's just a cross-link ... nothing significant, really. It's on ...

https://github.com/dotnet/AspNetCore.Docs/pull/25741

... which is still open. If you see any other low-hanging fruit, you can pitch it before I merge that.

PR #25741, opened 2022-04-26
Improve Blazor change detection description

guardrex:
Addresses #23084

### DIFF
--- a/aspnetcore/blazor/components/rendering.md
+++ b/aspnetcore/blazor/components/rendering.md
@@ -28,17 +28,10 @@ By default, Razor components inherit from the <xref:Microsoft.AspNetCore.Compone
 
 Components inherited from <xref:Microsoft.AspNetCore.Components.ComponentBase> skip rerenders due to parameter updates if either of the following are true:
 
-* All of the parameters are of the following immutable types, whether nullable or nonnullable, and haven't changed since the previous set of parameters were set:
-
-  * <xref:System.DateTime>
-  * [`decimal`](xref:System.Decimal)
-  * <xref:System.Guid>
-  * [`string`](xref:System.String)
-  * <xref:System.Type>
-  * Any [primitive type](/dotnet/api/system.type.isprimitive)
+* All of the parameters are of known immutable types or any [primitive type](/dotnet/api/system.type.isprimitive), whether nullable or nonnullable, and haven't changed since the previous set of parameters were set.
 
   > [!NOTE]
-  > Blazor's framework uses a set of built-in rules, including the list of parameter types, for change detection. These rules are subject to change at any time. For more information, see the [`ChangeDetection` API in the ASP.NET Core reference source](https://github.com/dotnet/aspnetcore/blob/main/src/Components/Components/src/ChangeDetection.cs).
+  > Blazor's framework uses a set of built-in rules and parameter types for change detection. These rules and the types are subject to change at any time. For more information, see the [`ChangeDetection` API in the ASP.NET Core reference source](https://github.com/dotnet/aspnetcore/blob/main/src/Components/Components/src/ChangeDetection.cs).
   >
   > [!INCLUDE[](~/includes/aspnetcore-repo-ref-source-links.md)]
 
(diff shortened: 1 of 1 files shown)


guardrex:
@szalapski .... I agree ... let's try primitive and complex types. Just a sec ... I'm working on **THREE things** at once again 😨, which never goes well 🙈 with the docs!

guardrex:
mmmmmmmmmmmm 🤔 ... nah ... `DateTime` is a struct value type.

Let me come back to this tomorrow, Wednesday. I'm buried in other things at the moment and want to compose this better for the framework's logic with accurate wording and perhaps additional cross-links to API and/or C# Guide topics.

I'll mark this **_DRAFT_** for the moment.

szalapski:
Besides `string` and `DateTime`, what non-primitive types are in play here?  That would help me reason about this more.

guardrex:
The ones in the API AFAICT ...

https://github.com/dotnet/aspnetcore/blob/main/src/Components/Components/src/ChangeDetection.cs#L39-L43

Ok ... I'm going to return to this tomorrow ... I think ... if the sky doesn't fall between now and then. We have high priority Blazor Hybrid things to do these days ... and other emergencies come up regularly around here 🚑🚒🚓.

szalapski:
PERFECT.  Finally, the actual answer.
So I'd propose just being explicit and fully correct, perhaps:

* All of the parameters are of the types `string`, `decimal`, `DateTime`, `Guid`, the `Type` type itself, or any [primitive type](/dotnet/api/system.type.isprimitive), whether nullable or nonnullable), and haven't changed since the previous set of parameters were set.

guardrex:
That's already going in on the other PR, where it will say ...

> Blazor's framework uses a set of built-in rules for change detection, which are subject to change at any time. For more information, see the \[\`ChangeDetection\` API in the ASP.NET Core reference source](https://github.com/dotnet/aspnetcore/blob/main/src/Components/Components/src/ChangeDetection.cs).

szalapski:
Ah, I would not have associated that with where I need to go to see the detail of what types are relevant.  Can the two be tied together somehow, or the link repeated?

guardrex:
I don't follow ... "two" what? ... "repeated" where?

szalapski:
The two parts that we are talking about:  the documentation on the types and the link to show the detail of the types.  When in different paragraphs, the reader would not think to associate the two.

guardrex:
Yes, that's right. I was too 🏃 to look again at the code yesterday. I see what you mean now. I'll move the link and remarks on change detection API directly to the bullet point.

guardrex:
Last commit has the explicit list, but I really, *REALLY* don't like it. It's going to break 💥😢. I'm going to have to go 'see the API' on this one.

guardrex:
Ok ... I feel a lot better about the latest update. I'd like to leave this where it *probably* won't 💥 .... not where it probably will 😢. *Future generations will thank us!*

I opened an issue with Bill to ask what we can link the word "immutable" to at ...

**Immutable (vs mutable) types defined**
https://github.com/dotnet/docs/issues/29187

... so we'll hold here a sec to see what he says.

guardrex:
Yes, I think it will go with a *see note* approach, but it will use a dagger (&dagger;) to connect it. I'm awaiting feedback from Bill Wagner on this, and I'm going to go with whatever text he suggests. I'm also asking him if for the word "known" that this can say that's a type that implements `System.Type`.

guardrex:
Ok ... it's going to be discussed further for the C# Guide. We can go ahead here for now, and I'm going to open a follow-up docs issue to see what they figure out over there. The two concepts that we want to cross-link are ...

* known vs. unknown types
* immutable vs. mutable types

We'll let them figure out what to say and where to say it. I'll circle around on the new issue to cross-link these words, or I'll add a 'For more info, see ...'-type of remark.

The NOTE is attached to the bullet point, so the ref source cross-link isn't some disconnected bit of guidance.

The Binding topic now has a cross-link to this topic with suitable link text ...

> \[Parameter change detection and additional guidance on Razor component rendering](xref:blazor/components/rendering)

Ok ... I think we're good for now. Thanks again for your help, @szalapski! 🎸

PR #25743, opened 2022-04-26
Ra/shawnwilde/static files

### LINKED ISSUE #25744: asp-append-version only works for StaticFiles in wwwroot
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When I am adding a new `AddStaticFiles` middleware, the asp-append-version attribute only works if the file is in **wwwroot**.  My repro case has two AddStaticFiles:

```csharp
app.UseStaticFiles();
app.UseStaticFiles(new StaticFileOptions()
{
  FileProvider = new PhysicalFileProvider(
    Path.Combine(builder.Environment.ContentRootPath, @"Other"))
});
```

Then my index.cshtml has:

```razor
  <script src="~/js/site.js" asp-append-version="true"></script>
  <script src="~/site.js" asp-append-version="true"></script>
```

- The first script works fine (it is in `wwwroot/js/`).
- The second script renders without a version (it is in `/Other/`);

I suspect that this line in the DefaultFileVersionProvider is at fault:

https://github.com/dotnet/aspnetcore/blob/c85baf8db0c72ae8e68643029d514b2e737c9fae/src/Mvc/Mvc.Razor/src/Infrastructure/DefaultFileVersionProvider.cs#L72

### Expected Behavior

Both .js files should have versions (or at least a console message explaining why it isn't).

### Steps To Reproduce

See the repro here:

https://github.com/shawnwildermuth/AppendVersionRepro

Should see that the second .js file doesn't not get version appended.

### Exceptions (if any)

None.

### .NET Version

6.0.202

### Anything else?

Just using default ASP.NET Core version with .NET 6:

```
<Project Sdk="Microsoft.NET.Sdk.Web">

  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
    <Nullable>disable</Nullable>
    <ImplicitUsings>enable</ImplicitUsings>
  </PropertyGroup>

  <ItemGroup>
    <Content Include="Other\site.js">
      <ExcludeFromSingleFile>true</ExcludeFromSingleFile>
      <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
    </Content>
  </ItemGroup>

</Project>

```

Rick-Anderson:
Fixes #25744
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/static-files?view=aspnetcore-6.0&branch=pr-en-us-25743#serve-files-from-multiple-locations)


aspnetcore/fundamentals/static-files/samples/6.x/StaticFilesSample/Program.cs
```diff
@@ -385,7 +385,51 @@
...
+// Update the default provider.
+app.Environment.WebRootFileProvider = compositeProvider;
+
+app.UseStaticFiles();
+
+app.UseStaticFiles(new StaticFileOptions()
+{
+    FileProvider = new PhysicalFileProvider(
+             Path.Combine(builder.Environment.ContentRootPath, "MyStaticFiles"))
 });
```
Tratcher, line 433: You shouldn't need a second call to UseStaticFiles, the first one will look at WebRootFileProvider and check both locations from the CompositeFileProvider.

Tratcher (commented):
I see one doc that mentions this feature only works for content in wwwroot. @mkArtakMSFT are there other docs about customizing this?
https://docs.microsoft.com/en-us/aspnet/core/mvc/views/tag-helpers/built-in/image-tag-helper?view=aspnetcore-6.0#asp-append-version

Rick-Anderson:
@Tratcher should I just delete the sample that shows calling `UseStaticFiles` twice and ***doesn't*** work with tag helper versions? It's a little simpler and I'm guessing most folks adding additional directories aren't using Tag Helper version stamping.

Tratcher:
Having two calls to UseStaticFiles is fine for most usages, and the most flexible since you can change other settings.

Rick-Anderson:
> I see one doc that mentions this feature only works for content in wwwroot. @mkArtakMSFT are there other docs about customizing this? https://docs.microsoft.com/en-us/aspnet/core/mvc/views/tag-helpers/built-in/image-tag-helper?view=aspnetcore-6.0#asp-append-version

Script TH too. I'll look for others with version. See #25760

PR #25746, opened 2022-04-26
Add DI to SignalR hub methods

BrennanConroy:
Documenting feature from https://github.com/dotnet/aspnetcore/pull/34047

### DIFF
--- a/aspnetcore/signalr/configuration.md
+++ b/aspnetcore/signalr/configuration.md
@@ -12,7 +12,7 @@ uid: signalr/configuration
 
 # ASP.NET Core SignalR configuration
 
-:::moniker range=">= aspnetcore-6.0"
+:::moniker range="= aspnetcore-6.0"
 
 ## JSON/MessagePack serialization options
 
@@ -71,7 +71,6 @@ The following table describes options for configuring SignalR hubs:
 | `StreamBufferCapacity` | `10` | The maximum number of items that can be buffered for client upload streams. If this limit is reached, the processing of invocations is blocked until the server processes stream items.|
 | `MaximumReceiveMessageSize` | 32 KB | Maximum size of a single incoming hub message. |
 | `MaximumParallelInvocationsPerClient` | 1 | The maximum number of hub methods that each client can call in parallel before queueing. |
-| `DisableImplicitFromServicesParameters` | false | Hub method arguments will be resolved from DI if possible. |
 
 Options can be configured for all hubs by providing an options delegate to the `AddSignalR` call in `Startup.ConfigureServices`.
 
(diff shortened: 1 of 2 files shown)


aspnetcore/signalr/configuration.md
```diff
@@ -71,6 +71,7 @@ The following table describes options for configuring SignalR hubs:
 | `StreamBufferCapacity` | `10` | The maximum number of items that can be buffered for client upload streams. If this limit is reached, the processing of invocations is blocked until the server processes stream items.|
 | `MaximumReceiveMessageSize` | 32 KB | Maximum size of a single incoming hub message. |
 | `MaximumParallelInvocationsPerClient` | 1 | The maximum number of hub methods that each client can call in parallel before queueing. |
+| `DisableImplicitFromServicesParameters` | false | Hub method arguments will be resolved from DI if possible. |
```
bradygaster, line 74: 👍

bradygaster (approved):

aspnetcore/signalr/configuration.md
guardrex, line 74: To avoid localization by MT ...

```suggestion
| `DisableImplicitFromServicesParameters` | `false` | Hub method arguments will be resolved from DI if possible. |
```

aspnetcore/signalr/hubs.md
```diff
@@ -120,6 +120,56 @@ By default, a server hub method name is the name of the .NET method. To change t
 
 :::code language="csharp" source="~/../AspNetCore.Docs.Samples/signalr/hubs/samples/6.x/SignalRHubsSample/Snippets/Hubs/ChatHub.cs" id="snippet_HubMethodName" highlight="1":::
 
+## Inject services into your hub
+
+> [!NOTE]
+> The following feature is only available in .NET 7.0+
```
BrennanConroy, line 126: Did this instead of duplicating everything and using a moniker range. AFAIK there is not way to nest moniker ranges.

aspnetcore/signalr/hubs.md
```diff
@@ -120,6 +120,56 @@ By default, a server hub method name is the name of the .NET method. To change t
 
 :::code language="csharp" source="~/../AspNetCore.Docs.Samples/signalr/hubs/samples/6.x/SignalRHubsSample/Snippets/Hubs/ChatHub.cs" id="snippet_HubMethodName" highlight="1":::
 
+## Inject services into your hub
```
guardrex, line 123: To avoid 2nd person where possible ...

```suggestion
## Inject services into a hub
```

or it could be "Hub service injection" ... or similar.

aspnetcore/signalr/configuration.md
BrennanConroy: How should we handle this? It's a new option in 7.0 but will show up in 6.0 docs since we're trying to stay away from moniker hell.

aspnetcore/signalr/hubs.md
guardrex, line 126: I recommend going ahead and versioning the whole topic at this point. Rick will activate it in the version selector. However, be sure to put the new 7.0 piece at the **_end of the topic_** to avoid broken links during preview.

aspnetcore/signalr/hubs.md
```diff
@@ -120,6 +120,56 @@ By default, a server hub method name is the name of the .NET method. To change t
...
+
+> [!NOTE]
+> The following feature is only available in .NET 7.0+
+
+Hub constructors can accept services from DI as parameters which can be stored in properties on the class for use in a hub method.
```
guardrex, line 128:
```suggestion
Hub constructors can accept services from DI as parameters, which can be stored in properties on the class for use in a hub method.
```

aspnetcore/signalr/hubs.md
```diff
@@ -120,6 +120,56 @@ By default, a server hub method name is the name of the .NET method. To change t
...
+
+When injecting multiple services for different hub methods or as an alternative way of writing your code, hub methods can also accept services from DI.
```
guardrex, line 130: Since we favor imperative tone and dropping 2nd person where possible ...

```suggestion
When injecting multiple services for different hub methods or as an alternative way of writing code, hub methods can also accept services from DI.
```

aspnetcore/signalr/hubs.md
```diff
@@ -120,6 +120,56 @@ By default, a server hub method name is the name of the .NET method. To change t
...
+By default hub method parameters will be inspected and resolved from DI if possible.
```
guardrex, line 131: Avoiding past tense where possible ...

```suggestion
By default, hub method parameters are inspected and resolved from DI if possible.
```

aspnetcore/signalr/hubs.md
```diff
@@ -120,6 +120,56 @@ By default, a server hub method name is the name of the .NET method. To change t
...
+{
+    public Task SendMessage(string user, string message, IDatabaseService dbService)
+    {
+        var userName = dbService.GetUserName(user);
+        return Clients.All.SendAsync("ReceiveMessage", userName, message);
+    }
+}
+```
+
+If this behavior is undesired it can be disabled with [DisableImplicitFromServicesParameters](xref:signalr/configuration#configure-server-options).
```
guardrex, line 148: Prob best to restate the noun over using a pronoun. How about ...

```suggestion
If implicit resolution of parameters from services isn't desired, disable it with [`DisableImplicitFromServicesParameters`](xref:signalr/configuration#configure-server-options).
```

aspnetcore/signalr/hubs.md
```diff
@@ -120,6 +120,56 @@ By default, a server hub method name is the name of the .NET method. To change t
...
+If you want to be explicit about what parameters get resolved from DI in your hub methods, you can use the `DisableImplicitFromServicesParameters` option and use the `[FromServices]` attribute or a custom attribute that implements `IFromServiceMetadata` on your parameters.
```
guardrex, line 149:
```suggestion
To explicitly specify which parameters are resolved from DI in hub methods, use the `DisableImplicitFromServicesParameters` option and use the `[FromServices]` attribute or a custom attribute that implements `IFromServiceMetadata` on parameters.
```

aspnetcore/signalr/hubs.md
```diff
@@ -120,6 +120,56 @@ By default, a server hub method name is the name of the .NET method. To change t
...
+services.AddSignalR(options =>
+{
+    options.DisableImplicitFromServicesParameters = true;
+});
+
+// ...
+
+public class ChatHub : Hub
+{
+    public Task SendMessage(string user, string message, [FromServices] IDatabaseService dbService)
```
guardrex, line 162: GH won't count the chars for me, but this might be a hair long. If it's over 85, I suggest ...

```suggestion
    public Task SendMessage(string user, string message, 
        [FromServices] IDatabaseService dbService)
```

aspnetcore/signalr/hubs.md
```diff
@@ -120,6 +120,56 @@ By default, a server hub method name is the name of the .NET method. To change t
...
+    {
+        var userName = dbService.GetUserName(user);
+        return Clients.All.SendAsync("ReceiveMessage", userName, message);
+    }
+}
+```
+
+> [!NOTE]
+> This feature makes use of [IServiceProviderIsService](/dotnet/api/Microsoft.Extensions.Dependencyinjection.IServiceProviderIsService) which is optionally implemented by Dependency Injection implementations. If the DI container used by your app does not support this feature then injecting services into your hub methods will not work.
```
guardrex, line 171:
```suggestion
> This feature makes use of <xref:Microsoft.Extensions.DependencyInjection.IServiceProviderIsService>, which is optionally implemented by DI implementations. If the app's DI container doesn't support this feature, injecting services into hub methods isn't supported.
```

guardrex (approved):
Either the topic should be versioned now for 7.0, or a tracking issue for 7.0 for this topic should sit to version the content later. Use of NOTE for versioning isn't our normal approach, and I think it will end in tears 😢 lol.

aspnetcore/signalr/configuration.md
guardrex: The whole topic should be versioned. If not the whole topic, then we're forced to version the **whole table**. Can do individual rows of a table.

aspnetcore/signalr/hubs.md
guardrex: Correct ... they can't be nested. I recommend versioning the whole topic now and signaling Rick to activate the version selector for the whole doc set. Place the new versioned content at the **end of the article tho**. Otherwise, it will break bookmark links during preview.

aspnetcore/signalr/configuration.md
BrennanConroy: We can't version just a topic, the whole page has to be duplicated and versioned

guardrex: That's what I meant ... the content for `aspnetcore-7.0` ..... or just the whole table here. We still go that way when there are only a few version flips.

aspnetcore/signalr/hubs.md
guardrex: I see ... then there's a stray CRLF at the end of Line148.

guardrex: ... oh ... Line 640.

Actually, I think without **two spaces** at the end of the line, it will end up rendering as one paragraph anyway ... even if the CRLF stays there. The DocFX trick that we use to merely place a `<br>` is two spaces. Otherwise, I think DocFX is going to make `<p>`s out these two lines.

BrennanConroy: Sounds good. Looks like it's a paragraph on the review site.

BrennanConroy:
Any more feedback or is this good to go?

aspnetcore/signalr/configuration.md
```diff
@@ -12,7 +12,7 @@ uid: signalr/configuration
 
 # ASP.NET Core SignalR configuration
 
-:::moniker range=">= aspnetcore-6.0"
+:::moniker range="= aspnetcore-6.0"
```
BrennanConroy, line 15:
```suggestion
:::moniker range="< aspnetcore-7.0"
```

aspnetcore/signalr/hubs.md
```diff
@@ -12,7 +12,7 @@ uid: signalr/hubs
 
 # Use hubs in SignalR for ASP.NET Core
 
-:::moniker range=">= aspnetcore-6.0"
+:::moniker range="= aspnetcore-6.0"
```
BrennanConroy, line 15:
```suggestion
:::moniker range="< aspnetcore-7.0"
```

guardrex:
🚢 **IT**

BrennanConroy:
That doesn't work... < 7.0 means 5.0, 2.1, etc. will also still run, so they would need to be updated as well

guardrex:
I'll have to look, but yes ... it shouldn't throw that unique heading warning merely for dup headings. It is versioning. I'll take a look.

guardrex:
Oh ... ok ... I see. Yes, it conflicts with the other blocks **that I didn't notice**. Revert the last change ... you were good!

Can I blame just getting back from the DENTIST for this??!! 🤕

guardrex:
# 😅

Ok ... build system is happy with it now.

Issue #25776, opened 2022-04-29
Immutable/mutable & known/unknown C# Guide cross-links


guardrex:
Tracking on https://github.com/dotnet/docs/issues/29187 for cross-link opportunities ... or perhaps a bit of text that makes sense to use for the line ...

> * All of the parameters are known immutable types or any \[primitive type](/dotnet/api/system.type.isprimitive) that hasn't changed since the previous set of parameters were set.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: a8201d0b-a6c4-7656-a8d9-9ae1fee26d10
* Version Independent ID: 27742f9a-95e5-a2d0-aa54-0ef24cad3d2a
* Content: [ASP.NET Core Razor component rendering](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/rendering?view=aspnetcore-6.0)
* Content Source: [aspnetcore/blazor/components/rendering.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/rendering.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
LICK'D IT 🐮👅 ... I mustn't take any personal credit, of course, it's entirely due to Bill Wagner's influence!!! 🤣 ... to misquote a Tom Baker Dr. Who line! 😄

@szalapski ... Bill says to use the following ...

> All of the parameters are [verifiably immutable types](/dotnet/csharp/language-reference/builtin-types/unmanaged-types) or any [primitive type](/dotnet/api/system.type.isprimitive) that hasn't changed since the previous set of parameters were set.

... so I'll shoot that in right now.

szalapski:
That still seems wrong, as those lists don't include string or DateTime.

guardrex:
Ask Bill these questions on the issue that I opened over there ...

https://github.com/dotnet/docs/issues/29187

PR #25780, opened 2022-04-29
Update parameter change detection remarks

guardrex:
Addresses #25776

cc: @szalapski

This text is plagued with a nasty **_gremlin_** 😈! 😄 I'll ping Steve on Monday to discuss this further and review the PR.

If we can just say 'set of known types' and 'see the ref source' for them, that's really the best. I've been around here five years now, and I'm telling u that it will all end in **_tears_** 😢 if we try to explicitly list these types ... devs will get 😠 after they churn ... I'll forever need to be looking over my shoulder everywhere I go for the rest of my life! 😨😆


kacharyea (approved):

guardrex:
I'm not going to call Steve in on this one at this time 🏃. I think we've resolved this. Bill took a look at this text on the other issue and didn't flag a concern about it. Let's see if additional feedback comes in 👂. Thanks everyone!

PR #25781, opened 2022-04-30
fix grpc loadbalancing doc based on current APIs

MoienTajik:
Fixed gRPC load balancing docs based on the current APIs.


JamesNK (approved):

guardrex:
Thanks @MoienTajik! :rocket:

PR #25825, opened 2022-05-10
Update doc with gRPC JSON transcoding

JamesNK:
gRPC JSON transcoding is coming in .NET 7 preview 4. This updates the doc with the right package name and makes various other updates.

PR #25852, opened 2022-05-12
Update httpapi sample with gRPC JSON transcoding

ArminShoeibi:
Hello there folks.
According to this PR https://github.com/dotnet/AspNetCore.Docs/pull/25825
I updated the sample with a newer code.

### DIFF
--- a/aspnetcore/grpc/httpapi/Startup.cs
+++ b/aspnetcore/grpc/httpapi/Startup.cs
@@ -1,14 +1,11 @@
 #region snippet_1
 public void ConfigureServices(IServiceCollection services)
 {
-    services.AddGrpc();
-    services.AddGrpcHttpApi();
-    
+    services.AddGrpc().AddJsonTranscoding();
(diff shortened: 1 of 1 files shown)


ArminShoeibi:
I have a question, Do we need to change the folder name to something else?
https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/grpc/httpapi

JamesNK:
Thanks. Where is this file used from?

ArminShoeibi:
> Thanks. Where is this file used from?

Thank you for second commit.
I think we can't change the folder name because it has some references in other issues and pull requests

Rick-Anderson:
@ArminShoeibi  I can't see where this file is used in the documents so I'm inclined to close the PR.

ArminShoeibi:
> @ArminShoeibi I can't see where this file is used in the documents so I'm inclined to close the PR.

Okay thank you

PR #25894, opened 2022-05-18
Update building-net-docker-images.md

Rust1k:
It seems to me that here are the typos


Rust1k:
Maybe I don't understand something, but this code only copies the dll file, while other files, such as *.json, are not copied. So when I try to run the image, I get the following error message
```
Cannot use file stream for [/app/aspnetapp.deps.json]: No such file or directory

A fatal error was encountered. The library 'libhostpolicy.so' required to execute the application was not found in '/app/'.

Failed to run as a self-contained app.

   - The application was run as a self-contained app because '/app/aspnetapp.runtimeconfig.json' was not found.

   - If this should be a framework-dependent app, add the '/app/aspnetapp.runtimeconfig.json' file and specify the appropriate framework.
```

Rick-Anderson (approved):

PR #25901, opened 2022-05-19
Add gRPC performance section on flow control

JamesNK:
From internal performance investigation.

Internal preview: https://review.docs.microsoft.com/en-us/aspnet/core/grpc/performance?view=aspnetcore-3.0&branch=pr-en-us-25901

PR #25911, opened 2022-05-19
Flow control feedback

JamesNK:
Follow up feedback on https://github.com/dotnet/AspNetCore.Docs/pull/25901

### DIFF
--- a/aspnetcore/grpc/performance.md
+++ b/aspnetcore/grpc/performance.md
@@ -180,7 +180,7 @@ builder.WebHost.ConfigureKestrel(options =>
 Recommendations:
 
 * If a gRPC service often receives messages larger than 96 KB, Kestrel's default stream window size, then consider increasing the connection and stream window size.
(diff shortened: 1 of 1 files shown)


halter73 (approved):

aspnetcore/grpc/performance.md
```diff
@@ -177,10 +177,15 @@ builder.WebHost.ConfigureKestrel(options =>
 });
 ```
 
+Recommendations:
+
+* If a gRPC service often receives messages larger than 96 KB, Kestrel's default stream window size, consider increasing the connection and stream window size.
+* The connection window size should always be equal or greater than the stream window size. A stream is part of the connection and the sender is limited by both.
```
guardrex, line 183:
```suggestion
* The connection window size should always be equal to or greater than the stream window size. A stream is part of the connection, and the sender is limited by both.
```

guardrex (approved):
I think just a missing "to" there.

Issue #25837, opened 2022-05-11
Re-add documentation on `Microsoft.AspNetCore.Grpc.HttpApi`


adeal:
A change (https://github.com/dotnet/AspNetCore.Docs/pull/25825) was made to this page to add information about gRPC JSON transcoding. However, this removed the existing info for `Microsoft.AspNetCore.Grpc.HttpApi`. Since json transcoding is a new dotnet 7 feature, is it possible to have the old details for net5/net6 alongside this new info?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: dfa3c933-30c4-eb49-d9a5-507aff44ce00
* Version Independent ID: 1485c456-eb24-a446-e296-dd3686456837
* Content: [gRPC JSON transcoding](https://docs.microsoft.com/en-us/aspnet/core/grpc/httpapi)
* Content Source: [aspnetcore/grpc/httpapi.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/grpc/httpapi.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-grpc**
* GitHub Login: @JamesNK
* Microsoft Alias: **jamesnk**

JamesNK:
GrpcHttpApi is an experiment that's getting replaced. I don't think docs.microsoft.com is the right place to keep docs for it long term. However, there is some docs at https://github.com/aspnet/AspLabs/tree/main/src/GrpcHttpApi. Does it have all the information you are looking for?

We could add a link to the GitHub page from https://docs.microsoft.com/en-us/aspnet/core/grpc/httpapi and remove it in a couple of years once .NET 7 and greater is more widely available.

wadepickett:
@JamesNK, [gRPC JSON transcoding](https://docs.microsoft.com/en-us/aspnet/core/grpc/httpapi) is not version specific yet.  I can give it a <= 6.0 version that includes https://docs.microsoft.com/en-us/aspnet/core/grpc/httpapi  and 7.0 version that does not include it.  That work?

JamesNK:
Sure. That sounds like a good idea.

adeal:
[...]

PR #25920, opened 2022-05-21
gRPC JSON transcoding: Move openAPIto gRPC HTTP API doc and link out

wadepickett:
[Internal Review](https://review.docs.microsoft.com/en-us/aspnet/core/grpc/httpapi?view=aspnetcore-7.0&branch=pr-en-us-25920)

Fixes: #25837 

Adding a note on gRPC HTTP API and linking out to its gitHub readme.


wadepickett:
@JamesNK , I kept this short, but in a note where it would not be missed.  Alternatively, I could add that gRPC JSON transcoding was introduced with .NET 7.0 and that gRPC HTTP API continues to be available for previous versions of .NET, or something of that sort.

JamesNK (approved):

PR #25924, opened 2022-05-23
Correct typo error

rjobidon:
"Preceding" instead of "Precedng"


guardrex (approved):
Thanks @rjobidon! :rocket:

PR #25928, opened 2022-05-23
Update block quote bare returns

guardrex:
Note to @Rick-Anderson and @wadepickett on this subject in passing:

I have several spots where I show console output in block quotes. I thought that in the the olden days of docs 👴, one could just CRLF at the ends of the lines for bare returns in the output (`<br>`) ...

```
> Blah blah blah.
> Blah blah blah 2.
> Blah blah blah 3.
```

... would render as ...

> Blah blah blah.
> Blah blah blah 2.
> Blah blah blah 3.

 Whether or not I'm recalling that correctly, it certainly isn't working today. The rendered output is ...

> Blah blah blah.Blah blah blah 2.Blah blah blah 3.

Soooo ... we're back to our usual rock-and-hard-place choice: Use HTML bare returns (`<br>`) (which are strictly required in table cells) or use markdown's double-space at the end of the line. I'm going with the double-space approach, as I've done elsewhere. I'm going to leave a hidden comment at the top of any topic that adopts this approach to surface it to future doc authors.

### DIFF
--- a/aspnetcore/blazor/fundamentals/logging.md
+++ b/aspnetcore/blazor/fundamentals/logging.md
@@ -11,6 +11,12 @@ uid: blazor/fundamentals/logging
 ---
 # ASP.NET Core Blazor logging
 
+<!--
+
+NOTE: The console output block quotes in this topic use a double-space at the ends of lines to generate a bare return in block quote output.
+
+-->
+
 This article explains logging in Blazor apps, including configuration and how to write log messages from Razor components.
 
 :::moniker range=">= aspnetcore-6.0"
--- a/aspnetcore/blazor/fundamentals/logging.md
+++ b/aspnetcore/blazor/fundamentals/logging.md
@@ -78,7 +78,7 @@ logger.LogWarning("Someone has clicked me!");
 
 Developer tools console output:
 
-> :::no-loc text="warn: CustomCategory[0]":::
+> :::no-loc text="warn: CustomCategory[0]":::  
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
It stopped working sometime ago, PITA, so `<br>` is the best option.

guardrex:
The double-space thing works tho. That's what I just used. For example, scroll to the bottom of ...

https://docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-dotnet-from-javascript?view=aspnetcore-6.0#call-net-generic-class-methods

It will be highlighted to future doc authors by an HTML comment at the top of these topics when the double-space thing is in use. I was merely favoring it because they don't seem to care much for HTML in docs and seem a bit more friendly/accepting to legitimate markdown per markdown spec.

Rick-Anderson:
https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/javascript-interoperability/call-dotnet-from-javascript.md?plain=1#L427

guardrex:
Of course, we're still stuck with `<br>` in table cells ... and other markdown things that don't work in table cells.

I'll go with double-spaces and my HTML comment until you let me know that the team has worked out another approach. For now, I'm saying at the tops of these topics under the H1 ...

```
<!--
NOTE: The console output block quotes in this topic use a double-space at the ends of lines to generate a bare return in block quote output.
-->
```

PR #25964, opened 2022-05-25
Improve overwritten parameter guidance

### LINKED ISSUE #23704: Why render/not render when having/not having ChildContent set?

guardrex:
Fixes #23704

Thanks @omuleanu! :rocket:

Referenced content:

https://docs.microsoft.com/aspnet/core/blazor/components/rendering#rendering-conventions-for-componentbase

The updates to the *Overwritten parameters* section doesn't try to list the explicit parameters that are checked because that's subject to change over time. I think it best to refer to the reference source to harden the doc. Devs can then 👀 the exact list of param types checked for any given release, and the PU is free to change the list at any time without breaking the doc.

I'm moving the *Overwritten parameters* one section lower, below the *Child content* section. The *Child content* section introduces the `RenderFragment`, so that should appear first.

I'm also improving cross-links.

I considered moving the whole example from the Components overview to the newer Rendering topic, but change detection/rendering for state changes is a foundational concept for component parameters. I think it's important enough to leave here. The further details exist in the *Rendering* topic that devs will reach later in their reading (or via the cross-link that I'm adding).

**UPDATE**: Yikes! That was a lot of churn, but I think 🤔 I'm happy with it now.

### DIFF
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -519,7 +519,6 @@ Consider the following `Expander` component that:
 
 * Renders child content.
 * Toggles showing child content with a component parameter (`Expanded`).
-* The component writes directly to the `Expanded` parameter, which demonstrates the problem with overwritten parameters and should be avoided.
 
 After the following `Expander` component demonstrates the incorrect approach for this scenario, a modified `Expander` component is shown to demonstrate the correct approach. The following examples can be placed in a local sample app to experience the behaviors described.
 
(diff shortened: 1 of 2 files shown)


guardrex:
@Rick-Anderson @wadepickett ... Something has gone sideways here with the build system ...

> Line 0: [Error] Failure to clone the repository `https://github.com/xamarin/xamarin-forms-samples#main`. This could be caused by an incorrect repository URL, please verify the URL on the Docs Portal (https://ops.microsoft.com). This could also be caused by not having the proper permission the repository, please confirm that the GitHub group/team that triggered the build has access to the repository.
>
> Line 0: [Error] Can't find branch, tag, or commit 'main' for repo https://github.com/Microsoft/templates.docs.msft.

..... **_OR_** .....

*THIS IS IT!* The **_ATTACK ON EARTH_** 👽🛸 has begun! Run for your LIVES! 🏃🏃🏃

PR #25965, opened 2022-05-25
Fixed incorrect lower case letter, and incorrect plural

mcasey329:
Minimum Host and Generic House are proper nouns in the document like Jim Smith and Bob Smith are proper nouns. 

You would not say "Jim and Bob smiths" -- you'd say "Jim and Bob Smith" it should be "Minimal and Generic Host" not "Minimal and Generic hosts"


wadepickett (approved):
Looks good!  Thanks!

PR #25966, opened 2022-05-25
Added Missing using statement

codestellar:
Added using Azure.Extensions.AspNetCore.Configuration.Secrets;


wadepickett (commented):
@codestellar , the sample you are editing is the 6.x version.  .NET 6.x takes advantage of implicit gobal using directives.  In this case, adding `using Azure.Extensions.AspNetCore.Configuration.Secrets;` is unnecessary.  If you are targeting .NET 6.0 and look at the code in VS 2022 you will notice that using statement would get greyed out after entering it, and VS would indicate "Using directive is unnecessary". 

For reference:  [Implict using directives](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/overview#implicit-using-directives)

Closing this PR.  Feel free to re-open if you feel I was missing something.  Thanks for the help

wadepickett:
Closing PR, see review comment.  Feel free to re-open if you feel I missed something.

PR #25979, opened 2022-05-27
Update intro.md

Rick-Anderson:
Is this the right way to handle plural of a type?


guardrex (approved):
Yes, but "Tag Helper" is a proper noun ........ *or is it??!* 😄

I'm not sure about "kind" tho. Would it be ok as something like ...

> ... to configure specific Tag Helper instances. Consider the ...

... or use "type" ...

> ... to configure Tag Helper instances of a specific type. Consider the ...

PR #25993, opened 2022-05-28
Correction of a possible typo

dhairya-cse:
The filtering section example logs Information instead of warning.


guardrex (approved):
Thanks @dhairya-cse! :rocket:

PR #25996, opened 2022-05-28
Fixed invalid attribute name.

ridercz:


guardrex (approved):
Thanks @ridercz! :rocket:

PR #26002, opened 2022-05-30
Add information to add access token usage in blazor webassembly

kbeaugrand:
Fix issue when starting with the Blazor template. 

Adding the access token request was confusing with the default documentation and causes additional discussions in an issue : [https://github.com/dotnet/aspnetcore/issues/39311](https://github.com/dotnet/aspnetcore/issues/39311).

This PR aims to document what is really needed to change in the app to attach access tokens.


aspnetcore/blazor/security/webassembly/hosted-with-azure-active-directory-b2c.md
```diff
@@ -317,6 +317,7 @@ All scopes must belong to the same app per Azure Active Directory rules. Additio
 builder.Services.AddMsalAuthentication(options =>
 {
     ...
+    options.ProviderOptions.ResponseType = "id_token token";
     options.ProviderOptions.DefaultAccessTokenScopes.Add("{SCOPE URI}");
```
damienbod, line 321: The implicit flow is no longer recommended and I don't think this should be used in the docs. Using code flow with PKCE or a confidential client would be better?

Greetings Damien

kbeaugrand: In that case, should I document to use code flow instead of ``id_token token`` ?

kbeaugrand: Here is a new version of the PR

guardrex:
Hello @kbeaugrand ... The default auth flow is PKCE/code now (5.0 or later), so this change isn't necessary. Also, we're not seeking to discuss or cover all of the edge cases here (e.g., obtaining an id token for a hybrid flow) or get into much *security theory* 😄. MS tutorials are more about getting the reader up and going with a base case, then the tutorial refers to the reference documentation for additional scenarios and detail on the framework. We're just trying to get devs up and running here.

This tutorial was placed in this doc set merely because there were no other Blazor docs for Identity/Azure at the time that Blazor came out several years ago. Since then, the Identity and Azure docs have come a long way, including with Blazor tutorials. One of the things that we'll be looking at shortly is **_dropping this topic_** 🔪 from this doc set because MS now maintaining duplicate Blazor coverage for base case scenarios for MSAL/Azure. IIRC, one or a few of them even have the Blazor app access Graph API.  We might be dropping our Graph API coverage here, too, and pointing to their coverage on Graph API. In any case, we leave the details and advanced scenarios to the MS Identity Platform and Azure documentation, which are cross-linked. Those cross-links go right to the code flow doc; and IIRC, there's a cross-link for implicit flow scenarios. For Okta scenarios (or any other IdP) not covered by MS, Okta should cover them, not MS. We can't keep up with every IdP framework.

Quick note for future reference: Thanks for asking about this and commenting on the topic, but it's best to open an issue first to discuss doc updates for anything other than a patch update.

Thanks @damienbod 🎸 for your quick assistance. You're correct. We agree with your assessment.

guardrex:
OH! ... and one other thing, @kbeaugrand ... I do try and look at all of the issues that have come in for a general subject like this when a given doc (or doc node ... a folder of articles) comes up for one of my massive overhauls, which take place every few years for any given article or node of articles.

I **_DID_** make a note of this subject on my tracking issue for that effort. You can see if you scroll down the issue to the *UE pass tracking* part that this whole security node will be assessed ... what to keep? ... what to drop? ... does everything still work? 🙈😆 ... and taking a look at subjects like this, where I closed the issue as a **won't fix** but would like to review it because I might change my mind or I might find something to cross-link for it that would be handy for devs.

We're working on Hybrid docs at the moment, but I'm angling to get to this node before all of the work lands for .NET 7 in a few months. Otherwise, the work on the security node will be pushed back to 23Q1 or Q2. Anyway ... I will review this subject and your PU repo issues and discussion with Artak and Javier when I reach this node of topics.

**Blazor project tracking 2022**
https://github.com/dotnet/AspNetCore.Docs/issues/24615

kbeaugrand:
Many thanks for this clarification @guardrex !

PR #26015, opened 2022-06-01
Blazor default ASP.NET Core Identity UI updates

### LINKED ISSUE #26014: Move section to overview
Per offline discussion, move the *Blazor Server uses ASP.NET Core Identity* section to the overview with updates.

---
#### Document Details

* ID: b6032832-e645-24a1-583e-7b1898019fd1
* Version Independent ID: 8da62490-e36f-8f38-8f06-d5d6bceaafde
* Content: [Secure ASP.NET Core Blazor Server apps](https://docs.microsoft.com/en-us/aspnet/core/blazor/security/server/?view=aspnetcore-6.0&tabs=visual-studio#blazor-server-uses-aspnet-core-identity)
* Content Source: [aspnetcore/blazor/security/server/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/security/server/index.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #26014

[Internal Review Topic (links to section)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/security/index?view=aspnetcore-3.1&branch=pr-en-us-26015#blazor-server-authentication)

The PR ...

* Moves the section to the *Security and Identity* overview.
* Because the overview is meant to primarily address *common* guidance, I felt it best to integrate the content into the topic's existing section on *Blazor Server authentication*.
* Seeks to extend and clarify prior remarks in the section. This is purely *first-cup* ☕ language ideas to get us started. I expect it to change, perhaps quite a bit.👂

I defer on pinging any of the others from the discussion.

### DIFF
--- a/aspnetcore/blazor/security/index.md
+++ b/aspnetcore/blazor/security/index.md
@@ -25,10 +25,7 @@ Blazor WebAssembly apps run on the client. Authorization is *only* used to deter
 [Razor Pages authorization conventions](xref:security/authorization/razor-pages-authorization) don't apply to routable Razor components. If a non-routable Razor component is [embedded in a page](xref:blazor/components/prerendering-and-integration), the page's authorization conventions indirectly affect the Razor component along with the rest of the page's content.
 
 > [!NOTE]
-> <xref:Microsoft.AspNetCore.Identity.SignInManager%601> and <xref:Microsoft.AspNetCore.Identity.UserManager%601> aren't supported in Razor components. Blazor Server apps use ASP.NET Core Identity. For more information, see the following guidance:
-> 
-> * <xref:blazor/security/server/index>
-> * [Scaffold ASP.NET Core Identity into a Blazor Server app](xref:security/authentication/scaffold-identity#scaffold-identity-into-a-blazor-server-project)
+> ASP.NET Core apps that use ASP.NET Core Identity for user management should use Razor Pages instead of Razor components for identity-related UI, such as user registration, login, logout, and other user management tasks. ASP.NET Core Identity is designed to work in the context of an HTTP request and response, which isn't generally available in Blazor apps. ASP.NET Core abstractions, such as <xref:Microsoft.AspNetCore.Identity.SignInManager%601> and <xref:Microsoft.AspNetCore.Identity.UserManager%601>, aren't supported in Razor components. For more information on using ASP.NET Core Identity with Blazor, see [Scaffold ASP.NET Core Identity into a Blazor Server app](xref:security/authentication/scaffold-identity#scaffold-identity-into-a-blazor-server-project).
 
 ## Authentication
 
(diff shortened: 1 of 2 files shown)


danroth27 (changes requested):
I think the goal here is to explain to Blazor users why they have to deal with Razor Pages for the default ASP.NET Core Identity UI when the rest of their app is built with Blazor. This isn't a Blazor Server specific concern. It applies to Blazor WebAssembly apps as well.

Instead of adding the proposed text to the Blazor Server section of the Blazor security overview topic, how about we expand on the existing note at the top of the doc that explains why you can't use `SignInManager<TUser>` and `UserManager<TUser>` from Blazor, since the topics are related. Something like this:

> [!NOTE]
> ASP.NET Core apps that use ASP.NET Core Identity for user management should use Razor Pages instead of Razor components for identity related UI, like user registration, login, logout, etc. This is because ASP.NET Core Identity is designed to work in the context of an HTTP request and response, which isn't generally available in Blazor apps. ASP.NET Core abstractions, like <xref:Microsoft.AspNetCore.Identity.SignInManager%601> and <xref:Microsoft.AspNetCore.Identity.UserManager%601>, aren't supported in Razor components. For more information on using ASP.NET Core Identity with Blazor, see [Scaffold ASP.NET Core Identity into a Blazor Server app](xref:security/authentication/scaffold-identity#scaffold-identity-into-a-blazor-server-project).

guardrex:
Ok, but what about keeping the first piece and placing it over in the Blazor Server overview? .... because that seemed to be what the thrust of the offline discussion was addressing. Just this bit ...

> When ASP.NET Core stores and manages user accounts, Blazor Server uses \[ASP.NET Core Identity](xref:security/authentication/identity). When user accounts are stored and managed externally, Blazor Server uses:
>
> \* \[Microsoft identity platform](/azure/active-directory/develop/) for Azure Active Directory (AAD) or AAD B2C, which use the [OAuth](https://oauth.net) and \[OpenID Connect (OIDC)](https://openid.net/connect/) protocols.
> \* Typically, \[OAuth](https://oauth.net)/\[OpenID Connect (OIDC)](https://openid.net/connect/) directly for non-Microsoft identity providers, often via a third-party API/package that simplifies implementation.

danroth27:
I don't think we need that extra text. I think it was only a point of confusion because the previous text sounded like Blazor Server only supported ASP.NET Core Identity. We're removing the problematic text, so I think we should be good to go.

guardrex:
Ok ... it's almost ready. *Commit on the way!* 🏃

guardrex:
Ok ... I think that *lick'd IT* 🐄👅.

danroth27 (approved):
:shipit:

PR #26018, opened 2022-06-01
Fix article description showing twice

TFellowes:
The article description shows twice when the selected moniker is < .net 6.
This moves all instances of the article description within moniker range selections, so there is only ever one description displayed.

Example incorrect behaviour:
image

Also fix a typo in the article description to read "create navigation links in Blazor apps" rather than "create **a** navigation links in Blazor apps".


guardrex (approved):
Thanks @TFellowes! :rocket:

PR #26036, opened 2022-06-02
Institute global no-loc

### LINKED ISSUE #25187: metadata in docfx.json seems a solid solution,
See details in #25066
Internal issue [here](https://ceapex.visualstudio.com/Engineering/_workitems/edit/571334)
previous [internal issue]( https://portal.microsofticm.com/imp/v3/incidents/details/290559630/home)

Per @guardrex 

1. Create a fake doc that isn't in the ToC (`testdoc.md`) and doesn't have a `no-loc` metadata entry that just has a single paragraph sprinkled with a few common words ... "computer" ... "cat" ... whatever.
1. Add `fileMetadata` for it in `docfx.json` ...

   ```json
   "fileMetadata": {
     "no-loc": { "testdoc.md": "[ computer, cat ]" },
     ...
   }
   ```
1. Given the MT runs on a post-publish schedule, go ahead and take it live.
1. Check back in a few weeks ... did the "computer" and the "cat" words escape MT?

Even if that works tho, HT might not catch it because translators might not be looking at the `docfx.json` file for the string of words to avoid. The whole translation system is opaque to the community AFAIK ... it's ✨😄.

guardrex:
Fixes #25187

# 🎉 _HERE it is!_ 🍾

Testing looks good for using `fileMetadata` with the live test entry that I set up earlier. The test topic is at ...

https://docs.microsoft.com/es-es/aspnet/core/migration/no-loc-test?view=aspnetcore-6.0

... for ...

```json
"no-loc": {
  "**/**.md": [ "toaster", "bowl", "toothbrush", "magnet", "oak" ]
},
```

... in `docfx.json`. This PR sets up ...

```json
"no-loc": {
  "**/**.md": [ ".NET MAUI", "Mac Catalyst", "Blazor Hybrid", "Home" , "Privacy" , "Kestrel", "appsettings.json", "ASP.NET Core Identity", "cookie", "Cookie", "Blazor", "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", "Razor", "SignalR" ]
},
```

... and BTW ... I think I'll alphabetize that. Not tonight tho. I'm 🧠🔥again ... it's late here and time for 🛌💤. I'll alphabetize the list in the morning. 

This is a good time to adjust the list. I'm not sure why `appsettings.json` is here. Can't that be a find-'n-replace with a code-fenced filename? Anyway ... this is a good time to make adjustments.👂 

There were no custom `no-loc` entries in the doc set AFAICT. It looks like my find-'n-replace 🔪 every entry. **_Adios `no-loc`s!_** 😄

There are ~600 file updates here, so we either need to just scan-👁️ it ... or break out the **_EYE DROPS_**.

If there are any concerns at all about this, then let's just close it. If there are no concerns, *GrEaT!* 🎉 This will save real ⏱️ and 💰, so I hope we can proceed. However, I don't want to create a giant 💥 with this.


aspnetcore/docfx.json
```diff
@@ -66,7 +66,7 @@
         "**/tutorials/**/**.md": "tutorial"
       },
       "no-loc": {
-        "**/**.md": [ "toaster", "bowl", "toothbrush", "magnet", "oak" ]
+        "**/**.md": [ ".NET MAUI", "Mac Catalyst", "Blazor Hybrid", "Home" , "Privacy" , "Kestrel", "appsettings.json", "ASP.NET Core Identity", "cookie", "Cookie", "Blazor", "Blazor Server", "Blazor WebAssembly", "Identity", "Let's Encrypt", "Razor", "SignalR" ]
```
Rick-Anderson, line 69: Here's the magic.
image

Rick-Anderson (approved):

Rick-Anderson:
cc @serpent5

guardrex:
**_⬆️📅_**:

* I was able to find some custom `no-loc`s after all to get *OUT* 🔪. My search phrase last night must have been hosed 🙈.
* In most cases, I was able to move custom entries to the global list in `docfx.json`. In some cases, I was able to code-fence the entries in the topic to resolve it.
* I didn't need to do the following for the updates, but individual topics don't need the full custom `no-loc` list with custom entries for just a few words/phrases. I think it best in those cases to use the **_inline syntax_**: `:::no-loc text="{TEXT}":::`. That approach eliminates the need to ever use a topic-based `no-loc` metadata entry.
* Some topics merely didn't have the entry code fenced in headings. It's fine to code fence in headings, so that's how I resolved those.
* I also alphabetized the global list.
* I code-fenced a few "appsettings.json" entries around the repo in order to avoid having that in the global list. I also eyeballed the dev, staging, and production versions of the file. *We're good!* :+1:

wadepickett:
I will have to leave the review up to Rick, who knows the history on this and what we are testing here.  

Just a side note:  I think we will soon need to add ".NET Multi-platform App UI" as a no-loc product name if that is what we are to officially use for .NET Maui going forward.  It is already being used, however, it seems a bit long and clunky to me so I wonder if it will change later.

serpent5:
Great stuff. It's awesome to see this finally made it. ⭐

PR #26058, opened 2022-06-05
Fix redundant equal sign

LuohuaRain:


guardrex (approved):
Thanks @LuohuaRain! :rocket:

PR #26063, opened 2022-06-06
WASM culture guidance

### LINKED ISSUE #25973: "*The app's culture is set by matching the first requested language that matches a supported culture of the app.*", but how to specify supported cultures?
Section [§ Dynamically set the culture from the Accept-Language header](https://docs.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-6.0&pivots=webassembly) for Blazor WASM has this to say:

> The app's culture is set by matching the first requested language that matches a supported culture of the app.

However, it then fails to give instructions how to specify supported cultures.

Note that the same section for Blazor Server does give instructions how to specify supported cultures:

> Specify the app's supported cultures in Program.cs immediately after Routing Middleware is added to the processing pipeline. The following example configures supported cultures for United States English and Chilean Spanish:
> 
> ```cs
> app.UseRequestLocalization(new RequestLocalizationOptions()
>     .AddSupportedCultures(new[] { "en-US", "es-CL" })
>     .AddSupportedUICultures(new[] { "en-US", "es-CL" }));
> ```

This gives me the impression that analogous information was intended to be put in the WASM version of this section, but was omitted by accident.

If my impression is incorrect and the current state of this section is intentional, then please close my issue.

Alternatively, if it is not possible to specify supported cultures, then I feel this line should be reworded to something like this:

> The app's culture is set by matching the first request language that matches a supported culture of the framework. Unlike in Blazor Server, in Blazor WASM it is not possible to narrow down the supported cultures on a per-application basis.

Side note: What happens if none of the languages listed in the Accept-Language header match cultures supported by the app? Does the app, I don't know, default to CultureInfo.InvariantCulture? It might be a good idea to document this perhaps?

---
#### Document Details

* Content: [ASP.NET Core Blazor globalization and localization](https://docs.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-6.0)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #25973

Thanks @gaazkam! :rocket:

Well, it took a while for me to test thru the scenario that I was thinking of ... based lightly on some of your, @hishamco, Oqtane framework pieces. However, that is a NOOP without setting the default loc cookie (*somehow*) in dev code. For WASM with Loc API services added (`builder.Services.AddLocalization();`), there is no OOB default loc cookie (`.AspNetCore.Culture`). Otherwise, the following might have been a demo of a *for-further-dev-customization* approach ...

> To set the culture using custom logic based on the localization cookie:
>
> * Use [JS interop](xref:blazor/js-interop/call-javascript-from-dotnet) to obtain the localization cookie's value. The default localization cookie name is provided by <xref:Microsoft.AspNetCore.Localization.CookieRequestCultureProvider.DefaultCookieName?displayProperty=nameWithType>.
> * Parse the cookie value to obtain its first value, which is the user's default requested culture.
> * Attempt to match the cookie's culture to the app's supported cultures list. If the cookie isn't present or the cookie's first culture doesn't match a supported culture, set the app's culture to the app's default culture.
> 
> The following example is the basis for further customization.
> 
> In the `<head>` content of `wwwroot/index.html` immediately after the script reference for the Blazor script (`blazor.webassembly.js`):
> 
> ```html
> <script>
>   window.getCookie = (name) => {
>     name = name + "=";
>     var decodedCookie = decodeURIComponent(document.cookie);
>     var ca = decodedCookie.split(';');
>     for (var i = 0; i < ca.length; i++) {
>         var c = ca[i];
>         while (c.charAt(0) === ' ') {
>             c = c.substring(1);
>         }
>         if (c.indexOf(name) === 0) {
>             return c.substring(name.length, c.length);
>         }
>     }
>     return "";
>   };
> </script>
> ```
> 
> In `Program.Main` (`Program.cs`):
> 
> ```csharp
> using Microsoft.AspNetCore.Localization;
> using Microsoft.JSInterop;
> 
> ...
> 
> var host = builder.Build();
> 
> var supportedCultures = new[] { "en-US", "es-CL" };
> var js = host.Services.GetRequiredService<IJSRuntime>();
> var localizationCookie = await js.InvokeAsync<string>("getCookie", 
>     CookieRequestCultureProvider.DefaultCookieName);
> var culture = CookieRequestCultureProvider.ParseCookieValue(localizationCookie)?
>     .UICultures?[0].Value;
> 
> if (culture == null || !supportedCultures.Any(c => c.Equals(culture, 
>     StringComparison.OrdinalIgnoreCase)))
> {
>     culture = supportedCultures[0];
> }
> 
> var cultureInfo = CultureInfo.GetCultureInfo(culture);
> CultureInfo.DefaultThreadCurrentCulture = cultureInfo;
> CultureInfo.DefaultThreadCurrentUICulture = cultureInfo;
> 
> await host.RunAsync();
> ```

Anyway ... I merely post that here for any potential further discussion. **_It's a NOOP without addl work_**, and we don't want to present something as complex as whatever Oqtane does ✨, including it's `ILocalizationService` and other bits and pieces.

In a NOTE in the *Dynamically set the culture from the `Accept-Language` header* section, I refer readers to the supported example approach for dynamically setting the culture by user preference. That's what the PU sent over, so I bet that's all that they'd like to show.

Let's **_NOT_** put my 🙈 **_RexHAQS_**&trade; 🙈 into the doc for JS interop for the `Accept-Encoding` header value (`js.InvokeAsync<string[]>("eval", "navigator.languages")`). If anyone opens an issue asking for tips or coverage for their own edge-case custom code on that, I'll just refer them to my remarks at https://github.com/dotnet/AspNetCore.Docs/issues/25973#issuecomment-1138869776.

@hishamco ... Unless you have anything else that you'd like to address, I think we're good with the little bit that I placed for this.

@gaazkam ... Use my **hacks** as an 'at your own risk' 😆. That's unsupported and not really recommended. We won't mention or cover that in the doc. It might work tho in cases where the browser will report `navigator.languages` and one codes it carefully with a supported fallback culture when `navigator.languages` isn't supported.


guardrex:
I'm going to go ahead with this because it's just a cross-reference update recommending what the PU already approved for the topic and that's already here for the use case where the dev want to limit the scope of cultures and/or provide user choice. I don't think we want to get into the weeds on anything as complex as the Oqtane framework's bits, which if devs want to see in whole or in part they can find via Awesome Blazor at https://github.com/AdrienTorris/awesome-blazor#others. As for my *hacks* with either possibly reading a custom implementation of a `.AspNetCore.Culture` cookie or for my other bit on `navigator.languages`, I think devs can explore and implement those advanced approaches on their own.

PR #26126, opened 2022-06-12
Update first-web-api

pea-sys:
Updated aspnet core-6.0 checklist
Postman →　http-repl


guardrex (approved):
Thanks @pea-sys ... and I'll update the cross-link to use XREF syntax while we're here.

PR #26171, opened 2022-06-18
Update protocols.md

JamesNK:
Update version numbers to match those at https://docs.microsoft.com/en-us/aspnet/core/grpc/supported-platforms?view=aspnetcore-6.0#supported-aspnet-core-servers

### DIFF
--- a/aspnetcore/host-and-deploy/iis/protocols.md
+++ b/aspnetcore/host-and-deploy/iis/protocols.md
@@ -34,7 +34,7 @@ Requirements to run gRPC on IIS:
 
 * In-process hosting.
 * Windows 11 Build 22000 or Windows Server 2022 Build 20348 or later.
-* TLS 1.2 or later connection
+* TLS 1.2 or later connection.
 
 ### Trailers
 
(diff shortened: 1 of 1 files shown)


aspnetcore/host-and-deploy/iis/protocols.md
```diff
@@ -33,8 +33,8 @@ Additional HTTP/2 features in IIS support gRPC, including support for response t
 Requirements to run gRPC on IIS:
 
 * In-process hosting.
-* Windows 10, OS Build 20300.1000 or later. May require use of a [Windows Insider](https://insider.windows.com) build.
-* TLS 1.2 or later connection
+* Windows 11 Build 22000 or Windows Server 2022 Build 20348 or later.
```
guardrex, line 36: It's not clear if the "or later" applies to both. Can the following work? ...

```suggestion
* Windows 11 Build 22000 or later, Windows Server 2022 Build 20348 or later.
```

guardrex (changes requested):
The double-conjunction is a bit awkward, but it's not very unclear.

guardrex (approved):

PR #26176, opened 2022-06-20
Update _Layout.cshtml to correct missing files

leighghunt:
Layout for sample is referring to minimized JavaScript and CSS files which aren't present in the project.

Apologies if I've missed a step and they should be created, but I removed the .min part of the filename and everything then looked right.


Rick-Anderson (approved):

Rick-Anderson:
Thanks for the contribution. Be sure to check out your name as a contributor about the 3rd of next month on the [What's new page](https://docs.microsoft.com/en-us/aspnet/core/whats-new/) in the **Community contributors** section at the bottom.

PR #26216, opened 2022-06-23
Remove .min ref from all non-Blazor projects /2

### LINKED ISSUE #26180: remove all ".min" from layout files
See #26176

Rick-Anderson:
Fixes #26180

image


guardrex (approved):
I can only sign off on the Blazor aspects. There are changes here that are of concern, but idk if they're real problems. For example, `bundleconfig.json` changes are here, which doesn't seem right. I also have a concern about the `libman.json` updates. Additionally, the SignalR changes don't seem right ... I think those are supposed to be minified assets for those commands/apps/examples. Anyway, I'm not 100% sure on those aspects, but the Blazor pieces not being here are good because those apps will surely 💥 with these updates.

PR #26221, opened 2022-06-24
I think the comment should be Certificate

sammychinedu2ky:
The initial definition is Certificate, so I guess it would be proper to have the **Certificate** comment instead of the **Managed** comment


sammychinedu2ky:
@serpent5

serpent5:
Hi @sammychinedu2ky. The comment there is to show that the other option for the `#define` is `Managed`, so it's correct as it is. That's a convention used in the docs samples to show the available options for a sample. In this case, there's a `Certificate` symbol and a `Managed` symbol. The `Managed` symbol is checked on line 37.

sammychinedu2ky:
Oh Ok, thanks for the clarification. I didn't know about that 😅

PR #26223, opened 2022-06-24
docs: Fix Kubernetes example

chertby:
Everywhere in the document, the _/health**z**_ endpoint is used (which, as I understand it, is the name by convention), and in the example for Kubernetes, just _/health_. I made changes to unify endpoints and get rid of misleadings.


guardrex:
Thanks @chertby ... Your versioning wasn't correct, so I fixed that up. That part of the doc was already 6.0 or later (and we can't nest the versioning). If this update is only for 6.0 or later, then only the code for that example requires an update in that part of the doc. If the <6.0 examples for that code should stay the same then further down the doc the examples remain as they are.

guardrex (approved):

PR #26236, opened 2022-06-26
Global no-store/no-cache guidance

### LINKED ISSUE #26235: Disabling Caching
Would appreciate a paragraph devoted to explaining how best to disable caching throughout a Core MVC Web application.

---
#### Document Details

* ID: b6728e59-93ec-891e-0416-94819296ad72
* Version Independent ID: 5d581acc-c651-2112-844a-6b85dc5a99a6
* Content: [Response caching in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/response?view=aspnetcore-6.0)
* Content Source: [aspnetcore/performance/caching/response.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/performance/caching/response.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

guardrex:
Fixes #26235

Thanks @xenophilios! :rocket:

Rick, a few points on this one ...

* If I've gone off the rails here, close this and take it over for a *better* PR than what I've set up here 🙈😄.
* The `Duration = 0` setting I think 🤔 is just a no-op in the scenario and can be removed/avoided.
* The RP approach is messy compared to the MVC approach ... *apparently*. It looks like something might be worked for 7.0 on [Make ResponseCacheMiddleware endpoint aware (dotnet/aspnetcore #37995)](https://github.com/dotnet/aspnetcore/issues/37995), so it might be good to have a follow-up 7.0 docs issue to make sure that this bit gets a 7.0 update if they make a change for MVC/RP/both.
* Although the "Razor Pages page responses" lingo isn't what I normally use, I think we must clarify that it's the page responses only from the RP app that get the attribute IIRC. Therefore, we should distinguish the app type ("RP") and the response type ("pages"), which also parallels the MVC language (i.e., "MVC controller").
* Head-check me on the `services.AddMvc().AddMvcOptions` piece for <6.0. I think that's right. *I can't compile that here on the new PC.* 😝 I haven't had the need (*yet*) to put 3.1/5.0 frameworks on here.
* I cross-link to https://github.com/dotnet/aspnetcore/issues/18890#issuecomment-584290537 for the exact PU example that Pranav provided at the time, but that's 3.1-era stuff. If there's any delta per the [Filter methods for Razor Pages](https://docs.microsoft.com/aspnet/core/razor-pages/filter) article, then I recommend that I make a change here to just link the Filters article instead of the PU issue comment. If we do that tho, we will have the reader work out the exact code for the global attribute approach on their own ... but that might be ok and is certainly preferable if the code shown by Pranav doesn't apply past 3.1.

  **_UPDATE_**: I just checked Pranav's code for 6.0 with Minimal APIs, and it's :+1: as he wrote it with an updated Minimal APIs registration in `Program.cs`. I think the cross-link with one more sentence will work there, so I'll add that sentence on the next commit.

  **_UPDATE_**: **Done!**

### DIFF
--- a/aspnetcore/performance/caching/response.md
+++ b/aspnetcore/performance/caching/response.md
@@ -97,7 +97,7 @@ The preceding code requires adding the Response Caching Middleware services <xre
 
 [!code-csharp[](response/samples/6.x/WebRC/Program.cs?name=snippet&highlight=4,13)]
 
-### NoStore and Location.None
+### `NoStore` and `Location.None`
 
 <xref:Microsoft.AspNetCore.Mvc.CacheProfile.NoStore> overrides most of the other properties. When this property is set to `true`, the `Cache-Control` header is set to `no-store`. If <xref:Microsoft.AspNetCore.Mvc.CacheProfile.Location> is set to `None`:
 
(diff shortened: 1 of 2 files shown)


Rick-Anderson (approved):

xenophilios:
Thank you very much for the fast and enthusiastic response, especially in the early hours of a Sunday morning!

xenophilios:
Would appreciate some comment on the following consideration:
If caching is turned off after some caching has taken place, will the cached files continue to be used?

PR #26255, opened 2022-06-29
Fix wrong Polly policy name

GitHubPang:
The policy name in the text was wrong.
It should be `Regular` and `Long`,
not `Regular` and `Short`.
Doc: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-6.0#add-policies-from-the-polly-registry


guardrex (approved):
Thanks @GitHubPang! 🎷

PR #26274, opened 2022-06-30
Update signalr-blazor.md

MoziVi:
Updating namespace 'using BlazorServerSignalRApp.Hubs;' to correct 'using BlazorServerSignalRApp.Server.Hubs;' to fix an error when building a project.


guardrex (approved):
Thanks for the catch, @MoziVi! Yes, I must have deleted it for the 6.x version of the topic because it is correct for the 5.x and 3.x versions. I guess that my 👁️ caught "Server" in the namespace name for the Blazor Server app without realizing that left the namespace alone across hosting model versions of the topic ... so I deleted that segment by mistake. Thanks again!

PR #26310, opened 2022-07-04
Update grpc-start.md

pea-sys:
Fix typo


guardrex:
Thanks @pea-sys! :rocket:

James, if we drop the word "service" here, we'll need to drop it for 5.x and 3.x content further down. I'll take care of it after I hear back from you.

JamesNK (approved):

PR #26333, opened 2022-07-06
Removed spaces to make command one line

yhscherber:
The existing script has a line break which prevents the script from working if copying directly. My change removed the spaces so the script is usable without editing.


guardrex:
Hello @yhscherber ... Thanks for the PR. It was done originally because we have a line length limit on code of 85 characters. *Indeed, tho!* We can't be breaking folks. 🙈😄 I'm going to move this to an issue and work it right now because this update is going to touch several topics here.

guardrex:
Ah! Good news, @yhscherber ... I only found **_ONE_** other instance of this. I was concerned that there might be a half-dozen or more, but no ... just one. Let's merge your PR here, and then I'll knock out that other one on a new PR.

Thanks again! :rocket:

guardrex (approved):

Issue #13582, opened 2019-07-31
Project Naming Guidelines


PradeepSinghMakwana:
I am not sure, if there is a documentation already available from Microsoft for this. Watch the below video to understand the issue better.

[Video](https://youtu.be/KM5o9M4cuQA)

PradeepSinghMakwana:
I mean we can have, project structure and naming guidelines. So that it is easier to create and scale projects. I recently faced problems in naming and structuring my own projects.

Rick-Anderson:
@PradeepSinghMakwana can you open that issue in the .NET repo?

PradeepSinghMakwana:
> @PradeepSinghMakwana can you open that issue in the .NET repo?

@Rick-Anderson  Do you mean microsoft/dotnet or aspnet/AspNet. I have opened the issue in both and then closed it in aspnet/AspNet as I think you didn't really mean that.

Rick-Anderson:
https://github.com/dotnet/docs

Rick-Anderson:
You'll have to write up exactly what's needed and not cite a video.

PR #26420, opened 2022-07-13
Redirect users when content isn't found

### LINKED ISSUE #26415: Add to documentation how to Redirect inside the <NotFound> element
When a route is not found, the following documentation is supplied for App.razor:

```
<NotFound>
        <h1>Sorry</h1>
        <p>Sorry, there's nothing at this address.</p> b
</NotFound>

```

As great as that may be, another very common scenario is to redirect the user to another page when landing at a page that is not a match. Please could you provide an example in the documentation of how to Redirect inside the <NotFound> element as I think that will be super useful and it is not really clear to me how I would do that. 

```
<NotFound>
   @{ NavigationManager.NavigateTo("/Home", false); }
</NotFound>

```

This is something I tried. This does have the desired effect, however can't be correct as it throws a NavigationException initially while the app is loading up when debugging. So what is the the correct way? Thank you.

---
#### Document Details

* Content: [ASP.NET Core Blazor routing and navigation](https://docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-6.0)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #26415

[Internal Review Topic (links to section)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-3.1&branch=pr-en-us-26420#provide-custom-content-when-content-isnt-found) ... scroll down to the bottom of the *Provide custom content when content isn't found* section that loads via this link.

### RE: Redirect users when content isn't found

Thanks @dirkesquire! :rocket:

Steve ... a couple of points (one from me, one from @dirkesquire) ...

* Your [PU discussion (dotnet/aspnetcore #13582)](https://github.com/dotnet/aspnetcore/issues/13582#issuecomment-527383363) left me with the impression that using a custom redirect component and calling for navigation in its `OnAfterRender` would be a sane approach to avoiding the `NavigationException` in Blazor Server apps during debugging, noting that ...
  * Seems like it would be fine for production.
  * Doesn't check `firstRender` because there isn't a double event call, seemingly because it navigates on the prerender ... a short-circuit.

* From Dirk ......

  > Drawing from inspiration from the React world, specifically React-Router, you have not only a NavigationManager you can invoke functions on but there is also a built in element `<Navigate>` element that I think would make sense here which I was kinda expecting you might mention (if it existed - but I take it it doesn't). This would avoid the user having to 'roll their own' `<RedirectToUser>` element.
  >
  > This is the type of solution I would expect:
  >
  > ```razor
  > <NotFound>
  >     <Navigate to="/" />
  > </NotFound>
  > ```
  >
  > I am of course happy building my own Navigate page as it is quite easy, but I think it would be a beneficial addition to the framework.

  So, I told Dirk on the issue that if you're interested that he should open it on the PU repo as a request but that you might say 'nah' to that here and save him the trouble.

### DIFF
--- a/aspnetcore/blazor/fundamentals/routing.md
+++ b/aspnetcore/blazor/fundamentals/routing.md
@@ -76,6 +76,47 @@ In the `App` component, set custom content in the <xref:Microsoft.AspNetCore.Com
 
 Arbitrary items are supported as content of the `<NotFound>` tags, such as other interactive components. To apply a default layout to <xref:Microsoft.AspNetCore.Components.Routing.Router.NotFound> content, see <xref:blazor/components/layouts#apply-a-layout-to-arbitrary-content-layoutview-component>.
 
+To redirect users when content isn't found, Blazor WebAssembly apps can:
+
+* Inject a [Navigation Manager](#uri-and-navigation-state-helpers) into the `App` component:
+
+  ```razor
+  @inject NavigationManager NavigationManager
+  ```
+
+* Call <xref:Microsoft.AspNetCore.Components.NavigationManager.NavigateTo%2A> with the destination. The following example navigates to the default component, typically the `Index` component (`Pages/Index.razor`):
+
+  ```razor
+  <NotFound>
+      @{ NavigationManager.NavigateTo("/", false); }
+  </NotFound>
+  ```
+  
+To redirect users in a Blazor Server app, calling <xref:Microsoft.AspNetCore.Components.NavigationManager.NavigateTo%2A> directly in the `App` component results in a <xref:Microsoft.AspNetCore.Components.NavigationException> during debugging, which is thrown during prerendering. To avoid the exception, call for the navigation in [`OnAfterRender`](xref:blazor/components/lifecycle#after-component-render-onafterrenderasync) of a custom redirect component.
+
+The following Blazor Server example navigates users to the app's default component, typically the `Index` component:
+
+* Create a `RedirectUsers` component (`Shared/RedirectUsers.razor`):
+
+  ```razor
+  @inject NavigationManager NavigationManager
+
+  @code {
+      protected override void OnAfterRender(bool firstRender)
+      {
+          NavigationManager.NavigateTo("/", false);
+      }
+  }
+  ```
+
+* In the `App` component (`App.razor`), place the custom redirect component in `<NotFound>` child content:
+
+  ```razor
+  <NotFound>
+      <RedirectUsers />
+  </NotFound>
+  ```
+
 ## Route to components from multiple assemblies
 
 Use the <xref:Microsoft.AspNetCore.Components.Routing.Router.AdditionalAssemblies> parameter to specify additional assemblies for the <xref:Microsoft.AspNetCore.Components.Routing.Router> component to consider when searching for routable components. Additional assemblies are scanned in addition to the assembly specified to `AppAssembly`. In the following example, `Component1` is a routable component defined in a referenced [component class library](xref:blazor/components/class-libraries). The following <xref:Microsoft.AspNetCore.Components.Routing.Router.AdditionalAssemblies> example results in routing support for `Component1`.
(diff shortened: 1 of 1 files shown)


SteveSandersonMS:
I don't think the code shown here does exactly what it suggests. The only reason why it's not throwing during prerendering now is because the redirection is done from `OnAfterRenderAsync`, which is not executed during prerendering. Creating a child component doesn't make any difference either way (it's an unnecessary complication), as far as I can tell. Please correct me if I'm wrong!

People could simply do a redirection from `OnAfterRenderAsync` without needing this, but would have to bear in mind this is no longer doing the redirection during prerendering. Instead, **the entire page is starting up on the client, and *then* it's redirecting**.

My guess is that unless a lot of developers are continuing to ask about this, we should avoid making it seem like they have to do some extra work. They can much more easily remove the `NavigationException` from the list of exceptions that VS will break on.

If this does become a thing that a lot of developers ask about, we could look at enhancing this to communicate the termination of the render flow without using an exception. It would technically be a breaking change but probably a safe enough one.

dirkesquire:
Thanks Steve.
It seems that quite a few developers have been asking about this from a few other Issues I've seen.
So it seems that we are back to:

<NotFound>
   @{ NavigationManager.NavigateTo("/Home", false); }
</NotFound>

I'll admit this works very well, and a bit faster actually because it seems to redirect earlier than when using redirecting from inside a ChildComponent. So your comments do make sense.

The only questions remaining are:
1) Do we not want to make a comment about this in the [documentation](https://docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-6.0#provide-custom-content-when-content-isnt-found) (that is originally how this conversation all started - with me giving feedback about the documentation), and secondly,

2) How do we turn off catching of the NavigationException in VS Code? I did a google search but I'm still not sure of the answer.

At the moment I get a horrible popup everytime I debug, and it really does break the flow of development to have to click through it every time you restart the application, which is very often when you are making changes.

image

SteveSandersonMS:
```razor
<NotFound>
   @{ NavigationManager.NavigateTo("/Home", false); }
</NotFound>
```

Could you clarify why anybody would want to use this? It seems like a really bad user experience. Instead of telling the user that the content they're looking for can't be found, it simply takes them to the homepage with no explanation. In general I don't think people should commonly be doing this.

> Do we not want to make a comment about this in the [documentation](https://docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-6.0#provide-custom-content-when-content-isnt-found) (that is originally how this conversation all started - with me giving feedback about the documentation)

Unless I'm missing something important, I don't think we should have an example like this in the docs, as it would imply we think it's a sensible thing to do.

> At the moment I get a horrible popup everytime I debug

Could you clarify - why is it that you're always going to a "not found" page every time you start debugging? That seems unusual. 

In any case, I think the problem would not exist if you weren't trying to redirect during prerendering.

> How do we turn off catching of the NavigationException in VS Code

It's [easy enough to do in VS](https://docs.microsoft.com/en-us/visualstudio/debugger/managing-exceptions-with-the-debugger?view=vs-2022#add-and-delete-exceptions), but I don't know about VS Code. If you're unable to find any docs about how to do it in VS Code, it might be a worthwhile thing to submit as a feature request there.

dirkesquire:
Alright fair enough. I think it is just two ways of going about it. Is there really a point to show users a page with nothing on it other than telling people that nothing is there? It is just easier to take the user to the home page if they typed a URL in wrong. But this is just a minor point, and not the real issue. I think both approaches should be supported, but I'm willing to change to your suggestion.

Okay so to clarify, my situation is slightly more complicated but perhaps not TOO unusual.. I've got a big existing corporate product website which is completely built in MVC. So to get with the times I've added Blazer server side to the existing project so that I can start building any new pages in Blazer and slowly start upgrading existing MVC pages to Blazer one by one as I have time.

So the problem is that the home page is a MVC page, but the second I add Blazer server side to it, it kinda hijacks that home page. So immediately after logging on users are taken to / which shows the 'Not found there's nothing here!'. Clearly this is also a terrible user experience. Hence I wanted to redirect them to a MVC page.

UPDATE
Finally figured out that its _Host.cshtml that was hijacking the "/" route. So I managed to solve the problem by commenting out `@page "/"` inside of _Host.cshtml. I previously tried commenting out `endpoints.MapFallbackToPage("/_Host")` but that didn't prevent _Host.cshtml responding to '/' and they would still see the 'There's nothing here!' as the home page.

PR #26374, opened 2022-07-09
Added sample Code for an issue

sammychinedu2ky:
Added Sample Code for https://github.com/dotnet/AspNetCore.Docs/issues/26346 and #26370

***@Rick-Anderson EDIT:***
Fixes #26346 Fixes #26370

PR #26437, opened 2022-07-15
Minimal api stream sample 

### LINKED ISSUE #26346: .NET 7: Add sample for Minimal API Bind the request body as a Stream or PipeReader
See https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-net-7-preview-1/

Sample goes in 
https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples

```
var builder = WebApplication.CreateBuilder(args);
var app = builder.Build();

app.MapGet("/", () => "Hello World!");

app.MapPost("/fileUp", async (IFormFile file) =>
{
    string tempfile = Path.GetTempFileName();
    using var stream = File.OpenWrite(tempfile);
    await file.CopyToAsync(stream);
});

app.MapPost("/FilesUp", async (IFormFileCollection myFiles) =>
{
    foreach (var file in myFiles)
    {
        string tempfile = Path.GetTempFileName();
        using var stream = File.OpenWrite(tempfile);
        await file.CopyToAsync(stream);
    }
});

app.Run();
```

sammychinedu2ky:
Added sample code for an issue #26374

Fixes #26346


aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/BackgroundQueueService.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    }
+}
+
+class Person
+{
+    public string Name { get; set; } = String.Empty;
+    public int Age { get; set; }
+    public string Country { get; set; } = String.Empty;
+}
```
Rick-Anderson, line 47:
```suggestion
using System.Text.Json;
using System.Threading.Channels;

namespace BackGroundQueueService;

class BackGroundQueue : BackgroundService
{
    private Channel<Stream> _queue;
    private readonly ILogger<BackGroundQueue> _logger;


    public BackGroundQueue(Channel<Stream> queue, ILogger<BackGroundQueue> logger)
    {
        _queue = queue;
        _logger = logger;
    }
    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
    {

        await foreach (var dataStream in _queue.Reader.ReadAllAsync())
        {
            // Reset the stream to the beginning.
            dataStream.Position = 0;

            var reader = new StreamReader(dataStream);
            try
            {   
                var Person = JsonSerializer.Deserialize<Person>(await reader.ReadToEndAsync())!;
                // Do more work with the data.
                 _logger.LogInformation($"{Person.Name} is {Person.Age} years and from {Person.Country}");
            }
            catch (Exception ex)
            {
                _logger.LogError(ex.Message);
            }
        }

        return;
    }
}

class Person
{
    public string Name { get; set; } = String.Empty;
    public int Age { get; set; }
    public string Country { get; set; } = String.Empty;
}
```

Rick-Anderson, line 47: Git like a blank line at the end of each file

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
```diff
@@ -0,0 +1,48 @@
...
+   reusableStream.Position = 0;
+
+   // set the response body to the reusable stream
+    req.Body = new MemoryStream(reusableStream.ToArray());
+
+   return "registered";
+});
+
+
+app.Run();
```
Rick-Anderson, line 48:
```suggestion
using System.Threading.Channels;
using BackGroundQueueService;
var builder = WebApplication.CreateBuilder(args);
// Create a channel to send data to the background queue.
builder.Services.AddSingleton<Channel<Stream>>((_) => Channel.CreateUnbounded<Stream>());

// Create a background queue service.
builder.Services.AddHostedService<BackGroundQueue>();
var app = builder.Build();

app.Use(async (context, next) =>
{
    await next();
    var reader = new StreamReader(context.Request.Body);
    app.Logger.LogInformation("Reading the stream from the request body.");
});

app.MapGet("/", () => "Hello World!");

// curl --request POST 'http://localhost:5256/register' --header 'Content-Type: application/json' --data-raw '{ "Name":"Samson", "Age": 23, "Country":"Nigeria" }'
app.MapPost("/register", async (Stream body, HttpRequest req, Channel<Stream> queue) =>
{
    // Create a rewindable stream to be able to reuse the body stream.
    var reusableStream = new MemoryStream();

    // Copy the request body to the reusable stream.
    await body.CopyToAsync(reusableStream);

    // Reset the stream to the beginning.
    reusableStream.Position = 0;

    // Send the stream to the background queue.
    await queue.Writer.WriteAsync(reusableStream);

    // Reset the stream to the beginning.
    reusableStream.Position = 0;

    // Set the response body to the reusable stream.
    req.Body = new MemoryStream(reusableStream.ToArray());

    return "registered";
});


app.Run();

```

Rick-Anderson (changes requested):
Super minor changes

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
Rick-Anderson: Wait, is that the right sample? I had super minor suggestions

Rick-Anderson: Yes, that's the right program. Git is doing a bad job at the diff file. Just added logging and changed the comments a bit.

sammychinedu2ky: Yh I have added the lines. thanks

Rick-Anderson:
@sammychinedu2ky  I did suggestions so you just have to "Add suggestion to batch" and then "Commit"

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/BackgroundQueueService.cs
```diff
@@ -0,0 +1,47 @@
...
+namespace BackGroundQueueService;
+
+class BackGroundQueue : BackgroundService
+{
+    private Channel<Stream> _queue;
+
+    public BackGroundQueue(Channel<Stream> queue)
+    {
+        _queue = queue;
+    }
```
Rick-Anderson, line 13:
```suggestion
    private readonly ILogger<BackGroundQueue> _logger;

    public BackGroundQueue(Channel<Stream> queue, ILogger<BackGroundQueue> logger)
    {
        _queue = queue;
        _logger = logger;
    }
```

sammychinedu2ky:
> @sammychinedu2ky I did suggestions so you just have to "Add suggestion to batch" and then "Commit" Then just do a `git pull ` to bring in the changes locally.

I've added the changes

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/BackgroundQueueService.cs
```diff
@@ -0,0 +1,47 @@
...
+        await foreach (var dataStream in _queue.Reader.ReadAllAsync())
+        {
+            // reset the stream to the beginning
+            dataStream.Position = 0;
+
+            var reader = new StreamReader(dataStream);
+            try
+            {   
+                var Person = JsonSerializer.Deserialize<Person>(await reader.ReadToEndAsync())!;
+                  Console.WriteLine($"{Person.Name} is {Person.Age} years and from {Person.Country}");
```
Rick-Anderson, line 26:
```suggestion
                 _logger.LogInformation($"{Person.Name} is {Person.Age} years and from {Person.Country}");
```

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/BackgroundQueueService.cs
```diff
@@ -0,0 +1,47 @@
...
+                // you could do something else with the data
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine(ex.Message);
```
Rick-Anderson, line 31:
```suggestion
                _logger.LogError(ex.Message);
```

sammychinedu2ky:
ooh I see the logger you added, cool

Rick-Anderson:
On the other PR, See https://github.com/dotnet/AspNetCore.Docs/blob/main/.gitignore#L15-L16 and check in the solution file. Usually you don't need the sln file so we ignore it.

sammychinedu2ky:
ooh ok let me work on that

sammychinedu2ky:
Please how did you get to see the sln file when it wasn't pushed?

Rick-Anderson (approved):

Rick-Anderson:
> Please how did you get to see the sln file when it wasn't pushed?

I didn't, I just figured you had one sln for both projects because that's typical.

sammychinedu2ky:
> > Please how did you get to see the sln file when it wasn't pushed?
> 
> I didn't, I just figured you had one sln for both projects because that's typical.

ooh I see 😅

sammychinedu2ky:
So I should work on putting the test in a single project?

Rick-Anderson:
@halter73 new sample code for review that I'll put in What's new in 7 and it will also be used in the Minimal API docs.

sammychinedu2ky:
So I should work on putting the test in a single project with the web app? @Rick-Anderson

Rick-Anderson:
> So I should work on putting the test in a single project with the web app? @Rick-Anderson

I think that's the easiest for customers to download and run. I'm open to suggestions, but typically on test samples both projects are in the same sln and reference each other.

sammychinedu2ky:
> > So I should work on putting the test in a single project with the web app? @Rick-Anderson
> 
> I think that's the easiest for customers to download and run. I'm open to suggestions, but typically on test samples both projects are in the same sln and reference each other.

Alright cool.. I would do that

davidfowl:
I'm confused about what this example is trying to be showing.

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/BackgroundQueueService.cs
```diff
@@ -0,0 +1,48 @@
+using System.Text.Json;
+using System.Threading.Channels;
+
+namespace BackGroundQueueService;
```
davidfowl, line 4:
```suggestion
namespace BackgroundQueueService;
```

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/BackgroundQueueService.cs
```diff
@@ -0,0 +1,48 @@
...
+
+class BackGroundQueue : BackgroundService
```
davidfowl, line 6:
```suggestion
class BackgroundQueue : BackgroundService
```

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/BackgroundQueueService.cs
```diff
@@ -0,0 +1,48 @@
...
+                _logger.LogError(ex.Message);
+            }
+
+        }
+
+
+        return;
+
+    }
+}
```
davidfowl, line 41:
```suggestion
    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
    {

        await foreach (var dataStream in _queue.Reader.ReadAllAsync())
        {
            // reset the stream to the beginning
            dataStream.Position = 0;

            try
            {   
                var person = JsonSerializer.Deserialize<Person>(dataStream)!;
                 _logger.LogInformation($"{person.Name} is {person.Age} years and from {person.Country}");
                // you could do something else with the data
            }
            catch (Exception ex)
            {
                _logger.LogError(ex.Message);
            }
        }
    }
}
```

sammychinedu2ky:
> I'm confused about what this example is trying to be showing.

Trying to show how to make use of the piped stream and channels to read a request body

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/BackgroundQueueService.cs
```diff
@@ -0,0 +1,42 @@
...
+
+    public BackGroundQueue(Channel<Stream> queue, ILogger<BackGroundQueue> logger)
+    {
+        _queue = queue;
+        _logger = logger;
+    }
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+
+        await foreach (var dataStream in _queue.Reader.ReadAllAsync())
```
davidfowl, line 18:
```suggestion
    }
    
    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
    {
        await foreach (var dataStream in _queue.Reader.ReadAllAsync())
```
```suggestion
    }
    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
    {

        await foreach (var dataStream in _queue.Reader.ReadAllAsync())
```

sammychinedu2ky:
Thanks @davidfowl I've learnt a good deal...🙌

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+builder.Services.AddSingleton<Channel<Stream>>((_) => Channel.CreateUnbounded<Stream>());
+
+// Create a background queue service.
+builder.Services.AddHostedService<BackGroundQueue>();
+var app = builder.Build();
+
+app.Use(async (context, next) =>
+{
+    await next();
+    var reader = new StreamReader(context.Request.Body);
```
davidfowl, line 14: What's this for?

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
```diff
@@ -0,0 +1,47 @@
+using System.Threading.Channels;
+using BackGroundQueueService;
+var builder = WebApplication.CreateBuilder(args);
+// Create a channel to send data to the background queue.
+builder.Services.AddSingleton<Channel<Stream>>((_) => Channel.CreateUnbounded<Stream>());
+
+// Create a background queue service.
+builder.Services.AddHostedService<BackGroundQueue>();
```
davidfowl, line 8:
```suggestion
builder.Services.AddHostedService<BackgroundQueue>();
```

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
```diff
@@ -0,0 +1,47 @@
+using System.Threading.Channels;
+using BackGroundQueueService;
+var builder = WebApplication.CreateBuilder(args);
+// Create a channel to send data to the background queue.
+builder.Services.AddSingleton<Channel<Stream>>((_) => Channel.CreateUnbounded<Stream>());
```
davidfowl, line 5: This shouldn't be unbounded.

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+    await next();
+    var reader = new StreamReader(context.Request.Body);
+    app.Logger.LogInformation("Reading the stream from the request body.");
+});
+
+app.MapGet("/", () => "Hello World!");
+
+// curl --request POST 'http://localhost:5256/register' --header 'Content-Type: application/json' --data-raw '{ "Name":"Samson", "Age": 23, "Country":"Nigeria" }'
+app.MapPost("/register", async (Stream body, HttpRequest req, Channel<Stream> queue) =>
```
davidfowl, line 21: This API should return a 202 accepted. I'm not sure why the stream is being reused.

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+    reusableStream.Position = 0;
+
+    // Send the stream to the background queue.
+    await queue.Writer.WriteAsync(reusableStream);
+
+    // Reset the stream to the beginning.
+    reusableStream.Position = 0;
+
+    // Set the response body to the reusable stream.
+    req.Body = new MemoryStream(reusableStream.ToArray());
```
davidfowl, line 39: Why?

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
sammychinedu2ky: Just in case you want to do some logging👀

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
sammychinedu2ky: for logging and also for using the stream in a queue

sammychinedu2ky: > This API should return a 202 accepted. I'm not sure why the stream is being reused.

Ok would change that

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
sammychinedu2ky: To be able to consume the body stream more than once

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
davidfowl: Lets remove it 😄

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
davidfowl: I think we should remove this from the sample.

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
sammychinedu2ky: Alright 😅

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
```diff
@@ -0,0 +1,47 @@
+using System.Threading.Channels;
+using BackgroundQueueService;
+var builder = WebApplication.CreateBuilder(args);
+// Create a channel to send data to the background queue.
+builder.Services.AddSingleton<Channel<Stream>>((_) => Channel.CreateBounded<Stream>(100));
```
davidfowl, line 5: Maybe this should be a `Channel<byte[]>` instead?

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
sammychinedu2ky: done boss

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/BackgroundQueueService.cs
```diff
@@ -0,0 +1,43 @@
...
+        _queue = queue;
+        _logger = logger;
+    }
+    
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        await foreach (var dataStream in _queue.Reader.ReadAllAsync(stoppingToken))
+        {
+            // reset the stream to the beginning
+            dataStream.Position = 0;
```
davidfowl, line 22:
```suggestion
```

Can we assume this has been done. We should change this to a Memory<byte> though.

sammychinedu2ky: alright.. let me try that out

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
sammychinedu2ky: let me try that out

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
```diff
@@ -0,0 +1,33 @@
...
+
+// Create a background queue service.
+builder.Services.AddHostedService<BackgroundQueue>();
+var app = builder.Build();
+
+
+app.MapGet("/", () => "Hello World!");
+
```
davidfowl, line 13:
```suggestion
```

Don't need this.

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/BackgroundQueueService.cs
sammychinedu2ky: done

sammychinedu2ky: It makes sense now on why it should be a memory<byte> and not a stream 😅

sammychinedu2ky: > It makes sense now why it should be a memory<byte> and not a stream 😅

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
```diff
@@ -0,0 +1,31 @@
+using System.Threading.Channels;
+using BackgroundQueueService;
+var builder = WebApplication.CreateBuilder(args);
+// Create a channel to send data to the background queue.
+builder.Services.AddSingleton<Channel<Memory<byte>>>((_) => Channel.CreateBounded<Memory<byte>>(100));
```
davidfowl, line 5: I messed up, this should be `ReadOnlyMemory<byte>` 😄. We're close!

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
```diff
@@ -0,0 +1,31 @@
...
+
+// Create a background queue service.
+builder.Services.AddHostedService<BackgroundQueue>();
+var app = builder.Build();
+
+
```
davidfowl, line 11: nit: Delete one of these spaces.

```suggestion

```

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
```diff
@@ -0,0 +1,31 @@
...
+app.MapPost("/register", async (Stream body, Channel<Memory<byte>> queue) =>
+{
+    // Create a rewindable stream to be able to reuse the body stream.
+    var reusableStream = new MemoryStream();
+
+    // Copy the request body to the reusable stream.
+    await body.CopyToAsync(reusableStream);
+
+    // Reset the stream to the beginning.
+    reusableStream.Position = 0;
```
davidfowl, line 22: Not needed as we're throwing away the stream now.

davidfowl, line 22: Not needed as we're throwing away the stream now.

```suggestion
```

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
sammychinedu2ky: Alright 🥂

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
```diff
@@ -0,0 +1,31 @@
...
+
+    // Send the stream to the background queue.
+    await queue.Writer.WriteAsync(reusableStream.ToArray());
```
davidfowl, line 23: Now the magic to avoid double allocations:

```suggestion
        reusableStream.TryGetBuffer(out var buffer);
        await queue.Writer.WriteAsync(buffer);
```

sammychinedu2ky: Clean Code 🥂

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
sammychinedu2ky: I've learnt some new things. Thanks again @davidfowl

davidfowl:
@sammychinedu2ky The code is *almost* complete, but we need to discuss the hardest part of this PR, the performance and potential security issues:
- We're buffering the data into memory. The default max request body size on Kestrel is ~28.6 MB. That's massive.
- We're potentially going on the large object heap if we have a single buffer over 85K (this isn't great).
- Since we're using a bounded channel (yay!), we're restricting the queue to 100 items at a time. Anything about that will wait on enqueue (do we want to instead drop the request and return a 429 status code?)

To avoid making the sample more complex than it needs to be, a happy medium would be to limit the size of the buffer we use for the memory stream.

sammychinedu2ky:
> * 429

 I can use a 429 if trywrite returns false

sammychinedu2ky:
> @sammychinedu2ky The code is _almost_ complete, but we need to discuss the hardest part of this PR, the performance and potential security issues:
> 
> * We're buffering the data into memory. The default max request body size on Kestrel is ~28.6 MB. That's massive.
> * We're potentially going on the large object heap if we have a single buffer over 85K (this isn't great).
> * Since we're using a bounded channel (yay!), we're restricting the queue to 100 items at a time. Anything about that will wait on enqueue (do we want to instead drop the request and return a 429 status code?)
> 
> To avoid making the sample more complex than it needs to be, a happy medium would be to limit the size of the buffer we use for the memory stream.

But won't I still need to return a 429 if it exceeds 20?

davidfowl:
We could do some back of the napkin math here and come up with some numbers that are reasonable and show readers how those calculations affect the limits. Lets say we're willing to use 500MB on this message processing (we're on a budget here). If the max message size is 1MB, we can support up to 500 concurrent requests. We can bound the channel to 500 and use `TryWrite` to write to the channel, if it fails, we can reject the request with a 429.

Now limiting the request body to 1MB is interesting, as this will still land us in the LOH range. We also don't want to start with a 1MB buffer as that's wasteful for small buffers but we want to resize up to 1MB and then explode for buffers over that size. I'm still thinking about the most reasonable way to accomplish that without exploding people's heads.

The other option is to stay below the LOH size (85K) and allow up to 80K bodies for this endpoint (which would leave a bounded channel around 6,400).

sammychinedu2ky:
> We could do some back of the napkin math here and come up with some numbers that are reasonable and show readers how those calculations affect the limits. Lets say we're willing to use 500MB on this message processing (we're on a budget here). If the max message size is 1MB, we can support up to 500 concurrent requests. We can bound the channel to 500 and use `TryWrite` to write to the channel, if it fails, we can reject the request with a 429.
> 
> Now limiting the request body to 1MB is interesting, as this will still land us in the LOH range. We also don't want to start with a 1MB buffer as that's wasteful for small buffers but we want to resize up to 1MB and then explode for buffers over that size. I'm still thinking about the most reasonable way to accomplish that without exploding people's heads.
> 
> The other option is to stay below the LOH size (85K) and allow up to 80K bodies for this endpoint (which would leave a bounded channel around 6,400).

ok but the body size for the sample is 63 bytes

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
```diff
@@ -0,0 +1,30 @@
+using System.Threading.Channels;
+using BackgroundQueueService;
+var builder = WebApplication.CreateBuilder(args);
```
davidfowl, line 3: nit: Spacing

```suggestion
using BackgroundQueueService;

var builder = WebApplication.CreateBuilder(args);
```

davidfowl:
> ok but the body size for the sample is 63 bytes

But people copy and paste sample code into applications and will blame Microsoft when there's a security breach 😄. The sample payload if 63 bytes, but an attacked can send any size post (well up to 28MB by default).

sammychinedu2ky:
> > ok but the body size for the sample is 63 bytes
> 
> But people copy and paste sample code into applications and will blame Microsoft when there's a security breach 😄. The sample payload if 63 bytes, but an attacked can send any size post (well up to ~28.6 MB by default).

You're right 😅

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
```diff
@@ -0,0 +1,31 @@
+using System.Threading.Channels;
+using BackgroundQueueService;
+
+var builder = WebApplication.CreateBuilder(args);
+// Create a channel to send data to the background queue.
+builder.Services.AddSingleton<Channel<ReadOnlyMemory<byte>>>((_) => Channel.CreateBounded<ReadOnlyMemory<byte>>(6400));
```
davidfowl, line 6: OK lets add some comments and constants:

```suggestion
// The max memory we're willing to use for the upload endpoint on this instance
var maxMemory = 500 * 1024 * 1024;

// The max size of a single message, we're staying below the default LOH size of 85K
var maxMessageSize = 80 * 1024 * 1024;

// The max size of the queue based on those restrictions
var maxQueueSize = maxMemory / maxMessageSize;

// Create a channel to send data to the background queue.
builder.Services.AddSingleton<Channel<ReadOnlyMemory<byte>>>((_) => Channel.CreateBounded<ReadOnlyMemory<byte>>(maxQueueSize));
```

sammychinedu2ky: Yh this is better.. would help the people copying the sample code to understand the given constraint

sammychinedu2ky: but this leaves it at 6.25 bounded space which ought to be an int

davidfowl: oops, fixed.

sammychinedu2ky: okay cool

sammychinedu2ky: saw the automatic type conversion 😅

aspnetcore/fundamentals/minimal-apis/bindStreamPipeReader/7.0-samples/PipeStreamToBackgroundQueue/Program.cs
```diff
@@ -0,0 +1,40 @@
...
+    await body.CopyToAsync(reusableStream);
+
+    // get buffer to avoid double allocation
+    reusableStream.TryGetBuffer(out var buffer);
+ 
+    // Send the buffer to the background queue.
+    await queue.Writer.WriteAsync(buffer);
+    
+    return Results.Accepted();
+});
```
davidfowl, line 37:
```suggestion
app.MapPost("/register", async (HttpRequest req, Stream body, Channel<ReadOnlyMemory<byte>> queue) =>
{
    if (req.ContentLength is not null && req.ContentLength > maxMessageSize)
    {
        // Message size exceeded
        return Results.BadRequest();
    }

    // We're not above the message size and we have a content length, or
    // we're a chunked request and we're going to read up to the maxMessageSize + 1. 
    // We add one to the message size so that we can detect when a chunked request body
    // is bigger than our configured max.
    var readSize = (int?)req.ContentLength ?? (maxMessageSize + 1);

    var buffer = new byte[readSize];

    // Read at least that many bytes from the body
    var read = await body.ReadAtLeastAsync(buffer, readSize, throwOnEndOfStream: false);

    // We read more than the max, so this is a bad request
    if (read > maxMessageSize)
    {
        // Message size exceeded
        return Results.BadRequest();
    }

    // Attempt to send the buffer to the background queue.
    if (queue.Writer.TryWrite(buffer.AsMemory(0..read)))
    {
        return Results.Accepted();
    }

    // We couldn't accept the message since we're overloaded
    return Results.StatusCode(StatusCodes.Status429TooManyRequests);
});
```

Untested, you'll need to run this for me, but I think it works. Review my codez please.

sammychinedu2ky: ok. let me do that

sammychinedu2ky: yh it works

sammychinedu2ky: 🥂

sammychinedu2ky: please is this necessary ?
```
   // We read more than the max, so this is a bad request
    if (read > maxMessageSize)
    {
        // Message size exceeded
        return Results.BadRequest();
    }
```

doesn't this handle it
```
    if (req.ContentLength is not null && req.ContentLength > maxMessageSize)
    {
        // Message size exceeded
        return Results.BadRequest();
    }
```

sammychinedu2ky: Also, I don't think the +1 is needed👀

davidfowl: If the request is chunked and the content length is null, then no. I think ReadAtLeast can only read as much as the buffer that was passed in.

With your curl test, try adding `-H "Transfer-Encoding: chunked"` to try that code path.

sammychinedu2ky: Okay thanks. I would try it out. Wasn't familiar with the chunked request 😅

sammychinedu2ky: Ok, I just tried it now. And it worked when chunked (the content length was null). But my question is I'm not really clear on the relevance of the +1.

sammychinedu2ky: > if (req.ContentLength is not null && req.ContentLength > maxMessageSize)
>     {
>         // Message size exceeded
>         return Results.BadRequest();
>     }

Also, should I remove this since, ReeadAtLeast can't read more than the buffer that was passed in

sammychinedu2ky: Also, should I remove this since ReadAtLeast can't read more than the buffer that was passed in
```
    if (req.ContentLength is not null && req.ContentLength > maxMessageSize)
    {
        // Message size exceeded
        return Results.BadRequest();
    }
```

davidfowl: You need `ReadAtLeast` because you need to make sure the the buffer is filled without having to manually write a stream reading loop. With `ReadAsync`, a stream today can return less than requested, we only want to do that if we're at the end of the stream. Here's the scenario you want to test, chunked encoding plus sending a payload > 80K. That's why the +1 is needed. If the message is > 80K then we want to throw and tell the user they're sending too much information.

sammychinedu2ky: Ooh, that makes sense now. Thanks, sensei @davidfowl. Learnt something new

davidfowl (approved):
This looks clean @sammychinedu2ky. I hope you had fun with it!

sammychinedu2ky:
> This looks clean @sammychinedu2ky. I hope you had fun with it!

Yh it was quite insightful 😎and fun as well

PR #26441, opened 2022-07-15
Update call-dotnet-from-javascript.md

benjaminsampica:
Fix a reference mistake. The method is called GetHelloMessage, not GetHelloMethod.


guardrex (approved):
Thanks @benjaminsampica! :rocket: ... *Indeed!* That's another typo 😈 *gremlin* 😈 that got into the doc! 😄

Issue #26434, opened 2022-07-14
Use public static bool TryParse(string value, IFormatProvider provider, T out result);


Rick-Anderson:
Use `public static bool TryParse(string value, IFormatProvider provider, T out result);` in one of the samples
See #26433 which did minor formatting and [Bind using TryParse in MVC and API Controllers](https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-dotnet-7-preview-3/#bind-using-tryparse-in-mvc-and-api-controllers)

Probably `DateRange` is the easiest to convert.

```
public class TryParseController : ControllerBase
{
    // GET /tryparse?data=MyName
    [HttpGet]
    public ActionResult Get([FromQuery]CustomTryParseObject data) => Ok();

    public class CustomTryParseObject    // This class wouldn't be in the controller
    {
        public string? Name { get; set; }

        public static bool TryParse(string s, out CustomTryParseObject result)
        {
            if (s is null) 
            {
                result = default;
                return false;
            }

            result = new CustomTryParseObject { Name = s };
            return true;
        }
    }
}
```

cc @fiyazbinhasan

fiyazbinhasan:
I'm unsure if I understood the use case, but I assume you need to call the `TryParse` method inside the `DateRange` class with a specific `formatProvider`. In that case, you might need to add another controller like this,

```
/// <summary>
        /// /WeatherForecast/RangeWithFormatter?culture=en-GB&range=07/12/2022-07/14/2022
        /// </summary>
        /// <param name="culture"></param>
        /// <param name="range"></param>
        /// <returns></returns>
        public IActionResult RangeWithFormatter(Culture? culture, string? range)
        {
            try
            {
                if (!DateRange.TryParse(range, new CultureInfo(culture?.DisplayName ?? "en-US"), out var dateRange))
                {
                    throw new ValidationException($"Invalid date range {range}");
                }

                var weatherForecasts = Enumerable
                    .Range(1, 5).Select(index => new WeatherForecast
                    {
                        Date = DateTime.Now.AddDays(index),
                        TemperatureC = Random.Shared.Next(-20, 55),
                        Summary = Summaries[Random.Shared.Next(Summaries.Length)]
                    })
                    .Where(wf => DateOnly.FromDateTime(wf.Date) >= (dateRange?.From ?? DateOnly.MinValue)
                                 && DateOnly.FromDateTime(wf.Date) <= (dateRange?.To ?? DateOnly.MaxValue))
[...]

PR #26469, opened 2022-07-19
Tryparse with dynamic format provider supplied from client

fiyazbinhasan:
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/26434

@Rick-Anderson I've modified the sample to make the snippets more meaningful and compact.

* `/WeatherForecast?culture=en-GB` can be called with or without the `culture=en-GB`
* `/WeatherForecast/Range?range=07/12/2022-07/14/2022`, here dateRange is a required field
* `/WeatherForecast/GetByRangeWithCulture?culture=en-GB&range=07/12/2022-07/14/2022`, both culture and range is required (but the range is in a string format, we make the call `DateRange.TryParse(range, new CultureInfo(culture?.DisplayName ?? "en-US"), out var dateRange)` inside the controller to validate the parameter based on the culture privided by the client

These three examples should suffice for using both,

`public static bool TryParse(string value, T out result);`
`public static bool TryParse(string value, IFormatProvider provider, T out result);`


Rick-Anderson (approved):

Rick-Anderson:
@fiyazbinhasan  really nice work.

PR #26474, opened 2022-07-19
tryparse bind sample clean up

Rick-Anderson:


aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseMVC/Models/DateRange.cs
```diff
@@ -12,7 +12,8 @@ public DateRange(DateOnly fromDate, DateOnly toDate)
             To = toDate;
         }
 
-        public static bool TryParse(string? value, IFormatProvider? provider, out DateRange? dateRange)
+        public static bool TryParse(string? value, IFormatProvider? provider,
```
Rick-Anderson, line 15: image
@fiyazbinhasan 
 Wrong provider

Rick-Anderson: Error message is great
image

aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseMVC/Program.cs
```diff
@@ -1,11 +1,17 @@
...
 var app = builder.Build();
 
-// Configure the HTTP request pipeline.
+//         // GET /WeatherForecast/RangeWithCulture?culture=en-GB&range=07/12/2022-07/14/2022
+
+var redirectDateRange = $"/WeatherForecast/RangeWithCulture?range={DateTime.Now.ToShortDateString()}" +
+                        $"-{DateTime.Now.AddDays(5).ToShortDateString()}" +
+                        "&culture=en-GB";
+
+app.MapGet("/", () => Results.Redirect(redirectDateRange));
```
Rick-Anderson, line 13: @fiyazbinhasan  please review. This makes it easy to test so you don't need to enter date range.

fiyazbinhasan: you would want to change the above code to the following,

```
//         // GET /WeatherForecast/RangeWithCulture?culture=en-GB&range=20/07/2022-25/07/2022

var redirectDateRange = $"/WeatherForecast/RangeWithCulture?culture=en-GB&range=" +
                        $"{DateTime.Now.ToString("d", new CultureInfo("en-GB"))}" +
                        $"-{DateTime.Now.AddDays(5).ToString("d", new CultureInfo("en-GB"))}";

app.MapGet("/", () => Results.Redirect(redirectDateRange));
```
@Rick-Anderson I would assume if a user wants to pass a different culture, he would also need to pass the DateTime in that culture as well e.g. for an en-GB user the DateTime should be dd/MM/yyyy whereas for en-US MM/dd/yyyy

Rick-Anderson: @fiyazbinhasan  Thanks! Take a look at https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseMVC/Program.cs
Is that OK. I returns a valid endpoint you can test with any culture. Just copy/paste the endpoint string. The `DRC` class is clunky, it would be better to do something like the following local function
`string HelloName(string name) => $"Hello, {name}!";`

PR #26484, opened 2022-07-20
Removed block-all-mixed-content directive

### LINKED ISSUE #26483: Do not promote using the CSP directive "block-all-mixed-content"
Based on the relevant MDN articles, [block-all-mixed-content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/block-all-mixed-content) and
[upgrade-insecure-requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/upgrade-insecure-requests), the `block-all-mixed-content` is considered obsolete, and using both is never recommended.

------

> **CSP: block-all-mixed-content**
> **Deprecated**: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/block-all-mixed-content#browser_compatibility) at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

> **Warning**: This directive is marked as obsolete in the specification: all mixed content is now blocked if it can't be autoupgraded.

-------

> **CSP: upgrade-insecure-requests**
> **Note**: The upgrade-insecure-requests directive is evaluated before [block-all-mixed-content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/block-all-mixed-content) and if it is set, the latter is effectively a no-op. It is recommended to set either directive, but not both, unless you want to force HTTPS on older browsers that do not force it after a redirect to HTTP.

I suggest removing the CSP directive `block-all-mixed-content` from the page and leave only `upgrade-insecure-requests`, as it will be as secure as the alternative, but won't produce errors on the page, and will rewrite URLs to HTTPS endpoints.

---
#### Document Details

* ID: 88177a7d-c9ff-c245-5629-a462c9258abf
* Version Independent ID: 6e0b5c52-90a1-5ca6-bfad-df33a8beae6c
* Content: [Enforce a Content Security Policy for ASP.NET Core Blazor](https://docs.microsoft.com/en-us/aspnet/core/blazor/security/content-security-policy?view=aspnetcore-6.0)
* Content Source: [aspnetcore/blazor/security/content-security-policy.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/security/content-security-policy.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

yugabe:
Removed the block-all-mixed-content directive from the list of recommended policies.

Fixes #26483

### DIFF
--- a/aspnetcore/blazor/security/content-security-policy.md
+++ b/aspnetcore/blazor/security/content-security-policy.md
@@ -38,7 +38,6 @@ CSP is supported in most modern desktop and mobile browsers, including Chrome, E
 Minimally, specify the following directives and sources for Blazor apps. Add additional directives and sources as needed. The following directives are used in the *Apply the policy* section of this article, where example security policies for Blazor WebAssembly and Blazor Server are provided:
 
 * [base-uri](https://developer.mozilla.org/docs/Web/HTTP/Headers/Content-Security-Policy/base-uri): Restricts the URLs for a page's `<base>` tag. Specify `self` to indicate that the app's origin, including the scheme and port number, is a valid source.
-* [block-all-mixed-content](https://developer.mozilla.org/docs/Web/HTTP/Headers/Content-Security-Policy/block-all-mixed-content): Prevents loading mixed HTTP and HTTPS content.
 * [default-src](https://developer.mozilla.org/docs/Web/HTTP/Headers/Content-Security-Policy/default-src): Indicates a fallback for source directives that aren't explicitly specified by the policy. Specify `self` to indicate that the app's origin, including the scheme and port number, is a valid source.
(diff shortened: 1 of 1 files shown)


guardrex:
@yugabe ... We need to discuss this first on the issue. Let's hold there for a minute. I'll ping the right 🐈 to take a look. We can re-open this if these are the changes to make.

Issue #26423, opened 2022-07-13
.NET 7: Bind using TryParse in MVC and API Controllers


Rick-Anderson:
blog [Bind using TryParse in MVC and API Controllers](https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-dotnet-7-preview-3/#bind-using-tryparse-in-mvc-and-api-controllers)

Code from #26378

PR #26488, opened 2022-07-20
Bind with TryParse /5

Rick-Anderson:
Contributes to https://github.com/dotnet/AspNetCore.Docs/issues/26423

[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/mvc/models/model-binding?view=aspnetcore-7.0&branch=pr-en-us-26488#bind-with-tryparse)

Sample code at https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse


aspnetcore/mvc/models/model-binding.md
```diff
@@ -189,6 +189,25 @@ The simple types that the model binder can convert source strings into include t
 * [Uri](xref:System.UriTypeConverter)
 * [Version](xref:System.ComponentModel.VersionConverter)
 
+## Bind with TryParse
+
+The following APIs support binding controller action parameter values using a `TryParse`:
```
brunolins16, line 194: I am not sure if it is in the docs yet, but do you think we need to mention about `IParseable<T>`

Rick-Anderson: @brunolins16  can you point me to a sample with `IParseable<T>` and we'll add that.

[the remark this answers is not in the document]
Rick-Anderson: @fiyazbinhasan  PLEASE review

aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseMVC/Models/Culture.cs
```diff
@@ -0,0 +1,38 @@
...
+
+            try
+            {
+                CultureInfo ci = new CultureInfo(value);
+            }
+            catch
+            {
+                culture = default;
+                return false;
+            }
```
Rick-Anderson, line 31: @fiyazbinhasan  I had to put in this hacky try catch to get `TryParse ` to work correctly.  With your latest code when you enter `12` for the culture
`/12/WeatherForecast/RangeWithCulture?range=07/12/2022-07/14/2022`
you get a run-time exception. With my hack code, I get a nice error message.
I think I'll remove all the culture code unless you can figure out a way to make it clean.

Nguyendaotyty:
Vào 7:41, Th 6, 22 thg 7, 2022 Rick Anderson ***@***.***> đã
viết:

aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseMVC/Models/DateRange.cs
```diff
@@ -15,7 +12,7 @@ public DateRange(DateOnly fromDate, DateOnly toDate)
             To = toDate;
         }
 
-        public static bool TryParse(string? value, IFormatProvider? provider,
+        public static bool TryParse(string? value,
```
Rick-Anderson, line 15: @fiyazbinhasan  I removed the format provider since it's not used. Actually, when I added
`<PackageReference Include="System.Runtime.Experimental" Version="7.0.0-preview.2.22152.2" />`
I got a compile error with the format provider in the code.

aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseMVC/Models/Point.cs
```diff
@@ -0,0 +1,38 @@
...
+            && double.TryParse(segments[1], out var y))
+        {
+            point = new Point { X = x, Y = y };
+            return true;
+        }
+
+        point = new Point();
+        return false;
+    }
+}
```
Rick-Anderson, line 37: @brunolins16  wrote:
> I am not sure if it is in the docs yet, but do you think we need to mention about `IParseable<T>`

@fiyazbinhasan  suggestion, I added @DamianEdwards sample here. But when I run the app I get
```
Unhandled exception. System.Reflection.ReflectionTypeLoadException: Unable to load one or more of the requested types.
Could not load type 'System.IParseable`1' from assembly 'System.Runtime, Version=7.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
   at System.Reflection.RuntimeModule.GetTypes(RuntimeModule module)
```
It doesn't like like `IParseable<T>` is ready.

aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseMVC/Models/Culture.cs
fiyazbinhasan: @Rick-Anderson I suggested throwing out this Culture class altogether and using it as a string to demonstrate the second example of DateRange.

aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseMVC/Models/Point.cs
fiyazbinhasan: I agree.

aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseMVC/Models/DateRange.cs
```diff
@@ -26,8 +23,8 @@ public DateRange(DateOnly fromDate, DateOnly toDate)
             
             var range = value.Split('-');
 
-            if (!DateOnly.TryParse(range[0], provider, out var fromDate) 
-             || !DateOnly.TryParse(range[1], provider, out var toDate))
+            if (!DateOnly.TryParse(range[0],  out var fromDate)
```
fiyazbinhasan, line 26: removing the provider doesn't demonstrate the second tryparse method,
`public static bool TryParse(string value, IFormatProvider provider, T out result);`

aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseMVC/Models/Culture.cs
fiyazbinhasan: I'm not sure why we would want to pass an integer like 12, where the culture display name should be something en-GB, en-US, etc.

`/en-US/WeatherForecast/RangeWithCulture?range=07/12/2022-07/14/2022`

aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseMVC/Models/DateRange.cs
Rick-Anderson: @fiyazbinhasan great catch. I removed it because when I added the experimental package, I was getting compile errors with it. Reverted!

aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseMVC/Models/Culture.cs
Rick-Anderson: @fiyazbinhasan 
> I'm not sure why we would want to pass an integer like 12, where the culture display name should be something en-GB, en-US, etc.

But that's the entire point of `TryParse`, it returns false if it can't do the conversion. It doesn't have to be `12`, it can be any miss-typed string that can't be turned into a culture.

> @Rick-Anderson I suggested throwing out this Culture class altogether and using it as a string to demonstrate the second example of DateRange.

Good point. Maybe we should leave it out and let it throw when someone enters a culture that can't be converted. 
Maybe you or @brunolins16 can think of a clever way to `TryParse` culture

fiyazbinhasan: @Rick-Anderson Oh got it. I thought you wanted to remove the culture class. In that case, the way you converted the string to culture using try/catch seems ok. I guess it's the only way you can do it. :)

Rick-Anderson: But the code could be cleaned up to replace 
```
 public string? DisplayName { get; }

        public Culture(string displayName)
        {
            DisplayName = displayName;
        }
```

with saving `new CultureInfo(value);` and not `DisplayName`
let me know if you can do that.

fiyazbinhasan: @Rick-Anderson you can do something like this,

```
public class Culture
    {
        public CultureInfo? CultureInfo { get; }

        public Culture(CultureInfo cultureInfo)
        {
            CultureInfo = cultureInfo;
        }

        public static bool TryParse(string? value, out Culture? culture)
        {
            if (string.IsNullOrEmpty(value))
            {
                culture = default;
                return false;
            }

            try
            {
                var ci = new CultureInfo(value);
                culture = new Culture(ci);
                return true;
            }
            catch
            {
                culture = default;
                return false;
            }
        }
    }
```

Also, I figured out that some culture uses dash (-) as a separator symbol. In that case, it's better to go with a comma for a range separator. Let me know if you want me to create a pull request with those changes.

Rick-Anderson: > Also, I figured out that some culture uses dash (-) as a separator symbol. In that case, it's better to go with a comma for a range separator. Let me know if you want me to create a pull request with those changes.

Yes, please do. I'll `git pull origin main` to bring in your changes.

aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseMVC/Models/Point.cs
DamianEdwards: There's no "e" in it now: `IParsable`

aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseMVC/Models/Culture.cs
fiyazbinhasan: @Rick-Anderson could you take a look at the branch, and see if the samples are enough. If I get a go I'll create a PR afterwards,

https://github.com/fiyazbinhasan/Docs/tree/try-parse-with-iparsable/aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse

Rick-Anderson:
Moved to #26608

PR #26534, opened 2022-07-25
TryParse with IParsable<TSelf>

fiyazbinhasan:
@Rick-Anderson please take a look. I've made some changes to reflect the use of the IParsable interface. However, I came across this issue, https://github.com/dotnet/aspnetcore/issues/42286 tagged with .NET 8 planning. 

https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-dotnet-7-preview-3/ in this blog it is mentioned that we have two methods to deal with the parsing,

```
public static bool TryParse(string value, T out result);
public static bool TryParse(string value, IFormatProvider provider, T out result);
```

With IParsable<T>, we have the following two methods that need to implement

```
static abstract TSelf Parse(string s, IFormatProvider? provider);
static abstract bool TryParse([NotNullWhen(true)] string? s, IFormatProvider? provider, out TSelf result);

```

The product team might be able to give a better idea here, like which one to use and what should be reflected in the documentation


Rick-Anderson:
@brunolins16 please comment on `IParsable`

brunolins16:
> However, I came across this issue, https://github.com/dotnet/aspnetcore/issues/42286 tagged with .NET 8 planning.

The issue is an improvement in our detection mechanism, to avoid Reflection and use the interface as first option. however, it is not mentioned in the blog post but the method from the IParsable is allowed as well.

```c#
static abstract bool TryParse([NotNullWhen(true)] string? s, IFormatProvider? provider, out TSelf result);
```

brunolins16:
> The product team might be able to give a better idea here, like which one to use and what should be reflected in the documentation

I would love to have both in the docs but maybe it will be confusing, and I think it will be more interesting to advocate the `IParsable` interface. 

@dotnet/minimal-apis any thoughts about which one we would prefer to have in docs?

aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseAPI/Controllers/WeatherForecastController.cs
```diff
@@ -57,46 +48,48 @@ public IActionResult GetByRange([FromQuery] DateRange range)
...
-        // </snippet>
+        // </snippet_2>
 
-        // GET /en-gb/WeatherForecast/GetByRangeWithCulture?range=21/07/2022-23/07/2022
+        // <snippet_3>
+        // GET /WeatherForecast/ByCulturalRange?culture=af-ZA&range=2022-07-24,2022-07-29
         [HttpGet]
-        [Route("/{culture}/[controller]/GetByRangeWithCulture")]
-        public IActionResult GetByRangeWithCulture([FromRoute] string culture, [FromQuery] string range)
+        [Route("ByCulturalRange")]
```
brunolins16, line 68: You can keep the `culture` in the route if you want to.

aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseAPI/Models/Culture.cs
```diff
@@ -0,0 +1,42 @@
+using System.Globalization;
+
+namespace BindTryParseAPI.Models
+{
+    // <snippet>
+    public class Culture : IParsable<Culture>
```
brunolins16, line 6: Could it just be `Culture: CultureInfo, IParsable<Culture>`, so you don't need to have a `CultureInfo` property?

aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseAPI/Models/DateRange.cs
```diff
@@ -1,36 +1,35 @@
 namespace BindTryParseAPI.Models
 {
     // <snippet>
-    public class DateRange
+    public class DateRange : IParsable<DateRange>
     {
-        public DateOnly? From { get; }
-        public DateOnly? To { get; }
+        public DateOnly? From { get; set; }
```
brunolins16, line 6:
```suggestion
        public DateOnly? From { get; init; }
```

Or keep it readonly by making it private?

aspnetcore/mvc/controllers/bind-tryparse/7.0-samples/BindUsingTryParse/BindTryParseAPI/Models/DateRange.cs
```diff
@@ -1,36 +1,35 @@
...
+        public DateOnly? To { get; set; }
```
brunolins16, line 7:
```suggestion
        public DateOnly? To { get; init; }
```

Or keep it readonly by making it private?

brunolins16: Maybe keep the ctor and not change the properties is another option.

brunolins16 (approved):
I have small suggestions, but the change look good to me. Can you point me to where the snippets are been used?

Rick-Anderson:
> I have small suggestions, but the change look good to me. Can you point me to where the snippets are been used?
@brunolins16 
I'll write the article after the code has been accepted, so the snippets aren't being used right now.

fiyazbinhasan:
@Rick-Anderson, modified the codebase based on @brunolins16 suggestions. One thing I did is changed the Culture class to Locale. There is a middleware called `AddRequestLocalization`, which by default adds the culture to the query string like this
`/?culture=en-Gb`. I don't want to confuse the user between these two, as someone might think that this is the actual way to add localization to your app. But that's a completely different use case. So, I decided to go with an URL like, `/en-gb/WeatherForecast?range=....`

PR #26595, opened 2022-07-30
Remove duende from the core identity docs

davidfowl:
While Duende does use Identity to manage users that is unrelated to explaining is at it's core. 
We already have an external link to the Dueude identity server, this doc shouldn't mention it when explaining what identity is.


davidfowl:
cc @blowdart @HaoK

Rick-Anderson (approved):

HaoK (approved):

Issue #26515, opened 2022-07-22
Rate limiting middleware


Rick-Anderson:
* [RateLimiting middleware](https://devblogs.microsoft.com/dotnet/announcing-rate-limiting-for-dotnet/#ratelimiting-middleware) from the blog [Announcing Rate Limiting for .NET](https://devblogs.microsoft.com/dotnet/announcing-rate-limiting-for-dotnet/)
* [Rate limiting middleware](https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-dotnet-7-preview-4/#rate-limiting-middleware) 
* [Blog update](https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-dotnet-7-preview-6/#updates-to-rate-limiting-middleware) 
* https://github.com/dotnet/aspnetcore/issues/41667

[System.Threading.RateLimiting](https://www.nuget.org/packages/System.Threading.RateLimiting) will be documented in  https://github.com/dotnet/docs/issues/30426  and this article will point to that. This article will document the [Microsoft.AspNetCore.RateLimiting](https://www.nuget.org/packages/Microsoft.AspNetCore.RateLimiting) NuGet package.

@BrennanConroy do you have any ASP.NET Core samples I can start with?

Jhstvf4t:
I don't have money anymore to go out to the workshop, I want to eat, let alone me

Rick-Anderson:
@Kahbazi are you interested in writing the sample code for rate limiting middleware?

Kahbazi:
@Rick-Anderson I have not write any docs yet, but I would like to give it a try. I'm not familiar with this repo. Where should I start? Are there any specifics samples you have in mind?

Jhstvf4t:
Can I call you?

Pada tanggal Sab, 30 Jul 2022 13:49, Kahbazi ***@***.***>
menulis:

Rick-Anderson:
@Kahbazi  I'll handle the document. It would be great if you could write a sample showing how to use [System.Threading.RateLimiting](https://www.nuget.org/packages/System.Threading.RateLimiting).  I have a basic sample [here](https://github.com/Rick-Anderson/WebRateLimitMw) but [`MyCustomLimiter : RateLimiter`](https://github.com/Rick-Anderson/WebRateLimitMw/blob/main/MyCustomLimiter.cs) isn't implemented.

Kahbazi:
[...]

PR #26623, opened 2022-08-04
Rate limit middleware /5

Rick-Anderson:
Contributes to #26515
Code only in the PR. Once code is approved and merge, I'll write the doc.


aspnetcore/fundamentals/middleware/rate-limit/samples/WebRateLimitAuth/Program.cs
```diff
@@ -0,0 +1,197 @@
+#define ADMIN // FIRST ADMIN
```
Rick-Anderson, line 1: @Kahbazi @bugproof @pentp please review rate limiting. Do we need more samples? I think I'll leave the [custom rate limiter](https://github.com/Rick-Anderson/WebRateLimitMw/blob/main/MyCustomLimiter.cs) for another PR. Most folks won't need to implement `RateLimiter`

bugproof: Seems fine for a start.

aspnetcore/fundamentals/middleware/rate-limit/samples/WebRateLimitAuth/Program.cs
```diff
@@ -0,0 +1,197 @@
...
+app.UseRateLimiter(new RateLimiterOptions()
+    .AddConcurrencyLimiter(policyName: getPolicyName,
+          new ConcurrencyLimiterOptions(permitLimit: 2,
+          queueProcessingOrder: QueueProcessingOrder.OldestFirst, queueLimit: 2))
+    .AddNoLimiter(policyName: adminPolicyName)
+    .AddPolicy(policyName: postPolicyName, partitioner: httpContext =>
+    {
+        if (!StringValues.IsNullOrEmpty(httpContext.Request.Headers["my-token"]))
+        {
+            return RateLimitPartition.CreateTokenBucketLimiter("token", key =>
```
Kahbazi, line 166: Is this suppose to limit based on each individual token or just one limiter for all tokens. If it's the first one I think the code needs to change.
```suggestion
            return RateLimitPartition.CreateTokenBucketLimiter(httpContext.Request.Headers["my-token"], key =>
```

aspnetcore/fundamentals/middleware/rate-limit/samples/WebRateLimitAuth/Program.cs
```diff
@@ -0,0 +1,197 @@
...
+
+var userPolicyName = "user";
+
+var options = new RateLimiterOptions()
+    .AddPolicy<string>(userPolicyName, context =>
+    {
+        if (!context.User?.Identity?.IsAuthenticated ?? false)
+        {
+            var username = context?.User?.Identity?.Name ?? "null user";
+
```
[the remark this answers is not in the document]
bugproof: `unlimitedUsers` is not clear enough name. I would name it like `usernamesWithNoRateLimit`

aspnetcore/fundamentals/middleware/rate-limit/samples/WebRateLimitAuth/Program.cs
```diff
@@ -0,0 +1,197 @@
...
+// <snippet>
+app.UseAuthentication();
+app.UseAuthorization();
+
+var userPolicyName = "user";
+
+var options = new RateLimiterOptions()
+    .AddPolicy<string>(userPolicyName, context =>
+    {
+        if (!context.User?.Identity?.IsAuthenticated ?? false)
```
pentp, line 50: How is this supposed to work? For unauthenticated users the Identity.Name will be empty. And this check seems to skip null User/Identity entirely...

aspnetcore/fundamentals/middleware/rate-limit/samples/WebRateLimitAuth/Program.cs
```diff
@@ -0,0 +1,197 @@
...
+        }
+    });
+
+options.GlobalLimiter = PartitionedRateLimiter.Create<HttpContext, IPAddress>(context =>
+{
+    var remoteIPaddress = context?.Connection?.RemoteIpAddress ?? new IPAddress(5001);
+
+    if (!IPAddress.IsLoopback(remoteIPaddress )) //context.Connection.RemoteIpAddress))
+    {
+        return RateLimitPartition.CreateSlidingWindowLimiter<IPAddress>
```
pentp, line 75: It would be more appropriate to use a TokenBuckedRateLimiter here and it would also provide an example of a different kind of rate limiter.

aspnetcore/fundamentals/middleware/rate-limit/samples/WebRateLimitAuth/Program.cs
Rick-Anderson: If user is anonymous, they are rate limited under the `"anonymous user"` partition key. Yes, null user is under this partition.

aspnetcore/fundamentals/middleware/rate-limit/samples/WebRateLimitAuth/Program.cs
```diff
@@ -0,0 +1,196 @@
...
+                  segmentsPerWindow: 1
+                ));
+        }
+        else
+        {
+            return RateLimitPartition.CreateNoLimiter<string>(string.Empty);
+        }
+    });
+
+options.GlobalLimiter = PartitionedRateLimiter.Create<HttpContext, IPAddress>(context =>
```
Rick-Anderson, line 69: @pentp  wrote
> It would be more appropriate to use a TokenBuckedRateLimiter here and it would also provide an example of a different kind of rate limiter.

I changed this to `TokenBuckedRateLimiter`. Because it uses the `remoteIPaddress` for the partition key, we'll have too many partitions. Shouldn't this use a fixed partition key for non-loopback IPs? Why do we need all those partition keys when the rate limiter is the same?

aspnetcore/fundamentals/middleware/rate-limit/samples/WebRateLimitAuth/Program.cs
pentp: But the current code means that authenticated users and requests with no user info (unauthenticated) will end up using `CreateNoLimiter` partition, while unauthenticated users with empty user info will all end up using the same `CreateSlidingWindowLimiter` partition. What's the point of that?

Rick-Anderson: > But the current code means that authenticated users and requests with no user info (unauthenticated) will end up using `CreateNoLimiter` partition, while unauthenticated users with empty user info will all end up using the same `CreateSlidingWindowLimiter` partition. What's the point of that?

No, it's the opposite, that's what `!` means.

Rick-Anderson: @pentp oops, you're right. I forgot to change the  null-coalescing operator `?? ` from `false` to `true`. It's fixed now
`if (!context.User?.Identity?.IsAuthenticated ?? true)`

aspnetcore/fundamentals/middleware/rate-limit/samples/WebRateLimitAuth/Program.cs
Rick-Anderson: > It would be more appropriate to use a TokenBuckedRateLimiter here and it would also provide an example of a different kind of rate limiter.

Changed it. Please review. Are parameters reasonable?

aspnetcore/fundamentals/middleware/rate-limit/samples/WebRateLimitAuth/Controllers/HomeController.cs
```diff
@@ -0,0 +1,33 @@
...
+        public IActionResult Index()
+        {
+            return View();
+        }
+
+        public async Task<IActionResult> Privacy()
+        {
+            await Task.Delay(100);
+            return View();
+        }
```
Rick-Anderson, line 25: The MVC controller only uses the global rate limiter. Using Azure load testing with a JMeter jmx script, I start getting 503 errors pretty quickly.

It looks like you can't apply the rate controller to specific actions, controllers, or razor pages. Is that right?

Rick-Anderson: I tried to test this in chrome by refreshing all tabs but chrome must use a single thread or something because even with 50 tabs all refreshing I couldn't get a 503 error and each tab would show busy. They eventually all responded with no errors.

aspnetcore/fundamentals/middleware/rate-limit/samples/WebRateLimitAuth/Program.cs
```diff
@@ -0,0 +1,226 @@
...
+            ((int)retryAfter.TotalSeconds).ToString(NumberFormatInfo.InvariantInfo);
+        }
+
+        context.HttpContext.Response.StatusCode = StatusCodes.Status429TooManyRequests;
+        context.HttpContext.Response.Headers.Add("Retry-After", "100");
+        app.Logger.LogWarning($"policy: {userPolicyName} {GetUserEndPoint(context.HttpContext)}");
+
+        return new ValueTask();
+    }
+}
```
Rick-Anderson, line 66: @pentp @Kahbazi I added `OnRejected` and logging for testing only. How do I set  `RetryAfter` so I can create `Response.Headers.RetryAfter` ?

Kahbazi: AFAIK you don't need to set `RetryAfter`. It will be added by the `RateLimiter`.

aspnetcore/fundamentals/middleware/rate-limit/samples/WebRateLimitAuth/Controllers/HomeController.cs
Kahbazi: > The MVC controller only uses the global rate limiter.

Right now there's no `RequireRateLimiting` on `MapDefaultControllerRoute`. 
https://github.com/dotnet/AspNetCore.Docs/blob/e6dbddea9c9765b06b2026596f903bc89801bd47/aspnetcore/fundamentals/middleware/rate-limit/samples/WebRateLimitAuth/Program.cs#L111

> It looks like you can't apply the rate controller to specific actions, controllers, or razor pages. Is that right?

You will be able to do it after https://github.com/dotnet/aspnetcore/pull/43053 get merged.

>  tried to test this in chrome by refreshing all tabs but chrome must use a single thread or something because even with 50 tabs all refreshing I couldn't get a 503 error and each tab would show busy.

Right now there's no limiter with loopback IP. I'm guessing maybe you are calling from 127.0.0.1 🤔

aspnetcore/fundamentals/middleware/rate-limit/samples/WebRateLimitAuth/Program.cs
Kahbazi: > Shouldn't this use a fixed partition key for non-loopback IPs? Why do we need all those partition keys when the rate limiter is the same?

It depends on how we want the limiter to work. Yes there would be too many partitions but in this scenario each IP would be limited separately. 

If there's only one partition key for non-loopback IPs. One IP could send too many requests and block the system for all other IPs.

aspnetcore/fundamentals/middleware/rate-limit/samples/WebRateLimitAuth/Controllers/HomeController.cs
Rick-Anderson: You're right. Now I get the timeout:
```

    //if (!IPAddress.IsLoopback(remoteIPaddress!))
    //{
        return RateLimitPartition.CreateTokenBucketLimiter<IPAddress>
           (remoteIPaddress!, key =>
                 new TokenBucketRateLimiterOptions(tokenLimit: 5,
                     queueProcessingOrder: QueueProcessingOrder.OldestFirst,
                     queueLimit: 1,
                     replenishmentPeriod: TimeSpan.FromSeconds(15),
                     tokensPerPeriod: 1,                    
                     autoReplenishment: true));
    //}
    //else
    //{
    //    return RateLimitPartition.CreateNoLimiter<IPAddress>(IPAddress.Loopback);
    //}
```

Rick-Anderson:
@pentp @Kahbazi  I moved this to https://github.com/dotnet/AspNetCore.Docs.Samples/pull/13

PR #26656, opened 2022-08-08
Fix incorrect indefinite article - "a" vs "an"

GitHubPang:


guardrex (approved):
Thanks @GitHubPang! ⛵

Issue #26655, opened 2022-08-07
.NET 7.0 Update: Use hubs in ASP.NET Core SignalR: Introduce dependency injection feature.


wadepickett:
In the 7.0 version of the topic Use hubs in ASP.NET Core SignalR:  add a new section introducing dependency injection for SignalR. 
 This feature added in .NET 7.0 preview 2.  

This issue has a parent tracking issue: #26477

Making use of the introduction work already provided in the blog preview docs where possible:  [Dependency injection for SignalR hub methods](https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-dotnet-7-preview-2/#dependency-injection-for-signalr-hub-methods)

Then follow up by providing what is needed in [What's new in ASP.NET Core 7.0](https://docs.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-7.0?view=aspnetcore-6.0) to represent it, and a link back.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1387f556-b4b7-70f9-5bd0-f0bccc597689
* Version Independent ID: fb016cdc-d281-6b5b-7466-2c718e7274b4
* Content: [Use hubs in ASP.NET Core SignalR](https://docs.microsoft.com/en-us/aspnet/core/signalr/hubs?view=aspnetcore-7.0)
* Content Source: [aspnetcore/signalr/hubs.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/signalr/hubs.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-signalr**
* GitHub Login: @bradygaster
* Microsoft Alias: **bradyg**

BrennanConroy:
Feature has already been documented: https://github.com/dotnet/AspNetCore.Docs/pull/25746

PR #26658, opened 2022-08-08
Prep 7.0 moniker section copy Hubs in SignalR

wadepickett:
Adding a 7.0 version of the topic in prep for Adding .NET 7.0 specific related features for #26655


wadepickett:
Closing draft, just realized 7.0 moniker was already previously added at bottom of doc rather than top.

PR #22770, opened 2021-07-19
Binding select multiple

guardrex:
Fixes #22448
Addresses #22448

🛑 ~**HOLD FOR RELEASE OF PREVIEW 7**~ 🛑 We're going to go ahead with the coverage **_BUT_** with a NOTE that it applies to Preview 7. This is blocking other work. I'll revert the NOTE when Preview 7 releases.

* [Internal Review Topic (Data Binding, links to section)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/components/data-binding?view=aspnetcore-3.1&branch=pr-en-us-22770#multiple-option-selection-with-input-elements)
* [Internal Review Topic (Forms and Validation, links to section)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/forms-validation?view=aspnetcore-3.1&branch=pr-en-us-22770#multiple-option-selection-with-the-inputselect-component)

On the PR:

* The multiple select feature for Preview 7 release. *I'll hold this PR until Preview 7 comes out.*
* Moves the `null` binding behavior section from the *Forms Validation* topic to the *Data Binding* topic and cross-links to it from *Forms Validation*. Cross-links also from each topic's new multiple select feature coverage.

Question ❓:

For `InputSelect`, the `multiple` attribute is inferred when bound to an array type. However, the example for binding `<select>` doesn't exhibit the same setup in the PU repo sample ...

https://github.com/dotnet/aspnetcore/blob/main/src/Components/test/testassets/BasicTestApp/SelectVariantsComponent.razor

Is inferred `multiple` a thing for binding `<select>`, too, and the sample just doesn't show it? If so, I'll mirror the `InputSelect` coverage remark in the `<select>` coverage and probably modify the `<select>` coverage example to drop `multiple` in that scenario, too (i.e., let it infer the attribute as the `InputSelect` example does).

PR #26691, opened 2022-08-10
Changed <input> to <select>

mrlife:


wadepickett (approved):
Looks good.  Verified we are using select in the example.

wadepickett:
@mrlife, thanks for taking the time to identify and contribute this fix!

wadepickett (changes requested):
We have more than one version of this doc in this file.  I'll update the others the same way.

wadepickett (approved):
Both 6.0 and 7.0 versions fixed.  This section did not exist before version 6.0 in the doc.

guardrex:
Thanks guys!!!! Sorry about this. Both Mackinnion and I didn't catch it when I put the original content up over a year ago on [Binding select multiple (dotnet/AspNetCore.Docs #22770)](https://github.com/dotnet/AspNetCore.Docs/pull/22770). Woe! It amazes me how long these 😈 *grimilins* 😈 go before a sharp 👁️ catches them. Thanks again, @mrlife! 🎷

PR #26768, opened 2022-08-18
.NET 7 Preview 7 update: WebSockets support in ASP.NET Core

### LINKED ISSUE #26366: Doc HTTP/2 WebSockets support
This is now available for Kestrel in 7.0-preview6. Notes from the blog post:

[WebSockets](https://www.rfc-editor.org/rfc/rfc6455.html) were originally designed for HTTP/1.1 but have since been [adapted](https://www.rfc-editor.org/rfc/rfc8441.html) to work over HTTP/2. Using WebSockets over HTTP/2 lets you take advantage of new features like header compression and multiplexing that reduce the time and resources needed when making multiple requests to the server. That support is now available in Kestrel on all HTTP/2 enabled platforms. The version negotiation is automatic in browsers and Kestrel so no new APIs are needed, you can follow the [existing samples](https://docs.microsoft.com/aspnet/core/fundamentals/websockets). Note HTTP/2 WebSockets use CONNECT requests rather than GET, so your routes and controllers may need updating. Chrome and Edge have HTTP/2 WebSockets enabled by default and you can enable it in FireFox on the about:config page with the network.http.spdy.websockets flag. SignalR server and the browser JavaScript client also support it.
Support is coming to [ClientWebSocket](https://github.com/dotnet/runtime/issues/69669) and [YARP](https://github.com/microsoft/reverse-proxy/issues/1375) soon. 

---
#### Document Details

* Content: [WebSockets support in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/websockets?view=aspnetcore-6.0)
* Content Source: [aspnetcore/fundamentals/websockets.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/websockets.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

wadepickett:
[Internal Review](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/websockets?view=aspnetcore-7.0&branch=pr-en-us-26768)

Fixes #26366


aspnetcore/fundamentals/websockets.md
```diff
@@ -14,17 +14,50 @@ uid: fundamentals/websockets
...
+## Http/2 WebSockets support
+
+Using WebSockets over HTTP/2 takes advantage of new features such as:
+
+* Header compression.
+* Multiplexing, which reduces the time and resources needed when making multiple requests to the server.
+
+These supported features are available in Kestrel on all HTTP/2 enabled platforms. The version negotiation is automatic in browsers and Kestrel, so no new APIs are needed.
+
+.NET 7 introduced Websockets over HTTP/2 support for Kestrel, SignalR JavaScript client and SignalR with Blazor WebAssembly.
```
tdykstra, line 26:
```suggestion
.NET 7 introduced Websockets over HTTP/2 support for Kestrel, the SignalR JavaScript client, and SignalR with Blazor WebAssembly.
```

aspnetcore/fundamentals/websockets.md
```diff
@@ -14,17 +14,50 @@ uid: fundamentals/websockets
...
+
+> [!NOTE]
+> HTTP/2 WebSockets use CONNECT requests rather than GET, so your own routes and controllers may need updating.
+> Chrome and Edge have HTTP/2 WebSockets enabled by default, and you can enable it in FireFox on the about:config page with the network.http.spdy.websockets flag.
```
tdykstra, line 30:
```suggestion
> Chrome and Edge have HTTP/2 WebSockets enabled by default, and you can enable it in FireFox on the about:config page with the `network.http.spdy.websockets` flag.
```

aspnetcore/fundamentals/websockets.md
```diff
@@ -14,17 +14,50 @@ uid: fundamentals/websockets
...
+
+* Header compression.
+* Multiplexing, which reduces the time and resources needed when making multiple requests to the server.
+
+These supported features are available in Kestrel on all HTTP/2 enabled platforms. The version negotiation is automatic in browsers and Kestrel, so no new APIs are needed.
+
+.NET 7 introduced Websockets over HTTP/2 support for Kestrel, SignalR JavaScript client and SignalR with Blazor WebAssembly.
+
+> [!NOTE]
+> HTTP/2 WebSockets use CONNECT requests rather than GET, so your own routes and controllers may need updating.
```
tdykstra, line 29: What needs to be done to them? If they have HttpGet attribute there is no HttpConnect attribute, although this text seems to imply you can just change ...Get to ...Connect.

Also, I tried the Websockets sample app in .NET 7 and it tries to use Http/2 but it doesn't work, so something like what we're telling people to do in their own apps needs to be done in the sample app.

aspnetcore/fundamentals/websockets.md
```diff
@@ -14,17 +14,50 @@ uid: fundamentals/websockets
...
 
-For most applications, we recommend SignalR over raw WebSockets. SignalR provides transport fallback for environments where WebSockets isn't available. It also provides a basic remote procedure call app model. And in most scenarios, SignalR has no significant performance disadvantage compared to using raw WebSockets.
+For most applications, we recommend SignalR over raw WebSockets.
+
+SignalR provides:
+
+* Transport fallback for environments where WebSockets isn't available.
+* A basic remote procedure call app model.
+
+SignalR has no significant performance disadvantage compared to using raw WebSockets in most scenarios.
```
tdykstra, line 45:
```suggestion
SignalR:

* Provides transport fallback for environments where WebSockets isn't available.
* Provides a basic remote procedure call app model.
* Has no significant performance disadvantage compared to using raw WebSockets in most scenarios.
```

aspnetcore/fundamentals/websockets.md
```diff
@@ -14,17 +14,50 @@ uid: fundamentals/websockets
...
+
+WebSockets were originally designed for HTTP/1.1 but have since been adapted to work over HTTP/2. 
 
 ## SignalR
 
 [ASP.NET Core SignalR](xref:signalr/introduction) is a library that simplifies adding real-time web functionality to apps. It uses WebSockets whenever possible.
 
-For most applications, we recommend SignalR over raw WebSockets. SignalR provides transport fallback for environments where WebSockets isn't available. It also provides a basic remote procedure call app model. And in most scenarios, SignalR has no significant performance disadvantage compared to using raw WebSockets.
+For most applications, we recommend SignalR over raw WebSockets.
+
```
tdykstra, line 39: The original text all in one paragraph made it clear that the "SignalR provides..." text explained the recommendation. Splitting it into separate paragraphs obscures that. I suggest restoring it as one paragraph consisting of the bullet points with an intro.

"over Websockets" could be interpreted as "using Websockets"

```suggestion
For most applications, we recommend SignalR rather than raw WebSockets.

```

aspnetcore/fundamentals/websockets.md
```diff
@@ -14,17 +14,50 @@ uid: fundamentals/websockets
...
+SignalR provides:
+
+* Transport fallback for environments where WebSockets isn't available.
+* A basic remote procedure call app model.
+
+SignalR has no significant performance disadvantage compared to using raw WebSockets in most scenarios.
+
+WebSocket over HTTP/2 is supported for:
```
tdykstra, line 47:
```suggestion
WebSockets over HTTP/2 is supported for:
```

tdykstra (commented):
Looks good, a couple of comments and suggestions.

aspnetcore/fundamentals/websockets.md
BrennanConroy: What issue were you seeing? Did you use HTTPS?

tdykstra: Yes I used HTTPS, I got 405 Method Not Allowed

BrennanConroy: With the `app.Use` sample?

BrennanConroy: Because as you noted in the first paragraph there is no `HttpConnect` attribute (which is what would be needed for controllers).

To fix controllers in the sample we would need something like:
```csharp
public class HttpConnectAttribute : HttpMethodAttribute
{
    private static readonly IEnumerable<string> _supportedMethods = new[] { "CONNECT" };

    public HttpConnectAttribute()
        : base(_supportedMethods)
    {
    }
}
```

wadepickett:
@tdykstra, thanks!   For the app sample controller update I will need to create a new issue, if that is OK.  First round I was asked to get the updated info in these topics and link back from the What's new, but not update the topic fully for 7.0, that is round 2.  This topic does need a 7.0 app sample.

BrennanConroy:
The SignalR .NET client now also has support for this (starting in RC1). Although we can probably update that in a separate PR as this one is targeting preview7 changes.

tdykstra (commented):
I don't think we should publish a 7.0 version of this article pointing to a sample that doesn't work, so I would make this PR dependent on resolving the new issue #26812. And the 7.0 text in this PR that says you have to update your code is incomplete without guidance on how to update the code.

A thought on heading structure -- Consider restoring the original H2/H3 structure instead of moving everything to H3/H4. H4s are hard to differentiate from H3s in the rendered page and H3s don't show up in the internal TOC. The addition of **Websockets without SignalR** as an H2 heading seems unnecessary, since that's what the whole article is about.

Rick-Anderson:
> I don't think we should publish a 7.0 version of this article pointing to a sample that doesn't work

Correct, we don't publish samples that don't work. It's easy to comment out that section until it's fixed.

wadepickett:
Related to product issue https://github.com/dotnet/aspnetcore/issues/43501

wadepickett:
@BrennanConroy, thanks for the custom HttpConnect attribute class example!  Sorry to ask, but could you update this sample so it runs with the Controller change you suggested, so I can see how this is supposed to work?  I'm missing something.  I tried adding a HttpConnect attribute class, added a string for the attribute since the example is doing that  `[HttpConnect("/ws")]` and applied it to the existing controller.  In .NET 7.0 it would do http/2 but I could not get through the handshake, resulting in the error:

'wss://localhost:49613/ws' failed: Error during WebSocket handshake: Unexpected response code: 404

BrennanConroy:
The example was missing a ctor of `HttpConnectAttribute` for the route pattern, edited my comment above.

I just tried this and it works fine when I add `[HttpConnect("/ws")]` to the existing sample. Are you using the latest runtime?

wadepickett:
@BrennanConroy, thanks!  Yes, that all works.

I had added public HttpConnectAttribute(string template) earlier, but embarrassingly, I neglected to add "template" to the base.   So it was compiling, running, and failing a bit silently.

aspnetcore/fundamentals/websockets/samples/7.x/WebSocketsSample/Controllers/WebSocketController.cs
```diff
@@ -0,0 +1,47 @@
+using System.Net.WebSockets;
+using Microsoft.AspNetCore.Mvc;
+
+namespace WebSocketsSample.Controllers;
+
+// <snippet>
+public class WebSocketController : ControllerBase
+{
+    [HttpConnect("/ws")]
```
BrennanConroy, line 9: Have both `[HttpConnect("/ws")]` and `[HttpGet("/ws")]`

BrennanConroy: Why separate methods? The same code will go into both methods

aspnetcore/fundamentals/websockets.md
```diff
@@ -74,18 +102,30 @@ Object name: 'HttpResponseStream'.
...
-:::code language="csharp" source="websockets/samples/6.x/WebSocketsSample/Snippets/Program.cs" id="snippet_AcceptWebSocketAsyncBackgroundSocketProcessor":::
+:::code language="csharp" source="websockets/samples/7.x/WebSocketsSample/Snippets/Program.cs" id="snippet_AcceptWebSocketAsyncBackgroundSocketProcessor":::
 
 The WebSocket closed exception can also happen when returning too soon from an action method. When accepting a socket in an action method, wait for the code that uses the socket to complete before returning from the action method.
 
 Never use `Task.Wait`, `Task.Result`, or similar blocking calls to wait for the socket to complete, as that can cause serious threading issues. Always use `await`.
 
+## Add HTTP/2 WebSockets support for existing controllers
+
+.NET 7 introduced Websockets over HTTP/2 support for Kestrel, the SignalR JavaScript client, and SignalR with Blazor WebAssembly. HTTP/2 WebSockets use CONNECT requests rather than GET.  The previous controller action method example can be modified to use CONNECT by defining a `HTTPConnect` attribute class and applying that attribute to the controller method.
```
BrennanConroy, line 113: `HTTPConnect` -> `HttpConnect`

aspnetcore/fundamentals/websockets/samples/7.x/WebSocketsSample/Controllers/WebSocketController.cs
BrennanConroy: That's not my concern. What I'm saying is you need both `[HttpGet]` and `[HttpConnect]` on the same method otherwise you won't support both HTTP/1 and HTTP/2 websocket requests.

wadepickett: I get it.  Sorry, I misunderstood.  Thanks Brennan!

aspnetcore/fundamentals/websockets.md
```diff
@@ -74,18 +102,30 @@ Object name: 'HttpResponseStream'.
...
 
+## Add HTTP/2 WebSockets support for existing controllers
+
+.NET 7 introduced Websockets over HTTP/2 support for Kestrel, the SignalR JavaScript client, and SignalR with Blazor WebAssembly. HTTP/2 WebSockets use CONNECT requests rather than GET.  The previous controller action method example can be modified to use CONNECT by defining a `HttpConnect` attribute class and applying that attribute to the controller method.
+
+The following is an example custom attribute routing class named `HttpConnectAttribute`:
+
+[!code-csharp[](~/fundamentals/websockets/samples/7.x/WebSocketsSample/Controllers/HttpConnectAttribute.cs)]
+
+The custom `[HttpConnect]` attribute can be added to a controller action to enable Websocket support over HTTP/2:
```
BrennanConroy, line 119: I think there should be a sentence describing that it should be added **in addition** to the `HttpGet` attribute, otherwise you will only support HTTP/2 websockets and not HTTP/1 websockets.

aspnetcore/fundamentals/websockets.md
```diff
@@ -180,7 +220,9 @@ If using the WebSocket support in [socket.io](https://socket.io/) on [Node.js](h
 
 ## Sample app
 
-The [sample app](https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/fundamentals/websockets/samples) that accompanies this article is an echo app. It has a webpage that makes WebSocket connections, and the server resends any messages it receives back to the client. The sample app isn't configured to run from Visual Studio with IIS Express, so run the app in a command shell with [`dotnet run`](/dotnet/core/tools/dotnet-run) and navigate in a browser to `http://localhost:<port>`. The webpage shows the connection status:
+The [sample app](https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/fundamentals/websockets/samples) that accompanies this article is an echo app. It has a webpage that makes WebSocket connections, and the server resends any messages it receives back to the client. The sample app supports WebSockets over HTTP/2 when using a targeted framework of .NET 7 or later.
+
+The sample app isn't configured to run from Visual Studio with IIS Express, so run the app in a command shell with [`dotnet run`](/dotnet/core/tools/dotnet-run) and navigate in a browser to `http://localhost:<port>`. The webpage shows the connection status:
```
BrennanConroy, line 225: If you include the launchSettings.json file, you can remove the IIS Express section and only have the dotnet section so it is runnable from VS.

aspnetcore/fundamentals/websockets.md
```diff
@@ -136,7 +176,7 @@ If you're hosting your server on "https://server.com" and hosting your client on
 > [!NOTE]
 > The `Origin` header is controlled by the client and, like the `Referer` header, can be faked. Do **not** use these headers as an authentication mechanism.
 
-## IIS/IIS Express support
+### IIS/IIS Express support
 
 Windows Server 2012 or later and Windows 8 or later with IIS/IIS Express 8 or later has support for the WebSocket protocol.
```
BrennanConroy, line 181: But not for HTTP/2 WebSockets. cc @Tratcher

aspnetcore/fundamentals/websockets/samples/7.x/WebSocketsSample/Controllers/HttpConnectAttribute.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Microsoft.AspNetCore.Mvc.Routing;
+
+public class HttpConnectAttribute : HttpMethodAttribute
```
BrennanConroy, line 3: nit: `public sealed class`

Tratcher (approved):

aspnetcore/fundamentals/websockets.md
wadepickett: I did already add launchSettings.json, I'll also change the suggested text.  Good point, thanks!

aspnetcore/fundamentals/websockets.md
```diff
@@ -42,28 +70,28 @@ For some apps, [gRPC on .NET](xref:grpc/index) provides an alternative to WebSoc
...
 
-:::code language="csharp" source="websockets/samples/6.x/WebSocketsSample/Snippets/Program.cs" id="snippet_AcceptWebSocketAsync" highlight="7":::
+:::code language="csharp" source="websockets/samples/7.x/WebSocketsSample/Snippets/Program.cs" id="snippet_AcceptWebSocketAsync" highlight="7":::
 
 A WebSocket request could come in on any URL, but this sample code only accepts requests for `/ws`.
 
 A similar approach can be taken in a controller method:
 
-:::code language="csharp" source="websockets/samples/6.x/WebSocketsSample/Controllers/WebSocketController.cs" id="snippet":::
+:::code language="csharp" source="websockets/samples/7.x/WebSocketsSample/Controllers/WebSocketController.cs" id="snippet_Controller_Connect":::
```
BrennanConroy, line 94: I just noticed that we have two different controller examples, not sure why that's needed?

wadepickett: Are you referring to versioned examples?   6.x vs 7.x?

The 6.x sample is for the 6.x version of this topic with its own code snippet regions.   When the 6.x topic reaches retirement to an archive the whole thing will be archivable.

BrennanConroy: Eh, nvm. One is "you can use websockets with controllers", and the other is explaining HTTP/2 websockets and highlighting the specific line in the controller code.

BrennanConroy (approved):

PR #26809, opened 2022-08-23
Blazor WASM Options configuration guidance

### LINKED ISSUE #26439: Blazor WASM options configuration coverage
Based on the documentation, I can bind an options object using code like this:
`builder.Services.Configure<PositionOptions>(
    builder.Configuration.GetSection(PositionOptions.Position));`

Unfortunately, this doesn't work unless we include the `Microsoft.Extensions.Options.ConfigurationExtensions` package.  It would be great if the documentation mentioned this.

Thanks!

---
#### Document Details

* ID: 7367a651-2b27-0a57-e7d0-95686cc97f1c
* Version Independent ID: 8c769648-728d-72ce-ebf6-2a06e5076aa2
* Content: [ASP.NET Core Blazor configuration](https://docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/configuration?view=aspnetcore-6.0)
* Content Source: [aspnetcore/blazor/fundamentals/configuration.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/fundamentals/configuration.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #26439

Thanks @dougclutter! :rocket: ... I managed to reach this during this period before the first .NET 7 release candidate comes out 🏃 .

Pinging Hao for a remark on Options for Razor components. TL;DR from the issue, I can ask here ...

* I didn't test **_every_** Options scenario, but I know that basic config will work. General `IOptions<T>` and named options work as expected. From looking at a product unit issue, it looks like `IPostConfigureOptions` works as it should. 
* Configuration for Options (setting the initial values) can be provided by merely adding a package reference for `Microsoft.Extensions.Options.ConfigurationExtensions` to a WASM app.
* However, the framework doesn't wire up snapshot and monitoring features of Options. All that I can state thus far in the Blazor *Configuration* topic is that there's a lack of support for snapshot and monitored options features. I found an intriguing issue at https://github.com/dotnet/aspnetcore/issues/20092, where you said, "and then only modified if change notifications are setup," that I thought might make the Options recompute features work, but I didn't have any luck tinkering around with a WASM app getting something like Isaac's example code to work.

... so should we go with what's on this PR, or do you have alternative language for what I'm stating here? For example, is there a known way to make snapshot/monitored options recompute **_on the client_** if the `appsettings.json` file were to be changed there?

### DIFF
--- a/aspnetcore/blazor/fundamentals/configuration.md
+++ b/aspnetcore/blazor/fundamentals/configuration.md
@@ -235,6 +235,21 @@ Configuration files are cached for offline use. With [Progressive Web Applicatio
 
 For more information on how background updates are handled by PWAs, see <xref:blazor/progressive-web-app#background-updates>.
 
+## Options configuration
+
+[Options configuration](xref:fundamentals/configuration/options) for Blazor WebAssembly apps requires adding a package reference for the [`Microsoft.Extensions.Options.ConfigurationExtensions`](https://www.nuget.org/packages/Microsoft.Extensions.Options.ConfigurationExtensions) NuGet package.
+
+[!INCLUDE[](~/includes/package-reference.md)]
+
+Example:
+
+```csharp
+builder.Services.Configure<MyOptions>(
+    builder.Configuration.GetSection("MyOptions"));
+```
+
+Not all of the ASP.NET Core Options features are supported in Razor components. For example, <xref:Microsoft.Extensions.Options.IOptionsSnapshot%601> and <xref:Microsoft.Extensions.Options.IOptionsMonitor%601> configuration is supported, but recomputing option values for these interfaces aren't supported outside of reloading the app by either requesting the app in a new browser tab or selecting the browser's reload button. Merely calling [`StateHasChanged`](xref:blazor/components/lifecycle#state-changes-statehaschanged) doesn't update snapshot or monitored option values when the underlying configuration changes.
+
 :::moniker-end
 
 :::moniker range=">= aspnetcore-5.0 < aspnetcore-6.0"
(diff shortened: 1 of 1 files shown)


HaoK:
So the magic in how config reload works with options monitor is there's an overload when adding the config file:

```
                config.AddJsonFile("appsettings.json", optional: true, reloadOnChange: false)
                      .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true, reloadOnChange: false);
```

I am not sure if the stream configuration apis have this support for change detection

guardrex:
If you mean via (for example) `JsonStreamConfigurationProvider` ... it's something that the dev would need to compose on their own. We could say that "*it's possible*" if we knew. What we **_do know_** is that our Options patterns for snapshot and monitored Options values don't update in **components** OOB per our main doc set guidance ... not in WASM and not in Blazor Server apps. I know simply because I tested the scenarios to see what works and what doesn't.

I think we're safe to say what this PR says ... these features just aren't supported in components. It's not like it's an ask, per se. There are no docs issues calling for it; and from a search of PU repo issues, I don't see much beyond that annotations issue that I mentioned.

HaoK:
Okay that sounds reasonable to me, yeah only the initial state of options is supported for components (basically updates won't flow from the Snapshot/Monitor)

HaoK (approved):

guardrex:
Very well ... and btw ... I just discovered a further **_accuracy_** in the remark that I wrote.

I said (**_guessed_**) that snapshot and monitored option values don't update based on changes to the **_underlying configuration_**, not necessarily/specifically for a JSON file configuration provider-based change (i.e., `reloadOnChange`) ... or specifically for any other configuration established by any of the configuration providers. What I wrote is correct based on testing. Changing a configuration value on-the-fly in a component doesn't update the option value.

You can do a ...

```csharp
OptionsDelegate.Value.Option1 = "yack yack yack";
```

... but that defeats the whole purpose of snapshot/monitor updates based on underlying config changes and creates a delta between what the snapshot/monitor holds and what configuration holds ... setting that doesn't change the underlying configuration value for `Option1`.

Yeah ... I think what we have here on the PR is fine until/unless there's a feature expansion later for Options.

PR #26817, opened 2022-08-24
Add JS interop example

### LINKED ISSUE #26798: Add a new JS interop scenario
Per https://github.com/dotnet/aspnetcore/issues/43367#issuecomment-1220496956.

**UPDATE (8/23)**: Need to hold for a sec. I think we already have the requested coverage at ...

**Component instance .NET method helper class**
https://docs.microsoft.com/aspnet/core/blazor/javascript-interoperability/call-dotnet-from-javascript#component-instance-net-method-helper-class

I'm waiting to hear back if that approach will work for the dev's needs ... and thus if I can then close this out.

**UPDATE (8/23)**: Looks like there would be an alternative approach to the one that we show ...

> I think users fail to connect the dots on how to use JS interop and element reference to achieve the scenario.

I hope to be able to work up the existing list item example using JS interop with element reference so that devs can see SxS how they compose.

Thus far, I've hacked up a concept (akin to the current example that uses a helper class) as ...

```html
<script>
  window.setup = (element, dotNetHelper) => {
    element.dotNetHelper = dotNetHelper;
  }

  window.interopCall = async (element) => {
    await element.dotNetHelper.invokeMethodAsync('UpdateUI');
  }
</script>
```

`Shared/ListItem2.razor`:

```razor
@inject IJSRuntime JS

<li>
    <span @ref="spanRef" onclick="interopCall(this)">Select one of these list item buttons.</span>
    <span style="display:@display">Not Updated Yet!</span>
</li>

@code {
    private ElementReference spanRef;
    private string display = "inline-block";
    private DotNetObjectReference<ListItem2>? objRef;

    protected override async Task OnAfterRenderAsync(bool firstRender)
    {
        if (firstRender)
        {
            objRef = DotNetObjectReference.Create(this);
            await JS.InvokeVoidAsync("setup", spanRef, objRef);
        }
    }

    [JSInvokable]
    public void UpdateUI()
    {
        display = "none";
        StateHasChanged();
    }

    public void Dispose() => objRef?.Dispose();
}
```

`Pages/Example.razor`:

```razor
@page "/example"

<h1>Call .NET Example 2</h1>

<ul>
    <ListItem2 />
    <ListItem2 />
    <ListItem2 />
    <ListItem2 />
</ul>
```

---
#### Document Details

* ID: 2964d558-cd6d-7421-f7b9-cb939792c2cf
* Version Independent ID: a3517e56-69fc-1ec7-b55a-dbb54d68a807
* Content: [Call .NET methods from JavaScript functions in ASP.NET Core Blazor](https://docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-dotnet-from-javascript?view=aspnetcore-6.0#class-instance-examples)
* Content Source: [aspnetcore/blazor/javascript-interoperability/call-dotnet-from-javascript.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/javascript-interoperability/call-dotnet-from-javascript.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #26798

@mayur-ekbote ... Would you like to take a look at this?

See if this lays out the approach and would've helped with the editable `<div>`s scenario. I never did find out why Javier said "JSObjectReference." I don't see how `IJSObjectReference` is relevant. Perhaps, he meant to say `DotNetObjectReference`. The code for the `ListItem2` and `CallDotNetExample7` components can be reached via the links below. They're in the `dotnet/blazor-samples` repo, and the code is pulled into the article from there.

I made the example almost exactly like the preceding list item example, so readers can see these approaches in a SxS format. Because I'm using another list item component named "ListItem," you'll see that I'm renaming the one in the *Component instance .NET method helper class* section to `ListItem1` so that I can have the new example named `ListItem2`.

[Internal Review Topic (links to section)](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-dotnet-from-javascript?view=aspnetcore-7.0&branch=pr-en-us-26817#component-instance-net-method-called-from-dotnetobjectreference-assigned-to-element-property)

* [`ListItem2` at `dotnet/blazor-samples`](https://github.com/dotnet/blazor-samples/blob/main/7.0/BlazorSample_Server/Shared/call-dotnet-from-js/ListItem2.razor)
* [`CallDotNetExample7` at `dotnet/blazor-samples`](https://github.com/dotnet/blazor-samples/blob/main/7.0/BlazorSample_Server/Pages/call-dotnet-from-js/CallDotNetExample7.razor)

### DIFF
--- a/aspnetcore/blazor/javascript-interoperability/call-dotnet-from-javascript.md
+++ b/aspnetcore/blazor/javascript-interoperability/call-dotnet-from-javascript.md
@@ -1941,30 +1941,31 @@ The following image shows the rendered `CallDotNetExample6` parent component aft
 
 ![Rendered 'CallDotNetExample6' component example](~/blazor/javascript-interoperability/call-dotnet-from-javascript/_static/component-example-6.png)
 
-## Component instance .NET method called from `DotNetObjectReference` assigned to element property
+## Component instance .NET method called from `DotNetObjectReference` assigned to an element property
 
-In this scearnio:
+The assignment of a <xref:Microsoft.JSInterop.DotNetObjectReference> to a property of an HTML element permits calling .NET methods on a component instance:
 
 * An [element reference](xref:blazor/js-interop/call-javascript-from-dotnet#capture-references-to-elements) is captured (<xref:Microsoft.AspNetCore.Components.ElementReference>).
-* Inside the [`OnAfterRender` or `OnAfterRenderAsync` method](xref:blazor/components/lifecycle#after-component-render-onafterrenderasync), a JavaScript (JS) function is invoked with the element reference and the component instance as a <xref:Microsoft.JSInterop.DotNetObjectReference>. The JS function attaches the <xref:Microsoft.JSInterop.DotNetObjectReference> to the element in a property (for example, `element.dotNetHelper = dotNetHelper;`, where `element` is the element reference and `dotNetHelper` is the component instance).
+* In the component's [`OnAfterRender{Async}` method](xref:blazor/components/lifecycle#after-component-render-onafterrenderasync), a JavaScript (JS) function is invoked with the element reference and the component instance as a <xref:Microsoft.JSInterop.DotNetObjectReference>. The JS function attaches the <xref:Microsoft.JSInterop.DotNetObjectReference> to the element in a property.
 * When an element event is invoked in JS (for example, `onclick`), the element's attached <xref:Microsoft.JSInterop.DotNetObjectReference> is used to call a .NET method.
 
-Similar to the use case described in the [Component instance .NET method helper class](#component-instance-net-method-helper-class) section, this approach is useful in the following scenarios:
+Similar to the approach described in the [Component instance .NET method helper class](#component-instance-net-method-helper-class) section, this approach is useful in the following scenarios:
 
 * When several components of the same type are rendered on the same page.
 * In Blazor Server apps with multiple users concurrently using the same component.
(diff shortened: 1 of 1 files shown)


guardrex:
@mayur-ekbote ... I'm going to go ahead with this. I'd like to merge general Blazor PRs *quickly* 🏃 at this time, as I'm regularly placing new .NET 7 content and desire to avoid dreaded 💥 **_merge conflicts_** 💥.

I didn't hear back from Javier, but I think when he wrote "JSObjectReference" that he actually meant "DotNetObjectReference," which is what made sense to me from the discussion in the PU issue.

If you have feedback ... any problems that you detect with the new section ... please place them in a comment here. I can patch the new content easily enough at any time. Thanks again for calling out the need for this scenario.

mayur-ekbote:
Sorry I couldn't access some bits (internal to Microsoft I guess). I think it is as per what Javier had suggested. I too think that he meant `DotNetObjectReference` when he wrote `JSObjectReference`. I have tried the solution and it works. So as you suggested we can relook at it if unforeseen issues come up.

guardrex:
Thanks @mayur-ekbote ... sounds good. I'll get it posted back to the earlier versions as soon as I can 🏃‍♂️🏃‍♂️🏃‍♂️🏃‍♂️🏃‍♂️😅.

ng294:
Don't me e-mail

PR #26825, opened 2022-08-25
Fix capitalization and hyphenation

vanillajonathan:


guardrex (approved):
Thanks @vanillajonathan! 🚀 ... I added a NIT update for a folder and a file to code fence them. At some point a few years ago, we figured out that fencing files and folders is best for localization (i.e., they're never localized by the loc engine that runs on the docs).

PR #26863, opened 2022-08-29
JS interop with prerendering updates

### LINKED ISSUE #26862: "Detect when a Blazor Server app is prerendering" is not very clear
The paragraph "Detect when a Blazor Server app is prerendering" contains examples and a various warnings and notes about the examples but it does not give a clear answer on how to actually detect _prerendering_.

It would be helpful to add a bit more _prerendering_ specific text to this sentence:

> To delay JavaScript interop calls until a point where such calls are guaranteed to work, override the [OnAfterRender{Async} lifecycle event](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?view=aspnetcore-6.0#after-component-render-onafterrenderasync). This event is only called after the app is fully rendered.

For example, add this line from the component lifecycle page:

[OnAfterRender](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.components.componentbase.onafterrender) and [OnAfterRenderAsync](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.components.componentbase.onafterrenderasync) aren't called during the prerendering process on the server. The methods are called when the component is rendered interactively after prerendering.

Also the paragraph title could be improved

1. The paragraph title: remove the word "Server" because the topic applies to both Blazor Server _and_ Blazor WebAssembly
2. The actual _detction_ of prerendering is not really the subject here. A better match would be "Delay JavaScript interop calls when prerendering"

---
#### Document Details

* Content: [Call JavaScript functions from .NET methods in ASP.NET Core Blazor](https://docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-javascript-from-dotnet?view=aspnetcore-5.0#feedback)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #26862

Thanks @JefSchraag! :rocket: ... Those are great suggestions. A few notes on the update here ...

* The section can be renamed to merely "Prerendering." This content comes from an INCLUDE file to appear in two spots, the *Lifecycle* topic and the *Call JS from .NET* topic. I think that will work in both spots. I ran through the docs and updated the references everywhere for this section heading change.
* It's ok to use the word "server" (lowercase) because that's where prerendering happens in either a Blazor Server app or hosted Blazor WebAssembly (solution). "Server" isn't used in the docs to mean "Blazor Server." I do use it in spots with special formatting [**`Server`** project (or app)] when referring to that project/app in a hosted WASM solution.
* When I have the two methods separated by a conjunction (i.e., "and"), I go plural on the verb. When I combine them (i.e., `OnAfterRender{Async}`), I use a singular verb. I think this will work ok for us. I think I will update use of a slash (`OnAfterRender`/`OnAfterRenderAsync`) to that format with the braces on a follow-up commit here in a sec. **UPDATE**: _Done!_ 👍 

If you want to look at the PR before I merge it, please look quickly 👀🏃. I don't want to leave PRs lying around due to our rapid rollout of .NET 7 docs for the upcoming release ... 💥 **_merge conflicts ensue!_** 💥😈😆 

**UPDATE**: **_mmm_** 🤔 ... Actually, I think I do want to use a different section heading in the *Lifecycle* topic to pair better with the existing *Stateful reconnection after prerendering* section. Hold on for another commit to fix it. ..... Ok ... **_Lick'd IT!_** 🐄👅 ... let's see how that looks on the diff.

### DIFF
--- a/aspnetcore/blazor/components/lifecycle.md
+++ b/aspnetcore/blazor/components/lifecycle.md
@@ -101,7 +101,7 @@ Blazor apps that prerender their content on the server call <xref:Microsoft.AspN
 
 To prevent developer code in <xref:Microsoft.AspNetCore.Components.ComponentBase.OnInitializedAsync%2A> from running twice when prerendering, see the [Stateful reconnection after prerendering](#stateful-reconnection-after-prerendering) section. Although the content in the section focuses on Blazor Server and stateful SignalR *reconnection*, the scenario for prerendering in hosted Blazor WebAssembly apps (<xref:Microsoft.AspNetCore.Mvc.Rendering.RenderMode.WebAssemblyPrerendered>) involves similar conditions and approaches to prevent executing developer code twice. To preserve state during the execution of initialization code while prerendering, see <xref:blazor/components/prerendering-and-integration#persist-prerendered-state>.
 
-While a Blazor app is prerendering, certain actions, such as calling into JavaScript (JS interop), aren't possible. Components may need to render differently when prerendered. For more information, see the [Prerendering](#prerendering) section.
+While a Blazor app is prerendering, certain actions, such as calling into JavaScript (JS interop), aren't possible. Components may need to render differently when prerendered. For more information, see the [Prerendering](#prerendering-with-javascript-interop) section.
 
 If event handlers are provided in developer code, unhook them on disposal. For more information, see the [Component disposal with `IDisposable` `IAsyncDisposable`](#component-disposal-with-idisposable-and-iasyncdisposable) section.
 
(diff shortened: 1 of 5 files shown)


JefSchraag:
A nice improvement that makes this complex topic better to grasp. Thanks for the fast update.

JefSchraag (approved):

guardrex:
Thanks @JefSchraag ... and sorry about all of the churn pings. *I should've branched for this!* 😆

PR #26866, opened 2022-08-29
gRPC multithreading example

JamesNK:
gRPC multithreading example

### DIFF
--- a/aspnetcore/grpc/services.md
+++ b/aspnetcore/grpc/services.md
@@ -328,7 +328,7 @@ A gRPC call ends on the server once the gRPC method exits. The following argumen
 
 If a gRPC method starts background tasks that use these types, it must complete the tasks before the gRPC method exits. Continuing to use the context, stream reader, or stream writer after the gRPC method exists causes errors and unpredictable behavior.
 
-The following method could write to the response stream after the method has finished:
+In the following example, the server streaming method could write to the response stream after the call has finished:
 
 ```csharp
 public override async Task StreamingFromServer(ExampleRequest request,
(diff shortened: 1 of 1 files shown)


aspnetcore/grpc/services.md
```diff
@@ -328,6 +328,46 @@ A gRPC call ends on the server once the gRPC method exits. The following argumen
 
 If a gRPC method starts background tasks that use these types, it must complete the tasks before the gRPC method exits. Continuing to use the context, stream reader, or stream writer after the gRPC method exists causes errors and unpredictable behavior.
 
+In the following example, the server streaming method could write to the response stream after the call has finished:
```
wadepickett, line 331:
```suggestion
In the following example, the server streaming method writes to the response stream after the call has finished:
```

aspnetcore/grpc/services.md
```diff
@@ -328,6 +328,46 @@ A gRPC call ends on the server once the gRPC method exits. The following argumen
...
+            await responseStream.WriteAsync(new ExampleResponse());
+            await Task.Delay(TimeSpan.FromSeconds(1));
+        }
+    });
+
+    await PerformLongRunningWorkAsync();
+}
+```
+
+The solution is to await the write task before exiting the method:
```
wadepickett, line 350:
```suggestion
For the previous example, the solution is to await the write task before exiting the method:
```

wadepickett (approved):
Looks good.  I made a suggestion for a very minor item that you could take or leave.  :)

aspnetcore/grpc/services.md
JamesNK: I'll leave it as "could write" because it's not certain what will happen.

PR #26871, opened 2022-08-30
feat: use variable value directly

yasirmx:
### Changes

- Used the value of the variable directly


guardrex (approved):
Thanks @yasirmx! :rocket:

PR #26874, opened 2022-08-30
Improve dB factory/context section

### LINKED ISSUE #26873: Show example that uses DbFactory
[Update by guardrex to change the metadata.]

Could you please add an example where Identity uses a `DbFactory` instead of the `DbContext`? Using a factory is the recommended way in EF Core >5.0, see [link1](https://docs.microsoft.com/en-us/ef/core/what-is-new/ef-core-5.0/whatsnew#dbcontextfactory), [link2](https://docs.microsoft.com/en-us/ef/core/dbcontext-configuration/#using-a-dbcontext-factory-eg-for-blazor). The example on this page (and the template for new Blazor server side projects) uses
```
Services.AddEntityFrameworkStores<ApplicationDbContext>()
```
to pass the `DbContext` to Identity. There is a simple solution described [here](https://stackoverflow.com/questions/64169025/using-identity-with-adddbcontextfactory-in-blazor), which passes the `DbContext` obtained from the `DbFactory`:
```
services.AddScoped<ApplicationDbContext>(p => p.GetRequiredService<ApplicationDbContextFactory<ApplicationDbContext>>().CreateDbContext());
```

**But**, as described in detail [here](https://stackoverflow.com/questions/73498604/dependency-injected-usermanager-causes-concurrency-issue-how-to-make-usermanage), the `UserManager` obtained via dependency injection is the same in every Blazor component, which can lead to a concurrency issue because asynchronous components typically run on differnt threads.

I think that Identity should also accept a `DbContextFactory` to enable something like `services.AddScoped<ApplicationDbContextFactory>()` . So it might need more updates than just the documentaion...

---
#### Document Details

* ID: 4a101e4d-85a2-8013-0f62-40e13e45b3c6
* Version Independent ID: 66a0a920-ebd8-9b1d-c8c5-953ebb766de0
* Content: [ASP.NET Core Blazor Server with Entity Framework Core (EF Core)](https://docs.microsoft.com/en-us/aspnet/core/blazor/blazor-server-ef-core?view=aspnetcore-6.0)
* Content Source: [aspnetcore/blazor/blazor-server-ef-core.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/blazor-server-ef-core.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **jeliknes**

guardrex:
Fixes #26873

Thanks @KurtP20! 🚀 ... I think this section was the content that you were seeking, not per the guidance in the main doc set. However, I think we can make an improvement here with this PR that will help clarify the example a bit further. Of course, the whole sample app is quite great 🎉 ... built by a guru dB engineer and product manager 🙏, Jeremy Likness.

If you plan to look this over, please do so quickly 🏃, as I'm trying to avoid 😈 **_merge conflicts_** 😈 under a rapid .NET 7 content rollout.

### DIFF
--- a/aspnetcore/blazor/blazor-server-ef-core.md
+++ b/aspnetcore/blazor/blazor-server-ef-core.md
@@ -246,8 +246,6 @@ A `DbContext` is created using the factory (`DbFactory`) to delete a contact in 
 > [!NOTE]
 > `Filters` is an injected `IContactFilters`, and `Wrapper` is a [component reference](xref:blazor/components/index#capture-references-to-components) to the `GridWrapper` component. See the `Index` component (`Pages/Index.razor`) in the [sample app](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/samples/5.0/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/Index.razor).
 
-New <xref:Microsoft.EntityFrameworkCore.DbContext> instances can be created with a factory that allows you to configure the connection string per `DbContext`, such as when you use [ASP.NET Core's Identity model](xref:security/authentication/customize_identity_model). For more information, see [Multi-tenancy (EF Core documentation)](/ef/core/miscellaneous/multitenancy).
-
 ### Scope to the component lifetime
 
 You may wish to create a <xref:Microsoft.EntityFrameworkCore.DbContext> that exists for the lifetime of a component. This allows you to use it as a [unit of work](https://martinfowler.com/eaaCatalog/unitOfWork.html) and take advantage of built-in features, such as change tracking and concurrency resolution.
@@ -358,8 +356,6 @@ The following recommendations are designed to provide a consistent approach to u
 
 * For longer-lived operations that take advantage of EF Core's [change tracking](/ef/core/querying/tracking) or [concurrency control](/ef/core/saving/concurrency), [scope the context to the lifetime of the component](#scope-to-the-component-lifetime).
 
-<!-- HOLD: See information in the UE tracking issue at: https://github.com/dotnet/AspNetCore.Docs/issues/24615
-
 ### New `DbContext` instances
 
 The fastest way to create a new <xref:Microsoft.EntityFrameworkCore.DbContext> instance is by using `new` to create a new instance. However, there are scenarios that require resolving additional dependencies:
@@ -394,10 +390,6 @@ A `DbContext` is created using the factory (`DbFactory`) to delete a contact in 
 
 > [!NOTE]
 > `Filters` is an injected `IContactFilters`, and `Wrapper` is a [component reference](xref:blazor/components/index#capture-references-to-components) to the `GridWrapper` component. See the `Index` component (`Pages/Index.razor`) in the [sample app](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/samples/3.1/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/Index.razor).
-
-New <xref:Microsoft.EntityFrameworkCore.DbContext> instances can be created with a factory that allows you to configure the connection string per `DbContext`, such as when you use [ASP.NET Core's Identity model])(xref:security/authentication/customize_identity_model). For more information, see [Multi-tenancy (EF Core documentation)](/ef/core/miscellaneous/multitenancy).
-
--->
 
 ### Scope to the component lifetime
 
@@ -539,8 +531,6 @@ A `DbContext` is created using the factory (`DbFactory`) to delete a contact in 
 
 > [!NOTE]
 > `Filters` is an injected `IContactFilters`, and `Wrapper` is a [component reference](xref:blazor/components/index#capture-references-to-components) to the `GridWrapper` component. See the `Index` component (`Pages/Index.razor`) in the [sample app](https://github.com/dotnet/blazor-samples/blob/main/6.0/BlazorServerEFCoreSample/Pages/Index.razor).
-
-New <xref:Microsoft.EntityFrameworkCore.DbContext> instances can be created with a factory that allows you to configure the connection string per `DbContext`, such as when you use [ASP.NET Core's Identity model](xref:security/authentication/customize_identity_model). For more information, see [Multi-tenancy (EF Core documentation)](/ef/core/miscellaneous/multitenancy).
 
 ### Scope to the component lifetime
 
--- a/aspnetcore/blazor/blazor-server-ef-core.md
+++ b/aspnetcore/blazor/blazor-server-ef-core.md
@@ -244,7 +244,7 @@ A `DbContext` is created using the factory (`DbFactory`) to delete a contact in 
 :::code language="csharp" source="~/../blazor-samples/5.0/BlazorServerEFCoreSample/BlazorServerDbContextExample/Pages/Index.razor" id="snippet1":::
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


KurtP20:
Thanks for the update. The changes and the link to the example are nice 👍 . **But:** I still don't see how the injected `UserManager` uses a new `DbContext` each time it is injected as described in #26873 . The sample does not use `UserManager` at all. Could you please explain how I can inject a `DbFactory` to Identity/UserManager?

guardrex:
> I find it very confusing with Razor pages, Razor components that are called Blazor components but use Razor syntax and Razor pages that actually are Razor components

That's unfortunate, and I pitched to management that MS never use the term "Blazor component," but that ship had already sailed. A "Blazor component" is the unofficial name now for "Razor component" (the official name), so there are two names for exactly the same thing. Razor components and pages of Razor Pages both use Razor syntax, but that's where the similarity ends. The framework processing for pages and components is different.

> injected UserManager uses a new DbContext

It's not supported by MS given what they've told me. **_However_**, Javier did remark at one point ...

> Other parts of identity like UserManager work because they don't rely on the HTTP Cookie in their implementation.

Still tho, I was told to tell devs **_not to use it in components_**. More info is in the cross-linked scaffolding doc at ...

https://docs.microsoft.com/aspnet/core/security/authentication/scaffold-identity#custom-identity-components

... including a cross-link to a PU issue where Javier goes into great detail on this subject ...

* https://github.com/dotnet/aspnetcore/issues/34095#issuecomment-874012924
* https://github.com/dotnet/aspnetcore/issues/34095#issuecomment-874061914

`SignInManager`/`UserManager` work normally in a controller, so devs can have the component to call server-side to process Identity actions with them. We have an example, but it's hosted WASM at this point. I might make a Blazor Server version of it at some point, but the controller side of it would be very similar or exactly the same ...

https://docs.microsoft.com/aspnet/core/blazor/security/webassembly/hosted-with-azure-active-directory#usermanager-and-signinmanager

WRT ...

> use new DbContext but it also recommends to use a factory instead

Jeremy calls out a couple of example cases ...

> ... there are several scenarios that may require resolving additional dependencies. For example, you may wish to use `DbContextOptions` to configure the context.

... and later ... at the bottom of the section ...

> ... created with a factory that allows you to configure the connection string per `DbContext` ...

I need to add a commit to the PR that combines those at the top. I don't have any further examples to mention tho, nor do I have the code. I think both of those could receive a code example. I'll consider that this morning. What I'd like to do is find those exact factory use cases over in the EFCore docs and cross-link to them. That's really best for us here ... these are after all Blazor docs ... not EF Core docs 😄.

guardrex:
Ok ... looks good from here.

I even found a couple of broken sample links 🙈 due to the movement of the snippet samples going from this repo over to the `dotnet/blazor-samples` repo. I checked the **_whole repo_** for that problem. Fortunately, these seem to be the only ones that didn't get updated when the samples moved. You never know 100% for sure tho. These types of things are 😈 **_Sneaky Devils&trade;!_** 😈😆.

I also decided that these 3rd level headings really should go to H2 so that the sections will show up in the topic's ToC sidebar.

I'm going to merge now. I'll take more feedback on this subject as we go. The cross-links to the scaffolding topic have the further remarks on `SignInManager`/`UserManager` **_AND_** the cross-links to the PU issues where Javier made those detailed remarks. In the future, I might add more info from those PU discussions; but for now, I'll go with the current cross-links and remarks and take further feedback.

Thanks again for calling this subject out for further work. *A lot was fixed up here!* 🎉

KurtP20:
Okay, I need to ponder over this a bit longer. My initial thoughts are that it is not a good idea to say that `UserManager` cannot be used in Razor componets *at all*, if - as found out by trial and error - it *is* possible (maybe with certain restrictions). 

- [ ] Wouldn't it be better to mention the restrictions instead of stating something which is not entirely true?

Maybe we have a different understanding of how `UserManager` can be used, so please let me explain how I use it: In some Razor pages, the `NavMenu`, in `LoginDisplay.razor` and in some service classes, I use `UserManager` to obtain the Identity user from a `ClaimsPrincipalUser`, which I get from `AuthenticationStateProvider` like so:
```
var authenticationState = await AuthenticationStateProvider.GetAuthenticationStateAsync();
var cpUser = authenticationState.User;
var myUser = await UserManager.GetUserAsync(cpUser);
```
To my Razor components I pass this Identity user and use 
```
await UserManager.IsInRoleAsync(myUser, "teacher")
```
to find the roles the user has. Except for the concurrency issue, this work fine.

I might be able to get away by not only passing the Identity user to my components but also the `ClaimsPrincipleUser` user to eliminate the use of `UserManager` from the Razor components. 

- [ ] But to me the question is more why Identity uses `DbContext` and not `DbContextFactory` if a factory is defined. Identity could then use a fresh `DbContext` for each `UserManager` for dependency injection. 

If I am not mistaken, there could also be a concurrency issue when differnet services use `UserManager` via dependency injection, so the whole issue of Identity not using `DbFactory` might be unrelated to Razor components. And after all, using a factory is IMHO the recommended way in EF Core >5.0.

guardrex:
> Wouldn't it be better to mention the restrictions instead of stating something which is not entirely true?

IIRC, this is what they asked me to put into the docs. I'd need to research it out and see for sure. It would be best perhaps to ask specifically about `UserManager` (outside of `SignInManager`) given Javier's remark that I cited ...

> Other parts of identity like UserManager work because they don't rely on the HTTP Cookie in their implementation.

... at ...

https://github.com/dotnet/aspnetcore/issues/34095#issuecomment-874061914

You would let them know about what we currently have in the Blazor security guidance, the concurrency problem that you're facing, and then ask them about using it safely in components. If you do set up that issue for them to ask at ...

Please add ...

```
cc: @guardrex https://github.com/dotnet/AspNetCore.Docs/pull/26874
```

... to the bottom of your opening comment so that I follow along. I'll open a new docs issue if they want to make a change to the recommendations or call out a scenario whereby `UserManager` would be safe to use in components. I think they're going to say not to use it, but I'm not sure. Javier's remark on that over there was a bit unclear to me in relation to earlier guidance that I had from the team ... and it might have been Pranav or Ryan a long time ago. Whatever the case, probably either Tanay or Javier will provide a definitive remark on `UserManager`.

> But to me the question is more why Identity uses DbContext and not DbContextFactory if a factory is defined. Identity could then use a fresh DbContext for each UserManager for dependency injection.

I suggest that since you'd be opening an issue over there for `UserManager` in general that you add the question on this as well. If I end up opening a new docs issue later, I might be adding their response. It depends on how ... *weedy* 😄 ... their answer is. We don't really get into all of the ✨ *deep inner magic* ✨ on framework implementations. Also, I'd rather not get into covering those aspects that are in the main ASP.NET Core Identity docs or the MS Identity docs ... those are the main docs sets for Identity coverage that apply to the whole framework. I'd like to keep this node focused on the Blazor aspects (deltas if they exist) and cross-link content in the main doc sets as much as possible.

So ... in summary ... I think it would be ok if you open an issue for the product unit on these questions. I'll be 👂 on it to see what else we need to do over here.

KurtP20:
Thanks for your guidance 🥇 . I have opend #43687

guardrex:
Great PU issue. That should lead to some good feedback from Javier or Tanay. I'll keep an 👂 open and an 👁️ on it.

guardrex:
Yeah, he's not excited about it. 😄 I'm going to leave the documentation as it is. If devs want to try taking his advice on resolving a scope for it, that's fine. I don't think it's something that we want the docs to tell people to take a chance on. The second point is moot because it isn't supported in the first place. We kind'a mostly try to steer clear of framework design discussion in the docs.

PR #26892, opened 2022-08-31
Fix logic error in the explanation

guneyizol:
If `Model` or `LicenseId` is null, the expression evaluates to true, not false.

### DIFF
--- a/aspnetcore/mvc/views/tag-helpers/intro.md
+++ b/aspnetcore/mvc/views/tag-helpers/intro.md
@@ -152,7 +152,7 @@ The preceding code can be written as:
        disabled="@(Model?.LicenseId == null)" />
 ```
 
-Normally, the `@` operator inserts a textual representation of an expression into the rendered HTML markup. However, when an expression evaluates to logical `false`, the framework removes the attribute instead. In the preceding example, the `disabled` attribute is removed if neither `Model` nor `LicenseId` is `null`.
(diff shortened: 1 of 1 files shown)


aspnetcore/mvc/views/tag-helpers/intro.md
```diff
@@ -152,7 +152,7 @@ The preceding code can be written as:
        disabled="@(Model?.LicenseId == null)" />
 ```
 
-Normally, the `@` operator inserts a textual representation of an expression into the rendered HTML markup. However, when an expression evaluates to logical `false`, the framework removes the attribute instead. In the preceding example, the `disabled` attribute is removed if `Model` or `LicenseId` is `null`.
+Normally, the `@` operator inserts a textual representation of an expression into the rendered HTML markup. However, when an expression evaluates to logical `false`, the framework removes the attribute instead. In the preceding example, the `disabled` attribute is removed if neither `Model` nor `LicenseId` is `null`.
```
Rick-Anderson, line 155:
```suggestion
Normally, the `@` operator inserts a textual representation of an expression into the rendered HTML markup. However, when an expression evaluates to logical `false`, the framework removes the attribute instead. In the preceding example, the `disabled` attribute is set to `true` if either `Model` or `LicenseId` is `null`.
```

guneyizol: IMHO, the false case is more relevant to the preceding discussion.

Rick-Anderson: > IMHO, the false case is more relevant to the preceding discussion.

It's not about the attributes, it's about working with `bool`  with `@`.

guneyizol: It's about both but you are the boss 😃 thanks for the feedback

Issue #26540, opened 2022-07-25
Minimal API parameter binding for argument lists


Rick-Anderson:
[Minimal API parameter binding for argument lists](https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-dotnet-7-preview-5/#minimal-api-parameter-binding-for-argument-lists)

see https://github.com/dotnet/aspnetcore/pull/41325

Put code in https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/minimal-apis/arg-lists/readme.txt and delete readme.txt

Use
```
var connection = new SqliteConnection("DataSource=:memory:");
connection.Open();
```

`@brunolins16` to review

Rick-Anderson:
@fiyazbinhasan  @martincostello  anyone interested in writing this sample code?

martincostello:
Sure, I can put something together for this.

martincostello:
@Rick-Anderson I'll put together a specific example a bit later today or tomorrow, but I'd already written a sample app for an internal talk at my work last month, so I've just put the whole file for all the new features I did examples for as far as preview 5/6 [here](https://gist.github.com/martincostello/742dfe99eaa3d872bb90f5e3a295bba2). The specific one for `[AsParameters]` is [here](https://github.je-labs.com/martin-costello/aspnet-core-7-samples/blob/63b15e3c12418e52cf8520003c1a1e2d6b983f98/src/TodoApp/SampleEndpoints.cs#L200-L208).

Rick-Anderson:
@martincostello  can you put your sample in https://github.com/dotnet/AspNetCore.Docs.Samples/tree/main/fundamentals/minimal-apis/samples ?

Rick-Anderson:
@martincostello  do you have an ETA on when you'll be able to write this sample? If possible, use a variation of [todoitems](https://docs.microsoft.com/en-us/aspnet/core/tutorials/min-web-api)

martincostello:
Sorry for the delay with this - been quite busy with other things at the moment to get around to this. I'll aim to get you something for that other repo by the end of next week.

martincostello:
https://github.com/dotnet/AspNetCore.Docs.Samples/pull/26

PR #26897, opened 2022-09-01
Min API param binding with record /1

Rick-Anderson:
Contributes to #26540
[Internal review URL](https://review.docs.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis?view=aspnetcore-7.0&branch=pr-en-us-26897#parameter-binding-for-argument-lists)

### DIFF
--- a/aspnetcore/fundamentals/minimal-apis.md
+++ b/aspnetcore/fundamentals/minimal-apis.md
@@ -1099,15 +1099,15 @@ The <xref:Microsoft.AspNetCore.Http.AsParametersAttribute> enables parameter bin
 
 Consider the following `GET` endpoint:
 
-:::code language="csharp" source="~/../AspNetCore.Docs.Samples/fundamentals/minimal-apis/samples/arg-lists/Program.cs" id="snippet_id":::
+:::code language="csharp" source="~/../AspNetCore.Docs.Samples/fundamentals/minimal-apis/samples/arg-lists/Program.cs" id="snippet_id" highlight="2":::
 
 The following [`record`](/dotnet/csharp/language-reference/builtin-types/record) can be used to replace the two parameters:
 
-:::code language="csharp" source="~/../AspNetCore.Docs.Samples/fundamentals/minimal-apis/samples/arg-lists/Models/TodoDb.cs" id="snippet_1":::
+:::code language="csharp" source="~/../AspNetCore.Docs.Samples/fundamentals/minimal-apis/samples/arg-lists/Models/TodoDb.cs" id="snippet_1" :::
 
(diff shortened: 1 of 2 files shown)


Rick-Anderson:
## Public review  of the doc build:
- download the [HTML files](https://www.dropbox.com/sh/yy994551cje0kap/AAB0tRnm7gze2kYrSYPSFHIda?dl=0)
- unzip
- view HTML file in a browser.
- For links outside the doc, remove the `review.` from URL.
@martincostello  please review

martincostello:
The content here looks fine to me. I wonder if it's worth mentioning in the prose that the same attributes that apply to parameters (`[FromRoute]`, `[FromBody]`, `[FromServices]`, etc.) also work with the type used for `[AsParameters]` and cross-reference to wherever else in the docs that was originally explained for .NET 6 and parameter binding?

Rick-Anderson:
@martincostello I think we need to add `[AsParameters]` to [Binding Precedence](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis?view=aspnetcore-6.0#binding-precedence)

Do you know where it goes  on that list?

martincostello:
[Looking at the code](https://github.com/dotnet/aspnetcore/blob/7d1e7b192298f4c269de1c1afcc80ad5c5abb3dd/src/Http/Http.Extensions/src/RequestDelegateFactory.cs#L711-L720), it looks like it goes after `[FromServices]`.

brunolins16:
@Rick-Anderson since `struct` are the most optimized scenario, what do you think about show that usage in the examples instead? Maybe mentioning that a `class` also works.

brunolins16:
> The content here looks fine to me. I wonder if it's worth mentioning in the prose that the same attributes that apply to parameters (`[FromRoute]`, `[FromBody]`, `[FromServices]`, etc.) also work with the type used for `[AsParameters]` and cross-reference to wherever else in the docs that was originally explained for .NET 6 and parameter binding?

Maybe also include something related to the fact that it is not a complex model binding (recursive as MVC) and instead it iwll just expand the `properties` are handler `parameters`.

Rick-Anderson:
> @Rick-Anderson since `struct` are the most optimized scenario, what do you think about show that usage in the examples instead? Maybe mentioning that a `class` also works.

@martincostello  would you be able to PR your [arg-lists sample](https://github.com/dotnet/AspNetCore.Docs.Samples/commits/main/fundamentals/minimal-apis/samples/arg-lists) and change it to structs?  Feel free to keep that sample and add a struct version.

martincostello:
https://github.com/dotnet/AspNetCore.Docs.Samples/pull/41

PR #26945, opened 2022-09-07
Mention <base href="/" />

zanyar3:
This document missed mention `<base href="/" />`.
 because I was followed  this document but not work after searching then found this solation.


zanyar3:
#26945

captainsafia (changes requested):
This doesn't seem like the correct fix to your problem. The problem that you indicated in the other issue seems to be related to resolving Razor views given a particular route whereas the modification you are making here would affect resolution for all routes.

Can you post some more info in the original issue about the problem that you are running into? In particular, what does the structure of your main app look like?

zanyar3:
Hi @captainsafia thanks for reviewed issue.
I don't know how you decided that the issue I mentioned is not a issue, I have explained to you through code/issue what my problem was with **dotnetcore repo**.
I have explained to them very clearly what/where is problem !!?.
You can try test to see if there is the same problem.

PR #26950, opened 2022-09-08
Improve Anchor on Link to CSRF on Razor-Pages

MorgeMoensch:
The old one scrolls to the middle of a code block, instead of to the title. Try yourself: 

Old: https://docs.microsoft.com/en-us/aspnet/core/razor-pages/?view=aspnetcore-6.0&tabs=visual-studio#xsrfcsrf-and-razor-pages-1

New: https://docs.microsoft.com/en-us/aspnet/core/razor-pages/?view=aspnetcore-6.0&tabs=visual-studio#xsrfcsrf-and-razor-pages


guardrex:
Hello @MorgeMoensch ... This is a known issue with the current documentation versioning system that we use. We hope the approach will change eventually over to a new system where this won't be a problem. In the meantime, we won't be able to fix the problem for the current system this way.

PR #27008, opened 2022-09-15
Fix `[!NOTE]` syntax

GitHubPang:
The notes [were not rendered correctly](https://docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/signalr?view=aspnetcore-6.0#reflect-the-connection-state-in-the-ui-blazor-server) due to incorrect syntax of `[!NOTE]`.


guardrex (approved):
Good 👁️, @GitHubPang! ... That's a *rotten* 😈! 😆

PR #27014, opened 2022-09-15
Warn on MaximumReceiveMessageSize

### LINKED ISSUE #27013: Warn on MaximumReceiveMessageSize size (DOS risk)
Per https://github.com/dotnet/aspnetcore/issues/43974#issuecomment-1247486989.

---
#### Document Details

* Version Independent ID: d17e4c28-ea82-0dd8-295d-19f6ada147ea
* Content: [ASP.NET Core Blazor SignalR guidance](https://docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/signalr?view=aspnetcore-6.0)
* Content Source: [aspnetcore/blazor/fundamentals/signalr.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/fundamentals/signalr.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #27013

### DIFF
--- a/aspnetcore/blazor/fundamentals/signalr.md
+++ b/aspnetcore/blazor/fundamentals/signalr.md
@@ -167,7 +167,7 @@ builder.Services.AddServerSideBlazor()
 ```
 
 > [!WARNING]
-> We recommend setting <xref:Microsoft.AspNetCore.SignalR.HubOptions.MaximumReceiveMessageSize> to a small value in order to limit the risk of [Denial of service (DoS) attacks](xref:blazor/security/server/threat-mitigation#denial-of-service-dos-attacks). The default value is 32 KB (32,768 bytes).
+> We recommend setting <xref:Microsoft.AspNetCore.SignalR.HubOptions.MaximumReceiveMessageSize> to a small value in order to limit the risk of [Denial of service (DoS) attacks](xref:blazor/security/server/threat-mitigation#denial-of-service-dos-attacks). The default value is 32 KB.
 
 ## Blazor Hub endpoint route configuration (Blazor Server)
 
@@ -493,7 +493,7 @@ services.AddServerSideBlazor()
 ```
 
 > [!WARNING]
-> We recommend setting <xref:Microsoft.AspNetCore.SignalR.HubOptions.MaximumReceiveMessageSize> to a small value in order to limit the risk of [Denial of service (DoS) attacks](xref:blazor/security/server/threat-mitigation#denial-of-service-dos-attacks). The default value is 32 KB (32,768 bytes).
+> We recommend setting <xref:Microsoft.AspNetCore.SignalR.HubOptions.MaximumReceiveMessageSize> to a small value in order to limit the risk of [Denial of service (DoS) attacks](xref:blazor/security/server/threat-mitigation#denial-of-service-dos-attacks). The default value is 32 KB.
 
 ## Blazor Hub endpoint route configuration (Blazor Server)
 
@@ -816,7 +816,7 @@ services.AddServerSideBlazor()
 ```
 
 > [!WARNING]
-> We recommend setting <xref:Microsoft.AspNetCore.SignalR.HubOptions.MaximumReceiveMessageSize> to a small value in order to limit the risk of [Denial of service (DoS) attacks](xref:blazor/security/server/threat-mitigation#denial-of-service-dos-attacks). The default value is 32 KB (32,768 bytes).
+> We recommend setting <xref:Microsoft.AspNetCore.SignalR.HubOptions.MaximumReceiveMessageSize> to a small value in order to limit the risk of [Denial of service (DoS) attacks](xref:blazor/security/server/threat-mitigation#denial-of-service-dos-attacks). The default value is 32 KB.
 
 ## Blazor Hub endpoint route configuration (Blazor Server)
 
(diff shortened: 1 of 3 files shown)


guardrex:
Let's hold off on review. We're discussing this on the PU issue. We might do something different ... an INCLUDE used in the Blazor doc and the SignalR doc, for example. We might do this, but we might only do it in the SignalR doc.

guardrex:
Ok ... well ... the PU issue was closed right in the middle of the discussion on this.

Let's go with *Rick's Gambit*&trade; ... i.e. ... **_Call for review!_** 😆

Brennan, which way would you like to go?

* Just note this on the Blazor-side, as this PR does.
* Create a duplicate NOTE for the SignalR config topic ... I'll add it to this PR.
* Do something else.

BrennanConroy:
> * I can place it in the **Description** in the table at ...

Sure, similar to how `ApplicationMaxBufferSize` in a later table mentions memory usage can increase if the setting is changed.

guardrex:
@BrennanConroy ... Added a remark to the table. How's it look?

aspnetcore/blazor/fundamentals/signalr.md
```diff
@@ -166,6 +166,9 @@ builder.Services.AddServerSideBlazor()
     });
 ```
 
+> [!WARNING]
+> We recommend setting <xref:Microsoft.AspNetCore.SignalR.HubOptions.MaximumReceiveMessageSize> to a small value in order to limit the risk of [Denial of service (DoS) attacks](xref:blazor/security/server/threat-mitigation#denial-of-service-dos-attacks). The default value is 32 KB (32,768 bytes).
```
TanayParikh, line 170: I believe 32KB should be sufficient unless we follow the convention of writing total bytes elsewhere?

```suggestion
> We recommend setting <xref:Microsoft.AspNetCore.SignalR.HubOptions.MaximumReceiveMessageSize> to a small value in order to limit the risk of [Denial of service (DoS) attacks](xref:blazor/security/server/threat-mitigation#denial-of-service-dos-attacks). The default value is 32 KB.
```

aspnetcore/blazor/security/server/threat-mitigation.md
```diff
@@ -92,7 +92,7 @@ By default, there's no limit on the number of connections per user for a Blazor
 
 ## Denial of service (DoS) attacks
 
-Denial of service (DoS) attacks involve a client causing the server to exhaust one or more of its resources making the app unavailable. Blazor Server apps include default limits and rely on other ASP.NET Core and SignalR limits to protect against DoS attacks that are set on <xref:Microsoft.AspNetCore.Components.Server.CircuitOptions>:
+[Denial of service (DoS) attacks](https://developer.mozilla.org/docs/Glossary/DOS_attack) involve a client causing the server to exhaust one or more of its resources making the app unavailable. Blazor Server apps include default limits and rely on other ASP.NET Core and SignalR limits to protect against DoS attacks that are set on <xref:Microsoft.AspNetCore.Components.Server.CircuitOptions>:
```
TanayParikh, line 95:
```suggestion
[Denial of service (DoS) attacks](https://developer.mozilla.org/docs/Glossary/DOS_attack) involve a client causing the server to exhaust one or more of its resources making the app unavailable. Blazor Server apps include default limits and rely on other ASP.NET Core and SignalR limits that are set on <xref:Microsoft.AspNetCore.Components.Server.CircuitOptions> to protect against DoS attacks:
```

TanayParikh (commented):
LGTM, will leave the final sign-off for @BrennanConroy

aspnetcore/blazor/fundamentals/signalr.md
guardrex: > the convention

That's the problem. The standard was never adopted. They wrote a whole page on how to use bit/byte terms without even addressing it ...

https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/term-collections/bits-bytes-terms

I think ur aware technically (IEC and SI) that we should be using "KiB" (kibibyte) to eliminate the confusion (or "kB" for multiples of 1,000). Since they won't permit that approach, this is how I've been hardening docs to avoid confusion.

guardrex: I might change it tho ... I'm going to open an issue to discuss this with the docs 🐈 🐈 🐈.

I performed a search to see how consistent I've been. It looks like in more than half the cases, I did write out either format (X bytes or X * 1024). It looks like a number of cases without it exist, mostly in content sent over from the PU that I didn't change on publication.

I'll circle around and update depending on the outcome of that discussion ... but let's not hold this up over that.

TanayParikh (approved):
LGTM, but I'd await @BrennanConroy's signoff.

guardrex:
I will ... and btw ... I've talked myself out of using explicit byte conventions, as you recommended, on the issue that I opened ...

https://github.com/dotnet/AspNetCore.Docs/issues/27056

I wish MS/.NET foundation would adopt the standard. What I can do is place a remark on it in the Blazor *Fundamentals* topic and just use "KB"/"MB"/"GB" everywhere.

guardrex:
@BrennanConroy ... Are you OOF? No rush ... just checking.

aspnetcore/signalr/configuration.md
```diff
@@ -447,7 +447,7 @@ The following table describes options for configuring SignalR hubs:
 | `SupportedProtocols` | All installed protocols | Protocols supported by this hub. By default, all protocols registered on the server are allowed. Protocols can be removed from this list to disable specific protocols for individual hubs. |
 | `EnableDetailedErrors` | `false` | If `true`, detailed exception messages are returned to clients when an exception is thrown in a Hub method. The default is `false` because these exception messages can contain sensitive information. |
 | `StreamBufferCapacity` | `10` | The maximum number of items that can be buffered for client upload streams. If this limit is reached, the processing of invocations is blocked until the server processes stream items.|
-| `MaximumReceiveMessageSize` | 32 KB | Maximum size of a single incoming hub message. |
```
BrennanConroy, line 450: nit: the wording "We recommend setting" makes it sound like the user /should/ change the value.
"We recommend keeping the value small", "Increasing this value may increase the risk of DOS...", etc.

guardrex: The 2nd one sounds good :ear: to me. I'll try that on the next commit.

guardrex:
@BrennanConroy ... Ok ... ready for another 👁️.

BrennanConroy (approved):

Issue #22862, opened 2021-07-27
"Place extension methods in the Microsoft.Extensions.DependencyInjection namespace to encapsulate groups of service registrations." ??


StingyJack:
Why would we put any of our code in a Microsoft namespace?  What is the value in doing that?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0eb4d762-3910-2c57-8741-3306c15537b6
* Version Independent ID: ba2254a0-ea8e-9b28-e94b-edd0fc58c043
* Content: [Configuration in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-3.1#combining-service-collection)
* Content Source: [aspnetcore/fundamentals/configuration/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/configuration/index.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-fundamentals**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
convention

davidfowl:
That's right. They show up in intellisense automagically as more components are referenced

StingyJack:
> convention

> That's right. They show up in intellisense automagically as more components are referenced

By the convention you are advocating, it sounds like you would expect me to put any extension method I have for `System.String` into the `System` namespace so that intellisense can find it. TBH this advice is strange enough that it feels like you are having a laugh at my expense (which is fine if you need it, just dont take it too far).

Intellisense finds the extension methods no matter where I tend to put them. "My" code belongs in "my" namespaces so it can be located or identified correctly.

davidfowl:
I think in reality the IServiceCollection and the IApplicationBuilder aren't as pervasive as string 🙃. So you're argument while valid doesn't apply here. Also it's guidance but if you feel strongly about your component you can ignore it. It's something that all of the built in components do and something we will continue to do.

StingyJack:
> So you're argument 

My argument what?🤔

Autocorrect fun aside, your guidance here (to all users, not just MSFT ones) is to break long standing convention and guidance regarding namespaces. Do you know how many people that don't work at MSFT will blindly follow this and the arguments that will be had over it? And the reasons proponents will be giving is that intellisense performs better this way, and that it's Best Practice to do this. 

[...]

Issue #20535, opened 2020-11-12
Putting my code in a Microsoft namespace


ikyriak:
[EDIT by guardrex to add the metadata ..... @ikyriak, it's best to open new issues on the repo using the **This page** feedback button at the bottom of the topic, which adds this metadata automatically and pings the author.]

Related Closed Issue: dotnet/AspNetCore.Docs#11435

Can we have an update regarding the suggestion to "Place extension methods in the Microsoft.Extensions.DependencyInjection namespace to encapsulate groups of service registrations." [[REF](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/fundamentals/dependency-injection.md)].

The last comment was on Jun 27, 2019 from @guardrex : "Engineering has looked this topic over multiple times since then and never said to change it. We'll be reviewing updates again when we get to 3.0 updates. I'll ask explicitly in a few months when I work it."

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0d78f540-81fd-1ba9-cb90-591e4ec27d9b
* Version Independent ID: 2251f557-dd3a-ec35-2558-6243337db009
* Content: [Dependency injection in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-5.0)
* Content Source: [aspnetcore/fundamentals/dependency-injection.md](https://github.com/dotnet/AspNetCore.Docs/blob/master/aspnetcore/fundamentals/dependency-injection.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-fundamentals**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

guardrex:
They ask that devs don't **_publish NuGet packages with "Microsoft" in the name of the package._** They sharpen their 🔪😨 when they see that happen! :smile:

@davidfowl ... This guidance was attributed to you by @ardalis when the DI topic was written ...

> Place extension methods in the `Microsoft.Extensions.DependencyInjection` namespace to encapsulate groups of service registrations.

... is that the guidance that you want?

guardrex:
cc: @mkArtakMSFT ... probably can answer this definitively, too.

TL;DR; &mdash; The DI topic states ...

> Place extension methods in the `Microsoft.Extensions.DependencyInjection` namespace to encapsulate groups of service registrations.

[...]

Issue #11435, opened 2019-03-15
Putting my code in a Microsoft namespace


EldieTurner:
I would like more clarification on why someone would put code they wrote inside a Microsoft.Extensions.DependencyInjection namespace.  If I see an extension method with this namespace I assume it is code that was written by Microsoft.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0d78f540-81fd-1ba9-cb90-591e4ec27d9b
* Version Independent ID: 2251f557-dd3a-ec35-2558-6243337db009
* Content: [Dependency injection in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.2#feedback)
* Content Source: [aspnetcore/fundamentals/dependency-injection.md](https://github.com/aspnet/Docs/blob/master/aspnetcore/fundamentals/dependency-injection.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
@EldieTurner Hello ... Thanks for commenting. That guidance was provided by an earlier author, and we haven't had management or engineering take issue with it on the reviews of this topic.  That should be fine. The problem is when devs name (and **_publish_**) their NuGet packages **Microsoft.SomethingCool**. Yeah ... *My advice would be to get a good attorney!!* :smile: lol ... It really does mislead people when a published package has that naming.

spankr:
The issue I am dealing with now is that other engineers are taking the line, _"Place extension methods in the Microsoft.Extensions.DependencyInjection namespace to encapsulate groups of service registrations."_ and declaring it "the standard used by the ASP.NET vNext team" (https://github.com/OData/WebApi/issues/576).

I'm with @EldieTurner on assuming everything under Microsoft.* should be Microsoft-written. It should probably be clarified as either officially good or bad before too many teams take this earlier author's guidance as an industry standard.

spankr:
Okay, that OData link is a poor example(my bad). It makes perfect sense for them to use the namespace, as OData is part of Microsoft (TIL).
But I do have non-MS engineers pointing at examples like that (and this doc) and drawing the conclusion that everyone should be using that namespace for their extensions of **IServiceCollection**.

guardrex:
AFAIK, the answer is *Yes!* I was able to trace the remark back to this ...

[...]

PR #27018, opened 2022-09-15
Address dotnet/docs #31084 - contradicting recommendations

IEvangelist:
The .NET docs formally recommend _not_ using the `Microsoft.Extensions.DependencyInjection` namespace, [.NET docs: Options pattern guidance for .NET library authors — Namespace guidance](https://docs.microsoft.com/dotnet/core/extensions/options-library-authors#namespace-guidance). The sentence being removed in this PR are a contradiction to the .NET docs recommendation.

This PR removes the recommendation to use namespaces owned by Microsoft. Otherwise, doing so has a good chance of colliding with our future additions in these namespaces.

Fixes dotnet/docs#31084

### DIFF
--- a/aspnetcore/includes/combine-di6.md
+++ b/aspnetcore/includes/combine-di6.md
@@ -12,7 +12,4 @@ The remaining services are registered in a similar class. The following code use
 
 [!code-csharp[](~/fundamentals/configuration/index/samples/6.x/ConfigSample/program.cs?name=snippet3)]
 
-**_Note:_** Each `services.Add{GROUP_NAME}` extension method adds and potentially configures services. For example, <xref:Microsoft.Extensions.DependencyInjection.MvcServiceCollectionExtensions.AddControllersWithViews%2A> adds the services MVC controllers with views require, and <xref:Microsoft.Extensions.DependencyInjection.MvcServiceCollectionExtensions.AddRazorPages%2A> adds the services Razor Pages requires. We recommend that apps follow the naming convention of creating extension methods in the <xref:Microsoft.Extensions.DependencyInjection?displayProperty=fullName> namespace. Creating extension methods in the `Microsoft.Extensions.DependencyInjection` namespace:
-
-* Encapsulates groups of service registrations.
-* Provides convenient [IntelliSense](/visualstudio/ide/using-intellisense) access to the service.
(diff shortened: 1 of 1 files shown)


guardrex:
@IEvangelist ... I think we should research this back unless you've been in contact with Dan about it. There was a discussion on it when this guidance was provided, and this was **_truly the PU's guidance_** on this point back in those days. IIRC, I think even Fowler and Damian were cool with it. We'd need to research it back and find the discussion to confirm. Let me know if you want me to see if I can scare up the chat on it. I'll be free within two or three hours to see if I can find it.

IEvangelist:
Hey, no worries. Yeah, @davidfowl, @jkotas, and @stephentoub are the three who aligned on this update this morning.

jkotas:
We should not be recommending that library authors place extension methods in the system or Microsoft namespaces. It has a good chance of colliding with our future additions in these namespaces. It is general guidance for all libraries. I do not see why DI should be an exception.

guardrex:
It came up several times over the years. It wasn't discussed much other than to say 'it's fine ... go ahead' with some offhand remarks about the PU being ok with it. For example ...

* https://github.com/dotnet/AspNetCore.Docs/issues/20535
* https://github.com/dotnet/AspNetCore.Docs/issues/11435
* One of the earliest mentions ... **_2016!_** *Good grief! I'm getting really*👴 _!_ 😄: https://github.com/dotnet/AspNetCore.Docs/pull/2145#issuecomment-260189456

But this last one that a quick search turned up is probably the one that I was thinking of, where @davidfowl was discussing it with a community dev ...

> When you're using `IServiceCollection`, `IApplicationBuilder` and `IEndpointRouteBuilder` you already have the namespaces Microsoft.Extensions.DependencyInjection, Microsoft.AspNetCore.Builder and Microsoft.AspNetCore.Routing imported in the file. If you were to create a package that has extension methods [e.g.](https://github.com/abpframework/abp/blob/7488b929ba80c8535591a7be0fed15390ca11915/framework/src/Volo.Abp.MemoryDb/Microsoft/Extensions/DependencyInjection/AbpMemoryDbServiceCollectionExtensions.cs#L10), it would show up in the intellisense as an extension method on those types with no further work from the user.

* https://github.com/dotnet/AspNetCore.Docs/issues/22862

Rick-Anderson (approved):

PR #27039, opened 2022-09-18
Updated to .NET 7 and few code readability optimizations

nirzaf:
NuGet package updated from Preview to Stable Version


guardrex:
Hello @nirzaf ... This is the 6.0 sample app for the 6.0 version of the article. We create a new sample for each release when the folder structure indicates the version. What you'd need to do is create a 7.0 folder and place the full app into that folder. The article would then (either at the same time or after that) also get a new 7.0 copy of the 6.0 content with the links updated to point to the 7.0 sample.

PR #27048, opened 2022-09-19
Update signalr-blazor.md

d4shm1r:
There seems to be a missing code block at step 4


guardrex:
Thanks for pointing this out, @d4shm1r. That step shows the code below, which includes the line in question, along with the other line. The step is **_messy_** 🤮. I see what I'd like to do to fix up the whole step: I'm going to avoid individual code lines, leave the bullets, and highlight those two lines in the existing code block. I'll do that across all versions of the doc. I'll take care of this first thing Tuesday morning. Thanks again for pointing it out.

Moved to 👉 https://github.com/dotnet/AspNetCore.Docs/issues/27052

PR #27058, opened 2022-09-20
Improve section leading paragraphs

### LINKED ISSUE #27055: Is this misleading?
In this section : https://learn.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-6.0#use-key-to-control-the-preservation-of-elements-and-components

> When rendering a list of elements or components and the elements or components subsequently change, Blazor must decide  which of the previous elements or components can be retained and how model objects should map to them. _**Normally, this process is automatic and can be ignored**_, but there are cases where you may want to control the process.

This is misleading as it suggests that Blazor will "get it right" most of the time and that is good enough. The truth is that preservation of elements/components rendered in a list/loop in Blazor is _only guaranteed_ by use of the `@key` directive - as noted further down the page, but only if the reader gets that far.

> Use of [@key](https://learn.microsoft.com/en-us/aspnet/core/mvc/views/razor?view=aspnetcore-6.0#key) guarantees the preservation of elements or components based on the key's value.

The reason I am raising this issue is because of the number of times developers are still not using the [@key](https://learn.microsoft.com/en-us/aspnet/core/mvc/views/razor?view=aspnetcore-6.0#key) and posting questions on SO about why they are seeing strange behaviour in loops.

### Suggestion:

Change the wording in that opening section

> When rendering a list of elements or components and the elements or components subsequently change, Blazor must decide  which of the previous elements or components can be retained and how model objects should map to them. Without use of the `@key` directive, this process is automatic based on ordering of the rendered items. However, there are many cases where you may want to control the process.
---
#### Document Details

* Content: [ASP.NET Core Razor components](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-6.0)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes  #27055

Thanks @SQL-MisterMagoo! 🎷 ... I'll hold here for your feedback on the updates. I tried to steer it a bit more in the 'it may be required for you to control it' direction.

Also note that this is kind'a *draft-y-ish* at the moment. I need to run out for a couple of hours. I'll give it another 👁️ when I get back to perhaps provide further polish.

### DIFF
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -772,7 +772,7 @@ When rendering a list of elements or components and the elements or components s
 
 Consider the following example that demonstrates a collection mapping problem that's solved by using [`@key`][5].
 
-The following `Details` and `People` components:
+For the following `Details` and `People` components:
 
 * The `Details` component receives data (`Data`) from the parent `People` component, which is displayed in an `<input>` element. Any given displayed `<input>` element can receive the focus of the page from the user when they select one of the `<input>` elements.
 * The `People` component creates a list of person objects for display using the `Details` component. Every three seconds, a new person is added to the collection.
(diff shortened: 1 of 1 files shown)


SQL-MisterMagoo:
Thanks - I think that reads much better.

guardrex:
Ok ... that looks decent.

... and **_I HOPE_** 🙏 they read the *whole bleed'in thing!* 😆 What a pain! *Long-winded professors and their rambling lectures!*

Thanks again ... this is a nice improvement.

PR #27061, opened 2022-09-20

adaggarwal:
This of this as a suggestion. I wonder if the intent was to mention that the type of <some interface> is requested.


guardrex:
Hello @adaggarwal ... Actually, that's correct. It's a service that's requested from the DI container.

PR #27075, opened 2022-09-22
markdown style error

wuyuansushen:


guardrex (approved):
Thanks @wuyuansushen ... It's not a style error in quite that way. The code-fencing should be on the inside of the bold style. It's code-fenced because project template code (e.g., the name of the "Server" project in a hosted WASM solution) isn't localized into non-English languages. Thanks for the PR. The update will fix this here.

wuyuansushen:
OK. I agree your opnions. I have known that `Server` is the name of project now...

guardrex:
There's another way to do it that I'm probably going to implement for all of the Blazor docs. We have a triple-colon syntax that would like like this for these cases ...

```
**:::no-loc text="Server":::**
```

I'm leaning toward making the change later. This isn't a good time to do it given all of the doc churn right now. We're working on the doc updates for .NET 7's release, which is planned for November. I'll mark this on my tracking issue and get to it after the .NET 7 release.

PR #27120, opened 2022-09-28
Clarify AddEntityFrameworkStores behavior

HaoK:
Fixes https://github.com/dotnet/aspnetcore/issues/12511

Clarify AddEntityFrameworkStores behavior and workaround when not using a derived IdentityDbContext

### DIFF
--- a/aspnetcore/security/authentication/customize-identity-model.md
+++ b/aspnetcore/security/authentication/customize-identity-model.md
@@ -292,6 +292,8 @@ The context is used to configure the model in two ways:
 
 When overriding `OnModelCreating`, `base.OnModelCreating` should be called first; the overriding configuration should be called next. EF Core generally has a last-one-wins policy for configuration. For example, if the `ToTable` method for an entity type is called first with one table name and then again later with a different table name, the table name in the second call is used.
 
+NOTE: If the `DbContext` for your app does not derive from `IdentityDbContext`, `AddEntityFrameworkStores` potentially will not infer the correct POCO types, for TUserClaim, TUserLogin, TUserToken, you should directly add your stores via `services.AddScoped<IUser/RoleStore<TUser>, UserStore<...>>` with the correct types.
+
 ### Custom user data
 
 <!--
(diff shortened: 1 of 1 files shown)


aspnetcore/security/authentication/customize-identity-model.md
```diff
@@ -292,6 +292,8 @@ The context is used to configure the model in two ways:
 
 When overriding `OnModelCreating`, `base.OnModelCreating` should be called first; the overriding configuration should be called next. EF Core generally has a last-one-wins policy for configuration. For example, if the `ToTable` method for an entity type is called first with one table name and then again later with a different table name, the table name in the second call is used.
 
+NOTE: If the `DbContext` for your app does not derive from `IdentityDbContext`, `AddEntityFrameworkStores` potentially will not infer the correct POCO types, for TUserClaim, TUserLogin, TUserToken, you should directly add your stores via `services.AddScoped<IUser/RoleStore<TUser>, UserStore<...>>` with the correct types.
```
Rick-Anderson, line 295:
```suggestion
***NOTE***: If the `DbContext` doesn't derive from `IdentityDbContext`:
* `AddEntityFrameworkStores` potentially won't infer the correct POCO types for `TUserClaim`, `TUserLogin`, and `TUserToken`.
* Stores should be directly added via `services.AddScoped<IUser/RoleStore<TUser>` and `UserStore<...>>` with the correct types.
```

That might need some work but the sentence is too long.

HaoK: What about the last line being a separate sentence to the effect of its a workaround and can be used if `AddEntityFrameworkStores` isn't doing the right thing for them?

HaoK: Stores can always be directly added...with the correct types

Rick-Anderson: @HaoK  check my updated suggestion.

HaoK:
Looks much better, thanks for cleaning up the wording as always!

PR #27204, opened 2022-10-06
Fix snippets to keep @blowDart not red faced

Rick-Anderson:


Rick-Anderson:
@JamesNK I checked the build and the snippets are the same.

PR #27224, opened 2022-10-08
Update Index.cshtml.cs

jf-gbg:
Add missing using System.Collections.Generic; statement


jf-gbg:
Building a fresh download of completed project failed due to the missing using statement in Instructors\Index.cshtml.cs

PR #27226, opened 2022-10-09
Broken link update

jaroslavbenes:


guardrex (approved):
Hello @jaroslavbenes ... That's not the correct link. I'm not sure why the original was breaking for you. The tutorial is at ...

https://learn.microsoft.com/en-us/aspnet/core/getting-started

However, we should have an XREF-style link there that in addition to setting the destination also automatically pulls the title. That permits us to make title changes to an article and have the change reflected everywhere across the docs.

**UPDATE**: Changed it on https://github.com/dotnet/AspNetCore.Docs/pull/27227: I chop the end of the sentence, which doesn't make sense. Also, XREF doesn't work in unpublished markdown files anyway. I just woke up **_AND NEED MORE ☕☕☕☕ NOW!_** 🤣

PR #27228, opened 2022-10-09
Fix broken link in index page on ASP.NET documentation

wmundev:
Fixes broken link i found on the main ASP.NET documentation


guardrex (approved):
Thanks @wmundev! 🚀

btw @Rick-Anderson ... A number of these link query strings don't look right. Do the "breadcrumb" parts do anything? Why are some of these pegged to prior release versions? Why peg the VS doc set to 2022's release (i.e., they default to latest on their own)? These are rhetorical questions. I just wanted to let you know.

PR #27242, opened 2022-10-11
Fix its grammar mistake

Bouke:


guardrex (approved):
Thanks @Bouke! :rocket:

Issue #26819, opened 2022-08-25
Route groups


Rick-Anderson:
[Route groups](https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-dotnet-7-preview-4/#route-groups)

Probably do this in multiple PR's

- [x] Update this intro tutorial using sample code for a basic scenario
- [ ] sample code for more advanced scenarios
- [ ] Update [Minimal APIs overview](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis?view=aspnetcore-7.0) and [What's new in ASP.NET Core 7.0](https://docs.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-7.0?view=aspnetcore-7.0)

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ab408035-6313-a183-9940-32de00e0e977
* Version Independent ID: ab408035-6313-a183-9940-32de00e0e977
* Content: [Tutorial: Create a minimal web API with ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/tutorials/min-web-api?view=aspnetcore-6.0&tabs=visual-studio)
* Content Source: [aspnetcore/tutorials/min-web-api.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/tutorials/min-web-api.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-tutorials**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

fiyazbinhasan:
Sure. I would like to do that :)

fiyazbinhasan:
I think @tdykstra already started on this. Do let me know if you need a hand 😺

tdykstra:
Yes I do! The MapGroup sample code I have in the intro tutorial is too basic for the what's new article. For the what's new I need an example that includes code similar to what we have in the [Preview 4 blog post](https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-dotnet-7-preview-4/#route-groups), i.e., code that illustrates what you can do with route groups in more advanced scenarios. This sample would ideally also include code that [returns multiple result types from an endpoint](https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-dotnet-7-preview-4/#return-multiple-result-types-from-minimal-apis).

Rick-Anderson:
[...]

PR #27177, opened 2022-10-03
Route Group Fundamental Sample

fiyazbinhasan:
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/26819

@tdykstra please review, I was unable to create a relative path using linkGenerator. 

I tried with,

`linkGenerator.GetUriByRouteValues(context, null, new {id = todo.Id})`, but it always returns null. Do you have any idea or tips on generating a link using the technique mentioned in the blog? 

> Instead of using relative addresses for the Location header in the 201 Created result, it’s also possible to use GroupRouteBuilder.GroupPrefix to construct a root-relative address or use [GetPathByRouteValues or GetUriByRouteValues](https://docs.microsoft.com/dotnet/api/microsoft.aspnetcore.routing.linkgeneratorroutevaluesaddressextensions) with a routeName for even more options.

PR #27249, opened 2022-10-11
Route Group Fundamental Sample

fiyazbinhasan:
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/26819

@tdykstra I accidentally removed my fork, so this PR https://github.com/dotnet/AspNetCore.Docs/pull/27177 has closed automatically. Creating a new PR. 

However, it would be best if someone from the product team could intervene here. Maybe I'm missing some crucial configuration in order to make the `linkGenerator` work

> These are very different configurations,
> 
> ```
> app.MapGroup("/{accessibility}/todos")
>     .MapTodosApi()
>     .WithTags("Todo Endpoints");
> 
> group.MapGet("/{id}", GetTodo).WithName("GetTodo");
> ```
> 
> With the following approach, you **can't** assign any name to your group. It will throw you an exception saying the name should be globally unique, hence `linkGenerator` won't be able to generate any links.
> 
> ```
> app.MapGroup("/public/todos")
>     .MapTodosApi()
>     .WithTags("Public Todo Endpoints");
> 
> app.MapGroup("/private/todos")
>     .MapTodosApi()
>     .RequireAuthorization();
> ```

In the sample, we should also consider showing a nested route group as well. What do you think?


Rick-Anderson:
@captainsafia Please review, having problem with `linkGenerator`

captainsafia:
@halter73 Do we happen to have a workaround for the link generation issue mentioned in https://github.com/dotnet/AspNetCore.Docs/pull/27177 or should we document it as a known issue?

aspnetcore/fundamentals/minimal-apis/7.0-samples/todo-group2/Program.cs
```diff
@@ -0,0 +1,38 @@
...
+app.MapGroup("/public/todos")
+    .MapTodosApi()
+    .WithTags("Public Todo Endpoints");
+
+app.MapGroup("/private/todos")
+    .MapTodosApi()
+    .RequireAuthorization();
+
+// Configure the HTTP request pipeline.
+if (app.Environment.IsDevelopment())
```
captainsafia, line 32: Can you move this to L18?

Rick-Anderson: @captainsafia 
> Can you move this to L18?
Is this what you want?

```suggestion
var app = builder.Build();

app.MapGroup("/private/todos")
    .MapTodosApi()
    .RequireAuthorization();

using var scope = app.Services.CreateScope();
var db = scope.ServiceProvider.GetService<TodoGroupDbContext>();
db?.Database.MigrateAsync();

app.MapGroup("/public/todos")
    .MapTodosApi()
    .WithTags("Public Todo Endpoints");

// Configure the HTTP request pipeline.
if (app.Environment.IsDevelopment())
```

Rick-Anderson:
@fiyazbinhasan is link generation working now?

halter73:
> With the following approach, you can't assign any name to your group. It will throw you an exception saying the name should be globally unique, hence linkGenerator won't be able to generate any links.

Groups doesn't give you a free lunch when it comes to naming endpoints. Endpoints defined by `MapGet`, `MapPost`, etc... don't work with special `RouteValuesAddress`-based extension methods like `GetPathByPage` and `GetPathByAction`. It might be interesting to make these extension methods and make the key the combined `RoutePattern.RawText` (`GetPathByRoutePattern`?), but that's not something we've implemented yet.

As it stands, `GetPathByName` does work with `MapGet`, `MapPost`, etc... but that requires uniquely naming each endpoint using `WithName` or the `EndpointNameAttribute` which doesn't make sense on a group. This isn't a group limitation. This is just a current limitation of minimal route handlers. For now, I don't think we should demonstrate link generation in a group-focused sample.

> Do we happen to have a workaround for the link generation issue mentioned in #27177 or should we document it as a known issue?

It turns out that the problem with that test was not related to groups after all. Even without the MapGroup calls, the page failed to generate a contact link. I fixed the test. https://github.com/dotnet/aspnetcore/pull/44562

tdykstra (approved):
Thanks, @fiyazbinhasan

PR #27260, opened 2022-10-12
Minimal APIS Optional Parameters example /products/two will return a 404 and not throw

collinstevens:
In the example for [optional parameters](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis?view=aspnetcore-6.0#optional-parameters), the route registered is `app.MapGet("/products", (int? pageNumber) => $"Requesting page {pageNumber ?? 1}");`.

If you call the API with `/products/two` it will return a `404` and not throw an exception as the documentation indicates.

See the attached screenshots.

image

image

### DIFF
--- a/aspnetcore/fundamentals/minimal-apis.md
+++ b/aspnetcore/fundamentals/minimal-apis.md
@@ -496,7 +496,8 @@ The preceding code calls the method with a null product if no request body is se
 | --------- | -------------- |
 | `/products?pageNumber=3` | `3` returned |
 | `/products` | `1` returned |
-| `/products/two` | `BadHttpRequestException`: Failed to bind parameter `"Nullable<int> pageNumber"` from "two".|
+| `/products?pageNumber=two` | `BadHttpRequestException`: Failed to bind parameter `"Nullable<int> pageNumber"` from "two". |
+| `/products/two` | HTTP 404 error, no matching route |
 
 See the [Binding Failures](#bf) section for more information.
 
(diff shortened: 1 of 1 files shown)


aspnetcore/fundamentals/minimal-apis.md
```diff
@@ -1340,7 +1340,7 @@ The preceding code calls the method with a null product if no request body is se
 | --------- | -------------- |
 | `/products?pageNumber=3` | `3` returned |
 | `/products` | `1` returned |
-| `/products/two` | `BadHttpRequestException`: Failed to bind parameter `"Nullable<int> pageNumber"` from "two".|
+| `/products/two` | HTTP 404 error, no matching route |
```
Rick-Anderson, line 1343: @collinstevens why doesn't the route match? We should add this row and update the original row to something the throws. Maybe `/products/0x2` or  `/products/2x`

brunolins16: I agree with @Rick-Anderson, we could change the current row with `/products?pageNumber=two` and add this new line.

collinstevens: @Rick-Anderson I believe the route `/products/two` returns a `404` because there is no route with a configured route template of `/products/{*path}` or `/products/two/`; not because of model binding as the original documentation would indicate.

Here is a repository to use for this example. https://github.com/collinstevens/MinimalApiProductRoute

I'm not even sure if the documentation _should_ include something which would throw because the header of the example is `If invalid data is provided and the parameter is nullable, the route handler is not run.`, therefore, I would expect model binding to **never** have run to begin with, and not to see the model binding exception.

Nevertheless, @brunolins16 and @Rick-Anderson, I've updated the branch to include the suggested new line and the current line I modified. I also added the changes to the previous framework moniker I originally missed.

Issue #27298, opened 2022-10-15
OpenAPI support in minimal APIs v7 Examples


timdeschryver:
Hello, first of all thanks for the new `OpenAPI` API 🎉

While going through the docs, I tried out the examples in the docs and bumped against the following issue.

The examples in the v7 docs that are using `WithOpenApi` do not compile.

For example, the snippet at https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/openapi?view=aspnetcore-7.0#add-operation-ids-to-openapi-1

```cs
app.MapGet("/todos", async (TodoDb db) => await db.Todos.ToListAsync())
    .WithOpenApi(operation => new(operation)
    {
        OperationId = "GetTodos"
    });
```

Give the following error:

```
Constructor 'OpenApiOperation' has 0 parameter(s) but is invoked with 1 argument(s)
```

Screenshot:

![image](https://user-images.githubusercontent.com/28659384/195991916-a394adf2-c66d-40bb-98c1-e4ca39847444.png)

I'm using the latest version (7.0.0-rc.2.22476.2) of Microsoft.AspNetCore.OpenApi, which came with a newly generated project using `dotnet new`.

I suppose this should become the following, if so I'm happy to update the docs 😀

```cs
app.MapGet("/todos", async (TodoDb db) => await db.Todos.ToListAsync())
    .WithOpenApi(operation =>
    {
        operation.OperationId = "GetTodos";
        return operation;
    })
```

One small detail: I think the snippet is used for v6 and v7, but it would be great if the snippet contained the correct number - which makes it easier to copy paste 😊

![image](https://user-images.githubusercontent.com/28659384/195991723-762595e6-f62b-4fe4-a570-5f2e3f29d754.png)

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: b18acd86-50e3-2316-27be-27156651e5a8
* Version Independent ID: b18acd86-50e3-2316-27be-27156651e5a8
* Content: [How to use OpenAPI with minimal APIs](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/openapi?view=aspnetcore-7.0)
* Content Source: [aspnetcore/fundamentals/minimal-apis/openapi.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/minimal-apis/openapi.md)
[...]

PR #27299, opened 2022-10-15
Update openapi.md

captainsafia:
Addresses feedback from https://github.com/dotnet/AspNetCore.Docs/issues/27298

### DIFF
--- a/aspnetcore/fundamentals/minimal-apis/openapi.md
+++ b/aspnetcore/fundamentals/minimal-apis/openapi.md
@@ -75,7 +75,7 @@ ASP.NET Core provides the [`Microsoft.AspNetCore.OpenApi`](https://www.nuget.org
 
 [!code-xml[](7.0-samples/WebMinOpenApi/projectFile.xml?highlight=10)]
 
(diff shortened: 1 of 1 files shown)


Rick-Anderson (approved):

aspnetcore/fundamentals/minimal-apis/7.0-samples/WebMinOpenApi/projectFile.xml
```diff
@@ -8,7 +8,7 @@
 
   <ItemGroup>    
     <PackageReference Include="Microsoft.AspNetCore.OpenApi" Version="7.0.*-*" />
-    <PackageReference Include="Swashbuckle.AspNetCore" Version="6.2.3" />
+    <PackageReference Include="Swashbuckle.AspNetCore" Version="6.4.0" />
```
Rick-Anderson, line 11: This will be out of date soon. Can't we a [Floating Versions](https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#floating-versions)
`Version="6.*-*" />`

captainsafia: I'm a little hesitant to use floating versions for external dependencies. I'd rather pin to versions that we've tested against. This matches the version we are shipping with in the templates for 7.0...

Issue #26364, opened 2022-07-08
Blazor .NET 7 tracking issue


guardrex:
## Derived from the 7.0 roadmap

🟢 = Ready&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; ✔️ = Done&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; 🛑 = Wait on PU to finish&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; ❌ = No action required

Status | Live | Issue | PR | Item | PU Links/<br>Blog
:---: | :---: | :---: | :---: | --- | :---:
✔️ | [Live 1](https://docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-7.0#navigation-options)<br>[Live 2](https://docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-7.0#handleprevent-location-changes) | &mdash; | #26919<br>#26947 | Expose location changing event for NavigationManger | [42638](https://github.com/dotnet/aspnetcore/pull/42638) [42877](https://github.com/dotnet/aspnetcore/issues/42877) [43079](https://github.com/dotnet/aspnetcore/pull/43079) [Blog](https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-dotnet-7-rc-1/#handle-location-changing-events)
✔️ | [Live](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-7.0#invoke-component-methods-externally-to-update-state) | &mdash; | #26574 | Calling InvokeAsync(StateHasChanged) causes page to fallback to default culture | [28521](https://github.com/dotnet/aspnetcore/issues/28521)
✔️ | [Live 1](https://docs.microsoft.com/en-us/aspnet/core/blazor/tooling?view=aspnetcore-7.0&pivots=windows)<br>[Live 2](https://docs.microsoft.com/en-us/aspnet/core/blazor/project-structure?view=aspnetcore-7.0) | &mdash; | #26575 #26576 | Empty Blazor project template | [32832](https://github.com/dotnet/aspnetcore/issues/32832) [Blog](https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-dotnet-7-preview-6/#empty-blazor-project-templates)
[...]

PR #27311, opened 2022-10-18
Run .NET from JS article

### LINKED ISSUE #27016: New Blazor JS interop
Per https://github.com/dotnet/blazor-samples/issues/41 ...

> there is new [JavaScript interop available](https://github.com/dotnet/dotnet-api-docs/pull/8390) for .Net 7.
>
> I created simple demo to show the new JavaScript interop in Blazor.
>
> I'm not sure if you want to adopt it as is, but perhaps you could have look at it and extend one of the existing sample apps.
>
> https://github.com/pavelsavara/blazor-wasm-hands-pose
>
> There is also non-Blazor demo which has more complex interop logic
>
> https://github.com/pavelsavara/dotnet-wasm-todo-mvc
>
> (also please ping me on IM to discuss how to collaborate on further docs about the feature, thank you)

It's best if we have a section and extend the snippet sample apps, as opposed to **_adding_** a new sample app. Every sample app adds overhead costs 💸☹️.

cc: @pavelsavara

---
#### Document Details

* ID: 57ed3088-2f9b-f51a-fd17-0c0de2949c57
* Version Independent ID: 49fa31d8-72a6-01a1-027d-6edeac9e0fc4
* Content: [ASP.NET Core Blazor JavaScript interoperability (JS interop)](https://docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/?view=aspnetcore-6.0)
* Content Source: [aspnetcore/blazor/javascript-interoperability/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/javascript-interoperability/index.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #27016
Addresses #26364

[Internal Review Topic](https://review.learn.microsoft.com/en-us/aspnet/core/client-side/import-export-interop?view=aspnetcore-7.0&branch=pr-en-us-27311)

### DIFF
--- a/aspnetcore/client-side/import-export-interop.md
+++ b/aspnetcore/client-side/import-export-interop.md
@@ -53,9 +53,9 @@ dotnet workload install wasm-experimental
 
 For more information, see the [Experimental workload and project templates](#experimental-workload-and-project-templates) section.
 
-## App configuration
+## Project configuration
 
-To configure the app's project file (`.csproj`):
+To configure a project (`.csproj`) to enable JS interop:
 
 * Target `net7.0` or later:
 
@@ -75,9 +75,6 @@ To configure the app's project file (`.csproj`):
   <OutputType>Exe</OutputType>
   ```
   
-  > [!NOTE]
-  > The framework may support library projects without an entry point in a future release of .NET.
-
 * Enable the <xref:Microsoft.Build.Tasks.Csc.AllowUnsafeBlocks> property, which permits the code generator in the Roslyn compiler to use pointers for JS interop:
 
   ```xml
@@ -89,28 +86,12 @@ To configure the app's project file (`.csproj`):
 
   In apps generated from the `wasmbrowser` or `wasmconsole` templates, the <xref:Microsoft.Build.Tasks.Csc.AllowUnsafeBlocks> property is set in the project file (`.csproj`).
 
-* For the .NET 7 release, you must specify `WasmMainJSPath` to point to a file on disk. This file is published with the app, but use of the file isn't required if you're integrating .NET into an existing JS app. We might make this property optional in the future. For more information, see [Smooth out support for running .NET from JS via WebAssembly (dotnet/runtime #77191)](https://github.com/dotnet/runtime/issues/77191).
+* For the .NET 7 release, you must specify `WasmMainJSPath` to point to a file on disk. This file is published with the app, but use of the file isn't required if you're integrating .NET into an existing JS app.
 
   In the following example, the JS file on disk is `main.js`, but any JS filename is permissable:
 
   ```xml
   <WasmMainJSPath>main.js</WasmMainJSPath>
-  ```
-
-* During preview, set the C# language version:
-
-  ```xml
-  <LangVersion>preview</LangVersion>
-  ```
-
-* Deploy JS files (`.js`) and CSS files (`.css`):
-
-  ```xml
-  <ItemGroup>
-    <WasmExtraFilesToDeploy Include="index.html" />
-    <WasmExtraFilesToDeploy Include="*.js" />
-    <WasmExtraFilesToDeploy Include="*.css" />
-  </ItemGroup>
   ```
 
 Example .NET 7 preview release project file (`.csproj`) after configuration:
@@ -124,15 +105,8 @@ Example .NET 7 preview release project file (`.csproj`) after configuration:
     <OutputType>Exe</OutputType>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <WasmMainJSPath>main.js</WasmMainJSPath>
-    <LangVersion>preview</LangVersion>
     <Nullable>enable</Nullable>
   </PropertyGroup>
-
-  <ItemGroup>
-    <WasmExtraFilesToDeploy Include="index.html" />
-    <WasmExtraFilesToDeploy Include="*.js" />
-    <WasmExtraFilesToDeploy Include="*.css" />
-  </ItemGroup>
 
 </Project>
 ```
@@ -150,7 +124,7 @@ APIs in the following example are imported from `dotnet.js`. These APIs enable y
 Function calls in the following example:
 
 * `dotnet.create()` sets up the .NET WebAssembly runtime.
-* `setModuleImports` creates the `window.location.href` function, which returns the current page address (URL). The import designates a module name, which must match the name used with the `JSImportAttribute` (explained later in this article). The following example uses the module name `main.js`, which is merely a convention for the example based on the filename `main.js`. The `window.location.href` function is imported into C# and called by the C# method `GetHRef`. The `GetHRef` method is shown later in this section.
+* `setModuleImports` associates a name with a module of JS functions for import into .NET. In the example, the `main.js` module contains a `window.location.href` function, which returns the current page address (URL). The name of the module can be any string (it doesn't need to be a file name), but it must match the name used with the `JSImportAttribute` (explained later in this article). The `window.location.href` function is imported into C# and called by the C# method `GetHRef`. The `GetHRef` method is shown later in this section.
 * `exports.MyClass.Greeting()` calls into .NET (`MyClass.Greeting`) from JS. The `Greeting` C# method returns a string that includes the result of calling the `window.location.href` function. The `Greeting` method is shown later in this section.
 * `runMainAndExit` runs `Program.Main`.
 
@@ -195,13 +169,12 @@ await runMainAndExit(config.mainAssemblyName, ["dotnet", "is", "great!"]);
 ```
 
 > [!NOTE]
-> You can inspect the .NET host builder APIs at the [`dotnet/runtime` GitHub repository](https://github.com/dotnet/runtime/blob/main/src/mono/wasm/runtime/). The runtime configuration is in the [`dotnet.d.ts` file](https://github.com/dotnet/runtime/blob/main/src/mono/wasm/runtime/dotnet.d.ts).
 >
 > The preceding link to the .NET reference source loads the repository's default branch (`main`), which represents the current development for the next release of .NET. To select a tag for a specific release, use the **Switch branches or tags** dropdown list.
 
 To import a JS function so it can be called from C#, use the new `JSImportAttribute` on a matching method signature. The first parameter to the `JSImportAttribute` is the name of the JS function to import and the second parameter is the name of the module.
 
-In the following example, the `window.location.href` function is called from the `main.js` module when `GetHRef` method is called. In an app generated from either the `wasmbrowser` or `wasmconsole` templates, `GetHRef` is in `Program.cs` with the partial class namespace `MyClass`:
+In the following example, the `window.location.href` function is called from the `main.js` module when `GetHRef` method is called:
 
 ```csharp
 [JSImport("window.location.href", "main.js")]
@@ -268,11 +241,9 @@ The following conditions apply to type mapping and marshalled values:
 * It's not possible to export a .NET method that returns a `Span`. The `Span` is allocated on the call stack and has GC implications. When calling from JS to .NET, there's no C# stack after the call.
 * For an exported method that returns an `ArraySegment`, calling `dispose()` in `try-finally` block disposes the proxy and unpins the underlying C# byte array. We recommend calling `dispose()` on the object in developer JS code. If developer code doesn't dispose of the object, the JS GC eventually disposes the object. You can also marshal a byte array (`byte[]`) instead of an `ArraySegment`, which copies the bytes.
 
-&dagger;The link to the `dotnet.d.ts` file in the .NET reference source loads the repository's default branch (`main`), which represents the current development for the next release of .NET. To select a tag for a specific release, use the **Switch branches or tags** dropdown list.
-
 To export a .NET method so it can be called from JS, use the `JSExportAttribute`.
 
-In the following example, the `Greeting` method returns a string that includes the result of calling the `GetHRef` method. As shown earlier, the `GetHref` C# method calls into JS for the `window.location.href` function from the `main.js` module. `window.location.href` returns the current page address (URL). In an app generated from either the `wasmbrowser` or `wasmconsole` templates, `Greeting` is in the `Program.cs` file with the partial class namespace `MyClass`:
+In the following example, the `Greeting` method returns a string that includes the result of calling the `GetHRef` method. As shown earlier, the `GetHref` C# method calls into JS for the `window.location.href` function from the `main.js` module. `window.location.href` returns the current page address (URL):
 
 ```csharp
 [JSExport]
@@ -308,14 +279,7 @@ To demonstrate the JS interop functionality and obtain JS interop project templa
 dotnet workload install wasm-experimental
 ```
 
-The `wasm-experimental` workload contains two project templates: `wasmbrowser` and `wasmconsole`. These templates are experimental at this time, which means the developer workflow for the templates hasn't been fully designed. For example, these templates don't run in Visual Studio at this time. The .NET and JS APIs used in the templates are supported in .NET 7 and provide a foundation for using .NET on :::no-loc text="WASM"::: from JS.
-
-The templates can be examined in reference source:
-
-[`templates` assets in the `dotnet/runtime` GitHub repository](https://github.com/dotnet/runtime/tree/main/src/mono/wasm/templates/templates).
-
-> [!NOTE]
-> The preceding link to the .NET reference source loads the repository's default branch (`main`), which represents the current development for the next release of .NET. To select a tag for a specific release, use the **Switch branches or tags** dropdown list.
+The `wasm-experimental` workload contains two project templates: `wasmbrowser` and `wasmconsole`. These templates are experimental at this time, which means the developer workflow for the templates is still evolving. But the .NET and JS APIs used in the templates are supported in .NET 7 and provide a foundation for using .NET on :::no-loc text="WASM"::: from JS.
 
 ### Browser app
 
--- a/aspnetcore/client-side/import-export-interop.md
+++ b/aspnetcore/client-side/import-export-interop.md
@@ -356,4 +356,7 @@ In the preceding example, the `{TARGET FRAMEWORK}` placeholder is the target fra
 
 -->
 
+* In the `dotnet/runtime` GitHub repository:
+  * [.NET host builder API](https://github.com/dotnet/runtime/blob/main/src/mono/wasm/runtime/)
+  * [`dotnet.d.ts` file (runtime configuration)](https://github.com/dotnet/runtime/blob/main/src/mono/wasm/runtime/dotnet.d.ts)
 * [Use .NET from any JavaScript app in .NET 7](https://devblogs.microsoft.com/dotnet/use-net-7-from-any-javascript-app-in-net-7/)
--- a/aspnetcore/client-side/import-export-interop.md
+++ b/aspnetcore/client-side/import-export-interop.md
@@ -1,0 +1,271 @@
+---
+title: Run .NET from JavaScript using `[JSImport]`/`[JSExport]` interop
+author: pavelsavara
+description: Learn how to run .NET from JavaScript using `[JSImport]`/`[JSExport]` interop.
+monikerRange: '>= aspnetcore-7.0'
+ms.author: riande
+ms.custom: mvc
+ms.date: 10/18/2022
+uid: client-side/import-export-interop
+---
+# Run .NET from JavaScript using `[JSImport]`/`[JSExport]` interop
+
+This article explains how to run .NET from JavaScript (JS) using `[JSImport]`/`[JSExport]` interop.
+
+Existing JS apps can use the expanded client-side WebAssembly support in .NET 7 to reuse .NET libraries from JS or to build novel .NET-based apps and frameworks.
+
+> [!NOTE]
+> This article focuses on running .NET from JS apps without any dependency on [Blazor](xref:blazor/index). 
+
+<!--
+
+    HOLD: ADD TO NOTE AFTER BLAZOR ARTICLE IS MERGED ...
+
+    For guidance on using `[JSImport]`/`[JSExport]` interop in Blazor WebAssembly apps, see <xref:blazor/js-interop/import-export-interop>.
+
+-->
+
+These approaches are appropriate when you only expect to run on WebAssembly (:::no-loc text="WASM":::). Libraries can make a runtime check to determine if the app is running on :::no-loc text="WASM"::: by calling <xref:System.OperatingSystem.IsBrowser%2A?displayProperty=nameWithType>.
+
+## Prerequisites
+
+[!INCLUDE[](~/includes/7.0-SDK.md)]
+
+## Namespace
+
+The JS interop API described in this article is controlled by attributes in the <xref:System.Runtime.InteropServices.JavaScript?displayProperty=fullName> namespace.
+
+## Enable unsafe blocks
+
+Enable the <xref:Microsoft.Build.Tasks.Csc.AllowUnsafeBlocks> property in app's project file, which permits the code generator in the Roslyn compiler to use pointers for JS interop:
+
+```xml
+<PropertyGroup>
+  <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+</PropertyGroup>
+```
+
+> [!WARNING]
+> The JS interop API requires enabling <xref:Microsoft.Build.Tasks.Csc.AllowUnsafeBlocks>. Be careful when implementing your own unsafe code in .NET apps, which can introduce security and stability risks. For more information, see [Unsafe code, pointer types, and function pointers](/dotnet/csharp/language-reference/unsafe-code).
+
+In apps generated from the `wasmbrowser` or `wasmconsole` templates, the <xref:Microsoft.Build.Tasks.Csc.AllowUnsafeBlocks> property is set in the project file (`.csproj`).
+
+## .NET JavaScript interop on :::no-loc text="WASM":::
+
+The [JS ES6 module](xref:blazor/js-interop/index#javascript-isolation-in-javascript-modules) in `main.js` in a project created from the `wasmbrowser` project template demonstrates JS interop. The relevant APIs are imported from `dotnet.js`. These APIs enable you to set up named modules that can be imported into your C# code and call into methods exposed by your .NET code, including `Program.Main`.
+
+Function calls in the following example:
+
+* `dotnet.withDiagnosticTracing(false).withApplicationArgumentsFromQuery().create()` sets up the .NET WebAssembly runtime without diagnostic tracing and with query string arguments.
+* `setModuleImports` creates the `window.location.href` function, which returns the current page address (URL). The import designates a module name that matches the JS file name, `main.js`. The `window.location.href` function is imported into C# and called by the C# method `GetHRef`. The `GetHRef` method is shown later in this section.
+* `exports.MyClass.Greeting()` calls into .NET (`MyClass.Greeting`) from JS. The `Greeting` C# method returns a string that includes the result of calling the `window.location.href` function. The `Greeting` method is shown later in this section.
+* `runMainAndExit` runs `Program.Main`.
+
+<!--
+
+    NOTE: THE FOLLOWING CODE IS CHANGING FOR RELEASE:
+
+    https://github.com/dotnet/runtime/blob/release/7.0/src/mono/wasm/templates/templates/browser/main.js
+
+    For example, the runMainAndExit() is changing to dotnet.run().
+
+    This is tracked on the 7.0 doc tracking issue at https://github.com/dotnet/AspNetCore.Docs/issues/26364.
+    
+-->
+
+`main.js` of an app created from the `wasmbrowser` project template:
+
+```javascript
+import { dotnet } from './dotnet.js'
+
+const is_browser = typeof window != "undefined";
+if (!is_browser) throw new Error(`Expected to be running in a browser`);
+
+const { setModuleImports, getAssemblyExports, getConfig, runMainAndExit } = 
+  await dotnet
+    .withDiagnosticTracing(false)
+    .withApplicationArgumentsFromQuery()
+    .create();
+
+setModuleImports("main.js", {
+  window: {
+    location: {
+      href: () => globalThis.window.location.href
+    }
+  }
+});
+
+const config = getConfig();
+const exports = await getAssemblyExports(config.mainAssemblyName);
+const text = exports.MyClass.Greeting();
+console.log(text);
+
+document.getElementById("out").innerHTML = `${text}`;
+await runMainAndExit(config.mainAssemblyName, ["dotnet", "is", "great!"]);
+```
+
+To import a JS function so it can be called from C#, use the new `JSImportAttribute` on a matching method signature. The first parameter to the `JSImportAttribute` is the name of the JS function to import and the second parameter is the name of the module.
+
+In the following example, the `window.location.href` function is called from the `main.js` module when `GetHRef` method is called. In an app generated from either the `wasmbrowser` or `wasmconsole` templates, `GetHRef` is in `Program.cs` with the partial class namespace `MyClass`:
+
+```csharp
+[JSImport("window.location.href", "main.js")]
+internal static partial string GetHRef();
+```
+
+In the imported method signature, you can use .NET types for parameters and return values, which are marshalled automatically by the runtime. Use `JSMarshalAsAttribute<T>` to control how the imported method parameters are marshalled. For example, you might choose to marshal a `long` as <xref:System.Runtime.InteropServices.JavaScript.JSType.Number?displayProperty=nameWithType> or <xref:System.Runtime.InteropServices.JavaScript.JSType.BigInt?displayProperty=nameWithType>. You can pass <xref:System.Action>/<xref:System.Func%601> callbacks as parameters, which are marshalled as callable JS functions. You can pass both JS and managed object references, and they are marshaled as proxy objects, keeping the object alive across the boundary until the proxy is garbage collected. You can also import and export asynchronous methods with a <xref:System.Threading.Tasks.Task> result, which are marshaled as [JS promises](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Promise). Most of the marshalled types work in both directions, as parameters and as return values, on both imported and exported methods.
+
+To export a .NET method so it can be called from JS, use the `JSExportAttribute`.
+
+In the following example, the `Greeting` method returns a string that includes the result of calling the `GetHRef` method. As shown earlier, the `GetHref` C# method calls into JS for the `window.location.href` function from the `main.js` module. `window.location.href` returns the current page address (URL). In an app generated from either the `wasmbrowser` or `wasmconsole` templates, `Greeting` is in the `Program.cs` file with the partial class namespace `MyClass`:
+
+```csharp
+[JSExport]
+internal static string Greeting()
+{
+    var text = $"Hello, World! Greetings from {GetHRef()}";
+    Console.WriteLine(text);
+    return text;
+}
+```
+
+In the `index.html` file of an app generated from the `wasmbrowser` template, the `main.js` and `dotnet.js` modules are preloaded. The `<script>` tag for `main.js` appears inside the closing `</body>` tag, resulting the following:
+
+* `main.js` is loaded.
+* `dotnet.js` is imported by `main.js`.
+* The .NET runtime is started by calling `runMainAndExit()` in `main.js`.
+
+<!--
+
+    NOTE: In the preceding list, change runMainAndExit() to dotnet.run() for release.
+
+-->
+
+```html
+<html>
+
+<head>
+    ...
+    <link rel="modulepreload" href="./main.js" />
+    <link rel="modulepreload" href="./dotnet.js" />
+</head>
+
+<body>
+    ...
+    <script type='module' src="./main.js"></script>
+</body>
+
+</html>
+```
+
+In addition to preloading, it's possible to prefetch the largest binary files. In a production app, measure the impact of preloading and prefetching optimizations. For more information, see [Use .NET from any JavaScript app in .NET 7: Optimize the app](https://devblogs.microsoft.com/dotnet/use-net-7-from-any-javascript-app-in-net-7/#optimize-the-app).
+
+## Additional example
+
+For an additional example of the JS interop techniques described in this article, see the *Port of famous Todo-MVC to .NET on WASM* sample app:
+
+* [Reference source (`pavelsavara/dotnet-wasm-todo-mvc` GitHub repository)](https://github.com/pavelsavara/dotnet-wasm-todo-mvc)
+* [Live demonstration](https://pavelsavara.github.io/dotnet-wasm-todo-mvc/)
+
+> [!NOTE]
+> The [`pavelsavara/dotnet-wasm-todo-mvc` GitHub repository](https://github.com/pavelsavara/dotnet-wasm-todo-mvc) isn't owned, maintained, or supported by the .NET foundation or Microsoft.
+>
+> The *Port of famous Todo-MVC to .NET on WASM* sample app uses:
+>
+> * [TodoMVC](https://todomvc.com/)
+> * [Playwright for .NET](https://playwright.dev/dotnet/)
+>
+> The preceding frameworks aren't owned, maintained, or supported by the .NET foundation or Microsoft.
+
+## Experimental workload and project templates
+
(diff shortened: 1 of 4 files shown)


aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,271 @@
...
+
+    HOLD: ADD TO NOTE AFTER BLAZOR ARTICLE IS MERGED ...
+
+    For guidance on using `[JSImport]`/`[JSExport]` interop in Blazor WebAssembly apps, see <xref:blazor/js-interop/import-export-interop>.
+
+-->
+
+These approaches are appropriate when you only expect to run on WebAssembly (:::no-loc text="WASM":::). Libraries can make a runtime check to determine if the app is running on :::no-loc text="WASM"::: by calling <xref:System.OperatingSystem.IsBrowser%2A?displayProperty=nameWithType>.
+
+## Prerequisites
```
danroth27, line 30: To use .NET from JS using this model you need to:

- Install the `wasm-tools` workload, which brings in the related MSBuild targets (you only need the `wasm-experimental` workload if you want to use the project templates)
- Target `net7.0` or later
- Specify `browser-wasm` for the runtime identifier: `<RuntimeIdentifier>browser-wasm</RuntimeIdentifier>`
- Specify `<OutputType>Exe</OutputType>` (we may support library projects without an entry point later)
- Enable `<AllowUnsafeBlocks>true</AllowUnsafeBlocks>` (you've already got this covered below, but I'm listing it here for completeness)

Currently, you also need to specify `WasmMainJSPath` to point to some file on disk. This file will get published with the app, but it doesn't actually have to be used if you're going to integrate .NET into an existing JS app. We might make this property optional in the future: https://github.com/dotnet/runtime/issues/77191.

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,271 @@
...
+
+In apps generated from the `wasmbrowser` or `wasmconsole` templates, the <xref:Microsoft.Build.Tasks.Csc.AllowUnsafeBlocks> property is set in the project file (`.csproj`).
+
+## .NET JavaScript interop on :::no-loc text="WASM":::
+
+The [JS ES6 module](xref:blazor/js-interop/index#javascript-isolation-in-javascript-modules) in `main.js` in a project created from the `wasmbrowser` project template demonstrates JS interop. The relevant APIs are imported from `dotnet.js`. These APIs enable you to set up named modules that can be imported into your C# code and call into methods exposed by your .NET code, including `Program.Main`.
+
+Function calls in the following example:
+
+* `dotnet.withDiagnosticTracing(false).withApplicationArgumentsFromQuery().create()` sets up the .NET WebAssembly runtime without diagnostic tracing and with query string arguments.
```
danroth27, line 59: We could probably simplify this to just `dotnet.create()` and then document the other DotnetHostBuilder APIs separately. See the full API definiton here: https://github.com/dotnet/runtime/blob/main/src/mono/wasm/runtime/dotnet.d.ts.

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,271 @@
...
+description: Learn how to run .NET from JavaScript using `[JSImport]`/`[JSExport]` interop.
+monikerRange: '>= aspnetcore-7.0'
+ms.author: riande
+ms.custom: mvc
+ms.date: 10/18/2022
+uid: client-side/import-export-interop
+---
+# Run .NET from JavaScript using `[JSImport]`/`[JSExport]` interop
+
+This article explains how to run .NET from JavaScript (JS) using `[JSImport]`/`[JSExport]` interop.
```
danroth27, line 13: It might be good to clarify that "import" and "export" here are defined from the perspective of .NET. You import JS methods so they can be called from .NET and you export .NET methods so they can be called from JS.

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,271 @@
...
+
+<body>
+    ...
+    <script type='module' src="./main.js"></script>
+</body>
+
+</html>
+```
+
+In addition to preloading, it's possible to prefetch the largest binary files. In a production app, measure the impact of preloading and prefetching optimizations. For more information, see [Use .NET from any JavaScript app in .NET 7: Optimize the app](https://devblogs.microsoft.com/dotnet/use-net-7-from-any-javascript-app-in-net-7/#optimize-the-app).
```
danroth27, line 161: This section is pretty specific to how the templates are setup. For most users I expect they will have their own host HTML file and they can decide how they want to optimize how the app loads. I don't think we really need to go through the specific HTML file from the project template.

However, we do need to document where to find *dotnet.js* so that users can import it. This file will be generated as part of the build output of the app in the AppBundle folder. To integrate with an existing app, the user should copy the contents of the AppBundle folder so that it can be served along with the rest of the app. For production deployments, they should do a `dotnet publish -c Release` and deploy the AppBundle folder contents from the publish output.

```suggestion
```

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,271 @@
...
+internal static partial string GetHRef();
+```
+
+In the imported method signature, you can use .NET types for parameters and return values, which are marshalled automatically by the runtime. Use `JSMarshalAsAttribute<T>` to control how the imported method parameters are marshalled. For example, you might choose to marshal a `long` as <xref:System.Runtime.InteropServices.JavaScript.JSType.Number?displayProperty=nameWithType> or <xref:System.Runtime.InteropServices.JavaScript.JSType.BigInt?displayProperty=nameWithType>. You can pass <xref:System.Action>/<xref:System.Func%601> callbacks as parameters, which are marshalled as callable JS functions. You can pass both JS and managed object references, and they are marshaled as proxy objects, keeping the object alive across the boundary until the proxy is garbage collected. You can also import and export asynchronous methods with a <xref:System.Threading.Tasks.Task> result, which are marshaled as [JS promises](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Promise). Most of the marshalled types work in both directions, as parameters and as return values, on both imported and exported methods.
```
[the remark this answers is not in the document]
maraf: Here a table of supported mappings

| .NET                            | JavaScript | Nullable      | Task -> Promise | JSType.Missing | Array of |
|---------------------------------|------------|---------------|-----------------|----------------|----------|
| Boolean                         | Boolean    | ✅             | ✅               | ✅              |          |
| Byte                            | Number     | ✅             | ✅               | ✅              | ✅        |
| Char                            | String     | ✅             | ✅               | ✅              |          |
| Int16                           | Number     | ✅             | ✅               | ✅              |          |
| Int32                           | Number     | ✅             | ✅               | ✅              | ✅        |
| Int64                           | Number     | ✅             | ✅               |                |          |
| Int64                           | BigInt     | ✅             | ✅               |                |          |
| Single                          | Number     | ✅             | ✅               | ✅              |          |
| Double                          | Number     | ✅             | ✅               | ✅              | ✅        |
| IntPtr                          | Number     | ✅             | ✅               | ✅              |          |
| DateTime                        | Date       | ✅             | ✅               |                |          |
| DateTimeOffset                  | Date       | ✅             | ✅               |                | ✅        |
| Exception                       | Error      |               |                 | ✅              |          |
| JSObject                        | Object     |               |                 | ✅              | ✅        |
| String                          | String     |               |                 | ✅              |          |
| Object                          | Any        |               | ✅               |                | ✅        |
| Span&lt;Byte&gt;                | MemoryView |               |                 |                |          |
| Span&lt;Int32&gt;               | MemoryView |               |                 |                |          |
| Span&lt;Double&gt;              | MemoryView |               |                 |                |          |
| ArraySegment&lt;Byte&gt;        | MemoryView |               |                 |                |          |
| ArraySegment&lt;Int32&gt;       | MemoryView |               |                 |                |          |
| ArraySegment&lt;Double&gt;      | MemoryView |               |                 |                |          |
| Task                            | Promise    |               |                 |                |          |
| Action                          | Function   |               |                 |                |          |
| Action&lt;T1&gt;                | Function   |               |                 |                |          |
| Action&lt;T1, T2&gt;            | Function   |               |                 |                |          |
| Action&lt;T1, T2, T3&gt;        | Function   |               |                 |                |          |
| Func&lt;TResult&gt;             | Function   |               |                 |                |          |
| Func&lt;T1, TResult&gt;         | Function   |               |                 |                |          |
| Func&lt;T1, T2, TResult&gt;     | Function   |               |                 |                |          |
| Func&lt;T1, T2, T3, TResult&gt; | Function   |               |                 |                |          |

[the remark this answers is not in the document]
pavelsavara: @danroth27 I told you I'm not the only one who knows about this stuff :-D

Few more notes
 - `JSType.Missing` means that the `JSMarshalAs<>` annotation is optional. 
     - @guardrex you can figure out better name for that column.
     - I think it's also optional for void `Task`, there shuld be checkmark too.
 - The `Array of` is C# `string[]` mapped to JS Array full of strings. 
     - It will copy the array and marshal the values.
     - `DateTimeOffset` and `JSObject` should not have check 
     - `String` should have a check
 - when you are passing JS value to C# side and you give value of wrong type, it will throw in most cases. When running, we don't have compile time type checker in JS.
 - `JSObject`, `Exception`, `Task` and `ArraySegment` create GCHandle and a proxy. Could be disposed by user or GC would handle it later. They are not cheap.

@maraf could you please edit the grid ? Somehow I could not do it.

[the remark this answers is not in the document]
pavelsavara: `MemoryView` is a type which dotnet brings, not a native JS type. It's API is described in `dotnet.d.ts`
It doesn't copy the bytes during marshaling.
`MemoryView` created for `Span` is only valid for the duration of the interop call.
`MemoryView` created for `ArraySegment` is good for long term sharing of a buffer. It will survive after the interop call.
You can't create instances of `MemoryView` on JS side yourself. So marshaling them in this direction doesn't make sense.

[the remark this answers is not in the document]
guardrex: Can I "dagger" (&dagger;) the "`JSType.Missing`" column name (`JSType.Missing`&dagger;); and then under the table, dagger the remark like this ...

&dagger;`JSType.Missing` means that the `JSMarshalAs<T>` annotation is optional.

If that's ok, it keeps the column ***narrow***. On small device screens, we have trouble displaying wide tables. It's best to keep them as narrow as possible.

pavelsavara:
cc @radical

aspnetcore/client-side/import-export-interop.md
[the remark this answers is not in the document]
maraf: I have updated the table.

> `DateTimeOffset` and `JSObject` should not have check

The `DateTimeOffset` was a mistake, but `JSObject` is there https://github.com/maraf/runtime/blob/99d21b9276ebe8f7bea7fb3ba74dca9fca625fe2/src/libraries/System.Runtime.InteropServices.JavaScript/gen/JSImportGenerator/JSGeneratorFactory.cs#L213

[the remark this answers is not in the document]
pavelsavara: `Task<string>` 
`Task<Exception>`
`Task<JSObject>`
`Task<object>` are all supported and working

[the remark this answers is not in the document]
pavelsavara: Also for `Task<string>`  `Task<Exception>` `Task<JSObject>` the annotation is optional.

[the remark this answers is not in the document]
maraf: Yes, already marked in the table.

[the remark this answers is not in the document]
pavelsavara: @guardrex maybe we want to talk about "JSMarshalAs optional" rather as "these cases require JSMarshalAs annotation". I guess people would not annotate (and care about it) unless it's required.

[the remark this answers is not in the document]
FranklinWhale: > `MemoryView` is a type which dotnet brings, not a native JS type. It's API is described in `dotnet.d.ts` It doesn't copy the bytes during marshaling. `MemoryView` created for `Span` is only valid for the duration of the interop call. `MemoryView` created for `ArraySegment` is good for long term sharing of a buffer. It will survive after the interop call. You can't create instances of `MemoryView` on JS side yourself. So marshaling them back to C# doesn't make sense.

Does it mean that:

1. It is not practical (or not possible) to export a .NET method that returns `Span`?
2. For exported method that returns `ArraySegment`, is it important to call `dispose()` in the returned `MemoryView` in `try... finally`?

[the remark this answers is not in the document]
pavelsavara: @FranklinWhale 
1) it's not possible. The Span is allocated on stack and it has GC implications. When calling from JS to C#, there is no C# stack after the call.
2) calling `dispose()` disposes the proxy and unpins the underlying C# byte array. It is good idea to do so. If you don't do it, the JS GC will do it eventually.
3) you can also return `byte[]` which will copy the bytes, but it's easiest to think about it.

guardrex:
I took into account all of the feedback. Check the *language translations*  ... .NET guru to *HackRex* 🦖 ... for errors. I may have inadvertently broken a few concepts in the process. 🙈

The table will get WCAG updates later ... immediately prior to merging this. For now, we can look at the simple layout on the PR to troubleshoot it further. If I add the WCAG bits now, it will be much harder to read on the diff and update.

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+
+## Prerequisites
+
+[!INCLUDE[](~/includes/7.0-SDK.md)]
+
+## Namespace
+
+The JS interop API described in this article is controlled by attributes in the <xref:System.Runtime.InteropServices.JavaScript?displayProperty=fullName> namespace.
+
+## .NET SDK and WebAssembly workloads
```
danroth27, line 38: Could we just include the workloads in the Prerequisites section instead of in a separate section?

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+
+Optionally, install the `wasm-experimental` workload, which contains experimental project templates for getting started with .NET on WebAssembly in a browser app (WebAssembly Browser App) or in a Node.js-based console app (WebAssembly Console App). This workload isn't required if you plan to integrate `[JSImport]`/`[JSExport]` interop into an existing JS app.
+
+```dotnetcli
+dotnet workload install wasm-experimental
+```
+
+For more information, see the [Experimental workload and project templates](#experimental-workload-and-project-templates) section.
+
+## App configuration
```
danroth27, line 56: I think "Project" is more appropriate here than app because the app is probably a separate JS app.

```suggestion
## Project configuration
```

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+
+To configure the app's project file (`.csproj`):
```
danroth27, line 58: I think we should specify what we're configuring the project to do.

```suggestion
To configure a project (`.csproj`) to enable JS interop:
```

Feel free to adjust this language for consistency.

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+
+* Specify an executable output type:
+
+  ```xml
+  <OutputType>Exe</OutputType>
+  ```
+  
+  > [!NOTE]
+  > The framework may support library projects without an entry point in a future release of .NET.
+
```
danroth27, line 80: This note was just an FYI to you. I prefer to avoid promising future features when the future can always change.

```suggestion
```

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+* Deploy JS files (`.js`) and CSS files (`.css`):
+
+  ```xml
+  <ItemGroup>
+    <WasmExtraFilesToDeploy Include="index.html" />
+    <WasmExtraFilesToDeploy Include="*.js" />
+    <WasmExtraFilesToDeploy Include="*.css" />
+  </ItemGroup>
+  ```
+
```
danroth27, line 115: This isn't technically required and the `WasmExtraFilesToDeploy` item group may change in the future, so for now I prefer to leave it undocumented for now.

```suggestion
```

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <WasmExtraFilesToDeploy Include="index.html" />
+    <WasmExtraFilesToDeploy Include="*.js" />
+    <WasmExtraFilesToDeploy Include="*.css" />
+  </ItemGroup>
+
+</Project>
```
danroth27, line 137:
```suggestion
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net7.0</TargetFramework>
    <RuntimeIdentifier>browser-wasm</RuntimeIdentifier>
    <OutputType>Exe</OutputType>
    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
    <WasmMainJSPath>main.js</WasmMainJSPath>
    <Nullable>enable</Nullable>
  </PropertyGroup>

</Project>
```

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+  ```xml
+  <WasmMainJSPath>main.js</WasmMainJSPath>
+  ```
+
+* During preview, set the C# language version:
+
+  ```xml
+  <LangVersion>preview</LangVersion>
+  ```
+
```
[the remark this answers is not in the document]
danroth27: Maybe this is only an issue if you try to use a stable VS version with a .NET 7 preview? In general .NET 7 requires VS 17.4.

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+</Project>
+```
+
+## .NET JavaScript interop on :::no-loc text="WASM":::
+
+APIs in the following example are imported from `dotnet.js`. These APIs enable you to set up named modules that can be imported into your C# code and call into methods exposed by your .NET code, including `Program.Main`.
+
+> [!IMPORTANT]
+> "Import" and "export" throughout this article are defined from the perspective of .NET:
```
danroth27, line 145: :heart:

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+>
+> * An app imports JS methods so that they can be called from .NET.
+> * The app exports .NET methods so that they can be called from JS.
+
+Function calls in the following example:
+
+* `dotnet.create()` sets up the .NET WebAssembly runtime.
+* `setModuleImports` creates the `window.location.href` function, which returns the current page address (URL). The import designates a module name, which must match the name used with the `JSImportAttribute` (explained later in this article). The following example uses the module name `main.js`, which is merely a convention for the example based on the filename `main.js`. The `window.location.href` function is imported into C# and called by the C# method `GetHRef`. The `GetHRef` method is shown later in this section.
```
danroth27, line 153:
```suggestion
* `setModuleImports` associates a name with a module of JS functions that can then be imported into .NET. In the example, the `main.js` module contains a `window.location.href` function, which returns the current page address (URL). The name of the module can be any string (it doesn't need to be a file name), but it must match the name used with the `JSImportAttribute` (explained later in this article). The `window.location.href` function is imported into C# and called by the C# method `GetHRef`. The `GetHRef` method is shown later in this section.
```

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+const exports = await getAssemblyExports(config.mainAssemblyName);
+const text = exports.MyClass.Greeting();
+console.log(text);
+
+document.getElementById("out").innerHTML = `${text}`;
+await runMainAndExit(config.mainAssemblyName, ["dotnet", "is", "great!"]);
+```
+
+> [!NOTE]
+> You can inspect the .NET host builder APIs at the [`dotnet/runtime` GitHub repository](https://github.com/dotnet/runtime/blob/main/src/mono/wasm/runtime/). The runtime configuration is in the [`dotnet.d.ts` file](https://github.com/dotnet/runtime/blob/main/src/mono/wasm/runtime/dotnet.d.ts).
```
danroth27, line 198: How about we just link to dotnet.d.ts in the Additional resources section at the end of this article?
```suggestion
```

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+>
+> The preceding link to the .NET reference source loads the repository's default branch (`main`), which represents the current development for the next release of .NET. To select a tag for a specific release, use the **Switch branches or tags** dropdown list.
+
+To import a JS function so it can be called from C#, use the new `JSImportAttribute` on a matching method signature. The first parameter to the `JSImportAttribute` is the name of the JS function to import and the second parameter is the name of the module.
+
+In the following example, the `window.location.href` function is called from the `main.js` module when `GetHRef` method is called. In an app generated from either the `wasmbrowser` or `wasmconsole` templates, `GetHRef` is in `Program.cs` with the partial class namespace `MyClass`:
```
danroth27, line 204: We haven't really introduced the project templates yet, so it feels a bit out of place to reference them here.

```suggestion
In the following example, the `window.location.href` function is called from the `main.js` module when `GetHRef` method is called:
```

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+
+<!--
+    NOTE
+
+    I'll leave it in this plain format for reviews, which is easier to inspect.
+    Before merging, I'll make the table WCAG compliant. 
+
+-->
+
+The following table indicates the supported type mappings.
```
danroth27, line 221: :+1:

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+  * `MemoryView` created for an `ArraySegment` survives after the interop call and is useful for sharing a buffer.
+  * `MemoryView` doesn't have an analogous JS type, so marshalling a JS object to a .NET `MemoryView` isn't possible.
+* It's not possible to export a .NET method that returns a `Span`. The `Span` is allocated on the call stack and has GC implications. When calling from JS to .NET, there's no C# stack after the call.
+* For an exported method that returns an `ArraySegment`, calling `dispose()` in `try-finally` block disposes the proxy and unpins the underlying C# byte array. We recommend calling `dispose()` on the object in developer JS code. If developer code doesn't dispose of the object, the JS GC eventually disposes the object. You can also marshal a byte array (`byte[]`) instead of an `ArraySegment`, which copies the bytes.
+
+&dagger;The link to the `dotnet.d.ts` file in the .NET reference source loads the repository's default branch (`main`), which represents the current development for the next release of .NET. To select a tag for a specific release, use the **Switch branches or tags** dropdown list.
+
```
danroth27, line 272: Since MemoryView isn't mentioned in dotnet.d.ts, it doesn't look like this note is actually needed.

```suggestion
```

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+| `Func<T1, T2, T3, TResult>` | `Function`   |            |                   |                        | |
+
+The following conditions apply to type mapping and marshalled values:
+
+* The :::no-loc text="Array of"::: column indicates if the .NET type can be marshalled as a JS [`Array`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array). Example: C# `int[]` (`Int32`) mapped to JS `Array` of `Number`s.
+* When passing a JS value to C# with a value of the wrong type, the framework throws an exception in most cases. The framework doesn't perform compile-time type checking in JS.
+* `JSObject`, `Exception`, `Task` and `ArraySegment` create `GCHandle` and a proxy. You can trigger disposal in developer code or allow [.NET garbage collection (GC)](/dotnet/standard/garbage-collection/) to dispose of the objects later. These types carry significant performance overhead.
+* `MemoryView`
+  * `MemoryView` is a .NET type, not a native JS type. It's API is described in the [`dotnet.d.ts` file (`dotnet/runtime` GitHub repository)&dagger;](https://github.com/dotnet/runtime/blob/main/src/mono/wasm/runtime/dotnet.d.ts).
```
danroth27, line 263: I don't see any mention of MemoryView in dotnet.d.ts. I do see this: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Runtime.InteropServices.JavaScript/src/System/Runtime/InteropServices/JavaScript/JSType.cs#L90

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+
+&dagger;The link to the `dotnet.d.ts` file in the .NET reference source loads the repository's default branch (`main`), which represents the current development for the next release of .NET. To select a tag for a specific release, use the **Switch branches or tags** dropdown list.
+
+To export a .NET method so it can be called from JS, use the `JSExportAttribute`.
+
+In the following example, the `Greeting` method returns a string that includes the result of calling the `GetHRef` method. As shown earlier, the `GetHref` C# method calls into JS for the `window.location.href` function from the `main.js` module. `window.location.href` returns the current page address (URL). In an app generated from either the `wasmbrowser` or `wasmconsole` templates, `Greeting` is in the `Program.cs` file with the partial class namespace `MyClass`:
```
danroth27, line 275:
```suggestion
In the following example, the `Greeting` method returns a string that includes the result of calling the `GetHRef` method. As shown earlier, the `GetHref` C# method calls into JS for the `window.location.href` function from the `main.js` module. `window.location.href` returns the current page address (URL):
```

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+
+## Experimental workload and project templates
+
+To demonstrate the JS interop functionality and obtain JS interop project templates, install the `wasm-experimental` workload:
+
+```dotnetcli
+dotnet workload install wasm-experimental
+```
+
+The `wasm-experimental` workload contains two project templates: `wasmbrowser` and `wasmconsole`. These templates are experimental at this time, which means the developer workflow for the templates hasn't been fully designed. For example, these templates don't run in Visual Studio at this time. The .NET and JS APIs used in the templates are supported in .NET 7 and provide a foundation for using .NET on :::no-loc text="WASM"::: from JS.
```
danroth27, line 311: The templates should run now.

```suggestion
The `wasm-experimental` workload contains two project templates: `wasmbrowser` and `wasmconsole`. These templates are experimental at this time, which means the developer workflow for the templates is still evolving. But the .NET and JS APIs used in the templates are supported in .NET 7 and provide a foundation for using .NET on :::no-loc text="WASM"::: from JS.
```

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+
+The templates can be examined in reference source:
+
+[`templates` assets in the `dotnet/runtime` GitHub repository](https://github.com/dotnet/runtime/tree/main/src/mono/wasm/templates/templates).
+
+> [!NOTE]
+> The preceding link to the .NET reference source loads the repository's default branch (`main`), which represents the current development for the next release of .NET. To select a tag for a specific release, use the **Switch branches or tags** dropdown list.
+
```
danroth27, line 319: Instead of trying to read the template source on GitHub the user can just use the templates to create a project.

```suggestion
```

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,394 @@
...
+You can create a browser app by running the following command:
+
+```dotnetcli
+dotnet new wasmbrowser
+```
+
+The `wasmbrowser` template creates a web app that demonstrates using .NET and JS together in a browser.
+
+Build the app from Visual Studio or by using the .NET CLI passing the [`-c/--configuration`](/dotnet/core/tools/dotnet-build#options) build configuration option:
```
danroth27, line 330: These templates now build and run without the need to specify a specific build configuration. I think we can simplify this section by just telling folks to run `dotnet build` and `dotnet run` and then assuming the default build configuration is `Debug`.

danroth27 (changes requested):
Looking really good! Just a few more changes.

aspnetcore/client-side/import-export-interop.md
FranklinWhale: I think it's https://github.com/dotnet/runtime/blob/release/7.0/src/mono/wasm/runtime/marshal.ts#L379

pavelsavara: Perhaps we should speak about `IMemoryView `
https://github.com/dotnet/runtime/blob/main/src/mono/wasm/runtime/marshal.ts#L436

@maraf do you remember if we had it in the `dotnet.d.ts` ? Should we add it ?

guardrex: I'm going to cross-link to `IMemoryView` (without the line number tho ... that's too fragile for us). I'm refactoring the prior remark about selecting the ref source branch for a given release. To harden the doc, it's best for us to link to `main` and let the dev find their version.

guardrex:
***Updates made.***

***NOTE TO SELF***: 🎗️ *Don't forget to WCAG the table before merging!* 🎗️

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,348 @@
...
+| `Func<T1, T2, TResult>`     | `Function`   |            |                   |                        | |
+| `Func<T1, T2, T3, TResult>` | `Function`   |            |                   |                        | |
+
+The following conditions apply to type mapping and marshalled values:
+
+* The :::no-loc text="Array of"::: column indicates if the .NET type can be marshalled as a JS [`Array`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array). Example: C# `int[]` (`Int32`) mapped to JS `Array` of `Number`s.
+* When passing a JS value to C# with a value of the wrong type, the framework throws an exception in most cases. The framework doesn't perform compile-time type checking in JS.
+* `JSObject`, `Exception`, `Task` and `ArraySegment` create `GCHandle` and a proxy. You can trigger disposal in developer code or allow [.NET garbage collection (GC)](/dotnet/standard/garbage-collection/) to dispose of the objects later. These types carry significant performance overhead.
+* [`IMemoryView`](https://github.com/dotnet/runtime/blob/main/src/mono/wasm/runtime/marshal.ts)
```
danroth27, line 219: This is a link to the `IMemoryView` interface in TS, but the table above says `MemoryView`, which seems inconsistent.

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,348 @@
...
+  * `MemoryView` is a .NET type, not a native JS type.
```
danroth27, line 220: It's a bit weird to say `MemoryView` is a .NET type, when think we're actually talking about a JS data structure, correct?

aspnetcore/client-side/import-export-interop.md
danroth27: @pavelsavara Since `IMemoryView` isn't defined in dotnet.d.ts, should we document its structure here? It seems potentially problematic to refer to the   marshal.ts implementation.

aspnetcore/client-side/import-export-interop.md
pavelsavara: It is JavaScript "type" because they mostly don't have "types". It's a shape of a JS object which .NET consistently creates. I don't know how to say it nicely. "not a browser type." would be better

aspnetcore/client-side/import-export-interop.md
pavelsavara: Agreed we should not link the source. Let not link anything until we fix the dotnet.d.ts

aspnetcore/client-side/import-export-interop.md
guardrex: I can't grok what the sentence should state from that discussion. Can you provide the full sentence that you'd like to see there ...

> `MemoryView` is a ...

aspnetcore/client-side/import-export-interop.md
guardrex: I'll remove the link and go back to `MemoryView` until further notice.

aspnetcore/client-side/import-export-interop.md
pavelsavara: Sorry.

`IMemoryView` is a JS type of .NET interop, not a primitive JS type.

guardrex:
***Updates made.***

* I changed `MemoryView` to `IMemoryView` in the table and down the bulleted list.
* I updated the definition of `IMemoryView` and removed the cross-link to `marshal.ts`.
* I commented-out the .NET ref source remark just in case we need it later. If we end up not needing it, I'll 🔪 it.

FranklinWhale:
> _**Updates made.**_
> 
> * I changed `MemoryView` to `IMemoryView` in the table and down the bulleted list.
> * I updated the definition of `IMemoryView` and removed the cross-link to `marshal.ts`.
> * I commented-out the .NET ref source remark just in case we need it later. If we end up not needing it, I'll 🔪 it.
> 
> _**NOTE TO SELF 🦖**_: 🎗️ _Don't forget to WCAG the table before merging!_ 🎗️


> - `IMemoryView`
>   - `IMemoryView` is a JS type of .NET interop, not a primitive JS type.
>   - Bytes aren't copied during marshalling.
>   - `IMemoryView` created for a `Span` is only valid for the duration of the interop call.
>   - `IMemoryView` created for an `ArraySegment` survives after the interop call and is useful for sharing a buffer.
>   - `IMemoryView` doesn't have an analogous JS type, so marshalling a JS object to a .NET `IMemoryView` isn't possible.
> - It's not possible to export a .NET method that returns a `Span`. The `Span` is allocated on the call stack and has GC implications. When calling from JS to .NET, there's no C# stack after the call.
> - For an exported method that returns an `ArraySegment`, calling `dispose()` in `try-finally` block disposes the proxy and unpins the underlying C# byte array. We recommend calling `dispose()` on the object in developer JS code. If developer code doesn't dispose of the object, the JS GC eventually disposes the object. You can also marshal a byte array (`byte[]`) instead of an `ArraySegment`, which copies the bytes.

@pavelsavara @guardrex `MemoryView` implements `IMemoryView` and `IDisposable`, and that's why `MemoryView` has `dispose()` and `isDisposed`. By specifying `IMemoryView` only, I think developers may lose sight of `IDisposable`.

I also think that some information in the quoted text above may be further clarified. May I propose changing to the following?

- `Array`
  - Marshalling an array creates a copy of the array in JavaScript or .NET, as the case may be.
- `MemoryView`
  - `MemoryView` is a JavaScript class for the .NET WebAssembly runtime to marshal `Span` and `ArraySegment`.
  - Unlike marshalling an array, marshalling a `Span` or `ArraySegment` does not create a copy of the underlying memory.
  - An instance of `MemoryView` can only be constructed by the .NET WebAssembly runtime. Therefore, it is not possble to import a JavaScript function as a .NET method that has a parameter of `Span` or `ArraySegment`.
  - `MemoryView` created for a `Span` is only valid for the duration of the interop call. As `Span` is allocated on the call stack, which does not persist after the interop call, it is not possible to export a .NET method that returns a `Span`.
  - `MemoryView` created for an `ArraySegment` survives after the interop call and is useful for sharing a buffer. Calling `dispose()` on a `MemoryView` created for an `ArraySegment` disposes the proxy and unpins the underlying .NET array. We recommend calling `dispose()` in `try-finally` block for all such `MemoryView`.

I have add a point to explictly state that marshalling an array creates a copy of the array, and therefore the code below will not work as intended:

```cs
[JSExport]
// This method will not actually modify the array in JS, as a copy of it is passed to .NET during marshalling
static void ModifyArrayFromJS(string[] message) {
	message[0] = "Override";
	Console.WriteLine(String.Join(',', message));
}
```

I have deleted `If developer code doesn't dispose of the object, the JS GC eventually disposes the object.` because it is not clear when that will happen eventually.

aspnetcore/client-side/import-export-interop.md
```diff
@@ -0,0 +1,357 @@
...
+* `IMemoryView`
+  * `IMemoryView` is a JS type of .NET interop, not a primitive JS type.
+  * Bytes aren't copied during marshalling.
+  * `IMemoryView` created for a `Span` is only valid for the duration of the interop call.
+  * `IMemoryView` created for an `ArraySegment` survives after the interop call and is useful for sharing a buffer.
+  * `IMemoryView` doesn't have an analogous JS type, so marshalling a JS object to a .NET `IMemoryView` isn't possible.
+* It's not possible to export a .NET method that returns a `Span`. The `Span` is allocated on the call stack and has GC implications. When calling from JS to .NET, there's no C# stack after the call.
+* For an exported method that returns an `ArraySegment`, calling `dispose()` in `try-finally` block disposes the proxy and unpins the underlying C# byte array. We recommend calling `dispose()` on the object in developer JS code. If developer code doesn't dispose of the object, the JS GC eventually disposes the object. You can also marshal a byte array (`byte[]`) instead of an `ArraySegment`, which copies the bytes.
```
FranklinWhale, line 226:
```suggestion
* `Array`
  * Marshalling an array creates a copy of the array in JavaScript or .NET, as the case may be.
* `MemoryView`
  * `MemoryView` is a JavaScript class for the .NET WebAssembly runtime to marshal `Span` and `ArraySegment`.
  * Unlike marshalling an array, marshalling a `Span` or `ArraySegment` does not create a copy of the underlying memory.
  * An instance of `MemoryView` can only be constructed by the .NET WebAssembly runtime. Therefore, it is not possble to import a JavaScript function as a .NET method that has a parameter of `Span` or `ArraySegment`.
  * `MemoryView` created for a `Span` is only valid for the duration of the interop call. As `Span` is allocated on the call stack, which does not persist after the interop call, it is not possible to export a .NET method that returns a `Span`.
  * `MemoryView` created for an `ArraySegment` survives after the interop call and is useful for sharing a buffer. Calling `dispose()` on a `MemoryView` created for an `ArraySegment` disposes the proxy and unpins the underlying .NET array. We recommend calling `dispose()` in `try-finally` block for all such `MemoryView`.
```

guardrex: After any Pavel/Dan updates to that, I'll just hit a couple of grammar/style nits before committing it.

danroth27: We still seem to be missing a description of the shape of a MemoryView in JS, like what members it has and how they behave. @pavelsavara How do you want to address that?

maraf: The type is exported in this PR https://github.com/dotnet/runtime/pull/77371. Backport to 7 is in the queue.

guardrex: The existing link to the file will cover it then. I'll take this commit now.

guardrex:
@danroth27 ... WRT naming/titling: In this part of the ToC (client-side development), we don't need to distinguish this interop from an existing JS interop (i.e., `IJSRuntime` JS interop), so it isn't necessary to rely so much on the "import-export" phrasing for naming/titling. Also, ".NET interop" seems like a possible replacement for "JS interop" in this context (but only in titling/naming) because the primary usefulness is for JS apps/JS devs to call .NET ... based on how you titled it (i.e., 'run .NET from JS'). It's possible that the naming/titling should reflect that focus.

I made a change on the last commit to see how things might compose. What do you think about ...

* **Title:** Run .NET from JavaScript
* **Filename:** client-side/dotnet-interop.md
* **URL:** `https://learn.microsoft.com/aspnet/core/client-side/dotnet-interop`
* **UID:** client-side/dotnet-interop
* **ToC entry name:** Run .NET from JavaScript

danroth27:
> I made a change on the last commit to see how things might compose

Looks good to me.

> I still propose that we maintain the formal technology name in the articles' guidance as JavaScript (JS) `[JSImport]`/`[JSExport]` interop in order to have one name for it across client-side and Blazor articles

That's fine. It's a bit wordy, but I think it's workable for now. I'll let you know if I come up with something cleaner.

> It's easy to update the content later to just "JS interop" if this API replaces `IJSRuntime` interop

We still expect to steer Blazor users to`IJSRuntime` so that their JS interop logic works across all Blazor hosting models.

danroth27 (approved):
:shipit:

guardrex:
# 🎉 

One quick pass on Wednesday morning, and I'll merge this and take it live.

**NOTE TO SELF:** 🎗️ Don't forget to WCAG the table before merging! 🎗️

PR #27372, opened 2022-10-25
Add route groups to What's new

tdykstra:
Contributes to #26819


halter73 (approved):
I'm working on improving this sample and expanding on it in a separate document.

Can we get rid of `"~/fundamentals/minimal-apis/7.0-samples/todo-group` since it's so similar to `~/fundamentals/minimal-apis/7.0-samples/todo-group2`.

PR #27385, opened 2022-10-26
[7.0] Finalize JSImport/JSExport interop

guardrex:
Addresses https://github.com/dotnet/AspNetCore.Docs/issues/27016
Addresses https://github.com/dotnet/AspNetCore.Docs/issues/26364

Pavel ... I assume that we can place a copy of the .NET-JavaScript type mappings table in the Blazor JS interop article. I use a versioned INCLUDE file for it to avoid duplicated content. The file is also versioned, so it can be easily updated release-to-release.

btw - I noticed a *NIT* update 😈 to make for another INCLUDE file (`size-limits.md`). I forgot to place a 7.0 version of that file for some earlier updates. I'm just going to tack that on here to knock it out 🥊😵.


guardrex:
Note that because this touches the *What's New* and *Migration* articles that I can only leave it up for a few hours. We're trying not to step on each others' PRs that touch these topics. If you don't get a chance to review before I merge, you can post feedback here for a patch update later.

PR #27424, opened 2022-10-28
Endpoint names are not inferred

halter73:
This was the behavior in some .NET 6 preview releases, but was reverted in .NET 6 RC 2 due to potential uniqueness issues given overloaded method groups: dotnet/aspnetcore#36518. This is why `{linker.GetPathByName("Hi", values: null)}` was null.

It's surprising that this REVIEW note stayed in the doc for over a year without anyone who knows about this noticing, but that's partially on me.

Also, these are technically "endpoint" names, not "route" names.

### DIFF
--- a/aspnetcore/fundamentals/minimal-apis.md
+++ b/aspnetcore/fundamentals/minimal-apis.md
@@ -329,7 +329,7 @@ Route handlers are methods that execute when the route matches. Route handlers c
 
 [!code-csharp[](minimal-apis/samples/WebMinAPIs/Program.cs?name=snippet_sm)]
 
-### Name endpoints and link generation
+### Named endpoints and link generation
 
 Endpoints can be given names in order to generate URLs to the endpoint. Using a named endpoint avoids having to hard code paths in an app:
 
--- a/aspnetcore/fundamentals/minimal-apis.md
+++ b/aspnetcore/fundamentals/minimal-apis.md
@@ -337,12 +337,6 @@ Routes can be given names in order to generate URLs to the route. Using a named 
 
 The preceding code displays `The link to the hello route is /hello` from the `/` endpoint.
 
-Route names are inferred from method names if specified:
-
-[!code-csharp[](minimal-apis/samples/WebMinAPIs/Program.cs?name=snippet_nr2)]
-
-**REVIEW**: `{linker.GetPathByName("Hi", values: null)}` is null in the preceding code.
-
 **NOTE**: Route names are case sensitive.
 
 Route names:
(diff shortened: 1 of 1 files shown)


aspnetcore/fundamentals/minimal-apis.md
```diff
@@ -329,23 +329,17 @@ Route handlers are methods that execute when the route matches. Route handlers c
 
 [!code-csharp[](minimal-apis/samples/WebMinAPIs/Program.cs?name=snippet_sm)]
 
-### Name routes and link generation
+### Name endpoints and link generation
```
captainsafia, line 332:
```suggestion
### Named endpoints and link generation
```

❓

aspnetcore/fundamentals/minimal-apis.md
```diff
@@ -329,23 +329,17 @@ Route handlers are methods that execute when the route matches. Route handlers c
...
 
-Routes can be given names in order to generate URLs to the route. Using a named route avoids having to hard code paths in an app:
+Endpoints can be given names in order to generate URLs to the endpoint. Using a named endpoint avoids having to hard code paths in an app:
```
captainsafia, line 334: > Endpoints can be given names in order to generate URLs to the endpoint.

This sentence feels a little confusing. What exactly are we trying to say here?

halter73: That endpoint names provide some indirection when generating links? I didn't write this originally, so I'm guessing. Do you have a suggestion for how we should reword this?

halter73:
Replaced by #27458.

aspnetcore/fundamentals/minimal-apis.md
pstricks-fans: Providing the readers with a lot of examples makes it much clearer.  :-)

```csharp
var builder = WebApplication.CreateBuilder(args);

var app = builder.Build();

var users = new List<User>
 {
    new User(1,"A"),
    new User(2,"B"),
    new User(3,"C"),
 };

IResult GetUser(int id)
{
    return users.Find(u => u.id == id) is User user ?
        TypedResults.Ok(user) : TypedResults.NotFound();
}

IResult CreateUser(User user, LinkGenerator lg)
{
    users.Add(user);
    var path = lg.GetPathByName(nameof(GetUser), new { id = user.id })!;
    return TypedResults.Created(path, user);
}

app
    .MapGet("/users/{id}", GetUser)
    .WithName(nameof(GetUser));

app.MapPost("/users", CreateUser);

app.Run();

record class User(int id, string name);
```

PR #27439, opened 2022-10-31
JS interop disposal after circuit disconnection

### LINKED ISSUE #27094: How to properly destroy javascript component and avoid JSDisconnectedException
My blazor component wraps existing javascript component that needs some javascript cleaning when removed from DOM.

I call `jsObjectRef.InvokeVoidAsync("destroy")` when disposing blazor component, but it throws unhandled exception:

 >     Microsoft.JSInterop.JSDisconnectedException: JavaScript interop calls cannot be issued at this time. This is because the circuit has disconnected and is being disposed.

The logs are then bloated with this exception

```csharp
    public class RichTextEditor : InputTextArea, IAsyncDisposable
    {
        ElementReference? _elementRef;
        IJSObjectReference? _jsObjectRef;

        [Inject] IJSRuntime Js { get; set; } = default!;

        protected override async Task OnAfterRenderAsync(bool firstRender)
        {
            if (firstRender)
            {
                _jsObjectRef = await Js.InvokeAsync<IJSObjectReference>("CKEditorInterop.init", _elementRef;
            }
            await base.OnAfterRenderAsync(firstRender);
        }

        public async ValueTask DisposeAsync()
        {
            if (_jsObjectRef != null)
            {
                await _jsObjectRef.InvokeVoidAsync("destroy"); //this throws JSDisconnectedException when browser is reloaded  or closed or disconnected
                await _jsObjectRef.DisposeAsync();
            }
        }
    }
````

Is it possible to avoid this exception?

---
#### Document Details

* Content: [Call JavaScript functions from .NET methods in ASP.NET Core Blazor](https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-javascript-from-dotnet?view=aspnetcore-6.0)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

### PR DESCRIPTION
**Pull request by guardrex**
Fixes #27094

Thanks @Liero! 🚀 

* Covers JS disconnection exception for JS interop calls via an INCLUDE file for the the *Call .NET* and *Call JS* articles.
* Improved cross-linking.
* Avoids getting into the weeds on framework implementations.
* Calls out `MutationObserver` for client-side disposal in lost circuit scenarios. Thus far on the PR, the dev will need to implement that on their own, as it's an advanced scenario. There's one example in the framework that we could cross-link as an example at https://github.com/dotnet/aspnetcore/blob/main/src/Components/Web.JS/src/Virtualize.ts, or the PU can provide an example to show.
* ***IMPORTANT***: I'm not going to update **every** component disposal example with a trap for `JSDisconnectedException`. For one thing, it's probably not a Blazor WebAssembly concern. Second, the exception is simple enough to track down via search in the docs after this merges.

### CODE REVIEW

File: aspnetcore/blazor/includes/js-interop/circuit-disconnection.md [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+        }
+    }
+    catch (JSDisconnectedException)
+    {
+    }
+}
+```
+
+> [!NOTE]
+> The pattern for <xref:Microsoft.JSInterop.DotNetObjectReference> disposal is the same in scenarios where .NET is called from JS.
```
SteveSandersonMS, line 37: I'm not sure what this means. Could we have an example for disposal "where .NET is called from JS"?

aspnetcore/blazor/includes/js-interop/circuit-disconnection.md
```diff
@@ -0,0 +1,46 @@
...
+>
+> When a circuit fails and developer disposal code fails with a caught or uncaught <xref:Microsoft.JSInterop.JSDisconnectedException>, automatic garbage collection eventually disposes of any objects created by the developer in the component.
```
SteveSandersonMS, line 39: I'm not sure why we need to point this out here. It's the same as how GC behaves anywhere.

aspnetcore/blazor/includes/js-interop/circuit-disconnection.md
```diff
@@ -0,0 +1,46 @@
...
+
+If you must clean up your own JS objects or execute other JS code on the client after a circuit is lost or you prefer to dispose of one or more <xref:Microsoft.JSInterop.IJSObjectReference>s or <xref:Microsoft.JSInterop.DotNetObjectReference>s on the client, use the [`MutationObserver`](https://developer.mozilla.org/docs/Web/API/MutationObserver) pattern in JS on the client and avoid calling `Dispose`/`DisposeAsync` on the objects in the component.
```
SteveSandersonMS, line 41: > and avoid calling `Dispose`/`DisposeAsync` on the objects in the component.

I think that clause could be dropped, as it should be safe to request disposal from the server whether or not the client is also disposing the same object.

aspnetcore/blazor/includes/js-interop/circuit-disconnection.md
guardrex: It's the same disposal concept (trap `JSDisconnectedException`), but for a `DotNetObjectReference` in a `Dispose` method. Actually, I'm not sure if calling `Dispose` on an `DotNetObjectReference` will trigger a `JSDisconnectedException` if the circuit is dead. I'll try to run a test on that.

aspnetcore/blazor/includes/js-interop/circuit-disconnection.md
guardrex: Trapped or not, I felt it would be a good idea to tell readers that just because there's a `JSDisconnectedException` when they attempt to dispose (i.e., the dispose fails) that they don't need to sweat that object lingering on. This puts the mind at ease on it.

aspnetcore/blazor/includes/js-interop/circuit-disconnection.md
guardrex: Safe, yes ... but that misses the answer to @Liero's ask. The ask is like, 'How do I avoid the logging/`JSDisconnectedException`?' Javier said to either ...

* Catch it and do nothing ... or ...
* Use `MutationObserver` as an alternative to disposing in C# to avoid the exception completely.

aspnetcore/blazor/includes/js-interop/circuit-disconnection.md
Liero: > The JSDisconnectedException isn't thrown in the in the DotNetObjectReference case.
Oh, I think it is thrown unfortunatelly, at least in .NET 6

guardrex: I couldn't repro that here in a 6.0 test app. It didn't throw. You'd need to give me exact repro steps. Also, I recommend not commenting on this aspect further here if you are able to present a step-by-step repro of it because we don't work off of closed issues/PRs. It's best to open a new docs issue from the bottom of the topic.

aspnetcore/blazor/includes/js-interop/circuit-disconnection.md
Liero: > I'm not sure why we need to point this out here. It's the same as how GC behaves anywhere.

Unless there is a collection of all JSObjectReferences tracked somewhere in blazor internals, which would prevent GC. 

I believe there is a such collection must exists somewhere on the javascript side. But the way, Blazor keeps references to all DotNetObjectReference in .NET memory and they are not automatically disposed by GC, are they?

guardrex: @Liero ... We don't work closed issues/PRs. You can open a new docs issue from the bottom of the topic cross-linking to this PR's updates.

Note that we're on the edge of the .NET 7 release, and the PU can't be pinged on doc issues until a few weeks after the release. Even then, it's not clear if they'll be able to respond with devs taking off for holiday breaks. Such an issue might not be reached until 23Q1.

I'll need to lock this PR down as resolved. Our work process isn't efficient for closed issues/PRs.

PR #27461, opened 2022-11-01
Document IComponent

### LINKED ISSUE #27337: Document IComponent
I couldn't find any documentation about IComponent in the Blazor docs. A user pointed out this missing content in the Blazor eBook: https://github.com/dotnet/docs/issues/31426.

IComponent is the lowest level abstraction for components in Blazor. Typically, you implement Razor components using .razor files, which are based on ComponentBase, but you can build use components based on IComponent instead. It's an advanced scenario, but it should be documented somewhere.

@SteveSandersonMS Thoughts on how you'd like IComponent documented? Could you provide a sketch of the desired content?

---
#### Document Details

* Content: [ASP.NET Core Razor components](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-6.0)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #27337

I'm drafting some kind of content here for this ... *crappy content*??! 🙈😄. This is just to get the ball rolling with some ideas.

### DIFF
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -26,7 +26,7 @@ By default, <xref:Microsoft.AspNetCore.Components.ComponentBase> is the base cla
 
 [!INCLUDE[](~/includes/aspnetcore-repo-ref-source-links.md)]
 
-Developers typically create Razor components from Razor component files (`.razor`) or base their components on <xref:Microsoft.AspNetCore.Components.ComponentBase>, but components can also be built by implementing <xref:Microsoft.AspNetCore.Components.IComponent>. Developer-built components that implement <xref:Microsoft.AspNetCore.Components.IComponent> can take low-level control over rendering at the cost of having to manually trigger rendering with events and other lifecycle methods that the developer must create and maintain.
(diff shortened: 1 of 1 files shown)


SteveSandersonMS:
Keeping it brief like proposed in this PR sounds good to me.

One main reason why the `IComponent` interface exists in case people want something different from `ComponentBase`'s built-in lifecycle, but unless a developer knows they have a reason to want that, we don't need to imply they should come up with a reason. We could say that components that implement `IComponent` have more low-level control over when (or if) they render themselves, but at the cost that the developer must manually trigger rendering instead of having it happen automatically with events and other lifecycle methods.

The other main reason why `IComponent` exists, which we probably wouldn't document at all, is that it helps Blazor and the Razor compiler retain architectural clarity and not falsely build up interconnections between rendering internals and arbitrary implementation details of `ComponentBase`.

guardrex:
I've added the following remark right after it mentions devs building on `IComponent` ...

>  Developer-built components that implement \<xref:Microsoft.AspNetCore.Components.IComponent> can take low-level control over rendering at the cost of having to manually trigger rendering with events and other lifecycle methods that the developer must create and maintain.

PR #27532, opened 2022-11-08
Message size limit and prerendering state size

### LINKED ISSUE #27531: Mention limitations of pre-render size
I was debugging an issue with a page that rendered, but started showing the "reconnecting" dialog after ~20 seconds or so. There was no error server-side and the client-side contained rather unhelpful traceback:

> [Error] [2022-11-08T09:44:35.259Z] Error: Error: Circuit host not initialized.
	log (blazor.server.js:1:115822)
	(anonymous function) (blazor.server.js:1:122515)
	asyncFunctionResume
	(anonymous function)
	promiseReactionJobWithoutPromise
	promiseReactionJob

After some investigation I've found a similar issue with this message: https://github.com/dotnet/aspnetcore/issues/44280#issuecomment-1266747939. This is a severe problem in our situation, as we don't know beforehand the pre-rendered size of certain data. The failure mode here is completely wrong: when you go over the limit, the client will fail to initialize and you have no way of detecting this nor recovering. State preservation should fail server-side when trying to persist data that will be over the limit.

In any case: this limitation should be included in the documentation as it will cause serious problems for users.

---
#### Document Details

* ID: 329a19ef-fb46-1f6b-a734-fa1c1334b1e5
* Version Independent ID: 3fa4318a-3c11-c1d8-bf24-ddc6d8441ca2
* Content: [Prerender and integrate ASP.NET Core Razor components](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/prerendering-and-integration?view=aspnetcore-6.0#persist-prerendered-state)
* Content Source: [aspnetcore/blazor/components/prerendering-and-integration.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/prerendering-and-integration.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #27531

Thanks @Bouke! :rocket: ... I'll remain 👂 on your PU issue in case there's more to address. If so, I'll do that on a follow-up PR. This is ***NOT*** a good time to leave PRs lying around with the .NET 7 release and all of the docs work needed to activate the new release docs ... 😈 ***merge conflicts!*** 😈

I added the coverage both to the *Prendering* doc ***and*** in the *Persist component state* TH doc. That should surface it well.

I also included the exact client-side message about the circuit not initializing, which can help devs find this content via a string-based docs search.

Since most apps won't hit the limit  ... we hope 🤞🍀 and is consistent with the lack of issues generated on this particular error, here on the docs repo anyway, I'll hold off on placing this info into the *Fundamentals* > *SignalR* topic. If more feedback arrives on it 💥 in apps, I'll further expand this into that fundamentals topic.

### DIFF
--- a/aspnetcore/blazor/components/prerendering-and-integration.md
+++ b/aspnetcore/blazor/components/prerendering-and-integration.md
@@ -936,7 +936,7 @@ A large prerendered state size may exceed the SignalR circuit message size limit
 
 To resolve the problem, use ***either*** of the following approaches:
 
-* Reduce the amount of data that you are putting in prerrendered state.
+* Reduce the amount of data that you are putting into the prerrendered state.
 * Increase the [SignalR message size limit](xref:blazor/fundamentals/signalr#circuit-handler-options-for-blazor-server-apps). ***WARNING***: Increasing the limit may increase the risk of Denial of service (DoS) attacks.
 
 ## Additional Blazor Server resources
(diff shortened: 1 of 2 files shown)


Bouke:
Thank you for the fast turn-around on this and the clarification provided, should be helpful.

guardrex:
Sure thing ... I'll keep an 👂 on if Javier has further feedback that can be placed into these remarks for a follow-up PR.

PR #27535, opened 2022-11-08
Blazor .NET 7 release work

guardrex:
Addresses #26364


guardrex:
Rick, I'm going to go ahead and make this live and hope that the default 6.0 load doesn't break anything. Dan has placed the blog post links with bookmarks, so those will break from his blog post unless I get this in right now.

PR #27562, opened 2022-11-09
Update for ASP.NET Core 7.0

gewarren:
Follow up to https://github.com/dotnet/aspnetcore/issues/44973#issuecomment-1309211993.

### DIFF
--- a/aspnetcore/blazor/security/includes/redirecttologin-component.md
+++ b/aspnetcore/blazor/security/includes/redirecttologin-component.md
@@ -14,3 +14,19 @@ The `RedirectToLogin` component (`Shared/RedirectToLogin.razor`):
     }
 }
 ```
+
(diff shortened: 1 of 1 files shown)


guardrex (approved):
Thanks @gewarren ... I'll do a little follow-up on this separately soon.

* We have history state coverage that I can cross-link.
* I'll place a remark on this in the *Migration* topic.
* I'd like to maintain a versioned INCLUDE for this so that it will only appear in 7.0 content.
* We usually link to the Announcement because we want devs to learn the location for centrailized knowledge on breaking and other framework changes.

gewarren:
Thanks @guardrex. Would it make sense for you to start documenting the breaking change announcements at https://learn.microsoft.com/en-us/dotnet/core/compatibility/7.0#aspnet-core instead of me? In cases like this, you'd have a much better idea of any related content that needs to be updated based on the announcement. cc @mcleblanc

guardrex:
If it applies to Blazor, yes. That's my only responsibility here at this time.

That particular one just got by me 😈 ... It didn't make a breaking change list that I was using prior to Javier placing that Announcement. I better ***LOOK AGAIN*** 👀 in the morning to see if anything else slipped!

PR #27570, opened 2022-11-10
Update content about Azure SignalR

JialinXin:
Client result is supported in Azure SignalR SDK v1.19.0.

### DIFF
--- a/aspnetcore/signalr/hubs.md
+++ b/aspnetcore/signalr/hubs.md
@@ -693,9 +693,6 @@ hubConnection.on("GetMessage", async () => {
 });
 ```
 
-> [!NOTE]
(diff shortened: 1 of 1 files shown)


guardrex (approved):
Thanks @JialinXin! 🚀 

We can drop the NOTE. We don't remark on migration changes in the reference docs very much. Otherwise, these things linger ***FOREVER*** in future versions without getting removed. It keeps the docs lean. I think this is covered by ...

https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-7.0?view=aspnetcore-7.0#client-results

... for migration/release.

BrennanConroy:
Looks like it's already in [What's New](https://learn.microsoft.com/aspnet/core/release-notes/aspnetcore-7.0?view=aspnetcore-7.0#client-results)
And there is no migration needed, it's a new feature so you either choose to use it or don't.

PR #27581, opened 2022-11-10
Update signalr.md

kyhou:
Changed  ..."and select signalr.js and signalr.js." to ..." and select signalr.js and signalr.min.js."


guardrex (approved):
Thanks @kyhou! :rocket:

PR #27607, opened 2022-11-12
Fix nested groups example

benfoster:
Fixes the nested group example. The `inner` group should be created from the `outer` group


Rick-Anderson:
Thanks @benfoster, that's the problem with embedded snippets.

Rick-Anderson:
Fixed in #27615

PR #27628, opened 2022-11-14
Almost File By Version /2

Rick-Anderson:
[Internal review URL](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-7.0&branch=pr-en-us-27628)

Intermediate step to File By Version (FBV)

Versions < 6 were too much work to split out

```
:::moniker range=">= aspnetcore-7.0" 
:::moniker range="= aspnetcore-6.0"
:::moniker range=">= aspnetcore-3.1 < aspnetcore-6.0"   // 3.1 & 5
:::moniker range=">= aspnetcore-5.0 < aspnetcore-6.0"  // 5
:::moniker range=">= aspnetcore-3.0 < aspnetcore-6.0"  // 3.0, 3.1 & 5
```

No more tedious and error prone moniker prep, no more tedious and error prone moving moniker sections at GA. ~What we do loose is a diff file showing what has changed, but on most updates, that's not useful anyway.~

Losing the diff is optional, can use the old moniker version prep approach and then you don't lose that.


aspnetcore/fundamentals/configuration/index.md
```diff
@@ -5,1850 +5,16 @@ description: Learn how to use the Configuration API to configure AppSettings in
...
-## Additional resources
-
-* [Configuration source code](https://github.com/dotnet/runtime/tree/main/src/libraries/Microsoft.Extensions.Configuration)
-* [WebApplicationBuilder source code](https://github.com/dotnet/aspnetcore/blob/v6.0.1/src/DefaultBuilder/src/WebApplicationBuilder.cs)
-* [View or download sample code](https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/fundamentals/configuration/index/samples) ([how to download](xref:index#how-to-download-a-sample))
-* <xref:fundamentals/configuration/options>
-* <xref:blazor/fundamentals/configuration>
-
-:::moniker-end
+[!INCLUDE[](~/includes/configuration/index6.md)]
```
Rick-Anderson, line 17: @tdykstra @wadepickett  for now on, all updates should be IFBV (Intermediate or Include File By Version). That will make it trivial to:

* start the next version
* Promote the released version at GA

cc @guardrex

Rick-Anderson:
@tdykstra  no need to review the changes other than this new approach to using includes.

guardrex:
Is this what their guidance says to do? If so, I'm utterly shocked by it. Version by File was ... I thought ... based on the version ***as a URL segment*** that reflects the version and with the files being in version-named folders (e.g., "3.1", "5.0", "6.0", "7.0", etc.).

Rick-Anderson:
> Is this what their guidance says to do? If so, I'm utterly shocked by it. Version by File was ... I thought ... based on the version _**as a URL segment**_ that reflects the version and with the files being in version-named folders (e.g., "3.1", "5.0", "6.0", "7.0", etc.).

No, this is just to make it easier for FBV move, that's why I call it IFBV, intermediate or include File by version.

guardrex:
Oh, I see ... gotcha 👍

wadepickett:
A great creative solution for getting this a step closer to the goal!  Thanks!

**Resource versioning**
I think it would be good to call out what is happening with resources the .md file is using in addition to the include.  To get a distinct version we don't have to sort out later, I would expect the app samples and screenshot/art resources to be clearly identified with a version somehow.  That would mean in some cases copying them, however when needed to a new version id of some sort and would increase the size of the repo.  On the other hand, later, when archiving, which would significantly reduce the size of the repo we would not have previous resource versions as a dependency for newer ones.  Thoughts?

wadepickett:
A great creative solution for getting this a step closer to the goal!  Thanks!

**Resource versioning**
As a matter of applying this to all topics: I think it would be good to call out what is happening with resources the .md file is using in addition to the include.  To get a distinct version we don't have to sort out later, I would expect the app samples and screenshot/art resources to be clearly identified with a version somehow.  That would mean in some cases copying them, however when needed to a new version id of some sort and would increase the size of the repo.  On the other hand, later, when archiving, which would significantly reduce the size of the repo we would not have previous resource versions as a dependency for newer ones.  Thoughts?

Rick-Anderson:
> To get a distinct version we don't have to sort out later, I would expect the app samples and screenshot/art resources to be clearly identified with a version somehow.

Why, we don't do that now. No reason to start.

> That would mean in some cases copying them, however when needed to a new version id of some sort.

The worst thing we could do would be making duplicate images.

 >On the more immediate downside that would increase the size of the repo. Longer term however, archiving old versions would significantly reduce the size of the repo and we would not have previous resource versions as a dependency for newer ones to sort out. Thoughts?

> archiving old versions would significantly reduce the size of the repo

That doesn't reduce the size, it increases it. Deleting images doesn't reduce the size as the image is always in the history.

wadepickett:
Image history:  I am referring to the download size of the repo.  That would be smaller.  Yes, good point it would still be in history.

Why do .md resource versioning now? (we have not in the past):  So archiving and sorting out versions in the future is not messy.  We would start that now going forward to updated topics as we go to start reducing that spaghetti of dependencies.

wadepickett:
Image history:  I am referring to the download size of the repo.  That would be smaller.  Yes, good point it would still be in history.

Why do .md resource versioning now (we have not in the past):  So archiving and sorting out versions in the future is not messy.  We would start that now going forward to updated topics as we go to start reducing that spaghetti of dependencies.

I guess if we didn't version dependent resources, we could automate by tracing each file, find its dependency and copy it to an archive repo at some point to get a clean version.

wadepickett (approved):

wadepickett:
I would think this would all work for localizing, etc, but it might be good to run by the localization team just in case?  Or run it through that system once with a few examples before we change everything as we go?  Just a thought.

Rick-Anderson:
> Why do .md resource versioning now (we have not in the past): So archiving and sorting out versions in the future is not messy. We would start that now going forward to updated topics as we go to start reducing that spaghetti of dependencies.
> 
> I guess if we didn't version dependent resources, we could automate by tracing each file, find its dependency and copy it to an archive repo at some point to get a clean version. So maybe that would be the answer for that case.

That has nothing to do with FBV or IFBV.

>So archiving and sorting out versions in the future is not messy.

red herring, we aren't archiving and there is no version issue. That's not something we would prioritize.

Rick-Anderson:
> , but it might be good to mention to localization team just in case?

Yes, I'm working with them.

PR #27658, opened 2022-11-16
Change misleading claims description

jez9999:
This sentence is highly misleading.  Claims can contain "permission" information, there is no issue with this, and this article even goes on to talk about "role claims" which *are* permission information.

### DIFF
--- a/aspnetcore/security/authentication/claims.md
+++ b/aspnetcore/security/authentication/claims.md
@@ -14,7 +14,7 @@ By [Damien Bowden](https://github.com/damienbod)
 
 :::moniker range=">= aspnetcore-6.0"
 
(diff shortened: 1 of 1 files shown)


guardrex:
It doesn't say what they can do tho ... it only provides information to the system that will authorize them for what they can do. If a role claim says that the user is in the "admin" role, it really doesn't say anything about what an "admin" can do after being authenticated, and it certainly can be different app-to-app for exactly the same claim.

jez9999:
OK, but that's still a misleading sentence.  It makes it sound like the information in a claim isn't related to determining what a user can do.  Saying a user is in an "admin" role, or even that they have "edit X permission" is clearly related to determining what they can do, and a system may even not need to go to a database to determine what they can do based on this information.

PR #27668, opened 2022-11-17
Blazor loc shared resource approach

### LINKED ISSUE #27594: Add guidance on shared resource approach in Blazor loc article
[Enter feedback here]
For what concerns Blazor webassembly localization, an user is likely to want to use his/her own folders hierarchy, and may not want to have the .resx file name equal to one of the .razor class using it. Thus, I think the doc is missing the following points:

- In IStringLocalizer<CultureExample2>, CultureExample2 is the name of the .resx class providing translations, not of the .razor class using it. It took me about two hours to understand this.
- In order to let dotnet find the CultureExample2 .resx file, you need to make a .cs class, even a fake one, with the same qualified name. This may seems obvious, but for someone who has never used .resx files so far like me, it is not. It took me about one hour trying to figure this out.

---
#### Document Details

* Content: [ASP.NET Core Blazor globalization and localization](https://learn.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-7.0&pivots=webassembly)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #27594

Thanks @AlessandroMartinelli! :rocket:

@hishamco, I think we should have a tidbit on this in the Blazor loc doc. If we merely cross-link to the main doc set for this, it's just a bit too cryptic for a typical dev and certainly too cryptic for ***new-to-Blazor/new-to-loc/new-to-.NET devs***. This section makes it 💀 ***dead simple*** 💀 to set up with the existing examples in the topic.

@AlessandroMartinelli ... WRT ...

> * Localization keys (the terms to be translated) taken from a constants file. This would help guarantee uniqueness of keys but would introduce a little overhead since all terms have to be inserted in the constants file first.
> * Localization keys sparsed among the .razor files. This would allow to avoid the constants file overhead, but it would be harder to keep uniqueness of keys (important to avoid duplicated works during keys translation by the hand of the human translator)

That's beyond the scope for the Blazor node, as we just want to call out the general implementation approach here. What you're asking about applies to the whole subject of shared resources. When you say, 'sparsed among the .razor files,' that could just as easily be WRT pages of RP apps or views of MVC apps. It's a general concern. If you feel the current coverage in the main doc set is lacking guidance on it, please open a separate issue from the bottom of the [main doc set's localization article](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/localization) to be considered separately from this.

⚠️ ***NOTE TO SELF*** ⚠️: ~This is currently placed for 7.0 just for the feedback round on the PR. After any/all feedback is addressed, duplicate this section back to the 6.0, 5.0, and 3.1 zones of the article ***before merging this***.~ ***Done!*** :+1:

### DIFF
--- a/aspnetcore/blazor/globalization-localization.md
+++ b/aspnetcore/blazor/globalization-localization.md
@@ -880,15 +880,15 @@ To create localization shared resources, adopt the following approach. For addit
 * Create a dummy class with an arbitrary class name. In the following example:
 
   * The app uses the `BlazorSample` namespace, and localization assets use the `BlazorSample.Localization` namespace.
-  * The dummy class is named `ComponentLocalization`.
+  * The dummy class is named `SharedResource`.
   * The class file is placed in a `Localization` folder at the root of the app.
 
-  `Localization/ComponentLocalization.cs`:
+  `Localization/SharedResource.cs`:
 
   ```csharp
   namespace BlazorSample.Localization
   {
-      public class ComponentLocalization
+      public class SharedResource
       {
       }
   }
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/globalization-localization.md
```diff
@@ -873,6 +873,53 @@ To further understand how the Blazor framework processes localization, see the [
...
+  * The app uses the `BlazorSample` namespace, and localization assets use the `BlazorSample.Localization` namespace.
+  * The dummy class is named `ComponentLocalization`.
+  * The class file is placed in a `Localization` folder at the root of the app.
+
+  `Localization/ComponentLocalization.cs`:
+
+  ```csharp
+  namespace BlazorSample.Localization
+  {
+      public class ComponentLocalization
```
hishamco, line 891: `SharedResource` might be a better name, coz it could be used for components and other things

aspnetcore/blazor/globalization-localization.md
```diff
@@ -873,6 +873,53 @@ To further understand how the Blazor framework processes localization, see the [
...
+      {
+      }
+  }
+  ```
+
+* Create the shared resource files with a **Build Action** of `Embedded resource`. In the following example:
+
+  * The files are placed in the `Localization` folder with the dummy `ComponentLocalization` class (`Localization/ComponentLocalization.cs`).
```
hishamco, line 899: We could add a note that `Localization` is resource path that could be set in the `LocalizationOptions`

aspnetcore/blazor/globalization-localization.md
```diff
@@ -873,6 +873,53 @@ To further understand how the Blazor framework processes localization, see the [
...
+  ```razor
+  @using Localization
+  @inject IStringLocalizer<ComponentLocalization> Loc
+  ```
+
+  ```razor
+  @inject IStringLocalizer<Localization.ComponentLocalization> Loc
+  ```
+
+For additional guidance on using the preceding approach with <xref:Microsoft.AspNetCore.Mvc.Localization.IHtmlLocalizer> and <xref:Microsoft.Extensions.Localization.IStringLocalizerFactory>, see <xref:fundamentals/localization#make-the-apps-content-localizable>.
```
hishamco, line 921: No need for this, because we injecting `IStringLocalizer` not `IHtmlLocalizer`

hishamco (changes requested):

aspnetcore/blazor/globalization-localization.md
```diff
@@ -873,6 +873,56 @@ To further understand how the Blazor framework processes localization, see the [
 
 :::zone-end
 
+## Shared resources
+
+To create localization shared resources, adopt the following approach. For additional information, see the shared resources guidance in the <xref:fundamentals/localization#make-the-apps-content-localizable> article.
```
hishamco, line 878: For additional information, see the shared resources guidance in the <xref:fundamentals/localization#make-the-apps-content-localizable> article.

IMHO this should moved at the end

hishamco (changes requested):
Just a little change left, one more thing why you duplicate the changes?

hishamco:
I remember there's a special moniker or it's deprecated?!!

guardrex:
I don't know what you mean. It's the same ones that we've always used ...

```
::: moniker range=">= aspnetcore-7.0"

::: moniker-end

::: moniker range=">= aspnetcore-6.0 < aspnetcore-7.0"

::: moniker-end

::: moniker range=">= aspnetcore-5.0 < aspnetcore-6.0"

::: moniker-end

::: moniker range="< aspnetcore-5.0"

::: moniker-end
```

hishamco:
That what I meant ;) I didn't notice them in your PR

guardrex:
About ..... errrrr .... say 97% 😄 (*obviously guessing*) ... of the Blazor docs use them today. There are only a couple that don't have them.

guardrex:
Ok, I think we're good to go here.

Thanks 🥇 ***.NET Localization Maniac***™ 🥇 for  your on-going help with these.

I know just enough LOC to get a 👊 in the 😵‍💫 by the community without your help! 😆

Thanks again, @AlessandroMartinelli! 🚀

AlessandroMartinelli:
> Thanks again, @AlessandroMartinelli! 🚀

Thanks to you, glad to have been helpful!

Issue #26044, opened 2022-06-03
Blazor Hybrid API doc cross-links


guardrex:
**UPDATES**

* **11/3/22**: Four appeared in the API docs since this issue was opened.
* **11/10/22**: Added the `UrlLoadingStrategy` enum for .NET 7.
* **12/19/22**: API docs should be online now. This issue is unblocked for work, which I hope to reach this week.
* **12/21/22**: Updating today for the latest API doc adds. Still missing a few.
* **12/23/22**: Added a NOTE on the `InitializeComponent` method where it's called out in the Hybrid docs.

Currently MIA ...

* `ExternalNavigationStarting`
* `ExternalLinkNavigationEventArgs`
* `ExternalLinkNavigationPolicy`

Prior MIA API that became available. The following have already received their API doc cross-links:

* ~`InitializeComponent`~: Generated at build time.
* ✔️ `Microsoft.Maui.Hosting.MauiApp`: `<xref:Microsoft.Maui.Hosting.MauiApp>`
* ✔️ `UseMauiApp`: `<xref:Microsoft.Maui.Controls.Hosting.AppHostBuilderExtensions.UseMauiApp%2A>`
* ✔️ `MauiAppBuilder`: `<xref:Microsoft.Maui.Hosting.MauiAppBuilder>`
* ✔️ `Microsoft.Maui.Storage`: `<xref:Microsoft.Maui.Storage>`
* ✔️ `UrlLoading`
* ✔️ `UrlLoadingEventArgs`
* ✔️ `UrlLoadingStrategy`
* ✔️ `Microsoft.AspNetCore.Components.WebView.WindowsForms`
* ✔️ `Microsoft.AspNetCore.Components.WebView.Wpf`
* ✔️ `AddMauiBlazorWebView`
* ✔️ `BlazorWebView`
* ✔️ `BlazorWebViewInitializing`
* ✔️ `BlazorWebViewInitialized`

davidbritch:
@guardrex .NET MAUI's `InitializeComponent` method is generated by a source generator at app build time and added to the compilation object for the calling class. So there isn't an API doc for it.

guardrex:
@davidbritch ... Are you aware if there's anything going on in the background to get the following API exposed in the API docs?

* ExternalNavigationStarting
* ExternalLinkNavigationEventArgs
* ExternalLinkNavigationPolicy

I'd like to be able to cross-link those to API docs in order to close this issue out.

[...]

PR #27702, opened 2022-11-21
Exception handling and Glob/Loc in Blazor Hybrid

### LINKED ISSUE #25679: Blazor Hybrid: Globalization and localization topic
When content is provided in an issue comment here, please remove the PU member assignment and assign the issue to @guardrex 🦖. PU member will be pinged on the PR when it goes up.

```
aspnetcore/blazor/hybrid/globalization-localization.md
```

guardrex:
Addresses #26364
Fixes #25679

I can't find API coverage on `System.AppDomain.CurrentDomain.UnhandledException`. I've added it to a list of temporarily-MIA API on #26044.

### UPDATE

There's one more thing we can knock out here ... 

**Blazor Desktop tenets: Global readiness**
https://github.com/dotnet/maui/issues/2532

... and I'll add that right now 🏃‍♂️.

### UPDATE 2X

I had to make several phrasing guesses to get Pranav's remarks in here from ...

https://github.com/dotnet/maui/issues/2532#issuecomment-1026378294

It probably needs some work on review.

WRT ...

> There are some instances of users trying this in their [Xamarin.Forms apps](https://stackoverflow.com/a/44417380) which relies on triggering an event causing their app to re-render. A determined user could do something similar to re-render the root component in this case.

... that's cool; but if we're going to include that concept, I think we need some kind of implementation guidance. Do you want to include it? If so, can you supply the minimal text and code that that will make sense to readers?

### DIFF
--- a/aspnetcore/blazor/hybrid/index.md
+++ b/aspnetcore/blazor/hybrid/index.md
@@ -35,6 +35,28 @@ Blazor Hybrid exposes the underlying Web View configuration for different platfo
 
 Use the preferred patterns on each platform to attach event handlers to the events to execute your custom code.
 
+## Exception handling in Windows Forms and WPF apps
+
+*This section only applies to Windows Forms and WPF Blazor Hybrid apps.*
+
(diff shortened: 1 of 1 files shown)


TanayParikh (approved):
Thanks @guardrex!

> Do you want to include it?

I don't think that part's necessary for the official docs. It's more of an advanced scenario.

Had another question as well for which I've reached out offline.

PR #27704, opened 2022-11-21
Rename People component

### LINKED ISSUE #27703: Typo on Page
This is minor, but this code sample results in a compile error.  The error is

```
Error (active)	CS0542	'people': member names cannot be the same as their enclosing type.	
```
```
public List<Person> people =
        new()
       {
            { new Person { Data = "Person 1" } },
            { new Person { Data = "Person 2" } },
            { new Person { Data = "Person 3" } }
        };
```
---

The issue is that the name of the page is "People" and then the list name is also 'people'

Would suggest changing the name of the list to `peopleList`.

#### Document Details

* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #27703

Thanks @gregoryagu! 🎸 ... I'm going to (***super quickly*** 🏃‍♂️🏃‍♂️🏃‍♂️🏃‍♂️) change the component name/path to fix this. I figured that would be a bit faster than changing the `people` object, and I'm in a real rush this week to wrap up as much as I can before the 🦃 weekend 😄. I have a ***heavy schedule*** ⛰️⛏️😅 right thru the EOY. The related changes to the snippet sample app are at https://github.com/dotnet/blazor-samples/pull/59, and I guess it didn't break on snippet build. It ***WAS*** catching those because I fixed a few of these. Oh, well. No worries. It will be fixed ***NOW*** 😆. Thanks for the report.

### DIFF
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -780,10 +780,10 @@ When rendering a list of elements or components and the elements or components s
 
 Consider the following example that demonstrates a collection mapping problem that's solved by using [`@key`][5].
 
-For the following `Details` and `People` components:
+For the following `Details` and `PeopleExample` components:
 
-* The `Details` component receives data (`Data`) from the parent `People` component, which is displayed in an `<input>` element. Any given displayed `<input>` element can receive the focus of the page from the user when they select one of the `<input>` elements.
-* The `People` component creates a list of person objects for display using the `Details` component. Every three seconds, a new person is added to the collection.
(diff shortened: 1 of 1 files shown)


guardrex:
> Note, you actually can have the same name if the page name is Capitalized.

Actually, that seems to have been the problem here. I think YOU changed it on your side, right? The example as shown was valid, which explains why it built. We actually don't worry about it too much when the examples are changed by the dev that breaks the code. In this case tho, this will make sure that it doesn't happen again in this case.

PR #27708, opened 2022-11-21
Align scheme name with code snippet

jvandertil:
Aligns the name of the scheme in the text with the scheme name in the code


guardrex (approved):
Thanks @jvandertil! 🚀 

Confirmed for 6.0 or later ... 

https://github.com/dotnet/AspNetCore.Docs.Samples/blob/main/test/integration-tests/IntegrationTestsSample/tests/RazorPagesProject.Tests/IntegrationTests/AuthTests.cs#L109

Confirmed the scheme is "Test" prior to 6.0, so that part of the doc can remain the same ...

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/test/integration-tests/samples/3.x/IntegrationTestsSample/tests/RazorPagesProject.Tests/IntegrationTests/AuthTests.cs#L115

PR #27726, opened 2022-11-22
Example on how to automatically refresh the browser when Blazor Server reconnection fails

MackinnonBuck:
## Example on how to automatically refresh the browser when Blazor Server reconnection fails

This PR adds an example showing how to configure a Blazor Server app to automatically refresh the browser if reconnection to the server fails.

### Description

Customers have been asking for a way to keep the existing Blazor Server reconnection logic and UI, but modified slightly so that refreshing the browser happens automatically rather than requiring direct user action. We don't have an extensibility point in our current JavaScript API that allows this do be done without reimplementing the full reconnection logic and UI. That said, it's not _that_ much code, so it could be sufficient to provide an example in the docs showing how to achieve this.

I'm opening this initially as a draft PR because there are a few open questions:
1. How complete should the example code be?
    * Should we make the example more complete by including some example CSS for the `reconnect-modal`?
    * Or, should we take the opposite approach and reduce the example by simplifying the reconnect functionality?
      * I'm personally not in favor of this, because the customer ask is to retain existing reconnect functionality but with automatic browser refreshing.
2. Is pasting the code inline the best approach?
    * While the code is straightforward, it's a bit on the longer end. Should we put this code elsewhere and link to it from the docs?
3. Rather than having a docs example for this at all, should we address https://github.com/dotnet/aspnetcore/issues/32113 and add new APIs to enable this scenario without requiring reconnection logic to be reimplemented?

Fixes https://github.com/dotnet/aspnetcore/issues/44715

### DIFF
--- a/aspnetcore/blazor/fundamentals/signalr.md
+++ b/aspnetcore/blazor/fundamentals/signalr.md
@@ -326,6 +326,86 @@ To modify the connection events, register callbacks for the following connection
 </body>
 ```
 
+### Advanced scenario: Automatically refresh the page when reconnection fails
+
+The default reconnection behavior requires the user to take manual action to refresh the page after reconnection fails. However, a custom reconnection handler can be used to automatically refresh the page:
+
+`Pages/_Host.cshtml`:
+
+```cshtml
+<body>
+    ...
+
+    <div id="reconnect-modal" style="display: none;"></div>
+    <script src="_framework/blazor.server.js" autostart="false"></script>
+    <script src="boot.js"></script>
+</body>
+```
+
+`wwwroot/boot.js`:
+
+```js
+(() => {
+  const maximumRetryCount = 3;
+  const retryIntervalMilliseconds = 5000;
+  const reconnectModal = document.getElementById('reconnect-modal');
+  
+  const startReconnectionProcess = () => {
+    reconnectModal.style.display = 'block';
+
+    let isCanceled = false;
+
+    (async () => {
+      for (let i = 0; i < maximumRetryCount; i++) {
+        reconnectModal.innerText = `Attempting to reconnect: ${i + 1} of ${maximumRetryCount}`;
+
+        await new Promise(resolve => setTimeout(resolve, retryIntervalMilliseconds));
+
+        if (isCanceled) {
+          return;
+        }
+
+        try {
+          const result = await Blazor.reconnect();
+          if (!result) {
+            // The server was reached, but the connection was rejected; reload the page.
+            location.reload();
+            return;
+          }
+
+          // Successfully reconnected to the server.
+          return;
+        } catch {
+          // Didn't reach the server; try again.
+        }
+      }
+
+      // Retried too many times; reload the page.
+      location.reload();
+    })();
+
+    return {
+      cancel: () => {
+        isCanceled = true;
+        reconnectModal.style.display = 'none';
+      },
+    };
+  };
+
+  let currentReconnectionProcess = null;
+
+  Blazor.start({
+    reconnectionHandler: {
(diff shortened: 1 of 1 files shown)


danroth27 (approved):
This seems fine to me, and I don't think the code is too long to include inline.

danroth27:
@MackinnonBuck It looks like https://github.com/dotnet/aspnetcore/issues/32113 may be a duplicate of https://github.com/dotnet/aspnetcore/issues/44669, which is already being considered for .NET 8. Thoughts?

aspnetcore/blazor/fundamentals/signalr.md
```diff
@@ -326,6 +326,86 @@ To modify the connection events, register callbacks for the following connection
 </body>
 ```
 
+### Advanced scenario: Automatically refresh the page when reconnection fails
```
guardrex, line 329: We almost never tag headings this way. Can it just call out what this content covers? ...

```suggestion
### Automatically refresh the page when reconnection fails
```

aspnetcore/blazor/fundamentals/signalr.md
```diff
@@ -326,6 +326,86 @@ To modify the connection events, register callbacks for the following connection
...
+
+    <div id="reconnect-modal" style="display: none;"></div>
+    <script src="_framework/blazor.server.js" autostart="false"></script>
+    <script src="boot.js"></script>
+</body>
+```
+
+`wwwroot/boot.js`:
+
+```js
```
guardrex, line 347:
```suggestion
```javascript
```

guardrex (approved):
LGTM with three suggestions.

aspnetcore/blazor/fundamentals/signalr.md
```diff
@@ -326,6 +326,86 @@ To modify the connection events, register callbacks for the following connection
 </body>
 ```
 
+### Advanced scenario: Automatically refresh the page when reconnection fails
+
```
guardrex, line 330:
```suggestion

*This section applies to Blazor Server apps.*
```

guardrex:
> Should we make the example more complete by including some example CSS for the reconnect-modal?

Your call; but if you do include it and want to cross-link, we cover that at ...

\[Reflect the connection state in the UI (Blazor Server)](#reflect-the-connection-state-in-the-ui-blazor-server)

MackinnonBuck:
> @MackinnonBuck It looks like [dotnet/aspnetcore#32113](https://github.com/dotnet/aspnetcore/issues/32113) may be a duplicate of [dotnet/aspnetcore#44669](https://github.com/dotnet/aspnetcore/issues/44669), which is already being considered for .NET 8. Thoughts?

@danroth27 I think those seem like two different issues.

From my understanding, the former is asking for an extra `onReconnectionFailed` event on the default reconnection handler so you can add the automatic refreshing behavior in just a few lines of code.

The latter seems to be asking for a way for `Blazor.reconnect()` to signal whether reconnection failed due to the circuit ID being invalid or due to something like the server being down. Addressing this would still require that customers provide a full implementation of the reconnection logic in order to automatically refresh the browser on reconnection failure.

But now that you called out https://github.com/dotnet/aspnetcore/issues/44669, I think this is something we already have support for, and it's what my sample in this PR utilizes. Our own [default reconnection handler](https://github.com/dotnet/aspnetcore/blob/921f5417741ca8086bd67d437ad40a91575095a1/src/Components/Web.JS/src/Platform/Circuits/DefaultReconnectionHandler.ts#L78-L94) even handles both cases explicitly ("failed" vs "rejected"). Maybe that issue can be closed? CC @SteveSandersonMS

guardrex:
@MackinnonBuck ... You might be waiting on Steve to look. I'll let you pull the trigger on merging this when ur ready.

PR #27735, opened 2022-11-24
Update configuration.md

masitsa:
The description of MaxSendMessageSize and MaxReceiveMessageSize do not accurately describe whether the it's a client to server request or a server to client response


JamesNK:
These settings are for server configuration. They control the server side.

The new descriptions in the PR aren't accurate.

PR #27744, opened 2022-11-25
Simplified the code so it renders correct HTML when embedded

Sire:
Fixed broken code render when embedded in documation on:

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-7.0#environments

(ASP.NET Core 3.1 to 7.0 variations uses the same code sample)

The environment tag helper code  &lt; was rendered as actual HTML tags in the embedded code example, creating invalid  code.

### DIFF
--- a/aspnetcore/fundamentals/environments/samples/6.x/EnvironmentsSample/Pages/About.cshtml
+++ b/aspnetcore/fundamentals/environments/samples/6.x/EnvironmentsSample/Pages/About.cshtml
@@ -19,15 +19,12 @@
 
 <!-- <snippet_Environment> -->
 <environment include="Development">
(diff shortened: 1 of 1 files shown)


guardrex:
Hello @Sire ... Just an FYI: This was by-design and used to work in the old days under a prior version of the build engine. I agree with you that this is a reasonable fix.

Rick-Anderson:
Thanks for the contribution. Be sure to check out your name as a contributor about the 3rd of next month on the [What's new page](https://docs.microsoft.com/en-us/aspnet/core/whats-new/) in the **Community contributors** section at the bottom.

PR #27754, opened 2022-11-26
Add a note about privilege escalation attacks using Graph permissions and public clients

damienbod:
I thought it would be good to add a note in the docs about graph permissions and privilege escalation attacks in public clients.  I see many public clients with too many Graph permissions which can be misused. The intended business created by using the permission is normally much less than the whole Graph permission(s) made public and does not need to be made available to all authenticated users. People should be made aware of this. A confidential client would be a better solution for Graph permissions apart from the basic delegated **User.Read** etc.

### DIFF
--- a/aspnetcore/blazor/security/webassembly/graph-api.md
+++ b/aspnetcore/blazor/security/webassembly/graph-api.md
@@ -19,7 +19,7 @@ This article explains how to use [Microsoft Graph API](/graph/use-the-api), whic
 [Microsoft Graph SDKs](/graph/sdks/sdks-overview) are designed to simplify building high-quality, efficient, and resilient applications that access Microsoft Graph.
 
 > [!NOTE]
-> Using Graph permissions in a public application can lead to an privilege escalation attack. Validate that the specified Graph permissions do not allow the user to do more than the intended use. Use confidential clients for sensitive Graph permissions.
+> Using Graph permissions in a public application can lead to a privilege escalation attack. Validate that the specified Graph permissions do not allow the user to do more than the intended use. Use confidential clients for sensitive Graph permissions.
 
 The examples in this section require package references for the standalone or **:::no-loc text="Client":::** app:
 
--- a/aspnetcore/blazor/security/webassembly/graph-api.md
+++ b/aspnetcore/blazor/security/webassembly/graph-api.md
@@ -19,7 +19,7 @@ This article explains how to use [Microsoft Graph API](/graph/use-the-api), whic
 [Microsoft Graph SDKs](/graph/sdks/sdks-overview) are designed to simplify building high-quality, efficient, and resilient applications that access Microsoft Graph.
 
 > [!NOTE]
-> Using Graph permissions in a public application can lead to an evaluated privilege attack. Validate that the specified Graph permissions do not allow the user to do more than the intended use. Use confidential clients for sensitive Graph permissions.
+> Using Graph permissions in a public application can lead to an privilege escalation attack. Validate that the specified Graph permissions do not allow the user to do more than the intended use. Use confidential clients for sensitive Graph permissions.
 
 The examples in this section require package references for the standalone or **:::no-loc text="Client":::** app:
 
--- a/aspnetcore/blazor/security/webassembly/graph-api.md
+++ b/aspnetcore/blazor/security/webassembly/graph-api.md
@@ -18,6 +18,9 @@ This article explains how to use [Microsoft Graph API](/graph/use-the-api), whic
 
 [Microsoft Graph SDKs](/graph/sdks/sdks-overview) are designed to simplify building high-quality, efficient, and resilient applications that access Microsoft Graph.
 
+> [!NOTE]
+> Using Graph permissions in a public application can lead to an evaluated privilege attack. Validate that the Graph permission does not allow the user to do more than the intended use. Use confidential clients for sensitive Graph permissions.
+
 The examples in this section require package references for the standalone or **:::no-loc text="Client":::** app:
 
 * [`Microsoft.Extensions.Http`](https://www.nuget.org/packages/Microsoft.Extensions.Http)
--- a/aspnetcore/blazor/security/webassembly/graph-api.md
+++ b/aspnetcore/blazor/security/webassembly/graph-api.md
@@ -19,7 +19,7 @@ This article explains how to use [Microsoft Graph API](/graph/use-the-api), whic
 [Microsoft Graph SDKs](/graph/sdks/sdks-overview) are designed to simplify building high-quality, efficient, and resilient applications that access Microsoft Graph.
 
 > [!NOTE]
-> Using Graph permissions in a public application can lead to an evaluated privilege attack. Validate that the Graph permission does not allow the user to do more than the intended use. Use confidential clients for sensitive Graph permissions.
+> Using Graph permissions in a public application can lead to an evaluated privilege attack. Validate that the specified Graph permissions do not allow the user to do more than the intended use. Use confidential clients for sensitive Graph permissions.
 
 The examples in this section require package references for the standalone or **:::no-loc text="Client":::** app:
 
--- a/aspnetcore/blazor/security/webassembly/graph-api.md
+++ b/aspnetcore/blazor/security/webassembly/graph-api.md
@@ -18,8 +18,12 @@ This article explains how to use [Microsoft Graph API](/graph/use-the-api), whic
 
 [Microsoft Graph SDKs](/graph/sdks/sdks-overview) are designed to simplify building high-quality, efficient, and resilient applications that access Microsoft Graph.
 
-> [!NOTE]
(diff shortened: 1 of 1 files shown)


guardrex:
Hello @damienbod ... A few points on this one ...

* I think this topic is either going to completely disappear from the Blazor docs node ***OR*** get distilled down to merely the aspects that pertain to MS Graph use in Razor components because MS Graph with Blazor and MS Graph docs in general have come a long way since Blazor came out a few years ago and we don't want to maintain more coverage for a non-Blazor technology than is needed here. However, there's nothing wrong with placing a new remark on any aspect of this scenario for prior versions, which will be 6.0 or earlier (probably). I'm mostly mentioning this because the whole doc might disappear in favor of one or more cross-links. We'll see shortly when I reach this node for work on #24615.
* WRT Azure/MS Graph security in general, it's better for us to maintain a NOTE that instructs readers to access the MS Identity/Azure/MS Graph security docs in their node(s) because they'll maintain higher quality and more extensive coverage than we can over here. I think the cross-links to consider for such a NOTE would be to the following articles. These are new or heavily revised articles that were written/revised after the Blazor security docs were written a few years ago, which is part of the reason why they aren't currently cross-linked. This is also a general aspect of our Blazor docs that I intend to address on #24615. All Blazor security doc cross-linking requires a round of improvement given that about three years have passed since many of these Blazor security docs were written.
  * [*Enhance security with the principle of least privilege*](https://learn.microsoft.com/azure/active-directory/develop/secure-least-privileged-access)
  * [*Microsoft Security Best Practices: Securing privileged access*](https://learn.microsoft.com/security/compass/overview)
  * [*Microsoft Graph auth overview*](https://learn.microsoft.com/graph/auth/) and [*Microsoft Graph permissions reference*](https://learn.microsoft.com/en-us/graph/permissions-reference)

I also found what I think is a great post on the subject by Andy Robbins (@wald0 *I think* ... forgive me @wald0 if you're not ***THE Andy Robbins!*** 😄) at https://posts.specterops.io/azure-privilege-escalation-via-azure-api-permissions-abuse-74aee1006f48.

### ***... and mentioning as usual*** 😄 ...

mccoy

### *Damn it Jim! I'm a silly 🦖, not a security expert!*

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&mdash; Dr. Leonard H. "Bones" McCoy ([DeForest Kelley](https://www.imdb.com/name/nm0001420/)) [Star Trek](http://www.startrek.com/), ©1966-69 [Paramount Pictures](http://www.paramount.com)

I'll ping Barry for an 👁️ on this and follow his advice on exact wording and any cross-links to use. Let's wait until Monday tho. I'm not sure he'd be looking today (Sunday), but I don't want to disturb him right now on a holiday weekend 🦃🍗😋. It's been a *long, and grueling year* around .NET. We need all the *undistrurbed* days off for R&R as possible 🏖️.

guardrex:
### ... and we're ***BACK!*** 🦃

I shot in an update to make this a bit less specific and rely on the cross-links that I found.

Barry, I originally thought that "privilege escalation" and "least privilege" weren't exactly the same thing, but I could be mistaken on that initial impression.

* How do you like the text in the WARNING?
* Are these cross-links ok?
* What about this Andy Robbins post ...

  https://posts.specterops.io/azure-privilege-escalation-via-azure-api-permissions-abuse-74aee1006f48

  Does that cross the threshold for *authoritative-enough* and *stable-enough* to be included in the cross-link list?

blowdart:
I'd ask you to go ask the graph api folks for their recommendations, or remove the topic entirely and ask them to either help draft something they know is secure, or, more preferrably document it under their own docs, and just direct people there. If we have our own then it's possible as they update their guidance we'll miss it and be out of date.

As an aside "privilege escalation" and "least privilege" are different, but related. privilege escalation is when an attacker can gain more rights that they had previously, least privilege is only ever granting the minimum amount of rights to do an operation. However if the minimum amount of rights to do something is more than an attackers default, and they can do the thing, that's an escalation.

guardrex:
> I'd ask you to go ask the graph api folks for their recommendations, or remove the topic entirely a

There was no coverage on Graph SDK for Blazor at the time this was written, so that's why this is here. It was a bit tricky back then to get working well. Part of my analysis and work for https://github.com/dotnet/AspNetCore.Docs/issues/24615 ***for December*** is indeed to see if we can drop this article in favor of linking to their docs. ***I would greatly prefer to take that approach.***

> privilege escalation is when an attacker can gain more rights that they had previously

That's what I thought. 

@damienbod ... We're going drop this for now to see if I can drop the coverage and cross-link entirely to their docs. This is already marked on my tracking issue so it won't get lost in case Dan decides later that we need to keep this based on the result of my analysis.

guardrex:
FYI @blowdart ... I'm analyzing the MS Identity platform doc set coverage on WASM with Graph, and it's not looking good. 

Their coverage is at ...

* **Tutorial**: https://learn.microsoft.com/en-us/azure/active-directory/develop/tutorial-blazor-webassembly
* **Sample App**: https://github.com/Azure-Samples/ms-identity-blazor-wasm

Some of the problems I was already aware of, and others have come to light on closer inspection ...

* They don't use the standard ASP.NET Core WASM project template, which is updated all the way to 7.0, ***of course*** 😄. They have their own in `Microsoft.Identity.Web.ProjectTemplates`, which ...
  * [Has only been updated to 5.0](https://github.com/AzureAD/microsoft-identity-web/blob/master/ProjectTemplates/templates/ComponentsWebAssembly-CSharp/Client/ComponentsWebAssembly-CSharp.Client.csproj).
  * Mixes MS Build properties and doesn't use the latest SDK approach for the 5.0 release: `netstandard2.1` TFM with a `RazorLangVersion`, which is in the project file, went away at 5.0 in favor at first of `net5.0` TFM with `UseBlazorWebAssembly` and an RID under the `Microsoft.NET.Sdk.Web` SDK, which is also in the project file. However, that approach went away at 5.0 RC1 with the switch to the `Microsoft.NET.Sdk.BlazorWebAssembly` SDK with only the need to state the TFM of `net5.0`. I'm sure you're aware that we stick with that approach today. The project file there is a bit of a 🚂💥 for a demo, not showing the best practice SDK.
* Old 3.1-era code, auth flow guidance, and a .NET CLI command is used/shown: The step-by-step guidance focuses on using ***Implicit Grant*** with a note tacked on at some point later that mentions Code+PKCE. The newer .NET CLI command is `dotnet new install`, not `dotnet new {-i|--install}` in the tutorial.
* The approaches don't match: The tutorial uses the named `HttpClient` with `AuthorizationMessageHandler` approach, while the sample app uses an `IAuthenticationProvider` with Graph API `GraphServiceClient`.
* The tutorial doesn't deserialize with the more-modern approach of `JsonPropertyNameAttribute` and `GetFromJsonAsync`. It parses the `JsonDocument` and uses `RootElement`, `GetProperty`, etc. calls to obtain the user's data.
* The sample app uses base classes for example components, such as a `UserProfile` component/`UserProfileBase` class, to demonstrate obtaining and displaying user data from Graph. That's fine; however, it's probably best to demonstrate the simplest use case with a simple component that doesn't rely upon inheritance. Multiple inheritance, perhaps demonstrated with a library, isn't being used, so why bother with the extra ceremony? While the sample app does that, the tutorial doesn't demo with inheritance, which I think is best educational approach. Again, there's another delta here between the tutorial and sample app on the approaches used.
* In our topic, we also show how to customize the user with claims based on Graph data with an `AccountClaimsPrincipalFactory` using both the named `HttpClient` approach and the `GraphServiceClient` (Graph API) approach. They don't have it in either form.

I don't think that I need to go any further. I wasn't impressed a few years ago with cross-linking to their topics when new, and I'm even less excited about it today. I think our Graph API topic is considerably better than their tutorial and sample app, even with the need now for me to perform an update pass on our topic.

I'm going to email Dan and Artak this afternoon to discuss these findings. Either I should just perform a pass on our topic and keep going with it, keeping it updated as I normally try to do, or we probably need to alert the MSI/Azure docs folks of these concerns, get a lot of fixes in over there (probably full overhauls of their tutorial and sample), and wait to drop our topic here until after they polish up the 🍎 a bit for readers.

@damienbod ... If you don't mind ... and ***IF*** Dan says that we should keep what we have, I'm going to roll your suggesion and the links I found into that work. I don't like taking a doc contribution credit away from a reader, but there will be several updates to the topic (probably), and it's probably best if I take the update that you pitched here into those changes on my upcoming branch.

guardrex:
@damienbod ... This week, I worked on the whole topic. I still can't say for sure if Dan and Artak will want to keep it. They may say that we need to reach out to Azure/Graph doc cats to see if they will take the whole topic off of our hands. I doubt that they'll want to kill it 🔪 tho. Our Blazor coverage for Graph is tailored to Blazor WASM scenarios and makes the implementations dead simple for base use cases. I did expand cross-links to address your issue here. I'll ping you on the PR, but I think I'm going to take the updates live fairly fast. This is mostly window dressing on what's there. It's not a radical overhaul.

@blowdart ... I'll get Dan and Artak ... and Javier ... to look over the result of this effort and decide how they'd like to proceed. I'll ping you just in case. I know you've said that you'll defer to Javier on the security aspects. I just like to keep you informed of things like this just in case you have feedback.

PR #27779, opened 2022-11-29
Add video link

samuelwine:
This video was particularly helpful for me to get an understanding of what each of these technologies are and their differences.


guardrex:
Hello @samuelwine ... We can't (usually) cross-link to videos (or blogs for that matter) due to framework churn. They go ***stale*** rather quickly. We don't have the capacity to keep up with all of the content in videos and blogs relative to each release. The framework changes, but videos and blogs are rarely ever updated by their authors ... even MS video and blog resources.

samuelwine:
That makes sense!
Thanks for responding.

PR #27808, opened 2022-12-01
Update PetsController.cs

vcrobe:


Rick-Anderson (commented):
Original is correct, your change is wrong. See https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.controllerbase.createdataction

PR #27825, opened 2022-12-03
Fix spacing in code demo Name claim and role claim mapping

damienbod:
spacing in code demo fix

https://learn.microsoft.com/en-us/aspnet/core/security/authentication/claims


guardrex (approved):
Thanks @damienbod! 🚀

PR #25628, opened 2022-04-14
gRPC Authentication & Authorization Version 6.0 update

wadepickett:
[Internal Review](https://review.docs.microsoft.com/en-us/aspnet/core/grpc/authn-and-authz?view=aspnetcore-6.0&branch=pr-en-us-25628)

Fixes #25137 

- Added a 6.0 version app samples for Ticketer Service and TicketerClient apps. (See download link at top of  doc)
- Updated content and code snippets to reflect .NET 6.0.

PR #27830, opened 2022-12-04
use correct framework

kasperk81:
netcoreapp was replaced with net since v5.

### DIFF
--- a/aspnetcore/grpc/authn-and-authz/sample/6.x/TicketerClient/TicketerClient.csproj
+++ b/aspnetcore/grpc/authn-and-authz/sample/6.x/TicketerClient/TicketerClient.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp6.0</TargetFramework>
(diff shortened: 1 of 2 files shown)


guardrex:
Hello @kasperk81 ... This may have been an oversight. Indeed, that TFM [isn't supported](https://learn.microsoft.com/en-us/dotnet/standard/frameworks#supported-target-frameworks). This sample was updated from an [earlier version that targeted `netcoreapp3.0`](https://github.com/dotnet/AspNetCore.Docs/blob/2906acafca569dfecae638b3b71e407ce733f1f3/aspnetcore/grpc/authn-and-authz/sample/5.x/TicketerClient/TicketerClient.csproj), and it probably escaped an update for that line on https://github.com/dotnet/AspNetCore.Docs/pull/25628. I'll ping Wade and James.

kasperk81:
thanks! nuget understands this format https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Frameworks/NuGetFrameworkFactory.cs#L677-L694 but it's better to use `net#.#` recommended tfm everywhere.

JamesNK (approved):

PR #27854, opened 2022-12-07
Clarifications on Blazor Hybrid static web assets

### LINKED ISSUE #27819: Clarify how static web assets work in Blazor Hybrid apps
We need to clarify a bit how static web assets work in Blazor Hybrid apps.

In a Blazor Hybrid app you have one or more `BlazorWebView` controls with a configured host file (typically `wwwroot/Index.html`). The host file path is relative to the project. All static web assets (scripts, CSS files, images, etc.) that are then referenced from that `BlazorWebView` are then relative to its configured host file. 

Static web assets from libraries get special generated paths: `_content/{PACKAGE ID}/`. These paths are logically sub paths of the apps wwwroot folder, even though they are actually coming from other packages or projects. The component specific CSS stye bundle will also be built at the root of the wwwroot folder (@javiercn is there any way to change or configure this location that we should document?).

If you configure a host file for a `BlazorWebView` that is in a subdirectory of the wwwroot folder (like `wwwroot/foobar/Index.html`) then you'll need to account for this when specifying relative paths to static web assets from libraries, as those will still be rooted in the wwwroot folder. Similarly, if you try to configure a host file to come from a library, or from a location other than the wwwroot folder, then this will make relative paths to other static web assets even more awkward. For most apps we recommend that the host file be at the root of the wwwroot folder of the app. 

@javiercn @MackinnonBuck @Eilon Please chime in if I got any of this wrong.

---
#### Document Details

* ID: 1b14a096-b1a3-2270-e57f-8238a25601cd
* Version Independent ID: 1b14a096-b1a3-2270-e57f-8238a25601cd
* Content: [ASP.NET Core Blazor Hybrid static files](https://learn.microsoft.com/en-us/aspnet/core/blazor/hybrid/static-files?view=aspnetcore-7.0)
* Content Source: [aspnetcore/blazor/hybrid/static-files.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/hybrid/static-files.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #27819

I seek to retain your, Eilon, examples via the list you provided because they provide the greatest clarity. We might need to play with the language and examples to get them just right.


aspnetcore/blazor/hybrid/static-files.md
```diff
@@ -176,9 +176,19 @@ In the following example component:
 
 ## Static assets limited to Razor components
 
-In scenarios where the app only uses static assets in Razor components, the static assets can be supplied from the app's web root (`wwwroot` folder).
+One or more `BlazorWebView` controls have a configured host file (:::no-loc text="HostPage":::), typically `wwwroot/index.html`. The :::no-loc text="HostPage"::: path is relative to the project. All static web assets (scripts, CSS files, images, and other files) that are referenced from a `BlazorWebView` are relative to its configured :::no-loc text="HostPage":::.
```
Eilon, line 179: I'm not sure I get why it's talking about "One or more BlazorWebView..." --> what is that in reference to?

Eilon (changes requested):
I think this is overall good, just one question.

aspnetcore/blazor/hybrid/static-files.md
danroth27: @guardrex In this case I think it's fine to just say "A `BlazorWebView` control has a configured host file...".

aspnetcore/blazor/hybrid/static-files.md
```diff
@@ -176,9 +176,19 @@ In the following example component:
...
+* `wwwroot/{PATH}/index.html`: All assets in the app's `wwwroot/{PATH}` folder are available using app web root relative paths. RCL static assets in `wwwroot/{PATH}` are ***not available*** because they would be in a non-existent theoretical location, such as `../../_content/{PACKAGE ID}/{PATH}`, which is ***not a supported relative path***.
+* `wwwroot/_content/{PACKAGE ID}/index.html`: All assets in the RCL's `wwwroot/{PATH}` folder are available using RCL web root relative paths. The app's static assets in `wwwroot/{PATH}` are ***not available*** because they would be in a non-existent theoretical location, such as `../../{PATH}`, which is ***not a supported relative path***.
+
+For most apps, we recommend placing the :::no-loc text="HostPage"::: at the root of the `wwwroot` folder of the app, which provides the greatest flexibility for supplying static assets from the app, RCLs, and via subfolders of the app and RCLs.
+
+The following examples demonstrate referencing static assets from the app's web root (`wwwroot` folder) with a :::no-loc text="HostPage"::: rooted in the `wwwroot` folder.
```
danroth27, line 191: @guardrex  The image link in the example below looks good, but I think instead of the text file example we should be showing how to load some JavaScript as a static web asset.

danroth27: No need to block this PR though on updating the example.

danroth27 (approved):
LGTM

aspnetcore/blazor/hybrid/static-files.md
guardrex: > instead of the text file

How about ***in addition to*** because it wasn't obvious (to me 🦖) how it was done exactly for something like file processing, which is more akin to the text file example here. How about I ***ADD*** a load JS static web asset to what's there, as it will only take a bit of extra coverage?

danroth27: > How about in addition to ?

Sure :smile:

PR #27875, opened 2022-12-09
Blazor MS Graph article updates

guardrex:
Addresses #24615

Thanks @hanslai and @damienbod. I'm going to sleep on these updates overnight and make some final touches on Saturday morning. Because of the number of updates here, you might want to wait until after tomorrow morning and inspect the live topic after these updates merge.

### DIFF
--- a/aspnetcore/blazor/security/webassembly/additional-scenarios.md
+++ b/aspnetcore/blazor/security/webassembly/additional-scenarios.md
@@ -5034,6 +5034,7 @@ The preceding example sets redirect URIs with regular string literals. The follo
 
 ## Additional resources
 
+* <xref:blazor/security/webassembly/graph-api>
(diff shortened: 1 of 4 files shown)


guardrex:
Ok ... LGTM 🦖 for a set of cool updates 😎. The base use cases here are the same as before, so I'm not too crazy worried about a technical blunder at this point. I do want to get some eyes on this 👀, but I'm going to do that on the live topic with the new content layout for this guidance. For one thing, I'm making the main approaches ... Graph SDK vs. named `HttpClient`/Graph API ... less confusing by using a doc pivot. I want to see how that renders and see what peeps think about it.

I'm going to take this whole subject up with DR and Artak offline for further discussion about keeping/dropping the topic or moving it to the Azure/MSI docs.

PR #27909, opened 2022-12-14
Access services from a different DI scope

### LINKED ISSUE #27908: Document how to access Blazor Services from another DI service scope.
This issue applies to Blazor Server specifically.

We've received some issues like https://github.com/dotnet/aspnetcore/issues/25758 and https://github.com/dotnet/aspnetcore/issues/40336 where customers try to get Blazor services from an `IServiceProvider` created from a different DI scope. In both the previous linked issues, an `IHttpClientFactory` was used to create an `HttpClient` configured to utilize a Blazor service in some way. This fails because these `HttpClient`s will have their own DI service scopes, so they can't access Blazor services directly.

We should document how customers can access Blazor Server services from code where the Blazor `IServiceProvider` is not already accessible, specifically when that code is invoked from a Blazor component.

One approach to make this work is by using an `AsyncLocal` to capture the Blazor component's `IServiceProvider` so it can be retrieved later in the async call stack.

In the future, we hope to add a product feature that addresses this issue more fundamentally, making the code provided by this example unnecessary.

guardrex:
Fixes #27908

[Internal Review Topic (links to section)](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/dependency-injection?view=aspnetcore-8.0&branch=pr-en-us-27909#access-blazor-services-from-a-different-di-scope)

Mackinnon:

The Blazor DI topic is probably the best spot for this.

Questions ❓ ...

1. Is `IJSInterop` a thing? I can't find that API. Did you mean `IJSRuntime`?
1. Is the "Blazor `IServiceProvider`" a `System.IServiceProvider`? I ask to make sure that I have the API cross-links correct.
1. I removed the `_` from `hasCalledOnAfterRender`. I was told that for developer code not to sweat it ... that it was more for framework purposes. I might be a bit off on that tho because the discussion surrounding using the underscore on fields was focused on how we would ***display Razor components to readers***, not so much for classes that we display to readers. Do you want me to add it back?
1. What versions can I display this for? ... Can this go all the way back to the 3.1 version of the topic?

### DIFF
--- a/aspnetcore/blazor/fundamentals/dependency-injection.md
+++ b/aspnetcore/blazor/fundamentals/dependency-injection.md
@@ -970,20 +970,38 @@ There may be times when a Razor component invokes asynchronous methods that exec
 
 For example, <xref:System.Net.Http.HttpClient> instances created using <xref:System.Net.Http.IHttpClientFactory> have their own DI service scope. As a result, <xref:System.Net.Http.HttpMessageHandler> instances configured on the <xref:System.Net.Http.HttpClient> aren't able to directly inject Blazor services.
 
-Create a static class `BlazorServiceAccessor` that defines an [`AsyncLocal<IServiceProvider>`](xref:System.Threading.AsyncLocal`1), which stores the Blazor <xref:System.IServiceProvider> for the current asynchronous context. Asynchronous code invoked from a Razor component can use `BlazorServiceAccessor.Services` to access Blazor services. *This is only necessary if the <xref:System.IServiceProvider> available to the invoked code was created in a different DI service scope.*
+Create a class `BlazorServiceAccessor` that defines an [`AsyncLocal`](xref:System.Threading.AsyncLocal`1), which stores the Blazor <xref:System.IServiceProvider> for the current asynchronous context. A `BlazorServiceAcccessor` instance can be acquired from within a different DI service scope to access Blazor services.
 
 `BlazorServiceAccessor.cs`:
 
 ```csharp
-internal static class BlazorServiceAccessor
+internal sealed class BlazorServiceAccessor
 {
-    private static readonly AsyncLocal<IServiceProvider> s_blazorServices = new();
+    private static readonly AsyncLocal<BlazorServiceHolder> s_currentServiceHolder = new();
 
-    public static IServiceProvider Services
+    public IServiceProvider? Services
     {
-        get => s_blazorServices.Value ?? throw new InvalidOperationException(
-            "Blazor services are not available in the current context.");
-        set => s_blazorServices.Value = value;
+        get => s_currentServiceHolder.Value?.Services;
+        set
+        {
+            if (s_currentServiceHolder.Value is { } holder)
+            {
+                // Clear the current IServiceProvider trapped in the AsyncLocal.
+                holder.Services = null;
+            }
+
+            if (value is not null)
+            {
+                // Use object indirection to hold the IServiceProvider in an AsyncLocal
+                // so it can be cleared in all ExecutionContexts when it's cleared.
+                s_currentServiceHolder.Value = new() { Services = value };
+            }
+        }
+    }
+
+    private sealed class BlazorServiceHolder
+    {
+        public IServiceProvider? Services { get; set; }
     }
 }
 ```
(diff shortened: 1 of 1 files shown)


MackinnonBuck:
1. Yes, sorry that was a typo on my part. I meant `IJSRuntime` 😄 
2. Yup!
3. It's probably fine to remove the underscore. This example actually copy/pastes some code from our own `ComponentBase`, which uses the framework's naming conventions. But since we're presenting it as developer code, I agree it's fine to remove!
4. Yes, I believe this can go back to the 3.1 version of the topic.

MackinnonBuck:
Adding @javiercn as a reviewer since he has a lot of context in this area.

guardrex:
Ok ... I'll wait for Javier to take a look 👀.

🎗️ ***NOTE TO SELF***: *Duplicate for each of the earlier versions before merging!* 🎗️

aspnetcore/blazor/fundamentals/dependency-injection.md
```diff
@@ -961,6 +961,111 @@ The app can register transient disposables without throwing an exception. Howeve
...
+  
+## Access Blazor services from a different DI scope
+  
+*This section only applies to Blazor Server apps.**
+
+There may be times when a Razor component invokes asynchronous methods that execute code in a different DI scope. Without the correct approach, these DI scopes don't have access to Blazor's services, such as <xref:Microsoft.JSInterop.IJSRuntime> and <xref:Microsoft.AspNetCore.Components.Server.ProtectedBrowserStorage>.
+
+For example, <xref:System.Net.Http.HttpClient> instances created using <xref:System.Net.Http.IHttpClientFactory> have their own DI service scope. As a result, <xref:System.Net.Http.HttpMessageHandler> instances configured on the <xref:System.Net.Http.HttpClient> aren't able to directly inject Blazor services.
+
+Until this scenario is addressed as a product feature in a future Blazor release, the following workaround is available to configure accessible Blazor services from code in other DI scopes when invoked from a Razor component.
```
javiercn, line 973: I would remove this sentence, let's not make any promises (even if we plan to do this) and let's not call this a workaround since it is the proper way to do it.

aspnetcore/blazor/fundamentals/dependency-injection.md
```diff
@@ -961,6 +961,111 @@ The app can register transient disposables without throwing an exception. Howeve
...
+{
+    private static readonly AsyncLocal<IServiceProvider> s_blazorServices = new();
+
+    public static IServiceProvider Services
+    {
+        get => s_blazorServices.Value ?? throw new InvalidOperationException(
+            "Blazor services are not available in the current context.");
+        set => s_blazorServices.Value = value;
+    }
+}
```
javiercn, line 990: Let's follow the pattern that we use in other places. Register this as a singleton service in DI and have the component class inject this type (like HttpContextAccessor).

aspnetcore/blazor/fundamentals/dependency-injection.md
```diff
@@ -961,6 +961,111 @@ The app can register transient disposables without throwing an exception. Howeve
...
+            throw;
+        }
+
+        StateHasChanged();
+    }
+
+    private async Task InvokeWithBlazorServiceContext(Func<Task> func)
+    {
+        BlazorServiceAccessor.Services = Services;
+        await func();
```
javiercn, line 1061: Put this inside a `try..finally` and set `.Services` to null inside the finally

aspnetcore/blazor/fundamentals/dependency-injection.md
javiercn, line 988: I am wondering if we need a couple of things
```csharp
internal static class BlazorServiceHolder
{
    private static readonly AsyncLocal<IServiceProvider> s_blazorServices = new();
    public static IServiceProvider Services
    {
        get => s_blazorServices.Value ?? throw new InvalidOperationException(
            "Blazor services are not available in the current context.");
        set => s_blazorServices.Value = value;
    }
}

internal static class BlazorServiceAccessor : IDisposable
{
    public BlazorServiceAccessor(BlazorServiceHolder holder)
    {
       _holder = holder;
    }

    public T GetService<T>() => _holder.ServiceProvider.GetRequiredService<T>();

    Dispose()
    {
        _holder = null;
    }
}
```

The class that sets the context injects the holder, the class that uses the service injects the accessor.

I believe that HttpContextAccessor does the same thing or something very similar.

aspnetcore/blazor/fundamentals/dependency-injection.md
guardrex: IntelliSense is telling me that the injected Blazor `IServiceProvider` is non-nullable, so it's not happy with me trying to assign `null` to it.

MackinnonBuck:
Thanks, @javiercn! I've adjusted the sample to be similar to how `HttpContextAccessor` is implemented.

javiercn (approved):

javiercn:
@guardrex is this content live? I can find it if I search for it, but when I try to access it is not on the page.

guardrex:
😈 Thanks for catching it. My bad ... There's a ***NOTE TO SELF*** that I didn't catch while merging. It's going into each version right now and will be live in a few minutes.

Currently, it is available where it was placed ... 6.0 ...

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/dependency-injection?view=aspnetcore-6.0#access-blazor-services-from-a-different-di-scope

PR #27916, opened 2022-12-16
update link fragment

sei-jmattson:

### DIFF
--- a/aspnetcore/web-api/action-return-types.md
+++ b/aspnetcore/web-api/action-return-types.md
@@ -156,7 +156,7 @@ The `HttpResults` can be useful when sharing code between Minimal APIs and Web A
 
 The <xref:Microsoft.AspNetCore.Http.HttpResults> namespace contains classes that implement the <xref:Microsoft.AspNetCore.Http.IResult> interface. The `IResult` interface defines a contract that represents the result of an HTTP endpoint. The static [Results](<xref:Microsoft.AspNetCore.Http.Results>) class is used to create varying `IResult` objects that represent different types of responses.
 
-The [Built-in results](/aspnet/core/fundamentals/minimal-apis?view=aspnetcore-7.0&preserve-view=true#bimr7) table shows the common result helpers.
+The [Built-in results](/aspnet/core/fundamentals/minimal-apis?view=aspnetcore-7.0&preserve-view=true#built-in-results) table shows the common result helpers.
 
 Consider the following code:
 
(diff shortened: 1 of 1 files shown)


aspnetcore/web-api/action-return-types.md
```diff
@@ -156,7 +156,7 @@ The `HttpResults` can be useful when sharing code between Minimal APIs and Web A
 
 The <xref:Microsoft.AspNetCore.Http.HttpResults> namespace contains classes that implement the <xref:Microsoft.AspNetCore.Http.IResult> interface. The `IResult` interface defines a contract that represents the result of an HTTP endpoint. The static [Results](<xref:Microsoft.AspNetCore.Http.Results>) class is used to create varying `IResult` objects that represent different types of responses.
 
-The [Built-in results](/aspnet/core/fundamentals/minimal-apis?view=aspnetcore-7.0&preserve-view=true#bimr7) table shows the common result helpers.
+The [Built-in results](/aspnet/core/fundamentals/minimal-apis?view=aspnetcore-7.0&preserve-view=true#built-in-results) table shows the common result helpers.
```
Rick-Anderson, line 159: custom html links are preferred because they don't break with each version upgrade. What's wrong with `#bimr7` ?

soon2849: 감사합니다

sei-jmattson: Thanks. Sorry to bother.  I assumed there was a reason for the `#bimr7` fragment;  however in this case it just goes to the top of a long page, rather than directly to the table being referenced.  Is there some magic that should translate the `#bimr7` fragment to the `id="built-in-results"` on the destination page?

Rick-Anderson: It's a standard HTML in page anchor, we should figure out why it's not working. Perhaps there's a duplicate on the page. Use the F12 tools to see what's generated.

Rick-Anderson: Can you provide a link to `/aspnet/core/fundamentals/minimal-apis?view=aspnetcore-7.0&preserve-view=true#built-in-results`
I don't even see that H2.

sei-jmattson: Looks to be an H3.

Screen Shot 2023-01-06 at 11 26 55 AM

sei-jmattson: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis?view=aspnetcore-7.0&preserve-view=true#built-in-results

Rick-Anderson:
@sei-jmattson https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis?view=aspnetcore-7.0&preserve-view=true#binr7 works

aspnetcore/web-api/action-return-types.md
Rick-Anderson, line 159:
```suggestion
The [Built-in results](/aspnet/core/fundamentals/minimal-apis?view=aspnetcore-7.0&preserve-view=true#binr7) table shows the common result helpers.
```

sei-jmattson:
Ah, cool.  Didn't realize there was an anchor in preceding paragraph.

PR #27923, opened 2022-12-17
[Blazor] Add image upload preview guidance

### LINKED ISSUE #26276: The "Preview an image provided by the InputFile component" content seems strange
A couple of aspects of the "Preview an image provided by the InputFile component" content seem odd to me:

 * Why does it round-trip the data to .NET? That's potentially very expensive, and could be avoided. Instead, the .NET side already knows the `Name` of the file being uploaded, so it can send that to some JS function `setImage` along with a reference to the `<input type=file>` element, and then the JS code can read `inputFileElement.files` to look up the corresponding file data without it ever having to be sent to .NET
 * Why does it claim that "Because setImage is called from a form, it isn't necessary for the setImage function to call [revokeObjectURL](https://developer.mozilla.org/docs/Web/API/URL/revokeObjectURL)"? I don't know what "form" is being referred to here, or why that has any bearing on the fact that `createObjectURL` will leak memory if you fail to revoke the object URL later. As far as I'm aware, it *is* necessary to revoke the object URL.

---
#### Document Details

* ID: 66bb1712-60ac-77a6-10d2-f2d795126783
* Version Independent ID: 66bb1712-60ac-77a6-10d2-f2d795126783
* Content: [Work with images in ASP.NET Core Blazor](https://docs.microsoft.com/en-us/aspnet/core/blazor/images?view=aspnetcore-6.0#feedback)
* Content Source: [aspnetcore/blazor/images.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/images.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **taparik**

TanayParikh:
Fixes: https://github.com/dotnet/aspnetcore/issues/27983
Fixes #26276

### DIFF
--- a/aspnetcore/blazor/file-uploads.md
+++ b/aspnetcore/blazor/file-uploads.md
@@ -575,7 +575,7 @@ In JavaScript, add a function called with an HTML [`input`](https://developer.mo
 
 * Extracts the selected file.
 * Creates an object URL with [`createObjectURL`](https://developer.mozilla.org/docs/Web/API/URL/createObjectURL).
-* Includes an event listener to revoke the object URL with [`revokeObjectURL`](https://developer.mozilla.org/docs/Web/API/URL/revokeObjectURL) after the image is loaded, so memory isn't leaked.
+* Sets an event listener to revoke the object URL with [`revokeObjectURL`](https://developer.mozilla.org/docs/Web/API/URL/revokeObjectURL) after the image is loaded, so memory isn't leaked.
 * Sets the `img` element's source to display the image.
 
 ```javascript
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/file-uploads.md
```diff
@@ -548,6 +548,52 @@ In Blazor WebAssembly, file data is streamed directly into the .NET code within
 
 :::zone-end
 
+## Upload image preview
```
TanayParikh, line 551: Do we need to copy this below for .NET 6.0?

guardrex: Yes, but I'll take care of it on Monday morning.

Does this PR cover #26276?

guardrex (approved):

PR #27932, opened 2022-12-19
Highlight code for suppress map client errors

scottsauber:
Previously this was highlighting some code that wasn't super relevant (unless I'm missing something). Now this will highlight the line mentioned above the snippet.

Current state: 
image


guardrex (approved):
Thanks @scottsauber ... Let's go with that block (lines 4-7).

PR #27661, opened 2022-11-16
Add Output Caching in list of middleware

sebastienros:

PR #27940, opened 2022-12-20
Move and rename "ASP.NET Core Performance Best Practices" doc

### LINKED ISSUE #26643: Move and rename "ASP.NET Core Performance Best Practices" doc
[The "ASP.NET Core Performance Best Practices" doc](https://docs.microsoft.com/en-us/aspnet/core/performance/performance-best-practices?view=aspnetcore-6.0) covers more than just performance. We already include general guidance in sections like [Do not use the HttpContext after the request is complete](https://docs.microsoft.com/en-us/aspnet/core/performance/performance-best-practices?view=aspnetcore-6.0#do-not-use-the-httpcontext-after-the-request-is-complete) which is more of a correctness than a performance thing.

I think this guidance is too important to hide in the `Performance > Overview` section.

Performance > Overview

I propose renaming it to "ASP.NET Core Best Practices" and making it a top-level document.

See https://github.com/dotnet/aspnetcore/issues/43080#issuecomment-1205168252 for context. We want to add a section about properly handling a null `HttpRequest.ContentLength` and this seemed like the right document even though it's not perf-related, and that brought this up.

@davidfowl @Rick-Anderson @guardrex 

---
[Associated WorkItem - 55893](https://dev.azure.com/msft-skilling/Content/_workitems/edit/55893)

tdykstra:
Fixes #26643 

Moving this doc to Fundamentals left us with no overview article for the Performance TOC section, so this PR creates a simple one with a list of links.

[Internal review URL](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/best-practices?view=aspnetcore-8.0&branch=pr-en-us-27940)

### DIFF
--- a/.openpublishing.redirection.json
+++ b/.openpublishing.redirection.json
@@ -292,7 +292,12 @@
         },
         {
             "source_path": "aspnetcore/performance/measuring.md",
-            "redirect_url": "/aspnet/core/performance/",
+            "redirect_url": "/aspnet/core/performance/overview",
+            "redirect_document_id": false
+        },
+        {
+            "source_path": "aspnetcore/performance/performance-best-practices.md",
+            "redirect_url": "/aspnet/core/fundamentals/best-practices",
             "redirect_document_id": false
         },
         {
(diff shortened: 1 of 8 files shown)


Rick-Anderson (approved):

Rick-Anderson:
@tdykstra  BTW, UID's are like a PK, problematic when you change them. I just had to fix this https://github.com/dotnet/AspNetCore.Docs/pull/27661/commits/571e3efbb679e3691e68da36765ac804ddf1508b

No reason to change.

PR #27966, opened 2022-12-23
Add dotnet tools to PATH

themagiccog:

### DIFF
--- a/aspnetcore/tutorials/first-web-api.md
+++ b/aspnetcore/tutorials/first-web-api.md
@@ -297,6 +297,8 @@ The preceding commands:
 * Add NuGet packages required for scaffolding.
 * Install the scaffolding engine (`dotnet-aspnet-codegenerator`) after uninstalling any possible previous version.
 
+Note: add dotnet tools to PATH. (i.e. export PATH=$HOME/.dotnet/tools:$PATH)
+
 Build the project.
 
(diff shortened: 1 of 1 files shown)


guardrex:
Hello @themagiccog ... This shouldn't be necessary. The SDK installer takes care of it. If you can repro it ***not*** behaving, please open an issue here with the steps so that they can investigate the problem with you ...

https://github.com/dotnet/sdk/issues

themagiccog:
You can reproduce the issue by simply working thru the steps in the document.

guardrex:
It has nothing to do with the document.

How did you install .NET on your system?

themagiccog:
I followed this:
https://dotnet.microsoft.com/en-us/download

guardrex:
Good, that's the correct approach if one isn't using VS to receive .NET.

You can try uninstalling .NET, confirming that the PATH entry is removed (or remove it manually), and re-installing from the SDK installer ***as an admin*** to see if that successfully updates the PATH. That's the only thing that leaps to mind. If it still doesn't make the PATH entry, then you'll need to open an issue for the team that works on the SDK at ...

https://github.com/dotnet/sdk/issues

... which is (for the most part) a different team than the ASP.NET Core product unit. They'll investigate with you why the PATH isn't being set properly.

PR #27988, opened 2022-12-30
Fixed typo mistake in Default host configuration sources Section.

3ilal:
There is a typographical mistake in the Default host configuration sources where the word '**_the_**' is repeated multiple times.
Doc Link -> [Link](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-7.0#default-host-configuration-sources)


guardrex (approved):
Thanks @3ilal! :rocket:

PR #27990, opened 2022-12-30
Fix Options Delegate link

secondsleep-blaine:
The **Configure options with a delegate** section has a typo in the link to the `Test2.cshtml.cs` code sample (`optionssamples` in path should be `options/samples`).


guardrex (approved):
Thanks @secondsleep-blaine! 🚀 ... I'll merge this to live immediately given that it breaks the snippet in the live doc. It should show up on the live site within ten minutes or so.

PR #28000, opened 2023-01-03
Return DTO with an identifier generated by EF

ganesanarun:
In the DTO approach, right now the response from create API does not return the id, this change would fix it.

### DIFF
--- a/aspnetcore/tutorials/min-web-api/samples/7.x/todoDTO/Program.cs
+++ b/aspnetcore/tutorials/min-web-api/samples/7.x/todoDTO/Program.cs
@@ -45,7 +45,7 @@ static async Task<IResult> CreateTodo(TodoItemDTO todoItemDTO, TodoDb db)
     db.Todos.Add(todoItem);
     await db.SaveChangesAsync();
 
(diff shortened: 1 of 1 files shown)


ganesanarun:
I understand that this tutorial is about asp.net core minimal API, I can't resist ask, should we introduce the Deconstruct method in Todo class?

Rick-Anderson:
@wadepickett  be sure to test original code and this change.

wadepickett (changes requested):
The suggested change would not compile as is.

wadepickett:
@Rick-Anderson and @ganesanarun,  I may not be catching on to what needs to be done here.  id does return with the code as it already was. 

{
    "id": 1,
    "name": null,
    "isComplete": false
}

The changed line suggested in this PR at this moment however will not compile, nor would I expect it to.  todo is not defined.

ganesanarun:
I agree that 

return TypedResults.Created($"/todoitems/{todoItem.Id}", new TodoItemDTO(todo)); should be 
TypedResults.Created($"/todoitems/{todoItem.Id}", new TodoItemDTO(todoItem)); 

but I don't think the current create API returns id at all.

anyway it's up to you folks. @Rick-Anderson @wadepickett

PR #28050, opened 2023-01-10
Remove dup package ref

guardrex:
Noticed in passing.


X-Celcius:
How was this a duplicate package ref? 
`WindowsServiceHelpers` and `UseWindowsService` is clearly defined in Microsoft.**Extensions**.Hosting.WindowsServices. 

If I download and build the project it now fails: 
`error CS0234: The type or namespace name 'WindowsServices' does not exist in the namespace 'Microsoft.Extensions.Hosting' (are you missing an assembly reference?)`

I think you wanted to remove the implicit reference to `Microsoft.AspNetCore.Hosting.WindowsServices` and wanted to leave the reference to `Microsoft.Extensions.Hosting.WindowsServices`.

guardrex:
Thanks for the catch, @X-Celcius ... I'll fix this right now.

Issue #25572, opened 2022-04-08
.NET 6 changes for Host ASP.NET Core in a Windows Service


Rick-Anderson:
* Copy over [Create the Windows Service](https://docs.microsoft.com/en-us/dotnet/core/extensions/windows-service#create-the-windows-service)
* Update all the old SDK and other versions to the current version.
* [

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b0bd8cb8-7e28-8c4d-a550-3aea8f3ab4dd
* Version Independent ID: 4b60d24e-ef05-15e5-10b4-483094081bf3
* Content: [Host ASP.NET Core in a Windows Service](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/windows-service?view=aspnetcore-6.0&tabs=visual-studio)
* Content Source: [aspnetcore/host-and-deploy/windows-service.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/host-and-deploy/windows-service.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-hostdeploy**
* GitHub Login: @Rick-Anderson
* Microsoft Alias: **riande**

Rick-Anderson:
@halter73  please review outline for changes.

halter73:
The outline looks good to me. As for "What parts of the article change by using the new minimal hosting model?", it's still being designed. We're going to have to wait on the outcome of https://github.com/dotnet/runtime/pull/68580.

Rick-Anderson:
@halter73 can you take another look

PR #28048, opened 2023-01-10
Fix/add missing use windows service dotnet7

maroonedsia:
[EDIT by guardrex to change the issue reference.]

Fixes #28055

* Compiling \aspnetcore\host-and-deploy\windows-service\samples\7.x\WebAppServiceSample and trying to add a Windows service for the executable was failing.
* Adding builder.Host.UseWindowsService() to the program.cs fixed it for me.

PR #28059, opened 2023-01-10
Update windows-service.md

halter73:
I looked at this since I got pinged on #25572. `WebApplication.CreateBuilder` doesn't call `AddWindowsService`. You still need to call that manually. The worker template still uses `Host.CreateDefaultBuilder(args)`, so mentioning `WebApplication` could be confusing. Fortunately, `AddWindowsService` works both on `builder.Services` and in a `ConfigureServices` callback.

### DIFF
--- a/aspnetcore/host-and-deploy/windows-service.md
+++ b/aspnetcore/host-and-deploy/windows-service.md
@@ -30,8 +30,7 @@ The ASP.NET Core Worker Service template provides a starting point for writing l
 
 ## App configuration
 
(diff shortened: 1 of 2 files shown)


halter73:
I pushed another commit to call `builder.Services.AddWindowsService()` instead of `builder.Host.UseWindowsService()` which was added by #28048. They both do the same thing, but we want to avoid referencing `builder.Host` in places we can since the builder has to contort itself to support the old API which can lead to runtime issues in some situations like when modifying the content root path.

Rick-Anderson (approved):

PR #28077, opened 2023-01-12
Group all properties together

vanillajonathan:

PR #28082, opened 2023-01-12
Proposed new PaginatedList.cs

Rick-Anderson:
See #28077
@vanillajonathan we don't have time to test this and update the tutorial with the changes, but we'll consider it when we update this series for .NET 8

cc @tdykstra


aspnetcore/data/ef-rp/intro/samples/cu80/PaginatedList.cs
```diff
@@ -0,0 +1,34 @@
+using System;
```
Rick-Anderson, line 1:
```suggestion
// See https://github.com/dotnet/AspNetCore.Docs/pull/28082
// This is proposed
using System;
```

PR #28091, opened 2023-01-13
JS interop disposal guidance updates

### LINKED ISSUE #27536: Improve coverage of Blazor JS interop disposal patterns
Hey @guardrex, here's another important corner case.

The blazor docs  cover js module disposal [here](https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-javascript-from-dotnet?view=aspnetcore-6.0#javascript-isolation-in-javascript-modules).

However if javascript event handlers are registered, they should (probably?!) be [released](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener), else they're liable to cause hard to debug problems.

An interesting solution can be found [on SO](https://stackoverflow.com/a/70614374/9971404) by @jessegood. In summary:

`MyComponent.razor.js`
````cs
export function dispose()
{
  _myElement.removeEventListener("mousedown", handleMouseDown);
}
````

`MyComponent.razor`
````cs
@implements IAsyncDisposable
@code {
  private IJSObjectReference? _module;
  public async ValueTask DisposeAsync()
  { 
    if (_module != null)
    {
      await _module.InvokeVoidAsync("dispose");        // <---------------
      await _module.DisposeAsync();
    }
  }
}
````

---
#### Document Details

* Content: [Call JavaScript functions from .NET methods in ASP.NET Core Blazor](https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-javascript-from-dotnet?view=aspnetcore-6.0)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

### PR DESCRIPTION
**Pull request by guardrex**
Fixes #27536

Thanks @lonix1! 🎸 

There are two side updates that I think must be accomplished and are on this PR:

* As part of the work for this PR, I found an example that doesn't show the proper disposal pattern.
* I think it's a good idea to call out that trying to perform any kind of DOM cleanup on location changes is a BAD idea. I add an IMPORTANT note to the *Routing* article.

This PR adds new section to the *Lifecycle* topic and the two JS interop topics that ***reiterates to reinforce*** all of the existing demonstrations and written content on disposal of JS interop objects.

We exclusively show `IJSObjectReference`/`IJSInProcessObjectReference` created and disposed ***on the .NET side***. If there was something to say about disposal of that on the JS side, I'm not sure what to say and show in code. Currently, this new section states ...

> \* When calling JS from .NET, as described in \<xref:blazor/js-interop/call-javascript-from-dotnet>, dispose any created \<xref:Microsoft.JSInterop.IJSObjectReference>/\<xref:Microsoft.JSInterop.IJSInProcessObjectReference> either from .NET or from JS to avoid leaking JS memory.

... which mirrors your (Javier's) remarks at https://github.com/dotnet/aspnetcore/issues/44960#issuecomment-1308481061.

However, we're not explaining in any text or code example the disposal in JS of objects that triggers automatic .NET-side disposal (if I'm understanding this correctly).

Outside of that, I'm 👂 for anything you'd like to add and any changes to the language.

### DIFF
--- a/aspnetcore/blazor/components/lifecycle.md
+++ b/aspnetcore/blazor/components/lifecycle.md
@@ -238,6 +238,18 @@ Components shouldn't need to implement <xref:System.IDisposable> and <xref:Syste
 
 Developer code must ensure that <xref:System.IAsyncDisposable> implementations don't take a long time to complete.
 
+### Disposal of JavaScript interop object references
+
+Examples throughout the [JavaScript (JS) interop articles](xref:blazor/js-interop/index) demonstrate typical object disposal patterns:
+
+* When calling JS from .NET, as described in <xref:blazor/js-interop/call-javascript-from-dotnet>, dispose any created <xref:Microsoft.JSInterop.IJSObjectReference>/<xref:Microsoft.JSInterop.IJSInProcessObjectReference> either from .NET or from JS to avoid leaking JS memory.
+
+* When calling .NET from JS, as described in <xref:blazor/js-interop/call-dotnet-from-javascript>, dispose of a created <xref:Microsoft.JSInterop.DotNetObjectReference> either from .NET or from JS to avoid leaking .NET memory.
+
+JS interop object references are implemented as a map keyed by an identifier on the side of the JS interop call that creates the reference. When object disposal is initiated from either the .NET or JS side, Blazor removes the entry from the map, and the object can be garbage collected as long as no other strong reference to the object is present.
+
+At a minimum, always dispose objects created on the .NET side to avoid leaking .NET managed memory.
+
 ### Document Object Model (DOM) cleanup tasks during component disposal
 
 Don't execute JS interop code for DOM cleanup tasks during component disposal. Instead, use the [`MutationObserver`](https://developer.mozilla.org/docs/Web/API/MutationObserver) pattern in JavaScript on the client for the following reasons:
--- a/aspnetcore/blazor/components/lifecycle.md
+++ b/aspnetcore/blazor/components/lifecycle.md
@@ -242,7 +242,7 @@ Developer code must ensure that <xref:System.IAsyncDisposable> implementations d
 
 Examples throughout the [JavaScript (JS) interop articles](xref:blazor/js-interop/index) demonstrate typical object disposal patterns:
 
-* When calling JS from .NET, as described in <xref:blazor/js-interop/call-javascript-from-dotnet>, dispose any created <xref:Microsoft.JSInterop.IJSObjectReference>/<xref:Microsoft.JSInterop.IJSInProcessObjectReference> either from .NET or from JS to avoid leaking JS memory.
+* When calling JS from .NET, as described in <xref:blazor/js-interop/call-javascript-from-dotnet>, dispose any created <xref:Microsoft.JSInterop.IJSObjectReference>/<xref:Microsoft.JSInterop.IJSInProcessObjectReference>/`JSObjectReference` either from .NET or from JS to avoid leaking JS memory.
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by mkArtakMSFT** [comment]
@TanayParikh can you please review this? Thanks!

**Review by TanayParikh — APPROVED** [review]
Thanks

**Comment by guardrex** [comment]
@TanayParikh ... What about that one question tho ☝️ ...

We exclusively show `IJSObjectReference`/`IJSInProcessObjectReference` created and disposed on the .NET side. If there was something to say about disposal of that on the JS side, I'm not sure what to say and show in code. I really wasn't aware that it was possible to create/dispose of it on the JS side.

**Comment by TanayParikh** [comment]
The `IJSObjectReference` is itself `IAsyncDisposable`:

https://github.com/dotnet/aspnetcore/blob/bb2d9ac2c943a3dbc94144e2e07f48f60b65fe65/src/JSInterop/Microsoft.JSInterop/src/IJSObjectReference.cs#L12

Likewise, it should be possible to call `disposeJSObjectReference`:

https://github.com/dotnet/aspnetcore/blob/bb2d9ac2c943a3dbc94144e2e07f48f60b65fe65/src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts#L193-L204

Could you please try that out and confirm it's working as expected?

**Comment by guardrex** [comment]
Thanks @TanayParikh ... that's perfect. That 2nd one ... `disposeJSObjectReference` ... I wasn't aware of that. I'll research it, confirm a working example, and add coverage on it.

**Comment by guardrex** [comment]
Ah, ha! I see it now ... where we cover creating a `JSObjectReference` and have a missing example of the disposal pattern. Our coverage is at ...

https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-dotnet-from-javascript?view=aspnetcore-7.0#create-javascript-object-and-data-references-to-pass-to-net

... and I detect a ***NIT*** 😈 that I need to fix here, too.

The example is ...

```javascript
DotNet.invokeMethodAsync('{APP NAMESPACE}', 'ReceiveWindowObject', 
  DotNet.createJSObjectReference(window));
```

```csharp
[JSInvokable]
public void ReceiveWindowObject(IJSObjectReference objRef)
{
    ...
}
```

The ***NIT*** is that it should state for the placeholder: "`{ASSEMBLY NAME}`", not "`{APP NAMESPACE}`".

Probably a good SxS disposal pattern for that is ...

```javascript
var jsObjectReference = DotNet.createJSObjectReference(window);

DotNet.invokeMethodAsync('{APP NAMESPACE}', 'ReceiveWindowObject', jsObjectReference);

DotNet.disposeJSObjectReference(jsObjectReference);
```

... and now that I understand that pattern, ***Javier's remark finally makes sense to me*** 🦖 ***!*** 😆 

I'll put that on the next commit, and we'll see how it composes.

guardrex:
@TanayParikh ... I have something that's probably at least headed in the right direction.

The last commit has four copies of it because there are four versioned copies of the content, so you only need to look at the first one. I'm actively ***getting rid of the horrid&dagger; "Whole-topic Versioning,"*** but I haven't reached this topic yet.

&dagger;*horrid*: That's an IMO. There's a *civil debate* on it among doc authors 🤣. I can't stand broken links in the docs tho, so it has to go from the Blazor docs.

Anyway ... let me know where/how I went sideways. 🙈

btw --- I noted that it doesn't look like disposing of a `JSStreamReference` is a thing in Blazor ... I don't see a `disposeJSStreamReference` anywhere. That was a bit of a surprise.

guardrex:
# 🎗️ *NOTE TO SELF* 🎗️

I need ***one more commit*** before merging. I have that placeholder in two formats:

* `{APP ASSEMBLY}`
* `{ASSEMBLY NAME}`

Check docs everywhere and use ***ONE*** of those everywhere.

guardrex:
Ok ... this looks good now. Let's roll will it! *Thanks everyone!*

### ***BEST WISHES @TanayParikh!*** 🍻 ***Thanks for everything and see you around the Interwebs!***

PR #28103, opened 2023-01-15
Update first-web-api.md

masudme09:
Without exporting the path, it cannot recognize the aspnet-codegenerator.

### DIFF
--- a/aspnetcore/tutorials/first-web-api.md
+++ b/aspnetcore/tutorials/first-web-api.md
@@ -299,6 +299,12 @@ The preceding commands:
 
 Build the project.
 
+On macOS and Linux, export the scaffold tool path:
+
+```console
+export PATH=$HOME/.dotnet/tools:$PATH
+```
+
 Run the following command:
 
   ```dotnetcli


guardrex:
Hello @masudme09 ... This really shouldn't ... ***in theory*** 😄 ... be necessary. The path should be there based on the first run of the SDK installer.

Cross-ref: https://learn.microsoft.com/en-us/dotnet/core/tools/global-tools#install-a-global-tool

I'm not fully aware tho of all of the juicy details for when devs opened issues on the SDK repo for when global tools weren't found and a 💥 occurred, issues such as https://github.com/dotnet/sdk/issues/9415 and https://github.com/dotnet/sdk/issues/23165.

@peterhuene ... Are you able to remark on this, or should William Li perhaps be pinged for assistance?

btw @masudme09 ... Just for future reference, we ask that an issue be opened for anything that isn't a trivial change (e.g., spelling error, small code example error). These types of fundamental changes to the coverage almost always need a little discussion before proceeding to a PR. No worries here tho ... I'm just saying for future reference.

masudme09:
@guardrex I understood that I had made the PR too soon. Anyways, I am using ubuntu 22.04 and I have tried to prepare the environment on vscode. Following: https://learn.microsoft.com/en-us/aspnet/core/tutorials/first-web-api?view=aspnetcore-7.0&tabs=visual-studio-code

And everything was smooth until dotnet-aspnet-codegenerator. After lots of navigation to documents and checking older versions of the documentation - exporting path works. 

And another piece of information- I have installed dotnet 7.0.102

I hope this helps, or I will create a docker container to reproduce the issue.

guardrex:
Let's see if Peter (or William) can advise. They may want to move this over to the SDK repo for further investigation ... i.e., why is this happening and can it be solved in a fundamental way on the SDK installer side? If so, we prefer not to document workarounds in tutorials. We usually point to the SDK installer docs, such as the one that I mentioned. If there's a need to note this type of PATH setting for specific platforms, they'll call it out there ***for everyone***. This isn't the only spot that a workaround callout like this would need to be made around here and in other MS doc sets where it applies. The central location for such remarks would be where the .NET tooling docs are located. Let's see what they think about the next step.

btw wrt Docker ... IIRC, that was a whole additional can of 🪱 worms 😄 ... IIRC, there were separate SDK repo issues pertaining to that because there were other PATH-related problems finding tools. I think they worked on those problems separately, and I'm not aware if/when they solved the problems that were reported.

peterhuene:
Hi @guardrex. I actually left Microsoft back in 2019 and haven't been involved with .NET since then in any capacity.

@marcpopMSFT do you know who best to comment and route this issue to?

guardrex:
Thanks @peterhuene ... I should've checked your profile. I was going off of one of the issues related to this.

marcpopMSFT:
Filing an issue on the SDK repo is the correct action though I'm not sure why the path wouldn't be set during the first run of dotnet. The only reasons would be if there was an environment variable set not to configure the path or because we'd already done it: https://github.com/dotnet/sdk/blob/main/src/Cli/Microsoft.DotNet.Configurer/DotnetFirstTimeUseConfigurer.cs#L106

guardrex:
That's what they'll investigate ... or perhaps they already know the answer and can just remark directly on it without further investigation.

PR #28140, opened 2023-01-19
Document javaScriptInitializer

### LINKED ISSUE #28135: Document javaScriptInitializer for RegisterForJavaScript
Per ...

* https://github.com/dotnet/aspnetcore/issues/37074
* https://github.com/dotnet/aspnetcore/issues/37181
* https://github.com/dotnet/aspnetcore/pull/37598

cc: @SQL-MisterMagoo ... You [asked in 2021 about coverage for this](https://github.com/dotnet/aspnetcore/issues/37181#issuecomment-940567797), but I only just ***today*** saw that request over there while researching updates for the JS initializer content. I'll get to this as soon as I can 🏃😅. 

Current section

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-7.0#render-razor-components-from-javascript

API

https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.components.web.jscomponentconfigurationextensions.registerforjavascript?view=aspnetcore-7.0

---
#### Document Details

* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #28135

[Internal Review Topic (links to section)](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/index?view=aspnetcore-8.0&branch=pr-en-us-28140#render-razor-components-from-javascript)

Mackinnon: @SQL-MisterMagoo [asked in 2021 about coverage for javaScriptInitializer in a PU issue](https://github.com/dotnet/aspnetcore/issues/37181#issuecomment-940567797) that I didn't see until yesterday. I'm going to work on that here ***and*** use the opportunity to knock up a quick example of the whole feature, along the lines of my usual ***fully-working, cut-'n-paste demos*** 👍 that makes it easy for devs to see how features compose.

In addition to just checking the basic technical aspects, I've hit a knowledge gap (not surprising for the 🦖🙈) on part of the coverage. The API states the following:

> **javaScriptInitializer**: Specifies an optional identifier for a JavaScript function that will be called to register the custom element.

Cross-ref: https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.components.web.jscomponentconfigurationextensions.registerforjavascript?view=aspnetcore-7.0

In *my goofy MST quote demo* 😄, I can see my `initializeComponent` FN being called, but I'm not sure what I can do there given that I'm not a JS jockey for non-MS SPA frameworks ...

* Can that do *something* better than just `console.log(...)` that would be germane to the demo?
* What is meant by "to register the custom element" in the param definition? Whatever that means, I think we want to steer clear of "custom elements" language (i.e., "Blazor custom elements") right? ... which is in the very next section of the doc here. I thought it was more like a JS FN that's ***called for initialization/configuration purposes*** ... *guessing* 🤔 of course.
* WRT timing of the JS FN call: It seems to be called before the component is rendered. Is that correct? It seems like the text should say that if that's the case. I'd like to add some kind of remark on the timing of the call.

### DIFF
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -1530,12 +1530,28 @@ rootComponent.dispose();
 
 Blazor Server (`Program.cs`):
 
+Add the namespace for the location of the component. The following example assumes that the `Quote` component is in the app's `Shared` folder, and the app's namespace is `BlazorSample`.
+
+```csharp
+using BlazorSample.Shared;
+```
+
+Register the `Quote` component:
+
 ```csharp
 options.RootComponents.RegisterForJavaScript<Quote>(identifier: "quote", 
     javaScriptInitializer: "initializeComponent");
 ```
 
 Blazor WebAssembly (`Program.cs`):
+
+Add the namespace for the location of the component. The following example assumes that the `Quote` component is in the app's `Shared` folder, and the app's namespace is `BlazorSample`.
+
+```csharp
+using BlazorSample.Shared;
+```
+
+Register the `Quote` component:
 
 ```csharp
 builder.RootComponents.RegisterForJavaScript<Quote>(identifier: "quote", 
@@ -3257,12 +3273,28 @@ rootComponent.dispose();
 
 Blazor Server (`Program.cs`):
 
+Add the namespace for the location of the component. The following example places the `Quote` component in the app's `Shared` folder, and the app's namespace is `BlazorSample`.
+
+```csharp
+using BlazorSample.Shared;
+```
+
+Register the `Quote` component:
+
 ```csharp
 options.RootComponents.RegisterForJavaScript<Quote>(identifier: "quote", 
     javaScriptInitializer: "initializeComponent");
 ```
 
 Blazor WebAssembly (`Program.cs`):
+
+Add the namespace for the location of the component. The following example assumes that the `Quote` component is in the app's `Shared` folder, and the app's namespace is `BlazorSample`.
+
+```csharp
+using BlazorSample.Shared;
+```
+
+Register the `Quote` component:
 
 ```csharp
 builder.RootComponents.RegisterForJavaScript<Quote>(identifier: "quote", 
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -1528,30 +1528,22 @@ rootComponent.dispose();
 }
 ```
 
-Blazor Server (`Program.cs`):
-
-Add the namespace for the location of the component. The following example assumes that the `Quote` component is in the app's `Shared` folder, and the app's namespace is `BlazorSample`.
+In `Program.cs`, add the namespace for the location of the component. The following example assumes that the `Quote` component is in the app's `Shared` folder, and the app's namespace is `BlazorSample`:
 
 ```csharp
 using BlazorSample.Shared;
 ```
 
-Register the `Quote` component:
+In `Program.cs`, register the `Quote` component.
+
+Blazor Server:
 
 ```csharp
 options.RootComponents.RegisterForJavaScript<Quote>(identifier: "quote", 
     javaScriptInitializer: "initializeComponent");
 ```
 
-Blazor WebAssembly (`Program.cs`):
-
-Add the namespace for the location of the component. The following example assumes that the `Quote` component is in the app's `Shared` folder, and the app's namespace is `BlazorSample`.
-
-```csharp
-using BlazorSample.Shared;
-```
-
-Register the `Quote` component:
+Blazor WebAssembly:
 
 ```csharp
 builder.RootComponents.RegisterForJavaScript<Quote>(identifier: "quote", 
@@ -3271,30 +3263,22 @@ rootComponent.dispose();
 }
 ```
 
-Blazor Server (`Program.cs`):
-
-Add the namespace for the location of the component. The following example places the `Quote` component in the app's `Shared` folder, and the app's namespace is `BlazorSample`.
+In `Program.cs`, add the namespace for the location of the component. The following example assumes that the `Quote` component is in the app's `Shared` folder, and the app's namespace is `BlazorSample`:
 
 ```csharp
 using BlazorSample.Shared;
 ```
 
-Register the `Quote` component:
+In `Program.cs`, register the `Quote` component.
+
+Blazor Server:
 
 ```csharp
 options.RootComponents.RegisterForJavaScript<Quote>(identifier: "quote", 
     javaScriptInitializer: "initializeComponent");
 ```
 
-Blazor WebAssembly (`Program.cs`):
-
-Add the namespace for the location of the component. The following example assumes that the `Quote` component is in the app's `Shared` folder, and the app's namespace is `BlazorSample`.
-
-```csharp
-using BlazorSample.Shared;
-```
-
-Register the `Quote` component:
+Blazor WebAssembly:
 
 ```csharp
 builder.RootComponents.RegisterForJavaScript<Quote>(identifier: "quote", 
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -3230,55 +3230,12 @@ For more information, see the following resources:
 
 Razor components can be dynamically-rendered from JavaScript (JS) for existing JS apps.
 
-To render a Razor component from JS, register the component as a root component for JS rendering and assign the component an identifier:
-
-* In a Blazor Server app, modify the call to <xref:Microsoft.Extensions.DependencyInjection.ComponentServiceCollectionExtensions.AddServerSideBlazor%2A> in `Program.cs`:
-
-  ```csharp
-  builder.Services.AddServerSideBlazor(options =>
-  {
-      options.RootComponents.RegisterForJavaScript<Counter>(identifier: "counter");
-  });
-  ```
-  
-  > [!NOTE]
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/components/index.md
```diff
@@ -1509,6 +1509,72 @@ const rootComponent = await window.Blazor.rootComponents.add(...);
...
+  async function showQuote() {
+    let targetElement = document.getElementById('quoteContainer');
+    await Blazor.rootComponents.add(targetElement, 'quote', 
+    {
+        text: "Crow: I have my doubts that this movie is actually 'starring' " +
+        "anybody. More like, 'camera is generally pointed at.'"
+    });
+  }
+
+  async function initializeComponent() {
```
MackinnonBuck, line 1570:
```suggestion
  function initializeComponent() {
```

aspnetcore/blazor/components/index.md
```diff
@@ -1509,6 +1509,72 @@ const rootComponent = await window.Blazor.rootComponents.add(...);
 rootComponent.dispose();
 ```
 
+<xref:Microsoft.AspNetCore.Components.Web.JSComponentConfigurationExtensions.RegisterForJavaScript%2A> includes an overload that accepts a JS function to register the custom element (`javaScriptInitializer`). The following example demonstrates the dynamic registration of a component that renders a quote.
```
MackinnonBuck, line 1512:
```suggestion
<xref:Microsoft.AspNetCore.Components.Web.JSComponentConfigurationExtensions.RegisterForJavaScript%2A> includes an overload that accepts the name of a JS function that performs initialization logic after the component gets registered. This function can be used for integration with JavaScript technologies like HTML custom elements or a JavaScript-based SPA framework. The following example demonstrates the dynamic registration of a component that renders a quote.
```

aspnetcore/blazor/components/index.md
```diff
@@ -1509,6 +1509,72 @@ const rootComponent = await window.Blazor.rootComponents.add(...);
...
+
+> [!IMPORTANT]
+> Don't confuse the `javaScriptInitializer` parameter of <xref:Microsoft.AspNetCore.Components.Web.JSComponentConfigurationExtensions.RegisterForJavaScript%2A> with [JavaScript initializers](xref:blazor/fundamentals/startup#javascript-initializers). The name of the parameter and the JS initializers feature is coincidental.
+
+`Shared/Quote.razor`:
+
+```razor
```
MackinnonBuck, line 1519: More of a high-level question - would it be possible to reuse the example that registers a `Counter` component, and use this section to add a `javaScriptInitializer` to it? Is there a reason we need an entirely new example?

aspnetcore/blazor/components/index.md
MackinnonBuck: Might also be nice to have this function accept arguments and log their values, similar to what we do [here](https://github.com/dotnet/aspnetcore/blob/1dd6f451a0df2ee8e4575dd8dca9df77fda129b1/src/Components/test/testassets/BasicTestApp/wwwroot/js/jsRootComponentInitializers.js#L5).

MackinnonBuck: Also, does this example work? We find the `javaScriptInitializer` function by looking for it on the `window` object. Either way, could we modify this slightly, so we directly attach the function to the `window` object similar to what we do in the code I linked above?

MackinnonBuck (commented):
> I can see my initializeComponent FN being called, but I'm not sure what I can do there given that I'm not a JS jockey for non-MS SPA frameworks ...

The initializer function's primary purpose is to integrate with other technologies. This could involve registering a custom element (as the  `Microsoft.AspNetCore.Components.CustomElements` package does), or it could configure itself with another SPA framework (like React or Angular) so it knows how to render the component via `Blazor.rootComponents.add()`, for examaple.

> WRT timing of the JS FN call: It seems to be called before the component is rendered. Is that correct?

The JS function call is performed just after the Blazor app starts, before any components have been rendered.

> Can that do something better than just console.log(...) that would be germane to the demo?

The initializer function takes the identifier and parameters as arguments. Maybe the example could log those? Here's what our [test code](https://github.com/dotnet/aspnetcore/blob/1dd6f451a0df2ee8e4575dd8dca9df77fda129b1/src/Components/test/testassets/BasicTestApp/wwwroot/js/jsRootComponentInitializers.js#L5) does. I think doing anything more advanced in the function would be unnecssary as a doc example, but it could be worth calling out the potential real-world uses of it.

aspnetcore/blazor/components/index.md
guardrex: > does this example work?

👍 Yes, it does work.

> We find the javaScriptInitializer function by looking for it on the window object. Either way, could we modify this slightly, so we directly attach the function to the window object similar to what we do in the code I linked above?

Sure ... I'll follow your advice on that. I'll get back on this first thing Friday morning.

aspnetcore/blazor/components/index.md
guardrex: Well ... the `Counter` example is kind'a ... er .... ***boring*** 😆. I was hoping to jazz 🎷 this example up a bit.

guardrex:
Thanks for the review @MackinnonBuck ...

> The initializer function's primary purpose is ...

Perfect ... I'll refactor your remarks into doc text.

> The JS function call is performed just after the Blazor app starts, before any components have been rendered.

That's what I thought. I'll call that out in passing.

> The initializer function takes the identifier and parameters as arguments. Maybe the example could log those? Here's what our [test code](https://github.com/dotnet/aspnetcore/blob/1dd6f451a0df2ee8e4575dd8dca9df77fda129b1/src/Components/test/testassets/BasicTestApp/wwwroot/js/jsRootComponentInitializers.js#L5) does. I think doing anything more advanced in the function would be unnecssary as a doc example, but it could be worth calling out the potential real-world uses of it.

Excellent ... I'll work it along those lines.

I'll ping u back Friday ... but there's no rush really.

WRT what the example is ... I think it might be best to redraft the ***whole section*** so that there's ***one*** example ... my cool MST3k example. I really don't prefer the `Counter` component at all in this section. Let me see what I can do with it on Friday morning. *I'll get back to you!*

guardrex:
@MackinnonBuck ... Ok ... this seems to be coming along well.

***Yes! Definitely!*** I'd like to keep the MST3k example and ***NOT*** use the dreaded `Counter` component ***Ugh!*** 😩 I've done this across our docs. The Blazor docs are full of sci-fi TV/movie quotes and examples. IMO, this is more fun than most non-MS doc sets ... and even more fun than *most MS doc sets!* Dan seems ok with it, and I think readers like it. I'm actually *RE-watching* the entire [MST3k series on Netflix](https://www.netflix.com/title/80128275) right now, which is why this came to mind for this example.

Yes, I think ***one example*** for the whole section makes the most sense. I often use this hybrid reference-tutorial format for covrage. This approach, along with fully-working/cut-'n-paste examples 👍, generates *FEWER reader doc issues* ... *concerns* ... *complaints* ... and 💀 ***death threats!*** 💀🤣

A ❓ occurs setting it up: Is the idea for the initialization FN that ...

* There's one of them for all registrations.
* There's usually more than one ... i.e., one for each registration.
* 🤔 ***I think ...*** It depends on the app and the dev's needs ... i.e., there could be one or several.

Check the naming in light of the answer to that question: Right now the naming is `jsComponentInitializers.initializeComponent` in the vein of potentially having ***more than one*** (perhaps one per registration). If that's true, should we go with naming like `jsComponentInitializers.initializeQuoteComponent`, naming it specifically for the `Quote` component? ... and comment in the text something like ...

> Registrations for several root components may use the same initializer function or several initialzier functions dedicated to individual components. Whether there is one or several initializer functions depends on the app's requirements.

guardrex:
# 🦖 NOTE TO SELF 🦖

***Don't forget that there are TWO versioned areas for the updates BEFORE merging!***

MackinnonBuck (approved):
@guardrex This looks great!

> Is the idea for the initialization FN that ... It depends on the app and the dev's needs ... i.e., there could be one or several.

Yes, there could be one or multiple `javaScriptInitializer` functions depending on the app's needs.

> ... should we go with naming like jsComponentInitializers.initializeQuoteComponent ...

I believe the more typical use case will be reusing the same initializer function for multiple components. At least, that's what would be expected if the initializer function is configuring integration with custom elements or another JS SPA framework. For that reason, I think it's fine to leave it as `initializeComponent`. I could go either way regarding whether we should have a `jsComponentInitializers` container object rather than attaching `initializeComponent` directly to the `window` object. Maybe we could remove the contianer object to keep the example simpler, since the 'advanced' example you linked shows an alternative anyway. But I'll leave it up to you, I really don't have a strong preference here.

> For the BasicTestApp as an extra (advanced) example at the bottom of the section...

I think the links you included are good!

guardrex:
Orginal_John_Hannibal_Smith

### _"I love it when a plan comes together!"_

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;John "Hannibal" Smith ([George Peppard](http://www.imdb.com/name/nm0000577/)), [The A-Team](https://www.nbc.com/the-a-team)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;©1983-87 [Universal Television](http://www.nbcuniversal.com/business/universal-television) [Stephen J. Cannell](http://www.cannell.com/)

MackinnonBuck:
@guardrex Yeah, I don't think we need the IFFE, especially since the script doesn't define any global variables👍

Issue #26980, opened 2022-09-13
{ASSEMBLY NAME} appears to be optional when working within the same assembly


contractcooker:
The documentation doesn't make clear that {ASSEMBLY NAME} does not need to be specified when working within the same assembly. Perhaps this is common knowledge but it was not immediately clear to me.

dotNetHelper.invokeMethodAsync("MethodName", data); is how we are using it in our codebase.

I believe the documentation would be clearer if some mention were made that {ASSEMBLY NAME} is optional in certain cases.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 2964d558-cd6d-7421-f7b9-cb939792c2cf
* Version Independent ID: a3517e56-69fc-1ec7-b55a-dbb54d68a807
* Content: [Call .NET methods from JavaScript functions in ASP.NET Core Blazor](https://docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-dotnet-from-javascript?view=aspnetcore-6.0)
* Content Source: [aspnetcore/blazor/javascript-interoperability/call-dotnet-from-javascript.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/javascript-interoperability/call-dotnet-from-javascript.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Yes, @contractcooker ... Good idea. I'll make that update now.

guardrex:
Actually, I think I see that overload now ...

https://github.com/dotnet/aspnetcore/blob/7efa83061a4ceaf347bdf164c255c2c9af707e39/src/JSInterop/Microsoft.JSInterop.JS/src/src/Microsoft.JSInterop.ts#L482-L484

guardrex:
> @param assemblyName The short name (without key/version or .dll extension) of the .NET assembly holding the method to invoke. The value may be null when invoking instance methods.

👉 **The value may be null when invoking instance methods.**

contractcooker:
@guardrex thank you very much for your prompt and thoughtful response. I was not expecting such a quick response and concede that in the grand scheme of things this is a minor issue. I understand completely that this may not be the most pressing issue. Again, thank you for taking the time to look at it.

guardrex:
[...]

PR #28294, opened 2023-02-02
Patch JS interop article

guardrex:
Addresses #26980

Mackinnon ... thanks (*again!*) for the help. I think this is right ... ***or close to right*** 🙈😄.


MackinnonBuck (approved):
🎉

guardrex:
# 🍻

PR #28298, opened 2023-02-03
Clarify swap behvaior of Data Protection key rings

jennings:
This is the existing wording about Data Protection keys and deployment slots:

> Separate deployment slots, such as Staging and Production, don't share a key ring. When you swap between deployment slots, for example swapping Staging to Production or using A/B testing, any app using Data Protection won't be able to decrypt stored data using the key ring inside the previous slot.

After reading this, I'm still not confident how I should interpret it:

- When deployment slots are swapped, key rings are _not_ swapped, so the production slot will continue using the same keys. If you had encrypted data in the staging slot and try to decrypt it with the production slot after the swap, it will fail.
- When deployment slots are swapped, key rings _are_ swapped, so the production slot will now use the keys previously used by the staging slot. If you had encrypted data with the production slot before the swap, the production slot will no longer be able to decrypt it.

I believe the latter is the actual behavior, so this PR helps clarify that.


guardrex:
Hello @jennings ... The changes you're making on the PR aren't correct. Open an issue from the bottom of the topic if you wish to discuss it further. It will add metadata to your opening comment that cross-links the issue for tracking and automatically pings the author of the topic.

Capture

PR #28331, opened 2023-02-08
Switch VS Code launch.json URL to be http instead of https

danroth27:
Blazor WebAssembly apps used to launch by default with an HTTPS address, but users kept hitting issues with that so we switch back to HTTP by default. I copied the url property from the doc and it didn't work because the app was listening on an HTTP address, not HTTPS.


guardrex:
Is this change only for the current release of the framework? There are multiple versions here because I haven't reached this topic yet to remove its Whole-topic Versioning.

guardrex (approved):
I'll go ahead and merge. Let me know if this change applies to earlier versions. If so, I'll patch the doc separately.

PR #28343, opened 2023-02-09
SignalR clarify connectionID access options

### LINKED ISSUE #28081: SignalR - How to access connectionIds outside of Hub when using ISingleClientProxy.InvokeAsync?
Is there a way to access the `connectionId` list outside of a Hub (e.g., from a Controller with an injected IHubContext)?
If I call the `ISingleClientProxy.InvokeAsync` method from e.g., a controller, it naturally requires a connectionId.

Suppose, I'm trying to request data from the client inside a server controller with an injected IHubContext. Do I still need to keep an array of the connectionIds as Context is not accessible from the outside of Hub? (Or so do I believe).
As the `InvokeAsync` is a member of `Client`, I'm kind of forced to provide a `connectionId` outside of the Hub class.
Unfortunately (maybe for good reasons), I can't call InvokeAsync on `.All`.

I just read this [post](https://stackoverflow.com/a/50394456/4393935) that it's not possible to have access to the connectionIds from the outside of Hub. 
However, in the case of invoking a client method would require the IHubContext to have an immediate access to the connectionIds. Otherwise, I'm not entirely sure how it would work.

In [this doc](https://learn.microsoft.com/en-us/aspnet/core/signalr/hubcontext?view=aspnetcore-7.0#get-an-instance-of-ihubcontext-in-middleware), it states that it's not possible.

> When client methods are called from outside of the Hub class, there's no caller associated with the invocation. Therefore, there's no access to the ConnectionId, Caller, and Others properties.

In that case, I'm baffled as to how I could use InvokeAsync function when having no access to connectionIds.

The [documentation](https://learn.microsoft.com/sv-se/aspnet/core/signalr/hubs?view=aspnetcore-7.0#client-results) provides an example of InvokeAsync usage but nothing about how the `connectionID` is retrieved.

I'm guessing it must be from a class where it has direct access to Hub thus its connections... A comprehensive example would really be helpful.

```cs
async Task SomeMethod(IHubContext<MyHub> context)
{
    string result = await context.Clients.Client(connectionID).InvokeAsync<string>(
        "GetMessage");
}
```

---
[Associated WorkItem - 60042](https://dev.azure.com/msft-skilling/Content/_workitems/edit/60042)

wadepickett:
[Internal review](https://review.learn.microsoft.com/en-us/aspnet/core/signalr/hubcontext?view=aspnetcore-8.0&branch=pr-en-us-28343#get-an-instance-of-ihubcontext-in-middleware)

Fixes #28081

### DIFF
--- a/aspnetcore/signalr/hubcontext.md
+++ b/aspnetcore/signalr/hubcontext.md
@@ -57,9 +57,9 @@ app.Use(async (context, next) =>
 >
 > If your application needs to map a user to the connection ID and persist that mapping, you can do one of the following:
 >
-> Persist mapping of single or multiple connections as groups. See [Groups in SignalR](xref:signalr/groups#groups-in-signalr) for more information.
-> Store the connection ID in a singleton service. See [Inject services into a hub](xref:signalr/hubs#inject-services-into-a-hub) for more information.
-> Pass the connection ID between clients.
+> - Persist mapping of single or multiple connections as groups. See [Groups in SignalR](xref:signalr/groups#groups-in-signalr) for more information.
+> - Store the connection ID in a singleton service. See [Inject services into a hub](xref:signalr/hubs#inject-services-into-a-hub) for more information.
+> - Pass the connection ID between clients.
 
 ### Get an instance of `IHubContext` from IHost
 
--- a/aspnetcore/signalr/hubcontext.md
+++ b/aspnetcore/signalr/hubcontext.md
@@ -55,7 +55,7 @@ app.Use(async (context, next) =>
 > [!NOTE]
 > When client methods are called from outside of the `Hub` class, there's no caller associated with the invocation. Therefore, there's no access to the `ConnectionId`, `Caller`, and `Others` properties.
 >
-> If your application needs to map a user to the connection ID and persist that mapping, you can do one of the following:
+> Apps that need to map a user to the connection ID and persist that mapping can do one of the following:
 >
 > - Persist mapping of single or multiple connections as groups. See [Groups in SignalR](xref:signalr/groups#groups-in-signalr) for more information.
 > - Retain connection and user information through a singleton service. See [Inject services into a hub](xref:signalr/hubs#inject-services-into-a-hub) for more information. The singleton service can use any storage method, such as:
--- a/aspnetcore/signalr/hubcontext.md
+++ b/aspnetcore/signalr/hubcontext.md
@@ -58,7 +58,8 @@ app.Use(async (context, next) =>
 > If your application needs to map a user to the connection ID and persist that mapping, you can do one of the following:
 >
 > - Persist mapping of single or multiple connections as groups. See [Groups in SignalR](xref:signalr/groups#groups-in-signalr) for more information.
-> - Store the connection ID in a singleton service. See [Inject services into a hub](xref:signalr/hubs#inject-services-into-a-hub) for more information.
+> - Retain connection and user information in a singleton service. See [Inject services into a hub](xref:signalr/hubs#inject-services-into-a-hub) for more information.
+> - Retain connection and user information in a dictionary that is stored in memory.
 > - Pass the connection ID between clients.
 
 ### Get an instance of `IHubContext` from IHost
--- a/aspnetcore/signalr/hubcontext.md
+++ b/aspnetcore/signalr/hubcontext.md
@@ -58,11 +58,10 @@ app.Use(async (context, next) =>
 > If your application needs to map a user to the connection ID and persist that mapping, you can do one of the following:
 >
 > - Persist mapping of single or multiple connections as groups. See [Groups in SignalR](xref:signalr/groups#groups-in-signalr) for more information.
-> - Retain connection and user information in a singleton service. See [Inject services into a hub](xref:signalr/hubs#inject-services-into-a-hub) for more information.
-> - Pass the connection ID between clients.
-> - Retain connection and user information using any storage method, such as:
+> - Retain connection and user information through a singleton service. See [Inject services into a hub](xref:signalr/hubs#inject-services-into-a-hub). The singleton service can use any storage method, such as:
 >   - In-memory storage in a dictionary.
 >   - Permanent external storage.  For example, a database or Azure Table storage using the [Azure.Data.Tables NuGet package](https://www.nuget.org/packages/Azure.Data.Tables/).
+> - Pass the connection ID between clients.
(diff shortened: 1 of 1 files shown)


aspnetcore/signalr/hubcontext.md
```diff
@@ -54,6 +54,15 @@ app.Use(async (context, next) =>
...
 > When client methods are called from outside of the `Hub` class, there's no caller associated with the invocation. Therefore, there's no access to the `ConnectionId`, `Caller`, and `Others` properties.
+>
+> If your application needs to map a user to the connection id and persist that mapping, you can use one of the following:
+>
+> - [The User ID Provider](#IUserIdProvider)
+> - [In-memory storage](#inmemory), such as a dictionary
+> - [SignalR group for each user](#groups)
+> - [Permanent, external storage](#database), such as a database table or Azure table storage
+>
+> For more information see, [Mapping SignalR Users to Connections](xref:signalr/overview/guide-to-the-api/mapping-users-to-connections).
```
BrennanConroy, line 65: Wrong docs. That's for ASP.NET SignalR

wadepickett: Thanks, pulling back this draft.

wadepickett:
@BrennanConroy, I am a bit new to this area and I am not confident I have captured what you intended correctly.  Does this seem correct?

BrennanConroy:
I wonder if we should copy a lot of https://learn.microsoft.com/aspnet/signalr/overview/guide-to-the-api/mapping-users-to-connections#in-memory-storage so it's easier for users to get started on their own connection mapping.

wadepickett:
@BrennanConroy, I agree!  We don't have any of that available for the Core version in docs and it is needed.    I need to update the sample it is using.  If it is OK with you, I will publish what we have here so far in this PR now if you approve, and I will create a new issue for documenting the In-memory storage option for the .NET 7/8 that includes an updated sample which I will take care of (and dedicate more time for).

wadepickett:
@BrennanConroy:  Correction:  I guess it would be better to pull it in on this PR.  I don't want to give the impression that In-memory storage is not an option by not listing it.    I'll add an new section to this doc:  In-memory storage and point to it from the note.

aspnetcore/signalr/hubcontext.md
```diff
@@ -54,6 +54,13 @@ app.Use(async (context, next) =>
 
 > [!NOTE]
 > When client methods are called from outside of the `Hub` class, there's no caller associated with the invocation. Therefore, there's no access to the `ConnectionId`, `Caller`, and `Others` properties.
+>
+> If your application needs to map a user to the connection ID and persist that mapping, you can do one of the following:
+>
+> - Persist mapping of single or multiple connections as groups. See [Groups in SignalR](xref:signalr/groups#groups-in-signalr) for more information.
+> - Retain connection and user information in a singleton service. See [Inject services into a hub](xref:signalr/hubs#inject-services-into-a-hub) for more information.
+> - Retain connection and user information in a dictionary that is stored in memory.
```
BrennanConroy, line 62: It's not just in-memory, you can store the info in a database or any other storage method.

BrennanConroy (approved):
> It seems like the proper long-term solution is for me to create a modern Core version of the dedicated 2014 topic

Yeah, that seems like the right direction.

wadepickett:
@BrennanConroy, thanks, I updated this PR following your suggestion and created a tracking issue for a rewrite/update of 2014 topic [Mapping SignalR Users to Connections](https://learn.microsoft.com/en-us/aspnet/signalr/overview/guide-to-the-api/mapping-users-to-connections) which I can point to from this topic once published.

aspnetcore/signalr/hubcontext.md
```diff
@@ -54,6 +54,15 @@ app.Use(async (context, next) =>
...
 > [!NOTE]
 > When client methods are called from outside of the `Hub` class, there's no caller associated with the invocation. Therefore, there's no access to the `ConnectionId`, `Caller`, and `Others` properties.
+>
+> If your application needs to map a user to the connection ID and persist that mapping, you can do one of the following:
+>
+> - Persist mapping of single or multiple connections as groups. See [Groups in SignalR](xref:signalr/groups#groups-in-signalr) for more information.
+> - Retain connection and user information in a singleton service. See [Inject services into a hub](xref:signalr/hubs#inject-services-into-a-hub) for more information.
+> - Pass the connection ID between clients.
+> - Retain connection and user information using any storage method, such as:
+>   - In-memory storage in a dictionary.
```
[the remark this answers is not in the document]
wadepickett: I have moved the storing options to under the singleton service. I wasn't catching on that the storage options were referring only to the singleton service.  Thanks!

BrennanConroy (approved):

wadepickett:
Once I have the following topic published, we will point to that full guidance instead: #28410.  Thanks Brennan for all the help.

aspnetcore/signalr/hubcontext.md
```diff
@@ -54,6 +54,14 @@ app.Use(async (context, next) =>
 
 > [!NOTE]
 > When client methods are called from outside of the `Hub` class, there's no caller associated with the invocation. Therefore, there's no access to the `ConnectionId`, `Caller`, and `Others` properties.
+>
+> If your application needs to map a user to the connection ID and persist that mapping, you can do one of the following:
```
Rick-Anderson, line 58:
```suggestion
> Apps that need to map a user to the connection ID and persist that mapping, can do one of the following:
```

Rick-Anderson (changes requested):

aspnetcore/signalr/hubcontext.md
Rick-Anderson: [Use bigger ideas, fewer words](https://learn.microsoft.com/en-us/style-guide/top-10-tips-style-voice#use-bigger-ideas-fewer-words) : Our modern design hinges on crisp minimalism. Shorter is always better.
[Short sentences and fragments are easier to scan and read.](https://learn.microsoft.com/en-us/style-guide/brand-voice-above-all-simple-human#a-focus-on-the-customer) Prune every excess word.

What about consultants who are working on an app? This repo uses a more formal style.

wadepickett: I added a note to also fix those in the original published doc that text snippet was derived from, which I will have a chance to rewrite.  Thanks.

Rick-Anderson (approved):

PR #28399, opened 2023-02-15
Update Index.cshtml

SKarwasra:
Use the parameter @item instead of item in AuthorizationService.AuthorizeAsync(...) method
Otherwise it fails.

### DIFF
--- a/aspnetcore/security/authorization/secure-data/samples/final6/Pages/Contacts/Index.cshtml
+++ b/aspnetcore/security/authorization/secure-data/samples/final6/Pages/Contacts/Index.cshtml
@@ -63,7 +63,7 @@
                 </td>
                 <td>
                     @if ((await AuthorizationService.AuthorizeAsync(
-                     User, item,
+                     User, @item,
                      ContactOperations.Update)).Succeeded)
                     {
                         <a asp-page="./Edit" asp-route-id="@item.ContactId">Edit</a>
@@ -73,7 +73,7 @@
                     <a asp-page="./Details" asp-route-id="@item.ContactId">Details</a>
 
                     @if ((await AuthorizationService.AuthorizeAsync(
-                     User, item,
+                     User, @item,
                      ContactOperations.Delete)).Succeeded)
                     {
                         <text> | </text>


Rick-Anderson:
@Rabadash8820  @Neutrino-Sunset @JakubJanowski @Bartmax please review. It works without `@` and works with it.

It should have `@`, but why does it work without it?

See [Index.cshtml](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/security/authorization/secure-data/samples/final6/Pages/Contacts/Index.cshtml#L66) page.

Rick-Anderson:
Does `@if (` make it an [Explicit Razor expression](https://learn.microsoft.com/en-us/aspnet/core/mvc/views/razor?view=aspnetcore-7.0#explicit-razor-expressions)?

Rabadash8820:
@Rick-Anderson That would be my guess. The docs you linked don't say what happens when the _expression_ part of an `@if` statement spans multiple lines, but it makes total sense to me that the parser would extend the Razor expression across lines until it finds closing parentheses.

Neutrino-Sunset:
I don't believe this PR is correct.

Imo the variable `item` does not require a preceding `@` in the context of the `if` statement because `item` is declared in the enclosing `@foreach (var item in Model.Contact)`

Therefore `item` in an enclosed scope simply refers to that variable.

Using `@item` also works for the same reason that the following is valid

```
         for (int bob = 0; @bob <= 5; ++@bob)
         {
            int a = 1;
         }
```

The leading `@` is normally used to permit the use of keywords as variable names, but the key point from the documentation is that the `@` does not form part of the variable name. Therefore if you use it on a variable name that is not a keyword then it makes no difference whether it is present or not.

https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names

Rabadash8820:
So really I guess none of the `@`s inside `@foreach` are necessary, including the one before `if`. 🤔

Neutrino-Sunset:
The ones before the `if` statements are necessary, because they are what switches the parsing from HTML markup to C#, but once in the C# code variables would not require an `@` prefix unless they mirror keywords.

Neutrino-Sunset:
If `AuthorizationService.AuthorizeAsync(User, item` really doesn't work without using `User, @item` that would seem to me to be a parsing bug.

JakubJanowski:
I don't know why I was summoned here but I agree that the `@` is redundant here.

Rick-Anderson:
> If `AuthorizationService.AuthorizeAsync(User, item` really doesn't work without using `User, @item` that would seem to me to be a parsing bug.

It works either way, I've tested it. I'm going to update the [Explicit Razor expression](https://learn.microsoft.com/en-us/aspnet/core/mvc/views/razor?view=aspnetcore-7.0#explicit-razor-expressions) section to show this code.

PR #28425, opened 2023-02-16
Update anti-request-forgery.md

### LINKED ISSUE #13024: Misleading Implementation Recommendation
The documentation I am referencing is this one regarding CSRF Attack Prevention

https://docs.microsoft.com/en-us/aspnet/core/security/anti-request-forgery
I would recommend adding a note under the section quoted: 

*To send the token on subsequent requests, store the token in the browser's local storage. Don't be concerned about CSRF vulnerability if the token is stored in the browser's local storage. CSRF is a concern when the token is stored in a cookie.*

Although this implementation would in-fact prevent a CSRF attack, by using this kind of implementation you are immediately putting the reader's web application at risk for a XSS attack.  By placing an authentication token into local storage, you are putting the web application at risk as an attacker that has gained access to the token by injecting their own script onto the page (whether directly via HTML inputs or a indirectly via a 3rd party library).

My only suggestion would to be putting a disclaimer after this line to warn readers about such vulnerabilities.

Thanks,
Matt

---
#### Document Details

* ID: 47b14f57-82ac-a2e2-cbc7-22a81a60f4ac
* Version Independent ID: bffca13c-223f-c61f-9cb2-9da8811eecfa
* Content: [Prevent Cross-Site Request Forgery (XSRF/CSRF) attacks in ASP.NET Core](https://learn.microsoft.com/en-us/aspnet/core/security/anti-request-forgery?view=aspnetcore-7.0)
* Content Source: [aspnetcore/security/anti-request-forgery.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/security/anti-request-forgery.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

---
[Associated WorkItem - 64418](https://dev.azure.com/msft-skilling/Content/_workitems/edit/64418)

Rick-Anderson:
Fixes #13024


Rick-Anderson:
@mhintzke please review

skykick-cm-2 (changes requested):
This looks great! One typo: `eternal` I think is meant to be `external`. Other than that looks good.

mhintzke:
This looks great! One typo: `eternal I think is meant to be `external`. Other than that looks good.

aspnetcore/security/anti-request-forgery.md
```diff
@@ -69,7 +69,9 @@ When a user authenticates using their username and password, they're issued a to
...
.../security/claims-based-identity-model) or a reference token that points the app to user state maintained in the app. When a user attempts to access a resource that requires authentication, the token is sent to the app with an extra authorization header in the form of a Bearer token. This approach makes the app stateless. In each subsequent request, the token is passed in the request for server-side validation. This token isn't *encrypted*; it's *encoded*. On the server, the token is decoded to access its information. To send the token on subsequent requests, store the token in the browser's local storage. Placing a token in the browser local storage and retrieving it and using it as a bearer token provides protection against CSRF attacks. However, should the app be vulnerable to script injection via XSS or a compromised eternal javascript file, an attacker could retrieve any value from local storage and send it to themselves. ASP.NET Core encodes all server side output from variables by default, reducing the risk of XSS. If you override this behavior by using [Html.Raw](xref:System.Web.Mvc.HtmlHelper.Raw%2A) or custom code with untrusted input then you may increase the risk of XSS.
```
Rick-Anderson, line 72:
```suggestion
When a user is authenticated, they're issued a token (not an antiforgery token). The token contains user information in the form of [claims](/dotnet/framework/security/claims-based-identity-model) or a reference token that points the app to user state maintained in the app. When a user attempts to access a resource that requires authentication, the token is sent to the app with an extra authorization header in the form of a Bearer token. This approach makes the app stateless. In each subsequent request, the token is passed in the request for server-side validation. This token isn't *encrypted*; it's *encoded*. On the server, the token is decoded to access its information. To send the token on subsequent requests, store the token in the browser's local storage. Placing a token in the browser local storage and retrieving it and using it as a bearer token provides protection against CSRF attacks. However, should the app be vulnerable to script injection via XSS or a compromised external javascript file, an attacker could retrieve any value from local storage and send it to themselves. ASP.NET Core encodes all server side output from variables by default, reducing the risk of XSS. If you override this behavior by using [Html.Raw](xref:System.Web.Mvc.HtmlHelper.Raw%2A) or custom code with untrusted input then you may increase the risk of XSS.
```

PR #28492, opened 2023-02-24
Additional JS initializer scenarios

### LINKED ISSUE #26363: Can we add some concrete examples for JavaScript initializers in this page?
[EDIT by guardrex to change the metadata to point to the *Startup* topic, where we can place the examples with existing ones.]

Examples are:
* Ensuring that your library is loaded before Blazor Starts.
* Automatically loading the scripts that your library needs.
* Ensuring that Blazor has been initialized before calling into it from your library.

---
#### Document Details

* ID: ad891f84-d11e-e334-fcdc-b55c5171fca4
* Version Independent ID: 11f8db3d-9988-e0a1-89c5-c3552ee47b93
* Content: [ASP.NET Core Blazor startup](https://docs.microsoft.com/en-us/aspnet/core/blazor/fundamentals/startup?view=aspnetcore-6.0)
* Content Source: [aspnetcore/blazor/fundamentals/startup.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/fundamentals/startup.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #26363

Since the issue was opened (July, '22), PRs have at least partially addressed the ask. In the first version of JS initializer coverage, our examples merely called `console.log`. These days, we show how to append custom scripts to the head to load them.

Today, I'm trying to figure out what's missing. What's on this PR might not address all of the asks. What's on here seems to me to be the obvious scenarios.

Here's the original list from the issue ...

* Ensuring that your library is loaded before Blazor Starts.
* Automatically loading the scripts that your library needs.
* Ensuring that Blazor has been initialized before calling into it from your library.

Here are the sections that I thought to add today on top of prior PRs that have improved the coverage ...

* Ensure libraries are loaded in a specific order
* Import additional modules
* Export additional JavaScript in the JavaScript initializers file
* Import map
* Module extension (`.js`/`.mjs`)


aspnetcore/blazor/fundamentals/startup.md
```diff
@@ -72,6 +90,94 @@ For an examples of JS initializers, see the following resources:
...
+import { logMessage } from "/additionalModule.js";
+
+export function beforeStart(options, extensions) {
+  ...
+
+  logMessage();
+}
+```
+
+For browser compatibility, see [Can I use: JavaScript statement: import](https://caniuse.com/?search=JavaScript%20statement%3A%20import).
```
MackinnonBuck, line 143: Do we link to https://caniuse.com in other places in the docs? This information can also be found on MDN: https://developer.mozilla.org/docs/web/javascript/reference/statements/import#browser_compatibility

MackinnonBuck: Actually, I don't think it's necessary to leave the browser compatibility note. Browser versions that support dynamic imports already support the `import` statement. Since Blazor itself relies on dynamic imports to run the JS initializers, we can be certain that they're always supported in Blazor apps.

aspnetcore/blazor/fundamentals/startup.md
```diff
@@ -72,6 +90,94 @@ For an examples of JS initializers, see the following resources:
...
+
+[Dynamic import with the `import()` operator](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Operators/import) is supported with ASP.NET Core and Blazor:
+
+```javascript
+import("/additionalModule.js");
```
MackinnonBuck, line 148: I'm not sure we need an example showing dynamic imports. I can't really think of a real case where you'd want to use dynamic imports over import statements in the context of a JS initializer. Blazor dynamically imports your JS initializer just before calling its `beforeStart()` function, so using an import statement and doing a dynamic import in `beforeStart()` effectively do the same thing.

aspnetcore/blazor/fundamentals/startup.md
```diff
@@ -72,6 +90,94 @@ For an examples of JS initializers, see the following resources:
...
+}
+
+export function afterStarted(blazor) {
+  ...
+}
+
+export function customFunction() {
+  ...
+}
+```
```
MackinnonBuck, line 169: Hm, I'm trying to think of scenarios where this would be useful, but I can't think of any. The only consumer of the JS initializer is usually Blazor itself, which won't do anything with these exports. Did you have a scenario in mind when writing this?

aspnetcore/blazor/fundamentals/startup.md
```diff
@@ -72,6 +90,94 @@ For an examples of JS initializers, see the following resources:
...
+
+### Import map
+
+[Import maps](https://developer.mozilla.org/docs/Web/HTML/Element/script/type/importmap) are supported by ASP.NET Core and Blazor.
+
+For browser compatibility, see [Can I use: import maps](https://caniuse.com/import-maps).
+
+### Module extension (`.js`/`.mjs`)
+
+Both `.js` and `.mjs` module extensions are supported by ASP.NET Core and Blazor, which are sent with a MIME type of `text/javascript`.
```
MackinnonBuck, line 179: Being in the "JavaScript initializers" section of the document, I wonder if this could be misinterpreted as indicating that the .mjs module extension can be used for JS initializers? i.e., naming the initializer module `<my_library_name>.lib.module.mjs`. I'm pretty sure that won't work.

MackinnonBuck, line 179: Have customers shown any confusion about these features? I'm wondering if it's even worth pointing out that they work, since I think they should automatically be expected to work in any web app.

MackinnonBuck (commented):
For the scenario:

> Ensuring that Blazor has been initialized before calling into it from your library.

I don't think we currently cover this in the JS initializers section of the document. Could we have an example where we interact with one of the `Blazor` JS APIs in an `afterStarted()` callback? e.g., register a custom event, add a root component.

aspnetcore/blazor/fundamentals/startup.md
guardrex: I'll strike it from this section/article.

Is the dynamic import useful for loading modules based on a conditional check? If so, we can keep it with the rest of the 'where to place JS in a Blazor app (modules)' content (which isn't in the this article). If yes, I'll add it to the right spot. If not, I'll strike it here, and we won't keep any coverage on it.

aspnetcore/blazor/fundamentals/startup.md
guardrex: Indeed! I didn't mean to imply that. Yes, this should only appear in the 'where to keep your JS (modules)' content, not in this doc.

I'll strike this here.

guardrex: No, you're correct. Let's kill 💀 off any content that isn't helpful.

guardrex:
Thanks, @MackinnonBuck ...

* For custom events ... Yes, registration examples are only in a `<script>` in that coverage. The JS initializers approach isn't shown. I can add an example to that article's coverage, and then cross-link to it both ways.

* For a root component ... Same thing. We only have `Blazor.rootComponents.add` in an external scripts file. I'll draft an example in that coverage and cross-link both ways.

WRT the updates for this PR, let me get back to you later in the week. I'm working Blazor security topics today, and I'll be on the road moving to another city on Wednesday 🚚. I'll be back on Thursday/Friday to update this PR.

aspnetcore/blazor/fundamentals/startup.md
MackinnonBuck: > Is the dynamic import useful for loading modules based on a conditional check?

Yep, that does seem reasonable! I suppose it would be useful to mention 🙂

aspnetcore/blazor/fundamentals/startup.md
MackinnonBuck: I wonder if, rather than getting into specifics, it would be sufficient to say something like:

> JavaScript module loading works the same way in Blazor as it does for other types of web apps, and you're free to customize how modules are defined in your app. See https://developer.mozilla.org/docs/Web/JavaScript/Guide/Modules for a guide on how to use JavaScript modules.

Or, do you think it would be worth it to include our own guidance? We could also do both - briefly summarize some ways to define/use JS modules and then link to the MDN doc for more info.

guardrex: Well, so much for getting back to this Thursday/Friday. Moving across Florida has been a crazy whirlwind of activity 😫. I'm good now tho ... but I am ***living in a SEA of BOXES!*** 😆 

WRT the text/approach ... yes ... that's cool. I'll use that in the JS interop modules INCLUDE that I've sprinkled in where appropriate. I'll remove this `.mjs` bit.

guardrex: This is removed everywhere now on the next commit.

guardrex:
@MackinnonBuck ... So much for Thursday/Friday! 😄 ... The move across the state of Florida has been a ***BEAST*** 👹. Things are stabilizing here now.

WRT the updates thus far, I addressed the feedback points. I'm working the last two items, custom events and root components. When I started with custom events, I noted that our coverage is lacking. I couldn't get the base case described by the doc to work. I've taken Steve's advice at https://github.com/dotnet/aspnetcore/issues/37012#issuecomment-940907089, and that did help ... it did light-up the custom event attribute. However, I'm not quite there on full implementation. I'll work on it further tomorrow, Sunday. If I can't work it out soon, I'll put up a GH repro for you to look at and see what I'm doing wrong ... then quickly move to the root component item to see if I can at least get that one done.

guardrex:
@MackinnonBuck ... Ok ... there's at least enough there for you to ***punch me*** 👦🤛😵 on ...

The custom events example for 6.0/7.0 now doesn't use the `<script>` approach. It goes with a JS initializer, and I cross-linked it from the JS initializer section of the *Startup* doc.

The root component example keeps it's current `showQuote()` call to dynamically render the root component. I've ***added an example*** to the end where the root component is rendered via a JS initializer, which is also cross-linked from the *Startup* doc.

While updating the custom event section, I noticed that our current guidance just didn't work OOB following the steps. It really did need Steve's remarks from https://github.com/dotnet/aspnetcore/issues/37012#issuecomment-940907089 rolled into it, so I have that here now. Also ... a *NIT* ... it needed to have the `Microsoft.AspNetCore.Components` namespace to light-up the `EventHandlerAttribute` ... all good now 👍.

I'm more concerned about the root components part, but l'll see what you think.

It might be quicker/easier to look at the review topics. These link to the relevant sections ...

* [Event Handling: Custom event arguments section](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/event-handling?view=aspnetcore-8.0&branch=pr-en-us-28492#custom-event-arguments)
* [Razor components in JS apps: Render Razor components from JavaScript section (it's at the end)](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/js-spa-frameworks?view=aspnetcore-8.0&branch=pr-en-us-28492#render-razor-components-from-javascript)
* [Startup: JavaScript initializers section](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/startup?view=aspnetcore-8.0&branch=pr-en-us-28492#javascript-initializers)

aspnetcore/blazor/components/js-spa-frameworks.md
```diff
@@ -139,6 +139,34 @@ Quote &copy;1988-1999 Satellite of Love LLC: [*Mystery Science Theater 3000*](ht
...
+
+The following example builds on the preceding example, using the `Quote` component, the root component registration in `Program.cs`, and the initialization of `jsComponentInitializers.js`. The `showQuote()` function (and the `script.js` file) aren't used.
+
+In HTML, place the target container element, `quoteContainer2` for this example:
+
+```html
+<div id="quoteContainer2"></div>
+```
+
+In a [JavaScript initializer](xref:blazor/fundamentals/startup#javascript-initializers). The following example only handles pasting text, but you could use arbitrary JavaScript APIs to deal with users pasting other types of data, such as images.
```
MackinnonBuck, line 152: > In a JavaScript initializer.

I'm guessing something got left out of this sentence. Or am I misinterpreting it?

> The following example only handles pasting text, but you could use arbitrary JavaScript APIs to deal with users pasting other types of data, such as images.

Forgive me if I'm missing something, but how does the example relate to pasting?

aspnetcore/blazor/fundamentals/startup.md
```diff
@@ -62,16 +78,76 @@ export function afterStarted(blazor) {
...
+  logMessage();
+}
+```
+
+### Import map
+
+[Import maps](https://developer.mozilla.org/docs/Web/HTML/Element/script/type/importmap) are supported by ASP.NET Core and Blazor.
+
+For browser compatibility, see [Can I use: import maps](https://caniuse.com/import-maps).
+
```
MackinnonBuck, line 150: Were we going to remove this, too? This might be covered implicitly by the other statement added in this PR:

> JavaScript module loading works the same way in Blazor as it does for other types of web apps, and you're free to customize how modules are defined in your app. For a guide on how to use JavaScript modules, see [MDN Web Docs: JavaScript modules](https://developer.mozilla.org/docs/Web/JavaScript/Guide/Modules).

MackinnonBuck (approved):
Looks good to me! Just a couple remaining questions/suggestions, but I think this should be good to merge after those are addressed 🙂

aspnetcore/blazor/components/js-spa-frameworks.md
guardrex: Yes, good 👁️ ... This paragraph must be a stray bad paste 😈 while I was working on the text. The "paste" part refers to the custom paste example in the *Event Handling* article. This paragraph shouldn't be here. I'll 🔪 it out on the next commit.

aspnetcore/blazor/fundamentals/startup.md
guardrex: Yes, I'll 🔪 this as well.

guardrex:
Thanks, @MackinnonBuck! 🎸 I'll make those edits and merge this.

aspnetcore/blazor/components/js-spa-frameworks.md
guardrex: Actually ... I don't think it's a stray paste ... I think it's just unfinished business. It should say ...

> Using a \[JavaScript initializer](xref:blazor/fundamentals/startup#javascript-initializers), add the root component to the target container element.

I'll go with that.

PR #28535, opened 2023-03-01
Add doc for using Blazor with YARP

twsouthwick:


twsouthwick:
@Tratcher As I wrote this, it seems like a general purpose guidance for YARP users. Should this be with the YARP docs and linked from the incremental?

pundj:
@Tratcher @twsouthwick This doc is for Blazor Server, but doesn't work for Blazor WebAssembly. Is there documentation on how to resolve the same issue when you are using YARP as a fallback in the same app as Blazor WebAssembly?

gerneio:
@pundj I could be completely way off, and haven't tested this myself (outside of Blazor server), but after looking at [this doc](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-3.1#routing-with-urls-that-contain-dots:~:text=endpoints.MapFallbackToFile(%22/example/%7Bparam%3F%7D%22%2C%20%22index.html%22)%3B), I think it might be as simple as:

`app.MapBlazorPages("index.html");`

dino182:
I have tried this out for a Blazor WebAssembly project. I can confirm that it won't just work by tweaking the page name. A couple of other changes to the extension methods are also required:

1.
```cs
// Replace this
var assembly = Assembly.GetEntryAssembly();

// with this to load the Client assembly instead of the Server assembly
var assembly = Assembly.GetAssembly(typeof(Client.App));
```

2.
```cs
// Replace this
var conventionBuilder = endpoints.MapFallbackToPage(route, page);

// with this so that WebAssembly routing is used
var conventionBuilder = endpoints.MapFallbackToFile(route, page);
```

3. As @gerneio stated
```cs
// Change the call to
app.MapBlazorPages("/_Host");

// to this
app.MapBlazorPages("index.html");
```

@twsouthwick @Tratcher could you please add this WebAssembly adjustment to the documentation?

twsouthwick:
@javiercn Looks like the recommendation I've added only works for Blazor server. @dino182 got it to work by switching out the `MapFallbackToPage` to `MapFallbackToFile`. Is this a case where we need different recommendations for webassembly/server or is there a pattern we could use that would work for both?

javiercn:
@twsouthwick the problem is the `assembly.GetEntryAssembly()`.

We added a `MapRazorComponents<TAppComponent>()` in preview 3 that does this, but uses typeof(TAppComponent).Assembly as the source, which allows you to reference a webassembly app and do endpoints.MapRazorComponents<Client.App>()` which should work.

Note that this is for new "united" apps and is likely not going to work with an existing app, since the rendering process skips Razor Pages. However, it is expected that we can have a method that chains from `MapRazorComponents` like `ToRazorPage("/_Host") or `ToFile("index.html")` that replaces the endpoint delegate with one that serves the file/page.

twsouthwick:
What about downlevel (ie 6/7)?

javiercn:
@twsouthwick can you clarify what you mean?

PR #28577, opened 2023-03-04
Fixing typo in links to previous/next pages

snikolaj:
Removed a ```(``` that broke the links to the previous/next pages and added a colon so that the links are consistent with other pages.


snikolaj:
@dotnet-policy-service agree

PR #28580, opened 2023-03-05
Fixing paths for apache error and access logs

selvinfehric:
Paths were saved to:
/var/log/apache2helloapp-error.log
/var/log/apache2helloapp-access.log

Instead, it should be:
/var/log/apache2/helloapp-error.log
/var/log/apache2/helloapp-access.log


selvinfehric:
@dotnet-policy-service agree

wadepickett:
@selvinfehric, thanks for the contribution!  You are correct, it should have followed: ErrorLog ${APACHE_LOG_DIR}/example.log.

wadepickett (approved):
Verified the original path was incorrect and this new one is correct. I happened to be already be editing this topic in another PR and a VM running to test for an update.

PR #28593, opened 2023-03-06
Fix typo ("or" to "and" in "both x or y")

tommysor:


guardrex (approved):
Great catch ⚾, @tommysor! *Thanks!*

PR #28696, opened 2023-03-16
Component parameter value guidance updates

### LINKED ISSUE #28694: Clarify syntax for specifying component parameter values
I find the language about how to specify component parameters a bit confusing. The section on component parameters currently says: "Assign a C# field, property, or result of a method to a component parameter as an HTML attribute value using [Razor's reserved @ symbol](https://learn.microsoft.com/en-us/aspnet/core/mvc/views/razor?view=aspnetcore-7.0#razor-syntax)." But `@` isn't actually needed. Component parameter attributes are parsed as a C# context (unless the parameter is of type `string`) so a transition isn't typically necessary. Stylistically some people prefer to specify a leading `@` to clarify that the context is C# but doing so is generally optional. The currently language makes it sound like the leading `@` is required when it isn't.

Later the doc says: "The @ prefix is required for string parameters. Otherwise, the framework assumes that a string literal is set." This phrasing seems to contradict itself. Is the leading @ required or not? The answer is it isn't required, but the value of string parameters are treated as string literals by default. If you want a C# context instead you need to lead with an `@`.

The doc recommends leading with an `@` prefix when specifying nonliteral parameters values. This guidance contradicts the default behavior. If we really wanted everyone to specify a leading `@` then we would have made it required.  The C# context is unambiguous from a language perspective and is general evident when reading the code. There is some potential to confuse C# expressions with string literals if you don't know the parameter type, but it's not clear to me that justifies always including the `@`. I think we should remove this guidance and instead clarify that including the leading `@` is optional. Users can then make their own stylistic choice. Stylistic guidance is of course always up for debate, so adding @mkArtakMSFT in case engineering feels strongly about this.

---
#### Document Details

* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #28694

This topic is one of the handful that hasn't been reached to drop the *Whole-topic Versioning*, so there are four copies of the text here. Resolving topics down to ***one copy*** 👍 is an on-going task this year, and I'll get to this topic soon enough! 😄

The `ParameterParent2` component is in the Blazor sample app. Steve, I kept it the way that it was with separate `Title` assignments to plainly demonstrate that they all work to the reader. This is the first time that a new-to-Blazor dev is seeing this. I understand that from the framework's perspective that there's no difference in these assignments.

https://github.com/dotnet/blazor-samples/blob/main/7.0/BlazorSample_WebAssembly/Pages/index/ParameterParent2.razor

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/3f69826dca1257e74d9c00cc7e234d2c319dd6d4/aspnetcore/blazor/components/index.md) | [aspnetcore/blazor/components/index](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/index?branch=pr-en-us-28696) |

### DIFF
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -290,8 +290,8 @@ Quotes around parameter attribute values are optional in most cases per the HTML
 Throughout the documentation, code examples:
 
 * Always use quotes. Example: `Value="this"`.
-* Nonliterals always use the `@` prefix, ***even when it's optional***. Example: `Count="@ct"`, where `ct` is a number-typed variable. `Count="ct"` is a valid stylistic approach, but the documentation and examples don't adopt the convention.
-* Literals, outside of Razor expressions, always avoid `@`. Example: `IsFixed="true"`.
+* Use the `@` prefix with nonliterals, ***even when it's optional***. Example: `Count="@ct"`, where `ct` is a number-typed variable. `Count="ct"` is a valid stylistic approach, but the documentation and examples don't adopt the convention.
+* Always avoid `@` for literals, outside of Razor expressions. Example: `IsFixed="true"`.
 
 `Pages/ParameterParent2.razor`:
 
@@ -1822,8 +1822,8 @@ Quotes around parameter attribute values are optional in most cases per the HTML
 Throughout the documentation, code examples:
 
 * Always use quotes. Example: `Value="this"`.
-* Nonliterals always use the `@` prefix, ***even when it's optional***. Example: `Count="@ct"`, where `ct` is a number-typed variable. `Count="ct"` is a valid stylistic approach, but the documentation and examples don't adopt the convention.
-* Literals, outside of Razor expressions, always avoid `@`. Example: `IsFixed="true"`.
+* Use the `@` prefix with nonliterals, ***even when it's optional***. Example: `Count="@ct"`, where `ct` is a number-typed variable. `Count="ct"` is a valid stylistic approach, but the documentation and examples don't adopt the convention.
+* Always avoid `@` for literals, outside of Razor expressions. Example: `IsFixed="true"`.
 
 `Pages/ParameterParent2.razor`:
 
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/components/index.md
```diff
@@ -270,16 +270,18 @@ The following rendered HTML markup from the `ParameterParent` component shows `P
 </div>
 ```
 
-Assign a C# field, property, or result of a method to a component parameter as an HTML attribute value using [Razor's reserved `@` symbol](xref:mvc/views/razor#razor-syntax). The following `ParameterParent2` component displays four instances of the preceding `ParameterChild` component and sets their `Title` parameter values to:
+Assign a C# field, property, or result of a method to a component parameter as an HTML attribute value.
+
+A leading [Razor reserved `@` symbol](xref:mvc/views/razor#razor-syntax) isn't required for component parameters of any type. A leading `@` symbol is optional for non-string parameters and is only required for string parameters if you want a C# context instead of assigning a string literal value.
```
danroth27, line 275:
```suggestion
Assign a C# field, property, or result of a method to a component parameter as an HTML attribute value. The value of the attribute can typically be any C# expression that matches the type of the parameter. The value of the attribute can optionally lead with a [Razor reserved `@` symbol](xref:mvc/views/razor#razor-syntax), but it isn't required.
```

aspnetcore/blazor/components/index.md
```diff
@@ -270,16 +270,18 @@ The following rendered HTML markup from the `ParameterParent` component shows `P
...
+
+The following `ParameterParent2` component displays four instances of the preceding `ParameterChild` component and sets their `Title` parameter values to:
```
danroth27, line 277:
```suggestion

If the component parameter is of type string, then the attribute value is instead treated as a C# string literal by default.  If you want to specify a C# expression instead, then use the `@` prefix.

The following `ParameterParent2` component displays four instances of the preceding `ParameterChild` component and sets their `Title` parameter values to:
```

aspnetcore/blazor/components/index.md
```diff
@@ -270,16 +270,18 @@ The following rendered HTML markup from the `ParameterParent` component shows `P
...
 
 * The value of the `title` field.
 * The result of the `GetTitle` C# method.
 * The current local date in long format with <xref:System.DateTime.ToLongDateString%2A>, which uses an [implicit C# expression](xref:mvc/views/razor#implicit-razor-expressions).
 * The `panelData` object's `Title` property.
 
-The `@` prefix is required for string parameters. Otherwise, the framework assumes that a string literal is set.
-
-Outside of string parameters, we recommend the use of the `@` prefix for nonliterals, even when they aren't strictly required.
+The `@` prefix is required for string-typed parameters in a C# context. Otherwise, the framework assumes that a string literal is set. The `@` prefix is ***never*** required for other, non-literal types.
```
danroth27, line 284:
```suggestion
```

danroth27 (changes requested):
Some suggested edits.

guardrex:
@danroth27 ... I think that gets it. Let me know if you want me to wait for Steve to look 👀.

SteveSandersonMS (approved):
Looks good

PR #28750, opened 2023-03-23
Fix minor typos and formatting

garyng:
Fix minor typos and formatting

### DIFF
--- a/aspnetcore/mvc/controllers/routing.md
+++ b/aspnetcore/mvc/controllers/routing.md
@@ -302,7 +302,7 @@ Using `page` as a route parameter with attribute routing is a common error. Doin
 
 The special parameter names are used by the URL generation to determine if a URL generation operation refers to a Razor Page or to a Controller.
 
-* The following keywords are reserved in the context of a Razor view or a Razor Page:
+The following keywords are reserved in the context of a Razor view or a Razor Page:
 
 * `page`
 * `using`
--- a/aspnetcore/mvc/controllers/routing.md
+++ b/aspnetcore/mvc/controllers/routing.md
@@ -349,16 +349,16 @@ Consider the following controller:
 In the preceding code:
 
 * Each action contains the `[HttpGet]` attribute, which constrains matching to HTTP GET requests only.
-* The `GetProduct` action includes the `"{id}"` template, therefore `id` is appended to the `"api/[controller]"` template on the controller. The methods template is `"api/[controller]/"{id}""`. Therefore this action only matches GET requests for the form `/api/test2/xyz`,`/api/test2/123`,`/api/test2/{any string}`, etc.
+* The `GetProduct` action includes the `"{id}"` template, therefore `id` is appended to the `"api/[controller]"` template on the controller. The methods template is `"api/[controller]/{id}"`. Therefore this action only matches GET requests for the form `/api/test2/xyz`,`/api/test2/123`,`/api/test2/{any string}`, etc.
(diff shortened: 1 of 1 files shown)


guardrex (changes requested):
Hello @garyng ... Thanks for sending this in. This topic contains two versions of the content. Could you also update the second copies of this content (down at and below Line 1147)?

Rick-Anderson:
@guardrex  all we care about is the latest version and the latest version is on the top.

Rick-Anderson (approved):

Rick-Anderson:
@garyng  massive improvements. 
 Be sure to check out your name as a contributor about the 3rd of next month on the [What's new page](https://docs.microsoft.com/en-us/aspnet/core/whats-new/) in the **Community contributors** section at the bottom.

Hopefully you'll keep contributing.

guardrex:
Some of these aren't minor typos.

garyng:
Heyy, I didn't notice there are 2 copies of the same content. I created the https://github.com/dotnet/AspNetCore.Docs/pull/28760 for replicating the changes.

Thank you!

guardrex:
@garyng, this reminds me of something that Ryan Nowak (an Azure engineer/manager these days) said years ago pertaining to engineering guidelines that some things are under "***under civil debate***" 😄 as "camps" of engineers support one approach or another. The topic of that discussion was the use of the `this` keyword versus underscoring class variables. Today for docs pertaining to your ***two*** PRs, it's the use of what we call *Whole-topic Versioning*, where each release, all of the content of a doc is copied and updated.

Over time, the copies pile up. For a content change across versions, readers often don't understand that they should search and destroy the content problem in ***multiple copies*** of the doc. Doc authors are aware of this, but I think it's safe to say that we all find it tedious to make a single change multiple times down several copies of a doc, thus @Rick-Anderson's remark about just making the updates in prior versions for the most serious class of problems, security updates.

I've been converting Blazor docs back to normal versioning, *and I really like IT!* 👍😄 It's saving me time and aggravation, thus it's saving MS 💰. Readers who submit a PR for a Blazor doc only need to find and fix the *one spot* of trouble in a doc.

There might be a couple of docs that will benefit from the *Whole-topic Versioning* approach because if a doc flips wildly in and out of versioning tags, it can get ***very*** confusing and hard to maintain a doc. In those cases, I'll keep *Whole-topic Versioning*, but most Blazor docs are easier to maintain without it ... with ***one copy*** of the content and the normal use of versioning tags.

Going back to normal versioning is also avoiding a serious problem with copies of the content: ***broken links*** 🙈. When there are multiple full copies of the content, the headings are copied. Headings form the basis of bookmarked cross-links. The docs build system ***numbers*** the `id`s of the duplicate headings. For example, `## My heading` copied in markdown leads to ...

* `<h2 id="my-heading">My heading</h2>`
* `<h2 id="my-heading-1">My heading</h2>`
* ...
* `<h2 id="my-heading-X">My heading</h2>`

However, links to the section elsewhere in the docs ***aren't numbered*** (i.e., `.../#my-heading`), so bookmarks are fine for the current version but break 💥 in older versions. We get fairly regular negative reader feedback on this. It's also just flat-out ***embarrassing*** ... *MS can't build a website that isn't full of broken links!* 🙈😄 The Blazor docs that have dropped copies of content also drop the multiple headings, so the Blazor docs are avoiding the problem of broken links for prior versions.

*Generally, Ryan was right!* 😄 Some things, including on this subject, will remain ⚔️ "***under civil debate***" ⚔️ here. 😆 Versioning content is challenging, and there are several ways that it can be accomplished, each with pros and cons.

PR #25791, opened 2022-05-03
Blazor Hybrid security content

guardrex:
Fixes #25771

\[Internal Review Topic]()

Javier, notes and ❓ ... I'll number them in case you need to discuss/answer a point.

1. Plz review the cross-links. I'm not sure if they're the best or in the best order.
2. I couldn't find WinForms with AAD/AAD B2C guidance in WinForms, Azure, or MSAL docs.
3. Signal me to revert to your lingo if you like your phrasing better anywhere (or edit directly).
4. If this requires a lot of work, I recommend **_editing the PR directly_** over placing *many* suggestions. I'll circle around with a grammer/style commit after your updates.
5. You said that they can set the `Threading.Thread.Principal` ... did you mean `System.Threading.Thread.CurrentPrincipal`? ... or is there a property added for `Principal` by an API implementation somewhere?
6. WRT the line I asked in the email about ... "From anywhere in the app, we can resolve ..." I still think we'll need to break that out into a few sentences and explain it out a bit more. For example, don't we need to add a bit that says that the code shown is for the developer's implementation for authenticating users using the identity provider's SDK?
7. I removed underscores from the code example, but let me know if there was a reason that you want them ... I'll revert if that's the case.
8. Currently, I'm avoiding localization of "Web View" across the Hybrid docs with the inline markers `:::no-loc text="Web View":::`. I might move that to a global no-loc later. This is fine for now tho.

PR #28751, opened 2023-03-23
Change MainWindow.cs to MauiProgram.cs

thomasclaudiushuber:
Use the correct filename that contains the static `CreateMauiApp` method.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/hybrid/security/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/62d4ee4c510c6bc8f6f2244abb80772d05c6275a/aspnetcore/blazor/hybrid/security/index.md) | [aspnetcore/blazor/hybrid/security/index](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/hybrid/security/index?branch=pr-en-us-28751) |

### DIFF
--- a/aspnetcore/blazor/hybrid/security/index.md
+++ b/aspnetcore/blazor/hybrid/security/index.md
@@ -447,7 +447,7 @@ In the preceding example:
 
 :::zone pivot="maui"
 
-In the `MauiProgram.CreateMauiApp` method of `MainWindow.cs`, add the authorization services and the Blazor abstraction to the service collection:
+In the `MauiProgram.CreateMauiApp` method of `MauiProgram.cs`, add the authorization services and the Blazor abstraction to the service collection:
 
 ```csharp
 builder.Services.AddAuthorizationCore();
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks @thomasclaudiushuber ... This is probably a recent change that I didn't hear about. Do you know when they made the change?

thomasclaudiushuber:
Hi @guardrex, in my opinion it's not a recent change, it has always been that way for .NET MAUI apps.

There was even never a `MainWindow.cs` file in MAUI. MAUI contains a `MainPage.cs`, but not a `MainWindow.cs`. But the code with the static `CreateMauiProgram` method was always in a `MauiProgram.cs` file, this is true for the .NET 6 and for the .NET 7 project template. I think `MainWindow.cs` was a copy/paste issue from something about WPF, because there that file exists.

guardrex:
Yes, I think you're correct. The rest of the docs use the correct filename. This is the only doc that seems to have a ***grimlin*** 😈 in it. It came in on this PR ... https://github.com/dotnet/AspNetCore.Docs/pull/25791 ... and that was derived from PU notes. I think that when I was writing it up I just flat-out had the wrong filename in my head 🙈. It was reviewed, but it wasn't caught by Jaiver at the time. That's how the sneaky ones 😈😈😈 really do well ... I put them in ***AND*** they escape review.

Stand-by for a sec ... I just need to make sure that they're all out of this before we merge it.

thomasclaudiushuber:
Thank you @guardrex, looks good now. :)👍 ✨

Issue #28161, opened 2023-01-21
Blazor .NET 8 tracking issue


guardrex:
## 8.0 Post GA

***NOTE***: Probably best to have the PWA doc reviewed by the PU for any other 8.0 PWA changes.

Status | Issue | PR | Item
:---: | :---: | :---: | ---
🚧<br>**IN PROGRESS**<br>👷 | #32020 | &mdash; | Update Blazor Web App glob/loc approach
🟢 | #31691 | &mdash; | Update section on passing tokens in Blazor Web Apps<br><br>***Waiting on guidance from Halter/Jeremy.***
🟢 | #31381 | &mdash; | Additional or different bearer token remarks for WASM security
✔️ | #32008 | #32028 | Cookie security for Call web API article
✔️ | #32082 | #32083 | Describe dependency on DBContext (vs DBContextFactory) for Blazor Identity UI
✔️ | #32077 | #32081 | Cover nonce update for token refresh
✔️ | #31402 | #32048 | Interactive rendering required for `NotAuthorized` component/content to work ... and I'll probably work this in tandem with `NotFound` content behavioral differences between ASP.NET Core pipeline on the server and WASM on the client.
✔️ | #32054 | #32060 | Blazor framework static files served via endpoint routing
✔️ | #31383 | #31431 | Account confirmation+PW recovery article for BWA/Identity components **UPDATE**: I asked multiple times for PU review, and it was never followed up on by PU engineers.
✔️ | #31931 | #32035 | Using fallback authorization policy breaks Blazor Server after migrating to .NET 8 
❌ | #31899 | &mdash; | Cookie example does not work for .Net 8 template because the state container does not instantiate in user scope ***Dup tracked by #39691.***
✔️ | #30002 | #31894 | Dynamic culture approach updates (last one is maintaining state across interactivity location change)
✔️ | #31973 | #31991 | Security aspects in the Call web API article
✔️ | #31887 | #31961 | Cover web API approaches for the BWA Auto render mode scenario
✔️ | #31389 |  #31888 | Samesite + addl config and guidance for WASM+Identity different domain hosting
✔️ | #31379 | #31934 | QuickGrid article API doc cross-links
✔️ | #31205 | https://github.com/dotnet/blazor-samples/pull/136 | BlazorWebAssemblyStandaloneWithIdentity uses cookie auth but no antiforgerytoken
✔️ | #31741 | #31873 | Adopt 8.0 conventions for inline Blazor forms
[...]

Issue #28001, opened 2023-01-03
Blazor project tracking 2023


guardrex:
# 2023

## History

* For the 2022, see [Blazor project tracking 2022 (dotnet/AspNetCore.Docs #24615)](https://github.com/dotnet/AspNetCore.Docs/issues/24615).
* For the 2021, see [Blazor project tracking 2021 (dotnet/AspNetCore.Docs #19286)](https://github.com/dotnet/AspNetCore.Docs/issues/19286).

## Doc ideas

Not **ALL** of these will be worked. This is an idea list/check list that don't rise to the level of opening an issue at this time.

* [ ] Consider adding a code example for the `MutationObserver` pattern where its discussed.
* [ ] File uploads topic: Fix accessibility of the yes/no icons. Use: `<span aria-hidden="true">XXXXX</span><span class="visually-hidden">XXXXX</span>`
* [ ] Add guidance on ***the reason(s)*** for using partial classes and base classes (inheritance) with pros and cons of each (partial = only splitting, base class = splitting and multiple-inheritance, useful in library-provided API scenarios).
* [ ] Per Steve's .NET Conf talk, shouldn't we briefly cover [`FileSystem` API](https://developer.mozilla.org/en-US/docs/Web/API/FileSystem) use in Blazor apps?
* [ ] Let's run through the [`testassets` pieces in the framework](https://github.com/dotnet/aspnetcore/tree/main/src/Components/test/testassets) to see what might be helpful for devs to see in docs. It's a large task tho because there's so much there. This probably can't be worked in 23H1, possibly for 23Q3 tho.
* [ ] There's great PU-provided memory leak guidance on https://github.com/dotnet/aspnetcore/issues/43221. Take a look and see if/what could go into docs, possibly even just by linking the PU issue to cover it in the near-term.
* [ ] More for '23 than '22, I'm moving over a wrap-up item from the 7.0 tracking issue to this issue for work later: "Open an issue to replace in-text (usually non-working) examples with samples repo-based fully working cut-'n-paste examples. List the sections that require updates. This issue will be targeted for EOY or 23Q1."
* [ ] It would probably be nice to create a `QuickGrid` example that opens and closes a detail record component without losing the page and scroll position of the grid. For context, see https://github.com/dotnet/blazor-samples/issues/58.
[...]

PR #28767, opened 2023-03-24
Blazor QuickGrid coverage enhancements

guardrex:
Addresses #28001
Addresses #28161

@Nick-Stanton, do you want to take a look at *rough* 8.0 draft coverage for `QuickGrid`? The coverage is a WIP 🚧 that I'll be picking back up with late next week, as I'm going to be OOF Monday thru Wednesday. I'll probably merge this now but come back with another PR late next week. If you don't have time to provide feedback now but will have time later, I can ping u on the next PR next week.

Note in passing that we ***might*** want to have a more advanced example than this. However, we're linking to the [**QuickGrid for Blazor** app](https://aspnet.github.io/quickgridsamples/), and that might be all we need to do for an advanced example.

Note also that this whole subject (section) might be better off in a dedicated topic. I'm really feel'in the idea because the *Components* overview article is already ***too long***. I'm looking for ways to shorten the whole topic generally, and this would stand nicely as its own article in the *Components* node (folder). *Yeah!* The more I think about it, the more ***I LIKE IT!*** 😆

### DIFF
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -1480,7 +1480,53 @@ For more information, see the following resources:
 
 The `QuickGrid` component is a Razor component for quickly and efficiently displaying data in tabular form. `QuickGrid` provides a simple and convenient data grid component for common grid rendering scenarios and serves as a reference architecture and performance baseline for building data grid components. `QuickGrid` is highly optimized and uses advanced techniques to achieve optimal rendering performance.
 
-To get started with `QuickGrid`:
+To implement a `QuickGrid` component:
+
+* Specify tags for the `QuickGrid` component in Razor markup (`<QuickGrid>...</QuickGrid>`).
+* Name a queryable source of data for the grid. Use ***either*** of the following data sources:
+  * `Items`: A nullable `IQueryable<TGridItem>`, where `TGridItem` is the type of data represented by each row in the grid.
+  * `ItemsProvider`: A callback that supplies data for the grid.
+* `Class`: An optional CSS class name. If provided, the class name is included in the `class` attribute of the rendered table.
+* `Theme`: A theme name (default value: `default`). This affects which styling rules match the table.
+* `Virtualize`: If true, the grid is rendered with virtualization. This is normally used in conjunction with scrolling and causes the grid to fetch and render only the data around the current scroll viewport. This can greatly improve the performance when scrolling through large data sets. If you use `Virtualize`, you should supply a value for `ItemSize` and must ensure that every row renders with a constant height. Generally, it's preferable not to use `Virtualize` if the amount of data rendered is small or if you're using pagination.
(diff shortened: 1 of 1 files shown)


Nick-Stanton:
This looks great @guardrex! I would suggest is adding that for people using EF IQueryables, we suggest installing the separate `Microsoft.AspNetCore.Components.QuickGrid.EntityFrameworkAdapter` package.

guardrex:
Thanks @Nick-Stanton ...

WRT the package, that is versioned here, so that should be ok. I'll look it over one more time tomorrow to confirm.

WRT EF IQueryables, yes, I'll work that into the coverage.

guardrex:
@Nick-Stanton ... Is the `Microsoft.AspNetCore.Components.QuickGrid.EntityFrameworkAdapter` package going to come in transitively under .NET 8? If so, then I only need to instruct that it be explicitly added for .NET 7.

Also, is there an example somewhere of that API being used with an EF `IQueryable` that I can drop into the coverage as an example?

guardrex:
I added a line to install the EF adapter package for .NET 7 to use `IQueryable` with EF Core. I'm going to add a cross-link for the Basic Test App EF Core component next.

If the adapter package is transitively added for .NET 8, then the earlier bit on data sources probably don't require modification: An `IQueryable<T>` is for any query provider AFAIK, and I don't think we need to say anything specific about EF there. However, I'll add something if you like.

Nick-Stanton:
I'm not sure we have specific tests for the EF IQueryables, which may need to be reevaluated. Here's an example of how it might be consumed from Steve's samples.

https://aspnet.github.io/quickgridsamples/datasources

The project reference looks a little funny here because it's referring to the version in the repo instead of from NuGet.
https://github.com/aspnet/AspLabs/blob/cbca15978abbf6e0eb865d78dd9d8640790fbebf/src/QuickGrid/samples/QuickGridSamples.Server/QuickGridSamples.Server.csproj#L26

https://github.com/aspnet/AspLabs/blob/cbca15978abbf6e0eb865d78dd9d8640790fbebf/src/QuickGrid/samples/QuickGridSamples.Server/Program.cs#L13

FYI, I moved teams and am no longer working on Blazor. I'm happy to answer what I'm able to when I have time, but @SteveSandersonMS has deeper knowledge of how QG works.

guardrex:
Sorry for the delay to reply, @Nick-Stanton ... I'm OOF for a few days. I'll get back to this later this week when I'm back on my feet. 🤒

guardrex:
> Here's an example of how it might be consumed from Steve's samples.

I have a feel'in that that's going to be too *point-in-time* for us to document from. Those types of examples go stale and break. I think we can cross-link to the `SampleQuickGridComponent` component, as that will be updated over time as they go.

WRT ...

```csharp
builder.Services.AddQuickGridEntityFrameworkAdapter();
```

I see. Yes, I'll get that in here.

... and WRT the package reference, I see that I was incorrect about that from the following error language in the ref source (*emphasis added*) ...

> throw new InvalidOperationException($"The supplied {nameof(IQueryable)} is provided by Entity Framework. To query it efficiently, ***you must reference the package*** Microsoft.AspNetCore.Components.QuickGrid.EntityFrameworkAdapter and call AddQuickGridEntityFrameworkAdapter on your service collection.");

PR #28936, opened 2023-04-10
Dispatch exceptions to the sync context

### LINKED ISSUE #28933: Dispatch exceptions to Blazor’s sync context
* [PU Issue](https://github.com/dotnet/aspnetcore/issues/44920)
* [API Proposal](https://github.com/dotnet/aspnetcore/issues/46068)
* [PU PR](https://github.com/dotnet/aspnetcore/pull/46074)

This will be a H3 under the *Synchronization context* section.

---
#### Document Details

* Content: [ASP.NET Core Razor components](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-6.0#invoke-component-methods-externally-to-update-state)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #28933

Notes ...

* I felt building off of the prior example makes sense. Presumably, many readers will have just tested out the prior section's example, so they can just build on that experience to see this API in action.

  Here's what this new example builds off of ...

  https://learn.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-7.0#invoke-component-methods-externally-to-update-state

* There's one thing in Steve's remarks that I don't understand ...

  > ... the premise that you have an exception already implies you caught it. If you don't catch an external exception, then you have an unhandled exception which has nothing to do with Blazor.

  Cross-ref: https://github.com/dotnet/aspnetcore/issues/46068#issuecomment-1398207040

  ***... but*** error boundaries (getting them to light up here) is about trapping unhandled exceptions. The example that I have on this PR is throwing in the `TimerService` to show how the exception can be forwarded to the Blazor sync context.

  ... also Halter said right after that ...

  > This is not something you'd want to do by default when the caller to InvokeAsync may be able to handle the exception gracefully.

  ... but again, you can do both. You can take action on it, just in the sense that you're processing it in the external code (e.g., log it), and then flow it ***unhandled*** to Blazor for the error boundary to trip on purpose. I don't see why this isn't a valid approach.

  I guess what I'm saying is that if these PU remarks are right, we don't want to show this example based on an exception in `TimerService` ... `TimerService` should handle its exceptions entirely and flow back something to the caller for processing in the Razor component on an error, not allow the unhandled exception to trip the error boundary. What I have here might be considered as an implied ***anti-pattern*** 🙈 and should be replaced with a different example.

  Besides that, is there some other or additional scenario that you'd like to show? I like this one, but I'm just pitching it as a possibility because the prior example is so good (IMO) and feeds right into the usefulness of this API AFAICT.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/c969e6921acbe0ddf72a21ffb8d116c2352ee66b/aspnetcore/blazor/components/index.md) | [ASP.NET Core Razor components](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/index?branch=pr-en-us-28936) |
| [aspnetcore/blazor/fundamentals/handle-errors.md](https://github.com/dotnet/AspNetCore.Docs/blob/c969e6921acbe0ddf72a21ffb8d116c2352ee66b/aspnetcore/blazor/fundamentals/handle-errors.md) | [aspnetcore/blazor/fundamentals/handle-errors](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/handle-errors?branch=pr-en-us-28936) |

### DIFF
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -810,7 +810,13 @@ catch (Exception ex)
 }
 ```
 
-For an example use of `DispatchExceptionAsync` API, implement the timer notification example in the [Invoke component methods externally to update state](#invoke-component-methods-externally-to-update-state) section. The example uses a timer outside of Blazor's synchronization context, where an unhandled exception normally doesn't reach Blazor's synchronization context and isn't processed by Blazor's error handling mechanisms, such as an [error boundary](xref:blazor/fundamentals/handle-errors#error-boundaries).
+For an example use of `DispatchExceptionAsync` API, implement the timer notification example in the [Invoke component methods externally to update state](#invoke-component-methods-externally-to-update-state) section. In a Blazor app, add the following files from the timer notification example and register the services in `Program.cs` as the text explains:
+
+* `TimerService.cs`
+* `NotifierService.cs`
+* `Pages/ReceiveNotifications.razor`
+
+The example uses a timer outside of Blazor's synchronization context, where an unhandled exception normally doesn't reach Blazor's synchronization context and isn't processed by Blazor's error handling mechanisms, such as an [error boundary](xref:blazor/fundamentals/handle-errors#error-boundaries).
 
 First, change the code in `TimerService.cs` to create an artificial exception outside of Blazor's synchronization context. In the `while` loop of `TimerService.cs`, throw an exception when the `elapsedCount` reaches a value of two:
 
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -810,9 +810,9 @@ catch (Exception ex)
 }
 ```
 
-For an example use of `DispatchExceptionAsync` API, implement the timer notification example in the [Invoke component methods externally to update state](#invoke-component-methods-externally-to-update-state) section. The example uses a timer outside of Blazor's synchronization context, where an exception normally doesn't reach Blazor's synchronization context and isn't processed by Blazor's error handling mechanisms, such as an [error boundary](xref:blazor/fundamentals/handle-errors#error-boundaries).
+For an example use of `DispatchExceptionAsync` API, implement the timer notification example in the [Invoke component methods externally to update state](#invoke-component-methods-externally-to-update-state) section. The example uses a timer outside of Blazor's synchronization context, where an unhandled exception normally doesn't reach Blazor's synchronization context and isn't processed by Blazor's error handling mechanisms, such as an [error boundary](xref:blazor/fundamentals/handle-errors#error-boundaries).
 
-First, change the code in `TimerService.cs` to create an artificial exception outside of Blazor's synchronization context. In the `while` loop of `TimerService.cs`, throw an exception when the `elapsedCount` reaches a value of `2`:
+First, change the code in `TimerService.cs` to create an artificial exception outside of Blazor's synchronization context. In the `while` loop of `TimerService.cs`, throw an exception when the `elapsedCount` reaches a value of two:
 
 ```csharp
 if (elapsedCount == 2)
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -783,101 +783,11 @@ In the preceding example:
 > [!IMPORTANT]
 > If a Razor component defines an event that's triggered from a background thread, the component might be required to capture and restore the execution context (<xref:System.Threading.ExecutionContext>) at the time the handler is registered. For more information, see [Calling `InvokeAsync(StateHasChanged)` causes page to fallback to default culture (dotnet/aspnetcore #28521)](https://github.com/dotnet/aspnetcore/issues/28521).
 
-:::moniker-end
-
 :::moniker range=">= aspnetcore-8.0"
 
-### Dispatch exceptions to Blazor's synchronization context
-
-Use `ComponentBase.DispatchExceptionAsync` in Razor components to process exceptions thrown outside of Blazor's synchronization context. This permits the app's code to treat the exceptions as through they're lifecycle method exceptions. Thereafter, Blazor's error handling mechanisms, such as [error boundaries](xref:blazor/fundamentals/handle-errors#error-boundaries), can process the exceptions.
-
-> [!NOTE]
-> `ComponentBase.DispatchExceptionAsync` is used in Razor component files (`.razor`) that inherit from <xref:Microsoft.AspNetCore.Components.ComponentBase>. When creating components that [implement <xref:Microsoft.AspNetCore.Components.IComponent> directly](#component-classes), use `RenderHandle.DispatchExceptionAsync`.
-
-To dispatch exceptions to Blazor's synchronization context, use a `try-catch` block to pass the exception to `DispatchExceptionAsync` and await the result:
-
-```csharp
-try
-{
-    await InvokeAsync(() =>
-    {
-        ...
-    });
-}
-catch (Exception ex)
-{
-    await DispatchExceptionAsync(ex);
-}
-```
-
-For an example use of `DispatchExceptionAsync` API, implement the timer notification example in the [Invoke component methods externally to update state](#invoke-component-methods-externally-to-update-state) section. In a Blazor app, add the following files from the timer notification example and register the services in `Program.cs` as the text explains:
-
-* `TimerService.cs`
-* `NotifierService.cs`
-* `Pages/ReceiveNotifications.razor`
-
-The example uses a timer outside of Blazor's synchronization context, where an unhandled exception normally doesn't reach Blazor's synchronization context and isn't processed by Blazor's error handling mechanisms, such as an [error boundary](xref:blazor/fundamentals/handle-errors#error-boundaries).
-
-First, change the code in `TimerService.cs` to create an artificial exception outside of Blazor's synchronization context. In the `while` loop of `TimerService.cs`, throw an exception when the `elapsedCount` reaches a value of two:
-
-```csharp
-if (elapsedCount == 2)
-{
-    throw new Exception("I threw an exception! Somebody help me!");
-}
-```
-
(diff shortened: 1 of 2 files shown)


aspnetcore/blazor/components/index.md
```diff
@@ -783,6 +783,96 @@ In the preceding example:
...
+}
+catch (Exception ex)
+{
+    await DispatchExceptionAsync(ex);
+}
+```
+
+For an example use of `DispatchExceptionAsync` API, implement the timer notification example in the [Invoke component methods externally to update state](#invoke-component-methods-externally-to-update-state) section. The example uses a timer outside of Blazor's synchronization context, where an unhandled exception normally doesn't reach Blazor's synchronization context and isn't processed by Blazor's error handling mechanisms, such as an [error boundary](xref:blazor/fundamentals/handle-errors#error-boundaries).
+
+First, change the code in `TimerService.cs` to create an artificial exception outside of Blazor's synchronization context. In the `while` loop of `TimerService.cs`, throw an exception when the `elapsedCount` reaches a value of two:
```
mkArtakMSFT, line 815: I know this is similar to the existing pattern we follow in the docs, where we build on top of an example from earlier in the page, but this model is fragile. In case of a reader following the documentation from the top, they will sure be successful, but if they've landed on this particular section through a link, they will start guessing what `TimerService` is. Hence, I think it'll be good to explicitly call it out like so:

```suggestion
First, change the code in `TimerService.cs` (from an earlier example in this page) to create an artificial exception outside of Blazor's synchronization context. In the `while` loop of `TimerService.cs`, throw an exception when the `elapsedCount` reaches a value of two:
```

aspnetcore/blazor/components/index.md
```diff
@@ -783,6 +783,96 @@ In the preceding example:
...
+        <ErrorContent>
+            <p class="alert alert-danger" role="alert">
+                Oh, dear! Oh, my! - George Takei
+            </p>
+        </ErrorContent>
+    </ErrorBoundary>
+</article>
+```
+
+Normally, the preceding error boundary doesn't react to unhandled exceptions thrown outside of Blazor's synchronization context. If you run the app at this point, the exception is thrown when the elapsed count reaches a value of two, but the error boundary doesn't show the error content.
```
mkArtakMSFT, line 843: The `Normally` implies, that there is a "trick" to do and it will show, but given the current state of code, it won't. Hence, I think we should remove the `Normally,` part.

```suggestion
If you run the app at this point, the exception is thrown when the elapsed count reaches a value of two, but nothing on the UI changes and the error boundary doesn't show the error content.
```

mkArtakMSFT (approved):
Thanks @guardrex !

aspnetcore/blazor/components/index.md
guardrex: Sounds good, but we're running on a bit with a looooong sentence. Let's break it it up into separate thoughts (probably better for MT anyway) .....

```suggestion
If you run the app at this point, the exception is thrown when the elapsed count reaches a value of two. However, nothing in the UI changes. The error boundary doesn't show the error content.
```

aspnetcore/blazor/components/index.md
guardrex: Let me try and fix this by making it more explicit in the preceding paragraph. That's why I started by telling the reader there that they would be building on that example. I'll try and resolve this on the next commit.

aspnetcore/blazor/components/index.md
```diff
@@ -783,6 +783,102 @@ In the preceding example:
 > [!IMPORTANT]
 > If a Razor component defines an event that's triggered from a background thread, the component might be required to capture and restore the execution context (<xref:System.Threading.ExecutionContext>) at the time the handler is registered. For more information, see [Calling `InvokeAsync(StateHasChanged)` causes page to fallback to default culture (dotnet/aspnetcore #28521)](https://github.com/dotnet/aspnetcore/issues/28521).
 
+:::moniker-end
+
+:::moniker range=">= aspnetcore-8.0"
+
+### Dispatch exceptions to Blazor's synchronization context
+
+Use `ComponentBase.DispatchExceptionAsync` in Razor components to process exceptions thrown outside of Blazor's synchronization context. This permits the app's code to treat the exceptions as through they're lifecycle method exceptions. Thereafter, Blazor's error handling mechanisms, such as [error boundaries](xref:blazor/fundamentals/handle-errors#error-boundaries), can process the exceptions.
```
SteveSandersonMS, line 792: It's not about being outside the sync context. It's about being outside the lifecycle methods. People should use
DispatchExceptionAsync when:

 * Some code *outside the call stack of a lifecycle method* causes an exception, which they catch
 * They want it to be handled as if it was within the lifecycle of a specific component instance (for example, triggering the error boundaries for that component)

This is true independently of whether their code is running on the sync context. The sync context doesn't come into it.

aspnetcore/blazor/components/index.md
```diff
@@ -783,6 +783,102 @@ In the preceding example:
...
+
+> [!NOTE]
+> `ComponentBase.DispatchExceptionAsync` is used in Razor component files (`.razor`) that inherit from <xref:Microsoft.AspNetCore.Components.ComponentBase>. When creating components that [implement <xref:Microsoft.AspNetCore.Components.IComponent> directly](#component-classes), use `RenderHandle.DispatchExceptionAsync`.
+
+To dispatch exceptions to Blazor's synchronization context, use a `try-catch` block to pass the exception to `DispatchExceptionAsync` and await the result:
```
SteveSandersonMS, line 797: This seems a bit odd because:

1. There's not necessarily any need to use `InvokeAsync` here - that depends on what code they are going to call. Whether or not `InvokeAsync` is required is unrelated to the scenario.
2. We probably don't need to explain to people how to catch exceptions - that's a basic C# topic which we can assume readers already know.

Of course you *can* use try/catch in your example, and it makes sense to do so, but it's just odd to point it out in the description as if there's some other way to get exceptions.

aspnetcore/blazor/components/index.md
```diff
@@ -783,6 +783,102 @@ In the preceding example:
...
+}
+```
+
+For an example use of `DispatchExceptionAsync` API, implement the timer notification example in the [Invoke component methods externally to update state](#invoke-component-methods-externally-to-update-state) section. In a Blazor app, add the following files from the timer notification example and register the services in `Program.cs` as the text explains:
+
+* `TimerService.cs`
+* `NotifierService.cs`
+* `Pages/ReceiveNotifications.razor`
+
+The example uses a timer outside of Blazor's synchronization context, where an unhandled exception normally doesn't reach Blazor's synchronization context and isn't processed by Blazor's error handling mechanisms, such as an [error boundary](xref:blazor/fundamentals/handle-errors#error-boundaries).
```
SteveSandersonMS, line 819: This is also a bit off-track because it's talking about whether things are on the sync context, which is unrelated to the use for `DispatchExceptionAsync`. You are correct to use something like a timer as an example of something that is not within the call stack of a lifecycle method, though, and that is the relevant thing for `DispatchExceptionAsync`.

SteveSandersonMS:
> Besides that, is there some other or additional scenario that you'd like to show? I like this one, but I'm just pitching it as a possibility because the prior example is so good (IMO) and feeds right into the usefulness of this API AFAICT.

Another common scenario is if a component wants to start an async operation in a fire-and-forget way. So it's not going to await the Task, and it doesn't get back a result, **but** if the operation fails it wants to treat it as a lifecycle exception to put the whole component into a faulted state, or to terminate the circuit (if it's Blazor Server and there is no error boundary), and to trigger the same logging that occurs for lifecycle exceptions.

Example:

```razor
<button @onclick="SendReport">Send report</button>

@code {
    void SendReport()
    {
        // In most cases you'd want to await this task and update the UI afterwards to say the operation completed
        // But for this example, we're doing it in a fire-and-forget way, so we don't return a Task from this method
        _ = SendReportFireAndForgetAsync();
    }

    async Task SendReportFireAndForgetAsync()
    {
        try
        {
            await ReportSender.SendAsync();
        }
        catch (Exception ex)
        {
            // Because we intentionally discarded the task above, any async failures would occur off the normal
            // lifecycle call stack, and hence would not be seen by Blazor. We can treat them like lifecycle method
            // exceptions by explicitly dispatching them back to the component.
            await DispatchExceptionAsync(ex);
        }
    }
}
```

guardrex:
Thanks @SteveSandersonMS ... I'll make these updates Monday morning and ping u back.

guardrex:
@SteveSandersonMS ... Ready ...

* We may need to work on the wording a bit more, but I dropped the "sync context" lingo.
* Best to place it in the *Handle errors* article.
* I draft it with the non-dispatching "Send Report" example first, then the dispatching example.
* Best to put your example first, then the working demo with the timer.

aspnetcore/blazor/fundamentals/handle-errors.md
```diff
@@ -66,6 +66,139 @@ In a Blazor WebAssembly app, customize the experience in the `wwwroot/index.html
 
 The `blazor-error-ui` element is normally hidden due to the presence of the `display: none` style of the `blazor-error-ui` CSS class in the site's stylesheet (`wwwroot/css/site.css` for Blazor Server or `wwwroot/css/app.css` for Blazor WebAssembly). When an error occurs, the framework applies `display: block` to the element.
 
+:::moniker range=">= aspnetcore-8.0"
+
+## Handle caught exceptions outside of a Razor component's lifecycle
+
+Use `ComponentBase.DispatchExceptionAsync` in a Razor component to process exceptions thrown outside of the component's lifecycle call stack. This permits the component's code to treat exceptions as through they're lifecycle method exceptions. Thereafter, Blazor's error handling mechanisms, such as [error boundaries](xref:blazor/fundamentals/handle-errors#error-boundaries), can process the exceptions.
```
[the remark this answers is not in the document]
guardrex: # 🙈😆

PR #28950, opened 2023-04-12
Move Kestrel above IIS in TOC

JamesNK:
Kestrel is the recommended server. Moved it above IIS in TOC to help hint that it should be considered first.

### DIFF
--- a/aspnetcore/toc.yml
+++ b/aspnetcore/toc.yml
@@ -1180,6 +1180,36 @@ items:
           - name: Next steps
             displayName: azure, deploy, publish
             href: /dotnet/architecture/devops-for-aspnet-developers/next-steps
+      - name: Kestrel
+        displayName: deploy, publish, server, Kestrel
+        items:
+          - name: Overview
+            displayName: deploy, publish, server
+            uid: fundamentals/servers/kestrel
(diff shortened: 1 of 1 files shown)


aspnetcore/toc.yml
```diff
@@ -1180,6 +1180,36 @@ items:
           - name: Next steps
             displayName: azure, deploy, publish
             href: /dotnet/architecture/devops-for-aspnet-developers/next-steps
+      - name: Kestrel
+        displayName: deploy, publish, server, Kestrel
+        items:
+          - name: Overview
+            displayName: deploy, publish, server
```
JamesNK, line 1187: What does `displayName` do?

tdykstra: In the text box at the top of the TOC you can type in a filter expression.  `displayName` lets you specify additional words that will respond to the filter expression. If the item `name` is `Overview` and you don't have `displayName` specified, and you enter "deploy" in the filter box, you won't see this toc item in the filtered results.  If you have `displayName: deploy` you'll get this item in the list when you filter on "deploy".

tdykstra (approved):

Rick-Anderson (approved):

PR #28968, opened 2023-04-13
Initial URL for BlazorWebView

### LINKED ISSUE #28965: Specify initial URL for BlazorWebView to load
* https://github.com/dotnet/maui/issues/11746
* https://github.com/dotnet/maui/pull/11942

---
#### Document Details

* ID: 48146c6d-f864-0b56-a2d5-0493a843c7a7
* Version Independent ID: 48146c6d-f864-0b56-a2d5-0493a843c7a7
* Content: [ASP.NET Core Blazor Hybrid routing and navigation](https://learn.microsoft.com/en-us/aspnet/core/blazor/hybrid/routing?view=aspnetcore-7.0)
* Content Source: [aspnetcore/blazor/hybrid/routing.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/hybrid/routing.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #28965

cc: @mkArtakMSFT ... I can see that Surayya did the PR ... I'll ping for review.

Notes:

* Do we want to use this language to describe the default start path? ...

  > The default start path is the relative root URL path (`/`).

  ... because I'm not aware if it's possible for these apps to have a different base path. I'll modify to your best choice of language there.

* Just FYI ... Across the Hybrid docs, we use the approach for setting properties on a `BlazorWebView` that the tutorials adopt. In the case where the dev is looking at the XAML markup code, we show the setting on the element, not via the designer property tools. In the WinForms case, we set those properties in C#, so that's what we place here. We assume the reader used our tutorials, so we assume they'll work with the properties across the docs the same way as the tutorials have them work with them. Just letting you know. This is subject to change later perhaps, but this is the approach we're using thus far.

### DIFF
--- a/aspnetcore/blazor/hybrid/routing.md
+++ b/aspnetcore/blazor/hybrid/routing.md
@@ -112,6 +112,48 @@ blazorWebView.UrlLoading +=
 
 :::zone-end
 
+:::moniker range=">= aspnetcore-8.0"
+
+## Get or set a path for initial navigation
+
+Use the `BlazorWebView.StartPath` property to get or set the path for initial navigation within the Blazor navigation context when the Razor component is finished loading. The default start path is the relative root URL path (`/`).
+
+:::zone pivot="maui"
+
+In the `MainPage` XAML markup (`MainPage.xaml`), specify the start path. The following example sets the path to a welcome page at `/welcome`:
+
+```xaml
+<BlazorWebView ... StartPath="/welcome" ...>
+    ...
+<BlazorWebView>
+```
+
+:::zone-end
+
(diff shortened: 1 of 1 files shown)


surayya-MS (approved):
Looks good to me!

guardrex:
Shall we wait on @Eilon to look at that phrasing before merging?

Eilon (approved):
Yeah I think this is fine, thanks for checking. If for some reason we ever see in the future that it is unclear to someone, we can revisit, but I'd be surprised.

PR #28983, opened 2023-04-14
Monitor Blazor Server circuit activity

### LINKED ISSUE #28982: Monitor Blazor Server circuit activity
* https://github.com/dotnet/aspnetcore/issues/30287
* https://github.com/dotnet/aspnetcore/pull/46968

---
#### Document Details

* Version Independent ID: d17e4c28-ea82-0dd8-295d-19f6ada147ea
* Content: [ASP.NET Core Blazor SignalR guidance](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/signalr?view=aspnetcore-7.0)
* Content Source: [aspnetcore/blazor/fundamentals/signalr.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/fundamentals/signalr.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #28982

### DIFF
--- a/aspnetcore/blazor/fundamentals/signalr.md
+++ b/aspnetcore/blazor/fundamentals/signalr.md
@@ -251,6 +251,161 @@ There was a problem with the connection! (Current reconnect attempt: 3 / 8)
 
 By default, Blazor Server apps prerender the UI on the server before the client connection to the server is established. For more information, see <xref:mvc/views/tag-helpers/builtin-th/component-tag-helper>.
 
+:::moniker-end
+
+:::moniker range=">= aspnetcore-8.0"
+
+## Monitor Blazor Server circuit activity (Blazor Server)
+
+Monitor inbound circuit activity in Blazor Server apps using the `CreateInboundActivityHandler` method on <xref:Microsoft.AspNetCore.Components.Server.Circuits.CircuitHandler>. Inbound circuit activity is any activity sent from the browser to the server, such as UI events or JavaScript-to-.NET interop calls.
+
+For example, you can use a circuit activity handler to detect if the client is idle:
+
+```csharp
+public sealed class IdleCircuitHandler : CircuitHandler, IDisposable
+{
+    readonly Timer timer;
+    readonly ILogger logger;
+
+    public IdleCircuitHandler(IOptions<IdleCircuitOptions> options, 
+        ILogger<IdleCircuitHandler> logger)
+    {
+        timer = new Timer();
+        timer.Interval = options.Value.IdleTimeout.TotalMilliseconds;
+        timer.AutoReset = false;
+        timer.Elapsed += CircuitIdle;
+        this.logger = logger;
+    }
+
+    private void CircuitIdle(object? sender, System.Timers.ElapsedEventArgs e)
+    {
+        logger.LogInformation(nameof(CircuitIdle));
+    }
+
+    public override Func<CircuitInboundActivityContext, Task> CreateInboundActivityHandler(
+        Func<CircuitInboundActivityContext, Task> next)
+    {
+        return context =>
+        {
+            timer.Stop();
+            timer.Start();
+            return next(context);
+        };
+    }
+
+    public void Dispose()
+    {
+        timer.Dispose();
+    }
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/fundamentals/signalr.md
```diff
@@ -251,6 +251,161 @@ There was a problem with the connection! (Current reconnect attempt: 3 / 8)
...
+    public static IServiceCollection AddIdleCircuitHandler(
+        this IServiceCollection services)
+    {
+        services.AddScoped<CircuitHandler, IdleCircuitHandler>();
+        return services;
+    }
+}
+```
+
+Circuit activity handlers also provide a way to access scoped Blazor services from other non-Blazor dependency injection (DI) scopes, such as scopes created using <xref:System.Net.Http.IHttpClientFactory>. There's an existing pattern for accessing circuit-scoped services from other dependency injection scopes, but it requires using a custom base component type. With circuit activity handlers, we can use a better approach, which is demonstrated by the following example:
```
MackinnonBuck, line 327: > There's an existing pattern for accessing circuit-scoped services from other dependency injection scopes...

The docs for this pattern are [here](https://learn.microsoft.com/aspnet/core/blazor/fundamentals/dependency-injection?view=aspnetcore-7.0#access-blazor-services-from-a-different-di-scope). Given that we now have a better way of doing this, maybe we could modify the 8.0 version of those docs to point to this section?

MackinnonBuck: And we could probably go as far as removing that example entirely from the 8.0 docs

guardrex:
@MackinnonBuck ... I think we have a bit more to look at and resolve on this one.

Two things leap out now that I know what "there's an existing pattern for accessing circuit-scoped services" was referring to:

1. Although it's the same approach, these example approaches probably should be grouped with their subject matter and cross-linked ...

   * The SignalR topic has *Monitor circuit activity (Blazor Server)* with its `IdleCircuitHandler` example.
   * The DI topic has *Access Blazor services from a different DI scope* with its `CircuitServicesAccessor` example.
   * The Blazor Server *Additional scenarios* topic has *Access `AuthenticationStateProvider` in outgoing request middleware* with its `AuthenticationStateHandler` example.

1. The first two additions, SignalR and DI examples, seem straightforward. However, the third is something that I have less familiarity with because I spend much of my time in auth scenarios for the WASM hosting model and not implementing `DelegatingHandler` for Blazor Server apps. AFAICT, the `AuthenticationStateHandler` is for outgoing middleware request processing, as I have on the PR now, but I could be totally off base on that use case. If that is incorrect, what's the use case and what is the complete example for it? Even if my hunch is correct about the use case, note that I hacked a more complete example of it by making it create a custom header for the authenticated user's name. Whatever else we need to do with the overall layout, section naming, and section content, is this header addition a decent example for devs to see? I'm going for quick and simple but clear from there for other more advanced uses.

guardrex:
@MackinnonBuck ... I think you should take a look at the updated layout. I'm primarily concerned about item 2 above :point_up:. I may have guessed correctly on the implementation of the `AuthenticationStateHandler`, but I think u should confirm if I went the right way.

MackinnonBuck:
@guardrex I think the updated layout looks great. Regarding the `AuthenticationStateHandler` example, I wonder if we could get @javiercn's opinion on that. Javier, what do you think? Do we need a full example for this, or can we assume that if a customer needs this, they already have a specific scenario in mind?

guardrex:
Thanks @MackinnonBuck ... I like what I wrote (and the example) in the *Access `AuthenticationStateProvider` in outgoing request middleware* section. If it's good, I hope we can keep it. I think 🤔 it works ..... *maybe* 🤣 ... the 🙏 *server gods* 🙏 hopefully will smile on that coverage.

javiercn (approved):

guardrex:
Thanks @javiercn.

@danroth27 ... awaiting final sign-off.

Issue #28948, opened 2023-04-12
Discuss Kestrel vs IIS vs HTTP.sys


JamesNK:
## Explain why this topic is needed.

Offer a summary of the pros and cons of the various ASP.NET Core servers allow customers to make the best choice quickly.

## Suggest a location in the Table of Contents.

I think a section on the Kestrel overview page would be good. The section title could be "Recommended ASP.NET Core server" and discuss why Kestrel is recommended, but then also discuss scenarios where IIS or HTTP.sys is a better choice.

Alternatively, a new page under "Host and deploy".

## Write an abstract. In one **short** paragraph, describe what this topic will cover.

Kestrel is the recommended server because it's cross-platform, performance, configuration options, and extensibility. However, there are features that IIS and HTTP.sys offer that might make them a better choice. For example, process lifecycle management (IIS restarts the process on crash) or binding multiple apps to one port.

(this is a summary, there are more reasons)

Rick-Anderson:
#28937 contributed to this - right?
Or should this be closed?

JamesNK:
https://github.com/dotnet/AspNetCore.Docs/pull/28937 just discusses Kestrel. There isn't a place that compares the servers.

PR #29008, opened 2023-04-18
Move Servers to a top-level TOC section

### LINKED ISSUE #28988: Add a top-level Servers section in TOC and move Kestrel, IIS and HTTP.sys
Today Kestrel, IIS, and HTTP.sys docs sit in [Host and deploy](https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/?view=aspnetcore-7.0). I think a dedicated `Servers` section in ASP.NET Core docs would be better:

1. It's not obvious that server docs are in the [Host and deploy](https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/?view=aspnetcore-7.0) section. I think when devs see Host and deploy they think more about where they are hosting their app (OS platform, cloud vs local, container vs managed service vs bare metal), rather than the server choice. A `Servers` section would make it immediately obvious where to look.

2. [Host and deploy](https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/?view=aspnetcore-7.0) is a mix of random topics. VS publish profiles, web farms, Docker, Windows Service, Azure App Service, dev ops, plus documentation about servers. Moving the servers out helps focus this area on the hosting platform and how to deploy to it.

3. No page summarizes the different servers. We can add an `Overview` page if there is a dedicated `Servers` section. The overview says that Kestrel is the default in our template and recommended server. But then go into features that IIS and HTTP.sys provide that aren't in Kestrel.

tdykstra:
Fixes #28988 

>No page summarizes the different servers. We can add an `Overview` page if there is a dedicated `Servers` section. 

We have an overview of the servers in `Fundamentals` - I moved that to the `Overview` node of the new top-level `Servers` section. This overview article already [compares Kestrel to HTTP.sys](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/?view=aspnetcore-8.0&branch=pr-en-us-29008&tabs=windows#kestrel-vs-httpsys), and we can add comparison section(s) to address #28948.

[Internal review URL](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/?view=aspnetcore-8.0&branch=pr-en-us-29008&tabs=windows)

### DIFF
--- a/aspnetcore/toc.yml
+++ b/aspnetcore/toc.yml
@@ -250,8 +250,6 @@ items:
             uid: fundamentals/host/generic-host
           - name: Web Host
             uid: fundamentals/host/web-host
-      - name: Servers
-        uid: fundamentals/servers/index
       - name: Configuration
         uid: fundamentals/configuration/index
       - name: Options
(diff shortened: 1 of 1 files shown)


amcasey:
I don't really understand how we decided what gets to be a top-level node and I've never actually made a web app, so take this with a grain of salt.

I like the idea of grouping those pages together, but why would we make them top-level?  They don't seem like something an implementer would immediately need to understand?  If you're not on Windows, you don't have a choice anyway and if you are, the templates provide sensible defaults.  It seems like only Advanced users or people digging into Fundamentals would need more information about the different options and why you might choose one over another.

Maybe @JamesNK has more context on why this needed to be more discoverable?

JamesNK:
I think the current place (Host and deploy) is unexpected and unintuitive. Putting all this content under fundamentals doesn't seem right (although if we decide against top-level Servers in TOC, I think it would be an improvement). A new Server area seems the best solution to me.

re: not useful outside of Windows: Making the right server choice is the minority of content. There is lots of info that people need semi-regularly: Configuring Kestrel endpoints, TLS, troubleshooting, switching to IIS or HTTP.sys, etc. TLS and the dev cert is a common problem.

If we're worried about too many top-level items, I think "Code analysis" can be moved somewhere. Maybe "Code analysis" moves into Advanced? It doesn't seem important enough to be a featured.

JamesNK:
But I agree that we should discuss this and get a consensus before doing anything.

captainsafia:
My two cents on this: I'm not opposed to having more top-level nodes. I'd rather have a shallow content tree that makes it easier to discover topic areas as opposed to requiring users to parse a complex taxonomy to figure out where the info they need is.

I also suspect that shallower taxonomies play nicer with search engines, summarizers, etc.

Tratcher:
Love it!

Tratcher:
The overview page's Hosting Models section gives IIS specific details but isn't labeled as IIS specific. It should also cover self-hosting with Kestrel or Http.Sys. 'in-process` and 'out-of-process' should be prefixed by 'IIS'.

tdykstra:
We have 3 votes in favor of moving Servers to a top-level TOC node (@JamesNK, @Tratcher, @captainsafia).  @amcasey yours is the only "no" vote but the other comments came in after yours -- did their comments convince you that this is a good idea?

amcasey:
@tdykstra Don't block on me.  There is understandably limited appetite for rejiggering the whole tree and, in the absence of that, the location of one particular node isn't that important to me.

tdykstra:
That controls the landing page at https://learn.microsoft.com/en-us/aspnet/core/
To correspond to what we did in the TOC, I'll add a new Servers card.

PR #29106, opened 2023-04-27
Hashed routing to named elements

### LINKED ISSUE #29105: Hash routing to named element
* PU Issue: https://github.com/dotnet/aspnetcore/issues/8393
* PU PR: https://github.com/dotnet/aspnetcore/pull/47320

---
#### Document Details

* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #29105

Surayya ... I didn't guess about what general rules or *gotchas* 😈 that you want to call out. Let me know what you'd like to add in a comment, or you can `suggestion` them into the text.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/routing.md](https://github.com/dotnet/AspNetCore.Docs/blob/598a3718a3a34a773dc30e9a585d5276f5146fbb/aspnetcore/blazor/fundamentals/routing.md) | [ASP.NET Core Blazor routing and navigation](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?branch=pr-en-us-29106) |

### DIFF
--- a/aspnetcore/blazor/fundamentals/routing.md
+++ b/aspnetcore/blazor/fundamentals/routing.md
@@ -834,15 +834,29 @@ For more information on JavaScript isolation with JavaScript modules, see <xref:
 
 ## Hashed routing to named elements
 
-Navigate to a hashed (`#`) named element using the following approaches:
+Navigate to a named element using the following approaches with a hashed (`#`) reference to the element. Examples for each of the following approaches demonstrate navigation to an element with an `id` of `targetElement` in the `Counter` component:
 
-* Anchor element (`<a>`) with an `href`.
-* <xref:Microsoft.AspNetCore.Components.Routing.NavLink> component with an `href`.
-* <xref:Microsoft.AspNetCore.Components.NavigationManager.NavigateTo%2A?displayProperty=nameWithType> passing the URL.
+* Anchor element (`<a>`) with an `href`:
 
-The following example demonstrates hashed routing to named H2 headings.
+  ```razor
+  <a href="/counter#targetElement">
+  ```
 
-In the `Index` (`Pages/Index.razor`) and `Counter` (`Pages/Counter.razor`) components, place the following markup at the bottom of the existing component markup:
+* <xref:Microsoft.AspNetCore.Components.Routing.NavLink> component with an `href`:
+
+  ```razor
+  <NavLink href="counter#targetElement">
+  ```
+
+* <xref:Microsoft.AspNetCore.Components.NavigationManager.NavigateTo%2A?displayProperty=nameWithType> passing the relative URL:
+
+  ```csharp
+  Navigation.NavigateTo("counter#targetElement");
+  ```
+
+The following example demonstrates hashed routing to named H2 headings within a component and to external components.
+
+In the `Index` (`Pages/Index.razor`) and `Counter` (`Pages/Counter.razor`) components, place the following markup at the bottoms of the existing component markup to serve as navigation targets. The `<div>` creates artificial vertical space to demonstrate browser scrolling behavior:
 
 ```razor
 <div class="border border-info rounded bg-info" style="height:500px"></div>
@@ -851,7 +865,7 @@ In the `Index` (`Pages/Index.razor`) and `Counter` (`Pages/Counter.razor`) compo
 <p>Content!</p>
 ```
 
-Add the following `HashedRouting` component to the app:
+Add the following `HashedRouting` component to the app. Routes to elements within the component must use the route template of the component, which is `hashed-routing` in the following `HashedRouting` component. Routes to elements in external components use a root-relative path, which is `/` for the `Index` component and `/counter` for the `Counter` component.
 
 `Pages/HashedRouting.razor`:
 
--- a/aspnetcore/blazor/fundamentals/routing.md
+++ b/aspnetcore/blazor/fundamentals/routing.md
@@ -830,6 +830,83 @@ For more information on JavaScript isolation with JavaScript modules, see <xref:
 
 :::moniker-end
 
+:::moniker range=">= aspnetcore-8.0"
+
+## Hashed routing to named elements
+
+Navigate to a hashed (`#`) named element using the following approaches:
+
+* Anchor element with an `href`.
+* `Navlink` component with an `href`.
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/fundamentals/routing.md
```diff
@@ -830,6 +830,83 @@ For more information on JavaScript isolation with JavaScript modules, see <xref:
...
+@page "/hashed-routing"
+@inject NavigationManager Navigation
+
+<PageTitle>Hashed routing</PageTitle>
+
+<h1>Hashed routing to named elements</h1>
+
+<ul>
+    <li>
+        <a href="#targetElement">
```
surayya-MS, line 868: I changed this behavior recently because it was incorrect. Here it should be `href="/hashed-routing#targetElement"`. There is no scrolling to the element on the same page when clicking on `<a href="#..."></a>`. https://github.com/dotnet/aspnetcore/issues/47687

aspnetcore/blazor/fundamentals/routing.md
```diff
@@ -830,6 +830,83 @@ For more information on JavaScript isolation with JavaScript modules, see <xref:
...
+        <button @onclick="NavigateToElement">
+            Navigate with <code>NavigationManager</code> to the 
+            <code>Counter</code> component
+        </button>
+    </li>
+</ul>
+
+<div class="border border-info rounded bg-info" style="height:500px"></div>
+
+<h2 id="targetElement">Target H2 heading</h2>
```
surayya-MS, line 897: For some elements in the example let's use name instead of id to show that we support scrolling to the elements with id attribute and name attribute.

surayya-MS (approved):
Looks great!

aspnetcore/blazor/fundamentals/routing.md
guardrex:
```suggestion
        <a href="hashed-routing#targetElement">
```

aspnetcore/blazor/fundamentals/routing.md
```diff
@@ -830,6 +830,83 @@ For more information on JavaScript isolation with JavaScript modules, see <xref:
...
+            Anchor to the <code>Index</code> component
+        </a>
+    </li>
+    <li>
+        <a href="/counter#targetElement">
+            Anchor to the <code>Counter</code> component
+        </a>
+    </li>
+    <li>
+        <NavLink href="#targetElement">
```
surayya-MS, line 883:
```suggestion
        <NavLink href="/hashed-routing#targetElement">
```

aspnetcore/blazor/fundamentals/routing.md
surayya-MS: yes, like this

guardrex:
brb ... was pulled off on something else. 🏃

aspnetcore/blazor/fundamentals/routing.md
guardrex: Ok ... I'm *BACK!* 😄 

Named elements (`name` attribute) are depreciated under HTML5, which I'm certain that you know, and Blazor adopts the current standard HTML5 by doctype by default, which I'm also certain that you know. Therefore, mentioning this we ***should*** tell devs such is the case. I don't think we want to support their use in HTML5 ... or show it for that matter in examples. I admit two things about this ...

* Over the years as a professional dev, I really disliked it when orgs threw standards out the window. I swore a 🩸 *blood oath* 🩸 to the 🙏 *server gods* 🙏 that if I can ever stop organizations from doing it that I would.
* I ***have no power here*** 😆 to stop MS from publishing standards-breaking text or code.

Here's my recommendation if you want to mention it. I suggest in a NOTE to the effect ...

> \> \[!NOTE]
> \> Named elements for targeting anchor elements with the \`name\` attribute are \[depreciated as of HTML5](https://developer.mozilla.org/docs/Web/HTML/Element/a#deprecated_attributes), and Blazor adopts the current HTML specification by \[Doctype](https://developer.mozilla.org/docs/Glossary/Doctype) in \`wwwroot/index.html\` (Blazor WebAssembly) and \`Pages/_Host.cshtml\` (Blazor Server).
> \>
> \> Blazor scrolls to named targets in the same fashion that it scrolls to elements by \`id\`. However, we recommend using \`id\` to target elements in Blazor based on the current HTML specification.

Otherwise tho if you insist, please send code suggestion(s) for the lines that you want to update. I'm sure that I'll grind my teeth over it ... and my dentist 🦷👨‍⚕️ won't be pleased ... *but I'll live!* 🤣

surayya-MS: I actually didn't know that `name` attribute was deprecated. Thanks! In that case I may change our code to support only `id`. If the code will stay the same let's use the note to prefer id over name as you suggested. I'll give the final answer tomorrow.

surayya-MS: @guardrex, let's not provide example with name. I will remove support for name attribute.

guardrex:
@surayya-MS ... I feel we can flesh it out a bit more. See my last commit for what I think are improvements.

btw ...

Is the no leading forward slash for the self-links by design?

Does `<a href="/hashed-routing#targetElement">` (with a `/` on it) 💥 at runtime? If so, we might want to work that into the coverage.

surayya-MS:
> @surayya-MS ... I feel we can flesh it out a bit more. See my last commit for what I think are improvements.
> 
> btw ...
> 
> Is the no leading forward slash for the self-links by design?
> 
> Does `<a href="/hashed-routing#targetElement">` (with a `/` on it) 💥 at runtime? If so, we might want to work that into the coverage. If that doesn't break, I wonder if we should use it in the `href` cases for consistency with the `Index` and `Counter` component examples?

No, both "/hashed-routing#targetElement" and "hashed-routing#targetElement" will work.

guardrex:
I just realized that I did have the leading forward slash on the `NavigateTo` examples, at least in this topic. I'm going to note a tracking item to check the whole Blazor docs node and confirm consistent display later.

guardrex:
Ok ... this looks good. I'll merge it after it builds out. Thx for your help.

PR #29164, opened 2023-05-04
Easy test Problem /8

Rick-Anderson:
Contributes to #29136

### DIFF
--- a/aspnetcore/mvc/models/validation/samples/ValidationResultErrorMessage/ContactsController.cs
+++ b/aspnetcore/mvc/models/validation/samples/ValidationResultErrorMessage/ContactsController.cs
@@ -3,161 +3,163 @@ using Microsoft.EntityFrameworkCore;
 using ValidationResultErrorMessage.Data;
 using ValidationResultErrorMessage.Models;
 
-namespace ValidationResultErrorMessage;
-
-public class ContactsController : Controller
+namespace ValidationResultErrorMessage
 {
-    private readonly ValidationResultErrorMessageContext _context;
 
-    public ContactsController(ValidationResultErrorMessageContext context)
+    public class ContactsController : Controller
     {
-        _context = context;
-    }
+        private readonly ValidationResultErrorMessageContext _context;
 
-    // GET: Contacts
-    public async Task<IActionResult> Index()
-    {
-          return _context.Contact != null ? 
-                      View(await _context.Contact.ToListAsync()) :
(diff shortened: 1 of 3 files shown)


aspnetcore/mvc/models/validation/samples/ValidationResultErrorMessage/ContactsController.cs
```diff
@@ -1,164 +1,163 @@
...
-using System.Linq;
-using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc;
-using Microsoft.AspNetCore.Mvc.Rendering;
 using Microsoft.EntityFrameworkCore;
 using ValidationResultErrorMessage.Data;
 using ValidationResultErrorMessage.Models;
 
-namespace ValidationResultErrorMessage
+namespace ValidationResultErrorMessage;
```
Rick-Anderson: We should do this on any code we write.
But it totally messes up GitHub diff.

aspnetcore/mvc/models/validation/samples/ValidationResultErrorMessage/ContactsController.cs
```diff
@@ -151,14 +146,18 @@ public async Task<IActionResult> DeleteConfirmed(Guid id)
             {
                 _context.Contact.Remove(contact);
             }
-            
+            else
+            {
+                return Problem("Contact deleted.");
+            }
+
```
Rick-Anderson, line 153: I added this to make it easy to test. Click open in new tab on delete button. Then delete the item. In new tab, try delete again.

Rick-Anderson (commented):
This is a really great sample, but I'm wonder if we should convert it to a Web API app. You really don't want a UI app returning

{"type":"https://tools.ietf.org/html/rfc7231#section-6.6.1","title":"An error occurred while processing your request.","status":500,"detail":"Contact deleted.","traceId":"00-bfdefd61ac9c6f6548bc2d999ea2b685-28769c063ba71723-00"}

What do you think @sammychinedu2ky

aspnetcore/mvc/models/validation/samples/ValidationResultErrorMessage/ContactsController.cs
sammychinedu2ky: ok let me see

sammychinedu2ky: I could return the Error View while serializing the JSON into the Error model.
image
@Rick-Anderson

PR #29232, opened 2023-05-10
Add public to class Todo

sguitardude:
When adding public class TodoItemDTO, this message may display in Visual Studio: Inconsistent accessibility: parameter type: 'Todo' is less accessible than the method 'TodoItemDTO.TodoItemDTO (todo).

Adding public to class makes this message go away.


tdykstra (approved):
Thanks, @sguitardude !

PR #29240, opened 2023-05-11
Update loadbalancing.md

JamesNK:
Fix example. Port is constant. IP address changes.


wadepickett (approved):
Looks good James, thanks.  Nothing to change, but I did update the ms.date metadata.

PR #29249, opened 2023-05-12
Update Program.cs

nockscitney:
Formatting Update

Updated the formatting when adding the MapAreaControllerRoute with the name of MyAreaProducts to bring it inline with the other MapAreaControllerRoute declarations


nockscitney:
@dotnet-policy-service agree

PR #29264, opened 2023-05-14
Update enforcing-ssl.md

postmeback:
Updated the text so that people understand it better


guardrex:
Hello @postmeback ... All three end up creating a policy file with that filename, so let's just add that filename to the line preceding the list.

guardrex (approved):

senglory:
I'd also add text with advise to create the 'distribution' folder manually if it hasn't been created by setup earlier.

postmeback:
Thanks @guardrex @Rick-Anderson  😃

PR #29294, opened 2023-05-17
Remove 'unsafe-eval' in CSP guidance

### LINKED ISSUE #29293: Remove 'unsafe-eval' in CSP guidance
* https://github.com/dotnet/aspnetcore/issues/37787
* https://github.com/dotnet/aspnetcore/pull/48182

---
#### Document Details

* ID: 88177a7d-c9ff-c245-5629-a462c9258abf
* Version Independent ID: 6e0b5c52-90a1-5ca6-bfad-df33a8beae6c
* Content: [Enforce a Content Security Policy for ASP.NET Core Blazor](https://learn.microsoft.com/en-us/aspnet/core/blazor/security/content-security-policy?view=aspnetcore-7.0)
* Content Source: [aspnetcore/blazor/security/content-security-policy.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/security/content-security-policy.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #29293

Mackinnon ... I see that Pavel is still working on runtime bits, but it is ok to knock this out for preview coverage now?

It's not easy to see from the diff because it's adding versioned content. The changes here are to ...

* Remove the following line ...

  > \* Specify \`unsafe-eval\` to permit the Blazor WebAssembly Mono runtime to function.

* Change part of the WASM policy from ...

  > ```html
  > script-src 'self'
  >            'unsafe-eval';
  > ```

  ... to ...

  > ```html
  > script-src 'self';
  > ```

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/security/content-security-policy.md](https://github.com/dotnet/AspNetCore.Docs/blob/810b3158f9a8eb750ddc092b88131a7697d04bc6/aspnetcore/blazor/security/content-security-policy.md) | [Enforce a Content Security Policy for ASP.NET Core Blazor](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/content-security-policy?branch=pr-en-us-29294) |

### DIFF
--- a/aspnetcore/blazor/security/content-security-policy.md
+++ b/aspnetcore/blazor/security/content-security-policy.md
@@ -39,7 +39,9 @@ Minimally, specify the following directives and sources for Blazor apps. Add add
 * [object-src](https://developer.mozilla.org/docs/Web/HTTP/Headers/Content-Security-Policy/object-src): Indicates valid sources for the `<object>`, `<embed>`, and `<applet>` tags. Specify `none` to prevent all URL sources.
 * [script-src](https://developer.mozilla.org/docs/Web/HTTP/Headers/Content-Security-Policy/script-src): Indicates valid sources for scripts.
   * Specify `self` to indicate that the app's origin, including the scheme and port number, is a valid source.
-  * In a Blazor WebAssembly app, specify any additional hashes to permit your required *non-framework scripts* to load.
+  * In a Blazor WebAssembly app:
+    * Specify `wasm-unsafe-eval` to permit the Blazor WebAssembly Mono runtime to function.
+    * Specify any additional hashes to permit your required *non-framework scripts* to load.
   * In a Blazor Server app, specify hashes to permit required scripts to load.
 * [style-src](https://developer.mozilla.org/docs/Web/HTTP/Headers/Content-Security-Policy/style-src): Indicates valid sources for stylesheets.
   * Specify `self` to indicate that the app's origin, including the scheme and port number, is a valid source.
@@ -147,7 +149,8 @@ In the `<head>` content of the `wwwroot/index.html` host page, apply the directi
                default-src 'self';
                img-src data: https:;
                object-src 'none';
-               script-src 'self';
+               script-src 'self'
+                          'wasm-unsafe-eval';
                style-src 'self';
                upgrade-insecure-requests;">
 ```
--- a/aspnetcore/blazor/security/content-security-policy.md
+++ b/aspnetcore/blazor/security/content-security-policy.md
@@ -29,7 +29,26 @@ CSP is supported in most modern desktop and mobile browsers, including Chrome, E
 
 Minimally, specify the following directives and sources for Blazor apps. Add additional directives and sources as needed. The following directives are used in the *Apply the policy* section of this article, where example security policies for Blazor WebAssembly and Blazor Server are provided:
 
-:::moniker range=">= aspnetcore-7.0"
+:::moniker range=">= aspnetcore-8.0"
+
+* [base-uri](https://developer.mozilla.org/docs/Web/HTTP/Headers/Content-Security-Policy/base-uri): Restricts the URLs for a page's `<base>` tag. Specify `self` to indicate that the app's origin, including the scheme and port number, is a valid source.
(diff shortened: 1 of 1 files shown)


MackinnonBuck:
@guardrex While 'unsafe-eval' is no longer required, 'wasm-unsafe-eval' is required instead (because WebAssembly execution is disabled without it). See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#unsafe_webassembly_execution.

guardrex:
@MackinnonBuck ... Ok ... I put that on the last commit.

damienbod:
@MackinnonBuck  unsafe-eval is still required due to the blazor js script. I get errors when I test this with the latest preview

Did I miss something? 

Greetings Damien

guardrex:
@damienbod ... I recommend posting the rendered `meta` tags built by Lock's API and to show the errors.

damienbod:
Here's the error

```
Error:

browserLink:153 Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU='), or a nonce ('nonce-...') is required to enable inline execution.
```

I can't see any difference in the CSP which is strange.

damienbod:
Error:

blazor.webassembly.js:1 EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' 'sha256-v8v3RKRPmN4odZ1CWM5gw80QKPCCWMcpNeOmimNL2AA=' 'wasm-unsafe-eval'".

    at Function (<anonymous>)
    at dotnet.8.0.0-preview.3.23174.8.b7g991eq1o.js:3:165575
    at sc (dotnet.8.0.0-preview.3.23174.8.b7g991eq1o.js:3:165596)
    at dotnet.8.0.0-preview.3.23174.8.b7g991eq1o.js:3:168598
    at ic (dotnet.8.0.0-preview.3.23174.8.b7g991eq1o.js:3:169205)
    at dotnet.8.0.0-preview.3.23174.8.b7g991eq1o.js:3:164662
    at Object.ec [as call_assembly_entry_point] (dotnet.8.0.0-preview.3.23174.8.b7g991eq1o.js:3:165046)
    at Object.callEntryPoint (blazor.webassembly.js:1:42926)
    at qt (blazor.webassembly.js:1:59086)

guardrex:
When we used to have hashes in the topic (6.0 or earlier), I would update them on release day because they're potentially being churned by the PU all the way up to final release. That wasn't a problem at 7.0 because no hashes were required for the CSP.

damienbod:
@guardrex  The one which fails uses the 'wasm-unsafe-eval' and the error is about the missing 'unsafe-eval'

guardrex:
> the error is about the missing 'unsafe-eval'

... but perhaps that's only because it needs the hash for the script that the error mentions.

What happens if you use the one from the error message ...

```csharp
builder.AddScriptSrc()
    .Self()
    .WithHash256("sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=")
    .WasmUnsafeEval();
```

guardrex:
... and btw ... looks like you aren't on latest ...

> preview.3

Update to Pre4.

damienbod:
@guardrex  latest version does not start

https://github.com/dotnet/aspnetcore/issues/48298

yes, this is probably just a WIP problem and will probably be fixed in the next preview release. I just wanted to mention it so that it's visible. Released version works without problem. I will test with the other hash as well. 

Greetings Damien

damienbod:
Was doing some testing again and this does not work in .NET 7 either, the 'unsafe-eval' is still required

MackinnonBuck:
@damienbod The 'unsafe-eval' requirement will still be required for .NET 7 and earlier. Also, please note that the updates enabling improved CSP compliance aren't available in a public preview release yet. You could try installing a nightly build from https://github.com/dotnet/installer to test this new improvement. Thanks!

PR #29325, opened 2023-05-20
Improve .NET 7 JWT Bearer appsettings configuration doco

### LINKED ISSUE #29307: "Authentication and authorization in minimal APIs" documentation unclear configuration options
I was testing out setting up JWT bearer auth for a minimal .NET 7 API without using my own options/configuration section by trying to follow the documentation on this page: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/security?view=aspnetcore-7.0#configuring-authentication-strategy

I ended up getting the following validation failure on the bearer token when calling an endpoint `Bearer was not authenticated. Failure message: IDX10501: Signature validation failed. Unable to match key`. 
The issue in this case was the `Authority` not being set on the options (the error message isn't particularly helpful).

I ended up having to look at the source code for the [JwtBearerConfigureOptions](https://github.com/dotnet/aspnetcore/blob/c5a9c5973bf735f1bb9d611cd50c796f6c32557e/src/Security/Authentication/JwtBearer/src/JwtBearerConfigureOptions.cs#L56) class to find out how I could configure this setting.

I think the documentation sample should be updated to include/reference this options class (or all the properties loaded from config into it) so developers know what values can be defined in configuration.

Furthermore, the example on that minimal API page should probably be updated to define the `Authority` property, as that's the most common configuration used when setting up JWT bearer auth to my understanding. 

When using a user defined custom config section/option, the auth configuration usually looks something like this:

```c#
builder.Services.AddAuthentication(options =>
        {
            options.DefaultAuthenticateScheme = JwtBearerDefaults.AuthenticationScheme;
            options.DefaultChallengeScheme = JwtBearerDefaults.AuthenticationScheme;
        }).AddJwtBearer(options =>
        {
            var jwtOptions = builder.Configuration.GetRequiredSection("MyJwtSettings").Get<MyJwtOptions>();
            ArgumentNullException.ThrowIfNull(jwtOptions);
            options.Authority = jwtOptions.Issuer;
            options.Audience = jwtOptions.Audience;
        });
```

If the framework provided option is configured correctly in appsettings you do not need the above and can simply do
`builder.Services.AddAuthentication().AddJwtBearer();`

---
#### Document Details

* ID: 3a9d7eb8-6c1f-4619-00fa-9b69dbe3dcea
* Version Independent ID: 3a9d7eb8-6c1f-4619-00fa-9b69dbe3dcea
* Content: [Authentication and authorization in minimal APIs](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/security?view=aspnetcore-7.0#configuring-authentication-strategy)
* Content Source: [aspnetcore/fundamentals/minimal-apis/security.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/minimal-apis/security.md)
* Product: **aspnet-core**
* GitHub Login: @captainsafia
* Microsoft Alias: **safia**

laurencee:
Fixes #29307

In real world usage, you usually have a token-issuing server. To validate tokens, you must specify the `Authority` property in the authentication scheme, so instructions for how to do this has been added to the documentation.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/fundamentals/minimal-apis/security.md](https://github.com/dotnet/AspNetCore.Docs/blob/9ded4699751578ddaaa0e0d5d1c9387fdc9b79cd/aspnetcore/fundamentals/minimal-apis/security.md) | [Authentication and authorization in minimal APIs](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/security?branch=pr-en-us-29325) |

### DIFF
--- a/aspnetcore/fundamentals/minimal-apis/security.md
+++ b/aspnetcore/fundamentals/minimal-apis/security.md
@@ -76,6 +76,12 @@ The ASP.NET Core framework expects to find these options under the `Authenticati
 }
 ```
 
+For standard JWT Bearer authentication in a real application you may need to specify the `Authentication:Schemes:{SchemeName}:Authority` property which matches the address of the token-issuing authentication server. 
+
+In the case that you do not have access to the token-issuing server, the `Authority` property should **not be set** and instead you can specify the signing key as a base64 encoded string under the `Bearer` scheme `SigningKeys:{SigningKeyName}:Value` with the `SigningKeys:{SigningKeyName}:Issuer` matching the `ValidIssuer` specified in the authentication scheme configuration.
+
+For a full list of available configuration options for JWT bearer based authentication, see the [JwtBearerConfigureOptions](https://github.com/dotnet/aspnetcore/blob/c5a9c5973bf735f1bb9d611cd50c796f6c32557e/src/Security/Authentication/JwtBearer/src/JwtBearerConfigureOptions.cs#L56) class.
+
 In `Program.cs`, two JWT bearer-based authentication strategies are registered, with the:
 
 * "Bearer" scheme name.
--- a/aspnetcore/fundamentals/minimal-apis/security.md
+++ b/aspnetcore/fundamentals/minimal-apis/security.md
@@ -76,11 +76,13 @@ The ASP.NET Core framework expects to find these options under the `Authenticati
 }
 ```
 
-For standard JWT Bearer authentication in a real application you may need to specify the `Authentication:Schemes:{SchemeName}:Authority` property which matches the address of the token-issuing authentication server. 
+For JWT Bearer authentication in a production app, you may need to specify the `Authentication:Schemes:{SchemeName}:Authority` property, which matches the address of the token-issuing authentication server. The `{SchemeName}` placeholder in the preceding property is the scheme name.
 
-In the case that you do not have access to the token-issuing server, the `Authority` property should **not be set** and instead you can specify the signing key as a base64 encoded string under the `Bearer` scheme `SigningKeys:{SigningKeyName}:Value` with the `SigningKeys:{SigningKeyName}:Issuer` matching the `ValidIssuer` specified in the authentication scheme configuration.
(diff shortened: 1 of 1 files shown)


laurencee:
@dotnet-policy-service agree

guardrex (approved):
Thanks @laurencee! :rocket:

We have an INCLUDE that explains how to change the release branch for reference source `main` branch references. I don't recommend a line number ... the product code churns quite a bit making them rather unstable and prone to break 😈.

captainsafia (changes requested):
The JSON-based configuration option are design to be used in conjunction with the `user-jwts` CLI tool. I would not recommend using them at all for production environments. You should be able to configure this properties on the `AddJwtBearer` method invocation.

laurencee:
@captainsafia That's quite a shame if that's the case. 😞  

From my reading, there's nothing in this current documentation that indicates this appsettings configuration approach **shouldn't** be used for production purpose, in fact it reads as the opposite to me right now which is why I posted the issue and this PR in the first place as it looks like an official guide on "how to do authentication with minimal apis". 

So, to make the documentation clear, there should be a pretty big disclaimer that this configuration approach is not for production purposes?

For me, this configuration approach would be the preferred way to configure these standard forms of authentication, so it's quite a shame if the only current purpose for this is demo/sample code. It would have trimmed down unnecessary duplication across numerous projects for setting JWT Bearer options and Authentication options in app service configuration code (creating a custom options class and hooking that to the method invocations).

Out of interest, what would you say is the main reason this **shouldn't** be used in a production setting? I don't see anything obvious in the way the JWT Bearer options are wired up that connects it to the tooling when looking at your [JwtBearerConfigureOptions ](https://github.com/dotnet/aspnetcore/blob/c5a9c5973bf735f1bb9d611cd50c796f6c32557e/src/Security/Authentication/JwtBearer/src/JwtBearerConfigureOptions.cs#L56) implementation. 

I tested it out myself in a test environment and the auth flow all worked as I expected with a 3rd party JWT provider.

captainsafia:
@laurencee So, in .NET 7, we introduced the notion of "config-based authentication options" for some authentication models (see [this issue](https://github.com/dotnet/aspnetcore/issues/42170)).

In development scenarios, the the framework already handles populating the signing key information correctly (when used in conjunction with the `dotnet user-jwts` tool) so the recommendation to include this configuration manually is incorrect.

It's acceptable to use the same config-based strategy in production apps, although I don't think here is the right place to document those configuration options. Technically, anything that you would put in your JwtBearerOptions you can pass through configuration in your appsettings (or likely a more secure configuration strategy when in production).

PR #29440, opened 2023-06-06
Updated Cors middleware pages

poornimanayar:
Updated the docs for including UseCors after UseRouting but before UseAuthorization


poornimanayar:
I hope everything is okay with this PR?

PR #29555, opened 2023-06-18
Update rate-limit.md

VahidFarahmandian:
The description provided for the Token Bucket Rate Limiter is ambiguous, so I clarified it a little a bit more.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/performance/rate-limit.md](https://github.com/dotnet/AspNetCore.Docs/blob/3a001fa40782c07b80a40dee675b9a0f3ed617ad/aspnetcore/performance/rate-limit.md) | [Rate limiting middleware in ASP.NET Core](https://review.learn.microsoft.com/en-us/aspnet/core/performance/rate-limit?branch=pr-en-us-29555) |

### DIFF
--- a/aspnetcore/performance/rate-limit.md
+++ b/aspnetcore/performance/rate-limit.md
@@ -103,7 +103,7 @@ The following code uses the sliding window rate limiter:
 
 ### Token bucket limiter
 
-The token bucket limiter is similar to the sliding window limiter, but rather than adding back the requests taken from the expired segment, a fixed number of tokens are added each replenishment period. The tokens added each segment can't increase the available tokens to a number higher than the token bucket limit. The following table shows a token bucket limiter with a limit of 100 tokens and a 10-second replenishment period:
+The token bucket limiter is similar to the sliding window limiter, but rather than adding back the requests taken from the expired segment, a fixed number of tokens are added each replenishment period. The tokens added to each segment can't increase the available tokens to a number higher than the token bucket limit. When all the available tokens are consumed, then no other incoming request is allowed to access the resource, until the replenishment period reached and some extra tokens(maximum of `TokensPerPeriod`) added to the bucket. It is important to note that if in each replenishment period, no request comes in, then the available tokens count will be increased until it reach the `TokenLimit`. The following table shows a token bucket limiter with a limit of 100 tokens, a 10-second replenishment period and the ability to add maximum of 20 tokens per replenishment period:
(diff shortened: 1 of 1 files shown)


aspnetcore/performance/rate-limit.md
```diff
@@ -103,7 +103,7 @@ The following code uses the sliding window rate limiter:
...
+The token bucket limiter is similar to the sliding window limiter, but rather than adding back the requests taken from the expired segment, a fixed number of tokens are added each replenishment period. The tokens added to each segment can't increase the available tokens to a number higher than the token bucket limit. When all the available tokens are consumed, then no other incoming request is allowed to access the resource, until the replenishment period reached and some extra tokens(maximum of `TokensPerPeriod`) added to the bucket. It is important to note that if in each replenishment period, no request comes in, then the available tokens count will be increased until it reach the `TokenLimit`. The following table shows a token bucket limiter with a limit of 100 tokens, a 10-second replenishment period and the ability to add maximum of 20 tokens per replenishment period:
```
Rick-Anderson, line 106:
```suggestion
The token bucket limiter is similar to the sliding window limiter, but rather than adding back the requests taken from the expired segment, a fixed number of tokens are added each replenishment period. The tokens added to each segment can't increase the available tokens to a number higher than the token bucket limit. When all the available tokens are consumed, then no other incoming request is allowed to access the resource, until the replenishment period reached and some extra tokens(maximum of `TokensPerPeriod`) added to the bucket. It is important to note that if in each replenishment period, no request comes in, then the available tokens count will be increased until it reach the `TokenLimit`. The following table shows a token bucket limiter with a limit of 100 tokens, a 10-second replenishment period and the ability to add maximum of 20 tokens per replenishment period:
```

Rick-Anderson: What's ambiguous. We don't like to get too wordy.

VahidFarahmandian: for the given table, some foundation data (such as token per replenishment) is not presented and this causes the readers to become confused. in this clarification I tried to clarify the role of each specific property. what happen when replenishment period reached? how much token is allowed to be added?

Rick-Anderson: You can take my suggestion or I'll close the PR. Thanks for your contribution.

PR #29590, opened 2023-06-21
Update middleware.md

brockallen:
It might be helpful to mention that if these are already added, then they are not added automatically. It seems this is the case, much in the same way UseRouting is done, according to this code: https://github.com/dotnet/aspnetcore/blob/e1c4295531b5529aa1462ae5ea620f04a099f54d/src/DefaultBuilder/src/WebApplicationBuilder.cs#L371 and this: https://github.com/dotnet/aspnetcore/blob/e1c4295531b5529aa1462ae5ea620f04a099f54d/src/DefaultBuilder/src/WebApplicationBuilder.cs#L382.


Rick-Anderson (approved):

Rick-Anderson:
@brockallen  thanks for the PR

PR #29564, opened 2023-06-20
Improve Kestrel endpoints

JamesNK:
Based on https://github.com/dotnet/AspNetCore.Docs/pull/29501
Fixes #21117 

[The issue](https://github.com/dotnet/AspNetCore.Docs/issues/21117) initially proposed grouping topics by ways to configure (e.g. code, appsettings.json). Then under each way to configure, there would be subjects (e.g. HTTPS, HTTP protocol)

That layout didn't work well because subjects couldn't be introduced once (e.g. here are all the HTTP protocols and what is needed to support them) because the instructions to configure the subject was in 2-3 different places. The layout is now by subject, with an introduction to that subject, then ways to configure it as subheadings.

I've liberally added links between different sections, so people can jump directly to the content they want.

**Use this link to preview changes:** https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/endpoints?view=aspnetcore-8.0&branch=pr-en-us-29564

For reference, here is the current page: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/endpoints?view=aspnetcore-7.0

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/fundamentals/servers/kestrel/connection-middleware.md](https://github.com/dotnet/AspNetCore.Docs/blob/580e087162762e67791dcd0a7ddaf6353a5a8e50/aspnetcore/fundamentals/servers/kestrel/connection-middleware.md) | [aspnetcore/fundamentals/servers/kestrel/connection-middleware](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/connection-middleware?branch=pr-en-us-29564) |
| [aspnetcore/fundamentals/servers/kestrel/endpoints.md](https://github.com/dotnet/AspNetCore.Docs/blob/580e087162762e67791dcd0a7ddaf6353a5a8e50/aspnetcore/fundamentals/servers/kestrel/endpoints.md) | [Configure endpoints for the ASP.NET Core Kestrel web server](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/endpoints?branch=pr-en-us-29564) |

PR #29605, opened 2023-06-22
Kestrel endpoints follow up changes

JamesNK:
Feedback from https://github.com/dotnet/AspNetCore.Docs/pull/29564

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/fundamentals/servers/kestrel/endpoints.md](https://github.com/dotnet/AspNetCore.Docs/blob/b6268b3677b7b172dac8d6e8dca615b2bc7d5d81/aspnetcore/fundamentals/servers/kestrel/endpoints.md) | [Configure endpoints for the ASP.NET Core Kestrel web server](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/endpoints?branch=pr-en-us-29605) |

### DIFF
--- a/aspnetcore/fundamentals/servers/kestrel/endpoints.md
+++ b/aspnetcore/fundamentals/servers/kestrel/endpoints.md
@@ -21,7 +21,7 @@ Kestrel endpoints provide the infrastructure for listening to incoming requests 
 
 * The address specifies the network interface that the server listens on for incoming requests, such as a TCP port.
 * The protocol specifies the communication between the client and server, such as HTTP/1.1, HTTP/2, or HTTP/3.
-* An endpoint can be secured with HTTPS using the `HTTPS` scheme or `UseHttps` method.
+* An endpoint can be secured with HTTPS using the `https` scheme or `UseHttps` method.
 
 Endpoints can be configured using URLs, JSON in `appsettings.json`, and code. This article discusses how to use each option to configure an endpoint:
 
--- a/aspnetcore/fundamentals/servers/kestrel/endpoints.md
+++ b/aspnetcore/fundamentals/servers/kestrel/endpoints.md
@@ -21,7 +21,7 @@ Kestrel endpoints provide the infrastructure for listening to incoming requests 
 
 * The address specifies the network interface that the server listens on for incoming requests, such as a TCP port.
 * The protocol specifies the communication between the client and server, such as HTTP/1.1, HTTP/2, or HTTP/3.
-* An endpoint can be secured with HTTPS using the `https` scheme or `UseHttps` method.
+* An endpoint can be secured using the `https` URL scheme or `UseHttps` method.
 
 Endpoints can be configured using URLs, JSON in `appsettings.json`, and code. This article discusses how to use each option to configure an endpoint:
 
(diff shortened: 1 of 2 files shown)


aspnetcore/fundamentals/servers/kestrel/endpoints.md
```diff
@@ -20,7 +20,8 @@ uid: fundamentals/servers/kestrel/endpoints
 Kestrel endpoints provide the infrastructure for listening to incoming requests and routing them to the appropriate middleware. The combination of an address and a protocol defines an endpoint.
 
 * The address specifies the network interface that the server listens on for incoming requests, such as a TCP port.
-* The protocol defines the communication standard between the client and server, such as whether the endpoint is secured with HTTPS, and the HTTP version.
+* The scheme specifies if the endpoint is secured with HTTPS.
+* The protocol specifies the communication standard between the client and server, such as HTTP/1.1, HTTP/2, or HTTP/3.
```
JamesNK, line 24: Is scheme the right word? If you're configuring endpoints with code you'd call `UseHttps`. It's only URLs and config that set https via a scheme.

tdykstra: In some places, such as [here]( https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/endpoints?view=aspnetcore-8.0&branch=pr-en-us-29605#url-formats), we imply that protocol means HTTP vs. HTTPS. In other places it’s HTTP/1 vs. HTTP/2, etc., or TLS1.1 vs. TLS1.2 etc. But the introduction to this bulleted list uses "protocol" loosely to carry all these meanings. So would it also be more accurate to replace “The combination of an address and a protocol defines an endpoint.” with “The combination of an address and a scheme defines an endpoint.”? Or "The combination of an address, a scheme, and one or two protocols defines an endpoint."?

Tratcher: The uri prefix (scheme, address, port) define an endpoint. Endpoints support one or more HTTP protocol versions.

Tratcher: I especially wanted to split them out because the next list does too.

Tratcher: > Is scheme the right word? If you're configuring endpoints with code you'd call `UseHttps`. It's only URLs and config that set https via a scheme.

It's the closest we've got that is distinguishable from http protocol versions.

Tratcher: Hmm, maybe your right, the address and the port really do define the endpoint. An endpoint may have encryption enabled via scheme or UseHttps, and it will support one or more HTTP protocol versions.

JamesNK: I updated the bullet point to say:

> An endpoint can be secured with HTTPS using the `HTTPS` scheme or `UseHttps` method.

What do you think?

Tratcher: Maybe "`https` url scheme"?

Tratcher (approved):

aspnetcore/fundamentals/servers/kestrel/endpoints.md
```diff
@@ -20,7 +20,8 @@ uid: fundamentals/servers/kestrel/endpoints
 Kestrel endpoints provide the infrastructure for listening to incoming requests and routing them to the appropriate middleware. The combination of an address and a protocol defines an endpoint.
 
 * The address specifies the network interface that the server listens on for incoming requests, such as a TCP port.
-* The protocol defines the communication standard between the client and server, such as whether the endpoint is secured with HTTPS, and the HTTP version.
+* The protocol specifies the communication between the client and server, such as HTTP/1.1, HTTP/2, or HTTP/3.
+* An endpoint can be secured with HTTPS using the `https` scheme or `UseHttps` method.
```
JamesNK, line 24:
```suggestion
* An endpoint can be secured using the `https` URL scheme or the `UseHttps` method.
```

PR #29624, opened 2023-06-23
Added comment about handling of absence of query string values and bi…

mitchdenny:
Addresses: https://github.com/dotnet/aspnetcore/issues/45956


JamesNK:
This change is only in .NET 7 docs. Should it be present in .NET 8 docs as well?

mitchdenny:
Done.

PR #29705, opened 2023-07-05
SupplyParameterFromQuery 8.0 updates

### LINKED ISSUE #29547: Update `[SupplyParameterFromQuery]` guidance and examples
Starting in .NET 8 Preview 6, the `[SupplyParameterFromQuery]` attribute no longer has to be used with the `[Parameter]` attribute. Furthermore, components other than page components can read query parameters using `[SupplyParameterFromQuery]`.

---
#### Document Details

* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #29547

Mackinnon ...

* We never documented `[SupplyParameterFromForm]` for any release. I thought it was an internal framework API bit that devs wouldn't directly use, but I'm prepared to hear what we need to cover it if that's not the case. 👂
* This is the only spot in the Blazor docs that we cover the `[SupplyParameterFromQuery]` attribute. If there are no other places around the Blazor docs that need to refer to the attribute, then we'll be good to go after we get this spot updated.
* Although Pre6 isn't out yet, I'm going to merge this after you sign off. I'm already working on the major updates for full-stack Blazor, and I don't want to get bogged down in a 😈 ***nasty merge conflict*** 😈 situation 😆.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/routing.md](https://github.com/dotnet/AspNetCore.Docs/blob/5a758cd2e62f42e4143a1965193f23f0c6814c72/aspnetcore/blazor/fundamentals/routing.md) | [ASP.NET Core Blazor routing and navigation](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?branch=pr-en-us-29705) |

### DIFF
--- a/aspnetcore/blazor/fundamentals/routing.md
+++ b/aspnetcore/blazor/fundamentals/routing.md
@@ -571,14 +571,24 @@ For more information on obtaining the state associated with the target history e
 
 ## Query strings
 
-:::moniker range=">= aspnetcore-6.0"
+:::moniker range=">= aspnetcore-8.0"
 
-Use the [`[SupplyParameterFromQuery]` attribute](xref:Microsoft.AspNetCore.Components.SupplyParameterFromQueryAttribute) with the [`[Parameter]` attribute](xref:Microsoft.AspNetCore.Components.ParameterAttribute) to specify that a component parameter of a *routable* component can come from the query string.
+Use the [`[SupplyParameterFromQuery]` attribute](xref:Microsoft.AspNetCore.Components.SupplyParameterFromQueryAttribute) to specify that a component parameter comes from the query string.
+
+:::moniker-end
+
+:::moniker range=">= aspnetcore-6.0 < aspnetcore-8.0"
+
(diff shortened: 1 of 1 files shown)


MackinnonBuck (approved):
This looks great! I think this is still the only spot where we need to document `[SupplyParameterFromQuery]`.

Regarding `[SupplyParameterFromForm]`, the reason we haven't documented it for previous releases is that it's a very new feature 🙂 Its use is somewhat similar to that of `[SupplyParameterFromQuery]`, but it binds parameters to form request entries rather than query parameters. I would suggest content about `[SupplyParameterFromForm]` should be added to the "Forms and Input Components" document rather than the routing docs, where `[SupplyParameterFromQuery]` is documented.

Issues describing the design of this feature can be found [here](https://github.com/dotnet/aspnetcore/issues/46688) and [here](https://github.com/dotnet/aspnetcore/issues/48759). Some usage examples can be found in our [test components](https://github.com/dotnet/aspnetcore/tree/main/src/Components/test/testassets/Components.TestServer/RazorComponents/Pages/Forms). If this isn't the right info (or not enough info) for you to get started preparing content, please let me know!

CC @javiercn in case you have anything to add.

PR #29749, opened 2023-07-10
Simple Update for state management

Runaho:
In my own tests, using OnChange += StateHasChanged; unfortunately did not work.

Based on this, I did some research and came to the conclusion that when we write our own handler method, we can solve this problem by invoking asynchronously.

If we can update the document in this way, it will be better for people like me who come across this problem to learn directly with the already solved version.

Please write if you have any suggestions.
I would like to point out that I did not open an issue because I thought it was a simple suggestion.

### DIFF
--- a/aspnetcore/blazor/includes/state-container.md
+++ b/aspnetcore/blazor/includes/state-container.md
@@ -66,7 +66,7 @@ services.AddScoped<StateContainer>();
 @code {
     protected override void OnInitialized()
     {
-        StateContainer.OnChange += StateHasChanged;
+        StateContainer.OnChange += OnMyChangeHandler;
     }
 
     private void ChangePropertyValue()
@@ -77,7 +77,14 @@ services.AddScoped<StateContainer>();
 
     public void Dispose()
     {
-        StateContainer.OnChange -= StateHasChanged;
+        StateContainer.OnChange -= OnMyChangeHandler;
+    }
+
+    private async void OnMyChangeHandler()
+    {
+      await InvokeAsync(() => {
+          StateHasChanged();
+      });
     }
 }
 ```
@@ -104,7 +111,7 @@ services.AddScoped<StateContainer>();
 @code {
     protected override void OnInitialized()
     {
-        StateContainer.OnChange += StateHasChanged;
+        StateContainer.OnChange += OnMyChangeHandler;
     }
 
     private void ChangePropertyValue()
@@ -115,7 +122,14 @@ services.AddScoped<StateContainer>();
 
     public void Dispose()
     {
-        StateContainer.OnChange -= StateHasChanged;
+        StateContainer.OnChange -= OnMyChangeHandler;
+    }
+
+    private async void OnMyChangeHandler()
+    {
+      await InvokeAsync(() => {
+          StateHasChanged();
+      });
     }
 }
 ```


Runaho:
@dotnet-policy-service agree

guardrex:
Hello @Runaho ... This isn't a patch update to non-working/poorly-implemented code. The original code works. Normally, we ***do*** prefer an issue to discuss significant changes. It sounds more like the code didn't work in your use case, and perhaps what we should do is make this say that the demo code is a starting point for further customization.

@MackinnonBuck ... The section where this code appears is at ...
 
https://learn.microsoft.com/en-us/aspnet/core/blazor/state-management?view=aspnetcore-7.0&pivots=server#in-memory-state-container-service-server

... and I just confirmed that it works as shown. Do you want to take a look, or do you think Steve or Javier should look at this?

Ultimately, we'll do one of the following ...

* Accept the changes, perhaps with a few NIT updates.
  * Either only show it with @Runaho's changes.
  * Show both the original and @Runaho's approach SxS because the original works in some cases and @Runaho's approach is appropriate in a different situation.
* Just add that this is a starting point for further customization.
* Close as a `won't-fix`.

Runaho:
I have prepared an update that you can try on this issue, I am leaving a simple example repo link.
There is also a video in Readme.
@guardrex 

https://github.com/Runaho/BlazorStateManagement

guardrex:
I think that's covered in the *Rendering* article in more detail ...

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/rendering?view=aspnetcore-7.0#receiving-a-call-from-something-external-to-the-blazor-rendering-and-event-handling-system

I need to digest this scenario a bit more ... things like what happens if your `SendMessage` call in the component is wrapped in `InvokeAsync` to perhaps dispatch it. I'm not saying that will work. I'd like to compare and contrast the examples in the docs with your use case.

I'm working on a potentially complicated set of updates to our form docs, so I'll need to circle back around to this. I'll try to get back to it by tomorrow ...... *but no promises on that!* 😆🏃🏃🏃🏃😅

Runaho:
Thank you for your effort @guardrex 
As far as I understand, when we look at the example, it has been added as scoped, not as a singleton, in Server Project so it will only work on the same client and will work correctly.
But when we want to trigger more than one client, this example causes an error.

Maybe we can add it to the document instead of editing it directly?

Runaho:
Got it, thank you.
The main goal here is to be able to trigger other clients with a singleton state container rather than creating a new hub since Blazor has already opened the SignalR socket.

When you look at it this way, it actually provides a really easy implementation.
This topic is no longer about simply editing documents :)

I think with this approach, we are already using the advantage of the Blazor Server application running on a single server and we are building a pusher-like structure.

A dynamic Blazor Rendering sounds really nice and it's a nice privilege to be able to use it as (SSR) or (CSR) when I want.

Maybe we can close this thread or turn it into something else after the other people you mentioned have expressed their opinion about it.

guardrex:
We've scoped this issue to an ask for a more general state container service than what the topic provides, given that the article is chiefly concerned with state management for a single client (single circuit). To help unblock devs who hit the `The current thread is not associated with the Dispatcher.` error, this article should cross-link the *Rendering* article that explains how to deal with callbacks invoked outside of Blazor's synchronization context. I think it makes the most sense to compose the cross-link in a *Troubleshoot* section, as opposed to just having it in an *Additional Resources* section. I'm going to fix this issue that way. I'll also add a long-range tracking issue to review this issue for more work in 2024.

PR #29777, opened 2023-07-11
Update browser limitation note in SignalR code sample

BrennanConroy:


Rick-Anderson (approved):

Rick-Anderson:
@BrennanConroy  let me know if you'd like to call this out in https://learn.microsoft.com/en-us/aspnet/core/signalr/authn-and-authz

BrennanConroy:
It's already called out in that page.

PR #29885, opened 2023-07-20
Update not-latest-version.md

tdykstra:
Follow-up to #29839 
Provides a link to current version instead of directions on how to switch manually.


Rick-Anderson:
Can you add

```
:::moniker range="= aspnetcore-8.0"

This is a prerelease

:::moniker end
```

Rick-Anderson (approved):

tdykstra:
> Can you add
> 
> ```
> :::moniker range="= aspnetcore-8.0"
> 
> This is a prerelease
> 
> :::moniker end
> ```

It already has a > 7 moniker for the prerelease case, and it already has text for it -- how could we add "This is a prerelease"?

PR #29895, opened 2023-07-21
Default HostingModel is now InProcess but web.config docs say otherwise

stevetemple:
From reading the rest of the documentation it appears InProcess is now the default hosting model, updating this to reflect that

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/host-and-deploy/iis/web-config.md](https://github.com/dotnet/AspNetCore.Docs/blob/2f7582e029dea9f6db0ba3667aef58cedbdade19/aspnetcore/host-and-deploy/iis/web-config.md) | [`web.config` file](https://review.learn.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/web-config?branch=pr-en-us-29895) |


guardrex:
Hi @stevetemple  ... The default for the app, yes. When it isn't set in the file, the default is still for out-of-process ...

https://github.com/dotnet/aspnetcore/blob/main/src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/ShimOptions.cpp#L23-L26

guardrex:
It's true that the SDK makes ***projects*** adopt in-process by default ... that's right. However, the SDK puts the `hostingModel` attribute for in-process into the *web.config* file when it generates the file. It has to be assigned to `in-process` to get in-process. If it's taken out of the ANCM config manually by the dev or via some automated *web.config* transform process, the default as far as the ANCM config is concerned will go to out-of-process. This text is ok because it specifically refers to what happens if `hostingModel` isn't present.

PR #29906, opened 2023-07-25
Update _SelectLanguagePartial.cshtml

MohammadElKhatib:
Include QueryString in the returnUrl

### DIFF
--- a/aspnetcore/fundamentals/localization/sample/6.x/Localization/Views/Shared/_SelectLanguagePartial.cshtml
+++ b/aspnetcore/fundamentals/localization/sample/6.x/Localization/Views/Shared/_SelectLanguagePartial.cshtml
@@ -12,7 +12,7 @@
     var cultureItems = LocOptions.Value.SupportedUICultures
         .Select(c => new SelectListItem { Value = c.Name, Text = c.DisplayName })
         .ToList();
-    var returnUrl = string.IsNullOrEmpty(Context.Request.Path) ? "~/" : $"~{Context.Request.Path.Value}";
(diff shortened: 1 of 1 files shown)


aspnetcore/fundamentals/localization/sample/6.x/Localization/Views/Shared/_SelectLanguagePartial.cshtml
```diff
@@ -12,7 +12,7 @@
     var cultureItems = LocOptions.Value.SupportedUICultures
         .Select(c => new SelectListItem { Value = c.Name, Text = c.DisplayName })
         .ToList();
-    var returnUrl = string.IsNullOrEmpty(Context.Request.Path) ? "~/" : $"~{Context.Request.Path.Value}";
+    var returnUrl = string.IsNullOrEmpty(Context.Request.Path) ? "~/" : $"~{Context.Request.Path.Value + Context.Request.QueryString.Value}";
```
Rick-Anderson, line 15: @hishamco please review

hishamco: Seems you added the query string only for non-default URL. The second thing where we are using the query string in the sample?

MohammadElKhatib: Well, I believe it is good to add the query string here even though in the sample there is no query string but mainly when I was following the sample in my application there was a query string causing exceptions. Also if no query string in the URL it will not be affected as the value will be an empty string.

hishamco: I disagree to add a query string to the sample if it's not needed, but any dev can add it if his/her application use it

PR #30033, opened 2023-08-11
Improve documentation around .NET Framework and gRPC-Web in the client

JamesNK:
Fixes https://github.com/grpc/grpc-dotnet/issues/2197

@jtattermusch Is this what you're looking for?

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/grpc/netstandard.md](https://github.com/dotnet/AspNetCore.Docs/blob/32e76a58222347e7dd61c9b5f93277275adc9ecd/aspnetcore/grpc/netstandard.md) | [Use gRPC client with .NET Standard 2.0](https://review.learn.microsoft.com/en-us/aspnet/core/grpc/netstandard?branch=pr-en-us-30033) |
| [aspnetcore/grpc/supported-platforms.md](https://github.com/dotnet/AspNetCore.Docs/blob/32e76a58222347e7dd61c9b5f93277275adc9ecd/aspnetcore/grpc/supported-platforms.md) | [gRPC on .NET supported platforms](https://review.learn.microsoft.com/en-us/aspnet/core/grpc/supported-platforms?branch=pr-en-us-30033) |

### DIFF
--- a/aspnetcore/grpc/netstandard.md
+++ b/aspnetcore/grpc/netstandard.md
@@ -70,7 +70,7 @@ Requirements and restrictions to using `WinHttpHandler`:
 
 * Windows 11 or later, Windows Server 2022 or later.
   * gRPC client is fully supported on Windows 11 or later.
-  * gRPC client is partially supported on Windows Server 2022 or later. Unary and server streaming methods are supported. Client and bidirectional streaming methods are not supported.
+  * gRPC client is partially supported on Windows Server 2022. Unary and server streaming methods are supported. Client and bidirectional streaming methods are not supported.
 * A reference to [`System.Net.Http.WinHttpHandler`](https://www.nuget.org/packages/System.Net.Http.WinHttpHandler/) version 6.0.1 or later.
 * Configure `WinHttpHandler` on the channel using `GrpcChannelOptions.HttpHandler`.
 * .NET Framework 4.6.1 or later.
--- a/aspnetcore/grpc/netstandard.md
+++ b/aspnetcore/grpc/netstandard.md
@@ -70,7 +70,7 @@ Requirements and restrictions to using `WinHttpHandler`:
 
 * Windows 11 or later, Windows Server 2022 or later.
   * gRPC client is fully supported on Windows 11 or later.
-  * gRPC client is partially supported on Windows Server 2022. Unary and server streaming methods are supported. Client and bidirectional streaming methods are not supported.
+  * gRPC client is partially supported on Windows Server 2022. Unary and server streaming methods are supported. Client and bidirectional streaming methods are **_not_** supported.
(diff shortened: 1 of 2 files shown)


aspnetcore/grpc/netstandard.md
```diff
@@ -69,6 +69,8 @@ For more information, see [Configure gRPC-Web with the .NET gRPC client](xref:gr
 Requirements and restrictions to using `WinHttpHandler`:
 
 * Windows 11 or later, Windows Server 2022 or later.
+  * gRPC client is fully supported on Windows 11 or later.
+  * gRPC client is partially supported on Windows Server 2022. Unary and server streaming methods are supported. Client and bidirectional streaming methods are not supported.
```
Rick-Anderson, line 73: Pick one
```suggestion
  * gRPC client is partially supported on Windows Server 2022. Unary and server streaming methods are supported. Client and bidirectional streaming methods aren't supported.
```

```suggestion
  * gRPC client is partially supported on Windows Server 2022. Unary and server streaming methods are supported. Client and bidirectional streaming methods are **_not_** supported.
```

Rick-Anderson (approved):

jtattermusch:
@JamesNK Yes, thanks for updating the docs.

aspnetcore/grpc/netstandard.md
```diff
@@ -69,6 +69,8 @@ For more information, see [Configure gRPC-Web with the .NET gRPC client](xref:gr
 Requirements and restrictions to using `WinHttpHandler`:
 
 * Windows 11 or later, Windows Server 2022 or later.
+  * gRPC client is fully supported on Windows 11 or later.
+  * gRPC client is partially supported on Windows Server 2022. Unary and server streaming methods are supported. Client and bidirectional streaming methods are **_not_** supported.
```
jtattermusch, line 73: qq: is the mode of failure for "not supported" functionality documented somewhere else than the code itself?

Basically, 
1. What happens when you try any gRPC call on platform that doesn't support WinHttpHandler at all.
2. what happens when you attempt client streaming or bi streaming call on a system where that's not supported.
3. what happens when you try non-TLS with WinHttpHandler

I don't think this needs to be documented here specifically, but may well be worth mentioning somewhere else.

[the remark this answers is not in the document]
JamesNK: Probably the next version of Windows Server. 2024?

JamesNK: > What happens when you try any gRPC call on platform that doesn't support WinHttpHandler at all.

I think WinHttpHandler throws a platform not supported exception that Windows is required.

> what happens when you attempt client streaming or bi streaming call on a system where that's not supported.
> what happens when you try non-TLS with WinHttpHandler

I recall that in both these cases WinHttpHandler silently falls back to sending an HTTP/1.1 request which then fails.

wadepickett (approved):
Looks good with currently made.

PR #30109, opened 2023-08-20
Editing Asp.Net documents

onurpicakci:
While looking at a case in the documentation I saw that the Display attribute remains as a comment. It's not a big problem but I wanted to open pr.

Related link -> https://learn.microsoft.com/en-us/aspnet/core/tutorials/razor-pages/validation?view=aspnetcore-7.0&tabs=visual-studio#add-validation-rules-to-the-movie-model

Ekran Resmi 2023-08-20 15 15 31

### DIFF
--- a/aspnetcore/tutorials/razor-pages/razor-pages-start/sample/RazorPagesMovie80/Models/MovieDateRatingDA.cs
+++ b/aspnetcore/tutorials/razor-pages/razor-pages-start/sample/RazorPagesMovie80/Models/MovieDateRatingDA.cs
@@ -17,7 +17,7 @@ public class Movie
     // </snippet11>
 
     // <snippet2>
-    // [Display(Name = "Release Date")]
+    [Display(Name = "Release Date")]
     [DataType(DataType.Date)]
     public DateTime ReleaseDate { get; set; }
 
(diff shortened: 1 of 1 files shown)


aspnetcore/tutorials/razor-pages/razor-pages-start/sample/RazorPagesMovie80/Models/MovieDateRatingDA.cs
```diff
@@ -17,7 +17,7 @@ public class Movie
     // </snippet11>
 
     // <snippet2>
-    // [Display(Name = "Release Date")]
+    [Display(Name = "Release Date")]
```
Rick-Anderson, line 20:
```suggestion
```
My guess is that it should be removed. @wadepickett ?

Rick-Anderson: Or maybe not this late in the tutorial. I think it was used earlier so should be there. The added attributes need highlighting.

onurpicakci: Yes, it would be more appropriate to remove this part of the document.

Rick-Anderson (approved):

PR #30166, opened 2023-08-27
Update bUnit docs

### LINKED ISSUE #30165: bUnit Example Update
The current bUnit example on the page needs to be updated and reflect our (cc @egil) latest recommendation.
I do understand that it's "just" a 3rd party library and maintaining the docs might be too much. That is why, if accepted, I can file a pull request.

Our recommendation would be:
1. Use razor syntax instead of "plain old" C# syntax. It might make sense to have both examples prominent for the user.
2. Inherit from `TestContext` instead of creating a new instance per test

The first approach could be achieved thanks to docfx quite easily:

~~~no-class
# [Razor](#tab/razor]
```razor
Here goes razor code
```

# [Csharp](#tab/csharp]
```csharp
Here goes C# code
```

---
~~~

---
#### Document Details

* ID: 61f0e0ba-e807-c5e8-3120-dbcf35a98f59
* Version Independent ID: 2890676f-3c63-5ab7-dd31-6b223212da30
* Content: [Test Razor components in ASP.NET Core Blazor](https://learn.microsoft.com/en-us/aspnet/core/blazor/test?view=aspnetcore-7.0)
* Content Source: [aspnetcore/blazor/test.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/test.md)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

linkdotnet:
Fixes #30165 

Following changes were made:
 * Showcasing `.razor` file based tests
 * Showcasing that `TestContext` should not be created in each test
 * Moved wording from "selecting" a button to "clicking"

CC @egil

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/test.md](https://github.com/dotnet/AspNetCore.Docs/blob/ae58f0f09fd568ee7f23f286a8f767c6c17558cc/aspnetcore/blazor/test.md) | [Test Razor components in ASP.NET Core Blazor](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/test?branch=pr-en-us-30166) |

### DIFF
--- a/aspnetcore/blazor/test.md
+++ b/aspnetcore/blazor/test.md
@@ -105,7 +105,7 @@ The following demonstrates the structure of a bUnit test on the `Counter` compon
 }
 ```
 
-The following bUnit test verifies that the CUT's counter is incremented correctly when the button is clicked:
+The following bUnit test verifies that the CUT's counter is incremented correctly when the button is selected:
 
 ```razor
 @code {
@@ -127,7 +127,7 @@ The following bUnit test verifies that the CUT's counter is incremented correctl
 }
 ```
 
-Of course, tests can also be written in classic `.cs` files:
+Tests can also be written in a C# class file:
 
 ```csharp
 public class CounterTests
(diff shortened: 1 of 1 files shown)


egil:
Even though we recommend just inheriting from TestContext to make tests easier to write, I don't think we should show that here since it only works with xUnit unless you take special steps. Users are likely to drop by bUnit.dev anyway to learn more.

We could add it as a tip with a deep link to the bUnit page where we discuss it.

linkdotnet:
The section afterward references bUnit - so I guess that's covered.

I'll revert that part - even though I am inclined to say that for an example it doesn't have to be 100% applicable for every user out of the box.

guardrex:
I'll review and update now.

> Moved wording from "selecting" a button to "clicking"

I'll revert that on my edit pass. That's a MS Style Manual requirement because the docs are written for devices with and without keyboards. They felt that "select" was better than "click" for devices generally.

guardrex:
cc: @Rick-Anderson @tdykstra @wadepickett ... FYI: There's a bad link popping up on builds ...

> aspnetcore/migration/proper-to-2x/mvc2.md
>
> 'https://vswebessentials.com/features/browserlink' points to a page  that doesn't exist. Check the path or URL and update the link.

When I hit that address, Firefox isn't to keen on my going there 😨 ...

Screenshot 2023-08-28 080612

egil:
Looks good to me.

linkdotnet:
> When I hit that address, Firefox isn't too keen on my going there 😨 ...

Me neither :D : 

image

guardrex (approved):
Thanks @linkdotnet 🥁 and @egil 🎸. *Merging now ..............*

PR #30288, opened 2023-09-08
fix logger /7

Rick-Anderson:
Fixes https://github.com/dotnet/AspNetCore.Docs/issues/30281

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/fundamentals/logging/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/4914aff9b2c7bebad7ab2753d9bb04c41c9a5fac/aspnetcore/fundamentals/logging/index.md) | [Logging in .NET Core and ASP.NET Core](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/logging/index?branch=pr-en-us-30288) |

### DIFF
--- a/aspnetcore/fundamentals/logging/index.md
+++ b/aspnetcore/fundamentals/logging/index.md
@@ -233,8 +233,6 @@ Calling `CreateLogger` with a fixed name can be useful when used in multiple met
 
 `ILogger<T>` is equivalent to calling `CreateLogger` with the fully qualified type name of `T`.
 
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
@cijothomas removed note. It's now in a comment so won't show. But what about folks that call `loggerfactory.create`? Shouldn't they be warned? 

I could also removed the commented out note if it's inaccurate.

cijothomas:
>  But what about folks that call loggerfactory.create? Shouldn't they be warned?

Why would someone do that in asp.net core? I understand previously people did it, but that could be because this doc itself referring it! Now that we removed it, I don't think we need to give that warning.

Will open a separate issue to help answer the question "what about logging before DI is ready?", with the answer to it as well (and that involves LoggerFactory.Create, and hence should come with the warning!)

We are waiting for the OpenTelemetry docs to clarify it first, and then come back and fix it right here! https://github.com/open-telemetry/opentelemetry-dotnet/pull/4821

PR #30336, opened 2023-09-13
Enhanced navigation & form handling 8.0

### LINKED ISSUE #30322: Blazor enhanced navigation and refresh API 8.0
* https://github.com/dotnet/aspnetcore/issues/50014

... and the entry in the What's New content should ...

* Include the API mention.
* Have the cross-link to the *Routing* article.

---
#### Document Details

* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #30322
Addresses #28161

* I'm using an informal name, *Normal navigation*, in contrast to *Enhanced navigation*. Does that choice make sense?
* I'm concerned about scoping what enhanced nav applies to. We're saying "if available," but can we specify *exactly* when it's available?
* I'm going to add the What's New cross-link to this immediately before merging or on a separate PR. I can't hold up that article with other doc 🐈🐈🐈 also trying to get updates applied to it. 

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/forms-and-input-components.md](https://github.com/dotnet/AspNetCore.Docs/blob/54f1a957e20b8ef019aaaef93f1cd674c2324722/aspnetcore/blazor/forms-and-input-components.md) | [ASP.NET Core Blazor forms and input components](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/forms-and-input-components?branch=pr-en-us-30336) |
| [aspnetcore/blazor/fundamentals/routing.md](https://github.com/dotnet/AspNetCore.Docs/blob/54f1a957e20b8ef019aaaef93f1cd674c2324722/aspnetcore/blazor/fundamentals/routing.md) | [ASP.NET Core Blazor routing and navigation](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?branch=pr-en-us-30336) |
| [aspnetcore/blazor/fundamentals/startup.md](https://github.com/dotnet/AspNetCore.Docs/blob/54f1a957e20b8ef019aaaef93f1cd674c2324722/aspnetcore/blazor/fundamentals/startup.md) | [ASP.NET Core Blazor startup](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/startup?branch=pr-en-us-30336) |
| [aspnetcore/release-notes/aspnetcore-8.0.md](https://github.com/dotnet/AspNetCore.Docs/blob/54f1a957e20b8ef019aaaef93f1cd674c2324722/aspnetcore/release-notes/aspnetcore-8.0.md) | [aspnetcore/release-notes/aspnetcore-8.0](https://review.learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-8.0?branch=pr-en-us-30336) |


MackinnonBuck (commented):
> I'm using an informal name, Normal navigation, in contrast to Enhanced navigation. Does that choice make sense?

That seems reasonable. Another term explained [here](https://github.com/WICG/navigation-api#appendix-types-of-navigations) is "cross-document navigation" (and enhanced navigation would be an example of "same-document-navigation"). Maybe we could refer to that document to clarify that "normal" navigation means "cross-document navigation" and "enhanced" navigation is a form of "same-document navigation"?

> I'm concerned about scoping what enhanced nav applies to. We're saying "if available," but can we specify exactly when it's available?

Enhanced navigation is available when:
* `blazor.web.js` is used (not `blazor.server.js` or `blazor.webassembly.js`)
* It's not explicitly disabled via `Blazor.start({ ssr: { disableDomPreservation: true } })`
* The destination URL is within the internal base URI space

But it should be noted that even if enhanced navigation is available, navigations will still go through an interactive (client-side) router if one exists.

guardrex:
Thx @MackinnonBuck ... I'll make another pass using that information and ping u back.

guardrex:
* I'm drafting this now with "cross-document/cross-origin" and "same-document/same-origin" adjectives. Let's see if that works. If so, we probably don't need to cross-link.
* I'll also add the phrasing on "URI space" that we normally use (app base path) ... ***IF*** that's a valid equivalent expression. 👂
* I add to the *Startup* doc on disabling enhanced nav. 

❓ on ...

> But it should be noted that even if enhanced navigation is available, navigations will still go through an interactive (client-side) router if one exists.

I added that to the text, but why is that important to tell readers? It seems like it needs an additional remark as to why the reader would care.

guardrex:
# 🛑 

***NOTE TO SELF***: Add link to What's New (release notes) before merging!

aspnetcore/blazor/fundamentals/routing.md
```diff
@@ -500,6 +515,48 @@ Use <xref:Microsoft.AspNetCore.Components.NavigationManager> to manage URIs and
...
+:::moniker range=">= aspnetcore-8.0"
+
+## Enhanced navigation and form handling
+
+*This section applies to Blazor Web Apps.*
+
+Blazor Web Apps are capable of two types of routing for page navigation and form handling requests:
+
+* Normal navigation: For all cross-document/cross-origin navigation, a full-page reload is triggered for the request URL.
+* Enhanced navigation: For some same-document/same-origin navigation&dagger;, Blazor intercepts the request and performs a `fetch` request instead. Blazor then patches the response content into the page's DOM. Blazor's enhanced navigation and form handling avoid the need for a full-page reload and preserves more of the page state, so pages load faster, usually without losing the user's scroll position on the page.
```
MackinnonBuck, line 527: Just to clarify - cross-document and cross-origin are two different things. Cross-document means that you lose JavaScript/Document state between navigations. Cross-origin means that the navigation results in the origin changing. A navigation can be same-origin but still cross-document. Hitting the refresh button is an example of this.

Cross-origin navigations are always cross-document. Same-origin navigations have the _potential_ to be same-document. Enhanced navigation is an example of a same-origin and same-document navigation. If enhanced navigation is explicitly disabled (and there's no interactive router), then most same-origin navigations will be cross-document (link clicks, hitting the back/forward button, etc.).

It's probably not necessary to elaborate on all this in these docs, but maybe we could phrase these definitions like:
```suggestion
* Normal navigation: Cross-document navigation; a full-page reload is triggered for the request URL.
* Enhanced navigation: Same-document navigation;&dagger;, Blazor intercepts the request and performs a `fetch` request instead. Blazor then patches the response content into the page's DOM. Blazor's enhanced navigation and form handling avoid the need for a full-page reload and preserves more of the page state, so pages load faster, usually without losing the user's scroll position on the page.
```

I think the key is to be clear that the same-document/cross-document decision doesn't happen _before_ Blazor decides whether to do an enhanced or normal navigation. The same-origin/cross-origin decision happens first (by nature of the URL being navigated to). Then, if the navigation is same-origin, Blazor decides whether to perform a same-document or cross-document navigation.

Apologies if it seems like I'm explaining this to death, or if this is information you already knew!

MackinnonBuck:
> I added that to the text, but why is that important to tell readers? It seems like it needs an additional remark as to why the reader would care.

It's probably not going to be relevant to most customers. It really only applies if you were to have, for example, an interactive `<Router>` nested inside a server-side rendered `<Router>`. In such a case, the interactive router would take priority when handling navigations, so enhanced navigation wouldn't be used for navigation (since that's a server-side routing feature).

It's possible this is a detail we can leave out, since that's a very uncommon pattern - I'll leave that up to you!

guardrex:
We can keep it ... let me try with a remark that it's an uncommon concern but for full disclosure. I'll try that now ... brb ..........

guardrex:
Ok ... 

* I try to format the bullet list items as ...

> \* Normal navigation (cross-document navigation): a full-page ...
> \* Enhanced navigation (same-document navigation)&dagger;: Blazor intercepts ...

* I tried to roll in the full disclosure remark. If we don't do something like this, it's hard to understand why the remark (first sentence now) is there.

guardrex:
@MackinnonBuck ... I'll assume this is ok now and go ahead. I'm about to wrap up for the week, and I'd like to get this in before I update DR and Artak on where everything stands. Msg me if this needs a patch.

guardrex:
@MackinnonBuck ... One more side-question ...

WRT RC2 and forms to enable enhanced nav, the behavior will be opt-in ...

> Enhanced form handling can be optionally enabled for specific forms.

I'm looking at Steve's issue ...

https://github.com/dotnet/aspnetcore/pull/50551

... and will be documenting ***for RC2*** that ...

> You turn it on by adding `data-enhance` to the `<form>` element or `Enhance` if it's an `EditForm`. This is nonhierarchical since it's not desirable to enable this globally.

... so it will need to be done for every form that the dev wants enhanced nav for.

For "nonhierarchical," confirm if that it just means it must be enabled on a per-form basis. Any embedded forms in a form also needs the attribute/param.

MackinnonBuck:
> I had to play with it on a few commits, but that seems like a good draft. How does it look?

Looks great 🙂 

> For "nonhierarchical," confirm if that it just means it must be enabled on a per-form basis. Any embedded forms in a form also needs the attribute/param.

Yep, that's correct. You can't, for example, put `data-enhance` on an ancestor element to a `<form>` and expect enhanced navigation to be enabled for the form.

PR #30351, opened 2023-09-14
Update best-practices large objects threshold

WeihanLi:
according to

https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/large-object-heap#how-an-object-ends-up-on-the-loh

the 85,000 bytes object would be treat as large object also

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/fundamentals/best-practices.md](https://github.com/dotnet/AspNetCore.Docs/blob/e20fae5a35c7ef08603c05f394da75070ac699d5/aspnetcore/fundamentals/best-practices.md) | [ASP.NET Core Best Practices](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/best-practices?branch=pr-en-us-30351) |


guardrex (approved):
Thanks @WeihanLi! 🚀

PR #30430, opened 2023-09-21
Update introduction-to-aspnet-core.md

Rick-Anderson:
No longer use regions

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/introduction-to-aspnet-core.md](https://github.com/dotnet/AspNetCore.Docs/blob/780bb073179be7c06e76af16678ebd3fb7e4f3d6/aspnetcore/introduction-to-aspnet-core.md) | [Overview of ASP.NET Core](https://review.learn.microsoft.com/en-us/aspnet/core/introduction-to-aspnet-core?branch=pr-en-us-30430) |


tdykstra (commented):
We still have a lot of regions in the repo, it seems premature to delete this explanation without converting the remaining ones to XML tags.

tdykstra (approved):
But I'll approve anyway -- go ahead and merge if you think it's OK to do so while there are still regions in use. We should probably have a similar explanation for XML tags.

Rick-Anderson:
> We still have a lot of regions in the repo, it seems premature to delete this explanation without converting the remaining ones to XML tags.

There shouldn't be any in articles that have been edited, which are the mostly likely articles/code that customers will review.

> We should probably have a similar explanation for XML tags.

I don’t think `// snippet comments` need to be explained. `// snippet comments`  are fairly obvious and not as obtrusive.

Issue #30444, opened 2023-09-22
Details on keepalive for gRPC in Azure App Service


karpikpl:
Hi,
I've been running gRPC server on an App Service.
I noticed my long running streaming call has been failing with 504 issues. To fix it I've implemented a dummy keep-alive message to be sent from the server every 2 minutes and that fixed it.

But then I noticed this [documentation](https://learn.microsoft.com/en-us/aspnet/core/grpc/performance?view=aspnetcore-7.0#keep-alive-pings) and I switched to keepalive pings configured on the channel. After doing that, the issue started again - you can see it in attached screenshot (before - my dummy keepalive call that has to be ignored by the client, later - keepalive pings)

![image](https://github.com/dotnet/AspNetCore.Docs/assets/3539908/1f3086df-f1ad-45bb-b56b-a6c21bdb122e)

Can the docs be updated for how to use keepalives in Azure? 

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 484737e7-abb8-c142-35f8-22a35817270d
* Version Independent ID: 9e412677-c511-9f5b-4654-95e0177c7dd4
* Content: [Performance best practices with gRPC](https://learn.microsoft.com/en-us/aspnet/core/grpc/performance?view=aspnetcore-7.0)
* Content Source: [aspnetcore/grpc/performance.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/grpc/performance.md)
* Product: **aspnet-core**
* Technology: **aspnetcore-grpc**
* GitHub Login: @JamesNK
* Microsoft Alias: **jamesnk**

wadepickett:
Thanks @karpikpl for the details on your experience for the issue.

@JamesNK, is there any guidance or clarification to add specific to Azure App Service for this section?  [Performance best practices with gRPC: section: Keep alive pings](https://learn.microsoft.com/en-us/aspnet/core/grpc/performance?view=aspnetcore-7.0#keep-alive-pings)

karpikpl:
Here's my keep alive client configuration - copied from docs
```csharp
builder.Services
    .AddGrpcClient<Notifications.Notifier.NotifierClient>(o =>
    {
        o.Address = new Uri(downstreamApiUrl);
        o.ChannelOptionsActions.Add(channelOptions =>
        {
            channelOptions.HttpHandler = new SocketsHttpHandler
            {
[...]

PR #30489, opened 2023-09-27
Update performance.md

JamesNK:
Addresses https://github.com/dotnet/AspNetCore.Docs/issues/30444

Keep alive pings don't help to keep long-running gRPC calls alive.

Feel free to reword to use the correct language.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/grpc/performance.md](https://github.com/dotnet/AspNetCore.Docs/blob/e64f9ffaaaa2e7a70821f5a9d76e3cce5762ac1f/aspnetcore/grpc/performance.md) | [Performance best practices with gRPC](https://review.learn.microsoft.com/en-us/aspnet/core/grpc/performance?branch=pr-en-us-30489) |

### DIFF
--- a/aspnetcore/grpc/performance.md
+++ b/aspnetcore/grpc/performance.md
@@ -153,6 +153,9 @@ var channel = GrpcChannel.ForAddress("https://localhost:5001", new GrpcChannelOp
 
 The preceding code configures a channel that sends a keep alive ping to the server every 60 seconds during periods of inactivity. The ping ensures the server and any proxies in use won't close the connection because of inactivity.
 
+> [!NOTE]
+> Keep alive pings only help keep the connection alive. Long-running gRPC calls on the connection may still be killed by the server or intermediary proxies for inactivity.
+
 :::moniker-end
 
 ## Flow control
--- a/aspnetcore/grpc/performance.md
+++ b/aspnetcore/grpc/performance.md
@@ -154,7 +154,7 @@ var channel = GrpcChannel.ForAddress("https://localhost:5001", new GrpcChannelOp
 The preceding code configures a channel that sends a keep alive ping to the server every 60 seconds during periods of inactivity. The ping ensures the server and any proxies in use won't close the connection because of inactivity.
 
 > [!NOTE]
-> Keep alive pings only help keep the connection alive. Long-running gRPC calls on the connection may still be killed by the server or intermediary proxies for inactivity.
+> Keep alive pings only help keep the connection alive. Long-running gRPC calls on the connection may still be terminated by the server or intermediary proxies for inactivity.
 
 :::moniker-end
 


aspnetcore/grpc/performance.md
```diff
@@ -153,6 +153,9 @@ var channel = GrpcChannel.ForAddress("https://localhost:5001", new GrpcChannelOp
 
 The preceding code configures a channel that sends a keep alive ping to the server every 60 seconds during periods of inactivity. The ping ensures the server and any proxies in use won't close the connection because of inactivity.
 
+> [!NOTE]
+> Keep alive pings only help keep the connection alive. Long-running gRPC calls on the connection may still be killed by the server or intermediary proxies for inactivity.
```
chwarr, line 157: >  Long-running gRPC calls on the connection may still be killed by the server

Could this link to documentation for how to configure the service-side per-call timeout (for a .NET-based gRPC server)?

Rick-Anderson: > > Long-running gRPC calls on the connection may still be killed by the server
> 
> Could this link to documentation for how to configure the service-side per-call timeout (for a .NET-based gRPC server)?

@chwarr  if you supply a link I can add it.

JamesNK: I don't believe Kestrel has a limit by default.

wadepickett, line 157:
```suggestion
> Keep alive pings only help keep the connection alive. Long-running gRPC calls on the connection may still be terminated by the server or intermediary proxies for inactivity.
```

wadepickett: That word change might work a little better in translations.

wadepickett (approved):
Looks good. I made a minor suggestion to consider

Rick-Anderson (approved):
Accept wade's suggestion. Can't use killed in general.

aspnetcore/grpc/performance.md
chwarr: > @chwarr if you supply a link I can add it.

I don't know, which is why I asked. :-)

I gather from JameNK's reply that there is nothing in Kestrel nor gRPC.NET that will terminate an idle gRPC call (or a HTTP2 stream) on a HTTP2 connection that has keep alive enabled.

I was reading too much into phrasing like "Keep alive pings don't help to keep long-running gRPC calls alive." and "I don't believe Kestrel has a limit _by default_." (emphasis mine) and assuming that there was a Kestrel setting. It's entirely possible that App Service is doing something to the idle gRPC calls/stream, but I can't find any documentation about that either... :-(

JamesNK: I think the App Service proxy has a maximum limit on request time based on some discussion I saw a while ago. I haven't tried it out myself.

PR #30558, opened 2023-10-03
Removed IdentityServer direct links

damienbod:
The links are removed and now part of the community links like all other OSS providers

https://learn.microsoft.com/en-us/aspnet/core/security/authentication/community


tdykstra (approved):
Thanks, @damienbod !

PR #30616, opened 2023-10-08
fix: remove unnecessary toArray since it leads to overhead of memory consumpion especially when the collection is large

### LINKED ISSUE #30615: unnecessary toArray in action
### Type of issue

Other (describe below)

### Description

the Get action within `WeatherForecastController` returns` IEnumerable<WeatherForecast>`, no need to call `ToArray` since it leads to overhead of memory consumpion especially when the collection is large.

If some beginners copy these sample codes into their projects which has a large collection, may be introduce performance issue.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/apis?view=aspnetcore-6.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/apis.md

### Version Independent Document ID

4841921d-fe70-1264-88ff-b2f2d067b212

### Article author

tdykstra

mulei2023:
Fixes #30615

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/hybrid/class-libraries.md](https://github.com/dotnet/AspNetCore.Docs/blob/531eba7a5b2340d490eb5845017fee4fc02f1ed2/aspnetcore/blazor/hybrid/class-libraries.md) | [Share assets across web and native clients using a Razor class library (RCL)](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/hybrid/class-libraries?branch=pr-en-us-30616) |

### DIFF
--- a/aspnetcore/blazor/hybrid/class-libraries.md
+++ b/aspnetcore/blazor/hybrid/class-libraries.md
@@ -141,7 +141,7 @@ namespace SharedLibrary.Interfaces;
 
 public interface IWeatherForecastService
 {
-    Task<WeatherForecast[]?> GetForecastAsync(DateTime startDate);
+    Task<IEnumerable<WeatherForecast>?> GetForecastAsync(DateTime startDate);
 }
 ```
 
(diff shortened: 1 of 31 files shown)


guardrex:
Hello @mulei2023 ... Let's discuss this over on the issue. I don't think we'll be able to accept this PR. If we do tho, no problem ... we can just re-open this. I'll explain further on the issue.

Issue #30715, opened 2023-10-17
Workflow automation


guardrex:
The workflow under development is here ...

https://github.com/dotnet/AspNetCore.Docs/blob/main/.github/workflows/new-blazor-issue.yml

The example for the first commit workflow is here ...

https://github.com/dotnet/AspNetCore.Docs/issues/30714

... which provides the dev a comment and an email ping for me via my GH handle.

The `peter-evans/create-or-update-comment@v2.1.0` action is only for comments.

To perform the following, I'll need to track down and implement additional actions for ...

* Assigning me to the issue.
* Modifying the labels:
  * Remove the `Not Triaged` label.
  * Add the `Blazor` label.
* Add the Blazor project, which is project 35.

Further, we'll want to duplicate this for the Blazor Hybrid issues, except ...

* The assignment will be to Beth and me.
* The label add will be `Blazor Hybrid`.
* The project add is the Blazor Hybrid project, which is project 106.

guardrex:
Currently, I'm investigating dropping `peter-evans/create-or-update-comment@v2.1.0` in favor of GH's new built-in GH-provided add comment feature.

guardrex:
All good thus far. After my GH script issue is resolved for adding the Blazor project, I'll create Blazor Hybrid workflow.

PR #30738, opened 2023-10-17
Clean up prior Blazor automation

guardrex:
Addresses #30715


guardrex:
@Rick-Anderson ... Let me know if you want me to set up workflows for the SignalR and Grpc pieces. However, the only thing this was doing was adding the `gRPC` and `SignalR` labels. It looks like the new labels are being assigned to issues. If you know that we don't need those two any longer, then I'll carve them out of the `.repoman.yml` file, too, and we'll just let the new labels take over at this point. 👂

guardrex:
Confirmed, @Rick-Anderson! The new label for SignalR is `aspnetcore-signalr/tech`, and the new label for GRPC is `aspnetcore-grpc/tech`. I'll go ahead and kill the SignalR and gRPC config in the repoman file. Just let me know if you want me to automate the old `SignalR` and `gRPC` labels via the workflow approach ... if you want those labels applied in addition to the new `/tech` labels.

guardrex:
BTW ... I ran tests to confirm none of the SignalR/gRPC repoman labels are working, see ............

* https://github.com/dotnet/AspNetCore.Docs/issues/30739
* https://github.com/dotnet/AspNetCore.Docs/issues/30740

PR #30793, opened 2023-10-20
Clarifying the Minimal API Authentication example.

AddisonRogers:
Adding the nuget package that is being used in the main text.
https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/security?view=aspnetcore-7.0

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/fundamentals/minimal-apis/security.md](https://github.com/dotnet/AspNetCore.Docs/blob/2ddef46e9512ebcd9c8dc8915cf1df7827b63395/aspnetcore/fundamentals/minimal-apis/security.md) | [Authentication and authorization in minimal APIs](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/security?branch=pr-en-us-30793) |

### DIFF
--- a/aspnetcore/fundamentals/minimal-apis/security.md
+++ b/aspnetcore/fundamentals/minimal-apis/security.md
@@ -32,7 +32,7 @@ To enable authentication, call [`AddAuthentication`](/dotnet/api/microsoft.exten
 
 :::code language="csharp" source="~/fundamentals/minimal-apis/security/7.0-samples/MinApiAuth/MinApiAuth/Program.cs" id="snippet_1" highlight="2":::
 
-Typically, a specific authentication strategy is used. In the following sample, the app is configured with support for JWT bearer-based authentication. This example makes use of the [`Microsoft.AspNetCore.Authentication.JwtBearer`](https://www.nuget.org/packages/Microsoft.AspNetCore.Authentication.JwtBearer) NuGet package, which is provided to the app by the [shared framework](xref:fundamentals/metapackage-app).
+Typically, a specific authentication strategy is used. In the following sample, the app is configured with support for JWT bearer-based authentication. This example makes use of the [`Microsoft.AspNetCore.Authentication.JwtBearer`](https://www.nuget.org/packages/Microsoft.AspNetCore.Authentication.JwtBearer) NuGet package, which isn't provided by the [shared framework](xref:fundamentals/metapackage-app).
 
 :::code language="csharp" source="~/fundamentals/minimal-apis/security/7.0-samples/MinApiAuth/MinApiAuth/Program.cs" id="snippet_jwt1" highlight="2-3":::
 
--- a/aspnetcore/fundamentals/minimal-apis/security.md
+++ b/aspnetcore/fundamentals/minimal-apis/security.md
@@ -32,7 +32,7 @@ To enable authentication, call [`AddAuthentication`](/dotnet/api/microsoft.exten
 
 :::code language="csharp" source="~/fundamentals/minimal-apis/security/7.0-samples/MinApiAuth/MinApiAuth/Program.cs" id="snippet_1" highlight="2":::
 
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks, @AddisonRogers! 🚀 ... I've pinged Safia to take a look, as she's listed as the author of this one.

guardrex:
@captainsafia ... I have a mild concern that since it's in the shared framework that someone might think that they need to add it. Seems like if we keep it that it should mention that, too.

Rick-Anderson:
@guardrex  are you sure, See [this](https://learn.microsoft.com/en-us/aspnet/core/security/authentication/jwt-authn?view=aspnetcore-7.0&tabs=windows#examples)

Rick-Anderson:
Also [Add package references for removed assemblies](https://learn.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-7.0&tabs=visual-studio#add-package-references-for-removed-assemblies)

ASP.NET Core 3.0 removes some assemblies that were previously part of the Microsoft.AspNetCore.App package reference.

To continue using features provided by the removed assemblies, reference the 3.0 versions of the corresponding packages:

OpenID Connect bearer token ([Microsoft.AspNetCore.Authentication.JwtBearer](https://www.nuget.org/packages/Microsoft.AspNetCore.Authentication.JwtBearer))

guardrex:
*Ooooops!* 🙈 You're right. I didn't physically go look. I corrected the bit I added. We could revert back to @AddisonRogers's original language, but wouldn't telling the reader that they should add the package or that it isn't in the shared framework be best?

As an aside, I wonder if we should add guidance to the [App metapackage article](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/metapackage-app) on how to physically see the packages for any given release?

guardrex (approved):

aspnetcore/fundamentals/minimal-apis/security.md
```diff
@@ -32,7 +32,7 @@ To enable authentication, call [`AddAuthentication`](/dotnet/api/microsoft.exten
 
 :::code language="csharp" source="~/fundamentals/minimal-apis/security/7.0-samples/MinApiAuth/MinApiAuth/Program.cs" id="snippet_1" highlight="2":::
 
-Typically, a specific authentication strategy is used. In the following sample, the app is configured with support for JWT bearer-based authentication.
+Typically, a specific authentication strategy is used. In the following sample, the app is configured with support for JWT bearer-based authentication. This example makes use of the [`Microsoft.AspNetCore.Authentication.JwtBearer`](https://www.nuget.org/packages/Microsoft.AspNetCore.Authentication.JwtBearer) NuGet package, which isn't provided by the [shared framework](xref:fundamentals/metapackage-app).
```
captainsafia, line 35: This text feels confusing only because the API referenced isn't actually in the shared framework as the prior clause mentions, it's in a NuGet package.

```suggestion
Typically, a specific authentication strategy is used. In the following sample, the app is configured with support for JWT bearer-based authentication. This example makes use of the APIs available in the [`Microsoft.AspNetCore.Authentication.JwtBearer`](https://www.nuget.org/packages/Microsoft.AspNetCore.Authentication.JwtBearer) NuGet package.
```

guardrex: I changed it to say "isn't provided," but I defer to your choice of words.

captainsafia: Oof, didn't notice! The original verbiage sounds good!

captainsafia (approved):

aspnetcore/fundamentals/minimal-apis/security.md
guardrex: It's cool ... I don't think most readers will be confused. If we hear 👂 that readers are confused, we can add the INCLUDE on how to add a package ...

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/includes/package-reference.md

I place that under every instruction in Blazor docs that has the user add a package.

PR #30825, opened 2023-10-24
Indent first elements in WeatherForecastExample

frjol:
Indenting the first elements in the generated example of a new project ASP.NET Core Web API.


tdykstra (approved):
Thanks, @frjol

PR #30844, opened 2023-10-26
Fix ASP.NET metric instrument name in example

j3parker:
As [documented elsewhere](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics-aspnetcore#instrument-httpserveractive_requests) (and [implemented](https://github.com/dotnet/aspnetcore/blob/6c1b3dfb66a1b66cc32ee26bbc23f6472f1dc985/src/Hosting/Hosting/src/Internal/HostingMetrics.cs#L30)) the instrument name uses dots rather than slashes: 

It seems that slashes were used in the [initial proposal](https://github.com/dotnet/aspnetcore/issues/47536).


guardrex (approved):
Thanks @j3parker! 🚀

PR #30849, opened 2023-10-26
WASM debug scenario update

### LINKED ISSUE #30824: Update Blazor WebAssembly debugging limitations
### Description

In the .NET 7 docs and earlier doc versions the Blazor WebAssembly debugging docs has a section that says "Unsupported scenarios for Blazor WebAssembly apps include:", but as far as I can tell, everything listed is actually supported. It looks like this was already fixed for the .NET 8 docs, but not the earlier doc sets.

Also, in the limitations section:

Limitation | Status | Addressed in
--- | --- | ---
Break on unhandled exceptions. | The debugger can break on unhandled exceptions, but Blazor currently catches all exceptions so they are all handled (see #42510). The workaround is to break when the exception is thrown. I think we should remove this item as a limitation and add a note about how to debug exceptions. | .NET 6
Hit breakpoints during app startup before the debug proxy is running. This includes breakpoints in the Program file and breakpoints in the [OnInitialized{Async} lifecycle methods](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?view=aspnetcore-7.0#component-initialization-oninitializedasync) of components that are loaded by the first page requested from the app. | This is no longer a limitation | .NET 6
Debug in non-local scenarios (for example, [Windows Subsystem for Linux (WSL)](https://learn.microsoft.com/en-us/windows/wsl/) or [Visual Studio Codespaces](https://visualstudio.microsoft.com/services/github-codespaces/)). | I think these scenarios are still unsupported. @thaysg? | TBD
Use a [symbol server](https://learn.microsoft.com/en-us/aspnet/core/test/debug-aspnetcore-source?view=aspnetcore-7.0) for debugging. | Addressed in .NET 8 | .NET 8

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/debug?view=aspnetcore-7.0&tabs=visual-studio

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/debug.md

### Document ID

4d06e565-282d-6387-aa90-cb2c2986d35a

### Article author

guardrex

guardrex:
Fixes  #30824

For ***adding a note about how to debug exceptions***, is the guidance what Thays provided at ...

* https://github.com/dotnet/aspnetcore/issues/42510#issuecomment-1171706250
* https://github.com/dotnet/aspnetcore/issues/42510#issuecomment-1171716418 

... what we're passing on? ~My plan is to place a section on it, which I've started but haven't finished yet.~ **UPDATE**: I've made the PR say what I think it's supposed to say.

~BTW, I'm not sure what Thays took a screenshot of for that. Where is that "All Exceptions" setting? ... and note that her text remark doesn't exactly match what's shown in the screenshot.~ ***Nevermind!*** I found it. I haven't looked at that UI in several years ... **Debug** > **Windows** > **Exception Settings**. I thought it would be in **Options**.

**UPDATE**: I've tried it here, and it doesn't seem to work. I'm setting ...

image

No 🎲🎲 ... still just breaks in the browser with the stack trace without the debugger indicating that anything has happened. I'm using the code from the issue that throws on a button click ...

```razor
<p>
    <button @onclick="ThrowException">Throw Exception</button>
</p>

@code {
    public void ThrowException()
    {
        throw new Exception("throwing new ex");
    }
}
```

... and I tried the `Counter` component approach ...

```razor
private void IncrementCount()
{
    currentCount++;

    if (currentCount == 3)
    {
        throw new Exception("throwing new ex");
    }
}
```

cc: @thaystg  

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/debug.md](https://github.com/dotnet/AspNetCore.Docs/blob/cada5992cd83f493dae62409ce96ad7e67602e0e/aspnetcore/blazor/debug.md) | [Debug ASP.NET Core apps](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/debug?branch=pr-en-us-30849) |

### DIFF
--- a/aspnetcore/blazor/debug.md
+++ b/aspnetcore/blazor/debug.md
@@ -35,14 +35,12 @@ Available scenarios for Blazor WebAssembly debugging include:
 
 Unsupported scenarios include:
 
-* Break on unhandled exceptions.
-* Hit breakpoints during app startup before the debug proxy is running. This includes breakpoints in the `Program` file and breakpoints in the [`OnInitialized{Async}` lifecycle methods](xref:blazor/components/lifecycle#component-initialization-oninitializedasync) of components that are loaded by the first page requested from the app.
 * Debug in non-local scenarios (for example, [Windows Subsystem for Linux (WSL)](/windows/wsl/) or [Visual Studio Codespaces](https://visualstudio.microsoft.com/services/github-codespaces/)).
 * Debug in Firefox from Visual Studio or Visual Studio Code.
 
 :::moniker-end
 
-:::moniker range="< aspnetcore-8.0"
+:::moniker range=">= aspnetcore-6.0 < aspnetcore-8.0"
 
 Blazor Server apps can be debugged in an IDE, Visual Studio or Visual Studio Code.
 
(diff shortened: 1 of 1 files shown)


guardrex:
@thays ... If the guidance here is correct in the new *Break on unhandled exceptions* section, I can't get it to work. All I get is the dev tools console logging of the error and stack trace. Nothing happens in the debugger. Any idea what I'm not doing correctly? ... or are these instructions incomplete?

thaystg:
I have just tested.
Started the app, enabled the exceptions, clicked in the "Click me" button.
image

thaystg:
image

guardrex:
# 🐀🍔 *Roasted Rat Burgers!* 🐀🍔 

😆 ... well ... it's not working here. It must be a local problem for *ME* 🦖. The guidance on the PR is correct, so I'm going to go ahead and merge it. I'll check again on this after RTM, when I'll be moving off of Preview VS back to normal release VS. Perhaps, something is funky with my Preview VS install.

PR #30884, opened 2023-10-31
Handlers and enhanced navigation 8.0

### LINKED ISSUE #30876: LocationChanged event and location changing handlers interaction with enhanced navigation 8.0
### Description

> When an enhanced navigation occurs, LocationChanged event handlers registered in interactive runtimes (i.e., Blazor Server/WebAssembly) get invoked. However, location changing handlers may or may not intercept enhanced navigations, so it's important that app logic not rely on location changing handlers being invoked reliably. For example, the user might switch to another page before an interactive runtime becomes available. In .NET 8, if server-side routing and enhanced navigation are enabled, location changing handlers will only get invoked for programmatic navigations initiated from an interactive runtime. In future .NET versions, more types of navigations (e.g., link clicks) may also invoke location changing handlers"

Per https://github.com/dotnet/aspnetcore/issues/51436#issuecomment-1786018976

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-7.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/fundamentals/routing.md

### Document ID

b6cb2574-6326-a40d-0b6c-d1661328e828

### Article author

guardrex

guardrex:
Fixes #30876

Mackinnon ...

* Seems like there were two paragraphs in the content that you supplied and one was more general than the other, so I placed one a little higher than the other.
* Since these paragraphs refer to location changing handlers and that section was ***below*** the new content on enhanced nav, I moved the *Location changes* section above the new *Enhanced navigation and form handling* section.

❓Should the second paragraph end with a remark like ...

> If the app must rely on a location change handler firing, don't use enhanced navigation, which is explained later in this section.

**UPDATE**: ***Yes!*** Let's give that line a try. You can let me know about keeping or 🔪 it on review.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/routing.md](https://github.com/dotnet/AspNetCore.Docs/blob/0dc441e6ea375f502dd20c6132606fad2744038a/aspnetcore/blazor/fundamentals/routing.md) | [ASP.NET Core Blazor routing and navigation](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?branch=pr-en-us-30884) |


aspnetcore/blazor/fundamentals/routing.md
```diff
@@ -532,8 +576,12 @@ Blazor Web Apps are capable of two types of routing for page navigation and form
...
+
 If enhanced navigation is available for interactive client routing, navigation always goes through the interactive client-side router. This point is only relevant in an uncommon scenario where an interactive `<Router>` is nested inside a server-side rendered `<Router>`. In that case, the interactive router takes priority when handling navigation, so enhanced navigation isn't used for navigation because it's a server-side routing feature.
 
+When an enhanced navigation occurs, [`LocationChanged` event handlers](#location-changes) registered with interactive server and WebAssembly runtimes are typically invoked. There are cases when location changing handlers might not intercept an enhanced navigation. For example, the user might switch to another page before an interactive runtime becomes available. Therefore, it's important that app logic not rely on invoking a location changing handler, as there's no guarantee of the handler executing. If the app must rely on a location change handler firing, don't use enhanced navigation, which is explained later in this section.
```
MackinnonBuck, line 583: This looks good. The thing that might need further clarification is:
> If the app must rely on a location change handler firing, don't use enhanced navigation, which is explained later in this section.

This makes it sound like the solution is to disable enhanced navigation, but actually won't make a difference. If they need location changing handlers to be reliably invoked for internal navigations, then they will need to use an interactive (client-side) router instead of server-side routing.

MackinnonBuck (approved):
Looks good!

aspnetcore/blazor/fundamentals/routing.md
guardrex: I'm *still* becoming acquainted with these concepts, so for ...

> use an interactive client-side router instead of server-side routing

... how is that done? Must they make the entire component adopt interactive WebAssembly rendering?

PR #30922, opened 2023-11-02
JavaScript with Blazor static server rendering

### LINKED ISSUE #30921: Using JavaScript with Blazor Static Server Rendering
### Description

In this node, react to https://github.com/dotnet/aspnetcore/issues/51331#issuecomment-1791582878.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/?view=aspnetcore-7.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/javascript-interoperability/index.md

### Document ID

49fa31d8-72a6-01a1-027d-6edeac9e0fc4

### Article author

guardrex

guardrex:
Fixes #30921
Addresses #28161

Mackinnon ... Check everything because I made a few guesses along the way ... *and we know how that goes* 🦖🙈😆. 

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/render-modes.md](https://github.com/dotnet/AspNetCore.Docs/blob/6b94e61cd5dc244598e1da70a8128a6998834e73/aspnetcore/blazor/components/render-modes.md) | [ASP.NET Core Blazor render modes](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/render-modes?branch=pr-en-us-30922) |
| [aspnetcore/blazor/fundamentals/routing.md](https://github.com/dotnet/AspNetCore.Docs/blob/6b94e61cd5dc244598e1da70a8128a6998834e73/aspnetcore/blazor/fundamentals/routing.md) | [ASP.NET Core Blazor routing and navigation](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?branch=pr-en-us-30922) |
| [aspnetcore/blazor/fundamentals/startup.md](https://github.com/dotnet/AspNetCore.Docs/blob/6b94e61cd5dc244598e1da70a8128a6998834e73/aspnetcore/blazor/fundamentals/startup.md) | [ASP.NET Core Blazor startup](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/startup?branch=pr-en-us-30922) |
| [aspnetcore/blazor/javascript-interoperability/static-server-rendering.md](https://github.com/dotnet/AspNetCore.Docs/blob/6b94e61cd5dc244598e1da70a8128a6998834e73/aspnetcore/blazor/javascript-interoperability/static-server-rendering.md) | [aspnetcore/blazor/javascript-interoperability/static-server-rendering](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/static-server-rendering?branch=pr-en-us-30922) |

### DIFF
--- a/aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
+++ b/aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
@@ -10,7 +10,11 @@ uid: blazor/js-interop/ssr
 ---
 # ASP.NET Core Blazor JavaScript with Blazor static Server rendering
 
+<!-- UPDATE 9.0 Activate after release and INCLUDE is updated
+
 [!INCLUDE[](~/includes/not-latest-version.md)]
+
+-->
 
 This article explains general concepts on how to interact with JavaScript in a Blazor Web App with static Server rendering.
 
--- a/aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
+++ b/aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
@@ -151,10 +151,10 @@ export function afterWebStarted(blazor) {
 
 For the preceding module:
 
-* In `initializePageScriptModule`:
+* In the `initializePageScriptModule` function:
   * A relative path is normalized by by making it an absolute URL with document's app base path.
-  * If `pageScriptInfo.referenceCount <= 0`, all page-script elements with the same 'src' were unregistered while the module was loading.
-* In `onEnhancedLoad`:
+  * If `pageScriptInfo.referenceCount <= 0`, all `<page-script>` elements with the same `src` were unregistered while the module was loading.
+* In the `onEnhancedLoad` function:
   * Start by invoking `onDispose` on any modules that are no longer referenced.
   * Invoke `onUpdate` on the remaining modules.
 * In the `afterWebStarted` function, `attributeChangedCallback` is used instead of `connectedCallback` because a page script element might be reused between enhanced navigation operations. For more information, see [Using custom elements: Custom element lifecycle callbacks (MDN documentation)](https://developer.mozilla.org/docs/Web/API/Web_components/Using_custom_elements#custom_element_lifecycle_callbacks).
--- a/aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
+++ b/aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
@@ -1,0 +1,134 @@
+---
+title: ASP.NET Core Blazor JavaScript with Blazor static Server rendering
+author: guardrex
+description: Learn how to use JavaScript in a Blazor Web App with static Server rendering.
+monikerRange: '>= aspnetcore-8.0'
+ms.author: riande
+ms.custom: mvc
+ms.date: 11/02/2023
+uid: blazor/js-interop/ssr
+---
+# ASP.NET Core Blazor JavaScript with Blazor static Server rendering
+
+[!INCLUDE[](~/includes/not-latest-version.md)]
+
+This article explains general concepts on how to interact with JavaScript in a Blazor Web App with static Server rendering.
+
+Some applications may depend on JavaScript to perform initialization tasks that are specific to each page. However, when using Blazor's enhanced navigation feature, which allows the user to navigate between pages without reloading the entire page, JavaScript code may not be re-executed as expected. Therefore, page-specific `<script>` elements may not be executed again when the user navigates to a different page.
+
+To avoid this problem, it is not recommended to rely on having page-specific `<script>` elements outside the app's common layout. Instead, scripts should register an `afterWebStarted(blazor)` JavaScript initializer to perform initialization logic, and use `blazor.addEventListener("enhancedload", callback)` to listen to page updates caused by enhanced navigation.
+
+Following is an example demonstrating one way to configure JavaScript code that runs when a page gets initially loaded or updated.
+
+In the Blazor Web App:
+
+* Add the following `MyPage` component.
+* `home.js` JS file.
+
+`Components/Pages/MyPage.razor`:
+
+```razor
+@page "/my-page"
+@using BlazorPageScript
+
+<PageTitle>My page</PageTitle>
+
+<PageScript Src="js/pages/home.js" />
+
+Welcome to my page.
+```
+
+`wwwroot/js/pages/home.js`:
+
+```javascript
+// Called when the page first gets loaded
+export function onLoad() {
+    console.log('Loaded');
+}
+
+// Called each time an enhanced load occurs on the page,
+// plus once when the page first gets loaded
+export function onUpdate() {
+    console.log('Updated');
+}
+
+// Called when the user leaves the page due to an enhanced navigation
+export function onDispose() {
+    console.log('Disposed');
+}
+```
+
+In the BlazorPageScript (Razor Class Library):
+
+`wwwroot/BlazorPageScript.lib.module.js`:
+
+```javascript
+let currentPageModule = null;
+let currentPathname = null;
+
+export function afterWebStarted(blazor) {
+    customElements.define('page-script', class extends HTMLElement {
+        static observedAttributes = ['src'];
+
+        // We use attributeChangedCallback instead of connectedCallback
+        // because a page script element might get reused between enhanced
+        // navigations.
+        attributeChangedCallback(name, oldValue, newValue) {
+            if (name === 'src') {
+                loadPageScript(newValue);
+            }
+        }
+    });
+
+    blazor.addEventListener('enhancedload', onEnhancedLoad);
+}
+
+async function loadPageScript(src) {
+    const pathnameOnLoad = document.location.pathname;
+
+    if (pathnameOnLoad === currentPathname) {
+        // We don't reload the page script if we're already on the same page.
+        return;
+    }
+
+    currentPathname = pathnameOnLoad;
+    currentPageModule?.onDispose?.();
+    currentPageModule = null;
+
+    const module = await import(`${document.location.origin}/${src}`);
+
+    if (location.pathname !== pathnameOnLoad) {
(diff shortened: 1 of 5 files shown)


aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
```diff
@@ -0,0 +1,141 @@
...
+Welcome to my page.
+```
+
+In the Blazor Web App, add the following JS file (`home.js`):
+
+* The `onLoad` function is called when the page loads the first time.
+* The `onUpdate` function is called each time an enhanced load occurs on the page, plus once when the page is first loaded.
+* The `onDispose` function is called when the user leaves the page due to an enhanced navigation.
+
+The `home.js` file is placed in the `wwwroot` folder and not [collocated with the component](xref:blazor/js-interop/index#load-a-script-from-an-external-javascript-file-js-collocated-with-a-component) in order to make it reusable around the app from a common location.
```
MackinnonBuck, line 48: It could still be a collocated script - in fact it might be preferable that way. This example is meant to demonstrate page-specific initialization logic.

That said, it's an interesting idea to support the case where a script could be reusable between pages. The `<PageScript>` component could be changed so it works in a more generalized manner:

* It calls 'onLoad' when the script gets added to the page
* It calls 'onUpdate' when the script still exists on the page after an enhanced update
* It calls 'onDispose' when the script gets removed from the page after an enhanced update

So it works without regard to current URL. Doing it this way would mean the developer gets more control over the lifetime of the script.

For example, if you put a `<PageScript>` on the top-level of a page, then it the behavior is no different from the way the example is now. But if you put it in the app's layout (making it shared between pages), then it only runs "load" after a full page reload, and "update" when any enhanced page update occurs (including enhanced navigation).

Or, if you want to reuse the same module between pages, but get the "load" and "dispose" callbacks invoked on each page change, you could append a querystring to the end of the script so it's recognized as a different module:

`<PageScript Src="./page.js?counter" />`

I might go ahead and propose a new version of the implementation that has these characteristics.

MackinnonBuck:
Here's a new version of `BlazorPageScript.lib.module.js` with the behavior that I described in [this comment](https://github.com/dotnet/AspNetCore.Docs/pull/30922#discussion_r1382163044).

```js
const pageScriptInfoBySrc = new Map();

function registerPageScriptElement(src) {
    if (!src) {
        throw new Error('Must provide a non-empty value for the "src" attribute.');
    }

    let pageScriptInfo = pageScriptInfoBySrc.get(src);

    if (pageScriptInfo) {
        pageScriptInfo.referenceCount++;
    } else {
        pageScriptInfo = { referenceCount: 1, module: null };
        pageScriptInfoBySrc.set(src, pageScriptInfo);
        initializePageScriptModule(src, pageScriptInfo);
    }
}

function unregisterPageScriptElement(src) {
    if (!src) {
        return;
    }

    const pageScriptInfo = pageScriptInfoBySrc.get(src);
    if (!pageScriptInfo) {
        return;
    }

    pageScriptInfo.referenceCount--;
}

async function initializePageScriptModule(src, pageScriptInfo) {
    // If the path is relative, normalize it by by making it an absolute URL
    // with document's the base HREF.
    if (src.startsWith("./")) {
        src = new URL(src.substr(2), document.baseURI).toString();
    }

    const module = await import(src);

    if (pageScriptInfo.referenceCount <= 0) {
        // All page-script elements with the same 'src' were
        // unregistered while we were loading the module.
        return;
    }

    pageScriptInfo.module = module;
    module.onLoad?.();
    module.onUpdate?.();
}

function onEnhancedLoad() {
    // Start by invoking 'onDispose' on any modules that are no longer referenced.
    for (const [src, { module, referenceCount }] of pageScriptInfoBySrc) {
        if (referenceCount <= 0) {
            module?.onDispose?.();
            pageScriptInfoBySrc.delete(src);
        } 
    }

    // Then invoke 'onUpdate' on the remaining modules.
    for (const { module } of pageScriptInfoBySrc.values()) {
        module?.onUpdate?.();
    }
}

export function afterWebStarted(blazor) {
    customElements.define('page-script', class extends HTMLElement {
        static observedAttributes = ['src'];

        // We use attributeChangedCallback instead of connectedCallback
        // because a page-script element might get reused between enhanced
        // navigations.
        attributeChangedCallback(name, oldValue, newValue) {
            if (name !== 'src') {
                return;
            }

            this.src = newValue;
            unregisterPageScriptElement(oldValue);
            registerPageScriptElement(newValue);
        }

        disconnectedCallback() {
            unregisterPageScriptElement(this.src);
        }
    });

    blazor.addEventListener('enhancedload', onEnhancedLoad);
}

```

MackinnonBuck:
@guardrex If we think this sample is too large to put in the docs, I could post it somewhere else and we could link to it.

guardrex:
No ... it's fine. Give me a few minutes, and I'll have the commit up for a look :eyes:.

aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
guardrex: > It could still be a collocated script - in fact it might be preferable that way.

This was one of my *incorrect* guesses, due to you not colocating the script. It will be fixed up on my next commit.

guardrex:
@MackinnonBuck ... I didn't finish it because I can see that this isn't the way that we normally guide devs to load modules. I didn't update that aspect in the `PageWithScript` component ... I wanted to hear back from you on it first.

guardrex:
Note that I always need to move code comments into the text to get them localized. If they stay in the code, they remain stuck in English forever. Just wanted to let you know why you always see me do that.

guardrex:
Ok ... I'm done with little nits.

Let me know on the module load situation with `PageScript` ... if what you have here is what you want ... or do we make it like we normally advise ... `OnAfterRenderAsync`/`JS.InvokeAsync`/`import`/path to module.

aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
```diff
@@ -65,6 +63,12 @@ export function onDispose() {
...
+  * If `pageScriptInfo.referenceCount <= 0`, all page-script elements with the same 'src' were unregistered while the module was loading.
+* In `onEnhancedLoad`:
+  * Start by invoking `onDispose` on any modules that are no longer referenced.
+  * Invoke `onUpdate` on the remaining modules.
 * In the `afterWebStarted` function, `attributeChangedCallback` is used instead of `connectedCallback` because a page script element might be reused between enhanced navigation operations. For more information, see [Using custom elements: Custom element lifecycle callbacks (MDN documentation)](https://developer.mozilla.org/docs/Web/API/Web_components/Using_custom_elements#custom_element_lifecycle_callbacks).
 * The `loadPageScript` function is abandoned if the loaded path (`pathnameOnLoad`) is equal to the current path (`currentPathname`) to avoid reloading the page script if the user is already on the same page. The function is also exited if `location.pathname` isn't equal to the loaded path (`pathnameOnLoad`) because the user changed pages since the module started loading, so there's nothing left to do.
```
MackinnonBuck: This all look accurate, thanks for putting the effort in to document it!

I am wondering how necessary it is to include all this information though. For example, in [this sample](https://learn.microsoft.com/aspnet/core/blazor/fundamentals/signalr?view=aspnetcore-7.0#automatically-refresh-the-page-when-server-side-reconnection-fails), we don't go very in depth about what's happening (although maybe that sample is a bit more straightforward to understand).

MackinnonBuck: Nevermind, saw that you removed it!

aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
```diff
@@ -0,0 +1,186 @@
...
+
+For the preceding module:
+
+* In the `initializePageScriptModule` function:
+  * A relative path is normalized by by making it an absolute URL with document's app base path.
+  * If `pageScriptInfo.referenceCount <= 0`, all `<page-script>` elements with the same `src` were unregistered while the module was loading.
+* In the `onEnhancedLoad` function:
+  * Start by invoking `onDispose` on any modules that are no longer referenced.
+  * Invoke `onUpdate` on the remaining modules.
+* In the `afterWebStarted` function, `attributeChangedCallback` is used instead of `connectedCallback` because a page script element might be reused between enhanced navigation operations. For more information, see [Using custom elements: Custom element lifecycle callbacks (MDN documentation)](https://developer.mozilla.org/docs/Web/API/Web_components/Using_custom_elements#custom_element_lifecycle_callbacks).
```
MackinnonBuck, line 160: Are there other examples in the docs where we go in this much detail about a code sample? I think this makes sense when the sample demonstrates a framework feature, but I'd imagine this is just code people are mostly going to copy and paste. For example, [this sample](https://learn.microsoft.com/aspnet/core/blazor/fundamentals/signalr?view=aspnetcore-7.0#automatically-refresh-the-page-when-server-side-reconnection-fails) is somewhat lengthy but we don't comment much about what it's doing.

MackinnonBuck (approved):
Looks great to me after the pending feedback is addressed 🙂

aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
guardrex: Only placed because you had them in the code sample, where they don't get localized. I'll strip this bit out now.

guardrex:
I see now that this won't work for general calls because the JS object reference is required for JS interop calls. ***DUH!*** 🙈😄 I merely detected a difference without thinking it through last night.

I'll cross-link this as an example in the JS initializers section. We have a few other cross-linked use cases there.

guardrex:
I'm going to hold here for a sec so that you can confirm the updates on the last couple of commits are correct.

PR #30964, opened 2023-11-07
Correct KeyVaultSecretManager documentation

### LINKED ISSUE #26714: Documentation for Key Vault configuration is wrong
Documentation states,

https://github.com/dotnet/AspNetCore.Docs/blob/642a1e53966fe9535fbf6ae870e29ffe119a39e4/aspnetcore/security/key-vault-configuration.md?plain=1#L339

This is now how `KeyVaultSecretManager` is implemented currently, and based on [my comments](https://github.com/Azure/azure-sdk-for-net/issues/28789#issuecomment-1127916189) I think we should change the implementation little to none. Expired secrets may still be necessary to, say, decrypt (if a private or shared key), but I could see excluding disabled secrets since those are explicitly disabled.

/cc @JoshLove-msft

---
#### Document Details

* ID: 2cc0b77a-df21-7007-5cec-4399d98c6dc6
* Version Independent ID: efe5ae5e-ff64-f1c7-6920-4512fc8f36d1
* Content: [Azure Key Vault configuration provider in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/key-vault-configuration?view=aspnetcore-6.0)
* Content Source: [aspnetcore/security/key-vault-configuration.md](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/security/key-vault-configuration.md)
* Product: **aspnet-core**
* Microsoft Alias: **riande**

heaths:
Resolves #26714 

While disabled secrets are listed, values cannot be retrieved. Also updated text to use the Key Vault product name correctly, or just "vault" where it's mentioned as a resource. Sections for >= net6.0 and < net6.0 were also made mostly congruent, sans sample code / references.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/security/key-vault-configuration.md](https://github.com/dotnet/AspNetCore.Docs/blob/4cdab733121cbedf7c4187df738d6d1b8ee241f0/aspnetcore/security/key-vault-configuration.md) | [Azure Key Vault configuration provider in ASP.NET Core](https://review.learn.microsoft.com/en-us/aspnet/core/security/key-vault-configuration?branch=pr-en-us-30964) |

### DIFF
--- a/aspnetcore/security/key-vault-configuration.md
+++ b/aspnetcore/security/key-vault-configuration.md
@@ -12,10 +12,12 @@ uid: security/key-vault-configuration
 
 :::moniker range=">= aspnetcore-6.0"
 
-This article explains how to use the [Azure Key Vault](/azure/key-vault/) configuration provider to load app configuration values from Azure Key Vault secrets. Azure Key Vault is a cloud-based service that helps safeguard cryptographic keys and secrets used by apps and services. Common scenarios for using Azure Key Vault with ASP.NET Core apps include:
+This article explains how to use the [Azure Key Vault](https://azure.microsoft.com/services/key-vault/) configuration provider to load app configuration values from Azure Key Vault secrets. Azure Key Vault is a cloud-based service that helps safeguard cryptographic keys and secrets used by apps and services. Common scenarios for using Azure Key Vault with ASP.NET Core apps include:
 
 * Controlling access to sensitive configuration data.
 * Meeting the requirement for FIPS 140-2 Level 2 validated Hardware Security Modules (HSMs) when storing configuration data.
+
+[View or download sample code](https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/security/key-vault-configuration/samples) ([how to download](xref:index#how-to-download-a-sample))
 
 ## Packages
 
@@ -26,7 +28,7 @@ Add package references for the following packages:
 
 ## Sample app
 
-The [sample app](https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/security/key-vault-configuration/samples) runs in either of two modes determined by the `#define` preprocessor directive at the top of `Program.cs`:
+The sample app runs in either of two modes determined by the `#define` preprocessor directive at the top of `Program.cs`:
 
 * `Certificate`: Demonstrates using an Azure Key Vault Client ID and X.509 certificate to access secrets stored in Azure Key Vault. This sample can be run from any location, whether deployed to Azure App Service or any host that can serve an ASP.NET Core app.
 * `Managed`: Demonstrates how to use [Managed identities for Azure resources](/azure/active-directory/managed-identities-azure-resources/overview). The managed identity authenticates the app to Azure Key Vault with Azure Active Directory (AD) authentication without storing credentials in the app's code or configuration. The `Managed` version of the sample must be deployed to Azure. Follow the guidance in the [Use the managed identities for Azure resources](#use-managed-identities-for-azure-resources) section.
@@ -92,7 +94,7 @@ Complete the following steps to create an Azure Key Vault and store the sample a
 
 1. Create secrets in the vault as name-value pairs.
 
-   Azure Key Vault secret names are limited to alphanumeric characters and dashes. Hierarchical values (configuration sections) use `--` (two dashes) as a delimiter, as colons aren't allowed in vault secret names. Colons delimit a section from a subkey in [ASP.NET Core configuration](xref:fundamentals/configuration/index). The two-dash sequence is replaced with a colon when the secrets are loaded into the app's configuration.
+   Azure Key Vault secret names are limited to alphanumeric characters and dashes. Hierarchical values (configuration sections) use `--` (two dashes) as a delimiter, as colons aren't allowed in Key Vault secret names. Colons delimit a section from a subkey in [ASP.NET Core configuration](xref:fundamentals/configuration/index). The two-dash sequence is replaced with a colon when the secrets are loaded into the app's configuration.
 
    The following secrets are for use with the sample app. The values include a `_prod` suffix to distinguish them from the `_dev` suffix values loaded in the Development environment from Secret Manager. Replace `{KEY VAULT NAME}` with the name of the Key Vault you created in the prior step:
 
@@ -142,7 +144,7 @@ The X.509 certificate is managed by the OS. The app calls <xref:Microsoft.Extens
 
 Example values:
 
-* Key vault name: `contosovault`
+* Key Vault name: `contosovault`
 * Application ID: `627e911e-43cc-61d4-992e-12db9c81b413`
 * Certificate thumbprint: `fe14593dd66b2406c5269d742d04b6e1ab03adb1`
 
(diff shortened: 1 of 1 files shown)


jlichwa (approved):

jlichwa:
Looks good.

Key Vault PG

aspnetcore/security/key-vault-configuration.md
```diff
@@ -12,11 +12,13 @@ uid: security/key-vault-configuration
 
 :::moniker range=">= aspnetcore-6.0"
 
-This article explains how to use the [Azure Key Vault](/azure/key-vault/) configuration provider to load app configuration values from Azure Key Vault secrets. Azure Key Vault is a cloud-based service that helps safeguard cryptographic keys and secrets used by apps and services. Common scenarios for using Azure Key Vault with ASP.NET Core apps include:
+This article explains how to use the [Azure Key Vault](https://azure.microsoft.com/services/key-vault/) configuration provider to load app configuration values from Azure Key Vault secrets. Azure Key Vault is a cloud-based service that helps safeguard cryptographic keys and secrets used by apps and services. Common scenarios for using Azure Key Vault with ASP.NET Core apps include:
```
Rick-Anderson, line 15: Why replace the learn (documentation) page with the marketing page?

aspnetcore/security/key-vault-configuration.md
```diff
@@ -12,11 +12,13 @@ uid: security/key-vault-configuration
...
 
 * Controlling access to sensitive configuration data.
 * Meeting the requirement for FIPS 140-2 Level 2 validated Hardware Security Modules (HSMs) when storing configuration data.
 
+[View or download sample code](https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/security/key-vault-configuration/samples) ([how to download](xref:index#how-to-download-a-sample))
+
```
Rick-Anderson, line 21: OK but I prefer the original where we don't duplicate the very verbose `view or download` before it's needed.

Fine to keep, but I prefer the original
image

cc @tdykstra

Rick-Anderson (approved):
Nice improvements, thanks for the PR.

aspnetcore/security/key-vault-configuration.md
heaths: The other section had the full URL. Not sure which is better. Given it's the first instance Key Vault is really mentioned, seemed like the landing page might be better. @jlichwa which way do you prefer?

Rick-Anderson: Original https://learn.microsoft.com/en-us/azure/key-vault/ (a learn landing page)
your change: https://azure.microsoft.com/en-us/products/key-vault/ : A marketing page

we avoid marketing pages like the plague, but I'll let you decide.

heaths: > we avoid marketing pages like the plague, but I'll let you decide.

If @jlichwa doesn't have any objection as the product PM, I'll stick with that MO then.

Rick-Anderson: The learn page has two prominent and helpful links
image

aspnetcore/security/key-vault-configuration.md
heaths: Most of the article is actually the same, too. Is there an easier way to vary just the code snippets for `< net6.0` vs. `>= net6.0`? There were a few other differences in a few paragraphs but I only briefly reviewed. I didn't want to boil the ocean, but a quick glance shows they really don't substantially vary by version either.

Rick-Anderson: > Most of the article is actually the same, too. Is there an easier way to vary just the code snippets for `< net6.0` vs. `>= net6.0`? There were a few other differences in a few paragraphs but I only briefly reviewed. I didn't want to boil the ocean, but a quick glance shows they really don't substantially vary by version either.

No. Very few customer are on .NET 5 and lower. .NET 6 and .NET 8 are LTS. It's very problematic to try to reduce duplication in sample code. It's prover easier to just make a copy and update it for the latest version.

heaths: I'm fine with samples being separate. I mean, can I share most of the content in the page and only bifurcate the code snippets?

For example, would this work:

````markdown
This is common text for both pages.

:::moniker range=">= aspnetcore-6.0"
<!-- >= 6.0 sample code -->
:::moniker-end

:::moniker range="< aspnetcore-6.0"
<!-- < 5.0 sample code -->
:::moniker-end

More common text for both pages.
````

When I made two separate files from each section and diff'd them, there are almost no differences and those I briefly looked at probably don't need to be different. As I mentioned above: I just didn't want to boil the ocean. I could open a separate PR to make future edits easier, though. Not having to update duplicate text seems less error prone.

Rick-Anderson: > When I made two separate files from each section and diff'd them, there are almost no differences and those I briefly looked at probably don't need to be different

After maintaining an entire doc set for many versions we've found using multiple monikers in a file is hard to maintain and reason about. I'd say 90% of our docs have 97% duplication, and that has proven to be the most productive approach.

- Updating versions (where there are changes) is trivial and robust.
  - copy fileX.md to fileX/includes/filex7.md
  - remove meta data and change moniker from >= 7 to = 7
  - change moniker in fileX.md from >= 7 to >= 8
  - merge PR
  - create new PR with 8 changes. Now the diff file shows exactly what has changed.

We've found that mixing monikers:

- Is so difficult to reason about with professionals MSFT folks that mistakes are often made. 
- Is a significant hurdle for customers to update.

We've found out that:

- Most customers want the latest version. Older versions don't need fixes unless it's security related
- The eliminate duplication goal software folks embrace is highly problematic in docs.

heaths: The file already uses multiple monikers, but those cover the entirety of the content. I would like to reduce maintenance costs by only bifurcating what needs to be different, which is pretty much just the code snippets.

Rick-Anderson: > The file already uses multiple monikers, but those cover the entirety of the content. I would like to reduce maintenance costs by only bifurcating what needs to be different, which is pretty much just the code snippets.

Right, that's because our moniker model has evolved. The current model has proven the most productive, that is moniker/version per include file.

>  I would like to reduce maintenance costs by only bifurcating what needs to be different

We've got enormous experience saying the opposite. Maybe you missed my point we don't update older versions unless theirs a security issue (which is rare).

heaths: > We've got enormous experience saying the opposite. Maybe you missed my point we don't update older versions unless theirs a security issue (which is rare).

In this case, the text that disabled secrets were returned was wrong. You're suggesting I shouldn't update that older docs? I know .NET likes people to be on LTS, but even the Azure SDK targets `netstandard2.0` because a large percentage of customers still target .NET Framework. Having incorrect code even in older docs feels very wrong.

Again, I'm not suggesting trying to unify code snippets. I'm arguing that this:

````markdown
Common text.

:::moniker ">= aspnet-6.0"
newer code snippet
:::moniker-end

:::monker "< aspnet-6.0"
older code snippet
:::moniker-end

More common text
````

Seems better than this:

````markdown
:::moniker ">= aspnet-6.0"
Common text

newer code snippet

More common text
:::moniker-end

:::monker "< aspnet-6.0"
Common text

older code snippet

More common text
:::moniker-end
````

If we have to update "Common text", we do it only in one place vs. two. You're saying experience has shown the latter example is easier to maintain?

At least in the examples I'm seeing here, the only real reason we have different code snippets is because the configuration APIs were improved in latter versions and we can show more concise code.

Azure SDKs have a policy that the public API surface has to remain the same regardless of underlying support meaning, for example, we may have to throw a `NotSupportedException` against some targets or platforms, but the public API surface will still be there unless there's a good reason it can't be e.g., returns or accepts a new type that didn't previously exist - and even then, we try to abstract away those details if at all possible. So our "Common text" and "More common text" in the examples above will always be the same.

PR #31027, opened 2023-11-14
Fix up Redis Output Caching doc

eerhardt:
- Point to the correct nupkg
- Fix the method call


tdykstra (approved):
LGTM Thanks, @eerhardt.

PR #31057, opened 2023-11-16
Environments + client services w/prerendering

### LINKED ISSUE #31025: Environments and client services with prerendering 8.0
### Description

Check all of the scenarios in BWAs and adjust guidance as needed.

This work is going to be expanded to not only address the problems in the *Environments* article but also to call out the challenges with the need for client side services that don't exist server-side when client-side components prerender on the server.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/environments?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/fundamentals/environments.md

### Document ID

8a5b47e7-b29c-b049-30e8-5b405a5086b6

### Article author

guardrex

guardrex:
Fixes #31025 
Addresses #28161

Mackinnon ... I decided to draft this with a full explanation and examples for what's going on with client-side services in a BWA. It's a little long. However, I can understand it this way, so we have a good bet on not losing anyone.

In the *Render Modes* article, the *Set the render mode for the entire app* section was too far down the article. I moved the section up and changed the heading name to *Apply a render mode to the entire app*. The new section on *Client-side services fail to resolve during prerendering* makes reference to globally setting the render mode. I want to reader to have already seen that section.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/76f68fd99e58275e5ea18a1e16d15593dd1c196a/aspnetcore/blazor/components/index.md) | [ASP.NET Core Razor components](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/index?branch=pr-en-us-31057) |
| [aspnetcore/blazor/components/render-modes.md](https://github.com/dotnet/AspNetCore.Docs/blob/76f68fd99e58275e5ea18a1e16d15593dd1c196a/aspnetcore/blazor/components/render-modes.md) | [ASP.NET Core Blazor render modes](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/render-modes?branch=pr-en-us-31057) |
| [aspnetcore/blazor/fundamentals/configuration.md](https://github.com/dotnet/AspNetCore.Docs/blob/76f68fd99e58275e5ea18a1e16d15593dd1c196a/aspnetcore/blazor/fundamentals/configuration.md) | [ASP.NET Core Blazor configuration](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/configuration?branch=pr-en-us-31057) |
| [aspnetcore/blazor/fundamentals/environments.md](https://github.com/dotnet/AspNetCore.Docs/blob/76f68fd99e58275e5ea18a1e16d15593dd1c196a/aspnetcore/blazor/fundamentals/environments.md) | [aspnetcore/blazor/fundamentals/environments](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/environments?branch=pr-en-us-31057) |
| [aspnetcore/blazor/security/server/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/76f68fd99e58275e5ea18a1e16d15593dd1c196a/aspnetcore/blazor/security/server/index.md) | [Secure ASP.NET Core server-side Blazor apps](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/server/index?branch=pr-en-us-31057) |
| [aspnetcore/blazor/state-management.md](https://github.com/dotnet/AspNetCore.Docs/blob/76f68fd99e58275e5ea18a1e16d15593dd1c196a/aspnetcore/blazor/state-management.md) | [ASP.NET Core Blazor state management](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/state-management?branch=pr-en-us-31057) |

### DIFF
--- a/aspnetcore/blazor/components/render-modes.md
+++ b/aspnetcore/blazor/components/render-modes.md
@@ -155,7 +155,7 @@ For apps based on the Blazor Web App project template, a render mode assigned to
 <Routes @rendermode="InteractiveServer" />
 ```
 
-The Blazor router propagates its render mode to the pages it routes. The pages aren't technically child components of the router, but when the routes are discovered at runtime for the router, they inherit the router's render mode.
+The Blazor router propagates its render mode to the pages it routes.
 
 You also typically must set the same interactive render mode on the `HeadOutlet` component, which is also found in the `App` component of a Blazor Web App generated from the project template:
 
@@ -320,22 +320,22 @@ For example, consider the following `Home` component in the `.Client` project in
 
 ```razor
 @page "/"
-@inject IBlazorHostEnvironment Environment
+@inject IWebAssemblyHostEnvironment Environment
 
 <PageTitle>Home</PageTitle>
 
 <h1>Home</h1>
 
 <p>
-    Environment: @Environment.EnvironmentName
+    Environment: @Environment.Environment
 </p>
 ```
 
 No compile time error occurs, but a runtime error occurs during prerendering:
 
-> :::no-loc text="Cannot provide a value for property 'HostEnvironment' on type 'BlazorWebApp80EnvironmentTesting.Client.Pages.Home'. There is no registered service of type 'Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment'.":::
+> :::no-loc text="Cannot provide a value for property 'Environment' on type 'BlazorWebApp80EnvironmentTesting.Client.Pages.Home'. There is no registered service of type 'Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment'.":::
 
-This error occurs because the component must compile and execute on the server during prerendering but <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> isn't a registered service and has no implementation on on the server.
+This error occurs because the component must compile and execute on the server during prerendering, but <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> isn't a registered service on the server.
 
 There are a four approaches that you can take to address this scenario. The following are listed from most recommended to least recommended:
 
--- a/aspnetcore/blazor/components/render-modes.md
+++ b/aspnetcore/blazor/components/render-modes.md
@@ -142,6 +142,34 @@ Technically, `@rendermode` is both a Razor *directive* and a Razor *directive at
 > [!NOTE]
 > Component authors should avoid coupling a component's implementation to a specific render mode. Instead, component authors should typically design components to support any render mode or hosting model. A component's implementation should avoid assumptions on where it's running (server or client) and should degrade gracefully when rendered statically. Specifying the render mode in the component definition may be needed if the component isn't instantiated directly (such as with a routable page component) or to specify a render mode for all component instances.
 
+## Apply a render mode to the entire app
+
+To set the render mode for the entire app, indicate the render mode at the highest-level interactive component in the app's component hierarchy that isn't a root component.
+
+> [!NOTE]
+> Making a root component interactive, such as the `App` component, isn't supported because the Blazor script may be evaluated multiple times. Therefore, the render mode for the entire app can't be set directly by the `App` component.
+
+For apps based on the Blazor Web App project template, a render mode assigned to the entire app is typically specified where the `Routes` component is used in the `App` component (`Components/App.razor`):
+
+```razor
+<Routes @rendermode="InteractiveServer" />
+```
+
+The Blazor router propagates its render mode to the pages it routes. The pages aren't technically child components of the router, but when the routes are discovered at runtime for the router, they inherit the router's render mode.
+
+You also typically must set the same interactive render mode on the `HeadOutlet` component, which is also found in the `App` component of a Blazor Web App generated from the project template:
+
+```
+<HeadOutlet @rendermode="InteractiveServer" />
+```
+
+To enable global interactivity when creating a Blazor Web App:
+
+* Visual Studio: Set the **Interactivity location** dropdown list to **Global**.
+* .NET CLI: Use the `-ai|--all-interactive` option.
+
+For more information, see <xref:blazor/tooling>.
+
 ## Prerendering
 
 *Prerendering* is the process of initially rendering page content on the server without enabling event handlers for rendered controls. The server outputs the HTML UI of the page as soon as possible in response to the initial request, which makes the app feel more responsive to users. Prerendering can also improve [Search Engine Optimization (SEO)](https://developer.mozilla.org/docs/Glossary/SEO) by rendering content for the initial HTTP response that search engines use to calculate page rank.
@@ -283,6 +311,43 @@ In the following example, the component is interactive throughout the process. T
 ```
 
 If using the preceding component locally in a Blazor Web App, place the component in the client project's `Pages` folder. The client project is the solution's project with a name that ends in `.Client`. When the app is running, navigate to `/render-mode-4` in the browser's address bar.
+
+## Client-side services fail to resolve during prerendering
+
+Assuming that prerendering isn't disabled for a component or for the app, a component in the `.Client` project is prerendered on the server. Because the server doesn't have access to registered client-side Blazor services, it isn't possible to inject these services into a component without receiving an error that the service can't be found during prerendering.
+
+For example, consider the following `Home` component in the `.Client` project in a Blazor Web App with [global Interactive WebAssembly or Interactive Auto rendering](#apply-a-render-mode-to-the-entire-app). The component attempts to inject <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> to obtain the environment's name.
+
+```razor
+@page "/"
+@inject IBlazorHostEnvironment Environment
+
+<PageTitle>Home</PageTitle>
+
+<h1>Home</h1>
+
+<p>
+    Environment: @Environment.EnvironmentName
+</p>
+```
+
+No compile time error occurs, but a runtime error occurs during prerendering:
+
+> :::no-loc text="Cannot provide a value for property 'HostEnvironment' on type 'BlazorWebApp80EnvironmentTesting.Client.Pages.Home'. There is no registered service of type 'Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment'.":::
+
+This error occurs because the component must compile and execute on the server during prerendering but <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> isn't a registered service and has no implementation on on the server.
+
+There are a four approaches that you can take to address this scenario. The following are listed from most recommended to least recommended:
+
+* *Recommended*: Create a service abstraction and create implementations&dagger; for the service in the `.Client` and server projects. Register the services in each project. Inject the service in the component. For a demonstration of this approach, see <xref:blazor/fundamentals/environments#read-the-client-side-environment-in-a-blazor-web-app>.
+
+  &dagger;One implementation can be created if the implementation can be configured for execution in the `.Client` and server projects.
+
+* Add a check for <xref:System.OperatingSystem.IsBrowser%2A?displayProperty=nameWithType> in the [`OnInitialized{Async}` lifecycle method](xref:blazor/components/lifecycle#component-initialization-oninitializedasync) and use the check to determine whether to obtain a service instance from an injected <xref:System.IServiceProvider>.
+
+* Similar to the preceding approach that works during prerendering, you might be able to add a `.Client` project package reference to a server-side package and fall back to using the server-side API when prerendering on the server.
+
+* Disable prerendering for the component. For more information, see the [Prerendering](#prerendering) section.
 
 ## Render mode propagation
 
@@ -433,37 +498,9 @@ The following component results in a runtime error when the component is rendere
 
 > :::no-loc text="Cannot create a component of type 'BlazorSample.Components.SharedMessage' because its render mode 'Microsoft.AspNetCore.Components.Web.InteractiveWebAssemblyRenderMode' is not supported by Interactive Server rendering.":::
 
-## Set the render mode for the entire app
-
-To set the render mode for the entire app, indicate the render mode at the highest-level interactive component in the app's component hierarchy that isn't a root component.
-
-> [!NOTE]
-> Making a root component interactive, such as the `App` component, isn't supported because the Blazor script may be evaluated multiple times. Therefore, the render mode for the entire app can't be set directly by the `App` component.
-
-For apps based on the Blazor Web App project template, a render mode assigned to the entire app is typically specified where the `Routes` component is used in the `App` component (`Components/App.razor`):
-
-```razor
-<Routes @rendermode="InteractiveServer" />
-```
-
-The Blazor router propagates its render mode to the pages it routes. The pages aren't technically child components of the router, but when the routes are discovered at runtime for the router, they inherit the router's render mode.
-
-You also typically must set the same interactive render mode on the `HeadOutlet` component, which is also found in the `App` component of a Blazor Web App generated from the project template:
-
-```
-<HeadOutlet @rendermode="InteractiveServer" />
-```
-
-To enable global interactivity when creating a Blazor Web App:
-
-* Visual Studio: Set the **Interactivity location** dropdown list to **Global**.
-* .NET CLI: Use the `-ai|--all-interactive` option.
-
-For more information, see <xref:blazor/tooling>.
-
 ## Discover components from additional assemblies for Static Server rendering
 
-Configure additional assemblies to use for discovering routable Razor components for Static Server rendering using the `AddAdditionalAssemblies` method chained to <xref:Microsoft.AspNetCore.Builder.RazorComponentsEndpointRouteBuilderExtensions.MapRazorComponents%2A>.
+Configure additional assemblies to discover routable Razor components for Static Server rendering using the `AddAdditionalAssemblies` method chained to <xref:Microsoft.AspNetCore.Builder.RazorComponentsEndpointRouteBuilderExtensions.MapRazorComponents%2A>.
 
 The following example includes the assembly of the `DifferentAssemblyCounter` component:
 
--- a/aspnetcore/blazor/components/render-modes.md
+++ b/aspnetcore/blazor/components/render-modes.md
@@ -339,7 +339,7 @@ This error occurs because the component must compile and execute on the server d
 
 There are a four approaches that you can take to address this scenario. The following are listed from most recommended to least recommended:
 
-* *Recommended*: Create a service abstraction and create implementations for the service in the `.Client` and server projects. Register the services in each project. Inject the service in the component. For a demonstration of this approach, see <xref:blazor/fundamentals/environments#read-the-client-side-environment-in-a-blazor-web-app>.
+* *Recommended*: Create a service abstraction and create implementations for the service in the `.Client` and server projects. Register the services in each project. Inject the service in the component. For a demonstration of this approach, see <xref:blazor/fundamentals/environments#read-the-environment-in-a-blazor-web-app>.
 
 * Add a check for <xref:System.OperatingSystem.IsBrowser%2A?displayProperty=nameWithType> in the [`OnInitialized{Async}` lifecycle method](xref:blazor/components/lifecycle#component-initialization-oninitializedasync) and use the check to determine whether to obtain a service instance from an injected <xref:System.IServiceProvider>.
 
--- a/aspnetcore/blazor/components/render-modes.md
+++ b/aspnetcore/blazor/components/render-modes.md
@@ -333,7 +333,7 @@ For example, consider the following `Home` component in the `.Client` project in
 
 No compile time error occurs, but a runtime error occurs during prerendering:
 
-> :::no-loc text="Cannot provide a value for property 'Environment' on type 'BlazorWebApp80EnvironmentTesting.Client.Pages.Home'. There is no registered service of type 'Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment'.":::
+> :::no-loc text="Cannot provide a value for property 'Environment' on type 'BlazorWebAppSample.Client.Pages.Home'. There is no registered service of type 'Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment'.":::
 
 This error occurs because the component must compile and execute on the server during prerendering, but <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> isn't a registered service on the server.
 
--- a/aspnetcore/blazor/components/render-modes.md
+++ b/aspnetcore/blazor/components/render-modes.md
@@ -337,15 +337,43 @@ No compile time error occurs, but a runtime error occurs during prerendering:
 
 This error occurs because the component must compile and execute on the server during prerendering, but <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> isn't a registered service on the server.
 
+If the app doesn't require the value during prerendering, this problem can be solved by injecting <xref:System.IServiceProvider> to obtain the service instead of the service type itself:
+
+```razor
+@page "/"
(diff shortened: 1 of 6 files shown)


aspnetcore/blazor/fundamentals/environments.md
```diff
@@ -135,57 +152,208 @@ In the following example for IIS, the custom header (`Blazor-Environment`) is ad
...
+```csharp
+builder.Services.AddSingleton<IBlazorHostEnvironment>(
+    sp => new BlazorHostEnvironment(builder.Environment.EnvironmentName));
+```
+
+In the `.Client` project's `Program` file, register a `IBlazorHostEnvironment` service with the `BlazorHostEnvironment` implementation. Set the host environment with <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment.Environment%2A?displayProperty=nameWithType>, which is only available on the client:
+
+```csharp
+builder.Services.AddSingleton<IBlazorHostEnvironment>(
+    sp => new BlazorHostEnvironment(builder.HostEnvironment.Environment));
```
MackinnonBuck, line 297: This is definitely valid, albeit different than what I had in mind.

If the customer were to add additional members to the `IBlazorHostEnvironment` interface, it would require adding more parameters to the primary constructor and providing values in `Program.cs`. Or if the interface had any methods in it, it would require falling back to separate implementations (unless you do a thing where you dynamically check which platform you're on using `OperatingSystem.IsBrowser()`).

I think it would be a bit more conventional to have separate implementations of `IBlazorHostEnvironment` that can inject app-specific dependencies and use those to support aspects of the implementation. For example:

`.Client` project:
```csharp
// IBlazorHostEnvironment.cs
public interface IBlazorHostEnvironment
{
    string EnvironmentName { get; }
}

// BlazorWebAssemblyHostEnvironment.cs
public class BlazorWebAssemblyHostEnvironment(IWebAssemblyHostEnvironment hostEnvironment) : IBlazorHostEnvironment
{
    public string EnvironmentName => hostEnvironment.Environment;
}

// Program.cs
builder.Services.AddSingleton<IBlazorHostEnvironment, BlazorWebAssemblyHostEnvironment>();
```

Server project:
```csharp
// BlazorServerHostEnvironment.cs
public class BlazorServerHostEnvironment(IWebHostEnvironment hostEnvironment) : IBlazorHostEnvironment
{
    public string EnvironmentName => hostEnvironment.EnvironmentName;
}

// Program.cs
builder.Services.AddScoped<IBlazorHostEnvironment, BlazorServerHostEnvironment>();
```

What do you think?

aspnetcore/blazor/components/render-modes.md
```diff
@@ -284,6 +312,43 @@ In the following example, the component is interactive throughout the process. T
...
+
+> :::no-loc text="Cannot provide a value for property 'HostEnvironment' on type 'BlazorWebApp80EnvironmentTesting.Client.Pages.Home'. There is no registered service of type 'Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment'.":::
+
+This error occurs because the component must compile and execute on the server during prerendering but <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> isn't a registered service and has no implementation on on the server.
+
+There are a four approaches that you can take to address this scenario. The following are listed from most recommended to least recommended:
+
+* *Recommended*: Create a service abstraction and create implementations&dagger; for the service in the `.Client` and server projects. Register the services in each project. Inject the service in the component. For a demonstration of this approach, see <xref:blazor/fundamentals/environments#read-the-client-side-environment-in-a-blazor-web-app>.
+
+  &dagger;One implementation can be created if the implementation can be configured for execution in the `.Client` and server projects.
```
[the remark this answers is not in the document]
MackinnonBuck: Update: I now see what you mean by looking at your example - I left a comment down there 🙂

MackinnonBuck (commented):
Looks like a great start - I have some ideas and thoughts!

aspnetcore/blazor/components/render-modes.md
```diff
@@ -142,6 +142,34 @@ Technically, `@rendermode` is both a Razor *directive* and a Razor *directive at
...
+> [!NOTE]
+> Making a root component interactive, such as the `App` component, isn't supported because the Blazor script may be evaluated multiple times. Therefore, the render mode for the entire app can't be set directly by the `App` component.
+
+For apps based on the Blazor Web App project template, a render mode assigned to the entire app is typically specified where the `Routes` component is used in the `App` component (`Components/App.razor`):
+
+```razor
+<Routes @rendermode="InteractiveServer" />
+```
+
+The Blazor router propagates its render mode to the pages it routes. The pages aren't technically child components of the router, but when the routes are discovered at runtime for the router, they inherit the router's render mode.
```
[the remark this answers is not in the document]
guardrex: It's probably ok to 🔪 that. IIRC, that was one of Dan's remarks.

aspnetcore/blazor/components/render-modes.md
```diff
@@ -284,6 +312,43 @@ In the following example, the component is interactive throughout the process. T
...
 
+## Client-side services fail to resolve during prerendering
+
+Assuming that prerendering isn't disabled for a component or for the app, a component in the `.Client` project is prerendered on the server. Because the server doesn't have access to registered client-side Blazor services, it isn't possible to inject these services into a component without receiving an error that the service can't be found during prerendering.
+
+For example, consider the following `Home` component in the `.Client` project in a Blazor Web App with [global Interactive WebAssembly or Interactive Auto rendering](#apply-a-render-mode-to-the-entire-app). The component attempts to inject <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> to obtain the environment's name.
+
+```razor
+@page "/"
+@inject IBlazorHostEnvironment Environment
```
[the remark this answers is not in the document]
guardrex: Yes ... I was going toooooo fast. Normally, I'd sleep on something this big and fix 😈 bugs in it on the next day.

aspnetcore/blazor/components/render-modes.md
```diff
@@ -284,6 +312,43 @@ In the following example, the component is interactive throughout the process. T
...
+
+<p>
+    Environment: @Environment.EnvironmentName
+</p>
+```
+
+No compile time error occurs, but a runtime error occurs during prerendering:
+
+> :::no-loc text="Cannot provide a value for property 'HostEnvironment' on type 'BlazorWebApp80EnvironmentTesting.Client.Pages.Home'. There is no registered service of type 'Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment'.":::
+
```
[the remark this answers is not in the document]
guardrex: Yes ... that was a late change to the example that I didn't spot here.

aspnetcore/blazor/components/render-modes.md
```diff
@@ -284,6 +312,43 @@ In the following example, the component is interactive throughout the process. T
...
+<h1>Home</h1>
+
+<p>
+    Environment: @Environment.EnvironmentName
+</p>
+```
+
+No compile time error occurs, but a runtime error occurs during prerendering:
+
+> :::no-loc text="Cannot provide a value for property 'HostEnvironment' on type 'BlazorWebApp80EnvironmentTesting.Client.Pages.Home'. There is no registered service of type 'Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment'.":::
```
guardrex, line 336:
```suggestion
> :::no-loc text="Cannot provide a value for property 'Environment' on type 'BlazorWebApp80EnvironmentTesting.Client.Pages.Home'. There is no registered service of type 'Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment'.":::
```

aspnetcore/blazor/fundamentals/environments.md
```diff
@@ -135,57 +152,208 @@ In the following example for IIS, the custom header (`Blazor-Environment`) is ad
...
+Create a project reference for the class library in both the server and `.Client` projects:
 
-:::code language="razor" source="~/../blazor-samples/3.1/BlazorSample_WebAssembly/Pages/environments/ReadEnvironment.razor" highlight="3,7":::
+* Visual Studio: Right-click each project and select **Add** > **Project Reference**. Select the checkbox for the shared project and select **OK**.
+* Visual Studio Code/.NET CLI: Execute `dotnet add reference {PATH}` in a command shell from each of the server and `.Client` project folders. The `{PATH}` placeholder is the path to the shared class library project.
+
+In the shared (`.Shared`) project:
+
+* Add an interface for the service abstraction.
+* Add classes for the service implementation and host environment extensions.
```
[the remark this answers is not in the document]
guardrex: Yes, I was split in my mind on if that felt good, and that is what you said in the email.

I just narrowly was leaning in the direction of portability, using a class lib or package across BWAs. Also, it feels like we're dealing with two distinct projects that are sharing a central resource, even if that isn't the best way to think about it.

However, it would be simpler to explain that way ... in the `.Client` project, so let's shorten this up with that approach. I'll make the changes.

aspnetcore/blazor/fundamentals/environments.md
```diff
@@ -135,57 +152,208 @@ In the following example for IIS, the custom header (`Blazor-Environment`) is ad
...
+    public void Dispose()
+    {
+        if (_disposed)
+        {
+            return;
+        }
+
+        _disposed = true;
+    }
+}
```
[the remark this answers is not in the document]
guardrex: I'll answer on this below.

aspnetcore/blazor/fundamentals/environments.md
```diff
@@ -135,57 +152,208 @@ In the following example for IIS, the custom header (`Blazor-Environment`) is ad
...
+    public static bool IsEnvironment(
+        this IBlazorHostEnvironment hostEnvironment,
+        string environmentName)
+    {
+        return string.Equals(
+            hostEnvironment.EnvironmentName,
+            environmentName,
+            StringComparison.OrdinalIgnoreCase);
+    }
+}
```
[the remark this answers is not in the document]
guardrex: Well ... it's kind'a nice to get the methods. I wrote this so that devs can just use it just like this ... save them some time. It's fast to delete the bits they don't want. It's slower (and a pain) for them to have to create these.

aspnetcore/blazor/fundamentals/environments.md
guardrex: I did it that way because I felt there probably wouldn't be anything else, but I think the most general approach is what you're suggesting. Since we'll use this as a promonient example of how to address this problem, I think your idea is best.

WRT the dispose situation from above. I'm confused if the DI container is creating the `new`-ed up services like this. I thought that it was NOT creating them ... that the dev code is creating them. If that's right, then the disposal code is boilerplate code that doesn't dispose of anything but perhaps present as a good pattern so as to be clear that the service implementation needs to dispose of anything it creates ... just as a reminder ... a pattern. Alternatively if I'm incorrect about this service registration and what's actually happening is that the DI container is creating them, then yeah ... for sure there's no need to place this boilerplate do-nothing disposal code ... such as what we'll be changing over to when I adopt your suggestion here for the serivces.

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-8.0#disposal-of-services

aspnetcore/blazor/components/render-modes.md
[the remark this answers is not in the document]
guardrex: I'll fix this on the next commit when I update the example approach.

guardrex:
@MackinnonBuck ... I think I have most of the ....

# 🦖 ***RexHacks!***&trade; 🦖

... out of the PR. 🙈 ... but you can see why I like to get these technical reviews. You never know what kind of disgusting 💩 code a 🦖 will poop out into the doc! 😆

I shortened the API naming quite a bit to get everything in under our 85 character line length for doc code. It doesn't have a collision with `Microsoft.Extensions.Hosting.IHostEnvironment` AFAICT.

guardrex:
... and I changed the section name in the *Environments* topic ...

```diff
Read the client-side environment in a Blazor Web App
Read the environment in a Blazor Web App
```

... because the environment is being read in both and this has broader implications because it also applies to Auto components.

aspnetcore/blazor/fundamentals/environments.md
```diff
@@ -135,57 +152,150 @@ In the following example for IIS, the custom header (`Blazor-Environment`) is ad
...
 
-:::moniker-end
+* Add an interface for the service abstraction.
+* Add a service implementation that obtains the host environment from <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment.Environment%2A?displayProperty=nameWithType>.
 
-:::moniker range="< aspnetcore-5.0"
+`IHostEnvironment.cs`:
+
+```csharp
+public interface IHostEnvironment
```
MackinnonBuck, line 214: The `IHostEnvironment` name might clash with [`Microsoft.Extensions.Hosting.IHostEnvironment`](https://source.dot.net/#Microsoft.Extensions.Hosting.Abstractions/IHostEnvironment.cs,37d39d4dc6955525). Do you think we could find another name for it? Maybe `IAppHostEnvironment` or `IBlazorHostEnvironment`?

aspnetcore/blazor/fundamentals/environments.md
```diff
@@ -135,57 +152,150 @@ In the following example for IIS, the custom header (`Blazor-Environment`) is ad
...
+{
+    string EnvironmentName { get; set; }
```
MackinnonBuck, line 216: I think we could make this just `{ get; }`

aspnetcore/blazor/fundamentals/environments.md
```diff
@@ -135,57 +152,150 @@ In the following example for IIS, the custom header (`Blazor-Environment`) is ad
...
+```
+
+`WasmHostEnvironment.cs`:
+
+```csharp
+using Microsoft.AspNetCore.Components.WebAssembly.Hosting;
+
+public class WasmHostEnvironment(IWebAssemblyHostEnvironment env) : IHostEnvironment
+{
+    public string EnvironmentName { get; set; } = env.Environment;
```
MackinnonBuck, line 227: Could this be a computed property (using `=>` rather than `{ get; set; } =`)? Same suggestion for `ServerHostEnvironment`. I only suggest this because `IWebHostEnvironment` has a mutable `EnvironmentName` property, and we would want the `EnvironmentName` property here to reflect any changes to the underlying environment name.

MackinnonBuck (commented):
Looking good! I do also want to hear an opinion from @javiercn or @SteveSandersonMS before we officially suggest this approach to customers.

aspnetcore/blazor/fundamentals/environments.md
guardrex: Yes ... I just need to keep it short. It wasn't colliding in my test app, but I agree that this is a bit scary!

aspnetcore/blazor/fundamentals/environments.md
guardrex: I thought IntelliSense was complaining when I tried that, but ***No!*** It's cool. I'll make the update.

aspnetcore/blazor/fundamentals/environments.md
guardrex: Yes ... again, it looked like at first that the compiler was going to choke on it. ***No!*** The compiler is good. I'll make the change. 👍

guardrex:
@MackinnonBuck ... Looks good, but can we use `IAppEnvironment`? Trying to keep it short to keep the code line lengths at 85 or less chars.

MackinnonBuck:
> @MackinnonBuck ... Looks good, but can we use `IAppEnvironment`? Trying to keep it short to keep the code line lengths at 85 or less chars.

That sounds fine to me! 🙂

aspnetcore/blazor/components/render-modes.md
```diff
@@ -284,6 +312,41 @@ In the following example, the component is interactive throughout the process. T
...
+</p>
+```
+
+No compile time error occurs, but a runtime error occurs during prerendering:
+
+> :::no-loc text="Cannot provide a value for property 'Environment' on type 'BlazorWebAppSample.Client.Pages.Home'. There is no registered service of type 'Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment'.":::
+
+This error occurs because the component must compile and execute on the server during prerendering, but <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> isn't a registered service on the server.
+
+There are a four approaches that you can take to address this scenario. The following are listed from most recommended to least recommended:
```
SteveSandersonMS, line 340: Often a simpler option is to inject `IServiceProvider` instead of the service type itself, and then you can have code like:

```cs
if (Services.GetService<IWebAssemblyHostEnvironment>() is {} env)
{
   // ... do stuff with env
}
```

TBH that seems easier than making a new abstraction.

aspnetcore/blazor/fundamentals/environments.md
```diff
@@ -39,33 +36,41 @@ For general guidance on ASP.NET Core app configuration, see <xref:fundamentals/e
 
 When running an app locally, the environment defaults to `Development`. When the app is published, the environment defaults to `Production`.
 
+<!-- UPDATE 9.0 The underlying problem with app settings filename 
+                case sensitivity is tracked for 9.0 by ...
+                https://github.com/dotnet/aspnetcore/issues/25152 -->
+                
+We recommend the following conventions:
+
+* Always use the "`Development`" environment name for local development. This is because the Blazor framework expects exactly that name when configuring the app and tooling for local development runs of a Blazor app.
```
SteveSandersonMS, line 45: It's ASP.NET Core as a whole that's coupled to the name "Development", not Blazor specifically.

SteveSandersonMS:
@guardrex @MackinnonBuck Could you clarify which specific part of this doc is the "approach" that you consider at risk of being controversial, and are asking for feedback about? I've read through it and nothing instantly springs out as being problematic, but it is a long doc and maybe I've missed something.

SteveSandersonMS:
Oh wait, I think I see now. You're referring to the stuff about `IAppEnvironment`.

While this approach would certainly work, it has the drawback that it doesn't create compatibility with existing code that already depends on accessing `IWebAssemblyHostEnvironment`. Two other possibilities:

## Implement IWebAssemblyHostEnvironment on the server

You may have considered this, not sure, but why not simply create your own service implementation for `IWebAssemblyHostEnvironment` on the server?

```cs
public class ServerHostEnvironment(IWebHostEnvironment env, NavigationManager nav) : IWebAssemblyHostEnvironment
{
    public string Environment => env.EnvironmentName;
    public string BaseAddress => nav.BaseUri;
}
```

... with `builder.Services.TryRegisterScoped<IWebHostEnvironment, ServerHostEnvironment>()`. Now all existing code will just work.

## Fix it in the framework

Longer term, if this is a common pain point, we could change the framework code that registers the WebAssembly hosting endpoints etc so that it also registers an implementation of `IWebAssemblyHostEnvironment` just like the one above, so then the whole problem vanishes for everyone. If this seems right (@MackinnonBuck?) we should track an issue for it.

aspnetcore/blazor/components/render-modes.md
```diff
@@ -142,6 +142,34 @@ Technically, `@rendermode` is both a Razor *directive* and a Razor *directive at
 > [!NOTE]
 > Component authors should avoid coupling a component's implementation to a specific render mode. Instead, component authors should typically design components to support any render mode or hosting model. A component's implementation should avoid assumptions on where it's running (server or client) and should degrade gracefully when rendered statically. Specifying the render mode in the component definition may be needed if the component isn't instantiated directly (such as with a routable page component) or to specify a render mode for all component instances.
 
+## Apply a render mode to the entire app
+
+To set the render mode for the entire app, indicate the render mode at the highest-level interactive component in the app's component hierarchy that isn't a root component.
```
javiercn, line 147: A concrete example here will be good. It's impossible for anyone without context to understand what is being said here. (To give you an idea, I have context and I had to stop to think about it).

guardrex:
> create your own service implementation for IWebAssemblyHostEnvironment on the server?

I like it, but I recommend keeping the service-based approach listed in case it has utility in other cases.

One thing tho ... is `TryRegisterScoped` valid API? Did you mean `TryAddScoped`?

WRT your earlier remark ...

> ```csharp
> if (Services.GetService<IWebAssemblyHostEnvironment>() is {} env)
> {
>    // ... do stuff with env
> }
> ```

I need more information to understand what problem that's addressing in this context.

aspnetcore/blazor/components/render-modes.md
```diff
@@ -142,6 +142,34 @@ Technically, `@rendermode` is both a Razor *directive* and a Razor *directive at
...
+
+> [!NOTE]
+> Making a root component interactive, such as the `App` component, isn't supported because the Blazor script may be evaluated multiple times. Therefore, the render mode for the entire app can't be set directly by the `App` component.
```
guardrex, line 150:
```suggestion
> Making a root component interactive, such as the `App` component, isn't supported. Therefore, the render mode for the entire app can't be set directly by the `App` component.
```

aspnetcore/blazor/components/render-modes.md
guardrex: The definition of a render mode and how to set it for component instances and component definitions were just explained in the preceding sections. What's not clear?

javiercn: I did not go through the previous section. Let me read through it

guardrex:
@SteveSandersonMS ... I placed your approach as the top pick. I just moved the ***concept*** of custom services down to the second spot in the list, and it won't have an example in the docs. In the *Environments* article, I placed the full example that you provided.

aspnetcore/blazor/components/render-modes.md
guardrex: AH! I see ... yes ... ***BUT***, it won't resolve an environment at all during prerendering. If the dev is counting on the value there, this will 💥 break their logic, correct?

aspnetcore/blazor/components/render-modes.md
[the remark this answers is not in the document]
SteveSandersonMS: In my recollection, it was that the script was being evaluated multiple times.

Nonetheless we don't have to explain *why* it's unsupported for the root component to be interactive. We only need to communicate that it is unsupported.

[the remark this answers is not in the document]
guardrex: I fixed it by removing the reason.

SteveSandersonMS:
> One thing tho ... is TryRegisterScoped valid API? Did you mean TryAddScoped?

Yeah I probably did mean TryAddScoped. Sorry. I was just typing from memory and haven't verified that the code is correct and would compile.

guardrex:
I added the `IServiceProvider` approach as a approach, but I did so above the list of *complete* solutions to the problem ... i.e., it isn't useful for obtaining a value during prerendering (if that's what the dev's spec calls for).

@MackinnonBuck, your bullet number 2 was @SteveSandersonMS's idea in code form. I removed that bullet from the list given that I'm moving it up above the list.

I also moved your last bullet on disabling prerendering to just under the `IServiceProvider` approach because that also doesn't solve the problem where the dev wants prerendering and a value.

The list is meant to include only the approaches that fully solve the problem, which is to obtain the value during prerendering. I think the list has the three good approaches.

I recommend re-reviewing the entire section along with the new example in the *Environments* doc.

aspnetcore/blazor/fundamentals/environments.md
```diff
@@ -135,57 +152,102 @@ In the following example for IIS, the custom header (`Blazor-Environment`) is ad
 
 > [!NOTE]
 > To use a custom `web.config` file for IIS that isn't overwritten when the app is published to the `publish` folder, see <xref:blazor/host-and-deploy/webassembly#use-a-custom-webconfig>.
+>
+> Although the Blazor framework issues the header name in all lowercase letters (`blazor-enviornment`), you're welcome to use any casing that you desire. For example, a header name that capitalizes each word (`Blazor-Enviornment`) is supported.
```
javiercn, line 156:
```suggestion
> Although the Blazor framework issues the header name in all lowercase letters (`blazor-environment`), you're welcome to use any casing that you desire. For example, a header name that capitalizes each word (`Blazor-Environment`) is supported.
```

guardrex:
Going ahead with this. If there's additional feedback, I can create a new PR.

PR #31145, opened 2023-11-28
Custom Blazor WASM loading progress indicator

### LINKED ISSUE #31048: Document how to display a loading UI for interactive WebAssembly rendering
We don't currently have a great way to display a loading UI for interactive WebAssembly rendering, but as @javiercn [pointed out](https://github.com/dotnet/aspnetcore/issues/52076#issuecomment-1812541641) there is a way to do it:

* Prerender a wrapper component that displays the placeholder on the page.
  * Inside the wrapper, render your component with prerender: false
* Have the wrapper use `@if(OsPlatform.IsBrowser())` to distinguish between SSR and WebAssembly and render your placeholder or the actual component.

guardrex:
Fixes #31048
Addresses #28161 

I built something that doesn't match what Javier described at https://github.com/dotnet/aspnetcore/issues/52076#issuecomment-1812541641. The 🦖 ***RexHax!***&trade; approach on the PR seems to work, but I suspect it has some 🐉🐉🐉. It at least doesn't seem to be right for WASM components that don't prerender.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/rendering.md](https://github.com/dotnet/AspNetCore.Docs/blob/3d02c3358740da57251f2f2b1fdda65d222bdf30/aspnetcore/blazor/components/rendering.md) | [ASP.NET Core Razor component rendering](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/rendering?branch=pr-en-us-31145) |
| [aspnetcore/blazor/fundamentals/startup.md](https://github.com/dotnet/AspNetCore.Docs/blob/3d02c3358740da57251f2f2b1fdda65d222bdf30/aspnetcore/blazor/fundamentals/startup.md) | [ASP.NET Core Blazor startup](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/startup?branch=pr-en-us-31145) |


mkArtakMSFT:
@SteveSandersonMS can you please review this? Thanks!

SteveSandersonMS (approved):
This is a bit of a complicated solution but I appreciate it has the benefit that, if you want to, you can load data as normal and use persistent component state to transfer that data to the client.

In the other case where a developer definitely does not plan to transfer any loaded data via persistent component state, a much simpler and more efficient solution would be to disable prerendering and go back to using `builder.RootComponents.Add(...)` in the `Program.cs` of their WebAssembly project. They can then use exactly the same markup and CSS as is used in the Standalone Blazor WebAssembly template, and don't need to introduce extra wrapper components.

So ultimately the best option depends on the app scenario, but I do appreciate that what you have here will work.

guardrex:
Thanks. I'll go with this for now.

If you want coverage for the approach that you described, I'll work it via a new issue, but I'll need to see it fully implemented or explained further because it doesn't seem like its *exactly* like the markup of a standalone WASM app. What's the selector for the root component and where the selector is placed in the BWA? What are the details and gotchas 😈 for that approach in the BWA's layout and routing contexts? Let me know if you want it covered👂... if so, I'll open a new issue and we can go over how that works in more detail.

PR #31169, opened 2023-11-30
Improve render modes remark

Jimsparkle:
Fixing typo.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/render-modes.md](https://github.com/dotnet/AspNetCore.Docs/blob/bd8c404cb22064bff4940b81db3a2441bc41cd27/aspnetcore/blazor/components/render-modes.md) | [ASP.NET Core Blazor render modes](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/render-modes?branch=pr-en-us-31169) |

### DIFF
--- a/aspnetcore/blazor/components/render-modes.md
+++ b/aspnetcore/blazor/components/render-modes.md
@@ -163,7 +163,7 @@ You also typically must set the same interactive render mode on the [`HeadOutlet
 <HeadOutlet @rendermode="InteractiveServer" />
 ```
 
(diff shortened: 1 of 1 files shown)


guardrex (approved):
Thanks @Jimsparkle ... We can't use the exact wording of your suggestion because it would change the meaning of the line. It's important that the global application of the WASM/Auto render mode be retained. However, I was able to improve the line a bit. Thanks for calling out the confusion on it.

PR #31188, opened 2023-12-01
Update Program.cs to dispose HttpClient at the end

RyanLiu99:


Rick-Anderson:
See [HttpClient and lifetime management](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-8.0&viewFallbackFrom=aspnetcore-9.0&branch=pr-en-us-31188#httpclient-and-lifetime-management)

`HttpClient `instances can generally be treated as .NET objects **_not_** requiring disposal.

PR #31340, opened 2024-01-03
Example description position fix

Ususucsus:
Before fix example description was placed after code block in subsequent section. After fix its placed before example like in webapplication7.md


Ususucsus:
#sign-off

guardrex (approved):
Thanks @Ususucsus! 🚀

PR #31353, opened 2024-01-04
Prerendering and interactive/enhanced routing

### LINKED ISSUE #31351: Add remarks on prerendering with interactive routing
### Description

Per offline discussion with Mackinnon ... internal navigations are interactive with an interactive router and don't involve requesting new page content from the server. Therefore, prerendering doesn't occur.

Also ...

> The current persistent component state feature only works on the initial page load. That is, if you perform a full (non-enhanced) navigation to a page utilizing persistent component state, the persisted state will be made available for the app to use when it becomes interactive, and things will work as expected. But if an interactive circuit has already been established and an enhanced navigation gets performed to a page that renders persisted component state, that state will not be made available in the existing circuit.

Goals ...

* Add remarks on both behaviors.
* Add remarks to prerendering examples that if using an interactive router to use a full page reload to see the demo behavior explained by the article.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/prerender?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/prerender.md

### Document ID

ffdc8df3-55a8-b757-d4e6-38c104916002

### Article author

@guardrex

guardrex:
Fixes #31351

Mackinnon ...

Even if this is heading in the right direction, we probably have muddled wording for "interactive routing" and "enhanced navigation." Are they the same? ... because we're blending the language in discussion and in articles [e.g., "interactive (enhanced) routing" or "interactive (enhanced) navigation"]. If the following pairs reflect the same concepts, there's more work to do here ...

* "static" and "non-enhanced" ... ***same thing?***
* "interactive" and "enhanced" ... ***same thing?***
* "routing" and "navigation" ... ***same thing?***

If possible, I think we should use consistent terminology everywhere. If I'm grouping compatible concepts, which of the following is the best pair to use? ... and should we loop all of the Blazor PU into this discussion?

* Option 1
  * "static (non-enhanced) routing"
  * "interactive (enhanced) routing"
* Option 2
  * "static (non-enhanced) navigation"
  * "interactive (enhanced) navigation"
* Option 3
  * "static (non-enhanced) routing and navigation"
  * "interactive (enhanced) routing and navigation"
* Option 4
  * "static routing (non-enhanced navigation)"
  * "interactive routing (enhanced navigation)"
* Option 6 (probably not viable given that general discussion, convos, issues, etc. use "enhanced navigation" lingo quite a bit)
  * "static routing"
  * "interactive routing"
* Option 7 (probably not viable because "interactive routing/router" lingo is used quite a bit)
  * "enhanced navigation"
  * "non-enhanced navigation"

Additionally, we've introduced "static router" and "interactive router" in just a couple of spots, and I hope to limit the use of those terms to just the few spots we have them.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/cascading-values-and-parameters.md](https://github.com/dotnet/AspNetCore.Docs/blob/9fc1b0aaef87382044cbd255d098d5a74b20da73/aspnetcore/blazor/components/cascading-values-and-parameters.md) | [ASP.NET Core Blazor cascading values and parameters](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/cascading-values-and-parameters?branch=pr-en-us-31353) |
| [aspnetcore/blazor/components/prerender.md](https://github.com/dotnet/AspNetCore.Docs/blob/9fc1b0aaef87382044cbd255d098d5a74b20da73/aspnetcore/blazor/components/prerender.md) | [Prerender ASP.NET Core Razor components](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/prerender?branch=pr-en-us-31353) |
| [aspnetcore/blazor/fundamentals/routing.md](https://github.com/dotnet/AspNetCore.Docs/blob/9fc1b0aaef87382044cbd255d098d5a74b20da73/aspnetcore/blazor/fundamentals/routing.md) | [ASP.NET Core Blazor routing and navigation](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?branch=pr-en-us-31353) |


MackinnonBuck:
I would say there are two main categories of routing:
* Static (server-side) routing
* Interactive (client-side) routing

Static routing is where the page content gets determined by making an HTTP request to the server and displaying the response HTML on the page.

Interactive routing generates page content _without_ requiring the server to render HTML to get displayed on the page.

_**Note:** I've personally sometimes used the terms "server-side" and "client-side" rather than "static" and "interactive" because those are the terms that the rest of the industry tends to use. It just happens to be confusing in Blazor's case because a router with Server interactivity has the characteristics of a client-side router, while the routing logic is technically running on the server._

---

> * "routing" and "navigation" ... same thing?

I think of navigation as the act of moving from one page to another. Routing is the act of mapping a URL to page, which is one of the mechanisms required to enable navigation.

> * "static" and "non-enhanced" ... same thing?
> * "interactive" and "enhanced" ... same thing?

"Enhanced" navigation is a variation of "static" navigation. A _non-enhanced_ navigation is just a normal static navigation that clears all JS/DOM state as the navigation occurs. These are the types of navigations you would get in a Razor Pages app, for example. An enhanced navigation works similarly, except it attempts to preserve as much browser-side state as possible when displaying the new HTML returned from the server.

So the hierarchy of these terms is:
* Static/server-side navigation
  * Enhanced navigation
  * Non-enhanced navigation
* Interactive/client-side navigation

Hopefully that makes sense! I don't have a very strong opinion about whether "server/client" or "static/interactive" are the right set of terms to use. Maybe @danroth27 has thoughts on this?

aspnetcore/blazor/components/prerender.md
```diff
@@ -104,7 +104,10 @@ The following counter component example persists counter state during prerenderi
 
 :::code language="razor" source="~/../blazor-samples/8.0/BlazorSample_BlazorWebApp/Components/Pages/PrerenderedCounter2.razor":::
 
-When the component executes, `currentCount` is only set once during prerendering. The value is restored when the component is rerendered. The following is example output:
+When the component executes, `currentCount` is only set once during prerendering. The value is restored when the component is rerendered. The following is example output.
+
+> [!NOTE]
+> If the app adopts interactive (enhanced) routing and the page is reached via an internal navigation, prerendering doesn't occur. Therefore, you must perform a full page reload for the `PrerenderedCounter2` component to see the following output.
```
MackinnonBuck, line 110:
```suggestion
> If the app adopts interactive routing and the page is reached via an internal navigation, prerendering doesn't occur. Therefore, you must perform a full page reload for the `PrerenderedCounter2` component to see the following output.
```

aspnetcore/blazor/components/prerender.md
```diff
@@ -125,6 +128,14 @@ For components embedded into a page or view of a Razor Pages or MVC app, you mus
...
+
+A mechanism to deliver state updates for enhanced navigation is under consideration for .NET 9, which is targeted for release in late 2024. For more information, see [[Blazor] Support persistent component state across enhanced page navigations (dotnet/aspnetcore #51584)](https://github.com/dotnet/aspnetcore/issues/51584). For more information on enhanced navigation, see <xref:blazor/fundamentals/routing#enhanced-navigation-and-form-handling>.
```
MackinnonBuck, line 137: We may want to move these two paragraphs to a separate section, since they apply to cases where there's a static (non-interactive) router.

guardrex:
Thanks @MackinnonBuck ... I'll digest that for a round of updates to this PR on Friday morning. 

Is what's on the PR good thus far?

guardrex:
If we have a change thus far just for what's on the PR now, it will likely be for the language ...

> interactive (enhanced) routing

aspnetcore/blazor/components/prerender.md
MackinnonBuck, line 137: Should we say this? While I think it's likely we'll do this, there are some unknowns that may impact the strategy/degree to which this gets solved. I appreciate that the wording implies that we haven't fully committed to this yet, but some customers might still misinterpret the meaning. @danroth27 what do you think?

MackinnonBuck: (this feedback specifically applies to line 137)

guardrex: I can cut it. Yes, I was trying to say it ***might be*** addressed with the "under consideration for .NET 9" language.

guardrex: Yes, I agree ... I think the problem is that the persist component state service is only discussed in light of maintaining state ***for prerendering*** (i.e., in this article). It's broader than that, so these two paragraphs can go where the persist component state service is described elsewhere and then the use of the service for prerendering situations can be cross-linked from this article to that new location ......

I just need to see ***where*** that the service guidance should live.

danroth27: Yeah, let's not promise anything for .NET 9. Nothing is committed yet for .NET 9, and even when we "commit" to stuff it often ends up getting cut or punted to a future release.

guardrex, line 137:
```suggestion
```

guardrex:
@MackinnonBuck ... I looked over your remarks. I think we have most of that already covered fairly well in the *Routing* article per our pre-holiday work. It's in two sections ...

* [Static versus interactive routing](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-8.0#static-versus-interactive-routing)
* [Enhanced navigation and form handling](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-8.0#enhanced-navigation-and-form-handling)

However, we don't have the cross-cutting list that you provided ...

> So the hierarchy of these terms is:
>
> * Static/server-side navigation
>   * Enhanced navigation
>   * Non-enhanced navigation
> * Interactive/client-side navigation

... because we keep the subjects in two separate sections. I'm not sure if we should add more content to the *Enhanced navigation and form handling* section or not.

One thing came up in my review of the content. I had messaged you about what I perceived to be 👽 behavior of the Persistent Component State service, but ***it was covered in Routing after all*** ...

> Interactive routing also prevents prerendering because new page content isn't requested from the server with a normal page request.

... a bit buried as the last sentence of a paragraph and certainly something that would be easy to forget over the holiday break. This PR will now make it clear in the *Prerender* article. I'm going to add a cross-link from that remark in the *Routing* article to this new *Prerender* article content.

The requested changes thus far have been made to this PR. Would you give it a final look to see if it's good to go?

aspnetcore/blazor/components/prerender.md
```diff
@@ -125,6 +130,12 @@ For components embedded into a page or view of a Razor Pages or MVC app, you mus
 </body>
 ```
 
+## Interactive routing and prerendering
+
+Internal navigation for interactive routing doesn't involve requesting new page content from the server. Therefore, prerendering doesn't occur for internal page requests.
+
+The <xref:Microsoft.AspNetCore.Components.PersistentComponentState> service only works on the initial page load and not across enhanced page navigation events. If the app performs a full (non-enhanced) navigation to a page utilizing persistent component state, the persisted state is made available for the app to use when it becomes interactive. But if an interactive circuit has already been established and an enhanced navigation is performed to a page that renders persisted component state, that state isn't made available in the existing circuit. The <xref:Microsoft.AspNetCore.Components.PersistentComponentState> service isn't aware of enhanced navigation, and there's no mechanism to deliver state updates to components that are already running.
```
MackinnonBuck, line 137: Since this is about enhanced navigation, which is a type of static (non-interactive) navigation, should this paragraph go somewhere else? I'm also wondering if we should cross-reference any docs that discuss enhanced navigation in case this is the first time the reader is seeing this term.

MackinnonBuck (approved):
Looks good to me - I have one more piece of feedback but feel free to address it however you like 🙂

aspnetcore/blazor/components/prerender.md
MackinnonBuck: (referring to that last paragraph, I'm not sure why GitHub keeps including the previous paragraph in the review comment)

guardrex: I don't think so because this is the first place that the service is described and it only pertains to the subject of the article, which is prerendering.

Enhanced nav already cross-links to this, so I think I just need to cross-link this back out to the enhanced nav coverage in the *Routing* article.

After this PR, I'll keep an 👂 open as usual for reader feedback.

PR #31376, opened 2024-01-06
Update virtualization.md to inform about the need to set the renderMode to InteractiveServer

vpaulino:
I have come to the fact that I was trying to setup a project with .NET 8 to use virtualize and I only had the configuration on program.cs to:
 builder.Services.AddRazorComponents()
    .AddInteractiveServerComponents();  and 
app.MapRazorComponents<App>().AddInteractiveServerRenderMode();    

Only with that configuration Virtualize component in version 8.0 is not rendering the elements. 

For sake of clarity I belive that is important to inform developers that virtualize only works if the rendermode is interactiveServer explicitly on the page or component where is to be used.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/virtualization.md](https://github.com/dotnet/AspNetCore.Docs/blob/09533e4ad459b0e7bf0bd021d96400d4b86836ff/aspnetcore/blazor/components/virtualization.md) | [ASP.NET Core Razor component virtualization](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/virtualization?branch=pr-en-us-31376) |

### DIFF
--- a/aspnetcore/blazor/components/virtualization.md
+++ b/aspnetcore/blazor/components/virtualization.md
@@ -50,6 +50,7 @@ Instead of rendering the entire list of flights at once, replace the [`foreach`]
 * Specify `allFlights` as a fixed item source to <xref:Microsoft.AspNetCore.Components.Web.Virtualization.Virtualize%601.Items%2A?displayProperty=nameWithType>. Only the currently visible flights are rendered by the <xref:Microsoft.AspNetCore.Components.Web.Virtualization.Virtualize%601> component.
 * Specify a context for each flight with the `Context` parameter. In the following example, `flight` is used as the context, which provides access to each flight's members.
 
+
 ```razor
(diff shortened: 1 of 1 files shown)


guardrex:
Hello @vpaulino ... The articles already say that by virtue of an INCLUDE file that I use widely across the doc set. It states ...

> Documentation component examples usually don't configure an interactive render mode with an `@rendermode` directive in the component's definition file (`.razor`):
>
> * In a Blazor Web App, the component must have an interactive render mode applied, either in the component's definition file or inherited from a parent component. For more information, see [ASP.NET Core Blazor render modes](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/render-modes?view=aspnetcore-8.0).

Issue #31403, opened 2024-01-10
Add example that sets a render mode programatically


guardrex:
### Description

This is a second example, and it's better for a new section on render mode by area. The example can show setting the render mode by path ...

```razor
<Routes @rendermode="@RenderModeForPage" />

...

[CascadingParameter]
private HttpContext HttpContext { get; set; } = default!;

private IComponentRenderMode? RenderModeForPage => 
    HttpContext.Request.Path.StartsWithSegments("/Admin") ? InteractiveServer : null;
```

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/render-modes?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/render-modes.md

### Document ID

7c9948a3-9886-8097-6014-492a4c21ea4e

### Article author

@guardrex

Issue #31402, opened 2024-01-10
The content inside <NotAuthorized> tag in Routes component won't be displayed 


FengRuikai:
### Description

The content inside <NotAuthorized> tag won't be displayed  if we don't configure Interactiveserver render mode for entire app,but its was not mentioned in the document

A related issue:https://stackoverflow.com/questions/77748801/how-to-do-windows-authentication-i-cant-get-notauthorized-to-show/77750838?noredirect=1#comment137134432_77750838

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/security/server/?view=aspnetcore-8.0&tabs=visual-studio

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/security/server/index.md

### Document ID

8da62490-e36f-8f38-8f06-d5d6bceaafde

### Article author

@guardrex

guardrex:
Hello @FengRuikai ... I see what you mean ... and from looking at the PU issues. I'm looking at Javier's explanations at ...

https://github.com/dotnet/aspnetcore/issues/52176#issuecomment-1825677291

... in particular his remark ...

> During SSR we won't show NotAuthorized, because your component doesn't even get to render.

... which seems like the critical bit pertaining to this issue. However, I might need to provide more coverage on the behaviors based on his full set of remarks.

guardrex:
I've studied several PU issues. I'm not sure what to do yet in the docs. Are we merely going to state the reason, or are we also going to prescribe approaches to take to address the behaviors? If we're stating approaches, what approaches? I'll need to work on this further tomorrow.

I think what I'll need to do is play with some local scenarios to study BWA behaviors so I can see what devs are reporting firsthand ... ***and*** ... try a few things that are discussed to see their effect on the app. I can probably address the approaches aspect better if I work directly with use cases. Trying to understand it from dev descriptions and PU responses is only partially working for me right now. I'll pick back up tomorrow with this. I just hit a 🦖🧠🔥 for the day on this subject 😆.

guardrex:
**UPDATE** ... It's taken ***forever*** 😩 to get back to this. I've been 🏃‍♂️ on higher priority 8.0 issues over the last several weeks. This is one of the last issues to address. I'm going to work on it ***right now***.

PR #31405, opened 2024-01-10
Additional programmatic render mode guidance

guardrex:
Addresses #31403
Addresses #28161

Follow-up to PR earlier this morning #31404 that added a new section on programmatic render modes. There are some additional scenarios to cover in the section.

Mackinnon ... two problems with what I'm adding ...

* The description of the code is ***very likely incorrect*** 🙈 (or at least incomplete) ...

  > set the render mode of the app dynamically based on the render modes of the components in the app

  What is this code doing? ... because when the router becomes interactive it seems like whatever was set on `Routes` is what is then going to be used thereafter for each navigation ... well 🤔 ... is it that the router will continue working interactively until a route for a static SSR component can't be matched? ... and then a request is made back to the server, where the root component re-renders and then sucks in the render mode of the requested component and applies that to the `Routes` component?

* ***Why bother?*** 😄 ... I don't understand what this approach going to do for devs? I think we need to tell devs exactly what scenario(s) this approach applies to.

  This PR is based on the cross-linked discussion between a dev and Javier at https://github.com/dotnet/aspnetcore/issues/52176. It looks like a workaround to avoid the flash-over from the server-rendered error page to the `NotFound` router content. Also possibly related are these various `NotFound`/`NotAuthorized` no-op situations, where under static SSR the server isn't going to render content from those components ... described by Javier at https://github.com/dotnet/aspnetcore/issues/52176#issuecomment-1825677291.

  If you look just above what this PR is adding, I placed some content earlier this morning on what seems like a simpler scenario where there's an app that defaults to static SSR components or per-component render modes but has this one area of components that you would like to explicitly configure globally for interactive SSR. Let me know if that bit is correct, but I want to mention that it seems simpler to understand ***why*** one would do that ... e.g., the `Admin` area of pages/components that the example uses. The dev wants to globally apply interactive SSR to that area of administrator pages and default the rest to static SSR or per-component render mode assignment. For the new content that I'm seeking to add on this PR, it's not as clear ***why*** one would adopt it, and I hope we can add a why to the approach.

I'll also mention that I'm working a related issue at https://github.com/dotnet/AspNetCore.Docs/issues/31402. I'm going sleep on it because I just hit a *brain fry* 🧠🔥. I'll test locally soon with some scenarios to experience what's being discussed in PU issues (not authorized and not found scenarios) and figure out what updates to draft.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/render-modes.md](https://github.com/dotnet/AspNetCore.Docs/blob/c155473188b25d3535d0a45bb008381b5a959054/aspnetcore/blazor/components/render-modes.md) | [ASP.NET Core Blazor render modes](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/render-modes?branch=pr-en-us-31405) |

### DIFF
--- a/aspnetcore/blazor/components/render-modes.md
+++ b/aspnetcore/blazor/components/render-modes.md
@@ -188,32 +188,45 @@ A component definition can define a render mode via a private field:
 
 ```razor
 @rendermode componentRenderMode
+```
 
-...
-
-@code {
-    private static IComponentRenderMode componentRenderMode =
-        new InteractiveServerRenderMode();
-}
+```csharp
+private static IComponentRenderMode componentRenderMode = new InteractiveServerRenderMode();
 ```
 
 ### By component instance
 
-<xref:Microsoft.AspNetCore.Http.HttpContext> is available in statically-rendered root components, such as the `App` component (`App.razor`). You can use [`HttpContext.Request.Path`](xref:Microsoft.AspNetCore.Http.HttpContext.Request%2A) to specify a render mode for a group of pages.
+<xref:Microsoft.AspNetCore.Http.HttpContext> is available in statically-rendered root components, such as the `App` component (`App.razor`). You can use [`HttpContext.Request.Path`](xref:Microsoft.AspNetCore.Http.HttpContext.Request%2A) to specify a render mode for a group of pages from a statically-rendered root component.
 
 The following example applies interactive server-side rendering (interactive SSR) to any request for a component in the app's `Admin` folder (`Components/Admin`), including subfolders. Components at any other path don't receive a render mode (`null`) from the `Routes` component, and they either render statically, inherit a render mode from a parent component, or set their own render mode.
 
 ```razor
 <Routes @rendermode="@RenderModeForPage" />
+```
 
-...
-
+```csharp
 [CascadingParameter]
 private HttpContext HttpContext { get; set; } = default!;
 
 private IComponentRenderMode? RenderModeForPage => 
     HttpContext.Request.Path.StartsWithSegments("/Admin") ? InteractiveServer : null;
 ```
+
+You can also set the render mode of the app dynamically based on the render modes of the components in the app. Each component sets a render mode, and the `App` component assigns the render mode of the loaded component to the `Routes` component:
+
+```razor
+<Routes @rendermode="@RenderModeForPage" />
+```
+
+```csharp
+[CascadingParameter]
+private HttpContext HttpContext { get; set; } = default!;
+
+private IComponentRenderMode? RenderModeForPage =>
+    HttpContext.GetEndpoint()?.Metadata.GetMetadata<RenderModeAttribute>()?.Mode;
+```
+
+A final approach for applying a render mode programatically is to apply the render mode to a <xref:Microsoft.AspNetCore.Components.ComponentBase> base class and [use the base class](xref:blazor/components/index#specify-a-base-class) to create components.
 
 Additional information on render mode propagation is provided in the [Render mode propagation](#render-mode-propagation) section later in this article.
 


MackinnonBuck:
> What is this code doing? ... because when the router becomes interactive it seems like whatever was set on Routes is what is then going to be used thereafter for each navigation ... well 🤔 ... is it that the router will continue working interactively until a route for a static SSR component can't be matched? ... and then a request is made back to the server, where the root component re-renders and then sucks in the render mode of the requested component and applies that to the Routes component?

It sounds like your understanding is pretty accurate - as soon as the router becomes interactive, it will stay that way until the page gets reloaded. That's because the logic determining the `RenderModeForPage` only runs during SSR. If you want to "escape" back to being server-side rendered, you'll need to do that manually.

The way we've handled this in the Blazor Web app template (with individual auth and app-level interactivity) is we have a layout component that applies to all the "Account" pages whose purpose is to do a full page reload when the page gets navigated to interactively:

```razor
@inherits LayoutComponentBase
@layout MainLayout
@inject NavigationManager NavigationManager

@if (HttpContext is null)
{
    <p>Loading...</p>
}
else
{
    @Body
}

@code {
    [CascadingParameter]
    private HttpContext? HttpContext { get; set; }

    protected override void OnParametersSet()
    {
        if (HttpContext is null)
        {
            // If this code runs, we're currently rendering in interactive mode, so there is no HttpContext.
            // The identity pages need to set cookies, so they require an HttpContext. To achieve this we
            // must transition back from interactive mode to a server-rendered page.
            NavigationManager.Refresh(forceReload: true);
        }
    }
}
```

Customers will likely need to do something similar if they want part of their app to have top-level interactivity and the rest to using static server rendering.

> Why bother?

The best real-world example I can think of is the Blazor Web app template with Identity pages and top level interactivity enabled. In that template, the developer expressed that their desired rendering strategy is to have top-level interactivity. Maybe the customer's app is highly interactive, so it doesn't make sense to add the complexity of SSR when it doesn't provide much benefit. However, the identity pages are incompatible with interactive rendering - they rely heavily on the `HttpContext`, utilize cookies, etc.

So now you have two parts of the app that have fundamentally different requirements and rendering strategies, and you need a way to transition from one style to another. That's where this approach would be useful.

guardrex:
> do a full page reload when the page gets navigated to interactively:

... along with the other piece to make it static SSR on the reload ...

```razor
<Routes @rendermode="@RenderModeForPage" />

...

@code {
    ...

    private IComponentRenderMode? RenderModeForPage => HttpContext.Request.Path.StartsWithSegments("/Account")
        ? null
        : InteractiveServer;
}
```

> Customers will likely need to do something similar if they want part of their app to have top-level interactivity and the rest to using static server rendering.

Yes, I think this whole technique should be described. I'll work on it.

guardrex:
@MackinnonBuck ... You know what .... this approach ...

```razor
<Routes @rendermode="@RenderModeForPage" />
```

```csharp
[CascadingParameter]
private HttpContext HttpContext { get; set; } = default!;

private IComponentRenderMode? RenderModeForPage =>
    HttpContext.GetEndpoint()?.Metadata.GetMetadata<RenderModeAttribute>()?.Mode;
```

...  🤔... kind'a looks really good for the ***per-component*** approach. The project template is saying that if the static SSR component is in a defined area of the app (i.e., the `/Account` area), then decide to set the render mode on `Routes` to `null`. The code above says to use (reflection?) what the component set ... so ... it looks like you don't need to have a defined area (folder) for static SSR components if you're willing to do a per-component assignment and just leave your static SSR components without an `@rendermode` directive.

guardrex:
I'm going to reconstitute this on a local branch for further work.

PR #31497, opened 2024-01-18
Fine control of static SSR

guardrex:
Addresses #31403
Addresses #28161

* Now we're cook'in with ***gas*** 🔥.
* There's a section here named *Client-side services fail to resolve during prerendering* that you can ignore. That section is merely moving down in the article, and there are no changes to it.
* The two sections for review are:
  * ***Apply a render mode in code*** and its two subsections
  * ***Fine control of render modes*** and its two subsections
* We might want to change the section heading name, *Fine control of render modes*, but I'm not sure yet what would be better. I'll 🤔 about it.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/render-modes.md](https://github.com/dotnet/AspNetCore.Docs/blob/d005f79d64c26e4541fc704c300a054afb66928e/aspnetcore/blazor/components/render-modes.md) | [ASP.NET Core Blazor render modes](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/render-modes?branch=pr-en-us-31497) |
| [aspnetcore/blazor/security/server/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/d005f79d64c26e4541fc704c300a054afb66928e/aspnetcore/blazor/security/server/index.md) | [[Visual Studio](#tab/visual-studio)](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/server/index?branch=pr-en-us-31497) |

### DIFF
--- a/aspnetcore/blazor/components/render-modes.md
+++ b/aspnetcore/blazor/components/render-modes.md
@@ -178,7 +178,7 @@ To enable global interactivity when creating a Blazor Web App:
 
 For more information, see <xref:blazor/tooling>.
 
-## Apply a render mode in code
+## Apply a render mode programatically
 
 Properties and fields can assign a render mode.
 
--- a/aspnetcore/blazor/components/render-modes.md
+++ b/aspnetcore/blazor/components/render-modes.md
@@ -178,44 +178,48 @@ To enable global interactivity when creating a Blazor Web App:
 
 For more information, see <xref:blazor/tooling>.
 
-## Apply a render mode programatically
+## Apply a render mode in code
 
-Properties and fields can assign a render mode to a component or a group of components via inheritance.
+Properties and fields can assign a render mode.
 
-### By component definition
+The second approach described in this section, setting the render mode by component instance, is especially useful when your app specification calls for either of the following scenarios:
+
+* You have an area (folder) of the app with components that must adopt static server-side rendering (static SSR) and only run on the server. The app controls the render mode globally by setting the render mode on the `Routes` component in the `App` component based on the path to the folder.
+* You have components around the app in various locations (not in a single folder) that must adopt static SSR and only run on the server. The app controls the render mode on a per-component basis by setting the render mode with the `@rendermode` directive in component instances and then using reflection in the `App` component to set the render mode on the `Routes` component.
+
+In both cases, the component that must adopt static SSR must also force a full-page reload.
+
+The preceding two scenarios are covered with examples in the [Fine control of render modes](#fine-control-of-render-modes) section later in this article. The following two subsections focus on basic approaches for setting the render mode.
+
+### Set the render mode by component definition
 
 A component definition can define a render mode via a private field:
 
 ```razor
-@rendermode componentRenderMode
+@rendermode renderModeForPage
 
 ...
 
 @code {
-    private static IComponentRenderMode componentRenderMode =
-        new InteractiveServerRenderMode();
+    private static IComponentRenderMode renderModeForPage = InteractiveServer;
 }
 ```
 
-### By component instance
+### Set the render mode by component instance
 
-<xref:Microsoft.AspNetCore.Http.HttpContext> is available in statically-rendered root components, such as the `App` component (`App.razor`). You can use [`HttpContext.Request.Path`](xref:Microsoft.AspNetCore.Http.HttpContext.Request%2A) to specify a render mode for a group of pages.
-
-The following example applies interactive server-side rendering (interactive SSR) to any request for a component in the app's `Admin` folder (`Components/Admin`), including subfolders. Components at any other path don't receive a render mode (`null`) from the `Routes` component, and they either render statically, inherit a render mode from a parent component, or set their own render mode.
+The following example applies interactive server-side rendering (interactive SSR) to any request.
 
 ```razor
 <Routes @rendermode="@RenderModeForPage" />
 
 ...
 
-[CascadingParameter]
-private HttpContext HttpContext { get; set; } = default!;
-
-private IComponentRenderMode? RenderModeForPage => 
-    HttpContext.Request.Path.StartsWithSegments("/Admin") ? InteractiveServer : null;
+@code {
+    private IComponentRenderMode? RenderModeForPage => InteractiveServer;
+}
 ```
 
-Additional information on render mode propagation is provided in the [Render mode propagation](#render-mode-propagation) section later in this article.
+Additional information on render mode propagation is provided in the [Render mode propagation](#render-mode-propagation) section later in this article. The [Fine control of render modes](#fine-control-of-render-modes) section shows how to use the preceding component instance approach to adopt static SSR in either specific areas of the app (folders) or to specific components spread around the app with per-component render mode assignments.
 
 ## Prerendering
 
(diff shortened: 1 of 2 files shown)


aspnetcore/blazor/components/render-modes.md
```diff
@@ -178,44 +178,48 @@ To enable global interactivity when creating a Blazor Web App:
 
 For more information, see <xref:blazor/tooling>.
 
-## Apply a render mode programatically
+## Apply a render mode in code
```
MackinnonBuck, line 181: Hm, I actually think 'programmatically' might be a better work here, since one could argue that all render modes are applied "in code".

Or maybe 'dynamically' would be a better word, since it implies that the render mode is determined at runtime?

MackinnonBuck: Thinking about this more, I don't want the word 'dynamically' to give the illusion that this has anything to do with the auto render mode. Maybe 'programmatically' is better after all. What do you think?

aspnetcore/blazor/components/render-modes.md
```diff
@@ -178,44 +178,48 @@ To enable global interactivity when creating a Blazor Web App:
...
 
-Properties and fields can assign a render mode to a component or a group of components via inheritance.
+Properties and fields can assign a render mode.
 
-### By component definition
+The second approach described in this section, setting the render mode by component instance, is especially useful when your app specification calls for either of the following scenarios:
+
+* You have an area (folder) of the app with components that must adopt static server-side rendering (static SSR) and only run on the server. The app controls the render mode globally by setting the render mode on the `Routes` component in the `App` component based on the path to the folder.
+* You have components around the app in various locations (not in a single folder) that must adopt static SSR and only run on the server. The app controls the render mode on a per-component basis by setting the render mode with the `@rendermode` directive in component instances. Reflection is used in the `App` component to set the render mode on the `Routes` component.
```
MackinnonBuck, line 188: I'm not totally understanding why these are fundamentally different scenarios. Isn't the only difference between these what goes in the implementation for `RenderModeForPage`? i.e., in scenario 1, you decide the render mode by looking at the request path, and in scenario 2, you decide the render mode via reflection. If that's the case, could the code samples and descriptions for these be combined into one, where the potential ways to decide the render mode are explained in a single section? I could be missing something here, so let me know if I am!

MackinnonBuck (commented):
Looks good so far!

I still need to do a detailed review, but I'm waiting on that because there's a chance this feedback results in consolidating some content, and I think it would be better to wait until that happens, if it does! 🙂

aspnetcore/blazor/components/render-modes.md
guardrex: It's often best in complex use cases to format the guidance around the goal that the dev is trying to achieve. It's often critical to adopt more of a tutorial-type of approach for the new-to-programming, new-to-.NET, and new-to-Blazor folks. 

In addition to the delta on how `RenderModeForPage` is set, there are additional considerations for the force-loading layout and what the individual components need to do in regard to use of the layout and setting the render mode.

When you see me write out scenarios in this degree of detail, it's usually because I found it confusing myself and detect that readers would benefit from more of step-by-step, detailed explanation and set of examples. It's long, yes, but it's utterly clear what's going on and what the dev needs to do. If we aren't clear, it will upset readers 😡 and they'll come for the 🦖 with sharpened *TEETH* and *CLAWS* 👹😨😆. I think we'll head off doc issues by explaining it out.

aspnetcore/blazor/components/render-modes.md
guardrex, line 181:
```suggestion
## Apply a render mode programatically
```

aspnetcore/blazor/components/render-modes.md
MackinnonBuck: Sounds great to me, thanks for clarifying! I definitely trust your judgement about what's clear to readers, so let's go with your original strategy of having it documented as two separate scenarios 🙂

guardrex: Cool ... you still need time to fully analyze this?

aspnetcore/blazor/components/render-modes.md
```diff
@@ -579,6 +516,228 @@ The following component results in a runtime error when the component is rendere
...
+            NavigationManager.Refresh(forceReload: true);
+        }
+    }
+}
+```
+
+> [!NOTE]
+> In the Blazor Web App project template, there's a second layout file (`ManageLayout.razor` in the `Components/Account/Shared` folder) for Identity components in the `Components/Account/Pages/Manage` folder. The `Manage` folder has its own `_Imports.razor` file to apply to the `ManageLayout` to components in the folder. In your own apps, using nested `_Imports.razor` files is a useful approach for applying custom layouts to groups of pages.
+
+In the `App` component, any request for a component in the `Account` folder applies a `null` render mode, which enforces static SSR. Other component requests receive a global application of the interactive SSR render mode (`InteractiveServer`).
```
MackinnonBuck, line 581: > In the `App` component, any request for a component in the `Account` folder applies a `null` render mode, which enforces static SSR.

Just a small note about this: A `null` render mode is effectively the same as not specifying one at all, which results in the component inheriting its parent's render mode. In this case, the `App.razor` component gets rendered using static server rendering, so a `null` render mode makes the `<Routes>` component also use static server rendering. Whereas, if a `null` render mode were to be specified for a child component whose parent was interactive, the child would inherit the same interactive render mode.

All that to say, it's not always the case that `null` enforces static SSR, it just happens to behave that way in this example. Not sure whether it's worth calling out that caveat (maybe it would just create more confusion), so I'll leave it up to you to decide! Just want to make sure that readers don't see this statement and think that they can apply `null` as a render mode _anywhere_ to switch from interactive rendering to static server rendering.

aspnetcore/blazor/components/render-modes.md
```diff
@@ -579,6 +516,228 @@ The following component results in a runtime error when the component is rendere
...
+<Routes @rendermode="RenderModeForPage" />
+```
+
+```csharp
+[CascadingParameter]
+private HttpContext HttpContext { get; set; } = default!;
+
+private IComponentRenderMode? RenderModeForPage =>
+    HttpContext.GetEndpoint()?.Metadata.GetMetadata<RenderModeAttribute>()?.Mode;
+```
```
MackinnonBuck, line 652: Maybe there's a reason you didn't do this way, but would it make sense to just have a single ` ```razor ` block with a `@code` section rather than two different code blocks? Maybe that would make the example more copy-pastable.

MackinnonBuck (approved):
Approving now because I think this looks great - feel free to address my feedback however you like!

guardrex:
Thanks @MackinnonBuck ... I'll make further updates based on your feedback on Friday.

aspnetcore/blazor/components/render-modes.md
guardrex: Yes ... great idea to add this. I placed it in an IMPORTANT note immediately after the paragraph with the remark. This is a *teachable moment*, to borrow an educational slogan, on this because we have the reader's focus on how this code is working and we're right in the thick of the whole render mode explanation in the article. Great idea to call it out.

PR #31503, opened 2024-01-18
Improve auth state provider guidance

### LINKED ISSUE #31496: Improve auth state provider guidance for BWAs
### Description

Following the documentation I came across an issue configuring authentication and authorization for a Blazor Server app. The [Authorize] attribute kept redirecting to (I assume) my authentication scheme's fallback path instead of just displaying the fallback message "Not authorized. I used Cookie Authentication. Neither role-based, policy-based nor the default policy was recognized. 
After some research I've found this workaround to fix this problem. [https://github.com/dotnet/aspnetcore/issues/52317#issuecomment-1830673284](url)

The workaround just inserts a middleware step and assumingly skips the redirections issue.

Please add this information into the documentation since it seems to be a recent problem since the introduction of .NET 8. 
Thank you!

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/security/?view=aspnetcore-8.0#additional-resources

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/security/index.md

### Document ID

1247ee64-fbc9-ce29-0710-f6e59380f5ee

### Article author

@guardrex

guardrex:
Fixes #31496

@surayya-MS ... Ignore my ping. I just found out that you joined a different team.

Thanks @PMcVries! :rocket: ... This isn't the end of the story. There's a lot more work to do on BWA security. This at least calls out Surayya's suggestion to take an approach similar to what the BWA template does. I have a tracking work item separate from your issue to work further on this node of docs, and I'll get to it as soon as I can 🏃‍♂️🏃‍♂️🏃‍♂️🏃‍♂️🏃‍♂️🏃‍♂️🏃‍♂️🏃‍♂️.

Mackinnon, I stole 🚓👮 Surayya's remarks from https://github.com/dotnet/aspnetcore/issues/52317#issuecomment-1824624057 to get this coverage going. I don't have the workaround described at https://github.com/dotnet/aspnetcore/issues/52317#issuecomment-1830673284 because it sounds like it isn't the best way to manage the scenario. However, I can add it to this if you like, or I can cross-link to the PU issue (e.g., For more information, see XXXXXX.).

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/security/server/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/daca0f1cb8697625adf69e7588e4e99a37028417/aspnetcore/blazor/security/server/index.md) | [Secure ASP.NET Core server-side Blazor apps](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/server/index?branch=pr-en-us-31503) |

### DIFF
--- a/aspnetcore/blazor/security/server/index.md
+++ b/aspnetcore/blazor/security/server/index.md
@@ -173,6 +173,25 @@ For more information on the Blazor Identity UI and guidance on integrating exter
 
 [!INCLUDE[](~/includes/aspnetcore-repo-ref-source-links.md)]
 
+## Manage authentication state in Blazor Web Apps
+
+*This section applies to Blazor Web Apps that adopt:*
+
+* *Interactive server-side rendering (interactive SSR) and CSR.*
+* *Client-side rendering (CSR).*
+
+A client-side authentication state provider is only used within Blazor and isn't integrated with the ASP.NET Core authentication system. During prerendering, Blazor respects the metadata defined on the page and uses the ASP.NET Core authentication system to determine if the user is authenticated. When a user navigates from one page to another, a client-side authentication provider is used. When the user refreshes the page (full-page reload), the client-side authentication state provider isn't involved in the authentication decision on the server. Since the user's state isn't persisted by the server, any authentication state maintained client-side is lost.
+
+To address this, the best approach is to perform authentication within the ASP.NET Core authentication system. The client-side authentication state provider only takes care of reflecting the user's authentication state. Examples for how to accomplish this with authentication state providers are demonstrated by the Blazor Web App project template:
+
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/security/server/index.md
```diff
@@ -173,6 +173,25 @@ For more information on the Blazor Identity UI and guidance on integrating exter
 
 [!INCLUDE[](~/includes/aspnetcore-repo-ref-source-links.md)]
 
+## Manage authentication state in Blazor Web Apps
+
+*This section applies to Blazor Web Apps that adopt:*
+
+* *Interactive server-side rendering (interactive SSR) and CSR.*
+* *Client-side rendering (CSR).*
```
MackinnonBuck, line 181: Are these acronyms used elsewhere? I haven't seen them before.

MackinnonBuck (approved):
Just a small question, but other than that, LGTM!

aspnetcore/blazor/security/server/index.md
guardrex: Yes, right from the start ...

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/render-modes?view=aspnetcore-8.0

Set up in consultation with Dan. Based on his remarks about industry use, "client-side rendering"/"CSR" doesn't get the "interactive" word applied, and there's an dagger remark there under the table to call that out :point_up:.

I always spell these out (static SSR/interactive SSR) first in a section (e.g., "interactive server-side rendering (interactive SSR)"), and then I use "interactive SSR" throughout the section. I figured that "SSR" might be too cryptic for many devs and that it always should be spelled out initially before using the acronym.

PR #31573, opened 2024-01-25
Revise template recommendation

### LINKED ISSUE #31572: Calling Blazor Web App 'recommended' with no justification is confusing
### Description

There are several places in the Blazor documentation where Blazor Web App, as opposed to Blazor WebAssembly Standalone App, is marked as 'recommended' with no link or justification, which I find quite confusing.  Surely in many circumstances the Standalone App is appropriate, so why is it "not recommended"?  The documentation should be made much clearer, or if there's no actual reason to recommend Blazor Web App, 'recommended' should just be removed.

### Page URL

https://learn.microsoft.com/en-gb/aspnet/core/blazor/tooling?view=aspnetcore-8.0&pivots=windows

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/tooling.md

### Document ID

dd777362-b92d-cac6-2343-a7aaf31d0bdf

### Article author

@guardrex

guardrex:
Fixes #31572

Thanks @jez9999! 🚀 

As @jez9999 implies, recommending the BWA template doesn't mirror any remarks elsewhere (Blazor *Overview* and *Hosting Models* docs in particular) that the choice between BWA or WASM has anything other than to do with the specification required by the organization for the app. There's no obvious cosmic reason to avoid WASM apps if they're a perfect fit for the spec.

Alternatively if there is some reason that I'm unaware of to favor BWAs outside of the benefits that we've already covered, I think we should state what that is, probably by also including it in the Blazor *Overview* and *Hosting Models* articles.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/tooling.md](https://github.com/dotnet/AspNetCore.Docs/blob/ba92006ed5c34ecaefed0b0c53fdc5cf57753916/aspnetcore/blazor/tooling.md) | [Tooling for ASP.NET Core Blazor](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/tooling?branch=pr-en-us-31573) |

### DIFF
--- a/aspnetcore/blazor/tooling.md
+++ b/aspnetcore/blazor/tooling.md
@@ -25,7 +25,7 @@ To create a Blazor app on Windows, use the following guidance:
 
 * Create a new project using one of the available Blazor templates:
 
-  * **Blazor Web App** (*recommended*): Creates a Blazor web app that supports interactive server-side rendering (interactive SSR) and client-side rendering (CSR).
+  * **Blazor Web App**: Creates a Blazor web app that supports interactive server-side rendering (interactive SSR) and client-side rendering (CSR).
   * **Blazor WebAssembly Standalone App**: Creates a standalone client web app that can be deployed as a static site.
 
 Select **Next**.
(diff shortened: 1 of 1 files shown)


guardrex:
I chatted with Dan offline about it. It's recommended for getting started in order to learn about server-side and client-side features. I'm calling that out explicitly. We'll leave the removal of "recommended" in the second spot, as it really isn't needed there.

PR #31597, opened 2024-01-29
Fix typo in routing.md

ericmutta:
The sentence:

> There are several techniques and optimizations **can** be applied...

Should read:

> There are several techniques and optimizations **which** can be applied...

It may be worth going further and adding some commas somewhere in there so it reads:

> There are several techniques and optimizations`,` which can be applied to routes`,` that will largely improve this scenario:

...but comma placement is tricky so I defer to the team on that point 👍

### DIFF
--- a/aspnetcore/fundamentals/routing.md
+++ b/aspnetcore/fundamentals/routing.md
@@ -933,7 +933,7 @@ It is unlikely for an app to run into a situation where this is a problem unless
 
 #### How to address this issue
 
-There are several techniques and optimizations can be applied to routes that will largely improve this scenario:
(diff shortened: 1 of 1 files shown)


guardrex (approved):
Thanks @ericmutta! 🎸

guardrex:
We'll go with "that" over "which" to avoid the comma concern. I also removed "will" to keep it in present tense.

guardrex:
I'm going to do another patch on that ... I prefer "to" for the second "that" in the sentence. I'd also like to kill the "etc" in the list because that's expressly forbidden by our style manual.

PR #31625, opened 2024-01-31
Suggestion: Update best practice Fire-And-Forget example to use AsyncScope

jufa2401:
###  Fixes:  https://github.com/dotnet/AspNetCore.Docs/issues/31628

I think the usage of `AsyncServiceScope` should be encouraged in this example, given that the example shows usage of resolving a service that implements `IAsyncDisposable`.

I realize there is some documentation elsewhere in this repository referencing `IServiceScope` and I don't think there are other places referencing `AsyncServiceScope`, so this could be a little misleading/incohesive.

### DIFF
--- a/aspnetcore/performance/performance-best-practices/samples/3.0/Controllers/FireAndForgetSecondController.cs
+++ b/aspnetcore/performance/performance-best-practices/samples/3.0/Controllers/FireAndForgetSecondController.cs
@@ -59,7 +59,7 @@ namespace performance_best_practices.Controllers
             {
                 await Task.Delay(1000);
 
-                using (var scope = serviceScopeFactory.CreateScope())
+                await using (var scope = serviceScopeFactory.CreateAsyncScope())
                 {
                     var context = scope.ServiceProvider.GetRequiredService<ContosoDbContext>();
 


guardrex:
Hello @jufa2401 ... We prefer an issue for a change like this for a couple of reasons ...

* It will ping the author and maintainer of the example.
* There might be text changes required to go along with the change(s).

Can you open an issue from the bottom of ...

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/best-practices?view=aspnetcore-8.0

... using the **Open a documentation issue** link?

After the issue is opened, add a remark in your opening comment :point_up: with a "fixes" for the issue number ...

```
Fixes #........
```

jufa2401:
> Hello @jufa2401 ... We prefer an issue for a change like this for a couple of reasons ...
> 
> * It will ping the author and maintainer of the example.
> * There might be text changes required to go along with the change(s).
> 
> Can you open an issue from the bottom of ...
> 
> https://learn.microsoft.com/en-us/aspnet/core/fundamentals/best-practices?view=aspnetcore-8.0
> 
> ... using the **Open a documentation issue** link?
> 
> After the issue is opened, add a remark in your opening comment ☝️ with a "fixes" for the issue number ...
> 
> ```
> Fixes #........
> ```

Thank you, will do 👍

Rick-Anderson:
@jufa2401  looks like a nice improvement.

mjrousos (approved):
Looks good to me. Thanks for the improvement, @jufa2401!

PR #31662, opened 2024-02-02
Avoid referring to components or RenderFragment parameters as "tags"

hakenr:
I recommend not referring to RenderFragment parameters and components as "tags".
In other parts of the documentation, the term "template" is commonly used for RenderFragment parameters, and "content" is also an appropriate choice. Although "parameter" might be the most precise term, it could be considered too technical for some contexts.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/class-libraries.md](https://github.com/dotnet/AspNetCore.Docs/blob/76a70966921f0318e2b438ec5fc145e54257068e/aspnetcore/blazor/components/class-libraries.md) | [Consume ASP.NET Core Razor components from a Razor class library (RCL)](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/class-libraries?branch=pr-en-us-31662) |
| [aspnetcore/blazor/fundamentals/routing.md](https://github.com/dotnet/AspNetCore.Docs/blob/76a70966921f0318e2b438ec5fc145e54257068e/aspnetcore/blazor/fundamentals/routing.md) | [ASP.NET Core Blazor routing and navigation](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?branch=pr-en-us-31662) |
| [aspnetcore/blazor/security/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/76a70966921f0318e2b438ec5fc145e54257068e/aspnetcore/blazor/security/index.md) | [ASP.NET Core Blazor authentication and authorization](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/index?branch=pr-en-us-31662) |

### DIFF
--- a/aspnetcore/blazor/components/class-libraries.md
+++ b/aspnetcore/blazor/components/class-libraries.md
@@ -308,7 +308,7 @@ To provide `Component1`'s `my-component` CSS class, link to the library's styles
 
 To make routable components in the RCL available for direct requests, the RCL's assembly must be disclosed to the app's router.
 
-Open the app's `App` component (`App.razor`). Add or update the `AdditionalAssemblies` parameter of the `<Router>` tag to include the RCL's assembly. In the following example, the `ComponentLibrary.Component1` component is used to discover the RCL's assembly.
+Open the app's `App` component (`App.razor`). Add or update the `AdditionalAssemblies` parameter of the `<Router>` component to include the RCL's assembly. In the following example, the `ComponentLibrary.Component1` component is used to discover the RCL's assembly.
 
 ```razor
 AdditionalAssemblies="new[] { typeof(ComponentLibrary.Component1).Assembly }"
--- a/aspnetcore/blazor/components/class-libraries.md
+++ b/aspnetcore/blazor/components/class-libraries.md
@@ -308,7 +308,7 @@ To provide `Component1`'s `my-component` CSS class, link to the library's styles
 
 To make routable components in the RCL available for direct requests, the RCL's assembly must be disclosed to the app's router.
 
-Open the app's `App` component (`App.razor`). Add or update the `AdditionalAssemblies` parameter of the `<Router>` component to include the RCL's assembly. In the following example, the `ComponentLibrary.Component1` component is used to discover the RCL's assembly.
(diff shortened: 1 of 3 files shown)


guardrex:
Hello @hakenr ...

* On this one ...

  > Add or update the `AdditionalAssemblies` parameter of the `<Router>` tag ...

  ... is a mistake for a different reason. It means what it says about the `<Router>` tag, but I was probably thinking the word "attribute" (i.e., HTML tag attribute) when I typed "parameter" because the instruction is telling the reader what to do ***in markup***. Therefore, I'll make that update. When I refer to "tag" in spots around the docs and include the angle brackets in inline code for the markup, I do mean "tag" because I'm remarking on markup.

* On the rest, I might agree with your suggestion for "content" in several spots. I'm not too agreeable at the moment with the word "template" because "template," defined generally as 'a preset format', is conceptually the component with the render fragment parameter (e.g., our use of "templated components"), not what sets the content of the render fragment parameter. The render fragment itself is probably only a "template" if it were to accept arbitrary content from something outside of where the render fragment content is assigned to the parameter. I agree that "parameter" might be reasonable to consider in at least some cases. I don't think its too technical, and those tags could be replaced by API doc cross-links. IMO, the examples help new devs and new-to-.NET/new-to-Blazor devs a great deal when the API docs are cross-linked and I don't use inline tags. Perhaps, these can go that direction. Another consideration is that "tags" (plural) with ***opening and closing render fragment tags*** (not just the opening tag) in code examples is probably better than what I have ... I mean if "tag(s)" sticks around at all.

I'll make an edit pass on your suggestions and see if a change to "content" feels good. We can discuss this further after I make a commit.

hakenr:
What I didn't make clear in my original comment is that, for me, the term "tag" in the context of markup code strictly applies to "HTML tags," and I would never use it for Razor component markup (e.g., `<Router>`) or RenderFragment parameters (e.g., `<NotFound>`).
You might have a different opinion on this, but it was this perspective that led me to propose the PR.

Regarding the "parameter" term. If you don't think its too technical, go for it (definitely much better than "tag").

guardrex:
Ok .... looking good here.

This focuses on "paramter" and "content" now.

I'll wait a bit to see if you want to look again. I'll need to merge this fairly *FAST* tho 🏃‍♂️. Merge conflicts 😈 are a pain 😆, and PRs are going in hot right right now for further .NET 8 updates.

guardrex:
I'm going to roll with these two PRs. I don't want to leave these lying around for too long, as the ***gremlins*** 😈 will spot it and hammer me with a merge conflict 😆.

PR #31663, opened 2024-02-02
Blazor built-in components include components from project templates

hakenr:
The list of Razor components includes both built-in components provided by the Blazor framework and those that come with Blazor project templates.
It's important to note this distinction on the page to prevent confusion, especially when working on projects not based on the official Blazor project templates.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/built-in-components.md](https://github.com/dotnet/AspNetCore.Docs/blob/8282bac47d68247d3c6c6a1e5cc5d6a5660f0232/aspnetcore/blazor/components/built-in-components.md) | [ASP.NET Core built-in Razor components](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/built-in-components?branch=pr-en-us-31663) |

### DIFF
--- a/aspnetcore/blazor/components/built-in-components.md
+++ b/aspnetcore/blazor/components/built-in-components.md
@@ -12,7 +12,7 @@ uid: blazor/components/built-in-components
 
 [!INCLUDE[](~/includes/not-latest-version.md)]
 
-The following Razor components are built-in within the Blazor framework, or serve as structural elements of Blazor project templates:
+The following built-in Razor components are provided by the Blazor framework. For more information on built-in components that are used by the Blazor project templates, see <xref:blazor/project-structure>.
 
 :::moniker range=">= aspnetcore-8.0"
 
--- a/aspnetcore/blazor/components/built-in-components.md
+++ b/aspnetcore/blazor/components/built-in-components.md
@@ -12,7 +12,7 @@ uid: blazor/components/built-in-components
 
 [!INCLUDE[](~/includes/not-latest-version.md)]
 
-The following built-in Razor components are provided by the Blazor framework:
+The following Razor components are built-in within the Blazor framework, or serve as structural elements of Blazor project templates:
 
 :::moniker range=">= aspnetcore-8.0"
 


guardrex:
I don't agree with the exact recommendation for a couple of reasons, but I think we can call out this sentiment with some alternative text. Here are my concerns about the update ...

* It's not an *either-or* situation. If the component is in a project template, ***it is part of the framework*** because the framework includes the project templates. These are all inheriting (directly or indirectly) `ComponentBase` in the framework. The project templates are using them from the framework API, and devs are free to use them regardless of their presence in the project templates. Also, what's in the project templates is subject to change, while these (versioned for each release) are present in the framework regardless of project template changes.
* For describing their individual use in the project templates, that's traditionally been a job of the [Project Structure](https://learn.microsoft.com/en-us/aspnet/core/blazor/project-structure?view=aspnetcore-8.0) article. What this article could do is indicate on a per-component basis which ones are in a project template and cross-link to the *Project Structure* article. At this time, the article has the sole purpose of acting as a directory so that readers can get to the content associated with the introduction and general explanation of the framework-provided components ***QUICKLY*** 🏎️💨. Because of the churn of project templates, I don't want to take on cross-linking. Yes, it might be a good idea, but there is a maintenance cost 💰, and with management's recent decision (.NET 8 was the first time) to document every preview release, it's just too much to commit to at this time with the current staffing here (***ME!*** 🦖😆). I'm already insanely busy during the release cycle, and I must be very careful about placing additional burdens on myself. However, let's make a change here to call it out in a general way.

I'll place a similar remark to what you're proposing ..................... stand-by .............

guardrex:
Ok ... How about that? I think that expresses your sentiment and cross-links where readers can find out about the ones used by the templates ***without actually indicating individual components used in templates***. *That's the thing that scares me in terms of my yearly workload for releases.* 😨

hakenr:
@guardrex, I understand that this page serves as a quick guide to components. From my perspective, components that are not part of the Blazor NuGet packages but are "just" in the new project template cannot be classified as "built-in Blazor components," and here our views diverge, but I get it, I haven't convinced you. :-)

My view:
Then where would the line be drawn? Are `Counter`, `Home`, or `Weather` considered built-in Blazor components when they are practically in every project template?
Or what about `Routes`, which the documentation often refers to? Or components from security templates like `LoginDisplay` and others.
I truly believe that lumping these components together with components from NuGet packages is misleading. Many people new to the world of Blazor are explicitly looking to understand what the basic set of components is and what they will need to address with additional libraries.

guardrex:
Yes ... I see what you mean now. I think the ones that are only in the project templates should be removed from the list, allowing the *Project Structure* topic to cover them. The text I added for the cross-link probably should be updated. I'm OOF for the weekend, so I'll work on this further next week. I opened  #31679 to work on it further.

PR #31686, opened 2024-02-04
Routing - use @key in NavLink loop warning (instead of NavLink href)

hakenr:
I believe the `@key` directive attribute should be emphasized whenever there's a loop in the rendering process. Moreover, in this scenario, it could also serve to demonstrate the difference between using `@c` and `@current`. Specifically, using `<NavLink href="@c" ...>` might not be very clear for numeric values of `c` and could lead to confusion for readers.

### DIFF
--- a/aspnetcore/blazor/fundamentals/routing.md
+++ b/aspnetcore/blazor/fundamentals/routing.md
@@ -1592,8 +1592,8 @@ The following HTML markup is rendered:
 > @for (int c = 0; c < 10; c++)
 > {
 >     var current = c;
->     <li ...>
->         <NavLink ... href="@c">
+>     <li @key="@c" ...>
+>         <NavLink ...>
 >             <span ...></span> @current
 >         </NavLink>
 >     </li>
@@ -1607,8 +1607,8 @@ The following HTML markup is rendered:
 > ```razor
 > @foreach (var c in Enumerable.Range(0,10))
 > {
->     <li ...>
->         <NavLink ... href="@c">
+>     <li @key="@c" ...>
+>         <NavLink ...>
 >             <span ...></span> @c
 >         </NavLink>
 >     </li>


guardrex:
> whenever there's a loop in the rendering process

Not really ... `@key` shouldn't be required and isn't recommended when the rendered content ***is static*** because it isn't performant. Also, introducing it here is a bit distracting from the example because it introduces an additional concept that must be explained.

Also, I don't agree with removing the `href` because it demonstrates that the content of the `NavLink` that's at stake, not the C# value of `c` (`@c`). However, I propose a different change to make the `href` assignment relevant ...

```diff
- href="@c"
+ href="product-number/@c"
```

... with ...

```diff
- <span ...></span> @c
+ <span ...></span> Product #@current
```

 ... or in the second case with `foreach` ...

```diff
- <span ...></span> @c
+ <span ...></span> Product #@c
```

... and possibly a slight content addition mentioning that the nav menu lists a handful of product links.

WRT my understanding on the C# value of `c` ... I would like to ask Mackinnon to put an 👁️ on this coverage.

Mackinnon ... I don't think we should have `@key` here for the reason that I gave, but let me know if you disagree and why. If we include it, we're going to need to say why it's here and cross-link coverage on it.

Second ... IIRC/AFAIK, but I could be wrong ... the value of `c` problem to overcome is in the ***content*** of the `NavLink`, not in general use of `c` otherwise. I think that because when this was discussed with the PU and the content was placed we were careful to design the example to show `c` (`@c`) in use outside of the content (i.e., in the `href`) to demonstrate no ill effects. I don't do "***random***" 😄 ... I might be wrong about something 🙈 ... but I definitely don't do *random* things 😆.

hakenr:
You're right; for static SSR, the `@key` doesn't make much sense since there are no updates happening (or are there with enhanced forms or enhanced navigation?).

On the other hand, the importance of `@key` for interactive scenarios, in my opinion, is not sufficiently emphasized in the documentation, and in practice, I repeatedly encounter its underestimation (for example, with the argument "it only gains a little more performance sometimes"), even though `@key` is not just about performance and is absolutely necessary in many scenarios. See, for example:
* https://github.com/dotnet/aspnetcore/issues/22355
* https://github.com/havit/Havit.Blazor/issues/524

The [documentation for @key](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/element-component-model-relationships?view=aspnetcore-8.0&branch=pr-en-us-31686#when-to-use-key) states:
> Typically, it makes sense to use `@key` whenever a list is rendered (for example, in a `foreach` block) and a suitable value exists to define the `@key`.

...so it seemed good to me to use `@key` in a `for`/`foreach` example to spread best practices.

(BTW. as far as I know, in React, a _key_ is required in loops, and I have a feeling that Blazor will gradually move in that direction too, :-D I guess we will see at least code analyzer for this suggestion soon.)

guardrex:
Yeah ... but I didn't say that. I merely said that it wouldn't be performant in this example. I defer to Mackinnon's wisdom on just how far `@key` should be pushed/recommended.

Mackinnon ... let's focus just on this example/this article for the moment ... ***because*** .......................

WRT the wider issue of how `@key` is described and demonstrated ... @hakenr, open an issue from that article. More examples can be added to the article, and additional cross-links can be added around the repo to that article where it might be relevant. Mackinnion might leave comments on your new issue for when it's reached. This is just a patch PR, so let's not "work" the problem from here. Patch PRs are only for patches, and we require an issue for any significant work.

Open the issue from this one .......

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/element-component-model-relationships?view=aspnetcore-8.0

Note that working on it will be a lower-priority task. There's still a large amount of 8.0 churn right now, so I think working on that would be pushed off at least a month (***or three!*** 😆 ... probably three actually 🙈😆 I'm sorry to say) because the focus now is on shoring up all of the 8.0 scenarios that need critical work 🏃‍♂️🏃‍♂️🏃‍♂️🏃‍♂️🏃‍♂️🏃‍♂️.

MackinnonBuck:
I agree with @guardrex's assessment here - the content in the loop is effectively static in the sense that the same 10 items will always be rendered. If items between 0 and 9 could be dynamically removed, then `@key` _might_ be useful so that the right state gets preserved. But this example doesn't demonstrate that, so readers might scratch their heads as to why `@key` is necessary. I think we should leave `@key` out and keep things how they are.

Issue #31691, opened 2024-02-05
PU review of *Pass tokens to a server-side Blazor app* section


guardrex:
### Description

Per our offline discussion, either (or both) @halter73 and @JeremyLikness are to review the *Pass tokens to a server-side Blazor app* section ...

* Is the content correct?
* Are there any MIA remarks to make?
* Are there any ***gotchas*** 😈 to call out?

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/security/server/additional-scenarios?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/security/server/additional-scenarios.md

### Document ID

c98be365-408d-7ee6-cb74-14c44d01b0b8

### Article author

@guardrex

timohermans:
I just found this issue, so I'll just leave my two cents here.

I tried applying this to my own blazor server 8 app. When adding a **scoped** TokenProvider service, the set value will **not** be handed over to the circuit. The value will initially be set during the prerender, but on the second oninitializedasync null.

What did work was marking the service as **singleton**, though I don't think this is very secure.

GStoynev:
https://github.com/dotnet/AspNetCore.Docs/issues/31759 got closed, but let me share what worked for me, if helpful to patch the documentation or to others searching for solution to the issue:

Capturing a cookie inline like this and then passing it as a Parameter to the Routes component where it's stored in the scoped provider seems to do the trick:
```
<body>
@{
    var cookie = HttpContext?.Request.Cookies["UI"];
    Logger.LogInformation("Cookie captured: {UICookie}", cookie);
}

<Routes Cookie="@cookie" @rendermode="InteractiveServer"/>
<script src="_framework/blazor.web.js"></script>

</body>
```

GStoynev:
Another suggestion for that section of the documentation:

Mixing OpenIdConnect in the example creates the impression that there might be some magic happening there, that alleviates the issue with the pre-rendering and the duplicated TokenProvider.

A simple example with a cookie would have been simpler and to the point.

timohermans:
@GStoynev wow that is a very easy solution 🤔. 

One other thing I found out yesterday is that the httpcontext within a  HttpClient, wherever you inject it inside the circuit. Which is also what the documentation does, I believe.

guardrex:
[...]

Issue #31689, opened 2024-02-05
Transient disposable section updates


guardrex:
### Description

Per the offline discussion ...

* Revise the utility base class (`OwningComponentBase`) section. We might say it's more important on the server than on the client ... that the client-side use is still a problem. They should be avoided client-side, too.
* Move the code completely into the sample apps, making it a fully working demo. 🔪 the code from the article. Give readers links to the code in the sample apps.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/dependency-injection?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/fundamentals/dependency-injection.md

### Document ID

d9aabe95-69e9-0a03-81ff-429914137f9d

### Article author

@guardrex

PR #31693, opened 2024-02-05
Offline discussion updates for 8.0

### LINKED ISSUE #31688: Blazor 8.0 offline discussion updates
### Description

This is a general issue to track the updates from offline discussion for 8.0 content. I believe most (or all) of the updates can be batched on one PR. I'll open followup issues/PRs if needed for anything that can't be batched on a single PR.

* [x] Transient disposables ... **WORK SEPARATELY ON #31689**
* [x] BWA integration into the ASP.NET Core endpoint routing configuration. ... ***NO CHANGES NEEDED***
  * Optional params are covered: https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-8.0#route-parameters
  * Route constraints are covered (and optional params with constraints): https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?view=aspnetcore-8.0#route-constraints
  * "Custom constraints" and "Regex constraints" were never added to the article per our goal to generally avoid covering features that Blazor doesn't offer. If it becomes a ❓ for readers and we receive feedback, I'll add something later.
* [x] `MapFallBackToPage`
* [x] "Don't place a <script> tag in a component file"
* [x] *Passing tokens* section ... **WORK SEPARATELY ON #31691**
* [x] *Cross-origin protection* section of the *Threat mitigation* topic ... **SKIPPING FOR NOW** ... we might add code examples if reader feedback indicates confusion.
* [x] *Configure and use gRPC in components* section ... **PARTIALLY ADDRESSED** by resurfacing the content sans any remarks about hosted WASM or any code directly tied to the server app of a hosted WASM solution. Dan said he can/will supply a code example.
* [x] *Secure a SignalR hub* section ... **PARTIALLY ADDRESSED** by resurfacing content sans any hosted WASM bits. Jeremy and/or Stephen will need to address the token question by DR.

### Page URL

Various

### Content source URL

Various

### Document ID

None

### Article author

@guardrex

guardrex:
Fixes #31688
Addresses #28161 

## Items to be worked on separate issues

* ~[Transient disposable section updates](https://github.com/dotnet/AspNetCore.Docs/issues/31689): I have everything I need to resolve this, and I should reach it on Tuesday 🤞🍀.~ **UPDATE**: This is resolved now.
* [PU review of *Pass tokens to a server-side Blazor app* section](https://github.com/dotnet/AspNetCore.Docs/issues/31691): Currently assigned to both Stephen and Jeremy, but perhaps only one pass by one PU 🐱 will suffice.

## *Configure and use gRPC in components* section

Dan ... You said you'd provide a code example, but it 🤔 ***seems like*** the current code is ok. The updates to the section are on the DIFF, so you can see the current code there. The focus of the updates thus far is just to 🔪 OUT the little bits of text that pertain to hosted WASM and updating the versioning to return the content for 8.0+. Let me know if I should 🛑 HOLD this PR for new or additional code (or open a separate issue for it). 👂👂👂

## *Secure a SignalR hub* section

Dan asks Stephen/Jeremy ...

> Should we be recommending using the auth cookie instead? Are there any CSRF concerns here that would then need to be addressed?

The updates to the section are on the DIFF, so the current code is there for inspection. Thus far ... again ... the updates on the PR just focus on 🔪 OUT the little bits pertaining to hosted WASM and updating the versioning to return the content for 8.0+.

## Skipped bits

Everything else that we discussed that isn't on the PR doesn't require updates because the current coverage is good.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/host-and-deploy/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/246e5c0d17394c1930b6bf511b10ca433b9607ff/aspnetcore/blazor/host-and-deploy/index.md) | [Host and deploy ASP.NET Core Blazor](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/index?branch=pr-en-us-31693) |
| [aspnetcore/blazor/javascript-interoperability/call-dotnet-from-javascript.md](https://github.com/dotnet/AspNetCore.Docs/blob/246e5c0d17394c1930b6bf511b10ca433b9607ff/aspnetcore/blazor/javascript-interoperability/call-dotnet-from-javascript.md) | [Call .NET methods from JavaScript functions in ASP.NET Core Blazor](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-dotnet-from-javascript?branch=pr-en-us-31693) |
| [aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md](https://github.com/dotnet/AspNetCore.Docs/blob/246e5c0d17394c1930b6bf511b10ca433b9607ff/aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md) | [Call JavaScript functions from .NET methods in ASP.NET Core Blazor](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-javascript-from-dotnet?branch=pr-en-us-31693) |
| [aspnetcore/blazor/javascript-interoperability/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/246e5c0d17394c1930b6bf511b10ca433b9607ff/aspnetcore/blazor/javascript-interoperability/index.md) | [ASP.NET Core Blazor JavaScript interoperability (JS interop)](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/index?branch=pr-en-us-31693) |
| [aspnetcore/blazor/security/webassembly/additional-scenarios.md](https://github.com/dotnet/AspNetCore.Docs/blob/246e5c0d17394c1930b6bf511b10ca433b9607ff/aspnetcore/blazor/security/webassembly/additional-scenarios.md) | [aspnetcore/blazor/security/webassembly/additional-scenarios](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/additional-scenarios?branch=pr-en-us-31693) |
| [aspnetcore/blazor/security/webassembly/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/246e5c0d17394c1930b6bf511b10ca433b9607ff/aspnetcore/blazor/security/webassembly/index.md) | [Secure ASP.NET Core Blazor WebAssembly](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/index?branch=pr-en-us-31693) |
| [aspnetcore/blazor/tutorials/signalr-blazor.md](https://github.com/dotnet/AspNetCore.Docs/blob/246e5c0d17394c1930b6bf511b10ca433b9607ff/aspnetcore/blazor/tutorials/signalr-blazor.md) | [[Visual Studio](#tab/visual-studio)](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/tutorials/signalr-blazor?branch=pr-en-us-31693) |

### DIFF
--- a/aspnetcore/blazor/javascript-interoperability/call-dotnet-from-javascript.md
+++ b/aspnetcore/blazor/javascript-interoperability/call-dotnet-from-javascript.md
@@ -1422,10 +1422,19 @@ Using JS modules to load JS is described in this article in the [JavaScript isol
 
 :::moniker-end
 
-<!-- UPDATE 8.0 Confirm for >=8.0 now that the warning is being removed. -->
+:::moniker range=">= aspnetcore-8.0"
+
+> [!WARNING]
+> Only place a `<script>` tag in a component file (`.razor`) if the component is guaranteed to adopt [static server-side rendering (static SSR)](xref:blazor/fundamentals/index#client-and-server-rendering-concepts) because the `<script>` tag can't be updated dynamically.
+
+:::moniker-end
+
+:::moniker range="< aspnetcore-8.0"
 
 > [!WARNING]
 > Don't place a `<script>` tag in a component file (`.razor`) because the `<script>` tag can't be updated dynamically.
+
+:::moniker-end
 
 :::moniker range=">= aspnetcore-5.0"
 
--- a/aspnetcore/blazor/host-and-deploy/index.md
+++ b/aspnetcore/blazor/host-and-deploy/index.md
@@ -398,11 +398,9 @@ The examples in this section focus on supplying the app base path from app setti
 * <xref:blazor/fundamentals/configuration>
 * <xref:fundamentals/configuration/index>
 
-:::moniker range="< aspnetcore-8.0"
-
 ## Blazor Server `MapFallbackToPage` configuration
 
-<!-- UPDATE 8.0 Update for BWAs -->
+*This section only applies to Blazor Server apps. <xref:Microsoft.AspNetCore.Builder.RazorPagesEndpointRouteBuilderExtensions.MapFallbackToPage%2A> isn't supported in Blazor Web Apps and Blazor WebAssembly apps.*
 
 In scenarios where an app requires a separate area with custom resources and Razor components:
 
@@ -427,8 +425,6 @@ In scenarios where an app requires a separate area with custom resources and Raz
 
   app.Run();
   ```
-
-:::moniker-end
 
 :::moniker range="< aspnetcore-8.0"
 
--- a/aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md
+++ b/aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md
@@ -432,10 +432,19 @@ For information on isolating scripts in [JS modules](https://developer.mozilla.o
 
 :::moniker-end
 
-<!-- UPDATE 8.0 Confirm for >=8.0 now that the warning is being removed. -->
+:::moniker range=">= aspnetcore-8.0"
+
+> [!WARNING]
+> Only place a `<script>` tag in a component file (`.razor`) if the component is guaranteed to adopt [static server-side rendering (static SSR)](xref:blazor/fundamentals/index#client-and-server-rendering-concepts) because the `<script>` tag can't be updated dynamically.
+
+:::moniker-end
+
+:::moniker range="< aspnetcore-8.0"
 
 > [!WARNING]
 > Don't place a `<script>` tag in a component file (`.razor`) because the `<script>` tag can't be updated dynamically.
+
+:::moniker-end
 
 :::moniker range=">= aspnetcore-5.0"
 
--- a/aspnetcore/blazor/javascript-interoperability/index.md
+++ b/aspnetcore/blazor/javascript-interoperability/index.md
@@ -251,10 +251,19 @@ Load JavaScript (JS) code using any of the following approaches:
 
 :::moniker-end
 
-<!-- UPDATE 8.0 Confirm for >=8.0 now that the warning is being removed. -->
+:::moniker range=">= aspnetcore-8.0"
 
 > [!WARNING]
-> Don't place a `<script>` tag in a Razor component file (`.razor`) because the `<script>` tag can't be updated dynamically by Blazor.
+> Only place a `<script>` tag in a component file (`.razor`) if the component is guaranteed to adopt [static server-side rendering (static SSR)](xref:blazor/fundamentals/index#client-and-server-rendering-concepts) because the `<script>` tag can't be updated dynamically.
+
+:::moniker-end
+
+:::moniker range="< aspnetcore-8.0"
+
+> [!WARNING]
+> Don't place a `<script>` tag in a component file (`.razor`) because the `<script>` tag can't be updated dynamically.
+
+:::moniker-end
 
 :::moniker range=">= aspnetcore-5.0"
 
--- a/aspnetcore/blazor/security/webassembly/additional-scenarios.md
+++ b/aspnetcore/blazor/security/webassembly/additional-scenarios.md
@@ -1454,25 +1454,32 @@ If tacking on a segment to the authority isn't appropriate for the app's OIDC pr
 
 The list of claims in the ID token changes for v2.0 endpoints. Microsoft documentation on the changes has been retired, but guidance on the claims in an ID token is available in the [ID token claims reference](/entra/identity-platform/id-token-claims-reference).
 
-:::moniker range="< aspnetcore-8.0"
-
-<!-- UPDATE 8.0 This section was written for hosted WASM.
-     Check with PU to see if we'll maintain it with mods
-     in the BWA/WebAssembly world. -->
-
 ## Configure and use gRPC in components
 
 To configure a Blazor WebAssembly app to use the [ASP.NET Core gRPC framework](xref:grpc/index):
 
 * Enable gRPC-Web on the server. For more information, see <xref:grpc/grpcweb>.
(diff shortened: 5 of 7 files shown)


aspnetcore/blazor/javascript-interoperability/call-dotnet-from-javascript.md
```diff
@@ -1422,11 +1422,20 @@ Using JS modules to load JS is described in this article in the [JavaScript isol
 
 :::moniker-end
 
-<!-- UPDATE 8.0 Confirm for >=8.0 now that the warning is being removed. -->
+:::moniker range=">= aspnetcore-8.0"
+
+> [!WARNING]
+> Only place a `<script>` tag in a component file (`.razor`) if the component is guaranteed to adopt [static server-side rendering (static SSR)](xref:blazor/fundamentals/index#client-and-server-rendering-concepts) because the `<script>` tag can't be updated dynamically.
```
halter73, line 1428: @javiercn Is this warning worded strong enough for you? I remember you complaining about people trying to render `<script>` tags in triage.

halter73 (approved):

guardrex:
Thanks, @halter73 ... but you didn't answer Dan's question in the opening remarks ☝️.

halter73:
I think we should recommend using an auth cookie because that's the only way to prerender content that requires authentication.

As for the CSRF concerns, I think we should just link to https://learn.microsoft.com/en-us/aspnet/core/security/anti-request-forgery?view=aspnetcore-8.0. All Blazor endpoints that accept form posts that might be susceptible to CSRF attacks already require an antiforgery token as mentioned in https://learn.microsoft.com/en-us/aspnet/core/blazor/security/?view=aspnetcore-8.0#antiforgery-support, so I don't think there's much to add there. Pinging @javiercn @blowdart in case they disagree

Other types of endpoints that accept form posts either require antiforgery tokens by default (like razor pages, controllers with views, and minimal APIs) or have been long documented not to (like `AddControllers` without views). And the need for antiforgery checks for these other types of endpoints is unrelated to whether or not they are hosted along with a Blazor app.

guardrex:
> I think we should recommend using an auth cookie because that's the only way to prerender content that requires authentication.

It will be a bit out-of-sync with our SignalR guidance ...

> Cookies are specific to browsers. Sending them from other kinds of clients adds complexity compared to sending bearer tokens. Cookie authentication isn't recommended unless the app only needs to authenticate users from the browser client. Bearer token authentication is the recommended approach when using clients other than the browser client.

Cross-ref: https://learn.microsoft.com/en-us/aspnet/core/signalr/authn-and-authz?view=aspnetcore-8.0#cookies-vs-bearer-tokens

We seem to have some cookie bits at ...

https://learn.microsoft.com/en-us/aspnet/signalr/overview/security/hub-authorization#cookie

@halter73 ... Do you want me to try to 🦖 ***RexHack***&trade; 🙈 some kind of approach based on that, or do we want a ***real programmer*** to build out what we'll show for the cookie approach in the SignalR doc?

BTW ... This is the current code that we show is, which I suppose we can maintain for standalone WASM apps ...

```csharp
@using Microsoft.AspNetCore.Components.WebAssembly.Authentication
@inject IAccessTokenProvider TokenProvider
@inject NavigationManager Navigation

...

var tokenResult = await TokenProvider.RequestAccessToken();

if (tokenResult.TryGetToken(out var token))
{
    hubConnection = new HubConnectionBuilder()
        .WithUrl(Navigation.ToAbsoluteUri("/chathub"), 
            options => { options.AccessTokenProvider = () => Task.FromResult(token?.Value); })
        .Build();

  ...
}
```

guardrex:
## 🦖 NOTE TO SELF

After we finish the work for securing a SignalR hub, update the cross-links of the SignalR-Blazor tutorial, which currently only cross-link to the bearer token approach.

halter73:
I think it's fine to continue recommend tokens for cases where there isn't prerendering like standalone WASM or non-browser apps. In those kind of apps, you don't have a browser window to login using Razor pages/components.

You can have a .NET HttpClient login using cookies with `MapIdentityApi` endpoints, but I think it's fair to say if you don't have a browser to do an OIDC flow or sign in using Razor pages/components, then bearer tokens are the better option.

If you're up for it, I'm more than happy to review 🦖 ***RexHack***&trade; version of cookie based SignalR auth using a .NET client. I expect instead of `options.AccessTokenProvider` you'd want to use `HttpMessageHandlerFactory` and configure it with `request.SetBrowserRequestCredentials(BrowserRequestCredentials.Include)` as demonstrated in https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/signalr?view=aspnetcore-8.0#client-side-signalr-cross-origin-negotiation-for-authentication

guardrex:
@halter73 ... *Indeed!* I forgot that section was there. It seems to me that we can just cross-link that from the security article, as I just set up on ...

https://github.com/dotnet/AspNetCore.Docs/pull/31693/commits/fd91e23eac2345b159cb1e774352cf20d606d8c2

WRT ...

> Is this warning worded strong enough for you? I remember you complaining about people trying to render <script> tags in triage.

I already conversed with @javiercn on that point, and that's about what he told me to write into the docs.

WRT anything else ... @danroth27 / @mkArtakMSFT ... are you OK with going ahead and getting this merged?

aspnetcore/blazor/javascript-interoperability/call-dotnet-from-javascript.md
danroth27: Seems fine to me. The reason for not using script tags in interactive components is clearly stated.

danroth27:
> Dan ... You said you'd provide a code example, but it 🤔 seems like the current code is ok.

It looks like the gRPC sample code isn't used to initialize any component state, so yeah, I think it's fine as is. No need to persist and use prerendered state.

danroth27 (approved):
LGTM

PR #31705, opened 2024-02-06
Update transient services guidance

### LINKED ISSUE #31689: Transient disposable section updates

guardrex:
Fixes #31689
Addresses #28161 

Blazor samples were updated on https://github.com/dotnet/blazor-samples/pull/181.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/dependency-injection.md](https://github.com/dotnet/AspNetCore.Docs/blob/210e87575e012d865a5820ca8fc957aed6fdaa79/aspnetcore/blazor/fundamentals/dependency-injection.md) | [ASP.NET Core Blazor dependency injection](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/dependency-injection?branch=pr-en-us-31705) |


aspnetcore/blazor/fundamentals/dependency-injection.md
```diff
@@ -317,15 +317,19 @@ public IMyService MyService { get; set; }
...
 
-In ASP.NET Core apps, scoped services are typically scoped to the current request. After the request completes, any scoped or transient services are disposed by the DI system. Server-side, the request scope lasts for the duration of the client connection, which can result in transient and scoped services living much longer than expected. Client-side, services registered with a scoped lifetime are treated as singletons, so they live longer than scoped services in typical ASP.NET Core apps.
+In non-Blazor ASP.NET Core apps, scoped and transient services are typically scoped to the current request. After the request completes, scoped and transient services are disposed by the DI system.
+
+In interactive server-side Blazor apps, the request scope lasts for the duration of the circuit (the SignalR connection between the client and server), which can result in scoped and transient services living much longer than expected. Therefore, direct use of scoped services should be avoided, and transient services shouldn't be registered or used at all. An alternative approach based on the <xref:Microsoft.AspNetCore.Components.OwningComponentBase> type is described later in this section.
```
MackinnonBuck, line 322: > direct use of scoped services should be avoided

Was this guidance that we had before? Scoped services can be useful when using limiting the scope of a stateful service to a circuit/user (e.g., if you need to manage user-specific data).

MackinnonBuck (approved):
Looks good, just a question about the guidance for using scoped services with server interactivity.

aspnetcore/blazor/fundamentals/dependency-injection.md
guardrex: I see what you mean. It needs more work ... I think it really should say something like ...

> ... can result in scoped and transient services living much longer than expected if you only want the service to exist for the lifetime of a single component. Therefore, direct use of scoped services for single component lifetimes should be avoided, and transient services shouldn't be registered or used at all.

... does that seem better? ... more specific to what this is leading to for the use of `OwningComponentBase`.

guardrex: I'll drop that as a full suggestion here ...

```suggestion
In interactive server-side Blazor apps, the request scope lasts for the duration of the circuit (the SignalR connection between the client and server), which can result in scoped and transient services living much longer than expected for the lifetime of a single component. Therefore, direct use of scoped services for single component lifetimes should be avoided, and transient services shouldn't be registered or used at all. An alternative approach based on the <xref:Microsoft.AspNetCore.Components.OwningComponentBase> type is described later in this section.
```

MackinnonBuck: Yes, that looks great!

hakenr:
@guardrex @MackinnonBuck I'd recommend adding one or two sentences explaining why transient disposables are a problem and why it's important to avoid them.

guardrex:
Already done ... it's in the earlier section on `OwningComponentBase`. These sections are just detection for those who feel that devs might add transient services in the future for some reason or perhaps have a lib add them inadvertently (although I didn't explicitly call that out).

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/dependency-injection?view=aspnetcore-8.0#utility-base-component-classes-to-manage-a-di-scope

hakenr:
@guardrex Not sure if you're referring to this section?

> In ASP.NET Core apps, scoped services are typically tied to the current request. Once the request is completed, any scoped or transient services are disposed of by the DI system. On the server side, the request scope lasts for the duration of the client connection, potentially causing transient and scoped services to live much longer than anticipated. On the client side, services registered with a scoped lifetime behave as singletons, meaning they persist longer than scoped services in typical ASP.NET Core applications.

I've discussed this with several colleagues, and despite their current knowledge and after going through this section of the documentation, they're unsure why transient disposables are particularly problematic, mainly because they haven't been following the Blazor issue tracking and haven't had the chance to look into any related issues:
* https://github.com/dotnet/aspnetcore/issues/43234
* https://github.com/dotnet/aspnetcore/issues/2336
* https://github.com/dotnet/aspnetcore/issues/26676

(Just an observation from the field, you might have different view.)

guardrex:
I'll raise that with Mackinnon on the new PR.

guardrex:
Ok ... it's live now ...

These updates went in yesterday. *I guess I agreed with you!* 😆 ... because I did make quite an update to the lead-in bits.

In spite of this PR being closed, take a look at it now here and let me know if you still have concerns about it. We'll avoid opening ***yet another issue in a SEA of Blazor issues at the moment*** 😩. I'm working fast to get the backlog down because we're starting to document .NET 9 ***NOW*** 😅.

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/dependency-injection?view=aspnetcore-8.0#utility-base-component-classes-to-manage-a-di-scope

hakenr:
It's much clearer now, but it still doesn't explain why specifically transient disposables are problematic. (The introduction mentions transient services in general, which might leave readers questioning why **disposable** services, in particular, should be identified as an issue.)

guardrex:
I think the "disposables" part could be dropped. I'll open an issue to discuss this subject further with Mackinnon.

PR #31739, opened 2024-02-08
Enforce HTTPS in ASP.NET Core: Use Status308PermanentRedirect

0xced:
The previous paragraph says
> Use the fields of the [StatusCodes](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.statuscodes) class for assignments to `RedirectStatusCode`.

But the sample code then uses the status code defined in `System.Net` and a cast to `int`.

### DIFF
--- a/aspnetcore/security/enforcing-ssl/sample-snapshot/6.x/Program3.cs
+++ b/aspnetcore/security/enforcing-ssl/sample-snapshot/6.x/Program3.cs
@@ -1,4 +1,4 @@
-using System.Net;
+using static Microsoft.AspNetCore.Http.StatusCodes;
 
 var builder = WebApplication.CreateBuilder(args);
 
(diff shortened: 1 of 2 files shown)


guardrex:
LGTM, but I think Tom is managing the article on this, so I'll ping him for an 👁️.

aspnetcore/security/enforcing-ssl/sample-snapshot/6.x/Program3.cs
tdykstra, line 1: We should do the same thing for the temporary redirect in the preceding code snippet.

tdykstra (changes requested):
We should do the same thing for the temporary redirect in the preceding code snippet.

0xced:
Good point! Done in 34f3c2d6c8ed3e34242f0516d003c1a3308b7e50.

tdykstra (approved):
Thanks!

PR #31747, opened 2024-02-08
Blazor Startup - sample environment variable name

hakenr:
I think demonstrating how to pass environment variables using a variable named `CONFIGURE_RUNTIME` in the section titled _Configure the .NET WebAssembly runtime_ can be confusing. An inexperienced user might be puzzled about whether using `CONFIGURE_RUNTIME = true` activates something that allows configuring the .NET WebAssembly runtime.

cc @guardrex 

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/startup.md](https://github.com/dotnet/AspNetCore.Docs/blob/a3e456ef165af322e7d48ffd6420941e7e58acfd/aspnetcore/blazor/fundamentals/startup.md) | [ASP.NET Core Blazor startup](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/startup?branch=pr-en-us-31747) |

### DIFF
--- a/aspnetcore/blazor/fundamentals/startup.md
+++ b/aspnetcore/blazor/fundamentals/startup.md
@@ -734,7 +734,7 @@ In `wwwroot/index.html`, remove the default SVG round indicator in `<div id="app
 
 To configure the .NET WebAssembly runtime, use the `configureRuntime` function with the `dotnet` runtime host builder.
 
-The following example sets an environment variable, `CONFIGURE_RUNTIME`, to `true`:
+The following example sets an environment variable, `MY_VARIABLE`, to `my value`:
 
 Blazor Web App:
 
@@ -744,7 +744,7 @@ Blazor Web App:
   Blazor.start({
     webAssembly: {
       configureRuntime: dotnet => {
-        dotnet.withEnvironmentVariable("CONFIGURE_RUNTIME", "true");
+        dotnet.withEnvironmentVariable("MY_VARIABLE", "my value");
       }
     }
   });
@@ -758,7 +758,7 @@ Standalone Blazor WebAssembly:
 <script>
   Blazor.start({
     configureRuntime: dotnet => {
-      dotnet.withEnvironmentVariable("CONFIGURE_RUNTIME", "true");
+      dotnet.withEnvironmentVariable("MY_VARIABLE", "my value");
     }
   });
 </script>
--- a/aspnetcore/blazor/fundamentals/startup.md
+++ b/aspnetcore/blazor/fundamentals/startup.md
@@ -734,7 +734,7 @@ In `wwwroot/index.html`, remove the default SVG round indicator in `<div id="app
 
 To configure the .NET WebAssembly runtime, use the `configureRuntime` function with the `dotnet` runtime host builder.
 
-The following example sets an environment variable, `MY_VARIABLE`, to `my value`:
+In the following examples, the `{NAME}` placeholder is the environment variable's name, and the `{VALUE}` placeholder is the environment variable's value.
 
 Blazor Web App:
 
@@ -744,7 +744,7 @@ Blazor Web App:
   Blazor.start({
     webAssembly: {
       configureRuntime: dotnet => {
-        dotnet.withEnvironmentVariable("MY_VARIABLE", "my value");
+        dotnet.withEnvironmentVariable("{NAME}", "{VALUE}");
       }
     }
   });
@@ -758,7 +758,7 @@ Standalone Blazor WebAssembly:
 <script>
   Blazor.start({
     configureRuntime: dotnet => {
-      dotnet.withEnvironmentVariable("MY_VARIABLE", "my value");
+      dotnet.withEnvironmentVariable("{NAME}", "{VALUE}");
     }
   });
 </script>


guardrex (approved):
Thanks @hakenr! ... Let's avoid "my"-naming. I've been working for years to keep all the "my" things out of the Blazor docs 😆. Let's go with placeholders for these.

guardrex:
***WOE!*** We need to hold the horses 🐴🐴🐴 ... this isn't just a name ... it's a ***real env var*** for configuring the WASM runtime.

https://github.com/search?q=repo%3Adotnet%2Faspnetcore%20CONFIGURE_RUNTIME&type=code

I'll open an issue to follow-up with additional guidance on what it does.

hakenr:
@guardrex  I think you've been caught up in the confusion of that name, and even in the test project, it's used only as a test variable (the framework doesn't interact with it in any way):
https://github.com/dotnet/aspnetcore/blob/main/src/Components/test/testassets/BasicTestApp/ConfigureRuntime.razor#L5

😇

hakenr:
(It seems the code snippet in the docs was just copy-pasted from the TestApp and the "not so ideal name for test environment variable" initiated there.)

hakenr:
BTW, I like the placeholders-style for code examples, I just didn't think of it when I was proposing this PR. 😇

guardrex:
I put the placeholders together and try a NOTE on `Blazor.runtime` to set it aside from the main concept, including an example type for the build config.

guardrex:
I think the reason that the PU named it that way is that they're highlighting in the name that it's an env var specifically used to configure something about the runtime. However, I totally see what you mean and agree that someone ....*cough cough* 🦖 *cough* ....... will misunderstand that ***the dev*** would then go on to use it somehow in their own custom code. Another way to handle this is to explain it out. However, I don't think it's very important. This is, after all, an advanced scenario for advanced devs.

guardrex:
I think we better have Mackinnon look at this. In addition to any foul description 💩  that might be present, I'd like to be able to tell readers about any typical use cases, even if it's only mentioned in text and not with a code example. Before I do that, I'll go check the PU repo issues, including on the runtime repo, to see if the team told any devs *do this to solve such-and-such goal*.

guardrex:
This one is kind'a interesting ... idk where it logs ... console I suppose ...

```
dotnet.withEnvironmentVariable("MONO_LOG_LEVEL", "debug");
```

hakenr:
Hidden gems... 😁

https://github.com/dotnet/runtime/blob/339443b0408f45503380558aa2cf2a9ecc2cbaa9/src/mono/browser/test-main.js#L274-L277
```javascript
    // dotnet.withEnvironmentVariable("MONO_LOG_LEVEL", "debug")
    // dotnet.withEnvironmentVariable("MONO_LOG_MASK", "gc")
    // dotnet.withEnvironmentVariable("MONO_GC_DEBUG", "9")
    // dotnet.withEnvironmentVariable("DOTNET_DebugWriteToStdErr", "1")
```

https://github.com/search?q=dotnet.withEnvironmentVariable&type=code

hakenr:
It would be incredibly helpful to cross-link an article which describes all possible environment variables affecting runtime. These variables will vary between Mono and CoreCLR, but unfortunately, it seems like such a list doesn't exist yet.

guardrex:
Yes ... and that doesn't surprise me. Over the years, a few readers have asked them for documentation on design and configuration of the runtime. If they do have anything, it will be in markdown files on their repo. Searching for `MONO_`, which should turn up any such markdown content, one sees that they haven't published anything yet.

guardrex:
Ok ... I think that's good. I'll ping Mackinnon now for an 👁️.

hakenr:
> No list or documentation describing all of the runtime environment variables exists at this time.
Should we explicitly acknowledge this in the documentation? Personally, I'd lean towards just letting it be... 😇

guardrex:
Yeah, I guess that is upsetting the way I ***blatantly attacked them*** like that 😆.

The best thing for me to do is open an issue asking them if they'll create a simple markdown file with the environment vars. If they accept the issue, then I can cross-link the issue.

... ***and*** that's a lot better for heading off a stale remark. It's easier to spot that they've worked the issue via one of my long-range tracking issues that monitors such things.

I'll take care of this shortly. I've gotten myself ***buried*** ⛰️⛏️😅 in something else at the moment.

guardrex:
I opened https://github.com/dotnet/runtime/issues/98225 to ask about placing a list with descriptions in the `dotnet/runtime` docs or giving the content to me for publication.

guardrex:
@maraf @akoeplinger ... When you reach this for a look 👀, a basic question is if we want to carry this coverage here at all. AFAICT, this is of no general use to the typical LOB Blazor dev, and I wonder if it would be better to reduce it down to a single cross-link to a markdown file document in the `dotnet/runtime` repo, where it's probably more relevant for advanced devs working on advanced scenarios. Either I or someone else can create the markdown file (or add it to an existing file). 

Conversely, we can continue to keep this coverage here. However, it seems like it should have links to resources on the env vars that can be set for the runtime that make sense to tell readers about. That's why I asked on the `dotnet/runtime` issue if the env vars are documented. Thus far we have ...

* `MONO_*` env vars: https://github.com/mono/mono/blob/main/man/mono.1
* `DOTNET_*`/`COMPlus_*` env vars: https://github.com/dotnet/runtime/blob/f582c0b66b2b91b626ccce62ea0e37496446e79f/src/coreclr/inc/clrconfigvalues.h

... and I can add those cross-links to this if it's decided that this coverage should remain here and if you agree that the links should be added to the coverage.

guardrex:
*NOTE TO SELF* ... If we keep this and per @lambdageek's remark, add a bit on ...

> ... there are also other uses. for example if an app is built with a `NativeFileReference` in order to include some C library, the environment variable setting could be used to affect the behavior of that C library.

maraf:
> @maraf @akoeplinger ... When you reach this for a look 👀, a basic question is if we want to carry this coverage here at all. AFAICT, this is of no general use to the typical LOB Blazor dev, and I wonder if it would be better to reduce it down to a single cross-link to a markdown file document in the `dotnet/runtime` repo, where it's probably more relevant for advanced devs working on advanced scenarios. Either I or someone else can create the markdown file (or add it to an existing file).

We can remove the concrete sample for setting environment variable, but the general idea that this is the way how to configure runtime and how to use export runtime API (use API of `dotnet.d.ts`) is something I think should be here. Because how to get to those API is specific for Blazor.

The example with setting envvar was more about that you can set any env var and then read it in .NET. Maybe the name `CONFIGURE_RUNTIME` which I originally used in Blazor unit test is missleading

guardrex:
> We can remove the concrete sample for setting environment variable

> Maybe the name CONFIGURE_RUNTIME which I originally used in Blazor unit test is missleading.

Since we've generalized it with placeholders, I think we should go ahead with this PR.

I'll be back shortly to add @lambdageek's remark ...

> there are also other uses. for example if an app is built with a `NativeFileReference` in order to include some C library, the environment variable setting could be used to affect the behavior of that C library.

... and then if Mackinnon is happy 😄 we can proceed and merge this. I'll open a tracking item for myself over here to keep an 👁️ on that `dotnet/runtime` issue for a future update if action is taken on that issue.

akoeplinger:
> * `MONO_*` env vars: [mono/mono@`main`/man/mono.1](https://github.com/mono/mono/blob/main/man/mono.1?rgh-link-date=2024-02-12T15%3A48%3A17Z)

The `man/mono.1` is for the old mono executable that shipped standalone, unfortunately we don't have a better place which documents the environment variables, ideally they'd be in some markdown doc.

aspnetcore/blazor/fundamentals/startup.md
```diff
@@ -732,9 +732,18 @@ In `wwwroot/index.html`, remove the default SVG round indicator in `<div id="app
 
 ## Configure the .NET WebAssembly runtime
 
-To configure the .NET WebAssembly runtime, use the `configureRuntime` function with the `dotnet` runtime host builder.
+In advanced programming scenarios, the `configureRuntime` function with the `dotnet` runtime host builder is used to configure the .NET WebAssembly runtime. For example, `dotnet.withEnvironmentVariable` sets an environment variable that:
 
-The following example sets an environment variable, `CONFIGURE_RUNTIME`, to `true`:
+* Configures the .NET WebAssembly runtime.
+* Changes the behavior of a C library if the app is built with a [`NativeFileReference`](xref:blazor/webassembly-native-dependencies).
```
akoeplinger, line 738: this makes it sound like this is some special feature of NativeFileReference, but in reality it's just the C library reading an env var and then doing something :)

guardrex:
```suggestion
* Changes the behavior of a C library.
```

aspnetcore/blazor/fundamentals/startup.md
```diff
@@ -758,15 +767,14 @@ Standalone Blazor WebAssembly:
...
     }
   });
 </script>
 ```
 
-In the preceding example, the `{BLAZOR SCRIPT}` placeholder is the Blazor script path and file name. For the location of the script, see <xref:blazor/project-structure#location-of-the-blazor-script>.
-
-The .NET runtime instance can be accessed from `Blazor.runtime`.
+> [!NOTE]
+> The .NET runtime instance can be accessed using the Blazor WebAssembly Runtime API (`Blazor.runtime`). For example, the app's build configuration can be obtained using `Blazor.runtime.runtimeBuildInfo.buildConfiguration`.
```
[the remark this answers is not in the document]
guardrex: Sure ... how do you want to phrase the remark? ...

```suggestion
> The .NET runtime instance can be accessed using the Blazor WebAssembly Runtime API (`Blazor.runtime`). For example, the app's build configuration can be obtained using `Blazor.runtime.runtimeBuildInfo.buildConfiguration`.
>
> For more information on the .NET WebAssembly runtime configuration, see the [runtime's TypeScript definition file (`dotnet.d.ts`) in the `dotnet/runtime` GitHub repository](https://github.com/dotnet/runtime/blob/main/src/mono/browser/runtime/dotnet.d.ts).
>
> [!INCLUDE[](~/includes/aspnetcore-repo-ref-source-links.md)]
```

guardrex:
@MackinnonBuck ... I think we're good here. I'm going to go ahead with this. If you have a concern, I can issue a patch PR.

PR #31808, opened 2024-02-14
[Blazor] SignalR - remove "using System" reminder

hakenr:
It seems inappropriate and somewhat unnecessary to remind everyone to include `using System;` for every demo. The topic of SignalR, especially the demos that involve configuring specific options, is quite advanced, and it's reasonable to assume that the audience is already familiar with such details.

(By the way, the Blazor templates use `<ImplicitUsings>enable</ImplicitUsings>`.)

I've noticed this reminder in other topics, but this time, it really felt superfluous.

cc @guardrex 

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/signalr.md](https://github.com/dotnet/AspNetCore.Docs/blob/96a203748e942016e6369ff828d05ab2ecd79e84/aspnetcore/blazor/fundamentals/signalr.md) | [ASP.NET Core Blazor SignalR guidance](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/signalr?branch=pr-en-us-31808) |

### DIFF
--- a/aspnetcore/blazor/fundamentals/signalr.md
+++ b/aspnetcore/blazor/fundamentals/signalr.md
@@ -147,7 +147,7 @@ Configure the circuit with the <xref:Microsoft.AspNetCore.Components.Server.Circ
 
 :::moniker range=">= aspnetcore-8.0"
 
-Configure the options in the `Program` file with an options delegate to <xref:Microsoft.Extensions.DependencyInjection.ServerRazorComponentsBuilderExtensions.AddInteractiveServerComponents%2A>. The following example assigns the default option values shown in the preceding table. Confirm that the `Program` file uses the <xref:System> namespace (`using System;`).
+Configure the options in the `Program` file with an options delegate to <xref:Microsoft.Extensions.DependencyInjection.ServerRazorComponentsBuilderExtensions.AddInteractiveServerComponents%2A>. The following example assigns the default option values shown in the preceding table.
 
 In the `Program` file:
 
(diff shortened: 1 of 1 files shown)


guardrex (approved):
It's a holdover from the days when implicit usings weren't a feature (<6.0). I agree that we can delete these now, including in the <6.0 content because the early versions have fallen out of use for new development.

hakenr:
@guardrex I will try to clean up the other articles.

guardrex:
In the old days, we would get a fairly constant stream of issues from devs that couldn't find the assemblies.

Let's be careful about this. If it seems like it makes sense to drop a mention, let's leave the call out. `System` is a good example. The problem is that lots of devs don't use VS, so lots of devs get stuck and can't find the assembly because they don't to look at the API browser. I don't want to go back to the old days of having many issues pop up asking, "Where is this?!"

hakenr:
> Let's be careful about this. If it seems like it makes sense to drop a mention, let's leave the call out. System is a good example of one we can drop. The problem is that lots of devs don't use VS, so lots of devs get stuck and can't find the assembly because they don't to look at the API browser. I don't want to go back to the old days of having many issues pop up asking, "Where is this?!"

I couldn't find any other instructions that seemed as trivial for the `using` statement. ;-)

guardrex:
Ok ... that's cool. Thanks for checking.

PR #31866, opened 2024-02-20
Update component ref text and examples

### LINKED ISSUE #31860: [Blazor] Components overview - Capture references to components - misleading example with event handler delegate ReferenceParent2.razor
### Description

> A component reference is only populated after the component is rendered, and its output includes ReferenceChild's element. Until the component is rendered, there's nothing to reference.
> To manipulate component references after the component has finished rendering, use the OnAfterRender or OnAfterRenderAsync methods.

The preceding paragraphs explain that you can only use a component reference after the referred component is rendered, typically in the `OnAfterRender` method. I'm puzzled by the focus on event handlers in the next paragraph, suggesting two approaches:
* lambda expression
* event handler delegate

> To use a reference variable with an event handler, employ a lambda expression or assign the event handler delegate in the [OnAfterRender or OnAfterRenderAsync methods](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?view=aspnetcore-8.0#after-component-render-onafterrenderasync). This ensures the reference variable is assigned before the event handler is set up.

```razor
@page "/reference-parent-1"

<PageTitle>Reference Parent 1</PageTitle>

<h1>Reference Parent Example 1</h1>

<button @onclick="@(() => childComponent?.ChildMethod(5))">
    Call <code>ReferenceChild.ChildMethod</code> with an argument of 5
</button>

<ReferenceChild @ref="childComponent" />

@code {
    private ReferenceChild? childComponent;
}
```

The discussion here revolves not around event handlers and their syntax, but the general safety of using component references. Guarding the usage of a component reference with a null-conditional operator (`childComponent?.ChildMethod()`) is one approach and is not limited to lambda expressions.

The `ReferenceParent2.razor` example doesn't add value by using the `callChildMethod` event handler delegate, as the only code in `CallChildMethod()` is also guarded with a null-conditional operator.

```razor
@page "/reference-parent-2"

<PageTitle>Reference Parent 2</PageTitle>

<h1>Reference Parent Example 2</h1>

<button @onclick="@(() => callChildMethod?.Invoke())">
    Call <code>ReferenceChild.ChildMethod</code> with an argument of 5
</button>

<ReferenceChild @ref="childComponent" />

@code {
    private ReferenceChild? childComponent;
    private Action? callChildMethod;

    protected override void OnAfterRender(bool firstRender)
    {
        if (firstRender)
        {
            callChildMethod = CallChildMethod;
        }
    }

    private void CallChildMethod()
    {
        childComponent?.ChildMethod(5);
    }
}
```

The original intention might have been to present an alternative to the null-conditional operator guard, but I still argue that enabling/disabling event handlers by using a delegate assigned in `OnAfterRender` complicates readability and maintainability, even if it technically works.

To demonstrate an alternative approach to guarding the component reference with a null-conditional operator, consider using a `bool isReady` flag. This can be checked whenever you need to verify the component reference is usable. Note: The `OnAfterRender` method doesn't guarantee the component reference will be set and ready, e.g., if the referred component is conditionally rendered.

```razor
@page "/reference-parent-2"

<PageTitle>Reference Parent 2</PageTitle>

<h1>Reference Parent Example 2</h1>

<button @onclick="CallChildMethod">
    Call <code>ReferenceChild.ChildMethod</code> with an argument of 5
</button>

<ReferenceChild @ref="childComponent" />

@code {
    private ReferenceChild? childComponent;
    private bool isReady;

    protected override void OnAfterRender(bool firstRender)
    {
            isReady = true;
    }

    private void CallChildMethod()
    {
        if (!isReady)
        {
           return;
        }
        childComponent?.ChildMethod(5);
    }
}
```

My proposal:
1. Remove the focus on "event handlers" from the explanation.
2. Remove the whole `ReferenceParent2.razor` example or replace it with code using an `isReady` (or better named) flag if we want to show alternative approach to null-checking of component reference.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/index.md

### Document ID

31ad18df-1904-bbdd-cfed-8d0d0267c10e

### Article author

@guardrex

guardrex:
Fixes #31860

@hakenr ... This is my interpretation of Steve's guidance. He didn't remark on [Artak's discussion with @mrlife](https://github.com/dotnet/aspnetcore/issues/22658), so we'll let that go.

Note tho that for ...

> \
> `<button type="button" @onclick="myComponent.DoSomething">Do Something</button>`

... I'm leaning toward remarking on that in passing here. Yeah ... I like that idea. I think I'll commit a remark. ***UPDATE***: Draft remark is done 👍. Let's look here on review and see if the phrasing is correct and clear. **OK** ... I think I'm done with it for now. I had to play with the wording a bit more.

The examples (8.0 shown here) are now ...

`ReferenceParent1` component:

```razor
﻿@page "/reference-parent-1"

<PageTitle>Reference Parent 1</PageTitle>

<h1>Reference Parent Example 1</h1>

<button @onclick="@(() => childComponent!.ChildMethod(5))">
    Call <code>ReferenceChild.ChildMethod</code> with an argument of 5
</button>

<ReferenceChild @ref="childComponent" />

@code {
    private ReferenceChild? childComponent;
}
```

`ReferenceParent2` component:

```razor
﻿@page "/reference-parent-2"

<PageTitle>Reference Parent 2</PageTitle>

<h1>Reference Parent Example 2</h1>

<button @onclick="CallChildMethod">
    Call <code>ReferenceChild.ChildMethod</code> with an argument of 5
</button>

<ReferenceChild @ref="childComponent" />

@code {
    private ReferenceChild? childComponent;

    private void CallChildMethod()
    {
        childComponent!.ChildMethod(5);
    }
}
```

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/249267c2f4b28fcdfe90449263f8f0339fc77015/aspnetcore/blazor/components/index.md) | [ASP.NET Core Razor components](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/index?branch=pr-en-us-31866) |

### DIFF
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -1302,7 +1302,7 @@ A component reference is only populated after the component is rendered and its 
 
 To manipulate component references after the component has finished rendering, use the [`OnAfterRender` or `OnAfterRenderAsync` methods](xref:blazor/components/lifecycle#after-component-render-onafterrenderasync).
 
-The following lambda approach uses the preceding `ReferenceChild` component. The [`!` (null-forgiving) operator](/dotnet/csharp/language-reference/operators/null-forgiving) results in a clear error if `childComponent` isn't populated when `ChildMethod` is called, but `childComponent` is populated in all normal Blazor framework rendering scenarios. Don't attempt to call a referenced component method without a lambda expression (for example, `@onclick="childComponent!.ChildMethod(5)"`) because the reference variable may not be assigned by the time click event is assigned.
+The following lambda approach uses the preceding `ReferenceChild` component. The [`!` (null-forgiving) operator](/dotnet/csharp/language-reference/operators/null-forgiving) results in a clear error if `childComponent` isn't populated when `ChildMethod` is called, but `childComponent` is populated in all normal Blazor framework rendering scenarios. Don't attempt to call a referenced component method without a lambda expression (for example, `@onclick="childComponent!.ChildMethod(5)"`) because the reference variable may not be assigned at the time click event is assigned.
 
 `ReferenceParent1.razor`:
 
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -1302,7 +1302,7 @@ A component reference is only populated after the component is rendered and its 
 
 To manipulate component references after the component has finished rendering, use the [`OnAfterRender` or `OnAfterRenderAsync` methods](xref:blazor/components/lifecycle#after-component-render-onafterrenderasync).
 
-The following lambda approach uses the preceding `ReferenceChild` component. The [`!` (null-forgiving) operator](/dotnet/csharp/language-reference/operators/null-forgiving) results in a clear error if `childComponent` isn't populated when `ChildMethod` is called, but `childComponent` is populated in all normal Blazor framework rendering scenarios.
+The following lambda approach uses the preceding `ReferenceChild` component. The [`!` (null-forgiving) operator](/dotnet/csharp/language-reference/operators/null-forgiving) results in a clear error if `childComponent` isn't populated when `ChildMethod` is called, but `childComponent` is populated in all normal Blazor framework rendering scenarios. Don't attempt to call a a reference component's method without a lambda expression (for example, `@onclick="childComponent!.ChildMethod(5)"`) because the reference variable may not be assigned by the time click event is assigned.
 
 `ReferenceParent1.razor`:
 
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -1298,11 +1298,11 @@ Consider the following `ReferenceChild` component that logs a message when its `
 
 :::moniker-end
 
-A component reference is only populated after the component is rendered and its output includes `ReferenceChild`'s element. Until the component is rendered, there's nothing to reference.
+A component reference is only populated after the component is rendered and its output includes `ReferenceChild`'s element. Until the component is rendered, there's nothing to reference. Don't attempt to call a referenced component method to an event handler directly (for example, `@onclick="childComponent!.ChildMethod(5)"`) because the reference variable may not be assigned at the time click event is assigned.
 
-To manipulate component references after the component has finished rendering, use the [`OnAfterRender` or `OnAfterRenderAsync` methods](xref:blazor/components/lifecycle#after-component-render-onafterrenderasync).
+To manipulate component references after the component has finished rendering, use the [`OnAfterRender` or `OnAfterRenderAsync` methods](xref:blazor/components/lifecycle#after-component-render-onafterrenderasync). 
 
-The following lambda approach uses the preceding `ReferenceChild` component. The [`!` (null-forgiving) operator](/dotnet/csharp/language-reference/operators/null-forgiving) results in a clear error if `childComponent` isn't populated when `ChildMethod` is called, but `childComponent` is populated in all normal Blazor framework rendering scenarios. Don't attempt to call a referenced component method without a lambda expression (for example, `@onclick="childComponent!.ChildMethod(5)"`) because the reference variable may not be assigned at the time click event is assigned.
+The following examples use the preceding `ReferenceChild` component. 
 
 `ReferenceParent1.razor`:
 
@@ -1335,8 +1335,6 @@ The following lambda approach uses the preceding `ReferenceChild` component. The
 :::code language="razor" source="~/../blazor-samples/3.1/BlazorSample_WebAssembly/Pages/index/ReferenceParent1.razor":::
 
 :::moniker-end
-
-The following approach uses the preceding `ReferenceChild` component.
 
 `ReferenceParent2.razor`:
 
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/components/index.md
```diff
@@ -1302,9 +1302,7 @@ A component reference is only populated after the component is rendered and its
...
 
-To use a reference variable with an event handler, use a lambda expression or assign the event handler delegate in the [`OnAfterRender` or `OnAfterRenderAsync` methods](xref:blazor/components/lifecycle#after-component-render-onafterrenderasync). This ensures that the reference variable is assigned before the event handler is assigned.
-
-The following lambda approach uses the preceding `ReferenceChild` component.
+The following lambda approach uses the preceding `ReferenceChild` component. The [`!` (null-forgiving) operator](/dotnet/csharp/language-reference/operators/null-forgiving) results in a clear error if `childComponent` isn't populated when `ChildMethod` is called, but `childComponent` is populated in all normal Blazor framework rendering scenarios. Don't attempt to call a referenced component method without a lambda expression (for example, `@onclick="childComponent!.ChildMethod(5)"`) because the reference variable may not be assigned at the time click event is assigned.
```
hakenr, line 1305: The null-forgiving operator (`!') has no effect at run time.
Let me suggest a change, will be back in a minute.

guardrex: Well ... it shouldn't have any affect anyway according to Steve.

If we change this, then I'll need to ping him because we'll be moving away from his sentiment AFAICT. I'll wait to see your suggestion.

hakenr: Sorry, the tooling didn't allow me to suggest code changes to this PR, so I submitted a new PR: #31868 😇

guardrex: Use quote markdown in that case ... but I don't understand why GH is being such a pain this morning 😄.

hakenr:
@guardrex 

image

ORIGINAL:
>A component reference is only populated after the component is rendered and its output includes `ReferenceChild`'s element. Until the component is rendered, there's nothing to reference.
>
>To manipulate component references after the component has finished rendering, use the [`OnAfterRender` or `OnAfterRenderAsync` methods](xref:blazor/components/lifecycle#after-component-render-onafterrenderasync).
>
>To use a reference variable with an event handler, use a lambda expression or assign the event handler delegate in the [`OnAfterRender` or `OnAfterRenderAsync` methods](xref:blazor/components/lifecycle#after-component-render-onafterrenderasync). This ensures that the reference variable is assigned before the event handler is assigned.>
>
>The following lambda approach uses the preceding `ReferenceChild` component.

MY SUGGESTION:
>A component reference is only populated after the component is rendered and its output includes `ReferenceChild`'s element. Until the component is rendered, there's nothing to reference, for example don't attempt to assign a referenced component method to an event handler directly (`@onclick="childComponent.ChildMethod"`) because the reference variable may not be assigned at the time click event is assigned.
>
>The following lambda approach uses the preceding `ReferenceChild` component and postpones accessing the component reference to the time of click event:

ORIGINAL 2:
>The following delegate approach uses the preceding `ReferenceChild` component.

MY SUGGESTION 2:
> The following event handling method approach uses the preceding `ReferenceChild` component and also postpones accessing the component reference to the time of click event:

You will probably have to polish the English language, but I hope you get my intention.
- express what to avoid
- give examples how to use the captured component reference correctly
- [OPTIONAL] remove the `!' (the null-forgiving operator has no effect at runtime)

guardrex:
I have a feeling that the reason GH wouldn't allow suggestions is that you tried to suggest on an earlier commit.

guardrex:
... and BTW ... I kind'a like what I have here and would prefer not to ping Steve unless absolutely necessary.

hakenr:
I think the DIFF (and my "ORIGINAL") shows the current `main` branch, not just my commit vs. your PR.
Just my commit (https://github.com/dotnet/AspNetCore.Docs/pull/31868/commits/25c22a55010b7979f479467cee3b4348a62db33a) is this:
image

We can keep the `!` in sample code, but the

>The [`!` (null-forgiving) operator](/dotnet/csharp/language-reference/operators/null-forgiving) results in a clear error if `childComponent` isn't populated when `ChildMethod` is called

part is IMHO wrong. It just suppresses the nullable reference types warnings and has no effect in run time. There is no chance it will have any effect "when `ChildMethod` is called".

My suggestion after re-reading 😇:
> A component reference is only populated after the component is rendered and its output includes `ReferenceChild`'s element. Until the component is rendered, there's nothing to reference. Do not attempt to assign a referenced component method to an event handler directly (`@onclick="childComponent.ChildMethod"`) because the reference variable may not be assigned at the time click event is assigned.
>
>The following lambda approach uses the preceding `ReferenceChild` component and accesses the component reference at the time of click event: 
> ...
> The following event handling method approach uses the preceding `ReferenceChild` component and also accesses the component reference at the time of click event:

guardrex:
Right ... gotcha. Thanks for clarifying that. I misinterpreted Steve's remark ...

> so that you get a clear error if childComponent is somehow not yet populated

guardrex:
Perhaps, we don't need to get into the lambda versus method distinction. The examples seem self-exclamatory.

guardrex:
I feel both of these could be rolled into one component. I might make that update. 🤔

PR #31872, opened 2024-02-21
Update DispatchExceptionAsync example

### LINKED ISSUE #31765: [Blazor] Handle Errors - misleading examples on DispatchExceptionAsync (net8)
@guardrex I believe the `TimerService.cs` + `Notifications.razor` example on `DispatchExceptionAsync` in [Handle errors](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/handle-errors?view=aspnetcore-8.0) - [Handle caught exceptions outside of a Razor component's lifecycle](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/handle-errors?view=aspnetcore-8.0#handle-caught-exceptions-outside-of-a-razor-components-lifecycle) section is wrong.

Following the article's guidance (and the readily available example in the [blazor-samples repo](https://github.com/dotnet/blazor-samples/blob/main/8.0/BlazorSample_BlazorWebApp/Components/Pages/Notifications.razor) and easily applicable there), the exception added to `TimerService.cs` is reported to the user even without adding the `try-catch` and `DispatchExceptionAsync(ex)` to `OnNotify` method.
  * This happens due to the `await Timer.Start()` call in `StartTimer()` method, where the exception from the timer loop gets propagated to Blazor
    ```csharp
    private async Task StartTimer()
    {
        await Timer.Start();
    }
    ```
  * Changing this method to a fire-and-forget approach causes exceptions not to be reported to the user
    ```csharp
    private void StartTimer()
    {
        _ = Timer.Start();
    }
    ```
Unfortunately, adding a `try-catch` block to the `OnNotify()` method as suggested does not effectively capture or handle exceptions, since the try block does not encompass the potential source of the exception (either directly or by capturing exceptions from the task where they occur). Incorporating this try-catch into the modified example above does not route exceptions to the previously added `ErrorBoundary` as expected.

I can work on creating a functional example that incorporates `DispatchExceptionAsync(ex);` based on the `TimerService.cs` and `Notifications.razor` if that would help. However, I'd appreciate it if you could first validate the concerns I've outlined above.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/handle-errors?view=aspnetcore-8.0#handle-caught-exceptions-outside-of-a-razor-components-lifecycle

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/fundamentals/handle-errors.md

### Document ID

66a0b1c2-c45c-98ca-9808-6f340a861c44

### Article author

@guardrex

guardrex:
Fixes  #31765

@hakenr ... I'd like to try and fix what's here. It seems faster to just adopt Steve's pattern ... ***the correct approach*** 🙈 ... by merely ...

* 🔪 out the `OnNotify` piece.
* Call `DispatchExceptionAsync` for the timer in the component in the way that Steve proposed calling it.

This only requires a few small changes to what's here, which makes this a quick 🏃‍♂️ update.

WRT a component example for this in the sample apps, I placed a tracking item for it on my long-range tracking issue. I also cross-linked the example that you provided. It might be a good idea to include your example. The text could receive a cross-link and quick description of the alternative approach. I'll consider it further when I reach that item on my long-range tracking issue.

Sorry about the churn pings. I found a number of NITs 😈 to change after I pinged this time.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/handle-errors.md](https://github.com/dotnet/AspNetCore.Docs/blob/8164562d336c72a5151d50e776461bb83510f313/aspnetcore/blazor/fundamentals/handle-errors.md) | [Handle errors in ASP.NET Core Blazor apps](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/handle-errors?branch=pr-en-us-31872) |

### DIFF
--- a/aspnetcore/blazor/fundamentals/handle-errors.md
+++ b/aspnetcore/blazor/fundamentals/handle-errors.md
@@ -233,8 +233,8 @@ To dispatch exceptions from the timer service back to the `Notifications` compon
 * Instead of calling `Timer.Start` from `StartTimer`, the asynchronous method is called and the `Task` is intentionally discarded, which is often called *fire and forget* because the method is *fired* (started) and the result of the method, the `Task`, is *forgotten* (thrown away). The `StartTimer` method's signature changes to adopt the fire and forget pattern:
 
   ```diff
-  - Task StartTimer()
-  + void StartTimer()
+  - Task StartTimer();
+  + void StartTimer();
   ```
 
 In the `@code` block of the `Notifications` component (`Notifications.razor`):
--- a/aspnetcore/blazor/fundamentals/handle-errors.md
+++ b/aspnetcore/blazor/fundamentals/handle-errors.md
@@ -179,7 +179,7 @@ To treat failures like lifecycle method exceptions, explicitly dispatch exceptio
 }
 ```
 
-An alternative approach for sending the report and discarding the <xref:System.Threading.Tasks.Task> is to leverage <xref:System.Threading.Tasks.Task.Run%2A?displayProperty=nameWithType>:
+An alternative approach for sending the report and discarding the <xref:System.Threading.Tasks.Task> is to leverage <xref:System.Threading.Tasks.Task.Run%2A?displayProperty=nameWithType>. This pattern is often called *fire and forget* because the method (<xref:System.Threading.Tasks.Task.Run%2A?displayProperty=nameWithType>) is *fired* (started) and the result of the method (the <xref:System.Threading.Tasks.Task> returned by the <xref:System.Action> delegate) is *forgotten* (thrown away).
 
 ```csharp
 private void SendReport()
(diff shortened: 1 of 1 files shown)


hakenr:
@guardrex Just let me know when you've finished with the updates if you'd like me to check the final result.

guardrex:
I'm done ... ***I think*** 😆.

guardrex:
***Yes!*** I like the `Task.Run` approach ... less code is always better 👍.

So ...... I took the following approach ...

* Leave Steve's example there for the report sender. Add the `Task.Run` version of that.
* Use the `Task.Run` version for the example demo.

Yes on leaving the example demo here. When I reach my long-range tracking issue, I'll turn it into a sample app example and remove it here saying, 'for an example, see the sample app's X component' with a cross-link.

guardrex:
Ok ... I'm good. I'm a 🦖 ***PR churn monster*** 🦖 today on this one 🙈.

You can look now.

hakenr (approved):
👍 lgtm

Issue #31205, opened 2023-12-05
BlazorWebAssemblyStandaloneWithIdentity uses cookie auth but no antiforgerytoken


nathan-parkinson:
### Description

The BlazorWebAssemblyStandaloneWithIdentity sample uses a cookie to store the authentication token but I believe that this opens the site up to csrf attacks.

This is usually mitigated by using antiforgerytokens but I've found it difficult to apply this to the sample as the api and client run as different sites (although I may have just dome something wrong).

Is is right that antiforgery tokens should be used here or is there a reason it is not required?
If it is required will it be added to this sample or can someone point me in the right direction to get this working?

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/standalone-with-identity?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/security/webassembly/standalone-with-identity.md

### Document ID

c4e6ec41-7bea-e600-6473-c5c870aab082

### Article author

@guardrex

guardrex:
Thanks for opening the issue this way. We use the metadata to track work on these.

It should be automatic and transparent. If you look at a request in browser tools, you should see the antiforgery cookie named something like `.AspNetCore.Antiforgery.XXXXXXX` on each request.

I agree that it would be nice if the article addressed this subject to head off concerns about it.

@JeremyLikness ... Plz confirm for us that devs don't need to sweat XSRF with basic use of this solution, and are there any ***gotchas*** 😈 that are worth calling out?

nathan-parkinson:
I have run the `Backend` and `BlazorWasmAuth` projects, registered and logged in without the Antiforgery cookie being created.

There is no cookie in the dev tools called `.AspNetCore.Antiforgery.XXXXXXX`. This is, in fact, what I have been trying to implement. 

Am I missing something?

guardrex:
Let me see. I have that test app here ... I'll run it and see what happens. I thought I recalled seeing it, but I might be mistaken. *Stand-by ..........................*

nathan-parkinson:
Here is mine. I still see no antiforgery token.
Feel like I'm missing something obvious

![image](https://github.com/dotnet/AspNetCore.Docs/assets/18175816/7dba68fa-d394-4cfe-9b14-16eff2bc219c)

guardrex:
Let's clarify something ...

[...]

PR #31888, opened 2024-02-22
WASM+Identity same-site & antiforgery updates

### LINKED ISSUE #31389: Blazor WebAssembly With Identity same-site coverage
### Description

After going through [the BlazorWebAssemblyStandaloneWithIdentity example](https://github.com/dotnet/blazor-samples/tree/main/8.0/BlazorWebAssemblyStandaloneWithIdentity), which works great, I want to make sure ASP.NET Core Identity with Cookie Authentication is still the right choice after deploying this to Azure. 

Specifically, I have deployed my apps, following the approach in this example, like this:

* Blazor WASM standalone app hosted in Azure Static Web App (https://brave-field-0f649a310.4.azurestaticapps.net)
* ASP.NET Core API hosted in Azure App Service (https://gamestore01.azurewebsites.net) 

The apps work mostly without issues in Azure, but I did have to add this to my API for cookies to work properly:

```csharp
builder.Services.ConfigureApplicationCookie(options =>
{
    options.Cookie.SameSite = SameSiteMode.None;
    options.Cookie.SecurePolicy = CookieSecurePolicy.Always;
});
```

I want to make sure ASP.NET Core Identity is the recommended approach in this scenario, as opposed to using an OIDC server. I went through the guidance on [this official doc](https://learn.microsoft.com/aspnet/core/security/how-to-choose-identity-solution) and after following this diagram:

image

...it is not clear to me if my Blazor WASM standalone app, which lives in Azure Static Web App, should be classified as external or internal, which is the first decision point. I own both the API and the Blazor app, but it's not like they live in the same domain (given the different URLs provided by App Service and Static App).

Could you please clarify if the approach in the BlazorWebAssemblyStandaloneWithIdentity sample is recommended when deploying to Azure App Service + Azure Static Web App?

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/standalone-with-identity?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/security/webassembly/standalone-with-identity.md

### Document ID

c4e6ec41-7bea-e600-6473-c5c870aab082

### Article author

@guardrex

guardrex:
Fixes #31389
Addresses #31205
Addresses #28161

## Sample updates

FIRST ... before reviewing the article DIFF on this PR ... let's sort out the sample updates at ...

https://github.com/dotnet/blazor-samples/pull/210

* NIT updates on the PR
  * Move CORS in the pipeline above the logout endpoint.
  * Lowercase the logout endpoint from `/Logout` to `/logout`.
  * Change conditional check `empty != null` to `empty is not null`.
  * Place a line of spacing above `return`s ... that's by convention for Blazor example code.
  * Remove null-forgiving ops from `email` and `password`, as they can't be null.
  * Remove the `UserBasic` class, as it isn't used.
 * Same-site/secure policy settings: Add commented-out default-setting code for it.
 * Add a quick-'n-dirty data processing experience.
   * Form in a page that uses the `Auth` client to POST to the `Backend` app.
   * Minimal API endpoint that processes the data and returns a string.
   * ~NOTE: The same errors shown in https://github.com/dotnet/blazor-samples/issues/161 are seen here when that code processes an unauthenticated request to the `/data-processing` endpoint. I'll place the error at the bottom of this OP.~ We think now that this is resolved by the sample updates.

### Security Overview

This section applies to all Blazor app hosting models/templates. The primary goal here is to add a reminder that ...

* ~Antiforgery mitigation is for form submissions encoded as `application/x-www-form-urlencoded`, `multipart/form-data`, or `text/plain` with a cross-link to the Blazor Forms article coverage on forms antiforgery support.~ ***Updated on review.***
* ~Server API endpoints with `application/json`-encoded content and CORS don't require CSRF protection.~ ***Updated on review.***

### Standalone with Identity article

* Adds a new section on *Cross-domain hosting (same-site configuration)*.
* Adds a new section on antiforgery support.
  * Explain the situation with the logout endpoint.
  * Add the reminders on forms antiforgery and server API/JSON/CORS.

I'm still a bit concerned ... even if this is all correct (or close to correct) ... that the bit about forms submission in a standalone WASM situation could use more work. Using an HTTP client with a JSON request and CORS is what we focus on throughout the docs. This will be the first time that we explicitly say that antiforgery isn't a concern in that scenario. However, `EditForm` is useful client-side ***to just collect the data*** for a JSON POST to a server API. I don't want readers to think that anything is happening for `EditForm` in such an app vis-a-vis antiforgery. I might need to add more content on this point.

... and BTW Jeremy ... it seems to me that one day if I ever get the time that I should convert your plain forms over to `EditForm` forms with the nice validation ... and roll in the bits to trigger validation failures with the `ValidationMessage` component. I can do all that ... it's just a question of ***TIME*** ⏲️ ... and I won't have any for the foreseeable future 🏃‍♂️😄.

## Error

***This has been addressed by the updates to the sample app.***

~Error for unauthenticated hit on the `/data-processing` endpoint. What the sample is doing now is trapping it in a `try-catch` and showing a message. ***Can't we work out something more graceful than that?*** ... or does this all resolve back to your remarks here ...~

~https://github.com/dotnet/blazor-samples/issues/161#issuecomment-1887754626~

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/security/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/718c5e767f17850b39268243a61059eb477b3eb7/aspnetcore/blazor/security/index.md) | [ASP.NET Core Blazor authentication and authorization](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/index?branch=pr-en-us-31888) |
| [aspnetcore/blazor/security/webassembly/standalone-with-identity.md](https://github.com/dotnet/AspNetCore.Docs/blob/718c5e767f17850b39268243a61059eb477b3eb7/aspnetcore/blazor/security/webassembly/standalone-with-identity.md) | [Secure ASP.NET Core Blazor WebAssembly with ASP.NET Core Identity](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/standalone-with-identity?branch=pr-en-us-31888) |

### DIFF
--- a/aspnetcore/blazor/security/index.md
+++ b/aspnetcore/blazor/security/index.md
@@ -50,7 +50,7 @@ ASP.NET Core abstractions, such as <xref:Microsoft.AspNetCore.Identity.SignInMan
 
 ## Antiforgery support
 
-Blazor adds Antiforgery Middleware and requires endpoint [antiforgery protection](xref:security/anti-request-forgery) by default to mitigate the threats of Cross-Site Request Forgery (CSRF/XSRF).
+The Blazor template adds Antiforgery Middleware and requires endpoint [antiforgery protection](xref:security/anti-request-forgery) by default to mitigate the threats of Cross-Site Request Forgery (CSRF/XSRF).
 
 The <xref:Microsoft.AspNetCore.Components.Forms.AntiforgeryToken> component renders an antiforgery token as a hidden field, and this component is automatically added to form (<xref:Microsoft.AspNetCore.Components.Forms.EditForm>) instances. For more information, see <xref:blazor/forms/index#antiforgery-support>.
 
@@ -59,9 +59,7 @@ The <xref:Microsoft.AspNetCore.Components.Forms.AntiforgeryStateProvider> servic
 Blazor stores request tokens in component state, which guarantees that antiforgery tokens are available to interactive components, even when they don't have access to the request.
 
 > [!NOTE]
-> [Antiforgery mitigation](xref:security/anti-request-forgery) is only required when submitting form data to the server encoded as `application/x-www-form-urlencoded`, `multipart/form-data`, or `text/plain`. For more information, see <xref:blazor/forms/index#antiforgery-support>.
->
-> Requests to server API endpoints (web API) with `application/json`-encoded content and [CORS](xref:security/cors) enabled doesn't require CSRF protection.
+> [Antiforgery mitigation](xref:security/anti-request-forgery) is only required when submitting form data to the server encoded as `application/x-www-form-urlencoded`, `multipart/form-data`, or `text/plain` since these are the [only valid form enctypes](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fs-enctype). For more information, see <xref:blazor/forms/index#antiforgery-support>.
 
 :::moniker-end
 
--- a/aspnetcore/blazor/security/index.md
+++ b/aspnetcore/blazor/security/index.md
@@ -50,13 +50,18 @@ ASP.NET Core abstractions, such as <xref:Microsoft.AspNetCore.Identity.SignInMan
 
 ## Antiforgery support
 
-Blazor adds Antiforgery Middleware and requires endpoint [antiforgery protection](xref:security/anti-request-forgery) by default. 
+Blazor adds Antiforgery Middleware and requires endpoint [antiforgery protection](xref:security/anti-request-forgery) by default to mitigate the threats of Cross-Site Request Forgery (CSRF/XSRF).
 
 The <xref:Microsoft.AspNetCore.Components.Forms.AntiforgeryToken> component renders an antiforgery token as a hidden field, and this component is automatically added to form (<xref:Microsoft.AspNetCore.Components.Forms.EditForm>) instances. For more information, see <xref:blazor/forms/index#antiforgery-support>.
 
 The <xref:Microsoft.AspNetCore.Components.Forms.AntiforgeryStateProvider> service provides access to an antiforgery token associated with the current session. Inject the service and call its <xref:Microsoft.AspNetCore.Components.Forms.AntiforgeryStateProvider.GetAntiforgeryToken> method to obtain the current <xref:Microsoft.AspNetCore.Components.Forms.AntiforgeryRequestToken>. For more information, see <xref:blazor/call-web-api#antiforgery-support>.
 
 Blazor stores request tokens in component state, which guarantees that antiforgery tokens are available to interactive components, even when they don't have access to the request.
+
+> [!NOTE]
+> [Antiforgery mitigation](xref:security/anti-request-forgery) is only required when submitting form data to the server encoded as `application/x-www-form-urlencoded`, `multipart/form-data`, or `text/plain`. For more information, see <xref:blazor/forms/index#antiforgery-support>.
+>
+> Requests to server API endpoints (web API) with `application/json`-encoded content and [CORS](xref:security/cors) enabled doesn't require CSRF protection.
 
 :::moniker-end
 
--- a/aspnetcore/blazor/security/webassembly/standalone-with-identity.md
+++ b/aspnetcore/blazor/security/webassembly/standalone-with-identity.md
@@ -265,7 +265,7 @@ For more information on same-site cookie settings, see the following resources:
 
 Only the logout endpoint (`/logout`) in the `Backend` app requires attention to mitigate the threat of [Cross-Site Request Forgery (CSRF)](xref:security/anti-request-forgery).
 
-The logout endpoint checks for an empty body to prevent CSRF attacks. By requiring a body, the request must be made from JavaScript, which is the only way to access the authentication cookie. The cookie can't be accessed by a form-based POST. This prevents a malicious site from logging the user out.
+The logout endpoint checks for an empty body to prevent CSRF attacks. By requiring a body, the request must be made from JavaScript, which is the only way to access the authentication cookie. The logout endpoint can't be accessed by a form-based POST. This prevents a malicious site from logging the user out.
 
 Furthermore, the endpoint is protected by authorization (<xref:Microsoft.AspNetCore.Builder.AuthorizationEndpointConventionBuilderExtensions.RequireAuthorization%2A>) to prevent anonymous access.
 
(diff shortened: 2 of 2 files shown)


aspnetcore/blazor/security/index.md
```diff
@@ -50,14 +50,19 @@ ASP.NET Core abstractions, such as <xref:Microsoft.AspNetCore.Identity.SignInMan
...
 
 The <xref:Microsoft.AspNetCore.Components.Forms.AntiforgeryStateProvider> service provides access to an antiforgery token associated with the current session. Inject the service and call its <xref:Microsoft.AspNetCore.Components.Forms.AntiforgeryStateProvider.GetAntiforgeryToken> method to obtain the current <xref:Microsoft.AspNetCore.Components.Forms.AntiforgeryRequestToken>. For more information, see <xref:blazor/call-web-api#antiforgery-support>.
 
 Blazor stores request tokens in component state, which guarantees that antiforgery tokens are available to interactive components, even when they don't have access to the request.
 
+> [!NOTE]
+> [Antiforgery mitigation](xref:security/anti-request-forgery) is only required when submitting form data to the server encoded as `application/x-www-form-urlencoded`, `multipart/form-data`, or `text/plain`. For more information, see <xref:blazor/forms/index#antiforgery-support>.
```
halter73, line 62:
```suggestion
> [Antiforgery mitigation](xref:security/anti-request-forgery) is only required when submitting form data to the server encoded as `application/x-www-form-urlencoded`, `multipart/form-data`, or `text/plain` since these are the [only valid form enctypes](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fs-enctype). For more information, see <xref:blazor/forms/index#antiforgery-support>.
```

@blowdart @GrabYourPitchforks Do either of you think we should hedge this statement in some way?

aspnetcore/blazor/security/index.md
```diff
@@ -50,14 +50,19 @@ ASP.NET Core abstractions, such as <xref:Microsoft.AspNetCore.Identity.SignInMan
...
+>
+> Requests to server API endpoints (web API) with `application/json`-encoded content and [CORS](xref:security/cors) enabled doesn't require CSRF protection.
```
halter73, line 64:
```suggestion
```

Enabling CORS is exactly what could land you into trouble especially if you call `AllowAnyOrigin()`. Fortunately, the linked doc warns against doing that, but it's wrong to claim you need to "enable" CORS. The safest option is to not enable CORS at all unless you need it, and most apps don't.

aspnetcore/blazor/security/webassembly/standalone-with-identity.md
```diff
@@ -237,6 +237,44 @@ In the `Backend` server API's `Program` file, a [Minimal API](xref:fundamentals/
...
+For more information on same-site cookie settings, see the following resources:
+
+* [Set-Cookie: `SameSite=<samesite-value>` (MDN documentation)](https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value)
+* [Cookies: HTTP State Management Mechanism (RFC Draft 6265, Section 4.1)](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-03#section-4.1)
+
+## Antiforgery support
+
+Only the logout endpoint (`/logout`) in the `Backend` app requires attention to mitigate the threat of [Cross-Site Request Forgery (CSRF)](xref:security/anti-request-forgery).
+
+The logout endpoint checks for an empty body to prevent CSRF attacks. By requiring a body, the request must be made from JavaScript, which is the only way to access the authentication cookie. The cookie can't be accessed by a form-based POST. This prevents a malicious site from logging the user out.
```
halter73, line 268:
```suggestion
The logout endpoint checks for an empty body to prevent CSRF attacks. By requiring a body, the request must be made from JavaScript, which is the only way to access the authentication cookie. The logout endpoint can't be accessed by a form-based POST. This prevents a malicious site from logging the user out.
```

While a cross-domain form-based POST doesn't exactly "access" the cookie, it will definitely include it as part of the request with `SameSiteMode.None`. The key is that the `/logout` endpoint will not be triggered by a form-based POST even though it will include the cookie.

aspnetcore/blazor/security/webassembly/standalone-with-identity.md
```diff
@@ -237,6 +237,44 @@ In the `Backend` server API's `Program` file, a [Minimal API](xref:fundamentals/
...
+
+Furthermore, the endpoint is protected by authorization (<xref:Microsoft.AspNetCore.Builder.AuthorizationEndpointConventionBuilderExtensions.RequireAuthorization%2A>) to prevent anonymous access.
+
+The `BlazorWasmAuth` client app is simply required to pass an empty object `{}` in the body of the request.
+
+Outside of the logout endpoint, [antiforgery mitigation](xref:security/anti-request-forgery) is only required when submitting form data to the server encoded as `application/x-www-form-urlencoded`, `multipart/form-data`, or `text/plain`. Blazor manages CSRF mitigation for forms in most cases. For more information, see <xref:blazor/forms/index#antiforgery-support>.
+
+Requests to other server API endpoints (web API) with `application/json`-encoded content and [CORS](xref:security/cors) enabled doesn't require CSRF protection. No CSRF protection is required for the `Backend` app's roles (`/roles`) and data processing (`/data-processing`) endpoints.
```
halter73, line 276:
```suggestion
Requests to other server API endpoints (web API) with `application/json`-encoded content and [CORS](xref:security/cors) enabled doesn't require CSRF protection. This is why no CSRF protection is required for the `Backend` app's data processing (`/data-processing`) endpoint. The roles (`/roles`) endpoint does not need CSRF protection because it is a GET endpoint that does not modify any state.
```

halter73 (approved):
> Error for unauthenticated hit on the /data-processing endpoint. What the sample is doing now is trapping it in a try-catch and showing a message. Can't we work out something more graceful than that?

I left [a comment on the blazor-samples PR](https://github.com/dotnet/blazor-samples/pull/210#discussion_r1506924135) suggesting we remove the try/catch. If CORS is correctly configured, you should not get any exceptions for CORS errors. I asked some [follow up questions on the issue](https://github.com/dotnet/blazor-samples/issues/161#issuecomment-1970288937).

Whether the request is authenticated or not should have no impact on how CORS policies are enforced as long as you call `.AllowCredentials()` and use `credentials: 'include'` via `SetBrowserRequestCredentials(BrowserRequestCredentials.Include)` as noted in https://learn.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-8.0#credentials-in-cross-origin-requests. And looking at the latest copy of [BlazorWebAssemblyStandaloneWithIdentity](https://github.com/dotnet/blazor-samples/tree/main/8.0/BlazorWebAssemblyStandaloneWithIdentity), bot things seem to be happening.

guardrex:
Thanks @halter73. I'll wait a day for a response to your question ...

> @blowdart @GrabYourPitchforks Do either of you think we should hedge this statement in some way?

I can't wait too long ... I need to unblock things fairly fast these days for more work ⛰️⛏️😅.

aspnetcore/blazor/security/index.md
blowdart: I bow to Stephen's expertise in this area :)

PR #31894, opened 2024-02-23
Add BWA global Auto approach

### LINKED ISSUE #30002: Dynamic culture approach updates 8.0
This might not be reached until after RTM.

Two tasks ...

* [x] **Resolved on #30576**: The dynamic approach server-side relies on the loc cookie approach and `_Host.cshtml`. I've versioned that whole section out for <8.0. We need to update the approach for BWAs with the `App` component.
* [x] **Resolved on #30860**: Fix the server-side dynamic culture section, which had its `App` component code updated but not its `Program` file (and other instructions) updated.
* [x] Since it's common for localized apps to permit culture changes from anywhere in the app (via for example what we show using the `MainLayout` component), the dynamic approaches should be integrated for BWA client- and server-side user culture changes. Perhaps, the BWA should adopt a state preservation service.

---
#### Document Details

* Content: [ASP.NET Core Blazor globalization and localization](https://learn.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-7.0)
* Product: **aspnet-core**
* GitHub Login: @guardrex
* Microsoft Alias: **riande**

guardrex:
Fixes #30002
Addresses #28161

@hishamco ... I have another 🙈 ***RexHacks!***&trade; 🦖 approach here. It ***SEEMS*** to work, but IDK if there are some 🐉🐉🐉 with what I did or some lurking 😈😈😈 in here.

Because of the complexity of this, I thought you might want to pull it down and run it locally, so I placed my test app at ...

https://github.com/guardrex/BlazorCulturePerComponentInteractivity

### Question 1

There is one thing here that I don't understand. When changing the culture from a CSR component (in the sample app, changing it from the `CultureClient` component), it does seem to update the localization cookie and carryover to the server-rendered component (in the sample app if you then navigate to the `CultureServer` component). It's not clear to me where/how the localization cookie is being updated when that happens. The controller action in the server app isn't called when the culture is changed client-side, and I didn't think that Localization Middleware was reading the updated value from local storage. However, it seems to work. If you can see where/how the culture is being carried over to SSR, I'll add an explanation on it to the article text.

### Question 2

... and another ❓on the delta between the Chilean Spanish date format SSR vs. CSR.

SSR it's in the format ... 23-02-2024
CSR it's in the format ... 23/2/2024

Any idea on why that format is changing in an odd way? It's another odd behavioral difference that I'll explain in the article if it's known why it happens.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/globalization-localization.md](https://github.com/dotnet/AspNetCore.Docs/blob/c98c0ad7941d850291b574b1ac0188bd2d857e41/aspnetcore/blazor/globalization-localization.md) | [ASP.NET Core Blazor globalization and localization](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?branch=pr-en-us-31894) |

### DIFF
--- a/aspnetcore/blazor/globalization-localization.md
+++ b/aspnetcore/blazor/globalization-localization.md
@@ -185,9 +185,9 @@ Optionally, add a menu item to the navigation in the `NavMenu` component (`NavMe
 
 Add the [`Microsoft.Extensions.Localization`](https://www.nuget.org/packages/Microsoft.Extensions.Localization) package to the app.
 
-The [`Accept-Language` header](https://developer.mozilla.org/docs/Web/HTTP/Headers/Accept-Language) is set by the browser and controlled by the user's language preferences in browser settings. In browser settings, a user sets one or more preferred languages in order of preference. The order of preference is used by the browser to set quality values (`q`, 0-1) for each language in the header. The following example specifies United States English, English, and Chilean Spanish with a preference for United States English or English:
+The [`Accept-Language` header](https://developer.mozilla.org/docs/Web/HTTP/Headers/Accept-Language) is set by the browser and controlled by the user's language preferences in browser settings. In browser settings, a user sets one or more preferred languages in order of preference. The order of preference is used by the browser to set quality values (`q`, 0-1) for each language in the header. The following example specifies United States English, English, and Costa Rican Spanish with a preference for United States English or English:
 
-**Accept-Language**: en-US,en;q=0.9,es-CL;q=0.8
+**Accept-Language**: en-US,en;q=0.9,es-CR;q=0.8
 
 The app's culture is set by matching the first requested language that matches a supported culture of the app.
 
@@ -220,17 +220,17 @@ Add the following line to the `Program` file where services are registered:
 builder.Services.AddLocalization();
 ```
 
-In ***server-side development***, you can specify the app's supported cultures immediately after Routing Middleware is added to the processing pipeline. The following example configures supported cultures for United States English and Chilean Spanish:
+In ***server-side development***, you can specify the app's supported cultures immediately after Routing Middleware is added to the processing pipeline. The following example configures supported cultures for United States English and Costa Rican Spanish:
 
 ```csharp
 app.UseRequestLocalization(new RequestLocalizationOptions()
-    .AddSupportedCultures(new[] { "en-US", "es-CL" })
-    .AddSupportedUICultures(new[] { "en-US", "es-CL" }));
+    .AddSupportedCultures(new[] { "en-US", "es-CR" })
+    .AddSupportedUICultures(new[] { "en-US", "es-CR" }));
 ```
 
 For information on ordering the Localization Middleware in the middleware pipeline of the `Program` file, see <xref:fundamentals/middleware/index#middleware-order>.
 
-Use the `CultureExample1` component shown in the [Demonstration component](#demonstration-component) section to study how globalization works. Issue a request with United States English (`en-US`). Switch to Chilean Spanish (`es-CL`) in the browser's language settings. Request the webpage again.
+Use the `CultureExample1` component shown in the [Demonstration component](#demonstration-component) section to study how globalization works. Issue a request with United States English (`en-US`). Switch to Costa Rican Spanish (`es-CR`) in the browser's language settings. Request the webpage again.
 
 > [!NOTE]
 > Some browsers force you to use the default language setting for both requests and the browser's own UI settings. This can make changing the language back to one that you understand difficult because all of the setting UI screens might end up in a language that you can't read. A browser such as [Opera](https://www.opera.com/download) is a good choice for testing because it permits you to set a default language for webpage requests but leave the browser's settings UI in your language.
@@ -240,7 +240,7 @@ When the culture is United States English (`en-US`), the rendered component uses
 * **Date**: 6/7/2021 6:45:22 AM
 * **Number**: 1,999.69
 
-When the culture is Chilean Spanish (`es-CL`), the rendered component uses day/month date formatting (`7/6`), 24-hour time, and period separators in numbers with a comma for the decimal value (`1.999,69`):
+When the culture is Costa Rican Spanish (`es-CR`), the rendered component uses day/month date formatting (`7/6`), 24-hour time, and period separators in numbers with a comma for the decimal value (`1.999,69`):
 
 * **Date**: 7/6/2021 6:49:38
 * **Number**: 1.999,69
@@ -323,7 +323,7 @@ CultureInfo.DefaultThreadCurrentUICulture = new CultureInfo("en-US");
 > [!IMPORTANT]
 > Always set <xref:System.Globalization.CultureInfo.DefaultThreadCurrentCulture> and <xref:System.Globalization.CultureInfo.DefaultThreadCurrentUICulture> to the same culture in order to use <xref:Microsoft.Extensions.Localization.IStringLocalizer> and <xref:Microsoft.Extensions.Localization.IStringLocalizer%601>.
 
-Use the `CultureExample1` component shown in the [Demonstration component](#demonstration-component) section to study how globalization works. Issue a request with United States English (`en-US`). Switch to Chilean Spanish (`es-CL`) in the browser's language settings. Request the webpage again. When the requested language is Chilean Spanish, the app's culture remains United States English (`en-US`).
+Use the `CultureExample1` component shown in the [Demonstration component](#demonstration-component) section to study how globalization works. Issue a request with United States English (`en-US`). Switch to Costa Rican Spanish (`es-CR`) in the browser's language settings. Request the webpage again. When the requested language is Costa Rican Spanish, the app's culture remains United States English (`en-US`).
 
 ## Statically set the server-side culture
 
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/globalization-localization.md
```diff
@@ -777,6 +777,422 @@ If the app adopts ***per-page/component*** interactivity, make the following cha
...
+
+if (result != null)
+{
+    culture = new CultureInfo(result);
+}
+else
+{
+    culture = new CultureInfo("en-US");
+    await js.InvokeVoidAsync("blazorCulture.set", "en-US");
+}
```
hishamco, line 834:
```suggestion
const string defaultCulture = "en-US";

var js = host.Services.GetRequiredService<IJSRuntime>();
var result = await js.InvokeAsync<string>("blazorCulture.get");
var culture = CultureInfo.GetCulture(result ?? defaultCulture);

if (result == null)
{
    await js.InvokeVoidAsync("blazorCulture.set", defaultCulture);
}
```

hishamco (changes requested):

hishamco:
It's a long time from doing localization in Blazor :) but seems everything LGTM

guardrex:
Thanks @hishamco ... and that last commit was to update the client-only section earlier in the article where it also appears.

What about my ❓ in the opening comment :point_up:? ... any ideas on what I can say about those in the article?

hishamco (approved):

hishamco:
> What about my ❓ in the opening comment ☝️? ... any ideas on what I can say about those in the article?

I need to check, but how I can produce the two formats?

guardrex:
Run the sample app, set the culture to Chilean Spanish, and flip back and forth between the `CultureClient` and `CultureServer` components.

guardrex:
It can definitely wait. I wasn't going to merge it until Monday at the earliest because it needs to be edited again.

guardrex:
# ***Yes!*** 🎉 

If I switch to ***Mexican*** Spanish, I get a stable date format. I'll do that ... I'll go with Mexico 🇲🇽  for the code example. That eliminates "Question 2" from being a major concern.

guardrex:
Well ... I'm actually not too keen on Mexican Spanish because the number format (dot separator) is the same that we use in English ...

```
1999.69
```

Chilean Spanish uses a comma separator ...

```
1999,69
```

guardrex:
OK ... I have one ......

# 🇨🇷  ***Costa Rica*** (`es-CR`)

The number and date formats change, ***and*** it's stable across WASM and server rendering.

I'll go with that.

hishamco:
Frankly, I was busy with Orchard Core and something else in my home :)

I hope to find time this weekend, if you didn't merge this yet

hishamco:
@guardrex I did a quick look for what you did, I'm asking why you didn't rely on the cookie that send by ASP.NET Core, this SHOULD work correctly in both Server & Client

FYI we already did something similar long time back in Oqtane Framework

guardrex:
Only because traditionally we didn't take that approach for WASM (standalone or hosted). We always pushed local storage for WASM; so I just assumed that after the Blazor bundle comes down and Auto components go with CSR, they would switch over to local storage. I get the impression from you now that Auto components reaching CSR should just navigate to the server controller and then be redirected back, which is how we've always managed the cookie for Blazor Server <8.0.

I'll try again only adopting our Blazor Server approach (cookie) in a BWA, but I'm buried in call web API work right now and probably won't be able to get back to this until the middle of next week. I'll ping you back here then when I complete a new round of testing with a new test app.

hishamco:
I think both read from JS , right?

guardrex:
I think only on the client (CSR) is when the culture is read from local storage via JS interop.

I think for SSR that the Loc Middleware is reading the culture from the cookie.

When the user changes the culture from an SSR component, the local storage is updated and they're thrown to the controller to update the cookie.

The part that I'm not 100% clear on is what's happening when the culture selector is activated.

If the component (and thus culture selector) are in SSR mode, it seems like it sets the culture in local storage and redirects to the controller and back ... which is updating the cookie.

However ... For CSR with the culture selector, I have it set the new culture in local storage ... ok, that's fine ... and I still have the code there for the controller redirect. I guess that works, but I'm not sure on the mechanics of is the component being rerendered completely (on WASM) after it rendered the first time on WASM. I'm not sure if that's the correct way to go for CSR. It seems like the code should update the cookie, but perhaps it should just be calling a backend endpoint to update it without the redirect.

guardrex:
🤔 ... I guess that's ok for CSR ... it sets the culture into local storage and redirects (for the cookie to be set correctly) ... and then on reload after coming back it sucks the changed culture in from local storage and rerenders the component CSR.

AFAICT, this whole approach is just a bit hacky because there isn't a nice built-in way of managing things.

I guess an alternative to local storage is to deal directly with the loc cookie in JS on the client. Idk if that's any better than this current approach. It would require as much or more code than this, and it would be perhaps equally inelegant as this approach.

guardrex:
I'll place the bits here in one spot to make it simpler to see the overall setup ...

### Server project of the BWA

Set up loc in the `Program` file ...

```csharp
var supportedCultures = new[] { "en-US", "es-CR" };
var localizationOptions = new RequestLocalizationOptions()
    .SetDefaultCulture(supportedCultures[0])
    .AddSupportedCultures(supportedCultures)
    .AddSupportedUICultures(supportedCultures);

app.UseRequestLocalization(localizationOptions);
```

Controller that can set the culture to the loc cookie and redirect back ...

```csharp
[Route("[controller]/[action]")]
public class CultureController : Controller
{
    public IActionResult Set(string culture, string redirectUri)
    {
        if (culture != null)
        {
            HttpContext.Response.Cookies.Append(
                CookieRequestCultureProvider.DefaultCookieName,
                CookieRequestCultureProvider.MakeCookieValue(
                    new RequestCulture(culture, culture)));
        }

        return LocalRedirect(redirectUri);
    }
}
```

The `App` component has a script to get/set the culture for local storage ...

```javascript
window.blazorCulture = {
  get: () => window.localStorage['BlazorCulture'],
  set: (value) => window.localStorage['BlazorCulture'] = value
};
```

### Client project of the BWA

In the `Program` file, set a default or read an existing culture and set the culture for the app running CSR ...

```csharp
CultureInfo culture;
var js = host.Services.GetRequiredService<IJSRuntime>();
var result = await js.InvokeAsync<string>("blazorCulture.get");

if (result != null)
{
    culture = new CultureInfo(result);
}
else
{
    culture = new CultureInfo("en-US");
    await js.InvokeVoidAsync("blazorCulture.set", "en-US");
}

CultureInfo.DefaultThreadCurrentCulture = culture;
CultureInfo.DefaultThreadCurrentUICulture = culture;
```

... and here's the meat 🍖 ... the `CultureSelector` component that appears throughout the app via the `MainLayout` component and given an Interactive Auto render mode (`<CultureSelector @rendermode="InteractiveAuto" />`).

```razor
﻿@using System.Globalization
@using System.Runtime.InteropServices
@inject IJSRuntime JS
@inject NavigationManager Navigation

<p>
    <label>
        Select your locale:
        <select @bind="Culture">
            @foreach (var culture in supportedCultures)
            {
                <option value="@culture">@cultureDict[culture.Name]</option>
            }
        </select>
    </label>
</p>

@code
{
    private Dictionary<string, string> cultureDict = 
        new()
        {
            { "en-US", "English (United States)" },
            { "es-CR", "Spanish (Costa Rica)" }
        };

    private CultureInfo[] supportedCultures = new[]
    {
        new CultureInfo("en-US"),
        new CultureInfo("es-CR"),
    };

    private CultureInfo Culture
    {
        get => CultureInfo.CurrentCulture;
        set
        {
            if (CultureInfo.CurrentCulture != value)
            {
                JS.InvokeVoidAsync("blazorCulture.set", value.Name);

                var uri = new Uri(Navigation.Uri)
                    .GetComponents(UriComponents.PathAndQuery, UriFormat.Unescaped);
                var cultureEscaped = Uri.EscapeDataString(value.Name);
                var uriEscaped = Uri.EscapeDataString(uri);

                Navigation.NavigateTo(
                    $"Culture/Set?culture={cultureEscaped}&redirectUri={uriEscaped}",
                    forceLoad: true);
            }
        }
    }
}
```

The basic idea is that regardless of rendering location, the culture is kept up-to-date in local storage, and the loc cookie is kept up-to-date via the controller. When going into CSR, local storage is read. When in SSR, the loc cookie is used. In order to avoid deltas on naming of the cultures between SSR and CSR for the dropdown list, the component uses a custom dictionary for the text displayed to the user.

AFAICT, this is how this approach should work. The only other alternatives that I could think of would be trying to deal with the loc cookie directly for CSR without a redirect, but it seems like it would be no better than this in terms of the amount of code and complexity of code (but it would potentially save the need for a redirect) ... ***IF*** it could be make to work cleanly. I tested a bit, and that approach had some 🐉 that didn't even let me get it running correctly. Another idea would be to maintain loc outside of the ASP.NET Core loc bits ... try to use some kind of service-based loc management. I didn't investigate it tho.

guardrex:
@hishamco ... I'm going to go ahead with this. It's fairly close to what we had here before, merging the server and client approaches in a BWA for components rendering either way, server or client. It seems stable 🤞🍀. I propose to take reader feedback on it. I'm sure if devs run into trouble that I'll receive the normal threats of 🦖 ***dismemberment*** 😨😆.

hishamco:
If you look closely

```chsarp
 HttpContext.Response.Cookies.Append(
                CookieRequestCultureProvider.DefaultCookieName,
                CookieRequestCultureProvider.MakeCookieValue(
                    new RequestCulture(culture, culture)));
```

you set the cookie in the server, but you are using the local storage on the client, that's why I prefer to use the cookies everywhere

guardrex:
Yes, but I don't understand how it would work for CSR without using JS interop to interact directly with the cookie. Even then when I tried that approach, it didn't fully work for me. I'll let Dan know that we need a PU engineer to build out the approach in a sample app for me. I'll ping you on the new issue that I open later with a cross-link to whatever they provide.

hishamco:
I can create a demo if you want

guardrex:
Yes, please do.

BTW ... I don't know why you aren't on the article with a byline ("By Hisham Bin Ateya" ... cross-linked to your blog, X account, or company), but would you like a byline on the article with a cross-link? If so, what do you want it linked to?

guardrex:
Let's pick up with discussion on the issue.

PR #31913, opened 2024-02-27
Hub page update per data and OKR

### LINKED ISSUE #31578: Improve ASP.NET Hub page based on data available
For the ASP.NET hub navigation topic:
https://learn.microsoft.com/en-us/aspnet/core/?view=aspnetcore-8.0

Looking back on data for the ASP.NET Hub navigation topic we are seeing a clear need to move links most people are hitting to above the fold where needed and reduce the noise for links that are getting very little use, provided they are not an SEO issue in disguise.

Data to review for the design update includes:

- [x] Review the Hub pages current design with Kelley who is driving the larger HUB improvement effort across products.
- [x] Review successful changes made so far for hub pages for other developer focused audiences to see what we can take advantage of.
- [x] Heat map data review
- [x] Traffic flow for click-through's
- [x] Look for high bounce rates for click-through to make sure folks were really getting where they wanted to be.
- [x] Top search scenarios for ASP.NET Core to make sure those are still reflected in the hub design.

---
[Associated WorkItem - 203535](https://dev.azure.com/msft-skilling/Content/_workitems/edit/203535)

wadepickett:
[Internal review: Hub page for ASP.NET Core](https://review.learn.microsoft.com/en-us/aspnet/core/?branch=pr-en-us-31913&view=aspnetcore-9.0)

Fixes #31578 

The ASP.NET Core hub page was specified as one of the Hub pages that could use some changes per hub OKR.  It had some high traffic links below the fold that needed to move up, some low traffic links above the fold where we provided too many doc choices per app type, and too many duplicate links.

- I reviewed traffic, clickthrough, exit rate, bounce, heatmap, and general report data available.
- Verified what should move above the fold per that data.
- Reviewed current published HUB with OKR stakeholders for suggestions per site wide initiative for Hub pages.
- Removed some duplicate links where it made sense.

**Top section:**
Summary: We can move the overall choices here to a more reasonable 6 by removing a low traffic card and high bounce/somewhat low traffic duplicate link.

**Remove Give feedback on ASP.NET Docs**
Why: It was receiving low relative traffic, and we have a feedback link on every single doc.

**Remove ASP.NET 4.x Documentation card.**
Why:  
- It had high bounce. 
  - We suspect folks are going there expecting the whole doc set or latest then bouncing out when they don't find that.  Being at the top as one of the first cards that indicates "Documentation" may be causing some confusion.
- It is a duplicate.  We have that link in a clear context on the "Previous ASP.NET framework versions" card.

Follow up:
- Watch the bounce rate for this after this PR change for the month of March, if it goes down significantly then the problem is the context it is presented in.  If the bounce rate stays high, we may have a different issue.

**2nd Section: Develop ASP.NET Core apps**

**Move HTTP API apps to the first position and move Interactive client-side Blazor apps to the 2nd postion**
Why:  _API_ has 3 times as much traffic than the _Blazor_ apps card which was in the first position and _API_ is about a 3rd more than _Create your first web UI_ which has its own top section card.

**Remove long tail sub-subject's links from each app card** 
Why: Specifically where they are around 0.15% of link traffic where we have more than 4 links to fit more towards the top of the fold and to avoid overwhelming with too many choices for lower interest topics.

**3rd Section:**
I left this last section alone.  It is all below the fold. It didn't have any high traffic interest areas that needed to move up to be faster to find.  It also doesn't hurt anything to leave those topics there and discoverable even though overall they are low traffic links.

### DIFF
--- a/aspnetcore/index.yml
+++ b/aspnetcore/index.yml
@@ -53,6 +53,9 @@ conceptualContent:
     - title: "HTTP API apps"
       summary: Develop HTTP services with ASP.NET Core
       links:
+        - url: fundamentals/apis.md
+          itemType: overview
+          text: "Overview"
         - url: tutorials/min-web-api.md
           itemType: get-started
           text: "Create a minimal web API with ASP.NET Core"
@@ -62,12 +65,6 @@ conceptualContent:
         - url: tutorials/web-api-help-pages-using-swagger.md
           itemType: tutorial
           text: "Generate web API help pages with Swagger / OpenAPI"
-        - url: web-api/action-return-types.md
-          itemType: concept
-          text: "Controller action return types"
-        - url: fundamentals/apis.md
-          itemType: overview
-          text: "See more"
     # Card
     - title: "Interactive client-side Blazor apps"
       summary: Develop with reusable UI components that can take advantage of WebAssembly for near-native performance
@@ -75,15 +72,15 @@ conceptualContent:
         - url: blazor/index.md
           itemType: overview
           text: "Overview"
+        - url: blazor/hosting-models.md
+          itemType: concept
+          text: "Blazor hosting models"
         - url: https://dotnet.microsoft.com/learn/aspnet/blazor-tutorial/intro
           itemType: get-started
           text: "Build your first Blazor app"
         - url: blazor/tutorials/build-a-blazor-app.md
           itemType: get-started
           text: "Build your first Blazor app with reusable components"
-        - url: blazor/hosting-models.md
-          itemType: concept
-          text: "Blazor hosting models"
     # Card
     - title: "Page-focused web UI with Razor Pages"
       summary: "Develop page-focused web apps with a clean separation of concerns"
@@ -123,9 +120,6 @@ conceptualContent:
     - title: Data-driven web apps
       summary: "Create data-driven web apps in ASP.NET Core"
       links:
-        - url: tutorials/first-mvc-app/start-mvc.md
-          itemType: tutorial
-          text: "SQL with ASP.NET Core"
         - url: blazor/components/data-binding.md
           itemType: concept
           text: "Data binding in ASP.NET Core Blazor"
@@ -138,6 +132,9 @@ conceptualContent:
         - url: data/ef-mvc/intro.md
           itemType: tutorial
           text: "Entity Framework Core with ASP.NET Core MVC"
+        - url: tutorials/first-mvc-app/start-mvc.md
+          itemType: tutorial
+          text: "Get started with ASP.NET Core MVC (SQL Server Express and SQLite)"
       # Card
     - title: Real-time web apps with SignalR
       summary: "Add real-time functionality to your web app, enable server-side code to push content instantly"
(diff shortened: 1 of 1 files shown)


aspnetcore/index.yml
```diff
@@ -89,15 +65,25 @@ conceptualContent:
...
+        - url: blazor/tutorials/build-a-blazor-app.md
+          itemType: get-started
+          text: "Build your first Blazor app with reusable components"
+        - url: blazor/hosting-models.md
           itemType: concept
-          text: "Handle errors"
-        - url: tutorials/web-api-javascript.md
-          itemType: tutorial
-          text: "Call an ASP.NET Core web API with JavaScript"
+          text: "Blazor hosting models"
```
tdykstra, line 86: Seems like this would fit right after the overview.

aspnetcore/index.yml
```diff
@@ -111,15 +97,9 @@ conceptualContent:
         - url: mvc/views/razor.md
           itemType: concept
           text: "Razor syntax"
```
tdykstra, line 99: This is Razor views as used in MVC apps, not in Razor pages apps as the card title promises.

tdykstra (approved):
Looks good, a couple suggestions for you to consider.

aspnetcore/index.yml
```diff
@@ -89,15 +65,25 @@ conceptualContent:
         - url: web-api/action-return-types.md
           itemType: concept
           text: "Controller action return types"
-        - url: web-api/advanced/formatting.md
-          itemType: concept
-          text: "Format response data"
-        - url: web-api/handle-errors.md
+        - url: fundamentals/apis.md
+          itemType: overview
+          text: "See more"
```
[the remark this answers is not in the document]
wadepickett: Thanks.  Since we didn't have a landing page just for HTTP API apps, the overview was the only place I could point to as a "See more" starting place.  I hate to create landing pages for each area to go to as the "See more" since I think sub area landing pages will not really get very much traffic overall.  Thoughts?

I will move the overview (which was the "See more" link) above the tutorial and omit the return types link, but there will also no longer be a "See more".

aspnetcore/index.yml
wadepickett: OK thanks.  Moving Blazor hosting models to right after the overview.

aspnetcore/index.yml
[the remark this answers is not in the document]
tdykstra: I don't think a see more link is essential to have in the hub, as the overview and tutorial will both have see also sections.

aspnetcore/index.yml
wadepickett: The Razor syntax page is for Razor syntax anywhere you use it in both "ASP.NET Core MVC or Razor Pages app".  So it mentions both app types in the doc.  The doc intro could be more clear however.

aspnetcore/index.yml
```diff
@@ -133,18 +113,31 @@ conceptualContent:
...
           itemType: concept
           text: "Routing to controller actions"
-        - url: mvc/controllers/testing.md
+    # Card
+    - title: Data-driven web apps
+      summary: "Create data-driven web apps in ASP.NET Core"
+      links:
+        - url: tutorials/first-mvc-app/start-mvc.md
+          itemType: tutorial
+          text: "SQL with ASP.NET Core"
```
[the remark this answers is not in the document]
wadepickett: Good point.  Moving this to last, using the tutorial title, but adding (SQL Server Express and SQLite) so folks know what db is included and walked through.

PR #31924, opened 2024-02-28
Update redirecttologin-component.md

yogyogi:
It is inside "Layout" folder. I have checked it in .NET 8.0 version of blazor wasm project.


guardrex:
Hello @yogyogi ... We can't do this without versioning it, and I'm ok not doing so in this case as long as the text doesn't mention it elsewhere. In fact, I may have updated it to remove this on purpose earlier in order to avoid having literally hundreds (😨) of versioning blocks to specify locations of things that changed at 8.0.

guardrex:
Oh ... wait ... hold a sec. I see what you mean about it saying `Shared` NOW ... today.

Sorry ... I'm wrapping up most of my work for the day and trying to get out of here ... going tooooooo FAST 🏃.

The change will be to remove the folder reference. I'll make the update and merge this. Thanks for the PR.

guardrex (approved):

PR #31925, opened 2024-02-28
Update logindisplay-component.md

yogyogi:
It is inside "Layout" folder and not "Shared". I have checked it in .NET 8.0 version of blazor wasm project.


guardrex (approved):
Thanks @yogyogi! 🎸 ... As you can see, I'm dealing with this in a version-agnostic way to avoid ***more versioning blocks*** 😩.

PR #32028, opened 2024-03-11
Blazor WASM cookie security for web APIs

### LINKED ISSUE #32008: Cookie security for Call web API article
### Description

Need a section/cross-link(s) for cookie security, which is used for WASM security now (8.0+).

I think the BWA security aspects are at least a decent start. The security samples (BFF and non-BFF) are cross-linked, and the security articles explain how the security works.

What we're lacking is the cookie security aspects for WASM+OIDC. There's a brief POST example (`/data-processing`), but I might want more than that ... I think I'd like a roles-enabled POST endpoint, too. The WASM+Identity article needs a little coverage on how the cookie security works (if it isn't there ... I need to confirm what I have).

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/call-web-api?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/call-web-api.md

### Document ID

c7e59a08-1c60-32c2-75fd-33cb77ff7a5d

### Article author

@guardrex

guardrex:
Fixes #32008

Jeremy ... Three things on this one ...

* Do you want to say more than this in the *Call web API* article? I figure that this along with the *Standalone with Identity* article (+sample), which is cross-linked here, should be enough. Are there any :smiling_imp: ***gotchas*** to call out or anything else?
* AFAIK, the only versioning that this needs is for the *Standalone with Identity* article (>=8.0). I think the rest of this always worked. AFAIK, we just didn't have it because we were pushing token auth for just about everything. We have a bit of this in the next section, which is generalized for Fetch API options. We just didn't have it in a dedicated cookie auth section. Anyway, any concerns about showing this for all versions?
* The API engineering comment for the handler for `AddHttpMessageHandler` states emphatically ...

  > The handler type must be registered as a transient service.

   ... but it's registered scoped ...

  ```csharp
  builder.Services.AddScoped<CookieHandler>();
  ```

  What's the story on that? I feel like we need to explain this exception to the rule in the article where I show the service registration.

... and just an FYI that I added a second seeded test user to the sample app and a second roles-authorized endpoint with a page where Leela can call both endpoints and Harry can only call one. https://github.com/dotnet/blazor-samples/pull/241

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/call-web-api.md](https://github.com/dotnet/AspNetCore.Docs/blob/1996b9932364eaf218e0bfe45468b46a0837382d/aspnetcore/blazor/call-web-api.md) | [Call a web API from ASP.NET Core Blazor](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/call-web-api?branch=pr-en-us-32028) |
| [aspnetcore/blazor/security/webassembly/additional-scenarios.md](https://github.com/dotnet/AspNetCore.Docs/blob/1996b9932364eaf218e0bfe45468b46a0837382d/aspnetcore/blazor/security/webassembly/additional-scenarios.md) | [ASP.NET Core Blazor WebAssembly additional security scenarios](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/additional-scenarios?branch=pr-en-us-32028) |

### DIFF
--- a/aspnetcore/blazor/call-web-api.md
+++ b/aspnetcore/blazor/call-web-api.md
@@ -655,6 +655,56 @@ The `BlazorWebAppCallWebApi` [sample app](#sample-apps) demonstrates calling a w
 
 :::moniker-end
 
+## Cookie-based request credentials
+
+*The guidance in this section applies to client-side scenarios that rely upon an authentication cookie.*
+
+For cookie-based authentication, which is considered more secure than bearer token authentication, cookie credentials can be sent with each web API request by calling <xref:Microsoft.Extensions.DependencyInjection.HttpClientBuilderExtensions.AddHttpMessageHandler%2A> with a <xref:System.Net.Http.DelegatingHandler> on a preconfigured <xref:System.Net.Http.HttpClient>. The handler configures <xref:Microsoft.AspNetCore.Components.WebAssembly.Http.WebAssemblyHttpRequestMessageExtensions.SetBrowserRequestCredentials%2A> with <xref:Microsoft.AspNetCore.Components.WebAssembly.Http.BrowserRequestCredentials.Include?displayProperty=nameWithType>, which advises the browser to send credentials with each request, such as cookies or HTTP authentication headers, including for cross-origin requests.
+
+`CookieHandler.cs`:
+
+```csharp
+public class CookieHandler : DelegatingHandler
+{
+    protected override Task<HttpResponseMessage> SendAsync(
+        HttpRequestMessage request, CancellationToken cancellationToken)
+    {
+        request.SetBrowserRequestCredentials(BrowserRequestCredentials.Include);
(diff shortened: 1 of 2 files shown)


JeremyLikness (approved):
LGTM. To answer your questions:

- I don't believe we need to say more, and the cross-linking should be fine
- No concerns with showing it for all versions
- That's a great question I'll need to defer to @halter73 to answer

Regards, Jeremy

guardrex:
Thanks @JeremyLikness.

@halter73 ... The API comment for the delegate provided to `AddHttpMessageHandler` says that the service should be ***transient***. However, the `CookieHandler` is registered scoped. Because it goes against the API remark, can we explain something about that delta in the article?

PR #32043, opened 2024-03-13
Update parameter-binding8.md

jensj:
Binding form-posted data to the boolean property "IsCompleted" does not work.

The example given is not working at all.

The form-posted data includes the value "on" is not applicable to the expected value "true" or "false" of the target which is of boolean type.


jensj:
@dotnet-policy-service agree


Den ons 13 mars 2024 kl 10:55 skrev dotnet-policy-service[bot] <
***@***.***>:

guardrex:
@jensj ... An issue should be opened for this. We don't just delete a section that likely just has a bug in the code. An issue should be opened from the bottom of the page. I'll do that now.

PR #32075, opened 2024-03-18
Fix typo

Jessuhh:
---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/fundamentals/middleware/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/8bebed1b227cb1835a82ab1fb0458524e8adb653/aspnetcore/fundamentals/middleware/index.md) | [ASP.NET Core Middleware](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/middleware/index?branch=pr-en-us-32075) |


rickklaasboer:
This sentence still seems weird to me, it's essentially saying `Modifying the HttpResponse after the response has begun will result in an exception being thrown.` But it's so strangely worded it doesn't make sense at all.

guardrex:
Hello ...

First off, the verb is correct in its present form (throw) because "changes" is plural.

WRT the sentence structure and interpretation of what it means, I think moving the prepositional phrase might help. I'll do that now, and we can take a look. *Stand-by ..............................*

guardrex:
@Jessuhh & @rickklaasboer ... See how it looks now with the changes that I applied.

guardrex:
Ok ... looking good I think ...

> Don't call next.Invoke during or after the response has been sent to the client. After an \<xref:Microsoft.AspNetCore.Http.HttpResponse> has started, changes result in an exception. For example, setting headers and a status code throw an exception after the response starts. Writing to the response body after calling next:
>
> \* May cause a protocol violation, such as writing more than the stated Content-Length.
> \* May corrupt the body format, such as writing an HTML footer to a CSS file.

guardrex:
@Jessuhh ... I'm going to go ahead with the updates. Thanks for asking, and thanks @rickklaasboer for your help.

PR #32078, opened 2024-03-19
[Blazor] Event handling - button "select" => button "click"

hakenr:
"Selecting" the button to mean triggering the `click` event might seem unusual. The term "clicking" is more commonly used and understood.

@guardrex If you agree, I will change the corresponding comments in sample code.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/event-handling.md](https://github.com/dotnet/AspNetCore.Docs/blob/ae53b94808116f89454c4af0f9e393d4f23ff827/aspnetcore/blazor/components/event-handling.md) | [ASP.NET Core Blazor event handling](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/event-handling?branch=pr-en-us-32078) |

### DIFF
--- a/aspnetcore/blazor/components/event-handling.md
+++ b/aspnetcore/blazor/components/event-handling.md
@@ -29,7 +29,7 @@ For event handling:
 
 The following code:
 
-* Calls the `UpdateHeading` method when the button is selected in the UI.
+* Calls the `UpdateHeading` method when the button is clicked in the UI.
 * Calls the `CheckChanged` method when the checkbox is changed in the UI.
 
 :::moniker range=">= aspnetcore-8.0"
@@ -74,7 +74,7 @@ The following code:
 
 In the following example, `UpdateHeading`:
 
-* Is called asynchronously when the button is selected.
+* Is called asynchronously when the button is clicked.
 * Waits two seconds before updating the heading.
 
 :::moniker range=">= aspnetcore-8.0"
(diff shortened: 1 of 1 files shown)


guardrex:
Hi @hakenr ... It's by design. See 👉 https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/c/click. It would be ok for ***mice*** 🖱️, but we don't make that distinction around here because .NET runs across device types. Nothing is mouse-specific (AFAIK 🤔), so we just go with a blanket rule here and avoid "click." However, we do use "right-click," as there's no such thing as a "right-select" concept in the industry.

hakenr:
@guardrex Just some feedback for further consideration:

I understand the reasoning behind avoiding the term "click," but it feels odd to refer to it as "selecting a button." It seems like an attempt to replace a widely accepted simplification. The event name is click, and Blazor operates on top of HTML elements, where "click" carries a broader significance. This isn't about drafting specifications for new browser developers, where emphasizing a broader meaning might be relevant.

To my knowledge, the Blazor documentation might be the first instance where I've seen this approach. For comparison, consider these references:
- https://www.google.com/search?q=to+select+a+button
- https://learn.microsoft.com/en-us/dotnet/api/system.web.ui.webcontrols.button.click
- https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.control.click
- https://www.w3schools.com/jsref/event_onclick.asp
- https://developer.apple.com/documentation/swiftui/button - uses "to activate button"
- https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event (does not use click nor select)
- 
This perspective might prompt a reconsideration of the terminology, aligning more closely with established norms.

guardrex:
This is *WAY* above my pay grade 😆. This phrasing decision was made at a much higher level of docs management (a committee I believe) than anyone around here is part of, and it's been MS's convention for as long as I've been here. The only thing that I can recommend is to open an issue from the style manual page for "click" using the **This page** feedback button and issue form ...

https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/c/click

Unfortunately, I haven't had much luck submitting suggestions via those pages. I don't even bother opening issues there any longer, but you may have better luck.

PR #32086, opened 2024-03-19
[Blazor] Event handling - ParentChild2.razor without Task.Yield()

hakenr:
I believe the `await Task.Yield()` in the `ParentChild2.razor` example no longer makes sense. It probably remained from an earlier era of Blazor when these in-line lambdas had to be asynchronous.

cc @guardrex 

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/event-handling.md](https://github.com/dotnet/AspNetCore.Docs/blob/e692389396709bbe8bee0398aa98ac0dcc7ae3c3/aspnetcore/blazor/components/event-handling.md) | [ASP.NET Core Blazor event handling](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/event-handling?branch=pr-en-us-32086) |

### DIFF
--- a/aspnetcore/blazor/components/event-handling.md
+++ b/aspnetcore/blazor/components/event-handling.md
@@ -681,7 +681,8 @@ The following parent-child example demonstrates the technique.
 @code {
     private string message1 = string.Empty;
     private string message2 = string.Empty;
-}```
+}
+```
 
 The second occurrence of `Child2` component demonstrates an asynchronous callback and the new `message2` value will be assigned and rendered with a delay of two seconds.
 
--- a/aspnetcore/blazor/components/event-handling.md
+++ b/aspnetcore/blazor/components/event-handling.md
@@ -663,16 +663,27 @@ The following parent-child example demonstrates the technique.
 ```razor
 @page "/parent-child-2"
 
-<Child2 OnClickCallback="@((value) => { messageText = value; })" />
+<PageTitle>Parent Child 2</PageTitle>
 
-<p>
-    @messageText
-</p>
+<h1>Parent Child 2 Example</h1>
+
+<div>
+    <Child2 OnClickCallback="(value) => { message1 = value; }" />
+    @message1
+</div>
+
+<div>
+    <Child2 OnClickCallback=
+        "async (value) => { await Task.Delay(2000); message2 = value; }" /> 
+    @message2
+</div>
 
 @code {
-    private string messageText = string.Empty;
-}
-```
+    private string message1 = string.Empty;
+    private string message2 = string.Empty;
+}```
+
+The second occurrence of `Child2` component demonstrates an asynchronous callback and the new `message2` value will be assigned and rendered with a delay of two seconds.
 
 ## Prevent default actions
 
--- a/aspnetcore/blazor/components/event-handling.md
+++ b/aspnetcore/blazor/components/event-handling.md
@@ -684,7 +684,7 @@ The following parent-child example demonstrates the technique.
 }
 ```
 
-The second occurrence of `Child2` component demonstrates an asynchronous callback and the new `message2` value will be assigned and rendered with a delay of two seconds.
+The second occurrence of the `Child2` component demonstrates an asynchronous callback, and the new `message2` value is assigned and rendered with a delay of two seconds.
 
 ## Prevent default actions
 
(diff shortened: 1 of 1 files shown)


guardrex:
Let's have one or both of them look at this as well. I'll add this to my email.

guardrex:
@danroth27 ... I pinged Steve (email), but my message may have been lost in the shuffle.

Do you recall why we had this demo code set up this way? ...

```razor
<Child2 OnClickCallback="@(async (value) => { await Task.Yield(); messageText = value; })" />
```

IIRC, it was because we were trying to demo async ***in the lambda***. `Task.Yield()` was likely chosen purely for demo purposes. Lambdas never had to be async AFAIK ... it was just that we wanted to put something here async that's runnable. I'm 🤔 that the example should be changed.

The existing `Child2` component stays the same ...

```razor
<h3>Child2 Component</h3>

<button @onclick="TriggerEvent">Click Me</button>

@code {
    [Parameter]
    public EventCallback<string> OnClickCallback { get; set; }

    private async Task TriggerEvent()
    {
        await OnClickCallback.InvokeAsync("Blaze It!");
    }
}
```

The `ParentChild2` component is changed to the following. Note that we don't specify a render mode (e.g., `@rendermode InteractiveServer`) because we assume global render modes for our examples.

```razor
@page "/parent-child-2"

<PageTitle>Parent Child 2</PageTitle>

<h1>Parent Child 2 Example</h1>

<div>
    <Child2 OnClickCallback="@(async (value) => { await Task.Yield(); messageText1 = value; })" />
    @messageText1
</div>

<div>
    <Child2 OnClickCallback="@((value) => { messageText2 = value; })" /> 
    @messageText2
</div>

@code {
    private string messageText1 = string.Empty;
    private string messageText2 = string.Empty;
}
```

danroth27:
> Do you recall why we had this demo code set up this way?
> `<Child2 OnClickCallback="@(async (value) => { await Task.Yield(); messageText = value; })" />`

`Task.Yield()` will force the method to return asynchronously, so I assume this is trying to show how an async event handler that doesn't complete synchronously can still handled. Making the change to show both he sync and async versions makes sense to me.

Note that syntactically I don't think you need the lambda expressions to be explicit Razor expressions. You're already in a C# context so the wrapping `@()` can be dropped.

guardrex:
Thanks @danroth27.

> don't think you need the lambda expressions to be explicit Razor expressions

... that aspect might be due to an ancient limitation (i.e., experimental days of ancient Blazor times 👴). I just confirmed that all is well without them 🎉.

@hakenr ... Can you use my recommended replacement?

hakenr:
I still think the `Task.Yield()` is very rarely seen in today's codebases and most of the developers won't understand the "just to force the method to run asynchronously" purpose of it in this example. From my POV something like `await Task.Delay(2000)` works better (and we can add explanation "This version is asynchronous and the new `@messageText2` value will be assigned and rendered after two seconds").

So I would go for:
```razor
@page "/parent-child-2"

<PageTitle>Parent Child 2</PageTitle>

<h1>Parent Child 2 Example</h1>

<div>
    <Child2 OnClickCallback="(value) => { messageText1 = value; }" />
    @messageText1
</div>

<div>
    <Child2 OnClickCallback="async (value) => { await Task.Delay(2000); messageText2 = value; }" /> 
    @messageText2
</div>

@code {
    private string messageText1 = string.Empty;
    private string messageText2 = string.Empty;
}
```

(I also moved the asynchronous version to the second occurrence, but it is not that important.)

guardrex:
That sounds good to me.

guardrex:
We're going to have a problem with the line length. We need to keep that long line at <= 85 characters.

aspnetcore/blazor/components/event-handling.md
```diff
@@ -663,17 +663,29 @@ The following parent-child example demonstrates the technique.
...
+</div>
 
 @code {
-    private string messageText = string.Empty;
+    private string message1 = string.Empty;
+    private string message2 = string.Empty;
 }
 ```
 
+The second occurrence of `Child2` component demonstrates an asynchronous callback and the new `message2` value will be assigned and rendered with a delay of two seconds.
```
guardrex, line 687: Tiny NITs ......

```suggestion
The second occurrence of the `Child2` component demonstrates an asynchronous callback, and the new `message2` value is assigned and rendered with a delay of two seconds.
```

guardrex (changes requested):
Couple of tiny NITs and it will be ready to go! 🍻

guardrex (approved):
Excellent ... thx! 🍻

PR #32089, opened 2024-03-19
[Blazor] Lifecycle - clear formulation for conditions when rendering is avoided

hakenr:
The original text does not explicitly express that both conditions have to be met to avoid further rendering.

cc @guardrex


guardrex:
I'll move this suggestion to the issue for Steve/Javier to consider. We can hold this open here in case they agree with the change there.

PR #32090, opened 2024-03-19
[Blazor] Lifecycle - remove breaking-changes notice

hakenr:
Such changes, typically breaking, are now uncommon, and the potential for these changes is inherent in the definition of a "major release." I believe there's no need to intimidate developers in the documentation by explicitly mentioning this.

cc @guardrex 

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/lifecycle.md](https://github.com/dotnet/AspNetCore.Docs/blob/2bf246e565d58cc4c207d9bc5222e3322f0253d3/aspnetcore/blazor/components/lifecycle.md) | [ASP.NET Core Razor component lifecycle](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?branch=pr-en-us-32090) |

### DIFF
--- a/aspnetcore/blazor/components/lifecycle.md
+++ b/aspnetcore/blazor/components/lifecycle.md
@@ -18,7 +18,7 @@ This article explains the ASP.NET Core Razor component lifecycle and how to use 
 
 The Razor component processes Razor component lifecycle events in a set of synchronous and asynchronous lifecycle methods. The lifecycle methods can be overridden to perform additional operations in components during component initialization and rendering.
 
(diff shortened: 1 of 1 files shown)


guardrex:
I'd like to leave this remark for a couple of reasons ...

* This remark was approved by the PU. This is what they want to tell devs.
* Devs who work across versions of Blazor should be told that content versioning in the article might not capture all of the lifecycle processing changes that were made over time. It's not meant to intimidate devs. It's more like a warning that they might see differences that the article doesn't cover.

However, I think the remark can be improved. I'll change it here, and we can take a look. *Stand-by .......................*

guardrex:
I moved the sentiment into the first sentence. See how it looks :eyes:.

hakenr:
While I appreciate the omission of the last sentence, the new text added to the first sentence almost sounds like "this documentation is outdated, and we are aware of it." However, that's not the case.

It's all just cosmetic, so I'll leave it up to you.

guardrex:
There's a difference between outdated and *incomplete*. I think the phrasing leans more in the direction of incomplete, which is the purpose of the remark. Devs have discovered MIA content, but the PU has said repeatedly to devs that this article is a known simplification of complex framework behavior that they don't want to describe. Let's go with the update.

PR #32125, opened 2024-03-24
[Blazor] - Lifecycle - Incorrect usage of "unmanaged" term

hakenr:
In the context of Blazor, `IDisposable` and `IAsyncDisposable` enable components to perform cleanup logic. However, it's important to note that this still involves managed logic. Components are generally not designed to directly use unmanaged resources, like file handles, which would necessitate the use of finalizers (destructors).

Therefore, we should avoid using the term _unmanaged_ in the context of Blazor.

(I'll propose adding an explanation about why we don't use destructors in Blazor in a separate issue.)

cc @guardrex

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/lifecycle.md](https://github.com/dotnet/AspNetCore.Docs/blob/d625fce4a143322515385fcad264e7964d4fd178/aspnetcore/blazor/components/lifecycle.md) | [ASP.NET Core Razor component lifecycle](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?branch=pr-en-us-32125) |

### DIFF
--- a/aspnetcore/blazor/components/lifecycle.md
+++ b/aspnetcore/blazor/components/lifecycle.md
@@ -565,7 +565,7 @@ Although the content in this section focuses on Blazor Server and stateful Signa
 
 ## Component disposal with `IDisposable` and `IAsyncDisposable`
 
-If a component implements <xref:System.IDisposable>, <xref:System.IAsyncDisposable>, or both, the framework calls for unmanaged resource disposal when the component is removed from the UI. Disposal can occur at any time, including during [component initialization](#component-initialization-oninitializedasync).
+If a component implements <xref:System.IDisposable>, <xref:System.IAsyncDisposable>, or both, the framework calls for resource disposal when the component is removed from the UI. Disposal can occur at any time, including during [component initialization](#component-initialization-oninitializedasync).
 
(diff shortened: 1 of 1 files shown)


guardrex:
I think this language was originally sent over by the PU. Just to be sure that they're ok with the change, I'll ping Mackinnon.

MackinnonBuck (approved):
Yeah, we can _probably_ remove the use of "unmanaged" here. Despite the ["Implement a Dispose method"](https://learn.microsoft.com/dotnet/standard/garbage-collection/implementing-dispose) docs suggesting that the primary use of a `Dispose()` method is to clean up unmanaged resources, that's not as common a scenario for Blazor components (although that depends slightly on what your [definition of an unmanaged resource](https://stackoverflow.com/a/3607226) is).

The `Dispose()/DisposeAsync()` methods are more like lifecycle methods in the context of Blazor components. They're useful for doing things like:
* Cleaning up browser-side state when the component gets removed from the page (e.g., via JS interop)
* Unregistering event handlers (like `NavigationManager.LocationChanged`)
* Disposing resources shared with JavaScript (e.g., `DotNetObjectReference`)
* etc.

...and I wouldn't categorize these as cleaning up "unmanaged" resources. Although, you _could_ argue that JavaScript state is unmanaged...

Tbh, I could go either way, but leaning toward removing "unmanaged" because that's not the only purpose of a component's dispose method.

PR #32177, opened 2024-03-27
Update current version and latest version include files

### LINKED ISSUE #32176: Update the not-current-version and not-latest-version include files
Update from saying 7 is current to saying 8 is current

tdykstra:
Fixes #32176


tdykstra:
@Rick-Anderson I think it might be best to combine these two includes into one. Can you think of any reasons not to do that?

Rick-Anderson (approved):

PR #32206, opened 2024-03-29
QR code generation article for BWAs

### LINKED ISSUE #32186: Need blazor specific implementation since the blazor template points to this article
### Description

>Need blazor specific implementation since the blazor template points to this article

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/security/authentication/identity-enable-qrcodes?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/security/authentication/identity-enable-qrcodes.md

### Document ID

270bedc7-a636-5bd0-9e2f-e42235deec8c

### Article author

@Rick-Anderson

guardrex:
Fixes #32186

Mackinnon ... I'm done messing around with it now. It's ready for a look :eyes:.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/startup.md](https://github.com/dotnet/AspNetCore.Docs/blob/7a56eba818c31fc223962715585bc6fd233e3fee/aspnetcore/blazor/fundamentals/startup.md) | [ASP.NET Core Blazor startup](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/startup?branch=pr-en-us-32206) |
| [aspnetcore/blazor/javascript-interoperability/static-server-rendering.md](https://github.com/dotnet/AspNetCore.Docs/blob/7a56eba818c31fc223962715585bc6fd233e3fee/aspnetcore/blazor/javascript-interoperability/static-server-rendering.md) | [ASP.NET Core Blazor JavaScript with static server-side rendering (static SSR)](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/static-server-rendering?branch=pr-en-us-32206) |
| [aspnetcore/blazor/security/server/qrcodes-for-authenticator-apps.md](https://github.com/dotnet/AspNetCore.Docs/blob/7a56eba818c31fc223962715585bc6fd233e3fee/aspnetcore/blazor/security/server/qrcodes-for-authenticator-apps.md) | [aspnetcore/blazor/security/server/qrcodes-for-authenticator-apps](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/server/qrcodes-for-authenticator-apps?branch=pr-en-us-32206) |
| [aspnetcore/security/authentication/identity-enable-qrcodes.md](https://github.com/dotnet/AspNetCore.Docs/blob/7a56eba818c31fc223962715585bc6fd233e3fee/aspnetcore/security/authentication/identity-enable-qrcodes.md) | [Enable QR code generation for TOTP authenticator apps in ASP.NET Core](https://review.learn.microsoft.com/en-us/aspnet/core/security/authentication/identity-enable-qrcodes?branch=pr-en-us-32206) |

### DIFF
--- a/aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
+++ b/aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
@@ -65,89 +65,7 @@ In a [Razor Class Library (RCL)](xref:blazor/components/class-libraries) (the ex
 
 `wwwroot/BlazorPageScript.lib.module.js`:
 
-```javascript
-const pageScriptInfoBySrc = new Map();
-
-function registerPageScriptElement(src) {
-  if (!src) {
-    throw new Error('Must provide a non-empty value for the "src" attribute.');
-  }
-
-  let pageScriptInfo = pageScriptInfoBySrc.get(src);
-
-  if (pageScriptInfo) {
-    pageScriptInfo.referenceCount++;
-  } else {
-    pageScriptInfo = { referenceCount: 1, module: null };
-    pageScriptInfoBySrc.set(src, pageScriptInfo);
-    initializePageScriptModule(src, pageScriptInfo);
-  }
-}
-
-function unregisterPageScriptElement(src) {
-    if (!src) {
-        return;
-    }
-
-    const pageScriptInfo = pageScriptInfoBySrc.get(src);
-    if (!pageScriptInfo) {
-        return;
-    }
-
-    pageScriptInfo.referenceCount--;
-}
-
-async function initializePageScriptModule(src, pageScriptInfo) {
-  if (src.startsWith("./")) {
-    src = new URL(src.substr(2), document.baseURI).toString();
-  }
-
-  const module = await import(src);
-
-  if (pageScriptInfo.referenceCount <= 0) {
-    return;
-  }
-
-  pageScriptInfo.module = module;
-  module.onLoad?.();
-  module.onUpdate?.();
-}
-
-function onEnhancedLoad() {
-  for (const [src, { module, referenceCount }] of pageScriptInfoBySrc) {
-    if (referenceCount <= 0) {
-      module?.onDispose?.();
-      pageScriptInfoBySrc.delete(src);
-    }
-  }
-
-  for (const { module } of pageScriptInfoBySrc.values()) {
-    module?.onUpdate?.();
-  }
-}
-
-export function afterWebStarted(blazor) {
-  customElements.define('page-script', class extends HTMLElement {
-    static observedAttributes = ['src'];
-
-    attributeChangedCallback(name, oldValue, newValue) {
-      if (name !== 'src') {
-        return;
-      }
-
-      this.src = newValue;
-      unregisterPageScriptElement(oldValue);
-      registerPageScriptElement(newValue);
-    }
-
-    disconnectedCallback() {
-      unregisterPageScriptElement(this.src);
-    }
-  });
-
-  blazor.addEventListener('enhancedload', onEnhancedLoad);
-}
-```
+[!INCLUDE[](~/blazor/includes/js-interop/blazor-page-script.md)]
 
 In the RCL, add the following `PageScript` component.
 
@@ -165,11 +83,7 @@ In the RCL, add the following `PageScript` component.
 
 The `PageScript` component functions normally on the top-level of a page.
 
-If you place the `PageScript` component in an app's layout (for example, `MainLayout.razor`), which results in a shared `PageScript` among pages that use the layout, then the component only runs `onLoad` after a full page reload and `onUpdate` when any enhanced page update occurs, including enhanced navigation:
-
-```razor
-<PageScript Src="./Components/Pages/PageWithScript.razor.js" />
-```
+If you place the `PageScript` component in an app's layout (for example, `MainLayout.razor`), which results in a shared `PageScript` among pages that use the layout, then the component only runs `onLoad` after a full page reload and `onUpdate` when any enhanced page update occurs, including enhanced navigation.
 
 To reuse the same module among pages, but have the `onLoad` and `onDispose` callbacks invoked on each page change, append a query string to the end of the script so that it's recognized as a different module. An app could adopt the convention of using the component's name as the query string value. In the following example, the query string is "`counter`" because this `PageScript` component reference is placed in a `Counter` component. This is merely a suggestion, and you can use whatever query string scheme that you prefer.
 
--- a/aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
+++ b/aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
@@ -24,7 +24,7 @@ To avoid this problem, we don't recommended relying on page-specific `<script>` 
 
 The following example demonstrates one way to configure JS code to run when a statically-rendered page with enhanced navigation is initially loaded or updated.
 
-The following `PageWithScript` component example is a component in the app that requires scripts to run with static SSR and enhanced navigation. The following component example includes a `PageScript` component from a Razor class library (RCL) that is added to the solution later in this article.
+The following `PageWithScript` component example is a component in the app that requires scripts to run with static SSR and enhanced navigation. The following component example includes a `PageScript` component from a Razor class library (RCL) that's added to the solution later in this article.
 
 `Components/Pages/PageWithScript.razor`:
 
@@ -165,7 +165,7 @@ In the RCL, add the following `PageScript` component.
 
 The `PageScript` component functions normally on the top-level of a page.
 
-If you place the `PageScript` component in the app's layout (for example, `Components/Layout/MainLayout.razor`), which results in a shared `PageScript` among pages that use the layout, then the component only runs `onLoad` after a full page reload and `onUpdate` when any enhanced page update occurs, including enhanced navigation:
+If you place the `PageScript` component in an app's layout (for example, `MainLayout.razor`), which results in a shared `PageScript` among pages that use the layout, then the component only runs `onLoad` after a full page reload and `onUpdate` when any enhanced page update occurs, including enhanced navigation:
 
 ```razor
 <PageScript Src="./Components/Pages/PageWithScript.razor.js" />
--- a/aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
+++ b/aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
@@ -24,7 +24,7 @@ To avoid this problem, we don't recommended relying on page-specific `<script>` 
 
 The following example demonstrates one way to configure JS code to run when a statically-rendered page with enhanced navigation is initially loaded or updated.
 
-Add the following `PageWithScript` component.
+The following `PageWithScript` component example is a component in the app that requires scripts to run with static SSR and enhanced navigation. The following component example includes a `PageScript` component from a Razor class library (RCL) that is added to the solution later in this article.
 
 `Components/Pages/PageWithScript.razor`:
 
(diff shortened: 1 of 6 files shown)


aspnetcore/blazor/security/server/qrcodes-for-authenticator-apps.md
```diff
@@ -0,0 +1,224 @@
...
+```
+
+The `EnableAuthenticator` component, which is part of the QR code system in the app and displays the QR code to users, adopts static server-side rendering (static SSR) with enhanced navigation. Therefore, normal scripts can't execute when the component loads or updates under enhanced navigation. Extra steps are required to trigger the QR code to load in the UI when the page is loaded. To accomplish loading the QR code, the approach explained in <xref:blazor/js-interop/ssr> is adopted.
+
+Add the following [JavaScript initializer](xref:blazor/fundamentals/startup#javascript-initializers) to the server project's `wwwroot` folder. The `{NAME}` placeholder must be the name of the app's assembly in order for Blazor to locate and load the file automatically. If the server app's assembly name is `BlazorSample`, the file is named `BlazorSample.lib.module.js`.
+
+`wwwroot/{NAME}.lib.module.js`:
+
+```javascript
+const pageScriptInfoBySrc = new Map();
```
MackinnonBuck, line 52: To keep this section shorter, would we consider the one of the following alternatives?

1. Link to the article about setting up `PageScript` without including all that content here, or
2. Use an approach similar to the one you sent me via email, but rather than using the "load" event, use Blazor's `enhancedload` event, which runs after every enhanced navigation:
    ```js
    Blazor.addEventListener('enhancedload', () => { /* Generate the QR code if the target element exists */ });
    ```

I would probably prefer option [2] if possible because it requires less machinery than going all-out with the `<PageScript>` component.

By the way - I created https://github.com/MackinnonBuck/blazor-page-script as a convenient way to add `<PageScript>` to an app so that customers don't have to copy and paste code. I'm not sure if we ever point to non-Microsoft projects from the docs, but if we do, it might be worth mentioning that it's an easier alternative to copying all the code into your project (at the cost of not being able to customize its implementation).

aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
```diff
@@ -165,7 +165,11 @@ In the RCL, add the following `PageScript` component.
 
 The `PageScript` component functions normally on the top-level of a page.
 
-If you place the `PageScript` component in the app's layout (for example, `Components/Layout/MainLayout.razor`), which results in a shared `PageScript` among pages that use the layout, then the component only runs `onLoad` after a full page reload and `onUpdate` when any enhanced page update occurs, including enhanced navigation.
+If you place the `PageScript` component in an app's layout (for example, `MainLayout.razor`), which results in a shared `PageScript` among pages that use the layout, then the component only runs `onLoad` after a full page reload and `onUpdate` when any enhanced page update occurs, including enhanced navigation:
+
+```razor
+<PageScript Src="./Components/Pages/PageWithScript.razor.js" />
```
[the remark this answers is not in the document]
guardrex: Sure ... do you have a suggestion?

aspnetcore/blazor/security/server/qrcodes-for-authenticator-apps.md
```diff
@@ -0,0 +1,224 @@
...
+ms.date: 04/01/2024
+uid: blazor/security/server/qrcodes-for-authenticator-apps
+---
+# Enable QR code generation for TOTP authenticator apps in an ASP.NET Core Blazor Web App
+
+This article explains how to configure an ASP.NET Core Blazor Web App with QR code generation for TOTP authenticator apps.
+
+For an introduction to two-factor authentication (2FA) with authenticator apps using a Time-based One-time Password Algorithm (TOTP), see <xref:security/authentication/identity-enable-qrcodes>.
+
+## Scaffold the Enable Authenticator component into the app
```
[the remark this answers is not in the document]
guardrex: That's reasonable. I'll see about taking that course of action.

aspnetcore/blazor/security/server/qrcodes-for-authenticator-apps.md
guardrex: I'll refactor for `enhancedload`. I like that, too, because the JS node article doesn't have a working example of that approach.

On linking the external repo ... yes ... we do take that approach sometimes. I'm not sure about doing it in the JS article. I'd leave it up to you, Dan, and Artak to decide if that's what you guys want. Also, I'd work that as a separate lower-priority issue. I'm still swamped with higher priority work, and that's not super pressing at the moment, right?

guardrex: BTW ... I'm not 100% clear on the `blazor.addEventListener` approach. Where does that go? ... not collocated in a module? If not, that's why you're saying to check if the element exists? Don't devs consider that a code smell :poop: to check constantly for the existence for an element, when the element only appears on one or a few pages?

aspnetcore/blazor/security/server/qrcodes-for-authenticator-apps.md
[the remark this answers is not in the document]
guardrex: Yes! We do still have a bit on that, it's just cut down from what we used to say there. I'll cross-link to it here ...

https://learn.microsoft.com/en-us/aspnet/core/security/authentication/scaffold-identity?view=aspnetcore-8.0&tabs=visual-studio#client-side-blazor-apps

aspnetcore/blazor/security/server/qrcodes-for-authenticator-apps.md
MackinnonBuck: It could go either in `afterWebStarted` or a `<script>` element just after the one for `blazor.web.js`. Yeah, it could be considered a code smell for the logic to run more often than it needs to, but the alternative is a more sophisticated approach like `PageScript` that uses additional logic to only activate logic that's specific to the current page. I suppose it's somewhat of a balance between having a simpler example vs one that cleanly separates different concerns.

guardrex: Well ... if it were me, I'd go with the `PageScript` approach in my own stuff. Are you sure I shouldn't just link over to that article, dropping the `PageScript` here. After all, if they're cool with the `addEventListener` approach, they can go that way based on the content of the JS interop article (although I admit we don't have an actual fully working example there).

aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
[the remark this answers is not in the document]
MackinnonBuck: Thinking about this more, I'm not sure if having the code snippet adds much value. Maybe it's fine to remove it? We could alternatively come up with a name for a script containing "global" JS logic but maybe that raises more questions than it answers (where is the script defined, what does it do?).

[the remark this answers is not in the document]
guardrex: > the code snippet ... remove it?

I'm not sure which one you're referring to. They all look required to me given the text that's here.

aspnetcore/blazor/security/server/qrcodes-for-authenticator-apps.md
MackinnonBuck: > Are you sure I shouldn't just link over to that article (the JS interop one), dropping the PageScript here (from the QR code article).

Sure, that sounds fine to me! If you think it would help, maybe it would be worth mentioning (in just a sentence, maybe) that using `Blazor.addEventListener('enhancedload', ...)` is a viable alternative approach.

guardrex: Very well ... I'll do that and mention it in passing here.

aspnetcore/blazor/javascript-interoperability/static-server-rendering.md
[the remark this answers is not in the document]
MackinnonBuck: This one:

> If you place the `PageScript` component in the app's layout...
> ```razor
> <PageScript Src="./Components/Pages/PageWithScript.razor.js" />
> ```

The paragraph above the snippet talks about placing a `<PageScript>` in the app's layout, but then the snippet shows an unremarkable usage of `<PageScript>` without context showing where it's being rendered from. I'm guessing it's meant to be assumed that the `<PageScript>` in the snippet is placed somewhere in the layout (even though it's not explicitly shown in the snippet), but if that's the case, then what does the snippet show that the reader couldn't have already inferred?

Or maybe I'm just totally misinterpreting this section. Was the snippet intended to refer to the paragraph that specifically mentions placing the `<PageScript>` in the layout? Or is it supposed to show a general usage of `<PageScript>`? If it's the latter, that makes a lot more sense to me, although I'd maybe suggest moving the code snippet to be _above_ the paragraph that explains the nuance about placing a `<PageScript>` in the layout. And then we wouldn't need to rename the JS file because now the example clearly demonstrates the most common usage, which is to have `<PageScript>` be rendered from a page component.

Let me know if I'm still missing something! I think I was just wrongly associating the snippet with the comment about placing the component in the app's layout.

aspnetcore/blazor/security/server/qrcodes-for-authenticator-apps.md
guardrex: I was able to add the language on `Blazor.addEventListener('enhancedload', ...)` easily enough, but bisecting this guidance doesn't work well. I don't have a clean way to only refer readers to the JS initializer code in the middle of a sequence of steps that pertains to adding it to an RCL from a sequence of steps that focuses on adding it directly to the BWA. It would also be unnatural to introduce artificial sections to the JS interop article merely for this purpose. I'm not favorable to cross-linking to a separate code file that would open in a new browser window either. It's just not good user experience either way. The best thing for readers is just to get the code they need when they need it working straight down the steps.

However, I'll place the code into an INCLUDE file and link it from there. That way, we can change the code in one spot and update article dates to draw it into both articles.

aspnetcore/blazor/security/server/qrcodes-for-authenticator-apps.md
```diff
@@ -0,0 +1,143 @@
...
+```javascript
+export function onLoad() {
+  const uri = document.getElementById('qrCodeData').getAttribute('data-url');
+  new QRCode(document.getElementById('qrCode'), uri);
+}
+
+export function onUpdate() { }
+
+export function onDispose() { }
+```
```
MackinnonBuck, line 76: Nit: It's not necessary to include the `onUpdate()` and `onDispose()` callbacks if they don't do anything :) The code will still work.

aspnetcore/blazor/security/server/qrcodes-for-authenticator-apps.md
```diff
@@ -0,0 +1,143 @@
...
+
+Under the `<PageTitle>` component in the `EnableAuthenticator` component, add the `PageScript` component with the path to the collocated JS file:
+
+```razor
+<PageScript Src="./Components/Account/Pages/Manage/EnableAuthenticator.razor.js" />
+```
+
+> [!NOTE]
+> An alternative to using the approach with the `PageScript` component is to use an event listener (`Blazor.addEventListener("enhancedload", {CALLBACK})`) registered in an [`afterWebStarted` JS initializer](xref:blazor/fundamentals/startup#javascript-initializers) to listen for page updates caused by enhanced navigation. The callback (`{CALLBACK}` placeholder) performs the QR code initialization logic.
```
MackinnonBuck, line 85:
```suggestion
> An alternative to using the approach with the `PageScript` component is to use an event listener (`blazor.addEventListener("enhancedload", {CALLBACK})`) registered in an [`afterWebStarted` JS initializer](xref:blazor/fundamentals/startup#javascript-initializers) to listen for page updates caused by enhanced navigation. The callback (`{CALLBACK}` placeholder) performs the QR code initialization logic.
```

(because if this is happening in the `afterWebStarted` callback, then `blazor` is passed as an argument, so it would be preferable to use the argument instead of the `window.Blazor` object)

MackinnonBuck (approved):
Looks great!

PR #32285, opened 2024-04-09
Improve `SetParametersAsync` base method coverage

### LINKED ISSUE #32273: Clearly describe what happens when a call to `base.SetParametersAsync` is skipped
### Description

In the `SetParametersAsync` section, it is mentioned that the default implementation of `SetParametersAsync` and the ability to skip its call are as follows:

```
If [base.SetParametersAsync](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.components.componentbase.setparametersasync) isn't invoked, developer code can interpret the incoming parameters' values in any way required. For example, there's no requirement to assign the incoming parameters to the properties of the class.
```

However, if the call to `base.SetParametersAsync` is omitted, other lifecycle methods such as `OnInitialized` will not be called, so it is usually necessary to call `base.SetParametersAsync`. Therefore, I believe it is necessary to describe in more detail what happens when `base.SetParametersAsync` is not called.

### Page URL

https://learn.microsoft.com/ja-jp/aspnet/core/blazor/components/lifecycle?view=aspnetcore-8.0#component-disposal-with-idisposable-and-iasyncdisposable

### Content source URL

### Document ID

### Article author

@guardrex

guardrex:
Fixes #32273

Thanks @runceel! 🎸 ... Mackinnon will be along in a bit to take a look 👀.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/lifecycle.md](https://github.com/dotnet/AspNetCore.Docs/blob/719e6ee2964dac244d936119d82a6b417a3fb574/aspnetcore/blazor/components/lifecycle.md) | [ASP.NET Core Razor component lifecycle](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?branch=pr-en-us-32285) |

### DIFF
--- a/aspnetcore/blazor/components/lifecycle.md
+++ b/aspnetcore/blazor/components/lifecycle.md
@@ -78,7 +78,7 @@ If you choose not to call the base class method, you must call your own [compone
 
 [!INCLUDE[](~/includes/aspnetcore-repo-ref-source-links.md)]
 
-If you want to rely on the initialization and rendering API of <xref:Microsoft.AspNetCore.Components.ComponentBase.SetParametersAsync%2A?displayProperty=nameWithType> but not process incoming parameters, you have the option of passing an empty <xref:Microsoft.AspNetCore.Components.ParameterView> to the base method:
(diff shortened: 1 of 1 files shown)


guardrex:
Mackinnon ... I'm going to go ahead with this one as well before I bug out for the weekend. If you spot anything wrong, I'll issue a patch PR.

PR #32318, opened 2024-04-15
Revise setting parameters lifecycle remarks

### LINKED ISSUE #32314: Add another item to consider when not calling `SetParametersAsync` base class method
### Description

From @hakenr ...

> The `ComponentBase.SetParametersAsync()` method is also responsible for invoking `OnParametersSet[Async]()`. Should we mention this similarly to how we discuss `OnInitialized[Async]` here?

I agree, and I'll fix it in the morning. Thanks for the note @hakenr! 🎸 

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?view=aspnetcore-8.0

### Content source URL

### Document ID

### Article author

@guardrex

guardrex:
Fixes #32314

Thanks @hakenr! 🌴 

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/lifecycle.md](https://github.com/dotnet/AspNetCore.Docs/blob/5dad3516b78c1e8a450d2a9219fea254cb0de6d4/aspnetcore/blazor/components/lifecycle.md) | [ASP.NET Core Razor component lifecycle](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?branch=pr-en-us-32318) |


hakenr:
The paragraph still somehow simplifies what `SetParametersAsync` does (possible multiple calls of `StateHasChanged` and rendering), but introduces the issues one has to handle when not calling `base.SetParametersAsync()`.

Maybe a reference to `SetParametersAsync()` source code might be helpful at this place, but TBH I'm not sure whether it is a good idea to bring the "not calling `base.SetParametersAsync(..)` et al" as acceptable option. Most of the `ComponentBase` logic goes away and the resulting component will be "very special" :-D as it breaks the _Liskov substitution principle_ (the _Children classes inherit parent's behaviours._ requirement).

guardrex:
We can surface the 🐉🐉🐉 when avoiding the call in a simplified manner like the following (tacked on to the preceding paragraph where this concept is introduced) ...

> If \<xref:Microsoft.AspNetCore.Components.ComponentBase.SetParametersAsync%2A?displayProperty=nameWithType> isn't invoked with \`base.SetParametersAsync();\`, developer code can interpret the incoming parameters' values in any way required. For example, there's no requirement to assign the incoming parameters to the properties of the class. Refer to the \[\`ComponentBase\` reference source](https://github.com/dotnet/aspnetcore/blob/main/src/Components/Components/src/ComponentBase.cs) when structuring your code without calling \`base.SetParametersAsync();\` because the base class method calls other lifecycle methods and triggers rendering in a complex fashion.

Let's see what Mackinnon thinks. I'll ping him to take a look.

hakenr:
Not calling `base.SetParametersAsync()` is close to implementing your own `IComponent`
https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.components.icomponent?view=aspnetcore-8.0

Definitely an advanced concept which could be covered in other/separate topic (similar to https://learn.microsoft.com/en-us/aspnet/core/blazor/advanced-scenarios.

guardrex:
@hakenr ... We won't get a review on this one after all. The PU is toooooo busy ⛏️⛰️😅. I'd like to proceed and merge the PR. Do you have any final thoughts to consider. I think the revisions here make it clear that this is an advanced subject, that the dev will need to get into the weeds in the ref source, and that we don't generally recommend avoiding the base class method call.

guardrex:
I guess you're busy at the moment. I'm going to go ahead. I need to get this merged before taking off for the weekend.

hakenr (commented):
lgtm

The final state effectively emphasizes the importance of `base.SetParametersAsync(...)` and the associated logic.

PR #32332, opened 2024-04-16
Improve lifecycle method remarks

### LINKED ISSUE #31851: Route parameter processing difference between init and param set lifecycle methods
### Description

In case we want to follow-up on ...

**OptionalParameter.razor - misleading example for setting default parameter value (OnInitialized vs. OnParametersSet)**
#31845 

... with an update to the *Lifecycle* article, this issue will be worked after the main problems are resolved on #31845.

The *Lifecycle* article will probably benefit from any of the following actions ...

* A low-hanging-fruit 🍎 cross-link to the updated guidance.
* Remark in either or both of the init and param set sections.
* Add a short section on the difference between rout param behavior between init and param set methods immediately after the param set section.

We'll circle back to this later, so I'm going to mark it a **P2 - Medium** priority.

cc: @hakenr

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?view=aspnetcore-8.0

### Content source URL

### Document ID

### Article author

@guardrex

guardrex:
Fixes  #31851

Thanks @hakenr! 🏉 ... Let's start with this draft and see where it goes. I think placing an introductory paragraph for init makes sense with the guidance that you suggested. For param set, adding a paragraph with your language works. I don't think that we need a dedicated section on these concepts, but that's still TBD based on where this goes.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/lifecycle.md](https://github.com/dotnet/AspNetCore.Docs/blob/9ed910e03a87bba94d254006ffe1a6e002f156aa/aspnetcore/blazor/components/lifecycle.md) | [ASP.NET Core Razor component lifecycle](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?branch=pr-en-us-32332) |


hakenr (approved):
lgtm

PR #32335, opened 2024-04-16
[Blazor] js-collocation - remove `Pages` from destination folder

hakenr:
`Pages` is part of the `{Path}` in specification of the target folder.
(If the source component is not placed inside `Pages` folder, there is no such folder present in destination path.)

cc @guardrex


guardrex (approved):
# 👍

PR #32336, opened 2024-04-16
[Blazor] js-collocation - remove {EXTENSION} placeholder

hakenr:
The only possible value of `{EXTENSION}` placeholder is `razor`.
(Removing the placeholder also better aligns with the next paragraph.)

cc @guardrex


guardrex (approved):
# 👍

PR #32367, opened 2024-04-22
correct snippet for minimal api result

### LINKED ISSUE #32363: Wrong snippet in "TypedResults vs Results"
### Description

I think the first snippet in the "TypedResults vs Results" section doesn't match the snippet, instead it shows the same code as for the `TypedResults` example.

To fix this the first snippet should show the following code using `Results` as return value:

https://github.com/dotnet/AspNetCore.Docs/blob/90ed3d1bf188f066df900e542394b6d07875363d/aspnetcore/tutorials/min-web-api/samples/7.x/todo/Program.cs#L95-L99

If accepted I can update the page.

Screenshot of the snippets:
image

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/responses?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/minimal-apis/responses.md

### Document ID

f0e5da68-cd54-ce74-1ff3-ebd7d9fd09a4

### Article author

@brunolins16

timdeschryver:
Fixes #32363

- commit 8ec3552e990beee23181ad987a1ef3e5c3d0e9cb: updates the snippet for 7.x and 8.x 
- commit 9e9571712ce083eb1e1a39dee3b3af73587bd36f: updates the examples to use the same route from `todo` to `todoitems`, which were used in preceding examples. I can revert this if needed.

image


aspnetcore/tutorials/min-web-api/samples/8.x/todo/Program.cs
```diff
@@ -92,14 +92,16 @@ await db.Todos.FindAsync(id)
...
 app.MapGet("/todoitems/{id}", async (int id, TodoDb db) =>
     await db.Todos.FindAsync(id)
         is Todo todo
             ? Results.Ok(todo)
             : Results.NotFound());
+// </snippet_1a>
 
 // <snippet_1b>
- app.MapGet("/todos/{id}", async Task<Results<Ok<Todo>, NotFound>> (int id, TodoDb db) =>
+ app.MapGet("/todoitems/{id}", async Task<Results<Ok<Todo>, NotFound>> (int id, TodoDb db) =>
```
wadepickett, line 104: This would result in a route conflict build warning for  `/todoitems/{id}` since there are two of them now.  I think the original author worked around that problem for the sake of keeping it all in one place by commenting out the duplicate route snippet below.  You might have to do the same for this one.   (Same for version 7.0)

wadepickett (changes requested):

wadepickett:
Thanks @timdeschryver, greatly appreciated!   See my note on the route conflict.  

As a side note: There are two documents that use this same app sample, the other is a tutorial. I checked through that one as well and your change would not effect that one for this particular snippet.  So all good there.

aspnetcore/tutorials/min-web-api/samples/8.x/todo/Program.cs
timdeschryver: Thanks for catching the build warning Wade.

I've updated both versions by adding a comment around the second code block.

timdeschryver:
I noticed by renaming the route from `todo` to `todoitems` that the snippet has a weird alignment (see screenshot). Is this something that only happens locally and will this be fixed on the live version, or do I have to do something about this?

image

wadepickett:
No worries for the alignment.  It depends on how wide the browser page is at at some point starts to scroll instead:

Wide page:
image

Scrolling kicks in:
image

wadepickett (approved):
Looks good.  Thanks again!

PR #32447, opened 2024-04-28
Update Startup2.cs

Rick-Anderson:


wadepickett:
This came Saturday and I was out, I have some time to look at these Sunday morning though.

We already had env.IsDevelovelopment, plus " only in a development environment."  Maybe there was some other doc only using just that line so we needed another?
   
```csharp      
 if (env.IsDevelopment())
            {
                // Enable middleware to serve generated Swagger as a JSON endpoint, only in a development environment.
                app.UseSwagger();

                // Enable middleware to serve swagger-ui (HTML, JS, CSS, etc.),
                // specifying the Swagger JSON endpoint.
                app.UseSwaggerUI(c =>  // UseSwaggerUI is called only in Development.
                {
```

wadepickett:
..oh I see, you wanted to also explicitly for `UseSwaggerUI()` even so.  Nevermind.  Not awake yet I guess.

PR #32456, opened 2024-04-29
Add named and typed clients section

peter-csala:
This PR is basically a follow up of my other PR against the [dotnet/docs](https://github.com/dotnet/docs/pull/40359) repository.

It extends the [Make HTTP requests using IHttpClientFactory in ASP.NET Core](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/http-requests) page with named and typed client capability.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/fundamentals/http-requests.md](https://github.com/dotnet/AspNetCore.Docs/blob/325da1b054601408ffb04b991d6f5e9fec6635e9/aspnetcore/fundamentals/http-requests.md) | [Make HTTP requests using IHttpClientFactory in ASP.NET Core](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?branch=pr-en-us-32456) |

### DIFF
--- a/aspnetcore/fundamentals/http-requests.md
+++ b/aspnetcore/fundamentals/http-requests.md
@@ -30,6 +30,7 @@ There are several ways `IHttpClientFactory` can be used in an app:
 * [Basic usage](#basic-usage)
 * [Named clients](#named-clients)
 * [Typed clients](#typed-clients)
+* [Named and typed clients](#named-and-typed-clients)
 * [Generated clients](#generated-clients)
 
 The best approach depends upon the app's requirements.
@@ -113,6 +114,38 @@ The typed client can be injected and consumed directly:
 The configuration for a typed client can also be specified during its registration in `Program.cs`, rather than in the typed client's constructor:
 
 :::code language="csharp" source="http-requests/samples/6.x/HttpRequestsSample/Snippets/Program.cs" id="snippet_AddHttpClientTypedInline":::
+
+### Named and typed clients
+
+Named clients and typed clients have their own strengths and weaknesses. There is a way to combine these two client types to get the best of both worlds.
+
+The primary use case is the following: You want to register the exact same type client multiple times but with different settings (different base address, timeout, and credentials for example).
+
+The registration of the named and typed clients:
+
+:::code language="csharp" source="http-requests/samples/6.x/HttpRequestsSample/Snippets/Program.cs" id="snippet_AddHttpClientTypedInline":::
+
+In the preceding code: 
+
+* Two `GitHubService` typed clients were registered with different timeouts. 
+* The `Timeout` was used as an example for the sake of simplicity, in a real-world scenario the two underlying `HttpClient` could be setup in a very different way.
+
+The usage of the named and typed clients:
+
+:::code language="csharp" source="http-requests/samples/6.x/HttpRequestsSample/Pages/Consumption/NamedAndTypedClient.cshtml.cs" id="snippet_Class" highlight="4,6-7,19-20,23,25-26":::
+
+In the preceding code: 
+
+* The `NamedAndTypedClientModel`'s constructor received two parameters: an `IHttpClientFactory` and an `ITypedHttpClientFactory<GitHubService>`.
+  * The former type resides inside the `System.Net.Http` namespaces whereas the latter type inside the `Microsoft.Extensions.Http`.
+  * The `ITypedHttpClientFactory`'s type parameter was the implementation class (in the preceding example the `GitHubService`). 
+    * Even if you have an abstraction (like `IGitHubService` interface) as well, the type parameter must be the implementation type. 
+    * If you accidentally use the abstraction (`GitHubService`) then it will throw an `InvalidOperationException` when you call its `CreateClient`.
+* Inside the `OnGet` a `Quick` named and typed client will be created for the happy path. 
+  * First the `IHttpClientFactory` is used to create a `Quick` named client.
+  * Then the `ITypedHttpClientFactory`'s `CreateClient` is called with the previous named client to wrap it into a named and typed client.
+* If the request times out (takes longer than 1 second) then the `HttpClient` throws a `TaskCanceledException` with a `TimeoutException` inner.
+  * In this case, a `LatencyTolerant` named and typed client will be created.
 
 ### Generated clients
 
--- a/aspnetcore/fundamentals/http-requests/samples/6.x/HttpRequestsSample/Pages/Consumption/NamedAndTypedClient.cshtml.cs
+++ b/aspnetcore/fundamentals/http-requests/samples/6.x/HttpRequestsSample/Pages/Consumption/NamedAndTypedClient.cshtml.cs
@@ -1,0 +1,38 @@
+using HttpRequestsSample.GitHub;
+using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.Extensions.Http;
+
+namespace HttpRequestsSample.Pages;
+
+// <snippet_Class>
+public class NamedAndTypedClientModel : PageModel
+{
+    private readonly IHttpClientFactory _httpClientFactory;
+    private readonly ITypedHttpClientFactory<GitHubService> _gitHubServiceFactory;
+
+    public NamedAndTypedClientModel(IHttpClientFactory httpClientFactory, 
+        ITypedHttpClientFactory<GitHubService> gitHubServiceFactory)
+    {
+        _httpClientFactory = httpClientFactory;
+        _gitHubServiceFactory = gitHubServiceFactory;
+    }
+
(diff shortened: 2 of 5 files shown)


Rick-Anderson:
@IEvangelist  I'll do the edit review after you do the technical review.

peter-csala:
@IEvangelist Hi David, can I ask you to please review my PR?

IEvangelist:
I had a look, seems ok - but let's please have @CarnaViire review this for technical accuracy instead.

aspnetcore/fundamentals/http-requests.md
```diff
@@ -114,6 +115,38 @@ The configuration for a typed client can also be specified during its registrati
 
 :::code language="csharp" source="http-requests/samples/6.x/HttpRequestsSample/Snippets/Program.cs" id="snippet_AddHttpClientTypedInline":::
 
+### Named and typed clients
+
+Named clients and typed clients have their own strengths and weaknesses. There is a way to combine these two client types to get the best of both worlds.
+
+The primary use case is the following: You want to register the exact same type client multiple times but with different settings (different base address, timeout, and credentials for example).
```
CarnaViire, line 122: In general, I believe this is the case where you should use just an _ordinary Transient service_ instead of a typed client.

That service could either

1) use IHttpClientFactory and incapsulate multiple named clients inside

<details>

```c#
service.UseNamedClient();
service.UseOtherClient();

/* Output:
Name=NamedClient, BaseAddress=https://some-uri/
Name=OtherNamedClient, BaseAddress=https://other-uri/
*/

class Service(IHttpClientFactory _httpClientFactory /*, ... */)
{
    private void UseHttpClient(string name) => Console.WriteLine($"Name={name}, BaseAddress={_httpClientFactory.CreateClient(name).BaseAddress}");
    public void UseNamedClient() => UseHttpClient("NamedClient");
    public void UseOtherClient() => UseHttpClient("OtherNamedClient");
}
```

</details>

-OR-

2) Have an instance of HttpClient passed as a parameter to the method each time

<details>

```c#
foreach (var name in new[] {"NamedClient", "OtherNamedClient" })
{
    service.UseHttpClient(name, _httpClientFactory.CreateClient(name));
}

/* Output:
Name=NamedClient, BaseAddress=https://some-uri/
Name=OtherNamedClient, BaseAddress=https://other-uri/
*/

class Service(/* ... */)
{
    private void UseHttpClient(string name, HttpClient client) => Console.WriteLine($"Name={name}, BaseAddress={client.BaseAddress}");
}
```

</details>

-OR-

3) Leverage Keyed Services DI Pattern (and it can even be non-transient in such case):

```c#

var services = new ServiceCollection();

services.AddHttpClient("NamedClient", c => c.BaseAddress = new Uri("https://some-uri"));
services.AddHttpClient("OtherNamedClient", c => c.BaseAddress = new Uri("https://other-uri"));

services.AddKeyedSingleton<Service>(KeyedService.AnyKey); // instead of a typed client
services.AddSingleton<TwoNamesDependentService>();

var provider = services.BuildServiceProvider();

var dependentService = provider.GetRequiredService<TwoNamesDependentService>();
dependentService.UseServices();

/* Output:
Name=NamedClient, BaseAddress=https://some-uri/
Name=OtherNamedClient, BaseAddress=https://other-uri/
*/

class Service([ServiceKey] string _name, IHttpClientFactory _httpClientFactory)
{
    private HttpClient Client => _httpClientFactory.CreateClient(_name);
    public void UseHttpClient() => Console.WriteLine($"Name={_name}, BaseAddress={Client.BaseAddress}");
}

class TwoNamesDependentService([FromKeyedServices("NamedClient")] Service _service, [FromKeyedServices("OtherNamedClient")] Service _otherService)
{
    public void UseServices()
    {
        _service.UseHttpClient();
        _otherService.UseHttpClient();
    }
}
```

aspnetcore/fundamentals/http-requests/samples/6.x/HttpRequestsSample/Pages/Consumption/NamedAndTypedClient.cshtml.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        try
+        {
+            var quickClient = _httpClientFactory.CreateClient("Quick");
+            var gitHubService = _gitHubServiceFactory.CreateClient(quickClient);
+            GitHubBranches = await gitHubService.GetAspNetCoreDocsBranchesAsync();
+        }
+        catch (TaskCanceledException ex) when (ex.InnerException is TimeoutException)
+        {
+            var tolerantClient = _httpClientFactory.CreateClient("LatencyTolerant");
```
CarnaViire, line 32: This exact scenario should be handled not by multiple clients, but by resilience strategies applied via Polly

CarnaViire (changes requested):
Thanks for tagging me @IEvangelist!

While I appreciate your contribution @peter-csala I believe this is not a good guidance, see my comment https://github.com/dotnet/AspNetCore.Docs/pull/32456#discussion_r1634728785. If you strongly believe using multiple named clients should be highlighted in the docs, I'd propose to change the guidance to the one I provided in the comment.

I see that https://github.com/dotnet/docs/pull/40359 was merged but I was not aware on it -- @IEvangelist I think we would need to either revert it or substitute it with the guidance from https://github.com/dotnet/AspNetCore.Docs/pull/32456#discussion_r1634728785.

On a side note:
I also keep thinking that we need to make some effort to _reduce duplications_ between the docs -- as HttpClientFactory belongs to BCL now, it would be better to have the conceptual docs there, and only leave ASP.NET Core related stuff here, and redirect to .NET docs otherwise. But this needs to be done carefully, not to disrupt the users flow. Let's discuss it later @IEvangelist

peter-csala:
Thank you @CarnaViire for the review. I agree that my made up example is not the best. I wanted to create a simple example to showcase the usage of named and typed clients because (for some unknown reason) they are not listed on any of the documentation pages. Neither here nor on the dotnet docs.

Do you have any idea what would be an ideal sample scenario for this?

peter-csala:
Thank you Natalia for your "constructive" suggestions...

CarnaViire:
> I agree that my made up example is not the best.

@peter-csala The main problem is not with the _example_, but with the _approach_ in general.

> showcase the usage of named and typed clients because (for some unknown reason) they are not listed on any of the documentation pages

I assume by "the usage of named and typed clients" you mean using several named clients within a service, which you wanted to register and construct via typed clients infrastructure. As I mentioned in my comment, typed clients are simply not designed to be used this way, and `ITypedHttpClientFactory` interface is not intended to be used in the application code, it is public only for the sake of being able to substitute its implementation e.g. in tests. That's the reason behind it "not being listed" -- simply because we don't consider it being a recommended approach.

You can see the following remarks in the [docs](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.http.itypedhttpclientfactory-1?view=net-8.0#remarks):

> _The [ITypedHttpClientFactory<TClient>](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.http.itypedhttpclientfactory-1?view=net-8.0) is infrastructure that supports the [AddHttpClient<TClient>(IServiceCollection, String)](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.httpclientfactoryservicecollectionextensions.addhttpclient?view=net-8.0#microsoft-extensions-dependencyinjection-httpclientfactoryservicecollectionextensions-addhttpclient-1(microsoft-extensions-dependencyinjection-iservicecollection-system-string)) and [AddTypedClient<TClient>(IHttpClientBuilder)](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.httpclientbuilderextensions.addtypedclient?view=net-8.0#microsoft-extensions-dependencyinjection-httpclientbuilderextensions-addtypedclient-1(microsoft-extensions-dependencyinjection-ihttpclientbuilder)) functionality. This type should rarely be used directly in application code, use [GetService(Type)](https://learn.microsoft.com/en-us/dotnet/api/system.iserviceprovider.getservice?view=net-8.0#system-iserviceprovider-getservice(system-type)) instead to retrieve typed clients._

So, if your wish is to use several named clients within a service, I've given you several ways to do that; if for some reason neither of them suit your needs, or there are other problems you are trying to solve, please let me know. If there's a _real-world_ scenario which is _impossible_ or _significantly complicated_ with the existing API shape, we'd love to know that.

You can also check some of the issues (in [dotnet/runtime](https://github.com/dotnet/runtime)) that are already known to us, e.g. this one https://github.com/dotnet/runtime/issues/89755 could further simplify the use of multiple named clients. Please upvote if there's something you need, to help us prioritize. Thanks!

PR #32465, opened 2024-04-30
Clarify conditional HTML element attributes

### LINKED ISSUE #32452: Clarify conditional attributes section
### Description

Following up on a PR that we didn't merge at https://github.com/dotnet/blazor-samples/pull/273 ...

The [article section](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-8.0#conditional-html-element-attributes) states ...

> If the value is `false` or `null`, the property isn't set ...

However, dev tools indicates the property value is `false` either way, unless dev tools is just stating that the value is `false` while under-the-covers it isn't set (I assume `null` == "not set"). Dev tools doesn't state the value is `false` if ...

* The component is opened and the `<input>`'s `checked` property value is examined immediately.
* The code is changed to assign `null` when the button is clicked. The button is clicked, and the `<input>`'s `checked` property value is examined in dev tools.

Also, the original ask on the samples PR is about making the example use a field, and I think that's fine for this example. I plan to add a line or a phrase that makes it clear that the guidance applies to both parameters and fields.

cc: @ali50m ... I'll ask Steve Sanderson for assistance with this during the week.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-8.0#conditional-html-element-attributes

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/index.md

### Document ID

31ad18df-1904-bbdd-cfed-8d0d0267c10e

### Article author

@guardrex

guardrex:
Fixes #32452

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/809342833499560b96a021847a44ffcec97a1c91/aspnetcore/blazor/components/index.md) | [ASP.NET Core Razor components](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/index?branch=pr-en-us-32465) |

### DIFF
--- a/aspnetcore/blazor/components/index.md
+++ b/aspnetcore/blazor/components/index.md
@@ -1357,45 +1357,16 @@ Attributes can be applied to components with the [`@attribute`][7] directive. Th
 @attribute [Authorize]
 ```
 
-## Conditional HTML element attributes
+## Conditional HTML element attributes and DOM properties
 
-HTML element attribute properties are conditionally set based on the .NET field or property value. If the value is `false` or `null`, the property isn't set. If the value is `true`, the property is set.
+The distinction between HTML element attributes and DOM properties is subtle, and complicated rules determine how attributes and properties are related and updated. In some cases, updating an attribute updates the property automatically, while in other cases it doesn't.
 
-In the following example, `isCompleted` determines if the `<input>` element's `checked` property is set.
+Blazor adopts the following general behaviors:
 
-`ConditionalAttribute.razor`:
+* For HTML attributes, Blazor sets or removes the attribute conditionally based on the .NET value. If the .NET value is `false` or `null`, the attribute isn't set or is removed if it was previously set.
+* For DOM properties, such as `checked` or `value`, Blazor sets the DOM property based on the .NET value. If the .NET value is `false` or `null`, the DOM property is reset to a default value.
 
-:::moniker range=">= aspnetcore-8.0"
-
-:::code language="razor" source="~/../blazor-samples/8.0/BlazorSample_BlazorWebApp/Components/Pages/ConditionalAttribute.razor":::
-
-:::moniker-end
-
-:::moniker range=">= aspnetcore-7.0 < aspnetcore-8.0"
-
-:::code language="razor" source="~/../blazor-samples/7.0/BlazorSample_WebAssembly/Pages/index/ConditionalAttribute.razor":::
-
-:::moniker-end
-
-:::moniker range=">= aspnetcore-6.0 < aspnetcore-7.0"
-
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/components/index.md
```diff
@@ -1357,45 +1357,16 @@ Attributes can be applied to components with the [`@attribute`][7] directive. Th
 @attribute [Authorize]
 ```
 
-## Conditional HTML element attributes
+## Conditional HTML element attributes and DOM properties
 
-HTML element attribute properties are conditionally set based on the .NET value. If the value is `false` or `null`, the property isn't set. If the value is `true`, the property is set.
+The distinction between HTML element attributes and DOM properties is subtle, and complicated rules determine how attributes and properties are related and updated. In some cases, updating an attribute updates the property automatically, while in other cases it doesn't.
```
SteveSandersonMS, line 1362: I don't personally think it's helpful to make this remark about it being subtle/complicated. It's true that it's complicated, but JS/DOM is an external system that we don't have to make value judgements about. People can have their own opinion about this and we neither need to justify nor document it.

The other content below looks fine though, and makes just as much sense without this statement about complexity.

aspnetcore/blazor/components/index.md
```diff
@@ -1357,45 +1357,16 @@ Attributes can be applied to components with the [`@attribute`][7] directive. Th
...
-
-:::code language="razor" source="~/../blazor-samples/3.1/BlazorSample_WebAssembly/Pages/index/ConditionalAttribute.razor":::
-
-:::moniker-end
-
-For more information, see <xref:mvc/views/razor>.
+Which Razor syntax attributes correspond to HTML attributes and which ones correspond to DOM properties remains undocumented because it's an implementation detail that might change without notice.
 
 > [!WARNING]
 > Some HTML attributes, such as [`aria-pressed`](https://developer.mozilla.org/docs/Web/Accessibility/ARIA/Roles/button_role#Toggle_buttons), don't function properly when the .NET type is a `bool`. In those cases, use a `string` type instead of a `bool`.
```
SteveSandersonMS, line 1370: Saying things "don't function properly" makes it sound like a framework bug. Could we phrase it more like "Some HTML attributes, such as `aria-pressed` explicitly need to have the string values `"true"` and `"false"`. Since they require a string value and not a boolean, you must use a .NET `string` and not a `bool` for their value. This is a requirement set by browser DOM APIs."

SteveSandersonMS (approved):

PR #32482, opened 2024-04-30
Use OIDC keys for the OSPO API

tdykstra:
Copied from https://github.com/dotnet/docs/pull/40674


BillWagner (approved):
This LGTM @tdykstra 

I updated the YML files with the permissions fix from /dotnet/docs#40676. I'd missed that in the reference PR, and we didn't catch it until we merged.

PR #32492, opened 2024-05-01
Fix return type issues in "SignalR API design considerations"

### LINKED ISSUE #32505: when to use async in SignalR code
### Description

Maybe
```
public async Task<string> GetTotalLengthAsync(string param1)
{
    // GetStringAsync is an async method that returns Task<string>
    string result = await GetStringAsync(param1);
    return result.Length.ToString();
}
```

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/signalr/api-design?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/signalr/api-design.md

### Document ID

0f0d321d-46ed-25cb-6bb4-e4611343b131

### Article author

@bradygaster

JonathanBout:
Page: https://learn.microsoft.com/en-us/aspnet/core/signalr/api-design?view=aspnetcore-8.0
- A lot of functions were async and returned `Task` without awaiting anything which is 
- Some functions had a 'void' `Task` as return type but actually returned a value
- Some functions returned `string` while that should be `int`

***EDIT by @Rick-Anderson***
Fixes #32505

### DIFF
--- a/aspnetcore/signalr/api-design/sample/Samples.cs
+++ b/aspnetcore/signalr/api-design/sample/Samples.cs
@@ -1,12 +1,12 @@
 #region ParameterBasedOldVersion
-public async Task<string> GetTotalLength(string param1)
+public int GetTotalLength(string param1)
 {
     return param1.Length;
 }
 #endregion
 
 #region ParameterBasedNewVersion
-public async Task<string> GetTotalLength(string param1, string param2)
+public int GetTotalLength(string param1, string param2)
 {
     return param1.Length + param2.Length;
 }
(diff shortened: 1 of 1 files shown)


Rick-Anderson:
SignalR code is asynchronous to provide maximum scalability. I think it would be better to await.

BrennanConroy:
There is nothing to await in the code being changed. This looks fine.

BrennanConroy:
> * Some functions had a 'void' `Task` as return type but actually returned a value

Where?

Rick-Anderson:
Why not:
```
public async Task<string> GetTotalLengthAsync(string param1)
{
    // GetStringAsync is an async method that returns Task<string>
    string result = await GetStringAsync(param1);
    return result.Length.ToString();
}
```

BrennanConroy:
Because that's just adding a new function without any reason. Returning a quick result synchronously is fine. If we are lacking docs that show results being returned asynchronously then that should be a different change.

PR #32498, opened 2024-05-01
Min Web API: Corrected Put data in body

### LINKED ISSUE #32239: Min Web API: Two problems: ID in PUT body and Put Example returns Status 200 not 204
### Description

Update by wadepickett:  The issue brought up in the discussion of the ID being passed in the PUT body is an issue to address.  The 2nd issue of the wrong status code could not be recreated.

2nd part of the issue noted here after it came up in the discussion, which is to be addressed:

"The other item on that page that should be updated is in the **Test the PUT endpoint** section. 

The example provides the ID in the body of the request but it is not used by the logic:

```
Content-Type: application/json

{
  "id": 1,
  "name": "feed fish",
  "isComplete": false
}

```

The **"id": 1,** line should be removed. The ID is pulled from the URI, not the body of the request. "

---  Original descrption from community member ----

The document states that sending the PUT command will return a 204 status code. Instead it returns a 200. 

image

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/tutorials/min-web-api?view=aspnetcore-8.0&tabs=visual-studio

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/tutorials/min-web-api.md

### Document ID

ab408035-6313-a183-9940-32de00e0e977

### Article author

@wadepickett

---
[Associated WorkItem - 251080](https://dev.azure.com/msft-skilling/Content/_workitems/edit/251080)

wadepickett:
Fixes #32239

Minor change.

Two items were mentioned in the issue, only one turned out to need a fix which was to remove the id from the PUT data in the body.  This fix was not needed for previous versions, just the current v 8.0.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/tutorials/min-web-api.md](https://github.com/dotnet/AspNetCore.Docs/blob/a89bbc5b71292a4baa001b8bdb27af0b10e37d3d/aspnetcore/tutorials/min-web-api.md) | [Tutorial: Create a minimal API with ASP.NET Core](https://review.learn.microsoft.com/en-us/aspnet/core/tutorials/min-web-api?branch=pr-en-us-32498) |

### DIFF
--- a/aspnetcore/tutorials/min-web-api.md
+++ b/aspnetcore/tutorials/min-web-api.md
@@ -3,7 +3,7 @@ title: "Tutorial: Create a minimal API with ASP.NET Core"
 author: wadepickett
 description: Learn how to build a minimal API with ASP.NET Core.
 ms.author: wpickett
-ms.date: 03/11/2024
+ms.date: 05/02/2024
 ms.custom: engagement-fy24
 monikerRange: '>= aspnetcore-6.0'
 uid: tutorials/min-web-api
@@ -526,7 +526,6 @@ Update the to-do item that has `Id = 1` and set its name to `"feed fish"`.
   Content-Type: application/json
 
   {
-    "id": 1,
     "name": "feed fish",
     "isComplete": false
   }


tdykstra:
The issue mentions the status code being wrong, doesn't say anything about the id field. Does the link point to the right issue?

wadepickett:
The ID issue is detailed in the discussion portion of that issue.

"The other item on that page that should be updated is in the Test the PUT endpoint section.

The example provides the ID in the body of the request but it is not used by the logic:

Content-Type: application/json

{
  "id": 1,
  "name": "feed fish",
  "isComplete": false
}

The "id": 1, line should be removed. The ID is pulled from the URI, not the body of the request."

wadepickett:
Sorry, that is odd, I just edited your text rather than creating new text.  I will fix it.

tdykstra:
Thanks for the clarification. Was the part about 200 vs. 204 incorrect, so nothing needed to be done?

tdykstra (approved):

wadepickett:
> Thanks for the clarification. Was the part about 200 vs. 204 incorrect, so nothing needed to be done?

Correct.  I verified following the instruction steps and could not recreate the 200 vs. 204 problem, and in the discussion I provided a screen shots of the results.  Another community member also did the same.

wadepickett:
Sorry for the confusion, I should have also updated the original issue description to make it clear there were two issues brought up, but one was a non-issue, not reproducible.  Thanks.

PR #32511, opened 2024-05-03
Bound field or property expression convention

### LINKED ISSUE #30656: Undocumented {PARAMETER NAME}Expression feature in blazor data binding documentation.
### Type of issue

Missing information

### Description

I came across a hidden feature with data binding that I could not find anywhere in the documentation. If you have a public parameter of type string for example you can add a _{PARAMETER NAME}Expression_ property and this property will be filled with the actual expression of the bound property. I think this detail is missing from the documentation and comes in handy for people who author their own input controls.

```C#
[Parameter] public string? Text { get; set; }

[Parameter] public EventCallback<string?> TextChanged { get; set; }

[Parameter] public Expression<Func<string?>>? TextExpression { get; set; } // undocumented feature
```

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/data-binding?view=aspnetcore-7.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/data-binding.md

### Version Independent Document ID

3f912dfc-5ba5-6c6c-ba93-1ae78508a5cc

### Article author

guardrex

guardrex:
Fixes #30656

cc: @AntMaster7 ... I had a few minutes to whip *something* up, but I'm not necessarily keen on my example and text. I say that this would have an obvious use in creating custom validation components, which isn't something that the Blazor docs here cover. This is a bit of an advanced framework feature for custom scenarios. Can you look this over along with the sample components ...

* https://github.com/dotnet/blazor-samples/blob/main/8.0/BlazorSample_BlazorWebApp/Components/ChildParameterExpression.razor
* https://github.com/dotnet/blazor-samples/blob/main/8.0/BlazorSample_BlazorWebApp/Components/Pages/Parent3.razor

... and see if you think this will at least get it covered until Mackinnon is free. He's buried in work for the next few months, and I don't want to disturb him unless absolutely necessary. I'm going to make a tracking entry to make sure that I discuss this section and example with him later. For now tho, do you think this is about what you would've needed to see for your own purposes back when you went looking for doc coverage on it?

### DIFF
--- a/aspnetcore/blazor/components/data-binding.md
+++ b/aspnetcore/blazor/components/data-binding.md
@@ -892,6 +892,76 @@ In the following `NestedChild` component, the `NestedGrandchild` component:
 
 For an alternative approach suited to sharing data in memory and across components that aren't necessarily nested, see <xref:blazor/state-management>.
 
+## Bound field or property expression tree
+
+A bound field or property's [expression tree](/dotnet/csharp/advanced-topics/expression-trees/) is obtained by adding a property named for the field or property (`{PARAMETER OR FIELD NAME}Expression`).
+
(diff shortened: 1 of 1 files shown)


guardrex:
I'm going to go ahead with this. Mackinnon will look at it later when his schedule opens up a bit.

guardrex:
Thanks! At least, this will provide a little guidance until Mackinnon can look. He may be able to provide better descriptive language.

>  I do not have permission however to open the preview link.

That's normal. It's for internal review by Microsoft folks.

PR #32538, opened 2024-05-07
Update quest.yml to match quest.yml in dotnet/docs

tdykstra:
See https://github.com/dotnet/docs/blob/9293ff9d4430de1b1aa3a40312707f7eb7575876/.github/workflows/quest.yml#L72
Only Quest_bulk.yml has duration at the end of the file; Quest.yml doesn't

### DIFF
--- a/.github/workflows/quest.yml
+++ b/.github/workflows/quest.yml
@@ -76,4 +76,4 @@ jobs:
           org: ${{ github.repository_owner }}
           repo: ${{ github.repository }}
           issue: ${{ github.event.issue.number }}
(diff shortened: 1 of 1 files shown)


BillWagner:
@tdykstra 

This one isn't needed. The `duration` argument enables the bulk workflow to look back more than the previous 5 days. We added that switch in case a PAT expires, and isn't caught quickly enough.

I'll be updating the notes today to reflect the latest configuration changes.

tdykstra:
> This one isn't needed. The `duration` argument enables the bulk workflow to look back more than the previous 5 days.

This PR removes duration from the non-bulk Quest.yml. Sounds like it isn't needed in the non-bulk workflow (?)

Issue #32541, opened 2024-05-08
Register a server API app in Azure is missing a step


1iveowl:
### Description

[Enter feedback here]
For the Server API to show up in the "My APIs" section when adding the "API.Access" in last number list guide of "Register a client app in Azure", the user need to add themself as Owner of the Service API registration. 

See here: https://learn.microsoft.com/en-us/answers/questions/1336925/why-i-cannot-find-my-api-in-the-list-of-apis-in-my

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/hosted-with-microsoft-entra-id?view=aspnetcore-7.0&viewFallbackFrom=aspnetcore-8.0&source=docs

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/security/webassembly/hosted-with-microsoft-entra-id.md

### Document ID

46ac86e5-52ba-d8a5-149d-b2f5eb439bab

### Article author

@guardrex

guardrex:
Thanks @1iveowl ... I'll get this scheduled to include a remark on it in the article. Due to a high workload at this time 😅, it will take about a month to reach this issue. If you want to submit a PR against the article yourself, you're welcome to do so. If not, that's fine. Your issue won't get lost, and I'll work it as soon as I can 🏃‍♂️.

guardrex:
I think it would be wise given how long it will take for me to reach this to place a really QUICK remark there that cross-links the content that you mentioned. That way, at least the article will *surface something* on this in the meantime. I'll do that right now.

PR #32542, opened 2024-05-08
Remark on ownership of service API registration

guardrex:
Addresses #32541

cc: @1iveowl ... This is just to make devs aware of the situation with the new ME-ID requirement. I'll get back to this as soon as I can to place better guidance than this.


guardrex:
I placed two of these IMPORTANT notes:

* I placed the first one in the *Register a server API app in Azure* section where the API is registered.
* I reiterate it with a second note where the API permissions are described.

Issue #32477, opened 2024-04-30
[.NET 9]: Refactor and revise OpenAPI docs


captainsafia:
We're making some significant changes to the OpenAPI area in .NET 9. Namely:

- Adding built-in support for OpenAPI document generation
- Updating the web API templates to take advantage of the built-in support

As part of this, we'll want to refactor the docs on OpenAPI to reflect the state of the new universe. 

AFAICT, we reference OpenAPI in a few pages in the docs:

- [OpenAPI support in minimal APIs](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/openapi?view=aspnetcore-8.0)
- [Swagger/OpenAPI under controller-based APIs](https://learn.microsoft.com/en-us/aspnet/core/tutorials/web-api-help-pages-using-swagger?view=aspnetcore-8.0)
	- Includes "Getting started with NSwag" and "Getting started with Swashbuckle" sub-pages 
- [.NET OpenAPI tool reference](https://learn.microsoft.com/en-us/aspnet/core/web-api/microsoft.dotnet-openapi?view=aspnetcore-8.0)
- [Using OpenAPI with gRPC JSON transcoding](https://learn.microsoft.com/en-us/aspnet/core/grpc/json-transcoding-openapi?view=aspnetcore-8.0)
- [Getting started with Swashbuckle and ASP.NET Core sample](https://learn.microsoft.com/en-us/samples/dotnet/aspnetcore.docs/getstarted-swashbuckle-aspnetcore/?tabs=visual-studio)
-   [Generate a REST API client in a build](https://learn.microsoft.com/en-us/visualstudio/msbuild/tutorial-rest-api-client-msbuild?view=vs-2022)

As part of this work, I'm recommending a revised header structure for our OpenAPI-related docs.

- Controller-based APIs
	- OpenAPI 	
		- Overview
		- Getting started with Microsoft.AspNetCore.OpenApi
		- Getting started with Swashbuckle
		- Getting started with NSwag
		- Working with OpenAPI at build-time
- Minimal APIs
	- OpenAPI 	
		- Overview
		- Getting started with Microsoft.AspNetCore.OpenApi
		- Getting started with Swashbuckle
		- Getting started with NSwag
		- Working with OpenAPI at build-time

The content under each of these pages should be the same, which conditional headers that allows us to show controller and minimal API specific code when necessary.

[...]

PR #32576, opened 2024-05-14
Update Swashbuckle documentation

martincostello:
- Remove unsupported notice for .NET 8.
- Update package version numbers.
- Remove reference to resolved issue.

Resolves domaindrivendev/Swashbuckle.AspNetCore#2857.

See https://github.com/dotnet/AspNetCore.Docs/issues/32477#issuecomment-2099910191.

/cc @captainsafia

### DIFF
--- a/aspnetcore/tutorials/getting-started-with-swashbuckle.md
+++ b/aspnetcore/tutorials/getting-started-with-swashbuckle.md
@@ -5,17 +5,10 @@ description: Learn how to add Swashbuckle to your ASP.NET Core web API project t
 ms.author: scaddie
 monikerRange: '>= aspnetcore-3.1'
 ms.custom: mvc
-ms.date: 04/25/2024
+ms.date: 05/14/2024
 uid: tutorials/get-started-with-swashbuckle
 ---
 # Get started with Swashbuckle and ASP.NET Core
-
-:::moniker range=">= aspnetcore-8.0"
-
-> [!NOTE]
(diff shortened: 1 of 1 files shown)


captainsafia:
Looks good overall!

Do you mind also updating the references to Swashbuckle.AspNetCore versions elsewhere in the docs? I'm thinking specifically about 8.x samples that reference 6.5.x (see [here](https://grep.app/search?q=Swashbuckle.ASpNetCore&filter[repo][0]=dotnet/AspNetCore.Docs))?

martincostello:
Sure - I'll take a look - it didn't occur to me to look elsewhere in the docs 😄

martincostello:
@captainsafia I did a quick look, and most of the remaining references are all package versions in the samples. A lot of those depend on pre-release versions of .NET from when the content was originally authored, so I figured trying to update those too would snowball fast into a whole world of pain 😅

captainsafia (approved):

martincostello:
What we do in _our_ docs is different to what Microsoft do in _their_ docs - I just did the minimal change which is to bump them, rather than question why they specified a version in the first place.

Removing them entirely is a separate conversation.

captainsafia:
> Why even specify package version numbers in the docs? 

We manage docs for multiple versions of .NET in the Microsoft docs page. Providing specific package versions in each docs page allows us to maintain more high quality docs when package versions might differ across TFMs for example. This is especially important for communicating what version of Swashbuckle.AspNetCore works with .NET 8.

PR #32651, opened 2024-05-24
DataProtection Blob warning

### LINKED ISSUE #32193: Storing encryption keys in BLOB and securing them with Azure Key Vault
### Description

I recently run into issues when I used BLOB storage to store encryption key(s). 
Link to that issue is [DataProtectionProvider persist keys to BLOB does not use keys from BLOB](https://github.com/dotnet/aspnetcore/issues/54697).
I was lucky enough that issue resolution was simple enough and I think official documents, should include important things from things discussed in that issue.
Please let me know if you need anything else from me.

Here is the text that should be added to documentation page:
> [!NOTE]
> This article was written for an app that runs within a docker container, where it will always have the same path and, therefore, the same application discriminator. If you need to run an app in multiple environments (e.g. local and deployed), set the default application discriminator for the environment. You can find more details on application discriminator at [SetApplicationName](https://learn.microsoft.com/aspnet/core/security/data-protection/configuration/overview#setapplicationname)
> When BLOB is used to store encryption key then Azure Key Vault MUST be used to protect it, thus two calls, **PersistKeysToAzureBlobStorage** and **ProtectKeysWithAzureKeyVault** cannot be separated.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/security/data-protection/configuration/overview?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/security/data-protection/configuration/overview.md

### Document ID

69818122-e47a-3fed-ea37-8009f66e2a5c

### Article author

@Rick-Anderson

Rick-Anderson:
Fixes #32193
@mjoksa  please review

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/security/data-protection/configuration/overview.md](https://github.com/dotnet/AspNetCore.Docs/blob/bc5c4966a4bd6ca668ec7e90210b290171db4cde/aspnetcore/security/data-protection/configuration/overview.md) | [Configure ASP.NET Core Data Protection](https://review.learn.microsoft.com/en-us/aspnet/core/security/data-protection/configuration/overview?branch=pr-en-us-32651) |

### DIFF
--- a/aspnetcore/security/data-protection/configuration/overview.md
+++ b/aspnetcore/security/data-protection/configuration/overview.md
@@ -25,7 +25,7 @@ For these scenarios, the Data Protection system offers a rich configuration API.
 > The extension method <xref:Microsoft.Extensions.DependencyInjection.DataProtectionServiceCollectionExtensions.AddDataProtection%2A> returns an <xref:Microsoft.AspNetCore.DataProtection.IDataProtectionBuilder>. `IDataProtectionBuilder` exposes extension methods that you can chain together to configure Data Protection options.
 
 > [!NOTE]
-> This article was written for an app that runs within a docker container, where it will always have the same path and, therefore, the same application discriminator. If you need to run an app in multiple environments (e.g. local and deployed), set the default application discriminator for the environment.
+> This article was written for an app that runs within a docker container. In a docker container the app always has the same path and, therefore, the same application discriminator. Apps that need to run in multiple environments (e.g. local and deployed), must set the default application discriminator for the environment.
 > Running an app in multiple environments is beyond the scope of this article.
 
 The following NuGet packages are required for the Data Protection extensions used in this article:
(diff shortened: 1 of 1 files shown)


aspnetcore/security/data-protection/configuration/overview.md
```diff
@@ -24,6 +24,10 @@ For these scenarios, the Data Protection system offers a rich configuration API.
 >
 > The extension method <xref:Microsoft.Extensions.DependencyInjection.DataProtectionServiceCollectionExtensions.AddDataProtection%2A> returns an <xref:Microsoft.AspNetCore.DataProtection.IDataProtectionBuilder>. `IDataProtectionBuilder` exposes extension methods that you can chain together to configure Data Protection options.
 
+> [!NOTE]
+> This article was written for an app that runs within a docker container, where it will always have the same path and, therefore, the same application discriminator. If you need to run an app in multiple environments (e.g. local and deployed), set the default application discriminator for the environment.
```
Rick-Anderson, line 28:
```suggestion
> This article was written for an app that runs within a docker container. In a docker container the app always has the same path and, therefore, the same application discriminator. Apps that need to run in multiple environments (e.g. local and deployed), must set the default application discriminator for the environment.
```

mjoksa:
Everything seems ok :)
One more time, thank you for all your help with this.
Stay safe.

PR #32700, opened 2024-05-30
Fix command to generate a minimal webapi

chyipin:
Use `webapi` instead of `web`.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/tutorials/min-web-api.md](https://github.com/dotnet/AspNetCore.Docs/blob/a2f4d978889c93e6f5123212b926bde413aba432/aspnetcore/tutorials/min-web-api.md) | [Tutorial: Create a minimal API with ASP.NET Core](https://review.learn.microsoft.com/en-us/aspnet/core/tutorials/min-web-api?branch=pr-en-us-32700) |

### DIFF
--- a/aspnetcore/tutorials/min-web-api.md
+++ b/aspnetcore/tutorials/min-web-api.md
@@ -77,7 +77,7 @@ This tutorial creates the following API:
 * Run the following commands:
 
   ```dotnetcli
-  dotnet new web -o TodoApi
+  dotnet new webapi -o TodoApi
   cd TodoApi
   code -r ../TodoApi
   ```


wadepickett (commented):
(Correction added) I think I see what happened here: This tutorial was purposely using an empty project without any example code or control structure.  

For .NET 7 the `webapi` template included controllers and example code.  At the same time the .Net 7 the `web` template was simplified in its project structure to only include minimal necessary components and was used for creating minimal API's going forward and this topic was updated accordingly.

Then for .net 8 the `webapi` template created a minimal API by default without controllers, but it included example code.   

This tutorial for all three versions (VS, VS Code, VS for Mac) starts with an empty minimal template and adds what is needed.  If it used the` webapi` template then starting instructions would be to clear out the existing code or ignore it.  So it uses the `web` template, which is as minimal and empty as it gets.

It is obvious however that using the `web` without an explanation is confusing. 

@tdykstra, thoughts on this?  The options seem to me to be:
1. Rewrite the tutorial for the `webapi` template which includes sample code and just walk through the sample code instead. (fastest way to get your first minimal API working in just a few steps)
2. Keep the tutorial as is, but use the `webapi` template, then either instruct to remove the example code, or ignore it, leave it there, and add just new API's.
3. Keep the tutorial as is, use the `web` template, but explain why that is used here instead of `webapi`, in order to show how to create one from scratch.

My inclination would be to do (3) and I will create a new issue for that if you agree.

tdykstra:
I agree that option 3 is best.

PR #32701, opened 2024-05-30
ID change to Id

MrXhh:
[Add a data model](https://learn.microsoft.com/en-us/aspnet/core/tutorials/razor-pages/model?view=aspnetcore-8.0&tabs=visual-studio#add-a-data-model)


wadepickett (approved):
Thanks.

PR #32743, opened 2024-06-05
First MVC: Validation: Clear old null values

### LINKED ISSUE #32661: [UUF Port]: First MVC: Validation: Method cannot be called on Null value
### Description

Unified User Feedback Port:
Customer Feedback:
"
Tengo un error y no se como corregirlo: SqlNullValueException: Data is Null. This method or property cannot be called on Null values. El erro me indica el siguiente codigo. if (!string. IsNullOrEmpty(movieGenre))            {                movies = movies. Where(x => x.Genre == movieGenre);            }            var movieGenreVM = new MovieGenreViewModel            {                Genres = new SelectList(await genreQuery.Distinct(). ToListAsync()),                Movies = await movies. ToListAsync()            };            return View(movieGenreVM);

"

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/tutorials/first-mvc-app/validation?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/tutorials/first-mvc-app/validation.md

### Document ID

12fe9329-2cef-4056-4986-18ed7ae5c397

### Article author

@wadepickett

wadepickett:
Fixes #32661 

There is a problem that can occur if the developer happens to enter a null value for one of the movie items.  When we move on to add validation to the model, any entries that are null will cause an error to be thrown: 
SqlNullValueException: Data is Null. This method or property cannot be called on Null values.

To get around this, I added a line of instruction to delete the movie data entries just before we add validation in the next step so it will re-seed with clean data.

### DIFF
--- a/aspnetcore/tutorials/first-mvc-app/validation.md
+++ b/aspnetcore/tutorials/first-mvc-app/validation.md
@@ -28,6 +28,10 @@ One of the design tenets of MVC is [DRY](https://wikipedia.org/wiki/Don%27t_repe
 
 The validation support provided by MVC and Entity Framework Core Code First is a good example of the DRY principle in action. You can declaratively specify validation rules in one place (in the model class) and the rules are enforced everywhere in the app.
 
+## Delete any previous movie data items to re-seed the data.
+
+In the next step validation rules are added that will no longer allow null values. Run the app, navigate to `/Movies/Index` and delete all listed movies and stop the app.  The app will use the seed data the next time it is run.
+
 [!INCLUDE[](~/includes/RP-MVC/validation-net7.md)]
 
 ## Validation Error UI


wadepickett:
@tdykstra, does this seem like a good way to handle this to you?  It would take the developer more time to go into the database and delete directly than to just delete the view items right in the browser.

Another option I considered was to have them delete as a final step in the previous tutorial so the data would re-seed.

The problem occurs if the developer happens to delete the contents of a field so that a null is entered in the data set, which is allowed at the start of tutorial series for some of the properties in order to explain that.

aspnetcore/tutorials/first-mvc-app/validation.md
```diff
@@ -28,6 +28,11 @@ One of the design tenets of MVC is [DRY](https://wikipedia.org/wiki/Don%27t_repe
 
 The validation support provided by MVC and Entity Framework Core Code First is a good example of the DRY principle in action. You can declaratively specify validation rules in one place (in the model class) and the rules are enforced everywhere in the app.
 
+## Delete the previously edited data
+
+In the next step validation rules are added that will no longer allow null values.
```
tdykstra, line 33:
```suggestion
In the next step, validation rules are added that don't allow null values. 
```

aspnetcore/tutorials/first-mvc-app/validation.md
```diff
@@ -28,6 +28,11 @@ One of the design tenets of MVC is [DRY](https://wikipedia.org/wiki/Don%27t_repe
 
 The validation support provided by MVC and Entity Framework Core Code First is a good example of the DRY principle in action. You can declaratively specify validation rules in one place (in the model class) and the rules are enforced everywhere in the app.
 
+## Delete the previously edited data
+
+In the next step validation rules are added that will no longer allow null values. 
+Run the app, navigate to `/Movies/Index`, delete all listed movies and stop the app.  The app will use the seed data the next time it is run.
```
tdykstra, line 34:
```suggestion
Run the app, navigate to `/Movies/Index`, delete all listed movies, and stop the app.  The app will use the seed data the next time it is run.
```

tdykstra (approved):
This will work. Another option is to have them enter a value in any null field.

PR #32757, opened 2024-06-06
Update quickgrid.md

dsschnau:
Update the QuickGrid implementation directions that the generic type parameter must be a class and not a record type. 

I found that implementing a QuickGrid with a record type rather than a class caused some broken behaviors.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/quickgrid.md](https://github.com/dotnet/AspNetCore.Docs/blob/ab5ddbc846da15f409a03cf9aac9177c4f4610c3/aspnetcore/blazor/components/quickgrid.md) | [ASP.NET Core Blazor `QuickGrid` component](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/quickgrid?branch=pr-en-us-32757) |

### DIFF
--- a/aspnetcore/blazor/components/quickgrid.md
+++ b/aspnetcore/blazor/components/quickgrid.md
@@ -36,7 +36,7 @@ To implement a `QuickGrid` component:
 
 * Specify tags for the `QuickGrid` component in Razor markup (`<QuickGrid>...</QuickGrid>`).
 * Name a queryable source of data for the grid. Use ***either*** of the following data sources:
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks @dsschnau!

I'll ask Mackinnon on the product unit to take a quick look.

MackinnonBuck:
A record type is really just a class that desugars

MackinnonBuck:
Records are just syntax sugar for classes that encapsulate some data and emulate value equality (among some other things that I doubt have any impact on QuickGrid functionality). I don't think QuickGrid relies on any specific kind of equality semantics for the `TGridItem`, so I'm not sure why using a `record` would make any difference.

@dsschnau, could you please clarify what broken behavior you're seeing when using records? Thanks!

dsschnau:
Will do, thank you!

Issue #32692, opened 2024-05-29
Document what's new in ASP.NET Core for .NET 9 Preview 5


danroth27:
### Description

Update the "What's new in ASP.NET Core 9.0" for .NET 9 Preview 5

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-9.0?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/release-notes/aspnetcore-9.0.md

### Document ID

4e75ad25-2c3f-b28e-6a91-ac79a9c683b6

### Article author

@Rick-Anderson @tdykstra @guardrex

mkArtakMSFT:
@MackinnonBuck please add a comment documenting the ReconnectionUI work that you've done.

mkArtakMSFT:
@javiercn please add a comment with the Static Assets compression work that you've done.

mkArtakMSFT:
@halter73 please add a comment documenting the AuthenticationStateProvider work.

MackinnonBuck:
## New Blazor Server Reconnection Experience:

The following changes have been made to the default Blazor Server reconnection experience:
* Reconnect timing now uses an exponential backoff strategy. The first several reconnection attempts will happen in rapid succession before a delay gradually gets introduced between attempts. This behavior can be customized by specifying a function to compute the retry interval. For example:
  ```js
  Blazor.start({
    circuit: {
      reconnectionOptions: {
        retryIntervalMilliseconds: (previousAttempts, maxRetries) => previousAttempts >= maxRetries ? null : previousAttempts * 1000,
      },
    },
  });
  ```
* When the user navigates back to an app with a disconnected circuit, a reconnect attempt gets immediately made rather than waiting for the duration of the next reconnect interval. This especially improves the user experience when navigating to an app in a browser tab that has gone to sleep.
* If a reconnection attempt reaches the server, but reconnection fails because server had already released the circuit, a refresh occurs automatically. This prevents the user from having to manually refresh the page if it's likely going to result in a successful reconnection.
* The styling of the default reconnect UI has been modernized.

BethMassi:
# .NET MAUI Blazor Hybrid and Web Solution Template
[...]

Issue #31909, opened 2024-02-26
Blazor .NET 9 tracking issue


guardrex:
## ASP.NET Core 9: Web UI Milestones

https://github.com/orgs/dotnet/projects/320/views/3

## Derived from the 9.0 roadmap

🟢 = Ready&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; ✔️ = Done&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; 🛑 = Wait on PU to finish&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; ❌ = No action required

Status | Live | Issue | PR | Item | PU Links/<br>Blog
:---: | :---: | :---: | :---: | --- | :---:
✔  | [Live](https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/static-server-rendering?view=aspnetcore-9.0) | #34086 | #34101 | `enhancednavigationstart`/`enhancednavigationend` events | 
✔  | [Live 1](https://learn.microsoft.com/en-us/aspnet/core/blazor/file-uploads?view=aspnetcore-9.0#file-size-read-and-upload-limits)<br>[Live 2](https://learn.microsoft.com/en-us/aspnet/core/blazor/call-web-api?view=aspnetcore-9.0#client-side-request-streaming)<br>[Live 3](https://learn.microsoft.com/en-us/aspnet/core/blazor/file-uploads?view=aspnetcore-9.0#upload-files-to-a-server-with-client-side-rendering-csr) | #33692 | #33693 | WASM streaming upload | [PU PR](https://github.com/dotnet/runtime/pull/91295)
✔  | [Live](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/signalr?view=aspnetcore-9.0#reflect-the-server-side-connection-state-in-the-ui) | #34102 | #34104 | Update reconnection UI guidance | 
✔ | [Live](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/built-in-components?view=aspnetcore-9.0) | #34076 | #34077 | Check/confirm no new components for built-in components article | 
✔ | &mdash; | https://github.com/dotnet/blazor-samples/issues/373 | https://github.com/dotnet/blazor-samples/issues/374 | Update components and assets in sample apps |
✔ | &mdash; | #33869 | #33870<br>#33872 | Cross-link 9.0 sample code into articles | 
[...]

PR #32790, opened 2024-06-10
Blazor Server reconnection coverage

### LINKED ISSUE #32789: Blazor Server reconnection (9.0 Preview 5)
### Description

Per https://github.com/dotnet/AspNetCore.Docs/issues/32692#issuecomment-2155184689.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/signalr?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/fundamentals/signalr.md

### Document ID

d17e4c28-ea82-0dd8-295d-19f6ada147ea

### Article author

@guardrex

guardrex:
Fixes #32789
Addresses #32692
Addresses #31909

Mackinnon, ~I'm blocked at the moment on working on the *What's New* side of this until another PR merges, but I'll have the *What's New* part for this updated (some coverage was placed that I'll be touching up) by EOD on a separate PR.~ 

**UPDATE: Not blocked any longer on What's New. I'll take care of it after the four reference article PRs are merged.**

WRT this coverage: It seems like we'd call the default logic "a computed backoff strategy" instead of "an exponential backoff strategy." However, your custom example is exponential, so I retain that language for the example.

I was curious about the default logic, and I think devs will be curious, too. The link goes here ...

https://github.com/search?q=repo%3Adotnet%2Faspnetcore%20computeDefaultRetryInterval&type=code

... and that's probably a decent approach for avoiding a hard link with a specific set of line numbers. It will work well as long as the FN name `computeDefaultRetryInterval` is stable. I like the way that it finds the exact line in the file for them when they click on the search result, and I like the way that it would find this FN if it moved to a different file in the future. However, we can pull that link and those remarks if you don't want to risk it breaking at some point.

As usual, the INCLUDE that explains what the link loads follows the link, and the content of that INCLUDE can be seen here ...

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/includes/aspnetcore-repo-ref-source-links.md

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/signalr.md](https://github.com/dotnet/AspNetCore.Docs/blob/491ec9b718c2073baca0de381245f67282cf2635/aspnetcore/blazor/fundamentals/signalr.md) | [ASP.NET Core Blazor SignalR guidance](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/signalr?branch=pr-en-us-32790) |

### DIFF
--- a/aspnetcore/blazor/fundamentals/signalr.md
+++ b/aspnetcore/blazor/fundamentals/signalr.md
@@ -1127,6 +1127,31 @@ Blazor Server:
 
 **In the preceding example, the `{BLAZOR SCRIPT}` placeholder is the Blazor script path and file name.** For the location of the script and the path to use, see <xref:blazor/project-structure#location-of-the-blazor-script>.
 
+:::moniker range=">= aspnetcore-9.0"
+
+When the user navigates back to an app with a disconnected circuit, reconnection is attempted immediately rather than waiting for the duration of the next reconnect interval. This improves the user experience when navigating to an app in a browser tab that has gone to sleep.
+
+When a reconnection attempt reaches the server but the server has already released the circuit, a page refresh occurs automatically. This prevents the user from having to manually refresh the page if it's likely going to result in a successful reconnection.
+
+Reconnect timing uses a computed backoff strategy. The first several reconnection attempts occur in rapid succession without a retry interval before computed delays are introduced between attempts. The default logic for computing the retry interval is an implementation detail subject to change without notice, but you can find the default logic that the Blazor framework uses [in the `computeDefaultRetryInterval` function (reference source)](https://github.com/search?q=repo%3Adotnet%2Faspnetcore%20computeDefaultRetryInterval&type=code).
+
+[!INCLUDE[](~/includes/aspnetcore-repo-ref-source-links.md)]
+
+Customize the retry interval behavior by specifying a function to compute the retry interval. In the following exponential backoff example, the number of previous reconnection attempts is multiplied by 1,000 ms to calculate the retry interval. When the count of previous attempts to reconnect (`previousAttempts`) is greater than the maximum retry limit (`maxRetries`), `null` is assigned to the retry interval (`retryIntervalMilliseconds`) to cease further reconnection attempts:
+
+```javascript
+Blazor.start({
+  circuit: {
+    reconnectionOptions: {
+      retryIntervalMilliseconds: (previousAttempts, maxRetries) => 
+        previousAttempts >= maxRetries ? null : previousAttempts * 1000
+    },
+  },
+});
+```
+
+:::moniker-end
+
 For more information on Blazor startup, see <xref:blazor/fundamentals/startup>.
 
 :::moniker range=">= aspnetcore-6.0"
--- a/aspnetcore/blazor/fundamentals/signalr.md
+++ b/aspnetcore/blazor/fundamentals/signalr.md
@@ -1133,7 +1133,7 @@ When the user navigates back to an app with a disconnected circuit, reconnection
 
 When a reconnection attempt reaches the server but the server has already released the circuit, a page refresh occurs automatically. This prevents the user from having to manually refresh the page if it's likely going to result in a successful reconnection.
 
-Reconnect timing uses a computed backoff strategy. The first several reconnection attempts occur in rapid succession without a retry interval before computed delays are introduced between attempts. The default logic for computing the retry interval is an implementation detail subject to change without notice, but you can find the default logic that the Blazor framework uses [in the `computeDefaultRetryInterval` function (reference source)](https://github.com/search?q=repo%3Adotnet%2Faspnetcore%20computeDefaultRetryInterval&type=code).
+The default reconnect timing uses a computed backoff strategy. The first several reconnection attempts occur in rapid succession before computed delays are introduced between attempts. The default logic for computing the retry interval is an implementation detail subject to change without notice, but you can find the default logic that the Blazor framework uses [in the `computeDefaultRetryInterval` function (reference source)](https://github.com/search?q=repo%3Adotnet%2Faspnetcore%20computeDefaultRetryInterval&type=code).
 
 [!INCLUDE[](~/includes/aspnetcore-repo-ref-source-links.md)]
 
--- a/aspnetcore/blazor/fundamentals/signalr.md
+++ b/aspnetcore/blazor/fundamentals/signalr.md
@@ -1129,9 +1129,7 @@ Blazor Server:
 
 :::moniker range=">= aspnetcore-9.0"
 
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/fundamentals/signalr.md
```diff
@@ -1127,6 +1127,31 @@ Blazor Server:
 
 **In the preceding example, the `{BLAZOR SCRIPT}` placeholder is the Blazor script path and file name.** For the location of the script and the path to use, see <xref:blazor/project-structure#location-of-the-blazor-script>.
 
+:::moniker range=">= aspnetcore-9.0"
+
+When the user navigates back to an app with a disconnected circuit, reconnection is attempted immediately rather than waiting for the duration of the next reconnect interval. This improves the user experience when navigating to an app in a browser tab that has gone to sleep.
+
+When a reconnection attempt reaches the server but the server has already released the circuit, a page refresh occurs automatically. This prevents the user from having to manually refresh the page if it's likely going to result in a successful reconnection.
```
MackinnonBuck, line 1134: Hm, while these are noteworthy improvements to mention in release notes, I'm not sure if these details are super relevant to this section. Do you think it's fine to leave them out?

aspnetcore/blazor/fundamentals/signalr.md
```diff
@@ -1127,6 +1127,31 @@ Blazor Server:
...
+
+Reconnect timing uses a computed backoff strategy. The first several reconnection attempts occur in rapid succession without a retry interval before computed delays are introduced between attempts. The default logic for computing the retry interval is an implementation detail subject to change without notice, but you can find the default logic that the Blazor framework uses [in the `computeDefaultRetryInterval` function (reference source)](https://github.com/search?q=repo%3Adotnet%2Faspnetcore%20computeDefaultRetryInterval&type=code).
```
MackinnonBuck, line 1136:
```suggestion
The default reconnect timing uses a computed backoff strategy. The first several reconnection attempts occur in rapid succession before computed delays are introduced between attempts. The default logic for computing the retry interval is an implementation detail subject to change without notice, but you can find the default logic that the Blazor framework uses [in the `computeDefaultRetryInterval` function (reference source)](https://github.com/search?q=repo%3Adotnet%2Faspnetcore%20computeDefaultRetryInterval&type=code).
```

aspnetcore/blazor/fundamentals/signalr.md
```diff
@@ -1127,6 +1127,31 @@ Blazor Server:
...
+
+```javascript
+Blazor.start({
+  circuit: {
+    reconnectionOptions: {
+      retryIntervalMilliseconds: (previousAttempts, maxRetries) => 
+        previousAttempts >= maxRetries ? null : previousAttempts * 1000
+    },
+  },
+});
```
MackinnonBuck, line 1150: Totally optional, but we could also mention another neat little trick if you want more precise control over each retry interval:

```js
Blazor.start({
  circuit: {
    reconnectionOptions: {
      retryIntervalMilliseconds: Array.prototype.at.bind([0, 1000, 2000, 5000, 10000, 15000, 30000]),
    },
  },
});
```

This just lets you just list out the retry intervals directly. After the last specified retry interval occurs, retries stop (because the function returns `undefined`).

MackinnonBuck:
> It will work well as long as the FN name `computeDefaultRetryInterval` is stable

Yeah, I think that's probably fine. In the worst case, we accidentally rename it in the future, but we then will likely get reports of a broken link and fix it.

aspnetcore/blazor/fundamentals/signalr.md
guardrex: I think the first one is more relevant to the section, but I see what you mean about the second one.

Do you want me to kill both 🔪 ... or just the second one?

aspnetcore/blazor/fundamentals/signalr.md
guardrex: Yes, I'll add that. That's nice.

aspnetcore/blazor/fundamentals/signalr.md
MackinnonBuck: I think we can keep the first one, but maybe without the "improves the user experience" bit because "improves" sounds relative to how it worked before, which to me seems more suitable for release notes. But we could still mention the reasoning for it working that way, which is that if the user is actively navigating back to the app, they likely expect things to resume working as quickly as possible (we know they're not sitting idly and twiddling their thumbs).

aspnetcore/blazor/fundamentals/signalr.md
```diff
@@ -1127,6 +1127,41 @@ Blazor Server:
 
 **In the preceding example, the `{BLAZOR SCRIPT}` placeholder is the Blazor script path and file name.** For the location of the script and the path to use, see <xref:blazor/project-structure#location-of-the-blazor-script>.
 
+:::moniker range=">= aspnetcore-9.0"
+
+When the user navigates back to an app with a disconnected circuit, reconnection is attempted immediately rather than waiting for the duration of the next reconnect interval to resume  the connection as quickly as possible for the user.
```
MackinnonBuck:
```suggestion
When the user navigates back to an app with a disconnected circuit, reconnection is attempted immediately rather than waiting for the duration of the next reconnect interval to resume the connection as quickly as possible for the user.
```

guardrex:
Try it now, but I'm not sure if that wording is exactly right ... I've been working here from 6am to 8pm non-stop 😩.

aspnetcore/blazor/fundamentals/signalr.md
```diff
@@ -1127,6 +1127,41 @@ Blazor Server:
...
+  circuit: {
+    reconnectionOptions: {
+      retryIntervalMilliseconds: (previousAttempts, maxRetries) => 
+        previousAttempts >= maxRetries ? null : previousAttempts * 1000
+    },
+  },
+});
+```
+
+Specify the exact milliseconds of each interval for precise retry interval control. After the last specified retry interval, retries stop because the function returns `undefined`:
```
MackinnonBuck, line 1151: Nit:
```suggestion
An alternative is to specify the exact sequence of retry intervals. After the last specified retry interval, retries stop because the `retryIntervalMilliseconds` function returns `undefined`:
```

MackinnonBuck (approved):

aspnetcore/blazor/fundamentals/signalr.md
MackinnonBuck: Looks like you already caught this one!

guardrex: I think that one I already fixed.

guardrex:
Ok ... all good.

# 🍻 

I'll get this merged as soon as it builds.

PR #32822, opened 2024-06-11
Update route-handlers.md

DickBaker:
inconsistency between code and doc postscript


guardrex (approved):
Thanks @DickBaker! 🚀 

Cross-ref: https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/minimal-apis/samples/WebMinAPIs/Program.cs#L384

PR #32855, opened 2024-06-14
Change description of code behavior

### LINKED ISSUE #32713: Misleading sentece about use of HttpContext to determine interactivity location
### Description

In the article we read that :  
> The component makes use of [HttpContext](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httpcontext) to determine if the component is rendering on the server".  

In case `HttpContext` has value, we can be sure that component is rendered using SSR, so it is _on the server_. However, in case `HttpContext` is `null` all we get to know is that the render mode is not "static SSR". We can't determine if it is rendered on client (CSR) or on server (interactive SSR). Hence `HttpContext` can't be used to determine whether the component is rendering on server or not, as In case of null, it could be server or client.

What I told is based on:  [This official document](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/http-context?view=aspnetcore-8.0#ihttpcontextaccessorhttpcontext-in-razor-components-blazor)

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/render-modes?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/render-modes.md

### Document ID

7c9948a3-9886-8097-6014-492a4c21ea4e

### Article author

@guardrex

guardrex:
Fixes #32713

Thanks @ShahryarSaljoughi! 🚀 

Mackinnon ... In addition to this change, I'm still trying to find out if this remark is correct ...

> `IHttpContextAccessor` must be avoided with interactive rendering because there isn't a valid `HttpContext` available.

... in the section at ...

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/http-context?view=aspnetcore-8.0#ihttpcontextaccessorhttpcontext-in-razor-components-blazor

... because the BWA+OIDC sample seems to be using it in the Server Weather Forecaster service (`ServerWeatherForecaster`) ...

https://github.com/dotnet/blazor-samples/blob/main/8.0/BlazorWebAppOidcBff/BlazorWebAppOidc/ServerWeatherForecaster.cs

... and the app adopts global `InteractiveAuto`. It seems like the [`Weather` component](https://github.com/dotnet/blazor-samples/blob/main/8.0/BlazorWebAppOidcBff/BlazorWebAppOidc.Client/Pages/Weather.razor) that injects the service would be using it when rendering interactively on the server after prerendering.

It was being discussed over on the Blazor samples repo at https://github.com/dotnet/blazor-samples/issues/258. I've pinged/emailed several times, but I still don't have a answer or understanding on use of `IHttpContextAccessor` in that context and if that remark is correct ... or always correct.

### DIFF
--- a/aspnetcore/blazor/components/render-modes.md
+++ b/aspnetcore/blazor/components/render-modes.md
@@ -670,7 +670,7 @@ The folder also contains an `_Imports.razor` file, which applies a custom accoun
 @layout AccountLayout
 ```
 
-The `Shared` folder maintains the `AccountLayout` layout component. The component makes use of <xref:Microsoft.AspNetCore.Http.HttpContext> to determine if the component is rendering on the server. Identity components must render on the server with static SSR because they set Identity cookies. If the value of <xref:Microsoft.AspNetCore.Http.HttpContext> is `null`, the component is rendering interactively, and a full-page reload is performed by calling <xref:Microsoft.AspNetCore.Components.NavigationManager.Refresh%2A?displayProperty=nameWithType> with `forceLoad` set to `true`. This forces a full rerender of the page using static SSR.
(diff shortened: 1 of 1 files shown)


guardrex:
I'm going to go ahead with this because I'm OOF all of next week, and need to clear these PRs through. I think the language update here is correct.

However, I'm still concerned about the language on `IHttpContextAccessor` :point_up:. That's being tracked at the moment over in the Blazor samples repo at https://github.com/dotnet/blazor-samples/issues/258.

Mackinnon ... I think you'll see this when you get back (or get free). Could you take a look at my concern in the OP and let me know what you think because it looks to me like our remark is incorrect or incomplete that ...

> `IHttpContextAccessor` must be avoided with interactive rendering because there isn't a valid `HttpContext` available.

PR #32856, opened 2024-06-14
Import-Export interop: collocated JS with RCL

### LINKED ISSUE #32618: Make a clear description of file location for JSHost.ImportAsync
### Description

The documentation currently states

> The CallJavaScript1 module is imported asynchronously from the [collocated JS file](https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/location-of-javascript?view=aspnetcore-8.0#load-a-script-from-an-external-javascript-file-js-collocated-with-a-component) with [JSHost.ImportAsync](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.javascript.jshost.importasync).

It's not entirely true based on other file collocation that works in RCL. `JSHost.ImportAsync` always loads files relative to `wwwroot/_framework` directory. Examples for paths for `JSHost.ImportAsync`:
- If a file to load is `wwwroot/MyModule.js`, then relative path can be `../MyModule.js`
- If a file to load is in RCL `MyLibrary` in `wwwroot/MyLibraryModule.js`, then relative path can be `../_content/MyLibrary/MyLibraryModule.js` (this is true for both cases when the import is done from RLC itself or from the Blazor app)
Note: absolute paths could be used as well, but then it could cause problems when the site is not hosted in the root of the domain.

Based on feedback from https://github.com/dotnet/aspnetcore/issues/55817

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/import-export-interop?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/javascript-interoperability/import-export-interop.md

### Document ID

8aab92a0-c05f-ef01-fb74-a0bf525aae69

### Article author

@guardrex

guardrex:
Fixes #32618

Marek ... The article that covers JS location ...

https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/location-of-javascript?view=aspnetcore-8.0

... is meant to cover all of the location scenarios, and it looks like it's just this ***one thing*** that isn't supported: Import-Export interop with collocated JS in an RCL.

If that's a correct interpretation, then I think a section should be added to the Import-Export article that just calls out that the collocation feature isn't supported, leaving the JS location article to continue to serve its purpose for both JS interop APIs. I'm adding that section here, and I'm cross-linking it from the JS location article where it talks about collocated JS from an RCL. 

Also, it seemed like a little advice can be offered to devs trying to keep their JS organized by component because that's the whole purpose of collocated JS in the first place: They can name the JS file to match the component, and they can place component-specific JS files into a `Components` folder in `wwwroot` of the RCL. We can strike that bit if it seems obvious or inappropriate for some reason.

Note that your remark on how "JSHost.ImportAsync always loads files relative to `wwwroot/_framework` directory" is an implementation detail that we don't cover, so dropping it into the coverage is going to be a little confusing. The only place in the entire Blazor doc set that we talk about that is for loading boot resources.

Also note that that JS collocation coverage is actually provided by an INCLUDE file. It's not only in the JS location article. It also gets sucked into the class libraries article, so the new cross-link will appear there as well after this merges ...

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/class-libraries?view=aspnetcore-8.0&tabs=visual-studio#create-an-rcl-with-javascript-files-collocated-with-components

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/javascript-interoperability/import-export-interop.md](https://github.com/dotnet/AspNetCore.Docs/blob/5e3dd35037e00406b8c598eb115fdda00c59924b/aspnetcore/blazor/javascript-interoperability/import-export-interop.md) | [JavaScript `[JSImport]`/`[JSExport]` interop with ASP.NET Core Blazor](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/import-export-interop?branch=pr-en-us-32856) |

### DIFF
--- a/aspnetcore/blazor/includes/js-interop/js-collocation.md
+++ b/aspnetcore/blazor/includes/js-interop/js-collocation.md
@@ -144,7 +144,7 @@ export function showPrompt2(message) {
 }
 ```
 
-Use of scripts and modules for collocated JS in a Razor class library (RCL) is only supported for Blazor's JS interop mechanism based on the <xref:Microsoft.JSInterop.IJSRuntime> interface. If you're implementing [JavaScript `[JSImport]`/`[JSExport]` interop](xref:blazor/js-interop/import-export-interop), see <xref:blazor/js-interop/import-export-interop#collocated-js-in-a-razor-class-library-rcl-is-unsupported>.
+Use of scripts and modules for collocated JS in a Razor class library (RCL) is only supported for Blazor's JS interop mechanism based on the <xref:Microsoft.JSInterop.IJSRuntime> interface. If you're implementing [JavaScript `[JSImport]`/`[JSExport]` interop](xref:blazor/js-interop/import-export-interop), see <xref:blazor/js-interop/import-export-interop#razor-class-library-rcl-collocated-js-is-unsupported>.
 
 For scripts or modules provided by a Razor class library (RCL) using <xref:Microsoft.JSInterop.IJSRuntime>-based JS interop, the following path is used:
 
--- a/aspnetcore/blazor/includes/js-interop/js-collocation.md
+++ b/aspnetcore/blazor/includes/js-interop/js-collocation.md
@@ -144,7 +144,9 @@ export function showPrompt2(message) {
 }
 ```
 
-For scripts or modules provided by a Razor class library (RCL), the following path is used:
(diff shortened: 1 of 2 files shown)


maraf (approved):
Looks good to me. Thank you!

PR #32882, opened 2024-06-18
.NET JS interop article updates

### LINKED ISSUE #32660: [UUF Port] dotnet-interop: May need an update for .NET 8
### Description

Unified User Feedback Port:
Customer feedback:
"This content does not appear to be correct for .NET 8. The project file format is different from what is documented here when using .NET 8 experimental template. There also no longer seems to be an AppBundle.The new template uses Microsoft.NET.Sdk.WebAssembly vs Microsoft.NET.Sdk"

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/client-side/dotnet-interop?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/client-side/dotnet-interop.md

### Document ID

efff6954-a035-d1da-5975-af04ff7c05ed

### Article author

@pavelsavara

guardrex:
Fixes #32660

Marek ... ***UPDATE***: Making progress ...

* Version-specific language updates. Version new coverage for .NET 8+.
* Change from the `Microsoft.NET.Sdk` SDK to the `Microsoft.NET.Sdk.WebAssembly` SDK.
* Update code and language for .NET 8+.
* Update guidance for workload and template installation.  

See my latest comments on three items of concern :point_down:.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/client-side/dotnet-interop.md](https://github.com/dotnet/AspNetCore.Docs/blob/fca6875016950ab06398315fe2d1ec1f4a9e8ccf/aspnetcore/client-side/dotnet-interop.md) | [Run .NET from JavaScript](https://review.learn.microsoft.com/en-us/aspnet/core/client-side/dotnet-interop?branch=pr-en-us-32882) |

### DIFF
--- a/aspnetcore/client-side/dotnet-interop.md
+++ b/aspnetcore/client-side/dotnet-interop.md
@@ -168,15 +168,64 @@ In the following example:
 
 * `dotnet.create()` sets up the .NET WebAssembly runtime.
 
+:::moniker range=">= aspnetcore-9.0"
+
+* `setModuleImports` associates a name with a module of JS functions for import into .NET. The JS module contains a `dom.setInnerText` function, which accepts and element selector and time to display the current stopwatch time in the UI. The name of the module can be any string (it doesn't need to be a file name), but it must match the name used with the `JSImportAttribute` (explained later in this article). The `dom.setInnerText` function is imported into C# and called by the C# method `SetInnerText`. The `SetInnerText` method is shown later in this section.
+
+* `exports.StopwatchSample.Toggle()` calls into .NET (`StopwatchSample.Toggle`) from JS. The `Toggle` C# method starts or stops the stopwatch depending on if it's currently running or not. The `Toggle` method is shown later in this section.
+
+* `runMain()` runs `Program.Main` and keeps the runtime process running and executing further API calls.
+
+:::moniker-end
+
+:::moniker range="< aspnetcore-9.0"
+
 * `setModuleImports` associates a name with a module of JS functions for import into .NET. The JS module contains a `window.location.href` function, which returns the current page address (URL). The name of the module can be any string (it doesn't need to be a file name), but it must match the name used with the `JSImportAttribute` (explained later in this article). The `window.location.href` function is imported into C# and called by the C# method `GetHRef`. The `GetHRef` method is shown later in this section.
 
 * `exports.MyClass.Greeting()` calls into .NET (`MyClass.Greeting`) from JS. The `Greeting` C# method returns a string that includes the result of calling the `window.location.href` function. The `Greeting` method is shown later in this section.
 
 * `dotnet.run()` runs `Program.Main`.
 
+:::moniker-end
+
 JS module:
 
-:::moniker range=">= aspnetcore-8.0"
+:::moniker range=">= aspnetcore-9.0"
+
+```javascript
+import { dotnet } from './_framework/dotnet.js'
+
+const { setModuleImports, getAssemblyExports, getConfig, runMain } = await dotnet
+    .withApplicationArguments("start")
+    .create();
+
+setModuleImports('main.js', {
+    dom: {
+        setInnerText: (selector, time) => document.querySelector(selector).innerText = time
+    }
+});
+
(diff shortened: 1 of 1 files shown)


guardrex:
BTW @maraf ... Can the `wasm-experimental` workload be installed for VS? I was able to install the **.NET WebAssembly build tools** (`wasm-tools`) workload in VS, and I'm wondering if I can get the `wasm-experimental` workload as well? If so, I think our guidance in this article should be updated to include the use of VS.

Anyway ... thus far ... the CLI is a total failure here. The command output makes no sense. Installing the workload yields ...

> Successfully installed workload(s) wasm-tools android aspire ios maccatalyst maui-windows.

Then, I list the installed workload, and that workload doesn't show up in the list.

guardrex:
Publishing seems to be broken. The target location is different from where the app is actually published.

**Target location**:

image

**Actual publish location**:

image

image

guardrex:
@maraf ... I've been *partially* successful in updating the content.

Here are the problem areas:

1. I can't get the workload install commands to work here.

   What was successful for me was to use the VS installer with the **.NET WebAssembly build tools** option. That got the tools installed. Then, I used the NuGet package to get the templates ...

   ```dotnetcli
   dotnet new install Microsoft.NET.Runtime.WebAssembly.Templates
   ```

   After that, I was able to create the projects from either the command line or from VS.

   I've updated the guidance to explain both approaches; but note that AFIACT from here, there's a problem with the CLI workload install command approach for both the tools and the templates ... at least on my system there's some kind of problem. I'm 👂 on how you want to handle it.

1. The `dotnet serve` command doesn't seem to be a command. It's [not documented](https://learn.microsoft.com/dotnet/core/tools/dotnet), and it's not available on my system ...

   image

1. The publish output location doesn't match the **Target location** in VS during publishing. The SDK uses `bin/Release/net8.0/publish`, but the target location shows `bin/Release/net8.0/browser-wasm/publish` ... and note that the default was always for the "`publish`" folder, not a folder named "`AppBundle`", so IDK if that was a custom path that was set by the original author of the content that came over or not. I've updated the guidance on the PR to match the actual published assets path.

maraf:
I'm sorry for later reply, I'm leaving for next two weeks and I'm a bit under the water .

1. The approach with installing just templates is possible, but installing wasm-experimental workload seems working on my machine™️. Could you try some VM or sandbox or DevBox? In .NET 9 P4 (or P3) I have changed the code in the template and it would be good to reference the changed code in the docs.
2. The `dotnet serve` is a .NET tool and needs to be installed separatelly https://github.com/natemcmaster/dotnet-serve. Anyway, in .NET 9 and also 8 and 7 you can do just `dotnet run` for the wasmbrowser template app and it will spin up a built-in server.
3. Yes, the publish from VS is broken. We are tracking it in https://github.com/dotnet/runtime/issues/99159

guardrex:
Me, too ... I'm OOF all of next week. I'll try to wrap this up now and get it merged.

1. Ok ... but I need to find the changes on the repo because it might be difficult for me to get the preview template installed locally given the trouble that I'm having with this workload. I'll look in the repo at the template to see if I can make the updates. As for using a VM to test commands, I don't think I'll have time. I only have three hours today before my annual contract ends, then I'll be back to work on July 1. It's best to merge this PR now before I go.
2. Thanks ... got it ... PR updated.
3. I'll go with what the SDK does now on the PR, and I've added a tracking item to my .NET 9 issue to change this article when the publish path problem is fixed.

guardrex:
I see you changed over to a stopwatch example.

https://github.com/dotnet/runtime/blob/main/src/mono/wasm/templates/templates/browser/wwwroot/main.js

guardrex:
@maraf ... I wrote that up fairly fast given that I'm leaving in just a couple of hours. Hopefully, it's correct or not too far off.

guardrex:
I'm going to go ahead and merge this, and we can circle back around in a couple of weeks for additional updates.

PR #32988, opened 2024-07-02
Update integrationTests.md

ParsaMehdipour:
fix grammar issues


guardrex:
Hello @ParsaMehdipour ... These changes wouldn't be consistent with our style manual, so we won't be able to accept them. However, thanks for the suggestions.

ParsaMehdipour:
Got it
Thanks

PR #33006, opened 2024-07-08
[Blazor] Render-modes.md - InteractiveWebassembly => InteractiveWebAssembly

hakenr:
typo
cc @guardrex


guardrex (approved):
# 😈 

Thanks for the fix!

... and I checked the whole repo. This was the only instance of this particular 😈.

PR #33036, opened 2024-07-10
Improve DotNetStreamReference disposal remarks

### LINKED ISSUE #33013: Recommend patterns that avoid disposing DotNetStreamReference during transmission to the browser
### Description

For the *Stream from .NET to JavaScript* section per https://github.com/dotnet/aspnetcore/issues/53376#issuecomment-2215262524.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-javascript-from-dotnet?view=aspnetcore-8.0#stream-from-net-to-javascript

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md

### Document ID

9e197a8b-2616-8363-0a15-38cac802e82e

### Article author

@guardrex

guardrex:
Fixes #33013

Mackinnon ... Here are your original remarks :point_right: https://github.com/dotnet/aspnetcore/issues/53376#issuecomment-2215262524.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md](https://github.com/dotnet/AspNetCore.Docs/blob/675b5315d3c5ca566a4305f9d3c68cb57e68248b/aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md) | [Call JavaScript functions from .NET methods in ASP.NET Core Blazor](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-javascript-from-dotnet?branch=pr-en-us-33036) |

### DIFF
--- a/aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md
+++ b/aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md
@@ -1252,7 +1252,7 @@ In the preceding example:
 
 Disposing a <xref:Microsoft.JSInterop.DotNetStreamReference> instance is usually unnecessary. When `leaveOpen` is set to its default value of `false`, the underlying <xref:System.IO.Stream> is automatically disposed after transmission to JS.
 
-If `leaveOpen` is `true`, then disposing a <xref:Microsoft.JSInterop.DotNetStreamReference> doesn't dispose its underlying <xref:System.IO.Stream>. The app's code determines when to dispose the underlying <xref:System.IO.Stream>. When deciding how to dispose a <xref:Microsoft.JSInterop.DotNetStreamReference> and its underlying <xref:System.IO.Stream>, consider the following:
+If `leaveOpen` is `true`, then disposing a <xref:Microsoft.JSInterop.DotNetStreamReference> doesn't dispose its underlying <xref:System.IO.Stream>. The app's code determines when to dispose the underlying <xref:System.IO.Stream>. When deciding how to dispose the underlying <xref:System.IO.Stream>, consider the following:
 
 * Disposing a <xref:System.IO.Stream> while it's being transmitted to JS is considered an application error and may cause an unhandled exception to occur.
 * <xref:System.IO.Stream> transmission begins as soon as the <xref:Microsoft.JSInterop.DotNetStreamReference> is passed as an argument to a JS interop call, regardless of whether the stream is actually used in JS logic.
--- a/aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md
+++ b/aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md
@@ -1250,7 +1250,7 @@ In the preceding example:
 * The `{STREAM}` placeholder represents the <xref:System.IO.Stream> sent to JS.
 * `JS` is an injected <xref:Microsoft.JSInterop.IJSRuntime> instance.
 
-The `streamRef` varaible isn't declared with the `using` keyword because it would result in disposing the stream twice, once when transmission completes and again when `streamRef` exits the scope it was declared in. You can still use the `using` keyword if ***either*** of the following conditions are met:
+The `streamRef` varaible isn't declared with the `using` keyword because it would result in disposing the stream twice, once when transmission completes and again when `streamRef` exits the scope that it was declared in. You can still use the `using` keyword if ***either*** of the following conditions are met:
 
 * `leaveOpen` is set to `true` and .NET code waits for the stream to get completely consumed by JS before exiting the current scope. For example, it's valid to set `leaveOpen` to `true` and await a JS method invocation that returns a promise that completes only after the stream is completely consumed.
 * The `DotNetStreamReference` isn't passed to JS via JS interop.
--- a/aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md
+++ b/aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md
@@ -1252,10 +1252,10 @@ In the preceding example:
 
 Disposing a <xref:Microsoft.JSInterop.DotNetStreamReference> instance is usually unnecessary. When `leaveOpen` is set to its default value of `false`, the underlying <xref:System.IO.Stream> is automatically disposed after transmission to JS.
 
-If `leaveOpen` is `true`, then disposing a <xref:Microsoft.JSInterop.DotNetStreamReference> doesn't dispose its underlying <xref:System.IO.Stream>. The app's code determines when to dispose the underlying <xref:System.IO.Stream>. When deciding whether or not to manually dispose a <xref:Microsoft.JSInterop.DotNetStreamReference> and its underlying <xref:System.IO.Stream>, consider the following:
+If `leaveOpen` is `true`, then disposing a <xref:Microsoft.JSInterop.DotNetStreamReference> doesn't dispose its underlying <xref:System.IO.Stream>. The app's code determines when to dispose the underlying <xref:System.IO.Stream>. When deciding how to dispose a <xref:Microsoft.JSInterop.DotNetStreamReference> and its underlying <xref:System.IO.Stream>, consider the following:
 
-* Disposing a <xref:System.IO.Stream> while it's being transmitted to JS is considered an application error and may cause termination of the circuit or the WebAssembly app to crash.
-* <xref:System.IO.Stream> transmission begins as soon as the <xref:Microsoft.JSInterop.DotNetStreamReference> is passed as an argument to a JS interop call, regardless of whether the <xref:System.IO.Stream> is actually used in JS logic.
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md
```diff
@@ -1241,15 +1241,22 @@ In JavaScript, use an array buffer or a readable stream to receive the data:
...
+* The `{STREAM}` placeholder represents the <xref:System.IO.Stream> sent to JS.
 * `JS` is an injected <xref:Microsoft.JSInterop.IJSRuntime> instance.
 
+The `streamRef` varaible isn't declared with the `using` keyword because it would result in disposing the stream twice, once when transmission completes and again when `streamRef` exits the scope that it was declared in. Use the `using` keyword to manually dispose of the stream if ***either*** of the following conditions are met:
+
+* `leaveOpen` is set to `true` and .NET code waits for the stream to get completely consumed by JS before exiting the current scope. For example, it's valid to set `leaveOpen` to `true` and await a JS method invocation that returns a promise that completes only after the stream is completely consumed.
+* The `DotNetStreamReference` isn't passed to JS via JS interop.
+
+In most cases, `leaveOpen` is `false`, which is the default value, and the `DotNetStreamReference` is passed to JS, so you generally shouldn't manually dispose the `DotNetStreamReference`.
+
```
MackinnonBuck, line 1259: Looking this over again, I think we can simplify the guidance a bit further:
1. I realized that disposing a `DotNetStreamReference` is actually a no-op if `leaveOpen` is true (so we can remove the bit about disposing when `leaveOpen` is true).
2. We should suggest not creating a `DotNetStreamReference` at all unless it's passed to a JS interop invocation. I can't think of a valid pattern that requires this, as you can always create the `DotNetStreamReference` just before passing it to a JS interop call.

But then I was also thinking that it could be beneficial to expand on _why_ developers should be cautious about disposing `DotNetStreamReference`s.

So maybe we could instead say something like this:

> It's almost never necessary to dispose a `DotNetStreamReference` instance. When `leaveOpen` is set to its default value of `false`, the underlying stream gets automatically disposed after being transmitted to JS.

> If `leaveOpen` is `true`, then disposing a `DotNetStreamReference` does not dispose its `stream`. It's up to the application to decide when to dispose the underlying stream. In doing so, it's important to consider the following:
> 1. Disposing a stream while it's being transmitted to JS is considered an application error and may cause the circuit to get terminated or the WebAssembly app to crash.
> 2. Stream transmission begins as soon as the `DotNetStreamReference` gets passed as an argument to a JS interop call, regardless of whether the stream actually gets used in JS logic.

> Given these characteristics, we recommend disposing the underlying stream _only after it gets fully consumed by JS_ (i.e., the promise returned by `.arrayBuffer()` or `.stream()` resolves). It follows that a `DotNetStreamReference` should only get passed to JS if it is unconditionally going to be consumed by JS logic.

(feel free to adjust the wording to sounds more docs-ey 🙂)

guardrex:
@MackinnonBuck ... I may have hosed some of your explanation.

I was trying to distinguish the underlying stream, which I think is the `System.IO.Stream`, from the `DotNetStreamReference`.

aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md
```diff
@@ -1250,12 +1250,14 @@ In the preceding example:
...
+Disposing a <xref:Microsoft.JSInterop.DotNetStreamReference> instance is usually unnecessary. When `leaveOpen` is set to its default value of `false`, the underlying <xref:System.IO.Stream> is automatically disposed after transmission to JS.
 
-* `leaveOpen` is set to `true` and .NET code waits for the stream to get completely consumed by JS before exiting the current scope. For example, it's valid to set `leaveOpen` to `true` and await a JS method invocation that returns a promise that completes only after the stream is completely consumed.
-* The `DotNetStreamReference` isn't passed to JS via JS interop.
+If `leaveOpen` is `true`, then disposing a <xref:Microsoft.JSInterop.DotNetStreamReference> doesn't dispose its underlying <xref:System.IO.Stream>. The app's code determines when to dispose the underlying <xref:System.IO.Stream>. When deciding whether or not to manually dispose a <xref:Microsoft.JSInterop.DotNetStreamReference> and its underlying <xref:System.IO.Stream>, consider the following:
```
MackinnonBuck, line 1255:
```suggestion
If `leaveOpen` is `true`, then disposing a <xref:Microsoft.JSInterop.DotNetStreamReference> doesn't dispose its underlying <xref:System.IO.Stream>. The app's code determines when to dispose the underlying <xref:System.IO.Stream>. When deciding how to dispose a <xref:Microsoft.JSInterop.DotNetStreamReference> and its underlying <xref:System.IO.Stream>, consider the following:
```

(what I meant to convey in my original wording was how to decide the "point in time" at which the stream can be safely disposed)

aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md
```diff
@@ -1250,12 +1250,14 @@ In the preceding example:
...
 
-In most cases, `leaveOpen` is `false`, which is the default value, and the `DotNetStreamReference` is passed to JS, so you generally shouldn't manually dispose the `DotNetStreamReference`.
+* Disposing a <xref:System.IO.Stream> while it's being transmitted to JS is considered an application error and may cause termination of the circuit or the WebAssembly app to crash.
```
MackinnonBuck, line 1257: I know I'm editing my own wording here (sorry!) but maybe it's better to avoid language specific to a render mode/hosting model:
```suggestion
* Disposing a <xref:System.IO.Stream> while it's being transmitted to JS is considered an application error and may cause an unhandled exception to occur.
```

aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md
```diff
@@ -1250,12 +1250,14 @@ In the preceding example:
...
+* <xref:System.IO.Stream> transmission begins as soon as the <xref:Microsoft.JSInterop.DotNetStreamReference> is passed as an argument to a JS interop call, regardless of whether the <xref:System.IO.Stream> is actually used in JS logic.
```
MackinnonBuck, line 1258: Supernit: It's technically not a `System.IO.Stream` by the time it gets used in JS (I could be convinced to keep it how you have it though, so only accept this suggestion if you agree!).

```suggestion
* <xref:System.IO.Stream> transmission begins as soon as the <xref:Microsoft.JSInterop.DotNetStreamReference> is passed as an argument to a JS interop call, regardless of whether the stream is actually used in JS logic.
```

MackinnonBuck (approved):
Approving with some small suggestions 👍

aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md
MackinnonBuck: Oh, and since we're saying earlier that `DotNetStreamReference`s shouldn't get disposed, maybe we could shorten this to:

> When deciding how to dispose the underlying <xref:System.IO.Stream>, consider the following:

aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md
```diff
@@ -1241,15 +1241,24 @@ In JavaScript, use an array buffer or a readable stream to receive the data:
...
 
 In the preceding example:
 
-* The `{STREAM}` placeholder represents the <xref:System.IO.Stream> sent to JavaScript.
+* The `{STREAM}` placeholder represents the <xref:System.IO.Stream> sent to JS.
 * `JS` is an injected <xref:Microsoft.JSInterop.IJSRuntime> instance.
 
+Disposing a <xref:Microsoft.JSInterop.DotNetStreamReference> instance is usually unnecessary. When `leaveOpen` is set to its default value of `false`, the underlying <xref:System.IO.Stream> is automatically disposed after transmission to JS.
+
+If `leaveOpen` is `true`, then disposing a <xref:Microsoft.JSInterop.DotNetStreamReference> doesn't dispose its underlying <xref:System.IO.Stream>. The app's code determines when to dispose the underlying <xref:System.IO.Stream>. When deciding how to dispose a <xref:Microsoft.JSInterop.DotNetStreamReference> and its underlying <xref:System.IO.Stream>, consider the following:
```
guardrex, line 1255:
```suggestion
If `leaveOpen` is `true`, then disposing a <xref:Microsoft.JSInterop.DotNetStreamReference> doesn't dispose its underlying <xref:System.IO.Stream>. The app's code determines when to dispose the underlying <xref:System.IO.Stream>. When deciding how to dispose the underlying <xref:System.IO.Stream>, consider the following:
```

MackinnonBuck:
LGTM!

Issue #32138, opened 2024-03-25
Add additional Blazor migration guidance for upgrading to .NET 8


danroth27:
We've received some [feedback](https://github.com/dotnet/AspNetCore.Docs/issues/29522#issuecomment-2017263297) that our Blazor migration guidance is too minimalistic and doesn't cover all of the related concerns:

- Apps using lazy loading were impacted by the change to webcil, which changed the assembly extension from .dll to .wasm
- We changed our authentication guidance in .NET 8, but we didn't give detailed instructions on how to upgrade

SeanFeldman:
Why does the `<title>` has to be removed? This ensured a consistent title for the web app across all the pages.

![image](https://github.com/dotnet/AspNetCore.Docs/assets/1309622/f71d3301-32b3-4e28-8b6a-5c59ec364a3d)

guardrex:
@SeanFeldman ... The `App` component of a BWA doesn't set the page title that way ...

https://github.com/dotnet/aspnetcore/blob/main/src/ProjectTemplates/Web.ProjectTemplates/content/BlazorWeb-CSharp/BlazorWeb-CSharp/Components/App.razor

Page titles are specified via the `HeadOutlet`/`PageTitle` components ...

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/control-head-content?view=aspnetcore-8.0#control-head-content-in-a-razor-component

I'll cross-link that guidance for that change when I reach this issue.

It might still be possible to use a `<title>` tag for an app-wide fallback (possibly depending on ordering, too), but I don't recall if there are artifacts of using it ... i.e., 🐉😈. The idea here is to convert the app to a BWA, not try to maintain hosted WASM solution conventions, so that's why we're angling to have devs drop it for the migration guidance. It truly is a ***conversion*** to a BWA, not merely an updated/upgraded hosted WASM solution. ***HOWEVER, note that*** you can apply a `PageTitle` component to your layout to get it for all of the components that don't specify their own `PageTitle` component.

For example in `Components/Layout/MainLayout.razor` ...

```razor
<PageTitle>My layout title</PageTitle>
```

... any component without a page title specified will get that one. When I reach this issue, I'll see if we mention this fact in the *Control \<head> content* article. If not, I'll make a remark on it there.

guardrex:
@danroth27 ... 

> We changed our authentication guidance in .NET 8, but we didn't give detailed instructions on how to upgrade

[...]

PR #33038, opened 2024-07-10
Enhance Blazor 8.0 migration guidance

guardrex:
Addresses #32138

@mkArtakMSFT ... For https://github.com/dotnet/aspnetcore/issues/55678#issuecomment-2192302581 and the task of adding migration guidance, it's not clear from the discussion what a user would be migrating from or the conditions that create a problem for our base coverage. The dev seems to be asking about a specific custom approach that they took and a solution/workaround that Halter is giving them. What is the starting point for the migration scenario? I assume the solution to that problem is to call `Microsoft.AspNetCore.Builder.AuthorizationEndpointConventionBuilderExtensions.AllowAnonymous` with a correctly-configured `<AuthorizeRouteView>` in the `Routes` component.

@danroth27 For ...

> We changed our authentication guidance in .NET 8, but we didn't give detailed instructions on how to upgrade

Do you want migration guidance from Blazor Server with MS Identity Web to our BWA+OIDC non-BFF guidance? ... and noting that our BWA+OIDC guidance adopts global Interactive Auto rendering ... and noting further that we don't have BWA MS Identity Web guidance, only OIDC at this time. I mention that because the natural migration would be for MS Identity Web with Blazor Server to MS Identity Web with BWA; however, we don't have MS Identity Web BWA reference guidance. We'd be migrating to an app that we don't actually document. However again :smile:, it's strange to migrate away from MS Identity Web to OIDC if MS Identity Web is what the dev was using.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/host-and-deploy/webassembly-deployment-layout.md](https://github.com/dotnet/AspNetCore.Docs/blob/6a4e4dea552b26536e64efb8dcbbee653949d5b8/aspnetcore/blazor/host-and-deploy/webassembly-deployment-layout.md) | [Deployment layout for ASP.NET Core hosted Blazor WebAssembly apps](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/webassembly-deployment-layout?branch=pr-en-us-33038) |
| [aspnetcore/migration/70-80.md](https://github.com/dotnet/AspNetCore.Docs/blob/6a4e4dea552b26536e64efb8dcbbee653949d5b8/aspnetcore/migration/70-80.md) | [[Visual Studio for Mac](#tab/visual-studio-mac)](https://review.learn.microsoft.com/en-us/aspnet/core/migration/70-80?branch=pr-en-us-33038) |


danroth27:
> Do you want migration guidance from...

@guardrex We need to document how to upgrade existing .NET 7 Blazor apps that use auth to .NET 8. This includes both Blazor Server & Blazor WebAssembly apps that use ASP.NET Core Identity and the Microsoft Identity Platform (Entra). For Entra, @halter73 is planning to provide a sample in a few weeks. For apps that use ASP.NET Core Identity & Identity Server, I think we need guidance from @JeremyLikness.

guardrex:
I'll change this to an "addresses" PR and prepare to merge it after I hear back from @mkArtakMSFT on the other item ☝️ because it sounds like some of the auth targets for migration are still in the works. Later, I'll need clarification on the exact starting points.

Examples of scenarios that we didn't cover in the Blazor node:

* **Duende backend for Blazor WASM**, which I suppose is the 7.0 starting point for the 8.0 *Standalone with Identity* (`MapIdentityApi`) article+sample target.
* **Blazor Server+Duende**, which I suppose is the 7.0 starting point for the 8.0 *BWA+OIDC* (non-BFF?) article+sample.
* **Blazor Server+Entra**, which is obviously the 7.0 starting point for 8.0 BWA+Entra. IIRC, the Azure folks were going to put up samples for BWA+Entra and BWA+B2C. I see you're saying now that Halter will create the target (at least for Entra), so I'll just need to start with a Blazor Server+Entra app.

I think that WASM+Entra/B2C doesn't require more than package and TFM changes, which is already covered by the existing migration guidance.

guardrex:
I'll proceed with this in this form, and we can take up the other two items later on the issue and one or more new PRs.

PR #33061, opened 2024-07-12
Fix syntax error in openapi9.md

codedeliveryservice:
This PR removes an extra closing parenthesis to fix a syntax error.


guardrex (approved):
Thanks @codedeliveryservice! 🚀

PR #33098, opened 2024-07-16
Rename variable in async example code snippet 

gnerkus:
- rename variable in Index.cshtml.cs to indicate it represents a collection


guardrex:
Hello @gnerkus ... This can't be addressed this simply. For one thing, the property declaration also requires an update. More importantly, the article might make in-text references to this code and also require adjustment for the change.

Open a new issue from the bottom of the article using the **Open a documentation issue** link/feedback form. Use of the form adds metadata to your GitHub issue that cross-links the topic and pings the author automatically for a faster response. When the doc author responds (Wade I think), you can float your suggestion for this. If he approves the change, he'll direct you further from there on if the article also requires updates. If you're able to proceed with the update, you'll add "Fixes XXX" with the issue number to link the PR to the issue.

PR #33102, opened 2024-07-16
Clarify setting the resource path value

### LINKED ISSUE #33078: Misleading part about ResourcesPath
### Description

There's a note
> Localization is resource path that can be set via [LocalizationOptions](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.localization.localizationoptions).

In section [SharedResources](https://learn.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-8.0#shared-resources).

This is extremely misleading. =( It took me a while to udnerstand how things really work.

`ResourceManagerStringLocalizerFactory` has two [`Create`](https://github.com/dotnet/aspnetcore/blob/90a622d4f3f377dcf4eabbbbb8765ba9d3e65415/src/Localization/Localization/src/ResourceManagerStringLocalizerFactory.cs#L122) methods.
The one which accepts `Type` [is used inside `IStringLocalizer<T>`](https://github.com/dotnet/aspnetcore/blob/90a622d4f3f377dcf4eabbbbb8765ba9d3e65415/src/Localization/Abstractions/src/StringLocalizerOfT.cs#L26) which wraps the instance of `IStringLocalizer`

If you follow the code of `Create` method, you will see, that if `ResourcesPath` is not set, it [has no effect](https://github.com/dotnet/aspnetcore/blob/90a622d4f3f377dcf4eabbbbb8765ba9d3e65415/src/Localization/Localization/src/ResourceManagerStringLocalizerFactory.cs#L84) for `IStringLocalizer<T>`, that is resource files just need to be placed where `SharedResources` class is. Otherwise 'ResourePath` seem to be a path relative to the path of the class!
Look at [this comment](https://github.com/dotnet/aspnetcore/blob/90a622d4f3f377dcf4eabbbbb8765ba9d3e65415/src/Localization/Localization/src/ResourceManagerStringLocalizerFactory.cs#L88)

Here's how embedded resource is named in assembly:
image

Here's the part of the project:
image

And here's how class is defined:
```
namespace DutchVerbs.Spa.Xyz;

internal sealed class Strings
{
}
```

Look how name of the resource changes if I move resource files:
image

image

In this case I indeed need to set `ResourcesPath = "Resources"`

It also means that combining the usage of both `IStringLocalizer` and `IStringLocalizer<T>` when resources are placed differently won't work.

The absence of `Strings` class affects how resources get named (no `Xyz` namespace anymore):
image

image

IMO, Xmldoc (api reference) part of `Localization` package is missing a lot of details. We should also push the devs to add more.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/globalization-localization.md

### Document ID

d6f07538-228e-9f96-680f-6c324caf11d6

### Article author

@guardrex

guardrex:
Fixes #33078

Thanks @voroninp! 🚀 ... I'm not aware of how `IStringLocalizer` comes into the pattern, so I didn't remark on ...

> It also means that combining the usage of both `IStringLocalizer` and `IStringLocalizer<T>` when resources are placed differently won't work.

The section's approach only uses `IStringLocalizer<T>`. Do you want to suggest a line of text for the note on that? It can't state "combined usage of." It has to say which one isn't supported in this scenario, and it might add a remark on why.

cc: @hishamco ... Do you want to review?

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/globalization-localization.md](https://github.com/dotnet/AspNetCore.Docs/blob/31a2ef5da1a7cf4076e18953aca0b2f180a17a8a/aspnetcore/blazor/globalization-localization.md) | [ASP.NET Core Blazor globalization and localization](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?branch=pr-en-us-33102) |

### DIFF
--- a/aspnetcore/blazor/globalization-localization.md
+++ b/aspnetcore/blazor/globalization-localization.md
@@ -1618,7 +1618,7 @@ To create localization shared resources, adopt the following approach.
   * `Localization/SharedResource.es.resx`
 
   > [!NOTE]
-  > Use <xref:Microsoft.Extensions.Localization.LocalizationOptions.ResourcesPath%2A?displayProperty=nameWithType> to set a different path when shared resource files aren't in the same folder as the dummy class. **Supply a path relative to the location of the dummy class, not the root of the app.**
+  > Use <xref:Microsoft.Extensions.Localization.LocalizationOptions.ResourcesPath%2A?displayProperty=nameWithType> to set a different path for resource files that aren't in the same folder as the dummy class. When providing a resources path, make the path relative to the location of the dummy class. If you supply a resource path, you can't simultaneously use <xref:Microsoft.Extensions.Localization.IStringLocalizerFactory.Create%2A?displayProperty=nameWithType> to load resources.
 
 * To reference the dummy class for an injected <xref:Microsoft.Extensions.Localization.IStringLocalizer%601> in a Razor component, either place an [`@using`](xref:mvc/views/razor#using) directive for the localization namespace or include the localization namespace in the dummy class reference. In the following examples:
 
--- a/aspnetcore/blazor/globalization-localization.md
+++ b/aspnetcore/blazor/globalization-localization.md
@@ -1617,8 +1617,8 @@ To create localization shared resources, adopt the following approach.
   * `Localization/SharedResource.resx`
   * `Localization/SharedResource.es.resx`
 
-  > [!NOTE]
-  > Use <xref:Microsoft.Extensions.Localization.LocalizationOptions.ResourcesPath%2A?displayProperty=nameWithType> to set a different path for resource files that aren't in the same folder as the dummy class. When providing a resources path, make the path relative to the location of the dummy class. If you supply a resource path, you can't simultaneously use <xref:Microsoft.Extensions.Localization.IStringLocalizerFactory.Create%2A?displayProperty=nameWithType> to load resources.
+  > [!WARNING]
+  > When following the approach in this section, you can't simultaneously set <xref:Microsoft.Extensions.Localization.LocalizationOptions.ResourcesPath%2A?displayProperty=nameWithType> and use <xref:Microsoft.Extensions.Localization.IStringLocalizerFactory.Create%2A?displayProperty=nameWithType> to load resources.
 
 * To reference the dummy class for an injected <xref:Microsoft.Extensions.Localization.IStringLocalizer%601> in a Razor component, either place an [`@using`](xref:mvc/views/razor#using) directive for the localization namespace or include the localization namespace in the dummy class reference. In the following examples:
 
--- a/aspnetcore/blazor/globalization-localization.md
+++ b/aspnetcore/blazor/globalization-localization.md
@@ -1618,7 +1618,7 @@ To create localization shared resources, adopt the following approach.
   * `Localization/SharedResource.es.resx`
 
   > [!NOTE]
-  > `Localization` is resource path that can be set via <xref:Microsoft.Extensions.Localization.LocalizationOptions>.
+  > Use <xref:Microsoft.Extensions.Localization.LocalizationOptions.ResourcesPath%2A?displayProperty=nameWithType> to set a different path when shared resource files aren't in the same folder as the dummy class. **Supply a path relative to the location of the dummy class, not the root of the app.**
 
 * To reference the dummy class for an injected <xref:Microsoft.Extensions.Localization.IStringLocalizer%601> in a Razor component, either place an [`@using`](xref:mvc/views/razor#using) directive for the localization namespace or include the localization namespace in the dummy class reference. In the following examples:
 


voroninp:
In theory one can use `IStringLocalizerFactory` directly, to get the instance of IStringLocalizer.
I used this [method](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.localization.istringlocalizerfactory.create?view=net-8.0#microsoft-extensions-localization-istringlocalizerfactory-create(system-string-system-string))

For example, I can crete some folder and put resx files there (without defining a class - let's call them unbound).
The problem is that for these resources `ResourcesPath` is a path relative to the root of the project.
For bound resources which have a class `ResourcesPath` is a path relative to the file with the class.
If unbound resources placed in folder `Resources` then bound resources also should be put in `Resources` folder:
```
proj
|-- Resources
   | -- res1.resx
   | -- res2.resx
|-- Classes
   |--MyClass.cs
   | -- Resources
       | -- MyClass.resx      
```

Bu this won't work (because for either one or another `ResourcesPath` value will yield wrong final path.
(it must be `Resources`  for unbound resources, and empty for bound ones):
```
proj
|-- Resources
   | -- res1.resx
   | -- res2.resx
|-- Classes
   |-- MyClass.cs
   |-- MyClass.resx      
```

guardrex:
So, this ...

```
proj
|-- Resources
   |-- res1.resx
   |-- res2.resx
   |-- MyClass.cs
   |-- MyClass.resx
```

... would work, but it's an unnatural layout.

I'll have to think about it some more. It sounds to me like this is a poorly supported feature in .NET ... I mean that even having to have such a thing as a "dummy class" is weak design, and I noticed that the main doc set has it as well. I'll take another look at this later today.

voroninp:
No, this should not work.
`ResourcesPath` for `MyClass.resx` is still empty because it is colocated with `MyClass.cs`
But `ResourcesPath` is `Resources` for `res1.resx` and `res2.resx`.

voroninp:
You can play with this solution: https://github.com/voroninp/StringLocalizerAndResources
Change `ResorcesPath` between `"Resources"` and `""` - see the difference.

guardrex:
Cool ... Thanks for the repro. I'll pick back up with this tomorrow morning, as I'm out of time today.

aspnetcore/blazor/globalization-localization.md
```diff
@@ -1618,7 +1618,7 @@ To create localization shared resources, adopt the following approach.
   * `Localization/SharedResource.es.resx`
 
   > [!NOTE]
-  > `Localization` is resource path that can be set via <xref:Microsoft.Extensions.Localization.LocalizationOptions>.
+  > Use <xref:Microsoft.Extensions.Localization.LocalizationOptions.ResourcesPath%2A?displayProperty=nameWithType> to set a different path when shared resource files aren't in the same folder as the dummy class. **Supply a path relative to the location of the dummy class, not the root of the app.**
```
hishamco, line 1621: It's not necessary a shared resource, right?

guardrex: Yes, but this section only applies to shared resources, and the remark is only addressing the guidance provided here.

voroninp: I think there's no much difference between shared or non-shared resource.
What matters is where resx file is located relative to the class and whether class exists at all.

guardrex: It seems to me like a design problem, and I encourage someone ... not necessarily either of you ... to open a product unit issue on it. For now, I think I can get the article's remark correct. I'll try an update shortly.

guardrex:
@voroninp @hishamco ... See the latest update. Does that make sense, or do you have suggestions on how the passage should phrase the limitations?

guardrex:
I'm surprised that the factory's `location` can't be set to avoid this problem.

... and I see what you mean about at least some of the API remarks not providing much help. Describing the factory's `location` param as ...

> The location to load resources from.

... is just about a useless description.

guardrex:
Hold on a sec ... I think I just figured something out that might make sense. brb .................

guardrex:
Nah! It's ***crap*** 💩👎...

```
proj
|-- Resources
   |-- Resources
       |-- Foo.resx
   |-- res.resx
   |-- Foo.cs
```

... works if you provide a `ResourcesPath` of `Resources`.

image

Anyway ... let's just see if we can get the text correct for the limitations.

voroninp:
I do not have access to preview link, btw. But I assume, this is ok.

voroninp:
Yep, this wording at least warns about the caveats and locations of the files.

voroninp:
P.S. I am surprized they do not provide an altenrative implementation of `IStringLocalizer` for Blazor Wasm which would simply try to fetch static txt/json/toml file (just as an asset), cache the result it in the browser's local storage and use it instead of sattelite assemblies. The benefit would be that if language is not used its file is never downloaded.
If I correctly understand current behavior, all sattelite assemblies are fetched with the main app.

guardrex:
The preview link is for internal review (Microsoft folks). I asked them years ago why they wouldn't allow everyone to see the previews, but there was no response. All I got from them was 🦗 *crickets* 🦗.

I'll merge this now. Thanks again for your help with the article.

guardrex:
I've played with it a bit more, and I don't like this direction for the note.

When following the guidance in the section, it really doesn't matter where the shared resource dummy class and resource files are placed ***if the `ResourcePath` isn't set***. That's the base case IMO.

The limitation is that when following the guidance in this section one can't set the `ResourcePath` in order to use the factory approach. That's the caveat :smiling_imp: IMO.

I'll try another draft note in a sec .......................

guardrex:
To me, the following seems the most direct way to say it, given that the base case is the approach that the section is taking ...

> When following the approach in this section, you can't simultaneously set \<xref:Microsoft.Extensions.Localization.LocalizationOptions.ResourcesPath%2A?displayProperty=nameWithType> and use \<xref:Microsoft.Extensions.Localization.IStringLocalizerFactory.Create%2A?displayProperty=nameWithType> to load resources.

guardrex:
I'm going to go ahead and merge this because I'd like to get these remarks updated in the live doc ASAP 🏃‍♂️. Let me know if there's a problem, and we can talk about my placing a patch PR to update this further.

voroninp:
Thank you, the Guardian of the Docs!

PR #33110, opened 2024-07-18
ErrorBoundary coverage enhancements

### LINKED ISSUE #33109: ErrorBoundary approach for global exception handling with per-page/component rendering
### Description

Reported by @TimMurphy ...

> It's because of the way I think `<ErrorBoundary>` works that I started down the rabbit hole of using the alternative global exception handling section.
>
> I've created https://github.com/TimMurphy/ErrorBoundaryWithPerPageInteracvtity. It is BWA with per-page/component mode.
>
> In the generated /counter page I've wrapped the HTML with `<ErrorBoundary>`. Press the button 3 times and you'll see the `<ErrorBoundary>` has no effect.
>
> I've added a WorkingCounter page that works with `<ErrorBoundary>`. This solution involves having a page, ie WorkingCounter, with a `<ErrorBoundary>` and the actual content for the page in a separate file, ie. WorkingCounterComponent.
>
> Is this the correct way to use `<ErrorBoundary>`? Seems like more work than it should be.
>
> How I'm using per-page/component mode is the public side of the site SSR and the admin section, /admin, I'm using Interactive rendering. I'd like the entire site to be interactive rendering but from the documentation I've read this is not a good idea because it may require higher spec server and I'm trying to keep costs to a minimum and use the lowest server configuration Azure provides. Am I wrong in this approach?

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/handle-errors?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/fundamentals/handle-errors.md

### Document ID

66a0b1c2-c45c-98ca-9808-6f340a861c44

### Article author

@guardrex

guardrex:
Fixes #33109

Thanks @TimMurphy! ... again! Although you hit a pain point with the explanation on wrapping the component hierarchy, I'm glad that you wrote in about this section (and the other one). This is vastly improved now, and it should clear up the confusing about about what the error boundary should wrap.

When this goes live, I'll ping you here with the link to it.


guardrex:
## It's live ⚡ now ...

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/handle-errors?view=aspnetcore-8.0#error-boundaries

TimMurphy:
Thanks Luke for your quick responses. Not the answers I was hoping for but
at least I know why and how to proceed.

guardrex:
If the framework features are limiting what you'd like to do, open a new feature request with the product unit at ...

Perhaps, they'll consider your suggestion(s). First tho, search their open issues. I'm under the impression that they already know that the global exception handling experience is a bit rough. They might already be planning something for a future release. You can also look at their closed issues for any suggestions that they decided not to pursue.

TimMurphy:
Thanks, I'll check it out.

guardrex:
Sure thing ... and I forgot to mention that you filter the blazor issues with label:area-blazor. A search like this might find some things ...

https://github.com/dotnet/aspnetcore/issues?q=is%3Aissue+label%3Aarea-blazor+exception+handling

guardrex:
It turns things like this up, where there's a discussion that we were just having. Also, the revisions that I just put in should help devs understand and help with the confusion.

https://github.com/dotnet/aspnetcore/issues/56413

PR #33156, opened 2024-07-24
Enhanced JSImport/JSExport guidance

### LINKED ISSUE #32001: Additional JSImport/JSExport Examples
### Proposal and offer to produce a PR to include additional examples of using JSImport/JSExport.  

1) Optionally, I would propose creating a new article focused on .NET/JS Interop.   
   - Move the type mapping matrix to this new article
   - The idea being to factor out any information common between [JavaScript JSImport/JSExport interop with ASP.NET Core Blazor](https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/import-export-interop?view=aspnetcore-8.0) and [Run .NET from JavaScript](https://learn.microsoft.com/en-us/aspnet/core/client-side/dotnet-interop?view=aspnetcore-8.0)
   - As this new article potentially grows, it will serve to document JS interop based on System.Runtime.InteropServices.JavaScript that is agnostic of whether it is Blazor or non-Blazor.  Currently the Blazor specific pages are the only thing I get when googling "JS interop JSImport".  This can be improved by providing a generic document for JS interop.
   - The two technology specific articles would still cover setup, and keep some of their specialized minimal examples because they do a good job of tieing it all together and showing how the interop code fits into the application, such as the Blazor specific article's examples showing JSImport used inside a *.razor.cs page.

2) Regardless of the above decision point, my primary ask is to add additional concrete usage examples of JSImport:
  - Use of JSObject
    - Demonstrate calling a JS method to retrieve a JSObject reference, then calling another JSImport'd method and passing it.
    - Demonstrate calling a JS instance method by passing the JSObject to a static method that in turn calls the instance method on the object.
    - Accessing properties via the JSObject.GetPropertyAs* methods
  - Returning a Promise mapped to a Task, and awaiting it from C#
  - Subscribing to a click event to demonstrate passing a .NET Action<> listener to a JSImport'd method doing .bind/.addEventListener, and handling callbacks from JS to .NET.
    - Unsubscribing to an event, which highlights an important note about how passing the same .NET action produces different wrappers which will cause problems with methods like unsubscribe that depend on reference equality(the subscribe method needs to return a ref to the JSObject wrapping the Action for the later unsubscribe call to see it as the same event handler)

I would consider proposing some more advanced scenarios and possible troubleshooting section to address pitfalls for those new to JS interop, but there'd probably be some debate about some of those items so I'd defer to a future PR for the sake of getting agreement on the core ask here.  Also I didn't cover anything JSExport related.  Point being, I think there's room for it to grow.

Let me know a decision point on 1) and thoughts on 2), and I'd be happy to draft a PR.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/client-side/dotnet-interop?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/client-side/dotnet-interop.md

### Document ID

efff6954-a035-d1da-5975-af04ff7c05ed

### Article author

@pavelsavara

guardrex:
Fixes #32001

Fantastic article @SerratedSharp! 🎷

Notes:

* Code lines were adjusted to meet our 85 character line length limit.
* In a few spots, I changed "function" to "method" because it looked like the text was referring to calling the JS declaration on a JS object. As you review, file a suggestion for any line that should say "method" instead of "function." Check the remaining mentions of "function" to confirm that they're correct.
* In one of the code comments, there's an error message shown: "ToJS for System.Int64 is not implemented." What is "ToJS" there? 👂 
* The original client-side .NET interop article will need to stay on the PR because GH 💥😈 for some cryptic reason when I tried to merely move and rename the file. I'll investigate and deal with it separately before going live.
* I'm changing the prereqs for the Blazor article to VS, instead of saying the prereq is just the .NET SDK ... and I added that they should install the **.NET WebAssembly build tools** during a VS install. Let me know if I'm incorrect on that aspect. 👂 

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md](https://github.com/dotnet/AspNetCore.Docs/blob/872d722ea2772b765e373c8929c1c33ef66d2a42/aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md) | [Call JavaScript functions from .NET methods in ASP.NET Core Blazor](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-javascript-from-dotnet?branch=pr-en-us-33156) |
| [aspnetcore/blazor/javascript-interoperability/import-export-interop.md](https://github.com/dotnet/AspNetCore.Docs/blob/872d722ea2772b765e373c8929c1c33ef66d2a42/aspnetcore/blazor/javascript-interoperability/import-export-interop.md) | [JavaScript JSImport/JSExport interop with ASP.NET Core Blazor](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/import-export-interop?branch=pr-en-us-33156) |
| [aspnetcore/client-side/dotnet-interop.md](https://github.com/dotnet/AspNetCore.Docs/blob/872d722ea2772b765e373c8929c1c33ef66d2a42/aspnetcore/client-side/dotnet-interop.md) | [Run .NET from JavaScript](https://review.learn.microsoft.com/en-us/aspnet/core/client-side/dotnet-interop?branch=pr-en-us-33156) |
| [aspnetcore/client-side/dotnet-interop/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/872d722ea2772b765e373c8929c1c33ef66d2a42/aspnetcore/client-side/dotnet-interop/index.md) | [aspnetcore/client-side/dotnet-interop/index](https://review.learn.microsoft.com/en-us/aspnet/core/client-side/dotnet-interop/index?branch=pr-en-us-33156) |
| [aspnetcore/client-side/dotnet-interop/wasm-browser-app.md](https://github.com/dotnet/AspNetCore.Docs/blob/872d722ea2772b765e373c8929c1c33ef66d2a42/aspnetcore/client-side/dotnet-interop/wasm-browser-app.md) | [aspnetcore/client-side/dotnet-interop/wasm-browser-app](https://review.learn.microsoft.com/en-us/aspnet/core/client-side/dotnet-interop/wasm-browser-app?branch=pr-en-us-33156) |
| [aspnetcore/release-notes/aspnetcore-7.0.md](https://github.com/dotnet/AspNetCore.Docs/blob/872d722ea2772b765e373c8929c1c33ef66d2a42/aspnetcore/release-notes/aspnetcore-7.0.md) | [aspnetcore/release-notes/aspnetcore-7.0](https://review.learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-7.0?branch=pr-en-us-33156) |
| [aspnetcore/whats-new/dotnet-AspNetCore.Docs-mod0.md](https://github.com/dotnet/AspNetCore.Docs/blob/872d722ea2772b765e373c8929c1c33ef66d2a42/aspnetcore/whats-new/dotnet-AspNetCore.Docs-mod0.md) | [ASP.NET Core docs: What's new for June 2024](https://review.learn.microsoft.com/en-us/aspnet/core/whats-new/dotnet-AspNetCore.Docs-mod0?branch=pr-en-us-33156) |


SerratedSharp:
I'm looking at the "Files changed" tab on this PR and comparing to my original cannibalized article.  I imagine your looking for responses from reviewers, but for what it's worth here's a couple notes:

> * In one of the code comments, there's an error message shown: "ToJS for System.Int64 is not implemented." What is "ToJS" there? 👂

Something AFAIK that bubbles up from the [code generator](https://github.com/dotnet/runtime/blob/7d64842fdc41340440b00795c3f39f4c917acc62/src/libraries/System.Runtime.InteropServices.JavaScript/src/System/Runtime/InteropServices/JavaScript/Marshaling/JSMarshalerArgument.Object.cs#L132).  I wouldn't think devs are expected to know what ToJS("It's used by JSImport code generator and should not be used by developers in source code."), it's just one of those internal implementation details bubbling up through errors.   Wording wise it means that this specific handler doesn't support Int64, but a couple layers up where the consuming dev is sitting the wording sounds like a blanket statement.   Rather they simply need to map the type explicitly.

**Project template names**
I noticed this line was removed: "- Once you have installed the wasm-tools and wasm-experimental workloads, a new project template will be available in Visual Studio's new project dialog called "WebAssembly Browser App"." 

I don't object to this but maybe I'm bad at following directions, as I called this out in plain language because when I originally went through [Run .NET from JavaScript](https://learn.microsoft.com/en-us/aspnet/core/client-side/dotnet-interop?view=aspnetcore-8.0) it wasn't apparent to me what exactly I was getting from installing these.  I think at the time I was new "workloads" concept, it didn't register with me at first what I was getting from the workload, and that the project templates corresponded to the names in parenthesis.

Maybe consider changing this from the linked article:
> Optionally, install the `wasm-experimental` workload, which contains experimental project templates for getting started with .NET on WebAssembly in a browser app (WebAssembly Browser App) or in a Node.js-based console app (WebAssembly Console App).

To this:
> Optionally, install the `wasm-experimental` workload, which adds the experimental "WebAssembly Browser App" project template for getting started with .NET on WebAssembly in a browser app or the "WebAssembly Console App" project template in a Node.js-based console app.  After installing the workload, these new templates can be selected when creating a new project.

**CS filename**

image

- Typically filename would be "JSObjectInterop.cs" to align with classname (disregarding usage example).  I think Pavel had recommended the suffix \*Interop for the C# classname(I originally used \*Proxy).
- I think similar code examples on docs I used as reference had omitted filenames completely, because they are superfluous?  
- The only reason I included filenames for JS files is because it **must** line up the filename in ImportAsync calls(if that's the approach chosen to load the JS).  Less strictly by convention it is also the module name in JSImport, but that doesn't strictly need to match the filename.
- If you're wondering, the reason I made the choice not to bubble up the \*Shim suffix to the C# class name is as mentioned elsewhere in the article, you don't necessarily always need a JS shim.  You might be doing JSImport directly against pre-existing JS code.  The presence or lack of a shim would be an implementation detail hidden from C# consumers of \*Interop classes.  My naming conventions in this area could easily be debated, but that's my personal rational.  Because the framework is already generating "proxies" for managed types, etc. there's already layers in place using some of the terminology one might want to use instinctively, and it starts to become difficult to come up with distinct names for something that is another layer of glue between other layers of glue.

Similar suggestion applies to instances such as "`PromisesShim.cs`:"/"public partial class PromisesInterop". Filename should match classname or vice versa.

**JS Filename(s?) removed**
Along these lines I noticed at least one instance that the JS filename was removed, and again **if there's going to be filenames IMHO the *.js filenames are the important ones**

image

...because they align with this code:
image

That's the only things that jumped out at me.  

**Thanks for everyone's time in getting this integrated.**

guardrex:
Thanks ... I'll make updates. For the future, note that we use GH's suggestion feature for line changes, which is efficient for change requests on PRs.

MackinnonBuck (commented):
Just adding some thoughts - I think we should wait for a review from @pavelsavara before merging.

aspnetcore/blazor/javascript-interoperability/import-export-interop.md
```diff
@@ -22,15 +24,17 @@ Blazor provides its own JS interop mechanism based on the <xref:Microsoft.JSInte
...
 ## Obsolete JavaScript interop API
 
 Unmarshalled JS interop using <xref:Microsoft.JSInterop.IJSUnmarshalledRuntime> API is obsolete in ASP.NET Core in .NET 7 or later. Follow the guidance in this article to replace the obsolete API.
 
 ## Prerequisites
 
-[Download and install .NET 7 or later](https://dotnet.microsoft.com/download/dotnet) if it isn't already installed on the system or if the system doesn't have the latest version installed.
+[Visual Studio](https://visualstudio.microsoft.com/downloads/?utm_medium=microsoft&utm_source=learn.microsoft.com&utm_campaign=inline+link&utm_content=download+vs2022) with the **ASP.NET and web development** workload.
+
+When installing Visual Studio, select the **.NET WebAssembly build tools** option from the list of optional components.
```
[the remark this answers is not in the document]
guardrex: I placed it because the prereqs of the .NET JS article call it out ...

> Install the latest version of the [.NET SDK](https://dotnet.microsoft.com/download/dotnet/).
>
> Install the `wasm-tools` workload in an administrative command shell, which brings in the related MSBuild targets:
>
> ```dotnetcli
> dotnet workload install wasm-tools
> ```
>
> The tools can also be installed via Visual Studio's installer under the **ASP.NET and web development** workload in the Visual Studio installer. Select the **.NET WebAssembly build tools** option from the list of optional components.

I switched this article over to VS because I thought that would be the most common development approach for Blazor apps here, but I can add a tab control to hold both the VS and CLI guidance.

aspnetcore/client-side/dotnet-interop/index.md
```diff
@@ -0,0 +1,950 @@
...
+
+In `Program.Main`, `ConsoleLog` is called with the message to log:
+
+```csharp
+GlobalInterop.ConsoleLog("Hello World!");
+```
+
+The output appears in the browser's console.
+
+The following demonstrates importing a static method declared in JS.
```
guardrex, line 61:
```suggestion
The following demonstrates importing a method declared in JS.
```

aspnetcore/client-side/dotnet-interop/index.md
```diff
@@ -0,0 +1,950 @@
...
+
+The following custom static method (`globalThis.callAlert`) spawns an [alert dialog (`window.alert`)](https://developer.mozilla.org/docs/Web/API/Window/alert) with the message passed in `text`:
```
guardrex, line 63:
```suggestion
The following method (`globalThis.callAlert`) spawns an [alert dialog (`window.alert`)](https://developer.mozilla.org/docs/Web/API/Window/alert) with the message passed in `text`:
```

aspnetcore/client-side/dotnet-interop/index.md
```diff
@@ -0,0 +1,950 @@
...
+  // Primitives.GetCounter = function () { return counter; };
+
+  // Takes a parameter and returns nothing. JS doesn't restrict the parameter type, 
+  // but we can restrict it in the .NET proxy, if desired.
+  PrimitivesShim.LogValue = (value) => { console.log(value); };
+
+  // Called for various .NET types to demonstrate mapping to JS primitive types.
+  PrimitivesShim.LogValueAndType = (value) => { console.log(typeof value, value); };
+
+})(PrimitivesShim);
```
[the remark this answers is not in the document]
pavelsavara: it creates a namespace object. But let's get rid of it.

aspnetcore/client-side/dotnet-interop/index.md
```diff
@@ -0,0 +1,948 @@
...
+The preceding example displays the following output in the browser's debug console:
+
+> :::no-loc text="Date: Sat Dec 21 1968 07:51:00 GMT-0500 (Eastern Standard Time)":::  
+> :::no-loc text="Date: Sun Dec 22 1968 07:51:00 GMT-0500 (Eastern Standard Time)":::
+
+## JS object references
+
+Whenever a JS method returns an object reference, it's represented in .NET as a <xref:System.Runtime.InteropServices.JavaScript.JSObject>. The original JS object continues its lifetime within the JS boundary, while .NET code can access and modify it by reference through the <xref:System.Runtime.InteropServices.JavaScript.JSObject>. While the type itself exposes a limited API, the ability to hold a JS object reference and return or pass it across the interop boundary enables support for several interop scenarios.
+
+The <xref:System.Runtime.InteropServices.JavaScript.JSObject> provides methods to access properties, but it doesn't provide direct access to instance methods. As the following `Summarize` method demonstrates, instance methods can be accessed indirectly by implementing a static method that takes the instance as a parameter.
```
pavelsavara, line 387: this is useful pattern, thanks for documenting it.

aspnetcore/client-side/dotnet-interop/index.md
```diff
@@ -0,0 +1,948 @@
...
+The preceding example displays the following output in the browser's debug console:
+
+> :::no-loc text="Waited 2.0s.":::  
+> :::no-loc text="Waited .5s for WaitGetString: 'String From Resolve'":::  
+> :::no-loc text="Waited .5s for WaitGetDate: '11/24/1988 12:00:00 AM'":::  
+> :::no-loc text="responseText.Length: 582":::  
+> :::no-loc text="Waited 2.0s for AsyncFunction.":::  
+> :::no-loc text="JS Exception Caught: 'Reject: ShouldSucceed == false'":::
+
+## Subscribing to JS events
```
pavelsavara, line 644: somewhere we should mention that interacting with individual properties of DOM via JS interop will be slow and create lot of proxies (GC pressure). This pattern it not recommended. I'm voting for not including it in the official docs, it could be on somebody's blog instead.

pavelsavara: Now I see the perf section below. Maybe we can put the perf section first and keep this DOM section after it. With a note that dealing with more than handful of elements via JS interop is bad pattern.

aspnetcore/blazor/javascript-interoperability/import-export-interop.md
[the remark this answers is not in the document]
guardrex: @pavelsavara ... When using VS (for the Blazor Import-Export article), is the dev required to install the **.NET WebAssembly build tools**? ... because we say in the non-Blazor (client-side article) that it is required to '[bring] in the related MSBuild targets.'

aspnetcore/client-side/dotnet-interop/index.md
```diff
@@ -0,0 +1,948 @@
...
+  DateShim.IncrementDay = function (date) {
+    date.setDate(date.getDate() + 1);
+    return date;
+  };
+
+  DateShim.LogValueAndType = (value) => {
+    if (value instanceof Date) 
+      console.log("Date:", value)
+    else
+      console.log("Not a Date:", value)
```
guardrex, line 336: Is this the fix?

```suggestion
      console.log("Date:", value)
```

aspnetcore/client-side/dotnet-interop/index.md
```diff
@@ -0,0 +1,948 @@
...
+* `Array`: Marshaling an array creates a copy of the array in JS or .NET.
+* `MemoryView`
+  * `MemoryView` is a JS class for the .NET WebAssembly runtime to marshal `Span` and `ArraySegment`.
+  * Unlike marshaling an array, marshaling a `Span` or `ArraySegment` doesn't create a copy of the underlying memory.
+  * `MemoryView` can only be properly instantiated by the .NET WebAssembly runtime. Therefore, it isn't possible to import a JS method as a .NET method that has a parameter of `Span` or `ArraySegment`.
+  * `MemoryView` created for a `Span` is only valid for the duration of the interop call. As `Span` is allocated on the call stack, which doesn't persist after the interop call, it isn't possible to export a .NET method that returns a `Span`.
+  * `MemoryView` created for an `ArraySegment` survives after the interop call and is useful for sharing a buffer. Calling `dispose()` on a `MemoryView` created for an `ArraySegment` disposes the proxy and unpins the underlying .NET array. We recommend calling `dispose()` in a `try-finally` block for `MemoryView`.
```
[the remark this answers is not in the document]
guardrex: "We recommned" is valid per the MS style manual, and we use it a fair amount.

aspnetcore/client-side/dotnet-interop/index.md
```diff
@@ -0,0 +1,948 @@
...
+
+```csharp
+await JSHost.ImportAsync("ExampleShim", "/scripts/ExampleShim.js");
+```
+
+Depending on the platform that's loading WebAssembly, a dot-prefixed URL, such as `./scripts/`, might refer to an incorrect subdirectory, such as `/_framework/scripts/`, because the WebAssembly package is initialized by framework scripts under `/_framework/`. In that case, prefixing the URL with `../scripts/` refers to the correct path. Prefixing with `/scripts/` works if the site is hosted at the root of the domain. A typical approach involves configuring the correct base path for the given environment with an HTML `<base>` tag and using the `/scripts/` prefix to refer to the path relative to the base path. Tilde notation `~/` prefixes aren't supported by <xref:System.Runtime.InteropServices.JavaScript.JSHost.ImportAsync%2A?displayProperty=nameWithType>.
+
+> [!IMPORTANT] 
+> If JS is loaded from an ES6 module, then `[JSImport]` attributes must include the module name as the second parameter. For example, `[JSImport("globalThis.callAlert", "ExampleShim")]` indicates the imported method was declared in an ES6 module named "`ExampleShim`."
```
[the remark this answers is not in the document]
guardrex: What's the exact change request?

aspnetcore/client-side/dotnet-interop/index.md
```diff
@@ -0,0 +1,913 @@
...
+
+`DateShim.js`:
+
+```javascript
+DateShim.IncrementDay = function (date) {
+  date.setDate(date.getDate() + 1);
+  return date;
+};
+
+DateShim.LogValueAndType = (value) => {
```
guardrex, line 320: This isn't logging the type as the name implies. Should it just be the `LogValue` method?

aspnetcore/client-side/dotnet-interop/index.md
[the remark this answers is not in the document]
SerratedSharp: **Is there better terminology that would still distinguish the _specific_ module pattern needed for ImportAsync to work successfully?**  As far as I could tell ImportAsync will not work unless I declared the *.js as what I've always seen distinguished from other module patterns such as AMD as an "ES6 module", but I don't doubt it might be the wrong terminology.  Point being, ImportAsync won't happily accept anything, and it would be nice to convey in some way which **specific** module pattern it requires.  I find it a little frustrating when documentation uses the term "module" without specificity, only to find it is more specific than implied.  It could be argued that there's really only one true "module" syntax in official JS, but it kind of disregards the reality of the ecosystem of module patterns.  You will find other documentation that uses the term "module", but is implicitly referring to another module pattern that was the more pervasive pattern at the time, and for the same reason it can be a hurdle/time sink without knowing the specificity.

aspnetcore/client-side/dotnet-interop/index.md
pavelsavara: yes

aspnetcore/client-side/dotnet-interop/index.md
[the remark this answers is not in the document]
pavelsavara: my point was that it only requires a second parameter, when you loaded it via `JSHost.ImportAsync`
The fact that it needs to be ES6 in order to be loaded via `JSHost.ImportAsync` is a separate (useful) information.

aspnetcore/client-side/dotnet-interop/index.md
```diff
@@ -0,0 +1,948 @@
...
+  DateShim.IncrementDay = function (date) {
+    date.setDate(date.getDate() + 1);
+    return date;
+  };
+
+  DateShim.LogValueAndType = (value) => {
+    if (value instanceof Date) 
+      console.log("Date:", value)
+    else
+      console.log("Not a Date:", value)
```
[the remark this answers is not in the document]
SerratedSharp: I also felt the never hit `else` here is odd. I had maybe a silly notion of consistently demonstrating the type marshalling on the JS side, since you don't declare types in JS function signatures it's not as evident. I didn't want there to be any question that maybe the value was just a string serialization which could produce the same output. Unfortunately typeof doesn't work for non primitive types.

guardrex:
I still need an answer on if the **.NET WebAssembly build tools** option must be included ***for Visual Studio*** use (the Blazor ariticle). The client-side article states (emphasis added) ...

> Install the `wasm-tools` workload in an administrative command shell, which ***brings in the related MSBuild targets***:
>
> ```dotnetcli
> dotnet workload install wasm-tools
> ```
>
> The tools can also be installed via Visual Studio's installer under the **ASP.NET and web development** workload in the Visual Studio installer. Select the **.NET WebAssembly build tools** option from the list of optional components.

***Brings in the related MSBuild targets*** implies that devs need the tools for the Blazor article experience with VS.

aspnetcore/client-side/dotnet-interop/index.md
```diff
@@ -0,0 +1,910 @@
...
+
+The following example demonstrates `[JSImport]` leveraging type mappings of several primitive JS types and the use of [`JSMarshalAs`](xref:System.Runtime.InteropServices.JavaScript.JSMarshalAsAttribute%601), where explicit mappings are required at compile time.
+
+`PrimitivesShim.js`:
+
+```javascript
+globalThis.counter = 0;
+
+// Takes no parameters and returns nothing.
+PrimitivesShim.IncrementCounter = function () {
```
pavelsavara, line 193: the syntax is typically in JS with lowerCase first letter
`export function incrementCounter() {`

which means you need to also drop `PrimitivesShim` and lowerCase it in JSImport
[JSImport("incrementCounter", "PrimitivesShim")]

Please test it all after you are done with updates.

guardrex: @SerratedSharp ... I'll make the updates, but can you test? I'm completely buried in other high priority work ⛰️⛏️😅. I didn't test any of the code for the new article. If testing has to wait for me to get free, this PR will be held up for a while.

aspnetcore/client-side/dotnet-interop/index.md
```diff
@@ -613,34 +613,34 @@ The preceding example displays the following output in the browser's debug conso
...
 // Return a JSObject reference to a JS number array.
-[JSImport("PromisesShim.WaitGetArray", "PromisesShim")]
+[JSImport("waitGetArray", "PromisesShim")]
 [return: JSMarshalAs<JSType.Promise<JSType.Object>>()]
 public static partial Task<JSObject> WaitGetIntArrayAsObject();
 
 // Takes a JSObject reference to a JS number array, and returns the array as a C# 
 // int array.
-[JSImport("PromisesShim.WaitGetArray", "PromisesShim")]
+[JSImport("waitGetArray", "PromisesShim")]
```
guardrex, line 643: Should this be ...

```suggestion
[JSImport("unwrapJSObjectAsIntArray", "PromisesShim")]
```

guardrex: @SerratedSharp ... Could you check this one and see if my suggestion is correct or not?

Otherwise, we're just waiting on final checks of all of the code examples here. If you don't reach it, I hope to have time next week to do it.

guardrex: @SerratedSharp ... Still need to know if that update :point_up: is correct, and I have two questions. The article states ...

> If JS is loaded from a JavaScript module, then `[JSImport]` attributes must include the module name as the second parameter. For example, `[JSImport("globalThis.callAlert", "ExampleShim")]` indicates the imported method was declared in a JavaScript module named "`ExampleShim`."

... and that's the pattern that all of the examples adopt.

However, that's not what the project template does. The project template names the module ***with the file extension*** ...

```csharp
[JSImport("window.location.href", "main.js")]
```

I noticed if I change that to ...

```csharp
[JSImport("window.location.href", "main")]
```

No dice ... 💥. Should all of the module file parameter values be changed to include the file extension? For example ...

```diff
- [JSImport("incrementCounter", "PrimitivesShim")]
+ [JSImport("incrementCounter", "PrimitivesShim.js")]
```

Next question ... I hit a 💥 on my first test run with the Primatives Interop example:

> Loading module from “http://127.0.0.1:52460/PrimitivesShim.js” was blocked because of a disallowed MIME type (“”).

How do I resolve that?

SerratedSharp (changes requested):

guardrex:
Thanks @SerratedSharp ... Sorry that I didn't see your remark from three days ago. Thanks for providing the updates. I merged all of them except for the one that you're discussing with @pavelsavara on adding back "ES6," which I'll commit if he's cool with it.

Then ... I'll pick back up with testing within the next couple of days, hopefully tomorrow morning 🤞.

aspnetcore/client-side/dotnet-interop/index.md
SerratedSharp: > ... Still need to know if that update ☝️ is correct

It is correct but there was a similar issue in the same file. Not sure if my comments were not appearing.  Only just realized my reviews were probably only visible to me.  Hopefully you see several suggested changes that covers this issue as well as others.

> Should all of the module file parameter values be changed to include the file extension? 

This will not work unless you also change the ImportAsync calls to include the extension in the module name.

The project template uses a file extension in JSImport because it actually names its module with *.js in the module name when calling `setModuleImports` (I personally do not use this method for declaring modules because it isn't present in other .NET WASM platforms that support JSImport):

```JS
setModuleImports('main.js', {
    window: {
        location: {
            href: () => globalThis.window.location.href
        }
    }
});
```

guardrex: When a suggestion commit is made, GH hides the comment and marks it "resolved." I merged all of your suggestions, so I think we're good here.

guardrex: Seems nicer to not use the file extension ...

```diff
- await JSHost.ImportAsync("PrimitivesShim.js", "/PrimitivesShim.js");
+ await JSHost.ImportAsync("PrimitivesShim", "/PrimitivesShim.js");
```

...

```diff
- [JSImport("incrementCounter", "PrimitivesShim.js")]
+ [JSImport("incrementCounter", "PrimitivesShim")]
```

... but *Yes!* It's working here now. 🎉

guardrex: I placed the sample app in the Blazor samples repo, which makes sense because devs are going to want to play with this code ...

https://github.com/dotnet/blazor-samples/tree/main/8.0/WASMBrowserAppImportExportInterop

I'll cross-link the code into the article next.

I hit a new 😈 ... it's in the Events Usage bits ...

> MONO_WASM: ManagedError@http://127.0.0.1:60662/_framework/dotnet.runtime.8.0.7.s7d13jzdw4.js:3:26842
Jn@http://127.0.0.1:60662/_framework/dotnet.runtime.8.0.7.s7d13jzdw4.js:3:31614
Tl<@http://127.0.0.1:60662/_framework/dotnet.runtime.8.0.7.s7d13jzdw4.js:3:182417
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[5]:0x0
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[349]:0x1fad7
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[245]:0x1bf9f
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[238]:0xf16c
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[306]:0x1e7f1
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[327]:0x1efe7
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[217]:0xcfbc
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[772]:0x44213
Bl/e[r]@http://127.0.0.1:60662/_framework/dotnet.runtime.8.0.7.s7d13jzdw4.js:3:215551
Go@http://127.0.0.1:60662/_framework/dotnet.runtime.8.0.7.s7d13jzdw4.js:3:50974
callUserCallback@http://127.0.0.1:60662/_framework/dotnet.native.8.0.7.9cskr7qo55.js:8:112495
createDotnetRuntime/safeSetTimeout/createDotnetRuntime<@http://127.0.0.1:60662/_framework/dotnet.native.8.0.7.9cskr7qo55.js:8:112626
setTimeout handler*safeSetTimeout@http://127.0.0.1:60662/_framework/dotnet.native.8.0.7.9cskr7qo55.js:8:112588
Tl<@http://127.0.0.1:60662/_framework/dotnet.runtime.8.0.7.s7d13jzdw4.js:3:155252
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[771]:0x441af
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[349]:0x1fad7
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[245]:0x1bf9f
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[238]:0xf16c
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[272]:0x1d1bd
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[3185]:0xe8756
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[2505]:0xbe4ac
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[2511]:0xbecd0
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[2535]:0xc1327
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[128]:0xa4fa
createDotnetRuntime/Module._mono_wasm_invoke_method_bound@http://127.0.0.1:60662/_framework/dotnet.native.8.0.7.9cskr7qo55.js:8:123345
kr@http://127.0.0.1:60662/_framework/dotnet.runtime.8.0.7.s7d13jzdw4.js:3:35365
Nl/e.onRuntimeInitialized/l.javaScriptExports.complete_task@http://127.0.0.1:60662/_framework/dotnet.runtime.8.0.7.s7d13jzdw4.js:3:209726
mo/<@http://127.0.0.1:60662/_framework/dotnet.runtime.8.0.7.s7d13jzdw4.js:3:42372
promise callback*mo@http://127.0.0.1:60662/_framework/dotnet.runtime.8.0.7.s7d13jzdw4.js:3:42203
Yr/<@http://127.0.0.1:60662/_framework/dotnet.runtime.8.0.7.s7d13jzdw4.js:3:39743
Tl</S<@http://127.0.0.1:60662/_framework/dotnet.runtime.8.0.7.s7d13jzdw4.js:3:178043
Tl<@http://127.0.0.1:60662/_framework/dotnet.runtime.8.0.7.s7d13jzdw4.js:3:179198
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[3]:0x0
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[349]:0x1fafb
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[245]:0x1bf9f
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[238]:0xf16c
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[272]:0x1d1bd
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[3185]:0xe8756
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[2505]:0xbe4ac
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[2511]:0xbecd0
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[2535]:0xc1327
@http://127.0.0.1:60662/_framework/dotnet.native.wasm:wasm-function[128]:0xa4fa
createDotnetRuntime/Module._mono_wasm_invoke_method_bound@http://127.0.0.1:60662/_framework/dotnet.native.8.0.7.9cskr7qo55.js:8:123345
kr@http://127.0.0.1:60662/_framework/dotnet.runtime.8.0.7.s7d13jzdw4.js:3:35365
Nl/e.onRuntimeInitialized/l.javaScriptExports.complete_task@http://127.0.0.1:60662/_framework/dotnet.runtime.8.0.7.s7d13jzdw4.js:3:209726
mo/<@http://127.0.0.1:60662/_framework/dotnet.runtime.8.0.7.s7d13jzdw4.js:3:42652

SerratedSharp: @guardrex Not sure if source of error, but this code looks for btn1 and btn2 elements.  Probably needs a `// Assume button elements with IDs "btn1" and "btn2" exist on page`.  See this html (btn3 is not needed for the MS article):

https://github.com/SerratedSharp/CSharpWasmRecipes/blob/serrated/jsimport-msarticle/WasmBrowser.Recipes.WasmClient/wwwroot/index.html

guardrex:
@SerratedSharp ... That licked it 🎉 ...

https://github.com/dotnet/blazor-samples/pull/324

Anything else for this?

pavelsavara:
@maraf any feedback ? See also https://github.com/dotnet/blazor-samples/tree/main/8.0/WASMBrowserAppImportExportInterop

pavelsavara:
Unrelated to this PR, I'm thinking that Blazor users should be nudged to use workload and compatible the [flags/features](https://github.com/dotnet/runtime/blob/main/src/mono/wasm/features.md) which improve performances/size. Are the samples the right place ? 

Should we demo Blazor compatible things like `<InvariantTimezone>true</InvariantTimezone>` and `<InvariantGlobalization>true</InvariantGlobalization>` there ?

Should there be AOT demo ?

guardrex:
`<InvariantGlobalization>` is covered in the [glob/loc article](https://learn.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-8.0#invariant-globalization).

`<InvariantTimezone>` can be covered near it, so I've opened #33292 to address it.

Not necessarily an AOT "demo" (i.e., sample app), but we do have [AOT coverage](https://learn.microsoft.com/en-us/aspnet/core/blazor/webassembly-build-tools-and-aot?view=aspnetcore-8.0).

pavelsavara:
I was thinking sample, maybe one of the existing samples could have it all enabled ? No pressure, just idea.

guardrex:
I'm not feel'in it because ...

* It requires the tooling install, and I'd like the existing samples to ✨***Just Work!***&trade;✨ OOB for the scenarios that they cover.
* After the tooling is installed, the changes to an app to adopt AOT are trivial and don't demonstrate anything if done to an existing app or new app unless presented SxS with code to enable/disable AOT (or yet another sample) that would run slowly without AOT.
* AFAIK, DR doesn't want anything to adopt AOT OOB. That's my vague recollection from discussing how AOT would be covered when it was first introduced at 6.0.

pavelsavara:
>     It requires the tooling install, and I'd like the existing samples to ✨_**Just Work!**_™✨ OOB for the scenarios that they cover.

That's DX of a lazy developer ;)
If you care about UX of your app users, you could reconsider.

>enable/disable AOT (or yet another sample) that would run slowly without AOT.

AOT is faster in execution, but much slower for dev loop or first download. That's something that our developers need to understand (and experience).


>    AFAIK, DR doesn't want anything to adopt AOT OOB. 

Correct, it's not good as default, you need to know what you are doing.

guardrex:
@danroth27 will have to address it. It's not really a 🦖 call. Dan, see the last few remarks.

guardrex:
@maraf ... Do you want to review this?

@danroth27 ... Do you want to reverse your earlier decision on ***not*** enabling AOT OOB for one or more of our WASM sample apps? Even if you do want to reverse that decision, it would only be for the [`WASMBrowserAppImportExportInterop` sample](https://github.com/dotnet/blazor-samples/tree/main/8.0/WASMBrowserAppImportExportInterop) on this PR. If you want to reverse the decision for the other WASM samps, I'll open a new issue to work on them (and their coverage) separately.

aspnetcore/blazor/javascript-interoperability/import-export-interop.md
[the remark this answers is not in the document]
maraf: For non-Blazor scenario, wasm build tools were required in .NET 7 (because that was how we shipped the SDK build non-Blazor wasm app). In .NET 8 the Wasm SDK is present-by-default in .NET SDK and thus wasm build tools are not a requirement for non-Blazor as well. Anyway without `wasm-experimental` workload (I'm not sure if that is visible in VS installation) users won't have the `wasmbrowser` to create the app

danroth27:
> @danroth27 ... Do you want to reverse your earlier decision on not enabling AOT OOB for one or more of our WASM sample apps? Even if you do want to reverse that decision, it would only be for the [WASMBrowserAppImportExportInterop sample](https://github.com/dotnet/blazor-samples/tree/main/8.0/WASMBrowserAppImportExportInterop) on this PR. If you want to reverse the decision for the other WASM samps, I'll open a new issue to work on them (and their coverage) separately.

I don't see a need to have a sample for AOT given that the only code change needed to enable AOT is setting the `RunAOTCompilation` property.

guardrex:
Thanks, @danroth27.

I'm just waiting on feedback from @maraf. I think we can merge this after his review/updates.

maraf (approved):

PR #33244, opened 2024-08-02
new CODEOWNERS file

tdykstra:
Fixes #33160

PR #33262, opened 2024-08-05
Update CODEOWNERS

tdykstra:
Follow-up to #33244

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,33 +5,33 @@
 
 # The '*' pattern is global owners.
 
-*     
+# *     
 
-.github/  
+# .github/  
 
 # Order is important. The last matching pattern has the most precedence.
 
 /aspnetcore/blazor/ @guardrex
-/aspnetcore/breadcrumb/ 
-/aspnetcore/client-side/ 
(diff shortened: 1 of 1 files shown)


wadepickett (approved):
I can say the file is valid.  It is not real clear to me what the intended outcome goal is for review requests processes though.  It felt like a step back to obscure manual intentional requests with automatic ones.  A bit better this way divided by folder, but hopefully we can figure out something a little better later.

tdykstra:
@Rick-Anderson @wadepickett If you want to go back to fully manual review requesting I'm open to dispensing with the CODEOWNERS file.  I think it will work better now with only one owner for most of the folders with owners, but maybe even then it may not be giving us benefits worth the trouble.

wadepickett:
I would like to get back to deliberate manual review requests.  I am also fine with testing out something new for other benefits followed by a quick evaluation after we had a chance to try it out.

Rick-Anderson:
> @Rick-Anderson Do you want to ditch CODEOWNERS too?

I don't have a strong opinion. Maybe try it out with _a few with only one owner for most of the folders with owners_

Feel free to ditch it.

PR #33269, opened 2024-08-06
Removed double space in object initialization

aleksvujic:
Removed double space in `SelectListItem` object initialization.

Before:
```cs
new SelectListItem { Value = "US", Text = "USA"  }, // two spaces after "USA"
```

After:
```cs
new SelectListItem { Value = "US", Text = "USA" }, // one space after "USA"
```


guardrex (approved):
Thanks @aleksvujic! 🚀

Issue #33312, opened 2024-08-09
What's new: Data Protection adds support for deleting keys


tdykstra:
### Description

- [x]  Add a section to What's new covering [Data Protection adds support for deleting keys](https://github.com/dotnet/AspNetCore.Docs/issues/33031#issuecomment-2274532091)
- [x]   Create an issue for updating the permanent documentation, and then do that
- [ ] Hook up to WN9 `[!INCLUDE[](~/release-notes/aspnetcore-9/includes/delete_keys.md)]`

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-9.0?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/release-notes/aspnetcore-9.0.md

### Document ID

4e75ad25-2c3f-b28e-6a91-ac79a9c683b6

### Article author

@Rick-Anderson

---
[Associated WorkItem - 295447](https://dev.azure.com/msft-skilling/Content/_workitems/edit/295447)

PR #33324, opened 2024-08-11
delete keys for wn

Rick-Anderson:
Contributes to #33312

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/release-notes/aspnetcore-9.0.md](https://github.com/dotnet/AspNetCore.Docs/blob/72091180fa9205dbb46e7fa696a352c6573133b0/aspnetcore/release-notes/aspnetcore-9.0.md) | [What's new in ASP.NET Core 9.0](https://review.learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-9.0?branch=pr-en-us-33324) |


aspnetcore/security/data-protection/configuration/samples/9/deleteKeys/Program.cs
```diff
@@ -0,0 +1,23 @@
+using Microsoft.AspNetCore.DataProtection.KeyManagement;
+
+var builder = WebApplication.CreateBuilder(args);
+
+var app = builder.Build();
+
+app.MapDelete("/api/keys/cleanup-old-keys", (IServiceProvider services) =>
```
amcasey, line 7: Just to be explicit, exposing an endpoint for this would be an extremely bad idea.

aspnetcore/release-notes/aspnetcore-9/includes/delete_keys.md
```diff
@@ -0,0 +1,9 @@
+<!--
+[!INCLUDE[](~/release-notes/aspnetcore-9/includes/delete_keys.md)]
+-->
+
+### Data Protection support for deleting keys
+
+Previous to .NET 9, data protection keys were ___not___ deletable by design, to prevent data loss. Deleting a key renders its protected data inaccessible. Given their small size, the accumulation of these keys generally posed minimal impact. However, to accommodate extremely long-running services, we have introduced the option to delete keys. Generally, only old keys should be deleted. Only delete keys when you can accept the risk of data loss in exchange for storage savings. We recommend data protection keys should ___not___ be deleted.
```
amcasey, line 7:
```suggestion
Prior to .NET 9, data protection keys were ___not___ deletable by design, to prevent data loss. Deleting a key renders its protected data irretrievable. Given their small size, the accumulation of these keys generally posed minimal impact. However, to accommodate extremely long-running services, we have introduced the option to delete keys. Generally, only old keys should be deleted. Only delete keys when you can accept the risk of data loss in exchange for storage savings. We recommend data protection keys should ___not___ be deleted.
```

I don't feel strongly about it.

aspnetcore/security/data-protection/configuration/samples/9/deleteKeys/Program.cs
amcasey: The expected use case is in an out-of-band tool with no internet exposure.  One can imagine building an administration interface with this sort of endpoint, but I'd want to be very clear in the sample that this would only be accessible within a cluster or behind auth (at least).

Rick-Anderson: > The expected use case is in an out-of-band tool with no internet exposure. One can imagine building an administration interface with this sort of endpoint, but I'd want to be very clear in the sample that this would only be accessible within a cluster or behind auth (at least).

I deleted the project, the surrounding code (the endpoint). We like having samples that compile for obvious validation and maintenance reasons. In this case a snippet is justified. But I'm thinking dev's who need this can figure it out without a sample. We do know customers copy/paste code we right, and that seems to be a bigger risk.

Should I just delete the remaining code>

amcasey: I can appreciate the desire to have compiling code.  One way to accomplish that would be to write a console app that connects to an existing key repository and then deletes some keys.  Alternatively, I think the fragment is a reasonable compromise.

amcasey (approved):

PR #33346, opened 2024-08-13
Expand images coverage to include documents

### LINKED ISSUE #31948: Streaming sample works even for other document types (like PDF) with a small change
### Description

I recently had a conversion with blazor customer who has a need to preview other document types (like PDF). The documentation sample for streaming images works for other document types as well with two changes

- (expectedly) replace `img` tag with `iframe` or `embed`
- (this is hidden gem) provide content type when creating the `Blob` in javascript

I have put it together and created a sample https://github.com/maraf/blazor-file-stream.

I'm opening this issue as a consideration for mentioning pain-points in more general usage or providing the more general sample 

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/images?view=aspnetcore-8.0#stream-image-data

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/images.md

### Document ID

66bb1712-60ac-77a6-10d2-f2d795126783

### Article author

@guardrex

guardrex:
Fixes #31948

Thanks @maraf! 🎷

@blowdart ... TL;DR: This uses an `<iframe>` to display a trusted document from the `wwwroot` folder of an app. I'm including a passing CAUTION remark. Do you have any change suggestions for it?

> [!CAUTION]
> Use of the `<iframe>` element in the following example is safe and doesn't require \[sandboxing](https://developer.mozilla.org/docs/Web/HTML/Element/iframe#sandbox) because content is loaded from the app, a trusted source.
>
> When loading content from an untrusted source or user input, an improperly implemented `<iframe>` element risks creating security vulnerabilities.

### DIFF
--- a/.openpublishing.redirection.json
+++ b/.openpublishing.redirection.json
@@ -1307,6 +1307,11 @@
             "source_path": "aspnetcore/blazor/http-caching-issues.md",
             "redirect_url": "/aspnet/core/blazor/host-and-deploy/webassembly-caching/http-caching-issues",
             "redirect_document_id": false
+        },
+        {
+            "source_path": "aspnetcore/blazor/images.md",
+            "redirect_url": "/aspnet/core/blazor/images-and-documents",
+            "redirect_document_id": false
         }
     ]
 }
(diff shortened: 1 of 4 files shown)


blowdart:
LGTM

PR #33357, opened 2024-08-14
File uploads (writing to memory/Autofac content)

### LINKED ISSUE #33272: The upload stream is closing after the loop.
[EDIT by guardrex to fix the code block ... @rodrigo-lemela-duarte, it's triple-backticks on a line above and below code in a comment]
[EDIT by guardrex to update the report on the var access that results in the error about the closed stream. The error occurs on `content` access.]

### Description

At the `FileUpload2.razor` code:

This part of the code is throwning error:

```csharp
var fileContent = new StreamContent(file.OpenReadStream(maxFileSize));
```

After you get out of the loop and try to access ~`fileContent`~ [should be `content`, see comment below], .NET throws an error saying that the **stream is closed**.

The right piece of the code is:

```csharp
await using var fileStream = uploadedFile.OpenReadStream(maxAllowedSize: maxFileSize);
using var memoryStream = new MemoryStream();
await fileStream.CopyToAsync(memoryStream);
var fileContent = new ByteArrayContent(memoryStream.ToArray());
```

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/file-uploads?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/file-uploads.md

### Document ID

c11d981c-05af-c19d-a333-feedd5978639

### Article author

@guardrex

guardrex:
Fixes #33272

Notes ...

* The versioning assumes that the problems associated with multiple file uploads (and avoiding writing to a `MemoryStream`/byte array) and Autofac will be resolved for >=9.0.
* The <9.0 text won't dissuade writing to a `MemoryStream`/byte array any longer. However, I'm leaving the bits about how it can degrade performance and present a possible security risk.
* 🤔 I think the "security risk" that we were generically mentioning before is DoS, so I'm stating that explicitly now on this PR. Let me know if that's an incorrect guess. 👂 
* The sample app changes were made in the Blazor samples repo for Blazor Server and Blazor Web App samples. Here's an example :point_right: https://github.com/dotnet/blazor-samples/blob/main/8.0/BlazorSample_BlazorWebApp/Components/Pages/FileUpload2.razor#L80-L85

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/file-uploads.md](https://github.com/dotnet/AspNetCore.Docs/blob/3ea630aa017b2d398a39815a9264a2e70a167df0/aspnetcore/blazor/file-uploads.md) | [ASP.NET Core Blazor file uploads](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/file-uploads?branch=pr-en-us-33357) |

### DIFF
--- a/aspnetcore/blazor/file-uploads.md
+++ b/aspnetcore/blazor/file-uploads.md
@@ -49,7 +49,17 @@ To read data from a user-selected file, call <xref:Microsoft.AspNetCore.Componen
 
 <xref:Microsoft.AspNetCore.Components.Forms.IBrowserFile.OpenReadStream%2A> enforces a maximum size in bytes of its <xref:System.IO.Stream>. Reading one file or multiple files larger than 500 KB results in an exception. This limit prevents developers from accidentally reading large files into memory. The `maxAllowedSize` parameter of <xref:Microsoft.AspNetCore.Components.Forms.IBrowserFile.OpenReadStream%2A> can be used to specify a larger size if required.
 
-If you need access to a <xref:System.IO.Stream> that represents the file's bytes, use <xref:Microsoft.AspNetCore.Components.Forms.IBrowserFile.OpenReadStream%2A?displayProperty=nameWithType>. Avoid reading the incoming file stream directly into memory all at once. For example, don't copy all of the file's bytes into a <xref:System.IO.MemoryStream> or read the entire stream into a byte array all at once. These approaches can result in performance and security problems, especially for server-side components. Instead, consider adopting either of the following approaches:
+:::moniker range=">= aspnetcore-9.0"
+
+If you need access to a <xref:System.IO.Stream> that represents the file's bytes, use <xref:Microsoft.AspNetCore.Components.Forms.IBrowserFile.OpenReadStream%2A?displayProperty=nameWithType>. Avoid reading the incoming file stream directly into memory all at once. For example, don't copy all of the file's bytes into a <xref:System.IO.MemoryStream> or read the entire stream into a byte array all at once. These approaches can result in degraded app performance and potential [Denial of Service (DoS)](xref:blazor/security/server/interactive-server-side-rendering#denial-of-service-dos-attacks) risk, especially for server-side components. Instead, consider adopting either of the following approaches:
+
+:::moniker-end
+
+:::moniker range="< aspnetcore-9.0"
+
+To avoid coping all of the file's bytes into a <xref:System.IO.MemoryStream> or reading the entire stream into a byte array all at once, which can result in degraded app performance and potential [Denial of Service (DoS)](xref:blazor/security/server/interactive-server-side-rendering#denial-of-service-dos-attacks) risk, consider adopting either of the following approaches:
+
+:::moniker-end
 
 * Copy the stream directly to a file on disk without reading it into memory. Note that Blazor apps executing code on the server aren't able to access the client's file system directly. 
 * Upload files from the client directly to an external service. For more information, see the [Upload files to an external service](#upload-files-to-an-external-service) section.
@@ -63,6 +73,8 @@ var reader =
     await new StreamReader(browserFile.OpenReadStream()).ReadToEndAsync();
 ```
 
+:::moniker range=">= aspnetcore-9.0"
+
 <span aria-hidden="true">❌</span><span class="visually-hidden">Unsupported:</span> The following approach is **NOT recommended** for [Microsoft Azure Blob Storage](/azure/storage/blobs/storage-blobs-overview) because the file's <xref:System.IO.Stream> content is copied into a <xref:System.IO.MemoryStream> in memory (`memoryStream`) before calling <xref:Azure.Storage.Blobs.BlobContainerClient.UploadBlobAsync%2A>:
 
 ```csharp
@@ -71,6 +83,8 @@ await browserFile.OpenReadStream().CopyToAsync(memoryStream);
 await blobContainerClient.UploadBlobAsync(
     trustedFileName, memoryStream));
 ```
+
+:::moniker-end
 
 <span aria-hidden="true">✔️</span><span class="visually-hidden">Supported:</span> The following approach is **recommended** because the file's <xref:System.IO.Stream> is provided directly to the consumer, a <xref:System.IO.FileStream> that creates the file at the provided path:
 
@@ -87,6 +101,15 @@ await blobContainerClient.UploadBlobAsync(
 ```
 
 A component that receives an image file can call the <xref:Microsoft.AspNetCore.Components.Forms.BrowserFileExtensions.RequestImageFileAsync%2A?displayProperty=nameWithType> convenience method on the file to resize the image data within the browser's JavaScript runtime before the image is streamed into the app. Use cases for calling <xref:Microsoft.AspNetCore.Components.Forms.BrowserFileExtensions.RequestImageFileAsync%2A> are most appropriate for Blazor WebAssembly apps.
+
+
+:::moniker range="< aspnetcore-9.0"
+
+## Autofac Inversion of Control (IoC) container users
+
+If you're using the [Autofac Inversion of Control (IoC) container](https://autofac.org/) instead of the built-in ASP.NET Core dependency injection container, set <xref:Microsoft.AspNetCore.SignalR.HubOptions.DisableImplicitFromServicesParameters%2A> to `true` in the [server-side circuit handler hub options](xref:blazor/fundamentals/signalr#server-side-circuit-handler-options). For more information, see [FileUpload: Did not receive any data in the allotted time (`dotnet/aspnetcore` #38842)](https://github.com/dotnet/aspnetcore/issues/38842#issuecomment-1342540950).
+
+:::moniker-end
 
 ## File size read and upload limits
 
@@ -213,17 +236,7 @@ The component assumes that the Interactive WebAssembly render mode (`Interactive
 
 ## Upload files to a server with server-side rendering
 
-:::moniker range=">= aspnetcore-8.0"
-
-*This section applies to Interactive Server components in Blazor Web Apps.*
-
-:::moniker-end
-
-:::moniker range="< aspnetcore-8.0"
-
-*This section applies to Blazor Server apps.*
-
-:::moniker-end
+*This section applies to Interactive Server components in Blazor Web Apps or Blazor Server apps.*
 
 The following example demonstrates uploading files from a server-side app to a backend web API controller in a separate app, possibly on a separate server.
 
--- a/aspnetcore/blazor/file-uploads.md
+++ b/aspnetcore/blazor/file-uploads.md
@@ -65,6 +65,8 @@ To avoid coping all of the file's bytes into a <xref:System.IO.MemoryStream> or 
 * Copy the stream directly to a file on disk without reading it into memory. Note that Blazor apps executing code on the server aren't able to access the client's file system directly. 
 * Upload files from the client directly to an external service. For more information, see the [Upload files to an external service](#upload-files-to-an-external-service) section.
 
+:::moniker-end
+
 :::moniker range=">= aspnetcore-8.0 < aspnetcore-9.0"
 
 If the component limits file uploads to a single file at a time or if the component only adopts interactive client-side rendering (CSR, `InteractiveWebAssembly`), the component can avoid copying all of the file's bytes into a <xref:System.IO.MemoryStream> or reading the entire stream into a byte array:
--- a/aspnetcore/blazor/file-uploads.md
+++ b/aspnetcore/blazor/file-uploads.md
@@ -56,12 +56,6 @@ If you need access to a <xref:System.IO.Stream> that represents the file's bytes
 
 In the following examples, `browserFile` represents the uploaded file and implements <xref:Microsoft.AspNetCore.Components.Forms.IBrowserFile>. Working implementations for <xref:Microsoft.AspNetCore.Components.Forms.IBrowserFile> are shown in the file upload components later in this article.
 
-<span aria-hidden="true">✔️</span><span class="visually-hidden">Supported:</span> The following approach is **recommended** because the file's <xref:System.IO.Stream> is provided directly to the consumer, a new <xref:System.Net.Http.StreamContent> that contains the file's content:
-
-```csharp
-var fileContent = new StreamContent(browserFile.OpenReadStream(maxFileSize));
-```
-
 <span aria-hidden="true">✔️</span><span class="visually-hidden">Supported:</span> The following approach is **recommended** because the file's <xref:System.IO.Stream> is provided directly to the consumer, a <xref:System.IO.FileStream> that creates the file at the provided path:
 
 ```csharp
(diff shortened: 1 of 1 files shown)


MackinnonBuck (commented):
Thanks @guardrex!

The content you wrote here looks accurate, although I imagine that customers will likely arrive at these docs _after_ experiencing the `TimeoutException`. That is, I'm not sure they'll have the patience to grasp all the intricacies of what scenarios are expected to work determine whether their scenario meets those requirements, and then connect that back to the `TimeoutException` they're experiencing.

Furthermore, these intricacies are due to a bug in the framework (https://github.com/dotnet/aspnetcore/issues/47301), and will probably disappear when that bug gets fixed.

I wonder if we could keep the docs simpler, without explaining the details of the bug, and instead add a "Troubleshooting" section where we:
* List the `TimeoutException` error message (so it can be discovered via web search)
* Explain the possible causes:
  1. Autofac
  2. Not reading the stream to completion
  3. On Blazor server, calling `OpenReadStream` on multiple files before reading them to completion (https://github.com/dotnet/aspnetcore/issues/47301).
* Explain workarounds for each cause

Also: I wrote an example below for another workaround we can suggest other than reading the stream into memory. I would recommend we document that instead, so our guidance can be consistent that reading the stream into memory should be avoided.

The workaround is to define a custom stream type that lazily calls `OpenReadStream` just before the first bytes of the stream are requested. In other words, the stream won't start getting transmitted from the browser to the server until the stream starts to get read in .NET. Here's the example:

```csharp
internal sealed class LazyBrowserFileStream : Stream
{
    private readonly IBrowserFile _file;
    private readonly int _maxAllowedSize;

    private Stream? _underlyingStream;
    private bool _isDisposed;

    public override bool CanRead => true;

    public override bool CanSeek => false;

    public override bool CanWrite => false;

    public override long Length => _file.Size;

    public override long Position
    {
        get => _underlyingStream?.Position ?? 0;
        set => throw new NotSupportedException();
    }

    public LazyBrowserFileStream(IBrowserFile file, int maxAllowedSize)
    {
        _file = file;
        _maxAllowedSize = maxAllowedSize;
    }

    public override void Flush()
    {
        _underlyingStream?.Flush();
    }

    public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
    {
        EnsureStreamIsOpen();
        return _underlyingStream.ReadAsync(buffer, offset, count, cancellationToken);
    }

    public override ValueTask<int> ReadAsync(Memory<byte> buffer, CancellationToken cancellationToken = default)
    {
        EnsureStreamIsOpen();
        return _underlyingStream.ReadAsync(buffer, cancellationToken);
    }

    [MemberNotNull(nameof(_underlyingStream))]
    private void EnsureStreamIsOpen()
    {
        _underlyingStream ??= _file.OpenReadStream(_maxAllowedSize);
    }

    protected override void Dispose(bool disposing)
    {
        if (_isDisposed)
        {
            return;
        }

        _underlyingStream?.Dispose();
        _isDisposed = true;

        base.Dispose(disposing);
    }

    public override int Read(byte[] buffer, int offset, int count)
        => throw new NotSupportedException();

    public override long Seek(long offset, SeekOrigin origin)
        => throw new NotSupportedException();

    public override void SetLength(long value)
        => throw new NotSupportedException(); 

    public override void Write(byte[] buffer, int offset, int count)
        => throw new NotSupportedException();
}
```

And here's an example using it in a `MultipartFormDataContent`:
```csharp
using var content = new MultipartFormDataContent();

foreach (var file in e.GetMultipleFiles())
{
    var stream = new LazyBrowserFileStream(file, maxAllowedSize: 15_000_000);
    var fileContent = new StreamContent(stream);
    fileContent.Headers.ContentType = new MediaTypeHeaderValue(file.ContentType);

    content.Add(fileContent);
}
```

So, if the cause of the `TimeoutException` is item (iii.) listed above, our suggested workaround options would be:
1. After calling `OpenReadStream()` for a file, read the stream fully to completion before calling `OpenReadStream()` on the next file
2. If (1.) isn't possible (e.g., when using `MultipartFormDataContent`), you could create a custom stream wrapper (shown above as `LazyBrowserFileStream`).

Let me know what you think 🙂

MackinnonBuck:
> The problem with troubleshooting guidance is that it's too easily overlooked. Devs will 💥😢, and then we 🙏 to the server gods that they find their way back to the article for help.

I definitely agree that a troubleshooting section can be easily overlooked, but isn't that kind of by design?

I can only speak for myself, but my goal when reading documentation is to spend as little time as possible doing so. Ideally, the APIs I'm using are self-evident and don't require reading docs at all, but if I do have a question, I'll skim the docs only until I've grasped just enough information to unblock me. Rarely do I read a page of docs to completion before starting to use any given feature. And if the docs go into extreme detail, it hurts my ability to skim because there's more to sift through, and I'm even _less_ likely to read them to completion because most of the information there doesn't apply to me.

Given that workflow, I'd rather there be a troubleshooting section so that I can ctrl+F the exception I'm getting (or find it via web search) to see if I can find a workaround. Unfortunately, I don't think I'd have the patience to do a thorough pass of the docs to fully understand the intricacies and limitations of the feature, then use that information to deduce the root cause of my problem.

Obviously this is just my personal preference and I'm not sure how reflective it is of the typical reader, so it should be taken with a grain of salt. I also don't have nearly as much experience writing docs, so I'm totally willing to be overridden on this!

@danroth27, do you have any thoughts?

> Why not just cover the LazyBrowserFileStream approach as the base case for all <9.0 (server-side) implementations?

The challenge is that in most cases, this approach isn't necessary. Only in a small number of scenarios do the right conditions get hit to cause the problem to occur:
1. Using server interactivity, and
2. Supporting multiple file upload, and
3. Calling `OpenReadStream` multiple times in a row without fully consuming the stream first, and
4. Not controlling the code that consumes the stream (e.g., the `MultipartFormDataContent` case)

If our blanket recommendation is to always wrap the returned stream in a custom stream implementation, then it's making the common scenario much more complicated. From that perspective, the workaround does seem more like a "troubleshooting" step to me.

guardrex:
Two of those are built into the current coverage of the dedicated *Upload files to a server with server-side rendering* section ...

* Using server interactivity
* Supporting multiple file upload

... so it seems that even if the opening remarks don't cover the `LazyBrowserFileStream` approach use and we stick with the current guidance (+ a troubleshooting section), that section has to show the `LazyBrowserFileStream` approach.

I'll look at this further in the morning and revise the PR.

guardrex:
@MackinnonBuck ... Try it now with the `LazyBrowserFileStream` approach only in the section that applies to the scenario. I added a *Troubleshoot* section. The sample code was updated on https://github.com/dotnet/blazor-samples/pull/330.

aspnetcore/blazor/file-uploads.md
```diff
@@ -49,13 +49,33 @@ To read data from a user-selected file, call <xref:Microsoft.AspNetCore.Componen
...
 * Copy the stream directly to a file on disk without reading it into memory. Note that Blazor apps executing code on the server aren't able to access the client's file system directly. 
 * Upload files from the client directly to an external service. For more information, see the [Upload files to an external service](#upload-files-to-an-external-service) section.
 
 In the following examples, `browserFile` represents the uploaded file and implements <xref:Microsoft.AspNetCore.Components.Forms.IBrowserFile>. Working implementations for <xref:Microsoft.AspNetCore.Components.Forms.IBrowserFile> are shown in the file upload components later in this article.
 
+<span aria-hidden="true">✔️</span><span class="visually-hidden">Supported:</span> The following approach is **recommended** because the file's <xref:System.IO.Stream> is provided directly to the consumer, a new <xref:System.Net.Http.StreamContent> that contains the file's content:
+
+```csharp
+var fileContent = new StreamContent(browserFile.OpenReadStream(maxFileSize));
+```
```
MackinnonBuck, line 63: Constructing the `StreamContent` actually doesn't start consuming the stream, so _technically_ this doesn't fall in line with our recommended usage. The stream provided to the `StreamContent` only starts being consumed when the request is made. When uploading a single file, this is fine, but when uploading multiple files via `MultipartFormDataContent`, it can create a deadlock.

Maybe we could remove this section? I think the usage examples and note about `LazyBrowserFileStream` you wrote later in the docs are sufficient.

aspnetcore/blazor/file-uploads.md
```diff
@@ -49,13 +49,33 @@ To read data from a user-selected file, call <xref:Microsoft.AspNetCore.Componen
...
+
+<span aria-hidden="true">✔️</span><span class="visually-hidden">Supported:</span> The following approach is **recommended** for [Microsoft Azure Blob Storage](/azure/storage/blobs/storage-blobs-overview) because the file's <xref:System.IO.Stream> is provided directly to <xref:Azure.Storage.Blobs.BlobContainerClient.UploadBlobAsync%2A>:
+
+```csharp
+await blobContainerClient.UploadBlobAsync(
+    trustedFileName, browserFile.OpenReadStream());
+```
+
 <span aria-hidden="true">❌</span><span class="visually-hidden">Unsupported:</span> The following approach is **NOT recommended** because the file's <xref:System.IO.Stream> content is read into a <xref:System.String> in memory (`reader`):
```
MackinnonBuck, line 79: I know this is not something you change here, but would we consider the terminology "Not recommended" instead of "Unsupported"? Technically the framework supports this usage, but we wouldn't recommend it for most cases.

aspnetcore/blazor/file-uploads.md
```diff
@@ -72,21 +92,16 @@ await blobContainerClient.UploadBlobAsync(
...
-    trustedFileName, browserFile.OpenReadStream());
-```
+## Autofac Inversion of Control (IoC) container users
 
-A component that receives an image file can call the <xref:Microsoft.AspNetCore.Components.Forms.BrowserFileExtensions.RequestImageFileAsync%2A?displayProperty=nameWithType> convenience method on the file to resize the image data within the browser's JavaScript runtime before the image is streamed into the app. Use cases for calling <xref:Microsoft.AspNetCore.Components.Forms.BrowserFileExtensions.RequestImageFileAsync%2A> are most appropriate for Blazor WebAssembly apps.
+If you're using the [Autofac Inversion of Control (IoC) container](https://autofac.org/) instead of the built-in ASP.NET Core dependency injection container, set <xref:Microsoft.AspNetCore.SignalR.HubOptions.DisableImplicitFromServicesParameters%2A> to `true` in the [server-side circuit handler hub options](xref:blazor/fundamentals/signalr#server-side-circuit-handler-options). For more information, see [FileUpload: Did not receive any data in the allotted time (`dotnet/aspnetcore` #38842)](https://github.com/dotnet/aspnetcore/issues/38842#issuecomment-1342540950).
+
```
MackinnonBuck, line 103: Do you think we need this section, given there's a troubleshooting section at the bottom?

aspnetcore/blazor/file-uploads.md
```diff
@@ -867,6 +934,23 @@ Blazor relies on <xref:Microsoft.AspNetCore.SignalR.HubOptions.MaximumParallelIn
...
+Possible causes:
+
+* Using the [Autofac Inversion of Control (IoC) container](https://autofac.org/) instead of the built-in ASP.NET Core dependency injection container in versions of ASP.NET Core earlier than 9.0. To resolve the issue, set <xref:Microsoft.AspNetCore.SignalR.HubOptions.DisableImplicitFromServicesParameters%2A> to `true` in the [server-side circuit handler hub options](xref:blazor/fundamentals/signalr#server-side-circuit-handler-options). For more information, see [FileUpload: Did not receive any data in the allotted time (`dotnet/aspnetcore` #38842)](https://github.com/dotnet/aspnetcore/issues/38842#issuecomment-1342540950).
+
+* Not reading the stream to completion. This isn't a framework issue. Trap the exception and investigate it further in your local environment/network.
+
+* Using server-side rendering and calling <xref:Microsoft.AspNetCore.Components.Forms.IBrowserFile.OpenReadStream%2A> on multiple files before reading them to completion in versions of ASP.NET Core earlier than 9.0. To resolve the issue, adopt ***either*** of the following approaches:
+
+  * Upgrade the app to ASP.NET Core 9.0 or later.
```
MackinnonBuck, line 951: Most likely, this fix won't make it into .NET 9, so we might want to remove this bit.

aspnetcore/blazor/file-uploads.md
```diff
@@ -298,6 +335,34 @@ The following `FileUpload2` component:
...
+
+```diff
+- var stream = new LazyBrowserFileStream(file, maxFileSize);
+- var fileContent = new StreamContent(stream);
++ var fileContent = new StreamContent(file.OpenReadStream(maxFileSize));
+```
+
+Remove the `LazyBrowserFileStream` class (`LazyBrowserFileStream.cs`), as it isn't used.
+
+:::moniker-end
```
MackinnonBuck, line 364: Would we consider making the default example not use `LazyBrowserFileStream`, but then add a note explaining that if you're using server interactivity and support multiple file upload, the `LazyBrowserFileStream` approach might be necessary (and we link to the troubleshooting section which has the sample)? If you disagree, feel free to keep it how it is 🙂

aspnetcore/blazor/file-uploads.md
```diff
@@ -867,6 +934,23 @@ Blazor relies on <xref:Microsoft.AspNetCore.SignalR.HubOptions.MaximumParallelIn
 
 Increasing the value leads to a high probability that `CopyTo` operations throw `System.InvalidOperationException: 'Reading is not allowed after reader was completed.'`. For more information, see [MaximumParallelInvocationsPerClient > 1 breaks file upload in Blazor Server mode (`dotnet/aspnetcore` #53951)](https://github.com/dotnet/aspnetcore/issues/53951).
 
+## Troubleshoot
```
MackinnonBuck, line 937: (not sure which is the more standard terminology for our docs)
```suggestion
## Troubleshooting
```

aspnetcore/blazor/file-uploads.md
```diff
@@ -173,7 +188,7 @@ Because the example uses the app's [environment](xref:blazor/fundamentals/enviro
 
 The following example processes file bytes and doesn't send files to a destination outside of the app. For an example of a Razor component that sends a file to a server or service, see the following sections:
 
-* [Upload files to a server](#upload-files-to-a-server)
+* [Upload files to a server with client-side rendering (CSR)](#upload-files-to-a-server-with-client-side-rendering-csr)
```
MackinnonBuck, line 191: I always forget what our definitions for these terms are 😄 Is CSR limited to Blazor WebAssembly? Is that the term we prefer over "WebAssembly interactivity"? Including the term "rendering" here threw me off a bit because it's more about where the code is running than how the rendering happens. If this wording is consistent with what we already do elsewhere in the docs, then it's fine to leave it as-is.

MackinnonBuck (commented):
I like the changes! Just did a more thorough pass and have a few notes/suggestions.

aspnetcore/blazor/file-uploads.md
guardrex: It's safer to have it listed in both spots. Aren't Autofac users required to adopt the approach? Why don't we keep this for a little while? I'll mark it now for removal at .NET 10, so the section will go away permanently in a year and then only appear in the *Troubleshoot* section after that.

aspnetcore/blazor/file-uploads.md
guardrex: Oh, I see. Ok ... I'll fix this.

aspnetcore/blazor/file-uploads.md
guardrex: > if you're using server interactivity and support multiple file upload

That's exactly the scenario of this section. I think we should get a ruling on it from @danroth27. Broken devs can get angry 😡 with a 🦖 when we give them code that 💥. Personally, I'd like to try and prevent it from happening ... and head off the doc issues 💸.

aspnetcore/blazor/file-uploads.md
guardrex: No 🎲🎲 ... no gerunds in section headings per the docs style manual ... they don't translate well into other languages.

aspnetcore/blazor/file-uploads.md
MackinnonBuck: Sounds good, we can keep it as is!

aspnetcore/blazor/file-uploads.md
MackinnonBuck: Totally fine if that's the established language!

MackinnonBuck (approved):
LGTM!

PR #33221, opened 2024-08-01
Update Linux dev cert trust instructions

amcasey:
The dev experience should mostly match other platforms now, but I've added a section with caveats.

Firefox no longer seems to need special handling on Windows or macOS.  I strongly suspect that section could also be removed in the pre-6.0 doc, but I haven't had a chance to confirm.

Context: https://github.com/dotnet/aspnetcore/pull/56582

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/security/docker-compose-https.md](https://github.com/dotnet/AspNetCore.Docs/blob/e8c64bc0717f68cdaec278484e8e11e180bbc1ea/aspnetcore/security/docker-compose-https.md) | [Hosting ASP.NET Core images with Docker Compose over HTTPS](https://review.learn.microsoft.com/en-us/aspnet/core/security/docker-compose-https?branch=pr-en-us-33221) |
| [aspnetcore/security/docker-https.md](https://github.com/dotnet/AspNetCore.Docs/blob/e8c64bc0717f68cdaec278484e8e11e180bbc1ea/aspnetcore/security/docker-https.md) | [Hosting ASP.NET Core Images with Docker over HTTPS](https://review.learn.microsoft.com/en-us/aspnet/core/security/docker-https?branch=pr-en-us-33221) |
| [aspnetcore/security/enforcing-ssl.md](https://github.com/dotnet/AspNetCore.Docs/blob/e8c64bc0717f68cdaec278484e8e11e180bbc1ea/aspnetcore/security/enforcing-ssl.md) | [Enforce HTTPS in ASP.NET Core](https://review.learn.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?branch=pr-en-us-33221) |

PR #33472, opened 2024-08-29
Remove special Firefox instructions in >=.NET 6 sections

### LINKED ISSUE #33230: Remove FF special Trust the HTTPS cert instructions
### Description

Maybe remove it from 6.0+ and in earlier version add a note it may not be necessary?

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?view=aspnetcore-8.0&branch=pr-en-us-33221&tabs=visual-studio%2Clinux-sles

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/security/enforcing-ssl.md

### Document ID

12ca0b9a-2638-a4ce-c45f-d4579c83dafb

### Article author

@tdykstra

---
[Associated WorkItem - 292347](https://dev.azure.com/msft-skilling/Content/_workitems/edit/292347)

tdykstra:
Fixes #33230 

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/security/enforcing-ssl.md](https://github.com/dotnet/AspNetCore.Docs/blob/f45ab8b7b071b7b7b32019b676103e23cdffe852/aspnetcore/security/enforcing-ssl.md) | [Enforce HTTPS in ASP.NET Core](https://review.learn.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?branch=pr-en-us-33472) |

### DIFF
--- a/aspnetcore/security/enforcing-ssl.md
+++ b/aspnetcore/security/enforcing-ssl.md
@@ -815,7 +815,7 @@ dotnet dev-certs https --help
 
 ### Trust the HTTPS certificate with Firefox to prevent SEC_ERROR_INADEQUATE_KEY_USAGE error
 
-**Note:** The following special instructions for Firefox might not be necessary.
+In some instances the Firefox browser uses its own certificate store, and therefore doesn't trust the [IIS Express](/iis/extensions/introduction-to-iis-express/iis-express-overview) or [Kestrel](xref:fundamentals/servers/kestrel) developer certificates and requires the following instructions. 
 
 The Firefox browser uses its own certificate store, and therefore doesn't trust the [IIS Express](/iis/extensions/introduction-to-iis-express/iis-express-overview) or [Kestrel](xref:fundamentals/servers/kestrel) developer certificates.
 
--- a/aspnetcore/security/enforcing-ssl.md
+++ b/aspnetcore/security/enforcing-ssl.md
@@ -201,8 +201,6 @@ dotnet new webapp --no-https
 
 ## Trust the ASP.NET Core HTTPS development certificate on Windows and macOS
 
-For the Firefox browser, see the next section.
-
 The .NET Core SDK includes an HTTPS development certificate. The certificate is installed as part of the first-run experience. For example, `dotnet --info` produces a variation of the following output:
 
 ```cli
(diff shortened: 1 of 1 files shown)


aspnetcore/security/enforcing-ssl.md
```diff
@@ -905,6 +815,8 @@ dotnet dev-certs https --help
 
 ### Trust the HTTPS certificate with Firefox to prevent SEC_ERROR_INADEQUATE_KEY_USAGE error
 
+**Note:** The following special instructions for Firefox might not be necessary.
```
Rick-Anderson, line 818:
```suggestion
In some instances the Firefox browser uses its own certificate store, and therefore doesn't trust the [IIS Express](/iis/extensions/introduction-to-iis-express/iis-express-overview) or [Kestrel](xref:fundamentals/servers/kestrel) developer certificates and requires the following instructions. 
```

aspnetcore/security/enforcing-ssl.md
```diff
@@ -1003,6 +915,8 @@ For chromium browsers on Linux:
 
 ### Trust the certificate with Firefox on Linux
 
+**Note:** The following special instructions for Firefox might not be necessary.
```
Rick-Anderson, line 918:
```suggestion
In some instances the Firefox browser uses its own certificate store, and therefore doesn't trust the [IIS Express](/iis/extensions/introduction-to-iis-express/iis-express-overview) or [Kestrel](xref:fundamentals/servers/kestrel) developer certificates and requires the following instructions. 
```

Rick-Anderson (changes requested):

amcasey:
I think these changes were probably prompted by a comment from me, but I can't recall the context.  Would you happen to have a link or something?

aspnetcore/security/enforcing-ssl.md
```diff
@@ -231,55 +229,13 @@ dotnet dev-certs https --help
 
 <a name="trust-ff"></a>
 
-### Trust the HTTPS certificate with Firefox to prevent SEC_ERROR_INADEQUATE_KEY_USAGE error
```
amcasey, line 234: To confirm: these are the windows and macos instructions?  If so, I agree that removing them makes sense.

aspnetcore/security/enforcing-ssl.md
```diff
@@ -345,36 +301,6 @@ For chromium browsers on Linux:
 
 <a name="trust-ff-linux"></a>
 
-#### Trust the certificate with Firefox on Linux
```
amcasey, line 348: Why delete, rather than update, this section?  It seems like we could just tell people to run `certutil` as for Chromium.

amcasey: There's a note in an unedited part of this document (above) that says `certutil` is "legacy".  I'm not sure why - maybe just the specific version linked?

amcasey: Above, where it says to run `certutil` twice, it should probably only be run once - with `"P,,"` for Chromium or `"C,,"` for Firefox.

amcasey (changes requested):

Rick-Anderson:
> I think these changes were probably prompted by a comment from me, but I can't recall the context. Would you happen to have a link or something?

#33221

aspnetcore/security/enforcing-ssl.md
Rick-Anderson: > Why delete, rather than update, this section? It seems like we could just tell people to run `certutil` as for Chromium.

Are the instructions different using  `certutil`  on FF different than for Chromium? If not, we'll make sure the Chromium instructions say Chromium and FF.

Rick-Anderson: > There's a note in an unedited part of this document (above) that says `certutil` is "legacy". I'm not sure why - maybe just the specific version linked?

the *legacy* comment is only for the the [Red Hat Enterprise Linux tab](https://learn.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?view=aspnetcore-8.0&branch=pr-en-us-33472&tabs=visual-studio%2Clinux-rhel#trust-https-certificate-on-linux-using-edge-or-chrome)

aspnetcore/security/enforcing-ssl.md
```diff
@@ -1003,6 +915,8 @@ For chromium browsers on Linux:
 
 ### Trust the certificate with Firefox on Linux
 
+In some instances the Firefox browser uses its own certificate store, and therefore doesn't trust the [IIS Express](/iis/extensions/introduction-to-iis-express/iis-express-overview) or [Kestrel](xref:fundamentals/servers/kestrel) developer certificates and requires the following instructions. 
+
```
Rick-Anderson, line 919: @amcasey and @tdykstra  I think  in the FF section we should add something like the following:

In some instances theses instructions don't work with Firefox on Linux. See [Trust the HTTPS certificate with Firefox](https://learn.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?view=aspnetcore-5.0&tabs=visual-studio%2Clinux-ubuntu#trust-the-https-certificate-with-firefox-to-prevent-sec_error_inadequate_key_usage-error-1] for alternative instructions. The link goes back to .NET 5 and has all the previous instructions.

Rick-Anderson (changes requested):

aspnetcore/security/enforcing-ssl.md
amcasey: > Are the instructions different using certutil on FF different than for Chromium? If not, we'll make sure the Chromium instructions say Chromium and FF.

The two differences are the usage, as noted in my comment, and the location of the database.  Otherwise, it should be the same.

aspnetcore/security/enforcing-ssl.md
```diff
@@ -905,6 +815,8 @@ dotnet dev-certs https --help
 
 ### Trust the HTTPS certificate with Firefox to prevent SEC_ERROR_INADEQUATE_KEY_USAGE error
 
+In some instances the Firefox browser uses its own certificate store, and therefore doesn't trust the [IIS Express](/iis/extensions/introduction-to-iis-express/iis-express-overview) or [Kestrel](xref:fundamentals/servers/kestrel) developer certificates and requires the following instructions.
```
amcasey, line 818: This seems to overlap substantially with the following paragraph.

aspnetcore/security/enforcing-ssl.md
amcasey: AFAIK, the old instructions apply to old version of Firefox, rather than to old versions of .NET.  If you're on a new enough .NET that `--trust` works, no Firefox-specific steps should be required.  Otherwise, the instructions should be as for Chromium, but with `"C,,"` and different locations.

Or maybe I've misunderstood your suggestion?  I'm not entirely sure where in the document this text occurs.

Rick-Anderson: > AFAIK, the old instructions apply to old version of Firefox, rather than to old versions of .NET. If you're on a new enough .NET that `--trust` works, no Firefox-specific steps should be required. Otherwise, the instructions should be as for Chromium, but with `"C,,"` and different locations.
> 
> Or maybe I've misunderstood your suggestion? I'm not entirely sure where in the document this text occurs.

Great, we'll replace that with older versions of FF

tdykstra:
Addressed by #33221

Issue #33273, opened 2024-08-06
CS Effort: AspNetCore.Docs: Images:  Triage and remove flagged orphaned images


wadepickett:
CS Effort: AspNetCore.Docs: Images: 

Image scans triage:
59 images

Tasks:

- [x]  **Run Genevieve's Clean Repo Tool** to identify and get rid of all images that are orphaned.  That alone may help clear a number of these out.  ( Found and removed 15, see PR: #33501 )

- [x] Review and triage all images first.

- [x] For images that need to be fixed, set a plan for fixing them and schedule.

For details on how to triage, fix and note completion, see:
[SFI info in Teams](https://teams.microsoft.com/l/message/19:49ff5836203f4b358334bfcfa593f252@thread.tacv2/1722361260037?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=8d8475c9-f864-4ed0-825e-b2e419026017&parentMessageId=1722361260037&teamName=Content%20Team%20%7C%20C%2BE%20Skilling&channelName=General&createdTime=1722361260037)

Related to parent tracking issue for this repo: #33229

(As soon as we have a label for this to auto connect the appropriate ADO parent set it.)

---
[Associated WorkItem - 295459](https://dev.azure.com/msft-skilling/Content/_workitems/edit/295459)

wadepickett:
Running customized version of  CleanRepo I found the following image files that were flagged as an SFI issue as of 8/29/24 to be orphaned files I can delete.  PR to follow, and then the rest will be triaged into similar groups to fix and the spreadsheets updated.

15 images are orphaned listed below which can just be removed.  3 images need to be set to "Completed-No Change" since they are showing the Trust cert dialog for a local development environment, see further below.  So  out of the total  59 images that had issues. Leaving 41 images to actually fix.  It looks like most of those remaining to be fixed require some text to be blanked out in the image.

**Orphanced images to be removed (15):**
aspnetcore/tutorials/razor-pages/razor-pages-start/_static/6/trustcertmac6.png
aspnetcore/tutorials/min-web-api/_static/post.png
aspnetcore/tutorials/first-web-api/_static/pmcput.png
aspnetcore/tutorials/razor-pages/razor-pages-start/_static/6/certmac6.png
aspnetcore/test/http-files/_static/dotenv-completions.png
aspnetcore/tutorials/publish-to-azure-webapp-using-vs/_static/set.png
[...]

PR #33501, opened 2024-08-30
SFI:Delete orphaned images with SFI flag

wadepickett:
Contributes to #33273

Deleting 15 images that were determined to be orphaned that also were on the SFI flagged list for images to be addressed.


wadepickett:
Verified no missing image link warnings in bild.

tdykstra (approved):

PR #33536, opened 2024-09-05
[Blazor] Constructor injection updates

### LINKED ISSUE #33530: [Blazor] Constructor injection example - drop property usage
### Description

```csharp
public partial class ConstructorInjection(NavigationManager navigation)
{
    protected NavigationManager Navigation { get; } = navigation;
}
```

Do we still want to use a property when demonstrating constructor injection?

I think we've gotten used to seeing properties with injected services in all the Blazor samples so far. However, to be honest, that's just a result of the fact that Blazor didn’t support constructor injection until recently. When using standard constructor injection, we typically don't create properties for services. Instead, we use them as `private readonly` fields or, somewhat controversially, capture them directly from the primary constructor.

I think an example like this would be more appropriate:
```csharp
public partial class MyComponent(NavigationManager navigation)
{
    private void HandleClick()
    {
        navigation.NavigateTo(newUri);
    }
}
```

Can I update the example to reflect this approach?

cc @guardrex 

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/dependency-injection?view=aspnetcore-9.0#request-a-service-in-a-component

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/fundamentals/dependency-injection.md

### Document ID

d9aabe95-69e9-0a03-81ff-429914137f9d

### Article author

@guardrex

hakenr:
Fixes #33530.

cc @guardrex 

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/dependency-injection.md](https://github.com/dotnet/AspNetCore.Docs/blob/c2b2a29db051dcfd9c60a94581306b27b96e466d/aspnetcore/blazor/fundamentals/dependency-injection.md) | [ASP.NET Core Blazor dependency injection](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/dependency-injection?branch=pr-en-us-33536) |

### DIFF
--- a/aspnetcore/blazor/fundamentals/dependency-injection.md
+++ b/aspnetcore/blazor/fundamentals/dependency-injection.md
@@ -224,7 +224,7 @@ After services are added to the service collection, inject one or more services 
 ```razor
 @page "/constructor-injection"
 
-<button @onclick="@(() => Navigation.NavigateTo("/counter"))">
+<button @onclick="HandleClick">
     Take me to the Counter component
 </button>
 ```
@@ -236,7 +236,10 @@ using Microsoft.AspNetCore.Components;
 
 public partial class ConstructorInjection(NavigationManager navigation)
 {
-    protected NavigationManager Navigation { get; } = navigation;
+    private void HandleClick()
+    {
+        navigation.NavigateTo("/counter");
+    }
 }
 ```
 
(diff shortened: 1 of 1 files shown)


guardrex (changes requested):
The code updates must be versioned 8.0 or later ... use the versioning blocks that I provided :point_up:.

If it's unclear what to do, I'll edit the PR and add them. Let me know. 👂

hakenr:
> The code updates must be versioned 8.0 or later ... use the versioning blocks that I provided ☝️.

@guardrex The whole _constructor injection_ section is `:::moniker range=">= aspnetcore-9.0"`, there was no constructor injection in 8.0.

guardrex:
Yes ... sorry ... I was thinking of ***for C#***. 😄 I'm with you now!

There's something that slightly concerns me about this update: Can the field be used in component markup? ... because injecting and assigning to the property is opening up the possibility to use the injected service in the markup.

guardrex (approved):

guardrex:
That's a minor point. I agree that your composition is the *BEST* for this.

guardrex:
Let's go ahead with this. That was a very minor point. Even if the field isn't available, devs can use a property to get around it on their own. It doesn't matter for the concept being addressed here.

Thanks for doing this! I'm swamped as usual 😩😆.

hakenr:
>  Can the field be used in component markup? ... because injecting and assigning to the property is opening up the possibility to use the injected service in the markup.

@guardrex 
The code-behind is a partial class definition, and all the fields are usable in markup.
In my case, `navigation` is technically not a field, but a _captured primary constructor parameter_.
To be certain, I just tested it, and the captured parameter also works in markup as expected.

PR #33538, opened 2024-09-05
What's new .NET 9 - Blazor constructor injection sample

hakenr:
Relates to #33530.
...found one more instance of the sample (portion).

cc @guardrex


guardrex (approved):
When we were looking at the other PR, my search didn't include this folder 😢.

Issue #33546, opened 2024-09-05
WN: Add SignalR distributed tracing section and code sample


wadepickett:
1. Add the new section for SignalR distributed tracing based on the dotnet/release-notes-draft.
2. Edit sweep
3. If any code samples, then move inline code into a runnable project and link the code back in.

---
[Associated WorkItem - 317469](https://dev.azure.com/msft-skilling/Content/_workitems/edit/317469)

wadepickett:
Moving this out to a future sprint after we have completed all the SFI work

wadepickett:
Related to #34183 which was the follow up PR after this prep PR.

PR #33547, opened 2024-09-05
WN: SignalR distributed tracing: Pre-work moving as is to include

wadepickett:
Contributes to #33546

This is just the pre-work of moving the content as-is from the Product Unit into an include that is easier to diff from once I edit it.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/release-notes/aspnetcore-9.0.md](https://github.com/dotnet/AspNetCore.Docs/blob/a34c7e34d2dc92d468b22c33c05495c5d75452af/aspnetcore/release-notes/aspnetcore-9.0.md) | [What's new in ASP.NET Core 9.0](https://review.learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-9.0?view=aspnetcore-9.0&branch=pr-en-us-33547#improved-activities-for-signalr) |


wadepickett:
@BrennanConroy, this is just a prep for moving your original content to an include for the What's New before I do an edit and move inline code to a project and link it back in.

However, it looked like I should combine the new SignalR client ActivitySource and previous server ActivitySource into one section.  I have not edited yet, basically just added those two sections under one parent section.  Does that generally look OK?  If so I will merge this, and then you will see a diff in the next PR for any edit changes I make if needed.  Thanks.

wadepickett:
I will go ahead and combine the two ActivitySource sections into one grouping and merge, then create a new PR to edit and move the code out to a project and link back in.

This PR was just prep only so we have something to diff against for edit.

PR #33557, opened 2024-09-06
Override InputNumber type attribute

### LINKED ISSUE #33556: Override InputNumber type attribute
### Description

Per https://github.com/dotnet/AspNetCore.Docs/issues/33442#issuecomment-2330249873.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/forms/input-components?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/forms/input-components.md

### Document ID

52ee411b-69fe-c0bc-5305-846d9384ebbd

### Article author

@guardrex

guardrex:
Fixes #33556

Mackinnon ...

* This covers both the *9.0 What's New* article and the new section in the *Input Components* article.
* Sorry that I replaced your example. I'm trying to go with Star Trek-themed forms throughout the *Forms* node.
* Is there anything else that we need to tell devs in the article section?

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/forms/input-components.md](https://github.com/dotnet/AspNetCore.Docs/blob/378ef5264bf06c9a358fce7da5917206c62d0375/aspnetcore/blazor/forms/input-components.md) | [ASP.NET Core Blazor input components](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/forms/input-components?branch=pr-en-us-33557) |

### DIFF
--- a/aspnetcore/blazor/forms/input-components.md
+++ b/aspnetcore/blazor/forms/input-components.md
@@ -527,3 +527,87 @@ Assign a custom template to <xref:Microsoft.AspNetCore.Components.Forms.InputDat
 > The ProductionDate field has an incorrect date value.
 
 :::moniker-end
+
+:::moniker range=">= aspnetcore-9.0"
+
+### Override `InputNumber` `type` attribute
+
+The `InputNumber` component supports overriding the `type` attribute, where the `{TYPE}` placeholder represents the `type` value:
+
+```razor
+<InputNumber ... type="{TYPE}" ... />
+```
+
+For example, specify an [`<input>` element of `type="range"`](https://developer.mozilla.org/docs/Web/HTML/Element/input/range) with an `InputNumber` component to create a range input that supports model binding and form validation, typically rendered as a slider or dial control rather than a text box:
+
+```razor
+<InputNumber @bind-Value="..." max="..." min="..." step="..." type="range" />
+```
+
+If the user's browser doesn't support `type="range"`, the `<input>` element gracefully degrades back to a default text box.
+
+In the following example, the `InputNumber` components set their `type` attributes to `range`, resulting in a set of three sliders or dial controls by typical browsers.
+
+`EngineForm.razor`:
+
+```razor
+@page "/engine-form"
+@inject ILogger<EngineForm> Logger
+
+<EditForm Model="Model" OnSubmit="Submit" FormName="EngineForm">
+    <div>
+        <label>
+            Nacelle Count (2-6): 
+            <InputNumber @bind-Value="Model!.NacelleCount" max="6" min="2" 
+                step="1" type="range" />
+        </label>
+    </div>
+    <div>
+        <label>
+            Plasma Injector Count (per nacelle, 2-8): 
+            <InputNumber @bind-Value="Model!.PlasmaInjectorCount" max="8" min="2" 
+                step="1" type="range" />
+        </label>
+    </div>
+    <div>
+        <label>
+            Warp Field Coil Count (per nacelle, 4-20): 
+            <InputNumber @bind-Value="Model!.WarpFieldCoilCount" max="20" min="4" 
+                step="1" type="range" />
+        </label>
+    </div>
+    <div>
+        <button type="submit">Submit</button>
+    </div>
+</EditForm>
+
+@code {
+    [SupplyParameterFromForm]
+    private EngineSpecifications? Model { get; set; }
+
+    protected override void OnInitialized() => Model ??= new();
+
+    private void Submit()
+    {
+        Logger.LogInformation(
+            "Nacelle Count = {NacelleCount}, Plasma Injector Count = " +
+            "{PlasmaInjectorCount}, Warp Field Coil Count = {WarpFieldCoilCount}", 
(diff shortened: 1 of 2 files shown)


MackinnonBuck (commented):
I appreciate the new example - it's more fun than the one I drafted up 🙂 

However, I'm wondering if we should keep the docs for this feature a bit shorter. Overriding the `type` attribute on `InputNumber` is a pretty niche scenario that can likely be mentioned in passing rather than getting a section dedicated to it. Do you agree? I appreciate that it's a cool example, I'm just worried that it could make the docs a bit noisy when we're mostly trying to highlight the line(s) with `<InputNumber type="range" />`.

MackinnonBuck:
i.e., we could even just have a note somewhere like:

> [!NOTE]
> The `<InputNumber>` component can also act as a range input by adding the `type="range" attribute`.

guardrex:
I suppose so. We don't have a dedicated section for `InputNumber`, so I didn't have much choice if we wanted a full sample.

I'll see about knocking it down to something simple in the main coverage, and then I'll place a reduced full example component in the *What's New* coverage ... Star Trek, of course! 😆

guardrex:
I'll get back to this within a few hours. I'm OOF at the moment.

guardrex:
Nevermind ... I had time now to adjust it before leaving. Try it now.

guardrex:
Ok ... I'm done mess'in around with it. 😆

aspnetcore/blazor/forms/input-components.md
```diff
@@ -67,6 +67,18 @@ Input components provide default behavior for validating when a field is changed
...
+:::moniker range=">= aspnetcore-9.0"
+
+The <xref:Microsoft.AspNetCore.Components.Forms.InputNumber%601> component supports overriding the `type` attribute. For example, specify an [`<input>` element of `type="range"`](https://developer.mozilla.org/docs/Web/HTML/Element/input/range) to create a range input that supports model binding and form validation, typically rendered as a slider or dial control rather than a text box:
+
+```razor
+<InputNumber @bind-Value="..." max="..." min="..." step="..." type="range" />
+```
+
+If the user's browser doesn't support `type="range"`, the `<input>` element gracefully degrades back to a default text box.
+
```
MackinnonBuck, line 79: We can probably remove this since range inputs have been supported for a long time now (as long ago as 2008 in Safari).

aspnetcore/blazor/forms/input-components.md
```diff
@@ -67,6 +67,18 @@ Input components provide default behavior for validating when a field is changed
 
 Some components include useful parsing logic. For example, <xref:Microsoft.AspNetCore.Components.Forms.InputDate%601> and <xref:Microsoft.AspNetCore.Components.Forms.InputNumber%601> handle unparseable values gracefully by registering unparseable values as validation errors. Types that can accept null values also support nullability of the target field (for example, `int?` for a nullable integer).
 
+:::moniker range=">= aspnetcore-9.0"
+
+The <xref:Microsoft.AspNetCore.Components.Forms.InputNumber%601> component supports overriding the `type` attribute. For example, specify an [`<input>` element of `type="range"`](https://developer.mozilla.org/docs/Web/HTML/Element/input/range) to create a range input that supports model binding and form validation, typically rendered as a slider or dial control rather than a text box:
+
```
MackinnonBuck, line 73: I wonder if, rather than noting that `<InputNumber>` has an overridable `"type"`, we just mention that it supports specifying `type="range"` specifically. I can't think of another input type that customers would want to specify other than `"range"`, and I don't necessarily want to claim that we support anything other than `"range"` (because that's not tested and might have some weird behavior).

> For example, specify an [`<input>` element of `type="range"`](https://developer.mozilla.org/docs/Web/HTML/Element/input/range) to create a range input that supports model binding and form validation

Where does the customer specify an `<input>` element? They're using `<InputNumber>` right? Should this instead say:

> Specify the [`type="range"` attribute](https://developer.mozilla.org/docs/Web/HTML/Element/input/range) to create a range input that supports model binding and form validation

?

aspnetcore/blazor/forms/input-components.md
MackinnonBuck: (and Blazor definitely won't work on a 2008 browser anyway 🙂)

aspnetcore/blazor/forms/input-components.md
guardrex: I'm sorry about this. The wording is blown because I'm trying to get off of this PC 😆 ... a bit of 🦖🧠🔥hit me today.

Hopefully, I'll get it right on the next commit. 🤞🍀

guardrex:
@MackinnonBuck ... See if that (***finally*** 🙈) gets it.

I have 🍺 in the fridge. Maybe I need to drink one ............ ***or FIVE!*** 😆

MackinnonBuck (approved):

MackinnonBuck:
Looks great!

PR #33574, opened 2024-09-09
Remark on UseBlazorFrameworkFiles with MapStaticAssets

### LINKED ISSUE #33571: Document bad interaction between MapStaticAssets and UseBlazorFrameworkFiles
### Description

I spent a while trying to find out why the Blazor framework static assets files in my API project (built from a referenced Blazor WASM project) weren't being served when using `MapStaticAssets`, instead returning a 500 status code, as the requests reached the end of the request handling pipeline without being handled.

I found that the issue was due to having `UseBlazorFrameworkFiles` in my middleware pipeline, I'm not sure of the details as to why it caused a 500 rather than just serving the framework files.

I feel this would make sense as a warning in the Blazor/Fundementals/Static Files docs, in the section discussing using `MapStaticAssets`, I'm happy to contribute this, I'd just like to get a consensus that this is the correct move first. I _personally_ don't think this is a bug that needs fixing.

I've drafted something [here](https://github.com/hwoodiwiss/AspNetCore.Docs/blob/hwoodiwiss-33570/aspnetcore/blazor/fundamentals/static-files.md?plain=1#L55-L56).

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/static-files?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/fundamentals/static-files.md

### Document ID

25458428-030f-6acb-457e-1b4dfa722043

### Article author

@guardrex

guardrex:
Fixes #33571

Thanks @hwoodiwiss! 🚀 ... See if you think this captures the right sentiment. The added paragraph is versioned with the rest of the `MapStaticAssets` coverage, so it will appear for >=9.0.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/static-files.md](https://github.com/dotnet/AspNetCore.Docs/blob/6d58e5bc7f4b916fc0e3a2869206b228d1b70c71/aspnetcore/blazor/fundamentals/static-files.md) | [ASP.NET Core Blazor static files](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/static-files?branch=pr-en-us-33574) |

### DIFF
--- a/aspnetcore/blazor/fundamentals/static-files.md
+++ b/aspnetcore/blazor/fundamentals/static-files.md
@@ -40,6 +40,8 @@ Configure Map Static Assets Middleware by calling `MapStaticAssets` in the app's
 
 `MapStaticAssets` can replace <xref:Microsoft.AspNetCore.Builder.StaticFileExtensions.UseStaticFiles%2A> in most situations. However, `MapStaticAssets` is optimized for serving the assets from known locations in the app at build and publish time. If the app serves assets from other locations, such as disk or embedded resources, <xref:Microsoft.AspNetCore.Builder.StaticFileExtensions.UseStaticFiles%2A> should be used.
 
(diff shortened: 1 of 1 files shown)


hwoodiwiss:
That looks good to me 👍🏼 

Thank you :D

PR #33590, opened 2024-09-12
Fix hyperlinks caused due to wrong markdown syntax

SpicyCatGames:
Search for "xref:Microsoft.AspNetCore.Authorization.AuthorizeAttribute" on https://learn.microsoft.com/en-us/aspnet/core/security/authentication/windowsauth?view=aspnetcore-8.0&tabs=visual-studio to see the problem.


guardrex (approved):
Thanks @SpicyCatGames! 🐈

Issue #32199, opened 2024-03-28
Clarify and focus the recommendation to use the Azure SignalR Service with Blazor


danroth27:
### Description

The "Host and deploy server-side Blazor apps" topic currently recommends using the Azure SignalR Service:

"We recommend using the [Azure SignalR Service](https://learn.microsoft.com/en-us/aspnet/core/signalr/scale?view=aspnetcore-8.0#azure-signalr-service) for server-side Blazor apps. The service works in conjunction with the app's Blazor Hub for scaling up a server-side Blazor app to a large number of concurrent SignalR connections. In addition, the SignalR Service's global reach and high-performance data centers significantly aid in reducing latency due to geography."

The recommendation is a bit too broad and can be mistaken for a requirement. While the Azure SignalR Service does help with connection scale out and provides global reach, this may not be needed depending on your scenario. For example, Azure App Service can now support a large number of concurrent WebSocket connections and already provides global distribution, so the Azure SignalR Service is no longer needed for this case (it was recommended in the past for Azure App Service due to infrastructure limitations that have since been addressed).

In other more constrained hosting environments, the Azure SignalR Service may still be beneficial, so the current content for using the Azure SignalR Service is still relevant, but we should clarify that it's use is not required.

Proposed edit:

```diff
- We recommend using the [Azure SignalR Service](https://learn.microsoft.com/en-us/aspnet/core/signalr/scale?view=aspnetcore-8.0#azure-signalr-service) for server-side Blazor apps. The service works in conjunction with the app's Blazor Hub for scaling up a server-side Blazor app to a large number of concurrent SignalR connections. In addition, the SignalR Service's global reach and high-performance data centers significantly aid in reducing latency due to geography.
+ Consider using the [Azure SignalR Service](https://learn.microsoft.com/en-us/aspnet/core/signalr/scale?view=aspnetcore-8.0#azure-signalr-service) with Blazor apps using interactive server rendering. The service works in conjunction with the app's Blazor Hub for scaling up a server-side Blazor app to a large number of concurrent SignalR connections. In addition, the SignalR Service's global reach and high-performance data centers significantly aid in reducing latency due to geography. If your hosting environment already handles these concerns, that using the Azure SignalR Service is not necessary.
```

We should also update the "Azure App Service without the Azure SignalR Service" section so clarify that using the Azure SignalR Service when hosted in Azure App Service is not necessary.

Proposed edits:

```diff
[...]

PR #33593, opened 2024-09-12
Azure App Service/Azure SignalR Service updates

### LINKED ISSUE #33591: Clarify use of the session affinity swtich in Azure (Configuration > General settings)
### Description

Appears here and is mentioned in the *Fundamentals* > *Signal* article.

When is manually flipping the session affinity switch required? ... or if using any of the listed approaches, is the switch flipped auto-magically ✨ or is it ignored when set to **Off**?

Also, perform some re-org on the *Host and deploy* section. List the approaches plainly. Condense the NOTE content and drop the NOTE if possible in favor of article text.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/server?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/host-and-deploy/server.md

### Document ID

2712f7a2-90e7-0012-f7d9-f3a7e7e345d7

### Article author

@guardrex

guardrex:
Fixes #33591

### DIFF
--- a/aspnetcore/blazor/fundamentals/signalr.md
+++ b/aspnetcore/blazor/fundamentals/signalr.md
@@ -181,17 +181,19 @@ To resolve the problem, use ***either*** of the following approaches:
 
 ## Use session affinity (sticky sessions) for server-side webfarm hosting
 
-A Blazor app prerenders in response to the first client request, which creates UI state on the server. When the client attempts to create a SignalR connection, **the client must reconnect to the same server**. When more than one backend server is in use, the app should implement session affinity, also called *sticky sessions*, for SignalR connections.
+A Blazor app prerenders in response to the first client request, which creates UI state on the server. When the client attempts to create a SignalR connection, the client must reconnect to the same server. When more than one backend server is in use, the app should implement session affinity, also called *sticky sessions*, for SignalR connections.
 
 The following error is thrown by an app that hasn't enabled session affinity in a webfarm:
 
-> :::no-loc text="blazor.server.js:1 Uncaught (in promise) Error: Invocation canceled due to the underlying connection being closed.":::
+> :::no-loc text="Uncaught (in promise) Error: Invocation canceled due to the underlying connection being closed.":::
 
-## Server-side Azure SignalR Service
+For more information on session affinity with Azure App Service hosting, see <xref:blazor/host-and-deploy/server#azure-app-service>.
+
+## Azure SignalR Service
 
 We recommend using the [Azure SignalR Service](xref:signalr/scale#azure-signalr-service) for server-side development hosted in Microsoft Azure. The service works in conjunction with the app's Blazor Hub for scaling up a server-side app to a large number of concurrent SignalR connections. In addition, the SignalR Service's global reach and high-performance data centers significantly aid in reducing latency due to geography.
 
-[Session affinity](#use-session-affinity-sticky-sessions-for-server-side-webfarm-hosting) is enabled for the Azure SignalR Service by setting the service's `ServerStickyMode` option or configuration value to `Required`. For more information, see <xref:blazor/host-and-deploy/server#azure-signalr-service>.
+For more information, see <xref:blazor/host-and-deploy/server#azure-signalr-service>.
 
 ## Server-side circuit handler options
 
--- a/aspnetcore/blazor/host-and-deploy/server.md
+++ b/aspnetcore/blazor/host-and-deploy/server.md
@@ -66,87 +66,52 @@ Recommendations for global deployments to geographical data centers:
 
 :::moniker-end
 
+## Azure App Service
+
+Hosting on Azure App Service requires configuration for WebSockets and session affinity, also called Application Request Routing (ARR) affinity.
+
+See the following resources:
+
+* [Configure the app in Azure App Service](xref:signalr/publish-to-azure-web-app#configure-the-app-in-azure-app-service)
+* [App Service Plan Limits](xref:signalr/publish-to-azure-web-app#app-service-plan-limits)
+
+If the app is deployed to a single server to run in a single process, the app doesn't require using the [Azure SignalR Service](#azure-signalr-service).
+
+If the app is deployed to a webfarm (multiple servers), the [Azure SignalR Service](#azure-signalr-service) is recommended.
+
 ## Azure SignalR Service
+
+The [Azure SignalR Service](xref:signalr/scale#azure-signalr-service) works in conjunction with the app's Blazor hub for scaling up to a large number of concurrent SignalR connections. In addition, the service's global reach and high-performance data centers significantly aid in reducing latency due to geography. If your hosting environment already handles these concerns, using the Azure SignalR Service isn't necessary.
 
 :::moniker range=">= aspnetcore-8.0"
 
-For Blazor Web Apps that adopt interactive server-side rendering, consider using the [Azure SignalR Service](xref:signalr/scale#azure-signalr-service). The service works in conjunction with the app's Blazor Hub for scaling up to a large number of concurrent SignalR connections. In addition, the service's global reach and high-performance data centers significantly aid in reducing latency due to geography. If your hosting environment already handles these concerns, using the Azure SignalR Service isn't necessary.
-
-<!-- UPDATE 9.0 Remove when support is present -->
+<!-- UPDATE 9.0 Update section to only cross-link stateful
+                reconnect guidance after the feature is
+                supported with Azure SignalR Service. -->
 
 > [!NOTE]
 > [Stateful reconnect](xref:signalr/configuration#configure-stateful-reconnect) (<xref:Microsoft.AspNetCore.SignalR.Client.HubConnectionBuilderHttpExtensions.WithStatefulReconnect%2A>) was released with .NET 8 but isn't currently supported for the Azure SignalR Service. For more information, see [Stateful Reconnect Support? (`Azure/azure-signalr` #1878)](https://github.com/Azure/azure-signalr/issues/1878).
 
 :::moniker-end
 
-:::moniker range="< aspnetcore-8.0"
+For guidance on deployment and configuration for the Azure SignalR Service:
 
-Consider using the [Azure SignalR Service](xref:signalr/scale#azure-signalr-service), which works in conjunction with the app's Blazor Hub for scaling up to a large number of concurrent SignalR connections. In addition, the service's global reach and high-performance data centers significantly aid in reducing latency due to geography. If your hosting environment already handles these concerns, using the Azure SignalR Service isn't necessary.
+* Only when the app is in production on Azure App Service, see <xref:signalr/publish-to-azure-web-app>.
+* When you want the app to adopt the Azure SignalR Service during development, see [Build Blazor Server chat app: Further topic: Enable Azure SignalR Service in local development](https://github.com/aspnet/AzureSignalR-samples/tree/main/samples/BlazorChat#further-topic-enable-azure-signalr-service-in-local-development).
 
-:::moniker-end
+We recommend using WebSockets for server-side Blazor apps deployed to Azure App Service. If [WebSockets](https://wikipedia.org/wiki/WebSocket) are disabled, Azure App Service simulates a real-time connection using HTTP Long Polling. HTTP Long Polling is noticeably slower than running with WebSockets enabled, which doesn't use polling to simulate a client-server connection.
 
-> [!IMPORTANT]
-> When [WebSockets](https://wikipedia.org/wiki/WebSocket) are disabled, Azure App Service simulates a real-time connection using HTTP Long Polling. HTTP Long Polling is noticeably slower than running with WebSockets enabled, which doesn't use polling to simulate a client-server connection. In the event that Long Polling must be used, you may need to configure the maximum poll interval (`MaxPollIntervalInSeconds`), which defines the maximum poll interval allowed for Long Polling connections in [Azure SignalR Service](#azure-signalr-service) if the service ever falls back from WebSockets to Long Polling. If the next poll request does not come in within `MaxPollIntervalInSeconds`, Azure SignalR Service cleans up the client connection. Note that Azure SignalR Service also cleans up connections when cached waiting to write buffer size is greater than 1 MB to ensure service performance. Default value for `MaxPollIntervalInSeconds` is 5 seconds. The setting is limited to 1-300 seconds.
->
-> **We recommend using WebSockets for server-side Blazor apps deployed to Azure App Service.** The [Azure SignalR Service](xref:signalr/scale#azure-signalr-service) uses WebSockets. If the app doesn't use the Azure SignalR Service, see <xref:signalr/publish-to-azure-web-app#configure-the-app-in-azure-app-service>.
->
-> For more information, see:
->
-> * [What is Azure SignalR Service?](/azure/azure-signalr/signalr-overview)
-> * [Performance guide for Azure SignalR Service](/azure/azure-signalr/signalr-concept-performance#performance-factors)
-> * <xref:signalr/publish-to-azure-web-app>
+In the event that Long Polling must be used, you may need to configure the maximum poll interval (`MaxPollIntervalInSeconds`, default: 5 seconds, limit: 1-300 seconds), which defines the maximum poll interval allowed for Long Polling connections in the Azure SignalR Service if the service ever falls back from WebSockets to Long Polling. If the next poll request doesn't arrive within the maximum poll interval, Azure SignalR Service closes the client connection. The Azure SignalR Service also cleans up connections when the cached waiting-to-write buffer size is greater than 1 MB for improved service performance.
 
-The app must support *session affinity*, also called *sticky sessions*, where clients are redirected back to the same server. To support session affinity, an application setting in the Azure portal is automatically configured (**Configuration** > **Application settings**: **Name**: `Azure__SignalR__ServerStickyMode`, **Value**: `Required`). Therefore, you don't need to manually configure the app for session affinity.
+For more information, see:
 
-If you prefer not to use the app setting in Azure (and delete it in the Azure portal), two other approaches for configuring session affinity are (**use either approach, not both**):
-
-* Set the option in `Program.cs`:
-
-  ```csharp
-  builder.Services.AddSignalR().AddAzureSignalR(options =>
-  {
-      options.ServerStickyMode = 
-          Microsoft.Azure.SignalR.ServerStickyMode.Required;
-  });
-  ```
-
-* Configure the option in `appsettings.json`:
-
-  ```json
-  "Azure:SignalR:ServerStickyMode": "Required"
-  ```
-
-> [!NOTE]
-> The following error is thrown by an app that hasn't enabled session affinity:
->
-> > :::no-loc text="blazor.server.js:1 Uncaught (in promise) Error: Invocation canceled due to the underlying connection being closed.":::
-
-To provision the Azure SignalR Service for an app in Visual Studio:
-
-1. Create an Azure Apps publish profile in Visual Studio for the app.
-1. Add the **Azure SignalR Service** dependency to the profile. If the Azure subscription doesn't have a pre-existing Azure SignalR Service instance to assign to the app, select **Create a new Azure SignalR Service instance** to provision a new service instance.
-1. Publish the app to Azure.
-
-Provisioning the Azure SignalR Service in Visual Studio automatically adds the SignalR connection string to the app service's configuration.
-
-## Azure App Service
-
-:::moniker range=">= aspnetcore-8.0"
-
-Hosting a Blazor Web App that uses interactive server-side rendering on Azure App Service requires configuration for Application Request Routing (ARR) affinity and WebSockets. The App Service should also be appropriately globally distributed to reduce UI latency. Using the Azure SignalR Service when hosting on Azure App Service isn't required.
-
-:::moniker-end
-
-:::moniker range="< aspnetcore-8.0"
-
-Hosting a Blazor Server app on Azure App Service requires configuration for Application Request Routing (ARR) affinity and WebSockets. The App Service should also be appropriately globally distributed to reduce UI latency. Using the Azure SignalR Service when hosting on Azure App Service isn't required.
-
-:::moniker-end
-
-1. Use the following guidance to configure the service:
-
-  * [Configure the app in Azure App Service](xref:signalr/publish-to-azure-web-app#configure-the-app-in-azure-app-service).
-  * [App Service Plan Limits](xref:signalr/publish-to-azure-web-app#app-service-plan-limits).
+* [Azure SignalR Service](/products/signalr-service/)
+* [What is Azure SignalR Service?](/azure/azure-signalr/signalr-overview)
+* <xref:signalr/scale#azure-signalr-service>
+* [Performance guide for Azure SignalR Service](/azure/azure-signalr/signalr-concept-performance#performance-factors)
+* <xref:signalr/publish-to-azure-web-app>
+* <xref:signalr/configuration>
+* [Build Blazor Server chat app](https://github.com/aspnet/AzureSignalR-samples/tree/main/samples/BlazorChat)
 
 :::moniker range=">= aspnetcore-6.0"
 
--- a/aspnetcore/fundamentals/owin.md
+++ b/aspnetcore/fundamentals/owin.md
@@ -94,7 +94,7 @@ app.UseOwin(pipeline =>
 
 ## Run ASP.NET Core on an OWIN-based server and use its WebSockets support
 
-Another example of how OWIN-based servers' features can be leveraged by ASP.NET Core is access to features like WebSockets. The .NET OWIN web server used in the previous example has support for Web Sockets built in, which can be leveraged by an ASP.NET Core application. The example below shows a simple web app that supports Web Sockets and echoes back everything sent to the server through WebSockets.
+Another example of how OWIN-based servers' features can be leveraged by ASP.NET Core is access to features like WebSockets. The .NET OWIN web server used in the previous example has support for WebSockets built in, which can be leveraged by an ASP.NET Core application. The example below shows a simple web app that supports WebSockets and echoes back everything sent to the server through WebSockets.
 
 ```csharp
 public class Startup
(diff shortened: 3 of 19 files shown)


aspnetcore/blazor/host-and-deploy/server.md
```diff
@@ -66,87 +66,52 @@ Recommendations for global deployments to geographical data centers:
...
-> * <xref:signalr/publish-to-azure-web-app>
+* [Configure the app in Azure App Service](xref:signalr/publish-to-azure-web-app#configure-the-app-in-azure-app-service)
+* [App Service Plan Limits](xref:signalr/publish-to-azure-web-app#app-service-plan-limits)
 
-The app must support *session affinity*, also called *sticky sessions*, where clients are redirected back to the same server. To support session affinity, an application setting in the Azure portal is automatically configured (**Configuration** > **Application settings**: **Name**: `Azure__SignalR__ServerStickyMode`, **Value**: `Required`). Therefore, you don't need to manually configure the app for session affinity.
+If the app is deployed to a single server to run in a single process, the app doesn't require using the [Azure SignalR Service](#azure-signalr-service).
 
-If you prefer not to use the app setting in Azure (and delete it in the Azure portal), two other approaches for configuring session affinity are (**use either approach, not both**):
+If the app is deployed to a webfarm (multiple servers), the [Azure SignalR Service](#azure-signalr-service) is recommended.
```
danroth27, line 80: I don't think this is accurate. If you're Blazor app is hosted on Azure App Service, then you don't really need the Azure SignalR Service

aspnetcore/blazor/host-and-deploy/server.md
```diff
@@ -66,87 +66,52 @@ Recommendations for global deployments to geographical data centers:
...
 
 :::moniker-end
 
-:::moniker range="< aspnetcore-8.0"
+For guidance on deployment and configuration for the Azure SignalR Service:
 
-Hosting a Blazor Server app on Azure App Service requires configuration for Application Request Routing (ARR) affinity and WebSockets. The App Service should also be appropriately globally distributed to reduce UI latency. Using the Azure SignalR Service when hosting on Azure App Service isn't required.
+* Only when the app is in production on Azure App Service, see <xref:signalr/publish-to-azure-web-app>.
+* When you want the app to adopt the Azure SignalR Service during development, see [Build Blazor Server chat app: Further topic: Enable Azure SignalR Service in local development](https://github.com/aspnet/AzureSignalR-samples/tree/main/samples/BlazorChat#further-topic-enable-azure-signalr-service-in-local-development).
```
danroth27, line 100: Again, if your Blazor app is hosted on Azure App Service, then I don't believe the Azure SignalR Service is needed.

aspnetcore/blazor/host-and-deploy/server.md
```diff
@@ -66,87 +66,52 @@ Recommendations for global deployments to geographical data centers:
...
->
-> For more information, see:
->
-> * [What is Azure SignalR Service?](/azure/azure-signalr/signalr-overview)
-> * [Performance guide for Azure SignalR Service](/azure/azure-signalr/signalr-concept-performance#performance-factors)
-> * <xref:signalr/publish-to-azure-web-app>
+* [Configure the app in Azure App Service](xref:signalr/publish-to-azure-web-app#configure-the-app-in-azure-app-service)
+* [App Service Plan Limits](xref:signalr/publish-to-azure-web-app#app-service-plan-limits)
 
-The app must support *session affinity*, also called *sticky sessions*, where clients are redirected back to the same server. To support session affinity, an application setting in the Azure portal is automatically configured (**Configuration** > **Application settings**: **Name**: `Azure__SignalR__ServerStickyMode`, **Value**: `Required`). Therefore, you don't need to manually configure the app for session affinity.
```
danroth27, line 99: I believe sticky sessions are still required when using the Azure SignalR Service with Blazor.

aspnetcore/blazor/host-and-deploy/server.md
```diff
@@ -66,87 +66,52 @@ Recommendations for global deployments to geographical data centers:
...
 
-  * [Configure the app in Azure App Service](xref:signalr/publish-to-azure-web-app#configure-the-app-in-azure-app-service).
-  * [App Service Plan Limits](xref:signalr/publish-to-azure-web-app#app-service-plan-limits).
+* [Azure SignalR Service](/products/signalr-service/)
+* [What is Azure SignalR Service?](/azure/azure-signalr/signalr-overview)
+* <xref:signalr/scale#azure-signalr-service>
+* [Performance guide for Azure SignalR Service](/azure/azure-signalr/signalr-concept-performance#performance-factors)
+* <xref:signalr/publish-to-azure-web-app>
+* <xref:signalr/configuration>
+* [Build Blazor Server chat app](https://github.com/aspnet/AzureSignalR-samples/tree/main/samples/BlazorChat)
```
danroth27, line 114: I get a 404 for the first link in this list.

Some of the other links seem superfluous or irrelevant. How about this?:

```suggestion
* [What is Azure SignalR Service?](/azure/azure-signalr/signalr-overview)
* <xref:signalr/scale#azure-signalr-service>
* <xref:signalr/publish-to-azure-web-app>
```

danroth27 (changes requested):
Lots of good changes here. A few comments about the Blazor guidance for using the Azure SignalR Service. Related past issue: https://github.com/dotnet/AspNetCore.Docs/issues/32199

aspnetcore/blazor/host-and-deploy/server.md
guardrex: > For Blazor apps, we recommend using the Azure SignalR Service to:
>
> * Facilitate connection scale out
> * Handle global distribution

That's why I used the word "recommended." On more than one instance, isn't it recommended? And, doesn't "scale out" mean more than one instance?

The SignalR doc recommends it for scale out ... last line of the section (under the image) at ...

https://learn.microsoft.com/en-us/aspnet/core/signalr/scale?view=aspnetcore-8.0#scale-out

... and the first line of ...

https://learn.microsoft.com/en-us/aspnet/core/signalr/scale?view=aspnetcore-8.0#azure-signalr-service

aspnetcore/blazor/host-and-deploy/server.md
guardrex: Confused ... that isn't what this states. This is in the section on the Azure SignalR Service and states that ...

* If you want to use it in production, see the cross-linked guidance (Line 99).
* If you want to use it locally, see the cross-linked guidance (Line 100).

This doesn't make any statements about it being required. An earlier line even says ...

> If your hosting environment already handles these concerns, using the Azure SignalR Service isn't necessary.

aspnetcore/blazor/host-and-deploy/server.md
guardrex: WIP ... 

That first link should go here ... https://azure.microsoft.com/en-us/products/signalr-service/ ... to the main MS page that advertises it. I clipped it for the learn domain ... ***BAD*** 🦖! 😄 

The perf link was just a holdover from ages ago.

The chat app link was added because thus far I'm relying on that cross-link to explain the use of the service, including the app configuration for it. One of the things that this PR is seeking to do is cross-link for the guidance in more stable locations instead of trying to keep up with it here.

aspnetcore/blazor/host-and-deploy/server.md
guardrex: That's right, but this is covered by the cross-linked guidance in the Azure SignalR Service section. I'm trying to let the cross-linked content from the Azure SignalR Service section cover it. However, this is a WIP for tomorrow morning. Perhaps, I'll think in the morning that we need to duplicate the external guidance here ... or you can let me know that that's exactly what you'd like to do.

aspnetcore/blazor/host-and-deploy/server.md
danroth27: When you publish to Azure App Service with multiple instances you're effectively on a web farm, but I believe you still don't need the Azure SignalR Service.

With the Azure SignalR Service, connection scale out means that it will multiplex multiple connections to reduce the connection load on your servers: https://learn.microsoft.com/azure/azure-signalr/signalr-concept-internals#application-server-connections.

Global distribution means the Azure SignalR Service is available in multiple regions and backed by a fast network.

aspnetcore/blazor/host-and-deploy/server.md
danroth27: > The chat app link was added because thus far I'm relying on that cross-link to explain the use of the service, including the app configuration for it. One of the things that this PR is seeking to do is cross-link for the guidance in more stable locations instead of trying to keep up with it here.

I agree with this idea in principle, but that sample needs some love before I'm comfortable linking to it. The sample README says to install .NET Core 3.0 and VS 2019 😆.

Even if we eventually link to that sample, this page still needs to be our official guidance on how to use the Azure SignalR Service with Blazor, so that it's part of the official Blazor doc set (indexed for search, localized, etc.)

aspnetcore/blazor/host-and-deploy/server.md
danroth27: I'm saying that there is really never a need to use the Azure SignalR Service if your Blazor app is hosted on Azure App Service. So, the first bullet on line 99 is providing guidance for a non-scenario for Blazor apps. I also don't know of a scenario for using the Azure SignalR Service with Blazor for local development.

aspnetcore/blazor/host-and-deploy/server.md
guardrex: Yes 😄 ... This is a WIP, and my pattern is not to cross-link when I don't like what I see (e.g., Azure Graph/AD groups & roles ... *I still like our stuff better!*). I have a feeling that I'll drop the cross-link for our main coverage tomorrow and just carry the guidance here. However, I should look around a little more for guidance similar to what that README has that I can cross-link. It's just that I think they are on the right track with how they summarized everything.

aspnetcore/blazor/host-and-deploy/server.md
guardrex: Ok ... gotcha. According to the chat sample, it's a thing ...

https://github.com/aspnet/AzureSignalR-samples/tree/main/samples/BlazorChat#further-topic-enable-azure-signalr-service-in-local-development

... and my understanding is that if the app is deployed with the service (profile publish out of VS) ***the service isn't local***. The service works on the Azure App instance(s) after deployment. None of the items they cover in that section ☝️ need to be done with the app, and none of the app's settings need to be adjusted in the app or in Azure. That's what I'm gleaning from their coverage.

aspnetcore/blazor/host-and-deploy/server.md
guardrex: Yes, session affinity and WebSockets are the only requirements from what I'm reading.

Still tho, I was hoping for a firmer recommendation for using the service, which we know offloads those two items from the app. I guess I should just go with your text ...

> For Blazor apps, we recommend using the Azure SignalR Service to:
>
> Facilitate connection scale out
> Handle global distribution

... and the cross-links and let the reader decide from there. I was probably reading too much into the references. They seem to kind'a strongly recommend it for app scale-out (multiple instances/multiple servers), and maybe that's because $$ is at stake.

danroth27: There are other reasons to use the Azure SignalR Service when working directly with SignalR (like using the service as a SignalR backplane), but with Blazor the benefits are more scoped.

guardrex:
@danroth27 ... Should be an improvement 🤞. There are still some gray areas, and the condition on using the service still seems a bit vague.

Note that the multiple instance/server guidance for enabling session affinity currently only has the dev confirm **Session affinity** is **On** in the App Service registration. I know that the dev must set `ServerStickyMode` for the Azure SignalR Service, and that's covered here. If the app isn't using the Azure SignalR Service, do they still need to set `ServerStickyMode`?

danroth27:
> One more thing ... I have this generally recommending session affinity. At the end of the App Service section, it says that you can disable it for a single instance. Elsewhere, it reads like we generally recommend leaving session affinity ON. Let me know if you'd like that revised to tell single-instance folks to turn it OFF and turn it on for multiple-instance folks.

I think it's best to always recommend enabling session affinity for Blazor Server based apps. That way the app continues to work when they move from one instance to multiple and I don't know of any downsides for the single instance case.

guardrex:
Very well ... that's what the PR has it doing.

danroth27:
> If the app isn't using the Azure SignalR Service, do they still need to set ServerStickyMode?

ServerStickyMode is an Azure SignalR Service specific setting, so no, it's not needed if they aren't using that service.

aspnetcore/blazor/fundamentals/signalr.md
```diff
@@ -181,17 +181,24 @@ To resolve the problem, use ***either*** of the following approaches:
...
 
-[Session affinity](#use-session-affinity-sticky-sessions-for-server-side-webfarm-hosting) is enabled for the Azure SignalR Service by setting the service's `ServerStickyMode` option or configuration value to `Required`. For more information, see <xref:blazor/host-and-deploy/server#azure-signalr-service>.
+The optional [Azure SignalR Service](xref:signalr/scale#azure-signalr-service) works in conjunction with the app's SignalR hub for scaling up a server-side app to a large number of concurrent connections. In addition, the service's global reach and high-performance data centers significantly aid in reducing latency due to geography.
+
+The service isn't required for Blazor apps hosted in Azure App Service but can be helpful in other hosting environments:
+
+* To facilitate connection scale out.
+* Handle global distribution.
+
+For more information, see <xref:blazor/host-and-deploy/server#azure-signalr-service>.
```
danroth27, line 201: I'd like to generalize this guidance beyond just the Azure App Service case. Basically, you don't get any benefit out of using the Azure SignalR Service with your Blazor Server app if your environment already can handle the WebSocket connection load and is globally distributed. Azure App Service already handles these things, but so does Azure Container Apps and potentially other environments.

```suggestion
The service isn't required for Blazor apps hosted in Azure App Service or Azure Container Apps but can be helpful in other hosting environments:

* To facilitate connection scale out.
* Handle global distribution.

For more information, see <xref:blazor/host-and-deploy/server#azure-signalr-service>.
```

aspnetcore/blazor/host-and-deploy/server.md
```diff
@@ -62,91 +62,103 @@ Recommendations for global deployments to geographical data centers:
 
 * Deploy the app to the regions where most of the users reside.
 * Take into consideration the increased latency for traffic across continents. To control the appearance of the reconnection UI, see <xref:blazor/fundamentals/signalr#control-when-the-reconnection-ui-appears>.
-* For Azure hosting, use the [Azure SignalR Service](#azure-signalr-service).
+* For Azure hosting outside of Azure App Service, use the [Azure SignalR Service](#azure-signalr-service).
```
danroth27, line 65: We want to recommend the Azure SignalR Service only when it's needed. We know it's not needed on Azure App Service and Azure Container apps, but there may also be other Blazor hosting environments that don't benefit from it.

```suggestion
* Consider using the [Azure SignalR Service](#azure-signalr-service).
```

aspnetcore/blazor/host-and-deploy/server.md
```diff
@@ -62,91 +62,103 @@ Recommendations for global deployments to geographical data centers:
...
+* Session affinity to route requests from a user back to the same App Service instance. The default setting is **On**.
 
-:::moniker-end
+1. In the Azure portal, navigate to the web app in **App Services**.
+1. Open **Settings** > **Configuration**.
+1. Set **Web sockets** to **On**.
+1. Verify that **Session affinity** is set to **On**.
 
-:::moniker range="< aspnetcore-8.0"
+If you wish to disable the session affinity feature for a Blazor app that runs on a single Azure instance, see [Disable Session affinity cookie (ARR cookie) for Azure web apps](https://azure.github.io/AppService/2016/05/16/Disable-Session-affinity-cookie-(ARR-cookie)-for-Azure-web-apps.html).
```
danroth27, line 86: I'm not aware of any benefit of doing this, and the app will break if they later decide to scale it out. I think we can just remove this text.

```suggestion
If you wish to disable the session affinity feature for a Blazor app that runs on a single Azure instance, see [Disable Session affinity cookie (ARR cookie) for Azure web apps](https://azure.github.io/AppService/2016/05/16/Disable-Session-affinity-cookie-(ARR-cookie)-for-Azure-web-apps.html).
```

aspnetcore/blazor/host-and-deploy/server.md
```diff
@@ -62,91 +62,103 @@ Recommendations for global deployments to geographical data centers:
...
->
-> **We recommend using WebSockets for server-side Blazor apps deployed to Azure App Service.** The [Azure SignalR Service](xref:signalr/scale#azure-signalr-service) uses WebSockets. If the app doesn't use the Azure SignalR Service, see <xref:signalr/publish-to-azure-web-app#configure-the-app-in-azure-app-service>.
->
-> For more information, see:
->
-> * [What is Azure SignalR Service?](/azure/azure-signalr/signalr-overview)
-> * [Performance guide for Azure SignalR Service](/azure/azure-signalr/signalr-concept-performance#performance-factors)
-> * <xref:signalr/publish-to-azure-web-app>
+The service isn't required for Blazor apps hosted in Azure App Service but can be helpful in other hosting environments:
```
danroth27, line 92:
```suggestion
The service isn't required for Blazor apps hosted in Azure App Service or Azure Container Apps but can be helpful in other hosting environments:
```

aspnetcore/blazor/host-and-deploy/server.md
```diff
@@ -62,91 +62,103 @@ Recommendations for global deployments to geographical data centers:
...
+> [!NOTE]
+> [Stateful reconnect](xref:signalr/configuration#configure-stateful-reconnect) (<xref:Microsoft.AspNetCore.SignalR.Client.HubConnectionBuilderHttpExtensions.WithStatefulReconnect%2A>) was released with .NET 8 but isn't currently supported for the Azure SignalR Service. For more information, see [Stateful Reconnect Support? (`Azure/azure-signalr` #1878)](https://github.com/Azure/azure-signalr/issues/1878).
 
-* Configure the option in `appsettings.json`:
+:::moniker-end
 
-  ```json
-  "Azure:SignalR:ServerStickyMode": "Required"
-  ```
+We recommend using WebSockets for Blazor apps deployed to Azure App Service. If [WebSockets](https://wikipedia.org/wiki/WebSocket) are disabled, Azure App Service simulates a real-time connection using HTTP Long Polling. HTTP Long Polling is noticeably slower than running with WebSockets enabled, which doesn't use polling to simulate a client-server connection.
```
danroth27, line 108: This text seems to be more about Azure App Service than the Azure SignalR service.  Could it be removed if we provide sufficient context in the following paragraph?

```suggestion
```

aspnetcore/blazor/host-and-deploy/server.md
```diff
@@ -62,91 +62,103 @@ Recommendations for global deployments to geographical data centers:
...
 
-> [!NOTE]
-> The following error is thrown by an app that hasn't enabled session affinity:
->
-> > :::no-loc text="blazor.server.js:1 Uncaught (in promise) Error: Invocation canceled due to the underlying connection being closed.":::
+In the event that Long Polling must be used, you may need to configure the maximum poll interval (`MaxPollIntervalInSeconds`, default: 5 seconds, limit: 1-300 seconds), which defines the maximum poll interval allowed for Long Polling connections in the Azure SignalR Service if the service ever falls back from WebSockets to Long Polling. If the next poll request doesn't arrive within the maximum poll interval, the service closes the client connection. The service also cleans up connections when the cached waiting-to-write buffer size is greater than 1 MB for improved service performance.
```
danroth27, line 110: @BrennanConroy @mikekistler Is this guidance about long polling specific to using Blazor with the Azure SignalR Service, or does it really apply to all SignalR apps when used with the service?

```suggestion
In the event that HTTP Long Polling must be used instead of WebSockets with the Azure SignalR Service, you may need to configure the maximum poll interval (`MaxPollIntervalInSeconds`, default: 5 seconds, limit: 1-300 seconds), which defines the maximum poll interval allowed for Long Polling connections in the Azure SignalR Service if the service ever falls back from WebSockets to Long Polling. If the next poll request doesn't arrive within the maximum poll interval, the service closes the client connection. The service also cleans up connections when the cached waiting-to-write buffer size is greater than 1 MB for improved service performance.
```

BrennanConroy: Sounds generic. You probably should have ASRS folks look this doc over.

aspnetcore/blazor/host-and-deploy/server.md
```diff
@@ -62,91 +62,103 @@ Recommendations for global deployments to geographical data centers:
...
-To provision the Azure SignalR Service for an app in Visual Studio:
+### Add the service as a dependency to a production deployment
 
-1. Create an Azure Apps publish profile in Visual Studio for the app.
-1. Add the **Azure SignalR Service** dependency to the profile. If the Azure subscription doesn't have a pre-existing Azure SignalR Service instance to assign to the app, select **Create a new Azure SignalR Service instance** to provision a new service instance.
-1. Publish the app to Azure.
+For guidance, see <xref:signalr/publish-to-azure-web-app>.
 
-Provisioning the Azure SignalR Service in Visual Studio automatically adds the SignalR connection string to the app service's configuration.
+### Use the service locally
```
danroth27, line 116: I don't think we really need to document here how to use the Azure SignalR Service locally during development with a Blazor app. I just don't think it's a common scenario for Blazor developers.

@mikekistler @BrennanConroy Is this scenario important for SignalR devs? Should this content maybe move to the https://review.learn.microsoft.com/aspnet/core/signalr/publish-to-azure-web-app doc?

aspnetcore/blazor/host-and-deploy/server.md
guardrex: Let me know if I should try to find out where/who/when this came from.

aspnetcore/blazor/host-and-deploy/server.md
guardrex: BTW @danroth27 ... This was my attempt to avoid cross-linking the chat sample, which has this, given your remarks about that sample requiring an update.

aspnetcore/blazor/host-and-deploy/server.md
danroth27: @chenkennt Can you take a look at this guidance on using HTTP long polling with the Azure SignalR Service? I think this text was added in response to [this GitHub issue discussion](https://github.com/dotnet/aspnetcore/issues/42778#issuecomment-1188852129), but I'm not sure if this content is still relevant and if this is the right place for it.

aspnetcore/blazor/host-and-deploy/server.md
danroth27: > This was my attempt to avoid cross-linking the chat sample, which has this, given your remarks about that sample requiring an update.

Yup, I know. I don't think we need this content in the Blazor docs, but maybe it makes sense to add to the SignalR docs. Let's see what @mikekistler and @BrennanConroy say.

guardrex:
I'll commit the remaining updates after the discussions are concluded.

*NOTE TO SELF* ... Removed paragraph for changes to Line 108 (and there's an extra line to remove at 107) ...

> We recommend using WebSockets for Blazor apps deployed to Azure App Service. If \[WebSockets](https://wikipedia.org/wiki/WebSocket) are disabled, Azure App Service simulates a real-time connection using HTTP Long Polling. HTTP Long Polling is noticeably slower than running with WebSockets enabled, which doesn't use polling to simulate a client-server connection.

aspnetcore/blazor/host-and-deploy/server.md
danroth27: The `MaxPollIntervalInSeconds` content is consistent with the [related Azure SignalR Service docs](https://learn.microsoft.com/azure/azure-signalr/signalr-howto-use#maxpollintervalinseconds). However, it looks like the last statement about the cached waiting-to-write buffer size was [removed from the Azure SignalR Service docs](https://github.com/MicrosoftDocs/azure-docs-pr/pull/272885) by @vicancy, so I believe we should remove it here as well.

```suggestion
In the event that Long Polling must be used, you may need to configure the maximum poll interval (`MaxPollIntervalInSeconds`, default: 5 seconds, limit: 1-300 seconds), which defines the maximum poll interval allowed for Long Polling connections in the Azure SignalR Service if the service ever falls back from WebSockets to Long Polling. If the next poll request doesn't arrive within the maximum poll interval, the service closes the client connection.
```

aspnetcore/blazor/host-and-deploy/server.md
guardrex: I removed the section of local use of the service. In case the content that I placed is of use in the SignalR docs, here it is as a cut-'n-paste, and I can place it using this PR 👂 ...

> \#\#\# Use the service locally
>
> Add package reference to the app for \[\`Microsoft.Azure.SignalR\`](https://www.nuget.org/packages/Microsoft.Azure.SignalR).
> 
> \[!INCLUDE\[](~/includes/package-reference.md)]
> 
> Implement and configure the Azure SignalR Service in the app's service configuration of the `Program` file:
> 
> \`\`\`csharp
> builder.Services.AddSignalR().AddAzureSignalR();
> \`\`\`
> 
> Configure the following for the service in the app settings file (\`appsettings.json\`):
> 
> \* Session affinity (sticky sessions) with the \`ServerStickyMode\` key. Set the value to \`Required\`.
> \* Connection string (\`{CONNECTION STRING}\` placeholder in the following example). To avoid maintaining the connection string in plain text in the file, use the \[Secret Manager tool](xref:security/app-secrets) to supply the string.
> 
> In \`appsettings.json\`:
> 
> \`\`\`json
> "Azure": {
>   "SignalR": {
>     "Enabled": true,
>     "ServerStickyMode": "Required",
>     "ConnectionString": {CONNECTION STRING}
>   }
> }
> \`\`\`
> 
> Assign the Azure SignalR Service SDK as a hosting startup assembly. Edit \`Properties/launchSettings.json\`. Add an \`ASPNETCORE_HOSTINGSTARTUPASSEMBLIES\` entry to \`environmentVariables\` of the \`https\` profile with a value of \`Microsoft.Azure.SignalR\`:
> 
> \`\`\`json
> "environmentVariables": {
>     "ASPNETCORE_ENVIRONMENT": "Development",
>     "ASPNETCORE_HOSTINGSTARTUPASSEMBLIES": "Microsoft.Azure.SignalR"
> }
> \`\`\`

danroth27 (approved):
:shipit:

PR #26136, opened 2022-06-13
Update Program.cs

ObliviousSnake:
[EDIT by guardrex to tie this to an issue]

Fixes #26137

app.MapHub does not work

PR #33594, opened 2024-09-12
MapHub is not present on ApplicationBuilder but on EndpointRouteBuilder

Gutek:
`MapHub<>` extension method is not present on `IApplicationBuilder` anymore, instead its on `IEndpointRouteBuilder`

Even in [source code](https://github.com/dotnet/aspnetcore/tree/main/src/SignalR/server/SignalR/src) you do not have any other extensions than to `IEndpointRouteBuilder`


Gutek:
@dotnet-policy-service agree

Gutek:
@dotnet-policy-bot agree

BrennanConroy:
See https://github.com/dotnet/AspNetCore.Docs/pull/26136

Gutek:
@BrennanConroy ok, as long as you use the new format.

thanks, this can be closed.

PR #33620, opened 2024-09-16
[Blazor] handle-errors.md - overriding async method with sync code

hakenr:
@guardrex I don't think it makes sense to await `Task.CompletedTask`.  
(The first example won't even compile - you have to use `async` in the method definition when you want to use `await` inside.)

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/handle-errors.md](https://github.com/dotnet/AspNetCore.Docs/blob/64fcca43e8f1d2199487d708dfcb792b765f78b6/aspnetcore/blazor/fundamentals/handle-errors.md) | [Handle errors in ASP.NET Core Blazor apps](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/handle-errors?branch=pr-en-us-33620) |

### DIFF
--- a/aspnetcore/blazor/fundamentals/handle-errors.md
+++ b/aspnetcore/blazor/fundamentals/handle-errors.md
@@ -454,7 +454,7 @@ The following example permits the user to recover from the exception with a butt
 }
 ```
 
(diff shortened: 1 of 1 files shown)


guardrex:
In the first example, the missing the `async` was my problem. My intention was to be able to await because dev code may do something asynchronous there. However, devs can figure that out easily enough on their own.

WRT the question you asked on the closed PR, that subject didn't come up with Mackinnon. You can open a new issue to ask about that.

guardrex (approved):
Thanks @hakenr! 🎷

PR #28306, opened 2023-02-04
File uploads file size limitations

guardrex:
Fixes #27680

Thanks @Kevin-Andrew! 🚀 

[Internal Review Topic (links to section)](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/file-uploads?view=aspnetcore-8.0&branch=pr-en-us-28306#file-size-limit)

* Remove the existing NOTE and replace it with a dedicated section on this subject.
* Pivot and version the content for Server vs. WASM and >=6.0 and <6.0 (for changes that occurred to lift the limit that only pertained to the `InputFile` component.
* Because `BrowserHttpHandler` is internal, it isn't in the API docs. Therefore, I'm going to code-fence it and supply the reference source link so that interested devs can track it down.

❓ ...

* ***Mackinnon/Pavel*** ... On [one issue (dotnet/aspnetcore #31873)](https://github.com/dotnet/aspnetcore/issues/31873#issuecomment-821031760), Javier recommended [Streaming requests with the fetch API](https://developer.chrome.com/articles/fetch-streaming-requests/); however, on [another issue (dotnet/aspnetcore #35899)](https://github.com/dotnet/aspnetcore/issues/35899#issuecomment-908183705), he recommended HTTP Ranges. This PR currently only has the HTTP Ranges suggestion. Do you want me to ***swap*** that for streaming fetch API or ***add*** streaming fetch API as another alternative? I understand that it isn't available for Safari and FF yet, so I'd need to work that in and then have a follow-up issue on hold to revise it later.

* ***Pavel*** ... I specifically mention that devs can add a 👍 to the [runtime issue (dotnet/runtime #36634)](https://github.com/dotnet/runtime/issues/36634) because it's merely marked "Future" at this time and you might have interest in gauging feedback for possibly placing it on the .NET 8 roadmap. Do you want me to strike the 👍 suggestion part of that and just leave the cross-link? I think at least the cross-link should be here because otherwise devs will think that the PU isn't aware of the problem/suggestion. I think we can curtail folks opening ***more issues*** on either the ASP.NET Core or runtime repos if we at least mention the issue. However, we can strike the 'you can add a 👍' part if you like.

Cross-references:

* [Blazor Support Large File Uploads (dotnet/aspnetcore #33638)](https://github.com/dotnet/aspnetcore/issues/33638)
[...]

PR #33693, opened 2024-09-23
Request Streaming upload

### LINKED ISSUE #33692: Request Streaming upload for client-side Blazor 9.0
### Description

Prior work at https://github.com/dotnet/AspNetCore.Docs/pull/28306.

Product unit PRs:

* https://github.com/dotnet/runtime/pull/91295
* https://github.com/dotnet/runtime/pull/91699

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/file-uploads?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/file-uploads.md

### Document ID

c11d981c-05af-c19d-a333-feedd5978639

### Article author

@guardrex

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+c11d981c-05af-c19d-a333-feedd5978639)

guardrex:
Fixes #33692

# 🚧👷‍♂️ ***WIP*** 🏗️🚧

### DIFF
--- a/aspnetcore/blazor/file-uploads.md
+++ b/aspnetcore/blazor/file-uploads.md
@@ -101,9 +101,32 @@ If you're using the [Autofac Inversion of Control (IoC) container](https://autof
 
 ## File size read and upload limits
 
-:::moniker range=">= aspnetcore-6.0"
+:::moniker range=">= aspnetcore-9.0"
+
+Server-side or client-side, there's no file read or upload size limit for the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component.
+
+For Chromium-based browsers (Google Chrome and Microsoft Edge) using the HTTP/2 protocol, client-side Blazor uses [Streams API](https://developer.mozilla.org/docs/Web/API/Streams_API) to permit uploading large files (> 2 GB or larger than the device's available memory).
+
+For non-Chromium browsers or without HTTP/2 protocol, client-side Blazor reads the file's bytes into a single JavaScript array buffer when marshalling the data from JavaScript to C#, which is limited to 2 GB or to the device's available memory. Large file uploads (> 250 MB) may fail for client-side uploads using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component. For more information, see the following discussions:
+
+* [The Blazor InputFile Component should handle chunking when the file is uploaded (dotnet/runtime #84685)](https://github.com/dotnet/runtime/issues/84685)
+* [Request Streaming upload via http handler (dotnet/runtime #36634)](https://github.com/dotnet/runtime/issues/36634)
+
+For large client-side file uploads that fail when attempting to use the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component, we recommend chunking large files with a custom component using multiple [HTTP range requests](https://developer.mozilla.org/docs/Web/HTTP/Range_requests) instead of using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component.
+
+:::moniker-end
+
+:::moniker range=">= aspnetcore-6.0 < aspnetcore-9.0"
 
 Server-side or client-side, there's no file read or upload size limit specifically for the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component. However, client-side Blazor reads the file's bytes into a single JavaScript array buffer when marshalling the data from JavaScript to C#, which is limited to 2 GB or to the device's available memory. Large file uploads (> 250 MB) may fail for client-side uploads using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component. For more information, see the following discussions:
+
+* [The Blazor InputFile Component should handle chunking when the file is uploaded (dotnet/runtime #84685)](https://github.com/dotnet/runtime/issues/84685)
+* [Request Streaming upload via http handler (dotnet/runtime #36634)](https://github.com/dotnet/runtime/issues/36634)
+
+For large client-side file uploads that fail when attempting to use the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component, we recommend:
+
+* Adopting .NET 9 or later for Chromium-based browsers (Google Chrome and Microsoft Edge) with HTTP/2 protocol support, where this problem is internally addressed by Blazor with [Streams API](https://developer.mozilla.org/docs/Web/API/Streams_API).
+* For non-Chromium browsers, Chromium browsers without HTTP/2, or .NET 8 or earlier, chunking large files with a custom component using multiple [HTTP range requests](https://developer.mozilla.org/docs/Web/HTTP/Range_requests) instead of using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component.
 
 :::moniker-end
 
@@ -111,16 +134,13 @@ Server-side or client-side, there's no file read or upload size limit specifical
 
 The maximum supported file size for the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component is 2 GB. Additionally, client-side Blazor reads the file's bytes into a single JavaScript array buffer when marshalling the data from JavaScript to C#, which is limited to 2 GB or to the device's available memory. Large file uploads (> 250 MB) may fail for client-side uploads using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component. For more information, see the following discussions:
 
-:::moniker-end
-
 * [The Blazor InputFile Component should handle chunking when the file is uploaded (dotnet/runtime #84685)](https://github.com/dotnet/runtime/issues/84685)
 * [Request Streaming upload via http handler (dotnet/runtime #36634)](https://github.com/dotnet/runtime/issues/36634)
 
-For large client-side file uploads that fail when attempting to use the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component, we recommend chunking large files with a custom component using multiple [HTTP range requests](https://developer.mozilla.org/docs/Web/HTTP/Range_requests) instead of using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component.
+* Adopting .NET 9 or later for Chromium-based browsers (Google Chrome and Microsoft Edge) with HTTP/2 protocol support, where this problem is internally addressed by Blazor with [Streams API](https://developer.mozilla.org/docs/Web/API/Streams_API).
+* For non-Chromium browsers, Chromium browsers without HTTP/2, or .NET 5 or earlier, chunking large files with a custom component using multiple [HTTP range requests](https://developer.mozilla.org/docs/Web/HTTP/Range_requests) instead of using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component.
 
-<!-- UPDATE 9.0 PU PR: https://github.com/dotnet/runtime/pull/91295 -->
-
-Work is currently scheduled for .NET 9 (late 2024) to address the client-side file size upload limitation.
+:::moniker-end
 
 ## Examples
 
--- a/aspnetcore/release-notes/aspnetcore-9/includes/blazor.md
+++ b/aspnetcore/release-notes/aspnetcore-9/includes/blazor.md
@@ -235,3 +235,9 @@ The <xref:Microsoft.AspNetCore.Components.Forms.InputNumber%601> component now s
 Support for multiple Blazor Web Apps per server project will be considered for .NET 10 (November, 2025).
 
 For more information, see [Support for multiple Blazor Web apps per server project (`dotnet/aspnetcore` #52216)](https://github.com/dotnet/aspnetcore/issues/52216).
+
+### File size read and upload limits
+
+For Chromium-based browsers (Google Chrome and Microsoft Edge) using the HTTP/2 protocol, client-side Blazor uses [Streams API](https://developer.mozilla.org/docs/Web/API/Streams_API) to permit uploading large files (> 2 GB or larger than the device's available memory).
+
+For more information, see <xref:blazor/file-uploads?view=aspnetcore-9.0#file-size-read-and-upload-limits>.


guardrex:
@pavelsavara ... Do you know if https://github.com/dotnet/runtime/pull/91699 has made it into the latest SDK that I think I can get, which is `9.0.100-rc.2.24474.4`?

I'm still hitting the out of memory error with large file uploads using our published example code under the public RC1 release (`9.0.0-rc.1.24431.7`), which I think will work (in Edge) if I get the SDK that has https://github.com/dotnet/runtime/pull/91699 in it.

pavelsavara:
It should be there. Which sample code ?

guardrex:
This ...

https://learn.microsoft.com/en-us/aspnet/core/blazor/file-uploads?view=aspnetcore-8.0#upload-files-to-a-server-with-client-side-rendering-csr

There are two spots that we'll need coverage ...

* In the *File Uploads* article, which is on the PR in draft form. The question is if request streaming will ✨ ***Just Work!***&trade; ✨  without code changes to the `FileUpload2` component ***IF*** they're using a Chromium browser? If request streaming isn't available (Safari/FF), doesn't request streaming gracefully fall back? ... noting that we never maintained content on Javier's suggestion to developers looking for client-side large file uploads to create a custom HTTP Ranges-based file upload component. We say that's what to do, but we never showed an example of it. I wonder if we need to show how to use HTTP Ranges in case request streaming isn't available?
* We also need coverage in the *Call web API* article, which I'm working on now locally (i.e., it isn't on the PR yet). There, I'll be adding a section and describing the `SetBrowserRequestStreamingEnabled` extension method for `HttpRequestMessage`.

guardrex:
If the `FileUpload2` component ***IS*** going to change, then I'll need the updated component to place in versioned content. Can you (or someone) send me the new component to replace it for >=9.0?

I also have a side question WRT the `LazyBrowserFileStream` part that @MackinnonBuck provided ... Mackinnon ... is that to be versioned OUT 🔪 at >=9.0? IIRC, it was a temporary workaround that can go away at 9.0.

pavelsavara:
Something like below will enable that for Chrome/Edge.

```
var webAssemblyEnableStreamingRequestKey = new HttpRequestOptionsKey<bool>("WebAssemblyEnableStreamingRequest");
var req = new HttpRequestMessage(HttpMethod.Post, url);
req.Options.Set(webAssemblyEnableStreamingRequestKey, true);
```

guardrex:
So, you're saying that we'll have to move away from an `HttpClient.PostAsync(string, HttpContent)` to a `HttpClient.SendAsync(HttpRequestMessage)` ... that's going to be a firm requirement for it to work?

The reason that I thought the code wouldn't need to change is that I thought the inner workings of `HttpClient.PostAsync` would perform the checks for HTTP/2 and `IsBrowser` and automatically set the streaming request key ... that it would do internal magic ✨ to make streaming requests work. I guess you're saying that's not the case.

pavelsavara:
My point was to use the `WebAssemblyEnableStreamingRequest` option, not how to set it.

guardrex:
Latest code is ...

```csharp
var req = new HttpRequestMessage(HttpMethod.Post, "/Filesave");
req.SetBrowserRequestStreamingEnabled(true);
req.Version = HttpVersion.Version20;
req.Content = content;
var response = await Http.SendAsync(req);
```

Works for small files, but for a 1 GB file, I'm getting ...

```
"type": "https://tools.ietf.org/html/rfc9110#section-15.5.1",
"title": "One or more validation errors occurred.",
"status": 400,
"errors": {
  "": [
    "Failed to read the request form. Request body too large. The max request body size is 30000000 bytes."
  ]
```

pavelsavara:
>     "Failed to read the request form. Request body too large. The max request body size is 30000000 bytes."

That's actually server side error. Try setting [MaxRequestBodySize](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.server.kestrel.core.kestrelserverlimits.maxrequestbodysize?view=aspnetcore-8.0) on your server

pavelsavara:
`req.Version = HttpVersion.Version20;` is not necessary, doesn't do anything.

guardrex:
I hit another limit on the server ...

> "Failed to read the request form. Multipart body length limit 134217728 exceeded."

I'll see about extending that one.

guardrex:
# ***Lick'd IT!*** 🍻🕺💃🎉 

Ok ... I'll get this all into the PR. I'll ping for reviews Wednesday when it's all in place. 

image

guardrex:
@danroth27 @mkArtakMSFT @pavelsavara @maraf ... See the ***NOTES*** that I just added to the opening comment ☝️.

aspnetcore/blazor/call-web-api.md
```diff
@@ -236,6 +236,37 @@ You can address this by flowing prerendered state using the Persistent Component
 
 :::moniker-end
 
+:::moniker range=">= aspnetcore-9.0"
+
+## Client-side request streaming
+
+For Chromium-based browsers (for example, Google Chrome and Microsoft Edge) using the HTTP/2 protocol, HTTPS, and [CORS](xref:security/cors), client-side Blazor uses [Streams API](https://developer.mozilla.org/docs/Web/API/Streams_API) to permit [request streaming](https://developer.chrome.com/docs/capabilities/web-apis/fetch-streaming-requests).
```
danroth27, line 243: This makes it sound like CORS is always required for request streaming, even for same origin requests. Is this really correct? I suspect that CORS is only required for cross origin requests, just like with any other fetch request.

aspnetcore/blazor/call-web-api.md
```diff
@@ -236,6 +236,37 @@ You can address this by flowing prerendered state using the Persistent Component
...
+request.SetBrowserRequestStreamingEnabled(true);
+request.Content = content;
+
+var response = await Http.SendAsync(request);
+```
+
+Streaming requests:
+
+* Require HTTPS protocol and don't work on HTTP/1.x.
+* Include a body but not a `Content-Length` header. [CORS](xref:security/cors) is required, and a CORS preflight request is always issued.
```
danroth27, line 264: I see the CORS statement here is from https://developer.chrome.com/docs/capabilities/web-apis/fetch-streaming-requests, but again I suspect CORS is only really required for cross origin requests. I mean, does it work if you don't do anything to set up CORS on the server? If yes, then I suggest:

```suggestion
* Include a body but not a `Content-Length` header. [CORS](xref:security/cors) with a preflight reqeust is required for cross-origin streaming requests.
```

aspnetcore/blazor/file-uploads.md
```diff
@@ -101,26 +101,88 @@ If you're using the [Autofac Inversion of Control (IoC) container](https://autof
 
 ## File size read and upload limits
 
-:::moniker range=">= aspnetcore-6.0"
+:::moniker range=">= aspnetcore-9.0"
+
+Server-side or client-side, there's no file read or upload size limit for the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component.
```
danroth27, line 106: This statement sounds like it contradicts the earlier statement in this doc that "Reading one file or multiple files larger than 500 KB results in an exception." Should this statement just be removed?

```suggestion
```

aspnetcore/blazor/file-uploads.md
```diff
@@ -101,26 +101,88 @@ If you're using the [Autofac Inversion of Control (IoC) container](https://autof
...
+
+For Chromium-based browsers (for example, Google Chrome and Microsoft Edge) using the HTTP/2 protocol, HTTPS, and [CORS](xref:security/cors), client-side Blazor uses [Streams API](https://developer.mozilla.org/docs/Web/API/Streams_API) to permit uploading large files (over 250 MB or larger than the device's available memory) with [request streaming](xref:blazor/call-web-api#client-side-request-streaming).
```
danroth27, line 108: Clarify that request streaming isn't done automatically. You have to enable it.

```suggestion
For Chromium-based browsers (for example, Google Chrome and Microsoft Edge) using the HTTP/2 protocol, HTTPS, and [CORS](xref:security/cors), client-side Blazor supports using the [Streams API](https://developer.mozilla.org/docs/Web/API/Streams_API) to permit uploading large files (over 250 MB or larger than the device's available memory) with [request streaming](xref:blazor/call-web-api#client-side-request-streaming).
```

danroth27, line 108: Where does the 250 MB limit come from?

aspnetcore/blazor/file-uploads.md
```diff
@@ -101,26 +101,88 @@ If you're using the [Autofac Inversion of Control (IoC) container](https://autof
...
+
+For non-Chromium browsers or without HTTP/2 protocol/HTTPS/CORS, client-side Blazor reads the file's bytes into a single JavaScript array buffer when marshalling the data from JavaScript to C#, which is limited to 2 GB or to the device's available memory. Large file uploads, typically 250 MB or larger, may fail for client-side uploads using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component. For more information, see the following discussions:
+
+* [The Blazor InputFile Component should handle chunking when the file is uploaded (dotnet/runtime #84685)](https://github.com/dotnet/runtime/issues/84685)
+* [Request Streaming upload via http handler (dotnet/runtime #36634)](https://github.com/dotnet/runtime/issues/36634)
+
```
danroth27, line 114: I'd prefer to not ask users to read through GitHub issues to understand the related concepts. I read through the issues and I think you've got all the important details covered here.

```suggestion
For non-Chromium browsers or without HTTP/2 protocol/HTTPS/CORS, client-side Blazor reads the file's bytes into a single JavaScript array buffer when marshalling the data from JavaScript to C#, which is limited to 2 GB or to the device's available memory. Large file uploads, typically 250 MB or larger, may fail for client-side uploads using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component. 
```

aspnetcore/blazor/file-uploads.md
```diff
@@ -101,26 +101,88 @@ If you're using the [Autofac Inversion of Control (IoC) container](https://autof
...
+For large client-side file uploads that don't meet the preceding criteria for request streaming and fail when attempting to use the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component, we recommend chunking large files with a custom component using multiple [HTTP range requests](https://developer.mozilla.org/docs/Web/HTTP/Range_requests) instead of using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component.
```
danroth27, line 115: @javiercn Can you please confirm that we want to formalize the HTTP range request recommendation? Unless we've proven it's viable with a sample implementation, I'd prefer just to leave it out.

aspnetcore/blazor/call-web-api.md
guardrex:
```suggestion
For Chromium-based browsers (for example, Google Chrome and Microsoft Edge) using the HTTP/2 protocol, and HTTPS, client-side Blazor uses [Streams API](https://developer.mozilla.org/docs/Web/API/Streams_API) to permit [request streaming](https://developer.chrome.com/docs/capabilities/web-apis/fetch-streaming-requests).
```

aspnetcore/blazor/call-web-api.md
guardrex: Yes, @danroth27, that's the source. The PU referred to that Chrome doc.

@pavelsavara @campersau ... WRT Request Streaming Upload, the remarks are ...

> Requires CORS and triggers a preflight
>
> Streaming requests have a body, but don't have a `Content-Length` header. That's a new kind of request, so CORS is required, and these requests always trigger a preflight.
>
> Streaming `no-cors` requests are not allowed.

Cross-ref: https://developer.chrome.com/docs/capabilities/web-apis/fetch-streaming-requests#requires_cors_and_triggers_a_preflight

It's listed under "Restrictions" and makes it sound like a preflight will be issued. I take it now that that's ***incorrect*** for a same-origin request ... *correct?* 😄

aspnetcore/blazor/file-uploads.md
guardrex: It just means that there's no restriction from the component itself ... that it doesn't try to set any such limit. This is stated to make it clear that any limits aren't due to the use of the component. Do you still want to strike it, or perhaps it should be rephrased to something like ...

```suggestion
If any file size upload limits are imposed during a file upload, they aren't imposed by the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component, which never sets an upload limit server-side or client-side.
```

aspnetcore/blazor/call-web-api.md
campersau: This is what is listed in the chrome status feature description:

> - Only 'cors' and 'same-origin' requests allow streaming upload. You can't use streaming upload with 'navigate' and 'no-cors' requests.
> - This feature cannot be used with HTTP/1.x. If the server doesn't support HTTP/2 or HTTP/3, the request fails. This is for some compatibility concerns. See whatwg/fetch#966 for the past discussions.

https://chromestatus.com/feature/5274139738767360

So to me it looks like `no-cors` is not allowed but `cors` and `same-origin` are. I also tested it: same origin requests don't trigger preflight requests.

danroth27: Yup, that all makes sense. The Chrome blog post is referring to the allowed [request mode](https://developer.mozilla.org/en-US/docs/Web/API/Request/mode). But when we talk about CORS being required in ASP.NET Core, that makes it sound like you have to enable CORS on the server, which isn't really the case.

aspnetcore/blazor/file-uploads.md
```diff
@@ -101,26 +101,84 @@ If you're using the [Autofac Inversion of Control (IoC) container](https://autof
...
 ## File size read and upload limits
 
-:::moniker range=">= aspnetcore-6.0"
+:::moniker range=">= aspnetcore-9.0"
+
+Server-side or client-side, there's no file read or upload size limit for the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component.
+
+For Chromium-based browsers (for example, Google Chrome and Microsoft Edge) using the HTTP/2 protocol, HTTPS, and [CORS](xref:security/cors), client-side Blazor supports using the [Streams API](https://developer.mozilla.org/docs/Web/API/Streams_API) to permit uploading large files (over 250 MB or larger than the device's available memory) with [request streaming](xref:blazor/call-web-api#client-side-request-streaming).
+
+For non-Chromium browsers, without HTTP/2 protocol, or without HTTPS, client-side Blazor reads the file's bytes into a single JavaScript array buffer when marshaling the data from JavaScript to C#, which is limited to 2 GB or to the device's available memory. Large file uploads, typically 250 MB or larger, may fail for client-side uploads using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component.
```
guardrex, line 110:
```suggestion
For non-Chromium browsers, without HTTP/2 protocol, or without HTTPS, client-side Blazor reads the file's bytes into a single JavaScript array buffer when marshaling the data from JavaScript to C#, which is limited to 2 GB or to the device's available memory. Large file uploads, typically 250 MB or larger, may fail for client-side uploads using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component.

```

aspnetcore/blazor/file-uploads.md
danroth27: I think we should just remove the statement and focus the content on the limits that are actually relevant.

aspnetcore/blazor/file-uploads.md
```diff
@@ -101,26 +101,83 @@ If you're using the [Autofac Inversion of Control (IoC) container](https://autof
 
 ## File size read and upload limits
 
-:::moniker range=">= aspnetcore-6.0"
+:::moniker range=">= aspnetcore-9.0"
+
+For Chromium-based browsers (for example, Google Chrome and Microsoft Edge) using the HTTP/2 protocol, HTTPS, and [CORS](xref:security/cors), client-side Blazor supports using the [Streams API](https://developer.mozilla.org/docs/Web/API/Streams_API) to permit uploading large files (over 250 MB or larger than the device's available memory) with [request streaming](xref:blazor/call-web-api#client-side-request-streaming).
```
danroth27, line 106: Let's remove this 250MB statement from this doc as it seems to be just an arbitrary estimate that @pavelsavara made in https://github.com/dotnet/AspNetCore.Docs/pull/28306#discussion_r1097260587. Instead, let's stick with generic language about "large files".

```suggestion
For Chromium-based browsers (for example, Google Chrome and Microsoft Edge) using the HTTP/2 protocol, HTTPS, and [CORS](xref:security/cors), client-side Blazor supports using the [Streams API](https://developer.mozilla.org/docs/Web/API/Streams_API) to permit uploading large files with [request streaming](xref:blazor/call-web-api#client-side-request-streaming).
```

aspnetcore/blazor/file-uploads.md
```diff
@@ -101,26 +101,83 @@ If you're using the [Autofac Inversion of Control (IoC) container](https://autof
...
+
+For large client-side file uploads that don't meet the preceding criteria for request streaming and fail when attempting to use the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component, we recommend chunking large files with a custom component using multiple [HTTP range requests](https://developer.mozilla.org/docs/Web/HTTP/Range_requests) instead of using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component.
+
+:::moniker-end
+
+:::moniker range=">= aspnetcore-6.0 < aspnetcore-9.0"
 
 Server-side or client-side, there's no file read or upload size limit specifically for the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component. However, client-side Blazor reads the file's bytes into a single JavaScript array buffer when marshalling the data from JavaScript to C#, which is limited to 2 GB or to the device's available memory. Large file uploads (> 250 MB) may fail for client-side uploads using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component. For more information, see the following discussions:
```
danroth27, line 116:
```suggestion
Server-side or client-side, there's no file read or upload size limit specifically for the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component. However, client-side Blazor reads the file's bytes into a single JavaScript array buffer when marshalling the data from JavaScript to C#, which is limited to 2 GB or to the device's available memory. Large file uploads may fail for client-side uploads using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component. For more information, see the following discussions:
```

aspnetcore/blazor/file-uploads.md
```diff
@@ -101,26 +101,83 @@ If you're using the [Autofac Inversion of Control (IoC) container](https://autof
...
+* For non-Chromium browsers, Chromium browsers without HTTP/2, or .NET 8 or earlier, chunking large files with a custom component using multiple [HTTP range requests](https://developer.mozilla.org/docs/Web/HTTP/Range_requests) instead of using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component.
 
 :::moniker-end
 
+:::moniker range="< aspnetcore-6.0"
+
+The maximum supported file size for the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component is 2 GB. Additionally, client-side Blazor reads the file's bytes into a single JavaScript array buffer when marshalling the data from JavaScript to C#, which is limited to 2 GB or to the device's available memory. Large file uploads (> 250 MB) may fail for client-side uploads using the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component. For more information, see the following discussions:
+
 * [The Blazor InputFile Component should handle chunking when the file is uploaded (dotnet/runtime #84685)](https://github.com/dotnet/runtime/issues/84685)
 * [Request Streaming upload via http handler (dotnet/runtime #36634)](https://github.com/dotnet/runtime/issues/36634)
```
danroth27, line 133:
```suggestion
The maximum supported file size for the <xref:Microsoft.AspNetCore.Components.Forms.InputFile> component is 2 GB. Additionally, client-side Blazor reads the file's bytes into a single JavaScript array buffer when marshalling the data from JavaScript to C#, which is limited to 2 GB or to the device's available memory.
```

aspnetcore/blazor/file-uploads.md
```diff
@@ -101,26 +101,83 @@ If you're using the [Autofac Inversion of Control (IoC) container](https://autof
...
-Work is currently scheduled for .NET 9 (late 2024) to address the client-side file size upload limitation.
+## Security considerations
+
+### Avoid `IBrowserFile.Size` for file size limits
+
+Avoid using <xref:Microsoft.AspNetCore.Components.Forms.IBrowserFile.Size?displayProperty=nameWithType> to impose a limit on the file size.
+
+<span aria-hidden="true">❌</span> The following approach is ***insecure*** and must be avoided:
+
+```diff
```
danroth27, line 148: Is this use of `diff` here intentional? It doesn't appear to be an actual diff. The above examples that show anti-patterns use normal `csharp` formatting.

aspnetcore/blazor/file-uploads.md
```diff
@@ -480,6 +537,12 @@ The server app must register controller services and map controller endpoints. F
 
 The following example demonstrates uploading files to a backend web API controller in a separate app, possibly on a separate server, from a component in a Blazor Web App that adopts CSR or a component in a Blazor WebAssembly app.
 
+:::moniker range=">= aspnetcore-9.0"
+
+The example adopts [request streaming](xref:blazor/call-web-api#client-side-request-streaming) for a Chromium-based browser (for example, Google Chrome or Microsoft Edge) with HTTP/2 protocol, [CORS](xref:security/cors), and HTTPS. If request streaming can't be used, Blazor gracefully degrades to [Fetch API](https://developer.mozilla.org/docs/Web/API/Fetch_API) without request streaming. For more information, see the [File size read and upload limits](#file-size-read-and-upload-limits) section.
```
danroth27, line 542:
```suggestion
The example adopts [request streaming](xref:blazor/call-web-api#client-side-request-streaming) for a Chromium-based browser (for example, Google Chrome or Microsoft Edge) with HTTP/2 protocol and HTTPS. If request streaming can't be used, Blazor gracefully degrades to [Fetch API](https://developer.mozilla.org/docs/Web/API/Fetch_API) without request streaming. For more information, see the [File size read and upload limits](#file-size-read-and-upload-limits) section.
```

aspnetcore/blazor/file-uploads.md
guardrex: It was intentional to make it clear that this isn't the approach to take. It doesn't have to be a diff just to get the formatting.

I'll change this so that it's combined with the next code example and is an actual diff.

aspnetcore/blazor/file-uploads.md
```diff
@@ -336,7 +379,7 @@ The following `FileUpload2` component:
 
 :::moniker range=">= aspnetcore-8.0"
 
-If the component limits file uploads to a single file at a time or if the component only adopts interactive client-side rendering (CSR, `InteractiveWebAssembly`), the component can avoid the use of the `LazyBrowserFileStream` and use a <xref:System.IO.Stream>. The following demonstrates the changes for the `FileUpload2` component:
+If the component limits file uploads to a single file at a time or if the component only adopts client-side rendering (CSR, `InteractiveWebAssembly`), the component can avoid the use of the `LazyBrowserFileStream` and use a <xref:System.IO.Stream>. The following demonstrates the changes for the `FileUpload2` component:
```
danroth27, line 382: @guardrex It's not really clear to me what problem `LazyBrowserFileStream` is trying to solve. Why is it needed?

danroth27: @lewing @javiercn It looks like `LazyBrowserFileStream` was introduced per @MackinnonBuck's [recommendation](https://github.com/dotnet/AspNetCore.Docs/pull/33357#pullrequestreview-2240989426) to work around this framework bug: https://github.com/dotnet/aspnetcore/issues/47301. Perhaps we should prioritize getting https://github.com/dotnet/aspnetcore/issues/47301 fixed for .NET 10 so that this workaround is no longer needed?

guardrex: I'll make a 10.0 tracking entry for https://github.com/dotnet/aspnetcore/issues/47301 to keep it on my radar.

aspnetcore/blazor/file-uploads.md
```diff
@@ -687,15 +901,72 @@ public class FilesaveController(
...
 
-<!--
+:::moniker range=">= aspnetcore-9.0"
 
-HOLD: Tracking anti-request forgery work for this article in the UE tracking issue.
+Cross-Origin Requests (CORS) configuration on the server is required for [request streaming](https://developer.chrome.com/docs/capabilities/web-apis/fetch-streaming-requests), and a preflight request is always made by the client. In the service configuration of the server's `Program` file (the server project of a Blazor Web App or the backend server web API of a Blazor WebAssembly app), the following default CORS policy is suitable for testing with the examples in this article. The client makes the local request from port 5003. Change the port number to match the client app port that you're using:
```
danroth27, line 908:
```suggestion
Cross-Origin Requests (CORS) configuration on the server is required for [request streaming](https://developer.chrome.com/docs/capabilities/web-apis/fetch-streaming-requests) when the server is hosted at a different origin, and a preflight request is always made by the client. In the service configuration of the server's `Program` file (the server project of a Blazor Web App or the backend server web API of a Blazor WebAssembly app), the following default CORS policy is suitable for testing with the examples in this article. The client makes the local request from port 5003. Change the port number to match the client app port that you're using:
```

danroth27 (approved):
I'm going make some executive decisions to get this PR unblocked:
- For non-Chromium browsers let's go with option 1, which is to do nothing outside of just letting readers know that this feature is only supported for Chromium browsers at this time.
- Let's drop the recommendation to use HTTP range requests as a fallback behavior given that we haven't actually proven that approach with a sample.
- Let's not bother with trying to document how to deal specifically with the auto render mode.

@lewing @pavelsavara @javiercn If you disagree with any of these decisions, then please open a separate doc issue with how you'd like them handled instead.

guardrex:
@danroth27 ... I'll react to your decisions in the morning.

@lewing @pavelsavara @javiercn ... If you need to open an issue, please open it from the bottom of this article ...

https://learn.microsoft.com/en-us/aspnet/core/blazor/file-uploads?view=aspnetcore-9.0#additional-resources

guardrex:
@danroth27 ... Based on your remarks earlier about not cross-linking the PU discussions on this, I simplified the *File size read and upload limits* section down to just ...

> :::moniker range=">= aspnetcore-9.0"
>
> For Chromium-based browsers (for example, Google Chrome and Microsoft Edge) using the HTTP/2 protocol, HTTPS, and \[CORS](xref:security/cors), client-side Blazor supports using the \[Streams API](https://developer.mozilla.org/docs/Web/API/Streams_API) to permit uploading large files with \[request streaming](xref:blazor/call-web-api#client-side-request-streaming).
>
> Without a Chromium browser, HTTP/2 protocol, or HTTPS, client-side Blazor reads the file's bytes into a single JavaScript array buffer when marshaling the data from JavaScript to C#, which is limited to 2 GB or to the device's available memory. Large file uploads may fail for client-side uploads using the \<xref:Microsoft.AspNetCore.Components.Forms.InputFile> component.
>
> :::moniker-end
>
> :::moniker range="< aspnetcore-9.0"
>
> Client-side Blazor reads the file's bytes into a single JavaScript array buffer when marshaling the data from JavaScript to C#, which is limited to 2 GB or to the device's available memory. Large file uploads may fail for client-side uploads using the \<xref:Microsoft.AspNetCore.Components.Forms.InputFile> component. We recommend adopting [request streaming](xref:blazor/call-web-api?view=aspnetcore-8.0&preserve-view=true#client-side-request-streaming) with ASP.NET Core 9.0 or later.
>
> :::moniker-end

Otherwise, there are no changes for the other two items ...

* 'do nothing outside of just letting readers know that this feature is only supported for Chromium browsers at this time'
* 'trying to document how to deal specifically with the auto render mode'

aspnetcore/blazor/file-uploads.md
guardrex: References to HTTP range requests ... ***Removed!*** 🔪 on the latest updates.

danroth27: I spoke to @javiercn about this and he is fine with this removal.

danroth27:
:shipit:

PR #33802, opened 2024-10-06
Clarify best app base path (`<base>` tag) location

### LINKED ISSUE #33797: Blazor host and deploy: Configure the app base path should mention that app base path only applies to the following paths
### Description

While following the steps to configure the app base path, I noticed that it won't work after I added the tag to my index.html document. It took me a while to figure out that the base element needs to be added before every other element that uses relative sources.

https://www.seoclarity.net/resources/knowledgebase/glossary/base-tags

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/?view=aspnetcore-8.0&tabs=visual-studio#configure-the-app-base-path

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/host-and-deploy/index.md

### Document ID

c82cf121-831d-75dd-a796-c735aad3b3e9

### Article author

@guardrex

guardrex:
Fixes #33797

Thanks @tobbi! 🚀 ... I'll let this sit until Monday morning to give you a chance to look/comment on it.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/host-and-deploy/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/f4549e546f5aa4e6d0a2d5c7430a68e27a647578/aspnetcore/blazor/host-and-deploy/index.md) | [aspnetcore/blazor/host-and-deploy/index](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/index?branch=pr-en-us-33802) |

### DIFF
--- a/aspnetcore/blazor/host-and-deploy/index.md
+++ b/aspnetcore/blazor/host-and-deploy/index.md
@@ -198,7 +198,7 @@ To provide configuration for the Blazor app's base path of `https://www.contoso.
 
 By configuring the app base path, a component that isn't in the root directory can construct URLs relative to the app's root path. Components at different levels of the directory structure can build links to other resources at locations throughout the app. The app base path is also used to intercept selected hyperlinks where the `href` target of the link is within the app base path URI space. The <xref:Microsoft.AspNetCore.Components.Routing.Router> component handles the internal navigation.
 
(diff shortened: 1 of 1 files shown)


tobbi:
This works, I guess. I would've maybe phrased it a bit differently, because declaring the base path **before** any asset paths (js files, css files, etc.) is mandatory. It doesn't need to be at the start of the head tag, just before any relative paths are referenced. The way you wrote it makes it sound like it's optional.

Could be that it's just me. I'd like others to chime in as well.

PR #33818, opened 2024-10-08
[Blazor] prerender.md - no data protection on prerendered state

hakenr:
There’s no data protection on the prerendered state sent to the WASM client. The WASM client doesn’t have any keys to decrypt protected data.

You can use Data Protection on your own (and there may be scenarios where built-in features do this) if there are specific data pieces you don’t want the client to use directly but rather send back to the server unchanged (where it can be decrypted and processed).

I did a quick test. I took the value from the `<!--Blazor-WebAssembly-Component-State:{VALUE}-->` comment on the prerendered page, and it's just plain Base64-encoded JSON that I can decode and read.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/prerender.md](https://github.com/dotnet/AspNetCore.Docs/blob/78509bd1f1cbfba8b6f854839dfb87ff0bae24a3/aspnetcore/blazor/components/prerender.md) | [Prerender ASP.NET Core Razor components](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/prerender?branch=pr-en-us-33818) |
| [aspnetcore/blazor/components/prerendering-and-integration.md](https://github.com/dotnet/AspNetCore.Docs/blob/78509bd1f1cbfba8b6f854839dfb87ff0bae24a3/aspnetcore/blazor/components/prerendering-and-integration.md) | [Prerender and integrate ASP.NET Core Razor components](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/prerendering-and-integration?branch=pr-en-us-33818) |

### DIFF
--- a/aspnetcore/blazor/components/prerender.md
+++ b/aspnetcore/blazor/components/prerender.md
@@ -114,8 +114,6 @@ When the component executes, `currentCount` is only set once during prerendering
 
 By initializing components with the same state used during prerendering, any expensive initialization steps are only executed once. The rendered UI also matches the prerendered UI, so no flicker occurs in the browser.
 
-The prerendered data sent to the client by the <xref:Microsoft.AspNetCore.Components.PersistentComponentState> service is protected by [ASP.NET Core Data Protection](xref:security/data-protection/introduction), so the state can't be read by a malicious user.
-
 ## Components embedded into pages and views (Razor Pages/MVC)
 
 For components embedded into a page or view of a Razor Pages or MVC app, you must add the [Persist Component State Tag Helper](xref:mvc/views/tag-helpers/builtin-th/persist-component-state-tag-helper) with the `<persist-component-state />` HTML tag inside the closing `</body>` tag of the app's layout. **This is only required for Razor Pages and MVC apps.** For more information, see <xref:mvc/views/tag-helpers/builtin-th/persist-component-state-tag-helper>.
--- a/aspnetcore/blazor/components/prerendering-and-integration.md
+++ b/aspnetcore/blazor/components/prerendering-and-integration.md
@@ -920,8 +920,6 @@ else
 
 By initializing components with the same state used during prerendering, any expensive initialization steps are only executed once. The rendered UI also matches the prerendered UI, so no flicker occurs in the browser.
 
-The prerendered data sent to the client by the <xref:Microsoft.AspNetCore.Components.PersistentComponentState> service is protected by [ASP.NET Core Data Protection](xref:security/data-protection/introduction), so the state can't be read by a malicious user.
-
 :::zone pivot="webassembly"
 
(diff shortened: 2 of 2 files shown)


guardrex:
@javiercn ... This came from your response to a dev at https://github.com/dotnet/aspnetcore/issues/47669#issuecomment-1506909381. Did I misinterpret your remarks?

hakenr:
@guardrex It seems @javiercn meant Blazor Server, where both pre-rendering and interactive rendering happen on the server, and in such a scenario, it's possible that Data Protection is used. It's strange that it would be different for Blazor Server and Blazor WASM, but it's possible - I'll look into it.

I'll add the second occurrence to the PR.

hakenr:
@guardrex I think I found it:  
https://github.com/dotnet/aspnetcore/blob/ed4a75d098f10925e8bac0761524fba920a08e6b/src/Components/Endpoints/src/Rendering/EndpointHtmlRenderer.PrerenderingState.cs#L138-L154  

There are two different serialization modes: one for Server and one for WebAssembly. The Server mode is "protected," while the WASM mode is not. We might consider updating the documentation to cover this difference. However, I'm not sure what the specification says about this - whether it guarantees protection for Server mode or if it's just the current implementation, which could change in the future. @javiercn?

(+ there might be additional scenarios such as MAUI)

hakenr:
(@guardrex If we can't assemble the new paragraph version right now, I'd prefer to merge this PR to live immediately to avoid having security-sensitive misinformation in the documentation.)

guardrex:
> I'll add the second occurrence to the PR.

So ... it seems like you're suggesting that we leave the remark in the <8.0 article because for Blazor Server it is a correct remark. Is that your conclusion?

hakenr:
> So ... it seems like you're suggesting that we leave the remark in the <8.0 article because for Blazor Server it is a correct remark. Is that your conclusion?

Uh, no, sorry, I just forgot to push the commit from the temporary Codespace I used.

guardrex:
I sent Javier an email to ask him to look at this and the try-catch removal PR.

> If we can't assemble the new paragraph version right now

I think we can update this quickly 🏃. It would be nice to leave the original issue closed and not generate a new PR. Let's see if he's around this morning. If he doesn't respond within a couple of hours, I'll merge this and re-open the issue.

guardrex:
Let's go ahead. I'll re-open the original issue and discuss this with the team later.

guardrex (approved):

javiercn:
It's data protected only for server payloads. There's no point in using data protection on the client.

The reason it's protected for server payloads is because those are read and interpreted on the server as opposed to the client, which is considered untrusted from the server point of view.

guardrex:
Thanks ... I'll try again with a new PR.

PR #33900, opened 2024-10-22
remove erroneous text

alexravenna:


guardrex (approved):
Thanks @alexravenna! 🚀 ... Yes, indeed! An edit went sideways there! 😈

PR #33921, opened 2024-10-25
Update openAPI_completion.md

Burgyn:
This PR addresses two documentation inaccuracies:

Updated usage of the outdated method `UseSchemaTransformer`, which has been replaced by `AddSchemaTransformer` in recent versions of .NET.

Corrected type checking syntax from `if (c.Type == typeof(Product))` to `if (c.JsonTypeInfo.Type == typeof(Product))`, aligning with the current API structure for accessing type information.


guardrex:
`UseSchemaTransformer` is in another file, too, which may require an update ...

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/release-notes/aspnetcore-9/includes/transformer-registration.md

PR #33995, opened 2024-11-01
Add details on enums in OpenAPI section

mikekistler:
This PR adds just a bit more detail about enum metadata to the OpenAPI section of the docs. Specifically, it adds an example of the JsonConverter attribute and explains how flag enums are handled (and why).

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/fundamentals/openapi/include-metadata.md](https://github.com/dotnet/AspNetCore.Docs/blob/cb97e6a3ede43bfe949f898589554bae2eec5762/aspnetcore/fundamentals/openapi/include-metadata.md) | [aspnetcore/fundamentals/openapi/include-metadata](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/openapi/include-metadata?branch=pr-en-us-33995) |

### DIFF
--- a/aspnetcore/fundamentals/openapi/include-metadata.md
+++ b/aspnetcore/fundamentals/openapi/include-metadata.md
@@ -521,7 +521,14 @@ public enum DayOfTheWeekAsString
 }
 ```
 
-A special case is when an enum type has the [Flags] attribute, which indicates that the enum can be treated as a bit field; that is, a set of flags. A flags enums with a [JsonConverterAttribute] will be defined as `type: string` in the generated schema with no `enum` property, since the value could be any combination of the enum values.
+A special case is when an enum type has the [Flags] attribute, which indicates that the enum can be treated as a bit field; that is, a set of flags. A flags enum with a [JsonConverterAttribute] will be defined as `type: string` in the generated schema with no `enum` property, since the value could be any combination of the enum values. For example, the following enum:
+
+```csharp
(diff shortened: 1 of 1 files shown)


aspnetcore/fundamentals/openapi/include-metadata.md
```diff
@@ -504,9 +504,28 @@ Properties can also be marked as `required` with the [required](/dotnet/csharp/l
...
+    Monday,
+    Tuesday,
+    Wednesday,
+    Thursday,
+    Friday,
+    Saturday
+}
+```
+
+A special case is when an enum type has the [Flags] attribute, which indicates that the enum can be treated as a bit field; that is, a set of flags. A flags enums with a [JsonConverterAttribute] will be defined as `type: string` in the generated schema with no `enum` property, since the value could be any combination of the enum values.
```
mikekistler, line 524: Fix typo:
```suggestion
A special case is when an enum type has the [Flags] attribute, which indicates that the enum can be treated as a bit field; that is, a set of flags. A flags enum with a [JsonConverterAttribute] will be defined as `type: string` in the generated schema with no `enum` property, since the value could be any combination of the enum values.
```

captainsafia (approved):
One comment about more examples here. LGTM otherwise.

tdykstra (approved):
LGTM

PR #34054, opened 2024-11-07
[Blazor] Server security - Inject AuthenticationStateProvider for ser…

hakenr:
This entire section seems more confusing than helpful:

> Don't attempt to resolve AuthenticationStateProvider within a custom scope because it results in the creation of a new instance of the AuthenticationStateProvider that isn't correctly initialized.

What exactly do we mean by "custom scope" here? The article is focused on server-side Blazor, where the default scope is the circuit. Any additional scope, such as `OwningComponentBase`, is simply a nested scope and will resolve the correct `AuthenticationStateProvider` instance.
 
Yes, we could construct an edge case - like using a `BackgroundService` in the same process, where there is no circuit scope available. But in such scenarios, there's no user identity available for that service workload anyway, and passing an `AuthenticationStateProvider` from a component to a background service method doesn't make sense.

> `ExampleService.cs`
>
> Register the service as scoped. In a server-side Blazor app, scoped services have a lifetime equal to the duration of the client connection [circuit].

Firstly, the service itself is stateless (i.e., it has no instance data fields), which makes the choice of lifetime irrelevant.

Secondly, if both the service and `AuthenticationStateProvider` have lifetimes tied to the client connection circuit, why are we passing `AuthenticationStateProvider` as a method parameter? It seems unnecessary if they share the same scope.

Thirdly, in the example below (`InjectAuthStateProvider.razor`), a component scope is used to resolve `ExampleService`, which contradicts the earlier statements about circuit-scoped services.

Overall, this section introduces too much confusion and should probably be removed. The basic principles of scoped services and `OwningComponentBase` are already well-explained in the _Dependency Injection_ documentation, and they apply just as well to `AuthenticationStateProvider`.

If there's a specific behavior of `AuthenticationStateProvider` that needs to be covered here, we should rewrite the section with clear examples that illustrate this behavior effectively.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/security/server/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/7c17a3b75de3263148823a8bc537f2ecb6dd8912/aspnetcore/blazor/security/server/index.md) | [Secure ASP.NET Core server-side Blazor apps](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/server/index?branch=pr-en-us-34054) |


guardrex:
> This entire section seems more confusing than helpful:

Actually, the entire Blazor *Security and Identity* node is due for a major overhaul. However, it won't be possible this year. It will likely occur in 25Q1.

This section came in from the product unit for publication, so we'll get a technical analysis from them on your remarks and desire to remove it. However, we have the .NET 9 release next week, and I know for sure that they can't look at this right now. I'll mention this to Artak on Friday, and he'll likely place it on his schedule for a look *ASAP*.

guardrex:
Moving to #34134 for work on the section.

PR #34108, opened 2024-11-13
Add a not-latest-version INCLUDE

### LINKED ISSUE #34100: Not latest version INCLUDE breaks on out-of-support moniker line
@Rick-Anderson @tdykstra @wadepickett 

I went to activate the not-latest version INCLUDE in my >=8.0 articles, but the newer out-of-support section of the INCLUDE breaks the file ...

> Line 6: [Warning] No  intersection between zone and file level monikers. The result of zone  level range string '= aspnetcore-7.0 \|\| = aspnetcore-5.0 \|\| =  aspnetcore-3.0 \|\| = aspnetcore-3.1 \|\| = aspnetcore-2.0' is  'aspnetcore-2.0', 'aspnetcore-3.0', 'aspnetcore-3.1', 'aspnetcore-5.0',  'aspnetcore-7.0', while file level monikers is 'aspnetcore-10.0',  'aspnetcore-8.0', 'aspnetcore-9.0'.

That section of the file can't be there until .NET 9 (STS) falls out of support (i.e., a `|| = aspnetcore-9.0` is added to the conditional) for any >=8.0 articles, so either that section of the file has to go, or I'll need to create a Blazor-specific version of the file without that section. Which way do you want me to go to solve it? 👂

guardrex:
Fixes #34100

This version of the file doesn't include the not-supported content.

After this goes in, I'll update my Blazor PR to use this file, which only addresses the new Blazor articles at 8.0 that were versioned >=8.0 at the file level.

### DIFF
--- a/aspnetcore/includes/not-latest-version-without-not-supported-content.md
+++ b/aspnetcore/includes/not-latest-version-without-not-supported-content.md
@@ -11,11 +11,19 @@
 :::moniker-end
 
 <!--
-Include this file at the top of articles. When a new version is released,
-it might be necessary to temporarily comment out the current version
-moniker range section until the new moniker is created.
-Markdown to include this file:
-[!INCLUDE[](~/includes/not-latest-version-without-not-supported-content.md)]
+Include either this file or 'not-latest-version.md' at the top of articles.
 
-This version of the file doesn't include the not-supported-version content.
+'not-latest-version.md': Includes not-supported content.
+'not-latest-version-without-not-supported-content.md' (this file): Doesn't include not-supported content.
+
+Use this file in articles that target >=8.0 until 9.0 reaches EOL, and then update those
+articles to use 'not-latest-version.md'. For articles that target >=7.0, 'not-latest-version.md'
+can be used without creating a zone/file moniker range mismatch error.
+
+When a new version is released, it might be necessary to temporarily comment out the current version
+moniker range section until the new moniker is created.
+
+Markdown to include this file:
+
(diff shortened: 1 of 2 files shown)


aspnetcore/includes/not-latest-version-without-not-supported-content.md
```diff
@@ -0,0 +1,21 @@
...
+:::moniker-end
+
+<!--
+Include this file at the top of articles. When a new version is released,
+it might be necessary to temporarily comment out the current version
+moniker range section until the new moniker is created.
+Markdown to include this file:
+[!INCLUDE[](~/includes/not-latest-version-without-not-supported-content.md)]
+
+This version of the file doesn't include the not-supported-version content.
```
tdykstra, line 20: Consider adding a comment explaining when this version needs to be used instead of the original and then add that comment to the original as well.

tdykstra (approved):
Looks good, I left one suggestion

guardrex:
Ok ... I tried to explain it out. These comments will be updated each release along with the markdown/remarks in the files.

guardrex:
Rick ... Wade ... if you have further updates, go for a follow-up PR. I'm trying to clear my other PR through quickly to get on with things 🏃‍♂️🏃‍♂️🏃‍♂️🏃‍♂️🏃‍♂️🏃‍♂️🏃‍♂️.

wadepickett (commented):
Looks great, thanks!!

PR #34128, opened 2024-11-14
Fix `@bind:after` description

### LINKED ISSUE #34127: Blazor bind:after description is incorrect on the "Blazor Data binding" page.
### Description

The description of how to use `bind:after` contains the following text:

>To execute asynchronous logic after binding, use `@bind:after="{EVENT}"` with a DOM event for the {EVENT} placeholder. An assigned C# method isn't executed until the bound value is assigned synchronously.

I think this contains a (very confusing, IMO) cut-and-paste error from the earlier code about bind:event, which (correctly) reads:

>Bind a property or field on other DOM events by including an @bind:event="{EVENT}" attribute with a DOM event for the {EVENT} placeholder.

But the `@bind:after` attribute does **not** have a value which which is a DOM event.  It has a value that's the name of a method (or delegate) to be called after the event has occurred.   It should probably read something more like like:

>To execute asynchronous logic after binding, use `@bind:after="{DELEGATE}"` .  An assigned C# method [_perhaps this should read "delegate" instead?_] isn't executed until the bound value is assigned synchronously.

This problem was there in the 8.0 docs (maybe earlier) and is still in the 10-preview docs.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/data-binding?view=aspnetcore-9.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/data-binding.md

### Document ID

3f912dfc-5ba5-6c6c-ba93-1ae78508a5cc

### Article author

@guardrex

guardrex:
Fixes #34127

Thanks @willdean! 🚀 ... I'll hold it for sec while you check out the update. I'm not going to wait for the product unit on this one. I agree with your suggestion for the change.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/data-binding.md](https://github.com/dotnet/AspNetCore.Docs/blob/451eb67f2ffc4bd65e46871c647fb2d48346dbbb/aspnetcore/blazor/components/data-binding.md) | [aspnetcore/blazor/components/data-binding](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/data-binding?branch=pr-en-us-34128) |

### DIFF
--- a/aspnetcore/blazor/components/data-binding.md
+++ b/aspnetcore/blazor/components/data-binding.md
@@ -158,7 +158,7 @@ Razor attribute binding is case-sensitive:
 
 :::moniker range=">= aspnetcore-7.0"
 
-To execute asynchronous logic after binding, use `@bind:after="{EVENT}"` with a DOM event for the `{EVENT}` placeholder. An assigned C# method isn't executed until the bound value is assigned synchronously.
+To execute asynchronous logic after binding, use `@bind:after="{DELEGATE}"`, where the `{DELEGATE}` placeholder is a C# delegate (method). An assigned C# delegate isn't executed until the bound value is assigned synchronously.
 
 Using an [event callback parameter (`EventCallback`/`EventCallback<T>`)](xref:blazor/components/event-handling#eventcallback) with `@bind:after` isn't supported. Instead, pass a method that returns an <xref:System.Action> or <xref:System.Threading.Tasks.Task> to `@bind:after`.
 


willdean:
LGTM (the Preview Link isn't visible to outsiders, but it's clear enough in the .md)

Thx

guardrex:
Correct ... that link is for internal review. I asked management several years ago if they would open it up for everyone, but I got crickets 🦗  from them 😆.

This PR will probably go live by EOD.

I'll get to your other issue early next week. That one is a bit more nuanced. I'd like Mackinnon (at least) to chime in on if they prefer mentioning it in passing or actually changing the content/code to more forcefully recommend trapping `JSDisconnectedException`.

PR #34150, opened 2024-11-17
Update part-2.md

kennethmanuel:
Added Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore when using VS Code and .NET CLI on 

https://learn.microsoft.com/en-us/aspnet/core/blazor/tutorials/movie-database-app/part-2?view=aspnetcore-9.0&pivots=vsc

Not adding Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore will result in build error when following the tutorial.


guardrex (approved):
Hummm 🤔 ... Not sure how the package didn't find its way in here, as it is required for the `AddDatabaseDeveloperPageExceptionFilter` extension method, and I did test this last year with VSC. I'm going to add one line to call that out. Thanks for the PR!

guardrex:
@kennethmanuel ... If you agree with the terms, can you sign the CLA. We can't accept the PR without it.

kennethmanuel:
@dotnet-policy-service agree

PR #34202, opened 2024-11-22
[Blazor] Security - interactive-server-side-rendering - script tags

hakenr:
>Script tags aren't allowed and shouldn't be included in the app's component render tree. If a script tag is included in a component's markup, a compile-time error is generated.

I'm not aware of any such compile-time behavior.  
While it's not recommended to include script tags in your interactive components (since the behavior can be tricky), there's no built-in mechanism that prevents you from including a script tag in the render tree.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/security/interactive-server-side-rendering.md](https://github.com/dotnet/AspNetCore.Docs/blob/c62c24e5b1b222e5957031bc1a1d4d24cb61b738/aspnetcore/blazor/security/interactive-server-side-rendering.md) | [aspnetcore/blazor/security/interactive-server-side-rendering](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/interactive-server-side-rendering?branch=pr-en-us-34202) |


guardrex:
This goes back to the inception of the article on #13962, written by Javier and reviewed by both Steve Sanderson and Ryan Nowak when he worked on Blazor. It's at Line 194 of ...

https://github.com/dotnet/AspNetCore.Docs/pull/13962/files#diff-4962dbf450346ed225e83e082be3ddbae62e523fd1980145aa2af347ec62a2eaR194

Stand-by while I get them on here to take a look.

guardrex:
**UPDATE (11/25)**: There might be a delay. We're in the *Thanksgiving* 🦃 holiday season here, so it might take a week or longer to advance this. *Stand-by, and we'll get back to this ASAP!* 🏃‍♂

guardrex:
🐀🍔 *Roasted Rat Burgers* 🐀🍔

> RequestError [HttpError]: Resource not accessible by integration

Oh, well. It was worth a shot!

No worries. I doubt many PRs will come in over the next four days.

guardrex:
Will try again. I had the wrong permissions set on the workflow.

SteveSandersonMS:
See https://github.com/dotnet/razor/pull/8744

There used to be a compile-time error if you had a `<script>` inside .razor due to the tricky bits that @hakenr mentions, and because at that time we didn't think there were valid scenarios for people doing it.

However from .NET 8 onwards when we added full SSR support, we judged that there were important scenarios for using `<script>` with SSR so the balance tipped towards allowing it without any warnings or errors.

PR #34214, opened 2024-11-26
QuickGrid empty rows and border styling

### LINKED ISSUE #34211: Document that QuickGrid will render additional empty rows and cells to fill in the final page of data
### Description

QuickGrid will render additional empty rows to fill in the final page of data when used with a Paginator. In .NET 9, it will also fill in empty data cells. This is intended to facilitate styling the QuickGrid so that it has a stable height and styling across all pages. Something like this:

```css
::deep tr {
    height: 2em;
}
```

Note that for `::deep` to work when using component specific styles for whatever component is the parent of the QuickGrid, the QuickGrid needs to have a wrapper element.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/quickgrid?view=aspnetcore-9.0&tabs=visual-studio

### Content source URL

### Document ID

### Article author

@guardrex

guardrex:
Fixes #34211

This includes the border fix, if you wish to keep it. I'll track the empty data row template PU issue, possibly for .NET 10, which will lead to >=10 modifications to that part of the coverage here.

Do you want the movie dB tutorial to include the border workaround for empty rows?

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/quickgrid.md](https://github.com/dotnet/AspNetCore.Docs/blob/7232da59853a689f3bdecceb621e8a6f7b2821bc/aspnetcore/blazor/components/quickgrid.md) | [aspnetcore/blazor/components/quickgrid](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/quickgrid?branch=pr-en-us-34214) |

### DIFF
--- a/aspnetcore/blazor/components/quickgrid.md
+++ b/aspnetcore/blazor/components/quickgrid.md
@@ -138,6 +138,63 @@ To provide a UI for pagination, add a [`Paginator` component](xref:Microsoft.Asp
 
 In the running app, page through the items using a rendered `Paginator` component.
 
+QuickGrid renders additional empty rows to fill in the final page of data when used with a `Paginator` component. In .NET 9 or later, empty data cells (`<td></td>`) are added to the empty rows. The empty rows are intended to facilitate rendering the QuickGrid with stable row height and styling across all pages. You can apply styles to the rows using [CSS isolation](xref:blazor/components/css-isolation) by wrapping the `QuickGrid` component in a wrapper element, such as a `<div>`, and applying a row style with `::deep` [pseudo-elements](https://developer.mozilla.org/docs/Web/CSS/Pseudo-elements):
+
+```css
+::deep tr {
+    height: 2em;
+}
+```
+
+:::moniker range=">= aspnetcore-9.0"
+
+<!-- UPDATE 10.0 Check on https://github.com/dotnet/aspnetcore/issues/59078 to see
+                 if an empty row template feature is added that will result in an 
+                 update the following content for >=10.0. -->
+
+In .NET 9 or later, Bootstrap styling adds a bottom border to empty row data cells, which results in a UI artifact. To remove the border styling of the empty cells on these rows, use [CSS isolation](xref:blazor/components/css-isolation).
+
+Consider the following QuickGrid placed in the app's `Index` component. The `QuickGrid` component has a wrapper `<div>` element, which permits you to apply `::deep` [pseudo-elements](https://developer.mozilla.org/docs/Web/CSS/Pseudo-elements) to the QuickGrid.
+
+`Index.razor`:
+
+```razor
+<div>
+    <QuickGrid ... Pagination="pagination">
+        ...
+    </QuickGrid>
+</div>
+
+<Paginator State="pagination" />
+
+@code {
+    private PaginationState pagination = new PaginationState { ItemsPerPage = 10 };
+
+    ...
+}
+```
+
+In the following isolated CSS styles:
+
+* Row cells populated with data are displayed.
+* Empty row data cells aren't displayed, which avoids empty row cell borders rendering per the Bootstrap style.
+
+`Index.razor.css`:
+
+```css
+::deep tr:has(> td:not(:empty)) > td {
+    display: table-cell;
+}
+
+::deep td:empty {
+    display: none;
+}
+```
+
+:::moniker-end
+
+For more information on using `::deep` [pseudo-elements](https://developer.mozilla.org/docs/Web/CSS/Pseudo-elements) with CSS isolation, see <xref:blazor/components/css-isolation#child-component-support>.
+
 ## Custom attributes and styles
 
 QuickGrid also supports passing custom attributes and style classes (<xref:Microsoft.AspNetCore.Components.QuickGrid.QuickGrid%601.Class%2A>) to the rendered table element:
--- a/aspnetcore/blazor/components/quickgrid.md
+++ b/aspnetcore/blazor/components/quickgrid.md
@@ -152,27 +152,7 @@ QuickGrid renders additional empty rows to fill in the final page of data when u
                  if an empty row template feature is added that will result in an 
                  update the following content for >=10.0. -->
 
-In .NET 9 or later, Bootstrap styling adds a bottom border to empty row data cells, which results in a UI artifact. To remove the border styling of the empty cells on these rows, use [CSS isolation](xref:blazor/components/css-isolation).
-
-Consider the following QuickGrid placed in the app's `Index` component. The `QuickGrid` component has a wrapper `<div>` element, which permits you to apply `::deep` [pseudo-elements](https://developer.mozilla.org/docs/Web/CSS/Pseudo-elements) to the QuickGrid.
-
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/components/quickgrid.md
```diff
@@ -138,6 +138,63 @@ To provide a UI for pagination, add a [`Paginator` component](xref:Microsoft.Asp
...
+
+```css
+::deep tr {
+    height: 2em;
+}
+```
+
+:::moniker range=">= aspnetcore-9.0"
+
+<!-- UPDATE 10.0 Check on https://github.com/dotnet/aspnetcore/issues/59078 to see
```
danroth27, line 151: I believe this feature request is actually about having a template to render when no records are bound, so the QuickGrid is empty of any data.

aspnetcore/blazor/components/quickgrid.md
```diff
@@ -138,6 +138,63 @@ To provide a UI for pagination, add a [`Paginator` component](xref:Microsoft.Asp
...
+
+<Paginator State="pagination" />
+
+@code {
+    private PaginationState pagination = new PaginationState { ItemsPerPage = 10 };
+
+    ...
+}
+```
+
```
danroth27, line 176: I think this can be simplified:

````suggestion
If you prefer to hide the empty rows rendered by the `QuickGrid`, you may do so using CSS styling to hide the empty data cell in rows that that don't contain any data.

````

File: aspnetcore/blazor/components/quickgrid.md (hunk 1 above)
Comment (guardrex) [reply aspnetcore/blazor/components/quickgrid.md:151]: Check out his opening paragraph ... I think the empty record row template styling would allow styling to prevent these rendered rows from showing the borders.

https://github.com/dotnet/aspnetcore/issues/59078

**Review by danroth27 — APPROVED** [review]
I have one suggestion to simplify the content, but otherwise looks good to me.

File: aspnetcore/blazor/components/quickgrid.md (hunk 2 above)
Comment (guardrex) [reply aspnetcore/blazor/components/quickgrid.md:176]: I think it overly simplifies things for new devs and new-to-Blazor devs, but ok.

**Comment by danroth27** [comment]
> Do you want the movie dB tutorial to include the border workaround for empty rows?

I think we would use the fixed row height approach in the tutorial, so that all the pages of data have the same height, including the last one.

**Comment by guardrex** [comment]
But your fixed row height example only made all of the rows a fixed height, including the empty rows.

It ends up looking like this ...

image

**Comment by danroth27** [comment]
Right, the idea is that the grid maintains the same size and styling across all pages. Presumably you're using paging because you have a lot of data that you don't want to retrieve and display all at once, so it's really only the last page that shows the empty rows.

For the tutorial, if we only have five sample movies, maybe we should make the page size five?

As for centering the text, we can do that too if you think it's important. I admit that I'm not a designer! :laughing:

**Comment by danroth27** [comment]
I'd like to still include the paginator. The main addition is to guide the user to style the rows with a fixed height so that the paginator doesn't jump to a different position on the last page. That way, every page of data shows the exact same styling for the rows and the paginator doesn't move. The borders are going to be there when there's data in the rows, right? Why would you not expect the same styling to be there on the last page when some of the rows are empty?

**Comment by guardrex** [comment]
I see your point. I opened #34217, and I'll resolve it tomorrow ... ***before I leave for the break and die of a*** 🦃 ***overdose!*** 😆

**Comment by coderdnewbie** [comment]
I tries to add the following scoped css in my {COMPONENT}.razor.css (in my case it is Index.razor.css as I am doing the Movie example from the beginners guide). 

Unfortunately, he scoped css is not getting applied (even after I pressed Ctrl+F5). Is there anything new that has to be done, since the css is in Assets?

This is the scoped css, I tried:

```
::deep tr:has(> td:not(:empty)) > td {
    display: table-cell;
}

::deep td:empty {
    display: none;
}
```

I also tried:

```
::deep tr {
    height: 2em;
}
```

but the scoped css is not being applied to the quickgrid. Please help.

**Comment by guardrex** [comment]
Did u wrap the Quickgrid in a block element like a `<div>`?

**Comment by coderdnewbie** [comment]
No. I followed the tutorial, so I used the scaffold pages. Here is the code:

```
@page "/movies"

@implements IAsyncDisposable
@rendermode InteractiveServer

<PageTitle>Index</PageTitle>

<h1>Index</h1>

<p>
    <a href="movies/create">Create New</a>
</p>

<QuickGrid Class="table" Items="FilteredMovies" Pagination="state">
    <PropertyColumn Property="movie => movie.Title" Sortable="true"> 
        <ColumnOptions>
            <div>
                <input type="search" @bind="TitleFilter" @bind:event="oninput" autofocus />
            </div>
        </ColumnOptions>
    </PropertyColumn>
    <PropertyColumn Property="movie => movie.ReleaseDate" Sortable="true" />
    <PropertyColumn Property="movie => movie.Genre" Sortable="true" />
    <PropertyColumn Property="movie => movie.Price" Sortable="true" />

    <TemplateColumn Context="movie">
        <a href="@($"movies/edit?id={movie.Id}")">Edit</a> |
        <a href="@($"movies/details?id={movie.Id}")">Details</a> |
        <a href="@($"movies/delete?id={movie.Id}")">Delete</a>
    </TemplateColumn>
</QuickGrid>

<Paginator State="state" />

@code {
    private IDbContextFactory<BlazorMovieCRUDDemoContext> dbFactory { get; set; }
    PaginationState state = new PaginationState { ItemsPerPage = 5 };

    string TitleFilter = string.Empty;
    IQueryable<Movie> FilteredMovies => context.Movie.Where(s => s.Title!.Contains(TitleFilter));

    public Index(IDbContextFactory<BlazorMovieCRUDDemoContext> dbFactory)
    {
        this.dbFactory = dbFactory;
    }

    private BlazorMovieCRUDDemoContext context = default!;

    protected override void OnInitialized()
    {
        context = dbFactory.CreateDbContext();
    }

    public async ValueTask DisposeAsync() => await context.DisposeAsync();
}

```

guardrex:
Yeah ... go ahead and wrap that `QuickGrid` in a `<div>...</div>` block and see if it works for you.

The tutorial will be updated to cover that soon. It's on a PR at ...

https://github.com/dotnet/AspNetCore.Docs/pull/34227

You can look at the diff on that PR to see what I'm proposing thus far for new coverage. I think Dan and I will have that merged next week when we get back from the 🦃 holiday.

guardrex:
... the sample app has been updated already. You can see the final component here ...

https://github.com/dotnet/blazor-samples/blob/main/9.0/BlazorWebAppMovies/Components/Pages/MoviePages/Index.razor

... and the isolated stylesheet here ...

https://github.com/dotnet/blazor-samples/blob/main/9.0/BlazorWebAppMovies/Components/Pages/MoviePages/Index.razor.css

AJ1000:
Can the scaffolding be updated as well?

AJ1000:
Thanks for the info, the scaffolding css should, in my opinion, have the css so it looks the same as dotnet 8. Then Dan can demo this without having to add css to cover this issue.

PR #34221, opened 2024-11-27
Remove Nowin WebSocket sample

halter73:
I would prefer to revert https://github.com/dotnet/AspNetCore.Docs/pull/19693, because by moving the Nowin sample, we removed any documentation on how to run ASP.NET Core using an OWIN server. But if we decide that isn't an important enough scenario to still document, since we don't know of any supported OWIN servers that run outside of .NET Framework, the "Run ASP.NET Core on an OWIN-based server and use its WebSockets support" section makes no sense. All it shows is bog standard WebSocket code. The only thing interesting about it was the code to make it work on Nowin which is thankfully still in the [sample directory](https://github.com/dotnet/AspNetCore.Docs/tree/9771604bd406c2415514f09a890d972b5ffb9a5f/aspnetcore/fundamentals/owin/sample/src/NowinWebSockets).

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/fundamentals/owin.md](https://github.com/dotnet/AspNetCore.Docs/blob/9d585fc0a0f1982a4b4d0d3e4610db2a141601a3/aspnetcore/fundamentals/owin.md) | [aspnetcore/fundamentals/owin](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/owin?branch=pr-en-us-34221) |

### DIFF
--- a/aspnetcore/fundamentals/owin.md
+++ b/aspnetcore/fundamentals/owin.md
@@ -90,58 +90,6 @@ app.UseOwin(pipeline =>
 });
 ```
 
-<a name="hosting-on-owin"></a>
-
-## Run ASP.NET Core on an OWIN-based server and use its WebSockets support
-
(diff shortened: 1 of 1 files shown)


guardrex (approved):
I'll leave it to Rick to merge, given that he worked on the original issue/PR.

PR #34222, opened 2024-11-27
Update bootstrap.bundle.js

v-thepet:
Fix Policheck bug https://dev.azure.com/msft-skilling/learn-platform-policheck/_workitems/edit/345733


v-thepet:
@guardrex - Can you review and merge if approved? thanks

guardrex (approved):
Yes, @v-thepet, we can change it. Note in passing that Bootstrap isn't a MS product, and it's in their source code ...

https://github.com/twbs/bootstrap/blob/v4.3.1/dist/js/bootstrap.bundle.js#L1505

The good news is that they don't have this license remark with that word in their latest versions of this file. It shouldn't crop up again with future sample apps that use Bootstrap.

The article and sample was created by @damienbod with @Rick-Anderson as the MS author. v4.3.1 is fairly old. The current version is 5.3.3. Idk if they'll want to schedule a Bootstrap update for the sample app at some point.

PR #34241, opened 2024-11-29
[Blazor] webassembly-lazy-load-assemblies - code

hakenr:
It's not just the `@code` block - it can also be the code-behind.


guardrex (approved):
# 👍

PR #34242, opened 2024-11-29
[Blazor] webassembly-lazy-load-assemblies.md - Forms package already referenced

hakenr:
The `Microsoft.AspNetCore.Components.Forms` package is already transitively referenced by `Microsoft.AspNetCore.Components.Web` package (which is included in the RCL template). No need to add explicit reference to `Forms` package.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/webassembly-lazy-load-assemblies.md](https://github.com/dotnet/AspNetCore.Docs/blob/e2a24fd658a0bebf94855c5264b3cf84bcdfd8d4/aspnetcore/blazor/webassembly-lazy-load-assemblies.md) | [Lazy load assemblies in ASP.NET Core Blazor WebAssembly](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/webassembly-lazy-load-assemblies?branch=pr-en-us-34242) |

### DIFF
--- a/aspnetcore/blazor/webassembly-lazy-load-assemblies.md
+++ b/aspnetcore/blazor/webassembly-lazy-load-assemblies.md
@@ -339,10 +339,6 @@ Add an ASP.NET Core class library project to the solution:
 * Visual Studio: Right-click the solution file in **Solution Explorer** and select **Add** > **New project**. From the dialog of new project types, select **Razor Class Library**. Name the project `GrantImaharaRobotControls`. Do **not** select the **Support pages and view** checkbox.
 * Visual Studio Code/.NET CLI: Execute `dotnet new razorclasslib -o GrantImaharaRobotControls` from a command prompt. The `-o|--output` option creates a folder and names the project `GrantImaharaRobotControls`.
 
-The example component presented later in this section uses a [Blazor form](xref:blazor/forms/index). In the RCL project, add the [`Microsoft.AspNetCore.Components.Forms`](https://www.nuget.org/packages/Microsoft.AspNetCore.Components.Forms) package to the project.
-
-[!INCLUDE[](~/includes/package-reference.md)]
-
 Create a `HandGesture` class in the RCL with a `ThumbUp` method that hypothetically makes a robot perform a thumbs-up gesture. The method accepts an argument for the axis, `Left` or `Right`, as an [`enum`](/dotnet/csharp/language-reference/builtin-types/enum). The method returns `true` on success.
 
 `HandGesture.cs`:
--- a/aspnetcore/blazor/webassembly-lazy-load-assemblies.md
+++ b/aspnetcore/blazor/webassembly-lazy-load-assemblies.md
@@ -336,7 +336,7 @@ Create a standalone Blazor WebAssembly app to demonstrate lazy loading of a Razo
 
 Add an ASP.NET Core class library project to the solution: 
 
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/webassembly-lazy-load-assemblies.md
```diff
@@ -339,10 +339,6 @@ Add an ASP.NET Core class library project to the solution:
 * Visual Studio: Right-click the solution file in **Solution Explorer** and select **Add** > **New project**. From the dialog of new project types, select **Razor Class Library**. Name the project `GrantImaharaRobotControls`. Do **not** select the **Support pages and view** checkbox.
```
guardrex, line 339: There's a NIT 😈 here to fix ...

```suggestion
* Visual Studio: Right-click the solution file in **Solution Explorer** and select **Add** > **New project**. From the dialog of new project types, select **Razor Class Library**. Name the project `GrantImaharaRobotControls`. Do **not** select the **Support pages and views** checkbox.
```

aspnetcore/blazor/webassembly-lazy-load-assemblies.md
```diff
@@ -339,10 +339,6 @@ Add an ASP.NET Core class library project to the solution:
...
 * Visual Studio Code/.NET CLI: Execute `dotnet new razorclasslib -o GrantImaharaRobotControls` from a command prompt. The `-o|--output` option creates a folder and names the project `GrantImaharaRobotControls`.
 
-The example component presented later in this section uses a [Blazor form](xref:blazor/forms/index). In the RCL project, add the [`Microsoft.AspNetCore.Components.Forms`](https://www.nuget.org/packages/Microsoft.AspNetCore.Components.Forms) package to the project.
```
guardrex, line 342: It might be better to state what's going on with the packages here. This could instead say something like ...

> The example component presented later in this section uses a \[Blazor form](xref:blazor/forms/index). Because the RCL was created without supporting pages and views, the RCL includes a package reference for \[\`Microsoft.AspNetCore.Components.Web\`](https://www.nuget.org/packages/Microsoft.AspNetCore.Components.Web), which transitivity includes the \[\`Microsoft.AspNetCore.Components.Forms\`](https://www.nuget.org/packages/Microsoft.AspNetCore.Components.Forms) package.
>
> \[!INCLUDE\[](~/includes/package-reference.md)]

guardrex (changes requested):
See if my suggestion makes sense. I agree that the text is wrong, but a replacement could meet the original goal of making it clear what packages are important here.

aspnetcore/blazor/webassembly-lazy-load-assemblies.md
hakenr: If you create the RCL with "Support pages and view" checked, the new project still successfully compiles the `Robot.razor` component (with `<EditForm>` and `<InputRadioGroup>`), most likely because of `<FrameworkReference Include="Microsoft.AspNetCore.App" />`.

Why do you think we need to include package-related info in the article?

guardrex: Truly, it was just a passing concern. I didn't know what would happen in that configuration. Let's go ahead with your suggestion.

guardrex (approved):
Thanks for fixing this! 🎉

PR #19211, opened 2020-07-15
Blazor WASM lazy load assemblies topic

guardrex:
Fixes #19149

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/webassembly-lazy-load-assemblies?view=aspnetcore-5.0&branch=pr-en-us-19211)

* This is the *hybrid* tutorial-like reference topic approach. We could just plop the complete example down and write thereafter about how it works, but I think there's merit to the approach on the PR that breaks it down into logical sections. This is the approach that you took with the design notes on the PR, and I think it works well here. However, let me know if you'd like to go the other way. I'm not opposed to a rewrite if you think it best.
* I originally pitched this for the *Host and deploy* node, but this is *really, REALLY* important perf that many devs are going to want to use. Therefore, I now recommend that we place it in the open *Blazor* node near the existing WASM perf topic. The PR has been set up that way.
* "Sales" was a bit 🥱 lol, so I started with a kind of Easter egg with "Ion Control" (an `IonControl.dll` assembly and an `/admin/ioncontrol` endpoint). I guess *The Empire Strikes Back* was on my mind. I changed that tho: I think it would be nice to do something in the example to celebrate the life of Grant Imahara, who passed away a few days ago unexpectedly. :cry:

  Capture
  *Grant Imahara (1970-2020) &mdash; Will be greatly missed.*
  https://en.wikipedia.org/wiki/Grant_Imahara
  https://www.imdb.com/name/nm0408017/

Issue #19149, opened 2020-07-08
Add documentation on lazy-loading support for Blazor


captainsafia:
For preview8, we need to add documentation on Blazor's new support for lazy-loading.

@guardrex This [PR description](https://github.com/dotnet/aspnetcore/pull/23290#issue-439057914) can be a useful starting ground for this. Let me know if you need any more info.

guardrex:
@captainsafia ... Do you feel that this fits best as a new section in the performance best practices topic ...

https://docs.microsoft.com/en-us/aspnet/core/blazor/webassembly-performance-best-practices

... or as a new section in the Blazor WASM Host and deploy topic ...

https://docs.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/webassembly?view=aspnetcore-3.1

... or as a new topic in the Host and deploy node given that there's enough content to merit creating a new topic from your remarks on the engineering issue?

If a new topic, then perhaps ...

* **_Title:_** Lazy load assemblies in ASP.NET Core Blazor WebAssembly
* **_Physical location/filename:_** `aspnetcore/blazor/host-and-deploy/lazy-load-assemblies.md`
* **_TOC entry:_** Blazor > Host and deploy > Lazy load assemblies (as the last entry in the node)

Note: We try to avoid gerunds in titles, section headings, and TOC entries. When used as a noun/subject, they apparently don't translate well. Let's see if we can just use "load" in titles and section headings for this.

captainsafia:
> ... or as a new topic in the Host and deploy node given that there's enough content to merit creating a new topic from your remarks on the engineering issue?

I like this option the most since there is enough content to warrant an entire page.

> * **_Title:_** Lazy load assemblies in ASP.NET Core Blazor WebAssembly
> * **_Physical location/filename:_** `aspnetcore/blazor/host-and-deploy/lazy-load-assemblies.md`
> * **_TOC entry:_** Blazor > Host and deploy > Lazy load assemblies (as the last entry in the node)

This looks good!

guardrex:
ok ... I'll get to work on that very soon ... probably tomorrow (Friday) tho. The family :cat2:, *Merlin*, woke me up at **_4am_** (!!!! 😠) to be fed and let outside. I was wide awake, so started at 4am ... my workday is almost over. :smile:

captainsafia:
Cool! Ping me with any questions.

[...]

PR #34243, opened 2024-11-29
[Blazor] Lazy loading - Complete example - Prevent repeated loads

hakenr:
Even though repeated calls to `AssemblyLoader.LoadAssembliesAsync()` don't download the assembly multiple times or throw exception:  
- They still trigger asynchronous calls to JavaScript.  
- The `<Navigating>` content is rendered in the meantime, causing unnecessary UI flickering when navigating to the lazy-loaded area repeatedly.  
- The `lazyLoadedAssemblies` collection gets populated with duplicate entries.  

Therefore, lazy-loading should be guarded with an "already loaded" flag.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/webassembly-lazy-load-assemblies.md](https://github.com/dotnet/AspNetCore.Docs/blob/6d848125dd50a3e46c68b1ffc5b3243c5815803a/aspnetcore/blazor/webassembly-lazy-load-assemblies.md) | [aspnetcore/blazor/webassembly-lazy-load-assemblies](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/webassembly-lazy-load-assemblies?branch=pr-en-us-34243) |

### DIFF
--- a/aspnetcore/blazor/webassembly-lazy-load-assemblies.md
+++ b/aspnetcore/blazor/webassembly-lazy-load-assemblies.md
@@ -658,10 +658,10 @@ The assembly is assigned to <xref:Microsoft.AspNetCore.Components.Routing.Router
         {
             if ((args.Path == "robot") && !grantImaharaRobotControlsAssemblyLoaded)
             {
-                grantImaharaRobotControlsAssemblyLoaded = true;
                 var assemblies = await AssemblyLoader.LoadAssembliesAsync(
                     new[] { "GrantImaharaRobotControls.{FILE EXTENSION}" });
                 lazyLoadedAssemblies.AddRange(assemblies);
+                grantImaharaRobotControlsAssemblyLoaded = true;
             }
         }
         catch (Exception ex)
--- a/aspnetcore/blazor/webassembly-lazy-load-assemblies.md
+++ b/aspnetcore/blazor/webassembly-lazy-load-assemblies.md
@@ -604,10 +604,10 @@ The assembly is assigned to <xref:Microsoft.AspNetCore.Components.Routing.Router
         {
             if ((args.Path == "robot") && !grantImaharaRobotControlsAssemblyLoaded)
             {
-                grantImaharaRobotControlsAssemblyLoaded = true;
                 var assemblies = await AssemblyLoader.LoadAssembliesAsync(
                     new[] { "GrantImaharaRobotControls.{FILE EXTENSION}" });
                 lazyLoadedAssemblies.AddRange(assemblies);
+                grantImaharaRobotControlsAssemblyLoaded = true;
             }
         }
         catch (Exception ex)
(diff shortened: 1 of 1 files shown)


guardrex:
# 🤔 

This wasn't part of Safia's original bits for the feature ... 

https://github.com/dotnet/aspnetcore/pull/23290

... and it never made its way into the test component (but they might want to add it after they see what you wrote) ...

https://github.com/dotnet/aspnetcore/blob/main/src/Components/test/testassets/BasicTestApp/RouterTest/TestRouterWithLazyAssembly.razor

... and it didn't come up for discussion on the docs issue or PR ...

* https://github.com/dotnet/AspNetCore.Docs/issues/19149
* https://github.com/dotnet/AspNetCore.Docs/pull/19211

I've added the Blazor engineering group to see if anyone is free to take a look at this. If we don't hear back within a couple of days, I'll reach out offline to get someone on here. Might take a while because reviews seem hard to get during the 🦃 holiday season. *Everyone apparently ate so much* 🦃 *that they can't get to the computer to look!* 😆

SteveSandersonMS (approved):

SteveSandersonMS:
Adding this guard here is mainly beneficial for the sample code, since otherwise (like @hakenr mentions) it ends up duplicating data. That's not an issue in the product itself; it's just a quirk of the sample code in the docs.

Whether or not people want to use a guard clause to ensure they don't see any flash of "loading..." content is up to the app developer.

Altogether the changes suggested here are good and I think they should be added to the docs, as it gives readers more of a sense of things they might want to do to get good end-to-end behavior in their own apps.

aspnetcore/blazor/webassembly-lazy-load-assemblies.md
```diff
@@ -596,13 +596,15 @@ The assembly is assigned to <xref:Microsoft.AspNetCore.Components.Routing.Router
...
+    private bool grantImaharaRobotControlsAssemblyLoaded;
 
     private async Task OnNavigateAsync(NavigationContext args)
     {
         try
         {
-            if (args.Path == "robot")
+            if ((args.Path == "robot") && !grantImaharaRobotControlsAssemblyLoaded)
             {
+                grantImaharaRobotControlsAssemblyLoaded = true;
```
guardrex, line 607: NIT question on the location of this line ...

Because the component survives an exception on the next two lines, does it make sense to move this line (`grantImaharaRobotControlsAssemblyLoaded = true;`) to the end after the `lazyLoadedAssemblies.AddRange(assemblies);` line (i.e., it's only `true` if we know those two lines executed)?

guardrex: I'm going to go ahead and move the line down to get this in. I'm about to shoot a bunch of PRs into the live branch.

guardrex (approved):
# 🎉

aspnetcore/blazor/webassembly-lazy-load-assemblies.md
hakenr: @guardrex Neither of the two alternatives is perfect. I usually prefer to set the flag before an awaited operation - because with the `await` there, another navigation might occur while waiting for the assembly to load, leaving the flag still `false` for the second run.  

In this case, since the awaited assembly load won’t cause a critical failure if called repeatedly, I agree that your version is better.

PR #34281, opened 2024-12-03
[Blazor] Performance - Change detection - set parameters only

hakenr:
1. When I think back to what wasn’t entirely clear to me and I had to test and look up in the source code - the whole condition for primitive types in change detection applies to the parameters **used (set)** in a specific place, not to the component **having** non-primitive parameters.
2. Starting net9 EventCallbacks are included in change detection.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/performance.md](https://github.com/dotnet/AspNetCore.Docs/blob/b79fd5034b4902ee7429e140f1aedf21c6fa2a00/aspnetcore/blazor/performance.md) | [ASP.NET Core Blazor performance best practices](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/performance?branch=pr-en-us-34281) |

### DIFF
--- a/aspnetcore/blazor/performance.md
+++ b/aspnetcore/blazor/performance.md
@@ -35,7 +35,7 @@ The last two steps of the preceding sequence continue recursively down the compo
 
 To prevent rendering recursion into a particular subtree, use either of the following approaches:
 
-* Ensure that the *set* parameters of child components are of primitive immutable types, such as `string`, `int`, `bool`, `DateTime`, and other similar types. The built-in logic for detecting changes automatically skips rerendering if the primitive immutable parameter values haven't changed. Only the parameters that are explicitly set on the component are considered for change detection. If you render a child component with `<Customer CustomerId="item.CustomerId" />`, where `CustomerId` is an `int` type, then the `Customer` component isn't rerendered unless `item.CustomerId` changes.
+* Ensure that the **set** parameters of child components are of primitive immutable types, such as `string`, `int`, `bool`, `DateTime`, and other similar types. The built-in logic for detecting changes automatically skips rerendering if the primitive immutable parameter values haven't changed. Only the parameters that are explicitly set on the component are considered for change detection. If you render a child component with `<Customer CustomerId="item.CustomerId" />`, where `CustomerId` is an `int` type, then the `Customer` component isn't rerendered unless `item.CustomerId` changes.
 * Override <xref:Microsoft.AspNetCore.Components.ComponentBase.ShouldRender%2A>:
   * To accept nonprimitive parameter values, such as complex custom model types, event callbacks, or <xref:Microsoft.AspNetCore.Components.RenderFragment> values.
   * If authoring a UI-only component that doesn't change after the initial render, regardless of parameter value changes.
--- a/aspnetcore/blazor/performance.md
+++ b/aspnetcore/blazor/performance.md
@@ -35,7 +35,7 @@ The last two steps of the preceding sequence continue recursively down the compo
 
 To prevent rendering recursion into a particular subtree, use either of the following approaches:
 
-* Ensure that child component parameters are of primitive immutable types, such as `string`, `int`, `bool`, `DateTime`, and other similar types. The built-in logic for detecting changes automatically skips rerendering if the primitive immutable parameter values haven't changed. If you render a child component with `<Customer CustomerId="item.CustomerId" />`, where `CustomerId` is an `int` type, then the `Customer` component isn't rerendered unless `item.CustomerId` changes.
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/performance.md
```diff
@@ -33,12 +33,27 @@ At runtime, components exist in a hierarchy. A root component (the first compone
...
+
 To prevent rendering recursion into a particular subtree, use either of the following approaches:
 
-* Ensure that child component parameters are of primitive immutable types, such as `string`, `int`, `bool`, `DateTime`, and other similar types. The built-in logic for detecting changes automatically skips rerendering if the primitive immutable parameter values haven't changed. If you render a child component with `<Customer CustomerId="item.CustomerId" />`, where `CustomerId` is an `int` type, then the `Customer` component isn't rerendered unless `item.CustomerId` changes.
+* Ensure that the **set** parameters of child components are of primitive immutable types, such as `string`, `int`, `bool`, `DateTime`, and other similar types. The built-in logic for detecting changes automatically skips rerendering if the primitive immutable parameter values haven't changed. Only the parameters that are explicitly set on the component are considered for change detection. If you render a child component with `<Customer CustomerId="item.CustomerId" />`, where `CustomerId` is an `int` type, then the `Customer` component isn't rerendered unless `item.CustomerId` changes.
```
guardrex, line 40: This seems like a good change, but I don't think emphasizing the word "set" has much value. Can we just drop that bold text there?

```suggestion
* Ensure that the set parameters of child components are of primitive immutable types, such as `string`, `int`, `bool`, `DateTime`, and other similar types. The built-in logic for detecting changes automatically skips rerendering if the primitive immutable parameter values haven't changed. Only the parameters that are explicitly set on the component are considered for change detection. If you render a child component with `<Customer CustomerId="item.CustomerId" />`, where `CustomerId` is an `int` type, then the `Customer` component isn't rerendered unless `item.CustomerId` changes.
```

aspnetcore/blazor/performance.md
```diff
@@ -33,12 +33,27 @@ At runtime, components exist in a hierarchy. A root component (the first compone
...
+  * To accept nonprimitive parameter values, such as complex custom model types, or <xref:Microsoft.AspNetCore.Components.RenderFragment> values.
+  * If authoring a UI-only component that doesn't change after the initial render, regardless of parameter value changes.
+ 
+:::moniker-end
+
+:::moniker range="< aspnetcore-9.0"
+
+To prevent rendering recursion into a particular subtree, use either of the following approaches:
+
+* Ensure that the **set** parameters of child components are of primitive immutable types, such as `string`, `int`, `bool`, `DateTime`, and other similar types. The built-in logic for detecting changes automatically skips rerendering if the primitive immutable parameter values haven't changed. Only the parameters that are explicitly set on the component are considered for change detection. If you render a child component with `<Customer CustomerId="item.CustomerId" />`, where `CustomerId` is an `int` type, then the `Customer` component isn't rerendered unless `item.CustomerId` changes.
```
guardrex, line 51:
```suggestion
* Ensure that the set parameters of child components are of primitive immutable types, such as `string`, `int`, `bool`, `DateTime`, and other similar types. The built-in logic for detecting changes automatically skips rerendering if the primitive immutable parameter values haven't changed. Only the parameters that are explicitly set on the component are considered for change detection. If you render a child component with `<Customer CustomerId="item.CustomerId" />`, where `CustomerId` is an `int` type, then the `Customer` component isn't rerendered unless `item.CustomerId` changes.
```

aspnetcore/blazor/performance.md
```diff
@@ -33,12 +33,27 @@ At runtime, components exist in a hierarchy. A root component (the first compone
...
+* Override <xref:Microsoft.AspNetCore.Components.ComponentBase.ShouldRender%2A>:
+  * To accept nonprimitive parameter values, such as complex custom model types, or <xref:Microsoft.AspNetCore.Components.RenderFragment> values.
```
guardrex, line 42:
```suggestion
  * To accept nonprimitive parameter values, such as complex custom model types or <xref:Microsoft.AspNetCore.Components.RenderFragment> values.
```

guardrex (changes requested):
Looks good except for a couple of NIT suggestions.

One more thing ... can you move the lead-in sentence outside of the bullet list?

> To prevent rendering recursion into a particular subtree, use either of the following approaches:

Move that ☝️ to one line above the first `moniker-range`.

hakenr:
All your suggestions should be resolved now.

guardrex (approved):
Excellent!

PR #34291, opened 2024-12-03
[Blazor] Performance - Parameter values bundling vs. change detection

hakenr:
---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/performance.md](https://github.com/dotnet/AspNetCore.Docs/blob/1a61f450cfb5e14abf44ed7e7654abe052b32582/aspnetcore/blazor/performance.md) | [aspnetcore/blazor/performance](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/performance?branch=pr-en-us-34291) |

### DIFF
--- a/aspnetcore/blazor/performance.md
+++ b/aspnetcore/blazor/performance.md
@@ -228,7 +228,9 @@ To reduce parameter load, bundle multiple parameters in a custom class. For exam
 }
 ```
 
-However, consider that it might be an improvement not to have a table cell component, as shown in the preceding example, and instead [inline its logic into the parent component](#inline-child-components-into-their-parents).
+However, keep in mind that bundling primitive parameters into a class isn't always an advantage. While it can reduce parameter count, it can also impact how change detection and rendering behave. Setting non-primitive parameters causes the component to rerender every time, even if the values themselves haven't changed. This behavior can offset the potential benefits.
+
+Also, consider that it might be an improvement not to have a table cell component, as shown in the preceding example, and instead [inline its logic into the parent component](#inline-child-components-into-their-parents).
 
 > [!NOTE]
 > When multiple approaches are available for improving performance, benchmarking the approaches is usually required to determine which approach yields the best results.
--- a/aspnetcore/blazor/performance.md
+++ b/aspnetcore/blazor/performance.md
@@ -228,7 +228,7 @@ To reduce parameter load, bundle multiple parameters in a custom class. For exam
 }
 ```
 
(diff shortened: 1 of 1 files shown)


guardrex:
Doesn't seem like it's related to the thrust of the argument in the section. Let's have one of the engineers have a look at the idea of placing this remark here.

guardrex:
**UPDATE (12/6)**: *Wow!* Reviews are really slow 🐌 right now due to internal activity. They told me that they're working on getting to your PRs, and they'll be here as soon as they can.

aspnetcore/blazor/performance.md
```diff
@@ -228,7 +228,9 @@ To reduce parameter load, bundle multiple parameters in a custom class. For exam
 }
 ```
 
-However, consider that it might be an improvement not to have a table cell component, as shown in the preceding example, and instead [inline its logic into the parent component](#inline-child-components-into-their-parents).
+However, keep in mind that bundling primitive parameters into a class isn't always an advantage. While it can reduce parameter count, it can also impact how change detection and rendering behave. Setting non-primitive parameters causes the component to rerender every time, even if the values themselves haven't changed. This behavior can offset the potential benefits.
```
SteveSandersonMS, line 231:
```suggestion
However, keep in mind that bundling primitive parameters into a class isn't always an advantage. While it can reduce parameter count, it will also impact how change detection and rendering behave. Passing non-primitive parameters always triggers a re-render, because Blazor can't know whether arbitrary objects have internally mutable state, whereas passing primitive parameters only triggers a re-render if their values have actually changed.
```

SteveSandersonMS (approved):

aspnetcore/blazor/performance.md
```diff
@@ -228,7 +228,9 @@ To reduce parameter load, bundle multiple parameters in a custom class. For exam
 }
 ```
 
-However, consider that it might be an improvement not to have a table cell component, as shown in the preceding example, and instead [inline its logic into the parent component](#inline-child-components-into-their-parents).
+However, keep in mind that bundling primitive parameters into a class isn't always an advantage. While it can reduce parameter count, it will also impact how change detection and rendering behave. Passing non-primitive parameters always triggers a re-render, because Blazor can't know whether arbitrary objects have internally mutable state, whereas passing primitive parameters only triggers a re-render if their values have actually changed.
```
guardrex, line 231: Quick NIT just because we try not to write in future tense ... this won't change the meaning of the text ...

```suggestion
However, keep in mind that bundling primitive parameters into a class isn't always an advantage. While it can reduce parameter count, it also impacts how change detection and rendering behave. Passing non-primitive parameters always triggers a re-render, because Blazor can't know whether arbitrary objects have internally mutable state, whereas passing primitive parameters only triggers a re-render if their values have actually changed.
```

guardrex (approved):
# 🎉

PR #34293, opened 2024-12-04
[Blazor] Performance - Event callbacks in RenderFragments

hakenr:
I'm not sure what was meant by "event callbacks aren't supported" in this context.

This code with event callback works as expected:
```razor
@page "/counter"
@using Microsoft.Extensions.Logging
@inject ILogger<Counter> Logger

@MyRenderFragment

@code {
	private RenderFragment? MyRenderFragment => @<button @onclick="HandleValidSubmit">Submit</button>;

	private void HandleValidSubmit()
	{
		Logger.LogInformation("HandleValidSubmit called");
	}
}
```

(Don't worry, this is my last PR for today. 😇)

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/performance.md](https://github.com/dotnet/AspNetCore.Docs/blob/aef6f508c6b6ced745a461b5fa46faf324392a4c/aspnetcore/blazor/performance.md) | [ASP.NET Core Blazor performance best practices](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/performance?branch=pr-en-us-34293) |

### DIFF
--- a/aspnetcore/blazor/performance.md
+++ b/aspnetcore/blazor/performance.md
@@ -195,7 +195,7 @@ public static RenderFragment SayHello = @<h1>Hello!</h1>;
 }
 ```
 
-The preceding approach reuses rendering logic without per-component overhead. However, the approach doesn't permit refreshing the subtree of the UI independently, nor does it have the ability to skip rendering the subtree of the UI when its parent renders because there's no component boundary. Assignment to a <xref:Microsoft.AspNetCore.Components.RenderFragment> delegate is only supported in Razor component files (`.razor`), and [event callbacks](xref:blazor/components/event-handling#eventcallback) aren't supported.
+The preceding approach reuses rendering logic without per-component overhead. However, the approach doesn't permit refreshing the subtree of the UI independently, nor does it have the ability to skip rendering the subtree of the UI when its parent renders because there's no component boundary. Assignment to a <xref:Microsoft.AspNetCore.Components.RenderFragment> delegate is only supported in Razor component files (`.razor`).
 
 For a non-static field, method, or property that can't be referenced by a field initializer, such as `TitleTemplate` in the following example, use a property instead of a field for the <xref:Microsoft.AspNetCore.Components.RenderFragment>:
 


guardrex:
I don't think that's what this means. Doesn't it mean that you can't pass the event callback to the render fragment?

hakenr:
The code is a little bit obscure, but still works:
```razor
@page "/counter"
@using Microsoft.AspNetCore.Components.Forms
@using Microsoft.Extensions.Logging
@inject ILogger<Counter> Logger

@MyRenderFragment(EventCallback.Factory.Create(this, HandleValidSubmit))

@code {
	private RenderFragment<EventCallback>? MyRenderFragment => callback => @<button @onclick="callback">Submit</button>;

	private void HandleValidSubmit()
	{
		Logger.LogInformation("HandleValidSubmit called");
	}
}
```

guardrex:
Yes, that was more along the lines of what I was thinking.

What about this being a point-in-time remark? Can you check that in a 3.1 app and see if it works that far back?

hakenr:
🤯🙄🤣
Unfortunately, I don't have that old runtime/SDK installed on my machine, and I’d rather not clutter my system with it right now :-D.  
I tested it on .NET 6, and it works fine there.  
Are there any online Blazor fiddles that support 3.1?  
Hope you drop those ancient unsupported versions soon. 🙏

guardrex:
Me neither! I think I can go back to 5.0 ... that's it. 😄

If I have 5.0, I'll try it tomorrow on 5.0 and report back.

> Hope you drop those ancient unsupported versions soon. 

No 🎲🎲 ... By Microsoft decree, we're required to keep ALL prior versions of docs *forever*. What we might end up implementing later is *Version-by-file* versioning, where each version of an article is in a separate file. That would cut out all of this moniker-range flipping insanity.

hakenr:
> By Microsoft decree, we're required to keep ALL prior versions of docs forever.

You can keep it, but don’t let it interfere with updates to the supported version docs. My suggestion is to keep a separate snapshot of documentation for those old versions (separate from the "live" doc sources) and focus on maintaining, updating, and improving only the supported versions.

guardrex:
It's already interfering IMO 😄. Management has decided that they want all of the old content live. They have no capacity with the current system to separate the content, and they won't allow snapshots (e.g., PDFs) to be the docs for prior releases, supported or out-of-support. Our hands are tied on it at this level. The only thing we have is moniker-range, ***and that breaks cross-links to headings in prior release content.*** I've been screaming for five years to give us Version-by-file. There's interest in it, but I still don't see the wheels of progress turning.

> focus on maintaining, updating, and improving only the supported versions.

That's kind'a the way that the main doc set is maintained ... not Blazor tho ... not yet and not while I'm here. I maintain all versions of Blazor as best I can. It's tough 😩, but I'm hanging in there.

guardrex:
I'm surprised to see that I have 3.1 on this machine.

Yes, both of the examples run fine under 3.1. Let's proceed with this update! 🎉

guardrex (approved):
# 👍

PR #34315, opened 2024-12-09
Add CookieOidcRefresher section

### LINKED ISSUE #34235: BlazorWebAppOidc sample: incorrect CookieOidcRefresher implementation
### Description

`CookieOidcRefresher` expects an ID token from the token endpoint. This seems wrong, based on the [spec](https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse), [the Duende implementation](https://github.com/DuendeSoftware/foss/blob/551666e9c40975d8dd76c346f6b29321cacf791b/access-token-management/src/AccessTokenManagement.OpenIdConnect/UserTokenEndpointService.cs#L138) and the fact that it always fails in my app.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/security/blazor-web-app-with-oidc?view=aspnetcore-8.0&pivots=without-bff-pattern

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/security/blazor-web-app-with-oidc.md

### Document ID

c3346d3a-346b-8db4-2650-ec044b3f0dd9

### Article author

@guardrex

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+c3346d3a-346b-8db4-2650-ec044b3f0dd9)

guardrex:
Fixes #34235

Thanks @albi005! 🚀 ... Would you check this and see if you think it will be sufficient to have helped you? I realize that it would be the most helpful to show a bit of code that does what Halter is calling out, but I'd leave that to him to produce, and I think it's safe to say that he's going to leave that to the dev to work out based on his response (or based on looking at what's in the code he placed in that PU (product unit) issue discussion, which is going to be cross-linked here for devs to find. A dev wanting help implementing the approach he recommends on your doc issue could ask on SO for help or even open an issue on the PU's repo (if they hit a problem). Anyway ... I pieced the content together from the discussion. I'll wait to merge this until later today to give you a chance to take a look and see if you have any suggestions on improving it.

Notes

* There are two copies of this in the article because the article uses a *pivot* to display content for the BFF and non-BFF patterns. The section I'm adding will appear in each pivot of the article.
* I've made a note to myself in the content to check the PU issue for 10.0 updates. I'll probably hear about changes via normal 10.0 tracking, but having the note here will make sure that I put eyes 👀 on what they're going to do.
* I'm setting this up now for <10.0 because I assume that Halter (or someone else) will fix the issue(s) with the refresher at 10.0. I'll adjust that later if changes get pushed off to 11.0.

### DIFF
--- a/aspnetcore/blazor/security/blazor-web-app-with-oidc.md
+++ b/aspnetcore/blazor/security/blazor-web-app-with-oidc.md
@@ -232,6 +232,19 @@ Inspect the sample app for the following features:
 
 For more information on (web) API calls using a service abstractions in Blazor Web Apps, see <xref:blazor/call-web-api#service-abstractions-for-web-api-calls>.
 
+:::moniker range="< aspnetcore-10.0"
+
+<!-- UPDATE 10.0 - Check the PU issue for 10.0 work to resolve both issues.
+                   The docs issue is https://github.com/dotnet/AspNetCore.Docs/issues/34235. -->
+
+## Custom cookie refresher (`CookieOidcRefresher.cs`)
+
+The custom cookie refresher (`CookieOidcRefresher.cs`) implementation might need to be updated depending on your Identity Provider. For example, the refresher expects an ID token from the token endpoint. If the refresh token isn't an ID token with claims, the refresher can be updated with additional logic to continue to use the prior set of claims stored in the authentication cookie.
+
+Additionally, work is scheduled for .NET 10 (November, 2025) on [`BlazorWebAppOidc AddOpenIdConnect with GetClaimsFromUserInfoEndpoint = true doesn't propogate role claims to client` (`dotnet/aspnetcore` 58826)](https://github.com/dotnet/aspnetcore/issues/58826) to refresh claims from the `/userinfo` endpoint.
+
+:::moniker-end
+
 ## Client-side Blazor Web App project (`BlazorWebAppOidc.Client`)
 
 The `BlazorWebAppOidc.Client` project is the client-side project of the Blazor Web App.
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/security/blazor-web-app-with-oidc.md
```diff
@@ -654,6 +654,22 @@ The important changes to the `LogInOrOut` component are demonstrated in the foll
...
 ```
 
+:::moniker range="< aspnetcore-10.0"
+
+## Token refresh
+
+<!-- UPDATE 10.0 - Check the PU issue for 10.0 work to resolve both issues.
+                   The docs issue is https://github.com/dotnet/AspNetCore.Docs/issues/34235. -->
+
+The custom cookie refresher (`CookieOidcRefresher.cs`) implementation updates the user's claims automatically when they expire. The current implementation expects to receive an ID token from the token endpoint in exchange for the refresh token. This ID token is then used to overwrite the user's claims.
```
albi005, line 664:
```suggestion
The custom cookie refresher (`CookieOidcRefresher.cs`) implementation updates the user's claims automatically when they expire. The current implementation expects to receive an ID token from the token endpoint in exchange for the refresh token. The claims in this ID token are then used to overwrite the user's claims.
```

aspnetcore/blazor/security/blazor-web-app-with-oidc.md
```diff
@@ -654,6 +654,22 @@ The important changes to the `LogInOrOut` component are demonstrated in the foll
...
+
+The sample implementation doesn't include code for requesting claims from the `/userinfo` endpoint on token refresh. For more information, see [`BlazorWebAppOidc AddOpenIdConnect with GetClaimsFromUserInfoEndpoint = true doesn't propogate role claims to client` (`dotnet/aspnetcore` #58826)](https://github.com/dotnet/aspnetcore/issues/58826#issuecomment-2492738142).
```
albi005, line 666:
```suggestion
The sample implementation doesn't include code for requesting claims from the [UserInfo endpoint](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo) on token refresh. For more information, see [`BlazorWebAppOidc AddOpenIdConnect with GetClaimsFromUserInfoEndpoint = true doesn't propogate role claims to client` (`dotnet/aspnetcore` #58826)](https://github.com/dotnet/aspnetcore/issues/58826#issuecomment-2492738142).
```

aspnetcore/blazor/security/blazor-web-app-with-oidc.md
```diff
@@ -654,6 +654,22 @@ The important changes to the `LogInOrOut` component are demonstrated in the foll
...
+
+> [!NOTE]
+> Some identity providers only return an access token when using a [refresh token](https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse). The `CookieOidcRefresher` can be updated with additional logic to continue to use the prior set of claims stored in the authentication cookie.
```
albi005, line 669:
```suggestion
> Some identity providers [only return an access token when using a refresh token](https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse). The `CookieOidcRefresher` can be updated with additional logic to continue to use the prior set of claims stored in the authentication cookie or use the access token to request claims from the UserInfo endpoint.
```

albi005 (changes requested):

albi005 (changes requested):
The code for the BFF part is missing.

guardrex:
This is no longer in a pivot because I moved it down to your suggested location, so we only need one copy of the content.

albi005 (approved):

guardrex:
LGTM ... I'll go ahead and merge this.

@halter73, if you have any change suggestions, I'm 👂 here, and I'll put in a patch PR if needed. Otherwise, this is now tracked for .NET 10. I'll keep an 👁️ on it next year. 

Thanks again for your help, @albi005!

PR #34332, opened 2024-12-11
[Blazor] synchronous-js-interop-call-js.md - fully synchronous code sample

hakenr:
1. When mentioning `IJSInProcessObjectReference`, we should demonstrate its usage in the sample, not just retrieve it. I’ve added the line `var value = module.Invoke<string>("javascriptFunctionIdentifier");` to address this.

2. When discussing the performance benefits of synchronous WASM calls over asynchronous code (as this file is part of the Blazor Performance article), the entire sample should be synchronous for consistency.

### DIFF
--- a/aspnetcore/blazor/includes/js-interop/synchronous-js-interop-call-js.md
+++ b/aspnetcore/blazor/includes/js-interop/synchronous-js-interop-call-js.md
@@ -38,7 +38,7 @@ When working with <xref:Microsoft.JSInterop.IJSObjectReference> in ASP.NET Core 
         {
             var jsInProcess = (IJSInProcessRuntime)JS;
             module = await jsInProcess.Invoke<IJSInProcessObjectReference>("import", 
-            "./scripts.js");
+                "./scripts.js");
             var value = module.Invoke<string>("javascriptFunctionIdentifier");
         }
     }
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/includes/js-interop/synchronous-js-interop-call-js.md
```diff
@@ -36,18 +36,20 @@ When working with <xref:Microsoft.JSInterop.IJSObjectReference> in ASP.NET Core
     {
         if (firstRender)
         {
-            module = await JS.InvokeAsync<IJSInProcessObjectReference>("import", 
+            var jsInProcess = (IJSInProcessRuntime)JS;
+            module = await jsInProcess.Invoke<IJSInProcessObjectReference>("import", 
             "./scripts.js");
```
guardrex, line 41: Just going to move this over a bit ...

```suggestion
                "./scripts.js");
```

guardrex (approved):
Very nice! Tanay reviewed the example on #28091, but it was part of a larger PR with lots of other updates. I'm not surprised that it was overlooked. As for the original example before Tanay saw that PR, it was merely ...

```csharp
private IJSInProcessObjectReference module;

...

module = await JS.InvokeAsync<IJSInProcessObjectReference>("import", "./scripts.js");
```

That would go back to sometime in 2022, and it's too hard to track exactly where that came from because the PR doesn't reference the source of it.

Issue #34236, opened 2024-11-28
Add a section on cascading values/parameters for state management with multiple subscribers


hakenr:
### Description

I think the [Factor out the state preservation to a common location](https://learn.microsoft.com/en-us/aspnet/core/blazor/state-management?view=aspnetcore-9.0&pivots=server#factor-out-the-state-preservation-to-a-common-location) section has some issues that make it incomplete or problematic:

1. **Prerendering Issue**:  
   The provided example fails during prerendering because `ProtectedSessionStore.GetAsync` cannot be invoked from `OnInitialized`. To fix this, prerendering must be disabled for the example to function correctly. See [this code](https://github.com/dotnet/AspNetCore.Docs/blob/e425bba36f00b7b9959b5e5d82f370883e09851d/aspnetcore/blazor/state-management.md?plain=1#L501). A possible solution is changing the render mode to something like `new InteractiveServerRenderMode(prerender: false)`.

2. **Issue with Multiple Subscribers**:  
   The `CounterStateProvider` fails when more than one component subscribes to `[CascadingParameter] private CounterStateProvider? CounterStateProvider`. The cascading value in this case is the `CounterStateProvider` component instance (`this`), which doesn't change after being set (you could even set `IsFixed="true"`). Since the actual counter value (`CurrentCount`) changes internally, the `CascadingValue` component is not being re-rendered (allowing the new value to be distributed to subscribers) and updates to `CurrentCount` are not propagated to other components subscribed to the cascading value. (The only situation when `CascadingValue` component notifies the subscribers of changes is when the `CascadingValue` itself receives new parameters in `SetParametersAsync`.)

   **Steps to Reproduce**:  
   - Follow the documentation steps, fixing the prerendering issue as mentioned above.  
   - Create a new `SubCounter.razor` component, duplicating the content of `Counter.razor` but without the `@page` directive.  
   - Add `<SubCounter />` to `Counter.razor` so both components render simultaneously.  
   - Test the two _Increment_ buttons (one in `Counter.razor` and the other in `SubCounter.razor`). You'll notice that updating the counter in one component doesn't trigger updates in the other.  
   
   ![Screenshot of the issue](https://github.com/user-attachments/assets/f99423d7-80fa-44ee-9597-24477eab7eab)

[...]

PR #34337, opened 2024-12-11
State preservation (common provider) updates

guardrex:
Addresses #34236

@hakenr ... Let's see if we can make some progress with obvious problems in the section just to get the ball rolling.

I placed some 🦖 ***RexHacks!™*** 🙈 on here to at least improve things while the thrust of your main suggestion is discussed on the issue.

Three goals of this ...

* Change the section heading to say what this is. The state provider is in the "common location," so I'm seeing if the heading as "*Factor out state preservation to a common provider*" works.
* Discuss the prerendering situation and update the code to deal with it.
* Explicitly call out that this isn't good for multiple components that modify/render the state on the same page.

BTW ... I don't speak to getting access to state from a provider ***during prerendering***, which is something to consider at least remarking on. Given the backing storage, should the article include a line to the effect ...

> For state management during prerendering, choose a different storage mechanism and access state in an earlier lifecycle method.

Simple enough. Does that approach to covering it make sense? 🤔

If we get this polished up and place it in the article, then your issue is about adding a ***new section***, perhaps entitled, "*Factor out state preservation to a common location*," which is basically the old title of this section.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/cascading-values-and-parameters.md](https://github.com/dotnet/AspNetCore.Docs/blob/6f526abe36d7baa8e18d3a816c228803ae091c53/aspnetcore/blazor/components/cascading-values-and-parameters.md) | [aspnetcore/blazor/components/cascading-values-and-parameters](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/cascading-values-and-parameters?branch=pr-en-us-34337) |
| [aspnetcore/blazor/state-management.md](https://github.com/dotnet/AspNetCore.Docs/blob/6f526abe36d7baa8e18d3a816c228803ae091c53/aspnetcore/blazor/state-management.md) | [aspnetcore/blazor/state-management](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/state-management?branch=pr-en-us-34337) |

### DIFF
--- a/aspnetcore/blazor/components/cascading-values-and-parameters.md
+++ b/aspnetcore/blazor/components/cascading-values-and-parameters.md
@@ -517,4 +517,4 @@ The following `ExampleTabSet` component uses the `TabSet` component, which conta
 ## Additional resources
 
 * [Generic type support: Explicit generic types based on ancestor components](xref:blazor/components/generic-type-support#explicit-generic-types-based-on-ancestor-components)
-* [State management: Factor out the state preservation to a common location](xref:blazor/state-management?pivots=server#factor-out-the-state-preservation-to-a-common-location)
+* [State management: Factor out state preservation to a common provider](xref:blazor/state-management?pivots=server#factor-out-state-preservation-to-a-common-provider)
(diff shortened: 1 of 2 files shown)


hakenr:
added a small patch in #34353

PR #34375, opened 2024-12-16
Updates to XSS article

guardrex:
Tom ... Patch update as I was passing through looking for a spot to discuss Map Static Assets routing conventions and CSP headers. Blazor has a dedicated CSP article, but the main doc set doesn't, so I'm passing that issue to Rick for the main *Static Assets* article. Anyway, I was passing through here a noticed some NITs 😈 to touch in the *XSS* article.

* All of the characters for HTML attribute encoding can be listed, as there's only four of them (for the `HttpUtility` class anyway). Probably best to just list them. The current line ends ...

  > ... characters such as " and ".

  ... which doesn't make sense.
* HTML attribute encoding isn't a superset. It doesn't contain more characters. It contains fewer, so that should be a "subset."
* Missing period from one line.

I didn't convert any of the lists for auto-numbering because they're spaced out for readability. Alternatively, let me know if you'd like me to collapse them to one list-item per line with automatic numbering.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/security/cross-site-scripting.md](https://github.com/dotnet/AspNetCore.Docs/blob/f31b9a8883a240b4ea34d24079c44be1a08e242b/aspnetcore/security/cross-site-scripting.md) | [aspnetcore/security/cross-site-scripting](https://review.learn.microsoft.com/en-us/aspnet/core/security/cross-site-scripting?branch=pr-en-us-34375) |

### DIFF
--- a/aspnetcore/security/cross-site-scripting.md
+++ b/aspnetcore/security/cross-site-scripting.md
@@ -23,9 +23,9 @@ At a basic level, XSS works by tricking your application into inserting a `<scri
 
 1. Never put untrusted data into your HTML input, unless you follow the rest of the steps below. Untrusted data is any data that may be controlled by a cyberattacker, such as HTML form inputs, query strings, HTTP headers, or even data sourced from a database, as a cyberattacker may be able to breach your database even if they can't breach your application.
 
-2. Before putting untrusted data inside an HTML element, ensure it's HTML encoded. HTML encoding takes characters such as &lt; and changes them into a safe form like &amp;lt;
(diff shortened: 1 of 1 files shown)


aspnetcore/security/cross-site-scripting.md
tdykstra, line 1: >Alternatively, let me know if you'd like me to collapse them to one list-item per line with automatic numbering.

That is my preference.

tdykstra (approved):
LGTM, one comment.

PR #34432, opened 2025-01-02
Prohibition on `<script>` tags for modules

### LINKED ISSUE #34414: Add to the Documentation on Importing Javascript modules that the script file is automatically downloaded when the Page is Rendered
### Description

There is a feature that is not described in the docs on Javascript interop.  This feature is that this code will automatically download the js file when the page is navigated to.  In other words, there is no need to add a script tag to the App.razor file.  In fact, if you do add it, it will not download successfully. To see this issue, uncomment the script tag in App.razor to force the browser to attempt to download it.

```
module = await JS.InvokeAsync<IJSObjectReference>("import",
    "./Components/Pages/CallDotnet1.razor.js");
```

Here is a simple Repro: https://github.com/gregoryagu/CallDotNetFromJs/

I would suggest adding text to describe this feature so that the user does not try to manually add the script tag, and also it is a really cool feature. 

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/location-of-javascript?view=aspnetcore-9.0#load-a-script-from-an-external-javascript-file-js-collocated-with-a-component

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/javascript-interoperability/location-of-javascript.md

### Document ID

af240267-89c0-f192-7807-cf426001ad8d

### Article author

@guardrex

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+af240267-89c0-f192-7807-cf426001ad8d)

guardrex:
Fixes #34414

Thanks @gregoryagu! 🚀 ... I won't merge this until tomorrow morning (Friday) to give you a chance to review it.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/hosting-models.md](https://github.com/dotnet/AspNetCore.Docs/blob/dbb863e45c81eeaec222de05c3eed5ed77e24ab3/aspnetcore/blazor/hosting-models.md) | [aspnetcore/blazor/hosting-models](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/hosting-models?branch=pr-en-us-34432) |
| [aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md](https://github.com/dotnet/AspNetCore.Docs/blob/dbb863e45c81eeaec222de05c3eed5ed77e24ab3/aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet.md) | [aspnetcore/blazor/javascript-interoperability/call-javascript-from-dotnet](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/call-javascript-from-dotnet?branch=pr-en-us-34432) |

### DIFF
--- a/aspnetcore/blazor/hosting-models.md
+++ b/aspnetcore/blazor/hosting-models.md
@@ -97,7 +97,7 @@ A Blazor WebAssembly app built as a [Progressive Web App (PWA)](xref:blazor/prog
 
 The Blazor script handles:
 
-* Downloading the .NET runtime, Razor components, and the component's dependencies.
+* Downloading the .NET runtime, Razor components, and dependencies.
 * Initialization of the runtime.
 
 The size of the published app, its *payload size*, is a critical performance factor for an app's usability. A large app takes a relatively long time to download to a browser, which diminishes the user experience. Blazor WebAssembly optimizes payload size to reduce download times:
--- a/aspnetcore/blazor/hosting-models.md
+++ b/aspnetcore/blazor/hosting-models.md
@@ -98,7 +98,7 @@ A Blazor WebAssembly app built as a [Progressive Web App (PWA)](xref:blazor/prog
 The Blazor script handles:
 
 * Downloading the .NET runtime, Razor components, and dependencies.
-* Initialization of the runtime.
+* Runtime initialization.
 
 The size of the published app, its *payload size*, is a critical performance factor for an app's usability. A large app takes a relatively long time to download to a browser, which diminishes the user experience. Blazor WebAssembly optimizes payload size to reduce download times:
 
--- a/aspnetcore/blazor/includes/js-interop/js-collocation.md
+++ b/aspnetcore/blazor/includes/js-interop/js-collocation.md
@@ -153,7 +153,7 @@ export function showPrompt2(message) {
 ```
 
 > [!IMPORTANT]
-> Don't place a `<script>` tag for `JsCollocation2.razor.js` after the [Blazor script](xref:blazor/project-structure#location-of-the-blazor-script). The module is loaded automatically and cached. Adding a `<script>` tag results in the module not loading correctly.
+> Don't place a `<script>` tag for `JsCollocation2.razor.js` after the [Blazor script](xref:blazor/project-structure#location-of-the-blazor-script) because the module is loaded and cached automatically. Adding a `<script>` tag prevents the module from loading.
 
 Use of scripts and modules for collocated JS in a Razor class library (RCL) is only supported for Blazor's JS interop mechanism based on the <xref:Microsoft.JSInterop.IJSRuntime> interface. If you're implementing [JavaScript `[JSImport]`/`[JSExport]` interop](xref:blazor/js-interop/import-export-interop), see <xref:blazor/js-interop/import-export-interop#razor-class-library-rcl-collocated-js-is-unsupported>.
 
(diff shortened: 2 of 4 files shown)


guardrex:
@gregoryagu ... I think I'll need to do a little more work ...

* The `import` remark should be cross-linked, and I think that's a dynamic `import()` call, so I'll place that cross-link to MDN as a guess.
* Where I'm stating 'it won't load the module,' I think it's going to have to state somehow that that's merely because `type="module"` isn't in the `<script>` tag. Can you clarify from your testing if you placed the `<script>` tag with `type="module"` or not because I think if you have that it should load the script (from cache) when the dynamic import is executed by Blazor.

guardrex:
I think on the second point that it's best to remove the remark. I really think that you can probably add a `<script>` tag for the module ***IF*** you properly declare `type="module"`. It's a guess, but it seems like a good bet. The module would be loaded and cached. When the component invokes the import, the module should load from cache ... *based on what I've heard anyway* 😄.

I'll email Steve and ask him a few questions on these points.

gregoryagu:
Sorry for the late reply, I am just getting to my emails.

I did a little repro to verify.  Here is what is shows:

If you use`this.JSModule = await JS.InvokeAsync<IJSObjectReference>("import", "./Components/Pages/Counter.razor.js");` then there is no need to add a <script> tag in app.razor.

If you do add a script tag, then the file is downloaded twice, once when the app starts, and then once when you navigate to the page.

Thus there is no need to add the script tag unless you purposely wanted it loaded up front. 

Here is the Console window in the browser that shows it was downloaded and executed twice:
image

However, as you point out, if you add type="module" then it is only loaded the first time.

`<script type="module" src="/Components/Pages/Counter.razor.js"></script>`

This is the [Repository](https://github.com/gregoryagu/BlazorTestImportingScript/tree/main)

gregoryagu:
Here is how I would write it:

Don't place a `<script>` tag for `JsCollocation2.razor.js` after the [Blazor script](xref:blazor/project-structure#location-of-the-blazor-script) because the module is loaded and cached automatically. Adding a `<script>` tag causes the module to be loaded twice unless `type="module` is added to the script tag. Then it is only loaded from the script.

guardrex:
Thanks for the additional investigation.

> if you add type="module" then it is only loaded the first time

Good! My hunch was correct 😄, and that moves the latency to app startup, which most people probably wouldn't want.

> Adding a `<script>` tag causes the module to be loaded twice unless `type="module"` is added to the script tag. Then it is only loaded from the script.

I'll add something along those lines. I'd like to mention the bit about front-loading the latency in case someone really wants to do that, along the lines of ...

> ... loaded from the `<script>` tag at app startup, moving the module loading latency from the component to app startup.

... something like that. I'll play with the wording on Monday morning and see what composes well.

In the meantime, I'm still waiting on Steve to get back to me. We'll see what he says, especially on my first question confirming that an `import` in this context is a ***dynamic*** `import()`. I'm almost certain that I'm correct on that point. That will confirm that I have the correct MDN cross-link and link text.

*Have a great weekend!* We'll push this forward on Monday.

gregoryagu:
Have a great weekend too!  (And if you ever want to hire a documentation reviewer/tester, let me know!)

guardrex:
Here's where we're at with this. We'll only carry the remark warning devs off of placing a `<script>` tag. We won't get into adding a `<script>` tag with `type="module"` for latency at startup. If devs want to do that, there's nothing stopping them AFAIK. However, that's not a best/typical use case scenario for loading JS module code using the collocation approach. I'll take more feedback on it going forward. If a few more devs open issues asking about it in the coming months/years, then we probably will add something on that aspect.

Thanks for the issue! I'm glad we're improving the coverage to dispel any question about it. *Happy New Year!* 🎉

PR #34479, opened 2025-01-13
Make QuickGrid section "Display name support" use correct way to determine display name

georg-jung:
See <https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.displayattribute.name?view=net-8.0#remarks> which reads

> Do not use this property to get the value of the [Name](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.displayattribute.name?view=net-8.0) property. Use the [GetName](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.displayattribute.getname?view=net-8.0) method instead.

I think using `.Name` instead of `.GetName()` would return the resource key / non-localized version.

Note: I didn't verify the fix in a minimal project.

@guardrex 

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/quickgrid.md](https://github.com/dotnet/AspNetCore.Docs/blob/c3d023c3a3209672a38d1b2990efc873cdb5ab71/aspnetcore/blazor/components/quickgrid.md) | [aspnetcore/blazor/components/quickgrid](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/quickgrid?branch=pr-en-us-34479) |

### DIFF
--- a/aspnetcore/blazor/components/quickgrid.md
+++ b/aspnetcore/blazor/components/quickgrid.md
@@ -297,7 +297,7 @@ However, managing column titles (names) from bound model properties is usually a
 public DateTime ReleaseDate { get; set; }
 ```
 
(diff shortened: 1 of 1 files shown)


guardrex (approved):
Fantastic idea, @georg-jung! 🚀 Thanks for the PR!

guardrex:
I'm going to go ahead and merge this, and I'll open a follow-up issue to further investigate localization for the properties here. I think a little more work will be required on this bit of guidance.

guardrex:
Actually, no ... I don't think more work will be required. The section shows the `Display` approach first, shows the code using `DisplayNameAttribute` before `Display`, shows the `DisplayNameAttribute` approach, and then remarks ...

> However, the [Display] attribute is recommended because it makes additional properties available. For example, the [Display] attribute offers the ability to assign a resource type for localization.

So, the guidance is kind'a pushing devs to use `Display` first, and will capture the localized name with `GetName`.

I think the section will fine, and I'll take further feedback from the community.

Thanks again! *Happy New Year!* 🎉

PR #34518, opened 2025-01-18
[Blazor] Script in HeadContent vs. enhanced navigation

hakenr:
I think this isn’t about interactive vs. static rendering, but rather depends on whether the page instance is preserved. If the DOM is being patched (interactive or static with enhanced navigation), the instance is retained. However, if a new page instance is created (static SSR without enhanced navigation), it behaves differently.


guardrex:
Thanks, @hakenr ... This came from Sanderson earlier this week, and he didn't mention the additional qualifier in discussion. Let's get past the MLK holiday weekend, and I'll message him to take a look next week.

The qualifier that's changing for the "removes the `<script>` tag from the rendered `<head>` content but doesn't unload the script's JavaScript code" remark is ...

```diff
- a Blazor Web App using an interactive render mode (interactive SSR, CSR)
+ a Blazor Web App using either an interactive render mode (interactive SSR, CSR) or static SSR with enhanced navigation
```

PR #34523, opened 2025-01-20
chore: remove redundant words

sunxunle:
remove redundant words

### DIFF
--- a/aspnetcore/client-side/dotnet-interop/index.md
+++ b/aspnetcore/client-side/dotnet-interop/index.md
@@ -131,7 +131,7 @@ Some types don't have a default type mapping. For example, a `long` can be marsh
 
 The following type mapping scenarios are supported:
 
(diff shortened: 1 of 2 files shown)


sunxunle:
@dotnet-policy-service agree

guardrex (approved):
Thanks @sunxunle! 🚀 ... It might seem strange that this happens here, but we know why. When the text of an article is written and the double-words break across two lines, the human eye 👁️ and brain 🧠 gets tripped up and misses them. About once a year, I run a Regex expression on the entire docs repo to flush these 😈 out. Thanks for catching these two!

sunxunle:
> Thanks @sunxunle! 🚀 ... It might seem strange that this happens here, but we know why. When the text of an article is written and the double-words break across two lines, the human eye 👁️ and brain 🧠 gets tripped up and misses them. About once a year, I run a Regex expression on the entire docs repo to flush these 😈 out. Thanks for catching these two!

It is my pleasure.

PR #34678, opened 2025-02-08
fix: typos in documentation files

maximevtush:


guardrex:
Hello @maximevtush ... Could you submit these changes directly to jQuery because we source these scripts from them and the mistakes are in their files. For example ...

https://cdn.jsdelivr.net/npm/jquery-validation@1.19.5/dist/jquery.validate.js

PR #34734, opened 2025-02-13
Failure to preserve types enhancements

### LINKED ISSUE #34654: Update complex type trimming remarks
### Description

To resolve https://github.com/dotnet/AspNetCore.Docs/pull/31422, language was added to two articles to cover how some complex types might be trimmed away.

In the *Trimmer* article ...

> Trimming may have detrimental effects for the published app. In apps that use \[reflection](/dotnet/csharp/advanced-topics/reflection-and-attributes/), the IL Trimmer often can't determine the required types for runtime reflection and trim them away. For example, complex framework types for JS interop, such as \<xref:System.Collections.Generic.KeyValuePair>, might be trimmed by default and not available at runtime for JS interop calls. In these cases, we recommend creating your own custom types instead.

In the *JS interop overview* article ...

> Complex framework types, such as \<xref:System.Collections.Generic.KeyValuePair>, might be \[trimmed away by the IL Trimmer on publish](xref:blazor/host-and-deploy/configure-trimmer) and not present for JS interop. We recommend creating custom types for types that the IL Trimmer trims away by default.

@maraf and @lewing ... Two questions ...

1. Is there a complex type that I can swap in for `KeyValuePair` that would still be trimmed away?

2. https://github.com/dotnet/runtime/pull/112216 is for release 8.0+, correct?

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/configure-trimmer?view=aspnetcore-9.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/host-and-deploy/configure-trimmer.md

### Document ID

1487b5e7-a40c-d757-c530-d0e4017f189e

### Article author

@guardrex

### Metadata

* ID: 2203b81b-1b37-8599-eb50-f6284170d224 
* Service: **aspnet-core**
* Sub-service: **blazor**

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+1487b5e7-a40c-d757-c530-d0e4017f189e)

guardrex:
Fixes #34654

Pavel ... Could you take a look at this one?

Two of the approaches work as I describe.

One approach, the Root Descriptor approach, I can't get to work using several permutations of assembly, type, and method ... assembly only ... assembly and method only ... or all three. If it won't be easy to figure out why this approach isn't working, should we merge with just the `[DynamicDependency]` attribute approach and the custom type approach? If so, I'll clip the Root Descriptor approach out of this, and I'll make a tracking item entry to try again to cover it some other time.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/host-and-deploy/configure-trimmer.md](https://github.com/dotnet/AspNetCore.Docs/blob/fa5b76550f4f933579ac2113ad04d2e72d01b16b/aspnetcore/blazor/host-and-deploy/configure-trimmer.md) | [aspnetcore/blazor/host-and-deploy/configure-trimmer](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/configure-trimmer?branch=pr-en-us-34734) |
| [aspnetcore/blazor/javascript-interoperability/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/fa5b76550f4f933579ac2113ad04d2e72d01b16b/aspnetcore/blazor/javascript-interoperability/index.md) | [aspnetcore/blazor/javascript-interoperability/index](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/javascript-interoperability/index?branch=pr-en-us-34734) |


pavelsavara:
You can find many examples of `ILLink.Descriptors.xml` in the runtime repo.
It may make sense to have full path via `$(MSBuildThisFileDirectory)` in the `<TrimmerRootDescriptor Include="$(MSBuildThisFileDirectory)ILLink.Descriptors.xml" />`

To see if your file was noticed or not, you can make it invalid xml and you should get fail. If nothing happened, somehow your TrimmerRootDescriptor didn' work.

I don't know if `TrimmerRootDescriptor` in MSbuild also works well for libraries and transitive dependencies. Worth exploring.

Trimmer team prefers attributes over XML, because they don't have those problems.

The last option of using custom type instead is probably poor workaround, do we know why we say that ?
It would only work if the type was in assembly which doesn't have trimming enabled, which I guess is a (bad) default for app assemblies.
If we really want to propose that, I would treat it as very last resort.

guardrex:
@pavelsavara ... No luck here this morning.

Let's take these one at a time ...

### Breaking the XML

The first thing is breaking the XML to get an error to know if the file is being found. I took the opening `<linker>` tag out of the file; and either with or without `$(MSBuildThisFileDirectory)`, I don't see an error anywhere at compile time or runtime. The app publishes normally AFAICT. Where would the error appear?

### Method signature

After we confirm that my file is being picked up, we can move to the next phase ... the method signature. I'm not 100% sure what it should be. Nothing I've tried has worked. However, we don't even know if my file is being picked up yet.

### Custom type

> The last option of using custom type instead is probably poor workaround, do we know why we say that ?

It came from Javier here ...

https://github.com/dotnet/aspnetcore/issues/52947#issuecomment-1886346182

I'm going to remove that approach on the next commit. I see what you mean about the attribute, and the attribute is the approach thus far that works well here.

pavelsavara:
> I don't see an error anywhere at compile time 

It means it's not being noticed by the msbuild scripts.

Do you have it in `<ItemGroup> </ItemGroup>` ?

guardrex:
... with a broken file ...

```xml

  <assembly fullname="System.Runtime">
    <type fullname="System.Collections.Generic.KeyValuePair">
      <method signature="System.Void .ctor(System.String,System.String)" />
    </type>
  </assembly>
</linker>
```

pavelsavara:
> If you want, I can put the project up in a GH repo.

yes

guardrex:
# 🐀🍔 ***Roasted Rat Burgers!***  🐀🍔

Ok ... I just made a major discovery here. The project that I'm using is a per-component BWA. I had the descriptor file ***set up in the server project*** 🙈 and just discovered that it has to be in the `.Client` project.

Let me try again now that I know what I was doing wrong. I'll get back to you in a bit .....................

guardrex:
Yeah ... that seems ok now. I did get the broken XML file warning, fixed the file, and now it seems to be picking it up.

Next quesiton tho ... what is the correct method signature for the `KeyValuePair` example to make it work?

It isn't ...

```xml
<method signature="System.Void .ctor(System.String,System.String)" />
```

pavelsavara:
```
    <type fullname="System.Collections.Generic.KeyValuePair`2">
        <method signature="System.Void .ctor()" />
        <method signature="System.Void .ctor(System.String,System.String)" />
    </type>
```

Try which of them works, with or without params.

pavelsavara:
`<method signature="System.Void .ctor(TKey,TValue)" />` ?

pavelsavara:
> ConstructorContainsNullParameterNames

Why this is not working for you https://github.com/dotnet/runtime/pull/102850
You are testing Net9 right ?

Maybe for Net8 and older @javiercn 's advice to use custom type is necessary fallback.

guardrex:
Correct ... 9.0 app.

Do you want me to place the custom type back for <=8.0, or should this only have the `[DynamicDependency]` attribute approach for <=8.0?

guardrex:
... so the fix is really going to be for .NET 10? ... I say so because the original issue is still open at ...

https://github.com/dotnet/aspnetcore/issues/52947

... marked for .NET 10.

If so, then we only have the `[DynamicDependency]` attribute approach and the custom type approach, and those two are valid for all versions. Then, we'll add this file descriptor approach for .NET 10 or later?

guardrex:
I see.

Should I put my test app up just so that you can check and make sure I have everything correct?

You say a new issue, but https://github.com/dotnet/aspnetcore/issues/52947 is still open. Seems like they just wouldn't close that one and keep working on it.

guardrex:
I've noted in the article text using my tracking moniker ("`UPDATE 10.0`") to circle back around later for more work to bring back the file descriptor approach and drop the custom types approach.

pavelsavara:
Does the `DynamicDependency` solve the `ConstructorContainsNullParameterNames` issue ?

guardrex:
Yes, it does.

aspnetcore/blazor/host-and-deploy/configure-trimmer.md
```diff
@@ -42,6 +40,141 @@ The default trimmer granularity for Blazor apps is `partial`. To trim all assemb
...
+> **:::no-loc text="key 1":::**  
+> :::no-loc text="value 1":::  
+> **:::no-loc text="key 2":::**  
+> :::no-loc text="value 2":::
+
+When the app is published, <xref:System.Collections.Generic.KeyValuePair> is trimmed from the app, even in spite of setting the [`<PublishTrimmed>` property](#configuration) to `false` in the project file. Accessing the component throws the following exception:
+
+> :::no-loc text="Unhandled exception rendering component: ConstructorContainsNullParameterNames, System.Collections.Generic.KeyValuePair`2[System.String,System.String]":::
+
+To address lost types, consider the following approaches.
```
pavelsavara, line 92: ~~approaches~~ -> solution

In singular until we un-comment the rest

pavelsavara (approved):

aspnetcore/blazor/host-and-deploy/configure-trimmer.md
guardrex: There are two ... the dynamic dep attribute and the custom type.

guardrex: The custom type is currently written out with the example. The commented text for it will replace that when the descriptor approach comes back. Either way, this sentence will always indicate more than one approach.

pavelsavara: ok

PR #34779, opened 2025-02-21
Cascading values+params for state management

### LINKED ISSUE #34236: Add a section on cascading values/parameters for state management with multiple subscribers

guardrex:
Fixes #34236

Thanks @hakenr! 🎷 ... This is probably a bit rough in spots. If you have time to look, I'll address your feedback before pinging a PU engineer. If you're busy, no worries ... I'll get one of them on here early next week to put an 👁️ on this.

**UPDATE (Monday, 2/24)**: I pinged to ask if either Illona or Ondrej are available to take a look. If they're busy, I'll hit up either Pavel or Marek to look.

### Notes

* Add a `CascadingValueSource<T>` with notifications (`INotifyPropertyChanged`/`NotifyChangedAsync`) approach to the *Cascading values and parameters* article. The basis for the example is code provided by @kzu on https://github.com/dotnet/aspnetcore/issues/53257 (***Thanks @kzu!*** 🎸) . Sanderson said it's supported at https://github.com/dotnet/aspnetcore/issues/53257#issuecomment-1904282397, but he also commented elsewhere that "In practice it's extremely rare for people to want to do `INotifyPropertyChanged` with Blazor." However, that might be because it isn't documented (until now ***here*** 😄).

* I favor placing this coverage because it's easy to understand/implement and flexible/powerful with a class type. I demo both setting a property value directly and having a method in the class that updates a property. An app can process state quite easily using this approach. I make sure to call out that there's a perf hit and to be careful with this approach.

* Add a short section to the *State Management* article mentioning cascading values and parameters generally. For >=8.0, cross-link the root-level cascading value/`CascadingValueSource<T>` /`INotifyPropertyChanged`/`NotifyChangedAsync` approach.

* WRT ... 

  > the entire State Management page could simply be a list of approaches with links to other pages that explain each feature in detail

  My initial concern is simply that explaining the features and explaining state management using the features *in one spot* is problematic because state management is a more advanced use case for the two or three scenarios here where this idea applies. For root-level cascading values with notifications, this approach seems to work well because the use case is a state management approach. I've made a tracking item entry for this idea to look at it again later.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/cascading-values-and-parameters.md](https://github.com/dotnet/AspNetCore.Docs/blob/2d49bcb6a90a84eb8f3f0425cdb9cdafb7c4a445/aspnetcore/blazor/components/cascading-values-and-parameters.md) | [aspnetcore/blazor/components/cascading-values-and-parameters](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/cascading-values-and-parameters?branch=pr-en-us-34779) |
| [aspnetcore/blazor/state-management.md](https://github.com/dotnet/AspNetCore.Docs/blob/2d49bcb6a90a84eb8f3f0425cdb9cdafb7c4a445/aspnetcore/blazor/state-management.md) | [aspnetcore/blazor/state-management](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/state-management?branch=pr-en-us-34779) |

### DIFF
--- a/aspnetcore/blazor/components/cascading-values-and-parameters.md
+++ b/aspnetcore/blazor/components/cascading-values-and-parameters.md
@@ -195,8 +195,7 @@ The type's <xref:System.ComponentModel.PropertyChangedEventHandler> (`HandleProp
 In the `Program` file&dagger;, `NotifyingDalek` is passed to create a <xref:Microsoft.AspNetCore.Components.CascadingValueSource%601> with an initial `Unit` value of 888 units:
 
 ```csharp
-builder.Services.AddNotifyingCascadingValue<NotifyingDalek>(
-    new NotifyingDalek() { Units = 888 });
+builder.Services.AddNotifyingCascadingValue(new NotifyingDalek() { Units = 888 });
 ```
 
 > [!NOTE]
--- a/aspnetcore/blazor/components/cascading-values-and-parameters.md
+++ b/aspnetcore/blazor/components/cascading-values-and-parameters.md
@@ -98,7 +98,7 @@ Calling <xref:Microsoft.AspNetCore.Components.CascadingValueSource%601.NotifyCha
 
 In the following example:
 
-* `NotifyingDalek` implements <xref:System.ComponentModel.INotifyPropertyChanged> to notify clients that a property value has changed. When the `Units` property is set, the <xref:System.ComponentModel.PropertyChangedEventHandler> (`PropertyChanged`) is invoked.
+* `NotifyingDalek` implements <xref:System.ComponentModel.INotifyPropertyChanged> to notify clients that a property value has changed. When the `Units` property is set, the <xref:System.ComponentModel.PropertyChangedEventHandler> (`PropertyChanged`) is invoked. Keep in mind for production code that any change in state (any property value change of the class) causes all subscribed components to rerender, regardless of which part of the state they use. We recommend creating granular classes, cascading them separately with specific subscriptions to ensure that only components subscribed to a specific portion of the application state are affected by changes.
 * The `SetUnitsToOneThousand` method can be triggered by subscribers to set `Units` to 1,000 with a simulated processing delay.
 
 `NotifyingDalek.cs`:
--- a/aspnetcore/blazor/components/cascading-values-and-parameters.md
+++ b/aspnetcore/blazor/components/cascading-values-and-parameters.md
@@ -170,7 +170,7 @@ public class NotifyingCascadingValueSource<T> : IDisposable where T : INotifyPro
 
     private void OnValuePropertyChanged(object? sender, PropertyChangedEventArgs e)
     {
-        source.NotifyChangedAsync();
+        _ = source.NotifyChangedAsync();
     }
 
     public CascadingValueSource<T> Source => source;
--- a/aspnetcore/blazor/components/cascading-values-and-parameters.md
+++ b/aspnetcore/blazor/components/cascading-values-and-parameters.md
@@ -75,7 +75,7 @@ The following `Daleks` component displays the cascaded values.
 
 :::moniker range=">= aspnetcore-8.0"
 
-In the following example, `Dalek` is registered as a cascading value using [`CascadingValueSource<T>`](xref:Microsoft.AspNetCore.Components.CascadingValueSource%601), where `<T>` is the type. The `isFixed` flag indicates whether the value is fixed. If false, all recipients are subscribed for update notifications, which are issued by calling <xref:Microsoft.AspNetCore.Components.CascadingValueSource%601.NotifyChangedAsync%2A>. Subscriptions create overhead and reduce performance, so set `isFixed` to `true` if the value doesn't change.
+In the following example, `Dalek` is registered as a cascading value using [`CascadingValueSource<T>`](xref:Microsoft.AspNetCore.Components.CascadingValueSource%601), where `<T>` is the type. The `isFixed` flag indicates whether the value is fixed. If false, all recipients are subscribed for update notifications. Subscriptions create overhead and reduce performance, so set `isFixed` to `true` if the value doesn't change.
 
 ```csharp
 builder.Services.AddCascadingValue(sp =>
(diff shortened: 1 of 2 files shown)


guardrex:
Illona sent a better factory that unhooks the event handler. Unfortunately, what she sent isn't running correctly yet. I'll bang the keyboard for a bit 🐒 and see if I can get it running.

guardrex:
🎉 ***Better!*** ... This works ..............

```csharp
using System.ComponentModel;
using Microsoft.AspNetCore.Components;

namespace Microsoft.Extensions.DependencyInjection;

public static class CascadingAuthenticationStateServiceCollectionExtensions
{
    public static IServiceCollection AddCascadingApplicationState<T>(
        this IServiceCollection serviceCollection, T value, bool isFixed = false)
        where T : INotifyPropertyChanged
    {
        return serviceCollection.AddCascadingValue<T>(services =>
        {
            return new ApplicationStateCascadingValueSource<T>(value, isFixed);
        });
}

    private sealed class ApplicationStateCascadingValueSource<T>
        : CascadingValueSource<T>, IDisposable where T : INotifyPropertyChanged
    {
        private readonly T value;
        private readonly CascadingValueSource<T> source;

        public ApplicationStateCascadingValueSource(T value, bool isFixed = false)
            : base(value, isFixed = false)
        {
            this.value = value;
            source = new CascadingValueSource<T>(value, isFixed);
            this.value.PropertyChanged += HandleApplicationStateChanged;
        }

        private void HandleApplicationStateChanged(object? sender, PropertyChangedEventArgs e)
        {
            _ = NotifyChangedAsync();
        }

        public CascadingValueSource<T> Source => source;

        public void Dispose()
        {
            value.PropertyChanged -= HandleApplicationStateChanged;
        }
    }
}
```

... with ...

```csharp
builder.Services.AddCascadingApplicationState<NotifyingDalek>(new NotifyingDalek() { Units = 888 });
```

guardrex:
I put it on the PR and will revise further on Tuesday morning.

hakenr:
@guardrex I like the samples. The only thing that worries me is the name of the `AddCascadingApplicationState<T>()` method, which suggests the possibility of using it as a large state storage (aka _store_) with many rich properties representing different parts of the global application state.  

What we need to emphasize is that **ANY** change in such state (**ANY** property) will cause **ALL** subscribed components to re-render, regardless of which part of the state they use. So, we should guide users to avoid creating a single large class representing the entire global application state. Instead, they should create more granular classes and cascade them separately (to allow more specific subscriptions with cascading parameters), ensuring that only components subscribed to a specific portion of the application state are affected by changes.

guardrex:
*Totally agree!* I'll probably even ***STEAL*** your language 👮‍♂️🚓😄 when I make changes tomorrow to react to that in the code examples. The thing that I was just trying to accomplish today was creating a better approach with unhooking the handler and making it a service collection extension. *So far, so good!*

ilonatommy (approved):
Looking good!

guardrex:
Illona suggested naming ...

* `AddCascadingStateNotifier`
* `AddCascadingStateSubscription`
* `AddCascadingScopedState`
* `AddCascadingState`

I'll be working further on this PR shortly.

guardrex:
Made the updates.

@hakenr ... I placed your remarks in ***two*** spots for the greatest visibility ...

* Where `NotifyingDalek` is described. The remarks directly pertain to the class.
* At the very end of the section. It's an important concept that we'd like to leave them with at the end to reinforce the point.

The code updates are verifying working here; so unless there are additional change requests (e.g., naming in the code), I'm ready to merge this.

aspnetcore/blazor/components/cascading-values-and-parameters.md
```diff
@@ -75,7 +75,7 @@ The following `Daleks` component displays the cascaded values.
 
 :::moniker range=">= aspnetcore-8.0"
 
-In the following example, `Dalek` is registered as a cascading value using [`CascadingValueSource<T>`](xref:Microsoft.AspNetCore.Components.CascadingValueSource%601), where `<T>` is the type. The `isFixed` flag indicates whether the value is fixed. If false, all recipients are subscribed for update notifications, which are issued by calling <xref:Microsoft.AspNetCore.Components.CascadingValueSource%601.NotifyChangedAsync%2A>. Subscriptions create overhead and reduce performance, so set `isFixed` to `true` if the value doesn't change.
+In the following example, `Dalek` is registered as a cascading value using [`CascadingValueSource<T>`](xref:Microsoft.AspNetCore.Components.CascadingValueSource%601), where `<T>` is the type. The `isFixed` flag indicates whether the value is fixed. If false, all recipients are subscribed for update notifications. Subscriptions create overhead and reduce performance, so set `isFixed` to `true` if the value doesn't change.
```
hakenr, line 78: `false` should be styles as code

aspnetcore/blazor/components/cascading-values-and-parameters.md
```diff
@@ -94,6 +94,192 @@ builder.Services.AddCascadingValue(sp =>
...
+                OnPropertyChanged();
+            }
+        }
+    }
+
+    protected virtual void OnPropertyChanged(
+        [CallerMemberName] string? propertyName = default)
+            => PropertyChanged?.Invoke(this, new(propertyName));
+
+    public async Task SetUnitsToOneThousand()
```
hakenr, line 134: The method name should have `Async` suffix, `SetUnitsToOneThousandAsync()`.

aspnetcore/blazor/components/cascading-values-and-parameters.md
```diff
@@ -94,6 +94,192 @@ builder.Services.AddCascadingValue(sp =>
...
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+public static class CascadingStateServiceCollectionExtensions
+{
+    public static IServiceCollection AddCascadingStateNotifier<T>(
+        this IServiceCollection serviceCollection, T value, bool isFixed = false)
+        where T : INotifyPropertyChanged
+    {
+        return serviceCollection.AddCascadingValue<T>(services =>
```
hakenr, line 160: This should not be `services` but `serviceProvider` or `sp`.

aspnetcore/blazor/components/cascading-values-and-parameters.md
```diff
@@ -94,6 +94,192 @@ builder.Services.AddCascadingValue(sp =>
...
+```csharp
+using System.ComponentModel;
+using Microsoft.AspNetCore.Components;
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+public static class CascadingStateServiceCollectionExtensions
+{
+    public static IServiceCollection AddCascadingStateNotifier<T>(
+        this IServiceCollection serviceCollection, T value, bool isFixed = false)
```
hakenr, line 157: The usual name being used here is `services` instead of `serviceCollection`.
I also suggest to use `state` instead of `value`.

aspnetcore/blazor/components/cascading-values-and-parameters.md
```diff
@@ -94,6 +94,192 @@ builder.Services.AddCascadingValue(sp =>
...
+        return serviceCollection.AddCascadingValue<T>(services =>
+        {
+            return new CascadingStateValueSource<T>(value, isFixed);
+        });
+    }
+
+    private sealed class CascadingStateValueSource<T>
+        : CascadingValueSource<T>, IDisposable where T : INotifyPropertyChanged
+    {
+        private readonly T value;
```
hakenr, line 169: Again, `state` might work better than `value`.

aspnetcore/blazor/components/cascading-values-and-parameters.md
```diff
@@ -94,6 +94,192 @@ builder.Services.AddCascadingValue(sp =>
...
+
+```csharp
+using System.ComponentModel;
+using Microsoft.AspNetCore.Components;
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+public static class CascadingStateServiceCollectionExtensions
+{
+    public static IServiceCollection AddCascadingStateNotifier<T>(
```
hakenr, line 156: I kind of don't like the `AddCascadingStateNotifier` name. It suggests that we're registering something that should have `Notifier` in its type name or interface, which it doesn't.  

Maybe something like:  
- `AddNotifyingCascadingValue()`  
- `AddNotifyPropertyChangedCascadingValue()`  

The issue isn't with the definition but with how it's used:  

```csharp
builder.Services.AddCascadingStateNotifier<NotifyingDalek>(new NotifyingDalek() { Units = 888 });
```  

The registration should instead look something like this, leveraging generic type inference:  

```csharp
builder.Services.AddNotifyingCascadingValue(new NotifyingDalek() { Units = 888 });
```
...we are actually registering a _cascading value_, which should be clear.

aspnetcore/blazor/components/cascading-values-and-parameters.md
```diff
@@ -94,6 +94,192 @@ builder.Services.AddCascadingValue(sp =>
...
+<Daleks />
+
+<Daleks />
+
+<Daleks />
+```
+
+Because the <xref:Microsoft.AspNetCore.Components.CascadingValueSource%601>'s type in this example (`NotifyingDalek`) is a class type, you can meet virtually any state management feature specification requirement. However, subscriptions create overhead and reduce performance, so benchmark the performance of this approach in your app and compare it to other [state management approaches](xref:blazor/state-management) before adopting it in a production app with constrained processing and memory resources.
+
+Any change in state (any property value change of the class) causes all subscribed components to rerender, regardless of which part of the state that they use. **Avoid creating a single large class representing the entire global application state.** Instead, create granular classes and cascade them separately with specific subscriptions to cascading parameters, ensuring that only components subscribed to a specific portion of the application state are affected by changes.
```
hakenr, line 281: regardless of which part of the state ~that~ they use

aspnetcore/blazor/components/cascading-values-and-parameters.md
guardrex:
```suggestion
    public async Task SetUnitsToOneThousandAsync()
```

aspnetcore/blazor/components/cascading-values-and-parameters.md
guardrex, line 157:
```suggestion
        this IServiceCollection services, T state, bool isFixed = false)
```

aspnetcore/blazor/components/cascading-values-and-parameters.md
guardrex, line 160:
```suggestion
        return serviceCollection.AddCascadingValue<T>(sp =>
```

aspnetcore/blazor/components/cascading-values-and-parameters.md
guardrex, line 169:
```suggestion
        private readonly T state;
```

aspnetcore/blazor/components/cascading-values-and-parameters.md
```diff
@@ -94,6 +94,192 @@ builder.Services.AddCascadingValue(sp =>
...
+        });
+    }
+
+    private sealed class CascadingStateValueSource<T>
+        : CascadingValueSource<T>, IDisposable where T : INotifyPropertyChanged
+    {
+        private readonly T value;
+        private readonly CascadingValueSource<T> source;
+
+        public CascadingStateValueSource(T value, bool isFixed = false)
```
guardrex, line 172:
```suggestion
        public CascadingStateValueSource(T state, bool isFixed = false)
```

aspnetcore/blazor/components/cascading-values-and-parameters.md
```diff
@@ -94,6 +94,192 @@ builder.Services.AddCascadingValue(sp =>
...
+            : base(value, isFixed = false)
+        {
+            this.value = value;
+            source = new CascadingValueSource<T>(value, isFixed);
```
guardrex, line 176:
```suggestion
            source = new CascadingValueSource<T>(state, isFixed);
```

aspnetcore/blazor/components/cascading-values-and-parameters.md
```diff
@@ -94,6 +94,192 @@ builder.Services.AddCascadingValue(sp =>
...
+            this.value.PropertyChanged += HandlePropertyChanged;
```
guardrex, line 177:
```suggestion
            this.state.PropertyChanged += HandlePropertyChanged;
```

aspnetcore/blazor/components/cascading-values-and-parameters.md
```diff
@@ -94,6 +94,192 @@ builder.Services.AddCascadingValue(sp =>
...
+        }
+
+        private void HandlePropertyChanged(object? sender, PropertyChangedEventArgs e)
+        {
+            _ = NotifyChangedAsync();
+        }
+
+        public void Dispose()
+        {
+            value.PropertyChanged -= HandlePropertyChanged;
```
guardrex, line 187:
```suggestion
            state.PropertyChanged -= HandlePropertyChanged;
```

aspnetcore/blazor/components/cascading-values-and-parameters.md
guardrex, line 281:
```suggestion
Any change in state (any property value change of the class) causes all subscribed components to rerender, regardless of which part of the state they use. **Avoid creating a single large class representing the entire global application state.** Instead, create granular classes and cascade them separately with specific subscriptions to cascading parameters, ensuring that only components subscribed to a specific portion of the application state are affected by changes.
```

aspnetcore/blazor/components/cascading-values-and-parameters.md
guardrex: Ok ... I'll fix this separately after all of the other changes go in.

guardrex: ... and I'm mopping up the last few changes locally that I missed with the suggestion commit.

hakenr:
I think the method definition should remain as is:  

```csharp
public static IServiceCollection AddNotifyingCascadingValue<T>(
    this IServiceCollection services, T state, bool isFixed = false)
    where T : INotifyPropertyChanged
{
    return services.AddCascadingValue(sp =>
    {
        return new CascadingStateValueSource<T>(state, isFixed);
    });
}
```  

I just believe (and I haven't tested it 😇) that these two lines are equivalent, with the generic type inferred by the C# compiler:  

```csharp
builder.Services.AddNotifyingCascadingValue(new NotifyingDalek() { Units = 888 });
builder.Services.AddNotifyingCascadingValue<NotifyingDalek>(new NotifyingDalek() { Units = 888 });
```  

You should be able to verify this by hovering over the shortened call in Visual Studio - it should show the full version in the tooltip.  

There must be some other issue if there's a "no-op at runtime."  
Do you want me to double-check the whole sample, or would you like to give it another try on your end?  

(BTW, I don't insist on the shortened call - it's just more readable. What worries me is: *How the hell can the shortened call cause a no-op?* There must be something else going on...)

guardrex:
> There must be some other issue if there's a "no-op at runtime."

Yes ... I think that there's something else going on in the framework. I checked it a few times, and it indeed no-ops without the type. See how the initial value isn't appearing ...

image

I see what you mean about hovering over it ...

image

It's nice to hear that my lack of knowledge was for something that ***doesn't exist in the first place.*** 😆 Indeed, this was the only why that I knew the method signature should be composed.

I guess we won't hold this up over the shortened syntax. I'm ready to go ahead and merge it.

guardrex:
***I resolved the problem!*** 🎉 It was missing a `T` type ...

```diff
- return services.AddCascadingValue<T>(sp =>
+ return services.AddCascadingValue(sp =>
```

After that, it's working again ***and with*** the shortened syntax ...

```csharp
builder.Services.AddNotifyingCascadingValue(new NotifyingDalek() { Units = 888 });
```

I'll merge this now!

Thanks @hakenr and @ilonatommy! 🍻

Issue #34775, opened 2025-02-20
Doesn't work with .net core 9.x


kemundos:
### Description

Going through the BlazorWebAppMovies tutorial.

The command below is used in the scaffolding part. However, when running this command, it's stated that you need .net core 8. I'm on .net core 9.

Seems the tutorial is broken as you cannot progress without this in a sensible way.

dotnet aspnet-codegenerator blazor CRUD -dbProvider sqlite -dc BlazorWebAppMovies.Data.BlazorWebAppMoviesContext -m Movie -outDir Components/Pages

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/tutorials/movie-database-app/part-2?view=aspnetcore-9.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/tutorials/movie-database-app/part-2.md

### Document ID

f9acc6f3-d6e2-ca66-9a66-7e0ea09aa38f

### Article author

@guardrex

### Metadata

* ID: f9acc6f3-d6e2-ca66-9a66-7e0ea09aa38f
* PlatformId: ef97ed5a-9570-fb55-a33b-46ed7e991772 
* Service: **aspnet-core**
* Sub-service: **blazor**

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+f9acc6f3-d6e2-ca66-9a66-7e0ea09aa38f)

guardrex:
Hello @kemundos ...

> it's stated that you need .net core 8

Very strange. I've stepped through the whole thing with .NET 9 when I was working up .NET 9 updates for the tutorial series.

Can you screenshot where that error is appearing and place it into a GH comment here?

guardrex:
Ok ... I see a problem. I followed the tutorial down to the `` command, and I'm getting ...

> Generating a new DbContext class 'BlazorWebAppMovies.Data.BlazorWebAppMoviesContext'
> Attempting to compile the application in memory with the added DbContext.
> Attempting to figure out the EntityFramework metadata for the model and DbContext: 'Movie'
> Could not load file or assembly 'Microsoft.EntityFrameworkCore.Design, Version=9.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.
>    at Microsoft.VisualStudio.Web.CodeGeneration.ActionInvoker.<BuildCommandLine>b__6_0()
[...]

PR #34814, opened 2025-02-26
Update EF Core tooling breaking change

guardrex:
Addresses #34775

Let's see if this looks better. I think with a minor update to this approach at 10.0 to remove the remark about .NET 10's release (and to update the link so that it just points to the downloads page at https://dotnet.microsoft.com/download), this is a better way to handle a tooling breaking change that will be patched for 10.0. ***Maybe!*** 🤔 ... best to sleep on it for sure 🛌💤.

cc: @Rick-Anderson @tdykstra ... Discussed further with Wade offline. I'm looking at a different approach here for this because this really is a temporary tooling (SDK) problem.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/tutorials/movie-database-app/part-2.md](https://github.com/dotnet/AspNetCore.Docs/blob/5a97559abdeffc48dc2d1e4c1e0f918d90286721/aspnetcore/blazor/tutorials/movie-database-app/part-2.md) | [aspnetcore/blazor/tutorials/movie-database-app/part-2](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/tutorials/movie-database-app/part-2?branch=pr-en-us-34814) |

### DIFF
--- a/aspnetcore/blazor/tutorials/movie-database-app/part-2.md
+++ b/aspnetcore/blazor/tutorials/movie-database-app/part-2.md
@@ -134,36 +134,32 @@ dotnet add package Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore
 > [!IMPORTANT]
 > After the first eight commands execute, make sure that you press <kbd>Enter</kbd> on the keyboard to execute the last command.
 
-:::moniker range=">= aspnetcore-9.0 < aspnetcore-10.0"
+:::moniker range=">= aspnetcore-9.0"
 
-<!-- UPDATE 10.0 - Update the first bullet. -->
+<!-- UPDATE 10.0 - Remove this when the 9.0.300 SDK lands. -->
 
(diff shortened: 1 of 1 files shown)


guardrex:
Yeah ... I do think I like this approach better.

I'll just need to swing back into this at 10.0 GA to update that first bullet to just say to use a .NET 10 or later SDK and cross-link to the main downloads page.

That will harden this up for the future (.NET 10 or later), and it has the nice side benefit of less text to maintain.

guardrex:
@wadepickett @Rick-Anderson @tdykstra (... get well Tom! 🤕)

They just told me that they're reverting what they did that causes the breaking change. This whole problem will go away at 9.0.3.

I'll revise this to that effect.

guardrex:
OK! That licks IT! 🐮👅

PR #34818, opened 2025-02-27
[Blazor] What's new .NET10 - QuickGrid.RowClass sample adjustment

hakenr:
I think the current sample is overloaded with placeholders, making it hard to read and deviating from the usual way we describe features in the docs (by example).


guardrex (approved):
Thanks @hakenr! 👍 ... I'm going to use a little bit of this theme with the embedded example in the article section in a minute.

PR #34827, opened 2025-02-27
Update mfa.md to fix broken reference

Arceden:
Link referenced an older version of of .NET. This update bumps it from 6.0 to 8.0.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/security/authentication/mfa.md](https://github.com/dotnet/AspNetCore.Docs/blob/c77b042dd9ec8356b07f0e5670ca36d63391e5dd/aspnetcore/security/authentication/mfa.md) | [aspnetcore/security/authentication/mfa](https://review.learn.microsoft.com/en-us/aspnet/core/security/authentication/mfa?branch=pr-en-us-34827) |


guardrex (approved):
Thanks @Arceden! 🚀 ... I modified it to harden it completely against future version changes (so long as they don't ***move the code*** 😆). We have a special INCLUDE that we can drop in that tells the reader to grab the correct version from the version selector ...

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/includes/aspnetcore-repo-ref-source-links.md

PR #34842, opened 2025-02-27
Reconnect UI component

### LINKED ISSUE #34813: New reconnect UI component in the BWA template
### Description

* PR: https://github.com/dotnet/aspnetcore/pull/60376
* Issue: https://github.com/dotnet/aspnetcore/issues/57453
* Issue: https://github.com/dotnet/aspnetcore/issues/58629

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/signalr?view=aspnetcore-9.0#reflect-the-server-side-connection-state-in-the-ui

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/fundamentals/signalr.md

### Document ID

d17e4c28-ea82-0dd8-295d-19f6ada147ea

### Article author

@guardrex

### Metadata

* PlatformId: 1f59f6df-d4db-5205-778e-570938f04f3f 
* Service: **aspnet-core**
* Sub-service: **blazor**

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+d17e4c28-ea82-0dd8-295d-19f6ada147ea)

guardrex:
Fixes #34813

## Notes

* I'm versioning all of the 10.0 or later content into one moniker range, and I'm disabling all of the existing content for <10.0. In the past, I broke up coverage into multiple moniker ranges because parts of the coverage changed (e.g., filenames, file locations).
* I'm using our standard approach for cross-linking reference source. I link the `main` branch, and I include the [guidance on selecting the correct release branch/tag](https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/includes/aspnetcore-repo-ref-source-links.md).

## ❓Questions❓ 

* Are the existing images still accurate? ***If not,*** I'll make a tracking note to snap and place new shots on Preview 2 release day because I'm not on nightly releases (*yet* anyway 😄).
  * https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/fundamentals/signalr/_static/reconnection-ui-90-or-later.png
  * https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/fundamentals/signalr/_static/retry-ui-90-or-later.png
* The rejected scenario is a bit confusing: What does the `ReconnectModel` component display in that scenario? There's no such class (element marked with) "`components-reconnect-rejected-visible`," and doesn't `components-reconnect-failed-visible` go to `display: none` on a rejection? I don't see any styles for `components-reconnect-rejected` when it's applied to the dialog.
* Are there any *gotchas* 😈 to call out?

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/signalr.md](https://github.com/dotnet/AspNetCore.Docs/blob/9091b6cc12f041a14a322760323e71acb11697af/aspnetcore/blazor/fundamentals/signalr.md) | [aspnetcore/blazor/fundamentals/signalr](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/signalr?branch=pr-en-us-34842) |

### DIFF
--- a/aspnetcore/blazor/fundamentals/signalr.md
+++ b/aspnetcore/blazor/fundamentals/signalr.md
@@ -524,7 +524,7 @@ If reconnection succeeds, user state is often lost. Custom code can be added to 
 To create UI elements that track reconnection state, the following table describes:
 
 * A set of `components-reconnect-*` CSS classes (**Css class** column) that are set or unset by Blazor on an element with an `id` of `components-reconnect-modal`.
-* The `components-reconnect-state-changed` event value (**Event** column) that indicates a reconnection status change.
+* A `components-reconnect-state-changed` event (**Event** column) that indicates a reconnection status change.
 
 | CSS class | Event | Indicates&hellip; |
 | --- | --- | --- |
@@ -541,6 +541,26 @@ When the reconnection state change is `rejected`, the server was reached but ref
 * A crash in the server-side circuit occurs.
 * The client is disconnected long enough for the server to drop the user's state. Instances of the user's components are disposed.
 * The server is restarted, or the app's worker process is recycled.
+
+The developer adds an event listener on the reconnect modal element to monitor and react to reconnection state changes, as seen in the following example:
+
+```javascript
+const reconnectModal = document.getElementById("components-reconnect-modal");
+reconnectModal.addEventListener("components-reconnect-state-changed", 
+  handleReconnectStateChanged);
+
+function handleReconnectStateChanged(event) {
+  if (event.detail.state === "show") {
+    reconnectModal.showModal();
+  } else if (event.detail.state === "hide") {
+    reconnectModal.close();
+  } else if (event.detail.state === "failed") {
+    Blazor.reconnect();
+  } else if (event.detail.state === "rejected") {
+    location.reload();
+  }
+}
+```
 
 An element with an `id` of `components-reconnect-max-retries` displays the maximum number of reconnect retries:
 
@@ -560,7 +580,7 @@ An element with an `id` of `components-seconds-to-next-attempt` displays the num
 <span id="components-seconds-to-next-attempt"></span>
 ```
 
-The Blazor Web App project template includes a `ReconnectModal` component (`Components/Layout/ReconnectModal.razor`) with collocated stylesheet and JavaScript files (`ReconnectModal.razor.css`, `ReconnectModal.razor.js`) that can be customized as needed. These files can be examined in the ASP.NET Core reference source or by inspecting an app created from the Blazor Web App project template with Interactive WebAssembly or Interactive Auto rendering enabled:
+The Blazor Web App project template includes a `ReconnectModal` component (`Components/Layout/ReconnectModal.razor`) with collocated stylesheet and JavaScript files (`ReconnectModal.razor.css`, `ReconnectModal.razor.js`) that can be customized as needed. These files can be examined in the ASP.NET Core reference source or by inspecting an app created from the Blazor Web App project template. The component is added to the project when the project is created in Visual Studio with **Interactive render mode** set to **Server** or **Auto** or created with the .NET CLI with the option `--interactivity server` (default) or `--interactivity auto`.
 
 * [`ReconnectModal` component](https://github.com/dotnet/aspnetcore/blob/main/src/ProjectTemplates/Web.ProjectTemplates/content/BlazorWeb-CSharp/BlazorWeb-CSharp/Components/Layout/ReconnectModal.razor)
 * [Stylesheet file](https://github.com/dotnet/aspnetcore/blob/main/src/ProjectTemplates/Web.ProjectTemplates/content/BlazorWeb-CSharp/BlazorWeb-CSharp/Components/Layout/ReconnectModal.razor.css)
--- a/aspnetcore/blazor/fundamentals/signalr.md
+++ b/aspnetcore/blazor/fundamentals/signalr.md
@@ -553,7 +553,7 @@ Additional CSS classes to further control the style of rendered elements is desc
 | --- | --- | --- |
 | `components-reconnect-first-attempt-visible` | `retrying` | The first retry attempt. |
 | `components-reconnect-repeated-attempt-visible` | `retrying` | Subsequent retry attempts. |
-| `components-reconnect-failed-visible` | `failed`/`rejected` | Reconnection attempts that have failed or been rejected. |
+| `components-reconnect-failed-visible` | `failed` | A failed reconnection attempt. |
 
 An element with an `id` of `components-reconnect-max-retries` displays the maximum number of reconnect retries:
 
--- a/aspnetcore/blazor/fundamentals/signalr.md
+++ b/aspnetcore/blazor/fundamentals/signalr.md
@@ -521,39 +521,26 @@ If reconnection succeeds, user state is often lost. Custom code can be added to 
 
 :::moniker range=">= aspnetcore-10.0"
 
-Customize the UI and reconnection behavior in the `ReconnectModal` component (`Components/Layout/ReconnectModal.razor`), its collocated stylesheet file (`ReconnectModal.razor.css`), and its collocated JavaScript file (`ReconnectModal.razor.js`). These files can be examined in the ASP.NET Core reference source or by inspecting an app created from the Blazor Web App project template with Interactive WebAssembly or Interactive Auto rendering enabled:
+To create UI elements that track reconnection state, the following table describes:
 
-* [`ReconnectModal` component](https://github.com/dotnet/aspnetcore/blob/main/src/ProjectTemplates/Web.ProjectTemplates/content/BlazorWeb-CSharp/BlazorWeb-CSharp/Components/Layout/ReconnectModal.razor)
-* [Stylesheet file](https://github.com/dotnet/aspnetcore/blob/main/src/ProjectTemplates/Web.ProjectTemplates/content/BlazorWeb-CSharp/BlazorWeb-CSharp/Components/Layout/ReconnectModal.razor.css)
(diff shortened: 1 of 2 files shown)


oroztocil:
Sorry for keeping you waiting for the review. I first want to investigate and fix https://github.com/dotnet/AspNetCore-ManualTests/issues/3489 because it might involve changing some things mentioned in the docs.

aspnetcore/blazor/fundamentals/signalr.md
```diff
@@ -519,7 +519,63 @@ If reconnection fails, the user is instructed to retry or reload the page:
...
+
+| CSS class | Event | Indicates&hellip; |
+| --- | --- | --- |
+| `components-reconnect-show` | `show` | A lost connection. The client is attempting to reconnect. Show the modal. |
+| `components-reconnect-hide` | `hide` | An active connection is re-established to the server. Hide the modal. |
+| `components-reconnect-retrying` | `retrying` | The client is attempting to reconnect. |
+| `components-reconnect-failed` | `failed` | Reconnection failed, probably due to a network failure. To attempt reconnection, call `Blazor.reconnect()` in JavaScript. |
+| `components-reconnect-rejected` | `rejected` | Reconnection rejected. The server was reached but refused the connection, and the user's state on the server is lost. To reload the app, call `location.reload()` in JavaScript. This connection state may result when:<ul><li>A crash in the server-side circuit occurs.</li><li>The client is disconnected long enough for the server to drop the user's state. Instances of the user's components are disposed.</li><li>The server is restarted, or the app's worker process is recycled.</li></ul> |
```
oroztocil, line 548: It might be better to put the extended explanations and the hints about calling `Blazor.reconnect()` and `location.reload()` to the previous section about the `components-reconnect-state-changed` event, as it is a lot more straightforward for the user's to call those functions in a handler for the JS event than as a reaction to CSS class change. The CSS classes can be used to set up CSS rules that control showing/hiding/changing appearance of the UI, not behavior itself.

aspnetcore/blazor/fundamentals/signalr.md
```diff
@@ -519,7 +519,63 @@ If reconnection fails, the user is instructed to retry or reload the page:
...
+
+Additional CSS classes to further control the style of rendered elements is described in the following table. The **Event** column represents the value of the matching `components-reconnect-state-changed` JavaScript event.
+
+| CSS class | Event | Displayed on&hellip; |
+| --- | --- | --- |
+| `components-reconnect-first-attempt-visible` | `retrying` | The first retry attempt. |
+| `components-reconnect-repeated-attempt-visible` | `retrying` | Subsequent retry attempts. |
+| `components-reconnect-failed-visible` | `failed` | A failed reconnection attempt. |
```
oroztocil, line 556: I would not document these classes as they are only used in the `ReconnectModal` component which is just a showcase as discussed in the other comment.

The previous set of `components-reconnect-*` classes which you describe above is what is set/usnet by the `blazor.web.js` library on the `components-reconnect-modal` element.

aspnetcore/blazor/fundamentals/signalr.md
```diff
@@ -519,7 +519,63 @@ If reconnection fails, the user is instructed to retry or reload the page:
 
 If reconnection succeeds, user state is often lost. Custom code can be added to any component to save and reload user state across connection failures. For more information, see <xref:blazor/state-management>.
 
-:::moniker range=">= aspnetcore-8.0"
+:::moniker range=">= aspnetcore-10.0"
+
+Customize the UI and reconnection behavior in the `ReconnectModal` component (`Components/Layout/ReconnectModal.razor`), its collocated stylesheet file (`ReconnectModal.razor.css`), and its collocated JavaScript file (`ReconnectModal.razor.js`). These files can be examined in the ASP.NET Core reference source or by inspecting an app created from the Blazor Web App project template with Interactive WebAssembly or Interactive Auto rendering enabled:
```
oroztocil, line 524: I am not completely sure how to phrase it but I would like to emphasize that the `ReconnectModal` component is just a showcase of how things _can be done_ and is not otherwise special. The _special_ thing is the ID of the `components-reconnect-modal` element, and the JS event and CSS classes which are published by the `blazor.web.js` library. (Which you document nicely later.)

aspnetcore/release-notes/aspnetcore-10/includes/blazor.md
```diff
@@ -36,4 +36,15 @@ The [`[Route]` attribute](xref:Microsoft.AspNetCore.Components.RouteAttribute) n
 
 Previously, <xref:Microsoft.AspNetCore.Components.NavigationManager.NavigateTo%2A?displayProperty=nameWithType> scrolled to the top of the page for same-page navigations. This behavior has been changed in .NET 10 so that the browser no longer scrolls to the top of the page when navigating to the same page. This means the viewport is no longer reset when making updates to the address for the current page, such as changing the query string or fragment.
 
+### Reconnection UI component added to the Blazor Web App project template
+
+The Blazor Web App project template now includes a `ReconnectModal` component, including collocated stylesheet and JavaScript files, for improved developer control over the reconnection UI when the client loses the WebSocket connection to the server. The component isn't required to inline styles programatically, so it doesn't violate stricter Content Security Policy (CSP) settings for the `style-src` policy, thus avoiding CSP violations that occurred with the reconnection UI in prior releases.
```
oroztocil, line 41: Rather than "inline styles" I would say "insert styles". Also maybe add note that the previous JS-based solution is still present as the default/fallback.

E.g.: "The component does not insert styles programmatically so it doesn't violate stricter Content Security Policy (CSP) settings for the `style-src` policy. In prior releases, the default reconnection UI was created by the framework in a way that could cause CSP violations. Note that this behavior is still used as fallback when the application does not define the reconnection UI (using the `ReconnectModal` component or a similar user-provided component)."

guardrex:
Thanks @oroztocil ... I'll work on this now and ping you back on the next commit.

guardrex:
... and @oroztocil ... see my questions in my OP ☝️. The first question is important on retaining the current images. The second question might not be something that we need in the coverage ... it could just be something simple that I don't understand.

guardrex:
> On the rejected state event, the component calls location.reload which means the browser reloads the page. After the reload, the application is either displayed normally (if it reconnects to the Blazor server successfully), or the reconnection UI shows up again and goes through the various states again from start.

I see ... yeah ... I think that's logical from reading the code. You can let me know if you think we should have remark about it at the end of the section.

aspnetcore/blazor/fundamentals/signalr.md
```diff
@@ -519,7 +519,58 @@ If reconnection fails, the user is instructed to retry or reload the page:
...
+<span id="components-reconnect-current-attempt"></span>
+```
+
+An element with an `id` of `components-seconds-to-next-attempt` displays the number of seconds to the next reconnection attempt:
+
+```html
+<span id="components-seconds-to-next-attempt"></span>
+```
+
+The Blazor Web App project template includes a `ReconnectModal` component (`Components/Layout/ReconnectModal.razor`) with collocated stylesheet and JavaScript files (`ReconnectModal.razor.css`, `ReconnectModal.razor.js`) that can be customized as needed. These files can be examined in the ASP.NET Core reference source or by inspecting an app created from the Blazor Web App project template with Interactive WebAssembly or Interactive Auto rendering enabled:
```
oroztocil, line 563: I overlooked this the first time, but the component is added to the project when:

- In VS, it is created with "Interactive render mode" set to `Server` or `Auto` (see image)
- In dotnet CLI, it is created with the option `--interactivity server` or `--interactivity auto`. (The `server` option is the default and is used when no option is specified.)

It is not added for "Interactive render mode" set to `WebAssembly`.

aspnetcore/blazor/fundamentals/signalr.md
```diff
@@ -519,7 +519,58 @@ If reconnection fails, the user is instructed to retry or reload the page:
 
 If reconnection succeeds, user state is often lost. Custom code can be added to any component to save and reload user state across connection failures. For more information, see <xref:blazor/state-management>.
 
-:::moniker range=">= aspnetcore-8.0"
+:::moniker range=">= aspnetcore-10.0"
+
+To create UI elements that track reconnection state, the following table describes:
+
+* A set of `components-reconnect-*` CSS classes (**Css class** column) that are set or unset by Blazor on an element with an `id` of `components-reconnect-modal`.
+* The `components-reconnect-state-changed` event value (**Event** column) that indicates a reconnection status change.
```
oroztocil, line 527: The event is dispatched on the `components-reconnect-modal` element as well (i.e. user has to set up on the event handler on this element).

aspnetcore/blazor/fundamentals/signalr.md
oroztocil: image

aspnetcore/blazor/fundamentals/signalr.md
guardrex: Come to think of it, I don't think this works the way that I have it, given that the developer's code must set up the event listener and react to it.

Let me see if I can explain out exactly what to do and re-organize this coverage a bit.

guardrex:
@oroztocil ... Is this getting *better* or ***worse?*** 😆 

It looks like we should show example code for wiring up the event, so I added a simplified example.

Also, note that I did state the default for the .NET CLI ...

> `--interactivity server` (default)

... but not for VS because IIRC 🤔 VS remembers the last app setup ... so Server/Auto isn't really a "default" for VS IIRC. Let me know if I'm wrong about that.

oroztocil (approved):
I think it's good now 🙂 I'd say this piece of functionality is inherently a bit difficult to document. (And I guess I should have written some summary of how the entire thing works to make it less confusing from the start.)

guardrex:
No worries ... that's ok. It perhaps saved a little of your time for me to take a stab at it and just fix my poor guesses 🙈😄.

PR #34869, opened 2025-03-03
Update openApi.md

visschersm:
An extra brace that should not be there.

Small fix on #59480


guardrex (approved):
Thanks @visschersm! 🚀

visschersm:
No problem!

PR #34947, opened 2025-03-11
Update first-web-api.md

dominiqueplante:
For me, putting the id in the body was the only way to get past the check in  the method:

```
if (id != todoDTO.Id)
{
            return BadRequest();
}
```

Tracking item added by wadepickett:
Fixes ##34954

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/tutorials/first-web-api.md](https://github.com/dotnet/AspNetCore.Docs/blob/b64fff56b90aed3d277a78af324200b07456fba4/aspnetcore/tutorials/first-web-api.md) | [Tutorial: Create a controller-based web API with ASP.NET Core](https://review.learn.microsoft.com/en-us/aspnet/core/tutorials/first-web-api?branch=pr-en-us-34947) |

### DIFF
--- a/aspnetcore/tutorials/first-web-api.md
+++ b/aspnetcore/tutorials/first-web-api.md
@@ -690,6 +690,7 @@ Use the `PUT` method to update the `TodoItem` that has Id = 1 and set its name t
   Content-Type: application/json
 
   {
+    "id": 1,
     "name": "feed fish",
     "isComplete": false
   }
@@ -711,6 +712,7 @@ Use Swagger to send a PUT request:
 
   ```json
   {
+    "id": 1,
     "name": "feed fish",
     "isComplete": false
   }


dominiqueplante:
To clarify a bit more, when I was testing, todoDTO.Id is 0 unless it is set

timdeschryver:
Oops, this is something that I've missed when porting this to .NET 9.
The change is indeed required to make it a valid request
@wadepickett  I'm pinging you because I just saw you assigned yourself to this issue.

Before: 400 bad request

image


After: 204 no content
image

dominiqueplante:
So it sounds like this is the correct fix :)

I just looked at the walkthrough for minimal APIs and saw the same issue there also, FYI ->

https://learn.microsoft.com/en-us/aspnet/core/tutorials/min-web-api?view=aspnetcore-9.0&tabs=visual-studio

@timdeschryver and @wadepickett, is there anything else I need to do?

wadepickett:
Thanks very much @dominiqueplante, looks right.  Just give me a moment to test it for versions/vs/vs studio text...

wadepickett:
Perfect, thanks @dominiqueplante!  Thanks @timdeschryver for checking as well!

wadepickett:
Also fixing https://learn.microsoft.com/en-us/aspnet/core/tutorials/min-web-api?view=aspnetcore-9.0&tabs=visual-studio#examine-the-put-endpoint ....

wadepickett:
I should have asked first @dominiqueplante, if you wanted the chance to fix the other tutorial as well go for it.  Otherwise, I can make the fix right away.

PR #35035, opened 2025-03-24
[Blazor] static-files.md - Replacing "server-side Blazor" with "ASP.NET Core hosted Blazor"

hakenr:
I think with today's terminology, "server-side Blazor app" isn't really descriptive anymore. "ASP.NET Core hosted Blazor app" is probably the label we want to use. (Not sure the intro sentences are still needed et al.)

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/static-files.md](https://github.com/dotnet/AspNetCore.Docs/blob/4c1f3d47bfa1608ce8d19842164f858766fea053/aspnetcore/blazor/fundamentals/static-files.md) | [ASP.NET Core Blazor static files](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/static-files?branch=pr-en-us-35035) |

### DIFF
--- a/aspnetcore/blazor/fundamentals/static-files.md
+++ b/aspnetcore/blazor/fundamentals/static-files.md
@@ -20,7 +20,7 @@ For general information on serving static files with Map Static Assets routing e
 
 :::moniker-end
 
-## Static asset delivery in server-side Blazor apps
+## Static asset delivery in ASP.NET Core hosted Blazor apps
 
 :::moniker range=">= aspnetcore-9.0"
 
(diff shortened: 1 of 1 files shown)


guardrex:
It was decided at the outset to retire the "hosted" naming to prevent confusion with hosted WASM, and "server-side" was the only way to make docs address Blazor Server and BWAs (when running on the server). This was all worked out for the .NET 8 release and touches hundreds of references across the doc set. You can open an issue to discuss it further if you want, but Dan and I are fairly happy with the current approach.

hakenr:
It's a bit confusing, especially with the overlap in naming with Blazor Server, but fair enough — I can roll with it.

guardrex:
We had a discussion on it. If we had *version-by-file* documentation versioning, we would've converted references to "Blazor Server" over to "Blazor Web App." Since we don't have good versioning, "server-side Blazor" was the only reasonable way to avoid massive numbers of `moniker-range` blocks to cover the change. I'm still lobbying Microsoft to adopt version-by-file. Six years later, we're still here with me begging for it 😩. If I ever get version-by-file, we might convert most of these to "Blazor Web App."

PR #35049, opened 2025-03-26
Update adding-view.md

DejanPopovic1:
VS  17.10.5 defaults a certain user step without user having to take an additional step.

### DIFF
--- a/aspnetcore/tutorials/first-mvc-app/adding-view.md
+++ b/aspnetcore/tutorials/first-mvc-app/adding-view.md
@@ -45,7 +45,7 @@ Right-click on the *Views* folder, and then **Add > New Folder** and name the fo
 
 Right-click on the *Views/HelloWorld* folder, and then **Add > New Item**.
 
-In the **Add New Item** dialog select **Show All Templates**.
+In the **Add New Item** dialog select **Show All Templates**. Visual Studio should by default show all templates without having to select **Show All Templates**.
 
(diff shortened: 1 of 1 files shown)


wadepickett:
Thanks @DejanPopovic1 for bringing this to light!

The issue is that the compact version of the Add New Item dialog is assumed.  While it is the default in VS on first run after install, selecting between the compact version and the all template version of the Add New Item dialog is sticky:  Whatever you chose last is what you get next time.

In a moment I will add a screen shot of the compact version and edit your text a bit and keeping your PR to complete it.  Thaks for offering a fix!

wadepickett (changes requested):
Needs to account for both the compact and all template view somehow.  Will edit.

aspnetcore/tutorials/first-mvc-app/adding-view.md
```diff
@@ -45,7 +45,7 @@ Right-click on the *Views* folder, and then **Add > New Folder** and name the fo
 
 Right-click on the *Views/HelloWorld* folder, and then **Add > New Item**.
 
-In the **Add New Item** dialog select **Show All Templates**.
+In the **Add New Item** dialog select **Show All Templates**. Visual Studio should by default show all templates without having to select **Show All Templates**.
```
wadepickett, line 48:
```suggestion
In the **Add New Item** dialog, if the option to **Show All Templates** is available, select it.
```

PR #35160, opened 2025-04-07
Include DateOnly and TimeOnly in query string supported types

Owenll66:


guardrex:
Hello @Owenll66 ... Thanks ... this sneaky one 😈 got in past us at 6.0, so I'll version it on your PR.

PR #35187, opened 2025-04-09
Prep only: WN: Validation Support Min API: Orig draft

wadepickett:
For .NET 10 Preview 3 What's New, moving original draft as-is for **Validation Support for Minimal API** to an include as a prep only so it can be easily compared with changes in the PR that follows with an edit pass.


Rick-Anderson:
@tdykstra @wadepickett  these prep PR's should fix all the cosmetic edits. Remove newlines, future tense, etc. Edit style we don't need the PU reviewing. That makes for a more productive review when we take the 2nd pass.

PR #35269, opened 2025-04-18
replace old versions SDK with latest released /1

Rick-Anderson:

### DIFF
--- a/aspnetcore/includes/0-latest-SDK.md
+++ b/aspnetcore/includes/0-latest-SDK.md
@@ -1,0 +1,1 @@
+[Latest released .NET Core SDK](https://dotnet.microsoft.com/download/dotnet-core)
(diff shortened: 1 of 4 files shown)


tdykstra:
Test

Sent from iPhone

aspnetcore/includes/0-latest-SDK.md
tdykstra, line 1: This goes too far, I think. The options are to say:

1. Version x is required. (This is what we have currently -- x is the actual version the tutorial was tested with.
1. Version x or later is required. (x is still the actual version the tutorial was tested with but we're not implicitly excluding later versions, as we are in no.1).
1. The latest released version is required. (This is the proposed change in this PR. Leaves people who actually have the correct version for the tutorial thinking they can't use it.)

I prefer #2 as the most accurate statement of what is required.

tdykstra (approved):

PR #35434, opened 2025-05-12
Improve docs for WasmPerfInstrumentation

pavelsavara:
---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics.md](https://github.com/dotnet/AspNetCore.Docs/blob/b94174117590a427c1b760eaaecead5a9b60942c/aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics.md) | [aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/performance/webassembly-event-pipe-diagnostics?branch=pr-en-us-35434) |

### DIFF
--- a/aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics.md
+++ b/aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics.md
@@ -53,10 +53,10 @@ The following table describes permissable `<WasmPerfInstrumentation>` values.
 `all` | All assemblies
 `none` |	No assemblies
 `program` | Entry point assembly
-`assembly` | Specifies an assembly
-`M:Type:Method` | Specifies a method
-`N:Namespace` | Specifies a namespace
-`T:Type` | Specifies a type
+`{ASSEMBLY}` | Specifies an assembly (`{ASSEMBLY}`)
+`M:Type:{METHOD}` | Specifies a method (`{METHOD}`)
+`N:{NAMESPACE}` | Specifies a namespace (`{NAMESPACE}`)
+`T:{TYPE}` | Specifies a type (`{TYPE}`)
 `+EXPR` | Includes expression
 `-EXPR` | Excludes expression
 
--- a/aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics.md
+++ b/aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics.md
@@ -42,7 +42,7 @@ The MSBuild properties in the following table enable profiler integration.
 Property | Default | Set value to&hellip; | Description
 --- | :---: | :---: | ---
 `<WasmPerfTracing>` | `false` | `true` | Enables support for WebAssembly performance tracing.
-`<WasmPerfInstrumentation>` | `none` | See table&dagger; | Enables instrumentation necessary for the sampling profiler. The property follows the :::no-loc text="callspec"::: syntax. &dagger;For permissible values, see the following table.
+`<WasmPerfInstrumentation>` | No value | See table&dagger; | Enables instrumentation necessary for the sampling profiler. The property follows the :::no-loc text="callspec"::: syntax. &dagger;For permissible values, see the following table.
 `<MetricsSupport>` | `false` | `true` | Enables `System.Diagnostics.Metrics` support. For more information, see the [`System.Diagnostics.Metrics` namespace](/dotnet/api/system.diagnostics.metrics).
 `<EventSourceSupport>` | `false`| `true` | Enables `EventPipe` support. For more information, see [Diagnostics and instrumentation: Observability and telemetry](/dotnet/core/deploying/native-aot/diagnostics#observability-and-telemetry).
 
@@ -51,7 +51,6 @@ The following table describes permissable `<WasmPerfInstrumentation>` values.
 `<WasmPerfInstrumentation>` value | Description
 --- | ---
 `all` | All assemblies
-`none` |	No assemblies
 `program` | Entry point assembly
 `{ASSEMBLY}` | Specifies an assembly (`{ASSEMBLY}`)
 `M:Type:{METHOD}` | Specifies a method (`{METHOD}`)
--- a/aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics.md
+++ b/aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics.md
@@ -42,7 +42,7 @@ The MSBuild properties in the following table enable profiler integration.
 Property | Default | Set value to&hellip; | Description
 --- | :---: | :---: | ---
 `<WasmPerfTracing>` | `false` | `true` | Enables diagnostic server.
-`<WasmPerfInstrumentation>` | `false` | `true` | Enables instrumentation necessary for sampling profiler.
+`<WasmPerfInstrumentation>` | `false` | `all` | Enables instrumentation necessary for sampling profiler.
 `<MetricsSupport>` | `false` | `true` | Controls `System.Diagnostics.Metrics` support. For more information, see the [`System.Diagnostics.Metrics` namespace](/dotnet/api/system.diagnostics.metrics).
 `<EventSourceSupport>` | `false`| `true` | Controls `EventPipe` support. For more information, see [Diagnostics and instrumentation: Observability and telemetry](/dotnet/core/deploying/native-aot/diagnostics#observability-and-telemetry).
 
--- a/aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics.md
+++ b/aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics.md
@@ -41,7 +41,8 @@ The MSBuild properties in the following table enable profiler integration.
 
 Property | Default | Set value to&hellip; | Description
 --- | :---: | :---: | ---
-`<WasmPerfTracing>` | `false` | `true` | Controls diagnostic server tracing.
+`<WasmPerfTracing>` | `false` | `true` | Enables diagnostic server.
+`<WasmPerfInstrumentation>` | `false` | `true` | Enables instrumentation necessary for sampling profiler.
 `<MetricsSupport>` | `false` | `true` | Controls `System.Diagnostics.Metrics` support. For more information, see the [`System.Diagnostics.Metrics` namespace](/dotnet/api/system.diagnostics.metrics).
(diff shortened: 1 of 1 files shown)


guardrex (approved):

danroth27:
The `WasmPerfInstrumentation` property follows the callspec syntax. Can clarify that here as well and link to the callspec syntax like we do for the WasmProfilers property in the browser dev tools profiling doc? Also do we want to document that `true` is treated the same as `all`?

guardrex:
Here's a cut-'n-paste link for that ...

> The WasmPerfInstrumentation property follows the :::no-loc text="callspec"::: syntax. For more information, see \[Trace MonoVM profiler events during startup](https://github.com/dotnet/runtime/blob/main/docs/design/mono/diagnostics-tracing.md#trace-monovm-profiler-events-during-startup).

@pavelsavara ... If you want me to place it, let me know 👂.

guardrex:
@pavelsavara ... I'm not sure on this point ...

> Also do we want to document that `true` is treated the same as `all`?

What are the potential values? ... or where could a dev find them if they're published elsewhere?

pavelsavara:
Maybe we could copy&paste that table with callspec here directly. 
Becuase the document we link talks about "trace MonoVM profiler events during startup" but for wasm, we don't implement it during VM startup (in Net 10 at least)

pavelsavara:
Note the same link is also in the dev tools page

guardrex:
I did some work with placeholders following our usual conventions. I'm still not sure about ...

* If the default is `none` ... or is it `false`. If `false`, I'll revert back to that.
* `+EXPR`/`-EXPR` ... I think it should probably be `+{EXPRESSION}` and `-{EXPRESSION}`. I'll need a clarification ... do these refer to regex? If so, I'll make the table description call that out.

guardrex:
**UPDATE**: ***Not*** regex. We'll leave them as `+EXPR`/`-EXPR`.

If "`none`" turns out to be incorrect for the `MSBuild` property ... if it should be "`false`" for the MS Build property resulting in `none` being passed ... let me know 👂. What I'll do is update the first table's mention of the default value back to `false`, and I'll update the value in the second table to reflect passing `false` to get `none` passed, if that's what's happening internally.

aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics.md
```diff
@@ -41,10 +41,25 @@ The MSBuild properties in the following table enable profiler integration.
 
 Property | Default | Set value to&hellip; | Description
 --- | :---: | :---: | ---
-`<WasmPerfTracing>` | `false` | `true` | Controls diagnostic server tracing.
+`<WasmPerfTracing>` | `false` | `true` | Enables diagnostic server.
+`<WasmPerfInstrumentation>` | `none` | See table&dagger; | Enables instrumentation necessary for the sampling profiler. The property follows the :::no-loc text="callspec"::: syntax. &dagger;For permissible values, see the following table.
 `<MetricsSupport>` | `false` | `true` | Controls `System.Diagnostics.Metrics` support. For more information, see the [`System.Diagnostics.Metrics` namespace](/dotnet/api/system.diagnostics.metrics).
 `<EventSourceSupport>` | `false`| `true` | Controls `EventPipe` support. For more information, see [Diagnostics and instrumentation: Observability and telemetry](/dotnet/core/deploying/native-aot/diagnostics#observability-and-telemetry).
```
danroth27, line 47:
```suggestion
`<MetricsSupport>` | `false` | `true` | Enables `System.Diagnostics.Metrics` support. For more information, see the [`System.Diagnostics.Metrics` namespace](/dotnet/api/system.diagnostics.metrics).
`<EventSourceSupport>` | `false`| `true` | Enables `EventPipe` support. For more information, see [Diagnostics and instrumentation: Observability and telemetry](/dotnet/core/deploying/native-aot/diagnostics#observability-and-telemetry).
```

aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics.md
```diff
@@ -41,10 +41,25 @@ The MSBuild properties in the following table enable profiler integration.
 
 Property | Default | Set value to&hellip; | Description
 --- | :---: | :---: | ---
-`<WasmPerfTracing>` | `false` | `true` | Controls diagnostic server tracing.
+`<WasmPerfTracing>` | `false` | `true` | Enables diagnostic server.
```
danroth27, line 44: @pavelsavara How is this property different from what `EventSourceSupport` does?

I'll suggest something more generic for now:

```suggestion
`<WasmPerfTracing>` | `false` | `true` | Enables support for WebAssembly performance tracing.
```

aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics.md
```diff
@@ -41,10 +41,25 @@ The MSBuild properties in the following table enable profiler integration.
...
+`<WasmPerfInstrumentation>` | `none` | See table&dagger; | Enables instrumentation necessary for the sampling profiler. The property follows the :::no-loc text="callspec"::: syntax. &dagger;For permissible values, see the following table.
 `<MetricsSupport>` | `false` | `true` | Controls `System.Diagnostics.Metrics` support. For more information, see the [`System.Diagnostics.Metrics` namespace](/dotnet/api/system.diagnostics.metrics).
 `<EventSourceSupport>` | `false`| `true` | Controls `EventPipe` support. For more information, see [Diagnostics and instrumentation: Observability and telemetry](/dotnet/core/deploying/native-aot/diagnostics#observability-and-telemetry).
 
+The following table describes permissable `<WasmPerfInstrumentation>` values.
+
+`<WasmPerfInstrumentation>` value | Description
+--- | ---
+`all` | All assemblies
```
danroth27, line 53:
```suggestion
`all` or `true` | All assemblies
```

guardrex: `true`/`all` and I'll need clarification on `false`/`none` to get the 😈 ironed out.

aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics.md
```diff
@@ -41,10 +41,25 @@ The MSBuild properties in the following table enable profiler integration.
 
 Property | Default | Set value to&hellip; | Description
 --- | :---: | :---: | ---
-`<WasmPerfTracing>` | `false` | `true` | Controls diagnostic server tracing.
+`<WasmPerfTracing>` | `false` | `true` | Enables diagnostic server.
+`<WasmPerfInstrumentation>` | `none` | See table&dagger; | Enables instrumentation necessary for the sampling profiler. The property follows the :::no-loc text="callspec"::: syntax. &dagger;For permissible values, see the following table.
 `<MetricsSupport>` | `false` | `true` | Controls `System.Diagnostics.Metrics` support. For more information, see the [`System.Diagnostics.Metrics` namespace](/dotnet/api/system.diagnostics.metrics).
 `<EventSourceSupport>` | `false`| `true` | Controls `EventPipe` support. For more information, see [Diagnostics and instrumentation: Observability and telemetry](/dotnet/core/deploying/native-aot/diagnostics#observability-and-telemetry).
 
+The following table describes permissable `<WasmPerfInstrumentation>` values.
```
danroth27, line 49: @pavelsavara What are these values really specifying? Do they specify the parts of the app that should be instrumented for performance tracing? Can you only specify one of these values? Or can they be composed using some sort of list syntax?

aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics.md
```diff
@@ -41,10 +41,25 @@ The MSBuild properties in the following table enable profiler integration.
...
+`none` |	No assemblies
+`program` | Entry point assembly
+`{ASSEMBLY}` | Specifies an assembly (`{ASSEMBLY}`)
+`M:Type:{METHOD}` | Specifies a method (`{METHOD}`)
+`N:{NAMESPACE}` | Specifies a namespace (`{NAMESPACE}`)
+`T:{TYPE}` | Specifies a type (`{TYPE}`)
+`+EXPR` | Includes expression
+`-EXPR` | Excludes expression
+
 Enabling profilers has negative size and performance impact, so don't publish an app for production with profilers enabled. In the following example, a condition is set on a property group section that only enables profiling when the app is built with `/p:BlazorSampleProfilingEnabled=true` (.NET CLI) or `<BlazorSampleProfilingEnabled>true</BlazorSampleProfilingEnabled>` in a Visual Studio publish profile, where "`BlazorSampleProfilingEnabled`" is a custom symbol name that you choose and doesn't conflict with other symbol names.
```
guardrex, line 63:
```suggestion
Enabling profilers has negative size and performance impacts, so don't publish an app for production with profilers enabled. In the following example, a condition is set on a property group section that only enables profiling when the app is built with `/p:BlazorSampleProfilingEnabled=true` (.NET CLI) or `<BlazorSampleProfilingEnabled>true</BlazorSampleProfilingEnabled>` in a Visual Studio publish profile, where "`BlazorSampleProfilingEnabled`" is a custom symbol name that you choose and doesn't conflict with other symbol names.
```

aspnetcore/blazor/performance/webassembly-event-pipe-diagnostics.md
pavelsavara: > @pavelsavara What are these values really specifying? Do they specify the parts of the app that should be instrumented for performance tracing? 

App and runtime managed methods that should be instrumented. If they are not instrumented, they can't be used to take sample stack trace. Such method could still be visible in the profile, if they call some other method that is instrumented.

For the callspec I reused existing mono implementation, so I don't know all the details right now.
It seems to me that you can chain the filters via `+` and `-`.

guardrex: @pavelsavara ...  For `<WasmPerfInstrumentation>`, the tables may still have problems 😈 describing the settings `all`/`true` and `none`/`false`. Is setting `all` the same as passing `true` and is setting `none` the same as passing `false`?

pavelsavara: Not talking about true/false is easiest way out

guardrex:
AFAICT, the last couple of commits should resolve the last few problems.

I've also mirrored changes for the other article on https://github.com/dotnet/AspNetCore.Docs/pull/35451.

@danroth27 ... I'll merge both of these PRs after I hear back from you.

PR #35437, opened 2025-05-12
Improve callspec coverage

guardrex:
Addresses https://github.com/dotnet/AspNetCore.Docs/pull/35434#issuecomment-2872901402

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/performance/webassembly-browser-developer-tools-diagnostics.md](https://github.com/dotnet/AspNetCore.Docs/blob/91461920b498926b6f0d64634469013210a6a7ba/aspnetcore/blazor/performance/webassembly-browser-developer-tools-diagnostics.md) | [aspnetcore/blazor/performance/webassembly-browser-developer-tools-diagnostics](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/performance/webassembly-browser-developer-tools-diagnostics?branch=pr-en-us-35437) |

### DIFF
--- a/aspnetcore/blazor/performance/webassembly-browser-developer-tools-diagnostics.md
+++ b/aspnetcore/blazor/performance/webassembly-browser-developer-tools-diagnostics.md
@@ -87,7 +87,7 @@ Add `callspec` to the `browser` WebAssembly profiler in the `<WasmProfilers>` el
 <WasmProfilers>browser:callspec={KEYWORD};</WasmProfilers>
 ```
 
-Permissiable keyword values are shown in the following table.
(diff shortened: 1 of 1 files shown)


guardrex:
I did some work with the table content to specify placeholders.

pavelsavara:
> I'll want to say if "EXPR" (probably changed to "{EXPRESSION}`) means regex? I assume that's right ... that it's regex.

Most likely not regex, this probably just means you can do say "include namespace Bla but exclude class Foo"

guardrex:
@pavelsavara ... I see. No further changes needed then. I have that as "+EXPR" and "-EXPR" here matching the original table.

PR #35481, opened 2025-05-16
Add warning about keepalives

Smaug123:
I'm going to be as clear as I can in this PR description about the symptoms we observed, in case it helps people Googling in the future.

Here's what happens when you copy-paste the client configuration from the keepalive docs:

* The default KeepAlive policy [is set at the HTTP level](https://github.com/dotnet/runtime/blob/b8752af999a212847b43034e7ef44cf6292b98c3/src/libraries/System.Net.Http/src/System/Net/Http/HttpHandlerDefaults.cs#L17) to "Always", which means that if you don't set a `KeepAlivePingPolicy` in the gRPC client settings but you *do* override `KeepAlivePingDelay` from its default of infinity ("don't send pings"), then the client will inherit the default "Always" and will send keepalives for the duration of the connection.
* `PooledConnectionIdleTimeout` in the docs snippet in this section is overridden from [the default "1 minute"](https://github.com/dotnet/runtime/blob/b8752af999a212847b43034e7ef44cf6292b98c3/src/libraries/Common/src/System/Net/Http/HttpHandlerDefaults.cs#L29) to "infinity".
* Keepalives [require cooperation with the server](https://github.com/grpc/grpc.io/blob/4896f5ed74f505637bea8c7618f6c6c2d0cee22c/content/en/docs/guides/keepalive.md#L24), as I describe in the patch.

This means that any given connection is open forever, pinging the server periodically (whether or not there is an active stream on the connection); and the server is completely free to close such a connection.

We experienced a hang condition where we speculate that this is what happened:

* Client performs a unary gRPC call to server, causing an HTTP/2 connection to be created.
* Call completes; connection is returned to client's pool.
* Client transparently starts sending pings once per minute after a minute of inactivity.
* Server ignores the first couple of pings, and then sends GOAWAY, in compliance with gRPC docs.
* Client performs a unary gRPC call, and is handed the connection from the pool to do so.
* The connection closes due to the GOAWAY.
* Client hangs because its unary gRPC call never completes.

The fix is simply not to attempt keepalives, because the server doesn't support them.

In this PR I essentially add a warning that there are other docs that the user must understand before enabling the feature, and a one-line summary of the failure mode they should expect if they get this wrong.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/grpc/performance.md](https://github.com/dotnet/AspNetCore.Docs/blob/23d8dca834d6980a788307fc6f3e3190cfc7cd83/aspnetcore/grpc/performance.md) | [aspnetcore/grpc/performance](https://review.learn.microsoft.com/en-us/aspnet/core/grpc/performance?branch=pr-en-us-35481) |

### DIFF
--- a/aspnetcore/grpc/performance.md
+++ b/aspnetcore/grpc/performance.md
@@ -4,7 +4,7 @@ author: jamesnk
 description: Learn the best practices for building high-performance gRPC services.
 monikerRange: '>= aspnetcore-3.0'
 ms.author: wpickett
-ms.date: 04/11/2023
+ms.date: 05/16/2025
 uid: grpc/performance
 ---
 # Performance best practices with gRPC
--- a/aspnetcore/grpc/performance.md
+++ b/aspnetcore/grpc/performance.md
@@ -158,7 +158,7 @@ Consider using a transport like Unix domain sockets or named pipes for gRPC call
 ## Keep alive pings
 
 > [!IMPORTANT]
-> Keep alive pings require the cooperation of the server. Do not enable keep alive pings in the client without verifying that the server supports them. A server which does not support keep alive pings will usually ignore the first few pings, and will then send a GOAWAY message, closing the active HTTP/2 connection.
+> Keep alive pings require the cooperation of the server. Do not enable keep alive pings in the client without verifying that the server supports them. A server which does not support keep alive pings will usually ignore the first few pings, and will then send a `GOAWAY` message, closing the active HTTP/2 connection.
 
 Keep alive pings can be used to keep HTTP/2 connections alive during periods of inactivity. Having an existing HTTP/2 connection ready when an app resumes activity allows for the initial gRPC calls to be made quickly, without a delay caused by the connection being reestablished.
 
--- a/aspnetcore/grpc/performance.md
+++ b/aspnetcore/grpc/performance.md
@@ -157,6 +157,9 @@ Consider using a transport like Unix domain sockets or named pipes for gRPC call
 
 ## Keep alive pings
 
+> [!IMPORTANT]
+> Keep alive pings require the cooperation of the server. Do not enable keep alive pings in the client without verifying that the server supports them. A server which does not support keep alive pings will usually ignore the first few pings, and will then send a GOAWAY message, closing the active HTTP/2 connection.
+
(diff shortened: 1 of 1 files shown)


aspnetcore/grpc/performance.md
```diff
@@ -157,6 +157,9 @@ Consider using a transport like Unix domain sockets or named pipes for gRPC call
 
 ## Keep alive pings
 
+> [!IMPORTANT]
+> Keep alive pings require the cooperation of the server. Do not enable keep alive pings in the client without verifying that the server supports them. A server which does not support keep alive pings will usually ignore the first few pings, and will then send a GOAWAY message, closing the active HTTP/2 connection.
```
wadepickett, line 161:
```suggestion
> Keep alive pings require the cooperation of the server. Do not enable keep alive pings in the client without verifying that the server supports them. A server which does not support keep alive pings will usually ignore the first few pings, and will then send a `GOAWAY` message, closing the active HTTP/2 connection.
```
Very minor item, added missing formatting.

wadepickett (approved):
Looks good to me, thanks for your contribution!  I just made a very minor formatting change.  I will wait to see what @JamesNK thinks before merging.

wadepickett:
I may have somehow introduced a doc build error.  Looking at the error message, it may just be that it needs to rebuild.  To do that I am closing and reopenining

guardrex:
We want to get that other PR merged ***quickly*** 🏃‍♂️  in order to update that IARD article to get rid of these build errors.

You can override and merge this to `main` right now tho. We just don't want to merge to `live` until that other PR is merged. We're just waiting on Dan to get back to us on the ".NET X" lingo.

wadepickett:
@guardrex, PR number of what you want merged so I don't mess it up.

guardrex:
I'm referring to this PR. You don't need to let this build error on this PR prevent you from overriding and squash-merging it.

When this goes on the 'Merge to Live' PR, it will generate the same error there. After the other PR is merged (after we hear back from Dan), the error will go away on the 'Merge to Live' PR.

wadepickett:
I also see 35476 is the other PR in question, thanks.

wadepickett:
Thanks again @Smaug123.

wadepickett:
@JamesNK, in the rush, I forgot to wait for your input on this.  Sorry, if you decide what was addressed in this PR should change in some way let me know and I will immediately update.

Issue #35225, opened 2025-04-15
Add documentation for JWT authentication in Blazor


MarvinKlein1508:
## Tell us what search terms you used and how you searched docs.
I've searched for JWT and authentification.

## Tell us what docs you found that didn't address your concern.
Basically the entire tree here:
https://learn.microsoft.com/en-us/aspnet/core/blazor/security/?view=aspnetcore-9.0&tabs=visual-studio

# Describe the new topic

## Explain why this topic is needed.
Many people create ASP.NET Core Web Apis as backend for their applications. Those often rely on JWT authentification for their endpoints. Mixing blazor into all of this makes this a really complicated process. Literally any tutorial you find about this topic is differs from each other and there are no real best practices.

Should you even use JWT authentification in Blazor at all? If so how should you handle the different rendermodes?

In my example there endpoints for authentication right within the Web Api which return an access and refresh token which should be used to authenticate the user in the blazor app. The token shoudl also be applied to all requests from `HttpClient`. 

## Suggest a location in the Table of Contents.
This should be located right into the Blazor section.
https://learn.microsoft.com/en-us/aspnet/core/blazor/security/?view=aspnetcore-9.0&tabs=visual-studio

## Write an abstract. In one **short** paragraph, describe what this topic will cover.
The docs should cover how to save the access and refresh token in the different blazor scenarios. As well as automatically fetching a new access token when it has been expired. Last but not least it should show how the access token should be attached to the `HttpClient` requests.

guardrex:
Hello @MarvinKlein1508 ... For BWAs, this is somewhat of duplicate of an [existing issue](https://github.com/dotnet/AspNetCore.Docs/issues/31691) for the [passing tokens](https://learn.microsoft.com/en-us/aspnet/core/blazor/security/additional-scenarios?view=aspnetcore-9.0#pass-tokens-to-a-server-side-blazor-app) section of the server-side security article and a tracking item that I maintain in another docs issue. Currently, we demonstrate several approaches for calling web APIs in the articles that describe specific project scenarios for BWAs. See how the web APIs are set up for the BFF patterns in the following articles ...

* https://learn.microsoft.com/en-us/aspnet/core/blazor/security/blazor-web-app-with-entra?view=aspnetcore-9.0&pivots=non-bff-pattern
* https://learn.microsoft.com/en-us/aspnet/core/blazor/security/blazor-web-app-with-oidc?view=aspnetcore-9.0&pivots=bff-pattern
[...]

PR #35482, opened 2025-05-16
Add the latest recommended passing tokens approach

### LINKED ISSUE #31691: PU review of *Pass tokens to a server-side Blazor app* section

guardrex:
Fixes #31691
Addresses #35225

As usual, the updates are based on 🦖 ***RexHacks!***&trade; 🙈. Get ready for the code smells! 💩👃😝

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/call-web-api.md](https://github.com/dotnet/AspNetCore.Docs/blob/c9eac0603f3b290e3532b5aa82dbf55e881be0d2/aspnetcore/blazor/call-web-api.md) | [aspnetcore/blazor/call-web-api](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/call-web-api?branch=pr-en-us-35482) |
| [aspnetcore/blazor/security/additional-scenarios.md](https://github.com/dotnet/AspNetCore.Docs/blob/c9eac0603f3b290e3532b5aa82dbf55e881be0d2/aspnetcore/blazor/security/additional-scenarios.md) | [aspnetcore/blazor/security/additional-scenarios](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/additional-scenarios?branch=pr-en-us-35482) |
| [aspnetcore/blazor/security/blazor-web-app-with-entra.md](https://github.com/dotnet/AspNetCore.Docs/blob/c9eac0603f3b290e3532b5aa82dbf55e881be0d2/aspnetcore/blazor/security/blazor-web-app-with-entra.md) | [aspnetcore/blazor/security/blazor-web-app-with-entra](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/blazor-web-app-with-entra?branch=pr-en-us-35482) |
| [aspnetcore/blazor/security/blazor-web-app-with-oidc.md](https://github.com/dotnet/AspNetCore.Docs/blob/c9eac0603f3b290e3532b5aa82dbf55e881be0d2/aspnetcore/blazor/security/blazor-web-app-with-oidc.md) | [aspnetcore/blazor/security/blazor-web-app-with-oidc](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/blazor-web-app-with-oidc?branch=pr-en-us-35482) |
| [aspnetcore/blazor/security/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/c9eac0603f3b290e3532b5aa82dbf55e881be0d2/aspnetcore/blazor/security/index.md) | [aspnetcore/blazor/security/index](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/index?branch=pr-en-us-35482) |
| [aspnetcore/blazor/security/webassembly/additional-scenarios.md](https://github.com/dotnet/AspNetCore.Docs/blob/c9eac0603f3b290e3532b5aa82dbf55e881be0d2/aspnetcore/blazor/security/webassembly/additional-scenarios.md) | [aspnetcore/blazor/security/webassembly/additional-scenarios](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/additional-scenarios?branch=pr-en-us-35482) |


aspnetcore/blazor/call-web-api.md
```diff
@@ -33,7 +33,7 @@ For more information, see the following resources:
 
 ## Microsoft identity platform for web API calls
 
-Blazor Web Apps that use use [Microsoft identity platform](/entra/identity-platform/)/[Microsoft Identity Web packages](/entra/msal/dotnet/microsoft-identity-web/) for [Microsoft Entra ID](https://www.microsoft.com/security/business/microsoft-entra) can make streamlined web API calls with API provided by the [`Microsoft.Identity.Web.DownstreamApi` NuGet package](https://www.nuget.org/packages/Microsoft.Identity.Web.DownstreamApi).
+Blazor Web Apps that use use [Microsoft identity platform](/entra/identity-platform/) with [Microsoft Identity Web packages/API](/entra/msal/dotnet/microsoft-identity-web/) for [Microsoft Entra ID](https://www.microsoft.com/security/business/microsoft-entra) can make streamlined web API calls with API provided by the [`Microsoft.Identity.Web.DownstreamApi` NuGet package](https://www.nuget.org/packages/Microsoft.Identity.Web.DownstreamApi).
```
halter73, line 36: Why "packages/API" and not just "packages"? I feel like "packages" implies API.

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -23,12 +23,231 @@ This article explains how to configure server-side Blazor for additional securit
...
+    cookieOptions.Events.OnSigningIn = (context) =>
+    {
+        if (context.Principal?.Identity is not null && 
+            context.Principal.Identity.IsAuthenticated)
+        {
+            var accessToken = context.Properties.GetTokenValue("access_token");
+            var claimsIdentity = new ClaimsIdentity(context.Principal?.Identity,
+                [new Claim("AccessToken", accessToken ?? "No Access Token!")]);
+            context.Principal = new ClaimsPrincipal(claimsIdentity);
+            context.Properties.Items.Remove("access_token");
```
halter73, line 62: I think the reason this isn't working is because GetTokenValue isn't directly indexing, it's [prefixing the key with ".Token."](https://github.com/dotnet/aspnetcore/blob/7ba32e7e88a3b8013863b1316b0ca595029d0a26/src/Http/Authentication.Abstractions/src/TokenExtensions.cs#L64). `.Remove(".Token.access_token")` would probably work.

However, if we wind up needing to use IHttpContextAccessor anyway, I'd rather just leave the access token in the AuthenticationProperties rather than putting it into a Claim. I know it was my suggestion, but I was only willing to do it to get rid of IHttpContextAccessor. And I'm not sure if it would have been worth it even then on second thought.

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -23,12 +23,231 @@ This article explains how to configure server-side Blazor for additional securit
...
+    {
+        if (AuthState is not null)
+        {
+            var authState = await AuthState;
+            var user = authState?.User;
+
+            if (user is not null)
+            {
+                var accessToken = 
+                    user.Claims.First(c => c.Type == "AccessToken")?.Value;
```
halter73, line 229: I realize that since I'm going back on my suggestion to make an access token claim, we'll have to come up with another solution to get the access token during server interactive rendering. What about using IHttpContextAccessor in a CircuitHandler to store the access token in a scoped service? It seems like it'd be similar complexity of doing the same thing with PersistentComponentState but would be less likely to invite people with web assembly projects unwittingly exposing the wrong kind of access token to their public browser client.

guardrex: > What about using IHttpContextAccessor in a CircuitHandler to store the access token in a scoped service?

I'm not able to get the CircuitHandler approach to work. I think you'll have to give me the code to make it work, keeping in mind that I need it to work for Auto-rendering apps (CSR) ... but that's related to my next question ................

Could you explain this a bit more ...

> would be less likely to invite people with web assembly projects unwittingly exposing the wrong kind of access token to their public browser client.

Do you intend that devs should never pass an access token to an `Auto` component for use under CSR?

aspnetcore/blazor/security/additional-scenarios.md
guardrex: > leave the access token in the AuthenticationProperties rather than putting it into a Claim

... but that seems to make getting the token(s) around the app more challenging. It's easy with the claims-based approach to get them from anywhere in the app, and it seems like it's easy to have the cookie refresher update them.

aspnetcore/blazor/security/additional-scenarios.md
halter73: > If the answer is 'yes' ... 😨 don't do that (unsafe middle-tier use) ... 

I think the answer is yes. We should never be passing an access token acquired by the server to the client for CSR. I think you're bullet points are fine. I can try to come up with a sample using a CircuitHandler, but I won't have time for that in the next couple weeks.

guardrex:
@halter73 ... This is a good improvement thus far. The latest updates come directly from our discussions. I'd like to go ahead and get this live now, and I'll probably perform some touch-ups to it next week. Then, I hope you'll look at the two main sections that cover this subject in the live articles within the next couple of weeks. I'll send you the live article links.

WRT a CircuitHandler approach, I tried again today and ran into trouble again. I couldn't get it work. I even think that I should check our guidance at ...

https://learn.microsoft.com/en-us/aspnet/core/blazor/security/additional-scenarios?view=aspnetcore-9.0#access-authenticationstateprovider-in-outgoing-request-middleware

... to confirm that it still works as we describe because I tried using a similar approach to that in one of my efforts and it didn't even look like that was working properly. I'll check on it soon.

PR #35492, opened 2025-05-18
[Blazor] Rendering - Rerender logic adjustments

hakenr:
The component itself doesn't decide whether to rerender - this logic is handled by Blazor. A rerender happens if the received parameter values might have changed (e.g. they're mutable objects) **and** the component's `ShouldRender()` method returns `true` (which it does by default unless overridden).

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/performance/rendering.md](https://github.com/dotnet/AspNetCore.Docs/blob/246ad0ee29f01af43621c02e1ebc6150ecca6017/aspnetcore/blazor/performance/rendering.md) | [aspnetcore/blazor/performance/rendering](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/performance/rendering?branch=pr-en-us-35492) |


guardrex (approved):
Thanks @hakenr! 🎷

PR #35493, opened 2025-05-18
[Blazor] rendering - event callbacks change detection

hakenr:
Event callbacks are now part of the "parameter might have changed" detection logic.

### DIFF
--- a/aspnetcore/blazor/performance/rendering.md
+++ b/aspnetcore/blazor/performance/rendering.md
@@ -30,7 +30,7 @@ To prevent rendering recursion into a particular subtree, use either of the foll
 
 * Ensure that child component parameters are of primitive immutable types, such as `string`, `int`, `bool`, `DateTime`, and other similar types. The built-in logic for detecting changes automatically skips rerendering if the primitive immutable parameter values haven't changed. If you render a child component with `<Customer CustomerId="item.CustomerId" />`, where `CustomerId` is an `int` type, then the `Customer` component isn't rerendered unless `item.CustomerId` changes.
 * Override <xref:Microsoft.AspNetCore.Components.ComponentBase.ShouldRender%2A>:
-  * To accept nonprimitive parameter values, such as complex custom model types, event callbacks, or <xref:Microsoft.AspNetCore.Components.RenderFragment> values.
+  * To accept nonprimitive parameter values, such as complex custom model types, or <xref:Microsoft.AspNetCore.Components.RenderFragment> values.
   * If authoring a UI-only component that doesn't change after the initial render, regardless of parameter value changes.
 
 The following airline flight search tool example uses private fields to track the necessary information to detect changes. The previous inbound flight identifier (`prevInboundFlightId`) and previous outbound flight identifier (`prevOutboundFlightId`) track information for the next potential component update. If either of the flight identifiers change when the component's parameters are set in [`OnParametersSet`](xref:blazor/components/lifecycle#after-parameters-are-set-onparameterssetasync), the component is rerendered because `shouldRender` is set to `true`. If `shouldRender` evaluates to `false` after checking the flight identifiers, an expensive rerender is avoided:


guardrex:
@ilonatommy @hakenr  ... This was written by either Pranav or Steve several years ago ... I think 🤔 it was Pranav. Besides the change here, I don't like this sentence structure. If the change to drop "event callbacks" from the sentence is correct, can we change it to something easier to understand? How about something like ...

> When parameters receive nonprimitive values, such as complex custom model types or \<xref:Microsoft.AspNetCore.Components.RenderFragment> values.

ilonatommy:
I don't understand the need of this change. Events are non-primitive objects and presence of `[Parameter] public EventCallback` in the component will result in re-render by default. The docs say: override that setting in case you are dealing with exactly this type of objects if you want to block rendering. Which is correct.

At the same time, the justification of the change:

> Event callbacks are now part of the "parameter might have changed" detection logic.

Is also true. "parameter might have changed" => we are dealing with non-primitives, like events and we are re-rendering unless you override.

About the phrasing:
Sure, we can change it, I trust @guardrex's word order sounds better.

guardrex:
In that case, the potential revision is ...

> When parameters receive nonprimitive values, such as complex custom model types, event callbacks, or \<xref:Microsoft.AspNetCore.Components.RenderFragment> values.

Also, I'm going to change the API doc cross-link to point to our section on `ShouldRender`, like I did on the last PR. The API doc is linked from there.

hakenr:
@ilonatommy @guardrex The event callbacks are now treated as immutable.

> Events are non-primitive objects and the presence of `[Parameter] public EventCallback` in a component will result in a re-render by default.

This is no longer true. See [ChangeDetection.cs, line 59](https://github.com/dotnet/aspnetcore/blob/caf0f3f2dae893395a8a80989315e08b38ea11a6/src/Components/Components/src/ChangeDetection.cs#L59), introduced in [this commit](https://github.com/dotnet/aspnetcore/commit/998c6c64bc58e818b85211f63df9d248579d4662).

guardrex:
In that case, we'd need to version it out at 8.0. I'll put a commit here, and we can see how it looks. *Stand-by .............*

guardrex:
Can we list the types? 🤔 It would be nice to be specific. I realize this can be difficult to keep up with. An alternative is to link the reference source, where devs can see them directly for any release version. It would harden the article if we cross-link for them.

guardrex:
... and I added a line for overriding `ShouldRender` that includes the three primitive types that Blazor doesn't support for change detection ...

> When parameters receive primative \`IntPtr\`, \`UIntPtr\`, or \`Type\` values, which aren't supported for change detection.

guardrex:
Yeah ... I have a feeling that we're better off with a reference source link here to ...

https://github.com/dotnet/aspnetcore/blob/main/src/Components/Components/src/ChangeDetection.cs

... instead of trying to list them.

I'll try another update to this. *Stand-by ..............................*

guardrex:
It's in their code comments.

Mmmmmmm 🤔 ... Yes, we can drop that now that we're going to cross-link ref source.

hakenr:
```csharp
    // For performance reasons, the following immutable types are not supported: IntPtr, UIntPtr, Type.
```

I see.
But the comment says "following immutable types" not "following primitive types".
BTW, there are many other immutable types which are not "supported" here, such as `ImmutableList`, `ImmutableDictionary`, tuples, ...

hakenr (commented):
lgtm

ilonatommy (approved):
Thank you, @hakenr for providing more context. The approach with linking the source code sounds, indeed the best in the moment.

@guardrex, one nit: "nonprimitive" and "immutable" should be adjectives for "types", not "values".

guardrex:
Thanks @ilonatommy ... I'll make that change now.

PR #35495, opened 2025-05-19
[Blazor] rendering perf - ShouldRender nits

hakenr:
`ShouldRender` is a method, it is not being set, it returns a value.

I also added a part to clarify that you shouldn't always return false - only when the parameter values haven't changed. (Though feel free to cut that if it's too much detail.)

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/performance/rendering.md](https://github.com/dotnet/AspNetCore.Docs/blob/11816147f56c233c6c9a6058d2b884b6c2e2b6a3/aspnetcore/blazor/performance/rendering.md) | [ASP.NET Core Blazor rendering performance best practices](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/performance/rendering?branch=pr-en-us-35495) |


guardrex (approved):
# 👍

PR #35527, opened 2025-05-26
feat: sample for big-file-upload via `MultiPartReader` / `IPipeReader` / `IFormFeature` (net9)

### LINKED ISSUE #34575: Modify file upload sample to handle very large files
### Description

[Enter feedback here]

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/mvc/models/file-uploads?view=aspnetcore-9.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/mvc/models/file-uploads.md

### Document ID

c6787f0f-3e9a-6efb-0e6e-5ccb957d3f4a

### Article author

@tdykstra

### Metadata

* Service: **aspnet-core**
* Sub-service: **mvc**

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+c6787f0f-3e9a-6efb-0e6e-5ccb957d3f4a)

DeagleGross:
There are several issues related to large-file-uploads, and I am proposing a fresh sample of how that can work in net9+ (will work for lower versions as well).

It is a new solution with a single csproj, which shows 3 different implementations of how to read the contents of a file uploaded via `Content-Type: multipart/form-data` using:
- `MultiPartReader` (per ask from #34575)
- `PipeReader` ("from scratch" solution)
- `IFormFeature` (because it is an existing wrapper already)

Sample covers both Controllers and Minimal API approaches.

_Note: is a finish of #33809_

Fixes #34575
Fixes https://github.com/dotnet/aspnetcore/issues/58233

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/mvc/models/file-uploads.md](https://github.com/dotnet/AspNetCore.Docs/blob/bc2a547e71f090d50f8d6cbd7c9b9c9090d4cb94/aspnetcore/mvc/models/file-uploads.md) | [aspnetcore/mvc/models/file-uploads](https://review.learn.microsoft.com/en-us/aspnet/core/mvc/models/file-uploads?branch=pr-en-us-35527) |

### DIFF
--- a/aspnetcore/mvc/models/file-uploads.md
+++ b/aspnetcore/mvc/models/file-uploads.md
@@ -415,45 +415,6 @@ The preceding example is similar to a scenario demonstrated in the sample app:
 
 ### Upload large files with streaming
 
-For scenarios where large file uploads are required, streaming uploads allow you to process incoming multipart form data directly, without buffering the entire file in memory or on disk via model binding. This technique is especially important for files that could exceed server or framework buffering thresholds.
-
-The [sample application for 9.x](file-uploads/samples/9.x) demonstrates how a server can receive a file and stream the data directly to disk, supporting robust cancellation via the HTTP request's cancellation token. There are several approaches.
-
-[`MultipartReader`](/dotnet/api/microsoft.aspnetcore.webutilities.multipartreader) is an ASP.NET Core utility for reading files from incoming requests. The following snippet shows how to process a request and stream the file into an `outputStream` (such as a `FileStream`):
-```csharp
-// Read the boundary from the Content-Type header
-var boundary = HeaderUtilities.RemoveQuotes(MediaTypeHeaderValue.Parse(request.ContentType).Boundary).Value;
-
-// Use MultipartReader to stream data to a destination
-var reader = new MultipartReader(boundary, Request.Body);
-MultipartSection? section;
-while ((section = await reader.ReadNextSectionAsync(cancellationToken)) != null)
-{
-    var contentDisposition = section.GetContentDispositionHeader();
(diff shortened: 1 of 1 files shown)


adityamandaleeka (approved):

aspnetcore/mvc/models/file-uploads/samples/9.x/FileManagerSample/Services/FileManagerService.cs
```diff
@@ -0,0 +1,108 @@
...
+        private void CheckAndRemoveLocalFile(string filePath)
+        {
+            if (File.Exists(filePath))
+            {
+                File.Delete(filePath);
+                _logger.LogDebug($"Removed existing output file: {filePath}");
+            }
+        }
+
+        public async Task<string?> SaveViaPipeReaderAsync(PipeReader contentReader, CancellationToken cancellationToken)
```
BrennanConroy, line 72: This method is saving the raw request bytes including the multipart framing. Is that the expected behavior?

arielmoraes:
I tried to use the `IFormFeature` approach, but the body is buffered as soon as the method `ReadFormAsync` is called. Am I missing something?

I think the culprit is this line: https://github.com/dotnet/aspnetcore/blob/4f805f07dc65d9080940ce7a110e29eaef80f811/src/Http/Http/src/Features/FormFeature.cs#L254

If rewind is being enabled for files, the sample is not correct when used to stream large files. Even more, the rewind should not happen if `BufferBody` is disabled.

PR #35575, opened 2025-06-04
Replace <a name= with <a id= for deep linking

jzabroski:
* `<a name=` is a legacy HTML4 way to refer to sections of a document
* HTML5 recommends `<a id=`, and allows for `id` to be placed on more items for enhanced accessibility
* for some reason, on Google Chrome Version 137.0.7151.56 (Official Build) (64-bit), the deep link to https://learn.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-6.0#tcer does not scroll to that section of the document. This could be due to some peculiarities in how Microsoft Learn loads dynamic content e.g. via JavaScript and how the browser is processing those fragments in load events. I have not debugged why, as it is unclear to me that `<a name` is a recommended approach across Microsoft Learn. If it is, forgive me.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/security/cors.md](https://github.com/dotnet/AspNetCore.Docs/blob/988022d6695fdddf22a4cce9ad8aab7b6140ef44/aspnetcore/security/cors.md) | [Enable Cross-Origin Requests (CORS) in ASP.NET Core](https://review.learn.microsoft.com/en-us/aspnet/core/security/cors?branch=pr-en-us-35575) |

### DIFF
--- a/aspnetcore/security/cors.md
+++ b/aspnetcore/security/cors.md
@@ -59,7 +59,7 @@ Using the [[EnableCors]](#attr) attribute with a named policy provides the fines
 
 Each approach is detailed in the following sections.
 
-<a name="np"></a>
+<a id="np"></a>
 
 ## CORS with named policy and middleware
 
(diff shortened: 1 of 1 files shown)


guardrex:
Hello @jzabroski ... These are known limitations and internal features of our doc system, and we don't want to change the approach taken in a piecemeal fashion like this. Management might be considering a new type of versioning, called *group topic versioning*, that will eliminate these types of cross-linking problems.

cc: @cmastr

jzabroski:
ok, but can you confirm you also can't use the anchors in this page when they reference an `<a name`? What's the point in having something that is broken?

guardrex:
This is under active internal discussion. Management is aware of the problems associated with these crosslinks based on named anchors, including how they aren't HTML5 compliant and thus probably don't work in all browsers. No decision has been made on what direction to take, but I provided my recommendation to embrace group topic versioning without named anchor crosslinking.

jzabroski:
thank you. helpful.

PR #35610, opened 2025-06-11
[JsonPatch] Patch docs for STJ docs points to Newtonsoft package

tebeco:
the documentation for the "new" Http `Patch` using STJ points to the nuget package of the package still using `Newtonsoft.Json`


guardrex (approved):
Thanks @tebeco, and it doesn't look like we had time to edit this one, so I gave it a quick pass 🏃‍♂️.

tebeco:
better later than sorry ;)

thx for taking time here as always

PR #35636, opened 2025-06-17
Update docfx.json for ms.subservice entries

wadepickett:
Two added entries to ensure articles at these two paths receive an ms.subservice value of "diagnostics"

       "**/diagnostics/**/**.md": "diagnostics",
       "**/log-mon/**/**.md": "diagnostics",


wadepickett:
Since this is a new ms.subservice value we are adding here, it has to be registered first or we get build warning.

I registered `diagnostics` today and hopefully it will be approved and added within the next two days.

Closing this PR and will try again with a new PR once the new value is confirmed as registered.

PR #35771, opened 2025-07-18
Update for aspnetcore-10.0.md release notes with correct heading level

selvinfehric:
Update for aspnetcore-10.0.md release notes with the correct heading level for Authentication and authorization


guardrex (approved):
Thanks @selvinfehric! 🚀 ... I put a commit in to just drop that bit because there's only one subsection. We'll elevate the subsection to H2.

There will be a bit of a delay on getting this merged. We have an internal issue right now that we're sorting out. I'll get this merged as soon as we're back in good graces with the 🙏 *server gods* 🙏😆.

PR #35830, opened 2025-07-26
Fix invalid ms.author values

wadepickett:
Fix invalid ms.author & author metadata values with valid ms alias for owners responsible for the content going forward.


wadepickett:
@tdykstra, this is the first 50 for what we split up and assigned that needed a valid value.  I set author as well since we don't get notifications otherwise.  There are about 50 more for what we already worked out and I will do a PR for that right after this one merges. (I think best not to do all 106 at the same time.)

tdykstra (approved):

Issue #35827, opened 2025-07-25
Reassign topics with invalid ms.author


wadepickett:
Reassign topics that have no valid ms.author entry.  There are around 100+ of these to fix that are currently resulting in build warnings.

---
[Associated WorkItem - 474822](https://dev.azure.com/msft-skilling/Content/_workitems/edit/474822)

PR #35849, opened 2025-07-29
Fix invalid ms.author part 2

wadepickett:
Fix invalid ms.author & author metadata values with valid ms alias for owners responsible for the content going forward.

Contributes to https://github.com/dotnet/AspNetCore.Docs/issues/35827


wadepickett:
@tdykstra, here is another 53.  Looks like we have one more to do after this.

PR #35853, opened 2025-07-29
SQLite guidance for case insensitive filtering

### LINKED ISSUE #35851: Case sensitive comparison
### Description

Hi,

By default, the lambda expression m => m.Title!.Contains(TitleFilter ?? string.Empty) is case sensitive, at least, nothing was returned when I searched for "road" or "road warrior" but movies were returned when I searched for "Road" for example.

I had to update the lambda expression to: context.Movie.Where(m => m.Title!.ToLower().Contains((TitleFilter ?? string.Empty).ToLower())); for it to work properly with the lowercase search.

Thanks!

### Page URL

https://learn.microsoft.com/en-gb/aspnet/core/blazor/tutorials/movie-database-app/part-6?view=aspnetcore-9.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/tutorials/movie-database-app/part-6.md

### Document ID

6d15bd42-ea91-ec59-1ebc-185653dc7ef4

### Platform Id

bc8a6fac-a9c8-44e8-4883-26b69c934483

### Article author

@guardrex

### Metadata

* ID: 6d15bd42-ea91-ec59-1ebc-185653dc7ef4
* PlatformId: bc8a6fac-a9c8-44e8-4883-26b69c934483 
* Service: **aspnet-core**
* Sub-service: **blazor**

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+6d15bd42-ea91-ec59-1ebc-185653dc7ef4)

guardrex:
Fixes #35851

Thanks @rodbeck! 🚀 

Dan, we might ***not*** want to do this, or we might want to do something different. I started by trying to update the instructions to specifically tell SQLite folks to use case sensitive search terms (e.g., "Road Warrior" instead of "road warrior"). The problem with modifying all of the instructions that way is that it muddies up the text with distracting parenthetical information that still doesn't match the images. Therefore if we do anything, it might be best to show the workaround approach of using `ToLower` (on this PR), which is simpler and quicker than getting into the weeds with SQLite dB collation, which we do cross-link.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/tutorials/movie-database-app/part-6.md](https://github.com/dotnet/AspNetCore.Docs/blob/69268a422701f8aedfeb95fdbb832a9ea9e17c49/aspnetcore/blazor/tutorials/movie-database-app/part-6.md) | [aspnetcore/blazor/tutorials/movie-database-app/part-6](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/tutorials/movie-database-app/part-6?branch=pr-en-us-35853) |

### DIFF
--- a/aspnetcore/blazor/tutorials/movie-database-app/part-6.md
+++ b/aspnetcore/blazor/tutorials/movie-database-app/part-6.md
@@ -60,7 +60,7 @@ The `movie => movie.Title!.Contains(...)` code is a *lambda expression*. Lambdas
 
 The <xref:System.Data.Objects.DataClasses.EntityCollection%601.Contains%2A> method is run on the database, not in the C# code. The case sensitivity of the query depends on the database and the collation. For SQL Server, <xref:System.String.Contains%2A> maps to [SQL `LIKE`](/sql/t-sql/language-elements/like-transact-sql), which is case insensitive. SQLite with default collation provides a mixture of case-sensitive and case-insensitive filtering, depending on the query. The remainder of this tutorial assumes case-insensitive database collation.
 
-If you're following this tutorial using Visual Studio Code or the .NET CLI, you're using a SQLite database. To adopt case-insensitive collation, open the `Data/BlazorWebAppMoviesContext.cs` file. Inside the `BlazorWebAppMoviesContext` class, add the following code:
+To adopt case-insensitive collation when using SQLite (<xref:Microsoft.EntityFrameworkCore.SqliteDbContextOptionsBuilderExtensions.UseSqlite%2A> is called in `Program.cs`), open the `Data/BlazorWebAppMoviesContext.cs` file. Inside the `BlazorWebAppMoviesContext` class, add the following code:
 
 ```csharp
 protected override void OnModelCreating(ModelBuilder modelBuilder)


danroth27:
@AndriySvyryd @luisquintanilla @roji What do you recommend here in this intro tutorial for dealing with SQLite case sensitivity? Should we be configuring a collation on the model instead?

roji:
Yeah, the right solution here is to create your column with the NOCASE collation ([SQLite documentation on collations](https://sqlite.org/datatype3.html#collating_sequences), [EF documentation on collations](https://learn.microsoft.com/en-us/ef/core/miscellaneous/collations-and-case-sensitivity#column-collation)) - that's what we should be guiding users towards.

Now, I'm not aware of any support for a "database" collation in SQLite; this does exist in other databases, but AFAIK not in SQLite. SQLite allows you to define collations on the column level - that's the thing to show (so basically [this](https://learn.microsoft.com/en-us/ef/core/miscellaneous/collations-and-case-sensitivity#column-collation)) with NOCASE).

Other than that, FYI the "workaround" with using ToLower() is generally discouraged as it means any index on the column won't be used. We generally try to not show that anti-pattern as people invariably copy-paste it as the simple/easy/intuitive thing, and frequently don't realize they're killing perf by doing it. So I'd recommend against showing this.

BTW: this PR is referencing System.Data.Objects.DataClasses.EntityCollection, which is the old non-core EF6. I'm assuming you want to show EF Core, no?
 
/cc @cincuranet

danroth27:
> the right solution here is to create your column with the NOCASE collation

@roji We can't assume the user is using SQLite - if the user is on Windows in VS, I believe they'll get SQL Server by default. I assume the NOCASE collation is SQLite specific, so we'll need to add some text to tell the user to add that only if they're using SQLite, correct?

cincuranet:
Yes. `NOCASE` is SQLite specific. And only ASCII characters are case folded.

A cross-database "workaround" could be to have dedicated extra column for search and this column will be lower-cased before `SaveChanges` is called.

roji:
Sorry @danroth27, misread this as being specific to SQLite. Unfortunately the situation with collations really is different across databases - SQL Server is already case-insensitive by default (unless the user specifically tweaked the SQL Server installation, they get case-insensitive behavior), where PostgreSQL and SQLite are case-sensitive. Actual collation names are database-specific.

So I'd basically tell users to configure their column collation if they're not getting the results they want, and point them to the EF page above.

guardrex:
Thanks folks! ... I have enough info to fix the tutorial now.

@danroth27 ... It will be tomorrow morning, and I'll ping you back here when this is ready for an 👁️.

guardrex:
@danroth27 ... Updates complete. Ready for a look. 👀

aspnetcore/blazor/tutorials/movie-database-app/part-6.md
```diff
@@ -58,7 +58,16 @@ Change the `QuickGrid` component's <xref:Microsoft.AspNetCore.Components.QuickGr
...
+The <xref:System.Data.Objects.DataClasses.EntityCollection%601.Contains%2A> method is run on the database, not in the C# code. The case sensitivity of the query depends on the database and the collation. For SQL Server, <xref:System.String.Contains%2A> maps to [SQL `LIKE`](/sql/t-sql/language-elements/like-transact-sql), which is case insensitive. SQLite with default collation provides a mixture of case-sensitive and case-insensitive filtering, depending on the query. The remainder of this tutorial assumes case-insensitive database collation.
+
+If you're following this tutorial using Visual Studio Code or the .NET CLI, you're using a SQLite database. To adopt case-insensitive collation, open the `Data/BlazorWebAppMoviesContext.cs` file. Inside the `BlazorWebAppMoviesContext` class, add the following code:
```
danroth27:
```suggestion
To adopt case-insensitive collation when using SQLite (<xref:Microsoft.EntityFrameworkCore.SqliteDbContextOptionsBuilderExtensions.UseSqlite%2A> is called in `Program.cs`), open the `Data/BlazorWebAppMoviesContext.cs` file. Inside the `BlazorWebAppMoviesContext` class, add the following code:
```

guardrex: The reason that I wrote it that way is that I'm not sure how many new devs and new-to-.NET devs are going to know that they're using a SQLite dB at this point. What they do probably know is what IDE approach they've taken. Are you sure you want to risk it? 😄

guardrex: > change the default database provider we use.

I see ur point, but I'm worried that they won't know if they need the code or not stated that way.

guardrex: It's not really going to harden things much because the whole tutorial will need work in several spots if the default dB changes for either VSC or the .NET CLI.

danroth27: It's pretty easy to check if you're using SQLite by looking in *Program.cs* to see if you're calling `UseSqlite`. Do you want to just add a sentence to tell the user how to do that?

guardrex: 🤔 ... Yeah ... that's better. I'll place a short passage here to that effect.

guardrex:
That lick'd IT. It now reads ...

> To adopt case-insensitive collation when using SQLite (`UseSqlite` is called in `Program.cs`), open the `Data/BlazorWebAppMoviesContext.cs` file. Inside the `BlazorWebAppMoviesContext` class, add the following code:

*Thanks everyone!* 🍻

aspnetcore/blazor/tutorials/movie-database-app/part-6.md
```diff
@@ -58,7 +58,16 @@ Change the `QuickGrid` component's <xref:Microsoft.AspNetCore.Components.QuickGr
...
+The <xref:System.Data.Objects.DataClasses.EntityCollection%601.Contains%2A> method is run on the database, not in the C# code. The case sensitivity of the query depends on the database and the collation. For SQL Server, <xref:System.String.Contains%2A> maps to [SQL `LIKE`](/sql/t-sql/language-elements/like-transact-sql), which is case insensitive. SQLite with default collation provides a mixture of case-sensitive and case-insensitive filtering, depending on the query. The remainder of this tutorial assumes case-insensitive database collation.
+
+To adopt case-insensitive collation when using SQLite (<xref:Microsoft.EntityFrameworkCore.SqliteDbContextOptionsBuilderExtensions.UseSqlite%2A> is called in `Program.cs`), open the `Data/BlazorWebAppMoviesContext.cs` file. Inside the `BlazorWebAppMoviesContext` class, add the following code:
+
+```csharp
+protected override void OnModelCreating(ModelBuilder modelBuilder)
+{
+    modelBuilder.UseCollation("SQL_Latin1_General_CP1_CS_AS");
+}
+```
```
AndriySvyryd, line 70: This is incorrect on multiple levels. But crucially EF translates `Contains` to `instr` on SQLite and that **always** performs a case sensitive comparison independently of the collation. The correct solution is to call `ToLower` first.

guardrex: ## Don't reply here ... let's move the discussion to #35868.

I see. Where is this covered in EF Core docs because ...

https://learn.microsoft.com/en-us/ef/core/miscellaneous/collations-and-case-sensitivity

... isn't a good cross-link for this particular scenario, namely SQLite and case sensitivity.

guardrex: ## Don't reply here ... let's move the discussion to #35868.

BTW @AndriySvyryd ... It hasn't gone live, so you did catch it in time. Thanks for jumping on to let me know. I'll fix it right now.

guardrex: ## Don't reply here ... let's move the discussion to #35868.

I'm looking back over the earlier remarks, and there seems to be disagreement ...

> SQLite allows you to define collations on the column level - that's the thing to show (so basically [this](https://learn.microsoft.com/en-us/ef/core/miscellaneous/collations-and-case-sensitivity#column-collation)) with NOCASE).

... versus ...

> The correct solution is to call ToLower first.

So ... will the following work?

```csharp
protected override void OnModelCreating(ModelBuilder modelBuilder)
{
    modelBuilder.Entity<Movie>().Property(c => c.Title).UseCollation("NOCASE");
}
```

If it's unclear, I'll test it, but I'll do so in the morning. I'm just about at 🛌💤🐑🐑🐑 time.

guardrex: Let's take it up on the PR. I'll start with @roji, and then I have a feeling that we'll be moving to @AndriySvyryd over there.

https://github.com/dotnet/AspNetCore.Docs/pull/35868

PR #13364, opened 2019-07-17
Blazor state management topic

guardrex:
Fixes #13259 

[Internal Review Topic](https://review.docs.microsoft.com/en-us/aspnet/core/blazor/state-management?view=aspnetcore-3.0&branch=pr-en-us-13364)

* **HOLD** merge until package is on NuGet. I'll keep an :eye: on it.
* I recommend the generic title that I use here (sans "server-side") so that the article can address client-side state further later. "Server-side" is in the article description, and there are ample remarks to explain that this topic applies to server-side only. This is just a recommendation, of course. I'll react to your feedback, including if you prefer the title be more specific about "server-side" application. Just have a feel'in that it will change soon enough to say *something* more about Blazor client-side state management ... even if it's only to [mention and cross-link some 3rd party options (here are a few)](https://www.nuget.org/packages?q=blazor+localstorage+sessionstorage).
* Let's withhold a common sample app demo for now. The 🧀 could move on this one.
* The "circuits" concept is new to docs. We have an issue now for engineering to flesh that out, perhaps in the hosting models topic :point_right: [Document "circuits" for Blazor server-side (aspnet/AspNetCore.Docs \#13360)](https://github.com/aspnet/AspNetCore.Docs/issues/13360).

PR #35873, opened 2025-07-31
Blazor state management updates

### LINKED ISSUE #35717: [Pre6 Followup Work] Blazor state persistence
### Description

There's coverage in a few spots on circuit loss/state preservation that don't take into account the new circuit preservation features landing for 10.0. The issue at https://github.com/dotnet/AspNetCore.Docs/issues/35711 pertains to only new coverage for *What's New*. This follow-up issue is to track changes to existing coverage elsewhere across the Blazor docs. This will include *State Management* article updates and probably moving the component state preservation content from the *Render* article to the *State Management* article.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/state-management?view=aspnetcore-10.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/state-management.md

### Document ID

e5e1273b-195e-5da1-b4aa-66bbcff1425b

### Platform Id

fe5f15b3-5a89-60c5-6b83-4291dfac10d1

### Article author

@guardrex

### Metadata

* ID: ae14b3e3-3271-b919-9e0a-6f317a6b2d2d
* PlatformId: fe5f15b3-5a89-60c5-6b83-4291dfac10d1 
* Service: **aspnet-core**
* Sub-service: **blazor**

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+e5e1273b-195e-5da1-b4aa-66bbcff1425b)

guardrex:
Fixes #35717

### Notes

* Creates a node for *State Management* articles.
  * ***Overview***: Introduction and scenarios common to server-side Blazor and Blazor WebAssembly apps.
  * ***Prerendered state persistence***: Moves from a section of the *Prerendering* article to a new article in the *State Management* node.
  * ***Server***: Server-side Blazor scenarios; move the Pre6 coverage on circuit state persistence from the *What's New* article to a section in this article. Cross-link from *What's New* to the coverage.
  * ***WebAssembly***: Blazor WebAssembly scenarios.
  * ***Protected browser storage***: Split out browser storage into its own article because its a large enough subject to merit its own article.
* There was a bug in the Pre6 coverage that escaped Javier's review that came from Javier's example code in a PU issue discussion on circuit state preservation: The `[SupplyParameterFromPersistentComponentState]` attribute missing the word "parameter" in one part of the docs, which is corrected by this PR.

I'm aware of the upcoming Pre7 name change for the `[SupplyParameterFromPersistentComponentState]` attribute, which will be taken care of on the upcoming Pre7 docs PR.

---

#### Internal previews

<details><summary><strong>Toggle expand/collapse</strong></summary><br/>

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/call-web-api.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/call-web-api.md) | [aspnetcore/blazor/call-web-api](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/call-web-api?branch=pr-en-us-35873) |
| [aspnetcore/blazor/components/cascading-values-and-parameters.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/components/cascading-values-and-parameters.md) | [aspnetcore/blazor/components/cascading-values-and-parameters](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/cascading-values-and-parameters?branch=pr-en-us-35873) |
| [aspnetcore/blazor/components/data-binding.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/components/data-binding.md) | [aspnetcore/blazor/components/data-binding](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/data-binding?branch=pr-en-us-35873) |
| [aspnetcore/blazor/components/integration-hosted-webassembly.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/components/integration-hosted-webassembly.md) | [aspnetcore/blazor/components/integration-hosted-webassembly](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/integration-hosted-webassembly?branch=pr-en-us-35873) |
| [aspnetcore/blazor/components/integration.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/components/integration.md) | [aspnetcore/blazor/components/integration](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/integration?branch=pr-en-us-35873) |
| [aspnetcore/blazor/components/lifecycle.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/components/lifecycle.md) | [aspnetcore/blazor/components/lifecycle](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?branch=pr-en-us-35873) |
| [aspnetcore/blazor/components/prerender.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/components/prerender.md) | [aspnetcore/blazor/components/prerender](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/prerender?branch=pr-en-us-35873) |
| [aspnetcore/blazor/components/render-modes.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/components/render-modes.md) | [aspnetcore/blazor/components/render-modes](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/render-modes?branch=pr-en-us-35873) |
| [aspnetcore/blazor/components/rendering.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/components/rendering.md) | [aspnetcore/blazor/components/rendering](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/rendering?branch=pr-en-us-35873) |
| [aspnetcore/blazor/file-uploads.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/file-uploads.md) | [aspnetcore/blazor/file-uploads](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/file-uploads?branch=pr-en-us-35873) |
| [aspnetcore/blazor/fundamentals/dependency-injection.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/fundamentals/dependency-injection.md) | [aspnetcore/blazor/fundamentals/dependency-injection](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/dependency-injection?branch=pr-en-us-35873) |
| [aspnetcore/blazor/fundamentals/environments.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/fundamentals/environments.md) | [aspnetcore/blazor/fundamentals/environments](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/environments?branch=pr-en-us-35873) |
| [aspnetcore/blazor/fundamentals/routing.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/fundamentals/routing.md) | [aspnetcore/blazor/fundamentals/routing](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/routing?branch=pr-en-us-35873) |
| [aspnetcore/blazor/fundamentals/signalr.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/fundamentals/signalr.md) | [aspnetcore/blazor/fundamentals/signalr](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/signalr?branch=pr-en-us-35873) |
| [aspnetcore/blazor/host-and-deploy/configure-linker.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/host-and-deploy/configure-linker.md) | [aspnetcore/blazor/host-and-deploy/configure-linker](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/configure-linker?branch=pr-en-us-35873) |
| [aspnetcore/blazor/security/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/security/index.md) | [aspnetcore/blazor/security/index](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/index?branch=pr-en-us-35873) |
| [aspnetcore/blazor/security/interactive-server-side-rendering.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/security/interactive-server-side-rendering.md) | [aspnetcore/blazor/security/interactive-server-side-rendering](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/interactive-server-side-rendering?branch=pr-en-us-35873) |
| [aspnetcore/blazor/security/webassembly/additional-scenarios.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/security/webassembly/additional-scenarios.md) | [aspnetcore/blazor/security/webassembly/additional-scenarios](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/webassembly/additional-scenarios?branch=pr-en-us-35873) |
| [aspnetcore/blazor/state-management/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/state-management/index.md) | [aspnetcore/blazor/state-management/index](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/state-management/index?branch=pr-en-us-35873) |
| [aspnetcore/blazor/state-management/prerendered-state-persistence.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/state-management/prerendered-state-persistence.md) | [aspnetcore/blazor/state-management/prerendered-state-persistence](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/state-management/prerendered-state-persistence?branch=pr-en-us-35873) |
| [aspnetcore/blazor/state-management/protected-browser-storage.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/state-management/protected-browser-storage.md) | [aspnetcore/blazor/state-management/protected-browser-storage](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/state-management/protected-browser-storage?branch=pr-en-us-35873) |
| [aspnetcore/blazor/state-management/server.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/state-management/server.md) | [aspnetcore/blazor/state-management/server](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/state-management/server?branch=pr-en-us-35873) |
| [aspnetcore/blazor/state-management/webassembly.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/blazor/state-management/webassembly.md) | [aspnetcore/blazor/state-management/webassembly](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/state-management/webassembly?branch=pr-en-us-35873) |
| [aspnetcore/fundamentals/app-state.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/fundamentals/app-state.md) | [aspnetcore/fundamentals/app-state](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/app-state?branch=pr-en-us-35873) |
| [aspnetcore/fundamentals/servers/yarp/config-providers.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/fundamentals/servers/yarp/config-providers.md) | [aspnetcore/fundamentals/servers/yarp/config-providers](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/yarp/config-providers?branch=pr-en-us-35873) |
| [aspnetcore/mvc/views/tag-helpers/built-in/persist-component-state.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/mvc/views/tag-helpers/built-in/persist-component-state.md) | [aspnetcore/mvc/views/tag-helpers/built-in/persist-component-state](https://review.learn.microsoft.com/en-us/aspnet/core/mvc/views/tag-helpers/built-in/persist-component-state?branch=pr-en-us-35873) |
| [aspnetcore/release-notes/aspnetcore-5.0.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/release-notes/aspnetcore-5.0.md) | [aspnetcore/release-notes/aspnetcore-5.0](https://review.learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-5.0?branch=pr-en-us-35873) |
| [aspnetcore/release-notes/aspnetcore-8.0.md](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/release-notes/aspnetcore-8.0.md) | [aspnetcore/release-notes/aspnetcore-8.0](https://review.learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-8.0?branch=pr-en-us-35873) |
| [aspnetcore/toc.yml](https://github.com/dotnet/AspNetCore.Docs/blob/904cd3619322cd1a1e198db235dd956e346e2bcd/aspnetcore/toc.yml) | [aspnetcore/toc](https://review.learn.microsoft.com/en-us/aspnet/core/toc?branch=pr-en-us-35873) |

</details>


aspnetcore/blazor/state-management/index.md
```diff
@@ -0,0 +1,168 @@
...
+description: Learn how to persist user data (state) in Blazor apps.
+monikerRange: '>= aspnetcore-3.1'
+ms.author: wpickett
+ms.custom: mvc
+ms.date: 08/01/2025
+uid: blazor/state-management/index
+---
+# ASP.NET Core Blazor state management overview
+
+This article and the other articles in this node describe common approaches for maintaining a user's data (state) while they use an app and across browser sessions, including during server prerendering.
```
danroth27, line 13: I think it would be helpful to provide a broader overview of what state management in Blazor is all about and to frame the problems that it's intended to solve.

One broad problem is how to share state across multiple components. The doc already covers some of these options but doesn't really frame them as solutions to this more general problem. I think sharing state across components breaks down into something this:

- Parent to child: A parent component can pass state to a child component using parameters.
- Child to parent: A child component can enable data binding to its state or provide state through callbacks.
- Parent to descendants: A parent to share state with all its descendants using cascading values.
- App-wide: State can be shared across the entire app using configured app state services.
- Per circuit: State can be shared for a specific circuit using scoped app state services.

Then, there's also the broad problem of how to persist state so that it survives pages refreshes, resumed circuits, and prerending. This state can be stored in a bunch of different places on the client and on the server. A key aspect of state persistence is getting your state persisted and loaded at the right time and when it's needed.

And there's the problem of managing state so that it can be centrally managed, tracked, and tested. That's what state containers like Flux, Redux, and MobX do. Blazor doesn't provide any sort of opinionated state management (just like React doesn't), but it can be used with a variety of popular state management solutions. We could provide links to these solutions.

aspnetcore/blazor/state-management/persistent-component-state.md
```diff
@@ -0,0 +1,355 @@
...
+title: ASP.NET Core Blazor persistent component state
+author: guardrex
+description: Learn how to persist user data (state) in Blazor apps using Blazor's Persistent Component State service.
+monikerRange: '>= aspnetcore-8.0'
+ms.author: wpickett
+ms.custom: mvc
+ms.date: 08/04/2025
+uid: blazor/state-management/persistent-component-state
+---
+# ASP.NET Core Blazor persistent component state
```
danroth27: How about this for the title?

```suggestion
# ASP.NET Core Blazor persisting state
```

aspnetcore/blazor/state-management/persistent-component-state.md
```diff
@@ -0,0 +1,355 @@
...
+
+Without persisting component state, state used during prerendering is lost and must be recreated when the app is fully loaded. If any state is created asynchronously, the UI may flicker as the prerendered UI is replaced when the component is rerendered.
```
danroth27: Persisting component state isn't just about prerendering. Can we lead off with a more general introduction into why persisting component state is important? It's useful for reusing the state from prerendering, pausing & resuming circuits, and maintaining state across browser refreshes.

aspnetcore/blazor/state-management/persistent-component-state.md
guardrex: *Nevermind!* This is fine ... it's a present participle, not a gerund, in this title.

aspnetcore/blazor/state-management/persistent-component-state.md
```diff
@@ -0,0 +1,355 @@
...
+
+Consider the following `PrerenderedCounter1` counter component. The component sets an initial random counter value during prerendering in [`OnInitialized` lifecycle method](xref:blazor/components/lifecycle#component-initialization-oninitializedasync). After the SignalR connection to the client is established, the component rerenders, and the initial count value is replaced when `OnInitialized` executes a second time.
```
danroth27:
```suggestion
Consider the following `PrerenderedCounter1` counter component. The component sets an initial random counter value during prerendering in [`OnInitialized` lifecycle method](xref:blazor/components/lifecycle#component-initialization-oninitializedasync). When the component then renders interactively, the initial count value is replaced when `OnInitialized` executes a second time.
```

aspnetcore/blazor/state-management/persistent-component-state.md
danroth27: Maybe "Blazor state persistence"?

aspnetcore/blazor/state-management/persistent-component-state.md
```diff
@@ -0,0 +1,355 @@
...
+
+`PrerenderedCounter1.razor`:
+
+:::code language="razor" source="~/../blazor-samples/8.0/BlazorSample_BlazorWebApp/Components/Pages/PrerenderedCounter1.razor":::
```
danroth27: Can we have this code sample just use the default render mode instead of enabling prerendering explicitly? It's important that users understand that prerendering is enabled by default.

aspnetcore/blazor/state-management/persistent-component-state.md
guardrex: Yes, that sounds better. There's nothing really wrong with the former, too. We're careful with gerunds because they cause translation problems; however, that earlier one ("persisting") isn't a gerund, so it wasn't a problem.

aspnetcore/blazor/state-management/persistent-component-state.md
```diff
@@ -0,0 +1,355 @@
...
+
+To retain the initial value of the counter during prerendering, Blazor supports persisting state in a prerendered page using the <xref:Microsoft.AspNetCore.Components.PersistentComponentState> service (and for components embedded into pages or views of Razor Pages or MVC apps, the [Persist Component State Tag Helper](xref:mvc/views/tag-helpers/builtin-th/persist-component-state-tag-helper)).
+
+:::moniker range=">= aspnetcore-10.0"
+
+<!-- UPDATE 10.0 - API cross-links -->
+
+To preserve prerendered state, use the `[SupplyParameterFromPersistentComponentState]` attribute to persist state in properties. Properties with this attribute are automatically persisted using the <xref:Microsoft.AspNetCore.Components.PersistentComponentState> service during prerendering. The state is retrieved when the component renders interactively or the service is instantiated.
+
+By default, properties are serialized using the <xref:System.Text.Json?displayProperty=fullName> serializer with default settings. Serialization isn't trimmer safe and requires preservation of the types used. For more information, see <xref:blazor/host-and-deploy/configure-trimmer>.
```
danroth27: Can we clarify here that this guidance really only applies to WebAssembly-based rendering?

aspnetcore/blazor/state-management/persistent-component-state.md
```diff
@@ -0,0 +1,355 @@
...
+
+<h1>Prerendered Counter 2</h1>
+
+<p role="status">Current count: @State?.CurrentCount</p>
+
+<button class="btn btn-primary" @onclick="IncrementCount">Click me</button>
+
+@code {
+    [SupplyParameterFromPersistentComponentState]
+    public CounterState? State { get; set; }
```
danroth27: Can the persisted state just be an `int`? Why the wrapper object?

aspnetcore/blazor/state-management/persistent-component-state.md
guardrex: 🤔 ... I thought I did that for a reason ... maybe enhanced nav fouled its behavior. I'll check and see why it's there.

aspnetcore/blazor/state-management/persistent-component-state.md
```diff
@@ -0,0 +1,355 @@
...
+> If the app adopts [interactive routing](xref:blazor/fundamentals/routing#static-versus-interactive-routing) and the page is reached via an internal [enhanced navigation](xref:blazor/fundamentals/routing#enhanced-navigation-and-form-handling), prerendering doesn't occur. Therefore, you must perform a full page reload for the component to see the following output. For more information, see the [Interactive routing and prerendering](#interactive-routing-and-prerendering) section.
+
+> :::no-loc text="info: BlazorSample.Components.Pages.PrerenderedCounter2[0]":::  
+> :::no-loc text="      CurrentCount set to 96":::  
+> :::no-loc text="info: BlazorSample.Components.Pages.PrerenderedCounter2[0]":::  
+> :::no-loc text="      CurrentCount restored to 96":::
+
+In the following example that serializes state for multiple components of the same type:
+
+* Properties annotated with the `[SupplyParameterFromPersistentComponentState]` attribute are serialized and deserialized during prerendering.
```
danroth27:
```suggestion
* Properties annotated with the `[SupplyParameterFromPersistentComponentState]` attribute are serialized during prerendering.
```

aspnetcore/blazor/state-management/persistent-component-state.md
guardrex: There was a bug ... Javier fixed it for Preview 7. I'll square it away on the upcoming docs Pre7 PR later this week.

aspnetcore/blazor/state-management/persistent-component-state.md
guardrex: Resolved by https://github.com/dotnet/blazor-samples/pull/568.

aspnetcore/blazor/state-management/server.md
```diff
@@ -0,0 +1,239 @@
...
+* Table storage
+
+After data is saved, the user's state is retained and available in any new circuit.
+
+For more information on Azure data storage options, see the following:
+
+* [Azure Databases](https://azure.microsoft.com/product-categories/databases/)
+* [Azure Storage Documentation](/azure/storage/)
+
+## URL
```
danroth27: Does this section on using the URL to track state really need to be Server/WebAssembly specific? The content seems to be pretty much the same for both. Could this live in the state management overview instead?

aspnetcore/blazor/state-management/browser-storage.md
```diff
@@ -612,103 +505,6 @@ To persist many different state objects and consume different subsets of objects
 
 :::zone pivot="webassembly"
```
danroth27: I don't think we actually need this pivot for Blazor WebAssembly. Most of the content for Blazor WebAssembly is just documenting the browser platform storage features, which we don't provide any built-in framework support for.

aspnetcore/blazor/state-management/browser-storage.md
```diff
@@ -1,125 +1,18 @@
...
-
-[!INCLUDE[](~/includes/not-latest-version.md)]
-
-This article describes common approaches for maintaining a user's data (state) while they use an app and across browser sessions.
-
-> [!NOTE]
-> The code examples in this article adopt [nullable reference types (NRTs) and .NET compiler null-state static analysis](xref:migration/50-to-60#nullable-reference-types-nrts-and-net-compiler-null-state-static-analysis), which are supported in ASP.NET Core in .NET 6 or later. When targeting .NET 5 or earlier, remove the null type designation (`?`) from types in the article's examples.
-
-## Maintain user state
+# ASP.NET Core Blazor state management using browser storage
```
danroth27: I think this article should really just focus on the Blazor framework support for protected browser storage:

```suggestion
# ASP.NET Core Blazor protected browser storage
```

aspnetcore/blazor/state-management/server.md
```diff
@@ -0,0 +1,239 @@
...
+
+If a user experiences a temporary network connection loss, Blazor attempts to reconnect the user to their original circuit with their original state. However, reconnecting a user to their original circuit in the server's memory isn't always possible:
+
+* The server can't retain a disconnected circuit forever. The server must release a disconnected circuit after a timeout or when the server is under memory pressure.
+* In multi-server, load-balanced deployment environments, individual servers may fail or be automatically removed when no longer required to handle the overall volume of requests. The original server processing requests for a user may become unavailable when the user attempts to reconnect.
+* The user might close and reopen their browser or reload the page, which removes any state held in the browser's memory. For example, JavaScript variable values set through JavaScript interop calls are lost.
+
+When a user can't be reconnected to their original circuit, the user receives a new circuit with an empty state. This is equivalent to closing and reopening a desktop app.
```
danroth27:
```suggestion
When a user can't be reconnected to their original circuit, the user receives a new circuit with newly initialized state. This is equivalent to closing and reopening a desktop app.
```

aspnetcore/blazor/state-management/server.md
```diff
@@ -0,0 +1,239 @@
...
+
+:::moniker range=">= aspnetcore-10.0"
+
+## Circuit state (and prerendering state) preservation
```
danroth27: I think we're covering state persistence during prerendering separately.

```suggestion
## Circuit state persistence
```

danroth27: There's a lot of front matter in this section that covers a lot of details before covering the mainline scenario of attributing a component so that it's state can be persisted. Can we condense this content and order it so that the most common usage is covered first and configuration options and nuanced guidance are covered later?

aspnetcore/blazor/state-management/server.md
```diff
@@ -0,0 +1,239 @@
...
+    public List<Item> Items { get; set; }
+
+    protected override async Task OnInitializedAsync()
+    {
+        Items ??= await LoadItemsAsync();
+    }
+}
+```
+
+To persist State for scoped services, annotate service properties with `[SupplyParameterFromPersistentComponentState]`, add the service to the service collection, and call the <xref:Microsoft.Extensions.DependencyInjection.RazorComponentsRazorComponentBuilderExtensions.RegisterPersistentService%2A> extension method with the service:
```
danroth27:
```suggestion
To persist state for scoped services, annotate service properties with `[SupplyParameterFromPersistentComponentState]`, add the service to the service collection, and call the <xref:Microsoft.Extensions.DependencyInjection.RazorComponentsRazorComponentBuilderExtensions.RegisterPersistentService%2A> extension method with the service:
```

aspnetcore/blazor/state-management/server.md
```diff
@@ -0,0 +1,239 @@
...
+window.addEventListener('visibilitychange', () => {
+  if (document.visibilityState === 'hidden') {
+    Blazor.pauseCircuit();
+  } else if (document.visibilityState === 'visible') {
+    Blazor.resumeCircuit();
+  }
+});
+```
+
+## Declarative model for persisting state using a DI service
```
danroth27: I think state persistence for services was covered earlier in this doc (line 133). Maybe this section can be removed?

aspnetcore/blazor/state-management/server.md
```diff
@@ -0,0 +1,239 @@
...
+services.AddHybridCache()
+    .AddRedis("{CONNECTION STRING}");
+
+services.AddRazorComponents()
+    .AddInteractiveServerComponents();
+```
+
+In the preceding example, the `{CONNECTION STRING}` placeholder represents the Redis cache connection string, which should be provided using a secure approach, such as the [Secret Manager](xref:security/app-secrets#secret-manager) tool in the Development environment or [Azure Key Vault](/azure/key-vault/) with [Azure Managed Identities](/entra/identity/managed-identities-azure-resources/overview) for Azure-deployed apps in any environment.
+
+To proactively pause and resume circuits in custom resource management scenarios, call `Blazor.pauseCircuit` and `Blazor.resumeCircuit` from a JavaScript event handler. In the following example, changes in the the visibility of the app either pause or resume the user's circuit:
```
danroth27: I think this section on proactively pausing and resuming circuits merits its own section heading and should be expanded a bit. For the section heading, how about "Pause and resume circuits"?

We should explain a bit more about what it means to pause a circuit. Pausing a circuit will cause it to persist and the data will be stored *client-side* using browser storage. The circuit will then be evicted, freeing up the server resources. Resuming the circuit will then establish a new circuit and initialize it using the persisted state.

The intended use of these APIs is to enable the implementation of custom policies that improve the scalability of your app. The example show is a good one - if the app isn't visible then the circuit probably isn't currently needed.

aspnetcore/blazor/state-management/persistent-component-state.md
danroth27: Actually, now that I've read the article more fully, I've found it to be exclusively about persisting prerendered state, not a general state persistence overview. Should we entitle it as such?

```suggestion
# ASP.NET Core Blazor prerendered state persistence
```

aspnetcore/blazor/components/prerender.md
danroth27, line 1: It looks like there's not much left in this prerendering article with all the state persistence content removed 😆.

Can we add some content to this article to describe what prerendering is? Basically, we statically render the page from the server to deliver HTML to the browser as quickly as possible. Then any interactive content has to render again, replacing any content that was rendered previously. All our interactive render modes prerender by default, but you can turn prerendering off if you want to (show how to do that). As is the case in general with static server-server side renderin, the `OnAfterRender`/`OnAfterRenderAsync` component lifecycle event does NOT get called when prerendering. It will get called only after the component renders interactively.

I really dislike the giant deeply nested bulleted list of links at the end of the prerendering article 😝. It seems silly that someone would need to click all those links to see the related prerendering guidance. I think we should summarize the related and important prerendering guidance directly in the article so that it's easier to digest. We don't have to do that as part of this PR, but we should track cleaning that up.

danroth27 (changes requested):
Nice start! We should get a technical review from the engineering team as well once we get the structure worked out.

aspnetcore/blazor/components/prerender.md
guardrex: > I really dislike the giant deeply nested bulleted list of links at the end of the prerendering article

> It seems silly that someone would need to click all those links to see the related prerendering guidance.

That's not the intention. This is an organized index of prerendering coverage. The only way to find all of the prerendering guidance without an index is to use search. Relying on our documentation search is poor because AFAIK readers can't filter results to the Blazor node, and the results are disorganized.

**UPDATE**: I checked our search. It only sprinkles in a few non-Blazor node links here and there. It's still quite disorganized, however, compared to the index here.

Confirm that you want the index 🔪, and I'll 💀 it.

guardrex: If you decide to 🔪 the index of prerendering coverage (link list), then I think we should move the little bit that remains to the *Fundamentals* overview. There's already some language there on it, and we could just enhance that coverage a bit. Then, we'd 💀 this *Prerendering* article.

If you want to keep the index of cross-links, which I still think has value for folks when they're trying to find something in particular, then I think we keep the *Prerendering* article.

guardrex:
@danroth27 ... I'm done with the first-round updates.

Still need an answer on ...

* Keep the prerendering index ... and keep the *Prerendering* article.
* 💀 the index ... and 💀 the *Prerendering* article ... and integrate the bits that are left into the *Fundamentals* overview, where there are some [existing remarks](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/?view=aspnetcore-9.0#client-and-server-rendering-concepts).

aspnetcore/blazor/components/prerender.md
danroth27: I think we keep the prerendering doc. Prerendering is important enough and complicated enough that I think it merits its own article. I'd like to consolidate the prerendering related content and guidance into the prerendering doc and then link to it from elsewhere. Right now, we're doing the opposite: we've sprinkled prerendering related guidance throughout the docs and then link to those many locations from the prerendering doc. I think that makes it really hard to get a complete picture of prerendering concerns without clicking on a lot of links. 

There's a pretty substantial [section on prerendering in the render modes doc](https://review.learn.microsoft.com/aspnet/core/blazor/components/render-modesprerendering). I think that section covers a good chunk of what I'd expect to be covered in the prerendering doc. I think we should move most of that content into the prerendering doc. We should still have the prerendering section in the render modes doc, but it can be short and say something like: "The interactive render modes (InteractiveServer, InteractiveWebAssembly, InteractiveAut) all support prerendering by default, which initially renders the page content statically from the server to improve the initial load experience. See Prerendering for additional details." I think we do a similar refactoring for all the prerendering guidance across our docs.

guardrex: Because new-to-.NET and new-to-Blazor devs read the docs like a book ... down the ToC ... the *Prerendering* article would either have to spend a lot of time explaining features before they're covered or the *Prerendering* article has to appear very nearly at the end of all of the coverage.

Even if the article did explain enough of each feature to introduce it (and then describe prerendering factors/effects), the reverse "clicking a lot of links" problem appears: Readers must click away from the feature coverage section that they're reading to a section in the *Prerendering* article for the related content (unless covered in both spots). Personally, I'd rather hear about prerendering's effects for a feature right where the feature is discussed.

Whatever we do tho, I recommend that this isn't the right PR for a major shuffle. I'll work on the prerendering coverage between the *Render Modes* article and the *Prerendering* article on this PR. I think we should wait for anything else. I'm going to need to switch over to Pre7 coverage soon, and this PR really needs to clear the way for that ... ***and*** I'm buried in the static files overhaul right now, too. I propose we at least wrap this state management PR up close to its present form, and then I'll open a new issue to discuss changes for prerendering coverage.

danroth27: > the Prerendering article would either have to spend a lot of time explaining features before they're covered or the Prerendering article has to appear very nearly at the end of all of the coverage.

There's probably a balance here of consolidating the main prerendering concerns in one place and having some targeted prerendering related notes in other sections.

> I propose we at least wrap this state management PR up close to its present form, and then I'll open a new issue to discuss changes for prerendering coverage.

Agreed, we should track the prerendering discussion separately.

guardrex: By EOD, I'll see what I can do here with the *Render Modes* and *Prerendering* article coverage. I'll ping back by EOD when that commit goes in.

Otherwise, even the other updates here will need a look. I took a decent amount of your language for the last commit, but the state management coverage probably still needs a bit more work.

guardrex: @danroth27 ... The last two commits move ...

* The general prerendering coverage, primarily focused on disabling it, from the *Render Modes* article to the *Prerender* article.
* The *Client-side services fail to resolve during prerendering* section to the *Prerender* article.

I went ahead and moved that second one. It seemed like the right thing to do given the direction we're going with the *Render Modes* and *Prerendering* articles.

If we're going to have one or more engineers look, can we do it ... well ... ***now?!*** 😄 I need this merged to get into the Pre7 docs work.

aspnetcore/blazor/components/prerender.md
```diff
@@ -17,358 +17,134 @@ uid: blazor/components/prerender
...
 
-The first logged count occurs during prerendering. The count is set again after prerendering when the component is rerendered. There's also a flicker in the UI when the count updates from 41 to 92.
+                 "prerender: false" is ignored in child components
+                 https://github.com/dotnet/aspnetcore/issues/55635
 
-To retain the initial value of the counter during prerendering, Blazor supports persisting state in a prerendered page using the <xref:Microsoft.AspNetCore.Components.PersistentComponentState> service (and for components embedded into pages or views of Razor Pages or MVC apps, the [Persist Component State Tag Helper](xref:mvc/views/tag-helpers/builtin-th/persist-component-state-tag-helper)).
+                 ... for .NET 11 work in the following area. -->
 
-:::moniker range=">= aspnetcore-10.0"
+Disabling prerendering using the following techniques only takes effect for top-level render modes. If a parent component specifies a render mode, the prerendering settings of its children are ignored.
```
danroth27: This is an important detail, but I don't think it should be the first sentence in this section. Can we move it somewhere near the end of the section instead?

aspnetcore/blazor/components/prerender.md
```diff
@@ -17,358 +17,134 @@ uid: blazor/components/prerender
...
+Making a root component, such as the `App` component, interactive with the `@rendermode` directive at the top of the root component's definition file (`.razor`) isn't supported. Therefore, prerendering can't be disabled directly by the `App` component.
 
-* Properties annotated with the `[SupplyParameterFromPersistentComponentState]` attribute are serialized and deserialized during prerendering.
-* The [`@key` directive attribute](xref:blazor/components/key#use-of-the-key-directive-attribute) is used to ensure that the state is correctly associated with the component instance.
-* The `Element` property is initialized in the [`OnInitialized` lifecycle method](xref:blazor/components/lifecycle#component-initialization-oninitializedasync) to avoid null reference exceptions, similarly to how null references are avoided for query parameters and form data.
+:::moniker-end
 
-`PersistentChild.razor`:
+[`OnAfterRender{Async}` component lifecycle events](xref:blazor/components/lifecycle#after-component-render-onafterrenderasync) aren't called when prerendering, only after the component renders interactively.
```
danroth27: This doesn't seem related to disabling prerendering. Maybe move this bit to the end of the intro section?

aspnetcore/blazor/components/prerender.md
danroth27: Ideally, I think we'd start this section explaining why you might want to disable prerendering. Prerendering can complicate your app because your components have to render twice: once for prerendering and once for setting up interactivity. If the components are set up to run on WebAssembly, then you also have to design your components so that they can run from both the server and the client.

aspnetcore/blazor/components/prerender.md
```diff
@@ -17,358 +17,134 @@ uid: blazor/components/prerender
...
+> :::no-loc text="Cannot provide a value for property 'Environment' on type 'BlazorSample.Client.Pages.Home'. There is no registered service of type 'Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment'.":::
 
-The following counter component example persists counter state during prerendering and retrieves the state to initialize the component.
+This error occurs because the component must compile and execute on the server during prerendering, but <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> isn't a registered service on the server.
```
danroth27, line 104: Yeah, it's a bit annoying that we don't have a common service interface for getting the environment that works on both the server and client.

aspnetcore/blazor/components/prerender.md
```diff
@@ -17,358 +17,134 @@ uid: blazor/components/prerender
...
 
-`PrerenderedCounter3.razor`:
+If the app doesn't require the value during prerendering, this problem can be solved by injecting <xref:System.IServiceProvider> to obtain the service instead of the service type itself:
```
danroth27, line 106: If you just want to make the service injection optional, you can use constructor injection instead (someday we'll get [optional property injection](https://github.com/dotnet/razor/issues/7653)):

```csharp
public Home(IWebAssemblyHostEnvironment? env = null)
{
    environmentName = env?.Environment;
}
```

To get the environment whether you're running on the server or the client, you could optionally inject `IHostEnvironment` from Microsoft.Extensions.Hosting.Abstractions. You'll have to add the package reference to the link, but the assembly is trimmable, so that should be fine:

```csharp
public Home(IHostEnvironment? serverEnv = null, IWebAssemblyHostEnvironment? wasmEnv = null)
{
    environmentName = serverEnv?.EnvironmentName;
    environmentName ??= wasmEnv?.Environment;
}
```

aspnetcore/blazor/components/prerender.md
```diff
@@ -17,358 +17,134 @@ uid: blazor/components/prerender
...
 
-:::moniker range="< aspnetcore-10.0"
+There are a three approaches that you can take to address this scenario. The following are listed from most recommended to least recommended:
 
-To preserve prerendered state, decide what state to persist using the <xref:Microsoft.AspNetCore.Components.PersistentComponentState> service. <xref:Microsoft.AspNetCore.Components.PersistentComponentState.RegisterOnPersisting%2A?displayProperty=nameWithType> registers a callback to persist the component state during prerendering. The state is retrieved when the component renders interactively. Make the call at the end of initialization code in order to avoid a potential race condition during app shutdown.
+* *Recommended* for shared framework services: For shared framework services that merely aren't registered server-side in the main project, register the services in the main project, which makes them available during prerendering. For an example of this scenario, see the guidance for <xref:System.Net.Http.HttpClient> services in <xref:blazor/call-web-api?pivots=webassembly#client-side-services-for-httpclient-fail-during-prerendering>.
```
danroth27: The anchor linked to in the call a web API doc doesn't seem to exist. Were you trying to link to xref:blazor/call-web-api#service-abstractions-for-web-api-calls instead? That section recommends using a service abstraction, which is good. But that doesn't seem related to shared framework services like IWebAssemblyHostEnvironment unless you want to create your own interface abstraction for getting the environment.

aspnetcore/blazor/components/prerender.md
```diff
@@ -17,358 +17,134 @@ uid: blazor/components/prerender
...
 
-The following counter component example persists counter state during prerendering and retrieves the state to initialize the component.
+* *Recommended* for services outside of the shared framework: Create a custom service implementation for the service on the server. Use the service normally in interactive components of the `.Client` project. For a demonstration of this approach, see <xref:blazor/fundamentals/environments#read-the-environment-client-side-in-a-blazor-web-app>.
```
danroth27: Implementing `IWebAssemblyHostEnvironment` on the server seems kinda icky to me, but it does work. I didn't realize we had so much content about how to get the environment from a component in a hosting model agnostic way. Maybe we should just address this in the framework.

aspnetcore/blazor/components/prerender.md
```diff
@@ -17,358 +17,134 @@ uid: blazor/components/prerender
...
 
-`PrerenderedCounter2.razor`:
+* Create a service abstraction and create implementations for the service in the `.Client` and server projects. Register the services in each project. Inject the custom service in the component.
```
danroth27: This is the way in general 👍.

aspnetcore/blazor/components/prerender.md
```diff
@@ -17,358 +17,134 @@ uid: blazor/components/prerender
...
-When the component executes, `currentCount` is only set once during prerendering. The value is restored when the component is rerendered. The following is example output.
-
-> [!NOTE]
-> If the app adopts [interactive routing](xref:blazor/fundamentals/routing#static-versus-interactive-routing) and the page is reached via an internal [enhanced navigation](xref:blazor/fundamentals/routing#enhanced-navigation-and-form-handling), prerendering doesn't occur. Therefore, you must perform a full page reload for the component to see the following output. For more information, see the [Interactive routing and prerendering](#interactive-routing-and-prerendering) section.
-
-> :::no-loc text="info: BlazorSample.Components.Pages.PrerenderedCounter2[0]":::  
-> :::no-loc text="      currentCount set to 96":::  
-> :::no-loc text="info: BlazorSample.Components.Pages.PrerenderedCounter2[0]":::  
-> :::no-loc text="      currentCount restored to 96":::
+* You might be able to add a `.Client` project package reference to a server-side package and fall back to using the server-side API when prerendering on the server.
```
danroth27: This approach seems risky to recommend. I see a lot of Blazor WebAssembly mistakenly reference server assemblies from their client projects resulting in large client apps. I think we should remove this one.

```suggestion
```

aspnetcore/blazor/state-management/prerendered-state-persistence.md
```diff
@@ -0,0 +1,355 @@
...
+Consider the following `PrerenderedCounter1` counter component. The component sets an initial random counter value during prerendering in [`OnInitialized` lifecycle method](xref:blazor/components/lifecycle#component-initialization-oninitializedasync). When the component then renders interactively, the initial count value is replaced when `OnInitialized` executes a second time.
+
+`PrerenderedCounter1.razor`:
+
+:::code language="razor" source="~/../blazor-samples/8.0/BlazorSample_BlazorWebApp/Components/Pages/PrerenderedCounter1.razor":::
+
+Run the app and inspect logging from the component. The following is example output.
+
+> [!NOTE]
+> If the app adopts [interactive routing](xref:blazor/fundamentals/routing#static-versus-interactive-routing) and the page is reached via an internal [enhanced navigation](xref:blazor/fundamentals/routing#enhanced-navigation-and-form-handling), prerendering doesn't occur. Therefore, you must perform a full page reload for the `PrerenderedCounter1` component to see the following output. For more information, see the [Interactive routing and prerendering](#interactive-routing-and-prerendering) section.
```
danroth27: The placement of this note makes it read like the note is the example output. Suggested new placement:

```suggestion
> [!NOTE]
> If the app adopts [interactive routing](xref:blazor/fundamentals/routing#static-versus-interactive-routing) and the page is reached via an internal [enhanced navigation](xref:blazor/fundamentals/routing#enhanced-navigation-and-form-handling), prerendering doesn't occur. Therefore, you must perform a full page reload for the `PrerenderedCounter1` component to see the following output. For more information, see the [Interactive routing and prerendering](#interactive-routing-and-prerendering) section.

Run the app and inspect logging from the component. The following is example output.
```

aspnetcore/blazor/state-management/prerendered-state-persistence.md
```diff
@@ -0,0 +1,355 @@
...
+
+:::moniker range=">= aspnetcore-10.0"
+
+<!-- UPDATE 10.0 - API cross-links -->
+
+To preserve prerendered state, use the `[SupplyParameterFromPersistentComponentState]` attribute to persist state in properties. Properties with this attribute are automatically persisted using the <xref:Microsoft.AspNetCore.Components.PersistentComponentState> service during prerendering. The state is retrieved when the component renders interactively or the service is instantiated.
+
+By default for client-side prerendering in a Blazor Web App, properties are serialized using the <xref:System.Text.Json?displayProperty=fullName> serializer with default settings. Serialization isn't trimmer safe and requires preservation of the types used. For more information, see <xref:blazor/host-and-deploy/configure-trimmer>.
```
danroth27: I'm not sure with "client-side prerendering" refers to here. Prerendering happens on the server. How about this?:

```suggestion
By default, properties are serialized using the <xref:System.Text.Json?displayProperty=fullName> serializer with default settings and persisted in the prerendered HTML. Serialization isn't trimmer safe and requires preservation of the types used. For more information, see <xref:blazor/host-and-deploy/configure-trimmer>.
```

aspnetcore/blazor/state-management/protected-browser-storage.md
```diff
@@ -0,0 +1,499 @@
...
+```csharp
+protected override async Task OnInitializedAsync()
+{
+    currentCount = await ProtectedSessionStore.GetAsync<int>("count");
+}
+```
+
+:::moniker-end
+
+If the component's parameters include navigation state, call `ProtectedSessionStore.GetAsync` and assign a non-`null` result in <xref:Microsoft.AspNetCore.Components.ComponentBase.OnParametersSetAsync%2A>, not <xref:Microsoft.AspNetCore.Components.ComponentBase.OnInitializedAsync%2A>. <xref:Microsoft.AspNetCore.Components.ComponentBase.OnInitializedAsync%2A> is only called once when the component is first instantiated. <xref:Microsoft.AspNetCore.Components.ComponentBase.OnInitializedAsync%2A> isn't called again later if the user navigates to a different URL while remaining on the same page. For more information, see <xref:blazor/components/lifecycle>.
```
danroth27, line 140: I'm not sure I get why having navigation state passed as parameters changes when you should call `ProtectedSessionStore.GetAsync`. Does accessing session state somehow depend on the navigation state?

aspnetcore/blazor/state-management/protected-browser-storage.md
```diff
@@ -0,0 +1,499 @@
...
+* An interactive connection to the user's browser doesn't exist.
+* The browser doesn't yet have a page in which it can run JavaScript code.
+
+`localStorage` or `sessionStorage` aren't available during prerendering. If the component attempts to interact with storage, an error is generated explaining that JavaScript interop calls cannot be issued because the component is being prerendered.
+
+One way to resolve the error is to disable prerendering. This is usually the best choice if the app makes heavy use of browser-based storage. Prerendering adds complexity and doesn't benefit the app because the app can't prerender any useful content until `localStorage` or `sessionStorage` are available.
+
+:::moniker range=">= aspnetcore-8.0"
+
+To disable prerendering, indicate the render mode with the `prerender` parameter set to `false` at the highest-level component in the app's component hierarchy that isn't a root component.
```
danroth27, line 188: It seems like we should link from here to the section on how to disable prerendering in the prerendering doc.

aspnetcore/blazor/state-management/protected-browser-storage.md
```diff
@@ -0,0 +1,499 @@
...
+
+:::moniker-end
+
+:::moniker range=">= aspnetcore-6.0"
+
+When prerendering is disabled, [prerendering of `<head>` content](xref:blazor/components/control-head-content) is disabled.
+
+:::moniker-end
+
+Prerendering might be useful for other pages that don't use `localStorage` or `sessionStorage`. To retain prerendering, defer the loading operation until the browser is connected to the circuit. The following is an example for storing a counter value:
```
danroth27, line 225: It's good that we show that you can handle access storage in OnAfterRenderAsync. You could also detect whether the current render mode is interactive or not before accessing browser storage. I think that should work even in OnInitializedAsync.

aspnetcore/blazor/state-management/server.md
```diff
@@ -0,0 +1,235 @@
...
+                   Guidance changes for persistent component state with enhanced nav
+                   will be on the upcoming docs PR for Preview 7.
+
+                   API review for support persistent component state on enhanced navigation
+                   https://github.com/dotnet/aspnetcore/issues/62773
+                   -->
+
+*[Enhanced navigation](xref:blazor/fundamentals/routing#enhanced-navigation-and-form-handling) with circuit state persistence isn't currently supported but planned for a future release.*
+
+Persisting state requires fewer server resources than persisting circuits:
```
danroth27: I don't really get what's being compared here. It's not like we have support for persisting entire circuits. I think the intended point is that pausing circuits by persisting and evicting them saves server resources. Maybe something like this?:

```suggestion
Server resources can be freed up if the circuit state can be persisted and then resumed later:
```

danroth27 (approved):
A few suggestions, but otherwise this looks good to me.

aspnetcore/blazor/state-management/protected-browser-storage.md
guardrex: It goes back over five years ago when Steve provided the content ...

> Note that you should instead do this from `OnParametersSetAsync` if your component's parameters include navigation state, since `OnInitializedAsync` will only be called once when the component is first instantiated, and won't be called again later if the user navigates to a different URL while remaining on the same page.

Obviously, there were NIT changes over the years, but the remarks have been reviewed a few times, including by Mackinnon years later.

* https://github.com/dotnet/AspNetCore.Docs/pull/13364
* https://gist.github.com/SteveSandersonMS/ba16f6bb6934842d78c89ab5314f4b56

aspnetcore/blazor/includes/prerendering.md
```diff
@@ -2,8 +2,15 @@
 
 *This section applies to server-side apps that prerender Razor components. Prerendering is covered in <xref:blazor/components/prerender>.*
 
+<!-- UPDATE 10.0 - Persistent component state across enhanced nav
+                   is arriving for Preview 7. 
+                   
+                   The following remarks will be updated/versioned
+                   on the upcoming docs Preview 7 PR.
+                   -->
+
```
javiercn: Arriving for RC1

guardrex:
@javiercn ... I'm working on Dan's changes now, ~so there will be another commit coming in maybe an hour. If you want me to hold here while you look at the PR, I'll stop and wait for your feedback.~

Actually, that's the best thing to do. I'll stop here and wait for your feedback.

I guess you're not reviewing now. I'll pick back up with the work on Dan's suggestions.

aspnetcore/blazor/components/prerender.md
guardrex: The section moved to this article. I'll fix it on the next commit.

> \[Client-side services fail to resolve during prerendering](#client-side-services-for-httpclient-fail-during-prerendering)

aspnetcore/blazor/state-management/protected-browser-storage.md
guardrex: Immediately after this merges, I'll open an issue on this because it's going to require an example. There's no time left to flesh this out right now 🏃.

aspnetcore/blazor/state-management/protected-browser-storage.md
danroth27: OK. The statement is certainly generally correct. I just don't see the relationship to browser storage specifically or to the specific counter example shown. I guess the assumption is that you often use browser storage to setup state, which might also require navigation state. Let's just leave the text as is.

guardrex:
@danroth27 @javiercn ... I just reacted to all of the feedback thus far with the two exceptions for follow-up issues:

* Prerendering content from around the repo that can move (or be duplicated) in the *Prerendering* article to give the *Prerendering* article more of an overview quality. I still think the index is helpful to finding guidance on prerendering, but I'll 🔪 it for that work when the time comes if you still don't like it.
* A section or add-on to that existing section on "detect whether the current render mode is interactive or not before accessing browser storage." I think it will require an example, and I'm *very* out of time this week with Pre7 breathing down my neck 🏃‍♂️.

guardrex:
Hold a sec ... one more problem to fix. 😈

guardrex:
Ok ... I think that fixes the 😈 .

BTW @danroth27 ... WRT the three recommendations ...

> There are a three approaches that you can take to address this scenario for prerendering. The following are listed from most recommended to least recommended:
>
> \* \*Recommended\* for shared framework services: For shared framework services that merely aren't registered server-side in the main project, register the services in the main project, which makes them available during prerendering. For an example of this scenario, see the guidance for \<xref:System.Net.Http.HttpClient> services in the \[Blazor Web App external web APIs](xref:blazor/call-web-api#blazor-web-app-external-web-apis) section of the \*Call web API\* article.
>
> \* \*Recommended\* for services outside of the shared framework: Create a custom service implementation for the service on the server. Use the service normally in interactive components of the \`.Client\` project. For a demonstration of this approach, see the example earlier in this section.
>
> \* Create a service abstraction and create implementations for the service in the \`.Client\` and server projects. Register the services in each project. Inject the custom service in the component.

That came about in a discussion with Mackinnon. This is a good time to change that sequence and which ones are marked as "Recommended" 👂.

aspnetcore/blazor/components/prerender.md
```diff
@@ -17,357 +17,157 @@ uid: blazor/components/prerender
...
+{
+    environmentName = serverEnvironment?.EnvironmentName;
+    environmentName ??= wasmEnvironment?.Environment;
+}
 ```
 
-## Interactive routing and prerendering
-
-When the `Routes` component doesn't define a render mode, the app is using per-page/component interactivity and navigation. Using per-page/component navigation, internal&dagger; navigation is handled by [enhanced routing](xref:blazor/fundamentals/routing#enhanced-navigation-and-form-handling) after the app becomes interactive. &dagger;*Internal* in this context means that the URL destination of the navigation event is a Blazor endpoint inside the app.
+However, the preceding approaches aren't useful if your logic requires a value during prerendering.
```
danroth27: By optionally injecting both `IHostEnvironment` and `IWebAssemblyHostEnvironment` you can get the environment both when prerendering and when interactive.

```suggestion
```

aspnetcore/blazor/components/prerender.md
```diff
@@ -17,357 +17,157 @@ uid: blazor/components/prerender
...
 
-Additional work on the Blazor framework to address this scenario is under consideration for .NET 10 (November, 2025). For more information and community discussion of *unsupported workarounds*&Dagger;, see [Support persistent component state across enhanced page navigations (`dotnet/aspnetcore` #51584)](https://github.com/dotnet/aspnetcore/issues/51584). &Dagger;Unsupported workarounds aren't sanctioned by Microsoft for use in Blazor apps. *Use third-party packages, approaches, and code at your own risk.*
+* *Recommended* for services outside of the shared framework: Create a custom service implementation for the service on the server. Use the service normally in interactive components of the `.Client` project. For a demonstration of this approach, see the example earlier in this section.
```
danroth27: I don't really understand this recommendation. What example is it referring to?

aspnetcore/blazor/components/prerender.md
```diff
@@ -17,357 +17,157 @@ uid: blazor/components/prerender
...
 
-Disabling enhanced navigation, which reduces performance but also avoids the problem of loading state with <xref:Microsoft.AspNetCore.Components.PersistentComponentState> for internal page requests, is covered in <xref:blazor/fundamentals/routing#enhanced-navigation-and-form-handling>.
+* Create a service abstraction and create implementations for the service in the `.Client` and server projects. Register the services in each project. Inject the custom service in the component.
```
danroth27: Honestly, this is the most correct and flexible solution. One minor tweak:

```suggestion
* Create a service abstraction and create implementations for the service in the `.Client` and server projects. Register the services in each project. Inject the custom service abstraction in the component.
```

danroth27 (approved):
Since we're out of time, let's go ahead and get this merged. We can follow up with any feedback from @javiercn later.

aspnetcore/blazor/components/prerender.md
guardrex: Ok ... yeah ... there's so much stuff moving around here that this was lost for a sec.

This isn't in this article. It's over in *Environments* at ...

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/environments?view=aspnetcore-9.0#read-the-environment-client-side-in-a-blazor-web-app

I'll add the correct cross-link back to this one.

aspnetcore/blazor/components/prerender.md
guardrex: I attempted to research it back, but things have moved around several times over the years. I see where Mackinnon reviewed these back when they were moved from the *Render Modes* article to this article, but he didn't comment on these recommendations. I don't see a discussion there. I think I did discuss this with him, and it might have been in an email. The simplest thing to do is just figure out what they should be for an update now/soon.

danroth27: I think the available approaches are:

1. Register the same service on both the server and client if it supports both.
2. Make the service optional if it isn't always needed.
3. Create a service abstraction and implement separate implementations for the server and client.

guardrex: I just reorganized ... ***possibly*** to the tune of what you wrote. I'll check.

guardrex: Nope ... that's not what I have. I need to re-order again. 😩

guardrex: Ok ... the order is ***closer*** to what you said, but you left out Mackinnon's "custom service implementation for the service on the server" approach, which I now have in the last position.

The example is the one over in the *Environments* article ...

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/environments?view=aspnetcore-9.0#read-the-environment-client-side-in-a-blazor-web-app

The PR is updated now with what I think you want.

danroth27: Ah, that's the example where we suggest implementing and injecting an IWebAssemblyHostEnvironment service for the server, which is a bit confusing given that you're not actually running on WebAssembly. But it does work, so I guess that's fine.

guardrex: When we reach the next issue on prerendering, this bit is going to have to be reorganized. ***It's a bit of a mess!*** 🙈😆 Each of these should be in a section with their examples. This is a case where content in another article should move, specifically the *Environments* example should move back to this article for sure.

Anyway ... no time right now. I have to flip over to Pre7 work.

terminator

&mdash; **_"I'll be back."_** - Cyberdyne Systems T-800 Series Model 101 ([Arnold Schwarzenegger](https://www.imdb.com/name/nm0000216/))
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[*The Terminator*](https://www.imdb.com/title/tt0088247/)  1984 [Skydance Media](https://skydance.com/)

PR #35952, opened 2025-08-15
Surface .dev.localhost TLD Visual Studio option

### LINKED ISSUE #35936: [10.0] Blazor tutorial VS setup guidance: Add .dev.localhost TLD option
As of .NET 10, there's a new entry in the UI for the `.dev.localhost` TLD to add to VS guidance ...

```
* **Use the .dev.localhost TLD in the application URL**: Not selected
```

guardrex:
Fixes #35936

We don't (usually) version VS tooling guidance. We assume that the dev is using the latest tooling. This updates our guidance to include coverage for the new "**Use the .dev.localhost TLD in the application URL**" option. I made the update across the repo, using whatever styling the doc author had in place.

This is in VS Preview. I'm not aware if it has made its way into VS GA. I run 100% of the time on VS Preview these days, so can either of you confirm if it's in VS GA yet? If it hasn't reached VS GA yet, we can sit on this PR or merge it now because VS GA should get it soon. Personally, I'd like to clear this through ... I don't like *merge conflicts* 😈 if they can be avoided.

image

### DIFF
--- a/aspnetcore/release-notes/aspnetcore-7.0.md
+++ b/aspnetcore/release-notes/aspnetcore-7.0.md
@@ -669,6 +669,8 @@ With Visual Studio, select the new **Do not use top-level statements** checkbox 
 
 ![checkbox](https://user-images.githubusercontent.com/3605364/180587645-90f7cce5-d9f8-49d2-88cf-2258960394e1.png)
 
+Verify that **Use the .dev.localhost TLD in the application URL** is unchecked.
+
 ### Updated Angular and React templates
 
 The Angular project template has been updated to Angular 14. The React project template has been updated to React 18.2.
--- a/aspnetcore/blazor/tutorials/movie-database-app/part-1.md
+++ b/aspnetcore/blazor/tutorials/movie-database-app/part-1.md
@@ -59,35 +59,17 @@ In Visual Studio:
 
 * Confirm that the **Location** for the app is suitable. Set the **Place solution and project in the same directory** checkbox to match your preferred solution file location. Select the **Next** button.
 
-:::moniker range=">= aspnetcore-9.0"
-
 * In the **Additional information** dialog, use the following settings:
 
-  * **Framework**: Select **.NET 9.0 (Standard Term Support)**.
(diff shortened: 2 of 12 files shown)


aspnetcore/release-notes/aspnetcore-7.0.md
```diff
@@ -669,6 +669,8 @@ With Visual Studio, select the new **Do not use top-level statements** checkbox
 
 ![checkbox](https://user-images.githubusercontent.com/3605364/180587645-90f7cce5-d9f8-49d2-88cf-2258960394e1.png)
 
+Verify that **Use the .dev.localhost TLD in the application URL** is unchecked.
```
tdykstra: This seems likely to be confusing, as it's on a page that shows a screenshot of top level statements and does not show TLD. However, the page doesn't get much traffic.  I think you could just as well not make any change to this page, but it's up to you.

guardrex: Ah! I should've looked, and I will now look at either updating the image or dropping this change.

guardrex: Same deal with the Min web API tutorial.

I think dropping these out is best. It's a minor bit that can be addressed on the next round of tutorial updates for these articles. I'm buried in work as far as the 👁️ can see.

guardrex:
@tdykstra ... It's done. This is just for the Blazor tutorials then, which don't have screenshots of VS for that bit.

WRT the versioning aspects, I'll wait to hear from @danroth27 and @wadepickett before proceeding.

guardrex:
BTW @tdykstra @wadepickett ... I already messaged Cassie about this but received no response. Would you consider seeing if they'll 🛑 the constant PoliCheck Scan Report comments ***when the scan passes***? It's generating a lot of unnecessary comment churn and email pings.

wadepickett:
> BTW @tdykstra @wadepickett ... I already messaged Cassie about this but received no response. Would you consider seeing if they'll 🛑 the constant PoliCheck Scan Report comments _**when the scan passes**_? It's generating a lot of unnecessary comment churn and email pings.

There was a discussion on this that was going on and still active this morning.  I think as of this morning it may not have been resolved yet but stakeholders understand the problem and they may be close to changing it for the better.

guardrex:
Dan ... if you see this ... I don't need to bother you with this one after all, so I removed your review ping. I'm only going to call out the .dev.localhost TLD VS setting on the two Blazor tutorials.

Issue #35962, opened 2025-08-18
Freshness tier levels for related repos


wadepickett:
Due by August 31

We need to have freshness support levels set.
This can be repo wide set in the docfx.json

Set for all repos related to ASP.NET:
aspnetcore.docs
aspnetdocs
IIS
EF

Metadata:
ms.update-cycle

See C# or .NET docs repos for examples.

---
[Associated WorkItem - 482187](https://dev.azure.com/msft-skilling/Content/_workitems/edit/482187)

PR #35964, opened 2025-08-18
Set Freshness ms.update-cycle

wadepickett:
Contributes to #35962

Setting the metadata for ms.update-cycle: values across the repo


wadepickett:
@tdykstra,  I just set the whole repo for now to the default 365.  If you want different settings per directory we can, but I think for now it is best to just run with this default at the repo level.

Issue #32891, opened 2024-06-19
Static file, multiple locations


ventureaaron:
[EDIT by guardrex to update the metadata]

### Description

In the final blazor example for serving from multiple locations, the compiled css file seems to get lost (appname.styles.css). Debugging the webrootfileprovider seems to indicate the culprit is the Microsoft.AspNetCore.StaticWebAssets.ManifestStaticWebAssetFileProvider dropping out when we replace the webrootprovider with the compositeProvider in the example.

I worked around this with this below, but it's almost certain to not be the proper way:

```csharp
var secondaryProvider = new PhysicalFileProvider(
    builder.Configuration.GetValue<String>("serverDir:HomePath")!);
var oldProvider = app.Environment.WebRootFileProvider;
var compositeProvider = new CompositeFileProvider(oldProvider, secondaryProvider);
app.Environment.WebRootFileProvider = compositeProvider;
```

Cheers on fantastic documentation! Really appreciating all the efforts.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/static-files?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/fundamentals/static-files.md

### Document ID

25458428-030f-6acb-457e-1b4dfa722043

### Article author

@guardrex

guardrex:
Hello @ventureaaron ...

Please open this for the product unit at ...

https://github.com/dotnet/aspnetcore/issues

... with a few updates ...

* Tell them that you're referring to this section for use with a Blazor Web App :point_right: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/static-files?view=aspnetcore-8.0#serve-files-from-multiple-locations
* When you post your code use triple-backticks above and below your code block for formatting in a GH comment with the C# code language ("\`\`\`csharp" on a line before your code ... then your code ... then on another line after with "\`\`\`"). See: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code
* Please add the following to your OP so that I can follow the discussion. If we make a documentation change, it will be to add content on this to the Blazor *Static Files* article ...

  ```
  cc: @guardrex https://github.com/dotnet/AspNetCore.Docs/issues/32891
  ```

@Rick-Anderson ... I'll relabel this issue for the Blazor article and assign to myself for work.

[...]

PR #35967, opened 2025-08-20
Static files overhaul

### LINKED ISSUE #35810: Static assets article(s) updates
### Description

Pick up from round one ...

* ✨ *Blazor-fy!*&trade; ✨ the article.
* Edit pass on the whole doc.
* Combine the two static assets articles into one article.
* Include additional API/config per https://github.com/dotnet/aspnetcore/issues/63292 with coverage notes at https://github.com/dotnet/aspnetcore/issues/59315#issuecomment-3200171263.

WRT the heading ...

> Map Static Assets routing endpoint conventions (`MapStaticAssets`)

First ...

> This heading is incomprehensible; Can it be made clear, maybe by dividing it up into multiple sections?
>
> I think we do need some kind of introduction to this article, something that explains what the web root is and what the content root is, what static assets are, what static files are, what it means to "map to static assets," etc.

Second ...

> Somewhere we need a high level conceptual overview of/intro to "map static assets" vs. "use static files".

Third (Line 139) ...

> Static file authorization seems like a topic that would be more appropriate in this article than in the fundamentals overview

Additional feedback ...

> I don't think the "Serve files outside of web root" section has much relevance to the topic of the page, if any. It just repeats the information on the other page for static files. It doesn't actually tell how to serve assets outside the web root as one could assume, furthermore, it is NOT possible currently with just its own features, so all it does is just sew confusion. The same goes for "Set HTTP response headers", "Serve files from multiple locations" and "Serve files outside wwwroot by updating IWebHostEnvironment.WebRootPath". In fact, the 2d section is wrong still because it is written for static *files* except in the last line of the example code, where it finally invokes `MapStaticAssets()`. And the 3d section confuses the hell out of me, which is shared feedback to both pages, is it really "serving files outside `wwwroot`" if you change its very location? Neither of the pages really need it, and instead it could just be a section in the ASP.NET configuration page about the web root, if there is none right now, I forget.
>
> Also, the title of the page should use the word "assets" instead of "files".

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/map-static-files?view=aspnetcore-10.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/map-static-files.md

### Document ID

927e39a4-a6d4-d5b8-8aa2-4fa2e04ea75b

### Platform Id

22b989a7-6b18-6b75-3391-72de8d1b5a88

### Article author

@guardrex

### Metadata

* ID: 927e39a4-a6d4-d5b8-8aa2-4fa2e04ea75b
* PlatformId: 22b989a7-6b18-6b75-3391-72de8d1b5a88 
* Service: **aspnet-core**
* Sub-service: **fundamentals**

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+927e39a4-a6d4-d5b8-8aa2-4fa2e04ea75b)

### LINKED ISSUE #35833: [static-files] Missing information or sample code content
### Description

Section speaking of here:
https://learn.microsoft.com/de-de/aspnet/core/fundamentals/static-files?view=aspnetcore-9.0#serve-files-outside-wwwroot-by-updating-iwebhostenvironmentwebrootpath

in this part right below the sample code block it is introducing that the wwwroot would only be used in Development, but nothing in the code sample its speaking of, does actually have a if block to make it do that.

> In the preceding code, requests to `/`:
> - In the development environment return `wwwroot/Index.html`
> - In any environment other than development return `wwwroot-custom/Index.html`

### Page URL

https://learn.microsoft.com/de-de/aspnet/core/fundamentals/static-files?view=aspnetcore-9.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/static-files.md

### Document ID

3fec6e08-fc99-7a5c-796f-3f2347cad891

### Platform Id

5c42c2c8-be1b-60de-8cde-f9b8d8c24f5f

### Article author

@Rick-Anderson

### Metadata

* ID: d47f38e2-bc36-cfb8-b0cc-17a534689f8b
* PlatformId: 5c42c2c8-be1b-60de-8cde-f9b8d8c24f5f 
* Service: **aspnet-core**
* Sub-service: **fundamentals**

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+3fec6e08-fc99-7a5c-796f-3f2347cad891)

guardrex:
Fixes #35810
Fixes #35833

After Javier technical review, I'll ping Wade and Tom.
 
### Notes

* Let's combine articles. IMO, it's kind'a scatter-🧠 to have ***two*** static files articles.
* '***Look ma! No broken links!***' 😆 Let's drop the horribad INCLUDES versioning for two great outcomes ...
  * Links aren't broken in prior-version content.
  * The team and community are prevented from only updating the latest version of the article. That will still be a problem if we ever go to group-topic versioning (aka "Version-by-file"), but I don't think that's ever going to happen. If we're on the current versioning model forever, this approach stops the poor practice of only making bug fixes to the latest release content (i.e., INCLUDES files are routinely ignored for bug and enhancement updates).
* Let's move the samples into the `dotnet/AspNetCore.Docs.Samples` repo. The sample apps were placed in that repo on https://github.com/dotnet/AspNetCore.Docs.Samples/pull/282. I'll delete them from this repo on this PR ***after*** reviews to avoid cluttering the diff.
* We're still running with the separate [Blazor Static Files article](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/static-files) with Blazor-focused scenarios. Dan, I can move that content (or selected bits of it) into this article if you want. 👂

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/environments.md](https://github.com/dotnet/AspNetCore.Docs/blob/3111f365741233c8c177de6b7e331d7a59fbeedb/aspnetcore/blazor/fundamentals/environments.md) | [aspnetcore/blazor/fundamentals/environments](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/environments?branch=pr-en-us-35967) |
| [aspnetcore/blazor/fundamentals/static-files.md](https://github.com/dotnet/AspNetCore.Docs/blob/3111f365741233c8c177de6b7e331d7a59fbeedb/aspnetcore/blazor/fundamentals/static-files.md) | [aspnetcore/blazor/fundamentals/static-files](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/static-files?branch=pr-en-us-35967) |
| [aspnetcore/fundamentals/static-files.md](https://github.com/dotnet/AspNetCore.Docs/blob/3111f365741233c8c177de6b7e331d7a59fbeedb/aspnetcore/fundamentals/static-files.md) | [aspnetcore/fundamentals/static-files](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/static-files?branch=pr-en-us-35967) |
| [aspnetcore/migration/50-to-60-samples.md](https://github.com/dotnet/AspNetCore.Docs/blob/3111f365741233c8c177de6b7e331d7a59fbeedb/aspnetcore/migration/50-to-60-samples.md) | [aspnetcore/migration/50-to-60-samples](https://review.learn.microsoft.com/en-us/aspnet/core/migration/50-to-60-samples?branch=pr-en-us-35967) |
| [aspnetcore/migration/70-80.md](https://github.com/dotnet/AspNetCore.Docs/blob/3111f365741233c8c177de6b7e331d7a59fbeedb/aspnetcore/migration/70-80.md) | [aspnetcore/migration/70-80](https://review.learn.microsoft.com/en-us/aspnet/core/migration/70-80?branch=pr-en-us-35967) |
| [aspnetcore/toc.yml](https://github.com/dotnet/AspNetCore.Docs/blob/3111f365741233c8c177de6b7e331d7a59fbeedb/aspnetcore/toc.yml) | [aspnetcore/toc](https://review.learn.microsoft.com/en-us/aspnet/core/toc?branch=pr-en-us-35967) |


aspnetcore/release-notes/aspnetcore-10/includes/blazor.md
```diff
@@ -133,7 +133,7 @@ For more information, see [`HttpClient` and `HttpRequestMessage` with Fetch API
...
+Last year, the release of .NET 9 introduced [server-side fingerprinting](https://en.wikipedia.org/wiki/Fingerprint_(computing)) of static assets in Blazor Web Apps with the introduction of [Map Static Assets routing endpoint conventions (`MapStaticAssets`)](xref:fundamentals/static-files), the [`ImportMap` component](xref:blazor/fundamentals/static-files#importmap-component), and the <xref:Microsoft.AspNetCore.Components.ComponentBase.Assets?displayProperty=nameWithType> property (`@Assets["..."]`) to resolve fingerprinted JavaScript modules. For .NET 10, you can opt-into client-side fingerprinting of JavaScript modules for standalone Blazor WebAssembly apps.
```
danroth27:
```suggestion
The release of .NET 9 introduced [server-side fingerprinting](https://en.wikipedia.org/wiki/Fingerprint_(computing)) of static assets in Blazor Web Apps with the introduction of [Map Static Assets routing endpoint conventions (`MapStaticAssets`)](xref:fundamentals/static-files), the [`ImportMap` component](xref:blazor/fundamentals/static-files#importmap-component), and the <xref:Microsoft.AspNetCore.Components.ComponentBase.Assets?displayProperty=nameWithType> property (`@Assets["..."]`) to resolve fingerprinted JavaScript modules. For .NET 10, you can opt-into client-side fingerprinting of JavaScript modules for standalone Blazor WebAssembly apps.
```

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1104 @@
...
+:::moniker range=">= aspnetcore-9.0"
+
+## Map Static Assets (`MapStaticAssets`)
+
+<xref:Microsoft.AspNetCore.Builder.StaticAssetsEndpointRouteBuilderExtensions.MapStaticAssets%2A> maps static files produced during the build as endpoints with the following features:
+
+* An asset is served once until the file changes or the browser clears its cache. The [`ETag`](https://developer.mozilla.org/docs/Web/HTTP/Headers/ETag) and [Last-Modified](https://developer.mozilla.org/docs/Web/HTTP/Headers/Last-Modified) headers are set.
+* The browser is prevented from using stale assets after an app is updated. The [`Last-Modified`](https://developer.mozilla.org/docs/Web/HTTP/Headers/Last-Modified) header is set.
+* The appropriate [caching headers](https://developer.mozilla.org/docs/Web/HTTP/Headers/Cache-Control) are set on the response.
+* [Caching Middleware](xref:performance/caching/middleware) is used.
```
danroth27: The Caching middleware isn't actually a feature of MapStaticAssets. It's a separate independent feature.

```suggestion
```

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1104 @@
...
+* [Compressed](/aspnet/core/performance/response-compression) versions of the assets are served when possible. This optimization doesn't include minification.
+* A [CDN](/microsoft-365/enterprise/content-delivery-networks?view=o365-worldwide&preserve-view=true) can be used to serve the assets closer to the user.
```
danroth27: Whether you decide to use a CDN isn't really specific to MapStaticAssets you could use one with the static files middleware too.

```suggestion
```

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1104 @@
...
+
+## `MapStaticAssets` versus `UseStaticFiles`
+
+<xref:Microsoft.AspNetCore.Builder.StaticAssetsEndpointRouteBuilderExtensions.MapStaticAssets%2A> is available in ASP.NET Core in .NET 9 or later. <xref:Microsoft.AspNetCore.Builder.StaticFileExtensions.UseStaticFiles%2A> must be used in versions prior to .NET 9.
+
+<xref:Microsoft.AspNetCore.Builder.StaticFileExtensions.UseStaticFiles%2A> serves static files, but it doesn't provide the same level of optimization as <xref:Microsoft.AspNetCore.Builder.StaticAssetsEndpointRouteBuilderExtensions.MapStaticAssets%2A>. <xref:Microsoft.AspNetCore.Builder.StaticAssetsEndpointRouteBuilderExtensions.MapStaticAssets%2A> is optimized for serving assets that the app has knowledge of at runtime. If the app serves assets from other locations, such as disk or embedded resources, <xref:Microsoft.AspNetCore.Builder.StaticFileExtensions.UseStaticFiles%2A> should be used.
+
+Map Static Assets provides the following benefits that aren't available when calling <xref:Microsoft.AspNetCore.Builder.StaticFileExtensions.UseStaticFiles%2A>:
```
danroth27: Another benefit of MapStaticAssets is that it is based on endpoint routing, so it works with other endpoint aware features like authorization.

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1104 @@
...
+  * Even if an asset isn't fingerprinted, content based `ETags` are generated for each static asset using the fingerprint hash of the file as the `ETag` value. This ensures that the browser only downloads a file if its content changes (or the file is being downloaded for the first time).
+  * Internally, the framework maps physical assets to their fingerprints, which allows the app to:
+    * Find automatically-generated assets, such as Razor component scoped CSS for Blazor's [CSS isolation feature](xref:blazor/components/css-isolation) and JS assets described by [JS import maps](https://developer.mozilla.org/docs/Web/HTML/Element/script/type/importmap).
+    * Generate link tags in the `<head>` content of the page to preload assets.
+* During [Visual Studio Hot Reload](/visualstudio/debugger/hot-reload) development testing:
+  * Integrity information is removed from the assets to avoid issues when a file is changed while the app is running.
+  * Static assets aren't cached to ensure that the browser always retrieves current content.
```
danroth27: I don't think this is a benefit of MapStaticAssets over the static files middleware. I believe both systems accommodate for iterative development workflows. It looks like we talk about how things are adjusted during development in the control static file caching section. Maybe move this content there?

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1104 @@
...
+
+## Short-circuit the middleware pipeline
+
+To avoid running the entire middleware pipeline after a static asset is served, which matches the behavior of <xref:Microsoft.AspNetCore.Builder.StaticFileExtensions.UseStaticFiles%2A>, call <xref:Microsoft.AspNetCore.Builder.RouteShortCircuitEndpointConventionBuilderExtensions.ShortCircuit%2A> on <xref:Microsoft.AspNetCore.Builder.StaticAssetsEndpointRouteBuilderExtensions.MapStaticAssets%2A>. Calling <xref:Microsoft.AspNetCore.Builder.RouteShortCircuitEndpointConventionBuilderExtensions.ShortCircuit%2A> immediately executes the endpoint and returns the response, preventing other middleware from executing for static asset requests:
+
+```csharp
+app.MapStaticAssets().ShortCircuit();
+```
+
+## Control static file caching
```
danroth27:
```suggestion
## Control static file caching during development
```

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1104 @@
...
+
+During development, the framework overrides cache headers to prevent browsers from caching static files. This helps ensure that the latest version of files are used when files change, avoiding issues with stale content. In production, the correct cache headers are set, allowing browsers to cache static assets as expected.
```
danroth27: Can we be more explicit that were talking about Development vs Product environments?

```suggestion
When running in the Development environment, the framework overrides cache headers to prevent browsers from caching static files. This helps ensure that the latest version of files are used when files change, avoiding issues with stale content. In production, the correct cache headers are set, allowing browsers to cache static assets as expected.
```

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1104 @@
...
+
+To disable this behavior, set `EnableStaticAssetsDevelopmentCaching` to `false` in the Development environment's app setting file (`appsettings.Development.json`).
```
danroth27: Looks like you set this property to true to enable caching during development ([code](https://github.com/dotnet/aspnetcore/blob/main/src/StaticAssets/src/Development/StaticAssetDevelopmentRuntimeHandler.cs#L228-L232)):

```suggestion
To disable this behavior, set `EnableStaticAssetsDevelopmentCaching` to `true` in the Development environment's app setting file (`appsettings.Development.json`).
```

danroth27: I think this property might also be specific to MapStaticAssets, but I'm not 100% sure.

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1104 @@
...
+```csharp
+var builder = WebApplication.CreateBuilder(new WebApplicationOptions
+{
+    Args = args,
+    EnvironmentName = Environments.Staging,
+    WebRootPath = "wwwroot-custom"
+});
+```
+
+When developing a server-side Blazor app and testing locally, see <xref:blazor/fundamentals/static-files#static-files-in-non-development-environments>.
```
danroth27: This content in the Blazor static files article doesn't really seem to be Blazor specific. Does it really add anything that isn't already stated here? The static web assets support in ASP.NET Core is intended to be a generic feature that's not specific to Blazor, so I think most of that content should just be covered here.

guardrex: There were behavioral deltas when the article was set up. As a general principle, only deltas (or Blazor-focused scenarios) were covered by the Blazor-specific articles. I'll double-check and get back to you, but it might be Tuesday, given that I'm trying to get the *Environments* article work wrapped up before the holiday weekend.

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1104 @@
...
+<img src="static-files/images/red-rose.jpg" alt="A red rose" />
+```
+
+For the preceding example, tilde-slash notation is supported in Razor Pages and MVC views (`src="~/StaticFiles/images/red-rose.jpg"`), not for Razor components in Blazor apps.
+
+## Serve files from multiple locations
+
+:::moniker range=">= aspnetcore-6.0"
+
+*The guidance in this section applies to Razor Pages and MVC apps. For guidance that applies to Blazor Web Apps, see <xref:blazor/fundamentals/static-files#serve-files-from-multiple-locations>.*
```
danroth27, line 295: Isn't the guidance the same? Can we remove this duplication?

guardrex: That's what I meant about deltas. It wasn't the same IIRC, but I'll have to look. I'll check and get back to you ... probably Tuesday morning.

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1104 @@
...
+
+```csharp
+app.UseStaticFiles();
+
+app.UseAuthorization();
+
+app.MapStaticAssets();
+```
+
+## Static file authorization
```
danroth27, line 423: I'm a bit confused about what this section is trying to communicate. It sounds like it's trying to explain how you can add authorization checks for your static files. But then the first thing it shows is enabling anonymous access to static web assets. Shouldn't this section start off with showing that static web assets can have authorization policies applied just like any other endpoint?

The rest of the section seems to be about workarounds for enabling authorization when using the static files middleware. This content is certainly important for folks still using the static files middleware, but it dominates this section in a way that I don't think we want.

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1104 @@
...
+* [Compressed](/aspnet/core/performance/response-compression) versions of the assets are served when possible. This optimization doesn't include minification.
+* A [CDN](/microsoft-365/enterprise/content-delivery-networks?view=o365-worldwide&preserve-view=true) can be used to serve the assets closer to the user.
+* Assets are [fingerprinted](https://wikipedia.org/wiki/Fingerprint_(computing)) to prevent reuse of old files.
+
+<xref:Microsoft.AspNetCore.Builder.StaticAssetsEndpointRouteBuilderExtensions.MapStaticAssets%2A>:
+
+* Integrates the information gathered about static web assets during the build and publish process with a runtime library that processes this information to optimize file serving to the browser.
+* Are routing endpoint conventions that optimize the delivery of static assets in an app. It's designed to work with all UI frameworks, including Blazor, Razor Pages, and MVC.
+
+## Static Files Middleware (`UseStaticFiles`)
```
danroth27: I'd like to avoid presenting the reader with a choice of how to handle static files right at the start of this article. Instead, I think the article should start off focused on basic static file handling using MapStaticAssets, almost as if we were documenting static file handling in ASP.NET Core for the first time and MapStaticAssets is the only thing available. After we've covered all the basic stuff, we can then cover scenarios that aren't supported by MapStaticAssets and that require the static file middleware.

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1104 @@
...
 
-[!INCLUDE[](~/fundamentals/static-files/includes/static-files8.md)]
+Static files, also called static assets, are files an ASP.NET Core app that aren't dynamically generated and served directly to clients on request, such as HTML, CSS, image, and JavaScript files.
 
-[!INCLUDE[](~/fundamentals/static-files/includes/static-files6.md)]
+For Blazor static files guidance, which adds to or supersedes the guidance in this article, see <xref:blazor/fundamentals/static-files>.
+
+:::moniker range=">= aspnetcore-9.0"
+
+## Map Static Assets (`MapStaticAssets`)
```
danroth27: Can we start with some basics?:

- To enable static file handling in ASP.NET Core, call `MapStaticAssets`
- You put any files you want to be served as static files in your wwwroot folder (also called the web root)
- By default, only files in the wwwroot folder will be addressable, so you don't need to worry about the rest of your code.
- Only files with specific file extensions mapped to supported media types will be treated as static web assets.
- Your static web assets will get discovered at build time and optimized using content-based finger printing and precompession.
- At runtime, the discovered static web assets will be exposed as endpoints and will have appropriate HTTP headers applied, like caching headers, content type, etc.
- Note that you can also serve static web assets from referenced projects and packages. Link to related content.

danroth27 (changes requested):
Great start! I'd like to simplify the beginning of the static files doc as much as possible and gradually build up to more complicated and legacy scenarios. Dealing with static files is pretty common, so the basics need to come across as simple and straightforward. I also want to consolidate more of the general purpose static web asset content in the top-level static files doc instead of having a bunch of it duplicated in the Blazor. The Blazor static files doc really should only cover Blazor specific stuff, like the framework specific static web assets that it adds, the helpers it provides for referencing static web assets (`Assets[]`), etc.

guardrex:
~Cool ... I'll reach this for updates first thing Tuesday morning because I'd like to wrap up the *Environments* article work tomorrow morning.~

Updates for the next round are done.

aspnetcore/fundamentals/static-files.md
guardrex: IDK why; but `IWebHostBuilder.UseStaticWebAssets` may have never been covered in the main doc set's *Static Files* article. Instead (apparently), it was only covered in the *Razor Pages Class Libraries* article and left to a cross-link from the *Static Files* article to cover it ...

https://learn.microsoft.com/en-us/aspnet/core/razor-pages/ui-class?view=aspnetcore-9.0&tabs=visual-studio#consume-content-from-a-referenced-rcl

I'll fix this by ~moving the guidance on it from the Blazor article to the main doc set article.~ **UPDATE**: *Done!* 👍 What's in the RP article can stay, but ~I'll take a look at it and make sure that it's composed well~. **UPDATE**: *Yes. It seems OK as a passing point there.*

aspnetcore/fundamentals/static-files.md
guardrex: *Removed my last response. I was incorrect about what happened.*

A dev had a problem getting our main doc set guidance with the composite file provider to work in a BWA with two physical file providers (which BTW works for Blazor Server, just not BWAs) ...

https://github.com/dotnet/AspNetCore.Docs/issues/32891

I asked the dev to open a PU issue to see what we needed to do for BWAs ...

https://github.com/dotnet/AspNetCore.Docs/issues/32895

However, the issue was transferred to the docs repo without comment. I closed the transferred issue to work from the original docs issue (32891).

I presumed that what the dev did to solve the problem was a valid approach, so that's what's in the section. Instead of using two physical file providers in a composite file provider, the Blazor content pitches passing `app.Environment.WebRootFileProvider` as the primary provider ...

```csharp
var secondaryProvider = new PhysicalFileProvider(
    Path.Combine(builder.Environment.ContentRootPath, "AdditionalStaticAssets"));
app.Environment.WebRootFileProvider = new CompositeFileProvider(
    app.Environment.WebRootFileProvider, secondaryProvider);
```

So, we still don't have an "official" Javier ruling 👨‍⚖️😄 on this delta (two physical providers in a composite doesn't work in a BWA). I'll leave the current coverage as it is for now, and then we can see on technical review what to do about BWA coverage for serving files from multiple locations.

[Note in passing that what I said in my stricken comment about the main doc set's section partially focusing the Image Tag Helper in this scenario is true. When I added the Blazor article coverage, I stripped that language out, so the Blazor article section may or may not be the right pitch on how to address the multiple locations scenario for BWAs, but it is better tailored to BWAs ***without the Tag Helper remarks***. That's something that I'd try to address in the main doc set if we shed the Blazor article section after hearing back from Javier.]

Here are the two live sections for further analysis on their deltas to see if we should try combining them in the main doc set article ...

* [Main doc set: *Serve files from multiple locations*](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/static-files?view=aspnetcore-9.0#serve-files-from-multiple-locations)
* [Blazor doc set: *Serve files from multiple locations*](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/static-files?view=aspnetcore-9.0#serve-files-from-multiple-locations)

guardrex:
@danroth27 ... Ready for another look 👀. Notes on the changes ...

## Some discussion above

There are two points of discussion above ☝️.

## Focusing on Map Static Assets

> presenting the reader with a choice of how to handle static files right at the start of this article ... MapStaticAssets is the only thing available. After we've covered all the basic stuff, we can then cover scenarios that aren't supported by MapStaticAssets and that require the static file middleware.

For .NET 9 or later, Static Files Middleware is pushed down into the article immediately before where it's first needed.

I also added a < .NET 9 bit for introductory content on Static File Middleware, which I should've placed earlier.

## Duplicate content

> instead of having a bunch of it duplicated in the Blazor

"*A bunch*"?? 😖😆 ... That was never the intention for Blazor's fundamentals coverage. Only in limited cases is there subject duplication. In ancient Blazor times 👴, I checked scenarios from the main doc set's articles to see where deltas required special coverage. That's what ended up in the Blazor *Fundamentals* node articles. Currently, I'm wondering why any fundamental coverage (or host/deploy ... or security) is covered in the Blazor node, given Blazor's preeminent position. MVC/RP stack nodes never carried such content the way that we're carrying Blazor content.

Except for whatever we do with the *Serve static files from multiple locations* section (probably needs Javier's input, links are above ☝️), there's only one other section to address ... [*File mappings and static file options* section](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/static-files?view=aspnetcore-9.0#file-mappings-and-static-file-options). There was a major delta in versions prior to .NET 8. I'm updating the coverage now to drop the content in the Blazor article from .NET 8 forward. We're keeping the Blazor coverage for < .NET 8 due to the change in the way the Blazor script is served.

## Name of the middleware

BTW ... I think it should be "Static File Middleware," not "Files" (plural). The API isn't plural. I'm adjusting a handful of locations for this, including making sure that (by convention) that it's a proper noun.

aspnetcore/fundamentals/static-files.md
DevTKSS: [EDIT by guardrex to number the points for responses]

@guardrex overflowed that static-files.md file and would like to give you small feedback to consider maybe:

1. if I read correctly, you do advise now for the core minimal api guide to use `MapStaticAssets()` and `UseStaticFiles()` and to use the `WebRootFileProvider`. what Information I do miss in there, can we call them multiple times? like if an extensions does call `.UseStaticFiles()` and uses that provider and so on, which we might not see in our code since it might come by a nuget, should we expect problems?

2. reffering to ln 94ff you advise us to call `UseHttpsRedirection()` which I did read before in some MVP posts, but do we need or should we also call Hsts depending on the Development Environment ? just wanting to point out, that if you advise the one of them and assuming from what I readed, that they are important, maybe the code Sample should just give a full block like also what we are required to call before the `.UseStaticFiles()` and which we should keep in mind to put afterwards. for the httpsRedirection I readed that is e.g. important for when we use something like logging. I would miss a full ordered list somewhere in general also to such important list 👍 

3. Line 189 ff: I did set about in a project that was targeting net9.0 and `.UseStaticFiles()` from such `wwwroot/Index.html` and did set without thinking about it could cause issues like you tell there to `/` and instantly [my asp net core wasm host](https://github.com/DevTKSS/DevTKSS.MyManufacturerERP/blob/21bc8e2b3ed451bc90c6f5d8981faeddee7005a1/src/DevTKSS.MyManufacturerERP.Server/Program.cs#L147-L150)  was not longer building! this was not Razor in remark to your Warning there. the WebRootPath was also not set if you ask](https://github.com/DevTKSS/DevTKSS.MyManufacturerERP/blob/21bc8e2b3ed451bc90c6f5d8981faeddee7005a1/src/DevTKSS.MyManufacturerERP.Server/Program.cs#L147-L150) 
created an [issue on my repro](https://github.com/DevTKSS/DevTKSS.MyManufacturerERP/issues/5)
as soon as I removed that path, it was able to find these files again. but what I want to tell is, maybe there should be some remark about what would happen (maybe its an expected issue?) if we define such endpoint on the `/` path? would be helpfull for non professinals

4. line 227ff: this reads like a repeating and since your intro was about the assets usage to me this following section is absolutly confusing when I could not use the StaticFiles in .net9 or later for example.
e.g. lets think the reader wants to do a wasm app like me,  but is not in deep in this topic, but the wasm app is not Blazor and does not call AddRazorPages. the Framework files are imported by exactly that `.UseStaticFiles()` and I do not have a call to the assets. That documentation is now leaving me idea-less because, up from line 235 you seem to tell that the assets do not allow https headers, only static files, but the description from ln36 ff refers to Http Headers as feature + more of course. just as an example.

5. ln 411-416
you introduced us in the first section that we need to call the `MapStaticAssets()` before its `UseStaticFiles()` but now you switching it in the code sample??? and same here, I would miss any Dev environment condition for the redirection.

6. line 723:
and suddenly its again possible to have DirectoryBrowsing with `MapStaticAssets()`? thats not matching...

hoping this helps you improve it a bit 🚀 but for the rest, thanks! nice to see there is some action 👍

guardrex: 1. You'll have to cross-link "for the core minimal api guide" because I don't know what that refers to. I'm not aware that a lib would set up Static File Middleware. I don't think that's a supported scenario, so you'd have to explain what you mean by that.

2. HSTS guidance is [here](https://learn.microsoft.com/en-us/aspnet/core/security/enforcing-ssl?view=aspnetcore-9.0&tabs=visual-studio%2Clinux-sles#http-strict-transport-security-protocol-hsts). Because it's off-topic here, I don't think this article needs to comment on it. I'd prefer to allow the [startup article](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/startup?view=aspnetcore-9.0) show it first and then the [middleware doc](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/middleware/?view=aspnetcore-9.0#middleware-order) get into middleware ordering details. Technical review should turn up any problems here with regard to middleware/endpoint conventions ordering.

3. I defer to technical analysis by the product unit on it, and I'll add whatever coverage they suggest. However, they might end up asking you to open a separate issue on it. 

4. Yes, you're probably correct because the opening paragraph doesn't say the important thing that the second paragraph states, namely that Static File Middleware can't do all of the wonderful things that Map Static Assets endpoint conventions can do. ~I think I should work on that a bit more here now, so I'll make a few changes there for the next round of review.~ **UPDATE**: I reworked that section on the middleware (that only appears for the >=9.0 version of the article).

5. I'm not sure about the last bit of your remark, but on the first part: *Yes!* That's what the product unit's guidance is ***for this particular approach*** for a large collection of assets. Javier will generally review that, but he also can comment on the second part of your remark to see if anything should be changed.

6. I haven't personally tested that scenario. However, I don't see any problem with a dedicated, non-`wwwroot` folder of assets for directory browsing working side-by-side with Map Static Asset endpoint conventions. I defer to Javier on technical review to say if there would be any problem with it working.

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1150 @@
...
-ms.date: 03/18/2025
+ms.date: 09/02/2025
 uid: fundamentals/static-files
 ---
 # Static files in ASP.NET Core
 
 [!INCLUDE[](~/includes/not-latest-version.md)]
 
-[!INCLUDE[](~/fundamentals/static-files/includes/static-files8.md)]
+Static files, also called static assets, are files an ASP.NET Core app that aren't dynamically generated and served directly to clients on request, such as HTML, CSS, image, and JavaScript files.
```
danroth27:
```suggestion
Static files, also called static assets, are files an ASP.NET Core app that aren't dynamically generated. Instead, they're served directly to clients on request, such as HTML, CSS, image, and JavaScript files.
```

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1150 @@
...
-[!INCLUDE[](~/fundamentals/static-files/includes/static-files6.md)]
+For Blazor static files guidance, which adds to or supersedes the guidance in this article, see <xref:blazor/fundamentals/static-files>.
+
+:::moniker range=">= aspnetcore-9.0"
+
+To enable static file handling in ASP.NET Core, call <xref:Microsoft.AspNetCore.Builder.StaticAssetsEndpointRouteBuilderExtensions.MapStaticAssets%2A>. 
+
+By default, static files are stored within the project's [web root](xref:fundamentals/index#web-root) directory. The default directory is `{CONTENT ROOT}/wwwroot`, where the `{CONTENT ROOT}` placeholder is the app's [content root](xref:fundamentals/index#content-root). Only files in the wwwroot folder will be addressable, so you don't need to worry about the rest of your code.
+
+Only files with specific file extensions mapped to supported media types are treated as static web assets.
```
danroth27, line 25: @javiercn Is there more we should say about the default media type mapping for static web assets and how it can be configured?

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1150 @@
...
+
+Static web assets are discovered at build time and optimized using content-based [fingerprinting](https://wikipedia.org/wiki/Fingerprint_(computing)) to prevent the reuse of old file and [compressed](/aspnet/core/performance/response-compression) to reduce asset delivery time. [The optimizations don't include minification.]
```
danroth27:
```suggestion
Static web assets are discovered at build time and optimized using content-based [fingerprinting](https://wikipedia.org/wiki/Fingerprint_(computing)) to prevent the reuse of old files. Assets are also [compressed](/aspnet/core/performance/response-compression) to reduce asset delivery time.
```

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1150 @@
...
+
+Delivery of static assets is based on [endpoint routing](xref:fundamentals/routing), so it works with other endpoint-aware features, such as authorization. It's designed to work with all UI frameworks, including Blazor, Razor Pages, and MVC.
+
+Map Static Assets provides the following benefits:
+
+* Build-time compression for all the assets in the app, including JavaScript (JS) and stylesheets but excluding image and font assets that are already compressed. [Gzip](https://tools.ietf.org/html/rfc1952) (`Content-Encoding: gz`) compression is used during development. Gzip with [Brotli](https://tools.ietf.org/html/rfc7932) (`Content-Encoding: br`) compression is used during publish.
```
danroth27:
```suggestion
* Build-time compression for all the assets in the app, including JavaScript (JS) and stylesheets but excluding image and font assets that are already compressed. [Gzip](https://tools.ietf.org/html/rfc1952) (`Content-Encoding: gz`) compression is used during development. Gzip and [Brotli](https://tools.ietf.org/html/rfc7932) (`Content-Encoding: br`) compression are both used during publish.
```

aspnetcore/fundamentals/static-files.md
danroth27: OK. I'm going to assume the differences in this content for handling static files from multiple locations in Blazor Web Apps is correct then unless @javiercn states otherwise.

guardrex:
This PR is ready for @javiercn's technical review.

aspnetcore/fundamentals/static-files.md
javiercn: Yes, I included a section about that in my original technical description. it should be covered here if it's not covered later in the document. At leas a note/link here to where it can be configured later.

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1150 @@
...
+
+```cshtml
+<link rel="icon" type="image/png" href="~/images/favicon.png" />
+```
+
+:::moniker range=">= aspnetcore-9.0"
+
+## Short-circuit the middleware pipeline
+
+To avoid running the entire middleware pipeline after a static asset is served, which matches the behavior of <xref:Microsoft.AspNetCore.Builder.StaticFileExtensions.UseStaticFiles%2A>, call <xref:Microsoft.AspNetCore.Builder.RouteShortCircuitEndpointConventionBuilderExtensions.ShortCircuit%2A> on <xref:Microsoft.AspNetCore.Builder.StaticAssetsEndpointRouteBuilderExtensions.MapStaticAssets%2A>. Calling <xref:Microsoft.AspNetCore.Builder.RouteShortCircuitEndpointConventionBuilderExtensions.ShortCircuit%2A> immediately executes the endpoint and returns the response, preventing other middleware from executing for static asset requests:
```
javiercn: matched, not served. endpoint routing separates selection from execution.

aspnetcore/fundamentals/static-files.md
guardrex: I'm only going by what's here, so I need to understand where the *original technical description* is in order to add content ... or the content, possibly as just notes, has to be provided to me.

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1150 @@
...
+`EnableDefaultCompressedItems` | Enables default compression include/exclude patterns.
+`CompressionIncludePatterns` | Semicolon-separated list of file patterns to include for compression.
+`CompressionExcludePatterns` | Semicolon-separated list of file patterns to exclude from compression.
+`EnableDefaultCompressionFormats` | Enables default compression formats (Gzip and Brotli).
+`BuildCompressionFormats` | Compression formats to use during build.
+`PublishCompressionFormats` | Compression formats to use during publish.
+`DisableBuildCompression` | Disables compression during build.
+`CompressDiscoveredAssetsDuringBuild` | Compresses discovered assets during build.
+`BrotliCompressionLevel` | Compression level for the Brotli algorithm.
+`LinkAlternativeRepresentationsToOriginalResource` | How to link compressed assets to original resources.
```
javiercn: We should remove this property, I missed it when I created the list, it should not be here.

javiercn: This should ideally be on its own section as `Customizing compression options` and providing examples.

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1150 @@
...
+`StaticWebAssetPublishCompressAllAssets` | Compresses all assets during publish, not just assets discovered or computed during a build.
+
+Property | Description
+--- | ---
+`StaticWebAssetBasePath` | Base URL path for all the assets in a library.
+`StaticWebAssetsFingerprintContent` | Enables content fingerprinting for cache busting.
+`StaticWebAssetFingerprintingEnabled` | Enables fingerprinting feature for static web assets.
+`StaticWebAssetsCacheDefineStaticWebAssetsEnabled` | Enables caching for static web asset definitions.
+`StaticWebAssetBuildManifestPath` | Path to the build manifest JSON file.
+`StaticWebAssetsBuildManifestCacheFilePath` | Path to the build manifest cache file.
```
javiercn: We should remove these two properties, users shouldn't configure them

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1150 @@
...
+`StaticWebAssetEndpointExclusionPattern` | Pattern for excluding endpoints.
+
+Item group | Description | Metadata
+--- | ---
+`StaticWebAssetContentTypeMapping` | Maps file patterns to content types and cache headers for endpoints. | Pattern, Cache
```
javiercn, line 1126: This should be covered under a section about "adding new content type mappings at build time"

javiercn (approved):
Few comments, otherwise looks good.

aspnetcore/fundamentals/static-files.md
guardrex:
```suggestion
To avoid running the entire middleware pipeline after a static asset is matched, which is the behavior of <xref:Microsoft.AspNetCore.Builder.StaticFileExtensions.UseStaticFiles%2A>, call <xref:Microsoft.AspNetCore.Builder.RouteShortCircuitEndpointConventionBuilderExtensions.ShortCircuit%2A> on <xref:Microsoft.AspNetCore.Builder.StaticAssetsEndpointRouteBuilderExtensions.MapStaticAssets%2A>. Calling <xref:Microsoft.AspNetCore.Builder.RouteShortCircuitEndpointConventionBuilderExtensions.ShortCircuit%2A> immediately executes the endpoint and returns the response, preventing other middleware from executing for static asset requests:
```

aspnetcore/fundamentals/static-files.md
guardrex:
```suggestion
```

aspnetcore/fundamentals/static-files.md
guardrex:
```suggestion
```

aspnetcore/fundamentals/static-files.md
guardrex: I placed everything that you provided in https://github.com/dotnet/aspnetcore/issues/63292. I need content or a link to additional information if I'm going to write it up (and note that @danroth27 asked me not to write up new features; leave it to the PU he instructed).

Also, I placed this section on MSBuild properties without versioning any of them. They don't all go back to 3.1, so I leave it to you guys to decide if versioning is something you want to get into the weeds with. If several came in at various times (and because I can't version individual rows of tables), this could get messy fast.

If you want to version without markdown versioning, then a table column can indicate the release that the property appeared.

aspnetcore/fundamentals/static-files.md
```diff
@@ -1,17 +1,1147 @@
...
+
+Metadata Descriptions:
+
+* **Pattern**: A glob pattern used to match files. For `StaticWebAssetContentTypeMapping`, it matches files to determine their content type (for example, `*.js` for JavaScript files). For `StaticWebAssetFingerprintPattern`, it identifies multi-extension files that require special fingerprinting treatment (for example, `*.lib.module.js`).
+
+* **Cache**: Specifies the `Cache-Control` header value for the matched content type. This controls browser caching behavior (for example, `max-age=3600, must-revalidate` for media files).
+
+* **Expression**: Defines how the fingerprint is inserted into the filename. The default is `#[.{fingerprint}]`, which inserts the fingerprint before the extension.
+
+The following table describes the runtime configuration options.
```
guardrex, line 1139: Questions ...

* Are all of these >=10.0?
* Should these be given their own section heading?

aspnetcore/fundamentals/static-files.md
danroth27: @guardrex @javiercn I've opened https://github.com/dotnet/aspnetcore/issues/63698 to track providing the requested content. Let's merge this PR as is and add the content type mapping section separately.

guardrex: Ok ... I have one more pass to get the sample app out of here. It's in the samples repo now. I didn't do it yet because I didn't want to clutter the diff. I'll take care of it by EOD.

aspnetcore/fundamentals/static-files.md
danroth27: These all appear to be available in .NET 9.

Adding a section heading so that it can be easily linked to sounds good.

guardrex: > These all appear to be available in .NET 9 ... Adding a section heading so that it can be easily linked to sounds good.

Will take care of these on the next commit, including dropping the sample from this repo.

guardrex:
@danroth27 ... Ok, those two are done, and the samples are now only in the samples repo, not here.

I think we're good 👍. I'll see it live tomorrow morning and look for any lingering 😈.

aspnetcore/fundamentals/static-files.md
javiercn:
```xml
<ItemGroup>
    <StaticWebAssetContentTypeMapping Include="image/bmp" Cache="<<Cache-Header-To-Use-For-Non-Fingerprinted-Endpoints>>" Pattern="*.bmp" />
</ItemGroup>
```

PR #35969, opened 2025-08-20
Patch update date format

guardrex:
Quick patch update of the date format. It came up when Copilot suggested a change on my static files PR at https://github.com/dotnet/AspNetCore.Docs/pull/35967.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [.github/copilot-instructions.md](https://github.com/dotnet/AspNetCore.Docs/blob/b2d5d8e0456b32b02257508304b0f13dcb930e48/.github/copilot-instructions.md) | [.github/copilot-instructions](https://review.learn.microsoft.com/en-us/aspnet/core/.github/copilot-instructions?branch=pr-en-us-35969) |


wadepickett:
Good catch, thanks!  I remember changing that at one point, but I think I reintroduced it with a copilot edit, which of course is next level irony.

wadepickett (approved):

wadepickett:
I went ahead and merged it right away so the old instruction would not do any more damage today.

PR #35973, opened 2025-08-21
Correct comparison to @bind:set

georgehemmings:
Tweak to the comparison to @bind:set.

The current comparison doesn't make sense to me, I assume it's a copy-and-paste bug.

### DIFF
--- a/aspnetcore/blazor/components/data-binding.md
+++ b/aspnetcore/blazor/components/data-binding.md
@@ -228,7 +228,7 @@ Components support two-way data binding by defining a pair of `@bind` attributes
 
 The `:get` and `:set` modifiers are always used together.
 
-With `:get`/`:set` binding, you can react to a value change before it's applied to the DOM, and you can change the applied value, if necessary. Whereas with `@bind:event="{EVENT}"` attribute binding, where the `{EVENT}` placeholder is a DOM event, you receive the notification after the DOM is updated, and there's no capacity to modify the applied value while binding.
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/components/data-binding.md
```diff
@@ -228,7 +228,7 @@ Components support two-way data binding by defining a pair of `@bind` attributes
 
 The `:get` and `:set` modifiers are always used together.
 
-With `:get`/`:set` binding, you can react to a value change before it's applied to the DOM, and you can change the applied value, if necessary. Whereas with `@bind:event="{EVENT}"` attribute binding, where the `{EVENT}` placeholder is a DOM event, you receive the notification after the DOM is updated, and there's no capacity to modify the applied value while binding.
+With `:get`/`:set` binding, you can react to a value change before it's applied to the DOM, and you can change the applied value, if necessary. Whereas with `@on{DOM EVENT}="{DELEGATE}"`, where the `{DOM EVENT}` placeholder is a DOM event, you receive the notification after the DOM is updated, and there's no capacity to modify the applied value while binding.
```
guardrex: Thanks! Yes, you're right. Let's make two more NIT changes ...

```suggestion
With `:get`/`:set` binding, you can react to a value change before it's applied to the DOM, and you can change the applied value, if necessary. Whereas with `@on{DOM EVENT}="{DELEGATE}"`, where the `{DOM EVENT}` placeholder is a DOM event and the `{DELEGATE}` placeholder is the delegate, you receive the notification after the DOM is updated, and there's no way to modify the applied value while binding.
```

PR #35981, opened 2025-08-22
Prerendering guidance organization

### LINKED ISSUE #35904: Re-organize Blazor prerendering coverage
### Description

Discussion 👉 https://github.com/dotnet/AspNetCore.Docs/pull/35873#discussion_r2253008937

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/components/prerender?view=aspnetcore-9.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/components/prerender.md

### Document ID

ffdc8df3-55a8-b757-d4e6-38c104916002

### Platform Id

50827770-e09d-1eac-9f2e-d593f56579d6

### Article author

@guardrex

### Metadata

* ID: ffdc8df3-55a8-b757-d4e6-38c104916002
* PlatformId: 50827770-e09d-1eac-9f2e-d593f56579d6 
* Service: **aspnet-core**
* Sub-service: **blazor**

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+ffdc8df3-55a8-b757-d4e6-38c104916002)

guardrex:
Fixes #35904

Dan, I took a look at what we can obviously move, and I only see the *Environments* section that pertained to a server-side implementation of a normally client-side-only service. That makes a lot of sense to move because it's one of the ways devs can deal with prerendering that's listed in the *Prerendering* article (the fourth option). So ... that's moved here.

Otherwise, I think we hit a bit of a wall on the rest. The sprinkled-in coverage is tightly bound to the concepts covered for each feature/scenario. Speaking only to the prerendering aspects of these features well in advance of the feature coverage doesn't look like to me that it's going to work ... but ... perhaps we addressed your major concern from the other convo and moved most/all of what we need to move.

One more thing ... 

The index is still here. I do put a little blurb there for discussion on how we could 🔪💀 the index in favor of a filtered search, but note that ...

* It only works in Google.
* It doesn't work in Bing ... or least I don't know how to make it work correctly in Bing.
* It doesn't work with our doc's built-in search ... our docs search doesn't have the ability to filter to just the Blazor node AFAIK.

I still like the index because all I need to remember is that it's here. It helped me a few times. Anywho ... I'll leave it there or cut it out 🔪 it when I hear what you'd like to do with it.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/prerender.md](https://github.com/dotnet/AspNetCore.Docs/blob/65ed912bb2f8a73568e98a6f50c763a525992a67/aspnetcore/blazor/components/prerender.md) | [aspnetcore/blazor/components/prerender](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/prerender?branch=pr-en-us-35981) |
| [aspnetcore/blazor/fundamentals/environments.md](https://github.com/dotnet/AspNetCore.Docs/blob/65ed912bb2f8a73568e98a6f50c763a525992a67/aspnetcore/blazor/fundamentals/environments.md) | [aspnetcore/blazor/fundamentals/environments](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/environments?branch=pr-en-us-35981) |

### DIFF
--- a/aspnetcore/blazor/components/prerender.md
+++ b/aspnetcore/blazor/components/prerender.md
@@ -103,17 +103,42 @@ No compile time error occurs, but a runtime error occurs during prerendering:
 
 This error occurs because the component must compile and execute on the server during prerendering, but <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> isn't a registered service on the server.
 
-There are several approaches that you can take to address this scenario:
+Consider any of the following approaches to address this scenario:
 
-* If the service supports server execution, register the service on the server in addition to the client so that it's available during prerendering. For an example of this scenario, see the guidance for <xref:System.Net.Http.HttpClient> services in the [Blazor Web App external web APIs](xref:blazor/call-web-api#blazor-web-app-external-web-apis) section of the *Call web API* article.
+* [Register the service on the server in addition to the client](#register-the-service-on-the-server-in-addition-to-the-client)
+* [Inject a service that the app can use during prerendering](#inject-a-service-that-the-app-can-use-during-prerendering)
+* [Make the service optional](#make-the-service-optional)
+* [Create a service abstraction](#create-a-service-abstraction)
+* [Disable prerendering for the component](#disable-prerendering-for-the-component)
 
-* Make the service optional if it isn't always needed during prerendering.
+### Register the service on the server in addition to the client
 
-* If a different service implementation is needed on the server, create a service abstraction and create implementations for the service in the `.Client` and server projects. Register the services in each project. Inject the custom service abstraction in the component.
+If the service supports server execution, register the service on the server in addition to the client so that it's available during prerendering. For an example of this scenario, see the guidance for <xref:System.Net.Http.HttpClient> services in the [Blazor Web App external web APIs](xref:blazor/call-web-api#blazor-web-app-external-web-apis) section of the *Call web API* article.
 
-* Disable prerendering for the component or for the entire app.
+### Inject a service that the app can use during prerendering
 
-To optionally inject the service into the component, use constructor injection:
+In some cases, the app can use a service on the server during prerendering and a different service on the client.
+
+For example, the following code obtains the app's environment whether the code is running on the server or on the client by injecting <xref:Microsoft.Extensions.Hosting.IHostEnvironment> from the [`Microsoft.Extensions.Hosting.Abstractions` NuGet package](https://www.nuget.org/packages/Microsoft.Extensions.Hosting.Abstractions):
+
+```csharp
+private string? environmentName;
+
+public Home(IHostEnvironment? serverEnvironment = null, 
+    IWebAssemblyHostEnvironment? wasmEnvironment = null)
+{
+    environmentName = serverEnvironment?.EnvironmentName;
+    environmentName ??= wasmEnvironment?.Environment;
+}
+```
+
+However, this approach adds an additional dependency to the client project that isn't needed.
+
+### Make the service optional
+
+Make the service optional if it isn't required during prerendering using either of the following approaches.
+
+The following example uses constructor injection of <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment>:
 
 ```csharp
 private string? environmentName;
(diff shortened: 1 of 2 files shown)


danroth27:
> I took a look at what we can obviously move, and I only see the Environments section that pertained to a server-side implementation of a normally client-side-only service. That makes a lot of sense to move because it's one of the ways devs can deal with prerendering that's listed in the Prerendering article (the fourth option). So ... that's moved here.

Makes sense. I think the content about dealing with services when doing prerending with client-side rendering can still be simplified. Instead of trying to hash that out in this PR I think I'll take a crack at it in a separate PR and see what you think.

> Otherwise, I think we hit a bit of a wall on the rest. The sprinkled-in coverage is tightly bound to the concepts covered for each feature/scenario. Speaking only to the prerendering aspects of these features well in advance of the feature coverage doesn't look like to me that it's going to work ... but ... perhaps we addressed your major concern from the other convo and moved most/all of what we need to move.

I still don't believe the index is usable or needed. It's a *lot* of links and I don't see why a user would ever try to go through them all. I'll need to take a look at guidance we're linking to see what if anything is worth moving to the prerendering article, but I suspect many of them simply aren't needed. For example, the first link is a discussion of client vs server rendering, which also mentions prerendering. I don't see any guidance in this section, and I don't see anything added beyond what is already covered in the prerendering article already, so I don't think that link adds value. The second link points to a section on static vs interactive routing, which mentions how interactive routing avoids prerendering. Again, this is already covered in the prerendering article. Etc.

> I do put a little blurb there for discussion on how we could 🔪💀 the index in favor of a filtered search

I don't like the filtered search either 😄. Why should I need to search for prerendering related content from the article on prerendering?

guardrex:
Ok ... so we'll hold here for a sec to get your PR merged. After that, I'd like to take look at the organization to see if dealing with services not available during prerendering would benefit from placing each approach neatly into subsections. Then, we can pick back up here with anything else that needs work.

aspnetcore/blazor/components/prerender.md
```diff
@@ -103,7 +103,28 @@ No compile time error occurs, but a runtime error occurs during prerendering:
 
 This error occurs because the component must compile and execute on the server during prerendering, but <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> isn't a registered service on the server.
 
-If the app doesn't require the value during prerendering, this problem can be solved by injecting <xref:System.IServiceProvider> to obtain the service instead of the service type itself:
+There are several approaches that you can take to address this scenario:
+
+* If the service supports server execution, register the service on the server in addition to the client so that it's available during prerendering. For an example of this scenario, see the guidance for <xref:System.Net.Http.HttpClient> services in the [Blazor Web App external web APIs](xref:blazor/call-web-api#blazor-web-app-external-web-apis) section of the *Call web API* article.
+
+* Make the service optional if it isn't always needed during prerendering.
```
danroth27:
```suggestion
* If the service isn't needed during prerendering, make the service optional.
```

aspnetcore/blazor/components/prerender.md
```diff
@@ -103,7 +103,28 @@ No compile time error occurs, but a runtime error occurs during prerendering:
...
+
+* If a different service implementation is needed on the server, create a service abstraction and create implementations for the service in the `.Client` and server projects. Register the services in each project. Inject the custom service abstraction in the component.
+
+* Disable prerendering for the component or for the entire app.
```
danroth27: @guardrex Can you add a link to this bullet to the section above on disabling prerendering?

aspnetcore/blazor/components/prerender.md
guardrex: I'm working on it now. I'll add this suggestion (or something similar). I'm almost done ... just need a few more minutes.

aspnetcore/blazor/components/prerender.md
guardrex: Stand-by for a few more minutes ... almost done with the next round of updates.

danroth27 (approved):
A couple more minor nits, but otherwise this looks good to me. Thanks!

aspnetcore/blazor/components/prerender.md
danroth27: Ah, ok. I'll take another quick look once you're done.

guardrex:
@danroth27 ... I typed that out fairly *fast*, so a few NITs 😈 might need to be addressed. See if subsections organizes those approaches better.

PR #35997, opened 2025-08-26
Fix: update URL and correct text alignment

erkalai:
### Summary
This PR updates example URL and text alignment.

### Changes
- Updated example URL to correct ones
- Adjusted spacing for better readability

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [.github/copilot-instructions.md](https://github.com/dotnet/AspNetCore.Docs/blob/a0b7814e6f795cdad04e5756332a9c8ad14f94a5/.github/copilot-instructions.md) | [.github/copilot-instructions](https://review.learn.microsoft.com/en-us/aspnet/core/.github/copilot-instructions?branch=pr-en-us-35997) |

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -135,15 +135,15 @@ When creating a PR for an issue:
 - Use the following language code and indentation standards for markdown code blocks or the `language` attribute of code snippets:
 
   Content of the snippet | Language code | Indentation in spaces
-  :---: | :---: | :---:
-  C# | csharp | 4
-  HTML | html | 4
-  CSS | css | 4
-  JavaScript | javascript | 2 spaces (use 4 spaces for line continuation)
-  XML | xml | 2
-  JSON | json | 2
-  Console | console | 2
-  Text | - | 2
+  :--------------------: | :-----------: | :-------------------:
+  C#                     | csharp        | 4
+  HTML                   | html          | 4
+  CSS                    | css           | 4
+  JavaScript             | javascript    | 2 spaces (use 4 spaces for line continuation)
+  XML                    | xml           | 2
+  JSON                   | json          | 2
+  Console                | console       | 2
+  Text                   | -             | 2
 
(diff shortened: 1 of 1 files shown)


erkalai:
@dotnet-policy-service agree

wadepickett:
@erkalai, thanks for your contribution!

@tdykstra, while the longer table markers improve visual clarity, does the publishing system handle those fine or does it expect the original shorter markers?

wadepickett (approved):
Approved, but I differ to @tdykstra on the longer markers used in the table and if the publishing system handles those OK.

guardrex (approved):
Thanks @erkalai! 🚀 ... We'll take the link fix without the table changes.

guardrex:
Sorry @wadepickett! I didn't see you were working this.

I reverted the table updates because this isn't for human consumption, and there's been a move in recent years to get rid of extra spacing in tables.

I'll reset the table changes.

**UPDATE**: *Done!* 👍 ... and fixed a little bit of the alignment.

tdykstra:
> @tdykstra, while the longer table markers improve visual clarity, does the publishing system handle those fine or does it expect the original shorter markers?

I don't know but I doubt the publishing system would have a problem with the table markers being longer.

However, for files like this I'm not sure what they would get from a publishing system. Aren't these just plain Markdown files we're dealing with when we maintain them and when the model reads them?

wadepickett:
> > @tdykstra, while the longer table markers improve visual clarity, does the publishing system handle those fine or does it expect the original shorter markers?
> 
> I don't know but I doubt the publishing system would have a problem with the table markers being longer.
> 
> However, for files like this I'm not sure what they would get from a publishing system. Aren't these just plain Markdown files we're dealing with when we maintain them and when the model reads them?

Sorry, dealing with published content all day by brain did not really switch over to non-published copilot instruction files.  Yes, it does not matter as far as publishing.

wadepickett:
@erkalai, thanks again!

PR #35999, opened 2025-08-27
Minimal API responces.md: add missing IBindableFromHttpContext section

### LINKED ISSUE #32903: Document IBindableFromHttpContext
### Description

`IBindableFromHttpContext` was added here: https://github.com/dotnet/aspnetcore/issues/40927

There is no documentation for using it. It should be added to this page.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/responses?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/minimal-apis/responses.md

### Document ID

f0e5da68-cd54-ce74-1ff3-ebd7d9fd09a4

### Article author

@brunolins16

wadepickett:
Fixes #32903

For versions 7-10 of the responses.md topic:
- Added new section:
Custom parameter binding with IBindableFromHttpContext
- Added a new app sample.

### DIFF
--- a/aspnetcore/fundamentals/minimal-apis/includes/responses7-8.md
+++ b/aspnetcore/fundamentals/minimal-apis/includes/responses7-8.md
@@ -1,7 +1,7 @@
 ---
 author: tdykstra
 ms.author: tdykstra
-ms.date: 08/07/2024
+ms.date: 08/26/2025
 ---
 
 :::moniker range=">= aspnetcore-7.0 <= aspnetcore-8.0"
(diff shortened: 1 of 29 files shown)


aspnetcore/fundamentals/minimal-apis/includes/responses7-8.md
```diff
@@ -229,6 +229,31 @@ public static void PopulateMetadata(MethodInfo method, EndpointBuilder builder)
...
+}
+```
+
+By implementing the <xref:Microsoft.AspNetCore.Http.IBindableFromHttpContext%601>, you can create custom types that handle their own binding logic from the HttpContext. When a route handler includes a parameter of this type, the framework automatically calls the static BindAsync method to create the instance:
+
+:::code language="csharp" source="~/fundamentals/minimal-apis/8.0-samples/CustomBindingExample/Program.cs" id="snippet_IBindableFromHttpContext":::
+
+The following is an example implementation of a custom parameter that binds from an HTTP header:
+
+:::code language="csharp" source="~/fundamentals/minimal-apis/8.0-samples/CustomBindingExample/CustomBoundParameter.cs":::
```
wadepickett:
```suggestion
:::code language="csharp" source="~/fundamentals/minimal-apis/8.0-samples/CustomBindingExample/CustomBoundParameters.cs":::
```

tdykstra (approved):

aspnetcore/fundamentals/minimal-apis/responses.md
```diff
@@ -274,6 +274,32 @@ An alternative approach is using the <xref:Microsoft.AspNetCore.Mvc.ProducesAttr
 
 :::code language="csharp" source="~/fundamentals/minimal-apis/9.0-samples/Snippets/Program.cs" id="snippet_11":::
 
+## Custom parameter binding with `IBindableFromHttpContext`
```
mikekistler: This topic is showing up in an article titled "Create responses", but it is about parameter binding. I think this is in the wrong place -- it should appear within the article with the title "Parameter binding".

wadepickett: Thanks @mikekistler,  I moved the new section to the Parameter-binding.md topic, and updated the intro for Custom Binding to include it here: https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/parameter-binding?view=aspnetcore-10.0&branch=pr-en-us-35999#custom-binding

wadepickett:
> Do we need 4 different sample apps and 4 separate include files for this? I don't think the feature has changed substantially since ASP.NET Core 7 so I _think_ we only need one app and feature description.
> 
> Also I think that describing this as "for advanced scenarios" is a bit vague. I _think_ the point of this interface is to avoid the need for reflection and therefore to allow the app to be AOT compatible. If that's correct, I think we should state that explicitly.

You are correct Mike thanks, we can consolidate.  After comparing versions 7-10 it looks like I can do the following:
- Keep parameter-binding7.md as a separate file since it represents a significantly different feature set (no form binding support)
- Consolidate parameter-binding8.md, parameter-binding9.md, and parameter-binding10.md into a single file that covers .NET 8 and above.
- Same for App samples I think:  I will keep the .NET 7 app sample, but consolidate 8-10 to just using a .NET 10 app sample.

wadepickett:
Changed:

After comparing versions 7-10:

Kept parameter-binding7.md as a separate file since it represents a significantly different feature set (no form binding support)
Consolidated parameter-binding8.md, parameter-binding9.md, and parameter-binding10.md into a single file that covers .NET 8 and above (so 8-10+)
Same for App samples, I kept the .NET 7 app sample but consolidate 8-10 to just using a .NET 10 app sample since there is zero difference except the target .NET version.

mikekistler (approved):
Looks good! 👍

PR #36022, opened 2025-08-29
Environments article overhaul

### LINKED ISSUE #31962: Add an example of Docker File
### Description

The page lists may options but not Dockerfile. Please add examples of how to set this for Dockerfile

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/environments.md

### Document ID

1cb5eb30-dd84-e5e1-5826-d74243fabf04

### Article author

@tdykstra

### LINKED ISSUE #35807: Update the sample app for "Use multiple environments in ASP.NET Core" to be based on Blazor instead of Razor Pages
### Description

We want to lead in our docs with our modern recommended web frameworks. The ASP.NET Core environments doc currently references a sample app implemented in Razor Pages. We should use a Blazor based sample instead.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-9.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/fundamentals/environments.md

### Document ID

1cb5eb30-dd84-e5e1-5826-d74243fabf04

### Platform Id

3f914b6e-befc-c9a2-7712-d34a12a97fec

### Article author

@tdykstra

### Metadata

* ID: ceffa6a2-770c-1cd9-5145-dcd423f9dd22
* PlatformId: 3f914b6e-befc-c9a2-7712-d34a12a97fec 
* Service: **aspnet-core**
* Sub-service: **fundamentals**

guardrex:
Fixes #35807 
Fixes #31962 

* The outline is below. I figure going with the following general order: environments (named), control code, and then setting the environment. If you want to change the order a great deal, I recommend giving me a new outline in a comment.
* The (crummy IMO 😆) samples were placed into the samples repo on https://github.com/dotnet/AspNetCore.Docs.Samples/pull/283 at https://github.com/dotnet/AspNetCore.Docs.Samples/tree/main/fundamentals/environments. I'll 🔪 💀 them here at the end because I don't want the diff overwhelmed with file changes during review.
* This one doesn't adopt the dreaded INCLUDES files, but it did copy the entire article at .NET 6, which led (again 😩) to the poor update practice of only updating the current release content. The Azure Apps guidance wasn't updated, so looking at that with the <6.0 content and trying to use the portal would lead to an awfully 😕😖😠 dev. Besides, I'm 💀 that bit anyway in favor of a link to Azure docs to harden that section against future portal changes.
* Per reader request, this adds Docker guidance for setting the environment.
* WRT ✨ `Startup` class magic ✨ ... We'll keep it ***for now.*** It's going to go away at some point in the future.
* The [Blazor article (this link opens the 10.0 version because we have several section changes for 10.0+)](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/environments?view=aspnetcore-10.0) is specifically focused on Blazor scenarios.

The initial outline for the article on the PR ...

* Introductory remarks
* Environments
* Logging
* Environment variables that determine the runtime environment
* Control code execution by environment
* Control rendered content
* Set the environment in a command shell when the app is run (`dotnet run`)
* Set the environment with the launch settings file (`launchSettings.json`)
* Set the environment with an environment variable
  * Azure App Service
  * Set environment variable for a process
  * Set environment variable globally
  * Set the environment for apps deployed to IIS
  * Docker
    * Use a Dockerfile
    * Use Docker Compose
    * Use the `docker run` command
    * Docker environment file
* Set the environment in the app's startup code
* Load configuration by environment
* Environment-specific `Startup` class
* Environment-specific `Startup` class methods
* Additional resources

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/fundamentals/environments.md](https://github.com/dotnet/AspNetCore.Docs/blob/7c047a107b4cfea9d9e5910f2a4bdb025923756d/aspnetcore/fundamentals/environments.md) | [aspnetcore/fundamentals/environments](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/environments?branch=pr-en-us-36022) |


aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
+Although the environment can be any string value, the following environment values are provided by the framework:
 
-1. [DOTNET_ENVIRONMENT](xref:fundamentals/configuration/index#default-host-configuration)
-1. `ASPNETCORE_ENVIRONMENT` when the <xref:Microsoft.AspNetCore.Builder.WebApplication.CreateBuilder%2A?displayProperty=nameWithType> method is called. The default ASP.NET Core web app templates call `WebApplication.CreateBuilder`. The `ASPNETCORE_ENVIRONMENT` value overrides `DOTNET_ENVIRONMENT`.
+* <xref:Microsoft.Extensions.Hosting.Environments.Development>
+* <xref:Microsoft.Extensions.Hosting.Environments.Staging>
+* <xref:Microsoft.Extensions.Hosting.Environments.Production>
 
-:::moniker-end
+The Production environment is configured to maximize security, performance, and app reliability. Some common settings that differ from the Development environment include:
```
danroth27: This revision makes it sound like the framework handles these production configuration changes, but I think most of them are up to the app developer to handle. Can we make it clearer that this is an example list of production concerns that the developer may want to address, not a list of things the framework automatically does?

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
-* Calls <xref:Microsoft.AspNetCore.Builder.ExceptionHandlerExtensions.UseExceptionHandler%2A> when the value of `ASPNETCORE_ENVIRONMENT` is anything other than `Development`.
-* Provides an <xref:Microsoft.AspNetCore.Hosting.IWebHostEnvironment> instance in the <xref:Microsoft.AspNetCore.Builder.WebApplication.Environment> property of `WebApplication`.
+* [`DOTNET_ENVIRONMENT`](xref:fundamentals/configuration/index#default-host-configuration)
+* `ASPNETCORE_ENVIRONMENT` when the <xref:Microsoft.AspNetCore.Builder.WebApplication.CreateBuilder%2A?displayProperty=nameWithType> method is called. The ASP.NET Core web app project templates call `WebApplication.CreateBuilder`. The `DOTNET_ENVIRONMENT` value overrides `ASPNETCORE_ENVIRONMENT` when <xref:Microsoft.AspNetCore.Builder.WebApplicationBuilder> is used. For other hosts, such as <xref:Microsoft.Extensions.Hosting.GenericHostBuilderExtensions.ConfigureWebHostDefaults%2A> and <xref:Microsoft.AspNetCore.WebHost.CreateDefaultBuilder%2A?displayProperty=nameWithType>, `ASPNETCORE_ENVIRONMENT` has higher precedence.
```
danroth27: Can we simplify this language?

```suggestion
* `ASPNETCORE_ENVIRONMENT`

When using <xref:Microsoft.AspNetCore.Builder.WebApplication>, the `DOTNET_ENVIRONMENT` value take precedence over `ASPNETCORE_ENVIRONMENT`. When using <xref:Microsoft.AspNetCore.Builder.WebHost>, `ASPNETCORE_ENVIRONMENT` takes precedence.
```

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
+:::moniker range=">= aspnetcore-6.0 < aspnetcore-7.0"
 
-:::code language="cshtml" source="environments/samples/6.x/EnvironmentsSample/Pages/About.cshtml" id="snippet_Environment":::
+* [`DOTNET_ENVIRONMENT`](xref:fundamentals/configuration/index#default-host-configuration)
+* `ASPNETCORE_ENVIRONMENT` when the <xref:Microsoft.AspNetCore.Builder.WebApplication.CreateBuilder%2A?displayProperty=nameWithType> method is called. The ASP.NET Core web app project templates call `WebApplication.CreateBuilder`. The `ASPNETCORE_ENVIRONMENT` value overrides `DOTNET_ENVIRONMENT`.
 
-The [About page](https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/fundamentals/environments/samples/6.x/EnvironmentsSample/Pages/About.cshtml) from the [sample code](https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/fundamentals/environments/samples/6.x) includes the preceding markup and displays the value of `IWebHostEnvironment.EnvironmentName`.
+:::moniker-end
```
danroth27, line 67: Isn't this saying pretty much the same thing as the >= aspnetcore-7.0 version above?

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
 
-### Set environment on the command line
+Use <xref:Microsoft.AspNetCore.Builder.WebApplicationBuilder.Environment?displayProperty=nameWithType> or <xref:Microsoft.AspNetCore.Builder.WebApplication.Environment?displayProperty=nameWithType> to conditionally add services or middleware depending on the current environment.
 
-Use the `--environment` flag to set the environment. For example:
+The following code in the app's `Program` file:
 
-```dotnetcli
-dotnet run --environment Production
+* Uses <xref:Microsoft.AspNetCore.Builder.WebApplication.Environment?displayProperty=nameWithType> to distinguish the environment, specifically when the value of `ASPNETCORE_ENVIRONMENT` is anything other than `Development`. This approach is useful when the app only requires adjusting the `Program` file for a few environments with minimal code differences per environment. When many code differences exist per environment, consider using one of the [environment-based `Startup` class approaches](#environment-based-startup-class-and-methods), which is covered later in this article.
```
danroth27: I don't think we want to recommend the older `Startup` class pattern.

```suggestion
* Uses <xref:Microsoft.AspNetCore.Builder.WebApplication.Environment?displayProperty=nameWithType> to distinguish the environment.
```

aspnetcore/fundamentals/environments.md
guardrex: The override is reversed (at least it says so, as I never kept this tidbit in my head) ...

* .NET 7: `DOTNET_ENVIRONMENT` value overrides `ASPNETCORE_ENVIRONMENT`
* .NET 6: `ASPNETCORE_ENVIRONMENT` value overrides `DOTNET_ENVIRONMENT`

... and we can't version single bullet list items. The whole list has to be versioned, unless we space out the bullet list. We do that sometimes, particularly if there is a lot of text per bullet.

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
 
-The preceding JSON contains two profiles:
+:::moniker-end
 
-* `EnvironmentsSample`: The profile name is the project name. As the first profile listed, this profile is used by default. The `"commandName"` key has the value `"Project"`, therefore, the [Kestrel web server](xref:fundamentals/servers/kestrel) is launched.
+Within the app, the <xref:Microsoft.Extensions.Hosting.IHostEnvironment> provides general information about the app's hosting environment, and the <xref:Microsoft.Extensions.Hosting.IHostEnvironment.EnvironmentName?displayProperty=nameWithType> property indicates the app's current environment.
 
-* `IIS Express`: The `"commandName"` key has the value `"IISExpress"`, therefore, [IISExpress](/iis/extensions/introduction-to-iis-express/iis-express-overview) is the web server.
+## Control rendered content in Razor Pages and MVC
```
danroth27: If we're going to show samples with web UI rendering logic, then we need to lead with Blazor. Preferably, we only show Blazor for core scenario to avoid confusing new users with multiple ways of doing the same thing. But we still need to document MVC and Razor Pages because they are heavily used and still supported. The environment tag helper is already documented in the tag helper documentation. In this case, I think we should replace the example MVC and Razor Pages example with a Blazor example that shows how you can inject `IHostEnvironment` to get the current environment and then render some content conditionally based on the environment.

guardrex: Yes ... but again leading to the question of why ***all*** Blazor *Fundamentals* node content isn't going to move over here? Specifically in this case, all of Blazor's *Environments* article coverage ...

https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/environments?view=aspnetcore-9.0

I mean to really lead with Blazor, would any of Blazor's fundamentals coverage be squirreled back in what is the equivalent of the "UI stack" coverage for MVC or RP?

If we keep split coverage, the new Blazor *Passkeys* article is out-of-place. It should go back into the Blazor node with the rest of Blazor's security coverage. The *Passkeys* overview stays in the main doc set's security node.

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
+The preceding approaches open the **Launch Profiles** dialog where you can edit the environment variable settings in the `launchSettings.json` file. Changes made to project profiles may not take effect until the web server is restarted. Kestrel must be restarted before it can detect changes made to its environment.
 
-:::code language="json" source="environments/samples/6.x/EnvironmentsSample/.vscode/launch.json" range="4-10,32-38":::
+Profiles can be selected in the Visual Studio UI next to the Start button (►).
 
-The `.vscode/launch.json` file is used only by Visual Studio Code.
+When a solution contains multiple projects, only set the environment for the startup project.
 
-### Production
+Alternatively, use the [`dotnet run`](/dotnet/core/tools/dotnet-run) command with the [`-lp|--launch-profile` option](/dotnet/core/tools/dotnet-run#options) set to the profile's name. *This approach only supports Kestrel profiles.*
```
danroth27:
```suggestion
Alternatively, use the [`dotnet run`](/dotnet/core/tools/dotnet-run) command with the [`-lp|--launch-profile` option](/dotnet/core/tools/dotnet-run#options) set to the profile's name. *This approach only supports launch profiles based on the `Project` command.*
```

aspnetcore/fundamentals/environments.md
danroth27: Ah, I need to learn to read more carefully! Thanks!

aspnetcore/fundamentals/environments.md
danroth27: > Yes ... but again leading to the question of why all Blazor Fundamentals node content isn't going to move over here? Specifically in this case, all of Blazor's Environments article coverage ...

Yup, I think that's the right thing to do, but we need to do so tastefully. The fundamentals of ASP.NET Core should feel simple and broadly applicable. As we merge the Blazor fundamentals into the ASP.NET Core fundamentals we should be careful to remove any content duplication and to avoid excessive Blazor specific details. I don't want to overwhelm the ASP.NET Core fundamentals docs with a bunch of Blazor content that's specific to narrow scenarios. So, let's be judicious about how we do the merge. If half of the merged Environments doc ends up being about Blazor WebAssembly specific concerns, that's probably not ideal.

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
-## Set the environment by setting an environment variable
+```json
+"env": {
+    "ASPNETCORE_ENVIRONMENT": "Staging",
+    ...
+},
+```
 
-It's often useful to set a specific environment for testing with an environment variable or platform setting. If the environment isn't set, it defaults to `Production`, which disables most debugging features. The method for setting the environment depends on the operating system.
+The `.vscode/launch.json` file is only used by Visual Studio Code.
```
danroth27, line 223: Using *launch.json* isn't necessary with C# Dev Kit. C# Dev Kit in Visual Studio Code will pick up the available launch profiles from your *launchSettings.json* file.

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
 
-When the host is built, the last environment setting read by the app determines the app's environment. The app's environment can't be changed while the app is running.
+## Set the environment with an environment variable
```
danroth27, line 225: I really don't think we need to be documenting how to set an environment variable on various platforms, like Windows, macOS, and Linux. I'm ok keeping some pointers on how to do it for App Service, IIS, Docker, and `dotnet run`, although I'd prefer those pointers to mostly be links to the docs for the corresponding service to tech.

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
+:::moniker-end
 
-**Per IIS Application Pool**
+## Load configuration by environment
 
-To set the `ASPNETCORE_ENVIRONMENT` environment variable for an app running in an isolated Application Pool (supported on IIS 10.0 or later), see the *AppCmd.exe command* section of the [Environment Variables &lt;environmentVariables&gt;](/iis/configuration/system.applicationHost/applicationPools/add/environmentVariables/#appcmdexe) topic. When the `ASPNETCORE_ENVIRONMENT` environment variable is set for an app pool, its value overrides a setting at the system level.
+To load configuration by environment, see <xref:fundamentals/configuration/index#json-configuration-provider>.
 
-When hosting an app in IIS and adding or changing the `ASPNETCORE_ENVIRONMENT` environment variable, use any one of the following approaches to have the new value picked up by apps:
+## Environment-based `Startup` class and methods
```
danroth27:
```suggestion
## Access the environment from a `Startup` class
```

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
 
-* Execute `net stop was /y` followed by `net start w3svc` from a command prompt.
-* Restart the server.
+Since the release of .NET 6, the ASP.NET Core project templates place startup code for service configuration and request pipeline processing in the `Program` file. Use of a `Startup` class (`Startup.cs`) with [`Configure`](xref:Microsoft.AspNetCore.Hosting.StartupBase.Configure%2A) and [`ConfigureServices`](xref:Microsoft.AspNetCore.Hosting.StartupBase.ConfigureServices%2A) methods was required before the release of .NET 6 but remains supported in ASP.NET Core. You can use the guidance in this section to control code execution by environment using a `Startup` class in any release of ASP.NET Core when a large number of code changes are required for each environment in use by an app.
```
danroth27, line 403: I prefer to avoid history lessons in our docs. Let's just document our features and guidance.

```suggestion
```

danroth27: I prefer to avoid history lessons in our docs. Let's just document our features and guidance.

```suggestion
```

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
 
-#### macOS
+Except for the [`Startup` class conventions](#startup-class-conventions) approach, the `Program` class uses the following general approach to identify the `Startup` class, where the `Startup` class is the type for the <xref:Microsoft.AspNetCore.Hosting.WebHostBuilderExtensions.UseStartup%2A?displayProperty=nameWithType> method:
```
danroth27, line 405: It is odd that we no longer document the old Startup class pattern in our app startup doc. We should probably fix that. But I think we can assume the reader is familiar with it if they care to read this section.

danroth27: It is odd that we no longer document the old Startup class pattern in our app startup doc. We should probably fix that. But I think we can assume the reader is familiar with it if they care to read this section.

```suggestion
```

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
 
-Setting the current environment for macOS can be performed in-line when running the app:
+```csharp
+public class Program
+{
+    public static void Main(string[] args)
+    {
+        CreateHostBuilder(args).Build().Run();
+    }
```
danroth27:
````suggestion
````

File: aspnetcore/fundamentals/environments.md [hunk 13]
```diff
@@ -1,601 +1,580 @@
...
 
-```bash
-ASPNETCORE_ENVIRONMENT=Staging dotnet run
+    public static IHostBuilder CreateHostBuilder(string[] args) =>
+        Host.CreateDefaultBuilder(args)
+            .ConfigureWebHostDefaults(webBuilder =>
+            {
+                webBuilder.UseStartup<Startup>();
+            });
+}
```
danroth27:
````suggestion
````

File: aspnetcore/fundamentals/environments.md [hunk 14]
```diff
@@ -1,601 +1,580 @@
...
 ```
 
-Alternatively, set the environment with `export` prior to running the app:
-
-```bash
-export ASPNETCORE_ENVIRONMENT=Staging
+### Inject `IWebHostEnvironment` into the `Startup` class
```
danroth27:
```suggestion
```

aspnetcore/fundamentals/environments.md
guardrex: Ok, but that's not the intention. This is here to explain when the features can be used without having to flip through versioned content, see new content coverage, or study framework support for API in the API Browser. Perhaps, I should try to rephrase this into something shorter  that sounds less like a history lesson.

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
+        }
+        else
+        {
+            ...
+        }
+
+        ...
+    }
+}
 ```
```
danroth27, line 415: The primary constructor version is simpler and more idiomatic with modern C#, so let's just use that and have one sample.

````suggestion
Inject <xref:Microsoft.AspNetCore.Hosting.IWebHostEnvironment> into the `Startup` constructor to access the current environment during startup configuration.
```

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
 
-Machine-level environment variables are set in the *.bashrc* or *.bash_profile* file. Edit the file using any text editor. Add the following statement:
-
-```bash
-export ASPNETCORE_ENVIRONMENT=Staging
+The preceding example is simplified with the use of [primary constructors](/dotnet/csharp/whats-new/tutorials/primary-constructors) in C# 12 (.NET 8) or later:
```
danroth27:
```suggestion
```

aspnetcore/fundamentals/environments.md
guardrex: I did have this here for visibility to those who have never seen it or forgotten about it.

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
+            ...
+        }
+
+        ...
+    }
+}
 ```
 
-#### Linux
+### `Startup` class conventions
```
danroth27:
```suggestion
## Environment-specific `Startup` class
```

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
+    return Host.CreateDefaultBuilder(args)
+        .ConfigureWebHostDefaults(webBuilder =>
+        {
+            webBuilder.UseStartup(assemblyName);
+        });
+}
+```
 
-:::code language="csharp" source="environments/samples/3.x/EnvironmentsSample/StartupClassConventions.cs" id="snippet":::
+### `Startup` class method conventions
```
danroth27:
```suggestion
## Environment-specific `Startup` class methods
```

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
 
-Use the <xref:Microsoft.AspNetCore.Hosting.HostingAbstractionsWebHostBuilderExtensions.UseStartup(Microsoft.AspNetCore.Hosting.IWebHostBuilder,System.String)> overload that accepts an assembly name:
+The `Configure` and `ConfigureServices` methods support environment-specific versions of the form `Configure{ENVIRONMENT NAME}` and `Configure{ENVIRONMENT NAME}Services`, where the `{ENVIRONMENT NAME}` placeholder is the environment name. If a matching environment name isn't found for the named methods, the `ConfigureServices` or `Configure` method is used, respectively. This approach is useful when the app requires configuring startup for several environments with many code differences per environment:
```
danroth27:
```suggestion
The `Configure` and `ConfigureServices` methods support environment-specific versions of the form `Configure{ENVIRONMENT NAME}` and `Configure{ENVIRONMENT NAME}Services`, where the `{ENVIRONMENT NAME}` placeholder is the environment name. If a matching environment name isn't found for the named methods, the `ConfigureServices` or `Configure` method is used, respectively.
```

aspnetcore/fundamentals/environments.md
```diff
@@ -1,601 +1,580 @@
...
 ```
 
-#### Linux
+### `Startup` class conventions
 
-For Linux distributions, use the `export` command at a command prompt for session-based variable settings and *bash_profile* file for machine-level environment settings.
+An app can define multiple `Startup` classes for different environments with the naming convention `Startup{EnvironmentName}` class, where the `{ENVIRONMENT NAME}` placeholder is the environment name.
 
-### Set the environment in code
+The class whose name suffix matches the current environment is prioritized. If a matching `Startup{EnvironmentName}` class isn't found, the `Startup` class is used. This approach is useful when the app requires configuring startup for several environments with many code differences per environment.
```
danroth27:
```suggestion
The class whose name suffix matches the current environment is prioritized. If a matching `Startup{EnvironmentName}` class isn't found, the `Startup` class is used.
```

aspnetcore/fundamentals/environments.md
guardrex: Ok, but normally we would version it if we don't show both. However, I did start to only show primary ctors in the Blazor node. Version flipping was becoming annoying in spots.

danroth27: We don't have to show the primary constructor version. We could just show a normal constructor as our single sample. Thie startup pattern is an older pattern that is eventually going away, so sticking to the older pattern kinda makes sense. But I definitely don't want to show two version of the same sample using different C# features.

guardrex:
@danroth27 ... Thanks! I'll get the article updated in the morning and ping u. Then, Javier after you, followed by Tom.

guardrex:
@danroth27 ... Updates made. Another look before asking Javier to look?

~Actually ... there's one more thing for me to do ........~

~The *Control rendered content in Razor Pages and MVC* has to be addressed. I'll work on that and ping back shortly.~

***UPDATE***: That last commit addresses the RP/MVC piece.

guardrex:
@tdykstra ... I think we're still planning on getting a technical review from Javier, but you can review anytime.

danroth27 (approved):
Looks great! :shipit:

guardrex:
@javiercn ... No need to review. Dan said 'Go for liftoff!' 🚀 on this one.

guardrex:
@tdykstra ... I'm going to go ahead this morning and merge this heading out for the weekend for some much needed R&R. I'm *toast* 😩 ... busy week ... and still only getting two days of PTO per year, so I'm getting close to a 🧠🔥and burnout this year. Of course, it's your puppy, so I defer additional work and updates to you on future issues/PRs. I sought to drop unnecessary whole-topic versioning, get non-essential sample code out with code improvements, improve cross-linking (including for API), improve overall organization, drop named anchors, improve section headings, and improve phrasing. Please hit me up anytime for 'why the heck did you do that 🦖?!' concerns.

Issue #34437, opened 2025-01-06
Blazor .NET 10 Tracking Issue


guardrex:
## NOTE TO SELF 🦖 

For .NET 11, there's a sample app path change to make. https://github.com/dotnet/blazor-samples/pull/574#discussion_r2298384906

## ASP.NET Core 10: Web UI

* Primary (Larry's team): https://github.com/orgs/dotnet/projects/503/views/1
* Auth+Identity (Artak's team): https://github.com/orgs/dotnet/projects/478

## Derived from PU tracking

🟢 = Ready&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; ✔️ = Done&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; 🛑 = Wait on PU to finish&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; ❌ = No action required

Status | Live | P | Issue | PR | Item | PU Links/<br>Blog
:---: | :---: | :---: | :---: | :---: | --- | :---:
✔️ | [Live](https://learn.microsoft.com/en-us/aspnet/core/blazor/forms/validation?view=aspnetcore-10.0#use-validation-models-from-a-different-assembly) | RC1| #36240 | #36246 | Validation models from library in .NET 10 | [PU Issue](https://github.com/dotnet/aspnetcore/issues/63626#issuecomment-3422644050)
✔️  | [Live](https://learn.microsoft.com/en-us/aspnet/core/blazor/tutorials/movie-database-app/part-3?view=aspnetcore-10.0#details-component) | GA | #35702 | #36489 | Blazor Movie dB tutorial Not Found updates | [PU Issue](https://github.com/dotnet/Scaffolding/issues/3177) - Waiting on [this issue](https://github.com/dotnet/Scaffolding/issues/3177).
❌ | \[Live]() | RC2 | &mdash; | &mdash; | Disable preloading for enhanced navigation | [PU Issue](https://github.com/dotnet/aspnetcore/pull/63546) - ***Skipping per [DR guidance](https://github.com/dotnet/AspNetCore.Docs/issues/36182#issuecomment-3382109861)***
[...]

PR #36053, opened 2025-09-04
Blazor RC1 coverage

### LINKED ISSUE #35897: Complex framework types for JS interop (2)
### Description

Following #31421 and #34734 (and see https://github.com/dotnet/aspnetcore/issues/52947#issuecomment-2674438882), react to the feature work that was performed ...

* https://github.com/dotnet/aspnetcore/issues/52947#issuecomment-3163300386
* https://github.com/dotnet/aspnetcore/issues/63098

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/configure-trimmer?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/host-and-deploy/configure-trimmer.md

### Document ID

1487b5e7-a40c-d757-c530-d0e4017f189e

### Platform Id

4300acf4-0cea-1a83-bb85-7e567889b895

### Article author

@guardrex

### Metadata

* ID: 2203b81b-1b37-8599-eb50-f6284170d224
* PlatformId: 4300acf4-0cea-1a83-bb85-7e567889b895 
* Service: **aspnet-core**
* Sub-service: **blazor**

guardrex:
Addresses #34437
Fixes #35897

# 🚧 👷‍♀️ ***WIP*** 👷‍♂️ 🏗️


aspnetcore/blazor/forms/validation.md
```diff
@@ -365,6 +377,13 @@ When validation messages are set in the component, they're added to the validato
 
 :::moniker-end
 
+:::moniker range=">= aspnetcore-10.0"
+
+> [!NOTE]
+> The approach in this section relies on an MVC controller for validation on the server. Although the approach is supported for all current and future versions of .NET, we plan to improve the approach by adopting [Minimal APIs](xref:fundamentals/minimal-apis) for server-side validation. The work is tracked by [Server validation with a validator component upgrade (`dotnet/AspNetCore.Docs` #36051)](https://github.com/dotnet/AspNetCore.Docs/issues/36051).
+
+:::moniker-end
+
```
danroth27: I like the plan to switch from an API controller to a minimal API in this section, but I don't think we need to advertise that in the doc. 

```suggestion
```

aspnetcore/blazor/fundamentals/routing.md
```diff
@@ -704,6 +704,35 @@ The following component:
...
 
+:::moniker range=">= aspnetcore-9.0"
+
+## Navigation Manager redirect behavior during static server-side rendering (static SSR)
+
+:::moniker-end
+
+:::moniker range=">= aspnetcore-10.0"
+
+In .NET 9 for a redirect during static server-side rendering (static SSR), <xref:Microsoft.AspNetCore.Components.NavigationManager> relied on throwing an exception that was captured by the framework, which converted the error into a redirect. Code that existed after the call to <xref:Microsoft.AspNetCore.Components.NavigationManager.NavigateTo%2A> abruptly stopped, which also didn't match the behavior during interactive rendering. When using Visual Studio, the debugger stopped on the exception, which required deselecting the checkbox for **Break when this exception type is user-handled** in the Visual Studio UI to avoid the debugger stopping for future redirects.
```
danroth27: We ended up making the new static SSR navigation behavior opt-in for .NET 10. The default is still to throw an exception. Setting the `<BlazorDisableThrowNavigationException>true</BlazorDisableThrowNavigationException>` property will switch on the new behavior to no longer throw an exception. This behavior is enabled in the .NET 10 version of the Blazor Web App project template.

aspnetcore/blazor/host-and-deploy/configure-trimmer.md
```diff
@@ -42,73 +42,123 @@ For more information, see [Trimming options (.NET documentation)](/dotnet/core/d
 
 ## Failure to preserve types used by a published app
 
-Trimming may have detrimental effects for a published app leading to runtime errors. In apps that use [reflection](/dotnet/csharp/advanced-topics/reflection-and-attributes/), the IL Trimmer often can't determine the required types for runtime reflection and trims them away or trims away parameter names from methods. This can happen with complex framework types used for JS interop, JSON serialization/deserialization, and other operations.
+Trimming may have detrimental effects for a published app leading to runtime errors, even in spite of setting the [`<PublishTrimmed>` property](#configuration) to `false` in the project file. In apps that use [reflection](/dotnet/csharp/advanced-topics/reflection-and-attributes/), the IL Trimmer often can't determine the required types for runtime reflection and trims them away or trims away parameter names from methods. This can happen with complex framework types used for JS interop, JSON serialization/deserialization, and other operations.
```
danroth27, line 45: > even in spite of setting the [`<PublishTrimmed>` property](#configuration) to `false`

This doesn't sound right. Setting `PublishTrimmed` to false should disable trimming for the entire app. Does this not work? If it doesn't, I think it's a bug that we should fix, not document.

guardrex: I'll double check this in the morning and report back.

aspnetcore/blazor/host-and-deploy/configure-trimmer.md
```diff
@@ -42,73 +42,123 @@ For more information, see [Trimming options (.NET documentation)](/dotnet/core/d
 
 ## Failure to preserve types used by a published app
```
danroth27, line 43: @javiercn Can you review the recommendations made in this section for dealing with type preservation for JS interop related scenarios?

aspnetcore/blazor/state-management/prerendered-state-persistence.md
danroth27, line 1: @javiercn Is there more you want to add here for handling persistent component state with enhanced nav?

aspnetcore/migration/90-to-100/includes/blazor.md
```diff
@@ -3,3 +3,30 @@ Complete migration coverage for Blazor apps is scheduled for September and Octob
 ### Adopt passkey user authentication in an existing Blazor Web App
 
 For guidance, see <xref:security/authentication/passkeys/blazor?pivots=existing-app>.
+
+### Update the `IdentityRedirectManager` in apps based on the Blazor Web App template with Individual Accounts
+
+The `IdentityRedirectManager` previously threw an <xref:System.InvalidOperationException> in the `RedirectTo` method to ensure the method wasn't called from an interactive render mode and all the redirection methods were marked with the [`[DoesNotReturn]` attribute](xref:System.Diagnostics.CodeAnalysis.DoesNotReturnAttribute). For more information, see <xref:aspnetcore-10#navigationmanagernavigateto-no-longer-throws-a-navigationexception>.
```
danroth27: We decided to make the new behavior that doesn't throw on navigation opt-in, so users only need to do these migration steps if they choose to set `<BlazorDisableThrowNavigationException>true</BlazorDisableThrowNavigationException>`, which we would generally recommend.

aspnetcore/migration/90-to-100/includes/blazor.md
```diff
@@ -3,3 +3,30 @@ Complete migration coverage for Blazor apps is scheduled for September and Octob
...
+  ```diff
+  - [DoesNotReturn]
+  ```
+
+Alternatively, to revert to the previous behavior of throwing a <xref:Microsoft.AspNetCore.Components.NavigationException>, set the following <xref:System.AppContext> switch in the `Program` file:
+
+```csharp
+AppContext.SetSwitch(
+    "Microsoft.AspNetCore.Components.Endpoints.NavigationManager.DisableThrowNavigationException", 
+    isEnabled: false);
```
danroth27: No need to revert the new behavior. It's opt-in now.

```suggestion
```

aspnetcore/release-notes/aspnetcore-10/includes/blazor.md
```diff
@@ -374,13 +374,13 @@ For more information, see <xref:blazor/fundamentals/static-files?view=aspnetcore
 
 ### `NavigationManager.NavigateTo` no longer throws a `NavigationException`
```
danroth27: We need to reframe this as an opt-in behavior and highlight that the behavior is enabled by default in the Blazor project template.

aspnetcore/release-notes/aspnetcore-10.0.md
```diff
@@ -130,6 +130,78 @@ if (RedirectHttpResult.IsLocalUrl(url))
 
 Thank you [@martincostello](https://github.com/martincostello) for this contribution!
 
+### Validation improvements for Blazor and Minimal APIs
```
danroth27: Let's cover the new validation features for Minimal APIs & Blazor in their respective sections and avoid having this separate validation improvements subsection here in the Miscellaneous section. The What's New doc should specify what's new in .NET 10 when compared to the previous version. We don't need to detail deltas between the various .NET 10 previews here. Since complex object graph validation is completely new for both minimal APIs and Blazor, all the new validation functionality should be covered together above.

aspnetcore/release-notes/aspnetcore-10/includes/blazor.md
```diff
@@ -708,3 +721,44 @@ In the following example, a hidden input field is created for the form's `Parame
...
 }
 ```
+
+### Preservation of types used by a published app
+
+We continue to recommend use of custom types for JS interop, JSON serialization/deserialization, and other operations that rely on [reflection](/dotnet/csharp/advanced-topics/reflection-and-attributes/). However, using framework types that ordinarily are trimmed when publishing an app remains supported with the following approaches:
+
+* [Preserve the type as a dynamic dependency](xref:blazor/host-and-deploy/configure-trimmer#preserve-the-type-as-a-dynamic-dependency): Supported since .NET 5.
+* [Use a Root Descriptor](xref:blazor/host-and-deploy/configure-trimmer#use-a-root-descriptor): *A new approach for apps targeting .NET 10 or later.*
+
```
danroth27: I don't think this one really merits being covered here in the release notes. It's more of a bug fix than a new feature.

```suggestion
```

aspnetcore/release-notes/aspnetcore-10.0.md
guardrex: It's late here and Im in 🛌 dozing off, but Im not clear on which parts are covering preview deltas that need to be cut. I'll study it in the morning, but I might need more specific change guidance.

oroztocil: These paragraphs were meant for the RC1 announcement, not for the general .NET 10 release notes. @guardrex, feel free to ping me with any questions regarding the validation area.

aspnetcore/blazor/host-and-deploy/configure-trimmer.md
guardrex: @danroth27 ... ***Confirmed!***

image

image

guardrex:
@danroth27 @javiercn @oroztocil ... Updates made ... and there were quite a few, so a new look 👁️ is a good idea.

I still need ...

* Anything that you want to change about what is stated on the ineffectiveness of `<PublishTrimmed>`, which doesn't stop trimming of the types. We've published that remark for quite a while now, and I don't think we can pull it until the framework changes because devs wrote in confused why it wasn't working.
* DR's pings for Javier to comment ☝️.

If possible, I'd like to merge this PR no later than noon EST (9am PST). It won't go live until the Merge-to-Live PR goes in at release time.

aspnetcore/blazor/forms/validation.md
```diff
@@ -1638,26 +1650,18 @@ In the following `OrderPage` component, the <xref:Microsoft.AspNetCore.Component
...
 
 :::moniker range="< aspnetcore-10.0"
 
 ## Nested objects, collection types, and complex types
 
-Blazor provides support for validating form input using data annotations with the built-in <xref:Microsoft.AspNetCore.Components.Forms.DataAnnotationsValidator>. However, the <xref:Microsoft.AspNetCore.Components.Forms.DataAnnotationsValidator> only validates top-level properties of the model bound to the form that aren't collection- or complex-type properties.
+> [!NOTE]
+> For apps targeting .NET 10 or later, we no longer recommend using the *experimental* [`Microsoft.AspNetCore.Components.DataAnnotations.Validation`](https://www.nuget.org/packages/Microsoft.AspNetCore.Components.DataAnnotations.Validation) package and approach described in this section. We recommend using the built-in validation features of the the <xref:Microsoft.AspNetCore.Components.Forms.DataAnnotationsValidator> component, which received feature parity and behavioral compatibility updates with <xref:System.ComponentModel.DataAnnotations.Validator?displayProperty=nameWithType> at the release of .NET 10. For more information, see <xref:aspnetcore-10#validation-improvements-for-blazor-and-minimal-apis>.
```
oroztocil: > We recommend using the built-in validation features of the the <xref:Microsoft.AspNetCore.Components.Forms.DataAnnotationsValidator> component, which received feature parity and behavioral compatibility updates with <xref:System.ComponentModel.DataAnnotations.Validator?displayProperty=nameWithType> at the release of .NET 10.

I don't think it is necessary to mention the feature parity here - the important bit is that since .NET 10, the standard `DataAnnotationsValidator` supports validation of complex objects, and that the older experimental package should not be used.

aspnetcore/blazor/forms/validation.md
```diff
@@ -134,6 +134,18 @@ The <xref:Microsoft.AspNetCore.Components.Forms.DataAnnotationsValidator> compon
 * [`DataAnnotationsValidator`](https://github.com/dotnet/AspNetCore/blob/main/src/Components/Forms/src/DataAnnotationsValidator.cs)
 * [`EnableDataAnnotationsValidation`](https://github.com/dotnet/AspNetCore/blob/main/src/Components/Forms/src/EditContextDataAnnotationsExtensions.cs)
 
+:::moniker range=">= aspnetcore-10.0"
+
+The <xref:Microsoft.AspNetCore.Components.Forms.DataAnnotationsValidator> component has the same validation order and short-circuiting behavior as <xref:System.ComponentModel.DataAnnotations.Validator?displayProperty=nameWithType>. The following rules are applied when validating an instance of type `T`:
+
+1. Member properties of `T` are validated, including recursively validating nested objects.
+1. Type-level attributes of `T` are validated.
+1. The <xref:System.ComponentModel.DataAnnotations.IValidatableObject.Validate%2A?displayProperty=nameWithType> method is executed, if `T` implements it.
```
oroztocil: While I think it's good that we document the validation logic so that the user's don't have to reverse engineer the rules themselves, I would put it somewhere later as it's a more niche, low-level detail that it only interesting to a limited audience.

aspnetcore/release-notes/aspnetcore-10/includes/blazor.md
```diff
@@ -587,6 +589,19 @@ In the following `OrderPage` component, the <xref:Microsoft.AspNetCore.Component
 
 The requirement to declare the model types outside of Razor components (`.razor` files) is due to the fact that both the new validation feature and the Razor compiler itself are using a source generator. Currently, output of one source generator can't be used as an input for another source generator.
 
+Validation support now includes:
+
+* Placing validation attributes on classes and records.
+* The new `[SkipValidation]` attribute to omit selected properties, parameters, or types from validation. When applied to a property or a method parameter, the validator skips that value during validation. When applied to a type, the validator skips all properties and parameters of that type. This can be useful, in particular, when using the same model types in cases that require and don't require validation.
```
oroztocil: This is still written more as a "delta" from the previous .NET 10 preview release, not as a delta between .NET 9 and .NET 10 (from the user's perspective).

The actual improvements between .NET 9 and .NET 10 for users:

1. Validation of nested complex objects and collections is now supported.
    - This includes validation rules defined by property attributes, class attributes, and `IValidatableObject` implementation.
    - The new SkipValidation attribute was added to solve a problem that previously did not exist (nested objects were not validated in any case).
2. Validation now uses a source generator-based implementation instead of reflection-based implementation for improved performance and compatibility with AoT compilation.

guardrex: The following seems nonsensical. Is the following the correct phrasing (emphasis added)?

> The new `[SkipValidation]` attribute was added ***to solve a problem that previously did not exist*** (nested objects were not validated in any case).

I understand that because nested objects weren't validated that the attribute had no effect on such objects.

How about something like ...

> The `[SkipValidation]` attribute can control validation with the new nested complex object validation support.

aspnetcore/blazor/host-and-deploy/configure-trimmer.md
danroth27: @gaurdrex Can you create an issue for this in the dotnet/aspnetcore repo and share your project with us?

aspnetcore/release-notes/aspnetcore-10/includes/blazor.md
danroth27: > The [SkipValidation] attribute can control validation with the new nested complex object validation support.

I would be more specific about what "control" means", that it's for excluding properties or types from validation.

guardrex: > The `[SkipValidation]` attribute can exclude properties or types from validation.

aspnetcore/blazor/host-and-deploy/configure-trimmer.md
guardrex: @danroth27 ... Isn't this the same as (or at least has the elements of) https://github.com/dotnet/aspnetcore/issues/52947? It's just not `KeyValuePair` ... it's a `Tuple` ... but same situation/error. I thought @javiercn was well aware of this, and I placed the content that he requested to cover it. I just want to make sure I'm opening an issue that @javiercn won't close saying that it can never be fixed outside of the approaches provided ... in order: use a custom type (any .NET), Dynamic Dependency (.NET 5+), Root Descriptor (.NET 10+), `<_ExtraTrimmerArgs>` (.NET 8 workaround).

guardrex: ... and @danroth27 ... are we holding this PR past the RC1 release if @javiercn doesn't respond by then?

danroth27: >Isn't this the same as (or at least has the elements of) https://github.com/dotnet/aspnetcore/issues/52947? It's just not KeyValuePair ... it's a Tuple ... but same situation/error. I thought @javiercn was well aware of this, and I placed the content that he requested to cover it.

Right, that's why I thought @javiercn might want to review this content.

> are we holding this PR past the RC1 release if @javiercn doesn't respond by then?

Nope, no need to block on this. The content looks fine to me.

guardrex: I just asked because you were calling for his review. I'll react with a patch PR later if he has any changes.

... and I thought Javier would just re-open the existing PU issue on that subject, but I'm still 👂 if you/he want a new issue on `ConstructorContainsNullParameterNames`.

javiercn: We did KeyValuePair because people use that.

If get reports for other types, we might choose to add those too. The cost of preserving a type is small compared to the benefit of not having to force people into doing this type of stuff if enough people use the type.

aspnetcore/blazor/forms/validation.md
danroth27, line 1: I think we need to also get rid of the "Blazor data annotations validation package" section in the .NET 10 version of this doc. The only mention of that old experimental package should be to tell user to stop using it.

aspnetcore/blazor/fundamentals/routing.md
```diff
@@ -704,6 +704,31 @@ The following component:
...
+
+## Navigation Manager redirect behavior during static server-side rendering (static SSR)
+
+For a redirect during static server-side rendering (static SSR), <xref:Microsoft.AspNetCore.Components.NavigationManager> relies on throwing a <xref:Microsoft.AspNetCore.Components.NavigationException> that gets captured by the framework, which converts the error into a redirect. Code that exists after the call to <xref:Microsoft.AspNetCore.Components.NavigationManager.NavigateTo%2A> isn't called. When using Visual Studio, the debugger breaks on the exception, requiring you to deselect the checkbox for **Break when this exception type is user-handled** in the Visual Studio UI to avoid the debugger stopping for future redirects.
+
+:::moniker-end
+
+:::moniker range=">= aspnetcore-10.0"
+
+You can use the `<BlazorDisableThrowNavigationException>` MSBuild property set to `true` in the app's project file to opt-in to no longer throwing a <xref:Microsoft.AspNetCore.Components.NavigationException>. This behavior is enabled by default in the .NET 10 or later Blazor Web App project template:
```
danroth27, line 717: It's probably also worth mentioning that this means code after the navigation will now run when it wouldn't have before.

PR #36094, opened 2025-09-10
[Blazor] Clarify trimming behavior for custom types

hakenr:
I think saying "never trimmed" is a bit too strong, since you can still enable trimming for your own assemblies and types.

### DIFF
--- a/aspnetcore/blazor/host-and-deploy/configure-trimmer.md
+++ b/aspnetcore/blazor/host-and-deploy/configure-trimmer.md
@@ -96,7 +96,7 @@ To address lost types, consider adopting one of the following approaches.
 
 ### Custom types
 
-Custom types aren't trimmed by Blazor when an app is published, so we recommend using custom types for JS interop, JSON serialization/deserialization, and other operations that rely on reflection.
+Custom types aren't trimmed by Blazor when an app is published (unless explicitly opted in), so we recommend using custom types for JS interop, JSON serialization/deserialization, and other operations that rely on reflection.
 
 The following modifications create a `StringTuple` type for use by the component.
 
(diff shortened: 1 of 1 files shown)


guardrex:
> "never trimmed" is a bit too strong, since you can still enable trimming for your own assemblies and types

I don't think it's a significant problem for the sentence because "never trimmed" is followed by "by Blazor," indicating that it's referring to the framework. However, changing it is fine.👍 

WRT ...

> unless explicitly opted in

Would one create a type that's only used during dev/test and not by the published app?

Also, we don't cover opting-into trimming in dev assemblies for published apps. Even if we add a remark on it, we need to indicate how it's even done in the first place.

~This is leaning a bit in the direction of an issue first for discussion~, but we could leave this PR open here and come back to it after things are sorted out on an issue.

I opened an issue. Let's discuss further on the issue.

guardrex:
Oh! ... and while we're working on this, there's one little NIT 😈 that we can fix along the way here.

The list items in the blockquote are on one line ...

• 1:T1, 1:T2 • 2:T2, 2:T2

They should be on two lines ...

• 1:T1, 1:T2 
• 2:T2, 2:T2

That can be achieved by placing two spaces after the "T2" on the first block quote line.

Let's do that before we merge any changes here.

I think I should also fix the exception in the same way. Two-spaces added to each of the first two lines will fix it.

I'll edit this PR later and fix these items directly. No need for you to take action on them.

hakenr:
> Would one create a type that's only used during dev/test and not by the published app?

It's not just a dev/test scenario. For example, a custom type in a reusable class library (not necessarily a widely used NuGet package, but most dev teams create their own helper libraries used across all their projects).

> Also, we don't cover opting-into trimming in dev assemblies for published apps. Even if we add a remark on it, we need to indicate how it's even done in the first place.

That would be nice. :-D

danroth27 (approved):
LGTM

guardrex:
Thanks @danroth27!

I'll make a final edit now to fix that list items problem and get this merged.

PR #36111, opened 2025-09-16
Blazor prerendering edits

danroth27:
Proposing some simplifications to the Blazor prerendering doc.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/components/prerender.md](https://github.com/dotnet/AspNetCore.Docs/blob/d9da4601de62f730680e2c44886fafcaf2237c36/aspnetcore/blazor/components/prerender.md) | [aspnetcore/blazor/components/prerender](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/components/prerender?branch=pr-en-us-36111) |

### DIFF
--- a/aspnetcore/blazor/components/prerender.md
+++ b/aspnetcore/blazor/components/prerender.md
@@ -191,6 +191,6 @@ In the server project's `Program` file, register the service:
 builder.Services.TryAddScoped<IWebAssemblyHostEnvironment, ServerHostEnvironment>();
 ```
 
-At this point, the <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> service can be injected into an interactive WebAssembly or Auto component that is also prerendered from the server.
+At this point, the <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> service can be [injected into an interactive WebAssembly or Auto component that is also prerendered from the server](xref:blazor/fundamentals/environments#read-the-environment-in-a-blazor-webassembly-app).
 
 :::moniker-end
--- a/aspnetcore/blazor/components/prerender.md
+++ b/aspnetcore/blazor/components/prerender.md
@@ -25,7 +25,7 @@ This article explains Razor component prerendering scenarios for server-rendered
 
 Prerendering is enabled by default for interactive components.
 
-Internal navigation for interactive routing doesn't involve requesting new page content from the server. Therefore, prerendering doesn't occur for internal page requests, including for [enhanced navigation](xref:blazor/fundamentals/routing#enhanced-navigation-and-form-handling). For more information, see [Static versus interactive routing](xref:blazor/fundamentals/routing#static-versus-interactive-routing), [Interactive routing and prerendering](xref:blazor/state-management/prerendered-state-persistence#interactive-routing-and-prerendering), and [Enhanced navigation and form handling](xref:blazor/fundamentals/routing#enhanced-navigation-and-form-handling).
+Internal navigation with interactive routing doesn't use prerendering because the page is already interactive. For more information, see [Static versus interactive routing](xref:blazor/fundamentals/routing#static-versus-interactive-routing) and [Interactive routing and prerendering](xref:blazor/state-management/prerendered-state-persistence#interactive-routing-and-prerendering).
 
 [`OnAfterRender{Async}` component lifecycle events](xref:blazor/components/lifecycle#after-component-render-onafterrenderasync) aren't called when prerendering, only after the component renders interactively.
 
@@ -103,7 +103,28 @@ No compile time error occurs, but a runtime error occurs during prerendering:
 
 This error occurs because the component must compile and execute on the server during prerendering, but <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> isn't a registered service on the server.
 
-If the app doesn't require the value during prerendering, this problem can be solved by injecting <xref:System.IServiceProvider> to obtain the service instead of the service type itself:
+There are several approaches that you can take to address this scenario:
+
+* If the service supports server execution, register the service on the server in addition to the client so that it's available during prerendering. For an example of this scenario, see the guidance for <xref:System.Net.Http.HttpClient> services in the [Blazor Web App external web APIs](xref:blazor/call-web-api#blazor-web-app-external-web-apis) section of the *Call web API* article.
+
+* Make the service optional if it isn't always needed during prerendering.
+
+* If a different service implementation is needed on the server, create a service abstraction and create implementations for the service in the `.Client` and server projects. Register the services in each project. Inject the custom service abstraction in the component.
+
+* Disable prerendering for the component or for the entire app.
+
+To optionally inject the service into the component, use constructor injection:
+
+```csharp
+private string? environmentName;
+
+public Home(IWebAssemblyHostEnvironment? env = null)
+{
+    environmentName = env?.Environment;
+}
+```
+
+Alternatively, inject <xref:System.IServiceProvider> to optionally obtain the service if it's available:
 
 ```razor
 @page "/"
@@ -131,18 +152,7 @@ If the app doesn't require the value during prerendering, this problem can be so
 }
 ```
 
-If you merely want to make the service injection optional, you can use constructor injection:
-
-```csharp
-private string? environmentName;
-
-public Home(IWebAssemblyHostEnvironment? env = null)
-{
-    environmentName = env?.Environment;
-}
-```
-
-Another option for obtaining the environment whether the code is running on the server or on the client is to inject <xref:Microsoft.Extensions.Hosting.IHostEnvironment> from the [`Microsoft.Extensions.Hosting.Abstractions` NuGet package](https://www.nuget.org/packages/Microsoft.Extensions.Hosting.Abstractions). Add a package reference to the app and use the following approach:
+To obtain the environment whether the code is running on the server or on the client, you could optionally inject <xref:Microsoft.Extensions.Hosting.IHostEnvironment> from the [`Microsoft.Extensions.Hosting.Abstractions` NuGet package](https://www.nuget.org/packages/Microsoft.Extensions.Hosting.Abstractions):
 
 ```csharp
 private string? environmentName;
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/components/prerender.md
```diff
@@ -25,7 +25,7 @@ This article explains Razor component prerendering scenarios for server-rendered
 
 Prerendering is enabled by default for interactive components.
 
-Internal navigation for interactive routing doesn't involve requesting new page content from the server. Therefore, prerendering doesn't occur for internal page requests, including for [enhanced navigation](xref:blazor/fundamentals/routing#enhanced-navigation-and-form-handling). For more information, see [Static versus interactive routing](xref:blazor/fundamentals/routing#static-versus-interactive-routing), [Interactive routing and prerendering](xref:blazor/state-management/prerendered-state-persistence#interactive-routing-and-prerendering), and [Enhanced navigation and form handling](xref:blazor/fundamentals/routing#enhanced-navigation-and-form-handling).
+Internal navigation with interactive routing doesn't use prerendering because the page is already interactive. For more information, see [Static versus interactive routing](xref:blazor/fundamentals/routing#static-versus-interactive-routing) and [Interactive routing and prerendering](xref:blazor/state-management/prerendered-state-persistence#interactive-routing-and-prerendering).
```
danroth27, line 28: Content might still come from the server when using interactive server rendering, and prerendering can still happen when doing an enhanced navigation.

aspnetcore/blazor/components/prerender.md
```diff
@@ -103,7 +103,28 @@ No compile time error occurs, but a runtime error occurs during prerendering:
 
 This error occurs because the component must compile and execute on the server during prerendering, but <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> isn't a registered service on the server.
 
-If the app doesn't require the value during prerendering, this problem can be solved by injecting <xref:System.IServiceProvider> to obtain the service instead of the service type itself:
+There are several approaches that you can take to address this scenario:
```
danroth27, line 106: I think it helps to list the available approaches up front instead of at the end.

aspnetcore/blazor/components/prerender.md
```diff
@@ -103,7 +103,28 @@ No compile time error occurs, but a runtime error occurs during prerendering:
...
+
+* If the service supports server execution, register the service on the server in addition to the client so that it's available during prerendering. For an example of this scenario, see the guidance for <xref:System.Net.Http.HttpClient> services in the [Blazor Web App external web APIs](xref:blazor/call-web-api#blazor-web-app-external-web-apis) section of the *Call web API* article.
+
+* Make the service optional if it isn't always needed during prerendering.
+
+* If a different service implementation is needed on the server, create a service abstraction and create implementations for the service in the `.Client` and server projects. Register the services in each project. Inject the custom service abstraction in the component.
+
+* Disable prerendering for the component or for the entire app.
+
+To optionally inject the service into the component, use constructor injection:
```
danroth27, line 116: I think showing the constructor injection approach is simpler than using `IServiceProvider`, so I think it should be suggested first.

aspnetcore/blazor/components/prerender.md
```diff
@@ -155,27 +165,11 @@ public Home(IHostEnvironment? serverEnvironment = null,
...
 
-Because the server doesn't have a registered <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> service, it isn't possible to inject the service and use the service implementation's host environment extension methods and properties during server prerendering. Injecting the service into an interactive WebAssembly or Auto component results in the following runtime error:
+A better approach is to create a custom service abstraction and implementations for the server and client. Implement the service interface in the server and client projects and register the implementations in each project. The component then depends solely on the custom service abstraction.
 
-> :::no-loc text="There is no registered service of type 'Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment'.":::
-
-To address this, create a custom service implementation for <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> on the server. Add the following class to the server project.
+In the case of `IWebAssemblyHostEnvironment`, we can reuse the existing interface instead of creating a new one:
```
danroth27, line 172: I feel a bit icky about suggesting users implement and register `IWebAssemblyHostEnvironment` on the server. The server certainly isn't running on WebAssembly! But maybe for simplicity and demonstration purposes it's ok? Alternatively, we could show a data retrieval service that involves an API call on the client and direct data access on the server.

aspnetcore/blazor/components/prerender.md
```diff
@@ -197,85 +191,6 @@ In the server project's `Program` file, register the service:
 builder.Services.TryAddScoped<IWebAssemblyHostEnvironment, ServerHostEnvironment>();
 ```
 
-At this point, the <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> service can be injected into an interactive WebAssembly or Auto component and used as shown in the [Read the environment in a Blazor WebAssembly app](xref:blazor/fundamentals/environments#read-the-environment-in-a-blazor-webassembly-app) section of the *Environments* article.
-
-The preceding example can demonstrate that it's possible to have a different server environment than client environment, which isn't recommended and may lead to arbitrary results. When setting the environment in a Blazor Web App, it's best to match server and `.Client` project environments. Consider the following scenario in a test app:
```
danroth27, line 202: I don't think the prerendering article is really the place to discuss the problems that might happen if you somehow set up a different environment on the client and the server. If we think this content is important to keep, maybe it should go in one of the environment docs?

aspnetcore/blazor/components/prerender.md
```diff
@@ -197,85 +191,6 @@ In the server project's `Program` file, register the service:
...
-     possibly leading to dropping the index in favor of the filtered 
-     search engine link.
-     
-     Note that the approach won't work in Bing, only Google. -->
-
-Prerendering guidance is organized by feature. Use the following index to find specific guidance on prerendering. Alternatively, use a search engine with a filter to the Blazor documentation node. To obtain English-US results, search with `site:learn.microsoft.com/en-us/aspnet/core/ "prerendering"`. Change the locale segment (`en-us`) for a different language/region.
-
-[Google filtered search results for "prerendering" (English-US articles)](https://www.google.com/search?q=site:learn.microsoft.com/en-us/aspnet/core/+%22prerendering%22)
-
-* Fundamentals
```
danroth27, line 233: I haven't gone through all these links yet, but I still prefer to remove them.

aspnetcore/blazor/components/prerender.md
guardrex: IIRC, @MackinnonBuck and I worked on this language for the internal navigation behavior. The enhanced nav part is only a bit of added info here to explain that it has no bearing on the behavior (the topic sentence is the second sentence of the paragraph), not that prerendering can't happen with enhanced nav. @MackinnonBuck may remark on this since (IIRC) he worked with me on the original language. IIRC/AFAIK, he said that this paragraph is accurate.

aspnetcore/blazor/components/prerender.md
guardrex: ~I'll move this bit over to the *Environments* article for the next commit and round of reviews there.~

Actually, I opened #36112 to work on it. I probably need to modify the example, and there might be cross-links to add. Best to work it separately.

aspnetcore/blazor/components/prerender.md
```diff
@@ -197,85 +191,6 @@ In the server project's `Program` file, register the service:
...
-When prerendering occurs on the server, the component is rendered in the `Development` environment:
-
-> **:::no-loc text="Environment:":::** :::no-loc text="Development":::
-
-When the component is rerendered just a second or two later, after the Blazor bundle is downloaded and the .NET WebAssembly runtime activates, the value changes to reflect that the client is operating in the `Staging` environment on the client:
-
-> **:::no-loc text="Environment:":::** :::no-loc text="Staging":::
-
-The preceding example demonstrates why we recommend setting the server environment to match the client environment for development, testing, and production deployments.
+At this point, the <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> service can be injected into an interactive WebAssembly or Auto component that is also prerendered from the server.
```
guardrex:
```suggestion
At this point, the <xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment> service can be [injected into an interactive WebAssembly or Auto component that is also prerendered from the server](xref:blazor/fundamentals/environments#read-the-environment-in-a-blazor-webassembly-app).
```

aspnetcore/blazor/components/prerender.md
guardrex: Ok, and you're also not keen on the (better-than-docs filtering) Google filtered search ...

https://www.google.com/search?q=site:learn.microsoft.com/en-us/aspnet/core/+%22prerendering%22

... to help devs find content on the subject? That was the point of the index here. It was never intended to be used as a learning path, just an index.

aspnetcore/blazor/components/prerender.md
guardrex: If this is kept, I'm concerned about removing my lead-in, which I think is better for newbies. Also, the example error can helpful when searching the doc set to find this content.

It seems like each of these approaches should be under an H3 heading to better organize them. Perhaps, we could look at a commit that would do that after these edits are merged into the PR.

Otherwise, I'm 👂 to hear if we're keeping the existing-interface approach.

aspnetcore/blazor/components/prerender.md
guardrex: Oh! I see your remark ...

> I don't like the filtered search either 😄. Why should I need to search for prerendering related content from the article on prerendering?
>
> I still don't believe the index is usable or needed.

... because the dev can't remember exactly where they read something. It's just an index to help find things. Without the index and without a filtered search, it's more difficult to track something down.

Let's kill it ... *No more Halloween* 🎃 *nightmares of indexes chasing developers* 🔪💀 *through the woods!* 🤣

aspnetcore/blazor/components/prerender.md
MackinnonBuck: I think the new wording is an improvement - it more concisely conveys what the rules are around when prerendering happens.

MackinnonBuck: > The enhanced nav part is only a bit of added info here to explain that it has no bearing on the behavior

This is true in that the enhanced navigation feature does not come into play during interactive navigation. However, the current wording makes it sound like prerendering doesn't happen during enhanced navigation, which isn't accurate.

aspnetcore/blazor/components/prerender.md
danroth27: We introduce the error and solutions at the beginning of "Client-side services fail to resolve during prerendering" section, so I don't think we need to repeat the error again here. I'm open to introducing some h3 headings to break up this content. I'll go ahead and merge this and we can discuss that on your PR.

danroth27: And let's just go with the `IWebAssemblyHostEnvironment` based example for having a common service interface. We can change it if someone else complains about it.

PR #36170, opened 2025-09-30
Minor paraphrasing

ahmadalshobaiki:
I think this is a more direct and clear description of the DI process during runtime


guardrex:
Hello @ahmadalshobaiki ... Please open an issue for this from the bottom of the article. Use the **Open a documentation issue** link. This suggested change will need to be made in multiple files, and the metadata for the issue that you open will ping the correct doc author. Thanks!

PR #36178, opened 2025-10-02
Globalization/ICU file updates

### LINKED ISSUE #36155: Globalization docs cleanup

# Issue description

Our globalization-localization docs are missing a few bits that are in https://github.com/dotnet/runtime/blob/89bba37a92053f52e38d5a5c81a1f3510319dabf/docs/design/features/globalization-icu-wasm.md.

Suggested location to add these bits:
 https://learn.microsoft.com/en-us/aspnet/core/blazor/globalization-localization
OR
https://learn.microsoft.com/en-us/dotnet/core/extensions/globalization-icu
(not sure what is neater)

We want to move the information about:
- shards (`icudt.*.dat`), the data each shard contains (1st paragraph)
- The 2nd paragraph can be copied but using blazor propery `BlazorIcuDataFileName` instead of `WasmIcuDataFileName`. The custom ICU can be mentioned with a link to the github doc - https://github.com/dotnet/runtime/blob/89bba37a92053f52e38d5a5c81a1f3510319dabf/docs/design/features/globalization-icu-wasm.md#custom-icu. It is considered a workaround way for corner cases and manual ICU build is not guaranteed to work with all filters configuration, so I would prefer to treat it as instruction for "advanced" users.
- The paragraph describing what `BlazorWebAssemblyLoadAllGlobalizationData` does.

# Software versions

It should be applied to version .NET Core 8 and later.

Other remarks:
- https://learn.microsoft.com/en-us/dotnet/core/extensions/globalization-icu#icu-on-webassembly is very hard to understand. I am suspecting it contains limitations that were valid for an experimental `HybridGlobalization` feature that was disabled in https://github.com/dotnet/runtime/pull/110567. It should be removed for net10. If you choose to place the blazor globalization switches update there, it can replace fully "icu-on-webassembly" paragraph.
- In case we would be adding the updates in "globalization-icu", let's add a cross link to it to "blazor/globalization-localization" and the other way around.

---
#### Document Details

* Platform ID: bd2a8fe1-da71-2bfb-045d-1b5f49d944a4
* Content: [ASP.NET Core Blazor globalization and localization](https://learn.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-9.0#net-globalization-and-international-components-for-unicode-icu-support-blazor-webassembly)
* Service: **aspnet-core**
* Sub-service: **blazor**
* GitHub Login: @guardrex
* Microsoft Alias: **wpickett**

guardrex:
Fixes #36155

See if this will work to fix the content in this doc set.

For updates to https://learn.microsoft.com/en-us/dotnet/core/extensions/globalization-icu, an issue would need to be opened from that article ... it goes to a different doc author on that doc team.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/globalization-localization.md](https://github.com/dotnet/AspNetCore.Docs/blob/a8cbbb22dddfceb84198ff5e89ecc5113527d488/aspnetcore/blazor/globalization-localization.md) | [aspnetcore/blazor/globalization-localization](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?branch=pr-en-us-36178) |

### DIFF
--- a/aspnetcore/blazor/globalization-localization.md
+++ b/aspnetcore/blazor/globalization-localization.md
@@ -5,7 +5,7 @@ description: Learn how to render globalized and localized content to users in di
 monikerRange: '>= aspnetcore-3.1'
 ms.author: wpickett
 ms.custom: mvc
-ms.date: 11/12/2024
+ms.date: 10/02/2025
 uid: blazor/globalization-localization
 ---
 # ASP.NET Core Blazor globalization and localization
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/globalization-localization.md
```diff
@@ -75,15 +75,40 @@ The following field types have specific formatting requirements and aren't suppo
 
 For current browser support of the preceding types, see [Can I use](https://caniuse.com).
 
+By default, Blazor only loads globalization data for the app's culture. To load all globalization data, set `<BlazorWebAssemblyLoadAllGlobalizationData>` to `true` in the app's project file (`.csproj`):
```
ilonatommy: I don't like the word "only". Because it "only" loads up to a hundred of cultures, out of which one is the app's culture.

The second sentence is fully true but I have a problem with forming the 1st one so that it does not sound misleading and at the same time it won't be overwhelming when it comes to details.

We are mentioning the details of shards later in this doc, so maybe something like: `By default, Blazor loads a subset of globalization data that contains the app's culture.`?

Or maybe a "shard of globalization data", so that we use the same wording everywhere?

guardrex:
```suggestion
By default, Blazor loads a subset of globalization data that contains the app's culture. To load all globalization data, set `<BlazorWebAssemblyLoadAllGlobalizationData>` to `true` in the app's project file (`.csproj`):
```

guardrex: "Shard" isn't defined, so I like your suggestion.

guardrex:
I'll merge this as soon as we get the sample live.

guardrex:
I'll go ahead. This one doesn't refer to the samp. We're good! 👍

PR #36186, opened 2025-10-06
Fix SetDefaultPolicy being used like SetFallbackPolicy

KaisoBits:
The description mistakenly says `SetDefaultPolicy` twice, despite describing the behavior of `SetFallbackPolicy` in the first sentence and then the behavior of `SetDefaultPolicy` in the next sentence. I suppose this is a typo.

What it used to say is false, as changing `SetDefaultPolicy` will never impact endpoints without an `[Authorize]` attribute.

This is especially dangerous, as someone copying the example code may just naively assume their API is already secure without using `[Authorize]`, while it obviously isn't.

Additionally, changed the example below to show how to require authenticated user by default, similar to how it's presented in the linked require-authenticated-users article section.

This was done as the previous example using `SetDefaultFallback` essentially did nothing, as requiring authenticated users is what the default policy does already with nothing set, so the old example is of no use.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/security/authentication/configure-jwt-bearer-authentication.md](https://github.com/dotnet/AspNetCore.Docs/blob/46880d0300af3f12fcc916ee9dfa65295a39e011/aspnetcore/security/authentication/configure-jwt-bearer-authentication.md) | [Configure JWT bearer authentication in ASP.NET Core](https://review.learn.microsoft.com/en-us/aspnet/core/security/authentication/configure-jwt-bearer-authentication?branch=pr-en-us-36186) |

### DIFF
--- a/aspnetcore/security/authentication/configure-jwt-bearer-authentication.md
+++ b/aspnetcore/security/authentication/configure-jwt-bearer-authentication.md
@@ -180,7 +180,7 @@ APIs often need to accommodate access tokens from various issuers. Supporting mu
 
 ### Forcing the bearer authentication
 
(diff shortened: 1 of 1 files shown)


guardrex:
Ping @damienbod for a look 👀.

damienbod:
Thanks @KaisoBits for the fix, LGTM.

PR #36231, opened 2025-10-17
[Blazor] Handle Errors - CascadingValue to include IsFixed attribute

hakenr:
`this` never changes, so `IsFixed="true"` should be used to avoid subscribing to changes (PERF)

https://learn.microsoft.com/en-us/aspnet/core/blazor/performance/rendering?view=aspnetcore-9.0#ensure-cascading-parameters-are-fixed

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/handle-errors.md](https://github.com/dotnet/AspNetCore.Docs/blob/aa0324d6f09d08a246b5d9b5ca15484b10306e81/aspnetcore/blazor/fundamentals/handle-errors.md) | [aspnetcore/blazor/fundamentals/handle-errors](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/handle-errors?branch=pr-en-us-36231) |

### DIFF
--- a/aspnetcore/blazor/fundamentals/handle-errors.md
+++ b/aspnetcore/blazor/fundamentals/handle-errors.md
@@ -555,6 +555,8 @@ The following `ProcessError` component example merely logs errors, but methods o
 
 > [!NOTE]
 > For more information on <xref:Microsoft.AspNetCore.Components.RenderFragment>, see <xref:blazor/components/index#child-content-render-fragments>.
+>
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks @hakenr! 🎷 ... Do you feel good about the passing remark that I added to the existing NOTEs under the examples? We could cross-link to the article that covers it, but I think a quick remark on it in passing is fine here.

hakenr (commented):
@guardrex The remark is correct.

PR #36234, opened 2025-10-19
Update wasm-tools installation instructions for .NET 8

### LINKED ISSUE #36235: Update wasm-tools installation instructions for .NET 8
### Description

Consider adding a note about .NET WebAssembly build tools compatibility and installation instructions for .NET 8 projects.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/webassembly-build-tools-and-aot?view=aspnetcore-9.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/webassembly-build-tools-and-aot.md

### Document ID

bbbdf15e-1918-cee9-ff03-0ebbca80e86e

### Platform Id

0a8e8b8e-2e88-684e-022e-56238545d038

### Article author

@guardrex

### Metadata

* ID: bbbdf15e-1918-cee9-ff03-0ebbca80e86e
* PlatformId: 0a8e8b8e-2e88-684e-022e-56238545d038 
* Service: **aspnet-core**
* Sub-service: **blazor**

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+bbbdf15e-1918-cee9-ff03-0ebbca80e86e)

candritzkynet:
[EDIT by guardrex to add the issue number]

Fixes #36235

Added note about .NET WebAssembly build tools compatibility and installation instructions for .NET 8 projects.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/webassembly-build-tools-and-aot.md](https://github.com/dotnet/AspNetCore.Docs/blob/b46f26b34bc4260e3f050bee0648d3f34e309078/aspnetcore/blazor/webassembly-build-tools-and-aot.md) | [ASP.NET Core Blazor WebAssembly build tools and ahead-of-time (AOT) compilation](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/webassembly-build-tools-and-aot?branch=pr-en-us-36234) |

### DIFF
--- a/aspnetcore/blazor/webassembly-build-tools-and-aot.md
+++ b/aspnetcore/blazor/webassembly-build-tools-and-aot.md
@@ -32,6 +32,17 @@ The .NET WebAssembly build tools are based on [Emscripten](https://emscripten.or
 > dotnet workload install wasm-tools-net6
 > ```
 
+> [!NOTE]
+> .NET WebAssembly build tools for .NET 8 projects
+>
+> The `wasm-tools` workload installs the build tools for the latest release. However, the current version of the build tools are incompatible with existing projects built with .NET 8. Projects using the build tools that must support both .NET 8 and a later release must use multi-targeting.
+>
+> Use the `wasm-tools-net8` workload for .NET 8 projects when developing apps with the .NET 9 SDK. To install the `wasm-tools-net8` workload, execute the following command from an administrative command shell:
+>
+> ```dotnetcli
+> dotnet workload install wasm-tools-net8
+> ```
+
 ## Ahead-of-time (AOT) compilation
 
 Blazor WebAssembly supports ahead-of-time (AOT) compilation, where you can compile your .NET code directly into WebAssembly. AOT compilation results in runtime performance improvements at the expense of a larger app size.


guardrex:
Hello @candritzkynet ... Please keep our guidelines in mind that we require an issue before a PR is submitting for anything other than a patch change. I'll open an issue for this now, so no worries about setting one up for this. The best way to open an issue for us in the future is from the bottom of the article using the **Open a documentation issue** link, which adds tracking metadata. When a PR goes up, it gets a "Fixes {ISSUE NUMBER}" line in the opening post. I'll take care of that, too.

**UPDATE**: *Done!* 👍

guardrex:
@javiercn ... I didn't think that this was a problem for .NET 8 as it was for .NET 6, where we have a special note for .NET 6 (purple NOTE box at) ...

https://learn.microsoft.com/en-us/aspnet/core/blazor/webassembly-build-tools-and-aot?view=aspnetcore-9.0#net-webassembly-build-tools

If devs are supposed to install matched build tools, then I want to cover this differently than adding a NOTE for each release. Is this ***pattern*** correct? ...

* `dotnet workload install wasm-tools` for the latest release
* `wasm-tools-net9` for .NET 9 (after .NET 10 releases)
* `wasm-tools-net8` for .NET 8
* `wasm-tools-net7` for .NET 7
* `wasm-tools-net6` for .NET 6

guardrex:
cc: @maraf ... Can you shed any light on this WRT matching WASM build tools to the app's TFM? See my last comment ☝️.

maraf:
The table is about SDK version and TFM

- .NET 10 SDK 
  - targeting .NET 10 requires wasm-tools
  - targeting .NET 9 requires wasm-tools-net9
  - targeting .NET 8 requires wasm-tools-net8
- .NET 9 SDK
  - targeting .NET 9 requires wasm-tools
  - targeting .NET 8 requires wasm-tools-net8
- .NET 8 SDK
  - targeting .NET 8 requires wasm-tools

The workload always installs into the SDK. If you install wasm-tools workload into .NET 9 SDK, the .NET 10 SDK won't have it. You'll need to install it again. Although multiple rows contain wasm-tools-net8 workload, it fact it is always a slightly different workload for the particular .NET SDK.

If you install the workload using checkbox in Visual Studio setup, the VS installer know about the workload and when it installs newer .NET SDK, it will install the workload automatically.

Let me know if anything needs more detail

guardrex:
Thanks for that @maraf. That's great!

@candritzkynet ... This is going to require a bit more work/thought on how to provide and version the content than a patch edit can handle here. I don't want to cover this in a series of NOTEs that we keep adding to the coverage each year. Unfortunately, I won't be able to accept this PR. I'll need to work on this one myself. However, thanks for bringing it to my attention.

PR #36311, opened 2025-11-09
Clarify the Blazor response streaming heading in the What's New doc

danroth27:
The current heading sounds backwards: response streaming is now enabled by default and is no longer opt-in. Here's my attempt to clarify the heading.


guardrex (approved):
👍

PR #36355, opened 2025-11-14
Not-latest includes updated for post .NET 10

### LINKED ISSUE #36356: Update Note-latest-version includes
The not-latest-version includes need to be updated to account for .NET 10 GA

- aspnetcore/includes/not-latest-version.md
- aspnetcore/includes/not-latest-version-without-not-supported-content.md

wadepickett:
Fixes #36356 

Update the Not-latest.. includes to account for .NET 10 GA

### DIFF
--- a/aspnetcore/includes/not-latest-version-without-not-supported-content.md
+++ b/aspnetcore/includes/not-latest-version-without-not-supported-content.md
@@ -26,6 +26,5 @@ When a new version is released, it might be necessary to temporarily comment out
 moniker range section until the new moniker is created.
 
 Markdown to include this file:
-+
 [!INCLUDE[](~/includes/not-latest-version-without-not-supported-content.md)]
 -->


tdykstra (approved):
I think you're right, we should add 11 now.

wadepickett:
@guardrex, do you know how to get around the catch for this situation. (See build errors) It seems early to list a  v 11 in the selector list (I would think) , but I suspect referring to > 10.0 in moniker range may be the issue.  Or maybe I am missing something else?

wadepickett:
@tdykstra & @guardrex.  I think the solution here is to exclude > v10 until the the next preview version is closer. then we add that version to the not-latest-version includes and to the version selector

wadepickett:
@tdykstra, made a change since you looked last.  I excluded > v10 until we are ready to add a new forward looking preview version in the version selector.

Handy test link:
https://review.learn.microsoft.com/en-us/aspnet/core/blazor/advanced-scenarios?view=aspnetcore-10.0&branch=pr-en-us-36355

guardrex:
Sorry @wadepickett that I wasn't able to respond last week. I ran out of hours ... and hit a 🧠🔥. It's that time of year again ... NO vacation for going on nine years will do that to you! 😆 ***BUT*** ... 🦃 Days and ⛄ holidays are just about here for me to recharge with the tiny bit of time off that I do get. Anyway, are you OK here with this? I wasn't planning on looking unless you need me to, as I'm still cranking away on the final 10.0 NITs and BITs this week.

PR #36364, opened 2025-11-17
RemoveNotice of Pre-release

### LINKED ISSUE #36360: notice of pre-release
Image

---
[Associated WorkItem - 509536](https://dev.azure.com/msft-skilling/Content/_workitems/edit/509536)

wadepickett:
Fixes #36360 

This is a follow up to PR #36355.   There were commented out lines regarding a notice of pre-release that inteneded to be used later when the enxt pre-relase comes along, and while those commented out lines do not show in the build review, they do for some reason in the build for live.

The issue has been identified before as a problem with `:::`  that get read anyway during the process even when commented out per guardrex.  So I removed any occurance of `:::` witin the commented out lines.

### DIFF
--- a/aspnetcore/includes/not-latest-version-without-not-supported-content.md
+++ b/aspnetcore/includes/not-latest-version-without-not-supported-content.md
@@ -4,12 +4,12 @@
 :::moniker-end
 
 <!-- Exclude until .NET 11 preview is added to the version selector collection
-(add tripple colon here)moniker range="> aspnetcore-10.0"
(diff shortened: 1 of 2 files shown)


wadepickett:
Verified proper display on review build:
.NET 10 = No note.
image

.NET 9 = Note that it is not the latest
image

.NET 8 = Note that it is not the latest
image

wadepickett:
I put the commented-out text, back in but removed the `:::` as recommended by @guardrex

PR #36390, opened 2025-11-24
Further work on standalone WASM environments

### LINKED ISSUE #36371: [Blazor] Clarify WasmApplicationEnvironmentName for build/publish scenarios
From discord communication at https://discord.com/channels/732297728826277939/732297874062311424/1440652633755881522:

> Fynzie — 11:39
Hi,
DId someone encounter a change regarding blazor wasm and net10 ? it seems the blazor wasm app is no longer fetching the correct appsettings file based on the servers's environment, resulting in the wasm app to always fetch the production appsettings (default) ?
I can't find anything talking about this change even tho it's a huge one 
Marek Fisera — 12:22
What blazor mode you using? Is it Blazor Web app? 
Fynzie — 12:31
blazor wasm, the old template from net5 or net6, I never migrated to the blazor united and such
but according to this https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-10.0?view=aspnetcore-10.0#set-the-environment-in-standalone-blazor-webassembly-apps its seems the old way was removed entierly and you have to either drive the configuration from your backend on startup or product as many build as environments you have
[What's new in ASP.NET Core in .NET 10](https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-10.0?view=aspnetcore-10.0)
Learn about the new features in ASP.NET Core in .NET 10.
Image
Marek Fisera — 12:52
Yes. In .NET 10 we have removed the boot config which was the only file that was requested every time you start the app, and so there is not reliable way to pass the actual environment from server to client (by the framework) 
For Blazor Web app we have a mechanism to get it to the client, because the document hosting the WebAssembly isn't static 
Fynzie — 13:14
thanks for the info, I suggest to enrich the doc with this info as I this was quite unexpected (at least for me)
jacalvar — 13:20
We should clarify that there is a way to do this during build/publish?
Fynzie — 13:24
The documentation talks about WasmApplicationEnvironmentName but I believe that what most people liked in the first place was the ability for a single output build to run on any environment. I personnaly defaulted to having a publicly accessible BE endpoint that is feeding the FE configuration

guardrex:
Fixes #36371

Marek ...

* I didn't have the cross-link to [our reference content in the *Environments* article](https://learn.microsoft.com/aspnet/core/blazor/fundamentals/environments?view=aspnetcore-10.0#set-the-environment) in the *What's New* article, so I'm adding that cross-link on this PR.
* I'm adding content on a few ways that a dev can control the value. These are common approaches for this sort of task, so I'm not testing them all. I'm 👂 if you have any concerns.
* I'm adding a bit on using a web API endpoint that the WASM app can hit. Should we show some startup JS code that would make that work? ... or is it best to just consider it an 'advanced scenario' for the dev to work out on their own?

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/environments.md](https://github.com/dotnet/AspNetCore.Docs/blob/acc8ff059983364d8014ae33db555203b096043e/aspnetcore/blazor/fundamentals/environments.md) | [aspnetcore/blazor/fundamentals/environments](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/environments?branch=pr-en-us-36390) |

### DIFF
--- a/aspnetcore/blazor/fundamentals/environments.md
+++ b/aspnetcore/blazor/fundamentals/environments.md
@@ -66,7 +66,7 @@ There are several approaches for setting the environment in a standalone Blazor 
   </PropertyGroup>
   ```
 
-* The environment can be set based on the use of a publish profile. The first condition in the following example sets the environment to `Development` when no publish profile is used (applies to both build and publish operations without a profile), while the second condition covers setting the environment to `Production` when any publish profile is used:
+* The environment can be set based on the use of a publish profile. In the following example, the first condition sets the environment to `Development` when no publish profile is used (applies to both build and publish operations without a profile), while the second condition covers setting the environment to `Production` when any publish profile is used:
 
   ```xml
   <PropertyGroup Condition="'$(PublishProfile)' == ''">


aspnetcore/blazor/fundamentals/environments.md
```diff
@@ -38,14 +38,50 @@ On the client for a Blazor Web App, the environment is determined from the serve
...
+  ```xml
+  <PropertyGroup>
+    <WasmApplicationEnvironmentName Condition="'$(PublishProfile)' != ''">
+      Production
+    </WasmApplicationEnvironmentName>
+
+    <WasmApplicationEnvironmentName Condition="'$(PublishProfile)' == ''">
+      Development
+    </WasmApplicationEnvironmentName>
+  </PropertyGroup>
```
maraf: nit: I would use the same syntax with PropertyGroup and Condition on it as above

aspnetcore/blazor/fundamentals/environments.md
```diff
@@ -38,14 +38,50 @@ On the client for a Blazor Web App, the environment is determined from the serve
...
+  ```
+
+* Create a custom server-side web API endpoint. The standalone Blazor WebAssembly app requests its environment from the web API either at app startup or on-demand while it's running. For general information on calling a web API from a Blazor app, see <xref:blazor/call-web-api>.
```
maraf: The call needs to happen the code reaches .NET, it needs to happen from JavaScript. Adding the link seems misleading.

The value should be passed to https://github.com/dotnet/aspnetcore/blob/main/src/Components/Web.JS/src/Platform/WebAssemblyStartOptions.ts#L7 or https://github.com/dotnet/aspnetcore/blob/3b674da67a2fcf9bb98120e94470eeb177c33e2f/src/Components/dotnet-runtime-js/dotnet.d.ts#L110

maraf (approved):
> I'm adding content on a few ways that a dev can control the value. These are common approaches for this sort of task, so I'm not testing them all. I'm 👂 if you have any concerns.

It looks good to me

> I'm adding a bit on using a web API endpoint that the WASM app can hit. Should we show some startup JS code that would make that work? ... or is it best to just consider it an 'advanced scenario' for the dev to work out on their own?

I think it's fine leaving it as an advanced scenario module the comment below

Otherwise, it looks great! Thank you

PR #36398, opened 2025-11-25
Move section and add guidance on plain class libs

### LINKED ISSUE #36363: Need guidance on `ValidatableTypeAttribute` for class library in documentation, or a fix for experimental attribute
### Description

I have a class library which needs the ability to specify `[ValidatableType]` on its models, assuming this is required for nested object validation.

After following the documentation for "Use validation models from a different assembly", I am required to use the "Experimental" attribute and suppress the error with pragma tags. Is this intentional? If so, please update the documentation to specifically instruct that people are required to do this, because it feels like this is an oversight when trying to set up class libraries with the new validation. 

With that said, I hope there's some better fix than simply putting pragma tags. Upgrading our class libraries to .net 10 with the new validation did not feel seamless or like an upgrade with the need for pragma tags on all of our classes with nested validation.

Sample:

```csharp
#pragma warning disable ASP0029 // dotnet 10 release notes instruct to use this attribute for nested type validation.
[Microsoft.Extensions.Validation.ValidatableType]
#pragma warning restore ASP0029 // dotnet 10 release notes instruct to use this attribute for nested type validation.
public class MyExampleParentClass {
    public class MyExampleChildClass { get; set; }
}

public class MyExampleChildClass {
    [Required]
    public string? ExampleString { get; set; }
}
```

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/forms/validation?view=aspnetcore-10.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/forms/validation.md

### Document ID

bcb470e1-c859-0549-b4ef-c6c9b7ef2c72

### Platform Id

b4c1bad9-fc7a-21f5-2552-f0f3617f77db

### Article author

@guardrex

### Metadata

* ID: bcb470e1-c859-0549-b4ef-c6c9b7ef2c72
* PlatformId: b4c1bad9-fc7a-21f5-2552-f0f3617f77db 
* Service: **aspnet-core**
* Sub-service: **blazor**

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+bcb470e1-c859-0549-b4ef-c6c9b7ef2c72)

guardrex:
Fixes #36363

Thanks @kgar! 🚀 

Ondřej ... I wrote it out using a lot of your language, which was very helpful. Thanks for taking the time to explain it. We can version this out at 11.0; but for now, I recommend making it clear and easy to address. It's a bit long, but I think writing it out will make it easier for devs to understand.

Also, I see that this section is ***too high*** in the article. It should be moved down just past the *Nested objects and collection types* section.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/forms/validation.md](https://github.com/dotnet/AspNetCore.Docs/blob/9ae7aa99ed69f7147357e920c653c73be08f24d3/aspnetcore/blazor/forms/validation.md) | [aspnetcore/blazor/forms/validation](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/forms/validation?branch=pr-en-us-36398) |

### DIFF
--- a/aspnetcore/blazor/forms/validation.md
+++ b/aspnetcore/blazor/forms/validation.md
@@ -1602,7 +1602,6 @@ In the following `Order` class, the `[ValidatableType]` attribute is required on
 
 ```csharp
 using System.ComponentModel.DataAnnotations;
-using Microsoft.Extensions.Validation;
 
 [ValidatableType]
 public class Order


aspnetcore/blazor/forms/validation.md
```diff
@@ -1643,6 +1601,9 @@ In the following `Order` class, the `[ValidatableType]` attribute is required on
 `Order.cs`:
 
 ```csharp
+using System.ComponentModel.DataAnnotations;
+using Microsoft.Extensions.Validation;
```
oroztocil: I would remove the using for `Microsoft.Extensions.Validation`. This will give people ambiguous reference between the experimental `Microsoft.Extensions.Validation.ValidatableType` and the generated `Microsoft.Extensions.Validation.Embedded.ValidatableType`. At least in the context of a Web SDK or Razor SDK project.

This `using` cannot be avoided in some cases (e.g., when you want to use the `SkipValidation` attribute) and then the users need to resolve the conflict by specifying the type explicitly or declaring an alias. It is one of the pain points present in .NET 10 because of the unfinished nature of the new validation API. However, it seems to me that it is better that we try to protect people from needing to deal with it until they really need to.

oroztocil:
The new documentation looks good - for the purpose of mitigating confusion while using the current API. I would much prefer if we did not have to write all of this (a feature in a good shape would require as little documentation as possible). However, from the feedback it is obvious that we need to address it somehow in the meantime, so thank you for your help.

guardrex:
I agree; but if it isn't explained, it will probably lead to confusion ... more questions ... more lost time 💰. Let's plan to drop all of this at .NET 11.

aspnetcore/blazor/forms/validation.md
guardrex:
```suggestion
```

PR #36426, opened 2025-12-03
Fix IOpenApiDocumentProvider injection sample (keyed DI)

gabr0236:
Updates the IOpenApiDocumentProvider example in
`aspnetcore/fundamentals/openapi/using-openapi-documents.md` to:
* Inject the provider as a keyed service using [FromKeyedServices("v1")].
* Call GetOpenApiDocumentAsync with an optional CancellationToken instead of a document name.
* Note that the service key "v1" must match the name passed to AddOpenApi.

The previous sample did not compile (because `GetOpenApiDocumentAsync` does not take a document name parameter) and did not reflect the keyed registration used by `AddOpenApi("v1")`.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/fundamentals/openapi/using-openapi-documents.md](https://github.com/dotnet/AspNetCore.Docs/blob/a91c7f1ba86ab88b382f321fc897dbd95a925417/aspnetcore/fundamentals/openapi/using-openapi-documents.md) | [aspnetcore/fundamentals/openapi/using-openapi-documents](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/openapi/using-openapi-documents?branch=pr-en-us-36426) |

### DIFF
--- a/aspnetcore/fundamentals/openapi/using-openapi-documents.md
+++ b/aspnetcore/fundamentals/openapi/using-openapi-documents.md
@@ -5,7 +5,7 @@ description: Learn how to use OpenAPI documents in an ASP.NET Core app.
 ms.author: safia
 monikerRange: '>= aspnetcore-6.0'
 ms.custom: mvc
-ms.date: 12/04/2025
+ms.date: 01/13/2026
 uid: fundamentals/openapi/using-openapi-documents
 ---
 # Use openAPI documents
@@ -65,8 +65,8 @@ Enable document generation at build time by setting the following properties in 
 
 ```xml
 <PropertyGroup>
-    <OpenApiDocumentsDirectory>$(MSBuildProjectDirectory)</OpenApiDocumentsDirectory>
-    <OpenApiGenerateDocuments>true</OpenApiGenerateDocuments>
+  <OpenApiDocumentsDirectory>$(MSBuildProjectDirectory)</OpenApiDocumentsDirectory>
+  <OpenApiGenerateDocuments>true</OpenApiGenerateDocuments>
 </PropertyGroup>
 ```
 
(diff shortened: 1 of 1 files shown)


gabr0236:
@captainsafia FYI

guardrex:
Hi @gabr0236! Sorry for the delay. I don't think Safia saw the ping. I sent her an email. Give her a little time to respond due to the holidays. Most of us are either totally OOF or working limited days until the first week of January. *Happy New Year!* 🥳 

BTW ... Please resolve the conflict, so she can see the changes that you're requesting.

gabr0236:
Hi @guardrex, no worries. I resolved the conflicts. Happy new years!

guardrex:
Thanks! 

BTW ... You and Safia will work out the final details; but in passing, it seems like ...

* A primary ctor seems in order here.
* The method should remain async.
* The method should be an expression-bodied method.
* There's no need to underscore the parameter name.

For comparison and consideration (and I'm breaking the lines to adhere to our 85 character line length limit) ...

```csharp
public class CustomDocumentService(
    [FromKeyedServices("v1")] IOpenApiDocumentProvider documentProvider)
{
    public async Task<OpenApiDocument> GetApiDocumentAsync(
            CancellationToken cancellationToken = default) =>
        await documentProvider.GetOpenApiDocumentAsync(cancellationToken);
}
```

aspnetcore/fundamentals/openapi/using-openapi-documents.md
```diff
@@ -109,21 +109,22 @@ public class CustomDocumentService
...
+        return _documentProvider.GetOpenApiDocumentAsync(cancellationToken);
     }
 }
 ```
 
-Register the service in your DI container:
+Register the service in your DI container. Note that service key "v1" should match the document name passed to `AddOpenApi`:
 
 ```csharp
+builder.Services.AddOpenApi("v1");
```
captainsafia: `"v1"` is the default so we don't need to explicitly include it here. If we want to make the sample more explicit, we might want to add multiple `AddOpenApi` calls then showcase how `FromKeyedServices` can be used to target a specific document.

```
builder.AddOpenApi();
builder.AddOpenApi("v2");
```

aspnetcore/fundamentals/openapi/using-openapi-documents.md
```diff
@@ -109,21 +109,22 @@ public class CustomDocumentService
...
     private readonly IOpenApiDocumentProvider _documentProvider;
 
-    public CustomDocumentService(IOpenApiDocumentProvider documentProvider)
+    public CustomDocumentService([FromKeyedServices("v1")] IOpenApiDocumentProvider documentProvider)
     {
         _documentProvider = documentProvider;
     }
 
-    public async Task<OpenApiDocument> GetApiDocumentAsync()
+    public Task<OpenApiDocument> GetApiDocumentAsync(CancellationToken cancellationToken = default)
```
captainsafia: Removing the async here is probably fine since we were directly returning the result of the call and not doing anything else.

I wonder though if its a code smell and we should make the example more illustrative, perhaps by doing some operation on the returned document before returning it?

I think this diff is fine as is though.

guardrex:
@gabr0236 ... Please sign the CLA if you agree to its terms 👇. We can't accept the PR without it.

@captainsafia ... Please see the diff now. I made the changes that you suggested.

gabr0236:
@dotnet-policy-service agree

gabr0236:
@guardrex sorry for the delay - thank you for resolving the comments!

PR #36588, opened 2026-01-07
Opaque (reference) access token guidance

### LINKED ISSUE #36422: OIDC Solution Doesn't Work With Opaque Access Tokens
### Description

As far as I can tell, the .NET OIDC implementation for a separate Web project and backend API relies on the access token being a JWT. This is not something required by OIDC and thus this solution does not work with our SSO provider.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/security/blazor-web-app-with-oidc?view=aspnetcore-10.0&pivots=non-bff-pattern#sample-solution

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/security/blazor-web-app-with-oidc.md

### Document ID

c3346d3a-346b-8db4-2650-ec044b3f0dd9

### Platform Id

1dd51a4e-f734-569f-f6e9-19a27bcd9f4d

### Article author

@guardrex

### Metadata

* ID: c3346d3a-346b-8db4-2650-ec044b3f0dd9
* PlatformId: 1dd51a4e-f734-569f-f6e9-19a27bcd9f4d 
* Service: **aspnet-core**
* Sub-service: **blazor**

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+c3346d3a-346b-8db4-2650-ec044b3f0dd9)

guardrex:
Fixes #36422

cc: @mikekistler 

Stephen ... I hacked some nasty 🦖 code 🙈😆 with the help of AI to give you an idea of what I have in mind for the bits that call the auth server to validate the token.

Apparently, Entra doesn't support opaque access token validation, per [this MS answer](https://learn.microsoft.com/en-us/answers/questions/1462452/azure-ad-opaque-token-validation) as of 2023 and a local test that I just ran here with Entra.

I originally had this in the BWA-OIDC article, but it's more general than that, so I just moved it to the [additional scenarios article](https://learn.microsoft.com/aspnet/core/blazor/security/additional-scenarios) and cross-linked to it there from a few spots.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/security/additional-scenarios.md](https://github.com/dotnet/AspNetCore.Docs/blob/f6f6640f137396659baadfbc1798268510592870/aspnetcore/blazor/security/additional-scenarios.md) | [aspnetcore/blazor/security/additional-scenarios](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/additional-scenarios?branch=pr-en-us-36588) |
| [aspnetcore/blazor/security/blazor-web-app-with-entra.md](https://github.com/dotnet/AspNetCore.Docs/blob/f6f6640f137396659baadfbc1798268510592870/aspnetcore/blazor/security/blazor-web-app-with-entra.md) | [aspnetcore/blazor/security/blazor-web-app-with-entra](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/blazor-web-app-with-entra?branch=pr-en-us-36588) |
| [aspnetcore/blazor/security/blazor-web-app-with-oidc.md](https://github.com/dotnet/AspNetCore.Docs/blob/f6f6640f137396659baadfbc1798268510592870/aspnetcore/blazor/security/blazor-web-app-with-oidc.md) | [aspnetcore/blazor/security/blazor-web-app-with-oidc](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/blazor-web-app-with-oidc?branch=pr-en-us-36588) |
| [aspnetcore/blazor/security/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/f6f6640f137396659baadfbc1798268510592870/aspnetcore/blazor/security/index.md) | [aspnetcore/blazor/security/index](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/index?branch=pr-en-us-36588) |


guardrex:
@halter73 ... Just checking in on your idea to flesh out the code bits here more completely. If that would take a while to do, can we get started by merging this now with a follow-up later? I'd like to clear some of my PRs that have been sitting for a while.

guardrex:
@halter73 ... ~Checking in on your idea to flesh out the code bits here more completely ... you need more time, correct?~ I had AI help me hack up some code for the missing piece. I'll see if I can get it to run with Entra.

aspnetcore/blazor/security/blazor-web-app-with-oidc.md
```diff
@@ -1462,6 +1463,220 @@ At this point, Razor components can adopt [role-based and policy-based authoriza
...
+
+public class OpaqueTokenAuthenticationOptions : AuthenticationSchemeOptions
+{
+    public const string DefaultScheme = "OpaqueTokenAuthentication";
+    public string? IntrospectionEndpoint { get; set; }
+    public string? ClientId { get; set; }
+}
+```
+
+`OpaqueTokenAuthenticationHandler.cs`:
```
guardrex:
```suggestion
`Authentication/OpaqueTokenAuthenticationHandler.cs`:
```

aspnetcore/blazor/security/blazor-web-app-with-oidc.md
```diff
@@ -1462,6 +1463,220 @@ At this point, Razor components can adopt [role-based and policy-based authoriza
...
+using System.Net.Http.Headers;
+using System.Security.Claims;
+using System.Text.Encodings.Web;
+using System.Text.Json;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.Extensions.Options;
+using MinimalApiJwt.Authentication;
+using MinimalApiJwt.Extensions;
+
+namespace MinimalApiJwt.Services;
```
guardrex:
```suggestion
namespace MinimalApiJwt.Authentication;
```

aspnetcore/blazor/security/blazor-web-app-with-oidc.md
```diff
@@ -1462,6 +1463,220 @@ At this point, Razor components can adopt [role-based and policy-based authoriza
...
+`OpaqueTokenAuthenticationHandler.cs`:
+
+```csharp
+using System.Net.Http.Headers;
+using System.Security.Claims;
+using System.Text.Encodings.Web;
+using System.Text.Json;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.Extensions.Options;
+using MinimalApiJwt.Authentication;
```
guardrex:
```suggestion
```

aspnetcore/blazor/security/blazor-web-app-with-oidc.md
```diff
@@ -1462,6 +1463,220 @@ At this point, Razor components can adopt [role-based and policy-based authoriza
...
+            returned.
+
+            Values for the authentication server introspection URI
+            ('{AUTH SERVER INTROSPECTION URI}') and the API client ID
+            ('{API CLIENT ID}') can be supplied from app settings
+            or any other configuration source.
+        */
+
+        var introspectionUri = "{AUTH SERVER INTROSPECTION URI}";
+        var clientId = "{API CLIENT ID}";
```
guardrex:
```suggestion
        var introspectionUri = options.IntrospectionEndpoint;
        var clientId = options.ClientId;
```

aspnetcore/blazor/security/blazor-web-app-with-oidc.md
```diff
@@ -1462,6 +1463,220 @@ At this point, Razor components can adopt [role-based and policy-based authoriza
...
+        var clientSecret = config["Authentication:Schemes:AuthServer:ClientSecret"];
+
+        using var client = new HttpClient();
```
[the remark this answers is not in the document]
guardrex: Took this suggestion, but I used primary constructor injection for it.

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,212 @@ builder.Services.AddHttpClient("HttpMessageHandler")
...
+
+            If the token is valid, an AuthenticationTicket is created 
+            containing the user's claims.
+
+            If the token is invalid, a failed authorization result is 
+            returned.
+        */
+
+        var introspectionUri = options.IntrospectionEndpoint;
+        var clientId = options.ClientId;
```
halter73: This won't compile. `options` is the `IOptionsMonitor<OpaqueTokenAuthenticationOptions>` constructor parameter, which doesn't have `IntrospectionEndpoint`/`ClientId` properties. Use the base class's `Options` property (capital `O`) instead, which exposes the resolved options for the current scheme:

```suggestion
        var introspectionUri = Options.IntrospectionEndpoint;
        var clientId = Options.ClientId;
```

If you keep this fix, the `options` constructor parameter is no longer referenced directly and the `using Microsoft.Extensions.Options;` line can stay (still needed for the generic constraint passed to the base type).

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,212 @@ builder.Services.AddHttpClient("HttpMessageHandler")
...
+        {
+            options.IntrospectionEndpoint = "{AUTH SERVER INTROSPECTION URI}";
+            options.ClientId = "{API CLIENT ID}";
+        });
+```
+
+The preceding example's placeholders:
+
+* `{AUTH SERVER INTROSPECTION URI}`: Authentication server's introspection URI
+* `{API CLIENT ID}`: API Client ID
```
halter73: Nit. Sentence case per the style guide:

```suggestion
* `{API CLIENT ID}`: API client ID
```

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,212 @@ builder.Services.AddHttpClient("HttpMessageHandler")
...
+            If the token is valid, an AuthenticationTicket is created 
+            containing the user's claims.
+
+            If the token is invalid, a failed authorization result is 
+            returned.
+        */
+
+        var introspectionUri = options.IntrospectionEndpoint;
+        var clientId = options.ClientId;
+        var clientSecret = config["Authentication:Schemes:AuthServer:ClientSecret"];
```
halter73: The scheme name in this config key (`AuthServer`) doesn't match `OpaqueTokenAuthenticationOptions.DefaultScheme` (`"OpaqueTokenAuthentication"`). The standard `Authentication:Schemes:{SchemeName}:...` convention would make this:

```suggestion
        var clientSecret = config["Authentication:Schemes:OpaqueTokenAuthentication:ClientSecret"];
```

If you change this, also update the `dotnet user-secrets set` command on line 1095 and the explanatory text on line 1084 to use the same key.

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,212 @@ builder.Services.AddHttpClient("HttpMessageHandler")
...
+using Microsoft.AspNetCore.Authentication;
+
+namespace MinimalApiJwt.Authentication;
+
+public class OpaqueTokenAuthenticationOptions : AuthenticationSchemeOptions
+{
+    public const string DefaultScheme = "OpaqueTokenAuthentication";
+    public string? IntrospectionEndpoint { get; set; }
+    public string? ClientId { get; set; }
+}
```
halter73, line 1150: Related to the `Authentication:Schemes:AuthServer:ClientSecret` comment further down: the cleaner pattern is to add `ClientSecret` to `OpaqueTokenAuthenticationOptions` and let the framework's normal `Authentication:Schemes:{SchemeName}:*` binding pick it up. Then the handler doesn't need an `IConfiguration` dependency at all and the config key in the docs and the `dotnet user-secrets set` command would automatically match the scheme name. Optional, but worth considering since it eliminates the naming-inconsistency surface area entirely.

```suggestion
public class OpaqueTokenAuthenticationOptions : AuthenticationSchemeOptions
{
    public const string DefaultScheme = "OpaqueTokenAuthentication";
    public string? IntrospectionEndpoint { get; set; }
    public string? ClientId { get; set; }
    public string? ClientSecret { get; set; }
}
```

If you apply this, also:
- Drop the `IConfiguration config` parameter from the handler's constructor.
- Replace `config["Authentication:Schemes:AuthServer:ClientSecret"]` with `Options.ClientSecret`.
- Update the `dotnet user-secrets set` command to use `Authentication:Schemes:OpaqueTokenAuthentication:ClientSecret`.

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,212 @@ builder.Services.AddHttpClient("HttpMessageHandler")
...
+        using var doc = JsonDocument.Parse(responseString);
+
+        // The 'active' property determines if the token is valid and not expired
+        var tokenIsValid = doc.RootElement.GetProperty("active").GetBoolean();
+
+        if (tokenIsValid)
+        {
+            // TODO: Replace the '{USER ID}' placeholder with extracted claim value
+            // from the token introspection response
+            var claims = new[] { new Claim(ClaimTypes.Name, "{USER ID}") };
```
halter73: **This is the most important issue with the section.** I stood up Keycloak and ran the sample end-to-end with a real access token. With the handler as written, the call succeeds (`IsAuthenticated == true`) but produces a `ClaimsPrincipal` with `Name = "{USER ID}"`, no `NameIdentifier`, no email, and no roles. The `// TODO` is buried inside a code block and very easy to miss — a reader who copies this verbatim ships authentication that *looks* like it works but exposes no usable user information.

Please replace the placeholder with a real mapping from the standard RFC 7662 introspection fields. Suggested:

```suggestion
            // Map standard introspection response fields onto claims.
            // Field names below match what Keycloak, Duende IdentityServer,
            // Auth0, and Okta return; adjust the role source for your provider.
            var claims = new List<Claim>();

            string? Get(string name) =>
                doc.RootElement.TryGetProperty(name, out var v) &&
                v.ValueKind == JsonValueKind.String ? v.GetString() : null;

            var sub = Get("sub");
            var username = Get("preferred_username") ?? Get("username") ?? sub;

            if (sub is not null) claims.Add(new Claim(ClaimTypes.NameIdentifier, sub));
            if (username is not null) claims.Add(new Claim(ClaimTypes.Name, username));
            if (Get("email") is { } email) claims.Add(new Claim(ClaimTypes.Email, email));
            if ((Get("client_id") ?? Get("azp")) is { } cid)
                claims.Add(new Claim("client_id", cid));
            if (Get("scope") is { } scope)
                foreach (var s in scope.Split(' ', StringSplitOptions.RemoveEmptyEntries))
                    claims.Add(new Claim("scope", s));

            // Keycloak surfaces realm roles under realm_access.roles.
            // Duende/IdentityServer uses a flat "role" claim; Auth0 uses a
            // configurable custom claim. Adjust for your authorization server.
            if (doc.RootElement.TryGetProperty("realm_access", out var ra) &&
                ra.ValueKind == JsonValueKind.Object &&
                ra.TryGetProperty("roles", out var roles) &&
                roles.ValueKind == JsonValueKind.Array)
            {
                foreach (var r in roles.EnumerateArray())
                    if (r.ValueKind == JsonValueKind.String)
                        claims.Add(new Claim(ClaimTypes.Role, r.GetString()!));
            }

            var identity = new ClaimsIdentity(claims,
                OpaqueTokenAuthenticationOptions.DefaultScheme,
                nameType: ClaimTypes.Name,
                roleType: ClaimTypes.Role);
```

Side-by-side test result against the same Keycloak token (user `alice` in roles `reader`, `writer`):

| | Doc handler (verbatim) | With the mapping above |
|---|---|---|
| `User.Identity.Name` | `"{USER ID}"` | `"alice"` |
| `User.IsInRole("reader")` | `false` | `true` |
| Claim count | 1 | 9 |

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,212 @@ builder.Services.AddHttpClient("HttpMessageHandler")
 ```
 
 :::moniker-end
+
+## Opaque (reference) access token support
+
+*The following guidance requires an authentication server that supports opaque (reference) access tokens. Currently, Microsoft Entra doesn't support opaque access token validation.*
```
halter73: Based on testing the sample against Keycloak:

1. Keycloak and Okta also issue **JWT** access tokens by default — there isn't a first-class "opaque" toggle. The handler still works against them (introspection doesn't care whether the token is opaque or JWT), but readers reaching for this section because they're using Keycloak/Okta should know that "opaque" here refers to how the *client* treats the token, not how the *server* mints it. Duende IdentityServer is the main implementation in the .NET ecosystem that issues truly opaque reference tokens out of the box.

2. **Keycloak-specific footgun, worth a sentence:** when testing introspection against Keycloak, the API's introspection client must be different from the OIDC client that issued the user's access token. If they're the same client, Keycloak returns `{"active": false}` with `"Access token JWT check failed"` in the server log, and the reader will think the handler is broken. In the documented Blazor Web App scenario this doesn't happen naturally (`BlazorWebAppOidc` and `MinimalApiJwt` are separate clients), but a one-liner saves the next person an hour.

Also, italicized prose for a caveat reads a bit awkwardly here. Consider a NOTE block so it stands out the same way the IMPORTANT block below does::

```suggestion
> [!NOTE]
> The following guidance requires an authentication server that supports opaque (reference) access tokens. Currently, Microsoft Entra doesn't support opaque access token validation. Keycloak and Okta issue JWT access tokens by default; the handler still works against them because it relies only on RFC 7662 introspection, but "opaque" in this section describes how the client treats the token rather than how the server mints it. Duende IdentityServer issues true opaque reference tokens out of the box.
>
> When testing this pattern against Keycloak specifically, the API's introspection client must be a different OIDC client than the one that issued the user's access token. Introspecting a token using the client that minted it returns `{"active": false}`.
```

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,253 @@ builder.Services.AddHttpClient("HttpMessageHandler")
...
+
+The following <xref:Microsoft.AspNetCore.Authentication.AuthenticationHandler%601> and associated configuration and helper code is provided as a general approach, which might require further development to suit a specific authorization server's requirements. The following handler extracts the opaque token from the `Authorization` header for an HTTP call to an authorization server's introspection endpoint and creates an <xref:Microsoft.AspNetCore.Authentication.AuthenticationTicket> containing the user's claims.
+
+Calling an authorization server's introspection endpoint requires authentication. The following example relies on setting the client secret for authentication in the request's Authorization header (base64 encoded credentials) using the [Secret Manager tool](xref:security/app-secrets) for local development and testing.
+
+[!INCLUDE[](~/blazor/security/includes/secure-authentication-flows.md)]
+
+In the following handler, the authorization server's introspection endpoint client secret uses the configuration key `Authentication:Schemes:OpaqueTokenAuthentication:ClientSecret`.
+
```
damienbod, line 1094: Maybe add a note here: client assertions are recommended in production (Instead of client "shared" secrets)

damienbod:
Hi @guardrex nice work. If you are making docs about introspection, I think you should mention something about the revocation endpoint and that when using reference tokens, these are typically invalidated on a logout event using the revocation endpoint.

aspnetcore/blazor/security/additional-scenarios.md
guardrex: With what details/cross-link do you recommend? Is the MSAL doc a reasonable?

https://learn.microsoft.com/entra/msal/dotnet/acquiring-tokens/web-apps-apis/confidential-client-assertions

I'm considering the following for this paragraph ...

> In the following handler, the authorization server's introspection endpoint client secret uses the configuration key \`Authentication:Schemes:OpaqueTokenAuthentication:ClientSecret\`. For production apps, consider using \*client assertions\*. For more information, see \[Confidential client assertions (Microsoft Entra documentation)](/entra/msal/dotnet/acquiring-tokens/web-apps-apis/confidential-client-assertions).

guardrex:
Thanks, @damienbod ... I credit AI and Stephen to a great degree here on this one.

> I think you should mention something about the revocation endpoint and that when using reference tokens, these are typically invalidated on a logout event using the revocation endpoint.

~Can the article show a general implementation as a starting point? Do you have a snippet that you can provide?~

**UPDATE**: From AI with a little 🦖 hacking, I have some potential "starting point" code that I added on the last commit ...

```csharp
app.MapPost("/logout", 
    async ([FromForm] string? returnUrl, HttpContext context, 
    IHttpClientFactory httpClientFactory) =>
{
    var accessToken = await context.GetTokenAsync("access_token");

    if (!string.IsNullOrEmpty(accessToken))
    {
        // Prepare the revocation request (RFC 7009)
        var requestContent = 
            new FormUrlEncodedContent(new Dictionary<string, string>
            {
                { "token", accessToken },
                { "token_type_hint", "access_token" },
                { "client_id", "your_client_id" },
                { "client_secret", "your_client_secret" }
            });

        // POST to the revocation endpoint
        using var client = httpClientFactory.CreateClient();

        await client.PostAsync(
            "https://your-identity-server/connect/revocation", requestContent);
    }

    TypedResults.SignOut(GetAuthProperties(returnUrl), 
        [CookieAuthenticationDefaults.AuthenticationScheme]);
});
```

If that's fairly close, we could simply describe it as "a starting point for further development."

aspnetcore/blazor/security/additional-scenarios.md
guardrex: I added the remark on the last commit.

guardrex:
@halter73 ... There was considerable churn (nine commits) while updating this. It's best if you look at the latest version of the PR to determine if the updates are sane. Unfortunately, I can't squash the nine commits into one because the Copilot commit was a merge commit.

guardrex:
@damienbod ... Would you review the updates? Stephen might be unavailable right now, and this PR is very close to being ready to merge. If you'll review the last couple of updates that I made per your suggestions, I'll ask Wade to look things over after that, and these updates can be published.

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,304 @@ builder.Services.AddHttpClient("HttpMessageHandler")
...
+A failure occurs only when the opaque token acquired by <xref:Microsoft.Extensions.DependencyInjection.OpenIdConnectExtensions.AddOpenIdConnect%2A> is passed to another service that attempts to validate it with <xref:Microsoft.Extensions.DependencyInjection.JwtBearerExtensions.AddJwtBearer%2A>. Unlike self-contained JWTs, opaque tokens require a request to an authorization server to validate their status and retrieve claims. To work around this limitation, either use a third-party API, such as the [Duende Introspection Authentication Handler](https://docs.duendesoftware.com/introspection/), or create a [custom `AuthenticationHandler`](xref:security/authentication/index#authentication-handler) to validate the token.
```
damienbod: Unlike self-contained JWTs, opaque tokens require a request to an authorization server to validate their status and retrieve claims. 

=> Unlike self-contained JWTs, opaque tokens require a request to an authorization server to validate the status and to retrieve the claims.

guardrex:
```suggestion
A failure occurs only when the opaque token acquired by <xref:Microsoft.Extensions.DependencyInjection.OpenIdConnectExtensions.AddOpenIdConnect%2A> is passed to another service that attempts to validate it with <xref:Microsoft.Extensions.DependencyInjection.JwtBearerExtensions.AddJwtBearer%2A>. Unlike self-contained JWTs, opaque tokens require a request to an authorization server to validate the status and to retrieve the claims. To work around this limitation, either use a third-party API, such as the [Duende Introspection Authentication Handler](https://docs.duendesoftware.com/introspection/), or create a [custom `AuthenticationHandler`](xref:security/authentication/index#authentication-handler) to validate the token.
```

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,304 @@ builder.Services.AddHttpClient("HttpMessageHandler")
...
+        else
+        {
+            var failedResult = AuthenticateResult.Fail("Bearer token invalid.");
+
+            return failedResult;
+        }
+    }
+}
+```
+
```
damienbod: Maybe add a note, this code can be further improved by using the OpenID Connect discovery endpoint and adding a cache for the client HttpClient Introspection requests.

damienbod:
LGTM, just adds 2 comments

aspnetcore/blazor/security/additional-scenarios.md
guardrex:
```suggestion

> [!NOTE]
> The preceding approach can be further improved by using the OpenID Connect discovery endpoint and adding a cache for the client's <xref:System.Net.Http.HttpClient> introspection requests.
```

guardrex:
Thanks @damienbod! ... The NOTE is obviously going to be for advanced devs. I have nothing in our doc set that I can cross-link for that approach.

guardrex:
@wadepickett ... @halter73 has gone MIA again! 😆 *He's probably buried in code somewhere!* I made all of his suggested updates, and @damienbod reviewed (thx again, Damien! 🙏). I think we're in good shape thus far. If we get your suggestions addressed, we can merge this. No rush tho ... it doesn't have to be today ... or even this week for that matter.

halter73 (commented):
Only stage, and if you think it's good to go, feel free to leave no comments and tell me

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,306 @@ builder.Services.AddHttpClient("HttpMessageHandler")
...
+    : AuthenticationHandler<OpaqueTokenAuthenticationOptions>(optionsMonitor, 
+        logger, encoder)
+{
+    protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
+    {
+        var opaqueToken = Request.ExtractBearerToken();
+
+        if (opaqueToken is null)
+        {
+            var failedResult = AuthenticateResult.Fail(
```
halter73, line 1191: This won't compile. `options` is the `IOptionsMonitor<OpaqueTokenAuthenticationOptions>` constructor parameter, which doesn't have `IntrospectionEndpoint`/`ClientId` properties. Use the base class's `Options` property (capital `O`) instead, which exposes the resolved options for the current scheme:

```suggestion
        var introspectionUri = Options.IntrospectionEndpoint;
        var clientId = Options.ClientId;
```

If you keep this fix, the `options` constructor parameter is no longer referenced directly and the `using Microsoft.Extensions.Options;` line can stay (still needed for the generic constraint passed to the base type).

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,306 @@ builder.Services.AddHttpClient("HttpMessageHandler")
 ```
 
 :::moniker-end
+
+## Opaque (reference) access token support
+
+The guidance in this section explains how to implement opaque (reference) access token support, which offers the following advantages over JSON Web Tokens (JWTs):
```
halter73, line 1069: Italicized prose for a caveat reads a bit awkwardly here — consider a NOTE block so it stands out the same way the IMPORTANT block below does:

```suggestion
> [!NOTE]
> The following guidance requires an authentication server that supports opaque (reference) access tokens. Currently, Microsoft Entra doesn't support opaque access token validation.
```

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,306 @@ builder.Services.AddHttpClient("HttpMessageHandler")
...
+                claims.Add(new Claim("client_id", cid));
+            if (Get("scope") is { } scope)
+                foreach (var s in scope.Split(' ', StringSplitOptions.RemoveEmptyEntries))
+                    claims.Add(new Claim("scope", s));
+
+            // Keycloak surfaces realm roles under realm_access.roles.
+            // Duende/IdentityServer uses a flat "role" claim; Auth0 uses a
+            // configurable custom claim. Adjust for your authorization server.
+            if (doc.RootElement.TryGetProperty("realm_access", out var ra) &&
+                ra.ValueKind == JsonValueKind.Object &&
```
halter73, line 1270: Sentence case per the style guide:

```suggestion
* `{API CLIENT ID}`: API client ID
```

aspnetcore/blazor/security/blazor-web-app-with-oidc.md
```diff
@@ -126,7 +126,7 @@ For more information on using Aspire and details on the `.AppHost` and `.Service
 
 Confirm that you've met the prerequisites for Aspire. For more information, see the *Prerequisites* section of [Quickstart: Build your first Aspire solution](/dotnet/aspire/get-started/build-your-first-aspire-app?tabs=visual-studio#prerequisites).
 
-The sample app only configures an insecure HTTP launch profile (`http`) for use during development testing. For more information, including an example of insecure and secure launch settings profiles, see [Allow unsecure transport in Aspire (Aspire documentation)](/dotnet/aspire/troubleshooting/allow-unsecure-transport).
+The sample app only configures an insecure HTTP launch profile (`http`) for use during development testing. For more information on Aspire tooling security, see [Aspire dashboard security considerations: Aspire tooling (Aspire documentation)](https://aspire.dev/dashboard/security-considerations/#aspire-tooling).
```
halter73: Same broken fragment as in `blazor-web-app-with-entra.md` — `#aspire-tooling` doesn't exist on the target page.

```suggestion
The sample app only configures an insecure HTTP launch profile (`http`) for use during development testing. For more information on Aspire tooling security, see [Aspire dashboard security considerations (Aspire documentation)](https://aspire.dev/dashboard/security-considerations/).
```

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,306 @@ builder.Services.AddHttpClient("HttpMessageHandler")
...
+        logger, encoder)
+{
+    protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
+    {
+        var opaqueToken = Request.ExtractBearerToken();
+
+        if (opaqueToken is null)
+        {
+            var failedResult = AuthenticateResult.Fail(
+                "Bearer token not found in Authorization header.");
```
halter73, line 1192: The scheme name in this config key (`AuthServer`) doesn't match `OpaqueTokenAuthenticationOptions.DefaultScheme` (`"OpaqueTokenAuthentication"`). The standard `Authentication:Schemes:{SchemeName}:...` convention would make this:

```suggestion
        var clientSecret = config["Authentication:Schemes:OpaqueTokenAuthentication:ClientSecret"];
```

If you change this, also update the `dotnet user-secrets set` command on line 1095 and the explanatory text on line 1084 to use the same key.

aspnetcore/blazor/security/blazor-web-app-with-entra.md
```diff
@@ -85,7 +85,7 @@ For more information on using Aspire and details on the `.AppHost` and `.Service
 
 Confirm that you've met the prerequisites for Aspire. For more information, see the *Prerequisites* section of [Quickstart: Build your first Aspire solution](/dotnet/aspire/get-started/build-your-first-aspire-app?tabs=visual-studio#prerequisites).
 
-The sample app only configures an insecure HTTP launch profile (`http`) for use during development testing. For more information, including an example of insecure and secure launch settings profiles, see [Allow unsecure transport in Aspire (Aspire documentation)](/dotnet/aspire/troubleshooting/allow-unsecure-transport).
+The sample app only configures an insecure HTTP launch profile (`http`) for use during development testing. For more information on Aspire tooling security, see [Aspire dashboard security considerations: Aspire tooling (Aspire documentation)](https://aspire.dev/dashboard/security-considerations/#aspire-tooling).
```
halter73: The `#aspire-tooling` fragment doesn't exist on that page — the actual section headings are `#scenarios-for-running-the-dashboard`, `#secure-telemetry-endpoint`, etc. Either drop the fragment or point at a real anchor:

```suggestion
The sample app only configures an insecure HTTP launch profile (`http`) for use during development testing. For more information on Aspire tooling security, see [Aspire dashboard security considerations (Aspire documentation)](https://aspire.dev/dashboard/security-considerations/).
```

Also — is bundling this Aspire-link change in with the opaque-token PR intentional? It's unrelated to the rest of the changes and might be worth splitting out.

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,306 @@ builder.Services.AddHttpClient("HttpMessageHandler")
...
+
+        return null;
+    }
+}
+```
+
+`Authentication/OpaqueTokenAuthenticationOptions.cs`:
+
+```csharp
+using Microsoft.AspNetCore.Authentication;
```
halter73, line 1140: Related to the `Authentication:Schemes:AuthServer:ClientSecret` comment further down: the cleaner pattern is to add `ClientSecret` to `OpaqueTokenAuthenticationOptions` and let the framework's normal `Authentication:Schemes:{SchemeName}:*` binding pick it up — then the handler doesn't need an `IConfiguration` dependency at all and the config key in the docs and the `dotnet user-secrets set` command would automatically match the scheme name. Optional, but worth considering since it eliminates the naming-inconsistency surface area entirely.

```suggestion
public class OpaqueTokenAuthenticationOptions : AuthenticationSchemeOptions
{
    public const string DefaultScheme = "OpaqueTokenAuthentication";
    public string? IntrospectionEndpoint { get; set; }
    public string? ClientId { get; set; }
    public string? ClientSecret { get; set; }
}
```

If you apply this, also:
- Drop the `IConfiguration config` parameter from the handler's constructor.
- Replace `config["Authentication:Schemes:AuthServer:ClientSecret"]` with `Options.ClientSecret`.
- Update the `dotnet user-secrets set` command to use `Authentication:Schemes:OpaqueTokenAuthentication:ClientSecret`.

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,306 @@ builder.Services.AddHttpClient("HttpMessageHandler")
...
+        // Post to the introspection endpoint
+        var response = await client.PostAsync(introspectionUri, content);
+
+        if (!response.IsSuccessStatusCode)
+        {
+            var failedResult = AuthenticateResult.Fail(
+                "Introspection endpoint failure.");
+
+            return failedResult;
+        }
```
halter73, line 1232: **This is the most important issue with the section.** I stood up Keycloak and ran the sample end-to-end with a real access token. With the handler as written, the call succeeds (`IsAuthenticated == true`) but produces a `ClaimsPrincipal` with `Name = "{USER ID}"`, no `NameIdentifier`, no email, and no roles. The `// TODO` is buried inside a code block and very easy to miss — a reader who copies this verbatim ships authentication that *looks* like it works but exposes no usable user information.

Please replace the placeholder with a real mapping from the standard RFC 7662 introspection fields. Suggested:

```suggestion
            // Map standard introspection response fields onto claims.
            // Field names below match what Keycloak, Duende IdentityServer,
            // Auth0, and Okta return; adjust the role source for your provider.
            var claims = new List<Claim>();

            string? Get(string name) =>
                doc.RootElement.TryGetProperty(name, out var v) &&
                v.ValueKind == JsonValueKind.String ? v.GetString() : null;

            var sub = Get("sub");
            var username = Get("preferred_username") ?? Get("username") ?? sub;

            if (sub is not null) claims.Add(new Claim(ClaimTypes.NameIdentifier, sub));
            if (username is not null) claims.Add(new Claim(ClaimTypes.Name, username));
            if (Get("email") is { } email) claims.Add(new Claim(ClaimTypes.Email, email));
            if ((Get("client_id") ?? Get("azp")) is { } cid)
                claims.Add(new Claim("client_id", cid));
            if (Get("scope") is { } scope)
                foreach (var s in scope.Split(' ', StringSplitOptions.RemoveEmptyEntries))
                    claims.Add(new Claim("scope", s));

            // Keycloak surfaces realm roles under realm_access.roles.
            // Duende/IdentityServer uses a flat "role" claim; Auth0 uses a
            // configurable custom claim. Adjust for your authorization server.
            if (doc.RootElement.TryGetProperty("realm_access", out var ra) &&
                ra.ValueKind == JsonValueKind.Object &&
                ra.TryGetProperty("roles", out var roles) &&
                roles.ValueKind == JsonValueKind.Array)
            {
                foreach (var r in roles.EnumerateArray())
                    if (r.ValueKind == JsonValueKind.String)
                        claims.Add(new Claim(ClaimTypes.Role, r.GetString()!));
            }

            var identity = new ClaimsIdentity(claims,
                OpaqueTokenAuthenticationOptions.DefaultScheme,
                nameType: ClaimTypes.Name,
                roleType: ClaimTypes.Role);
```

Side-by-side test result against the same Keycloak token (user `alice` in roles `reader`, `writer`):

| | Doc handler (verbatim) | With the mapping above |
|---|---|---|
| `User.Identity.Name` | `"{USER ID}"` | `"alice"` |
| `User.IsInRole("reader")` | `false` | `true` |
| Claim count | 1 | 9 |

aspnetcore/blazor/security/additional-scenarios.md
halter73, line 1069: A couple of additions worth folding into this caveat, based on testing the sample against Keycloak:

1. Keycloak and Okta also issue **JWT** access tokens by default — there isn't a first-class "opaque" toggle. The handler still works against them (introspection doesn't care whether the token is opaque or JWT), but readers reaching for this section because they're using Keycloak/Okta should know that "opaque" here refers to how the *client* treats the token, not how the *server* mints it. Duende IdentityServer is the main implementation in the .NET ecosystem that issues truly opaque reference tokens out of the box.

2. **Keycloak-specific footgun, worth a sentence:** when testing introspection against Keycloak, the API's introspection client must be different from the OIDC client that issued the user's access token. If they're the same client, Keycloak returns `{"active": false}` with `"Access token JWT check failed"` in the server log, and the reader will think the handler is broken. In the documented Blazor Web App scenario this doesn't happen naturally (`BlazorWebAppOidc` and `MinimalApiJwt` are separate clients), but a one-liner saves the next person an hour.

Suggested rewrite:

```suggestion
> [!NOTE]
> The following guidance requires an authentication server that supports opaque (reference) access tokens. Currently, Microsoft Entra doesn't support opaque access token validation. Keycloak and Okta issue JWT access tokens by default; the handler still works against them because it relies only on RFC 7662 introspection, but "opaque" in this section describes how the client treats the token rather than how the server mints it. Duende IdentityServer issues true opaque reference tokens out of the box.
>
> When testing this pattern against Keycloak specifically, the API's introspection client must be a different OIDC client than the one that issued the user's access token. Introspecting a token using the client that minted it returns `{"active": false}`.
```

aspnetcore/blazor/security/blazor-web-app-with-entra.md
halter73: Tweaking my earlier comment — you were right that `#aspire-tooling` does resolve (Starlight emits a sibling `<a class="sl-anchor-link">`, not an `id=` on the heading, so my "broken anchor" claim was wrong). But agreed the cleaner move is to drop the Aspire reference entirely here — the sentence reads fine on its own and the link isn't really about opaque tokens.

```suggestion
The sample app only configures an insecure HTTP launch profile (`http`) for use during development testing.
```

(Please delete my earlier comment on this line — I can't remove it from here.)

aspnetcore/blazor/security/blazor-web-app-with-oidc.md
halter73: Same as in the Entra article — drop the Aspire link entirely rather than fix the fragment. My earlier "broken anchor" claim was wrong (Starlight's `sl-anchor-link` does work), but the link isn't pulling its weight in an opaque-token PR.

```suggestion
The sample app only configures an insecure HTTP launch profile (`http`) for use during development testing.
```

(Please delete my earlier comment on this line as well.)

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,306 @@ builder.Services.AddHttpClient("HttpMessageHandler")
...
+
+        // POST to the revocation endpoint
+        using var client = httpClientFactory.CreateClient();
+
+        await client.PostAsync("{AUTH SERVER TOKEN REVOCATION URI}", content);
+    }
+
+    TypedResults.SignOut(GetAuthProperties(returnUrl), 
+        [CookieAuthenticationDefaults.AuthenticationScheme]);
+});
```
halter73, line 1360: Two issues in this snippet:

1. `TypedResults.SignOut(...)` is constructed and discarded — there's no `return`, so the framework never executes it and the user is never actually signed out. The lambda needs to return the result (and stop being `async void`-shaped).
2. `GetAuthProperties` isn't defined or referenced anywhere else in this article. It comes from the BlazorWebAppOidc sample, so a reader landing here from search has no idea where it's from. A one-line callout (or inlining `new AuthenticationProperties { RedirectUri = ... }`) would save them a hunt.

Suggested minimal fix for #1:

```suggestion
app.MapPost("/logout", 
    async ([FromForm] string? returnUrl, HttpContext context, 
    IHttpClientFactory httpClientFactory) =>
{
    var accessToken = await context.GetTokenAsync("access_token");

    if (!string.IsNullOrEmpty(accessToken))
    {
        // Prepare the revocation request (RFC 7009)
        var content = 
            new FormUrlEncodedContent(new Dictionary<string, string>
            {
                { "token", accessToken },
                { "token_type_hint", "access_token" },
                { "client_id", "{API CLIENT ID}" },
                { "client_secret", "{CLIENT SECRET}" }
            });

        // POST to the revocation endpoint
        using var client = httpClientFactory.CreateClient();

        await client.PostAsync("{AUTH SERVER TOKEN REVOCATION URI}", content);
    }

    return TypedResults.SignOut(GetAuthProperties(returnUrl), 
        [CookieAuthenticationDefaults.AuthenticationScheme]);
});
```

aspnetcore/blazor/security/additional-scenarios.md
```diff
@@ -1063,3 +1063,306 @@ builder.Services.AddHttpClient("HttpMessageHandler")
...
+
+builder.Services.AddHttpClient();
+builder.Services.AddAuthentication()
+    .AddScheme<OpaqueTokenAuthenticationOptions, OpaqueTokenAuthenticationHandler>(
+        OpaqueTokenAuthenticationOptions.DefaultScheme,
+        options =>
+        {
+            options.IntrospectionEndpoint = "{AUTH SERVER INTROSPECTION URI}";
+            options.ClientId = "{API CLIENT ID}";
+        });
```
halter73, line 1322: As written, this snippet never sets `Options.ClientSecret`, so the handler's new guard clause fails every request with `"Opaque token authentication isn't fully configured."` and returns 401. `AddScheme<TOptions, THandler>` doesn't auto-bind the `Authentication:Schemes:{name}:*` section the way `AddJwtBearer`/`AddOpenIdConnect` do — that auto-binding lives in `JwtBearerConfigureOptions`/`OpenIdConnectConfigureOptions`, not the base API. So readers who only follow the user-secrets step earlier in the article will hit a broken sample.

Verified against Keycloak: with the line below added, alice signs in with all expected claims and roles; without it, every call 401s.

```suggestion
        options =>
        {
            options.IntrospectionEndpoint = "{AUTH SERVER INTROSPECTION URI}";
            options.ClientId = "{API CLIENT ID}";
            options.ClientSecret = builder.Configuration["Authentication:Schemes:OpaqueTokenAuthentication:ClientSecret"];
        });
```

And add `{SECRET}` to the placeholder list below alongside the other two.

halter73 (approved):

aspnetcore/blazor/security/additional-scenarios.md
guardrex: I'm not sure what you're referring to. This isn't a caveat, and it's no longer italicized. We usually don't start sections with NOTE remarks. It currently begins ...

> The guidance in this section explains how to implement opaque (reference) access token support, which offers the following advantages over JSON Web Tokens (JWTs):
>
> * Strict revocation: Invalidate access tokens at any time before they naturally expire.
> * Token size limits: Store a large number of user claims in the token to avoid a prohibitively large JWT.
> * Security: Prevent API consumers or third parties from reading access token claims.

I felt that that would be a good introduction to the section. Is that a poor introduction?

guardrex: Again, I wonder if you're looking at the latest commits. This was all updated per your prior remarks.

aspnetcore/blazor/security/additional-scenarios.md
guardrex: Your suggestions were already addressed.

aspnetcore/blazor/security/additional-scenarios.md
guardrex: Already addressed by an earlier commit.

aspnetcore/blazor/security/additional-scenarios.md
guardrex: Already addressed by an earlier commit.

aspnetcore/blazor/security/additional-scenarios.md
guardrex: Already addressed by an earlier commit.

aspnetcore/blazor/security/additional-scenarios.md
guardrex: Already addressed by an earlier commit.

aspnetcore/blazor/security/additional-scenarios.md
guardrex: Yes, this one needs to be addressed. Got it on my local branch and will commit it shortly.

aspnetcore/blazor/security/additional-scenarios.md
guardrex: Got it ... I'll put on the next commit.

aspnetcore/blazor/security/blazor-web-app-with-entra.md
guardrex: IIRC, the build report threw the error on the link. That's why it was changed on a PR that didn't directly focus on this part of the docs. I'm taking your suggestion now, which will be on the next commit.

halter73:
I think this all looks good to merge. Sorry for all the repeated/stale review content! I should have noticed that before posting.

PR #36620, opened 2026-01-13
Update license exclusion in docfx.json

wadepickett:
Docfx.json: Add exclusion for LICENSE.md jquery files.

- Added upper case LICENSE.md and README.md to the doc build checks exclude list.  

Why: jquery includes the uppercase version now and it was creating errors in the build worflow processes since they were being treated as MS documenation files.

### DIFF
--- a/aspnetcore/docfx.json
+++ b/aspnetcore/docfx.json
@@ -3,7 +3,7 @@
     "content": [
       {
         "files": [ "**/*.md", "**/*.yml" ], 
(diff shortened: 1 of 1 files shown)


wadepickett:
Closing.  We decided the issue was the status-checker action triggured at build rather then the build processing itself.  As status-checker creates the preview links from all new or changed .md files, it includes added LICENSE.md files included with jquery.
The fix will likely be to have status-checker use the exclusion list from docfx.json to exclude .md files from the preview link generation.

PR #36630, opened 2026-01-15
Update Blazor release notes

### LINKED ISSUE #36628: Tweaks to .NET 10.0 release notes for Blazor
### Description

I have notified two minor things in the release notes at https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-10.0?view=aspnetcore-10.0#client-side-fingerprinting
It's missing the "preloading feature", see [[Pre4] Preloaded Blazor framework assets by guardrex · Pull Request #35376 · dotnet/AspNetCore.Docs](https://github.com/dotnet/AspNetCore.Docs/pull/35376/files)
We are adding StaticWebAssetFingerprintPattern with .mjs , but the script `<script src="js/scripts#[.{fingerprint}].js"></script>` is .js.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-10.0?view=aspnetcore-10.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/release-notes/aspnetcore-10.0.md

### Document ID

a7ad4b01-2333-32d8-b759-e044edfb9102

### Platform Id

2f8fb8ab-5309-0a77-987a-c867c6517d2b

### Article author

@wadepickett

### Metadata

* ID: a7ad4b01-2333-32d8-b759-e044edfb9102
* PlatformId: 2f8fb8ab-5309-0a77-987a-c867c6517d2b 
* Service: **aspnet-core**
* Sub-service: **release-notes**

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+a7ad4b01-2333-32d8-b759-e044edfb9102)

guardrex:
Fixes #36628

Marek ... I might be incorrect, but it seems like the example that you were referring to is correct by design because it states the module is using the `.js` file extension ...

> ... a module named `scripts.js` in the app's `wwwroot/js` folder is fingerprinted by adding `#[.{fingerprint}]` before the file extension (`.js`):

... and then it goes on to say that if the dev wants to fingerprint `.mjs` files, they use `StaticWebAssetFingerprintPattern`.

> We are adding StaticWebAssetFingerprintPattern with .mjs

Yes, but that's after the example of a `.js` module ***IF*** they want to use a different extension (or fingerprint expression). Perhaps, I can fix that lead-in line and make it more specific like this ...

> To fingerprint additional JS modules with a different file extension or fingerprint expression in standalone Blazor WebAssembly apps, use the `<StaticWebAssetFingerprintPattern>` property in the app's project file (`.csproj`).

... which is on this PR.

However, I might be misunderstanding the feature/problem with what's in the article.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/static-files.md](https://github.com/dotnet/AspNetCore.Docs/blob/15e6c2e0c942a2df201fcd7798c726eda48b2b64/aspnetcore/blazor/fundamentals/static-files.md) | [aspnetcore/blazor/fundamentals/static-files](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/static-files?branch=pr-en-us-36630) |

### DIFF
--- a/aspnetcore/release-notes/aspnetcore-10/includes/blazor.md
+++ b/aspnetcore/release-notes/aspnetcore-10/includes/blazor.md
@@ -191,7 +191,7 @@ Specify the fingerprint expression with the `<StaticWebAssetFingerprintPattern>`
 </ItemGroup>
 ```
 
-Any JS module in `index.html` with the fingerprint marker is fingerprinted by the framework, including when the app is published.
+Any JS file (`*.js`) in `index.html` with the fingerprint marker is fingerprinted by the framework, including when the app is published.
 
 If you adopt the `.mjs` file extension for JS modules, set the file extension with the `Pattern` parameter:
 


maraf:
The twist is `.js` files are not fingerprinted by default, because we don't know if they will be loaded as modules as by other mechanism. We can fingeprint `.js` files only if they are modules and thus browser will apply importmap when resolving the URL.

guardrex:
The behavior doesn't seem to match that. What am I missing in the following scenario? .........

If I create a plain (non-module) JS script (`wwwroot/script.js`) with a JS FN in it ...

```javascript
function alertMessage() {
  alert("Hello from JavaScript!");
}
```

... and I add the fingerprint expression to its filename ...

```html
<script src="script#[.{fingerprint}].js"></script>
```

It gets fingerprinted and placed into the import map ...

image

image

maraf:
Does it work on publish output?

guardrex:
Not sure ... I didn't test. I'll check that and see what happens. brb

guardrex:
I see what you mean ... no ... it doesn't fingerprint that file after the app is published. I'm not sure if I was just being silly expecting it to work like that.

Anyway ... for the changes that we need here, let me try another pass on this to see if I can get the coverage right. ~I'm still a bit confused tho because `.js` files could be modules or not modules. Does the framework figure that out ... and then if the `.js` file is a module, it will do the fingerprinting?~

Rubber 🦆 says 'no' to that. I just tried with `type="module"`, and it doesn't fingerprint on publish.

guardrex:
I'm attempting to understand this feature better, and I tried the MSBuild config for `.mjs`, but I get ...

> The attribute "Include" in element `<StaticWebAssetFingerprintPattern>` is unrecognized. The project file is invalid on disk and was not reloaded.

~Any idea what might be going wrong?~

guardrex:
Ah .... rubber 🦆 says, 'that's an `<ItemGroup>` you silly 🦖!' 🙈😆 ... We better add that to the coverage.

guardrex:
@maraf ... Ok, I think 🤔 that I'm getting this feature now. See if the diff (I made two commits) covers what we need to say.

~If so, I'll go into the reference article coverage next and make similar updates.~

UPDATE (1/20): I updated the reference guidance to match the changes in the release notes.

guardrex:
@maraf ... I think this is ready for a final review and merge.

aspnetcore/blazor/fundamentals/static-files.md
```diff
@@ -262,12 +262,25 @@ In the project file (`.csproj`), the `<OverrideHtmlAssetPlaceholders>` property
...
+Specify the fingerprint expression with the `<StaticWebAssetFingerprintPattern>` property in the app's project file (`.csproj`):
+
+```xml
+<ItemGroup>
+  <StaticWebAssetFingerprintPattern Include="JSModule" Pattern="*.js" 
+    Expression="#[.{fingerprint}]!" />
+</ItemGroup>
+```
+
+Any JS module in `index.html` with the fingerprint marker is fingerprinted by the framework, including when the app is published.
```
maraf:
```suggestion
Any JS file (`*.js`)in `index.html` with the fingerprint marker is fingerprinted by the framework, including when the app is published.
```

maraf:
```suggestion
Any JS file (`*.js`) in `index.html` with the fingerprint marker is fingerprinted by the framework, including when the app is published.
```

PR #36673, opened 2026-01-23
Change navbar link controller from Movies to HelloWorld

Rei45c:
Cotnributes to #36706


wadepickett:
Thanks @Rei45c!

wadepickett:
The change needs to occur on the .NET 10 sample versio as well, but that was being worked on at the same time as this PR, so to avoid complication, I will approve this one and then update the .NET 10 after.

wadepickett:
closing and reopening to force a rebuild and merge check.

PR #36680, opened 2026-01-25
[Blazor] Preloaded Blazor framework script fixes

hakenr:
1. Fixed title level
2. Removed duplicate section (link to doc moved to first section).


hakenr:
@guardrex It seems this section should be also merged:
https://learn.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-10.0?view=aspnetcore-10.0#blazor-webassembly-static-asset-preloading-in-blazor-web-apps
...plus the link does not carry proper fragment (It should be `#blazor-webassembly-static-asset-preloading`)

PR #36449, opened 2025-12-05
[11.0 P1] Blazor release notes and initial migration guidance

guardrex:
Addresses #36448

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/migration/100-to-110.md](https://github.com/dotnet/AspNetCore.Docs/blob/fdc01a6c0542e599bd6faf7853f85e8aaf131772/aspnetcore/migration/100-to-110.md) | [aspnetcore/migration/100-to-110](https://review.learn.microsoft.com/en-us/aspnet/core/migration/100-to-110?branch=pr-en-us-36449) |

PR #36750, opened 2026-02-06
 [11.0 P1] `NavigateTo` and `NavLink` support `RelativeToCurrentUri`

ilonatommy:
[EDIT by guardrex to update the PU PR number to 64670]

Doc changes proposal for dotnet/aspnetcore#64670
Issue: dotnet/aspnetcore#23615

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/fundamentals/navigation.md](https://github.com/dotnet/AspNetCore.Docs/blob/c4d583c3deffcdd1a3308a3a786954f5360edb7a/aspnetcore/blazor/fundamentals/navigation.md) | [aspnetcore/blazor/fundamentals/navigation](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/navigation?branch=pr-en-us-36750) |

### DIFF
--- a/aspnetcore/blazor/fundamentals/navigation.md
+++ b/aspnetcore/blazor/fundamentals/navigation.md
@@ -80,13 +80,19 @@ There are two <xref:Microsoft.AspNetCore.Components.Routing.NavLinkMatch> option
 
 :::moniker range=">= aspnetcore-11.0"
 
-Set <xref:Microsoft.AspNetCore.Components.Routing.NavLink.RelativeToCurrentUri> to `true` to resolve the `href` relative to the current page path instead of the app's base URI:
+<!-- UPDATE 11.0 - API cross-link
+
+<xref:Microsoft.AspNetCore.Components.Routing.NavLink.RelativeToCurrentUri?displayProperty=nameWithType>
+
+-->
+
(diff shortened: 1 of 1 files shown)


guardrex:
@ilonatommy ... The edit that we need for the second part around Line 770 is too complicated to fix with GH suggestions, so I'll just edit the PR directly for the updates.

The problem with the second part is that ...

* It cuts into an existing versioned content.
* Individual bullets of a bullet list (a rendered unordered list) can't be versioned. The extra lines around it break out the bullet into a new list. Therefore, the whole list must be versioned.

I'll fix it up now, and you'll see what I mean. *Stand-by ....................*

guardrex:
@ilonatommy ... See if my update makes sense ...

* The whole list is versioned so that it doesn't render as two lists.
* Since there's an existing example for `NavigationOptions`, I roll the new API into it.
* Also, we won't have API Browser access to the API until later, so I just left myself an "UPDATE 11.0" note for later.

**UPDATE**: I added it to the release notes at https://github.com/dotnet/AspNetCore.Docs/pull/36449/changes/10a9406fe7962e275d9cdc20f7aef220b509fed1.

ilonatommy:
Thank you, I didn't see that! The update looks good.

PR #36805, opened 2026-02-25
Update routing.md

jz-k:
Updated language to clarify the purpose of Route constraint `required`. The phrase "non-parameter value" was awkward. The updated language clarifies that an explicit value is required, and I hope is less muddy. "Non-parameter value" gave me pause, since we *are* applying the attribute to a parameter.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/fundamentals/routing.md](https://github.com/dotnet/AspNetCore.Docs/blob/d39f9f5a904eb25a0dc4627cc62a2d0b1de4abd3/aspnetcore/fundamentals/routing.md) | [Routing in ASP.NET Core](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/routing?branch=pr-en-us-36805) |

### DIFF
--- a/aspnetcore/fundamentals/routing/includes/routing3-7.md
+++ b/aspnetcore/fundamentals/routing/includes/routing3-7.md
@@ -499,7 +499,7 @@ The following table demonstrates example route constraints and their expected be
 | `range(min,max)`    | `{age:range(18,120)}`                       | `91`                                   | Integer value must be at least 18 but no more than 120                                    |
 | `alpha`             | `{name:alpha}`                              | `Rick`                                 | String must consist of one or more alphabetical characters, `a`-`z` and case-insensitive. |
 | `regex(expression)` | `{ssn:regex(^\\d{{3}}-\\d{{2}}-\\d{{4}}$)}` | `123-45-6789`                          | String must match the regular expression. See tips about defining a regular expression.   |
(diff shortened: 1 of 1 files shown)


guardrex:
AI is correct to suggest updating the INCLUDE file (`routing3-7.md`) for prior versions with the change ...

> Also, the <=7.0 content included from fundamentals/routing/includes/routing3-7.md still uses the old “non-parameter value” phrasing for the required constraint; updating that include would keep the guidance consistent across monikers.

I recommend adding a commit to update it to match ...

> Enforces that an explicit value (not an ambient value) is present during URL generation.

guardrex:
Thanks @jz-k ... Please sign the CLA ☝️ if you agree with its terms. We can't accept a PR unless the CLA is reviewed and signed.

jz-k:
@dotnet-policy-service agree

guardrex:
@cmastr ... The CLA flagging has been broken like this for years ... really as long as I can remember. The CLA notice often appears well after the PR has been submitted and after the dev and our team has had interaction on it, even with additional commits arriving. We were told at the outset not to even look at the content of a PR without a signed CLA, but that's impossible with the CLA notice popping up so late.

PR #36886, opened 2026-03-17
Update .NET to JS interop support table with float array, span and arraysegment

ArcadeMode:
In https://github.com/dotnet/runtime/pull/123642 JS interop support for float[], Span\<float\> and ArraySegment\<float\> was implemented. This is to reflect these types in the support table.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/client-side/dotnet-interop/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/a63665676b52d2ccd0d8025cfaff9aa79aa1cfae/aspnetcore/client-side/dotnet-interop/index.md) | [JavaScript `[JSImport]`/`[JSExport]` interop in .NET WebAssembly](https://review.learn.microsoft.com/en-us/aspnet/core/client-side/dotnet-interop/index?branch=pr-en-us-36886) |

### DIFF
--- a/aspnetcore/client-side/dotnet-interop/index.md
+++ b/aspnetcore/client-side/dotnet-interop/index.md
@@ -154,6 +154,8 @@ Most of the marshalled types work in both directions, as parameters and as retur
 
 The following table indicates the supported type mappings.
 
+:::moniker range=">= aspnetcore-11.0"
+
 | .NET | JavaScript | `Nullable` | `Task` <span aria-hidden="true">➔</span><span class="visually-hidden">to</span> `Promise` | `JSMarshalAs` optional | `Array of` |
 | --- | --- | :---: | :---: | :---: | :---: |
 | `Boolean` | `Boolean` | <span aria-hidden="true">✅</span><span class="visually-hidden">Supported</span> | <span aria-hidden="true">✅</span><span class="visually-hidden">Supported</span> | <span aria-hidden="true">✅</span><span class="visually-hidden">Supported</span> | <span class="visually-hidden">Not supported</span> |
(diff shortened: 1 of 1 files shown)


guardrex:
Thanks, @ArcadeMode! We'll just get Pavel to take a quick look before merging this.

guardrex:
BTW tho ...

For `Span<Single>` and `ArraySegment<Single>`, you didn't update the "Not supported" entries of the table cells. Didn't you mean to make them supported in those cells?

```html
<span aria-hidden="true">✅</span><span class="visually-hidden">Supported</span>
```

guardrex:
One other question on versioning ...

 What release did the support appear? That's going to affect how we implement this change in the article. The entire table must be versioned if the support doesn't go all the way back to .NET 7.

ArcadeMode:
> Did you mean to make them "supported" across the board?

The columns are wether this type is supported in Tasks, Arrays and such. These types are not. The entries have to be the same as the other Span/ArraySegment types on the rows surrounding it. We should not make the suggested changes

ArcadeMode:
> One other question on versioning ...
> 
> What release did the support appear? That's going to affect how we implement this change in the article. The entire table must be versioned if the support doesn't go all the way back to .NET 7.

The types will be added in .NET 11 I think. @pavelsavara can probably answer this question.

What should we do to version the table?

guardrex:
To version the table, a complete copy of it has to be made and placed in moniker range blocks ...

```markdown
:::moniker range=">= aspnetcore-11.0"

----- Table here for >=11.0 with the content updates -----

:::moniker-end

:::moniker range="< aspnetcore-11.0"

----- Current table, which will be for <11.0 -----

:::moniker-end
```

ArcadeMode:
> To version the table, a complete copy of it has to be made and placed in moniker range blocks ...
> 
> ```
> :::moniker range=">= aspnetcore-11.0"
> 
> ----- Table here for >=11.0 with the content updates -----
> 
> :::moniker-end
> 
> :::moniker range="< aspnetcore-11.0"
> 
> ----- Current table, which will be for <11.0 -----
> 
> :::moniker-end
> ```

Thank you, its done @guardrex !

guardrex:
Thanks @ArcadeMode!

@pavelsavara ... When you see this, the first commit shows the changes that are being made to the table for >=11.0 ...

https://github.com/dotnet/AspNetCore.Docs/pull/36886/changes/4484c67fe8f6e72363286382657b8002912b0a26

PR #37045, opened 2026-04-22
Fixed typo

asummers1:
---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/razor-pages/ui-class.md](https://github.com/dotnet/AspNetCore.Docs/blob/85cafad216284d96ba9145e2fc1ebc5fe207cacd/aspnetcore/razor-pages/ui-class.md) | [Create reusable UI using the Razor class library project in ASP.NET Core](https://review.learn.microsoft.com/en-us/aspnet/core/razor-pages/ui-class?branch=pr-en-us-37045) |

PR #37048, opened 2026-04-23
Fix typo in project creation instructions

guardrex:
Addresses #37045

Wade, Tom ... a follow-up patch to #37045.

I fleshed it out a bit more with both approaches, Start Window and menu approaches.

Also, Copilot is recommending an article date update on the 50-character rule. That seems wrong to me, as I'm only looking at and patching a single line. I don't recommend updating the date, but I'll do it if you think it best.

cc: @asummers1

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/razor-pages/ui-class.md](https://github.com/dotnet/AspNetCore.Docs/blob/886530666fc28cebfc327f9e87f2a690fde46c24/aspnetcore/razor-pages/ui-class.md) | [aspnetcore/razor-pages/ui-class](https://review.learn.microsoft.com/en-us/aspnet/core/razor-pages/ui-class?branch=pr-en-us-37048) |

### DIFF
--- a/aspnetcore/razor-pages/ui-class/includes/ui-class6.md
+++ b/aspnetcore/razor-pages/ui-class/includes/ui-class6.md
@@ -9,7 +9,7 @@ For information on how to integrate npm and webpack into the build process for a
 
 # [Visual Studio](#tab/visual-studio)
 
-* Select **Create a new project**.
+* From Visual Studio's **Start Window**, select **Create a new project**. Alternatively, select the **File** > **New** > **Project/Solution** command.
 * Select **Razor Class Library** > **Next**.
 * Name the library (for example, "RazorClassLib"), > **Create**. To avoid a file name collision with the generated view library, ensure the library name doesn't end in `.Views`.
 * Select **Support pages and views** if you need to support views. By default, only Razor Pages are supported. Select **Create**.
@@ -175,7 +175,7 @@ Razor views, pages, controllers, page models, [Razor components](xref:blazor/com
 
 # [Visual Studio](#tab/visual-studio)
 
-* Select **Create a new project**.
+* From Visual Studio's **Start Window**, select **Create a new project**. Alternatively, select the **File** > **New** > **Project/Solution** command.
 * Select **Razor Class Library** > **Next**.
 * Name the library (for example, "RazorClassLib"), > **Create** > **Next**. To avoid a file name collision with the generated view library, ensure the library name doesn't end in `.Views`.
 * Select the **Target Framework**. Check **☑ Support pages and views** to support views. By default, only Razor components are supported. Select **Create**.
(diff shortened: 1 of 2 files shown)


aspnetcore/razor-pages/ui-class/includes/ui-class6.md
```diff
@@ -9,7 +9,7 @@ For information on how to integrate npm and webpack into the build process for a
 
 # [Visual Studio](#tab/visual-studio)
 
-* From Visual Studio select **Create new a new project**.
+* From Visual Studio's **Start Window**, select **Create a new project**. Alternatively, select the **File** > **New** > **Project/Solution** command.
```
wadepickett:
```suggestion
* From Visual Studio's **Start Window**, select **Create a new project**. Alternatively, select **File** > **New** > **Project/Solution**.
```
Minor item: This is a menu item really rather than a  "command" and since it is possible of course to use commands in VS, I suggest dropping that word to keep this a little more clear.

wadepickett (approved):
Approved, see suggested drop of "command".

aspnetcore/razor-pages/ui-class/includes/ui-class6.md
guardrex: I was going by the style manual, which seems to favor the use of "command" over a menu item ...

https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/m/menu-item

I noticed that I have used "menu bar" around the Blazor docs. Let's go with your suggestion because it's obvious that's what the sentence is referring to.

guardrex:
@wadepickett ... Do you think there's any value in a structure like this ...

> Alternatively, select **New** > **Project/Solution** from the **File** menu.

The style manual seems fine with that format ...

https://learn.microsoft.com/en-us/style-guide/procedures-instructions/formatting-text-in-instructions

... where they show ...

> On the **Design** menu, select **Colors**, and then select a color scheme.

wadepickett:
> @wadepickett ... Do you think there's any value in a structure like this ...
> 
> > Alternatively, select **New** > **Project/Solution** from the **File** menu.
> 
> The style manual seems fine with that format ...
> 
> https://learn.microsoft.com/en-us/style-guide/procedures-instructions/formatting-text-in-instructions
> 
> ... where they show ...
> 
> > On the **Design** menu, select **Colors**, and then select a color scheme.

For this:
Alternatively, select **New** > **Project/Solution** from the **File** menu.

My thoughts: Sounds great, but the instructions are backwards in practice.  You have to find the menu at the end of the instruction to select before following the rest of the selection instruction.  Personally I find that slower to work with in execution, but it is also not a huge deal and not something I have ever seen customer feedback from people on either way for over 20 years.  It gets to be more of a problem in the case where there are several sub-menues to select before you find out what menu it is all under in the first place.

My main concern (which was minor) was just calling it a command when it isn't one, in an IDE where commands are also an option through a terminal.

guardrex:
All good. I don't disagree with anything you wrote. I just mentioned it because it plays well with the style manual's use of the word "menu" in this context.

Let's go with what we have now. It's SUPER obvious in this day and age what that string of menu commands means. 😄

PR #37055, opened 2026-04-27
Add link to inspect WebView on Mac Catalyst

### LINKED ISSUE #31874: MacCat WebViewinspectable in Safari dev tools
### Description

Beth ... This is from Dave via email. I'm opening it as an issue for consideration. Idk if it's something that should be documented or not.

I’ve been looking at how you make a MacCat WebViewinspectable in Safari dev tools in .NET 8, which caused me to go back to how it’s inspectable in .NET 7. This then made me look at BlazorWebView. Here’s what I found.

WebView

In .NET 8 and 7, WebView is inspectable by adding the following code to your app (e.g. to MauiProgram.cs):

```csharp
#if MACCATALYST
    Microsoft.Maui.Handlers.WebViewHandler.Mapper.AppendToMapping("Inspect", (handler, view) =>
    {
        if (OperatingSystem.IsMacCatalystVersionAtLeast(16, 4)) handler.PlatformView.Inspectable = true;
    });
#endif
```

The `com.apple.security.get-task-allow` entitlement isn’t required.

BlazorWebView

In .NET 8, BlazorWebView is inspectable by default due to code added to the framework. While the project template is adding the com.apple.security.get-task-allow entitlement, this isn’t required to make the app inspectable. I’m going to log a bug against that.

In .NET 7, I can’t make a BlazorWebView inspectable at all on MacCat. Either through code or through the  com.apple.security.get-task-allow entitlement. Which I think is the point of the code addition to BlazorWebView for .NET 8 (see https://github.com/dotnet/maui/pull/14610 and https://github.com/dotnet/maui/pull/16631).

So what’s all this stuff about the com.apple.security.get-task-allow entitlement I hear you cry? Well, it was required at one point. But it was changes to Apple APIs (shipped via Xcode) that resulted in the entitlement no longer being required. It’s difficult because it’s entirely possible that some users do have combinations of Xcode/.NET that do require the entitlement approach. Quite what those combinations are, is beyond me. But that’s ok. The golden rule is that users always need the latest Xcode version to be able to submit apps to Apple’s store. So we generally ensure that anything in docs is correct as per the latest everything.

I’m mentioning this to you wrt your docs, and I’d recommend you double check by BlazorWebView findings before doing any updates!

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/hybrid/developer-tools?view=aspnetcore-8.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/hybrid/developer-tools.md

### Document ID

48e7c079-a7e4-e0a4-ed19-935436113b70

### Article author

@guardrex

guardrex:
Fixes #31874

This will add a cross-link to the MAUI docs section and briefly remark on the relevant versions. If the dev knows that their versions are beyond those indicated, they shouldn't sweat it.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/hybrid/developer-tools.md](https://github.com/dotnet/AspNetCore.Docs/blob/693ba65c30ea37b4ebb2d1434f7b31277b8fdfc2/aspnetcore/blazor/hybrid/developer-tools.md) | [aspnetcore/blazor/hybrid/developer-tools](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/hybrid/developer-tools?branch=pr-en-us-37055) |

### DIFF
--- a/aspnetcore/blazor/hybrid/developer-tools.md
+++ b/aspnetcore/blazor/hybrid/developer-tools.md
@@ -137,4 +137,4 @@ To use Safari developer tools with a macOS app:
 * [Chrome DevTools](https://developer.chrome.com/docs/devtools/)
 * [Microsoft Edge Developer Tools overview](/microsoft-edge/devtools-guide-chromium/)
 * [Safari Developer Help](https://support.apple.com/guide/safari-developer/welcome/mac)
-* [Inspect a WebView on Mac Catalyst (Mac Catalyst prior to 13.1 and iOS prior to 16.4)](/dotnet/maui/user-interface/controls/webview?pivots=devices-maccatalyst#inspect-a-webview-on-mac-catalyst)
+* [Inspect a `BlazorWebView` on Mac Catalyst (Mac Catalyst prior to 13.1 and iOS prior to 16.4)](/dotnet/maui/user-interface/controls/webview?pivots=devices-maccatalyst#inspect-a-webview-on-mac-catalyst)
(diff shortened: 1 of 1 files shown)


wadepickett (approved):
Approved.

PR #37061, opened 2026-04-28
Update performance notes for SetParametersAsync method

### LINKED ISSUE #35507: [Blazor] Rendering performance - Implement SetParametersAsync manually still valid?
### Description

https://learn.microsoft.com/en-us/aspnet/core/blazor/performance/rendering?view=aspnetcore-9.0#implement-setparametersasync-manually

Is the suggestion to replace `parameters.SetParameterProperties(this)` from `ComponentBase` with a `switch` over parameter names still valid?

Given all the `ComponentProperties.SetProperties()` optimizations, including cached `PropertyWriter`s with `Action`-based setter delegates, I'm wondering if this advice is outdated.

I ran some tests in Blazor WASM with 10,000 instances of a component in a loop, each having 10 parameters. Overriding `SetParametersAsync` with a `switch`-based implementation made no measurable difference.

@guardrex Could you please check with the PU whether it still makes sense to keep this section in the list of potential Blazor performance optimizations? I’m not saying there are *no* scenarios where this could help, but I think they’re rare enough that documenting this might just lead devs down an unproductive path.

The current text says:

> In extreme cases, it can improve rendering performance by 20-25%, but you should only consider this approach in the extreme scenarios listed earlier in this section.

If there are still valid scenarios, maybe we should explain those more clearly than what's currently in that section:

>In some extreme cases, you may wish to avoid the reflection and implement your own parameter-setting logic manually. This may be applicable when:
> 
> - A component renders extremely often, for example, when there are hundreds or thousands of copies of the component in the UI.
> - A component accepts many parameters.
> - You find that the overhead of receiving parameters has an observable impact on UI responsiveness.

My test component:
https://github.com/hakenr/BlazorPerformanceTuningDemos/blob/e2b1704ac95d9b887f6bd972ac4864c7352b9698/BlazorPerformanceTuningDemos/BlazorPerformanceTuningDemos.Client/ParametersCount/TableCell_ManyParameters_Optimized.razor

PS: This isn't my first attempt at creating a synthetic demo that shows a measurable benefit from this optimization. I've already tried several times in the past without success.

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/performance/rendering?view=aspnetcore-9.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/performance/rendering.md

### Document ID

f8a69843-1318-b2de-0161-d4e6a6e34b63

### Platform Id

0fb068b6-35c8-9f91-53eb-1695e45cca93

### Article author

@guardrex

### Metadata

* ID: f8a69843-1318-b2de-0161-d4e6a6e34b63
* PlatformId: 0fb068b6-35c8-9f91-53eb-1695e45cca93 
* Service: **aspnet-core**
* Sub-service: **blazor**

guardrex:
Fixes #35507

Thanks @hakenr! 🎷 

Wade/Tom ... Just need one review to get this one in.

Clarified the performance improvement potential of overriding `SetParametersAsync`, noting it typically yields under 10% improvement in .NET 10.

Notes ...

* I'm using the HTML named entity for the em dash because the GH markdown editor doesn't make it clear that the character is indeed an em dash.
* I'm going to peg the "current release" remark at .NET 10 because we don't know how this might change in the future and the "current release" could be anything when the content targets >=8.0. Benchmark testing was performed for .NET 10, so we can only state the figures for that release.
* This must be versioned because "Interactive Server" has no meaning prior to .NET 8.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/performance/rendering.md](https://github.com/dotnet/AspNetCore.Docs/blob/1b9150fd98a591f3d540613d055a7cba60a4e77a/aspnetcore/blazor/performance/rendering.md) | [aspnetcore/blazor/performance/rendering](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/performance/rendering?branch=pr-en-us-37061) |


guardrex:
Sorry for all of the update pings, @hakenr. The lingo is a bit tricky. I think I have it now ... or close to what we'll go with. Please feel free to suggest updates. Wade or Tom will be along shortly to review, and then we'll get this merged. *Thanks again!*

hakenr (approved):
lgtm

aspnetcore/blazor/performance/rendering.md
wadepickett:
```suggestion
ms.date: 04/28/2026
```
Reporting ends up coming back to haunt us if I don't udpate these when a relatively significant change occurs.

wadepickett (approved):
Looks great, approved.

Issue #35972, opened 2025-08-21
[RC1] Update Blazor form validation docs for .NET 10


oroztocil:
### Description

Suggestions for improvements of the Blazor form validation docs for .NET 10:

- Document [new features added in RC1](https://github.com/dotnet/aspnetcore/pull/63231/)
    - `SkipValidation` attribute
    - Support for type-level validation attributes (in addition to the already existing `IValidatableObject` support)

- Document behavior of the DataAnnotationsValidator
    - After RC1 the new validation infrastructure and consequently the DataAnnotationsValidator component have the same behavior with regards to validation order and short-circuiting as `System.ComponentModel.DataAnnotations.Validator`
    - We could add a brief description of this logic so that user's are not surprised about the validation outcome when they have a combination of property-level and type-level validation rules

- Remove information about `ObjectGraphDataAnnotationsValidator`
    - The docs for complex object validation contain a mixture of information describing both the new validation implementation and the deprecated `ObjectGraphDataAnnotationsValidator` from the unsupoorted NuGet package
    - We want to remove all mentions of the old package
    - Optionally, leave a short note stating that `ObjectGraphDataAnnotationsValidator` should not be used and point the users to the improved capabilities of the standard component

- ***MOVED TO #36051*** ~Idea: Replace use of MVC controller with Minimal API in the section "Server validation with a validator component" to showcase the unified validation experience~
    - ~The API controller uses MVC-based validation which also uses `DataAnnotations` attribute to specify validation rules. However, it is configured differently and might have different behavior in some cases (e.g. validation is skipped with `ValidateNever` attribute in MVC but with `SkipValidation` attribute in the unified validation code)~
    - ~Up for debate: Speaking of "server validation" and "client validation" might also be a bit confusing in the context of Blazor Server - maybe rephrase the section using terminology such as "data API", "backend", or "backend API"?~

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/forms/validation?view=aspnetcore-10.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/forms/validation.md

### Document ID

bcb470e1-c859-0549-b4ef-c6c9b7ef2c72

### Platform Id

b4c1bad9-fc7a-21f5-2552-f0f3617f77db

### Article author

@guardrex

### Metadata

* ID: bcb470e1-c859-0549-b4ef-c6c9b7ef2c72
[...]

PR #37113, opened 2026-05-06
Server validation with a validator component upgrade

### LINKED ISSUE #36051: Server validation with a validator component upgrade
### Description

Per Ondřej's suggestion in #35972 (for .NET 6 or later because Minimal APIs didn't arrive until .NET 6) ...

- Idea: Replace use of MVC controller with Minimal API in the section "Server validation with a validator component" to showcase the unified validation experience
    - The API controller uses MVC-based validation which also uses `DataAnnotations` attribute to specify validation rules. However, it is configured differently and might have different behavior in some cases (e.g. validation is skipped with `ValidateNever` attribute in MVC but with `SkipValidation` attribute in the unified validation code)
    - Up for debate: Speaking of "server validation" and "client validation" might also be a bit confusing in the context of Blazor Server - maybe rephrase the section using terminology such as "data API", "backend", or "backend API"?

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/blazor/forms/validation?view=aspnetcore-9.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/blazor/forms/validation.md

### Document ID

bcb470e1-c859-0549-b4ef-c6c9b7ef2c72

### Platform Id

b4c1bad9-fc7a-21f5-2552-f0f3617f77db

### Article author

@guardrex

### Metadata

* ID: bcb470e1-c859-0549-b4ef-c6c9b7ef2c72
* PlatformId: b4c1bad9-fc7a-21f5-2552-f0f3617f77db 
* Service: **aspnet-core**
* Sub-service: **blazor**

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+bcb470e1-c859-0549-b4ef-c6c9b7ef2c72)

guardrex:
Fixes #36051

Ondřej ... Per your suggestion in #35972, the general idea is to replace the MVC controller with a Minimal API validation, including the unified validation experience.

To simplify the section, the content is versioned into >=10.0, >=8.0/<10.0, and <8.0 blocks. ***I think we only need to focus on the new guidance in the >=10.0 coverage at the top of the section.*** I haven't received any complaints (or *death threats* 💀😨😆) from our readers on our MVC controller-based guidance.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/forms/validation.md](https://github.com/dotnet/AspNetCore.Docs/blob/d95a5c80e70cb14e11deec5c408f99378c60d63b/aspnetcore/blazor/forms/validation.md) | [aspnetcore/blazor/forms/validation](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/forms/validation?branch=pr-en-us-37113) |

### DIFF
--- a/aspnetcore/blazor/forms/validation.md
+++ b/aspnetcore/blazor/forms/validation.md
@@ -472,21 +472,28 @@ internal sealed class ClientFormValidator(HttpClient httpClient) : IFormValidato
             Status = StatusCodes.Status500InternalServerError
         };
 
-        using var response = await httpClient.PostAsJsonAsync("/starship-validation", starship);
+        try
+        {
+            using var response = await httpClient.PostAsJsonAsync("/starship-validation", starship);
 
-        if (response.IsSuccessStatusCode)
-        {
-            return await response.Content.ReadFromJsonAsync<ValidationProblemDetails>() ?? new ValidationProblemDetails
+            if (response.IsSuccessStatusCode)
             {
-                Title = "Validation succeeded",
-                Detail = "The starship form is valid.",
-                Instance = "BlazorSample.Client.ClientFormValidator",
-                Status = StatusCodes.Status500InternalServerError
-            };
+                return await response.Content.ReadFromJsonAsync<ValidationProblemDetails>() ?? new ValidationProblemDetails
+                {
+                    Title = "Validation succeeded",
+                    Detail = "The starship form is valid.",
+                    Instance = "BlazorSample.Client.ClientFormValidator",
+                    Status = StatusCodes.Status500InternalServerError
+                };
+            }
+            else if (response.StatusCode == HttpStatusCode.BadRequest)
+            {
+                return await response.Content.ReadFromJsonAsync<ValidationProblemDetails>() ?? problemDetails;
+            }
         }
-        else if (response.StatusCode == HttpStatusCode.BadRequest)
+        catch (Exception ex)
         {
-            return await response.Content.ReadFromJsonAsync<ValidationProblemDetails>() ?? problemDetails;
+            //Log the exception or handle it as needed
         }
 
         return problemDetails;
@@ -547,6 +554,8 @@ internal sealed class ServerFormValidator(IDownstreamApi downstreamApi)
         catch (HttpRequestException ex) when 
             (ex.StatusCode == HttpStatusCode.BadRequest)
         {
+            // The response content is only available via
+            // the Message property of the exception
             int index = ex.Message.IndexOf('{');
 
             if (index != -1)
@@ -598,7 +607,7 @@ builder.Services.AddHttpClient<IFormValidator, ClientFormValidator>(httpClient =
 
 The preceding example sets the base address with `builder.HostEnvironment.BaseAddress` (<xref:Microsoft.AspNetCore.Components.WebAssembly.Hosting.IWebAssemblyHostEnvironment.BaseAddress%2A?displayProperty=nameWithType>), which gets the base address for the app and is typically derived from the `<base>` tag's `href` value in the host page.
 
-In the `Program` file of the `MinimalApiJwt` project, add the following starship form validation endpoint. The endpoint validates that the model's `Description` property has a value when the model's `Classification` property is `Defense`. If validation fails, a dictionary of validation errors returns a dictionary entry with the failed field and a description of the error. If validation passes, an empty dictionary is returned. In a typical production app, any number of form model checks are made, and the validation errors dictionary can include multiple failures (`List<string>` value) for each model property (`validationErrors[nameof({PROPERTY})]`).
+In the `Program` file of the `MinimalApiJwt` project, add the following starship form validation endpoint. The endpoint validates that the model's `Description` property has a value when the model's `Classification` property is `Defense`. If validation fails, a `ValidationProblem` returns a dictionary with the failed field and a description of the error. If validation passes, a *204 - No Content* response is issued. In a typical production app, any number of custom form model checks are made, and the validation errors dictionary can include multiple failures (`string[]` value) for each model property.
 
 In the `Program` file of the Minimal API project:
 
--- a/aspnetcore/blazor/forms/validation.md
+++ b/aspnetcore/blazor/forms/validation.md
@@ -361,41 +361,78 @@ When validation messages are set in the component, they're added to the validato
 
 :::moniker range=">= aspnetcore-10.0"
 
-*This section is focused on Blazor Web App scenarios, but the approach for any type of app that uses server validation with web API adopts the same general approach. The approach in this section is based on a [Blazor Web App (Interactive Auto render mode) app hosted in Entra with Microsoft Azure packages](xref:blazor/security/blazor-web-app-entra).*
+*This section demonstrates server validation using a [Blazor Web App (Interactive Auto render mode) hosted in Entra with `Microsoft.Identity.Web` packages](xref:blazor/security/blazor-web-app-entra), but the same general approach is recommended for any Blazor app.*
 
 Server validation is supported in addition to client validation:
 
 * Process client validation in the form with the <xref:Microsoft.AspNetCore.Components.Forms.DataAnnotationsValidator> component.
-* When the form passes client validation (<xref:Microsoft.AspNetCore.Components.Forms.EditForm.OnValidSubmit> is called), send the <xref:Microsoft.AspNetCore.Components.Forms.EditContext.Model?displayProperty=nameWithType> to a backend server API for form processing.
-* Process model validation on the server.
-* The server API includes custom validation logic supplied by the developer.
-* Either disable the form on success or display the errors.
+* When the form passes client validation (<xref:Microsoft.AspNetCore.Components.Forms.EditForm.OnValidSubmit> is called), send the <xref:Microsoft.AspNetCore.Components.Forms.EditContext.Model?displayProperty=nameWithType> to a backend server web API for server-side validation.
+* Process model validation on the server with custom validation logic. If the model is valid, process the form on the server.
+* Send validation errors, if any, back to the client.
+* Either disable the form on success or display the errors, allowing the user to correct any problems with the form's field values.
 
-Basic validation is useful in cases where the form's model is defined within the component hosting the form, either as members directly on the component or in a subclass. Use of a validator component is recommended where an independent model class is used across several components.
+Basic validation is useful in cases where the form's model is defined within the component hosting the form, either as members directly on the component or in a subclass. Use of a *validator component* is recommended where an independent model class is used across several components. The approach demonstrated by the following guidance uses a validator component.
 
 The following example is based on:
 
-* A Blazor Web App with Interactive WebAssembly components created from the [Blazor Web App project template](xref:blazor/project-structure).
-* The `Starship` model  (`Starship.cs`) of the [Example form](xref:blazor/forms/input-components#example-form) section of the *Input components* article.
+* A Blazor Web App with global Interactive Auto components created from the [Blazor Web App project template](xref:blazor/project-structure).
 * The `CustomValidation` component shown in the [Validator components](#validator-components) section.
+* A [Minimal API](xref:fundamentals/minimal-apis) project validates that a ship description field (`Description`) has a value if the user selects the `Defense` ship classification (`Classification`).
 
-In the main project of the Blazor Web App, a Minimal API is added to validate that a value is provided for the ship's description (`Description`) if the user selects the `Defense` ship classification (`Classification`).
-
-The validation for the `Defense` ship classification only occurs on the server because the upcoming form doesn't perform the same validation client-side when the form is submitted to the server. Server validation without client validation is common in apps that require private business logic validation of user input on the server. For example, private information from data stored for a user might be required to validate user input. Private data obviously can't be sent to the client for client validation.
+The validation for the `Defense` ship classification only occurs on the server because the upcoming form doesn't perform the same validation client-side when the form is submitted to the server. Server validation without client validation is common in apps that require private business logic validation of user input on the server. For example, private information from data stored for a user might be required to validate user input. Private data is never sent to the client for client validation.
 
 > [!NOTE]
-> The Minimal API only accepts tokens for users that have the "`API.Access`" scope for this API. Additional customization is required if the API's scope name is different from `API.Access`.
+> For more information on security pertaining to the following example, see the following resources:
 >
-> For more information on security, see:
->
+> * <xref:blazor/security/blazor-web-app-entra>
 > * <xref:blazor/security/index> (and the other articles in the Blazor *Security and Identity* node)
 > * [Microsoft identity platform documentation](/entra/identity-platform/)
 
-Place the `Starship` model (`Starship.cs`) into a `Starship` folder in the `.Client` project. Since the model requires data annotations, confirm that the shared class library uses the shared framework or add the [`System.ComponentModel.Annotations` package](https://www.nuget.org/packages/System.ComponentModel.Annotations) to the shared project.
+> [!NOTE]
+> Long code lines in the following examples are broken into shorter lines to reduce the appearance of horizontal scrollbars on mobile devices. Where appropriate, you're welcome to combine lines to shorten the length of the code.
 
-[!INCLUDE[](~/includes/package-reference.md)]
+Create a `Starship` folder in the `.Client` project of the Blazor Web App.
 
-Add an interface for a form validation service to the `.Client` project.
+Place the following `StarshipModel` model (`StarshipModel.cs`) into the `Starship` folder ***and*** into the Minimal API project of the solution.
+
+> [!NOTE]
+> If you choose to place one copy of the `StarshipModel` into a shared class library project for use by both the Blazor Web App and the Minimal API project, confirm that the shared class library uses the shared framework or add the [`System.ComponentModel.Annotations` package](https://www.nuget.org/packages/System.ComponentModel.Annotations) to the shared project. This ensures that the model has access to data annotations.
+>
+> [!INCLUDE[](~/includes/package-reference.md)]
+
+In the two `StarshipModel` classes, set the namespace (`{NAMESPACE}`) appropriately for each project (for example, `BlazorSample.Starship` in the Blazor Web App and `MinimalApiJwt.Models` in the Minimal API project). Some developers prefer to use a different folder scheme. If you position the classes in the projects in different locations than what is demonstrated here, set the namespaces appropriately.
+
+`Starship/StarshipModel.cs` (Blazor Web App) or `Models/StarshipModel.cs` (Minimal API project):
+
+```csharp
+using System.ComponentModel.DataAnnotations;
+
+namespace {NAMESPACE};
+
+public class StarshipModel
+{
+    [Required]
+    [StringLength(16, ErrorMessage = "Identifier too long (16 character limit).")]
+    public string? Id { get; set; }
+
+    public string? Description { get; set; }
+
+    [Required]
+    public string? Classification { get; set; }
+
+    [Range(1, 100000, ErrorMessage = "Accommodation invalid (1-100000).")]
+    public int MaximumAccommodation { get; set; }
+
+    [Required]
+    [Range(typeof(bool), "true", "true", ErrorMessage = "Approval required.")]
+    public bool IsValidatedDesign { get; set; }
+
+    [Required]
+    public DateTime ProductionDate { get; set; }
+}
+```
+
+Add an interface for a form validation service to the `.Client` project in the `Starship` folder. The interface is used to register a service for server validation on the server or client validation on the client.
 
 `Starship/IFormValidator.cs`:
 
@@ -404,43 +441,12 @@ namespace BlazorSample.Client.Starship;
 
 public interface IFormValidator
 {
-    Task<Dictionary<string, List<string>>> ValidateStarshipFormAsync(StarshipModel starship);
+    Task<Dictionary<string, List<string>>> ValidateStarshipFormAsync(
+        StarshipModel starship);
 }
 ```
 
-Create a server form validator that implements the preceding interface in the Blazor Web App.
-
-`Starship/ServerFormValidator.cs`:
-
-```csharp
-using BlazorSample.Client.Starship;
-using Microsoft.Identity.Abstractions;
-
-namespace BlazorSample.Server.Starship;
-
-internal sealed class ServerFormValidator(IDownstreamApi downstreamApi) : IFormValidator
-{
-    public async Task<Dictionary<string, List<string>>> ValidateStarshipFormAsync(StarshipModel starship)
-    {
-        var validationErrors = await downstreamApi.CallApiForUserAsync<StarshipModel, Dictionary<string, List<string>>>(
-            "DownstreamApi", starship, 
-            options =>
-            {
-                options.HttpMethod = HttpMethod.Post.Method;
-                options.RelativePath = "/starship-validation";
-            });
-
-        return validationErrors ?? new Dictionary<string, List<string>>
-        {
-            { "Validation", [ "An error occurred during validation." ] }
-        };
-    }
-}
-```
-
-Confirm or update the namespace of the preceding class (`BlazorSample.Server.Starship`) to match the app's namespace.
-
-Add a client form validator class to the `.Client` project.
+Add a client form validator class to the `.Client` project's `Starship` folder. The client form validator is used when the app is running on the client. The validator class posts the form's model to the backend Minimal API for processing.
 
 `Starship/ClientFormValidator.cs`:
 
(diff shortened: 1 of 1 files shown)


oroztocil:
Hi @guardrex, sorry for not getting to this sooner. I did read through the whole PR but a first general comment is that by using a validated Minimal API endpoint (instead of a MVC controller) I meant using the built-in validation feature we added in .NET 10 that you enable simply by calling `builder.Services.AddValidation()`. A built-in filter in the request pipeline then validates all the `DataAnnotations` attributes on the request parameters automatically (including nested properties of model classes).

That should significantly simplify the examples because you don't need to build and register any custom validators (the `IFormValidator` types in your example), you don't need to create `ProblemDetails` manually, etc.

guardrex:
@oroztocil ... No worries on the delay. I've needed the time to work on this approach.

[**UPDATE (5/12)**: I see that you did a [Community Standup with Dan](https://www.youtube.com/watch?v=X-qAr4uqAzc) a few weeks ago. I'll watch that and roll concepts into the coverage from your remarks there.]

This is a complicated subject to even discuss here ... lots of moving parts. I'll try to break it down on how I'm thinking about it, and then you can tell me what I'm getting wrong or missing .........

I have the built-in Data Annotations (DA) validation feature (`builder.Services.AddValidation()`) included where the approach in this section states the following ...

> Also in the \`Program\` file of the Minimal API, register \[built-in validation services](xref:fundamentals/minimal-apis#validation-support-in-minimal-apis):
> 
> \`\`\`csharp
> builder.Services.AddValidation();
> \`\`\`
> 
> Built-in validation automatically intercepts the endpoint request and validates the types that the endpoint receives. If the model fails validation, the framework returns a \*400 - Bad Request\* response with error details without executing the endpoint's code.

The link goes here ...

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis?view=aspnetcore-10.0#enable-built-in-validation-support-for-minimal-apis

For one thing, that's not enough information even in this section. It should state that it's specifically DA validation that takes place. It also doesn't include links to the additional content that we placed at .NET 10's release on [*Nested objects and collection types*](https://learn.microsoft.com/en-us/aspnet/core/blazor/forms/validation?view=aspnetcore-10.0#nested-objects-and-collection-types) and [*Use validation models from a different assembly*](https://learn.microsoft.com/en-us/aspnet/core/blazor/forms/validation?view=aspnetcore-10.0#use-validation-models-from-a-different-assembly). I'll need to work on those remarks in this section.

Beyond that, I realize now that this isn't enough coverage in Blazor's *Validation* article on the built-in validation in general. I now think that there should be additional coverage on the built-in validation either in the introductory remarks of this article or in a new section near the top of the article.

But for your remark ...

> That should significantly simplify the examples because you don't need to build and register any custom validators (the IFormValidator types in your example), you don't need to create ProblemDetails manually, etc.

Don't we still need all of this infrastructure for truly *custom business logic validation* that can't be easily addressed by DA validation and where a custom `ValidationAttribute` isn't what the dev wants to use? For example ...

* When one or more server-side database queries are needed to process validation on a form field.
* When several field values must be analyzed in order to process validation on a form field.
* When complex operations must be performed on the user's input from a form field to process validation on that field.

... cases like those. In this example, the custom logic is to see what's in another field to determine if a different form field should pass validation.

Even if that's correct ......

Making matters worse in the global Auto BWA case, the server and client form validation services all need wiring up and `ProblemDetails` have to go through the server app from the client when running on the client. Downstream API's `CallApiForUserAsync` call has been a real PITA because the code ...

```csharp
var response = await downstreamApi.CallApiForUserAsync<StarshipModel, ValidationProblemDetails>(...);
```

... hasn't been behaving. It's either throwing an exception in some cases or I don't get back the expected `ValidationProblemDetails` from it ... `response` is `null` ... that's what I'm fighting with at the moment.

Anyway ... setting my current coding troubles aside for a minute ... do you agree that we still need a section like this for custom/complex validation that DA can't handle easily, where the dev wants to write code into the Minimal API to manually determine if a field should be validated?

guardrex:
@oroztocil ... Update on this ... I was pulled off for a couple of days for the Preview 4 release. I'll be getting right back on this on Monday. I'm making some progress with it. I hope to be able to present you with a good, working implementation soon! Have a great weekend! 🎉

guardrex:
@oroztocil ... Update on this and the latest commit ......

* I'm added a section that focuses on calling `AddValidation` for the built-in type validation server-side. It cross-links to https://learn.microsoft.com/aspnet/core/fundamentals/minimal-apis#enable-built-in-validation-support-for-minimal-apis.
* In the custom validation logic scenario on the server (again, Minimal API), I've really had a hard time getting Downstream API's POST (`PostForUserAsync`) to work with returning validation problem details. That API really likes to ...
  * Throw an exception for a 400, so you have to parse the exception to get the actual problem details JSON for use by the app.
  * Returning `null` instead of an expected `HttpResponseMessage` when called (or a `ValidationProblemDetails` for that matter). I would need to take a little more time looking at their docs to figure out what's going wrong.
Anyway, it doesn't really matter that the Downstream API is giving me trouble WRT what's on the PR now. What's on the PR now uses a vanilla `HttpClient` call with the access token and deserializes the response into `ValidationProblemDetails`. The `Errors` are then returned as a dictionary, where the component can use the `CustomValidation` component to display the errors.
* The code is still prototyped. It needs more work, even if this is a decent approach.
* The code seems to run OK on both Interactive Server and Interactive WebAssembly render modes, so I think it would be fine for Interactive Auto. I'll do a little more testing to confirm that that's true.
* Note again that the purpose of this section isn't merely DA validation. The use case is when the server has to perform complicated validation logic in developer code ... more complicated than one would want to put into a custom `ValidationAttribute`. However, I would expect a custom `ValidationAttribute` to work with `AddValidation`. I probably should both test that scenario and mention it to devs in the section ... or perhaps it warrants its own section, so there would be sister sections for these two approaches ... OR even three of these sections ......
  * *Server validation with a validator component (basic `AddValidation` with model properties approach)*
  * *Server validation with a validator component (custom `ValidationProblemDetails` approach)*
  * *Server validation with a validator component (custom `ValidationAttribute` approach)*

If I'm totally missing a concept for how the Minimal API's `return Results.ValidationProblem()` gets piped into the `CustomValidation` component in this BWA InteractiveAuto case, then I'm 👂 for what I'm missing in my understanding.

guardrex:
@oroztocil ... I confirmed that the server-side built-in validation (`AddValidation`) is working with the model.

When I removed the "Accommodation" field range (1-100000) DA validation from the BWA's `StarshipModel` class and placed a `StarshipModel` class in the backend Minimal API app with an "Accommodation" field range of 200000-1000000, it sent back a 400 with a problem details that was successfully used by the custom validation component to set the error on the form correctly ...

image

If I then correct that to get the built-in validation to pass but break the special check performed in code, it then also throws the 400 with the problem details showing the expected form result ...

image

AFAICT, what's on the PR is fine with perhaps a few more code improvements.

I also perform a quick test to confirm that a custom validation attribute in the Minimal API works well with the built-in validation. I'm sure it will work fine. Then, I'll add a quick remark (or a short section) to this PR on it.

*Stand-by* a bit longer ... making progress here! 🎉 ... and we might have some decent draft coverage to look at soon.

guardrex:
It works! 🎉 

In the Minimal API ...

```csharp
using System.ComponentModel.DataAnnotations;

namespace MinimalApiJwt;

public class CustomValidator : ValidationAttribute
{
    protected override ValidationResult IsValid(object? value,
        ValidationContext validationContext)
    {
        if (value is not null)
        {
            var intValue = (int)value;

            if (intValue > 500000)
            {
                return ValidationResult.Success!;
            }
        }

        return new ValidationResult("Accommodation must be greater than 500000.",
            [validationContext.MemberName!]);
    }
}
```

In the `StarshipModel` class in the Minimal API ...

```csharp
[CustomValidator]
public int MaximumAccommodation { get; set; }
```

Result when an invalid figure is provided ...

image

Correcting the figure, the validation passes ...

image

guardrex:
@oroztocil ... Made a final round of updates to the draft content. I added some remarks (not a section) on custom validation attributes. It's ready for review now.

aspnetcore/blazor/forms/validation.md
```diff
@@ -357,126 +358,582 @@ When validation messages are set in the component, they're added to the validato
...
 
-:::moniker range=">= aspnetcore-8.0"
+    [Required]
+    public DateTime ProductionDate { get; set; }
+}
+```
+
+Add an interface for a form validation service to the `.Client` project in the `Starship` folder. The interface is used to register a service for server validation on the server or client validation on the client.
+
+`Starship/IFormValidator.cs`:
```
oroztocil: Is it necessary to have the `IFormValidator` and the two implementations? If nothing else, the naming `*Validator` may confuse people who are accustomed to the `DataAnnotationValidator` (which is a component they insert into `EditForm`).

aspnetcore/blazor/forms/validation.md
```diff
@@ -357,126 +358,582 @@ When validation messages are set in the component, they're added to the validato
...
+            if (httpContextAccessor.HttpContext is null)
             {
-                ModelState.AddModelError(nameof(model.Description),
-                    "For a 'Defense' ship " +
-                    "classification, 'Description' is required.");
+                throw new Exception("HttpContext not available");
             }
-            else
+
+            var request = new HttpRequestMessage(HttpMethod.Post,
```
oroztocil, line 555: Why is the Server project delegating to a Minimal API endpoint if it can invoke the custom validation code directly? We could get rid of all the code involvingc `IHttpContextAccessor`, token passing, JSON de/seralization, and so on.

aspnetcore/blazor/forms/validation.md
```diff
@@ -357,126 +358,582 @@ When validation messages are set in the component, they're added to the validato
...
+}).RequireAuthorization();
+```
+
+Also in the `Program` file of the Minimal API, register [built-in validation services](xref:fundamentals/minimal-apis#validation-support-in-minimal-apis):
+
+```csharp
+builder.Services.AddValidation();
+```
+
+Built-in validation automatically intercepts the endpoint request and validates the types that the endpoint receives. If the model fails validation, the framework returns a *400 - Bad Request* response with error details without executing the endpoint's code.
```
oroztocil: Probably better to not mix the standard validation pattern with `AddValidation` and `DataAnnotations` validation attributes with the custom validation. Or if the intention is exactly to show how the custom validation can be added on top of the built-in one, then call this out explicitly somewhere here.

aspnetcore/blazor/forms/validation.md
```diff
@@ -357,126 +358,582 @@ When validation messages are set in the component, they're added to the validato
...
+`CustomValidation.cs`:
+
+```csharp
+using Microsoft.AspNetCore.Components;
+using Microsoft.AspNetCore.Components.Forms;
+using Microsoft.AspNetCore.Mvc;
+
+namespace BlazorSample.Client;
+
+public class CustomValidation : ComponentBase
```
oroztocil, line 679: Is it necessary to introduce the `ClientValidation` class that wraps work with the message store?

aspnetcore/blazor/forms/validation.md
```diff
@@ -357,126 +358,582 @@ When validation messages are set in the component, they're added to the validato
...
+The implementation automatically discovers types that are defined in Minimal API handlers or as base types of types defined in Minimal API handlers. An endpoint filter performs validation on these types and is added for each endpoint.
 
-*This section is focused on hosted Blazor WebAssembly scenarios, but the approach for any type of app that uses server validation with web API adopts the same general approach.*
+Built-in validation also supports [custom validation attributes](xref:mvc/models/validation#custom-attributes).
+
+For more information, see <xref:fundamentals/minimal-apis#enable-built-in-validation-support-for-minimal-apis>.
 
 :::moniker-end
 
+## Server validation with a validator component
```
oroztocil: This example still seems overly complex. I appreciate the work that went into this but I am a big fan of docs that are minimal in showing only what is specific for the scenario being taught and omitting custom infrastructure and ceremonies.

First, can we find a different name for this section? I know there are precedents in the docs but I would move away from calling this "server validation". Maybe "remote validation"? The pattern here is basically what the MVC's `Remote` validation attribute did, only with more manual plumbing. Also, where is the validator component mentioned in the title? The `ClientValidation` class is just a wrapper for the message store.

Second, can we reorder the sections in the long Blazor forma validation article so that the docs flow from the most basic/standard/default scenarios to the more advanced/custom ones? For typical Blazor apps the order of complexity is roughly:

1. Built-in validation rules using validation attributes and a `DataAnnotationsValidator` inside `EditForm`.
2. Custom validation rules using custom validation attributes or `IValidatableObject` and a `DataAnnotationsValidator` inside `EditForm`.
3. Information about `AddValidation`.
5. Custom validation logic using the API of `EditForm` and `EditContext`.
6. Only then something like this.

Thinking about it more, it might help us to break up this long article into one that cover 1-3 and another covering 5-6.

guardrex:
cc: @danroth27 ... You might want to weigh in on this one. This is the section that @oroztocil and I are discussing for updates ...

[Server validation with a validator component](https://learn.microsoft.com/en-us/aspnet/core/blazor/forms/validation?view=aspnetcore-10.0#server-validation-with-a-validator-component)

What I have on this PR works 🎉, but the whole section might have outlived your desire to maintain it, falling into the "an advanced scenario for devs to work out on their own" bucket. 😄

BTW @oroztocil ... We're still waiting on getting the P5 PRs merged. Wade and I are on the case, and we'll get the P5 content live *ASAP* 🏃‍♂️.

I opened [Improve section order and possibly split the article (`dotnet/AspNetCore.Docs` #37249)](https://github.com/dotnet/AspNetCore.Docs/issues/37249). [BTW ... I'm going to be OOF for two weeks at the end of this month. If I'm invited back for FY27, I'll be back in early July. If not, this might fall to the docs folks to take over depending on how much work needs to be done here.]

WRT this section ...

> I would move away from calling this "server validation". Maybe "remote validation"?

Yes ... I'll update the section heading. 👍 

> If nothing else, the naming \*Validator may confuse people who are accustomed to the DataAnnotationValidator

Ok ... easy fix. 👍 

> Probably better to not mix the standard validation pattern with AddValidation and DataAnnotations validation attributes with the custom validation. Or if the intention is exactly to show how the custom validation can be added on top of the built-in one, then call this out explicitly somewhere here.

I think calling it out makes sense. The dev can decide to drop built-in validation if they're only looking to implement custom remote validation. I'm glad to see both working well together! *It's pretty sweet IMO!*

> Why is the Server project delegating to a Minimal API endpoint if it can invoke the custom validation code directly? 

Under this paradigm, the Minimal API is part of the use case. There would be several user-facing apps ... BWAs and perhaps other types of apps ... all hitting this other web API app/project for their centralized validation. Also, this is probably the most complicated scenario with the BWA adopting Interactive Auto rendering 😵😄, given that the BWA requires the two-service pattern for server/client rendering.

If a dev is only adopting SSR ***or*** CSR across their apps, this is an example that they can deconstruct down to a simpler remote validation scenario without the Interactive Auto rendering complexity.

> This example still seems overly complex. I appreciate the work that went into this but I am a big fan of docs that are minimal in showing only what is specific for the scenario being taught and omitting custom infrastructure and ceremonies.

Yes, but it seems like a lot of new-to-.NET/new-to-Blazor/new-to-programming devs are going to be significantly challenged trying to do this on their own.

Of course, you could let me know that you feel this whole section's approach isn't welcome any longer for any of several reasons. Do you want to consider dropping the entire section? I leave it up to you and @danroth27 to let me know. 👂  I'll try to get it resolved either way before my OOF period starts on Tuesday.

guardrex:
@oroztocil ... Updates on the last commit ...

* Changed the following naming ...
  * `IFormValidator` to `IFormValidation`.
  * `ClientFormValidator` to `ClientFormValidation`
  * `ServerFormValidator` to `ServerFormValidation`
* Refactored "server validation" into "remote validation" language throughout the section across versions, including in the section heading (*Server validation with a validator component* 👉 *Remote validation with a validator component*).

The overall paradigm is unchanged ... multiple BWAs and other apps hitting a Minimal API for validation. The Minimal API uses built-in and custom validation. The built-in validation aspect is called out both in the section's lead-in remarks and by an added remark near the `AddValidation` call explaining that it can be dropped (don't call `AddValidation`) if the dev only wants custom validation. The BWA goes with Interactive Auto because it's the most complicated case, and I think devs can deconstruct it if they're adopting non-Auto render modes.

Still ready to 💀 the entire section if you and @danroth27 prefer to label this an advanced case for devs to work out on their own. Otherwise tho, please check my 🦖 *RexHacks!*&trade; 🦖 for poor coding 🙈 ... but ... what's here seems to at least work 🎉.

Let's take up reordering the sections and probably breaking the article into two articles on the other issue that I opened in July if I'm still here 🤞🍀.

aspnetcore/blazor/forms/validation.md
```diff
@@ -355,128 +356,584 @@ When validation messages are set in the component, they're added to the validato
...
+In the Blazor Web App, create a server form validator that implements the `IFormValidation` interface. Place the server form validator class in the server-side `Starship` folder. The server form validator is used when the Blazor Web App is running on the server. The validator class posts the form's model to the backend Minimal API for processing.
+
+`Starship/ServerFormValidation.cs`:
+
+```csharp
+using System.Net;
+using System.Net.Http.Headers;
+using System.Text.Json;
+using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Mvc;
```
wadepickett, line 528:
```suggestion
```
The class only uses `ComponentBase`, `EditContext`, `ValidationMessageStore`, and `EditForm`, all from the Components namespaces. The MVC using is unused as far as I can see and should be removed.  Maybe I'm missing something.

aspnetcore/blazor/forms/validation.md
```diff
@@ -355,128 +356,584 @@ When validation messages are set in the component, they're added to the validato
...
+
+```razor
+<div class="nav-item px-3">
+    <NavLink class="nav-link" href="starship-10">
+        <span class="bi bi-list-nested-nav-menu" aria-hidden="true"></span> Starship
+    </NavLink>
+</div>
+```
+
+When automatic model binding validation error occurs on the server, the framework returns a [default bad request response](xref:web-api/index#default-badrequest-response) with a <xref:Microsoft.AspNetCore.Mvc.ValidationProblemDetails>. The response contains more data than just the validation errors, as shown in the following example when all of the fields of the `Starfleet Starship Database` form aren't submitted and the form fails validation:
```
wadepickett:
```suggestion
When automatic model binding validation fails on the server, the framework returns a [default bad request response](xref:web-api/index#default-badrequest-response) with a <xref:Microsoft.AspNetCore.Mvc.ValidationProblemDetails>. The response contains more data than just the validation errors, as shown in the following example when all of the fields of the `Starfleet Starship Database` form aren't submitted and the form fails validation:
```
Very minor: Fixes missing article / subject-verb agreement, reads a bit better, and matches the described 400 behavior.

aspnetcore/blazor/forms/validation.md
```diff
@@ -355,128 +356,584 @@ When validation messages are set in the component, they're added to the validato
...
-* Process model validation on the server.
-* The server API includes both the built-in framework data annotations validation and custom validation logic supplied by the developer. If validation passes on the server, process the form and send back a success status code ([`200 - OK`](https://developer.mozilla.org/docs/Web/HTTP/Status/200)). If validation fails, return a failure status code ([`400 - Bad Request`](https://developer.mozilla.org/docs/Web/HTTP/Status/400)) and the field validation errors.
-* Either disable the form on success or display the errors.
+* When the form passes client validation (<xref:Microsoft.AspNetCore.Components.Forms.EditForm.OnValidSubmit> is called), send the <xref:Microsoft.AspNetCore.Components.Forms.EditContext.Model?displayProperty=nameWithType> to a backend Minimal API for remote validation.
+* Process remote model validation:
+  * Data annotations validation with built-in support for Minimal APIs.
+  * Custom validation logic.
+* Send validation errors, if any, back to the client.
+* Either disable the form on success or display the errors, allowing the user to correct any problems with the form's field values.
```
wadepickett:
```suggestion
* Either disable the form on success or display the errors so that the user can correct any problems with the form's field values.
```
Very minor: Reads a little better this way.

aspnetcore/blazor/forms/validation.md
```diff
@@ -355,128 +356,584 @@ When validation messages are set in the component, they're added to the validato
...
 ```
 
 :::moniker-end
 
-:::moniker range="< aspnetcore-8.0"
+:::moniker range=">= aspnetcore-8.0 < aspnetcore-10.0"
+
+*This section is focused on Blazor Web App scenarios, but the approach for any type of app that uses server-side validation with web API adopts the same general approach.*
+
+Remote validation is supported in addition to Blazor Web App client/server-side validation:
```
wadepickett:
```suggestion
Remote validation is supported in addition to Blazor Web App client-side and server-side validation:
```
Very minor: Was not clear if "-side" applies to both, better I think to clarify and generaly avoid slashes in prose when we can.

wadepickett (commented):
See note on using MVC that is not used, the rest are very minor suggestions.

aspnetcore/blazor/forms/validation.md
guardrex: `ValidationProblemDetails` on Line 582 requires the namespace.

https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.validationproblemdetails

I couldn't find it in any other package, and @oroztocil can probably confirm if it's available from any other source. The app could also host its own to avoid having to use that package. I appeal to @oroztocil to decide.

Of course, I'm still waiting to hear back what we're doing with this section. I like the section (if it's coded correctly for this scenario). I think it would save devs trying to implement this a lot of time and frustration. However, @oroztocil (and of course, @danroth27) are better judges of that. We might be dropping this entire section. If we keep it tho ... AFAICT ... the MVC package is needed here for the use of `ValidationProblemDetails`.

aspnetcore/blazor/forms/validation.md
guardrex: ***Yes!*** 😄 ... However, you're describing API for .NET 11, correct? Therefore, we'd still do what we're showing here for .NET 10 or earlier. I opened #37366 to trail this PR for changing this over to use the `ClientValidation` class for .NET 11 or later (and the *Validator component* section can be dropped for .NET 11 or later).

guardrex:
@wadepickett ... Thanks, and I took all of your feedback. The MVC package is required for `ValidationProblemDetails` AFAIK. @oroztocil will correct me if that API is available from a different package or if we should roll that one bit of API into the app and drop the MVC package.

@oroztocil ... I'll follow up this PR with .NET 11 work for adopting the `ClientValidation` class API, tracking the work via #37366. Are you OK with proceeding with this PR? I still like showing this setup, but I yield to you and @danroth27 to say if this is something that should be left to developers to work out on their own. Personally, I find it a bit complex ... I think this will save devs who are trying to do this some time with a basic implementation (or even just a basic learning experience). If we're keeping this, I'll need at least one GH sign-off ... you or Wade to get it over the finish line 🏎️🏁. If we're dropping the entire section, then I'll close this PR and open a new PR to shed this section.

wadepickett:
> @wadepickett ... Thanks, and I took all of your feedback. The MVC package is required for `ValidationProblemDetails` AFAIK. 

@guardrex, thanks for the update.  I will leave it to you and oroztocil from here unless you explicitly call me in.

guardrex:
He might be OOF at the moment. I'm standing by to see what we're doing to wrap this up.

aspnetcore/blazor/forms/validation.md
oroztocil: If you are referring to the [client-side validation feature](https://learn.microsoft.com/en-us/aspnet/core/blazor/forms/validation?view=aspnetcore-11.0#client-side-validation-in-static-ssr-forms) released in Preview 5, that is for static SSR, while this tutorial is for interactive mode.

That said, renaming the class to `CustomValidation` reduces the confusion.

guardrex: I updated the issue to clarify that we'd be adding remarks on the `ClientValidation` class for client-side validation under static SSR. I'm not sure where/how it will affect the current coverage, but I'll analyze it when I reach that issue.

guardrex:
@oroztocil ... Have you made a decision on if you will sign-off on this? ... or if you want to remove the entire section? This has been here since May, and I'd like to get it wrapped up either way as soon as we can.

oroztocil (approved):
@guardrex Sorry for not providing feedback here in a more timely manner. I don't want to block your changes forever.

I think the content is valuable, although I had a somewhat difficult time understanding the role and place for it. That said, I would like to do a thorough refresh of the validation docs before the 11 GA, which might include restructuring the related articles and sections. Considering that data validation features are a larger story in .NET 11 across the framework, I think it will be worth it.

Therefore, I propose that we merge your update now, and later (potentially) have a discussion about what is the best place for this content.

Thank you for your work on this!

PR #37120, opened 2026-05-08
Correct nested Map middleware example

yermekyerden:
## Summary

This PR corrects the nested `Map` middleware example.

The example currently calls `app.Run(...)` inside the nested branches. Since `app` is the root application builder, the terminal middleware is registered on the root pipeline instead of the nested branch.

The updated code uses `level2AApp.Run(...)` and `level2BApp.Run(...)`, which matches the intended nested `Map` behavior.

## Validation

Documentation-only change.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/fundamentals/middleware/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/eeaf722b3bbbd4e89a91d8def4608de07ef919ce/aspnetcore/fundamentals/middleware/index.md) | [ASP.NET Core Middleware](https://review.learn.microsoft.com/en-us/aspnet/core/fundamentals/middleware/index?branch=pr-en-us-37120) |

### DIFF
--- a/aspnetcore/fundamentals/middleware/index.md
+++ b/aspnetcore/fundamentals/middleware/index.md
@@ -228,13 +228,13 @@ var app = builder.Build();
 
 app.Map("/level1", level1App => {
     level1App.Map("/level2a", level2AApp => {
-        app.Run(async context =>
+        level2AApp.Run(async context =>
         {
             await context.Response.WriteAsync("Processing '/level1/level2a'");
         });
     });
     level1App.Map("/level2b", level2BApp => {
-        app.Run(async context =>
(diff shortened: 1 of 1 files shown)


wadepickett:
@yermekyerden,  Thank very much for identifying this issue and fixing it!  Greatly appreciated!!


The original code used app.Run(...) (root pipeline builder) instead of the branch-specific builders level2AApp.Run(...) and level2BApp.Run(...), which would cause incorrect runtime behavior inconsistent with the response table shown in the article.

wadepickett (approved):
Approved.

PR #37194, opened 2026-05-26
Added improvement in documentation for parallel streaming invocations limit

### LINKED ISSUE #37238: SignalR Hubs: Add documenation and samples on parallel streaming invocations limit
### Description

Dirived from product issue: dotnet/aspnetcore#66719

A community member has kindly volunteered to update for this issue along with a collaboration from the doc team, so I created this additional tracking issue.

- Update SignalR docs to clarify that HubOptions.MaximumParallelInvocationsPerClient does not apply to streaming invocations and add guidance for limiting per-connection streaming concurrency via a hub filter.
- Clarify MaximumParallelInvocationsPerClient behavior in configuration.md and versioned includes.
- Add a StreamConcurrencyFilter example to hubs.md
- Included guidance that streaming invocations are long-running and can run concurrently, and that hub filters are the right mechanism to enforce connection-level streaming limits

### Page URL

https://learn.microsoft.com/en-us/aspnet/core/signalr/hubs?view=aspnetcore-10.0

### Content source URL

https://github.com/dotnet/AspNetCore.Docs/blob/main/aspnetcore/signalr/hubs.md

### Document ID

fb016cdc-d281-6b5b-7466-2c718e7274b4

### Platform Id

836386f1-e0b0-4e86-1f74-5ed6ba1d9a91

### Article author

@wadepickett

### Metadata

* ID: 1387f556-b4b7-70f9-5bd0-f0bccc597689
* PlatformId: 836386f1-e0b0-4e86-1f74-5ed6ba1d9a91 
* Service: **aspnet-core**
* Sub-service: **aspnetcore-signalr**

[Related Issues](https://github.com/dotnet/AspNetCore.Docs/issues?q=is%3Aissue+is%3Aopen+fb016cdc-d281-6b5b-7466-2c718e7274b4)

---
[Associated WorkItem - 586392](https://dev.azure.com/msft-skilling/Content/_workitems/edit/586392)

### LINKED ISSUE #66719: Improve documentation on parallel streaming invocations limit
Streaming hub method invocations bypass the concurrency limit enforced by the `MaximumParallelInvocationsPerClient` setting. This is very intentional decision ([code](https://github.com/dotnet/aspnetcore/blob/main/src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs#L303-L314)), because streaming hub invocations are expected to be long-running and concurrent.

https://github.com/dotnet/AspNetCore.Docs/issues/27956#issuecomment-1362339682:
> NOTE: This concurrency limit does NOT apply to streaming invocations. The client can spawn concurrency streams without limits.

`MaximumParallelInvocationsPerClient` docs states: 
> By default a client is only allowed to invoke a single Hub method at a time. Changing this property will allow clients to invoke multiple methods at the same time before queueing.

There is no mention of streaming invocations being excluded from this limit. Update the docs and the conceptual docs. We could also demonstrate how to limit per-connection parallel streaming invocations using [hub filters](https://learn.microsoft.com/en-us/aspnet/core/signalr/hub-filters?view=aspnetcore-10.0).

As a long term, we could consider a separate `MaximumParalllelStreamingInvocations` limit.

classyk12:
Fixes https://github.com/dotnet/aspnetcore/issues/66719
Fixes #37238

1. Update SignalR docs to clarify that **HubOptions.MaximumParallelInvocationsPerClient** does not apply to streaming invocations and add guidance for limiting per-connection streaming concurrency via a hub filter.

2. Clarified **MaximumParallelInvocationsPerClient** behavior in **configuration.md** and versioned includes.

3. Added a **StreamConcurrencyFilter** example to **hubs.md**

4. Included guidance that streaming invocations are long-running and can run concurrently, and that hub filters are the right mechanism to enforce connection-level streaming limits

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/signalr/configuration.md](https://github.com/dotnet/AspNetCore.Docs/blob/35fbecbe5f74597f946a011ee3d6605c462b0753/aspnetcore/signalr/configuration.md) | [customer intent: As an ASP.NET developer, I want to configure ASP.NET Core SignalR, so I can specify my preferences for client and server options.](https://review.learn.microsoft.com/en-us/aspnet/core/signalr/configuration?branch=pr-en-us-37194) |
| [aspnetcore/signalr/hubs.md](https://github.com/dotnet/AspNetCore.Docs/blob/35fbecbe5f74597f946a011ee3d6605c462b0753/aspnetcore/signalr/hubs.md) | [Use hubs in ASP.NET Core SignalR](https://review.learn.microsoft.com/en-us/aspnet/core/signalr/hubs?branch=pr-en-us-37194) |


guardrex:
Hello @classyk12 ... That's not a correct issue number for this repo. We do need an issue for a contribution such as this. Do you have the correct issue number? If not, can you open an issue for this? 👇 

Use [THIS LINK](https://github.com/dotnet/aspnetcore.docs/issues/new?template=customer-feedback.yml&pageUrl=https%3A%2F%2Flearn.microsoft.com%2Fen-us%2Faspnet%2Fcore%2Fsignalr%2Fconfiguration%3Fview%3Daspnetcore-10.0%26tabs%3Ddotnet&pageQueryParams=%3Fview%3Daspnetcore-10.0%26tabs%3Ddotnet&contentSourceUrl=https%3A%2F%2Fgithub.com%2Fdotnet%2FAspNetCore.Docs%2Fblob%2Fmain%2Faspnetcore%2Fsignalr%2Fconfiguration.md&documentVersionIndependentId=66d252c3-6300-7e28-9aa0-ef2ff66e7a76&platformId=5e34ac2d-cc62-b969-c4ab-a1681eb6b5bd&feedback=%0A%0A%5BEnter+feedback+here%5D%0A&author=%40wadepickett&metadata=*+ID%3A+de8cdc99-48bb-3266-05bb-eda98f59f2e2%0A*+PlatformId%3A+5e34ac2d-cc62-b969-c4ab-a1681eb6b5bd+%0A*+Service%3A+**aspnet-core**%0A*+Sub-service%3A+**aspnetcore-signalr**&labels=Source+-+Docs.ms%2C%3Awatch%3A+Not+Triaged) to open an issue for that article.

After the issue is opened, update your OP☝️with the correct issue for this repo.

aspnetcore/signalr/hubs.md
```diff
@@ -254,13 +249,130 @@ public class ChatHub : Hub
...
+            1,
+            (_, current) => current + 1);
+
+        if (activeStreams > _maxConcurrentStreams)
+        {
+            Decrement(connectionId);
+            throw new HubException($"The connection is limited to {_maxConcurrentStreams} concurrent streaming invocations.");
+        }
+
+        try
```
wadepickett, line 320: I had to look at this a few times, but I think the limit would never be honored.

The count is decremented the moment the stream starts delivering, not when the stream actually finishes.
Every stream increments the counter, then decrements it almost instantly as soon as the stream object is returned. So by the time a second stream request comes in, the first one has already decremented. The counter is back to zero even though the first stream is still actively running.

So it seems for example, we could start 100 concurrent streams even though there is a limit of 1, because the counter never stays elevated long enough to block as intended, so no actual concurrency control.

@BrennanConroy, what would be ther correct pattern to suggest and document here?  Also, should this be using `InvokeStreamAsync` as opposed to `InvokeMethodAsync`?

aspnetcore/signalr/hubs.md
```diff
@@ -254,13 +249,130 @@ public class ChatHub : Hub
...
+> This feature makes use of <xref:Microsoft.Extensions.DependencyInjection.IServiceProviderIsService>, which is optionally implemented by DI implementations. If the app's DI container doesn't support this feature, injecting services into hub methods isn't supported.
 
-### Keyed services support in dependency injection
+## Limit per-connection streaming invocations
 
-The *keyed services* mechanism allows you to register and retrieve dependency injection services by using keys. A service is associated with a key by calling the <xref:Microsoft.Extensions.DependencyInjection.ServiceCollectionServiceExtensions.AddKeyedSingleton%2A> method to register it. As an alternative, you can call the `AddKeyedScoped` or `AddKeyedTransient` method.
+`HubOptions.MaximumParallelInvocationsPerClient` controls non-streaming hub invocations only. It does not apply to streaming hub invocations. Streaming invocations are expected to be long-running and can run concurrently. To enforce a per-connection limit for streaming invocations, use a hub filter to track active stream-returning hub methods.
+
+```csharp
+using System.Collections.Concurrent;
```
wadepickett, line 274: I think we also need:
`using System.Threading.Channels;`

aspnetcore/signalr/hubs.md
```diff
@@ -254,13 +249,130 @@ public class ChatHub : Hub
 ```
 
 > [!NOTE]
-> This feature makes use of <xref:Microsoft.Extensions.DependencyInjection.IServiceProviderIsService>, which is optionally implemented in dependency injection configurations. If the application dependency injection container doesn't support this feature, injecting services into hub methods isn't supported.
+> This feature makes use of <xref:Microsoft.Extensions.DependencyInjection.IServiceProviderIsService>, which is optionally implemented by DI implementations. If the app's DI container doesn't support this feature, injecting services into hub methods isn't supported.
 
-### Keyed services support in dependency injection
+## Limit per-connection streaming invocations
```
wadepickett, line 267: This new section would need to move to after the sub section _Keyed Services support in Dependancy Injection_.
"Keyed services support" is conceptually a subtopic of "Inject services into a hub"

So the organizaiton would like like this:

`## Inject services into a hub`
  (existing content)
`  ### Keyed services support in Dependency Injection`
    (existing content)

`## Limit per-connection streaming invocations`
  (new content)

`## Handle events for a connection`

BrennanConroy:
They're linking to https://github.com/dotnet/aspnetcore/issues/66719

classyk12:
yes @BrennanConroy mentioned, this issue on the aspnetcore repo: https://github.com/dotnet/aspnetcore/issues/66719 has to be fixed there and then the conceptual docs updated as well. 

Do let me know if this stays or you need me to open a new issue linking back to the aspnetcore issue.

wadepickett:
> yes @BrennanConroy mentioned, this issue on the aspnetcore repo: [dotnet/aspnetcore#66719](https://github.com/dotnet/aspnetcore/issues/66719) has to be fixed there and then the conceptual docs updated as well.
> 
> Do let me know if this stays or you need me to open a new issue linking back to the aspnetcore issue.

That tracking issue is good.  No need for any other context.  Thanks for the clarificaiton and thanks very much for taking on this issue!  Greatly appreciated.

aspnetcore/signalr/hubs.md
wadepickett: @BrennanConroy, could you weigh in on the correct pattern here for above?  It is not clear to me what it should be.

BrennanConroy: My initial thought would have been code in the Hub itself to handle this: (pseudo code) e.g.
```
class MyHub : Hub
{
    public IAsyncEnumerable<string> Stream()
    {
        return WithLimit(GetStream());
    }

    private async IAsyncEnumerable<string> WithLimit(IAsyncEnumerable<string> originalStream)
    {
        // initial limit check to throw before wrapping if at limit
        if (...) throw;

        IncrementStreamCount();

        try
        {
            await foreach (var item in originalStream)
            {
                yield return item;
            }
        }
        finally
        {
            DecrementStreamCount();
        }
    }
}
```

The hub filter is an interesting idea, unfortunately you would need to use some reflection to achieve the same behavior:

<details>
<summary>Reflection IHubFilter Pseudo Code</summary>

```
public sealed class StreamConcurrencyFilter : IHubFilter
{
    private readonly ConcurrentDictionary<string, int> _activeStreams = new();
    private readonly int _maxConcurrentStreams;

    public StreamConcurrencyFilter(int maxConcurrentStreams = 1)
    {
        if (maxConcurrentStreams < 1)
        {
            throw new ArgumentOutOfRangeException(nameof(maxConcurrentStreams));
        }

        _maxConcurrentStreams = maxConcurrentStreams;
    }

    public async ValueTask<object?> InvokeMethodAsync(
        HubInvocationContext invocationContext,
        Func<HubInvocationContext, ValueTask<object?>> next)
    {
        var streamItemType = GetStreamItemType(invocationContext.HubMethod.ReturnType);
        if (streamItemType is null)
        {
            return await next(invocationContext);
        }

        var connectionId = invocationContext.Context.ConnectionId;

        // Reserve a slot before invoking the method. If the cap is exceeded,
        // release the slot and fail the invocation so no stream is started.
        if (_activeStreams.AddOrUpdate(connectionId, 1, (_, current) => current + 1)
            > _maxConcurrentStreams)
        {
            Decrement(connectionId);
            throw new HubException(
                $"The connection is limited to {_maxConcurrentStreams} concurrent streaming invocations.");
        }

        object? result;
        try
        {
            result = await next(invocationContext);
        }
        catch
        {
            Decrement(connectionId);
            throw;
        }

        if (result is null)
        {
            Decrement(connectionId);
            return null;
        }

        // Wrap the result so the slot is released only after the client finishes
        // consuming the stream (or the connection is aborted).
        return WrapMethod
            .MakeGenericMethod(streamItemType)
            .Invoke(this, new[] { result, connectionId });
    }

    private static readonly MethodInfo WrapMethod = typeof(StreamConcurrencyFilter)
        .GetMethod(nameof(Wrap), BindingFlags.Instance | BindingFlags.NonPublic)!;

    private object Wrap<T>(object result, string connectionId)
    {
        return result switch
        {
            IAsyncEnumerable<T> asyncEnumerable => TrackAsyncEnumerable(asyncEnumerable, connectionId),
            ChannelReader<T> channelReader => TrackChannelReader(channelReader, connectionId),
            _ => ReleaseAndReturn(result, connectionId),
        };
    }

    private async IAsyncEnumerable<T> TrackAsyncEnumerable<T>(
        IAsyncEnumerable<T> source,
        string connectionId,
        [EnumeratorCancellation] CancellationToken cancellationToken = default)
    {
        try
        {
            await foreach (var item in source.WithCancellation(cancellationToken))
            {
                yield return item;
            }
        }
        finally
        {
            Decrement(connectionId);
        }
    }

    private ChannelReader<T> TrackChannelReader<T>(ChannelReader<T> source, string connectionId)
    {
        var channel = Channel.CreateUnbounded<T>(new UnboundedChannelOptions
        {
            SingleReader = true,
            SingleWriter = true,
        });

        _ = Task.Run(async () =>
        {
            try
            {
                await foreach (var item in source.ReadAllAsync())
                {
                    await channel.Writer.WriteAsync(item);
                }

                channel.Writer.TryComplete();
            }
            catch (Exception ex)
            {
                channel.Writer.TryComplete(ex);
            }
            finally
            {
                Decrement(connectionId);
            }
        });

        return channel.Reader;
    }

    private object ReleaseAndReturn(object result, string connectionId)
    {
        Decrement(connectionId);
        return result;
    }

    private static Type? GetStreamItemType(Type returnType)
    {
        if (returnType is null)
        {
            return null;
        }

        if (returnType.IsGenericType)
        {
            var def = returnType.GetGenericTypeDefinition();
            if (def == typeof(IAsyncEnumerable<>) || def == typeof(ChannelReader<>))
            {
                return returnType.GetGenericArguments()[0];
            }

            if (def == typeof(Task<>) || def == typeof(ValueTask<>))
            {
                return GetStreamItemType(returnType.GetGenericArguments()[0]);
            }
        }

        return null;
    }

    private void Decrement(string connectionId)
    {
        while (_activeStreams.TryGetValue(connectionId, out var current))
        {
            if (current <= 1)
            {
                if (_activeStreams.TryRemove(new KeyValuePair<string, int>(connectionId, current)))
                {
                    return;
                }
            }
            else if (_activeStreams.TryUpdate(connectionId, current - 1, current))
            {
                return;
            }
        }
    }
}
```

</details>

wadepickett:
@classyk12, do you want to keep going from here?  See Brennon's recomondaton above.  Or do you need help?  Your offer to do the PR and the work you put into this so far is very appreciated!   If you need some help to get over the finish line let me know and we can jump in.  Thanks again for the effort here.

classyk12:
> @classyk12, do you want to keep going from here? See Brennon's recomondaton above. Or do you need help? Your offer to do the PR and the work you put into this so far is very appreciated! If you need some help to get over the finish line let me know and we can jump in. Thanks again for the effort here.

I will sort this out today. I have been away from town running a Marathon. I will review the suggestions by @BrennanConroy

classyk12:
@wadepickett  The idea proposed by @BrennanConroy  is an interesting one. I think the question would be if we should document this in the **hub** or stick with reflection in the **hubfilter**. We can collectively agree on an approach and we run with it.

wadepickett:
The hub approach

> @wadepickett The idea proposed by @BrennanConroy is an interesting one. I think the question would be if we should document this in the **hub** or stick with reflection in the **hubfilter**. We can collectively agree on an approach and we run with it.

Follow the less complex hub solution BrennanConroy suggested.  

From here that might look like this:

1. Document that `MaximumParallelInvocationsPerClient` doesn't apply to streaming hub invocations.
2. Provide a Hub-based workaround pattern (no HubFilter/reflection approach).

---

#### File 1: `aspnetcore/signalr/hubs.md`

**Restore heading**
- **Line 13** — Restore `#` heading marker (`" Use hubs..."` → `"# Use hubs..."`)

**Restore deleted content**
- **Lines 259–265** (`### Keyed services support in dependency injection`) — Revert to original wording and xref links that the PR mangled/deleted.

**New section**
Insert after line 265 (end of keyed services content), before `## Handle events for a connection` at line 267:

Add `## Limit per-connection streaming invocations` containing:
- Explanatory text: `MaximumParallelInvocationsPerClient` controls non-streaming invocations only and does not apply to streaming hub invocations.
- Hub-based `WithLimit` wrapper pattern — wrapping `IAsyncEnumerable<T>` with `try/finally` + `yield return` to track active streams and enforce a per-connection limit. Using BrennanConroy's code example above as a guideline.
- Registration/usage example showing how to call `WithLimit` from a streaming hub method.

---

#### File 2: `aspnetcore/signalr/configuration.md`

Section: `## Configure server options`, options table
- **Line 80** (`MaximumParallelInvocationsPerClient` row) — Append to description: note that this limit doesn't apply to streaming hub invocations, with a cross-link to `xref:signalr/hubs#limit-per-connection-streaming-invocations`.

---

#### File 3: `aspnetcore/signalr/configuration/includes/configuration7.md`

Section: `## Configure server options`, options table
- **Line 60** (`MaximumParallelInvocationsPerClient` row) — Append to description: note that this limit doesn't apply to streaming hub invocations. No cross-link (the workaround section in `hubs.md` is scoped to `>= aspnetcore-8.0` and wouldn't resolve for 7.0 readers).

---

## Summary

| File | Location | Change |
|---|---|---|
| `aspnetcore/signalr/hubs.md` | Line 13 | **Fix**: Restore `#` heading marker |
| `aspnetcore/signalr/hubs.md` | Lines 259–265 | **Revert**: Restore original keyed services content |
| `aspnetcore/signalr/hubs.md` | Between line 265 and line 267 | **Add**: `## Limit per-connection streaming invocations` with Hub-based `WithLimit` pattern |
| `aspnetcore/signalr/configuration.md` | Line 80 | **Update**: Append streaming limitation note + cross-link |
| `aspnetcore/signalr/configuration/includes/configuration7.md` | Line 60 | **Update**: Append streaming limitation note only (no cross-link) |

> **Note:** Only `>= aspnetcore-7.0` moniker docs are updated. Older configuration includes (`configuration5.md`, `configuration6.md`, `configuration3.md`, `configuration3.1.md`, `configuration2.x.md`) are left unchanged.

classyk12:
@wadepickett have a look at the changes and for the code implementation as well using @BrennanConroy  approach. It is weird also that the pipeline is failing with this error on an untouched file

image


could you please advice? Cheers

wadepickett:
> @wadepickett have a look at the changes and for the code implementation as well using @BrennanConroy approach. It is weird also that the pipeline is failing with this error on an untouched file


You don't have to worry about the error.  It has nothing to do with your PR.  The "Look for build warnings" check just fails if any warning exists anywhere in the docset, not just your PR.

That broken snippet references to Contoso.API/appsettings.json and Contoso.Security.API/appsettings.json in the external samples repo have nothing to do with the SignalR streaming documentation being added here.

I will go check on getting it fixed.  Looks like someone moved some sample files around or renamed them on the samples repo ( I am going to be embarassed if it was me ).

wadepickett:
Update on build error:  It is a bit odd since the files appear to be there and named as expected.  It could be a transient build issue from a sync that is off between repos.  

Possible solution to try: 
You can try clicking "Close with comment" and then click the reopen button for this PR and that will force a new clean buld.  I will let you do it so I am not messing with it while you are potentioally commiting something at the same time.

guardrex:
@wadepickett ... I was in the process of moving a sample and broke a couple of PRs (I'm working on *Blazorfication* of the policies article), so I temporarily put the sample back. If this PR is rebuilt, it should pick up the presence of the sample. *Sorry for the trouble!!!!!* 🙈

wadepickett:
Closing and re-opening to start new build.

classyk12:
Thanks @wadepickett  everything looks good now

wadepickett:
I see the following is in, thanks:

- Adopted the hub-based WithLimit approach per BrennanConroy's recommendation
- Placed the new ## Limit per-connection streaming invocations section after keyed services and before "Handle events for a connection"
- Updated configuration.md with the streaming note + cross-link
Restored the # heading on line 13

There looks like there are a few items to address.  I will point them out inline.

aspnetcore/signalr/hubs.md
```diff
@@ -258,7 +258,99 @@ public class ChatHub : Hub
...
+The key point is that `WithLimit` wraps the original `IAsyncEnumerable<T>` and holds the counter elevated for the *full lifetime of the stream*, not just until the first item is yielded.
+The `finally` block runs only when the client finishes consuming the stream, cancels it, or the connection drops.
+
+> [!NOTE]
+> The `_activeStreams` dictionary is `static` so it is shared across all hub instances for
+a given connection. If you prefer DI-managed state, register a singleton service that owns
+> the dictionary and inject it into the hub constructor.
+
+
+### Keyed services support in Dependency Injection
```
wadepickett: This "Keyed services support.." section now appears twice.  This one from line 352 - 358 needs to be removed so the duplicate no longer exists, and to get rid of an extra line space.

aspnetcore/signalr/hubs.md
```diff
@@ -258,7 +258,99 @@ public class ChatHub : Hub
 
 ### Keyed services support in dependency injection
 
-The *keyed services* mechanism allows you to register and retrieve dependency injection services by using keys. A service is associated with a key by calling the <xref:Microsoft.Extensions.DependencyInjection.ServiceCollectionServiceExtensions.AddKeyedSingleton%2A> method to register it. As an alternative, you can call the `AddKeyedScoped` or `AddKeyedTransient` method.
+The keyed services mechanism allows you to register and retrieve dependency injection services by using keys. A service is associated with a key by calling the  xref:Microsoft.Extensions.DependencyInjection.ServiceCollectionServiceExtensions.AddKeyedSingleton%2A method to register it. As an alternative, you can call the `AddKeyedScoped` or `AddKeyedTransient` method.
```
wadepickett:
```suggestion
The keyed services mechanism allows you to register and retrieve dependency injection services by using keys. A service is associated with a key by calling the  <xref:Microsoft.Extensions.DependencyInjection.ServiceCollectionServiceExtensions.AddKeyedSingleton%2A> method to register it. As an alternative, you can call the `AddKeyedScoped` or `AddKeyedTransient` method.
```
Replaced missing angle brackets.

aspnetcore/signalr/hubs.md
```diff
@@ -258,7 +258,99 @@ public class ChatHub : Hub
...
+
+You access a registered service by specifying the key with the [FromKeyedServices] attribute. The following code shows how to use keyed services:
```
wadepickett:
```suggestion
You access a registered service by specifying the key with the [[FromKeyedServices](xref:Microsoft.Extensions.DependencyInjection.FromKeyedServicesAttribute)] attribute. The following code shows how to use keyed services:
```
Making this a link.

aspnetcore/signalr/hubs.md
```diff
@@ -258,7 +258,99 @@ public class ChatHub : Hub
...
+To enforce a per-connection limit on streaming invocations, wrap the stream inside the hub method itself using a private helper that increments a counter before yielding items and decrements it in a `finally` block:
+
+```csharp
+using System.Collections.Concurrent;
+using System.Runtime.CompilerServices;
+
+public class StreamingHub : Hub
+{
+    // Track the number of active streams per connection.
+    private static readonly ConcurrentDictionary _activeStreams = new();
```
wadepickett:
```suggestion
    private static readonly ConcurrentDictionary<string, int> _activeStreams = new();
```

I think everything in angle brackets may have been stripped accidently during editing?  The generic type parameters on IAsyncEnumerable, ConcurrentDictionary, etc. got stripped out. Compare your code block to the [official sample](https://github.com/dotnet/aspnetcore/blob/7f0d5e98459ff1e14bf1859e4409abe7ccd9d5a6/src/SignalR/samples/SignalRSamples/Hubs/Streaming.cs#L13). For example IAsyncEnumerable needs to be IAsyncEnumerable<int>, and ConcurrentDictionary needs to be ConcurrentDictionary<string, int>. Without these the sample won't compile.

aspnetcore/signalr/hubs.md
```diff
@@ -258,7 +258,99 @@ public class ChatHub : Hub
...
+    private const int MaxConcurrentStreams = 2;
+
+    public IAsyncEnumerable Counter(
```
wadepickett:
```suggestion
    public IAsyncEnumerable<int> Counter(
```
Bracket items seem to have been removed.

aspnetcore/signalr/hubs.md
```diff
@@ -258,7 +258,99 @@ public class ChatHub : Hub
...
+        int count,
+        int delay,
+        CancellationToken cancellationToken)
+    {
+        return WithLimit(GetCounter(count, delay, cancellationToken));
+    }
+
+    private async IAsyncEnumerable GetCounter(
```
wadepickett:
```suggestion
    private async IAsyncEnumerable<int> GetCounter(
```
Put missing generic type params back in.

aspnetcore/signalr/hubs.md
```diff
@@ -258,7 +258,99 @@ public class ChatHub : Hub
...
+    {
+        for (var i = 0; i < count; i++)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            yield return i;
+            await Task.Delay(delay, cancellationToken);
+        }
+    }
+
+    private async IAsyncEnumerable WithLimit(
```
wadepickett:
```suggestion
    private async IAsyncEnumerable<int> WithLimit(
```
Putting the missing generic type param back in.

aspnetcore/signalr/hubs.md
```diff
@@ -258,7 +258,99 @@ public class ChatHub : Hub
...
+        IAsyncEnumerable stream,
```
wadepickett:
```suggestion
        IAsyncEnumerable<int> stream,
```
I put the missing generic type params back in.

wadepickett (changes requested):
Almost done.  Found a few lines of code that were missing generic type params, I think because they were stripped somehow while editing.  I made suggestions inline.

classyk12:
Thank you for this @wadepickett. Changes have been applied.

aspnetcore/signalr/hubs.md
```diff
@@ -258,12 +258,97 @@ public class ChatHub : Hub
...
+    }
+}
+```
+
+The key point is that `WithLimit` wraps the original `IAsyncEnumerable<T>` and holds the counter elevated for the *full lifetime of the stream*, not just until the first item is yielded.
+The `finally` block runs only when the client finishes consuming the stream, cancels it, or the connection drops.
+
+> [!NOTE]
+> The `_activeStreams` dictionary is `static` so it is shared across all hub instances for
+a given connection. If you prefer DI-managed state, register a singleton service that owns
```
BrennanConroy:
```suggestion
> The `_activeStreams` dictionary is `static` so it is shared across all hub instances. If you prefer DI-managed state, register a singleton service that owns
```

aspnetcore/signalr/hubs.md
```diff
@@ -258,12 +258,97 @@ public class ChatHub : Hub
...
+    {
+        for (var i = 0; i < count; i++)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            yield return i;
+            await Task.Delay(delay, cancellationToken);
+        }
+    }
+
+    private async IAsyncEnumerable<int> WithLimit(
```
BrennanConroy: Should this use `IAsyncEnumerable<T>`?

aspnetcore/signalr/hubs.md
```diff
@@ -258,12 +258,97 @@ public class ChatHub : Hub
...
+            {
+                yield return item;
+            }
+        }
+        finally
+        {
+            _activeStreams.AddOrUpdate(
+                connectionId,
+                addValue: 0,
+                updateValueFactory: (_, count) => Math.Max(0, count - 1));
```
BrennanConroy: I think this also needs to remove itself from the dictionary to avoid a memory leak
```
_activeStreams.TryRemove(new (connectionId, 0));
```

Above as well in the error path.

wadepickett:
Closing and opening to force docs rebuild.  The doc build system had trouble yesterday.

aspnetcore/signalr/hubs.md
classyk12: This came from the suggestion by @wadepickett. Whats your thought on this @BrennanConroy. Do we make it generic of type T?

aspnetcore/signalr/hubs.md
classyk12: Updated now

aspnetcore/signalr/hubs.md
classyk12: I have fixed this in the code snippet. Made the decrement logic a private function that handles decrement/removal from dictionary depending on how many connections are left.

wadepickett:
Successful doc build.  Reviewing based on Brennan's last round of feedback...

wadepickett:
Here is a summary of where we are at currently:

| Reviewer | Item | Status |
|---|---|---|
| BrennanConroy | Memory leak — remove from dictionary in `Decrement` ([r3391760492](https://github.com/dotnet/AspNetCore.Docs/pull/37194#discussion_r3391760492)) | ✅Addressed in `550baf9` |
| BrennanConroy | Add `> ` prefix on wrapped second line of NOTE ([r3391620999](https://github.com/dotnet/AspNetCore.Docs/pull/37194#discussion_r3391620999)) | Not applied yet |
| BrennanConroy | Make `WithLimit` generic `IAsyncEnumerable<T>` ([r3391623436](https://github.com/dotnet/AspNetCore.Docs/pull/37194#discussion_r3391623436)) | In discussion: Question to Brennan ([r3401472711](https://github.com/dotnet/AspNetCore.Docs/pull/37194#discussion_r3401472711)) just a few hours ago|
| wadepickett (last round) | Stray unmatched `` ` `` before `<xref:…>` on `aspnetcore/signalr/hubs.md` line 269 | Still present |

classyk12:
@wadepickett For item2: I am not sure of what the issue may be but this is the current state of the line:

image

Item 4 has been sorted now

aspnetcore/signalr/configuration/includes/configuration7.md
```diff
@@ -57,7 +57,7 @@ The following table describes options for configuring SignalR hubs:
 | `EnableDetailedErrors` | `false` | If `true`, detailed exception messages are returned to clients when an exception is thrown in a Hub method. The default is `false` because these exception messages can contain sensitive information. |
 | `StreamBufferCapacity` | `10` | The maximum number of items that can be buffered for client upload streams. If this limit is reached, the processing of invocations is blocked until the server processes stream items.|
 | `MaximumReceiveMessageSize` | 32 KB | Maximum size of a single incoming hub message. Increasing the value may increase the risk of [Denial of service (DoS) attacks](https://developer.mozilla.org/docs/Glossary/DOS_attack). |
-| `MaximumParallelInvocationsPerClient` | 1 | The maximum number of hub methods that each client can call in parallel before queueing. |
+| `MaximumParallelInvocationsPerClient` | 1 | The maximum number of hub methods that each client can call in parallel before queueing. This limit does not apply to streaming hub invocations. Note that this limit doesn't apply to streaming hub invocations.|
```
wadepickett:
```suggestion
| `MaximumParallelInvocationsPerClient` | 1 | The maximum number of hub methods that each client can call in parallel before queueing. This limit does not apply to streaming hub invocations.|
```
removed duplicate sentance.

aspnetcore/signalr/hubs.md
wadepickett: @BrennanConroy,  I think this is the very last item/question to resolve.

wadepickett:
@BrennanConroy,  I think this item is the very last one to resolve and then we can publish:
https://github.com/dotnet/AspNetCore.Docs/pull/37194#discussion_r3401472711

aspnetcore/signalr/configuration/includes/configuration6.md
```diff
@@ -57,7 +57,10 @@ The following table describes options for configuring SignalR hubs:
 | `EnableDetailedErrors` | `false` | If `true`, detailed exception messages are returned to clients when an exception is thrown in a Hub method. The default is `false` because these exception messages can contain sensitive information. |
 | `StreamBufferCapacity` | `10` | The maximum number of items that can be buffered for client upload streams. If this limit is reached, the processing of invocations is blocked until the server processes stream items.|
 | `MaximumReceiveMessageSize` | 32 KB | Maximum size of a single incoming hub message. Increasing the value may increase the risk of [Denial of service (DoS) attacks](https://developer.mozilla.org/docs/Glossary/DOS_attack). |
-| `MaximumParallelInvocationsPerClient` | 1 | The maximum number of hub methods that each client can call in parallel before queueing. |
+| `MaximumParallelInvocationsPerClient` | 1 | The maximum number of hub methods that each client can call in parallel before queueing. This limit does not apply to streaming hub invocations. |
```
BrennanConroy: nit: 6.0 isn't supported

aspnetcore/signalr/configuration/includes/configuration5.md
```diff
@@ -57,7 +57,10 @@ The following table describes options for configuring SignalR hubs:
 | `EnableDetailedErrors` | `false` | If `true`, detailed exception messages are returned to clients when an exception is thrown in a Hub method. The default is `false` because these exception messages can contain sensitive information. |
 | `StreamBufferCapacity` | `10` | The maximum number of items that can be buffered for client upload streams. If this limit is reached, the processing of invocations is blocked until the server processes stream items.|
 | `MaximumReceiveMessageSize` | 32 KB | Maximum size of a single incoming hub message. Increasing the value may increase the risk of [Denial of service (DoS) attacks](https://developer.mozilla.org/docs/Glossary/DOS_attack). |
-| `MaximumParallelInvocationsPerClient` | 1 | The maximum number of hub methods that each client can call in parallel before queueing. |
+| `MaximumParallelInvocationsPerClient` | 1 | The maximum number of hub methods that each client can call in parallel before queueing. This limit does not apply to streaming hub invocations. |
```
BrennanConroy: nit: 5.0 isn't supported

aspnetcore/signalr/configuration/includes/configuration7.md
```diff
@@ -57,7 +57,7 @@ The following table describes options for configuring SignalR hubs:
 | `EnableDetailedErrors` | `false` | If `true`, detailed exception messages are returned to clients when an exception is thrown in a Hub method. The default is `false` because these exception messages can contain sensitive information. |
 | `StreamBufferCapacity` | `10` | The maximum number of items that can be buffered for client upload streams. If this limit is reached, the processing of invocations is blocked until the server processes stream items.|
 | `MaximumReceiveMessageSize` | 32 KB | Maximum size of a single incoming hub message. Increasing the value may increase the risk of [Denial of service (DoS) attacks](https://developer.mozilla.org/docs/Glossary/DOS_attack). |
-| `MaximumParallelInvocationsPerClient` | 1 | The maximum number of hub methods that each client can call in parallel before queueing. |
+| `MaximumParallelInvocationsPerClient` | 1 | The maximum number of hub methods that each client can call in parallel before queueing. This limit does not apply to streaming hub invocations.|
```
BrennanConroy, line 60: nit: 7.0 isn't supported

aspnetcore/signalr/hubs.md
BrennanConroy: I think we should make it generic. Also pass in the ConnectionId as an argument to make it more obvious how someone could extract this method into a service or utility class outside of the Hub.

aspnetcore/signalr/hubs.md
```diff
@@ -258,12 +258,105 @@ public class ChatHub : Hub
...
+            else if (_activeStreams.TryUpdate(connectionId, current - 1, current))
+            {
+                return;
+            }
+        }
+    }
+}
+```
+
+The key point is that `WithLimit` wraps the original `IAsyncEnumerable<T>` and holds the counter elevated for the *full lifetime of the stream*, not just until the first item is yielded.
```
BrennanConroy, line 353: We should also mention that `ChannelReader<T>` could have a similar helper method applied to it (and should use the same dictionary).

BrennanConroy (approved):

wadepickett:
@classyk12,  no rush here, but if this PR turned out to take more time than you could researve for it, I can try to take it from here if you like.  If you want to take it to the finish line yourself as you have more time, great, no worries, no rush.

classyk12:
> @classyk12, no rush here, but if this PR turned out to take more time than you could researve for it, I can try to take it from here if you like. If you want to take it to the finish line yourself as you have more time, great, no worries, no rush.

I will sort the comments out and update the PR in a bit

classyk12:
@wadepickett I have updated the PR now. All the comments by @BrennanConroy have been addressed in the update.

wadepickett (approved):
Approved.  Thanks @classyk12!!!

PR #37198, opened 2026-05-28
Enhance sign-out guidance

### LINKED ISSUE #36662: Enhance documentation for `RevalidatingServerAuthenticationStateProvider` in Blazor Apps
I know that you guys planning on improving the authentification docs for Blazor and I think I've found another bit which needs improvement.

I want to check if my current validation state is still valid and I came accross this docs here:

https://learn.microsoft.com/en-us/aspnet/core/blazor/security/?view=aspnetcore-10.0&tabs=visual-studio#additional-security-abstractions

So I implemented my custom `AuthenticationStateProvider` and inherited from `RevalidatingServerAuthenticationStateProvider` because I'm using a Blazor Web App with global Server interactivity. Here is my current implementation:

```cs
public sealed class HSOnlineAuthenticationStateProvider : RevalidatingServerAuthenticationStateProvider
{
    private readonly IDbConnectionFactory _dbFactory;
    private readonly UserRepository _userRepository;

    public HSOnlineAuthenticationStateProvider(IDbConnectionFactory dbFactory, UserRepository userRepository, ILoggerFactory loggerFactory) : base(loggerFactory)
    {
        _dbFactory = dbFactory;
        _userRepository = userRepository;
    }

    protected override TimeSpan RevalidationInterval => TimeSpan.FromSeconds(5);

    protected override async Task<bool> ValidateAuthenticationStateAsync(AuthenticationState authenticationState, CancellationToken cancellationToken)
    {
        Console.WriteLine("Authentification validation");

        Claim? claim = authenticationState.User.FindFirst("userId");

        if (claim is null)
        {
            // TODO: LOGOUT
            return false;
        }

        var userId = Convert.ToInt32(claim.Value, CultureInfo.InvariantCulture);
        using var connection = await _dbFactory.CreateConnectionAsync(cancellationToken);
        User? result = await _userRepository.GetAsync(userId, connection, cancellationToken: cancellationToken);

        if (result is null)
        {
            // User has been deleted from the database
            // TODO: LOGOUT
            return false;
        }

        if (!result.IsActive)
        {
            // User has been deactivated by an admin
            // TODO: LOGOUT
            return false;
        }

        return true;
    }

    // TODO: CHECK STATE ON FIRST PAGE LOAD IN ORDER TO NOT HAVE TO WAIT FOR RevalidationInterval

}
```

I login now into my app and set my user to disabled via SQL. After 10 seconds I can indeed see that I'm being redirected to the Account/Login page but when I go back to my auth required page then I see that I'm still logged in and after 10 seconds I get redirected again.

I'm using a cookie based authentifcation like this:
```cs
builder.Services.AddAuthentication(CookieAuthenticationDefaults.AuthenticationScheme)
                .AddCookie(CookieAuthenticationDefaults.AuthenticationScheme);

builder.Services.AddScoped<AuthenticationStateProvider, HSOnlineAuthenticationStateProvider>();
builder.Services.AddCascadingAuthenticationState();

/* .... */
app.UseAuthentication();
app.UseAuthorization();

/* ... */
var accountGroup = app.MapGroup("/Account");

accountGroup.MapGet("/Logout", async (HttpContext context) =>
{
    await context.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
    return Results.LocalRedirect("/Account/Login");
});
```

My main question know is how can I handle the `TODOs` from my `HSOnlineAuthenticationStateProvider`. Should the Logout be handled by the login page or is there any way todo this via the `HSOnlineAuthenticationStateProvider`. And how can I check if the authentification is still valid when the user first loads the page so I don't have to wait for my TimeSpan to kick in?

And lastly if it's valid how can I reseed the claims in case something has been changed within my user e.g. the roles without the need to logout and log in again?

CC: @guardrex @danroth27 @javiercn

guardrex:
Fixes #36662

Thanks, @MarvinKlein1508! 🚀 ... If we can get this right, we'll have closed an important gap in our coverage.

Stephen ... I may have butchered a little bit of your suggested direction 🙈, but I wanted to really flesh this one out in complete detail. The outline for this guidance explains the common `RevalidatingServerAuthenticationStateProvider.RevalidationInterval` setting in the main section. Then, there are a pair of subsections ...

* ASP.NET Core Identity subsection: Focuses on setting `SecurityStampValidatorOptions.ValidationInterval`.
* Cooke-based, non-Identity subsection: Focuses on the two approaches that you described. However, it does so with less code for the second approach, only describing the non-interval approach in a paragraph with API cross-links ... AND ... I'm not 100% sure that the code I'm showing for the first approach is exactly right. *We'll see on review where I've gone off the rails!* 😄

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [aspnetcore/blazor/security/index.md](https://github.com/dotnet/AspNetCore.Docs/blob/2fabf7bf6f3073bb432f49ca2c150a1bc0ea6561/aspnetcore/blazor/security/index.md) | [aspnetcore/blazor/security/index](https://review.learn.microsoft.com/en-us/aspnet/core/blazor/security/index?branch=pr-en-us-37198) |

### DIFF
--- a/aspnetcore/blazor/security/index.md
+++ b/aspnetcore/blazor/security/index.md
@@ -559,15 +559,12 @@ Implementing <xref:Microsoft.AspNetCore.Components.Server.RevalidatingServerAuth
 > [!NOTE]
 > Each browser tab requires a separate circuit, so additional tabs opened by a user don't observe an authentication state change until their circuits revalidate. However, the approach in this section sets the revalidation interval for all of the tabs opened by a user.
 
-To control the revalidation interval of the underlying authentication cookie in apps that adopt ASP.NET Core Identity, see the following [Sign out for ASP.NET Core Identity](#sign-out-for-aspnet-core-identity) subsection for details. For apps that adopt cookie-based authentication without Identity, see the following [Sign out for cookie-based authentication](#sign-out-for-cookie-based-authentication) subsection.
+To control the revalidation interval in apps that adopt ASP.NET Core Identity with cookie authentication, see the following [Sign out for ASP.NET Core Identity](#sign-out-for-aspnet-core-identity) subsection for details. For apps that adopt cookie-based authentication without Identity, see the following [Sign out for cookie-based authentication](#sign-out-for-cookie-based-authentication) subsection.
 
-The reference source for <xref:Microsoft.AspNetCore.Components.Server.RevalidatingServerAuthenticationStateProvider> is located in [`RevalidatingServerAuthenticationStateProvider.cs`](https://github.com/dotnet/aspnetcore/blob/main/src/Components/Server/src/Circuits/RevalidatingServerAuthenticationStateProvider.cs).
-
-[!INCLUDE[](~/includes/aspnetcore-repo-ref-source-links.md)]
 
 #### Sign out for ASP.NET Core Identity
 
-To proactively, completely sign a user off within less than the default 30 minute revalidation interval in apps that adopt ASP.NET Core Identity, use the guidance in this section.
+To force a complete sign-out within less than the default 30-minute revalidation interval in apps that adopt ASP.NET Core Identity, use the guidance in this section.
 
 For Blazor apps that target .NET 8 or later, reduce the default 30 minute <xref:Microsoft.AspNetCore.Components.Server.RevalidatingServerAuthenticationStateProvider.RevalidationInterval%2A> in the `IdentityRevalidatingAuthenticationStateProvider` class (`Components/Account/IdentityRevalidatingAuthenticationStateProvider.cs`). If the app targets .NET earlier than .NET 8, reduce the interval in `RevalidatingIdentityAuthenticationStateProvider`.
 
--- a/aspnetcore/blazor/security/index.md
+++ b/aspnetcore/blazor/security/index.md
@@ -535,21 +535,97 @@ Two additional abstractions participate in managing authentication state:
 
 * <xref:Microsoft.AspNetCore.Components.Server.ServerAuthenticationStateProvider> ([reference source](https://github.com/dotnet/aspnetcore/blob/main/src/Components/Endpoints/src/DependencyInjection/ServerAuthenticationStateProvider.cs)): An <xref:Microsoft.AspNetCore.Components.Authorization.AuthenticationStateProvider> used by the Blazor framework to obtain authentication state from the server.
 
-* <xref:Microsoft.AspNetCore.Components.Server.RevalidatingServerAuthenticationStateProvider> ([reference source](https://github.com/dotnet/aspnetcore/blob/main/src/Components/Server/src/Circuits/RevalidatingServerAuthenticationStateProvider.cs)): A base class for <xref:Microsoft.AspNetCore.Components.Authorization.AuthenticationStateProvider> services used by the Blazor framework to receive an authentication state from the host environment and revalidate it at regular intervals.
+* <xref:Microsoft.AspNetCore.Components.Server.RevalidatingServerAuthenticationStateProvider> ([reference source](https://github.com/dotnet/aspnetcore/blob/main/src/Components/Server/src/Circuits/RevalidatingServerAuthenticationStateProvider.cs)): A base class for <xref:Microsoft.AspNetCore.Components.Authorization.AuthenticationStateProvider> services used by the Blazor framework to receive an authentication state from the host environment and revalidate it at regular intervals, 30 minutes by default.
 
 [!INCLUDE[](~/includes/aspnetcore-repo-ref-source-links.md)]
-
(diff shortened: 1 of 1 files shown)


aspnetcore/blazor/security/index.md
```diff
@@ -535,22 +536,101 @@ Two additional abstractions participate in managing authentication state:
...
+To control the revalidation interval of the underlying authentication cookie in apps that adopt ASP.NET Core Identity, see the following [Sign out for ASP.NET Core Identity](#sign-out-for-aspnet-core-identity) subsection for details. For apps that adopt cookie-based authentication without Identity, see the following [Sign out for cookie-based authentication](#sign-out-for-cookie-based-authentication) subsection.
+
+The reference source for <xref:Microsoft.AspNetCore.Components.Server.RevalidatingServerAuthenticationStateProvider> is located in [`RevalidatingServerAuthenticationStateProvider.cs`](https://github.com/dotnet/aspnetcore/blob/main/src/Components/Server/src/Circuits/RevalidatingServerAuthenticationStateProvider.cs).
 
 [!INCLUDE[](~/includes/aspnetcore-repo-ref-source-links.md)]
```
halter73: This duplicates the reference-source link from line 539 and re-includes the same `aspnetcore-repo-ref-source-links.md` footer note that's already included at line 541, which will render the standard note twice on the page. Suggest removing both.

```suggestion
```

aspnetcore/blazor/security/index.md
```diff
@@ -535,22 +536,101 @@ Two additional abstractions participate in managing authentication state:
...
 
-Server-side Blazor persists user authentication state for the lifetime of the circuit, including across browser tabs. To proactively sign off a user across browser tabs when the user signs out on one tab, you must implement a <xref:Microsoft.AspNetCore.Components.Server.RevalidatingServerAuthenticationStateProvider> ([reference source](https://github.com/dotnet/aspnetcore/blob/main/src/Components/Server/src/Circuits/RevalidatingServerAuthenticationStateProvider.cs)) with a short <xref:Microsoft.AspNetCore.Components.Server.RevalidatingServerAuthenticationStateProvider.RevalidationInterval>.
+To control the revalidation interval of the underlying authentication cookie in apps that adopt ASP.NET Core Identity, see the following [Sign out for ASP.NET Core Identity](#sign-out-for-aspnet-core-identity) subsection for details. For apps that adopt cookie-based authentication without Identity, see the following [Sign out for cookie-based authentication](#sign-out-for-cookie-based-authentication) subsection.
```
halter73: Drop "the underlying authentication cookie" (the Identity subsection is about the security-stamp validator, not the cookie itself) and clarify that this applies to Identity with cookie authentication — the revalidation interval doesn't apply to Identity bearer tokens, which rely on the refresh interval instead.

```suggestion
To control the revalidation interval in apps that adopt ASP.NET Core Identity with cookie authentication, see the following [Sign out for ASP.NET Core Identity](#sign-out-for-aspnet-core-identity) subsection for details. For apps that adopt cookie-based authentication without Identity, see the following [Sign out for cookie-based authentication](#sign-out-for-cookie-based-authentication) subsection.
```

aspnetcore/blazor/security/index.md
```diff
@@ -535,22 +536,101 @@ Two additional abstractions participate in managing authentication state:
...
+
+The reference source for <xref:Microsoft.AspNetCore.Components.Server.RevalidatingServerAuthenticationStateProvider> is located in [`RevalidatingServerAuthenticationStateProvider.cs`](https://github.com/dotnet/aspnetcore/blob/main/src/Components/Server/src/Circuits/RevalidatingServerAuthenticationStateProvider.cs).
 
 [!INCLUDE[](~/includes/aspnetcore-repo-ref-source-links.md)]
 
+#### Sign out for ASP.NET Core Identity
+
+To proactively, completely sign a user off within less than the default 30 minute revalidation interval in apps that adopt ASP.NET Core Identity, use the guidance in this section.
```
halter73: "To proactively, completely sign a user off" reads awkwardly with the stacked adverbs.

```suggestion
To force a complete sign-out within less than the default 30-minute revalidation interval in apps that adopt ASP.NET Core Identity, use the guidance in this section.
```

halter73 (approved):
Copilot wanted to leave a bunch of inline comments for the following:

> Compound-modifier hyphenation. Multiple instances of "30 minute" / "five minute" / "four minute" used as adjectives — per Microsoft style they should be hyphenated when modifying a noun: "30-minute window/interval/revalidation interval", "five-minute interval", "four-minute interval". Affects lines 545 (twice), 549, 551, 568, 570, 572, 576, 591, 593.

Also,

> Repo guidance says to bump `ms.date` to today when more than 50 characters change. This PR adds ~80 lines, so this should be updated.
>
> ```suggestion
> ms.date: 05/28/2026
> ```

I personally don't care too much about this, but I guess it makes sense if that's what the style guide says.

guardrex:
Fair enough on the adjectives. I'll fix them up. I love the way Copilot has to be run multiple times to catch problems. I feel like it should be run two or three times on every PR to get ALL of its suggestions in.

aspnetcore/blazor/security/index.md
guardrex: It's in a different section. However ... yes ... let's delete on the grounds it was already mentioned and make the article shorter.

PR #37246, opened 2026-06-09
Update links to use HTML anchor tags

keenan137:
Changed the links to use HTML anchor tags so that they would open in a new tab instead.


guardrex:
Hello @keenan137 ... Note that we require an issue for most changes before submitting a PR. Anyway ... WRT opening links in new tabs ... it's not something we do in our articles. Devs should right-click a link to open in a new tab.

PR #37292, opened 2026-06-25
Prep only - Minimal Web API tutorial Samples

wadepickett:
This is prep only for a follow up PR.  Moving a copy of all 9.x code samples for the Minimal API tutorial to a 10.x folder.  That way we can see a diff as changes are med across all 8 app samples in the next PR that makes the actual changes.


wadepickett:
@tdykstra, this is prep only for a version update.  No changes in the app samples, they are a copy of the v9 to a v10 folder so we have a diff in the follow up PR.

tdykstra (approved):

